PR #548, opened 2023-03-08
Move Ventura support in scripts/install.sh to scripts/build.sh; fix for node build on ventura

eatonphil:
## Pre-merge checklist

### DIFF
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -8,6 +8,22 @@ else
     ZIG_OS="macos"
 fi
 
+target_set="false"
+case "$*" in
+    *-Dtarget*)
+	target_set="true"
+	;;
+esac
+
+if [ "$target_set" = "false" ]; then
+    # Default to specifying "native-macos" if the target is not provided.
+    # See https://github.com/ziglang/zig/issues/10478 (and note there's not a backport to 0.9.2).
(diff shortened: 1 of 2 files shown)


scripts/build.sh
```diff
@@ -8,6 +8,20 @@ else
...
+	target_set="true"
+	;;
+esac
+
+if [ "$target_set" = "false" ]; then
+    if [ "$(./zig/zig targets | grep '"os": "' | cut -d ":" -f 2 | cut -d '"' -f 2)" = "macos" ]; then
+	# Add target to args
+	set -- "$@" "-Dtarget=native-macos"
+    fi
+fi
```
kprotty: Moving the target check into build.sh makes more sense over requiring it externally. Could the target flag also be removed in `benchmark.sh` now?

eatonphil: Ok, I gave that a shot too!

cb22:
@eatonphil looks like the mac benchmark is broken?

eatonphil:
@kprotty @cb22 I dropped the benchmark.sh fix for now so we can get the build.sh in since there was an issue in benchmark.sh I didn't have time to look at.

cb22 (approved):

PR #549, opened 2023-03-08
vsr: use consistent name for type-parametrized clock

matklad:
That's very minor and I am not even sure if this is a good idea

## Pre-merge checklist


src/vsr/replica.zig
```diff
@@ -76,7 +76,7 @@ pub fn ReplicaType(
         const Self = @This();
 
         const Journal = vsr.JournalType(Self, Storage);
-        const Clock = vsr.Clock(Time);
```
matklad, line 79: These two lines are the whole motivation for the PR, it always seemed odd that they are inconsistent.

sentientwaffle (approved):
:+1: I guess `Clock` predated the "ThingType" naming convention.

PR #543, opened 2023-03-06
Test/Cluster: Don't open replica immediately after crash

sentientwaffle:
Reopening the replica immediately after crash was a shortcut that kept all of the cluster's `Replica`s valid. (Then the "crashed" replicas would just not be `tick`'d until they restarted).

But it makes more sense to `open` when the replica is actually restarted.

This allows `Cluster.restart_replica()` to report `open`'s error, rather than `Cluster.crash_replica()`.

## Pre-merge checklist

Performance:

* [x] I am very sure this PR could not affect performance.

PR #561, opened 2023-03-09
VSR: repair fast-forward

sentientwaffle:
## Summary

Follow-up to https://github.com/tigerbeetledb/tigerbeetle/pull/532.

- Backups can use SV messages to catch up to the latest log if their log intersects with the primary, even if the backup doesn't have the latest checkpoint.
- A new primary will immediately forfeit and start a new view if it is behind the latest checkpoint.
- Both of the above require reworknig DVC/SV generation a bit.
- And SV invariants are different — the body headers may now be disjoint (to allow inclusion of prior checkpoint triggers ~and the oldest repairable op~).

## Pre-merge checklist

Performance:


src/simulator.zig
```diff
@@ -486,31 +501,25 @@ pub const Simulator = struct {
...
-                        simulator.options.replica_crash_stability;
-                },
-                .down => {
-                    if (chance_f64(simulator.random, simulator.options.replica_restart_probability)) {
-                        simulator.cluster.restart_replica(replica.replica) catch unreachable;
-                        log_simulator.debug("{}: restart replica", .{replica.replica});
-                        simulator.replica_stability[replica.replica] =
-                            simulator.options.replica_restart_stability;
-                    }
+                        simulator.options.replica_restart_stability;
```
sentientwaffle, line 522: This is an unrelated fix; it should have been in #543.

Since `restart_replica` actually opens the replica (it used to be `crash_replica`), it also needs to handle the `fault` guard.

(Context: We disable faults if necessary to guarantee that the cluster will eventually recover, to avoid putting a majority of replicas in `status=recovering_head`.)

src/vsr.zig
```diff
@@ -254,10 +254,8 @@ pub const Header = extern struct {
     ///   For `create_accounts` and `create_transfers` this is the batch's highest timestamp.
     /// * A `reply` sets this to the corresponding `prepare`'s timestamp.
     ///   This allows the test workload to verify transfer timeouts.
-    /// * A `do_view_change` sets this to the latest normal view number.
+    /// * A `do_view_change` sets this to its latest log_view number.
```
sentientwaffle: `log_view` used to be called `view_normal`.

src/vsr/replica.zig
```diff
@@ -466,6 +449,29 @@ pub fn ReplicaType(
...
+                    assert(self.log_view >= header.view);
+                    // Typically if there is an op in the WAL higher than the durable headers'
+                    // head op, we must have crashed with a durable SV, not a durable DVC.
+                    //
+                    // However, we could have joined a view and finished some repair before ever
+                    // updating our view_durable.
+                    //
+                    // To avoid special-casing this all over, we pretend this higher op doesn't
+                    // exist. This is safe because the prior view-change didn't complete.
+                    maybe(self.log_view == self.view);
```
sentientwaffle, line 478: Note that this used to be:
```
                    assert(self.log_view == self.view or
                        self.replica == self.primary_index(self.view));
```

https://github.com/tigerbeetledb/tigerbeetle/pull/561/commits/0d31f2a1525b359131fa69f3d5f11f08176b3c81

src/vsr/replica.zig
```diff
@@ -1270,14 +1291,11 @@ pub fn ReplicaType(
...
+            self.view_jump(message.header);
             if (self.ignore_start_view_change_message(message)) return;
 
             assert(!self.solo());
             assert(self.status == .normal or self.status == .view_change);
-            assert(message.header.view >= self.view);
-
-            self.view_jump(message.header);
-
             assert(message.header.view == self.view);
```
sentientwaffle: Explained here: https://github.com/tigerbeetledb/tigerbeetle/commit/d32b3edf04b648b0c654e375f687942ba7c947ac.

src/vsr/replica.zig
```diff
@@ -1400,46 +1450,105 @@ pub fn ReplicaType(
...
+                });
 
-            assert(self.status == .view_change);
-            self.transition_to_normal_from_view_change_status(message.header.view);
-            self.send_prepare_oks_after_view_change();
+                for (self.view_headers.array.constSlice()) |*header| {
+                    log.err("{}: on_start_view: header={}", .{ self.replica, header.* });
+                }
+                unreachable;
+            }
```
sentientwaffle: In this branch, according to the SV, the primary could provide the checkpoint-trigger header that we need in order to repair our journal — but for some reason did not include it in the SV.
This _should_ be unreachable — primaries should always include the trigger header.

But from a Postel's law perspective, maybe we should ignore this (error message then return, to wait for another SV), or even fall back to state transfer.

src/vsr/replica.zig
```diff
@@ -1400,46 +1450,105 @@ pub fn ReplicaType(
...
+                    self.replica,
+                    self.log_view,
+                });
+                return;
+            }
 
             self.view_jump(message.header);
 
-            assert(self.status == .view_change);
+            assert(self.status == .view_change or self.status == .normal);
```
sentientwaffle: Normal-status (lagging) replicas will now request SV messages for their own view, when they need to advance their head op to some future `commit_max`.

src/vsr/replica.zig
```diff
@@ -183,12 +183,13 @@ pub fn ReplicaType(
...
-        /// The op number of the latest committed operation (according to the cluster):
+        /// The op number of the latest committed operation (according to the cluster).
         /// This is the commit number in terms of the VRR paper.
         ///
         /// Invariants:
         /// * `replica.commit_max ≥ replica.commit_min`.
-        /// * never decreases
+        /// * `replica.commit_max = replica.commit_min` for status=normal primary.
+        /// * never decreases.
         commit_max: u64,
```
sentientwaffle, line 197: When reviewing this code, remember that:

- `commit_max` _may exceed the head op_, but
- the head op must _never_ exceed `op_checkpoint_trigger`.

src/vsr/replica.zig
```diff
@@ -1485,14 +1594,7 @@ pub fn ReplicaType(
...
             const op = message.header.op;
             const slot = self.journal.slot_for_op(op);
-            const checksum: ?u128 = switch (message.header.timestamp) {
-                0 => null,
-                1 => message.header.context,
-                else => unreachable,
-            };
-
-            // Only the primary may respond to `request_prepare` messages without a checksum.
-            assert(checksum != null or self.primary_index(self.view) == self.replica);
```
sentientwaffle, line 1511: `request_prepare` with no checksum used to be used during repair to advance the head op. Now we use RSV/SV for that purpose instead, which means `on_request_prepare` doesn't need to handle to "no checksum" case. (Which is good, since it was very risky + I found a couple bugs while ripping it out).

src/vsr/replica.zig
```diff
@@ -2923,26 +3024,70 @@ pub fn ReplicaType(
...
             assert((self.status == .normal) == (command == .start_view));
             assert((self.status == .view_change) == (command == .do_view_change));
-            assert(self.view >= self.log_view);
             assert(self.view >= self.view_durable());
             assert(self.log_view >= self.log_view_durable());
-
-            assert(command != .do_view_change or self.log_view < self.view);
-            assert(command != .start_view or self.log_view == self.view);
+            assert((self.log_view < self.view) == (command == .do_view_change));
+            assert((self.log_view == self.view) == (command == .start_view));
```
sentientwaffle, line 3132: Hopefully more readable. Maybe I should move these into the `switch (command)` but I kind of like having all of them together as preconditions.

src/vsr/replica.zig
```diff
@@ -2923,26 +3024,70 @@ pub fn ReplicaType(
...
 
-            const headers = self.create_view_change_headers();
-            assert(headers.array.len > 0);
-            assert(headers.array.get(0).op == self.op);
+            switch (command) {
+                // The DVC headers are already up to date, either via:
+                // - transition_to_view_change_status(), or
+                // - superblock's vsr_headers (after recovery).
+                .do_view_change => {},
+                .start_view => {
```
sentientwaffle: `ViewChangeHeadersArray.build()` is no more.
DVC and SV headers are now constructed separately.

src/vsr/replica.zig
```diff
@@ -2923,26 +3024,70 @@ pub fn ReplicaType(
...
+                    // The SV includes headers corresponding to:
+                    // - the triggers for preceding checkpoints (as many as we have and can help
+                    //   repair, which is at most 2), and
+                    // - the oldest op that we can help repair.
+                    for ([_]u64{
+                        self.op_checkpoint_trigger() -|
+                            (constants.journal_slot_count - constants.lsm_batch_multiple),
+                        self.op_checkpoint_trigger() -|
+                            (constants.journal_slot_count - constants.lsm_batch_multiple) * 2,
+                        op_min,
```
sentientwaffle: The last (i.e. oldest) header in the SV is special.

`on_start_view` knows that if the oldest header in the SV has op `X`, then the primary can help it repair all ops `≥X`. (Or at least it could when the SV was generated). `on_start_view` uses this to determine when it needs to fall back to state synchronization.

src/vsr/replica.zig
```diff
@@ -2923,26 +3024,70 @@ pub fn ReplicaType(
...
+
+                    // The SV includes headers corresponding to:
+                    // - the triggers for preceding checkpoints (as many as we have and can help
+                    //   repair, which is at most 2), and
+                    // - the oldest op that we can help repair.
+                    for ([_]u64{
+                        self.op_checkpoint_trigger() -|
+                            (constants.journal_slot_count - constants.lsm_batch_multiple),
+                        self.op_checkpoint_trigger() -|
+                            (constants.journal_slot_count - constants.lsm_batch_multiple) * 2,
```
sentientwaffle, line 3169: There are at most 2 checkpoint triggers from the past in our log. We include both of them if possible — these are the headers than `repair()` is looking for when it sends its `request_start_view` when `op < op_repair_max`.

src/vsr/replica.zig
```diff
@@ -2954,56 +3099,21 @@ pub fn ReplicaType(
                 .timestamp = if (command == .do_view_change) self.log_view else 0,
                 .op = self.op,
                 // See the comment in `on_do_view_change()` for why `commit_min` is crucial:
-                .commit = if (command == .do_view_change) self.commit_min else self.commit_max,
+                .commit = self.commit_min,
```
sentientwaffle: Just a simplification. For SV, we are the primary, so `commit_max == commit_min` anyway.

src/vsr/replica.zig
```diff
@@ -3484,29 +3589,19 @@ pub fn ReplicaType(
...
-            assert(message.header.view > 0); // The initial view is already zero.
             assert(self.status != .recovering); // Single node clusters don't have view changes.
             assert(message.header.replica < self.replica_count);
 
             const command: []const u8 = @tagName(message.header.command);
 
-            if (self.status == .recovering_head and message.header.command != .start_view) {
-                log.debug("{}: on_{s}: ignoring (recovering_head)", .{ self.replica, command });
-                return true;
-            }
```
sentientwaffle, line 3511: Refactor; moved into the `do_view_change` branch of the switch.

src/vsr/replica.zig
```diff
@@ -3484,29 +3589,19 @@ pub fn ReplicaType(
...
-
             if (message.header.view < self.view) {
                 log.debug("{}: on_{s}: ignoring (older view)", .{ self.replica, command });
                 return true;
             }
 
-            if (message.header.view == self.view and self.status == .normal) {
-                log.debug("{}: on_{s}: ignoring (view started)", .{ self.replica, command });
-                return true;
-            }
```
sentientwaffle, line 3521: This now only impacts DVCs (moved into the switch) — status=normal is now permitted to handle same-view SVs.

src/vsr/replica.zig
```diff
@@ -3767,23 +3885,15 @@ pub fn ReplicaType(
             return op;
         }
 
-        /// Finds the header with the highest op number in a slice of headers from a replica.
-        /// The headers must be continuous, in reverse order, all connected, and with no gaps.
-        fn op_highest(headers: []const Header) u64 {
```
sentientwaffle, line 3789: Formerly just used for an assertion, but we already have an even stronger version of this via `message_body_as_headers_chain_disjoint`.

src/vsr/replica.zig
```diff
@@ -4032,6 +4116,7 @@ pub fn ReplicaType(
                 self.commit_journal(self.commit_max);
```
sentientwaffle, line 4253: This line creates a sort of loop — committing can trigger a checkpoint, which may advance `op_repair_max`, which would cause us to go through `if (self.op < self.op_repair_max()) {` again.

src/vsr/replica.zig
```diff
@@ -5271,15 +5356,19 @@ pub fn ReplicaType(
         /// `view_durable` and `log_view_durable` will update asynchronously, when their respective
         /// updates are durable.
         fn view_durable_update(self: *Self) void {
-            assert(self.status == .normal or self.status == .view_change);
+            assert(self.status == .normal or self.status == .view_change or
+                (self.status == .recovering and self.solo()));
```
sentientwaffle, line 5506: To allow for the "view-change" (just incrementing its view numbers, not a real view change) that a `replica_count=1` replica does during open.

src/vsr/replica.zig
```diff
@@ -475,6 +481,17 @@ pub fn ReplicaType(
...
+                // Solo replicas must increment their view after recovery.
+                // Otherwise, two different versions of an op could exist within a single view
+                // (the former version truncated as a torn write).
+                self.log_view += 1;
+                self.view += 1;
+                self.view_headers.array.len = 0;
+                self.view_headers.array.appendAssumeCapacity(
+                    self.journal.header_with_op(self.op).?.*,
+                );
+                self.view_durable_update();
```
sentientwaffle, line 507: https://github.com/tigerbeetledb/tigerbeetle/pull/561/commits/b3457197dc2ec00fa3d39146180a269476208e81 for more detail.

src/vsr/replica.zig
```diff
@@ -5298,15 +5387,16 @@ pub fn ReplicaType(
                     .commit_max = self.commit_max,
                     .view = self.view,
                     .log_view = self.log_view,
-                    .headers = self.create_view_change_headers(),
+                    .headers = &self.view_headers,
```
sentientwaffle, line 5536: No need to regenerate the headers here, they are already set up:
- If `log_view == view`, this is an SV — we set these to our suffix during `primary_start_view_as_the_new_primary`.
- If `log_view < view`, this is a DVC — we set these when we transitioned to `status=view_change`.

src/vsr/replica.zig
```diff
@@ -5448,6 +5537,8 @@ pub fn ReplicaType(
             assert(dvcs_canonical.len > 0);
 
             for (dvcs_all.constSlice()) |message| {
+                assert(message.header.op <= self.op_checkpoint_trigger());
```
sentientwaffle: If any of the DVCs had an op belonging to the next checkpoint, then we wouldn't have started this view as the primary, we would have forfeited in favor of that other replica.

src/vsr/replica.zig
```diff
@@ -5496,37 +5578,29 @@ pub fn ReplicaType(
 
             var headers_canonical = DVCQuorum.headers_canonical(self.do_view_change_from_all_replicas);
             const header_head = headers_canonical.next().?;
-            assert(header_head.op >= do_view_change_commit_min_max);
```
sentientwaffle, line 5516: Replaced by `for (dvcs_all.constSlice()) |dvc| assert(header_head.op >= dvc.header.commit);` which is the same check, just inline.

src/vsr/replica.zig
```diff
@@ -5807,11 +5897,64 @@ pub fn ReplicaType(
...
+                        //   installed a single (checkpoint trigger) hook header.
+                        if (op <= commit_max + 1) break;
+                    } else {
+                        // Skip over the gap.
+                    }
+                }
+                assert(self.view_headers.array.len > 0);
+                assert(self.view_headers.array.len <= constants.pipeline_prepare_queue_max);
+                // We could safely include any additional chained headers.
+                // However, that can make view-change bugs harder to trigger.
```
sentientwaffle: Generate the DVC headers!
(If we can't reuse just them from the previous view).

Note that this is much less aggressive than the old strategy — we include exactly as many headers as we need to to be correct, but no more.

src/vsr/replica.zig
```diff
@@ -5975,15 +6120,13 @@ pub fn ReplicaType(
...
+                // When we are recovering_head we can't participate in a view-change anyway.
+                // But there is a chance that the primary is still running, despite the SVC.
+                .start_view_change => if (self.status == .recovering_head) Status.normal else .view_change,
                 else => unreachable,
             };
 
-            switch (self.status) {
-                .normal, .view_change, .recovering_head => {},
-                .recovering => return,
-            }
```
sentientwaffle, line 6002: Refactor; folded into the next `switch`.

src/vsr/replica.zig
```diff
@@ -6299,43 +6461,26 @@ const DVCQuorum = struct {
...
 
-    /// Returns the highest `commit_min` from any DVC (this is not a `commit_max`).
-    fn commit_min_max(dvc_quorum: QuorumMessages, local: struct {
-        replica: u64,
-        commit_min: u64,
-    }) u64 {
-        assert(dvc_quorum[local.replica].?.header.commit <= local.commit_min);
-
-        var commit_min_max_: ?u64 = null;
+    fn commit_max(dvc_quorum: QuorumMessages) u64 {
```
sentientwaffle, line 6656: Because the DVC generation now explicitly uses `commit_max` to anchor (i.e. to determine how many headers can be included) we must be more aggressive about advancing `commit_max` as much as possible.

Of the three ways we can do that:

```
            commit_max_ = std.math.max(commit_max_, dvc_commit_max_connected);
            commit_max_ = std.math.max(commit_max_, dvc_commit_max_pipeline);
            commit_max_ = std.math.max(commit_max_, dvc.header.commit);
```

Only the middle one (`dvc_commit_max_pipeline`) is new; the others were previously handled by `commit_min_max`.

The insight behind `dvc_commit_max_pipeline` is: if you have a pipeline-max of 4 ops, and the head op is 10, then op 6 _must_ be committed (somewhere).

src/vsr/replica.zig
```diff
@@ -6548,22 +6693,6 @@ fn message_body_as_headers_chain_disjoint(message: *const Message) []const Heade
     return message_headers;
 }
 
-/// Asserts that the headers are in descending op order, and there are no gaps or breaks.
-fn message_body_as_headers_chain_consecutive(message: *const Message) []const Header {
```
sentientwaffle, line 6569: SV bodies are no longer consecutive headers.

sentientwaffle:
The simulator found a couple bugs I'll sort out tomorrow.

src/vsr/replica.zig
matklad: :thinking: don't yet fully understand this bit, but this last op feels suspect to me. My understanding is that this'll be the first op to be overwritten by an incoming request. So, it feels a bit useless to send it out --- there's a high chance that we _won't_ actually remeber this op by the time the message reaches the destination.

src/vsr/replica.zig
```diff
@@ -1400,46 +1443,105 @@ pub fn ReplicaType(
...
+                });
+                return;
+            }
 
             self.view_jump(message.header);
 
-            assert(self.status == .view_change);
+            assert(self.status == .view_change or self.status == .normal);
+            assert(self.status == .view_change or self.backup());
+            assert(self.status == .view_change or self.log_view == self.view);
```
matklad: Maybe

```
if (self.status == .normal) assert(self.backup() and self.log_view == self.view);
```

?

Still find `or` asserts hard to read

src/vsr/replica.zig
```diff
@@ -1400,46 +1443,105 @@ pub fn ReplicaType(
...
             assert(message.header.view == self.view);
-            assert(message.header.op == op_highest(message_body_as_headers(message)));
 
-            self.set_op_and_commit_max(message.header.op, message.header.commit, "on_start_view");
-            for (message_body_as_headers_chain_consecutive(message)) |*header| {
-                self.replace_header(header);
+            self.view_headers = vsr.Headers.ViewChangeArray.from_slice(message_body_as_headers_chain_disjoint(message));
```
matklad: Nit, but maybe `self.view_heares.init_from_slice(message_body_as_headers_chain_disjoint(message))` ?

The `=`-based API feels copy-happy.

src/vsr/replica.zig
sentientwaffle: It corresponds to: https://github.com/tigerbeetledb/tigerbeetle/blob/ff1b61f29f4e8c1fc78edc51f88308ead10fd0b8/src/vsr/replica.zig#L1485-L1508 on the `on_start_view` side. That is, the backup uses it to determine when it is definitely too far behind the primary and must transition to state transfer.

> there's a high chance that we won't actually remeber this op by the time the message reaches the destination.

True, when the cluster is healthy. Though after a view-change, there's a good chance that the cluster isn't healthy.

:thinking: If we go to fetch that header and it isn't available, then we will keep trying & maybe even get stuck right now. Maybe `command=commit` heartbeats should include a "minimum repairable message" which backups can use to transition to state transfer.

src/vsr/replica.zig
```diff
@@ -1400,46 +1443,105 @@ pub fn ReplicaType(
...
+                // Use "op ≤ primary_repair_min" instead of "=" to force the logs to intersect.
+                // This will cause a state transfer in the case where:
+                // - the repairing replica has a pristine log up to its op_checkpoint_trigger=X
+                // - the SV headers begin at op X+1
+                // But in exchange it avoids a special-case to update the commit_max and verify
+                // X's header.
+                const repairable = while (op <= primary_repair_min) : (op += 1) {
+                    const slot = self.journal.slot_with_op(op) orelse break false;
+                    if (self.journal.dirty.bit(slot)) break false;
+                } else true;
```
matklad: We are not using `find_latest_headers_break_between` here, because that only checks headers, and here we care about prepares, right? That is, `self.journal.dirty.bit(slot)` tell us whether we have a prepare, right? 

Maybe add this as helper fn in Journal, right next to `find_latest_headers_break_between` (unless I misunderstand the semantics, and the two are unrelated)?

Also, is `self.journal.dirty.bit(slot)` the right condition to use here? I _think_ the dirty bit is set, while we are writing a prepare. But that seems fine? If we have prepare somewhere in memory (in cache?), we shouldn't try to repair it?

src/vsr/replica.zig
matklad: >on_start_view knows that if the oldest header in the SV has op X, then the primary can help it repair all ops ≥X

Ok, I think this is backwards a bit, what we actually get here is:

> if the oldest header in the SV has op X, then it's not possible to repare any op `<x`

So, when a replica receives an SV with oldest op X, and it doesn't have X itself, then it knows it should switch to state sync. If it has op X, it might attempt to request missing prepares.

matklad: >heartbeats should include a "minimum repairable message" which backups can use to transition to state transfer.

I think a backup can infer that from `commit`? if heartbeat commit is at least a wrap ahead of our commit_min, we should switch to sync (give or take a pipeline worth of messages).

sentientwaffle: > I think a backup can infer that from commit? if heartbeat commit is at least a wrap ahead of our commit_min, we should switch to sync (give or take a pipeline worth of messages).

Good point; updated. `repair` now uses `commit_max`  to identify messages that it cannot possibly repair, and trigger state sync.

src/vsr/replica.zig
sentientwaffle: Modified, it now uses `find_latest_headers_break_between`, followed by `find_earliest_dirty_header_between` (which I just added). Also, this logic is now in `repair()` instead of `on_start_view`. `on_start_view` only triggers state transfer if it can't get the checkpoint trigger header it needs.

`dirty` is set during writing, yes. I'm inclined to allow that anyway, even if it could trigger a state transfer, just in the interest of keeping this code simpler. I might change my mind once I have worked on this a bit more though — it is a little rough right now.

sentientwaffle:
@matklad This is ready for review.

I'm not completely happy about putting more logic into `Replica.repair()`, but didn't find a better spot for it.

Also, https://github.com/tigerbeetledb/tigerbeetle/pull/561/commits/a6da03c6f4ccbe5966b6584391e00ffce6fc22a6 is a bug fix, but not directly related to this PR.

src/vsr/replica.zig
```diff
@@ -1270,14 +1293,11 @@ pub fn ReplicaType(
 
         fn on_start_view_change(self: *Self, message: *Message) void {
             assert(message.header.command == .start_view_change);
+            self.view_jump(message.header);
             if (self.ignore_start_view_change_message(message)) return;
```
```suggestion
            if (self.ignore_start_view_change_message(message)) return;
            self.view_jump(message.header);
```

It makes sense to try to ignore a message before reacting to it. 

We probably should change `if (header.view < self.view) return;` to assert and just consistently ignore before jump (I see that we don't have ignore_commit / ignore_prepare yet, but we probably should)

src/vsr/replica.zig
```diff
@@ -1400,46 +1445,92 @@ pub fn ReplicaType(
...
+            if (self.status == .recovering_head) {
+                self.view = message.header.view;
+            } else {
+                self.view_jump(message.header);
+                assert(self.view == message.header.view);
+
+                if (self.status == .normal) {
+                    assert(self.backup());
+                    assert(self.view == self.log_view);
+                }
```
matklad: Can pull these asserts outside of an if? They are correct for both branches.

src/vsr/replica.zig
```diff
@@ -1400,46 +1445,92 @@ pub fn ReplicaType(
...
+                return;
+            }
 
-            self.set_op_and_commit_max(message.header.op, message.header.commit, "on_start_view");
-            for (message_body_as_headers_chain_consecutive(message)) |*header| {
-                self.replace_header(header);
+            if (self.status == .recovering_head) {
+                self.view = message.header.view;
+            } else {
+                self.view_jump(message.header);
```
matklad: Could you add a comment explaining why we view_jump in one case, but not in the other, especially given that we do handle `recovering_head` status in `view_jump`

I also notice that doing a view_jump on start_view would request a redundant start view, perhaps we need something like this in `view_jump`?: 

```zig
                    if (header.command == .start_view) {
                        // No need to request SV if SV is what caused us to increase the view in the first place.
                    } else {
                        self.send_header_to_replica(self.primary_index(header.view), .{
                            .command = .request_start_view,
                            .cluster = self.cluster,
                            .replica = self.replica,
                            .view = header.view,
                        });
                    }
```

src/vsr.zig
```diff
@@ -1189,239 +1183,19 @@ const ViewChangeHeadersArray = struct {
...
-                    // truncated when the SV was installed.
-                    .chain_break => unreachable,
-                }
-                suffix_done = op <= op_dvc_anchor;
-            } else unreachable;
+    pub fn replace(headers: *ViewChangeHeadersArray, slice: []const Header) void {
+        Headers.ViewChangeSlice.verify(slice);
+        headers.array.len = 0;
+        for (slice) |*header| headers.array.appendAssumeCapacity(header.*);
+    }
```
matklad, line 1194: I _think_ we can assert here that the replacement headers are fresher? First element of `slice` should have (view, op) greater than first element of `headers`?

src/vsr/replica.zig
```diff
@@ -1400,46 +1445,92 @@ pub fn ReplicaType(
...
+                // in the WAL, precluding recovery.
+                // TODO State transfer.
+                @panic("unimplemented (state transfer)");
+            }
+
+            self.view_headers.replace(view_headers);
+            assert(self.view_headers.array.get(0).view <= self.view);
+            assert(self.view_headers.array.get(0).op == message.header.op);
+            maybe(self.view_headers.array.get(0).op > self.op_checkpoint_trigger());
+            vsr.Headers.ViewChangeSlice.verify(self.view_headers.array.constSlice());
```
matklad: I think we run `vsr.Headers.ViewChangeSlice.verify` three times on the same headers. I don't super like this, as this makes it less clear _where_ the validation should happen. But maybe that's ok.

src/vsr/replica.zig
```diff
@@ -1400,46 +1445,92 @@ pub fn ReplicaType(
...
+
+            for (view_headers) |*header| {
```
matklad, line 1609:
```suggestion
            for (self.view_headers.array) |*header| {
```

As we installed the headers, let's use the installed version.

src/vsr/replica.zig
matklad: Yeah, I am very confused here. I'd expect to see either `assert(self.status != .recovering_head)` in `fn view_jump`, or to see unconditional `view_jump` here.

src/vsr/replica.zig
```diff
@@ -1371,6 +1392,30 @@ pub fn ReplicaType(
...
+                // as primary.
+                //
+                // This serves a few purposes:
+                // 1. Availability: We pick a primary to minimize the number of WAL repairs, to
+                //    minimize the likelihood of a repair-deadlock.
+                // 2. Optimization: The cluster does not need to wait for a lagging replicas before
+                //    prepares/commits can resume.
+                // 3. Simplify repair: A new primary never needs to fast-forward to a new checkpoint.
+
+                log.debug("{}: on_do_view_change: lagging primary; forfeiting " ++
```
matklad, line 1489:
```suggestion
                log.debug("{}: on_do_view_change: lagging primary; abdicating " ++
```

Let's use `abdicate` as glossary term for when a (would be) primary volonturely decides to step down.

src/vsr/replica.zig
```diff
@@ -1400,46 +1445,92 @@ pub fn ReplicaType(
...
+                // the last checkpoint. If we wrap now, we overwrite un-checkpointed transfers
+                // in the WAL, precluding recovery.
+                // TODO State transfer.
+                @panic("unimplemented (state transfer)");
+            }
+
+            self.view_headers.replace(view_headers);
+            assert(self.view_headers.array.get(0).view <= self.view);
+            assert(self.view_headers.array.get(0).op == message.header.op);
+            maybe(self.view_headers.array.get(0).op > self.op_checkpoint_trigger());
```
matklad: Might assert that the _last_ header should be < checkpoint_trigger

src/vsr/replica.zig
sentientwaffle: > We probably should change if (header.view < self.view) return; to assert and just consistently ignore before jump (I see that we don't have ignore_commit / ignore_prepare yet, but we probably should)

:+1: That works. However,

> It makes sense to try to ignore a message before reacting to it.

In the case of a `view_jump`, sometimes we want to `view_jump` even when we ignore. That is, we often want to ignore the message from a newer view, but still increase our own view (or request a SV).

This is one of those cases:
https://github.com/tigerbeetledb/tigerbeetle/pull/561/commits/d32b3edf04b648b0c654e375f687942ba7c947ac

src/vsr/replica.zig
```diff
@@ -5807,8 +5940,64 @@ pub fn ReplicaType(
...
+            } else if (status_before == .normal or
+                (status_before == .recovering and self.log_view == self.view) or
+                (status_before == .view_change and self.log_view == self.view))
+            {
+                // Either:
+                // - Transition from normal status.
+                // - Recovering from normal status.
+                // - Retired primary that didn't finish repair.
+                assert(self.log_view == self.view);
+
```
matklad, line 6113: Can we `assert(view_new > self.view)` here, to make sure that, if we update `self.view_headers`, we do `self.view_durable_update` below?

src/vsr/replica.zig
```diff
@@ -5807,8 +5940,64 @@ pub fn ReplicaType(
...
+                self.view_headers.array.len = 0;
+
+                const commit_max =
+                    std.math.max(self.commit_max, self.op -| constants.pipeline_prepare_queue_max);
```
matklad: This feels weird a bit, can we just makit it an invariant that op can't be more thatn pipeline ahead of the commit_max?

src/vsr/replica.zig
sentientwaffle: I pulled out `assert(self.view == message.header.view);`.

But these:
```
                    assert(self.backup());
                    assert(self.view == self.log_view);
```
are not correct outside of that branch.

`backup()` asserts that `status == normal`.
And when `status == recovering_head or status == view_change`, `self.view > self.log_view`.

sentientwaffle: Actually, `status == recovering_head` means that `self.view >= self.log_view`.

src/vsr/replica.zig
matklad: > In the case of a view_jump, sometimes we want to view_jump even when we ignore. 

Hm, maybe then we should always view_jump before checking if the message should be ignored? Although I don't like it -- view_jumping on, eg, misdirected messages should be correct, but I'd rather we not do that, to simplify the state space.

src/vsr/replica.zig
sentientwaffle: > I also notice that doing a view_jump on start_view would request a redundant start view, perhaps we need something like this in view_jump?:

`view_jump` in this case is effectively just a `transition_to_view_change_status`.
So it does do unnecessary work, but it is a view-durable update, not a RSV.
(I'm not sure how to solve that without an extra `view_starting` status though, which seems like overkill for a relatively rare case).

I'm going to split `view_jump` into `view_jump_to_normal()` (which requests a SV) and `view_jump_to_view_change()` (which transitions to `view_change`), since right now it is confusing.

matklad: >So it does do unnecessary work, but it is a view-durable update, not a RSV.

Right, I mixed up control flow here, it's ok as is.

src/vsr/replica.zig
sentientwaffle: I agree that we should be consistent about when we `view_jump`.

What about:
- Split `view_jump` into `view_jump_to_normal` (the RSV case) and `view_jump_to_view_change` (the transition-to-view-change case).
- `view_jump_*()` returns `bool` — whether or not `message.header.view != self.view`. If `true`, `return` from the message handler (it is like an extra `ignore_` helper, which as a side effect will try to RSV/transition).
- Change `ignore_*` to assert that `view` matches current view.

That likewise would jump on misdirected messages too, though. :thinking: 
I guess we could filter out `message.header.replica == replica` easily enough in `view_jump_*`.

sentientwaffle: Or `ignore_{message}`/`on_{message}` could `error { Misdirected, ViewOld, ViewNew }!void`, and `on_message` would turn `ViewNew` into a `view_jump`.

src/vsr/replica.zig
```diff
@@ -2923,26 +3005,65 @@ pub fn ReplicaType(
...
+                    }
+                    assert(self.view_headers.array.len + 2 <= constants.view_change_headers_max);
+
+                    // Determine the consecutive extent of the log that we can help recover.
+                    // This may precede op_repair_min if we haven't had a view-change recently.
+                    const range_min = (self.op + 1) -| constants.journal_slot_count;
+                    const range = self.journal.find_latest_headers_break_between(range_min, self.op);
+                    const op_min = if (range) |r| r.op_max + 1 else range_min;
+                    assert(op_min <= op);
+                    assert(op_min <= self.op_repair_min());
```
matklad, line 3161: Hm, this is interesting -- we send out SV, but we actually don't persist that info anywhere. So backups might learn new headers without learning new prepares. I think this means that we need CTRL for liveness even if we don't have disk faults, something I didn't realize before. 

The situation I am thinking is that a primary accepts a request, increments its op, but the prepare subesquently gets dropped by the network, and doesn't make it to disk. Now, if the primary sends out SVs which inlcude it's current `op` and then dies, we are in the situation where:

* the cluster as a whole knows the `op`
* no data file contains a `prepare` for this `op`
* no disk faults occurred.

src/vsr/replica.zig
sentientwaffle: My thinking was:
- We already use "abdicating" for the case where the primary isn't receiving acks and has paused sending heartbeats. But this case is recoverable — the primary may receive some acks and _stop_ abdicating.
- Whereas "forfeiting" is immediate and irrevocable.

The terms aren't great, but that is the reason for the distinction.

src/vsr/replica.zig
matklad: Another option is to kick `view_jump` from within ignore_message. This makes sense from control-flow perspective, but is not great given the current names.

sentientwaffle: Also I like how the current `ignore_` functions have no side effects

src/vsr/replica.zig
```diff
@@ -5357,13 +5451,14 @@ pub fn ReplicaType(
...
+            assert(self.status == .view_change or self.status == .normal or
+                self.status == .recovering_head);
 
             // Uncommitted ops may not survive a view change so we must assert `op` against
             // `commit_max` and not `self.op`. However, committed ops (`commit_max`) must survive:
-            assert(op >= self.commit_max);
-            assert(op >= commit_max);
             assert(op <= self.op_checkpoint_trigger());
+            maybe(op >= self.commit_max);
+            maybe(op >= commit_max);
```
matklad: Hm, the comment probably needs to be updated to match new assertions? Or maybe rather assertions needs to be updated to something like `if (op < self.op) assert(op >= max(commit_max, self.commit_max))`. That is, if we truncate, we never truncate commited stuff.

src/vsr/replica.zig
```diff
@@ -1371,6 +1392,30 @@ pub fn ReplicaType(
             assert(!self.do_view_change_quorum);
             self.do_view_change_quorum = true;
 
+            const dvc_latest = DVCQuorum.dvc_latest(self.do_view_change_from_all_replicas);
+            if (dvc_latest.header.op > self.op_checkpoint_trigger()) {
```
matklad: Hm, this doesn't feel right. `self.op_checkpoint_trigger()` can be very close to our op (eg, the next op might be checkpoint trigger), and it feels like this can reduce availability? 

Let's say we trigger a view change very close to checkpoint. Elected primary adds a couple of request to the pipeline just above the checkpoint, but those requests never commit. Still, if we do another view change on top of that, DVC from that primary would include those two uncommitted requests. 

Now that DVC would prevent any other primary from getting elected in this view, even if it has everything exctept those two uncommitted requests? I think even in the next view no different primary would be elected, because the poisoned primary with two outstanding uncommited requests would re-generate the same poisoned DVC for the new view?

sentientwaffle: If the DVC quorum includes only one replica that is on the next checkpoint, then yes, that replica is the only one that can become the primary.

The view change will still eventually succeed (possibly after cycling a bit).

This is a deliberate tradeoff — the view change is (potentially) slower, but this is much simpler.
The alternative is that the lagging backup needs to catch-up to the latest checkpoint before it can even install the DVC headers to its log. Unless I'm missing something, that would be _really_ complicated.

matklad: So, what worries me is that I think in my example there's no replicas on the next checkpoint? The two prepares which cause the primary to forfeit are uncommitted, and would actually be truncated by CTRL, should this primary decide not to forfeit.

matklad: :thinking: yeah, I think this might actually formally break liveness? Just a single faulty node can prevent progress, if it faults in such a way that it is able to send DVC for every view, but isn't able to become primary itself (which, admitedly, not a too realistic failure mode). 

I think if we _actually_ checked for "is there any one node which checkpointed the next trigger", we would be fine --- if there's one node at a chcekpoint, then at least quorum of nodes would be at a checkpoint eventually. The problem here is that the `dvc_latest` node might not be at a checkpoint, but rather have some uncommitted ops.

sentientwaffle: Ah I see, this is quite tricky.

If there are no ops from the next checkpoint in the DVC quorum, then we are fine. (This is true even if some replicas do have ops from the next checkpoint — if they aren't part of the DVC quorum, they will just be truncated per usual).

If there are ops from the next checkpoint in the DVC quorum, if the (now-retired) primary that prepared them keeps participating in view-changes, then it will eventually succeed.

The problem is that a lagging backup might receive a SV, then use those SV headers to participate in a DVC, even though it has not caught up. If the primary (that sent the SV) is now down, then those ops are part of the DVC quorum, but no replica can become the primary, they will just keep cycling. :thinking:

sentientwaffle: Yeah, I think as-is this is both an availability and a liveness issue.

src/vsr.zig
sentientwaffle: Not necessarily, due to truncation

src/vsr/replica.zig
jorangreef: I don't think we should split `view_jump()` across sub-protocols, but would rather that we keep it all contained.

Also, would not like to increase the dimensionality of the `ignore_()` functions by moving from boolean to error set. The boolean return type is less dimensionality at the call site.

Let's keep this as is.

But it would be good to document why we always check to see if we should view jump, but may then decide thereafter to ignore a message.

sentientwaffle: "less dimensionality at the call site" is the primary the reason I think that `view_jump` should be split.
The caller often needs to know which branch is being taken, but previously that was only implicitly known (based on the context, i.e. the message command).

The other reason is that the two branches don't actually share code.

See:

```zig
        fn view_jump_to_normal(self: *Self, header: *const Header) void {
            assert(header.command == .prepare or header.command == .commit or
                header.command == .start_view_change);

            if (header.command == .start_view_change) {
                // When we are recovering_head we can't participate in a view-change anyway.
                // But there is a chance that the primary is still running and reachable,
                // despite the SVC.
                assert(self.status == .recovering_head);
            }

            if (header.view < self.view) return;

            // We need a start_view from any other replica — don't request it from ourselves.
            if (self.primary_index(header.view) == self.replica) return;

            switch (self.status) {
                // If the transition is to `.normal`, then ignore if for the same view:
                .normal => if (header.view == self.view) return,
                // This is an interesting special case:
                // If the transition is to `.normal` in the same view, then we missed the
                // `start_view` message and we must also consider this a view jump:
                .view_change => {},
                .recovering_head => {},
                .recovering => return,
            }

            if (header.view == self.view) {
                assert(self.status == .view_change or self.status == .recovering_head);

                log.debug("{}: view_jump: waiting to exit view change", .{self.replica});
            } else {
                assert(header.view > self.view);
                assert(self.status == .view_change or self.status == .recovering_head or
                    self.status == .normal);

                log.debug("{}: view_jump: waiting to jump to newer view", .{self.replica});
            }

            // TODO Debounce and decouple this from `on_message()` by moving into `tick()`:
            // (Using request_start_view_message_timeout).
            log.debug("{}: view_jump: requesting start_view message", .{self.replica});
            self.send_header_to_replica(self.primary_index(header.view), .{
                .command = .request_start_view,
                .cluster = self.cluster,
                .replica = self.replica,
                .view = header.view,
            });
        }

        fn view_jump_to_view_change(self: *Self, header: *const Header) void {
            // status=recovering_head cannot transition to view-change — without a head, it cannot
            // safely persist its log suffix headers.
            assert(self.status != .recovering_head);
            assert(header.command == .do_view_change or header.command == .start_view_change);

            if (header.view <= self.view) return;

            if (header.view == self.view + 1) {
                log.debug("{}: view_jump: jumping to view change", .{self.replica});
            } else {
                log.debug("{}: view_jump: jumping to next view change", .{self.replica});
            }
            self.transition_to_view_change_status(header.view);
        }
```

Aside from the conceptual similarity of "response to higher view number" they don't actually have anything in common.

jorangreef: Thanks @sentientwaffle,

By "reducing dimensionality at the call site" I was referring to increasing "the dimensionality of the ignore_() functions by moving from boolean to error set."

i.e. `error { Misdirected, ViewOld, ViewNew }!void` is more dimensionality than `void` or `boolean` so I don't prefer it to what we already have at present.

We used to have error set return types and/or at least optional return types in an early version of VSR and I actively got rid of them to go for `void` (best) or `boolean` return types as far as possible, to reduce branching at the call site.

This experience was the inspiration for this bit in TigerStyle:

> Use simpler function signatures and return types to reduce dimensionality at the call site, the number of branches that need to be handled at the call site, because this dimensionality can also be viral, propagating through the call chain. For example, as a return type, void trumps bool, bool trumps u64, u64 trumps ?u64, and ?u64 trumps !u64.

Regarding `view_jump()`, the rationale was that:

* It's not meant as an imperative (e.g. jump to view change or jump to normal), but rather as a hook/listener to jump if necessary, but in a way that the caller makes its own decisions thereafter, without relying on any return value. This is the least coupling.
* The caller shouldn't need to decide which version of the function to call.
* I also wanted to keep all the view jump logic contained and together, so that one can reason about all the enumerations (given source/target sub-protocols) in one place. Spreading these out increases the risk of a spatial TOCTOU. For example, instead of being able to audit the one function for all enumerations and how they are handled, we would then need to audit all the call sites, that we've called the right variant and handled its return value (if any) correctly.

sentientwaffle: > It's not meant as an imperative (e.g. jump to view change or jump to normal), but rather as a hook/listener to jump if necessary, but in a way that the caller makes its own decisions thereafter, without relying on any return value. This is the least coupling.

Gotcha, that makes sense.

In that case though, should `view_jump` move up into `on_message` so that each message handler does not need to call it separately? And then rather than asserting `header.command`, `view_jump` would just ignore irrelevent commands (i.e. any command without an attached `view`).

jorangreef: Great idea, let's do it.

src/constants.zig
```diff
@@ -197,16 +197,30 @@ comptime {
...
+///   suffix of our journal.
+///   - These help a lagging replica catch up when its `op < commit_max`.
+///   - There are at most two of these in the journal.
+///     (There are 2 immediately after we checkpoint, until we prepare enough to overwrite one).
+pub const view_change_headers_max = view_change_headers_suffix_max + 2;
+
+/// Maximum number of headers from the WAL suffix to include in an SV message.
+/// Must at least cover the full pipeline.
+/// Increasing this reduces likelihood that backups will need to repair their suffix's headers.
+pub const view_change_headers_suffix_max = config.cluster.view_change_headers_suffix_max;
```
matklad: Let's swap the order of the two definitions around, such that: 

* they are defined in natural order, without forward reference
* the CRITICAL comment applies to the suffix_max. My understanding that the suffix condition is critical for safety. The two hooks help with availability, by allowing us to avoid state transfer, but I don't think they are required for safety?

src/vsr/README.md
matklad, line 1: In `Protocol: View-Change`, lets mention forfeiting code path.

src/vsr/replica.zig
```diff
@@ -466,6 +454,38 @@ pub fn ReplicaType(
...
+                    // updating our view_durable.
+                    //
+                    // To avoid special-casing this all over, we pretend this higher op doesn't
+                    // exist. This is safe because the prior view-change didn't complete.
+                    maybe(self.log_view == self.view);
+
+                    if (self.log_view == self.view) op_head = header.op;
+                }
+            }
+            self.op = op_head;
```
matklad, line 483: I wonder if `init` should set `.op = undefined`, so that we assert if we touch it before actually initializing.

src/vsr/replica.zig
```diff
@@ -5992,15 +6308,12 @@ pub fn ReplicaType(
...
+                .do_view_change, .start_view_change => if (self.status == .recovering_head) Status.normal else .view_change,
+                else => return,
             };
 
-            switch (self.status) {
-                .normal, .view_change, .recovering_head => {},
-                .recovering => return,
-            }
-
             if (header.view < self.view) return;
```
matklad: As we no longer have asserts in the `switch` above, let's move this `if` to be the first statement in the function. 

And also maybe add a `if (standby(message.header)) { return }`? 

In general, I have very mixed feelings about moving `view_change` prior to `ignore_message` --- I'd rather that we drop misdirected messages and such. 

At the same time, yes, I also see Joran's point that we don't want to make control flow here more complicated, `void > bool > ? > !`. 

Let's maybe leave as is, but let it cooking in the background, I feel like there's maybe some elegant way to do this.

src/vsr/replica.zig
```diff
@@ -5992,15 +6308,12 @@ pub fn ReplicaType(
         fn view_jump(self: *Self, header: *const Header) void {
             const to: Status = switch (header.command) {
                 .prepare, .commit => .normal,
-                .start_view_change, .do_view_change, .start_view => .view_change,
-                else => unreachable,
+                // When we are recovering_head we can't participate in a view-change anyway.
+                // But there is a chance that the primary is actually running, despite the DVC/SVC.
+                .do_view_change, .start_view_change => if (self.status == .recovering_head) Status.normal else .view_change,
```
matklad, line 6324: Maybe spell `.start_view => return` explicitly with a comment that that case we handle manually in `on_start_view`?

src/vsr/replica.zig
```diff
@@ -1031,6 +1065,22 @@ pub fn ReplicaType(
...
+            assert(self.primary() or self.backup());
+            assert(message.header.replica == self.primary_index(message.header.view));
+            assert(message.header.op > self.op_checkpoint());
+            assert(message.header.op > self.op);
+            assert(message.header.op > self.commit_min);
+
+            defer {
+                if (self.backup()) {
+                    // A prepare may already be committed if requested by repair() so take the max:
+                    self.commit_journal(std.math.max(message.header.commit, self.commit_max));
```
```suggestion
                    self.commit_journal(message.header.commit);
```

I might be missing something, but it seems `self.commit_journal` already have the max logic internally?

src/vsr/replica.zig
```diff
@@ -1114,13 +1150,7 @@ pub fn ReplicaType(
 
             // Wait until we have `f + 1` prepare_ok messages (including ourself) for quorum:
```
matklad: `f+1` phrasing seems to predate flexible quorums?

src/vsr/replica.zig
```diff
@@ -6295,6 +6611,24 @@ const DVCQuorum = struct {
...
+            if (checkpoint_max == null or checkpoint_max.? < dvc_checkpoint) {
+                checkpoint_max = dvc_checkpoint;
+            }
+        }
+        return checkpoint_max.?;
+    }
+
+    /// Returns the highest op of any header from any canonical-view DVC.
+    /// Includes ops that will be truncated due to gaps.
+    fn op_canonical_max(dvc_quorum: QuorumMessages) u64 {
```
matklad: Note that we also have `op_max_canonical`, so we probably should rename at least one of them.

sentientwaffle: Oof I forgot about that, how confusing :facepalm:

sentientwaffle: I inlined `op_canonical_max` — it was probably going to be removed as part of handling corrupt headers in DVCs (future PR) anyhow.

src/vsr/replica.zig
```diff
@@ -6435,7 +6747,7 @@ const DVCQuorum = struct {
 
             var dvcs_log_view: ?u32 = null;
             for (dvcs.constSlice()) |message| {
-                const log_view = @intCast(u32, message.header.timestamp);
+                const log_view = message.header.request;
```
matklad: Might simplify the whole thing to 

```zig
            const dvcs_log_view = dvcs.get(0).header.request;
            for (dvcs.constSlice()) |message| {
                const log_view = message.header.request;
                assert(dvcs_log_view == log_view);
            }
```

src/vsr/replica.zig
matklad: Maybe exapnd the comment a bit then? I actually can't find the comment explainig why `commit_min` is crucial, so it might make sense to duplicate that here, and mention that for start view the point is moot.

src/vsr/replica.zig
```diff
@@ -5511,39 +5699,48 @@ pub fn ReplicaType(
...
-                        // Use `do_view_change_op_head` instead of `replica.op` since the former is
-                        // about to become the new `replica.op`.
-                        header_head.op -| constants.pipeline_prepare_queue_max,
-                    ),
-                ),
-                "on_do_view_change",
-            );
             // "`replica.op` exists" invariant may be broken briefly between set_op_and_commit_max()
             // and replace_header().
             self.replace_header(&header_head);
```
matklad: micro nit, but we may emphasize the extent of broken invariant a bit

```zig
            {
                // "`replica.op` exists" invariant may be broken briefly between set_op_and_commit_max()
                // and replace_header().

                self.set_op_and_commit_max(header_head.op, commit_max, "on_do_view_change");
                assert(self.commit_max <= self.op_checkpoint_trigger());
                assert(self.commit_max <= self.op);

                self.replace_header(&header_head);
                assert(self.journal.header_with_op(self.op) != null);
            }
```

src/vsr/replica.zig
```diff
@@ -5713,10 +5923,54 @@ pub fn ReplicaType(
...
+                "{}: transition_to_normal_from_recovering_head_status: view={}..{} backup",
+                .{
+                    self.replica,
+                    self.view,
+                    view_new,
+                },
+            );
+
+            self.status = .normal;
+            if (self.log_view == view_new and self.view == view_new) {
```
matklad: `and self.view == view_new)` I think this is just always true? 

```
         assert(self.view >= self.log_view);
            assert(self.view <= view_new);
```

src/vsr/replica.zig
sentientwaffle: Good idea, and I can add a `maybe(self.journal.header_with_op(self.op) == null);` too.

src/vsr/replica.zig
```diff
@@ -5713,10 +5923,54 @@ pub fn ReplicaType(
...
+                // Recovering to the same view we lost the head in.
+            } else {
+                self.view = view_new;
+                self.log_view = view_new;
+                self.view_durable_update();
+            }
+
+            assert(!self.prepare_timeout.ticking);
+            assert(!self.primary_abdicate_timeout.ticking);
```
matklad, line 5962: :thinking: can we `assert(self.backup())` here? If not, should we start this timeout if we are primary?

sentientwaffle: :+1: A replica in recovering_head always becomes a backup, never a primary.

matklad (dismissed):
LGTM modulo https://github.com/tigerbeetledb/tigerbeetle/pull/561/files#r1145343531

sentientwaffle:
@matklad Can you re-approve? Your approval was auto-dismissed when I pushed the last commit.

matklad (approved):

PR #562, opened 2023-03-10
Trace number of io completions queued in sqe and number actually submitted to the kernel

jamii:
## Pre-merge checklist

Performance:

    ``` sh
    # benchmark results before
    ...
    
    # benchmark results after
    ...
    ```
OR

### DIFF
--- a/src/io/linux.zig
+++ b/src/io/linux.zig
@@ -22,6 +22,9 @@ pub const IO = struct {
     /// Completions that are ready to have their callbacks run.
     completed: FIFO(Completion) = .{ .name = "io_completed" },
 
+    ios_queued: u64 = 0,
+    ios_in_kernel: u64 = 0,
+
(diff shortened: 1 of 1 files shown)


src/io/linux.zig
```diff
@@ -188,6 +199,18 @@ pub const IO = struct {
                 },
                 else => return err,
             };
+
+            self.ios_in_kernel += self.ios_queued;
```
kprotty: Should this use what's returned by `submit_and_wait()` above? Could be less than `ios_queued`.

jamii: Good point.

If I do:

```
const submitted = ...;
self.ios_queued -= submitted;
self.ios_in_kernel += submitted;
```

Then the `-=` underflows because we're also counting timer submits. Separating them seems hard so I just started counting timers too.

kprotty (approved):

PR #568, opened 2023-03-13
vsr: re-enable clock tests

matklad:
## Pre-merge checklist

Performance:

### DIFF
--- a/src/vsr/clock.zig
+++ b/src/vsr/clock.zig
@@ -1,7 +1,20 @@
 const std = @import("std");
+const builtin = @import("builtin");
 const assert = std.debug.assert;
-const log = std.log.scoped(.clock);
 const fmt = std.fmt;
+
+const log = if (builtin.is_test)
+    // Downgrade `err` to `warn` for tests.
+    // Zig fails any test that does `log.err`, but we want to test those code paths here.
+    struct {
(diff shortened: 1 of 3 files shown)


src/vsr/clock.zig
```diff
@@ -769,7 +786,10 @@ const ClockSimulator = struct {
 };
 
 test "clock: fuzz test" {
-    std.testing.log_level = .err; // silence all clock logs
+    // Silence all clock logs.
+    var level = std.log.Level.err;
```
sentientwaffle: Ah, this was a mistake in the code snippet I sent! It should be 
```zig
var level = std.testing.log_level;
```

matklad: Also can be const. 

I wonder if we should add somthing akin to Rust's `std::mem::replace`?

```zig
    const level = stdx.swap_with(&std.testing.log_level, std.log.Level.err);
    defer std.testing.log_level = level;
```

We have 

```zig
const foo = bar;
bar = baz
```

logic fairly often, and my eyes skip over it automatically, so abstracting it away _might_ reduce trivial bugs. 

OTOH, I feel like the zig's way would be not to wrap such trivial constructs...

sentientwaffle (approved):

PR #546, opened 2023-03-07
vsr: store configuration in the superblock

matklad:
- today, configuration is replica+replica_count
- store both in VSR State (move replica, add count)
- plumbing to specify `--replica-count` on format

## Pre-merge checklist

* [X] I am very sure this PR could not affect performance.

PR #569, opened 2023-03-13
Remove --standby-count and update example for --replica-count

cb22:
It seems `--standby-count` was removed in #546 - also bring `--replica-count` into the CLI help format example

## Pre-merge checklist


matklad (approved):
I wonder what else did I forget... I did grep for usages and updated the docs, but it's odd that I missed cli.zig of all the things!

PR #570, opened 2023-03-14
Docs fixes, add sample programs and a runner to help with ephemeral TigerBeetle instances

eatonphil:
This PR:

* Drops the global flag from the recommended npm install step in the top-level README
* Adds a helper for parsing AccountFlags and TransferFlags in the Go client
* Go, Node and Java all have the basic and two-phase sample programs
  * Runs all of these in CI against real TigerBeetle instance on windows, mac and linux (except for node on windows)
* Adds a `run_with_tb.zig` program to schedule ephemeral TigerBeetle instances while running another program (mainly a TigerBeetle client)
* Breaks the two-phase FAQ entry into its own page and adds some info based on the recent discussion in community Slack

## Pre-merge checklist


docs/usage/two-phase-transfer.md
```diff
@@ -0,0 +1,68 @@
...
+2. Later, the application creates another transfer — with flags either
+marked `post_pending_transfer` or `void_pending_transfer`. The former
+moves all (or part) of the reserved funds to the destination; the
+latter reverts them to the original account.
+
+## `credits_must_not_exceed_debits` and `debits_must_not_exceed_credits`
+
+The pending transfer's amount is reserved in a way that the second
+step in a two-phase transfer will never cause the account's configured
+balance invariants to be broken.
```
sentientwaffle:
```suggestion
step in a two-phase transfer will never cause the accounts' configured
balance invariants to be broken.
```

(The invariants of both the debit and credit accounts are enforced.)

Also might be worth clarifying that the invariants hold whether the pending transfer is posted or voided.

docs/usage/two-phase-transfer.md
```diff
@@ -0,0 +1,68 @@
...
+
+### Optimistic pending transfers
+
+If an account with `debits_must_not_exceed_credits` has
+`credits_posted = 100` and `debits_posted = 70` and a pending transfer
+is started causing the account to have `debits_pending = 50`, the
+*pending* transfer will still succeed.
```
sentientwaffle: This is incorrect; the pending transfer would fail with `exceeds_credits`.
I would consider pending transfers pessimistic.

docs/usage/two-phase-transfer.md
```diff
@@ -0,0 +1,68 @@
+# Two-phase Transfer
+
+This is a reference to the [two-phase commit protocol for distributed
+transactions](https://en.wikipedia.org/wiki/Two-phase_commit_protocol).
```
sentientwaffle: Maybe something like:

> The name "two-phase transfer" is a reference to ...(etc)

To disambiguate since "This" _page_ is a reference to two-phase transfers, not 2PC.

docs/usage/two-phase-transfer.md
eatonphil: Joran suggested the opposite. I didn't run a test program but I will do so now. :D

docs/usage/two-phase-transfer.md
eatonphil: Gotcha, can do

docs/usage/two-phase-transfer.md
eatonphil: Yeah you're right, DJ! The pending transfer failed for me.

.github/workflows/clients-go.yml
```diff
@@ -52,34 +52,9 @@ jobs:
         run: ./zgo.bat test ./...
         working-directory: ./src/clients/go
 
-  build_sample:
```
eatonphil, line 55: The build process for samples of all languages has been unified. They are now not only built but run. For Node, Go, and Java.

build.zig
```diff
@@ -444,10 +458,26 @@ fn link_tracer_backend(
     }
 }
 
+// Zig cross-targets plus Dotnet RID (Runtime Identifier):
```
eatonphil, line 461: Tried to unify target platforms across client libraries.

scripts/build.sh
```diff
@@ -33,5 +32,5 @@ GLOBAL_CACHE_ROOT="$HOME/.cache/zig"
 # Using the {os} directly instead of "native" avoids hitting an abort on macos Ventura.
 # TODO The abort was fixed in 0.10 while we're on 0.9. This script can be removed once we update zig.
 $ZIG_EXE run ./scripts/build_runner.zig -target native-$ZIG_OS --main-pkg-path $BUILD_ROOT \
-    -- $ZIG_EXE $BUILD_ROOT $CACHE_ROOT "$GLOBAL_CACHE_ROOT" \
+    -- $ZIG_EXE $BUILD_ROOT $CACHE_ROOT "$GLOBAL_CACHE_ROOT" $TARGET \
```
eatonphil, line 35: Makes sure that $TARGET is passed in at the right layer, before any other user passed in arguments which might include `--`

docs/FAQ.md
```diff
@@ -82,25 +82,7 @@ data and a mission critical database to store it.
...
-they cannot be used by either the payer or payee.
-
-2. Later (in a future "commit", i.e. a separate request), the
-application creates another transfer — either a post-pending transfer
-or a void-pending transfer. The former moves all (or part) of the
-reserved funds to the destination; the latter reverts them to the
-original account. The pending transfer's amount is reserved in a way
-that this second step will never cause the account's configured
-balance invariants (e.g. debits < credits) to be broken.
+See [Two-phase Transfer](./usage/two-phase-transfer.md).
```
sentientwaffle: What do you think of including a snippet of the original text here as a summary?

e.g.:

```
Single-phase transfers post funds to accounts immediately when they
are created.

In contrast to single-phase transfers, a two-phase transfer moves
funds in stages.

See [Two-phase Transfer](./usage/two-phase-transfer.md) for more detail.

Comment (eatonphil) [reply docs/FAQ.md]: Sure, updated.

File: src/clients/node/README.md [hunk 8]
```diff
@@ -280,6 +290,14 @@ for (let i = 0; i < transfers.length; i += BATCH_SIZE) {
 }
 ```
 
+### Queues and Workers
+
+If you are making requests to TigerBeetle from workers
+pulling jobs from a queue, you can batch requests to
+TigerBeetle by having the worker act on multiple jobs from
+the queue at once rather than one at a time. i.e. pulling
+multiple jobs from the queue rather than just one.
```
eatonphil, line 299: @sentientwaffle calling attention to new sections for queues and workers.

I think we should also put this in other global docs but since this section (expand to see above it) is about Batching, might as well add here now.

sentientwaffle: :+1:  (https://docs.tigerbeetle.com/usage/integration#batching-events)

eatonphil: Nice! Should I just add the same section to the end of that doc, basically?

sentientwaffle: Yes, perfect :+1:

eatonphil: Done, thanks.

docs/usage/two-phase-transfer.md
```diff
@@ -0,0 +1,67 @@
...
+
+1. First, the pending transfer reserves funds. While reserved, they
+cannot be used by either the payer or payee. Only the `debits_pending`
+and `credits_pending` fields are modified on the relevant accounts at
+this point.
+
+2. Later, the application creates another transfer — with flags either
+marked `post_pending_transfer` or `void_pending_transfer`. The former
+moves all (or part) of the reserved funds to the destination; the
+latter reverts them to the original account.
```
sentientwaffle: Since `post_pending_transfer`/`void_pending_transfer` are important and this is the first mention of them in this document, can you change them to link to the flags in the `transfers.md` reference document?

Also (unrelated), what do you think of:

2. Later, the application creates another transfer — with either of the following flags:
    - `post_pending_transfer`: Move all (or part) of the reserved funds to the destination.
    - `void_pending_transfer`: Revert all of the reserved funds to the original account.

docs/usage/two-phase-transfer.md
```diff
@@ -0,0 +1,67 @@
...
+
```
sentientwaffle, line 64: This document should also talk about timeouts.
Admittedly, they aren't implemented yet, but they will be prior to our release.

(If the pending transfer is created with a timeout (which is optional), then if it has not been posted/voided by the time the timeout expires, then the full amount will be voided.)

docs/usage/two-phase-transfer.md
```diff
@@ -0,0 +1,67 @@
...
+funds in stages:
+
+1. First, the pending transfer reserves funds. While reserved, they
+cannot be used by either the payer or payee. Only the `debits_pending`
+and `credits_pending` fields are modified on the relevant accounts at
+this point.
+
+2. Later, the application creates another transfer — with flags either
+marked `post_pending_transfer` or `void_pending_transfer`. The former
+moves all (or part) of the reserved funds to the destination; the
```
sentientwaffle: s/destination/"pending transfer's destination", just to be extra-clear (since this is an intro)

docs/usage/two-phase-transfer.md
```diff
@@ -0,0 +1,67 @@
...
+If an account with `debits_must_not_exceed_credits` has
+`credits_posted = 100` and `debits_posted = 70` and a pending transfer
+is started causing the account to have `debits_pending = 50`, the
+*pending* transfer will fail. It will not wait to get to *posted*
+status to fail.
+
+## All transfers are immutable
+
+To reiterate, completing a two-phase transfer (by either marking it
+void or posted) does not involve modifying a transfer. Instead you
```
sentientwaffle:
```suggestion
void or posted) does not involve modifying the pending transfer. Instead you
```

docs/usage/two-phase-transfer.md
sentientwaffle: Also worth mentioning: A pending transfer can only be "completed" once.
It can't be posted twice, it can't be voided then posted, etc.

docs/usage/two-phase-transfer.md
```diff
@@ -0,0 +1,67 @@
...
+create a new transfer.
+
+The first transfer that is marked pending will always have its pending
+flag set.
+
+The second transfer will have a `post_pending_transfer` or
+`void_pending_transfer` flag set and a `pending_id` field set to the
+ID of the first transfer. The ID of the second transfer will be
```
sentientwaffle: Instead of "ID", what about `id` (with backtick-quotes)? (That makes it more clear that we aren't informally referring to the `pending_id`, which was just mentioned.)

docs/usage/two-phase-transfer.md
eatonphil: great!

docs/usage/two-phase-transfer.md
eatonphil: Good idea! Added.

docs/usage/two-phase-transfer.md
eatonphil: Good idea, adjusted.

docs/usage/two-phase-transfer.md
eatonphil: Accepted locally :+1:

docs/usage/two-phase-transfer.md
eatonphil: Added

docs/usage/two-phase-transfer.md
eatonphil: Added all fields as links.

docs/usage/two-phase-transfer.md
```diff
@@ -5,50 +5,68 @@ protocol for distributed
 transactions](https://en.wikipedia.org/wiki/Two-phase_commit_protocol).
 
 Single-phase transfers post funds to accounts immediately when they
-are created. That is, they affect the `debits_posted` and
-`credits_posted` fields on relevant accounts.
+are created. That is, they affect the
```
sentientwaffle: nit: s/affect/increase (to be a little more specific)

eatonphil: :+1: done, thanks!

docs/usage/two-phase-transfer.md
```diff
@@ -5,50 +5,68 @@ protocol for distributed
...
 funds in stages:
 
 1. First, the pending transfer reserves funds. While reserved, they
-cannot be used by either the payer or payee. Only the `debits_pending`
-and `credits_pending` fields are modified on the relevant accounts at
-this point.
+cannot be used by either the payer or payee. Only the
+[`debits_pending`](../usage/accounts.md#debits_pending) and
+[`credits_pending`](../usage/accounts.md#credits_pending) fields are
+modified on the relevant accounts at this point.
```
sentientwaffle: s/modified/increased
s/relevant/respective

docs/usage/two-phase-transfer.md
```diff
@@ -5,50 +5,68 @@ protocol for distributed
...
 
-2. Later, the application creates another transfer — with flags either
-marked `post_pending_transfer` or `void_pending_transfer`. The former
-moves all (or part) of the reserved funds to the destination; the
-latter reverts them to the original account.
+2. Later, the application creates another transfer — with either of
+   the following flags:
+* [`post_pending_transfer`](../usage/transfers.md#flagspost_pending_transfer): Move all (or part) of the reserved funds to the pending transfer's destination.
+* [`void_pending_transfer`](../usage/transfers.md#flagsvoid_pending_transfer): Revert all of the reserved funds to the original account.
```
sentientwaffle: Ok I know I suggested this change.. but I think we should expand this a bit (either here or in a follow-up paragraph).
Lets make it more clear that the money being moved/reverted is what is was originally added to the pending debits/credits, and that the pending transfer's modification to the pending debits/credits is undone.

eatonphil: Ok, took a stab at that in the paragraph following bullet point 2.

docs/usage/two-phase-transfer.md
eatonphil: Taken, thanks

docs/usage/two-phase-transfer.md
```diff
@@ -17,13 +17,22 @@ funds in stages:
...
 * [`void_pending_transfer`](../reference/transfers.md#flagsvoid_pending_transfer): Revert all of the reserved funds to the original account.
 
+Once the pending transfer is posted or voided, the
+[`debits_pending`](../reference/accounts.md#debits_pending) and
+[`credits_pending`](../reference/accounts.md#credits_pending) fields
+on the relevant accounts are decreased by the
+[`amount`](../reference/transfer.md#amount) of the transfer. And the
+[`debits_posted`](../reference/accounts.md#debits_posted) and
+[`credits_posted`](../reference/accounts.md#credits_posted) fields on
+the relevant accounts are increased by that same amount.
```
sentientwaffle: The pending part is right, but the posted part needs more detail.
Transfers can be partially posted, in which case they may increase but only a subset of the pending amount.
Or when voided, then the posted debits/credits don't change at all.

eatonphil: I added a section on "Posting a subset of pending `amount`" which maybe addresses your first concern. But I don't understand the latter.

sentientwaffle: How about something like this?

When the pending transfer is resolved (posted or voided),
[`debits_pending`](../reference/accounts.md#debits_pending) and
[`credits_pending`](../reference/accounts.md#credits_pending) fields
on the respective accounts are decreased by the
[`amount`](../reference/transfer.md#amount) of the _pending_ transfer.

When the pending transfer is posted,
[`debits_posted`](../reference/accounts.md#debits_posted) and
[`credits_posted`](../reference/accounts.md#credits_posted) fields on
the respective accounts are increased by the _posting_ transfer's `amount`
(which cannot exceed the pending amount).

When the pending transfer is voided,
[`debits_posted`](../reference/accounts.md#debits_posted) and
[`credits_posted`](../reference/accounts.md#credits_posted) are not modified.

eatonphil: Looks good, added a short phrase to clarify posted amount need not equal pending amount either

sentientwaffle (commented):
LGTM aside from the squash.
(I only reviewed the docs changes).

eatonphil:
Thanks DJ!

src/clients/run_with_tb.zig
```diff
@@ -0,0 +1,192 @@
...
+    return file_name.items;
+}
+
+fn free_port() !u16 {
+    var port: u16 = 1025;
+    while (true) {
+        // Address.resolveIp doesn't end up working on Windows. It's
+        // ok because this is always going to be an IPv4 address anyway.
+        const self_addr = try std.net.Address.parseIp4("127.0.0.1", port);
+        var listener = &std.net.StreamServer.init(.{});
```
batiati:
```suggestion
        var listener = std.net.StreamServer.init(.{});
        defer listener.close();
```

.github/workflows/clients-java.yml
```diff
@@ -173,3 +173,48 @@ jobs:
...
+      uses: actions/download-artifact@v3
+      with:
+        name: jar-artifact
+        path: src/clients/java/target/
+        
+    # Grab Zig
+    - run: ./scripts/install_zig.${{ matrix.os == 'windows-latest' && 'bat' || 'sh' }}
+
+    # Grab Maven
+    - run: curl -o maven.zip https://dlcdn.apache.org/maven/maven-3/3.9.1/binaries/apache-maven-3.9.1-bin.zip
```
batiati, line 214: Just checking, `actions/setup-java@v3` already has maven installed.
If we are downloading it here to ensure the specific `3.9.1` version, it'd be good to add a comment explaining that, otherwise we can just remove this step.

batiati (commented):
Just a general comment, in some zig scripts (e.g `integration.zig` and `run_with_tb.zig`) I've noticed points where we might have memory leaks if the caller replaces the arena allocator with something else. Not a big deal in this case, since it's just a zig script.

Should we enforce a convention for explicitly making the allocator argument an `arena` whenever the function expects an arena?

Example:

```diff
-pub fn script_filename(allocator: std.mem.Allocator, parts: []const []const u8) ![]const u8
+pub fn script_filename(arena: *std.heap.ArenaAllocator, parts: []const []const u8) ![]const u8 
```

.github/workflows/clients-java.yml
eatonphil: No it wasn't that. The self hosted runners did not have maven at least. I'm not sure why.

eatonphil:
Great point about the arena signature, @batiati !   I will adjust that.

.github/workflows/clients-java.yml
batiati: Wow, that's really strange, and even more worthy of a comment 👍

src/clients/run_with_tb.zig
eatonphil: There was a close() already below but yeah this simplifies it I suppose. thanks!

.github/workflows/clients-java.yml
eatonphil: Added.

src/clients/run_with_tb.zig
eatonphil: Added in a local commit, pushed.

batiati (dismissed):
LGTM 👍

eatonphil:
Squashed commits, @batiati can I get a reapproval?

batiati (approved):

eatonphil:
Thanks Rafael and DJ for the review!

PR #571, opened 2023-03-14
Bring in self-hosted m1 runners

cb22:
This brings in some self-hosted M1 runners, on macOS 12.6 and 13.2.

Performance:


eatonphil (approved):
Awesome!

PR #573, opened 2023-03-15
vopr: sometimes crash replicas permanently

matklad:
Before, any replica would eventually restart with some probability,
which might mask some a liveness failure.

Now we allow the replica to crash permanently.

* This is parametrized by a conditional probability of death upon crash
* We make sure that there's at least a quorum of alive (but maybe
  crashed) replicas
* We exclude dead replicas from health checks in `fn done()`
* 

## Pre-merge checklist


matklad:
Closing in favor of https://github.com/tigerbeetledb/tigerbeetle/pull/880, as that's more general (allows nasty things like permanent assymetric partitions. )

PR #575, opened 2023-03-16
LSM: ManifestLevel fuzzer

kprotty:
Prototype of a fuzzer which simulates ManifestLevel usage during Manifest under various interleavings. Needs a bit more optimizations and verification, table visibility, and maybe better FuzzOp generation.

## Pre-merge checklist

Performance:

### DIFF
--- a/src/lsm/manifest_level_fuzz.zig
+++ b/src/lsm/manifest_level_fuzz.zig
@@ -367,35 +367,52 @@ fn EnvironmentType(comptime table_count_max: u32, comptime node_size: u32) type 
             var update_amount = amount;
             assert(amount > 0);
 
-            // TODO: use level.iterator() + binary_search(env.tables) instead.
-            // Iterate tables in a random order (until enough tables have been updated):
-            // https://lemire.me/blog/2017/09/18/visiting-all-values-in-an-array-exactly-once-in-random-order/
-            const iter_range = @intCast(u32, env.tables.items.len);
-            var iter_count = iter_range;
-            var index = env.random.uintLessThanBiased(u32, iter_range);
-            while (iter_count > 0) : (iter_count -= 1) {
-                assert(index < iter_range);
-                defer {
-                    index += iter_range - 1;
-                    if (index >= iter_range) index -= iter_range;
-                }
+            // Only update the snapshot_max of those visible to snapshot_latest.
+            // Those visible to env.snapshot would include tables with updated snapshot_max.
+            const snapshots = @as(*const [1]u64, &lsm.snapshot_latest);
 
-                // Update the snapshot_max of tables which are visible to snapshot_latest:
-                const table = &env.tables.items[index];
-                if (table.visible(lsm.snapshot_latest)) {
-                    assert(table.visible(env.snapshot));
+            var it = env.level.iterator(.visible, snapshots, .descending, null);
+            while (it.next()) |level_table| {
+                assert(level_table.visible(env.snapshot));
+                assert(level_table.visible(lsm.snapshot_latest));
 
-                    const snapshot_max = env.snapshot;
-                    env.level.set_snapshot_max(snapshot_max, table);
-                    assert(table.snapshot_max == snapshot_max);
-                    assert(!table.visible(lsm.snapshot_latest));
+                const env_table = env.find_exact(level_table);
+                assert(level_table.equal(env_table));
 
-                    update_amount -= 1;
-                    if (update_amount == 0) break;
-                }
+                env.level.set_snapshot_max(env.snapshot, env_table);
+                assert(env_table.snapshot_max == env.snapshot);
+                assert(!env_table.visible(lsm.snapshot_latest));
+                assert(env_table.visible(env.snapshot));
+
+                update_amount -= 1;
+                if (update_amount == 0) break;
             }
 
             assert(update_amount == 0);
+        }
+
+        fn find_exact(env: *Environment, level_table: *const TableInfo) *TableInfo {
+            const index = binary_search.binary_search_values_raw(
+                Key,
+                TableInfo,
+                key_min_from_table,
+                compare_keys,
+                env.tables.items,
+                level_table.key_min,
+                .{},
+            );
+
+            assert(index < env.tables.items.len);
+            const tables = env.tables.items[index..];
+
+            assert(compare_keys(tables[0].key_min, level_table.key_min) == .eq);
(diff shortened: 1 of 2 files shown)


src/lsm/manifest_level_fuzz.zig
```diff
@@ -0,0 +1,426 @@
+//! Fuzz ManifestLevel insert_table()/set_snapshot_max()/remove_table*()/iterator().
+//!
+//! Invariants checked:
+//! 
+//! - Inserted tables become visible (and invisible, given the snapshot) from the iterator.
+//! - Tables with updated snapshot_max become invisible to future snapshots from the iterator.
+//! - Tables visible to the current/future snapshot can be removed.
+//! - Tables invisible to the current/future/past snapshots can be removed.
+//! - Tables can be inserted, updated, and removed in any order.
```
jorangreef:
```
+ All tables can be updated.
+ All tables can be removed.
```

src/lsm/manifest_level_fuzz.zig
```diff
@@ -0,0 +1,426 @@
+//! Fuzz ManifestLevel insert_table()/set_snapshot_max()/remove_table*()/iterator().
```
jorangreef: We should indicate whether all public methods are covered here, so that the reader doesn't have to work this out.

For example, `All public methods are covered.`

src/lsm/manifest_level_fuzz.zig
jorangreef: We're still missing `Strategy` here, a short sentence or paragraph to tell the story of how the fuzzer maximizes coverage-to-test-code ratio.

src/lsm/manifest_level_fuzz.zig
```diff
@@ -0,0 +1,426 @@
...
+                    Key,
+                    TableInfo,
+                    key_min_from_table,
+                    compare_keys,
+                    env.tables.items,
+                    table.key_max,
+                    .{},
+                );
+
+                // Can't be equal as the tables may not overlap
```
jorangreef: Comments are sentences and should always end with punctuation.

For example, either `.` if they refer clearly to the next line in isolation, or `:` if they refer to multiple lines beneath.

src/lsm/manifest_level_fuzz.zig
```diff
@@ -0,0 +1,426 @@
...
+                    .gt => unreachable,
+                }
+            } else std.math.maxInt(Key);
+
+            const max_delta = @minimum(32, next_key_min - 1 - new_key_min);
+            const new_key_max = new_key_min + env.random.uintAtMostBiased(Key, max_delta);
+
+            return .{
+                .checksum = env.random.int(u128),
+                // zero addresses are used to indicate the table being removed.
```
jorangreef:
```suggestion
                // Zero addresses are used to indicate the table being removed.
```

Proper capitalization tells the reader we're sharp about the little things.

jorangreef (commented):
Nice work @kprotty.

src/lsm/manifest_level_fuzz.zig
```diff
@@ -0,0 +1,497 @@
...
+//!
+//! Invariants:
+//!
+//! - Inserted tables are visible to the current snapshot and snapshot_latest.
+//! - Updated tables are visible to the current snapshot but no longer snapshot_latest.
+//! - Updated tables become completely invisible when the current snapshot is bumped.
+//!
+//! - Tables visible to both snapshot_latest and the current snapshot can be removed.
+//! - Tables invisible to snapshot_latest but still the current snapshot cannot be removed.
+//!     - The current snapshot must be bumped which puts them in the next category:
```
batiati: Not a big deal, but I felt like this `tab` made the comment harder to understand than it actually is. At least for me, I had to read 3 times to be sure that I wasn't missing the relationship between those 3 last invariants.

I guess just a plain list would be clear enough:
```suggestion
//! - The current snapshot must be bumped which puts them in the next category:
```

src/lsm/manifest_level_fuzz.zig
```diff
@@ -0,0 +1,497 @@
...
+                // we either create new ones for future removal or remove invisible ones.
+                const visible_latest = (ctx.inserted - ctx.invisible) - ctx.updated;
+                if (visible_latest == 0) {
+                    if (ctx.inserted < ctx.max_inserted) return ctx.next(.insert_tables);
+                    return ctx.next(.remove_invisible);
+                }
+
+                // Decide if all tables visible ot snapshot_latest should be removed.
+                var amount = ctx.random.intRangeAtMostBiased(usize, 1, visible_latest);
+                if (ctx.random.boolean()) amount = visible_latest;
```
batiati: I got confused by this second assignment on `amount`.

If the intention is to have 50% chance for `visible_latest` and 50% chance for between `1` and `visible_latest`,  I'd suggest this instead:

```zig
 // Decide if all tables visible ot snapshot_latest should be removed.
var amount = if (ctx.random.boolean()) visible_latest else ctx.random.intRangeAtMostBiased(usize, 1, visible_latest);
```

src/lsm/manifest_level_fuzz.zig
```diff
@@ -0,0 +1,497 @@
...
+            .remove_visible => {
+                // If there are no tables visible to snapshot_latest for removal,
+                // we either create new ones for future removal or remove invisible ones.
+                const visible_latest = (ctx.inserted - ctx.invisible) - ctx.updated;
+                if (visible_latest == 0) {
+                    if (ctx.inserted < ctx.max_inserted) return ctx.next(.insert_tables);
+                    return ctx.next(.remove_invisible);
+                }
+
+                // Decide if all tables visible ot snapshot_latest should be removed.
```
batiati, line 212:
```suggestion
                // Decide if all tables visible at snapshot_latest should be removed.
```

batiati (commented):
Neat work @kprotty!

src/lsm/manifest_level_fuzz.zig
kprotty: That was the original plan, but it didn't fit in the 100 column limit. What about a full if statement?
```zig
var amount = if (ctx.random.boolean())
    visible_latest
else
    ctx.random.intRangeAtMostBiased(usize, 1, visible_latest);
```

batiati: Works great!

batiati (approved):

PR #576, opened 2023-03-17
docs: update docker instructions to use the new --replica-count arg

matklad:
Missed that in the original refactor, as I was grepping for `tigerbeetle format`, and `\` got me.

Thanks for heads up about this Phil!

## Pre-merge checklist


eatonphil:
Thanks! I had this fixed in my WIP open PR but it's taking a while so good to have this is now.

eatonphil (approved):

PR #577, opened 2023-03-17
vopr: fix partitioning logic

matklad:
auto_partition_network in PacketSimulator assumes that nodes go first (only nodes can be partitioned, not clients).

However, network allows adding clients&replicas in arbitrary order, and clients were in fact first. Fix and assert that.

## Pre-merge checklist


sentientwaffle (approved):

sentientwaffle:
CI failure is unrelated; see https://github.com/tigerbeetledb/tigerbeetle/pull/574

PR #580, opened 2023-03-20
Clients/Node: fix CreateTransferError definition

zetanumbers:
Yep, that's right


sentientwaffle (approved):
Thanks!

eatonphil:
Note for myself, the CreateTransferError object is an enum so order matters. We should generate this from Zig source code like we do for the Go and Java clients.

PR #582, opened 2023-03-21
Assert that no additional writes are pending during superblock/forest checkpoints.

jamii:
This reveals that there are sometimes writes pending from superblock.view_change during forest.checkpoint eg `zig build simulator_run -- 11414157658330439066`.

## Pre-merge checklist

Performance:

    ``` sh
    # benchmark results before
    ...
    
    # benchmark results after
    ...
    ```
OR

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -246,6 +246,8 @@ pub fn build(b: *std.build.Builder) void {
 
         const simulator = b.addExecutable("simulator", "src/simulator.zig");
         simulator.setTarget(target);
+        // Ensure that we get stack traces even in release builds.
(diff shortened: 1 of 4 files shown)


jamii:
~Looks like the superblock fuzzer alone is enough to repro this - https://github.com/tigerbeetledb/tigerbeetle/actions/runs/4474247522/jobs/7862512167?pr=582~

The superblock fuzzer was a false positive - I had the assert when the op entered the queue rather than when it was executed. 

The original simulator seed is still finding concurrent writes between forest.checkpoint and superblock.view_change, but that is apparently ok.

PR #584, opened 2023-03-21
vsr: standby clocks do not affect the cluster

matklad:
Backstory for implementation: originally I started with adding `standby_count` parameter to the clock, to push the logic of not caring about standbys to the clock itself. But that meant that I need to extend the testing for clocks significantly, as they now have all this extra logic to care about. 

Upon reflection, I find that I can think about the clock as an abstract structure which cares about N time sources, and its up to the call-site to setup those sources as it likes. So I used the trick where active replicas use N sources, sandbys use N+1 source, and standby's repilca_index is remapped to `replica_count`. 

## Pre-merge checklist

Performance:

### DIFF
--- a/src/vsr/clock.zig
+++ b/src/vsr/clock.zig
@@ -160,13 +160,9 @@ pub fn ClockType(comptime Time: type) type {
         /// * the remote replica's `realtime()` timestamp `t1`, and
         /// * our monotonic timestamp `m2` as captured by our `Replica.on_pong()` handler.
         pub fn learn(self: *Self, replica: u8, m0: u64, t1: i64, m2: u64) void {
-            if (self.synchronization_disabled) return;
+            assert(replica != self.replica);
 
-            // A network routing fault must have replayed one of our outbound messages back against us:
-            if (replica == self.replica) {
-                log.warn("{}: learn: replica == self.replica", .{self.replica});
-                return;
-            }
+            if (self.synchronization_disabled) return;
 
             // Our m0 and m2 readings should always be monotonically increasing if not equal.
             // Crucially, it is possible for a very fast network to have m0 == m2, especially where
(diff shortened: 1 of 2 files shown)


src/vsr/clock.zig
```diff
@@ -160,13 +160,9 @@ pub fn ClockType(comptime Time: type) type {
         /// * the remote replica's `realtime()` timestamp `t1`, and
         /// * our monotonic timestamp `m2` as captured by our `Replica.on_pong()` handler.
         pub fn learn(self: *Self, replica: u8, m0: u64, t1: i64, m2: u64) void {
-            if (self.synchronization_disabled) return;
+            assert(replica != self.replica);
```
matklad, line 163: What's our code style for such "defense in depth" situations? 

That is, the current code with `log.warn` is not wrong, but it is effectively dead due to us already doing the check at the call-site. 

My current thinking is that defense in-depth is not great, as checking something _twice_ makes it unclear which check is actually responsible, and which is "just in case". So I like to turn secondary checks into asserts, like I did here.

matklad: (i guess this could be called "offense in depth")

src/vsr/replica.zig
```diff
@@ -571,15 +571,25 @@ pub fn ReplicaType(
...
                 allocator,
-                node_count,
+                replica_count,
                 replica_index,
                 &self.time,
+            ) else Clock.init(
+                allocator,
+                replica_count + 1,
+                replica_count,
+                &self.time,
```
sentientwaffle, line 592: I was considering:
```zig
            self.clock = try Clock.init(
                allocator,
                @boolToInt(replica_index >= replica_count) + replica_count,
                @minimum(replica_index, replica_count),
                &self.time,
            );
```
.. More succinct, but probably harder to tell what's going on

src/vsr/clock.zig
sentientwaffle: I agree that asserting is the right approach.

sentientwaffle (approved):

PR #586, opened 2023-03-21
forest_fuzz: Instead of crashing between ops, crash in the middle of an op.

jamii:
Attempting to reproduce the bug I'm seeing in `jamii-compaction`, with no luck. But still a nice improvement to the fuzzer.

## Pre-merge checklist

Performance:

    ``` sh
    # benchmark results before
    ...
    
    # benchmark results after
    ...
    ```
OR

### DIFF
--- a/src/lsm/forest_fuzz.zig
+++ b/src/lsm/forest_fuzz.zig
@@ -28,7 +28,7 @@ const SuperBlock = vsr.SuperBlockType(Storage);
 
 pub const tigerbeetle_config = @import("../config.zig").configs.test_min;
 
-const FuzzOp = union(enum) {
+const FuzzOpAction = union(enum) {
     // TODO Test secondary index lookups and range queries.
     compact: struct {
         op: u64,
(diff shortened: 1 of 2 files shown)


jamii:
The diff is not as bad as it looks, I just dedented Model.

src/lsm/forest_fuzz.zig
```diff
@@ -535,28 +585,57 @@ pub fn generate_fuzz_ops(random: std.rand.Random, fuzz_op_count: usize) ![]const
...
+        // But this is difficult to test, so for now we'll avoid it.
+        const modifer_tag =
+            if (action == .compact and !action.compact.checkpoint)
+            fuzz.random_enum(
+                random,
+                FuzzOpModifierTag,
+                modifier_distribution,
+            )
+        else
+            FuzzOpModifierTag.normal;
```
sm2n: The indentation on this if branch is weird, but I see that zig fmt thinks that's the right way to do it. I think it'd be more readable with a block even though it's a single expression. What do you think?

sm2n:
I really like the trick with gating the crashes on ticks. On the whole this LGTM.

src/lsm/forest_fuzz.zig
jamii: Oh weird. I'll just push the if up onto the previous line.

sm2n (approved):

PR #587, opened 2023-03-21
Make compaction completion-based instead of readiness-based.

jamii:
This is the first step of https://github.com/tigerbeetledb/tigerbeetle/issues/269. 

Compactions are still concurrent per tree but are uncoupled from the pace of beats. Instead, they start at the beginning of each half-bar, and at the end of each half-bar we wait for them to finish.

As a minor win, the merging itself can now be a really tight loop over large batches of values and has vanished from the perf profile.

```
# before
1223 batches in 94.13 s
load offered = 1000000 tx/s
load accepted = 106236 tx/s
batch latency p00 = 5 ms
batch latency p10 = 31 ms
batch latency p20 = 32 ms
batch latency p30 = 32 ms
batch latency p40 = 32 ms
batch latency p50 = 32 ms
batch latency p60 = 32 ms
batch latency p70 = 32 ms
batch latency p80 = 37 ms
batch latency p90 = 37 ms
batch latency p100 = 3885 ms
transfer latency p00 = 5 ms
transfer latency p10 = 3373 ms
transfer latency p20 = 9848 ms
transfer latency p30 = 16901 ms
transfer latency p40 = 25117 ms
transfer latency p50 = 34823 ms
transfer latency p60 = 44714 ms
transfer latency p70 = 54532 ms
transfer latency p80 = 64492 ms
transfer latency p90 = 74352 ms
transfer latency p100 = 84099 ms

# after
1223 batches in 93.21 s
load offered = 1000000 tx/s
load accepted = 107289 tx/s
batch latency p00 = 5 ms
batch latency p10 = 31 ms
batch latency p20 = 32 ms
batch latency p30 = 32 ms
batch latency p40 = 32 ms
batch latency p50 = 32 ms
batch latency p60 = 37 ms
batch latency p70 = 37 ms
batch latency p80 = 37 ms
batch latency p90 = 37 ms
batch latency p100 = 3690 ms
transfer latency p00 = 5 ms
transfer latency p10 = 3558 ms
transfer latency p20 = 9408 ms
transfer latency p30 = 16161 ms
transfer latency p40 = 24337 ms
transfer latency p50 = 33938 ms
transfer latency p60 = 43993 ms
transfer latency p70 = 54132 ms
transfer latency p80 = 64300 ms
transfer latency p90 = 74033 ms
transfer latency p100 = 83175 ms
```

I tried optionally bypassing the cache in `jamii-bypass-cache` but it's only a small win for extra grid complexity. Worth retrying after merging https://github.com/tigerbeetledb/tigerbeetle/pull/564 though.

Follow-up work:
* Make compactions serial.
* Queue multiple reads inside table and level iterators.
* Queue multiple writes inside compaction.


src/lsm/forest.zig
```diff
@@ -194,6 +201,11 @@ pub fn ForestType(comptime Storage: type, comptime groove_config: anytype) type
         }
 
         pub fn checkpoint(forest: *Forest, callback: Callback) void {
+            if (Storage == @import("../testing/storage.zig").Storage) {
+                // We should have finished all pending io before checkpointing.
+                forest.grid.superblock.storage.assert_no_pending_io(.grid);
+            }
```
sentientwaffle: I think that once we have grid repair implemented that this needs to be removed, since a block could be repaired (a write) or read (via an incoming `request_block` message, to help another replica repair its grid) written during the checkpoint. Maybe add a "TODO Grid repair: remove" if you want to leave it in for now.

(Likewise for the other `assert_no_pending_io` earlier in this file).

src/vsr/superblock.zig
```diff
@@ -1528,6 +1528,14 @@ pub fn SuperBlockType(comptime Storage: type) type {
                 superblock.queue_head = context;
                 log.debug("{s}: started", .{@tagName(context.caller)});
 
+                if (Storage == @import("../testing/storage.zig").Storage) {
+                    // We should have finished all pending io before starting any more.
+                    superblock.storage.assert_no_pending_io(.superblock);
+                    if (context.caller == .checkpoint) {
+                        superblock.storage.assert_no_pending_io(.grid);
```
sentientwaffle, line 1536: Another TODO that this assertion will fail once we have grid repair.

src/testing/storage.zig
```diff
@@ -485,6 +490,41 @@ pub const Storage = struct {
 
         return storage.memory[block_offset..][0..constants.block_size];
     }
+
+    pub fn log_pending_io(storage: *Storage) void {
```
sentientwaffle:
```suggestion
    pub fn log_pending_io(storage: *const Storage) void {
```

src/lsm/range_iterator.zig
```diff
@@ -0,0 +1,182 @@
...
+
+        pub fn init(
+            allocator: mem.Allocator,
+        ) !RangeIterator {
+            _ = allocator; // TODO(jamii) Will need this soon for pipelining.
+            return RangeIterator{
+                .context = undefined,
+                .level_iterator = try LevelIterator.init(allocator),
+                .table_iterator = try TableIterator.init(allocator),
+                .index_block = try alloc_block(allocator),
```
sentientwaffle: Pull these up so we can `errdefer Foo.deinit(...)` them. In practice we'll fail on an allocation error anyway, but it is convention :shrug:

src/lsm/table_iterator.zig
```diff
@@ -6,335 +6,111 @@ const assert = std.debug.assert;
...
-        };
-
         pub fn start(
             it: *TableIterator,
             context: Context,
-            read_done: fn (*TableIterator) void,
         ) void {
-            assert(!it.read_pending);
-            assert(it.index_block_callback == null);
+            assert(it.callback == .none);
```
sentientwaffle: Can we additionally assert
```zig
assert(it.empty());
```

src/lsm/table_iterator.zig
```diff
@@ -6,335 +6,111 @@ const assert = std.debug.assert;
...
-
-            const addresses = Table.index_data_addresses(it.index_block);
-            const checksums = Table.index_data_checksums(it.index_block);
-            const address = addresses[it.data_block_index];
-            const checksum = checksums[it.data_block_index];
-
-            assert(!it.read_pending);
-            it.read_pending = true;
-            it.grid.read_block(on_read, &it.read, address, checksum, .data);
+        pub fn empty(it: *const TableIterator) bool {
```
sentientwaffle: Assert `address.len == checksums.len`.

src/lsm/table_iterator.zig
```diff
@@ -6,335 +6,111 @@ const assert = std.debug.assert;
...
-            const value = values[it.value];
-
-            it.value += 1;
-            if (it.value == values.len) {
-                it.value = 0;
-                it.data_blocks.advance_head();
-            }
+        fn on_next_tick(next_tick: *Grid.NextTick) void {
+            const it = @fieldParentPtr(TableIterator, "next_tick", next_tick);
+            assert(it.callback == .next_tick);
```
sentientwaffle:
```zig
assert(it.empty());
```

src/lsm/table_iterator.zig
```diff
@@ -6,335 +6,111 @@ const assert = std.debug.assert;
...
         const TableIterator = @This();
 
         const Grid = GridType(Storage);
-        const Manifest = ManifestType(Table, Storage);
-        const ValuesRingBuffer = RingBuffer(Table.Value, Table.data.block_value_count_max, .pointer);
+        const BlockPtrConst = Grid.BlockPtrConst;
 
-        const BlockPtrConst = *align(constants.sector_size) const [constants.block_size]u8;
-        const IndexBlockCallback = fn (it: *TableIterator, index_block: BlockPtrConst) void;
+        pub const Callback = fn (it: *TableIterator, data_block: ?BlockPtrConst) void;
```
sentientwaffle: Document that `data_block` is only valid/usable within the callback.

src/lsm/range_iterator.zig
```diff
@@ -0,0 +1,182 @@
...
+                // For range queries, we only want data blocks that might contain the search range.
+                // So filter those out here.
+                it.table_iterator.start(.{
+                    .grid = it.context.grid,
+                    .addresses = Table.index_data_addresses_used(it.index_block),
+                    .checksums = Table.index_data_checksums_used(it.index_block),
+                });
+
+                const on_index = callback.on_index;
+                on_index(it, table_info.?, index_block.?);
```
sentientwaffle: Would it make more sense for `TableIterator` to own the index block, instead of passing it `index_data_addresses_used`/`index_data_checksums_used`? (And for the TODO about filtering, pass a key range). TableIterator's non-ownership of the context's addresses/checksums memory makes me nervous.

src/lsm/range_iterator.zig
```diff
@@ -0,0 +1,182 @@
...
+
+        fn on_level_next(
+            level_iterator: *LevelIterator,
+            table_info: ?TableInfo,
+            index_block: ?BlockPtrConst,
+        ) void {
+            const it = @fieldParentPtr(RangeIterator, "level_iterator", level_iterator);
+            assert(it.callback == .level_next);
+            const callback = it.callback.level_next;
+            it.callback = .none;
```
sentientwaffle: `assert(it.table_iterator.empty());`

src/lsm/range_iterator.zig
```diff
@@ -0,0 +1,182 @@
...
+            }
+        }
+
+        fn on_level_next(
+            level_iterator: *LevelIterator,
+            table_info: ?TableInfo,
+            index_block: ?BlockPtrConst,
+        ) void {
+            const it = @fieldParentPtr(RangeIterator, "level_iterator", level_iterator);
+            assert(it.callback == .level_next);
```
sentientwaffle: Since this assert is immediately followed by `it.callback.level_next` I think it is redundant.

src/lsm/range_iterator.zig
```diff
@@ -0,0 +1,182 @@
...
+        fn table_next(it: *RangeIterator, callback: Callback) void {
+            assert(it.callback == .none);
+
+            it.callback = .{ .table_next = callback };
+            it.table_iterator.next(on_table_next);
+        }
+
+        fn on_table_next(table_iterator: *TableIterator, data_block: ?Grid.BlockPtrConst) void {
+            const it = @fieldParentPtr(RangeIterator, "table_iterator", table_iterator);
+            assert(it.callback == .table_next);
```
sentientwaffle: Redundant because of `it.callback.table_next`.

src/lsm/range_iterator.zig
```diff
@@ -0,0 +1,182 @@
...
+                on_index(it, table_info.?, index_block.?);
+            } else {
+                // If there are no more index blocks, we can just leave `table_iterator` empty.
+            }
+
+            it.table_next(callback);
+        }
+
+        fn table_next(it: *RangeIterator, callback: Callback) void {
+            assert(it.callback == .none);
```
sentientwaffle:
```zig
assert(!it.table_iterator.empty());
```

src/lsm/level_iterator.zig
```diff
@@ -6,327 +6,137 @@ const assert = std.debug.assert;
 const constants = @import("../constants.zig");
 
 const Direction = @import("direction.zig").Direction;
-const TableIteratorType = @import("table_iterator.zig").TableIteratorType;
-const RingBuffer = @import("../ring_buffer.zig").RingBuffer;
 const ManifestType = @import("manifest.zig").ManifestType;
 const GridType = @import("grid.zig").GridType;
 
-/// A LevelIterator sequentially iterates the values of every table in the level.
+/// A LevelIterator iterates the index blocks of every table in a key range.
```
sentientwaffle: Is this not sequential anymore?

src/lsm/range_iterator.zig
```diff
@@ -0,0 +1,182 @@
...
+
+const stdx = @import("../stdx.zig");
+const ManifestType = @import("manifest.zig").ManifestType;
+const alloc_block = @import("grid.zig").alloc_block;
+const GridType = @import("grid.zig").GridType;
+const TableIteratorType = @import("table_iterator.zig").TableIteratorType;
+const LevelIteratorType = @import("level_iterator.zig").LevelIteratorType;
+
+/// A RangeIterator iterates the data blocks of every table in a key range.
+pub fn RangeIteratorType(comptime Table: type, comptime Storage: type) type {
```
sentientwaffle: I keep the iterators mixed up — what do you think of these names instead? 

- s/RangeIterator/LevelDataIterator
- s/TableIterator/TableDataIterator
- s/LevelIterator/LevelIndexIterator

src/lsm/level_iterator.zig
```diff
@@ -6,327 +6,137 @@ const assert = std.debug.assert;
...
-            if (it.manifest_iterated) return null;
-
-            const next_table = it.manifest.next_table(
-                it.level,
-                it.snapshot,
-                it.key_min,
-                it.key_max,
+            // NOTE We must ensure that between calls to `next`,
+            //      no changes are made to the manifest that are visible to `it.context.snapshot`.
+            const next_table_info = it.context.manifest.next_table(
```
sentientwaffle: It would be quite verbose, but to check this we could:
- In `start()`, use a `it.context.manifest.next_table` loop to count how many tables we should get in total. Store it on the iterator. (Or maybe thread the `CompactionRange`'s count through `RangeIterator` to here).
- Keep a counter as state, incrementing it in `on_read`.
- When the iterate completes, assert that the two counters match up.
- (Probably gate under `constants.verify`).

src/lsm/compaction.zig
```diff
@@ -34,616 +34,749 @@ const std = @import("std");
...
+            level_b: u8,
+            range_b: CompactionRange,
+            callback: fn (*Compaction) void,
+        };
 
-        const Status = enum {
-            idle,
-            processing,
-            done,
+        const Side = enum(u1) {
```
sentientwaffle: Would "InputLevel" make sense as a name?

src/lsm/compaction.zig
```diff
@@ -34,616 +34,749 @@ const std = @import("std");
...
-            errdefer iterator_b.deinit(allocator);
+            data_blocks[0] = try alloc_block(allocator);
+            errdefer allocator.free(data_blocks[0]);
+
+            data_blocks[1] = try alloc_block(allocator);
+            errdefer allocator.free(data_blocks[1]);
 
-            var table_builder = try Table.Builder.init(allocator);
+            const table_builder = try Table.Builder.init(allocator);
             errdefer table_builder.deinit(allocator);
```
sentientwaffle: No need to change this, but I think its funny that zig allows `const` here — `const` only works because the `errdefer` can never trigger, so it gets compiled out. If these 2 lines were shifted up, it would require `var` instead of `const`.

src/lsm/compaction.zig
```diff
@@ -34,616 +34,749 @@ const std = @import("std");
...
 const constants = @import("../constants.zig");
 
+const stdx = @import("../stdx.zig");
 const GridType = @import("grid.zig").GridType;
+const alloc_block = @import("grid.zig").alloc_block;
+const TableInfoType = @import("manifest.zig").TableInfoType;
 const ManifestType = @import("manifest.zig").ManifestType;
-const MergeIteratorType = @import("merge_iterator.zig").MergeIteratorType;
 const TableIteratorType = @import("table_iterator.zig").TableIteratorType;
 const LevelIteratorType = @import("level_iterator.zig").LevelIteratorType;
```
sentientwaffle: Don't need to import `LevelIteratorType`.

src/lsm/compaction.zig
```diff
@@ -34,616 +34,749 @@ const std = @import("std");
...
         }
 
-        fn iterator_a_io_callback(iterator_a: *IteratorA) void {
-            const compaction = @fieldParentPtr(Compaction, "iterator_a", iterator_a);
-            compaction.io_finish();
+        fn on_iter_init_a(read: *Grid.Read, index_block: BlockPtrConst) void {
+            const compaction = @fieldParentPtr(Compaction, "read", read);
+            assert(compaction.state == .iter_init_a);
+
+            // `index_block` is only valid for this callback, so copy it's contents.
```
```suggestion
            // `index_block` is only valid for this callback, so copy its contents.
```

src/lsm/compaction.zig
```diff
@@ -34,616 +34,749 @@ const std = @import("std");
...
-                    if (constants.verify) {
-                        // We've finished writing so the block should now be zeroed.
-                        assert(mem.allEqual(u8, block_write.block.*, 0));
+        fn iter_next(compaction: *Compaction, data_block: ?BlockPtrConst) void {
+            assert(compaction.state == .iter_next);
+            const side = compaction.state.iter_next;
+            const index = @enumToInt(side);
+
+            if (data_block) |block| {
+                // `data_block` is only valid for this callback, so copy it's contents.
```
```suggestion
                // `data_block` is only valid for this callback, so copy its contents.
```

src/lsm/compaction.zig
```diff
@@ -34,616 +34,749 @@ const std = @import("std");
...
 
-                    compaction.iterator_a.start(iterator_a_context, iterator_a_io_callback);
-                    compaction.iterator_b.start(iterator_b_context, iterator_b_io_callback);
-                },
-                // Nothing else to start when only moving table_a between levels.
-                .move => {},
+                log.debug(
+                    "Moving table: tree={s}, level_b={}",
+                    .{ compaction.tree_name, context.level_b },
+                );
```
sentientwaffle: Use the `log.debug(tree_name ++ ": ` convention instead (i.e. `tree_name` prefix).

src/lsm/compaction.zig
```diff
@@ -34,616 +34,749 @@ const std = @import("std");
...
+
+                compaction.state = .next_tick;
+                compaction.context.grid.on_next_tick(done_on_next_tick, &compaction.next_tick);
+            } else {
+                // Otherwise, start merging.
+
+                log.debug(
+                    "Merging table: tree={s}, level_b={}",
+                    .{ compaction.tree_name, context.level_b },
+                );
```
sentientwaffle: Use the `log.debug(tree_name ++ ": ` convention instead (i.e. `tree_name` prefix).

src/lsm/compaction.zig
```diff
@@ -34,616 +34,749 @@ const std = @import("std");
...
+            allocator.free(compaction.index_block_a);
             compaction.iterator_b.deinit(allocator);
             compaction.iterator_a.deinit(allocator);
         }
 
+        pub fn reset(compaction: *Compaction) void {
+            assert(compaction.state == .done);
+            compaction.state = .idle;
+            if (compaction.grid_reservation) |grid_reservation| {
+                compaction.context.grid.forfeit(grid_reservation);
```
sentientwaffle, line 204: `compaction.grid_reservation = null;` (not strictly necessary, but as an extra guarantee that the reservation won't be reused.

src/lsm/compaction.zig
```diff
@@ -34,616 +34,749 @@ const std = @import("std");
...
-            level_b: u8,
-            iterator_a_context: IteratorA.Context,
+            context: Context,
         ) void {
-            assert(compaction.status == .idle);
-            assert(compaction.callback == null);
-            assert(compaction.io_pending == 0);
-            assert(!compaction.merge_done and compaction.merge_iterator == null);
-            assert(compaction.tracer_slot == null);
+            assert(compaction.state == .idle);
```
sentientwaffle, line 216: `assert(compaction.grid_reservation == null);` (as an extra sanity-check)

src/lsm/compaction.zig
```diff
@@ -34,616 +34,749 @@ const std = @import("std");
...
         }
 
-        const BlockWriteField = enum { data, filter, index };
-
-        fn io_write_start(compaction: *Compaction, comptime field: BlockWriteField) void {
-            const write_callback = struct {
-                fn callback(write: *Grid.Write) void {
-                    const block_write = @fieldParentPtr(BlockWrite, "write", write);
+        fn iter_next_a(iterator_a: *TableIterator, data_block: ?BlockPtrConst) void {
+            const compaction = @fieldParentPtr(Compaction, "iterator_a", iterator_a);
```
sentientwaffle, line 419: `assert(std.meta.eql(compaction.state, .{ .iter_next = .a }));`

src/lsm/compaction.zig
```diff
@@ -34,616 +34,749 @@ const std = @import("std");
...
+            compaction.iter_next(data_block);
+        }
 
-                    assert(block_write.state == .writing);
-                    block_write.state = .building;
+        fn iter_next_b(iterator_b: *RangeIterator, data_block: ?BlockPtrConst) void {
+            const compaction = @fieldParentPtr(Compaction, "iterator_b", iterator_b);
```
sentientwaffle, line 425: `assert(std.meta.eql(compaction.state, .{ .iter_next = .b }));`

src/lsm/compaction.zig
```diff
@@ -34,616 +34,749 @@ const std = @import("std");
...
-                    compaction.io_write_start(.data);
-                    compaction.io_write_start(.filter);
-                    compaction.io_write_start(.index);
-                },
-                .move => {
-                    const snapshot_max = snapshot_max_for_table_input(compaction.op_min);
-                    const level_b = compaction.level_b;
-                    const level_a = level_b - 1;
+        /// If `values_in[index]` is empty and more values are available, read them.
+        fn iter_check(compaction: *Compaction, side: Side) void {
```
sentientwaffle: Will any index/data block prefetching be the responsibility of the sub-iterators?

src/lsm/range_iterator.zig
```diff
@@ -0,0 +1,182 @@
...
+            level: u8,
+            snapshot: u64,
+            key_min: Key,
+            key_max: Key,
+        };
+
+        pub const IndexCallback = fn (
+            it: *RangeIterator,
+            table_info: ?TableInfo,
+            index_block: ?BlockPtrConst,
```
sentientwaffle: Where are these ever passed as `null`?

src/lsm/range_iterator.zig
sentientwaffle: (Also, could pass `block` here instead of `index_block.?`.)

src/lsm/compaction.zig
```diff
@@ -34,616 +34,749 @@ const std = @import("std");
...
+        fn copy(compaction: *Compaction, side: Side) void {
+            const values_in = compaction.values_in[@enumToInt(side)];
+            const values_out = compaction.table_builder.data_block_values();
+            var values_out_index = compaction.table_builder.value_count;
+
+            const len = @minimum(values_in.len, values_out.len - values_out_index);
+            stdx.copy_disjoint(
+                .exact,
+                Value,
+                values_out[values_out_index .. values_out_index + len],
```
```suggestion
                values_out[values_out_index..][0..len],
```

src/lsm/compaction.zig
```diff
@@ -34,616 +34,749 @@ const std = @import("std");
...
+            compaction.values_in[@enumToInt(side)] = values_in[len..];
+            compaction.table_builder.value_count += @intCast(u32, len);
+        }
 
-            const callback = compaction.callback.?;
-            compaction.callback = null;
-            callback(compaction);
+        fn copy_drop_tombstones(compaction: *Compaction) void {
+            // Copy variables locally to ensure a tight loop.
+            var values_in_a = compaction.values_in[0];
```
```suggestion
            const values_in_a = compaction.values_in[0];
```

src/lsm/compaction.zig
```diff
@@ -34,616 +34,749 @@ const std = @import("std");
...
-            // TODO Implement pacing here by deciding if we should do another compact_tick()
-            // instead of invoking the callback, using compaction.range.table_count as the heuristic.
+            compaction.values_in[@enumToInt(side)] = values_in[len..];
+            compaction.table_builder.value_count += @intCast(u32, len);
+        }
 
-            const callback = compaction.callback.?;
-            compaction.callback = null;
-            callback(compaction);
+        fn copy_drop_tombstones(compaction: *Compaction) void {
```
sentientwaffle, line 535: Assert state is `compacting`.
Also assert `compaction.values_in[1].len == 1`

src/lsm/compaction.zig
```diff
@@ -34,616 +34,749 @@ const std = @import("std");
...
-                &compaction.tracer_slot,
-                .{ .tree_compaction_tick = .{
-                    .tree_name = compaction.tree_name,
-                    .level_b = compaction.level_b,
-                } },
+
+            compaction.write_blocks();
+        }
+
+        fn copy(compaction: *Compaction, side: Side) void {
```
sentientwaffle: Assert state is `compacting`.

src/lsm/compaction.zig
```diff
@@ -34,616 +34,749 @@ const std = @import("std");
...
+            const values_in = compaction.values_in[@enumToInt(side)];
```
sentientwaffle: Assert `values_in.len > 0`.
Assert `compaction.values_in[@enumToInt(side) +% 1].len == 0`.

src/lsm/compaction.zig
```diff
@@ -34,616 +34,749 @@ const std = @import("std");
...
+            const values_out = compaction.table_builder.data_block_values();
+            var values_out_index = compaction.table_builder.value_count;
+
+            const len = @minimum(values_in.len, values_out.len - values_out_index);
```
sentientwaffle, line 522: Can we assert that `len > 0`? i.e. there is at least one free position in the output values.

src/lsm/compaction.zig
```diff
@@ -34,616 +34,749 @@ const std = @import("std");
...
-            // Build up a data block with values merged from the read iterators.
-            // This skips tombstone values if compaction was started with the intent to drop them.
-            while (!compaction.table_builder.data_block_full()) {
-                const value = merge_iterator.pop() orelse break;
-                if (compaction.drop_tombstones and tombstone(&value)) continue;
-                compaction.table_builder.data_block_append(&value);
+        fn merge(compaction: *Compaction) void {
+            // Copy variables locally to ensure a tight loop.
+            var values_in_a = compaction.values_in[0];
+            var values_in_b = compaction.values_in[1];
```
sentientwaffle: Assert these are nonempty.

src/lsm/compaction.zig
```diff
@@ -34,616 +34,749 @@ const std = @import("std");
...
+            remaining,
+            exhausted,
+        },
+        state: union(enum) {
+            idle,
+            compacting,
+            iter_init_a,
+            iter_next: Side,
+            writing: struct {
+                remaining: usize,
```
sentientwaffle: Maybe rename to `pending`, since there is also `input_state.remaining`.

src/lsm/compaction.zig
```diff
@@ -34,616 +34,749 @@ const std = @import("std");
...
-
-            // TODO(Beat Pacing) This should really be where the compaction callback is invoked,
-            // but currently that can occur multiple times per beat.
-            switch (compaction.strategy) {
-                .move => assert(compaction.grid_reservation == null),
-                .merge => {
-                    compaction.grid.forfeit(compaction.grid_reservation.?);
-                    compaction.grid_reservation = null;
+        fn write_finish(compaction: *Compaction) void {
+            assert(compaction.state == .writing);
```
sentientwaffle, line 732: Assert `compaction.state.writing.remaining == 0`

src/lsm/compaction.zig
```diff
@@ -34,616 +34,749 @@ const std = @import("std");
...
+
+                    compaction.state = .next_tick;
+                    compaction.context.grid.on_next_tick(done_on_next_tick, &compaction.next_tick);
                 },
             }
+        }
+
+        fn loop_on_next_tick(next_tick: *Grid.NextTick) void {
+            const compaction = @fieldParentPtr(Compaction, "next_tick", next_tick);
+            assert(compaction.state == .next_tick);
```
sentientwaffle, line 777: Assert `compaction.input_state == .remaining`.

src/lsm/tree.zig
```diff
@@ -564,70 +571,62 @@ pub fn TreeType(comptime TreeTable: type, comptime Storage: type, comptime tree_
...
                 constants.lsm_batch_multiple,
             });
 
-            if (start) tree.manifest.reserve();
+            if (start) {
+                if (constants.verify) {
+                    tree.manifest.verify(tree.compaction_op);
+                }
+
+                tree.manifest.reserve();
```
sentientwaffle: Not necessarily a change for this PR, since it might change as part of the next, but if we pass `reserve()` the number of compactions that will run during this half-bar, `ManifestLog.reserve` can reserve fewer blocks, which will help with space amplification. (Currently we always reserve the worst-case number of blocks).

src/lsm/tree.zig
```diff
@@ -564,70 +571,62 @@ pub fn TreeType(comptime TreeTable: type, comptime Storage: type, comptime tree_
...
             }
 
-            // Try to start compacting the other levels.
-            var it = CompactionTableIterator{ .tree = tree };
-            while (it.next()) |context| {
-                if (start) tree.compact_start_table(op_min, context);
+            if (end) {
+                // At the end of a half-bar, we have to wait for all compactions to finish.
+                tree.compaction_callback = .{ .awaiting = callback };
+                tree.compact_finish_join();
```
sentientwaffle: Send `end` implies `!start`, this might read easier:
```
if (end) {
    ...
    return
}

if (start) {
    ...
}

tree.lookup_snapshot_max = ...
...
```

Or maybe

```zig
if (end) {
    ...
} else {
    if (start) {
         ...
    }
    ...
}
```

src/lsm/table_iterator.zig
jamii: We can at the moment, but we won't be able to when these iterators are used in eg range queries, which might not exhaust their input because they filled their output space first.

src/lsm/tree.zig
jamii: Might be easier in the next one. I'm not sure how easy it will be to calculate the number of compactions up front though. Do all the reservations have to happen at the start of the half-bar? Is anything else writing to the grid outside of compaction?

src/lsm/tree.zig
jamii: I tried something a little different and made it an enum. I don't love the formatting that zig fmt insisted on though.

src/lsm/tree.zig
```diff
@@ -577,56 +577,62 @@ pub fn TreeType(comptime TreeTable: type, comptime Storage: type, comptime tree_
...
-                tree.lookup_snapshot_max = tree.compaction_op + 1;
+            const BeatKind = enum { half_bar_start, half_bar_end, other };
+            const beat_kind = if (compaction_beat == 0 or
+                compaction_beat == half_bar_beat_count)
+                BeatKind.half_bar_start
+            else if (compaction_beat == half_bar_beat_count - 1 or
+                compaction_beat == constants.lsm_batch_multiple - 1)
+                BeatKind.half_bar_end
+            else
+                BeatKind.other;
```
sentientwaffle: Maybe `s/other/half_bar_middle`, both the be more descriptive and to match the other names.

src/lsm/tree.zig
```diff
@@ -577,56 +577,62 @@ pub fn TreeType(comptime TreeTable: type, comptime Storage: type, comptime tree_
...
+                .other => {
+                    tree.lookup_snapshot_max = tree.compaction_op + 1;
 
-                // At the end of other beats, we'll callback on the next tick.
-                tree.compaction_callback = .{ .next_tick = callback };
-                tree.grid.on_next_tick(compact_finish_next_tick, &tree.compaction_next_tick);
+                    // At the end of other beats, we'll callback on the next tick.
+                    tree.compaction_callback = .{ .next_tick = callback };
+                    tree.grid.on_next_tick(compact_finish_next_tick, &tree.compaction_next_tick);
+                },
```
sentientwaffle: Previously, this was run for both "middle" and "start", but now it only runs for "middle". I _think_ it was correct before, but may be missing something.
```
                    tree.lookup_snapshot_max = tree.compaction_op + 1;

                    // At the end of other beats, we'll callback on the next tick.
                    tree.compaction_callback = .{ .next_tick = callback };
                    tree.grid.on_next_tick(compact_finish_next_tick, &tree.compaction_next_tick);
```

src/lsm/tree.zig
sentientwaffle: > Do all the reservations have to happen at the start of the half-bar?

The key here is that if a bunch of different actions need reservations (e.g. multiple concurrent compactions across trees/levels), then those reservations must be invoked in a deterministic order. (Otherwise the storage layout becomes non-deterministic).

> Is anything else writing to the grid outside of compaction?

There's grid repair, but that won't need reservations, since it only writes to already-reserved blocks.
There's state transfer, but that won't run concurrently with compaction.

src/lsm/level_iterator.zig
jamii: Maybe cheaper to put a version number on the manifest and increment it on mutation?

sentientwaffle: Yeah, that works too.

src/lsm/range_iterator.zig
jamii: My thinking was that when I add pipelining, the pipeline object will own all the blocks and will hand out pointers that remain valid until you call read again.

src/lsm/range_iterator.zig
jamii: We can reach this with an empty table_iterator. Then we call `on_data(it, null)` in `on_table_next`.

src/lsm/range_iterator.zig
jamii: :+1:

src/lsm/compaction.zig
jamii: It's going to be a bit weird because the iterators take Table at comptime, but we want only have one compaction worth of allocated state. So I'm thinking about making ReadPipeline and WritePipeline objects that own blocks and have read methods. Then those get passed to LevelIndexIterator and TabelDataIterator and they use them to prefetch.

src/lsm/range_iterator.zig
jamii: Oh, good point.

src/lsm/tree.zig
jamii: :flushed:

sentientwaffle (dismissed):
:+1:

sentientwaffle (approved):

PR #591, opened 2023-03-22
vsr: assert that op order is descending

matklad:
## Pre-merge checklist

Performance:


sentientwaffle (approved):
Good catch!

PR #593, opened 2023-03-22
WIP query design doc

jamii:


docs/design/queries.md
```diff
@@ -0,0 +1,177 @@
...
+where transfer.pending_id = ?
+```
+
+``` sql
+select transfer.pending_id
+where transfer.id = ?
+```
+
+
+TODO other uses.
```
sentientwaffle, line 89: A couple other important use cases:

- Get accounts/transfer(s) with `id` matching a given prefix.
- Get accounts/transfers with `user_data` matching a given prefix.

matklad:
Closing, but cc @batiati. I am not sure whether we have any internal docs for scans infrastruture, but it might be a good idea to add those!

PR #594, opened 2023-03-23
LSM: `TableMutable` benchmark

### LINKED ISSUE #550: Sorting table_mutable starves IO
Most of the time during compaction we're keeping the io queue pretty full. But when one compaction finishes and we sort the mutable table, this occupies the cpu for a long time without yielding. As a result, all the other compactions are unable to schedule more io and the queue empties.

image

kprotty:
In an attempt to resolve #550, this implements a separate version of TableMutable which uses a self-balancing binary-search-tree to keep sort() times low. It also implements `benchmark_table_mutable` to see how it compares against the reference implementation (atm, scales great until `table_count_max ~= 32k` using an [AA-Tree](https://en.wikipedia.org/wiki/AA_tree))

## Pre-merge checklist

Performance:

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -371,6 +371,11 @@ pub fn build(b: *std.build.Builder) void {
             .file = "src/lsm/segmented_array_benchmark.zig",
             .description = "SegmentedArray search",
         },
+        .{
+            .name = "benchmark_table_mutable",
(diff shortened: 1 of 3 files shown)


kprotty:
Superseded by #606

PR #598, opened 2023-03-24
POC: mix config into root's prepare hash

matklad:
Mostly just got curious what  will happen.

On `main`, if I start three nodes, one of which thinks there are only two nodes, the cluster "works" and can silently lead to corruption

With this PR, the cluster correctly shuts itself down.

Don't think it's worth to merge this as is though, as I think we'll move the point where we know configuration to `open`, rather than `format`


matklad:
Specifically, we die in this assert: 

https://github.com/tigerbeetledb/tigerbeetle/blob/47a78cb987191d79ab22f517251b02d702b89895/src/vsr/replica.zig#L6803-L6804

PR #601, opened 2023-03-26
Leave docs_website alone

eatonphil:
I had to move the docs build system into Docker to make the builds reproducible but not everyone has Docker permissions set up correctly so the files built inside the docs builder come out owned by `root` so you need a `sudo rm -rf docs_website` but I don't like putting `sudo` into scripts since not everyone has `sudo`. So this just stops trying to remove it and adds it to gitignore. You can remove it if you want.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,5 @@ simulator
 zig/
 zig-cache/
 zig-out/
+
+docs_website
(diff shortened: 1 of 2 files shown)


batiati (approved):
👍

PR #602, opened 2023-03-27
vsr: remove dead code

matklad:
## Pre-merge checklist

### DIFF
--- a/src/vsr/replica.zig
+++ b/src/vsr/replica.zig
@@ -324,6 +324,7 @@ pub fn ReplicaType(
         /// Seeded with the replica's index number.
         prng: std.rand.DefaultPrng,
 
+        /// Used by `Cluster` in the simulator.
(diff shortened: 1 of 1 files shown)


src/vsr/replica.zig
```diff
@@ -324,7 +324,6 @@ pub fn ReplicaType(
         /// Seeded with the replica's index number.
         prng: std.rand.DefaultPrng,
 
-        context: ?*anyopaque = null,
```
sentientwaffle, line 327: I think `cluster.zig` uses this as an anchor for `@fieldParentPtr` to turn a `*Replica` back into a `*Cluster`.

matklad: Indeed, added a comment instead (it's _not_ `@fieldParentPtr`, so that's why we need context).

sentientwaffle (approved):

PR #603, opened 2023-03-27
Simulator: Log detailed state

sentientwaffle:
See `src/testing/README.md` for example + explanation.

Motivation: Make the non-debug vopr logs expose useful cluster state (rather than only commits).
(The message/state checksums that part of the existing vopr output are now omitted — they are verbose, and on occasions where they are useful, they can be easily found in the debug logs.)

Previously we logged "maximum commit op by any replica", but now log "commit_max per replica" — the former can be determined from the latter.

## Pre-merge checklist

Performance:

### DIFF
--- a/src/testing/README.md
+++ b/src/testing/README.md
@@ -1,0 +1,60 @@
+## VOPR Output
+
+### Columns
+
+1. Replica index.
+2. Event:
+    - `$`: crash
+    - `^`: recover
+    - ` `: commit
+    - `[`: checkpoint start
+    - `]`: checkpoint done
+3. Role (according to the replica itself):
+    - `/`: primary
+    - `\`: backup
+    - `|`: standby
+    - `#`: (crashed)
+4. Status:
+    - The column (e.g. `.   ` vs `   .`) corresponds to the replica index. (This can help identify events' replicas at a quick glance.)
+    - The symbol indicates the `replica.status`.
+    - `.`: `normal`
+    - `v`: `view_change`
+    - `r`: `recovering`
+    - `h`: `recovering_head`
+5. View: e.g. `74V` indicates `replica.view=74`.
+6. Commit: e.g. `150/160C` indicates `replica.commit_min=150` and `replica.commit_max=160`.
+7. Journal op: e.g. `87:150Jo` indicates that the minimum op in the journal is `87` and the maximum is `150`.
+8. Journal dirty: `0Jd` indicates that the journal has 0 dirty headers.
+9. Journal faulty: `0Jf` indicates that the journal has 0 faulty headers.
+10. WAL prepare ops: e.g. `85:149Wo` indicates that the op of the oldest prepare in the WAL is `85` and the op of the newest prepare in the WAL is `149`.
+11. Grid blocks free: e.g. `122695Gf` indicates that the grid has `122695` blocks free.
+12. Pipeline prepares (primary-only): e.g. `2/4Pp` indicates that the primary's pipeline has 2 prepares queued, out of a capacity of 4.
+13. Pipeline requests (primary-only): e.g. `0/3Pq` indicates that the primary's pipeline has 0 requests queued, out of a capacity of 3.
+
+### Example
+
+```
+ 1 2 3 4------- 5-- 6-------  7-------  8--  9--  10------  11------   12---  13---
+
+ 0   \ .        74V 146/146C  86:149Jo  2Jd  0Jf  86:149Wo  122695Gf
+ 4   /     .    74V 147/147C  86:149Jo  0Jd  0Jf  86:149Wo  122695Gf   3/4Pp  0/3Pq
+ 0   \ .        74V 147/147C  86:149Jo  2Jd  0Jf  86:149Wo  122695Gf
+ 3   \    .     74V 147/147C  84:147Jo  0Jd  0Jf  84:147Wo  122695Gf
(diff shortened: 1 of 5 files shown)


src/testing/README.md
```diff
@@ -0,0 +1,60 @@
...
+    - `$`: crash
+    - `^`: recover
+    - ` `: commit
+    - `[`: checkpoint start
+    - `]`: checkpoint done
+3. Role (according to the replica itself):
+    - `/`: primary
+    - `\`: backup
+    - `|`: standby
+    - `#`: (crashed)
```
matklad, line 16: These looks a bit mysterious. Maybe `P b s c` / `P B S X` ? Ahhh, I think we want to avoid alphanumerics to not clash with the hash, right?

matklad: Ah, but we don't have hashes

src/testing/README.md
```diff
@@ -0,0 +1,60 @@
...
+    - `h`: `recovering_head`
+5. View: e.g. `74V` indicates `replica.view=74`.
+6. Commit: e.g. `150/160C` indicates `replica.commit_min=150` and `replica.commit_max=160`.
+7. Journal op: e.g. `87:150Jo` indicates that the minimum op in the journal is `87` and the maximum is `150`.
+8. Journal dirty: `0Jd` indicates that the journal has 0 dirty headers.
+9. Journal faulty: `0Jf` indicates that the journal has 0 faulty headers.
+10. WAL prepare ops: e.g. `85:149Wo` indicates that the op of the oldest prepare in the WAL is `85` and the op of the newest prepare in the WAL is `149`.
+11. Grid blocks free: e.g. `122695Gf` indicates that the grid has `122695` blocks free.
+12. Pipeline prepares (primary-only): e.g. `2/4Pp` indicates that the primary's pipeline has 2 prepares queued, out of a capacity of 4.
+13. Pipeline requests (primary-only): e.g. `0/3Pq` indicates that the primary's pipeline has 0 requests queued, out of a capacity of 3.
```
matklad, line 32: Do we need capacities here? they are constant, right?

src/testing/README.md
```diff
@@ -0,0 +1,60 @@
...
+    - The symbol indicates the `replica.status`.
+    - `.`: `normal`
+    - `v`: `view_change`
+    - `r`: `recovering`
+    - `h`: `recovering_head`
+5. View: e.g. `74V` indicates `replica.view=74`.
+6. Commit: e.g. `150/160C` indicates `replica.commit_min=150` and `replica.commit_max=160`.
+7. Journal op: e.g. `87:150Jo` indicates that the minimum op in the journal is `87` and the maximum is `150`.
+8. Journal dirty: `0Jd` indicates that the journal has 0 dirty headers.
+9. Journal faulty: `0Jf` indicates that the journal has 0 faulty headers.
```
matklad, line 28: Maybe merge the two into `0/2J`?

src/testing/cluster.zig
```diff
@@ -450,5 +470,105 @@ pub fn ClusterType(comptime StateMachineType: fn (comptime Storage: type, compti
...
             std.os.exit(@enumToInt(failure));
         }
+
+        fn log_replica(cluster: *const Self, event: enum {
+            crash,
+            recover,
+            commit,
+            checkpoint_start,
+            checkpoint_done,
+        }, replica_index: u8) void {
```
matklad: Somewhat confusing layout, maybe

```zig
        fn log_replica(
            cluster: *const Self,
            event: enum { crash, recover, commit, checkpoint_start, checkpoint_done },
            replica_index: u8,
        ) void {
```

matklad: Or even

```zig
        fn log_replica(
            cluster: *const Self,
            event: enum(u8) {
                crash = '$',
                recover = '^',
                commit = ' ',
                checkpoint_start = '[',
                checkpoint_done = ']',
            },
            replica_index: u8,
        ) void {
```

src/testing/README.md
matklad, line 1: Should we add a column for checkpoints? previous checkpoints and #ops since last checkpoint?

matklad (dismissed):

src/testing/README.md
matklad: looking at the actual output, the aesthetics of \|/ is just too good :D

src/testing/README.md
matklad: Ah, we have [] for checkpoints!

matklad:
removed from the queu just in case you want to apply some formatting suggestions, but this LGTM as is

src/testing/README.md
sentientwaffle: Yeah, I wanted something that is easy to spot at a glance — even if it a little mysterious initially, I think we use the simulator often enough that it is learnable.

Mnemonics:
- `|` standby is standing straight up
- `\` backup backslash (leaning back)
- `/` primary forward-slash (leaning ahead)

src/testing/README.md
sentientwaffle: They are comptime-constant, but ideally we will eventually fuzz different capacities for these (the prepare queue and request queue).

e.g. the request queue is entirely optional — I sometimes manually fuzz it at zero.
And increasing/decreasing the prepare queue size can make different behaviors more/less likely.
In particular, there is different behavior when `prepare queue < batch multiple` vs `prepare queue > batch multiple`.

matklad (approved):

PR #605, opened 2023-03-27
VSR: Explicit blank/faulty DVC headers; DVC-nack

sentientwaffle:
Change DVC messages and protocol:
- Previously, gaps were implicit — the headers would have ops 8,7,5,4 (implicit gap at op=6).
- Now, gaps are explicit ("blanks"). And these are now distinct from "faults" — gaps caused by WAL corruption.
- The former (blanks) count as nacks; the latter (faults) do not count as nacks.
- If the new primary collects a view-change quorum of DVCs, and 1+ headers are both missing and have insufficient nacks,
  then it will wait for even more DVC messages before it will complete the view-change.
  (At this point it warn-logs `quorum received, awaiting repair`.
- If the new primary collects DVCs from every replica in the cluster, but it still isn't enough
  (i.e. too much corruption to nack, but no one has the header),
  then we error-log `quorum received, deadlocked`.

There is a minor change the SVs as well:
- Since an SV can be "converted" to a DVC, it needs to include an extra header from before.
  (This is why `view_change_headers_suffix_max` now has a +1 in `config.zig`).
- Note that DVCs cannot have "hook" headers (like SVs have), since that would leave implicit gaps.
  When an SV is converted to a DVC, the hook headers are dropped.

## Pre-merge checklist

Performance:

### DIFF
--- a/src/constants.zig
+++ b/src/constants.zig
@@ -202,24 +202,34 @@ comptime {
 /// - We must provide enough headers to cover all uncommitted headers so that the new
 ///   primary (if we are in a view change) can decide whether to discard uncommitted headers
 ///   that cannot be repaired because they are gaps. See DVCQuorum for more detail.
+/// - +1 to leave room for commit_max, in case a backup converts the SV to a DVC.
 pub const view_change_headers_suffix_max = config.cluster.view_change_headers_suffix_max;
 
 /// The number of prepare headers to include in the body of a DVC/SV.
 ///
+/// start_view:
+///
 /// - We must include all uncommitted headers.
+/// - +1 We must include the highest cluster-committed header (in case the SV is converted to a DVC
+///   by the backup). (This is part of view_change_headers_suffix_max).
 /// - +2: We must provide the header corresponding to each checkpoint-trigger in the intact
 ///   suffix of our journal.
 ///   - These help a lagging replica catch up when its `op < commit_max`.
 ///   - There are at most two of these in the journal.
 ///     (There are 2 immediately after we checkpoint, until we prepare enough to overwrite one).
+///
+/// do_view_change:
+///
+/// - We must include all uncommitted headers.
+/// - +1 We must include the highest cluster-committed header, so that the new primary still has a
+///   head op if it truncates the entire pipeline.
 pub const view_change_headers_max = view_change_headers_suffix_max + 2;
 
 comptime {
-    assert(view_change_headers_suffix_max > 0);
-    assert(view_change_headers_suffix_max >= pipeline_prepare_queue_max);
+    assert(view_change_headers_suffix_max >= pipeline_prepare_queue_max + 1);
 
     assert(view_change_headers_max > 0);
-    assert(view_change_headers_max >= pipeline_prepare_queue_max + 2);
+    assert(view_change_headers_max >= pipeline_prepare_queue_max + 3);
     assert(view_change_headers_max <= journal_slot_count);
     assert(view_change_headers_max <= @divFloor(message_body_size_max, @sizeOf(vsr.Header)));
     assert(view_change_headers_max > view_change_headers_suffix_max);
--- a/src/vsr.zig
+++ b/src/vsr.zig
@@ -1058,10 +1058,13 @@ pub fn quorums(replica_count: u8) struct {
     assert(quorum_view_change >= @divFloor(replica_count, 2) + 1);
     assert(quorum_view_change + quorum_replication > replica_count);
 
+    const quorum_nack_prepare = replica_count - quorum_replication + 1;
+    assert(quorum_nack_prepare + quorum_replication > replica_count);
+
     return .{
         .replication = quorum_replication,
         .view_change = quorum_view_change,
-        .nack_prepare = replica_count - quorum_replication + 1,
+        .nack_prepare = quorum_nack_prepare,
     };
 }
 
@@ -1071,6 +1074,7 @@ test "quorums" {
     const expect_replication = [_]u8{ 1, 2, 2, 2, 3, 3, 3, 3 };
     const expect_view_change = [_]u8{ 1, 2, 2, 3, 3, 4, 5, 6 };
     const expect_nack_prepare = [_]u8{ 1, 1, 2, 3, 3, 4, 5, 6 };
+
     for (expect_replication[0..]) |_, i| {
         const replicas = @intCast(u8, i) + 1;
         const actual = quorums(replicas);
@@ -1099,7 +1103,7 @@ pub const Headers = struct {
 
     fn dvc_blank(op: u64) Header {
         return std.mem.zeroInit(Header, .{
-            .command = .prepare,
+            .command = .reserved,
             .op = op,
             .checksum = 0,
         });
@@ -1107,17 +1111,17 @@ pub const Headers = struct {
 
     fn dvc_fault(op: u64) Header {
         return std.mem.zeroInit(Header, .{
-            .command = .prepare,
+            .command = .reserved,
             .op = op,
             .checksum = 1,
         });
     }
 
     pub fn dvc_header_type(header: *const Header) enum { blank, fault, valid } {
-        assert(header.command == .prepare);
-
         if (std.meta.eql(header.*, Headers.dvc_blank(header.op))) return .blank;
         if (std.meta.eql(header.*, Headers.dvc_fault(header.op))) return .fault;
+
+        assert(header.command == .prepare);
         if (constants.verify) assert(header.valid_checksum());
         return .valid;
     }
@@ -1143,47 +1147,45 @@ const ViewChangeHeadersSlice = struct {
         assert(headers.slice.len > 0);
         assert(headers.slice.len <= constants.view_change_headers_max);
 
-        const op_head = headers.slice[0].op;
+        const head = &headers.slice[0];
+        // A DVC's head op is never a gap or faulty.
+        // A SV never includes gaps or faulty headers.
+        assert(Headers.dvc_header_type(head) == .valid);
 
         if (headers.command == .start_view) {
             assert(headers.slice.len >= @minimum(
                 constants.view_change_headers_suffix_max,
-                op_head + 1, // +1 to include the head itself
+                head.op + 1, // +1 to include the head itself.
             ));
         }
 
-        var child: ?*const Header = null;
-        for (headers.slice) |*header, i| {
-            assert(header.command == .prepare);
+        var child = head;
+        for (headers.slice[1..]) |*header, i| {
+            const index = i + 1;
+            assert(header.command == .prepare or header.command == .reserved);
+            assert(header.op < child.op);
 
             // DVC: Ops are consecutive (with explicit blank headers).
             // SV: The first "pipeline + 1" ops of the SV are consecutive.
             if (headers.command == .do_view_change or
-                (headers.command == .start_view and i < constants.pipeline_prepare_queue_max + 1))
+                (headers.command == .start_view and
+                index < constants.pipeline_prepare_queue_max + 1))
             {
-                assert(header.op == op_head - i);
+                assert(header.op == head.op - index);
             }
 
-            if (child) |child_header| {
-                assert(header.op < child_header.op);
-
-                switch (Headers.dvc_header_type(header)) {
-                    .blank, .fault => {
-                        assert(headers.command == .do_view_change);
-                        continue; // Don't update "child".
-                    },
-                    .valid => {
-                        assert(header.view <= child_header.view);
-                        assert(header.timestamp < child_header.timestamp);
-                        if (header.op + 1 == child_header.op) {
-                            assert(header.checksum == child_header.parent);
-                        }
-                    },
-                }
-            } else {
-                // A DVC's head op is never a gap or faulty.
-                // A SV never includes gaps or faulty headers.
-                assert(Headers.dvc_header_type(header) == .valid);
+            switch (Headers.dvc_header_type(header)) {
+                .blank, .fault => {
+                    assert(headers.command == .do_view_change);
+                    continue; // Don't update "child".
+                },
+                .valid => {
+                    assert(header.view <= child.view);
+                    assert(header.timestamp < child.timestamp);
+                    if (header.op + 1 == child.op) {
+                        assert(header.checksum == child.parent);
+                    }
+                },
             }
             child = header;
         }
@@ -1219,7 +1221,9 @@ const ViewChangeHeadersSlice = struct {
             }
             break :blk &headers.slice[oldest.?];
         };
-        assert(header_oldest.op <= header_newest.op);
+        assert(header_newest.view <= log_view);
+        assert(header_newest.view >= header_oldest.view);
+        assert(header_newest.op >= header_oldest.op);
 
         if (op < header_oldest.op) return .{ .min = 0, .max = header_oldest.view };
         if (op > header_newest.op) return .{ .min = log_view, .max = log_view };
@@ -1350,6 +1354,10 @@ const ViewChangeHeadersArray = struct {
 
     pub fn start_view_into_do_view_change(headers: *ViewChangeHeadersArray) void {
         assert(headers.command == .start_view);
+        // This function is only called by a replica that is lagging behind the primary's
+        // checkpoint, so the start_view has a full suffix of headers.
+        assert(headers.array.get(0).op >= constants.journal_slot_count);
+        assert(headers.array.len >= constants.view_change_headers_suffix_max);
         headers.command = .do_view_change;
 
         // Remove the "hook" checkpoint trigger(s), since they would create gaps in the ops.
@@ -1360,7 +1368,11 @@ const ViewChangeHeadersArray = struct {
         {
             headers.array.len -= 1;
         }
+
         assert(removed <= 2);
+        // The remaining headers may be larger than the suffix because the latest hook may chain
+        // with the suffix, so we can keep it.
+        assert(headers.array.len >= constants.view_change_headers_suffix_max);
         headers.verify();
     }
 
--- a/src/vsr.zig
+++ b/src/vsr.zig
@@ -1096,33 +1096,94 @@ pub const Headers = struct {
     /// - DVC headers (disjoint chain)
     pub const ViewChangeSlice = ViewChangeHeadersSlice;
     pub const ViewChangeArray = ViewChangeHeadersArray;
+
+    fn dvc_blank(op: u64) Header {
+        return std.mem.zeroInit(Header, .{
+            .command = .prepare,
+            .op = op,
+            .checksum = 0,
+        });
+    }
+
+    fn dvc_fault(op: u64) Header {
+        return std.mem.zeroInit(Header, .{
+            .command = .prepare,
+            .op = op,
+            .checksum = 1,
+        });
+    }
+
+    pub fn dvc_header_type(header: *const Header) enum { blank, fault, valid } {
+        assert(header.command == .prepare);
+
+        if (std.meta.eql(header.*, Headers.dvc_blank(header.op))) return .blank;
+        if (std.meta.eql(header.*, Headers.dvc_fault(header.op))) return .fault;
+        if (constants.verify) assert(header.valid_checksum());
+        return .valid;
+    }
 };
 
+pub const ViewChangeCommand = enum { do_view_change, start_view };
+
 const ViewChangeHeadersSlice = struct {
+    command: ViewChangeCommand,
     /// Headers are ordered from high-to-low op.
     slice: []const Header,
 
-    pub fn init(slice: []const Header) ViewChangeHeadersSlice {
-        ViewChangeHeadersSlice.verify(slice);
-
-        return .{ .slice = slice };
+    pub fn init(command: ViewChangeCommand, slice: []const Header) ViewChangeHeadersSlice {
+        const headers = ViewChangeHeadersSlice{
+            .command = command,
+            .slice = slice,
+        };
+        headers.verify();
+        return headers;
     }
 
-    pub fn verify(slice: []const Header) void {
-        assert(slice.len > 0);
-        assert(slice.len <= constants.view_change_headers_max);
+    pub fn verify(headers: ViewChangeHeadersSlice) void {
+        assert(headers.slice.len > 0);
+        assert(headers.slice.len <= constants.view_change_headers_max);
+
+        const op_head = headers.slice[0].op;
+
+        if (headers.command == .start_view) {
+            assert(headers.slice.len >= @minimum(
+                constants.view_change_headers_suffix_max,
+                op_head + 1, // +1 to include the head itself
+            ));
+        }
 
         var child: ?*const Header = null;
-        for (slice) |*header| {
-            assert(header.valid_checksum());
+        for (headers.slice) |*header, i| {
             assert(header.command == .prepare);
+
+            // DVC: Ops are consecutive (with explicit blank headers).
+            // SV: The first "pipeline + 1" ops of the SV are consecutive.
+            if (headers.command == .do_view_change or
+                (headers.command == .start_view and i < constants.pipeline_prepare_queue_max + 1))
+            {
+                assert(header.op == op_head - i);
+            }
 
             if (child) |child_header| {
                 assert(header.op < child_header.op);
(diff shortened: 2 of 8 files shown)


src/constants.zig
```diff
@@ -202,24 +202,34 @@ comptime {
 /// - We must provide enough headers to cover all uncommitted headers so that the new
 ///   primary (if we are in a view change) can decide whether to discard uncommitted headers
 ///   that cannot be repaired because they are gaps. See DVCQuorum for more detail.
+/// - +1 to leave room for commit_max, in case a backup converts the SV to a DVC.
```
matklad, line 205: which +1 is that?

sentientwaffle: `view_change_headers_suffix_max` comes from the configuration — it must be at least `pipeline_prepare_queue_max + 1`.

src/vsr.zig
```diff
@@ -1060,6 +1061,7 @@ pub fn quorums(replica_count: u8) struct {
     return .{
         .replication = quorum_replication,
         .view_change = quorum_view_change,
+        .nack_prepare = replica_count - quorum_replication + 1,
```
matklad: Perhaps it makes sense to introduce a variable here and write the quorum intersection assert we need here? `nack_prepare + replication > repic_count`. On the one hand, trivial, on the other hand, not really, as `+1` could be replaced with `+2`, but not with `+0`

src/vsr.zig
```diff
@@ -1068,11 +1070,20 @@ test "quorums" {
 
     const expect_replication = [_]u8{ 1, 2, 2, 2, 3, 3, 3, 3 };
     const expect_view_change = [_]u8{ 1, 2, 2, 3, 3, 4, 5, 6 };
-
+    const expect_nack_prepare = [_]u8{ 1, 1, 2, 3, 3, 4, 5, 6 };
     for (expect_replication[0..]) |_, i| {
```
matklad:
```suggestion
    const expect_nack_prepare = [_]u8{ 1, 1, 2, 3, 3, 4, 5, 6 };

    for (expect_replication[0..]) |_, i| {
```

src/vsr.zig
```diff
@@ -1085,33 +1096,94 @@ pub const Headers = struct {
     /// - DVC headers (disjoint chain)
     pub const ViewChangeSlice = ViewChangeHeadersSlice;
     pub const ViewChangeArray = ViewChangeHeadersArray;
+
+    fn dvc_blank(op: u64) Header {
+        return std.mem.zeroInit(Header, .{
+            .command = .prepare,
+            .op = op,
+            .checksum = 0,
```
matklad, line 1108: I guess we are just not worrying that something checksums to zero, right? Makes sense, but still gives me a pause. No actionable suggestions :)

src/vsr.zig
```diff
@@ -1085,33 +1096,94 @@ pub const Headers = struct {
...
+                        assert(header.timestamp < child_header.timestamp);
+                        if (header.op + 1 == child_header.op) {
+                            assert(header.checksum == child_header.parent);
+                        }
+                    },
+                }
+            } else {
+                // A DVC's head op is never a gap or faulty.
+                // A SV never includes gaps or faulty headers.
+                assert(Headers.dvc_header_type(header) == .valid);
```
matklad: Maybe just special-case the first one? 

(uncompiled)

```zig
    pub fn verify(headers: ViewChangeHeadersSlice) void {
        assert(headers.slice.len > 0);
        assert(headers.slice.len <= constants.view_change_headers_max);

        const head = &headers.slice[0];
        // A DVC's head op is never a gap or faulty.
        // A SV never includes gaps or faulty headers.
        assert(Headers.zdvc_header_type(head) == .valid);

        if (headers.command == .start_view) {
            assert(headers.slice.len >= @min(
                constants.view_change_headers_suffix_max,
                head.op + 1, // +1 to include the head itself
            ));
        }

        var child: ?*const Header = head;
        for (headers.slice[1..]) |*header, i| {

            // DVC: Ops are consecutive (with explicit blank headers).
            // SV: The first "pipeline + 1" ops of the SV are consecutive.
            if (headers.command == .do_view_change or
                (headers.command == .start_view and i < constants.pipeline_prepare_queue_max))
            {
                assert(header.op == head.op - i - 1);
            }

            assert(header.op < child.op);

            switch (Headers.dvc_header_type(header)) {
                .blank, .fault => {
                    assert(headers.command == .do_view_change);
                    // Don't update "child".
                },
                .valid => {
                    assert(header.view <= child.view);
                    assert(header.timestamp < child.timestamp);
                    if (header.op + 1 == child.header.op) {
                        assert(header.checksum == child.parent);
                    }
                    child = header;
                },
            }
        }
    }
```

src/vsr.zig
```diff
@@ -1136,32 +1208,97 @@ const ViewChangeHeadersSlice = struct {
     ///   a view prior to the log_view, they would already be part of the headers.
     pub fn view_for_op(headers: ViewChangeHeadersSlice, op: u64, log_view: u32) ViewRange {
```
matklad, line 1211: At the call-site, we do `view_range.max = std.math.min(view_range.max, log_view);`. But it looks like we can assert that?

src/constants.zig
matklad: Ahh, see now

src/vsr.zig
```diff
@@ -1174,27 +1311,85 @@ test "Headers.ViewChangeSlice.view_for_op" {
...
-    pub fn replace(headers: *ViewChangeHeadersArray, slice: []const Header) void {
-        Headers.ViewChangeSlice.verify(slice);
+    pub fn start_view_into_do_view_change(headers: *ViewChangeHeadersArray) void {
+        assert(headers.command == .start_view);
+        headers.command = .do_view_change;
+
+        // Remove the "hook" checkpoint trigger(s), since they would create gaps in the ops.
+        // (There are at most 2 hooks).
+        var removed: usize = 0;
+        while (headers.array.get(headers.array.len - 2).op !=
```
matklad, line 1339: Hm, do we always have at least to?

```
        if (headers.command == .start_view) {
            assert(headers.slice.len >= @minimum(
                constants.view_change_headers_suffix_max,
                op_head + 1, // +1 to include the head itself
            ));
        }
```

makes me thing that one is possible. If it isn't, that should also assert that `op_head >= 1`

src/vsr.zig
sentientwaffle: Even if a header does legitimately have `checksum=0|1`, some field would be different, so the `std.meta.eql`s in `dvc_header_type` wouldn't match.

src/vsr/README.md
```diff
@@ -172,7 +176,8 @@ During repair, missing/damaged prepares are requested & repaired chronologically
 In response to a `request_prepare`:
 
 - Reply the `command=prepare` with the requested prepare, if available and valid.
-- Reply `command=nack_prepare` if the request origin is the primary of the ongoing view-change and we never received the prepare. (This enables the primary to truncate uncommitted messages and remain available).
+- Reply `command=nack_prepare` if the request origin is the primary of the ongoing view-change and we never received the prepare.
```
matklad, line 179: :thinking: could we in theory remove `command=nack_prepare` altogether, and just reply with a DVC?

src/vsr.zig
matklad: But maybe set `.command` to something other than `.prepare`, such that, if we accidently use a fake header, we'd blow up? Eg, I think if we pass these blaks by accident to `replace_header` stuff will "just work" rather than "just bork"?

src/vsr.zig
```diff
@@ -1136,32 +1208,97 @@ const ViewChangeHeadersSlice = struct {
...
+                if (header_prev.op < op and op < header_next.op) {
+                    return .{ .min = header_prev.view, .max = header_next.view };
+                }
+                header_next = header_prev;
             }
         }
         unreachable;
     }
+
+    pub const Iterator = struct {
```
matklad: I wonder if Iterator pulls its weight? The alternative would be something like

```zig
if (header.is_fake()) continue;                 
```

condition at the two call-sites we have.

Though, yeah, we assert here more thoroughly.

src/vsr/replica.zig
```diff
@@ -6700,152 +6717,186 @@ const DVCQuorum = struct {
...
+    /// Otherwise:
+    /// - Return the reason the view cannot begin.
+    fn headers_canonical(dvc_quorum: QuorumMessages, options: struct {
+        quorum_nack_prepare: u8,
+        quorum_view_change: u8,
+        replica_count: u8,
+    }) union(enum) {
+        // The quorum has fewer than "quorum_view_change" DVCs.
+        // We are waiting for DVCs from the remaining replicas.
+        awaiting_quorum: void,
```
matklad: maybe just 

```suggestion
        awaiting_quorum,
```

? 

That should be equivalent, right? When I have syntactic choice, I usually lean towards the minimal specialized option as a tie-breaker.

src/vsr.zig
sentientwaffle: `start_view.headers.len == 1` is possible in general (and in `verify()`), but not in `start_view_into_do_view_change`.

Adding this explanation:

```zig
    pub fn start_view_into_do_view_change(headers: *ViewChangeHeadersArray) void {
        assert(headers.command == .start_view);
        // This function is only called by a replica that is lagging behind the primary's
        // checkpoint, so the start_view has a full suffix of headers.
        assert(headers.array.get(0).op >= constants.journal_slot_count);
        assert(headers.array.len >= constants.view_change_headers_suffix_max);
```

and this postcondition after the hooks are removed:

```zig
        assert(headers.array.len == constants.view_change_headers_suffix_max);
```

src/vsr/replica.zig
```diff
@@ -3139,46 +3070,17 @@ pub fn ReplicaType(
...
-                    }) |op_hook| {
-                        if (op > op_hook and op_hook >= op_min) {
-                            op = op_hook;
-                            self.view_headers.array.appendAssumeCapacity(
-                                self.journal.header_with_op(op).?.*,
-                            );
-                        }
-                    }
+                    self.primary_update_view_headers();
+                    assert(self.view_headers.command == .start_view);
```
matklad, line 3082:
```suggestion
                    assert(self.view_headers.command == .start_view);
                    assert(self.view_headers.array.get(0).op == self.op);
```

src/vsr.zig
sentientwaffle: Good catch!

src/vsr/replica.zig
```diff
@@ -6700,152 +6717,186 @@ const DVCQuorum = struct {
...
+        // The quorum collected so far is insufficient to determine which headers can be nacked
+        // (due to an excess of faults).
+        // We must wait for DVCs from one or more remaining replicas.
+        awaiting_repair: void,
+        // All replicas have contributed a DVC, but there are too many faults to start a new view.
+        // The cluster is deadlocked, unable to ever complete a view change.
+        complete_invalid: void,
+        // The quorum is complete, and sufficient to start the new view.
+        complete_valid: HeaderIterator,
+    } {
```
matklad: Originally it felt a bit odd that we have "wait for quorum of _messages_" at all, as it seemed like we should be coming to an agreement on a particular op. But this actually makes sense -- we have an invariant that the head op can't be corrupted (replica in recovering_head sends hardly a message). So, getting a quorum of messages means getting the quorum for the head op.

src/vsr.zig
matklad: I guess another concern here is that we have DVCQuorum.HeaderIterator and ViewChangeHeadersSlice.Iterator, and I feel like we should either name them exactly the same, or make naming more distinct.

src/vsr/replica.zig
```diff
@@ -6700,152 +6717,186 @@ const DVCQuorum = struct {
...
+                        .valid => break true,
+                    }
                 }
+            } else false;
+
+            if (found) {
+                // This op is eligible to be the view's head.
             } else {
-                break :blk op_head_max;
+                // Never nack op_head_min (aka commit_max).
```
matklad: Shouldn't this be an assert? The op is committed, so it can't be nacked, right?

src/vsr/replica.zig
```diff
@@ -6700,152 +6717,186 @@ const DVCQuorum = struct {
...
+                    }
+                }
             }
-        };
+        } else op_head_max;
         assert(op_head >= op_head_min);
         assert(op_head <= op_head_max);
 
-        return HeaderIterator.init(dvcs, op_head);
+        return .{ .complete_valid = .{
```
```suggestion
        return .{ .complete_valid = HeaderIterator{
```

after groking that intricate while/for loop, i completely forgot what  we were supposed to return here :D

src/vsr/replica.zig
```diff
@@ -5779,6 +5752,48 @@ pub fn ReplicaType(
             }
         }
 
+        fn primary_print_do_view_change_quorum(
```
```suggestion
        fn primary_log_do_view_change_quorum(
```

src/vsr.zig
sentientwaffle: Removed. The asserts were redundant since `verify()` should be called any time the slice is initialized.

src/vsr/replica.zig
sentientwaffle: Good point! There may be a few nacks, but never a quorum.

src/vsr/replica.zig
sentientwaffle: (TIL this is valid syntax!)

src/vsr.zig
matklad: I think you forgot to remove/push removal of Iterator itself?

sentientwaffle: Oops; removed.

src/vsr/replica.zig
```diff
@@ -2285,34 +2277,11 @@ pub fn ReplicaType(
...
-            }
-
-            // This is not the first time we have had quorum, the state transition has already happened:
-            if (count > threshold) {
-                log.debug("{}: on_{s}: ignoring (quorum received already)", .{
-                    self.replica,
-                    command,
-                });
-                return null;
+            } else {
```
matklad, line 2271: `// Do not allow duplicate messages to trigger multiple passes through a state transition:` commend no longer makes sense, as we now return `void`

src/vsr/replica.zig
```diff
@@ -2205,24 +2207,14 @@ pub fn ReplicaType(
...
+        fn add_message_to_quorum(self: *Self, messages: *QuorumMessages, message: *Message) void {
             assert(messages.len == constants.replicas_max);
             assert(message.header.cluster == self.cluster);
             assert(message.header.replica < self.replica_count);
             assert(message.header.view == self.view);
             switch (message.header.command) {
                 .do_view_change => {
                     assert(!self.solo());
-                    if (self.replica_count == 2) assert(threshold == 2);
-
```
matklad, line 2225: while we are at it, might replace switch with assert?

matklad: Actually, it seems like we in general have some dead code here, `if (message.header.command == .do_view_change) {` bellow is always true now, no? Probably makes sense to refactor separately then.

src/vsr/replica.zig
```diff
@@ -6550,13 +6587,6 @@ const DVCQuorum = struct {
                 log_views_all.appendAssumeCapacity(log_view);
             }
         }
-
-        // Verify that DVCs with the same log_view do not conflict.
-        for (log_views_all.constSlice()) |log_view| {
-            const view_dvcs = dvcs_with_log_view(dvc_quorum, log_view);
-            var view_headers = HeaderIterator.init(view_dvcs, null);
-            while (view_headers.next()) |_| {}
-        }
```
matklad, line 6559: The loop above is dead code as well?

src/vsr/replica.zig
sentientwaffle: :+1: I'll update it. This function used to handle all of the quorums, but all of the others were changed to simple bitsets (for less RAM use).

src/vsr/replica.zig
```diff
@@ -6659,19 +6691,19 @@ const DVCQuorum = struct {
 
         var commit_max_: u64 = 0;
         for (dvcs.constSlice()) |dvc| {
-            const dvc_headers = message_body_as_headers_chain_disjoint(dvc);
-            // The DVC is connected to the committed ops.
-            const dvc_commit_max_connected = dvc_headers[dvc_headers.len - 1].op -| 1;
+            const dvc_headers = message_body_as_view_headers(dvc);
+            // DVC generation stops when a header with op ≤ commit_max is appended.
+            const dvc_commit_max_tail = dvc_headers.slice[dvc_headers.slice.len - 1].op;
```
matklad, line 6708: Can `dvc_headers.slice[dvc_headers.slice.len - 1]` be invalid?

src/vsr/replica.zig
```diff
@@ -5826,15 +5857,9 @@ pub fn ReplicaType(
...
-            self.view_headers.array.len = 0;
-            var op = self.op + 1;
-            while (op > 0) {
-                op -= 1;
-                assert(op >= self.op_repair_min());
-
-                self.view_headers.array.append(self.journal.header_with_op(op).?.*) catch break;
-            }
-            vsr.Headers.ViewChangeSlice.verify(self.view_headers.array.constSlice());
+            self.view_headers.command = .start_view;
```
matklad, line 5849: Delete commit above, we now will include the hooks I think?

src/vsr/replica.zig
sentientwaffle: Ah, I removed that intending to reimplement that check (since it can't use `HeaderIterator` anymore) but forgot to come back to it. I'll update that now.

src/vsr/replica.zig
```diff
@@ -6492,36 +6532,34 @@ pub fn ReplicaType(
 ///
```
matklad, line 6523: Unrelated to the PR, but I continue to be mystified by the canonical terminology as it is explained here. 

My take on the explanation: 

Canonical messages: messages from DVC quorum with the highest log view. By the quorum intersection property, each op that could have been committed is included in at least one canonical message

Canonical headers: subset of headers from canonical messages, which could have been committed. This might be a strict subset: even if a header is present in some canonical message, we might conclude that it is not committed if it is missing from enough other canonical messages.

src/vsr/replica.zig
```diff
@@ -6492,36 +6532,34 @@ pub fn ReplicaType(
...
-///     - a DVC of 7,8 is invalid if replica.commit_min=5.
-///     - a DVC of 7,8 is valid if replica.commit_min=6.
-///     - a DVC of 5,7,8 is valid. (5,_,7,8)
-///     - a DVC of 5,8 is valid.   (5,_,_,8)
-///     - a DVC of 0,1 is valid.
+///   - When `replica.commit_max ≤ replica.op`,
+///     the DVC must include a valid/blank/fault header for every op in that range.
+///   - When `replica.commit_max > replica.op`, only a single header (`replica.op`) is included.
+///   - The DVC will need a valid `commit_max` to complete, since the entire pipeline may be
+///     truncated, and the new primary still needs a head op.
```
matklad: This talks about commit_min, right?

matklad: Ah, no, its me being confused

src/vsr/replica.zig
```diff
@@ -6700,152 +6733,187 @@ const DVCQuorum = struct {
...
-        return HeaderIterator.init(dvcs, op_head);
+        return .{ .complete_valid = HeaderIterator{
+            .dvcs = dvcs,
+            .op_max = op_head,
+            .op_min = op_head_min,
+        } };
     }
 
     /// Iterate the headers of a set of (same-log_view) DVCs, from high-to-low op.
     const HeaderIterator = struct {
```
matklad, line 6848: lovely simplification here!

src/vsr/replica.zig
sentientwaffle: Invalid in what sense?

`dvc_headers.slice.len` is always at least 1.
`dvc_headers.slice[dvc_headers.slice.len - 1]` may correspond to a blank or a fault, but in those cases we would still know that their op (which is set even for blanks/faults) corresponds to `commit_max` according to the sender.

src/vsr/replica.zig
sentientwaffle: I consider "canonical headers" to be the union of all headers from all canonical messages. (Regardless of whether they are definitely-committed, maybe-committed, or definitely-uncommitted).

I'll update the docs — they mix up "canonical message" and "canonical header" which probably made it extra hard to understand. :sweat:

src/vsr/replica.zig
matklad: blank or fault, but we indeed set `op` correctly for them

src/vsr/replica.zig
sentientwaffle: At the end of a new primary's repair, `commit_min == commit_max`.

But immediately after the DVC quorum finishes, it might decide to truncate the entire pipeline.
If we do truncate our entire pipeline, then the new head op == `commit_max`.
In order to repair (or do anything), our head op must always correspond to a valid header in the journal.
To guarantee that the new primary _has_ that valid header, DVCs must always include the valid `commit_max` header (if they are able to provide it — by the intersection property we will at least one).

I rewrote this to (hopefully) clarify:
```
///   - The DVC will need a valid header corresponding to its `commit_max` to complete, since the
///     entire pipeline may be truncated, and the new primary still needs a header for its head op.
```

src/vsr/replica.zig
matklad: >I consider "canonical headers" to be the union of all headers from all canonical messages.

But that's not how headers_canonical actually works, right? It strips uncommitted

matklad (commented):
Finished full review, LGTM modulo that dead loop for log_view

src/vsr/replica.zig
sentientwaffle: Good point. I guess I need to update the docs _and_ my mental model. :stuck_out_tongue:

sentientwaffle: I ended up leaving my mental model intact, but renaming `headers_canonical()` to `quorum_headers()`.
This is more consistent overall — there is a more intuitive correspondence between the canonical headers/messages/view.

src/vsr/replica.zig
```diff
@@ -6578,6 +6578,34 @@ const DVCQuorum = struct {
                 log_views_all.appendAssumeCapacity(log_view);
             }
         }
+
+        // Verify that DVCs with the same log_view do not conflict.
+        for (log_views_all.constSlice()) |log_view| {
+            const view_dvcs = dvcs_with_log_view(dvc_quorum, log_view);
+            for (view_dvcs.constSlice()) |dvc_a, i| {
+                for (view_dvcs.constSlice()[i + 1 ..]) |dvc_b| {
```
matklad: Nit, but I like to wirte such loops as

```
for i in 0..n:
  for j in 0..i:
```

as that minimizes +1

```zig
            for (view_dvcs.constSlice()) |dvc_a, i| {
                for (view_dvcs.constSlice()[0..i]) |dvc_b| {
```

src/vsr/replica.zig
```diff
@@ -6578,6 +6578,34 @@ const DVCQuorum = struct {
...
+                    const headers_a = message_body_as_view_headers(dvc_a);
+                    const headers_b = message_body_as_view_headers(dvc_b);
+                    // Find the intersection of the ops covered by each DVC.
+                    const op_max = std.math.min(dvc_a.header.op, dvc_b.header.op);
+                    const op_min = std.math.max(
+                        headers_a.slice[headers_a.slice.len - 1].op,
+                        headers_b.slice[headers_b.slice.len - 1].op,
+                    );
+
```
matklad: `assert(op_min <= op_max)`

src/vsr/replica.zig
matklad: Actually, maybe we can fold both loops in one?

```zig
for (view_dvcs.constSlice()) |dvc_a, i| {
    for (view_dvcs.constSlice()[0..i]) |dvc_b| {
        if (dvc_a.log_view == dvc_b.log_view)
```

?

src/vsr/replica.zig
sentientwaffle: I think it is `maybe(op_min > op_max);` — it is possible that a replica is lagging so far behind that its headers do not overlap at all.

matklad: Ah, that's correct,

matklad (approved):

PR #608, opened 2023-03-29
Client/Header: Use actual enum value, not variant index

sentientwaffle:
Currently our enum variants' values match their respective indexes within the enum. But that will soon not (necessarily) be true.

## Pre-merge checklist

Performance:

### DIFF
--- a/src/clients/c/tb_client_header.zig
+++ b/src/clients/c/tb_client_header.zig
@@ -71,6 +71,7 @@ fn to_uppercase(comptime input: []const u8) []const u8 {
 
 fn emit_enum(
     buffer: *std.ArrayList(u8),
+    comptime Type: type,
     comptime type_info: anytype,
     comptime c_name: []const u8,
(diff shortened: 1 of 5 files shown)


batiati:
I think all clients' binding generators suffer from this same problem:

- [dotnet_bindings.zig](https://github.com/tigerbeetledb/tigerbeetle/blob/main/src/clients/dotnet/dotnet_bindings.zig#L188-L199)
- [go_bindings.zig](https://github.com/tigerbeetledb/tigerbeetle/blob/main/src/clients/go/go_bindings.zig#L104-L111)
- [java_bindings](https://github.com/tigerbeetledb/tigerbeetle/blob/main/src/clients/java/java_bindings.zig#L186-L194)
- [node_bindings](https://github.com/tigerbeetledb/tigerbeetle/blob/main/src/clients/node/node_bindings.zig#L98-L107)

batiati (approved):
Nice!

PR #609, opened 2023-03-29
VSR: Reserve 128 operations for VSR

sentientwaffle:
Reserve the first 128 operations for VSR, leaving the remaining 128 for the state machine.

## Pre-merge checklist

Performance:


src/constants.zig
```diff
@@ -32,6 +32,14 @@ pub const standbys_max = 6;
 /// The maximum number of nodes (either standbys or active replicas) allowed in a cluster.
 pub const nodes_max = replicas_max + standbys_max;
 
+/// All operations <vsr_operations_reserved are reserved for the control protocol.
+/// All operations ≥vsr_operations_reserved are available for the state machine.
+pub const vsr_operations_reserved = 128;
+
+comptime {
+    assert(vsr_operations_reserved <= std.math.maxInt(u8));
+}
```
matklad: would

```suggestion
pub const vsr_operations_reserved: u8 = 128;
```

work?

src/state_machine.zig
```diff
@@ -222,15 +223,15 @@ pub fn StateMachineType(comptime Storage: type, comptime constants_: struct {
...
 
             /// Operations exported by TigerBeetle:
-            create_accounts,
-            create_transfers,
-            lookup_accounts,
-            lookup_transfers,
+            create_accounts = constants_.vsr_operations_reserved + 0,
+            create_transfers = constants_.vsr_operations_reserved + 1,
+            lookup_accounts = constants_.vsr_operations_reserved + 2,
+            lookup_transfers = constants_.vsr_operations_reserved + 3,
```
matklad:
```suggestion
            create_accounts = constants_.vsr_operations_reserved + 0,
            create_transfers,
            lookup_accounts,
            lookup_transfers,
```

I'd maybe just rely on auto-numbering behavior, as that guarantees absense of gaps

src/state_machine.zig
```diff
@@ -22,10 +22,11 @@ const CreateTransfersResult = tb.CreateTransfersResult;
...
 const CreateTransferResult = tb.CreateTransferResult;
 
-pub fn StateMachineType(comptime Storage: type, comptime constants_: struct {
-    message_body_size_max: usize,
-    lsm_batch_multiple: usize,
-}) type {
+pub fn StateMachineType(comptime Storage: type, comptime constants_: type) type {
+    assert(constants_.message_body_size_max > 0);
+    assert(constants_.lsm_batch_multiple > 0);
+    assert(constants_.vsr_operations_reserved > 0);
```
matklad: Note sure how I feel about this, we seem to lose some type-safety here. Perhaps we can

```zig
pub fn StateMachineType(comptime Storage: type, comptime constants_: struct {
    message_body_size_max: usize = @import("constants").message_body_size_max,
}) type {
```

?

src/vsr.zig
```diff
@@ -132,7 +132,8 @@ pub const Operation = enum(u8) {
     /// The value 2 is reserved to register a client session with the cluster.
     register = 2,
 
-    /// Operations exported by the state machine (all other values are free):
+    /// Operations <vsr_operations_reserved are reserved for the control plane.
+    /// Operations ≥vsr_operations_reserved are available for the state machine.
     _,
```
matklad, line 137: I _think_ some validation is missing here. If the client sends us operation=92, what will happen? Seems like it'd just blow-up in intToEnum casts. 

I _think_ we should check in `invalid_request` and such that, if operation is < 128, than it's one of the reserved ones. 

:thinking: I think we are not validating this on main either?

matklad (commented):
Let's figure the validation story while we are at it? (or maybe I am missing the place where we do the validation of client-originating operations?)

src/state_machine.zig
sentientwaffle: I kind of like avoiding auto-numbering for this since it helps ensure that the values are stable — reordering an enum is normally ok, but in this case it makes it clear that reordering is a breaking change.

sentientwaffle:
> Let's figure the validation story while we are at it? (or maybe I am missing the place where we do the validation of client-originating operations?)

Yeah, not much validation yet (here or in main). I can check that it isn't an invalid vsr-reserved operation in `invalid_request()`, but the `Header` doesn't know the state machine's operation type, so it can't check for an invalid state machine operation.

Maybe `StateMachine.cast` should be fallable — return an error if it is >128 but not a legal state machine operation?

src/vsr.zig
matklad, line 1: Let's also check in `Client.request` that the operation is not reserved, instead of separtely checking for 

```
            assert(operation != .reserved);
            assert(operation != .root);
            assert(operation != .register);
```

matklad (approved):
LGTM!

Not sure what's the status of fed's CI investigation though

sentientwaffle:
Thanks! He narrowed it down to a bug in the macos test runners' caching. He added a workaround, so this is good to merge.

PR #612, opened 2023-03-30
Test: Cluster consensus tests

sentientwaffle:
Add a new suite of tests to replicate certain recovery and view-change scenarios that are hard or in some cases impossible for the vopr to replicate.

For example
- Scenarios where the cluster is unable to ever safely recover, and make sure that it behaves as expected.
- Corruption too serious for the replica to recover from.
- Major disk corruption near checkpoints.
- Asymmetric partition liveness.
- View-change forfeit
- View-change truncate post-checkpoint uncommitted ops

Right now `TestClients` is always used as a full pool of clients. There are some cases I will want to test in the future (e.g. client eviction) where more fine-grained control is needed, hence the current design.

## Pre-merge checklist

Performance:

### DIFF
--- a/src/stdx.zig
+++ b/src/stdx.zig
@@ -1,6 +1,7 @@
 //! Extensions to the standard library -- things which could have been in std, but aren't.
 
 const std = @import("std");
+const builtin = @import("builtin");
 const assert = std.debug.assert;
 
 pub inline fn div_ceil(numerator: anytype, denominator: anytype) @TypeOf(numerator, denominator) {
@@ -184,3 +185,20 @@ const TimeIt = struct {
         );
     }
 };
+
+pub const log = if (builtin.is_test)
+    // Downgrade `err` to `warn` for tests.
+    // Zig fails any test that does `log.err`, but we want to test those code paths here.
+    struct {
+        pub fn scoped(comptime scope: @Type(.EnumLiteral)) type {
+            const base = std.log.scoped(scope);
+            return struct {
+                pub const err = warn;
+                pub const warn = base.warn;
+                pub const info = base.info;
+                pub const debug = base.debug;
+            };
+        }
+    }
+else
+    std.log;
--- a/src/testing/state_machine.zig
+++ b/src/testing/state_machine.zig
@@ -109,7 +109,7 @@ pub fn StateMachineType(comptime Storage: type, comptime constants_: struct {
                 .reserved, .root => unreachable,
                 .register => return 0,
                 .echo => {
-                    stdx.copy_disjoint(.exact, u8, output[0..input.len], input);
+                    stdx.copy_disjoint(.inexact, u8, output, input);
                     return input.len;
                 },
             }
--- a/src/testing/state_machine.zig
+++ b/src/testing/state_machine.zig
@@ -109,7 +109,7 @@ pub fn StateMachineType(comptime Storage: type, comptime constants_: struct {
                 .reserved, .root => unreachable,
                 .register => return 0,
                 .echo => {
-                    stdx.copy_disjoint(.exact, u8, output, input);
+                    stdx.copy_disjoint(.exact, u8, output[0..input.len], input);
                     return input.len;
                 },
             }
--- a/src/vsr/replica_test.zig
+++ b/src/vsr/replica_test.zig
@@ -41,7 +41,7 @@ test "Cluster: recovery: WAL prepare corruption (R=3, corrupt right of head)" {
     defer t.deinit();
 
     var c = t.clients(0, t.cluster.clients.len);
-    try expectEqual(c.request(20), 20);
+    try c.request(20, 20);
     t.replica(.R_).stop();
     t.replica(.R0).corrupt(.{ .wal_prepare = 22 });
 
@@ -49,10 +49,10 @@ test "Cluster: recovery: WAL prepare corruption (R=3, corrupt right of head)" {
     try expectEqual(t.replica(.R0).open(), .ok);
     try expectEqual(t.replica(.R0).status(), .recovering_head);
     try expectEqual(t.replica(.R1).open(), .ok);
-    try expectEqual(c.request(24), 20);
+    try c.request(24, 20);
     // With the aid of the last replica, the cluster can recover.
     try expectEqual(t.replica(.R2).open(), .ok);
-    try expectEqual(c.request(24), 24);
+    try c.request(24, 24);
     try expectEqual(t.replica(.R_).commit(), 24);
 }
 
@@ -63,7 +63,7 @@ test "Cluster: recovery: WAL prepare corruption (R=3, corrupt left of head, 3/3 
     defer t.deinit();
 
     var c = t.clients(0, t.cluster.clients.len);
-    try expectEqual(c.request(20), 20);
+    try c.request(20, 20);
     t.replica(.R_).stop();
     t.replica(.R_).corrupt(.{ .wal_prepare = 10 });
     try expectEqual(t.replica(.R_).open(), .ok);
@@ -81,12 +81,12 @@ test "Cluster: recovery: WAL prepare corruption (R=3, corrupt root)" {
     defer t.deinit();
 
     var c = t.clients(0, t.cluster.clients.len);
-    try expectEqual(c.request(20), 20);
+    try c.request(20, 20);
     t.replica(.R0).stop();
     t.replica(.R0).corrupt(.{ .wal_prepare = 0 });
     try expectEqual(t.replica(.R0).open(), .ok);
 
-    try expectEqual(c.request(21), 21);
+    try c.request(21, 21);
     try expectEqual(t.replica(.R_).commit(), 21);
 }
 
@@ -96,21 +96,21 @@ test "Cluster: recovery: WAL prepare corruption (R=3, corrupt checkpoint…head)
 
     var c = t.clients(0, t.cluster.clients.len);
     // Trigger the first checkpoint.
-    try expectEqual(c.request(63), 63);
+    try c.request(checkpoint_trigger_1, checkpoint_trigger_1);
     t.replica(.R0).stop();
 
     // Corrupt op_checkpoint (59) and all ops that follow.
-    var slot: usize = constants.journal_slot_count - constants.lsm_batch_multiple - 1;
-    while (slot < constants.journal_slot_count) : (slot += 1) {
+    var slot: usize = slot_count - constants.lsm_batch_multiple - 1;
+    while (slot < slot_count) : (slot += 1) {
         t.replica(.R0).corrupt(.{ .wal_prepare = slot });
     }
     try expectEqual(t.replica(.R0).open(), .ok);
     try expectEqual(t.replica(.R0).status(), .recovering_head);
 
-    try expectEqual(c.request(64), 64);
+    try c.request(slot_count, slot_count);
     try expectEqual(t.replica(.R0).status(), .normal);
     t.replica(.R1).stop();
-    try expectEqual(c.request(65), 65);
+    try c.request(slot_count + 1, slot_count + 1);
 }
 
 test "Cluster: recovery: WAL prepare corruption (R=1, between checkpoint and head)" {
@@ -119,7 +119,7 @@ test "Cluster: recovery: WAL prepare corruption (R=1, between checkpoint and hea
     defer t.deinit();
 
     var c = t.clients(0, t.cluster.clients.len);
-    try expectEqual(c.request(20), 20);
+    try c.request(20, 20);
     t.replica(.R0).stop();
     t.replica(.R0).corrupt(.{ .wal_prepare = 15 });
     try expectEqual(t.replica(.R0).open(), .WALCorrupt);
@@ -131,11 +131,11 @@ test "Cluster: recovery: WAL header corruption (R=1)" {
     defer t.deinit();
 
     var c = t.clients(0, t.cluster.clients.len);
-    try expectEqual(c.request(20), 20);
+    try c.request(20, 20);
     t.replica(.R0).stop();
     t.replica(.R0).corrupt(.{ .wal_header = 15 });
     try expectEqual(t.replica(.R0).open(), .ok);
-    try expectEqual(c.request(30), 30);
+    try c.request(30, 30);
 }
 
 test "Cluster: recovery: WAL torn prepare, standby with intact prepare (R=1 S=1)" {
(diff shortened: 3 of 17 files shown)


src/testing/state_machine.zig
```diff
@@ -109,7 +109,7 @@ pub fn StateMachineType(comptime Storage: type, comptime constants_: struct {
                 .reserved, .root => unreachable,
                 .register => return 0,
                 .echo => {
-                    stdx.copy_disjoint(.exact, u8, output, input);
+                    stdx.copy_disjoint(.exact, u8, output[0..input.len], input);
```
matklad:
```suggestion
                    stdx.copy_disjoint(.inexact, u8, output, input);
```

I think we have the API specifically for such cases, right?

src/stdx.zig
```diff
@@ -184,3 +185,20 @@ const TimeIt = struct {
         );
     }
 };
+
+pub const log = if (builtin.is_test)
```
matklad, line 189:
```
const log = stdx.log.scoped(.replica);
const log =  std.log.scoped(.replica);
```

Look very similar. Should we make the difference more obvious by renaming? I lean towards no

src/vsr/replica_test.zig
```diff
@@ -0,0 +1,920 @@
...
+    try expectEqual(t.replica(.R_).commit(), 21);
+}
+
+test "Cluster: recovery: WAL prepare corruption (R=3, corrupt checkpoint…head)" {
+    const t = try TestContext.init(.{ .replica_count = 3 });
+    defer t.deinit();
+
+    var c = t.clients(0, t.cluster.clients.len);
+    // Trigger the first checkpoint.
+    try expectEqual(c.request(63), 63);
```
matklad: Maybe add something like 

```zig
try expectEqual(t.replica(R_).checkpoint_op(), 59);
```

63 is a magic number, would be nice to verify it is indeed a checkpoint.

src/vsr/replica_test.zig
```diff
@@ -0,0 +1,920 @@
...
+    const t = try TestContext.init(.{
+        .replica_count = 1,
+        .standby_count = 1,
+    });
+    defer t.deinit();
+
+    var c = t.clients(0, t.cluster.clients.len);
+    try expectEqual(c.request(20), 20);
+    t.replica(.R0).corrupt(.{ .wal_header = 20 });
+    t.replica(.R0).stop();
```
matklad: Does the order of stop/corrupt matter here? If not, let's stop-then-corrupt, as that looks more logical, and also guaratees that we dont' depend on our special crash logic.

src/vsr/replica_test.zig
```diff
@@ -0,0 +1,920 @@
...
+test "Cluster: network: partition 2-1 (isolate backup, asymmetric, receive-only)" {
+    const t = try TestContext.init(.{ .replica_count = 3 });
+    defer t.deinit();
+
+    var c = t.clients(0, t.cluster.clients.len);
+    try expectEqual(c.request(20), 20);
+    t.replica(.B2).drop_all(.__, .outgoing);
+    try expectEqual(c.request(40), 40);
+    try expectEqual(t.replica(.A0).commit(), 40);
+}
```
matklad:
```suggestion
    try expectEqual(t.replica(.A0).commit(), 40);
    try expectEqual(t.replica(.B1).commit(), 40);
    try expectEqual(t.replica(.B2).commit(), 20);
}
```

I am actually a bit surprised at that 20 -- if backup is receive-only, it should still commit fine, no?

src/vsr/replica_test.zig
```diff
@@ -0,0 +1,920 @@
...
+    // The primary cannot communicate with either backup, but the backups can communicate with one
+    // another. The backups will perform a view-change since they don't receive heartbeats.
+    const t = try TestContext.init(.{ .replica_count = 3 });
+    defer t.deinit();
+
+    var c = t.clients(0, t.cluster.clients.len);
+    try expectEqual(c.request(20), 20);
+    t.replica(.A0).drop_all(.B1, .bidirectional);
+    t.replica(.A0).drop_all(.B2, .bidirectional);
+    try expectEqual(c.request(30), 30);
```
matklad: Maybe 

```zig
    const p = t.replica(.A0);
    p.drop_all(.B1, .bidirectional);
    p.drop_all(.B2, .bidirectional);
    try expectEqual(c.request(30), 30);
    try expectEqual(p.commit(), 20);
```

to double check that the primary is indeed stuck?

src/vsr/replica_test.zig
```diff
@@ -0,0 +1,920 @@
...
+    // The primary can send to the backups, but not receive.
+    // After a short interval of not receiving messages (specifically prepare_ok's) it will abdicate
+    // by pausing heartbeats, allowing the next replica to take over as primary.
+    const t = try TestContext.init(.{ .replica_count = 3 });
+    defer t.deinit();
+
+    var c = t.clients(0, t.cluster.clients.len);
+    try expectEqual(c.request(20), 20);
+    t.replica(.A0).drop_all(.B1, .incoming);
+    t.replica(.A0).drop_all(.B2, .incoming);
```
matklad, line 232: would be a good place for something like 

```zig
    t.replica(.A0).will_log("primary abdicating");
    try expectEqual(c.request(30), 30);
    try t.check_logs();
```

src/vsr/replica_test.zig
```diff
@@ -0,0 +1,920 @@
...
+// test "Cluster: network: partititon client-primary (symmetric)" {
+//     // Clients cannot communicate with the primary, but they still request/reply via a backup.
+//     const t = try TestContext.init(.{ .replica_count = 3 });
+//     defer t.deinit();
+//
+//     var c = t.clients(0, t.cluster.clients.len);
+//     try expectEqual(c.request(20), 20);
+//     t.replica(.A0).drop_all(.C_, .bidirectional);
+//     try expectEqual(c.request(40), 40);
+// }
```
matklad: Let's maybe do something like 

```zig
test "Cluster: network: partititon client-primary (symmetric)" {
    // Clients cannot communicate with the primary, but they still request/reply via a backup.
    const t = try TestContext.init(.{ .replica_count = 3 });
    defer t.deinit();

    var c = t.clients(0, t.cluster.clients.len);
    try expectEqual(c.request(20), 20);
    t.replica(.A0).drop_all(.C_, .bidirectional);
    // TODO: https://github.com/tigerbeetledb/tigerbeetle/issues/444
    // try expectEqual(c.request(40), 40);
    try expectEqual(c.request(40), 40);
}
```

That'll protect us from bit-rot, and might be a nice reminder in case we unexpectedly fix the issue.

src/vsr/replica_test.zig
```diff
@@ -0,0 +1,920 @@
...
+    var c = t.clients(0, t.cluster.clients.len);
+    try expectEqual(c.request(20), 20);
+    try expectEqual(t.replica(.R_).commit(), 20);
+
+    var r_a0 = t.replica(.A0);
+    var r_b1 = t.replica(.B1);
+    var r_b2 = t.replica(.B2);
+
+    // B2 cannot see .commit messages, so it lags behind by pipeline_prepare_queue_max commits.
+    r_b1.drop_all(.__, .bidirectional);
```
matklad: Comment for the wrong line?

src/vsr/replica_test.zig
```diff
@@ -0,0 +1,920 @@
...
+    const t = try TestContext.init(.{ .replica_count = 3 });
+    defer t.deinit();
+
+    var c = t.clients(0, t.cluster.clients.len);
+    try expectEqual(c.request(20), 20);
+    try expectEqual(t.replica(.R_).commit(), 20);
+
+    var r_a0 = t.replica(.A0);
+    var r_b1 = t.replica(.B1);
+    var r_b2 = t.replica(.B2);
```
matklad: Maybe we should drop `r_` prefix? It makes it harder to distinguish one from the other.

src/vsr/replica_test.zig
```diff
@@ -0,0 +1,920 @@
...
+    try expectEqual(c.request(checkpoint_trigger_1 - 1), checkpoint_trigger_1 - 1);
+    try expectEqual(r_a0.commit(), checkpoint_trigger_1 - 1);
+    try expectEqual(r_b1.commit(), 20);
+    try expectEqual(r_b2.commit(), checkpoint_trigger_1 - 1);
+
+    r_b2.drop(.__, .incoming, .commit);
+    try expectEqual(c.request(checkpoint_trigger_1 + 1), checkpoint_trigger_1 + 1);
+    try expectEqual(r_a0.op_checkpoint(), checkpoint_1);
+    try expectEqual(r_b1.op_checkpoint(), 0);
+    try expectEqual(r_b2.op_checkpoint(), checkpoint_1);
```
matklad: everything here starts with `try expectEqual`, so it's a bit unclear which operations are state changing, and which are just queries. I wonder if we shoud do something like

```zig
c.request(checkpoint_trigger_1 + 1, checkpoint_trigger_1 + 1)
```

just to make it visually more distinct?

src/vsr/replica_test.zig
```diff
@@ -0,0 +1,920 @@
...
+    t.replica(.R2).corrupt(.{ .wal_prepare = 22 });
+
+    try expectEqual(t.replica(.R_).open(), .ok);
+    // TODO Explicit code coverage marks: This should hit the "quorum received, deadlocked"
+    // log line in on_do_view_change().
+    t.run();
+    try expectEqual(t.replica(.R_).status(), .view_change);
+}
+
+const ProcessSelector = enum {
```
matklad, line 510: Beautiful API

src/vsr/replica_test.zig
```diff
@@ -0,0 +1,920 @@
...
+
+    pub fn init(options: struct {
+        replica_count: u8,
+        standby_count: u8 = 0,
+        client_count: u8 = constants.clients_max,
+    }) !*TestContext {
+        var log_level_original = std.testing.log_level;
+        std.testing.log_level = log_level;
+
+        var prng = std.rand.DefaultPrng.init(123);
```
matklad, line 549: :thinking: I am wondering if this should be genuinely random, to make sure the tests are deterministic for real, and not just accidetally?

src/vsr/replica_test.zig
```diff
@@ -0,0 +1,920 @@
...
+                    }
+                }
+                if (selector == .__ or selector == .C_) {
+                    for (t.cluster.clients) |*client| {
+                        array.appendAssumeCapacity(.{ .client = client.id });
+                    }
+                }
+            },
+        }
+        return array;
```
matklad, line 700:
```suggestion
        assert(array.len > 0);
        return array;
```

src/vsr/replica_test.zig
```diff
@@ -0,0 +1,920 @@
...
+        t: *const TestReplicas,
+        peer: ProcessSelector,
+        direction: LinkDirection,
+        command: vsr.Command,
+    ) void {
+        const paths = t.peer_paths(peer, direction);
+        for (paths.constSlice()) |path| t.cluster.network.link_filter(path).remove(command);
+    }
+
+    const paths_max = constants.nodes_max * (constants.nodes_max + constants.clients_max);
```
matklad:
```suggestion
    const paths_max = constants.nodes_max * (constants.nodes_max - 1 + constants.clients_max);
```

src/vsr/replica_test.zig
```diff
@@ -0,0 +1,920 @@
...
+            var requests_remaining = requests - t.requests;
+            while (requests_remaining > 0) {
+                for (t.clients.constSlice()) |c| {
+                    t.context.client_requests[c] += 1;
+                    requests_remaining -= 1;
+                    if (requests_remaining == 0) break;
+                }
+            }
+            t.requests = requests;
+        }
```
matklad, line 909: It feels like this could be simplified a bit, perhaps to 

```zig
    pub fn request(t: *TestClients, requests: usize) usize {
        assert(t.requests <= requests);
        defer assert(t.requests == requests);

        outer: while (true) {
            for (t.clients.constSlice()) |c| {
                if (t.requests == requests) break :outer;
                t.context.client_requests[c] += 1;
                t.requests += 1;
            }
        }
```

src/stdx.zig
sentientwaffle: I like that they are the same name. I'm not too worried about mixing them up since it only makes a difference during tests.

src/vsr/replica_test.zig
sentientwaffle: Oops, good catch. When some of these tests were first written it actually _did_ matter — corruption had to precede the crash — but I fixed that a while back (https://github.com/tigerbeetledb/tigerbeetle/pull/543).

src/vsr/replica_test.zig
sentientwaffle: :+1: I changed 63 to `checkpoint_trigger_1`.

src/vsr/replica_test.zig
sentientwaffle: Looks like:
1. Ops 22 and 23 are pipelined/prepared concurrent
2. B1 receives 23. (advancing its head)
3. B1 receives 22. Since it precedes the head (23), it considers it a "repair" (`on_repair()`). `on_repair()` doesn't forward to the next replica in the chain.
4. So B2 never sees op 22.
5. B2 can't request any of its missing headers or repairs because it is partitioned, so it is unable to commit (`commit_journal_next: waiting for repair (hash chain)`).

src/vsr/replica_test.zig
sentientwaffle: I was thinking that (eventually) metrics counters might be an even better fit for explicit code coverage, to get a little type safety. (e.g. if metric names are an enum). But I'm listing the log messages for now to make it easier to add later.

matklad: I am wondering if we maybe should decouple logging in replica?

so that, rather that `log.error(“string”, args)` we do `event(.tag, args)`, and it’s event who then makes a format string for log (or directly matches on tag in test)

src/vsr/replica_test.zig
sentientwaffle: Oops, leftover from a different test.

src/vsr/replica_test.zig
sentientwaffle: :thinking: ~Maybe with `Event = union(enum) { tag: args, ... }`?~

The `event()` could handle logging, metrics, tracing, and even [simulator hooks](https://github.com/tigerbeetledb/tigerbeetle/blob/887a01f5ccbae47544d0ddfc7d8ec06a7bf958dc/src/vsr/replica.zig#L333-L345).

src/vsr/replica_test.zig
sentientwaffle: Maybe with a seed from an env variable? I don't actually want to get a random number from the OS here since we want CI to be deterministic (w/ a consistent seed).

src/vsr/replica_test.zig
sentientwaffle: I think depending on the seed, B2 might actually be able to commit 1 or 2 more ops, so I will put
```
    try expect(t.replica(.B2).commit() >= 20);
```
instead of a precise `expectEqual`.

matklad (approved):

PR #613, opened 2023-03-31
vsr: introduce u128 replica ids

matklad:
vsr: introduce u128 replica ids

Before replicas were identified by their index in the configuration.
This is susceptible to operator errors, and does not really work in a
world with reconfiguration (if a replica joins a cluster after being
offline through several reconfigurations, neither the replica nor the
cluster have any idea what this replica's index is).

What we do now instead is identify replica by a globally unique u128
number, which is generated during format.

It's not clear how _exactly_ we generate the ID though. The bullet proof
option is to pick the id randomly, but that makes creating a cluster
less ergonomic in a significant way (as you don't know ids until you
format). An alternative, for initial cluster members, it so pre-generate
ids based on initial replica indexes. It's more convenient, but less
safe.

In this PR I pick the second approach for now, to avoid disturbing  the
CLI until we fully implement reconfiguration. This isn't as useless as
it might seem -- although we generate id based on initial `replica`,
with ID generated and persisted we can now change `replica` over epochs!

## Pre-merge checklist

Performance:

### DIFF
--- a/src/vsr.zig
+++ b/src/vsr.zig
@@ -1097,6 +1097,49 @@ test "quorums" {
     }
 }
 
+/// Deterministically assigns replica_ids for the initial configuration.
+///
+/// Eventualy, we want to identify replicas using random u128 ids to prevent operator errors.
+/// However, that requires unergonomic two-step process for spinning a new cluster up.  To avoid
+/// needlessly compromizing the experience until reconfiguration is fully implemented, derive
+/// replica ids for the initial cluster deterministically.
+pub fn root_members(cluster: u32) [constants.nodes_max]u128 {
+    const IdSeed = packed struct {
+        cluster_config_checksum: u128 = config.configs.current.cluster.checksum(),
+        cluster: u32,
+        replica: u8,
+    };
+
+    var result = [_]u128{0} ** constants.nodes_max;
+    var replica: u8 = 0;
+    while (replica < constants.nodes_max) : (replica += 1) {
+        result[replica] = checksum(std.mem.asBytes(&IdSeed{ .cluster = cluster, .replica = replica }));
+    }
+
+    assert_valid_members(&result);
+    return result;
+}
+
+/// Check that:
+///  - all non-zero elements are different
+///  - all zero elements are trailing
+pub fn assert_valid_members(members: *const [constants.nodes_max]u128) void {
+    for (members) |replica_i, i| {
+        for (members[0..i]) |replica_j| {
+            if (replica_j == 0) assert(replica_i == 0);
+            if (replica_j != 0) assert(replica_j != replica_i);
+        }
+    }
+}
+
+pub fn assert_valid_member(members: *const [constants.nodes_max]u128, replica_id: u128) void {
+    assert(replica_id != 0);
+    assert_valid_members(members);
+    for (members) |member| {
+        if (member == replica_id) break;
+    } else unreachable;
+}
+
 pub const Headers = struct {
     pub const Array = std.BoundedArray(Header, constants.view_change_headers_max);
     /// The SuperBlock's persisted VSR headers.
--- a/src/vsr/superblock.zig
+++ b/src/vsr/superblock.zig
@@ -102,7 +102,7 @@ pub const SuperBlockHeader = extern struct {
     /// The number of headers in vsr_headers_all.
     vsr_headers_count: u32,
 
-    reserved: [3144]u8 = [_]u8{0} ** 3144,
+    reserved: [2936]u8 = [_]u8{0} ** 2936,
 
     /// SV/DVC header suffix. Headers are ordered from high-to-low op.
     /// Unoccupied headers (after vsr_headers_count) are zeroed.
@@ -117,6 +117,16 @@ pub const SuperBlockHeader = extern struct {
         /// The vsr.Header.checksum of commit_min's message.
         commit_min_checksum: u128,
 
+        /// Globally unique identifier of the replica, must be non-zero.
+        replica_id: u128,
+
+        /// Set of replica_ids of cluster members, where order of ids determines replica indexes.
+        ///
+        /// First replica_count elements are active replicas,
+        /// then standby_count standbys, the rest are zeros.
+        /// Order determines ring topology for replication.
+        members: [constants.nodes_max]u128,
+
         /// The last operation committed to the state machine. At startup, replay the log hereafter.
         commit_min: u64,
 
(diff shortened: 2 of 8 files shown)


matklad:
Marking as ready for review: I haven't done self-review yet, so I might want to adjust various details (like not compiling tests :D ), but I think I am ok with overall shape here, so it might make sense to take a look @sentientwaffle!

matklad:
Overall, I have mixed feelings here --- I do believe that that's the right way to go if we optimize for robustness, but the new CLI is pretty ugly. My herd now looks like this:

```zig
const replicas = await par(
  $.read`./tigerbeetle format --cluster=0 --replica-count=3 0_0.tigerbeetle`,
  $.read`./tigerbeetle format --cluster=0 --replica-count=3 0_1.tigerbeetle`,
  $.read`./tigerbeetle format --cluster=0 --replica-count=3 0_2.tigerbeetle`,
);
const configuration = replicas.join(",");

await par(
  $`./tigerbeetle start --addresses=3000,3001,3002 ${`--configuration=${configuration}`} 0_0.tigerbeetle`,
  $`./tigerbeetle start --addresses=3000,3001,3002 ${`--configuration=${configuration}`} 0_1.tigerbeetle`,
  $`./tigerbeetle start --addresses=3000,3001,3002 ${`--configuration=${configuration}`} 0_2.tigerbeetle`,
  $`zig/zig build benchmark -- --account-count 2 --transfer-count 10 --transfer-count-per-second 1`,
);
```

src/vsr.zig
```diff
@@ -1097,6 +1097,47 @@ test "quorums" {
     }
 }
 
+/// Deterministically assigns replica_ids for the initial configuration.
+///
+/// Eventualy, we want to identify replicas using random u128 ids to prevent operator errors.
+/// However, that requires unergonomic two-step process for spinning a new cluster up.  To avoid
+/// needlessly compromizing the experience until reconfiguration is fully implemented, derive
+/// replica ids for the initial cluster deterministically.
+pub fn initial_configuration(cluster: u32) [constants.nodes_max]u128 {
```
matklad: Debating whether I should introduce 

```zig
const Configuration = struct {
  replicas: [constants.nodes_max]u128,
  replica_count: u8,
}
```

already.

The main reason why I don't want to do that is that in the VSR header we want to manually align the fields.

src/vsr.zig
```diff
@@ -1097,6 +1097,47 @@ test "quorums" {
...
+    return result;
+}
+
+/// Check that:
+///  - all non-zero elements are different
+///  - all zero elements are trailing
+pub fn assert_valid_configuration(configuration: *const [constants.nodes_max]u128) void {
+    for (configuration) |replica_i, i| {
+        for (configuration[0..i]) |replica_j| {
+            if (replica_j == 0) assert(replica_i == 0);
```
matklad, line 1129: I do plan to encode length into the configuration by just zeroing out suffix, but in this implementaiton configurations _always_ contain nodes_max entries. This is because we don't want to pass `--standby-count` to `format`, so we just "pre-allocate" all standbys ids. 

Even if we keep "initial cluster memebres get deterministic ids" trick, long term standbys would definitely use random ids, and we would just forbid standbys in the initial config. But, as we can't add sbys dynamically yet, we preallocate instaed.

src/vsr.zig
```diff
@@ -1097,6 +1097,48 @@ test "quorums" {
     }
 }
 
+/// Deterministically assigns replica_ids for the initial configuration.
+///
+/// Eventualy, we want to identify replicas using random u128 ids to prevent operator errors.
+/// However, that requires unergonomic two-step process for spinning a new cluster up.  To avoid
+/// needlessly compromizing the experience until reconfiguration is fully implemented, derive
+/// replica ids for the initial cluster deterministically.
+pub fn initial_configuration(cluster: u32) [constants.nodes_max]u128 {
```
sentientwaffle: Thoughts on the term "root configuration" vs "initial configuration"? Similar to how we have a "root prepare".

src/vsr.zig
```diff
@@ -1097,6 +1097,48 @@ test "quorums" {
...
+    const IdSeed = packed struct {
+        cluster: u32,
+        replica: u8,
+    };
```
sentientwaffle: Should we fold in the cluster config too?

src/vsr.zig
```diff
@@ -1097,6 +1097,48 @@ test "quorums" {
...
+
+    var result = [_]u128{0} ** constants.nodes_max;
+    var replica: u8 = 0;
+    while (replica < constants.nodes_max) : (replica += 1) {
```
sentientwaffle, line 1115: Could be
```
for (&result) |*id, replica| {
    id.* = checksum(std.mem.asBytes(&IdSeed{ .cluster = cluster, .replica = @intCast(u8, replica) }));
}
```
...though with the intcast I'm not sure if it's actually simpler...

src/vsr/superblock.zig
```diff
@@ -117,6 +117,16 @@ pub const SuperBlockHeader = extern struct {
         /// The vsr.Header.checksum of commit_min's message.
         commit_min_checksum: u128,
 
+        /// Globally unique identifier of the replica, must be non-zero.
+        replica_id: u128,
+
+        /// Ordered set of replica_ids of cluster members.
```
sentientwaffle: When I first read this I assumed that it meant that `configuration` is sorted.
What about something like "Set of unique replica_ids, ordered to correspond to their respective replica indexes."?

src/vsr.zig
sentientwaffle: So to see if I understand: Right now `initial_configuration` creates ids for every possible standby, even though the initial configuration is not allowed any standbys?

src/vsr/superblock.zig
```diff
@@ -129,23 +139,26 @@ pub const SuperBlockHeader = extern struct {
...
+            assert(@sizeOf(VSRState) == 256);
             // Assert that there is no implicit padding in the struct.
             assert(@bitSizeOf(VSRState) == @sizeOf(VSRState) * 8);
         }
 
-        pub fn root(options: struct { cluster: u32, replica: u8, replica_count: u8 }) VSRState {
+        pub fn root(options: struct {
+            cluster: u32,
+            replica_id: u128,
+            configuration: [constants.nodes_max]u128,
```
sentientwaffle: I wonder if we should be calling this `vsr_configuration` or `replica_configuration` to disambiguate from e.g. `config.zig`'s configuration.

src/vsr.zig
matklad: > So to see if I understand: Right now initial_configuration creates ids for every possible standby, even though the initial configuration is not allowed any standbys?

Not exactly, it is quite a bit less elegant. 

In this pr, the root configuration looks as if we have a maximum number of standbys configured. However, we ignore the number of standbys implicitly encoded in the configuration, and instead derive that from `--addresses` passed to `.start`.

This is hacky, but avoids the need to encode the number of standbys during format. So, you can format the cluster, start it with 3 standbys, then restart with zero standbys, and that'll work, as the standby count right now is purely in-memory thing. 

But the bigger reason is that before this PR we don't pass `--standby-count` to format, and I want to keep it that way, just to avoid having to adjust user-visible interface while it is still in flux. 

---

Long term, I do want to rely on `superblock.vsr_state.configuration` to tell me the number of standbys, but, to make that useful, we need to have an API to actually change standys dynamically.

src/vsr.zig
matklad: Root configuration is lovely, thanks!

src/vsr.zig
matklad: Wouldn't harm!

src/vsr.zig
matklad: Left explicit indexing it, feels simpler that way.

sentientwaffle (approved):

PR #615, opened 2023-04-03
Add Node steps to client_integration rather than manual setup

eatonphil:
Brings Node CI in line with the Go and Java integration test paths. I also made a mistake in a refactor of build.zig's run_with_tb step so the Node integration tests actually weren't running before. This allowed in a bug after a different PR was merged in the Node client when detecting invalid operations. This PR fixes that bug. Node integration steps are running.

### DIFF
--- a/.github/workflows/clients-node.yml
+++ b/.github/workflows/clients-node.yml
@@ -25,18 +25,6 @@ jobs:
           name: node-binaries
           path: src/clients/node/tigerbeetle-node-*.tgz
 
-      # Integration tests depend on dist rather than the .tgz version
-      # since these files refer to a local directory. There's not a
-      # way to refer to a .tgz file in a package.json file with a
-      # wildcard in it. If we hardcoded a version in the .tgz name of
-      # the package.json dependency then it would break every time we
-      # bumped the version.
-      - name: Upload dist
-        uses: actions/upload-artifact@v3
-        with:
-          name: node-dist
-          path: src/clients/node/dist
-
   test_install:
     # We use self hosted runners for M1 here. See macos.yml for an explaination
     permissions:
@@ -95,6 +83,10 @@ jobs:
         os: [macos-latest, ubuntu-latest]
         sample: [basic, two-phase]
 
+    defaults:
+      run:
+        working-directory: ${{ github.workspace }}
+
     runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v3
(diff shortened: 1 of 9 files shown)


src/clients/integration.zig
```diff
@@ -131,7 +131,7 @@ fn prepare_go_sample_integration_test(
     // Make sure go_client is built, though not necessarily up-to-date.
     const root = try git_root(arena);
     try std.os.chdir(root);
-    if (!file_or_directory_exists(arena, "src/clients/go/pkg/native")) {
+    if (!file_or_directory_exists(arena, "src/clients/go/pkg/native/x86_64-linux")) {
```
eatonphil, line 134: The previous check wasn't actually checking an uncommitted directory. native will always be there. The arch-specific folders are not committed, which is what we want to check before rebuilding.

src/clients/node/src/node.zig
```diff
@@ -579,7 +579,11 @@ fn request(env: c.napi_env, info: c.napi_callback_info) callconv(.C) c.napi_valu
     const context = contextCast(context_raw.?) catch return null;
     const operation_int = translate.u32_from_value(env, argv[1], "operation") catch return null;
 
-    if (operation_int >= @typeInfo(Operation).Enum.fields.len) {
+    const greatest_enum_value = @enumToInt(@field(
+        Operation,
+        @typeInfo(Operation).Enum.fields[@typeInfo(Operation).Enum.fields.len - 1].name,
+    ));
+    if (operation_int > greatest_enum_value) {
```
sentientwaffle: I think you could use `@intToEnum(vsr.Operation, operation).valid(StateMachine)`, which is a little more precise of a check.

eatonphil: Interesting, I wasn't sure what would happen if you tried casting an integer outside of the enum value space. If it would crash or something.

sentientwaffle: I _think_ in this case it won't only because `vsr.Operation` is a non-exhaustive enum (note the `_,` variant): 

https://github.com/tigerbeetledb/tigerbeetle/blob/e5bf5393e7a658cf5ca6fc4d7291cedc83bf9ec6/src/vsr.zig#L126-L137

And https://ziglang.org/documentation/master/#toc-Non-exhaustive-enum:

> [@intToEnum](https://ziglang.org/documentation/master/#intToEnum) on a non-exhaustive enum involves the safety semantics of [@intCast](https://ziglang.org/documentation/master/#intCast) to the integer tag type, but beyond that always results in a well-defined enum value.

(But I haven't tested it)

eatonphil: Cool, got it working and also tested that it still catches an invalid operation. Thanks!

sentientwaffle: Perfect :+1:

.github/workflows/clients-node.yml
```diff
@@ -95,19 +83,22 @@ jobs:
...
-      - run: cd ../../../ && ./scripts/install.sh
+          name: node-binaries
+          path: src/clients/node
 
-      # Set up sample dependencies
-      - run: cd samples/${{ matrix.sample }} && npm install
+      # Grab Zig
+      - run: ./scripts/install.${{ matrix.os == 'windows-latest' && 'bat' || 'sh' }}
 
-      - run: cd ../../../ && R_CWD=$(pwd)/src/clients/node/samples/${{ matrix.sample }} ./scripts/build.sh run_with_tb -- node main.js
```
eatonphil, line 113: In a previous PR I changed the run_with_tb build step not to actually run when it's built. That's why this break got in.

https://github.com/tigerbeetledb/tigerbeetle/blob/main/build.zig#L762

sentientwaffle (commented):
`node.zig` changes LGTM

cb22 (approved):

eatonphil:
Thanks Fed and DJ!

PR #616, opened 2023-04-03
VSR/SuperBlock: Fix client table recovery

sentientwaffle:
By not resetting the `copy` prior to `read_client_table()`, the recovery potentially skips over multiple (valid!) copies of the client table trailer. This can trigger an incorrect `superblock client table lost` panic.

(Bug discovered by the simulator, but in a different branch, so there isn't a seed here).

## Pre-merge checklist

### DIFF
--- a/src/vsr/superblock.zig
+++ b/src/vsr/superblock.zig
@@ -1413,6 +1413,7 @@ pub fn SuperBlockType(comptime Storage: type) type {
                 superblock.verify_manifest_blocks_are_acquired_in_free_set();
 
                 // TODO Repair any impaired copies before we continue.
+                context.copy = 0;
                 superblock.read_client_table(context);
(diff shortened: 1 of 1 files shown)


src/vsr/superblock.zig
```diff
@@ -1413,6 +1413,7 @@ pub fn SuperBlockType(comptime Storage: type) type {
                 superblock.verify_manifest_blocks_are_acquired_in_free_set();
 
                 // TODO Repair any impaired copies before we continue.
+                context.copy = 0;
```
matklad, line 1416: A good one!

Don't see a super great way to make this bug jump out, only something relatively heavy-weigh like

```zig
            copy: ?union(enum) {
                cient_table: u8,
                free_set: u8,
            } = null,
```

matklad (approved):

PR #618, opened 2023-04-03
VSR: Nack during DVC-phase of view-change, not repair-phase

sentientwaffle:
Move the nack-prepare phase into the DVC step.
How this works:
- Use `dvc.header.context` as a "missing" bitset to indicate nack-able (or situationally-nackable) messages (see DVCQuorum's updated docs for detail).
- Don't send "fault" entries in the DVC headers anymore — send the real header if available, and a blank otherwise.
- The new primary uses the combination of the "missing" bitset with the headers/blanks to nack+truncate before the repair phase of the view change even begins.

Note that this design means that `pipeline_prepare_queue_max` now has a hard limit of `@bitSizeOf(u128) - 1`.
We won't use a value that high even in production.

## Pre-merge checklist

Performance:

### DIFF
--- a/src/vsr.zig
+++ b/src/vsr.zig
@@ -113,7 +113,6 @@ pub const Command = enum(u8) {
     request_headers,
     request_prepare,
     headers,
-    nack_prepare,
 
     eviction,
 
@@ -227,7 +226,6 @@ pub const Header = extern struct {
     ///   in the message body which it has definitely not prepared (i.e. "missing").
     ///   The corresponding header may be an actual prepare header, or it may be a "blank" header.
     /// * A `request_prepare` sets this to the checksum of the prepare being requested.
-    /// * A `nack_prepare` sets this to the checksum of the prepare being nacked.
     ///
     /// This allows for cryptographic guarantees beyond request, op, and commit numbers, which have
     /// low entropy and may otherwise collide in the event of any correctness bugs.
@@ -353,7 +351,6 @@ pub const Header = extern struct {
             .request_prepare => self.invalid_request_prepare(),
             .request_block => null, // TODO
             .headers => self.invalid_headers(),
-            .nack_prepare => self.invalid_nack_prepare(),
             .eviction => self.invalid_eviction(),
             .block => null, // TODO
         };
@@ -648,19 +645,6 @@ pub const Header = extern struct {
         if (self.op != 0) return "op != 0";
         if (self.commit != 0) return "commit != 0";
         if (self.timestamp != 0) return "timestamp != 0";
-        if (self.operation != .reserved) return "operation != .reserved";
-        return null;
-    }
-
-    fn invalid_nack_prepare(self: *const Header) ?[]const u8 {
-        assert(self.command == .nack_prepare);
-        if (self.parent != 0) return "parent != 0";
-        if (self.client != 0) return "client != 0";
-        if (self.request != 0) return "request != 0";
-        if (self.commit != 0) return "commit != 0";
-        if (self.timestamp != 0) return "timestamp != 0";
-        if (self.checksum_body != checksum_body_empty) return "checksum_body != expected";
-        if (self.size != @sizeOf(Header)) return "size != @sizeOf(Header)";
         if (self.operation != .reserved) return "operation != .reserved";
         return null;
     }
@@ -1078,6 +1062,8 @@ pub fn quorums(replica_count: u8) struct {
     assert(quorum_view_change >= @divFloor(replica_count, 2) + 1);
     assert(quorum_view_change + quorum_replication > replica_count);
 
+    // We need to have enough nacks to guarantee that `quorum_replication` was not reached,
+    // because if the replication quorum was reached, then it may have been committed.
     const quorum_nack_prepare = replica_count - quorum_replication + 1;
     assert(quorum_nack_prepare + quorum_replication > replica_count);
 
--- a/src/vsr.zig
+++ b/src/vsr.zig
@@ -1325,6 +1325,7 @@ const ViewChangeHeadersArray = struct {
         // checkpoint, so the start_view has a full suffix of headers.
         assert(headers.array.get(0).op >= constants.journal_slot_count);
         assert(headers.array.len >= constants.view_change_headers_suffix_max);
+        assert(headers.array.len >= constants.pipeline_prepare_queue_max + 1);
 
         const commit_max = std.math.max(
             headers.array.get(0).op -| constants.pipeline_prepare_queue_max,
@@ -1340,10 +1341,7 @@ const ViewChangeHeadersArray = struct {
         //   (SV headers are determined by view_change_headers_suffix_max,
         //   but DVC headers must stop at commit_max.)
         headers.command = .do_view_change;
-        headers.array.len = std.math.min(
-            headers.array.len,
-            constants.pipeline_prepare_queue_max + 1,
-        );
+        headers.array.len = constants.pipeline_prepare_queue_max + 1;
 
         headers.verify();
     }
--- a/src/vsr.zig
+++ b/src/vsr.zig
@@ -223,6 +223,9 @@ pub const Header = extern struct {
     /// * A `prepare` sets this to the checksum of the client's request.
     /// * A `prepare_ok` sets this to the checksum of the prepare being acked.
     /// * A `commit` sets this to the checksum of the latest committed prepare.
+    /// * A `do_view_change` sets this to a bitset, with set bits indicating headers
+    ///   in the message body which it has definitely not prepared (i.e. "missing").
+    ///   The corresponding header may be an actual prepare header, or it may be a "blank" header.
     /// * A `request_prepare` sets this to the checksum of the prepare being requested.
     /// * A `nack_prepare` sets this to the checksum of the prepare being nacked.
     ///
@@ -581,7 +584,6 @@ pub const Header = extern struct {
         assert(self.command == .do_view_change);
         if (self.parent != 0) return "parent != 0";
         if (self.client != 0) return "client != 0";
-        if (self.context != 0) return "context != 0";
         if (self.operation != .reserved) return "operation != .reserved";
         return null;
     }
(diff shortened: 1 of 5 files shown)


src/vsr.zig
```diff
@@ -1076,6 +1062,8 @@ pub fn quorums(replica_count: u8) struct {
     assert(quorum_view_change >= @divFloor(replica_count, 2) + 1);
     assert(quorum_view_change + quorum_replication > replica_count);
 
+    // We need to have enough nacks to guarantee that `quorum_replication` was not reached,
+    // because if the replication quorum was reached, then it may have been committed.
```
matklad, line 1066:
```suggestion
    // because if the replication quorum was reached, then it is committed.
```

It is impossible for an entry to not be committed if it collected replication quorum of prepare-oks, right?

src/vsr/replica.zig
```diff
@@ -6541,36 +6336,51 @@ pub fn ReplicaType(
...
+///   - Yes: The corresponding header in the message body was definitely not
+///     prepared in the latest view. (The corresponding header may be blank or ¬blank).
+///   - No: The corresponding header in the message body was either prepared during the latest view,
+///     or _might_ have been prepared, but due to WAL corruption we can't tell.
+/// - Nack: based on Blank/Missing, whether the new primary counts it as a nack.
+/// - The header corresponding to the sender's `replica.op` is always "valid", never a "blank".
 ///   (Otherwise the replica would be in status=recovering_head and unable to participate).
 ///
-/// Across all DVCs in the quorum:
+/// Invariants (across all DVCs in the quorum):
```
matklad, line 6391: Let's add invariant which was violated in that op 68 case:

> Any pipeline message which could have been committed is included in some canonical message

src/vsr/replica.zig
```diff
@@ -6541,36 +6336,51 @@ pub fn ReplicaType(
...
-/// - The header corresponding to `replica.op` is always "valid", never a "blank"/"fault".
+///     the DVC must include a valid/blank header for every op in that range.
+///   - When `replica.commit_max > replica.op`, only a single header is included
+///     (`replica.commit_max` if available in the SV, otherwise `replica.op`).
+///   - (The DVC will need a valid header corresponding to its `commit_max` to complete, since the
+///     entire pipeline may be truncated, and the new primary still needs a header for its head op.)
+///
+/// Each header in the DVC body is one of:
+///
+/// | Blank | Missing |  Nack | Description
```
matklad: Some slight misgivings about terminology here, as "Missing" I think doesn't capture what we want here (that replica guarantees that it didn't send prepare-ok).

It feels like the right name for missing is probably "nack", as that's replcia acknowledging not sending  the ok out. 

The third column is an observation that we can derive nacks for different non-blank messages.

src/vsr/replica.zig
```diff
@@ -3092,6 +2932,29 @@ pub fn ReplicaType(
...
+                                missing.set(i);
+                            }
+                        } else {
+                            missing.set(i);
+                        }
+                    } else {
+                        if (!self.journal.faulty.bit(slot)) {
+                            missing.set(i);
+                        }
+                    }
```
matklad: Having  bit of a hard time convincing myself that the case analysis is correct here. In particular, in 

```zig
else {
    missing.set(i);
}
```

branch, shouldn't we check that the faulty bit?

matklad: Ah, it shouldn't, because, even if it faulty, we don't care, because we know it to be different. Maybe it makes sense to try to flatten this out a bit?

```zig
const slot = self.journal.slot_for_op(header.op);
const journal_header = self.journal.header_for_op(header.op);
const dirty = self.journal.dirty.bit(slot);
const faulty = self.journal.faulty.bit(slot);

// Case 1: we have this prepare in memory, but haven't persisted it to disk yet.
if (journal_header != null and journal_header.?.checksum == header.checksum and dirty and !faulty) {
    missing.set(i);
}
// Case 2: we have a _different_ prepare, safe to nack even if it is faulty.
if (journal_header != null and journal_header.?.checksum != header.checksum) {
    missing.set(i);
}
// Case 3: we don't have a prepare all, and that's not due to a fault.
if (journal_header == null and !faulty) {
    missing.set(i);
}
```

matklad: Also, the difference between `header_for_op/header_with_op` is subtle. I woneder if we should use `header_for_slot` instead?

matklad: Also, for case2, can we assert that that's an older prepare? That it is either for a different wrap, or has a smaller view? (not sure if both are possible here)

src/vsr/replica.zig
```diff
@@ -6827,40 +6656,47 @@ const DVCQuorum = struct {
...
-                        .fault => {},
-                        .valid => break true,
-                    }
-                }
-            } else false;
+                    // The replica's prepare is available, but for a different header.
+                    .valid => found != null and found.?.checksum != header.checksum,
+                    // The replica's prepare is faulty, with an unknown header.
+                    .blank => false,
+                });
```
matklad: Doesn't look to pretty, maybe an explicit if would read better here?

src/vsr/replica.zig
```diff
@@ -6802,23 +6619,35 @@ const DVCQuorum = struct {
...
+
+                const headers = message_body_as_view_headers(dvc);
+                const header_index = dvc.header.op - op;
+                assert(header_index <= headers.slice.len);
+
+                const header = &headers.slice[header_index];
+                assert(header.op == op);
+
+                if (vsr.Headers.dvc_header_type(header) == .valid) break header;
+            } else null;
```
matklad: So here we only check that we have a header for an op, not necessary the body, right? can this create an availability issue?

That is, can we be in a situation such that:

* if we looked at DVCs from the whole cluster, we would have truncated the op
* but, looking at the DVC for the quorum we've received, we don't _nack_ this header, so it gets to be a new head
* so we send an SV with a message which should have been nacked? 

Should we also perhaps check that the `found` header is not `missing`?

matklad: Yeah, I think it is an issue under a selective partition, when only some of the messages get through, and that just checking for "is not `missing`" is not going to help. 

Step 1, R0 is the primary, it prepares 1, 2, and replicates 2, but not 1

```
R0 1 2
R1   2
R2   2
```

Step 2, view change, R1 receives DVC from itself and R0. The new head is 2. 1 is found, and is not nacked (we need R2's for that). So R1 sends `SV 1 2`.

Step 3, R0 gets partitioned in such a way that its DVCs go through, but not repairs. So any subsequent view change where R0 participates gets stuck, because it doesnt' truncate  1, but can't repair it either. 

:thinking: it seems like we need that nasty "head can go backwards" bit if we want to make this 100% bullet-proof.

matklad: I guess it can be less weird if `R0` crashes and corrupts `1`. Then any view change where it participates becomes stuck. Although _that_ case I think we can fix with checking for !missing.

sentientwaffle: I agree we need to check that `found` is not `missing`. (I'm thinking the DVC quorum will keep waiting for DVCs if needed).

> Step 3, R0 gets partitioned in such a way that its DVCs go through, but not repairs. So any subsequent view change where R0 participates gets stuck, because it doesnt' truncate 1, but can't repair it either.

In step 2, R1 sent `SV 1 2` — if R1 sent an SV, that means that it finished repair, so by the end of step 2, the state is

```
R0 1 2
R1 1 2
R2   2
```

But I think your example applies if R1 didn't finish repair (and didn't send a SV), since its state would still be

```
R0 1 2
R1   2
R2   2
```

In that case, a view-change with R1/R2 would truncate 1,2.
But any view change involving R0 would require a repair of 1.

But isn't this outside our fault model though? (`(f+1)/(2f+1)` faults)
I guess we could artificially delay completion of the DVC quorum to allow all non-replicas a chance to register their nacks :thinking:

src/vsr.zig
sentientwaffle: There might be ops preceding it in the pipeline which do not have a replication quorum of prepare-oks.
Or alternately, it the prepare oks might have been lost, and the op is "logically" committed (i.e. it will eventually be committed), but no replica has physically executed it in that state machine yet.

src/vsr/replica.zig
matklad: >But isn't this outside our fault model though? ((f+1)/(2f+1) faults)

I am not sure. If R0 is thoroughly dead, that's definitely inside our fault model (1/3 is unavailable) and works. If R0 is more available than fully dead, it feels like it shouldn't make the matters worse. I think that, if we have a valid quorum, it should be able to make progress even if other nodes behave in arbitrary non-byzantine ways. 

>I guess we could artificially delay completion of the DVC quorum to allow all non-replicas a chance to register their nacks thinking

I guess we might actually get that out-of-the-box if we wait for !missing and found?

src/vsr.zig
matklad: Wrong: it might be truncated if some preceding entry is not committed.

src/vsr/replica.zig
sentientwaffle: > I guess we might actually get that out-of-the-box if we wait for !missing and found?

Right, the only issue left is if R0 is partially available — participating in DVCs, but not helping with repair.
But then again, maybe that would be ok too — eventually the view-change would cycle over to R0 as primary, and it already _has_ op 1.

matklad: >eventually the view-change would cycle over to R0 as primary, and it already has op 1.

Right, if it's not corrupted, we'll progress, and if it _is_ corrupted, !missing would cause us to wait for full DVC set&truncate. Sounds good!

src/vsr/replica.zig
sentientwaffle: > Also, for case2, can we assert that that's an older prepare? That it is either for a different wrap, or has a smaller view? (not sure if both are possible here)

Tricky case. I think we can't assert this, since successive view-changes can swing the version of an op forward-and-back, and crash/recovery restores the journal state. So even though the header in the view_headers is logically newer — i.e. part of a higher log_view — it might have originated in an older view.

matklad (approved):

PR #622, opened 2023-04-06
IO/Storage: CPU-bound and cross-thread scheduling

kprotty:
This introduces the concept of an `intent` enum for `Storage.on_next_tick` which controls how the NextTick is scheduled. The enum includes: `yield` which defaults to `IO.timeout(0)` as usual, `cpu_work` which runs the NextTick on a thread pool, and `cpu_inject` which brings the NextTick back from a random (pool) thread to the IO thread. This is done with `tb_client.Signal` which now uses proper IO events (eventfd, EVFILT_USER, PostQueuedCompletionStatus) instead of a hacky, local TCP server.

The following LSM bits now utilize `cpu_work` to offload heavy tasks from the main thread:
* Sorting of `TableMutable` to `TableImmutable`.
* `TableBuilder.data_block_finish` in `Compaction` (this calls `vsr.calculate_checksum_body`)
* `Grid.read_block_validate` after a `Storage.read_sectors` occurs (this also calls `vsr.calculate_checksum_body`).

## Pre-merge checklist

Performance:

* [x] Compare `zig benchmark` on linux before and after this pr.
    ``` sh
    # benchmark results before
    1223 batches in 135.30 s
    load offered = 1000000 tx/s
    load accepted = 73908 tx/s
    batch latency p00 = 0 ms
    batch latency p10 = 28 ms
    batch latency p20 = 29 ms
    batch latency p30 = 30 ms
    batch latency p40 = 30 ms
    batch latency p50 = 31 ms
    batch latency p60 = 32 ms
    batch latency p70 = 33 ms
    batch latency p80 = 34 ms
    batch latency p90 = 36 ms
    batch latency p100 = 5517 ms
    transfer latency p00 = 0 ms
    transfer latency p10 = 4166 ms
    transfer latency p20 = 10805 ms
    transfer latency p30 = 21106 ms
    transfer latency p40 = 34218 ms
    transfer latency p50 = 49408 ms
    transfer latency p60 = 64719 ms
    transfer latency p70 = 74813 ms
    transfer latency p80 = 90404 ms
    transfer latency p90 = 106386 ms
    transfer latency p100 = 125316 ms
    
    # benchmark results after
    1223 batches in 82.40 s
    load offered = 1000000 tx/s
    load accepted = 121362 tx/s
    batch latency p00 = 1 ms
    batch latency p10 = 27 ms
    batch latency p20 = 28 ms
    batch latency p30 = 29 ms
    batch latency p40 = 30 ms
    batch latency p50 = 32 ms
    batch latency p60 = 33 ms
    batch latency p70 = 34 ms
    batch latency p80 = 35 ms
    batch latency p90 = 37 ms
    batch latency p100 = 3281 ms
    transfer latency p00 = 1 ms
    transfer latency p10 = 2101 ms
    transfer latency p20 = 5877 ms
    transfer latency p30 = 11111 ms
    transfer latency p40 = 17961 ms
    transfer latency p50 = 26298 ms
    transfer latency p60 = 35204 ms
    transfer latency p70 = 44132 ms
    transfer latency p80 = 53201 ms
    transfer latency p90 = 62509 ms
    transfer latency p100 = 72410 ms
    ```

### DIFF
--- a/src/clients/c/tb_client/signal.zig
+++ b/src/clients/c/tb_client/signal.zig
@@ -13,274 +13,73 @@ const log = std.log.scoped(.tb_client_signal);
 /// It does this by using OS sockets (which are thread safe) 
 /// to resolve IO.Completions on the tigerbeetle thread.
 pub const Signal = struct {
-    io: *IO,
-    server_socket: os.socket_t,
-    accept_socket: os.socket_t,
-    connect_socket: os.socket_t,
-
-    completion: IO.Completion,
-    recv_buffer: [1]u8,
-    send_buffer: [1]u8,
-
-    on_signal_fn: fn (*Signal) void,
-    state: Atomic(enum(u8) {
-        running,
+    const State = enum(u8) {
         waiting,
         notified,
-        shutdown,
-    }),
+        running,
+    };
+
+    io: *IO,
+    on_signal_fn: fn (*Signal) void,
+
+    event: u32,
+    event_completion: IO.Completion,
+
+    state: Atomic(State),
+    terminated: Atomic(bool),
 
     pub fn init(self: *Signal, io: *IO, on_signal_fn: fn (*Signal) void) !void {
         self.io = io;
-        self.server_socket = os.socket(
-            os.AF.INET,
-            os.SOCK.STREAM | os.SOCK.NONBLOCK,
-            os.IPPROTO.TCP,
-        ) catch |err| {
-            log.err("failed to create signal server socket: {}", .{err});
-            return switch (err) {
-                error.PermissionDenied, error.ProtocolNotSupported, error.SocketTypeNotSupported, error.AddressFamilyNotSupported, error.ProtocolFamilyNotAvailable => error.NetworkSubsystemFailed,
-                error.ProcessFdQuotaExceeded, error.SystemFdQuotaExceeded, error.SystemResources => error.SystemResources,
-                error.Unexpected => error.Unexpected,
-            };
-        };
-        errdefer os.closeSocket(self.server_socket);
-
-        // Windows requires that the socket is bound before listening
-        if (builtin.target.os.tag == .windows) {
-            const addr = std.net.Address.initIp4(.{ 127, 0, 0, 1 }, 0); // zero port lets the OS choose
-            os.bind(self.server_socket, &addr.any, addr.getOsSockLen()) catch |err| {
-                log.err("failed to bind the server socket to a local random port: {}", .{err});
-                return switch (err) {
-                    error.AccessDenied => unreachable,
-                    error.AlreadyBound => unreachable,
-                    error.AddressInUse, error.AddressNotAvailable => unreachable,
-                    error.SymLinkLoop => unreachable,
-                    error.NameTooLong => unreachable,
-                    error.FileNotFound, error.FileDescriptorNotASocket => unreachable,
-                    error.NotDir => unreachable,
-                    error.ReadOnlyFileSystem => unreachable,
-                    error.SystemResources, error.NetworkSubsystemFailed, error.Unexpected => |e| e,
-                };
-            };
-        }
-
-        os.listen(self.server_socket, 1) catch |err| {
-            log.err("failed to listen on signal server socket: {}", .{err});
-            return switch (err) {
-                error.AddressInUse => unreachable,
-                error.FileDescriptorNotASocket => unreachable,
-                error.AlreadyConnected => unreachable,
-                error.SocketNotBound => unreachable,
-                error.OperationNotSupported, error.NetworkSubsystemFailed => error.NetworkSubsystemFailed,
-                error.SystemResources => error.SystemResources,
-                error.Unexpected => error.Unexpected,
-            };
-        };
-
-        var addr = std.net.Address.initIp4(undefined, undefined);
-        var addr_len = addr.getOsSockLen();
-        os.getsockname(self.server_socket, &addr.any, &addr_len) catch |err| {
-            log.err("failed to get address of signal server socket: {}", .{err});
-            return switch (err) {
-                error.SocketNotBound => unreachable,
-                error.FileDescriptorNotASocket => unreachable,
-                error.SystemResources => error.SystemResources,
-                error.NetworkSubsystemFailed => error.NetworkSubsystemFailed,
-                error.Unexpected => error.Unexpected,
-            };
-        };
-
-        self.connect_socket = self.io.open_socket(
-            os.AF.INET,
-            os.SOCK.STREAM,
-            os.IPPROTO.TCP,
-        ) catch |err| {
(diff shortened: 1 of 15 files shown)


src/clients/c/tb_client/signal.zig
```diff
@@ -13,274 +13,73 @@ const log = std.log.scoped(.tb_client_signal);
 /// It does this by using OS sockets (which are thread safe)
```
cb22: Believe this comment is out of date now?

src/io/linux.zig
```diff
@@ -960,6 +960,84 @@ pub const IO = struct {
...
+                switch (on_event(_completion)) {
+                    .listen => _completion.io.listen_event(_event, _completion, on_event),
+                    .cancel => {},
+                }
+            }
+        };
+
+        self.read(
+            *anyopaque,
+            @intToPtr(*anyopaque, event),
```
jamii: Why does io.read take a type argument if the only type that works is `*anyopaque`?

src/io/linux.zig
```diff
@@ -960,6 +960,84 @@ pub const IO = struct {
...
+        // eventfd initialized with no (zero) previous write value,
+        const event_fd = os.eventfd(0, linux.EFD.CLOEXEC) catch |err| switch (err) {
+            error.SystemResources, error.SystemFdQuotaExceeded, error.ProcessFdQuotaExceeded => return error.SystemResources,
+            error.Unexpected => return error.Unexpected,
+        };
+        assert(event_fd != INVALID_EVENT);
+        errdefer os.close(event_fd);
+
+        // Start listening for notifications on this event.
+        const event = @intCast(u32, event_fd);
```
jamii: Why cast this from `fd_t`?

src/storage.zig
```diff
@@ -69,77 +71,178 @@ pub const Storage = struct {
...
     };
 
+    const ThreadPool = struct {
+        mutex: std.Thread.Mutex = .{},
+        cond: std.Thread.Condition = .{},
+        queue: FIFO(NextTick) = .{ .name = "thread-pool" },
+        notified: bool = false,
+        idle: u16 = 0,
+
+        var instance = ThreadPool{};
```
jamii: Why does this need to be global, rather than a field on storage?

src/storage.zig
```diff
@@ -69,77 +71,178 @@ pub const Storage = struct {
...
+                pool.notified = true;
+                pool.cond.signal();
+            }
+            pool.mutex.unlock();
+        }
+    };
+
     io: *IO,
     fd: os.fd_t,
+    main_thread_id: std.Thread.Id,
```
jamii, line 190: Can you use this for `assert(storage.is_background_thread())` in callbacks that are run in the threadpool (eg `cpu_sort_callback`), to call out that they need to be read carefully.

src/storage.zig
```diff
@@ -69,77 +71,178 @@ pub const Storage = struct {
...
+        assert(storage.is_local());
         storage.io.tick() catch |err| {
             log.warn("tick: {}", .{err});
             std.debug.panic("io.tick(): {}", .{err});
         };
     }
 
+    pub const NextTickIntent = enum {
+        yield,
+        cpu_work,
```
jamii: `yield/work/inject` is not super obvious to me. Maybe `same_thread/move_to_foreground_thread/move_to_background_thread`?

jamii: Similarly `is_local` doesn't connect to `work/inject`. Maybe `is_foreground_thread`.

src/storage.zig
```diff
@@ -69,77 +71,178 @@ pub const Storage = struct {
...
+        cpu_inject,
+    };
+
     pub fn on_next_tick(
```
jamii, line 247: In the doc comment here, could you explain the rules we have to follow for thread safety.

src/testing/storage.zig
```diff
@@ -275,11 +275,24 @@ pub const Storage = struct {
...
+        cpu_inject,
+    };
+
     pub fn on_next_tick(
         storage: *Storage,
         callback: fn (next_tick: *Storage.NextTick) void,
         next_tick: *Storage.NextTick,
+        intent: NextTickIntent,
     ) void {
+        // All on_next_tick requests are fullened through next_tick_queue for determinism.
```
jamii: 'fulfilled'?

jamii:
@kprotty I added some tracing. 

image

Utilization is kind of low. I wonder if that's due to the control plane not supplying enough work, or to the overhead of communication.

jamii:
`tracer` is not thread-safe because of https://github.com/tigerbeetledb/tigerbeetle/blob/6e4ed3bacdb865f1b71979a26ac601cdce7c7875//src/tracer.zig#L337. Since the yield and thread-pool queue are traced, this can cause UB and crashes when tracing benchmarks.

Making the interns and message_buffer thread-local might be sufficient? Tracy itself is thread-safe.

src/io/linux.zig
kprotty: Wanted a unified "event" type across impls so went with `u32`. I can go with i32 so fd_t is compatible.

src/storage.zig
kprotty: Good note. Accidentally carried over from when multiple storages was up in the air and Storage didn't support intrusive initialization.

src/storage.zig
kprotty: This would need to be exposed/work on `testing/storage.zig` too (as Tree is used on both Storage types). But test storage doesn't use threads.

src/io/linux.zig
kprotty: It only supports a pointer type, hence the use of `anyopaque` just for representing the `eventfd` as pointer.

jamii: Oh, I saw that it doesn't cast `context` [here](https://github.com/tigerbeetledb/tigerbeetle/blob/0b15d2209f6d51f2f56e7f7a597540b78ddacdf6//src/io/linux.zig#L652), but I didn't know that there is an implicit conversion from *T to *anyopaque.

src/io/linux.zig
jamii: I guess the cast can't fail because the stdlib already checked for errors?

src/storage.zig
jamii: Good point. We could have testing/storage store a boolean flag that's true when it's executing a background callback?

Just thinking ahead to when I inevitably mess up thread safety :)

src/io/linux.zig
kprotty: Yea, although I've changed it to be i32 now for compatibility.

src/storage.zig
kprotty: Thoughts on `main_thread` instead of `foreground_thread`?

jamii: Sure :)

kprotty:
We decided to focus on improving single threaded performance first. There's still a few more things to implement there.

PR #624, opened 2023-04-07
WIP: reconfiguration

matklad:
First stab at reconfiguration, basically looking around at what would break if we do reconfig. A bunch of stuff breaks!

The interesting bits here are:

* The Readme update
* `epoch_initiate_new`, `epoch_switch_to_new` and `epoch_retire_old` family of functions in replica
* the test in `replica_tests` which demonstrates that we can actually switch epoch 

The overall flow is:

* we _initiate_ new epoch when we prepare a `.reconfigure` (primary additionaly promises to not prepare anything else in this epoch)
* we _switch_ to the new epoch when we commit
* we _retire_ old epoch once we get a checkpoint in the new epoch

---

My overall plan here is 

* implement simplest possible approach from VRR paper without any optimization and ignoring repairs
* (i wanted to initially skip the persistance bit, and test with non-crashing replicas, but it seems we have enough assertions in the superblock that just doing persistance is easier)
* implement cross-epoch repair
* implement reconfiguration safety checks (ie, primary rejects reconfigure request if standbys are not caught up)
* implement cross-epoch pipelining (so that, rather stopping prepares, we send them in the next epoch)
* implement a faster way to retire old epoch then waiting for a checkpoint
* implement adding and removing standbys once we have state sync

### DIFF
--- a/src/clients/c/tb_client.h
+++ b/src/clients/c/tb_client.h
@@ -148,6 +148,7 @@ typedef struct tb_create_transfers_result_t {
 } tb_create_transfers_result_t;
 
 typedef enum TB_OPERATION {
+    TB_OPERATION_RECONFIGURE = 3,
     TB_OPERATION_CREATE_ACCOUNTS = 128,
     TB_OPERATION_CREATE_TRANSFERS = 129,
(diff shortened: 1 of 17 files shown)


matklad:
Will get to this one day :)

PR #630, opened 2023-04-11
Add dry run release to pipeline, fix build

eatonphil:

### DIFF
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -1,5 +1,5 @@
 #!/usr/bin/env sh
-set -e
+set -ex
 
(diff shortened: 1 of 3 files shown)


scripts/build.sh
```diff
@@ -15,6 +15,7 @@ case "$*" in
 	;;
 esac
 
+TARGET=""
```
eatonphil, line 18: This is the actual fix.

eatonphil: TARGET is set and then that variable is reused in here for something else. So this resets.

cb22 (approved):

PR #631, opened 2023-04-13
LSM and VSR glossary updates

eatonphil:
Maybe we don't want to make this glossary bigger or more verbose @sentientwaffle so we can close this if so. Otherwise was just trying to clarify for myself as I'm looking.

### DIFF
--- a/src/lsm/README.md
+++ b/src/lsm/README.md
@@ -3,13 +3,13 @@
 - _bar_/_measure_: `lsm_batch_multiple` beats; unit of incremental compaction.
 - _beat_: `op % lsm_batch_multiple`; Single step of an incremental compaction.
 - _groove_: A collection of LSM trees, storing objects and their indices.
-- _immutable table_: in-memory table; one per tree. Used to periodically flush the mutable table to
+- _immutable table_: In-memory table; one per tree. Used to periodically flush the mutable table to
   disk.
-- _level_: Between `0` and `lsm_levels - 1` (usually `lsm_levels = 7`).
-- _forest_: a collection of grooves.
-- _manifest_: index of table and level metadata; one per tree.
-- _mutable table_: in-memory table; one per tree. All tree updates are applied only to this table.
-- _snapshot_: sequence number which selects the queryable partition of on-disk tables.
+- _level_: A collection of on-disk tables, numbering between `0` and `config.lsm_levels - 1` (usually `config.lsm_levels = 7`).
+- _forest_: A collection of grooves.
+- _manifest_: Index of table and level metadata; one per tree.
+- _mutable table_: In-memory table; one per tree. All tree updates (e.g. `Tree.put`) directly modify just this table.
+- _snapshot_: Sequence number which selects the queryable partition of on-disk tables.
 
 # Tree
 ## Tables
--- a/src/vsr/README.md
+++ b/src/vsr/README.md
@@ -1,3 +1,17 @@
+# Glossary
+
+- _checkpoint_: Ensure that all updates from the past wrap of the WAL are durable in the _grid_, then advance the replica's recovery point by updating the superblock. After a checkpoint, the checkpointed WAL entries are safe to be overwritten by the next wrap. (Sidenote: in consensus literature this is sometimes called snapshotting. But we use that term to mean something else.)
+- _grid_: The zone on disk where LSM trees and metadata for them reside.
+- _header_: Identifier for many kinds of messages, including each entry in the VSR log. Passed around instead of the entry when the full entry is not needed (such as view change).
+- _journal_: The in-memory data structure that manages the WAL.
+- _nack_: Short for negative acknowledgement. Used to determine (during a view change) which entries can be dropped from the log. See [Protocol Aware Recovery](https://www.usenix.org/system/files/conference/fast18/fast18-alagappan.pdf).
+- _op_: Short for op-number. An op is assigned to each request that is submitted by the user before being stored in the log. An op is a monotonically increasing integer identifying each message to be handled by consensus. When messages with the same op in different views conflict, view change picks one version to commit. Each user batch (which may contain many batch entries) corresponds to one op. Each op is identified (once inside the VSR log) by a _header_.
+- _state sync_: The process of syncing not WAL entries but long-term storage (for us, the _grid_). If the WAL has been truncated and a lagging replica wants to catch up, syncing the WAL is not sufficient it must also sync what is in the long-term storage. What has been _checkpoint_-ed.
+- _superblock_: All local state for the replica that cannot be replicated remotely. Loss is protected against by storing `config.superblock_copies` copies of the superblock.
+- _zone_: The TigerBeetle data file is made up of zones. The superblock is one zone.
+- _view_: A replica is _primary_ for one view. Views are monotonically increasing integers that are incremented each time a new primary is selected.
+- _WAL_: Write-ahead log. It is implemented as two ring buffers where entries are only ever overwritten after they have been checkpointed.
+
 # Protocols
 
 ### Commands
@@ -209,7 +223,7 @@ See also:
 
 TODO (Unimplemented)
 
-## Protocol: Repair: State Transfer
+## Protocol: Repair: State Sync
 
 TODO (Unimplemented)
 
@@ -235,3 +249,8 @@ See also:
 
   - `constants.quorum_replication_max` for configuration.
   - [Flexible Paxos](https://fpaxos.github.io/)
+
+## Further reading
+
+* [Viewstamped Replication Revisited](https://pmg.csail.mit.edu/papers/vr-revisited.pdf)
+* [Protocol Aware Recovery](https://www.usenix.org/system/files/conference/fast18/fast18-alagappan.pdf)
--- a/scripts/.cspell.json
+++ b/scripts/.cspell.json
@@ -124,6 +124,9 @@
 	"ilpp",
 	"tprt",
 	"preimage",
-	"threadpools"
+	"threadpools",
+	"snapshotting",
+	"checkpointing",
+	"checkpointed"
     ]
 }


src/lsm/README.md
```diff
@@ -2,13 +2,13 @@
...
-- _groove_: A collection of LSM trees, storing objects and their indices.
+- _groove_: a collection of LSM trees for an object: one tree for the object and one tree for each field on the object (so that every field on the object is indexed). If an object has two fields, the groove will produce three trees.
 - _immutable table_: in-memory table; one per tree. Used to periodically flush the mutable table to
   disk.
-- _level_: Between `0` and `lsm_levels - 1` (usually `lsm_levels = 7`).
+- _level_: a collection of immutable tables, numbering between `0` and `config.lsm_levels - 1` (usually `config.lsm_levels = 7`).
 - _forest_: a collection of grooves.
 - _manifest_: index of table and level metadata; one per tree.
-- _mutable table_: in-memory table; one per tree. All tree updates are applied only to this table.
+- _mutable table_: in-memory table; one per tree. All tree updates (i.e. `Tree.put`) directly modify just this table.
```
sentientwaffle: Nit, since other methods can update the table too.

```suggestion
- _mutable table_: in-memory table; one per tree. All tree updates (e.g. `Tree.put`) directly modify just this table.
```

src/lsm/README.md
```diff
@@ -2,13 +2,13 @@
 
 - _bar_/_measure_: `lsm_batch_multiple` beats; unit of incremental compaction.
 - _beat_: `op % lsm_batch_multiple`; Single step of an incremental compaction.
-- _groove_: A collection of LSM trees, storing objects and their indices.
+- _groove_: a collection of LSM trees for an object: one tree for the object and one tree for each field on the object (so that every field on the object is indexed). If an object has two fields, the groove will produce three trees.
```
sentientwaffle: I would revert this change. Not all fields are necessarily indexed, and there can be indexes that don't directly correspond to fields. (Though we do not use this latter feature yet).

src/lsm/README.md
```diff
@@ -2,13 +2,13 @@
...
 - _immutable table_: in-memory table; one per tree. Used to periodically flush the mutable table to
   disk.
-- _level_: Between `0` and `lsm_levels - 1` (usually `lsm_levels = 7`).
+- _level_: a collection of immutable tables, numbering between `0` and `config.lsm_levels - 1` (usually `config.lsm_levels = 7`).
```
sentientwaffle:
```suggestion
- _level_: a collection of on-disk tables, numbering between `0` and `config.lsm_levels - 1` (usually `config.lsm_levels = 7`).
```
To differentiate from the "immutable table", which is not part of a level.

src/lsm/README.md
eatonphil: Done! Thanks

eatonphil:
There are a few suggestions I wanted to make to the VSR doc while I'm at it.

src/vsr/README.md
```diff
@@ -1,3 +1,13 @@
...
+- _journal_: Our write-ahead log (WAL). It is implemented as a ring buffer but entries are never dropped/overwritten. They are removed during compaction/snapshotting.
+- _nack_: Short for negative acknowledgement. Used to determine (through a separate consensus) if entries can be dropped from a log. See [Protocol Aware Recovery](https://www.usenix.org/system/files/conference/fast18/fast18-alagappan.pdf).
+- _op_: Short for op-number. Tied to an entry or command, submitted by the user, that is stored in the log by consensus. Identified once inside a log by a _header_. Unique, monotonically increasing integer identifying each message to be handled by VSR consensus. Each user batch (which may contain many messages) corresponds to one op.
+- _superblock_: All local state for the replica that cannot be replicated remotely. Loss is protected against by storing `config.superblock_copies` copies of the superblock.
+- _zone_: The TigerBeetle data file is made up of zones. The superblock is one zone.
+- _view_: Short for view-number. Each time a leader is elected it is elected for one view. Views are monotonically increasing integers that are incremented for each leader election.
```
sentientwaffle: > _journal_: Our write-ahead log (WAL).
- I like to distinguish between "journal" and "WAL". I use the former to refer to our in-memory structure that manages the WAL. And the latter to refer to the actual durable WAL that is on-disk.

> It is implemented as a ring buffer but entries are never dropped/overwritten. They are removed during compaction/snapshotting.

They are overwritten, but only after they have been checkpointed.

> Unique, monotonically increasing integer

I think this is a little misleading — exactly one message per op is committed, but there can be multiple messages with the same op (in different views). A view change picks one version to try to commit.

Also, "submitted by the user" makes it sounds like the user/client picks the op number, which is not the case — they are assigned by the primary.

>  - _view_: Short for view-number.

Depending on the context, "view" isn't short for anything. Though we do sometimes refer to the view number as a view too :sweat_smile: 


Lastly, I think we should avoid using "leader election" terminology in the technical documentation.

src/vsr/README.md
```diff
@@ -1,3 +1,15 @@
+# Glossary
+
+- _checkpoint_: The process of moving items from the WAL to permanent storage. (Sidenote: in consensus literate this is often called snapshotting. But we use that term to mean something else.)
```
sentientwaffle: (Apologies for the nitpicking!)
> - _checkpoint_: The process of moving items from the WAL to permanent storage. 

This is tricky. I think this is a little misleading as it is right now, but I'm not sure how to explain it succinctly enough for a glossary.
For one, the "checkpoint" itself doesn't actually remove/overwrite/move anything in the WAL — it doesn't touch the WAL at all. The "moving" (really copying) is handled by compaction, and then after the checkpoint, we can safely overwrite the WAL.
Also, s/"permanent storage"/grid, since we can re-write that data (via compaction) later.

s/"literate"/"literature"

eatonphil: Good to be nit-picky!

What about

> _checkpoint_: The process whereby once entries in the WAL are compacted into the LSM tree, those entries are removed from the WAL. ( ... )

sentientwaffle: We don't proactively "remove" anything from the WAL. Checkpointing just means that we are free to overwrite them (which we do as we wrap around the WAL).

How about:

_checkpoint_: Ensure that all updates from the past wrap of the WAL are durable in the grid, then advance the replica's recovery point by updating the superblock. After a checkpoint, the checkpointed WAL entries are safe to be overwritten by the next wrap.

eatonphil: Cool that's fine! I realize `grid` makes more sense than saying `LSM tree` in here. I want to clarify grid though (I'm adding an entry). Grid is "the collection of all lsm trees and metadata about them", is that right?

src/vsr/README.md
eatonphil: Ok! Gave a shot adjusting for these suggestions.

src/vsr/README.md
sentientwaffle: I would say that the grid is the section of the data file where the lsm trees and metadata reside. :+1:

eatonphil: Awesome, added

src/vsr/README.md
```diff
@@ -1,3 +1,17 @@
...
+- _state transfer_: The process of syncing not WAL entries but long-term storage (for us, the _grid_). If the WAL has been truncated and a lagging replica wants to catch up, syncing the WAL is not sufficient it must also sync what is in the long-term storage. What has been _checkpoint_-ed.
+- _superblock_: All local state for the replica that cannot be replicated remotely. Loss is protected against by storing `config.superblock_copies` copies of the superblock.
+- _zone_: The TigerBeetle data file is made up of zones. The superblock is one zone.
+- _view_: A replica is _primary_ for one view. Views are monotonically increasing integers that are incremented each time a new primary is selected.
+- _wal_: Write-ahead log. It is implemented as two ring buffers where entries are only ever overwritten after they have been checkpointed.
```
sentientwaffle: Capitalize "WAL". We sometimes use lower-case in code just because of caps conventions.

src/vsr/README.md
```diff
@@ -1,3 +1,17 @@
+# Glossary
+
+- _checkpoint_: Ensure that all updates from the past wrap of the WAL are durable in the _grid_, then advance the replica's recovery point by updating the superblock. After a checkpoint, the checkpointed WAL entries are safe to be overwritten by the next wrap. (Sidenote: in consensus literature this is sometimes called snapshotting. But we use that term to mean something else.)
+- _grid_: The zone on disk where LSM trees and metadata for them reside.
+- _header_: Identifier for an entry in the VSR log. Passed around instead of the entry when the full entry is not needed (such as view change).
```
sentientwaffle: _header_ can be this specific (log entry) in context, but it is more general (all messages/blocks have a header).

src/vsr/README.md
```diff
@@ -1,3 +1,17 @@
...
+- _journal_: The in-memory data structure that manages the WAL.
+- _nack_: Short for negative acknowledgement. Used to determine (through a separate consensus) if entries can be dropped from a log. See [Protocol Aware Recovery](https://www.usenix.org/system/files/conference/fast18/fast18-alagappan.pdf).
```
sentientwaffle:
```suggestion
- _nack_: Short for negative acknowledgement. Used to determine (during a view change) which entries can be dropped from the log. See [Protocol Aware Recovery](https://www.usenix.org/system/files/conference/fast18/fast18-alagappan.pdf).
```

src/vsr/README.md
```diff
@@ -1,3 +1,17 @@
...
+- _op_: Short for op-number. An op is assigned to each entry/command that is submitted by the user before being stored in the log. An op is a monotonically increasing integer identifying each message to be handled by consensus. When messages with the same op in different views conflict, view change picks one version to commit. Each user batch (which may contain many batch entries) corresponds to one op. Each op is identified (once inside the VSR log) by a _header_.
+- _state transfer_: The process of syncing not WAL entries but long-term storage (for us, the _grid_). If the WAL has been truncated and a lagging replica wants to catch up, syncing the WAL is not sufficient it must also sync what is in the long-term storage. What has been _checkpoint_-ed.
```
sentientwaffle: We are moving to the term "state sync" instead of "state transfer" to avoid overloading "transfer".

src/vsr/README.md
```diff
@@ -1,3 +1,17 @@
...
+- _grid_: The zone on disk where LSM trees and metadata for them reside.
+- _header_: Identifier for an entry in the VSR log. Passed around instead of the entry when the full entry is not needed (such as view change).
+- _journal_: The in-memory data structure that manages the WAL.
+- _nack_: Short for negative acknowledgement. Used to determine (through a separate consensus) if entries can be dropped from a log. See [Protocol Aware Recovery](https://www.usenix.org/system/files/conference/fast18/fast18-alagappan.pdf).
+- _op_: Short for op-number. An op is assigned to each entry/command that is submitted by the user before being stored in the log. An op is a monotonically increasing integer identifying each message to be handled by consensus. When messages with the same op in different views conflict, view change picks one version to commit. Each user batch (which may contain many batch entries) corresponds to one op. Each op is identified (once inside the VSR log) by a _header_.
```
sentientwaffle:
```suggestion
- _op_: Short for op-number. An op is assigned to each request that is submitted by the user before being stored in the log. An op is a monotonically increasing integer identifying each message to be handled by consensus. When messages with the same op in different views conflict, view change picks one version to commit. Each user batch (which may contain many batch entries) corresponds to one op. Each op is identified (once inside the VSR log) by a _header_.
```

"request" matches https://docs.tigerbeetle.com/design/client-requests

src/vsr/README.md
eatonphil: How about

> Identifier for many kinds of messages, including each entry in the VSR log. Passed around instead of the entry when the full entry is not needed (such as view change).

src/vsr/README.md
eatonphil: Should I change it in the header line 226 "Protocol: Repair: State Transfer" too? Or just leave that one for now.

sentientwaffle: Yeah, feel free to update that

src/vsr/README.md
sentientwaffle: Yeah, that works

src/vsr/README.md
eatonphil: Cool, thanks

sentientwaffle:
Squash? Then this LGTM

eatonphil:
Done, thanks!

sentientwaffle (approved):

PR #634, opened 2023-04-14
VSR: Incremental client table

sentientwaffle:
NOTE: This PR does not implement repair for the `client_replies` zone.
(Accordingly, the simulator does not inject storage faults into the `client_replies` zone yet.)
(That will be my next PR.)

Previously the client table was a superblock trailer:
- We keep four copies of it per replica.
- We rewrite (four copies of) the whole thing every checkpoint (i.e. WAL wrap).
- We keep it all in memory — one message per client (for each of `clients_max`).

This has some issues:
- Redundant writes, causing a latency spike at checkpoint.
- High RAM overhead per-client.

To address this:

- Split the client table out of the superblock — move it to its own zone.
- (Only keep 1 copy of the client table per replica.)
- The client table superblock trailer just stores reply metadata — slot index (implicitly), session, and reply header.
- Update the client table zone incrementally, in-place on disk.
- Drop the in-memory client table, just read from the zone as needed.
- Instead of replicating the client table locally, repair it via a new `command=request_reply` message (similar to how the WAL repairs with `request_prepare`).

As a bonus, this will help state sync be more lazy about transferring the client table.
(The client table is frequently-rewritten and infrequently-used.)

## Pre-merge checklist

Performance:

* [x] Compare `zig benchmark` on linux before and after this pr.

``` sh
    # benchmark results before
1223 batches in 545.86 s
load offered = 1000000 tx/s
load accepted = 18319 tx/s
batch latency p00 = 3 ms
batch latency p10 = 65 ms
batch latency p20 = 69 ms
batch latency p30 = 73 ms
batch latency p40 = 77 ms
batch latency p50 = 94 ms
batch latency p60 = 105 ms
batch latency p70 = 109 ms
batch latency p80 = 115 ms
batch latency p90 = 130 ms
batch latency p100 = 52104 ms
transfer latency p00 = 3 ms
transfer latency p10 = 11193 ms
transfer latency p20 = 35379 ms
transfer latency p30 = 70407 ms
transfer latency p40 = 117706 ms
transfer latency p50 = 176211 ms
transfer latency p60 = 240513 ms
transfer latency p70 = 299360 ms
transfer latency p80 = 362602 ms
transfer latency p90 = 430966 ms
transfer latency p100 = 535868 ms

    
    # benchmark results after
1223 batches in 531.52 s
load offered = 1000000 tx/s
load accepted = 18813 tx/s
batch latency p00 = 3 ms
batch latency p10 = 64 ms
batch latency p20 = 66 ms
batch latency p30 = 69 ms
batch latency p40 = 72 ms
batch latency p50 = 88 ms
batch latency p60 = 101 ms
batch latency p70 = 105 ms
batch latency p80 = 109 ms
batch latency p90 = 122 ms
batch latency p100 = 51412 ms
transfer latency p00 = 3 ms
transfer latency p10 = 11389 ms
transfer latency p20 = 35604 ms
transfer latency p30 = 71834 ms
transfer latency p40 = 119412 ms
transfer latency p50 = 178786 ms
transfer latency p60 = 240835 ms
transfer latency p70 = 299428 ms
transfer latency p80 = 362978 ms
transfer latency p90 = 422016 ms
transfer latency p100 = 521531 ms
```

### DIFF
--- a/src/vsr.zig
+++ b/src/vsr.zig
@@ -37,6 +37,7 @@ pub const Status = @import("vsr/replica.zig").Status;
 pub const Client = @import("vsr/client.zig").Client;
 pub const ClockType = @import("vsr/clock.zig").ClockType;
 pub const JournalType = @import("vsr/journal.zig").JournalType;
+pub const ClientRepliesType = @import("vsr/client_replies.zig").ClientRepliesType;
 pub const SlotRange = @import("vsr/journal.zig").SlotRange;
 pub const SuperBlockType = superblock.SuperBlockType;
 pub const VSRState = superblock.SuperBlockHeader.VSRState;
@@ -51,17 +52,20 @@ pub const Zone = enum {
     superblock,
     wal_headers,
     wal_prepares,
+    client_replies,
     grid,
 
     const size_superblock = superblock.superblock_zone_size;
     const size_wal_headers = constants.journal_size_headers;
     const size_wal_prepares = constants.journal_size_prepares;
+    const size_client_replies = constants.client_replies_size;
 
     comptime {
         for (.{
             size_superblock,
             size_wal_headers,
             size_wal_prepares,
+            size_client_replies,
         }) |zone_size| {
             assert(zone_size % constants.sector_size == 0);
         }
@@ -76,7 +80,8 @@ pub const Zone = enum {
             .superblock => 0,
             .wal_headers => size_superblock,
             .wal_prepares => size_superblock + size_wal_headers,
-            .grid => size_superblock + size_wal_headers + size_wal_prepares,
+            .client_replies => size_superblock + size_wal_headers + size_wal_prepares,
+            .grid => size_superblock + size_wal_headers + size_wal_prepares + size_client_replies,
         };
     }
 
@@ -85,6 +90,7 @@ pub const Zone = enum {
             .superblock => size_superblock,
             .wal_headers => size_wal_headers,
             .wal_prepares => size_wal_prepares,
+            .client_replies => size_client_replies,
             .grid => null,
         };
     }
--- a/src/vsr.zig
+++ b/src/vsr.zig
@@ -76,14 +76,16 @@ pub const Zone = enum {
             assert(offset_logical < zone_size);
         }
 
-        var offset_ = offset_logical;
-        // TODO(Zig): "inline for" when "unable to evaluate constant expression" is fixed.
-        for (std.enums.values(Zone)) |z| {
-            if (zone == z) return offset_;
-            offset_ += z.size().?;
-        } else {
-            unreachable;
+        return zone.start() + offset_logical;
+    }
+
+    pub fn start(zone: Zone) u64 {
+        comptime var start_offset = 0;
+        inline for (comptime std.enums.values(Zone)) |z| {
+            if (z == zone) return start_offset;
+            start_offset += comptime size(z) orelse 0;
         }
+        unreachable;
     }
 
     pub fn size(zone: Zone) ?u64 {
--- a/src/vsr.zig
+++ b/src/vsr.zig
@@ -76,13 +76,14 @@ pub const Zone = enum {
             assert(offset_logical < zone_size);
         }
 
-        return offset_logical + switch (zone) {
-            .superblock => 0,
-            .wal_headers => size_superblock,
-            .wal_prepares => size_superblock + size_wal_headers,
-            .client_replies => size_superblock + size_wal_headers + size_wal_prepares,
-            .grid => size_superblock + size_wal_headers + size_wal_prepares + size_client_replies,
-        };
+        var offset_ = offset_logical;
+        // TODO(Zig): "inline for" when "unable to evaluate constant expression" is fixed.
+        for (std.enums.values(Zone)) |z| {
+            if (zone == z) return offset_;
+            offset_ += z.size().?;
+        } else {
+            unreachable;
+        }
     }
 
     pub fn size(zone: Zone) ?u64 {
--- a/src/vsr/client_replies.zig
+++ b/src/vsr/client_replies.zig
@@ -1,0 +1,262 @@
+const std = @import("std");
+const assert = std.debug.assert;
+const mem = std.mem;
+const log = std.log.scoped(.client_replies);
+
+const stdx = @import("../stdx.zig");
+const constants = @import("../constants.zig");
+const RingBuffer = @import("../ring_buffer.zig").RingBuffer;
+const IOPS = @import("../iops.zig").IOPS;
+const vsr = @import("../vsr.zig");
+const Message = @import("../message_pool.zig").MessagePool.Message;
+const MessagePool = @import("../message_pool.zig").MessagePool;
+const Slot = @import("superblock_client_sessions.zig").ReplySlot;
+const Entry = @import("superblock_client_sessions.zig").ClientSessions.Entry;
+
+const client_replies_iops_max =
+    constants.client_replies_iops_read_max + constants.client_replies_iops_write_max;
+
+fn slot_offset(slot: Slot) usize {
+    return slot.index * constants.message_size_max;
+}
+
+pub fn ClientRepliesType(comptime Storage: type) type {
+    return struct {
+        const ClientReplies = @This();
+
+        const Read = struct {
+            client_replies: *ClientReplies,
+            completion: Storage.Read,
+            callback: fn (client_replies: *ClientReplies, reply: ?*Message) void,
+            message: *Message,
+            /// The header of the expected reply.
+            header: vsr.Header,
+        };
+
+        const Write = struct {
+            client_replies: *ClientReplies,
+            completion: Storage.Write,
+            callback: fn (client_replies: *ClientReplies, wrote: *Message) void,
+            slot: Slot,
+            message: *Message,
+        };
+
+        storage: *Storage,
+        message_pool: *MessagePool,
+        replica: u8,
+
+        reads: IOPS(Read, constants.client_replies_iops_read_max) = .{},
+        writes: IOPS(Write, constants.client_replies_iops_write_max) = .{},
+
+        /// Track which slots have a write currently in progress.
+        writing: std.StaticBitSet(constants.clients_max) =
+            std.StaticBitSet(constants.clients_max).initEmpty(),
+
+        /// Guard against multiple concurrent writes to the same slot.
+        /// Pointers are into `writes`.
+        write_queue: RingBuffer(*Write, constants.client_replies_iops_write_max, .array) = .{},
+
+        pub fn init(storage: *Storage, message_pool: *MessagePool, replica_index: u8) ClientReplies {
+            return .{
+                .storage = storage,
+                .message_pool = message_pool,
+                .replica = replica_index,
+            };
+        }
+
+        pub fn deinit(client_replies: *ClientReplies) void {
+            {
+                var it = client_replies.reads.iterate();
+                while (it.next()) |read| client_replies.message_pool.unref(read.message);
+            }
+            {
+                var it = client_replies.writes.iterate();
+                while (it.next()) |write| client_replies.message_pool.unref(write.message);
+            }
+            // Don't unref `write_queue`'s Writes — they are a subset of `writes`.
+        }
+
+        pub fn read_reply(
+            client_replies: *ClientReplies,
+            slot: Slot,
+            entry: *const Entry,
+            callback: fn (*ClientReplies, ?*Message) void,
+        ) void {
+            const client = entry.header.client;
+
+            if (client_replies.writing.isSet(slot.index)) {
+                var writes = client_replies.writes.iterate();
+                var write_latest: ?*const Write = null;
+                while (writes.next()) |write| {
+                    if (write.message.header.client == client) {
+                        if (write_latest == null or
+                            write_latest.?.message.header.request < write.message.header.request)
+                        {
+                            write_latest = write;
+                        }
+                    }
+                }
+                assert(write_latest.?.message.header.checksum == entry.header.checksum);
+
+                callback(client_replies, write_latest.?.message);
+                return;
+            }
+
+            const read = client_replies.reads.acquire() orelse {
+                log.debug("{}: read_reply: busy (client={} reply={})", .{
+                    client_replies.replica,
+                    entry.header.client,
+                    entry.header.checksum,
+                });
+
+                callback(client_replies, null);
+                return;
+            };
+
+            const message = client_replies.message_pool.get_message();
+            defer client_replies.message_pool.unref(message);
+
+            read.* = .{
+                .client_replies = client_replies,
+                .completion = undefined,
+                .message = message.ref(),
+                .callback = callback,
+                .header = entry.header,
+            };
+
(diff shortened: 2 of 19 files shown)


sentientwaffle:
MacOS CI failure appears to be a flaky test.

src/vsr/superblock_client_sessions.zig
```diff
@@ -0,0 +1,240 @@
...
+    };
+
+    pub fn encode(client_sessions: *const ClientSessions, target: []align(@alignOf(vsr.Header)) u8) u64 {
+        assert(target.len >= encode_size_max);
+
+        var size: u64 = 0;
+
+        // Write all headers:
+        var new_size = std.mem.alignForward(size, @alignOf(vsr.Header));
+        std.mem.set(u8, target[size..new_size], 0);
```
matklad, line 95: This is probably not important, but it doesn't feel maximally optimal that we zero-out and then immediately overwrite. Maybe `memset` just the gap created by alignForward?

src/vsr/superblock_client_sessions.zig
```diff
@@ -0,0 +1,240 @@
...
+    /// If the entry is from a newly-registered client, the caller is responsible for ensuring
+    /// the ClientSessions has available capacity.
+    pub fn put(client_sessions: *ClientSessions, entry: *const Entry) ReplySlot {
+        assert(entry.session != 0);
+        assert(entry.header.command == .reply);
+        const client = entry.header.client;
+
+        defer if (constants.verify) assert(client_sessions.entries_by_client.contains(client));
+
+        if (client_sessions.entries_by_client.get(client)) |entry_index| {
```
matklad: get_or_put here to avoid double hash map lookup?

src/vsr/superblock_client_sessions.zig
```diff
@@ -0,0 +1,240 @@
...
+            const existing = &client_sessions.entries[entry_index];
+            assert(existing.session == entry.session);
+            assert(existing.header.client == client);
+            assert(existing.header.commit > entry.header.commit);
+
+            existing.header = entry.header;
+            return ReplySlot{ .index = entry_index };
+        } else {
+            for (client_sessions.entries) |*e, i| {
+                if (e.session == 0) {
```
matklad: Perhaps it makes sense to do a free list here, if we are thinking about 1k clients anyways?

src/vsr/superblock_client_sessions.zig
```diff
@@ -0,0 +1,240 @@
...
+                    return ReplySlot{ .index = i };
+                }
+            } else unreachable;
+        }
+    }
+
+    pub fn remove(client_sessions: *ClientSessions, client: u128) void {
+        const entry_index = client_sessions.entries_by_client.get(client).?;
+
+        const removed = client_sessions.entries_by_client.remove(client);
```
matklad: `fetchRemove` to save one hashmap lookup and make assert superflows?

src/vsr/superblock_client_sessions.zig
```diff
@@ -0,0 +1,240 @@
...
+const mem = std.mem;
+
+const constants = @import("../constants.zig");
+const vsr = @import("../vsr.zig");
+const stdx = @import("../stdx.zig");
+
+/// There is a slot corresponding to every active client (i.e. a total of clients_max slots).
+pub const ReplySlot = struct { index: usize };
+
+pub const ClientSessions = struct {
```
matklad, line 15: Might be helpful to add a comment to say that this holds _in memory_ part of client session, and point towards client replies for the persistent part?

src/vsr/client_replies.zig
```diff
@@ -0,0 +1,342 @@
...
+                var it = client_replies.writes.iterate();
+                while (it.next()) |write| client_replies.message_pool.unref(write.message);
+            }
+            // Don't unref `write_queue`'s Writes — they are a subset of `writes`.
+        }
+
+        pub fn read_reply(
+            client_replies: *ClientReplies,
+            slot: Slot,
+            entry: *const Entry,
```
matklad: Maybe spell this as


```suggestion
            entry: *const CientSession.Entry,
```

The difference between solt&entry is not obvious without context, and context is in the other file

matklad: Maybe even call it `session: *const ...` ?

src/vsr/client_replies.zig
```diff
@@ -0,0 +1,342 @@
...
+                        if (write_latest == null or
+                            write_latest.?.message.header.request < write.message.header.request)
+                        {
+                            write_latest = write;
+                        }
+                    }
+                }
+                assert(write_latest.?.message.header.checksum == entry.header.checksum);
+
+                callback(client_replies, write_latest.?.message);
```
matklad: shouldn't we schedule this to run on the next tick? I was under the impression that we, by default, try to make async APIs always async, to guarantee bounded stack depth.

src/vsr/client_replies.zig
```diff
@@ -0,0 +1,342 @@
...
+            slot: Slot,
+            entry: *const Entry,
+            callback: fn (*ClientReplies, ?*Message) void,
+        ) void {
+            const client = entry.header.client;
+
+            if (client_replies.writing.isSet(slot.index)) {
+                var writes = client_replies.writes.iterate();
+                var write_latest: ?*const Write = null;
+                while (writes.next()) |write| {
```
matklad, line 99: Hm, shouldn't we have a symmetric in loop `write_reply` (or at least an assert)?

What happens if we schedule fall through here, schedule a read, and then try to wirte to the same slot before the read is completed?

src/vsr/superblock_client_sessions.zig
sentientwaffle: > Maybe memset just the gap created by alignForward?

:confused: Isn't this what we are doing already? 

```
        new_size = std.mem.alignForward(size, @alignOf(u64));
        std.mem.set(u8, target[size..new_size], 0);
        size = new_size;
```

src/vsr/client_replies.zig
```diff
@@ -0,0 +1,342 @@
...
+            );
+        }
+
+        fn read_reply_callback(completion: *Storage.Read) void {
+            const read = @fieldParentPtr(ClientReplies.Read, "completion", completion);
+            const client_replies = read.client_replies;
+
+            defer {
+                client_replies.message_pool.unref(read.message);
+                client_replies.reads.release(read);
```
matklad: Might as well release a read immediately, to be symmetric with write?

src/vsr/client_replies.zig
```diff
@@ -0,0 +1,342 @@
...
+
+            const write = client_replies.writes.acquire().?;
+            write.* = .{
+                .client_replies = client_replies,
+                .completion = undefined,
+                .message = message.ref(),
+                .slot = slot,
+            };
+
+            // TODO Optimization — if there is already a write for this slot queued (but not started), replace it.
```
matklad: Interestingly, there's more to this than just avoiding redundant writes. If I am reading the code correctly, then, if _some_ client gets two writes in the write_queue, then all _other_ clients would be blocked, because we won't go past this laggy write in the 


```
            while (client_replies.write_queue.head()) |write| {
                if (client_replies.writing.isSet(write.slot.index)) break;
```

loop. 

This actually seems pretty bad, as it creates two equilibrium states. Normally, we can serve `client_replies_iops_write_max` clients in concurrently. 

Now, if one of those clients bursts and causes two writes in the queue, all other clients stall. So it becomes likely that some other cilent queues two writes. And that can lock us into usining only one write iop, despite the fact that we have a bunch of enqueued writes, and a bunch of iops available.

src/vsr/superblock_client_sessions.zig
matklad: Yup, I can't read!

src/vsr.zig
```diff
@@ -76,7 +80,8 @@ pub const Zone = enum {
             .superblock => 0,
             .wal_headers => size_superblock,
             .wal_prepares => size_superblock + size_wal_headers,
-            .grid => size_superblock + size_wal_headers + size_wal_prepares,
+            .client_replies => size_superblock + size_wal_headers + size_wal_prepares,
+            .grid => size_superblock + size_wal_headers + size_wal_prepares + size_client_replies,
```
matklad: One more here, and we probably should switch to inline-for or inline-switch

src/vsr/superblock_client_sessions.zig
sentientwaffle: I'll change it to a bitset instead

src/vsr/client_replies.zig
sentientwaffle: Yeah, that is definitely the best practice. But we need a `NextTick` context to do it, and there isn't a way to bound the number of concurrent `read_reply`s with the current API.

I think instead I'll just split `read_reply` into `read_reply_sync` and `read_reply`, and the caller is responsible for checking the former before resorting to the latter. That way we never need `next_tick` here. (There is only one call site (or 2 once repair is implemented) anyway.)

src/vsr/superblock_client_sessions.zig
```diff
@@ -0,0 +1,240 @@
...
+    }
+
+    pub fn get_slot(client_sessions: *ClientSessions, client: u128) ?ReplySlot {
+        const index = client_sessions.entries_by_client.get(client) orelse return null;
+        return ReplySlot{ .index = index };
+    }
+
+    /// If the entry is from a newly-registered client, the caller is responsible for ensuring
+    /// the ClientSessions has available capacity.
+    pub fn put(client_sessions: *ClientSessions, entry: *const Entry) ReplySlot {
```
matklad: microoptimization, but maybe `put(client_sessions: *ClientSessions, session: u64, header: *const Header)`? I feel like groupping a header into an addressable Entry might force an extra memcpy. Or at least compiler would have to do an extra sroa to undo that.

src/vsr/replica.zig
```diff
@@ -2882,41 +2913,40 @@ pub fn ReplicaType(
...
-                        if (entry.reply.header.commit < evictee_reply.header.commit) {
-                            evictee = entry.reply;
+                        if (entry.header.commit < evictee_reply.commit) {
+                            evictee = &entry.header;
                         }
                     } else {
-                        evictee = entry.reply;
+                        evictee = &entry.header;
                     }
                 }
```
matklad: Maybe move the whole loop to the client_sessions? It is indeed pretty crucial that the iteration order is consistent across replicas, but that's not obvious just from the session's code

src/vsr/client_replies.zig
```diff
@@ -0,0 +1,342 @@
...
+                assert(client_replies.writes.available() > 0);
+
+                client_replies.ready_callback = null;
+                callback(client_replies);
+            } else {
+                // Another write finished before the next_tick triggered.
+            }
+        }
+
+        pub fn write_reply(client_replies: *ClientReplies, slot: Slot, message: *Message) void {
```
matklad, line 256:
```zig
            assert(client_replies.ready_callback == null)
```

it woul be incorrect to write a reply if you are waiting for a slot

matklad (commented):
looks good!

https://github.com/tigerbeetledb/tigerbeetle/pull/634/files#r1168805259 I think is a genuine correctness issue (race), but the rest are nits. 

I think this infra can serve as a grate exampel of how we do things in tigerbeetle. It is very self-contained, and displays in-memory data-structure, on disk data structure, and budgeting of IO!

src/vsr/client_replies.zig
sentientwaffle: :+1:  I'll implement this TODO. But if the blocking does end up as a perf issue, we can switch out the ring buffer queue for a linked list. (I just used a ring buffer for simplicity.)

src/vsr.zig
```diff
@@ -76,13 +76,14 @@ pub const Zone = enum {
...
 
-        return offset_logical + switch (zone) {
-            .superblock => 0,
-            .wal_headers => size_superblock,
-            .wal_prepares => size_superblock + size_wal_headers,
-            .client_replies => size_superblock + size_wal_headers + size_wal_prepares,
-            .grid => size_superblock + size_wal_headers + size_wal_prepares + size_client_replies,
-        };
+        var offset_ = offset_logical;
+        // TODO(Zig): "inline for" when "unable to evaluate constant expression" is fixed.
```
matklad: Got it to compile with 

```
        comptime var offset_ = 0;
        inline for (comptime std.enums.values(Zone)) |z| {
            if (zone == z) return offset_;
            offset_ += comptime z.size() orelse 0;
        } else {
            unreachable;
        }
```

`comptime` in `for` seems like compiler bug, but `orelse 0` seems essential, at least according to my mental model of comptime

sentientwaffle: Can you elaborate on the `orelse 0`? `size()` is `null` only for the grid, which is the last zone, so we should never add that last size, hence `.?`. (It does need `orelse 0` to compile, but I interpreted that as another compiler bug).

matklad (dismissed):

src/vsr.zig
matklad: The expanded code looks like this:

```zig
const offset1 = 0
if (zone == .superblock) return offset1;
const offset2 = offset1 + .superblock.size().?

if (zone == .wal_headers) return offse2_;
const offset3 = offset2 + .size_wal_headers().?

...

if (zone == .grid) return offse2_;
const offsetN = offsetN-1 + .grid.size_wal_headers().?
unreachable
```

even though the code is unreachable at runtime, at comptime zig still has to evaluate

```
const offsetN = offsetN-1 + .grid.size_wal_headers().?
```

and that fails

sentientwaffle: Ah, that makes sense, thanks!

matklad: actually ,wait a sec, I think this is not _exactly_ what we want here

matklad:
```zig
        comptime var offset_zone = 0;
        inline for (comptime std.enums.values(Zone)) |z| {
            if (zone == z) return offset_zone + offset_logical;
            offset_zone += comptime z.size() orelse 0;
        } else {
            unreachable;
        }
```

this is what we want (to force zone start offset to be evaluated at compile time).

matklad: Or maybe even

```zig
    pub fn offset(zone: Zone, offset_logical: u64) u64 {
        if (zone.size()) |zone_size| {
            assert(offset_logical < zone_size);
        }

        return zone.start() + offset_logical;
    }

    pub fn start(zone: Zone) u64 {
        comptime var start_offset = 0;
        inline for (comptime std.enums.values(Zone)) |z| {
            if (z == zone) return start_offset;
            start_offset += comptime size(z) orelse 0;
        }
        unreachable;
    }

    pub fn size(zone: Zone) ?u64 {
        return switch (zone) {
            .superblock => size_superblock,
            .wal_headers => size_wal_headers,
            .wal_prepares => size_wal_prepares,
            .client_replies => size_client_replies,
            .grid => null,
        };
    }
```

if we go for small functions. Sorry for drawing attention to irrelevant :sweat_smile:

sentientwaffle: Ooh, great catch!

sentientwaffle: Oh I just barely missed that last comment, I'll update it again

matklad (dismissed):

matklad (approved):

PR #635, opened 2023-04-14
Add READMEs for sample projects, run client README validation on macos and windows in addition to Linux

eatonphil:
This PR:

* Drops the Docker requirement from the docs_generate script's validation which makes it faster
* Enables the docs_generate compile step on Mac . This wouldn't be that crazy except for that this includes building the client which is a first outside of Linux since we elsewhere cross compile.
* Adds READMEs for sample projects so that you can understand what they do and **how to run them**.

Unrelated


.github/workflows/macos.yml
```diff
@@ -40,19 +40,32 @@ jobs:
...
+        with:
+          java-version: '11'
+          distribution: 'temurin'
+          server-id: github
+          settings-path: ${{ github.workspace }}
+
+      # Although Maven is supposed to be installed in the
+      # actions/setup-java@v3 step, it wasn't available in at least the
+      # self-hosted mac runner. So we grab it manually.
+      - if: matrix.language == 'java'
```
batiati: Can we save some CI time here by adding `if matrix.target == 'self-hosted'` ?

.github/workflows/windows.yml
```diff
@@ -15,4 +15,24 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - run: .\scripts\install_zig.bat
-      - run: zig\zig.exe build c_sample -Drelease-safe
\ No newline at end of file
+      - run: zig\zig.exe build c_sample -Drelease-safe
```
batiati: Can we use `.\scripts\build.bat` instead?

src/clients/README.md
```diff
@@ -19,43 +19,56 @@ Existing `Docs` struct implementations are in:
 
 ### Run
 
-To run the generator:
+Go to the repo root.
+
+If you don't already have the TigerBeetle version of `zig` run:
+
+```console
+$ ./scripts/install_zig.[sh|bat]
```
batiati, line 27: Would it be more precise if instead of using `[.sh|.bat]` we explicitly had two different snippets for Linux/macOS and Windows? 

Rationale:

- The user can copy and paste.
- Windows accepts `/`, but still, it should be `\`.
- PowerShell accepts `$(pwd)`, but if the user is running in cmd, it doesn't work.

src/clients/node/samples/basic/README.md
```diff
@@ -0,0 +1,66 @@
+This file is generated by
+[/src/clients/docs_generate.zig](/src/clients/docs_generate.zig).
+
+# Basic Node.js Sample
+
+Code for this sample is in [./main.js](./main.js).
+
+## Prerequisites
+
+Linux >= 5.6 is the only production environment we
```
batiati, line 10: For node, we could add a note that "Windows is not supported at the moment".

src/clients/go/docs.zig
```diff
@@ -296,31 +295,23 @@ pub const GoDocs = Docs{
...
     \\./scripts/build.sh go_client -Drelease-safe
     \\cd src/clients/go
-    \\[ "$TEST" = "true" ] && ./zgo.sh test || echo "Skipping client unit tests"
+    \\if [ "$TEST" = "true" ]; then ./zgo.sh test; else echo "Skipping client unit tests"; fi
     ,
 
     .current_commit_pre_install_commands = "",
     .current_commit_post_install_commands = 
-    \\printf 'module tbtest\nreplace github.com/tigerbeetledb/tigerbeetle-go => '$(pwd)'/tigerbeetle/src/clients/go\ngo 1.18' > go.mod
+    \\printf 'module tbtest\nreplace github.com/tigerbeetledb/tigerbeetle-go => '$TB_ROOT'/src/clients/go\ngo 1.18' > go.mod
```
batiati: Running `.\scripts\build.bat client_docs -- --language go` on Windows, I got the error `printf is not recognized as an internal command`

src/clients/java/docs.zig
```diff
@@ -92,21 +92,16 @@ pub const JavaDocs = Docs{
     .install_prereqs = "apk add -U maven openjdk11",
 
     .current_commit_pre_install_commands = 
-    \\mvn deploy:deploy-file -Durl=file://$(pwd) -Dfile=$(find . -name '*-SNAPSHOT.jar') -DgroupId=com.tigerbeetle -DartifactId=tigerbeetle-java -Dpackaging=jar -Dversion=0.0.1-3431
+    \\mvn deploy:deploy-file -Durl=file://$SAMPLE_ROOT -Dfile=$(find . -name '*-SNAPSHOT.jar') -DgroupId=com.tigerbeetle -DartifactId=tigerbeetle-java -Dpackaging=jar -Dversion=0.0.1-3431
     \\mkdir -p $HOME/.m2/
-    \\echo '<settings xmlns="http://maven.apache.org/SETTINGS/1.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/SETTINGS/1.0.0 http://maven.apache.org/xsd/settings-1.0.0.xsd"><localRepository>'$(pwd)'</localRepository></settings>' > $HOME/.m2/settings.xml
+    \\echo '<settings xmlns="http://maven.apache.org/SETTINGS/1.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/SETTINGS/1.0.0 http://maven.apache.org/xsd/settings-1.0.0.xsd"><localRepository>'$SAMPLE_ROOT'</localRepository></settings>' > $HOME/.m2/settings.xml
```
batiati: It's ok for a CI runner, but overwriting the `$HOME/.m2/settings.xml` file in a developer machine seems odd. We should instead specify it on maven command line `mvn --global-settings PATH_TO_ALTERNATIVE_SETTINGS_XML`.

.github/workflows/macos.yml
eatonphil: I don't actually understand how to target the self-hosted host because we use an array to describe it.

.github/workflows/windows.yml
eatonphil: I didn't actually make a change to this line. In any future PR we could make that change yes but it's unrelated to my changes.

src/clients/README.md
eatonphil: This is internal developer docs not docs users should read. I agree it would be nice but I wanted more to emphasize everywhere the symmetry and I did try a version splitting out Windows and Linux/macOS but then it got super verbose. This seemed like a nice in between.

src/clients/node/samples/basic/README.md
eatonphil: Yeah good point. Done

.github/workflows/windows.yml
eatonphil: Ok I made this change for this PR but it doesn't actually matter since scripts\build.bat is a dumb wrapper. There's no point in making this change everywhere IMO.

src/clients/go/docs.zig
eatonphil: Hmm yeah that's a great point. Why on earth was this passing tests.

src/clients/java/docs.zig
eatonphil: Yeah I agree this is not good. I even use `mvn -s local-settings.xml` elsewhere. I will remove this but not necessarily in this PR.

eatonphil: I ended up making this change in this pr.

src/clients/go/docs.zig
eatonphil: This code is no longer used.

batiati (dismissed):

eatonphil:
I've had so many issues with getting Java and Go to work on Windows I'm dropping that part for now.

It's basically back to where you reviewed it @batiati.

eatonphil:
Nevermind! Windows tests for Go and Java are now committed and working.

.github/workflows/clients-go.yml
```diff
@@ -20,45 +20,9 @@ jobs:
           name: go-binaries-${{ github.sha }}
           path: src/clients/go/pkg/native
 
-  tests:
```
eatonphil, line 23: Tests are now run as part of the docs validation step.

Proven that errors in tests fail the build here: https://github.com/tigerbeetledb/tigerbeetle/pull/754

batiati (approved):

PR #636, opened 2023-04-14
aof: Handle op=1 being replayed in the VOPR

cb22:
Performance:

### DIFF
--- a/src/testing/aof.zig
+++ b/src/testing/aof.zig
@@ -22,8 +22,12 @@ const InMemoryAOF = struct {
     }
 
     pub fn readAll(self: *Self, buf: []u8) !usize {
-        stdx.copy_disjoint(.inexact, u8, buf, self.backing_store[self.index .. self.index + buf.len]);
-        return buf.len;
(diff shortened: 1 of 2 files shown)


src/testing/aof.zig
```diff
@@ -73,6 +73,13 @@ pub const AOF = struct {
             const header = entry.header();
             if (self.validation_checksums.count() != 0) {
                 assert(self.validation_checksums.get(header.parent) != null);
+            } else {
+                // For op=1, put its parent in our list of seen checksums too.
+                // This handles the case where it gets replayed, so count() != 0,
+                // but we don't record op=0 so the assert above would fail.
+                // It's needed for simulator validation only (aof merge uses a
+                // different method to walk down AOF entries).
+                try self.validation_checksums.put(header.parent, {});
```
sentientwaffle, line 86: Can we assert that `header.op == 1`?

cb22: I've changed the `if` condition rather, to make it more explicit.

sentientwaffle (approved):

PR #639, opened 2023-04-18
fix(node-client): result mapping

sabineschaller:
## Changes
When upgrading to the newest version of tigerbeetle + node client in Rafiki, I realized that the typescript types `CreateAccountsError` and `CreateTransfersError` are now auto-generated and contain `{index, result}` instead of `{index, code}`. Auto generation and hence the change in the property name was introduced [here](https://github.com/tigerbeetledb/tigerbeetle/commit/6d2bce9ea05761e410f3962504e96a3a5a721822#diff-7ae23758d5faa544c4157bb7b436683f4d81444ed15f3aa241128eac098f7f1f). 

However, the mapping from zig to typescript still uses `code` instead of `result`. This PR updates it to also use `result`.

## Pre-merge checklist

Performance:

* [ ] Compare `zig build benchmark` on linux before and after this pr.
    ``` sh
    # benchmark results before
    ...
    
    # benchmark results after
    ...
    ```
OR


batiati (changes requested):
Hello @sabineschaller, nice catch, thank you for submitting a PR!

Just a minor adjustment:
The code examples presented in the README.md file were auto-generated too. Could you please change them in `/src/clients/node/docs.zig` instead of changing the markdown directly?

For example:
```diff
  .create_accounts_errors_example = 
  \\let account2 = { /* ... account values ... */ };
  \\let account3 = { /* ... account values ... */ };
  \\let account4 = { /* ... account values ... */ };
  \\accountErrors = await client.createAccounts([account2, account3, account4]);
  \\for (const error of accountErrors) {
- \\  switch (error.code) {
+ \\  switch (error.result) {
  \\    case CreateAccountError.exists:
  \\      console.error(`Batch account at ${error.index} already exists.`);
  \\	  break;
  \\    default:
- \\      console.error(`Batch account at ${error.index} failed to create: ${CreateAccountError[error.code]}.`);
+ \\      console.error(`Batch account at ${error.index} failed to create: ${CreateAccountError[error.result]}.`);
  \\  }
  \\}
```

sabineschaller:
@batiati, thank you for pointing out the auto-generation of docs. @eatonphil was so kind and took care of it for me 🙂

batiati (dismissed):
That's great! Thank you once again.

The remaining failing CI is due to it not being able to correctly handle PRs from forks. I'm putting it on the merge queue.

batiati:
I need to investigate it more, it looks like we're facing a [CI timeout](https://github.com/tigerbeetledb/tigerbeetle/actions/runs/4741864778/jobs/8420109914) during docs generation in the merge queue.

image

src/clients/node/docs.zig
```diff
@@ -307,7 +307,7 @@ pub const NodeDocs = Docs{
     \\cd src/clients/node
     \\npm install --include dev
     \\npm pack
-    \\[ "$TEST" = "true" ] && mvn test || echo "Skipping client unit tests"
+    \\[ "$TEST" = "true" ] && npm test || echo "Skipping client unit tests"
```
eatonphil: @batiati since you're already pushing to this branch, just delete this entire line if you don't mind.

I'm clarifying in #635 that this never was a path that was tested and never worked in CI.

batiati (approved):

PR #640, opened 2023-04-18
VSR: Repair client replies

sentientwaffle:
- Send `request_reply` messages to recover corrupt replies.
- Handle `request_reply` messages to help other replicas repair.
- Incoming `reply` messages repair corrupt ClientReplies entries.
- VOPR: Enable corruption of the `client_replies` zone.

## Pre-merge checklist

Performance:

### DIFF
--- a/src/vsr/client_replies.zig
+++ b/src/vsr/client_replies.zig
@@ -61,7 +61,8 @@ pub fn ClientRepliesType(comptime Storage: type) type {
         /// Track which slots have a write currently in progress.
         writing: std.StaticBitSet(constants.clients_max) =
             std.StaticBitSet(constants.clients_max).initEmpty(),
-        /// Track which slots hold a corrupt reply.
+        /// Track which slots hold a corrupt reply, or are otherwise missing the reply
+        /// that ClientSessions believes they should hold.
         faulty: std.StaticBitSet(constants.clients_max) =
             std.StaticBitSet(constants.clients_max).initEmpty(),
 
@@ -70,7 +71,10 @@ pub fn ClientRepliesType(comptime Storage: type) type {
         write_queue: RingBuffer(*Write, constants.client_replies_iops_write_max, .array) = .{},
 
         ready_next_tick: Storage.NextTick = undefined,
-        ready_callback: ?fn (*ClientReplies) void = null,
+        ready_callback: ?union(enum) {
+            next_tick: fn (*ClientReplies) void,
+            write: fn (*ClientReplies) void,
+        } = null,
 
         checkpoint_next_tick: Storage.NextTick = undefined,
         checkpoint_callback: ?fn (*ClientReplies) void = null,
@@ -239,35 +243,34 @@ pub fn ClientRepliesType(comptime Storage: type) type {
             callback: fn (client_replies: *ClientReplies) void,
         ) void {
             assert(client_replies.ready_callback == null);
-            client_replies.ready_callback = callback;
 
             if (client_replies.writes.available() > 0) {
+                client_replies.ready_callback = .{ .next_tick = callback };
                 client_replies.storage.on_next_tick(
                     ready_next_tick_callback,
                     &client_replies.ready_next_tick,
                 );
             } else {
                 // ready_callback will be called the next time a write completes.
+                client_replies.ready_callback = .{ .write = callback };
             }
         }
 
         fn ready_next_tick_callback(next_tick: *Storage.NextTick) void {
             const client_replies = @fieldParentPtr(ClientReplies, "ready_next_tick", next_tick);
+            assert(client_replies.writes.available() > 0);
 
-            if (client_replies.ready_callback) |callback| {
-                assert(client_replies.writes.available() > 0);
-
-                client_replies.ready_callback = null;
-                callback(client_replies);
-            } else {
-                // Another write finished before the next_tick triggered.
-            }
+            const callback = client_replies.ready_callback.?.next_tick;
+            client_replies.ready_callback = null;
+            callback(client_replies);
         }
 
         pub fn remove_reply(client_replies: *ClientReplies, slot: Slot) void {
             client_replies.faulty.unset(slot.index);
         }
 
+        /// The caller is responsible for ensuring that the ClientReplies is able to write
+        /// by calling `write_reply()` after `ready()` finishes.
         pub fn write_reply(client_replies: *ClientReplies, slot: Slot, message: *Message) void {
             assert(client_replies.ready_callback == null);
             assert(client_replies.writes.available() > 0);
@@ -349,10 +352,13 @@ pub fn ClientRepliesType(comptime Storage: type) type {
             client_replies.write_reply_next();
 
             if (client_replies.ready_callback) |ready_callback| {
-                assert(client_replies.writes.available() > 0);
-
-                client_replies.ready_callback = null;
-                ready_callback(client_replies);
+                switch (ready_callback) {
+                    .write => |callback| {
+                        client_replies.ready_callback = null;
+                        callback(client_replies);
+                    },
+                    .next_tick => {},
+                }
             }
 
             if (client_replies.checkpoint_callback != null and
--- a/src/vsr/client_replies.zig
+++ b/src/vsr/client_replies.zig
@@ -31,11 +31,17 @@ pub fn ClientRepliesType(comptime Storage: type) type {
         const Read = struct {
             client_replies: *ClientReplies,
             completion: Storage.Read,
-            callback: fn (client_replies: *ClientReplies, reply: ?*Message) void,
+            callback: fn (
+                client_replies: *ClientReplies,
+                reply_header: *const vsr.Header,
+                reply: ?*Message,
+                destination_replica: ?u8,
+            ) void,
             slot: Slot,
             message: *Message,
             /// The header of the expected reply.
             header: vsr.Header,
+            destination_replica: ?u8,
         };
 
         const Write = struct {
@@ -55,6 +61,9 @@ pub fn ClientRepliesType(comptime Storage: type) type {
         /// Track which slots have a write currently in progress.
         writing: std.StaticBitSet(constants.clients_max) =
             std.StaticBitSet(constants.clients_max).initEmpty(),
+        /// Track which slots hold a corrupt reply.
+        faulty: std.StaticBitSet(constants.clients_max) =
+            std.StaticBitSet(constants.clients_max).initEmpty(),
 
         /// Guard against multiple concurrent writes to the same slot.
         /// Pointers are into `writes`.
@@ -66,11 +75,15 @@ pub fn ClientRepliesType(comptime Storage: type) type {
         checkpoint_next_tick: Storage.NextTick = undefined,
         checkpoint_callback: ?fn (*ClientReplies) void = null,
 
-        pub fn init(storage: *Storage, message_pool: *MessagePool, replica_index: u8) ClientReplies {
+        pub fn init(options: struct {
+            storage: *Storage,
+            message_pool: *MessagePool,
+            replica_index: u8,
+        }) ClientReplies {
             return .{
-                .storage = storage,
-                .message_pool = message_pool,
-                .replica = replica_index,
+                .storage = options.storage,
+                .message_pool = options.message_pool,
+                .replica = options.replica_index,
             };
         }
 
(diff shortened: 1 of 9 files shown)


src/vsr/client_replies.zig
```diff
@@ -55,6 +61,9 @@ pub fn ClientRepliesType(comptime Storage: type) type {
         /// Track which slots have a write currently in progress.
         writing: std.StaticBitSet(constants.clients_max) =
             std.StaticBitSet(constants.clients_max).initEmpty(),
+        /// Track which slots hold a corrupt reply.
+        faulty: std.StaticBitSet(constants.clients_max) =
```
matklad, line 66: Is it `faulty` or just `missing`?

With state-sync, I think it would be normal for replies to be missing without necessary being faulty?

Although it is faulty today, so :shrug:

src/vsr/client_replies.zig
```diff
@@ -253,8 +264,13 @@ pub fn ClientRepliesType(comptime Storage: type) type {
             }
         }
 
+        pub fn remove_reply(client_replies: *ClientReplies, slot: Slot) void {
+            client_replies.faulty.unset(slot.index);
+        }
+
         pub fn write_reply(client_replies: *ClientReplies, slot: Slot, message: *Message) void {
             assert(client_replies.ready_callback == null);
+            assert(client_replies.writes.available() > 0);
```
matklad, line 274: Hm, why woundn't `create_client_table_entry` trip this assert?

src/vsr/replica.zig
```diff
@@ -1811,6 +1863,84 @@ pub fn ReplicaType(
...
+                    entry.header.checksum,
+                });
+                return;
+            }
+            assert(entry.header.size != @sizeOf(Header));
+            assert(entry.header.op == message.header.op);
+
+            const slot = self.client_sessions().get_slot_for_header(&entry.header).?;
+            if (self.client_replies.read_reply_sync(slot, entry) catch |err| {
+                assert(err == error.Busy);
```
matklad: Hm, now that I look at this, shoudn't we get `error.Busy` from `read_reply` rather than `sync` version?

matklad: Also, should probably log why we ignore here?

src/vsr/replica.zig
```diff
@@ -3477,9 +3622,25 @@ pub fn ReplicaType(
...
+        ) void {
             const self = @fieldParentPtr(Self, "client_replies", client_replies);
-            const reply = reply_ orelse return;
+            assert(destination_replica == null);
+
+            const reply = reply_ orelse {
+                if (self.client_sessions().get_slot_for_header(reply_header)) |slot| {
+                    self.client_replies.faulty.set(slot.index);
+                } else {
+                    // The read may have be a repair for an older op,
```
```suggestion
                    // The read may have been a repair for an older op,
```

src/vsr/replica.zig
```diff
@@ -3477,9 +3622,25 @@ pub fn ReplicaType(
...
+            reply_: ?*Message,
+            destination_replica: ?u8,
+        ) void {
             const self = @fieldParentPtr(Self, "client_replies", client_replies);
-            const reply = reply_ orelse return;
+            assert(destination_replica == null);
+
+            const reply = reply_ orelse {
+                if (self.client_sessions().get_slot_for_header(reply_header)) |slot| {
+                    self.client_replies.faulty.set(slot.index);
```
matklad, line 3661: Huh, it's a bit too subtle that we set faulty outise of `client_replies`. I am wondering, can we just unconditionally set faulty in `client_replies`? 

If we have updated the header for this client already, than we also should have a write queued, and that would clean the faulty bit?

ahhh.... that's not the case --- if the reply is header only, then we won't queue the write, and won't set the faulty bit.

src/vsr/replica.zig
```diff
@@ -3477,9 +3622,25 @@ pub fn ReplicaType(
             }
         }
 
-        fn on_request_read_reply_callback(client_replies: *ClientReplies, reply_: ?*Message) void {
+        fn on_request_read_reply_callback(
```
matklad: NB: we have 

```
fn on_request_reply_read_callback
fn on_request_read_reply_callback
```

matklad: Maybe factor out the whole on_request_ignore_resend logic into a separate function, so that we:

* don't do send directly from ignore, which feels a bit weird
* have a name for this flow?

src/vsr/client_replies.zig
sentientwaffle: I'll update the comment to
```
        /// Track which slots hold a corrupt reply, or are otherwise missing the reply
        /// that ClientSessions believes they should hold.
```
to clarify.

I think `faulty` is a decent name even with state sync, and is more descriptive in the corruption case.
After state sync we can pretend that all of the entries are corrupt and then repair them.

src/vsr/replica.zig
sentientwaffle: I'll add a log. The caller must invoke `read_reply` immediately after `read_reply_sync` fails. The reason for handling the `Busy` case in `_sync` (and more generally, the reason for splitting out `_sync` at all) is to avoid needing `NextTick` in the read path. We would need an upper bound for `NextTick` contexts, and I don't think we can do better than `messages_max`.

src/vsr/replica.zig
sentientwaffle: Yeah, ClientReplies doesn't know which header should be in the slot. If a lagging replica requests reply X, but we have a newer reply Y, then we don't want to set `faulty` just because `X` is missing. We can't just compare the op of the reply that we read, because we don't write header-only messages. I could give `ClientReplies` access to `ClientSessions`, but I kind of like the separation. :thinking:

src/vsr/client_replies.zig
sentientwaffle: I forgot to add a comment, but:

```
        /// The caller is responsible for ensuring that the ClientReplies is able to write
        /// by calling `write_reply()` after `ready()` finishes.
```

Replica inverted this order, calling `ready` _after_ instead of _before_. This was correct but confusing, so I am moving it to before.

src/vsr/replica.zig
matklad: But I think Busy can be returned from either of the two functions? 

```diff
diff --git a/src/vsr/client_replies.zig b/src/vsr/client_replies.zig
index aac53c6e..fd82b59f 100644
--- a/src/vsr/client_replies.zig
+++ b/src/vsr/client_replies.zig
@@ -107,7 +107,7 @@ pub fn ClientRepliesType(comptime Storage: type) type {
             client_replies: *ClientReplies,
             slot: Slot,
             session: *const ClientSessions.Entry,
-        ) error{Busy}!?*Message {
+        ) ?*Message {
             const client = session.header.client;
 
             if (client_replies.writing.isSet(slot.index)) {
@@ -127,16 +127,6 @@ pub fn ClientRepliesType(comptime Storage: type) type {
                 return write_latest.?.message;
             }
 
-            if (client_replies.reads.available() == 0) {
-                log.debug("{}: read_reply: busy (client={} reply={})", .{
-                    client_replies.replica,
-                    session.header.client,
-                    session.header.checksum,
-                });
-
-                return error.Busy;
-            }
-
             return null;
         }
 
@@ -148,8 +138,18 @@ pub fn ClientRepliesType(comptime Storage: type) type {
             session: *const ClientSessions.Entry,
             callback: fn (*ClientReplies, *const vsr.Header, ?*Message, ?u8) void,
             destination_replica: ?u8,
-        ) void {
-            assert(client_replies.read_reply_sync(slot, session) catch unreachable == null);
+        ) error{Busy}!void {
+            assert(client_replies.read_reply_sync(slot, session) == null);
+
+            if (client_replies.reads.available() == 0) {
+                log.debug("{}: read_reply: busy (client={} reply={})", .{
+                    client_replies.replica,
+                    session.header.client,
+                    session.header.checksum,
+                });
+
+                return error.Busy;
+            }
 
             const read = client_replies.reads.acquire().?;
 
diff --git a/src/vsr/replica.zig b/src/vsr/replica.zig
index 079bf84e..99b6f6b9 100644
--- a/src/vsr/replica.zig
+++ b/src/vsr/replica.zig
@@ -1889,14 +1889,7 @@ pub fn ReplicaType(
             assert(entry.header.op == message.header.op);
 
             const slot = self.client_sessions().get_slot_for_header(&entry.header).?;
-            if (self.client_replies.read_reply_sync(slot, entry) catch |err| {
-                assert(err == error.Busy);
-
-                log.debug("{}: on_request_reply: ignoring, client_replies busy", .{
-                    self.replica,
-                });
-                return;
-            }) |reply| {
+            if (self.client_replies.read_reply_sync(slot, entry)) |reply| {
                 on_request_reply_read_callback(
                     &self.client_replies,
                     &entry.header,
@@ -1909,7 +1902,14 @@ pub fn ReplicaType(
                     entry,
                     on_request_reply_read_callback,
                     message.header.replica,
-                );
+                ) catch |err| {
+                    assert(err == error.Busy);
+
+                    log.debug("{}: on_request_reply: ignoring, client_replies busy", .{
+                        self.replica,
+                    });
+                    return;
+                };
             }
         }
 
@@ -3625,14 +3625,7 @@ pub fn ReplicaType(
             }
 
             const slot = self.client_sessions().get_slot_for_client(message.header.client).?;
-            if (self.client_replies.read_reply_sync(slot, entry) catch |err| {
-                assert(err == error.Busy);
-
-                log.debug("{}: on_request: ignoring (client_replies busy)", .{
-                    self.replica,
-                });
-                return;
-            }) |reply| {
+            if (self.client_replies.read_reply_sync(slot, entry)) |reply| {
                 on_request_repeat_reply_callback(
                     &self.client_replies,
                     &entry.header,
@@ -3645,7 +3638,14 @@ pub fn ReplicaType(
                     entry,
                     on_request_repeat_reply_callback,
                     null,
-                );
+                ) catch |err| {
+                    assert(err == error.Busy);
+
+                    log.debug("{}: on_request: ignoring (client_replies busy)", .{
+                        self.replica,
+                    });
+                    return;
+                };
             }
         }
 
```

This version looks a bit more natural to me, as we get an error when we try to schedule async work, and there isn't any resources for that. 

I guess the drawback here is that the callback is now called either zero or once, and that could be confusing?

sentientwaffle: The zero-or-once-callback was the reason I was avoiding it, but this is indeed cleaner — we can even remove the `return` from within the catch. I'll update it.

src/vsr/client_replies.zig
matklad: Ah, I guess the problem is that this maybe is wrong then?

https://github.com/tigerbeetledb/tigerbeetle/blob/73ddb9b9acc8c5db76ab729393f87626b0e1a842/src/vsr/replica.zig#L2800

sentientwaffle: Ah, that can be
```
            assert(self.client_replies.writes.available() > 0);
```
now

src/vsr/client_replies.zig
```diff
@@ -176,17 +194,23 @@ pub fn ClientRepliesType(comptime Storage: type) type {
                 client_replies.write_reply_next();
             }
 
-            if (!message.header.valid_checksum()) {
+            if (!message.header.valid_checksum() or
+                !message.header.valid_checksum_body(message.body()))
+            {
                 log.warn("{}: read_reply: corrupt header (client={} reply={})", .{
```
matklad:
```suggestion
                log.warn("{}: read_reply: corrupt reply (client={} reply={})", .{
```

matklad: Well, not  sure what's the best phrasing,but its not only the header that can be an issue

src/vsr/client_replies.zig
```diff
@@ -228,33 +241,37 @@ pub fn ClientRepliesType(comptime Storage: type) type {
...
+            client_replies.ready_callback = null;
+            callback(client_replies);
+        }
 
-                client_replies.ready_callback = null;
-                callback(client_replies);
-            } else {
-                // Another write finished before the next_tick triggered.
-            }
+        pub fn remove_reply(client_replies: *ClientReplies, slot: Slot) void {
```
matklad, line 266: Hm, when we call this from `replica`, what guarantees that we don't have any in-flight ops for the Slot? We could assert that here?

sentientwaffle: I think it doesn't actually matter if we do have in-flight writes to this op.
Writes will queue up anyway, and `faulty` is only set (by the replica) after checking `ClientSessions`, which is always up-to-date.

matklad: In the read path, when we fetch message from in-flight write, we assert that the checksum matches. Can we hir that?

sentientwaffle: I'm not sure what you mean :thinking: 
Are you suggesting we pass the entry into `remove_reply`, and assert that the latest write (if it exists) matches that entry?
(Ignoring `header.size == sizeOf(Header)` I guess.)

sentientwaffle: Oh just realized that s/hir/hit

But now I don't understand how we could hit that assert :sweat_smile:

matklad: Yeah, I think I convinced myself that it is indeed unreachable

matklad (approved):

PR #641, opened 2023-04-18
There are some links 404 in readme

Miloas:


README.md
```diff
@@ -29,9 +29,12 @@ Check out [docs.tigerbeetle.com](https://docs.tigerbeetle.com/).
 Here are a few key pages you might be interested in:
 
 - Deployment
-  - [Hardware](https://docs.tigerbeetle.com/deployment/hardware)
-- Usage
-  - [Integration](https://docs.tigerbeetle.com/usage/integration)
+  - [Hardware](https://docs.tigerbeetle.com/deploy/hardware)
+- Quick start
```
eatonphil: Thanks for the catch! We validate links to internal files but since these are absolute links they aren't caught yet.

Instead of adding the Quick Start links could you keep them out as you have since the README itself already has parts of the quick start. And these links were not meant to be a replacement for the index page at docs.tigerbeetle.com.

Could you keep the links as they were but change Integration to point to https://docs.tigerbeetle.com/#designing-for-tigerbeetle ?

Thank you!

Miloas: done

eatonphil:
Thanks for the quick fix @Miloas ! :)

eatonphil (approved):

PR #645, opened 2023-04-19
VSR: Retain fault information during on_start_view

sentientwaffle:
Found by VOPR seed `12162487559395126725` on commit `5cebb0d9488e8da89681fad084d43035c697dc05`.
Manifested as a panic at:

    // Never nack op_head_min (aka commit_max).
    assert(op > op_head_min);

But I think it could be a correctness bug depending on the state of the other replicas that participate in the view-change.

## Pre-merge checklist

Performance:

### DIFF
--- a/src/vsr/replica.zig
+++ b/src/vsr/replica.zig
@@ -1665,7 +1665,25 @@ pub fn ReplicaType(
 
             for (view_headers.slice) |*header| {
                 if (header.op <= self.op_checkpoint_trigger()) {
+                    // Consider this case:
+                    // 1. Replica A prepares and commits op X (as backup).
+                    // 2. Replica A crashes.
+                    // 3. Prepare X is corrupted in the WAL.
+                    // 4. Replica A recovers into status=recovering_head.
+                    // 5. Replica A receives a start_view, which includes the header of X.
+                    // 6. Replica A participates in a view-change.
+                    // During this view-change, replica A *must not* nack op X.
+                    // Otherwise, op X might be truncated from the new view.
+                    // To ensure this, leave the entry as faulty.
+                    const slot = self.journal.slot_for_op(header.op);
+                    const fault = self.journal.faulty.bit(slot);
+
                     self.replace_header(header);
+
+                    if (fault) {
+                        assert(self.journal.dirty.bit(slot));
+                        self.journal.faulty.set(slot);
+                    }
                 }
             }
 
--- a/src/vsr/replica.zig
+++ b/src/vsr/replica.zig
@@ -470,8 +470,8 @@ pub fn ReplicaType(
                 maybe(self.journal.op_maximum() > vsr_headers.slice[0].op);
             }
 
-            // Try to use vsr_headers to update our head op.
-            // Loading headers prior to the head would risk the following scenario:
+            // Try to use vsr_headers to update our head op and its header.
+            // To avoid the following scenario, don't load headers prior to the head:
             // 1. Replica A prepares[/commits] op X.
             // 2. Replica A crashes.
             // 3. Prepare X is corrupted in the WAL.
--- a/src/vsr/replica.zig
+++ b/src/vsr/replica.zig
@@ -440,77 +440,69 @@ pub fn ReplicaType(
             if (self.journal.faulty.count == constants.journal_slot_count) return error.WALInvalid;
 
             const vsr_headers = self.superblock.working.vsr_headers();
-            for (vsr_headers.slice) |*header| {
-                if (vsr.Headers.dvc_header_type(header) != .valid) continue;
-
-                const slot = .{ .index = header.op % constants.journal_slot_count };
-                if (header.op > self.op_checkpoint_trigger()) {
-                    // Ignore an op that is too far head for our journal.
-                } else if (self.journal.has(header)) {
-                    // Header is already in the WAL.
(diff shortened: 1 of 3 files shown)


src/vsr/replica.zig
```diff
@@ -1665,7 +1665,25 @@ pub fn ReplicaType(
...
+                    // Consider this case:
+                    // 1. Replica A prepares and commits op X (as backup).
+                    // 2. Replica A crashes.
+                    // 3. Prepare X is corrupted in the WAL.
+                    // 4. Replica A recovers into status=recovering_head.
+                    // 5. Replica A receives a start_view, which includes the header of X.
+                    // 6. Replica A participates in a view-change.
+                    // During this view-change, replica A *must not* nack op X.
+                    // Otherwise, op X might be truncated from the new view.
+                    // To ensure this, leave the entry as faulty.
```
matklad: Could the same scenario happen in `primary_set_log_from_do_view_change_messages`?

If `4` is

* Replica A recovers into `.view_change`
* Replica `A` receives DVC quorum and bemoves primary
* Replica replaces `X` via canonical headers
* Another view change happens, and replica `A` erroneously nacks `X`

matklad: Hm, actually, shouldn't

```
            if (journal.has(header)) {
                assert(journal.dirty.bit(slot));
                // Do not clear any faulty bit for the same entry.
            } else {
```

logic in Journal already handle both of these cases?

sentientwaffle: To answer your second question — that journal logic doesn't always help.
In the example I put in the code, we have no header at all (`command=reserved`) prior to calling `replace_header` (the WAL was corrupt).

For your former question:
We could put this same logic in `primary_set_log_from_do_view_change_messages`, but I don't _think_ it is necessary.
The difference is that the headers of this new view cannot be lost, even if we crash + recover the WAL — because they are stored in our superblock's `vsr_headers`, and restored (if necessary) during `open`.

matklad: :thinking: tbh, I am not a fan of this save-restore logic, especially given that it papers over similar logic on journal. 

Could we do something like

```zig
            if (journal.has(header) or journal.faulty(slot) and journal.headers[slot].command == .reserved) {
                assert(journal.dirty.bit(slot));
                // Do not clear any faulty bit for potentially the same entry.
```

in the journal perhaps?

sentientwaffle: I need to play with this a bit, I think there are some tricky cases.
I think we need to leave `set_header_as_dirty` as-is since leaving faulty intact in all cases could impact availability.
But we could add a new function that does leave faulty intact, which the replica can use in some circumstances. (Hand-waiving here b/c I'm not sure which all circumstances those are yet...)

Also I think we can't check `command == reserved` because `Replica.open` loads headers from vsr_headers but leaves `faulty` intact.

sentientwaffle:
Updated with a different approach.
This is a more complicated change but the end result is a little simpler imo.

See commit message for explanation.

src/vsr/replica.zig
```diff
@@ -440,77 +440,69 @@ pub fn ReplicaType(
...
+                    op_head = vsr_header.op;
 
-            var op_head = for (vsr_headers.slice) |*header| {
-                if (vsr.Headers.dvc_header_type(header) != .valid) continue;
-                if (header.op <= self.op_checkpoint_trigger()) break header.op;
-            } else op_head: {
+                    if (!self.journal.has(vsr_header)) {
+                        self.journal.set_header_as_dirty(vsr_header);
+                    }
+                    break;
```
matklad, line 493: Maybe 

```
                    if (!self.journal.has(vsr_header)) {
                        assert(op_head == null or op_head.? < vsr_header.op)
                        self.journal.set_header_as_dirty(vsr_header);
                    }
                   op_head = vsr_header.op;

```

?

To double-check that vsr header and journal header don't disagree

src/vsr/replica.zig
```diff
@@ -440,77 +440,69 @@ pub fn ReplicaType(
...
+                //
+                // To avoid special-casing this all over, we pretend this higher op doesn't
+                // exist. This is safe because we never prepared any ops in the view we joined just
+                // before the crash.
+                assert(self.log_view < self.view);
+                maybe(self.journal.op_maximum() > vsr_headers.slice[0].op);
+            }
+
+            // Try to use vsr_headers to update our head op.
+            // Loading headers prior to the head would risk the following scenario:
```
matklad: This comment is a bit confusing (at least, I didn't get it until I also read the commit message). Maybe 

```
            // Try to use vsr_headers to update our head op and its header.
            // Don't load headers prior to the head to avoid the following scenario:
```

matklad (dismissed):
Two nits, otherwise LGTM!

src/vsr/replica.zig
sentientwaffle: :thinking: I think that assertion is actually not right though — if `op_head.? == vsr_header.op`, then we do want to `set_header_as_dirty` to replace the existing journal header with the one from vsr_headers. That is, our WAL has an old version of the op, but we serialized a view change with a newer header, but did not yet repair the WAL, then crashed.

matklad: Ah, right, journal and  SV may actually diagree with each other!

matklad (approved):

PR #646, opened 2023-04-20
vsr: remove dead code

matklad:
We now nack in DVC, not in request_prepare

## Pre-merge checklist

Performance:


sentientwaffle (approved):
Ah, thanks for catching this.
(And great branch name :grin:!)

PR #649, opened 2023-04-21
Trigger benchmark runs from CI

cb22:
This submits a job to run our benchmark suite on each commit, and brings in a very simple statsd implementation for us to push metrics. Currently, all that's done are a few in `benchmark.zig` that are used in our dashboard.

The statsd implementation does use our IO subsystem, but this is more of a stop-gap measure until we have a more holistic metric view in mind.

It also adds `--addresses` (and `--statsd`, `--print-batch-timings`) support to `benchmark.zig`.

You can see benchmark runs for any PR / commit by going into the Benchmark check, clicking summary, and following the link to Grafana. The benchmark parameters aren't configurable easily at the moment, but you can edit them in `scripts/benchmark/client.hcl` and push a commit, if you'd like to test a specific configuration.

## Pre-merge checklist

Performance:
* [x] I am very sure this PR could not affect performance. (heh)

### DIFF
--- a/src/statsd.zig
+++ b/src/statsd.zig
@@ -1,13 +1,18 @@
 const std = @import("std");
 
 const IO = @import("io.zig").IO;
+const FIFO = @import("fifo.zig").FIFO;
 
 pub const StatsD = struct {
     buffer: []u8,
     socket: std.os.socket_t,
-    address: std.net.Address,
-    address_size: u32,
+    io: *IO,
+    completions: []IO.Completion,
+    completions_fifo: FIFO(IO.Completion) = .{ .name = "statsd" },
 
+    /// Creates a statsd instance, which will send UDP packets via the IO instance provided.
+    /// Not thread safe, since it uses a single buffer shared between gauge and timing methods
+    /// with no locking.
     pub fn init(allocator: std.mem.Allocator, io: *IO, address: std.net.Address) !StatsD {
         // Limit the max packet size to 1000.
         const buffer = try allocator.alloc(u8, 1000);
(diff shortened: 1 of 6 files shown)


src/statsd.zig
```diff
@@ -0,0 +1,85 @@
...
+            .buffer_completions = buffer_completions,
+        };
+
+        for (buffer_completions) |*buffer_completion| {
+            buffer_completion.next = null;
+            statsd.buffer_completions_fifo.push(buffer_completion);
+        }
+
+        // 'Connect' the UDP socket, so we can just send() to it normally.
+        try std.os.connect(socket, &address.any, address.getOsSockLen());
```
kprotty, line 42: darwin/windows have `open_socket` set `SOCK_NONBLOCK`. For UDP sockets, does this just bind it to the address for future `send()`s and never return `error.WouldBlock`?

cb22: Good question - I know on Linux, it just binds it to the address and I believe it does the same on Darwin, both should never block. On Windows, https://learn.microsoft.com/en-us/windows/win32/api/winsock2/nf-winsock2-connect seems to suggest that we'll get the same behaviour:

> For a connectionless socket (for example, type SOCK_DGRAM), the operation performed by connect is merely to establish a default destination address that can be used on subsequent [send](https://learn.microsoft.com/en-us/windows/desktop/api/winsock2/nf-winsock2-send)/ [WSASend](https://learn.microsoft.com/en-us/windows/desktop/api/winsock2/nf-winsock2-wsasend) and [recv](https://learn.microsoft.com/en-us/windows/desktop/api/winsock/nf-winsock-recv)/ [WSARecv](https://learn.microsoft.com/en-us/windows/desktop/api/winsock2/nf-winsock2-wsarecv) calls

kprotty (approved):

PR #651, opened 2023-04-21
VSR: use Aegis128 over Blake3 for checksum

kprotty:
Ported from the stdlib's `std.crypto.aegis.State128L` with the output/cipher portions removed and only returning the mac. This version exists already as `std.crypto.aegis.Aegis128LMac` but it's not available in Zig 0.9.1. 

## Pre-merge checklist

Performance:

    ``` sh
    # benchmark results before
    1223 batches in 136.04 s
    load offered = 1000000 tx/s
    load accepted = 73510 tx/s
    batch latency p00 = 1 ms
    batch latency p10 = 27 ms
    batch latency p20 = 29 ms
    batch latency p30 = 29 ms
    batch latency p40 = 30 ms
    batch latency p50 = 31 ms
    batch latency p60 = 31 ms
    batch latency p70 = 32 ms
    batch latency p80 = 33 ms
    batch latency p90 = 34 ms
    batch latency p100 = 5740 ms
    transfer latency p00 = 1 ms
    transfer latency p10 = 4074 ms
    transfer latency p20 = 10701 ms
    transfer latency p30 = 21192 ms
    transfer latency p40 = 34426 ms
    transfer latency p50 = 49705 ms
    transfer latency p60 = 65163 ms
    transfer latency p70 = 75389 ms
    transfer latency p80 = 91266 ms
    transfer latency p90 = 107060 ms
    transfer latency p100 = 126048 ms
    
    # benchmark results after
    1223 batches in 103.34 s
    load offered = 1000000 tx/s
    load accepted = 96769 tx/s
    batch latency p00 = 0 ms
    batch latency p10 = 23 ms
    batch latency p20 = 25 ms
    batch latency p30 = 26 ms
    batch latency p40 = 26 ms
    batch latency p50 = 27 ms
    batch latency p60 = 28 ms
    batch latency p70 = 28 ms
    batch latency p80 = 29 ms
    batch latency p90 = 30 ms
    batch latency p100 = 3627 ms
    transfer latency p00 = 0 ms
    transfer latency p10 = 3597 ms
    transfer latency p20 = 8693 ms
    transfer latency p30 = 16661 ms
    transfer latency p40 = 26522 ms
    transfer latency p50 = 37787 ms
    transfer latency p60 = 49125 ms
    transfer latency p70 = 56922 ms
    transfer latency p80 = 68434 ms
    transfer latency p90 = 79917 ms
    transfer latency p100 = 93351 ms
    ```

### DIFF
--- a/src/vsr.zig
+++ b/src/vsr.zig
@@ -187,7 +187,17 @@ pub const Operation = enum(u8) {
 /// We reuse the same header for both so that prepare messages from the primary can simply be
 /// journalled as is by the backups without requiring any further modification.
 pub const Header = extern struct {
-    const checksum_body_empty = checksum(&.{});
+    var checksum_empty_once = std.once(checksum_empty_init);
+    var checksum_empty: u128 = undefined;
+
+    fn checksum_empty_init() void {
+        checksum_empty = checksum(&.{});
+    }
+
+    fn checksum_body_empty() u128 {
+        checksum_empty_once.call();
+        return checksum_empty;
+    }
 
     comptime {
         assert(@sizeOf(Header) == 128);
@@ -396,7 +406,7 @@ pub const Header = extern struct {
         if (self.view != 0) return "view != 0";
         if (self.commit != 0) return "commit != 0";
         if (self.timestamp != 0) return "timestamp != 0";
-        if (self.checksum_body != checksum_body_empty) return "checksum_body != expected";
+        if (self.checksum_body != checksum_body_empty()) return "checksum_body != expected";
         if (self.size != @sizeOf(Header)) return "size != @sizeOf(Header)";
         if (self.operation != .reserved) return "operation != .reserved";
         return null;
@@ -411,7 +421,7 @@ pub const Header = extern struct {
         if (self.view != 0) return "view != 0";
         if (self.commit != 0) return "commit != 0";
         if (self.timestamp == 0) return "timestamp == 0";
-        if (self.checksum_body != checksum_body_empty) return "checksum_body != expected";
+        if (self.checksum_body != checksum_body_empty()) return "checksum_body != expected";
         if (self.size != @sizeOf(Header)) return "size != @sizeOf(Header)";
         if (self.operation != .reserved) return "operation != .reserved";
         return null;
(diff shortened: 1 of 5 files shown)


src/vsr/superblock_fuzz.zig
```diff
@@ -89,11 +91,16 @@ fn run_fuzz(allocator: std.mem.Allocator, seed: u64, transitions_count_total: us
     var sequence_states = Environment.SequenceStates.init(allocator);
     defer sequence_states.deinit();
 
-    var env = Environment{
-        .sequence_states = sequence_states,
-        .superblock = &superblock,
-        .superblock_verify = &superblock_verify,
-    };
+    var env = Environment{ .sequence_states = sequence_states, .superblock = &superblock, .superblock_verify = &superblock_verify, .latest_vsr_state = SuperBlockHeader.VSRState{
```
cb22: Nit - line wrapping

src/vsr.zig
```diff
@@ -187,7 +187,18 @@ pub const Operation = enum(u8) {
...
 /// journalled as is by the backups without requiring any further modification.
 pub const Header = extern struct {
-    const checksum_body_empty = checksum(&.{});
+    var checksum_empty_once = std.once(checksum_empty_init);
+    var checksum_empty: u128 = undefined;
+    fn checksum_empty_init() void {
+        checksum_empty = checksum(&.{});
+    }
+
+    /// Aegis128 (used by checksum) uses hardware accelerated AES via inline asm which isn't
```
cb22, line 197: Did a bit of digging in to this - Zig's AES will use hardware accelerated AES either via AESNI or arm64's extensions if supported by the target CPU.

However - in benchmark.sh, we don't specify a target CPU - so I believe it'll use whatever is available on the machine building (I could be wrong on this - at the very least it uses something more than `baseline`) - but everywhere else we explicitly specify `baseline` as our target CPU, which doesn't get AES hardware acceleration.

I think we should specify a more modern CPU target all round...

src/vsr.zig
```diff
@@ -1055,11 +1066,79 @@ pub fn sector_ceil(offset: u64) u64 {
 }
 
 pub fn checksum(source: []const u8) u128 {
-    @setEvalBranchQuota(4000);
+    // From std.crypto.aegis.State128L:
```
cb22: Personal preference, so I'm happy to go either way, but I generally think it's nicer to keep backported code like this in a separate file.

src/vsr.zig
cb22: After chatting a bit with Joran (this is mostly off topic for this PR haha):
- We should build our binaries for a relatively modern CPU target (we do expect people to have NVMe storage, after all)
- If you're on an older CPU, or something bleeding edge, you can compile yourself and it'll use `native` CPU. We should document this.

Coming back to this PR - I think we can bump our baseline to `x86_64_v3+aes` and implement the above, I'm happy to do that in a separate PR

cb22 (approved):
LGTM! If we're going to change the hash function again though, perhaps we should hold off on merging this?

kprotty:
Don't think so. Aegis seems fast enough.

PR #654, opened 2023-04-21
Fix vopr debug builds - -Ddebug breaks zig build

cb22:
## Pre-merge checklist

Performance:

### DIFF
--- a/src/vopr.zig
+++ b/src/vopr.zig
@@ -138,8 +138,8 @@ fn build_simulator(
     mode: std.builtin.Mode,
 ) void {
     const mode_str = switch (mode) {
-        .Debug => "-Ddebug",
-        .ReleaseSafe => "-Drelease-safe",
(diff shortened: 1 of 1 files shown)


src/vopr.zig
```diff
@@ -138,7 +138,7 @@ fn build_simulator(
     mode: std.builtin.Mode,
 ) void {
     const mode_str = switch (mode) {
-        .Debug => "-Ddebug",
+        .Debug => "--",
```
sentientwaffle: What about using `-Drelease-safe=false` and `-Drelease-safe=true`? `--` works but might be surprising if we add other arguments in the future.

sentientwaffle (approved):

src/vopr.zig
cb22: Completely forgot that was possible :smile:

PR #682, opened 2023-04-25
LSM: Avoid sorting TableMutable when updated in sorted order.

kprotty:
Implements a wrapper around `[]Value`, `HashMap(u24)`, and `[]u8` for TableMutable. Values are appended to the array, their indexes are added to the HashMap, and there's a bitset to track which ones were removed. When iterating, removed Values are ignored using the bitset and we track if the values were inserted in sorted order. If so, we can skip `std.sort.sort`.

## Pre-merge checklist

Performance:

    ``` sh
    # benchmark results before
    1223 batches in 133.25 s
    load offered = 1000000 tx/s
    load accepted = 75047 tx/s
    batch latency p00 = 0 ms
    batch latency p10 = 27 ms
    batch latency p20 = 28 ms
    batch latency p30 = 29 ms
    batch latency p40 = 29 ms
    batch latency p50 = 30 ms
    batch latency p60 = 31 ms
    batch latency p70 = 32 ms
    batch latency p80 = 33 ms
    batch latency p90 = 34 ms
    batch latency p100 = 5404 ms
    transfer latency p00 = 0 ms
    transfer latency p10 = 4066 ms
    transfer latency p20 = 10537 ms
    transfer latency p30 = 20729 ms
    transfer latency p40 = 33777 ms
    transfer latency p50 = 48780 ms
    transfer latency p60 = 63997 ms
    transfer latency p70 = 74013 ms
    transfer latency p80 = 89395 ms
    transfer latency p90 = 104911 ms
    transfer latency p100 = 123261 ms
    
    # benchmark results after
    1223 batches in 123.42 s
    load offered = 1000000 tx/s
    load accepted = 81021 tx/s
    batch latency p00 = 0 ms
    batch latency p10 = 26 ms
    batch latency p20 = 27 ms
    batch latency p30 = 28 ms
    batch latency p40 = 28 ms
    batch latency p50 = 29 ms
    batch latency p60 = 30 ms
    batch latency p70 = 30 ms
    batch latency p80 = 31 ms
    batch latency p90 = 33 ms
    batch latency p100 = 5406 ms
    transfer latency p00 = 0 ms
    transfer latency p10 = 3215 ms
    transfer latency p20 = 9149 ms
    transfer latency p30 = 18426 ms
    transfer latency p40 = 30364 ms
    transfer latency p50 = 44402 ms
    transfer latency p60 = 58471 ms
    transfer latency p70 = 67372 ms
    transfer latency p80 = 81779 ms
    transfer latency p90 = 96208 ms
    transfer latency p100 = 113436 ms
    ```

### DIFF
--- a/src/lsm/table_mutable.zig
+++ b/src/lsm/table_mutable.zig
@@ -40,134 +40,168 @@ pub fn TableMutableType(comptime Table: type, comptime tree_name: [:0]const u8) 
         );
 
         const ValuesMap = struct {
-            const Values = struct {
-                active: [value_count_max / 8]u8,
-                items: [value_count_max]Value,
-            };
-
             const load_factor = 50;
-            const IndexMap = std.HashMapUnmanaged(u32, void, IndexMapContext, load_factor);
+            const capacity = math.ceilPowerOfTwo(u32, value_count_max * 100 / load_factor) catch unreachable;
 
-            const IndexMapContext = struct {
-                values: *const Values,
-                search_key: Key,
-                search_index: u32 = math.maxInt(u32),
+            const Slot = u24;
+            const slot_bytes = @bitSizeOf(Slot) / 8;
+            comptime {
+                assert(value_count_max <= std.math.maxInt(Slot));
+            }
 
-                inline fn key_from_index(ctx: *const IndexMapContext, index: u32) Key {
-                    if (index == ctx.search_index) return ctx.search_key;
-                    return key_from_value(&ctx.values.items[index]);
-                }
-
-                pub inline fn eql(ctx: IndexMapContext, a: u32, b: u32) bool {
-                    return compare_keys(ctx.key_from_index(a), ctx.key_from_index(b)) == .eq;
-                }
-
-                pub inline fn hash(ctx: IndexMapContext, index: u32) u64 {
-                    const key = ctx.key_from_index(index);
-                    return std.hash.Wyhash.hash(0, mem.asBytes(&key));
-                }
+            const Data = struct {
+                tags: [capacity]u8,
+                slots: [capacity * slot_bytes]u8,
+                active: [value_count_max / 8]u8,
+                values: [value_count_max]Value,
             };
 
-            index_map: IndexMap,
-            values: *Values,
-            active: u32,
+            len: u32 = 0,
+            used: u32 = 0,
+            data: *Data,
 
             pub fn init(allocator: mem.Allocator) !ValuesMap {
-                var index_map = IndexMap{};
-                try index_map.ensureTotalCapacityContext(allocator, value_count_max, undefined);
-                errdefer index_map.deinit(allocator);
-
-                const values = try allocator.create(Values);
-                errdefer allocator.destroy(values);
-
-                return ValuesMap{
-                    .index_map = index_map,
-                    .values = values,
-                    .active = 0,
-                };
+                var map = ValuesMap{ .data = try allocator.create(Data) };
+                map.clear();
+                return map;
             }
 
             pub fn deinit(map: *ValuesMap, allocator: mem.Allocator) void {
-                allocator.destroy(map.values);
-                map.index_map.deinit(allocator);
+                allocator.destroy(map.data);
             }
 
             pub inline fn count(map: *const ValuesMap) u32 {
-                return map.index_map.count();
+                return map.len;
             }
 
-            pub inline fn clear(map: *ValuesMap) void {
-                map.index_map.clearRetainingCapacity();
-                map.active = 0;
+            pub fn clear(map: *ValuesMap) void {
+                map.len = 0;
+                map.used = 0;
+                @memset(&map.data.tags, 0, @sizeOf(@TypeOf(map.data.tags)));
+                @memset(&map.data.active, 0, @sizeOf(@TypeOf(map.data.active)));
             }
 
-            pub fn find(map: *const ValuesMap, key: Key) ?*const Value {
-                const ctx = IndexMapContext{ .values = map.values, .search_key = key };
-                const index = map.index_map.getKeyAdapted(ctx.search_index, ctx) orelse return null;
-                return &map.values.items[index];
-            }
+            pub const Search = struct {
+                pos: u32,
+                tag: u8,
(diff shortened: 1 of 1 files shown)


src/lsm/table_mutable.zig
```diff
@@ -174,10 +297,16 @@ pub fn TableMutableType(comptime Table: type, comptime tree_name: [:0]const u8)
...
-            var it = table.values.keyIterator();
+            var sorted = true;
+            var it = table.values.iterator();
             while (it.next()) |value| : (i += 1) {
                 values_max[i] = value.*;
 
+                if (i > 0 and sorted) {
+                    const prev_key = key_from_value(&values_max[i - 1]);
+                    sorted = compare_keys(prev_key, key_from_value(value)) != .gt;
+                }
```
jorangreef: This is expensive to do in the data plane, since we're branching on `sorted` and comparing all keys.

The idea is that by this point, TableMutable should already know whether it needs to do any last minute sort finalization, or else if all keys are already sorted, and if so, it should be aiming to do a single memcpy here.

src/lsm/table_mutable.zig
```diff
@@ -127,23 +254,19 @@ pub fn TableMutableType(comptime Table: type, comptime tree_name: [:0]const u8)
...
                     }
                 },
                 .general => {
-                    // If the key is already present in the hash map, the old key will not be overwritten
-                    // by the new one if using e.g. putAssumeCapacity(). Instead we must use the lower
-                    // level getOrPut() API and manually overwrite the old key.
-                    const upsert = table.values.getOrPutAssumeCapacity(value.*);
-                    upsert.key_ptr.* = tombstone_from_key(key_from_value(value));
+                    // Overwrite the old key and value if any.
+                    table.values.upsert(value).* = tombstone_from_key(key_from_value(value));
```
jorangreef: Nice API improvement here with `upsert()`.

jorangreef: We could improve the comment, since this not only overwrites the old key if any, but will otherwise also insert.

src/lsm/table_mutable.zig
```diff
@@ -42,7 +39,142 @@ pub fn TableMutableType(comptime Table: type, comptime tree_name: [:0]const u8)
...
+
+            pub fn find(map: *const ValuesMap, key: Key) ?*const Value {
+                const ctx = IndexMapContext{ .values = map.values, .search_key = key };
+                const index = map.index_map.getKeyAdapted(ctx.search_index, ctx) orelse return null;
+                return &map.values.items[index];
+            }
+
+            pub fn upsert(map: *ValuesMap, value: *const Value) *Value {
+                const ctx = IndexMapContext{ .values = map.values, .search_key = key_from_value(value) };
+                const result = map.index_map.getOrPutAssumeCapacityContext(ctx.search_index, ctx);
```
jorangreef: Is this correct? Always passing `maxInt(u32)` (`ctx.search_index`) as the key to `getOrPutAssumeCapacityContext()`?

We need a fuzz test for this structure to make sure we're covering all the functionality we expect from it.

src/lsm/table_mutable.zig
```diff
@@ -42,7 +39,142 @@ pub fn TableMutableType(comptime Table: type, comptime tree_name: [:0]const u8)
...
+                map.values.active[index / 8] |= @as(u8, 1) << @intCast(u3, index % 8);
+                return &map.values.items[index];
+            }
+
+            pub fn fetch_remove(map: *ValuesMap, value: *const Value) ?*Value {
+                const ctx = IndexMapContext{ .values = map.values, .search_key = key_from_value(value) };
+                const kv = map.index_map.fetchRemoveContext(ctx.search_index, ctx) orelse return null;
+                const index = kv.key;
+
+                map.values.active[index / 8] &= ~(@as(u8, 1) << @intCast(u3, index % 8));
```
jorangreef:
```zig
[index / 8]
```

As per TigerStyle, let's always use `@divExact()` etc. to indicate what we intend precisely with respect to flooring.

jorangreef: Also, is there anything we expect of the existing value for `active[index / 8]` here?

Or do we intend to unilaterally `AND` it regardless of the existing value?

jorangreef: Is there anything we need to still do with respect to `map.active`?

jorangreef (changes requested):

src/lsm/table_mutable.zig
kprotty: `IndexMap` deals with indexes instead of Keys. `maxInt(u32)` is used as a dummy index to represent the key being passed in for lookup instead of the Key of a Value already in the array. `IndexMapContext` does the appropriate conversion for the hashmap.

src/lsm/table_mutable.zig
kprotty: Yes, we expect its bit to already be set. Ill add that.

src/lsm/table_mutable.zig
kprotty: The comparison of all keys to check if it's sorted (stops comparing when it's not sorted) either happens here or during `upsert/insert_no_clobber`. We can't do a memcpy as some Values may be deleted from `fetch_remove` and shouldn't get copied into `values_max` (the `active` bitset is used to detect this). It's also already traversed for updating the values_cache.

src/lsm/table_mutable.zig
kprotty: This was replaced with a custom map implementation.

src/lsm/table_mutable.zig
jorangreef: > The comparison of all keys to check if it's sorted (stops comparing when it's not sorted) either happens here or during upsert/insert_no_clobber.

Yes, however, doing it during upsert/insert would be more incremental.

> We can't do a memcpy as some Values may be deleted from fetch_remove and shouldn't get copied into values_max (the active bitset is used to detect this).

I mean that we should *aim* to do a single memcpy, i.e. we can iterate the bits quickly to find the biggest memcpy we can do. Then skip. Then big memcpy etc.

kprotty:
superseded by #784

PR #692, opened 2023-04-26
vsr: commit concurrently with repair

matklad:
Before, repair would only trigger commit if we collected all of the missing prepares. This is not optimal: if we have a WAL-full of prepares, and only the last couple are faulty, we want to start committing earlier prepares immediately, so that, by the time we get the requested prepares, those can be immediately committed as well.

SEED=16825366195277209080 shows how we don't commit earlier prepares if a later one is missing (although I _think_ this PR does not fix the root issue for that VOPR failure; rather it fixes a mostly unrelated problem).

## Pre-merge checklist

Performance:

### DIFF
--- a/src/vsr/replica.zig
+++ b/src/vsr/replica.zig
@@ -4363,8 +4363,6 @@ pub fn ReplicaType(
                 });
             }
 
-            assert(self.commit_max <= self.op);
-
             if (self.commit_min < self.commit_max) {
                 // Try to the commit prepares we already have, even if we don't have all of them.
                 // This helps when a replica is recovering from a crash and has a mostly intact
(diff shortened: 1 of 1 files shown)


src/vsr/replica.zig
```diff
@@ -4360,18 +4361,25 @@ pub fn ReplicaType(
...
-                return;
             }
-            assert(self.commit_max <= self.op);
 
-            if (self.status == .view_change and self.primary_index(self.view) == self.replica) {
+            if (self.status == .view_change and
+                self.primary_index(self.view) == self.replica and
+                self.commit_min == self.commit_max)
+            {
+                assert(self.journal.dirty.count == 0);
```
sentientwaffle, line 4380: I was a little concerned that this assert could be hit, because `repair_prepares` cleans up the "final" dirty bits [at the end](https://github.com/tigerbeetledb/tigerbeetle/blob/4e724158e006d7eb70b362aae84f28a9676527a6/src/vsr/replica.zig#L4809-L4838).

But I think its ok since `repair_prepares` would only early-return if all of the journal's write budget is used, and those writes imply `dirty > 0`.

sentientwaffle (approved):

PR #694, opened 2023-04-27
vopr: add logging for which WAL slot is corrupted

matklad:
## Pre-merge checklist

### DIFF
--- a/src/testing/storage.zig
+++ b/src/testing/storage.zig
@@ -439,11 +439,33 @@ pub const Storage = struct {
     fn fault_sector(storage: *Storage, zone: vsr.Zone, sector: usize) void {
         storage.faults.set(sector);
         if (storage.options.replica_index) |replica_index| {
-            log.debug("{}: corrupting sector at zone={} offset={}", .{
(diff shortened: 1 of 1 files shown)


src/testing/storage.zig
```diff
@@ -439,11 +439,31 @@ pub const Storage = struct {
...
+                    log.err(
+                        "{}: corrupting sector at zone={} offset={} slot={}",
+                        .{ replica_index, zone, offset, slot },
+                    );
+                },
+                .wal_headers => {
+                    comptime assert(constants.sector_size % @sizeOf(vsr.Header) == 0);
+                    const slot = @divFloor(offset, @sizeOf(vsr.Header));
+                    log.err(
+                        "{}: corrupting sector at zone={} offset={} slot={}",
```
sentientwaffle: Each `wal_headers` zone sector corresponds to `@divExact(constants.sector_size, @sizeOf(vsr.Header))` slots, so I think "slot" should be printed as a range.

sentientwaffle (approved):

PR #735, opened 2023-04-27
Trigger release after merge, not weekly

eatonphil:
Until we decide on a good naming convention we can share across all artifacts we release (which basically must be semver) this is just another random naming scheme for the release and built binaries.

### DIFF
--- a/scripts/next_version.sh
+++ b/scripts/next_version.sh
@@ -1,0 +1,16 @@
+#!/usr/bin/env sh
+
+set -eu
+
+# Update these as needed.
+MAJOR=0
+MINOR=13
+
+git fetch --tags --all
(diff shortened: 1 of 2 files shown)


scripts/next_version.sh
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env sh
+
+set -eu
+
+# Update these as needed.
+MAJOR=0
+MINOR=13
```
eatonphil, line 7: This isn't being used yet but once we can agree on a version scheme I'd start moving everything to be based on this.

cb22 (approved):
LGTM. When we add in the client versions (which we can do in a separate PR), I think we can make each client job output the final package name as an action output, and make this workflow called by our pipeline explicitly.

PR #736, opened 2023-04-27
Test/StateMachine: Use a simple LSM tree

sentientwaffle:
## Pre-merge checklist

Performance:

### DIFF
--- a/src/testing/state_machine.zig
+++ b/src/testing/state_machine.zig
@@ -1,10 +1,12 @@
 const std = @import("std");
 const assert = std.debug.assert;
+const log = std.log.scoped(.state_machine);
 
 const stdx = @import("../stdx.zig");
 const constants = @import("../constants.zig");
 const vsr = @import("../vsr.zig");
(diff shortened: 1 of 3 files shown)


src/testing/state_machine.zig
```diff
@@ -109,58 +165,56 @@ pub fn StateMachineType(
...
                 .register => return 0,
                 .echo => {
+                    const thing = state_machine.forest.grooves.things.get(123);
+                    const key: u64 = if (thing) |t| t.timestamp else timestamp;
+
+                    state_machine.forest.grooves.things.put(&.{
+                        .timestamp = key,
+                        .id = 123,
+                        .value = @truncate(u64, vsr.checksum(input)),
+                    });
```
kprotty, line 171: As the variables are now used, their `_ = variable` discard statements can be removed.

sentientwaffle: Good catch; updated.

kprotty (approved):

PR #737, opened 2023-04-27
cli: fix validation of --replica flag

matklad:
We properly assert bounds in replica.open, but of course we should give a nice error message during CLI parsing as well.

## Pre-merge checklist


eatonphil:
Looks better!

image

eatonphil (approved):

matklad:
I think PR waits for ages for mack builders, and the just times out?

PR #738, opened 2023-04-27
LSM/Grid: Block repair

sentientwaffle:
`Grid.Options.on_read_fault`: A callback invoked when any "normal" read fails. Will signal the `Replica` that the `read_faulty_queue` is non-empty to trigger block repair protocol.

Add `read_block_repair()`, which will be called by the replica when another replica asks for a particular block. Unlike `read_block()`, we do not necessarily expect/require the block to exist.

Add `write_block_repair()`, which replicas use to repair faulty blocks.

Repair reads do *not* queue up (`resolves`) behind non-repair reads. And repair reads targeting the same address but different checksums also do not queue up together. This is purely to keep the implementation of "on read" much simpler.

## Pre-merge checklist

Performance:

### DIFF
--- a/src/lsm/grid.zig
+++ b/src/lsm/grid.zig
@@ -1,5 +1,6 @@
 const std = @import("std");
 const assert = std.debug.assert;
+const maybe = stdx.maybe;
 const mem = std.mem;
 
 const constants = @import("../constants.zig");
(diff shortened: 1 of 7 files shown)


src/lsm/grid.zig
```diff
@@ -156,12 +162,16 @@ pub fn GridType(comptime Storage: type) type {
...
+        read_faulty_queue: FIFO(Read) = .{ .name = "grid_read_faulty" },
         // True if there's a read thats resolving callbacks. If so, the read cache must not be invalidated.
         read_resolving: bool = false,
 
-        pub fn init(allocator: mem.Allocator, superblock: *SuperBlock) !Grid {
+        on_read_fault: fn (*Grid, *Grid.Read) void,
+
+        pub fn init(allocator: mem.Allocator, options: struct {
+            superblock: *SuperBlock,
+            on_read_fault: fn (*Grid, *Grid.Read) void,
```
kprotty: Could this be optional (same inside Grid) and default initialize to null?

kprotty (approved):

PR #744, opened 2023-05-01
Grid: Allow read_block_repair of never-written blocks; Fuzz grid corruption/repair

sentientwaffle:
(See individual commit messages for more detail)

- [LSM/Grid: Allow read_block_repair of never-written blocks](https://github.com/tigerbeetledb/tigerbeetle/commit/ee410fcff3831209df42c537dafd5de1376f9405)
- [Test/Storage: Add grid faults (currently disabled)](https://github.com/tigerbeetledb/tigerbeetle/commit/e75668c1e0e52c784994d44c0257a6f04bfef890)
- [Test/FuzzTree: Test grid corruption](https://github.com/tigerbeetledb/tigerbeetle/commit/3f632c5b041dba7d38e42c7f958940d96a6195f5)

## Pre-merge checklist

Performance:

### DIFF
--- a/src/vsr/replica_test.zig
+++ b/src/vsr/replica_test.zig
@@ -714,7 +714,7 @@ const TestContext = struct {
     }
 
     pub fn run(t: *TestContext) void {
-        const tick_max = 2_400;
+        const tick_max = 3_000;
         var tick_count: usize = 0;
         while (tick_count < tick_max) : (tick_count += 1) {
             if (t.tick()) tick_count = 0;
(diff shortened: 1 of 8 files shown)


src/vsr/replica_test.zig
```diff
@@ -1000,7 +1001,7 @@ const TestClients = struct {
             }
         }
 
-        const tick_max = 2_400;
+        const tick_max = 3_000;
```
kprotty, line 1004: Out of curiosity, what's the reason for the bump?

sentientwaffle: The rng got permuted, and the new permutation takes a little longer to run one of the test cases.
Long-term rather than a `tick_max` counter, this should actually detect state changes across replicas and keep ticking until they stop. But that is not trivial so the counter will do for now.
We can set this counter as high as we like, but keeping it lower makes the tests faster.

kprotty (approved):

PR #745, opened 2023-05-01
Add single version for all releases

eatonphil:
This PR:

* Changes all client libraries, and the docker images, to only be published on a release.
* Releases happen after every merge to main now so this is effectively the same thing.
* The process grabs a new name from ./scripts/next_version.sh
* All client libraries and the docker images use the current git tag for their version.

Since we've already published 0.12.x of the Node client, we need to tag 0.13.0 to be the first tag to release.


.github/workflows/release_artifacts.yml
```diff
@@ -44,3 +43,31 @@ jobs:
           -H "Content-Type: application/zip" \
           --data-binary @./tigerbeetle-${{ matrix.target }}-${GIT_TAG}${{ matrix.debug }}.zip \
           "https://uploads.github.com/repos/tigerbeetledb/tigerbeetle/releases/$RELEASE_ID/assets?name=tigerbeetle-${{ matrix.target }}-${GIT_TAG}${{ matrix.debug }}.zip"
+
```
cb22, line 70: @eatonphil since we don't currently have de-dupe / caching of jobs, I'm worried about the extra jobs this will create since it'll essentially run the full pipeline again on main.

.github/workflows/linux.yml
```diff
@@ -186,7 +187,7 @@ jobs:
           username: ${{ github.repository_owner }}
           password: ${{ secrets.GITHUB_TOKEN }}
 
-      - if: github.ref == 'refs/heads/main'
+      - if: startsWith(github.ref, 'refs/tags/')
```
cb22: Will `github.ref` have the correct value here? Reason I ask is that release_artifacts.yml is invoked through `workflow_call` as a workaround and not `publish`, since the `hub release create` command is running with a `GITHUB_TOKEN`

.github/workflows/clients-dotnet.yml
```diff
@@ -34,17 +34,19 @@ jobs:
           /p:Threshold=\"95,85,95\" \
           /p:ThresholdType=\"line,branch,method\"
       - name: Pack
-        if: startsWith(github.ref, 'refs/tags/')
         working-directory: src/clients/dotnet
         run: |
+          version="$(git tag --points-at ${{ github.sha }})"
+          if [[ "$version" == "" ]]; then
```
batiati: Instead of using a dummy 0.0.1 version, you can just skip setting `/p:AssemblyVersion=$version /p:Version=$version` in this case. 
We only need to set a version number when publishing to Nuget. For the validation tests, the version number isn't required.

.github/workflows/release_artifacts.yml
eatonphil: Any suggestions?

.github/workflows/clients-dotnet.yml
eatonphil: Got it, updating, thanks!

.github/workflows/linux.yml
eatonphil: I've redone it so that the version is set once and then passed around to workflows as input.

.github/workflows/release_artifacts.yml
eatonphil: I've added more restrictions

cb22: Great!

cb22 (dismissed):
Not sure if you want to add the client versions to the release notes in this too?

eatonphil:
> Not sure if you want to add the client versions to the release notes in this too?

Good idea. I'll do that in a followup since there will likely be mistakes here.

eatonphil:
Nevermind, @cb22 I've added a message now.
image

eatonphil:
I had to dismiss your review to squash commits anyway.

eatonphil:
Ping @cb22 or @batiati for review.

cb22:
@eatonphil LGTM - but seems there's some conflicts now :sweat_smile:

eatonphil:
@cb22 ready to rip!

cb22 (approved):

PR #746, opened 2023-05-01
VSR: Grid repair protocol

sentientwaffle:
Repair corrupt grid blocks.

(Note that grid corruption is still disabled in the VOPR, pending state sync, because grid repair alone may stall if the block is unavailable. However, I have tested the VOPR locally to check that there are no VOPR issues besides the liveness issue.)

The `Replica.open()` sequence changes somewhat, because grid repair must run during `StateMachine.open()`. So where before the replica could run `StateMachine.open()` "synchronously", it must now run asynchronously. Commits are blocked until this finishes.

Related: a solo replica that starts with `op == commit_min` will now transition immediately to `status=normal` rather than waiting for `view_durable_update()` to complete. And `on_request` will reject incoming requests until `view_durable_update()` completes, to guard against the solo replica committing a message, then restarting and discarding it because it is part of a newer view.

## Pre-merge checklist

Performance:

### DIFF
--- a/scripts/.cspell.json
+++ b/scripts/.cspell.json
@@ -127,6 +127,7 @@
 	"threadpools",
 	"snapshotting",
 	"checkpointing",
-	"checkpointed"
+	"checkpointed",
+        "freeset"
     ]
 }
--- a/src/lsm/grid.zig
+++ b/src/lsm/grid.zig
@@ -728,6 +728,17 @@ pub fn GridType(comptime Storage: type) type {
         /// If the block is not present (or corrupt), we do not attempt to repair it — the repair's
         /// address/checksum is not assumed to match what "should" be in our block.
         /// Relatedly we still attempt to read free blocks — they might still hold the requested data.
+        ///
+        /// Even though this block is the block we expected to read, we can't safely
+        /// cache this result:
+        /// 1. Replica A write block X₁ to address X.
+        /// 2. Replica A write block X₂ to address X (write is lost/misdirected).
+        /// 3. Replica B requests block X₁ from replica A.
+        /// It is safe for A to send back X₁, but A must not allow it to poison its cache.
+        ///
+        /// Additionally, we don't cache these reads for performance reasons:
+        /// - it would fill the cache with non-temporally-local blocks, and
+        /// - it would force an extra memcpy.
         pub fn read_block_repair(
             grid: *Grid,
             callback: fn (*Grid.ReadRepair, error{BlockNotFound}!void) void,
@@ -755,13 +766,6 @@ pub fn GridType(comptime Storage: type) type {
                 .grid = grid,
             };
 
-            grid.on_next_tick(read_block_repair_tick_callback, &read.next_tick);
-        }
-
-        fn read_block_repair_tick_callback(next_tick: *Grid.NextTick) void {
-            const read = @fieldParentPtr(ReadRepair, "next_tick", next_tick);
-            const grid = read.grid;
-
             if (grid.cache.get_index(read.address)) |cache_index| {
                 const cache_block = grid.cache_blocks[cache_index];
 
@@ -771,10 +775,13 @@ pub fn GridType(comptime Storage: type) type {
 
                 if (header.checksum == read.checksum) {
                     stdx.copy_disjoint(.inexact, u8, read.block, cache_block[0..header.size]);
-                    read.callback(read, {});
                 } else {
-                    read.callback(read, error.BlockNotFound);
+                    // Signal to read_block_repair_tick_callback() that we found a block,
+                    // but not the one we wanted.
+                    std.mem.set(u8, read.block[0..header.size], 0);
                 }
+
+                grid.on_next_tick(read_block_repair_tick_callback, &read.next_tick);
             } else {
                 grid.superblock.storage.read_sectors(
                     read_block_repair_callback,
@@ -786,6 +793,17 @@ pub fn GridType(comptime Storage: type) type {
             }
         }
 
+        fn read_block_repair_tick_callback(next_tick: *Grid.NextTick) void {
+            const read = @fieldParentPtr(ReadRepair, "next_tick", next_tick);
+            const header = mem.bytesAsValue(vsr.Header, read.block[0..@sizeOf(vsr.Header)]);
+
+            if (header.op > 0) {
+                read.callback(read, {});
+            } else {
+                read.callback(read, error.BlockNotFound);
+            }
+        }
+
         fn read_block_repair_callback(completion: *Storage.Read) void {
             const read = @fieldParentPtr(ReadRepair, "completion", completion);
 
@@ -794,12 +812,6 @@ pub fn GridType(comptime Storage: type) type {
                 .checksum = read.checksum,
                 .block_type = null,
             })) {
-                // Even though this block is the block we expected to read, we can't safely
-                // cache this result:
-                // 1. Replica A write block X₁ to address X.
-                // 2. Replica A write block X₂ to address X (write is lost/misdirected).
-                // 3. Replica B requests block X₁ from replica A.
-                // It is safe for A to send back X₁, but A must not allow it to poison its cache.
                 read.callback(read, {});
             } else {
                 read.callback(read, error.BlockNotFound);
--- a/src/lsm/grid.zig
+++ b/src/lsm/grid.zig
@@ -85,14 +85,10 @@ pub fn GridType(comptime Storage: type) type {
         };
 
         pub const Read = struct {
-            callback: union(enum) {
-                normal: fn (*Grid.Read, BlockPtrConst) void, // repair == false
-                repair: fn (*Grid.Read, ?BlockPtrConst) void, // repair == true
-            },
+            callback: fn (*Grid.Read, BlockPtrConst) void,
             address: u64,
             checksum: u128,
-            block_type: ?BlockType,
-            repair: bool,
+            block_type: BlockType,
 
             pending: ReadPending = .{},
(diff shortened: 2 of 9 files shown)


src/vsr/replica.zig
```diff
@@ -666,6 +719,12 @@ pub fn ReplicaType(
             });
             errdefer self.grid.deinit(allocator);
 
+            for (self.grid_write_blocks) |*block, i| {
+                errdefer for (self.grid_write_blocks[0..i]) |b| allocator.free(b);
+                block.* = try alloc_block(allocator);
+            }
+            errdefer for (self.grid_write_blocks) |b| allocator.free(b);
```
matklad, line 728: I wonder if it makes sense to have

```zig
self.grid_write_blocks = alloc_blocks(allocator, constants.grid_repair_writes_max);
defere free_blocks(allocator, self.grid_write_blocks);
```

if only to avoid partial errdefers ?

I guess we can't easily express it in zig? We need "return array if size is comptime known, and a slice otherwise".

src/vsr/replica.zig
```diff
@@ -1939,6 +2008,203 @@ pub fn ReplicaType(
...
+                });
+
+                read.* = .{
+                    .replica = self,
+                    .destination = message.header.replica,
+                    .read = undefined,
+                };
+
+                self.grid.read_block_repair(
+                    on_block_read_repair,
```
matklad: Naming convention wise, shouldn't this be `on_request_blocks_read` (on_request_prepare / on_request_prepare_read)?

src/vsr/replica.zig
```diff
@@ -1939,6 +2008,203 @@ pub fn ReplicaType(
...
+                    &read.read,
+                    request.block_address,
+                    request.block_checksum,
+                );
+            }
+        }
+
+        fn on_block(self: *Self, message: *const Message) void {
+            maybe(!self.state_machine_opened);
```
```suggestion
            maybe(self.state_machine_opened);
```

I feel like semantically maybe already hints at both possibilities.

src/vsr/replica.zig
```diff
@@ -1939,6 +2008,203 @@ pub fn ReplicaType(
...
+            assert(message.header.command == .block);
+            assert(message.header.size <= constants.block_size);
```
matklad, line 2142: Should this be `assert(message.header.size <= constants.block_size + @sizeOf(Header));`?

matklad: Oh, no, it actually is correct, a block is a message! :exploding_head: we set `.command = .block,` in `lsm/table.zig`

src/vsr/replica.zig
```diff
@@ -1939,6 +2008,203 @@ pub fn ReplicaType(
...
+                message.header.checksum,
+            });
+
+            write.* = .{
+                .replica = self,
+                .write = undefined,
+            };
+
+            const block = &self.grid_write_blocks[self.grid_writes.index(write)];
+            stdx.copy_disjoint(.inexact, u8, block.*, message.buffer[0..message.header.size]);
```
matklad, line 2190: :thinking: so we have a memcpy on both read and write path here. This feels like a relatively big deal, as block repair is our most perf-sensitive operation (when we repair, we might have a lot of data to transfer). Why do we not use our pre-allocated message pool here?

src/vsr/replica.zig
```diff
@@ -1939,6 +2008,203 @@ pub fn ReplicaType(
...
+
+            self.grid.write_block_repair(
+                on_block_write_repair,
+                &write.write,
+                block,
+                message.header.op,
+            );
+        }
+
+        fn on_block_read_repair(grid_read: *Grid.Read, block_: ?Grid.BlockPtrConst) void {
```
matklad: Move that next to `on_request_blocks`, so that they can be read sequentially, as if we had async/await.

src/vsr/replica.zig
```diff
@@ -2181,6 +2447,57 @@ pub fn ReplicaType(
...
+            }
+            assert(requests_count > 0);
+            assert(requests_count <= constants.grid_repair_request_max);
+
+            message.header.* = .{
+                .command = .request_blocks,
+                .cluster = self.cluster,
+                .replica = self.replica,
+                .view = self.view,
+                .size = @sizeOf(Header) + @intCast(u32, requests_count) * @sizeOf(vsr.BlockRequest),
```
matklad: Maybe `var requests_count: u32 = 0;` instead?

src/vsr/replica.zig
```diff
@@ -6647,6 +6985,15 @@ pub fn ReplicaType(
...
                 read.block_type,
             });
+
+            if (self.solo()) @panic("grid is corrupt");
+
+            if (!self.grid_repair_message_timeout.ticking) {
+                assert(self.grid.read_faulty_queue.count == 1);
+
+                self.grid_repair_message_timeout.start();
+                self.on_grid_repair_message_timeout();
```
matklad: Hm, invoking the timeout manually is not something we do elsewhere, it is a bit surprising. 

Perhaps we should wait until the next replica tick? Right now we'll immediately send a request for a single block, but it doesn't seem unreasonable that more than one block would fail at the same time?

matklad (commented):
Two non-trivial questions:

* can we avoid memcpys here? feels relatively important
* manually invoking `on_grid_repair_message_timeout` feels iffy. Like, I don't think there are any specific problems there, but I don't like how we can reach that function both from replica.tick, and from grid's callback.

src/vsr/replica.zig
sentientwaffle: We could make it `alloc_blocks(slice_of_blocks);` (`!void` return type), which would be compatible with arrays or slices (caller must heap-allocate the slice for non-comptime-known capacity). But I think this is a common enough Zig idiom that single allocation is ok, even if it is ugly.

src/vsr/replica.zig
sentientwaffle: Yeah, I was on the fence about "wrapping" the block in an additional header when converting it to a message, but there wasn't a strong reason to.

src/lsm/grid.zig
```diff
@@ -768,33 +707,102 @@ pub fn GridType(comptime Storage: type) type {
...
             // of the callback as any nested Grid calls cannot synchronously update the cache).
-            switch (read.callback) {
-                .normal => |callback| callback(read, block.?),
-                .repair => |callback| callback(read, block),
+            read.callback(read, block);
+        }
+
+        /// If the block is not present (or corrupt), we do not attempt to repair it — the repair's
+        /// address/checksum is not assumed to match what "should" be in our block.
+        /// Relatedly we still attempt to read free blocks — they might still hold the requested data.
```
kprotty, line 730: Is there a semantic difference between "not present" and "free" (does the former refer to the cache)?

sentientwaffle: "Not present" means that we don't have the block in our cache _or_ disk.
"Free" means that the block's address is not held by the freeset.

src/vsr/replica.zig
```diff
@@ -6992,8 +6952,71 @@ pub fn ReplicaType(
                 assert(self.grid.read_faulty_queue.count == 1);
 
                 self.grid_repair_message_timeout.start();
-                self.on_grid_repair_message_timeout();
+                self.send_request_blocks();
+                self.grid.on_next_tick(
+                    on_grid_read_fault_next_tick,
+                    &self.grid_read_fault_next_tick,
+                );
```
matklad: Hm, this:

* requests blocks immediately
* and also schedules requesting blocks on the next tick

We should do only the latter, right?

src/lsm/grid.zig
```diff
@@ -768,33 +707,102 @@ pub fn GridType(comptime Storage: type) type {
...
+                .address = read.address,
+                .checksum = read.checksum,
+                .block_type = null,
+            })) {
+                // Even though this block is the block we expected to read, we can't safely
+                // cache this result:
+                // 1. Replica A write block X₁ to address X.
+                // 2. Replica A write block X₂ to address X (write is lost/misdirected).
+                // 3. Replica B requests block X₁ from replica A.
+                // It is safe for A to send back X₁, but A must not allow it to poison its cache.
```
matklad: Maybe move that to doc comment for `read_block_repair`, and also mention that caching would be bad from perf perspective because: 

* it will fill cache with non-temporally-local blocks
* it will force an extra memcpy.

src/lsm/grid.zig
```diff
@@ -768,33 +707,102 @@ pub fn GridType(comptime Storage: type) type {
...
+            } else {
+                grid.superblock.storage.read_sectors(
+                    read_block_repair_callback,
+                    &read.completion,
+                    read.block,
+                    .grid,
+                    block_offset(read.address),
+                );
+            }
+        }
```
matklad: I guess this could be a tiny bit more optimal and avoid introduction of next_tick delay if we are going to rid from storage anyway? 

```
if in_cache {
    memcpy()
    on_next_tick(|| read.callback())
} else {
    read_sectors
}
```

That is, copy data to message synchroneously, and then I think it would be safe to next-tick?

src/vsr/replica.zig
sentientwaffle: Good catch; I left the `send_request_blocks` call in by mistake. :facepalm:

matklad (approved):

scripts/.cspell.json
```diff
@@ -127,6 +127,7 @@
 	"threadpools",
 	"snapshotting",
 	"checkpointing",
-	"checkpointed"
+	"checkpointed",
+        "freeset"
```
eatonphil, line 131: Nice, glad this was somewhat obvious to edit

PR #751, opened 2023-05-02
VSR: Ack committed prepares

sentientwaffle:
Previously we did not ack messages that are committed by the cluster.
But that could that result in unnecessary view changes.

e.g. (replica_count=3):

1. R0 (primary) prepares op=10.
2. R1 prepares op=10.
3. R0 commits op=10.
4. (View change due to random network whatever. R1 and R2 participate in DVC quorum. R1 is new primary)
5. R1 and R2 currently believe the commit_max is 9.
6. If R2 is now partitioned, then R1 will keep trying to prepare op=10 — R0 has prepared (and committed!) op=10, but doesn't ack it. Eventually R1 gives up. The cluster can progress when R0 becomes primary or R2 un-partitions and prepares op=10.

Additionally, fix the root prepare's ack's validation. (It was previously dead code since we would never ack the root prepare.)

## Pre-merge checklist

Performance:

### DIFF
--- a/src/vsr/replica.zig
+++ b/src/vsr/replica.zig
@@ -5123,15 +5123,10 @@ pub fn ReplicaType(
             }
 
             assert(self.status == .normal);
-            // After a view change, replicas send prepare_oks for uncommitted ops with older views:
+            // After a view change, replicas send prepare_oks for ops with older views.
             // However, we only send to the primary of the current view (see below where we send).
             assert(header.view <= self.view);
             assert(header.op <= self.op);
-
-            if (header.op <= self.commit_max) {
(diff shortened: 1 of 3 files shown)


src/vsr/replica.zig
```diff
@@ -5128,11 +5128,6 @@ pub fn ReplicaType(
             assert(header.view <= self.view);
             assert(header.op <= self.op);
 
-            if (header.op <= self.commit_max) {
```
matklad, line 5131: Adjust the 

>            // After a view change, replicas send prepare_oks for uncommitted ops with older views:

comment, as we no longer ok only uncomitted ops.

matklad (approved):

PR #752, opened 2023-05-03
vsr: fix reference count discipline for `primary_pipeline_prepare`

matklad:
By convention, functions accepting a Message "borrow" the argument, and functions returning a Message return an owned copy.

primary_pipeline_prepare and push_prepare violated this guideline.

Found this out during reconfig work, where `primary_pipeline_prepare` might actually drop the request if reconfig is in progress.

## Pre-merge checklist

Performance:

### DIFF
--- a/src/vsr/replica.zig
+++ b/src/vsr/replica.zig
@@ -2601,19 +2601,22 @@ pub fn ReplicaType(
             assert(self.commit_min <= self.commit_max);
 
             if (self.status == .normal and self.primary()) {
-                const prepare = self.pipeline.queue.pop_prepare().?;
+                {
+                    const prepare = self.pipeline.queue.pop_prepare().?;
+                    defer self.message_bus.unref(prepare.message);
+
+                    assert(prepare.message.header.command == .prepare);
+                    assert(prepare.message.header.checksum == self.commit_prepare.?.header.checksum);
+                    assert(prepare.message.header.op == self.commit_min);
+                    assert(prepare.message.header.op == self.commit_max);
+                }
+
                 if (self.pipeline.queue.pop_request()) |request| {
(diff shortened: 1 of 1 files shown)


src/vsr/replica.zig
```diff
@@ -4126,6 +4129,8 @@ pub fn ReplicaType(
             self.pipeline.queue.verify();
 
             const message = request.message;
+            defer self.message_bus.unref(message);
```
sentientwaffle, line 4132: Doesn't this go against "functions accepting a Message borrow the argument"?

matklad: We accept a `Request` here, not the message. When call-site cosntructs a request, it does

```zig
            const request = .{
                .message = message.ref(),
                .realtime = realtime,
            };

            if (self.pipeline.queue.prepare_queue.full()) {
                self.pipeline.queue.push_request(request);
            } else {
            }
```

which makes sense --- every "stored" message is owned. 


But, yeah, I am not 100% happy here. But I am more happy with this than with 

```
pipeline_queue.push_prepare(prepare.ref());
```

sentientwaffle: Ok I'm following now.

Before, we would "move" the `request.message` from the request queue to the prepare queue (`pipeline.queue.push_prepare`).
Now, `primary_pipeline_prepare` unrefs the message, and `pipeline.queue.push_prepare` takes its own reference.

sentientwaffle (approved):

src/vsr/replica.zig
matklad: >Before, we would "move" the request.message from the request queue to the prepare queue (pipeline.queue.push_prepare).

Yeah, and that was heavily dependent on there not being any stray `return`s on the path between the two.

PR #753, opened 2023-05-04
Add timestamp to the list of fields in the modes table

eatonphil:
Some folks got confused by this field not being here.


sentientwaffle (approved):
Ah, I didn't include `timestamp` b/c this section was intended as "what the user/client needs to pass in when creating a transfer". But I can see how that would be confusing. :+1:

PR #764, opened 2023-05-08
LSM: Manifest(/Log) storage determinism

sentientwaffle:
Fix (grid) storage determinism (at checkpoint) of ManifestLog and SuperBlock Manifest.

These are the last two (known) storage determinism issues, so the `StorageChecker` is now enabled at checkpoint.
`StorageChecker` is still disabled for other compaction beats though (see code for explanation).

### Problems:

1. Order of `ManifestLog` appends depends on the ordering of compaction IO.
2. Order of `SuperBlock.Manifest` appends+removes depends on the ordering of compaction IO.

##### 1.

`Compaction` queues up `ManifestLog` appends. It holds onto them until `finish()` (formerly `reset()`) is invoked, rather than just appending as it progresses.

`compact_finish_join()` previously handled all of the compaction "prelude".
This is now in a separate step: `compact_end()`.
`compact_end()` is called in a consistent order (by `Groove` and `Forest`) regardless of compaction concurrency.

##### 1b:

`finish()` is now called _before_ `remove_invisible_tables` to ensure the invisible table cleanup doesn't lag behind.

##### 2.

`SuperBlock.Manifest` appends are performed by the `ManifestLog` when it flushes blocks to the disk.

During `ManifestLog.checkpoint()` this is a non-issue, since it happens synchronously with respect to the `Forest.checkpoint()`/`Tree.checkpoint()` invocation.
However, `ManifestLog.compact()` was previously called asynchronously with respect to `Forest.compact()`/`Tree.compact()`, and thus could be interleaved unpredictably.

`ManifestLog.compact()` now runs concurrently to compaction. This is safe now because `Compaction` is queuing its `ManifestLog` appends.

## Pre-merge checklist

Performance:

    ``` sh
    # benchmark results before
    1223 batches in 537.43 s
    load offered = 1000000 tx/s
    load accepted = 18607 tx/s
    batch latency p00 = 3 ms
    batch latency p10 = 56 ms
    batch latency p20 = 59 ms
    batch latency p30 = 62 ms
    batch latency p40 = 65 ms
    batch latency p50 = 88 ms
    batch latency p60 = 99 ms
    batch latency p70 = 103 ms
    batch latency p80 = 111 ms
    batch latency p90 = 125 ms
    batch latency p100 = 52028 ms
    transfer latency p00 = 3 ms
    transfer latency p10 = 10274 ms
    transfer latency p20 = 33719 ms
    transfer latency p30 = 69368 ms
    transfer latency p40 = 117064 ms
    transfer latency p50 = 178294 ms
    transfer latency p60 = 239797 ms
    transfer latency p70 = 300107 ms
    transfer latency p80 = 362484 ms
    transfer latency p90 = 421176 ms
    transfer latency p100 = 527439 ms

    
    # benchmark results after
    1223 batches in 492.30 s
    load offered = 1000000 tx/s
    load accepted = 20312 tx/s
    batch latency p00 = 3 ms
    batch latency p10 = 58 ms
    batch latency p20 = 60 ms
    batch latency p30 = 62 ms
    batch latency p40 = 65 ms
    batch latency p50 = 86 ms
    batch latency p60 = 97 ms
    batch latency p70 = 101 ms
    batch latency p80 = 105 ms
    batch latency p90 = 110 ms
    batch latency p100 = 26695 ms
    transfer latency p00 = 3 ms
    transfer latency p10 = 10475 ms
    transfer latency p20 = 34136 ms
    transfer latency p30 = 70119 ms
    transfer latency p40 = 118096 ms
    transfer latency p50 = 177697 ms
    transfer latency p60 = 239884 ms
    transfer latency p70 = 301274 ms
    transfer latency p80 = 364407 ms
    transfer latency p90 = 425338 ms
    transfer latency p100 = 482315 ms
    ```

### DIFF
--- a/src/lsm/README.md
+++ b/src/lsm/README.md
@@ -88,7 +88,7 @@ Invariants:
 
 ### Compaction Selection Policy
 
-Compaction targets the table from level `A` which overlaps the fewest tables of level `B`.
+Compaction targets the table from level `A` which overlaps the fewest visible tables of level `B`.
 
 For example, in the following table (with `lsm_growth_factor=2`), each table is depicted as the range of keys it includes. The tables with uppercase letters would be chosen for compaction next.
 
@@ -103,6 +103,31 @@ Links:
 - [`Manifest.compaction_table`](manifest.zig)
 - [Constructing and Analyzing the LSM Compaction Design Space](http://vldb.org/pvldb/vol14/p2216-sarkar.pdf) describes the tradeoffs of various data movement policies. TigerBeetle implements the "least overlapping with parent" policy.
 - [Option of Compaction Priority](https://rocksdb.org/blog/2016/01/29/compaction_pri.html)
+
+#### Compaction Move Table
+
+When the [selected input table](#compaction-selection-policy) from level `A` does not overlap _any_
+input tables in level `B`, the input table can be "moved" to level `B`.
+That is, instead of sort-merging `A` and `B`, just update the input table's metadata in the manifest.
+
+This is referred to as the _move table_ optimization.
+
+#### Compaction Table Overlap
+
+When compacting a table from level `A` to level `B`, what is the maximum number of level-`B` tables
+that may overlap the level-`A` table (i.e. the "worst case")?
+
+Perhaps surprisingly, this is `lsm_growth_factor`:
+
+- Tables within a level are disjoint.
+- Level `B` has at most `lsm_growth_factor` times as many tables as level `A`.
+- To trigger compaction, level `A`'s visible-table count exceeds
+  `table_count_max_for_level(lsm_growth_factor, level_a)`.
+- The [selection policy](#compaction-selection-policy) chooses the table from level `A`
+  which overlaps the fewest visible tables in level `B`.
+- If any table in level `A` overlaps _more than_ `lsm_growth_factor` tables in level `B`,
+  that implies the existence of a table in level `A` with _less than_ `lsm_growth_factor` overlap.
+  The latter table would be selected over the former.
 
 ## Snapshots
 
--- a/src/lsm/README.md
+++ b/src/lsm/README.md
@@ -88,7 +88,7 @@ Invariants:
 
 ### Compaction Selection Policy
 
-Compaction targets the table from level `A` which overlaps the fewest visible tables of level `B`.
+Compaction selects the table from level `A` which overlaps the fewest visible tables of level `B`.
 
 For example, in the following table (with `lsm_growth_factor=2`), each table is depicted as the range of keys it includes. The tables with uppercase letters would be chosen for compaction next.
 
@@ -111,6 +111,9 @@ input tables in level `B`, the input table can be "moved" to level `B`.
 That is, instead of sort-merging `A` and `B`, just update the input table's metadata in the manifest.
 
 This is referred to as the _move table_ optimization.
+
+Where a tree performs inserts mostly in sort order, with a minimum of updates, this _move table_
+optimization should enable the tree's performance to approach that of an append-only log.
 
 #### Compaction Table Overlap
 
--- a/src/lsm/compaction.zig
+++ b/src/lsm/compaction.zig
@@ -154,7 +154,8 @@ pub fn CompactionType(
                 pending: usize,
             },
             next_tick,
-            done,
+            done_writing_tables,
+            done_applied_manifest,
         },
 
         next_tick: Grid.NextTick = undefined,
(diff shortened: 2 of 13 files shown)


src/lsm/README.md
```diff
@@ -88,7 +88,7 @@ Invariants:
 
 ### Compaction Selection Policy
 
-Compaction targets the table from level `A` which overlaps the fewest tables of level `B`.
+Compaction targets the table from level `A` which overlaps the fewest visible tables of level `B`.
```
jorangreef:
```suggestion
Compaction selects the table from level `A`, which overlaps the fewest visible tables of level `B`.
```

src/lsm/README.md
```diff
@@ -104,6 +104,31 @@ Links:
 - [Constructing and Analyzing the LSM Compaction Design Space](http://vldb.org/pvldb/vol14/p2216-sarkar.pdf) describes the tradeoffs of various data movement policies. TigerBeetle implements the "least overlapping with parent" policy.
 - [Option of Compaction Priority](https://rocksdb.org/blog/2016/01/29/compaction_pri.html)
 
+#### Compaction Move Table
+
+When the [selected input table](#compaction-selection-policy) from level `A` does not overlap _any_
+input tables in level `B`, the input table can be "moved" to level `B`.
+That is, instead of sort-merging `A` and `B`, just update the input table's metadata in the manifest.
+
+This is referred to as the _move table_ optimization.
```
jorangreef, line 113:
```suggestion
This is referred to as the _move table_ optimization.

Where a tree performs inserts mostly in sort order, with a minimum of updates, this _move table_ optimization should enable the tree's performance to approach that of an append-only log.
```

jorangreef: Great to be building up our docs like this.

src/lsm/README.md
```diff
@@ -104,6 +104,31 @@ Links:
...
+
+- Tables within a level are disjoint.
+- Level `B` has at most `lsm_growth_factor` times as many tables as level `A`.
+- To trigger compaction, level `A`'s visible-table count exceeds
+  `table_count_max_for_level(lsm_growth_factor, level_a)`.
+- The [selection policy](#compaction-selection-policy) chooses the table from level `A`
+  which overlaps the fewest visible tables in level `B`.
+- If any table in level `A` overlaps _more than_ `lsm_growth_factor` tables in level `B`,
+  that implies the existence of a table in level `A` with _less than_ `lsm_growth_factor` overlap.
+  The latter table would be selected over the former.
```
jorangreef, line 133: Crystal clear.

src/lsm/compaction.zig
```diff
@@ -133,7 +154,8 @@ pub fn CompactionType(
                 pending: usize,
             },
             next_tick,
-            done,
+            done_writing_tables,
+            done_applied_manifest,
```
jorangreef:
```suggestion
            wrote_tables,
            applied_to_manifest,
```

To keep the tense consistent.

Would be great to also fix up the other enums here. For example `iter_init_a` should be `iterator` instead of `iter` (no abbreviations per TigerStyle, unless we're using `it` in a local context).

Even better, how can we use MSB order, and relate all the enums to each other? For example, `tables_writing` and `tables_written` (or `tables_writing_done`) instead of `writing` and `done_writing_tables`?

src/lsm/compaction.zig
```diff
@@ -198,11 +220,13 @@ pub fn CompactionType(
...
         pub fn reset(compaction: *Compaction) void {
-            assert(compaction.state == .done);
+            assert(compaction.state == .done_applied_manifest);
+
             compaction.state = .idle;
+            compaction.manifest_entries.len = 0;
             if (compaction.grid_reservation) |grid_reservation| {
-                compaction.context.grid.forfeit(grid_reservation);
                 compaction.grid_reservation = null;
+                compaction.context.grid.forfeit(grid_reservation);
```
jorangreef: Nulling out `grid_reservation` before passing it to `forfeit()`, and then relying on the capture to copy, was a surprising read. Would feel better to pass the memory directly, and then null.

src/lsm/tree.zig
```diff
@@ -110,6 +110,14 @@ pub fn TreeType(comptime TreeTable: type, comptime Storage: type, comptime tree_
 
         manifest: Manifest,
 
+        compaction_phase: enum {
+            idle,
+            skipped,
+            skipped_done,
+            running,
+            running_done,
+        } = .idle,
```
jorangreef, line 119: Reads nicely.

jorangreef (approved):

kprotty (approved):

PR #766, opened 2023-05-09
[Dotnet client] Set rollForward flag for runtime compatibility

batiati:
This change enables building and running the client and tests using newer versions of DotNet SDK, however the minimum required version was kept as is.

## Pre-merge checklist

Performance:


src/clients/dotnet/TigerBeetle.Benchmarks/TigerBeetle.Benchmarks.csproj
```diff
@@ -4,6 +4,7 @@
     <TargetFramework>net6.0</TargetFramework>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
     <UseCurrentRuntimeIdentifier>true</UseCurrentRuntimeIdentifier>
+		<RollForward>LatestMajor</RollForward>
```
eatonphil: Wanna fix the indentation in these two csproj files?

eatonphil (approved):

PR #767, opened 2023-05-09
vsr: correctly infer presence from DVC messages

matklad:
When a primary receives a DVC quorum, it should wait until each piepeline message is either nacked, or known to be present. In the ambiguous casees, the primary must wait for further DVCs. Otherwise, if the primary doesn't truncate a prepare which is not present, it will get stuck trying to repair it.

Before this PR, a prepare was considered as present if it wasn't nacked or a blank. This is wrong: if the prepare is faulty, but the header for it is available, it woun't be nacked, won't be a blank, and would not be present.

To fix this, use `client` field to send an explicit bitset of present prepares instead of inferring it.

The VOPRs seed makes the cluster to enter a view-change loop, where the cluster tries to repair a prepare which should have been truncated.

SEED: 4596119782490370426

## Pre-merge checklist

Performance:

### DIFF
--- a/src/vsr/replica.zig
+++ b/src/vsr/replica.zig
@@ -3398,7 +3398,14 @@ pub fn ReplicaType(
             const BitSet = std.bit_set.IntegerBitSet(128);
             comptime assert(BitSet.MaskInt == std.meta.fieldInfo(Header, .context).field_type);
 
+            // Collect nack and presence bits for the headers, so that the new primary can run CTRL
+            // protocol to truncate uncomitted headers. When:
+            // - a header has quorum of nacks -- the header is truncated
+            // - a header isn't truncated and is present -- the header gets into the next view
+            // - a header is neither truncated nor present -- the primary waits for more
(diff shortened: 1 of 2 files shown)


src/vsr/replica.zig
```diff
@@ -7395,12 +7411,15 @@ const DVCQuorum = struct {
                 assert(header.op == op);
 
                 const header_nacks = std.bit_set.IntegerBitSet(128){ .mask = dvc.header.context };
+                const header_present = std.bit_set.IntegerBitSet(128){ .mask = dvc.header.client };
```
sentientwaffle, line 7419: There is a "nacks" check in `DVCQuorum.verify_message()` that can be replicated for "present":
```
        const nacks = message.header.context;
        comptime assert(@TypeOf(nacks) == u128);
        assert(@popCount(u128, nacks) <= headers.slice.len);
        assert(@clz(u128, nacks) + headers.slice.len >= @bitSizeOf(u128));
```

sentientwaffle (approved):

PR #772, opened 2023-05-11
Thread version through release

eatonphil:

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -13,12 +13,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v3
-      with:
-        submodules: recursive
-        fetch-depth: 0
     - id: calculate_version
       run: |
-        git fetch --tags
         echo "version=$(./scripts/next_version.sh) >> "$GITHUB_OUTPUT"
   
   scheduled_release:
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -7,17 +7,31 @@ on:
...
+      version: ${{ steps.calculate_version.outputs.version }}
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+      with:
+        submodules: recursive
+        fetch-depth: 0
+    - id: calculate_version
+      run: |
+        git fetch --tags
```
batiati: The command `git fetch --tags` is already done inside `next_version.sh`.

.github/workflows/release.yml
```diff
@@ -7,17 +7,31 @@ on:
...
 
 jobs:
+  version:
+    outputs:
+      version: ${{ steps.calculate_version.outputs.version }}
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+      with:
+        submodules: recursive
```
batiati: Can we avoid loading the submodules for this step? It can save us a bit of CI time.

.github/workflows/release.yml
eatonphil: Good catch, thank you!

.github/workflows/release.yml
eatonphil: Done!

batiati (approved):

PR #779, opened 2023-05-12
Need to specify needs if using needs variables

eatonphil:
Trying to figure out why [this](https://github.com/tigerbeetledb/tigerbeetle/actions/runs/4961183702/jobs/8877677985) still does not trigger the publish step and I noticed that the workflow calls that reference `needs.version.outputs.version` doesn't actually set `need: [version]`.

### DIFF
--- a/.github/workflows/release_artifacts.yml
+++ b/.github/workflows/release_artifacts.yml
@@ -74,6 +74,7 @@ jobs:
   # Publish all clients and Docker image.
 
   client-dotnet:
+    needs: version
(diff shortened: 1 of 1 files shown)


eatonphil:
Sorry @batiati there's not a way to test this other than merging since it will end up publishing real versions to package managers.

batiati (approved):
> Sorry @batiati there's not a way to test this other than merging since it will end up publishing real versions to package managers.

Don't worry!!

PR #785, opened 2023-05-15
Propagate secrets into release_artifacts workflow call

eatonphil:


batiati (dismissed):
I'm just thinking out loud, we could standardize the secrets names/prefixes in a next opportunity, (e.g `NODE`, `GO`, `JAVA`, `DOTNET`)

eatonphil:
Yeah good point.

eatonphil:
Sorry had to fix a redundant call to the release artifacts step in pipeline.yml. Needs a reapproval when you can.

batiati (approved):

.github/workflows/pipeline.yml
```diff
@@ -25,10 +25,6 @@ jobs:
     uses: ./.github/workflows/windows.yml
   macos:
     uses: ./.github/workflows/macos.yml
-  release_test:
```
eatonphil, line 28: This was more necessary before when releases were once a week. But now this runs much of the whole steps twice on every run. When uploads should already be gated by the equivalent of when dry-run was false. So just dropping completely.

eatonphil:
Thank you!

PR #564, opened 2023-03-11
Increase the grid cache size in production to 16gb

jamii:
Dev and test are left at 128mb, as before.

Doesn't improve performance, at least for small benchmarks, as discussed in https://github.com/tigerbeetledb/tigerbeetle/issues/551#issuecomment-1462853718. Even though the db file is 13gb.

## Pre-merge checklist

Performance:

* [x] Compare `zig benchmark` on linux before and after this pr.
``` sh
# benchmark results before
1223 batches in 105.79 s
load offered = 1000000 tx/s
load accepted = 94527 tx/s
batch latency p00 = 5 ms
batch latency p10 = 31 ms
batch latency p20 = 32 ms
batch latency p30 = 32 ms
batch latency p40 = 32 ms
batch latency p50 = 36 ms
batch latency p60 = 37 ms
batch latency p70 = 37 ms
batch latency p80 = 37 ms
batch latency p90 = 37 ms
batch latency p100 = 4049 ms
transfer latency p00 = 5 ms
transfer latency p10 = 3565 ms
transfer latency p20 = 10669 ms
transfer latency p30 = 17949 ms
transfer latency p40 = 26392 ms
transfer latency p50 = 37194 ms
transfer latency p60 = 48950 ms
transfer latency p70 = 60220 ms
transfer latency p80 = 72309 ms
transfer latency p90 = 84444 ms
transfer latency p100 = 95760 ms
    
# benchmark results after
1223 batches in 102.93 s
load offered = 1000000 tx/s
load accepted = 97149 tx/s
batch latency p00 = 5 ms
batch latency p10 = 31 ms
batch latency p20 = 32 ms
batch latency p30 = 32 ms
batch latency p40 = 32 ms
batch latency p50 = 32 ms
batch latency p60 = 36 ms
batch latency p70 = 37 ms
batch latency p80 = 37 ms
batch latency p90 = 37 ms
batch latency p100 = 5151 ms
transfer latency p00 = 5 ms
transfer latency p10 = 3496 ms
transfer latency p20 = 10126 ms
transfer latency p30 = 17715 ms
transfer latency p40 = 25784 ms
transfer latency p50 = 35901 ms
transfer latency p60 = 47219 ms
transfer latency p70 = 57791 ms
transfer latency p80 = 70066 ms
transfer latency p90 = 82694 ms
transfer latency p100 = 92901 ms
```
OR
* [ ] I am very sure this PR could not affect performance.

PR #790, opened 2023-05-16
Bump default grid cache to 1GB, make it a CLI arg, and relax SetAssociativeCache sizing

### LINKED ISSUE #277: CLI: cache-blocks argument
Similar to https://github.com/tigerbeetledb/tigerbeetle/issues/276.

Right now the number of blocks in the grid cache is a constant: https://github.com/tigerbeetledb/tigerbeetle/blob/d15acc663f8882cb02413129e8351bf3238335e6/src/lsm/grid.zig#L154

Instead there should be a runtime CLI argument (`--cache-blocks`, parsed with `parse_size`) and a default (in `config.zig`).

`Grid` should get the count from an `options: GridOptions`.

cb22:
This adds a new parameter, `--cache-grid`, which now defaults to 1GB (up from 128MB). Thought it would be more ergonomic for the operator to be able to specify a size in MB/GB and compute the final count, rather than specifying a count directly.

It also relaxes the assertion in `SetAssociateCache` that `value_count_max` must be a power of 2. This would leave a lot of memory on the table for larger cache sizes. Instead, `value_count_max_multiple` is exposed on a returned Cache type, of which `value_count_max` must be a multiple.

Fixes #277 and replaces #564

## Pre-merge checklist

Performance:

(well, it will improve performance, but only on larger datasets than our current benchmark.sh runs)


src/tigerbeetle/cli.zig
```diff
@@ -59,6 +59,12 @@ const usage = fmt.comptimePrint(
     \\        will be used.
     \\        "addresses[i]" corresponds to replica "i".
     \\
+    \\  --grid-cache-size=<size>
+    \\        Set the grid cache size, size prefixes (MB, GB) are supported. The grid cache acts
+    \\        like a page cache for TigerBeetle, and should be set as large as possible.
+    \\        On a machine running only TigerBeetle, this is somewhere around
+    \\        (Total RAM) - 3GB (TigerBeetle) - 1GB (System), eg 12GB for a 16GB machine.
```
sentientwaffle: What about

```
--grid-cache-size=<number><MB|GB>
```
? Doesn't have to be exactly that, but we should specify the expected format more clearly somehow.

src/tigerbeetle/cli.zig
```diff
@@ -312,6 +324,32 @@ fn parse_storage_size(size_string: ?[]const u8) u64 {
...
+    // bound. We need to be a multiple of our Set Associative Cache's ways and still
+    // meet other assertions, but we relax the power of 2 requirement for the grid
+    // cache specifically.
+    // 2048 is the lowest multiple that can satisfy all those requirements for the
+    // current grid configuration.
+    const sac_multiple = 2048;
+    const block_count = @divFloor(
+        @divFloor(target_size, constants.block_size),
+        sac_multiple,
+    ) * sac_multiple;
```
sentientwaffle: We try to avoid acronyms (except for very common ones like VSR/LSM) ­— what about calling this "cache_entry_multiple"?

src/tigerbeetle/cli.zig
```diff
@@ -21,7 +21,7 @@ const usage = fmt.comptimePrint(
     \\
     \\  tigerbeetle format --cluster=<integer> --replica=<index> --replica-count=<integer> <path>
     \\
-    \\  tigerbeetle start --addresses=<addresses> <path>
+    \\  tigerbeetle start --addresses=<addresses> [--grid-cache-size=<size>] <path>
```
sentientwaffle: Rename to `--cache-grid` to match the value cache (`--cache-transfers`)

src/tigerbeetle/cli.zig
sentientwaffle: Can you split these up into e.g.

```
    const block_count_limit = @divFloor(target_size, constants.block_size);
    const block_count_rounded = @divFloor(block_count_limit, sac_multiple) * sac_multiple;
```

src/lsm/set_associative_cache.zig
```diff
@@ -128,14 +129,21 @@ pub fn SetAssociativeCache(
...
+            const sets = @divExact(value_count_max, layout.ways);
+
+            switch (value_count_max_requirement) {
+                .power_of_2 => {
+                    assert(math.isPowerOfTwo(value_count_max));
+                    assert(math.isPowerOfTwo(sets));
+                },
+
+                .relaxed => {},
+            }
```
sentientwaffle: Nit: Since these are more "specific" checks, I think they should follow 

```
            assert(value_count_max > 0);
            assert(value_count_max >= layout.ways);
            assert(value_count_max % layout.ways == 0);
```

src/tigerbeetle/cli.zig
```diff
@@ -312,6 +324,32 @@ fn parse_storage_size(size_string: ?[]const u8) u64 {
...
+fn parse_grid_cache_size(size_string: ?[]const u8) u64 {
+    const target_size = if (size_string) |s| parse_size(s) else return constants.grid_cache_blocks_default;
+
+    // Calculate a grid cache size, using the target_size provided as an upper
+    // bound. We need to be a multiple of our Set Associative Cache's ways and still
+    // meet other assertions, but we relax the power of 2 requirement for the grid
+    // cache specifically.
+    // 2048 is the lowest multiple that can satisfy all those requirements for the
+    // current grid configuration.
+    const sac_multiple = 2048;
```
sentientwaffle: Let's turn this into a constant that is exported by the `SetAssociativeCache`. Then we would access it via `Grid.Cache.(constant)`.

Also, ideally it could be computed rather than hard-coded at 2048.

src/lsm/set_associative_cache.zig
```diff
@@ -128,14 +129,21 @@ pub fn SetAssociativeCache(
...
 
         pub fn init(allocator: mem.Allocator, value_count_max: u64) !Self {
-            assert(math.isPowerOfTwo(value_count_max));
+            const sets = @divExact(value_count_max, layout.ways);
+
+            switch (value_count_max_requirement) {
+                .power_of_2 => {
+                    assert(math.isPowerOfTwo(value_count_max));
+                    assert(math.isPowerOfTwo(sets));
+                },
```
sentientwaffle: Could we remove this constraint for all caches? I don't understand what the power-of-2 buys us.

src/tigerbeetle/cli.zig
```diff
@@ -312,6 +326,28 @@ fn parse_storage_size(size_string: ?[]const u8) u64 {
     return size;
 }
 
+fn parse_cache_grid(size_string: ?[]const u8) u64 {
```
sentientwaffle: `parse_size_to_count` should probably be updated as well — maybe remove `parse_size_to_count` and then rename `parse_cache_grid` to `parse_cache_size_to_count`? It seems like these do pretty much the same thing now.

src/lsm/manifest_log_fuzz.zig
```diff
@@ -84,11 +84,13 @@ fn run_fuzz(
 
     var grid = try Grid.init(allocator, .{
         .superblock = &superblock,
+        .cache_blocks_count = 2048,
```
sentientwaffle: All of these `2048`s should be replaced by `Grid.value_count_max_multiple` (in this file and elsewhere).

src/tigerbeetle/cli.zig
```diff
@@ -89,6 +97,7 @@ pub const Command = union(enum) {
         cache_transfers: u32,
         cache_transfers_posted: u32,
         storage_size_limit: u64,
+        grid_cache_blocks_count: u64,
```
sentientwaffle: s/grid_cache_blocks_count/cache_grid_blocks (to match the value cache parameters)

src/tigerbeetle/cli.zig
sentientwaffle: Though it will probably need to take the respective `comptime SetAssociativeCache: type` as a parameter, rather than always using `Grid.Cache`

src/lsm/manifest_log_fuzz.zig
sentientwaffle: (Including in `simulator.zig`.)

src/tigerbeetle/cli.zig
cb22: I've done this - but it ended up being a bit convoluted to get access to the underlying `SetAssociateCache` from `cli.zig` - not sure if that's the best way 🫤

sentientwaffle (dismissed):

sentientwaffle:
The changes LGTM, but can you squash to get rid of the `Merge branch 'main' into bump-cache` commit? I think it would mess up git-blame if it merged.

cb22:
@sentientwaffle sure, done!

sentientwaffle (approved):

PR #792, opened 2023-05-18
fix: undefined reference to `__zig_probe_stack'

kassane:
Allows another C compiler link to this library.

If the library allowed build `ReleaseFast` then this command would be useless.

cc: @batiati @kprotty 

**Reference:**
- https://github.com/ziglang/zig/issues/6817#issuecomment-736129115

> Now you can pass -fcompiler-rt to include compiler-rt in static libraries. Feel free to open a proposal to change the default, which is currently only dynamic libraries and executables get compiler-rt included. - **Andrewrk**

## Pre-merge checklist

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -774,6 +774,7 @@ fn c_client(
 
         const shared_lib = b.addSharedLibrary("tb_client", "src/clients/c/tb_client.zig", .unversioned);
         const static_lib = b.addStaticLibrary("tb_client", "src/clients/c/tb_client.zig");
+        static_lib.bundle_compiler_rt = true;
 
         for ([_]*std.build.LibExeObjStep{ shared_lib, static_lib }) |lib| {
             lib.setMainPkgPath("src");
--- a/build.zig
+++ b/build.zig
@@ -775,6 +775,7 @@ fn c_client(
         const shared_lib = b.addSharedLibrary("tb_client", "src/clients/c/tb_client.zig", .unversioned);
         const static_lib = b.addStaticLibrary("tb_client", "src/clients/c/tb_client.zig");
         static_lib.bundle_compiler_rt = true;
+        static_lib.pie = true;
 
         for ([_]*std.build.LibExeObjStep{ shared_lib, static_lib }) |lib| {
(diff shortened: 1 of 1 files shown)


eatonphil:
> If the library allowed build ReleaseFast then this command would be useless.

I'm pretty sure we have comptime checks that fail the build if you try to do ReleaseFast.

batiati (changes requested):
Thanks Matheus,

It's worth adding a comment:

```zig
// Bundle rt to allow linking this static library
// with the final artifact using something other than Zig.
```

Could you please add `static_lib.pie = true` as well? So we follow the same configuration used by Go client's static libraries:

https://github.com/tigerbeetledb/tigerbeetle/blob/1c4000cbd5c4fb8c892019dd29787d1da0ae1e09/build.zig#L597-L598

kassane:
Hi @batiati ,

I have added the suggestion and the build has worked with gcc 13 (archlinux).
A doubt, does this change overwrite other bindings (duplicate parameters)?

At first sight, I think that the settings between bindings are standalone. Right?

batiati:
> A doubt, does this change overwrite other bindings (duplicate parameters)?
> At first sight, I think that the settings between bindings are standalone. Right?

I'm not sure if I'm following,  but the other bindings (clients) compile their libs independently, so it's not duplicated but only following the same build configuration for similar use cases (i.e. both c_client and go_client produce static libraries).

Did I get your question answered?

(...)

Maybe we can extract this part in a helper function, just like `link_tracer_backend` is called by every artifact, we can then create a `fn static_lib_config(...)` for doing that repetitive work of setting up correctly a static lib.

We can also place the comment there:

```Zig
// Bundle rt to allow linking this static library
// with the final artifact using something other than Zig.
```

Looking at the code, shared libs have even more repeated code, but it can be done by another PR!

kassane:
>Did I get your question answered?

Yeah!!

> Looking at the code, shared libs have even more repeated code, but it can be done by another PR!

Ok.

batiati (approved):

PR #793, opened 2023-05-19
[C Client] tb_client refactor

batiati:
## This PR:

The goal is to provide a simpler way to implement language clients, not requiring the target programming language to implement synchronization abstractions to handle the packet list from multiple concurrent application threads.

NOTE: The main use case is for refactoring the Node.js client, which will be part of a next PR (to make the review more convenient).

### Changes:

- Introduces `tb_client_acquire_packet` and `tb_client_release_packet` functions.
Those functions are thread-safe and intended to simplify the concurrency control that had to be done by the language client (e.g. using channels, semaphores, etc).
This is a **breaking API change for all clients** (a good one) since the language client has to handle `packet == null` if `maxConcurrency` is exceeded and notify the application through a new error/exception.
The status quo implementation for go/java/c# would block/wait until a packet became available, which is undesirable because it makes the max concurrency setting almost pointless from the application point of view.

- `tb_client_deinit` waits for all inflight requests to finish before stopping the io thread and freeing the client.
During the deinitialization, all available packets are "drained" and any call to `tb_client_acquire_packet` will return `status == .shutdown`, preventing the application from placing new requests.
 Previously, each programming language had to ensure that there was no thread waiting for a reply before closing the client.

- Refactored: tests, c sample, java client, dotnet client.

- Renamed `packet_count` and `max_concurrency` to `concurrency_max`.
Also, `PacketCountInvalid` and `packet_count_invalid` were renamed accordingly.
Note: `concurrencyMax` is validated to be >0 and <4096 (arbitrary upper limit).

- `tb_client_submit` accepts a single `tb_packet_t`, removing the type `tb_packet_list_t` from the public API.
It makes the API simpler since there's no support for the application to submit more than one packet at once.

## Pre-merge checklist

Performance:


src/clients/c/tb_client/packet.zig
```diff
@@ -50,31 +51,64 @@ pub const Packet = extern struct {
...
+                    head,
+                    next,
+                    .Release,
+                    .Monotonic,
+                ) orelse {
+                    head.?.next = null;
+                    return head;
+                };
+            }
+        }
```
kprotty: The head node must be loaded with `Acquire` to ensure push()'s `packet.next` is written before it reads it. There's also an [ABA issue](https://en.wikipedia.org/wiki/ABA_problem#Examples) here. For single dequeues, itll be simpler to use a `std.Thread.Mutex` which protects `head` (instead of atomics).

batiati: Thanks for pointing out the ABA issue!
Should we use an `AtomicMutex` to get rid of pthread mutexes? At least for the happy path, it seems a better choice.

kprotty: pthread mutexes and AtomicMutex have the same speed uncontended. The latter *may* be faster for contended cases, but pthread mutex is fast enough.

kprotty (approved):

zetanumbers:
Making my rust client bindings, for which i decided to defer client deinitialization to the `on_completion` call via (atomic) reference counting, instead of blocking during destruction.

batiati:
> Making my rust client bindings, for which i decided to defer client deinitialization to the `on_completion` call via (atomic) reference counting, instead of blocking during destruction.

That's a nice approach! I think that in general, blocking on `tb_client_deinit` makes it easier for the application to know for how long the `tb_client_t` reference is still valid.

During `tb_client_deinit`, any other call using `tb_client_t` may segfault due to use after free. This reference is only guaranteed to be valid until all ongoing requests have been processed, so, it still requires some control on the application's side to prevent _new_ threads from using an invalid handle during destruction, even before calling `tb_client_deinit`.

PR #804, opened 2023-05-24
tb_client: split C FFI interface and Zig native interface

rbino:
The init and init_echo functions can now be used to initialize the client from Zig, using some more idiomatic Zig conventions, namely being able to pass an explicit allocator, using a slice instead of ptr + len for the addresses string and returning an error.
The tb_client_ prefix was also removed from all the functions, while leaving it in the functions exported by the C FFI.
This commit also adapts the Java client (which uses Zig bindings) to the new API.

## Pre-merge checklist

Performance:

### DIFF
--- a/src/clients/c/tb_client.zig
+++ b/src/clients/c/tb_client.zig
@@ -33,6 +33,7 @@ const StateMachine = @import("../../state_machine.zig").StateMachineType(Storage
 
 const ContextType = @import("tb_client/context.zig").ContextType;
 const ContextImplementation = @import("tb_client/context.zig").ContextImplementation;
+pub const InitError = @import("tb_client/context.zig").Error;
 
 const DefaultContext = blk: {
     const Client = @import("../../vsr/client.zig").Client(StateMachine, MessageBus);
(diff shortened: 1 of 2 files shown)


src/clients/c/tb_client.zig
```diff
@@ -52,6 +53,57 @@ fn client_to_context(tb_client: tb_client_t) *ContextImplementation {
...
+        std.testing.allocator
+    else if (builtin.link_libc)
+        std.heap.c_allocator
+    else if (builtin.target.os.tag == .windows)
+        (struct {
+            var gpa = std.heap.HeapAllocator.init();
+        }).gpa.allocator()
+    else
+        @compileError("tb_client must be built with libc");
+    _ = allocator;
```
batiati: Pointless discard

batiati (changes requested):
> as usual naming things is hard :smile:

It's a bit more work refactoring (tests and clients that uses zig code), but I just had an idea: what if we dropped the `tb_client` prefix for all functions and only exposed them with this naming convention in the export section?

We can encapsulate all FFI specifics into an inner struct, so there is no naming collision. 

Like this:
```zig

const ffi = struct {
    pub fn c_init( .... ) tb_status_t {
        // Any FFI translation goes here:
    }

    pub fn c_init_echo( ... ) tb_status_t {
        // Any FFI translation goes here:
    }
}

comptime {
    if (builtin.link_libc) {
        @export(ffi.c_init, .{ .name = "tb_client_init", .linkage = .Strong });
        @export(ffi.c_init_echo, .{ .name = "tb_client_init_echo", .linkage = .Strong });
        @export(acquire_packet, .{ .name = "tb_client_acquire_packet", .linkage = .Strong });
        @export(release_packet, .{ .name = "tb_client_release_packet", .linkage = .Strong });
        @export(submit, .{ .name = "tb_client_submit", .linkage = .Strong });
        @export(deinit, .{ .name = "tb_client_deinit", .linkage = .Strong });
    }
}
```

PS: Don't worry about the failing CI, some checks fail when the PR comes from a branch in another repository.

rbino:
@batiati implemented your suggestion above, let me know if that's what you had in mind

src/clients/java/src/client.zig
```diff
@@ -88,25 +88,18 @@ const ReflectionHelper = struct {
...
 
-        var exception = env.newObject(
-            initialization_exception_class,
-            initialization_exception_ctor_id,
-            &[_]jui.jvalue{jui.jvalue.toJValue(@bitCast(jui.jint, @enumToInt(status)))},
-        ) catch {
-            // It's unexpected here: we did not initialize correctly or the JVM is out of memory.
-            @panic("JNI: Unexpected error creating a new exception.");
+        var buf: [512]u8 = undefined;
+        var msg = std.fmt.bufPrintZ(&buf, "{s}", .{@errorName(err)}) catch {
```
batiati: We can't do that, our `InitializationException` class expects an integer value on the constructor, that's why we don't use `env.throwNew` but instead we construct a new instance.

We should use the function `context_error_to_status` here, and keep passing the integer value to the Java side.

src/clients/c/tb_client.zig
```diff
@@ -52,131 +53,199 @@ fn client_to_context(tb_client: tb_client_t) *ContextImplementation {
...
-        addresses_len,
+        addresses,
         packets_count,
         on_completion_ctx,
         on_completion_fn,
     );
 }
 
-fn init(
+fn init_with_context(
```
batiati: I think it's ok to repeat the pattern `Context.init( ... ) catch |err| return context_error_to_status(err);`. We don't need an `init_with_context` function.

batiati (changes requested):
Thank you, @rbino
Just two minor adjustments, and we are good to merge 🎉

batiati (approved):
Grazie Riccardo 😎👨‍🍳🤌

PR #818, opened 2023-05-25
VSR: `commit_status`; `commit_dispatch()`

sentientwaffle:
## `commit_status`

Replace the `Replica.committing` boolean with a `commit_status` enum.
State sync will need to track the progress through the commit so that it can tell how to cancel the commit.
(e.g. Whether it can cancel the grid, or whether it needs to wait.)

## `commit_dispatch()`
Previously the commit "chain" of callbacks was like:

    fn foo()
      foo_start(foo_callback)
    fn foo_callback()
      bar_start(bar_callback)
    fn bar_callback()
      ... etc

But now it is:

    fn dispatch(next_state)
      switch (state)
        .foo -> foo_start(foo_callback)
        .bar -> bar_start(bar_callback)
        ...etc

    fn foo_callback()
      dispatch(.bar)
    fn bar_callback()
      dispatch(...etc)

This is preparation for state sync. State sync needs to be able to "interrupt" commits between steps.
How this will work: in `commit_dispatch()`, check whether we are syncing.
If we are, instead of continuing to the next step of commit, abort and signal to state sync that we have reached a stopping point.

Another approach would be to put a `if (syncing) { signal sync; return }` hook before every commit state transition, but that is more brittle & intrusive.

## Pre-merge checklist

Performance:

### DIFF
--- a/src/vsr/replica.zig
+++ b/src/vsr/replica.zig
@@ -2844,11 +2844,7 @@ pub fn ReplicaType(
             assert(self.commit_min <= self.op);
 
             if (self.status == .normal and self.primary()) {
-                if (self.pipeline.queue.prepare_queue.empty()) {
-                    self.commit_ops_done();
-                } else {
-                    self.commit_pipeline_next();
-                }
+                self.commit_pipeline_next();
             } else {
                 self.commit_journal_next();
             }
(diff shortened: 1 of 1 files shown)


src/vsr/replica.zig
```diff
@@ -41,6 +41,24 @@ pub const Status = enum {
     recovering_head,
 };
 
+const CommitStatus = enum {
```
matklad: I wonder if there's a better name, to reflect that we lineary move throughout the stages. `CommitProgress` or `CommitStage` perhaps? Or, to name the pattern "flow"

```
fn commit_flow(self: *Self, flow: CommitFlow) void {
```

Status also feels ok as well though.

src/vsr/replica.zig
```diff
@@ -580,12 +589,7 @@ pub fn ReplicaType(
                 }
             }
 
-            assert(
-                (self.status == .recovering and self.solo()) or
-                    self.status == .normal or
-                    self.status == .view_change or
-                    self.status == .recovering_head,
-            );
```
matklad, line 588: My original thinking here was to signify that we can be at any status after `open`. So perhaps we could add 

```
maybe(self.status == .normal);
maybe(self.status == .view_change);
maybe(self.status == .recovering_head);
```

matklad (approved):

matklad (dismissed):

matklad:
Love how this makes the overall flow clearer!

src/vsr/replica.zig
sentientwaffle: Good point! I went with `CommitStage`, and left the function name as-is (`commit_dispatch`).

matklad (approved):

PR #825, opened 2023-05-26
Fix next_version.sh

batiati:
Sort next version in numeric order.
This was causing the publish job to fail.

## Pre-merge checklist


eatonphil (approved):
Thanks Rafael!

PR #828, opened 2023-05-29
vsr: don't accept old SV when .recovering_head

matklad:
This fixes a correctness bug in the following scenario:

```
.normal
✓ 1 2 3
      ^

crash & restart

.recovering_head
✓ 1 x x
  ^ op uncertain

get an *outdated* start_view:

✓ 1 2
    ^

recover after start view:

.normal
✓ 1 2 x
    ^

view change: we send over 1&2 and _implicitly_ nack 3
```

That is, a `.recovering_head` replica accepts _any_ SV to set its head op. If the old SV is accepted, then replica's `op` post crash might be smaller than the one before it, resulting in the replica nacking subsequent ops.

What we want here is to guarantee that, when we are .recovering_head, that we only accept SV produced by the primary after our restart. To do that, include a nonce in a `request_start_view` message.

## Pre-merge checklist

Performance:


matklad:
I have yet to catch this one with a test though...

src/vsr/replica.zig
```diff
@@ -146,6 +146,8 @@ pub fn ReplicaType(
         /// header/message.
         quorum_nack_prepare: u8,
 
+        nonce: Nonce = 0,
```
sentientwaffle: Why is the default value needed?

src/vsr/replica.zig
```diff
@@ -4151,6 +4159,22 @@ pub fn ReplicaType(
                         log.warn("{}: on_{s}: misdirected message (self)", .{ self.replica, command });
                         return true;
                     }
+
+                    if (self.status == .recovering_head) {
+                        if (message.header.view > self.view or
+                            message.header.op >= self.op_checkpoint_next() or
```
sentientwaffle: I think `op >= op_checkpoint_next()` here should instead be `op >= op_checkpoint_trigger()` (which is a few ops after "next").

`op_checkpoint_trigger` is the highest op that could possibly be in our log.
So we need to guard against SVs with ops between `op_checkpoint_next` and `op_checkpoint_trigger`,

src/testing/cluster.zig
```diff
@@ -258,7 +258,8 @@ pub fn ClusterType(comptime StateMachineType: fn (comptime Storage: type, compti
 
             for (cluster.replicas) |_, replica_index| {
                 errdefer for (replicas[0..replica_index]) |*r| r.deinit(allocator);
-                try cluster.open_replica(@intCast(u8, replica_index), .{
+                const nonce = random.int(u128);
+                try cluster.open_replica(@intCast(u8, replica_index), nonce, .{
```
sentientwaffle: Why not generate the nonce within `open_replica`?
Or the replica could generate its own nonce during `Replica.open/init`?

matklad: To be deterministic: we either need to inject a nonce, or an rng, and injecting the nonce seems simpler. We can't use RNG we already have, becaues that's deterministically seeded from replica index, and woun't actualy generate unique nonces.

src/vsr/replica.zig
```diff
@@ -146,6 +146,10 @@ pub fn ReplicaType(
         /// header/message.
         quorum_nack_prepare: u8,
 
+        /// A globally unique integer generated by a crypto rng during replcia process startup.
```
sentientwaffle: s/replcia/replica

sentientwaffle (dismissed):
(other than the nit typo, this LGTM)

matklad:
@sentientwaffle PTAL, I added a test. The following diff shows how to introduce a bug:

```diff
diff --git a/src/vsr/replica.zig b/src/vsr/replica.zig
index 20f0ab31..f3d25991 100644
--- a/src/vsr/replica.zig
+++ b/src/vsr/replica.zig
@@ -1704,7 +1704,7 @@ pub fn ReplicaType(
             assert(message.header.op - message.header.commit <=
                 constants.pipeline_prepare_queue_max);
 
-            if (message.header.view == self.log_view and message.header.op <= self.op) {
+            if (message.header.view == self.log_view and message.header.op < self.op) {
                 // We were already in this view prior to receiving the SV.
                 assert(self.status == .normal or self.status == .recovering_head);
 
@@ -4177,7 +4177,7 @@ pub fn ReplicaType(
                                 self.replica,
                                 command,
                             });
-                            return true;
+                            // return true;
                         }
                     }
                 },
```

as discussed on slack, the `<= vs <` is probably another (liveness) bug, but I'll look into it separately.

sentientwaffle (approved):

PR #839, opened 2023-06-02
replica-test: improve readability

matklad:
This is command-query separation. In this tests, some statements are asserts over the state, and some are commands to change the state.

    try expectEqual(replicas.open(), .ok)

is both, and is too easy to confuse with a pure assertion.

## Pre-merge checklist

Performance:


matklad:
This is pure cosmetics, but I do miss every single one `.open` in tests on the first read.

sentientwaffle (dismissed):

sentientwaffle (approved):

PR #840, opened 2023-06-02
vsr: head can be recovered in an idle cluster 

matklad:
When we are in `.recovering_head`, the `op == message.header.op` case
isn't necessary outdated, it tells us that our provisional op is in fact
ok. This is especially important in idle cluster, where we might not get
a fresher op.

We _could_ return here if are not in .recovering_head, but let's rather
hit the rest of assertions.

### DIFF
--- a/src/vsr/replica_test.zig
+++ b/src/vsr/replica_test.zig
@@ -609,9 +609,10 @@ test "Cluster: repair: crash, corrupt committed pipeline op, repair it, view-cha
     try expectEqual(b1.status(), .recovering_head);
     t.run();
 
-    a0.stop();
     b1.pass_all(.R_, .bidirectional);
     b2.pass_all(.R_, .bidirectional);
+    a0.stop();
+    a0.drop_all(.R_, .outgoing);
     t.run();
 
     // The cluster is stuck trying to repair op=30 (requesting the prepare).
@@ -621,6 +622,7 @@ test "Cluster: repair: crash, corrupt committed pipeline op, repair it, view-cha
     try expectEqual(b1.op_head(), 30);
 
     // A0 provides prepare=30.
+    a0.pass_all(.R_, .outgoing);
     try a0.open();
     t.run();
     try expectEqual(t.replica(.R_).status(), .normal);
--- a/src/vsr/replica_test.zig
+++ b/src/vsr/replica_test.zig
@@ -264,6 +264,29 @@ test "Cluster: recovery: recovering_head, outdated start view" {
     try c.request(22, 22);
 }
 
+test "Cluster: recovery: recovering head: idle cluster" {
+    const t = try TestContext.init(.{ .replica_count = 3 });
+    defer t.deinit();
+
+    var c = t.clients(0, t.cluster.clients.len);
+    var b = t.replica(.B1);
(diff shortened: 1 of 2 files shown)


src/vsr/replica_test.zig
```diff
@@ -550,6 +550,8 @@ test "Cluster: repair: crash, corrupt committed pipeline op, repair it, view-cha
     t.run();
 
     a0.stop();
+    t.run();
```
matklad: Not sure what's the appropriate fix here tbh. Perhaps its `.stop()` which should be delivering/dropping any residual messages?

sentientwaffle: I don't understand what you're referencing. And what is this ~stop~ run for?

matklad: Sorry, I think I forgot to type the beginning of the comment here :)


So without this .run(), the test now fails. b1 *somehow* gets prepare for 30, so the cluster doesnt actually get stuck.

what I think is happening here is that, while we shut down a, we still get a repairing prepare from a in the packet simulator. So, when `pass_all` below, this prepare gets delivered, despite a being down (didn’t verify that this is exactly what’s happening).

🤔 hm, I guess I still don’t understand why the `<` change has this exact effect here, will revisit on Monday

matklad: Took a closer look here. It indeed is the case that, *after* we call `a0.stop()`, there could still be a pending `.prepare` message in-flight, and that a subsequent `t.run()` gets rid of it. 

Now, why _exactly_ the `<=` change causes the `.prepare` to be queued, is unclear to me. But it seems that it's basically a "timing" issue. 

`<=` doesn't cause any meaningful changes. In particular, the handling of `.recovering_head` for `op=30` is the same. What is different is the handling of an earlier `.start_view` for `op=0`. We either ignore this start view (with `<=`), or trigger a repair (with `<`). The repair is no-op, _except_ that it resets repair timer. And it seems that the timer ultimately affects whether a repairing prepare for `op=30` gets into the window around the crash of the primary. 

So, yeah, I think our testing infra (or maybe just a specific test) is buggy, in a sense that, when you do `a.stop()`, you can't be sure which messages would be delivered.

matklad: went for a slightly cleaner tacktical fix of cutting the network in addition to stoping the replica itself.

src/vsr/replica_test.zig
```diff
@@ -581,9 +604,10 @@ test "Cluster: repair: crash, corrupt committed pipeline op, repair it, view-cha
     try expectEqual(b1.status(), .recovering_head);
     t.run();
 
-    a0.stop();
     b1.pass_all(.R_, .bidirectional);
     b2.pass_all(.R_, .bidirectional);
+    a0.stop();
+    a0.drop_all(.R_, .outgoing);
```
sentientwaffle, line 615: Makes sense. In the future maybe we could fuzz different seeds for `replica_test` to catch these problems preemptively, so that unrelated changes don't trigger test failures due to permuting the rng.

sentientwaffle (dismissed):

sentientwaffle (approved):

PR #841, opened 2023-06-02
nuke pull-request template

matklad:
We now have continuous benchmarking setup, so it probably doesn't make sense to remind about this on every PR.

## Pre-merge checklist

* [X] I am very sure this PR could not affect performance and hopeful that this is the last time I need to check this checkbox :D


eatonphil (approved):

eatonphil:
One last time :)
image

PR #842, opened 2023-06-03
test: speed up unit tests

matklad:
`mem.set` is quite slow, removing it cuts replica_tests time from 30s to 15s.

My first thought here is to just remove the memset and rely on mmap to lazily initialize with zeros, _but_ it turns out that Zig intentionally doesn't provide this API in the allocator:

https://github.com/ziglang/zig/issues/2090#issuecomment-475641804

That's surprising, but also makes sense. In particular, in this case this pushed me to a better solution --- just randomly initializing the sectors on the first access. This has the same perf bonus, while also increasing the coverage (as we now return random data, rather than just zero).

Technically, this is not initializing, as subsequent init reads of the same sector can return different data, but I think we shouldn't rely on the fact that un-initialized data is frozen. This is certainly not true for memory (MADV_FREE), and I think it could reasonably be true for disks as well.

### DIFF
--- a/src/testing/storage.zig
+++ b/src/testing/storage.zig
@@ -158,8 +158,6 @@ pub const Storage = struct {
         const sector_count = @divExact(size, constants.sector_size);
         const memory = try allocator.allocAdvanced(u8, constants.sector_size, size, .exact);
         errdefer allocator.free(memory);
-        // TODO: random data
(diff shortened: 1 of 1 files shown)


sentientwaffle (approved):
Excellent! And this should improve vopr performance, too.

PR #843, opened 2023-06-03
vopr: catch replica-local bugs earlier

matklad:
State checker catches bug once they break cluster-wide linearizability. However not every correctness bug immediately leads to cluster divergance.

For example, if a replica forgets its prepare_ok, this might lead to erroneous truncation during a view change, but it is more likely to be just repaired over.

In this PR, we enhance the state checker to take a closer look at individual state of each replica, and prevent replicas from rolling back their commitments.

Note that we can't track this purely with asserts inside the replica, as the invariant might be broken across restarts.

I've verified that it makes discovering problem in

https://github.com/tigerbeetledb/tigerbeetle/pull/828

more likely.

### DIFF
--- a/src/testing/cluster.zig
+++ b/src/testing/cluster.zig
@@ -393,6 +393,7 @@ pub fn ClusterType(comptime StateMachineType: fn (comptime Storage: type, compti
             replica.on_compact = on_replica_compact;
             replica.on_checkpoint_start = on_replica_checkpoint_start;
             replica.on_checkpoint_done = on_replica_checkpoint_done;
+            replica.on_message_sent = on_replica_message_sent;
             cluster.network.link(replica.message_bus.process, &replica.message_bus);
         }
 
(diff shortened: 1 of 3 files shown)


sentientwaffle (approved):

matklad:
Got this crashing on main if I increase replica crash prob. Investigating

```
11:00:24|~/p/tigerbeetle|main⚡*
λ git rev-parse HEAD 
f144807bd9be3290e36bfcb9143ae631d702a771

11:00:33|~/p/tigerbeetle|main⚡*
λ git diff
diff --git a/src/simulator.zig b/src/simulator.zig
index 8477a70d..ffbbc10c 100644
--- a/src/simulator.zig
+++ b/src/simulator.zig
@@ -158,9 +158,9 @@ pub fn main() !void {
         .cluster = cluster_options,
         .workload = workload_options,
         // TODO Swarm testing: Test long+few crashes and short+many crashes separately.
-        .replica_crash_probability = 0.00002,
+        .replica_crash_probability = 0.002,
         .replica_crash_stability = random.uintLessThan(u32, 1_000),
-        .replica_restart_probability = 0.0002,
+        .replica_restart_probability = 0.02,
         .replica_restart_stability = random.uintLessThan(u32, 1_000),
         .requests_max = constants.journal_slot_count * 3,
         .request_probability = 1 + random.uintLessThan(u8, 99),

11:00:36|~/p/tigerbeetle|main⚡*
λ ./zig/zig build simulator_run -Dsimulator-log=short -- 8419954452427260277
```

PR #845, opened 2023-06-05
Use a union for holding prefetch contexts.

batiati:
Implements this TODO note:

https://github.com/tigerbeetledb/tigerbeetle/blob/b5fb287e2cdf82cb7d12ad372e40b77395237f10/src/state_machine.zig#L254-L260

- Removes multiple fields for each specialized PrefetchContext in favor of an external union holding all possible contexts in the same memory space.
We must use an external union because even untagged unions had an implicit tag in Debug mode for raising undefined behavior. Packed unions would be the best choice, but they require all fields to be packed types.

- Adds a helper function to retrieve the parent pointer, since Zig 0.9.1 doesn't support fieldParentPtr on unions.

### DIFF
--- a/src/state_machine.zig
+++ b/src/state_machine.zig
@@ -245,19 +245,40 @@ pub fn StateMachineType(
             cache_entries_posted: u32,
         };
 
+        // Since only one context can be used at time,
+        // we use an union to save memory.
+        const PrefetchContext = extern union {
+            accounts_immutable: AccountsImmutableGroove.PrefetchContext,
+            accounts_mutable: AccountsMutableGroove.PrefetchContext,
+            transfers: TransfersGroove.PrefetchContext,
+            posted: PostedGroove.PrefetchContext,
+
+            // TODO(Zig): No need for this function once Zig is upgraded
+            // and @fieldParentPtr() can be used for unions.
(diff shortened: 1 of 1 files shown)


src/state_machine.zig
```diff
@@ -245,19 +245,39 @@ pub fn StateMachineType(
...
         prefetch_input: ?[]align(16) const u8 = null,
         prefetch_callback: ?fn (*StateMachine) void = null,
-        // TODO(ifreund): use a union for these to save memory, likely an extern union
-        // so that we can safetly @ptrCast() until @fieldParentPtr() is implemented
-        // for unions. See: https://github.com/ziglang/zig/issues/6611
-        prefetch_accounts_immutable_context: AccountsImmutableGroove.PrefetchContext = undefined,
-        prefetch_accounts_mutable_context: AccountsMutableGroove.PrefetchContext = undefined,
-        prefetch_transfers_context: TransfersGroove.PrefetchContext = undefined,
-        prefetch_posted_context: PostedGroove.PrefetchContext = undefined,
+        prefetch_context: PrefetchContext = undefined,
```
matklad, line 280: Could we also set `prefetch_context` to `undefined` in `prefetch_finished`?

matklad (approved):

PR #848, opened 2023-06-05
Move integration tests until after merge to main

eatonphil:


eatonphil:
I'm not sure if this is a great thing to jump into without more cleanup work around releases that fail. Do we delete the release or mark it as invalid? Closing until we've thought through it a bit more.

PR #834, opened 2023-06-01
VSR: State Sync (Part 1/?)

sentientwaffle:
State sync synchronizes the state of a lagging/divergent replica with the healthy cluster.

Check out `docs/internals/sync.md` for details.

The commit history is broken up into logical changes, but I wouldn't necessarily recommend reviewing them in that order, since the changes are bottom-up until the "Implement sync protocol" commit.

## Overview

0. Sync is needed. (`open`, `on_repair_sync_timeout`, `on_start_view`, `jump_sync_target`)
1. Trigger sync. (`sync_start_from_recovering`, `sync_start_from_committing`)
2. Wait for non-grid commit operation to finish. (`sync_cancel_commit_callback`)
3. Wait for grid IO to finish. (`Grid.cancel`, `sync_cancel_grid_callback`, `{StateMachine,Forest,Tree,etc}.reset`)
4. Wait for a usable sync target to arrive. (`jump_sync_target`, `command=commit`, `command=ping`)
5. Update superblock headers. (`sync_superblock_start`, `sync_superblock_start_callback`, `SuperBlock.sync_start`)
6. Request superblock trailers. (`on_request_sync_manifest`, `on_request_sync_free_set`, `on_request_sync_client_sessions`, `on_sync_manifest`, `on_sync_free_set`, `on_sync_client_sessions`, `on_sync_message_timeout`, `SuperBlock.Manifest.clear`, `SuperBlock.FreeSet.clear`, `SuperBlock.ClientSessions.clear`)
7. Request and write manifest log blocks. (`sync_request_manifest_logs`, `sync_request_manifest_logs_callback`, grid repair).
8. Request and write table blocks. (This step is not implemented yet. Currently table blocks are repair lazily, using grid repair.)
9. Update superblock headers and trailers. (`sync_superblock_done`, `sync_superblock_done_callback`, `SuperBlock.sync_done`)
10. Done.

## Next Steps

(In no particular order:)

- Async checkpoints, to fix a known availability bug (currently worked around in `Replica.ignore_request_message`). See below.
- Proactive table (and maybe client replies) sync. (This will allow the currently disabled checks in
  the `StorageChecker` and `Manifest` to be re-enabled.)
- Unified manifest: Don't partition the LSM manifest log by tree.
  (Not directly related to state sync except to make the "sync manifest" step more efficient.)
- Manifest log prefetching: Prefetch manifest log blocks during `StateMachine.open()`.
  (Especially important when coupled with the unified manifest. This optimizes the "sync manifest
[...]

PR #855, opened 2023-06-06
Use a union for holding lookup contexts.

batiati:
Pretty much the same as [#845 Use a union for holding prefetch contexts](https://github.com/tigerbeetledb/tigerbeetle/pull/845), sorry for making it two PRs 😅.

Implements a TODO note:

https://github.com/tigerbeetledb/tigerbeetle/blob/c612e90d0ad47e443425f7b5002c50c4a4a3d0a5/src/lsm/groove.zig#L675-L679

- Removes individual fields for each LookupContext in favor of an external union, holding them all in the same memory space.

- Adds a helper function to retrieve the parent pointer, since Zig 0.9.1 doesn't support fieldParentPtr on unions

### DIFF
--- a/src/lsm/groove.zig
+++ b/src/lsm/groove.zig
@@ -671,12 +671,31 @@ pub fn GrooveType(
         };
 
         pub const PrefetchWorker = struct {
+
+            // Since lookup contexts are used one at a time, it's safe to access
+            // the union's fields and reuse the same memory for all context instances.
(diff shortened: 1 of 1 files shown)


matklad (approved):
>sorry for making it two PRs sweat_smile.

Naah, smaller, separable PRs are better in general (looks at #834 :-) )

PR #863, opened 2023-06-08
[stdx][perf] div_ceil alternative implementation.

ttsugriy:
In C++ this leads to a more [efficient assembly](https://godbolt.org/z/f35PrfP76) and [~30% faster execution](https://quick-bench.com/q/om3CRcHNHO7hQld5mZaJRxPoM4w).

The disadvantage of this approach is that unlike previous implementation it can overflow if `nominator - 1 + denominator` is greater than the largest possible int value. Whether it matters or not for your workloads is up to you, so feel free to
- add additional parameter to control which version to use
- add extra asserts detecting this overflow
- reject the PR

### DIFF
--- a/src/stdx.zig
+++ b/src/stdx.zig
@@ -21,8 +21,7 @@ pub inline fn div_ceil(numerator: anytype, denominator: anytype) @TypeOf(numerat
 
     assert(denominator > 0);
 
-    if (numerator == 0) return 0;
-    return @divFloor(numerator - 1, denominator) + 1;
+    return @divFloor(numerator - 1 + denominator, denominator);
 }
 
 test "div_ceil" {


eatonphil:
Thanks @ttsugriy ! One thing folks do when contributing is run our benchmark script before and after and post the output in the PR. Since you're making a performance suggestion would you mind doing so?

https://github.com/tigerbeetledb/tigerbeetle#benchmarks

ttsugriy:
> Thanks @ttsugriy ! One thing folks do when contributing is run our benchmark script before and after and post the output in the PR. Since you're making a performance suggestion would you mind doing so?
> 
> https://github.com/tigerbeetledb/tigerbeetle#benchmarks

I'd be happy to, but, unfortunately, it looks like the overflow I've mentioned in the summary does affect some of the tests and benchmarks, which fail with

```bash
% scripts/benchmark.sh
Code Generation [7059/10374] .vsr.lsm.binary_search.binary_search_values_raw...Formatting replica 0...662] io.darwin.IO.submit... locator... tBytes... .. from_.scripts/benchmark.sh: line 55: 44464 Abort trap: 6           ./tigerbeetle format --cluster=0 --replica="$I" --replica-count=1 "$FILE" > benchmark.log 2>&1

Error running benchmark, here are more details (from benchmark.log):

info(io): creating "0_0.tigerbeetle.benchmark"...
info(io): allocating 1.13494873046875GiB...
thread 15801951 panic: integer overflow
tigerbeetle/src/vsr/superblock.zig:0:0: 0x102295403 in .vsr.vsr.superblock.SuperBlockType(.vsr.storage.Storage).init (tigerbeetle)
tigerbeetle/src/tigerbeetle/main.zig:110:45: 0x10228a0c7 in main (tigerbeetle)
        var superblock = try SuperBlock.init(
                                            ^
tigerbeetle/zig/lib/std/start.zig:561:37: 0x10228fec3 in std.start.main (tigerbeetle)
                                    ^
???:?:?: 0x184d8ff27 in ??? (???)
???:?:?: 0x1c0c7fffffffffff in ??? (???)
Stopping replica 0...
scripts/benchmark.sh: line 51: kill: %1: no such job
```

So I'll close this PR, but if at any point this code gets into the hot path, consider using proposed approach :)

PR #867, opened 2023-06-09
LSM,VSR: Backport std HashMap fix

sentientwaffle:
In zig 0.9.1 (and zig master, as of writing this) `fetchRemove` does not increment `available`.
So it leaks slots.
Currently we only use `fetchRemove` in the mutable table and the client sessions.

Note that the VOPR didn't find this bug in client sessions because evictions are not tested.
It probably didn't find it in the mutable table because we often use `createRetainingCapacity()`, which restores the leaked slots.

Backport the fix: https://github.com/ziglang/zig/pull/15989.

### DIFF
--- a/src/lsm/table_mutable.zig
+++ b/src/lsm/table_mutable.zig
@@ -22,7 +22,13 @@ pub fn TableMutableType(comptime Table: type, comptime tree_name: [:0]const u8) 
         const TableMutable = @This();
 
         const load_factor = 50;
-        const Values = std.HashMapUnmanaged(Value, void, Table.HashMapContextValue, load_factor);
+        /// TODO(Zig): Import from std. See https://github.com/ziglang/zig/pull/15989.
+        const Values = @import("../hash_map.zig").HashMapUnmanaged(
+            Value,
+            void,
+            Table.HashMapContextValue,
+            load_factor,
+        );
 
         pub const ValuesCache = SetAssociativeCache(
             Key,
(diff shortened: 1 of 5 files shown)


src/lsm/table_mutable.zig
```diff
@@ -22,7 +22,13 @@ pub fn TableMutableType(comptime Table: type, comptime tree_name: [:0]const u8)
         const TableMutable = @This();
 
         const load_factor = 50;
-        const Values = std.HashMapUnmanaged(Value, void, Table.HashMapContextValue, load_factor);
+        /// TODO(Zig): Import from std. See https://github.com/ziglang/zig/pull/15989.
+        const Values = @import("../hash_map.zig").HashMapUnmanaged(
```
jorangreef: We could import this in `stdx`, centralize the TODO there, and then use `stdx.HashMapUnmanaged` here and in `superblock_client_sessions.zig`.

jorangreef (commented):
Do we need to update a few other uses of std HashMap?

For example, `(posted_)groove.zig`, `superblock_manifest.zig`, `auditor.zig` and `message_bus.zig`?

sentientwaffle:
> For example, (posted_)groove.zig, superblock_manifest.zig, auditor.zig and message_bus.zig?

The backported hashmap is only needed when we use `fetchRemove`. So only the mutable table's value cache and the client sessions hashmap need to be updated.

jorangreef (approved):

PR #875, opened 2023-06-13
LSM Compaction: Use stable ManifestLevel TableInfo references to reduce ManifestLevel queries

chaitanyabhandari:
### Reduce queries to ManifestLevel during compaction

Specifically, this change have reduces the number of queries to the ManifesLevel at two places:

1.  [CompactionType.iterator_check](https://github.com/tigerbeetledb/tigerbeetle/blob/6c1f34bd69b07419542c9214159add629ea17243/src/lsm/compaction.zig#L403-L408), invokes [LevelDataIteratorType.next](https://github.com/tigerbeetledb/tigerbeetle/blob/13cc39fadd226864910c581642edb388bf2a5de7/src/lsm/level_data_iterator.zig#L121-L139), which further invokes [LevelndexIteratorType.next](https://github.com/tigerbeetledb/tigerbeetle/blob/13cc39fadd226864910c581642edb388bf2a5de7/src/lsm/level_index_iterator.zig#L86-L122). [LevelndexIteratorType.next](https://github.com/tigerbeetledb/tigerbeetle/blob/13cc39fadd226864910c581642edb388bf2a5de7/src/lsm/level_index_iterator.zig#L86-L122) again iterates through the tables in Level B which intersect with the range returned by [ManifestType.compaction_range](https://github.com/tigerbeetledb/tigerbeetle/blob/6c1f34bd69b07419542c9214159add629ea17243/src/lsm/manifest.zig#L487-L546). Now, instead of querying for the "next intersecting table" using [LevelndexIteratorType.next](https://github.com/tigerbeetledb/tigerbeetle/blob/13cc39fadd226864910c581642edb388bf2a5de7/src/lsm/level_index_iterator.zig#L86-L122), we're directly using pointers to the intersecting tables in the ManifestLevel in [LevelBDataIterator.next](https://github.com/tigerbeetledb/tigerbeetle/blob/1b64cac51bc103db6c673ad59dc69d7c12f04e6e/src/lsm/compaction.zig#L861).
2.  [CompactionType.apply_to_manifest](https://github.com/tigerbeetledb/tigerbeetle/blob/13cc39fadd226864910c581642edb388bf2a5de7/src/lsm/compaction.zig#L817) invokes [Compaction.update_table](https://github.com/tigerbeetledb/tigerbeetle/blob/13cc39fadd226864910c581642edb388bf2a5de7/src/lsm/manifest.zig#L213), which further invokes calls [ManifestLevel.set_snapshot_max](https://github.com/tigerbeetledb/tigerbeetle/blob/13cc39fadd226864910c581642edb388bf2a5de7/src/lsm/manifest_level.zig#L89). [ManifestLevel.set_snapshot_max](https://github.com/tigerbeetledb/tigerbeetle/blob/13cc39fadd226864910c581642edb388bf2a5de7/src/lsm/manifest_level.zig#L89) performs another query to find the table to be updated in the ManifestLevel, and then updates it. In the new version, we're directly using pointers to the intersecting tables in the ManifestLevel and updating it using [ManifestLevel.set_snapshot_max_stable_ptr](https://github.com/tigerbeetledb/tigerbeetle/blob/1b64cac51bc103db6c673ad59dc69d7c12f04e6e/src/lsm/manifest_level.zig#L104) (I've currently retained both functions, we can make this the default one too).

### Miscellaneous Changes

1. Implements [this](https://github.com/tigerbeetledb/tigerbeetle/blob/6c1f34bd69b07419542c9214159add629ea17243/src/lsm/compaction.zig#L775) TODO for releasing grid blocks associated with Level A. This is now done [here](https://github.com/tigerbeetledb/tigerbeetle/blob/1b64cac51bc103db6c673ad59dc69d7c12f04e6e/src/lsm/compaction.zig#L364), right after we copy the index block for the optimal compaction table in Level A.
2. Moves the core logic for [Manifest.compaction_table](https://github.com/tigerbeetledb/tigerbeetle/blob/6c1f34bd69b07419542c9214159add629ea17243/src/lsm/manifest.zig#L430), [Manifest.compaction_range](https://github.com/tigerbeetledb/tigerbeetle/blob/6c1f34bd69b07419542c9214159add629ea17243/src/lsm/manifest.zig#L500), [Manifest.next_table](https://github.com/tigerbeetledb/tigerbeetle/blob/6c1f34bd69b07419542c9214159add629ea17243/src/lsm/manifest.zig#L375), and [Manifest.compaction_must_drop_tombstones](https://github.com/tigerbeetledb/tigerbeetle/blob/6c1f34bd69b07419542c9214159add629ea17243/src/lsm/manifest.zig#LL549C16-L549C47) into ManifestLevel (while still retaining the proxy functions in Manifest for use by Compaction). This is because the logic in these functions strictly related to the in-memory ManifestLevel. Moving this logic into ManifestLevel would allow us to test & benchmark all this functionality more easily (without involving ManifestLog, which accompanies by Manifest).
3. Removes [Table.index_block_address](https://github.com/tigerbeetledb/tigerbeetle/blob/6c1f34bd69b07419542c9214159add629ea17243/src/lsm/table.zig#L710), since it is the same as [Table.block_address](https://github.com/tigerbeetledb/tigerbeetle/blob/6c1f34bd69b07419542c9214159add629ea17243/src/lsm/table.zig#L870).
4. Small changes to [EnvironmentType](https://github.com/tigerbeetledb/tigerbeetle/blob/1b64cac51bc103db6c673ad59dc69d7c12f04e6e/src/lsm/manifest_level_fuzz.zig#L229) in manifest_level_fuzz.zig to make it more amenable for use from other test/benchmarking scripts. 

### Tests Conducted

1. ./scripts/build.sh fuzz_lsm_tree ✅
2. ./scripts/build.sh fuzz_lsm_forest ✅


src/lsm/manifest_level_fuzz.zig
```diff
@@ -224,47 +226,50 @@ const GenerateContext = struct {
     }
 };
 
-fn EnvironmentType(comptime table_count_max: u32, comptime node_size: u32) type {
+pub fn EnvironmentType(comptime table_count_max: u32, comptime node_size: u32) type {
```
sentientwaffle, line 229: Why change visibility?

src/lsm/compaction.zig
```diff
@@ -825,19 +832,176 @@ pub fn CompactionType(
...
+        const LevelBDataIterator = @This();
+
+        const Grid = GridType(Storage);
+        const BlockPtr = Grid.BlockPtr;
+        const BlockPtrConst = Grid.BlockPtrConst;
+        const Manifest = ManifestType(Table, Storage);
+        const TableInfo = Manifest.TableInfo;
+        const TableDataIterator = TableDataIteratorType(Storage);
+
+        pub const Context = struct { grid: *Grid, level: u8, snapshot: u64, tables: std.BoundedArray(*TableInfo, constants.lsm_growth_factor) };
```
sentientwaffle: Split long line

src/lsm/manifest_level.zig
```diff
@@ -388,6 +419,164 @@ pub fn ManifestLevelType(
...
+        ///   range.key_min = min(a.key_min, b.key_min)
+        ///   range.key_max = max(a.key_max, b.key_max)
+        ///
+        /// This last invariant is critical to ensuring that tombstones are dropped correctly.
+        pub fn compaction_range(
+            level: *const Self,
+            key_min: Key,
+            key_max: Key,
+        ) CompactionRange {
+            var range = CompactionRange{ .table_count = 1, .key_min = key_min, .key_max = key_max, .tables = .{ .buffer = undefined } };
```
sentientwaffle: Line too long, please wrap.

src/lsm/manifest_level.zig
```diff
@@ -388,6 +419,164 @@ pub fn ManifestLevelType(
...
+                }
+                // This const cast is safe as we know that the memory pointed to is in fact
+                // mutable. That is, the table is not in the .text or .rodata section.
+                // We set the capacity of range.tables to lsm_growth_factor, since that is the
+                // maximum number of tables that can intersect with a table in Level A. We needn't
+                // add any more table pointers to this array, since this table won't get selected
+                // if it intersects with > lsm_growth_factor number of tables.
+                if (range.tables.len < range.tables.capacity()) {
+                    range.tables.appendAssumeCapacity(@intToPtr(*TableInfo, @ptrToInt(table)));
+                }
```
sentientwaffle: Under what circumstances would we ever append more than `lsm_growth_factor` tables to `range.tables`?

src/lsm/manifest_level.zig
```diff
@@ -388,6 +419,164 @@ pub fn ManifestLevelType(
...
+        /// Returns the next table in the range, after `key_exclusive` if provided.
+        ///
+        /// * The table returned is visible to `snapshot`.
+        pub fn next_table(
+            self: *const Self,
+            snapshot: u64,
+            key_min: Key,
+            key_max: Key,
+            key_exclusive: ?Key,
+            direction: Direction,
```
sentientwaffle, line 472:
```suggestion
            parameters: struct {
            snapshot: u64,
            key_min: Key,
            key_max: Key,
            key_exclusive: ?Key,
            direction: Direction,
            },
```
(Since there are so many parameters, naming them guards against mixups & makes the callsite easier to read.)

src/lsm/manifest_level.zig
```diff
@@ -388,6 +419,164 @@ pub fn ManifestLevelType(
             }
             return false;
         }
+
+        /// Given two levels, returns the most optimal compaction table in the
+        /// level for which this function is invoked. The optimal table is one
+        /// that overlaps with the least number of tables in the second level.
+        /// This function exits early if it finds a table that doesn't
+        ///  overlap with any tables in the second level.
```
sentientwaffle:
```suggestion
        /// overlap with any tables in the second level.
```

src/lsm/manifest_level.zig
```diff
@@ -81,6 +97,22 @@ pub fn ManifestLevelType(
             assert(level.keys.len() == level.tables.len());
         }
 
+        /// Set snapshot_max for the given table in the ManifestLevel.
+        /// This function assumes that the argument `table` is a stable pointer to
+        /// a table within the segmented array.
+        /// * Asserts that the table currently has snapshot_max of math.maxInt(u64).
+        pub fn set_snapshot_max_stable_ptr(level: *Self, snapshot: u64, table: *TableInfo) void {
```
sentientwaffle: Can you add a `constants.verify`-gated verification that actually checks that `table` really is a valid pointer into the manifest level?

Also, are there any other functions in the manifest level that could benefit from the same guard (or its inverse).

chaitanyabhandari:
### Output of scripts/benchmark.sh

#### Before (on branch _main_)

```python
1223 batches in 109.32 s
load offered = 1000000 tx/s
load accepted = 91471 tx/s
batch latency p00 = 1 ms
batch latency p10 = 22 ms
batch latency p20 = 24 ms
batch latency p30 = 24 ms
batch latency p40 = 25 ms
batch latency p50 = 26 ms
batch latency p60 = 27 ms
batch latency p70 = 28 ms
batch latency p80 = 28 ms
batch latency p90 = 30 ms
batch latency p100 = 4505 ms
transfer latency p00 = 1 ms
transfer latency p10 = 4191 ms
transfer latency p20 = 8537 ms
transfer latency p30 = 15783 ms
transfer latency p40 = 25487 ms
transfer latency p50 = 37590 ms
transfer latency p60 = 49517 ms
transfer latency p70 = 57964 ms
transfer latency p80 = 70396 ms
transfer latency p90 = 83550 ms
transfer latency p100 = 99337 ms
```

#### After (on branch _cbb-lsm-manifest-iteration_)

```python
~/Desktop/tigerbeetle_summer23/tigerbeetle cbb-lsm-manifest-iteration* ⇡ ❯ sudo scripts/benchmark.sh                  

1223 batches in 107.07 s
load offered = 1000000 tx/s
load accepted = 93399 tx/s
batch latency p00 = 2 ms
batch latency p10 = 22 ms
batch latency p20 = 23 ms
batch latency p30 = 24 ms
batch latency p40 = 25 ms
batch latency p50 = 25 ms
batch latency p60 = 26 ms
batch latency p70 = 27 ms
batch latency p80 = 28 ms
batch latency p90 = 30 ms
batch latency p100 = 4192 ms
transfer latency p00 = 2 ms
transfer latency p10 = 4122 ms
transfer latency p20 = 8632 ms
transfer latency p30 = 15791 ms
transfer latency p40 = 25148 ms
transfer latency p50 = 36490 ms
transfer latency p60 = 48335 ms
transfer latency p70 = 56904 ms
transfer latency p80 = 69647 ms
transfer latency p90 = 82354 ms
transfer latency p100 = 97078 ms
```

src/lsm/manifest_level.zig
```diff
@@ -38,6 +38,22 @@ pub fn ManifestLevelType(
...
+        pub const CompactionRange = struct {
+            /// The total number of tables in the compaction across both levels, always at least 1.
+            table_count: usize,
+            /// The minimum key across both levels.
+            key_min: Key,
+            /// The maximum key across both levels.
+            key_max: Key,
+            // References to tables in level B that intersect with the chosen table in level A.
+            tables: std.BoundedArray(*TableInfo, constants.lsm_growth_factor),
+        };
```
sentientwaffle: Not 100% this is a good idea (@jorangreef thoughts?) but:

- Pull `CompactionRange` and `CompactionTableRange` out into a separate file (`compaction_range.zig`).
- Move `ManifestLevel.compaction_table` and `ManifestLevel.compaction_range` to  `CompactionTableRange.init` and `CompactionRange.init` respectively.

Motivation is: I agree that computing compaction tables/ranges isn't the `Manifest`'s responsibility (it has nothing to do with the manifest log). But also, it maybe isn't the manifest level's responsibility either.. I picture `ManifestLevel` as a data structure that doesn't really care about or depend on compaction. (Similar to how `Compaction` is separate from `Tree`.)

src/lsm/compaction.zig
```diff
@@ -411,29 +404,27 @@ pub fn CompactionType(
...
+            compaction.release_table_blocks(index_block);
+        }
 
-            // Tables that we've compacted should become invisible at the end of this half-bar.
-            compaction.manifest_entries.appendAssumeCapacity(.{
-                .operation = .update_in_level_b,
-                .table = table_info,
-            });
-
+        fn release_table_blocks(compaction: *Compaction, index_block: BlockPtrConst) void {
```
sentientwaffle, line 430: Add a TODO that to support LSM snapshots we need to only remove the blocks if they are actually invisible.

src/lsm/compaction.zig
```diff
@@ -814,9 +794,36 @@ pub fn CompactionType(
...
+                    .immutable => {},
+                    .disk => |table_info| {
+                        manifest.update_table(level_b - 1, snapshot_max, null, table_info);
+                    },
+                }
+                for (compaction.context.range_b.tables.slice()) |table| {
+                    manifest.update_table(level_b, snapshot_max, null, table);
+                }
+            }
+        }
```
```suggestion
        }

```

sentientwaffle: I think this function should just be inlined into `apply_to_manifest`.

src/lsm/compaction.zig
```diff
@@ -814,9 +794,36 @@ pub fn CompactionType(
...
+        // Updates snapshot_max for the optimal compaction table in level A, as well as the
+        // tables in level B that intersect with it. Updating snapshot_max marks the table as
+        // invisible, these tables are then deleted from the level using remove_invisible_tables.
+        // * Assumption: This function directly uses pointers to the tables in the ManifestLevel
+        // to perform updates. This avoids extra queries to the ManifestLevel for these tables.
+        pub fn update_snapshot_max(compaction: *Compaction) void {
+            const snapshot_max = snapshot_max_for_table_input(compaction.context.op_min);
+            const can_move_table =
+                compaction.context.table_info_a == .disk and
+                compaction.context.range_b.table_count == 1;
```
sentientwaffle: Maybe store this on the compaction during `start` instead of recomputing, to ensure they don't get out of sync. (I thought we did that already but don't see it.)

src/lsm/compaction.zig
```diff
@@ -825,19 +832,176 @@ pub fn CompactionType(
...
+// tables to be iterated over are passed via the `tables` field in the context,
+// in the start() function.
+// * Assumption: This iterator works under the assumption that the references to
+// the tables are stable. For references fetched from the ManifestLevel, this
+// assumption only holds true till move_table & insert_table are called as a part
+// of apply_to_manifest. This is because move_table & insert_table are both operations
+// that insert/remove tables from the ManifestLevel segmented arrays, which could cause
+// rearrangements in the memory layout of the arrays. These rearrangements could
+// cause our table references to get invalidated.
+pub fn LevelBDataIteratorType(comptime Table: type, comptime Storage: type) type {
```
sentientwaffle: Why add `LevelBDataIteratorType` to `compaction.zig` instead of replacing `level_data_iterator.zig`?
If we will still need the old iterator for something (queries?) I'm not convinced that 2% throughput is worth the complexity of an extra iterator implementation.

src/lsm/compaction.zig
```diff
@@ -411,29 +404,27 @@ pub fn CompactionType(
...
-            compaction.manifest_entries.appendAssumeCapacity(.{
-                .operation = .update_in_level_b,
-                .table = table_info,
-            });
-
+        fn release_table_blocks(compaction: *Compaction, index_block: BlockPtrConst) void {
             // Release the table's block addresses in the Grid as it will be made invisible.
-            // This is safe; iterator_b makes a copy of the block before calling us.
+            // This is safe; iterator_b makes a copy of the block before calling us, and
+            // iterator_a holds a copy for table A's index block.
```
```suggestion
            // iterator_a owns a copy of table A's index block.
```

sentientwaffle (commented):
First pass; will give it a more detailed read-through tomorrow.

src/lsm/manifest_level_fuzz.zig
chaitanyabhandari: This is related to the the 4th miscellaneous change:

Small changes to [EnvironmentType](https://github.com/tigerbeetledb/tigerbeetle/blob/1b64cac51bc103db6c673ad59dc69d7c12f04e6e/src/lsm/manifest_level_fuzz.zig#L229) in manifest_level_fuzz.zig to make it more amenable for use from other test/benchmarking scripts.

I am working on some ManifestLevel microbenchmarks where I want to use this!

src/lsm/manifest.zig
```diff
@@ -208,18 +209,27 @@ pub fn ManifestType(comptime Table: type, comptime Storage: type) type {
             }
         }
 
-        /// Updates the snapshot_max on the provide table for the given level.
-        /// The table provided is mutable to allow its snapshot_max to be updated.
+        /// Updates the snapshot_max on the provided table for the given level.
+        /// The table provided could either a pointer to the table in the
```
kprotty: > could either *be*

src/lsm/manifest.zig
```diff
@@ -208,18 +209,27 @@ pub fn ManifestType(comptime Table: type, comptime Storage: type) type {
...
+        /// ManifestLevel or a pointer to a copy of that table.
         pub fn update_table(
             manifest: *Manifest,
             level: u8,
             snapshot: u64,
-            table: *TableInfo,
+            table_copy: ?*TableInfo,
+            table_manifest_level: ?*TableInfo,
```
kprotty: Maybe use a union or enum here:
```zig
table_ref: union(enum){ copy: *TableInfo, from_level: *TableInfo },
// or
table_type: enum{ copy, from_level },
table: *TableInfo,
```

src/lsm/compaction.zig
chaitanyabhandari: I added it as a separate iterator since I was under the impression that Rafael is going to consume LevelDataIterator to implement scans. @batiati: could you please confirm?

src/lsm/manifest_level.zig
chaitanyabhandari: As per my understanding: since [ManifestLevel.compaction_range](https://github.com/tigerbeetledb/tigerbeetle/blob/1b64cac51bc103db6c673ad59dc69d7c12f04e6e/src/lsm/manifest_level.zig#L534) iterates through all tables in Level B that intersect with a given table in Level A. That count could be more that `lsm_growth_factor` (even though we wouldn't end up choosing that as the optimal compaction table), right? I might be missing some context here. 😄 

Also, if my understanding of the above is correct, then I could add a small change that exits this function as soon as we find that a table in Level A intersects with more than `lsm_growth_factor` number of tables.

src/lsm/manifest_level.zig
chaitanyabhandari: Agreed, thank you for the input! Adding this to Manifest.next_table & ManifestLevel.next_table.

src/lsm/compaction.zig
chaitanyabhandari: Oops, this was a typo, I meant _**"compaction.index_block_a holds a copy for table A's index block"**_.

src/lsm/compaction.zig
chaitanyabhandari: Ah, yes. That'd be cleaner. I'm adding this as a property in start() so we can use it as _compaction.can_move_table_.

src/lsm/manifest.zig
```diff
@@ -370,8 +380,6 @@ pub fn ManifestType(comptime Table: type, comptime Storage: type) type {
         }
 
         /// Returns the next table in the range, after `key_exclusive` if provided.
-        ///
-        /// * The table returned is visible to `snapshot`.
```
kprotty: `ManifestLevel.next_table` says it's visible to the snapshot. Is it no longer the case for this function?

src/lsm/manifest.zig
```diff
@@ -383,53 +391,22 @@ pub fn ManifestType(comptime Table: type, comptime Storage: type) type {
...
-            return null;
         }
 
         /// Returns the most optimal table for compaction from a level that is due for compaction.
         /// Returns null if the level is not due for compaction (table_count_visible < count_max).
-        pub fn compaction_table(manifest: *const Manifest, level_a: u8) ?CompactionTableRange {
+        pub fn compaction_table(manifest: *const Manifest, level_a: u8) ?Level.CompactionTableRange {
             // The last level is not compacted into another.
             assert(level_a < constants.lsm_levels - 1);
+            assert(level_a >= 0);
```
kprotty: This is always true for `u8`.

src/lsm/manifest.zig
chaitanyabhandari: That's a great idea; I'll use this, thank you!

src/lsm/manifest.zig
chaitanyabhandari: Ack. 👍🏼

src/lsm/manifest_level.zig
sentientwaffle: Gotcha. Could we return a different value if we need to append a table but the array is already full? I think its misleading to return what is technically an incomplete range, even if we are guaranteed that the caller won't actually use it.

src/lsm/manifest_level.zig
chaitanyabhandari: Mhmmm, I had observed this as well -- ManifestLevel is currently oblivious of all compaction logic, and ManifestLog is oblivious of compaction as well, except for [this](https://github.com/tigerbeetledb/tigerbeetle/blob/1b64cac51bc103db6c673ad59dc69d7c12f04e6e/src/lsm/manifest_log.zig#L85-L97) small logic. I was wondering if that was by design. :) 



I am on board with keeping ManifestLevel oblivious of compaction and moving these out to `compaction_range.zig`. 

~~Another option could be moving them to `compaction.zig`?~~

Just realized that Compaction is unaware of ManifestLevel & ManifestLog, it only interacts with them through Manifest, so this does not seem like a good idea...

chaitanyabhandari: However, on second thought -- _compaction_range_ finds all tables in Level X that intersect with a given range. This fits the bill of a generic ManifestLevel operation rather than something strictly related to compaction... 

_compaction_table_, on the other hand, is definitely logic very specific to compaction as it implements our table selection strategy.

src/lsm/manifest_level.zig
chaitanyabhandari: Agreed, returning an incomplete range does seem misleading! 

Shall we add an argument called `max_intersecting_tables`/`intersection_threshold` to ManifestLevel.compaction_range and change its return type to `?CompactionRange`? `max_intersecting_tables` would be set to `lsm_growth_factor`. We could then return `null` if the # of tables that are intersecting cross that threshold. However, we would have to keep `CompactionRange.tables` as an `ArrayList` instead of a `BoundedArray` since the number of elements in the array would be driven by `max_intersecting_tables`/`intersection_threshold`.

src/lsm/compaction.zig
batiati: I'm ok with that, this implementation is generic enough to be used by both compaction and scans. In fact I think there is no reason to add `LevelB` to its name, we can just call it `LevelDataIterator` or even better just `LevelIterator`.

The particular case of having `tables: std.BoundedArray` as argument can also be used for scans, since we need to filter which TableInfos may contain the range we're looking for. I'm just not sure if a BoundedArray will suit scans (maybe we can abstract over a type when the time comes), but the idea seems good.

chaitanyabhandari: Thanks for the reply, @batiati. 😄 

That's perfect, I can replace LevelDataIterator with this iterator, we can rename it to _LevelIterator_ based on the discussion that you, me, & Joran had!

src/lsm/manifest_level.zig
chaitanyabhandari: Thanks for this great idea! 

I augmented [ManifestLevel.contains](https://github.com/tigerbeetledb/tigerbeetle/blob/3c815185cc24828b6e344b9d1d631064c9e53678/src/lsm/manifest_level.zig#L411) to check for the validity of stable TableInfo pointers and added a `constants.verify`-gated verification in [Manifest.update_table](https://github.com/tigerbeetledb/tigerbeetle/blob/3c815185cc24828b6e344b9d1d631064c9e53678/src/lsm/manifest.zig#L226-L231). 

All functions in Manifest that invoke ManifestLevel functions with TableInfo references ([Manifest.insert_table](https://github.com/tigerbeetledb/tigerbeetle/blob/3c815185cc24828b6e344b9d1d631064c9e53678/src/lsm/manifest.zig#L195), [Manifest.update_table](https://github.com/tigerbeetledb/tigerbeetle/blob/3c815185cc24828b6e344b9d1d631064c9e53678/src/lsm/manifest.zig#L226-L231), [Manifest.move_table](https://github.com/tigerbeetledb/tigerbeetle/blob/3c815185cc24828b6e344b9d1d631064c9e53678/src/lsm/manifest.zig#L244), [Manifest.remove_invisible_tables](https://github.com/tigerbeetledb/tigerbeetle/blob/3c815185cc24828b6e344b9d1d631064c9e53678/src/lsm/manifest.zig#L279)) now have `constants.verify`-gated verifications. However, none of the ManifestLevel functions that are invoked have any verification. I shall add these verifications to the ManifestLevel functions as well, for defense-in-depth.

src/lsm/manifest_level.zig
jorangreef: Great to think this through from different angles like this:
* To see where we can keep inner abstractions oblivious of outer abstractions.
* To move in-memory only logic from abstractions that also do I/O, to abstractions that only do in-memory logic for easier testing.

Perhaps the way forward is to keep these in Manifest, but to improve the names to be more descriptive, since they could also be seen as generic manifest utilities that compaction uses:
* `table_with_least_overlap_in_the_next_level()`
* `table_range()` or `overlapping_tables()` (or a name that also works well as a noun so that it can be used for the type result)

chaitanyabhandari: @jorangreef @sentientwaffle as part of https://github.com/tigerbeetledb/tigerbeetle/pull/875/commits/1132d35312a948249330de01ebf94dd2044a123c, I've made the functions in ManifestLevel more generic, they're now [ManifestLevel.table_with_least_overlap](https://github.com/tigerbeetledb/tigerbeetle/blob/1132d35312a948249330de01ebf94dd2044a123c/src/lsm/manifest_level.zig#L464) and [ManifestLevel.overlapping_tables](https://github.com/tigerbeetledb/tigerbeetle/blob/1132d35312a948249330de01ebf94dd2044a123c/src/lsm/manifest_level.zig#L584)!

[Tree.compact_start_tables](https://github.com/tigerbeetledb/tigerbeetle/blob/1132d35312a948249330de01ebf94dd2044a123c/src/lsm/tree.zig#L721) still invokes [Manifest.compaction_table](https://github.com/tigerbeetledb/tigerbeetle/blob/1132d35312a948249330de01ebf94dd2044a123c/src/lsm/manifest.zig#L424) → [ManifestLevel.table_with_least_overlap](https://github.com/tigerbeetledb/tigerbeetle/blob/1132d35312a948249330de01ebf94dd2044a123c/src/lsm/manifest_level.zig#L464) → [ManifestLevel.overlapping_tables](https://github.com/tigerbeetledb/tigerbeetle/blob/1132d35312a948249330de01ebf94dd2044a123c/src/lsm/manifest_level.zig#L584).

src/lsm/compaction.zig
chaitanyabhandari: FYI, I merged LevelBDataIterator into LevelDataIterator and changed the name to LevelIterator as a part of https://github.com/tigerbeetledb/tigerbeetle/pull/875/commits/f68290b8568dc9e410b6e1635a14e3573cdaa618 & https://github.com/tigerbeetledb/tigerbeetle/pull/875/commits/762412889e04a9456076c69e8121ef1620f1c97a.

src/lsm/manifest_level.zig
sentientwaffle: > Perhaps the way forward is to keep these in Manifest, but to improve the names to be more descriptive, since they could also be seen as generic manifest utilities that compaction uses:

@jorangreef Can you help me understand why these belong in the `Manifest`? Neither depends at all on the disk/IO.
Or did you mean `ManifestLevel`?

src/lsm/level_index_iterator.zig
```diff
@@ -89,12 +89,14 @@ pub fn LevelIndexIteratorType(comptime Table: type, comptime Storage: type) type
...
-                it.key_exclusive,
-                it.context.direction,
+                .{
+                    .level = it.context.level,
+                    .snapshot = it.context.snapshot,
+                    .key_min = it.context.key_min,
+                    .key_max = it.context.key_max,
+                    .key_exclusive = it.key_exclusive,
+                    .direction = it.context.direction,
+                },
```
sentientwaffle, line 99: Nit, but this can be formatted as (ignoring my mangled indentation):

```zig
            const next_table_info = it.context.manifest.next_table(.{
                  .level = it.context.level,
                  .snapshot = it.context.snapshot,
                  .key_min = it.context.key_min,
                  .key_max = it.context.key_max,
                  .key_exclusive = it.key_exclusive,
                  .direction = it.context.direction,
              });
```

src/lsm/level_iterator.zig
```diff
@@ -0,0 +1,177 @@
...
+        const TableInfo = Manifest.TableInfo;
+        const TableDataIterator = TableDataIteratorType(Storage);
+
+        pub const Context = struct {
+            grid: *Grid,
+            level: u8,
+            snapshot: u64,
+            index_block: BlockPtr,
+            tables: std.BoundedArray(
+                *TableInfo,
```
sentientwaffle: Please add a comment explaining that these pointers are into `ManifestLevel`.

src/lsm/manifest.zig
```diff
@@ -119,6 +119,21 @@ pub fn ManifestType(comptime Table: type, comptime Storage: type) type {
...
+        pub const CompactionRange = struct {
+            /// The total number of tables in the compaction across both levels, always at least 1.
+            table_count: usize,
+            /// The minimum key across both levels.
+            key_min: Key,
+            /// The maximum key across both levels.
+            key_max: Key,
+            // References to tables in level B that intersect with the chosen table in level A.
+            tables: std.BoundedArray(*TableInfo, constants.lsm_growth_factor),
+        };
```
sentientwaffle, line 138:
```suggestion
        };

```

src/lsm/manifest_level.zig
```diff
@@ -204,9 +244,15 @@ pub fn ManifestLevelType(
                     }
                 } else {
                     switch (direction) {
-                        .ascending => break :blk level.tables.iterator_from_index(0, direction),
```
sentientwaffle, line 207: Can you undo the unnecessary reformats? Our column width is 100, anything under that is fine as-is, and the reformats make the relevant diff harder to review.

src/lsm/manifest_level.zig
```diff
@@ -38,6 +38,23 @@ pub fn ManifestLevelType(
...
+            /// The maximum key across both levels.
+            key_max: Key,
+            // References to tables in level B that intersect with the chosen table in level A.
+            tables: std.BoundedArray(*TableInfo, constants.lsm_growth_factor),
+        };
+        pub const TableInfoPtr = union(enum) { copy: *TableInfo, from_level: *TableInfo };
+        pub const TableInfoPtrConst = union(enum) {
+            copy: *const TableInfo,
+            from_level: *const TableInfo,
+        };
```
sentientwaffle, line 55:
```suggestion
        };

```

sentientwaffle: Rather than creating these types, what about passing `*TableInfo`/`*const TableInfo`, with an extra `TableInfoOwner: enum { external, internal }`? Then there's no need for

```zig
            const table = switch (table_union) {
                .copy => table_union.copy,
                .from_level => table_union.from_level,
            };
```

Alternatively, what about

```
TableInfoReference = union(enum) {
  internal: struct { table_info: *TableInfo, generation: u64 },
  external: struct { table_info: *TableInfo },
  fn table() *TableInfo
}
TableInfoReferenceConst = union(enum) {
  internal: struct { table_info: *const TableInfo, generation: u64 },
  external: struct { table_info: *const TableInfo },
  fn table() *const TableInfo
}
```

(use a generation number (modified on insert/remove, but not in-place snapshot update) so that we can inexpensively verify that an invalid internal reference isn't ever used.)

src/lsm/manifest_level_fuzz.zig
```diff
@@ -224,47 +226,50 @@ const GenerateContext = struct {
...
         snapshot: u64,
 
-        pub fn run_fuzz_ops(random: std.rand.Random, fuzz_ops: []const FuzzOp) !void {
+        pub fn init(random: std.rand.Random) !Environment {
             var env: Environment = undefined;
 
             const node_pool_size = ManifestLevel.Keys.node_count_max +
                 ManifestLevel.Tables.node_count_max;
             env.pool = try NodePool.init(allocator, node_pool_size);
-            defer env.pool.deinit(allocator);
```
sentientwaffle, line 254: Don't remove these; change them to `errdefer`.

src/lsm/manifest_level.zig
```diff
@@ -374,20 +423,216 @@ pub fn ManifestLevelType(
...
+            key_max: Key,
+            max_overlapping_tables: usize,
+        ) ?OverlapRange {
+            assert(max_overlapping_tables <= constants.lsm_growth_factor);
+            var range = OverlapRange{
+                .key_min = key_min,
+                .key_max = key_max,
+                .tables = .{ .buffer = undefined },
+            };
+            const snapshots = [_]u64{lsm.snapshot_latest};
```
sentientwaffle: Extract this into a parameter to pass into `overlapping_tables`.

src/lsm/manifest_level.zig
```diff
@@ -374,20 +423,216 @@ pub fn ManifestLevelType(
...
+        ///   overlapping tables. This is because OverlapRange.tables is a
+        ///   BoundedArray of size lsm_growth_factor. This works with our current
+        ///   compaction strategy that is guaranteed to choose a table with that
+        ///   intersects with <= lsm_growth_factor tables in the next level.
+        pub fn overlapping_tables(
+            level: *const Self,
+            key_min: Key,
+            key_max: Key,
+            max_overlapping_tables: usize,
+        ) ?OverlapRange {
```
sentientwaffle: The word 'overlap' brings to mind that these are tables whose key ranges intersect with one another, when in fact the ranges are disjoint. What about e.g. `select_tables`?

src/lsm/manifest_level.zig
```diff
@@ -374,20 +423,216 @@ pub fn ManifestLevelType(
...
+            assert(max_overlapping_tables <= constants.lsm_growth_factor);
```
sentientwaffle, line 551:
```suggestion
            assert(max_overlapping_tables <= constants.lsm_growth_factor);

```

src/lsm/manifest.zig
```diff
@@ -434,115 +428,65 @@ pub fn ManifestType(comptime Table: type, comptime Storage: type) type {
...
+            const manifest_level: *const Level = &manifest.levels[level_b];
+
+            // We are guaranteed to get a non-null range because Level 0 has
+            // growth_max number of tables, so the number of tables that intersect
+            // with the immutable table can be no more than growth_max.
+            const range = manifest_level.overlapping_tables(
+                key_min,
+                key_max,
+                growth_factor,
+            ) orelse unreachable;
```
sentientwaffle:
```suggestion
            ).?;
```

src/lsm/level_data_iterator.zig
sentientwaffle, line 1: Alas, I prefer the `LevelDataIterator`/`LevelIndexIterator` split and naming, but I guess I'm outvoted...

Even so, could the rename (or at least the file move) be deferred to its own PR? Github makes it very inconvenient to see where the file actually changed, because it is showing up as fully removed + recreated rather than just renamed + modified.

src/lsm/manifest_level.zig
chaitanyabhandari: I assumed Joran meant `ManifestLevel` even though he typed `Manifest`; but that's just because we had a discussion about moving the purely in-memory functionality from `Manifest` → `ManifestLevel` in our one-on-one yesterday morning. I hope I'm right! 😅

src/lsm/manifest_level.zig
```diff
@@ -82,43 +99,62 @@ pub fn ManifestLevelType(
...
+                .copy => table_union.copy,
+                .from_level => table_union.from_level,
+            };
+            if (constants.verify) {
+                switch (table_union) {
+                    .copy => assert(level.contains(.{ .copy = table })),
+                    .from_level => assert(level.contains(.{ .from_level = table })),
+                }
+            }
+            assert(table.snapshot_max == math.maxInt(u64));
```
sentientwaffle, line 129: Move this back up to below `assert(snapshot < ...);` so that the snapshot asserts are grouped together.

src/lsm/manifest_level.zig
```diff
@@ -374,20 +423,216 @@ pub fn ManifestLevelType(
...
+                    },
+                    // If the table provided is a pointer to the table in the ManifestLevel,
+                    // we check for equality of the pointers. This is to ensure
+                    // that the provided table is a valid pointer in the ManifestLevel.
+                    .from_level => {
+                        if (level_table == table) return true;
+                    },
+                }
             }
             return false;
```
sentientwaffle, line 409: Can we `assert(table_union == .copy);` here? i.e. if it is `from_level`, then we would expect to have found it in our level.

src/lsm/compaction.zig
```diff
@@ -135,6 +132,10 @@ pub fn CompactionType(
         // Passed by `start`.
         context: Context,
 
+        // Pased by `start`. Defines whether the move optimization applies to this
+        // compaction. This field is set to True if the optimal compaction
+        // table in level A can simply be moved to level B.
```
```suggestion
        // Whether this compaction will use the move-table optimization.
```

src/lsm/compaction.zig
```diff
@@ -135,6 +132,10 @@ pub fn CompactionType(
...
+        can_move_table: bool,
```
sentientwaffle: Maybe we should rename to `move_table`. `can_move_table` might be confusing, since it implies the question "under what circumstances are we able to move-table, but don't".

src/lsm/compaction.zig
```diff
@@ -410,40 +425,50 @@ pub fn CompactionType(
...
-            table_info: TableInfo,
-            index_block: BlockPtrConst,
-        ) void {
-            const compaction = @fieldParentPtr(Compaction, "iterator_b", iterator_b);
+        fn on_index_block(iterator_b: *LevelIterator) void {
+            const compaction = @fieldParentPtr(
+                Compaction,
+                "iterator_b",
+                iterator_b,
+            );
```
sentientwaffle: Don't wrap, this isn't a long line.

src/lsm/compaction.zig
```diff
@@ -319,7 +327,10 @@ pub fn CompactionType(
                 });
 
                 compaction.state = .next_tick;
-                compaction.context.grid.on_next_tick(done_on_next_tick, &compaction.next_tick);
+                compaction.context.grid.on_next_tick(
+                    done_on_next_tick,
+                    &compaction.next_tick,
+                );
```
sentientwaffle: Don't wrap, this isn't a long line.

src/lsm/manifest_level_fuzz.zig
```diff
@@ -385,7 +390,7 @@ fn EnvironmentType(comptime table_count_max: u32, comptime node_size: u32) type
                 const env_table = env.find_exact(level_table);
                 assert(level_table.equal(env_table));
 
-                env.level.set_snapshot_max(env.snapshot, env_table);
+                env.level.set_snapshot_max(env.snapshot, .{.copy=env_table});
```
sentientwaffle: Weird that `zig fmt` didn't fix this
```suggestion
                env.level.set_snapshot_max(env.snapshot, .{.copy = env_table});
```

src/lsm/manifest_level.zig
```diff
@@ -349,7 +395,10 @@ pub fn ManifestLevelType(
             key_cursor: Keys.Cursor,
             direction: Direction,
         ) Keys.Cursor {
-            var reverse = level.keys.iterator_from_cursor(key_cursor, direction.reverse());
+            var reverse = level.keys.iterator_from_cursor(
+                key_cursor,
+                direction.reverse(),
+            );
```
sentientwaffle: Don't wrap, this isn't a long line.

src/lsm/level_iterator.zig
```diff
@@ -0,0 +1,177 @@
...
+            const callback = it.callback.level_next;
+            it.callback = .none;
+            // `index_block` is only valid for this callback, so copy it's contents.
+            // TODO(jamii) This copy can be avoided if we bypass the cache.
+            stdx.copy_disjoint(
+                .exact,
+                u8,
+                it.context.index_block,
+                index_block,
+            );
```
sentientwaffle: Not a long line, &c

src/lsm/compaction.zig
```diff
@@ -410,40 +425,50 @@ pub fn CompactionType(
...
-            });
-
+        // TODO: Support for LSM snapshots would require us to only remove blocks
+        // that are invisible.
+        fn release_table_blocks(compaction: *Compaction, index_block: BlockPtrConst) void {
             // Release the table's block addresses in the Grid as it will be made invisible.
-            // This is safe; iterator_b makes a copy of the block before calling us.
+            // This is safe; compaction.index_block_b holds a of the index block for a
+            // table in Level B. Additionally, compaction.index_block_a holds
+            // a copy of the index block for the Level A table being compacted.
```
sentientwaffle: Typo "holds a of the"

src/lsm/compaction.zig
```diff
@@ -767,28 +797,27 @@ pub fn CompactionType(
...
-                            .table = table,
-                        }),
-                    }
-
                     compaction.state = .next_tick;
-                    compaction.context.grid.on_next_tick(done_on_next_tick, &compaction.next_tick);
+                    compaction.context.grid.on_next_tick(
+                        done_on_next_tick,
+                        &compaction.next_tick,
+                    );
```
sentientwaffle: Don't wrap

src/lsm/compaction.zig
```diff
@@ -362,13 +372,18 @@ pub fn CompactionType(
 
             // `index_block` is only valid for this callback, so copy its contents.
             // TODO(jamii) This copy can be avoided if we bypass the cache.
-            stdx.copy_disjoint(.exact, u8, compaction.index_block_a, index_block);
+            stdx.copy_disjoint(
+                .exact,
+                u8,
+                compaction.index_block_a,
+                index_block,
+            );
```
sentientwaffle: Unnecessary wrap

src/lsm/manifest_level.zig
```diff
@@ -374,20 +423,216 @@ pub fn ManifestLevelType(
...
+        /// * Exits early if it finds a table that doesn't overlap with any
+        ///   tables in the second level.
+        pub fn table_with_least_overlap(
+            level_a: *const Self,
+            level_b: *const Self,
+            max_overlapping_tables: usize,
+        ) ?LeastOverlapTable {
+            assert(max_overlapping_tables <= constants.lsm_growth_factor);
+            var optimal: ?LeastOverlapTable = null;
+            const snapshots = [1]u64{lsm.snapshot_latest};
```
sentientwaffle: Pass this in through a parameter.

src/lsm/manifest_level.zig
```diff
@@ -82,43 +99,62 @@ pub fn ManifestLevelType(
         }
 
         /// Set snapshot_max for the given table in the ManifestLevel.
-        ///
+        /// The table provided could either be a pointer to the table (from_level) in the
+        /// ManifestLevel or a pointer to a copy of that table (copy).
```
sentientwaffle:
```suggestion
        ///
```

This is implied by the type signature, no need to explain again.

src/lsm/manifest_level.zig
```diff
@@ -189,7 +225,11 @@ pub fn ManifestLevelType(
                 if (key_range) |range| {
                     assert(compare_keys(range.key_min, range.key_max) != .gt);
 
-                    if (level.iterator_start(range.key_min, range.key_max, direction)) |start| {
+                    if (level.iterator_start(
+                        range.key_min,
+                        range.key_max,
+                        direction,
+                    )) |start| {
```
sentientwaffle: Don't wrap

src/lsm/manifest.zig
```diff
@@ -434,115 +428,65 @@ pub fn ManifestType(comptime Table: type, comptime Storage: type) type {
...
             key_max: Key,
         ) CompactionRange {
-            assert(level_b < constants.lsm_levels);
             assert(compare_keys(key_min, key_max) != .gt);
+            const level_b = 0;
+            const manifest_level: *const Level = &manifest.levels[level_b];
+
+            // We are guaranteed to get a non-null range because Level 0 has
+            // growth_max number of tables, so the number of tables that intersect
+            // with the immutable table can be no more than growth_max.
```
sentientwaffle:
```suggestion
            // lsm_growth_factor number of tables, so the number of tables that intersect
            // with the immutable table can be no more than lsm_growth_factor.
```

src/lsm/compaction.zig
```diff
@@ -826,11 +858,30 @@ pub fn CompactionType(
...
             const snapshot_max = snapshot_max_for_table_input(compaction.context.op_min);
+
+            // Update snapshot_max only if table in level A has been compacted
+            // with tables from level B. If no compaction is required, i.e.
+            // if a table in level A can simply be moved to level B, we needn't
+            // update snapshot_max.
+            // This update MUST be done before insert_table() and move_table()
+            // since it directly uses pointers to the ManifestLevel tables to
+            // perform updates. Calling insert_table() and move_table() before
+            // this update will cause these pointers to become invalid.
```
sentientwaffle, line 832:
```suggestion
            // Update snapshot_max iff table in level A was compacted
            // with tables from level B. If no compaction was required (move-table), we needn't
            // update snapshot_max.
            // Apply the updates before insert_table() or move_table().
            // Calling insert_table() and move_table() before
            // these updates would invalidate the TableInfo pointers.
```

src/lsm/manifest.zig
```diff
@@ -204,22 +219,29 @@ pub fn ManifestType(comptime Table: type, comptime Storage: type) type {
...
-        pub fn update_table(
-            manifest: *Manifest,
-            level: u8,
-            snapshot: u64,
-            table: *TableInfo,
-        ) void {
+        /// Updates the snapshot_max on the provided table for the given level.
+        /// The table provided could either be a pointer to the table (from_level) in the
+        /// ManifestLevel or a pointer to a copy of that table (copy).
+        pub fn update_table(manifest: *Manifest, level: u8, snapshot: u64, table_union: Level.TableInfoPtr) void {
```
sentientwaffle: Afaict we only use the `copy` branch from a test. Can we drop it and just always pass in an internal reference? `struct { *TableInfo, generation }`.

src/lsm/manifest_level.zig
sentientwaffle: Yeah `ManifestLevel` makes sense to me, especially since the naming is now generalized away from compaction.

src/lsm/manifest_level.zig
chaitanyabhandari: Argh, sorry for the reformats. I got thrown off by a VSCode setting that displays the parameter names as hints. Every line seemed so long! I have finally disabled that setting and I'm fixing all of these. 😄

src/lsm/level_data_iterator.zig
chaitanyabhandari: Ack, I shall defer the rename to another PR. For this PR, I'll just add my code to level_data_iterator.zig so that it appears as a change to the file!

src/lsm/manifest_level.zig
chaitanyabhandari: We could also make the meaning of overlap more explicit, how about `tables_overlapping_with_range` or `tables_intersecting_with_range`?

src/lsm/level_index_iterator.zig
chaitanyabhandari: @sentientwaffle I actually realized that with my iterator, we now don't need LevelIndexIterator at all... Shall I go ahead and delete that file? Just wanted to make sure if we have some future use cases where it might end up being useful!

src/lsm/manifest_level.zig
chaitanyabhandari: If we were to use _internal_ & _external_ to categorize *TableInfo, then all invocations to insert_table & remove_table use _external *const TableInfo_. With my changes, compaction would now use an _internal *TableInfo_ to update snapshot_max.

Since the only place _internal *TableInfo_  is being used is _update_table_, I didn't understand how we can leverage the generation number to verify that an invalid internal reference isn't ever used! Could you please expand on it a bit? I would also be happy to jump on a quick call to understand things more clearly. :D

src/lsm/manifest_level.zig
chaitanyabhandari: This assertion has to be below our declaration of the _table_ variable, so I moved the other asserts below to group them together!

src/lsm/manifest_level.zig
chaitanyabhandari: Ah, maybe I misunderstood what you said in https://github.com/tigerbeetledb/tigerbeetle/pull/875#discussion_r1228536880? I thought we wanted to validate whether a pointer that the invoker of `update_table` claims is `from_level` is really a *TableInfo within ManifestLevel. Wouldn't that be useful to check?

Or maybe I didn't completely understand what you meant in this comment! Could you please explain it a bit more? :D

src/lsm/manifest_level.zig
sentientwaffle: This code isn't 100% in my head so maybe this wouldn't work.
But what I was imagining is:

- Add a `generation: usize` field to `ManifestLevel` (initially `0`). 
- When we are querying the compaction range, include the current value of `generation` in the result.
- Whenever `ManifestLevel.insert_table`/`remove_table` is called, increment the `ManifestLevel`'s `generation`.
- In `ManifestLevel.set_snapshot_max`, assert that the parameter's `generation` matches the `ManifestLevel`'s own generation. If they don't match, it means that there was a insert/remove in the mean time, so the internal pointer is not valid.

`contains()` checks that the pointer is indeed internal, but it doesn't check that the underlying table being pointed to is the same that it was when the compaction first started.

src/lsm/manifest_level.zig
sentientwaffle: > I thought we wanted to validate whether a pointer that the invoker of update_table claims is from_level is really a *TableInfo within ManifestLevel. Wouldn't that be useful to check?

Yes, that's right.

To expand on:

> Can we assert(table_union == .copy); here? i.e. if it is from_level, then we would expect to have found it in our level.


We are about to `return false`, i.e. "the manifest level does not contain the table that was passed in".
But if `table_union == .from_level` (i.e. an internal pointer), then by definition the pointer should be in our level.

chaitanyabhandari: > We are about to return false, i.e. "the manifest level does not contain the table that was passed in".
But if table_union == .from_level (i.e. an internal pointer), then by definition the pointer should be in our level.

Ah, understood! I was under the impression that returning `False` here is valid even if `table_union == .from_level`, because maybe the TableInfo* pointer doesn't exist in the ManifestLevel due to rearrangements in the Segmented Array causing the pointer to get invalidated?

sentientwaffle: > I was under the impression that returning False here is valid even if table_union == .from_level, because maybe the TableInfo* pointer doesn't exist in the ManifestLevel due to rearrangements in the Segmented Array causing the pointer to get invalidated?

Can you elaborate on this case? If the segmentedarray was rearranged then it the caller shouldn't be using `from_level`, right?

src/lsm/manifest_level.zig
chaitanyabhandari: @sentientwaffle: Ah, I understand the idea now! We already have `table_count_visible` that is incremented in [ManifestLevel.insert_table](https://github.com/tigerbeetledb/tigerbeetle/blob/8bb876834790e440e251626cc28a106a94692c8d/src/lsm/manifest_level.zig#L79) and decremented in [ManifestLevel.set_snapshot_max](https://github.com/tigerbeetledb/tigerbeetle/blob/8bb876834790e440e251626cc28a106a94692c8d/src/lsm/manifest_level.zig#L117) and [ManifestLevel.remove_table_visible](https://github.com/tigerbeetledb/tigerbeetle/blob/8bb876834790e440e251626cc28a106a94692c8d/src/lsm/manifest_level.zig#L130). However, this doesn't track all calls to `insert_table` & `remove_table`. 

We could maintain a `table_count` that could be used as a generation number. We could return the `table_count` when we're finding the optimal compaction table and cache it. Later, when we are about to update the tables by reference, we could validate if the ManifestLevel's `table_count` is the same as the one that we have cached!

Does this approach sound reasonable?

sentientwaffle: I think I would make it an explicit generation number that is always incremented. That way a balanced insert/remove is still detected as invalid.

src/lsm/manifest_level.zig
chaitanyabhandari: Yes, the caller shouldn't be using ManifestLevel references if the Segmented Array was rearranged! I thought that returning `False` from `ManifestLevel.contains` would be a good way to catch issues where the caller is using`.from_level` but the pointer being used has been invalidated...

src/lsm/manifest_level.zig
chaitanyabhandari: Ah yes, a monotonically increasing generation number would indeed be more robust. Thanks a lot for expanding on the idea, dj!

chaitanyabhandari:
Thanks for the super thorough review, @sentientwaffle! FYI, I have taken a stab at the change that we discussed in https://github.com/tigerbeetledb/tigerbeetle/pull/875#discussion_r1230195704 as a part of https://github.com/tigerbeetledb/tigerbeetle/pull/875/commits/ef8b53e44d3b93176b2d7bf257bb2f8637ed06e8.

It spans changes across, manifest.zig, manifest_level.zig, compaction.zig & also manifest_level_fuzz.zig, so I ran the following tests again:

- ./scripts/build.sh fuzz_lsm_tree ✅
- ./scripts/build.sh fuzz_lsm_forest ✅
- ./scripts/build.sh fuzz_lsm_manifest_level ✅

src/lsm/manifest.zig
chaitanyabhandari: Yes, [ManifestLevel.set_snapshot_max](https://github.com/tigerbeetledb/tigerbeetle/blob/ef8b53e44d3b93176b2d7bf257bb2f8637ed06e8/src/lsm/manifest_level.zig#L162) is only invoked with an `external TableInfoReference` via [this](https://github.com/tigerbeetledb/tigerbeetle/blob/ef8b53e44d3b93176b2d7bf257bb2f8637ed06e8/src/lsm/manifest_level.zig#L973) ManifestLevel test! 

Are you recommending that we allow updating a table only using an `internal TableInfoReference`? Specifically, are you suggesting we `assert(table_ref == .internal)` in ManifestLevel.set_snapshot_max & get rid of [this](https://github.com/tigerbeetledb/tigerbeetle/blob/ef8b53e44d3b93176b2d7bf257bb2f8637ed06e8/src/lsm/manifest_level.zig#L175-L194) piece of code for updating via `external TableInfoReference`?

sentientwaffle: A rule of thumb I use is to narrow the types as much as possible.

So, for example
- Right now `insert_table` takes a `TableInfoReference` which is always external. So instead it should just take a `*TableInfo`.
- `LeastOverlapTable` and `OverlapRange` contain `TableInfoReference`s. But those references are always internal. So it should just contain `struct { table_info: *const TableInfo, generation: u64 }` instead.
- `set_snapshot_max` is always called internal (ignoring the test, which can be modified). So it should just receive a `struct { table_info: *const TableInfo, generation: u64 }` instead. 
- And so forth. I did a quick look through and it seemed like no operations actually need to handle both internal and external, always one or the other.

If no operations actually need to handle both internal _and_ external pointers, then we don't need to `union(enum)`s at all. I would create a `struct { table_info: *const TableInfo, generation: u64 }` (e.g. named `TableInfoReference` since that is now free) to use for internal tables, and just use `*const TableInfo` for external.

chaitanyabhandari: Thank you for the advice, @sentientwaffle! 

So as of now, these functions need to handle both internal & external references:

* `remove_table` - [Manifest.remove_invisible_tables](https://github.com/tigerbeetledb/tigerbeetle/blob/ef8b53e44d3b93176b2d7bf257bb2f8637ed06e8/src/lsm/manifest.zig#L324) uses internal references & [Manifest.move_table](https://github.com/tigerbeetledb/tigerbeetle/blob/ef8b53e44d3b93176b2d7bf257bb2f8637ed06e8/src/lsm/manifest.zig#L285) uses external references. But there is no difference in the behaviour of `remove_table` based on this, except we can only  `assert(!level.contains(table))` in `remove_table` if the removal is done via an external reference (since an internal reference is invalidated). 

*  `set_snapshot_max` 

`insert_table` only needs to handle external references.

If we're change [this](https://github.com/tigerbeetledb/tigerbeetle/blob/ef8b53e44d3b93176b2d7bf257bb2f8637ed06e8/src/lsm/manifest_level.zig#L973) ManifestLevel test to use internal references, then I do agree that exposing `TableInfoReference` & `TableInfoReferenceConst` would be overkill just for `remove_table`! 

I can revert to a simpler version where:

- `TableInfoReference` is defined as `struct { table_info: *TableInfo, generation: u64 }` instead of an `enum`
-  `Manifest.update_table` & `ManifestLevel.set_snapshot_max` take `TableInfoReference` as argument since they only work with internal references
- `LeastOverlapTable` & `TableOverlapRange` uses `TableInfoReference`
- `ManifestLevel.insert_table` & `ManifestLevel.remove_table` take `const* TableInfo` as input, like they did before
- `ManifestLevel.contains` currently contains our logic for ensuring that internal pointers are valid, I can keep that logic within `Manifest.update_table` & `ManifestLevel.set_snapshot_max` since they're the only functions that work with internal pointers and need that validation: https://github.com/tigerbeetledb/tigerbeetle/blob/ef8b53e44d3b93176b2d7bf257bb2f8637ed06e8/src/lsm/manifest_level.zig#L484-L492

Does the aforementioned design look okay?

src/lsm/manifest_level.zig
```diff
@@ -50,6 +107,11 @@ pub fn ManifestLevelType(
         /// Used to enforce table_count_max_for_level().
         table_count_visible: u32 = 0,
 
+        /// A monotonically increasing generation number that is incremented in insert_table() &
+        /// decremented in remove_table(). This is used detect invalid TableInfo references
+        /// when a caller is using an `internal` TableInfoReference/TableInfoReferenceConst.
```
sentientwaffle:
```suggestion
        /// A monotonically increasing generation number that is used detect invalid internal TableInfo references.
```

Don't need to be quite so detailed here imo; "generation" is pretty self-explanatory. Also, the "decremented" part wasn't correct.

chaitanyabhandari: Oops, my bad, the "decreasing" is indeed wrong; I'll fix this!

src/lsm/manifest.zig
sentientwaffle: > Does the aforementioned design look okay?

Yeah, looks good.

> So as of now, these functions need to handle both internal & external references:
> - `set_snapshot_max`

Where does the external `set_snapshot_max` call originate from? I missed it earlier.

chaitanyabhandari: It's invoked from the ManifestLevel test, here: https://github.com/tigerbeetledb/tigerbeetle/blob/ef8b53e44d3b93176b2d7bf257bb2f8637ed06e8/src/lsm/manifest_level.zig#L973

sentientwaffle: Ah right, I discounted that since I figured it was worth just changing the test if necessary. Thanks!

chaitanyabhandari: @sentientwaffle: FYI, I've implemented the changes that we discussed in this thread as part of https://github.com/tigerbeetledb/tigerbeetle/pull/875/commits/2c745c71bf24b14ee4120dd0dc9e7df690f74210!

src/lsm/level_data_iterator.zig
```diff
@@ -64,128 +70,101 @@ pub fn LevelDataIteratorType(comptime Table: type, comptime Storage: type) type
 
         pub fn init(
             allocator: mem.Allocator,
```
batiati: Can we get rid of `allocator` here? It isn't used by `TableDataIterator` either, but there's a comment reserving it for future use 🤔.

https://github.com/tigerbeetledb/tigerbeetle/blob/8bb876834790e440e251626cc28a106a94692c8d/src/lsm/table_data_iterator.zig#L41

If so, we can also remove `deinit` for these structs.

src/lsm/level_data_iterator.zig
```diff
@@ -10,37 +10,43 @@ const ManifestType = @import("manifest.zig").ManifestType;
...
+// tables to be iterated over are passed via the `tables` field in the context,
+// in the start() function.
+// * Assumption: This iterator works under the assumption that the references to
+// the tables are stable. For references fetched from the ManifestLevel, this
+// assumption only holds true till move_table & insert_table are called as a part
+// of apply_to_manifest. This is because move_table & insert_table are both operations
+// that insert/remove tables from the ManifestLevel segmented arrays, which could cause
+// rearrangements in the memory layout of the arrays. These rearrangements could
+// cause our table references to get invalidated.
+pub fn LevelIteratorType(comptime Table: type, comptime Storage: type) type {
```
batiati: Rename the file to `level_iterator.zig`

chaitanyabhandari: Hey, Rafael! As part of https://github.com/tigerbeetledb/tigerbeetle/pull/875#discussion_r1230204218, we decided to defer file rename to a separate PR, because GitHub shows the rename as a file deletion + file addition, which makes reviewing cumbersome...

src/lsm/manifest_level.zig
```diff
@@ -38,6 +38,22 @@ pub fn ManifestLevelType(
 
         pub const Tables = SegmentedArray(TableInfo, NodePool, table_count_max, .{});
 
+        pub const TableInfoReference = struct { table_info: *TableInfo, generation: u64 };
```
sentientwaffle: Either `u64` or `u32` is fine for `generation`, but be consistent. (`ManifestLevel.generation` is currently `u32`, `TableInfoReference.generation` is `u64`).

src/lsm/level_data_iterator.zig
chaitanyabhandari: @batiati I don't exactly know what _pipelining_ refers to here. Does it refer to the _Compaction Pacing_ work that Fed is working on, or is it something different? If we're sure we wouldn't need this in the future, then I'm happy to remove it. 😃 

cc: @cb22 @sentientwaffle

src/lsm/manifest_level.zig
chaitanyabhandari: Ah, my bad! I'll make both of them `u32`.

src/lsm/level_data_iterator.zig
cb22: @chaitanyabhandari I think it has to do with potential future read pipelining? But I'm not sure - so far I've only focused on the write path!

chaitanyabhandari: Got it, thank you Fed! 

FYI, I've removed the unused allocator from `LevelIterator` & `TableDataIterator` as part of https://github.com/tigerbeetledb/tigerbeetle/pull/875/commits/376120a5eb12a8923eddb0d1d72e1f026c254b94! We can add it back if/when we require it for read pipelining.

cc: @batiati

src/lsm/compaction.zig
sentientwaffle: Wasn't applied. If you want to keep the existing sentences that's fine, but fix the typo in the first sentence.

src/lsm/level_data_iterator.zig
```diff
@@ -50,142 +56,113 @@ pub fn LevelDataIteratorType(comptime Table: type, comptime Storage: type) type
...
+            it.table_data_iterator.deinit();
             it.* = undefined;
         }
 
         pub fn start(
-            it: *LevelDataIterator,
+            it: *LevelIterator,
             context: Context,
         ) void {
             assert(it.callback == .none);
```
sentientwaffle, line 91:
```suggestion
            assert(it.callback == .none);
            assert(context.level < constants.lsm_levels);
            assert(context.tables.len > 0);
```

src/lsm/compaction.zig
```diff
@@ -203,7 +210,7 @@ pub fn CompactionType(
 
                 .input_state = .remaining,
                 .state = .idle,
-
+                .move_table = false,
```
sentientwaffle: Nit: I would initialize this to `undefined` to make it clear that this initial value is never used.

src/lsm/manifest.zig
```diff
@@ -119,6 +119,23 @@ pub fn ManifestType(comptime Table: type, comptime Storage: type) type {
...
+
+        pub const CompactionRange = struct {
+            /// The total number of tables in the compaction across both levels, always at least 1.
+            table_count: usize,
+            /// The minimum key across both levels.
+            key_min: Key,
+            /// The maximum key across both levels.
+            key_max: Key,
+            // References to tables in level B that intersect with the chosen table in level A.
+            tables: std.BoundedArray(TableInfoReference, constants.lsm_growth_factor),
```
sentientwaffle, line 137: Maybe `tables_b`? Not as nice a name, but otherwise it is confusing that `table_count == tables.len + 1`.

src/lsm/manifest.zig
```diff
@@ -119,6 +119,23 @@ pub fn ManifestType(comptime Table: type, comptime Storage: type) type {
 
         const ManifestLog = ManifestLogType(Storage, TableInfo);
 
+        pub const TableInfoReference = Level.TableInfoReference;
+
+        const CompactionTableRange = struct {
+            table: TableInfoReference,
```
sentientwaffle: `table_a`? This isn't obvious otherwise.

src/lsm/compaction.zig
chaitanyabhandari: Oops, I missed this somehow! I have addressed this in https://github.com/tigerbeetledb/tigerbeetle/pull/875/commits/3ebed30f3c4db81fe5890a064128f9eb90cd265e.

src/lsm/tree.zig
```diff
@@ -678,8 +678,7 @@ pub fn TreeType(comptime TreeTable: type, comptime Storage: type, comptime tree_
             assert(values_count > 0);
 
             const level_b: u8 = 0;
-            const range = tree.manifest.compaction_range(
-                level_b,
+            const range = tree.manifest.immutable_table_compaction_range(
                 tree.table_immutable.key_min(),
                 tree.table_immutable.key_max(),
             );
```
sentientwaffle, line 684:
```suggestion
            );
            assert(range.table_count == range.tables.len + 1);
```

src/lsm/manifest.zig
```diff
@@ -556,16 +503,16 @@ pub fn ManifestType(comptime Table: type, comptime Storage: type) type {
             assert(compare_keys(range.key_min, range.key_max) != .gt);
 
             var level_c: u8 = level_b + 1;
-            while (level_c < constants.lsm_levels) : (level_c += 1) {
-                const snapshots = [_]u64{snapshot_latest};
```
sentientwaffle: Nit:
```suggestion
```
(Normally I wouldn't bother, but the extra newline makes git-diff incorrectly infer that the `while` loop line changed, which bugged me :stuck_out_tongue:).

src/lsm/manifest.zig
```diff
@@ -372,60 +398,29 @@ pub fn ManifestType(comptime Table: type, comptime Storage: type) type {
...
-                };
-                if (next) return table;
-            }
-
-            return null;
+        }) ?*const TableInfo {
+            assert(parameters.level < constants.lsm_levels);
+            assert(compare_keys(parameters.key_min, parameters.key_max) != .gt);
+            const manifest_level: *const Level = &manifest.levels[parameters.level];
+            return manifest_level.next_table(.{
```
sentientwaffle:
```suggestion
            return manifest.levels[parameters.level].next_table(.{
```

src/lsm/manifest.zig
chaitanyabhandari: Maybe we could contain these changes to [CompactionTableRange](https://github.com/tigerbeetledb/tigerbeetle/blob/d88d9cdf32f29ee3e90ed2822bb88df9cd0e0e3c/src/lsm/manifest.zig#L124-L127) and change it to:

```C
        const CompactionTableRange = struct {
            table_a: TableInfoReference,
            range_b: CompactionRange,
        };

```

Also, to make things more explicit, we could change the usage [here](https://github.com/tigerbeetledb/tigerbeetle/blob/d88d9cdf32f29ee3e90ed2822bb88df9cd0e0e3c/src/lsm/tree.zig#L681) to be `const range_b = tree.manifest.immutable_table_compaction_range` so that the assertions make more sense.

Does that sound better?

src/lsm/level_data_iterator.zig
chaitanyabhandari: I only added `assert(context.level < constants.lsm_levels);`. We can't assert `assert(context.tables.len > 0);` since we invoke `LevelTableValueBlockIteratorType.start()` for the case where the Immutable Table is being compacted into Level-0, even if there are no tables in Level 0 that intersect with the Immutable Table.

src/lsm/manifest.zig
chaitanyabhandari: FYI, I implemented these changes in https://github.com/tigerbeetledb/tigerbeetle/commit/032573c463deac065a9544983f8d691019cd8e13!

src/lsm/manifest.zig
chaitanyabhandari: I implemented these changes in https://github.com/tigerbeetledb/tigerbeetle/commit/032573c463deac065a9544983f8d691019cd8e13!

src/lsm/level_data_iterator.zig
sentientwaffle: Does `assert(context.tables.len > 0);` not work for some reason?

sentientwaffle:
~https://github.com/tigerbeetledb/tigerbeetle/pull/875#discussion_r1234693396 and https://github.com/tigerbeetledb/tigerbeetle/pull/875#discussion_r1234693670 (from the previous review) weren't addressed.~

sentientwaffle:
Nevermind, I was mixing up the files :facepalm:

src/lsm/level_data_iterator.zig
chaitanyabhandari: Ah, this is strange. I did reply to this thread but instead of commenting I believe I accidentally hit `Start Review`, so maybe you can't see my comment? It also says `Pending` on the comment. 🤦🏼 

Pasting the comment here:

I only added assert(context.level < constants.lsm_levels);. We can't assert assert(context.tables.len > 0); since we invoke LevelTableValueBlockIteratorType.start() for the case where the Immutable Table is being compacted into Level-0, even if there are no tables in Level 0 that intersect with the Immutable Table.

chaitanyabhandari: I had to hit `Finish Review` to post all of the comments that I made 4 days ago, I believe they'd all visible to you now. 😅

src/lsm/manifest.zig
sentientwaffle: Hmm, I think using `const range_b = ...` is confusing, since `range_b.table_count` actually counts the table in level A too.

chaitanyabhandari: Oh yes, that's one thing I've been meaning to ask as well! What was the rationale behind counting the table in Level A within table_count? I see that the only use is to decide whether the move-table optimization applies, by checking `context.range_b.table_count == 1;`. 

Now that we're maintaining the intersecting tables in `range_b.tables`, maybe could just use the length of that array for the same and remove the `table_count` property altogether?

That way, all information related to Level A would be encapsulated in `table_a ` and Level B would be encapsulated in `range_b`.

sentientwaffle: Yeah, I think that makes sense. Maybe in a follow-up PR though? This on is pretty overloaded as-is.

chaitanyabhandari: Sounds good to me, I'll take this up in a separate PR.

Also, I think I've addressed all pending reviews on this PR, do let me know if there's anything more that I can improve, @sentientwaffle! :D

sentientwaffle (approved):

PR #880, opened 2023-06-14
vopr: liveness checking

matklad:
This allows VOPR to catch more liveness bugs, using fewer tick.

The idea is to split simulator run into two phases, safety and liveness.

The "safety" phase is almost exactly as the previous implementation. The only difference is that we have a (small) tick budget to process a single request, rather than a large budget for the whole run.

The "liveness" phase is different --- we designate a subset of replicas, "core", and make sure they are up & connected. For other replicas, we "freeze" their current failure mode. Eg, if a non-core replica is crashed, it remains crashed forever. This is in contrast to safety phase, where each replica is eventually available.

### DIFF
--- a/src/simulator.zig
+++ b/src/simulator.zig
@@ -239,9 +239,9 @@ pub fn main() !void {
     var simulator = try Simulator.init(allocator, random, simulator_options);
     defer simulator.deinit(allocator);
 
-    // Safety: replicas crash and restart; at any given point in time arbitrary many replicas may be
-    // crashed, but each replica restarts eventually. The cluster must process all requests without
-    // split-brain.
+    // Safety: replicas crash and restart; at any given point in time arbitrarily many replicas may
+    // be crashed, but each replica restarts eventually. The cluster must process all requests
+    // without split-brain.
     const ticks_max_requests = 5_000_000;
     var tick: u64 = 0;
     while (tick < ticks_max_requests) : (tick += 1) {
@@ -379,8 +379,12 @@ pub const Simulator = struct {
     }
 
     pub fn done(simulator: *const Simulator) bool {
-        assert(simulator.requests_sent <= simulator.options.requests_max);
         assert(simulator.core.count() > 0);
+        assert(simulator.requests_sent == simulator.options.requests_max);
+        assert(simulator.reply_sequence.empty());
+        for (simulator.cluster.clients) |*client| {
+            assert(client.request_queue.count == 0);
+        }
 
         for (simulator.cluster.replicas) |*replica| {
             if (simulator.core.isSet(replica.replica)) {
@@ -392,13 +396,6 @@ pub const Simulator = struct {
 
         simulator.cluster.state_checker.assert_cluster_convergence();
 
-        if (!simulator.reply_sequence.empty()) return false;
-        if (simulator.requests_sent < simulator.options.requests_max) return false;
-
-        for (simulator.cluster.clients) |*client| {
-            if (client.request_queue.count > 0) return false;
-        }
-        assert(simulator.requests_replied == simulator.requests_sent);
         return true;
     }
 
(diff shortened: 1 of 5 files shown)


matklad:
This currently uncoveres a real liveness issue with `SEED=11182076433755387547`.

matklad:
One big change since the original implementation: actively shutdown non-core replicas, to prevent a primary to end up outside of the core. 

:thinking: actually, I think I could do better here, let me check this....

matklad:
Ok, I now just special case "primary not in core" case in VOPR and `stdx.unimplemented` there.

src/testing/cluster/network.zig
```diff
@@ -126,6 +134,26 @@ pub const Network = struct {
         network.packet_simulator.tick();
     }
 
+    pub fn transition_to_liveness_mode(network: *Network, core: Core) void {
```
sentientwaffle, line 137: `assert(core.count() > 0)`

src/testing/cluster/state_checker.zig
```diff
@@ -166,16 +166,14 @@ pub fn StateCheckerType(comptime Client: type, comptime Replica: type) type {
...
-                    return false;
-                }
-            }
+            const b = state_checker.commit_mins[replica_index];
+            return a == b;
+        }
 
+        pub fn assert_cluster_convergence(state_checker: *Self) void {
             for (state_checker.commits.items) |commit, i| {
-                assert(commit.replicas.count() == state_checker.replica_count);
```
sentientwaffle, line 178: `assert(commit.replicas.count() > 0);`

src/simulator.zig
```diff
@@ -237,23 +239,59 @@ pub fn main() !void {
     var simulator = try Simulator.init(allocator, random, simulator_options);
     defer simulator.deinit(allocator);
 
-    const ticks_max = 50_000_000;
+    // Safety: replicas crash and restart; at any given point in time arbitrary many replicas may be
+    // crashed, but each replica restarts eventually. The cluster must process all requests without
+    // split-brain.
```
```suggestion
    // Safety: replicas crash and restart; at any given point in time arbitrarily many replicas may be
    // crashed, but each replica restarts eventually. The cluster must process all requests without
    // split-brain.
```

src/simulator.zig
```diff
@@ -338,18 +380,25 @@ pub const Simulator = struct {
...
 
-        if (!simulator.cluster.state_checker.convergence()) return false;
+        simulator.cluster.state_checker.assert_cluster_convergence();
+
         if (!simulator.reply_sequence.empty()) return false;
         if (simulator.requests_sent < simulator.options.requests_max) return false;
 
         for (simulator.cluster.clients) |*client| {
             if (client.request_queue.count > 0) return false;
         }
```
sentientwaffle: Since `done()` is now called after all requests are sent & replies are received, I think these conditions can be asserted.

src/simulator.zig
```diff
@@ -581,6 +684,46 @@ fn random_partition_symmetry(random: std.rand.Random) PartitionSymmetry {
...
+
+    var need = replica_core_count;
+    var left = replica_count;
+    var replica: u8 = 0;
+    while (replica < replica_count + standby_count) : (replica += 1) {
+        if (random.uintLessThan(u8, left) < need) {
+            result.set(replica);
+            need -= 1;
+        }
+        left -= 1;
```
sentientwaffle, line 705: Very elegant!

sentientwaffle:
> One big change since the original implementation: actively shutdown non-core replicas, to prevent a primary to end up outside of the core.

I don't see where this crash takes place; did you change your mind on this?

matklad:
Sorry, I was too cryptic. I did change that to just special casing "primary outside of the core" case over here: 

https://github.com/tigerbeetledb/tigerbeetle/blob/06f59f41e5e41513a91062134fee4410585d3e7f/src/simulator.zig#L275-L277

matklad:
for posterity, the "kill non-core" logic was here: https://github.com/tigerbeetledb/tigerbeetle/commit/62ed7665df97831b8a381e34a27ecd850dda75d8

src/simulator.zig
matklad: :exploding_head: these are two different words!

sentientwaffle (approved):

PR #882, opened 2023-06-17
VSCode debug settings

batiati:
Although we are IDE/Editor agnostic, we can take advantage of some convenience tools for developers who want to use VSCode and extensions.

### This PR

This PR adds settings files, scripts, and other changes required to use [the debugging features](https://code.visualstudio.com/docs/cpp/cpp-debug) provided by VSCode [cpptools](https://marketplace.visualstudio.com/items?itemName=ms-vscode.cpptools) extensions.

- `./vscode` folder contains all settings files required by VSCode.

- `./vscode/extensions.json` simply recommends the required VSCode extensions to be installed by the user. This file is optional.

- `./vscode/launch.json` contains all entry points we might want to run. Note that `"type": "cppdbg"` makes it a debugger entry point, which is shown under the debug tab:
image

- `./vscode/tasks.json` contains tasks that can be automatized. For now, it just contains the build steps required by the debugger launcher.

- `settings.json` contains the setting `"debug.allowBreakpointsEverywhere": true`, which enables placing breakpoints in .zig files, which by default are not recognized by the `cpptools` extension as valid code. 

- `scripts/format_debug_server.sh` is used as a helper to compile and format a replica that will be started by the debugger.
This script can be improved if we want to keep the data file between runs. For now, each time the debugger is started a fresh replica is started.

- `build.zig` minor changes allowing to build the unit tests as an executable, so we can attach the debugger on test cases:
image

### Tips

Inspecting variables works in most cases, except for zig slices, in which only the first element is shown.
Other elements can be inspected from the debugger console by doing simple pointer arithmetic:

`*(slice.ptr + index)` 

Example:

image


matklad:
I guess that's on my to review? Missed the PR initially, will take a look tomorrow!

scripts/format_debug_server.sh
```diff
@@ -0,0 +1,15 @@
...
+if [ ! -d "zig" ]; then
+    scripts/install_zig.sh
+fi
+
+if [ -f "$FILE" ]; then
+    rm $FILE
+fi
+
+./scripts/build.sh install -Dconfig=production
+./tigerbeetle format --cluster=0 --replica=0 --replica-count=1 $FILE
```
matklad: let's add trailing `\n` here and in other files

.vscode/launch.json
```diff
@@ -0,0 +1,27 @@
...
+            "request": "launch",
+            "program": "${workspaceFolder}/tigerbeetle",
+            "args": [ "start", "--addresses=127.0.0.1", "0_0.tigerbeetle.debug" ],
+            "stopAtEntry": false,
+            "cwd": "${workspaceFolder}",
+            "environment": [],
+            "preLaunchTask": "format debug server",
+        },
+        {
+            "name": "debug tests",
```
matklad, line 16: :tada:

.vscode/launch.json
```diff
@@ -0,0 +1,27 @@
...
+            "type": "cppdbg",
+            "request": "launch",
+            "program": "${workspaceFolder}/zig-out/bin/tests",
+            "args": [ "." ],
```
matklad, line 20: why do we need a `"."`?

.vscode/extensions.json
```diff
@@ -0,0 +1,6 @@
+{
+	"recommendations": [
+        "ziglang.vscode-zig",
+        "ms-vscode.cpptools"
```
matklad, line 4: huh, I always tried (and faild) to use [vscode-lldb](https://marketplace.visualstudio.com/items?itemName=vadimcn.vscode-lldb). But yeah, it does seem that sticking with gdb is better.

matklad (dismissed):
We might also mention this in https://github.com/tigerbeetledb/tigerbeetle/blob/main/docs/HACKING.md?

.vscode/launch.json
batiati: What do you think?

1. Configure an exhaustive list of all debuggable artifacts we produce (i.e. simulator, fuzz tests, etc.), and then keep that in sync every time we add a new artifact.

2. Just start minimal AS IS, and add new entries as it is needed.

.vscode/launch.json
batiati: Yes, the test runner expects arguments:

```bash
> ./zig-out/bin/tests
Usage: ./zig-out/bin/tests path/to/zig
thread 197643 panic: Wrong number of command line arguments
/home/batiati/tigerbeetle/zig/lib/std/special/test_runner.zig:20:9: 0x4c6840 in std.special.processArgs (tests)
        @panic("Wrong number of command line arguments");
```

.vscode/extensions.json
batiati: Yes, I also tried lldb on Windows, but GDB just works.
I should have mentioned GDB's dependency in the docs.

.vscode/launch.json
matklad: I'd maybe start with the minimal set. I am also wondering if there's some way to attach debugger without specifying the exact command to lanunch in settings.json every time... 

Like, ideally we'd just run `zig build test`, and have the debugger figure out the process to attach to automatically.... But I don't think there's a nice way to have that.

batiati: Yes, we can configure `request=attach`, and works like this:

image

Idk if we can attach it from the command line, so we can debug since the first instruction executed

batiati:
Can you approve it again please, @matklad?

matklad (approved):

PR #884, opened 2023-06-20
vsr: simplify empty checksum

matklad:
No _real_ motivation here, it just feels awkward to do runtime synchronization here.

The test is right next to `checksum_body_empty`, so, if any of our tests use this constant, the test is guaranteed to be included.


kprotty:
Neat, is there a way to do something like this for `Environment.members()` in `superblock_fuzz.zig` or is that not needed?

matklad:
I guess so, ptal!

kprotty (approved):

PR #885, opened 2023-06-20
VOPR: Fix total tick logging

sentientwaffle:


matklad (approved):

matklad:
When writing this, I was also undecided whether to do two loops, or to squash everything into just one. This is another reason to go for one loop perhaps...

PR #887, opened 2023-06-20
state_machine: completely hide reserved operations

matklad:


src/state_machine.zig
```diff
@@ -226,11 +226,6 @@ pub fn StateMachineType(
         });
 
         pub const Operation = enum(u8) {
-            /// Operations reserved by VR protocol (for all state machines):
-            reserved = 0,
-            root = 1,
-            register = 2,
-
             /// Operations exported by TigerBeetle:
             create_accounts = config.vsr_operations_reserved + 0,
```
matklad, line 230: I don't like how the user is supposed to use literal `reserved` operations. I wonder if there's a nice antonym here, so that we can write 

```
create_accounts = config.vsr_operations_not_reserved + 0,
```

sentientwaffle: What do you mean by 

> user is supposed to use literal `reserved` operations

?

matklad: I mean

```
create_accounts = config.vsr_operations_reserved
```

reads like we use `reserved` value for `create_accounts`. `_reserved` is actually the first not reserved one.

sentientwaffle: Ah, gotcha. What if state machine operations start from 0, and vsr operations start from `config.vsr_operations_reserved`? 
Since `config.vsr_operations_reserved` is 128 that means that all protocol operations have the high-order bit set, and state machine operations don't.

matklad: We want `0` to be reserved, so that we don't confuse accidentally memzeroed op with something we intentionally set

sentientwaffle: `0` could be a separate (unrelated to the constant) builtin (vsr protocol) operation that just happens to be named `reserved`

sentientwaffle: Ah nevermind, I see what you mean, state machine would have to do that too...

matklad: `config.state_machine_operations_min` might be a wordy option here

sentientwaffle (approved):

PR #888, opened 2023-06-21
vopr: relax liveness check

matklad:
If a primary is partially connected to the core, replicas in the core
might be in different views, when a replica partitioned from the primary
ends up in an _earlier_ view. It won't view jump because it doesn't
receive commits from the primary, and it won't DVC itself, because it'll
fail to collect an SVC quorum.


sentientwaffle (approved):
Confirmed to fix a previously-failing state sync seed. Thanks!

PR #890, opened 2023-06-21
state_machine: add missing alignment attributes

matklad:
It's a part of state machine contract that it operates on well-aligned buffers.

### DIFF
--- a/src/tigerbeetle.zig
+++ b/src/tigerbeetle.zig
@@ -21,6 +21,7 @@ pub const Account = extern struct {
     comptime {
         assert(@sizeOf(Account) == 128);
         assert(@bitSizeOf(Account) == @sizeOf(Account) * 8);
+        assert(@alignOf(Account) == 16);
     }
 
     pub fn debits_exceed_credits(self: *const Account, amount: u64) bool {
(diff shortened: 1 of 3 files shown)


src/tigerbeetle.zig
```diff
@@ -19,8 +19,9 @@ pub const Account = extern struct {
     timestamp: u64 = 0,
 
     comptime {
+        assert(@alignOf(Account) == 16);
         assert(@sizeOf(Account) == 128);
-        assert(@bitSizeOf(Account) == @sizeOf(Account) * 8);
+        assert(@alignOf(Transfer) == 16);
```
sentientwaffle: Did you mean to change this? The `bitSizeOf` assert says "we expect that there is no implicit struct padding".

matklad: Nope, just was not sure whether I should assert alignment first or last, and accidetly messed up the whole thing in two different ways while trying to decide :sweat_smile:

sentientwaffle (approved):

matklad:
https://github.com/ziglang/zig/issues/16132

PR #901, opened 2023-06-22
vsr: add reconfiguration request

matklad:
This adds the basic boilerplate to handle reconfiguration:

   - `.reconfigure` operation
   - Corresponding `ReconfigurationRequest` which forms the body of the request.
   - Basic validation logic (the bit where primary makes a judgement call whether reconfiguration is a good idea based on status of the cluster is yet TODO)
   - Skeleton of request handling by the primary (during prepare and commit).

The `ReconfigurationRequest` is somewhat future proof --- it includes facilities for changing the number of standbys/actives, although we don't support actual changes here yet (only permutations are allowed). I _think_ we could re-use the same request shape for both:

* logical reconfiguration (this PR; advance epoch, permute members, but keep the member set the same).
* physical reconfiguration (keep the epoch and active replicas, but add or remove a standby).

### DIFF
--- a/src/stdx.zig
+++ b/src/stdx.zig
@@ -293,6 +293,21 @@ pub inline fn hash_inline(value: anytype) u64 {
     });
 }
 
+/// Returns a copy of `base` with fields changed according to `diff`.
+///
+/// Intended exclusively for table-driven prototype-based tests. Write
+/// updates explicitly in production code.
+pub fn update(base: anytype, diff: anytype) @TypeOf(base) {
+    assert(builtin.is_test);
+    assert(@typeInfo(@TypeOf(base)) == .Struct);
+
+    var updated = base;
+    inline for (std.meta.fields(@TypeOf(diff))) |f| {
+        @field(updated, f.name) = @field(diff, f.name);
+    }
+    return updated;
+}
+
 /// Inline version of Google Abseil "LowLevelHash" (inspired by wyhash).
 /// https://github.com/abseil/abseil-cpp/blob/master/absl/hash/internal/low_level_hash.cc
 inline fn low_level_hash(seed: u64, input: anytype) u64 {
--- a/src/vsr.zig
+++ b/src/vsr.zig
@@ -1119,92 +1119,50 @@ test "ReconfigurationRequest" {
     };
 
     try t.check(r, .ok);
-
-    try t.check(
-        stdx.update(r, .{ .replica_count = 0 }),
-        .replica_count_zero,
-    );
-
-    try t.check(
-        stdx.update(r, .{ .replica_count = 255 }),
-        .replica_count_max_exceeded,
-    );
-
+    try t.check(stdx.update(r, .{ .replica_count = 0 }), .replica_count_zero);
+    try t.check(stdx.update(r, .{ .replica_count = 255 }), .replica_count_max_exceeded);
     try t.check(
         stdx.update(r, .{ .standby_count = constants.standbys_max + 1 }),
         .standby_count_max_exceeded,
     );
-
     try t.check(
         stdx.update(r, .{ .members = Test.to_members(.{ 4, 1, 4, 3 }) }),
         .members_invalid,
     );
-
     try t.check(
         stdx.update(r, .{ .members = Test.to_members(.{ 4, 1, 0, 2, 3 }) }),
         .members_invalid,
     );
-
     try t.check(
         stdx.update(r, .{ .epoch = 0, .members = Test.to_members(.{ 4, 1, 0, 2, 3 }) }),
         .members_invalid,
     );
-
     try t.check(
         stdx.update(r, .{ .epoch = 1, .members = Test.to_members(.{ 4, 1, 0, 2, 3 }) }),
         .members_invalid,
     );
-
-    try t.check(
-        stdx.update(r, .{ .replica_count = 4 }),
-        .members_count_invalid,
-    );
-
-    try t.check(
-        stdx.update(r, .{ .reserved = [_]u8{1} ** 52 }),
-        .reserved_field,
-    );
-
-    try t.check(
-        stdx.update(r, .{ .result = .ok }),
-        .result_must_be_reserved,
-    );
-
-    try t.check(
-        stdx.update(r, .{ .epoch = 0 }),
-        .epoch_in_the_past,
-    );
-
-    try t.check(
-        stdx.update(r, .{ .epoch = 3 }),
-        .epoch_in_the_future,
-    );
-
+    try t.check(stdx.update(r, .{ .replica_count = 4 }), .members_count_invalid);
+    try t.check(stdx.update(r, .{ .reserved = [_]u8{1} ** 52 }), .reserved_field);
+    try t.check(stdx.update(r, .{ .result = .ok }), .result_must_be_reserved);
+    try t.check(stdx.update(r, .{ .epoch = 0 }), .epoch_in_the_past);
+    try t.check(stdx.update(r, .{ .epoch = 3 }), .epoch_in_the_future);
     try t.check(
         stdx.update(r, .{ .members = Test.to_members(.{ 1, 2, 3 }), .replica_count = 2 }),
         .different_replica_count,
     );
-
     try t.check(
         stdx.update(r, .{ .members = Test.to_members(.{ 1, 2, 3, 4, 5 }), .standby_count = 2 }),
         .different_standby_count,
     );
-
     try t.check(
         stdx.update(r, .{ .members = Test.to_members(.{ 8, 1, 2, 3 }) }),
         .different_member_set,
     );
-
     try t.check(
         stdx.update(r, .{ .epoch = 1, .members = Test.to_members(.{ 1, 2, 3, 4 }) }),
         .configuration_applied,
     );
-
-    try t.check(
-        stdx.update(r, .{ .epoch = 1 }),
-        .configuration_conflict,
-    );
-
+    try t.check(stdx.update(r, .{ .epoch = 1 }), .configuration_conflict);
     try t.check(
         stdx.update(r, .{ .members = Test.to_members(.{ 1, 2, 3, 4 }) }),
         .configuration_is_no_op,
--- a/src/vsr.zig
+++ b/src/vsr.zig
@@ -161,6 +161,8 @@ pub const Operation = enum(u8) {
     root = 1,
     /// The value 2 is reserved to register a client session with the cluster.
     register = 2,
+    /// The value 3 is reserved for reconfiguration request.
+    reconfigure = 3,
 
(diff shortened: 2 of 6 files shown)


src/vsr.zig
```diff
@@ -142,6 +142,8 @@ pub const Operation = enum(u8) {
     root = 1,
     /// The value 2 is reserved to register a client session with the cluster.
     register = 2,
+    /// The value 3 is reserved for reconfiguration request.
+    reconfigure = 3,
```
matklad, line 165: Didn't find a better name for the request here, the problem with reconfiguration is that configuration is too generic. On the plus side, this maintains the invariant that all reserved operations start with r!

src/vsr.zig
```diff
@@ -1189,13 +1417,20 @@ test "quorums" {
     }
 }
 
+/// Set of replica_ids of cluster members, where order of ids determines replica indexes.
+///
+/// First replica_count elements are active replicas,
+/// then standby_count standbys, the rest are zeros.
+/// Order determines ring topology for replication.
+pub const Members = [constants.nodes_max]u128;
```
matklad, line 1578: I am wondering if I should wrap this in a struct just to have something to put methods to, instead of doing c-style `members_this`, `members_that`. Decided to stick to C-style, as that's simpler

src/vsr/replica.zig
```diff
@@ -4637,6 +4672,23 @@ pub fn ReplicaType(
...
+        fn primary_prepare_reconfiguration(self: *const Self, request: *Message) void {
+            assert(self.primary());
+            assert(request.header.command == .request);
+            assert(request.header.operation == .reconfigure);
+            assert(request.header.size == @sizeOf(vsr.Header) + @sizeOf(vsr.ReconfigurationRequest));
+            var reconfiguration_request = std.mem.bytesAsValue(
+                vsr.ReconfigurationRequest,
+                request.body()[0..@sizeOf(vsr.ReconfigurationRequest)],
+            );
+            reconfiguration_request.*.result = reconfiguration_request.validate(.{
```
matklad, line 5173: Here, we curiously take advantage of our ability to do some computation during prepare, rather than commit. 

For reconfiguration, primary needs to make a decision whether its good or not during prepare, not during commit; preparing reconfigration takes effect immediately on the primary, as it must stop processing any further requests.

src/stdx.zig
```diff
@@ -216,6 +216,19 @@ pub inline fn hash_inline(value: anytype) u64 {
     });
 }
 
+/// Returns a copy of `base` with fields changed according to `diff`.
+///
+/// Intended exclusively for table-driven prototype-based tests. Write
+/// updates explicitly in production code.
+pub fn update(base: anytype, diff: anytype) @TypeOf(base) {
+    assert(builtin.is_test);
```
sentientwaffle, line 301: Maybe add `assert(@TypeOf(base) == .Struct);` -- that way passing a pointer for `base` will fail rather than mutating.

src/vsr.zig
```diff
@@ -811,6 +817,228 @@ pub const BlockRequest = extern struct {
...
+            } else return .different_member_set;
+        }
+
+        if (std.meta.eql(request.members, current.members.*)) return .no_op;
+
+        return .ok;
+    }
+};
+
+pub const ReconfigurationResult = enum(u32) {
```
sentientwaffle, line 1036: I know we don't do this now for other protocol enums, but what do you think of setting the values explicitly?
e.g. `enum { ok = 0, replica_count_zero = 1, ... }`.

Motivation: Make it extra obvious that renumbering is a breaking protocol change, i.e. new operations should be appended.

src/vsr.zig
```diff
@@ -811,6 +817,228 @@ pub const BlockRequest = extern struct {
...
+    replica_count: u8,
+    /// The new standby count.
+    ///
+    /// At the moment, we require this to be equal to the old count.
+    standby_count: u8,
+    reserved: [10]u8 = [_]u8{0} ** 10,
+    /// The result of this request. Set to zero by the client and filled-in by the primary when it
+    /// accepts a reconfiguration request.
+    result: ReconfigurationResult,
+
```
sentientwaffle, line 973:
```suggestion
comptime {
    assert(@bitSizeOf(ReconfigurationRequest) == @sizeOf(ReconfigurationRequest) * 8);
}
```
Verify there is no implicit padding.

src/vsr.zig
```diff
@@ -811,6 +817,228 @@ pub const BlockRequest = extern struct {
...
+            members: *const Members,
+            epoch: u32,
+            replica_count: u8,
+            standby_count: u8,
+        },
+    ) ReconfigurationResult {
+        assert(valid_members(current.members));
+        assert(member_count(current.members) == current.replica_count + current.standby_count);
+
+        if (request.replica_count == 0) return .replica_count_max_exceeded;
```
sentientwaffle:
```suggestion
        if (request.replica_count == 0) return .replica_count_zero;
```

src/vsr.zig
```diff
@@ -811,6 +817,228 @@ pub const BlockRequest = extern struct {
...
+
+        if (current.epoch == std.math.maxInt(u32)) return .epoch_overflow;
+        if (request.epoch <= current.epoch) return .epoch_in_the_past;
+        if (request.epoch > current.epoch + 1) return .epoch_in_the_future;
+        assert(request.epoch == current.epoch + 1);
+
+        if (request.replica_count != current.replica_count) return .different_replica_count;
+        if (request.standby_count != current.standby_count) return .different_standby_count;
+
+        // Note: neither set can have duplicates at this point.
```
sentientwaffle: Do you mean that it is impossible for there to be duplicates at this point?
Or that if there are duplicates, then it is an error?
(I'm think it is the former, but I think this could be made more clear.)

src/vsr.zig
```diff
@@ -811,6 +817,228 @@ pub const BlockRequest = extern struct {
...
+        stdx.update(r, .{ .epoch = 0 }),
+        .epoch_in_the_past,
+    );
+
+    try t.check(
+        stdx.update(r, .{ .epoch = 3 }),
+        .epoch_in_the_future,
+    );
+
+    t.epoch = 0xFFFF_FFFF;
```
sentientwaffle: `std.math.maxInt(u32)` might make the test's intent slightly more clear, but either is fine.

src/vsr.zig
```diff
@@ -811,6 +817,228 @@ pub const BlockRequest = extern struct {
...
+    /// At the moment, we require this to be equal to the old count.
+    replica_count: u8,
+    /// The new standby count.
+    ///
+    /// At the moment, we require this to be equal to the old count.
+    standby_count: u8,
+    reserved: [10]u8 = [_]u8{0} ** 10,
+    /// The result of this request. Set to zero by the client and filled-in by the primary when it
+    /// accepts a reconfiguration request.
+    result: ReconfigurationResult,
```
sentientwaffle, line 972: I wonder if `0` (set by client) should correspond to a special `ReconfigurationResult` (i.e. not `ok`)?
e.g. `request`.

src/vsr.zig
sentientwaffle: Brainstorming: shape/reshape; roster; roll.

(Roll is a pun — "roll" as in a list of names/items, or "roll" as in circular, i.e. a ring — which is not a reason to use it, but is funny anyway :grin:)

src/vsr.zig
```diff
@@ -811,6 +817,228 @@ pub const BlockRequest = extern struct {
...
+        // Note: neither set can have duplicates at this point.
+        assert(member_count(current.members) == member_count(&request.members));
+        for (current.members) |member_current| {
+            if (member_current == 0) break;
+            for (request.members) |member| {
+                if (member == member_current) break;
+            } else return .different_member_set;
+        }
+
+        if (std.meta.eql(request.members, current.members.*)) return .no_op;
```
sentientwaffle: Maybe use `std.mem.eql` instead, to avoid copying the arrays onto the stack.
```suggestion
        if (std.mem.eql(u128, &request.members, current.members)) return .no_op;
```

src/vsr.zig
matklad: From today's chat with Joran --- an important point here is that this is our public visible interface, rather than just our private API. In the docs, we do want to call this "reconfiguration" as that's the appropriate term here.

src/vsr/replica.zig
```diff
@@ -3344,6 +3346,36 @@ pub fn ReplicaType(
...
+            assert(prepare.header.command == .prepare);
+            assert(prepare.header.operation == .reconfigure);
+            assert(prepare.header.size == @sizeOf(vsr.Header) + @sizeOf(vsr.ReconfigurationRequest));
+            assert(prepare.header.op == self.commit_min + 1);
+            assert(prepare.header.op <= self.op);
+
+            const reconfiguration_request = std.mem.bytesAsValue(
+                vsr.ReconfigurationRequest,
+                prepare.body()[0..@sizeOf(vsr.ReconfigurationRequest)],
+            );
```
sentientwaffle, line 3695:
```suggestion
            );
            assert(reconfiguration_request.result != .reserved);
```

src/vsr.zig
sentientwaffle: Good point; I hadn't considered that.

src/vsr.zig
```diff
@@ -811,6 +817,271 @@ pub const BlockRequest = extern struct {
...
+    );
+
+    try t.check(
+        stdx.update(r, .{ .members = Test.to_members(.{ 1, 2, 3, 4 }) }),
+        .configuration_is_no_op,
+    );
+
+    assert(t.tested.count() < ResultSet.initFull().count());
+    t.tested.insert(.reserved);
+    assert(t.tested.count() == ResultSet.initFull().count());
```
sentientwaffle, line 1173: Nice coverage check!

src/vsr/replica.zig
sentientwaffle: Did this assert get missed? I see it in `primary_prepare_reconfiguration` but I think it applies to `commit_reconfiguration` too.

matklad: indeed, added!

sentientwaffle (dismissed):

matklad:
Status: waiting for @jorangreef to take a look as well

matklad:
Walked through the PR with jorangreef, should be good to go now. We might still want to work on naming for various error variants a bit, but that's related to `ok_created_already` discussion I also have in flight ....

sentientwaffle (approved):
LGTM, though maybe in a follow-up add a comptime assert to `ReconfigurationResult` to ensure that there are no skipped variant values in the enum:

```
    comptime {
        for (std.enums.values(ReconfigurationResult)) |result, index| {
            assert(@enumToInt(result) == index);
        }
    }
```

PR #905, opened 2023-06-23
[Java client] Fix tb_client_release_packet usage.

batiati:
The flaky test was indeed a bug:

The request's `tb_packet_t` was released by the background thread after notifying the application thread about the completion.

In most cases, it worked seamlessly, but under heavy contention (i.e `tasks_qty == concurrency_max`), it was possible for a request to finish and the immediately next one to fail with `ConcurrencyMaxExceeded`, because the background thread was still holding the packet that should have been released when the previous request completed.

~The fix was to expose `release` to the Java side, to be called by the JVM thread just before signaling completion.~

The fix was to change the responsibility of copying and releasing the packet to the JNI side. Java's side endRequest now deals with JVM-owned memory only.

Other minor non-related changes:

- Reduce the number of iterations EchoTest processes.

- Improve the way some tests expect an exception.

- Improve comments.


src/clients/java/src/client.zig
```diff
@@ -332,7 +346,6 @@ const NativeClient = struct {
         _ = client;
 
         var context = @intToPtr(*JNIContext, context_ptr);
-        defer context.release_packet(packet);
```
matklad, line 335: Heh, this is subtle! Tbh, I am a bit unnerved at how we need to release this on the java side, but this indeed seems subtle: we want to release _after_ we've done memcpy, but _before_ we return from the Java method.

I am wondering, can we perhaps push `Request.memcpy` into zig code, so that, when we call `Request.endRequest`, it only gets JVM-owned objects?

src/clients/java/src/main/java/com/tigerbeetle/NativeClient.java
```diff
@@ -49,6 +49,12 @@ public void submit(final Request<?> request) throws ConcurrencyExceededException
         }
     }
 
+    public void release(final long packet) {
+        if (contextHandle == 0L)
```
matklad: Hm, this feels suspicious. If we don't call release, then I think we leak the packet, and than the 

```zig
                    // We need to drain all free packets, to ensure that all
                    // inflight requests have finished.
                    while (self.packets_free.pop() != null) {
```

loop just loops forever? 

Or is it guaranteed that Java-side `close` won't return until all of the callbacks will have executed? But if that's the case, then we can assert here that context is not 0?

src/clients/java/src/client.zig
matklad: Ah, I think another way we can do that is, instead of introducing `release`, split `end_request` into `copy_result` and `end_request`. So that the sequence of calls on the native side is

```
callJvm(“copy_result”)
release(packet)
callJVM(“end_request”)
```

the main thing I want to avoid here is splitting resource management between native and JVM, as that seems error prone.

batiati: That's an interesting bug.

We want to avoid switching between Java and JNI worlds as much as possible because [it interferes with JVM's ability to optimize code](https://developer.ibm.com/articles/j-jni/#choosing). With that in mind, differently from Go and C# clients, in Java we group a lot of things (e.g. acquiring a packet, populating and submitting it) into a single JNI call.

>I am wondering, can we perhaps push Request.memcpy into zig code, so that, when we call Request.endRequest, it only gets JVM-owned objects?

This change only exposes `tb_client_release_packet` (+1 JNI call), following the same logic as the other clients. However, it still keeps `tb_client_acquire_packet` hidden from the JVM, so it would make sense to expose something other than `release` as you suggested. We just have to make sure that we are not introducing too many JNI boundaries.

(...)

This is where we release the packet from the Java side in this PR (replacing `defer` at the zig side):
https://github.com/tigerbeetledb/tigerbeetle/blob/c1cc4b53b81a2341069467ade49617f2431c3f7c/src/clients/java/src/main/java/com/tigerbeetle/Request.java#L141-L145

And these are how we do exactly the same in C# and Go:
https://github.com/tigerbeetledb/tigerbeetle/blob/9703bc2e2a383e39a99bd1e15cb02dfb01282740/src/clients/dotnet/TigerBeetle/Request.cs#L117-L120

https://github.com/tigerbeetledb/tigerbeetle/blob/9703bc2e2a383e39a99bd1e15cb02dfb01282740/src/clients/go/tb_client.go#L178-L179

It's worth noting that this bug always existed, but it was silent before #793  because we didn't have `ConcurrenctExceededException`, instead, we just blocked until a packet became available.

src/clients/java/src/main/java/com/tigerbeetle/NativeClient.java
batiati: Good catch. It's suspicious indeed ... I tried to write a comment, but I failed to explain it in simple terms.

We don't need to check `if (contextHandle == 0L)`  at the Java side for few reasons: 
- It will panic at Zig side if zero.
- It is always called from the JNI side, so it's unlike to be `0` without having segfaulted somewhere else before.
- Even if it's zero due to a `close` bug, we don't want to throw an exception there because it is called from the native thread, and we will have to store this exception to be popped up by the caller thread after completion. This kind of error is better represented with a `@panic`.

The truth is, we do  `if (contextHandle == 0L) return` for the sake of unit tests.

Some tests mock a callback simulator, passing arbitrary packet values which would segfault if the zig code tried to cast to `*tb_packet_t`, like this one:
https://github.com/tigerbeetledb/tigerbeetle/blob/c1cc4b53b81a2341069467ade49617f2431c3f7c/src/clients/java/src/test/java/com/tigerbeetle/AsyncRequestTest.java#L112-L115

Another solution would be a mocked client, but it seemed to be too much trouble to decouple a mock and still have `NativeClient.endRequest` properly asserted.

I agree that we need to `assert` if it happens outside unit tests.
Let me think of something else ...

src/clients/java/src/client.zig
matklad: >And these are how we do exactly the same in C# and Go:

Ok, I think if we do explicit release everywhere, it's probably better that way!

src/clients/java/src/main/java/com/tigerbeetle/NativeClient.java
matklad: So, just to clarify, what would be the behavior on the java side if we do something like this?

```
var client = new Client();
CompletableFuture<CreateAccountResultBatch> accounts  = client.createAccountsAsync(...);
client.close();
var result = accounts.wait();
```

src/clients/java/src/client.zig
batiati: I liked the idea of doing memcpy on the JNI side and then releasing the package.
It would solve both the issues we are talking about in this PR.

src/clients/java/src/main/java/com/tigerbeetle/NativeClient.java
batiati: `client.close()` waits until all requests are finished. So, it will block until `accounts` is processed. After that, `accounts.wait()` will return synchronously. Further requests will fail with `client closed`.

batiati:
I had to update [Jui submodule to fix a simple typo](https://github.com/zig-java/jui/commit/a957d9376db101cb136a62d89341abdfef1529ab) and now our lint may fail due to zig fmt divergences between the zig version used by both repos.

I think it's time to strip out just the parts we use and merge into our repo.

src/clients/java/src/client.zig
```diff
@@ -110,21 +115,59 @@ const ReflectionHelper = struct {
         assert(env.hasPendingException());
     }
 
-    pub fn buffer(env: *jui.JNIEnv, this_obj: jui.jobject) ?[]u8 {
+    pub fn get_send_buffer(env: *jui.JNIEnv, this_obj: jui.jobject) ?[]u8 {
```
matklad, line 118: This doesn't need to be `pub` I think? `pub` is for visibility across files, and we only use this method in this file.

src/clients/java/src/client.zig
```diff
@@ -110,21 +115,59 @@ const ReflectionHelper = struct {
...
+        assert(request_send_buffer_field_id != null);
+        assert(request_send_buffer_len_field_id != null);
 
-        var buffer_obj = env.getField(.object, this_obj, request_buffer_field_id) orelse return null;
+        var buffer_obj = env.getField(.object, this_obj, request_send_buffer_field_id) orelse return null;
         defer env.deleteReference(.local, buffer_obj);
 
-        var buffer_len = env.getField(.long, this_obj, request_buffer_len_field_id);
+        var buffer_len = env.getField(.long, this_obj, request_send_buffer_len_field_id);
         var address = env.getDirectBufferAddress(buffer_obj);
```
matklad: Let's also use `getDirectBufferCapacity` to sanity check that we do in fact have at least `buffer_len` bytes.

src/clients/java/src/main/java/com/tigerbeetle/Request.java
```diff
@@ -157,25 +153,12 @@ void endRequest(final byte receivedOperation, final ByteBuffer buffer, final lon
...
-        assertTrue(source.isDirect(), "Source buffer must be direct");
-
-        final var capacity = source.capacity();
-        assertTrue(capacity > 0, "Source buffer cannot be empty");
-
-        final var copy = ByteBuffer.allocate(capacity);
-        copy.put(source);
-
-        return copy.position(0).asReadOnlyBuffer();
+    byte getOperation() {
```
matklad, line 164: this seems unused?

src/clients/java/src/main/java/com/tigerbeetle/Request.java
```diff
@@ -157,25 +153,12 @@ void endRequest(final byte receivedOperation, final ByteBuffer buffer, final lon
         }
     }
 
-    byte getOperation() {
-        return this.operation.value;
+    void setReplyBuffer(byte[] buffer) {
```
matklad, line 158: comment that we only need this for tests

src/clients/java/src/test/java/com/tigerbeetle/BlockingRequestTest.java
```diff
@@ -483,22 +454,36 @@ public void testFailedCompletion() throws InterruptedException {
 
     }
 
+    private static NativeClient getDummyClient() {
+        var client = NativeClient.initEcho(0, "3000", 1);;
+
+        try {
+            // Calling close() to avoid the native client
+            // trying to release the dummy packet pointer.
+            client.close();
```
matklad: I don't really understand this bit... If we must call close after `initEcho`, should we perhaps also do this in `public final class EchoClient implements AutoCloseable {` which I think is our public api?

src/clients/java/src/client.zig
batiati: `env.getDirectBufferAddress` already calls `getDirectBufferCapacity` to return a slice.
We could explicitly `assert(buffer_len <= address.len)` but Zig does it when we take `address[0..buffer_len]`.

I think the best thing to do is that:
```suggestion
        var buffer: []u8 = env.getDirectBufferAddress(buffer_obj);
```

src/clients/java/src/main/java/com/tigerbeetle/Request.java
batiati: The JNI side uses it.

src/clients/java/src/test/java/com/tigerbeetle/BlockingRequestTest.java
batiati: No, we don't!!! This code is obsolete, `client.close()` was to force the `if (contextHandle == 0L)` that [we discussed](https://github.com/tigerbeetledb/tigerbeetle/pull/905#discussion_r1241285744).

src/clients/java/src/client.zig
batiati: I didn't know about that :+1:, but  I think we should keep using `pub` because it makes it visible that this function is part of the API used by other types and not an internal implementation.

batiati:
Finally, it's passing CI tests 🎉

- Thanks @cb22 for cleaning the self-hosted runner and solving the submodule drama!

- Random CI timeouts were due to insufficient memory on Windows runners, after #790. The Maven test itself consumes >2GB of RAM, making the tests impossible to run in 20 minutes.
Setting `--cache-grid=128MB` for integration tests fixed the issue.

- Still there's a flaky Go test [here](https://github.com/tigerbeetledb/tigerbeetle/actions/runs/5447258487/jobs/9909393762?pr=905), not the focus of this PR.

cc @matklad

src/clients/java/src/client.zig
```diff
@@ -110,19 +115,58 @@ const ReflectionHelper = struct {
...
 
-        var buffer_obj = env.getField(.object, this_obj, request_buffer_field_id) orelse return null;
+        var buffer_obj = env.getField(.object, this_obj, request_send_buffer_field_id) orelse
+            return null;
         defer env.deleteReference(.local, buffer_obj);
 
-        var buffer_len = env.getField(.long, this_obj, request_buffer_len_field_id);
-        var address = env.getDirectBufferAddress(buffer_obj);
+        var buffer_len = env.getField(.long, this_obj, request_send_buffer_len_field_id);
+        var buffer: []u8 = env.getDirectBufferAddress(buffer_obj);
```
matklad, line 128: Ahh! This seem like a suboptimal naming in jui, feels better to leave `getDirectBufferAddress` for `[*]u8`, and name this thing `getDirectBufferSlice` or something.

src/clients/java/src/client.zig
```diff
@@ -110,19 +115,58 @@ const ReflectionHelper = struct {
...
+
+        var reply_buffer_obj = env.newPrimitiveArray(
+            .byte,
+            @intCast(c_int, reply.len),
+        ) catch {
+            // Cannot allocate an array, it's likely the JVM has run out of resources
+            // Printing the buffer size here just to help diagnosing how much memory was required
+            env.describeException();
+            log.err("Unexpected error allocating a new array len={}", .{reply.len});
+            @panic("JNI: Unexpected error creating a new array.");
```
matklad, line 147: Would it be possible to rethrow the exception to the JVM here? Crashing the whole process on OOM feels suboptimal (in contrast, crashing after `setPrimitiveArrayRegion` feels fine, as an excetpion there means a programming error in our code).

Anywhay, this is unrelated to the PR in question, as we had similar behavior before.

matklad (approved):

src/clients/java/src/client.zig
batiati: A new JNI binding is coming 🎉

src/clients/java/src/client.zig
batiati: On the JNI side, the only place that throws exceptions is the initialization (for validating the parameters).
Panicking the entire process feels suboptimal and can be inconvenient for the application's developers.

a) Environment errors, like OOM. We could try to throw an exception, but it can fail due to OOM anyway 🤪, so we don't bother with that and panic;

b) Unexpected conditions in our code are treated with `assert`.

c) We cannot throw exceptions inside the native callback.

(...)

I'll review this strategy for minimizing the places where we panic at library code in general.

batiati:
Thanks @matklad for the detailed review 🏆

PR #920, opened 2023-06-26
Use stable archive url for maven

eatonphil:
The other one only stores the latest minor version.

https://archive.apache.org/dist/maven/maven-3/ this one has them all.

### DIFF
--- a/.github/workflows/clients-java.yml
+++ b/.github/workflows/clients-java.yml
@@ -215,14 +215,14 @@ jobs:
     # self-hosted mac runner. So we grab it manually.
 
     # Grab Maven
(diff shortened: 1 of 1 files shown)


matklad (approved):

matklad:
>    # Although Maven is supposed to be installed in the
    # actions/setup-java@v3 step, it wasn't available in at least the
    # self-hosted mac runner. So we grab it manually.

It's probably worth it to open an issue about that at https://github.com/actions/setup-java/issues

actions/setup-java are provided by the platform we use (ie, provided by GitHub, rather than by a third party), so, long term, relying on that might be less maintainance.

eatonphil:
> It's probably worth it to open an issue about that at https://github.com/actions/setup-java/issues

I agree, I haven't had the time to see what the issue is though so I would not be able to make a good report.

PR #921, opened 2023-06-26
Simplify install and build from source

eatonphil:
Much simpler.

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,116 +38,53 @@ Here are a few key pages you might be interested in:
 
 ## Quickstart
 
-TigerBeetle is easy to run with or without Docker, depending on your
-preference. First, we'll cover running the [Single
-Binary](#single-binary). And below that is how to run [with
-Docker](#with-docker).
+First, download a prebuilt copy of TigerBeetle.
 
-### Single Binary
+On macOS/Linux:
 
-Install TigerBeetle by grabbing the latest release from
-GitHub.
-
-x86_64 and aarch64 builds are available for macOS and Linux. Only
-x86_64 builds are available for Windows.
-
-For example:
-
-```bash
-$ curl -LO https://github.com/tigerbeetle/tigerbeetle/releases/download/2023-03-27-weekly/tigerbeetle-x86_64-linux-2023-03-27-weekly.zip
-$ unzip tigerbeetle-x86_64-linux-2023-03-27-weekly.zip
-$ sudo cp tigerbeetle /usr/local/bin/tigerbeetle # On Windows, add $(pwd) to $env:PATH instead.
-$ tigerbeetle version --verbose | head -n6
-TigerBeetle version experimental
-
-git_commit="55c8fdf1f52c7a174d1bc9d9785cf4e327cae182"
-
-build.mode=Mode.ReleaseSafe
-build.zig_version=0.9.1
+```console
+$ git clone https://github.com/tigerbeetle/tigerbeetle; ./tigerbeetle/bootstrap.sh
 ```
 
-NOTE: This example version is not kept up-to-date. So always check the
-[releases](https://github.com/tigerbeetle/tigerbeetle/releases) page
-for the latest version. You can also find debug builds for each
-arch/OS combo on the release page as well.
+On Windows:
 
-#### Building from source
-
-Or to build from source, clone the repo, checkout a release, and run
-the install script.
-
-You will need POSIX userland, curl or wget, tar, and xz.
-
-```bash
-$ git clone https://github.com/tigerbeetle/tigerbeetle.git
-$ cd tigerbeetle
-$ git checkout 2022-11-16-weekly # Or latest tag
-$ scripts/install.sh
+```console
+$ git clone https://github.com/tigerbeetle/tigerbeetle; .\tigerbeetle\bootstrap.ps1
 ```
 
-Don't worry, this will only make changes within the `tigerbeetle`
-directory. No global changes. The result will place the compiled
-`tigerbeetle` binary into the current directory.
+Want to build from source locally? Add `-build` as an argument to the bootstrap script.
 
 #### Running TigerBeetle
 
 Then create the TigerBeetle data file.
 
-```bash
+```console
 $ ./tigerbeetle format --cluster=0 --replica=0 --replica-count=1 0_0.tigerbeetle
 info(io): creating "0_0.tigerbeetle"...
 info(io): allocating 660.140625MiB...
 ```
 
-And start the server.
+And start the replica.
 
-```bash
+```console
 $ ./tigerbeetle start --addresses=3000 0_0.tigerbeetle
 info(io): opening "0_0.tigerbeetle"...
 info(main): 0: cluster=0: listening on 127.0.0.1:3000
 ```
 
-Now skip ahead to [Use Node as a CLI](#use-node-as-a-cli).
-
-### With Docker
-
-First provision TigerBeetle's data directory.
-
-```bash
-$ docker run -v $(pwd)/data:/data ghcr.io/tigerbeetle/tigerbeetle \
-    format --cluster=0 --replica=0 --replica-count=1 /data/0_0.tigerbeetle
-info(io): creating "0_0.tigerbeetle"...
-info(io): allocating 660.140625MiB...
-```
-
-Then run the server.
-
-```bash
-$ docker run -p 3000:3000 -v $(pwd)/data:/data ghcr.io/tigerbeetle/tigerbeetle \
-    start --addresses=0.0.0.0:3000 /data/0_0.tigerbeetle
-info(io): opening "0_0.tigerbeetle"...
-info(main): 0: cluster=0: listening on 0.0.0.0:3000
-```
-
-Note: if you are on macOS, you will need to call the Docker run
-command with `--cap-add IPC_LOCK` or `--ulimit memlock=-1:-1`. See
-[here](https://docs.tigerbeetle.com/deployment/with-docker#error-systemresources-on-macos) for
-more information.
+Now let's connect to the replica and do some accounting!
 
 ### Use Node as a CLI
 
-Now that you've got the server running with or without Docker, let's
-connect to the running server and do some accounting!
-
 First install the Node client.
 
-```javascript
+```console
 $ npm install tigerbeetle-node
 ```
 
 Then create a client connection.
 
-```javascript
+```console
 $ node
 Welcome to Node.js v16.14.0.
 Type ".help" for more information.
--- a/bootstrap.ps1
+++ b/bootstrap.ps1
@@ -1,0 +1,20 @@
+param(
+    [switch]$build = $false
+)
+
+# Make sure we're in the repository root.
+cd "$PSScriptRoot"
+
+git fetch --tags --force --quiet
+
+$version = $( git tag --sort=committerdate | Select-Object -Last 1 )
+
+if ($build) {
+    git checkout $version
+    .\scripts\install.bat
+} else {
+    curl -o tigerbeetle.zip "https://github.com/tigerbeetle/tigerbeetle/releases/download/$version/tigerbeetle-x86_64-windows-$version.zip"
+    unzip -qo tigerbeetle.zip
+}
+
+echo "Installed at $(pwd)\tigerbeetle.exe: $(.\tigerbeetle.exe version)."
--- a/src/clients/docs_generate.zig
+++ b/src/clients/docs_generate.zig
@@ -765,10 +765,8 @@ const Generator = struct {
 
             mw.header(2, "Start the TigerBeetle server");
             mw.paragraph(
-                \\Follow steps in the repo README to start a **single
-                \\server** [from a single
-                \\binary](/README.md#single-binary) or [in a Docker
-                \\container](/README.md#with-docker).
+                \\Follow steps in the repo README to [run
+                \\TigerBeetle](/README.md#running-tigerbeetle).
                 \\
                 \\If you are not running on port `localhost:3000`, set
                 \\the environment variable `TB_ADDRESS` to the full
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -1,0 +1,45 @@
+#!/bin/sh
+
+# Since this script potentially updates itself, putting the script in
+# braces forces sh to parse the entire script before executing it so
+# it isn't concurrently parsing a script that is changing on disk.
+# https://stackoverflow.com/a/2358432/1507139
+{
+    from_source=""
+    if [ "$1" = "-build" ]; then
+	from_source="t"
+    fi
+
+    set -e
+
+    # Make sure we're in the repository root.
+    cd "$( dirname "$0" )"
+
+    git fetch --tags --force --quiet
+    version="$(git describe --tags "$(git rev-list --tags --max-count=1)")"
+    os="$(uname)"
+    arch="$(uname -m)"
+    if [ "$from_source" = "" ]; then
+	if [ "$os" = "Darwin" ]; then
+	    arch="universal"
+	    os="macos"
+	elif [ "$os" = "Linux" ]; then
+	    os="linux"
+	else
+	    echo "Unsupported OS."
+	    exit 1
+	fi
+
+	curl -sLO "https://github.com/tigerbeetle/tigerbeetle/releases/download/$version/tigerbeetle-$arch-$os-$version.zip"
+	unzip -qo "tigerbeetle-$arch-$os-$version.zip"
+	chmod +x tigerbeetle
+    else
+	git checkout "$version"
+	./scripts/install.sh
+    fi
+
+    echo "Installed at $(pwd)/tigerbeetle: $(./tigerbeetle version)."
+
+    # See https://stackoverflow.com/a/2358432/1507139.
+    exit 0
+}
--- a/src/clients/dotnet/samples/basic/README.md
+++ b/src/clients/dotnet/samples/basic/README.md
@@ -30,10 +30,8 @@ $ dotnet add package tigerbeetle
 
 ## Start the TigerBeetle server
 
-Follow steps in the repo README to start a **single
-server** [from a single
-binary](/README.md#single-binary) or [in a Docker
-container](/README.md#with-docker).
+Follow steps in the repo README to [run
+TigerBeetle](/README.md#running-tigerbeetle).
 
 If you are not running on port `localhost:3000`, set
 the environment variable `TB_ADDRESS` to the full
--- a/src/clients/dotnet/samples/two-phase/README.md
+++ b/src/clients/dotnet/samples/two-phase/README.md
@@ -30,10 +30,8 @@ $ dotnet add package tigerbeetle
 
 ## Start the TigerBeetle server
 
-Follow steps in the repo README to start a **single
-server** [from a single
-binary](/README.md#single-binary) or [in a Docker
-container](/README.md#with-docker).
+Follow steps in the repo README to [run
+TigerBeetle](/README.md#running-tigerbeetle).
 
 If you are not running on port `localhost:3000`, set
 the environment variable `TB_ADDRESS` to the full
--- a/src/clients/go/samples/basic/README.md
+++ b/src/clients/go/samples/basic/README.md
@@ -29,10 +29,8 @@ $ go get github.com/tigerbeetle/tigerbeetle-go
 
 ## Start the TigerBeetle server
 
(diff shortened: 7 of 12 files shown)


README.md
```diff
@@ -38,116 +38,53 @@ Here are a few key pages you might be interested in:
...
-$ git checkout 2022-11-16-weekly # Or latest tag
-$ scripts/install.sh
+```console
+git clone https://github.com/tigerbeetledb/tigerbeetle; ./tigerbeetle/scripts/bootstrap.ps1
 ```
 
-Don't worry, this will only make changes within the `tigerbeetle`
-directory. No global changes. The result will place the compiled
-`tigerbeetle` binary into the current directory.
+Want to build from source? Add `-build` to the bootstrap script arguments:
```
eatonphil: `-build`  as opposed to `--from-source` because Windows deals with `--` flags in a weird way.

README.md
```diff
@@ -38,116 +38,53 @@ Here are a few key pages you might be interested in:
...
 
-You will need POSIX userland, curl or wget, tar, and xz.
-
-```bash
-$ git clone https://github.com/tigerbeetledb/tigerbeetle.git
-$ cd tigerbeetle
-$ git checkout 2022-11-16-weekly # Or latest tag
-$ scripts/install.sh
+```console
+$ git clone https://github.com/tigerbeetledb/tigerbeetle; ./tigerbeetle/scripts/bootstrap.ps1
```
eatonphil: ps1 as opposed to bat since powershell has been around since 2012 and batch scripts are horrifying to work with.

README.md
```diff
@@ -38,116 +38,53 @@ Here are a few key pages you might be interested in:
...
 ```
 
-Don't worry, this will only make changes within the `tigerbeetle`
-directory. No global changes. The result will place the compiled
-`tigerbeetle` binary into the current directory.
+Want to build from source? Add `-build` to the bootstrap script arguments.
```
cb22: Looks like the shell script just checks for `build`? (perhaps we could make it accept `-build` and `--build`)

scripts/bootstrap.sh
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/env sh
```
matklad: `#!/usr/bin/env bash` or `#!/bin/sh` would make more sense to me. If `/usr/bin/env`  is a thing, then `/bin/sh` should also be a thing. Or are there systems where `env` is in the standard location, but `sh` isn't?

scripts/bootstrap.ps1
```diff
@@ -0,0 +1,20 @@
...
+
+# Switch to TigerBeetle directory.
+cd "$PSScriptRoot\.."
+
+git fetch --tags --force --quiet
+
+$version = $( git tag --sort=committerdate | Select-Object -Last 1 )
+
+if ($build) {
+    git checkout $version
```
matklad: I wonder if this would break if `git checkout` overwrites bootstrap.ps1. Windows doesn't like replacing file which is being executed (but maybe this is relaxed for scripts which are not executed themselves?)

README.md
```diff
@@ -38,116 +38,53 @@ Here are a few key pages you might be interested in:
...
-$ sudo cp tigerbeetle /usr/local/bin/tigerbeetle # On Windows, add $(pwd) to $env:PATH instead.
-$ tigerbeetle version --verbose | head -n6
-TigerBeetle version experimental
-
-git_commit="55c8fdf1f52c7a174d1bc9d9785cf4e327cae182"
-
-build.mode=Mode.ReleaseSafe
-build.zig_version=0.9.1
+```console
+$ git clone https://github.com/tigerbeetledb/tigerbeetle; ./tigerbeetle/scripts/bootstrap.sh
```
matklad: That's a big improvement over the current quick start!

Not sure if that's the right long-term shape --- I expect that "installing tigerbeetle on CI" shouldn't involve clonning the repo. But it's not line we should solve each and every problem right now.

README.md
```diff
@@ -38,116 +38,53 @@ Here are a few key pages you might be interested in:
...
-```bash
+```console
 $ ./tigerbeetle start --addresses=3000 0_0.tigerbeetle
 info(io): opening "0_0.tigerbeetle"...
 info(main): 0: cluster=0: listening on 127.0.0.1:3000
 ```
 
-Now skip ahead to [Use Node as a CLI](#use-node-as-a-cli).
-
-### With Docker
```
matklad, line 112: :tada: 

I think my love for docker equals eatonphil's love for WSL :-)

src/clients/docs_generate.zig
```diff
@@ -731,8 +731,7 @@ const Generator = struct {
             mw.paragraph(
                 \\Follow steps in the repo README to start a **single
                 \\server** [from a single
-                \\binary](/README.md#single-binary) or [in a Docker
-                \\container](/README.md#with-docker).
+                \\binary](/README.md#quickstart).
```
matklad: nit: could fit in a single line?

scripts/bootstrap.sh
eatonphil: > If /usr/bin/env is a thing, then /bin/sh should also be a thing.

As far as I understood it this definitely isn't the case. However, I don't personally know of an example. I just had it drilled in that /usr/bin/env sh or /usr/bin/env bash was the safe cross platform way to go.

scripts/bootstrap.ps1
eatonphil: Yes it would probably crash in that scenario. But our bootstrap script wouldn't work already if you copied and pasted it since the directory would also already exist. So the one-liner is really only meant as a one-time thing. Right?

src/clients/docs_generate.zig
eatonphil: Done.

README.md
eatonphil: ✅

README.md
eatonphil: Lots of systems marketing one-liner is not the way you are suggested to install in CI fwiw. We'll want to be in all the package managers eventually.

README.md
eatonphil: Whoops good catch! I adjusted it to `-build`. I am not sure being more flexible is better.

scripts/bootstrap.sh
matklad: Yeah, I also don't really know, https://stackoverflow.com/questions/53116226/what-is-the-recommended-posix-sh-shebang is relevant. 

My current understanding is that it is a difference between sh and bash. 

You can't rely on `/bin/bash`, because not all systems put bash there, and not all system even have bash in the first place. (in particular, my system doesn't have `/bin/bash`). 

However, `/usr/bin/env` and `/bin/sh` are relatively safe. It is not guaranteed that these two paths exist, but, unlike with bash, they actually do exist in practice. Eg, on my system

```
14:26:53|~/p/tb/main|matklad/overfault⚡?
λ exa -lT /usr/
drwxr-xr-x  - root 13 Dec  2022 /usr
drwxr-xr-x  - root 27 Jun 11:45 └── bin
lrwxrwxrwx 65 root 27 Jun 11:45    └── env -> /nix/store/02dr9ymdqpkb75vf0v1z2l91z2q3izy9-coreutils-9.3/bin/env

14:26:57|~/p/tb/main|matklad/overfault⚡?
λ exa -lT /bin
drwxr-xr-x  - root 27 Jun 11:45 /bin
lrwxrwxrwx 75 root 27 Jun 11:45 └── sh -> /nix/store/7xhrcm6a01qri8z2wa4lxzkpc2wfnagm-bash-interactive-5.2-p15/bin/sh
```

Just checked, and https://sh.rustup.rs  and https://install.determinate.systems/nix also use `/bin/sh`. This does seem like will rely on fewer moving parts, but I don't thing ultimately this matters much.

scripts/bootstrap.ps1
matklad: >So the one-liner is really only meant as a one-time thing. Right?

The one-liner itself does checkout. So what I am thinking here is the case where we update script on master, relative to the tagged version. So the one-liner will start executing master version, but then overwrite it with the tagged version.

scripts/bootstrap.sh
matklad: Might also want to  steal `# shellcheck shell=dash` from both of those scripts?

matklad: (although I am maybe 0.7 sure that the nix-one was copied from the rust one... (ah, it even says so in the comment :))

scripts/bootstrap.ps1
eatonphil: While I've seen something along the lines of what you are talking about... I can't remember the rules. Can't delete while open for sure. But it seems like this works? Or what am I forgetting?

image

scripts/bootstrap.sh
eatonphil: I've swapped it to /bin/sh

scripts/bootstrap.ps1
cb22: It seems like on Windows it might be completely fine, but a problem on Linux... https://stackoverflow.com/questions/60692684/edit-powershell-script-while-its-running

(I've seen super weird behaviour first hand, editing shell scripts while they're running)

scripts/bootstrap.sh
```diff
@@ -0,0 +1,33 @@
...
+	os="linux"
+    else
+	echo "Unsupported OS."
+	exit 1
+    fi
+
+    curl -sLO "https://github.com/tigerbeetledb/tigerbeetle/releases/download/$version/tigerbeetle-$arch-$os-$version.zip"
+    unzip -qo "tigerbeetle-$arch-$os-$version.zip"
+else
+    git checkout "$version"
```
cb22: One way that I think is safe to do this on Linux (although, it won't echo the final message...):
```
exec /bin/sh -c "git checkout \"$version\" && ./scripts/install.sh"
```

cb22: (there are some others [here](https://stackoverflow.com/questions/2285403/how-to-make-shell-scripts-robust-to-source-being-changed-as-they-run/2358432#2358432))

README.md
jorangreef: Huge leap forward!

Let's do:

`git clone https://github.com/tigerbeetledb/tigerbeetle; ./tigerbeetle/bootstrap.sh`

To move `bootstrap` to root for its importance.

README.md
```diff
@@ -38,116 +38,53 @@ Here are a few key pages you might be interested in:
 
 ## Quickstart
 
-TigerBeetle is easy to run with or without Docker, depending on your
-preference. First, we'll cover running the [Single
-Binary](#single-binary). And below that is how to run [with
-Docker](#with-docker).
+First, grab a prebuilt copy of TigerBeetle.
```
jorangreef:
```suggestion
First, download a prebuilt copy of TigerBeetle.
```

"grab" is really nice/better for the feel, but we need to hint that this step will download a prebuilt binary rather than build locally.

README.md
jorangreef:
```suggestion
Want to build from source locally? Add `-build` as an argument to the bootstrap script.
```

To emphasize again, that the previous step bootstraps from a binary built remotely.

README.md
```diff
@@ -38,116 +38,53 @@ Here are a few key pages you might be interested in:
...
-directory. No global changes. The result will place the compiled
-`tigerbeetle` binary into the current directory.
+Want to build from source? Add `-build` to the bootstrap script arguments.
 
 #### Running TigerBeetle
 
 Then create the TigerBeetle data file.
 
-```bash
+```console
```
jorangreef, line 61: Slick.

README.md
```diff
@@ -38,116 +38,53 @@ Here are a few key pages you might be interested in:
...
-    start --addresses=0.0.0.0:3000 /data/0_0.tigerbeetle
-info(io): opening "0_0.tigerbeetle"...
-info(main): 0: cluster=0: listening on 0.0.0.0:3000
-```
-
-Note: if you are on macOS, you will need to call the Docker run
-command with `--cap-add IPC_LOCK` or `--ulimit memlock=-1:-1`. See
-[here](https://docs.tigerbeetle.com/deployment/with-docker#error-systemresources-on-macos) for
-more information.
+Now let's connect to the replica and do some accounting!
```
jorangreef, line 75: 🥳

scripts/bootstrap.sh
```diff
@@ -0,0 +1,33 @@
...
+	os="macos"
+    elif [ "$os" = "Linux" ]; then
+	os="linux"
+    else
+	echo "Unsupported OS."
+	exit 1
+    fi
+
+    curl -sLO "https://github.com/tigerbeetledb/tigerbeetle/releases/download/$version/tigerbeetle-$arch-$os-$version.zip"
+    unzip -qo "tigerbeetle-$arch-$os-$version.zip"
```
jorangreef: Since this will overwrite any duplicates, is there any safety/sanity checking that we need to do, to verify that we're in the correct cwd before executing this step?

jorangreef: It seems far-fetched, but is there any scenario where this might not err on the side of doing the safe thing?

scripts/bootstrap.sh
jorangreef: Would `set -e` also be needed here?

cb22:
I wonder if GitHub does transparent HTTP gzip compression if the payload isn't already compressed. That could allow us to drop a requirement for either gzip / unzip binaries, while still benefiting from the download size improvement.

If not I still think we should perhaps switch to gz, either curl might be able to decompress it inline, or a system is more likely to have gzip than unzip (not sure about Windows, however). I'll look in to this a bit more.

jorangreef:
Great idea. One way to check is to do `curl -v` to see the HTTP headers when downloading from GitHub.

scripts/bootstrap.sh
cb22: We'll also need to `chmod +x` here if using `unzip`.

cb22:
Unfortunately they don't auto-gzip, I think the best would be to just switch to `gzip`ing our artifacts, and use `gunzip` in these scripts (Windows has `tar` support built in from the same time as `curl` - which I presume includes gunzip)

README.md
eatonphil: Can do.

scripts/bootstrap.sh
eatonphil: Hmm... Showing the end message would be good though...

README.md
eatonphil: ✅ pulled in locally

README.md
eatonphil: pulled in locally, thanks!

scripts/bootstrap.sh
eatonphil: > Since this will overwrite any duplicates, is there any safety/sanity checking that we need to do, to verify that we're in the correct cwd before executing this step?

There's one `cd` in the script to make sure we're in the TigerBeetle root. So there shouldn't be anything more to check.

eatonphil: > We'll also need to chmod +x here if using unzip.

Done!

eatonphil:
> (Windows has tar support built in from the same time as curl - which I presume includes gunzip)

@cb22 Windows `curl` is a bedeviled alias for Invoke-WebRequest. Its parameters are nothing like real curl.

scripts/bootstrap.sh
cb22: Yeah - I think there are better methods in the SO post I linked above. We should be able to put it in a block and add the echo, then exit after the block to be safe (or wrap the whole script in a block)

cb22:
> Windows curl is a bedeviled alias for Invoke-WebRequest

Ah, I thought it was a real curl for a bit now - https://learn.microsoft.com/en-us/virtualization/community/team-blog/2017/20171219-tar-and-curl-come-to-windows - but I don't run Windows much, so can't vouch for this!

eatonphil:
> Ah, I thought it was a real curl for a bit now - https://learn.microsoft.com/en-us/virtualization/community/team-blog/2017/20171219-tar-and-curl-come-to-windows - but I don't run Windows much, so can't vouch for this!

While I remember seeing this too, I run latest Windows and when I looked up curl -v or whatever it showed Invoke-WebRequest.

scripts/bootstrap.sh
eatonphil: :+1: done thanks!

cb22:
@eatonphil just thought of two annoying edge cases unfortunately:
* We publish a broken release that has no binaries (like we currently have)
* A user tries to grab a release in between the release being created and the binaries being uploaded...

Not sure what a good solution would be at the moment...

eatonphil:
Yeah I was thinking about them too because of yesterday. As soon as we want to we get to mark releases as not pre-release but release. Then we can use the github API to grab the latest prerelease. Although using the Github API is a potential area for failure because if the user somehow messes up and keeps retrying, maybe they then get rate limited. But it's something we could do first off and then back off if it doesn't work to just latest release.

Or alternatively we could also just build it for you if there is not release. Or we could list through releases and keep checking their assets until we find one that exists.

However, we could leave that to a second pass!

eatonphil:
@cb22 ping for review or thoughts :)

bootstrap.ps1
```diff
@@ -0,0 +1,20 @@
...
+
+git fetch --tags --force --quiet
+
+$version = $( git tag --sort=committerdate | Select-Object -Last 1 )
+
+if ($build) {
+    git checkout $version
+    .\scripts\install.bat
+} else {
+    curl -o tigerbeetle.zip "https://github.com/tigerbeetledb/tigerbeetle/releases/download/$version/tigerbeetle-x86_64-windows-$version.zip"
```
cb22: In both cases, should we `rm` this file when done? Perhaps using `New-TemporaryFile` / `mktemp` too? Alternatively if we switch to `gzip` we can `| gunzip` directly (not sure about PowerShell though...)

bootstrap.ps1
```diff
@@ -0,0 +1,20 @@
...
+
+# Make sure we're in the repository root.
+cd "$PSScriptRoot"
+
+git fetch --tags --force --quiet
+
+$version = $( git tag --sort=committerdate | Select-Object -Last 1 )
+
+if ($build) {
+    git checkout $version
```
cb22, line 13: I'm not sure on PowerShell quoting rules - does this need to be quoted?

cb22:
@eatonphil 

> Or we could list through releases and keep checking their assets until we find one that exists.

I think we should do this - we could stop after the first 5 and error out if we don't find anything? I don't think it can be left to a second pass because even though the window for failure might be small (I think ~10 minutes) - we do merge quite often currently, and it's not a great experience if it fails on your first try!

bootstrap.ps1
eatonphil: I'm not sure. But I tested it does work and there won't be special characters because we control the file name.

eatonphil:
> Or we could list through releases and keep checking their assets until we find one that exists.

> I think we should do this - we could stop after the first 5 and error out if we don't find anything? I don't think it can be left to a second pass because even though the window for failure might be small (I think ~10 minutes) - we do merge quite often currently, and it's not a great experience if it fails on your first try!

Doing this makes the script much more complicated and makes me nervous about it breaking.

I think it would be best if we just had somewhere that marked what the last good release was? So we could just check one place rather than duplicating retry logic in shell and powershell?

eatonphil:
> I think it would be best if we just had somewhere that marked what the last good release was? So we could just check one place rather than duplicating retry logic in shell and powershell?

Also, if we did this, we might not need to download the entire repo just to get the tags!

Could we have the CI autocommit a successful build (since we've just built up all that logic too) to either this repo or some other repo? And then we can just check github usercontent / raw files to see the last good release and just download that?

cb22:
> Doing this makes the script much more complicated and makes me nervous about it breaking.

Very true.

> Could we have the CI autocommit a successful build (since we've just built up all that logic too) to either this repo or some other repo? And then we can just check github usercontent / raw files to see the last good release and just download that?

I _think_ we can fix this (and get an added benefit of not having incomplete releases on the release page) by creating our release as draft and then when the validation passes, marking it as non-draft. I'll check this out.

cb22:
It looks like GH will only actually cut the tag when the release is marked as a non-draft, too!

`This tag will be created from the target when you publish this release.`

eatonphil:
That's amazing. Then I think this is ready to go @cb22 ?

cb22 (approved):

cb22:
> That's amazing. Then I think this is ready to go @cb22 ?

Yep! Some small improvements but we can do those in a second pass - this is a massive win already!

PR #923, opened 2023-06-26
Raise Java timeouts

eatonphil:
One reason things keep failing on main is these timeouts.

https://github.com/tigerbeetledb/tigerbeetle/actions/runs/5379438558

### DIFF
--- a/.github/workflows/clients-java.yml
+++ b/.github/workflows/clients-java.yml
@@ -176,7 +176,7 @@ jobs:
         docker run -v $(pwd)/test:/test -w /test ${{ matrix.image }} java -cp "./*" Main.java
 
   samples_integration_test:
-    timeout-minutes: 20
+    timeout-minutes: 30
 
     if: inputs.version == ''
     needs: build_and_publish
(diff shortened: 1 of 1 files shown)


eatonphil:
Raised the timeouts and the build failed on timeouts. Raised them again.

image

cb22:
@eatonphil I think there's something else that's causing these tests to flake?

batiati:
It looks like it's stuck during IntegrationTest.
I think it isn't a matter of timeout since the time-consuming part of the IntegrationTest is spinning up servers, and we don't have the same problem with dotnet tests, which do basically the same thing.

It's most likely a bug (maybe related to #905) or something else that closes the server during the test and makes the client wait indefinitely for replies.

I'll try to break down IntegrationTest into individual tests to isolate which one is causing that.

eatonphil:
Cool, I agree raising timeouts doesn't make sense here. Closing.

batiati:
Hey @eatonphil, I think it's good evidence to reopen this PR:
https://github.com/tigerbeetle/tigerbeetle/actions/runs/5464536722/jobs/9946705261?pr=957

This job just exceeded the timeout but wasn't stuck or failing.

cb22:
@batiati do you think 30 min should be reasonable, or do we need more?

batiati:
I think it's related to #961, so once it's fixed 30 min should be good enough.

PR #924, opened 2023-06-26
vopr: disable livenss checking with many faults

matklad:
With the addition of core/liveness checking logic, it became possible for cluster to fail to converge if a prepare is available only outside the core.

In theory, we _could_ patch FaultAtlas to not put too many faults into the core, but that's a bit complicated, and also makes faults less interesting.

Instead, here we allow FaultAtlas to fault whatever, and then check, at the end, that, if we are not live, there indeed exists a prepare which is faulty in core.

We don't assert that we have this prepare outside of the core for two reasons:

* Some replicas might be down and might not have a journal to consult at all.
* Ideally, we'd allow simulator to corrupt all copies of prepare, to have extra checking that we remain unavailable in this case

However, we do verify that the state_checker is aware about the prepare, which means that we check that the prepare was committed at one point, which implies that we had appropriate durability at some point.

### DIFF
--- a/src/simulator.zig
+++ b/src/simulator.zig
@@ -274,26 +274,28 @@ pub fn main() !void {
         if (simulator.done()) {
             break;
         }
-    } else {
-        if (simulator.primary_outside_of_core()) {
-            stdx.unimplemented("repair requires reachable primary");
-        }
-
-        output.info("no liveness, final cluster state (core={b})", .{simulator.core.mask});
-        simulator.cluster.log_cluster();
-        output.err("you can reproduce this failure with seed={}", .{seed});
-        fatal(.liveness, "no state convergence", .{});
     }
 
-    assert(simulator.done());
-
-    const commits = simulator.cluster.state_checker.commits.items;
-    const last_checksum = commits[commits.len - 1].header.checksum;
-    for (simulator.cluster.aofs) |*aof, replica_index| {
-        if (simulator.core.isSet(replica_index)) {
-            try aof.validate(last_checksum);
+    if (simulator.done()) {
+        const commits = simulator.cluster.state_checker.commits.items;
+        const last_checksum = commits[commits.len - 1].header.checksum;
+        for (simulator.cluster.aofs) |*aof, replica_index| {
+            if (simulator.core.isSet(replica_index)) {
+                try aof.validate(last_checksum);
+            } else {
+                try aof.validate(null);
+            }
+        }
+    } else {
+        if (simulator.core_missing_primary()) {
+            stdx.unimplemented("repair requires reachable primary");
+        } else if (simulator.core_missing_prepare()) |header| {
+            output.warn("no liveness, op={} is not available in core", .{header.op});
         } else {
-            try aof.validate(null);
+            output.info("no liveness, final cluster state (core={b})", .{simulator.core.mask});
+            simulator.cluster.log_cluster();
+            output.err("you can reproduce this failure with seed={}", .{seed});
+            fatal(.liveness, "no state convergence", .{});
         }
     }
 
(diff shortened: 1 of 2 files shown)


src/simulator.zig
```diff
@@ -457,6 +459,49 @@ pub const Simulator = struct {
         return false;
     }
 
+    // Returns a header for a prepare which can't be repaired by the core due to storage faults.
+    //
+    // When generating a FaultAtlas, we don't try to protect core from excessive errors. Instead,
+    // if the core gets stuck, we verify that this is indeed due to storage faults.
+    pub fn core_missing_prepare(simulator: *const Simulator) ?vsr.Header {
```
sentientwaffle, line 466: (Out of scope for this PR:)

This is sufficient in the `main` branch b/c grid corruption is disabled in the vopr.

In the state sync branch, is possible that we have the prepare in our WAL, but one or more grid blocks (required for either committing or compaction) are corrupt. So we will additionally need a `core_missing_grid_block`.

src/simulator.zig
```diff
@@ -457,6 +459,49 @@ pub const Simulator = struct {
...
+
+        for (simulator.cluster.replicas) |replica| {
+            if (simulator.core.isSet(replica.replica) and
+                replica.replica < simulator.cluster.replica_count)
+            {
+                const slot = replica.journal.slot_for_op(missing_header.op);
+                if (replica.journal.prepare_inhabited[slot.index] and
+                    replica.journal.prepare_checksums[slot.index] == missing_header.checksum and
+                    !replica.journal.faulty.bit(slot))
+                {
```
sentientwaffle: Simplified to:

```suggestion
                if (replica.journal.has_clean(missing_header)) {
```

Unlike the original version, this expects `!dirty` (in addition to `!faulty`) but I think that's actually good.

src/simulator.zig
```diff
@@ -457,6 +459,49 @@ pub const Simulator = struct {
...
+            }
+        }
+
+        if (missing_op == null) return null;
+
+        const missing_header = simulator.cluster.state_checker.header_with_op(missing_op.?);
+
+        for (simulator.cluster.replicas) |replica| {
+            if (simulator.core.isSet(replica.replica) and
+                replica.replica < simulator.cluster.replica_count)
```
sentientwaffle: Instead of the `replica.replica < simulator.cluster.replica_count`, what about checking `!replica.standby()`?
Or iterating `for (simulator.cluster.replicas[0..simulator.cluster.replica_count])`.

src/simulator.zig
matklad: Yeah, I think we should check that a block is in repair queue of all replicas (similarly to faulty check  here, we expect the entirety of the core to be aware that they in fact are missing the said block)

src/simulator.zig
matklad: Huh, forgot that we have `.standby`...

This is actually something I am struggling with respect to our code style, what's our stance wrt [the law of demeter](https://www.tedinski.com/2018/12/18/the-law-of-demeter.html)? If I want to peek through several abstraction layers, should I just go and reach inside through a layer of fields, or should I introduce a helper function?

In the `has_clean` case, I copied the code from `on_request_prepare`, which open-coded the logic for "we might have a specific prepare, even if we are not sure about" in `replica.zig`, and I missed that we have similar (and better) logic in journal.zig.

sentientwaffle: > what's our stance wrt [the law of demeter](https://www.tedinski.com/2018/12/18/the-law-of-demeter.html)

I don't think we have a consistent stance.
When writing tests I am fine doing whichever is simpler (which is usually just reaching through layers of fields).

sentientwaffle (approved):

src/simulator.zig
matklad: >When writing tests I am fine doing whichever is simpler (which is usually just reaching through layers of fields).

Yeah, same here, though for me rational is not so much simplicity, as the desire to keep production code as simple as possible (eg, rather than buggify _in the source_, we prefer to buggify in the simulator, to keep the source pristine).

PR #930, opened 2023-06-27
[Java client] Request's completion life-cycle changes. 

batiati:
change the responsibility of copying and releasing the packet to the JNI side. Java's side endRequest now deals with JVM-owned memory only.

This fix also closes the related #905.

Other minor non-related changes:

- Reduce the number of iterations EchoTest processes.

- Improve the way some tests expect an exception.


batiati:
This PR is just a CI test to fix the submodule dependency, that will be closed in favor of #905 (which has much more detailed infos/discussion)

batiati:
Closing in favor of #905.

PR #933, opened 2023-06-28
VOPR: Handle recovering_head core replica

sentientwaffle:
If one or more replicas in the core is in `status=recovering_head`, then they may stuck – unable to complete a view change.

### DIFF
--- a/src/simulator.zig
+++ b/src/simulator.zig
@@ -289,6 +289,8 @@ pub fn main() !void {
     } else {
         if (simulator.core_missing_primary()) {
             stdx.unimplemented("repair requires reachable primary");
+        } else if (simulator.core_missing_quorum()) {
+            output.warn("no liveness, core replicas cannot view-change", .{});
         } else if (simulator.core_missing_prepare()) |header| {
             output.warn("no liveness, op={} is not available in core", .{header.op});
         } else {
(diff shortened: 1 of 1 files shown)


sentientwaffle:
I haven't seen a seed in `main` that hits this exact case, but I think it's just rare.
I did hit it in the state sync branch, since `recovering_head` is much more common there.

src/simulator.zig
```diff
@@ -459,6 +461,28 @@ pub const Simulator = struct {
         return false;
     }
 
+    /// The core contains at least a view-change quorum of replicas. But if one or more of those
+    /// replicas are in status=recovering_head (due to corruption or state sync), then that may be
+    /// insufficient.
```
matklad, line 466: Hm, this is interesting. I think for `.recovering_head` after a crash this is a fair game --- the replica isn't considered fully started until it recovered its head (VRR paper says "interesting, recovery protocol requires F+2 replicas to communicate"). 

But for the state sync, this doesn't seem to be entirely well-justified. Replica can get into `.recovering_head` due to state sync in normal processing without crashing if the network is laggy. So, ideally, if some replica is in this condition due to state sync, we should recover all-right. 

I think  a scenario here is: 3 replicas lag, 3 replicas race ahead, and then their leader crashes. As the three laggards are in `.recovering_head`, the cluster will be stuck, although only one replica has really crashed. 

This feels somewhat realistic, tbh, but also can be tackled separately (perhaps a `.recovering_head` replica can participate in a view change with a completely empty DVC, which could be treated as a bunch of NACKs if it's checkpoint is too old?)

matklad (approved):

PR #934, opened 2023-06-28
vsr: avoid resonance during repair

matklad:
The attached seed shows an interesting failure mode where in a cluster of three a replica is missing two prepares, and it always asks for op=192 from replica=1, and for op=193 from replica=2. This is because we round-robin through replicas during repairs, and, with 2 prepares and 2 other replicas, we get a perfect resonance.

As a simple, robust solution we now instead pick an other replica at random.

Perhaps it would be better to try to account for connectivity information here, try to repair from the closest replicas first, and fall back to random only when our neighbors fail us.

However, at the moment we don't have real connectivity information, and can rely only on closeness of replica's ids. However, that information is already somewhat accounted for by our ring-replication topology, and, given that this is a repair, we already _know_ that it doesn't work.

SEED: 13510902032428962439

### DIFF
--- a/src/vsr/replica.zig
+++ b/src/vsr/replica.zig
@@ -362,10 +362,6 @@ pub fn ReplicaType(
         /// (grid.read_faulty_queue.count>0)
         grid_repair_message_timeout: Timeout,
 
-        /// Used to provide deterministic entropy to `choose_any_other_replica()`.
-        /// Incremented whenever `choose_any_other_replica()` is called.
-        choose_any_other_replica_ticks: u64 = 0,
-
         /// Used to calculate exponential backoff with random jitter.
         /// Seeded with the replica's index number.
         prng: std.rand.DefaultPrng,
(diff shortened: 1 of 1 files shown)


src/vsr/replica.zig
```diff
@@ -2700,20 +2696,17 @@ pub fn ReplicaType(
...
-                const replica = @mod(
-                    self.replica + self.choose_any_other_replica_ticks,
-                    self.replica_count,
-                );
-                if (replica == self.replica) continue;
-                return @intCast(u8, replica);
-            }
-            unreachable;
+            const shift = self.prng.random().intRangeAtMost(u8, 1, self.replica_count - 1);
+            const other_replica = @mod(self.replica + shift, self.replica_count);
```
sentientwaffle: :thinking: If we are a standby, this might not send a request to the (active) replica at `self.replica % self.replica_count`.

matklad: Agonized between writing a retry loop, which naturally handles both casese but looks non-minimal, writing an explicit `if (self.standby())` which looks ugly, and writing cryptic modulo arithmetic, which happens to work for standby case, picked the last solution in the end. 

To bad it's hard to assert that we can pick _any_ replica here.

sentientwaffle (approved):

PR #939, opened 2023-06-30
vsr: improve CTRL availability

matklad:
I was looking into `nack=true present=true` case in our CTRL protocol implementation, and noticed two bugs reducing availability.

First, `nack=true present=true` case is actually possible: the replica received a prepare and it is writing it to disk (slot is dirty). In this case, the replica can nack prepare as it hasn't sent prepare_ok, but, at the same time, the prepare is actually available on the replica.

We correctly account for this case when producing a DVC, but we miss this case when we receiving a DVC. There, we assume that nacks means that the prepare is not actually present.

This is fixed by counting nacks and copies separately.

The second issue is that we compute present bits incorrectly. We only look whether we have the header, and don't check the prepare itself.

### DIFF
--- a/src/vsr/replica.zig
+++ b/src/vsr/replica.zig
@@ -3522,9 +3522,17 @@ pub fn ReplicaType(
                         nacks.set(i);
                     }
 
-                    if (journal_header != null and journal_header.?.checksum == header.checksum and
-                        !faulty)
+                    // Presence bit: the prepare is on disk, is being written to disk, or is cached
+                    // in memory. These conditions mirror logic in `on_request_prepare` and imply
+                    // that we can help the new primary to repair this prepare.
+                    if ((self.journal.prepare_inhabited[slot.index] and
+                        self.journal.prepare_checksums[slot.index] == header.checksum) or
+                        self.journal.writing(header.op, header.checksum) or
+                        self.pipeline_prepare_by_op_and_checksum(header.op, header.checksum) != null)
                     {
+                        if (journal_header != null) {
+                            assert(journal_header.?.checksum == header.checksum);
+                        }
                         maybe(nacks.isSet(i));
                         present.set(i);
                     }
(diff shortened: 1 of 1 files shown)


src/vsr/replica.zig
```diff
@@ -3522,11 +3522,19 @@ pub fn ReplicaType(
...
+                    // is being written to disk, or is cached in memory.
                     if (journal_header != null and journal_header.?.checksum == header.checksum and
                         !faulty)
                     {
-                        maybe(nacks.isSet(i));
-                        present.set(i);
+                        if ((self.journal.prepare_inhabited[slot.index] and
+                            self.journal.prepare_checksums[slot.index] == header.checksum) or
+                            self.journal.writing(header.op, header.checksum) or
+                            self.pipeline_prepare_by_op_and_checksum(header.op, header.checksum) != null)
```
matklad: This is interesting. I _think_ we want this condition to be more or less symmetric to that in `on_request_prepare`. But we don't have `self.journal.writing(header.op, header.checksum)` case in `on_request_prepare`; there, we only try cache and disk, we don't try in-flight writes. 

I am wondering though, can it be the case that the prepare is being written, but is absent from the cache? Maybe `self.journal.writing(header.op, header.checksum)` condition is fully covered by `self.pipeline_prepare_by_op_and_checksum(header.op, header.checksum) != null` ?

sentientwaffle: Cache is updated in `write_prepare()`:

```zig

            // Criteria for caching:
            // - The primary does not update the cache since it is (or will be) reconstructing its
            //   pipeline.
            // - Cache uncommitted ops, since it will avoid a WAL read in the common case.
            if (self.pipeline == .cache and
                self.replica != self.primary_index(self.view) and
                self.commit_min < message.header.op)
            {
                const prepare_evicted = self.pipeline.cache.insert(message.ref());
                if (prepare_evicted) |m| self.message_bus.unref(m);
            }
```

So I think there _are_ cases where we write a prepare which is not in our cache.

src/vsr/replica.zig
```diff
@@ -3522,11 +3522,19 @@ pub fn ReplicaType(
                         nacks.set(i);
                     }
 
+                    // Presence bit: We have the header, and the prepare is on disk,
+                    // is being written to disk, or is cached in memory.
```
sentientwaffle: Maybe add a little more detail?:

```suggestion
                    // Presence bit: We have the header, and the prepare is on disk,
                    // is being written to disk, or is cached in memory.
                    // If the new primary needs to repair this prepare, we can assist them.
```

sentientwaffle: Or maybe that last line is misleading :thinking: it sort of implies that the check shouldn't consider the `journal_header`.

matklad: :thinking: thinking even more about it, it seems like we indeed shouldn't consider `journal_header`? In the situation where:

* we have the header from persisted DVC/SV
* but we don't have journal_header because it is marked faulty
* but we none-the-less have the prepare itself (so, the slot is faulty due to redundant header being faulty)

we should signal that we are capable of reparing the prepare. 

I think it goes even further than this:

If we we have prepare matching the header in the DVC then:

* either the journal header is null
* or the journal header matches the header in the DVC

Journal header can't be newer than the header in the DVC (because DVC must included freshest info) and it can't be older than the prepare (because we update journal header before writting the prepare)

sentientwaffle (approved):

PR #941, opened 2023-06-30
Renamed grid.alloc_block() to grid.allocate_block() #927

ItIsOHM:

PR #942, opened 2023-06-30
Rename grid.alloc_block() to grid.allocate_block()

### LINKED ISSUE #927: Rename grid.alloc_block() to grid.allocate_block()
Per TigerStyle, “Do not abbreviate [...] names” and the reason we use `snake_case`, to encourage more descriptive, complete names.

eatonphil:
Switched to this PR instead of #941 so that @ItIsOHM still gets credit but so that I could run `zig/zig fmt .` on top and collapse @ItIsOHM 's commits into one rather than including the merge commit in #941.

Closes #927

### DIFF
--- a/src/lsm/compaction.zig
+++ b/src/lsm/compaction.zig
@@ -43,7 +43,7 @@ const constants = @import("../constants.zig");
 
 const stdx = @import("../stdx.zig");
 const GridType = @import("grid.zig").GridType;
-const alloc_block = @import("grid.zig").alloc_block;
+const allocate_block = @import("grid.zig").allocate_block;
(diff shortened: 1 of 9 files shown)


batiati (approved):
That's great @ItIsOHM. Thank you for contributing to TigerBeetle!
It reads a lot better now, doesn't it?  😉

ItIsOHM:
> That's great @ItIsOHM. Thank you for contributing to TigerBeetle! It reads a lot better now, doesn't it? 😉

It sure does! Thanks a ton for guiding me so well throughout this issue, really learnt so many new things :D

PR #945, opened 2023-07-03
Cleanup root, remove Tracy and JUI from submodules

cb22:
Remove a few things from our root directory:

- `.gitattributes`
    - Moved under `src/` and `scripts/` 
- `Dockerfile` and `.dockerignore`
    - Moved under `tools/docker`
- `.vscode/`
    - Moved under `tools/vscode`. Not sure if it belongs here, but a `contrib` or similar dir is quite common.
- `.gitmodules`
    - We now use commands in `build.zig` to check out Tracy / JUI as needed, if they don't exist.
- `demos/`
    - These were really more demos explaining the performance history and demos of design decisions of TigerBeetle, than current demos (which exist under `src/demos`) and could be a source of confusion... I've deleted them for now. We could bring them back in a subfolder or somewhere else however.


scripts/.gitattributes
```diff
@@ -0,0 +1,2 @@
+* text=auto eol=lf
+*.bat text=auto eol=crlf
```
matklad, line 2: I think we recently added `.ps1` for quick start. I wonder if we should migrate all `.bat`s to `.ps1` maybe? Abstractly, I think it's better to use one scripting langauge per system, and `.ps1` feels like what _should_ be used on windows. Maybe there are _specific_  reasons to stick with `.bat` though!

matklad:
> Moved under tools/docker
> Moved under tools/vscode. Not sure if it belongs here, but a contrib or similar dir is quite common.

I wouldn't want to create a bunch of special-case folders with one or two files. At the same time, I very much in favor of removing cruft from top-level. I wonder if a more forward-compatible choice would be `etc/` folder (in the original meaning of et cetera). Basically, https://matklad.github.io/2021/05/12/design-pattern-dumping-ground.html. 


> Moved under tools/vscode. Not sure if it belongs here, but a contrib or similar dir is quite common.

It seems to me that, if we just move this to some subfolder, we might as well just delete stuff, and it's unlikely that somebody would notice this. OTOH, there's not much harm in this being available somewhere...

build.zig
```diff
@@ -555,6 +555,11 @@ fn link_tracer_backend(
                 exe.linkSystemLibrary("dbghelp");
                 exe.linkSystemLibrary("ws2_32");
             }
+
+            // We might need to clone Tracy, if it doesn't exist
+            const git_clone = exe.builder.addSystemCommand(&.{ "sh", "-c", "test -d tools/tracy ||" ++
```
matklad: I'd rather we avoid `sh` here -- it doesn't seem to be _fundamentally_ required, as we are full-fledged Zig process at this point and can do whatever. It might be shorter to write in the short-term, but long term it would be faster to start building a library for scripting in Zig perhaps.

OTOH, we _already_ use "sh" in the build.zig, so this doesn't introduce a new dependency. I guess I'll look into expunging sh from build.zig after this is merged, along the lines of 

https://github.com/matklad/xshell

scripts/.gitattributes
cb22: +1 for moving to `ps1`. I believe PowerShell has been supported by default on Windows since 2012!

build.zig
cb22: Yeah, it was my laziness coming through here :laughing:. All I'm using it for is to check if a dir exists, which is simple enough in pure Zig. I'll refactor this and the other instance to do that.

eatonphil: I have some wrappers for running commands cross platform in shells in src/clients/shutil.zig already as well.

cb22: Thanks for the heads up @eatonphil - I've modified `file_or_directory_exists` to no longer need an allocator, and using it now

cb22:
> I wouldn't want to create a bunch of special-case folders with one or two files. At the same time, I very much in favor of removing cruft from top-level. I wonder if a more forward-compatible choice would be etc/ folder (in the original meaning of et cetera). Basically, https://matklad.github.io/2021/05/12/design-pattern-dumping-ground.html.

That idea is the same as what I was thinking / using the `tools/` folder for now (which isn't the right name). I think `etc/` might cause too much confusion given its prevailing usage however, given I've seen quite a few projects that use that for config samples and the like.

> It seems to me that, if we just move this to some subfolder, we might as well just delete stuff, and it's unlikely that somebody would notice this. OTOH, there's not much harm in this being available somewhere...

It's certainly up for debate - personally, I think if something isn't part of the core experience, but would be widely used (we don't want things bit-rotting in our repo - or at least used by someone on our team) and is otherwise unobtrusive, it can have a place.

README.md
```diff
@@ -288,17 +288,6 @@ $ scripts/benchmark.sh
 
 Read [docs/HACKING.md](docs/HACKING.md).
 
-## Performance Demos
```
eatonphil, line 291: Joran wanted these in last time I made changes to the readme. We're all cool with removing this now? (I am, but just asking.)

build.zig
```diff
@@ -2,9 +2,14 @@ const std = @import("std");
 const builtin = @import("builtin");
 const CrossTarget = std.zig.CrossTarget;
 const Mode = std.builtin.Mode;
+const file_or_directory_exists = @import("src/clients/shutil.zig").file_or_directory_exists;
```
eatonphil, line 5: If this is the only function you're pulling from that file and it's only three lines, maybe we should just copy the code instead?

eatonphil:
Mostly looks good to me!

build.zig
```diff
@@ -648,6 +660,17 @@ fn java_client(
     bindings.setMainPkgPath("src");
     const bindings_step = bindings.run();
 
+    // We might need to clone JUI, if it doesn't exist
+    if (!file_or_directory_exists("src/clients/java/lib/jui") and !git_clone_jui) {
+        const git_clone = b.addSystemCommand(&.{ "git", "clone", "https://github.com/zig-java/jui.git", "src/clients/java/lib/jui" });
```
batiati: It feels so weird to inform the commit hash here, but at the same time, it's the safer choice for reproducible builds.

For now, we can also check out the branch "zig-0.9.1", that dependency will be removed soon anyway.
```suggestion
        const git_clone = b.addSystemCommand(&.{ "git", "clone", "-b", "zig-0.9.1", "https://github.com/zig-java/jui.git", "src/clients/java/lib/jui" });
```

.gitignore
```diff
@@ -12,5 +12,9 @@ simulator
 zig/
 zig-cache/
 zig-out/
+tools/tracy
+
+# We have a recommended vscode config under tools/vscode, which you can symlink to .vscode
+.vscode/
```
batiati, line 18: :+1:

build.zig
matklad: A draft for further reducing sh dependencies https://github.com/tigerbeetledb/tigerbeetle/pull/950.

scripts/.gitattributes
jorangreef: Good call also re: one scripting language per system.

README.md
jorangreef: Yes, it's been good so far to keep these around for posterity. I know people who've been running them or using them to get into Zig, e.g. how to work with io_uring.

Jamie actually had a comment on his blog from way back that he liked how we had preserved them—usually these kinds of things get lost along the way.

At the same time, I'm happy to prune these. They require ongoing maintenance. And it makes sense not to confuse people by lumping them in the `demos` directory.

Let's say adios!

build.zig
```diff
@@ -555,6 +560,13 @@ fn link_tracer_backend(
                 exe.linkSystemLibrary("dbghelp");
                 exe.linkSystemLibrary("ws2_32");
             }
+
+            // We might need to clone Tracy, if it doesn't exist
```
jorangreef: Per TigerStyle, comments are sentences and end with a full stop or colon (same for the JUI comment below):

```suggestion
            // We might need to clone Tracy, if it doesn't exist.
```

src/clients/shutil.zig
```diff
@@ -219,10 +219,11 @@ pub fn binary_filename(arena: *std.heap.ArenaAllocator, parts: []const []const u
     return file_name.items;
 }
 
-pub fn file_or_directory_exists(arena: *std.heap.ArenaAllocator, f_or_d: []const u8) bool {
-    _ = std.fs.cwd().realpathAlloc(arena.allocator(), f_or_d) catch {
+pub fn file_or_directory_exists(f_or_d: []const u8) bool {
+    var file = std.fs.cwd().openFile(f_or_d, .{}) catch {
         return false;
     };
```
jorangreef: Which errors are we ignoring here? Would be good to handle these here explicitly/exhaustively.

Otherwise, if the file does exist, but we encounter an error, we might then do something dangerous.

cb22: There are quite a few - I've rather made it explicitly `panic` if it encounters something other than `FileNotFound` now.

build.zig
```diff
@@ -2,9 +2,14 @@ const std = @import("std");
 const builtin = @import("builtin");
 const CrossTarget = std.zig.CrossTarget;
 const Mode = std.builtin.Mode;
+const file_or_directory_exists = @import("src/clients/shutil.zig").file_or_directory_exists;
 
 const config = @import("./src/config.zig");
 
+// We need to ensure we only add a clone dependency once.
+var git_clone_tracy = false;
+var git_clone_jui = false;
```
matklad, line 11: :thinking: this is not quite right I think. Zig build system is a two-phase system:

* there's "configure" phase which means running `pub fn build` and constructing a graph of steps
* then there's "execution" phase, where a subset of the step graph is run in the dependency order

I think what happens here is mixing the two phases: you check the flag in the configure phase, but you actually need to check if the dir exists at the execution phase. 

I think that'd be easy to do with https://github.com/tigerbeetledb/tigerbeetle/pull/950, so I suggest maybe merging as is, and I clean this right after that?

matklad (dismissed):

build.zig
cb22: > I think what happens here is mixing the two phases: you check the flag in the configure phase, but you actually need to check if the dir exists at the execution phase.

Yes, it's exactly that - this is just a workaround :) Fixing it with #950 after sounds good to me!

matklad (approved):

src/clients/shutil.zig
jorangreef: Good call. And now we fail fast at least if we get something unexpected. We only return `false` if it definitely does not exist.

PR #878, opened 2023-06-14
Client built into the binary

eatonphil:
Happy to address feedback! The current Node CLI doesn't work on Windows so would be great to have a real one.

Check out the updated README to see it in action: https://github.com/tigerbeetledb/tigerbeetle/blob/pe/cli2/README.md#using-the-cli.

This has unit tests for the parser and integration tests for the whole thing.

PR #949, opened 2023-07-03
Pull in basic kcov support

cb22:
This brings in `kcov` from #878.

To use it locally: run with `COV=1 zig build test`

On CI:
* See the Summary page for the Core pipeline, it'll have both a diff to main of coverage, and absolute coverage, as well as HTML reports for download.
* It'll give inline annotations on PRs about lines that aren't covered.


build.zig
```diff
@@ -187,13 +187,28 @@ pub fn build(b: *std.build.Builder) void {
...
         const unit_tests = b.addTest("src/unit_tests.zig");
+        unit_tests.addPackage(vsr_package);
         unit_tests.setTarget(target);
         unit_tests.setBuildMode(mode);
         unit_tests.addOptions("vsr_options", options);
         unit_tests.step.dependOn(&tb_client_header_generate.step);
         unit_tests.setFilter(test_filter);
         link_tracer_backend(unit_tests, tracer_backend, target);
 
+        // Turn on test coverage if COV env var is not blank.
```
eatonphil, line 195: This looks familiar. :D

eatonphil: @cb22 could you also pull in my updates to HACKING.md that talk about how to use this?

build.zig
```diff
@@ -190,13 +190,28 @@ pub fn build(b: *std.build.Builder) void {
...
+        if (std.process.getEnvVarOwned(b.allocator, "COV")) |coverage| {
+            b.allocator.free(coverage);
+            unit_tests.setExecCmd(&[_]?[]const u8{
+                "kcov",
+                "--clean",
+                "--include-path=src/",
+                "kcov-output",
```
eatonphil, line 202: Should we add this directory to the .gitignore as well?

build.zig
cb22: Done!

build.zig
cb22: Yep, it is there

eatonphil (approved):

PR #950, opened 2023-07-03
build: avoid sh dependency to build tigerbeetle

matklad:


matklad:
Draft: I need to merge this with `client/shutil.zig` before merging, but I am curious whether we wan't something like this at all (a stateful `Shell` with an arena which is a one-stop-shop for scripting).

build.zig
```diff
@@ -150,9 +152,8 @@ pub fn build(b: *std.build.Builder) void {
         lint_tigerstyle_step.dependOn(&run_cmd.step);
 
         // lint_shellcheck
-        const lint_shellcheck = b.addSystemCommand(&.{ "sh", "-c", "command -v shellcheck >/dev/null" ++
-            " || (echo -e '\\033[0;31mPlease install shellcheck - https://www.shellcheck.net/\\033[0m' && exit 1)" ++
-            " && shellcheck $(find ./src ./scripts -type f -name '*.sh')" });
+        var lint_shellcheck = allocator.create(ShellcheckStep) catch unreachable;
+        lint_shellcheck.* = ShellcheckStep.init(allocator);
         const lint_shellcheck_step = b.step("lint_shellcheck", "Run shellcheck on **.sh");
```
matklad: I don't like how enterprisey doing anything custom in build.zig looks. I _think_ this all can be condensed down to something like

```zig
        // lint_shellcheck
        const lint_shellcheck_step = custom_step(b, "lint_shellcheck", "Run shellcheck on **.sh", allocator, struct {
            fn make(allocator: std.mem.Allocator, step: *std.build.Step) anyerror!void {
                const self = @fieldParentPtr(ShellcheckStep, "step", step);

                var shell = try Shell.init(self.gpa);
                defer shell.deinit();

                if (!shell.exec_status_ok("shellcheck --version")) {
                    shell.echo("{ansi-red}Please install shellcheck - https://www.shellcheck.net/{ansi-reset}", .{});
                    return error.NoShellcheck;
                }

                const scripts = try shell.find(.{
                    .where = &.{ "src", "scripts" },
                    .ends_with = ".sh",
                });

                try shell.exec("shellcheck {scripts}", .{
                    .scripts = scripts,
                });
            }
        });
```

Haven't gone this path just yet --- there's a bunch of breaking changes in build.zig for 0.11, so it might be premature to build thick abstractions here.

eatonphil:
I've gone the route of implementing a cross-platform shell like this for the build system before as well (in DataStation) and I'm not sure I'm a huge fan of it versus just using builtin sh and powershell commands as I've exposed in src/clients/shutil.zig. For example, your shell doesn't (yet) support command combinations with ";" or "&&". I also use `src/clients/shutil.zig` to execute `if`s when validating shell scripts work in docs. (This aspect isn't cross-platform, but it does need access to `sh`.) And of course you can express combined commands at the Zig layer but you just end up being fairly limited in your own shell implementation and working having to do strange workarounds.

Also dealing purely with strings as the API here is a little more cumbersome than dealing with arguments as arrays but idk.

It's easy to see a lot of this is opinion so I don't really want to get into an opinion debate. Just mentioning some thoughts. :)

matklad:
I do feel myself having some fairly strong thoughts here :-) I see bash as a pragmatic choice for many UNIX-first project, but our context is slightly different --- while we are pretty tightly bound to UNIX (and even to Linux, via io_uring), we also have zero dependency policy. And `sh` is a huuge dependency. So I am eyeing this from perspective of reducing dependencies and owning all our bugs, rather from niceness and ergonomics perspective. 

Though, to me 

```zig
        const scripts = try shell.find(.{
            .where = &.{ "src", "scripts" },
            .ends_with = ".sh",
        });

        try shell.exec("shellcheck {scripts}", .{
            .scripts = scripts,
        });
```

reads better than

```sh
shellchek $(find ./src ./scripts -type f -name '*.sh')
```

as I don't have to worry about substitution behavior. I don't know how the shell version behaves if file names contain `\s` or `\n`, while the Zig's version behavior is relatively bullet-proof. Similarly, I don't have to worry that the caller might not have `sh` or `find`.

Though, the situation with testing client readmes is different. There, the property we want to test is that copy-pasting commands into user's _shell_ works. In that context, concatenating all commands into string and submitting it to `sh -c` feels like an appropriate solution. So, yeah, having a closer look, I don't think that suggested `Shell` would be a good solution there.

cb22:
> but I am curious whether we want something like this at all

I'm a fan - I think there's a gap between writing a full blown shell script, and these little inline hack-jobs I've been doing that this serves well. To me, the downside in being less concise is worth the upsides (less platform dependency, less worrying about error handling and edge cases etc)

I think if we get to the point where we need too much more logic, we can reevaluate how we're doing things (perhaps we're working at the wrong level of abstraction or it should be a standalone script)

jorangreef:
Great to move in the direction of eliminating shell dependencies.

> while we are pretty tightly bound to UNIX

And this will also broaden as we make Windows and macOS first class production targets (and then new OS'es in 15 years' time, yet to be born). So any shell dependencies will explode dimensionality, as more platforms are supported, compared to having everything in Zig.

Positively, also to emphasize that there's also the safety risk. Eventually, a bash script in a project will do `rm -rf` on production data. For example, our benchmark script used to unilaterally do this on the target data file. And this probability is multiplied by the number of platforms we support. We only have to get it wrong on Windows or Linux, and not everyone on the team will know the intricacies of platform-specific scripting.

Of course, this kind of thing can happen with Zig scripts too. But writing in the same language as the codebase should help to increase the probability that TigerStyle and explicitness are kept front of mind.

src/Shell.zig
```diff
@@ -0,0 +1,303 @@
+//! Collection of utilities for scripting: an in-process sh+coreutils combo.
```
cb22: > I need to merge this with client/shutil.zig before merging

Are you still planning to do so?

cb22: Should we name this `shell.zig` rather to keep the convention? Eg, `static_allocator.zig` which isn't generic.

src/Shell.zig
```diff
@@ -0,0 +1,303 @@
...
+}
+
+const ansi = .{
+    .red = "\x1b[0;31m",
+    .reset = "\x1b[0m",
+};
+
+/// Prints formatted input to stderr.
+/// Newline symbol is appended automatically.
+/// ANSI colors are supported via `"{ansi-red}my colored text{ansi-reset}"` syntax.
```
cb22: Love this!

src/Shell.zig
```diff
@@ -0,0 +1,303 @@
...
+        .Exited => |code| if (code != 0) return error.NonZeroExitStatus,
+        else => return error.CommandFailed,
+    }
+}
+
+/// Returns `true` if the command executed successfully with a zero exit code and an empty stderr.
+///
+/// The intended use-case is sanity-checking that an executable is present, by running
+/// `my-tool --version`.
+pub fn exec_status_ok(shell: *Shell, comptime cmd: []const u8, args: anytype) !bool {
```
cb22: Seems very specific - and some utils might output on `stderr`... We could just use `exec` at the caller, since it seems we return `false` for any `exec` error here? Or, we could make this just wrap `exec` above and return `true` on void, `false` on `NonZeroExitStatus` and pass through the error otherwise.

src/Shell.zig
```diff
@@ -0,0 +1,303 @@
...
+    return switch (res.term) {
+        .Exited => |code| code == 0 and res.stderr.len == 0,
+        else => false,
+    };
+}
+
+/// Run the command and return its stdout.
+///
+/// Returns an error if the command exists with a non-zero status or a non-empty stderr.
+pub fn exec_stdout(shell: *Shell, comptime cmd: []const u8, args: anytype) ![]const u8 {
```
cb22: Bit of a train of thought as I go through this haha, what if we make a single `exec` that returns a struct, and always captures `stderr`, `stdout` and exit code?

src/Shell.zig
```diff
@@ -0,0 +1,303 @@
...
+        .Exited => |code| if (code == 0) return error.NonZeroExitStatus,
+        else => return error.CommandFailed,
+    }
+
+    if (res.stderr.len > 0) return error.NonEmptyStderr;
+    return res.stdout;
+}
+
+/// Returns current git commit hash as an ASCII string.
+pub fn git_commit(shell: *Shell) ?[40]u8 {
```
cb22: On the fence (in general) if we should bubble the error rather than null for these? Caller can always do eg `git_commit() catch null`

(but in shell tools like this that can be an ergonomic penalty)

src/Shell.zig
```diff
@@ -0,0 +1,303 @@
...
+}
+
+/// Returns current git tag.
+pub fn git_tag(shell: *Shell) ?[]const u8 {
+    const stdout = shell.exec_stdout("git describe --tags", .{}) catch return null;
+    return stdout;
+}
+
+const Argv = std.ArrayList([]const u8);
+fn format_argv(argv: *Argv, comptime cmd: []const u8, args: anytype) !void {
```
cb22: Perhaps there's a better name for this - I found myself wondering if it were vulnerable to shell injection / quote attacks when I first saw a string being used - but all we're doing is splitting and formatting.

build.zig
```diff
@@ -978,3 +949,75 @@ fn client_docs(
...
+    const Options = struct {
+        repo: []const u8,
+        tag: []const u8,
+        path: []const u8,
+    };
+
+    fn add(b: *std.build.Builder, options: Options) *GitCloneStep {
+        var result = b.allocator.create(GitCloneStep) catch unreachable;
+        result.* = .{
+            .step = std.build.Step.init(.custom, "run shellcheck", b.allocator, GitCloneStep.make),
```
cb22: "run shellcheck" -> "run git clone"

build.zig
```diff
@@ -153,9 +152,7 @@ pub fn build(b: *std.build.Builder) void {
         lint_tigerstyle_step.dependOn(&run_cmd.step);
 
         // lint_shellcheck
-        const lint_shellcheck = b.addSystemCommand(&.{ "sh", "-c", "command -v shellcheck >/dev/null" ++
-            " || (echo -e '\\033[0;31mPlease install shellcheck - https://www.shellcheck.net/\\033[0m' && exit 1)" ++
-            " && shellcheck $(find ./src ./scripts -type f -name '*.sh')" });
+        const lint_shellcheck = ShellcheckStep.add(b);
```
cb22, line 154: :tada:

src/Shell.zig
matklad: >Are you still planning to do so?

No, upon a closer look, `shutil` does a different thing --- it concatenates strings into a single .sh or .bat script, for the purpose of testing sh/bat fragments from our readme.

src/Shell.zig
matklad: Yeah, this is obviously better, fixed!

src/Shell.zig
matklad: Yeah, injection-proofness is one of the core ideas here, but it could be made more explicit with better naming/comments, added both.

src/Shell.zig
matklad: Yeah, the slightly sprawling design here is intentional. I've added an explanatory comment to the start of the file. There's already a fully-flexible, give-caller-all-control API in std.ChildProcess. The problem with that is that the caller has too much control over the redirection, status codes, etc, and it's easy to forget something. 

So here we rather try to orient API towards specific use-cases, so that the caller doesn't have any knobs, and needs to just select the correct function.


Now, how exactly do we turn the knobs is an open question (should we check if stderr is zero? would it be a good idea to echo the command itself in `exec`?), but I think we can adjust that as we go.

src/shell.zig
```diff
@@ -0,0 +1,315 @@
...
+        } else {
+            @compileError("Unsupported argument type");
+        }
+    }
+
+    const args_provided = std.meta.fields(@TypeOf(cmd_args)).len;
+    if (args_used != args_provided) @compileError("Unused argument");
+}
+
+test "format argv" {
```
cb22: Sorry one more nit :D
```suggestion
test "shell: expand argv" {
```

src/unit_tests.zig
```diff
@@ -42,4 +42,6 @@ test {
     _ = @import("vsr/superblock_quorums.zig");
 
     _ = @import("aof.zig");
+
+    _ = @import("Shell.zig");
```
cb22: (I lied when I said one more)
```suggestion
    _ = @import("shell.zig");
```

cb22 (approved):
🚀

PR #951, opened 2023-07-03
vsr: reconfigure request for simulator and client

matklad:
This adds reconfiguration request to the client and simulator. 

For the client, we go with a strongly typed API -- `.requset` and `.request_reconfigure` are different functions, and they accept callbacks with different signatures. I am not entirely sure that's a good idea --- as an alternative, we could use more "dynamically-typed" API, where it is just a request, and its on the user to cast the result to ReconfigurationResult. 

Note that `request_reconfigure` accepts a `Message` argument. It _could_ aquire the message itself internally, given that it's going to memcpy ReconfigurationRequest anyway, but, as acquiring a message is a failable operation (there are only so many messages available), pushing that to the caller seems advantageous.

### DIFF
--- a/src/vsr/client.zig
+++ b/src/vsr/client.zig
@@ -26,9 +26,16 @@ pub fn Client(comptime StateMachine_: type, comptime MessageBus: type) type {
                 operation: StateMachine.Operation,
                 results: []const u8,
             ) void;
+
+            pub const ReconfigureCallback = *const fn (
+                user_data: u128,
+                result: vsr.ReconfigurationResult,
+            ) void;
+
             user_data: u128,
+
             // Null iff operation=register.
-            callback: ?Callback,
+            callback: ?union { request: Callback, reconfigure: ReconfigureCallback },
             message: *Message,
         };
 
@@ -205,6 +212,7 @@ pub fn Client(comptime StateMachine_: type, comptime MessageBus: type) type {
             message_body_size: usize,
         ) void {
             assert(@enumToInt(operation) >= constants.vsr_operations_reserved);
+            assert(message_body_size <= constants.message_body_size_max);
 
             self.register();
             assert(self.request_number > 0);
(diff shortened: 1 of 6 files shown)


matklad:
Notably, this does not expose reconfiguration in the non-zig clients.

src/vsr/client.zig
```diff
@@ -205,65 +217,70 @@ pub fn Client(comptime StateMachine_: type, comptime MessageBus: type) type {
...
-            });
-            if (self.request_queue.full()) assert(self.messages_available == 0);
+        pub fn request_reconfigure(
+            self: *Self,
+            user_data: u128,
+            callback: Request.ReconfigureCallback,
+            reconfiguration: *const vsr.ReconfigurationRequest,
+            message: *Message,
+        ) void {
+            self.register();
```
sentientwaffle, line 240: Why is this `.register()` necessary if `raw_request` is going to call it?

src/vsr/client.zig
```diff
@@ -205,65 +217,70 @@ pub fn Client(comptime StateMachine_: type, comptime MessageBus: type) type {
...
-            const was_empty = self.request_queue.empty();
-
-            self.request_number += 1;
-            self.request_queue.push_assume_capacity(.{
-                .user_data = user_data,
-                .callback = callback,
-                .message = message.ref(),
-            });
-            if (self.request_queue.full()) assert(self.messages_available == 0);
+        pub fn request_reconfigure(
```
sentientwaffle, line 233: > For the client, we go with a strongly typed API -- .requset and .request_reconfigure are different functions, and they accept callbacks with different signatures. I am not entirely sure that's a good idea --- as an alternative, we could use more "dynamically-typed" API, where it is just a request, and its on the user to cast the result to ReconfigurationResult.

My first response is that it feels a bit inconsistent to use a separate API for just reconfiguration. 

:thinking: Does the `Client` needs to "know" about reconfigurations, the same way that it needs to "know" about `register`s? i.e. when a client sends/receives a reconfiguration does the client itself need to take some special action that it wouldn't take for a different message?

matklad:
Closing, will revisit some time in the future, but, with our elf-powered upgrade infra, there's no _burning_ need for recongfig.

PR #953, opened 2023-07-04
Build a single universal macOS binary for aarch64 and x86_64

cb22:


cb22:
```
~/g/t/tigerbeetle> file tigerbeetle                                                             (universal-macos|?) 14:54
tigerbeetle: Mach-O universal binary with 2 architectures: [x86_64:\012- Mach-O 64-bit x86_64 executable, flags:<NOUNDEFS|DYLDLINK|TWOLEVEL|PIE|HAS_TLV_DESCRIPTORS>] [\012- arm64:\012- Mach-O 64-bit arm64 executable, flags:<NOUNDEFS|DYLDLINK|TWOLEVEL|PIE|HAS_TLV_DESCRIPTORS>]
```

I've attached a sample built locally, since these commands would only get run on publish. This should work on both x86-64 and aarch64 macOS, but I can't test that easily:
[tigerbeetle.zip](https://github.com/tigerbeetledb/tigerbeetle/files/11949487/tigerbeetle.zip)

jorangreef:
This is great @cb22.

IIRC, the motivation is to increase the probability of success for anyone downloading a TB binary for macOS? To reduce the amount of platform detection we need to do?

Would be good to document this _why_ inline with the build step, to motivate.

cb22:
> IIRC, the motivation is to increase the probability of success for anyone downloading a TB binary for macOS? To reduce the amount of platform detection we need to do?

Exactly that! I've added it as a comment giving the full motivation:
```
  # Split out the macOS building from the step above, so we can build a universal binary.
  # This means 2 less binaries (one for debug, one for release) on the release page,
  # less chance of confusion by a macOS user downloading the wrong arch, and less
  # to do in a quick start script (no need to detect system arch).
```

eatonphil:
Dumb question: is the resulting binary twice the size?

cb22:
It is - but that works out around 4MB (2MB a pop) which I think is OK?

eatonphil:
Yeah that's not bad!

.github/workflows/release_artifacts.yml
```diff
@@ -87,6 +87,56 @@ jobs:
...
+      if: ${{ inputs.dry-run != 'true' }}
+      run: |
+        export GIT_TAG="${{ needs.version.outputs.version }}" && \
+        export RELEASE_ID=`curl -H 'Authorization: Bearer ${{ secrets.GITHUB_TOKEN }}' "https://api.github.com/repos/tigerbeetledb/tigerbeetle/releases/tags/${GIT_TAG}" | jq '.id'` && \
+        export BINARY="tigerbeetle" && \
+        zip -9 tigerbeetle-universal-macos-${GIT_TAG}${{ matrix.debug }}.zip $BINARY && \
+        curl --fail \
+          -H "Authorization: Bearer ${{ secrets.GITHUB_TOKEN }}" \
+          -H "Content-Type: application/zip" \
+          --data-binary @./tigerbeetle-universal-macos-${GIT_TAG}${{ matrix.debug }}.zip \
```
eatonphil, line 137: Let's just drop universal then from the uploaded file then? If they don't need to think about it let's just not say anything?

cb22: Hmmm, thought now that removing it might cause confusion as to if they're getting a native binary or not? (I'm not a Mac User - if it's the usual to have a universal binary, then I think dropping it is fine)

jorangreef: Yes, I think it may prevent confusion to say `universal` then it's explicit. Otherwise, users might wonder if we have their arch covered.

jorangreef: It should be `macos-universal`, putting the qualifier last.

cb22: @jorangreef that was my initial thought, but we have the rest as eg `tigerbeetle-aarch64-linux.zip` or `tigerbeetle-x86_64-windows.zip` (`{arch}-{platform}`) since that's what Zig build uses?

jorangreef: Ah, thanks for the clarification. I guess Zig is ahead of us here. Let's leave `macos` as the qualifier then.

eatonphil (approved):

PR #954, opened 2023-07-04
build: remove extra allocator

matklad:
It seems that using `b.allocator` should be fine, semantic-wise, and also faster, as that's an arena:

https://ziggit.dev/t/how-to-allocate-in-build-zig/1044/2

The problem with

```
var gpa = std.heap.GeneralPurposeAllocator(.{}){};
```

is that's a local variable, and so will be destroyed by the time the steps are run. We don't need allocation in steps right now, but we'll soon need it!

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -13,13 +13,11 @@ var git_clone_jui = false;
 pub fn build(b: *std.build.Builder) void {
     const target = b.standardTargetOptions(.{});
     const mode = b.standardReleaseOptions();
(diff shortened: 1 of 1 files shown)


cb22:
> https://ziggit.dev/t/how-to-allocate-in-build-zig/1044/2

:format(webp):no_upscale()/cdn.vox-cdn.com/uploads/chorus_asset/file/23265504/Spider_Man_meme.jpg)

cb22 (approved):

PR #955, opened 2023-07-05
Shuffle scripts around

cb22:
- Scripts that are only used by CI now live under `.github/ci`
- Deleted `upgrade_ubuntu_kernel.sh` as the latest Ubuntu 20.04 point releases have 5.15 natively
- Deleted `vr_state_enumerate` - it wasn't referenced anywhere


eatonphil:
Is shellcheck now still going to find these?

cb22:
It will now!

eatonphil (approved):

PR #957, opened 2023-07-05
[Java client] Removes the dependency on the JUI project.

batiati:
## Motivation:

While [JUI](https://github.com/zig-java/jui) is an excellent library that has served us well, we made this decision to align with the [TIGER_STYLE guidelines](https://github.com/tigerbeetle/tigerbeetle/blob/main/docs/TIGER_STYLE.md#dependencies), which emphasize a foundational approach with no external dependencies.

## The new JNI bindings:

The JNI binding code was rewritten from scratch.

Unlike JUI, which prioritized ergonomics, our focus is on low-level control over the JNI API, with no abstractions for handling Zig datatypes and errors but rather a thin layer that only translates the C API into idiomatic Zig datatypes and follows TB's naming convention.

The common approach for most languages to deal with JNI's vtable is either:

a) Following the C API:

```C
str = env->functions->NewStringUTF(env, char_ptr);
```

b) Following the C++ API:

```C++
str = env->NewStringUTF(char_ptr);
```
----
We want our implementation to be a _thin layer_ over the JNI API, so it makes sense to use the first approach like this:

```Zig
str = env.functions.NewStringUTF(env, slice.ptr);
```

But at the same time, we also want to expose our naming convention and provide all pointer semantics Zig allows within the C ABI (e.g `[*]`, `[*:0]`, `?*`, `?[*]`), instead of using the raw C translation that represents every pointer as `[*c]`. 

In order to do that, we need to carefully (and manually) review and rewrite the code generated from `jni.h`, and since we went into all this trouble, there's no reason to keep the more verbose C API, so our API can work like the C++ one:

```Zig
str = env.new_string_utf(slice.ptr);
```

Under the hood, we call the underlying vtable translated from `jni.h`:
```Zig
// Traditional implementation:
const JniEnv = struct {
    //Reviewed API using idiomatic Zig and TB's naming convention.
    pub fn new_string_utf(env: *JniEnv, bytes: [*:0]const u8) JInt {
        // Raw API is hidden from the user uses [*c].
        return env.functions.NewStringUTF(env, bytes);
    }
};
```
---
It turns out that we need to maintain two declarations for the same function, which is error-prone, especially because the vtable ABI is **implicitly** defined by the ordering in which each function is declared.

One of those declarations can be `comptime` generated, in our case, the one translated from `jni.h` which contains less information. So we just need one declaration + the function index inside the vtable, which is well documented by the JNI spec.

That's the final result:

```Zig
str = env.new_string_utf(slice.ptr);
```

Now we call the vtable directly at the function's index:
```Zig
// Comptime implementation:
const JniEnv = opaque {
   // Holds only the function index.
   const TableIndex = enum(u8) {
       get_env = 4,
       /// ... omitted ...
       new_string_utf = 167,
   }

    //Declaration using idiomatic Zig and TB's naming convention.
    /// https://docs.oracle.com/en/java/javase/17/docs/specs/jni/functions.html#newstringutf.
    pub inline fn new_string_utf(
        env: *JNIEnv,
        bytes: [*:0]const u8,
    ) JString {
        // Comptime generated, accessing the vtable at the specific index.
        return interface_call(
            env,
            .new_string_utf,
            .{bytes},
        );
    }
};
```

## Changes:

- Some refactoring was needed on the Java client to use the new JNI bindings, but most changes were due to a general review for better code organization, including error handling with panics and exceptions.

- Unit tests for the new JNI bindings were written inspired by the same principle used for testing JUI: a JVM hosted by our process to be able to call JNI functions during the Zig test.

In order to run the test, one need to have the Java SDK installed and `JAVA_HOME` environment variable defined:

```bash
export JAVA_HOME=$(dirname $(dirname $(readlink -f $(which java))))
./scripts/build.sh test:jni
```


build.zig
```diff
@@ -8,7 +8,6 @@ const config = @import("./src/config.zig");
 
 // We need to ensure we only add a clone dependency once.
 var git_clone_tracy = false;
-var git_clone_jui = false;
```
matklad: this probably wants a rebase

build.zig
```diff
@@ -978,3 +970,70 @@ fn client_docs(
...
+        // On Windows we need to set the DLL directory.
+        const set_dll_directory = struct {
+            pub extern "kernel32" fn SetDllDirectoryA(path: [*:0]const u8) callconv(.C) std.os.windows.BOOL;
+        }.SetDllDirectoryA;
+
+        var java_bin_path = std.fs.path.joinZ(
+            b.allocator,
+            &.{ java_home, "\\bin" },
+        ) catch unreachable;
+        defer b.allocator.free(java_bin_path);
```
matklad: Let's consistently _not_ free stuff and remove this defer (eg, b.pathJoin above does that).

build.zig
```diff
@@ -978,3 +970,70 @@ fn client_docs(
...
+            &.{ java_home, "\\bin\\server" },
+        ) catch unreachable;
+        defer b.allocator.free(java_bin_server_path);
+        _ = set_dll_directory(java_bin_server_path);
+
+        // TODO(zig): The function `JNI_CreateJavaVM` tries to detect
+        // the stack size and causes a SEGV that is handled by Zig's panic handler.
+        // https://bugzilla.redhat.com/show_bug.cgi?id=1572811#c7
+        //
+        // The workarround is run the tests in "ReleaseFast" mode.
```
matklad:
```suggestion
        // The workaround is run the tests in "ReleaseFast" mode.
```

build.zig
```diff
@@ -978,3 +970,70 @@ fn client_docs(
...
+    mode: Mode,
+) void {
+    var test_step = b.step("test:jni", "Run the JNI tests");
+    const jni_abi = if (builtin.os.tag == .linux) b.option(
+        enum { gnu, musl },
+        "jni-test-abi",
+        "Which ABI the JVM will run the test uses",
+    ) orelse .gnu else {};
+
+    const java_home = b.env_map.get("JAVA_HOME") orelse {
```
matklad: I wonder if we should make this part of `java_client`, rather then `test`, and then just fail if JAVA_HOME is not set? To make sure this is properly executed by CI.

matklad (commented):
Didn't get to full review today!

build.zig
batiati: I think we might not want this to be part of `./scripts/build.sh java_client`, just to keep the dependency of JDK not a requirement for building the client library.

Although it's a good idea to include `./scripts/build.sh test:jni` as a step inside Maven, so `mvn package` will run all the tests, both from CI and from the developer machine.

src/clients/java/src/client.zig
```diff
@@ -1,62 +1,361 @@
...
+            tb.submit(context.client, packet);
+        } else {
+            assert(acquire_status != .ok);
+        }
 
-    pub fn load(env: *jui.JNIEnv) !void {
+        return acquire_status;
+    }
 
+    /// Completion callback, aways called from the native thread.
```
matklad:
```suggestion
    /// Completion callback, always called from the native thread.
```

src/clients/java/src/client.zig
```diff
@@ -1,62 +1,361 @@
...
+    var initialization_exception_class: jni.JClass = null;
+    var initialization_exception_ctor_id: jni.JMethodID = null;
+    var assertion_error_class: jni.JClass = null;
+
+    var request_class: jni.JClass = null;
+    var request_send_buffer_field_id: jni.JFieldID = null;
+    var request_send_buffer_len_field_id: jni.JFieldID = null;
+    var request_reply_buffer_field_id: jni.JFieldID = null;
+    var request_operation_method_id: jni.JMethodID = null;
+    var request_end_request_method_id: jni.JMethodID = null;
```
matklad, line 278: I guess potentially we can do something like

```
const ReflectionHelper = struct {
    var instance: ?@This() = null;

    initialization_exception_class: jni.JClass,
    initialization_exception_ctor_id: jni.JMethodID,
    assertion_error_class: jni.JClass,

    request_class: jni.JClass,
    request_send_buffer_field_id: jni.JFieldID,
    request_send_buffer_len_field_id: jni.JFieldID,
    request_reply_buffer_field_id: jni.JFieldID,
    request_operation_method_id: jni.JMethodID,
    request_end_request_method_id: jni.JMethodID,
```

to get some compile-time checking that we try to initialize everything. But as jni.JClass and such are themselves nullable, I don't think it is worth it (that is, we need to check _both_ that we tried to initialized a field, and that `find_class` returned a non-null value).

src/clients/java/src/client.zig
```diff
@@ -93,415 +378,305 @@ const ReflectionHelper = struct {
...
-    pub fn get_send_buffer(env: *jui.JNIEnv, this_obj: jui.jobject) ?[]u8 {
+    pub fn get_send_buffer_slice(env: *jni.JNIEnv, this_obj: jni.JObject) ?[]u8 {
         assert(this_obj != null);
         assert(request_send_buffer_field_id != null);
         assert(request_send_buffer_len_field_id != null);
 
-        var buffer_obj = env.getField(.object, this_obj, request_send_buffer_field_id) orelse
+        var buffer_obj = env.get_object_field(this_obj, request_send_buffer_field_id) orelse
+            return null;
+        defer env.delete_local_ref(buffer_obj);
```
matklad, line 416: Hm, this smells suspiciously to me, although I don't know exactly how this works, so might be wrong. 

It seems that `get_object_field` registers a refenrence with JVM, which pins the object in place and prevents GC from freeing it. That's why we need `delete_local_ref` to signal that we are done.

But we are returning a slice pointing inside the buffer! Sholdn't we delete the local ref after we are done with it? Or would it be corret to delete_local_ref here immediately, without defer, because we are holding onto the request object anyway?

src/clients/java/src/client.zig
```diff
@@ -93,415 +378,305 @@ const ReflectionHelper = struct {
...
-        );
-        return @bitCast(jui.jlong, @ptrToInt(context));
-    }
+    pub inline fn get_direct_buffer(
+        env: *jni.JNIEnv,
+        buffer_obj: jni.JObject,
+    ) ?[]u8 {
+        var buffer_capacity = env.get_direct_buffer_capacity(buffer_obj);
+        if (buffer_capacity < 0)
+            return null;
```
matklad:
```suggestion
        if (buffer_capacity < 0) return null;
```

let's maybe format on a single line? Multiline ifs without braces are prone to "goto fail"-style bugs (we use zigfmt, so they are unlikely, but still)

src/clients/java/src/client.zig
```diff
@@ -93,415 +378,305 @@ const ReflectionHelper = struct {
...
-        env: *jui.JNIEnv,
-        class: jui.jclass,
-        cluster_id: jui.jint,
-        addresses: jui.jstring,
-        max_concurrency: jui.jint,
-    ) callconv(.C) jui.jlong {
-        _ = class;
-        var context = NativeClient.client_init(
-            false,
+    pub inline fn find_method(
```
matklad, line 637: Can these find_  methods also assert that the result is not-null?

src/clients/java/src/client.zig
```diff
@@ -65,23 +364,9 @@ const ReflectionHelper = struct {
...
-            @panic("JNI: Unexpected error creating a global reference for " ++ class_name);
-        };
-    }
-
-    pub fn unload(env: *jui.JNIEnv) void {
-        env.deleteReference(.global, initialization_exception_class);
-        env.deleteReference(.global, request_class);
+    pub fn unload(env: *jni.JNIEnv) void {
+        env.delete_global_ref(initialization_exception_class);
+        env.delete_global_ref(request_class);
```
matklad, line 359: `assertion_error_class` needs a similar treatment here?

src/clients/java/src/client.zig
batiati: Since the type definition is `?*anyopaque`, I think it's just [feng shui programming](https://github.com/SerenityOS/serenity/blob/60ac254df62408bf1831d28c9a712ee3d27a1366/CONTRIBUTING.md?plain=1#L70) 🤣

src/clients/java/src/client.zig
batiati: > That's why we need delete_local_ref to signal that we are done

Actually, no. Local references live on the JVM stack and are freed (decrease the Rc) automatically when a function ends. Normally only global references need to be explicitly deleted. In our case, since we run code from an unmanaged thread, we need to free both global and local references, otherwise, the JVM would not know when to pop the reference from the call stack.

> But we are returning a slice pointing inside the buffer! 

It's correct to free the buffer object reference here. The buffer's content is a stable address from unmanaged memory allocated by [UNSAFE.alloc(...)](https://github.com/openjdk/jdk21/blob/830279e0540c01e2012c60b724857a7fe4a450b1/src/java.base/share/classes/java/nio/Direct-X-Buffer.java.template#L131), and the buffer object is only a Rc that is still being held by the Request object.

Note this is valid only for DirectBuffers, for primitive arrays and strings we need to keep the reference and pin/unpin the content.

src/clients/java/src/client.zig
batiati: It's done by the `orelse` clause:

https://github.com/tigerbeetle/tigerbeetle/blob/120e6a62ea6320d1f23b0c3b89e701693140ca57/src/clients/java/src/client.zig#L654-L659

src/clients/java/src/jni.zig
```diff
@@ -0,0 +1,3230 @@
...
+        );
+    }
+
+    /// Raises a fatal error and does not expect the VM to recover.
+    /// This function does not return.
+    /// https://docs.oracle.com/en/java/javase/17/docs/specs/jni/functions.html#fatalerror.
+    pub inline fn fatal_error(
+        env: *JNIEnv,
+        msg: [*]const u8,
+    ) void {
```
matklad: I guess this could be 

```suggestion
    ) noreturn {
```

src/clients/java/src/jni.zig
```diff
@@ -0,0 +1,3230 @@
+///! Based on Java Native Interface Specification and Java Invocation API.
+///! https://docs.oracle.com/en/java/javase/17/docs/specs/jni.
+const std = @import("std");
```
matklad, line 16: Could you add a motivation here? Basically, that because interface is stable and pinned down by java langauge specification, that jni.h has an unfortunate license, and so we just "manually" code this and rely on exhaustive tests to catch out any bugs?

build.zig
matklad: So, what's the CI job that tests this?

src/clients/java/src/jni.zig
batiati: Those are valid points, but I think it's not only about that.

If you inspect the very own `jni.h` [C++ declarations](https://github.com/openjdk/jdk/blob/ec7da91bd83803b7d91a4de3a01caf0ba256c037/src/java.base/share/native/include/jni.h#L780-L811), you'll see it's equally error-prone.

I mean, nothing prevents a human error from calling the wrong vtable function or passing the wrong argument.
In that respect, even if we distribute the .h file or a translated version of it, we're still in the same situation when our API needs to call the vtable pointer (like the C++ API does).

That's why I implemented the vtable using only indexes to minimize the amount of code we need to maintain.

```C++
/*
 * We use inlined functions for C++ so that programmers can write:
 *
 *    env->FindClass("java/lang/String")
 *
 * in C++ rather than:
 *
 *    (*env)->FindClass(env, "java/lang/String")
 *
 * in C.
 */

struct JNIEnv_ {
    const struct JNINativeInterface_ *functions;
#ifdef __cplusplus

    jint GetVersion() {
        return functions->GetVersion(this);
    }
    jclass DefineClass(const char *name, jobject loader, const jbyte *buf,
                       jsize len) {
        return functions->DefineClass(this, name, loader, buf, len);
    }
    jclass FindClass(const char *name) {
        return functions->FindClass(this, name);
    }
    jmethodID FromReflectedMethod(jobject method) {
        return functions->FromReflectedMethod(this,method);
    }
    jfieldID FromReflectedField(jobject field) {
        return functions->FromReflectedField(this,field);
    }

    // ...
}
```

matklad: I mean that we need to include our motivation as a source code comment `//!`, so that the next person doesn’t need to wonder why we do it this way. I agree that what we do is the right approach.

src/clients/java/src/client.zig
matklad: :face-palm: :sweat_smile:

src/clients/java/src/jni.zig
```diff
@@ -0,0 +1,3230 @@
...
+            self: *T,
+            comptime function: T.FunctionTable,
+            args: anytype,
+        ) return_type: {
+            const type_info = @typeInfo(JniFn(function));
+            break :return_type type_info.Fn.return_type.?;
+        } {
+            const Fn = JniFn(function);
+            const VTable = extern struct {
+                functions: [*]*anyopaque,
```
batiati: It should be const ... 

```suggestion
                functions: [*]const *const anyopaque,
```

build.zig
batiati: It just landed.
Since Maven knows `JAVA_HOME`, it's called from `mvn` build. 

I've made changes on `build.zig` to detect if we need to run the test using musl, it's not a build argument anymore.

```
❯ mvn package
[INFO] Scanning for projects...
[INFO] Inspecting build with total of 1 modules...
[INFO] Installing Nexus Staging features:
[INFO]   ... total of 1 executions of maven-deploy-plugin replaced with nexus-staging-maven-plugin
[INFO] 
[INFO] ------------------< com.tigerbeetle:tigerbeetle-java >------------------
[INFO] Building TigerBeetle Java client 0.0.1-SNAPSHOT
[INFO] --------------------------------[ jar ]---------------------------------
[INFO] 
[INFO] --- exec-maven-plugin:3.1.0:exec (zig-jni-test) @ tigerbeetle-java ---
1/33 test "JNI: check jvm"... OK
2/33 test "JNI: GetVersion"... OK
3/33 test "JNI: FindClass"... OK
4/33 test "JNI: GetSuperclass"... OK
5/33 test "JNI: IsAssignableFrom"... OK
6/33 test "JNI: GetModule"... OK
7/33 test "JNI: Throw"... OK
8/33 test "JNI: ThrowNew"... OK
9/33 test "JNI: ExceptionOccurred"... OK
10/33 test "JNI: ExceptionDescribe"... Exception in thread "main" java.lang.Exception: EXCEPTION DESCRIBED CORRECTLY
OK
11/33 test "JNI: ExceptionClear"... OK
12/33 test "JNI: ExceptionCheck"... OK
13/33 test "JNI: References"... OK
14/33 test "JNI: LocalFrame"... OK
15/33 test "JNI: AllocObject"... OK
16/33 test "JNI: NewObject"... OK
17/33 test "JNI: IsInstanceOf"... OK
18/33 test "JNI: GetFieldId"... OK
19/33 test "JNI: Get<Type>Field, Set<Type>Field"... OK
20/33 test "JNI: GetMethodId"... OK
21/33 test "JNI: Call<Type>Method,CallNonVirtual<Type>Method"... OK
22/33 test "JNI: GetStaticFieldId"... OK
23/33 test "JNI: GetStatic<Type>Field, SetStatic<Type>Field"... OK
24/33 test "JNI: GetStaticMethodId"... OK
25/33 test "JNI: CallStatic<Type>Method"... OK
26/33 test "JNI: strings"... OK
27/33 test "JNI: strings utf"... OK
28/33 test "JNI: GetStringRegion"... OK
29/33 test "JNI: GetStringUTFRegion"... OK
30/33 test "JNI: GetStringCritical"... OK
31/33 test "JNI: DirectByteBuffer"... OK
32/33 test "JNI: object array"... OK
33/33 test "JNI: primitive arrays"... OK
All 33 tests passed.
[INFO] 
[INFO] --- exec-maven-plugin:3.1.0:exec (zig-build) @ tigerbeetle-java ---
```

build.zig
```diff
@@ -954,6 +952,99 @@ fn client_docs(
...
+                // the libjvm.so is linked against libc or musl.
+                // It's reasonable to assume that ldd will be present.
+                const abi = builder.exec(&.{
+                    "sh",
+                    "-c",
+                    builder.fmt(
+                        "ldd {s} | grep -q musl && echo musl || echo libc",
+                        .{builder.pathJoin(&.{ libjvm_path, "libjvm.so" })},
+                    ),
+                }) catch unreachable;
```
matklad: Let's reduce our dependencies to just `ldd` here, `builder.exec(.{"ldd", builder.pathJoin(&.{ libjvm_path, "libjvm.so" }})`, and then use `std.mem.IndexOf` to find musl/gnu.

build.zig
```diff
@@ -954,6 +952,99 @@ fn client_docs(
...
+                self.tests.target.abi = if (std.mem.startsWith(u8, abi, "musl")) .musl else .gnu;
```
matklad: :thinking: would it be correct to set step's `abi` at execution time, rather than at configuration time? I think it would work fine, given the current impl, but I am not sure it'll be guaranteed to work. 

I guess it doesn't really matter, and the current solution seems simple enough.

build.zig
```diff
@@ -954,6 +952,99 @@ fn client_docs(
...
+            .macos => try builder.env_map.put("DYLD_LIBRARY_PATH", libjvm_path),
+            .linux => {
+                // On Linux, detects the abi by calling `ldd` to check if
+                // the libjvm.so is linked against libc or musl.
+                // It's reasonable to assume that ldd will be present.
+                const abi = builder.exec(&.{
+                    "sh",
+                    "-c",
+                    builder.fmt(
+                        "ldd {s} | grep -q musl && echo musl || echo libc",
```
matklad: Maybe change the logic here from `if 'musl' .musl else .glic` to `if 'musl' .musl elseif 'libc' .gnu else "error"`? Just as a defense in depth, to safely bail out if we can't actually parse the abi.

build.zig
batiati: Not only `tests.target.abi` but `tests.addLibPath` is called during the execution phase of the parent step, before the test step runs, so I guess it is still in its configuration phase.

Unless `zig build` changes this behavior in the future to render all commands at once before calling `step.make(...)` on the parent step.

matklad (approved):

PR #959, opened 2023-07-05
RFC: add snapshot testing infra

matklad:


src/shell.zig
```diff
@@ -313,3 +313,50 @@ test "shell: expand_argv" {
...
+            try want.diff(buf.items);
+        }
+    };
+
+    try T.check("zig version", .{}, snap(@src(),
+        \\["zig","version"]
+    ));
+    try T.check("  zig  version  ", .{}, snap(@src(),
+        \\["zig","version"]
+    ));
```
matklad, line 304: Here's how robustness to refactors plays out. Suppose that I want to extend the syntax of `expand_argv` to support not only the command line itself, but also environment variables: 

```zig
expand_argv("LD_PRELOAD={faulty_libc} tigerbeetle format", .{ faulty_libc = "..." })
```

This requires me to change the return type from `[]const []const u8` to something like `struct { env: EnvMap, args: []const []const u8}`. The `snap` thing would allow me to re-run the existing test suit with `SNAP_UPDATE=1` to fix all existing tests, without actually touching the code.

src/testing/snaptest.zig
```diff
@@ -0,0 +1,220 @@
...
+//!   - encourages creation of reusable visualizations for data structures.
+//!
+//! Implementation-wise, `snaptest` provides a `Snap` type, which can be thought of as a Zig string
+//! literal which also remembers its location in the source file, can be diffed with other strings,
+//! and, crucially, can _update its own source code_ to match the expected value.
+//!
+//! Example usage:
+//!
+//! ```
+//! const Snap = @import("snaptest.zig");
```
sentientwaffle:
```suggestion
//! const Snap = @import("snaptest.zig").Snap;
```

src/testing/snaptest.zig
```diff
@@ -0,0 +1,220 @@
...
+//! const snap = Snap.snap;
+//!
+//! fn check_addition(x: u32, y: u32, want: Snap) !void {
+//!     const got = x + y;
+//!     try want.diff_fmt("{}", .{got});
+//! }
+//!
+//! test "addition" {
+//!     check_addition(2, 2, snap(@src(),
```
sentientwaffle:
```suggestion
//!     try check_addition(2, 2, snap(@src(),
```

src/testing/snaptest.zig
```diff
@@ -0,0 +1,220 @@
...
+    } else return null;
+
+    return Range{ .start = snap_start, .end = snap_end };
+}
+
+fn is_multiline_string(line: []const u8) bool {
+    for (line) |c, i| {
+        switch (c) {
+            ' ' => {},
+            '\\' => return (i < line.len and line[i + 1] == '\\'),
```
sentientwaffle: Should the first clause be `i + 1 < line.len`?

src/testing/snaptest.zig
```diff
@@ -0,0 +1,220 @@
...
+    const snap_start = while (lines.next()) |line| : (line_number += 1) {
+        if (line_number == snap_line) {
+            assert(std.mem.indexOf(u8, line, "@src()") != null);
+        }
+        if (line_number == snap_line + 1) {
+            assert(is_multiline_string(line));
+            break offset;
+        }
+        offset += line.len + 1; // 1 for \n
+    } else return null;
```
sentientwaffle: Should this be `else unreachable`? How could this be hit?

src/testing/snaptest.zig
```diff
@@ -0,0 +1,220 @@
...
+
+        try std.fs.cwd().writeFile(snapshot.location.file, file_text_updated.items);
+
+        std.debug.print("Updated {s}\n", .{snapshot.location.file});
+        return error.SnapUpdated;
+    }
+};
+
+const Range = struct { start: usize, end: usize };
+fn snap_range(text: []const u8, snap_line: u32) ?Range {
```
sentientwaffle: I'm assuming that string replacement ends up simpler than parsing the AST, mutating, and then rendering it?

src/testing/snaptest.zig
matklad: Yeah, I am inconsistent here. This _could_ be fixed if the file is modified in-place before the test is run. In general, there's fundamental fragility to update workflow. But it should be fine in practice, and it probably is better to assert consistently here.

src/testing/snaptest.zig
matklad: Exactly: I originally wanted to do `std.zig.Ast` thing here because, wow, I can, but then figured out that Zig's syntax is simple enough to not-overthink it. Added motivational comment.

sentientwaffle (approved):

src/testing/snaptest.zig
matklad: Indeed! In Rust, I'd spell this as `line[i..].starts_with(r"\\")` to be more robust, but in Zig `std.mem.startsWith(u8, ` feels like a mouthful and doesn't feel minimal enough.

PR #960, opened 2023-07-06
[Client] Fix cross-compilation issues

### LINKED ISSUE #762: dotnet build doesn't reliably build on m1 mac
See the log here: https://gist.github.com/eatonphil/df78a29152a9ce3f2f4af35401dcea0e. Even after clearing the global and local cache it doesn't build very reliably. Sometimes after running `dotnet build` twice it works. Sometimes after running `dotnet build` six times.

batiati:
This PR solves a common cross-compiling problem for multiple targets in the same `zig build` command.
See https://github.com/ziglang/zig/issues/9711

Also, for the dotnet client, it fixed the `csproj` exec tag, which invoked the zig compiler twice (once for each target framework), causing zig to fail with `cannot write file XYZ`, because two zig instances were running at the same time.  
Closes #762


cb22 (approved):

eatonphil:
Awesome!

PR #961, opened 2023-07-06
Reduce cache-grid for integration tests

batiati:
This PR sets the flag `--cache-grid=128MB` for all integration tests.
We start many ephemeral TB instances in this use case, so we want low memory usage rather than a large cache for performance, especially when running by CI.

Also, it changes one of the Java tests to avoid [flaky results](https://github.com/tigerbeetle/tigerbeetle/actions/runs/5468389491/jobs/9956628305?pr=957) on slow machines.

### DIFF
--- a/src/clients/dotnet/TigerBeetle.Tests/IntegrationTests.cs
+++ b/src/clients/dotnet/TigerBeetle.Tests/IntegrationTests.cs
@@ -986,7 +986,7 @@ namespace TigerBeetle.Tests
         private const string TB_FILE = "dotnet-tests.tigerbeetle";
         private const string TB_SERVER = TB_PATH + "/" + TB_EXE;
         private const string FORMAT = $"format --cluster=0 --replica=0 --replica-count=1 ./" + TB_FILE;
-        private const string START = $"start --addresses=" + TB_PORT + " ./" + TB_FILE;
+        private const string START = $"start --addresses=" + TB_PORT + " --cache-grid=128MB ./" + TB_FILE;
 
         private readonly Process process;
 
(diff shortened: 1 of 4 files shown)


cb22 (approved):
@batiati I'm happy with this as-is, but we could also add support to TigerBeetle directly for something like `--cache-grid=min`?

batiati:
> @batiati I'm happy with this as-is, but we could also add support to TigerBeetle directly for something like `--cache-grid=min`?

Do you mean, making `min` a keyword for the `--cache-grid` flag? It's an elegant solution.
There is a formula for the min size, or we should just declare const min=128MB arbitrarily?

cb22:
> Do you mean, making min a keyword for the --cache-grid flag? It's an elegant solution.
> There is a formula for the min size, or we should just declare const min=128MB arbitrarily?

Yes - exactly that. 128MB is arbitrary, but it is the current min for the grid cache. The value comes from a combination of `value_count_max_multiple` (exported by `set_associative_cache.zig`) - currently 2048 for the config the grid uses, plus the grid's block size of 64K = 128MB.

batiati:
Too complicated to be a trivial change, I'm going to merge this as-is and we can tackle that in another PR.

- Introducing `min`, should we validate that the value the user informed is >= min?

- We have multiple `size_default` for each config type (e.g. `production`, `development`, `test_min`).
Should we have a different `size_min` for each one as well?

- Other settings that are CLI overridable, should they support `min` too?
For example `--cache-accounts`, `--cache-transfers`, `--cache-transfers-posted`.

PR #965, opened 2023-07-07
Don't call .install() in build.zig for aof and benchmark

cb22:
By calling .install(), these binaries get added to our `zig build install` target, which is used in lots of places, but nowhere does it need these binaries (they're always called via `zig build benchmark` or `zig build aof` when needed).

This speeds up build times by ~10-20%.

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -100,7 +100,6 @@ pub fn build(b: *std.build.Builder) void {
         benchmark.setTarget(target);
         benchmark.setBuildMode(mode);
         benchmark.addPackage(vsr_package);
-        benchmark.install();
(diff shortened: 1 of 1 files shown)


batiati (approved):
Nice!

Just a question ... since now `zig build aof` is only executing the program, do we need to output it in `./zig-out/bin`?

cb22:
> Just a question ... since now `zig build aof` is only executing the program, do we need to output it in `./zig-out/bin`?

I think it's OK as is - the `aof` stuff is a production ready stop-gap; so ideally we'll be deprecating it entirely sooner rather than later! It'll only be run on servers we control completely.

PR #968, opened 2023-07-07
LSM: Table{Index,Filter,Data}Schema

sentientwaffle:
(This PR has lots of lines changed, but for the most part they are just moved.)

### Motivation

The scrubber and `GridRepairQueue` will need to decode table index blocks.
(Specifically, they need to enumerate all referenced filter/data blocks in an index block).

But the scrubber and the `GridRepairQueue` are not specialized per-tree.

## Overview

Rather than switching between decoders based on the tree id (which would be complicated -- each `Table`/`Tree` is a different specialized type), encode the schema's parameters directly into the table index block.

Then the scrubber/repair-queue constructs the relevant decoder schema at runtime.
This change also lends itself well to backwards-compatible table layout changes.

(Aside: Functions that use `Key` are still specialized. The scrubber/repair-queue don't need these anyway.)

### DIFF
--- a/src/lsm/compaction.zig
+++ b/src/lsm/compaction.zig
@@ -46,6 +46,7 @@ const GridType = @import("grid.zig").GridType;
 const allocate_block = @import("grid.zig").allocate_block;
 const TableInfoType = @import("manifest.zig").TableInfoType;
 const ManifestType = @import("manifest.zig").ManifestType;
+const TableIndexSchema = @import("table.zig").TableIndexSchema;
 const TableDataIteratorType = @import("table_data_iterator.zig").TableDataIteratorType;
 const LevelTableValueBlockIteratorType = @import("level_data_iterator.zig").LevelTableValueBlockIteratorType;
 
@@ -410,10 +411,12 @@ pub fn CompactionType(
             // `index_block` is only valid for this callback, so copy its contents.
             // TODO(jamii) This copy can be avoided if we bypass the cache.
             stdx.copy_disjoint(.exact, u8, compaction.index_block_a, index_block);
+
+            const index_a = TableIndexSchema.from_index_block(compaction.index_block_a);
             compaction.iterator_a.start(.{
                 .grid = compaction.context.grid,
-                .addresses = Table.index_data_addresses_used(compaction.index_block_a),
-                .checksums = Table.index_data_checksums_used(compaction.index_block_a),
+                .addresses = index_a.data_addresses_used(compaction.index_block_a),
+                .checksums = index_a.data_checksums_used(compaction.index_block_a),
             });
             compaction.release_table_blocks(compaction.index_block_a);
             compaction.state = .compacting;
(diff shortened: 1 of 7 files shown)


src/lsm/level_data_iterator.zig
```diff
@@ -139,21 +140,19 @@ pub fn LevelTableValueBlockIteratorType(comptime Table: type, comptime Storage:
             read: *Grid.Read,
             index_block: BlockPtrConst,
         ) void {
-            const it = @fieldParentPtr(
-                LevelTableValueBlockIterator,
-                "read",
-                read,
-            );
+            const it = @fieldParentPtr(LevelTableValueBlockIterator, "read", read);
```
sentientwaffle, line 143: (Just reformatting).

src/vsr.zig
```diff
@@ -795,9 +795,9 @@ pub const Header = extern struct {
         assert(self.command == .block);
         if (self.parent != 0) return "parent != 0";
         if (self.client != 0) return "client != 0";
-        if (self.context != 0) return "context != 0";
         if (self.view != 0) return "view != 0";
         if (self.op == 0) return "op == 0"; // address ≠ 0
+        if (self.commit != 0) return "commit != 0";
```
sentientwaffle, line 800: The data this field used to hold is now in `context` (along with some others).

src/lsm/table.zig
```diff
@@ -239,52 +243,22 @@ pub fn TableType(
...
 
-            const padding_offset = values_offset + values_size;
-            const padding_size = block_size - padding_offset;
-        };
+        pub const data = TableDataSchema.init(.{
+            .key_count = layout.block_key_count,
+            .key_layout_size = layout.block_key_layout_size,
+            .value_count_max = layout.block_value_count_max,
+            .value_size = value_size,
+        });
```
sentientwaffle: These are the schemas which encode _new_ blocks, so they are still comptime-known.

src/lsm/table.zig
```diff
@@ -361,7 +335,7 @@ pub fn TableType(
                         filter.padding_size,
 
                         data.key_count,
-                        data.block_value_count_max,
+                        data.value_count_max,
```
sentientwaffle, line 304: Renamed this to be more consistent.

src/lsm/compaction.zig
```diff
@@ -410,10 +411,12 @@ pub fn CompactionType(
             // `index_block` is only valid for this callback, so copy its contents.
             // TODO(jamii) This copy can be avoided if we bypass the cache.
             stdx.copy_disjoint(.exact, u8, compaction.index_block_a, index_block);
+
+            const index_a = TableIndexSchema.from_index_block(compaction.index_block_a);
```
jorangreef: This is great. 🚀 

```suggestion
            const schema_index_block_a = SchemaIndexBlock.from(compaction.index_block_a);
```

* Clarify that `schema_index_block_a` is a schema and not an index block.
* Shorten the `from` method name since the type name is clear.
* Use MSB order to put Schema first.

batiati: It's great to clarify that it's a schema, not an index block. However, I think `Schema` may not be the MSB, e.g `filter_block` vs `filter_bytes` vs `filter_schema` and `data_block` vs `data_schema`.

batiati (commented):
We just need to settle the naming choice 🤔.
Apart from that, LGMT.

batiati (dismissed):

batiati (approved):
:+1:

PR #970, opened 2023-07-07
VSR: Sync follow-up refactors

sentientwaffle:
Follow-up changes from comments on https://github.com/tigerbeetle/tigerbeetle/pull/834 that didn't make it in.
(Mostly renames.)

I recommend reviewing commit-by-commit.

### DIFF
--- a/src/testing/cluster.zig
+++ b/src/testing/cluster.zig
@@ -486,7 +486,7 @@ pub fn ClusterType(comptime StateMachineType: fn (comptime Storage: type, compti
                 .message_sent => |message| {
                     cluster.state_checker.on_message(message);
                 },
-                .commit => {
+                .committed => {
                     cluster.log_replica(.commit, replica.replica);
                     cluster.state_checker.check_state(replica.replica) catch |err| {
                         fatal(.correctness, "state checker error: {}", .{err});
--- a/src/testing/cluster.zig
+++ b/src/testing/cluster.zig
@@ -528,7 +528,7 @@ pub fn ClusterType(comptime StateMachineType: fn (comptime Storage: type, compti
                         cluster.log_replica(.sync_commenced, replica.replica);
                         cluster.sync_checker.replica_sync_start(replica);
                     },
-                    .not_syncing => {
+                    .idle => {
                         cluster.log_replica(.sync_completed, replica.replica);
                         cluster.sync_checker.replica_sync_done(replica);
                         if (cluster.replica_diverged.isSet(replica.replica)) {
@@ -584,7 +584,7 @@ pub fn ClusterType(comptime StateMachineType: fn (comptime Storage: type, compti
 
             const role: u8 = role: {
                 if (cluster.replica_health[replica_index] == .down) break :role '#';
-                if (replica.sync_stage != .not_syncing) break :role '~';
+                if (replica.sync_stage != .idle) break :role '~';
                 if (replica.standby()) break :role '|';
                 if (replica.primary_index(replica.view) == replica.replica) break :role '/';
                 break :role '\\';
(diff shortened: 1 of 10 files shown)


src/vopr_hub/SETUP.md
```diff
@@ -120,7 +120,7 @@ export CURRENT_VOPR=$1
 go run ./src/vopr_hub/scheduler/main.go
 
 # Run the VOPR a few times before we go back to the top:
-zig/zig run ./src/vopr.zig -- --send="<address>" --simulations=5
+zig/zig run ./src/vopr.zig -- --send="<address>" --simulations=100
```
sentientwaffle, line 123: This helps avoid running into rate limits from github. It is already running on our voprhub instance.

sentientwaffle: (It's unrelated to state sync... just included in this PR for convenience.)

src/testing/cluster.zig
```diff
@@ -486,7 +486,7 @@ pub fn ClusterType(comptime StateMachineType: fn (comptime Storage: type, compti
                 .message_sent => |message| {
                     cluster.state_checker.on_message(message);
                 },
-                .commit => {
+                .committed => {
```
matklad, line 489: Tangentially related, but one cute naming convention for events is `onDidDoSomething/onWillDoSomething` from VS Code (https://code.visualstudio.com/api/references/vscode-api), eg `onDidSaveTextDocument/onWillSaveTextDocument`. They name all events that way, to remove ambiguity about when, exactly, the hoot is called relative to the event.

src/vsr/client_replies.zig
```diff
@@ -1,3 +1,15 @@
...
+//! This allows them to be resent to the corresponding client if the client missed the original
+//! reply message (e.g. dropped packet).
+//!
+//! - Client replies' headers are redundantly stored in the SuperBlock's ClientSessions trailer.
+//! - Client replies are only stored when `header.size ≠ sizeOf(Header)`
+//!   (the header is in ClientSessions).
+//! - Corrupt client replies can be repaired from other replicas.
+//!
+//! ClientReplies may defer writing the reply immediately – often the reply will be replaced
+//! quickly. But all writes must flush during a checkpoint.
```
matklad, line 12: I think we can maybe tell a little bit more about how things fit together, and how this relates to `ClientSessions`, something along the lines of:

Superblock's `ClientSessions` stores the _header_ of the latest reply. The body of the reply (if it is non-empty) can be fond in the `.replies` zone of the datafile through `ClientReplies` type. Replies are written to disc asynchronously and subsequent writes for the same client are coalesced (as we care only about the latest reply). We guarantee that replies are durable(*)

---

I am actually curious what's our exact durability guarantee for replies? I _think_ a cluster can lose a reply (eg, checkpointed reply gets corrupted on all replicas), but, unlike lost prepare, this won't render the cluster unavailable, right?

src/vsr/replica.zig
```diff
@@ -2396,7 +2396,7 @@ pub fn ReplicaType(
         fn on_sync_trailer(
             self: *Self,
             comptime trailer: vsr.SuperBlockTrailer,
-            message: *Message,
+            message: *const Message,
```
matklad, line 2399: `*const Message` means "can't manipulate reference count", but it actually still is possible to mutate the body of the message, right? In situations where we do mutate the body, but don't touch rc, do we use `*const Message` or `*Message`? I'd say `const` --- language says that `const`=shallowly immutable, and assigning any extra sematics would be error prone.

matklad (approved):

src/vsr/client_replies.zig
sentientwaffle: > I am actually curious what's our exact durability guarantee for replies? I think a cluster can lose a reply (eg, checkpointed reply gets corrupted on all replicas), but, unlike lost prepare, this won't render the cluster unavailable, right?

Exactly right. I'll add this in a follow-up:

```zig
//! If the same reply is corrupted by all replicas, the cluster is still available.
//! If the respective client also never received the reply (due to a network fault), the client may
//! be "locked out" of the cluster – continually retrying a request which has been executed, but
//! whose reply has been permanently lost. This can be resolved by the operator restarting the
//! client to create a new session.
```

PR #971, opened 2023-07-09
Pass Context.table by reference.

batiati:
`LevelTableValueBlockIterator.Context.table` has been updated as a slice instead of a bounded array. 

Rationale:

- To avoid taking a copy of the array.
- To make the caller responsible for sizing the array (future use case for Scans).


sentientwaffle (approved):
:+1:

PR #972, opened 2023-07-10
LSM: Manifest Fuzzer

kprotty:
Implements TreeType using only Keys instead of Values. This fuzzes the expected usecase of Manifest without Grid/Storage plumbing. It also gives TreeType and CompactionType a simpler model to compare against.


matklad:
Closing for now!

PR #973, opened 2023-07-10
Switch release process for binaries to use gh cli rather than curl

cb22:
This was failing silently (!) because of our repo name change, which was causing the `curl` to fail. To be followed up by a separate PR that validates our release completed successfully.

### DIFF
--- a/.github/workflows/release_artifacts.yml
+++ b/.github/workflows/release_artifacts.yml
@@ -78,14 +78,10 @@ jobs:
       if: ${{ inputs.dry-run != 'true' }}
       run: |
         export GIT_TAG="${{ needs.version.outputs.version }}" && \
-        export RELEASE_ID=`curl -H 'Authorization: Bearer ${{ secrets.GITHUB_TOKEN }}' "https://api.github.com/repos/tigerbeetledb/tigerbeetle/releases/tags/${GIT_TAG}" | jq '.id'` && \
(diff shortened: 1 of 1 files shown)


batiati (approved):

matklad (approved):
Makes sense, also from the angle that `gh` is the tool by the vendor of our CI platform, so this actually reduces the amount of moving parts. 

Looking at a wider context here, I see that we do full cross-compilation here (we run everything on ubuntu-latest). Which makes me wonder --- do we need to put release logic in .yml files here? 

I am imagining that we can move the matrix into `build.zig`, such that

```
zig build dist
```

produces a "distribution" of a particular version of tigerbeetle, which includes binaries for all platforms we use. 

What we loose here is distributed builds --- a single machine builds everything instead of distributing the work across different workers. What we gain is simplicity --- we are no longer bound to GHA to produce a set of releasable artifacts, anyone can run&debug `zig build dist` locally. Similarly, releases are more atomic ---  we check that we can build all platforms before publishing a single platform. 

Given that TigerBeetle is not only fast, but also small, sticking to non-distributed builds long-term feels right. And I don't think this will even be necessary slower --- once we upgrade to new Zig which supports running build steps in parallel, I think doing multicore build on a single machine would be faster for the distributed COST reasons.

cb22:
@matklad agreed. The build time for each platform isn't even particularly long, and our release is decoupled from merge blocking - which we care more about the duration of.

PR #975, opened 2023-07-10
Release validation: Check all binary artifacts after publishing

cb22:
This introduces a step, `release_validate` that runs after release (with `if: '!cancelled()` to run even if something has failed) which checks that all the artifacts we expect to have published are indeed there.

If not, it sends a message on Slack to notify us.


cb22:
This is ready for review, but going to hold off on merging for a little bit until we're sure our release pipeline is working again as it should.

I've tested the validation checking here manually, by shuffling it to a different point in the pipeline, so hopefully we don't have to have too much testing-on-main with this one!

.github/workflows/release_validate.yml
```diff
@@ -0,0 +1,67 @@
...
+        run: |
+          mkdir -p /tmp/node && cd /tmp/node && \
+          npm install tigerbeetle-node@${{ inputs.version }}
+
+      - name: Alert if anything failed
+        if: failure()
+        run: |
+          export URL="${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}" && \
+          echo "Failed!" && \
+          # curl -d "text=Release process for ${{ inputs.version }} failed! See ${URL} for more information." -d "channel=C04RWHT9EP5" -H "Authorization: Bearer ${{ secrets.SLACK_TOKEN }}" -X POST https://slack.com/api/chat.postMessage
```
eatonphil: && \ but then a comment on the last line?

cb22: Slipped in from testing - thanks!

cb22:
Releases are going again, so this should be good to merge when ready.

eatonphil (dismissed):

cb22:
We now pull in all client packages via their actual package managers, rather than hacking around with curl for Java / Dotnet. CC @batiati if you could have a quick look at the steps for those 2!

eatonphil (approved):

PR #978, opened 2023-07-10
VSR: Explicit protocol enum values

sentientwaffle:
Motivation:
This makes it extra clear that removing or reordering `Command`/`CreateAccountResult`/`CreateTransferResult` variants is a breaking protocol change.

### DIFF
--- a/src/tigerbeetle.zig
+++ b/src/tigerbeetle.zig
@@ -99,108 +99,120 @@ pub const TransferFlags = packed struct {
 /// When errors do not have an obvious/natural precedence (e.g. "*_must_be_zero"),
 /// the ordering matches struct field order.
 pub const CreateAccountResult = enum(u32) {
-    ok,
-    linked_event_failed,
-    linked_event_chain_open,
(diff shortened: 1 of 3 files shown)


src/tigerbeetle.zig
```diff
@@ -99,108 +99,120 @@ pub const TransferFlags = packed struct {
...
+    exists_with_different_flags = 15,
+    exists_with_different_user_data = 16,
+    exists_with_different_ledger = 17,
+    exists_with_different_code = 18,
+    exists = 19,
+
+    comptime {
+        for (std.enums.values(CreateAccountResult)) |result| {
+            assert(@enumToInt(result) < std.enums.values(CreateAccountResult).len);
+        }
```
matklad: maybe

```
        for (std.enums.values(CreateAccountResult)) |result, index| {
            assert(@enumToInt(result) == index);
        }
```

?

I think the current assertion only allows permutations, not gaps, but I don't think we would ever _want_ a permutation here anyway?

matklad (approved):

PR #979, opened 2023-07-10
VOPR minor quality of life improvements

matklad:


sentientwaffle (dismissed):

sentientwaffle:
Ah looks like `fmt` failed

build.zig
```diff
@@ -1018,7 +1018,7 @@ const JniTestStep = struct {
                 else if (std.mem.indexOf(u8, ldd_result, "libc") != null)
                     .gnu
                 else {
-                    std.log.err("{s}", .{ ldd_result });
+                    std.log.err("{s}", .{ldd_result});
```
matklad, line 1021: @cb22 hm, haven't we fixed `zig fmt` not gating the main branch? This is pre-existing I think?

sentientwaffle (approved):

build.zig
cb22: Hmmm, it only runs on `src/`

sentientwaffle: It looks like we only check the zig files in `src/`, so `build.zig` itself is not checked.

PR #980, opened 2023-07-10
build: format all the code

matklad:
I _think_ originally we scoped this to src because we were checking out some third-party zig code, but we don't do that anymore now.

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -129,8 +129,8 @@ pub fn build(b: *std.build.Builder) void {
     // The meta-target lint runs them all
     {
         // lint_zig_fmt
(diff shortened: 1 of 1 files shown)


matklad:
I think this _might_ break something, as I don't remember _exactly_ why we scoped this to src, so lets see what CI thinks.

cb22 (approved):
That sounds about right! CI seems green, so LGTM :+1:

eatonphil:
> I think this might break something, as I don't remember exactly why we scoped this to src, so lets see what CI thinks.

I think it's the `jui` code which we are dropping.

PR #981, opened 2023-07-10
bus: ensure we don't depend on messages being zeroed-out

matklad:


matklad:
Not entirely sure this is a good idea TBH.

The primary win here is that, if we accidentally depend on stuff being zero, `@memset(undefined)` will catch us out. The secondady win here is slight perf boost. A lot of our messages are header-only, and zeroing-out the whole body for them seems wasteful. 

The primary drawback here I think is that we expose messages to the clients, and our _users_ might not be prepared to deal with genuinely uninit data, as it usually isn't idiomatic thing in non-Zig languages...

sentientwaffle:
I think this is a good idea -- messages are available to the language client implementations, but iirc the language clients themselves don't expose them to users.

What do you think of setting just the header to undefined when `!constants.verify`?

matklad:
Yeah, great idea!

sentientwaffle (approved):

PR #984, opened 2023-07-10
LSM: Don't specialize TreeType on tree_name; explicit tree ids

sentientwaffle:
###  Motivation

This improves compile times slightly (~10%) and shrinks the binary from 18M to 13M.
Possibly this could marginally improve benchmarks due to better instruction caching, but I didn't measure it. (Probably not significant.)
None of those reasons are too important... but there's no particular reason to specialize based on tree name in the first place. ("Avoid unnecessary specialization" seems like a good rule of thumb.)

### Other Changes

- Rename "tree hash" to "tree id".
- Instead of assigning tree ids as `hash(tree_name)`, the `StateMachine` is responsible for setting explicit unique ids.
  This better matches how other identifiers are assigned (e.g. `Operation`, `Command`).
  And it makes it obvious that changing the id would be a breaking change (i.e. backwards incompatible with the storage format).
  (Previously, renaming a tree would change the id, breaking the layout.)
- Fix: Previously the tree hashes were only checked for uniqueness *within the context of a each Groove*.
  That is, collisions between grooves were possible.
  Uniqueness is now properly verified (still at comptime) across the entire forest.

### DIFF
--- a/src/lsm/groove.zig
+++ b/src/lsm/groove.zig
@@ -119,7 +119,6 @@ comptime {
 fn IndexTreeType(
     comptime Storage: type,
     comptime Field: type,
-    comptime tree_name: [:0]const u8,
     comptime value_count_max: usize,
 ) type {
     const Key = CompositeKey(IndexCompositeKeyType(Field));
@@ -135,7 +134,7 @@ fn IndexTreeType(
         .secondary_index,
     );
 
-    return TreeType(Table, Storage, tree_name);
+    return TreeType(Table, Storage);
 }
 
 /// A Groove is a collection of LSM trees auto generated for fields on a struct type
@@ -147,6 +146,10 @@ pub fn GrooveType(
     comptime Storage: type,
     comptime Object: type,
     /// An anonymous struct instance which contains the following:
+    ///
+    /// - ids: { .tree = u128 }:
+    ///     An anonymous struct which maps each of the groove's trees to a stable, forest-unique,
+    ///     tree identifier.
     ///
     /// - value_count_max: { .field = usize }:
     ///     An anonymous struct which contains, for each field of `Object`,
@@ -184,11 +187,9 @@ pub fn GrooveType(
         }
 
         if (!ignored) {
-            const tree_name = @typeName(Object) ++ "." ++ field.name;
             const IndexTree = IndexTreeType(
                 Storage,
                 field.field_type,
-                tree_name,
                 @field(groove_options.value_count_max, field.name),
             );
             index_fields = index_fields ++ [_]std.builtin.TypeInfo.StructField{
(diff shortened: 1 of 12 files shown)


src/state_machine.zig
```diff
@@ -1299,7 +1326,7 @@ pub fn StateMachineType(
                     .tree_options_object = .{
                         .cache_entries_max = options.cache_entries_accounts,
                     },
-                    .tree_options_id = {}, // No ID tree as there's one already for AccountsMutable.
+                    .tree_options_id = {}, // No ID tree as there's one already for AccountsImmutable.
```
sentientwaffle, line 1341: This didn't change; it was incorrect before.

src/lsm/groove.zig
```diff
@@ -313,21 +305,6 @@ pub fn GrooveType(
...
-        assert(std.mem.indexOf(u128, hashes, hash) == null);
-        hashes = hashes ++ hash;
-    }
-    inline for (std.meta.fields(_IndexTrees)) |field| {
-        const IndexTree = @TypeOf(@field(@as(_IndexTrees, undefined), field.name));
-        const hash: []const u128 = &.{IndexTree.hash};
-        assert(std.mem.indexOf(u128, hashes, hash) == null);
-        hashes = hashes ++ hash;
-    }
-
```
sentientwaffle, line 330: This check was run at the groove scope, not the forest scope.
So it could detect collisions within a single groove, but it would miss hash collisions between multiple grooves.

src/state_machine.zig
```diff
@@ -136,6 +136,13 @@ pub fn StateMachineType(
             Storage,
             AccountImmutable,
             .{
+                .ids = .{
+                    .timestamp = 1,
```
batiati: Suggestion:

Since it's a stable id for the lsm schema, it can be useful to reference those constants from somewhere that isn't a specialized type.

Just an example, it may not be the best place:

```zig
pub const AccountImmutable = extern struct {
    pub const schema_ids = .{
        .timestamp = 1,
        .id = 2,
        .user_data = 3,
        // ...
    };
}
```

The downside is losing the nice sight of all increasing ids declared together, but anyway, it can become unordered if we evolve our schema in the future.

batiati (dismissed):

src/state_machine.zig
sentientwaffle: Great idea!

sentientwaffle:
Ack, just realized that this can be improved more

batiati (dismissed):
Nice :+1:

sentientwaffle:
Ok I just moved `tree_ids` into `constants`.  Apologies for the last-second change :facepalm:

batiati (approved):

PR #985, opened 2023-07-10
VSR/SuperBlock: Trailer IO dedup

sentientwaffle:
Also:
- Move the request/response command mapping from `superblock.Trailer` to `sync.Trailer`. (More relevant).
- Split `SuperBlock.trailer()` into `SuperBlock.trailer_buffer`, `SuperBlockHeader.trailer_checksum`, and `SuperBlockHeader.trailer_size`. (`.trailer()` previously always used `.staging` as its header, but now that isn't always what we want. This is more explicit too.)

The diff in `superblock.zig` makes this refactor look much more extensive than it is.

### DIFF
--- a/src/testing/storage.zig
+++ b/src/testing/storage.zig
@@ -502,7 +502,7 @@ pub const Storage = struct {
         copy_: u8,
     ) *const superblock.SuperBlockHeader {
         const offset = vsr.Zone.superblock.offset(superblock.Area.header.offset(copy_));
-        const bytes = storage.memory[offset..][0..superblock.Area.header.size_max()];
+        const bytes = storage.memory[offset..][0..comptime superblock.Area.header.size_max()];
         return mem.bytesAsValue(superblock.SuperBlockHeader, bytes);
     }
 
--- a/src/testing/storage.zig
+++ b/src/testing/storage.zig
@@ -501,8 +501,8 @@ pub const Storage = struct {
         storage: *const Storage,
         copy_: u8,
     ) *const superblock.SuperBlockHeader {
-        const offset = vsr.Zone.superblock.offset(superblock.areas.header.offset(copy_));
-        const bytes = storage.memory[offset..][0..superblock.areas.header.size_max];
+        const offset = vsr.Zone.superblock.offset(superblock.Area.header.offset(copy_));
+        const bytes = storage.memory[offset..][0..superblock.Area.header.size_max()];
         return mem.bytesAsValue(superblock.SuperBlockHeader, bytes);
     }
 
@@ -610,8 +610,8 @@ pub const Area = union(enum) {
         switch (area) {
             .superblock => |data| SectorRange.from_zone(
                 .superblock,
-                @field(superblock.areas, data.area).offset(data.copy),
-                @field(superblock.areas, data.area).size_max,
+                data.area.offset(data.copy),
+                data.area.size_max(),
             ),
             .wal_headers => |data| SectorRange.from_zone(
                 .wal_headers,
@@ -814,10 +814,10 @@ pub const ClusterFaultAtlas = struct {
         const copy = @divFloor(offset_in_zone, superblock.superblock_copy_size);
         const offset_in_copy = offset_in_zone % superblock.superblock_copy_size;
         const area: superblock.Area = switch (offset_in_copy) {
-            superblock.areas.header.base => .header,
-            superblock.areas.manifest.base => .manifest,
-            superblock.areas.free_set.base => .free_set,
-            superblock.areas.client_sessions.base => .client_sessions,
+            superblock.Area.header.base() => .header,
+            superblock.Area.manifest.base() => .manifest,
+            superblock.Area.free_set.base() => .free_set,
+            superblock.Area.client_sessions.base() => .client_sessions,
             else => unreachable,
         };
 
--- a/src/testing/storage.zig
+++ b/src/testing/storage.zig
@@ -501,8 +501,8 @@ pub const Storage = struct {
         storage: *const Storage,
         copy_: u8,
     ) *const superblock.SuperBlockHeader {
-        const offset = vsr.Zone.superblock.offset(superblock.Area.header.offset(copy_));
-        const bytes = storage.memory[offset..][0..comptime superblock.Area.header.size_max()];
+        const offset = vsr.Zone.superblock.offset(superblock.SuperBlockZone.header.start_for_copy(copy_));
+        const bytes = storage.memory[offset..][0..comptime superblock.SuperBlockZone.header.size_max()];
         return mem.bytesAsValue(superblock.SuperBlockHeader, bytes);
     }
 
@@ -600,7 +600,7 @@ fn verify_alignment(buffer: []const u8) void {
 }
 
 pub const Area = union(enum) {
-    superblock: struct { area: superblock.Area, copy: u8 },
+    superblock: struct { zone: superblock.SuperBlockZone, copy: u8 },
     wal_headers: struct { sector: usize },
     wal_prepares: struct { slot: usize },
     client_replies: struct { slot: usize },
@@ -610,8 +610,8 @@ pub const Area = union(enum) {
         switch (area) {
             .superblock => |data| SectorRange.from_zone(
                 .superblock,
-                data.area.offset(data.copy),
-                data.area.size_max(),
+                data.zone.start_for_copy(data.copy),
+                data.zone.size_max(),
             ),
             .wal_headers => |data| SectorRange.from_zone(
                 .wal_headers,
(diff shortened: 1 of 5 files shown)


src/vsr/superblock.zig
```diff
@@ -1823,37 +1741,33 @@ pub const Area = enum {
...
-    pub const free_set = AreaRange{
-        .base = manifest.base + manifest.size_max,
-        .size_max = superblock_trailer_free_set_size_max,
-    };
-
-    pub const client_sessions = AreaRange{
-        .base = free_set.base + free_set.size_max,
-        .size_max = superblock_trailer_client_sessions_size_max,
-    };
+    pub fn base(area: Area) u64 {
```
matklad: Let's use the same `inline for` here we use for `vsr.Zone`. Also, maybe call this `SuperblockZone`? I think `Zone` and `Area` are the same kind of thing, so using different words for two feels redundant. 

Ah, we also have testing/storage/Area, and that is vsr.Zone + a specific entry in the zone. So, we have Zone and two Areas, and the semantics between the two Areas is different :)

Yeah, it feels like this should be `SuperBlockZone` or maybe `SuperBlock.Zone`

matklad: I am also not happy with base/offset naming here. `superblock.Area.offset` is `vsr.Zone.start`, and `vsr.Zone.offset` is something completely different. And, well, in 

```
return superblock_copy_size * @as(u64, copy) + area.base();
```

`superblock_copy_size * @as(u64, copy)` is base, and `area.base()` is offset :rofl: 

maybe `SuperBlockZone.start` and `SuperBlockZone.start_for_copy()`?

src/vsr/superblock.zig
```diff
@@ -978,133 +979,60 @@ pub fn SuperBlockType(comptime Storage: type) type {
...
-
-            superblock.assert_bounds(offset, buffer.len);
-
-            if (buffer.len == 0) {
-                write_client_sessions_callback(&context.write);
-                return;
+            switch (trailer_) {
+                .manifest => context.superblock.write_trailer(context, .free_set),
+                .free_set => context.superblock.write_trailer(context, .client_sessions),
+                .client_sessions => context.superblock.write_header(context),
```
matklad: Awesome simplification here!

src/vsr/superblock.zig
```diff
@@ -1823,37 +1741,33 @@ pub const Area = enum {
...
+            .free_set => base_free_set,
+            .client_sessions => base_client_sessions,
+        };
+    }
 
-    const AreaRange = struct {
-        base: u64,
-        size_max: u64,
+    pub fn offset(area: Area, copy: u8) u64 {
+        return superblock_copy_size * @as(u64, copy) + area.base();
```
matklad: assert that `copy < superblock_copies`

src/vsr/superblock.zig
```diff
@@ -978,133 +979,60 @@ pub fn SuperBlockType(comptime Storage: type) type {
...
+            const trailer_data_ = superblock.trailer_data(trailer);
 
-            mem.set(u8, buffer[superblock.staging.manifest_size..], 0); // Zero sector padding.
+            const trailer_size_ceil = vsr.sector_ceil(trailer_data_.size);
+            assert(trailer_size_ceil <= trailer.area().size_max());
 
-            assert(superblock.staging.manifest_checksum == vsr.checksum(
-                superblock.manifest_buffer[0..superblock.staging.manifest_size],
-            ));
+            const offset = trailer.area().offset(context.copy.?);
```
matklad: Move offset computation below zerroing-out the buffer, right now is looks like `offset` is related to buffer, but it is not.

src/vsr/superblock.zig
```diff
@@ -924,7 +925,7 @@ pub fn SuperBlockType(comptime Storage: type) type {
 
             context.copy = 0;
             if (context.caller.updates_trailers()) {
-                superblock.write_manifest(context);
+                superblock.write_trailer(context, .manifest);
```
matklad: for clarity, maybe add `superblock.write_trailers()` which delegates to `write_trailer` (which we could rename to `write_next_trailer` to ehphasize that it writes a chain of trailers, rather than just one?)

src/vsr/superblock.zig
```diff
@@ -978,133 +979,60 @@ pub fn SuperBlockType(comptime Storage: type) type {
...
             assert(superblock.queue_head == context);
+            assert(context.write_trailer == null);
 
-            const size = vsr.sector_ceil(superblock.staging.manifest_size);
-            assert(size <= superblock_trailer_manifest_size_max);
+            context.write_trailer = trailer;
 
-            const buffer = superblock.manifest_buffer[0..size];
-            const offset = areas.manifest.offset(context.copy.?);
+            const trailer_data_ = superblock.trailer_data(trailer);
```
matklad: Can we make any area-specific assertions here? Eg, `if area == client_sessions { assert(data.size % ??? == 0)}`?

src/testing/storage.zig
```diff
@@ -502,7 +502,7 @@ pub const Storage = struct {
         copy_: u8,
     ) *const superblock.SuperBlockHeader {
         const offset = vsr.Zone.superblock.offset(superblock.Area.header.offset(copy_));
-        const bytes = storage.memory[offset..][0..superblock.Area.header.size_max()];
+        const bytes = storage.memory[offset..][0..comptime superblock.Area.header.size_max()];
```
matklad: :zap:

src/vsr/replica.zig
```diff
@@ -8192,16 +8192,18 @@ pub fn ReplicaType(
...
-            });
-            assert(trailer_data.size >= parameters.offset);
+            const trailer = for (comptime std.enums.values(vsr.SuperBlockTrailer)) |trailer| {
+                if (command == SyncTrailer.requests.get(trailer)) break trailer;
+            } else unreachable;
+
+            const trailer_buffer_all = self.superblock.trailer_buffer(trailer);
+            const trailer_checksum = self.superblock.staging.trailer_checksum(trailer);
+            const trailer_size = self.superblock.staging.trailer_size(trailer);
+            assert(trailer_size >= parameters.offset);
```
matklad: Should this be a strict inequality? Doesn't make sense to request empty chunk

src/vsr/replica.zig
```diff
@@ -8192,16 +8192,18 @@ pub fn ReplicaType(
...
+            assert(trailer_size <= trailer.area().size_max());
 
             const body_size = @intCast(u32, @minimum(
-                trailer_data.size - parameters.offset,
+                trailer_size - parameters.offset,
                 constants.sync_trailer_message_body_size_max,
             ));
             assert(body_size <= constants.message_body_size_max);
```
matklad, line 8210: similarly, assert that body_size > 0

src/vsr/replica.zig
matklad: We repeat similarly-shaped bit of code in superblock. Maybe we need a pair of helpers? 

```
self.superblock.trailer_data(trailer) -> []u8
self.superblock.trailer_data_sector_aligned(trailer) -> [][constants.sector_size]u8
```

(unrelated, but, yeah, I wonder if it makes sense to enforce, at compile time, that write_sectors accepts some number of sectrors, and not just []u8

src/vsr/superblock.zig
```diff
@@ -1024,11 +1042,11 @@ pub fn SuperBlockType(comptime Storage: type) type {
 
         fn write_trailer_callback(write: *Storage.Write) void {
             const context = @fieldParentPtr(Context, "write", write);
-            const trailer_ = context.write_trailer.?;
+            const trailer = context.trailer.?;
 
-            context.write_trailer = null;
+            context.trailer = null;
```
matklad: Alternatively, instead of going `null -> .manifest -> null -> .free_set -> null -> .client_session -> null`, we could go `null -> .manifest -> .free_set -> .client_session -> null`, as that would narrow down the state transition somewhat.

sentientwaffle:
Oops I missed some of you comment somehow. Will update shortly.

src/vsr/replica.zig
sentientwaffle: `body_size` can be zero in the case of e.g. an empty manifest -- when offset is 0 the syncing replica isn't just requesting the body, they are also discovering the size and checksum of the whole trailer.

I'll add `assert(body_size > 0 or parameters.offset == 0);`.

src/vsr/replica.zig
sentientwaffle: Hmm it seems like all of the uses are just slightly different :thinking: 
1. replica here just wants the bytes within the data; doesn't care about alignment
2. `superblock.write_trailer` wants to round up to the nearest sector -- but I think handling that explicitly there is a good thing, since it needs to be careful to zero out the sector padding
3. `superblock.read_trailer` wants to round up to the nearest sector

I think I'll leave it as-is.

sentientwaffle:
Ok, updated.

src/testing/storage.zig
```diff
@@ -501,8 +501,8 @@ pub const Storage = struct {
         storage: *const Storage,
         copy_: u8,
     ) *const superblock.SuperBlockHeader {
-        const offset = vsr.Zone.superblock.offset(superblock.areas.header.offset(copy_));
-        const bytes = storage.memory[offset..][0..superblock.areas.header.size_max];
+        const offset = vsr.Zone.superblock.offset(superblock.SuperBlockZone.header.offset(copy_));
```
matklad: This shouldn't compile? We renamed `offset` I think?

sentientwaffle: That is a `vsr.Zone`; I didn't rename `vsr.Zone.offset`.

sentientwaffle: Oh wait, I didn't see the inner one

matklad (approved):

PR #987, opened 2023-07-11
Rename tigerbeetledb to tigerbeetle

batiati:
Replaced:

- github.com/tigerbeetledb
- api.github.com/repos/tigerbeetledb
- ghcr.io/tigerbeetledb

Preserved:

Twitter and Youtube links


batiati:
It's failing due to a circular dependency running the Go client samples ... the file `go.mod` in `tigerbeetle-go` still refers to the old name, and only will be updated when this PR is merged.

Shall we update `go.mod` directly in `tigerbeetle-go`?

cb22:
> Shall we update go.mod directly in tigerbeetle-go?

I think we can do that for now. @eatonphil any thoughts on how we could get the go testing to pull from the local repo? If that's even possible?

eatonphil:
> Shall we update go.mod directly in tigerbeetle-go?

Yep

> I think we can do that for now. @eatonphil any thoughts on how we could get the go testing to pull from the local repo? If that's even possible?

The integration tests test from a built version of the repo. There are actually no tests at the moment that test against the "package manager" version of builds. Since you may at any point be making changes to client code you want to make sure you're testing the changes too. That does leave a gap in testing of course.

cb22:
> Since you may at any point be making changes to client code you want to make sure you're testing the changes too. That does leave a gap in testing of course.

I definitely agree with this, but...

> The integration tests test from a built version of the repo. There are actually no tests at the moment that test against the "package manager" version of builds

I'm not sure this is the case? Eg, [here](https://github.com/tigerbeetle/tigerbeetle/actions/runs/5520723336/jobs/10084796160), it references the `tigerbeetle-go` repo which is essentially the package manager for go.

cb22 (approved):

eatonphil:
> I'm not sure this is the case? Eg, [here](https://github.com/tigerbeetle/tigerbeetle/actions/runs/5520723336/jobs/10084796160), it references the tigerbeetle-go repo which is essentially the package manager for go.

It later gets modified to only reference the local version. So I guess judging from there it does catch *some* issues but it wouldn't catch most compile-time or runtime issues.

PR #988, opened 2023-07-11
Adds direction to table_data_iterator

batiati:
This PR adds support to `direction` to `table_data_iterator.zig`, as it is needed for the Scan API, and doing it by now reduces the amount of work needed for further rebase.

Also, changes `level_data_iterator` to accept `direction` and `compaction` to pass always `ascending`.

A few non-related changes:

- Removes the error result for `TableDataIterator.init`.
- Adds `schema.header_from_block` as a helper to obtain the `vsr.Header` pointer from a block, replacing in many places `mem.bytesAsValue(...)`.

### DIFF
--- a/src/lsm/level_data_iterator.zig
+++ b/src/lsm/level_data_iterator.zig
@@ -2,6 +2,7 @@ const std = @import("std");
 const mem = std.mem;
 const math = std.math;
 const assert = std.debug.assert;
+const schema = @import("schema.zig");
 
 const constants = @import("../constants.zig");
 
@@ -9,7 +10,7 @@ const stdx = @import("../stdx.zig");
 const ManifestType = @import("manifest.zig").ManifestType;
 const allocate_block = @import("grid.zig").allocate_block;
 const GridType = @import("grid.zig").GridType;
-const schema = @import("schema.zig");
+const Direction = @import("direction.zig").Direction;
 const TableDataIteratorType = @import("table_data_iterator.zig").TableDataIteratorType;
 
 // Iterates over the data blocks in a level B table. References to the level B
@@ -41,6 +42,7 @@ pub fn LevelTableValueBlockIteratorType(comptime Table: type, comptime Storage: 
 
             // `tables` contains TableInfo references from ManifestLevel.
             tables: []const Manifest.TableInfoReference,
+            direction: Direction,
         };
 
         pub const IndexCallback = fn (it: *LevelTableValueBlockIterator) void;
@@ -66,8 +68,8 @@ pub fn LevelTableValueBlockIteratorType(comptime Table: type, comptime Storage: 
             table_next: Callback,
         },
 
-        pub fn init() !LevelTableValueBlockIterator {
-            var table_data_iterator = try TableDataIterator.init();
+        pub fn init() LevelTableValueBlockIterator {
+            var table_data_iterator = TableDataIterator.init();
             errdefer table_data_iterator.deinit();
 
             return LevelTableValueBlockIterator{
@@ -107,6 +109,7 @@ pub fn LevelTableValueBlockIteratorType(comptime Table: type, comptime Storage: 
                 .grid = context.grid,
                 .addresses = &.{},
                 .checksums = &.{},
+                .direction = context.direction,
             });
         }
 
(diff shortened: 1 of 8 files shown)


src/lsm/schema.zig
```diff
@@ -58,6 +58,14 @@ const block_body_size = block_size - @sizeOf(vsr.Header);
 const BlockPtr = *align(constants.sector_size) [block_size]u8;
 const BlockPtrConst = *align(constants.sector_size) const [block_size]u8;
 
+pub inline fn header_from_block(block: anytype) switch (@TypeOf(block)) {
+    BlockPtr => *vsr.Header,
+    BlockPtrConst => *const vsr.Header,
+    else => unreachable,
+} {
+    return mem.bytesAsValue(vsr.Header, block[0..@sizeOf(vsr.Header)]);
+}
```
sentientwaffle: Excellent helper function!

Right now we use this helper in two different scenarios:
- where we know for certain that the block is valid, and
- in e.g. grid where the block isn't yet validated (i.e. we just read it off disk, and it might be corrupt).

Most of the call sites are the former case, with just a handful from the latter.

If we split it into two different helpers, then the first case can include extra validation (`command == .block`, `op > 0`, `size <= block_size`).

src/lsm/level_data_iterator.zig
```diff
@@ -107,6 +109,7 @@ pub fn LevelTableValueBlockIteratorType(comptime Table: type, comptime Storage:
                 .grid = context.grid,
                 .addresses = &.{},
                 .checksums = &.{},
+                .direction = .ascending,
```
sentientwaffle:
```suggestion
                .direction = context.direction,
```

src/lsm/table_data_iterator.zig
```diff
@@ -87,12 +91,24 @@ pub fn TableDataIteratorType(comptime Storage: type) type {
...
+                } = switch (it.context.direction) {
+                    .ascending => .{
+                        .address = it.context.addresses[0],
+                        .checksum = it.context.checksums[0],
+                    },
+                    .descending => .{
+                        .address = it.context.addresses[it.context.addresses.len - 1],
+                        .checksum = it.context.checksums[it.context.checksums.len - 1],
+                    },
+                };
```
sentientwaffle: Maybe refactor:

```
                const index = switch (it.context.direction) {
                    .ascending => 0,
                    .descending => it.context.addresses.len - 1,
                };
                
                it.callback = .{ .read = callback };
                it.context.grid.read_block(
                    on_read,
                    &it.read,
                    it.context.addresses[index],
                    it.context.checksums[index],
                    .data,
                );
```

src/lsm/table_data_iterator.zig
```diff
@@ -102,12 +118,35 @@ pub fn TableDataIteratorType(comptime Storage: type) type {
...
+                    if (constants.verify) {
+                        const header = schema.header_from_block(block);
+                        assert(header.op == it.context.addresses[it.context.addresses.len - 1]);
+                        assert(header.checksum == it.context.checksums[it.context.checksums.len - 1]);
+                    }
+
+                    it.context.addresses = it.context.addresses[0 .. it.context.addresses.len - 1];
+                    it.context.checksums = it.context.checksums[0 .. it.context.checksums.len - 1];
+                },
+            }
```
sentientwaffle: Since we iterate in either direction now, it might be simpler to add a `Context.index: usize` field and increment/decrement that instead of re-slicing `addresses`/`checksums`.

batiati: If you don't mind, I'd rather keep the slicing logic, it's easy to understand and prevents off-by-one bugs.

src/lsm/schema.zig
batiati: Instead of two different helpers, I restored the use cases where the block can be invalid with `mem.bytesAsValue`, and now the helper signature is `*const` only, which prevents misuse.

I did that mostly because I couldn't find good names for splitting the function into two, and I realized that `mem.bytesAsValue` is in fact great for what we are doing in those places.
Suggestions are accepted 😁

sentientwaffle:
This LGTM, but can you get rid of the merge commits? e.g. by squashing

sentientwaffle (approved):

PR #989, opened 2023-07-11
Binary search over ranges

batiati:
This PR adds the ability to binary search over ranges, following the same semantics of the existing `binary_search_*` functions.

- Renames `binary_search_*_raw` to `binary_search_*_upsert_index`.

- Removes `config.verify` in favor of `constants.verify`, and removes the validation if the input array is sorted.

- Introduces `config.mode = .lower_bound | .upper_bound` to define the behavior of the binary search when the array contains duplicated entries.

- Adds tests for both `.lower_bound` and `.upper_bound` modes, including new test cases for duplicated matches.

- Clarify the comments.

The new functions `binary_search_*_range` and `binary_search_*_range_upsert_indexes` will be used by the Scan API. The point of adding it now is to make the review simpler and to reduce the amount of code to rebase later.

### DIFF
--- a/src/lsm/binary_search.zig
+++ b/src/lsm/binary_search.zig
@@ -5,6 +5,7 @@ const math = std.math;
 const constants = @import("../constants.zig");
 
 pub const Config = struct {
+    verify_sorted: bool = false,
     mode: enum { lower_bound, upper_bound } = .lower_bound,
 };
 
@@ -37,7 +38,7 @@ pub fn binary_search_values_upsert_index(
 ) u32 {
     if (values.len == 0) return 0;
 
-    if (constants.verify) {
+    if (config.verify_sorted) {
         // Input must be sorted by key.
         for (values) |_, i| {
             assert(i == 0 or
@@ -164,6 +165,10 @@ pub inline fn binary_search_keys(
     };
 }
 
+pub const ConfigRange = struct {
+    verify_sorted: bool = false,
+};
+
 pub const BinarySearchRangeUpsertIndexes = struct {
     start: u32,
     end: u32,
@@ -186,6 +191,7 @@ pub inline fn binary_search_values_range_upsert_indexes(
     values: []const Value,
     key_min: Key,
     key_max: Key,
+    comptime config: ConfigRange,
 ) BinarySearchRangeUpsertIndexes {
     assert(compare_keys(key_min, key_max) != .gt);
 
@@ -196,7 +202,7 @@ pub inline fn binary_search_values_range_upsert_indexes(
         compare_keys,
         values,
         key_min,
-        .{ .mode = .lower_bound },
+        .{ .verify_sorted = config.verify_sorted, .mode = .lower_bound },
     );
 
     if (start == values.len) return .{
@@ -211,7 +217,7 @@ pub inline fn binary_search_values_range_upsert_indexes(
         compare_keys,
         values[start..],
         key_max,
-        .{ .mode = .upper_bound },
+        .{ .verify_sorted = config.verify_sorted, .mode = .upper_bound },
     );
 
     return .{
@@ -226,6 +232,7 @@ pub inline fn binary_search_keys_range_upsert_indexes(
     keys: []const Key,
     key_min: Key,
     key_max: Key,
+    comptime config: ConfigRange,
 ) BinarySearchRangeUpsertIndexes {
     return binary_search_values_range_upsert_indexes(
         Key,
@@ -239,6 +246,7 @@ pub inline fn binary_search_keys_range_upsert_indexes(
         keys,
         key_min,
         key_max,
+        config,
     );
 }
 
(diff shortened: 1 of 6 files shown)


src/lsm/binary_search.zig
```diff
@@ -19,7 +19,7 @@ pub const Config = struct {
 /// * key_value_from(values[i-1]) < key or i == 0
 ///
 /// Doesn't perform the extra key comparison to determine if the match is exact.
-pub fn binary_search_values_raw(
+pub fn binary_search_values_upsert_index(
```
matklad, line 29: While we are at it, could you clarify in the comment which index do we return if there are equal elements? (I wonder if we could/should assert that all elements are unique? Might be good for things like ids and such).

src/lsm/binary_search.zig
```diff
@@ -129,13 +129,148 @@ pub inline fn binary_search_keys(
...
+        }.key_from_key,
+        compare_keys,
+        keys,
+        key_min,
+        key_max,
+        config,
+    );
+}
+
+pub inline fn binary_search_values_range_upsert_indexes(
```
matklad, line 173: Could you add a short doc comment here? In particular, looking just at the signature, the meaning of `key_min` / `key_max` is not immediately obvious to me.

matklad: Ahhhh, this is not `equal_range`, we are looking for two keys at the same time. Still worth a short comment

src/lsm/binary_search.zig
```diff
@@ -129,13 +129,148 @@ pub inline fn binary_search_keys(
...
+    comptime Key: type,
+    comptime Value: type,
+    comptime key_from_value: fn (*const Value) callconv(.Inline) Key,
+    comptime compare_keys: fn (Key, Key) callconv(.Inline) math.Order,
+    values: []const Value,
+    key_min: Key,
+    key_max: Key,
+    comptime config: Config,
+) BinarySearchRangeUpsertIndexes {
+    const start = binary_search_values_upsert_index(
```
matklad, line 184: `assert(key_min <= key_max)`

matklad: Or maybe strict equality, to let the caller branch to non-range version?

src/lsm/binary_search.zig
```diff
@@ -129,13 +129,148 @@ pub inline fn binary_search_keys(
...
+        key_min,
+        config,
+    );
+
+    if (start == values.len) return .{
+        .start = start,
+        .end = start,
+    };
+
+    const end = binary_search_values_upsert_index(
```
matklad, line 199: Yeah, its unclear what's the semantics here if there are duplicates. If we are looking for  `(3, 5)`  in `0, 1, 3, 3, 3, 3, 4, 4, 5, 5, 5, 5, 6, 7`, what result do we expect? 

Seems that we should either assert that there are no dups, OR actually do the lower/upper bound to return the largest range (so, we should use slightly different algos for min and max bound)

src/lsm/binary_search.zig
```diff
@@ -339,3 +562,128 @@ test "binary search: random" {
         try test_binary_search.random_search(rng.random(), i);
     }
 }
+
+test "binary search: range" {
```
matklad: Let's write this in the same style as `binary search: explicit`. Instead of manually `try std.desting.expect` for each result, define a helper functions which takes both the input and the output, and asserts inside. 

Duplication of expectations in the tests makes future refactors painful, so I like to introduce a helper testing function (and it also good to write tests in the same style)

src/lsm/binary_search.zig
batiati: ~Clarified in the comment that we don't expect duplicated entries, and included an `assert` for that when `constants.verify == true`.~

batiati: Clarified the comment and added tests to assert that if we are looking for `(3, 5)` in `0, 1, 3, 3, 3, 3, 4, 4, 5, 5, 5, 5, 6, 7`, the expected result is the first occurrence of `3` (index 2) and the last occurrence of `5` (index 11)

src/lsm/binary_search.zig
```diff
@@ -30,7 +37,7 @@ pub fn binary_search_values_raw(
 ) u32 {
     if (values.len == 0) return 0;
 
-    if (config.verify) {
+    if (constants.verify) {
```
matklad: Hm, not sure that using `constants.verify` is right here, because it changes big-O complexity from `log(N)` to `N`, so this'll probably slow down simulator and such too much. 

It makes sense that binary search has it's own toggle for full verification.

matklad: All the other instancies of verify I think are fine though, as they preserve logarithmic behavior.

batiati: You're right. Now it makes sense why running the fuzz tests was taking so long 😅.

The unit test itself is the only place setting this flag, so I'm not convinced if we should keep this assertion at all, unless we had plans to toggle this flag in a less intrusive way, not dependent on each call site.

matklad: yeah, let's just remove it. This assert checks the caller, but if the caller are tests, then it doesn't really matter.

matklad (approved):

PR #990, opened 2023-07-11
client: correctly limit the number of outstanding requests

matklad:
Before, we used `TooManyOutstandingRequests` to signal this condition. The problem with that is that by the time we get to invoking user's callback, its already too late --- the code might have blown up with an assertion in `get_message`.

To fix this, check the number of outstanding requests in `get_message` instead. This also allows us to remove the error condition from the request callback altogether.

### DIFF
--- a/src/clients/c/tb_client/context.zig
+++ b/src/clients/c/tb_client/context.zig
@@ -81,7 +81,7 @@ pub fn ContextType(
             return null;
         }
 
-        const PacketError = Client.Error || error{
+        const PacketError = error{
             TooMuchData,
             InvalidOperation,
             InvalidDataSize,
@@ -95,7 +95,6 @@ pub fn ContextType(
         addresses: []const std.net.Address,
         io: IO,
         message_pool: MessagePool,
-        messages_available: usize,
         client: Client,
 
         on_completion_ctx: usize,
@@ -179,7 +178,6 @@ pub fn ContextType(
             );
             errdefer context.client.deinit(context.allocator);
 
-            context.messages_available = constants.client_request_queue_max;
             context.on_completion_ctx = on_completion_ctx;
             context.on_completion_fn = on_completion_fn;
             context.implementation = .{
@@ -242,12 +240,10 @@ pub fn ContextType(
         }
 
         pub fn request(self: *Context, packet: *Packet) void {
-            assert(self.messages_available > 0);
+            assert(self.client.messages_available > 0);
 
             const message = self.client.get_message();
             defer self.client.unref(message);
-
-            self.messages_available -= 1;
 
             // Get the size of each request structure in the packet.data:
             const event_size: usize = operation_event_size(packet.operation) orelse {
@@ -286,7 +282,7 @@ pub fn ContextType(
         fn on_result(
             raw_user_data: u128,
             op: Client.StateMachine.Operation,
-            results: Client.Error![]const u8,
+            results: []const u8,
         ) void {
             const user_data = @bitCast(UserData, raw_user_data);
             const self = user_data.self;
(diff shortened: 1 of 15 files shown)


src/clients/c/tb_client/context.zig
```diff
@@ -179,7 +178,6 @@ pub fn ContextType(
             );
             errdefer context.client.deinit(context.allocator);
 
-            context.messages_available = constants.client_request_queue_max;
```
matklad, line 182: I _think_ there was a subtle bug  here --- there's also `.register` request, which _won't_ be counted by this counter. New impl should be fine (but `.register` is an extra gotcha for the users to be aware of).

src/clients/c/tb_client/context.zig
```diff
@@ -301,19 +297,14 @@ pub fn ContextType(
             packet: *Packet,
             result: PacketError![]const u8,
         ) void {
-            self.messages_available += 1;
-            assert(self.messages_available <= constants.client_request_queue_max);
+            assert(self.client.messages_available <= constants.client_request_queue_max);
 
             // Signal to resume sending requests that was waiting for available messages.
-            if (self.messages_available == 1) self.thread.signal.notify();
+            if (self.client.messages_available == 1) self.thread.signal.notify();
```
matklad, line 303: This is subtle. In  the `Client.zig`, I actaully allow the users to retain messages beyond request-response itself. Eg, the user can do

```
const message = client.get_message();
some_global_var.message = message.ref();
client.request(message...)
```

So a request might become available either when a reply arrives, or when the user drops their outstanding reference. If the latter case happened here, that would have been a lost wakeup. 

However, tb_client fully controlls message and it doesn't retain references elsewher.

src/vsr/client.zig
```diff
@@ -276,13 +282,23 @@ pub fn Client(comptime StateMachine_: type, comptime MessageBus: type) type {
             if (was_empty) self.send_request_for_the_first_time(message);
         }
 
-        /// Acquires a message from the message bus if one is available.
-        pub fn get_message(self: *Self) *Message {
+        /// Acquires a message from the message bus.
+        ///
+        /// The caller must ensure that a message is available.
+        // FIXME: change back to get_message.
+        pub fn touch_get_message(self: *Self) *Message {
```
matklad: Renamed just to have the CI yell at me if I missed a callsite, will rename back before merge.

src/vsr/client.zig
```diff
@@ -18,15 +20,11 @@ pub fn Client(comptime StateMachine_: type, comptime MessageBus: type) type {
...
-        pub const Error = error{
-            TooManyOutstandingRequests,
-        };
-
         pub const Request = struct {
             pub const Callback = fn (
                 user_data: u128,
                 operation: StateMachine.Operation,
-                results: Error![]const u8,
+                results: []const u8,
```
sentientwaffle, line 27: :tada:

sentientwaffle (commented):
LGTM once the temporary rename is fixed.

matklad:
Adjusted the node client as well. 

For all other clients, there's no other user-visible changes, because they are based on C client, which already:

- emits too-much-concurrency synchroneously
- transparentl retries failed requests

For node, the situation is different --- submitting request can now throw a new error, and request completion callback now always gets undefined as an error. I didn't change the signature of callbacks, though, because I think we are rewriting this to be based on the C client anyway?

matklad:
Ohhh, I can now actually kill the whole retry thing in our C client!

batiati (approved):
It was nice to get rid of the `retry` list, now the boundary between threads is much more simple to understand: application --> submitted at one side and submitted --> request at the other side.

PR #991, opened 2023-07-11
LSM/Grid: Split read_block_from_{cache,storage,cache_or_storage}

sentientwaffle:
Currently all callers use `read_block_from_cache_or_storage`, which is equivalent to the old behavior (next tick, cache, storage). But the new APIs allow for reading synchronously from the cache (useful for prefetch optimization), or skipping the cache entirely (useful for prefetch optimization and disk scrubbing).

### DIFF
--- a/src/lsm/grid.zig
+++ b/src/lsm/grid.zig
@@ -97,6 +97,7 @@ pub fn GridType(comptime Storage: type) type {
             /// Used to verify that the checkpoint does not advance while the read is in progress.
             checkpoint_id: u128,
 
+            read_cache: bool,
             pending: ReadPending = .{},
             resolves: FIFO(ReadPending) = .{ .name = null },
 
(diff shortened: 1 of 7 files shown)


src/lsm/grid.zig
```diff
@@ -160,6 +161,8 @@ pub fn GridType(comptime Storage: type) type {
         cache: Cache,
         /// Each entry in cache has a corresponding block.
         cache_blocks: []BlockPtr,
+        // TODO
```
cb22: TODO?

cb22: Ah, seems `cache_borrows` isn't used yet?

sentientwaffle: Oops, I didn't mean to commit that at all. Removed. (Also rebased since there were a couple conflicts with `main`.)

cb22 (approved):

PR #997, opened 2023-07-12
k_way_merge refactor

batiati:
This PR adds support to streams that need to be refilled during iteration by `pop()`.
~Also, renames errors `Empty` and `Drained` to `EOF` and `Again`.~

The common pattern for using it in a `while` loop was kept unchanged:
```zig
while(kway.pop() catch buffer_read()) |value| {

}
```

This feature is needed by the Scan API. It's being added now to simplify the review and reduce the amount of code to be rebased later.

### DIFF
--- a/src/lsm/k_way_merge.zig
+++ b/src/lsm/k_way_merge.zig
@@ -5,7 +5,7 @@ const mem = std.mem;
 
 const Direction = @import("direction.zig").Direction;
 
-pub fn KWayMergeIterator(
+pub fn KWayMergeIteratorType(
     comptime Context: type,
     comptime Key: type,
     comptime Value: type,
@@ -14,6 +14,9 @@ pub fn KWayMergeIterator(
     comptime k_max: u32,
     /// Peek the next key in the stream identified by stream_index.
     /// For example, peek(stream_index=2) returns user_streams[2][0].
+    /// Returns Drained if the stream was consumed and
+    /// must be refilled before calling peek() again.
+    /// Returns Empty if the stream was fully consumed and reached the end.
     comptime stream_peek: fn (
         context: *const Context,
         stream_index: u32,
@@ -86,8 +89,8 @@ pub fn KWayMergeIterator(
             return it.k == 0;
         }
 
-        pub fn pop(it: *Self) ?Value {
-            while (it.pop_internal()) |value| {
+        pub fn pop(it: *Self) error{Drained}!?Value {
+            while (try it.pop_internal()) |value| {
                 const key = key_from_value(&value);
                 if (it.previous_key_popped) |previous| {
                     switch (compare_keys(previous, key)) {
@@ -104,7 +107,7 @@ pub fn KWayMergeIterator(
             return null;
         }
 
-        fn pop_internal(it: *Self) ?Value {
+        fn pop_internal(it: *Self) error{Drained}!?Value {
             if (it.k == 0) return null;
 
             // We update the heap prior to removing the value from the stream. If we updated after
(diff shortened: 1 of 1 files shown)


src/lsm/k_way_merge.zig
```diff
@@ -225,10 +228,10 @@ fn TestContext(comptime k_max: u32) type {
             return math.order(a, b);
         }
 
-        fn stream_peek(context: *const Self, stream_index: u32) error{ Empty, Drained }!u32 {
+        fn stream_peek(context: *const Self, stream_index: u32) error{ Again, EOF }!u32 {
             // TODO: test for Drained somehow as well
```
cb22: Is this comment still relevant?

src/lsm/k_way_merge.zig
```diff
@@ -14,10 +14,13 @@ pub fn KWayMergeIterator(
     comptime k_max: u32,
     /// Peek the next key in the stream identified by stream_index.
     /// For example, peek(stream_index=2) returns user_streams[2][0].
+    /// returns error.Again if the stream was consumed and
```
cb22:
```suggestion
    /// Returns error.Again if the stream was consumed and
```

src/lsm/k_way_merge.zig
```diff
@@ -14,10 +14,13 @@ pub fn KWayMergeIterator(
...
+    /// must be refilled before calling peek() again.
+    /// returns error.EOF if the stream was fully consumed and reached EOF.
```
cb22:
```suggestion
    /// Returns error.EOF if the stream was fully consumed and reached EOF.
```

cb22 (commented):
Just some minor nits, LGTM otherwise!

src/lsm/k_way_merge.zig
batiati: It is, we still need to write a test case for consuming/refilling the streams during iteration.
I'm going to do that as part of the Scan PR.

cb22 (dismissed):

src/lsm/k_way_merge.zig
```diff
@@ -71,8 +74,8 @@ pub fn KWayMergeIterator(
                 it.keys[it.k] = stream_peek(context, stream_index) catch |err| switch (err) {
                     // On initialization, the streams should either have data already
                     // buffered up to peek or be empty and have no more values to produce.
-                    error.Drained => unreachable,
-                    error.Empty => continue,
+                    error.Again => unreachable,
+                    error.EOF => continue,
```
sentientwaffle: Imo these new names are not an improvement.

- `EOF` stands for "End of File", which is not relevant to tigerbeetle.
- `Again` conveys less information than "drained". Unlike drained, it doesn't convey why no value was returned.

sentientwaffle (changes requested):

src/lsm/k_way_merge.zig
batiati: I agree, `EOF` is a bad choice. But not sure about `Again` vs `Drained`.
The point of renaming was to reduce the confusion between `Empty` and `Drained`, since if it's drained, it's also empty, so the difference is not obvious. 

The name `Again` came up following Linux's `EAGAIN` idea: there's no data available, we need to do some external work and call it again. Before that, I was calling it `Refill`.

-----------

But now I realized that [we already had this conversation before](https://github.com/tigerbeetle/tigerbeetle/commit/bf23d70a188ac3c3233de6553c61cfbb32507012), where it was named `Buffering` 🙃, I guess we are just bikeshedding at this point, let's keep `Empty` and `Drained`.

sentientwaffle (approved):
LGTM; thanks!

PR #998, opened 2023-07-13
Handle partial release failures in next_version.sh

cb22:
Currently `next_version.sh` only looks at git tags. However, when using draft releases (#992) this causes issues because a client release can fail and then the git tag is never created.

With this, we check our package managers too.

### DIFF
--- a/.github/ci/next_version.sh
+++ b/.github/ci/next_version.sh
@@ -6,10 +6,31 @@ set -eu
 MAJOR=0
 MINOR=13
 
+# Our Docker tags and GH releases can be overwritten, but if we've had a partial release
+# due to a failure for dotnet, npm or Java, we need to take that version into account too.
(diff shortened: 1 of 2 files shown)


matklad:
Modern distributed systems is implementing two-phase commit in .yaml :D

matklad (approved):
Looks good to me, but I am wondering if we can do something less complex here...

cb22:
> Modern distributed systems is implementing two-phase commit in .yaml :D

Don't tempt me to write `vsr.sh` :D

cb22:
> Looks good to me, but I am wondering if we can do something less complex here...

Yeah - I'm not too happy with it, but I think a lot of these problems will go away when we have a less frequent / more curated release process, so I'm not too sure how much time we should spend on it...

PR #1001, opened 2023-07-13
LSM/Grid: Fix header_from_block crash on repair

sentientwaffle:
`read_block_repair_tick_callback` may be called with a zero-header block (`command ≠ .block`) to indicate that a block was found, but not the one we want. Fixes vopr seed `5456263921243472075`.


batiati (approved):
From https://github.com/tigerbeetle/tigerbeetle/pull/988#discussion_r1261213429
> now the helper signature is *const only, which prevents misuse.

I wonder if I covered all cases correctly since my assumption was obviously false!

PR #1002, opened 2023-07-13
VSR/Sync: Fix trailer_size assert

sentientwaffle:
See `8379206930496250648` fails on `assert(trailer_size > parameters.offset);` in `send_sync_trailer()`. `trailer_size == parameters.offset` when both are `0`.


matklad:
Maybe explicitly check that both are zeros to make sure that the only reason for an empty part if  the whole thing is empty?

matklad (approved):

PR #1006, opened 2023-07-14
Add retries for npm and Java

cb22:
Should hopefully be the last of this...

### DIFF
--- a/.github/workflows/release_validate.yml
+++ b/.github/workflows/release_validate.yml
@@ -19,6 +19,18 @@ jobs:
           submodules: recursive
           fetch-depth: 0
 
+      - name: Set up Dotnet
+        uses: actions/setup-dotnet@v3
+        with:
+          dotnet-version: 6.0.x
+
+      - name: Set up Java
+        uses: actions/setup-java@v3
+        with:
+          java-version: '11'
+          distribution: 'temurin'
+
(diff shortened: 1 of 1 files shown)


.github/workflows/release_validate.yml
```diff
@@ -41,37 +53,36 @@ jobs:
...
-          java-version: '11'
-          distribution: 'temurin'
-
+      # Java, Node and Dotnet might take a bit to reflect.
       - name: Check Client Java
         run: |
           mkdir -p /tmp/java && \
           cat src/clients/java/validation_pom.xml | sed 's|{{VERSION}}|${{ inputs.version }}|' > /tmp/java/pom.xml && \
-          cd /tmp/java && mvn dependency:copy-dependencies && \
-          test -e target/dependency/tigerbeetle-java-${{ inputs.version }}.jar
```
batiati, line 67: You dropped the part `test -e target/dependency/tigerbeetle-java-${{ inputs.version }}.jar` ... was that intentional?

batiati (approved):
Please check my comment about the dropped line. LGTM if it was intentional.

.github/workflows/release_validate.yml
cb22: Yep! There was no need for it since `mvn` will exit with a non-zero exit code, and chaining it together would make the line more confusing (or require extra brackets)

PR #1010, opened 2023-07-14
use constants.zig as a source of truth for config

matklad:
Couple of bits of code were using `vsr.config.configs.current` to get user-specified compile time configuration. This reads kafkaesque, and also makes it less clear what should one use config.zig or constants.zig?

To fix that, remove `pub const config` from the vsr interface, and re-export required bits from the `constants`.

As a positive side effect, we also gained an actual `config.BuildOptions` type which clearly specifies all bits that can be tweaked at compile time using build.zig options or `vsr_options` struct in the root file.

### DIFF
--- a/src/config.zig
+++ b/src/config.zig
@@ -8,11 +8,49 @@
 
 const builtin = @import("builtin");
 const std = @import("std");
+const assert = std.debug.assert;
 
 const root = @import("root");
-// Allow setting build-time config either via `build.zig` `Options`, or via a struct in the root file.
-const build_options =
-    if (@hasDecl(root, "vsr_options")) root.vsr_options else @import("vsr_options");
+
+const BuildOptions = struct {
+    config_base: ConfigBase,
+    config_log_level: std.log.Level,
+    tracer_backend: TracerBackend,
+    hash_log_mode: HashLogMode,
+    config_aof_record: bool,
+    config_aof_recovery: bool,
+};
+
+// Allow setting build-time config either via `build.zig` `Options`, or via a struct in the root
+// file.
+const build_options: BuildOptions = blk: {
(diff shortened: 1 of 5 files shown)


src/vsr.zig
```diff
@@ -16,7 +16,6 @@ pub const storage = @import("storage.zig");
 pub const tigerbeetle = @import("tigerbeetle.zig");
 pub const time = @import("time.zig");
 pub const tracer = @import("tracer.zig");
-pub const config = @import("config.zig");
```
matklad, line 19: This is the important line

src/vsr.zig
```diff
@@ -1325,7 +1324,7 @@ test "quorums" {
 /// replica ids for the initial cluster deterministically.
 pub fn root_members(cluster: u32) [constants.nodes_max]u128 {
     const IdSeed = packed struct {
-        cluster_config_checksum: u128 = config.configs.current.cluster.checksum(),
```
matklad, line 1328: and that's what motivated this change during today's call with jorangreef

src/config.zig
```diff
@@ -8,11 +8,49 @@
...
+            @field(result, field.name) = launder_type(
+                field.field_type,
+                @field(vsr_options, field.name),
+            );
+        }
+        break :blk result;
+    }
+};
+
+fn launder_type(comptime T: type, comptime value: anytype) T {
```
sentientwaffle, line 44: Great function name :grin:

sentientwaffle (approved):

src/config.zig
matklad: Inspired by https://en.cppreference.com/w/cpp/utility/launder

PR #1024, opened 2023-07-17
build: add infrastructure for finding memcpy calls

matklad:
This adds a script, copyhound, which parses LLVM IR `.ll` files, finds
calls to `memcpy` and prints names of functions which do large memcpys.

This isn't integrated anywhere, so it is likely to bitrot. However, it
is already quite useful for ad-hoc analysis, so I'd rather we add it to
main already.

I can also see this growing to more feature-full LLVM analysis in the
future (eg, we can try to guard against recursion, or compute stack
sizes).

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -9,6 +9,7 @@ const Shell = @import("./src/shell.zig");
 pub fn build(b: *std.build.Builder) void {
     const target = b.standardTargetOptions(.{});
     const mode = b.standardReleaseOptions();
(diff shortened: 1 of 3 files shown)


stack-copies.zig
```diff
@@ -0,0 +1,19 @@
+const Big = struct {
```
cb22: Left this in?

matklad: :sweat_smile:

cb22 (approved):
This is great :tada:

PR #1026, opened 2023-07-17
perf: replace slow meta.eql with fast eql_repr

matklad:
meta.eql has two problems:

- it often forces a memcpy because it takes arguments by value
- it does field-wise short-circuiting comparison

However, in our _actual_ usage, we use it to compare bag-of-bytes types. So it would be faster and simpler to just compare the underlying memory directly.

### DIFF
--- a/src/stdx.zig
+++ b/src/stdx.zig
@@ -237,6 +237,55 @@ pub const log = if (builtin.is_test)
 else
     std.log;
 
+/// Compare two values by directly comparing the underlying memory.
+///
+/// Assert at compile time that this is a reasonable thing to do for a given `T`. That is, check
+/// that:
+///   - `T` doesn't have any non-deterministic padding,
+///   - `T` doesn't embed any pointers.
+pub fn equal_bytewise(comptime T: type, a: *const T, b: *const T) bool {
+    comptime assert(std.meta.trait.hasUniqueRepresentation(T));
+    comptime assert(!has_pointers(T));
+    comptime assert(@sizeOf(T) * 8 == @bitSizeOf(T));
+
+    // Pick the biggest "word" for word-wise comparison, and don't try to early-return on the first
+    // mismatch, so that a compiler can vectorize the loop.
+
+    const Word = inline for (.{ u64, u32, u16, u8 }) |Word| {
+        if (@alignOf(T) >= @alignOf(Word) and @sizeOf(T) % @sizeOf(Word) == 0) break Word;
+    } else unreachable;
+
+    const a_words = std.mem.bytesAsSlice(Word, std.mem.asBytes(a));
+    const b_words = std.mem.bytesAsSlice(Word, std.mem.asBytes(b));
+    assert(a_words.len == b_words.len);
+
+    var total: Word = 0;
+    for (a_words) |a_word, i| {
+        const b_word = b_words[i];
+        total |= a_word ^ b_word;
+    }
+
+    return total == 0;
+}
+
+fn has_pointers(comptime T: type) bool {
+    switch (@typeInfo(T)) {
+        .Pointer => return true,
+        // Be conservative.
+        else => return true,
+
+        .Bool, .Int, .Enum => return false,
+
+        .Array => |info| return comptime has_pointers(info.child),
+        .Struct => |info| {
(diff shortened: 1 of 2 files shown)


matklad:
I am a bit fried-brained right now, so want to revisit this tommorrow, but this looks basically good to me.

src/stdx.zig
```diff
@@ -237,6 +237,55 @@ pub const log = if (builtin.is_test)
 else
     std.log;
 
+/// Compare two values by directly comparing the underlying memory.
+///
+/// Assert at compile time that this is a reasonable thing to do for a given `T`. That is, check
+/// that:
+///   - `T` doesn't have any non-deterministic padding.
+///   - `T` doesn't embed any pointers.
+pub fn eql_repr(comptime T: type, a: *const T, b: *const T) bool {
```
sentientwaffle: This function LGTM overall, though I don't love this name. But I also can't think of anything better :shrug:

src/stdx.zig
```diff
@@ -237,6 +237,55 @@ pub const log = if (builtin.is_test)
...
+    var total: Int = 0;
+    for (a_words) |a_word, i| {
+        const b_word = b_words[i];
+        total |= a_word ^ b_word;
+    }
+
+    return total == 0;
+}
+
+fn pointer_free(comptime T: type) bool {
```
sentientwaffle: Nit, but when I hear "free" in the context of pointers, I typically think of alloc/free.
But in this case it is pretty easy to tell what is going on, so it's ok as-is.

sentientwaffle (dismissed):

matklad:
Came up with somewhat better names.

I didn't replace other calls to `mem.eql`, as they don't show up in my memcpy profiler. Though, maybe we should in general perfer `equal_bytewise` to `meta.eql`? I like how it has a more tight semantics, additionally asserting that the type is a contiguous slice of bytes.

src/stdx.zig
jorangreef: I was also wondering about how we can improve the name, also to avoid abbreviation.

`stdx.equal` would be fine I think, because we're testing for equality in the strictest possible sense here, the actual memory representation. So any relaxed version that does less can pay the cost of a longer name.

jorangreef: Is there anyway we can determine the type of `a` or `b` and comptime assert they are the same, to avoid passing in the comptime type? Perhaps there are some downsides to this.

jorangreef: I really like how we're moving to `*const` style here.

matklad: The new name is `equal_bytewise`. 

>Is there anyway we can determine the type of a or b and comptime assert they are the same, to avoid passing in the comptime type? Perhaps there are some downsides to this.

I think we could, but I chose not to. The problem with this sort of functions is that it's not clearer whethere they operate on bytes pointed to by the pointer, or on the pointer itself. 

The best worst example here is asBytes, which is a genuinely scary footgun: 

https://github.com/ziglang/zig/issues/2904#issuecomment-1483914130

I _think_ in this particular case we can actually be safe, because we forbid pointers altogether, but it still feels overall best to type it out in full.

src/stdx.zig
```diff
@@ -237,6 +237,55 @@ pub const log = if (builtin.is_test)
 else
     std.log;
 
+/// Compare two values by directly comparing the underlying memory.
+///
+/// Assert at compile time that this is a reasonable thing to do for a given `T`. That is, check
+/// that:
+///   - `T` doesn't have any non-deterministic padding,
+///   - `T` doesn't embed any pointers.
+pub fn equal_bytewise(comptime T: type, a: *const T, b: *const T) bool {
```
sentientwaffle, line 246: Great name!

sentientwaffle (approved):

src/stdx.zig
jorangreef: I prefer `equal` (for the reasons above) or even `equal_bytes` to `equal_bytewise` (less syllables, shorter).

jorangreef: Agreed on not dropping the type, to be explicit.

matklad: love equal_bytes, submitted a PR!

PR #1036, opened 2023-07-18
vsr: assert that ViewChange headers are valid

matklad:
The most interesting case here is when we load superblock from disk. We do check the overall checksum, so seeing an invalid header there would be surprising, but still feels better to double check.

Two other paths to load a header from disk are Journal.read_prepare, and Journal.redundant_headers, and we do verify `.invalid() == null` there.

### DIFF
--- a/src/vsr.zig
+++ b/src/vsr.zig
@@ -1380,8 +1380,9 @@ pub const Headers = struct {
     pub fn dvc_header_type(header: *const Header) enum { blank, valid } {
         if (std.meta.eql(header.*, Headers.dvc_blank(header.op))) return .blank;
 
-        assert(header.command == .prepare);
         if (constants.verify) assert(header.valid_checksum());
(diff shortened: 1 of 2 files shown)


src/vsr.zig
```diff
@@ -1381,6 +1381,7 @@ pub const Headers = struct {
         if (std.meta.eql(header.*, Headers.dvc_blank(header.op))) return .blank;
 
         assert(header.command == .prepare);
+        assert(header.invalid() == null);
         if (constants.verify) assert(header.valid_checksum());
```
sentientwaffle: Tests are failing with `operation == .reserved` :joy: 

Also, unrelated to this change, but it probably makes sense to move 

```
        if (constants.verify) assert(header.valid_checksum());
```

up above the other two  assertions.

sentientwaffle (approved):

PR #1038, opened 2023-07-19
Add ACID question to FAQ

eatonphil:


docs/FAQ.md
```diff
@@ -144,3 +144,23 @@ TigerBeetle](https://github.com/tigerbeetle/tigerbeetle#quickstart),
...
+
+Yes. Let's discuss each part:
+
+1. Atomicity: TigerBeetle allows events within a batch to be
+   [linked](./reference/transfers.md#flagslinked). When an event fails
+   within a linked chain of events, all events are rolled back.
+
+2. Consistency: TigerBeetle's state machine ensures that accounts and
+   transfers remain consistent. When you create a transfer, associated
+   account credits and debits are guaranteed to be up-to-date.
```
sentientwaffle: I'd like to work "invariant" and/or "double-entry" in here somehow. I think we can be more precise than "up-to-date".

:thinking: Something along the lines of "creating a transfer maintains all account invariants (e.g. balanced debits/credits)"? 

Also, maybe link to `./design/consistency.md`.

eatonphil: > 2. Consistency: Accounts and transfers are immutable. Once created,
   users can no longer directly modify them. Creating a transfer
   maintains all account invariants (e.g. balanced
   debits/credits). See
   [./design/consistency.md](./design/consistency.md) for more
   details.

Any better? Feels sort of not to the point to me.

sentientwaffle: Hmm, maybe I'm misremembering ACID :sweat_smile: 
I thought consistency was about maintaining database invariants.

I agree that the "users can no longer directly modify them" isn't super relevant though.
Maybe just your original text + the link at the end? Unless I'm misunderstanding and ACID consistency != `consistency.md`'s consistency, in which case we can leave it out.

eatonphil: No I think you're right I just am not sure if everyone will connect account invariants to consistency without like an additional phrase or sentence.

docs/FAQ.md
```diff
@@ -144,3 +144,23 @@ TigerBeetle](https://github.com/tigerbeetle/tigerbeetle#quickstart),
...
+
+3. Isolation: All events within a batch, across all clients, are
+   serialized into TigerBeetle's state machine. There are no
+   concurrent events in TigerBeetle, even across clients.
+
+4. Durability: TigerBeetle `fsync()`s data and replicates it among a
+   quorum of replicas in a cluster before indicating success to the
+   user.
```
matklad: This doesn't feel exactly right: 

- we provide _more_ guarantees that `fsync` does 
- we do that by _not_ using `fsync`and relying on direct IO

from DESIGN.md:

> TigerBeetle **exceeds the fsync durability of a single disk** and the hardware of a single server because disk firmware can contain bugs and because single server systems fail.

matklad: Marketing-wise, I think it would build trust if we specify the extent of durability. ACI are binary properties, while D is a slider (there's no durability if a sufficiently large asteroid hits the Earth). By showing that we know exactly where our D ends, we build trust. There's also angle that we tolerate storage faults, so that we again explicitly thinking through durability limits

We can convey  the following: 

- on a single replica, we use direct IO, so:
  - a replica only acknowledges transaction once the data hits the disk (fsync gives you this)
  - if a hardware error occurs, the replcia noticies it and doesn't ack (fsync _dosen't_ give you this)
- on a cluster level, a transaction is only acknowledge if a quorum of replicas acknoledge it
- once a cluster acks a transaction, it is never "lost", even if subsequently all those quorum replicas fail (the cluster would stop processing transactions when that happens, )
- after saving transaction to disk, we don't blidly assume that it is there, but take into account the possibility that the data gets corrupted on some subsequent read, and actively repair it. If at least one replcia has a comitted transaction, it will be repaired to the whole cluster. 


Not sure how to best compress that to a couple of pithy sentences, but I think the core idea we want to convey is:

> unlike literary everyone else, we know that disk fail, and the data we write today might not be available tomorrow. We actively work against such entropy by taking advantage of the cluster-wide storage: a bit of data should get corrupted on all machines to get lost, and even in that case the system would safely halt.

docs/FAQ.md
jorangreef: Great point re: not conflating with C in ACID.

Here, the C in ACID that we provide is Strict Serializability.

The "Financial Consistency" of the State Machine is something even stronger than this, it's domain-specific, and at a higher layer. But not related to ACID here. Still worth pointing to it though.

We typically speak of both “Raw Data Consistency” and “Financial Consistency” intentionally so as to contrast them, clarify by contrast, and thereby show also why double entry primitives are the right layer of abstraction (as opposed to the lower level general purpose DB primitives of SQL).

For example, something like (needs work, but to give the idea): “the C in ACID provides raw data consistency across multiple row reads/writes.... however, financial consistency is something more than this... [proceed to explain double entry invariants... such as money cannot be created nor destroyed, merely transferred from one account to another, with the transfer history immutable etc.]"

People might be surprised to learn that Strict Serializability alone, while great, is actually not enough for Financial Consistency, and that TB provides this too.

jorangreef: In other words, the argument is "at least ACID, and the best kind of ACID, but actually also more than ACID, not only ACID”

eatonphil: > the C in ACID provides raw data consistency across multiple row reads/writes.... however, financial consistency is something more than this... [proceed to explain double entry invariants... such as money cannot be created nor destroyed, merely transferred from one account to another, with the transfer history immutable etc.]

Makes sense!

<s>But I think in your discussion you didn't mention (or I didn't understand) what part of TigerBeetle actually guarantees raw data consistency?</s>

Edit: Nevermind, reread and you wrote this in the second sentence.

eatonphil: Ah or is it that "Strict Serializability", being how TigerBeetle is designed, is the answer to the Raw Data Consistency part?

docs/FAQ.md
eatonphil: Got it, that makes sense. I pushed another shot at it incorporating your suggestions.

docs/FAQ.md
eatonphil: Pushed another version clarifying about strict serializability for traditional ACID guarantees at parts of your snippet for financial data guarantees.

docs/FAQ.md
matklad: Hm, I'd say this needs some wordsmithing more, opening with "durability is a sliding scale" is a little bit too blunt I feel :) 

Maybe something like this "Absolute durability is impossible, because all hardware can ultimately fail. TigerBeetle embraces limited disk reliability and maximizes data durability in spite of imperfect disk. Data we write today might not be available tomorrow ....". 

And maybe more emphasis on TigerBeetle rather on we

eatonphil: Cool, updated again. :)

matklad (dismissed):

sentientwaffle (approved):

PR #1040, opened 2023-07-19
Add benchmark notes

eatonphil:


src/benchmark.zig
```diff
@@ -1,3 +1,24 @@
+// This benchmark uses a single client to 1) create `account_count`
```
matklad: Let's use `//!` to make this into a doc comment.

matklad: Let's maybe start with a big picture, how this file fits with the rest of the codebase (is this a micro benchmark? a macro benchmark? a whole bench or a part), and only then go into the detail of what actually this benchmarks. 

Something like 

> benchmark.zig is a program which connects to a tigerbeetle cluster and runs a predefined load against it, measuring observed latencies and throughput. See benchmark.{sh,bat} for the full benchmarking setup.

matklad (dismissed):
Leave some comments, but that's awesome as it. 

At some point, I feel we should add a lint rule that enforces that each `.zig` file starts with `//!` which explains why it exists and what does it do.

src/benchmark.zig
eatonphil: TIL about this. Thanks!

eatonphil: Good idea! Done.

eatonphil: Requesting a re-review after taking your suggestions.

matklad (approved):
:heart:

PR #1041, opened 2023-07-20
size: avoid needless monomorphizations

matklad:
By the amount of LLVM IR, Command.version is one of the larger functions we have. That's because `print_value` is monomoprihsed for each field name and distinct value (as opposed to every type).

### DIFF
--- a/src/tigerbeetle/main.zig
+++ b/src/tigerbeetle/main.zig
@@ -269,8 +269,8 @@ const Command = struct {
 
 fn print_value(
     writer: anytype,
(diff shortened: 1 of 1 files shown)


matklad:
This is mostly to sanity-check that counting llvm-lines is not completely useless. Perhaps the best solution here would be compute the whole string at compile time, but that feels like a more complex change. 

Ah, forgot to mention that it shaves off 336 bytes from the release build for me :)

sentientwaffle (approved):

PR #1043, opened 2023-07-20
Make sure version is set

eatonphil:
I'm not sure why, but the version in releases has stopped being accurate:

```
$ curl -LO https://github.com/tigerbeetle/tigerbeetle/releases/download/0.13.71/tigerbeetle-universal-macos-0.13.71.zip
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
100 1226k  100 1226k    0     0  2099k      0 --:--:-- --:--:-- --:--:-- 2099k
$ unzip tigerbeetle-universal-macos-0.13.71.zip
Archive:  tigerbeetle-universal-macos-0.13.71.zip
replace tigerbeetle? [y]es, [n]o, [A]ll, [N]one, [r]ename: y
  inflating: tigerbeetle
$ ./tigerbeetle version
TigerBeetle version experimental%
```

So this PR makes it a little more explicit.

```
$ ./scripts/build.sh -Dversion=GREAT
$ ./tigerbeetle version
TigerBeetle version GREAT%
```

### DIFF
--- a/.github/workflows/release_artifacts.yml
+++ b/.github/workflows/release_artifacts.yml
@@ -65,7 +65,7 @@ jobs:
         fetch-depth: 0
 
     - name: Build tigerbeetle
-      run: ./scripts/install.sh ${{ matrix.debug }}
+      run: ./scripts/install.sh ${{ matrix.debug }} -Dversion=${{ needs.version.outputs.version }}
       env:
         TARGET: ${{ matrix.target }}
         CPU: ${{ matrix.target == 'aarch64-linux' && 'baseline+aes+neon' || matrix.target == 'aarch64-macos' && 'baseline+aes+neon' || 'x86_64_v3+aes' }}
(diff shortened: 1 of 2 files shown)


eatonphil:
Throwing you in here too @matklad since you're thinking about `tigerbeetle version` I see. 😆 

Now a separate thing is that at some point recently this `tigerbeetle version` reporting broke.

Do you happen to know or see why? I am wondering if it was a `src/shell.zig` change since even locally,  not just during the release, does the `git_tag` flow no longer seem to work at all.

Maybe this PR shouldn't be necessary.

eatonphil:
Ok, I talked myself out of this. Since the git_tag isn't even being set locally correctly we should just fix that. Closing.

PR #1049, opened 2023-07-20
client-node: more useful errors for integration tests

matklad:


eatonphil:
While this is good, I'd suggest we go with https://github.com/tigerbeetle/tigerbeetle/pull/1050 not just because it will show string error messages rather than enum integers, it also gets readers used to the idea that they do actually need to write this code themselves for error handling in real code anyway. Same thing all the other language client sample projects do.

PR #1052, opened 2023-07-21
Update Zig from 0.9.1 to 0.10.1

kprotty:
In preparation for the upcoming Zig 0.11 release, this branch updates the codebase to support 0.10.1 first which had major changes from 0.9. In particular, the self hosted compiler (stage2) became the default instead of the C++ compiler (stage1). Some notes on the upgrade process: https://ziglang.org/download/0.10.0/release-notes.html#How-to-Upgrade


build.zig
```diff
@@ -874,6 +874,8 @@ fn maybe_execute(
     binary_name: []const u8,
 ) void {
     var to_run = std.ArrayList([]const u8).init(allocator);
+    defer to_run.deinit();
```
matklad, line 880: No need to change this, but I am curious, isn't it idiomatic to _not_ deallocate stuff in build.zig? The allocator here is an arena anyway.

src/clients/c/tb_client_header.zig
```diff
@@ -20,43 +20,45 @@ const type_mappings = .{
...
-
-            inline for (type_mappings) |type_mapping| {
-                const ZigType = type_mapping[0];
-                const c_name = type_mapping[1];
-
-                if (info.child == ZigType) {
-                    const prefix = if (@typeInfo(ZigType) == .Struct) "struct " else "";
-                    return prefix ++ c_name ++ "*";
+    comptime {
+        switch (@typeInfo(Type)) {
```
matklad: We can avoid indent here, `comptime switch () {};`

matklad:
After this lands, we should take advantage of the new ability to specify backing integer for packed structs: 

```
pub const AccountFlags = packed struct(u16) {
```

or might as well do that here?

kprotty:
We can do it here.

src/lsm/forest.zig
```diff
@@ -83,8 +83,8 @@ pub fn ForestType(comptime Storage: type, comptime groove_config: anytype) type
             open,
         };
 
-        pub const groove_config = groove_config;
-        pub const Grooves = Grooves;
+        pub const groove_config = groove_cfg;
+        pub const Grooves = _Grooves;
```
matklad, line 87: No need to fix up front, but I wonder if we could revisit this and come up with some better names? `groove_cfg` -> `config`, `_Grooves` -> `GroovesType`

src/shell.zig
```diff
@@ -73,7 +73,7 @@ pub fn echo(shell: *Shell, comptime fmt: []const u8, fmt_args: anytype) void {
         }
         pos += 1;
     }
-    fmt_ansi = fmt_ansi ++ fmt[pos_start..pos] ++ "\n";
+    fmt_ansi = fmt_ansi ++ fmt[pos_start..@min(pos, fmt.len)] ++ "\n";
```
matklad: Wait what? This is compiler bug, right? It _shouldn't_ be incrementing `pos += 1` on last time? Maybe

```zig
    next_pos: inline while (pos < fmt.len) {
        if (fmt[pos] == '{') {
            inline for (comptime std.meta.fieldNames(@TypeOf(ansi))) |field_name| {
                const tag = comptime "{ansi-" ++ field_name ++ "}";
                if (comptime std.mem.startsWith(u8, fmt[pos..], tag)) {
                    fmt_ansi = fmt_ansi ++ fmt[pos_start..pos] ++ @field(ansi, field_name);
                    pos += tag.len;
                    pos_start = pos;
                    continue :next_pos;
                }
            } else {
                pos += 1;
            }
        } else {
            pos += 1;
        }
    }
    // Note: We duplicate `pos += 1` above, otherwise this hits a compiler bug. 
    comptime assert(pos == fmt.len);
    fmt_ansi = fmt_ansi ++ fmt[pos_start..pos] ++ "\n";
```

instead? 

Sorry for writing convoluted code which confused the compiler :fearful:

src/io/windows.zig
```diff
@@ -464,15 +462,27 @@ pub const IO = struct {
...
+                            null,
+                            null,
+                        )) {
+                            os.windows.ws2_32.SOCKET_ERROR => switch (os.windows.ws2_32.WSAGetLastError()) {
+                                .WSAEOPNOTSUPP => unreachable,
+                                .WSAENOTSOCK => unreachable,
+                                else => |err| return os.windows.unexpectedWSAError(err),
+                            },
+                            else => assert(num_bytes == @sizeOf(LPFN_CONNECTEX)),
+                        }
```
matklad, line 487: Could you clarify why we are doing that? That helper stills seems to be there? 

https://github.com/ziglang/zig/blob/e8fa19960243a99cef5d9af94c3db52173787a4e/lib/std/os/windows.zig#L2375

src/lsm/node_pool.zig
```diff
@@ -67,7 +67,8 @@ pub fn NodePool(comptime _node_size: u32, comptime _node_alignment: u13) type {
             assert(pool.free.isSet(node_index));
             pool.free.unset(node_index);
 
-            return pool.buffer[node_index * node_size ..][0..node_size];
+            const node = pool.buffer[node_index * node_size ..][0..node_size];
+            return @alignCast(node_alignment, node);
```
matklad, line 71: unrelated to the PR, but perhaps it woudl make sense to define buffer as 


```
buffer: []align(node_alignment) [node_size]u8,
```

src/lsm/grid.zig
```diff
@@ -834,8 +834,8 @@ pub fn GridType(comptime Storage: type) type {
 
             if (header.checksum != checksum) {
                 log.err(
-                    "expected address={} checksum={} block_type={}, " ++
-                        "found address={} checksum={} block_type={}",
+                    "expected address={} checksum={} block_type={?}, " ++
+                        "found address={} checksum={} block_type={?}",
```
matklad: I think the second ` ?` is not needed here? But maybe its a bug that we do `@enumToInt(header.operation),`?

src/io/windows.zig
kprotty: The `T` it takes in needs to be `fn(...)` as it does `&function`, but it stores the variable as `var function: T` (instead of correctly as `var function: *const T`) which errors out on args incorrectly needing comptime. I just reimplemented it inline for now.

matklad: Ah, so that's a bug in stdlib. Let's add a todo to swith to  the upstream impl once that is fixed.

build.zig
jorangreef: I would rather prefer us to have a habit of always doing defer/deinit. This way the code is correct if it's ever moved to a new context or switched to another allocator.

src/shell.zig
kprotty: I think continues/breaks inside `inline for/while` are not handled properly

src/lsm/tree_fuzz.zig
```diff
@@ -1,13 +1,13 @@
...
 const testing = std.testing;
-const allocator = testing.allocator;
 const assert = std.debug.assert;
 
 const stdx = @import("../stdx.zig");
 const constants = @import("../constants.zig");
 const fuzz = @import("../testing/fuzz.zig");
 const vsr = @import("../vsr.zig");
 const schema = @import("schema.zig");
+const allocator = fuzz.fuzz_allocator;
```
matklad: maybe just `fuzz.allocator` to mirror `testing.allocator`?

src/vsr/superblock_client_sessions.zig
```diff
@@ -134,10 +134,10 @@ pub const ClientSessions = struct {
         assert(source.len <= encode_size_max);
 
         size = std.mem.alignForward(size, @alignOf(vsr.Header));
-        const headers = mem.bytesAsSlice(
+        const headers = @alignCast(@alignOf(vsr.Header), mem.bytesAsSlice(
```
matklad, line 137: Oh wow, `@align(u128) == 8 and @align(extern struct { x: u128}) == 16`, so alignment of header stays the same. Also, we are at 0 here, so we don't depend on the value anyways. 

But still, now that I look at it, using current arch alignment to define serialization format looks deeply suspicious to me. After this lands, I think we should rewrite this to somethign like

```zig
comptime assert @alignOf(vsr.Header) == 16;
size = std.mem.alignForward(size, 16);
```

just so we catch at compile time that the alignment is fixed.

build.zig
matklad: :thinking: not sure I agree here. I think it's pretty challenging to write correct generic allocating zig code. Like, it's not hard to get 9 out of 10 defers right, but getting them all is something we are not particularly good at. 

So I'd say a more robust strategy is to avoid writing generic allocating code --- either allocate all memory at startup, or figure out how to use arenas to do bulk deallocation. 

From this angle, code that clearly _doesn't_ deallocate because we statically know we are using an arena seems preferable to me. 

I like convention in the zig compiler here, there you often have both

```zig
gpa: Allocator, 
arena: Allocator, 
```

:thinking: it seems that Build.zig is actually a great example of what I am worrying here. It has a destroy method, so it looks like maybe it should be a managed object: 

https://github.com/ziglang/zig/blob/77b96231a6bc195cc482d05599e8c20ee01645a6/lib/std/Build.zig#L427-L431

But it also has a bunch of fields which aren't included with the destroy: 

https://github.com/ziglang/zig/blob/77b96231a6bc195cc482d05599e8c20ee01645a6/lib/std/Build.zig#L235-L237

I would say that such inconsistency is a natural consequence of trying to always init/deinit, and that it is worse than clearly not de-initing and using arenas by design instead.

src/state_machine.zig
```diff
@@ -304,7 +304,8 @@ pub fn StateMachineType(
 
         // Since prefetch contexts are used one at a time, it's safe to access
         // the union's fields and reuse the same memory for all context instances.
-        const PrefetchContext = extern union {
```
matklad, line 307: Hm, why did we use `extern union` in the first place here?

src/state_machine.zig
```diff
@@ -484,6 +485,10 @@ pub fn StateMachineType(
             for (accounts) |*a| {
                 self.forest.grooves.accounts_immutable.prefetch_enqueue(a.id);
             }
+
+            // Set the union tag so access via &self.prefetch_context.* doesn't trap.
+            self.prefetch_context = .{ .accounts_immutable = undefined };
```
matklad: Maybe use "Activate" word instead of "Set", as "active union variant" seems to be more specific bit of terminology.

src/tigerbeetle/cli.zig
```diff
@@ -145,7 +145,8 @@ pub fn parse_args(allocator: std.mem.Allocator) !Command {
     var verbose: ?bool = null;
 
     var args = try std.process.argsWithAllocator(allocator);
-    defer args.deinit();
+    // NOTE: do not deinit() as we return the path and this could free/invalidate it.
+    // defer args.deinit();
```
matklad, line 149: Now that ArgsIterator owns the data, it semes we can remove `args_allocated` and use

```
    pub const Start = struct {
        args_allocated: std.process.ArgIterator,
```

?

src/state_machine/workload.zig
```diff
@@ -167,12 +166,20 @@ const transfer_templates = table: {
 pub fn WorkloadType(comptime AccountingStateMachine: type) type {
     const Operation = AccountingStateMachine.Operation;
 
-    const Action = enum(u8) {
-        create_accounts = @enumToInt(Operation.create_accounts),
-        create_transfers = @enumToInt(Operation.create_transfers),
-        lookup_accounts = @enumToInt(Operation.lookup_accounts),
-        lookup_transfers = @enumToInt(Operation.lookup_transfers),
-    };
+    const Action = @Type(.{
```
matklad, line 177: Add comment why we are doing this (some comptime limitation I suppose)?

src/vopr.zig
```diff
@@ -411,19 +407,15 @@ fn parse_args(allocator: mem.Allocator) !Flags {
...
-    // Keep track of the args from the ArgIterator above that were allocated
-    // then free them all at the end of the scope.
+    // Keep track of the args from the ArgIterator above that were allocated.
+    // `args` will free them all at the end of the scope.
     var args_allocated = std.ArrayList([:0]const u8).init(allocator);
-    defer {
-        for (args_allocated.items) |arg| allocator.free(arg);
-        args_allocated.deinit();
-    }
+    defer args_allocated.deinit();
```
matklad, line 413: Similarly to cli.zig, I _think_ that just `args` can now play the role we used `args_allocated` for before.

scripts/install_zig.bat
```diff
@@ -1,6 +1,6 @@
 @echo off
 
-set ZIG_RELEASE_DEFAULT=0.9.1
+set ZIG_RELEASE_DEFAULT=0.10.1
```
matklad, line 3: :tada:

src/tigerbeetle/main.zig
```diff
@@ -156,11 +156,7 @@ const Command = struct {
             });
         }
 
-        const nonce = while (true) {
-            // `random.uintLessThan` does not work for u128 as of Zig 0.9.1.x.
-            const nonce_or_zero = std.crypto.random.int(u128);
-            if (nonce_or_zero != 0) break nonce_or_zero;
-        } else unreachable;
+        const nonce = std.crypto.random.int(u128) +| 1;
```
matklad, line 159: This is a great simplification, thanks!

scripts/install.sh
```diff
@@ -34,8 +34,8 @@ fi
 # shellcheck disable=SC2086
 if [ "$debug" = "true" ]; then
     echo "Building Tigerbeetle debug..."
-    ./scripts/build.sh install $cpu $target
+    zig/zig build install $cpu $target
```
matklad, line 37: :tada: :tada: :tada:

src/clients/docs_generate.zig
```diff
@@ -802,16 +802,14 @@ pub fn main() !void {
...
-    _ = args.next(global_arena.allocator());
-    while (args.next(global_arena.allocator())) |arg_or_err| {
-        const arg = arg_or_err catch {
-            std.debug.print("Could not parse all arguments.\n", .{});
-            return error.CouldNotParseArguments;
-        };
+    var args = try std.process.argsWithAllocator(global_arena.allocator());
+    defer args.deinit();
+
+    std.debug.assert(args.skip());
```
matklad: let's add `const assert = ` to keep this consistent with the rest of the code

scripts/lint_tigerstyle.zig
```diff
@@ -26,8 +26,11 @@ var general_purpose_allocator = std.heap.GeneralPurposeAllocator(.{}){};
 const gpa = general_purpose_allocator.allocator();
 
 pub fn main() !void {
-    const argv = std.os.argv;
-    for (argv[1..]) |raw_path| {
+    var args = try std.process.argsWithAllocator(gpa);
+    defer args.deinit();
+
+    std.debug.assert(args.skip());
```
matklad: same about `const assert = `

.gitattributes
```diff
@@ -0,0 +1 @@
+*.sh text=auto eol=lf
```
matklad, line 1: This shouldn't be necessary? We already have 

```
λ cat scripts/.gitattributes 
* text=auto eol=lf
*.bat text=auto eol=crlf
```

src/clients/go/tb_client.go
```diff
@@ -6,7 +6,7 @@ package tigerbeetle_go
 #cgo darwin,amd64 LDFLAGS: ${SRCDIR}/pkg/native/x86_64-macos/libtb_client.a -ldl -lm
 #cgo linux,arm64 LDFLAGS: ${SRCDIR}/pkg/native/aarch64-linux/libtb_client.a -ldl -lm
 #cgo linux,amd64 LDFLAGS: ${SRCDIR}/pkg/native/x86_64-linux/libtb_client.a -ldl -lm
-#cgo windows,amd64 LDFLAGS: -L${SRCDIR}/pkg/native/x86_64-windows -ltb_client -lws2_32
+#cgo windows,amd64 LDFLAGS: -L${SRCDIR}/pkg/native/x86_64-windows -ltb_client -lws2_32 -lssp
```
matklad: what's ssp?

src/clients/java/src/client.zig
```diff
@@ -51,19 +51,19 @@ const NativeClient = struct {
...
         max_concurrency: u32,
-    ) *Context {
+    ) ?*Context {
         const addresses = JNIHelper.get_string_utf(env, addresses_obj) orelse {
             ReflectionHelper.initialization_exception_throw(
                 env,
                 tb.tb_status_t.address_invalid,
             );
-            return undefined;
+            return null;
```
matklad, line 60: Hm, why Zig says that `return undefined` is illegal behavior? I tried simpler example (eg, https://github.com/ziglang/zig/issues/13178), and returning undefined there works. Is this some bug in compiler?

matklad: Although, given that we bitcast at the compile site, returning `null` seems better anyway.

matklad:
Got to the end! Looks good to me overall, most of comments are nits. Let me double check our usage of extern struct, pact struct, and whether we rely on alignment of u128, and it should be good to go.

src/vsr/superblock_client_sessions.zig
matklad: also, probably makes sense to add stdx.alignmentPreservingAsBytes and friends, but better to do it after this lands.

matklad (dismissed):
Ok, Looks Awesome To Me!

There are still a few nits left, but I'd be happy to merge this as is, and fix the nits in follow ups, to avoid merge conflits.

build.zig
jorangreef: > I would say that such inconsistency is a natural consequence of trying to always init/deinit

Agreed. On a related note (but not related to the Build.zig example), I do like your thinking on create/destroy etc. and we should use that where appropriate. I don't actually mean that we must always only use `init/deinit()`.

> So I'd say a more robust strategy is to avoid writing generic allocating code --- either allocate all memory at startup, or figure out how to use arenas to do bulk deallocation.

Agreed. However, there are parts of our code, which may end up in different contexts outside of TigerBeetle, which we can't anticipate now. For example, clients that we expose in different languages, where we may not control the allocator policy. Or embedded TigerBeetle mode... in a not-to-distant galaxy.

So what I meant is that wherever we allocate internally, we _must_ make a habit of defer'ing or errdefer'ing the correct deallocation code, without breaking encapsulation and assuming anything about the allocator. I have always thought of this as good hygiene, to build muscle around thinking about deallocations, and when reviewing, but the "don't lose flexibility" is the big one.

And again, being strict about deallocation also doesn't preclude these defense-in-depth strategies such as static allocation or arenas. We can do both, but I would like us to write the code assuming it needs to be correct regardless of allocator or context.

As far as possible, we should allow the user to set the allocator policy externally to the implementation, and prefer unmanaged interfaces. This preserves flexibility and minimizes contract.

At the same time, where it makes sense, we can be coupled to allocator strategy, for example to use the arena trick for resetting the storage engine.

> Like, it's not hard to get 9 out of 10 defers right, but getting them all is something we are not particularly good at.

“Do not go gentle into that good night!”

For example, this is why TigerStyle encourages a newline after allocation, so that the reviewer can more easily check that the corresponding deallocation is there.

I think we could also leverage Zig's test allocator more to get comfort here?

Finally, given that our outer policy is static allocation, we should also be able to detect leaks ourselves by measuring any memory increase through repeated resets?

src/state_machine.zig
kprotty: To treat it as a C union to save space iirc.

.gitattributes
kprotty: Sorry, certified windows moment for `src/clients/[go|node]/scripts/*.sh`

src/clients/go/tb_client.go
kprotty: Golang on windows is trying to use `-fstack-protector` in the CI, but libtb_client isn't built with that and it causes missing linker symbols. Can't seem to repro locally but this was one of the solutions I was trying out.

src/clients/java/src/client.zig
kprotty: Its a bug in the compiler. There's another bug with `[*c]const u8` not coercing to `[*:0]const u8` as before.

kprotty:
Nice! Let me wrap up some review comments for the small stuff.

matklad (dismissed):

matklad (approved):

eatonphil:
That [Go on Windows](https://github.com/tigerbeetle/tigerbeetle/actions/runs/5657829012/job/15327796695) failure looks real fwiw:

```
# github.com/tigerbeetle/tigerbeetle-go.test
C:\hostedtoolcache\windows\go\1.20.6\x64\pkg\tool\windows_amd64\link.exe: running D:\a\tigerbeetle\tigerbeetle\zig\zig.exe failed: exit status 1
lld-link: error: undefined symbol: __stack_chk_fail
>>> referenced by D:\a\tigerbeetle\tigerbeetle\src\clients\c\tb_client.zig:98
>>>               tb_client.lib(tb_client.lib.obj):(tb_client_init)
>>> referenced by D:\a\tigerbeetle\tigerbeetle\src\clients\c\tb_client.zig:133
>>>               tb_client.lib(tb_client.lib.obj):(tb_client_init_echo)
>>> referenced by D:\a\tigerbeetle\tigerbeetle\zig\lib\std\net.zig:89
>>>               tb_client.lib(tb_client.lib.obj):(net.Address.parseIp4)
>>> referenced 130 more times
```

PR #1055, opened 2023-07-22
tidy: allow long multiline strings

matklad:
For multiline strings, we want to check that the string itself is short-enough, but we don't mind if it is indented.

### DIFF
--- a/src/tidy.zig
+++ b/src/tidy.zig
@@ -71,7 +71,14 @@ fn find_long_line(file_text: []const u8) !?usize {
     var line_number: usize = 0;
     while (line_iterator.next()) |line| : (line_number += 1) {
         const line_length = try std.unicode.utf8CountCodepoints(line);
-        if (line_length > 100 and !is_url(line)) {
+        if (line_length > 100) {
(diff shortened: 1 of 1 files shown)


sentientwaffle:
Imo we should not limit multiline string length at all. Or set the limit high enough that none of the statemachine tests hit it (~180).

The purpose of limiting the line length is code readability. But splitting these long lines would only inhibit readability:

https://github.com/tigerbeetle/tigerbeetle/blob/19f97c36ea2a438c760abe2493c2d90682481592/src/state_machine.zig#L2048-L2088

matklad:
My motivation here was that a common use-case for such literals is for things which end up printed to the terminal, and there I think it’s pretty important to keep this under 100.

I do agree that state machine tests should use longer lines, I think I’d add a targeted exception there, once we move to newer zig

matklad:
(If you are ok with it, might land this to reduce annoyance for phil)

sentientwaffle (approved):
Yeah, the state machine tests were my only concern -- an exception there is fine.

PR #1061, opened 2023-07-25
Split out ACID FAQ into raw data considerations and financial data considerations

eatonphil:


docs/FAQ.md
```diff
@@ -149,27 +149,55 @@ and [give feedback](https://join.slack.com/t/tigerbeetle/shared_invite/zt-1gf3qn
...
-   can ultimately fail. Data we write today might not be available
-   tomorrow. TigerBeetle embraces limited disk reliability and
-   maximizes data durability in spite of imperfect disks. We actively
-   work against such entropy by taking advantage of cluster-wide
-   storage. A bit of data would need to get corrupted on all replicas
-   in a cluster to get lost, and even in that case the system would
-   safely halt.
+### Atomicity
+
+As part of replication, each operation is durably stored in a majority
```
sentientwaffle:
```suggestion
As part of replication, each operation is durably stored in a quorum
```

Due to flexible quorums, a quorum is not necessarily a majority.

docs/FAQ.md
```diff
@@ -149,27 +149,55 @@ and [give feedback](https://join.slack.com/t/tigerbeetle/shared_invite/zt-1gf3qn
...
+of replicas' Write-Ahead Log before indicating to the primary that the
+operation is successful. Only over time do WAL entries make their way
+through a state machine and into the LSM storage layer.
+
+The WAL is what allows TigerBeetle to achieve atomicity and durability
+since once the WAL is the source of truth. If TigerBeetle crashes, the
```
sentientwaffle: "since once the WAL is the source of truth"

Typo? I'm not sure what this is supposed to say.

docs/FAQ.md
```diff
@@ -149,27 +149,55 @@ and [give feedback](https://join.slack.com/t/tigerbeetle/shared_invite/zt-1gf3qn
...
+fault model, which we then verify and test with incredible levels of
+corruption, something which few distributed systems
+were ever designed to handle. Our emphasis on protecting Durability is
+what sets TigerBeetle apart, not only as a ledger but as a DBMS.
+
+However, absolute durability is impossible, because all hardware can
+ultimately fail. Data we write today might not be available
+tomorrow. TigerBeetle embraces limited disk reliability and maximizes
+data durability in spite of imperfect disks. We actively work against
+such entropy by taking advantage of cluster-wide storage. A bit of
```
sentientwaffle: Is "bit" being used colloquially here, or as a literal 1/8-of-a-byte? If the former I'd use a different word.

docs/FAQ.md
eatonphil: Fixed, thanks!

docs/FAQ.md
eatonphil: Oops yes, typo. Reworded this paragraph to be clearer (about what I meant anyway).

docs/FAQ.md
eatonphil: This was @matklad 's phrase. I also am not completely sure if it was meant as a literal bit or rather "a chunk".

docs/FAQ.md
```diff
@@ -149,27 +149,55 @@ and [give feedback](https://join.slack.com/t/tigerbeetle/shared_invite/zt-1gf3qn
...
-   work against such entropy by taking advantage of cluster-wide
-   storage. A bit of data would need to get corrupted on all replicas
-   in a cluster to get lost, and even in that case the system would
-   safely halt.
+### Atomicity
+
+As part of replication, each operation is durably stored in at least a
+quorum of replicas' Write-Ahead Log (WAL) before the primary will mark
+the operation as successful. Only over time do WAL entries make their
+way through a state machine and into the LSM storage layer.
```
jorangreef:
```suggestion
As part of replication, each operation is durably stored in at least a
quorum of replicas' Write-Ahead Logs (WAL) before the primary will acknowledge
the operation as committed. WAL entries are executed through the state machine business logic and the resulting state changes are stored in TigerBeetle's LSM-Forest local storage engine.
```

docs/FAQ.md
```diff
@@ -149,27 +149,55 @@ and [give feedback](https://join.slack.com/t/tigerbeetle/shared_invite/zt-1gf3qn
...
+
+The WAL is what allows TigerBeetle to achieve atomicity and durability
+since the WAL is the source of truth. If TigerBeetle crashes, the WAL
+is replayed on startup against the last checkpoint on disk.
```
jorangreef:
```suggestion
is replayed at startup from the last checkpoint on disk.
```

docs/FAQ.md
```diff
@@ -149,27 +149,55 @@ and [give feedback](https://join.slack.com/t/tigerbeetle/shared_invite/zt-1gf3qn
...
+
+Financial atomicity goes further than this though: events and
+transfers can be linked when created so they all succeed or fail
+together.
+
+### Consistency
+
+TigerBeetle guarantees strict serializability. And at the cluster
+level, stale reads are not possible since all operations (i.e. reads,
+not just writes) go through consensus.
```
jorangreef:
```suggestion
level, stale reads are not possible since all operations (not only writes, but also reads) go through the global consensus protocol.
```

docs/FAQ.md
```diff
@@ -149,27 +149,55 @@ and [give feedback](https://join.slack.com/t/tigerbeetle/shared_invite/zt-1gf3qn
...
+As part of replication, each operation is durably stored in at least a
+quorum of replicas' Write-Ahead Log (WAL) before the primary will mark
+the operation as successful. Only over time do WAL entries make their
+way through a state machine and into the LSM storage layer.
+
+The WAL is what allows TigerBeetle to achieve atomicity and durability
+since the WAL is the source of truth. If TigerBeetle crashes, the WAL
+is replayed on startup against the last checkpoint on disk.
+
+Financial atomicity goes further than this though: events and
```
jorangreef:
```suggestion
However, financial atomicity goes further than this: events and
```

To use the same structure as the next section.

docs/FAQ.md
```diff
@@ -149,27 +149,55 @@ and [give feedback](https://join.slack.com/t/tigerbeetle/shared_invite/zt-1gf3qn
...
+
+### Consistency
+
+TigerBeetle guarantees strict serializability. And at the cluster
+level, stale reads are not possible since all operations (i.e. reads,
+not just writes) go through consensus.
+
+However, financial consistency requires more than this. TigerBeetle
+exposes a double-entry accounting API where money is never created or
+destroyed, only transferred between accounts. And transfer history is
```
jorangreef:
```suggestion
exposes a double-entry accounting API to guarantee that money cannot be created or
destroyed, but only transferred from one account to another. And transfer history is
```

To follow Einstein's wording a little more closely with the law of conservation of energy:

> "Energy cannot be created or destroyed, it can only be changed from one form to another"

docs/FAQ.md
```diff
@@ -149,27 +149,55 @@ and [give feedback](https://join.slack.com/t/tigerbeetle/shared_invite/zt-1gf3qn
...
+immutable. You can read more about our consistency guarantees
+[here](./design/consistency.md).
+
+### Isolation
+
+All events within a batch (across all clients) are
+serialized at the state machine level. There are no concurrent
+events in TigerBeetle, even across clients.
```
jorangreef:
```suggestion
All client requests (and all events within a client request batch) are executed with the highest level of isolation, serially through the state machine, one after another, before the next operation begins. Counterintuitively, the use of batching and serial execution means that TigerBeetle can also provide this level of isolation optimally, without the cost of locks for all the individual events within a batch.
```

We also serialize at the global consensus protocol layer beneath.

docs/FAQ.md
```diff
@@ -149,27 +149,55 @@ and [give feedback](https://join.slack.com/t/tigerbeetle/shared_invite/zt-1gf3qn
...
+
+### Durability
+
+Up until 2018, traditional DBMS durability has focused on the crash
+consistency model, however, Fsyncgate and [Protocol Aware
```
jorangreef:
```suggestion
Up until 2018, traditional DBMS durability has focused on the Crash
Consistency Model, however, Fsyncgate and [Protocol Aware
```

docs/FAQ.md
```diff
@@ -149,27 +149,55 @@ and [give feedback](https://join.slack.com/t/tigerbeetle/shared_invite/zt-1gf3qn
...
+Recovery](https://www.usenix.org/conference/fast18/presentation/alagappan)
+have shown that this model can lead to real data loss for users in the
+wild. TigerBeetle therefore adopts an explicit storage
+fault model, which we then verify and test with incredible levels of
+corruption, something which few distributed systems
+were ever designed to handle. Our emphasis on protecting Durability is
```
jorangreef:
```suggestion
corruption, something which few distributed systems historically
were designed to handle. Our emphasis on protecting Durability is
```

jorangreef (dismissed):
LATM!

docs/FAQ.md
matklad: Colloquial sense, but I now see that this is ambiguous. Maybe just call it a "record"? 

> A record can only be lost if it gets corrupted on all replica, and even in that case the system would safely halt.

eatonphil: Switched to record. Thanks!

sentientwaffle (approved):

PR #1062, opened 2023-07-26
stdx: move hash testing vectors out of the way

matklad:
Two goals here:

- reduce clutter in `stdx` --- we append to that file fairly often, so kepeing it readable is nice
- enforce linelength check for stdx


kprotty:
Would it make sense to move stdx.hash_inline into its own file like vsr.checksum?

matklad:
Not sure! For vsr.checksum, this is a very specific, protocol defining hash.

If I understand the semantics of `inline_hash` correctly, it is just a general-purpose hash function, which we use for in-memory hash maps (but never leak to disk or the network). As such, `stdx` seems like the right place for it. 

At some point, we _might_ want to split stdx further, to have, eg `stdx.mem`, `stdx.meta`, `stdx.hash`, but I don't think we are at that point yet.

kprotty (approved):

PR #1064, opened 2023-07-26
style: document one more divergence from Zig style

matklad:
Specifically, that we `XMLHTTPRequest` rather than `XmlHttpRequest`. Also, rewrap at 100 while we are at it.

### DIFF
--- a/docs/TIGER_STYLE.md
+++ b/docs/TIGER_STYLE.md
@@ -2,196 +2,367 @@
 
 ## The essence of style
 
-> “There are three things extremely hard: steel, a diamond, and to know one's self.” — Benjamin Franklin
+> “There are three things extremely hard: steel, a diamond, and to know one's self.”
+> — Benjamin Franklin
 
-TigerBeetle's coding style is evolving. A collective give-and-take at the intersection of engineering and art. Numbers and human intuition. Reason and experience. First principles and knowledge. Precision and poetry. Just like music. A tight beat. A rare groove. Words that rhyme and rhymes that break. Biodigital jazz. This is what we've learned along the way. The best is yet to come.
+TigerBeetle's coding style is evolving. A collective give-and-take at the intersection of
+engineering and art. Numbers and human intuition. Reason and experience. First principles and
+knowledge. Precision and poetry. Just like music. A tight beat. A rare groove. Words that rhyme and
+rhymes that break. Biodigital jazz. This is what we've learned along the way. The best is yet to
+come.
 
 ## Why have style?
 
 Another word for style is design.
 
-> “The design is not just what it looks like and feels like. The design is how it works.” — Steve Jobs
+> “The design is not just what it looks like and feels like. The design is how it works.”
+> — Steve Jobs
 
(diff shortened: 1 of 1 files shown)


docs/TIGER_STYLE.md
```diff
@@ -2,196 +2,368 @@
...
+  significantly affect performance, and avoids use-after-free. As a second-order effect, it is our
+  experience that this also makes for more efficient, simpler designs that are more performant and
+  easier to maintain and reason about, compared to designs that do not consider all possible memory
+  usage patterns upfront as part of the design.
+
+* Declare variables at the **smallest possible scope**, and **minimize the number of variables in
+  scope**, to reduce the probability that variables are misused.
+
+* Restrict the length of function bodies to reduce the probability of poorly structured code. We
+  enforce a **hard limit of 70 lines per function**.
```
matklad, line 141: Reading this... I bet that's not actually the case for our code base at the moment...

sentientwaffle: Maybe we should remove that bullet? Fwiw I often [prefer to inline](http://number-none.com/blow/john_carmack_on_inlined_code.html) when there is a single caller.

jorangreef: We should try to get back to it, as a place to aim, allowing some exceptions that make sense for inlining, but not remove it.

I found when working on VSR that whenever I would tackle these, the code would get better. It can help to reduce the probability of variables being declared and then used only further along, and incorrectly.

jorangreef: This one originally comes from NASA's The Power of Ten Rules. The reasoning in the PDF on this is gold. The idea is that functions should be able to be printed on a page, or fit on a monitor.

jorangreef: They originally had 60 lines and we bumped it to 70... screens are a bit bigger these days.

matklad: I am personally torn here. That "on inline code" is in my "top articles of all time" and that's also how I prefer to write code by default. _But_ if we are really striving for perfect code, and also use `fn foo()`, `fn foo_bar()`, `fn foo_baz()` naming, then we likely actually could improve readability by introducing new functions.

docs/TIGER_STYLE.md
```diff
@@ -2,196 +2,368 @@
...
-
-* When choosing related names, try hard to find names with the same number of characters so that related variables all line up in the source. For example, as arguments to a memcpy function, `source` and `target` are better than `src` and `dest` because they have the second-order effect that any related variables such as `source_offset` and `target_offset` will all line up in calculations and slices. This makes the code symmetrical, with clean blocks that are easier for the eye to parse and for the reader to check.
-
-* When a single function calls out to a helper function or callback, prefix the name of the helper function with the name of the calling function to show the call history. For example, `read_sector()` and `read_sector_callback()`.
+* Do not abbreviate variable names, unless the variable is a primitive integer type used as an
+  argument to a sort function or matrix calculation. Use proper capitalization for acronyms
+  (`VSRState`, not `VsrState`).
```
jorangreef, line 222: `VSRState`—sigh of relief! 😅

jorangreef (approved):

PR #1066, opened 2023-07-26
vsr: rename nodes_max -> members_max

matklad:
cf vsr.Members


sentientwaffle (approved):
Good catch! "members" is much less overloaded than "nodes". :+1:

PR #1067, opened 2023-07-26
zig: fix cli args memory management

matklad:


src/benchmark.zig
```diff
@@ -178,11 +178,12 @@ fn parse_arg_addresses(
...
     if (!std.mem.eql(u8, arg, arg_name)) return false;
 
-    allocator.free(arg_value.*);
-
     const address_string = args.next() orelse
         panic("Expected an argument to {s}", .{arg_name});
-    arg_value.* = try vsr.parse_addresses(allocator, address_string, constants.nodes_max);
+
+    const addresses = try vsr.parse_addresses(allocator, address_string, constants.nodes_max);
+    allocator.free(arg_value.*);
```
matklad, line 185: Mostly a drive-by-fix, but I belive we have a double-free here if the `try` actually errors.

src/tigerbeetle/cli.zig
```diff
@@ -106,26 +106,27 @@ pub const Command = union(enum) {
...
-            .format => |cmd| cmd.args_allocated,
-            .start => |cmd| cmd.args_allocated,
-            .version => return,
-        };
-
-        for (args_allocated.items) |arg| allocator.free(arg);
-        args_allocated.deinit();
+    pub fn deinit(command: *Command, allocator: std.mem.Allocator) void {
+        switch (command.*) {
+            .start => |*start| allocator.free(start.addresses),
```
matklad, line 123: Seems like we were leaking addresses before

src/vopr.zig
```diff
@@ -407,17 +407,10 @@ fn parse_args(allocator: mem.Allocator) !Flags {
     var args = try std.process.argsWithAllocator(allocator);
     defer args.deinit();
 
-    // Keep track of the args from the ArgIterator above that were allocated.
-    // `args` will free them all at the end of the scope.
-    var args_allocated = std.ArrayList([:0]const u8).init(allocator);
```
matklad, line 412: Just dead code?

src/tigerbeetle/cli.zig
```diff
@@ -144,21 +145,16 @@ pub fn parse_args(allocator: std.mem.Allocator) !Command {
...
-    var args_allocated = std.ArrayList([:0]const u8).init(allocator);
+    errdefer args.deinit(allocator);
 
     // Skip argv[0] which is the name of this executable
     const did_skip = args.skip();
     assert(did_skip);
 
     const raw_command = args.next() orelse
         fatal("no command provided, expected 'start', 'format', or 'version'", .{});
-    defer allocator.free(raw_command);
```
matklad, line 161: I think this slipped during upgrade, and its probably UB: the `raw_command` might be libc argv, which shouldn't be freed.

src/tigerbeetle/cli.zig
```diff
@@ -106,26 +106,27 @@ pub const Command = union(enum) {
...
-
-        for (args_allocated.items) |arg| allocator.free(arg);
-        args_allocated.deinit();
+    pub fn deinit(command: *Command, allocator: std.mem.Allocator) void {
+        switch (command.*) {
+            .start => |*start| allocator.free(start.addresses),
+            else => {},
+        }
+        switch (command.*) {
+            inline else => |*cmd| cmd.args_allocated.deinit(),
```
kprotty, line 127: just 0.10 things ⚡

src/vopr.zig
kprotty: Seems like it. May have previously been used to return.

kprotty (approved):

src/tigerbeetle/cli.zig
matklad: yup-yup!

PR #1068, opened 2023-07-26
Make quickstart higher

eatonphil:
Also drops the consensus challenge line for now since that hasn't picked back up yet. Everything else has just been moved to after the quickstart.

image


jorangreef (approved):
LATM!

PR #1071, opened 2023-07-26
Simplifies the pattern for holding multiple contexts inside a union.

batiati:
- Adds the function `get()`, so it's not necessary to initialize the tag.

- Moves `union_field_parent_ptr` to `stdx`, making it more generic and comptime checking if we can use a simple cast instead of doing pointer arithmetic.

- Adds unit tests.

### DIFF
--- a/src/stdx.zig
+++ b/src/stdx.zig
@@ -432,40 +432,25 @@ pub fn update(base: anytype, diff: anytype) @TypeOf(base) {
 // TODO(Zig): No need for this function once Zig is upgraded
 // and @fieldParentPtr() can be used for unions.
 // See: https://github.com/ziglang/zig/issues/6611.
-
-pub fn UnionFieldParentPtrType(
-    comptime Union: type,
-    comptime field: std.meta.FieldEnum(Union),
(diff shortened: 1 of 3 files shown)


src/stdx.zig
```diff
@@ -428,3 +428,65 @@ pub fn update(base: anytype, diff: anytype) @TypeOf(base) {
...
+    const Field = std.meta.fieldInfo(Union, field).field_type;
+
+    assert(std.meta.trait.isSingleItemPtr(Child));
+    assert(std.meta.Child(Child) == Field);
+
+    return if (std.meta.trait.isConstPtr(Child))
+        *const Union
+    else
+        *Union;
+}
```
kprotty: Could this be omitted in favor of assuming `*Union` for `union_field_parent_ptr` usages?

batiati: Yes, we can do that, the previous implementation removed the *const qualifier by doing intToPtr, so I wanted to fix it and keep the stdx implementation generic for future use. But anyway, it’s not our use case.

kprotty (approved):

PR #1073, opened 2023-07-27
RFC: introduce argument parsing library

matklad:


src/copyhound.zig
```diff
@@ -31,6 +31,7 @@
 
 const std = @import("std");
 const stdx = @import("./stdx.zig");
+const flags = @import("./flags.zig");
```
jorangreef, line 34: Love the name.

src/flags.zig
```diff
@@ -0,0 +1,127 @@
...
+        while (std.mem.indexOf(u8, field[index..], "_")) |i| {
+            result = result ++ field[index..][0..i] ++ "-";
+            index = index + i + 1;
+        }
+        result = result ++ field[index..];
+    }
+    return result;
+}
+
+fn FlagsCounts(comptime Flags: type) type {
```
jorangreef:
```suggestion
fn FlagsCountsType(comptime Flags: type) type {
```

We've started doing this pattern in the forest. It would be good to add to TigerStyle too:

Suffix type constructor names with `Type`. This enables the resulting type to be given the name that the constructor might otherwise have shadowed.

src/copyhound.zig
```diff
@@ -41,53 +42,6 @@ const size_thershold = 1024;
 const CliArgs = union(enum) {
```
sentientwaffle, line 40: s/size_thershold/size_threshold (a couple lines up)

src/flags.zig
```diff
@@ -0,0 +1,127 @@
+const std = @import("std");
+const assert = std.debug.assert;
+
+/// Format and print an error message followed by the usage string to stderr,
+/// then exit with an exit code of 1.
```
sentientwaffle: > followed by the usage string

Is the comment wrong or the code? I don't see a usage string.

src/flags.zig
```diff
@@ -0,0 +1,127 @@
...
+
+    assert(@typeInfo(Flags) == .Struct);
+
+    var result: Flags = undefined;
+    var counts: FlagsCounts(Flags) = .{};
+
+    next_arg: while (args.next()) |arg| {
+        inline for (comptime std.meta.fields(Flags)) |field| {
+            const flag = comptime field_to_flag(field.name);
+            if (std.mem.startsWith(u8, arg, flag)) {
```
sentientwaffle: Maybe additionally check that the next character after the flag is either nothing, `=`, or a space?

Right now:

```
$ zig run  src/copyhound.zig -- memcpy --bytes_except_this_is_a_different_flag
error: --bytes argument requires a value
```

which is kind of weird, and I think would be broken if one flag was a prefix of another.

src/flags.zig
```diff
@@ -0,0 +1,127 @@
...
+        if (args.next()) |arg| {
+            fatal("unexpected argument: '{s}'", .{arg});
+        }
+        return {};
+    }
+
+    assert(@typeInfo(Flags) == .Struct);
+
+    var result: Flags = undefined;
+    var counts: FlagsCounts(Flags) = .{};
```
sentientwaffle: Or maybe

```suggestion
    var counts: std.enums.EnumFieldStruct(std.meta.FieldEnum(Flags), u32, 0) = .{};
```

src/flags.zig
sentientwaffle: Actually I think `tigerbeetle/cli.zig` has this exact problem too! 

```zig
        } else if (mem.startsWith(u8, arg, "--cache-transfers")) {
            if (command != .start) fatal("--cache-transfers: supported only by 'start' command", .{});
            cache_transfers = parse_flag("--cache-transfers", arg);
        } else if (mem.startsWith(u8, arg, "--cache-transfers-posted")) {
            if (command != .start) fatal("--cache-transfers-posted: supported only by 'start' command", .{});
            cache_transfers_posted = parse_flag("--cache-transfers-posted", arg);
```

src/flags.zig
matklad: :boom:  :exploding_head: :exploding_head: :exploding_head: :boom:

src/flags.zig
matklad: >which is kind of weird, and I think would be broken if one flag was a prefix of another.

Yeah, this is intentional bug-for-bug compatibility with our `cli.zig` for the moment. 

I do want to say "let's do the right thing when parsing CLI arguments", but at this point I am not quite sure what that right thing is

src/flags.zig
matklad: wrong both here and in the cli.zig which is where this was copyied from.

src/flags.zig
matklad: Ah, well, this actually hits the benchmarks, which have `--transfer-count` and `--transfer-count-per-second`. I was thinking about just banning prefix args, but seems better to handle this properly.

src/flags.zig
```diff
@@ -0,0 +1,201 @@
...
+//!      Don't use `--key value`, as that can be ambiguous (e.g., `--key --verbose`).
+//!    - Use subcommand syntax when appropriate.
+//!    - Use positional arguments when appropriate.
+//!
+//! Design choices for this particular `flags` library:
+//!
+//! - Be a 80% solution. Parsing arguments is a surprisingly vast topic: auto-generated help,
+//!  bash completions, typo correction. Rather than providing a definitive solution, `flags`
+//!  is just one possible option. It is ok to re-implement arg parsing in a different way, as long
+//!  as the CLI guidelines are observed.
```
sentientwaffle:
```suggestion
//!   bash completions, typo correction. Rather than providing a definitive solution, `flags`
//!   is just one possible option. It is ok to re-implement arg parsing in a different way, as long
//!   as the CLI guidelines are observed.
```

(To match the other list item's indentation.)

src/flags.zig
```diff
@@ -0,0 +1,201 @@
...
+        const T = field.field_type;
+        if (T == bool) {
+            assert(default_value(field) == false); // boolean flags should have explicit default
+        } else {
+            assert(T == []const u8 or @typeInfo(T) == .Int); // unsupported CLI argument type
+        }
+    };
+
+    var result: Flags = undefined;
+    var counts: std.enums.EnumFieldStruct(Flags, u32, 0) = .{};
```
sentientwaffle, line 107: Even better! I didn't realize you could just pass a struct like this directly, since it is using `meta.fields()` instead of `enums.values()`! :100:

src/flags.zig
```diff
@@ -0,0 +1,201 @@
...
+
+fn default_value(comptime field: std.builtin.Type.StructField) ?field.field_type {
+    return if (field.default_value) |default|
+        @ptrCast(
+            *const field.field_type,
+            @alignCast(@alignOf(field.field_type), default),
+        ).*
+    else
+        null;
+}
```
sentientwaffle: If this effectively "cast field.default_value to field.field_type"?

matklad: Yup! Default is *anyopaque, so we need to cast both the alignment and he actual type, will add a comment

sentientwaffle (approved):

PR #1054, opened 2023-07-21
Scan API and get_account_transfers

batiati:
## This PR

Scan API, including fuzz tests and the new state machine operation `get_account_transfers` for querying transfers for a given account.

<details>
<summary>Scan API</summary>

### Scan API:

`ScanTreeType`: The basic building block of the range queries, specialized over `Tree`, allowing to seek for `key_min`..`key_max` in ascending or descending key order in **any** LSM tree.

`ScanBuilderType`: Specialized type over `Groove`, grouping `ScanTreeType`s for all `IndexTrees` in the groove, exposing them in a common API not specialized by `Tree` (the `Scan` Interface), allowing conditions where the predicate is the tree's `CompositeKey` prefix and the value produced is the `ObjectTree`'s timestamp.

Merge operations (union, intersection, difference) can be performed with scans that are agnostic about its source (e.g. it can be a `ScanTreeType` specialized over any `Tree` or it can be another merge operation). The `Scan` type holds all possible concrete implementations as a tagged Zig `union` for dynamic dispatch through comptime-generated code.
> Note: Only merge union is currently implemented.

`ScanLookupType`: Implements the lookup logic for retrieving the groove objects from a scan. Unlike the regular prefetch mechanism, this lookup acts only over `timestamp` and loads the objects directly into a provided buffer, keeping the same sort order of the underlying scan. The buffer size can be used as the `LIMIT` operator.

This diagram illustrates how multiple scans can be combined to create a complex query:
> WARNING: outdated diagram.

image

## API limitations
This API is intended to be used internally by the state machine, not to be exposed directly to the end user, it has some sensible limitations:
> Note: All SQL examples are only illustrative, it's not the actual query language provided by the Scan API.

#### 1. Only secondary indexes contained in `IndexTree` can be queried by `ScanGrooveType`.

Examples:

✔ Supported:
```SQL
WHERE code=$1 AND ledger=$2
```

✖ Not supported:
```SQL
WHERE id=$1 AND code=$2
```
>Rationale: `id` isn't an `IndexTrees` field, this kind of query may be split into two steps:
1 - Performing a regular lookup for `id=$1`
2 - And then check if `code=$2` in the instance found in the lookup step.

#### 2. The field `timestamp` can be used in the query for paging, not as an index.

Examples:

✔ Supported:
```SQL
WHERE code=$1 AND timestamp>$2
```

✖ Not supported:
```SQL
WHERE timestamp>$2
```
[...]

PR #1076, opened 2023-07-28
[RFC] Automatically voiding pending transfers based on timeout.

batiati:
## This RFC:

This RFC is based on [WIP] Scan API #1054. The focus is to discuss the design since the implementation will likely change as the Scan API gets reviewed.

### State machine:

- During the operation's prefetch, search for all pending transfers such `timestamp + timeout <= prepare_timestamp` and prefetch the debit and credit accounts.

- Before the operation is executed, iterate over all expired transfers and (a) Reverse the pending credit/debit for the mutable account involved; (b) Insert the `Posted` object with `.fulfillment = .voided`; and (c) Remove the `expires_at` index entry so that this transfer won't be returned again in the future.

- Unlike the manual process where the user inserts a new transfer for voiding the pending one, in the automatic process no `void` transfer is inserted, only the balance and the `Posted` groove are updated. 

### Changes:

- New derived index `Transfers.expires_at`, which is inserted only for pending transfers with `timeout>0` and calculated based on `timestamp + timeout`. This index allows us to search all transfers about to expire in a given timestamp.

- `TransferPendingExpiry` buffer to hold the maximum number of pending transfers we can process in a single operation.
This point is very sensible for the design since we are expiring transfers passively (triggered by operations). The query can return unlimited pending transfers to process at a given timestamp.

### Testing:

- Benchmark gets 30% of pending transfers chance, to stress the expiration process.

- Demo unit test implemented for the Java client: creating a pending transfer, waiting for it to expire, and looking up again.


jorangreef:
Great @batiati!

What guarantees will we give re: _exactly when_ a pending transfer is timed out? Will we timeout everything before beginning to execute the batch? What if an expiry timestamp intersects the middle of the batch?

batiati:
> Great @batiati!
> 
> What guarantees will we give re: _exactly when_ a pending transfer is timed out? Will we timeout everything before beginning to execute the batch? What if an expiry timestamp intersects the middle of the batch?

I pondered that, but as long we are not recording a new transfer to void the pending one (which would be the evidence of _when_ we did it), it is virtually free of temporality. We can expire all transfers before the first commit, or one by one at the exact timestamp, and it won't change the result that is visible to the user at the end.

"If a tree falls in a forest and no one is around to hear it, does it make a sound?" 😁

sentientwaffle:
> I pondered that, but as long we are not recording a new transfer to void the pending one (which would be the evidence of when we did it), it is virtually free of temporality. We can expire all transfers before the first commit, or one by one at the exact timestamp, and it won't change the result that is visible to the user at the end.

I think this _does_ impact what is visible to the user. Suppose I create a batch of transfers, and they are assigned timestamps 10,11,12,...,20.

If there was already a pending transfer due to time out at T=15, then depending on whether the expiring transfer is voided before or after these transfers, the balance in the expiring transfers will be different. Any of these new incoming transfers may reuse the same account(s) as the pending transfer, so differences in the balance are user-visible.

sentientwaffle:
StateMachine's timestamps and timeouts are both nanoseconds.
If timeouts were set with millisecond precision instead (leaving timestamps as nanoseconds), then we could guarantee that timeouts never land in the middle of a batch. 

That is, between every millisecond, there are 1e6 nanoseconds, which is more than 8191 (transfers/batch). So if a timeout would land in the middle of a batch, expire the timeout + bump the statemachine timestamp before executing the batch.

batiati:
> I think this _does_ impact what is visible to the user.

@sentientwaffle , you are absolutely right, I explained the thing very poorly 😬.
I was referring to the moment at which the transfer is voided not being visible to the end user, but the way transfers are processed in this RFC is **flawed** regarding the `*_pending` components of the balance, whether their `expires_at` timestamp intersects with the batch or not:

> `TransferPendingExpiry` buffer to hold the maximum number of pending transfers we can process in a single operation.
> This point is very sensible for the design since we are expiring transfers passively (triggered by operations). The query can return unlimited pending transfers to process at a given timestamp.

Considering the following example (exaggerated for illustration purposes, but it might happen at any time scale) :

- Suppose the user sets a fixed timeout of 24 hours for each pending transfer and then submits 100k pending transfers during the entire day.

- Now, let's say the day after is a holiday, so there's no activity in the cluster.

- At Day+2's first operation, it will find 100k transfers to expire, but the state machine can only deal with a limited number of them during each operation, ignoring the rest.

I was thinking of solving this issue by processing only the pending transfers for accounts involved in the current batch (we could even check if `credits_pending > 0 or debits_pending > 0` before querying). In this case, if an account had an expired transfer, it would remain pending indefinitely until some operation touched it (eg a new transfer or an account lookup).
However, it adds a lot of complexity to prefetch logic, and I feel like I'm just moving the problem around because we can still find an arbitrarily large number of pending transfers per account basis.

-----------

>If timeouts were set with millisecond precision instead (leaving timestamps as nanoseconds), then we could guarantee that timeouts never land in the middle of a batch.
>That is, between every millisecond, there are 1e6 nanoseconds, which is more than 8191 (transfers/batch). So if a timeout would land in the middle of a batch, expire the timeout + bump the statemachine timestamp before executing the batch.

The `timeout` being expressed as milliseconds makes much more sense to the business logic :+1: 
Also, tricking the state machine timestamp is an elegant solution for the intersection problem, ~although I think it would work orthogonally to the timeout being `ms`.~

sentientwaffle:
Ah, I see; thank you for the explanation! That does indeed make sense.

> but the state machine can only deal with a limited number of them during each operation, ignoring the rest.

Gotcha, so since the `TransferPendingExpiry` buffer may not fit the (unbounded) number of transfers that can possible expire at a single millisecond timestamp, we may need to support multiple fetch/expire rounds per commit.

I think that's not actually that bad, though:

- In the common case, the buffer should be large enough that no more than one round is required. (i.e. the cluster is not idle for long periods, and timeouts don't often land together on the same exact (millisecond) timestamp).
- That buffer could be sized with a runtime argument (e.g. from the CLI), so if a user's workload (somehow) runs into the buffer's limit often, they could run their cluster with a higher limit. The buffer's size has no impact on the state machine's determinism.

batiati:
> I think that's not actually that bad

The only issue is that we will need to compact/checkpoint in between rounds. Otherwise, the mutable table will get full (specifically the tables from `PostedGroove` and the index `expire_at`).

batiati:
Implemented by #1657

PR #1080, opened 2023-07-28
vsr: extra test for checksum

matklad:
vsr.checksum is pretty important, as it cements our on-disk representation and over-the-wire protocol (so, we won't be able to easily change it).

This PR adds a "change detector" test to ensure we don't accidentally change the result. The idea of the test is basically to hash a bunch of random stuff, and then assert that the final hash has a specific value.

While we are at it, extend our fuzz test to do more sanity checking:

- allow arbitrary bitflips (rather than just +%= 1)
- also test avalanching (that every bit of input affects every bit of the output)

While doing all that, I've noticed that the results we get are actually different from the results in Zig master. That's because we treat bytes-to-hash as a secret message, rather than as associated data. Adjust our algo to do that. I've also committed&reverted test against Zig master, which passes the change-detector test with updated expected value.

### DIFF
--- a/src/vsr/checksum.zig
+++ b/src/vsr/checksum.zig
@@ -1,86 +1,48 @@
+//! This file implements vsr.checksum. TigerBeetle uses this checksum to:
+//!
+//! - detect bitrot in data on disk,
+//! - validate network messages before casting raw bytes to an `extern struct` type,
+//! - hash-chain prepares and client requests to have strong consistency and ordering guarantees.
+//!
+//! As this checksum is stored on disk, it is set in stone and impossible to change.
+//!
+//! We need this checksum to be fast (it's in all our hotpaths) and strong (it's our ultimate line
+//! of defense against storage failures and some classes of software bugs).
+//!
+//! Our checksum of choice is based on Aegis:
+//!
+//! <https://datatracker.ietf.org/doc/draft-irtf-cfrg-aegis-aead/>
+//!
+//! We use the implementation from the Zig standard library, but here's the overall overview of the
+//! thing works:
+//!
+//! - AES-block is a widely supported in hardware symmetric encryption primitive (`vaesenc`,
+//!   `vaesdec` instructions). Hardware acceleration is what provides speed.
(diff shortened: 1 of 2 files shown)


matklad:
Should be ready for review @kprotty! Dotnet tests fail, but that seems to be pre-existing problem on main which is being fixed independently.

src/vsr/checksum.zig
```diff
@@ -135,28 +138,85 @@ test "Aegis simple fuzzing" {
...
+        prng.fill(message);
+
+        cases[case_index] = checksum(message);
+        case_index += 1;
+    }
+
+    // Sanity check that we are not getting trivial answers.
+    for (cases) |case_a, i| {
+        assert(case_a != 0);
+        assert(case_a != std.math.maxInt(u128));
```
kprotty, line 158: The zero case makes sense, but never considered if the full bits pattern was unlikely too.

kprotty (dismissed):

matklad:
Remixed this to use the upstream impl!

kprotty:
Looks good. Good enough that it needs to be off `tidy.zig`'s naughty list for the test/CI.

matklad:
Fixed!

matklad:
wait, no, test legitimately fail...

matklad:
Ok, now it seems to actually be green, ptal @kprotty

src/vsr/checksum.zig
```diff
@@ -1,132 +1,89 @@
...
+//!   And that's what we do here!
+
 const std = @import("std");
 const builtin = @import("builtin");
 const mem = std.mem;
 const testing = std.testing;
+const assert = std.debug.assert;
 
-// TODO(king): Replace with std.crypto.aead.aegis.Aegis128LMac_128 once Zig is updated.
-const Aegis128 = struct {
```
kprotty, line 7: Side note: wondering if we'll need to bring this back for checksum() to work at comptime (thinking `@inComptime()` switch for `aes.Block`)

kprotty (approved):

PR #1081, opened 2023-07-29
cli: move main argument parser to flags.zig

matklad:
Got carried away a bit here :rofl:


src/flags.zig
```diff
@@ -206,21 +281,108 @@ fn parse_flag_value_int(comptime T: type, comptime flag: []const u8, value: [:0]
...
+        .{ 100 * gib, "100GB" },
+        .{ 1000 * gib, "1000gb" },
+        .{ 1 * mib, "1MiB" },
+        .{ 10 * mib, "10mib" },
+        .{ 100 * mib, "100MB" },
+        .{ 1000 * mib, "1000mb" },
+        .{ 1 * kib, "1KiB" },
+        .{ 10 * kib, "10kib" },
+        .{ 100 * kib, "100KB" },
+        .{ 1000 * kib, "1000kb" },
```
matklad, line 358: NB: I intentionally removed support for spaces in the arguments. `--size='6 Gib'` feels very error-prone, if it is printed somewhere and then copy-pasted.

src/tigerbeetle/cli.zig
```diff
@@ -18,80 +19,110 @@ const StateMachine = vsr.state_machine.StateMachineType(
...
+        \\
+        \\  tigerbeetle version --verbose
+        \\
+    , .{
+        .default_address = constants.address,
+        .default_port = constants.port,
+    });
+};
 
 pub const Command = union(enum) {
```
matklad, line 122: Note that we have both `CliArgs` with "syntactically valid" arguments, and `Command` with semantically-validated ones. I actually like this overall pattern when doing any kind of parsing/deserialization --- rather than trying to cram all the validation into auto-generatied code, do only shallow syntax validation when parsing from a string to a type, and then have a second layer of manually written dumb validation which transforms between two types (but doesn't have to do parsing itself).

src/tigerbeetle/cli.zig
```diff
@@ -133,330 +164,152 @@ pub const Command = union(enum) {
...
-    var verbose: ?bool = null;
-
     // This iterator owns the arguments and is passed to the caller as a part of `Command`.
     var args = try std.process.argsWithAllocator(allocator);
     errdefer args.deinit(allocator);
 
     // Skip argv[0] which is the name of this executable
-    const did_skip = args.skip();
-    assert(did_skip);
+    assert(args.skip());
```
matklad, line 167: I suggest `assert(args.skip())` as the canonicall way to skip the argument, given that we require ReleaseSafe (that is, I am ok with simple side-effects in assertions).

src/tigerbeetle/cli.zig
```diff
@@ -133,330 +164,152 @@ pub const Command = union(enum) {
...
                 .version = .{
                     .args_allocated = args,
-                    .verbose = verbose orelse false,
+                    .verbose = version.verbose,
                 },
             };
         },
-        .format => {
-            const replica_count_parsed = parse_replica_count(
-                replica_count orelse fatal("required: --replica-count", .{}),
```
matklad, line 221: I tried to make all error message consistent, and to pub the problematic flag name first.

src/flags.zig
```diff
@@ -233,3 +396,409 @@ fn default_value(comptime field: std.builtin.Type.StructField) ?field.field_type
...
+    var t = try T.init(std.testing.allocator);
+    defer t.deinit();
+
+    // Test-cases are roughly in the source order of the corresponding features.
+
+    try t.check(&.{"empty"}, snap(@src(),
+        \\stdout:
+        \\empty
+        \\
+    ));
```
matklad, line 590: This actually is a great motivational example for snapshot testing! I can imaging us wanting to tweak various error messages over time, and now its trivial to update this. 

I also thought about using `shell.zig` to compile&run the test executable, to create a singularity out of our three micro-libraries, but decided not to. The "run" bit is not a usual shell-scripting, because there failure is very much expected, and we want to get everything there is about process exectuion, which `std.ChildProcess.exec` gives you. `shell.zig` would've work for compiling bit, but decided to be locally consistent instead.

matklad:
NB: our java integration tests caught an unintentional change in the argument names here, so this needs a somewhat careful review.

src/flags.zig
```diff
@@ -54,22 +52,43 @@ pub fn fatal(comptime fmt_string: []const u8, args: anytype) noreturn {
...
+///    format: struct {
+///        verbose: bool = false,
+///        positional: struct {
+///            path: []const u8,
+///        }
+///    },
+///
+///    pub const help =
+///        \\ tigerbeetle start --addresses=<addresses> --replica=<replica>
+///        \\ tigerbeetle format [--verbose]
```
sentientwaffle:
```suggestion
///        \\ tigerbeetle format [--verbose] <path>
```

src/flags.zig
```diff
@@ -91,15 +110,27 @@ pub fn parse_flags(args: *std.process.ArgIterator, comptime Flags: type) Flags {
 
     assert(@typeInfo(Flags) == .Struct);
 
-    const fields = comptime std.meta.fields(Flags);
-    comptime assert(fields.len >= 1);
+    comptime var fields: [16]std.builtin.Type.StructField = undefined;
```
sentientwaffle: Since this is all comptime, instead of defining a capacity `[16]`/`field_count`, you could do:

```zig
const StructField = std.builtin.Type.StructField;
comptime var fields: []const StructField = &[_]StructField{};
...
fields = fields ++ &[_]StructField{field};
```

src/flags.zig
```diff
@@ -91,15 +110,27 @@ pub fn parse_flags(args: *std.process.ArgIterator, comptime Flags: type) Flags {
...
+
+    comptime var positional_fields: []const std.builtin.Type.StructField = &.{};
 
-    comptime for (fields) |field| {
-        const T = field.field_type;
-        if (T == bool) {
-            assert(default_value(field) == false); // boolean flags should have explicit default
+    comptime for (std.meta.fields(Flags)) |field| {
+        if (std.mem.eql(u8, field.name, "positional")) {
+            assert(@typeInfo(field.field_type) == .Struct);
```
sentientwaffle, line 120:
```suggestion
            assert(@typeInfo(field.field_type) == .Struct);
            assert(positional_fields.len == 0);
```

src/flags.zig
```diff
@@ -233,3 +393,409 @@ fn default_value(comptime field: std.builtin.Type.StructField) ?field.field_type
     else
         null;
 }
+
+// CLI parsing makes a liberal use of `fatal`, so testing it within the process is impossible.
+//
+// Rater than making the code more complicated by abstracting over termination logic, test it out
```
sentientwaffle:
```suggestion
// Rather than making the code more complicated by abstracting over termination logic, test it out
```

src/flags.zig
```diff
@@ -233,3 +393,409 @@ fn default_value(comptime field: std.builtin.Type.StructField) ?field.field_type
...
+            const flags_exe_buf = try gpa.create([std.fs.MAX_PATH_BYTES]u8);
+            errdefer gpa.destroy(flags_exe_buf);
+
+            const src_dir = comptime std.fs.path.dirname(@src().file).?;
+            const argv = [_][]const u8{
+                zig_exe,
+                "build-exe",
+                "--main-pkg-path",
+                src_dir,
+                src_dir ++ "/testing/flags_test_program.zig",
```
sentientwaffle: Or maybe put the file directly in `src/` (alongside `flags.zig`)? `src/testing/` doesn't contain many tests, it is mostly tools for writing tests. Though I guess this is sort of both :thinking:

src/flags.zig
```diff
@@ -168,34 +230,45 @@ fn parse_flag(comptime T: type, comptime flag: []const u8, arg: [:0]const u8) T
...
+    comptime unreachable;
+}
+
 /// Parse string value into an integer, providing a nice error message for the user.
-fn parse_flag_value_int(comptime T: type, comptime flag: []const u8, value: [:0]const u8) T {
+fn parse_value_int(comptime T: type, flag: []const u8, value: [:0]const u8) T {
+    assert((flag[0] == '-' and flag[1] == '-') or flag[0] == '<');
+
     return std.fmt.parseInt(T, value, 10) catch |err| {
         fatal("{s}: expected an integer value, but found '{s}' ({s})", .{
```
sentientwaffle: The Overflow error message should convey what limit that is being exceeded (e.g. include the name of the integer type in the message).

Previously `parse_cluster`:

```
error.Overflow => fatal("--cluster: value exceeds a 32-bit unsigned integer", .{}),
```

src/flags.zig
matklad: My motivation for stuffing that into `testing` was to keep non-test code more readable. Top-level src is pretty valuable, and adding one more entry there seems a bit wasteful. 

Ideally, I want to specify that main directly in this file (`flags.zig`), and that actually works, but then it sorta adds unnecessary things to production builds (I belive it doesn't _actually_ adds things due to lazy compiliation, but still makes me uneasy enough). 

Ohhh, wait, I think I know how to do this cleanly actually!

src/flags.zig
matklad: Yeah, good point, let me think about a better message here... I think that `exceeds a 32-bit unsigned integer` is perhaps less then ideal here, as this feels too implementation oriented (as a user, I'd rather not care how many bits do you need). But also for something like `--cluster` its not like we could say something better.

src/flags.zig
matklad: Ah, that doesn't work directly --- it looks like `++` returns an `[]const` slice, so I can't subseqently sort it. I could however use the right length here.

sentientwaffle (approved):

PR #1082, opened 2023-08-02
CLI: Use proper type for cache_transfers_posted size

sentientwaffle:


src/tigerbeetle/cli.zig
```diff
@@ -254,7 +254,7 @@ pub fn parse_args(allocator: std.mem.Allocator) !Command {
                         start.cache_transfers,
                     ),
                     .cache_transfers_posted = parse_cache_size_to_count(
-                        u256, // TODO(#264): Use actual type here, once exposed.
```
matklad, line 257: Heh, so we actually even change the size here? A good one!

matklad (approved):

src/tigerbeetle/cli.zig
sentientwaffle: Yeah, the `PostedGrooveValue` used to be the size of a `u256` (it was originally indexed by the u128 id instead of the u64 timestamp).

PR #1084, opened 2023-08-02
LSM: Remove `Tree.lookup_snapshot_max`

sentientwaffle:
- Remove `Tree.lookup_snapshot_max`; just use `snpashot_latest`.
- Clean up some compaction logic.

### DIFF
--- a/src/lsm/tree.zig
+++ b/src/lsm/tree.zig
@@ -709,27 +709,18 @@ pub fn TreeType(comptime TreeTable: type, comptime Storage: type) type {
 
             tree.compaction_op = op;
 
-            if (op < constants.lsm_batch_multiple) {
-                // There is nothing to compact for the first measure.
-                // We skip the main compaction code path first compaction bar entirely because it
-                // is a special case — its first beat is 1, not 0.
-                tree.compaction_phase = .skipped;
-
-                if (op + 1 == constants.lsm_batch_multiple) {
-                    tree.compact_mutable_table_into_immutable();
-                }
-
-                tree.compaction_callback = .{ .next_tick = callback };
-                tree.grid.on_next_tick(compact_finish_next_tick, &tree.compaction_next_tick);
-                return;
-            }
-
-            if (tree.grid.superblock.working.vsr_state.op_compacted(op)) {
-                // We recovered from a checkpoint, and must avoid replaying one bar of
-                // compactions that were applied before the checkpoint. Repeating these ops'
-                // compactions would actually perform different compactions than before,
-                // causing the storage state of the replica to diverge from the cluster.
-                // See also: compaction_op_min().
+            if (op < constants.lsm_batch_multiple or
+                tree.grid.superblock.working.vsr_state.op_compacted(op))
+            {
+                // Either:
+                // - There is nothing to compact for the first measure.
+                //   We skip the main compaction code path first compaction bar entirely because it
+                //   is a special case — its first beat is 1, not 0.
+                // - We recovered from a checkpoint, and must avoid replaying one bar of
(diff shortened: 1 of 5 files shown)


sentientwaffle:
(:arrow_up: Rebased to resolve conflicts.)

src/lsm/tree.zig
```diff
@@ -325,19 +312,24 @@ pub fn TreeType(comptime TreeTable: type, comptime Storage: type) type {
             key: Key,
             fingerprint: Fingerprint,
         ) LookupMemoryResult {
-            if (snapshot >= tree.lookup_snapshot_max.?) {
+            assert(snapshot <= snapshot_latest);
+
+            if (snapshot == snapshot_latest) {
```
batiati, line 317: Overall, getting rid of `Tree.lookup_snapshot_max` makes a lot of sense to me, but I think I'm missing some context here.

With this PR, the mutable table is only visible when `snapshot_latest` is passed. How are we going to handle cases where trees had different `lookup_snapshot_max` (eg IdTree and ObjectTree not sharing the same `lookup_snapshot_max`), but the snapshot provided could be found in both (eg `snapshot != snapshot_latest and snapshot >= tree.lookup_snapshot_max.?`)?

sentientwaffle: :thinking: Can you elaborate? I'm not following.

- By `snapshot_max` do you mean `snapshot_latest`?
- `lookup_snapshot_max` previously had the invariant "Between beats, all of a groove's trees share the same lookup_snapshot_max." -- I'm not sure what you mean by "eg IdTree and ObjectTree not sharing the same...".

This is a bit speculative since querying snapshots isn't implemented. Note the comment in `tree.zig`:
```
                // The mutable table is converted to an immutable table when a snapshot is created.
                // This means that a past snapshot will never be able to see the mutable table.
                // This simplifies the mutable table and eliminates compaction for duplicate puts.
```
So any queries to old snapshots will never hit the mutable table.

batiati: > * By `snapshot_max` do you mean `snapshot_latest`?

Yes, I fixed the typo 😳

> * `lookup_snapshot_max` previously had the invariant "Between beats, all of a groove's trees share the same lookup_snapshot_max." -- I'm not sure what you mean by "eg IdTree and ObjectTree not sharing the same...".

Exactly that invariant, if "between beats, all of a groove's trees share the same lookup_snapshot_max.", So I suppose it can diverge at some point. But I'm not sure if we can do lookups during this state (that's the context I might be missing).

> This is a bit speculative since querying snapshots isn't implemented. Note the comment in `tree.zig`:

Got it :+1:

batiati (approved):

src/lsm/tree.zig
sentientwaffle: > But I'm not sure if we can do lookups during this state (that's the context I might be missing).

We did perform lookups in this state, which is why each `Tree` had its own `lookup_snapshot_max`, rather than sharing a single snapshot between all `Tree`.

But since we now only hit the mutable table when `snapshot == snapshot_latest` that shouldn't be necessary.

---

For a little more context: The reason we needed `lookup_snapshot_max` in the first place is that (some weeks/months ago) compaction's output tables were added to the manifest before the tables' blocks were written to disk. But as part of an unrelated change that is no longer the case -- we only update the manifest _after_ the corresponding writes are finished. 

`lookup_snapshot_max`'s job was to guard those un-written tables from being read, so it isn't necessary anymore.

batiati: Thanks for the explanation @sentientwaffle 🎉

PR #1087, opened 2023-08-07
Add info about other subcommands when none are passed.

eatonphil:
Before switching to src/flags.zig we used to have a little more information about what is available if you pass in incorrect or missing arguments.

Though this still doesn't help the user very much yet if they do something like `./tigerbeetle what`.

```
./tigerbeetle
error: subcommand required.

Expected one of: format, start, version, client.

Or use -h, --help for more information.
```

### DIFF
--- a/src/flags.zig
+++ b/src/flags.zig
@@ -619,7 +619,11 @@ test "flags" {
     try t.check(&.{}, snap(@src(),
         \\status: 1
         \\stderr:
-        \\error: subcommand required
+        \\error: subcommand required.
+        \\
+        \\Expected one of: empty, prefix, pos, required, values.
+        \\
+        \\Or use -h, --help for more information.
         \\
     ));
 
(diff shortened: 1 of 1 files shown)


src/flags.zig
```diff
@@ -72,7 +72,34 @@ pub fn fatal(comptime fmt_string: []const u8, args: anytype) noreturn {
 pub fn parse_commands(args: *std.process.ArgIterator, comptime Commands: type) Commands {
     assert(@typeInfo(Commands) == .Union);
 
-    const first_arg = args.next() orelse fatal("subcommand required", .{});
+    const first_arg = args.next() orelse {
+        const fields = comptime blk: {
+            var fields: [:0]const u8 = "";
+            inline for (std.meta.fields(Commands)) |field, i| {
```
matklad: I _think_ this could be just `for`, without `inline`, because we are already in comptime.

eatonphil: You're right! Dropped.

matklad:
Let’s maybe use original single line phrasing we had before flags.zig? All other error messages are compact one-liners, so a line separated Proper sentence feels a bit out of place here.

eatonphil:
Ok, before switching to src/flags.zig it used to be:

```
$ ./tigerbeetle
error: no command provided, expected 'start', 'format', or 'version'
```

Now it is:

```
$ ./tigerbeetle
error: subcommand required, expected 'format', 'start', or 'version'
```

matklad (approved):

PR #1092, opened 2023-08-08
tests: fix docs testing with git subtrees

matklad:
Manually walking the file tree does not work with gitsubtrees, where `.git` might be elsewhere. Let's rather get the answer straight from the horses mouth with:

    $ git rev-parse --show-top-level

This adds dependency on the git executable, but this is fine, as we need it to clone the repo anyway, and git itself is reasonably cross platform.


eatonphil (approved):
Thanks!

PR #1093, opened 2023-08-08
Minor fingerprint improvements

batiati:
- Change the fingerprint for composite keys, taking only the prefix, not the entire key.
Rationale: We don't use composite keys for lookups. However, the bloom filter can be useful for scans with equality conditions by prefix.

- Propagate the key's fingerprint during the prefetch path.
Rationale: Hashing the key only once.

- Change PrefetchKeys' implementation from HashMap to ArrayHashMap.
Rationale: It's better for iteration.

### DIFF
--- a/src/lsm/groove.zig
+++ b/src/lsm/groove.zig
@@ -12,9 +12,11 @@ const TreeType = @import("tree.zig").TreeType;
 const GridType = @import("grid.zig").GridType;
 const CompositeKey = @import("composite_key.zig").CompositeKey;
 const NodePool = @import("node_pool.zig").NodePool(constants.lsm_manifest_node_size, 16);
+const Fingerprint = @import("bloom_filter.zig").Fingerprint;
 
 const snapshot_latest = @import("tree.zig").snapshot_latest;
 const compaction_snapshot_for_op = @import("tree.zig").compaction_snapshot_for_op;
+const key_fingerprint = @import("tree.zig").key_fingerprint;
 
 fn ObjectTreeHelpers(comptime Object: type) type {
     assert(@hasField(Object, "timestamp"));
@@ -387,12 +389,16 @@ pub fn GrooveType(
 
         const primary_field = if (has_id) "id" else "timestamp";
         const PrimaryKey = @TypeOf(@field(@as(Object, undefined), primary_field));
-        const PrefetchKey = struct { key: union(enum) {
-            id: PrimaryKey,
-            timestamp: u64,
-        }, level: u8 };
+        const PrefetchKey = struct {
+            key: union(enum) {
+                id: PrimaryKey,
+                timestamp: u64,
+            },
+            fingerprint: Fingerprint,
+            level: u8,
+        };
 
-        const PrefetchKeys = std.AutoHashMapUnmanaged(PrefetchKey, void);
+        const PrefetchKeys = std.AutoArrayHashMapUnmanaged(PrefetchKey, void);
 
         const PrefetchObjectsContext = struct {
             pub inline fn hash(_: PrefetchObjectsContext, object: Object) u64 {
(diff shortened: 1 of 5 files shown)


src/lsm/groove.zig
```diff
@@ -387,12 +389,16 @@ pub fn GrooveType(
...
+            key: union(enum) {
+                id: PrimaryKey,
+                timestamp: u64,
+            },
+            fingerprint: Fingerprint,
+            level: u8,
+        };
 
-        const PrefetchKeys = std.AutoHashMapUnmanaged(PrefetchKey, void);
+        const PrefetchKeys = std.ArrayListUnmanaged(PrefetchKey);
```
sentientwaffle: What is the reason for changing `PrefetchKeys` from a hashset to an arraylist? The former is nice since it automatically deduplicates prefetches. (e.g. imagine that all of the transfers in the batch share the same debit account id -- each one needs to prefetch that same account).

batiati: > The former is nice since it automatically deduplicates prefetches

That's so true, but I suppose the prefetch path now can take care of deduping ids efficiently.

The main reason for using an ArrayList is to remove the inherent cost of hashing the `PrefetchKey` struct and iterating over the hashtable keys, which seems costly compared with the ArrayList.

But I agree it's easier to reason about duplicated ids relying on a hashtable, not the prefetch path.

sentientwaffle: > That's so true, but I suppose the prefetch path now can take care of deduping ids efficiently.

:thinking: Ah, since for e.g. two reads (X and Y) to the same account, either:
- reads X and Y are issued concurrently (and thus coalesced by the grid's read path), or
- read X finishes before Y starts, so Y's lookup_from_memory hits

Is that what you mean?

batiati: Yes, although I haven't measured the worst case (eg a batch using the same two accounts, or even worse one entire lookup batch with the same id for a transfer that doesn't exists).

This approach is too risky, instead I'm benchmarking a ArrayHashMap which can provide both dedup and easy iteration over items.

src/lsm/tree.zig
```diff
@@ -347,8 +353,12 @@ pub fn TreeType(comptime TreeTable: type, comptime Storage: type) type {
...
         /// the existence of the key in the filter/data block. If any of the blocks needed to
         /// ascertain the existence of the key are not in the Grid cache, it bails out.
-        fn lookup_from_levels_cache(tree: *Tree, snapshot: u64, key: Key) LookupMemoryResult {
-            const fingerprint = bloom_filter.Fingerprint.create(stdx.hash_inline(key));
+        fn lookup_from_levels_cache(
+            tree: *Tree,
+            snapshot: u64,
+            key: Key,
+            fingerprint: Fingerprint,
+        ) LookupMemoryResult {
```
sentientwaffle, line 361: Can you add a `if (constants.verify)` check that verifies that `fingerprint` is indeed the fingerprint of the passed `key`? Along with a note that it is provided as an argument (rather than computed here) as an optimization.

src/lsm/groove.zig
```diff
@@ -387,12 +389,16 @@ pub fn GrooveType(
...
+            key: union(enum) {
+                id: PrimaryKey,
+                timestamp: u64,
+            },
+            fingerprint: Fingerprint,
+            level: u8,
+        };
 
-        const PrefetchKeys = std.AutoHashMapUnmanaged(PrefetchKey, void);
+        const PrefetchKeys = std.AutoArrayHashMapUnmanaged(PrefetchKey, void);
```
sentientwaffle: Probably wouldn't make much of a perf difference, but if instead of a `(PrefetchKey, void)` hashmap it was a `(union(enum) {id,timestamp}, struct {fingerprint, level}})` hashmap that would slightly reduce the amount of data being hashed.

batiati: I've tried that, not changed, just measurement noise.
Maybe it's a good design for correctness since it should be unique per `union(enum) {id,timestamp}` and not by the whole struct.

src/lsm/groove.zig
```diff
@@ -747,38 +752,49 @@ pub fn GrooveType(
...
                 };
 
                 // prefetch_enqueue() ensures that the tree's cache is checked before queueing the
                 // object for prefetching. If not in the LSM tree's cache, the object must be read
                 // from disk and added to the auxiliary prefetch_objects hash map.
-                switch (prefetch_key.key) {
+                switch (prefetch_entry.key_ptr.*) {
                     .id => |id| {
+                        if (constants.verify) {
+                            assert(std.meta.eql(
```
sentientwaffle, line 766: Nit, but I'd suggest `stdx.equal_bytes` to avoid the stack copy.

batiati: Not possible for this type:
```
comptime assert(std.meta.trait.hasUniqueRepresentation(T));
```

sentientwaffle: Ah, nevermind!

sentientwaffle (approved):

PR #1094, opened 2023-08-08
docs: move to flags & assorted fixes

matklad:

PR #1096, opened 2023-08-08
clients: use flags for integration.zig

matklad:


matklad:
follow up for https://github.com/tigerbeetle/tigerbeetle/pull/1094

src/clients/integration.zig
```diff
@@ -5,83 +5,56 @@
...
-        std.debug.print("--sample not set.\n", .{});
-        return error.SampleNotSet;
-    }
-
-    if (language == null) {
-        std.debug.print("--language not set.\n", .{});
-        return error.LanguageNotSet;
-    }
+    const language: Docs = switch (cli_args.language) {
+        .java => @import("./java/docs.zig").JavaDocs,
```
eatonphil: Do we not have a policy of keeping imports at the top of the file for readability? It seems weird to me, even though these aren't dynamic imports, that we "hide" the imports until down here.

matklad: Good question, I don't know! My guess would be that we would, in general, prefer to keep imports on top, but this isn't really an import of code, it's just a table with data, so its significantly less noisy to not repeat it in two places.

eatonphil: There is actually code in them too though, the hooks anywawy.

matklad: Raised discussion point in Slack. I don't feel too strongly here, but on balance compactness wins here for me. 

I was toying with the idea of defining 

```zig
pub const docs_for = .{
     .java = @import("./java/docs.zig"),
}
```

in the `doc_types.zig` so that we have a single source of truth there, instead of integration.zig and docs_generate repeating each other, but I don't like how this introduces circular dependencies.

eatonphil:
One of your steps is failing legitimately:

```
 zig/zig build client_docs -- --no-validate
[go] Generating main README
[go] Generating sample READMEs
[javascript] Generating main README
[javascript] Generating sample READMEs
[java] Generating main README
[java] Generating sample READMEs
[cs] Generating main README
[cs] Generating sample READMEs
➜  tigerbeetle git:(matklad/more-flags) ✗ gs
On branch matklad/more-flags
Your branch is up to date with 'origin/matklad/more-flags'.

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
        modified:   src/clients/dotnet/README.md

no changes added to commit (use "git add" and/or "git commit -a")
➜  tigerbeetle git:(matklad/more-flags) ✗ git diff
diff --git a/src/clients/dotnet/README.md b/src/clients/dotnet/README.md
index 8ceed7b8..c66801f6 100644
--- a/src/clients/dotnet/README.md
+++ b/src/clients/dotnet/README.md
@@ -446,3 +446,4 @@ $ dotnet clean --verbosity minimal
 $ dotnet build
 $ if ($env:TEST -eq 'true') { dotnet test } else { echo "Skipping client unit test" }
 ```
+
```

eatonphil:
GitHub's UI rendering broke when I pasted that. Here's a screenshot as a backup.
image

matklad:
Yeah, that's a regression introduced in #1094 and fixed in #1098, let me do a quick rebase here...

matklad:
Ok, lost my fight with GitHub here, re-opened at https://github.com/tigerbeetle/tigerbeetle/pull/1100

PR #1098, opened 2023-08-09
docs: more fixes & cleanups

matklad:


src/clients/docs_generate.zig
```diff
@@ -83,58 +83,42 @@ const MarkdownWriter = struct {
         mw.buf.clearRetainingCapacity();
     }
 
-    fn diff_on_disk(mw: *MarkdownWriter, filename: []const u8) !bool {
-        const file = std.fs.cwd().createFile(
-            filename,
-            .{ .read = true, .truncate = false },
-        ) catch |e| {
+    fn diff_on_disk(mw: *MarkdownWriter, filename: []const u8) !enum { same, different } {
```
eatonphil, line 86: Just checking: did you make sure this will correctly overwite if docs.zig has changed and that it will correctly not overwrite (keeping the old mtime) if it hasn't?

matklad: Yup, I manually checked that it did false overwrites before, and that it is correct now,  by inserting a print statement to where we actually write the file.

eatonphil (approved):

PR #1100, opened 2023-08-09
clients: use flags for integration.zig

matklad:


eatonphil:
What happened to the old one? :D 

Edit: Nevermind, read the messages over there.

matklad:
I originally implemented https://github.com/tigerbeetle/tigerbeetle/pull/1098 on top of `matklad/more-falgs` branch, but then decided to split out it to a separate PR, as it was unrelated. 

I have a `git spinoff` script for this use-case, which takes several commits on top of the current branch and moves them to a new branch (removing them from the old branch), exactly for this use-case of "I changed my mind and want a branch here after all". That script however works only with `main`/`master`, so in this case I did just `git switch -c matklad/diff-diff`, which left the old commits also on that old branch. 

So when I came back and rebased on top of master, the tip was the same commit as on the `diff-diff` branch. So when I pushed that, github noticed that #1096 now ends in a commit which is in the mastre branch, and hopefully closed the PR. Removing the offending commit and pushing the branch didn't reopen the PR though, as the logic here seems to be edge triggered. So I had to create a new PR from the same branch. 

:0)

eatonphil:
Looks ok, just want to make sure more of the otherwise optional tests pass before approving.

eatonphil (approved):

PR #1102, opened 2023-08-10
docs: make concurrency advice more clear

matklad:
~~Clients are thread safe because we want to provide hard-to-mess-up API for the users. However, each client is inherently sequential inside, as there could be only one request in flight for each particular client. So, if you want max throughput, you need to have multiple clients (but note that you loose relative ordering between requests of different clients).~~

Each client is spawns a thread, so creating many clients would be pretty bad. So let's rephrase this by removing the but.

### DIFF
--- a/src/clients/docs_generate.zig
+++ b/src/clients/docs_generate.zig
@@ -473,9 +473,8 @@ const Generator = struct {
             \\ID and replica addresses are both chosen by the system that
             \\starts the TigerBeetle cluster.
             \\
-            \\Clients are thread-safe. But for better
-            \\performance, a single instance should be shared between
-            \\multiple concurrent tasks.
+            \\Clients are thread-safe and a single instance should be shared
+            \\between multiple concurrent tasks.
             \\
             \\Multiple clients are useful when connecting to more than
             \\one TigerBeetle cluster.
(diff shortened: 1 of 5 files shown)


src/clients/docs_generate.zig
```diff
@@ -473,11 +473,11 @@ const Generator = struct {
             \\ID and replica addresses are both chosen by the system that
             \\starts the TigerBeetle cluster.
             \\
-            \\Clients are thread-safe. But for better
-            \\performance, a single instance should be shared between
-            \\multiple concurrent tasks.
+            \\Clients are thread-safe but serialize requests internally.
```
matklad: Any better word that "serialize" here? I want to convey that each client is thread-safe, but sequential, serialize has that sense, but also half of the people reading that would thing about JSON serialization or something...

eatonphil: "but order requests serially internally"?

eatonphil: Or separate the lly-es and "but internally order requests serially"

src/clients/docs_generate.zig
```diff
@@ -473,11 +473,11 @@ const Generator = struct {
...
+            \\For maximal throughput, each concurrent task should have
+            \\a separate client.
```
eatonphil: Isn't this the complete opposite of what the readme was saying before?

matklad: Yes, I think before we were saying the wrong thing, as we missed the "not" (the "But" sortof hints that the "not" should be there, and as far as I understand, you indeed want more clients for better perf).

batiati (approved):

PR #1106, opened 2023-08-10
LSM/SegmentedArray: SA.iterator_from_cursor(SA.search, ascending)

sentientwaffle:
Correctly handle cursors at the very end of the SegmentedArray.

This is analogous to beginning to forward-iterate an N-element array from index N. That is, no elements should be iterated. (Previously this panicked).

I think this isn't actually reachable at the moment. But the `ForestTableIterator` (not part of this changeset) needs this, since it passes the result of `SegmentedArray.search()` to `SegmentedArray.iterator_from_cursor(_, .ascending)`.

### DIFF
--- a/src/lsm/segmented_array.zig
+++ b/src/lsm/segmented_array.zig
@@ -806,11 +806,30 @@ fn SegmentedArrayType(
                 assert(cursor.node == 0);
                 assert(cursor.relative_index == 0);
 
-                return Iterator{
+                return .{
                     .array = array,
                     .direction = direction,
                     .cursor = .{ .node = 0, .relative_index = 0 },
                     .done = true,
+                };
+            } else if (cursor.node == array.node_count - 1 and
+                cursor.relative_index == array.count(cursor.node))
+            {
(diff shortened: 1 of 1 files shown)


src/lsm/segmented_array.zig
```diff
@@ -806,12 +806,31 @@ fn SegmentedArrayType(
...
                     .done = true,
                 };
+            } else if (cursor.node == array.node_count - 1 and
+                cursor.relative_index == array.count(cursor.node))
+            {
+                return switch (direction) {
+                    .ascending => .{
+                        .array = array,
+                        .direction = direction,
+                        .cursor = .{ .node = 0, .relative_index = 0 },
```
batiati: Is there any particular reason for resetting the cursor instead of preserving it + `done = true`?
```suggestion
                        .cursor = cursor,
```

sentientwaffle: Nope, no difference. Updated.

batiati (approved):

PR #1107, opened 2023-08-10
LSM/ManifestLevel: Order tables by (key, snapshot_min) instead of (key)

sentientwaffle:
Previously tables with the same key (specifically `key_max`) but a different `snapshot_min` were not ordered. Ordering them properly will enable the `ForestTableIterator` (not part of this changeset) to use a `(key, snapshot)` tuple as a cursor for iteration interleaved with mutations without skipping tables.

This has no measurable performance impact.

### DIFF
--- a/src/lsm/manifest_level.zig
+++ b/src/lsm/manifest_level.zig
@@ -3,6 +3,7 @@ const assert = std.debug.assert;
 const math = std.math;
 const mem = std.mem;
 const meta = std.meta;
+const maybe = stdx.maybe;
 
 const stdx = @import("../stdx.zig");
 const constants = @import("../constants.zig");
@@ -37,7 +38,39 @@ pub fn ManifestLevelType(
             .{},
         );
 
-        pub const Tables = SegmentedArray(TableInfo, NodePool, table_count_max, .{});
+        pub const Tables = SortedSegmentedArray(
+            TableInfo,
+            NodePool,
+            table_count_max,
+            KeyAndSnapshot,
+            struct {
+                inline fn key_from_value(table_info: *const TableInfo) KeyAndSnapshot {
+                    return .{
+                        .key_max = table_info.key_max,
+                        .snapshot_min = table_info.snapshot_min,
+                    };
+                }
+            }.key_from_value,
+            struct {
+                inline fn compare_key_and_snapshot(
+                    key_a: KeyAndSnapshot,
+                    key_b: KeyAndSnapshot,
+                ) math.Order {
+                    const order = compare_keys(key_a.key_max, key_b.key_max);
+                    if (order == .eq) {
+                        return std.math.order(key_a.snapshot_min, key_b.snapshot_min);
+                    } else {
+                        return order;
+                    }
+                }
+            }.compare_key_and_snapshot,
+            .{},
+        );
+
+        const KeyAndSnapshot = struct {
+            key_max: Key,
+            snapshot_min: u64,
+        };
 
         // A direct reference to a TableInfo within the Tables array.
         pub const TableInfoReference = struct { table_info: *TableInfo, generation: u32 };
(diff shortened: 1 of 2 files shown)


src/lsm/manifest_level.zig
```diff
@@ -37,7 +38,39 @@ pub fn ManifestLevelType(
...
+                        return std.math.order(key_a.snapshot_min, key_b.snapshot_min);
+                    } else {
+                        return order;
+                    }
+                }
+            }.compare_key_and_snapshot,
+            .{},
+        );
+
+        const KeyAndSnapshot = struct {
```
jorangreef:
```suggestion
        const KeyMaxSnapshotMin = struct {
```

src/lsm/manifest_level.zig
```diff
@@ -37,7 +38,39 @@ pub fn ManifestLevelType(
...
+                    key_b: KeyAndSnapshot,
+                ) math.Order {
+                    const order = compare_keys(key_a.key_max, key_b.key_max);
+                    if (order == .eq) {
+                        return std.math.order(key_a.snapshot_min, key_b.snapshot_min);
+                    } else {
+                        return order;
+                    }
+                }
+            }.compare_key_and_snapshot,
```
jorangreef:
```suggestion
            }.compare_key_max_snapshot_min,
```

src/lsm/manifest_level.zig
```diff
@@ -37,7 +38,39 @@ pub fn ManifestLevelType(
...
+                inline fn key_from_value(table_info: *const TableInfo) KeyAndSnapshot {
+                    return .{
+                        .key_max = table_info.key_max,
+                        .snapshot_min = table_info.snapshot_min,
+                    };
+                }
+            }.key_from_value,
+            struct {
+                inline fn compare_key_and_snapshot(
+                    key_a: KeyAndSnapshot,
```
jorangreef: How can we rework these interfaces to use `*const` here, so as to make explicit to the compiler that we don't require a copy?

src/lsm/manifest_level.zig
```diff
@@ -172,10 +205,16 @@ pub fn ManifestLevelType(
...
-            level.tables.insert_elements(node_pool, absolute_index, &[_]TableInfo{table.*});
+            const absolute_index_keys = level.keys.insert_element(node_pool, table.key_max);
+            assert(absolute_index_keys < level.keys.len());
+
+            const absolute_index_tables = level.tables.insert_element(node_pool, table.*);
+            assert(absolute_index_tables < level.tables.len());
+
+            // `keys` may have duplicate entries due to tables with the same key_max, but different
+            // snapshots.
+            maybe(absolute_index_keys != absolute_index_tables);
```
jorangreef: This is so critical that we should add a unit test and/or online verification that:

* The `key_max`/`snapshot_min` tuples are sorted correctly.
* The new tuple was inserted and can be found.
* The length of the arrays was increased by 1 without losing any other entries.

src/lsm/manifest_level.zig
sentientwaffle: I'll add a TODO to `SegmentedArray` to benchmark it. Based on [this talk](https://www.youtube.com/watch?v=dEIsJPpCZYg) I think there's a good chance that since the function is inline the copy is being optimized away anyhow.

jorangreef: Yes, I'm wondering more _how_ can we change the interface to make the `*const` explicit, to move in that direction, not to detect (i.e. via benchmark) whether the copy is happening or not.

My thinking on this is that we actually want to tell the compiler that we don't need a copy. Often, it can figure that out, but better to be explicit to help it out.

sentientwaffle: :thinking: Don't `key_a: KeyAndSnapshot` and `key_a: *const KeyAndSnapshot` both tell the compiler that we don't need a copy? i.e. (since `KeyAndSnapshot` has no pointers) both of those make it clear that the function does not mutate `key_a`'s memory, so it is free to use a pointer of a copy (whichever the optimizer decides is better).

jorangreef: The two are subtly different. The former _may_ optimize the copy (if the compiler can see we don't need it). The latter explicitly says we _must_ and are willing to work off a const reference.

We have a few places where we've leaned on the compiler, but lately I think we should move away from this to be explicit and tell the compiler exactly what we want. This way, we don't risk performance regressions in future if optimizations decide differently.

But most of all, it's more precise. We have the opportunity to instruct the compiler that we can work off a const reference. i.e. That we're sure there are no aliasing issues.

jorangreef: It's also important, since it changes interface, and these tend to be viral.

sentientwaffle: > The latter explicitly says we _must_

This sounds like a drawback to me -- for small keys (such as `KeyAndSnapshot`) copying may well be faster than the indirection of a pointer.

batiati: I think we rely on a well-defined compiler behavior here: **inline functions**.

Since inline functions in Zig are the closest equivalent of C macros, `inline fn func(x: u32) { return x + 1; }` should behave _almost_ as `#DEFINE func(x) (x+1)`, that is, there's no stack involved to copy the argument in the first place.

jorangreef: > This sounds like a drawback to me -- for small keys (such as KeyAndSnapshot) copying may well be faster than the indirection of a pointer.

Touché, agreed.

Thanks @batiati also for the clarification on inline functions.

This is like a bugbear for me that keeps coming up. 😛

batiati (approved):

PR #1108, opened 2023-08-10
LSM/Schema: Include `timestamp=table.snapshot_min` in Filter and Data blocks

sentientwaffle:
(Table index blocks already set `timestamp=snapshot_min`).

This additional information is useful for troubleshooting missing blocks following a state sync + table repair.

---
Also add some simple validation to `header_from_block()`.


batiati (approved):
👍

PR #1111, opened 2023-08-11
LSM: Include tree_id in table block headers

sentientwaffle:
The assertions here aren't particularly powerful, but this is useful for debugging state sync's table repair.

I included the id in an `extern struct` rather than just directly using the `parent: u128` since tree id will be shrunk down to a `u16` as part of the upcoming "unified manifest" changes.

### DIFF
--- a/src/lsm/compaction.zig
+++ b/src/lsm/compaction.zig
@@ -104,8 +104,7 @@ pub fn CompactionType(
         };
 
         // Passed by `init`.
-        tree_name: []const u8,
-        tree_id: u128,
+        tree_config: Tree.Config,
 
         // Allocated during `init`.
         iterator_a: TableDataIterator,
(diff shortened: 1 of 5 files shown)


src/lsm/schema.zig
```diff
@@ -15,6 +15,7 @@
 //! Index block schema:
 //! │ vsr.Header                   │ operation=BlockType.index,
 //! │                              │ context=schema.TableIndex.Context,
```
batiati: Unrelated: `context=schema.TableFilter.Context` as described in the "Index block" comment section is inconsistent with the lowercase `context=schema.TableFilter.context` used by the "Filter block" and "Data block" ones.

This made me wonder if `Parent` should be lowercase as well or if it's the other way around.

Maybe we could adopt `:` to indicate "type of" and `=` to indicate "value of" for those comments, for example:

```
context : schema.TableIndex.Context,
parent : schema.TableIndex.Parent,
request=@sizeOf(Key)
timestamp=snapshot_min
```

src/lsm/compaction.zig
```diff
@@ -105,6 +105,7 @@ pub fn CompactionType(
 
         // Passed by `init`.
         tree_name: []const u8,
+        tree_id: u128,
```
batiati: Suggestion: instead of adding the fields `tree_name` and `tree_id` we could reference `tree_config: Tree.Config` (or even better `*const Tree.Config` if possible).

IMO it reads better and decouples these fields from the compaction logic.

batiati (approved):

PR #1112, opened 2023-08-11
Update Zig from 0.10.1 to 0.11.0

kprotty:
With this, we should finally be tracking the latest tagged release. View the [release notes](https://ziglang.org/download/0.11.0/release-notes.html) for some relevant information on upgrading.


src/aof.zig
```diff
@@ -59,13 +59,13 @@ pub const AOFEntry = extern struct {
     }
 
     pub fn header(self: *AOFEntry) *Header {
-        return @ptrCast(*Header, &self.message);
+        return @as(*Header, @ptrCast(&self.message));
```
matklad: After this PR is merger, we need to do a pass over the codebase to check how much of these `@as` could be elliminated. No need to do this in this PR though!

src/hash_map.zig
```diff
@@ -104,7 +104,7 @@ pub const StringIndexContext = struct {
     }
 
     pub fn hash(self: @This(), x: u32) u64 {
-        const x_slice = mem.sliceTo(@ptrCast([*:0]const u8, self.bytes.items.ptr) + x, 0);
+        const x_slice = mem.sliceTo(@as([*:0]const u8, @ptrCast(self.bytes.items.ptr)) + x, 0);
```
matklad: for hash map, I think we can now just switch to upstream impl?

matklad: I guess I'll see this in a further commit!

src/testing/storage.zig
```diff
@@ -506,19 +506,13 @@ pub const Storage = struct {
     ) *const superblock.SuperBlockHeader {
         const offset = vsr.Zone.superblock.offset(superblock.SuperBlockZone.header.start_for_copy(copy_));
         const bytes = storage.memory[offset..][0..comptime superblock.SuperBlockZone.header.size_max()];
-        return @alignCast(
-            @alignOf(superblock.SuperBlockHeader),
-            mem.bytesAsValue(superblock.SuperBlockHeader, bytes),
-        );
+        return @alignCast(mem.bytesAsValue(superblock.SuperBlockHeader, bytes));
```
matklad, line 509: Oh, I think I need to check if we now can introduce `stdx.bytes_as_value_but_give_error_at_compile_time_if_I_mess_alignment_please`!

src/constants.zig
```diff
@@ -298,7 +298,7 @@ pub const connection_delay_min_ms = config.process.connection_delay_min_ms;
 pub const connection_delay_max_ms = config.process.connection_delay_max_ms;
 
 /// The maximum number of outgoing messages that may be queued on a replica connection.
-pub const connection_send_queue_max_replica = std.math.max(std.math.min(clients_max, 4), 2);
+pub const connection_send_queue_max_replica = @max(@min(clients_max, 4), 2);
```
matklad, line 301: After this PR we should add `stdx.clamp`, the minmax combo is too easy to mess up,

src/vsr/replica.zig
```diff
@@ -8569,10 +8569,10 @@ const DVCQuorum = struct {
             const dvc_commit_max_pipeline =
                 dvc.header.op -| constants.pipeline_prepare_queue_max;
 
-            commit_max_ = std.math.max(commit_max_, dvc_commit_max_tail);
-            commit_max_ = std.math.max(commit_max_, dvc_commit_max_pipeline);
-            commit_max_ = std.math.max(commit_max_, dvc.header.commit);
-            commit_max_ = std.math.max(commit_max_, dvc_headers.slice[0].commit);
```
matklad, line 8579: TODO for after PR: this would be _perfect_ to use vararg capability of `@max`

src/config.zig
```diff
@@ -168,7 +168,7 @@ const ConfigCluster = struct {
     /// It is used to assert that all cluster members share the same config.
     pub fn checksum(comptime config: ConfigCluster) u128 {
         @setEvalBranchQuota(10_000);
-        var hasher = std.crypto.hash.Blake3.init(.{});
+        var hasher = std.crypto.hash.blake2.Blake2s256.init(.{});
```
matklad, line 174: Is Blake3 == Blake2s256? If it is, might make sense to add a comment here. Also, maybe `TODO: switch to vsr.checksum once that works at comptime`? (I tihnk I have an issue tracking that,but comment won't hurt)

src/vsr/superblock_quorums.zig
```diff
@@ -97,7 +97,7 @@ pub fn QuorumsType(comptime options: Options) type {
 
             for (copies, 0..) |*copy, index| quorums.count_copy(copy, index, threshold);
 
-            std.sort.sort(Quorum, quorums.slice(), {}, sort_priority_descending);
```
matklad, line 100: Unreleated, but I think `sort.insertion` actually would be good here, as the size is small, and we could save on codegen here. I wonder if should add `stdx.tiny_sorts` which asserts that the size of the slice is `<= 32`?

matklad (commented):
Firs batch it!

kassane:
Is it intended add **zig-pkg** support for `c_client`?

Missing `lib.installHeader("Path/tb_client.h", "tb_client.h");` copying the Tigerbeetle package header to the client application.

### My next update to tb-cpp using zig-pkg

#### My `build.zig` app:

```zig
const libtb_dep = b.dependency("tigerbeetle", .{
        .target = info.target,
        .optimize = info.optimize,
    });
const libtb = libtb_dep.artifact("tigerbeetle");


// get includepath from pkg to zig-cache/i/{hash-pkg}/include
for (libtb.include_dirs.items) |include| {
            exe.include_dirs.append(include) catch @panic("Unavailable include_dirs");
        }

// or to zig-out/include
lib.installLibraryHeaders(libtb);

exe.linkLibrary(libtb);
```

matklad:
>Is it intended add zig-pkg support for c_client?

Probably not right off the bat, as the immediate goal is to just move the codebase over to the new zig. 

Once we are over that hurdle though, we could indeed extract relevant bits of TigerBeetle into proper zig packages.

build.zig
```diff
@@ -1149,9 +1250,9 @@ const GitCloneStep = struct {
 /// - From MacOS, similar to https://github.com/ziglang/zig/issues/9711#issuecomment-1090071087.
 /// Workaround: Just setting different cache folders for each platform.
 fn builder_with_isolated_cache(
```
matklad, line 1258: After the PR lands, we need to check if we can remove this altogether, perhaps the underlying bugs were fixed?

src/stdx.zig
```diff
@@ -7,8 +7,8 @@ const assert = std.debug.assert;
 /// TODO(Zig): Remove these and import directly from std.
 /// See https://github.com/ziglang/zig/pull/15989.
 /// This backported HashMap is only needed when fetchRemove() will be used.
-pub const HashMapUnmanaged = @import("./hash_map.zig").HashMapUnmanaged;
-pub const AutoHashMapUnmanaged = @import("./hash_map.zig").AutoHashMapUnmanaged;
+pub const HashMapUnmanaged = std.HashMapUnmanaged;
+pub const AutoHashMapUnmanaged = std.AutoHashMapUnmanaged;
```
matklad, line 11: :tada:

src/lsm/set_associative_cache.zig
```diff
@@ -570,11 +569,11 @@ fn PackedUnsignedIntegerArray(comptime UInt: type) type {
 
     assert(builtin.target.cpu.arch.endian() == .Little);
     assert(@typeInfo(UInt).Int.signedness == .unsigned);
-    assert(@typeInfo(UInt).Int.bits < @bitCount(u8));
+    assert(@typeInfo(UInt).Int.bits < @bitSizeOf(u8));
     assert(math.isPowerOfTwo(@typeInfo(UInt).Int.bits));
 
-    const word_bits = @bitCount(Word);
```
matklad: Wait, what happened here? `@bitCount` isn't a Zig builtin... Is this some ancient dead code?

src/stdx.zig
```diff
@@ -428,50 +428,3 @@ pub fn update(base: anytype, diff: anytype) @TypeOf(base) {
     }
     return updated;
 }
-
-// TODO(Zig): No need for this function once Zig is upgraded
-// and @fieldParentPtr() can be used for unions.
-// See: https://github.com/ziglang/zig/issues/6611.
-pub inline fn union_field_parent_ptr(
```
matklad, line 435: :tada: :tada: :tada:

src/aof.zig
```diff
@@ -431,7 +431,7 @@ pub fn aof_merge(
     // located.
     try stdout.print("Building checksum map...\n", .{});
     var current_parent: ?u128 = null;
-    for (aofs[0..aof_count], 0..) |*aof, i| {
+    for (&aofs[0..aof_count], 0..) |*aof, i| {
```
matklad: TODO: we also probably want convert a bunch of while loops to `for (0..n) |i|` form...

src/vsr/replica.zig
```diff
@@ -3893,7 +3893,7 @@ pub fn ReplicaType(
             }
 
             message.header.* = .{
-                .size = @as(u32, @intCast(@sizeOf(Header) * (1 + self.view_headers.array.len))),
+                .size = @as(u32, @intCast(@as(usize, @sizeOf(Header)) * (1 + self.view_headers.array.len))),
```
matklad, line 3897: I wonder we could reduce this to a signle `@as`

src/clients/dotnet/dotnet_bindings.zig
```diff
@@ -134,23 +134,21 @@ fn get_mapped_type_name(comptime Type: type) ?[]const u8 {
 }
 
 fn to_case(comptime input: []const u8, comptime case: enum { camel, pascal }) []const u8 {
-    comptime {
```
matklad, line 137: Why do we need to return `comptime` here? Seems important to signal that this code is intended to be run at comptime.

src/clients/go/go_bindings.zig
```diff
@@ -46,7 +46,9 @@ fn get_mapped_type_name(comptime Type: type) ?[]const u8 {
 }
 
 fn to_pascal_case(comptime input: []const u8, comptime min_len: ?usize) []const u8 {
-    comptime {
+    // TODO: Cleanup when this is fixed after Zig 0.11.
```
matklad: Let's use more greppable `TODO(Zig)`

src/flags.zig
```diff
@@ -489,7 +489,13 @@ test "flags" {
...
 
         fn init(gpa: std.mem.Allocator) !T {
-            const zig_exe = std.os.getenv("ZIG_EXE") orelse return error.SkipZigTest;
+            // TODO: Avoid std.os.getenv() as it currently causes a linker error on windows.
+            // See: https://github.com/ziglang/zig/issues/8456
+            const zig_exe = std.process.getEnvVarOwned(gpa, "ZIG_EXE") catch |e| switch (e) {
+                error.EnvironmentVariableNotFound => return error.SkipZigTest,
+                error.InvalidUtf8 => unreachable,
+                error.OutOfMemory => return error.OutOfMemory,
+            };
```
matklad, line 565: `defer gpa.free(zig_exe)`

src/testing/cluster.zig
```diff
@@ -42,7 +42,9 @@ pub const Failure = enum(u8) {
 /// with a replica index.
 const client_id_permutation_shift = constants.members_max;
 
-pub fn ClusterType(comptime StateMachineType: fn (comptime Storage: type, comptime constants: anytype) type) type {
+// TODO: Once Zig is upgraded from 0.11, change StateMachineType from anytype back to
```
matklad: `TODO(Zig)`

src/config.zig
kprotty: Good point on the comment. It was changed due to Blake3 in stdlib having a compile error internally.

src/clients/go/docs.zig
```diff
@@ -11,11 +12,20 @@ fn go_current_commit_pre_install_hook(
...
 ) !void {
     for (&[_][]const u8{ "go.mod", "go.sum" }) |file| {
-        std.os.unlink(try std.fmt.allocPrint(
+        const path = try std.fmt.allocPrint(
             arena.allocator(),
             "{s}/{s}",
             .{ sample_dir, file },
-        )) catch {
+        );
+        // TODO: Mismatching internal std.os.unlink() error in Zig 0.11.0
```
matklad: TODO(Zig)

src/lsm/set_associative_cache.zig
kprotty: It was `std.meta.bitCount` which I did a find-and-replace to `@bitCount` then realized "wait, that doesn't exist" and fixed it back to `@bitSizeOf`.

src/clients/dotnet/dotnet_bindings.zig
kprotty: `comptme blk:` has the same effect (the code in the block runs at comptime), but doesn't hit the issue of the compiler thinking `return output[0..len]` happens at runtime (even inside `comptime {}`) for some reason.

matklad:
Ok, this looks good to me! Compiled post-merge checklist from non-blocking comments here: 

* https://github.com/tigerbeetle/tigerbeetle/pull/1112#discussion_r1291758901 needless `@as` cast
* https://github.com/tigerbeetle/tigerbeetle/pull/1112#discussion_r1291766926 add alignment-preseving versions of asBytes
* https://github.com/tigerbeetle/tigerbeetle/pull/1112#discussion_r1291773386 stdx.clamp.
* https://github.com/tigerbeetle/tigerbeetle/pull/1112#discussion_r1291774463 `@max` is now vararg
* !!! https://github.com/tigerbeetle/tigerbeetle/pull/1112#discussion_r1291794038 audit what's up with our config checksum. I _think_ we leak that into various places on disk
* https://github.com/tigerbeetle/tigerbeetle/pull/1112#discussion_r1291797718 stdx.tiny_sort
* https://github.com/tigerbeetle/tigerbeetle/pull/1112#discussion_r1293390649 remove builder_works_with_isolated_cache
* https://github.com/tigerbeetle/tigerbeetle/pull/1112#discussion_r1293409722 use `for(0..n) |i|`
* https://github.com/tigerbeetle/tigerbeetle/pull/1112#discussion_r1293412872 more `@as`

Client builds seem to fail though.

kprotty:
For the client builds, seems like an error in build.zig when porting `setOutputDir()` as the binaries aren't going to their correct folders.

src/lsm/manifest_level.zig
```diff
@@ -546,7 +546,7 @@ pub fn ManifestLevelType(
                 if (optimal == null or range.tables.len < optimal.?.range.tables.len) {
                     optimal = LeastOverlapTable{
                         .table = TableInfoReference{
-                            .table_info = @intToPtr(*TableInfo, @ptrToInt(table)),
```
batiati, line 549: We can replace this pattern `@intToPtr(*X, @ptrToInt(x_const))` with `@constCast` introduced in Zig 0.11.0.

matklad (approved):
Approving!

This currently fails client builds, but I think that it would be less pain overall fixing that after the merge, rather than rebasing this continiously (less paint still means a lot of pain though!)

PR #1114, opened 2023-08-11
testing: introduce TmpTigerBeetle

matklad:
TmpTigerBeetle is an utility for integration tests, which spawns a single node TigerBeetle
cluster in a temporary directory. This generalizes `run_with_tb` a bit.

Backstory: I am looking into simplifying how we do cilent doc generation. This is one intermediate step, but it seems pretty valuable as is --- it actually is surprising that we went so far without "tmpdir, but for tigerbeetle" utility. Guess that shows the extent we prefer simulation tests over, you know, real tests :D

### DIFF
--- a/src/shell.zig
+++ b/src/shell.zig
@@ -427,23 +427,29 @@ test "shell: expand_argv" {
 ///
 /// Caller is responsible for closing the dir.
 fn discover_project_root() !std.fs.Dir {
-    var current = try std.fs.cwd().openDir(".", .{}); // dup cwd so that the caller can close
-    errdefer current.close();
+    if (false) { // TODO(Zig): https://github.com/ziglang/zig/issues/16779
+        var current = try std.fs.cwd().openDir(".", .{}); // dup cwd so that the caller can close
+        errdefer current.close();
 
-    var level: u32 = 0;
-    while (level < 16) : (level += 1) {
-        if (current.statFile("build.zig")) |_| {
-            assert(try subdir_exists(current, ".github"));
-            return current;
-        } else |err| switch (err) {
-            error.FileNotFound => {
-                var parent = try current.openDir("..", .{});
-                current.close();
-                current = parent;
-            },
-            else => return err,
+        var level: u32 = 0;
+        while (level < 16) : (level += 1) {
+            if (current.statFile("build.zig")) |_| {
+                assert(try subdir_exists(current, ".github"));
+                return current;
+            } else |err| switch (err) {
+                error.FileNotFound => {
+                    var parent = try current.openDir("..", .{});
+                    current.close();
+                    current = parent;
+                },
+                else => return err,
+            }
         }
-    }
 
-    return error.DiscoverProjectRootDepthExceeded;
+        return error.DiscoverProjectRootDepthExceeded;
+    } else {
+        const this_file = @src().file;
+        const root_dir = std.fs.path.dirname(std.fs.path.dirname(this_file).?).?;
+        return try std.fs.cwd().openDir(root_dir, .{});
+    }
 }
--- a/src/shell.zig
+++ b/src/shell.zig
@@ -18,6 +18,8 @@ const std = @import("std");
 const builtin = @import("builtin");
 const assert = std.debug.assert;
 
+const Shell = @This();
+
 /// For internal use by the `Shell` itself.
 gpa: std.mem.Allocator,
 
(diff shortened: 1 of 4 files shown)


matklad:
cc @eatonphil

src/shell.zig
```diff
@@ -200,6 +222,70 @@ pub fn exec_stdout(shell: *Shell, comptime cmd: []const u8, cmd_args: anytype) !
...
+    // XXX: child.argv is undefined when we return. This is fine though, as it is only used during
+    // `spawn`.
+    return child;
+}
+
+/// Runs the zig compiler.
+pub fn zig(shell: Shell, comptime cmd: []const u8, cmd_args: anytype) !void {
+    const zig_exe = try shell.project_root.realpathAlloc(
+        shell.gpa,
+        comptime "zig/zig" ++ builtin.target.exeFileExt(),
```
eatonphil, line 269: > builtin.target.exeFileExt(),

TIL. Very nice.

eatonphil: Does the path separator not need to be dependent on the OS? i.e. zig\zig.

src/shell.zig
```diff
@@ -319,3 +405,28 @@ test "shell: expand_argv" {
         ),
     );
 }
+
+/// Finds the root of TigerBeetle repo.
+///
+/// Caller is responsible for closing the dir.
+fn discover_project_root() !std.fs.Dir {
+    var current = try std.fs.cwd().openDir(".", .{}); // dup cwd so that the caller can close
```
eatonphil: What happened to the git command you just switched `git_root` to using?

src/shell.zig
matklad: >TIL

Took me a while to hunt this one down! Obviously, there _should_ be some way to get the exe extention in the stdlib, but I had to grep my local copy of zig repo along and across to find it! `rg '".exe"'` did the trick

>Does the path separator not need to be dependent on the OS? i.e. zig\zig.

Windows APIs treat `/` and `\` the same. I _think_ `cmd.exe` might be confused by `\`, but I didn't check that .

src/shell.zig
matklad: `git rev-parse --show-top-level` is still the correct way to find git root. But here, I don't actually need git root. I need the root of this particular project. As we control the layout here, we can go for a simpler solution of looking for our specific root file. 

Plus, I want this to be a `Dir` rather than `[]const u8` --- Zig in general is happier when you operate with `fd`s rather than paths.

src/shell.zig
eatonphil: > Windows APIs treat / and \ the same. I think cmd.exe might be confused by \, but I didn't check that .

I was under the impression if you `exec` with `/` it will fail. Maybe I'm wrong. We'll see in CI :)

eatonphil:
We could also simplify integration tests of the CLI client/repl if we supported TmpTigerBeetle calling a Zig function, not just a shell command to run. Something to consider, not a requirement.

matklad:
>We could also simplify integration tests of the CLI client/repl if we supported TmpTigerBeetle calling a Zig function, not just a shell command to run. Something to consider, not a requirement.

The idea here is that TmpTigerBeetle doesn't run anything. It just creates a cluster, and let's the caller use it however they want. 

I'd imagine for repl the test could look like this:

```zig
const tb = try TmbTigerBetle.init(gpa);
defer tb.deinit();

const repl = shell.spawn(
    "tigerbeetle repl --addresses={port}", 
    { .port = tb.port_str.slice() },
);

for (test_cases) |test_case| {
   repl.stdin.writeAll(test_case.command);
   const response_got = repl.stdout.readLine();
   try std.testing.assertEqualStrings(response_got, test_case.response_want)
}
```

eatonphil:
> I'd imagine for repl the test could look like this:

Cool!

src/shell.zig
matklad: Heh, turns out that Zig on windows breaks on `..`!

https://github.com/ziglang/zig/issues/16779

matklad: >We'll see in CI :)

Seems ok!

src/testing/tmp_tigerbeetle.zig
```diff
@@ -0,0 +1,129 @@
...
+    // Parse stderr to learn the assigned port. After that, spawn a dedicated thread to read the
+    // rest of stderr to avoid blocking stderr. No need to read stdout as we don't print there.
+    var buf: [4096]u8 = undefined;
+    var buf_cursor: usize = 0;
+    var limit: u32 = 0;
+    const port = while (limit < 32) : (limit += 1) {
+        buf_cursor += try process.stderr.?.read(buf[buf_cursor..]);
+        if (parse_port(buf[0..buf_cursor])) |port| break port;
+    } else {
+        log.err("can't start TigerBeetle", .{});
```
cb22: Perhaps
```suggestion
        log.err("Failed to find port for TigerBeetle", .{});
```

src/testing/tmp_tigerbeetle.zig
```diff
@@ -0,0 +1,129 @@
...
+}
+
+test parse_port {
+    try std.testing.expectEqual(
+        parse_port("info(main): 0: cluster=1: listening on 127.0.0.1:39047\n"),
+        39047,
+    );
+}
+
+fn read_stderr_in_background(process: *const std.ChildProcess) !std.Thread {
```
cb22: What about making this print out what it reads back to stderr, so the output isn't lost?

cb22 (commented):
Just some minor nits. Looking forward to stealing this for the built-in benchmark subcommand :)

src/testing/tmp_tigerbeetle.zig
matklad: Yeah, I originally punted on that, but seems like a clear case of YGNI, added

src/testing/tmp_tigerbeetle.zig
matklad: :+1: , but change to lowercase first, I think that's what we use for logs.

cb22 (dismissed):

cb22:
@matklad LGTM, just needs a conflict resolved!

cb22 (approved):

PR #1117, opened 2023-08-13
Changed exception type to ArgumentOutOfRangeException

YohDeadfall:
`ArgumentOutOfRangeException` is a more specific about the failure reason than generalized `ArgumentException` and from which it derives. It suits better in the current place due to `concurrencyMax` having a range starting from zero.

### DIFF
--- a/src/clients/dotnet/TigerBeetle.Tests/IntegrationTests.cs
+++ b/src/clients/dotnet/TigerBeetle.Tests/IntegrationTests.cs
@@ -100,14 +100,14 @@ namespace TigerBeetle.Tests
         }
 
         [TestMethod]
-        [ExpectedException(typeof(ArgumentException))]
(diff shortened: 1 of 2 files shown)


batiati (changes requested):
Thank you, good catch! That's indeed [the most suitable exception](https://learn.microsoft.com/en-us/dotnet/api/system.argumentoutofrangeexception?view=net-7.0&redirectedfrom=MSDN#remarks) for this case.

Could you please change the integration tests to reflect the new expected exception as well?

https://github.com/tigerbeetle/tigerbeetle/blob/dd3ad231397ed853ee5c05ead57541a9dc6c4c6c/src/clients/dotnet/TigerBeetle.Tests/IntegrationTests.cs#L102;L114

YohDeadfall:
Done, changed expected types in the tests too.

batiati (approved):
👍

PR #1118, opened 2023-08-14
Consistent directory and print binary location

eatonphil:
I was confused since the Powershell version did keep you in the new directory but the bash version did not. So just make them both explicitly enter the TigerBeetle directory.

Also, it's confusing to see the success message but then not know where/what the binary is for sure. So this prints out the binary name.

### DIFF
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -44,7 +44,7 @@
 	./scripts/install.sh
     fi
 
-    echo "Successfully set up $(./tigerbeetle version).
+    echo "Successfully set up $(./tigerbeetle version) at ./tigerbeetle.
 
 To get started running TigerBeetle and interacting with it, see:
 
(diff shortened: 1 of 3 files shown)


bootstrap.sh
```diff
@@ -44,7 +44,7 @@
 	./scripts/install.sh
     fi
 
-    echo "Successfully set up $(./tigerbeetle version).
+    echo "Successfully set up $(./tigerbeetle version) at ./tigerbeetle.
```
matklad: Let's do `$(realpath ./tigerbeetle)` so that's unambiguous

matklad: Note sure what's the ps1 alternative off the top of my head.

eatonphil: Is realpath available everywhere? Alternatively I can do `$(pwd)/tigerbeetle`.

eatonphil: And same for Windows for that matter.

matklad: Yeah, that's _much_ better!

matklad (approved):

PR #1119, opened 2023-08-14
Do not deinit walker until after function is over

eatonphil:
entry.path needs to live on past the initializer block.

See https://github.com/tigerbeetle/tigerbeetle/actions/runs/5856689278/job/15877532448 for the failure.

### DIFF
--- a/src/clients/dotnet/docs.zig
+++ b/src/clients/dotnet/docs.zig
@@ -49,10 +49,12 @@ fn current_commit_post_install_hook(
     const generated_csproj_file_exists =
         file_or_directory_exists(generated_csproj_filename);
 
-    const csproj_filename = blk: {
-        var walker = try dir.walk(arena.allocator());
-        defer walker.deinit();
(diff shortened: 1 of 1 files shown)


src/clients/dotnet/docs.zig
```diff
@@ -49,10 +49,9 @@ fn current_commit_post_install_hook(
     const generated_csproj_file_exists =
         file_or_directory_exists(generated_csproj_filename);
 
+    var walker = try dir.walk(arena.allocator());
+    defer walker.deinit();
```
matklad, line 55: This needs a comment so that it's not refactored away.

eatonphil: (It will be caught in tests, for what it's worth, as it already was.)

eatonphil: I've added a comment.

matklad (approved):

PR #1120, opened 2023-08-14
VSR/Grid: read_block(.from_local,{_or_global})

sentientwaffle:
(Context: The grid scrubber will use this.)

- `read_block(.{ .from_local = ... }, ...)`: Read from the disk. If there is an issue (e.g. checksum failure), pass the encountered error into the callback.
- `read_block(.{ .from_local_or_global = ... }, ...)`: Read from the disk. If there is an issue (e.g. checksum failure), queue the read into the "faulty queue" for remote repair. Don't invoke the callback until the correct block is available.

Both of these have `cache_read`/`cache_update` options:
- `cache_read` allows the scrubber to skip the cache (since it only cares about disk content).
- `cache_read` also allows the caller to skip the cache if they are handling it separately via `read_block_from_cache` as an optimization (to avoid a nexttick).
- `cache_write` allows the scrubber to avoid filling the cache with cold data.

---

Remove `cache_coherent`. `Grid.verify_read()` can now depend on the `coherent` option, which is much simpler (albeit slightly less comprehensive).
But the old `cache_coherent` strategy won't work soon (due to `fulfill_block`) and would have needed to be _much_ more complicated (depending on a whole second bitset) so this is a good compromise.

### DIFF
--- a/src/lsm/compaction.zig
+++ b/src/lsm/compaction.zig
@@ -394,7 +394,7 @@ pub fn CompactionType(
                     .disk => |table_ref| {
                         compaction.state = .iterator_init_a;
                         compaction.context.grid.read_block(
-                            .{ .from_local_or_global= on_iterator_init_a },
+                            .{ .from_local_or_global_storage = on_iterator_init_a },
                             &compaction.read,
                             table_ref.table_info.address,
                             table_ref.table_info.checksum,
--- a/src/lsm/compaction.zig
+++ b/src/lsm/compaction.zig
@@ -393,8 +393,8 @@ pub fn CompactionType(
                     },
                     .disk => |table_ref| {
                         compaction.state = .iterator_init_a;
-                        compaction.context.grid.read_block_from_cluster(
-                            on_iterator_init_a,
+                        compaction.context.grid.read_block(
+                            .{ .from_local_or_global= on_iterator_init_a },
                             &compaction.read,
                             table_ref.table_info.address,
                             table_ref.table_info.checksum,
--- a/src/lsm/compaction.zig
+++ b/src/lsm/compaction.zig
@@ -393,12 +393,12 @@ pub fn CompactionType(
                     },
                     .disk => |table_ref| {
                         compaction.state = .iterator_init_a;
-                        compaction.context.grid.read_block_from_cache_or_storage(
+                        compaction.context.grid.read_block_from_cluster(
                             on_iterator_init_a,
                             &compaction.read,
                             table_ref.table_info.address,
                             table_ref.table_info.checksum,
-                            .index,
+                            .{ .cache_check = true, .cache_update = true },
                         );
                     },
                 }
--- a/src/lsm/compaction.zig
+++ b/src/lsm/compaction.zig
@@ -398,7 +398,7 @@ pub fn CompactionType(
                             &compaction.read,
                             table_ref.table_info.address,
                             table_ref.table_info.checksum,
-                            .{ .cache_check = true, .cache_update = true },
+                            .{ .cache_read = true, .cache_write = true },
                         );
                     },
                 }
--- a/src/lsm/level_data_iterator.zig
+++ b/src/lsm/level_data_iterator.zig
@@ -124,8 +124,8 @@ pub fn LevelTableValueBlockIteratorType(comptime Table: type, comptime Storage: 
                 // Refill `table_data_iterator` before calling `table_next`.
                 const table_ref = it.context.tables[it.table_index];
                 it.callback = .{ .level_next = callback };
-                it.context.grid.read_block_from_cluster(
-                    on_level_next,
+                it.context.grid.read_block(
+                    .{ .from_local_or_global = on_level_next },
                     &it.read,
                     table_ref.table_info.address,
                     table_ref.table_info.checksum,
--- a/src/lsm/level_data_iterator.zig
+++ b/src/lsm/level_data_iterator.zig
@@ -129,7 +129,7 @@ pub fn LevelTableValueBlockIteratorType(comptime Table: type, comptime Storage: 
                     &it.read,
                     table_ref.table_info.address,
                     table_ref.table_info.checksum,
-                    .{ .cache_check = true, .cache_update = true },
+                    .{ .cache_read = true, .cache_write = true },
                 );
             } else {
                 it.table_next(callback);
--- a/src/lsm/level_data_iterator.zig
+++ b/src/lsm/level_data_iterator.zig
@@ -125,7 +125,7 @@ pub fn LevelTableValueBlockIteratorType(comptime Table: type, comptime Storage: 
                 const table_ref = it.context.tables[it.table_index];
                 it.callback = .{ .level_next = callback };
                 it.context.grid.read_block(
-                    .{ .from_local_or_global = on_level_next },
+                    .{ .from_local_or_global_storage = on_level_next },
                     &it.read,
                     table_ref.table_info.address,
                     table_ref.table_info.checksum,
--- a/src/lsm/level_data_iterator.zig
+++ b/src/lsm/level_data_iterator.zig
@@ -124,12 +124,12 @@ pub fn LevelTableValueBlockIteratorType(comptime Table: type, comptime Storage: 
                 // Refill `table_data_iterator` before calling `table_next`.
                 const table_ref = it.context.tables[it.table_index];
                 it.callback = .{ .level_next = callback };
-                it.context.grid.read_block_from_cache_or_storage(
+                it.context.grid.read_block_from_cluster(
                     on_level_next,
                     &it.read,
                     table_ref.table_info.address,
                     table_ref.table_info.checksum,
-                    .index,
+                    .{ .cache_check = true, .cache_update = true },
                 );
             } else {
                 it.table_next(callback);
--- a/src/lsm/tree.zig
+++ b/src/lsm/tree.zig
@@ -532,7 +532,7 @@ pub fn TreeType(comptime TreeTable: type, comptime Storage: type) type {
                 assert(context.index_block_count <= constants.lsm_levels);
 
                 context.tree.grid.read_block(
-                    .{ .from_local_or_global = read_index_block_callback },
+                    .{ .from_local_or_global_storage = read_index_block_callback },
                     &context.completion,
                     context.index_block_addresses[context.index_block],
                     context.index_block_checksums[context.index_block],
@@ -556,7 +556,7 @@ pub fn TreeType(comptime TreeTable: type, comptime Storage: type) type {
                 };
 
                 context.tree.grid.read_block(
-                    .{ .from_local_or_global = read_filter_block_callback },
+                    .{ .from_local_or_global_storage = read_filter_block_callback },
                     completion,
                     blocks.filter_block_address,
                     blocks.filter_block_checksum,
@@ -582,7 +582,7 @@ pub fn TreeType(comptime TreeTable: type, comptime Storage: type) type {
                     );
 
                     context.tree.grid.read_block(
-                        .{ .from_local_or_global = read_data_block_callback },
+                        .{ .from_local_or_global_storage = read_data_block_callback },
                         completion,
                         context.data_block.?.address,
                         context.data_block.?.checksum,
--- a/src/lsm/tree.zig
+++ b/src/lsm/tree.zig
@@ -356,6 +356,7 @@ pub fn TreeType(comptime TreeTable: type, comptime Storage: type) type {
                 const index_block = tree.grid.read_block_from_cache(
                     table.address,
                     table.checksum,
+                    .{ .coherent = true },
                 ) orelse {
                     // Index block not in cache. We cannot rule out existence without I/O,
                     // and therefore bail out.
@@ -409,7 +410,11 @@ pub fn TreeType(comptime TreeTable: type, comptime Storage: type) type {
             checksum: u128,
             fingerprint: Fingerprint,
         ) enum { negative, possible, block_not_in_cache } {
-            if (tree.grid.read_block_from_cache(address, checksum)) |filter_block| {
+            if (tree.grid.read_block_from_cache(
+                address,
+                checksum,
+                .{ .coherent = true },
+            )) |filter_block| {
                 const filter_schema = schema.TableFilter.from(filter_block);
                 const filter_bytes = filter_schema.block_filter_const(filter_block);
                 if (!bloom_filter.may_contain(fingerprint, filter_bytes)) {
@@ -432,7 +437,11 @@ pub fn TreeType(comptime TreeTable: type, comptime Storage: type) type {
             negative,
             block_not_in_cache,
         } {
-            if (tree.grid.read_block_from_cache(address, checksum)) |data_block| {
+            if (tree.grid.read_block_from_cache(
+                address,
+                checksum,
+                .{ .coherent = true },
+            )) |data_block| {
                 if (Table.data_block_search(data_block, key)) |value| {
                     return .{ .positive = value };
                 } else {
@@ -522,12 +531,12 @@ pub fn TreeType(comptime TreeTable: type, comptime Storage: type) type {
                 assert(context.index_block_count > 0);
                 assert(context.index_block_count <= constants.lsm_levels);
 
-                context.tree.grid.read_block_from_cache_or_storage(
+                context.tree.grid.read_block_from_cluster(
                     read_index_block_callback,
                     &context.completion,
                     context.index_block_addresses[context.index_block],
                     context.index_block_checksums[context.index_block],
-                    .index,
+                    .{ .cache_check = true, .cache_update = true },
                 );
             }
 
@@ -546,12 +555,12 @@ pub fn TreeType(comptime TreeTable: type, comptime Storage: type) type {
                     .checksum = blocks.data_block_checksum,
                 };
 
-                context.tree.grid.read_block_from_cache_or_storage(
+                context.tree.grid.read_block_from_cluster(
                     read_filter_block_callback,
                     completion,
                     blocks.filter_block_address,
                     blocks.filter_block_checksum,
-                    .filter,
+                    .{ .cache_check = true, .cache_update = true },
                 );
             }
 
@@ -572,12 +581,12 @@ pub fn TreeType(comptime TreeTable: type, comptime Storage: type) type {
                         @as(f64, @floatFromInt(context.tree.filter_block_hits)),
                     );
 
-                    context.tree.grid.read_block_from_cache_or_storage(
+                    context.tree.grid.read_block_from_cluster(
                         read_data_block_callback,
                         completion,
                         context.data_block.?.address,
                         context.data_block.?.checksum,
-                        .data,
+                        .{ .cache_check = true, .cache_update = true },
                     );
                 } else {
                     context.tree.filter_block_misses += 1;
--- a/src/lsm/tree.zig
+++ b/src/lsm/tree.zig
@@ -536,7 +536,7 @@ pub fn TreeType(comptime TreeTable: type, comptime Storage: type) type {
                     &context.completion,
                     context.index_block_addresses[context.index_block],
                     context.index_block_checksums[context.index_block],
-                    .{ .cache_check = true, .cache_update = true },
+                    .{ .cache_read = true, .cache_write = true },
                 );
             }
 
@@ -560,7 +560,7 @@ pub fn TreeType(comptime TreeTable: type, comptime Storage: type) type {
                     completion,
                     blocks.filter_block_address,
                     blocks.filter_block_checksum,
-                    .{ .cache_check = true, .cache_update = true },
+                    .{ .cache_read = true, .cache_write = true },
                 );
             }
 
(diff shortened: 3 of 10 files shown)


src/lsm/compaction.zig
```diff
@@ -393,12 +393,12 @@ pub fn CompactionType(
                     },
                     .disk => |table_ref| {
                         compaction.state = .iterator_init_a;
-                        compaction.context.grid.read_block_from_cache_or_storage(
+                        compaction.context.grid.read_block_from_cluster(
                             on_iterator_init_a,
                             &compaction.read,
                             table_ref.table_info.address,
                             table_ref.table_info.checksum,
-                            .index,
```
sentientwaffle, line 401: Passing the `BlockType` into the grid read allowed us to validate it, but I decided with wasn't worth including in the API -- that validation will be performed when decoding the block anyways.

src/vsr/grid.zig
```diff
@@ -141,11 +154,6 @@ pub fn GridType(comptime Storage: type) type {
         cache: Cache,
         /// Each entry in cache has a corresponding block.
         cache_blocks: []BlockPtr,
-        /// Each entry in cache has a corresponding bit.
-        /// This bit tracks whether the cache block is definitely valid relative to the current
-        /// checkpoint (matches what should be on disk). At the conclusion of state sync, all blocks
-        /// are marked as invalid. This enables additional validation of cache consistency.
-        cache_coherent: std.DynamicBitSetUnmanaged,
```
sentientwaffle, line 148: Rather than tracking which cache entries are coherent ourselves (internal to the grid), each read must signal whether or not it can guarantee that its target is coherent.

Specifically, coherent reads are:
- opening the manifest log
- commit prefetch
- compaction

Other reads pass `coherent=false`:
- repair requests (via the network)
- scrubbing (in a follow-up PR). Scrubbing is technically coherent at the time that the read is submitted, but marking it as incoherent allows the read to not block checkpoints. (Which simplifies things, since this means that reads can never block checkpoints.)

src/vsr/grid.zig
```diff
@@ -765,76 +892,57 @@ pub fn GridType(comptime Storage: type) type {
...
-
+        }) ReadBlockResult {
             const header = mem.bytesAsValue(vsr.Header, block[0..@sizeOf(vsr.Header)]);
 
-            if (!header.valid_checksum()) {
-                log.err("invalid checksum at address {} (expected={})", .{ address, checksum });
-                return false;
-            }
+            if (!header.valid_checksum()) return .invalid_checksum;
+            if (header.command != .block) return .unexpected_command;
```
sentientwaffle, line 906: Sneaky bug fix here!

We need to verify the command before we validate the body checksum.

If this is a misdirected `command=prepare` or `command=reply`, its header is valid (`valid_checksum()`) -- but checking the body checksum could panic! Blocks (and `block`) have a 64KiB capacity, but prepares/replies may be as large as 1MiB.

src/vsr/grid.zig
```diff
@@ -765,76 +892,57 @@ pub fn GridType(comptime Storage: type) type {
...
-            if (block_type) |block_type_| {
-                assert(header.operation == block_type_.operation());
-            }
-            return true;
+            assert(header.op == expect.address);
+            return .{ .valid = block };
         }
 
-        fn read_block_resolve(grid: *Grid, read: *Grid.Read, block: BlockPtrConst) void {
+        fn read_block_resolve(grid: *Grid, read: *Grid.Read, result: ReadBlockResult) void {
```
sentientwaffle, line 919: Previously `read_block_resolve` always received a `block`. Now it can receive a block _or_ an error (e.g. "invalid checksum").

If it receives a block, the behavior is unchanged.

If it receives a block, for each read in the resolve-chain:

- If the read originated from `read_block_from_replica` (`tag(callback) == .local`) report the error. (For the scrubber, this will trigger repair).
- If the read originated from `read_block_from_cluster` (`tag(callback) == .remote`) queue the read in `read_faulty_queue` and signal to the replica that we are waiting for a repair.

src/vsr/grid.zig
```diff
@@ -981,11 +1028,16 @@ pub fn GridType(comptime Storage: type) type {
...
-                return;
+            assert(constants.verify);
+
+            const TestStorage = @import("../testing/storage.zig").Storage;
+            if (Storage != TestStorage) return;
 
             const actual_block = grid.superblock.storage.grid_block(address);
+            const actual_header = schema.header_from_block(actual_block);
+            const cached_header = schema.header_from_block(cached_block);
+            assert(cached_header.checksum == actual_header.checksum);
```
sentientwaffle, line 1015: Checking the checksum before the entire block buffer makes errors a little easier to figure out. (i.e. "wrong block" vs "padding mismatch")

jorangreef:
Thanks for the excellent overview!

src/lsm/tree.zig
```diff
@@ -531,12 +531,12 @@ pub fn TreeType(comptime TreeTable: type, comptime Storage: type) type {
                 assert(context.index_block_count > 0);
                 assert(context.index_block_count <= constants.lsm_levels);
 
-                context.tree.grid.read_block_from_cluster(
-                    read_index_block_callback,
+                context.tree.grid.read_block(
+                    .{ .from_local_or_global = read_index_block_callback },
```
jorangreef:
```suggestion
                    .{ .from_local_or_global_storage = read_index_block_callback },
```

These changes look great!

sentientwaffle: Alternatively, what do you think of `callback_block` and `callback_block_or_error`? That leaves the "local vs global" storage as an internal concern to the grid.

jorangreef: Those are also great.

If it tips the scales, then I think it is good to be explicit that a network request might be made. Perhaps also it's less likely to use the wrong field at the callsite if it's `from_local_or_global_storage`.

What do you think?

src/vsr/grid.zig
```diff
@@ -67,16 +67,26 @@ pub fn GridType(comptime Storage: type) type {
...
+            callback: ReadBlockCallback,
             address: u64,
             checksum: u128,
-            block_type: BlockType,
             /// The current checkpoint when the read began.
             /// Used to verify that the checkpoint does not advance while the read is in progress.
             checkpoint_id: u128,
 
-            read_cache: bool,
+            coherent: bool,
```
matklad, line 95: Comment about semantics of coherent. Also, is there a better word here? I _think_ "cache coherency" refers to consistency of data among different caches, not to the consistency between cache and the slower store. But that's nitpicking, coherence is also fine!

src/lsm/tree.zig
sentientwaffle: > If it tips the scales, then I think it is good to be explicit that a network request might be made. Perhaps also it's less likely to use the wrong field at the callsite if it's `from_local_or_global_storage`.

Ah, that is a very good point.

src/vsr/grid.zig
```diff
@@ -335,9 +390,9 @@ pub fn GridType(comptime Storage: type) type {
             return false;
         }
 
-        const Writing = enum { init, repair, none };
+        const Writing = enum { create, repair, none };
```
matklad: Maybe `s/none/not_writing/`? I misread `none` as "writing for an unknown reason" rather than "not writing".

src/vsr/grid.zig
```diff
@@ -571,117 +661,112 @@ pub fn GridType(comptime Storage: type) type {
...
                 }
             }
 
-            // When Read.read_cache is set, the caller of read_block() is responsible for calling
+            // When Read.cache_read is set, the caller of read_block() is responsible for calling
             // us via next_tick().
-            if (read.read_cache) {
-                if (grid.read_block_from_cache(read.address, read.checksum)) |cache_block| {
-                    grid.read_block_resolve(read, cache_block);
+            if (read.cache_read) {
```
matklad, line 775: Heh, that's some fully control flow here. I was wondering if maybe we could push the loop above to the synchronous path, so that this callback has to deal just with the cache, but that doesn't work. 

I guess this could be a bit better if we learn to pin data in cache -- that way we can synchronously decide whether we need to read from cache or from disk, simplifying the call-site.

src/vsr/grid.zig
```diff
@@ -571,117 +661,112 @@ pub fn GridType(comptime Storage: type) type {
...
 
             read.* = .{
                 .callback = callback,
                 .address = address,
                 .checksum = checksum,
-                .block_type = block_type,
-                .read_cache = true,
+                .coherent = callback == .from_local_or_global_storage,
+                .cache_read = options.cache_read,
+                .cache_write = options.cache_write,
```
matklad, line 725: Is it true that `if (cache_write) assert(coherent)` ?

sentientwaffle: In practice yes, but it isn't strictly necessary.

src/vsr/grid.zig
sentientwaffle: I agree that "coherent" isn't a perfect name here, I just couldn't figure out anything better.

src/vsr/grid.zig
```diff
@@ -742,11 +835,18 @@ pub fn GridType(comptime Storage: type) type {
...
-            std.mem.swap(BlockPtr, iop_block, cache_block);
-            @memset(iop_block.*, 0);
-            grid.cache_coherent.set(cache_index);
+            const cache_index =
+                if (read.cache_write) grid.cache.insert_index(&read.address) else null;
+            const block = block: {
+                if (read.cache_write) {
+                    const cache_block = &grid.cache_blocks[cache_index.?];
+                    std.mem.swap(BlockPtr, iop_block, cache_block);
+                    @memset(iop_block.*, 0);
```
matklad, line 848: Unrelated, but this memset feels weird. First, we are doing memset here at all, and we generally try to avoid memsets. 

Second, this means that we have a spare block to memeset to begin with! I'd maybe expect for cached reads to always use the block from the cache, and for the uncached reads for the caller to provide the block.

src/vsr/grid.zig
```diff
@@ -765,76 +865,57 @@ pub fn GridType(comptime Storage: type) type {
...
+                .{
+                    @tagName(result),
+                    read.address,
+                    read.checksum,
+                    header.op,
+                    header.checksum,
+                },
+            );
+
             // Don't cache a corrupt or incorrect block.
```
matklad: We cache the block conditionally on `read.cache_wire`, but here we remove it unconditionally. This feels wrong: if a scrubber finds a bad block, but it is actually still cached, we don't want to evict the cached copy

src/vsr/grid.zig
```diff
@@ -765,76 +865,57 @@ pub fn GridType(comptime Storage: type) type {
...
+            const result = read_block_validate(block.*, .{
                 .address = read.address,
                 .checksum = read.checksum,
-                .block_type = read.block_type,
-            })) {
-                grid.read_block_resolve(read, cache_block.*);
+            });
+
+            if (result == .valid) {
+                grid.read_block_resolve(read, .{ .valid = block.* });
```
matklad: This is `grid.read_block_resolve(read, result);`, right?

Would the control flow we simpler as

```zig
if (result != valid) {

}
grid.read_block_resolve(read, result);
```

so that there's only one call to resolve?

src/vsr/grid.zig
sentientwaffle: Great catch!

src/vsr/grid.zig
sentientwaffle: Right now we rely on blocks being zeroed in a bunch of places, even though there is only one where it really matters (constructing filter blocks). I'm planning on fixing all of that, probably as part of the grid block pool changes.

Swapping the iop block into the cache like this is nice since it means we don't have to "reserve" a slot in the cache for the duration of the read. It is worth reexamining that once we have a grid block pool, though.

src/vsr/replica.zig
```diff
@@ -2210,7 +2205,7 @@ pub fn ReplicaType(
                             request.block_address,
                             request.block_checksum,
                         });
-                        continue :request_loop;
+                        continue :next_request;
```
matklad, line 2208: Love  how smoothly this reads!

src/vsr/replica.zig
```diff
@@ -2243,17 +2238,20 @@ pub fn ReplicaType(
                     .message = reply.ref(),
                 };
 
-                self.grid.read_block_repair(
-                    on_request_blocks_read_repair,
+                self.grid.read_block(
+                    .{ .from_local_storage = on_request_blocks_read_repair },
```
```suggestion
                    .{ .from_local_storage = on_request_blocks_read_block },
```

src/vsr/replica.zig
```diff
@@ -2263,25 +2261,27 @@ pub fn ReplicaType(
...
-            log.debug("{}: on_request_blocks: block found " ++
-                "(destination={} address={} checksum={})", .{
+            log.debug("{}: on_request_blocks: success: (destination={} address={} checksum={})", .{
                 self.replica,
                 read.destination,
                 grid_read.address,
                 grid_read.checksum,
             });
 
+            stdx.copy_disjoint(.inexact, u8, read.message.buffer, result.valid);
```
matklad, line 2283: It feels like we should be able to eliminate this copy in theory, as we could read into this buffer in the first place. 

If we go for refcounted blocks, we could even make block=message

src/lsm/level_data_iterator.zig
```diff
@@ -124,12 +124,12 @@ pub fn LevelTableValueBlockIteratorType(comptime Table: type, comptime Storage:
...
                 it.callback = .{ .level_next = callback };
-                it.context.grid.read_block_from_cache_or_storage(
-                    on_level_next,
+                it.context.grid.read_block(
+                    .{ .from_local_or_global_storage = on_level_next },
                     &it.read,
                     table_ref.table_info.address,
                     table_ref.table_info.checksum,
-                    .index,
+                    .{ .cache_read = true, .cache_write = true },
```
matklad, line 132: :thinking: it seems lsm pretty-much always wants a coherent reaad. 

Maybe its worth to add a helper function for this, which just delegates to `read_block` with appropriate arguments? 

I think from the perspective of LSM we want to work with a simple grid, which just has blocks. We don't want to thing about local_or_global storage and cache coherency. 

Basiaclly, we only need advanced read functionality for scrubber & replica's repiar, lsm could use simpler API. 


But also having `read_block` and `read_block_advanced` sound kinda stupid, so maybe we don't need it? But also, grepping for `read_block_advanced` is easier than grepping for `read_block` and noticing odd arguments...

src/vsr/replica.zig
sentientwaffle: Interesting! :thinking: 
So the `MessagePool` would have a bunch of `Message`s, some of which are 1MiB and some which are 64KiB.
Ironically this is similar to the original messagepool design, which treated header-only messages specially.

src/vsr/grid.zig
```diff
@@ -87,14 +105,17 @@ pub fn GridType(comptime Storage: type) type {
...
+            valid: BlockPtrConst,
+            /// Checksum of block header is invalid.
+            invalid_checksum,
+            /// Checksum of block body is invalid.
+            invalid_checksum_body,
+            /// The block header is valid, but its `header.command` is not `block`.
+            /// (This is possible due to misdirected IO).
+            unexpected_command,
+            /// The block is valid, but it is not the block we expected.
+            unexpected_checksum,
```
matklad, line 120: I am slightly uneasy here, as this errors look like something you might want to branch on, but we actually never branch IIUC, just log. Maybe

```
union(enum) {
    valid: BlockPtrConst,
    invalid: []const u8, // static string describing what's wrong
}
```

src/lsm/level_data_iterator.zig
sentientwaffle: Yes, LSM always wants a coherent read.

I went back and forth on this with the exact same reasons you gave :laughing: 
I ended up on just `read_block()` since it keeps things uniform, but I don't feel too strongly about it.

src/vsr/grid.zig
matklad: (the risk here is that at some point in the future some code starts distinguishing between `unexpected_checksum` vs `invalid_checksum`, and that uncovereds a sleeper bug where we compute the exact variant wrong, because it only affected the logs before. Also, the precedence is not entirely obvious here, the code could get refactored to return `unexpected_checksum` before, eg,  `invalid_checsum_body`)

src/lsm/level_data_iterator.zig
matklad: Yeah, let's keep as is then. But just one more option in case its a clear tiebreaker for you (it isn't for me):


We could go very specific here, make `read` private, and add

```
pub fn read_lsm
pub fn read_scrubber
pub fn read_replica
```

with appropriate specialization. That way all the callers of grid are sort-of inside the grid, and the logic of what's to cache or not cache stays encapsulated. 

Though, don't like it much --- I bet at some point something somewhere would want to use an odd combination of read flags.

sentientwaffle: Yeah, in particular I could see us experimenting with different cache options (and even adding more granular ones) for compaction in the future. For example, don't cache compaction input tables since they won't be needed for long.

src/vsr/grid.zig
sentientwaffle: What about:

```zig
union(enum) {
    valid: BlockPtrConst,
    invalid: enum {
            invalid_checksum,
            invalid_checksum_body,
            unexpected_command,
            unexpected_checksum,
    },
}
```

I'm on the fence about this.
It doesn't really fix the issues you mentioned, but maybe hints at them.

In particular, I like the `enum` for my rule-of-thumb of "don't discard information".
For example, in the future, we will have metrics, and an enum is nice since it is a static guarantee against metric label cardinality issues. (Though maybe we could assert that the string has a static lifetime to achieve something similar there...)

matklad: Good point about metrics! :thinking: I think that maybe the current code is best overall, as it's simple and flat, and those are big advantages! Maaybe add a comment at the definition site that "although we do distinguish between the reasons why the block is invalid, we only use this info for logging".

src/vsr/replica.zig
```diff
@@ -2178,24 +2179,18 @@ pub fn ReplicaType(
                 return;
             }
 
+            if (self.grid.canceling) |_| {
```
matklad, line 2182: Hm, shouldnt' we look at `self.syncing` here instead and assert grid.cancelling? `self.syncing` matches the current abstraction (replica) better.

src/vsr/grid.zig
```diff
@@ -217,14 +233,59 @@ pub fn GridType(comptime Storage: type) type {
...
             allocator.free(grid.cache_blocks);
 
-            grid.cache_coherent.deinit(allocator);
             grid.cache.deinit(allocator);
 
             grid.* = undefined;
         }
 
+        /// Invoke the callback after all pending repair-writes to blocks that are about to be
+        /// released have completed.
```
matklad: Could make this a bit more useful by adding "why" context

> ... so as to avoid data races with blocks allocated in the new checkpoint

src/vsr/grid.zig
```diff
@@ -571,117 +672,112 @@ pub fn GridType(comptime Storage: type) type {
...
-
             assert(address > 0);
-            assert(block_type != .reserved);
+
+            switch (callback) {
+                .from_local_storage => {
+                    maybe(grid.checkpointing == null);
+                    // We try to read the block even when it is free — if we recently released it,
+                    // it might be found on disk anyway.
+                    maybe(grid.superblock.free_set.is_free(address));
```
matklad, line 709: Just so that I understand the context better: 


When we read from_local_storage, for repair, we might be reading while we are checkpointing, and even while the corresponding block is free. This means we cold get a genuine data race on this block if, while the read is in flight, we checkpoint and overwrite the block. That is, in the kernel there might be be overlapping read-write requests. 

But that's OK! Because everything is checksummed, even if we get some torn garbage here, we just safely discard it.

src/vsr/grid.zig
```diff
@@ -571,117 +672,112 @@ pub fn GridType(comptime Storage: type) type {
...
+                    maybe(grid.writing(address, null) == .create);
+                },
+                .from_local_or_global_storage => {
+                    assert(grid.checkpointing == null);
+                    assert(!grid.superblock.free_set.is_free(address));
+                    assert(grid.writing(address, null) != .create);
```
matklad, line 715: We have assymetry here: in `read`, we do `grid.writing() != .create`, but in `write`, we do `grid.not_reading()`. It would be more symmetrict to introduce `grid.reading() enum { reading, not_reading, reading_repair}`. But probably not reall worth it.

matklad (dismissed):

src/vsr/replica.zig
sentientwaffle: I was thinking we want to be liberal in terms of "who can respond to `request_blocks`", since it is technically possible for the majority of the cluster (2/3) to state sync simultaneously.

See: https://github.com/tigerbeetle/tigerbeetle/blob/2e7f7b6f2ce5c2a2932a6fa9130549be8bdcd15d/src/vsr/replica_test.zig#L719-L726 and https://github.com/tigerbeetle/tigerbeetle/blob/2e7f7b6f2ce5c2a2932a6fa9130549be8bdcd15d/src/vsr/replica.zig#L4280-L4306

src/vsr/grid.zig
sentientwaffle: Yes, that's exactly correct.

matklad (approved):

PR #1121, opened 2023-08-15
VSR: Fix overflow due to zig 11 upgrade

sentientwaffle:
In Zig 0.9/0.10, `BoundedArray.len` is a `usize`.
In Zig 0.11, `BoundedArray.len` is a `std.math.IntFittingRange(0, buffer_capacity)`!

So adding `1` without first casting to `usize` can (potentially) overflow.

### DIFF
--- a/src/vsr/replica.zig
+++ b/src/vsr/replica.zig
@@ -3894,7 +3894,8 @@ pub fn ReplicaType(
             }
 
             message.header.* = .{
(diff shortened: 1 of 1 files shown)


matklad (approved):

matklad:
:fearful: :fearful: :fearful: this one's really scary, I think I've found at least one more case of this

matklad:
Important note: this can overflow because the comptime size of the array is `4 + 1 + 2 = 7 = 2^3 - 1`

PR #1122, opened 2023-08-15
zig: protect from BoundedArray.len overflow

matklad:
`@TypeOf(BoundedArray.len)` is the smallest integer that can hold the array length. As such, even something like `array.len + 1` can overflow, if (comptime) array len happens to be exactly power of two less one.

Protect against this by explicitly casting to a larger type. I've look for cases by grepping for `.len +`, `\+.*\n.len` and looking for BoundedArray

Not sure what's the principled solution here. Make our own bounded array which has a `.len()` which always returns at least `u32`?

### DIFF
--- a/src/stdx/bounded_array.zig
+++ b/src/stdx/bounded_array.zig
@@ -1,0 +1,76 @@
+const std = @import("std");
+const assert = std.debug.assert;
+
+/// A version of standard `BoundedArray` with TigerBeetle-idiomatic APIs.
+///
+/// See <https://github.com/tigerbeetle/tigerbeetle/pull/1121> for the original reason for
+/// wrapping --- we need an `fn count` which returns an `usize`, instead of potentially much smaller
+/// type which stores the length internally.
+pub fn BoundedArray(comptime T: type, comptime capacity: usize) type {
+    const Inner = @import("std").BoundedArray(T, capacity); // smuggle the std version past tidy
+
+    return struct {
+        inner: Inner = Inner{},
+
+        const Self = @This();
+
+        pub inline fn from_slice(items: []const T) error{Overflow}!Self {
+            return .{ .inner = try Inner.fromSlice(items) };
+        }
+
+        pub inline fn count(array: *const Self) usize {
+            return array.inner.len;
+        }
+
+        /// Returns count of elements in this BoundedArray in the specified integer types,
+        /// checking at compile time that it indeed can represent the length.
+        pub inline fn count_as(array: *const Self, comptime Int: type) Int {
+            return array.inner.len;
+        }
+
+        pub inline fn full(self: Self) bool {
+            return self.count() == capacity;
+        }
+
+        pub inline fn empty(self: Self) bool {
+            return self.count() == 0;
+        }
+
+        pub inline fn get(array: *const Self, index: usize) T {
+            return array.inner.get(index);
+        }
+
+        pub inline fn slice(array: *Self) []T {
+            return array.inner.slice();
+        }
+
+        pub inline fn const_slice(array: *const Self) []const T {
+            return array.inner.constSlice();
+        }
+
+        pub inline fn add_one_assume_capacity(array: *Self) *T {
+            return array.inner.addOneAssumeCapacity();
+        }
+
+        pub inline fn append_assume_capacity(array: *Self, item: T) void {
+            array.inner.appendAssumeCapacity(item);
+        }
+
+        pub inline fn writer(self: *Self) Inner.Writer {
+            return self.inner.writer();
+        }
+
+        pub inline fn swap_remove(array: *Self, index: usize) T {
+            return array.inner.swapRemove(index);
+        }
+
+        pub inline fn truncate(array: *Self, count_new: usize) void {
+            assert(count_new <= array.count());
+            array.inner.len = @intCast(count_new); // can't overflow due to check above.
+        }
+
+        pub inline fn clear(array: *Self) void {
+            array.inner.len = 0;
+        }
+    };
+}
--- a/src/lsm/compaction.zig
+++ b/src/lsm/compaction.zig
@@ -117,7 +117,7 @@ pub fn CompactionType(
 
         /// Manifest log appends are queued up until `finish()` is explicitly called to ensure
         /// they are applied deterministically relative to other concurrent compactions.
-        manifest_entries: std.BoundedArray(struct {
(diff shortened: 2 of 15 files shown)


jorangreef:
> Not sure what's the principled solution here. Make our own bounded array which has a .len() which always returns at least u32?

This was also my first thought, seeing @sentientwaffle's issue. The less we use stdlib long term, the better, since subtle changes can break determinism or introduce bugs like this, or worse.

sentientwaffle (dismissed):

matklad:
I guess what I'll do here rather is introduce `stdx.BoundedArray` which API wraps the std impl (so, don't copy code, but copy API, because its the API that's a footgun here)

``` zig
const BoundedArray = struct {
    impl: std.BoundedArray
}
```

sentientwaffle:
I agree that `BoundedArray.len() -> usize` is the right solution for this particular issue.

> The less we use stdlib long term, the better, since subtle changes can break determinism or introduce bugs like this, or worse.

:thinking: I chalk this up to using a stdlib that has currently has no stability guarantees, which is a hazard of Zig being pre-1.0.

In general though I still like to use stdlib when possible -- it gives us (or _will_ give us, as Zig matures) a great multiplier in terms of "eyes on bugs" and also "eyes on performance".

There is a risk that stdlib changes in surprising ways, but I think we can mitigate that and get the best of both worlds by:
- Write a test suite for stdlib data structures + functionality that we use. We could add snapshot tests to verify both results and determinism.
- More long-term: Add a test suite to check determinism between TB versions. e.g. "Run a randomly generated workload against version X and version X+1; assert that storage matches" (except when we expect it to mismatch).

jorangreef:
> I chalk this up to using a stdlib that has currently has no stability guarantees, which is a hazard of Zig being pre-1.0.

My concern is more for things like hash tables. I actually think Zig's hash map implementation is excellent, state of the art, and the API is already one of the best I've seen.

At the same time, we've hit some pretty interesting performance/liveness issues around subtle things like decisions on linear probing, and the effects on tombstones for performance and also for static memory allocation. None of these would break determinism, and they are areas where the std lib would be free to change policy.

We also have stricter demands on our data structures than the typical std lib user, most probably. Plus we don't use too many structures in the first place from the std lib (I was cautious around this in the beginning). For example, I think there's value in eventually having our own hash map which we understand deeply in terms of behavior, and which we test as well as anything, since it's a foundational structure in TigerBeetle.

If we see any interesting bugs patched upstream in a similar structure, there's also nothing stopping us from cherry picking.

jorangreef:
To add, we should definitely do the tests you describe as defense-in-depth, whether we vendor or not.

matklad:
I also concur with @sentientwaffle here --- long term, I do expect HashMap tests in Zig stdlib to be more thorough than what we have here. I guess my model is that we exercise, say, hash map, pretty deep, but we also do it along only our particular current workflow. So, when we refactor code, we could start exercising different code-paths, and leaning on Zig's stdlib.

This issue is actually a cool example of something we couldn't have caught in our data structure with just testing. To trigger this, we need a very specific comptime length of an array. So we could totally have written our own BoundedArray with the same problem, which could have manifested itself as a sleeper bug until one day we tweak our comptime config. 

With widely used libraries, I'd actually expect fewer of such sleeper bugs to arise, because they are used in a wider variety of contexts. 

But also, this I think is a function of how popular Zig ends up being. At the current level of popularity, given that we fix pretty obvious std bugs fairly often, it does seem that DIY could be more reliable.

src/stdx/bounded_array.zig
```diff
@@ -0,0 +1,68 @@
...
+    return struct {
+        inner: Inner = Inner{},
+
+        const Self = @This();
+
+        pub inline fn from_slice(items: []const T) error{Overflow}!Self {
+            return .{ .inner = try Inner.fromSlice(items) };
+        }
+
+        pub inline fn count(array: *const Self) usize {
```
matklad, line 21: Not sure whethere we use `fn count` or `fn len`, but I think it's count, based on RingBuffer, and the fact that `"count".len == "index".len`

src/stdx/bounded_array.zig
```diff
@@ -0,0 +1,68 @@
...
+        /// checking at compile time that it indeed can represent the length.
+        pub inline fn count_as(array: *const Self, comptime Int: type) Int {
+            return array.inner.len;
+        }
+
+        pub inline fn full(self: Self) bool {
+            return self.count() == capacity;
+        }
+
+        pub inline fn empty(self: Self) bool {
```
matklad, line 35: I've added an `empty()` method, but I only use it in postive sense. I find `array.count() > 0` to be more readable than `!array.empty()`, because negations are hard.

src/stdx/bounded_array.zig
```diff
@@ -0,0 +1,72 @@
+const std = @import("std");
+const assert = std.debug.assert;
+
+/// A version of `std.BoundedArray` with TigerBeetle-idiomatic APIs.
```
sentientwaffle: Maybe comment that the impetus for this wrapper is to add `fn count() usize` since `BoundedArray.len`'s type isn't a `usize`?

---

Also, would it be preferable to just use Zig's APIs for functions that are trivial wrappers? e.g. `appendAssumeCapacity`. That way if (hopefully) zig adds `count()` to its stdlib, it is trivial to remove the wrapper. (Similar to un-vendoring hashmap.)

matklad: Yeah, that's a design question! My thinking here is that we don't actually want to remove this wrapper at least until Zig std is stable --- wrapper will help us with, eg, Zig version upgrades, by decoupling changes in the std from changes in our code (eg, if a method is renamed in a new version of Zig, its enough to rename just the wrapper, not every usage). The overall guideline I have in mind here is:

- we wrap std types to fix API errors, or otherwise tweak them for our
  particular usage
- we do this lazily, only when there's an actual problem
- we do this minimally (wrap API rather than vendor the whole copy)
- we _might_ unfork once Zig stdlib is stable (and switches to
  snake_case :-))

sentientwaffle: > - we do this lazily, only when there's an actual problem
> - we do this minimally (wrap API rather than vendor the whole copy)

100% agree! That's why I suggested mirroring the underlying function names. :wink: 
I don't feel strongly about it though.

sentientwaffle (approved):

PR #1123, opened 2023-08-15
Explicitly fail the pipeline if a dependency fails

cb22:


matklad (dismissed):
:see_no_evil:

eatonphil:
I guess I missed a conversation. Why are we commenting out most of linux.yml? When will this setExecCmd be fixed?

cb22:
Ah sorry - we're commenting out the kcov coverage in CI for now; it seems `setExecCmd` is broken, so we can't run our compiled test binary easily under `kcov` for now (I should probably dig into it a bit more, but tried executing it unconditionally, with a broken cmd, etc and there was no change)

eatonphil:
Is it worth it to only disable the coverage checks within the test_on_bunutu step or it doesn't matter for now since this is only one of these sorts of tests anyway?

matklad:
>I should probably dig into it a bit more,

5s glance at Zig sources tells me that it's probably fine, and that the problem is maybe elsewhere...

cb22:
I don't quite follow?

matklad:
I mean, nothing seems obviously broken with `setExecCmd` internally, so I have a high prior that it works, and something's broken elsewhere.

cb22:
> I mean, nothing seems obviously broken with setExecCmd internally, so I have a high prior that it works, and something's broken elsewhere.

Oh oops, my comment was @eatonphil - I didn't see yours before replying!

eatonphil:
Nevermind. :)

eatonphil (approved):

eatonphil:
Maybe open an issue to add back code coverage?

PR #1124, opened 2023-08-15
docs: fix spellcheck error

matklad:


matklad:
Should fix https://github.com/tigerbeetle/tigerbeetle/actions/runs/5866652093/job/15905865077#step:3:269

cb22 (approved):
Yep, CI for that step is now good!

PR #1125, opened 2023-08-15
java: fix tests for Zig 0.11

matklad:


matklad:
java_docs fail with

>run ./zig-out/bin/client_docs: error: unable to spawn ./zig-out/bin/client_docs: FileNotFound

but that's an unrelated race in now parallel build.zig

cb22:
I see

```
error while loading shared libraries: libjvm.so: cannot open shared object file: No such file or directory
``` 

now? - https://github.com/tigerbeetle/tigerbeetle/actions/runs/5867558463/job/15908487703?pr=1125

matklad:
Oh wow, I assumed that that's my local NixOS problems! Will look into that! But we've stepped further! Now we get a linking error, rather than a compilation error!

batiati:
The path is set here, depending on the JAVA_HOME environment variable:
https://github.com/tigerbeetle/tigerbeetle/blob/70fcb51c40e4ef259b7f51f83e88baec803c5922/build.zig#L1119-L1124

I guess this custom build step is not setting the lib path as expected, and maybe it needs to be refactored for Zig 0.11.0.

matklad:
Yeah, I am 0.9 sure is that it’s just karma catching up with us for setting paths in execute phase :)

batiati:
> Yeah, I am 0.9 sure is that it’s just karma catching up with us for setting paths in execute phase :)

More karma-context here: https://github.com/tigerbeetle/tigerbeetle/pull/957#discussion_r1255740976

matklad:
https://github.com/tigerbeetle/tigerbeetle/pull/1128 fixes the unrelated jni failure

matklad:
Actually, probably better to add this here, because, although the fixes are orthogonal, you can't actually check one without the other.

cb22 (approved):

PR #1126, opened 2023-08-15
build: fix race codnition in maybe_execute

matklad:
We need happens-before between running a binary and building it. This was a latent bug, but in Zig 0.11 we hit it due to parallel build.


cb22 (approved):

eatonphil:
> matklad/off-to-the-races

Award for best branch names in general goes to matklad

PR #1128, opened 2023-08-15
build: fix jni tests for Zig 0.11

matklad:


build.zig
```diff
@@ -353,8 +353,63 @@ pub fn build(b: *std.Build) void {
...
+                else {
+                    std.log.err("{s}", .{ldd_result});
+                    return error.JavaAbiUnrecognized;
+                };
+            }
+
+            switch (builtin.os.tag) {
+                .windows => set_windows_dll(b.allocator, java_home),
+                .macos => try b.env_map.put("DYLD_LIBRARY_PATH", libjvm_path),
+                .linux => try b.env_map.put("LD_LIBRARY_PATH", libjvm_path),
```
matklad, line 400: NB: `LD_LIBRARY_PATH` on linux is new, but I think we need that, as JAVA_HOME in general might not be on the runtime search path.

matklad:
let's do this in https://github.com/tigerbeetle/tigerbeetle/pull/1125 rather.

PR #1138, opened 2023-08-17
Fix tigerbeetle version, correct zero check in src/shell.zig

### LINKED ISSUE #1097: Version isn't being correctly set for `tigerbeetle version`
See https://github.com/tigerbeetle/tigerbeetle/pull/1043.

eatonphil:
In addition to the zero check being backwards, we weren't propagating .git into the build environment in a few places so the version wouldn't have been set correctly in Docker for example. It may never have been.

Closes https://github.com/tigerbeetle/tigerbeetle/issues/1097

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -20,12 +20,8 @@ pub fn build(b: *std.Build) !void {
     defer shell.destroy();
 
     // The "tigerbeetle version" command includes the build-time commit hash.
-    options.addOption(
-        ?[]const u8,
-        "git_commit",
-        if (shell.git_commit()) |commit| @as([]const u8, &commit) else |_| null,
-    );
-    options.addOption(?[]const u8, "git_tag", shell.git_tag() catch null);
+    options.addOption([]const u8, "git_commit", try shell.git_commit());
+    options.addOption([]const u8, "git_tag", try shell.git_tag());
 
     options.addOption(
         config.ConfigBase,
--- a/src/shell.zig
+++ b/src/shell.zig
@@ -192,7 +192,17 @@ pub fn exec_stdout(shell: *Shell, comptime cmd: []const u8, cmd_args: anytype) !
     errdefer shell.arena.allocator().free(res.stdout);
 
     switch (res.term) {
-        .Exited => |code| if (code == 0) return error.NonZeroExitStatus,
+        .Exited => |code| if (code != 0) {
+            std.debug.print(
+                \\Debugging failure (stdout): {s}
+                \\Debugging failure (stderr): {s}
+                \\
+            , .{
+                res.stdout,
+                res.stderr,
+            });
+            return error.NonZeroExitStatus;
+        },
         else => return error.CommandFailed,
     }
 
@@ -201,16 +211,17 @@ pub fn exec_stdout(shell: *Shell, comptime cmd: []const u8, cmd_args: anytype) !
 }
 
 /// Returns current git commit hash as an ASCII string.
-pub fn git_commit(shell: *Shell) ![40]u8 {
+pub fn git_commit(shell: *Shell) ![]const u8 {
     const stdout = try shell.exec_stdout("git rev-parse --verify HEAD", .{});
     // +1 for trailing newline.
     if (stdout.len != 40 + 1) return error.InvalidCommitFormat;
-    return stdout[0..40].*;
+    return stdout[0..40];
 }
 
 /// Returns current git tag.
 pub fn git_tag(shell: *Shell) ![]const u8 {
-    const stdout = try shell.exec_stdout("git describe --tags", .{});
+    // --always is necessary in cases where we haven't yet `git fetch`-ed.
+    const stdout = try shell.exec_stdout("git describe --tags --always", .{});
     return stdout;
 }
 
--- a/.github/ci/docs_check.sh
+++ b/.github/ci/docs_check.sh
@@ -5,6 +5,9 @@ set -eu
 docker run -v "$(pwd)":/build:ro -w /tmp node:19 bash -c '
 set -eux
 
+# Deals with an error git throws within Docker when a git repo is
+# volume mounted.:
+#   fatal: detected dubious ownership in repository at '/wrk'
 git config --global --add safe.directory /build
 
 # Using a fork of this validate-links check to include checks on
--- a/.github/ci/tests_on_alpine.sh
+++ b/.github/ci/tests_on_alpine.sh
@@ -4,6 +4,13 @@ set -e
 
 docker run --entrypoint sh -v "$(pwd)":/wrk -w /wrk alpine -c "
 set -e
+apk add -U git
+
+# Deals with an error git throws within Docker when a git repo is
+# volume mounted.:
+#   fatal: detected dubious ownership in repository at '/wrk'
+git config --global --add safe.directory /wrk
+
 ./scripts/install_zig.sh
 ./zig/zig build test
 "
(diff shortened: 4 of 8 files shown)


src/shell.zig
```diff
@@ -192,7 +192,7 @@ pub fn exec_stdout(shell: *Shell, comptime cmd: []const u8, cmd_args: anytype) !
     errdefer shell.arena.allocator().free(res.stdout);
 
     switch (res.term) {
-        .Exited => |code| if (code == 0) return error.NonZeroExitStatus,
+        .Exited => |code| if (code != 0) return error.NonZeroExitStatus,
```
eatonphil: @matklad this is the meat of the problem. :)

src/shell.zig
```diff
@@ -201,11 +201,11 @@ pub fn exec_stdout(shell: *Shell, comptime cmd: []const u8, cmd_args: anytype) !
 }
 
 /// Returns current git commit hash as an ASCII string.
-pub fn git_commit(shell: *Shell) ![40]u8 {
+pub fn git_commit(shell: *Shell) ![]const u8 {
```
eatonphil, line 214: I'm not sure if this is important. Returning a slice made build.zig simpler.

build.zig
```diff
@@ -20,12 +20,8 @@ pub fn build(b: *std.Build) !void {
...
 
     // The "tigerbeetle version" command includes the build-time commit hash.
-    options.addOption(
-        ?[]const u8,
-        "git_commit",
-        if (shell.git_commit()) |commit| @as([]const u8, &commit) else |_| null,
-    );
-    options.addOption(?[]const u8, "git_tag", shell.git_tag() catch null);
+    options.addOption(?[]const u8, "git_commit", try shell.git_commit());
+    options.addOption([]const u8, "git_tag", try shell.git_tag());
```
eatonphil, line 24: Switch these to `try` so they don't hide the error message on failure.

src/shell.zig
eatonphil: Oh, you already knew that. :)

src/shell.zig
```diff
@@ -201,16 +211,18 @@ pub fn exec_stdout(shell: *Shell, comptime cmd: []const u8, cmd_args: anytype) !
...
+    try shell.exec("git fetch --tags", .{});
```
matklad: fetching tags from the network is a slow, failable operation. Given that we run this on every build, it seems better to avoid maybe ask the caller to fetch the repo with tags instead?

build.zig
matklad: Yeah, I think it's OK to assume that we have git during building!

src/shell.zig
eatonphil: > Given that we run this on every build

Yes, however, locally it's going to rarely do anything. Do you still think we should drop it?

```
$ time git fetch --tags
remote: Enumerating objects: 16, done.
remote: Counting objects: 100% (16/16), done.
remote: Compressing objects: 100% (6/6), done.
remote: Total 16 (delta 10), reused 13 (delta 10), pack-reused 0
Unpacking objects: 100% (16/16), 5.68 KiB | 528.00 KiB/s, done.
From github.com:tigerbeetledb/tigerbeetle
 + 77184c64...b87c81c8 batiati-eytzinger                                                       -> origin/batiati-eytzinger  (forced update)
 * [new branch]        gh-readonly-queue/main/pr-1140-6378bbe359c0ae2721344ac867ac521f2e5db66a -> origin/gh-readonly-queue/main/pr-1140-6378bbe359c0ae2721344ac867ac521f2e5db66a
 + 80a01ed5...16fd8150 matklad/alligned-free                                                   -> origin/matklad/alligned-free  (forced update)
 + c5ec87c0...8f9bbff5 matklad/tmp-tigerbeetle                                                 -> origin/matklad/tmp-tigerbeetle  (forced update)
git fetch --tags  0.08s user 0.05s system 18% cpu 0.688 total
$ time git fetch --tags
git fetch --tags  0.07s user 0.04s system 18% cpu 0.620 total
$ time git fetch --tags
git fetch --tags  0.07s user 0.04s system 18% cpu 0.593 total
$ time git fetch --tags
git fetch --tags  0.07s user 0.04s system 18% cpu 0.572 total
```

matklad: It's going to ping GitHub server every time, and, depending on where you are, that might be quite slow: 

```
λ n 3 t git fetch --tags
Run 1

real 2.13s
cpu  196.23ms (140.76ms user + 55.47ms sys)
rss  96.70mb
Run 2

real 2.25s
cpu  196.53ms (156.44ms user + 40.09ms sys)
rss  96.45mb
Run 3

real 2.16s
cpu  195.90ms (147.55ms user + 48.35ms sys)
rss  96.89mb
```

eatonphil: Ok! I'll move it into the ci steps then.

matklad: I guess there's also a considiration that we want `zig build` to work if there's no internet connection.

eatonphil: That will depend on how you cloned it then. Since although we clone in GitHub actions, it doesn't have any tags and got describe --tags fails, so the build fails.

However, if we drop the hard check then it can be easy to reintroduce this bug of TigerBeetle version not being set correctly. Unless we add new tests for after a release.

matklad: FWIW, that's what I use to fetch tags to gh builders:

https://github.com/matklad/once_cell/blob/3cd65494663f003b9b2e0be5b37b3426e543f72a/.github/workflows/ci.yaml#L22-L23

Maybe there's a better way, I don't remeber how I came up with that snippet

eatonphil: Ok!

eatonphil: Well now it started succeeding without me needing to fetch anything. Wonderful! That means it may fail sometime in the future. But I'm not sure which steps would fail.

eatonphil: Nevermind! It's `git tag --describe --always` that handles it without needing to fetch. We're all set.

build.zig
```diff
@@ -20,12 +20,8 @@ pub fn build(b: *std.Build) !void {
     defer shell.destroy();
 
     // The "tigerbeetle version" command includes the build-time commit hash.
-    options.addOption(
-        ?[]const u8,
-        "git_commit",
-        if (shell.git_commit()) |commit| @as([]const u8, &commit) else |_| null,
-    );
-    options.addOption(?[]const u8, "git_tag", shell.git_tag() catch null);
+    options.addOption(?[]const u8, "git_commit", try shell.git_commit());
```
matklad:
```suggestion
    options.addOption([]const u8, "git_commit", try shell.git_commit());
```

src/shell.zig
```diff
@@ -192,7 +192,17 @@ pub fn exec_stdout(shell: *Shell, comptime cmd: []const u8, cmd_args: anytype) !
     errdefer shell.arena.allocator().free(res.stdout);
 
     switch (res.term) {
-        .Exited => |code| if (code == 0) return error.NonZeroExitStatus,
+        .Exited => |code| if (code != 0) {
+            std.debug.print(
+                \\Debugging failure (stdout): {s}
+                \\Debugging failure (stderr): {s}
```
matklad, line 198: :heart:

matklad (dismissed):

build.zig
eatonphil: Accepted locally along with a related change needed in src/tigerbeetle/main.zig. Thanks!

matklad (approved):

PR #1139, opened 2023-08-17
VSR: Extract Op helpers; rework "checkpoint trigger" terminology

sentientwaffle:
#### s/op_checkpoint_trigger/op_checkpoint_next_trigger

Clarify checkpoint terminology:
- No need to refer to a checkpoint's "previous" or "next" (past/future) trigger.
- Instead, every checkpoint has exactly one trigger (except checkpoint=0, which has no trigger).
- This corresponds to the "past" trigger. To refer to the "next" trigger, compute the trigger of
the next checkpoint.

#### Extract Op math helpers

`checkpoint_before_checkpoint()` isn't used yet except for validation, but it will be used as part of table-sync.

### DIFF
--- a/src/vsr.zig
+++ b/src/vsr.zig
@@ -1918,3 +1918,52 @@ const ViewChangeHeadersArray = struct {
         headers.array.appendAssumeCapacity(Headers.dvc_blank(op));
     }
 };
+
+pub const Op = struct {
+    pub fn checkpoint_before_checkpoint(checkpoint: u64) ?u64 {
+        assert(checkpoint_valid(checkpoint));
+
+        if (checkpoint == 0) {
+            return null;
+        }
+
+        if (checkpoint == constants.journal_slot_count - constants.lsm_batch_multiple - 1) {
+            return 0;
+        }
+
+        return checkpoint - (constants.journal_slot_count - constants.lsm_batch_multiple);
+    }
+
+    pub fn checkpoint_after_checkpoint(checkpoint: u64) u64 {
+        assert(checkpoint_valid(checkpoint));
+
+        const checkpoint_next = op: {
+            if (checkpoint == 0) {
+                // First wrap: op_checkpoint_next = 8-2-1 = 5
+                break :op constants.journal_slot_count - constants.lsm_batch_multiple - 1;
+            } else {
+                // Second wrap: op_checkpoint_next = 5+8-2 = 11
+                // Third wrap: op_checkpoint_next = 11+8-2 = 17
+                break :op checkpoint + constants.journal_slot_count - constants.lsm_batch_multiple;
+            }
+        };
+
+        assert((checkpoint_next + 1) % constants.lsm_batch_multiple == 0);
+        assert(checkpoint_valid(checkpoint_next));
+        assert(checkpoint_before_checkpoint(checkpoint_next) == checkpoint);
+
+        return checkpoint_next;
+    }
+
+    pub fn trigger_from_checkpoint(checkpoint_nonzero: u64) u64 {
+        assert(checkpoint_nonzero != 0);
+        assert(checkpoint_valid(checkpoint_nonzero));
+
+        return checkpoint_nonzero + constants.lsm_batch_multiple;
+    }
+
+    pub fn checkpoint_valid(op: u64) bool {
+        return op == 0 or
+            (op + 1) % (constants.journal_slot_count - constants.lsm_batch_multiple) == 0;
+    }
+};
--- a/src/vsr.zig
+++ b/src/vsr.zig
@@ -170,19 +170,19 @@ pub const Operation = enum(u8) {
     _,
 
     pub fn from(comptime StateMachine: type, op: StateMachine.Operation) Operation {
-        check_state_machine_operations(StateMachine.Operation);
+        check_state_machine_operations(StateMachine);
         return @as(Operation, @enumFromInt(@intFromEnum(op)));
     }
 
     pub fn cast(self: Operation, comptime StateMachine: type) StateMachine.Operation {
-        check_state_machine_operations(StateMachine.Operation);
+        check_state_machine_operations(StateMachine);
         assert(self.valid(StateMachine));
         assert(!self.vsr_reserved());
         return @as(StateMachine.Operation, @enumFromInt(@intFromEnum(self)));
     }
 
     pub fn valid(self: Operation, comptime StateMachine: type) bool {
-        check_state_machine_operations(StateMachine.Operation);
+        check_state_machine_operations(StateMachine);
 
         inline for (.{ Operation, StateMachine.Operation }) |Enum| {
             const ops = comptime std.enums.values(Enum);
@@ -213,12 +213,13 @@ pub const Operation = enum(u8) {
         unreachable;
     }
 
-    fn check_state_machine_operations(comptime Op: type) void {
+    fn check_state_machine_operations(comptime StateMachine: type) void {
         comptime {
-            assert(@typeInfo(Op).Enum.is_exhaustive);
-            assert(@typeInfo(Op).Enum.tag_type == @typeInfo(Operation).Enum.tag_type);
-            for (@typeInfo(Op).Enum.fields) |field| {
-                const op = @field(Op, field.name);
+            assert(@typeInfo(StateMachine.Operation).Enum.is_exhaustive);
+            assert(@typeInfo(StateMachine.Operation).Enum.tag_type ==
+                @typeInfo(Operation).Enum.tag_type);
+            for (@typeInfo(StateMachine.Operation).Enum.fields) |field| {
+                const op = @field(StateMachine.Operation, field.name);
                 if (@intFromEnum(op) < constants.vsr_operations_reserved) {
                     @compileError("StateMachine.Operation is reserved");
                 }
--- a/src/vsr.zig
+++ b/src/vsr.zig
@@ -1920,7 +1920,7 @@ const ViewChangeHeadersArray = struct {
 };
 
 pub const Op = struct {
-    pub fn checkpoint_before_checkpoint(checkpoint: u64) ?u64 {
+    pub fn checkpoint_before(checkpoint: u64) ?u64 {
         assert(checkpoint_valid(checkpoint));
 
         if (checkpoint == 0) {
@@ -1934,7 +1934,7 @@ pub const Op = struct {
         return checkpoint - (constants.journal_slot_count - constants.lsm_batch_multiple);
     }
 
-    pub fn checkpoint_after_checkpoint(checkpoint: u64) u64 {
+    pub fn checkpoint_after(checkpoint: u64) u64 {
         assert(checkpoint_valid(checkpoint));
 
         const checkpoint_next = op: {
@@ -1950,7 +1950,7 @@ pub const Op = struct {
 
         assert((checkpoint_next + 1) % constants.lsm_batch_multiple == 0);
         assert(checkpoint_valid(checkpoint_next));
-        assert(checkpoint_before_checkpoint(checkpoint_next) == checkpoint);
+        assert(checkpoint_before(checkpoint_next) == checkpoint);
 
         return checkpoint_next;
     }
--- a/src/vsr.zig
+++ b/src/vsr.zig
@@ -1919,25 +1919,46 @@ const ViewChangeHeadersArray = struct {
     }
 };
 
-pub const Op = struct {
+/// For a replica with journal_slot_count=8 and lsm_batch_multiple=2:
+///
+///   checkpoint() call           0   1   2   3
+///   op_checkpoint               0   5  11  17
+///   op_checkpoint_next          5  11  17  23
+///   op_checkpoint_next_trigger  7  13  19  25
+///
+///     commit log (ops)           │ write-ahead log (slots)
+///     0   4   8   2   6   0   4  │ 0---4---
+///   0 ─────✓·%                   │ 01234✓6%   initial log fill
+///   1 ───────────✓·%             │ 890✓2%45   first wrap of log
+///   2 ─────────────────✓·%       │ 6✓8%0123   second wrap of log
+///   3 ───────────────────────✓·% │ 4%67890✓   third wrap of log
+///
+/// Legend:
+///
+///   ─/✓  op on disk at checkpoint
+///   ·/%  op in memory at checkpoint
+///     ✓  op_checkpoint
+///     %  op_checkpoint's trigger
+///
+pub const Checkpoint = struct {
+    comptime {
+        assert(constants.journal_slot_count > constants.lsm_batch_multiple);
+        assert(constants.journal_slot_count % constants.lsm_batch_multiple == 0);
+    }
+
     pub fn checkpoint_before(checkpoint: u64) ?u64 {
-        assert(checkpoint_valid(checkpoint));
+        assert(valid(checkpoint));
 
-        if (checkpoint == 0) {
-            return null;
-        }
-
-        if (checkpoint == constants.journal_slot_count - constants.lsm_batch_multiple - 1) {
-            return 0;
-        }
+        if (checkpoint == 0) return null;
+        if (checkpoint == constants.journal_slot_count - constants.lsm_batch_multiple - 1) return 0;
 
         return checkpoint - (constants.journal_slot_count - constants.lsm_batch_multiple);
     }
 
     pub fn checkpoint_after(checkpoint: u64) u64 {
-        assert(checkpoint_valid(checkpoint));
+        assert(valid(checkpoint));
 
-        const checkpoint_next = op: {
+        const result = op: {
             if (checkpoint == 0) {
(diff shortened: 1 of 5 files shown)


src/vsr.zig
```diff
@@ -170,19 +170,19 @@ pub const Operation = enum(u8) {
     _,
 
     pub fn from(comptime StateMachine: type, op: StateMachine.Operation) Operation {
-        check_state_machine_operations(StateMachine.Operation);
+        check_state_machine_operations(StateMachine);
```
jorangreef, line 173: Nice change.

src/vsr.zig
```diff
@@ -1917,3 +1918,52 @@ const ViewChangeHeadersArray = struct {
         headers.array.appendAssumeCapacity(Headers.dvc_blank(op));
     }
 };
+
+pub const Op = struct {
```
jorangreef:
```suggestion
pub const Checkpoint = struct {
```

Since all the method names in here include `checkpoint`, and since checkpoints or checkpoint calculations are what this is really about.

To riff on Brian Eno's "Honor your mistake as a hidden intention" (e.g. when @kprotty kept saying "groove" we just had fun and together decided to rename `Grove` to `Groove`), this one could be an instance of "Honor your repetition as a namespace suggestion". 😛

src/vsr.zig
```diff
@@ -1917,3 +1918,52 @@ const ViewChangeHeadersArray = struct {
...
+    pub fn checkpoint_before(checkpoint: u64) ?u64 {
+        assert(checkpoint_valid(checkpoint));
+
+        if (checkpoint == 0) {
+            return null;
+        }
```
jorangreef:
```suggestion
        if (checkpoint == 0) return null;
```

If it fits on a single line, it's okay to do a one liner return. I know the kernel doesn't like this, but it's not something we enforce because it can make for leaner code, according to context (for example, in the state machine).

src/vsr.zig
```diff
@@ -1917,3 +1918,52 @@ const ViewChangeHeadersArray = struct {
         headers.array.appendAssumeCapacity(Headers.dvc_blank(op));
     }
 };
+
+pub const Op = struct {
+    pub fn checkpoint_before(checkpoint: u64) ?u64 {
+        assert(checkpoint_valid(checkpoint));
```
jorangreef: If the namespace becomes `Checkpoint` then we can also use only `assert(valid(checkpoint))` here, which will be nice.

src/vsr.zig
```diff
@@ -1917,3 +1918,52 @@ const ViewChangeHeadersArray = struct {
...
+
+        if (checkpoint == 0) {
+            return null;
+        }
+
+        if (checkpoint == constants.journal_slot_count - constants.lsm_batch_multiple - 1) {
```
jorangreef: What are our local expectations here w.r.t. the relation between `journal_slot_count` and `lsm_batch_multiple`?

We should assert them here, regardless of where else they are asserted, as defense in depth.

jorangreef: If you think it's necessary, we could comment to the effect that `checkpoint` is an index, whereas slot count and multiple are counts, therefore `- 1` to convert count to index.

jorangreef: Since this distance is also shared by both `before` and `after` (we do the same subtraction in a few places), we could use a named local constant instead of the magic calculation.

jorangreef: Then it could become `+ ops_between_checkpoints` or `- ops_between_checkpoints`.

src/vsr.zig
```diff
@@ -1917,3 +1918,52 @@ const ViewChangeHeadersArray = struct {
...
+                // Third wrap: op_checkpoint_next = 11+8-2 = 17
+                break :op checkpoint + constants.journal_slot_count - constants.lsm_batch_multiple;
+            }
+        };
+
+        assert((checkpoint_next + 1) % constants.lsm_batch_multiple == 0);
+        assert(checkpoint_valid(checkpoint_next));
+        assert(checkpoint_before(checkpoint_next) == checkpoint);
+
+        return checkpoint_next;
```
jorangreef:
```suggestion
        return result;
```

`checkpoint_after` and `checkpoint_next` are different only because of shadowing. Therefore, use a simpler local variable like `result` or `op`.

src/vsr.zig
```diff
@@ -1917,3 +1918,52 @@ const ViewChangeHeadersArray = struct {
...
+    }
+
+    pub fn trigger_from_checkpoint(checkpoint_nonzero: u64) u64 {
```
jorangreef: It feels like this should have an optional return type, following `checkpoint_before()`.

jorangreef: Then it can accept `checkpoint` instead of `checkpoint_nonzero`.

src/vsr.zig
```diff
@@ -1917,3 +1918,52 @@ const ViewChangeHeadersArray = struct {
...
+        assert(checkpoint_nonzero != 0);
+        assert(checkpoint_valid(checkpoint_nonzero));
+
+        return checkpoint_nonzero + constants.lsm_batch_multiple;
+    }
+
+    pub fn checkpoint_valid(op: u64) bool {
+        return op == 0 or
+            (op + 1) % (constants.journal_slot_count - constants.lsm_batch_multiple) == 0;
```
jorangreef, line 1991: Do we have an off-by-one here?

It feels like where we modulo an index by a count, we don't actually want to convert the index to a count?

src/vsr.zig
sentientwaffle: They are all checkpoint functions right now, but this namespace will also shortly (as part of table sync) get `snapshot_from_op`, which doesn't have anything to do with checkpoints.

src/vsr.zig
sentientwaffle: I'd prefer to leave that constant inlined, since "ops_between_checkpoints" is misleading -- it is really "ops between checkpoints except for between the origin and next checkpoint". (The transition from "checkpoint=0" to its next checkpoint is a special case.)

src/vsr.zig
sentientwaffle: This "+1" is a shift to the right, to counteract the "-1" applied to the first wrap's checkpoint.

src/vsr.zig
jorangreef: I do still much prefer `vsr.Checkpoint` over `Op` to keep these checkpoint-related functions together, and would then rather put op-related helpers in another namespace.

Also, do we need `snapshot_from_op`, or can we do without it? The `snapshot` should be the same as the `op` number for a simple mental model (we did it like this at the time for this reason).

src/vsr.zig
jorangreef: It does seem that the subtraction is performed in enough places that it could be factored out, perhaps with a better name.

src/vsr.zig
jorangreef: Could you help me understand this better? I don't follow.

src/vsr.zig
sentientwaffle: Ops are snapshots. But there are different ways they can correspond. Perhaps `snapshot_from_commit` would be a better name.

```zig
    /// A table with TableInfo.snapshot_min=S was written during some commit with op<S.
    /// A block with snapshot_min=S is definitely readable at op=S.
    pub fn snapshot_from_commit(op: u64) u64 {
        return op + 1;
    }
```

Roughly, "block.snapshot=X" is "X is the first op that can read `block`".

> Also, do we need snapshot_from_op, or can we do without it? 

We could inline that, but I think the label describing what is happening is useful for readability, even though the implementation is trivial.

src/vsr.zig
sentientwaffle: In this diagram, `✓` are each checkpoints. Each is one minus a multiple of `constants.journal_slot_count - constants.lsm_batch_multiple`.

```zig
        /// For a replica with journal_slot_count=8 and lsm_batch_multiple=2:
...
        ///     commit log (ops)           │ write-ahead log (slots)
        ///     0   4   8   2   6   0   4  │ 0---4---
        ///   0 ─────✓·%                   │ 01234✓6%   initial log fill
        ///   1 ───────────✓·%             │ 890✓2%45   first wrap of log
        ///   2 ─────────────────✓·%       │ 6✓8%0123   second wrap of log
        ///   3 ───────────────────────✓·% │ 4%67890✓   third wrap of log
        ///
        /// Legend:
        ///
        ///   ─/✓  op on disk at checkpoint
        ///   ·/%  op in memory at checkpoint
        ///     ✓  op_checkpoint
        ///     %  op_checkpoint's trigger
```

The `+ 1` in `(op + 1) % (constants.journal_slot_count - constants.lsm_batch_multiple)` cancels out the `- 1` that is part of the first wrap's checkpoint (7 in the diagram):

```
    pub fn checkpoint_after_checkpoint(checkpoint: u64) u64 {
        assert(checkpoint_valid(checkpoint));

        const checkpoint_next = op: {
            if (checkpoint == 0) {
                // First wrap: op_checkpoint_next = 8-2-1 = 5
                break :op constants.journal_slot_count - constants.lsm_batch_multiple - 1;
            } else {
```

Since every successive checkpoint after the first wrap adds `constants.journal_slot_count - constants.lsm_batch_multiple`, they all have that `-1` in them also.

src/vsr.zig
jorangreef: To clarify if we are on the same page (I think we are)... the intended sense of `table.snapshot_min` should be "the minimum op number that can see the table", as your comment above the function has it.

However, I don't follow the `return op + 1` above. It should be an identity function rather.

sentientwaffle: For table/reply sync, we will store `commit_unsynced_min` and `commit_unsynced_max` in the superblock's VSRState:

```
        /// See `commit_unsynced_max`.
        commit_unsynced_min: u64,

        /// When zero, all of the grid and replies are synced.
        /// (When zero, `commit_unsynced_min` is also zero.)
        ///
        /// When nonzero, we must repair grid-blocks/client-replies that would have been written
        /// during the commits between `commit_unsynced_min` and `commit_unsynced_max` (inclusive).
        /// (They were not written normally because we skipped past them via state sync.)
        commit_unsynced_max: u64,
```

Then as we iterate the tables to figure out which need to be synced, and which we already have:

```
            assert(self.superblock.working.vsr_state.commit_unsynced_max > 0);

            const commit_unsynced_min = self.superblock.working.vsr_state.commit_unsynced_min;
            const commit_unsynced_max = self.superblock.working.vsr_state.commit_unsynced_max;
            while (self.sync_tables.?.next(&self.state_machine.forest)) |table_info| {
                if (table_info.snapshot_min >= vsr.Op.snapshot_from_commit(commit_unsynced_min) and
                    table_info.snapshot_min <= vsr.Op.snapshot_from_commit(commit_unsynced_max))
                {
                    // (sync the table)
                } else {
                    // (table is already synced)
                }
```

We could apply the `+1` when we set `commit_unsynced_min`/`max` (instead of in `snapshot_from_commit`), but I think that results in a less intuitive mental model of what those ops mean.

matklad: >to rename Grove to Groove)

:exploding_head: these are different words!

src/vsr/replica.zig
```diff
@@ -4943,47 +4945,39 @@ pub fn ReplicaType(
         ///   ─/✓  op on disk at checkpoint
         ///   ·/%  op in memory at checkpoint
         ///     ✓  op_checkpoint
-        ///     %  op_checkpoint_trigger
+        ///     %  op_checkpoint's trigger
         ///
```
matklad: Might move this infographic to the `const OP = `?

src/vsr/replica.zig
```diff
@@ -4943,47 +4945,39 @@ pub fn ReplicaType(
...
-        /// checkpointed.
+        /// Receiving and storing an op higher than `op_checkpoint_next_trigger()` is forbidden;
+        /// doing so would overwrite a message (or the slot of a message) that has not yet been
+        /// committed and checkpointed.
         ///
         /// See `op_checkpoint_next` for more detail.
-        fn op_checkpoint_trigger(self: *const Self) u64 {
-            return self.op_checkpoint_next() + constants.lsm_batch_multiple;
+        fn op_checkpoint_next_trigger(self: *const Self) u64 {
+            return vsr.Op.trigger_from_checkpoint(self.op_checkpoint_next()).?;
```
matklad: maybe `trigger_for_checkpoint`? `for` is more specific about the relationship between the two.

src/vsr.zig
jorangreef: Thanks for the explanation.

Stepping back from state sync, for snapshot usage in general (e.g. for read transactions), I don't think we should use a helper (or it should be an identity function), and the snapshot value should be the committed op at which the snapshot was acquired.

I would rather change the inclusive/exclusive semantics of the unsynced range, than change the identity function between op/snapshot.

matklad: I guess we could go full OOP here?

```
pub const Checkpoint = struct {
    op: u64,
    trigger: u64,

    fn next(self: Checkpoint) Checkpoint {

    }

    fn prev(self: Checkpoint) Checkpoint {

    }
};
```

The call site would be `replica.checkpoint().next().trigger`

sentientwaffle: > the snapshot value should be the committed op at which the snapshot was acquired.

I don't understand what you are suggesting. (We haven't actually implemented snapshot acquisition so maybe our mental models don't match up?)

Do you mean that, when we send a message to acquire a user snapshot, that the op on the reply message is our snapshot? (Which means that we would "backdate" the snapshots of writes of data that is currently in memory, I guess?)

> then change the identity function between op/snapshot.

These are all different ops:
1. when data arrives at a replica (the WAL + mutable table),
2. when data starts writing to disk (the first beat of a bar, but delayed by ~bar from arrival),
3. when data is definitely written to disk (the end of a bar, again delayed by ~bar from arrival) 

So wrt "the identity function between op/snapshot", which gets the identity function? Because it can be at most one of them :sweat_smile: 

(Right now it is 3, which is imo the most natural choice implementation-wise (which was my reason for choosing it).)

---

> I would rather change the inclusive/exclusive semantics of the unsynced range, then change the identity function between op/snapshot.

Fwiw, this is really the most natural semantics for the range, e.g. when determining which client replies to sync:

```zig
                    entry.header.op >= self.superblock.working.vsr_state.commit_unsynced_min and
                    entry.header.op <= self.superblock.working.vsr_state.commit_unsynced_max;
```

jorangreef: > I don't understand what you are suggesting. (We haven't actually implemented snapshot acquisition so maybe our mental models don't match up?)

Thanks. Yes, I'm referring to snapshots in the API when implemented. What `snapshot` will mean w.r.t. to the `op` that acquires it.

> So wrt "the identity function between op/snapshot", which gets the identity function?

I'm thinking not at the storage layer but at the state machine layer. For the op that acquires the snapshot, whatever the op can see (if it were executing say a read transaction) is what the snapshot should see when used in subsequent read operations that reference the snapshot.

By "identity function", I mean that if an op is to acquire a snapshot, then the value of the snapshot acquired is that op number.

My thinking has been that how we implement this under the hood will require pulsing, to get the snapshot acquiring op onto a bar boundary.

src/vsr/replica.zig
```diff
@@ -4923,67 +4925,37 @@ pub fn ReplicaType(
...
-        /// would overwrite a message (or the slot of a message) that has not yet been committed and
-        /// checkpointed.
+        /// Receiving and storing an op higher than `op_checkpoint_next_trigger()` is forbidden;
+        /// doing so would overwrite a message (or the slot of a message) that has not yet been
+        /// committed and checkpointed.
         ///
         /// See `op_checkpoint_next` for more detail.
-        fn op_checkpoint_trigger(self: *const Self) u64 {
-            return self.op_checkpoint_next() + constants.lsm_batch_multiple;
+        fn op_checkpoint_next_trigger(self: *const Self) u64 {
```
matklad, line 4960: This particular rename is golden! So much better now!

matklad (dismissed):

sentientwaffle:
(Rebased to resolve conflicts with `main`.)

src/vsr.zig
sentientwaffle: Ok, thanks for the clarification, I think I understand now!

---

Taking a step back: As the "user-visible" snapshot id, why not use (state machine) timestamps instead of op numbers?

- Op-numbers are currently "internal" to the vsr-replica/vsr-client world -- op-snapshots would expose them to users.
- Timestamps have more semantic meaning -- they are self-describing.
- We plan to have "scheduled future snapshots", which also play very well with this idea. You schedule the snapshot with a particular timestamp, and then if you query it before then it just says "scheduled but not queryable yet".
- We could map the snapshot-timestamps to snapshot-ops internally where we keep a record of persisted snapshots (in the superblock header iirc), so we can still use the ops to manage table visibility.

matklad (approved):

src/vsr.zig
jorangreef: That's a great idea.

As you say, physical timestamps, over logical op timestamps, also have the value of being auditable/relatable in the real world.

PR #1140, opened 2023-08-18
message_pool: fix wrong alignment in free

matklad:
We must pass the same alignment in free as the one used in AlignedAlloc. Removing the cast should fix the  issue

### DIFF
--- a/src/message_pool.zig
+++ b/src/message_pool.zig
@@ -123,7 +123,7 @@ pub const MessagePool = struct {
         var free_count: usize = 0;
         while (pool.free_list) |message| {
             pool.free_list = message.next;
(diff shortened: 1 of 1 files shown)


matklad:
wait a second, I think I've found another onew

matklad:
Actually, no, upon closer reading, we are all fine here, and the underlying Zig API is reliable, as long as you don't explictily cast the alingent away as we do in this single buggy case.

matklad:
Heh, look who messed up the alignment here in the first place!

https://github.com/tigerbeetle/tigerbeetle/commit/51436d87bc88f9f1554876027531d9dc61758d87

cb22 (approved):

PR #1141, opened 2023-08-18
Zig 0.11.0 misc

batiati:
- Remove the pattern `@ptrFromInt(@intFromPtr(...))` using `@constCast` or `@ptrCast`.

- Remove unnecessary `@as`, preferring explicit types declarations instead.

Note: This is not an exhaustive `@as` replacement, in some places it may still need future work.
Only occurrences related to `@ptrCast` or `@ptrFromInt` were analyzed.

### DIFF
--- a/src/benchmark.zig
+++ b/src/benchmark.zig
@@ -435,7 +435,7 @@ const Benchmark = struct {
         );
 
         b.client.request(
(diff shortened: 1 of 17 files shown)


matklad (approved):
:+1:

PR #1142, opened 2023-08-18
Explicit check for prefix characters in release to use in Linux version check

### LINKED ISSUE #1131: TigerBeetle crashes on startup on Fedora while trying to parse Fedora kernel release name
On my Fedora machine I get:

```
Running: ./tigerbeetle format --cluster=0 --replica=0 --replica-count=1 /home/phil/tb/tigerbeetle/zig-cache/tmp/PbRC13ZcO2qopxVz/0_0.tigerbeetle
info(io): creating "0_0.tigerbeetle"...                                                                                                   info(io): allocating 1.13494873046875GiB...
error: InvalidVersion
/home/phil/tb/tigerbeetle/zig/lib/std/SemanticVersion.zig:117:73: 0x30d6ec in parse (tigerbeetle)
            for (id) |c| if (!std.ascii.isAlphanumeric(c) and c != '-') return error.InvalidVersion;
                                                                        ^
/home/phil/tb/tigerbeetle/src/io/linux.zig:35:25: 0x310029 in init (tigerbeetle)
        const version = try std.SemanticVersion.parse(release);
                        ^
/home/phil/tb/tigerbeetle/src/tigerbeetle/main.zig:90:22: 0x3111d5 in init (tigerbeetle)
        command.io = try IO.init(128, 0);
                     ^
/home/phil/tb/tigerbeetle/src/tigerbeetle/main.zig:110:9: 0x317c3f in format (tigerbeetle)
        try command.init(allocator, path, true);
        ^
/home/phil/tb/tigerbeetle/src/tigerbeetle/main.zig:52:28: 0x364e01 in main (tigerbeetle)
        .format => |*args| try Command.format(allocator, .{
```

Here's my kernel name:

```
$ uname --kernel-release
6.3.8-100.fc37.x86_64
```

eatonphil:
Closes #1131

### DIFF
--- a/src/io/linux.zig
+++ b/src/io/linux.zig
@@ -11,6 +11,7 @@ const tracer = @import("../tracer.zig");
 const constants = @import("../constants.zig");
 const FIFO = @import("../fifo.zig").FIFO;
 const buffer_limit = @import("../io.zig").buffer_limit;
+const parse_dirty_semver = @import("../stdx.zig").parse_dirty_semver;
 
 pub const IO = struct {
     ring: IO_Uring,
@@ -31,8 +32,7 @@ pub const IO = struct {
     pub fn init(entries: u12, flags: u32) !IO {
         // Detect the linux version to ensure that we support all io_uring ops used.
         const uts = std.os.uname();
-        const release = std.mem.sliceTo(&uts.release, 0);
-        const version = try std.SemanticVersion.parse(release);
+        const version = try parse_dirty_semver(&uts.release);
         if (version.order(std.SemanticVersion{ .major = 5, .minor = 5, .patch = 0 }) == .lt) {
             @panic("Linux kernel 5.5 or greater is required for io_uring OP_ACCEPT");
         }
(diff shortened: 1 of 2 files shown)


matklad:
I _think_ @kprotty would be a better reviewer here --- what we do here with uname feels shady, and indeed something like feature detection might be better, but I don't how pedantically correct io_uring support detection works.

src/io/linux.zig
```diff
@@ -31,7 +31,24 @@ pub const IO = struct {
...
+            // std.SemanticVersion requires there be no extra
+            // characters after the major/minor/patch numbers. But
+            // some distributions have extra characters, such as this
+            // Fedora one: 6.3.8-100.fc37.x86_64.
+            var last_valid_version_character_index: usize = 0;
+            for (uts.release) |c| {
+                if (c == '.' or (c >= '0' and c <= '9')) {
+                    last_valid_version_character_index += 1;
+                    continue;
+                }
```
kprotty: Believe this should handle the case of more than 3 dots (i.e. custom kernel) by ignoring what's after as `5.15.90.1-microsoft-standard-WSL2` fails `std.SemanticVersion.parse`.

eatonphil: Gotcha!

eatonphil: Moved this into stdx.zig so I could add some unit tests.

kprotty (approved):

PR #1144, opened 2023-08-21
Prefetch binary search

batiati:
Implements `@prefetch` to binary search accordingly to https://arxiv.org/abs/1509.05053.

No divergences from the original paper, except that in our case a `Value` can be larger than a single cache line, and we need to prefetch multiple addresses for a single value.

Prefetch is opt-out (e.g. `.{ .prefetch = false }`).
Benchmark `./zig/zig build benchmark_eytzinger` compares both versions.

### DIFF
--- a/src/lsm/binary_search.zig
+++ b/src/lsm/binary_search.zig
@@ -2,14 +2,15 @@ const std = @import("std");
 const assert = std.debug.assert;
 const math = std.math;
 
+const stdx = @import("../stdx.zig");
+
 const constants = @import("../constants.zig");
 
 pub const Config = struct {
     mode: enum { lower_bound, upper_bound } = .lower_bound,
+    prefetch: bool = true,
 };
 
-// TODO Add prefeching when @prefetch is available: https://github.com/ziglang/zig/issues/3600.
-//
 // TODO The Zig self hosted compiler will implement inlining itself before passing the IR to llvm,
 // which should eliminate the current poor codegen of key_from_value/compare_keys.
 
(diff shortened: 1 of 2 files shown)


src/lsm/binary_search.zig
```diff
@@ -48,6 +49,46 @@ pub fn binary_search_values_upsert_index(
...
+            //
+            // Prefetching the two possible positions we'd need for the next iteration:
+            //
+            // [....................mid....................]
+            //           ^                       ^
+            //       one quarter          three quarters.
+
+            // We need to use pointer arithmetic and disable runtime safety to avoid bounds checks,
+            // otherwise prefetching would harm the performance instead of improving it.
+            // Since these pointers are never derefered, it's safe to dismiss this extra cost here.
```
sentientwaffle:
```suggestion
            // Since these pointers are never dereferenced, it's safe to dismiss this extra cost here.
```

src/lsm/binary_search.zig
```diff
@@ -48,6 +49,46 @@ pub fn binary_search_values_upsert_index(
...
+            @setRuntimeSafety(false);
```
sentientwaffle: Maybe leave runtime safety on when `constants.verify` is set?

sentientwaffle (approved):

jorangreef:
@batiati, for posterity here, how does performance compare before/after?

batiati:
> @batiati, for posterity here, how does performance compare before/after?

We have a benchmark comparing with/without prefetching: `./zig/zig build benchmark_binary_search`.

Running this benchmark on my laptop was inconclusive. I can only say that prefetching does not hurt the performance, but any improvement can be just measurement noise ... I assume the CPU can do a really good job on hardware prefetching. Overall, I think we may take advantage of prefetching in some CPUs, but this gain can be easily diluted by other layers such as the hypervisor and the OS scheduler.

Arrays with larger values (i.e. 128 bytes per element) are where the prefetch seems to be more consistent, but again, it can easily vary between executions.

`PB` is the prefetch binary search and `_B` is the regular binary search without prefetching.
¹ _Unfortunately, the performance counters are disabled on AWS Linux, so I could only measure time._
₂ _Each line is running 500_000 searches_
```
K=_8B V=128B N=_256 PB: WT=__3805ns UT=__3804ns CY=_____0 IN=_____0 CR=____0 CM=____0 BM=0
K=_8B V=128B N=_256 _B: WT=__3822ns UT=__3821ns CY=_____0 IN=_____0 CR=____0 CM=____0 BM=0
K=_8B V=128B N=_512 PB: WT=__7968ns UT=__7968ns CY=_____0 IN=_____0 CR=____0 CM=____0 BM=0
K=_8B V=128B N=_512 _B: WT=__8120ns UT=__8119ns CY=_____0 IN=_____0 CR=____0 CM=____0 BM=0
K=_8B V=128B N=1024 PB: WT=_16734ns UT=_16733ns CY=_____0 IN=_____0 CR=____0 CM=____0 BM=0
K=_8B V=128B N=1024 _B: WT=_17763ns UT=_17762ns CY=_____0 IN=_____0 CR=____0 CM=____0 BM=0
K=_8B V=128B N=2048 PB: WT=_40944ns UT=_40941ns CY=_____0 IN=_____0 CR=____0 CM=____0 BM=0
K=_8B V=128B N=2048 _B: WT=_42079ns UT=_42075ns CY=_____0 IN=_____0 CR=____0 CM=____0 BM=0
K=_8B V=128B N=4096 PB: WT=_85441ns UT=_85433ns CY=_____0 IN=_____0 CR=____0 CM=____0 BM=0
K=_8B V=128B N=4096 _B: WT=_87048ns UT=_87039ns CY=_____0 IN=_____0 CR=____0 CM=____0 BM=0
```

PR #1148, opened 2023-08-21
VSR/SuperBlock: SuperBlock.replica_index

sentientwaffle:
By convention, logs start with the prefix: `{replica_index}: ...`.
This makes it easy to distinguish and filter the cluster members in simulator debug log output.
But many components don't currently have access to the replica index.

Add it to `SuperBlock` so that `SuperBlock` itself, and also LSM and Grid, can log the replica index.

### DIFF
--- a/src/vsr/superblock.zig
+++ b/src/vsr/superblock.zig
@@ -1045,8 +1045,9 @@ pub fn SuperBlockType(comptime Storage: type) type {
             @memset(buffer[trailer_size_..], 0); // Zero sector padding.
 
             const offset = trailer.zone().start_for_copy(context.copy.?);
-            log.debug("{s}: write_trailer_next: " ++
+            log.debug("{?}: {s}: write_trailer_next: " ++
                 "trailer={s} copy={} checksum={x:0>32} size={} offset={}", .{
+                superblock.replica_index,
                 @tagName(context.caller),
                 @tagName(trailer),
                 context.copy.?,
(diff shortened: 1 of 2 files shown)


matklad:
I think, long term, we'd want to print git-style short hash of replica random ID here (and also rig the simulator to produce sequential "random" ids). 

And we _might_ want to have some nicer logging API here, so that we do

```
replica.log.info("method", .{ args });
```

or 

```
log.info(replica.id, "method", .{ args });
```


rather than

```
log.info("{}: method", .{ self.id, args});
```

Just some simmering thoughts ....

src/vsr/superblock.zig
```diff
@@ -184,7 +184,7 @@ pub const SuperBlockHeader = extern struct {
             assert(state.view >= state.log_view);
             assert(state.replica_count > 0);
             assert(state.replica_count <= constants.replicas_max);
-            vsr.assert_valid_member(&state.members, state.replica_id);
```
matklad, line 187: :heart:

src/vsr/superblock.zig
```diff
@@ -1509,6 +1531,10 @@ pub fn SuperBlockType(comptime Storage: type) type {
                 .open => {
                     assert(!superblock.opened);
                     superblock.opened = true;
+                    superblock.replica_index = vsr.member_index(
```
matklad, line 1537: Let's also set this in `pub fn format`?

matklad (approved):

PR #1149, opened 2023-08-22
zig: fully drop previously vendored HashMap

matklad:


sentientwaffle (approved):
:tada:

jorangreef:
🥳 

(Here's to thinking long term about a hash map that's optimized for static memory allocation, knowing it never needs to resize...)

PR #1151, opened 2023-08-22
Refactor tree_id to u16.

batiati:
Refactors the `// TODO u16 + padding.` at `schema.zig`.
This refactor was started by #1145, but it left two places still as `u128`.


sentientwaffle (approved):
The superblock manifest still encodes ids as u128 but I have a branch that removes it entirely, so we may as well leave it for now.

PR #1154, opened 2023-08-22
Node: support building `tb_nodeclient.dll`

kprotty:
Generates the `node.lib` for `zig build node_client` to work/target windows.

In order to dynamically link `node.dll` on windows for NAPI calls, a `node.lib` + `linkSystemLibrary('node')` is required. A `.def` file is one way to describe exported functions in a dll for the linker to generate the corresponding `.lib` internally. Zig's current `std.Build` API doesn't seem to support passing in `.def` so we must provide it ourselves.

Visual Studio developer tools has [lib.exe](https://learn.microsoft.com/en-us/cpp/build/reference/overview-of-lib?view=msvc-170) which can be used to create a `.lib` from a `.def` file via `lib.exe /def:node.def /out:node.lib`. Zig (0.11) exposes [llvm-lib](https://llvm.org/docs/CommandGuide/llvm-lib.html) but can't seem to find any files you give it through `/DEF:`. Fortunately, Zig also exposes [dlltool](https://sourceware.org/binutils/docs/binutils/dlltool.html) which works and can do the same: `dlltool -m i386:x86-64 -d node.def -l node.lib`.


src/clients/node/scripts/build_lib.js
```diff
@@ -0,0 +1,32 @@
...
+}
+
+// Write a '.def' file for node.dll.
+const defFile = path.resolve(__dirname, '../node.def')
+const libFile = path.resolve(__dirname, '../node.lib')
+const allSymbolsArr = Array.from(allSymbols)
+fs.writeFileSync(defFile, 'LIBRARY node\nEXPORTS\n' + allSymbolsArr.join('\n'))
+
+// Compile '.def' file to '.lib' file for zig build node_client to link to.
+execSync(`${zig} dlltool -m i386:x86-64 -d ${defFile} -l ${libFile}`)
```
batiati, line 28: Can we skip the def/lib process if not `isWindows`?

src/clients/node/package.json
```diff
@@ -28,7 +28,7 @@
     "test": "node dist/test",
     "build": "npm run build_tsc && npm run build_lib",
     "build_tsc": "./node_modules/typescript/bin/tsc",
-    "build_lib": "node scripts/windows_generate_node_lib.js && cd ../../.. && ./zig/zig build node_client -Doptimize=ReleaseSafe",
+    "build_lib": "node scripts/build_lib.js",
```
batiati, line 31: 🥳

src/clients/node/scripts/build_lib.js
kprotty: Considered that, but then it could mean non-windows targets would fail their cross-compile build.

batiati (dismissed):

batiati (approved):

PR #1156, opened 2023-08-23
Add Java sample project for many in-flight two-phase transfers

### LINKED ISSUE #943: Add sample project to language clients that shows multiple two-phase transfers in flight all at once
See https://tigerbeetle.slack.com/archives/C04375JUAMS/p1688158327999759 for background.

- [x] Go client #1051
- [x] Java client https://github.com/tigerbeetle/tigerbeetle/pull/1156
- [x] Dotnet client https://github.com/tigerbeetle/tigerbeetle/pull/1153
- [x] Node client https://github.com/tigerbeetle/tigerbeetle/pull/1147

eatonphil:
Closes https://github.com/tigerbeetle/tigerbeetle/issues/943.


src/clients/java/samples/two-phase/src/main/java/Main.java
```diff
@@ -67,7 +67,6 @@ public static void main(String[] args)
 	    assert accounts.getCapacity() == 2;
 
 	    while (accounts.next()) {
-		accounts.getTimestamp();
```
eatonphil, line 70: @batiati I don't know why these `.getTimestamp()` calls were in here in the first place. But they were in many places.

eatonphil: They do not seem to be necessary. So I got rid of them. Mentioning it in case they are actually necessary.

src/clients/java/samples/two-phase/src/main/java/Main.java
```diff
@@ -96,7 +95,6 @@ public static void main(String[] args)
 	    transfers.setCode(1);
 	    transfers.setAmount(500);
 	    transfers.setFlags(TransferFlags.POST_PENDING_TRANSFER);
-	    transfers.getTimestamp();
```
batiati, line 99: Yeah, this one can be removed too. They are no-op.

eatonphil: Thanks!

batiati (approved):

PR #1157, opened 2023-08-23
Refactor Accounts and Transfers to u128 amounts

batiati:
## This PR:

- Account balances fields are now `u128`.

- Transfers amounts are `u128`.

- Timeouts are `u32` and expressed as `seconds` instead of `nanoseconds`.

- Three `user_data` fields, as `u128`, `u64` and `u32`.

- `AccountMutable` and `AccountImmutable` are now a single groove `Account`.

- Reordered the `Transfer` and `Account` fields to keep `@alignOf(T) == 16`.

## Direct changes:

**Docs:** All changes have been reflected, and new examples have been added for the `user_data_{128|64|32}` fields.

**Benchmarks, Demos, Samples, Tests, and Repl:** The code has been refactored.

**Clients:** Besides the naming and types refactor, these are specific changes for each client:

- Node client: No extensive changes were needed, as `BigInt` was already being used for `u64` values.

- Java client: The API now defaults all `get` acessors to `byte[]` for IDs and user data, and `BigInteger` for balances and amounts. Additionally, overloaded accessors allow the application to get/set values as a pair of primitive `long`. It allows integrating with third-party numerical libraries without the need to allocate temporary objects like arrays or BigIntegers.

- C# client: .NET 7.0 comes with support for `UInt128`. Users can easily opt-in to use the new `UInt128` or to use a replacement type we distribute for .NET Standard 2.1 (which is supported by other runtimes like .NET Core, Xamarin, Unity, and .NET up to 6.0). Also, there are extension methods for converting to/from `BigInteger`.

- Go client: We expose the type `Uint128` with a set of conversion functions, including `math/big.Int`. Likewise, users have the option to use `big.Int` or any other implementation to handle or format 128-bit values.

## Other changes:

- Increased `block_size` to 128K.
Without this, we can't fit everything in a single block due to the size of each object and `lsm_batch_multiple`:
```
src/lsm/schema.zig:180:51: error: overflow of integer type 'u32' with value '-1048'
        const padding_size = constants.block_size - padding_offset;
```

- Argument `--cache-grid` is now validated for the minimum size.
Currently, it's `--cache-grid=256MB` instead of `128MB` due to `block_size` changes.

- Repl parser now accepts identifiers containing numbers (but not starting with numbers).
Example: `user_data_128` would be rejected by the previous version.

<details>
<summary>Benchmarks</summary>

### Before:

```
❯ ./scripts/benchmark.sh
Formatting replica 0...
Starting replica 0...

Benchmarking...
steps [2/4] zig build-exe benchmark ReleaseSafe native-linux... LLVM Emit Object... info: Benchmark running against { 127.0.0.1:43705 }
1225 batches in 73.57 s
load offered = 1000000 tx/s
load accepted = 135926 tx/s
batch latency p00 = 0 ms
batch latency p10 = 18 ms
batch latency p20 = 22 ms
batch latency p30 = 25 ms
batch latency p40 = 26 ms
batch latency p50 = 28 ms
batch latency p60 = 29 ms
batch latency p70 = 29 ms
batch latency p80 = 30 ms
batch latency p90 = 31 ms
batch latency p100 = 3236 ms
transfer latency p00 = 0 ms
transfer latency p10 = 3435 ms
transfer latency p20 = 7277 ms
transfer latency p30 = 12325 ms
transfer latency p40 = 18875 ms
transfer latency p50 = 25889 ms
transfer latency p60 = 33611 ms
transfer latency p70 = 41591 ms
transfer latency p80 = 49480 ms
transfer latency p90 = 56775 ms
transfer latency p100 = 63583 ms
```

### After:

```
❯ ./scripts/benchmark.sh
Formatting replica 0...
Starting replica 0...

Benchmarking...
info: Benchmark running against { 127.0.0.1:38517 }
1226 batches in 66.92 s
load offered = 1000000 tx/s
load accepted = 149436 tx/s
batch latency p00 = 0 ms
batch latency p10 = 17 ms
batch latency p20 = 19 ms
batch latency p30 = 23 ms
batch latency p40 = 24 ms
batch latency p50 = 25 ms
batch latency p60 = 26 ms
batch latency p70 = 28 ms
batch latency p80 = 30 ms
batch latency p90 = 31 ms
batch latency p100 = 2754 ms
transfer latency p00 = 0 ms
transfer latency p10 = 3181 ms
transfer latency p20 = 6575 ms
transfer latency p30 = 11852 ms
transfer latency p40 = 17253 ms
transfer latency p50 = 23811 ms
transfer latency p60 = 30330 ms
transfer latency p70 = 37195 ms
transfer latency p80 = 43995 ms
transfer latency p90 = 50735 ms
transfer latency p100 = 56930 ms
```

</details>


cb22:
Interesting benchmark results, too. I remember when testing a pure block size increase, that bumped performance too due to the situation wrt compaction pacing (think the sweet spot was 256K).

Would be interesting to run those again (eg, compare `main` @ 128K).

batiati:
> Interesting benchmark results, too. I remember when testing a pure block size increase, that bumped performance too due to the situation wrt compaction pacing (think the sweet spot was 256K).

The combination of `block_size = 256k` + `lsm_batch_multiple = 128` gives the best TPS. IMO, mostly because it uses more NVMe bandwidth. I think it may change once the compaction pacing is in place, so it's not worth tuning it yet.

> Would be interesting to run those again (eg, compare `main` @ 128K).

The `main` branch with 128k blocks showed the same ~150k TPS we have measured with these PR changes.
The difference is that now we're writing all the 128 bytes of each account during balance updates (no more split between `AccountMutable` and `AccountImmutable`). I could observe about 2GB more writes during the same benchmark, but I think the NVMe could absorb the extra IO and we still gained by not having to do two lookups for the same object.

```
❯ ./scripts/benchmark.sh
Formatting replica 0...
Starting replica 0...

Benchmarking...
steps [2/4] zig build-exe benchmark ReleaseSafe native-linux... LLVM Emit Object... info: Benchmark running against { 127.0.0.1:33091 }
1224 batches in 67.04 s
load offered = 1000000 tx/s
load accepted = 149174 tx/s
batch latency p00 = 0 ms
batch latency p10 = 19 ms
batch latency p20 = 22 ms
batch latency p30 = 25 ms
batch latency p40 = 27 ms
batch latency p50 = 29 ms
batch latency p60 = 30 ms
batch latency p70 = 31 ms
batch latency p80 = 33 ms
batch latency p90 = 35 ms
batch latency p100 = 2569 ms
transfer latency p00 = 0 ms
transfer latency p10 = 3714 ms
transfer latency p20 = 7157 ms
transfer latency p30 = 12014 ms
transfer latency p40 = 17586 ms
transfer latency p50 = 23668 ms
transfer latency p60 = 29589 ms
transfer latency p70 = 36722 ms
transfer latency p80 = 43803 ms
transfer latency p90 = 50881 ms
transfer latency p100 = 57049 ms
```

src/state_machine.zig
```diff
@@ -64,140 +64,74 @@ pub fn StateMachineType(
...
-            AccountImmutable,
+            Account,
             .{
-                .ids = constants.tree_ids.accounts_immutable,
+                .ids = constants.tree_ids.accounts,
                 .value_count_max = .{
                     .timestamp = config.lsm_batch_multiple * @as(usize, @max(
                         constants.batch_max.create_accounts,
                         // ×2 because creating a transfer will update 2 accounts.
                         2 * constants.batch_max.create_transfers,
```
sentientwaffle: `2 * constants.batch_max.create_transfers,` is correct here now that the account is mutable, but previously (i.e. in `main`) where it is immutable, it should just be `constants.batch_max.create_transfers`.

(I'm not posting a fix to main since after this PR it will be correct anyway. :smile:)
Also it shouldn't impact the performance tradeoffs of the `AccountsImmutable`/`AccountMutable` split in either direction.

docs/design/data-modeling.md
```diff
@@ -47,19 +47,19 @@ ledger might look something like this:
 
 ## `user_data`
 
-`user_data` is the most flexible field in the schema (for both
-[accounts](../reference/accounts.md#user_data) and [transfers](../reference/transfers.md#user_data)).
+`user_data_128`, `user_data_64` and `user_data_32` are the most flexible fields in the schema (for both
+[accounts](../reference/accounts.md) and [transfers](../reference/transfers.md)).
 `user_data`'s contents are arbitrary, interpreted only by the application.
 
 `user_data` is indexed for efficient point and range queries.
```
sentientwaffle:
```suggestion
Each `user_data` field's contents are arbitrary, interpreted only by the application.

Each `user_data` field is indexed for efficient point and range queries.
```

docs/reference/transfers.md
```diff
@@ -203,10 +214,49 @@ Constraints:
 * Type is 128-bit unsigned integer (16 bytes)
 * Must be zero if neither void nor pending transfer flag is set
 * Must match an existing transfer's [`id`](#id) if non-zero
+### `user_data_128`
```
sentientwaffle:
```suggestion
* Must match an existing transfer's [`id`](#id) if non-zero

### `user_data_128`
```

docs/reference/transfers.md
```diff
@@ -217,10 +267,10 @@ Non-pending transfers cannot have a timeout.
 
 Constraints:
 
-* Type is 64-bit unsigned integer (8 bytes)
+* Type is 32-bit unsigned integer (4 bytes)
 * Must be zero if `flags.pending` is *not* set
-* Must not overflow a 64-bit unsigned integer when summed with the transfer's timestamp
-  (`error.overflows_timeout`)
+* Must not overflow a 64-bit unsigned integer when converted to nanoseconds
+  and summed with the transfer's timestamp (`error.overflows_timeout`)
```
sentientwaffle: I guess this warning is less important now that the user passes in seconds, which we convert to nanoseconds -- unless the TB cluster's clock is 400 years in the future! :wink:
May as well leave it in for completeness, though.

src/clients/go/pkg/types/main_test.go
```diff
@@ -39,3 +42,31 @@ func Test_HexStringToUint128_LittleEndian(t *testing.T) {
 		t.Errorf("Expected %s to produce bytes %v, got bytes %v", test, expected, res.Bytes())
 	}
 }
+
+func Test_BigIntToUint128(t *testing.T) {
+big.Int.
```
sentientwaffle: Is this a typo?

src/tigerbeetle.zig
```diff
@@ -124,10 +128,12 @@ pub const CreateAccountResult = enum(u32) {
     credits_posted_must_be_zero = 14,
```
sentientwaffle: Error codes of the same "type" should be ordered to match their precedence in the state machine.
And that precedence should match the order of the fields in the struct.

For example, now that the balance fields precede `ledger`/`code` in the struct layout, the balance field "must_be_zero" error codes should precede (and take precedence over) the `ledger`/`node` "must_not_be_zero" error codes.

Likewise, `exists_with_different_flags` should follow `exists_with_different_code` (I think this was actually wrong before too).

In the state machine unit tests, we test error precedence by trying to create a transfer with N errors and checking which code comes back, then creating a transfer with N-1 errors, then N-2, etc.

src/tigerbeetle/main.zig
```diff
@@ -154,8 +155,21 @@ const Command = struct {
         }
 
         const grid_cache_size = @as(u64, args.cache_grid_blocks) * constants.block_size;
+
+        // The amount of bytes in `--cache-grid` must be a multiplle of
```
```suggestion
        // The amount of bytes in `--cache-grid` must be a multiple of
```

src/tigerbeetle/main.zig
```diff
@@ -154,8 +155,21 @@ const Command = struct {
...
+        // `constants.block_size` and `SetAssociativeCache.value_count_max_multiple`,
+        // and it may be converted to zero if a value too small is informed.
```
sentientwaffle: I'm not sure that "informed" makes sense in this context. Maybe instead something like:
```suggestion
        // and it may be converted to zero if a smaller value is passed in.
```

src/tigerbeetle/repl.zig
```diff
@@ -121,8 +121,12 @@ pub const Parser = struct {
...
 
         while (parser.offset < parser.input.len and
-            (std.ascii.isAlphabetic(parser.input[parser.offset]) or
-            parser.input[parser.offset] == '_'))
+            (parser.input[parser.offset] == '_' or
+            // Identifiers may contain numbers, but not start with a number.
+            if (after_whitespace == parser.offset)
+            std.ascii.isAlphabetic(parser.input[parser.offset])
+        else
+            std.ascii.isAlphanumeric(parser.input[parser.offset])))
```
sentientwaffle: Maybe split this out into an `if { break }` statement within the `while` body, to keep the condition readable.

src/state_machine.zig
```diff
@@ -64,140 +64,74 @@ pub fn StateMachineType(
...
-                    .credits_posted = 10,
+                    .user_data_128 = 3,
+                    .user_data_64 = 4,
+                    .user_data_32 = 5,
+                    .ledger = 6,
+                    .code = 7,
+                    .debits_pending = 8,
+                    .debits_posted = 9,
+                    .credits_pending = 10,
+                    .credits_posted = 11,
```
sentientwaffle: Reorder these to match the field order in the `tigerbeetle.Account` struct.

src/state_machine.zig
```diff
@@ -64,140 +64,74 @@ pub fn StateMachineType(
...
+                    .debit_account_id = 14,
+                    .credit_account_id = 15,
+                    .user_data_128 = 16,
+                    .user_data_64 = 17,
+                    .user_data_32 = 18,
+                    .pending_id = 19,
+                    .timeout = 20,
+                    .ledger = 21,
+                    .code = 22,
+                    .amount = 23,
```
sentientwaffle: Reorder these to match the field order in the `tigerbeetle.Transfer` struct.

src/state_machine.zig
```diff
@@ -64,140 +64,74 @@ pub fn StateMachineType(
...
+                    .ledger = config.lsm_batch_multiple * @as(usize, @max(
+                        constants.batch_max.create_accounts,
+                        2 * constants.batch_max.create_transfers,
+                    )),
+                    .code = config.lsm_batch_multiple * @as(usize, @max(
                         constants.batch_max.create_accounts,
-                        // ×2 because creating a transfer will update 2 accounts.
                         2 * constants.batch_max.create_transfers,
                     )),
                     // Transfers mutate the secondary indices for debits/credits pending/posted.
```
sentientwaffle: Reorder fields to match `tigerbeetle.Account` field order.

src/state_machine.zig
```diff
@@ -64,140 +64,74 @@ pub fn StateMachineType(
...
-                .value_count_max = .{
-                    .timestamp = config.lsm_batch_multiple * @as(usize, @max(
+                    .id = config.lsm_batch_multiple * @as(usize, @max(
+                        constants.batch_max.create_accounts,
+                        2 * constants.batch_max.create_transfers,
+                    )),
+                    .user_data_128 = config.lsm_batch_multiple * @as(usize, @max(
+                        constants.batch_max.create_accounts,
+                        2 * constants.batch_max.create_transfers,
+                    )),
```
sentientwaffle: Remove the `2 * constants.batch_max.create_transfers` component for all account fields _except_ `timestamp` and the balance fields.

e.g.:

```suggestion
                    .user_data_128 =
                        config.lsm_batch_multiple * constants.batch_max.create_accounts,
```

Non-balance account secondary index fields are not mutated by `create_transfers`, and since they aren't mutated,  we won't update them.

See: https://github.com/tigerbeetle/tigerbeetle/blob/4e69eb3f9f20484d30ae9dda9fbf24004cb58d31/src/lsm/groove.zig#L923-L934

docs/reference/transfers.md
batiati: Agreed!
I removed it from here and instead it's now mentioned in `create_transfers.md#overflows_timeout`

src/tigerbeetle/main.zig
```diff
@@ -154,8 +155,21 @@ const Command = struct {
         }
 
         const grid_cache_size = @as(u64, args.cache_grid_blocks) * constants.block_size;
+
+        // The amount of bytes in `--cache-grid` must be a multiple of
+        // `constants.block_size` and `SetAssociativeCache.value_count_max_multiple`,
+        // and it may be converted to zero if a smaller value is passed in.
```
```suggestion
        // and it may have been converted to zero if a smaller value is passed in.
```

(Since this conversion happens in `cli.zig`, not here.)

src/tigerbeetle/main.zig
```diff
@@ -154,8 +155,21 @@ const Command = struct {
...
+        if (grid_cache_size == 0) {
+            fatal("Grid cache must be greater than {}MB. See --cache-grid", .{
+                @divExact(
+                    constants.block_size * Grid.Cache.value_count_max_multiple,
```
sentientwaffle: Maybe extract this into `grid_cache_size_min` (since we have `grid_cache_size` and `grid_cache_size_warn`). Then the `@divExact` will fit on one line, too.

And if `grid_cache_size != 0`, then we can assert that `grid_cache_size >= grid_cache_size_min`.

src/state_machine.zig
```diff
@@ -64,142 +64,48 @@ pub fn StateMachineType(
...
+                    .debits_pending = 2,
+                    .debits_posted = 3,
+                    .credits_pending = 4,
+                    .credits_posted = 5,
+                    .user_data_128 = 6,
+                    .user_data_64 = 7,
+                    .user_data_32 = 8,
+                    .ledger = 9,
+                    .code = 10,
+                    .timestamp = 11,
```
sentientwaffle, line 78: :thinking: Despite the name "timestamp" here, what this actually corresponds to is the groove's object tree. Imo that should take precedence over the field order, and this tree's id should be listed/assigned before `id`. (Likewise for `transfers`).

src/state_machine.zig
```diff
@@ -1383,31 +1211,40 @@ pub fn StateMachineType(
...
 
-fn zeroed_48_bytes(a: [48]u8) bool {
-    const x = @as([6]u64, @bitCast(a));
+inline fn zeroed_48_bytes(a: [48]u8) bool {
+    const x: [6]u64 = @bitCast(a);
     return (x[0] | x[1] | x[2] | x[3] | x[4] | x[5]) == 0;
 }
 
+/// Optimizes for the common case, where the arrays are equal. Completely branchless.
+inline fn equal_4_bytes(a: [4]u8, b: [4]u8) bool {
```
sentientwaffle: It looks like these `equal_*_bytes` and `zeroed_*_bytes` aren't actually used; can they be removed?

src/state_machine.zig
```diff
@@ -1947,49 +1800,53 @@ test "create_transfers/lookup_transfers" {
...
-        \\ transfer T2 A3 A1  _  _  _   _ L1 C2 _   _ _ _ _ _  _    7 _ ok
-        \\ transfer T3 A1 A3  _  _  _   _ L1 C2 _   _ _ _ _ _  _    3 _ ok
+        \\ transfer   T1 A1 A3  123   _  _  _  _    1 L2 C1   _ PEN   _   _   _   _  _ _ transfer_must_have_the_same_ledger_as_accounts
+        \\ transfer   T1 A1 A3   -0   _ U1 U1 U1    _ L1 C2   _   _   _   _   _   _  _ _ exists_with_different_flags
+        \\ transfer   T1 A3 A1   -0   _ U1 U1 U1    1 L1 C2   _ PEN   _   _   _   _  _ _ exists_with_different_debit_account_id
+        \\ transfer   T1 A1 A4   -0   _ U1 U1 U1    1 L1 C2   _ PEN   _   _   _   _  _ _ exists_with_different_credit_account_id
+        \\ transfer   T1 A1 A3   -0   _ U1 U1 U1    1 L1 C1   _ PEN   _   _   _   _  _ _ exists_with_different_amount        
+        \\ transfer   T1 A1 A3  123   _ U1 U1 U1    1 L1 C2   _ PEN   _   _   _   _  _ _ exists_with_different_user_data_128
+        \\ transfer   T1 A1 A3  123   _  _ U1 U1    1 L1 C2   _ PEN   _   _   _   _  _ _ exists_with_different_user_data_64
+        \\ transfer   T1 A1 A3  123   _  _  _ U1    1 L1 C2   _ PEN   _   _   _   _  _ _ exists_with_different_user_data_32
```
sentientwaffle, line 1799: Perfect!

src/tigerbeetle.zig
```diff
@@ -6,32 +6,34 @@ const stdx = @import("./stdx.zig");
 
 pub const Account = extern struct {
```
sentientwaffle, line 7: Also need to update `docs/DESIGN.md` with these account/transfer schema changes.

src/state_machine.zig
jorangreef: > what this actually corresponds to is the groove's object tree

However, our usage of `timestamp` as internal identifier for the groove's object tree is a secondary implementation detail. The primary usage of `timestamp` is external, as a totally ordered hybrid logical clock. Beyond that, the reason it's last, is because it's the only field that's "filled in"—in the sense that it's appended to the existing data.

src/state_machine.zig
jorangreef: These test vectors are always so nice and concise.

src/state_machine.zig
jorangreef: Also, ahead of `timestamp` is `id`, which is the external identifier and which comes first (it's the "key", followed by the value). Users provide this much of the key/value, consecutively, and then the time is "stamped" on after.

sentientwaffle: > Users provide this much of the key/value, consecutively, and then the time is "stamped" on after.

This code is defining/assigning the LSM tree ids though, it doesn't relate to anything that users see.
From the LSM's perspective, the object tree is the "primary" tree -- object and id trees are special cases in the groove, separate from all of the secondary index trees.

batiati: Imo, there's not much benefit that justifies not following the object's declaration order here.
We already handle `timestamp` as a special case in the LSM, which is very clear,  there's no need to reinforce it.

batiati: Anyway, those fields are intended to be stable IDs and _may_ become unordered in the future if we ever change our schema. Either we will follow the declaration order or the numerical order. So, I don't have strong feelings about it.

batiati:
Note: Dotnet tests are failing for Macos self-hosted.
Please ignore it for this PR, I suspect it is due to cache issues because the self-hosted runner has no isolation when building .Net 7.0 for this PR and .Net 6.0 for everything else.

sentientwaffle (approved):

src/clients/java/src/main/java/com/tigerbeetle/AccountBatch.java
```diff
@@ -112,197 +118,373 @@ public byte[] getUserData() {
...
+     */
+    public void setUserData128(final long leastSignificant) {
+        putUInt128(at(Struct.UserData128), leastSignificant, 0);
+    }
+
+    /**
+     * @throws IllegalStateException if not at a {@link #isValidPosition valid position}.
+     * @see <a href="https://docs.tigerbeetle.com/reference/accounts/#user_data_64">user_data_64</a>
+     */
+    public long getUserData64() {
```
matklad, line 380: :thinking: it feels like this API makes it too easy to ignore high order bits. Should we perhaps through an exception here if this is not the entirety of user data?

matklad: And this is also inconsistetn with how, eg, `getDebitsPosted` works (that API makes sense more to me personally). 

@batiati it feels like maybe we have an incomplete refactor here?

PR #1160, opened 2023-08-23
VSR: State Sync (Part 3/?): sync_op_{min,max}; sync replies

sentientwaffle:
## `sync_op_{min,max}`

When state sync completes, the replica still needs to repair the LSM table blocks and VSR client replies that it skipped over. (This PR only sync replies, for ease of review.)

How it works:

As part of the new superblock installed by state sync, set `vsr_state.sync_op_{min,max}`.
(If `sync_op_max ≠ 0` _before_ this sync, it means we were still repairing from a previous sync. In that case, we grow to range to encompass both sync's commits.)

On every checkpoint, either:
- If all of the tables/replies are synced, set `sync_op_{min,max}` to zero. (See `sync_content_done()`).
- If all of the tables/replies _aren't_ synced, reuse the same values for `sync_op_{min,max}`.

When we open the state machine, if `sync_op_max ≠ 0`, start syncing replies and tables.

`sync_op_max`:

    /// When zero, all of the grid blocks and replies are synced.
    /// (When zero, `sync_op_min` is also zero.)
    ///
    /// When nonzero, we must repair grid-blocks/client-replies that would have been written
    /// during the commits between `sync_op_min` and `sync_op_max` (inclusive).
    /// (Those grid-blocks and client-replies were not written normally because we "skipped"
    /// past them via state sync.)

## Sync replies

Use `sync_op_{min,max}` to repair replies immediately after syncing the superblock.

In a follow-up PR this same machinery will sync LSM table blocks too, but that is separate (for ease of review).


src/vsr/replica.zig
```diff
@@ -7734,6 +7788,41 @@ pub fn ReplicaType(
             self.sync_dispatch(.idle);
         }
 
+        /// We have just:
+        /// - finished state sync,
```
matklad: Maybe call it "supreblock sync"? It is a bit surprising that "state sync" doesn't actually sync the state itself, only meta info.

src/vsr/replica.zig
```diff
@@ -7734,6 +7788,41 @@ pub fn ReplicaType(
...
+        /// Now we sync:
+        /// - the missed LSM table blocks (index/filter/data), and
+        /// - client reply messages.
+        fn sync_content(self: *Self) void {
+            assert(self.syncing == .idle);
+            assert(self.state_machine_opened);
+            assert(self.superblock.working.vsr_state.sync_op_max > 0);
+
+            var entry_slot: usize = 0;
+            while (entry_slot < constants.clients_max) : (entry_slot += 1) {
```
```suggestion
            for (0..constants.clients_max) |entry_slot| {
```

src/vsr/replica.zig
```diff
@@ -7734,6 +7788,41 @@ pub fn ReplicaType(
...
+                const entry = &self.superblock.client_sessions.entries[entry_slot];
+                const entry_free = self.superblock.client_sessions.entries_free.isSet(entry_slot);
+                const entry_faulty = !entry_free and
+                    entry.header.size > @sizeOf(Header) and
+                    entry.header.op >= self.superblock.working.vsr_state.sync_op_min and
+                    entry.header.op <= self.superblock.working.vsr_state.sync_op_max;
+
+                self.client_replies.faulty.setValue(entry_slot, entry_faulty);
```
matklad: Hm, wouldn't this falsely clear faulty status of entries outside o f`sync_op_min..sync_op_max`?

sentientwaffle: Great catch!

docs/internals/sync.md
```diff
@@ -20,25 +20,26 @@ In the context of state sync, "state" refers to:
 6. the grid (LSM data; acquired blocks only)
 7. client replies
 
-However, state sync protocol itself only syncs the first 5.
-6/7 are replicated after state sync protocol, driven by grid repair and the disk scrubber.
+State sync consists of two protocols:
+- [Sync Superblock](./vsr.md#protocol-sync-superblock) (syncs 1-5)
```
jorangreef, line 24:
```suggestion
- [Sync SuperBlock](./vsr.md#protocol-sync-superblock) (syncs 1-5)
```

docs/internals/sync.md
```diff
@@ -20,25 +20,26 @@ In the context of state sync, "state" refers to:
...
+- [Sync Content](./vsr.md#protocol-sync-content) (syncs 6-7)
```
jorangreef:
```suggestion
- [Sync Grid](./vsr.md#protocol-sync-grid) (syncs 6-7)
```

"Content" was not clear/descriptive enough to me at first glance, to get an idea as to what might be synced. "Content" is more general, like "Data". Whereas, if we are syncing the grid blocks here, then I think it would be great to call it that, rather than introduce a new term.

sentientwaffle: "Grid" is too specific; we also sync the client replies during this phase.

docs/internals/sync.md
sentientwaffle: The capitalization of "Superblock" here is deliberate -- I am using the word "superblock" as a noun, not referring to the `SuperBlock` struct.

docs/internals/sync.md
jorangreef: Yes, I noticed that. Sync Client Replies could also be decoupled out of that.

docs/internals/sync.md
jorangreef: I would prefer we just use the same capitalization as the struct (I think I do this even in our other writing about the SuperBlock).

For example, we also do "TigerBeetle". 😁

docs/internals/sync.md
jorangreef: Then the different sync strategies all relate to the zone they sync.

docs/internals/sync.md
sentientwaffle: I changed it, but fwiw I don't think "TigerBeetle" is the same situation. "TigerBeetle" is a company, whereas "tigerbeetle" is an insect. :wink: 

But in the superblock case, `SuperBlock` is an implementation of the "superblock" ("Superblock") concept from file systems.

By analogy, the protocol for repairing client replies is labeled "Protocol: Repair Client Replies", not "Protocol: Repair ClientReplies". "ClientReplies" is an implementation concern, "Client Replies" is a protocol concern. Likewise for the [Ss]uper[Bb]lock. :grin:

src/testing/cluster/storage_checker.zig
```diff
@@ -81,121 +79,95 @@ pub fn StorageCheckerType(comptime Replica: type) type {
...
-                const trailer_size = @field(working, @tagName(trailer) ++ "_size");
-                var copy: u8 = 0;
-                while (copy < constants.superblock_copies) : (copy += 1) {
-                    const trailer_start = trailer.zone().start_for_copy(copy);
-                    @field(checkpoint, "checksum_superblock_" ++ @tagName(trailer)) |=
-                        vsr.checksum(storage.memory[trailer_start..][0..trailer_size]);
-                }
+            const syncing = superblock.working.vsr_state.sync_op_max > 0;
+            if (!syncing) {
+                checkpoint_actual.put(.client_replies, checksum_client_replies(superblock));
```
matklad, line 93: I was wondering how this works in the presence of faults. That's because we don't actually overwrite memory with garbage on a fault, but rather just set a bit in an out-of-band bitset, right? So we might have to re-think the approach here onece we start doign misdirected writes (which I think would have to modify the memory directly)

src/testing/storage.zig
```diff
@@ -500,6 +511,27 @@ pub const Storage = struct {
...
+        area: Area,
+    ) []align(constants.sector_size) const u8 {
+        const sectors = area.sectors();
+        const area_min = sectors.min * constants.sector_size;
+        const area_max = sectors.max * constants.sector_size;
+        return @alignCast(storage.memory[area_min..area_max]);
+    }
+
+    /// Returns whether any sector in the area is corrupt.
+    pub fn area_faulty(storage: *const Storage, area: Area) bool {
```
matklad, line 525: unused?

src/testing/storage.zig
```diff
@@ -610,7 +642,7 @@ pub const Area = union(enum) {
     grid: struct { address: u64 },
 
     fn sectors(area: Area) SectorRange {
-        switch (area) {
```
matklad, line 613: o_O was this some bug in compiler?

src/vsr/client_replies.zig
```diff
@@ -326,6 +326,11 @@ pub fn ClientRepliesType(comptime Storage: type) type {
                 },
             }
 
+            // Clear the fault *before* the write completes, not after.
+            // Otherwise, a replica exiting state sync might mark a reply as faulty, then the
+            // ClientReplies clears that bit due to an unrelated write that was already queued.
+            client_replies.faulty.unset(slot.index);
```
matklad, line 334: Assert in `read_reply_sync` that this bit is cleared if there's a pending write.

src/vsr/replica.zig
```diff
@@ -7734,6 +7788,43 @@ pub fn ReplicaType(
...
+                }
+            }
+        }
+
+        pub fn sync_content_done(self: *const Self) bool {
+            if (self.superblock.staging.vsr_state.sync_op_max == 0) {
+                return true;
+            } else {
+                // TODO Return whether all tables are synced.
+                return self.client_replies.faulty.count() == 0;
```
matklad: :thinking: we could have faulty replies outside of the min-max range. So perhaps this should `.iterate` and check if the bit is in range?

docs/internals/sync.md
jorangreef: Thanks for the context and motivation. You're right. I thought that other file systems used "super block" as in "tiger beetle" (the insect!) 😜 but they do in fact do "superblock" and "Superblock" in their docs. Let's bikeshed one last time and go with what you had!

jorangreef: This is going to be funny for posterity...

src/testing/cluster/storage_checker.zig
sentientwaffle: > That's because we don't actually overwrite memory with garbage on a fault, but rather just set a bit in an out-of-band bitset, right?

Correct!

> So we might have to re-think the approach here onece we start doign misdirected writes (which I think would have to modify the memory directly)

I think for misdirected writes we should do something similar, i.e. store the "incorrect but visible" versions in a separate data structure that we consult on reads, while storing the "true but not visible" versions in `storage.memory`. That is, keep `storage.memory` as pristine as possible, to keep checkers simple.

src/testing/storage.zig
sentientwaffle: Ah, hazards of `git add --patch`. It will be used in one of the next PRs (by the simulator, as part of `core_missing_blocks`).

src/testing/storage.zig
sentientwaffle: I think that this code was not being used before, so I guess the compiler ignored it.

sentientwaffle:
Note https://github.com/tigerbeetle/tigerbeetle/pull/1160/commits/8b2cfc250664658fae68d7bd0d403dd78158b582
Previously we used `vsr.Checkpoint.checkpoint_before()`, but that was removed (https://github.com/tigerbeetle/tigerbeetle/pull/1159) -- this is more flexible anyway.

matklad (approved):

PR #1161, opened 2023-08-24
Add note about oomkiller to docker-compose docs

eatonphil:


batiati:
We can also suggest using --cache-grid=128MB for development setups

eatonphil:
>  We can also suggest using --cache-grid=128MB for development setups

Ok! Added.

cb22 (approved):

PR #1162, opened 2023-08-24
VSR/SuperBlock: Fix grid_blocks_max, data_file_size_min

sentientwaffle:
`grid_blocks_max` and `data_file_size_min` were not counting the client replies.

(This was missed because we don't have any tests that fill the whole grid.) (We don't test a full grid yet since it isn't currently handled properly.)

Also, in `replica_test.zig`, use `storage_size_limit == storage_size_max`. This is needed because a test (`Cluster: recovery: grid corruption (disjoint)`) iterates over every block in the grid and marks faults.

### DIFF
--- a/src/vsr/replica_format.zig
+++ b/src/vsr/replica_format.zig
@@ -154,7 +154,10 @@ test "format" {
 
     var storage = try Storage.init(
         allocator,
-        superblock_zone_size + constants.journal_size_headers + constants.journal_size_prepares,
+        superblock_zone_size +
+            constants.journal_size_headers +
+            constants.journal_size_prepares +
+            constants.client_replies_size,
         .{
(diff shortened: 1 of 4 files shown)


src/vsr/replica_format.zig
```diff
@@ -154,7 +154,10 @@ test "format" {
 
     var storage = try Storage.init(
         allocator,
-        superblock_zone_size + constants.journal_size_headers + constants.journal_size_prepares,
+        superblock_zone_size +
+            constants.journal_size_headers +
+            constants.journal_size_prepares +
```
matklad:
```suggestion
            constants.journal_size_max
```

Also, it should probably be just `constants.journal_size`? The size is not variable

matklad (approved):

PR #1163, opened 2023-08-25
vsr: use stable checksum for ClusterConfig

### LINKED ISSUE #1009: Revisit vsr.root_members and ConfigCluster.checksum
These two things want to do the same --- compress structured data into a single u128 hash for comparision. They use two different approaches (Blake hashing field-by-field vs Aegis hashing mem.asBytes), and we might want to unify this. As of Zig 0.9, this runs into comptime limitations though, revisit after we upgrade zig.

matklad:
This checksum could leak to disk or network, so it must be stable across releases of TigerBeetle.

Closes: #1009

### DIFF
--- a/src/config.zig
+++ b/src/config.zig
@@ -167,18 +167,14 @@ const ConfigCluster = struct {
     /// It is used to assert that all cluster members share the same config.
     pub fn checksum(comptime config: ConfigCluster) u128 {
         @setEvalBranchQuota(10_000);
-        // TODO(Zig): Cleanup when this is fixed after Zig 0.11.
-        // Blake3 in the stdlib has a compile error so currently using Blake2 as a workaround.
-        // We should eventually switch this to vsr.checksum() once that works at comptime.
-        var hasher = std.crypto.hash.blake2.Blake2s256.init(.{});
-        inline for (std.meta.fields(ConfigCluster)) |field| {
+        comptime var config_bytes: []const u8 = &.{};
+        comptime for (std.meta.fields(ConfigCluster)) |field| {
             const value = @field(config, field.name);
             const value_64 = @as(u64, value);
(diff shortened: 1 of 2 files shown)


src/vsr.zig
```diff
@@ -1593,17 +1593,20 @@ pub const Members = [constants.members_max]u128;
 /// replica ids for the initial cluster deterministically.
 pub fn root_members(cluster: u32) Members {
     const IdSeed = extern struct {
-        cluster_config_checksum: u128 align(1) = constants.config.cluster.checksum(),
+        cluster_config_checksum: u128 align(1),
```
kprotty, line 1596: Nice, this avoids having `vsr.checksum()` support being called at comptime.

src/config.zig
```diff
@@ -167,18 +167,14 @@ const ConfigCluster = struct {
...
+        comptime for (std.meta.fields(ConfigCluster)) |field| {
             const value = @field(config, field.name);
             const value_64 = @as(u64, value);
-            hasher.update(std.mem.asBytes(&value_64));
-        }
-        var target: [32]u8 = undefined;
-        hasher.final(&target);
-        return @as(u128, @bitCast(target[0..@sizeOf(u128)].*));
+            assert(builtin.target.cpu.arch.endian() == .Little);
+            config_bytes = config_bytes ++ std.mem.asBytes(&value_64);
```
kprotty, line 175: Random thought: could the comptime interpreter be defined to run in a setting that's Big endian which is invisible to `builtin.target`?

kprotty (approved):

src/config.zig
matklad: I don't think so --- my understanding is that's the shtick of Zig that it fully emulates _target_ architecture at comptime.

kprotty: Yea you're right.

PR #1164, opened 2023-08-26
bootstrap.ps1: make curl follow redirects

rbino:
The download for Windows was failing in a Github CI

### DIFF
--- a/bootstrap.ps1
+++ b/bootstrap.ps1
@@ -18,8 +18,8 @@ if ($build) {
 } else {
     echo "Downloading pre-built TigerBeetle binary for your machine."
     echo ""
-    
-    curl -o tigerbeetle.zip "https://github.com/tigerbeetle/tigerbeetle/releases/download/$version/tigerbeetle-x86_64-windows-$version.zip"
+
+    curl.exe -Lo tigerbeetle.zip "https://github.com/tigerbeetle/tigerbeetle/releases/download/$version/tigerbeetle-x86_64-windows-$version.zip"
     unzip -qo tigerbeetle.zip
 }
 


bootstrap.ps1
```diff
@@ -19,7 +19,7 @@ if ($build) {
     echo "Downloading pre-built TigerBeetle binary for your machine."
     echo ""
     
-    curl -o tigerbeetle.zip "https://github.com/tigerbeetle/tigerbeetle/releases/download/$version/tigerbeetle-x86_64-windows-$version.zip"
+    curl -Lo tigerbeetle.zip "https://github.com/tigerbeetle/tigerbeetle/releases/download/$version/tigerbeetle-x86_64-windows-$version.zip"
```
batiati:
```suggestion
    curl.exe -Lo tigerbeetle.zip "https://github.com/tigerbeetle/tigerbeetle/releases/download/$version/tigerbeetle-x86_64-windows-$version.zip"
```


Nice catch @rbino!

Just a nit, PowerShell has aliases for some Unix commands, and [curl is one of them, even after Microsoft started to distribute cURL as a Windows component](https://curl.se/windows/microsoft.html).
I couldn't find exactly if/when they removed this alias, but I'm pretty sure that in older versions of Windows, it will call `Invoke-WebRequest` instead of `curl`, so it's better to disambiguate calling `curl.exe`.

batiati (changes requested):

bootstrap.ps1
rbino: Fixed and pushed

batiati (approved):
Thank you ⚡😎

PR #1165, opened 2023-08-28
lsm: assert that data's block padding is zero

matklad:


src/lsm/table.zig
```diff
@@ -432,18 +398,56 @@ pub fn TableType(
...
+                            assert(compare_keys(key_from_value(a), key_from_value(b)) == .lt);
+                            a = b;
+                        }
+                    }
+
+                    assert(builder.value_count == values.len);
+                    assert(block_size - header.size ==
+                        (data.value_count_max - values.len) * @sizeOf(Value) + data.padding_size);
+                    // Padding is short on average, so assert unconditionally.
+                    assert(stdx.zeroed(block[header.size..]));
```
matklad, line 424: That's the main value in this PR. As far as I can tell, we don't have an equivalent assert anywhere. 

It wasn't exactly clear what's the best place to add this assert in this method initially, so I reshufled code a bit to make it clear when we change what.

matklad: :thinking: I wonder if we do actually zero the padding here? I haven't found that code yet!

matklad: Hm, yeah, I think this is a buffer bleed! In a table with two data blocks, if the second block is only partially filled, the suffix will include the data from the previous data block!

jorangreef: There may have been an implicit zeroing by Grid's `allocate_block()`, but within compaction, and when cycling blocks, that indeed sounds like a bleed. Nice find!

jorangreef: Since this is data plane, we should gate the assertion on `verify`, as part of online verification.

matklad: See the comment above this line! I originally gated this with verify, but then reasoned that this is padding, so it actually is fine to check unconditionally: 

* if the block is full, there's not much to check
* if the block is empty, we'd have already saved some time due to not filling the block, so we can use that credit to check the padding!

src/lsm/table.zig
```diff
@@ -332,9 +332,9 @@ pub fn TableType(
                 errdefer allocator.free(data_block);
 
                 return Builder{
-                    .index_block = index_block[0..block_size],
-                    .filter_block = filter_block[0..block_size],
-                    .data_block = data_block[0..block_size],
+                    .index_block = index_block,
+                    .filter_block = filter_block,
+                    .data_block = data_block,
```
jorangreef, line 337: I believe the original intention of the comptime-known slice here is to upgrade type safety, to an array over a slice?

Can you help me understand the reason for dropping the slice now (Zig 0.11)?

src/lsm/table.zig
```diff
@@ -432,18 +398,56 @@ pub fn TableType(
                     .op = options.address,
                     .timestamp = options.snapshot_min,
                     .request = builder.value_count,
-                    .size = block_size - @as(u32, @intCast(values_padding.len + block_padding.len)),
+                    .size = @sizeOf(vsr.Header) + builder.value_count * @sizeOf(Value),
```
jorangreef, line 401: Would be good to assert these two calculations against each other, that they're the same.

src/lsm/table.zig
```diff
@@ -386,41 +386,7 @@ pub fn TableType(
...
-                    for (values[1..]) |*b| {
-                        assert(compare_keys(key_from_value(a), key_from_value(b)) == .lt);
-                        a = b;
-                    }
-                }
-
-                const values_padding = mem.sliceAsBytes(values_max[builder.value_count..]);
-                const block_padding = block[data.padding_offset..][0..data.padding_size];
-                assert(compare_keys(key_from_value(&values[values.len - 1]), key_max) == .eq);
-
```
jorangreef, line 423: Are there any invariants asserted/verified here that are not checked below? i.e. Are we dropping any checks?

src/lsm/table.zig
matklad: `allocate_block` already returns an array, so this slicing is a no-op in the current code. 

Considering potential evolution of code over time, this slicing is a safety hazard --- if at some point `data_block` local variables become slices (or arrays of different lengths), the code with the cast will continue to compile. Without the cast, it will fail to compile, which is probably good. 

This is a more general rule: in Zig, avoid casts, if possible. Zig is fairly strongly typed when it comes to coercions, so a missing cast is always a compilation error. With the cast though, you opt-out of this extra safety (and slicing by a comptime argument is a syntax for a cast in disguise). 

Compare to https://github.com/tigerbeetle/tigerbeetle/pull/1140 --- that what a bug caused by a similar cast.

src/lsm/table.zig
matklad: The

```zig
                    assert(block_size - header.size ==
                        (data.value_count_max - values.len) * @sizeOf(Value) + data.padding_size);
```

after we checksum is exactly that.

src/lsm/table.zig
jorangreef: Thanks for the excellent explanation. 🕺

src/lsm/table.zig
jorangreef: Ah, I had (somehow!) missed that comment. This is much better as you have it now.

src/lsm/table.zig
matklad: I dropped

```zig
assert(compare_keys(key_from_value(&values[values.len - 1]), key_max) == .eq);
```

as it is just an obfuscated way to write

```zig
assert(compare_keys(key_from_value(&values[values.len - 1]), key_from_value(&values[values.len - 1])) == .eq);
```

which I think doesn't pull its weight. I added

```zig
                assert(compare_keys(builder.key_max, sentinel_key) == .lt);
```

which is more meaningful.

jorangreef: Ah, I see the careful commit log showed this too.

We probably added this assertion while pairing, and it's bizarre that it made it past two people (of which I was almost certainly one!). Nice find! 🥳

src/lsm/table.zig
jorangreef: You got this.

src/lsm/table.zig
matklad: For posterity: there's no buffer bleed here, here's where we do the zeroing,

https://github.com/tigerbeetle/tigerbeetle/blob/8657375bb8f18bc022f0a5c6b5b04f8db6ad7b26/src/vsr/grid.zig#L604-L605

, but it takes a [two star programmer](https://wiki.c2.com/?ThreeStarProgrammer) to see that.

sentientwaffle (approved):

PR #1166, opened 2023-08-28
Fix CI benchmark client args

cb22:


matklad (dismissed):

matklad:
actually maybe rename the other way around and change the flag name? `--statsd` looks nicer

matklad:
ping @cb22

cb22:
@matklad sorry, with the self-assign I thought you wanted to do something here :smile: sounds good to me, will do.

matklad:
We _assign_ reviewers so that review notfications are level-triggered.

cb22:
@matklad for context behind the latest commit, @jorangreef and I discussed it would be best to disable the CI benchmark / associated AWS infra for now, until we make it more usable and tie it in better!

matklad (approved):

PR #1169, opened 2023-08-28
Some table cleanup

sm2n:
The main thing here is the compaction assert under `constants.verify`. They might not be worth it in terms of making CI even slower. But I'm not entirely sure how much of a difference it will be.

### DIFF
--- a/src/lsm/compaction.zig
+++ b/src/lsm/compaction.zig
@@ -511,6 +511,11 @@ pub fn CompactionType(
                 // Assert that we're reading data blocks in key order.
                 const values_in = compaction.values_in[index];
                 if (values_in.len > 0) {
+                    if (constants.verify) {
+                        for (values_in[0 .. values_in.len - 1], values_in[1..]) |*value, *value_next| {
+                            assert(compare_keys(key_from_value(value), key_from_value(value_next)) == .lt);
+                        }
+                    }
                     const first_key = key_from_value(&values_in[0]);
                     const last_key = key_from_value(&values_in[values_in.len - 1]);
                     if (compaction.last_keys_in[index]) |last_key_prev| {
--- a/src/lsm/compaction.zig
+++ b/src/lsm/compaction.zig
@@ -510,22 +510,21 @@ pub fn CompactionType(
 
                 // Assert that we're reading data blocks in key order.
                 const values_in = compaction.values_in[index];
-                if (values_in.len > 0) {
-                    if (constants.verify) {
-                        for (values_in[0 .. values_in.len - 1], values_in[1..]) |*value, *value_next| {
-                            assert(compare_keys(key_from_value(value), key_from_value(value_next)) == .lt);
-                        }
+                assert(values_in.len > 0);
+                if (constants.verify) {
+                    for (values_in[0 .. values_in.len - 1], values_in[1..]) |*value, *value_next| {
+                        assert(compare_keys(key_from_value(value), key_from_value(value_next)) == .lt);
                     }
-                    const first_key = key_from_value(&values_in[0]);
-                    const last_key = key_from_value(&values_in[values_in.len - 1]);
-                    if (compaction.last_keys_in[index]) |last_key_prev| {
-                        assert(compare_keys(last_key_prev, first_key) == .lt);
-                    }
-                    if (values_in.len > 1) {
-                        assert(compare_keys(first_key, last_key) == .lt);
-                    }
-                    compaction.last_keys_in[index] = last_key;
                 }
+                const first_key = key_from_value(&values_in[0]);
+                const last_key = key_from_value(&values_in[values_in.len - 1]);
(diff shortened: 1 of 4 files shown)


sm2n:
Tagging @matklad

src/lsm/table.zig
```diff
@@ -677,13 +677,16 @@ pub fn TableType(
             if (result.exact) {
                 const value = &values[result.index];
                 if (constants.verify) {
+                    for (values[0..result.index]) |*value_before| {
+                        assert(compare_keys(key, key_from_value(value_before)) != .eq);
+                    }
```
matklad: I am wondering whether this slows down things in `verify` mode too much?

@batiati, I _think_ you hit the "O(N) binary search if `constants.verify`" problem. Do you remeber which "benchmark" shows the slowdown, so that we can check it here?

matklad: An obvious ask could be "well, but we do full verification on non-found path anyway", but I think we might not be hitting that all that often? 

Though, yeah, this feels very inconsistent, not yet sure what's the right call.

src/lsm/schema.zig
```diff
@@ -499,13 +499,12 @@ pub const TableData = struct {
...
     ) []align(16) const u8 {
         const header = header_from_block(data_block);
-        // TODO we should be able to cross-check this with the header size
-        // for more safety.
         const used_values: u32 = header.request;
         assert(used_values > 0);
         assert(used_values <= schema.value_count_max);
 
         const used_bytes = used_values * schema.value_size;
+        assert(@sizeOf(vsr.Header) + used_bytes == header.size);
```
matklad, line 507: As a tripple-check, let's also look at `padding_offset` here to make sure we are not stepping into the paddign bytes.

src/lsm/compaction.zig
```diff
@@ -511,6 +511,11 @@ pub fn CompactionType(
                 // Assert that we're reading data blocks in key order.
                 const values_in = compaction.values_in[index];
                 if (values_in.len > 0) {
+                    if (constants.verify) {
+                        for (values_in[0 .. values_in.len - 1], values_in[1..]) |*value, *value_next| {
```
matklad: I was wondering whether this also changes the overall complexity here, like with the binary search, but I think this is fine. We do this once for data block, so this is O(N). 

Though, I am wondering --- why we need `values_in.len > 0` conditional? I would assume that data blocks are always not-empty

src/stdx.zig
```diff
@@ -141,7 +141,6 @@ test "disjoint_slices" {
     try std.testing.expectEqual(false, disjoint_slices(u32, u8, b, std.mem.sliceAsBytes(b)));
 }
 
-// TODO(Performance): Iterate over words.
```
matklad, line 144: Let's add a godbolt link (I tihnk one can get a shortened one?) as a comment here, as every person will be wondering how well this vectorizes. For godbolt, 

```

pub export fn f(bytes_ptr: [*]const u8, bytes_len: usize) bool {
    return zeroed(bytes_ptr[0..bytes_len]);
}
```

might be a simpler trick to cause the compiler to generate code, without including the whole main with the allocator and such. 

What we could perhaps do here as well is to take `[]const T`, and then take advantage of `T`'s size/alignment to generate tighter code wioutht `_split`, but yeah, this stone probably doesn't warrant more squeezing.

matklad (commented):
LGTM, excetp that linear search assert. I feel like it _should_ make our tests prohibitively expensive (and, if it doesn't, we are missing some tests)

src/stdx.zig
sm2n: I added a godbolt link. Good tip on `export fn` it's a lot cleaner.

src/lsm/compaction.zig
sm2n: I switched the branch to an assert after the discussion. The {tree, forest} fuzzers appear to still pass, so it seems like it's good.

src/lsm/schema.zig
sm2n: Added. It's not as useful as I thought though because if we calculate the real padding it would end up being a (syntactic) tautology. But it's still good to check.

src/lsm/table.zig
sm2n: I'm going to pull this out for now and see if I can add this check as a separate fuzzer. That way it doesn't slow all the other tests down.

matklad (approved):

src/lsm/table.zig
batiati: > @batiati, I _think_ you hit the "O(N) binary search if `constants.verify`" problem. Do you remeber which "benchmark" shows the slowdown, so that we can check it here?

Sorry, I missed this notification!

It was the fuzz_tree and fuzz_forest tests where I noticed a huge slowdown caused by O(N) in binary search when `constants.verify == true`.
For reference https://github.com/tigerbeetle/tigerbeetle/pull/989#discussion_r1266670365

PR #1170, opened 2023-08-29
Node client support on Windows

batiati:
This PR is a continuation of #1154 

- Adds `win32` to `platformMap` fixing the error `Unsupported platform: win32`.
- Fixes the `.lib` file that was generated pointing to node.dll instead of node.exe, causing the `dlopen` to fail trying to load `node.dll` that does not exist.
- Converts the `sh` scripts to `js` for cross-platform compatibility.
- Adds `windows-latest` to the CI tests.

### DIFF
--- a/src/clients/node/README.md
+++ b/src/clients/node/README.md
@@ -11,7 +11,7 @@ The TigerBeetle client for Node.js.
 ### Prerequisites
 
 Linux >= 5.6 is the only production environment we
-support. But for ease of development we also support macOS. Windows is not yet supported.
+support. But for ease of development we also support macOS and Windows.
 * NodeJS >= `14`
 
 ## Setup
@@ -615,4 +615,14 @@ npm pack
 
 ### On Windows
 
-Not yet supported.
+In PowerShell run:
+
+```console
+git clone https://github.com/tigerbeetle/tigerbeetle
+cd tigerbeetle
+git submodule update --init --recursive
+.\scripts\install_zig.bat
+cd src/clients/node
+npm install --include dev
+npm pack
+```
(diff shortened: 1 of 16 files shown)


src/clients/node/docs.zig
```diff
@@ -88,7 +88,10 @@ pub const NodeDocs = Docs{
     ,
 
     .install_commands = "npm install tigerbeetle-node",
-    .build_commands = "npm install typescript @types/node && npx tsc --allowJs --noEmit main.js",
+    .build_commands =
+    \\npm install typescript @types/node
+    \\npx tsc --allowJs --noEmit main.js
```
eatonphil, line 93: If you make a change to these docs.zig files, the build will fail until you:

```
zig/zig build client_docs -- --no-validate
```

as mentioned here https://github.com/tigerbeetle/tigerbeetle/tree/main/src/clients#documentation

batiati: Thanks @eatonphil, you saved me some time figuring it out!

One problem we have is that the node client is running on Windows, but it can´t build on Windows yet because of the `sh` scripts used in `package.json`.
I'm just checking if it's easy enough make them cross-platform scripts, otherwise for now we can partially support running but not building on Windows.

src/clients/node/README.md
```diff
@@ -21,7 +21,7 @@ First, create a directory for your project and `cd` into the directory.
 Then, install the TigerBeetle client:
 
 ```console
-npm install tigerbeetle-node
+npm install tigerbeetle-node@latest
```
eatonphil: If we encourage installing any version, I don't think we should encourage installing a symbolic version versus a tagged version since we require the client and server versions to line up.

batiati: That's true, ideally, we should update the documents on each build to include the current tag.
I added `@latest` because otherwise, it would install version `0.12.52`, which is too old to run the samples on Windows.

Instead, should I update the `package.json` file for all samples?

kprotty (approved):

PR #1172, opened 2023-08-29
LSM: Unified manifest

sentientwaffle:
## Summary

- Store tree id in TableInfo in the manifest log blocks on disk.
- Store `TableInfo` of every tree in each manifest log block.
- Move the `ManifestLog` from one-per-`Tree` to one-per-`Forest`.

#### Results

- Average case manifest log writes for a checkpoint: 21 blocks to 2 or 3 blocks (ballpark).
- Worst-case manifest log writes for a compaction half-bar: 63 blocks to 6 blocks.
- Worst-case manifest log writes for a checkpoint: 63 blocks to 6 blocks.
- Memory owned by manifest log: 63 blocks to 6 blocks.
- `state_machine.checkpoint()` latency drops by ~70% in the benchmark (on my laptop). (Note that this is an _extremely_ small sample size, though -- the benchmark only checkpoints once per run.)

<details>
<summary>(Checkpoint latency samples)</summary>

(Measuring `state_machine.checkpoint()` from invocation to callback.)

```
main:
0.041102385s
0.044551919s
0.033833146s
0.033007621s

unified manifest:
0.010172937s
0.012081376s
0.009965735s
```
</details>

## Context

Each tree's `TableInfo` is currently (abbreviated):

    fn TableInfoType(Key) type
        return extern struct {
            checksum: u128,
            address: u64,
            flags: u64, // (Currently unused.)
            snapshot_min: u64,
            snapshot_max: u64,
            key_min: Key,
            key_max: Key,
        };
    }

Some relevant constants:

    block_size                             = 64KiB
    lsm_batch_multiple                     = 64 ops
    journal_slot_count                     = 1024 ops
    storage_size_max                       = 16TiB
    tree_count                             = 21 trees
    lsm_levels                             = 7
    bars_per_checkpoint                    = 16 = 1024 / 64

    // Worst-case: fully-loaded LSM, with every possible compaction running every bar:
    compactions_pre_bar_per_tree           = 112  = lsm_levels * bars_per_checkpoint
    compactions_per_bar                    = 2352 = lsm_levels * bars_per_checkpoint * tree_count

    // This is a pretty arbitary constant – we can tune it up or down (preferrable down) as necessary.
    // But it is part of the disk layout, so we need to pick a value and stick with it.
    // (See superblock_trailer_manifest_size_max.)
    maximum superblock manifest size       = 36_184_064B = ~34.5MiB

And currently:

    SuperBlock.Manifest.BlockReferenceSize = 40B = @sizeof(tree_id) + @sizeof(checksum) + @sizeof(address)

    // Pretending that every manifest log block is full (which they aren't in practice – see below):
    maximum manifest log blocks            = ⌊36_184_064B / 40B⌋ = 904_601 blocks
    maximum manifest log bytes             = 904_601 * block_size = ~55.2GiB
    // Pretending that all Keys are 16B (as they are for the majority of trees).
    maximum manifest updates               = 723_680_800 updates = 800 updates/block * 904_601 blocks

    Trees:
    | @sizeof(Tree.Key) | @sizeof(TableInfo) | manifest updates/block | Tree Count |
    |                8B |                64B |                    992 |          4 |
    |               16B |                80B |                    800 |         12 |
    |               32B |               112B |                    576 |          5 |

However, the _updates-per-block_ numbers (and thus _maximum manifest updates_) above are a misleading ballpark:

- Different trees have different manifest log blocks.
- Different levels of the same tree have the same manifest log blocks.
- Lower bound: the LSM writes one manifest log block per tree (that is written to during the checkpoint) per checkpoint. (21 trees → 21 blocks written/checkpoint).
- Not counting manifest log compaction, each of those blocks has _at most_ 112 (`compactions_pre_bar_per_tree`) blocks – a write/space amplification of ~7 (`TableInfo/block ÷ compactions_pre_bar_per_tree`).
- Usually we don't compact every tree of every level. So the blocks are even emptier! (More space/write amplification).
- The amplification isn't a huge deal, but those 21 writes are a lower bound, and they occur during the checkpoint, contributing to the checkpoint latency spike.
- Additionally, the incomplete manifest log blocks increase the amount of manifest log compaction work outside of checkpoints.

## Unified Manifest Log

Instead of dividing the manifest log between trees, we could put all updates to all trees in the same manifest log blocks.

For this to work, `TableInfo` would need:

- To include a tree identifier (recently migrated from `u128` to `u16`).
- To accommodate any table, the `Key`s must be padded to the largest key size of any tree (`32B`).
- Then `@sizeOf(TableInfo) == 112B` _for every tree_.
- Lower bound: the LSM writes one manifest log block total per checkpoint.

On the superblock manifest side:
- No need to store the `tree_id` in the superblock manifest anymore, so `SuperBlock.Manifest.BlockReferenceSize == 24B` (down from `40B`).

Then:

    // Using the same value as a base, then working backwards:
    maximum manifest updates               = 723_680_800 updates
    maximum manifest log blocks            = 722_680_800 updates ÷ 576 updates/block = 1_256_390 blocks

    // 24B = @sizeof(checksum) + @sizeof(address)
    SuperBlock.Manifest.BlockReferenceSize = 24B
    maximum superblock manifest size       = 1_256_390 blocks * 24 bytes/block = 30_153_360B = 28.8MiB

This means that despite paying for the largest key size for trees with small keys, the overall superblock manifest size shrinks as we drop the tree id.
(Note that this is because we currently encode tree id as a `u128` even though it fits in a `u16`.)

#### Pros

- Smaller superblock manifest upper bound.
- Smaller superblock manifest in practice (due to less space amplification).
  - This means less data for checkpoints to write.
  - This means a little less data for state sync to move.
- Lower manifest log space/read/write amplification.
  - This reduced the amount of work necessary when starting the replica (i.e. opening the LSM).
- Fewer manifest log writes, especially at checkpoint.
- Faster recovery (fewer, denser manifest blocks to read).
- Lower `ManifestLog` RAM usage (approximately `21*3*64KiB ≈ 4MiB` → `6*64KiB = 384KiB`).
- De-specialize `ManifestLog`.

#### Cons

- We can't add a tree with a key larger than `32B`.
  - (Well, we can, but we would need to re-version the manifest block format).
  - (For secondary indexes, that means the maximum is `8B timestamp + 24B indexed key`).
  - (If we anticipate one a larger key, we could pad in advance, but the above calculations all assume 32B is the limit.)
- Unified manifest log compaction can't separate hot and cold manifests – all of the entries in a manifest log block are rewritten when it is compacted.
  - (So if one tree is modified infrequently, while the other is modified often, rewriting the cold tree's manifest entries is now write amplification).

---

<details>
<summary>Unified Manifest Level?</summary>

## Unified Manifest Level?

All of the above is entirely about the manifest _on disk_.

For completeness: **What about the in-memory manifest?**

Two options:
1. Specialized. (This is what we already have implemented.)
2. Unified.

I lean towards _specialized manifest levels_.
But either way, the decision doesn't impact the storage format, so it is flexible in the future.

### Specialized

Right now we have per-tree `TableInfo`, with a per-tree specialized `ManifestLevel`.

#### Pros
- This is more RAM-efficient – no overhead due to padding keys up to 32B.
- (Also, this is already implemented.)

#### Cons
- We need separate `TableInfo` type(s) from the `TableInfo` that is serialized into the manifest log blocks. (`schema.TableInfo` vs `ManifestLevel(Tree).TableInfo`)
- Per-tree specialization of `ManifestLevel` (i.e. more code generation).

### Unified

Change `ManifestLevel` (and `SegmentedArray`) to take a `*const fn` for `compare_keys` (instead of a `comptime fn callconv(.Inline)`).

#### Pros
- No pre-tree specialization of `ManifestLevel` (i.e. less code generation).

#### Cons
- More RAM overhead – tables with 8B or 16B keys must still pad their keys to 32B.
- `compare_keys()` can't be inlined.
- My guess is that despecializing `SegmentedArray` will have a noticeable (CPU) performance impact when searching levels with many tables.

</details>


matklad:
> We can't add a tree with a key larger than 32B

Here, could we in theory make manifest log entries variable in size? I think we can --- the data structure here is a log, it is processed sequentially and doesn't need random access. If we can make entries var-sized, then yeah, I don't think are any forward compatibility concerns here --- if at some point we decide to change `TableInfo` for a particular table size, we will be able to do that.

sentientwaffle:
> Here, could we in theory make manifest log entries variable in size? I think we can --- the data structure here is a log, it is processed sequentially and doesn't need random access. If we can make entries var-sized, then yeah, I don't think are any forward compatibility concerns here --- if at some point we decide to change `TableInfo` for a particular table size, we will be able to do that.

:thinking: I think I see what you mean. As an added bonus, only 5/21 of our trees use the maximum key size -- so in practice we could pack more than 576 (variable-sized) `TableInfo`s into each block. (We still have to allocate enough blocks to handle the worst (maximum-key-size) case.)

But: It would make APIs more complicated, since we can't have a `schema.ManifestLog.TableInfo` anymore -- we would need to pass + encode the tree's specialized `TableInfo`s.

Also the manifest log currently schedules via `entry_count < entry_count_max`. But with variable-sized `TableInfo`s, `entry_count_max` isn't that useful -- we would rather just check whether the next `TableInfo` can fit into the current block.

And when computing `blocks_count_max` (the maximum number of blocks that the manifest log must ever buffer) we need to assume the worst case, where every `TableInfo` has the largest key size.

Overall I'm not sure whether this is worth it right now, since we don't anticipate requiring larger keys.

matklad:
Yeah, we definitely don't need this right now. I am thinking more from an angle of "would we regret unified manifest if, at some point, we would like to have table-specific log entries of different sizes". Seems the answer is no, because you could have _both_ unified manifest, and also table-specific log entries.

src/lsm/forest.zig
```diff
@@ -57,16 +57,16 @@ pub fn ForestType(comptime Storage: type, comptime groove_cfg: anytype) type {
...
     {
         // Verify that every tree id is unique.
-        comptime var ids: []const u128 = &.{};
+        comptime var ids: []const u16 = &.{};
 
         inline for (std.meta.fields(_Grooves)) |groove_field| {
             const Groove = groove_field.type;
 
             for (std.meta.fields(@TypeOf(Groove.config.ids))) |field| {
                 const id = @field(Groove.config.ids, field.name);
```
matklad, line 69: could also check that `id > 0` as per our usual "make `mem.ZeroInit` to be an invalid object".

src/lsm/schema.zig
```diff
@@ -220,11 +220,13 @@ pub const TableIndex = struct {
...
     }
 
-    pub fn tree_id(index_block: BlockPtrConst) u128 {
+    pub fn tree_id(index_block: BlockPtrConst) u16 {
         const header = header_from_block(index_block);
         assert(BlockType.from(header.operation) == .index);
 
-        return @as(Parent, @bitCast(header.parent)).tree_id;
+        const parent = @as(Parent, @bitCast(header.parent));
+        assert(stdx.zeroed(&parent.reserved));
```
matklad, line 235: Excellent catch!

src/ring_buffer.zig
```diff
@@ -87,6 +87,14 @@ pub fn RingBuffer(
             return &self.buffer[(self.index + self.count - 1) % self.buffer.len];
         }
 
+        pub fn get(self: *Self, index: usize) ?T {
```
matklad: could be const? Ah, no, because we use get_ptr in the impl, and that one can't be const.

src/ring_buffer.zig
```diff
@@ -33,15 +38,15 @@ pub fn RingBuffer(
                     return .{ .buffer = undefined };
                 }
             },
-            .pointer => struct {
-                pub fn init(allocator: mem.Allocator) !Self {
-                    const buffer = try allocator.create([count_max]T);
-                    errdefer allocator.destroy(buffer);
+            .slice => struct {
+                pub fn init(allocator: mem.Allocator, capacity: usize) !Self {
```
matklad, line 42: I am wondering if we should bad `capacity == 0`? The code handles it, and _compile time_ capacity zero probably makes sense to me, but runtime empty capacity feels suspicions.

src/lsm/compaction.zig
```diff
@@ -271,6 +271,8 @@ pub fn CompactionType(
             if (compaction.grid_reservation) |grid_reservation| {
                 compaction.context.grid.forfeit(grid_reservation);
                 compaction.grid_reservation = null;
+            } else {
+                assert(compaction.move_table);
```
matklad, line 274: Shouldn't this also `compaction.move_table = false` to be a clean reset?

src/lsm/forest.zig
```diff
@@ -155,101 +203,205 @@ pub fn ForestType(comptime Storage: type, comptime groove_cfg: anytype) type {
...
+        fn manifest_log_open_event(
+            manifest_log: *ManifestLog,
+            level: u7,
+            table: *const schema.ManifestLog.TableInfo,
+        ) void {
+            const forest = @fieldParentPtr(Forest, "manifest_log", manifest_log);
+            assert(forest.progress.? == .open);
+            assert(level < constants.lsm_levels);
+
+            // TODO Is there a faster (or cleaner) way to map tree-id to the tree?
```
matklad: Wasn't able to do a working poc in a few minutes, but I think something of this shape should be possible: 

```
/// Given a comptime tree id, lookup the tree type.
/// This is basically lookup in tree_infos
fn TreeType(tree_id: comptime_int) type

/// Given a comptime tree id and a runtime forset, return a pointer to the tree
/// This is `inline for` we already we have in manifest_log_open
fn tree(forest: *Forest, tree_id: comptime_int) *TreeType(tree_id) {}

/// The largest tree ID we have, for loop over tree infos.
const tree_id_max: comptime_int = ...

fn manifest_log_open_event(table) {
    switch (table.tree_id) {
        // key idea -- covert to comptime-known tree id here
        inline 1..tree_id_max => |tree_id| tree(forest, tree_id).open_table()
        else => @panic()
    }

}

```

src/ring_buffer.zig
sentientwaffle: :thinking: Ah, it is probably worth inlining that logic so that it can be `*const Self`, since it is just a few lines.

src/lsm/compaction.zig
sentientwaffle: `move_table` will be cleaned up (along with most other fields) when the next compaction starts (`Compaction.start()`, `compaction.* = ...`). Maybe instead `compaction.manifest_entries.clear();` should be removed from `transition_to_idle` -- that way the only thing that gets reset there is the `state` and the grid reservation (which needs to be forfeited at that point).

src/lsm/forest.zig
sentientwaffle: Does that make a difference though? I would expect `switch (...) { inline 1..tree_id_max => ... }` to compile down to essentially the same chain of `if (id == 1) { ... }; if (id == 2) { ... } etc`.

matklad: I think inline case is guaranteed to be a jump table. Here's a debug example in godbolt, where `inline for` generates an if-chain, and `inline case` gets a jump table (in release for this small example, both get a jumptable):

https://godbolt.org/z/Eq5EcsPPn

src/lsm/manifest.zig
```diff
@@ -70,35 +65,71 @@ pub fn TableInfoType(comptime Table: type) type {
...
                 compare_keys(table.key_min, other.key_min) == .eq and
                 compare_keys(table.key_max, other.key_max) == .eq;
         }
+
+        pub fn decode(table: *const TableInfo) TreeTableInfo {
+            assert(stdx.zeroed(&table.reserved));
+
+            const key_min = std.mem.bytesAsValue(Key, table.key_min[0..@sizeOf(Key)]);
+            const key_max = std.mem.bytesAsValue(Key, table.key_max[0..@sizeOf(Key)]);
+
```
matklad, line 102: assert `key_min <= key_max`. And probably on the encode path as wel for simmetry?

src/lsm/tree.zig
```diff
@@ -658,26 +658,28 @@ pub fn TreeType(comptime TreeTable: type, comptime Storage: type) type {
...
-            tree.manifest.open(manifest_open_callback);
+            tree.manifest.open_commence(manifest_log);
         }
 
-        fn manifest_open_callback(manifest: *Manifest) void {
-            const tree = @fieldParentPtr(Tree, "manifest", manifest);
-            assert(tree.open_callback != null);
+        pub fn open_table(tree: *Tree, level: u8, table: *const schema.ManifestLog.TableInfo) void {
+            assert(tree.compaction_op == null);
+            assert(tree.key_range == null);
```
matklad, line 671: `assert table.tree_id == tree.id`

src/lsm/tree.zig
```diff
@@ -658,26 +658,28 @@ pub fn TreeType(comptime TreeTable: type, comptime Storage: type) type {
...
+        }
 
+        pub fn open_complete(tree: *Tree) void {
+            assert(tree.compaction_op == null);
             assert(tree.key_range == null);
-            tree.key_range = manifest.key_range();
-            maybe(tree.key_range == null);
 
-            callback(tree);
+            tree.compaction_op = tree.grid.superblock.working.vsr_state.commit_min;
```
matklad, line 680: Let's assert that `tree.compaction_op` points at the bar boundary. 

And let's assert here that the manifest is consistent (no interior empty levels, key ranges are disjoint for the latest snapshot, etc).

sentientwaffle: :thinking: What do you mean by "interior empty levels"?

matklad: I think that if level i is not empty, then level i-1 can’t be empty. That is, empty levels are always a suffix of the atray

sentientwaffle: Ah, I see! I think we can only safely check this for `TableUsage.general` tables (object and id trees):

```zig
            switch (Table.usage) {
                // Interior levels are non-empty.
                .general => {
                    var empty: bool = false;
                    for (&manifest.levels) |*level| {
                        var level_iterator =
                            level.iterator(.visible, &.{snapshot}, .ascending, null);
                        if (level_iterator.next()) |_| {
                            assert(!empty);
                        } else {
                            empty = true;
                        }
                    }
                },
                // In the secondary index TableUsage, it is possible (albeit unlikely!) that every
                // table in an interior level is deleted.
                //
                // Unlike general-usage tables, secondary-index tombstones need not compact down to
                // the last level of the tree before they are deleted. (Rather, the tombstones are
                // deleted as soon as they merge with their corresponding "put").
                // In this way, enough object deletions may lead to compactions where the both input
                // tables entirely cancel each other out, and no output table is written at all.
                // See `TableUsage` for more detail.
                .secondary_index => {},
            }
```

src/lsm/forest.zig
```diff
@@ -155,101 +208,216 @@ pub fn ForestType(comptime Storage: type, comptime groove_cfg: anytype) type {
...
+            table: *const schema.ManifestLog.TableInfo,
+        ) void {
+            const forest = @fieldParentPtr(Forest, "manifest_log", manifest_log);
+            assert(forest.progress.? == .open);
+            assert(level < constants.lsm_levels);
+
+            const tree_id_max: u16 = comptime tree_id_max: {
+                var tree_id_max: u16 = 0;
+                for (tree_infos) |tree_info| tree_id_max = @max(tree_id_max, tree_info.tree_id);
+                assert(tree_id_max == tree_infos.len); // There are no gaps in the tree ids.
```
matklad: Do we assert that ids are non-zero somewhere?  I guess we can make this more general and compute both `tree_id_max` and `tree_id_min`?

src/lsm/forest.zig
```diff
@@ -155,101 +208,216 @@ pub fn ForestType(comptime Storage: type, comptime groove_cfg: anytype) type {
...
-            // Start a compacting join.
-            const Join = JoinType(.compacting);
-            Join.start(forest, callback);
+            assert(forest.progress == null);
+
+            inline for (std.meta.fields(Grooves)) |field| {
+                @field(forest.grooves, field.name).compact(compact_groove_callback(field.name), op);
+            }
+
+            if (op % @divExact(constants.lsm_batch_multiple, 2) == 0) {
```
matklad, line 291: We have `half_bar_beat_count` defined in tree.zig, should either use that, or inline it. 

(also, `const half_bar_beat_count = @divExact(constants.lsm_batch_multiple, 2);` suggests `bar_beat_count` for `lsm_batch_multiple`)

src/lsm/forest.zig
```diff
@@ -155,101 +208,216 @@ pub fn ForestType(comptime Storage: type, comptime groove_cfg: anytype) type {
...
+                forest.manifest_log_progress = .idle;
+            }
+
+            const callback = progress.callback;
+            forest.progress = null;
+            callback(forest);
+        }
+
+        fn GrooveFor(comptime groove_field_name: []const u8) type {
+            return @TypeOf(@field(@as(Grooves, undefined), groove_field_name));
```
matklad: std.meta.FieldType would work?

src/lsm/schema.zig
```diff
@@ -509,3 +520,122 @@ pub const TableData = struct {
...
+        level: u7,
+        event: enum(u1) { insert, remove },
+
+        comptime {
+            assert(@bitSizeOf(Label) == @sizeOf(Label) * 8);
+        }
+    };
+
+    /// See manifest.zig's TreeTableInfoType declaration for field documentation.
+    pub const TableInfo = extern struct {
```
matklad, line 573: In theory, we can benefit from SOA layout on the `address` field here.

src/lsm/schema.zig
```diff
@@ -509,3 +520,122 @@ pub const TableData = struct {
...
+
+/// A ManifestLog block's body is a SoA of Labels and TableInfos.
+/// Each Label/TableInfo pair is an "entry".
+// TODO Store timestamp (snapshot) in header.
+pub const ManifestLog = struct {
+    const entry_size = @sizeOf(Label) + @sizeOf(TableInfo);
+    const entry_count_max_unaligned = @divFloor(block_body_size, entry_size);
+
+    pub const entry_count_max =
+        @divFloor(entry_count_max_unaligned, @alignOf(TableInfo)) * @alignOf(TableInfo);
```
matklad: Maybe add `entry_count_max_unaligned * @sizeOf(Label)`, which would work for labels on any size. This code currently implictily assumese that label is  single byte, and that's a bit surprisign to read.

src/lsm/schema.zig
```diff
@@ -509,3 +520,122 @@ pub const TableData = struct {
...
+        assert(labels_size_max % @alignOf(TableInfo) == 0);
+        assert(tables_size_max % @alignOf(TableInfo) == 0);
+
+        // Bit 7 is reserved to indicate whether the event is an insert or remove.
+        assert(constants.lsm_levels <= std.math.maxInt(u7) + 1);
+
+        assert(@sizeOf(Label) == @sizeOf(u8));
+
+        // All TableInfo's should already be 16-byte aligned because of the leading checksum.
+        const alignment = 16;
```
matklad, line 549: assert `alignment >= @sizeOf(vsr.Header)`.

src/lsm/schema.zig
matklad: Alternatively, maybe layout stuff in the order of decreasing alignment? There won't be any gaps that way, so no adjustments are necessary. 

I think this migh be a good rule-of-thumb for all on-disk SoA layouts?

matklad: Ahhh, so the current layout is actually tricky --- we always store `entry_max` labels, but only as much table infos as we need. It seems that we are trying to do both: 

* use simple layout, where the tables are at the constant offset from the begining of the block
* optimize for partially filled blocks

I think I am slightly in favor of maybe changeing the layout to make all arrays varlength, so it's more unfiform.

jorangreef: > Alternatively, maybe layout stuff in the order of decreasing alignment? There won't be any gaps that way, so no adjustments are necessary.

That's generally a good rule. However, sometimes we trump that intentionally. For example, packing fields of a `user_data/reserved` nature together contiguously so that they can be modified by flags in future and regrouped, or to put "well known" fields first.

src/lsm/forest.zig
sentientwaffle: > Do we assert that ids are non-zero somewhere?

There is `assert(id > 0);` in the `// Verify that every tree id is unique.` block.

>  I guess we can make this more general and compute both tree_id_max and tree_id_min?

:+1:

src/lsm/schema.zig
sentientwaffle: > I think I am slightly in favor of maybe changeing the layout to make all arrays varlength, so it's more unfiform.

I think the gotcha here is that we construct manifest log blocks incrementally. Right now we can do that in-place with no extra copies (i.e. writing the label and entry directly to the memory of the block we are building), since the two arrays (labels and tableinfos) are in fixed positions/offsets within the block body. But if the labels array (or whichever array comes first) is variable-length, then that breaks down.

sentientwaffle: (Not directly related to this thread, but I think I will directly store the `entry_count: u32` in the manifest log block header's `context` field, rather than deriving it from `size`.)

matklad: Ahh, yeah. I guess this could be added as a comment to motivate the current layout? 

But maybe we could do this:

* use vararray soa layout in the order of decrease alignment. 
* construct the block incrementally, _assuming that its going to be full_. 
* when the block is finished and it is not full, `copy_left` to compact it before writing

sentientwaffle: Hmm yeah that should work -- we only write at most 1 partial manifest log blocks per wrap of the log (at checkpoint).

src/lsm/schema.zig
sentientwaffle: :thinking: Why `>=` instead of `alignment <= @sizeOf(vsr.Header)`? As in: "tables should not be aligned more strictly than the headers that offset them"?

matklad: :rofl: was thinking about alignment vs log-of-alignment, and mixed up comparison direction instead!

src/lsm/schema.zig
sentientwaffle: :thinking: I'm not sure what you mean?

src/lsm/manifest_log.zig
```diff
@@ -266,9 +237,9 @@ pub fn ManifestLogType(comptime Storage: type, comptime TableInfo: type) type {
             const block_reference = manifest_log.read_block_reference.?;
             verify_block(block, block_reference.checksum, block_reference.address);
 
-            const entry_count = Block.entry_count(block);
-            const labels_used = Block.labels_const(block)[0..entry_count];
-            const tables_used = Block.tables_const(block)[0..entry_count];
+            const entry_count = schema.ManifestLog.entry_count(block);
+            const labels_used = schema.ManifestLog.labels_const(block)[0..entry_count];
+            const tables_used = schema.ManifestLog.tables_const(block)[0..entry_count];
```
matklad: tbh, not a fan of `_used` terminology: i'd rather use `tables` for `_used` version and `tables_buffer` for "potentially undefined". But this is consistent with other uses of this terminology!

src/vsr/superblock.zig
```diff
@@ -435,12 +435,12 @@ pub const superblock_trailer_size_max = blk: {
 // A manifest block reference of 40 bytes contains a tree hash, checksum, and address.
 // These references are stored in struct-of-arrays layout in the trailer for the sake of alignment.
 const superblock_trailer_manifest_size_max = blk: {
-    assert(SuperBlockManifest.BlockReferenceSize == 16 + 16 + 8);
```
matklad, line 438: comment needs updating

src/lsm/forest.zig
sentientwaffle: I like `constants.lsm_bar_beat_count` and `constants.lsm_half_bar_beat_count`, but I'll defer it to a follow-up PR for ease of review.

src/lsm/schema.zig
matklad: Sorry, GitHub ate half of the comment. When processing manifest log entries, we skip over entries for addresses we have seen “in the future”, because we know they are outdated. If addresses were a separate array, then we could skip loading other files of table info in cache for skipped over entries

src/lsm/manifest_log.zig
sentientwaffle: :thinking: I don't love `buffer` since it implies to me that we are getting back bytes, not tables/labels/etc What about `_free`?

src/lsm/schema.zig
sentientwaffle: Ah, I see

src/lsm/manifest_log.zig
matklad: 🤔 ”getting back bytes” isn’t exactly wrong here, because what we are getting here is raw memory indeed, it doesn’t necessary contain *valid* tables.

free doesn’t feel right, because *some* entries might not be free.

src/lsm/forest.zig
```diff
@@ -57,43 +60,93 @@ pub fn ForestType(comptime Storage: type, comptime groove_cfg: anytype) type {
...
 
-        join_op: ?JoinOp = null,
-        join_pending: usize = 0,
-        join_callback: ?Callback = null,
+        progress: ?union(enum) {
+            open: struct { callback: Callback },
+            checkpoint: struct { callback: Callback },
+            compact: struct {
+                op: u64,
+                pending: usize,
```
matklad: Comment what this is a counter for. I wonder if it also would be more clear to use a bitset here, to lean on "never lose the information" principle. Maybe even yes? That way we can assert that we don't count the same groove twice.

src/lsm/groove.zig
```diff
@@ -417,9 +414,10 @@ pub fn GrooveType(
         };
         const PrefetchObjects = std.HashMapUnmanaged(Object, void, PrefetchObjectsContext, 70);
 
-        join_op: ?JoinOp = null,
-        join_pending: usize = 0,
-        join_callback: ?Callback = null,
+        compacting: ?struct {
+            pending: usize,
```
matklad: Same here, comment what pending is, consider bitset.

src/lsm/manifest.zig
```diff
@@ -138,6 +168,8 @@ pub fn ManifestType(comptime Table: type, comptime Storage: type) type {
         };
 
         node_pool: *NodePool,
+        config: TreeConfig,
+        manifest_log: ?*ManifestLog = null,
```
matklad, line 174: This ["lateinit"](https://kotlinlang.org/docs/properties.html#late-initialized-properties-and-variables) field feels error prone. I guess we are doing it this way because we need a stable address for manifest log, and we only have that in open_commence, not in init? 

Perhaps a comment on the field is worth it, but also I am wondering if we could allocate the log struct on the heap in the init?

src/lsm/manifest_log.zig
```diff
@@ -266,25 +239,20 @@ pub fn ManifestLogType(comptime Storage: type, comptime TableInfo: type) type {
...
                 const label = labels_used[entry];
                 const table = &tables_used[entry];
 
-                if (manifest.insert_table_extent(
-                    manifest_log.tree_id,
-                    table.address,
-                    block_reference.address,
-                    entry,
-                )) {
+                if (manifest.insert_table_extent(table.address, block_reference.address, entry)) {
```
matklad, line 259: This gives me a pause here: if we have seen this address already, we might skip over bogus, non-existing `tree_id`s here. So I was going to suggest to check the `tree_id` proactively before the if.

But I am wondering if we perhaps should do some more thorough validation? Perhaps `TableExtent` should rember `tree_id` (or maybe the entire `TableInfo`) to make sure that insert/remove events are consistetn?

matklad: Yeah, I think we want to store `checksum`s in an extent, verify that checksums are consistent, and also proactively check tree id (so that we don't ignore removals of unknown tables)

src/lsm/manifest_log.zig
```diff
@@ -637,38 +574,62 @@ pub fn ManifestLogType(comptime Storage: type, comptime TableInfo: type) type {
...
+            assert(!manifest_log.reading);
+            assert(!manifest_log.writing);
+            assert(manifest_log.read_callback == null);
+            assert(manifest_log.write_callback == null);
+
+            if (manifest_log.grid_reservation) |grid_reservation| {
+                manifest_log.grid.forfeit(grid_reservation);
+                manifest_log.grid_reservation = null;
+            } else {
+                // Compaction was skipped for this half-bar.
```
matklad, line 630: Hm, would it be better to assert that reservation is not-null here, and let the caller special case skipped compaction. Am I correct that this is special case for the very first half-bar, `lsm_batch_multiple / 2 - 1`?

src/lsm/schema.zig
```diff
@@ -509,3 +521,137 @@ pub const TableData = struct {
...
+            block[@sizeOf(vsr.Header)..][0 .. schema.entry_count * @sizeOf(TableInfo)],
+        );
+    }
+
+    pub fn tables_used(schema: *const Manifest, block: BlockPtrConst) []const TableInfo {
+        return mem.bytesAsSlice(
+            TableInfo,
+            block[@sizeOf(vsr.Header)..][0 .. schema.entry_count * @sizeOf(TableInfo)],
+        );
+    }
```
matklad: These are two identical functions, no? Seems like we can remove the `_used` version?

matklad: Ahhhh, the differ in constness. That's pretty confusing, maybe `_const` then?

src/lsm/manifest_log.zig
```diff
@@ -437,142 +392,128 @@ pub fn ManifestLogType(comptime Storage: type, comptime TableInfo: type) type {
...
+            const block_index = manifest_log.writes_pending;
+            const block = manifest_log.blocks.get_ptr(block_index).?;
             verify_block(block.*, null, null);
 
+            const block_schema = schema.Manifest.from(block.*);
+            assert(block_schema.entry_count > 0);
+
             const header = schema.header_from_block(block.*);
-            const address = Block.address(block.*);
+            const address = header.op;
```
matklad, line 411: `fn close_block` misses this `const address = header.op` alias, might want to add it, as it does make the code quite a bit more clear.

src/lsm/schema.zig
```diff
@@ -509,3 +521,137 @@ pub const TableData = struct {
...
+        const header = header_from_block(manifest_block);
+        assert(header.command == .block);
+        assert(BlockType.from(header.operation) == .manifest);
+        assert(header.op > 0);
+
+        const context = @as(Context, @bitCast(header.context));
+        assert(context.entry_count > 0);
+        assert(context.entry_count <= entry_count_max);
+        assert(stdx.zeroed(&context.reserved));
+
```
matklad, line 614: assert that header.size is consistent with entry count.

src/lsm/schema.zig
```diff
@@ -509,3 +521,137 @@ pub const TableData = struct {
...
+        comptime {
+            assert(@bitSizeOf(Label) == @sizeOf(Label) * 8);
+        }
+    };
+
+    entry_count: u32,
+
+    pub fn from(manifest_block: BlockPtrConst) Manifest {
+        const header = header_from_block(manifest_block);
```
matklad, line 604: Unrelated, but in `header_from_block` we do `assert(header.size >= @sizeOf(vsr.Header));`. I wonder if that could be strict inequality --- do we ever use empty blocks for anything?

src/vsr/superblock_manifest.zig
```diff
@@ -375,90 +341,83 @@ pub const Manifest = struct {
...
 
     pub const IteratorReverse = struct {
         manifest: *const Manifest,
-        tree: u128,
         count: u32,
 
         pub fn next(it: *IteratorReverse) ?BlockReference {
             assert(it.count <= it.manifest.count);
 
             while (it.count > 0) {
```
matklad: This `while` could be and `if`?

src/lsm/manifest.zig
sentientwaffle: The reason not to heap allocate is the forest needs to be able to `@fieldParentPtr(Forest, "manifest_log", manifest_log)`. The alternative is to lazy-init a pointer within the `ManifestLog` itself to the `Forest` (e.g. "context") but that seems at least as bad, and additionally adds a dependency.

I'll add a comment.

src/lsm/manifest_log.zig
sentientwaffle: Good idea wrt checking the table id! I'll add that.

Including `checksum`s in the extent doesn't work, though: During `ManifestLog.append()` we need to create the table extent, but we don't yet know the block checksum (since the block in question isn't constructed yet).

src/lsm/manifest_log.zig
sentientwaffle: The reason for doing it this way (i.e. let manifest log handle the skip internally, rather than skip at a higher level) is that it mirrors `Tree.compact()`.

`Tree.compact()` must skip internally because it does work even when compaction is skipped (`tree.compact_mutable_table_into_immutable();`).

> Am I correct that this is special case for the very first half-bar, lsm_batch_multiple / 2 - 1?

It is for the first full bar _after recovery from checkpoint_ -- not just from op=0.

src/lsm/schema.zig
sentientwaffle: Nice! I don't think we ever use empty blocks. It might be redundant but what do you think of:
```zig
    assert(header.size >= @sizeOf(vsr.Header)); // Every block has a header.
    assert(header.size > @sizeOf(vsr.Header)); // Every block has a non-empty body.
```

src/lsm/forest.zig
```diff
@@ -328,7 +326,16 @@ pub fn ForestType(comptime Storage: type, comptime groove_cfg: anytype) type {
                     const grooves: *align(@alignOf(Grooves)) Grooves =
                         @alignCast(@fieldParentPtr(Grooves, groove_field_name, groove));
                     const forest = @fieldParentPtr(Forest, "grooves", grooves);
-                    forest.progress.?.compact.pending -= 1;
+
+                    inline for (std.meta.fields(Grooves), 0..) |groove_field, i| {
+                        if (std.mem.eql(u8, groove_field.name, groove_field_name)) {
```
matklad: :thinking: I think this generates a runtime comparison (which could be optimized away by LLVM, but I think the semantics here is runtime). 

I think we want rather something like:

```
const groove_index = comptime for ((std.meta.fields(Grooves), 0..)) |groove_field, i| {
    if (std.mem.eql(u8, groove_field.name, groove_field_name)) break i;
} else unreachable;
```

src/lsm/groove.zig
```diff
@@ -1028,6 +1031,20 @@ pub fn GrooveType(
             ids,
             objects,
             index: []const u8,
+
+            fn offset(field: TreeField) usize {
+                switch (field) {
+                    .objects => return 0,
+                    .ids => return 1,
+                    .index => |index_tree_name| {
+                        inline for (std.meta.fields(IndexTrees), 0..) |index_tree_field, i| {
```
matklad: Could be just a `for`

src/lsm/groove.zig
```diff
@@ -1007,6 +1008,9 @@ pub fn GrooveType(
                         },
                     };
 
+                    assert(groove.compacting.?.pending.isSet(tree_field.offset()));
+                    groove.compacting.?.pending.unset(tree_field.offset());
```
matklad: Needs `.unset(comptime tree_field.offset());`

matklad (dismissed):

sentientwaffle:
Rebased to fix conflict with `main`.

matklad (approved):

PR #1175, opened 2023-08-31
No REPL printing in tests

eatonphil:


src/tigerbeetle/repl.zig
```diff
@@ -611,8 +637,7 @@ pub fn ReplType(comptime MessageBus: type) type {
                     try repl.do_statement(statement);
                 }
             } else {
-                try display_help();
-                std.os.exit(0);
+                try repl.display_help();
```
eatonphil, line 640: It used to be the case that `statements` in the if () above this was an optional type. And this case would come only if the user typed in `--command=""` and left the flag blank.

After refactors though this ended up just being the actual REPL case. So only the --command="" approach to using the REPL worked (which is what the integration tests test).

Nobody can use the REPL without this and without --command="" as is.

sentientwaffle (approved):

PR #1177, opened 2023-09-01
FIFO: Assert !contains(elem) before push(elem)

sentientwaffle:
This is useful to be certain that e.g. `NextTick` or `Read` contexts are not used multiple times concurrently.


jorangreef:
Great to have this, and also gated by `verify`.

kprotty (approved):

PR #1179, opened 2023-09-07
VSR: State Sync (Part 4/?): Sync tables

sentientwaffle:
# Context

State sync allows a replica to "skip" past a bunch of commits.

Immediately after [superblock sync](https://docs.tigerbeetle.com/internals/sync#algorithm) completes:

- The replica can participate in consensus (i.e. prepare/commit, view changes, etc).
- Manifest blocks are already synced (via normal [grid repair](https://docs.tigerbeetle.com/internals/vsr/#protocol-repair-grid)).
- The replica is missing the client replies that were skipped over. (These are repaired by the [sync client replies](https://docs.tigerbeetle.com/internals/vsr/#protocol-sync-client-replies) protocol.)
- The replica is missing the LSM table blocks (index/filter/data) that would have been created by the commits that were skipped over.

Previously TigerBeetle repaired the LSM table blocks lazily. That is, it goes about its normal business executing commits and compaction, and when it tries to read a block belonging to a skipped table, it interprets the missing block as corruption and repairs it.

This approach works (and is quite simple!), but has several downsides:

- This lazy approach is slow – the replica will often be blocked waiting for blocks from the network, and due to access patterns these delays will be sequential, not parallel.
- This lazy approach restores cluster durability incrementally in incompletely – there is no guarantee that the replica will eventually repair all missing blocks.

This changeset supplements (but does not replace!) this lazy approach with a proactive sync of skipped tables.

# Design

In pseudocode (see `Replica.sync_request_tables()`):

    for table in replica.state_machine.forest: # See ForestTableIterator.
        if table.snapshot_min ∈ (superblock.sync_op_min … superblock.sync_op_max):
            # Fetch the index block first.
            index_block := request table index block from cluster
            # Write it to disk:
            grid.repair_block(index_block)

            # Request and repair every block referenced by the table index block.
            for content in index_block:
                content_block := request content block from cluster
                grid.repair_block(content_block)

    # Then when all tables are fully repaired:
    superblock.sync_op_min = 0
    superblock.sync_op_max = 0

# Future

Right now this still leans on grid-repair protocol, requesting each block by address/checksum.

In the future we can require just a table index address/checksum along with a (compressed) bitset of which content blocks we require, and attach that to e.g. pings.


docs/internals/testing.md
```diff
@@ -36,37 +36,39 @@ Documentation for (roughly) code in the `src/testing` directory.
...
+ 6   |       .          1V 127/127C  64:127Jo  0/_2J!  63:126Wo   <0:__0>   44857Gf  0G!   0G?
+11   |            .     1V 127/127C  64:127Jo  0/_0J!  64:127Wo   <0:__0>   44857Gf  0G!   0G?
+10 ^ |           h      1V   0/_11C  15:_15Jo 31/31J!   0:_15Wo   <0:__0>   45056Gf  0G!   0G?
+ 0   \ .                1V 127/127C  64:127Jo  0/_1J!  63:126Wo   <0:__0>   44857Gf  0G!   0G?
+ 4   \     .            1V 127/127C  64:127Jo  0/_1J!  63:126Wo   <0:__0>   44857Gf  0G!   0G?
+ 5   \      .           1V 127/127C  64:127Jo  0/_0J!  64:127Wo   <0:__0>   44857Gf  0G!   0G?
+ 3   \    .             1V 127/127C  64:127Jo  0/_1J!  63:126Wo   <0:__0>   44857Gf  0G!   0G?
+ 8   |         .        1V 127/127C  64:127Jo  0/_1J!  63:126Wo   <0:__0>   44857Gf  0G!   0G?
+10 < ~           h      1V   0/_11C  15:_15Jo 31/31J!   0:_15Wo   <0:__0>   45056Gf  0G!   0G?
+10 > |           h      1V 119/119C  15:_15Jo 31/31J!   0:_15Wo   <0:123>   44877Gf  0G!   0G?
```
sentientwaffle, line 73: Using `<min:max>` for the sync min/max since `<` and `>` denote the start and end of superblock sync in the "event" column.

docs/internals/vsr.md
```diff
@@ -49,8 +49,8 @@ Storage:
 |                `request_reply` | replica |      replica | [Repair Client Replies](#protocol-repair-client-replies), [Sync Client Replies](#protocol-sync-client-replies) |
 |                      `headers` | replica |      replica | [Repair Journal](#protocol-repair-journal)                                       |
 |                     `eviction` | primary |       client | [Client](#protocol-client)                                                       |
-|               `request_blocks` | replica |      replica | [Sync Grid](#protocol-sync-grid), [Repair Grid](#protocol-repair-grid)           |
-|                        `block` | replica |      replica | [Sync Grid](#protocol-sync-grid), [Repair Grid](#protocol-repair-grid)           |
+|               `request_blocks` | replica |      replica | [Sync Tables](#protocol-sync-tables), [Repair Grid](#protocol-repair-grid)       |
+|                        `block` | replica |      replica | [Sync Tables](#protocol-sync-tables), [Repair Grid](#protocol-repair-grid)       |
```
sentientwaffle, line 53: Renamed this protocol since it doesn't actually sync manifest blocks, only index/filter/data blocks.

src/lsm/schema.zig
```diff
@@ -308,9 +305,37 @@ pub const TableIndex = struct {
...
     }
 
-    inline fn blocks_used(index: *const TableIndex, index_block: BlockPtrConst) u32 {
-        return index_block_count + index.filter_blocks_used(index_block) +
-            data_blocks_used(index_block);
+    pub fn content_blocks_used(index: *const TableIndex, index_block: BlockPtrConst) u32 {
+        return index.filter_blocks_used(index_block) + index.data_blocks_used(index_block);
+    }
+
+    pub fn content_block(
```
sentientwaffle: Once we drop filter blocks we can get rid of these functions.

src/lsm/tree_fuzz.zig
```diff
@@ -598,8 +535,8 @@ pub fn main() !void {
         .read_latency_mean = 0 + fuzz.random_int_exponential(random, u64, 20),
         .write_latency_min = 0,
         .write_latency_mean = 0 + fuzz.random_int_exponential(random, u64, 20),
-        .read_fault_probability = random.uintLessThan(u8, 100),
-        .write_fault_probability = random.uintLessThan(u8, 100),
+        .read_fault_probability = 0,
+        .write_fault_probability = 0,
```
sentientwaffle, line 539: Imo testing corruption at this layer is more trouble than it's worth -- it is well covered by the VOPR.

src/testing/cluster/storage_checker.zig
```diff
@@ -171,5 +177,42 @@ pub fn StorageCheckerType(comptime Storage: type) type {
             }
             return checksum;
         }
+
+        fn checksum_grid(checker: *Self, superblock: *const SuperBlock) u128 {
```
sentientwaffle, line 182: At long last, the storage checker can test grid storage determinism at checkpoint! (When sync isn't running, that is.)

src/vsr.zig
```diff
@@ -1983,3 +1984,13 @@ pub const Checkpoint = struct {
...
+
+pub const Snapshot = struct {
+    /// A table with TableInfo.snapshot_min=S was written during some commit with op<S.
+    /// A block with snapshot_min=S is definitely readable at op=S.
+    pub fn readable_at_commit(op: u64) u64 {
+        // TODO: This is going to become more complicated when snapshot numbers match the op
+        // acquiring the snapshot.
+        return op + 1;
+    }
+};
```
sentientwaffle, line 1998: I don't love `readable_at_commit` for a name but couldn't figure out anything better. I am planning on reworking snapshots soon so hopefully will come up for something better then.

src/vsr/grid.zig
```diff
@@ -264,11 +276,14 @@ pub fn GridType(comptime Storage: type) type {
...
             grid.assert_only_repairing();
 
-            // TODO(Grid repair queue): This check is too conservative – we only need to wait for
-            // all released-block writes to finish.
-            if (grid.write_iops.executing() > 0) return;
+            // We are still repairing some blocks that were released at the checkpoint.
+            if (!grid.repair_queue.checkpoint_complete()) {
+                assert(grid.write_iops.executing() > 0);
+                return;
+            }
```
sentientwaffle: Before: Any in-progress repairs (i.e. syncing blocks) would stall the checkpoint until they finish.
Now: We are more precise -- the checkpoint only needs to stall for writes to blocks that are about to be released.

(The goal is that we never write to a block _after_ it has been released, since we may re-acquire and re-write it in the new checkpoint.)

src/vsr/grid_repair_queue.zig
```diff
@@ -0,0 +1,489 @@
...
+const BlockPtrConst = *align(constants.sector_size) const [constants.block_size]u8;
+
+/// Immediately after state sync we want access to all of the grid's write bandwidth to rapidly sync
+/// table blocks.
+const grid_repair_writes_max = constants.grid_iops_write_max;
+
+/// The maximum number of blocks that can possibly be referenced by any table index block.
+///
+/// - This is a very conservative (upper-bound) calculation that doesn't rely on the StateMachine's
+///   tree configuration. (To avoid prevent Grid from depending on StateMachine).
```
matklad: Let's maybe move this up to top-level `constants.zig`, and also asset, in state machines, that each particular index block fits into this upper bound.

src/vsr/grid_repair_queue.zig
```diff
@@ -0,0 +1,489 @@
+//! Track corrupt/missing grid blocks.
+//!
+//! - The GridRepairQueue is LSM-aware: it can repair entire tables.
```
matklad: Let's mention that it is non-generic (all types of tables share the same queue)

src/vsr/grid_repair_queue.zig
```diff
@@ -0,0 +1,489 @@
...
+
+const allocate_block = @import("./grid.zig").allocate_block;
+const GridType = @import("./grid.zig").GridType;
+const FIFO = @import("../fifo.zig").FIFO;
+const IOPS = @import("../iops.zig").IOPS;
+const BlockPtrConst = *align(constants.sector_size) const [constants.block_size]u8;
+
+/// Immediately after state sync we want access to all of the grid's write bandwidth to rapidly sync
+/// table blocks.
+const grid_repair_writes_max = constants.grid_iops_write_max;
```
matklad: This seems to be unused (together with the WriteSet) and duplicated in constants.zig. 

Also, what happens if we want to saturate writes via repairs, but also have a write from the commit path? Am I correct that, depending on timings, either commit or repair might proceed?

src/vsr/grid_repair_queue.zig
```diff
@@ -0,0 +1,489 @@
...
+    const FaultProgress = union(enum) {
+        /// Repair a single block.
+        block,
+        /// Repair the table and all of its content. Awaiting table index block.
+        table_index: TableIndex,
+        /// Repair the table and all of its content. Awaiting table content blocks.
+        table_content: TableContent,
+
+        const TableIndex = struct { table: *RepairTable };
+        const TableContent = struct { table: *RepairTable, index: usize };
```
matklad: Maybe `index: u32`, to avoid dependency on host architecture (the dimensionality of `index` is defined by our on-disk format, not the virtual memory size), and maybe to shrink the size of `FaultProgress`a bit, as I imagine there could be a fair amount of those in memory via faultyBlocks.

src/vsr/grid_repair_queue.zig
```diff
@@ -0,0 +1,489 @@
...
+        /// block-by-block.
+        content_blocks_received: TableContentBlocksSet = TableContentBlocksSet.initEmpty(),
+        /// This count includes the index block.
+        /// Invariants:
+        /// - table_blocks_written ≤ table_blocks_total
+        table_blocks_written: usize = 0,
+        /// When null, the table is awaiting an index block.
+        /// When non-null, the table is awaiting content blocks.
+        /// This count includes the index block.
+        table_blocks_total: ?usize = null,
```
matklad: Same here, consider using `u32` for counting table blocks.

src/vsr/grid_repair_queue.zig
```diff
@@ -0,0 +1,489 @@
+//! Track corrupt/missing grid blocks.
+//!
+//! - The GridRepairQueue is LSM-aware: it can repair entire tables.
+//! - The GridRepairQueue is "coherent" – that is, all of the blocks in the queue belong in the
+//!   replica's current checkpoint:
+//!   - The GridRepairQueue will not repair freed blocks.
+//!   - The GridRepairQueue will repair released blocks, until they are freed at the checkpoint.
```
matklad: Maybe mention a little bit about the flow here:

- that sync calls `request_table`
- that `request_block` is called either by the grid on the read path, or, asynchronously, by `request_table`

src/vsr/grid_repair_queue.zig
```diff
@@ -0,0 +1,489 @@
...
+        const faults_total = queue.faulty_blocks.count();
+        if (faults_total == 0) return 0;
+        assert(faults_total > queue.faulty_blocks_repair_index);
+
+        const fault_addresses = queue.faulty_blocks.entries.items(.key);
+        const fault_data = queue.faulty_blocks.entries.items(.value);
+
+        var requests_count: usize = 0;
+        var fault_offset: usize = 0;
+        while (requests_count < requests.len and
```
matklad: maybe `for(0..faults_total) |fault_offset|` and hadle `requests_count` separately?

src/vsr/grid_repair_queue.zig
```diff
@@ -0,0 +1,489 @@
...
+        queue.faulty_blocks.deinit(allocator);
+
+        queue.* = undefined;
+    }
+
+    /// When the queue wants more blocks than fit in a single request message, successive calls
+    /// to this function cycle through the pending BlockRequests.
+    pub fn block_requests(queue: *GridRepairQueue, requests: []vsr.BlockRequest) usize {
+        assert(!queue.canceling);
+
```
matklad: assert(requests.len > 0)

src/vsr/grid_repair_queue.zig
```diff
@@ -0,0 +1,489 @@
...
+                    .block_address = fault_addresses[fault_index],
+                    .block_checksum = fault_data[fault_index].checksum,
+                };
+                requests_count += 1;
+            }
+        }
+
+        queue.faulty_blocks_repair_index =
+            (queue.faulty_blocks_repair_index + fault_offset) % faults_total;
+        return requests_count;
```
matklad:
```zig
assert(requests_cont <= requests.len)
assert(requests_cont <= faults_total)
if (faults_total <= requests_len) assert(requests_cont <= faults_total)
```

src/vsr/grid_repair_queue.zig
```diff
@@ -0,0 +1,489 @@
...
+
+        var requests_count: usize = 0;
+        var fault_offset: usize = 0;
+        while (requests_count < requests.len and
+            fault_offset < faults_total) : (fault_offset += 1)
+        {
+            const fault_index =
+                (queue.faulty_blocks_repair_index + fault_offset) % faults_total;
+
+            const fault = &fault_data[fault_index];
```
matklad: do we need this alias? There's a single usage only

src/vsr/grid_repair_queue.zig
```diff
@@ -0,0 +1,489 @@
...
+            if (fault.state == .waiting) {
+                requests[requests_count] = .{
+                    .block_address = fault_addresses[fault_index],
+                    .block_checksum = fault_data[fault_index].checksum,
+                };
+                requests_count += 1;
+            }
```
matklad: I am wondering if there are any useful asserts we can add on the `else` branch here. This seems like a good place for assert, as it periodically sweeps through everything, but I don't think we can actually assert anything useful here?

src/vsr/grid_repair_queue.zig
```diff
@@ -0,0 +1,489 @@
...
+
+    /// Invariants:
+    /// - For every block address in faulty_blocks, ¬free_set.is_free(address).
+    faulty_blocks: FaultyBlocks,
+    /// Index within `faulty_blocks`, used to cycle through block-repair requests.
+    ///
+    /// Invariants:
+    /// - faulty_blocks.count() > 0 implies faulty_blocks_repair_index < faulty_blocks.count()
+    /// - faulty_blocks.count() = 0 implies faulty_blocks_repair_index = faulty_blocks.count()
+    faulty_blocks_repair_index: usize = 0,
```
matklad: Hm, I am slightly concerned with fairness here -- given that we swap_remove when a block is repaired, a particular unlucky block can dodge the sweep several times in a row. 

I think that's fine though:

- we guarantee that we'll request it eventually, because `swap_remove` always moves blocks to the left,  and the zeroth block is guaranteed to be requested
- I don't think we care about fairness all that much, as we must repair everything for the repair to be considered complete.

src/vsr/grid_repair_queue.zig
```diff
@@ -0,0 +1,489 @@
...
+        assert(!queue.canceling);
+
+        queue.faulty_blocks.deinit(allocator);
+
+        queue.* = undefined;
+    }
+
+    /// When the queue wants more blocks than fit in a single request message, successive calls
+    /// to this function cycle through the pending BlockRequests.
+    pub fn block_requests(queue: *GridRepairQueue, requests: []vsr.BlockRequest) usize {
```
matklad: Not the best name, as this actively mutates the state of the queue, rather than just being a query. "request" as in `BlockRequest` also clashes with `fn request_block`

Maybe something along the lines of:

* `enqueue_block` / `enqueue_table` / `enqueue_table_contents` for adding to the queue (or maybe just `push` as that's shorter, and, strictly speaking, we are not a queue
* `repair_commence` / `repair_complete` sgtm. `repair_ready` checks for `.waiting` state, so it feels fliped. Maybe `repair_in_progress` / `_scheduled` / `enqueued`?  
* `next_batch_of_block_requests` for seding requests

src/vsr/grid_repair_queue.zig
```diff
@@ -0,0 +1,489 @@
...
+    pub fn request_blocks_available(queue: *const GridRepairQueue) usize {
+        assert(!queue.canceling);
+        assert(queue.faulty_tables.count <= queue.options.tables_max);
+
+        const faulty_blocks_free =
+            queue.faulty_blocks.capacity() -
+            queue.faulty_blocks.count() -
+            queue.options.tables_max * lsm_table_content_blocks_max;
+        return faulty_blocks_free;
+    }
```
matklad: I am confused by the logic here. `queue.faulty_blocks.count()` could be equal to `queue.faulty_blocks.capacity()`, so `queue.faulty_blocks.capacity() - queue.faulty_blocks.count()` could be zero, and further subtracing `queue.options.tables_max *` could overflow. 

I don't think we can do what this function wants to do without maintaining some extra counter. If we have a certain amount of faulty blocks, we can't know which blocks are attributed to table repair, and which ones are due to explicit repair requests.

src/vsr/grid_repair_queue.zig
```diff
@@ -0,0 +1,489 @@
...
+
+    /// Queue a faulty block to request from the cluster and repair.
+    pub fn request_block(queue: *GridRepairQueue, address: u64, checksum: u128) void {
+        assert(!queue.canceling);
+        assert(queue.request_blocks_available() > 0);
+        assert(queue.faulty_tables.count <= queue.options.tables_max);
+
+        if (queue.faulty_blocks.getPtr(address)) |fault| {
+            assert(fault.checksum == checksum);
+            assert(fault.state != .aborting);
```
matklad: Can we assert that `.progress` here is not `.block`? I _think_ grid's `resolves` logic should prevent us from requesting the same block twice?

src/vsr/grid_repair_queue.zig
```diff
@@ -0,0 +1,489 @@
...
+        table: *RepairTable,
+        address: u64,
+        checksum: u128,
+    ) void {
+        assert(!queue.canceling);
+        assert(queue.faulty_tables.count < queue.options.tables_max);
+        assert(!queue.request_table_queued(address, checksum));
+
+        var tables = queue.faulty_tables.peek();
+        while (tables) |queue_table| : (tables = queue_table.next) assert(queue_table != table);
```
matklad: Hm, this `while` just inlines the assert above?

src/vsr/grid_repair_queue.zig
```diff
@@ -0,0 +1,489 @@
...
+
+        table.* = .{
+            .callback = callback,
+            .index_address = address,
+            .index_checksum = checksum,
+        };
+        queue.faulty_tables.push(table);
+
+        if (queue.faulty_blocks.getPtr(address)) |fault| {
```
matklad: Can get_or_put here?

src/vsr/grid_repair_queue.zig
```diff
@@ -0,0 +1,489 @@
...
+                .checksum = checksum,
+                .progress = .{ .table_index = .{ .table = table } },
+            });
+        }
+    }
+
+    pub fn repair_ready(queue: *const GridRepairQueue, address: u64, checksum: u128) bool {
+        const fault_index = queue.faulty_blocks.getIndex(address) orelse return false;
+        const fault = &queue.faulty_blocks.entries.items(.value)[fault_index];
+        return fault.checksum == checksum and fault.state == .waiting;
```
matklad: `.waiting` and  `ready` read like antonyms to me.

src/vsr/grid_repair_queue.zig
```diff
@@ -0,0 +1,489 @@
...
+        }
+
+        if (switch (fault.progress) {
+            .block => null,
+            .table_index => |progress| progress.table,
+            .table_content => |progress| progress.table,
+        }) |table| {
+            assert(table.table_blocks_total != null); // We already received the index block.
+            assert(table.table_blocks_written < table.table_blocks_total.?);
+            assert(table.content_blocks_received.count() <= table.table_blocks_total.? - 1);
```
matklad: If this could be done without duplciating this bit, assert that `content_blocks_received.isSet(fault.progerss.index)` on the `table_content` branch, and that it is zero on `table_index` branhc.

src/vsr/grid_repair_queue.zig
```diff
@@ -0,0 +1,489 @@
...
+        const index_block_header = schema.header_from_block(index_block_data);
+        assert(index_block_header.op == table.index_address);
+        assert(index_block_header.checksum == table.index_checksum);
+        assert(schema.BlockType.from(index_block_header.operation) == .index);
+
+        const content_blocks_total = index_schema.content_blocks_used(index_block_data);
+        table.table_blocks_total = 1 + content_blocks_total;
+
+        var content_block_index: usize = 0;
+        while (content_block_index < content_blocks_total) : (content_block_index += 1) {
```
matklad: `for(0..content_blocks_total) |content_block_index|`

src/vsr/grid_repair_queue.zig
```diff
@@ -0,0 +1,489 @@
...
+        assert(queue.request_blocks_available() > 0);
+        assert(queue.faulty_tables.count <= queue.options.tables_max);
+
+        if (queue.faulty_blocks.getPtr(address)) |fault| {
+            assert(fault.checksum == checksum);
+            assert(fault.state != .aborting);
+            return;
+        }
+
+        queue.faulty_blocks.putAssumeCapacityNoClobber(address, .{
```
matklad: could `get_or_put` here to avoid two hash map lookups. I think we also have three call-sites where we add entries for faulty_blocks (request{_table,_table_content,_block}). Maybe it makes sense to extract a helper such that there's only a single place where we physically add block to the hash map, so that there's a single place where we check validity of all transitions (eg, if we concurrently request the block for different reasons).

src/vsr/grid.zig
```diff
@@ -611,29 +652,14 @@ pub fn GridType(comptime Storage: type) type {
...
-                    {
-                        grid.read_global_queue.remove(read);
-                        grid.read_block_resolve(read, .{ .valid = cache_block.* });
-                        break;
-                    }
-                } else {
-                    unreachable;
-                }
-            }
+            if (completed_write.repair) grid.repair_queue.repair_complete(cache_block.*);
```
matklad: Maybe move this to after 

```zig
completed_write.callback(completed_write);
```

?

`repair_complete` internally also invokes user-supplied callback, and I think we want to sequence that after all other effects of a write are carried out. 

I don't understand specifics here yet, but it also seems suspicious that we now essentially attach two different callbacks to the write...

src/vsr/grid_repair_queue.zig
matklad: (allocate_block is also unused)

sentientwaffle: > Am I correct that, depending on timings, either commit or repair might proceed?

Yes. `write_iops_max` determines how many physical disk writes can execute concurrently. If we start more `grid.write_block()`s than that (due to repairs or commits or some mix of the two) then they will queue up in `Grid`'s write queue (which is a FIFO linked list). 

`grid_repair_writes_max` (the correct one, not the unused one!) acts as an upper bound -- at most how much of the IOPS are willing to spend on repair? The higher we set it, the more we favor repair over commit. (Or vice-versa.)

Right now `constants.zig` sets

```zig
/// Immediately after state sync we want access to all of the grid's write bandwidth to rapidly sync
/// table blocks.
pub const grid_repair_writes_max = grid_iops_write_max;
```

but I could imagine us tuning them separately in the future.

src/vsr/grid_repair_queue.zig
sentientwaffle: Adding:

```
//! - GridRepairQueue.request_table() is called immediately after superblock sync.
//! - GridRepairQueue.request_block() is called by the grid when non-repair reads encounter corrupt
//!   blocks.
```

> or, asynchronously, by `request_table`

`request_table` doesn't use `request_block`, it just queues the blocks manually.

matklad:
Big picture question: do we have the logic for

> check if we have the block on disk already, but, if we don't, request it from peers

?

It seems like we go straight to the network, which feels a bit odd --- after a sync, I would still expect us to have most of the blocks on disk already.

sentientwaffle:
> It seems like we go straight to the network, which feels a bit odd --- after a sync, I would still expect us to have most of the blocks on disk already.

We go straight to the network, but keep in mind we are specifically only repairing tables (and all associated blocks) that were created during our `sync_op_{min,max}` range. (Likewise for client replies.)

```zig
            while (self.sync_tables.?.next(&self.state_machine.forest)) |table_info| {
                if (table_info.snapshot_min >= snapshot_from_commit(sync_op_min) and
                    table_info.snapshot_min <= snapshot_from_commit(sync_op_max))
                {
                ...
```

So we usually do _not_ already have the blocks we are syncing.
The exceptions (i.e. where we _do_ already have the blocks) are:
- Some overlap necessary to properly sync when recovering from a divergent checkpoint. (See the "TODO" below).
- We did some sync work, then crashed, then recovered, and are now repeating that sync work.

```zig
                // We must re-sync the tables created between `vsr_state.commit_min_canonical` and
                // our current (pre-sync) checkpoint, since the current checkpoint might be
                // divergent, in which case the tables we created during compaction are invalid.
                // TODO: When we know our checkpoint is canonical (via commit/ping), we don't need
                // to resync it.
                const sync_min_new = if (vsr.Checkpoint.trigger_for_checkpoint(
                    self.superblock.working.vsr_state.commit_min_canonical,
                )) |trigger|
                    // +1 because `vsr_state.commit_min_canonical` is definitely canonical, and
                    // the range is inclusive.
                    trigger + 1
                else
                    0;
```

src/vsr/grid_repair_queue.zig
sentientwaffle: It might be empty if all of the requests were taking by the `read_global_queue`:

```zig
        pub fn block_requests(grid: *Grid, requests: []vsr.BlockRequest) usize {
            const request_faults_count = @min(grid.read_global_queue.count, requests.len);
            const request_repairs_count =
                grid.repair_queue.block_requests(requests[request_faults_count..]);
```

src/vsr/grid_repair_queue.zig
sentientwaffle: I reworked the loop into

```zig
        const requests_count = @min(requests.len, faults_total);
        for (requests[0..requests_count], 0..) |*request, fault_offset| {
```

instead.

sentientwaffle: Wait nvm, that doesn't work (since some faulty blocks are skipped over b/c they are being written).

matklad:
> we are specifically only repairing tables (and all associated blocks) that were created during our sync_op_{min,max} range.

Aha, I was missing that! Nifty, that means all the logic for avoiding redundant work is handled by the iterator. I guess we could also use the iterator to get back the benefits of “index block implies all data blocks” invariant? That is, we can make the iterator state durable, so that, after restart, we can continue from a specific position.

sentientwaffle:
> Aha, I was missing that! Nifty, that means all the logic for avoiding redundant work is handled by the iterator. I guess we could also use the iterator to get back the benefits of “index block implies all data blocks” invariant? That is, we can make the iterator state durable, so that, after restart, we can continue from a specific position.

Yeah, there are a few different ways we could go in that direction. Another is to make multiple passes with the iterator, incrementally walking (and updating in the superblock) `sync_op_min` forward. e.g. if we have `vsr_state.sync_op_min=100` and `vsr_state.sync_op_max=900`, then we could sync tables with ops 100-500 first, then set `vsr_state.sync_op_min=500`, then  sync 500-900, etc.

src/vsr/grid_repair_queue.zig
sentientwaffle: Actually I'm going to keep the `while` loop -- we need `fault_offset` after the loop finishes to update the `queue.faulty_blocks_repair_index` cursor, and I want that update outside of the loop since it mutates the queue.

src/vsr/grid_repair_queue.zig
sentientwaffle: My motivation here was that fairness doesn't matter too much -- we care about sending out requests for lots of different blocks, but if there are some misses (or repeats) then it isn't a big deal. In the future, the EWAH-bitset-of-table-content-blocks will be more fair though.

src/vsr/grid_repair_queue.zig
sentientwaffle: :+1: When we encounter an `aborting` block, we can assert that `assert(queue.checkpointing.? > 0)`.

src/vsr/grid_repair_queue.zig
sentientwaffle: :+1: I like "enqueue" -- even though the grid repair queue isn't physically a queue, my (simplified) mental model of it is a queue.

src/vsr/grid_repair_queue.zig
sentientwaffle: Good catch; that's a bug! I'm going to add both
```
    enqueued_single_blocks: usize = 0,
    enqueued_table_blocks: usize = 0,
```

-- we only need one, but having both means they can be checked against each other:

```
        assert(queue.faulty_blocks.count() ==
            queue.enqueued_single_blocks + queue.enqueued_table_blocks);
```

src/vsr/grid_repair_queue.zig
sentientwaffle: Not quite:

- `queue.request_table_queued()` checks whether we have already queued that particular `address` and `checksum`.
- The `while` loop checks that we don't accidentally (illegally) reuse the `table: *RepairTable` context object. (E.g. if, somehow, the `Replica` released the table from its `IOPS` prematurely and tried to reuse it.)

src/vsr/grid_repair_queue.zig
matklad: The only problem with enqueue is that it makes me think silly thoughts https://xkcd.com/853/ :D 

But maybe that’s an advantage!

src/vsr/grid_repair_queue.zig
sentientwaffle: Not necessarily, e.g.

1. Queue block X
2. Block X arrives and the grid fulfills its read, but we don't write it because all of the repair-writes are busy.
3. Another read fails, queue block X again.

src/vsr/grid_repair_queue.zig
sentientwaffle: What about `repair_waiting`? repair_in_progress / _scheduled / enqueued (as you suggest elsewhere) sound like they would return true when the block has arrived but is being written.

src/vsr/grid_repair_queue.zig
sentientwaffle: Hmm, I am also changing `content_block_index` from a `usize` to a `u32` -- is there any way to avoid a bunch of `@intCast`s? Even when the top of the range `content_blocks_total` is a `u32`, `content_block_index` is a `usize`.

src/vsr/grid_repair_queue.zig
matklad: Ah, then we can assert this in `Grid.block_requests`! I am actually unsure here --- on the one hand, empty slice is a natural neutral element here, on the other hand, we should never issue "empty" network messages or "empty" disk writes, and deep asserts seem useful.

src/vsr/grid_repair_queue.zig
matklad: Yeah, that's the simplest&best

src/vsr/grid_repair_queue.zig
matklad: TIL, that's unfortunate, indeed it seems like there's no way to loop over `u32` directly. 

I'd maybe still write something like

```zig
for (0..content_bllocks_total) |content_block_index_usize| {
    const content_block_index: u32 = @intCast(content_block_index_usize);
}
```

That's the same amount of lines as the while, but it makes it clear that there aren't any funny mutations of the index inside the loop. This is most valuable in that it makes counting `while` loops signal "hey, I am doing something extra funky with the index here".

matklad: Asked the hive mind just in case, but I think @intCast vs while is basically it:

https://ziggit.dev/t/idiomatic-counting-loop-over-u32/1731

src/vsr/grid_repair_queue.zig
```diff
@@ -0,0 +1,548 @@
+//! Track corrupt/missing grid blocks.
+//!
+//! - The GridRepairQueue is LSM-aware: it can repair entire tables.
+//! - The GridRepairQueue is shared by all Trees.
+//! - The GridRepairQueue is "coherent" – that is, all of the blocks in the queue belong in the
+//!   replica's current checkpoint:
```
matklad: I wonder if we should add `checkpoint_id: u128` field here then? I think it could make the control flow around cancelation & checkpointing more obvious. 

I think we can even use that for extra asserts? 

We could have an API like

```zig
fn assert_coheret(checkpoint_id: u128, block_addr: u64, block_checksum: u128)
```

which would be a no-op in production, but, for a test storage, it'll check that (across cluster) everything is coheret. I guess that this'll mean that we have to parametrize `GridRepairQueue` over storage, making it generic, which would be unfortunate. OTOH, I feel that asserting that we don't mixup different checkpoints is pretty valuable, both because such a mixup feels like a plausbile bug with bad consequences, and also because assertions in place would make the control flow clearer.

src/vsr/grid_repair_queue.zig
```diff
@@ -0,0 +1,548 @@
...
+    /// - For every block address in faulty_blocks, ¬free_set.is_free(address).
+    faulty_blocks: FaultyBlocks,
+    /// Index within `faulty_blocks`, used to cycle through block-repair requests.
+    ///
+    /// Invariants:
+    /// - faulty_blocks.count() > 0 implies faulty_blocks_repair_index < faulty_blocks.count()
+    /// - faulty_blocks.count() = 0 implies faulty_blocks_repair_index = faulty_blocks.count()
+    faulty_blocks_repair_index: usize = 0,
+
+    /// Invarants:
```
matklad: Typo, `Invariants`

src/vsr/grid_repair_queue.zig
```diff
@@ -0,0 +1,548 @@
...
+    pub fn next_batch_of_block_requests(
+        queue: *GridRepairQueue,
+        requests: []vsr.BlockRequest,
+    ) usize {
+        assert(!queue.canceling);
+        maybe(requests.len == 0);
+
+        const faults_total = queue.faulty_blocks.count();
+        if (faults_total == 0) return 0;
+        assert(faults_total > queue.faulty_blocks_repair_index);
```
matklad: Super trivial, but let's maybe flip it around?


```zig
assert(queue.faulty_blocks_repair_index < faults_total);
```

`index < len` is how this invariant spelled most of the time (eg, in C-style for loop), comparison in the opposite direction feels odd.

At one point, I actually used to completely ban `>` and `>=` operators, to canonicalize the direction of all comparisons in a way which is consistent with the number line (this is especially nice for `lo < x && x < hi` range checks).

I now think that that goes too far, and do use `>` occasionally, but I still find canonicalizing to `<` useful, especially for index-len checks. For example, I prefer

```
if (index < len) {
} else {
}
```

and even

```
if (!(index < len)) {
}
```

to

```
if (index >= len)
```

I find it useful when invariant is stated positively and negation is pushed to the top-level.

src/vsr/grid_repair_queue.zig
sentientwaffle: :thinking: I think this is covered when grid writes complete: `assert(!grid.superblock.free_set.is_free(completed_write.address));`.

(But I'm also not really sure how this assertion you're suggesting would work, so possibly I'm missing something. Maybe it is something that could go in the `StorageChecker` (or a similar checker) instead?)

matklad: What I am thinking here is that we use two "coordinates" to refer to blocks: address & checksum. But actually, there are three: address, checksum, and checkpoint_id.

We have an invariant that at most one checkpoint is active at any given point in time, but, as far as I can tell, this invariant is implicit. That is, when someone calls

```
pub fn enqueue_block(queue: *GridRepairQueue, address: u64, checksum: u128) void {
```

they call it at a particular `checkpoint_id`, and that `checkpoint_id` must match the current `checkpoint_id` of the `GridRepairQueue`. 

Similarly, calling `checkpoint_commence` implicitly advances this current `checkpoint_id`. 

The problem I see here is that this is very implicit --- we don't _actually_ pass the new `checkpoint_id`. If someone calls `enqueue_block` with arguments from a wrong checkpoint, this will probably trip some assert somewhere eventually, but that would be a generic "state of the grid is bad" assert we get on a subsequent read, not "you are trying to write data for checkpoint X, but the current checkpoint is Y" on the immediate problematic write. 

We do have `is_free` checks which I think solve part of the problem here. If I understand this correctly, a block at particular address evolves as `free -> checksumA -> free -> checksumB`, it can't evolve as `free -> checksumA -> -> checksumB`. So, it's impossible to confuse two neighbouring checkpoints, because they pass through a "free" state. But I think this ABA problem exists across several checkpoints, where a block has different contents at checkpoints `X` and `X+2`, but it is not free in either (but free in `X+1`). 

So, my overall thought here is to add

```zig
fn assert_coheret(checkpoint_id: u128, block_addr: u64, block_checksum: u128)
```

to record the fact that `block_addr` at `checkpoint_id` has contents `block_checksum`, and call that function whenever convenient, on write and read paths. Implementation wise, I think this assert could maintain a global backing hashmap keyed by `(addr, checkpoint)` tuple, which would `get_or_put` checksum and verify that it's never overwritten with conflicting values. 

---


To maybe arrive at the same point from a different angle, imagine writing an Ada-Spark style proof here that the code works correctly. I think that, to reason about correctness, we would actually have to both prove and also rely on the fact that all the faulty blocks in repair queue are coherent (belong to the same checkpoint). And to write such a proof, we'll probably need to introduce a "shadow state" --- a logical variable that tracks the current checkpoint, which is updated in the proof, but doesn't actually exist in code

And my thinking is that maybe we should actually go and materialize this shadow variable into an actual variable which tracks the current checkpoint from repair queue point of view?  Strictly speaking, this is redundant, but I think it might create space for more thorough assertions.

src/vsr/grid_repair_queue.zig
```diff
@@ -0,0 +1,548 @@
...
+
+            switch (fault_data[fault_index].state) {
+                .waiting => {
+                    requests[requests_count] = .{
+                        .block_address = fault_addresses[fault_index],
+                        .block_checksum = fault_data[fault_index].checksum,
+                    };
+                    requests_count += 1;
+
+                    if (requests_count == requests.len) break;
```
matklad: I think this `break` is fully redundant with `requests_count < requests.len` while condition, so we  could remove either the break, or the extra condition in while (I'd prune the while)

src/vsr/grid_repair_queue.zig
```diff
@@ -0,0 +1,548 @@
...
+        }
+
+        queue.faulty_blocks.swapRemoveAt(fault_index);
+
+        if (queue.faulty_blocks_repair_index == queue.faulty_blocks.count()) {
+            queue.faulty_blocks_repair_index = 0;
+        } else {
+            if (queue.faulty_blocks_repair_index > fault_index) {
+                queue.faulty_blocks_repair_index -= 1;
+            }
```
matklad: Why do we need this `-= 1`? We are swap-removing, so we don't need to shift the index outside of `== .count()` case handledd above?

I think, as written, this will cause us to send the same block in two consequitive requests?

src/vsr/grid_repair_queue.zig
```diff
@@ -0,0 +1,548 @@
...
+        assert(queue.checkpointing == null);
+
+        queue.faulty_blocks.clearRetainingCapacity();
+        queue.faulty_blocks_repair_index = 0;
+        queue.enqueued_single_blocks = 0;
+        queue.enqueued_table_blocks = 0;
+
+        queue.canceling = true;
+        while (queue.faulty_tables.pop()) |table| {
+            (table.callback)(table, .canceled);
```
matklad: Hm, invoking callbacks here looks suspicious. In `Replica.sync_request_table_callback`, it's not clear why we won't enter

```
            if (self.sync_tables) |_| {
                self.sync_request_tables();
            }
```

and then re-enter GridRepairQueue. Let's explicitly match over `RepairTableResult` in `sync_request_table_callback` and assert that we don't do anything funny in the canceled case.

src/vsr/grid_repair_queue.zig
```diff
@@ -0,0 +1,548 @@
...
+        }
+        queue.canceling = false;
```
matklad: I am wondering if it makes sense to write this as 

```
queue.* = .{
}
```

just to double-check that we didn't forget to reset any fields?

src/vsr/grid_repair_queue.zig
```diff
@@ -0,0 +1,548 @@
...
+                }
+            }
+        }
+
+        var tables: FIFO(RepairTable) = .{ .name = queue.faulty_tables.name };
+        while (queue.faulty_tables.pop()) |table| {
+            assert(!free_set.is_free(table.index_address));
+
+            if (free_set.is_released(table.index_address)) {
+                (table.callback)(table, .released);
```
matklad: Hm, this callback also feels complicated. Am i correct that this callback could cause the replica to call `enqueued_table`, so `enqueued_table` might be running while we are in hte middle of `checkpoint_commence`? Is there anyway  we could simplify that? Such re-entrancy I think goes against

> Use **only very simple, explicit control flow** for clarity.

:thinking: 

What if we remove table-synced callback, and instead ask the replica to explicitly poll for synced tables?

matklad: Yeah, I guess we could poll in `grid_repair_block_callback` ?

src/vsr/grid_repair_queue.zig
sentientwaffle: Ah, you're right, this is unnecessary -- I wasn't considering the swap-remove!

src/vsr/grid_repair_queue.zig
sentientwaffle: ~I think on cancellation we _will_ enter `sync_request_tables()`, and that is good -- we are queuing tables that belong to the checkpoint (since they are in our manifest levels), so they are safe to repair concurrent to the checkpoint.~

sentientwaffle: Oh wait nevermind, I was thinking of the checkpoint case, didn't realize this was cancellation!

src/vsr/grid_repair_queue.zig
sentientwaffle: :thinking: So then `Replica` would not own the `RepairTable`s, `GridRepairQueue` would own them, and `Replica` would poll for their availability?

Fwiw the analogue right now is quite similar to e.g. `read_block`, where it is safe for the caller to call `read_block` within the callback of another `read_block`, reusing the same `*Read`.

matklad: No, what I had in mind is the same ownership structure as today. Rather:

* we remove callback from repair table
* GridRepaiRQueue maintains a FIFO of repaired tables
* when the table is repaired, we push it onto this ready queue, instead of invoking calllback
* We add API to drain the ready tables


So the replica doesn’t get notified when a table is repaired. Rather, it reaps ready table at its own leisure.

sentientwaffle: Ah, that makes sense

matklad: Don’t react to stimuli basically — poll instead of callback. The question is, can the replica know when to poll?

sentientwaffle: We could put it in the repair timeout.

matklad: That, and I think callback from on_block would do the trick.

matklad: >Fwiw the analogue right now is quite similar to e.g. read_block, where it is safe for the caller to call read_block within the callback of another read_block, reusing the same *Read.

I would worry about this as well. What makes me uneasy is that this is re-entrancy -- an object's method is called, while another method is executed, and this happens indirectly via callback. In theory, this is bad because it might cause unanticipated violation of the invariants: inside the first method, invariants might be _temporarily_ suspended, but we might expect the invariants to hold when we enter the nested-through-callback method call. In practice, I noticed that two-thirds of unsound `unsafe` blocks in Rust look like this:

```
fn allegedly_safe_abstraction(callback: fn) {
  unsafe {
   ...
   callback()
   ...
  }
}
```

where UB happens if the (safe) user-supplied callback manages to call `allegedly_safe_abstraction` re-entrantly.

src/vsr/grid.zig
```diff
@@ -491,66 +495,102 @@ pub fn GridType(comptime Storage: type) type {
...
+                {
+                    grid.read_global_queue.remove(read);
+                    grid.read_block_resolve(read, .{ .valid = block });
+                    return true;
+                }
+            }
+            return false;
+        }
+
+        pub fn repair_block_ready(grid: *Grid, address: u64, checksum: u128) bool {
```
matklad: s/ready/waiting/ ?

src/vsr/replica.zig
```diff
@@ -2304,86 +2324,77 @@ pub fn ReplicaType(
...
-                .write = undefined,
-            };
 
-            const block = &self.grid_write_blocks[self.grid_writes.index(write)];
-            stdx.copy_disjoint(.inexact, u8, block.*, message.buffer[0..message.header.size]);
+                    stdx.copy_disjoint(
+                        .inexact,
+                        u8,
+                        self.grid_repair_write_blocks[write_index],
+                        message.buffer[0..message.header.size],
```
matklad, line 2365: Assert that the rest of the buffer is zeroed.

sentientwaffle: I assume you mean the block buffer, not the message buffer? I don't remember any guarantees about the message buffer beyond `header.size`.

src/lsm/forest_table_iterator.zig
```diff
@@ -0,0 +1,179 @@
...
+
+        trees: TreeTableIterators = default: {
+            var iterators: TreeTableIterators = undefined;
+            for (std.meta.fields(TreeTableIterators)) |field| @field(iterators, field.name) = .{};
+            break :default iterators;
+        },
+
+        pub fn next(iterator: *ForestTableIterator, forest: *Forest) ?TableInfo {
+            while (iterator.level < constants.lsm_levels) : (iterator.level += 1) {
+                inline for (Forest.tree_infos) |tree_info| {
```
matklad: This `inline for` feels problematic --- iterating the last tree requires skipping over all other trees in O(number of trees) and, what's more, due to advancing generations, this might actually need some non-trivial work. Let's store `tree_id` as a field, and directly inline switch to the correct tree? The same pattern we use for `manifest_log_open_event`.

src/lsm/forest_table_iterator.zig
```diff
@@ -0,0 +1,179 @@
...
+                    }
+                } else {
+                    break :tables manifest_level.tables.iterator_from_cursor(
+                        manifest_level.tables.first(),
+                        .ascending,
+                    );
+                }
+            };
+
+            const table = table_iterator.next() orelse return null;
```
matklad, line 177: Took me this morning to understand that `+ 1` in `snapshot_min + 1` above :-) But now I know about snapshots! I think to make this easier to grok, we can assert after this line that `table` is correctly ordered after `position.previous` (that is, that it has ge key-max, and, in case of a tie, a larger snapshot).

Or maybe even, to make this more robust --- use `tables.search` to re-find the current table, and then call `.next`. That way, the code works even if we change the order of tables in the manifest.tables. :thinking: ah, I think this doesn't actually work --- we might not necessary be able to find the previous table, as it might have been deleted during the checkpoint.

src/vsr/replica.zig
```diff
@@ -7831,9 +7876,90 @@ pub fn ReplicaType(
...
                     }
                 }
+                return self.sync_tables == null and self.grid_repair_tables.executing() == 0;
+            }
+        }
 
-                // TODO Return whether all tables are synced.
-                return true;
+        /// State sync finished, and we must repair all of the tables we missed.
+        fn sync_request_tables(self: *Self) void {
```
matklad: s/request/enqueueu

src/vsr/replica.zig
```diff
@@ -7831,9 +7876,90 @@ pub fn ReplicaType(
...
+            assert(self.syncing == .idle);
+            assert(self.sync_tables != null);
+            assert(self.state_machine_opened);
+            assert(self.superblock.working.vsr_state.sync_op_max > 0);
+            assert(self.grid_repair_tables.available() > 0);
+
+            const snapshot_from_commit = vsr.Snapshot.readable_at_commit;
+            const sync_op_min = self.superblock.working.vsr_state.sync_op_min;
+            const sync_op_max = self.superblock.working.vsr_state.sync_op_max;
+            while (self.sync_tables.?.next(&self.state_machine.forest)) |table_info| {
```
matklad, line 7901: Would be great to assert that we skip over all tables with generation larger than then initial one. Though I don't think we have the data about which generation the table was added to the manifest.

src/vsr/replica.zig
```diff
@@ -7831,9 +7876,90 @@ pub fn ReplicaType(
...
+                if (table_info.snapshot_min >= snapshot_from_commit(sync_op_min) and
+                    table_info.snapshot_min <= snapshot_from_commit(sync_op_max))
+                {
+                    if (self.grid.repair_queue.enqueued_table(
+                        table_info.address,
+                        table_info.checksum,
+                    )) {
+                        assert(self.grid_repair_tables.executing() > 0);
+                        continue;
+                    }
```
matklad: Hm, why could we hit this case? The iterator can't return the same table twice I think? And when we reset the iterator in `sync_content`, I think there should be no pending repairing tables?

src/vsr/replica.zig
```diff
@@ -7801,6 +7839,13 @@ pub fn ReplicaType(
             assert(self.syncing == .idle);
             assert(self.state_machine_opened);
             assert(self.superblock.working.vsr_state.sync_op_max > 0);
+            assert(self.sync_tables == null);
+            assert(self.grid_repair_tables.available() == constants.grid_repair_tables_max);
+
```
matklad, line 7851: Can we assert the state of `self.grid.repair_queue` here? I _think_ it should be idle after cancellation.

src/constants.zig
```diff
@@ -277,19 +277,41 @@ pub const grid_repair_request_max = config.process.grid_repair_request_max;
...
-/// The number of grid writes allocated to handle incoming command=block messages.
-pub const grid_repair_writes_max = config.process.grid_repair_writes_max;
+/// Immediately after state sync we want access to all of the grid's write bandwidth to rapidly sync
+/// table blocks.
+pub const grid_repair_writes_max = grid_iops_write_max;
 
 /// The default sizing of the grid cache. It's expected for operators to override this on the CLI.
 pub const grid_cache_size_default = config.process.grid_cache_size_default;
 
+/// The maximum capacity (in blocks) of the GridRepairQueue.
```
matklad: I think this is imprecise, as the queue covers both on-demand lazy repairs (this constant) and eager syncs (contorlled by repair_tables_max). 

Maybe that's a sign that `GridRepairQueue` is not the best name, as it confuses repair with sync? As a bikeshed, maybe just `MissingBlocks`? It also tracks missing tables, but only as bundles of missing blocks.

src/constants.zig
```diff
@@ -277,19 +277,41 @@ pub const grid_repair_request_max = config.process.grid_repair_request_max;
...
+
 comptime {
     assert(grid_repair_request_max > 0);
     assert(grid_repair_request_max <= @divFloor(message_body_size_max, @sizeOf(vsr.BlockRequest)));
     assert(grid_repair_request_max <= grid_repair_reads_max);
 
     assert(grid_repair_reads_max > 0);
     assert(grid_repair_writes_max > 0);
+
+    assert(grid_repair_blocks_max > 0);
```
matklad: Should we sanity check that `grid_repair_writes_max <= grid_repair_blocks_max`? Basically, that we don't over-allocate writes. Though, to do this precisely we need to redo the capacity computation repair queue is doing.

src/vsr/grid.zig
```diff
@@ -491,66 +495,102 @@ pub fn GridType(comptime Storage: type) type {
...
-                    assert(cache_block != block.*);
-                }
-            }
+            grid.repair_queue.repair_commence(block_header.op, block_header.checksum);
+            grid.write_block(callback, write, block, .repair);
+        }
 
+        /// Write a block for the first time.
+        /// NOTE: This will consume `block` and replace it with a fresh block.
+        pub fn create_block(
```
matklad, line 544: An absolutely brilliant addition! There are only too call-sites in our codebase, one in compaction, and one in manifest, and those two sites tell _the whole_ story about how data is being created!

src/lsm/schema.zig
matklad: Yeah, would be good! I am not a fan of the current impl, as its mechanically apathetic --- we take a nicely-arranged non-polymorphic batch, and turn into into a bunch of singular blocks with an `if` guarding each block.

src/tigerbeetle/main.zig
```diff
@@ -185,6 +185,7 @@ const Command = struct {
                 .io = &command.io,
             },
             .grid_cache_blocks_count = args.cache_grid_blocks,
+            .grid_checker = null,
```
matklad: I guess we could make `GridChecker` type a member of `Storage`, so that in prod we just use `void` here? If that's not too much trouble, I'd do that --- I don't feel good about this "runtime" `if (grid_checker == null)` branches in prod. There's no _specific_ problems there, but just feels odd.

src/vsr/grid.zig
```diff
@@ -987,33 +1033,76 @@ pub fn GridType(comptime Storage: type) type {
...
+                    grid.repair_queue.enqueue_block(
+                        read_remote_head.address,
+                        read_remote_head.checksum,
+                    );
                 }
             }
         }
 
+        pub fn next_batch_of_block_requests(grid: *Grid, requests: []vsr.BlockRequest) usize {
+            assert(requests.len > 0);
```
matklad, line 1052: can we assert that we are not canceling here?

src/vsr/replica.zig
```diff
@@ -2304,86 +2324,75 @@ pub fn ReplicaType(
...
 
-            if (self.grid.canceling != null) {
-                log.debug("{}: on_block: ignoring; grid is canceling (address={} checksum={}", .{
+            const block = message.buffer[0..constants.block_size];
+
+            const grid_fulfill = self.grid.fulfill_block(block);
+            if (grid_fulfill) {
+                assert(!self.superblock.free_set.is_free(message.header.op));
+
+                log.debug("{}: on_block: fulfilling address={} checksum={}", .{
```
matklad: `fulfilled`, as this happens after we actually invoked the callbacks.

src/vsr/replica.zig
```diff
@@ -2304,86 +2324,75 @@ pub fn ReplicaType(
...
-            self.grid.write_block_repair(
-                on_block_write_repair,
-                &write.write,
-                block,
-                message.header.op,
-            );
+                    write.* = .{ .replica = self };
+                    self.grid.repair_block(grid_repair_block_callback, &write.write, write_block);
+                } else {
+                    log.debug("{}: on_block: ignoring; no write available " ++
```
matklad, line 2372: Let's add a log when the block is used neither to fulfil an outstanding read, nor for reapir, so that we have some traces in the logs in case we just send wrong blocks.

src/vsr/replica.zig
matklad: Let's also `assert(self.grid_repair_tables.available() > 0);` at the start of the loop (we do it on function entry, but we use it on every iter)

src/vsr.zig
```diff
@@ -813,6 +813,7 @@ pub const Header = extern struct {
 
     fn invalid_block(self: *const Header) ?[]const u8 {
         assert(self.command == .block);
+        if (self.size > constants.block_size) return "size > block_size";
```
matklad, line 817: Could we also check `if (self.size == @sizeOf(vsr.Header))` ?

src/testing/cluster/storage_checker.zig
```diff
@@ -171,5 +178,42 @@ pub fn StorageCheckerType(comptime Storage: type) type {
...
+                    });
+
+                    blocks_missing += 1;
+                    continue;
+                };
+
+                const block_header = schema.header_from_block(block);
+                assert(block_header.op == block_address);
+
+                checksum ^= vsr.checksum(block[0..block_header.size]);
```
matklad: I may be paranoid, but let's add some kind of a counter or nonce here, so that identical blocks don't cancel each other out.

src/testing/cluster/storage_checker.zig
matklad: :tada:

src/simulator.zig
```diff
@@ -331,6 +331,7 @@ pub const Simulator = struct {
         request_idle_off_probability: u8, // percent
     };
 
+    allocator: std.mem.Allocator,
```
matklad: Let's either pass this to `core_missing_blocks` or remove `alocator` argument from deinit.

matklad (commented):
Finallly got to the last page!

src/vsr/replica.zig
sentientwaffle: :thinking: Agreed, this isn't necessary. 

I don't remember what my reasoning for adding it was. :shrug:

src/vsr/replica.zig
sentientwaffle: The grid repair queue isn't necessarily idle here: At this point, we just finished opening the state machine, but there might be writes to repair the manifest log blocks still in progress.

I can add an assert before we call `StateMachine.open()`, though.

src/constants.zig
sentientwaffle: :thinking: Hmm, good point, that is confusing. I think I'll go with `GridMissingBlocks` though -- I want "Grid" in there to make the relationship with the `Grid` obvious.

(I was considering `GridRequestQueue` as an alternative, but we have other "request queues" elsewhere, and the thing being queued is missing blocks, not requests, so that more makes sense.)

sentientwaffle:
https://github.com/tigerbeetle/tigerbeetle/pull/1179/commits/4522994463f27bd7b645f9cd2cd3fa490acc0f4f is a bug the VOPR found. Not a great solution, but I'm not sure what the "proper" solution is. Maybe use the grid scrubber to check blocks in our repair queue periodically?

src/constants.zig
jorangreef: If I understand correctly, this is used for repairs (not missing, but corrupt) or for gaps (missing)? If so, perhaps we can still improve on `Missing`?

Per TigerStyle, the qualifier should go last, e.g. `GridBlocks[Missing|Outstanding|Faulty]`

jorangreef: Another idea: `GridGaps`

matklad:
>Maybe use the grid scrubber to check blocks in our repair queue periodically?

:thinking: heh, that's indeed an interesting one. Formally, I think we are fine pretending that these blocks don't exist, as we commit to guaranteeing existence of a block only when we checkpoint. This is similar to how we sometimes truncate some prepares from WAL on restart for simplicity, in that we forget about some data we stored, but we rely on cluster guarantees to ensure this is safe. 

It is a bit more dicey in this case (as grid blocks are definitely committed data), but I think should also be fine --- the replica would completely lose the knowledge about this blocks if it restarts. Grid writes "don't count" until they are reachable from the superblock. 

I think I convinced myself that checking for this in the VOPR is the proper solution here. I guess ideally we'd check that the block in question is indeed from a future checkpoint, but I don't think we can actually do that (or maybe we can? the GridChecker remebres which blocks are associated with which checkponts)

src/lsm/forest_table_iterator.zig
```diff
@@ -64,18 +64,43 @@ pub fn ForestTableIteratorType(comptime Forest: type) type {
             break :default iterators;
         },
 
+        tree_id: ?u16 = null,
```
matklad: Do we need `null` here? can't we say `tree_id: u16 = Forest.tree_infos[0].tree_id` and then write numeric loop from `id_range.min` to `id_range.max`, the way we do for levels .

src/testing/cluster/storage_checker.zig
```diff
@@ -210,6 +210,8 @@ pub fn StorageCheckerType(comptime Storage: type) type {
                 assert(block_header.op == block_address);
 
                 checksum ^= vsr.checksum(block[0..block_header.size]);
+                // Extra guard against identical blocks:
+                checksum ^= vsr.checksum(std.mem.asBytes(&block_address));
```
matklad: I think it's not quite it, this will give identical checksums:

R0: addr=1, content=X; addr=2, content=X
R1: addr=1, content=Y; addr=2, content=Y

basically, you are independency checksumming block contents and block addresses, but we want checksum them entangled, so that indentical contents doesn't cancel out. We can do it like this:

```zig
checksum ^= vsr.checksum(block[0..block_header.size] ++ std.mem.asBytes(&block_address))
```

but, looking at the bigger context, perhaps we need a streaming variant of our checksum?

```zig
pub fn checksum(source: []const u8) u128 {
    var stream = ChecksumStream.init();
    stream.add(source);
    stream.checksum();
}

pub const ChecksumStream = struct {
    state: Aegis128LMac_128,

    pub fn init() ChecksumStream {
        seed_init();
        return ChecksumStream{ .state = seed_state };
    }

    pub fn add(stream: *ChecksumStream, bytes: []const u8) void {
        stream.state.update(bytes);
    }

    pub fn checksum(stream: *ChecksumStream) u128 {
        var result: u128 = undefined;
        stream.state.final(mem.asBytes(&result));
        stream.* = undefined;
        return result;
    }
};
```

matklad (commented):
I love `GridGaps` suggestions from joran!

src/constants.zig
sentientwaffle: :thinking: "Grid gaps" makes me think of the unallocated parts of the grid, though.

src/vsr/checksum.zig
```diff
@@ -48,14 +48,37 @@ fn seed_init() void {
...
-    var state_copy = seed_state;
-    state_copy.update(source);
-    var result: u128 = undefined;
-    state_copy.final(mem.asBytes(&result));
-    return result;
+    //var state_copy = seed_state;
+    //state_copy.update(source);
+    //var result: u128 = undefined;
+    //state_copy.final(mem.asBytes(&result));
+    //return result;
```
matklad: Did you intend to leave these commented out?

sentientwaffle: Oops :facepalm:

sentientwaffle:
Rebased to resolve a weird type error. There were no conflicts though -- and only the last two commits have any changes:
- Changing how `tree_infos` are sorted at comptime. (This was where the type errors were).
- Removing the commented-out code in `checksum()`.

matklad (approved):

PR #1188, opened 2023-09-12
style: state invariants positively

matklad:


docs/TIGER_STYLE.md
```diff
@@ -156,6 +156,25 @@ Beyond these rules:
   branches and cases clear. Again, consider whether a single `if` does not also need a matching
   `else` branch, to ensure that the positive and negative spaces are handled or asserted.
 
+* Negations are not easy! In the source code, state invariants positively. When
```
matklad: > Negations are not easy!

Pun intended.

jorangreef:
```suggestion
* Negations are not easy! Therefore, state invariants positively. When
```

Since the context for most of these is source code.

docs/TIGER_STYLE.md
```diff
@@ -156,6 +156,25 @@ Beyond these rules:
...
+  working with lengths and indexes, this form is easy to get right:
```
jorangreef:
```suggestion
  working with lengths and indexes, this form is easy to get right (and understand):
```

docs/TIGER_STYLE.md
```diff
@@ -156,6 +156,25 @@ Beyond these rules:
...
+
+  ```zig
+  if (index < length) {
+    // The invariant holds.
+  } else {
+    // The invariant doesn't hold.
+  }
+  ```
+
+  This one is harder:
```
jorangreef:
```suggestion
  This form is harder, and also goes against the grain of how `index` would typically be compared to `length`, for example, in a loop condition:
```

docs/TIGER_STYLE.md
matklad: What I want to explain here is that, when `foo and bar` is an invariant, `if (!(foo and bar))` is clearer than `if (!foo or !bar)`. "In the source code" refers to the `foo and bar` substring literally present in the file. 

But it's not like the reader could understand what I mean here, unless the reader is me

matklad: I'll just more forcefully `. State invariants positively`, as a counter-weight and double-pun for `not easy`.

jorangreef: How would one triple-pun this?

matklad: Attempted Hemingway pun in the commit message: https://github.com/tigerbeetle/tigerbeetle/pull/1188/commits/4321b8ccf5821e0052ce7d69efed0fd769408494

jorangreef (approved):
Looks terse to me!

jorangreef:
Conditions are truly a moveable feast...

docs/TIGER_STYLE.md
jorangreef: I saw it and raised you a Heming'egg: https://github.com/tigerbeetle/tigerbeetle/pull/1188#issuecomment-1715636569

PR #1194, opened 2023-09-19
Update fractional-amounts.md

NeilW:
Update 64 to 128 bit in the documentation


batiati (approved):
Thank you for spotting that!

PR #1180, opened 2023-09-08
LSM: Unified {table_memory, prefetch + cache}, Scopes, Cache Map

cb22:
The primary purpose behind this PR is to optimize, and enable future optimizations for the CPU-bound parts of compaction, specifically including the ability to pace the previously one-off mutable table sort.

This PR unifies our `TableMutable` and `TableImmutable` backing implementations into a single `TableMemory`, hoists our Tree cache to rather be a Groove cache (keyed by ID instead of timestamp, and cutting out coherency complexity) as well as unifies the new Groove cache with the prefetch objects.

Additionally, there's a new data structure - `CacheMap` - a union between our `SetAssociativeCache` and a `HashMap`. There's also a new feature, for state machines, Scopes, enabled by the above changes that make it much simpler to write code that involves rollbacks.

# Show-me-the-numbers
|                                                      | main tps | this pr tps | main p100 | this pr p100 |
|------------------------------------------------------|----------|-------------|-----------|--------------|
| benchmark as is (10k acc), nvme                      | 121k     | 150k        | 5199ms    | 4286ms       |
| benchmark as is (10k acc), tmpfs                     | 230k     | 316k        | 1778ms    | 1616ms       |
| benchmark 10 accounts, nvme                          | 132k     | 163k        | 4932ms    | 4165ms       |
| benchmark 10 accounts, tmpfs                         | 266k     | 377k        | 1532ms    | 1740ms       |
| benchmark 100k acc, nvme                             | 117k     | 145k        | 5531ms    | 4516ms       |
| benchmark 100k acc, tmpfs                            | 215k     | 296k        | 1627ms    | 1590ms       |
| (bonus) benchmark 10 accounts, sequential ids, tmpfs | 335k     | 488k        | 1704ms    | 952ms        |

## Cache Map
The CacheMap is backed by two data structures: our `SetAssociativeCache` providing the Cache layer, and a `HashMap` providing the Stash layer. Functionally, it behaves as a cache where a value can be _guaranteed_ to be present for a certain period of time. 

This allows use cases beyond caching; namely where a value is required to be present if it has been inserted. A set associative cache, even if large enough, could suffer from collisions and not maintain this property. In this PR, it's the backing for the new combined Groove prefetch + cache. 

Additionally, it has scope support and a fuzzer.

## Unified, log based `TableMemory`
[...]

PR #1196, opened 2023-09-21
StateMachine: Prefetch from posted groove

sentientwaffle:
Regression via https://github.com/tigerbeetle/tigerbeetle/pull/1157


cb22 (approved):
Ha, did the VOPR catch this, or was it something you noticed?

Reason I ask is on #1180 I've been running it pretty extensively locally, and never bumped in to this, and I've got a feeling that the relatively large minimum object cache size (2048 vs 0/2048) is to blame.

sentientwaffle:
The VOPR found it -- I was testing a different branch with some manifest changes and assumed they were to blame at first :grin: 

> I've got a feeling that the relatively large minimum object cache size (2048 vs 0/2048) is to blame.

Yeah, that is definitely plausible.

PR #1198, opened 2023-09-21
Docs: Recommend ULID

sentientwaffle:


docs/FAQ.md
```diff
@@ -101,7 +101,7 @@ payment system.
 
 You might keep account metadata in your existing database (say,
```
cb22, line 102: The change is good from what it was, but why wouldn't we recommend to use your existing database's primary key as the same in TigerBeetle and avoid a mapping?

cb22: Ah, we actually do - https://github.com/tigerbeetle/tigerbeetle/blob/main/docs/design/data-modeling.md#reuse-foreign-identifier - maybe we should reference that here rather?

docs/design/data-modeling.md
```diff
@@ -78,20 +78,33 @@ When selecting an `id` scheme:
 - An account and a transfer may share the same `id` — they belong to different "namespaces".
 - Avoid requiring a central oracle to generate each unique `id` (e.g. an auto-increment field in SQL).
   A central oracle may become a performance bottleneck when creating accounts/transfers.
+- Sequences of identifiers with long runs of strictly increasing values are amenable to very
```
cb22: I'm not sure if we want to speak about internals like compaction at the data-modeling level of docs?

batiati:
Just a reminder for later, we should remove/replace all usage of Guid in our samples. There are a few cases only

docs/design/data-modeling.md
```diff
@@ -159,3 +172,8 @@ This technique enables a simple range query to iterate every object
 with a target role. While you can't yet do prefix queries within
 TigerBeetle, you could do a prefix query in an external database to
 resolve `id`s and pass the resolved `id`s into TigerBeetle.
+
+#### Random Identifier
```
cb22, line 172: Just below the "Examples (Advanced)" section, we still stay Random IDs are recommended for most applications and link here.

(GitHub won't let me comment there :sweat:)

docs/design/data-modeling.md
sentientwaffle: I wanted to provide the motivation for the recommendation -- especially since it is a complete change from our former recommendation.

But I can be a bit more abstract, just saying that they lead to higher database throughput. Since that is visible to the user, and this is a section about choosing (or designing your own) id schema, I think it is relevant info.

docs/design/data-modeling.md
```diff
@@ -78,20 +78,33 @@ When selecting an `id` scheme:
...
+
+A ULID ("Universally Unique Lexicographically Sortable identifier") consists of:
+
+- 48 bits of (millisecond) timestamp (high-order bits)
+- 80 bits of randomness (low-order bits)
 
-Randomly-generated identifiers are recommended for most applications.
+**Important**: When creating multiple objects during the same millisecond, increment the random bytes
+(instead of generating new random bytes). This ensures that a sequence of objects within a
+[batch](./client-requests.md#batching-events) have strictly increasing ids. (Batches of strictly
```
jorangreef:
```suggestion
[batch](./client-requests.md#batching-events) has strictly increasing ids. (Batches of strictly
```

docs/design/data-modeling.md
```diff
@@ -78,20 +78,33 @@ When selecting an `id` scheme:
...
+increasing ids are amenable to LSM optimizations, leading to higher database throughput).
 
-- Random identifiers require coordination with a secondary database to implement idempotent
+- ULIDs require coordination with a secondary database to implement idempotent
   [application crash recovery](./consistency.md#consistency-with-foreign-databases).
```
jorangreef: This is not so much a property of ULIDs or UUIDs, rather the link is dealing with how applications think about idempotency.

Idempotency here also technically doesn't require a secondary database. It may. It may not. It's more about how the application handles the `.exists` or `.ok` codes.

Would be great to point to this link, with something like a “Remember to think about idempotency in your application.”

jorangreef: At present, this paragraph could be misunderstood as saying that using UUIDs or ULIDs in TigerBeetle therefore requires a secondary database (to coordinate idempotency), but this is not correct.

docs/design/data-modeling.md
```diff
@@ -159,3 +172,8 @@ This technique enables a simple range query to iterate every object
 with a target role. While you can't yet do prefix queries within
 TigerBeetle, you could do a prefix query in an external database to
 resolve `id`s and pass the resolved `id`s into TigerBeetle.
+
+#### Random Identifier
+
+Random identifiers are not recommended – they are incompatible with some LSM optimizations.
```
jorangreef: They are not incompatible (they still work), but they don't take advantage of all of TigerBeetle's optimizations.

jorangreef: Would be great to motivate this more (and quantify impact). For example, UUIDs require more work to do negative lookups, resulting in +/- 5-10% reduction in performance.

jorangreef (commented):
Well spotted DJ.

docs/design/data-modeling.md
cb22: > resulting in +/- 5-10% reduction in performance.

Some real numbers from `main`, on my laptop:
35s sequential vs 43s random on tmpfs
52s sequential vs 60s random on nvme

docs/design/data-modeling.md
```diff
@@ -78,20 +78,31 @@ When selecting an `id` scheme:
 - An account and a transfer may share the same `id` — they belong to different "namespaces".
```
jorangreef: I remember this was a discussion a few months back.

Can you remember if there was any advantage to allowing/suggesting that accounts/transfers may collide IDs?

I would rather we go the other way on this, to warn against schemes that do this, because it:

* increases surface area (we can always expand, but it's harder to restrict later),
* may be incompatible with single namespaces when exporting into other systems, and
* breaks the principle of “universally” unique (doesn't feel right).

jorangreef: The reason it doesn't feel right to me is that, for example, in cryptography, there's the principle of "single use":

Only ever use something (like a symmetric key) for one purpose, one intent, and bind the use to that intent (e.g. for encryption in transit, use a HKDF to take a master symmetric key and derive 4 keys, 2 per direction, bound to the direction of traffic along the network connection). When that's violated, a CVE is typically not far off.

In our context, if a transfer is accidentally specified as an account, or vice-versa, and reinterpreted like that against the wrong data, then that could be bad. There's one less defense-in-depth, compared to if the IDs are unique.

docs/design/data-modeling.md
```diff
@@ -78,20 +78,31 @@ When selecting an `id` scheme:
...
 - Avoid requiring a central oracle to generate each unique `id` (e.g. an auto-increment field in SQL).
   A central oracle may become a performance bottleneck when creating accounts/transfers.
+- Sequences of identifiers with long runs of strictly increasing values are amenable to
+  optimization, leading to higher database throughput.
```
jorangreef: We also optimize for strictly decreasing values.

docs/design/data-modeling.md
```diff
@@ -78,20 +78,31 @@ When selecting an `id` scheme:
...
+
+- 48 bits of (millisecond) timestamp (high-order bits)
+- 80 bits of randomness (low-order bits)
+
+**Important**: When creating multiple objects during the same millisecond, increment the random bytes
+(instead of generating new random bytes). This ensures that a sequence of objects within a
+[batch](./client-requests.md#batching-events) has strictly increasing ids. (Batches of strictly
+increasing ids are amenable to LSM optimizations, leading to higher database throughput).
+
+- ULIDs have an insignificant risk of collisions.
```
jorangreef:
```suggestion
- ULIDs have an insignificant risk of collision.
```

jorangreef (dismissed):

docs/design/data-modeling.md
sentientwaffle: I completely agree that we shouldn't recommend it. My intent here was not to suggest that users deliberately collide account+transfer IDs, just to state that it is possible (because it is).

Reworded to:

> - An account and a transfer may share the same `id` (they belong to different "namespaces"),
>  but this is not recommended.

Alternatively, I can remove that bullet entirely.

jorangreef: Thanks, that makes sense. We should just motivate why it's not recommended. If we say why, then it gives people a framework to evaluate the recommendation and makes them more likely to understand.

For example, "but this is not recommended because other systems (that you may later connect to TigerBeetle) may use a single "namespace" for all objects."

batiati:
I'd suggest we use the term "Time-based IDs" in contrast to "Random IDs".
This broader category includes ULID and other implementations such as [UUID v7](https://www.ietf.org/archive/id/draft-peabody-dispatch-new-uuid-format-01.html#name-uuidv7-timestamp-usage), [SnowFlake ID](https://en.wikipedia.org/wiki/Snowflake_ID), [KSUID](https://segment.com/blog/a-brief-history-of-the-uuid/), etc.

* I'm not suggesting we recommend any of those IDs; some are unsuitable (eg, KSUID is 20 bytes long, SnowFlake is 64-bit), but UUID v7 is a suitable option and has gained traction lately.

jorangreef (approved):

PR #1201, opened 2023-09-22
build: redo releases & client testing

matklad:
The original goal here was just to improve our publishing process, but this escalated into general de-GitHub-actions-ification.

Not fully finished yet, here's where it is going:

Overall philosophy
------------------

- Avoid building a distributed system if we can. Zig can cross-compile, so we have a fighting chance of keeping everything on a single machine. Maaaybe things like glibc ABI will push us against cross compiling literally everything, but we should try!

- Use Zig, both for positive reasons (build up a re-usable ecosystem for everything zig) and negative (avoid GHA lock-in).

- Use CI _only_ for distributing work across machines with different configurations (as a consequence of the first two points).

- Embrace essential variability. We are not building software to run on our tightly controlled hardware. We are building software for other people to run elsewhere. So, don't try to precisely control the environment, and instead try to ask as little of it as possible.

Build process for releases
--------------------------

Three steps:

* `zig build dist` creates a `./dist` folder with a "distribution" of TigerBeetle. Distribution includes TigreBeetle binary cross-compiled for all supported targets, as well as package archive for all clients (`.nuget` for dotnet, `.jar` for java, `.tar.gz` for node and just a folder with sources and pre-built binaries for go).

  `zig build dist` runs on a single computer, but, because the result is just a directory with files, we should be able to distribute this, if needed (but the plan is to try to avoid that, if at all possible).

* `zig build dist --publish` publishes relevant parts of the distribution to the relevant registries.

* `zig build client_ci --verify-release` runs _after_ publishing and sanity checks that the stuff in remote registries makes sense.

* There's intentionally no testing during release process. All the testing we currently do should be done _before_ the code gets to a releasable state. Nonetheless, the dist/publish separation will allow us to run extra QA on the actual binaries we release.

Release process
---------------

* Releases process is triggered once a new commit is pushed to the `release` branch.
* Releases are manual and weekly. Every Monday, someone pushes current `main` to `release`
* Weekly release is accompanied by a changelog. Changelogs are stored in `changelog/` folder using `changelog/$(date --iso).md` naming scheme (with weekly changelogs, using a single file will not last us many years).
* There are no nightly releases --- weekly cadence should be fast enough.
* Because releases are weekly, CI's run number is used for the patch version. Version doesn't appear in the source code, as weekly releases would require too many changes.

Details
-------

`src/scripts/dist.zig` is a script which manages building a distribution and publishing. This is a stand-alone script, rather than a build.zig task, because it orchestrates different build systems.

`src/clients/ci.zig` is a script which orchestrates CI for clients. It delegates to `src/clients/$LANG/ci.zig`.

Client testing is now run as a part of core pipeline. If we want releases, we _must_ test our code properly. I am fairly certain we can make this non-flaky and fast, especially given that tests now don't try to do things, unrelated to testing.

Client testing includes running "unit tests" withing the project itself, as well as running all our samples with a real tigerbeetle.

I am not sure what matrix we should go for! Right now, I am using GitHub and self-hosted runners, but I think it might make sense to use upstream langauge-specific docker containers, as in
https://docs.github.com/en/actions/using-jobs/running-jobs-in-a-container for the purpose of embracing essential variability

Publishing docker images is TODO


.github/workflows/clients.yml
```diff
@@ -0,0 +1,70 @@
...
+    strategy:
+      matrix:
+        os: [ubuntu-latest, windows-latest, macos-latest, [self-hosted, ARM64, macos-12.6], [self-hosted, ARM64, macos-13.2]]
+
+      runs-on: ${{ matrix.os }}
+      steps:
+        - uses: actions/checkout@v3
+        - uses: actions/setup-node@v3
+        - run: ./scripts/install_zig.sh
+        - run: ./zig/zig build client_ci -- --language=node
```
matklad: I guess this is the core thing here, conceptually. This is how I relatively strongly feel our CI should look: define the matrix in `.yaml`, use upstream dependencies (`setup-x` actions), just like how our users would, launch the zig process and then stack away from yml.

sentientwaffle:
> Weekly release is accompanied by a changelog. Changelogs are stored in `changelog/` folder using `changelog/$(date --iso).md` naming scheme (with weekly changelogs, using a single file will not last us many years).

Apologies for bikeshedding, but can you elaborate on this? I would expect that "one big changelog file" would scale better to frequent releases than "one small changelog file per release".

It's also more convinient (at least imo) if you are an operator trying to see what has changed since your last version. Release `X` includes all changes `< X`, so you just keep reading down until you hit the version you're upgrading from (or vice-versa, depending whether the file is ordered chronologically or reverse-chronologically).

matklad:
That’s not bikeshedding, that’s important, thanks for raising the point.

rust lang uses a single for the changelog, and, with release every six weeks, GitHub no longer can always display this file:

https://internals.rust-lang.org/t/rust-1-72-1-pre-release-testing/19566/4?u=matklad

With our aim to have more frequent releases, I think we’ll hit this problem sooner.

My other bit of thinking here is that we might want to pull reach separate changelog to a page on the website, so that it can be tweeted and such, a-la https://rust-analyzer.github.io/thisweek. This also feels easier to do with separate files.

no super strong feelings here, can go with traditional single file!

sentientwaffle:
:+1: I don't feel strongly about it either.

> My other bit of thinking here is that we might want to pull reach separate changelog to a page on the website, so that it can be tweeted and such, a-la https://rust-analyzer.github.io/thisweek. This also feels easier to do with separate files.

Very good point, this tipped me in favor of multiple files.

:thinking: I wonder if the changelog belongs under `docs/` -- and if we are going to be linking to the changelog externally, it could be built into docs.tigerbeetle.com too.

Aside: https://github.com/rust-lang/rust/blob/master/RELEASES.md is only 13k lines, which is not that much more than `replica.zig`!

---

One more thing to consider is how the changelog gets updated:
- Is it updated as part of the same PR that implements the change?
- Or do we go and update it later, all at once, just before each release?

matklad:
Yeah, this PR adds docs/changelog, and, yeah, the idea is to expose that to tb.com!

> One more thing to consider is how the changelog gets updated:

Unclear! There’s value in both “everyone adds stuff to changelog” and “changelog editor compiles the weeks worth of PRs”.

id maybe go with the second one by default: that’s a lot more work, but higher quality: if someone *else* writers about your PR in the changelog, this acts as a forcing function to create externally relevant&understandable descriptions

jorangreef:
> There’s value in both “everyone adds stuff to changelog” and “changelog editor compiles the weeks worth of PRs”.

Let's give it a try and do both!

* In the spirit of "everything on main is good to go", PRs also add to `changelog` or else put `Changelog: Skipped` in the commit, and we lint that.

* In the spirit of "editing and quality", before a release there can be a commit to edit the changelog if need be.

jorangreef:
> each separate changelog to a page on the website, so that it can be tweeted and such, a-la https://rust-analyzer.github.io/thisweek. 

To solve this, and still have a single changelog file, we could have the latest release on the one pager changelog have a "latest" anchor.

If it helps, I'd prefer a single changelog file (one less directory, and less files)—this also increases the probability that someone carries on reading into the next one. If a single changelog doesn't work, we could then split it out later if need be. Or archive previous years' changelogs etc.

> Yeah, this PR adds docs/changelog, and, yeah, the idea is to expose that to tb.com!

Yes, that is excellent, and to have it at `docs/changelog`.

.github/workflows/clients-dotnet.yml
```diff
@@ -1,242 +0,0 @@
...
-
-      - name: Build dotnet
-        working-directory: src/clients/dotnet
-        run: dotnet build --verbosity normal --configuration Release
-
-      - name: Tests
-        working-directory: src/clients/dotnet
-        run: |
-          dotnet test \
-          /p:CollectCoverage=true \
```
matklad, line 41: Dropped dotnet coverage from CI, as it failed a couple of times, and, as we now gate main on clients' CI passing, flakes are not allowed.

.github/workflows/clients-dotnet.yml
```diff
@@ -1,242 +0,0 @@
...
-          name: nuget-artifact-${{ inputs.version }}
-          path: src/clients/dotnet/TigerBeetle/bin/Release/*.nupkg
-
-  prepare_validation_tests:
-    if: inputs.version == ''
-    needs: build
-    runs-on: ubuntu-latest
-    name: Prepare validation tests
-    steps:
-      - name: Create a simple test project
```
matklad, line 67: Dropped this. It is redundant with checking that our samples work.

.github/workflows/clients-dotnet.yml
```diff
@@ -1,242 +0,0 @@
...
-    runs-on: ${{ matrix.os }}
-    steps:
-      - uses: actions/checkout@v3
-
-      - uses: actions/setup-dotnet@v3
-        with:
-          # Min supported and latest versions.
-          dotnet-version: |
-            3.1.x
-            7.0.x
```
matklad, line 192: TBH, I don't understand which dotnet version would actually be used here and elsewhere.

.github/workflows/clients-dotnet.yml
```diff
@@ -1,242 +0,0 @@
...
-          name: nuget-artifact-${{ inputs.version }}
-          path: ./nuget
-      - run: |
-          dotnet nuget push ./nuget/*.nupkg \
-          --api-key $NUGET_KEY \
-          --source https://api.nuget.org/v3/index.json
-        env:
-          NUGET_KEY: ${{ secrets.NUGET_KEY }}
-
-  samples_integration_test:
```
matklad, line 169: samples are now tested as a part of `zig build client_ci -- --language=dotnet`

.github/workflows/clients-java.yml
```diff
@@ -1,280 +0,0 @@
...
-        java-version: '11'
-        distribution: 'temurin'
-        server-id: ossrh
-        server-username: MAVEN_USERNAME
-        server-password: MAVEN_CENTRAL_TOKEN
-        gpg-private-key: ${{ secrets.MAVEN_GPG_SECRET_KEY }}
-        gpg-passphrase: MAVEN_GPG_PASSPHRASE
-        settings-path: ${{ github.workspace }}
-
-    - name: Format linter
```
matklad, line 45: Moved to `client_ci`

.github/workflows/clients-java.yml
```diff
@@ -1,280 +0,0 @@
...
-    - name: Restore local package
-      uses: actions/download-artifact@v3
-      with:
-        name: jar-artifact-${{ inputs.version }}
-        path: test/
-
-    - name: Create a simple test program
-      run: |
-        mkdir -p test
-        cat <<"EOF" > test/Main.java
```
matklad, line 100: redundant with samples testing

.github/workflows/clients-dotnet.yml
```diff
@@ -1,242 +0,0 @@
...
-          dotnet add package tigerbeetle -s ../nuget
-      - uses: actions/download-artifact@v3
-        with:
-          name: test-project-artifact
-          path: ./test-project
-      - name: Test nuget package
-        working-directory: ./test-project
-        run: dotnet run
-
-  run_validation_tests_on_containers:
```
matklad, line 116: I am killing container tests for now and will restore then in the follow up. 

My original plan was to treat containers as OSes, and just add more lines to GHA matrix. That doesn't work, for two reasons:

* practically, on, eg, amazoncorretto, checkout action fails because the container includes neither git nor tar
* philosophically, what `zig build client_ci` tests is that the code both builds _and_ runs on host. I don't think we should constraint our build process to work on all containers. So, we fundamentally need to compile stuff in one environment, and then deploy it into the container. I imagining adding a `--container` flag to `client_ci`, which would then build on host, and then spawn docker internally to run things.

.github/workflows/clients-dotnet.yml
```diff
@@ -1,242 +0,0 @@
...
-          path: ./nuget
-
-      # Grab Zig
-      - run: ./scripts/install.${{ matrix.os == 'windows-latest' && 'bat' || 'sh' }}
-
-      # Build and run the integration tests
-      - run: ./zig/zig build client_integration
-      - run: ./zig-out/bin/client_integration${{ matrix.os == 'windows-latest' && '.exe' || '' }} --language=dotnet --sample=${{ matrix.sample }}
-
-  docs:
```
matklad, line 206: I keep the client docs setup mostly as is. I do want to fold that into `client_ci`, but that's going to be a follow  up.

.github/workflows/clients.yml
```diff
@@ -0,0 +1,109 @@
...
+      - uses: actions/checkout@v3
+      - uses: actions/setup-dotnet@v3
+        with:
+          dotnet-version: ${{ matrix.dotnet-version }}
+
+      - run: ./scripts/install_zig.${{ matrix.os == 'windows-latest' && 'bat' || 'sh' }}
+      - run: ./zig/zig build client_ci -- --language=dotnet
+      - run: ./zig/zig build client_docs -- --language=dotnet
+      - if: matrix.os == 'ubuntu-latest'
+        run: ./.github/ci/fail_on_diff.sh
```
matklad: I want to fold docs into `client_ci`, but that's for a follow up, as I want to do a somewhat large refactor there (basically, the way the code works today is that pieces of reademe are concatenated to form an executable example. I'd rather we have a full example in the repo, and generate the readme by extracting parts of that.

.github/workflows/clients.yml
```diff
@@ -0,0 +1,109 @@
+name: Clients
```
matklad, line 1: This is intentionally a single file now, it shouldn't get larger, and there shouldn't be any differences in logic between the clients.

.github/workflows/pipeline-core.yml
```diff
@@ -34,15 +37,16 @@ jobs:
     needs:
       - smoke
       - linux
+      - clients
```
matklad, line 40: Big one: main is now gated on clients' passing.

.github/workflows/release.yml
```diff
@@ -1,113 +1,52 @@
...
-        required: true
-      SLACK_TOKEN:
-        required: true
-    inputs:
-      version:
-        required: false
-        type: string
+  push:
+    branches:
+      - release
```
matklad, line 6: Releases are triggered by pushes to the release branch. long term, we want something here to make sure at least two people sing off on a release, to protect from any single person's laptop being hacked.

.github/workflows/release.yml
```diff
@@ -1,113 +1,52 @@
...
+          server-password: MAVEN_CENTRAL_TOKEN
+          gpg-private-key: ${{ secrets.MAVEN_GPG_SECRET_KEY }}
+          gpg-passphrase: MAVEN_GPG_PASSPHRASE
+
+      # No special setup for Go.
+
+      - uses: actions/setup-node@v3
+        with:
+          node-version: '14.x'
+          registry-url: 'https://registry.npmjs.org'
```
matklad: Release action needs all languages.

.github/workflows/release.yml
```diff
@@ -1,113 +1,52 @@
...
+
+      - run: ./scripts/install_zig.sh
+      - run: ./zig/zig build dist -- --build --publish --version 0.9.${{ github.run_number }} --sha ${{ github.sha }}
```
matklad, line 43: That's the single place in code where we define version. use 0.9 for now to not mess our latest release.

matklad, line 43: Both `--build` and `--publish` at the same time for now. That's the place where we would be able to insert `--and-thoroughly-test` in the future. This is the single place where we might want to use `upload-artifact` action.

.github/workflows/release_validate.yml
```diff
@@ -1,101 +1,33 @@
...
-        uses: actions/setup-java@v3
-        with:
-          java-version: '11'
-          distribution: 'temurin'
+      - run: ./scripts/install_zig.sh
+      
+      - run: ./zig/zig build dotnet_client:ci -- --release-validate
+      - run: ./zig/zig build go_client:ci     -- --release-validate
+      - run: ./zig/zig build java_client:ci   -- --release-validate
+      - run: ./zig/zig build node_client:ci   -- --release-validate
```
matklad, line 23: That's going to be completely broken for know, as we don't have a proper release with u128 balances. 


But yeah, the idea here is to run this on the clock, to test latest releases.

src/clients/dotnet/samples/basic/Basic.csproj
```diff
@@ -5,10 +5,15 @@
     <ImplicitUsings>enable</ImplicitUsings>
     <Nullable>enable</Nullable>
     <RollForward>LatestMajor</RollForward>
+    <UseCurrentRuntimeIdentifier>true</UseCurrentRuntimeIdentifier>
+    <InvariantGlobalization>true</InvariantGlobalization>
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="tigerbeetle" Version="0.13.88" />
+    <Content Include="..\..\TigerBeetle\runtimes\$(RuntimeIdentifier)\native\*.*">
```
matklad, line 13: Switch from using a random release, to using the code from the repo.

src/clients/dotnet/samples/basic/Basic.csproj
```diff
@@ -5,10 +5,15 @@
     <ImplicitUsings>enable</ImplicitUsings>
     <Nullable>enable</Nullable>
     <RollForward>LatestMajor</RollForward>
+    <UseCurrentRuntimeIdentifier>true</UseCurrentRuntimeIdentifier>
+    <InvariantGlobalization>true</InvariantGlobalization>
```
matklad, line 9: Needed to avoid linking ICU

src/clients/go/samples/basic/go.mod
```diff
@@ -2,4 +2,6 @@ module basic
 
 go 1.17
 
-require github.com/tigerbeetle/tigerbeetle-go v0.13.78
+require github.com/tigerbeetle/tigerbeetle-go v0.0.0
+
+replace github.com/tigerbeetle/tigerbeetle-go => ../../
```
matklad, line 7: Same here, remove mentioning of the version, use the code from the repo instead.

src/clients/java/samples/basic/pom.xml
```diff
@@ -39,8 +39,7 @@
     <dependency>
       <groupId>com.tigerbeetle</groupId>
       <artifactId>tigerbeetle-java</artifactId>
-      <!-- Grab the latest commit from: https://repo1.maven.org/maven2/com/tigerbeetle/tigerbeetle-java/maven-metadata.xml -->
-      <version>0.0.1-3431</version>
+      <version>0.0.0</version>
```
matklad: I wasn't able to find a clean way to use "path dependencies" with maven, so this is just a placeholder, and some extra zig code is needed so that maven finds the correct code. 

Still, we avoid mentioning a random version.

src/clients/java/ci.zig
```diff
@@ -0,0 +1,60 @@
...
+
+    try shell.exec(
+        \\mvn deploy:deploy-file --quiet
+        \\  -Durl=file://{maven_repo_path}
+        \\  -Dfile=target/tigerbeetle-java-0.0.1-SNAPSHOT.jar
+        \\  -DgroupId=com.tigerbeetle -DartifactId=tigerbeetle-java -Dversion=0.0.0
+        \\  -Dpackaging=jar
+    , .{
+        .maven_repo_path = maven_repo_path,
+    });
```
matklad: Work-around for the lack of path deps in maven.

src/clients/node/samples/basic/package.json
```diff
@@ -1,5 +1,5 @@
 {
   "dependencies": {
-    "tigerbeetle-node": "^0.13"
+    "tigerbeetle-node": "file:../../"
```
matklad, line 3: Remove random version, use a path dependency.

src/shell.zig
```diff
@@ -317,39 +384,89 @@ pub fn git_tag(shell: *Shell) ![]const u8 {
     return stdout;
 }
 
-const Argv = std.ArrayList([]const u8);
+const Argv = struct {
```
matklad, line 387: Argv is changed to suport syntax like `git tag v{version}`, which requires runtime-concatenation of `v` to stringified interpolation variable.

src/shell.zig
```diff
@@ -97,6 +102,18 @@ pub fn echo(shell: *Shell, comptime fmt: []const u8, fmt_args: anytype) void {
...
+}
+
+pub fn env_get(shell: *Shell, var_name: []const u8) ![]const u8 {
+    errdefer {
+        log.err("environment variable '{s}' not defined", .{var_name});
+    }
+
+    return try std.process.getEnvVarOwned(shell.arena.allocator(), var_name);
+}
+
```
matklad, line 116: Somewhat random utilities, I am still not entirely sure what's the desired end state for the shell is. But the guidelines are:

* convencience
* error reporting
* avoiding error-prone memory management (by allocating into shell-owned arena)

.github/workflows/clients-dotnet.yml
cb22: cc @batiati - from what I understand, we support a minimum of `3.1.x`, `6.0.x` is the LTS release, and `7.0.x` is the latest. We define a custom UInt128 handler if < 7: https://github.com/tigerbeetle/tigerbeetle/pull/1157/files#diff-a6230f0d0a0e6f295c398ae2edaab541ce487794961e802a71038c408ceb2172

.github/workflows/clients-dotnet.yml
cb22: > practically, on, eg, amazoncorretto, checkout action fails because the container includes neither git nor tar

I think we could install dependencies like that that we'd expect a user to install or have installed. I think the original idea behind testing inside them was along the lines of "does Amazon Linux ship some weird janky version of Node that breaks building TigerBeetle".

There are two separate things here:
* Using the container as a proxy for a full OS (the paragraph mentioned above) to test the build process.
* Using the container as an end user would actually use it, to test that TigerBeetle works inside of it.

Debatable if the first point is worth it!

.github/workflows/release.yml
cb22: I feel we should definitely set it up as a protected branch on GitHub. I know I'll try accidentally push directly to it at least once :smile: 

Not sure if you've checked out the rules there but might be easy enough to require two person signoff etc too.

.github/workflows/release.yml
cb22: > That's the place where we would be able to insert --and-thoroughly-test in the future.

(if you're speaking about `--and-thoroughly-test` testing still-bounded-tests-that-take-a-while as opposed to things like fuzzing, VOPR, than the below doesn't apply!)

Jobs are limited to ~6 hours on GitHub, so we would need to split it up (also, we don't want to incur the waste of an idle runner while the tests are really being run on different hardware!)

.github/workflows/release_validate.yml
cb22: > But yeah, the idea here is to run this on the clock, to test latest releases.

Do we not also want to run it after publish, to catch any issues quicker?

src/clients/java/samples/basic/pom.xml
cb22: Perhaps a comment saying that?

src/clients/java/ci.zig
cb22: I know it's a placeholder, but for `target/tigerbeetle-java-0.0.1-SNAPSHOT.jar` should we not use `target/tigerbeetle-java-0.0.0-SNAPSHOT.jar`?

.github/workflows/clients-dotnet.yml
batiati: Our baseline support is .Net Standard 2.1, which isn't a runtime but a spec supported by multiple runtimes such as Mono, Unity, .Net Core, and  Xamarin. We also support  >= .Net 7.0, to take advantage of some features like U128 type and unmanaged function pointers.

That's why we need two different runtimes for validating the client. We actually distribute two different binaries in the same package.

.github/workflows/release.yml
cb22: To add to this: it would be great to move our secrets across from being global secrets, to have a specific `release` environment that matches the branch and has the secrets.

That way, any random branch won't be able to access them.

.github/workflows/clients-dotnet.yml
batiati: We have two main reasons for testing multiple OS:

- Test if our build works on Linux, MacOS, and Windows.
- Test if our binary can run on multiple OS/distro/kernels/client runtimes.

Those container tests are supposed to test the artifact already produced, I think they don't need to include any of the build dependencies, only the target client dependencies e.g. java, node, dotnet.

.github/workflows/clients.yml
```diff
@@ -0,0 +1,124 @@
...
+  group: clients-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+on:
+  workflow_call:
+
+jobs:
+  client-dotnet:
+    strategy:
+      matrix:
```
cb22, line 14: We might want to sprinkle `max-parallel:` around the matrix jobs here - so that a single commit / run doesn't use all the runners available to us.

.github/workflows/clients.yml
```diff
@@ -0,0 +1,124 @@
...
+      - run: ./zig/zig build client_ci -- --language=go
+      # - run: ./zig/zig build client_docs -- --language=go
+      # - if: matrix.os == 'ubuntu-latest'
+        # run: ./.github/ci/fail_on_diff.sh
+
+  client-java:
+    strategy:
+      matrix:
+        os: [ubuntu-latest, windows-latest, macos-latest, [self-hosted, ARM64, macos-12.6], [self-hosted, ARM64, macos-13.2]]
+        java-version: ['11' ]
```
cb22:
```suggestion
        java-version: ['11']
```

.github/workflows/clients.yml
```diff
@@ -0,0 +1,124 @@
...
+      - run: ./scripts/install_zig.${{ matrix.os == 'windows-latest' && 'bat' || 'sh' }}
+      - run: ./zig/zig build client_ci -- --language=dotnet
+      # - run: ./zig/zig build client_docs -- --language=dotnet
+      # - if: matrix.os == 'ubuntu-latest'
+        # run: ./.github/ci/fail_on_diff.sh
+
+  client-go:
+    strategy:
+      matrix:
+        os: [ubuntu-latest, windows-latest, macos-latest, [self-hosted, ARM64, macos-12.6], [self-hosted, ARM64, macos-13.2]]
```
cb22: Since macOS 14 is out now, I think we can drop the 12 ones (and we should add 14). We can also use GitHub's own hosted M1 runners now...

src/clients/go/ci.zig
```diff
@@ -0,0 +1,71 @@
...
+const log = std.log;
+const assert = std.debug.assert;
+
+const flags = @import("../../flags.zig");
+const fatal = flags.fatal;
+const Shell = @import("../../shell.zig");
+const TmpTigerBeetle = @import("../../testing/tmp_tigerbeetle.zig");
+
+pub fn tests(shell: *Shell, gpa: std.mem.Allocator) !void {
+    // No unit tests for Go :-(
```
cb22, line 12: Been a while since I've seen a nose :smile:

docs/internals/releases.md
```diff
@@ -0,0 +1,57 @@
...
+- artifacts are uploaded to GitHub releases, npm, Maven Central, and Nuget. For Go, a new commit
+  is pushed to <https://github.com/tigerbeetle/tigerbeetle-go>
+- if publishing to all registries were successfully, the release is marked as non-draft.
+
+All publishing keys are stored as GitHub Actions secrets.
+
+## Triggering a Release
+
+Releases are triggered manually, on Monday, by pushing the the current tip of `origin/main` to
+`origin/releases`. This triggers the `release.yml` workflow. The workflow doesn't run unit-tests,
```
cb22:
```suggestion
`origin/release`. This triggers the `release.yml` workflow. The workflow doesn't run unit-tests,
```

cb22:
> There's intentionally no testing during release process. All the testing we currently do should be done before the code gets to a releasable state. Nonetheless, the dist/publish separation will allow us to run extra QA on the actual binaries we release.

and 

> Releases are triggered manually, on Monday, by pushing the the current tip of `origin/main` to `origin/releases`. This triggers the `release.yml` workflow. The workflow doesn't run unit-tests, [not rocket science](https://graydon2.dreamwidth.org/1597.html) rule is assumed.

Do you have something in mind for how we'll handle the testing-before-releasing code? I'm thinking of the long running stuff specifically when we have a time budget for testing (bounded things like unit tests should be every commit and simple enough).

matklad:
>Do you have something in mind for how we'll handle the testing-before-releasing code? I'm thinking of the long running stuff specifically when we have a time budget for testing (bounded things like unit tests should be every commit and simple enough).

Yup! The "push to release" workflow runs `dist --build`, but not `dist --publish`. The resulting dist is then uploaded as GHA artifact. Any further testing works with this uploaded artifact, to enforce the invariant that we release exactly the thing we've tested. After that extra round of testing (which might be arbitrary long), another action runs `dist --publish` (but *not* `dist --build`) to upload the tested artifacts.

.github/workflows/release_validate.yml
matklad: Yes, though what gives me pause is "eventual consistency" of publishing process --- my understnading is that some package registiries don't guarantee that a new version becomes available immediately after publishing. But we probably should do something along the lines of "run 10 minutes after the publish".

src/clients/java/ci.zig
matklad: Figured out how to solve this in a better way, with less version patching!

.github/workflows/clients.yml
matklad: Left for the follow up

cb22: If anyone comes back to this thread from the future, it was done here in the end!

cb22 (approved):
@matklad looks good to me! Only two things are if you want to do these now or leave for a follow-up:
* Use a GH environment for limiting secret access for `release` branch
* Protecting the `release` branch

(I actually think protecting it should wait... It'll make testing `release` much easier for you with the push-cycle)

matklad:
Those are for follow ups! But I remember e v e r y t h i n g!

PR #1202, opened 2023-09-24
Add eager parsing with clear indication of input

### LINKED ISSUE #1185: REPL: Eager parser erroring / clear indication of expected input
When hitting enter on the REPL, no prompt is displayed, but commands can still be entered:

```
>

lookup_accounts id=1;
{
  "id": "1",
  "user_data": "0",
  "ledger": "700",
  "code": "10",
  "flags": [],
  "debits_pending": "0",
  "debits_posted": "10",
  "credits_pending": "0",
  "credits_posted": "0",
  "timestamp": "1694419483145145501"
}
```

Hitting enter with no previous input should still print out the same prompt (`>`).

On a similar note, there's no indication a semi-colon is expected to end and process the line:
```
> create_accounts id=1 code=10 ledger=700,
[waiting for input here]
```

We should have a continuation prompt, similar to how Python uses `>>>` but switches to `...`. Lastly, entering garbage and hitting enter gives no indication until a semicolon is entered:

```
> asdf

;
Fail near line 1, column 0:
```

pragmaticanon:
This fixes #1185, with the following changes:
1. Make \ a separating delimiter, so when you hit enter after a trailing \, the cli will show >>> (this will fix 1 and 3)
2. Still parse each statement and throw error if there's a syntax error (this will fix 2)
3. Run all queries when a ; is entered at the end of any statement


cb22:
Hey @pragmaticanon - thanks for the PR!

Do you think you could make it so `\` isn't a required delimiter, but it still eagerly parses on newline and aborts if there's any error? Maybe the newline could be special cased?

We're going for roughly the same behaviour as eg `psql` / `mysql`. (although, neither of those will give you an early indication something is wrong)

pragmaticanon:
Sure yeah makes sense, will update

PR #1203, opened 2023-09-25
Test/StorageChecker: Cleanup; verify grid block padding

sentientwaffle:
I recommend reviewing the commits independently -- the first looks big because it un-indents the `StorageChecker` methods.

### DIFF
--- a/src/vsr/grid.zig
+++ b/src/vsr/grid.zig
@@ -615,10 +615,14 @@ pub fn GridType(comptime Storage: type) type {
                 .write = write,
             };
 
+            const write_header = schema.header_from_block(write.block.*);
+            assert(write_header.size > @sizeOf(vsr.Header));
+            assert(write_header.size <= constants.block_size);
+
             grid.superblock.storage.write_sectors(
                 write_block_callback,
                 &iop.completion,
-                write.block.*,
+                write.block.*[0..vsr.sector_ceil(write_header.size)],
                 .grid,
                 block_offset(write.address),
             );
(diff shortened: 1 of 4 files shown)


src/vsr/grid.zig
```diff
@@ -957,6 +957,10 @@ pub fn GridType(comptime Storage: type) type {
             if (!header.valid_checksum_body(block_body)) return .invalid_checksum_body;
             if (header.checksum != expect.checksum) return .unexpected_checksum;
 
+            if (constants.verify) {
+                assert(stdx.zeroed(block[header.size..vsr.sector_ceil(header.size)]));
```
matklad, line 965: Why do we go only to the sector_ceil, rather than to the end of the block? I think we always read/write whole blocks, so we should care about padding after the sector boundary as well?

sentientwaffle: :thinking: Is the current behavior (write full block even when `block.header.size < constants.block_size`) desirable? I don't really understand the purpose. Is it worth the write amplification for small blocks?

matklad: Writing / reading full blocks for sure seems simpler, but, yeah, this doesn't feel optimal. 

OTOH, optimizing for blocks always being full also makes sense, as, if writing partial blocks is a big win, the problem is that we have too much partial blocks.

matklad: Let's maybe change `.write_block` to write only what's necessary (but still the whole number of sectors) here?

matklad (approved):

PR #1204, opened 2023-09-25
LSM: Fix SuperBlock.Manifest.tables overflow

sentientwaffle:
Presently manifest compaction can fall behind, eventually overflowing `SuperBlock.Manifest.tables`.

To fix this, only store "active" table extents, instead of both active and inactive table extents.

## Context

`SuperBlock.Manifest.tables` is a hash map tracking each table's latest "extent" – that is, tracking the most recent manifest block which referenced that table.

`SuperBlock.Manifest.tables` presently tracks both:
1. "active" tables: i.e. tables that are referenced by a `ManifestLevel`
2. "inactive" tables: tables that were removed from their LSM tree, but which have not yet been garbage collected by manifest log compaction.

Right now `SuperBlock.Manifest.tables` capacity is defined as:

```zig
try tables.ensureTotalCapacity(allocator, forest_table_count_max);
```

This bound is the maximum number of active tables – it doesn't take into account any inactive tables.
And if `tables` overflows, the replica will crash.

(This is currently reproducible with the forest fuzzer because (in an unrelated bug) manifest compaction can't keep up and falls behind.)

## Fix

Two possible fixes:
1. Count inactive tables in `tables`' worst-case capacity. (~Too expensive; see below.~ Implemented in this PR.)
2. Exclude inactive tables from `tables`, but leave the capacity as-is. (~Implemented in this PR.~)

### Count inactive tables

(This approach is implemented in this PR.)

I originally thought that this approach was too expensive to be viable:

<details>
<summary>Original (incorrect!) worst-case calculation</summary>

The worst-case capacity for `tables` counting both active and inactive tables
(calculation performed with 64KiB blocks for convenience, but it doesn't make a difference in the final result):

      65.5e3 manifest blocks
    × 576    events/block
    × 8+8+8  bytes/extent
    × 2      (hash map load factor)
    = 1.7 GiB

...Ouch.

Counting only active tables:

      2.3e6  tables
    × 8+8+8  bytes/extent
    × 2      (hash map load factor)
    = 105.3 MiB

That 1.7GiB could be tuned down by increasing the manifest block load factor (i.e. compacting the manifest more frequently, reducing the maximum number of manifest blocks) but that can only go so far.

</details>

@maklad found a tighter bound! :fire:

For a particular table in the manifest, the sequence of events is:

    insert(0|1), update(0+), remove(0|1)

We only need to track the addresses of removed tables _until the corresponding "insert" is encountered_.
That, combined with the fact that the maximum number of active tables at any given moment is a constant (`tree.table_count_max`), means that there are at most `table_count_max` "unpaired" removes to count.

Those removes are tracked in a separate hashmap from the table extents – this saves a little memory, and lends itself to farther RAM optimization in the future.

### Exclude inactive tables

(This approach is ~implemented~ not implemented in this PR.)

#### Reverse-chronological manifest recovery

Why does `tables` need inactive tables right now?

The manifest recovery algorithm (`ManifestLog.open()`) is:

    tables = {}
    for manifest_block in reverse(superblock.manifest.blocks):
        for entry in reverse(manifest_block.entries):
            if tables[entry.table]:
                # This table has already been restored.
            else:
                tables[entry.table] = true
                switch entry.event:
                    insert: manifest_levels.insert(entry.table)
                    remove: {}

Manifest entries are replayed in reverse-chronological order.
If a table `T` is deleted at time `Y`, then if we encounter an insert for table `T` at time `X` (`X < Y`), we must ignore it – we must not restore it to the manifest levels.

By replaying events in reverse chronological order, recovery only needs to update the manifest levels once per surviving table.

#### Chronological manifest recovery

The downside of reverse-chronological recovery is that the replica pays (RAM) tracking those (many) inactive tables at all times, not just during recovery.

By replaying events chronologically, the inactive tables are no longer necessary:

    for manifest_block in superblock.manifest.blocks:
        for entry in manifest_block.entries:
            switch entry.event:
                insert: manifest_levels.insert(entry.table)
                remove: manifest_levels.remove(entry.table)

The tradeoff is that we must now update the manifest levels for each event in the manifest log – even events which are later superseded.

## Also in this PR

- `ManifestChecker`: Verify that the `ManifestLevel`s tables are constructed consistently across replicas and after recovering from a restart.
- `Forest.verify_tables_recovered()`: An intrusive (`constants.verify` and `TestStorage`-gated) check which reimplements the chronological manifest recovery algorithm to check `ManifestLevel`s and `SuperBlock.Manifest.tables`.
- `Forest.verify_table_extents()`: An inexpensive sanity-check between `ManifestLevel`s and `SuperBlock.Manifest.tables`.


matklad:
What's the CPU perf impact here? Would be interesting to check if our original reverse chronological recovery actually helped

sentientwaffle:
> What's the CPU perf impact here?

After running the benchmark there are only two manifest blocks (and the second is nearly empty) so I don't have much of a sample size to pull from here :wink: 

That said, for the single full block the measurement is surprisingly stable:
- reverse-chronological is ~250μs
- chronological is ~1200μs

So chronological is about 5x slower. Though that is extremely dependent on the "load factor" of the manifest blocks -- the more often we do manifest compaction, the more chronological's time will approach reverse-chronological.

src/lsm/forest.zig
```diff
@@ -439,5 +451,145 @@ pub fn ForestType(comptime _Storage: type, comptime groove_cfg: anytype) type {
...
+            var forest_tables_iterator = ForestTableIterator{};
+            while (forest_tables_iterator.next(forest)) |forest_table_item| {
+                const table_latest = tables_latest.get(forest_table_item.table.checksum).?;
+                assert(table_latest.level == forest_table_item.level);
+                assert(std.meta.eql(table_latest.table, forest_table_item.table));
+
+                const table_removed = tables_latest.remove(forest_table_item.table.checksum);
+                assert(table_removed);
+            }
+        }
```
matklad, line 594: assert that `tables_latest` is empty after the `while`.

src/lsm/forest.zig
```diff
@@ -439,5 +451,145 @@ pub fn ForestType(comptime _Storage: type, comptime groove_cfg: anytype) type {
...
         }
+
+        pub fn tree_for_id_const(
+            forest: *const Forest,
+            comptime tree_id: u16,
+        ) *const TreeForIdType(tree_id) {
+            const tree_info = tree_infos[tree_id - tree_id_range.min];
+            assert(tree_info.tree_id == tree_id);
+
+            var groove = &@field(forest.grooves, tree_info.groove_name);
```
matklad:
```suggestion
            const groove = &@field(forest.grooves, tree_info.groove_name);
```

matklad: Sometimes I with that Zig just didn't have const/var for locals --- with shallow const, I am skeptical whether it actually prevents bugs.

src/lsm/forest.zig
```diff
@@ -439,5 +451,145 @@ pub fn ForestType(comptime _Storage: type, comptime groove_cfg: anytype) type {
...
+                const table_extent = forest.manifest_log.tables.get(table.table.address).?;
+                assert(table.manifest_block == table_extent.block);
+                assert(table.manifest_entry == table_extent.entry);
+
+                table_extent_counts += 1;
+            }
+            assert(table_extent_counts == forest.manifest_log.tables.count());
+
+            // Verify the tables in `tables` are exactly the tables recovered by the Forest.
+            var forest_tables_iterator = ForestTableIterator{};
```
matklad, line 584: :+1: great check here!

src/lsm/manifest_level.zig
```diff
@@ -261,36 +262,51 @@ pub fn ManifestLevelType(
...
+            if (table.invisible(snapshots)) {
+                return .invisible;
+            } else {
+                level.table_count_visible -= 1;
+                level.key_range.exclude(.{
+                    .key_min = table.key_min,
+                    .key_max = table.key_max,
+                });
+                return .visible;
+            }
```
matklad, line 302: Hm, the API changes here looks suspicious. `level.key_range` should track key range for the latest snapshot, but here the decision to update the key_range also depends on the snapshots passed in, which feels wrong. 

In particular, if a table is not visible to the latest snapshot, but is visible to some older snapshot, we won't update the key range here, which would be wrong. 

And then there's an issue about needing to copy table-info, and the whole "caller pass parameter, we `if` on the parameter, and then the caller asserts the `if` branch". 

Can we simplify this?

It _seems_ that we have two code paths here:

* removing a "tip" table visible to the snapshot_latest
* removing a table which is not visible to snapshot_latest

Whether we have some _other_ snapshots which could see the table I think shouldn't be relevant for this function. The caller should assert that whatever is being removed isn't visible to some old, but still active, snapshot, but it should exists outside of level. 

Relatedly, signature of `remove_invisible_tables` doesn't make sense to me. It should take a slice of snapshots, rather just a single one. At the call site, we pass in `snapshot_latest` to `remove_invisible_tables`, so that, as a result, we call `Table.invisible(&.{snapshot_latest})`. But that's weird, as `Table.invisible` special-cases `snapshot_latest` already

src/lsm/manifest_log.zig
```diff
@@ -115,6 +124,25 @@ pub fn ManifestLogType(comptime Storage: type) type {
 
         next_tick: Grid.NextTick = undefined,
 
+        /// A map from table address to the manifest block and entry that is the latest extent
+        /// version. Used to determine whether a table should be dropped in a compaction.
+        tables: Tables,
```
matklad: Was tripped more than once by the naming here! This aren't tables actually, just extents. `extents: Extents` or `table_extents` would make more sense to me.

src/lsm/manifest_log.zig
```diff
@@ -255,14 +309,29 @@ pub fn ManifestLogType(comptime Storage: type) type {
...
-                if (manifest.insert_table_extent(table.address, block_reference.address, entry)) {
-                    switch (label.event) {
-                        .insert => manifest_log.open_event(manifest_log, label.level, table),
-                        .remove => manifest.queue_for_compaction(block_reference.address),
-                    }
+                if (label.event == .remove) {
+                    assert(manifest_log.tables_removed.get(table.address) == null);
+
+                    manifest.queue_for_compaction(block_address);
+                    manifest_log.tables_removed.putAssumeCapacity(table.address, {});
```
matklad: Maybe `fetchPut` / `getOrPut` to avoid the second hashmap lookup for the sake of the assert?

src/lsm/manifest_log.zig
```diff
@@ -255,14 +309,29 @@ pub fn ManifestLogType(comptime Storage: type) type {
...
+                    if (manifest_log.tables_removed.get(table.address)) |_| {
+                        manifest.queue_for_compaction(block_address);
+
+                        if (label.event == .insert) {
+                            assert(manifest_log.tables_removed.remove(table.address));
+                        }
+                    } else {
+                        var extent = manifest_log.tables.getOrPutAssumeCapacity(table.address);
+                        if (extent.found_existing) {
+                            manifest.queue_for_compaction(block_address);
```
matklad, line 331: `assert(label.event == .update)`

sentientwaffle: I don't think this holds -- the event that created the existing extent was an `.update`, but the one encountered now precedes it chronologically, so it may be an `insert`.

src/lsm/manifest_level.zig
sentientwaffle: Very good point.

`table_count_visible` is in a similar boat right now -- I added a TODO to move it up from `ManifestLevel` to `Manifest`, since in that case we actually _do_ care about visibility relative to all persistent snapshots, not just `snapshot_latest`.

src/lsm/manifest_level.zig
```diff
@@ -160,10 +160,12 @@ pub fn ManifestLevelType(
         tables: Tables,
 
         /// The range of keys in this level covered by tables visible to snapshot_latest.
-        key_range: LevelKeyRange = .{ .key_range = null },
+        key_range_latest: LevelKeyRange = .{ .key_range = null },
```
matklad, line 163: :+1: :+1: :+1:

src/lsm/manifest_level.zig
```diff
@@ -256,29 +258,22 @@ pub fn ManifestLevelType(
...
                 .key_min = table.key_min,
                 .key_max = table.key_max,
             });
         }
 
         /// Remove the given table.
-        /// - If the exact table does not exist in the ManifestLevel: return null.
-        /// - If the exact table exists in the ManifestLevel: remove it, and return whether the
-        ///   table is visible to any of the given snapshots or `snapshot_latest`.
+        /// Returns whether the exact table was present, (and therefore removed).
```
matklad: Maybe just assert that the table is present? I don't think removing non-existant table would make sense, as there are just no tables outside the manifest (and we key removal by the table itself, not by a key)

matklad (approved):

PR #1208, opened 2023-10-02
Benchmark: add the ability to specify ID ordering as a CLI flag

cb22:
When benchmarking, it's very handy to be able to switch this out at runtime to see what different workloads look like.

### DIFF
--- a/src/benchmark.zig
+++ b/src/benchmark.zig
@@ -45,6 +45,7 @@ const StateMachine = vsr.state_machine.StateMachineType(Storage, constants.state
 const Client = vsr.Client(StateMachine, MessageBus);
 const tb = @import("tigerbeetle.zig");
 const StatsD = @import("statsd.zig").StatsD;
+const IdPermutation = @import("testing/id.zig").IdPermutation;
 
 const account_count_per_batch = @divExact(
     constants.message_size_max - @sizeOf(vsr.Header),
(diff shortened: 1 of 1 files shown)


src/benchmark.zig
```diff
@@ -55,11 +55,17 @@ const transfer_count_per_batch = @divExact(
     @sizeOf(tb.Transfer),
 );
 
+/// The ID order can affect the results of a benchmark significantly. Specifically, sequential is
+/// expected to be the best (since it can take advantage of various optimizations such as avoiding
+/// negative prefetch) while random / reversed can't.
+const IdOrder = enum { sequential, random, reversed };
```
batiati, line 62: Wondering if we can use the logic already implemented here, or at least the same names.

https://github.com/tigerbeetle/tigerbeetle/blob/ac79c9c692c2b552ee5d520366eb51213c30569c/src/testing/id.zig#L8-L26

cb22: Ah! I didn't know about `IdPermutation` - I'll see if that's workable.

batiati: The benefit is that you don't need to store an array of IDs because with `IdPermutation` you can retrieve the Id vs index with `decode` and `encode`.

cb22:
@batiati I've swapped in `IdPermutation` - but I think that keeping different names for the CLI is more clear (IMO `--id-order=sequential` / `--id-order=reversed` vs `--id-order=identity` / `--id-order=inversion`)

batiati (approved):

PR #1209, opened 2023-10-02
Fix the client's available messages counter.

### LINKED ISSUE #1206: `c_sample` get errors on debug/release mode
OS: (Arch) Linux
src: https://github.com/tigerbeetle/tigerbeetle/blob/main/src/clients/c/samples/main.c
TB version: Latest (branch main) [0660e5525de03558898153fb33c098fb50c21050]

full-log: [build.log](https://github.com/tigerbeetle/tigerbeetle/files/12753130/build.log)

```bash
debug(client): 1528359687425172202647238492214602118: request: user_data=1735980174421467519501868295783088 request=102 size=160 lookup_accounts
debug(client): 1528359687425172202647238492214602118: send_request_for_the_first_time: request=102 checksum=111310465394886694512212779770963941453
debug(vsr): 1528359687425172202647238492214602118: request_timeout started
debug(client): 1528359687425172202647238492214602118: sending request to replica 0: vsr.Header{ .checksum = 111310465394886694512212779770963941453, .checksum_body = 93813099630386861813656996729618955930, .parent = 165991785639304660942628622715852329879, .client = 1528359687425172202647238492214602118, .context = 1, .request = 102, .cluster = 0, .epoch = 0, .view = 1, .op = 0, .commit = 0, .timestamp = 0, .size = 160, .replica = 0, .command = vsr.Command.request, .operation = vsr.Operation(130), .version = 0 }
debug(client): 1528359687425172202647238492214602118: on_message: vsr.Header{ .checksum = 56396782973848753888938160004986599172, .checksum_body = 67071762243317685171395707297930255356, .parent = 111310465394886694512212779770963941453, .client = 1528359687425172202647238492214602118, .context = 129474760937880992992288287422616544251, .request = 102, .cluster = 0, .epoch = 0, .view = 1, .op = 103, .commit = 103, .timestamp = 1695928489873303648, .size = 384, .replica = 0, .command = vsr.Command.reply, .operation = vsr.Operation(130), .version = 0 }
debug(client): 1528359687425172202647238492214602118: on_reply: user_data=1735980174421467519501868295783088 request=102 size=384 lookup_accounts
debug(vsr): 1528359687425172202647238492214602118: request_timeout stopped
thread 65270 panic: reached unreachable code
/home/kassane/Documentos/tb_test/build/_deps/tb-src/zig/lib/std/debug.zig:343:14: 0x55972546aa0c in assert (tb_client)
             ^
/home/kassane/Documentos/tb_test/build/_deps/tb-src/src/vsr/client.zig:154:19: 0x55972547c1bd in deinit (tb_client)
            assert(self.messages_available == constants.client_request_queue_max);
                  ^
/home/kassane/Documentos/tb_test/build/_deps/tb-src/src/clients/c/tb_client/context.zig:202:35: 0x559725482f85 in deinit (tb_client)
                self.client.deinit(self.allocator);
                                  ^
/home/kassane/Documentos/tb_test/build/_deps/tb-src/src/clients/c/tb_client/context.zig:351:27: 0x55972547c167 in on_deinit (tb_client)
            context.deinit();
                          ^
/home/kassane/Documentos/tb_test/build/_deps/tb-src/src/clients/c/tb_client.zig:222:24: 0x55972546a96a in deinit (tb_client)
    (context.deinit_fn)(context);
                       ^
/home/kassane/Documentos/tb_test/src/main.c:266:5: 0x559725469e39 in main (/home/kassane/Documentos/tb_test/src/main.c)
    tb_client_deinit(client);
    ^
???:?:?: 0x7f8017e8accf in ??? (libc.so.6)
Unwind information for `libc.so.6:0x7f8017e8accf` was not available, trace may be incomplete
```

**Reference:**
- https://github.com/kassane/tigerbeetle-cpp/issues/20
- https://github.com/tigerbeetle/tigerbeetle/issues/487

batiati:
A subtle bug when a request is timed-out and its reply arrives when it is being retransmitted, causes the counter not to be incremented because the request is still referenced by the send queue.

The most common effect could be observed during `client.deinit()` when the assertion `self.messages_available == constants.client_request_queue_max` failed.

- Reworks `client.messages_available` to not rely on the message's reference counter.

- Renames `client.unref` to `client.release` and changes the call site. When a message is acquired from `client.get_message`, it should never call `message.ref()` or `message_bus.unref()` directly.

- Adds `-Dconfig=production` in all the client's builds.

- Adds the missing `TooMuchData` test to the dotnet unit tests.

- Adds support to the environment variable `TB_ADDRESS` to the `c_sample`.
It allows the C sample to work along `run_with_tb`.

Closes #1206


src/vsr/client.zig
```diff
@@ -293,11 +266,58 @@ pub fn Client(comptime StateMachine_: type, comptime MessageBus: type) type {
 
         /// Releases a message back to the message bus.
         pub fn unref(self: *Self, message: *Message) void {
+            // Only request messages are unreferenced by this function,
+            // all others should use `message_bus.unref` directly.
+            assert(message.header.command == .request);
```
batiati: This assumption that `unref` can only accept `.request` heavily depends on how `tb_client` works.
One using `vsr.client` directly can call `get_message` and then `unref` without setting the message header, causing a panic for accessing undefined memory.

Even so, I believe this assertion is worth keeping to avoid misuse between `client.unref`, `bus.unref` and `pool.unref`.

src/vsr/client.zig
```diff
@@ -151,37 +154,8 @@ pub fn Client(comptime StateMachine_: type, comptime MessageBus: type) type {
...
-                .reply => self.on_reply(message),
-                .eviction => self.on_eviction(message),
-                else => {
-                    log.warn("{}: on_message: ignoring misdirected {s} message", .{
-                        self.id,
-                        @tagName(message.header.command),
-                    });
-                    return;
-                },
-            }
```
batiati: Assuming that it's possible for a client to receive a misdirected `.request`, it would incorrectly increment the message counter when unref`d.
We could mitigate it by checking the client and cluster id, but it's still possible to receive a request formatted like that (I can't think of a way for it to happen without a malicious agent, but it's not impossible).

I'm starting to wonder if it's worth keeping `client.messages_available`, or if we could just rely on `client.request_queue.count` to check if we are out of messages.

Assuming that we have calculated the correct maximum number of messages to never exhaust the message pool, even in odds cases such as retransmission (the message is still held by the send queue after leaving the request queue).

batiati: > I can't think of a way for it to happen without a malicious agent, but it's not impossible

😂 Just discovered a particular case when the `recv_message` is unref'd during `deinit`, causing the message counter to be double incremented because the message header was left with the previous `command == .request`.

See the commit message for more info.

src/vsr/client.zig
batiati: Assertion removed, since `client.release` can be called by `errdefer` with an uninitialized message.

src/vsr/client.zig
```diff
@@ -292,11 +292,10 @@ pub fn Client(comptime StateMachine_: type, comptime MessageBus: type) type {
         }
 
         /// Releases a message back to the message bus.
-        pub fn unref(self: *Self, message: *Message) void {
```
matklad, line 295: Let's adjust the docs for get_message to say that the message must end up in release or request, and that it must _not_ be `ref`ed, as this flow is different from how one typically uses the `Message`. "`.ref()` if you store the message" is the general rule we use elsewhere, and here we are deliberately violating that.

src/vsr/client.zig
```diff
@@ -388,12 +387,13 @@ pub fn Client(comptime StateMachine_: type, comptime MessageBus: type) type {
...
             // Eagerly release request message, to ensure that user's callback can submit a new
             // request.
-            self.unref(inflight.message);
-            inflight.message = undefined;
+            self.release(inflight.message);
+            assert(self.messages_available > 0);
 
-            // Even though we release our reference to the message, the user might have retained
+            // Even though we release our reference to the message, we might have retained
             // another one.
```
matklad: Let's mention the `send` queue here; as written, the reader will wonder how _exactly_ do we retain the message.

src/vsr/client.zig
```diff
@@ -234,7 +234,7 @@ pub fn Client(comptime StateMachine_: type, comptime MessageBus: type) type {
             self.request_queue.push_assume_capacity(.{
                 .user_data = user_data,
                 .callback = callback,
-                .message = message.ref(),
+                .message = message,
```
matklad, line 237: I am debating whether we should add `errdefer self.release` for readability and for simmetry with register. I think we'd better not.

src/clients/java/pom.xml
```diff
@@ -243,6 +243,7 @@
                 <argument>build</argument>
                 <argument>java_client</argument>
                 <argument>-Doptimize=ReleaseSafe</argument>
+                <argument>-Dconfig=production</argument>
```
matklad, line 246: Such a great find!

src/vsr/client.zig
batiati: I prefer the pattern `errdefer self.release()` only after `client.get_message()`.

src/clients/java/pom.xml
batiati: It was with @kassane 's help!
Thanks for reporting the bug.

matklad (approved):

PR #1210, opened 2023-10-02
VSR/Format: Zero client_replies zone

sentientwaffle:
Also (unrelated): Don't use `std.mem.zeroInit` in `dvc_blank()`.

The behavior of `std.mem.zeroInit()` changed in zig 0.10:
- previously struct default values were ignored,
- now they aren't -- default values win over zeroes

In this case, `size` was `0` in 0.9, but `128` in 0.10 + 0.11.

Since stability is crucial for `dvc_blank()`, be more explicit about its construction.

### DIFF
--- a/src/vsr.zig
+++ b/src/vsr.zig
@@ -1656,11 +1656,11 @@ pub const Headers = struct {
     pub const ViewChangeArray = ViewChangeHeadersArray;
 
     fn dvc_blank(op: u64) Header {
-        return std.mem.zeroInit(Header, .{
-            .command = .reserved,
-            .op = op,
-            .checksum = 0,
-        });
+        var header: Header = undefined;
+        @memset(std.mem.asBytes(&header), 0);
+        header.command = .reserved;
+        header.op = op;
+        return header;
     }
 
     pub fn dvc_header_type(header: *const Header) enum { blank, valid } {
--- a/src/vsr/replica_format.zig
+++ b/src/vsr/replica_format.zig
@@ -2,10 +2,13 @@ const std = @import("std");
 const assert = std.debug.assert;
 
 const constants = @import("../constants.zig");
+const stdx = @import("../stdx.zig");
 const vsr = @import("../vsr.zig");
 const Header = vsr.Header;
 const format_wal_headers = @import("./journal.zig").format_wal_headers;
 const format_wal_prepares = @import("./journal.zig").format_wal_prepares;
+
+// TODO Parallelize formatting IO.
 
 /// Initialize the TigerBeetle replica's data file.
 pub fn format(
@@ -19,6 +22,9 @@ pub fn format(
     var replica_format = ReplicaFormat{};
 
     try replica_format.format_wal(allocator, options.cluster, storage);
+    assert(!replica_format.formatting);
+
+    try replica_format.format_replies(allocator, storage);
     assert(!replica_format.formatting);
 
     superblock.format(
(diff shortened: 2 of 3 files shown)


src/vsr.zig
```diff
@@ -1656,11 +1656,11 @@ pub const Headers = struct {
     pub const ViewChangeArray = ViewChangeHeadersArray;
 
     fn dvc_blank(op: u64) Header {
-        return std.mem.zeroInit(Header, .{
-            .command = .reserved,
-            .op = op,
-            .checksum = 0,
-        });
+        var header: Header = undefined;
+        @memset(std.mem.asBytes(&header), 0);
```
matklad, line 1660: What’s the difference between `@memset` and `std.mem.zero` (note: zero, *not* zeroInit)?

if the latter gives us extra compile-time checks that zeroing makes sense, that might be preferable.


Although I am hesitant about relying on std functions with subtle and shifting semantics.

memset certainly is simpler!

sentientwaffle: Truthfully I forgot that `std.mem.zeroes` exists when I was writing this code. :wink: 

It looks like `std.mem.zeroes` would do what we want in this case. That said, considering this change is due to being bitten by a quiet stdlib change, I think being explicit is best.

src/vsr/replica_format.zig
```diff
@@ -128,8 +132,32 @@ fn ReplicaFormatType(comptime Storage: type) type {
...
+            storage: *Storage,
+        ) !void {
+            // Direct I/O requires the buffer to be sector-aligned.
+            var message_buffer =
+                try allocator.alignedAlloc(u8, constants.sector_size, constants.message_size_max);
+            defer allocator.free(message_buffer);
+            @memset(message_buffer, 0);
+
+            for (0..constants.clients_max) |slot| {
+                storage.write_sectors(
```
matklad, line 153: `assert(!self.formatting)`

src/vsr/replica_format.zig
```diff
@@ -128,8 +132,32 @@ fn ReplicaFormatType(comptime Storage: type) type {
...
+                    write_sectors_callback,
+                    &self.write,
+                    message_buffer,
+                    .client_replies,
+                    slot * constants.message_size_max,
+                );
+                self.formatting = true;
+                while (self.formatting) storage.tick();
```
matklad, line 161: unrelated to PR, but should we parallelize formatting? As in, I think we could perhaps issue more than one write at a time? 

We trust that parallel IO is safe during normal operation, so I don't think we get any extra reliability here by formatting sequentially? I guess this also ties into the idea about randomizing formatting: we can create a list of all formatting operations we want to do, and then execute them in parallel, in random order. 

formatting is going to be a bottleneck for at least some workflows, so making it faster might be meaningful.

matklad: (implementation wise, I guess this means changing `formatting: bool` to a counter and just issuing several ops?)

src/vsr/replica_format.zig
matklad: And I guess the same for `format_wall`. We assert !formating on exit, but, if we also do this on entry, we get nice interlocked asserts

src/vsr/replica_format.zig
sentientwaffle: :+1: Added a TODO.

matklad (approved):

PR #1213, opened 2023-10-03
IO: Allow running on tmpfs without Direct IO, regardless of direct_io_required

cb22:
Running on `tmpfs` with production settings and a production binary is useful to get system measurement numbers without disk speed playing a role.

This tries to strike a balance between allowing that flexibility and preventing footguns. If `direct_io_required` is true, the code will panic as usual _except_ on a `tmpfs` file system. `tmpfs` is inherently non-durable, so we give an additional warning about this.

### DIFF
--- a/src/io/linux.zig
+++ b/src/io/linux.zig
@@ -9,9 +9,10 @@ const log = std.log.scoped(.io);
 const tracer = @import("../tracer.zig");
 
 const constants = @import("../constants.zig");
+const stdx = @import("../stdx.zig");
 const FIFO = @import("../fifo.zig").FIFO;
 const buffer_limit = @import("../io.zig").buffer_limit;
-const parse_dirty_semver = @import("../stdx.zig").parse_dirty_semver;
+const parse_dirty_semver = stdx.parse_dirty_semver;
 
 pub const IO = struct {
     ring: IO_Uring,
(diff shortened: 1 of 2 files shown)


src/stdx.zig
```diff
@@ -487,3 +487,44 @@ test "stdx.zig: parse_dirty_semver" {
...
+            @as(usize, @bitCast(@as(isize, fd))),
+            @intFromPtr(statfs_buf),
+        );
+    } else {
+        return std.os.linux.syscall2(
+            .fstatfs,
+            @as(usize, @bitCast(@as(isize, fd))),
+            @intFromPtr(statfs_buf),
+        );
+    }
```
batiati:
```suggestion
    return std.os.linux.syscall2(
        if (@hasField(std.os.linux.SYS, "fstatfs64")) .fstatfs64 else .fstatfs,
        @as(usize, @bitCast(@as(isize, fd))),
        @intFromPtr(statfs_buf),
    );
```

src/io/linux.zig
```diff
@@ -1002,12 +1003,20 @@ pub const IO = struct {
...
             if (direct_io_supported) {
                 flags |= os.O.DIRECT;
             } else if (!constants.direct_io_required) {
                 log.warn("file system does not support Direct I/O", .{});
+            } else if (constants.direct_io_required and dir_on_tmpfs) {
+                // Special case. Normally we would panic here (see below), but being able to
+                // benchmark production workloads on tmpfs is very useful for removing disk speed
+                // from the equation.
+                log.warn("tmpfs is not durable, and your data will be lost on reboot", .{});
+                log.warn("tmpfs does not support Direct I/O - continuing", .{});
```
batiati: This warning of running on tmpfs seems critical. I think we should move the condition outside `if (constants.direct_io)`, therefore not bothering to check `fs_supports_direct_io` in the case of tmpfs. 

That's to prioritize showing the warning when running with `constants.direct_io_required == false`, or even in the eventuality the kernel [changes mind](https://lore.kernel.org/lkml/45A29EC2.8020502@tmr.com/t/) and decides that tmpfs should support `O_DIRECT`.

batiati (commented):
It would be great to trick `benchmark.sh` to have an option to create the tmpfs partition 🚀.

cb22:
> It would be great to trick `benchmark.sh` to have an option to create the tmpfs partition 🚀.

Definitely - it's part of what I want to do when we integrate it fully as a subcommand (`tigerbeetle benchmark ...`).

src/stdx.zig
cb22: Ah yeah - I was just copying the Zig stdlib style, but it's a simple change so no need to keep it consistent with that internally. (should put up a Zig PR :smile:)

src/io/linux.zig
cb22: Good point! I've made the warning about running on `tmpfs` always show, and we assume Direct I/O isn't supported.

batiati (approved):

PR #1217, opened 2023-10-06
Use integer keys

batiati:
### This PR

Makes `Table.Key` always an integer type, removing the need for the `compare_keys` function.

**Rationale:** 

Removes unnecessary extra comparisons performed by `std.math.order` when we need to check only one condition, e.g. `a > b` instead of `std.math.order(a, b) == .gt`.

Future benefits may be explored when dealing with integer keys.

**Results:**

- Binary size decreased from 11757K to 10934K. Removing the inline function resulted in a **823K** smaller binary.

- Slightly better performance for both sequential and random IDs:

<table>
<th>
<td>Before</td><td>After</td>
</th>
<tr>
<td>Sequential IDs</td>
<td>
<pre>
Benchmarking...
info: Benchmark running against { 127.0.0.1:44605 }
1226 batches in 59.76 s
load offered = 1000000 tx/s
load accepted = 167324 tx/s
batch latency p00 = 0 ms
batch latency p10 = 13 ms
batch latency p20 = 14 ms
batch latency p30 = 19 ms
batch latency p40 = 19 ms
batch latency p50 = 20 ms
batch latency p60 = 23 ms
batch latency p70 = 23 ms
batch latency p80 = 24 ms
batch latency p90 = 25 ms
batch latency p100 = 2494 ms
transfer latency p00 = 0 ms
transfer latency p10 = 3242 ms
transfer latency p20 = 6370 ms
transfer latency p30 = 10733 ms
transfer latency p40 = 15543 ms
transfer latency p50 = 20723 ms
transfer latency p60 = 26259 ms
transfer latency p70 = 31955 ms
transfer latency p80 = 38095 ms
transfer latency p90 = 44121 ms
transfer latency p100 = 49773 ms
</pre>
</td>
<td>
<pre>
Benchmarking...
info: Benchmark running against { 127.0.0.1:47107 }
1226 batches in 58.06 s
load offered = 1000000 tx/s
load accepted = 172247 tx/s
batch latency p00 = 0 ms
batch latency p10 = 12 ms
batch latency p20 = 13 ms
batch latency p30 = 15 ms
batch latency p40 = 19 ms
batch latency p50 = 20 ms
batch latency p60 = 22 ms
batch latency p70 = 23 ms
batch latency p80 = 24 ms
batch latency p90 = 25 ms
batch latency p100 = 2684 ms
transfer latency p00 = 0 ms
transfer latency p10 = 2675 ms
transfer latency p20 = 5603 ms
transfer latency p30 = 9744 ms
transfer latency p40 = 14511 ms
transfer latency p50 = 19412 ms
transfer latency p60 = 25255 ms
transfer latency p70 = 30772 ms
transfer latency p80 = 36665 ms
transfer latency p90 = 42643 ms
transfer latency p100 = 48067 ms
</pre>
</td>
</tr>
<tr>
<td>Random IDs</td>
<td>
<pre>
Benchmarking...
info: Benchmark running against { 127.0.0.1:36915 }
1226 batches in 68.28 s
load offered = 1000000 tx/s
load accepted = 146456 tx/s
batch latency p00 = 0 ms
batch latency p10 = 18 ms
batch latency p20 = 19 ms
batch latency p30 = 22 ms
batch latency p40 = 26 ms
batch latency p50 = 28 ms
batch latency p60 = 30 ms
batch latency p70 = 31 ms
batch latency p80 = 32 ms
batch latency p90 = 34 ms
batch latency p100 = 2824 ms
transfer latency p00 = 0 ms
transfer latency p10 = 3396 ms
transfer latency p20 = 6709 ms
transfer latency p30 = 12640 ms
transfer latency p40 = 18089 ms
transfer latency p50 = 24672 ms
transfer latency p60 = 30580 ms
transfer latency p70 = 37002 ms
transfer latency p80 = 44130 ms
transfer latency p90 = 51227 ms
transfer latency p100 = 58288 ms
</pre>
</td>
<td>
<pre>
Benchmarking...
info: Benchmark running against { 127.0.0.1:45447 }
1226 batches in 66.87 s
load offered = 1000000 tx/s
load accepted = 149544 tx/s
batch latency p00 = 0 ms
batch latency p10 = 17 ms
batch latency p20 = 19 ms
batch latency p30 = 23 ms
batch latency p40 = 26 ms
batch latency p50 = 28 ms
batch latency p60 = 29 ms
batch latency p70 = 31 ms
batch latency p80 = 32 ms
batch latency p90 = 34 ms
batch latency p100 = 2506 ms
transfer latency p00 = 0 ms
transfer latency p10 = 3035 ms
transfer latency p20 = 6561 ms
transfer latency p30 = 11519 ms
transfer latency p40 = 16549 ms
transfer latency p50 = 23144 ms
transfer latency p60 = 29478 ms
transfer latency p70 = 35383 ms
transfer latency p80 = 42422 ms
transfer latency p90 = 49794 ms
transfer latency p100 = 56878 ms
</pre>
</td>
</tr>
</table>

### DIFF
--- a/src/lsm/binary_search.zig
+++ b/src/lsm/binary_search.zig
@@ -12,26 +12,28 @@ pub const Config = struct {
 };
 
 // TODO The Zig self hosted compiler will implement inlining itself before passing the IR to llvm,
-// which should eliminate the current poor codegen of key_from_value/compare_keys.
+// which should eliminate the current poor codegen of key_from_value.
 
 /// Returns either the index of the value equal to `key`,
 /// or if there is no such value then the index where `key` would be inserted.
 ///
 /// In other words, return `i` such that both:
-/// * key_from_value(values[i])  >= key or i == values.len
-/// * key_value_from(values[i-1]) < key or i == 0
+/// * key <= key_from_value(values[i]) or i == values.len
+/// * key_value_from(values[i-1]) <= key or i == 0
 ///
-/// Expects `values` to be sorted by key.
 /// If `values` contains duplicated matches, then returns
 /// the first index when `Config.mode == .lower_bound`,
 /// or the last index when `Config.mode == .upper_bound`.
+/// This invariant can be expressed as:
+/// * key_value_from(values[i-1]) < key or i == 0 when Config.mode == .lower_bound.
+/// * key < key_value_from(values[i+1]) or i == values.len when Config.mode == .upper_bound.
 ///
+/// Expects `values` to be sorted by key.
 /// Doesn't perform the extra key comparison to determine if the match is exact.
 pub fn binary_search_values_upsert_index(
     comptime Key: type,
     comptime Value: type,
     comptime key_from_value: fn (*const Value) callconv(.Inline) Key,
-    comptime compare_keys: fn (Key, Key) callconv(.Inline) math.Order,
     values: []const Value,
     key: Key,
     comptime config: Config,
@@ -42,10 +44,15 @@ pub fn binary_search_values_upsert_index(
     var length: usize = values.len;
     while (length > 1) {
         if (constants.verify) {
-            assert(offset == 0 or
-                compare_keys(key_from_value(&values[offset - 1]), key) != .gt);
-            assert(offset + length == values.len or
-                compare_keys(key_from_value(&values[offset + length]), key) != .lt);
+            assert(offset == 0 or switch (comptime config.mode) {
+                .lower_bound => key_from_value(&values[offset - 1]) < key,
+                .upper_bound => key_from_value(&values[offset - 1]) <= key,
+            });
+
+            assert(offset + length == values.len or switch (comptime config.mode) {
+                .lower_bound => key <= key_from_value(&values[offset + length]),
+                .upper_bound => key < key_from_value(&values[offset + length]),
+            });
         }
 
         const half = length / 2;
@@ -69,7 +76,7 @@ pub fn binary_search_values_upsert_index(
 
             // @sizeOf(Value) can be greater than a single cache line.
             // In that case, we need to prefetch multiple cache lines for a single value:
-            comptime stdx.maybe(constants.cache_line_size >= @sizeOf(Value));
+            comptime stdx.maybe(@sizeOf(Value) > constants.cache_line_size);
             const CacheLineBytes = [*]const [constants.cache_line_size]u8;
             const cache_lines_per_value = comptime stdx.div_ceil(
                 @sizeOf(Value),
@@ -98,8 +105,8 @@ pub fn binary_search_values_upsert_index(
             // For exact matches, takes the first half if `mode == .lower_bound`,
             // or the second half if `mode == .upper_bound`.
             @intFromBool(switch (comptime config.mode) {
-                .lower_bound => compare_keys(key_from_value(&values[mid]), key) == .lt,
-                .upper_bound => compare_keys(key_from_value(&values[mid]), key) != .gt,
+                .lower_bound => key_from_value(&values[mid]) < key,
+                .upper_bound => key_from_value(&values[mid]) <= key,
             })
         ];
 
@@ -108,25 +115,31 @@ pub fn binary_search_values_upsert_index(
 
     if (constants.verify) {
         assert(length == 1);
-        assert(offset == 0 or
-            compare_keys(key_from_value(&values[offset - 1]), key) != .gt);
-        assert(offset + length == values.len or
-            compare_keys(key_from_value(&values[offset + length]), key) != .lt);
+
+        assert(offset == 0 or switch (comptime config.mode) {
+            .lower_bound => key_from_value(&values[offset - 1]) < key,
+            .upper_bound => key_from_value(&values[offset - 1]) <= key,
+        });
+
+        assert(offset + length == values.len or switch (comptime config.mode) {
+            .lower_bound => key <= key_from_value(&values[offset + length]),
+            .upper_bound => key < key_from_value(&values[offset + length]),
+        });
     }
 
-    offset += @intFromBool(compare_keys(key_from_value(&values[offset]), key) == .lt);
+    offset += @intFromBool(key_from_value(&values[offset]) < key);
 
     if (constants.verify) {
         assert(offset == 0 or switch (config.mode) {
-            .lower_bound => compare_keys(key_from_value(&values[offset - 1]), key) == .lt,
-            .upper_bound => compare_keys(key_from_value(&values[offset - 1]), key) != .gt,
+            .lower_bound => key_from_value(&values[offset - 1]) < key,
+            .upper_bound => key_from_value(&values[offset - 1]) <= key,
         });
         assert(offset >= values.len - 1 or switch (config.mode) {
-            .lower_bound => compare_keys(key_from_value(&values[offset + 1]), key) != .lt,
-            .upper_bound => compare_keys(key_from_value(&values[offset + 1]), key) == .gt,
+            .lower_bound => key <= key_from_value(&values[offset + 1]),
+            .upper_bound => key < key_from_value(&values[offset + 1]),
         });
         assert(offset == values.len or
-            compare_keys(key_from_value(&values[offset]), key) != .lt);
+            key <= key_from_value(&values[offset]));
     }
 
     return @as(u32, @intCast(offset));
@@ -134,7 +147,6 @@ pub fn binary_search_values_upsert_index(
 
 pub inline fn binary_search_keys_upsert_index(
     comptime Key: type,
-    comptime compare_keys: fn (Key, Key) callconv(.Inline) math.Order,
     keys: []const Key,
     key: Key,
     comptime config: Config,
@@ -147,7 +159,6 @@ pub inline fn binary_search_keys_upsert_index(
                 return k.*;
             }
         }.key_from_key,
-        compare_keys,
         keys,
         key,
         config,
@@ -163,7 +174,6 @@ pub inline fn binary_search_values(
     comptime Key: type,
     comptime Value: type,
     comptime key_from_value: fn (*const Value) callconv(.Inline) Key,
-    comptime compare_keys: fn (Key, Key) callconv(.Inline) math.Order,
     values: []const Value,
     key: Key,
     comptime config: Config,
@@ -172,28 +182,26 @@ pub inline fn binary_search_values(
         Key,
         Value,
         key_from_value,
-        compare_keys,
         values,
         key,
         config,
     );
     return .{
         .index = index,
-        .exact = index < values.len and compare_keys(key_from_value(&values[index]), key) == .eq,
+        .exact = index < values.len and key_from_value(&values[index]) == key,
     };
 }
 
 pub inline fn binary_search_keys(
     comptime Key: type,
-    comptime compare_keys: fn (Key, Key) callconv(.Inline) math.Order,
     keys: []const Key,
     key: Key,
     comptime config: Config,
 ) BinarySearchResult {
-    const index = binary_search_keys_upsert_index(Key, compare_keys, keys, key, config);
+    const index = binary_search_keys_upsert_index(Key, keys, key, config);
     return .{
         .index = index,
-        .exact = index < keys.len and compare_keys(keys[index], key) == .eq,
+        .exact = index < keys.len and keys[index] == key,
     };
 }
 
@@ -215,18 +223,16 @@ pub inline fn binary_search_values_range_upsert_indexes(
     comptime Key: type,
     comptime Value: type,
     comptime key_from_value: fn (*const Value) callconv(.Inline) Key,
-    comptime compare_keys: fn (Key, Key) callconv(.Inline) math.Order,
     values: []const Value,
     key_min: Key,
     key_max: Key,
 ) BinarySearchRangeUpsertIndexes {
-    assert(compare_keys(key_min, key_max) != .gt);
+    assert(key_min <= key_max);
 
     const start = binary_search_values_upsert_index(
         Key,
         Value,
         key_from_value,
-        compare_keys,
         values,
         key_min,
         .{ .mode = .lower_bound },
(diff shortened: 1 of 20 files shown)


src/lsm/binary_search.zig
```diff
@@ -43,9 +42,9 @@ pub fn binary_search_values_upsert_index(
     while (length > 1) {
         if (constants.verify) {
             assert(offset == 0 or
-                compare_keys(key_from_value(&values[offset - 1]), key) != .gt);
+                !(key_from_value(&values[offset - 1]) > key));
```
batiati: It's an excellent opportunity to put https://github.com/tigerbeetle/tigerbeetle/pull/1179#discussion_r1322873550 in practice, any thoughts?

| before | after |
--------|--------
`compare_keys(a,b) == .eq` | `a == b`
`compare_keys(a,b) == .lt` | `a < b`
`compare_keys(a,b) == .gt` | `a > b`
`compare_keys(a,b) != .eq` | `a != b`
`compare_keys(a,b) != .lt` | `!(a < b)`
`compare_keys(a,b) != .gt` | `!(a > b)`

In some cases, especially in asserts, `!(a > b)` reads better than `a <= b` imo.

src/lsm/composite_key.zig
```diff
@@ -27,44 +27,43 @@ pub fn CompositeKey(comptime Field: type) type {
...
-            }
-            return order;
-        }
-
-        pub inline fn key_from_value(value: *const Value) Self {
-            return .{
+        pub inline fn key_from_value(value: *const Self) Key {
+            return @bitCast(Self{
                 .field = value.field,
                 .timestamp = @as(u63, @truncate(value.timestamp)),
```
matklad:
```suggestion
                .timestamp = @as(u63, @truncate(value.timestamp)),
                .field = value.field,
```

src/lsm/composite_key.zig
```diff
@@ -27,44 +27,43 @@ pub fn CompositeKey(comptime Field: type) type {
...
 
-        pub inline fn tombstone_from_key(key: Self) Value {
-            return .{
-                .field = key.field,
-                .timestamp = key.timestamp | tombstone_bit,
+        pub inline fn tombstone_from_key(key: Key) Self {
+            const value: Self = @bitCast(key);
+            return Self{
+                .field = value.field,
+                .timestamp = value.timestamp | tombstone_bit,
```
matklad: This makes me wonder if perhaps tombstone could be the least significant bit of the timestamp, rather than the MSB. That way, when we do binary search, we won't have to mask, and instead use the integer as is. This in turn would allow us to parametrize binary search not by Key/Value types, but rather by:

1) size of value
2) offset of the key in value
3) size of the key

The benefit here would be:

* smaller code size
* potentially faster runtime

For the on-disk format, LSB tombstone feels like a win to me. However, I don't like this as an over-the-wire format --- we never allow tombstones on the wire, and an unset LSB would look weird.

OTOH, its always the replica who assigns timestamps, and I think we could always assign even timestamps? Ie, we won't even document that we use LSB for tombstones, it's just so happens that all alive objects get assigned even timestamps.

src/lsm/binary_search.zig
matklad: Heh, using the rationale from that comment, I would say that

```zig
assert(
  key_from_value(&values[offset - 1]) <= key
)
```

is the most readable version:

* it states the invariant positively 
* it follows the natural number-line order of comparison, where the lesser number is on the left

That is, using the same antecedent, I seem to arrive at the opposite conclusion :rofl: 


Either version is much more readable than the one with `.gt` though!

matklad: Also, spelled this way, it's clear that the assertion is actually wrong! We want 

```zig
key_from_value(&values[offset - 1]) < key
```

here, as per invariant from the docs: 

```zig
key_value_from(values[i-1]) < key or i == 0
```

incidentally, I was thinking about these invariants during my PTO:

https://matklad.github.io/2023/10/06/what-is-an-invariant.html

src/lsm/binary_search.zig
```diff
@@ -43,9 +42,9 @@ pub fn binary_search_values_upsert_index(
...
             assert(offset + length == values.len or
-                compare_keys(key_from_value(&values[offset + length]), key) != .lt);
+                !(key_from_value(&values[offset + length]) < key));
```
matklad: This one is correct, but I'd spell this as 

```
key <= key_from_value(&values[offset + length])
```

src/lsm/binary_search.zig
matklad: Ahh, no, I think the docs are incorrect: now that we have lower/upper in config, the invariant actually depends on which bound we are looking for. So I think we need to

* adjust the docs
* use different asserts depenring on `comptime switch (config.mode)` (the current assert works for both cases, but it is not as tight as it could be)

matklad: >In some cases, especially in asserts, !(a > b) reads better than a <= b imo.

Could you give an intuition why `!(a > b)` reads better for you? I just can't see it, to me it's obviously a worse readable version, so I am wondering if I am missing some different way to look at comparisons?

src/lsm/composite_key.zig
```diff
@@ -27,44 +27,43 @@ pub fn CompositeKey(comptime Field: type) type {
...
+        pub const Key = std.meta.Int(.unsigned, @bitSizeOf(Self));
+
+        // The keys are ordered by (field, timestamp),
+        // fields are declared from the least siginificant to the most significant:
 
-        field: Field align(field_bitsize_alignment),
-        /// The most significant bit must be unset as it is used to indicate a tombstone.
-        timestamp: u64,
         /// [0]u8 as zero-sized-type workaround for https://github.com/ziglang/zig/issues/16394.
         padding: @TypeOf(pad) = pad,
```
matklad: padding first feels weird, and I think it actually is wrong, as it breaks alignment. If Field is u128, then, with padding first, it is alligned only as good as u64. Which is actually OK, as the abi for u128 goes, but still feels less than obvious.

src/lsm/forest_table_iterator.zig
```diff
@@ -157,12 +158,14 @@ fn TreeTableIteratorType(comptime Tree: type) type {
...
-                                // (The tables are ordered by (key_max,snapshot_min).)
-                                .snapshot_min = position.previous.snapshot_min + 1,
-                            }),
+                            manifest_level.tables.search(@bitCast(
+                                KeyMaxSnapshotMin{
+                                    .key_max = position.previous.key_max,
+                                    // +1 to skip past the previous table.
+                                    // (The tables are ordered by (key_max,snapshot_min).)
+                                    .snapshot_min = position.previous.snapshot_min + 1,
+                                },
```
matklad: I guess flip the order here as well? I am torn on this, but I think bitcasting data in the wrong order is more confusing than reversed lexicographic comparison

src/lsm/forest_table_iterator.zig
```diff
@@ -178,7 +181,7 @@ fn TreeTableIteratorType(comptime Tree: type) type {
 
             if (iterator.position) |position| {
                 const table_order =
-                    Tree.Table.compare_keys(position.previous.key_max, table.key_max);
+                    std.math.order(position.previous.key_max, table.key_max);
                 assert(table_order.compare(.lte));
```
matklad: Maybe `switch` over `table_order` here, with an explicit unreachable branch for `ge`?

src/lsm/composite_key.zig
```diff
@@ -4,16 +4,16 @@ const math = std.math;
 
 const stdx = @import("../stdx.zig");
 
-pub fn CompositeKey(comptime Field: type) type {
+pub fn CompositeKeyType(comptime Field: type) type {
     assert(Field == u128 or Field == u64);
 
     return extern struct {
```
matklad, line 17: I _think_ this now wants to be a `packed` struct? Packed structs are "integres with extra semantics", and that's what we have here essentially.

matklad (commented):
Awesome!, but I am genuinely puzzled by the `assert(!(a > b))` form, as it seems to me exactly the opposite of readable. Curious what other folks think here!

src/lsm/binary_search.zig
batiati: > Could you give an intuition why !(a > b) reads better for you?

Not a general case, but for example `assert(!(key > last_key))`, my inner voice reads it as "it cannot be greater than the last key", which I think is easier to reason than `assert(key <= last_key)`.
IMO dealing with `>` and `<` is natural, but `>=` and `<=` are more subject to subtle "off-by-one" errors.

However, I'm not strong about it. The main motivation for converting `compare_keys(a,b) != .gt` to `!(a > b)` was to make it easier to mechanically refactor the code. I'd be happier if we could define a canonical way and document it in the TigerStyle.

batiati: Unrelated but relevant:

What is the expected semantics of `maybe(<condition>)`?
Should we state something that is likely to always happen, or should we state the rule's exception?

For example, after `message_bus.unref()` we likely have freed the message, but maybe not, so the condition should be `maybe(messages.references > 0)` or `maybe(messages.references == 0)`?

I'm asking that because we could introduce a new function called `assert_not` for that kind of usage like `assert_not(key > last_key)`.

matklad: >What is the expected semantics of maybe(<condition>)?

The intended semantics is that the condition can be either true or false, and, once we have branch coverage, we should have tests tripping both sides.

I don't think we assign any  probabilities to maybe, so both `maybe(cond)` and `maybe(!cond)` are equivalent. Though, I think we usually state the more surprising thing in the `cond`, as a cognitive help for the reader.

src/lsm/composite_key.zig
batiati: Packed struct simplifies the alignment, and `packet struct(IntType)` communicates the intent nicely.

Also:
- Added asserts for `padding == pad`
- Added asserts for `timestamp & tombstone_bit == 0`
- Padding now can be `void` instead of `[0]u8`

src/lsm/composite_key.zig
```diff
@@ -4,67 +4,73 @@ const math = std.math;
...
+    const Pad = @TypeOf(pad);
+
+    // The composite key's backing integer type.
+    const Int = std.meta.Int(
+        .unsigned,
+        @bitSizeOf(u64) + @bitSizeOf(Field) + @bitSizeOf(Pad),
+    );
 
-    return extern struct {
+    return packed struct(Int) {
```
matklad: Didn't realize that one can manually specify the backing integer type, that's ninfty!

src/lsm/composite_key.zig
```diff
@@ -4,67 +4,73 @@ const math = std.math;
...
+        u64 => {},
+        else => @compileError("invalid Field for CompositeKey: " ++ @typeName(Field)),
+    };
+    const Pad = @TypeOf(pad);
+
+    // The composite key's backing integer type.
+    const Int = std.meta.Int(
+        .unsigned,
+        @bitSizeOf(u64) + @bitSizeOf(Field) + @bitSizeOf(Pad),
+    );
```
matklad: As we only support u128 and u64 fields, maybe this will be more readable as an explicit switch with hard-coded `u128` and `u256` arms? Works either way for me.

src/lsm/composite_key.zig
```diff
@@ -4,67 +4,73 @@ const math = std.math;
...
 
-        field: Field align(field_bitsize_alignment),
         /// The most significant bit must be unset as it is used to indicate a tombstone.
         timestamp: u64,
-        /// [0]u8 as zero-sized-type workaround for https://github.com/ziglang/zig/issues/16394.
-        padding: @TypeOf(pad) = pad,
+
+        field: Field,
+
+        padding: Pad = pad,
```
matklad, line 38: :facepalm: I think I mixed things up when I said that padding first breaks alignment. I think we actually _need_ to have padding first to preserve alignment, _because_ timestamp precedes Field. 

For `Field = u128, timestamp = u64, pad = u64`, the traditional layout would be:

```
Field     // u128
timestamp // u64
Pad       // u64
```

This layout is well-aligned, but doesn't fit little-endian comparison order we need. 

With timestamp first,  we need this layout for alignment to add up:

```
Pad       // u64
timestamp // u64
Field     // u128
```


:thinking: honestly, I think it doesn't really matter either way. 

To confuse us with even more options: I realized that we don't actually need to change composite key layout here. We are doing non-trivial masking operation in `key_from_value` anyway, and that can read like:

```
return @as(Key, value.timestamp & ~tombstone_bin) | (@as(Key, value.field) << 64)
```

This then would be completely independent on the actual layout of data on disk.

I think I am slightly in favor of this latter option -- what makes me slightly uneasy about the current state in the PR, is that our "integer keys" affects the data layout on disk somewhat (as we rearrange the order of bytes), but it is _still_ not just "take these bytes as a key", as we need to mask away the tombstone. 

But we actually don't have to change the layout on disk! `fn key_from_value` allows us to cook up the integer key in whichever way we want. So we preserve both:

* the natural, aligned `field, timestamp, padding` layout on disk
* the integer keys

I don't have a _strong_ argument for why "natural" layout is actually good though, except that it is, well, non-weird :) 

Sorry for back and forth here!

batiati: > return @as(Key, value.timestamp & ~tombstone_bin) | (@as(Key, value.field) << 64)

That's a great solution. We don't particularly care about keeping the data layout on disk as a little-endian integer, in fact, I think it gives more flexibility to the LSM layout.

> Sorry for back and forth here!

Not at all!!!

batiati: I restored the `CompositeKey`'s layout as an `extern struct` to be able to keep the alignment.

However, I kept `KeyMaxSnapshotMin` as a `packed struct(Int)` because we don't store it.
Just reduced the code "weirdness" by not `@bitcast`ing it around.

matklad (dismissed):

batiati:
Sorry @matklad, just added some tests for the CompositeKey logic.

matklad (approved):

PR #1218, opened 2023-10-09
Node:  Reimplement node client using `tb_client`

kprotty:
The node client interacted with `vsr.Client` in its own weird way; TigerBeetle's `IO.run_for_ns()` is the heart of its event loop and is needed by the `vsr.Client`. Unfortunately for language clients, IO blocks the calling thread. To solve this, we wrote `tb_client` that runs IO + vsr.Client in another thread with cross-thread communication. 

The node client, however, was written before that and instead called `IO.tick()` (nonblocking) using a `setInterval` in JS. While correct, it results in requests getting delayed until the next tick (which is currently 10ms) hurting latency and in-turn throughput. This PR rewrites the node client to use `tb_client` instead.


src/clients/c/tb_client_exports.zig
batiati, line 1: It's nice to move the FFI to another file 🥳.

Let's drop the `ffi_` prefix then (except for `ffi_allocator()`)

src/clients/c/tb_client/context.zig
```diff
@@ -178,9 +178,9 @@ pub fn ContextType(
             );
             errdefer context.client.deinit(context.allocator);
 
-            context.on_completion_ctx = on_completion_ctx;
             context.on_completion_fn = on_completion_fn;
             context.implementation = .{
+                .completion_ctx = on_completion_ctx,
```
batiati: We can call it just `completion_ctx`.
Pls check and rename other places too.

src/clients/c/tb_client_exports.zig
```diff
@@ -0,0 +1,76 @@
...
+    return if (builtin.is_test)
+        std.testing.allocator
+    else if (builtin.link_libc)
+        std.heap.c_allocator
+    else if (builtin.target.os.tag == .windows)
+        (struct {
+            var gpa = std.heap.HeapAllocator.init();
+        }).gpa.allocator()
+    else
+        @compileError("tb_client must be built with libc");
```
batiati: That's something I think we can simplify: We always link FFI clients using libc, so even on Windows we use the `c_allocator`.
Should we move Windows up or simply remove this special case?

src/clients/node/src/node.zig
```diff
@@ -15,783 +19,465 @@ const CreateTransfersResult = tb.CreateTransfersResult;
...
+        "Failed to get client context pointer.",
+    );
+    const client: tb.tb_client_t = @ptrCast(@alignCast(client_ptr.?));
+    defer tb.deinit(client);
 
-    // We take a slice on `output` to ensure that its length is a multiple of @sizeOf(T) to prevent
-    // a safety-checked runtime panic from `bytesAsSlice` for non-multiple sizes.
-    var results = std.mem.bytesAsSlice(T, output[0..body_length]);
+    const on_completion_ctx = tb.completion_context(client);
+    const on_completion_tsfn: c.napi_threadsafe_function = @ptrFromInt(on_completion_ctx);
```
batiati: Removing this alias makes clear that `completion_ctx` is the pointer to a Node's thread-safe function.

```suggestion
    const on_completion_tsfn: c.napi_threadsafe_function = @ptrFromInt(
        tb.completion_context(client),
    );
```

src/clients/node/src/node.zig
```diff
@@ -15,783 +19,465 @@ const CreateTransfersResult = tb.CreateTransfersResult;
...
+        context,
+        "Failed to get client context pointer.",
     );
+    const client: tb.tb_client_t = @ptrCast(@alignCast(client_ptr.?));
+
+    const packet = blk: {
+        var packet_ptr: ?*tb.tb_packet_t = undefined;
+        switch (tb.acquire_packet(client, &packet_ptr)) {
+            .ok => break :blk packet_ptr.?,
+            .shutdown => return translate.throw(env, "Client was prematurely shutdown."),
```
batiati: Not sure if "_prematurely_ shutdown" is the correct message here.
`.shutdown` is returned if one thread starts `tb_deinit()` and another tries to acquire a packet.
I think that just `Client was shutdown.` is more adequate.

src/clients/node/src/node.zig
```diff
@@ -15,783 +19,465 @@ const CreateTransfersResult = tb.CreateTransfersResult;
...
+        .too_much_data => unreachable, // We limit packet data size during request().
+        .invalid_operation => unreachable, // We check the operation during request().
+        .invalid_data_size => unreachable, // We set correct data size during request().
     }
 
-    if (argc != 4) translate.throw(
-        env,
-        "Function request() requires 4 arguments exactly.",
-    ) catch return null;
+    // Copy the results given to use into the packet dataa.
```
batiati:
```suggestion
    // Copy the results given to use into the packet data.
```

src/clients/node/src/node.zig
```diff
@@ -15,783 +19,465 @@ const CreateTransfersResult = tb.CreateTransfersResult;
...
-    const context_raw = translate.value_external(
-        env,
-        argv[0],
-        "Failed to get Client Context pointer.",
-    ) catch return null;
-    const context = contextCast(context_raw.?) catch return null;
-    const operation_int = translate.u32_from_value(env, argv[1], "operation") catch return null;
+    // Stuff the event size along with the packet result size into the packet.
+    // The former is needed to know the full heap allocation size in packet.data.
+    // The latter is needed to know how many Results in packet.data to serialize for JS.
```
batiati: Neat solution using the same buffer for both!

Are "former" and "latter" precise here?
For example, for `create_accounts` the data size should be used to know the full heap allocation size, however for `lookup_accounts` it's the opposite, since the result is much larger than the event.

src/clients/node/src/node.zig
```diff
@@ -15,783 +19,465 @@ const CreateTransfersResult = tb.CreateTransfersResult;
...
+        .invalid_operation => unreachable, // We check the operation during request().
+        .invalid_data_size => unreachable, // We set correct data size during request().
     }
 
-    if (argc != 4) translate.throw(
-        env,
-        "Function request() requires 4 arguments exactly.",
-    ) catch return null;
+    // Copy the results given to use into the packet dataa.
+    const events = @as([*]align(16) u8, @ptrCast(@alignCast(packet.data.?)))[0..packet.data_size];
```
batiati:
```suggestion
    const events: [*]align(16) u8 = @ptrCast(@alignCast(packet.data.?))[0..packet.data_size];
```

src/clients/node/src/node.zig
```diff
@@ -15,783 +19,465 @@ const CreateTransfersResult = tb.CreateTransfersResult;
...
-    }
-    const operation = @as(Operation, @enumFromInt(@as(u8, @intCast(operation_int))));
+// (De)Serialization
 
-    if (context.client.messages_available == 0) {
-        translate.throw(
-            env,
-            "Too many outstanding requests - message pool exhausted.",
-        ) catch return null;
+fn decode_array_into_data(
```
batiati: Naming things is hard ...
Instead of `decode_array_into_data`, we could use a name that express two important things the reader should know about:
a) It's allocating memory
b) The buffer returned will be used both for sending (data) and receiving (results).

Same thing with `encode_results_into_data` and `encode_data_into_array`

batiati (changes requested):
Amazing work @kprotty

src/clients/node/src/node.zig
kprotty: For a given op, packet memory is allocated to hold `max(sizeof(Event/Request) * n, sizeof(Result/Response))`. As long as the op and `n` are known, the allocation size can be computed.

When submitting the packet, the data_size is `sizeof(Event) * n`. On packet result, we overwrite the memory with `sizeof(Result) * completed` where completed <= n. If less Results are completed than Events, the completed size no longer reflects "n". So instead, data_size stores `packed(n, completed)`: arg1 to know alloc size, arg2 to know what to parse back to JS. 

Admittedly got a bit too "clever" here to avoid allocations.

src/clients/node/src/node.zig
kprotty: Thoughts on the following?
* `alloc_packet_memory_from_js()`
* `copy_results_into_packet_memory()`
* `free_packet_memory_into_js()`

src/clients/c/tb_client_exports.zig
kprotty: Will try just removing it.

src/clients/node/src/node.zig
batiati: > Admittedly got a bit too "clever" here to avoid allocations.

😎

You should document it here, or at `DataSize`.
Btw, what about renaming `DataSize` to `BufferSize` or something else communicating that?

src/clients/node/src/node.zig
batiati: That reads great!

Packet is a name that we're thinking to change, because it's so "network-ish" ... but let's tackle that on #1226

jorangreef: Chiming in a little later here, but let's keep the naming as is.

The context is Node's N-API, where we are allocating for JS as needed, so I think that's fine.

What's important/intended for these methods is that they highlight the JS type: i.e. `array` and the encode/decode direction (without boilerplate).

```
decode_array_into_data
encode_data_into_array
```

They were also chosen carefully to line up symmetrically in the source.

src/clients/node/src/node.zig
kprotty: `packet.data` is cast into a many-pointer, then sliced on to actually make `const events: []align(16) u8`

src/clients/c/tb_client/context.zig
```diff
@@ -97,7 +98,6 @@ pub fn ContextType(
         message_pool: MessagePool,
         client: Client,
 
-        on_completion_ctx: usize,
         on_completion_fn: tb_completion_t,
```
batiati: Remove the prefix on `on_completion_fn`

src/clients/node/src/node.zig
```diff
@@ -15,783 +19,472 @@ const CreateTransfersResult = tb.CreateTransfersResult;
...
+        error.OutOfMemory => return translate.throw(env, "Batch allocation ran out of memory."),
     };
-}
-
-fn tick(env: c.napi_env, info: c.napi_callback_info) callconv(.C) c.napi_value {
-    var argc: usize = 1;
-    var argv: [1]c.napi_value = undefined;
-    if (c.napi_get_cb_info(env, info, &argc, &argv, null, null) != c.napi_ok) {
-        translate.throw(env, "Failed to get args.") catch return null;
+    errdefer allocator.free(data);
```
batiati: Suggestion: rename `data` to `buffer` and use a block to keep the related code/comments together.
I think `data` is ambiguous with `events`.

```zig
    const buffer: []align(16) u8 = buffer: {
        // Allocate enough memory to contain the Events and the Results.
        const alloc_size = @max(@sizeOf(Event) * array_length, @sizeOf(Result) * array_length);
        if (@sizeOf(vsr.Header) + alloc_size > constants.message_size_max) {
            return translate.throw(env, "Batch is larger than the maximum message size.");
        }

        const alloc_align = @max(@alignOf(Event), @alignOf(Result));
        comptime assert(alloc_align <= 16); // Assume 16-byte alignment. Makes free() simpler.

        break :buffer allocator.alignedAlloc(u8, 16, alloc_size) catch |err| switch (err) {
            error.OutOfMemory => return translate.throw(env, "Batch allocation ran out of memory."),
        };
    };
    errdefer allocator.free(buffer);
```

src/clients/node/src/node.zig
```diff
@@ -15,783 +19,472 @@ const CreateTransfersResult = tb.CreateTransfersResult;
...
+    const event_count = @divExact(event_data.len, @sizeOf(Event));
+    const results = std.mem.bytesAsSlice(Result, event_data.ptr[0 .. @sizeOf(Result) * event_count]);
+
+    const packet_results = std.mem.bytesAsSlice(Result, result_bytes);
+    assert(packet_results.len <= results.len);
+    @memcpy(results[0..packet_results.len], packet_results);
+
+    return .{
+        .event_count = @intCast(event_count),
+        .result_count = @intCast(packet_results.len),
```
batiati: Suggestions:

- Make names consistent `results_bytes` and `events_bytes` instead of `events_data`.
- Don't introduce new names like `packet_results`, instead let's just use `buffer` and naturally call it `results`.
- Explain why we're doing it in the comments.

```suggestion
fn encode_results_into_data(
    comptime Event: type,
    comptime Result: type,
    result_bytes: []const u8,
    event_bytes: []align(16) u8,
) BufferSize {
    const event_count = @divExact(event_bytes.len, @sizeOf(Event));
    const buffer: []Result = buffer: {
        // The results must be stored until the JS callback is called, we use the same buffer for
        // both events and results.
        // Reinterpreting `event_bytes` is safe since the buffer should have been allocated with
        // enough space to also contain the Results.
        const buffer_size = @sizeOf(Result) * event_count;
        break :buffer std.mem.bytesAsSlice(Result, event_bytes.ptr[0..buffer_size]);
    };

    const results = std.mem.bytesAsSlice(Result, result_bytes);
    assert(results.len <= buffer.len);
    @memcpy(buffer[0..results.len], results);

    return .{
        .event_count = @intCast(event_count),
        .result_count = @intCast(results.len),
```

src/clients/node/src/node.zig
```diff
@@ -15,783 +19,472 @@ const CreateTransfersResult = tb.CreateTransfersResult;
...
-    const context = contextCast(context_raw.?) catch return null;
+// For a given op, packet memory is allocated to hold
+// `max(sizeof(Event/Request) * n, sizeof(Result/Response))`. As long as the op and n are known, the
+// allocation size can be computed.
+//
+// When submitting the packet, the data_size is `sizeof(Event) * n`. On packet results, we overwrite
+// the memory with `sizeof(Result) * completed` where completed <= n. If less Results are completed
+// than Events, the completed size no longer reflects "n". So instead, data_size stores
+// `packed(n, completed)`: arg1 to know alloc size, arg2 to know what to parse back to JS.
+const DataSize = packed struct(u32) {
```
batiati: Let's call it `BufferSize`, "Data" can be misleading since it holds events and results.

```suggestion
const BufferSize = packed struct(u32) {
```

src/clients/node/src/node.zig
batiati: I agree, we must keep `encode` and `decode` :+1: 

That's **one suggestion**:

`decode_array_into_data` to `decode_from_array` and `encode_data_into_array` to `encode_into_array`.
Rationale: `data` isn't a thing anymore, we are using a dual-purpose buffer that can be used for both `events` and `results` (results need to be stored until the JS callback is executed).

`encode_results_into_data` to `copy_results_to_buffer`
Rationale: this function does not encode data. It only reinterprets the `events` memory to retrieve the buffer to copy the `results`.

**Another suggestion** would be making `Buffer` an explicit type that holds memory and has all the functions needed to reinterpret and free memory.

Example:

```zig
// On request:
var buffer: Buffer = decode_array(js_obj);
const events: []Event = buffer.events();

//On completion:
var buffer = Buffer.from_ptr(events_bytes);
var results: []Result = buffer.results();

// On JS callback
var buffer = Buffer.from_ptr(results_bytes);
defer buffer.free();

var results: []Result = buffer.results();
encode_array(results[0..result_count]);
```

batiati (changes requested):
@kprotty can you please check the failing CI?

src/clients/node/src/node.zig
```diff
@@ -15,783 +19,470 @@ const CreateTransfersResult = tb.CreateTransfersResult;
...
+                .count = event_count,
+            };
+
+            const Result = StateMachine.Result(op);
+            const completed: []const Result = @alignCast(std.mem.bytesAsSlice(Result, results));
+            @memcpy(buffer.results()[0..completed.len], completed);
+
+            packet.data_size = @bitCast(BufferSize{
+                .allocated = @intCast(event_count),
+                .completed = @intCast(completed.len),
```
batiati: I feel like the previous names `events_count` and `results_count` are more appropriate than `allocated` and `completed`.
Especially because the actual `allocated` amount needs to be calculated based on `events_count`.

src/clients/node/src/node.zig
```diff
@@ -15,783 +19,470 @@ const CreateTransfersResult = tb.CreateTransfersResult;
...
+            });
+        },
```
batiati: Suggestion: no need for taking an alias for `results` and avoid introducing new names such as `completed`;

```suggestion
    switch (@as(Operation, @enumFromInt(packet.operation))) {
        inline else => |op| {
            const event_count = @divExact(packet.data_size, @sizeOf(StateMachine.Event(op)));
            const buffer: BufferType(op) = .{
                .ptr = @ptrCast(packet.data.?),
                .count = event_count,
            };

            const Result = StateMachine.Result(op);
            const results: []const Result = @alignCast(
                std.mem.bytesAsSlice(Result, result_ptr.?[0..result_len]),
            );
            @memcpy(buffer.results()[0..results.len], results);

            packet.data_size = @bitCast(BufferSize{
                .allocated = @intCast(event_count),
                .completed = @intCast(results.len),
            });
        },
```

src/clients/node/src/node.zig
```diff
@@ -15,783 +19,470 @@ const CreateTransfersResult = tb.CreateTransfersResult;
...
+    packet_argument: ?*anyopaque,
+) callconv(.C) void {
+    _ = unused_js_cb;
+    _ = unused_context;
+
+    const packet: *tb.tb_packet_t = @ptrCast(@alignCast(packet_argument.?));
+    assert(packet.status == .ok);
+
+    // Decode the packet's Buffer results into an array then free the Buffer.
+    const buffer_size: BufferSize = @bitCast(packet.data_size);
```
batiati: Reduce the variable scope by moving `buffer_size` to the switch block.

batiati (commented):
Just a few adjustments and we're good to merge!
Thanks @kprotty!

batiati (approved):

PR #1219, opened 2023-10-10
vsr: fix integer overflow in address parsing code

matklad:
Also add simplistic fuzzer while we are at it. Parsing _really_ wants coverage guided fuzzing, but even this test is enough to uncover the bug.

Just something I've noticed while working through backlog of emails!

### DIFF
--- a/src/vsr.zig
+++ b/src/vsr.zig
@@ -1412,6 +1412,7 @@ pub fn parse_address_and_port(string: []const u8) !std.net.Address {
 }
 
 fn parse_address(string: []const u8, port: u16) !std.net.Address {
+    if (string.len == 0) return error.AddressInvalid;
     if (string[string.len - 1] == ':') return error.AddressHasMoreThanOneColon;
 
     if (string[0] == '[' and string[string.len - 1] == ']') {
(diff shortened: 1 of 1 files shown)


src/vsr.zig
```diff
@@ -1525,6 +1529,30 @@ test "parse_addresses" {
...
+    var input_max: [len_max]u8 = .{0} ** len_max;
+    for (0..test_count) |_| {
+        const len = random.uintAtMost(usize, len_max);
+        var input = input_max[0..len];
+        for (input) |*c| {
+            c.* = alphabet[random.uintAtMost(usize, alphabet.len)];
+        }
+        if (parse_addresses(std.testing.allocator, input, 3)) |addresses| {
+            std.testing.allocator.free(addresses);
+        } else |_| {
```
sentientwaffle: Does zig require this branch since it is an error?

sentientwaffle (dismissed):

src/vsr.zig
matklad: Yup, otherwise Zig assumes it's an option. I _think_ the idea here is to make error discards syntactically apparent --- `catch {}` and `else |_|` are the only forms that can discard errors (and they only discard errors). Something like `_ = err` can'd discard error. This is in contrast  to Rust, which doesn't distinguish between discarding errors and discarding any other value.

matklad: but `zig fmt` disagrees with formatting here

src/vsr.zig
```diff
@@ -1525,6 +1529,29 @@ test "parse_addresses" {
     }
 }
 
+test "parse_addresses: fuzz" {
+    const test_count = 1024;
+    const len_max = 32;
+    const alphabet = " \t\n,:[]0123456789abcdefgABCDEFGXx";
+
+    const seed = std.crypto.random.int(u64);
```
sentientwaffle, line 1537: Only other note is whether or not we should use a CPRNG in unit tests -- I think all of our existing unit tests are deterministic, and then we have separate fuzzers to test with real randomness.

But a separate fuzzer feels like overkill in this case, so I'm happy with it as-is.

sentientwaffle (approved):

src/vsr.zig
matklad: This is copy-pasted from `test "k_way_merge: fuzz"`, so there's a precedent (I didn't add `errdefer` line, as we only test for crashes here, and panics don't call defers).

In general, I am pretty strongly in favor of using real random seed in these kinds of tests --- if the test is only flaky when the code is incorrect, I'd rather have such a test. 

Though, what gives me a pause here is that `zig build test` actually caches the result of the test, defeating the purpose of CPRNG here a bit.

PR #1221, opened 2023-10-12
dotnet: fix flaky tests

matklad:


src/clients/dotnet/TigerBeetle.Tests/IntegrationTests.cs
```diff
@@ -1130,8 +1132,34 @@ public TBServer()
...
+                var found = line.IndexOf(listening);
+                if (found != -1)
+                {
+                    if (addressLogged != "") throw new InvalidOperationException("have already read the port");
+                    addressLogged = line.Substring(found + listening.Length).Trim();
+                    readPortEvent.Set();
+                }
+            };
+            process.BeginErrorReadLine();
+            readPortEvent.WaitOne(60000);
```
batiati: I think we should call `process.CancelErrorRead()` to stop listening to the stderr after initialization, just in case!

```suggestion
            process.BeginErrorReadLine();
            readPortEvent.WaitOne(60_000);
            process.CancelErrorRead();
```

src/clients/dotnet/TigerBeetle.Tests/IntegrationTests.cs
```diff
@@ -1109,19 +1111,19 @@ private static void AssertTransfer(Transfer a, Transfer b)
...
         private const string TB_EXE = "tigerbeetle";
         private const string TB_FILE = "dotnet-tests.tigerbeetle";
         private const string TB_SERVER = TB_PATH + "/" + TB_EXE;
         private const string FORMAT = $"format --cluster=0 --replica=0 --replica-count=1 ./" + TB_FILE;
-        private const string START = $"start --addresses=" + TB_PORT + " --cache-grid=256MB ./" + TB_FILE;
+        private const string START = $"start --addresses=0  --cache-grid=256MB ./" + TB_FILE;
 
         private readonly Process process;
 
+        public readonly string address;
```
batiati:
```suggestion
        public string Address { get; }
```

Idiomatic way of publishing a read-only field is via property 🤷‍♂️, just replace `server.address` with `server.Address`.

src/clients/dotnet/TigerBeetle.Tests/IntegrationTests.cs
```diff
@@ -1130,8 +1132,34 @@ public TBServer()
...
+            process.StartInfo.Arguments = START;
+            process.StartInfo.RedirectStandardError = true;
+            process.Start();
             if (process.WaitForExit(100)) throw new InvalidOperationException("Tigerbeetle server failed to start");
+
+            var readPortEvent = new AutoResetEvent(false);
+            var addressLogged = "";
+            var stderrLogged = new StringBuilder();
+            process.ErrorDataReceived += (sender, e) => {
+                var line = e.Data;
```
batiati: Running `dotnet format`:

```suggestion
            process.ErrorDataReceived += (sender, e) =>
            {
                var line = e.Data;

```

We might add `dotnet format --verify-no-changes` to our CI.

batiati (changes requested):
It's great to have dynamic port support!
I'll also add the same logic to the other clients in a separate PR.

src/clients/dotnet/TigerBeetle.Tests/IntegrationTests.cs
matklad: Good call, will add to https://github.com/tigerbeetle/tigerbeetle/pull/1201

src/clients/dotnet/TigerBeetle.Tests/IntegrationTests.cs
```diff
@@ -1141,18 +1173,6 @@ public void Dispose()
...
-            {
-                if (Process.GetProcessesByName(TB_EXE) is Process[] runningList)
-                {
-                    foreach (var runningProcess in runningList)
-                    {
-                        runningProcess.Kill();
-                    }
-                }
-            }
-            catch { }
```
batiati, line 1154: Humm... I remembered why this code is there: In the earlier days of the dotnet client it was common to the client to segfault leaving the server running... so I put it as a band-aid to avoid manually killing the process each time but forgot to remove 🙈

batiati (dismissed):

src/clients/dotnet/TigerBeetle.Tests/IntegrationTests.cs
matklad: I am wondering what _should_ cleanup stray processes? My guess is that actions are run as a separate process group, with a shell attached to a pty, so hopefully everything is cleaned up across jobs. 

But yeah, this code is why I suggest "tigerbeetle exits if stdin is closed"

matklad:
Heh, that's this befuddling crash: 

https://github.com/tigerbeetle/tigerbeetle/actions/runs/6498235776/job/17649591312?pr=1221

Added some more prints around stuff.

src/clients/dotnet/TigerBeetle.Tests/IntegrationTests.cs
```diff
@@ -1130,8 +1132,42 @@ public TBServer()
...
+                var listening = "listening on ";
+                var found = line.IndexOf(listening);
+                if (found != -1)
+                {
+                    if (addressLogged != "") throw new InvalidOperationException("have already read the port");
+                    addressLogged = line.Substring(found + listening.Length).Trim();
+                    readPortEvent.Set();
+                }
+            };
+            process.BeginErrorReadLine();
```
batiati, line 1159: > https://github.com/tigerbeetle/tigerbeetle/actions/runs/6498235776/job/17649591312?pr=1221

It's bizarre, it couldn't find any "listening on" string on the stderr:

```
System.InvalidOperationException: failed to read the port, exited=False stderr:
info(io): opening "dotnet-tests.tigerbeetle"...warning(main): Grid cache size of 256MB is small. See --cache-grid
```

My guess (and it's only a guess, I haven't debugged), is that the event only fires when the stderr receives a `\n`... being "listening on" the last message, it does not have a line break to fire the event. It's easy to test this hypothesis, by just putting a `\n` on TB's side.

matklad: We actually do add `\n` here -- `listening on` is a log line, logs go to unbufferred stderr and include `\n`. What's more, I included `.WaitForExit()` in the throwing branch, as, according to the docs, it should collect all the output there is. And there's no output after this line. 

Actually, I think I get it! The next line should read:

```
info(main): 0: Allocated 5099MB in 15 regions during replica init (Grid Cache: 256MB)
```

And we are not getting that! 

I bet we just OOM killed!

batiati: Makes sense, so try to print the exit code, MacOS always gives 137 when the OOM hits

batiati: I think I encountered this problem before when the grid cache was updated to 256MB:
https://github.com/tigerbeetle/tigerbeetle/blob/998668b8bcd6ededef09ee28060df76dc8848da6/.github/workflows/clients-java.yml#L273

matklad:
Ok, I belive the failures currently exist on main, as it seems related to the recently increased RAM consumption, and not to the current change. Let's merge it, so that we can easier debug further flakiness?

src/clients/dotnet/TigerBeetle.Tests/IntegrationTests.cs
```diff
@@ -1130,8 +1132,42 @@ public TBServer()
...
+
+            var readPortEvent = new AutoResetEvent(false);
+            var addressLogged = "";
+            var stderrLogged = new StringBuilder();
+            process.ErrorDataReceived += (sender, e) =>
+            {
+                var line = e.Data;
+                if (line == null) return;
+                stderrLogged.Append(line);
+                var listening = "listening on ";
```
batiati: Just a nit I noticed while trying to debug the mystery:

```suggestion
                const string listening = "listening on ";
```

batiati (approved):

PR #1224, opened 2023-10-13
java: don't hard-code port in integration tests

matklad:
The same approach we use in tmp_tigerbeetle and C# --- let the OS pick an open port for US.


batiati (approved):
:+1:

PR #1228, opened 2023-10-17
vsr: align grid to block

matklad:
Aligning blocks this way makes it more likely that they are aligned to the underlying physical sector size, and is plainly nicer. 

While doing that, I've noticed that we have divergent assertions between the real storage and the testing storage. While it a fun puzzle to ponder, which assertions are more important, from the testing storage, or from the real one, I decided to just merge the two :)

### DIFF
--- a/src/vsr.zig
+++ b/src/vsr.zig
@@ -61,12 +61,25 @@ pub const Zone = enum {
     wal_headers,
     wal_prepares,
     client_replies,
+    // Add padding between `client_replies` and `grid`, to make sure grid blocks are aligned to
+    // block size and not just to sector size. Aligning blocks this way makes it more likely that
+    // they are aligned to the underlying physical sector size. This padding is zeroed during
+    // format, but isn't used otherwise.
+    grid_padding,
     grid,
 
     const size_superblock = superblock.superblock_zone_size;
     const size_wal_headers = constants.journal_size_headers;
     const size_wal_prepares = constants.journal_size_prepares;
     const size_client_replies = constants.client_replies_size;
+    const size_grid_padding = size_grid_padding: {
+        const grid_start_unaligned = size_superblock +
+            size_wal_headers +
+            size_wal_prepares +
+            size_client_replies;
+        const grid_start_aligned = std.mem.alignForward(usize, grid_start_unaligned, constants.block_size);
+        break :size_grid_padding grid_start_aligned - grid_start_unaligned;
+    };
 
     comptime {
         for (.{
@@ -74,9 +87,15 @@ pub const Zone = enum {
             size_wal_headers,
             size_wal_prepares,
             size_client_replies,
+            size_grid_padding,
         }) |zone_size| {
             assert(zone_size % constants.sector_size == 0);
         }
+
+        for (std.enums.values(Zone)) |zone| {
+            assert(Zone.start(zone) % constants.sector_size == 0);
+        }
+        assert(Zone.start(.grid) % constants.block_size == 0);
     }
 
     pub fn offset(zone: Zone, offset_logical: u64) u64 {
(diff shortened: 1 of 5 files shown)


src/vsr.zig
```diff
@@ -61,13 +61,25 @@ pub const Zone = enum {
     wal_headers,
     wal_prepares,
     client_replies,
+    // Note: there might be padding between `client_replies` and `grid`, to make sure grid blocks
+    // are aligned to block size and not just to sector size. Aligning blocks this way makes it more
+    // likely that they are aligned to the underlying physical sector size.
```
matklad: Originally I went with just adding an explicit `.gird_padding` Zone, but but that would require some non-sensial assertions that we _dont_ actually use this zone.

src/vsr.zig
```diff
@@ -105,6 +125,21 @@ pub const Zone = enum {
...
+    /// Ensures that the read or write is aligned correctly for Direct I/O.
+    /// If this is not the case, then the underlying syscall will return EINVAL.
+    /// We check this only at the start of a read or write because the physical sector size may be
+    /// less than our logical sector size so that partial IOs then leave us no longer aligned.
+    pub fn verify_iop(zone: Zone, buffer: []const u8, offset_in_zone: u64) void {
+        if (zone.size()) |zone_size| {
+            assert(offset_in_zone + buffer.len <= zone_size);
+        }
+        assert(@intFromPtr(buffer.ptr) % constants.sector_size == 0);
+        assert(buffer.len % constants.sector_size == 0);
```
sentientwaffle, line 138: Also could check `buffer.len > 0` as a sanity-check.

src/vsr.zig
```diff
@@ -105,6 +125,21 @@ pub const Zone = enum {
             .grid => null,
         };
     }
+
+    /// Ensures that the read or write is aligned correctly for Direct I/O.
+    /// If this is not the case, then the underlying syscall will return EINVAL.
+    /// We check this only at the start of a read or write because the physical sector size may be
+    /// less than our logical sector size so that partial IOs then leave us no longer aligned.
+    pub fn verify_iop(zone: Zone, buffer: []const u8, offset_in_zone: u64) void {
```
sentientwaffle, line 133: It's great to have all of these assertions in one place! :tada:

src/vsr.zig
```diff
@@ -61,13 +61,25 @@ pub const Zone = enum {
...
+    // are aligned to block size and not just to sector size. Aligning blocks this way makes it more
+    // likely that they are aligned to the underlying physical sector size.
     grid,
 
     const size_superblock = superblock.superblock_zone_size;
     const size_wal_headers = constants.journal_size_headers;
     const size_wal_prepares = constants.journal_size_prepares;
     const size_client_replies = constants.client_replies_size;
 
+    const size_grid_padding = size_grid_padding: {
```
sentientwaffle, line 75: Should we zero this padding in `replica_format.zig`, when we zero the client replies zone?

jorangreef:
Do we happen to have the grid already aligned with our constants as is?

I realize that the WAL prepares should also be aligned to prepare size (WAL headers should already be aligned to 4 KiB).

But if WAL prepares are aligned, (and assuming client replies don't break grid alignment, or can be moved to after superblock, with padding in the superblock to compensate), then we might be able to avoid grid padding entirely, in favor of an assertion that there is none?

matklad:
The padding is definitely non-zero for at least some configurations. But it’s a great point that we can extend existing superblock padding, rather than add an extra padding here.

sentientwaffle:
(I had assumed that the reason we were padding client replies rather than the superblock is that the WAL headers are a multiple of the sector size, but not necessarily a multiple of the block size.)

matklad:
Yeah, I've looked into padding superblock, and it is messy: 

* the padding somewhat substantially increases the size of `SuperBlockHeader` (more than 2x). The size of the superblock including trailers is not affected much, as that's ~170 blocks anyway
* for production config, because journal_slot_count is a nice power-of-two, it's enough to make the size of the superblock to be a multiple of block size to align the grid.
* even though WAL heades are aligned to sector, we can still pad the superblock zone such that they two together are just right in size
* *But* it might not be possible to pad just the superblock -- given that we have multiple copies of the superblock, you can't in general pad the superblock size by the arbitrary number of sectors. 
* And even padding superblock to be a multiple of block size is tricky implementation wise due to interdependence between freeset trailer size and the size of the superblockblock header (with padding)

I'd say let's go with the strategy from this PR (+ zeroing in fromat), and revisit after we remove trailers.

sentientwaffle (approved):

PR #1229, opened 2023-10-18
dotnet: check formatting on CI

matklad:
Also remove BOM and translate CRLF -> LF


matklad:
@batiati do you know by any chance to force dotnet format use \n without explicit config through .editorconfig?

batiati:
> @batiati do you know by any chance to force dotnet format use \n without explicit config through .editorconfig?

I'm afraid we need a .editorconfig file to parameterize it. The bright side is that we now can have a consistent format across Linux and Windows, and VSCode or Visual Studio as well.
Sometimes I use Visual Studio on Windows, and it's so frustrating when it creates new files with CRLF despite all other files being LF.

src/clients/dotnet/.editorconfig
```diff
@@ -0,0 +1,7 @@
+# By default, `dotnet format` uses `\r\n` line endings on Windows. To override this behavior, we
+# need this editorconfig file.
+root = true
+
+[*]
+end_of_line = lf
+insert_final_newline = true
```
batiati, line 7: Let's be more opinionated:

```suggestion
end_of_line = lf
insert_final_newline = true
charset=utf-8
indent_style=space
indent_size=4
```

Reference:

- [charset](https://github.com/editorconfig/editorconfig/wiki/EditorConfig-Properties#charset)
- [indent_style](https://github.com/editorconfig/editorconfig/wiki/EditorConfig-Properties#indent_style)
- [indent_size](https://github.com/editorconfig/editorconfig/wiki/EditorConfig-Properties#indent_size)

matklad: Originally I thought to stick to defaults as much as possible, but, indeed, if we have to configure something, we could as well pin down more things!

batiati (approved):

PR #1230, opened 2023-10-18
CI: disable large instances

matklad:
They are too expensive. We need to figure _something_ out, but, for now, let's scale down to just ubuntu latest.


matklad:
Closing in favor of https://github.com/tigerbeetle/tigerbeetle/pull/1231, as my branch name broke our linting shell scripts :facepalm:

PR #1234, opened 2023-10-19
LSM: Manifest linked-list; Superblock CheckpointState

sentientwaffle:
Context: The superblock trailer is a SoA of references to every manifest block.

In this commit:
- Remove the superblock manifest trailer.
- Instead, store it implicitly – each manifest block stores the reference to the (chronologically) previous manifest block.
- Add `VSRState.checkpoint: CheckpointState` field. `CheckpointState` is a struct which holds the subset of `VSRState` that is _deterministic at a checkpoint_.
- Add the `request_sync_checkpoint` and `sync_checkpoint` message. The message body of `sync_checkpoint` is exactly a `CheckpointState`!

Not in this commit:
- The upper-bound of manifest blocks is still 65K — that will be fixed in a follow-up PR.
- The client sessions and freeset trailers will also be removed. (The latter as part of the "manifest freeset".)
- Client sessions will be stored directly in the superblock header. They will also need to be included in the `sync_checkpoint` message.

The reason `CheckpointState` is stored within (rather than alongside) `VSRState` is so that `monotonic` and `would_be_updated_by` still work.

Using `CheckpointState` both in the superblock storage and the `sync_checkpoint` message makes it very clear exactly what state is required to sync the replica to a new checkpoint.


src/vsr/replica_test.zig
```diff
@@ -1506,29 +1507,18 @@ const TestReplicas = struct {
...
+            // updated its sync_op_max.
+            maybe(replica.superblock.staging.vsr_state.sync_op_max > 0);
 
-                try expectEqual(
-                    std.mem.bytesToValue(vsr.Header, block_want[0..@sizeOf(vsr.Header)]),
-                    std.mem.bytesToValue(vsr.Header, block_got[0..@sizeOf(vsr.Header)]),
-                );
-            }
+            try t.cluster.storage_checker.replica_sync(&replica.superblock);
         }
```
sentientwaffle, line 1522: The previous strategy for checking the grid doesn't always work (we were "lucky" before) for the same reason that the storage checker cannot verify determinism except at checkpoints:
Manifest block addresses are acquired and then held onto for potentially multiple bars before they are written.

Instead, we just trigger the storage checker, which already knows how to verify all of this properly!

docs/internals/lsm.md
sentientwaffle, line 1: (Most of the changes here were due to stuff that was outdated since the unified manifest.)

src/vsr/replica.zig
```diff
@@ -2469,11 +2489,6 @@ pub fn ReplicaType(
                 stage.trailers.get(trailer).offset,
                 target_size,
             });
-
-            if (stage.done()) {
-                assert(progress == .complete);
-                self.sync_requesting_trailers_callback();
-            }
```
sentientwaffle, line 2476: `on_sync_checkpoint()` just checks `if (stage.done())` before accepting the "checkpoint state".

Since the other trailer steps are going to go away soon the performance impact of potentially ignoring a `sync_checkpoint` that arrives before the trailers won't be an issue, and it keeps things simpler.

src/vsr/replica.zig
```diff
@@ -2402,13 +2402,48 @@ pub fn ReplicaType(
...
+            });
+
+            const checkpoint_state = std.mem.bytesAsValue(
+                vsr.CheckpointState,
+                message.body()[0..@sizeOf(vsr.CheckpointState)],
+            );
+
+            if (stage.done()) {
+                self.sync_requesting_trailers_callback(checkpoint_state);
+            }
```
sentientwaffle, line 2437: When the client sessions trailer is removed, the client sessions will move into the `sync_checkpoint` message body too, immediately following the `CheckpointState` in the message body.

docs/internals/vsr.md
```diff
@@ -271,9 +271,9 @@ Sync missed client replies using [Protocol: Repair Grid](#protocol-repair-client
 (Runs immediately after [Protocol: Sync Superblock](#protocol-sync-superblock).)
 See [State Sync](./sync.md) for details.
 
-## Protocol: Sync Tables
+## Protocol: Sync Forest
 
-Sync missed LSM table blocks using [Protocol: Repair Grid](#protocol-repair-grid).
+Sync missed LSM manifest and table blocks using [Protocol: Repair Grid](#protocol-repair-grid).
```
sentientwaffle, line 276: This is sort of aspirational atm. Right now manifest blocks are still repaired via regular grid repair -- i.e., the syncing replica reads the block, sees that it's missing, and repairs that block. But that limits manifest repair to be sequential.

Long-term:
- the healthy replica should see notice when a manifest block is requested, and send long with it the handful of manifest blocks that precede it
- the syncing replica should cache blocks that it receives before it has requested, just in case it needs them momentarily.

src/lsm/manifest_log.zig
```diff
@@ -42,15 +39,6 @@ const RingBuffer = @import("../ring_buffer.zig").RingBuffer;
...
 const TableInfo = schema.Manifest.TableInfo;
 
-comptime {
-    // We need enough manifest blocks to accommodate the upper-bound of tables.
-    // *10 is an arbitrary lower-bound of available space for amplification, which allows compaction
-    // to be spaced more efficiently.
-    const block_entries = schema.Manifest.entry_count_max;
-    const manifest_entries = block_entries * vsr.superblock.manifest_blocks_max;
-    assert(manifest_entries >= tree.table_count_max * 10);
-}
```
sentientwaffle, line 52: I didn't bother updating this since it will be unnecessary (and incorrect) once the manifest block upper-bound is in place.

src/lsm/manifest_log_fuzz.zig
```diff
@@ -3,13 +3,10 @@
...
-//!   the state of the ManifestLog/SuperBlock.Manifest immediately after the latest checkpoint.
-//! - SuperBlock.Manifest.open() only returns the latest version of each table.
-//! - SuperBlock.Manifest's compaction queue contains any blocks which:
-//!   - contain fewer than entry_count_max entries, or
-//!   - contain a "remove" entry, or
-//!   - contain an overridden entry.
+//! - The state of the ManifestLog immediately after recovery matches
+//!   the state of the ManifestLog immediately after the latest checkpoint.
+//! - ManifestLog.open() only returns the latest version of each table.
+//! - The ManifestLog performs enough compaction to not "fall behind" (i.e. run out of blocks).
```
sentientwaffle, line 9: > The ManifestLog performs enough compaction to not "fall behind" (i.e. run out of blocks).

Ah, this last point snuck into my patch; it will actually apply during the next PR when I introduce the manifest log upper bound.

src/testing/cluster/storage_checker.zig
```diff
@@ -86,59 +87,115 @@ pub const StorageChecker = struct {
...
+        if (!syncing) assert(checker.checkpoints.count() > 0);
+    }
 
-        const syncing = superblock.working.vsr_state.sync_op_max > 0;
-        if (!syncing) {
-            checkpoint_actual.put(.client_replies, checksum_client_replies(superblock));
-            checkpoint_actual.put(.grid, checker.checksum_grid(superblock));
-        }
+    /// Invoked when both superblock and content sync is complete.
+    pub fn replica_sync(checker: *StorageChecker, superblock: *const SuperBlock) !void {
```
sentientwaffle, line 108: See `replica_test.zig` for why this is needed.

src/vsr.zig
```diff
@@ -320,8 +323,6 @@ pub const Header = extern struct {
     /// * A `do_view_change` sets this to a bitset of "present" prepares. If a bit is set, then
     ///   the corresponding header is not "blank", the replica has the prepare, and the prepare
     ///   is not known to be faulty.
-    /// * A `sync_free_set` sets this to the vsr_state.previous_checkpoint_id.
-    /// * A `sync_client_sessions` sets this to the vsr_state.commit_min_checksum.
```
sentientwaffle, line 324: These parameters are now just part of the `CheckpointState` sent in `sync_checkpoint`.

matklad:
Nice PR number!

docs/internals/lsm.md
```diff
@@ -254,7 +253,8 @@ The manifest log is periodically compacted to remove older entries that have bee
 newer entries. For example, if a table is created and later deleted, manifest log compaction
 will eventually remove any reference to the table from the log blocks.
 
-All manifest blocks are tracked in the superblock manifest.
+Each manifest block has a reference to the (chronologically) previous manifest block.
```
matklad, line 256: Mention that this reference may "dangle" after compaction.

src/vsr/superblock.zig
```diff
@@ -255,8 +272,37 @@ pub const SuperBlockHeader = extern struct {
...
+
+        manifest_head_checksum: u128,
+        manifest_tail_checksum: u128,
+        manifest_head_address: u64,
+        manifest_tail_address: u64,
+
+        /// The last operation committed to the state machine. At startup, replay the log hereafter.
+        commit_min: u64,
+
+        manifest_count: u32,
```
matklad: Unclear what we count here: live tables? manifest log events? manifest log blocks?

src/vsr/superblock.zig
```diff
@@ -378,19 +423,52 @@ pub const SuperBlockHeader = extern struct {
...
+};
+
+pub const ManifestReferences = struct {
+    /// The chronologically first manifest block in the chain.
+    head_checksum: u128,
+    head_address: u64,
+    /// The chronologically last manifest block in the chain.
+    tail_checksum: u128,
+    tail_address: u64,
+    /// The number if manifest blocks in the chain.
```
matklad: s/if/of

src/lsm/schema.zig
```diff
@@ -458,9 +460,29 @@ pub const Manifest = struct {
...
+            assert(header.parent == 0);
+
+            return null;
+        } else {
+            return .{
+                .checksum = header.parent,
+                .address = header.commit,
+            };
+        }
+    }
```
matklad, line 486: Mostly unrelated, but maybe call this whole struct `schema.ManifestNode`, to not confuse with actual manifest?

src/vsr/superblock.zig
sentientwaffle: :+1: I'll rename to `manifest_block_count`.

src/vsr/superblock.zig
```diff
@@ -255,8 +272,37 @@ pub const SuperBlockHeader = extern struct {
...
             // If commit_min is 0, we have never checkpointed, so no compactions are checkpointed.
-            return state.commit_min > 0 and
-                op <= vsr.Checkpoint.trigger_for_checkpoint(state.commit_min).?;
+            return state.checkpoint.commit_min > 0 and
+                op <= vsr.Checkpoint.trigger_for_checkpoint(state.checkpoint.commit_min).?;
+        }
+    };
+
+    /// The content of CheckpointState is deterministic for the corresponding checkpoint.
+    pub const CheckpointState = extern struct {
```
matklad, line 291: One alternative would be to do the dual thing, and introduce `ReplicaState` for things that _vary_ per replica. This makes a tiny bit more sense to me logically, as "unique" state feels like more of the special case. 

OTOH, the reason why we need this struct is because we need a struct to send over the wire during the sync. 

Maybe reflect that in the doc comment? "This drives the sync" is more fundamental than "this is consistent across cluster".

src/vsr/superblock.zig
```diff
@@ -201,42 +201,59 @@ pub const SuperBlockHeader = extern struct {
...
             assert(state.sync_op_max >= state.sync_op_min);
             assert(state.view >= state.log_view);
             assert(state.replica_count > 0);
             assert(state.replica_count <= constants.replicas_max);
             assert(vsr.member_index(&state.members, state.replica_id) != null);
+
+            if (state.checkpoint.manifest_count == 0) {
+                assert(state.checkpoint.manifest_head_address == 0);
+                assert(state.checkpoint.manifest_tail_address == 0);
+            }
```
matklad: I think this should assert in both direction?

`assert((state.checkpoint.manifest_count == 0) == (state.checkpoint.manifest_head_address == 0) == (state.checkpoint.manifest_tail_address == 0))`

src/vsr/superblock.zig
```diff
@@ -201,42 +201,59 @@ pub const SuperBlockHeader = extern struct {
...
+
+            if (state.checkpoint.manifest_head_address == 0) {
+                assert(state.checkpoint.manifest_head_checksum == 0);
+            }
+
+            if (state.checkpoint.manifest_tail_address == 0) {
+                assert(state.checkpoint.manifest_tail_checksum == 0);
+                assert(state.checkpoint.manifest_head_address == 0);
+            }
```
matklad, line 228: Let's also assert that `(head_addres == tail_address) == (head_checksum == tail_checksum)`

src/vsr/superblock.zig
```diff
@@ -201,42 +201,59 @@ pub const SuperBlockHeader = extern struct {
...
         }
 
         pub fn monotonic(old: VSRState, new: VSRState) bool {
             old.assert_internally_consistent();
             new.assert_internally_consistent();
-            if (old.commit_min == new.commit_min) {
-                if (old.commit_min_checksum == 0 and old.commit_min == 0) {
+            if (old.checkpoint.commit_min == new.checkpoint.commit_min) {
```
matklad, line 234: Can we, in this branch, assert that the checkpoints are `.equal_bytes`?

src/vsr/superblock.zig
```diff
@@ -821,18 +864,27 @@ pub fn SuperBlockType(comptime Storage: type) type {
...
+            assert(update.commit_min_checksum !=
+                superblock.staging.vsr_state.checkpoint.commit_min_checksum);
             assert(update.sync_op_min <= update.sync_op_max);
 
             var vsr_state = superblock.staging.vsr_state;
-            vsr_state.commit_min_checksum = update.commit_min_checksum;
-            vsr_state.commit_min_canonical = vsr_state.commit_min;
-            vsr_state.commit_min = update.commit_min;
+            vsr_state.checkpoint = .{
+                .previous_checkpoint_id = superblock.staging.checkpoint_id(),
```
matklad, line 886: I am wondering if we could use `CheckpointState` here, and remove `UpdateCheckpoint` struct altogethrer? Feels ok to let the caller to set previous_checkpoint_id.

src/vsr/replica.zig
```diff
@@ -8394,12 +8420,59 @@ pub fn ReplicaType(
...
+            );
+
+            reply.header.* = .{
+                .command = .sync_checkpoint,
+                .cluster = self.cluster,
+                .replica = self.replica,
+                .size = @sizeOf(Header) + @sizeOf(vsr.CheckpointState),
+                .parent = self.superblock.working.checkpoint_id(),
+                .op = self.op_checkpoint(),
+            };
```
matklad: I am wondering, can we _first_ set the `header`, and then use `reply.body()` with `copy_disjoint(.exact)` to double-checkthat we didn't mess up the lengths?

src/vsr.zig
```diff
@@ -873,20 +869,21 @@ pub const Header = extern struct {
         return null;
     }
 
-    fn invalid_request_sync_free_set(self: *const Header) ?[]const u8 {
-        assert(self.command == .request_sync_free_set);
+    fn invalid_sync_checkpoint(self: *const Header) ?[]const u8 {
+        assert(self.command == .sync_checkpoint);
+        if (self.size != @sizeOf(Header) + @sizeOf(CheckpointState)) return "size != @sizeOf(Header) + @sizeOf(CheckpointState)";
```
matklad: tidy line length

src/lsm/manifest_log.zig
```diff
@@ -312,42 +348,78 @@ pub fn ManifestLogType(comptime Storage: type) type {
...
+            log.debug("{}: opened: checksum={} address={} entries={}", .{
+                manifest_log.superblock.replica_index.?,
+                block_checksum,
+                block_address,
+                block_schema.entry_count,
+            });
+
+            const checkpoint_state = &manifest_log.superblock.working.vsr_state.checkpoint;
+            if (checkpoint_state.manifest_head_address == block_address) {
+                // When we find the "head" block, stop iterating the linked list – any more blocks
```
matklad: Wait, so the "head" block is the oldest one, and the "tail" block is the newest one? And `tail`' `previous` points (transitively) to the head? I was thinking the opposite this whole time! I guess it might make sense to duplicate comment from `ManifestReferences` onto `CheckPointState`, as the latter is the ground truth. 

I am also wondering if this is genuinelly flipped, mainly due to the `previous` link. If I start with the head, and follow previous links, I expect to reach the tail, and not vice verse.

The current naming _also_ makes sense to me from the perspective that we add entries to the `tail`. Maybe just call them `newest` and `oldest`?

src/lsm/manifest_log.zig
```diff
@@ -756,24 +807,57 @@ pub fn ManifestLogType(comptime Storage: type) type {
...
+            assert(manifest_log.write_callback == null);
+            assert(manifest_log.grid_reservation == null);
+            assert(manifest_log.log_block_checksums.count ==
+                manifest_log.log_block_addresses.count);
+            assert(manifest_log.blocks.count == 0);
+            assert(manifest_log.blocks_closed == 0);
+            assert(manifest_log.entry_count == 0);
+
+            if (manifest_log.log_block_addresses.count == 0) {
+                return std.mem.zeroes(vsr.SuperBlockManifestReferences);
```
matklad, line 836: We are inconsistent about "neutral" element of `SuperBlockManifestReferences)`. Here we return zeros, but in `superblock.manifest_references` we use `null`. 

Given that we need to store _something_ on disk, let's maybe get rid of `null` and add `fn empty() bool` method to `SuperBlockManifestReferences)` for branches&asserts?

src/lsm/manifest_log.zig
```diff
@@ -219,87 +244,100 @@ pub fn ManifestLogType(comptime Storage: type) type {
...
 
             manifest_log.open_event = event;
-            manifest_log.open_iterator = manifest_log.superblock.manifest.iterator_reverse();
-
             manifest_log.reading = true;
             manifest_log.read_callback = callback;
 
-            manifest_log.open_read_block();
+            if (manifest_log.superblock.working.manifest_references()) |references| {
+                manifest_log.open_read_block(.{
```
matklad, line 265: Assert `references.count` here (eg, that it's smaller than 65536. With a true bound we'll turn this into a true assert)

src/lsm/manifest_log.zig
```diff
@@ -684,26 +743,20 @@ pub fn ManifestLogType(comptime Storage: type) type {
             }
 
             log.debug("compacted: checksum={} address={} frees={}/{}", .{
```
matklad: Can we work replica_index in here?

src/vsr/superblock.zig
sentientwaffle: No, but I'll add a `maybe()` with an explanation to the code. The reason is that if we are syncing to recover from a checkpoint divergence, then the state might be different even if the `commit_min`s match.

src/lsm/manifest_log.zig
matklad:
```suggestion
            log.debug("compacted: checksum={} address={} free={}/{}", .{
```

I think we are counting the total amount of free entries, rather than the amount of entries we additionally freed in this round. 

Although, the original formulation, where we start from zero, makes more sense to me. We already can infer how full the blocks are from `entry_count`. What we can't infer is exactly the number of free event, the delta in the block size.

src/lsm/manifest_log.zig
```diff
@@ -684,26 +743,20 @@ pub fn ManifestLogType(comptime Storage: type) type {
...
 
-            // Blocks may be compacted if they contain frees, or are not completely full.
-            // (A partial block may be flushed as part of a checkpoint.)
-            assert(frees > 0 or block_schema.entry_count < schema.Manifest.entry_count_max);
-            // At most one block could have been filled by the compaction.
+            // Blocks are compacted in sequence – not skipped, even if no entries will be freed.
+            // (That should be rare though, since blocks are large.)
+            // This is necessary to update the block's "previous block" pointer in the header.
+            maybe(frees == 0);
             assert(manifest_log.blocks_closed <= 1);
```
matklad, line 765: Unreleated, but why do we size the blocks dynamically? Could use just two buffer blocks, and flush the block immediately as it gets full?

src/vsr/superblock.zig
sentientwaffle: We can't remove `UpdateCheckpoint` because we need to update the `commit_max` and `sync_op_{min,max}`, which are not part of the `CheckpointState`.

src/lsm/manifest_log.zig
```diff
@@ -756,24 +809,57 @@ pub fn ManifestLogType(comptime Storage: type) type {
...
             assert(!manifest_log.blocks.full());
 
             manifest_log.blocks.advance_tail();
 
             const block: BlockPtr = manifest_log.blocks.tail().?;
             const block_address = manifest_log.grid.acquire(manifest_log.grid_reservation.?);
-            assert(!manifest_log.superblock.manifest.queued_for_compaction(block_address));
+
+            const tail_checksum = manifest_log.log_block_checksums.tail();
+            const tail_address = manifest_log.log_block_addresses.tail();
```
matklad:
```
            const tail_checksum = manifest_log.log_block_checksums.tail() orelse 0;
            const tail_address = manifest_log.log_block_addresses.tail() orelse 0;
```

reads a tiny bit better to me, as it keeps the logic closer to the count assert that the two are synchronized.

src/lsm/manifest_log.zig
```diff
@@ -819,16 +905,14 @@ pub fn ManifestLogType(comptime Storage: type) type {
             header.set_checksum();
             verify_block(block, null, null);
 
-            manifest_log.superblock.manifest.append(header.checksum, address);
-            if (entry_count < schema.Manifest.entry_count_max) {
-                manifest_log.superblock.manifest.queue_for_compaction(address);
-            }
+            manifest_log.log_block_checksums.push_assume_capacity(header.checksum);
```
matklad, line 919: Maybe add an explicit assert against capacity here, we'll need it anyway once we have true upper bound.

src/lsm/schema.zig
```diff
@@ -458,10 +460,30 @@ pub const Manifest = struct {
...
 
+        if (header.commit == 0) {
+            assert(header.parent == 0);
+        }
+
         return .{ .entry_count = context.entry_count };
     }
 
-    pub fn size(schema: *const Manifest) u32 {
+    pub fn previous(manifest_block: BlockPtrConst) ?struct { checksum: u128, address: u64 } {
```
matklad, line 472: We _could_ pass-in the head we know from the superblock, so that we can return `null` it it is really `null` or just dangling. 

The beneifit here is that the call-site will be aware that the chain of blocks might be cut short.

Though, this feels like overengieering to be honest, the current API is simple (but maybe add a comment that this might point to a garbage block?)

matklad: I am also wondering -- by the time we reach the last block when opening manifest, can we assert anything about that entry with respect to the free set? Ie, that the precious entry should always be free?

src/lsm/segmented_array.zig
```diff
@@ -1,6 +1,7 @@
 const std = @import("std");
 
 const assert = std.debug.assert;
+const maybe = stdx.maybe;
```
matklad, line 4: dead code, but better leave a is, we'll use it as some point anyway!

src/lsm/manifest_log.zig
sentientwaffle: > If I start with the head, and follow previous links, I expect to reach the tail, and not vice verse.

Ahh that makes so much sense! (For context: I named the head/tail because I was thinking of it from the perspective of the manifest log's `log_block` ring buffers, where "log_block_{addresses,checksums}.tail" is the newest.)

I renamed them to `newest`/`oldest` instead -- that way they can't be mixed up.

src/testing/cluster/storage_checker.zig
```diff
@@ -86,59 +87,115 @@ pub const StorageChecker = struct {
...
+            if (mismatch) return error.StorageMismatch;
+        } else {
+            // This replica is the first to reach op_checkpoint.
+            // Save its state for other replicas to check themselves against.
+            var checkpoint: Checkpoint = undefined;
+            for (std.enums.values(CheckpointArea)) |area| {
+                if (checkpoint_actual.get(area)) |area_checksum| {
+                    checkpoint.set(area, area_checksum);
+                } else {
+                    return;
```
matklad: Hm, this `return` looks suspicions --- if a replica has a partial checkpoint, and is the first one with a checkpoint, we wont' be veryfying it. If this is OK, the comment would be helpful. 

Although even with a comment, it feels like something which could easily mutate to skipping all checkpoints. Can we add some sanity checks that `checker.checkpoints` eventually have all of the checkpoints?

src/vsr/replica.zig
```diff
@@ -7820,14 +7846,13 @@ pub fn ReplicaType(
 
             const stage: *const SyncStage.UpdatingSuperBlock = &self.syncing.updating_superblock;
 
-            assert(self.superblock.working.vsr_state.commit_min == stage.target.checkpoint_op);
             assert(self.superblock.working.checkpoint_id() == stage.target.checkpoint_id);
-            assert(self.superblock.working.vsr_state.previous_checkpoint_id ==
-                stage.previous_checkpoint_id);
-            assert(self.superblock.working.vsr_state.commit_min_checksum ==
-                stage.checkpoint_op_checksum);
+            assert(std.meta.eql(
```
matklad: stdx.equal_bytes would be simpler & faster

src/vsr/replica.zig
```diff
@@ -2402,13 +2402,48 @@ pub fn ReplicaType(
...
+
+            self.send_sync_checkpoint(.{ .replica = message.header.replica });
+        }
+
+        fn on_sync_checkpoint(self: *Self, message: *const Message) void {
+            assert(message.header.replica < self.replica_count);
+            assert(message.header.command == .sync_checkpoint);
+            if (self.ignore_sync_response_message(message)) return;
+
+            const stage: *SyncStage.RequestingTrailers = &self.syncing.requesting_trailers;
```
```suggestion
            const stage: *const SyncStage.RequestingTrailers = &self.syncing.requesting_trailers;
```

src/vsr/sync.zig
```diff
@@ -29,28 +29,18 @@ pub const Stage = union(enum) {
...
-            assert(self.checkpoint_op_checksum != null);
             return true;
         }
     };
 
     pub const UpdatingSuperBlock = struct {
         target: Target,
-        previous_checkpoint_id: u128,
-        checkpoint_op_checksum: u128,
+        checkpoint_state: vsr.CheckpointState,
```
matklad, line 43: Looked at where we might be copying Stage as I think sizeof here is not tiny. Found 

```
                // We had a sync target already, but we discovered a newer one.
                // Re-sync to the new target.
                .requesting_trailers => |stage| {
```

I think we might want to use `|*stage|` there

matklad (commented):
Finished review! Awesome!

The only two larger-than tiny coments are on the head/tail naming, and on the nullability/zeroability of MainfestReferences.

src/lsm/manifest_log.zig
sentientwaffle: We need enough blocks to accommodate all of the appends by table compaction (inserts, update snapshot_max, etc).
The number of blocks this can fill (in the worst case, where we compact all trees+levels) is dependent on how many entries can fit in each block -- i.e., it depends on the block size.
For example, in the production config, with 128KiB blocks, we need as many as 3 blocks.

In addition, right now we always compact exactly one block per half-bar. The next manifest PR will change that, to speed up compaction as the manifest size increases.

src/lsm/segmented_array.zig
sentientwaffle: Oops! I was tinkering with some asserts but ended up not committing them, I guess I left that in.

src/lsm/schema.zig
sentientwaffle: > I am also wondering -- by the time we reach the last block when opening manifest, can we assert anything about that entry with respect to the free set? Ie, that the precious entry should always be free?

I think that would be possible right now, but it definitely won't work once manifest compaction pacing is in (next PR).
The reason is that when the manifest is small (has few blocks) the manifest might decide to not do any compaction during a particular half-bar. That gives the freeset the opportunity to reuse the block even while the oldest manifest block still references it.

src/lsm/manifest_log.zig
matklad: >We need enough blocks to accommodate all of the appends by table compaction (inserts, update snapshot_max, etc).

But why we need it? Can't we spill blocks to disk as they are being filled, instead of writing them in one go in the end?

src/testing/cluster/storage_checker.zig
sentientwaffle: > Can we add some sanity checks that checker.checkpoints eventually have all of the checkpoints?

Hmm yeah this is confusing + possibly a footgun. I'm going to rework the storage checker so that it incrementally adds whatever info it can, even if (for example) it first encounters the checkpoint from a syncing replica.

(Context:
As it is now, it is possible (albeit unlikely) for checkpoints to be entirely missed from the storage checker's point of view.
The reason is that if a replica crashes part-way through checkpointing (for example, after it has written 3 out of 4 superblock copies), and then the replica recovers (into the new 3/4 checkpoint), then it never invokes the "on checkpoint" hook.)

src/lsm/manifest_log.zig
sentientwaffle: One constraint here is that we need to make sure that all of the inserts/updates/etc are appended in a deterministic order, regardless of the order that the table compactions actually finish in.

The way we handle that right now is: we wait until all of the (half-bar's) table compactions finish, and then we synchronously apply all the compactions' appends at once.

matklad (approved):

PR #1235, opened 2023-10-20
java: improve samples

matklad:
Two changes here, one big and one small:

* Remove "throws Exception" from our close method. As per AutoClosable docs:

     While this interface method is declared to throw Exception,
     implementers are strongly encouraged to declare concrete
     implementations of the close method to throw more specific
     exceptions, or to throw no exception at all if the close operation
     cannot fail.

  Covariant return types for the win!

* Assign `ClusterID` name to the magical `0` constant.


batiati (approved):
🚀

PR #1236, opened 2023-10-20
CI: touch up release process

matklad:
Final tweaks to the release process, the result of today's live debugging on the `release` branch.


.github/workflows/release.yml
```diff
@@ -8,7 +8,7 @@ on:
 jobs:
   release:
     runs-on: ubuntu-latest
-    secrets:
+    environment: release
```
matklad, line 11: Most secrets are still global, but we can now move them over to this environment. I didn't do that, as I don't actually know our node publish keys and such. But no changes to workflow would be required in any case.

.github/workflows/release.yml
```diff
@@ -39,15 +39,19 @@ jobs:
           node-version: '14.x'
           registry-url: 'https://registry.npmjs.org'
 
+      # We need llvm-lipo utility from LLVM to build a universal mac binary.
+      - run: wget https://apt.llvm.org/llvm.sh && chmod +x llvm.sh && sudo ./llvm.sh 16
       - run: ./scripts/install_zig.sh
-      - run: ./zig/zig build dist -- --build --publish --version 0.9.${{ github.run_number }} --sha ${{ github.sha }}
+      - run: ./zig/zig build release -- --build --publish --version=0.14.${{ github.run_number }} --sha=${{ github.sha }}
```
matklad, line 45: NB: Intend to bump our minor to 0.14 to signify new release process.

src/scripts/release.zig
```diff
@@ -153,10 +153,16 @@ fn build_tigerbeetle(shell: *Shell, info: VersionInfo, dist_dir: std.fs.Dir) !vo
     log.info("building tigerbeetle", .{});
     try shell.project_root.setAsCwd();
 
-    const llvm_lipo_version = shell.exec_stdout("llvm-lipo -version", .{}) catch {
+    const llvm_lipo = for (@as([2][]const u8, .{ "llvm-lipo-16", "llvm-lipo" })) |llvm_lipo| {
```
matklad, line 156: This tool is known as `llvm-lipo-$LLVM_VERSION` on ubunty, and just `llvm-lipo` for me locally. Wrote the code to try both for convenience.

.github/workflows/release.yml
```diff
@@ -39,15 +39,19 @@ jobs:
           node-version: '14.x'
           registry-url: 'https://registry.npmjs.org'
 
+      # We need llvm-lipo utility from LLVM to build a universal mac binary.
+      - run: wget https://apt.llvm.org/llvm.sh && chmod +x llvm.sh && sudo ./llvm.sh 16
```
matklad, line 43: Installing through LLVM rather than ubuntu package, to make sure the version is alinged with the one used by zig (which is usually fresher than what's publsihed with ubuntu)

.github/workflows/release.yml
```diff
@@ -39,15 +39,19 @@ jobs:
...
       - run: ./scripts/install_zig.sh
-      - run: ./zig/zig build dist -- --build --publish --version 0.9.${{ github.run_number }} --sha ${{ github.sha }}
+      - run: ./zig/zig build release -- --build --publish --version=0.14.${{ github.run_number }} --sha=${{ github.sha }}
         env:
           NUGET_KEY: ${{ secrets.NUGET_KEY }}
-          TIGERBEETLE_GO_GITHUB_TOKEN: ${{ secrets.TIGERBEETLE_GO_GITHUB_TOKEN }}
+          TIGERBEETLE_GO_PAT: ${{ secrets.TIGERBEETLE_GO_PAT }}
```
sentientwaffle, line 48: What is "PAT" short for?

matklad: Personal Access Token: https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/managing-your-personal-access-tokens

We use this token to clone & push to tigerbeetle-go repostiory over here:

https://github.com/tigerbeetle/tigerbeetle/pull/1236/files#diff-4dfde4c7ef7e6790beb4b0ef50fd5644d15d3dee53eeb2020046eb0de80e11ddL463

matklad: (the token itself starts with something like `github-pat`, so I consider `PAT` to be an appropriate abbreviation here)

sentientwaffle (approved):

PR #1237, opened 2023-10-20
ci: group output

matklad:
Now that our CI is driven by a single Zig step, as opposed to multiple yml steps, we get a problem with a single large printout for the whole job, which might be hard to navigate.

Luckily, GitHub actions have `::group::` command to solve this problem. And, to kill two birds with one stone, we can use the same groups to track elapsed time which is pretty important to motivate us to make our CI go faster!

### DIFF
--- a/src/shell.zig
+++ b/src/shell.zig
@@ -40,6 +40,9 @@ project_root: std.fs.Dir,
 
 env: std.process.EnvMap,
 
+/// True if the process is run in CI (the CI env var is set)
+ci: bool,
+
 pub fn create(gpa: std.mem.Allocator) !*Shell {
     var arena = std.heap.ArenaAllocator.init(gpa);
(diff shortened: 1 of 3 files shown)


src/shell.zig
```diff
@@ -40,19 +40,30 @@ project_root: std.fs.Dir,
...
     var project_root = try discover_project_root();
     errdefer project_root.close();
 
+    var env = try std.process.getEnvMap(gpa);
+    errdefer env.deinit();
+
+    const ci = env.get("CI") != null;
+
     var result = try gpa.create(Shell);
+    errdefer gpa.destroy(result);
```
matklad, line 59: This was a memory leak on error path before :weary:

cb22 (approved):
Looks great!

PR #1239, opened 2023-10-23
LSM: Revert groove to use update + insert and vendor hasUniqueRepresentation

cb22:
# Motivation
Whenever we update an object, we always have the old object around. Previously, we used to do a lookup to find the old one in any case, to give a simpler `upsert` function signature to the state machine - but this was resulting in duplicated work.

# Changes
Revert that, going back to having an explicit `update` and `insert` with the state machine choosing which is better for its needs. Additionally, switch to using `stdx.equal_bytes` in the Groove `update` hot path. This required vendoring a small fix for https://github.com/ziglang/zig/issues/17592.

# Benchmarks
These two things combined give a nice performance bump:

```sh
# on tmpfs
./scripts/benchmark.sh --print-batch-timings --transfer-count=10_000_000 --account-count=10 --id-order=sequential
```

Before:
```
1362 batches in 29.25 s
load offered = 1000000 tx/s
load accepted = 341867 tx/s
```

After:
```
1356 batches in 25.52 s
load offered = 1000000 tx/s
load accepted = 391846 tx/s
```

### DIFF
--- a/src/lsm/groove.zig
+++ b/src/lsm/groove.zig
@@ -899,8 +899,12 @@ pub fn GrooveType(
 
             if (constants.verify) {
                 const old_from_cache = groove.objects_cache.get(@field(old, primary_field)).?;
-                assert(@field(old_from_cache, primary_field) == @field(old, primary_field));
-                assert(old_from_cache.timestamp == old.timestamp);
+
+                // While all that's actually required is that the _contents_ of the old_from_cache
+                // and old objects are identical, in current usage they're always the same piece of
+                // memory. We'll assert that for now, and this can be weakened in future if
+                // required.
+                assert(old_from_cache == old);
             }
 
             // Sanity check to ensure the caller didn't accidentally pass in an alias.
@@ -925,9 +929,6 @@ pub fn GrooveType(
                 assert(!tombstone(old) and !tombstone(new));
             }
 
-            groove.objects.put(new);
-            groove.objects_cache.upsert(new);
-
             inline for (std.meta.fields(IndexTrees)) |field| {
                 const Helper = IndexTreeFieldHelperType(field.name);
                 const old_index = Helper.derive_index(old);
@@ -946,6 +947,13 @@ pub fn GrooveType(
                     }
                 }
             }
+
+            // Putting the objects_cache upsert after the index tree updates is critical:
+            // We diff the old and new objects, but the old object will be a pointer into the
+            // objects_cache. If we upsert first, old.* == new.* and no secondary indexes will
+            // be updated!
+            groove.objects_cache.upsert(new);
+            groove.objects.put(new);
         }
 
         /// Asserts that the object with the given PrimaryKey exists.
--- a/src/lsm/groove.zig
+++ b/src/lsm/groove.zig
@@ -890,7 +890,13 @@ pub fn GrooveType(
 
         /// Update the value. Requires the old object to be provided.
         /// Update the object and index trees by diff'ing the old and new values.
-        pub fn update(groove: *Groove, new: *const Object, old: *const Object) void {
+        pub fn update(
+            groove: *Groove,
+            values: struct { old: *const Object, new: *const Object },
+        ) void {
+            const old = values.old;
+            const new = values.new;
+
             if (constants.verify) {
                 const old_from_cache = groove.objects_cache.get(@field(old, primary_field)).?;
                 assert(@field(old_from_cache, primary_field) == @field(old, primary_field));
--- a/src/lsm/groove.zig
+++ b/src/lsm/groove.zig
@@ -863,10 +863,12 @@ pub fn GrooveType(
             }
         };
 
-        /// Insert the value into the objects tree and associated index trees, asserting that it doesn't
-        /// already exist.
+        /// Insert the value into the objects tree and associated index trees. It's up to the
+        /// caller to ensure it doesn't already exist.
         pub fn insert(groove: *Groove, object: *const Object) void {
-            assert(!groove.objects_cache.has(@field(object, primary_field)));
+            if (constants.verify) {
+                assert(!groove.objects_cache.has(@field(object, primary_field)));
+            }
 
             groove.objects_cache.upsert(object);
 
@@ -886,37 +888,43 @@ pub fn GrooveType(
             }
         }
 
-        /// Insert the value (or update it, if it exists).
+        /// Update the value. Requires the old object to be provided.
         /// Update the object and index trees by diff'ing the old and new values.
-        pub fn upsert(groove: *Groove, new: *const Object) void {
-            // Explicit stack copy needed - otherwise old will == new!
-            const old: Object = (groove.objects_cache.get(@field(new, primary_field)) orelse {
-                return groove.insert(new);
-            }).*;
+        pub fn update(groove: *Groove, new: *const Object, old: *const Object) void {
+            if (constants.verify) {
+                const old_from_cache = groove.objects_cache.get(@field(old, primary_field)).?;
+                assert(@field(old_from_cache, primary_field) == @field(old, primary_field));
+                assert(old_from_cache.timestamp == old.timestamp);
+            }
 
+            // Sanity check to ensure the caller didn't accidentally pass in an alias.
+            assert(new != old);
+
+            // TODO: This will assert timestamp twice if `has_id` is false.
             assert(@field(old, primary_field) == @field(new, primary_field));
             assert(old.timestamp == new.timestamp);
 
-            groove.objects_cache.upsert(new);
-
             // The ID can't change, so no need to update the ID tree.
-
             // Update the object tree entry if any of the fields (even ignored) are different.
-            if (!std.mem.eql(u8, std.mem.asBytes(&old), std.mem.asBytes(new))) {
+            if (!stdx.equal_bytes(Object, old, new)) {
                 // Unlike the index trees, the new and old values in the object tree share the
                 // same key. Therefore put() is sufficient to overwrite the old value.
                 groove.objects.put(new);
+                groove.objects_cache.upsert(new);
+            } else {
+                // TODO: Should we assert we actually updated something...?
+                return;
             }
 
             inline for (std.meta.fields(IndexTrees)) |field| {
                 const Helper = IndexTreeFieldHelperType(field.name);
-                const old_index = Helper.derive_index(&old);
+                const old_index = Helper.derive_index(old);
                 const new_index = Helper.derive_index(new);
 
                 // Only update the indexes that change.
                 if (!std.meta.eql(old_index, new_index)) {
                     if (old_index) |index| {
-                        const old_index_value = Helper.derive_value(&old, index);
+                        const old_index_value = Helper.derive_value(old, index);
                         @field(groove.indexes, field.name).remove(&old_index_value);
                     }
 
--- a/src/lsm/groove.zig
+++ b/src/lsm/groove.zig
@@ -910,9 +910,7 @@ pub fn GrooveType(
             // Sanity check to ensure the caller didn't accidentally pass in an alias.
             assert(new != old);
 
-            if (has_id) {
-                assert(old.id == new.id);
-            }
+            if (has_id) assert(old.id == new.id);
             assert(old.timestamp == new.timestamp);
 
             // The ID can't change, so no need to update the ID tree. Update the object tree entry
@@ -950,8 +948,8 @@ pub fn GrooveType(
 
             // Putting the objects_cache upsert after the index tree updates is critical:
             // We diff the old and new objects, but the old object will be a pointer into the
-            // objects_cache. If we upsert first, old.* == new.* and no secondary indexes will
-            // be updated!
+            // objects_cache. If we upsert first, there's a high chance old.* == new.* (always,
+            // unless old comes from the stash) and no secondary indexes will be updated!
             groove.objects_cache.upsert(new);
             groove.objects.put(new);
         }
--- a/src/lsm/groove.zig
+++ b/src/lsm/groove.zig
@@ -900,21 +900,27 @@ pub fn GrooveType(
             // Sanity check to ensure the caller didn't accidentally pass in an alias.
             assert(new != old);
 
-            // TODO: This will assert timestamp twice if `has_id` is false.
-            assert(@field(old, primary_field) == @field(new, primary_field));
+            if (has_id) {
+                assert(old.id == new.id);
+            }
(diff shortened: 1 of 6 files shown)


src/lsm/groove.zig
```diff
@@ -886,37 +888,43 @@ pub fn GrooveType(
...
-
             // Update the object tree entry if any of the fields (even ignored) are different.
-            if (!std.mem.eql(u8, std.mem.asBytes(&old), std.mem.asBytes(new))) {
+            if (!stdx.equal_bytes(Object, old, new)) {
                 // Unlike the index trees, the new and old values in the object tree share the
                 // same key. Therefore put() is sufficient to overwrite the old value.
                 groove.objects.put(new);
+                groove.objects_cache.upsert(new);
+            } else {
+                // TODO: Should we assert we actually updated something...?
```
cb22: We never did in the past - is updating an object to an identical state something we would expect is able to happen?

src/stdx.zig
```diff
@@ -520,3 +520,46 @@ pub fn fstatfs(fd: i32, statfs_buf: *StatFs) usize {
...
+
+        .Int => |info| return @sizeOf(T) * 8 == info.bits,
+
+        .Pointer => |info| return info.size != .Slice,
+
+        .Array => |info| return comptime hasUniqueRepresentation(info.child),
+
+        .Struct => |info| {
+            // Packed structs are always unique.
+            if (info.backing_integer != null) {
```
cb22: If a struct is packed, we don't need to recursively check it like the unpacked case below, since Zig ensures packed structs can only contain other packed structsa.

src/state_machine.zig
```diff
@@ -1006,8 +1006,8 @@ pub fn StateMachineType(
                 cr_account_new.credits_posted += amount;
             }
 
-            self.forest.grooves.accounts.upsert(&dr_account_new);
-            self.forest.grooves.accounts.upsert(&cr_account_new);
+            self.forest.grooves.accounts.update(&dr_account_new, dr_account);
```
jorangreef:
```suggestion
            self.forest.grooves.accounts.update(dr_account, &dr_account_new);
```

So that progress moves from left to right, from old to new.

src/lsm/groove.zig
jorangreef: Great question. Thanks for flagging this.

If the state machine calls `update()` then it should be unilateral (and we assert under `verify` that there is a change). Rather than the Groove doing magic to compare, to double check that the update is necessary, the Groove should trust the state machine to already know that the update is necessary.

If we later have state machines that might call `update()` but not know if the update is necessary, then we can add an extra `update_if_different()`.

jorangreef: This hoists the comparison into the explicit policy/responsibility of the state machine, and then if it later needs a method to cede this responsibility to the groove, then that is also explicit.

src/state_machine.zig
jorangreef: A risk to my suggestion here though is that if an `insert(new)` is swapped to an `update(old, new)` then the caller needs to swap arguments.

However, if the interface remains `update(new, old)` then there is the risk that the caller does `(old, new)`... thinking progress moves positively, from left to right.

cb22: Yeah indeed - my ordering was based off the first case case, which also imitates our `target` first style for copies... A third option could be to make it an explicit struct? 

```zig
self.forest.grooves.accounts.update(.{ .old = dr_account, .new = &dr_account_new });
```

From what I can see, we don't have any other `update` functions floating around to base the style off.

> Two parameters of the same type here makes me nervous 😅

~ DJ, 1 year ago, on my AOF PR :smile:

jorangreef: Yes, for copies, we should also rather be moving from left to right, from `source` to `target` (I know the Zig std lib does it the other way, but we can break rhyme).

batiati:
Can we add `hasUniqueRepresentation` to `tidy.banned(...)`?

src/state_machine.zig
cb22: After a bit more discussion:
- `target` first style for copies gets `const` protection, which we can't take advantage of here.
- There's no real convention, because this `.update()` method doesn't update a parameter with an existing one, but rather updates the groove based off of both.
- There's no performance cost to using a struct (asm generated on godbolt was identical and double checked on the benchmark).

I've gone with a struct.

src/stdx.zig
```diff
@@ -520,3 +520,46 @@ pub fn fstatfs(fd: i32, statfs_buf: *StatFs) usize {
         @intFromPtr(statfs_buf),
     );
 }
+
+// TODO(Zig): https://github.com/ziglang/zig/issues/17592.
+/// True if every value of the type `T` has a unique bit pattern representing it.
+/// In other words, `T` has no unused bits and no padding.
+pub fn hasUniqueRepresentation(comptime T: type) bool {
```
batiati: When vendoring, we've been adopting our snake_case convention.
Arguable here, since we should move back to std as soon we update Zig again 🤔.

```suggestion
pub fn has_unique_representation(comptime T: type) bool {
```

src/stdx.zig
matklad: Let's add a couple of tests here though, just to sanity-check things.

matklad: Also, for a good measure, let's assert that the bitsize of all fields adds up to the sizeof the whole struct (that is, that we don't have padding _bits_).

I am thinking about cases like 

```zig
const S = packed struct {
    x: u24,
    y: u3,
};
```

I think the language may work both as "the spare bits after `y` are zeroed" and "the spare bits after `y` are arbitrary". I _think_ they are actually zeroed, but let's assert that we just don't have such weird structs altogether.

src/lsm/groove.zig
```diff
@@ -886,37 +888,55 @@ pub fn GrooveType(
...
+            groove.objects_cache.upsert(new);
+
             inline for (std.meta.fields(IndexTrees)) |field| {
                 const Helper = IndexTreeFieldHelperType(field.name);
-                const old_index = Helper.derive_index(&old);
+                const old_index = Helper.derive_index(old);
                 const new_index = Helper.derive_index(new);
 
                 // Only update the indexes that change.
                 if (!std.meta.eql(old_index, new_index)) {
```
matklad, line 936: I wonder if comparing just `values` (and ignoring timestamps) would make sense here, benchmark-wise?.

src/lsm/groove.zig
```diff
@@ -886,37 +888,55 @@ pub fn GrooveType(
...
+                const tombstone = ObjectTreeHelpers(Object).tombstone;
+                const key_from_value = ObjectTreeHelpers(Object).key_from_value;
+
+                assert(!stdx.equal_bytes(Object, old, new));
+                assert(key_from_value(old) == key_from_value(new));
+                assert(!tombstone(old) and !tombstone(new));
             }
 
+            groove.objects.put(new);
+            groove.objects_cache.upsert(new);
```
matklad: This makes my [my Rust eye](https://static.wikia.nocookie.net/p__/images/7/73/Mad-Eye_Moody%27s_%22mad_eye%22.jpeg/revision/latest?cb=20160215100736&path-prefix=protagonist) twitch a bit. Isn't this shared mutable aliasing? I _think_ the `old` object might be the object from the cache (`maybe(old == old_from_cache)` in the if above). If that's the case, I think we _might_ be clobbering old here, in which case the `inline for` below would get wrong results when accessing `old` fields?

cb22: We have an explicit guard against this above; since I've been bitten by this before here :smile::

> assert(new != old);

I think that's enough to prevent it from happening?

src/lsm/groove.zig
cb22: `derive_index` does return `IndexType(field_name)` which boils down to just the `value`. It's only `derive_value` that creates the `CompositeKey` with both. `std.meta.eql` should boil away to `==` in these cases too.

src/lsm/groove.zig
matklad: No, I think that's different. The `new` is an object on state_machine stack:

https://github.com/tigerbeetle/tigerbeetle/blob/c367d1dfdba783b8eb171cdfec5d9de0188517c5/src/state_machine.zig#L870

the `old` is an object from the cache:

https://github.com/tigerbeetle/tigerbeetle/blob/c367d1dfdba783b8eb171cdfec5d9de0188517c5/src/state_machine.zig#L813

So `new` and `old`  are different. 

However, when we do `groove.objects_cache.upsert(new);` this might invalidate `old` still, if it was in the cache. That is, after this line I would expect `new.*` and `old.*` to be equal, because the `old` memory was overwritten when we upserted `new`.

Not 100% sure that this could happen, but I think it does?

cb22: Ahhh yes, that is the case. Good catch!

In the past, I took an explicit stack copy of `old` within the function which prevented that. We could move the object cache upsert to the end - there's no reason it needs to be done before the index trees - with an explicit comment explaining the above. That way we can get correct behaviour without a stack copy, but it still feels a bit fragile...

matklad: I think that's ok, as long as we document that with a `maybe`. 

I wonder if we even can _require_ that this is the object from the cache? It _must_ be, if we are updating, right?

matklad: I am also _reeeeeally_ wondering why CI is green here :thinking: in `release`, I can see compiler hoisting the loads to before the cache update, which would give the correct behavior. 

But in debug, I expect the overwrite to be reflected robustly. Could you also take a look what's up with tests passing here? Either my theory about the overwrite is wrong (good case), or we somehow don't test this case at all (which would be pretty scary).

src/lsm/groove.zig
```diff
@@ -886,37 +888,55 @@ pub fn GrooveType(
...
+        ) void {
+            const old = values.old;
+            const new = values.new;
 
-            assert(@field(old, primary_field) == @field(new, primary_field));
-            assert(old.timestamp == new.timestamp);
+            if (constants.verify) {
+                const old_from_cache = groove.objects_cache.get(@field(old, primary_field)).?;
+                assert(@field(old_from_cache, primary_field) == @field(old, primary_field));
+                assert(old_from_cache.timestamp == old.timestamp);
```
sentientwaffle: Can you assert that `stdx.bytes_equal(Object, old, old_from_cache);`?

src/lsm/groove.zig
cb22: > Could you also take a look what's up with tests passing here? Either my theory about the overwrite is wrong (good case), or we somehow don't test this case at all (which would be pretty scary).

As far as I can tell, we don't have any tests (unit or fuzz) that actually check the result of the secondary index trees, and in this case, if `old.* == new.*` it'll just skip updating the secondary index so there's nothing that's hit on the way down :grimacing:...

Also just noticed the forest fuzz (which I forgot to run / update in this PR) has also been broken for a few weeks. I need to look at proper infra for fuzzing ASAP, but maybe as a workaround for now we can just run each fuzz for a fixed time on CI?

We previously did run some more fuzzes in CI, but with a fixed seed for a single iteration.

cb22: > I wonder if we even can require that this is the object from the cache? It must be, if we are updating, right?

It could in theory be a stack copy, but we don't have an explicit use case for that yet so I've made the check `assert(old_from_cache == old);`.

src/lsm/groove.zig
cb22: I'm now aserting they're actually one and the same: https://github.com/tigerbeetle/tigerbeetle/pull/1239#discussion_r1369774962

src/stdx.zig
cb22: >  Arguable here, since we should move back to std as soon we update Zig again 🤔.

It's a small enough change either way, so I'm happy to adopt it.

src/lsm/groove.zig
```diff
@@ -946,6 +947,13 @@ pub fn GrooveType(
                     }
                 }
             }
+
+            // Putting the objects_cache upsert after the index tree updates is critical:
+            // We diff the old and new objects, but the old object will be a pointer into the
+            // objects_cache. If we upsert first, old.* == new.* and no secondary indexes will
+            // be updated!
+            groove.objects_cache.upsert(new);
+            groove.objects.put(new);
```
matklad, line 954: For an extra tripple check, can we assert here that now old & new _contents_ is the same?

src/stdx.zig
```diff
@@ -540,26 +540,133 @@ pub fn hasUniqueRepresentation(comptime T: type) bool {
...
-        .Array => |info| return comptime hasUniqueRepresentation(info.child),
+        .Array => |info| return comptime has_unique_representation(info.child),
 
         .Struct => |info| {
-            // Packed structs are always unique.
-            if (info.backing_integer != null) {
-                return true;
+            // Only consider packed structs unique if they are byte aligned.
+            if (info.backing_integer) |backing_integer| {
+                return @sizeOf(@Type(.{ .Struct = info })) * 8 == @bitSizeOf(backing_integer);
```
matklad: `@sizeOf(T)` would be simpler I think?

cb22: ... yes, I haven't had my Monster yet. :rofl:

jorangreef: I think it's time to go pull an espresso too... ☕

src/lsm/groove.zig
cb22: Ok - so that won't actually hold; if the old value comes from the stash of the cache_map, and we insert a new value, it'll go into the cache - so now `old.* != new.*`. 

The assertion at the start, `assert(old_from_cache == old);` will still hold because if it comes from the stash from the caller, it should still come from the stash inside.

src/lsm/groove.zig
```diff
@@ -910,9 +910,7 @@ pub fn GrooveType(
             // Sanity check to ensure the caller didn't accidentally pass in an alias.
             assert(new != old);
 
-            if (has_id) {
-                assert(old.id == new.id);
```
matklad: :+1: I actually do line one-line without braces form quite a bit more, when it works

matklad (approved):

src/lsm/groove.zig
jorangreef: Yes, we don't need to use kernel style indented returns etc.

src/lsm/groove.zig
sentientwaffle: > Also just noticed the forest fuzz (which I forgot to run / update in this PR) has also been broken for a few weeks.

I think that is fixed as part of https://github.com/tigerbeetle/tigerbeetle/pull/1234, which merged this morning.

PR #1240, opened 2023-10-23
LSM: Approximate pacing

cb22:
Until compaction IO pacing lands, we'll show higher p100 latencies than we should, since we're not properly distributing work across the bar.

In the meantime, bump up our block size to 1MB and decrease `lsm_batch_multiple` to 32. This has the effect of approximating pacing; since {1-3} blocks are done per beat regardless. The end result is a nice decrease in p100 latencies, at least until a few levels have filled up. The only remaining spikes are due to CPU compaction and checkpoint work.

## Grid cache impact
With the new, (very) large block size, we also need to make a compromise on our Grid cache - otherwise with the current minimum size of 2048 objects, it would require at least 2GB of RAM for the Grid cache alone. 

Lower the `cache_line_size` to a sub-optimal 16; it has less impact on performance (~2%)  than lowering the number of ways.

# Benchmarks
```
# On NVMe
./scripts/benchmark.sh --print-batch-timings --transfer-count=10_000_000 --account-count=10 --id-order=sequential
```

## Before
```
1266 batches in 48.07 s
load offered = 1000000 tx/s
load accepted = 208011 tx/s
batch latency p00 = 0 ms
batch latency p10 = 5 ms
batch latency p20 = 5 ms
batch latency p30 = 6 ms
batch latency p40 = 7 ms
batch latency p50 = 7 ms
batch latency p60 = 8 ms
batch latency p70 = 9 ms
batch latency p80 = 10 ms
batch latency p90 = 12 ms
batch latency p100 = 2954 ms
transfer latency p00 = 0 ms
transfer latency p10 = 1169 ms
transfer latency p20 = 4141 ms
transfer latency p30 = 7364 ms
transfer latency p40 = 11636 ms
transfer latency p50 = 16378 ms
transfer latency p60 = 21089 ms
transfer latency p70 = 26211 ms
transfer latency p80 = 31996 ms
transfer latency p90 = 33857 ms
transfer latency p100 = 38091 ms
```

## After
```
1243 batches in 41.73 s
load offered = 1000000 tx/s
load accepted = 239645 tx/s
batch latency p00 = 1 ms
batch latency p10 = 5 ms
batch latency p20 = 6 ms
batch latency p30 = 8 ms
batch latency p40 = 13 ms
batch latency p50 = 15 ms
batch latency p60 = 17 ms
batch latency p70 = 19 ms
batch latency p80 = 21 ms
batch latency p90 = 25 ms
batch latency p100 = 829 ms
transfer latency p00 = 1 ms
transfer latency p10 = 1762 ms
transfer latency p20 = 4226 ms
transfer latency p30 = 7265 ms
transfer latency p40 = 10305 ms
transfer latency p50 = 12667 ms
transfer latency p60 = 16133 ms
transfer latency p70 = 20150 ms
transfer latency p80 = 24443 ms
transfer latency p90 = 28760 ms
transfer latency p100 = 31738 ms
```

### DIFF
--- a/.github/ci/test_aof.sh
+++ b/.github/ci/test_aof.sh
@@ -36,7 +36,7 @@ rm -f aof-test.tigerbeetle
 rm -f aof-test.tigerbeetle.aof
 
 ./tigerbeetle-aof format --cluster=0 --replica=0 --replica-count=1 aof-test.tigerbeetle > aof.log 2>&1
-./tigerbeetle-aof start --cache-grid=256MB --addresses=3001 aof-test.tigerbeetle >> aof.log 2>&1 &
+./tigerbeetle-aof start --cache-grid=512MB --addresses=3001 aof-test.tigerbeetle >> aof.log 2>&1 &
 
 # Wait for replicas to start, listen and connect:
 sleep 1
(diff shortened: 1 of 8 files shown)


jorangreef (approved):
Looks good!

Related, it was interesting to see how many sites we had to update, to bump `--cache-grid` to `512MB` (and we should in fact be using `MiB` which will also parse), which is a magic number.

In a follow-up, we should rather set `--cache-grid` implicitly (if not provided by the operator explicitly) to a minimum default informed directly by the various constants—rather than hard coding the implications of the constant here. Alternatively, to guide users to thinking about and setting the cache, we could have this as an example in one place (QuickStart) only.

PR #1243, opened 2023-10-24
LSM: Switch forest fuzz to stdx.equal_bytes

cb22:
`std.mem.eql` and `asBytes` hid the fact that we were now incorrectly comparing a double pointer to a single pointer.

Change to `stdx.equal_bytes` for some nice type safety, and fix the bug.

### DIFF
--- a/src/lsm/forest_fuzz.zig
+++ b/src/lsm/forest_fuzz.zig
@@ -4,6 +4,7 @@ const assert = std.debug.assert;
 
 const constants = @import("../constants.zig");
 const fuzz = @import("../testing/fuzz.zig");
+const stdx = @import("../stdx.zig");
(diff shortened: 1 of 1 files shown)


cb22:
(something funky was up with my build, just double checking)

Ah - #1239 ~is~ was still in the merge queue.

matklad:
Reminds me that I wanted to vendor better version of `asBytes` / `bytesAsSlice` which:

* would take an explicit `T` argument to prevent two-star bugs
* would check that the alignment is correct

This was blocked by https://github.com/ziglang/zig/pull/16245 before, but I think this now is in 0.11? Anway, that's future work!

matklad (approved):

PR #1244, opened 2023-10-24
build: specify CPU features once

matklad:


matklad:
The main philosophical thing here is that the `-Dcpu` is no longer user controlled -- you can use any CPU, as long as that's what we explicitly support. 

I think that's the right call for tigerbeetle --- we don't intend to be as portable as possible, rather, we aim to be as releable as fast as feasible on the specific architectures we support.

build.zig
```diff
@@ -6,11 +6,50 @@ const Mode = std.builtin.Mode;
...
+        "x86_64-windows",
+    };
+    const cpus = .{
+        "baseline+aes+neon",
+        "baseline+aes+neon",
+        "x86_64_v3+aes",
+        "x86_64_v3+aes",
+        "x86_64_v3+aes",
+    };
+    for (tripples, cpus) |tripple, cpu| {
```
jorangreef:
```suggestion
    for (tuples, cpus) |tuple, cpu| {
```

Or `triples|triple`.

jorangreef: Since `aarch64-linux` is more two parts, and also since `tuple` is more general: "a data structure consisting of multiple parts" so it doesn't matter if there are two or three.

jorangreef (commented):
If I understand correctly, this adds an explicit allowlist of targets and CPU "features", and then selects whichever of these the build machine supports?

build.zig
matklad: "triple" is a term of art here: https://www.flother.is/til/llvm-target-triple/

It can contain anywhere from two to five components, but definitely only a single "p"!

matklad:
>If I understand correctly, this adds an explicit allowlist of targets and CPU "features", and then selects whichever of these the build machine supports?

From the perspective of the API, yes, the user can not longer compile TigerBeetle for "whatever CPU" (unless they want to patch build.zig), they have to pick from the available menu. On the other hand, only the CPU features which we actually test get enabled. (actually,  as we were passing `-Dcpu` only for releases and benchmarks, our tests themselves were testing the "wrong" CPU (IIRC, zig uses 'native' by default)). 

From the perspective of the implementation, this is actually a bit more hacky. There _is_ `whitelist` argument for `standardTargetOptions`, but I don't think it does what I want here (it only [kicks in during explicit cross-compilation](https://github.com/ziglang/zig/blob/d1230842ac882bbd6456081c474658b293aaf20a/lib/std/Build.zig#L1215-L1217)). I _could_ write our own, non-standard target options, but I don't want to do that just now (as that would be better as a part of larger build.zig cleanup). So instead I just overwrite whatever `-Dcpu` user may have passed in with the value we actually support --- my primary goal here is not to cleanup build.zig interface, but to get rid of fundamental duplication of cpu knowledge between our release process, benchmark, and docker file (and, as I noticed while writing this comment, our tests, which _should_ have set the CPU, but didnt)

build.zig
jorangreef: Ah, thanks for the "triple" pointer!

cb22: They missed the chance to call it a trippple. :disappointed:

build.zig
```diff
@@ -6,11 +6,50 @@ const Mode = std.builtin.Mode;
...
+    var target = b.standardTargetOptions(.{});
+    // Patch the target to use the right CPU. This is a somewhat hacky way to do this, but the core
+    // idea here is to keep this file as the source of truth for what we need from the CPU.
+    for (supported_targets) |supported_target| {
+        if (target.getCpuArch() == supported_target.getCpuArch()) {
+            target.cpu_model = supported_target.cpu_model;
+            target.cpu_features_add = supported_target.cpu_features_add;
+            target.cpu_features_sub = supported_target.cpu_features_sub;
+        }
+    }
```
sentientwaffle: Should we fail on unsupported targets? e.g.
```suggestion
        if (target.getCpuArch() == supported_target.getCpuArch()) {
            target.cpu_model = supported_target.cpu_model;
            target.cpu_features_add = supported_target.cpu_features_add;
            target.cpu_features_sub = supported_target.cpu_features_sub;
            break;
        }
    } else @panic("error: unsupported target");
```

build.zig
```diff
@@ -6,11 +6,50 @@ const Mode = std.builtin.Mode;
...
+        "x86_64-macos",
+        "x86_64-windows",
+    };
+    const cpus = .{
+        "baseline+aes+neon",
+        "baseline+aes+neon",
+        "x86_64_v3+aes",
+        "x86_64_v3+aes",
+        "x86_64_v3+aes",
+    };
```
sentientwaffle, line 28: An array of `CrossTarget.ParseOptions` (or just tuples) might be more readable, though it probably doesn't matter with only 5 pairs.

matklad: I've actually recently started to use "columnar" format for small tables in the source code to _improve_ readability, as it requires less symbols and usually formats better, so this is intentional

build.zig
matklad: Good call!

build.zig
sentientwaffle: (Related: https://github.com/ziglang/zig/issues/17145)

sentientwaffle (approved):

PR #1245, opened 2023-10-24
CI: restore publishing of docker images

matklad:
- move publishing to Zig
- put the exact release binary into container, instead of building
  inside docker


tools/docker/Dockerfile
```diff
@@ -1,33 +1,15 @@
...
-RUN ./scripts/install_zig.sh
-
-COPY .git .git
-COPY scripts ./scripts
-COPY src ./src
-COPY build.zig ./build.zig
-
 ARG TARGETPLATFORM
-RUN test ${TARGETPLATFORM} = "linux/arm64" && CPU="baseline+aes+neon" || CPU="x86_64_v3+aes"; CI=true ./scripts/install.sh
+RUN unzip tigerbeetle-$(test ${TARGETPLATFORM} = "linux/arm64" && echo "aarch64" || echo "x86_64")-linux$(test ${DEBUG} = "true" && echo "-debug" || echo "").zip
```
matklad: Open to suggestions how to make this more readable!

cb22: I think you can replace the whole Dockerfile with:

```dockerfile
FROM alpine

ARG TARGETARCH
COPY dist/docker/tigerbeetle-${TARGETARCH} /tigerbeetle

ENTRYPOINT ["/tigerbeetle"]
```

If in the `publish_docker` (or a `build_docker`) you unzip the desired eg aarch64 tigerbeetle to `dist/docker/tigerbeetle-arm64`.

(even if you don't want to go down that route entirely, it would still be good to eliminate the multi-stage build since it's no longer needed!)

cb22: `FROM alpine`

Bit more context on that - the only reason we're not `FROM scratch` is so that if you're debugging or poking around you have a somewhat familiar environment to do that in.

src/scripts/release.zig
```diff
@@ -615,6 +616,52 @@ fn publish_node(shell: *Shell, info: VersionInfo) !void {
...
+    try shell.project_root.setAsCwd();
+    assert(try shell.dir_exists("dist/tigerbeetle"));
+
+    try shell.exec(
+        \\docker login --username tigerbeetle --password {password} ghcr.io
+    , .{
+        .password = try shell.env_get("GITHUB_TOKEN"),
+    });
+
+    try shell.exec(
```
cb22: If you want to be able to test these before pushing, you can do the following:

```
# Build the Dockerfile to an OCI tarball.
docker buildx build --file tools/docker/Dockerfile . --platform linux/amd64,linux/arm64 --output type=oci,dest=dist/docker/oci.tar

# Load each platform separately, so we can test it locally.
docker import dist/docker/oci.tar --platform arm64 tigerbeetle-arm64
docker import dist/docker/oci.tar --platform amd64 tigerbeetle-amd64
docker run tigerbeetle-arm64 version --verbose
docker run tigerbeetle-amd64 version --verbose

# Build and push - this will come straight from build cache.
docker buildx build --file tools/docker/Dockerfile . --platform linux/amd64,linux/arm64 --build-arg=DEBUG=false --tag ghcr.io/tigerbeetle/tigerbeetle:{version} --push
```

src/scripts/release.zig
```diff
@@ -615,6 +616,52 @@ fn publish_node(shell: *Shell, info: VersionInfo) !void {
     });
 }
 
+fn publish_docker(shell: *Shell, info: VersionInfo) !void {
```
cb22, line 621: I know you said your reasoning for not having a `build_docker` step - but there are still things that can go wrong in the `Dockerfile` build even though we aren't really doing much besides copying a binary...

tools/docker/Dockerfile
cb22: Ah, either if doing my suggestion below or with the current setup, I don't think `tigerbeetle` will be `chmod +x`'d anywhere (IIRC zip doesn't keep permission...)

matklad: I think it gets the `x` bit actually, at least locally `docker run` seems to work sometimes (it seems podman picks a random arch from a multiarch builds and sometises you get lucky)

matklad: @cb22 

>I think you can replace the whole Dockerfile with:

This still needs `ARG DEBUG`, no?

cb22: > This still needs ARG DEBUG, no?

We build our debug image with a separate invocation of `docker build`, so we can just change the binaries out and specify a different tag!

We could also make it a build arg too if we want to make it more explicit...

matklad:
Manually smoke tested that 

```zig
fn publish_docker(shell: *Shell, info: VersionInfo) !void {
    var section = try shell.open_section("publish docker");
    defer section.close();

    try shell.project_root.setAsCwd();
    assert(try shell.dir_exists("dist/tigerbeetle"));

    // try shell.exec(
    //     \\docker login --username tigerbeetle --password {password} ghcr.io
    // , .{
    //     .password = try shell.env_get("GITHUB_TOKEN"),
    // });

    for (@as([]const bool, &.{ true, false })) |debug| {
        const triples: []const []const u8 = &.{ "aarch64-linux", "x86_64-linux" };
        const docker_arches: []const []const u8 = &.{ "arm64", "amd64" };
        for (triples, docker_arches) |triple, docker_arch| {
            // We need to unzip binaries from dist. For simplicity, don't bother with a temporary
            // directory.
            shell.project_root.deleteFile("tigerbeetle") catch {};
            try shell.exec("unzip ./dist/tigerbeetle/tigerbeetle-{triple}{debug}.zip", .{
                .triple = triple,
                .debug = if (debug) "-debug" else "",
            });
            try shell.project_root.rename(
                "tigerbeetle",
                try shell.print("tigerbeetle-{s}", .{docker_arch}),
            );
        }
        try shell.exec(
            \\docker buildx build --file tools/docker/Dockerfile . --platform linux/amd64
            \\   --tag dockerlol --load
        , .{  });

        // Sadly, there isn't an easy way to locally build & test a multiplatform image without
        // pushing it out to the registry first. As docker testing isn't covered under not rocket
        // science rule, let's do a best effort after-the-fact testing here.
        const version_verbose = try shell.exec_stdout(
            \\docker run dockerlol version --verbose
        , .{
            // .version = info.version,
        });
        const mode = if (debug) "Debug" else "ReleaseSafe";
        assert(std.mem.indexOf(u8, version_verbose, mode) != null);
        assert(std.mem.indexOf(u8, version_verbose, info.version) != null);
    }
}
```

works locally, so we don't need `chmod +x` it seems.

src/scripts/release.zig
```diff
@@ -615,6 +618,54 @@ fn publish_node(shell: *Shell, info: VersionInfo) !void {
...
+    defer section.close();
+
+    try shell.project_root.setAsCwd();
+    assert(try shell.dir_exists("dist/tigerbeetle"));
+
+    try shell.exec(
+        \\docker login --username tigerbeetle --password {password} ghcr.io
+    , .{
+        .password = try shell.env_get("GITHUB_TOKEN"),
+    });
```
matklad, line 632: @cb22 just to highlight this login part, as I have never run this thing even once and don't know what it should look like.

cb22: It seems good to me! Technically any process on the system could read `password` via `ps` or so, but on a dedicated runner where we would have bigger problems if compromised, I think that's OK.

The other way is to use `--password-stdin` and pipe it in.

tools/docker/Dockerfile
```diff
@@ -1,37 +1,11 @@
...
-FROM alpine:3.17 AS release
-WORKDIR /
-
-COPY --from=build /opt/beta-beetle/tigerbeetle ./tigerbeetle
+# We _could_ use `FROM scratch` here, but let's use alpine to ease debugging in case someone needs
+# to poke inside the container.
+FROM alpine:3.17
+
+# NB: We are doing something unconventional here. To ensure that docker images contain the exact
+# same TigreBeetle binary we publish as our release, we don't build inside a container, and copy a
```
cb22:
```suggestion
# same TigerBeetle binary we publish as our release, we don't build inside a container, and copy a
```

cb22 (approved):

PR #1250, opened 2023-10-25
ci: fix docker publish

matklad:
- include `-debug` in the image tag if appropriate
- don't fail due to `0.14.155: Pulling from tigerbeetle/tigerbeetle` output on stderr from Docker over here:

  https://github.com/tigerbeetle/tigerbeetle/blob/468007a0a37c3646f2bb6ace5cabe27ed7eba7b8/src/scripts/release.zig#L662-L666

### DIFF
--- a/src/shell.zig
+++ b/src/shell.zig
@@ -283,7 +283,7 @@ pub fn exec(shell: Shell, comptime cmd: []const u8, cmd_args: anytype) !void {
     child.stdout_behavior = .Inherit;
     child.stderr_behavior = .Inherit;
 
-    echo_command(&child);
+    echo_command(argv.slice());
     const term = try child.spawnAndWait();
 
     switch (term) {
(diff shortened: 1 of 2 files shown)


src/shell.zig
```diff
@@ -326,35 +326,27 @@ pub fn exec_stdout(shell: *Shell, comptime cmd: []const u8, cmd_args: anytype) !
...
-        },
-        else => return error.CommandFailed,
+    defer shell.arena.allocator().free(child.stderr);
+    errdefer {
+        log.err("command failed", .{});
+        echo_command(argv.slice());
+        log.err("stdout:\n{s}\nstderr:\n{s}", .{ child.stdout, child.stderr });
     }
 
-    if (res.stderr.len > 0) return error.NonEmptyStderr;
```
matklad, line 351: Just dropped this bit

cb22 (approved):

PR #1251, opened 2023-10-25
LSM/ManifestLog: Manifest block count upper-bound

sentientwaffle:
## Summary

(Borrowed from docs):

> The goals of manifest log compaction are (in no particular order):
>
> 1. Free enough manifest blocks such that there are always enough free slots in the manifest
>    trailer to accommodate the appends by table compaction.
> 2. Shrink the manifest: A smaller manifest means that fewer blocks need to be replayed during
>    recovery, or repaired during state sync.
> 3. Don't shrink the manifest too much: The more manifest compaction work is deferred, the more
>    "efficient" compaction is. Put another way: deferring manifest compaction means that more
>    entries are freed per block compacted.
> 4. Spread compaction work evenly between half-bars, to avoid latency spikes.
>
> To address goal 1, we must (on average) "remove" as many blocks from the manifest log as we add.
> But when we compact a block, only a subset of its entries can be freed/dropped – the remainder
> must be re-appended to the manifest log.
>
> The upper-bound number of manifest blocks is related to the rate at which we compact blocks.
> Put simply, the more compaction work we do, the smaller the upper bound.

Prior to this PR:
1. The maximum number of manifest blocks has been hard-coded at 65536.
2. We compact exactly one manifest block per half-bar.

But the current state of things is not ideal. In particular:
1. We had no guarantee that manifest compaction was doing enough work to avoid eventually running out of space. (It wasn't!)
2. One block per half-bar is more than enough when the forest is small, but as trees fill up their lower levels, it falls behind.

This commit solves both issues – the upper bound and compaction pacing – by leveraging the relationship between the two factors.

More specifically:
- the maximum compaction rate is now defined in terms of the number of blocks appended per half-bar
- the manifest block count upper bound is defined in terms of the compaction rate

For details, check out the (very long) explanation in `manifest_log.zig`. 

## Compaction vs Upper Bound

This chart plots the manifest upper-bound block count (Y axis) against the number of _extra_ blocks compacted per half-bar (i.e. the number of blocks compacted beyond the absolute minimum required for an upper bound to exist at all).

Caveats:
- This chart assumes a 128 KiB block size, not the 1 MiB block size that is currently in `main`.
- In order to keep the calculation (relatively) simple, several factors are ignored, resulting in the upper bound being overestimated. See `manifest_log.zig` for details.

Screenshot 2023-10-25 at 09-57-19 Scratch

<details>
<summary>Data</summary>

`lsm_manifest_compact_extra_blocks` | `log_blocks_max`
--: | --:
1 | 9439
2 | 6163
3 | 4972
4 | 4330
5 | 3928
6 | 3646
7 | 3436
8 | 3277
9 | 3148
10 | 3043
11 | 2956
12 | 2884
13 | 2821
14 | 2764
15 | 2719
16 | 2677
17 | 2638
18 | 2605
19 | 2575
20 | 2548
21 | 2521
22 | 2500
23 | 2479
24 | 2458
25 | 2440
26 | 2425
27 | 2410
28 | 2395
29 | 2383

</details>

### Default

I selected `1` as the default production `constants.lsm_manifest_compact_extra_blocks`.

My reasoning is that optimizing throughput is a higher priority than optimizing time-to-recover.

### DIFF
--- a/src/lsm/manifest_log.zig
+++ b/src/lsm/manifest_log.zig
@@ -83,8 +83,12 @@ pub fn ManifestLogType(comptime Storage: type) type {
         superblock: *SuperBlock,
         grid: *Grid,
         options: Options,
+        pace: Pace,
 
         grid_reservation: ?Grid.Reservation = null,
+
+        /// The number of blocks (remaining) to compact during the current half-bar.
+        compact_blocks: ?u32 = null,
 
         /// This is a struct-of-arrays of `BlockReference`s.
         /// It includes:
@@ -149,20 +153,41 @@ pub fn ManifestLogType(comptime Storage: type) type {
         pub fn init(allocator: mem.Allocator, grid: *Grid, options: Options) !ManifestLog {
             assert(options.tree_id_min <= options.tree_id_max);
 
-            // TODO Replace with the actual computed upper-bound.
-            const log_blocks_max = 65536;
+            const pace = Pace.init(.{
+                .tree_count = options.forest_tree_count(),
+                .tables_max = options.forest_table_count_max,
+                .compact_extra_blocks = constants.lsm_manifest_compact_extra_blocks,
+            });
+
+            inline for (std.meta.fields(Pace)) |pace_field| {
+                log.debug("{?}: Manifest.Pace.{s} = {d}", .{
+                    grid.superblock.replica_index,
+                    pace_field.name,
+                    @field(pace, pace_field.name),
+                });
+            }
 
             var log_block_checksums =
-                try RingBuffer(u128, .slice).init(allocator, log_blocks_max);
+                try RingBuffer(u128, .slice).init(allocator, pace.log_blocks_max);
             errdefer log_block_checksums.deinit(allocator);
 
             var log_block_addresses =
-                try RingBuffer(u64, .slice).init(allocator, log_blocks_max);
+                try RingBuffer(u64, .slice).init(allocator, pace.log_blocks_max);
             errdefer log_block_addresses.deinit(allocator);
+
+            // The upper-bound of manifest blocks we must buffer.
+            //
+            // `blocks` must have sufficient capacity for:
+            // - a leftover open block from the previous ops (+1 block)
+            // - table updates copied from a half bar of manifest compactions
+            // - table updates from a half bar of table compactions
+            const half_bar_buffer_blocks_max =
+                1 + pace.half_bar_compact_blocks_max + pace.half_bar_append_blocks_max;
+            assert(half_bar_buffer_blocks_max >= 3);
 
             // TODO RingBuffer for .slice should be extended to take care of alignment:
             var blocks =
-                try RingBuffer(BlockPtr, .slice).init(allocator, options.blocks_count_max());
+                try RingBuffer(BlockPtr, .slice).init(allocator, half_bar_buffer_blocks_max);
             errdefer blocks.deinit(allocator);
 
             for (blocks.buffer, 0..) |*block, i| {
@@ -171,7 +196,7 @@ pub fn ManifestLogType(comptime Storage: type) type {
             }
             errdefer for (blocks.buffer) |b| allocator.free(b);
 
-            var writes = try allocator.alloc(Write, options.blocks_count_max());
+            var writes = try allocator.alloc(Write, half_bar_buffer_blocks_max);
             errdefer allocator.free(writes);
             @memset(writes, undefined);
 
@@ -187,6 +212,7 @@ pub fn ManifestLogType(comptime Storage: type) type {
                 .superblock = grid.superblock,
                 .grid = grid,
                 .options = options,
+                .pace = pace,
                 .log_block_checksums = log_block_checksums,
                 .log_block_addresses = log_block_addresses,
                 .blocks = blocks,
@@ -220,6 +246,7 @@ pub fn ManifestLogType(comptime Storage: type) type {
                 .superblock = manifest_log.superblock,
                 .grid = manifest_log.grid,
                 .options = manifest_log.options,
+                .pace = manifest_log.pace,
                 .log_block_checksums = manifest_log.log_block_checksums,
                 .log_block_addresses = manifest_log.log_block_addresses,
                 .blocks = .{ .buffer = manifest_log.blocks.buffer },
@@ -630,6 +657,10 @@ pub fn ManifestLogType(comptime Storage: type) type {
         }
 
         /// `compact` does not close a partial block; that is only necessary during `checkpoint`.
+        ///
+        /// The (production) block size is large, so the number of blocks compacted per half-bar is
+        /// relatively small (e.g. ~4). We read them in sequence rather than parallel to spread the
+        /// work more evenly across the half-bar's beats.
         // TODO Make sure block reservation cannot fail — before compaction begins verify that
         // enough free blocks are available for all reservations.
         pub fn compact(manifest_log: *ManifestLog, callback: Callback, op: u64) void {
@@ -641,6 +672,7 @@ pub fn ManifestLogType(comptime Storage: type) type {
             assert(manifest_log.grid_reservation == null);
             assert(manifest_log.blocks.count ==
                 manifest_log.blocks_closed + @intFromBool(manifest_log.entry_count > 0));
+            assert(manifest_log.compact_blocks == null);
             assert(op % @divExact(constants.lsm_batch_multiple, 2) == 0);
 
             if (op < constants.lsm_batch_multiple or
@@ -651,12 +683,22 @@ pub fn ManifestLogType(comptime Storage: type) type {
                 return;
             }
 
-            // +1: During manifest-log compaction, we will create at most one block.
-            manifest_log.grid_reservation =
-                manifest_log.grid.reserve(1 + manifest_log.options.blocks_count_appends()).?;
+            manifest_log.compact_blocks = @min(
+                manifest_log.pace.half_bar_compact_blocks(.{
+                    .log_blocks_count = @intCast(manifest_log.log_block_checksums.count),
+                }),
+                // Never compact closed blocks. (They haven't even been written yet.)
+                manifest_log.log_block_checksums.count - manifest_log.blocks_closed,
+            );
+            assert(manifest_log.compact_blocks.? <= manifest_log.pace.half_bar_compact_blocks_max);
+
+            manifest_log.grid_reservation = manifest_log.grid.reserve(
+                manifest_log.compact_blocks.? +
+                    manifest_log.pace.half_bar_append_blocks_max,
+            ).?;
 
             manifest_log.read_callback = callback;
-            manifest_log.flush(compact_flush_callback);
+            manifest_log.flush(compact_next_block);
         }
 
         fn compact_tick_callback(next_tick: *Grid.NextTick) void {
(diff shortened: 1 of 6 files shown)


src/lsm/manifest_log.zig
```diff
@@ -954,47 +1000,237 @@ pub fn ManifestLogType(comptime Storage: type) type {
...
+    }) u32 {
+        if (options.log_blocks_count < pace.log_blocks_cycle_max) {
+            // We have enough free manifest blocks that we could go a whole "cycle" without
+            // compacting any. It doesn't strictly matter how much compaction we do in this case, so
+            // just try to pace the work evenly, erring on the side of doing less work than
+            // necessary.
+            return @divFloor(
+                pace.half_bar_compact_blocks_max * options.log_blocks_count,
+                pace.log_blocks_cycle_max,
+            );
```
sentientwaffle: Some other possible policies here (pretty much anything works):
- Round instead of floor.
- `max(1, ...)` (compact at least 1 block).
- Even `return 0;` would be safe here, i.e. defer all compaction until our manifest is very large.

src/lsm/manifest_log.zig
```diff
@@ -954,47 +1000,237 @@ pub fn ManifestLogType(comptime Storage: type) type {
...
+///     means that "A" might in truth be fractional, but we would round up. For example, if "A" is
+///     2.1, for the purposes of the upper-bound it is 3. Because `C` is computed (below) as
+///     "A + compact_extra_blocks", the result is that we perform more compaction (relative to
+///     appends) than the block-granular constants indicate.
+///     As a result, we overestimate the upper-bound (or, equivalently, perform compaction more
+///     quickly than strictly necessary).
+///   - The calculation does *not* consider the "padding" appends in to a partial block written
+///     during a checkpoint. This oversight is masked because "A" is overestimated (see previous
+///     bullet).
+///
```
sentientwaffle, line 1109: I went back and forth a lot on whether we should implement the stricter upper bound. But it is quite a bit more complicated, and the benefit is only marginal. (A larger manifest is inexpensive memory-wise. Even the current overestimate of ~10k manifest blocks is still 6x lower than our original manifest block count.)

src/lsm/manifest_log.zig
```diff
@@ -630,6 +657,10 @@ pub fn ManifestLogType(comptime Storage: type) type {
         }
 
         /// `compact` does not close a partial block; that is only necessary during `checkpoint`.
+        ///
+        /// The (production) block size is large, so the number of blocks compacted per half-bar is
+        /// relatively small (e.g. ~6). We read them in sequence rather than parallel to spread the
+        /// work more evenly across the half-bar's beats.
         // TODO Make sure block reservation cannot fail — before compaction begins verify that
         // enough free blocks are available for all reservations.
         pub fn compact(manifest_log: *ManifestLog, callback: Callback, op: u64) void {
```
sentientwaffle, line 666: The changes to the compaction flow are to change it from compacting exactly 1 block per half-bar, to compacting a variable number of blocks per half-bar (`compact_blocks` is the count).

src/lsm/manifest_log_fuzz.zig
sentientwaffle, line 1: The modifications to the fuzzer are to:
- generate more realistic sequences of table compaction events
- try to generate sequences of events that approach the upper bound

src/lsm/manifest_log.zig
```diff
@@ -954,47 +1000,237 @@ pub fn ManifestLogType(comptime Storage: type) type {
...
         return (options.tree_id_max - options.tree_id_min) + 1;
     }
+};
 
-    /// The maximum number of table updates to the manifest by a half-measure of table
-    /// compaction.
+/// The goals of manifest log compaction are (in no particular order):
+///
+/// 1. Free enough manifest blocks such that there are always enough free slots in the manifest
+///    trailer to accommodate the appends by table compaction.
```
matklad: We don't have manifest trailer anymore, I think the only limit here is the memory requirement for `log_block_checksums` RingBuffer (that is, on disk our manifest log can be arbitrary large and is limited only by u64 addres space for blocks

src/lsm/manifest_log.zig
```diff
@@ -954,47 +1000,237 @@ pub fn ManifestLogType(comptime Storage: type) type {
...
+/// 2. Shrink the manifest: A smaller manifest means that fewer blocks need to be replayed during
```
matklad: Maybe shrink "manifest log"? Manifest is the cumulative effect of manifest log, and the manifest doesn't actually depend on compaction --- it's the definition of compaction that it shrinks the log while keeping the end result the same.

src/lsm/manifest_log.zig
matklad: This is super hand-wavy as I don't yet fully understand the logic here, but dont' we need an extra `option` here? It feels like, to pace compaction, we need to know not only the size of the manifest (`log_blocks_count`), but also the number of tables?

Consider this two situations:

|`log_blocks_count` | `table_count` |
-|-
|  100              |   10          |
|  100              |   1000        |

It feels like we should compact quite a bit in the first case (few tables), but maybe not as much in the second case. 

The current logic can't do that, because it has access only to the first column of this table. 

In other words, the calculation we do are relative to `T` --- the maximum number of tables that could in theory be supported by the data file. But in practice for the vast majority of data files would have some actual number of tables `t`, such that `t ≪ T`. 

It's not clear to me what would be the behavior of our pacing in this, non-extreme case? I think it could be one of:

* the manifest log grows to its maximal extent (as if it has full T tables)
* the manifest log shrinks to something very small
* the manifest log maintains maintains the same load factor as it would have in t=T case

I _think_ the behavior we want here is the last one --- if your data file is small, we want the log to be small as well (speedier recovery), but also we don't want to waste compaction work churning through "live" tables. I would be a bit surprised if we can achieve that behavior without passing the current number of tables, but maybe that's how it works out?

sentientwaffle: Good point!

So in short we to pace compaction relative to the log's density (the ratio of live tables to total entries).
Though at the same time, when the absolute number of log blocks is low, even if it is low density we probably don't want to compact at the maximum rate.

This feels more complicated than it "should" be, but:

```zig
            // We have enough free manifest blocks that we could go a whole "cycle" without
            // compacting any. It doesn't strictly matter how much compaction we do in this case, so
            // just try to pace the work evenly, maintaining a constant load factor with respect to
            // the cycle-max.

            // Our "target" block count extrapolates a log block count from our table count and the
            // log's maximum load factor.
            const log_blocks_target = @max(1, @divFloor(
                pace.log_blocks_cycle_max * options.tables_count,
                pace.tables_max,
            ));

            // Scale the compaction rate according to the total number of blocks.
            // e.g. if our manifest log only has 10 blocks, we shouldn't compact at the same
            // rate we would if we had 1000 blocks, even if the tables are low density.
            const compact_blocks_limit = @min(
                pace.log_blocks_cycle_max,
                stdx.div_ceil(
                    pace.half_bar_compact_blocks_max * options.log_blocks_count,
                    pace.log_blocks_cycle_max,
                ),
            );

            // Finally, scale the compaction rate based on the current load factor.
            return @min(
                compact_blocks_limit,
                @divFloor(
                    compact_blocks_limit * options.log_blocks_count,
                    log_blocks_target,
                ),
            );
```

src/lsm/manifest_log.zig
```diff
@@ -954,47 +1000,237 @@ pub fn ManifestLogType(comptime Storage: type) type {
...
-    fn blocks_count_max(options: *const Options) usize {
-        const count = 1 + 1 + options.blocks_count_appends();
-        assert(count >= 3);
-
-        return count;
+    fn half_bar_compact_blocks(pace: Pace, options: struct {
+        /// The number of manifest blocks that *currently* exist.
+        log_blocks_count: u32,
+    }) u32 {
+        if (options.log_blocks_count < pace.log_blocks_cycle_max) {
```
matklad: I _think_ this might want to be `pace.log_blocks_cycle_max - (half_bar_append_blocks_max)`.

If I understand the equations correctly, the invariant we prove is:

> If currently the manifest has `log_blocks_cycle_max` blocks, and you compact at max pace (pace.half_bar_compact_blocks_max), then the size of the manifest would not grow beyond `log_blocks_max`

But we might not get a chance to kick compaction into full gear exactly at `log_blocks_cycle_max` blocks! Eg, if we don't compact at all until we have `log_blocks_cycle_max`, we might be in a situation where we have exactly `log_blocks_cycle_max - 1` blocks, don't compact, and append `half_bar_append_blocks_max`. So we only start compacting from 

`log_blocks_cycle_max + half_bar_append_blocks_max` blocks, and that might hit a tiny bit higher? 


(given that the bound is conservative, it probably works out even without this adjustment, but this I think doesn't follow from our calcualtions).

matklad: Also, at the end of the doc comment, let's formulate the postivie invariat we have:

> If anytime manifest log at least `X` long you compact at last `Y` blocks, the total size of the manifest can't be larger than `Z`

That is, on the first reading it was not obvious that what's at play here is not only the speed we compact and at the speed we append at, but also the point when we _start_ compacting.

src/lsm/manifest_log.zig
matklad: SGTM! 

I think that maybe scaling by the total number of blocks is not _that_ necessary though --- if there are few blocs, but the load factor is high, then compaction is probably going to reduce the load factor significantly, as, for short manifest, we compact larger fraction of the manifest. 

If we don't scale by manifest log size, we can look _just_ at the load factor, and that'll naturally handle the case where we exceed pace.log_blocks_cycle_max, which would be simplification.

sentientwaffle: > I think that maybe scaling by the total number of blocks is not that necessary though


In that case I'd rather simplify!:

```zig
        // We have enough free manifest blocks that we could go a whole "cycle" without
        // compacting any. It doesn't strictly matter how much compaction we do in this case, so
        // just try to pace the work evenly, maintaining a constant load factor with respect to
        // the cycle-max.

        // Our "target" block count extrapolates a log block count from our table count and the
        // log's maximum load factor.
        const log_blocks_target = @max(1, @divFloor(
            pace.log_blocks_cycle_max * options.tables_count,
            pace.tables_max,
        ));

        return @min(
            pace.half_bar_compact_blocks_max,
            @divFloor(
                pace.half_bar_compact_blocks_max * options.log_blocks_count,
                log_blocks_target,
            ),
        );
```

i.e. scale `half_bar_compact_blocks_max` to the ratio of "current log blocks" to "target log blocks".

> If we don't scale by manifest log size, we can look just at the load factor, and that'll naturally handle the case where we exceed pace.log_blocks_cycle_max, which would be simplification.

I think I like handling the

```zig
        // Pretend we have an extra half_bar_append_blocks_max blocks so that we always switch to
        // the maximum compaction rate before we exceed the cycle-max.
        if (pace.log_blocks_cycle_max <=
            options.log_blocks_count + pace.half_bar_append_blocks_max)
        {
            return pace.half_bar_compact_blocks_max;
        }
```
branch explicitly + independently though, since it is critical for the upper-bound calculation, whereas the other branch is "anything goes".

---

So all together:

```zig
    fn half_bar_compact_blocks(pace: Pace, options: struct {
        /// The number of manifest blocks that *currently* exist.
        log_blocks_count: u32,
        /// The number of live tables.
        tables_count: u32,
    }) u32 {
        assert(options.tables_count <= pace.tables_max);

        // Pretend we have an extra half_bar_append_blocks_max blocks so that we always switch to
        // the maximum compaction rate before we exceed the cycle-max.
        if (pace.log_blocks_cycle_max <=
            options.log_blocks_count + pace.half_bar_append_blocks_max)
        {
            return pace.half_bar_compact_blocks_max;
        }

        // We have enough free manifest blocks that we could go a whole "cycle" without
        // compacting any. It doesn't strictly matter how much compaction we do in this case, so
        // just try to pace the work evenly, maintaining a constant load factor with respect to
        // the cycle-max.

        // Our "target" block count extrapolates a log block count from our table count and the
        // log's maximum load factor.
        const log_blocks_target = @max(1, @divFloor(
            pace.log_blocks_cycle_max * options.tables_count,
            pace.tables_max,
        ));

        return @min(
            pace.half_bar_compact_blocks_max,
            @divFloor(
                pace.half_bar_compact_blocks_max * options.log_blocks_count,
                log_blocks_target,
            ),
        );
    }
```

matklad: > I think I like handling the branch explicitly

:+1: I was also thinking that at the back of my mind!

matklad (approved):

Issue #1186, opened 2023-09-11
REPL: Control-C behavior


cb22:
When hitting Control-C inside the REPL, it exits immediately, even if you're in the middle of typing or a command. Rather, if you're midway through a command, it should clear whatever you've entered, and if you're at a blank prompt, only then should it exit the REPL (maybe after twice in a row, like Node does).

alxgh:
https://github.com/tigerbeetle/tigerbeetle/pull/1255

swaingotnochill:
Hi, I would like to pick this up. Can you assign me?
I will triage and see what's the issue on my end.

fidelicura:
Looks like this issue is fixed, see https://github.com/tigerbeetle/tigerbeetle/pull/1255#issuecomment-2566379581

ryukyi:
> even if you're in the middle of typing or a command

This part is fixed - it clears and jumps to new line

> if you're at a blank prompt, only then should it exit the REPL (maybe after twice in a row, like Node does).

Ctrl+C doesn't exit the repl on blank prompts:

<img width="584" height="704" alt="Image" src="https://github.com/user-attachments/assets/20f0652c-bbcd-4da7-93ff-11e8ac9a72b4" />

matklad:
Yeah, this works as intended, Ctrl-C clears input, Ctrl-D exits Repl, following conventions of `python`

PR #1255, opened 2023-10-26
 REPL: Control-C behavior

alxgh:
https://github.com/tigerbeetle/tigerbeetle/issues/1186

Handles Ctrl-C so it won't stops the client, only terminates the current prompt and opens a new one.

It is also possible to handle two consecutive Ctrl-C without typing in between but right now Ctrl-D does the work and I think that's fine :-)

NOTE: The part related to `windows` is not tested.


matklad:
Thanks for the PR @alxgh!

I am wondering, do we have reference which explains the correct UNIX way to handle this? Signals are tricky! I _think_ what we do here, printing to stdout, is actually signal safe (as Zig doesn't have a lock around stdout), but it seems like it doesn't behave quite like we want it to? This is what I get with the PR applied, if I `^C` after I hit `Enter` in the middle of the command:


```
λ ./tigerbeetle client --cluster=1 --addresses=3000
TigerBeetle CLI Client 0.14.157-20-gb84d25f2

Examples:
  create_accounts id=1 code=10 ledger=700,
                  id=2 code=10 ledger=700;
  lookup_accounts id=1;
  lookup_accounts id=1, id=2;

> invalid command

^C
> ;
Fail near line 1, column 0:

invalid command
^ Near here.

Operation must be help, create_accounts, lookup_accounts,
create_transfers, or lookup_transfers. Got: 'invalid'.
```

See how the input is not cleared --- 'invalid command' still gets to repl, even though ^C was supposed to clear it. 

I am not sure what's the right pattern to handle this though.

alxgh:
@matklad I think i got that bug fixed, the problem was when you hit enter it reads the input but and stores it so ctrl-c has no effect. I re-wrote the code so now it reads byte-by-byte and in case of ctrl-c clearing out the input.

I just had to use a global variable to handle the scenario which I don't know if it is acceptable or not but couldn't find a way to pass the Repl struct pointer to signal handler.

Also didn't find another way for handling this aside using signal handlers.

alxgh:
Also no idea how to test it by code :-)

cb22:
The REPL has this now!

Issue #489, opened 2023-02-22
Client: Batching


sentientwaffle:
### Batching Background

Batching requests is essential for good throughput & latency.

Right now applications must batch accounts/transfers/queries themselves before invoking the client. The client ships that whole batch to the cluster in a single message.

The goal of this design is to build transparent batching into all client implementations so that users don't need to think about them so much. (At least until they need to eke out more performance).

### Client Background

Currently the client architecture is:

  1. Language client, which is built atop the
  2. C client, which is built atop the
  3. Zig client (`vsr.Client`).

`vsr.Client`'s responsibilities are currently:

  - Sending requests, receiving replies.
  - Retrying requests
  - Automatic session registration
  - Handle session eviction
  - Request queueing
  - View discovery (ping/pong)

The C client's responsibilities:

  - Concurrency for language runtimes.

The language client's responsibilities:

  - Encode requests into bytes.
  - Decode replies from bytes.

## Terminology

- _logical batch_: A sequence of one or more events which will be sent together in a single request message.
    - Logical batches are created by the application.
    - The client is not allowed to split a logical batch across multiple request messages.
    - All of the events in a logical batch have a common `Operation`.
- _physical batch_: A sequence of one or more logical batches which are grouped together in a single request message.
    - Physical batches are created by the client itself.
    - All of the logical batches in a physical batch have a common `Operation`.

## Design

`vsr.Client` is additionally responsible for batching.

Currently,

```
const message: *Message = client.get_message();
... write events to message.body()
client.request(user_data, callback, operation, message, message_body_size)
```

Replace with:

```
const batch: BatchLogical(operation) = client.get_batch(operation, number of events);
... write events (the logical batch) to batch.slice() (a []u8.. or maybe it could be a []StateMachine.Event(operation) if that works)
... The memory returned by `Batch.slice()` points to a slice of message body memory.
Client.batch(user_data, callback, batch)
```

When `.get_batch()` is called:

1. Find the first `Batch` (aka batch) in the queue with a matching `operation` that has room for the entire `batch_logical`.
2. Copy the `batch_logical`'s bytes into that memory (or deserialize the accounts/transfers into that memory, depending on the language).
3. Save the `callback`+ `user_data` along with the range of events so that they can be used to demux the reply.
4. (If there is no such batch/message, create a new one and append it to the queue).

In `.batch()`:
1. Any time there is no request in flight, send the first batch immediately.

In the client `tick`:
[...]

PR #1257, opened 2023-10-27
VSR: Batch pending client requests

kprotty:
Implements #489

### DIFF
--- a/src/benchmark.zig
+++ b/src/benchmark.zig
@@ -404,7 +404,7 @@ const Benchmark = struct {
         const batch = b.client.batch_get(operation, event_size) catch unreachable;
         stdx.copy_disjoint(.exact, u8, batch.slice(), payload);
 
-        b.client.batch_sumit(
+        b.client.batch_submit(
             @intCast(@intFromPtr(b)),
             send_complete,
             batch,
--- a/src/benchmark.zig
+++ b/src/benchmark.zig
@@ -397,22 +397,17 @@ const Benchmark = struct {
     ) void {
         b.callback = callback;
 
-        const message = b.client.get_message();
-        errdefer b.client.release(message);
+        const event_size = switch (operation) {
+            inline else => |op| @divExact(payload.len, @sizeOf(StateMachine.Event(op))),
+        };
 
-        stdx.copy_disjoint(
-            .inexact,
-            u8,
-            message.buffer[@sizeOf(vsr.Header)..],
-            payload,
-        );
+        const batch = b.client.batch_get(operation, event_size) catch unreachable;
+        stdx.copy_disjoint(.exact, u8, batch.slice(), payload);
 
-        b.client.request(
+        b.client.batch_sumit(
             @intCast(@intFromPtr(b)),
             send_complete,
-            operation,
-            message,
-            payload.len,
+            batch,
         );
     }
 
--- a/src/benchmark.zig
+++ b/src/benchmark.zig
@@ -397,11 +397,11 @@ const Benchmark = struct {
     ) void {
         b.callback = callback;
 
-        const event_size = switch (operation) {
+        const event_count = switch (operation) {
             inline else => |op| @divExact(payload.len, @sizeOf(StateMachine.Event(op))),
         };
 
-        const batch = b.client.batch_get(operation, event_size) catch unreachable;
+        const batch = b.client.batch_get(operation, event_count) catch unreachable;
         stdx.copy_disjoint(.exact, u8, batch.slice(), payload);
 
         b.client.batch_submit(
--- a/src/clients/c/tb_client/echo_client.zig
+++ b/src/clients/c/tb_client/echo_client.zig
@@ -7,6 +7,7 @@ const constants = @import("../../../constants.zig");
 const vsr = @import("../../../vsr.zig");
 const Header = vsr.Header;
 
+const IOPS = @import("../../../iops.zig").IOPS;
 const RingBuffer = @import("../../../ring_buffer.zig").RingBuffer;
 const MessagePool = @import("../../../message_pool.zig").MessagePool;
 const Message = @import("../../../message_pool.zig").MessagePool.Message;
@@ -26,6 +27,7 @@ pub fn EchoClient(comptime StateMachine_: type, comptime MessageBus: type) type 
             };
         };
 
+        const DemuxIOPS = IOPS(Self.Request.Demux, constants.client_request_queue_max);
         const RequestQueue = RingBuffer(Self.Request, .{
             .array = constants.client_request_queue_max,
         });
@@ -34,6 +36,7 @@ pub fn EchoClient(comptime StateMachine_: type, comptime MessageBus: type) type 
         cluster: u128,
         messages_available: u32 = constants.client_request_queue_max,
         request_queue: RequestQueue = RequestQueue.init(),
+        demux_iops: DemuxIOPS = .{},
         message_pool: *MessagePool,
 
         pub fn init(
@@ -75,6 +78,9 @@ pub fn EchoClient(comptime StateMachine_: type, comptime MessageBus: type) type 
             };
             assert(body_size <= constants.message_body_size_max);
 
+            const demux = self.demux_iops.acquire() orelse return error.TooManyOutstanding;
+            errdefer self.demux_iops.release(demux);
+
             if (self.messages_available == 0) return error.TooManyOutstanding;
             const message = self.get_message();
             errdefer self.release(message);
@@ -92,12 +98,14 @@ pub fn EchoClient(comptime StateMachine_: type, comptime MessageBus: type) type 
             return Batch{
                 .message = message_request,
                 .event_count = @intCast(event_count),
+                .demux = demux,
             };
         }
 
         pub const Batch = struct {
             message: *Message.Request,
             event_count: u16,
+            demux: *Self.Request.Demux,
 
             pub fn slice(batch: Batch) []u8 {
                 return batch.message.body();
@@ -112,14 +120,18 @@ pub fn EchoClient(comptime StateMachine_: type, comptime MessageBus: type) type 
         ) void {
             const message = batch.message;
 
-            assert(!self.request_queue.full());
-
-            self.request_queue.push_assume_capacity(.{ .message = message, .demux = .{
+            batch.demux.* = .{
                 .user_data = user_data,
                 .callback = callback,
                 .event_count = batch.event_count,
                 .event_offset = 0,
-            } });
+            };
+
+            assert(!self.request_queue.full());
+            self.request_queue.push_assume_capacity(.{ .message = message });
+
+            const request = self.request_queue.tail_ptr().?;
+            request.demux_queue.push(batch.demux);
         }
 
         pub fn get_message(self: *Self) *Message {
@@ -150,9 +162,15 @@ pub fn EchoClient(comptime StateMachine_: type, comptime MessageBus: type) type 
                 // callback. This necessitates a `copy_disjoint` above.
                 self.release(inflight.message.base());
 
-                assert(inflight.demux_queue.empty());
-                inflight.demux.callback.?(
-                    inflight.demux.user_data,
+                const demux = inflight.demux_queue.peek().?;
+                assert(inflight.demux_queue.count == 1);
+
+                const user_data = demux.user_data;
+                const callback = demux.callback.?;
+                self.demux_iops.release(demux);
+
+                callback(
+                    user_data,
                     reply_message.header.operation.cast(Self.StateMachine),
                     reply_message.body(),
                 );
--- a/src/clients/c/tb_client/echo_client.zig
+++ b/src/clients/c/tb_client/echo_client.zig
@@ -27,8 +27,13 @@ pub fn EchoClient(comptime StateMachine_: type, comptime MessageBus: type) type 
             };
         };
 
-        const DemuxIOPS = IOPS(Self.Request.Demux, constants.client_request_queue_max);
-        const RequestQueue = RingBuffer(Self.Request, .{
+        const EchoRequest = struct {
+            message: *Message.Request,
+            user_data: u128,
+            callback: Self.Request.Callback,
+        };
+
+        const RequestQueue = RingBuffer(EchoRequest, .{
             .array = constants.client_request_queue_max,
         });
 
@@ -36,7 +41,6 @@ pub fn EchoClient(comptime StateMachine_: type, comptime MessageBus: type) type 
         cluster: u128,
         messages_available: u32 = constants.client_request_queue_max,
         request_queue: RequestQueue = RequestQueue.init(),
-        demux_iops: DemuxIOPS = .{},
         message_pool: *MessagePool,
 
         pub fn init(
@@ -74,12 +78,11 @@ pub fn EchoClient(comptime StateMachine_: type, comptime MessageBus: type) type 
             event_count: usize,
         ) Self.BatchError!Batch {
             const body_size = switch (operation) {
-                inline else => |op| @sizeOf(Self.StateMachine.Event(op)) * event_count,
+                inline else => |batch_operation| blk: {
+                    break :blk @sizeOf(Self.StateMachine.Event(batch_operation)) * event_count;
+                },
             };
             assert(body_size <= constants.message_body_size_max);
-
-            const demux = self.demux_iops.acquire() orelse return error.TooManyOutstanding;
-            errdefer self.demux_iops.release(demux);
 
             if (self.messages_available == 0) return error.TooManyOutstanding;
             const message = self.get_message();
@@ -95,17 +98,11 @@ pub fn EchoClient(comptime StateMachine_: type, comptime MessageBus: type) type 
                 .size = @intCast(@sizeOf(Header) + body_size),
             };
 
-            return Batch{
-                .message = message_request,
-                .event_count = @intCast(event_count),
-                .demux = demux,
-            };
+            return Batch{ .message = message_request };
         }
 
         pub const Batch = struct {
             message: *Message.Request,
-            event_count: u16,
-            demux: *Self.Request.Demux,
 
             pub fn slice(batch: Batch) []u8 {
                 return batch.message.body();
@@ -118,20 +115,12 @@ pub fn EchoClient(comptime StateMachine_: type, comptime MessageBus: type) type 
             callback: Self.Request.Callback,
             batch: Batch,
         ) void {
-            const message = batch.message;
-
-            batch.demux.* = .{
+            assert(!self.request_queue.full());
+            self.request_queue.push_assume_capacity(.{
+                .message = batch.message,
                 .user_data = user_data,
                 .callback = callback,
-                .event_count = batch.event_count,
-                .event_offset = 0,
-            };
-
-            assert(!self.request_queue.full());
-            self.request_queue.push_assume_capacity(.{ .message = message });
-
-            const request = self.request_queue.tail_ptr().?;
-            request.demux_queue.push(batch.demux);
+            });
         }
 
         pub fn get_message(self: *Self) *Message {
@@ -147,31 +136,25 @@ pub fn EchoClient(comptime StateMachine_: type, comptime MessageBus: type) type 
         }
 
         fn reply(self: *Self) void {
-            while (self.request_queue.pop()) |inflight| {
+            while (self.request_queue.pop()) |request| {
                 const reply_message = self.message_pool.get_message(.request);
                 defer self.message_pool.unref(reply_message.base());
 
+                const operation = request.message.header.operation.cast(Self.StateMachine);
                 stdx.copy_disjoint(
                     .exact,
                     u8,
                     reply_message.buffer,
-                    inflight.message.buffer,
+                    request.message.buffer,
                 );
 
                 // Similarly to the real client, release the request message before invoking the
                 // callback. This necessitates a `copy_disjoint` above.
-                self.release(inflight.message.base());
+                self.release(request.message.base());
 
-                const demux = inflight.demux_queue.peek().?;
-                assert(inflight.demux_queue.count == 1);
-
-                const user_data = demux.user_data;
-                const callback = demux.callback.?;
-                self.demux_iops.release(demux);
-
-                callback(
-                    user_data,
-                    reply_message.header.operation.cast(Self.StateMachine),
+                request.callback(
+                    request.user_data,
+                    operation,
                     reply_message.body(),
                 );
             }
--- a/src/clients/c/tb_client/echo_client.zig
+++ b/src/clients/c/tb_client/echo_client.zig
@@ -22,6 +22,7 @@ pub fn EchoClient(comptime StateMachine_: type, comptime MessageBus: type) type 
                 pub const StateMachine = Client.StateMachine;
                 pub const Error = Client.Error;
                 pub const Request = Client.Request;
+                pub const BatchError = Client.BatchError;
             };
         };
 
@@ -29,6 +30,8 @@ pub fn EchoClient(comptime StateMachine_: type, comptime MessageBus: type) type 
             .array = constants.client_request_queue_max,
         });
 
+        id: u128,
+        cluster: u32,
         messages_available: u32 = constants.client_request_queue_max,
         request_queue: RequestQueue = RequestQueue.init(),
         message_pool: *MessagePool,
@@ -42,12 +45,12 @@ pub fn EchoClient(comptime StateMachine_: type, comptime MessageBus: type) type 
             message_bus_options: MessageBus.Options,
         ) !Self {
             _ = allocator;
-            _ = id;
-            _ = cluster;
             _ = replica_count;
             _ = message_bus_options;
 
             return Self{
+                .id = id,
+                .cluster = cluster,
                 .message_pool = message_pool,
             };
         }
@@ -62,31 +65,53 @@ pub fn EchoClient(comptime StateMachine_: type, comptime MessageBus: type) type 
             self.reply();
         }
 
-        pub fn request(
+        pub fn batch_get(
+            self: *Self,
+            operation: Self.StateMachine.Operation,
+            event_count: usize,
+        ) Self.BatchError!Batch {
+            const body_size = switch (operation) {
+                inline else => |op| @sizeOf(Self.StateMachine.Event(op)) * event_count,
+            };
+            assert(body_size <= constants.message_body_size_max);
+
+            if (self.messages_available == 0) return error.TooManyOutstanding;
+            const message = self.get_message();
+            errdefer self.release(message);
+
+            message.header.* = .{
+                .client = self.id,
+                .request = 0,
+                .cluster = self.cluster,
+                .command = .request,
+                .operation = vsr.Operation.from(Self.StateMachine, operation),
+                .size = @intCast(@sizeOf(Header) + body_size),
+            };
+
+            return Batch{ .message = message };
+        }
+
+        pub const Batch = struct {
+            message: *Message,
+
+            pub fn slice(batch: Batch) []u8 {
+                return batch.message.body();
+            }
+        };
+
+        pub fn batch_submit(
             self: *Self,
             user_data: u128,
             callback: Self.Request.Callback,
-            operation: Self.StateMachine.Operation,
-            message: *Message,
-            message_body_size: usize,
+            batch: Batch,
         ) void {
-            const message_request = message.build(.request);
-
-            message_request.header.* = .{
-                .client = 0,
-                .request = 0,
-                .cluster = 0,
-                .command = .request,
-                .operation = vsr.Operation.from(Self.StateMachine, operation),
-                .size = @as(u32, @intCast(@sizeOf(Header) + message_body_size)),
-            };
+            const message = batch.message;
 
             assert(!self.request_queue.full());
 
             self.request_queue.push_assume_capacity(.{
-                .user_data = user_data,
-                .callback = callback,
-                .message = message_request,
+                .info = .{ .user_data = user_data, .callback = callback },
+                .message = message,
             });
         }
 
@@ -118,8 +143,8 @@ pub fn EchoClient(comptime StateMachine_: type, comptime MessageBus: type) type 
                 // callback. This necessitates a `copy_disjoint` above.
                 self.release(inflight.message.base());
 
-                inflight.callback.?(
-                    inflight.user_data,
+                inflight.info.callback.?(
+                    inflight.info.user_data,
                     reply_message.header.operation.cast(Self.StateMachine),
                     reply_message.body(),
                 );
--- a/src/clients/c/tb_client/echo_client.zig
+++ b/src/clients/c/tb_client/echo_client.zig
@@ -31,7 +31,7 @@ pub fn EchoClient(comptime StateMachine_: type, comptime MessageBus: type) type 
         });
 
         id: u128,
-        cluster: u32,
+        cluster: u128,
         messages_available: u32 = constants.client_request_queue_max,
         request_queue: RequestQueue = RequestQueue.init(),
         message_pool: *MessagePool,
@@ -78,8 +78,9 @@ pub fn EchoClient(comptime StateMachine_: type, comptime MessageBus: type) type 
             if (self.messages_available == 0) return error.TooManyOutstanding;
             const message = self.get_message();
             errdefer self.release(message);
-
-            message.header.* = .{
+            
+            const message_request = message.build(.request);
+            message_request.header.* = .{
                 .client = self.id,
                 .request = 0,
                 .cluster = self.cluster,
@@ -88,11 +89,11 @@ pub fn EchoClient(comptime StateMachine_: type, comptime MessageBus: type) type 
                 .size = @intCast(@sizeOf(Header) + body_size),
             };
 
-            return Batch{ .message = message };
+            return Batch{ .message = message_request };
         }
 
         pub const Batch = struct {
-            message: *Message,
+            message: *Message.Request,
 
             pub fn slice(batch: Batch) []u8 {
                 return batch.message.body();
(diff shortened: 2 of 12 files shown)


batiati (commented):
TIL `switch inline x,y,z` was possible⚡

src/testing/state_machine.zig
```diff
@@ -22,6 +22,23 @@ pub fn StateMachineType(
...
+        }
+
+        /// TODO: Support batching in test StateMachine (?)
+        pub fn batch_demux(
+            operation: Operation,
+            results: []const u8,
+            demuxed: []u8,
+            event: struct { index: u32, count: u32 },
+        ) u32 {
+            _ = .{ operation, results, demuxed, event };
```
batiati: Nice pattern 👍

src/vsr/client.zig
```diff
@@ -26,12 +28,25 @@ pub fn Client(comptime StateMachine_: type, comptime MessageBus: type) type {
...
+                event_count: u16,
+                event_offset: u16,
+            };
+
+            info: Info,
             message: *Message,
+            demux_queue: FIFO(Demux) = .{ .name = null },
         };
 
+        const DemuxPool = IOPS(Request.Demux, constants.client_request_queue_max);
```
batiati: Not sure if we want to use `client_request_queue_max` here.
Theoretically, we could have 8k single-item batches per request * `client_request_queue_max`, which is a lot to preallocate.
That's why we need a `DemuxPool` and a new limit `batch_logical_max`.

We can receive `batch_logical_max` as a comptime parameter in `vsr.Client`, allowing the user to set it to `0` if they want to disable batching completely.

src/vsr/client.zig
```diff
@@ -196,50 +226,142 @@ pub fn Client(comptime StateMachine_: type, comptime MessageBus: type) type {
...
+                    };
+                } else blk: {
+                    assert(batch.request.demux_queue.empty());
+                    break :blk 0;
+                };
+
+                const body = batch.request.message.body();
+                return body[body_offset..];
+            }
+        };
```
batiati: Let's move this struct to the top, among the other decls.

src/vsr/client.zig
```diff
@@ -196,50 +226,142 @@ pub fn Client(comptime StateMachine_: type, comptime MessageBus: type) type {
...
             self: *Self,
-            user_data: u128,
-            callback: Request.Callback,
             operation: StateMachine.Operation,
-            message: *Message,
-            message_body_size: usize,
-        ) void {
+            event_count: usize,
+        ) BatchError!Batch {
             assert(@intFromEnum(operation) >= constants.vsr_operations_reserved);
```
batiati: Can you document here that a `Batch` cannot be stored, and once acquired it cannot be released back and the user must call `batch_submit` before the next `tick()`?

It's a quite sensitive API change.

src/vsr/client.zig
```diff
@@ -196,50 +226,142 @@ pub fn Client(comptime StateMachine_: type, comptime MessageBus: type) type {
...
+            if (StateMachine.batch_logical(operation)) {
+                var it = self.request_queue.iterator_mutable();
+                while (it.next_ptr()) |request| {
+                    if (request == self.request_inflight) continue;
+                    if (request.message.header.command != .request) continue;
+                    if (request.message.header.operation.cast(StateMachine) != operation) continue;
+                    if (request.message.header.size + body_size > constants.message_size_max) continue;
+
+                    // Reserve a Demux node for attaching this Batch onto the existing Request.
+                    // TODO: Should this return TooManyOutstanding? Currently, it falls back to new.
```
batiati: It should fallback to a new request, which can return `TooManyOutstanding` if there's no message available.

src/vsr/client.zig
```diff
@@ -196,50 +226,142 @@ pub fn Client(comptime StateMachine_: type, comptime MessageBus: type) type {
...
+                        .info = undefined, // Set during batch_submit().
+                        .event_count = @intCast(event_count),
+                        // Offset from either the last queued Demux or the request initial size.
+                        .event_offset = blk: {
+                            if (request.demux_queue.peek_last()) |tail| {
+                                break :blk tail.event_offset + tail.event_count;
+                            }
+                            const request_body_size = request.message.header.size - @sizeOf(Header);
+                            break :blk @intCast(@divExact(request_body_size, event_size));
+                        },
```
batiati: - Use `else`
- Can we use `request.message.body().len` instead of recalculating it?

src/state_machine.zig
```diff
@@ -100,6 +100,47 @@ pub fn StateMachineType(
...
+        /// into the demuxed byte array, returning the amount of bytes written to it.
+        pub fn batch_demux(
+            operation: Operation,
+            results: []const u8,
+            demuxed: []u8,
+            event: struct { index: u32, count: u32 },
+        ) u32 {
+            switch (operation) {
+                inline .create_accounts, .create_transfers => |op| {
+                    const out_results = std.mem.bytesAsSlice(Result(op), demuxed);
```
batiati: add `assert(batch_logical(op))`

src/vsr/client.zig
```diff
@@ -78,10 +93,20 @@ pub fn Client(comptime StateMachine_: type, comptime MessageBus: type) type {
         /// requests can be submitted.
         messages_available: u32 = constants.client_request_queue_max,
 
+        /// Pool of nodes which represent batched Infos on a given Request.
+        demux_pool: DemuxPool = .{},
+
+        /// StateMachine can't take advantage of Request.demux_queue order for in-place demuxing,
+        /// so instead it demuxes into a scratch buffer and that's passed into Request.Callback.
+        demux_buffer: *[constants.message_body_size_max]u8,
```
batiati: This buffer makes the support for batching generic for any state machine implementation, but it comes with the cost of one more copy ...

Since the Ops `create_accounts` and `create_transfers` can be demuxed in place, I think it's ok to assume that any state machine that supports batches also needs a way to demux in place.

Let's remove this buffer and if it's going to be a problem in the future we can always reconsider and introduce it back.

src/state_machine.zig
batiati, line 1: Could you add unit tests for all demux/batching logic?

batiati (changes requested):

src/vsr/client.zig
```diff
@@ -26,10 +28,83 @@ pub fn Client(comptime StateMachine_: type, comptime MessageBus: type) type {
...
+                return null;
+            }
+
+            fn release(pool: *DemuxPool, demux: *Request.Demux) void {
+                assert(@intFromPtr(demux) < @intFromPtr(pool.memory.ptr + pool.memory.len));
+                assert(@intFromPtr(demux) >= @intFromPtr(&pool.memory[0]));
+                demux.next = pool.free;
+                pool.free = demux;
+            }
         };
```
batiati: Pretty cool, lazily populating the `free` list during `release` and not during the initialization 🤓🙌
Just one thing feels strange: `pool.used` shouldn't be named `used` since it's not decremented during `release`.  It's more like "the number of items that now live in the linked list". Can you think of something better?

On the other hand, we are adding an unnecessary `if` branch in the hot path `acquire` instead of populating the linked list during `init` time. Wdyt?

src/vsr/client.zig
```diff
@@ -196,50 +282,117 @@ pub fn Client(comptime StateMachine_: type, comptime MessageBus: type) type {
...
+            // a batched Request.
+            TooManyOutstanding,
+        };
+
+        /// Allocate space in a vsr Message to write [event_count]StateMachine.Event(operation).
+        /// Unlike `raw_request`, a Batch will attempt to piggy-back onto an existing Message.
+        /// The returned Batch provides a slice of memory to write the events to. Once populated,
+        /// use `batch_submit` to commit the writes to the Message and send it through the Client.
+        /// NOTE: A Batch cannot be stored and must be passed to batch_submit() before any calling
+        /// any other methods on the Client.
```
batiati:
```suggestion
        /// NOTE: A Batch cannot be stored and must be passed to batch_submit() before calling
        /// any other methods on the Client.
```

src/state_machine.zig
```diff
@@ -42,6 +42,18 @@ pub fn StateMachineType(
...
+            pub const batch_logical_max = blk: {
+                var max: usize = 0;
+                inline for (std.enums.values(Operation)) |operation| {
+                    if (@intFromEnum(operation) < config.vsr_operations_reserved) continue;
+                    if (!batch_logical_allowed(operation)) continue;
+                    max = @max(max, batch_max.operation_batch_max(operation));
+                }
+                break :blk max * global_constants.client_request_queue_max;
+            };
+
```
batiati: The `batch_logical_max` calculation is correct. However, we might not want to allocate that much memory (about 12MB, right?), given it's an odd worst-case scenario.

I suggest we keep it configurable per `vsr.Client`, since the caller site knows better the expected batching usage (eg. anything from `0` to `batch_logical_max` inclusive). 
For our `tb_client`, we can define an arbitrary default value not greater than a few thousand batches.

batiati (changes requested):
The unit test for the client + message bus is great! Can we add units at the state machine as well to test the specific demux logic?

src/vsr/client.zig
kprotty: What about `pool.memory_allocated`?

As for the `memory.len` check, it happened lazily to avoid faulting/iterating 12MB worth of tiny items on init. It's always predicted false (excluding the odd worst-case scenario down below) so it's similar to `+` operator which does overflow checks in ReleaseSafe.

src/state_machine.zig
sentientwaffle: I think `batch_logical_max` doesn't belong in the `StateMachine` -- it should be in `Client` instead, if it is needed.

The algorithm to compute `batch_logical_max` would be identical across any state machine implementation. We want to minimize the surface area for implementing new state machines.

> I suggest we keep it configurable per vsr.Client,

That is an improvement, but why not go with a context-style API:
- Then we can get rid of `batch_logical_max` and `client_request_queue_max`? (The limit moves up a layer. Clients could have no limit (by using dynamic allocator), or their own static limit.)
- Reduced `Client` surface area: Back to a single `request` function, instead of both `batch_get` and `batch_submit`.
- Reduced `Client` surface area: I think there would be no possible `error` on the request method.

src/state_machine.zig
```diff
@@ -100,6 +112,51 @@ pub fn StateMachineType(
...
         };
 
+        /// Returns whether or not the operation can be batched at the VSR layer.
+        /// If so, the StateMachine must support demuxing batched operations below.
+        pub fn batch_logical_allowed(operation: Operation) bool {
+            return switch (operation) {
+                .create_accounts, .create_transfers => true,
+                .lookup_accounts, .lookup_transfers => false, // Don't batch lookups for now.
+            };
+        }
```
sentientwaffle: Rather than a function, this should be a `std.enums.EnumArray(Operation, bool)`.
That encodes the intent more directly -- this isn't a runtime decision, it shouldn't every vary.

src/state_machine.zig
```diff
@@ -100,6 +112,51 @@ pub fn StateMachineType(
...
+        pub fn DemuxerType(comptime operation: Operation) type {
+            comptime assert(batch_logical_allowed(operation));
+            return struct {
+                const Demuxer = @This();
+                const CreateResult = Result(operation);
+
+                results: []CreateResult,
+
+                /// Create a Demuxer which can extract Results out of the reply bytes in-place.
+                pub fn init(reply: []align(@alignOf(CreateResult)) u8) Demuxer {
```
sentientwaffle: Why not take `[]CreateResult` as an argument?

src/state_machine.zig
```diff
@@ -100,6 +112,51 @@ pub fn StateMachineType(
...
+                }
+
+                /// Returns a slice into the the original reply bytes with Results matching the
+                /// Event range (offset and size). Each subsequent call to demux() must have ranges
+                /// that are disjoint and increase monotonically.
+                pub fn decode(
+                    self: *Demuxer,
+                    event_offset: u32,
+                    event_size: u32,
+                ) []align(@alignOf(CreateResult)) u8 {
```
sentientwaffle: Maybe return `[]CreateResult` (or `[]const CreateResult`, if that would work), and let the caller cast to bytes? (To move logic out of the state machine, and make the type signature reflect more closely the actual data.)

src/vsr/client.zig
```diff
@@ -26,10 +28,83 @@ pub fn Client(comptime StateMachine_: type, comptime MessageBus: type) type {
...
+            };
+
+            info: Info,
             message: *Message,
+            demux_queue: FIFO(Demux) = .{ .name = null },
+        };
+
+        pub const Batch = struct {
+            request: *Request,
+            demux: ?*Request.Demux, // Null iff batch_get() newly created `request`
```
sentientwaffle: Comment seems ungrammatical, or at least I can't parse it. Also missing punctuation.
Does it mean "Null iff batch_get() **is a** newly created `request`."?

src/vsr/client.zig
```diff
@@ -196,50 +282,117 @@ pub fn Client(comptime StateMachine_: type, comptime MessageBus: type) type {
...
 
-            // If the queue was empty, then there is no request inflight and we must send this one:
-            if (was_empty) self.send_request_for_the_first_time(message);
+            return Batch{
+                .request = self.request_queue.tail_ptr().?,
+                .demux = null,
+            };
+        }
+
+        /// After writing Event to batch.slice(), commit the batched
```
sentientwaffle: Incomplete sentence.

src/vsr/client.zig
```diff
@@ -196,50 +282,117 @@ pub fn Client(comptime StateMachine_: type, comptime MessageBus: type) type {
...
+            callback: Request.Callback,
+            batch: Batch,
+        ) void {
+            const info_ptr = if (batch.demux) |demux| &demux.info else &batch.request.info;
+            info_ptr.* = .{
+                .user_data = user_data,
+                .callback = callback,
+            };
+
+            // Newly made Requests are populated and potentially pushed through VSR if no inflight.
```
sentientwaffle: Incomplete sentence.

src/vsr/client.zig
```diff
@@ -568,9 +797,11 @@ pub fn Client(comptime StateMachine_: type, comptime MessageBus: type) type {
             self.message_bus.send_message_to_replica(replica, message);
         }
 
-        fn send_request_for_the_first_time(self: *Self, message: *Message) void {
-            assert(self.request_queue.head_ptr().?.message == message);
+        fn send_request_for_the_first_time(self: *Self) void {
```
sentientwaffle: I think I'd prefer if there was no `request_inflight` field and when needed we just check the head of the request queue. Otherwise there is an implicit (rather than explicit) contract in how `send_request_for_the_first_time` is used, which I think makes the code harder to follow.

src/vsr/client.zig
```diff
@@ -196,50 +282,117 @@ pub fn Client(comptime StateMachine_: type, comptime MessageBus: type) type {
...
+            assert(body_size <= constants.message_body_size_max);
+
+            // Check-in with the StateMachine to see if this operation can even be batched.
+            // If so, find an existing Request with the same Op that has room in its Message.
+            // The request must not be the one currently inflight in VSR as its Message is sealed.
+            if (StateMachine.batch_logical_allowed(operation)) {
+                var it = self.request_queue.iterator_mutable();
+                while (it.next_ptr()) |request| {
+                    if (request == self.request_inflight) continue;
+                    if (request.message.header.command != .request) continue;
```
sentientwaffle: How is the command ever not `request`?

src/vsr/client.zig
```diff
@@ -26,10 +28,83 @@ pub fn Client(comptime StateMachine_: type, comptime MessageBus: type) type {
...
+            info: Info,
             message: *Message,
+            demux_queue: FIFO(Demux) = .{ .name = null },
+        };
+
+        pub const Batch = struct {
+            request: *Request,
+            demux: ?*Request.Demux, // Null iff batch_get() newly created `request`
+
+            pub fn slice(batch: Batch) []u8 {
```
sentientwaffle, line 49: Assert the request's message's command.

src/vsr/client.zig
```diff
@@ -26,10 +28,83 @@ pub fn Client(comptime StateMachine_: type, comptime MessageBus: type) type {
...
+                if (pool.free) |demux| {
+                    pool.free = demux.next;
+                    return demux;
+                }
+                if (pool.used < pool.memory.len) {
+                    defer pool.used += 1;
+                    return &pool.memory[pool.used];
+                }
+                return null;
+            }
```
sentientwaffle: Why use the "lazy fifo" pattern rather than an IOPS? :thinking:

batiati: IOPS is limited to 2^6 elements 🙁

sentientwaffle: Hmm, is that arbitrary? We could just bump that capacity `u6` up to a large enough int to fix whatever the expected capacity here is. Admittedly it introduces a bitset (which the current `DemuxPool` avoids), but it would also remove ~30 lines of code.

kprotty: IOPS uses an array for the elements and `batch_logical_max` was so large (12MB worth of Demux nodes) that it would make the vsr.Client overflow the stack.

src/vsr/client.zig
kprotty: Need to avoid batching into the head Request of the queue if it's currently inflight (message has already been sent to MessageBus). The implicit contract is indeed kind of a hack.

src/vsr/client.zig
kprotty: Good point: `.register` can be put into the queue, but if so it's always `request_inflight`.

sentientwaffle: `register` is the  operation; the command is still `request` even then.

src/vsr/client.zig
kprotty: "If `demux` is null, the `request` is newly create."

"If `demux` is _not_ null, the `request` already exists".

```zig
const Batch = union(enum) {
  newly_created: *Request,
  piggyback_onto_existing: struct { *Request, *Demux },
}; 
```

src/testing/state_machine.zig
```diff
@@ -3,14 +3,13 @@ const assert = std.debug.assert;
...
 const stdx = @import("../stdx.zig");
-const constants = @import("../constants.zig");
 const vsr = @import("../vsr.zig");
 const GrooveType = @import("../lsm/groove.zig").GrooveType;
 const ForestType = @import("../lsm/forest.zig").ForestType;
 
 pub fn StateMachineType(
     comptime Storage: type,
-    comptime config: constants.StateMachineConfig,
+    comptime config: @import("../constants.zig").StateMachineConfig,
```
batiati: Let's use the same pattern `const global_constants = @import("constants.zig");` seen in the real  StateMachine.zig code.

src/vsr/client.zig
batiati: I like the `union(enum)` idea.
We already have to check `demux == null` everywhere we use a batch, so let's use a proper `switch` statement instead.

The field's names can be shorter:
```Zig
const Batch = union(enum) {
  new: *Request,
  append: struct { *Request, *Demux },
};
```

src/vsr/client.zig
```diff
@@ -26,10 +29,97 @@ pub fn Client(comptime StateMachine_: type, comptime MessageBus: type) type {
...
+                const body_offset = if (batch.demux) |demux| blk: {
+                    const operation = message.header.operation.cast(StateMachine);
+                    switch (operation) {
+                        inline else => |op| {
+                            const event_size: u32 = @sizeOf(StateMachine.Event(op));
+                            break :blk event_size * demux.event_offset;
+                        },
+                    }
+                } else blk: {
+                    assert(batch.request.demux_queue.empty());
```
batiati: We can also assert the inverse `assert(!batch.request.demux_queue.empty())` when a `demux` is present.

Maybe we could assert if the `demux` is contained in the `request.demux_queue` (enclosed by `if (constants.verify)`)

src/state_machine.zig
batiati: > That is an improvement, but why not go with a context-style API:
>
> - Then we can get rid of batch_logical_max and client_request_queue_max? (The limit moves up a layer. Clients could have no limit (by using dynamic allocator), or their own static limit.)
> - Reduced Client surface area: Back to a single request function, instead of both batch_get and batch_submit.
> - Reduced Client surface area: I think there would be no possible error on the request method.

I agree we can improve the client API, however, I thought it was out of the scope of this PR.
Can you elaborate on the suggestion of a context-style API a bit further?

...

The current implementation of the `tb_client` can take advantage of 4096 batches at most since it's the hard-coded upper `concurrency_max` limit.
That's why I think it makes sense for the caller to save memory specifying a new limit, lower than `batch_logical_max`.

batiati:
### Benchmark

I just created an [async benchmark](https://gist.github.com/batiati/8bd19479480ea0962d783bc794ff907c) running 4096 concurrent tasks (via thread pool), each one calling `create_transfers` with **only a single transfer**: 

- branch main, 10K transfers without client batching:
```
547 transfers per second with 4096 concurrent tasks.
create_transfers max p100 latency per transfer = 14_499ms.
total 10_000 transfers in 18_272ms.
```

- branch king-client-batching, 1M transfers:
```
165_837 transfers per second with 4096 concurrent tasks.
create_transfers max p100 latency per transfer = 93ms.
total 1_000_000 transfers in 6030ms.
```

The TPS numbers are impressively close to the benchmark with 8191 transfers per batch! Well done!
The throughput depends only on the application's capacity to submit enough requests to fill a batch, limited by `concurrency_max`.

src/state_machine.zig
sentientwaffle: > I agree we can improve the client API, however, I thought it was out of the scope of this PR.

Sounds good to me.

src/vsr/client.zig
```diff
@@ -433,22 +626,74 @@ pub fn Client(comptime StateMachine_: type, comptime MessageBus: type) type {
...
+                },
+            };
+
+            // Make sure to add the inflight_demux at the front of its queue.
+            inflight_demux.next = inflight.demux_queue.out;
+            inflight.demux_queue.out = &inflight_demux;
+            inflight.demux_queue.count += 1;
+
+            switch (operation) {
+                inline else => |op| {
```
sentientwaffle: Let's not shorten "operation" to "op", since op numbers are a thing. Maybe `operation_comptime`?

src/vsr/client.zig
```diff
@@ -433,22 +626,74 @@ pub fn Client(comptime StateMachine_: type, comptime MessageBus: type) type {
...
+                    if (comptime !StateMachine.batch_logical_allowed.get(op)) {
+                        std.debug.panic("invalid batched op: {s}", .{
+                            inflight_operation.tag_name(StateMachine),
+                        });
+                    }
```
sentientwaffle: Can this error be hit somehow? I don't understand why this has a `if () { panic }` rather than a simple assertion, e.g. `assert(StateMachine.batch_logical_allowed.get(op));`.

Also it might be more clear if it was moved up to after we fall through `if (inflight.demux_queue.empty())`?

src/state_machine.zig
```diff
@@ -100,6 +100,48 @@ pub fn StateMachineType(
...
+                    return Demuxer{ .results = reply };
+                }
+
+                /// Returns a slice into the the original reply bytes with Results matching the
+                /// Event range (offset and size). Each subsequent call to demux() must have ranges
+                /// that are disjoint and increase monotonically.
+                pub fn decode(self: *Demuxer, event_offset: u32, event_size: u32) []CreateResult {
+                    // Count all results from the start of our slice which match the Event range,
+                    // Updating the result.indexes to be relative to the Event range in the process.
+                    const demuxed = for (self.results, 0..) |*result, i| {
```
sentientwaffle: Could this implementation be more efficient if we remember the previous batch's offset within `results`, rather than re-iterating over all of `results` for each batch that we demux?

e.g. right now I think if each batch has exactly one event, then demultiplexing is `O(n²)`.

src/vsr/client.zig
```diff
@@ -26,10 +29,97 @@ pub fn Client(comptime StateMachine_: type, comptime MessageBus: type) type {
                 operation: StateMachine.Operation,
                 results: []const u8,
             ) void;
-            user_data: u128,
-            // Null iff operation=register.
-            callback: ?Callback,
+
+            const Info = struct {
```
sentientwaffle: `Info` is a very vague/generic name, is there a more descriptive name we could use?

Or even better, could we get rid of it entirely?:
- Inline `user_data`/`callback` into `Demux`.
- Use a single-item `demux_queue` for events that cannot be batched.

And maybe this would reduce dimensionality in `on_reply`, too?
- More shared code between batching and non-batching operations.
- In `StateMachine`, implement trivial `Demuxer`s for non-batchable operations.

src/state_machine.zig
kprotty: The `results` itself is re-sliced with the found batch offset in the `defer = self.results = self.results[demuxed..]` below. This achieves a similar effect IIUC?

sentientwaffle: You're correct, thanks!

src/vsr/client.zig
kprotty: Demux should be removed when switching to a Context-based API so could this change be deferred until then?

batiati: It's a small change that simplifies the code, so it's worth including now.
We can benefit from that when migrating to Context-based. Demux will be removed from the vsr. Client, but some equivalent mechanism may be inspired by the current way it works.

src/vsr/client.zig
```diff
@@ -424,28 +632,63 @@ pub fn Client(comptime StateMachine_: type, comptime MessageBus: type) type {
...
+                    );
+
+                    if (!StateMachine.batch_logical_allowed.get(operation)) {
+                        assert(inflight.demux_queue.empty());
+                    }
+
+                    // Push the inflight's demux to the front of the demux_queue to simplify handling.
+                    inflight.demux.next = inflight.demux_queue.out;
+                    inflight.demux_queue.out = &inflight.demux;
+                    inflight.demux_queue.count += 1;
```
batiati: Suggestion: Rename `Request.demux` to something like `Request.demux_queue_head` and add it immediately to the `demux_queue` when creating a new request.
Here, we can just call `demux_queue.pop()` without extra logic.

kprotty: It would be self-referential and `request_queue.pop()` returns a (RLS) copy from its ring buffer, so `demux_queue_head` in the FIFO would need a fixup after that call.

batiati: Sadly true 😅
Well, in this case we have the option of acquiring from the DemuxPool in either cases, what do you think? No queue head required!

I think we can't exceed the DemixPool’s capacity the way it is calculated.

kprotty: It could introduce a failure/unreachable in `raw_request` if `get_message()` doesn't pair well with demux nodes, although  that's highly unlikely. `raw_request` is also only used by testing code (aof/cluster) which should already ensure message limits. Ill go ahead.

src/vsr/client.zig
```diff
@@ -1,10 +1,13 @@
...
 const stdx = @import("../stdx.zig");
-const assert = std.debug.assert;
 const testing = std.testing;
 const maybe = stdx.maybe;
 const mem = std.mem;
 
+noinline fn assert(cond: bool) void {
+    if (!cond) @panic("assert failed");
+}
+
```
batiati: I'm curious about that. 🤓
Why can't we use `const assert = std.debug.assert;` here?

src/vsr/client.zig
```diff
@@ -414,10 +417,10 @@ pub fn Client(comptime StateMachine_: type, comptime MessageBus: type) type {
 
             // Newly made Requests are populated and potentially pushed through VSR if there exists
             // no currently inflight request.
-            // A Batch is newly made if its demux node is the same one inlined in the Request.
-            // NOTE: This sets the batch.request.message.request number.
-            if (batch.demux == &batch.request.demux) {
-                assert(batch.request.demux_queue.empty());
+            // A Batch is newly made if its Request only contains its Demux.
+            // NOTE: This sets the `batch.request.message.request` number.
```
batiati:
```zig
// NOTE: This sets the `batch.request.message.request` number.
```
I don't think this comment is up to date, the request number was already set during `batch_get`.

Also, we can swap the order of `self.register()` with the `header` initialization, so we don't need to use `.request = undefined` when initializing the header anymore.

src/vsr/client.zig
```diff
@@ -382,18 +389,13 @@ pub fn Client(comptime StateMachine_: type, comptime MessageBus: type) type {
...
-                .demux = .{
-                    .user_data = undefined, // Set during batch_submit().
-                    .callback = undefined, // Set during batch_submit().
-                    .event_count = @intCast(event_count),
-                    .event_offset = 0,
-                },
             });
 
             const request = self.request_queue.tail_ptr().?;
+            request.demux_queue.push(demux);
```
batiati, line 394: Assert that `demux_queue.count == 0` here.
Also, we can assert that `demux_queue.count >1` above when appending a demux into an existing request.

batiati (commented):
Great job @kprotty 🥳
There are just some minor comments to be addressed here, otherwise LGTM.
I think @sentientwaffle wants to give it one more pass before approval.

src/vsr/client.zig
batiati: >```zig
> // NOTE: This sets the `batch.request.message.request` number.
>```
> I don't think this comment is up to date, the request number was already set during `batch_get`.

In fact, I DO think this comment was correct!

https://github.com/tigerbeetle/tigerbeetle/blob/74312b226bbb0921e006993c6c3298e0a8457bb7/src/vsr/client.zig#L382-L386

It raised my concern about calling `register()` during `batch_get()`.
I think `batch_get()` was supposed to behave like `message_get()`,  and `register()` should only be called by `batch_submit`.

src/vsr/client.zig
kprotty: Oops, forgot to remove debugging code. This was for stack trace on windows (`unreachable` didn't provide any)

src/vsr/client.zig
kprotty: `batch_get()` needs to call `register()` so that the latter's Request is pushed to the `request_queue` first. `batch_get()` pushes to the queue to obtain a `*Request` from the tail. Maybe it could have its own `Request`, populate that, and push to queue on `batch_submit()`? This would mean the following isn't batchable (Should we tighten doc-comment to make this not even a valid use-case?)
```zig
const a = batch_get(.batchable_op, max_events - 1);
const b = batch_get(.batchable_op, 1);
```

batiati: I think the docs are clear enough that `batch_get` must be followed by `batch_submit` immediately, so it's not a valid use case (I can't think of an `assert` here, but it'd be much better than docs).

> he latter's Request is pushed to the `request_queue` first. `batch_get()` pushes to the queue to obtain a `*Request` from the tail. Maybe it could have its own Request, populate that, and push to queue on batch_submit()? 

I remember I asked this same question in the past, and I ended up deferring to enqueue the request to `batch_submit`, holding either a `Message` or a `Request` depending on the case.

https://github.com/tigerbeetle/tigerbeetle/blob/14b7afc2aa389394630d13545a38734e64271d5f/src/vsr/client.zig#L31-L46

But you're right, I was most concerned about keeping it with the same semantics of `message_get`, but I don't think it makes any difference with the intended usage.

Just adjust the outdated comment, and we're good to go :+1:

src/benchmark.zig
```diff
@@ -397,22 +397,17 @@ const Benchmark = struct {
...
-
-        stdx.copy_disjoint(
-            .inexact,
-            u8,
-            message.buffer[@sizeOf(vsr.Header)..],
-            payload,
-        );
+        const event_size = switch (operation) {
+            inline else => |op| @divExact(payload.len, @sizeOf(StateMachine.Event(op))),
+        };
```
sentientwaffle: Shouldn't this be named `event_count`?

src/clients/c/tb_client/echo_client.zig
```diff
@@ -62,32 +68,70 @@ pub fn EchoClient(comptime StateMachine_: type, comptime MessageBus: type) type
...
+                .size = @intCast(@sizeOf(Header) + body_size),
             };
 
-            assert(!self.request_queue.full());
+            return Batch{
+                .message = message_request,
+                .event_count = @intCast(event_count),
+                .demux = demux,
+            };
+        }
```
sentientwaffle: So to check my understanding, the echo client's `batch_get`/`batch_submit` don't actually do any batching, right? i.e. each "batch_get" ends up in a separate request message no matter what.

But then why does `Batch` have a `demux` property at all?

src/clients/c/tb_client/echo_client.zig
```diff
@@ -62,32 +68,70 @@ pub fn EchoClient(comptime StateMachine_: type, comptime MessageBus: type) type
...
-            callback: Self.Request.Callback,
             operation: Self.StateMachine.Operation,
-            message: *Message,
-            message_body_size: usize,
-        ) void {
-            const message_request = message.build(.request);
+            event_count: usize,
+        ) Self.BatchError!Batch {
+            const body_size = switch (operation) {
+                inline else => |op| @sizeOf(Self.StateMachine.Event(op)) * event_count,
```
sentientwaffle:
```suggestion
                inline else => |operation| @sizeOf(Self.StateMachine.Event(operation)) * event_count,
```

"op" typically refers to an op number, not the state machine operation type.

src/vsr/client.zig
```diff
@@ -196,52 +296,131 @@ pub fn Client(comptime StateMachine_: type, comptime MessageBus: type) type {
...
+                .event_offset = 0,
+            };
+
+            // Check-in with the StateMachine to see if this operation can even be batched.
+            // If so, find an existing Request with the same Op that has room in its Message.
+            // The request must not be the one currently inflight in VSR as its Message is sealed.
+            if (StateMachine.batch_logical_allowed.get(operation)) {
+                var it = self.request_queue.iterator_mutable();
+                while (it.next_ptr()) |request| {
+                    if (request == self.request_inflight) continue;
```
sentientwaffle: Re: https://github.com/tigerbeetle/tigerbeetle/pull/1257#discussion_r1384075559:

> Need to avoid batching into the head Request of the queue if it's currently inflight (message has already been sent to MessageBus). The implicit contract is indeed kind of a hack.

Could we accomplish this (minus the hack) by just calling `_ = it.next_ptr();` immediately before this loop starts? (With a comment to explain.)

batiati (approved):
Awesome work @kprotty!

PR #1258, opened 2023-10-30
LSM/ManifestNode: Move Label into TableInfo

sentientwaffle:
Context:

The Manifest block schema previously arranged the `Label`s (level + event) and `TableInfo`s in a SoA. This choice dates back to the pre-unified manifest – the schema's `TableInfo` was also used in memory for `ManifestLevel`s. But with the unified manifest, the `ManifestLevel`'s `TableInfo` type is a separate type.

In this PR:

Replace the SoA with AoS -- that is, an array of `TableInfo` structs, each of which has the `Label` embedded within. 

As a bonus simplification: Including the `Label` in `TableInfo` removes the `copy_left` from ManifestLog's block encoding.

Worth mentioning: This change bubbles up to the `ManifestLog`'s API. Previously it had `ManifestLog.{insert,remove,update}(level, table_info)`. But now it is just `ManifestLog.append(table_info)`, since `table_info` includes the insert/remove/update event (and level).

### DIFF
--- a/src/lsm/forest_table_iterator.zig
+++ b/src/lsm/forest_table_iterator.zig
@@ -66,10 +66,7 @@ pub fn ForestTableIteratorType(comptime Forest: type) type {
             break :default iterators;
         },
 
-        pub fn next(iterator: *ForestTableIterator, forest: *const Forest) ?struct {
-            level: u6,
-            table: TableInfo,
-        } {
+        pub fn next(iterator: *ForestTableIterator, forest: *const Forest) ?TableInfo {
             while (iterator.level < constants.lsm_levels) : (iterator.level += 1) {
                 for (iterator.tree_id..Forest.tree_id_range.max + 1) |tree_id_runtime| {
                     iterator.tree_id = @intCast(tree_id_runtime);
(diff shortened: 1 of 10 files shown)


src/lsm/forest_table_iterator.zig
```diff
@@ -84,10 +81,14 @@ pub fn ForestTableIteratorType(comptime Forest: type) type {
...
+                                return table.encode(.{
+                                    .tree_id = tree_id,
                                     .level = iterator.level,
-                                    .table = table.encode(tree_id),
-                                };
+                                    // Dummy event, doesn't really mean anything in this context.
+                                    // (We are reusing the schema's TableInfo type since it is
+                                    // shared by all Tree types.)
+                                    .event = .insert,
+                                });
```
batiati: Could we use instead `.event = .reserved` for cases where a dummy event is required?


Also here:
https://github.com/tigerbeetle/tigerbeetle/blob/aceb8f736526e00834ce660356cc1bf6b936c4c7/src/vsr/replica.zig#L7931

batiati (approved):

PR #1260, opened 2023-11-02
CI: tag the published docker image as 'latest'

matklad:
Another regression introduced in the new release process! The current latest release is the one from a month ago. 

I am not fully familiar with how `:latest` is implemented in docker, so I am not sure that this is the right fix, but I've verified manually that at least passing `--tag` twice doesn't immediately return an error.

### DIFF
--- a/src/scripts/release.zig
+++ b/src/scripts/release.zig
@@ -651,10 +651,16 @@ fn publish_docker(shell: *Shell, info: VersionInfo) !void {
         }
         try shell.exec(
             \\docker buildx build --file tools/docker/Dockerfile . --platform linux/amd64,linux/arm64
-            \\   --tag ghcr.io/tigerbeetle/tigerbeetle:{version}{debug} --push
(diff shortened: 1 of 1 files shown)


KMJ-007:
waiting for this to get merged!

so can continue my work!

matklad:
@KMJ-007 note that, even before this is merged, you can use explicit version of the container, `ghcr.io/tigerbeetle/tigerbeetle:0.14.158` rather than just `ghcr.io/tigerbeetle/tigerbeetle` (which is implicitly latest).

KMJ-007:
cool

cb22 (approved):
> I am not fully familiar with how :latest is implemented in docker, so I am not sure that this is the right fix, but I've verified manually that at least passing --tag twice doesn't immediately return an error.

It's just a special tag by convention - tags in Docker aren't immutable. :upside_down_face:

PR #1261, opened 2023-11-02
docs: move website generation into this repository

matklad:
This PR moves the brains of https://github.com/tigerbeetledb/docs into the main repo, leaving the `docs` repo as a stograge for pre-rendered GitHub pages. 

The main conceptual simplification here is that we no longer need to do this repo clone dance:

* https://github.com/tigerbeetle/tigerbeetle/blob/b65a28d01a02054b5ca1f18ef80a9a82442dbf46/scripts/validate_docs.sh#L9-L21
* https://github.com/tigerbeetle/docs/blob/main/scripts/build_dont_call_this_directly.sh#L8-L19

Originally, I had the ambition to replace the whole `build.sh` with `zig`, but I scaled that back --- our doc generation is still quite messy (esp with the manual link rewrites), and I'd love to, as usual, approach that holistically, instead of just rewriting the current logic in Zig. Notably, the documentation generator we are using, docusaurus, released a new major version like yesterday  (https://docusaurus.io/blog/releases/3.0), which gives me a lot of food for thought, esp coupled with the pretty large package-lock.json we are importing here. But let that simmer for quite a bit!


kprotty (approved):
Code looks good. Want to note that `zig build ci -- --language=node` works locally on my native linux machine but fails with `unable to get absolute bin path: getwd: no such file or directory` in docker on WSL2 for some reason.

matklad:
>Want to note that zig build ci -- --language=node works locally on my native linux machine but fails with unable to get absolute bin path: getwd: no such file or directory in docker on WSL2 for some reason.

Huh, but not really _that_ unexpected --- I've noticed that Zig API are not as robust when used in cross-platform way. As we are not actively testing on WSL, I'd say that's an OK failure to have, and we can debug it lazily (eg, when it actually prevents us from debugging some real issue).

PR #1264, opened 2023-11-03
ci: implement release verification

matklad:
This adds last-mile testing of the thing that was actually released (GitHub release, nmp package, etc). This is too late to _prevent_ issues, but helps to discover things otherwise impossible to discover (eg, some files were lost during upload to GitHub), and also helps with finding divergence of upstream dependencies (e.g., if something breaks on a new version of node js).


src/scripts/ci.zig
```diff
@@ -97,3 +97,44 @@ fn build_docs(shell: *Shell) !void {
...
+    const version = try shell.exec_stdout("./tigerbeetle version --verbose", .{});
+    assert(std.mem.indexOf(u8, version, tag) != null);
+    assert(std.mem.indexOf(u8, version, "ReleaseSafe") != null);
+
+    const tigerbeetle_absolute_path = try std.fs.cwd().realpathAlloc(gpa, "tigerbeetle");
+    defer gpa.free(tigerbeetle_absolute_path);
+
+    inline for (comptime std.enums.values(Language)) |language| {
+        // TODO: these two don't work yet
+        if (language == .go or language == .dotnet) continue;
```
matklad: Yes, these two are broken! Looking into fixing them, but want to do that independently from this work to add basic verification.

src/testing/tmp_tigerbeetle.zig
```diff
@@ -33,24 +33,34 @@ stderr_reader_thread: std.Thread,
...
-    _ = shell.project_root.statFile(tigerbeetle_exe) catch {
-        log.info("building TigerBeetle", .{});
-        try shell.zig("build", .{});
+    if (options.prebuilt == null) {
+        const tigerbeetle_exe = comptime "tigerbeetle" ++ builtin.target.exeFileExt();
 
-        _ = try shell.project_root.statFile(tigerbeetle_exe);
-    };
+        // If tigerbeetle binary does not exist yet, build it.
+        // TODO: just run `zig build run` unconditionally here, when that doesn't do spurious rebuilds.
```
cb22: @matklad tidy.zig complaining about line length here

matklad:
fix for go: https://github.com/tigerbeetle/tigerbeetle/pull/1265

src/testing/tmp_tigerbeetle.zig
matklad: Should be fine now!

matklad:
and for .net!

https://github.com/tigerbeetle/tigerbeetle/pull/1267

src/clients/java/ci.zig
```diff
@@ -33,11 +33,73 @@ pub fn tests(shell: *Shell, gpa: std.mem.Allocator) !void {
...
+        \\        </configuration>
+        \\      </plugin>
+        \\    </plugins>
+        \\  </build>
+        \\
+        \\  <dependencies>
+        \\    <dependency>
+        \\      <groupId>com.tigerbeetle</groupId>
+        \\      <artifactId>tigerbeetle-java</artifactId>
+        \\      <version>0.14.158</version>
```
batiati: Shouldn't we use the targetted version here?

```suggestion
        \\      <version>{version}</version>
```

matklad: Most excellent catch, thanks!

Fun fact, when writing this code I've actually thought that `options: struct` pattern has a drawback that it doesn't warn about unused field, as I made a similar mistake for some other language!

batiati (approved):

PR #1270, opened 2023-11-06
VSR/ClientReplies: Fix "Write" availability race

sentientwaffle:
`ClientReplies.Write`s are managed by the `ClientReplies` itself, not the caller (`Replica`).

The replica acquires `ClientReplies.Write`s in two different paths:

  1. commit: Writing a new client reply as the result of a commit.
  2. repair: Repairing a corrupt client reply, or acquiring a client reply after state sync.

At the beginning of the async commit chain, we wait for a `ClientReplies.Write` to be available (`client_replies.ready(callback)`). If there is one already available, then `.ready()` just waits for `next_tick()` before invoking the callback. But during that "next tick" interval, a reply-repair might arrive and steal the write.

~To fix this, don't allow repairs to steal the last available write.~

To fix this, split `ready()` into a sync and an async method. If there is a `Write` available synchronously, the caller can use it immediately -- no `next_tick` race.

An alternative solution, we could instead make `Replica` the owner of the `Write`s (similar to how most other "context" objects work). This is a more involved change though, and pulls more logic up into the replica. In particular, it adds dimensionality to the async commit change. (Or rather, exposes that dimensionality to the replica, rather than hiding it within `ClientReplies`.

This fails on commit `2ee06059911fced41bbd479ccc5fa260cf350aef` with simulator seed `5538896752928653533`.

### DIFF
--- a/src/vsr/client_replies.zig
+++ b/src/vsr/client_replies.zig
@@ -100,11 +100,7 @@ pub fn ClientRepliesType(comptime Storage: type) type {
         /// Pointers are into `writes`.
         write_queue: WriteQueue = WriteQueue.init(),
 
-        ready_next_tick: Storage.NextTick = undefined,
-        ready_callback: ?union(enum) {
-            next_tick: *const fn (*ClientReplies) void,
-            write: *const fn (*ClientReplies) void,
-        } = null,
+        ready_callback: ?*const fn (*ClientReplies) void = null,
 
         checkpoint_next_tick: Storage.NextTick = undefined,
         checkpoint_callback: ?*const fn (*ClientReplies) void = null,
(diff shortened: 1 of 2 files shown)


matklad:
:thinking: I am not quite happy about the flow we have here: 


```zig
        pub fn ready(
            client_replies: *ClientReplies,
            callback: *const fn (client_replies: *ClientReplies) void,
        ) void {
            assert(client_replies.ready_callback == null);

            if (client_replies.writes.available() > 0) {
                client_replies.ready_callback = .{ .next_tick = callback };
                client_replies.storage.on_next_tick(
                    .vsr,
                    ready_next_tick_callback,
                    &client_replies.ready_next_tick,
                );
            } else {
                // ready_callback will be called the next time a write completes.
                client_replies.ready_callback = .{ .write = callback };
            }
        }

        fn ready_next_tick_callback(next_tick: *Storage.NextTick) void {
            const client_replies = @fieldParentPtr(ClientReplies, "ready_next_tick", next_tick);
            assert(client_replies.writes.available() > 0);

            const callback = client_replies.ready_callback.?.next_tick;
            client_replies.ready_callback = null;
            callback(client_replies);
        }
```


That is still TOCTOU, which feels error prone. Is there some way we can fix this specific TOCTOU without that larger refactor? Some ideas:

* perhaps we can split `ready` into immediate and async parts, so that the call site looks roughly like

    ```zig
                .setup_client_replies => {
                    if (self.client_replies.ready()) self.commit_op()
                    else self.client_replies.when_ready(commit_op_client_replies_ready_callback);
                },
    ```
* in `ready`, could we perhaps just go and `.aquire` a write `if (client_replies.writes.available() > 0) {`, and then release this write immediately in `ready_next_tick_callback`? (so, we don't actually use this write, we just resreve it until the next tick)

matklad:
While we are at it, we have this assert:

```
        pub fn ready( ... ) void {
            assert(client_replies.ready_callback == null);
```

It prevents concurrent calls to `ready` in case we go through the `write_reply_callback` callback. It would be great if we could assert the same thing for the `next_tick` flow. 

That is, if I call `.ready`, that schedule `next_tick`, and then I call `.ready` _again_ before that next tick fires, I'd love to get an assert.

sentientwaffle:
> That is still TOCTOU, which feels error prone. Is there some way we can fix this specific TOCTOU without that larger refactor? 

:+1: Good catch! I went with "split ready into immediate and async parts" -- that way, the acquired `writes` can still be safely iterated/accessed (e.g. by `read_reply_sync`, to check whether we have a particular reply buffered in memory).

> That is, if I call .ready, that schedule next_tick, and then I call .ready again before that next tick fires, I'd love to get an assert.

:thinking: I think this is already the case? `ready_callback` is a union, and it holds a callback in both the "write" path and the "next tick" path.

matklad:
>🤔 I think this is already the case? 

:facepalm: still can't read the code, of course you are correct!

src/constants.zig
```diff
@@ -436,7 +436,8 @@ comptime {
     assert(journal_iops_read_max > 0);
     assert(journal_iops_write_max > 0);
     assert(client_replies_iops_read_max > 0);
-    assert(client_replies_iops_write_max > 0);
+    // Require at least two: one repair, and one commit.
+    assert(client_replies_iops_write_max >= 2);
```
matklad: I think we no longer need this? Now, even with one IOP it should be fine?

sentientwaffle: Good point!

src/constants.zig
```diff
@@ -436,7 +436,7 @@ comptime {
     assert(journal_iops_read_max > 0);
     assert(journal_iops_write_max > 0);
     assert(client_replies_iops_read_max > 0);
-    assert(client_replies_iops_write_max > 0);
+    // Require at least two: one repair, and one commit.
```
matklad: Deleted wrong line?

sentientwaffle: :facepalm: I deleted the line but it should have been an edit to restore the original condition (and remove the comment).

matklad (approved):

PR #1272, opened 2023-11-06
ci: fix wrong working directory when unmarking release as a draft

matklad:
`publish_docs` step changes the dir to that of the docs website, but that directory is removed completly when publish_docs completes.

So we end up in a non-existing working directory, which breaks git.

The proper fix here would be to bite the bullet and add some sort of misuse resitant cd API to shell, but let's just fix the bug for now.

### DIFF
--- a/src/scripts/release.zig
+++ b/src/scripts/release.zig
@@ -447,6 +447,7 @@ fn publish(shell: *Shell, languages: LanguageSet, info: VersionInfo) !void {
     }
 
     if (languages.contains(.zig)) {
+        try shell.project_root.setAsCwd();
         try shell.exec(
             \\gh release edit --draft=false
             \\  {tag}


src/scripts/release.zig
```diff
@@ -447,6 +447,7 @@ fn publish(shell: *Shell, languages: LanguageSet, info: VersionInfo) !void {
     }
 
     if (languages.contains(.zig)) {
+        try shell.project_root.setAsCwd();
```
kprotty, line 450: Should this be inside the node if statement instead?

matklad: So far, functions don't generally try to not mess up with cwd --- rather each function sets the cwd it needs. 

I am thinking about a larger refactor which would enforce the invariant that cwd doesn't change from under your feet, but that's going to be a larger refactor :)

For the code as is, I think doing cd at this point is right --- `gh` is the thing that needs correct `cd`, so it's its job to enforce that.

kprotty (approved):

PR #1274, opened 2023-11-07
docs: simplify website folder layout

matklad:
Originally, we had the `/docs` folder with the rendered website, and the docusaurus stuff at `/` to actually generate `/docs`. Now with docusaurus moved into the main repo, we no longer need this extra level of `/docs` anymore.

On the day of the next release, I'll change the config on the docs repo to serve the content from the root.


kprotty (approved):

matklad:
Realized that this doesn't even need to be a task for the next release, as I can just go and move the docs now.

Once you start automating things, you forget that doing something manually is an option :sweat_smile:

PR #1275, opened 2023-11-07
ci: more robust handling of directories in shell.zig 

matklad:
We use shell.zig for various "scripting" tasks. One problem there is
managing environment --- current directory and environmental variables.

So far, the strategy was to avoid doing anything special about it and
just use std API to mutate global process state. That's not great, for
two reasons:

- First, mutating process state is inherently fragile. For example,
  calling setenv in a multithreaded program is almost always UB

- Second, mutating global state might lead to logical errors. For
  example, we had a bug recently where `foo` called `bar`, `bar` changed
  process working directory, and, upon return, `foo` ended up in an
  unexpected state. (https://github.com/tigerbeetle/tigerbeetle/pull/1272)

To fix the issues, this PR phases our global API in favor of `Shell.cwd`
field.

Trivia: in most other languages, adding a second cwd would be quite
confusing, as that would make the `std` API behave in unexpected ways.
However in Zig the cwd is not ambient, but rather passed explicitly to
almost all APIs:

    std.fs.cwd().readFle("Readme.me")

So there's no confusion!

### DIFF
--- a/src/clients/dotnet/ci.zig
+++ b/src/clients/dotnet/ci.zig
@@ -9,6 +9,7 @@ const Shell = @import("../../shell.zig");
 const TmpTigerBeetle = @import("../../testing/tmp_tigerbeetle.zig");
 
 pub fn tests(shell: *Shell, gpa: std.mem.Allocator) !void {
+    assert(shell.file_exists("TigerBeetle.sln"));
     try shell.exec("dotnet format --verify-no-changes", .{});
 
     // Unit tests.
@@ -24,13 +25,8 @@ pub fn tests(shell: *Shell, gpa: std.mem.Allocator) !void {
 
     // Integration tests.
     inline for (.{ "basic", "two-phase", "two-phase-many" }) |sample| {
-        var sample_dir = try shell.project_root.openDir(
-            "src/clients/dotnet/samples/" ++ sample,
-            .{},
-        );
-        defer sample_dir.close();
-
-        try sample_dir.setAsCwd();
+        try shell.pushd("./samples/" ++ sample);
+        defer shell.popd();
 
         var tmp_beetle = try TmpTigerBeetle.init(gpa, .{});
         defer tmp_beetle.deinit(gpa);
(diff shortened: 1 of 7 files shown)


src/clients/dotnet/ci.zig
```diff
@@ -9,6 +9,7 @@ const Shell = @import("../../shell.zig");
 const TmpTigerBeetle = @import("../../testing/tmp_tigerbeetle.zig");
 
 pub fn tests(shell: *Shell, gpa: std.mem.Allocator) !void {
+    assert(shell.file_exists("TigerBeetle.sln"));
```
matklad, line 12: Fun fact: I've added these asserts as a drive by change, because why not, and they've already caught a bug in this PR!

src/scripts/release.zig
```diff
@@ -132,8 +132,6 @@ fn build_tigerbeetle(shell: *Shell, info: VersionInfo, dist_dir: std.fs.Dir) !vo
     var section = try shell.open_section("build tigerbeetle");
     defer section.close();
 
-    try shell.project_root.setAsCwd();
```
matklad, line 135: Before, we were defensively setting up cwd, because it could have been changed from under our feet. With pushd/popd we can rather assume that the cwd is reasonable, and just assert that when appropriate.

src/shell.zig
```diff
@@ -31,13 +33,19 @@ arena: std.heap.ArenaAllocator,
...
 /// on the first access, but, given that we always use `Shell` in the context of our repository,
 /// eager initialization is more ergonomic.
 project_root: std.fs.Dir,
 
+/// Shell's logical cwd which is used for all functions in this file. It might be different from
+/// `std.fs.cwd()` and is set to `project_root` on init.
+cwd: std.fs.Dir,
+
+// Stack of working directories backing pushd/popd.
+cwd_stack: [cwd_stack_max]std.fs.Dir,
```
matklad, line 46: Open-coded bounded array here:

* std.BoundedArray is banned in favor of stdx.BoundedArray
* but, as we use shell.zig in build.zig, we don't want  to drag any deps here
* and, also, std.BoundedArray doesn't really by us all that much.

src/shell.zig
```diff
@@ -302,9 +356,14 @@ pub fn exec_status_ok(shell: *Shell, comptime cmd: []const u8, cmd_args: anytype
 
     try expand_argv(&argv, cmd, cmd_args);
 
+    // TODO(Zig): use cwd_dir once that is available https://github.com/ziglang/zig/issues/5190
+    var buffer: [std.fs.MAX_PATH_BYTES]u8 = undefined;
+    const cwd_path = try shell.cwd.realpath(".", &buffer);
+
```
matklad, line 362: Not happy about all these copy-paste here, but also not sure if its worth to try to dry it out -- it's hard to move stack-allocated `bufer` to a helper. Hopefully, `cwd_dir` will be available in the next release.

kprotty (approved):

PR #1277, opened 2023-11-07
ci: smoke-run simulator on CI

matklad:
Normally, a seed is specified as a base-10 integer. However, as a special case, we allow using a Git hash (a hex string 40 character long). This is used by our CI, which passes current commit hash as a seed --- that way, we run simulator on CI, we run it with different, "random" seeds, but the the failures remain reproducible just from the commit hash!

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -213,15 +213,13 @@ jobs:
       - run: zig/zig build build_benchmark_segmented_array
       - run: zig/zig build -Dtracer-backend=tracy
 
-  # This is just a canary to make sure that the simulator compiles
-  # It would be a good idea to also _run_ a single iteration,
-  # but that currently has some false failures :-)
-  # TODO: Make so we have a single zig build simulator that does both
+  # Run simulator once for each state machine, using commit hash as a random, but also deterministic
+  # seed.
   simulate:
     if: inputs.version == ''
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
       - run: ./scripts/install_zig.sh
-      - run: zig/zig build simulator -Dsimulator-state-machine=accounting
-      - run: zig/zig build simulator -Dsimulator-state-machine=testing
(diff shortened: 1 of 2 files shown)


.github/workflows/linux.yml
```diff
@@ -213,15 +213,12 @@ jobs:
...
   simulate:
     if: inputs.version == ''
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
       - run: ./scripts/install_zig.sh
-      - run: zig/zig build simulator -Dsimulator-state-machine=accounting
-      - run: zig/zig build simulator -Dsimulator-state-machine=testing
+      - run: zig/zig build simulator_run -Dsimulator-state-machine=accounting -Doptimize=ReleaseSafe -- ${{ github.sha }}
+      - run: zig/zig build simulator_run -Dsimulator-state-machine=testing    -Doptimize=ReleaseSafe -- ${{ github.sha }}
```
sentientwaffle, line 225: :thinking: Should we use the PR number instead? That way you can't "fix" a failing seed by just force-pushing with a different commit message.

matklad: I think I like commit hash more, as it is more hermetic. It is a bit more fragile, but hopefully everyone is diligent enough to create a follow-up issue anyway!

matklad: Hermetic in a sense that you only need commit hash for a repro: with PR number, you still need commit hash, but you also need PR, which also is a tie to GitHub

.github/workflows/linux.yml
```diff
@@ -213,15 +213,12 @@ jobs:
       - run: zig/zig build build_benchmark_segmented_array
       - run: zig/zig build -Dtracer-backend=tracy
 
-  # This is just a canary to make sure that the simulator compiles
-  # It would be a good idea to also _run_ a single iteration,
-  # but that currently has some false failures :-)
-  # TODO: Make so we have a single zig build simulator that does both
+  # Run simulator once, using commit hash as a random, but also deterministic seed.
```
sentientwaffle: Nit: once (per state machine)

sentientwaffle (dismissed):

sentientwaffle (approved):

PR #1281, opened 2023-11-08
build: squash warning from ldd

matklad:
This gets rid of

ldd: warning: you do not have execution permission for `/nix/store/njjfhnvg90czp2jz37dkfyrhr97rvlii-openjdk-19.0.2+7/lib/openjdk/lib/server/libjvm.so'

warning

I've also spend some time trying to figure out how to get rid of ldd here altogether, and didnt' find a good way :-)

I think it is true that, on a given system, both musl and glibc jvms might be present, and that we need to inspect the .so to figure out the ABI, and that we need to do this during "configure" phase of our build process.

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -401,10 +401,13 @@ pub fn build(b: *std.Build) !void {
                 // On Linux, detects the abi by calling `ldd` to check if
                 // the libjvm.so is linked against libc or musl.
                 // It's reasonable to assume that ldd will be present.
-                const ldd_result = b.exec(&.{
-                    "ldd",
-                    b.pathJoin(&.{ libjvm_path, "libjvm.so" }),
-                });
+                var exit_code: u8 = undefined;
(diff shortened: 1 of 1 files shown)


batiati:
> I think it is true that, on a given system, both musl and glibc jvms might be present, and that we need to inspect the .so to figure out the ABI

Yes, it is possible but most unexpected. Do you have something simpler to probe the ABI? We also use this same strategy to load the correct binary at runtime for both on Node and Dotnet, except we inspect the running process instead of the libjvm

matklad:
Nope, I guess the most relevant project to learn from is https://github.com/mapbox/node-pre-gyp, but what they do is determining the ABI of the "system" (https://github.com/lovell/detect-libc/blob/main/lib/detect-libc.js), which feels even more fragile than what we do.

batiati (approved):

PR #1289, opened 2023-11-08
node: streamline scripts for building node 

matklad:
It seems like just `prepare` should be enough for everything.


src/clients/node/scripts/prepare.js
```diff
@@ -28,5 +30,33 @@ fs.writeFileSync(defFile, 'EXPORTS\n    ' + allSymbolsArr.join('\n    '))
 execSync(`${zig} dlltool -m i386:x86-64 -D node.exe -d ${defFile} -l ${libFile}`)
 
 // Build the tigerbeetle node client.
-process.chdir(path.resolve('../../../'))
```
kprotty, line 31: Is this script called in the context of `src/clients/node`? If so, wondering how `zig build` is able to run below.

matklad: We get the path to zig as `const zig = path.resolve('../../../zig/zig' + (isWindows ? '.exe' : ''))` so that is an aboslute path. And `zig build` crawls the directory tree upwards, looking for build.zig. 

The following works for me locally :

```
$  cd src/clients/node/ && ../../../zig/zig build node_client
```

kprotty (approved):

PR #1294, opened 2023-11-09
ci: pass GITHUB_TOKEN for release validation

matklad:
We use `gh` to download releases, and, somewhat surprisingly, it needs a token even for public operations like fetching a release.


matklad:
https://github.com/tigerbeetle/tigerbeetle/actions/runs/6814006694/job/18529866314#step:8:14

Both `GH_` and `GITHUB_` work, and we already use the latter in `release.yml`

sentientwaffle (approved):

PR #1295, opened 2023-11-09
VSR: 256-byte headers; Header specialization

sentientwaffle:
## 256-byte headers

The extra space is mostly just reserved space for now. (See "Next Steps".)

#### Aside:

Counter-intuitively, when we double the header size, we can _halve_ the minimum WAL size.
This is because the minimum journal size is driven by the redundant headers -- we need at least 2 sectors of them.
Doubling the header size means those 2 sectors hold half as many headers. (See `journal_slot_count_min`).

Thus, a VOPR run of the same number of commits will checkpoint twice as often!

### Context

Presently, we have a single header type (`vsr.Header`) which serves as the header of all message types (i.e. "commands") – including all grid block types.

Some fields are common across all message types – e.g. `checksum` or `cluster`.
But not all message types need to store the same information.

Consider `Header.context`:

```
const Header = extern struct {
    ...
    /// The checksum of the message to which this message refers.
    ///
    /// We use this cryptographic context in various ways, for example:
    ///
    /// * A `request` sets this to the client's session number.
    /// * A `reply` sets this to a "stable" checksum.
    ///   Replies for a specific op may have different views (and checksums),
    ///   but are guaranteed to have the same context.
    /// * A `prepare` sets this to the checksum of the client's request.
    /// * A `prepare_ok` sets this to the checksum of the prepare being acked.
    /// * A `commit` sets this to the checksum of the latest committed prepare.
    /// * A `do_view_change` sets this to a bitset, with set bits indicating headers
    ///   in the message body which it has definitely not prepared (i.e. "nack").
    ///   The corresponding header may be an actual prepare header, or it may be a "blank" header.
    /// * A `request_start_view` sets this to a unique nonce, to detect outdated SVs.
    /// * A `start_view` sets this to zero for a new view, and to a nonce from an RSV when
    ///   responding to the RSV.
    /// * A `request_prepare` sets this to the checksum of the prepare being requested.
    /// * A `request_reply` sets this to the checksum of the reply being requested.
    /// * A `sync_free_set` sets this to the complete FreeSet's checksum.
    /// * A `sync_client_sessions` sets this to the complete ClientSessions's checksum.
    ///
    /// This allows for cryptographic guarantees beyond request, op, and commit numbers, which have
    /// low entropy and may otherwise collide in the event of any correctness bugs.
    context: u128 = 0,
    ...
};
```

The field's name applies to `prepare` and `reply`. All of the other message types reuse the field because they need a `u128`. The documentation specifying these uses easily gets out of sync, though. (Grid blocks also use `context` – and each block type encodes data into `context` with a different schema.)

### Solution

- Some fields are shared by all (or most) messages – these are called the _frame_ fields.
- `Header` still exists as a "generic" header (it may hold any message type). But it holds the frame fields, and then a `reserved_command: [...]u8` slice of bytes which are command-specific.
- For each `Command`, add a new `extern struct` header type:
    - Exposed via `Header.Type(command)`.
    - The frame fields are identical in all header types (in both type and offset).
    - In place of `reserved_command`, each header type defines its own specific fields. For example, `Header.Type(.request_prepare)` has a `prepare_checksum` field instead of `context`.
- There is a corresponding `Message.Type(command)`, which has `message.header: *Header.Type(command)`. This means that most functions (e.g. message handlers) don't need to cast headers at all.
- For grid blocks, there is a `Header.Type(.block)` shared by all block types, with a `block_type: BlockType` field (no more reusing `Operation`!) and `reserved_block` bytes, which are decoded by each block schema (see `schema.zig`).
- To avoid duplicating methods like `set_checksum()` onto each header type, every header type has a `frame()` and `frame_const()` method, which casts to a base `Header`.
- Casts from the base header type to a specific header use one of:
    - `pub fn into(self: *Header, comptime command: Command) ?*Type(command)`
        - Returns `null` if the command doesn't match `header.command`.
        - The caller almost always writes `header.into(...).?` -- it makes the assertion explicit at the call site.
    - `pub fn into_const(self: *const Header, comptime command: Command) ?*const Type(command)`
    - `pub fn into_any(self: *const Header) AnyHeaderPointer`
        - (returns a `union(Command) { command → *const Type(command) }`).
- Note that there are some frame fields that are unused by some message types. (e.g. `view`, `replica`).

### Polymorphism

This approach raises the question of how to handle message polymorphism – code which operates over multiple specific header/message types.

On a case-by-case basis:

- WAL prepares and reserved prepares: Instead of using `command=reserved` for reserved prepares, use `command=prepare operation=reserved`. No more polymorphism needed here.
- DVC prepares and gaps: Same as the WAL.
- Superblock trailer sync messages: These are going to removed anyway, so for now the split just uses code generation for each command.
- Different block types: Each block type uses `Header.Type(.block)`. Some fields (e.g. `address`, `block_type`) are shared by all block types. For fields that are block-type-specific (e.g. "how many values are in this data block"), there is `metadata_bytes: [...]u8`, which is cast to block-type-specific `Metadata` structs.

### Misc

Also included in this commit:

- Remove many default fields from header types – now that we can simply remove extraneous fields from headers that don't use them, default values are a bit of a footgun (since you can forget to set a required field).
- Change `Header.version` from a `u8` to a `u16`.
- In the `Header(*).invalid()` functions, when we are verifying that a header has no body, we test both `size == @sizeof(Header)` and `checksum_body == checksum(.{})`. Previously we checked the checksum first, but now we check the size first. I _think_ this would provide a little more information if one of them found an error.

The frame/header schemas used in this commit is very much not final, but this diff is big enough already:

## Next Steps

(In follow-up PRs).

- Double the checksum size so that we can use the checksum as an AEAD tag eventually.
- Add checkpoint id to `prepare`s. (As mentioned in the [state sync PR](https://github.com/tigerbeetle/tigerbeetle/pull/834).)

### DIFF
--- a/src/lsm/schema.zig
+++ b/src/lsm/schema.zig
@@ -4,40 +4,18 @@
 //! relevant parameters directly into the block's header. This allows the decoders to not be
 //! generic. This is convenient for compaction, but critical for the scrubber and repair queue.
 //!
-//! Every block begins with a `vsr.Header` that includes:
-//!
-//! * `checksum`, `checksum_body` verify the data integrity.
-//! * `cluster` is the cluster id.
-//! * `command` is `.block`.
-//! * `op` is the block address.
-//! * `size` is the block size excluding padding.
-//!
-//! Index block schema:
-//! │ vsr.Header                   │ operation=BlockType.index,
-//! │                              │ context: schema.TableIndex.Context,
-//! │                              │ parent: schema.TableIndex.Parent,
-//! │                              │ request=@sizeOf(Key)
-//! │                              │ timestamp=snapshot_min
+//! Index block body schema:
 //! │ [data_block_count_max]u128   │ checksums of data blocks
 //! │ [data_block_count_max]Key    │ the minimum/first key in the respective data block
 //! │ [data_block_count_max]Key    │ the maximum/last key in the respective data block
 //! │ [data_block_count_max]u64    │ addresses of data blocks
 //! │ […]u8{0}                     │ padding (to end of block)
 //!
-//! Data block schema:
-//! │ vsr.Header               │ operation=BlockType.data,
-//! │                          │ context: schema.TableData.Context,
-//! │                          │ parent: schema.TableData.Parent,
-//! │                          │ request=values_count
-//! │                          │ timestamp=snapshot_min
+//! Data block body schema:
 //! │ [≤value_count_max]Value  │ At least one value (no empty tables).
 //! │ […]u8{0}                 │ padding (to end of block)
 //!
-//! ManifestNode block schema:
-//! │ vsr.Header                  │ operation=BlockType.manifest
-//! │                             │ context: schema.ManifestNode.Context
-//! │                             │ parent=previous_manifest_block.checksum else 0
-//! │                             │ commit=previous_manifest_block.address else 0
+//! ManifestNode block body schema:
 //! │ [entry_count]TableInfo      │
 //! │ […]u8{0}                    │ padding (to end of block)
 const std = @import("std");
@@ -58,70 +36,53 @@ const block_body_size = block_size - @sizeOf(vsr.Header);
 const BlockPtr = *align(constants.sector_size) [block_size]u8;
 const BlockPtrConst = *align(constants.sector_size) const [block_size]u8;
 
-pub inline fn header_from_block(block: BlockPtrConst) *const vsr.Header {
-    const header = mem.bytesAsValue(vsr.Header, block[0..@sizeOf(vsr.Header)]);
+pub inline fn header_from_block(block: BlockPtrConst) *const vsr.Header.Type(.block) {
+    const header = mem.bytesAsValue(vsr.Header.Type(.block), block[0..@sizeOf(vsr.Header)]);
     assert(header.command == .block);
-    assert(header.op > 0);
+    assert(header.address > 0);
     assert(header.size >= @sizeOf(vsr.Header)); // Every block has a header.
     assert(header.size > @sizeOf(vsr.Header)); // Every block has a non-empty body.
     assert(header.size <= block.len);
-    assert(BlockType.valid(header.operation));
-    assert(BlockType.from(header.operation) != .reserved);
+    assert(header.block_type.valid());
+    assert(header.block_type != .reserved);
     return header;
 }
 
 /// A block's type is implicitly determined by how its address is stored (e.g. in the index block).
 /// BlockType is an additional check that a block has the expected type on read.
 ///
-/// The BlockType is stored in the block's `header.operation`.
+/// The BlockType is stored in the block's `header.block_type`.
 pub const BlockType = enum(u8) {
-    /// Unused; verifies that no block is written with a default 0 operation.
+    /// Unused; verifies that no block is written with a default 0 block type.
     reserved = 0,
 
     manifest = 1,
     index = 2,
     data = 3,
 
-    pub fn valid(vsr_operation: vsr.Operation) bool {
-        _ = std.meta.intToEnum(BlockType, @intFromEnum(vsr_operation)) catch return false;
+    pub fn valid(block_type: BlockType) bool {
+        _ = std.meta.intToEnum(BlockType, @intFromEnum(block_type)) catch return false;
 
         return true;
-    }
-
-    pub inline fn from(vsr_operation: vsr.Operation) BlockType {
-        return @as(BlockType, @enumFromInt(@intFromEnum(vsr_operation)));
-    }
-
-    pub inline fn operation(block_type: BlockType) vsr.Operation {
-        return @as(vsr.Operation, @enumFromInt(@intFromEnum(block_type)));
     }
 };
 
 pub const TableIndex = struct {
-    /// Stored in every index block's header's `context` field.
+    /// Stored in every index block's header's `metadata_bytes` field.
     ///
     /// The max-counts are stored in the header despite being available (per-tree) at comptime:
     /// - Encoding schema parameters enables schema evolution.
     /// - Tables can be decoded without per-tree specialized decoders.
     ///   (In particular, this is useful for the scrubber and the grid repair queue).
-    pub const Context = extern struct {
+    pub const Metadata = extern struct {
         data_block_count: u32,
         data_block_count_max: u32,
-        padding: [8]u8 = [_]u8{0} ** 8,
-
-        comptime {
-            assert(@sizeOf(Context) == @sizeOf(u128));
-            assert(stdx.no_padding(Context));
-        }
-    };
-
-    pub const Parent = extern struct {
+        key_size: u32,
         tree_id: u16,
-        reserved: [14]u8 = [_]u8{0} ** 14,
+        reserved: [130]u8 = [_]u8{0} ** 130,
 
         comptime {
-            assert(@sizeOf(Parent) == @sizeOf(u128));
-            assert(stdx.no_padding(Parent));
+            assert(stdx.no_padding(Metadata));
         }
     };
 
@@ -187,26 +148,27 @@ pub const TableIndex = struct {
     pub fn from(index_block: BlockPtrConst) TableIndex {
         const header = header_from_block(index_block);
         assert(header.command == .block);
-        assert(BlockType.from(header.operation) == .index);
-        assert(header.op > 0);
-        assert(header.timestamp > 0);
+        assert(header.block_type == .index);
+        assert(header.address > 0);
+        assert(header.snapshot > 0);
 
-        const context = @as(Context, @bitCast(header.context));
-        assert(context.data_block_count <= context.data_block_count_max);
+        const header_metadata = metadata(index_block);
 
         return TableIndex.init(.{
-            .key_size = header.request,
-            .data_block_count_max = context.data_block_count_max,
+            .key_size = header_metadata.key_size,
+            .data_block_count_max = header_metadata.data_block_count_max,
         });
     }
 
-    pub fn tree_id(index_block: BlockPtrConst) u16 {
+    pub fn metadata(index_block: BlockPtrConst) *const Metadata {
         const header = header_from_block(index_block);
-        assert(BlockType.from(header.operation) == .index);
+        assert(header.command == .block);
+        assert(header.block_type == .index);
 
-        const parent = @as(Parent, @bitCast(header.parent));
-        assert(stdx.zeroed(&parent.reserved));
-        return parent.tree_id;
+        const header_metadata = std.mem.bytesAsValue(Metadata, &header.metadata_bytes);
+        assert(header_metadata.data_block_count <= header_metadata.data_block_count_max);
+        assert(stdx.zeroed(&header_metadata.reserved));
+        return header_metadata;
     }
 
     pub inline fn data_addresses(index: *const TableIndex, index_block: BlockPtr) []u64 {
@@ -246,35 +208,24 @@ pub const TableIndex = struct {
     }
 
     pub inline fn data_blocks_used(index: *const TableIndex, index_block: BlockPtrConst) u32 {
-        const header = header_from_block(index_block);
-        const context = @as(Context, @bitCast(header.context));
-        const value = @as(u32, @intCast(context.data_block_count));
-        assert(value > 0);
-        assert(value <= index.data_block_count_max);
-        return value;
+        const header_metadata = metadata(index_block);
+        assert(header_metadata.data_block_count > 0);
+        assert(header_metadata.data_block_count <= index.data_block_count_max);
+        return header_metadata.data_block_count;
     }
 };
 
 pub const TableData = struct {
-    /// Stored in every data block's header's `context` field.
-    pub const Context = extern struct {
+    /// Stored in every data block's header's `metadata_bytes` field.
+    pub const Metadata = extern struct {
         value_count_max: u32,
+        value_count: u32,
         value_size: u32,
-        padding: [8]u8 = [_]u8{0} ** 8,
-
-        comptime {
-            assert(@sizeOf(Context) == @sizeOf(u128));
-            assert(stdx.no_padding(Context));
-        }
-    };
-
-    pub const Parent = extern struct {
         tree_id: u16,
-        reserved: [14]u8 = .{0} ** 14,
+        reserved: [130]u8 = [_]u8{0} ** 130,
 
         comptime {
-            assert(@sizeOf(Parent) == @sizeOf(u128));
-            assert(stdx.no_padding(Parent));
+            assert(stdx.no_padding(Metadata));
         }
     };
 
@@ -289,7 +240,12 @@ pub const TableData = struct {
     padding_offset: u32,
     padding_size: u32,
 
-    pub fn init(parameters: Context) TableData {
+    pub const Parameters = struct {
+        value_count_max: u32,
+        value_size: u32,
+    };
+
+    pub fn init(parameters: Parameters) TableData {
         assert(parameters.value_count_max > 0);
         assert(parameters.value_size > 0);
         assert(std.math.isPowerOfTwo(parameters.value_size));
@@ -315,20 +271,30 @@ pub const TableData = struct {
     pub fn from(data_block: BlockPtrConst) TableData {
         const header = header_from_block(data_block);
         assert(header.command == .block);
-        assert(BlockType.from(header.operation) == .data);
-        assert(header.op > 0);
-        assert(header.timestamp > 0);
+        assert(header.block_type == .data);
+        assert(header.address > 0);
+        assert(header.snapshot > 0);
 
-        return TableData.init(@as(Context, @bitCast(header.context)));
+        const header_metadata = metadata(data_block);
+
+        return TableData.init(.{
+            .value_count_max = header_metadata.value_count_max,
+            .value_size = header_metadata.value_size,
+        });
     }
 
-    pub fn tree_id(data_block: BlockPtrConst) u16 {
+    pub fn metadata(data_block: BlockPtrConst) *const Metadata {
         const header = header_from_block(data_block);
-        assert(BlockType.from(header.operation) == .data);
+        assert(header.command == .block);
+        assert(header.block_type == .data);
 
-        const parent = @as(Parent, @bitCast(header.parent));
-        assert(stdx.zeroed(&parent.reserved));
-        return parent.tree_id;
+        const header_metadata = std.mem.bytesAsValue(Metadata, &header.metadata_bytes);
+        assert(header_metadata.value_size > 0);
+        assert(header_metadata.value_count > 0);
+        assert(header_metadata.value_count <= header_metadata.value_count_max);
+        assert(header_metadata.tree_id > 0);
+        assert(stdx.zeroed(&header_metadata.reserved));
+        return header_metadata;
     }
 
     pub inline fn block_values_bytes(
@@ -350,9 +316,9 @@ pub const TableData = struct {
         data_block: BlockPtrConst,
     ) []align(16) const u8 {
         const header = header_from_block(data_block);
-        assert(BlockType.from(header.operation) == .data);
+        assert(header.block_type == .data);
 
-        const used_values: u32 = header.request;
+        const used_values: u32 = metadata(data_block).value_count;
         assert(used_values > 0);
         assert(used_values <= schema.value_count_max);
 
@@ -364,7 +330,7 @@ pub const TableData = struct {
 };
 
 /// A Manifest block's body is an array of TableInfo entries.
-// TODO Store timestamp (snapshot) in header.
+// TODO Store snapshot in header.
 pub const ManifestNode = struct {
     const entry_size = @sizeOf(TableInfo);
 
@@ -388,14 +354,15 @@ pub const ManifestNode = struct {
         assert(@sizeOf(TableInfo) % alignment == 0);
     }
 
-    /// Stored in every manifest block's header's `context` field.
-    pub const Context = extern struct {
+    /// Stored in every manifest block's header's `metadata_bytes` field.
+    pub const Metadata = extern struct {
+        previous_manifest_block_checksum: u128,
+        previous_manifest_block_address: u64,
         entry_count: u32,
-        reserved: [12]u8 = .{0} ** 12,
+        reserved: [116]u8 = .{0} ** 116,
 
         comptime {
-            assert(@sizeOf(Context) == @sizeOf(u128));
-            assert(stdx.no_padding(Context));
+            assert(stdx.no_padding(Metadata));
         }
     };
 
@@ -439,22 +406,28 @@ pub const ManifestNode = struct {
     entry_count: u32,
 
     pub fn from(manifest_block: BlockPtrConst) ManifestNode {
+        const header_metadata = metadata(manifest_block);
+        return .{ .entry_count = header_metadata.entry_count };
+    }
+
+    pub fn metadata(manifest_block: BlockPtrConst) *const Metadata {
         const header = header_from_block(manifest_block);
         assert(header.command == .block);
-        assert(BlockType.from(header.operation) == .manifest);
-        assert(header.op > 0);
+        assert(header.block_type == .manifest);
+        assert(header.address > 0);
 
-        const context = @as(Context, @bitCast(header.context));
-        assert(context.entry_count > 0);
-        assert(context.entry_count <= entry_count_max);
-        assert(context.entry_count == @divExact(header.size - @sizeOf(vsr.Header), entry_size));
-        assert(stdx.zeroed(&context.reserved));
+        const header_metadata = std.mem.bytesAsValue(Metadata, &header.metadata_bytes);
+        assert(header_metadata.entry_count > 0);
+        assert(header_metadata.entry_count <= entry_count_max);
+        assert(header_metadata.entry_count ==
+            @divExact(header.size - @sizeOf(vsr.Header), entry_size));
+        assert(stdx.zeroed(&header_metadata.reserved));
 
-        if (header.commit == 0) {
-            assert(header.parent == 0);
+        if (header_metadata.previous_manifest_block_address == 0) {
+            assert(header_metadata.previous_manifest_block_checksum == 0);
         }
 
-        return .{ .entry_count = context.entry_count };
+        return header_metadata;
     }
 
     /// Note that the returned block reference is no longer be part of the manifest if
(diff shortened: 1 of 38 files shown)


src/vsr/client.zig
```diff
@@ -329,15 +331,10 @@ pub fn Client(comptime StateMachine_: type, comptime MessageBus: type) type {
...
 
-        fn on_pong_client(self: *Self, pong: *const Message) void {
+        fn on_pong_client(self: *Self, pong: Message.Type(.pong_client)) void {
             assert(pong.header.command == .pong_client);
             assert(pong.header.cluster == self.cluster);
 
-            if (pong.header.client != 0) {
-                log.debug("{}: on_pong: ignoring (client != 0)", .{self.id});
-                return;
-            }
```
sentientwaffle, line 339: This check is unnecessary (even before this PR). We discard `pong_client` messages with a nonzero `client` as `invalid()` in `on_message().

src/vsr/replica.zig
```diff
@@ -2041,17 +2058,15 @@ pub fn ReplicaType(
             const response = self.message_bus.get_message();
             defer self.message_bus.unref(response);
 
-            response.header.* = .{
+            response.build(.headers).header.* = .{
                 .command = .headers,
-                // We echo the context back to the replica so that they can match up our response:
-                .context = message.header.context,
```
sentientwaffle, line 2047: This does not appear to have ever been used – `request_headers` always set the outgoing `context` to `0`.

src/vsr/message_header.zig
```diff
@@ -0,0 +1,1430 @@
...
+    replica: u8,
+    reserved_frame: [12]u8 = [_]u8{0} ** 12,
+
+    /// A bitset of "present" prepares. If a bit is set, then the corresponding header is not
+    /// "blank", the replica has the prepare, and the prepare is not known to be faulty.
+    present_bitset: u128,
+    /// A bitset, with set bits indicating headers in the message body which it has definitely not
+    /// prepared (i.e. "nack"). The corresponding header may be an actual prepare header, or it may
+    /// be a "blank" header.
+    nack_bitset: u128,
```
sentientwaffle: We could pick a different bitset size now that we are no longer constrained by reusing fields. But 128 bits seems fine. :shrug: 

(This means that 128 is the absolute maximum pipeline size that TB could be could be configured for.)

src/vsr/replica.zig
```diff
@@ -3878,125 +3927,44 @@ pub fn ReplicaType(
...
-        /// Construct a SV/DVC message, including attached headers from the current log_view.
-        ///
+        /// Construct a SV message, including attached headers from the current log_view.
         /// The caller owns the returned message, if any, which has exactly 1 reference.
-        fn create_view_change_message(self: *Self, command: Command, nonce: ?u128) *Message {
-            assert(self.status == .normal or self.status == .view_change);
-            assert((self.status == .normal) == (command == .start_view));
-            assert((self.status == .view_change) == (command == .do_view_change));
-            assert((command == .do_view_change) == (nonce == null));
+        fn create_start_view_message(self: *Self, nonce: u128) Message.Type(.start_view) {
```
sentientwaffle: `start_view` and `do_view_change` messages have a lot less in common than they used to (due to DVC bitsets, &c), and keeping the code paths intertwined was more confusing than helpful, so I split them up. The DVC side now lives in `send_do_view_change()`. (I inlined it there instead of creating a `create_do_view_change_message` function to avoid duplicating a bunch of assertions.)

src/vsr/replica.zig
```diff
@@ -9189,14 +9338,15 @@ const PipelineQueue = struct {
...
+    fn prepare_by_prepare_ok(pipeline: *PipelineQueue, ok: Message.Type(.prepare_ok)) ?*Prepare {
         assert(ok.header.command == .prepare_ok);
 
         const prepare = pipeline.prepare_by_op_and_checksum(
             ok.header.op,
-            ok.header.context,
+            ok.header.prepare_checksum,
         ) orelse return null;
         assert(prepare.message.header.command == .prepare);
+        assert(prepare.message.header.parent == ok.header.parent);
```
sentientwaffle, line 9354: `parent` was temporarily stolen (by state sync) to fit the checkpoint id in replies. But now we can fit both, so we can go back to echoing the corresponding `prepare`'s `parent`.

src/vsr/message_header.zig
```diff
@@ -0,0 +1,1430 @@
...
+const checksum_body_empty: u128 = 0x49F174618255402DE6E7E3C40D60CC83;
+
+test "empty checksum" {
+    assert(checksum_body_empty == vsr.checksum(&.{}));
+}
+
+/// Network message and journal entry header:
+/// We reuse the same header for both so that prepare messages from the primary can simply be
+/// journalled as is by the backups without requiring any further modification.
+pub const Header = extern struct {
```
sentientwaffle, line 23: Currently planned changes to frame fields:

- Change all checksums to `u256`. (For future AEAD tags.)
- Change nonce to `u256`. (For future AEAD nonce.)
- Change `cluster` to `u128`. (I still don't understand the reasoning here, but we have room so why not. :shrug:)

src/lsm/schema.zig
```diff
@@ -4,40 +4,18 @@
 //! relevant parameters directly into the block's header. This allows the decoders to not be
 //! generic. This is convenient for compaction, but critical for the scrubber and repair queue.
 //!
-//! Every block begins with a `vsr.Header` that includes:
```
sentientwaffle, line 7: Removing the header descriptions -- now that the `Metadata` struct is simple + uniform, I'd prefer that it be the "source of truth" for block header schemas.

src/vsr/message_header.zig
jorangreef: Great!

A 128-bit `cluster` also improves our "resolution" for interoperability with other systems that want to use UUIDs for cluster ID to allow for random generation of cluster ID without the cost of having to introduce a centralized ID coordinator. Plus it gives us a further security guarantee that clusters can't talk to other clusters unless they can literally “guess” the identifier, which is now cryptographically "tough" (granted, we might not be able to imagine now why this could be important, but it's a stronger property nevertheless—especially from an online verification point of view).

(The motivation for AEADs moving to nonces bigger than 96-bit... is also why 128-bit is a “bare minimum” for cluster ID for us).

src/vsr/message_header.zig
```diff
@@ -0,0 +1,1430 @@
...
+        return checksum_value;
+    }
+
+    /// This must be called only after set_checksum_body() so that checksum_body is also covered:
+    pub fn set_checksum(self: *Header) void {
+        self.checksum = self.calculate_checksum();
+    }
+
+    pub fn set_checksum_body(self: *Header, body: []const u8) void {
+        self.checksum_body = self.calculate_checksum_body(body);
```
matklad, line 128: can we `assert(self.checksum == 0)`?

src/vsr/message_header.zig
```diff
@@ -0,0 +1,1430 @@
...
+        if (self.command != command) return null;
+        return std.mem.bytesAsValue(Type(command), std.mem.asBytes(self));
+    }
+
+    /// Returns null if all fields are set correctly according to the command, or else a warning.
+    /// This does not verify that checksum is valid, and expects that this has already been done.
+    pub fn invalid(self: *const Header) ?[]const u8 {
+        if (self.version != vsr.Version) return "version != Version";
+        if (self.size < @sizeOf(Header)) return "size < @sizeOf(Header)";
+        if (self.epoch != 0) return "epoch != 0";
```
matklad, line 169: Missing `self.reserved_nonce == 0` check

src/vsr/message_header.zig
```diff
@@ -0,0 +1,1430 @@
...
+                assert(@offsetOf(CommandHeader, command_field.name) ==
+                    @offsetOf(Header, header_field.name));
+            }
+        }
+    }
+}
+
+/// This type isn't ever actually a constructed, but makes Type() simpler by providing a header type
+/// for each command.
+const Reserved = extern struct {
```
matklad: Given that this is never cosntructed, shouldn't `invalid` be unreachable (or always return `"reserved is invalid"`)? If we need a real reserved here, we should also check view, cluster, replica and such.

src/vsr/message_header.zig
```diff
@@ -0,0 +1,1430 @@
...
+        return std.mem.bytesAsValue(Header, std.mem.asBytes(header));
+    }
+
+    fn invalid(self: *const @This()) ?[]const u8 {
+        assert(self.command == .ping);
+        if (self.size != @sizeOf(Header)) return "size != @sizeOf(Header)";
+        if (self.checksum_body != checksum_body_empty) return "checksum_body != expected";
+        if (self.view != 0) return "view != 0";
+        if (self.ping_timestamp_monotonic == 0) return "ping_timestamp_monotonic != expected";
+        if (!stdx.zeroed(&self.reserved)) return "reserved != 0";
```
matklad: Could check that `checkpoint_op` is valid.

src/vsr/message_header.zig
```diff
@@ -0,0 +1,1430 @@
...
+        return std.mem.bytesAsValue(Header, std.mem.asBytes(header));
+    }
+
+    fn invalid(self: *const @This()) ?[]const u8 {
+        assert(self.command == .ping_client);
+        if (self.size != @sizeOf(Header)) return "size != @sizeOf(Header)";
+        if (self.checksum_body != checksum_body_empty) return "checksum_body != expected";
+        if (self.replica != 0) return "replica != 0";
+        if (self.view != 0) return "view != 0";
+        if (self.client == 0) return "client == 0";
```
matklad: Curiously, we create client id as:

```
context.client_id = std.crypto.random.int(u128);
```

I am still unsure whether this should check for zero explicitly...

src/vsr/message_header.zig
```diff
@@ -0,0 +1,1430 @@
...
+    reserved_frame: [12]u8,
+
+    /// This data's schema is different depending on the `Header.command`.
+    /// (No default value – `Header`s should not be constructed directly.)
+    reserved_command: [176]u8,
+
+    comptime {
+        assert(@sizeOf(Header) == 256);
+        assert(stdx.no_padding(Header));
+        assert(@offsetOf(Header, "reserved_command") % @alignOf(u128) == 0);
```
matklad:
```suggestion
        assert(@offsetOf(Header, "reserved_command") % 16 == 0);
```

`@alignOf(u128) == 8`

matklad: I wonder if we should patch `no_padding` to check also that all fields are aligned at least to their sizeof.

src/vsr/message_header.zig
```diff
@@ -0,0 +1,1430 @@
...
+                        return "size != @sizeOf(Header) + @sizeOf(ReconfigurationRequest)";
+                    }
+                } else if (@intFromEnum(self.operation) < constants.vsr_operations_reserved) {
+                    return "operation is reserved";
+                }
+                // Thereafter, the client must provide the session number:
+                // These requests should set `parent` to the `checksum` of the previous reply.
+                if (self.session == 0) return "session == 0";
+                if (self.request == 0) return "request == 0";
+            },
```
matklad: Let's add comment here that we check that the operation is a valid enum variant in replia, as that requires knowing the StateMachine type.

src/vsr/message_header.zig
```diff
@@ -0,0 +1,1430 @@
...
+    cluster: u32,
+    size: u32 = @sizeOf(Header),
+    epoch: u32 = 0,
+    view: u32,
+    version: u16 = vsr.Version,
+    command: Command,
+    replica: u8 = 0,
+    reserved_frame: [12]u8 = [_]u8{0} ** 12,
+
+    /// A backpointer to the previous request or prepare checksum for hash chain verification.
```
matklad: This now only points to other prepares.

src/vsr/message_header.zig
```diff
@@ -0,0 +1,1430 @@
...
+                if (self.parent != 0) return "reserved: parent != 0";
+                if (self.client != 0) return "reserved: client != 0";
+                if (self.context != 0) return "reserved: context != 0";
+                if (self.request != 0) return "reserved: request != 0";
+                if (self.view != 0) return "reserved: view != 0";
+                if (self.commit != 0) return "reserved: commit != 0";
+                if (self.timestamp != 0) return "reserved: timestamp != 0";
+                if (self.size != @sizeOf(Header)) return "reserved: size != @sizeOf(Header)";
+                if (self.replica != 0) return "reserved: replica != 0";
+                if (self.checksum_body != checksum_body_empty) {
```
matklad: Le'ts sort this in order of fields.

matklad: and maybe add explicit maybe for op

src/vsr/message_header.zig
```diff
@@ -0,0 +1,1430 @@
...
+    }
+
+    pub fn frame_const(header: *const @This()) *const Header {
+        return std.mem.bytesAsValue(Header, std.mem.asBytes(header));
+    }
+
+    fn invalid(self: *const @This()) ?[]const u8 {
+        assert(self.command == .prepare_ok);
+        if (self.checksum_body != checksum_body_empty) return "checksum_body != expected";
+        if (self.size != @sizeOf(Header)) return "size != @sizeOf(Header)";
```
matklad: Flip the order to be consistent with other cases

src/vsr/message_header.zig
```diff
@@ -0,0 +1,1430 @@
...
+    size: u32 = @sizeOf(Header),
+    epoch: u32 = 0,
+    view: u32,
+    version: u16 = vsr.Version,
+    command: Command,
+    replica: u8,
+    reserved_frame: [12]u8 = [_]u8{0} ** 12,
+
+    checkpoint_id: u128,
+    parent: u128,
```
matklad: Comment what parent points to, it can be confused with prepare_checksum

src/vsr/message_header.zig
```diff
@@ -0,0 +1,1430 @@
...
+    pub fn frame_const(header: *const @This()) *const Header {
+        return std.mem.bytesAsValue(Header, std.mem.asBytes(header));
+    }
+
+    fn invalid(self: *const @This()) ?[]const u8 {
+        assert(self.command == .commit);
+        if (self.size != @sizeOf(Header)) return "size != @sizeOf(Header)";
+        if (self.checksum_body != checksum_body_empty) return "checksum_body != expected";
+        if (self.timestamp_monotonic == 0) return "timestamp_monotonic == 0";
+        if (!stdx.zeroed(&self.reserved)) return "reserved != 0";
```
matklad: I think we could compare commit and checkpoint_op?

src/vsr/message_header.zig
```diff
@@ -0,0 +1,1430 @@
...
+    present_bitset: u128,
+    /// A bitset, with set bits indicating headers in the message body which it has definitely not
+    /// prepared (i.e. "nack"). The corresponding header may be an actual prepare header, or it may
+    /// be a "blank" header.
+    nack_bitset: u128,
+    op: u64,
+    /// Set to `commit_min`, to indicate the sending replica's progress.
+    /// The sending replica may continue to commit after sending the DVC.
+    commit_min: u64,
+    checkpoint_op: u64,
```
matklad: I guess we could check checkpoint_op <= commit_min <= op ?

src/vsr/message_header.zig
```diff
@@ -0,0 +1,1430 @@
...
+    replica: u8,
+    reserved_frame: [12]u8 = [_]u8{0} ** 12,
+
+    /// Set to zero for a new view, and to a nonce from an RSV when responding to the RSV.
+    nonce: u128,
+    op: u64,
+    /// Set to `commit_min`/`commit_max` (they are the same).
+    commit: u64,
+    /// The replica's `op_checkpoint`.
+    checkpoint_op: u64,
```
matklad: ditto

src/vsr/message_header.zig
```diff
@@ -0,0 +1,1430 @@
...
+
+    pub fn frame_const(header: *const @This()) *const Header {
+        return std.mem.bytesAsValue(Header, std.mem.asBytes(header));
+    }
+
+    fn invalid(self: *const @This()) ?[]const u8 {
+        assert(self.command == .request_start_view);
+        if (self.size != @sizeOf(Header)) return "size != @sizeOf(Header)";
+        if (self.checksum_body != checksum_body_empty) return "checksum_body != expected";
+        if (!stdx.zeroed(&self.reserved)) return "reserved != 0";
```
matklad: check that nonce != 0. We do requrie it to be non-zero here:

```
        /// Initializes and opens the provided replica using the options.
        pub fn open(self: *Self, parent_allocator: std.mem.Allocator, options: OpenOptions) !void {
            assert(options.storage_size_limit <= constants.storage_size_max);
            assert(options.storage_size_limit % constants.sector_size == 0);
            assert(options.nonce != 0);
```

src/vsr/message_header.zig
```diff
@@ -0,0 +1,1430 @@
...
+        return std.mem.bytesAsValue(Header, std.mem.asBytes(header));
+    }
+
+    pub fn frame_const(header: *const @This()) *const Header {
+        return std.mem.bytesAsValue(Header, std.mem.asBytes(header));
+    }
+
+    fn invalid(self: *const @This()) ?[]const u8 {
+        assert(self.command == .request_headers);
+        if (self.size != @sizeOf(Header)) return "size != @sizeOf(Header)";
```
matklad: makes me wonder if we should check that the view is zero here and for RequestPrepare (and for RequestReply)? I _think_ we are not actually using RequestHeader's view for anything?

src/vsr/message_header.zig
```diff
@@ -0,0 +1,1430 @@
...
+        return std.mem.bytesAsValue(Header, std.mem.asBytes(header));
+    }
+
+    pub fn frame_const(header: *const @This()) *const Header {
+        return std.mem.bytesAsValue(Header, std.mem.asBytes(header));
+    }
+
+    fn invalid(self: *const @This()) ?[]const u8 {
+        assert(self.command == .eviction);
+        if (self.size != @sizeOf(Header)) return "size != @sizeOf(Header)";
```
matklad: check that client != 0

src/vsr/message_header.zig
```diff
@@ -0,0 +1,1430 @@
...
+        return std.mem.bytesAsValue(Header, std.mem.asBytes(header));
+    }
+
+    pub fn frame_const(header: *const @This()) *const Header {
+        return std.mem.bytesAsValue(Header, std.mem.asBytes(header));
+    }
+
+    fn invalid(self: *const @This()) ?[]const u8 {
+        assert(self.command == .do_view_change);
+        if (!stdx.zeroed(&self.reserved)) return "reserved != 0";
```
matklad: Could check for size multiple, like we do for RequestBlocks.

matklad:
`Message.Type(.request)` looks a bit ugly given its frequency. Maybe reshuffle the code so that just `Message(.request)` works? (and `Message(.any)` for when we want to be polymorphic. I _think_ we don't have to use the literal `.command` enum there). 

Not feeling too strongly here though.

src/vsr/message_header.zig
sentientwaffle: It looks like `Client.init()` verifies `assert(id > 0);`. So maybe in `context.zig` we should just check whether our generated id is zero, and if so, regenerate it.

On the other hand, if _I_ generated a `0` from a cryptographically-random u128, I would probably panic too... :stuck_out_tongue: :four_leaf_clover:

src/vsr/message_header.zig
sentientwaffle: No, sometimes (albeit rarely) we call `set_checksum_body`/`set_checksum` on a head that already has its checksum set:

In `on_request_repeat_reply`:

```zig
            if (entry.header.size == @sizeOf(Header)) {
                const reply = self.create_message_from_header(@bitCast(entry.header)).into(.reply).?;
                defer self.message_bus.unref(reply.base);

                self.send_reply_message_to_client(reply);
                return;
            }
```

(`create_message_from_header` computes the checksum, which in most call sites hasn't been computed yet.)

src/message_pool.zig
```diff
@@ -65,6 +66,34 @@ comptime {
 /// initialization and reused thereafter. The messages_max values determine the size of this pool.
 pub const MessagePool = struct {
     pub const Message = struct {
+        pub const Reserved = MessageType(.reserved);
```
matklad, line 69: Hm, it seems that `Message` and  `Message.Prepare` differ in the level of indirection: `Message.Type` is a pointer to `Message`. This is not ideal, I think:

* first, because there's a difference between a raw and a type message
* second, because it adds one more level of indirection. Eg, in aof, when reading
  
  ```
  pub fn to_message(self: *AOFEntry, target: Message.Prepare) void {
  ```

  there's now extra effort required to reason "wait, why we are passing a message by value? Ahh, that's because a message is secretly a pointer!"

src/message_pool.zig
```diff
@@ -65,6 +66,34 @@ comptime {
...
+        pub const RequestBlocks = MessageType(.request_blocks);
+        pub const Block = MessageType(.block);
+        pub const RequestSyncCheckpoint = MessageType(.request_sync_checkpoint);
+        pub const SyncCheckpoint = MessageType(.sync_checkpoint);
+        pub const RequestSyncFreeSet = MessageType(.request_sync_free_set);
+        pub const RequestSyncClientSessions = MessageType(.request_sync_client_sessions);
+        pub const SyncFreeSet = MessageType(.sync_free_set);
+        pub const SyncClientSessions = MessageType(.sync_client_sessions);
+
         // TODO: replace this with a header() function to save memory
```
matklad: This is mostly unrelated to the PR, but I think there's a better solution here, which also gets rid of an extra level of indirection:

```zig
pub const MessageRc = extern struct align(constant.sector_size) {
    padding: [constant.sector_size - @sizeOf(u32) - @sizeOf(?*Message)]u8 = undefined,
    references: u32 = 0,
    next: ?*Message,
    message: Message,
};

pub const Message = extern struct {
    header: Header,
    body: [constants.message_size_max - @sizeOf(Header)]u8,
};
```

We'll still be passing around a `*Message`, but it'll now be a pointer directly into message bytes (as opposed to a pointer to  a pointer to a body). `message.header` and `message.body` would be direct references into the underlying memory (reducing the need to slice by `sizeOf(Header)`), and `mem.asBytes(message)` would give the raw memory of the entire message. 

The reference count and next link would be accessed through `@fieldParentPtr`. We'll lose some memory for left-padding the rc wrapper to sector boundary, but overall it feels like it should be a win, as we have less indirection.

matklad: alternatively, we could just tack `references` and `next` to the end of the message, after the body, but then `mem.asBytes(message)` won't work.

sentientwaffle: > TODO: replace this with a header() function to save memory

Doesn't the `padding` field in `MessageRc` undo any memory won from not needing the `header` pointer in `Message`? I don't understand what this design would buy us.

src/message_pool.zig
sentientwaffle: :thinking: The `MessageType(command)` types could cast the `Message` memory (instead of holding a pointer). Then the `MessageType(command)`s would be pointers too.

Thoughts? Fwiw I didn't go that route originally since it seems like extra "magic". Plus `Message`s aren't `extern struct`, so I don't think their layouts are actually guaranteed to be equivalent. (Though in practice I assume they would be.)

matklad: > Then the MessageType(command)s would be pointers too.

Yeah, I think we should do this: I don't like "hiding" pointers in structs, by-value call site just looks too weird. 

Good point about extern structs! I think we could make message an extern one?

src/message_pool.zig
matklad: Yeah, the improvement here is not in saving memory, but in:

* removing the indirection (right now, `message.header.replica` is two pointer dereferences, but it could be one),
* simplifying the call-site, as `.body` is a field access, rather than a function call,

It will also:

* maaybe improving const-correcness, as `*const Message` would mean "immutable message", rather than "mutable message with immutable reference count" we have today. 
* reduce the _number_ of allocations, as we no longer need to allocate rc+pointers in a separate (tiny) allocation.

src/lsm/schema.zig
```diff
@@ -58,70 +36,53 @@ const block_body_size = block_size - @sizeOf(vsr.Header);
...
-    pub const Parent = extern struct {
+        key_size: u32,
         tree_id: u16,
-        reserved: [14]u8 = [_]u8{0} ** 14,
+        reserved: [130]u8 = [_]u8{0} ** 130,
 
         comptime {
-            assert(@sizeOf(Parent) == @sizeOf(u128));
-            assert(stdx.no_padding(Parent));
+            assert(stdx.no_padding(Metadata));
```
matklad, line 85: could assert that the size is 144

src/lsm/schema.zig
```diff
@@ -315,20 +271,30 @@ pub const TableData = struct {
...
+        assert(header.block_type == .data);
+        assert(header.address > 0);
+        assert(header.snapshot > 0);
+
+        const header_metadata = metadata(data_block);
 
-        return TableData.init(@as(Context, @bitCast(header.context)));
+        return TableData.init(.{
+            .value_count_max = header_metadata.value_count_max,
+            .value_size = header_metadata.value_size,
```
matklad, line 284: Do we want to assert at this point that values don't go beyond header.size?

src/lsm/table.zig
```diff
@@ -307,20 +307,20 @@ pub fn TableType(
 
                 const block = builder.data_block;
                 const header = mem.bytesAsValue(vsr.Header, block[0..@sizeOf(vsr.Header)]);
-                header.* = .{
+                header.* = @bitCast(vsr.Header.Block{
```
matklad: Can we do

```
const header = mem.bytesAsValue(vsr.Header.Block, block[0..@sizeOf(vsr.Header)]);
```

to avoid bitcast here and elsewhere (this seems to be a fairly frequent pattern.

src/lsm/schema.zig
sentientwaffle: I'll add this to `metadata()`:

```
        assert(@sizeOf(vsr.Header) + header_metadata.value_size * header_metadata.value_count ==
            header.size);
```

src/vsr/replica_format.zig
```diff
@@ -113,13 +112,12 @@ fn ReplicaFormatType(comptime Storage: type) type {
...
 
-                for (std.mem.bytesAsSlice(Header, wal_buffer[0..size])) |*header| {
+                for (std.mem.bytesAsSlice(Header.Prepare, wal_buffer[0..size])) |*header| {
                     assert(header.valid_checksum());
-                    if (header.op == 0) {
-                        assert(header.command == .prepare);
-                        assert(header.operation == .root);
+
+                    if (header.operation == .root) {
+                        assert(header.op == 0);
```
matklad: Shouldn't we rather `assert((header.operation == .root) == (header.op == 0))`?

src/vsr/replica.zig
```diff
@@ -9063,13 +9216,13 @@ fn message_body_as_view_headers(message: *const Message) vsr.Headers.ViewChangeS
 
 /// Asserts that the headers are in descending op order.
 /// The headers may contain gaps and/or breaks.
-fn message_body_as_headers(message: *const Message) []const Header {
+fn message_body_as_prepare_headers(message: *const Message) []const Header.Prepare {
```
matklad, line 9219: In the spirit of pushing complexity to the definiton site, might take in an `anytype` here and comptime assert that the type is right, primarily to get rid of .base() calls at the call-site. Not sure it is worth it though.

sentientwaffle: I lean towards leaving these functions as-is -- they aren't used as often as (e.g.) `unref()`.
`anytype` cleans up the call site, but at the cost of worse type signatures (for both readability and type errors).

matklad (approved):

matklad:
Awesome refactor @sentientwaffle! I especially like the block refactors, removing those `const address = message.op` is sooooo good!

PR #1298, opened 2023-11-13
ci: skip release validation for failed releases

matklad:
We already get a message when release fails, no need to duplicate it.


cb22:
Ah took me a sec to figure out what was happening here. Does making the `if` on the `validate:` job play nicely with the `if` on the `alert_failure`?

matklad:
Good question! I don't know, but I expect it to be fine. I think the status for skipped job would be "skipped", so `contains(needs.*.result, 'failure')` would evaluate to false.

cb22 (approved):

PR #1302, opened 2023-11-14
Fixes the Node.js client validation messages.

### LINKED ISSUE #1290: Node library throws unhelpful error if your `code`s are too big
If you submit an `Account` or `Transfer` with a `code` that's bigger than a 16-bit integer, the library will throw an unhelpful error:

```
tigerbeetle-schema-gen/node_modules/tigerbeetle-node/dist/index.js:59
                binding.submit(context, operation, batch, (error, result) => {
                        ^

Error: code
    at tigerbeetle-schema-gen/node_modules/tigerbeetle-node/dist/index.js:59:25
    at new Promise (<anonymous>)
    at request (tigerbeetle-schema-gen/node_modules/tigerbeetle-node/dist/index.js:57:16)
    at Object.createAccounts (tigerbeetle-schema-gen/node_modules/tigerbeetle-node/dist/index.js:77:40)
    at file:///tigerbeetle-schema-gen/dist/index.js:23:34
    at ModuleJob.run (node:internal/modules/esm/module_job:194:25)

Node.js v18.17.1
```

It would be useful for the library to check the sizes of integer fields and return specific errors if you exceed the allowable size.

batiati:
Fixes the Node.js client validation messages.

We added an extra sentinel to the field's name `field.name ++ "\x00"` causing the message to be incorrectly concatenated like `"code\0x00 must be a u16."` and therefore printed as just `"code"` at Node's side.

Closes #1290

### DIFF
--- a/src/clients/node/src/node.zig
+++ b/src/clients/node/src/node.zig
@@ -372,8 +372,7 @@ fn decode_array(comptime Event: type, env: c.napi_env, array: c.napi_value, even
                     const value = try @field(translate, @typeName(FieldInt) ++ "_from_object")(
                         env,
                         object,
-                        // Use `comptimePrint` to ensure the sentinel terminated string.
-                        std.fmt.comptimePrint("{s}", .{field.name}),
+                        add_trailing_null(field.name),
                     );
 
                     if (std.mem.eql(u8, field.name, "timestamp") and value != 0) {
@@ -440,6 +439,15 @@ fn encode_array(comptime Result: type, env: c.napi_env, results: []const Result)
     }
 
     return array;
+}
+
+fn add_trailing_null(comptime input: []const u8) [:0]const u8 {
+    // Concatenating `[]const u8` with an empty string `[0:0]const u8`,
+    // gives us a null-terminated string `[:0]const u8`.
+    const output = input ++ "";
+    comptime assert(output.len == input.len);
+    comptime assert(output[output.len] == 0);
+    return output;
 }
 
 /// Each packet allocates enough room to hold both its Events and its Results.
--- a/src/clients/node/src/node.zig
+++ b/src/clients/node/src/node.zig
@@ -372,7 +372,8 @@ fn decode_array(comptime Event: type, env: c.napi_env, array: c.napi_value, even
                     const value = try @field(translate, @typeName(FieldInt) ++ "_from_object")(
                         env,
                         object,
-                        @ptrCast(field.name ++ "\x00"),
+                        // Use `comptimePrint` to ensure the sentinel terminated string.
+                        std.fmt.comptimePrint("{s}", .{field.name}),
                     );
 
                     if (std.mem.eql(u8, field.name, "timestamp") and value != 0) {
--- a/src/clients/node/src/translate.zig
+++ b/src/clients/node/src/translate.zig
@@ -92,7 +92,7 @@ pub fn slice_from_object(
     comptime key: [:0]const u8,
 ) ![]const u8 {
     var property: c.napi_value = undefined;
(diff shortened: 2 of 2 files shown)


src/clients/node/src/node.zig
```diff
@@ -372,7 +372,8 @@ fn decode_array(comptime Event: type, env: c.napi_env, array: c.napi_value, even
                     const value = try @field(translate, @typeName(FieldInt) ++ "_from_object")(
                         env,
                         object,
-                        @ptrCast(field.name ++ "\x00"),
+                        // Use `comptimePrint` to ensure the sentinel terminated string.
+                        std.fmt.comptimePrint("{s}", .{field.name}),
```
batiati: It's not clear if `field.name` already contains a sentinel, since it's a `comptime` string, it should, but just removing the `++ "\x00"` didn't work.

matklad:
Hm, does this actually fix the issue? I think the following code is still problematic: 

```zig
pub fn u128_from_value(env: c.napi_env, value: c.napi_value, comptime name: [:0]const u8) !u128 {

    throw(env, name ++ " must be a BigInt"),
}
```

It semes that `++` to `[:0]` slice doesn't do what we want it to do:


```zig
const std = @import("std");

pub fn main() void {
    const a: []const u8 = "x";
    const b: [:0]const u8 = a ++ "\x00";
    const c = b ++ "lol";
    for (c, 0..) |byte, index| {
        std.debug.print("c[{}] = {}\n", .{index, byte});
    }
}
```

```console
λ ./zig/zig run main.zig
c[0] = 120
c[1] = 0  <- problem!
c[2] = 108
c[3] = 111
c[4] = 108
```

In general, `comptime [:0]const u8` feel strange to me a bit. Maybe thread through just `comptime []const u8` and append `\x00` at the latest possible moment (when calling napi)?

matklad:
Whoa, wait, 

```
const std = @import("std");

pub fn main() void {
    const a: []const u8 = "x";
    const b: [:0]const u8 = std.fmt.comptimePrint("{s}", .{a});
    const c = b ++ "lol";
    for (c, 0..) |byte, index| {
        std.debug.print("c[{}] = {}\n", .{ index, byte });
    }
}
```

```
c[0] = 120
c[1] = 108
c[2] = 111
c[3] = 108
```

what happens here at all?

matklad:
Ah, so apparently `name ++ ""` is how one converts from `comptime []` to `comptime [:0]`. That makes sense! The type of `""` is `*const [0:0]u8`, that is, it already is null-terminated. So, just `"\x00"` already has two nulls.

src/clients/node/src/node.zig
matklad: As an alternative, consider defining a helper

```zig
fn add_trailing_nul(comptime s: []const u8) [:0] const u8 {
    const result = s ++ "";
    assert(result.len == s.len);
    assert(result[result.len] = 0;
    return result;
}
```

That `++ ""` looks mystifying, but I think it's simpler than `comptimePrint`, and the signature & asserts hopefully make it clear what's going on here!

batiati: Yeah, we just added two sentinels. Comptime concat ++ knows how to deal with sentinels when adding two [:0], but in our case the sentinel was part of the string!

I didn't like ++ "" very much, as you said, too mystical. But yes, with a good comment it seems reasonable.

matklad (approved):

src/clients/node/src/node.zig
matklad: >I didn't like ++ "" very much, as you said, too mystical

My thoughts exactly! But then I figured that I don't actually understand why `comptimePrint` works here any more than I do with `++ ""` --- it is equally mystic, but also has enough stuff going on to serve as a [thought terminator](https://en.wikipedia.org/wiki/Thought-terminating_cliché).

PR #1303, opened 2023-11-16
superblock: clarify what storage_size means

matklad:
- remove `SuperBlock.storage_size` which can be confused with `SuperBlockHeader.storage_size` and is a comptime constant anyway.

  The former is the size of the supreblock itself, the latter is the size of the entire datafile.

- remove `SuperBlock.storage_offset` for the same reason.

- Add docs and maybes to `SuperBlock.storage_size_limit`. After removal of the other two sizes, it is much less confusing, but extra clarification wouldn't hurt!

I was soooo confused by `storage_size` until I realized there are two different ones!

### DIFF
--- a/src/vsr/superblock.zig
+++ b/src/vsr/superblock.zig
@@ -586,12 +586,6 @@ pub fn SuperBlockType(comptime Storage: type) type {
 
         storage: *Storage,
 
-        /// The first logical offset that may be written to the superblock storage zone.
-        storage_offset: u64 = 0,
-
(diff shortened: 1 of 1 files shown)


matklad:
Related work --- with free set in the grid, the storage_size moves from the superblock header to the `CheckpointState`, as it can no longer _always_ be derived from the free set (as the free set needs to be explicitely opened)

sentientwaffle (approved):
:exploding_head: I hadn't realized that `SuperBlock` tracked this separately from the header!

PR #1304, opened 2023-11-16
VSR: Change cluster id from u32 to u128

sentientwaffle:
Rationale (via https://github.com/tigerbeetle/tigerbeetle/pull/1295#discussion_r1389061252):

> A 128-bit cluster also improves our "resolution" for interoperability with other systems that want to use UUIDs for cluster ID to allow for random generation of cluster ID without the cost of having to introduce a centralized ID coordinator. Plus it gives us a further security guarantee that clusters can't talk to other clusters unless they can literally “guess” the identifier, which is now cryptographically "tough" (granted, we might not be able to imagine now why this could be important, but it's a stronger property nevertheless—especially from an online verification point of view).

### DIFF
--- a/src/clients/java/src/client.zig
+++ b/src/clients/java/src/client.zig
@@ -48,7 +48,7 @@ const NativeClient = struct {
     fn client_init(
         comptime echo_client: bool,
         env: *jni.JNIEnv,
-        cluster_id: u32,
+        cluster_id: u128,
         addresses_obj: jni.JString,
         max_concurrency: u32,
     ) ?*Context {
@@ -194,15 +194,20 @@ comptime {
         fn client_init(
             env: *jni.JNIEnv,
             class: jni.JClass,
-            cluster_id: jni.JInt,
+            cluster_id: jni.JByteArray,
             addresses: jni.JString,
             max_concurrency: jni.JInt,
         ) callconv(jni.JNICALL) jni.JLong {
             _ = class;
+            assert(env.get_array_length(cluster_id) == 16);
+
+            const cluster_id_elements = env.get_byte_array_elements(cluster_id, null).?;
+            defer env.release_byte_array_elements(cluster_id, cluster_id_elements, .abort);
+
             var context = NativeClient.client_init(
                 false,
                 env,
-                @as(u32, @bitCast(cluster_id)),
+                @as(u128, @bitCast(cluster_id_elements[0..16].*)),
                 addresses,
                 @as(u32, @bitCast(max_concurrency)),
             );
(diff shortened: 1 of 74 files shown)


sentientwaffle:
Thanks @batiati for the help on this PR!

src/tigerbeetle/main.zig
```diff
@@ -63,10 +63,6 @@ pub fn main() !void {
     }
 }
 
-// Pad the cluster id number and the replica index with 0s
-const filename_fmt = "cluster_{d:0>10}_replica_{d:0>3}.tigerbeetle";
-const filename_len = fmt.count(filename_fmt, .{ 0, 0 });
```
sentientwaffle, line 68: Dead code!

src/clients/java/src/client.zig
```diff
@@ -194,15 +194,20 @@ comptime {
...
+            assert(env.get_array_length(cluster_id) == 16);
+
+            const cluster_id_elements = env.get_byte_array_elements(cluster_id, null).?;
+            defer env.release_byte_array_elements(cluster_id, cluster_id_elements, .abort);
+
             var context = NativeClient.client_init(
                 false,
                 env,
-                @as(u32, @bitCast(cluster_id)),
+                @as(u128, @bitCast(cluster_id_elements[0..16].*)),
```
batiati, line 210: Zig is so beautiful!

src/clients/java/src/main/java/com/tigerbeetle/Client.java
```diff
@@ -25,16 +25,17 @@ public final class Client implements AutoCloseable {
      * @throws InitializationException if an error occurred initializing this client. See
      *         {@link InitializationStatus} for more details.
      *
-     * @throws IllegalArgumentException if {@code clusterID} is negative.
+     * @throws IllegalArgumentException if {@code clusterID} is not a UInt128.
```
batiati, line 29:
```suggestion
     * @throws NullPointerException if {@code clusterID} is null.
     * @throws IllegalArgumentException if {@code clusterID} is not a UInt128.
```

src/clients/java/src/main/java/com/tigerbeetle/Client.java
```diff
@@ -25,16 +25,17 @@ public final class Client implements AutoCloseable {
...
      *         is null.
      * @throws IllegalArgumentException if {@code concurrencyMax} is zero or negative.
      */
-    public Client(final int clusterID, final String[] replicaAddresses, final int concurrencyMax) {
-        if (clusterID < 0)
-            throw new IllegalArgumentException("ClusterID must be positive");
+    public Client(final byte[] clusterID, final String[] replicaAddresses,
+            final int concurrencyMax) {
+        if (clusterID.length != UInt128.SIZE)
+            throw new IllegalArgumentException("ClusterID must be 16 bytes long");
```
batiati, line 40:
```suggestion
        Objects.requireNonNull(clusterID, "ClusterID cannot be null");
        if (clusterID.length != UInt128.SIZE)
            throw new IllegalArgumentException("ClusterID must be 16 bytes long");
```

src/tigerbeetle/cli.zig
```diff
@@ -82,7 +82,7 @@ const CliArgs = union(enum) {
         \\        Print this help message and exit.
         \\
         \\  --cluster=<integer>
-        \\        Set the cluster ID to the provided 32-bit unsigned integer.
+        \\        Set the cluster ID to the provided 128-bit unsigned integer.
```
batiati: We should say which format the u128 is expected (eg decimal, hex, guid).

src/clients/java/src/main/java/com/tigerbeetle/Client.java
batiati: LGTM, just waiting for this one ☝️

sentientwaffle: Ah forgot to commit that! Thank you.

batiati (approved):

PR #1306, opened 2023-11-20
ci: run release validation on the release branch

matklad:
If we run workflow from main, then the there might be a divergence between code in our sample and the current published release.


batiati (approved):
:+1:

PR #1309, opened 2023-11-20
vsr: store free set in the grid

matklad:
Goal: get rid of storage_size_max

Currently, free set is stored as a part of superblock in a trailer. As a result, the worst-case size of the free set affects the layout of the data file. As the worst-case size is proprtional to the size of the grid, this requires limiting the size of the data file up-front, using `storage_size_max` parameter. This is bad operator experience -- it's hard to predict which size would be enough, and failure to allocate enough would be very costly to fix down the line.

The solution to make free set elastic and store it in the grid.

 ## Overview

FreeSetEncoded is the persistent component of the free set. It defines defines the layout of the free set as stored in the grid between checkpoints.

Free set is stored as a linked list of blocks containing EWAH-encoding of a bitset of allocated blocks. The length of the linked list is proportional to the degree of fragmentation, rather that to  size of the data file. The common case is a single block.

The blocks holding free set itself are marked as free in the on-disk encoding, because the number of blocks required to store the compressed bitset becomes known only after encoding.

Linked list is a FIFO. While the blocks are written in the direct order, they have to be read in the reverse order.

 ## Implementation notes

- Separate structs to hold persistent and in-memory components of the free set, as the shape of data is quite different between the two.

- FreeSet Trailer and associated messages are gone.

- `storage_size_max` is _not_ gone. While it's the entire purpose of the work, the actual removal of the parameter can happen separately.

- FreeSetEncoded field is stored as a field in the superblock alongside the free set. This isn't great, as it needs access to the grid, but the superblock is above the grid. As a result, the implementation has a bunch of fairly awkward paths like `set.grid.superblock.free_set`.

  I want to move both free set and free set encoded to be just a part of the grid, but that's for a follow up to not mix code motion with logic changes.

- I noticed that we `@intCast` the encoded bit set size to `u32`. I wonder if we should maybe just store bitset lenght as u64? With u32 we can support a merely 16PiB data file.

- Testing this is awkward. In normal use, the free set size just doesn't
  go beyond a single block, so the entire linked list code is
  effectively unused. To cover this, `fragment_free_set` is added to
  forest fuzz which manually injects fragmentation into the free set to
  make sure it occupies more than one block

### DIFF
--- a/src/lsm/forest_fuzz.zig
+++ b/src/lsm/forest_fuzz.zig
@@ -15,12 +15,15 @@ const Transfer = @import("../tigerbeetle.zig").Transfer;
 const Account = @import("../tigerbeetle.zig").Account;
 const Storage = @import("../testing/storage.zig").Storage;
 const StateMachine = @import("../state_machine.zig").StateMachineType(Storage, constants.state_machine_config);
+const Reservation = @import("../vsr/superblock_free_set.zig").Reservation;
 const GridType = @import("../vsr/grid.zig").GridType;
 const GrooveType = @import("groove.zig").GrooveType;
 const Forest = StateMachine.Forest;
 
 const Grid = GridType(Storage);
 const SuperBlock = vsr.SuperBlockType(Storage);
+const FreeSet = vsr.FreeSet;
+const FreeSetEncoded = vsr.FreeSetEncodedType(Storage);
 
 pub const tigerbeetle_config = @import("../config.zig").configs.fuzz_min;
 
@@ -64,6 +67,9 @@ const Environment = struct {
         .cache_entries_posted = cache_entries_max,
     });
 
+    const free_set_fragments_max = 2048;
+    const free_set_fragment_size = 67;
+
     // We must call compact after every 'batch'.
     // Every `lsm_batch_multiple` batches may put/remove `value_count_max` values per index.
     // Every `FuzzOp.put_account` issues one remove and one put per index.
@@ -82,11 +88,13 @@ const Environment = struct {
         init,
         superblock_format,
         superblock_open,
+        free_set_open,
         forest_init,
         forest_open,
         fuzzing,
         forest_compact,
         forest_checkpoint,
+        free_set_checkpoint,
         superblock_checkpoint,
     };
 
@@ -161,13 +169,37 @@ const Environment = struct {
 
     fn open(env: *Environment) !void {
         env.superblock.open(superblock_open_callback, &env.superblock_context);
-        try env.tick_until_state_change(.superblock_open, .forest_init);
+        try env.tick_until_state_change(.superblock_open, .free_set_open);
+
+        env.superblock.free_set_encoded.open(
+            &env.grid,
+            env.superblock.working.free_set_reference(),
+            free_set_open_callback,
+        );
+        try env.tick_until_state_change(.free_set_open, .forest_init);
 
         env.forest = try Forest.init(allocator, &env.grid, node_count, forest_options);
         env.change_state(.forest_init, .forest_open);
         env.forest.open(forest_open_callback);
 
         try env.tick_until_state_change(.forest_open, .fuzzing);
+
+        env.fragmentate_free_set();
+    }
+
+    /// Allocate a sparse subset of grid blocks to make sure that the encoded free set needs more
+    /// than one block to exercise the block linked list logic from FreeSetEncoded.
+    fn fragmentate_free_set(env: *Environment) void {
+        var reservations: [free_set_fragments_max]Reservation = undefined;
+        for (&reservations) |*reservation| {
+            reservation.* = env.superblock.free_set.reserve(free_set_fragment_size).?;
+        }
+        for (reservations) |reservation| {
+            _ = env.superblock.free_set.acquire(reservation).?;
+        }
+        for (reservations) |reservation| {
+            env.superblock.free_set.forfeit(reservation);
+        }
     }
 
     fn close(env: *Environment) void {
@@ -181,7 +213,13 @@ const Environment = struct {
 
     fn superblock_open_callback(superblock_context: *SuperBlock.Context) void {
         const env = @fieldParentPtr(@This(), "superblock_context", superblock_context);
-        env.change_state(.superblock_open, .forest_init);
+        env.change_state(.superblock_open, .free_set_open);
+    }
+
+    fn free_set_open_callback(set: *FreeSetEncoded) void {
+        const superblock = @fieldParentPtr(SuperBlock, "free_set_encoded", set);
+        const env = @fieldParentPtr(@This(), "superblock", superblock);
+        env.change_state(.free_set_open, .forest_init);
     }
 
     fn forest_open_callback(forest: *Forest) void {
@@ -206,14 +244,10 @@ const Environment = struct {
 
         env.change_state(.fuzzing, .forest_checkpoint);
         env.forest.checkpoint(forest_checkpoint_callback);
-        try env.tick_until_state_change(.forest_checkpoint, .superblock_checkpoint);
-        try env.tick_until_state_change(.superblock_checkpoint, .fuzzing);
-    }
+        try env.tick_until_state_change(.forest_checkpoint, .free_set_checkpoint);
 
-    fn forest_checkpoint_callback(forest: *Forest) void {
-        const env = @fieldParentPtr(@This(), "forest", forest);
-        const op = env.checkpoint_op.?;
-        env.checkpoint_op = null;
+        env.superblock.free_set_encoded.checkpoint(free_set_checkpoint_callback);
+        try env.tick_until_state_change(.free_set_checkpoint, .superblock_checkpoint);
 
         {
             // VSRState.monotonic() asserts that the previous_checkpoint id changes.
@@ -222,24 +256,38 @@ const Environment = struct {
             var reply = std.mem.zeroInit(vsr.Header.Reply, .{
                 .cluster = cluster,
                 .command = .reply,
-                .op = op,
-                .commit = op,
+                .op = env.checkpoint_op.?,
+                .commit = env.checkpoint_op.?,
             });
             reply.set_checksum_body(&.{});
             reply.set_checksum();
 
             _ = env.superblock.client_sessions.put(1, &reply);
         }
-
-        env.change_state(.forest_checkpoint, .superblock_checkpoint);
         env.superblock.checkpoint(superblock_checkpoint_callback, &env.superblock_context, .{
-            .manifest_references = forest.manifest_log.checkpoint_references(),
+            .manifest_references = env.forest.manifest_log.checkpoint_references(),
             .commit_min_checksum = env.superblock.working.vsr_state.checkpoint.commit_min_checksum + 1,
-            .commit_min = op,
-            .commit_max = op + 1,
+            .commit_min = env.checkpoint_op.?,
+            .commit_max = env.checkpoint_op.? + 1,
             .sync_op_min = 0,
             .sync_op_max = 0,
         });
+        try env.tick_until_state_change(.superblock_checkpoint, .fuzzing);
+
+        env.checkpoint_op = null;
+    }
+
+    fn forest_checkpoint_callback(forest: *Forest) void {
+        const env = @fieldParentPtr(@This(), "forest", forest);
+        assert(env.checkpoint_op != null);
+        env.change_state(.forest_checkpoint, .free_set_checkpoint);
+    }
+
+    fn free_set_checkpoint_callback(set: *FreeSetEncoded) void {
+        const superblock = @fieldParentPtr(SuperBlock, "free_set_encoded", set);
+        const env = @fieldParentPtr(@This(), "superblock", superblock);
+        assert(env.checkpoint_op != null);
+        env.change_state(.free_set_checkpoint, .superblock_checkpoint);
     }
 
     fn superblock_checkpoint_callback(superblock_context: *SuperBlock.Context) void {
--- a/src/lsm/schema.zig
+++ b/src/lsm/schema.zig
@@ -56,9 +56,10 @@ pub const BlockType = enum(u8) {
     /// Unused; verifies that no block is written with a default 0 block type.
     reserved = 0,
 
-    manifest = 1,
-    index = 2,
-    data = 3,
+    free_set = 1,
+    manifest = 2,
+    index = 3,
+    data = 4,
 
     pub fn valid(block_type: BlockType) bool {
         _ = std.meta.intToEnum(BlockType, @intFromEnum(block_type)) catch return false;
@@ -340,6 +341,68 @@ pub const TableData = struct {
     }
 };
 
+pub const FreeSetNode = struct {
+    pub const Word = u64;
+
+    pub const Metadata = extern struct {
+        previous_free_set_block_checksum: u128,
+        previous_free_set_block_checksum_padding: u128 = 0,
+        previous_free_set_block_address: u64,
+        reserved: [56]u8 = .{0} ** 56,
+
+        comptime {
+            assert(stdx.no_padding(Metadata));
+            assert(@sizeOf(Metadata) == vsr.Header.Block.metadata_size);
+        }
+    };
+
+    fn metadata(free_set_block: BlockPtrConst) *const Metadata {
+        const header = header_from_block(free_set_block);
+        assert(header.command == .block);
+        assert(header.block_type == .free_set);
+        assert(header.address > 0);
+        assert(header.snapshot == 0);
+
+        const header_metadata = std.mem.bytesAsValue(Metadata, &header.metadata_bytes);
+        assert(header_metadata.previous_free_set_block_checksum_padding == 0);
+        assert(stdx.zeroed(&header_metadata.reserved));
+
+        if (header_metadata.previous_free_set_block_address == 0) {
+            assert(header_metadata.previous_free_set_block_checksum == 0);
+        }
+
+        assert(header.size > @sizeOf(vsr.Header));
+        assert((header.size - @sizeOf(vsr.Header)) % @sizeOf(Word) == 0);
+
+        return header_metadata;
+    }
+
+    pub fn assert_valid_header(free_set_block: BlockPtrConst) void {
+        _ = metadata(free_set_block);
+    }
+
+    pub fn previous(free_set_block: BlockPtrConst) ?struct { checksum: u128, address: u64 } {
+        const header_metadata = metadata(free_set_block);
+
+        if (header_metadata.previous_free_set_block_address == 0) {
+            assert(header_metadata.previous_free_set_block_checksum == 0);
+            return null;
+        } else {
+            return .{
+                .checksum = header_metadata.previous_free_set_block_checksum,
+                .address = header_metadata.previous_free_set_block_address,
+            };
+        }
+    }
+
+    pub fn encoded_words(block: BlockPtrConst) []align(@alignOf(Word)) const u8 {
+        assert_valid_header(block);
+
+        const header = header_from_block(block);
+        return block[@sizeOf(vsr.Header)..header.size];
+    }
+};
+
 /// A Manifest block's body is an array of TableInfo entries.
 // TODO Store snapshot in header.
 pub const ManifestNode = struct {
@@ -429,6 +492,7 @@ pub const ManifestNode = struct {
         assert(header.command == .block);
         assert(header.block_type == .manifest);
         assert(header.address > 0);
+        assert(header.snapshot == 0);
 
         const header_metadata = std.mem.bytesAsValue(Metadata, &header.metadata_bytes);
         assert(header_metadata.entry_count > 0);
--- a/src/vsr/replica.zig
+++ b/src/vsr/replica.zig
@@ -63,6 +63,7 @@ const CommitStage = enum {
     checkpoint_state_machine,
     checkpoint_client_replies,
     checkpoint_grid_repair,
+    checkpoint_free_set,
     checkpoint_superblock,
     /// A commit just finished. Clean up before proceeding to the next.
     cleanup,
@@ -128,6 +129,7 @@ pub fn ReplicaType(
         const Self = @This();
 
         pub const SuperBlock = vsr.SuperBlockType(Storage);
+        const FreeSetEncoded = vsr.FreeSetEncodedType(Storage);
         const Journal = vsr.JournalType(Self, Storage);
         const ClientReplies = vsr.ClientRepliesType(Storage);
         const Clock = vsr.ClockType(Time);
@@ -658,9 +660,13 @@ pub fn ReplicaType(
             maybe(self.status == .recovering_head);
             if (self.status == .recovering) assert(self.solo());
 
-            // Asynchronously open the (Forest inside) StateMachine so that we can repair grid
-            // blocks if necessary:
-            self.state_machine.open(state_machine_open_callback);
+            // Asynchronously open the free set and then the (Forest inside) StateMachine so that we
+            // can repair grid blocks if necessary:
+            self.superblock.free_set_encoded.open(
+                &self.grid,
+                self.superblock.working.free_set_reference(),
+                free_set_open_callback,
+            );
         }
 
         fn superblock_open_callback(superblock_context: *SuperBlock.Context) void {
@@ -675,8 +681,20 @@ pub fn ReplicaType(
             self.opened = true;
         }
 
+        fn free_set_open_callback(free_set_encoded: *FreeSetEncoded) void {
+            const superblock = @fieldParentPtr(SuperBlock, "free_set_encoded", free_set_encoded);
+            const self = @fieldParentPtr(Self, "superblock", superblock);
+            assert(!self.state_machine_opened);
+            assert(self.commit_stage == .idle);
+            assert(self.syncing == .idle);
+            assert(self.sync_tables == null);
+            assert(self.grid_repair_tables.executing() == 0);
+            self.state_machine.open(state_machine_open_callback);
+        }
+
         fn state_machine_open_callback(state_machine: *StateMachine) void {
             const self = @fieldParentPtr(Self, "state_machine", state_machine);
+            assert(self.superblock.free_set.opened);
             assert(!self.state_machine_opened);
             assert(self.commit_stage == .idle);
             assert(self.syncing == .idle);
@@ -1119,9 +1137,7 @@ pub fn ReplicaType(
                 .block => |m| self.on_block(m),
                 .request_sync_checkpoint => |m| self.on_request_sync_checkpoint(m),
                 .sync_checkpoint => |m| self.on_sync_checkpoint(m),
-                .request_sync_free_set => self.on_request_sync_trailer(message),
                 .request_sync_client_sessions => self.on_request_sync_trailer(message),
-                .sync_free_set => self.on_sync_trailer(.free_set, message),
                 .sync_client_sessions => self.on_sync_trailer(.client_sessions, message),
                 // A replica should never handle misdirected messages intended for a client:
                 .pong_client, .eviction => {
@@ -2454,14 +2470,11 @@ pub fn ReplicaType(
         }
 
         fn on_request_sync_trailer(self: *Self, message: *Message) void {
-            assert(message.header.command == .request_sync_free_set or
-                message.header.command == .request_sync_client_sessions);
+            assert(message.header.command == .request_sync_client_sessions);
             if (self.ignore_sync_request_message(message)) return;
 
             switch (message.header.into_any()) {
-                inline .request_sync_client_sessions,
-                .request_sync_free_set,
-                => |message_header, command| {
+                inline .request_sync_client_sessions => |message_header, command| {
                     assert(message_header.checkpoint_op ==
                         self.superblock.staging.vsr_state.checkpoint.commit_min);
                     assert(message_header.checkpoint_id == self.superblock.staging.checkpoint_id());
@@ -2485,9 +2498,7 @@ pub fn ReplicaType(
             if (self.ignore_sync_response_message(message)) return;
 
             switch (message.header.into_any()) {
-                inline .sync_client_sessions,
-                .sync_free_set,
-                => |message_header| {
+                inline .sync_client_sessions => |message_header| {
                     const stage: *SyncStage.RequestingTrailers = &self.syncing.requesting_trailers;
                     assert(stage.target.checkpoint_id == message_header.checkpoint_id);
 
@@ -3155,6 +3166,7 @@ pub fn ReplicaType(
                     self.client_replies.checkpoint(commit_op_checkpoint_client_replies_callback);
                 },
                 .checkpoint_grid_repair => self.commit_op_checkpoint_grid_repair(),
+                .checkpoint_free_set => self.commit_op_checkpoint_free_set(),
                 .checkpoint_superblock => self.commit_op_checkpoint_superblock(),
                 .cleanup => self.commit_op_cleanup(),
                 .idle => assert(self.commit_prepare == null),
@@ -3547,10 +3559,31 @@ pub fn ReplicaType(
             assert(self.commit_stage == .checkpoint_grid_repair);
             assert(self.commit_prepare.?.header.op == self.op);
 
+            self.commit_dispatch(.checkpoint_free_set);
+        }
+
+        fn commit_op_checkpoint_free_set(self: *Self) void {
+            assert(self.commit_stage == .checkpoint_free_set);
+            assert(self.commit_prepare.?.header.op == self.op);
+            assert(self.superblock.free_set.opened);
+
+            self.superblock.free_set_encoded.checkpoint(commit_op_checkpoint_free_set_callback);
+        }
+
+        fn commit_op_checkpoint_free_set_callback(set: *SuperBlock.FreeSetEncoded) void {
+            const superblock = @fieldParentPtr(SuperBlock, "free_set_encoded", set);
+            const self = @fieldParentPtr(Self, "superblock", superblock);
+            assert(self.commit_stage == .checkpoint_free_set);
+            assert(self.commit_prepare.?.header.op == self.op);
+            assert(self.superblock.free_set.opened);
+            assert(self.superblock.free_set.count_released() ==
+                self.superblock.free_set_encoded.block_count);
+
             self.commit_dispatch(.checkpoint_superblock);
         }
 
         fn commit_op_checkpoint_superblock(self: *Self) void {
+            assert(self.superblock.free_set.opened);
             assert(self.state_machine_opened);
             assert(self.commit_stage == .checkpoint_superblock);
             assert(self.commit_prepare.?.header.op == self.op);
@@ -4739,7 +4772,6 @@ pub fn ReplicaType(
         // (or just be inlined into on_request_sync_checkpoint).
         fn ignore_sync_request_message(self: *const Self, message: *const Message) bool {
             assert(message.header.command == .request_sync_checkpoint or
-                message.header.command == .request_sync_free_set or
                 message.header.command == .request_sync_client_sessions);
 
             const command = @tagName(message.header.command);
@@ -4765,7 +4797,6 @@ pub fn ReplicaType(
             switch (message.header.command) {
                 inline .request_sync_checkpoint,
                 .request_sync_client_sessions,
-                .request_sync_free_set,
                 => |command_comptime| {
                     const message_header = message.header.into_const(command_comptime).?;
 
@@ -4800,7 +4831,6 @@ pub fn ReplicaType(
         // (or just be inlined into on_sync_checkpoint).
         fn ignore_sync_response_message(self: *const Self, message: *const Message) bool {
             assert(message.header.command == .sync_checkpoint or
-                message.header.command == .sync_free_set or
                 message.header.command == .sync_client_sessions);
             assert(message.header.replica < self.replica_count);
 
@@ -4817,7 +4847,7 @@ pub fn ReplicaType(
             const stage: *const SyncStage.RequestingTrailers = &self.syncing.requesting_trailers;
 
             switch (message.header.command) {
-                inline .sync_checkpoint, .sync_client_sessions, .sync_free_set => |command| {
+                inline .sync_checkpoint, .sync_client_sessions => |command| {
                     const message_header = message.header.into_const(command).?;
 
                     if (message_header.checkpoint_op != stage.target.checkpoint_op) {
@@ -6675,7 +6705,6 @@ pub fn ReplicaType(
                     assert(!self.standby());
                 },
                 .request_sync_checkpoint,
-                .request_sync_free_set,
                 .request_sync_client_sessions,
                 => {
                     maybe(self.standby());
@@ -6684,7 +6713,6 @@ pub fn ReplicaType(
                     assert(message.header.replica != replica);
                 },
                 .sync_checkpoint,
-                .sync_free_set,
                 .sync_client_sessions,
                 => {
                     assert(!self.standby());
@@ -7673,6 +7701,7 @@ pub fn ReplicaType(
                 .setup_client_replies,
                 .checkpoint_client_replies,
                 .checkpoint_grid_repair,
+                .checkpoint_free_set,
                 .checkpoint_superblock,
                 => self.sync_dispatch(.canceling_commit),
 
@@ -7820,6 +7849,7 @@ pub fn ReplicaType(
                 .setup_client_replies,
                 .checkpoint_client_replies,
                 .checkpoint_grid_repair,
+                .checkpoint_free_set,
                 .checkpoint_superblock,
                 => {},
             }
@@ -7889,7 +7919,6 @@ pub fn ReplicaType(
                 "manifest_oldest_address={[manifest_oldest_address]} " ++
                 "manifest_newest_checksum={[manifest_newest_checksum]} " ++
                 "manifest_newest_address={[manifest_newest_address]} " ++
-                "free_set_checksum={[free_set_checksum]x:0>32} " ++
                 "client_sessions_checksum={[client_sessions_checksum]x:0>32}", .{
                 .replica = self.replica,
                 .checkpoint_op = stage.target.checkpoint_op,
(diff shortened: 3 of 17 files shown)


src/lsm/schema.zig
```diff
@@ -334,6 +335,65 @@ pub const TableData = struct {
     }
 };
 
+pub const FreeSetNode = struct {
+    pub const Word = u64;
```
matklad, line 345: This seemingly duplicates `FreeSet.Word`, but I think that's false sharing --- there's no _fudnamental_ reason when in-memory and on-disk words must match. We require it in the current implementation, and indeed, in encoded free set, we assert that the two are the same.

src/lsm/schema.zig
```diff
@@ -334,6 +335,65 @@ pub const TableData = struct {
...
+        assert((header.size - @sizeOf(vsr.Header)) % @sizeOf(Word) == 0);
+
+        return header_metadata;
+    }
+
+    pub fn assert_valid_header(free_set_block: BlockPtrConst) void {
+        _ = metadata(free_set_block);
+    }
+
+    pub fn next(free_set_block: BlockPtrConst) ?struct { checksum: u128, address: u64 } {
```
matklad: I think we should add `vsr.BlockReference` as it is used in a couple of different places. But that's for a follow up.

src/vsr/superblock.zig
```diff
@@ -85,29 +78,28 @@ pub const SuperBlockHeader = extern struct {
     /// The checksum of the previous superblock to hash chain across sequence numbers.
     parent: u128,
 
-    /// The checksum over the actual encoded block free set in the superblock trailer.
-    free_set_checksum: u128,
```
matklad, line 88: While this is not strictly necessary, I think I'll add back the checksum covering the entire free set, as it can protect from encoding/decoding bugs. I think I'll leave it to a follow up though.

src/vsr/superblock.zig
```diff
@@ -304,22 +300,30 @@ pub const SuperBlockHeader = extern struct {
...
         manifest_oldest_address: u64,
         manifest_newest_address: u64,
         snapshots_block_address: u64,
+        free_set_head_address: u64,
 
         /// The last operation committed to the state machine. At startup, replay the log hereafter.
         commit_min: u64,
 
+        // Storage size in bytes.
+        storage_size: u64,
```
matklad, line 345: Storage size is now part of checkpoint state, as it is no longer a function of the rest of the superblock, and we want to know storage size even without having access to the grid.

src/vsr/superblock.zig
```diff
@@ -623,12 +630,12 @@ pub fn SuperBlockType(comptime Storage: type) type {
         /// Staging trailers. These are modified between checkpoints, and are persisted on
         /// checkpoint and sync.
         free_set: FreeSet,
+        free_set_encoded: FreeSetEncoded,
         client_sessions: ClientSessions,
 
         /// Updated when:
         /// - the trailer is serialized immediately before checkpoint or sync, and
         /// - used to construct the trailer during state sync.
-        free_set_buffer: []align(constants.sector_size) u8,
```
matklad, line 665: This buffer is effectively moved inside FreeSetEncoded.

src/vsr/superblock.zig
```diff
@@ -1326,9 +1319,7 @@ pub fn SuperBlockType(comptime Storage: type) type {
                         &working.vsr_state.members,
                         working.vsr_state.replica_id,
                     ) != null);
-                } else if (context.caller == .checkpoint) {
-                    superblock.free_set.checkpoint();
-                }
+                } else if (context.caller == .checkpoint) {}
```
matklad: The free set is now delicately check-pointed in the middle of free-set-encoded checkpointing process.

src/vsr/superblock_free_set.zig
```diff
@@ -578,12 +651,15 @@ test "FreeSet checkpoint" {
     const expectEqual = std.testing.expectEqual;
     const blocks_count = FreeSet.shard_bits;
     var set = try FreeSet.init(std.testing.allocator, blocks_count);
+    set.opened = true;
```
matklad: It is tempting to say that the free set is opened after init (as it indeed is in a valid state where no blocks are allocated), but I'd rather not do it -- while it is a valid state, if we actually do have a non-empty free set on disk, _using_ that valid state can lose data!

src/lsm/schema.zig
```diff
@@ -334,6 +335,65 @@ pub const TableData = struct {
...
+        assert(header.snapshot == 0);
+
+        const header_metadata = std.mem.bytesAsValue(Metadata, &header.metadata_bytes);
+        assert(stdx.zeroed(&header_metadata.reserved));
+
+        if (header_metadata.next_free_set_block_address == 0) {
+            assert(header_metadata.next_free_set_block_checksum == 0);
+        }
+
+        assert((header.size - @sizeOf(vsr.Header)) % @sizeOf(Word) == 0);
```
sentientwaffle, line 375:
```suggestion
        assert(header.size > @sizeOf(vsr.Header));
        assert((header.size - @sizeOf(vsr.Header)) % @sizeOf(Word) == 0);
```

src/lsm/schema.zig
```diff
@@ -334,6 +335,65 @@ pub const TableData = struct {
...
+
+        return header_metadata;
+    }
+
+    pub fn assert_valid_header(free_set_block: BlockPtrConst) void {
+        _ = metadata(free_set_block);
+    }
```
sentientwaffle, line 382: I like this pattern!

src/lsm/schema.zig
```diff
@@ -334,6 +335,65 @@ pub const TableData = struct {
     }
 };
 
+pub const FreeSetNode = struct {
+    pub const Word = u64;
+
+    pub const Metadata = extern struct {
+        next_free_set_block_checksum: u128,
+        next_free_set_block_address: u64,
```
sentientwaffle: :thinking: So the "next" free set block will be decoded into bytes at to an offset less than the current block? Maybe worth a comment here, since that might be surprising.

src/vsr/replica.zig
```diff
@@ -63,6 +63,7 @@ const CommitStage = enum {
     checkpoint_state_machine,
     checkpoint_client_replies,
     checkpoint_grid_repair,
+    checkpoint_free_set,
```
sentientwaffle, line 66: Could `checkpoint_free_set` precede `checkpoint_grid_repair`? That way, if a block happens to be repairing during `checkpoint_free_set`, we might avoid a pause for `checkpoint_grid_repair` by repairing concurrent to the freeset checkpoint.

src/vsr/replica.zig
```diff
@@ -658,9 +660,13 @@ pub fn ReplicaType(
...
-            // Asynchronously open the (Forest inside) StateMachine so that we can repair grid
-            // blocks if necessary:
-            self.state_machine.open(state_machine_open_callback);
+            // Asynchronously open the free set and then the (Forest inside) StateMachine so that we
+            // can repair grid blocks if necessary:
+            self.superblock.free_set_encoded.open(
+                &self.grid,
+                self.superblock.working.free_set_reference(),
+                free_set_open_callback,
+            );
```
sentientwaffle, line 669: Somewhat off-topic, but I think we are inconsistent about whether a callback is the first or last parameter. (This is probably my fault :sweat_smile:  -- I reflexively put callbacks last, but I think most uses in the project put it first. e.g. `Grid.write_block()`).

src/vsr/replica.zig
```diff
@@ -8005,7 +8036,11 @@ pub fn ReplicaType(
                 self.transition_to_recovering_head();
             }
 
-            self.state_machine.open(state_machine_open_callback);
+            self.superblock.free_set_encoded.open(
+                &self.grid,
+                self.superblock.working.free_set_reference(),
+                free_set_open_callback,
+            );
```
sentientwaffle, line 8043: On the fence about pulling this into a `fn free_set_open(self: *Self, callback: ..) void {` (so that the `self.superblock.working.free_set_reference()` is not duplicated with the other calls site. That is, we always open the working freeset, never staging). Maybe not worth it since there are only two callers though.

src/vsr/superblock.zig
```diff
@@ -304,22 +300,30 @@ pub const SuperBlockHeader = extern struct {
...
 
         manifest_oldest_address: u64,
         manifest_newest_address: u64,
         snapshots_block_address: u64,
+        free_set_head_address: u64,
 
         /// The last operation committed to the state machine. At startup, replay the log hereafter.
         commit_min: u64,
 
+        // Storage size in bytes.
```
sentientwaffle: Maybe elaborate along the lines of:
- If `storage_size` is less than the data file size, then the grid blocks beyond `storage_size` were used previously, but have since been freed.
- If `storage_size` is more than the data file size, then the data file might have been truncated/corrupted.

src/vsr/superblock.zig
jorangreef: Fantastic idea to verify the roundtrip.

src/vsr/superblock_free_set.zig
jorangreef: 💯 critical.

src/vsr/superblock.zig
```diff
@@ -1202,8 +1191,11 @@ pub fn SuperBlockType(comptime Storage: type) type {
             assert(superblock.staging.cluster == superblock.working.cluster);
             assert(superblock.staging.vsr_state.replica_id == superblock.working.vsr_state.replica_id);
 
-            assert(superblock.staging.storage_size >= data_file_size_min);
-            assert(superblock.staging.storage_size <= superblock.staging.storage_size_max);
+            const storage_size = superblock.staging.vsr_state.checkpoint.storage_size;
+            assert(storage_size >= data_file_size_min);
+            assert(
+                storage_size <= superblock.staging.storage_size_max,
+            );
```
sentientwaffle: One line?

src/vsr/superblock_free_set.zig
```diff
@@ -270,8 +322,6 @@ pub const FreeSet = struct {
         ) orelse return null;
         assert(block >= reservation.block_base);
         assert(block <= reservation.block_base + reservation.block_count);
-        assert(!set.blocks.isSet(block));
-        assert(!set.staging.isSet(block));
```
sentientwaffle: Why remove these?

src/vsr/superblock_free_set.zig
```diff
@@ -307,15 +367,45 @@ pub const FreeSet = struct {
...
+    /// free set, the blocks can be simultaneously released.
+    fn mark_released(set: *FreeSet, addresses: []const u64) void {
+        assert(!set.opened);
+        var address_previous: u64 = 0;
+        for (addresses) |address| {
+            assert(address > 0);
+
+            // Assert that addresses are sorted and unique. Sortedness is not a requirement, but
+            // a consequence of "first free" allocation algorithm.
+            assert(address > address_previous);
```
sentientwaffle, line 394: Clever!

src/vsr/superblock_free_set.zig
```diff
@@ -307,15 +367,45 @@ pub const FreeSet = struct {
...
+            address_previous = address;
+
+            const block = address - 1;
+            assert(!set.blocks.isSet(block));
+            assert(!set.staging.isSet(block));
+
+            const shard = @divFloor(block, shard_bits);
+            if (set.find_free_block_in_shard(shard) == null) set.index.set(shard);
```
sentientwaffle: Could instead set the bit unconditionally, and add a `maybe()` beforehand.

src/vsr/superblock_free_set.zig
```diff
@@ -364,7 +457,8 @@ pub const FreeSet = struct {
 
     /// Decodes the compressed bitset in `source` into `set`.
     /// Panics if the `source` encoding is invalid.
-    pub fn decode(set: *FreeSet, source: []align(@alignOf(usize)) const u8) void {
+    pub fn decode(set: *FreeSet, source: []align(@alignOf(Word)) const u8) void {
+        assert(!set.opened);
         // Verify that this FreeSet is entirely unallocated.
         assert(set.index.count() == 0);
         assert(set.blocks.count() == 0);
```
sentientwaffle, line 467: Should we assert `source.len % @sizeOf(Word) == 0`, or leave it opaque at this layer since EWAH checks it?

src/vsr/superblock_free_set_encoded.zig
```diff
@@ -0,0 +1,431 @@
...
+        // Reference to the grid is late-initialized in the open, because the free set is part of
+        // the superblock, which doesn't have access to grid. It is set to null by reset, to verify
+        // that the free set is not used before it is opened during sync.
+        grid: ?*Grid = null,
+
+        next_tick: Grid.NextTick = undefined,
+        read: Grid.Read = undefined,
+        write: Grid.Write = undefined,
+        // As the free set is expected to fit in one block, it is written sequentialy, one block at
+        // a time. This is the memory used for writting.
```
sentientwaffle:
```suggestion
        // a time. This is the memory used for writing.
```

src/vsr/superblock_free_set_encoded.zig
```diff
@@ -0,0 +1,431 @@
...
+        write_block: Grid.BlockPtr,
+
+        // SoA representation of block references holding the free set itself.
+        //
+        // After the set is read from disk and decoded, these blocks are manually marked as
+        // allocated.
```
sentientwaffle: s/allocated/acquired, to use consistent terminology.

src/vsr/superblock_free_set_encoded.zig
```diff
@@ -0,0 +1,431 @@
...
+        block_addresses: []u64,
+        block_checksums: []u128,
+        block_count: u32 = 0,
+        // The current block that is being read or written It goes from 0 to block_count - 1 during
+        // checkpoint, and from block_count - 1 to zero during open.
```
sentientwaffle:
```suggestion
        // The current block that is being read or written. It counts from 0 to block_count - 1 during
        // checkpoint, and from block_count - 1 to zero during open.
```

src/vsr/superblock_free_set_encoded.zig
```diff
@@ -0,0 +1,431 @@
...
+            open: *const fn (set: *Self) void,
+            checkpoint: *const fn (set: *Self) void,
+        } = .none,
+
+        comptime {
+            assert(FreeSet.Word == schema.FreeSetNode.Word);
+            assert(chunk_size_max % @sizeOf(FreeSet.Word) == 0);
+        }
+
+        pub fn init(allocator: mem.Allocator, block_count_limit: usize) !Self {
```
sentientwaffle: :thinking: Maybe `address_count_limit`? The relationship between `block_count_limit` and `block_count_max` is tricky right now, since the former counts _all_ blocks, but the latter counts only freeset blocks.

Alternatively, we could rename `block_count_max` to `chunk_count_max` (`chunk_addresses`, `chunk_checksums`, &c). (Though then that gets confusing since `chunk_size_max` doesn't count the header...)

src/vsr/superblock_free_set_encoded.zig
```diff
@@ -0,0 +1,431 @@
...
+            };
+        }
+
+        // These data are stored in the superblock header.
+        pub fn checkpoint_reference(set: *const Self) FreeSetReference {
+            assert(set.size == set.size_transferred);
+            assert(set.callback == .none);
+
+            if (set.block_count == 0) {
+                assert(set.size == 0);
```
sentientwaffle, line 174: Can we assert that `set.grid.?.superblock.free_set` has no acquired blocks?

src/vsr/superblock_free_set_encoded.zig
```diff
@@ -0,0 +1,431 @@
...
+            assert(set.callback == .none);
+            defer assert(set.callback == .open);
+
+            assert(set.size == 0);
+            assert(set.size_transferred == 0);
+            assert(set.block_count == 0);
+            assert(set.block_index == 0);
+
+            set.size = reference.size;
+            set.size_transferred = 0;
```
sentientwaffle: `size_transferred` is already 0.

src/vsr/superblock_free_set_encoded.zig
```diff
@@ -0,0 +1,431 @@
...
+            const chunk = Chunk.for_block(.{
+                .block_index = set.block_index,
+                .block_count = set.block_count,
+                .free_set_size = set.size,
+            });
+
+            stdx.copy_disjoint(
+                .exact,
+                u8,
+                set.buffer[chunk.start..chunk.end],
```
sentientwaffle, line 271: Alternatively, 
```suggestion
                set.buffer[chunk.offset..][0..chunk.size],
```
so that there is no need for `Chunk`'s "end - start == size" invariant.

src/vsr/superblock_free_set_encoded.zig
```diff
@@ -0,0 +1,431 @@
...
+            assert(set.block_count == 0);
+            assert(set.size == 0);
+            set.open_done();
+        }
+
+        fn open_read_next(set: *Self, address: u64, checksum: u128) void {
+            assert(set.callback == .open);
+            assert(set.block_index < set.block_count);
+            assert(address != 0);
+            assert((set.size_transferred == 0) == (set.block_index == set.block_count - 1));
```
sentientwaffle: Assert `size > 0`.

src/vsr/superblock_free_set_encoded.zig
```diff
@@ -0,0 +1,431 @@
...
+                address,
+                checksum,
+                .{ .cache_read = true, .cache_write = false },
+            );
+        }
+
+        fn open_read_next_callback(read: *Grid.Read, block: Grid.BlockPtrConst) void {
+            const set = @fieldParentPtr(Self, "read", read);
+            assert(set.callback == .open);
+            assert(set.block_index < set.block_count);
```
sentientwaffle, line 259: Assert `size > 0`.

src/vsr/superblock_free_set_encoded.zig
```diff
@@ -0,0 +1,431 @@
...
+                set.block_count = stdx.div_ceil(set.size, chunk_size_max);
+            }
+
+            assert(set.grid.?.superblock.free_set.count_reservations() == 0);
+            const reservation = set.grid.?.superblock.free_set.reserve(set.block_count).?;
+            for (0..set.block_count) |index| {
+                const address = set.grid.?.superblock.free_set.acquire(reservation).?;
+                set.block_addresses[index] = address;
+                set.block_checksums[index] = undefined;
+            }
```
sentientwaffle: Nit; multi-loop?
```
for (
    set.block_addresses,
    set.block_checksums,
) |*address, *checksum|
```

src/vsr/superblock_free_set_encoded.zig
```diff
@@ -0,0 +1,431 @@
...
+            set.grid.?.superblock.free_set.forfeit(reservation);
```
sentientwaffle: Maybe wrap this in a `{ ... }` block, and use `defer ... forfeit` immediately after the `.reserve()`.

src/vsr/superblock_free_set_encoded.zig
sentientwaffle, line 343: Immediately after the loop, you can assert that `.acquire(reservation)` returns `null` (i.e. the entire reservation was used).

src/lsm/forest_fuzz.zig
```diff
@@ -161,13 +169,37 @@ const Environment = struct {
...
         env.change_state(.forest_init, .forest_open);
         env.forest.open(forest_open_callback);
 
         try env.tick_until_state_change(.forest_open, .fuzzing);
+
+        env.fragmentate_free_set();
+    }
+
+    /// Allocate a sparse subset of grid blocks to make sure that encoded free set needs more than
+    /// block to exercise block linked list logic from FreeSetEncoded.
```
sentientwaffle:
```suggestion
    /// Allocate a sparse subset of grid blocks to make sure that the encoded free set needs more than
    /// one block to exercise the block linked list logic from FreeSetEncoded.
```

src/vsr/superblock_free_set_encoded.zig
```diff
@@ -0,0 +1,431 @@
...
+                const last_block = options.block_index == options.block_count - 1;
+                const chunk_size = if (last_block)
+                    options.free_set_size - (options.block_count - 1) * chunk_size_max
+                else
+                    chunk_size_max;
+
+                const chunk_start = chunk_size_max * options.block_index;
+                const chunk_end = chunk_start + chunk_size;
+                assert(chunk_end <= options.free_set_size);
+                assert(chunk_size % @sizeOf(FreeSet.Word) == 0);
```
sentientwaffle, line 68:
```suggestion
                assert(chunk_size % @sizeOf(FreeSet.Word) == 0);
                assert(chunk_size > 0);
```

src/vsr/superblock_free_set_encoded.zig
```diff
@@ -0,0 +1,431 @@
...
+        fn checkpoint_done(set: *Self) void {
+            assert(set.callback == .checkpoint);
+            defer assert(set.callback == .none);
+
+            if (set.block_count == 0) {
+                assert(set.size == 0);
+                assert(set.block_index == 0);
+            } else {
+                assert(set.block_index == set.block_count - 1);
+            }
```
sentientwaffle: Alternatively, if `checkpoint_write_next_callback` increments `block_index` unconditionally, then we can simply assert `set.block_count == set.block_index`.

src/vsr/superblock_free_set_encoded.zig
```diff
@@ -0,0 +1,431 @@
...
+/// Free set is stored as a linked list of blocks containing EWAH-encoding of a bitset of allocated
+/// blocks. The length of the linked list is proportional to the degree of fragmentation, rather
+/// that to  size of the data file. The common case is a single block.
+///
+/// The blocks holding free set itself are marked as free in the on-disk encoding, because the
+/// number of blocks required to store the compressed bitset becomes known only after encoding.
+/// This might or might not be related to Russel's paradox.
+///
+/// Linked list is a FIFO. While the blocks are written in the direct order, they have to be read in
+/// the reverse order.
```
sentientwaffle:
```suggestion
/// FreeSetEncoded is the persistent component of the free set. It defines the layout of
/// the free set as stored in the grid between checkpoints.
///
/// Free set is stored as a linked list of blocks containing EWAH-encoding of a bitset of allocated
/// blocks. The length of the linked list is proportional to the degree of fragmentation, rather
/// that to the size of the data file. The common case is a single block.
///
/// The blocks holding free set itself are marked as free in the on-disk encoding, because the
/// number of blocks required to store the compressed bitset becomes known only after encoding.
/// This might or might not be related to Russel's paradox.
///
/// Linked list is a FIFO. While the blocks are written in the direct order, they have to be read in
/// the reverse order.
```

> This might or might not be related to Russel's paradox.

:joy:

src/vsr/replica.zig
```diff
@@ -3547,10 +3559,31 @@ pub fn ReplicaType(
...
+            self.superblock.free_set_encoded.checkpoint(commit_op_checkpoint_free_set_callback);
+        }
+
+        fn commit_op_checkpoint_free_set_callback(set: *SuperBlock.FreeSetEncoded) void {
+            const superblock = @fieldParentPtr(SuperBlock, "free_set_encoded", set);
+            const self = @fieldParentPtr(Self, "superblock", superblock);
+            assert(self.commit_stage == .checkpoint_free_set);
+            assert(self.commit_prepare.?.header.op == self.op);
+            assert(self.superblock.free_set.opened);
+            assert(self.superblock.free_set.count_released() ==
```
sentientwaffle, line 3579: In `commit_op_checkpoint_superblock_callback`, assert that `free_set.count_released() == 0`.

src/vsr/superblock_free_set_encoded.zig
```diff
@@ -0,0 +1,431 @@
...
+        pub fn checkpoint(set: *Self, callback: *const fn (set: *Self) void) void {
+            assert(set.callback == .none);
+            defer assert(set.callback == .checkpoint);
+
+            // Checkpoint process is delicate:
+            //   - encode free set,
+            //   - derive the number of blocks required to store the encoding,
+            //   - allocate blocks for the encoding (in the old checkpoint),
+            //   - checkpoint the free set,
+            //   - release the freshly acquired blocks in the new checkpoint.
```
sentientwaffle: To emphasize/check the end state:
```
defer assert(set.grid.?.superblock.free_set.count_released() == set.block_count);
```

Also maybe number the list, since the order is quite important!

src/vsr/superblock.zig
```diff
@@ -885,6 +886,22 @@ pub fn SuperBlockType(comptime Storage: type) type {
                 superblock.staging.vsr_state.checkpoint.commit_min_checksum);
             assert(update.sync_op_min <= update.sync_op_max);
 
+            const free_set_reference = superblock.free_set_encoded.checkpoint_reference();
+
+            var storage_size = data_file_size_min;
+            if (superblock.free_set.highest_address_acquired()) |address| {
+                storage_size += address * constants.block_size;
+                assert(free_set_reference.head_address != 0);
+                assert(!superblock.free_set.is_free(free_set_reference.head_address));
```
sentientwaffle: Why is this non-free? Doesn't `FreeSetEncoded.checkpoint` free these before it is done? (`set.grid.?.superblock.free_set.release(address);`).

Aha, we release it, but it is only staged, and won't actually be freed until the _next_ checkpoint!
Maybe add an `assert(set.count_released() > 0);` (and `== 0` in the other branch) to emphasize this.

src/vsr/superblock_free_set_encoded.zig
matklad: Yeah, I am unsure here. Originally it spelled something equivalent to `[chunk.offset..chunk.size]`, which is buggy, unless there’s only a single block. 

I was quite scared by that bug (as it happened before I figured the fragmentate trick for fuzzing), and tried to design an API which is less prone to indexing errors. The result is the Chunk API, with the idea that we compute “denormalized” form with free fields so that the call-sites could be most natural.

src/vsr/replica.zig
matklad: Heh, yeah I went back and forth here as well. I also have “context first, lambdas last” rule hard-wired in my brain, but it isn’t really applicable to zig, since there are no lambdas…

@jorangreef any tiger style insights about the order of the callback argument?

src/vsr/superblock_free_set_encoded.zig
sentientwaffle: I don't feel strongly about it; `Chunk.for_block()` is a simple enough function that I think it is fine either way.

src/lsm/schema.zig
matklad: The main idea here is calling `assert_valid_header` when we _write_ a data block, not only during reading.

src/vsr/superblock_free_set.zig
matklad: Good catch! Accidentally cut this out when copying over to mark_released.

src/vsr/superblock_free_set.zig
matklad: Oh dear, this is actually a bug! I should _first_ set the bit in the `blocks`, and only afterwards, if that fills up the shard, mark the index.

src/vsr/superblock_free_set.zig
matklad: I am tempted to do a larger refactor in the follow up and just pass in an `[] Word`, in a case of pushing an `if` up :) 

It seems like the original idea was to pass in the slice of bytes, so that the details of encoding are abstracted away, but we are already leaking alingment, and `[]Word` is just going to be shorter.

src/vsr/superblock_free_set.zig
matklad: I originally had the function operating on a single block at a time, but then I realized that I can pass in the whole slice and do deeper assertions, and got distracted writing the "ifs up fors down" post :rofl:

src/vsr/superblock_free_set_encoded.zig
matklad: Ahh, the idea here is that, every time you set `size`, you should also set `size_transfers` to zero (In a more object oriented code-base, both fields would be groupped into an object that tracks transfer progress). So this is sort-of intentional, even given the assert above:

- the asserts asserts the invariant about `(Open Reset)*` trajectory 
- the assignment establishes the invariant that `size_transfered` is reset on updates to size.

The two end up the same by accident. 


But it's totally true that I'll forget about this deep reasoning next week, and would myself wonder why are we duplciating code here, removed!

src/vsr/superblock_free_set_encoded.zig
matklad: Added a comment with the "why"!

src/vsr/replica.zig
jorangreef: I'm so glad you both raise this... because... I also reflexively put callbacks last in my mind. 😁 

"context first, lambdas last" makes sense also because we tend to "order according to logical use" and callbacks are the last thing that happens.

Beyond this, if you have a situation where you also have interim event callbacks that get emitted multiple times (e.g. as part of a streaming API) in addition to "done" callbacks, then putting the callback last follows the progression better.

So... let's do it, and start to move things over as we go.

src/vsr/superblock_free_set_encoded.zig
matklad: Great catch that we have two kinds of blocks here --- blocks storing encoded free set itself and the grid blocks. I renamed the init parameter to `grid_block_count_limit`

src/vsr/superblock_free_set_encoded.zig
matklad: Yeah, that's better, thanks! I remeber _intentionally_ picking the indexed loop here, but I don't remeber why, and today the multi-loop one looks better to me!

src/vsr/superblock_free_set_encoded.zig
matklad: :pinched_fingers:

src/vsr/superblock.zig
matklad: Will check for `superblock.free_set.is_released` directly!

src/vsr/replica.zig
matklad: Ahhh, I tried that, an then got a weird error in vopr, and then realized that this doesn't work, at least without some extra legwork. 

`checkpoint_free_set` is what "buffer-flips" the free set right now. So, if we run grid-repair _after_ checkpointing the free set, it'll see the new checkpoint and won't know which blocks are released. 

What we can do is change the logic of repair checkpoint to _assume_ that it happens after buffer flip and use is_free rather than is_released:

https://github.com/tigerbeetle/tigerbeetle/commit/b06bf4a882e85e3b9042790177f20e985b350f54

In some sense, this even feels clearer, as we are waiting the writes to the free blocks, which is easier to see how it could lead to troubles.

matklad: I think I'll do it, but in a follow up. One thing I don't like is how subtle it is when the data structures "flip" to the next checkpoint, maybe I find some better way to assert that...

src/vsr/replica.zig
matklad: I'd say the free set (encoded) should go from the superblock entirely maybe? And move into grid, so that we have `grid.open` / `grid.checkpoint` 

:thinking: in https://github.com/tigerbeetle/tigerbeetle/pull/1309#discussion_r1402525289, I wrote

> maybe I find some better way to assert that (that the grid checkpoing uses correct state of the free set)...

This might actually be it? If the grid owns the free set, than checkpointing the free set and waiting for repairs is handled by the single `grid.checkpoing` function, and that can _locally_ sequence the two correctly (as opposed to the current situtation, where the replica has to know to checkpoing free set before the grid).

src/vsr/replica.zig
matklad: Hm, I don't get this? In `commit_op_checkpoint_superblock_callback`, the free set had been checkpointed already, so `free_set.count_released()  == free_set_encoded.block_count`?

src/vsr/replica.zig
sentientwaffle: Yeah, that works. At the same time maybe instead of the replica getting the `FreeSetReference` via a method on `FreeSetEncoded` (or the `Grid`), the `FreeSetReference` could be passed into the `checkpoint` callback (which is then passed into `superblock.checkpoint()`).

src/vsr/replica.zig
sentientwaffle: Ah right, nevermind!

src/vsr/superblock_free_set_encoded.zig
matklad: I initially avoided that because then the reverse case, in the open, looks weird, as the index is one greater than the real index. But that's actually an idiomatic form of a downwards counted loop, so switched to that to avoid special cases.

src/vsr/replica.zig
matklad: Yeah, thought the same as well, but didn't do for two reasons:

* to use the same pattern as in manifest log
* to avoid requiring the replica to keep this bit of state around between callbacks: the replica goes through commit dispatch, so it can't immediately just pass the checkpoint along, it would have to store it as a field, which feels awkward. 

Perhaps in the both cases we rather need a state enum?

```
cosnt state = union(enum) {
   closed,
   at_checkpoint: CheckpointReference,
   working,
}
```

then, checkpoint reference would not be just a getter, but would actually change the state from `at_checkpoint` to `working`.

sentientwaffle: Ah yeah.. it kind of seems like `CommitStage` should be a `union(enum)`, and the checkpoint stages would accumulate the references. But maybe that is more verbose than just doing what we do for the manifest log references :shrug:

src/vsr/superblock.zig
```diff
@@ -320,23 +316,38 @@ pub const SuperBlockHeader = extern struct {
...
 
         // TODO Reserve some more extra space before locking in storage layout.
-        reserved: [28]u8 = [_]u8{0} ** 28,
+        reserved: [8]u8 = [_]u8{0} ** 8,
 
         comptime {
-            assert(@sizeOf(CheckpointState) == 224);
-            assert(@sizeOf(CheckpointState) % @sizeOf(u256) == 0);
+            assert(@sizeOf(CheckpointState) == 256);
+            assert(@sizeOf(CheckpointState) % @sizeOf(u128) == 0);
```
sentientwaffle, line 359: :thinking: Stepping back, what is the purpose for this assertion? I changed it without thinking as part of the pseudo-256-bit-checksum change -- i.e., we are pretending that `manifest_oldest_checksum + manifest_oldest_checksum_padding` is a `u256`. (In the future it will be, but for the time being we don't need it and Zig doesn't support it).

src/vsr/superblock.zig
```diff
@@ -224,6 +219,7 @@ pub const SuperBlockHeader = extern struct {
             assert(state.checkpoint.manifest_oldest_checksum_padding == 0);
             assert(state.checkpoint.manifest_newest_checksum_padding == 0);
             assert(state.checkpoint.snapshots_block_checksum_padding == 0);
+            assert(state.checkpoint.free_set_last_block_checksum_padding == 0);
```
sentientwaffle, line 231: Additionally can assert that `free_set_last_block_address == 0` implies `free_set_last_block_checksum == 0`.

Also `storage_size >= data_file_size_min`.

src/vsr/superblock.zig
matklad: Ah, this is was unintentional change, but  I think this is correct: I think we want to align to u128, as many things are u128, and we want to align to 16 bytes, as that might help with SIMD? 

Basically, we want most things to be aligned "nicely", and 16 seems like a reasonable bound for niceness

src/vsr/superblock.zig
```diff
@@ -320,23 +316,38 @@ pub const SuperBlockHeader = extern struct {
         manifest_newest_checksum_padding: u128 = 0,
         snapshots_block_checksum: u128,
         snapshots_block_checksum_padding: u128 = 0,
+        free_set_last_block_checksum: u128,
+        free_set_last_block_checksum_padding: u128 = 0,
 
         manifest_oldest_address: u64,
         manifest_newest_address: u64,
         snapshots_block_address: u64,
+        free_set_last_block_address: u64,
```
sentientwaffle: Very much a nit, but maybe "free set" could precede "manifest" in this order, since it is accessed sooner, and is "lower level" (grid rather than LSM).

src/vsr/superblock_free_set.zig
matklad: Hm, curiously, the following diff

```
diff --git a/src/vsr/superblock_free_set.zig b/src/vsr/superblock_free_set.zig
index 3ef3db6c..b9250b54 100644
--- a/src/vsr/superblock_free_set.zig
+++ b/src/vsr/superblock_free_set.zig
@@ -325,9 +325,9 @@ pub const FreeSet = struct {
         assert(!set.blocks.isSet(block));
         assert(!set.staging.isSet(block));
 
-        set.blocks.set(block);
         // Update the index when every block in the shard is allocated.
+        set.blocks.set(block);
 
         const address = block + 1;
         return address;
@@ -399,9 +399,9 @@ pub const FreeSet = struct {
             assert(!set.staging.isSet(block));
 
             const shard = @divFloor(block, shard_bits);
-            set.blocks.set(block);
             // Update the index when every block in the shard is allocated.
+            set.blocks.set(block);
             set.staging.set(block);
         }
     }
@@ -426,6 +426,11 @@ pub const FreeSet = struct {
         assert(set.reservation_count == 0);
         assert(set.reservation_blocks == 0);
 
+        var shard: usize = 0;
+        while (shard < set.index.bit_length) : (shard += 1) {
+            assert(set.index.isSet(shard) == (set.find_free_block_in_shard(shard) == null));
+        }
+
         var it = set.staging.iterator(.{ .kind = .set });
         while (it.next()) |block| {
             set.staging.unset(block);
```

trips neither vopr nor forest fuzz...

sentientwaffle: :thinking: I would understand the VOPR not hitting it, but the forest fuzzer is surprising. Shards are 4096 blocks.

sentientwaffle: In `FreeSet.acquire`:

```zig
        const shard = find_bit(
            set.index,
            @divFloor(reservation.block_base, shard_bits),
            div_ceil(reservation.block_base + reservation.block_count, shard_bits),
            .unset,
        ) orelse return null;
        assert(!set.index.isSet(shard));

        const reservation_start = @max(
            shard * shard_bits,
            reservation.block_base,
        );
        const reservation_end = reservation.block_base + reservation.block_count;
        const block = find_bit(
            set.blocks,
            reservation_start,
            reservation_end,
            .unset,
        ) orelse return null;
```

I think we aren't maintaining the shard index strictly! i.e. if we picked a `shard` which is (unexpectedly) full, that would silently work, as long as it wasn't the last shard covered by the reservation. Which it usually wouldn't be, since reservations are worst-case and typically much larger than necessary.

matklad: Yeah, it makes sense that just not updating the index precisely won’t blow up immediately (it could still explode when a replica restarts and gets a different, precise index by reading from disk). That’s why in the diff above I added a loop which asserts, at a checkpoint, that the index is consistent with blocks. That one I would expect to get tripped…

will look more into this tomorrow…

src/lsm/schema.zig
matklad: Ah, that's the same trap as with the manifest log! Renamed to `free_set_last_block / previous_block`

src/vsr/superblock.zig
matklad: :+1: that's actually exactly the reason why I put FreeSetNode before ManifestNode in the scheme!

src/vsr/superblock_free_set.zig
matklad: Mystery solved, I think! 

The reason why this isn't tripped by normal fuzzers is that the shard size, 4096, is pretty large, and we need to fill all of shard's bits to hit this. 

It _is_ tiggered by free_set_fuzz though, via https://github.com/tigerbeetle/tigerbeetle/pull/1319

Thanks again for noticing that something's off here dj, this would have been a horrible bug to discover after the fact.

src/vsr/superblock_free_set.zig
matklad: Tried that, and it looks like `[]align(u8) u8` is actually a better API here: while `[]Word` feels simpler, there are just way to many places to mix up the bytes-vs-words indexing, as _some_ amount of conversions has to happen "in the middle" -- we definitely want to use byte-size in CheckpointState.

I've added % asserts to FreeSetEncoded open&checkpoint, that feels like the right place to assert that

src/vsr/superblock_free_set_encoded.zig
```diff
@@ -0,0 +1,449 @@
...
+            stdx.copy_disjoint(
+                .exact,
+                u8,
+                set.buffer[chunk.start..chunk.end],
+                encoded_words,
+            );
+            set.size_transferred += chunk.size;
+
+            if (schema.FreeSetNode.previous(block)) |previous| {
+                assert(set.block_index > 0);
```
sentientwaffle, line 277: Assert that `previous.{address,checksum}` is not equal to the `last_block_address`/`last_block_checksum`.

sentientwaffle (approved):

src/vsr/superblock_free_set_encoded.zig
matklad: We don't have `last_block_address`, but I've added an assert that all read elements are distinct.

matklad: Huh, 

image

```
λ git show origin/matklad/free-set-in-the-grid 
commit 80bcce2fd1a06fddd2008649b1441566f4c133ed (HEAD -> matklad/free-set-in-the-grid, origin/matklad/free-set-in-the-grid)
Author: Alex Kladov <aleksey.kladov@gmail.com>
Date:   Mon Nov 27 19:51:02 2023 +0000

    freeset: assert no duplicate blocks in open
```

did I broke GitHub by racing with the merge queue? If this doesn't get in here, will sneak this into some of the follow ups!

sentientwaffle: I was wondering about that, it seems like it queued despite a new commit after approval.

matklad:
closing in favor of https://github.com/tigerbeetle/tigerbeetle/pull/1328

PR #1310, opened 2023-11-20
VSR: Pad checksum from u128 to u256

sentientwaffle:
- Checksums are still `u128`.
- But in storage (message/block headers, table index blocks, etc), those checksums are padded to `u256`.
- Don't actually use the `u256` type though, since Zig doesn't support `u256` in `extern struct`s (yet).
- Instead of a custom `Checksum` type (e.g. `[32]u8` or `struct { u128, u128 }`), include explicit `checksum_padding: u128` fields for the time being. This ensures that formatting (`log.debug("header={}", .{message.header})`) and equality checks (`header.checksum == other.checksum`) both work as expected. (In particular, changing to a custom `.equal()` function would be inconvenient, especially with the intent to switch back).

In the future, the whole `u256` can be used for AEAD tags.

### DIFF
--- a/src/lsm/table_data_iterator.zig
+++ b/src/lsm/table_data_iterator.zig
@@ -25,7 +25,7 @@ pub fn TableDataIteratorType(comptime Storage: type) type {
             /// Table data block addresses.
             addresses: []const u64,
             /// Table data block checksums.
-            checksums: []const u128,
+            checksums: []const schema.Checksum,
             direction: Direction,
         };
 
@@ -99,12 +99,14 @@ pub fn TableDataIteratorType(comptime Storage: type) type {
                     .descending => it.context.addresses.len - 1,
                 };
 
+                assert(it.context.checksums[index].padding == 0);
+
                 it.callback = .{ .read = callback };
                 it.context.grid.read_block(
                     .{ .from_local_or_global_storage = on_read },
                     &it.read,
                     it.context.addresses[index],
-                    it.context.checksums[index],
+                    it.context.checksums[index].value,
                     .{ .cache_read = true, .cache_write = true },
                 );
             } else {
@@ -126,7 +128,7 @@ pub fn TableDataIteratorType(comptime Storage: type) type {
                     if (constants.verify) {
                         const header = schema.header_from_block(block);
                         assert(header.address == it.context.addresses[0]);
-                        assert(header.checksum == it.context.checksums[0]);
+                        assert(header.checksum == it.context.checksums[0].value);
                     }
 
                     it.context.addresses = it.context.addresses[1..];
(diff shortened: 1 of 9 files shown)


src/vsr/message_header.zig
```diff
@@ -635,26 +667,33 @@ pub const Header = extern struct {
...
 
         checkpoint_id: u128,
+        checkpoint_id_reserved: u128 = 0,
         /// The corresponding request message's checksum.
-        parent: u128,
-        client: u128,
+        /// TODO: Add this back and remove checkpoint_id instead, once checkpoint_id is added to
+        /// the Prepare header.
+        // parent: u128,
+        // parent_reserved: u128 = 0,
```
sentientwaffle: I'm just commenting these out since I expect to add them all back in my next PR.

(Swapping `checkpoint_id` over to `Prepare` is a more extensive change than simply temporarily disabling `parent` verification.)

jorangreef:
Good call on the `checksum_padding` decision. 👍

sentientwaffle:
I missed a few checksums in the superblock's `CheckpointState`; will update shortly.

src/vsr/replica.zig
```diff
@@ -6216,7 +6216,7 @@ pub fn ReplicaType(
                     @bitCast(Header.PrepareOk{
                         .command = .prepare_ok,
                         .checkpoint_id = checkpoint_id,
-                        .parent = header.parent,
+                        // .parent = header.parent,
```
matklad, line 6219: Did you intend to leave this commented out?

sentientwaffle: Yes; see https://github.com/tigerbeetle/tigerbeetle/pull/1310#discussion_r1399787823

src/vsr/superblock.zig
```diff
@@ -79,17 +80,20 @@ pub const SuperBlockHeader = extern struct {
     /// A monotonically increasing counter to locate the latest superblock at startup.
     sequence: u64,
 
-    /// Protects against writing to or reading from the wrong data file.
-    cluster: u128,
-
     /// The checksum of the previous superblock to hash chain across sequence numbers.
     parent: u128,
+    parent_padding: u128 = 0,
```
matklad, line 85: Is it intentional that we sometimes use `checksum_padding` and sometimes `checksum_reserved`?

src/lsm/table_data_iterator.zig
```diff
@@ -104,7 +104,7 @@ pub fn TableDataIteratorType(comptime Storage: type) type {
                     .{ .from_local_or_global_storage = on_read },
                     &it.read,
                     it.context.addresses[index],
-                    it.context.checksums[index],
```
matklad, line 107: It seems like this could be the right place to assert that the padding is zero? I don't think we assert this anywhere else?

src/vsr/superblock.zig
sentientwaffle: No, not intentional. I'll change them all to "padding".

matklad (approved):

src/vsr/superblock.zig
jorangreef: Related, we could do `nonce_reserved` instead of `reserved_nonce`.

PR #1311, opened 2023-11-21
Docs: Add 3️⃣-node single binary quickstart

cb22:
I feel we should move the Docker sections out of the quickstart entirely, but I'm not sure where a good place to put them would be.

They're not really for production deployment, so `Deploy` doesn't fit. `Recipes` is more client level stuff. We could add something, but not sure if that's great either. Open to suggestions :smile:.

### DIFF
--- a/docs/quick-start/single-binary-three.md
+++ b/docs/quick-start/single-binary-three.md
@@ -1,0 +1,54 @@
+---
+sidebar_position: 2
+---
+
+# Three-node cluster with a single binary
+
+First, download a prebuilt copy of TigerBeetle.
+
+On macOS/Linux:
+
+```console
+git clone https://github.com/tigerbeetle/tigerbeetle; ./tigerbeetle/bootstrap.sh
+```
+
+On Windows:
+
+```console
+git clone https://github.com/tigerbeetle/tigerbeetle; .\tigerbeetle\bootstrap.ps1
+```
+
+Want to build from source locally? Add `-build` as an argument to the bootstrap script.
+
+## Running TigerBeetle
+
+Now create the TigerBeetle data file for each replica
+
+```console
+./tigerbeetle format --cluster=0 --replica=0 --replica-count=3 0_0.tigerbeetle
+./tigerbeetle format --cluster=0 --replica=1 --replica-count=3 0_1.tigerbeetle
+./tigerbeetle format --cluster=0 --replica=2 --replica-count=3 0_2.tigerbeetle
+```
+
+And start each server in a new terminal window:
+
+```console
+./tigerbeetle start --addresses=3000,3001,3002 0_0.tigerbeetle
+```
+
+```console
+./tigerbeetle start --addresses=3000,3001,3002 0_1.tigerbeetle
+```
+
+```console
+./tigerbeetle start --addresses=3000,3001,3002 0_2.tigerbeetle
+```
+
+TigerBeetle uses the `--replica` that's stored in the data file as an index into the `--addresses`
+provided.
+
+### Connect with the CLI
+
+Now you can connect to the running server with any client. For a quick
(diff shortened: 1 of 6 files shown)


jorangreef:
Great to de-Docker QuickStart and put the spotlight on a single binary experience.

I think if people are looking for Docker steps, they would expect it under Deploy. We can put it there and improve it while we are pre-production?

docs/quick-start/single-binary-three.md
```diff
@@ -0,0 +1,54 @@
...
+```
+
+On Windows:
+
+```console
+git clone https://github.com/tigerbeetle/tigerbeetle; .\tigerbeetle\bootstrap.ps1
+```
+
+Want to build from source locally? Add `-build` as an argument to the bootstrap script.
+
```
matklad, line 22: This feels a bit odd. Three-node-cluster feels like a "real" thing, not just a quick-start, so it probably should recomend "production" way of getting tigerbeetle, and not our bootstrap script (the assumption here is that `git clone https://github.com/tigerbeetle/tigerbeetle; ./tigerbeetle/bootstrap.sh` is _not_ how we recommend folks to get TB in prod). 

So I expect to see here something like "Download tigerbeetle binary", where Download is a link to the docs which explains all the ways to get tigerbetele (with the primary one being a github release).

docs/quick-start/single-binary-three.md
```diff
@@ -0,0 +1,54 @@
...
+```console
+./tigerbeetle start --addresses=3000,3001,3002 0_2.tigerbeetle
+```
+
+TigerBeetle uses the `--replica` that's stored in the data file as an index into the `--addresses`
+provided.
+
+### Connect with the CLI
+
+Now you can connect to the running server with any client. For a quick
```
matklad, line 52: Connect how? I think it's crucial to show the cluster address string here. Could be add something like "for example, using a repl: "?

cb22:
> I think if people are looking for Docker steps, they would expect it under Deploy. We can put it there and improve it while we are pre-production?

Another one of my concerns was only having a "Docker" section under deploy, might give the impression that Docker's the only way to deploy. Maybe a disclaimer there would be good enough, or could do a pretty quick systemd example file?

matklad:
I think both "here's the docker image" and "we have github releases" should go under "Instllation" section maybe?

docs/quick-start/single-binary-three.md
cb22: > This feels a bit odd. Three-node-cluster feels like a "real" thing, not just a quick-start, so it probably should recomend "production" way of getting tigerbeetle

Hmm perhaps there's some expectation mismatch here. My thinking was that this isn't how you would run TigerBeetle in production - not yet - but if you want to experiment with running a full cluster on a single machine. Maybe that's not a valid use case to be highlighting in a quickstart at all?

matklad: Yeah, that's also true. I guess what makes me uneasy is that:

* "installation" feels off-topic for "topical guides". My expectation here is, when the sure is looking at three-node tutorial, they've already run the basic tutorial and have tigerbeetle
* at the same time, for people who do want to get the answer to the question of "how do I get a TigerBeetle", the bootstrap instructions feel not very useful. The bit about "want to build from source" is not something I am immediately intersted. At the same time, it's clear that `git | .sh` is _not_ how I should be installing tigerbeetle in prod, but there's no link to the explanation of _how_ to download the binary properly.

matklad: Added install page in https://github.com/tigerbeetle/tigerbeetle/pull/1321, hat tip to jorangreef for suggesting that we just add a markdown table with releases to our docs!

matklad (approved):
To not let the perfect be the enemy of the good, let's maybe merge as is, and iterate on bootstrap and what not separately?

PR #1313, opened 2023-11-21
don't mix tabs and spaces in bootstrap.sh

matklad:
bootstarp.sh is probably something people would look at to learn how to download tigerbeetle, it should be readable.

### DIFF
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -7,7 +7,7 @@
 {
     from_source=""
     if [ "$1" = "-build" ]; then
-	from_source="t"
+        from_source="t"
     fi
 
     set -e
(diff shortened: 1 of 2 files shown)


cb22:
While you're here, can you add it to the `lint_shellcheck` too!

matklad:
Good call, though it seems that shellcheck is fine about mixing tabs&spaces here?

sentientwaffle:
As an aside, ignoring style preferences, there is one place where "spaces vs tabs" actually matters in shell scripts -- heredocs work better with tabs. (e.g. https://tldp.org/LDP/abs/html/here-docs.html example 19-4).

cb22:
> though it seems that shellcheck is fine about mixing tabs&spaces here?

I was quite surprised by this too, given how opinionated they are on other things!

cb22 (approved):

PR #1314, opened 2023-11-21
Fix storage fault probabilities and add paper reference

jorangreef:
Before, in the DESIGN doc, we indicated the probabilities as "per disk" instead of as a "percentage of disks" experiencing a storage fault.

The referenced paper was also to Bairavasundaram's LSE analysis, instead of their corruption analysis, so use the direct link (and add the paper also to our references).

### DIFF
--- a/docs/DESIGN.md
+++ b/docs/DESIGN.md
@@ -14,15 +14,15 @@ TigerBeetle is designed to a higher safety standard than a general-purpose relat
 
 * Strict consistency, CRCs and crash safety are not enough.
 
-* TigerBeetle **recovers correctly from Latent Sector Errors** (e.g. [1.4% of Enterprise HDD disks per year on average](https://www.usenix.org/legacy/events/fast08/tech/full_papers/bairavasundaram/bairavasundaram.pdf)) **detects and repairs disk corruption or misdirected I/O where firmware reads/writes the wrong sector** (e.g. [0.466% of Nearline HDD disks per year on average](https://www.usenix.org/legacy/events/fast08/tech/full_papers/bairavasundaram/bairavasundaram.pdf)), and **detects data tampering** with hash-chained cryptographic checksums.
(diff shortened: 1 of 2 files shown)


docs/DESIGN.md
```diff
@@ -14,7 +14,7 @@ TigerBeetle is designed to a higher safety standard than a general-purpose relat
 
 * Strict consistency, CRCs and crash safety are not enough.
 
-* TigerBeetle **detects and repairs disk corruption** ([3.45% per 32 months, per disk](https://research.cs.wisc.edu/wind/Publications/latent-sigmetrics07.pdf)), **detects and repairs misdirected writes** where the disk firmware writes to the wrong sector ([0.042% per 17 months, per disk](https://research.cs.wisc.edu/wind/Publications/latent-sigmetrics07.pdf)), and **prevents data tampering** with hash-chained cryptographic checksums.
+* TigerBeetle **recovers correctly from Latent Sector Errors** (e.g. [1.4% of Enterprise HDD disks per year on average](https://www.usenix.org/legacy/events/fast08/tech/full_papers/bairavasundaram/bairavasundaram.pdf)) **detects and repairs disk corruption or misdirected I/O where firmware reads/writes the wrong sector** (e.g. [0.466% of Nearline HDD disks per year on average](https://www.usenix.org/legacy/events/fast08/tech/full_papers/bairavasundaram/bairavasundaram.pdf)), and **detects data tampering** with hash-chained cryptographic checksums.
```
sentientwaffle: There's also [A Study of SSD Reliability in Large Scale Enterprise Storage Deployments](https://www.usenix.org/system/files/fast20-maneas.pdf) which is a more recent study that covers SSDs instead. See page 5 for an excellent summary of the error types + frequencies.

matklad: >detects

I think we actually do prevent it, in a sense that we proactively check hash chains and don't accept data which doesn't chain? 

I'd even say that we don't detect tampering per se, as tampering looks indistinguishable from latent sector errors for us?

jorangreef: Good points. We had `prevents` and I changed it, because we don't *prevent* the tampering per se (someone can do it while the TB process is off) but rather *recover* from it.

It's indistinguishable from corruption, as corruption is indistinguishable from tampering, except that if the tamperer has access to more than 1 machine they can break things (if they really know the format and can time everything). i.e. We're not Byzantine-fault tolerant. But it would be hard to tamper with a TB data file (and across machines).

sentientwaffle (approved):

PR #1315, opened 2023-11-21
Fix the broken golang test

batiati:
Current issue running Go tests:

```
$ go test
# github.com/tigerbeetle/tigerbeetle-go [github.com/tigerbeetle/tigerbeetle-go.test]
./tb_client_test.go:74:27: undefined: ToUint128
FAIL    github.com/tigerbeetle/tigerbeetle-go [build failed]
```

After fixing the reference `type.ToUint128`:

```
$ go test
# github.com/tigerbeetle/tigerbeetle-go [github.com/tigerbeetle/tigerbeetle-go.test]
./tb_client_test.go:74:43: cannot use TIGERBEETLE_CLUSTER_ID (constant 0 of type int) as type uint in argument to types.ToUint128
FAIL    github.com/tigerbeetle/tigerbeetle-go [build failed]
```

This PR also adjusts the parameter of `ToUint128` from `int` to `uint64`.


src/clients/go/pkg/types/main.go
```diff
@@ -78,7 +78,7 @@ func BigIntToUint128(value big.Int) Uint128 {
 }
 
 // ToUint128 converts a integer to a Uint128.
-func ToUint128(value int) Uint128 {
-	values := [2]uint64{uint64(value), 0}
+func ToUint128(value uint64) Uint128 {
```
kprotty, line 81: Would it make sense to have `(low, high uint64)`?

matklad:
Ah, so we do have tests for go!

Let's maybe add them to our CI script over here

https://github.com/tigerbeetle/tigerbeetle/blob/d1284b0798e0d57225017efe7af47c492afc2e71/src/clients/go/ci.zig#L14

?

src/clients/go/pkg/types/main.go
batiati: Honestly, I don't know 🤔
This helper is intended to be an ergonomic way to convert single integers and literals to U128, like `ToUint128(42)`.

I'm not sure if `ToUint128(42, 0)` is a good name, maybe we can call it `U64ToU128`.
Anyway, there's no helper function to convert a U128 back to a pair of U64, so we should add one as well if we decide to accept two parameters, otherwise, it would be inconsistent.

kprotty (approved):

PR #1296, opened 2023-11-13
release: 2023-11-13

sentientwaffle:

PR #1316, opened 2023-11-21
tidy: validate changelog

matklad:
Check that lines are not too long and don't end in whitespace using the existing tidy script. One alternative would be to use a general markdown linter, I don't know a good one off the top of my head.

As [promised](https://github.com/tigerbeetle/tigerbeetle/pull/1296#discussion_r1391170235) a couple of weeks ago!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@
   uninitialized bytes can be stored or transmitted over the network. Previously, we checked only
   if the struct had no padding, but not its fields.
 
-### Internals  
+### Internals
 
 - [#1299](https://github.com/tigerbeetle/tigerbeetle/pull/1299)
 
@@ -42,7 +42,7 @@
   repurpose and reinterpret fields so that the same header could hold different data depending on
   the context. Now, commands have their own specialized data type containing the fields that are
   only pertinent to the context, making the API much safer and intent-clear.
-  
+
 - [#1304](https://github.com/tigerbeetle/tigerbeetle/pull/1304)
 
   With larger headers (see #1295) we have enough room to make the cluster ID a 128-bit integer,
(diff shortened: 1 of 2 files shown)


src/tidy.zig
```diff
@@ -89,8 +110,8 @@ fn is_naughty(path: []const u8) bool {
 
 fn find_long_line(file_text: []const u8) !?usize {
     var line_iterator = mem.split(u8, file_text, "\n");
-    var line_number: usize = 0;
-    while (line_iterator.next()) |line| : (line_number += 1) {
+    var line_index: usize = 0;
```
matklad, line 114: Finally figured out a resonable naming scheme for user-visible line numbers: `line_index` is zero-based, `linu_number` is one-based.

src/tidy.zig
```diff
@@ -59,6 +59,27 @@ test "tidy" {
     }
 }
 
+test "tidy changelog" {
+    const allocator = std.testing.allocator;
+
+    const max_bytes = 1024 * 1024;
+    const changelog = try fs.cwd().readFileAlloc(allocator, "CHANGELOG.md", max_bytes);
```
sentientwaffle: Maybe `changelog_size_max`?

matklad: Good call! I was mechanically applying "named parameter pattern", but of course we have more context here for the name than at the `readFileAlloc`'s definition site!

CHANGELOG.md
```diff
@@ -221,4 +222,5 @@ history, but any notable change from this point on shall be captured by this doc
 For archeological inquiries, check out the state of the repository at the time of the first
 changelog:
 
-[https://github.com/tigerbeetle/tigerbeetle/](https://github.com/tigerbeetle/tigerbeetle/tree/d2d6484188ecc57680e8bde446b5d09b6f2d83ca)
+[https://github.com/tigerbeetle/tigerbeetle/](
+https://github.com/tigerbeetle/tigerbeetle/tree/d2d6484188ecc57680e8bde446b5d09b6f2d83ca)
```
sentientwaffle, line 226: Hard-wrapping links is annoying, but keeping the lint simple is probably more useful for the time being.
Maybe reconsider adding some exceptions if we ever need to include a URL that is >100 chars.

sentientwaffle (dismissed):

src/tidy.zig
```diff
@@ -59,6 +59,27 @@ test "tidy" {
     }
 }
 
+test "tidy changelog" {
+    const allocator = std.testing.allocator;
+
+    const changelog_size_max = 1024 * 1024;
+    const changelog = try fs.cwd().readFileAlloc(allocator, "CHANGELOG.md", changelog_size_max);
```
sentientwaffle, line 66: Ah, CI is complaining that this needs to free the memory.

sentientwaffle (approved):

PR #1320, opened 2023-11-23
vsr: lift BlockReference up

matklad:
We use an isomorphic struct in a couple of places, so makes sense to
lift it up to the VSR layer, rather then redeclare everywhere.

We could also consider changing things like ManifestLogReferences from

```
  pub const ManifestReferences = struct {
      oldest_checksum: u128,
      oldest_address: u64,
      newest_checksum: u128,
      newest_address: u64,
      block_count: u32,
  }
```

to something akin to

```
  pub const ManifestReferences = struct {
      oldest: BlockReference,
      newest: BlockReference,
      block_count: u32,
  }
```

but I think it won't be a big win, flat is simpler. Similarly, I think
it's better if `grid.read_block` and such accept two flat parameters, as
that simpler, and also avoids the need to re-pack SoAs which we use a
lot.

### DIFF
--- a/src/lsm/manifest_log.zig
+++ b/src/lsm/manifest_log.zig
@@ -38,6 +38,7 @@ const tree = @import("tree.zig");
 const RingBuffer = @import("../ring_buffer.zig").RingBuffer;
 const schema = @import("schema.zig");
 const TableInfo = schema.ManifestNode.TableInfo;
+const BlockReference = vsr.BlockReference;
 
 const block_builder_schema = schema.ManifestNode{
     .entry_count = schema.ManifestNode.entry_count_max,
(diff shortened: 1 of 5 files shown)


matklad:
(I originally expected this to be bigger, as I wanted to use BlockReference everywhere I could, but than I asked myself whether I should, and reasoned that I needn't)

sentientwaffle (dismissed):
LGTM. One other place where `BlockReference` might help, is to turn the `StorageChecker.checksum_grid` `while (true)` loop into a `while (free_set_block) |block_reference|` loop.

matklad:
Ah, right. I don't think getting rid of while per se is worth...while, but it does simplify the code overall!

sentientwaffle (approved):

PR #1322, opened 2023-11-24
freeset: defense in depth, include an extra checksum

matklad:


src/vsr/superblock.zig
```diff
@@ -318,6 +321,11 @@ pub const SuperBlockHeader = extern struct {
         commit_min_checksum: u128,
         commit_min_checksum_padding: u128 = 0,
 
+        /// Checksum covering the entire encoded free set. Strictly speaking it is redundant:
+        /// free_set_last_block_checksum indirectly covers the same data. It is still useful
+        /// to protect from encoding-decoding bugs as a defense in depth.
+        free_set_checksum: u128,
+        free_set_checksum_padding: u128 = 0,
```
sentientwaffle: I think you can omit the padding -- I haven't pushed it yet, I'm going to remove the padding from other non-"tag" checksums (like checkpoint ids). That is, the only padded checksums are: message `checksum`, message `checksum_body`, and any reference to either of those (e.g. `commit_min_checksum` or `parent`).

(We needed to free up some space in Prepare messages.)

matklad: Ah, that makes more sense to me!

So we commit to `u128` checksums. However, for messages traveling over the network we also leave space so that replicas can tag the message as "yes, it really is me who is sending the message". 

Though I guess I am still a bit confused about our AEAD plans, guess that might be a nice topic for the next W&T?

sentientwaffle: I think that once we switch to AEAD, `header.checksum` and `header.checksum_body` will probably be renamed to better reflect that they aren't simple checksums. Whereas checksums that really are just a hash will stay a `u128`.
Yeah, this would be a good W&T discussion.

sentientwaffle (approved):

PR #1323, opened 2023-11-27
LSM: Fix tree fuzz using outdated pointers

cb22:
Our tree fuzzer had a `lookup_value: ?*const Value` which stored the result from the `lookup_from_levels_storage` callback. However, this result is returned after a `tick_until_state_change()` which can result in the value changing between the callback and it being read.

This `*const Value` ultimately points into a data block, so it's subject to the same restrictions. The other call site of `lookup_from_levels_storage` is in the Groove, which does correctly take a copy for the object cache.

This started manifesting from 32f438a35018f456de93a77c56562382deeea9b5 - likely because the increased checksum size caused things to spill over.

Failing fuzz:
```bash
zig build -Doptimize=ReleaseSafe fuzz_lsm_tree -- --seed 8745333161403778570
```

### DIFF
--- a/src/lsm/tree.zig
+++ b/src/lsm/tree.zig
@@ -296,6 +296,7 @@ pub fn TreeType(comptime TreeTable: type, comptime Storage: type) type {
         /// This function is intended to never be called by regular code. It only
         /// exists for fuzzing, due to the performance overhead it carries. Real
         /// code must rely on the Groove cache for lookups.
+        /// The returned Value pointer is only valid synchronously.
         pub fn lookup_from_memory(tree: *Tree, key: Key) ?*const Value {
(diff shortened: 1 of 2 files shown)


src/lsm/tree.zig
```diff
@@ -369,6 +371,7 @@ pub fn TreeType(comptime TreeTable: type, comptime Storage: type) type {
         }
 
         /// Call this function only after checking `lookup_from_memory()`.
+        /// The returned Value pointer is only valid synchronously.
```
sentientwaffle: Maybe instead:

```suggestion
        /// The callback's Value pointer is only valid synchronously within the callback.
```

sentientwaffle (approved):

PR #1324, opened 2023-11-27
LSM: Fix segmented array fuzz not having any `insert()`s

cb22:
It's possible (but unlikely) for the fuzz code above `context.remove_all` to not generate any insert ops, which will cause `remove_all` to fail.

Ensure there's always at least one insert before calling `remove_all`.

Failing fuzz:
```bash
zig build -Doptimize=Debug fuzz_lsm_segmented_array -- --seed 11767020692845232958
```

### DIFF
--- a/src/lsm/segmented_array.zig
+++ b/src/lsm/segmented_array.zig
@@ -1027,7 +1027,10 @@ fn TestContext(
                 }
             }
 
-            try context.remove_all();
+            // Rarely, the code above won't generate an insert at all.
+            if (context.inserts > 0) {
+                try context.remove_all();
+            }
 
             if (element_order == .unsorted) {
                 // Insert at the beginning of the array until the array is full.


src/lsm/segmented_array.zig
```diff
@@ -1027,6 +1027,11 @@ fn TestContext(
                 }
             }
 
+            // Ensure we have at least one insert to try remove_all() against.
+            if (context.inserts == 0) {
+                try context.insert();
+            }
+
             try context.remove_all();
```
matklad: Maybe 

```zig
            if (context.inserts > 0) {
                try context.remove_all();
            }
```

?

To make sure we still cover no-inserts case. 

alternatively, maybe replace `try testing.expect(context.inserts > 0);` with a `maybe`?

Why are we asserting that at all? To make sure we don't do no-op fuzzing by accident?

cb22: > Why are we asserting that at all? To make sure we don't do no-op fuzzing by accident?

I'm not sure - I reasoned that since we always tried to call `context.remove_all()` it was intended that we are expecting at least one call to `.insert()` and it was an oversight that there isn't always.

But I'm not familiar with the segmented array code at all :smile:

matklad: `remove_all` is not the code from `SegmentedArray`, it's the code from the fuzzer, which just calls remove in a loop. So I don't think there are any internal invariants for the data structure that it must be non-empty. And testing the empty case seems useful, it is an edge case after all. 

:thinking: the assert is still useful to make sure this is not a no-op _by accident_. Eg, if you call `remove_all` on a code path where the array is _always_ empty, that _should_ be an error. I guess I convinced myself that adding an  `if (context.inserts > 0) {` guard is the right solution here, as it signals, at the call-site, whether we expect the empty case to be possible.

matklad:
Status: waiting on `if (context.inserts > 0) {` as I think that's meaningfully better (we still test an important "only removals" case).

matklad (approved):

PR #1327, opened 2023-11-27
vsr: panic if crypto.random.int returns zero

matklad:
While getting zero here by accident is possible, it is by far likelier that the CSPRNG is broken and panicking would be prudent.

Context: https://github.com/tigerbeetle/tigerbeetle/pull/1295#discussion_r1391495628

There's a kernel of truth to every joke!

### DIFF
--- a/src/clients/c/tb_client/context.zig
+++ b/src/clients/c/tb_client/context.zig
@@ -115,8 +115,8 @@ pub fn ContextType(
             errdefer allocator.destroy(context);
 
             context.allocator = allocator;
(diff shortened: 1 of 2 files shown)


sentientwaffle (approved):
LGTM.

Mixed feelings here though:

If the CPRNG is truly random, then `0` is indeed a valid result.
And I think we _need_ to trust that the CPRNG is truly random -- otherwise, we have to worry about client id collisions, which could cause all kinds of confusing problems.

But as you said, `P(id=0 | CPRNG is buggy) > P(id=0 | CPRNG hit zero by chance)` so we should crash. :+1:

PR #1330, opened 2023-11-28
VSR/FreeSet: Fix shard index update

sentientwaffle:
In `FreeSet.acquire()`:

```zig
    const shard = find_bit(
        set.index,
        @divFloor(reservation.block_base, shard_bits),
        div_ceil(reservation.block_base + reservation.block_count, shard_bits),
        .unset,
    ) orelse return null;

    // ...

    // Update the index when every block in the shard is allocated.
```

The bug here is: The block that we acquire is not necessarily from `shard`. If our reservation intersects more than one shard, then the first shard might have free bits, but _not within our reservation_. If we acquire a bit from the reservation's _second_ shard, then we may fail to mark the shard as exhausted!

The added test fails prior to this bugfix.

Also: This bug was found by reading the code, not any fuzzer failures. Because shards are so large (4096 bits) relative to the number of blocks acquired per reservation, it is hard to trigger accidentally. Perhaps we should make the shard size a cluster-configurable constant, and set a much lower value during fuzzing.

### DIFF
--- a/src/vsr/superblock_free_set.zig
+++ b/src/vsr/superblock_free_set.zig
@@ -307,16 +307,16 @@ pub const FreeSet = struct {
         assert(reservation.block_base + reservation.block_count <= set.reservation_blocks);
         assert(reservation.session == set.reservation_session);
 
-        const shard = find_bit(
+        const shard_start = find_bit(
             set.index,
             @divFloor(reservation.block_base, shard_bits),
             div_ceil(reservation.block_base + reservation.block_count, shard_bits),
             .unset,
         ) orelse return null;
-        assert(!set.index.isSet(shard));
+        assert(!set.index.isSet(shard_start));
 
         const reservation_start = @max(
-            shard * shard_bits,
+            shard_start * shard_bits,
             reservation.block_base,
         );
         const reservation_end = reservation.block_base + reservation.block_count;
@@ -330,6 +330,12 @@ pub const FreeSet = struct {
         assert(block <= reservation.block_base + reservation.block_count);
         assert(!set.blocks.isSet(block));
         assert(!set.staging.isSet(block));
+
+        // Even if "shard_start" has free blocks, we might acquire our block from a later shard.
+        // (This is possible because our reservation begins part-way through the shard.)
+        const shard = @divFloor(block, shard_bits);
+        maybe(shard == shard_start);
+        assert(shard >= shard_start);
 
         set.blocks.set(block);
         // Update the index when every block in the shard is allocated.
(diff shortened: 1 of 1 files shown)


matklad:
:volcano: :exploding_head: :volcano: 

Most excellent catch @sentientwaffle !

>not any fuzzer failures. 

Another reason here would be https://github.com/tigerbeetle/tigerbeetle/pull/1319 --- I think we also didn't have any asserts to catch this at all, short of full replica  restart (where in-memory and reconstructed-from-disk indexes could be different). I think with that assert in this should be discover able at least in theory by the free set fuzzer. But it also is extremely slow, and we are not running it on CI yet :shrug: 

_someone_ should invent a not-rocket-science analogue for fuzzing --- fuzzers are _so_ annoying to keep fresh :weary:

matklad:
Ah, but look at this (at ad627ebf11d2d471feae26c49f67dfc35c980575)!

```
Run 10
info(fuzz): Fuzz seed = 17693596620990535439
info(fuzz_vsr_superblock_free_set): event_distribution = enums.EnumFieldStruct(@typeInfo(vsr.superblock_free_set_fuzz.FreeSetEvent).Union.tag_type.?,f64,null){ .reserve = 1.90e+01, .forfeit = 1.00e+00, .acquire = 2.08e+02, .release = 5.63e+01, .checkpoint = 2.13e+01 }
info(fuzz_vsr_superblock_free_set): event_count = 2000000
thread 211992 panic: reached unreachable code
/home/matklad/p/tb/sim/zig/lib/std/debug.zig:343:14: 0x23f982 in assert (fuzz_vsr_superblock_free_set)
             ^
/home/matklad/p/tb/sim/src/vsr/superblock_free_set.zig:180:19: 0x251a45 in verify_index (fuzz_vsr_superblock_free_set)
            assert((set.find_free_block_in_shard(shard) == null) == set.index.isSet(shard));
                  ^
/home/matklad/p/tb/sim/src/vsr/superblock_free_set.zig:444:25: 0x242d4f in checkpoint (fuzz_vsr_superblock_free_set)
        set.verify_index();
                        ^
/home/matklad/p/tb/sim/src/vsr/superblock_free_set_fuzz.zig:97:36: 0x241075 in run_fuzz (fuzz_vsr_superblock_free_set)
                free_set.checkpoint();
                                   ^
/home/matklad/p/tb/sim/src/vsr/superblock_free_set_fuzz.zig:28:17: 0x243546 in main (fuzz_vsr_superblock_free_set)
    try run_fuzz(allocator, prng.random(), blocks_count, events);
                ^
/home/matklad/p/tb/sim/zig/lib/std/start.zig:574:37: 0x23ca8e in posixCallMainAndExit (fuzz_vsr_superblock_free_set)
                                    ^
/home/matklad/p/tb/sim/zig/lib/std/start.zig:243:5: 0x23c571 in _start (fuzz_vsr_superblock_free_set)
    ^
???:?:?: 0x2 in ??? (???)
Unwind information for `???:0x2` was not available, trace may be incomplete
```

So this is actually findable now, good!

matklad:
(that's 10th run of ` ./zig/zig build fuzz_vsr_superblock_free_set -Dconfig-log-level=info -- --events-max 90`)

sentientwaffle:
Perfect; I guess I spoke too soon!

matklad (approved):

matklad:
Confirmed that that was indeed this bug (and not some yet unknown indexing update issue)

PR #1331, opened 2023-11-28
vsr: remove storage_size_max

matklad:
Well, not really. `constants.storage_size_max` is still a thing, but it serves only as a hard-limit for the `--storage-size-limit` CLI argument.

`constants.storage_size_max` is also used during tests as a convenient shortcut for "storage size reasonable for tests". Perhaps this could be phased away in favor of an explicit runtime parameter, so that `fuzz_min` config is not needed, but this is mostly orthogonal.

### DIFF
--- a/src/constants.zig
+++ b/src/constants.zig
@@ -468,10 +468,12 @@ comptime {
 /// The maximum size of a local data file.
 /// This should not be much larger than several TiB to limit:
 /// * blast radius and recovery time when a whole replica is lost,
-/// * replicated storage overhead, since all data files are mirrored,
-/// * the size of the superblock storage zone, and
+/// * replicated storage overhead, since all data files are mirrored, and
 /// * the static memory allocation required for tracking LSM forest metadata in memory.
-pub const storage_size_max = config.cluster.storage_size_max;
+///
+/// This is a "firm" limit --- while it is a compile-time constant, it does not affect data file
+/// layout and can be safely changed for an existing cluster.
+pub const storage_size_max = config.process.storage_size_max;
 
 /// The unit of read/write access to LSM manifest and LSM table blocks in the block storage zone.
 ///
--- a/src/testing/storage.zig
+++ b/src/testing/storage.zig
@@ -33,6 +33,7 @@ const FIFO = @import("../fifo.zig").FIFO;
 const constants = @import("../constants.zig");
 const vsr = @import("../vsr.zig");
 const superblock = @import("../vsr/superblock.zig");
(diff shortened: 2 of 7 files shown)


src/vsr/superblock.zig
```diff
@@ -88,11 +88,6 @@ pub const SuperBlockHeader = extern struct {
     /// State stored on stable storage for the Viewstamped Replication consensus protocol.
     vsr_state: VSRState,
 
-    /// The maximum possible size of the data file.
-    /// The maximum allowed runtime storage_size_limit.
-    /// The FreeSet's on-disk size is a function of storage_size_max.
-    storage_size_max: u64,
```
matklad, line 94: :tada:

src/constants.zig
```diff
@@ -468,9 +468,11 @@ comptime {
...
 /// This should not be much larger than several TiB to limit:
 /// * blast radius and recovery time when a whole replica is lost,
-/// * replicated storage overhead, since all data files are mirrored,
-/// * the size of the superblock storage zone, and
+/// * replicated storage overhead, since all data files are mirrored, and
 /// * the static memory allocation required for tracking LSM forest metadata in memory.
+///
+/// This is a "firm" limit --- while it is a compile-time constant, it does not affect data file
+/// layout and can be safely changed for an existing cluster.
 pub const storage_size_max = config.cluster.storage_size_max;
```
sentientwaffle: Should `storage_size_max` be under `config.process` instead of `config.cluster` now? `process` is intended to indicate values that can safely vary between replicas (and typically on the same replica over time).

src/testing/storage.zig
```diff
@@ -54,6 +54,8 @@ const log = std.log.scoped(.storage);
 // - likely that surrounding sectors also corrupt
 // - likely that stuff written at the same time is also corrupt even if written to a far away sector
 pub const Storage = struct {
+    pub const grid_blocks_max = stdx.div_ceil(constants.storage_size_max, constants.block_size);
```
sentientwaffle: Should this use `storage_size_max - Zone.grid.start()` as the numerator? Also, why `div_ceil` instead of `@divFloor`?

matklad: My original thinking was to make this some simple upper-bound -- the actual storage size is defined by the runtime `size` parameter to the `init` anyway, so being overly precise here didn't feel important. 

But, given that for convenience I need to round it to a free set shard size anyhow, I might as well compute this precisely. 

The high order bit here is that this now exists only in the testing code --- `superblock.grid_blocks_max` is no more, because the limit that matters now depends on the runtime value.

sentientwaffle (approved):

PR #1332, opened 2023-11-29
vsr: move free set to the grid

matklad:
As we moved free set into the grid, it's only natural to move free set into the grid!

That is, right now free set _data_ is stored in the grid, _but_ the actual free set Zig struct lives as a field of the SuperBlock. This doesn't make much sense, leads to longish paths like

`free_set_encoded.grid.superblock.free_set`

and, most annoyingly, makes it harder to test the superblock (as it pulls the entire free set in).

With free set in the grid, it becomes possible to be smarter during checkpointing, as the grid could start writing out the free set even _before_ it waits for pending clobber-prone writes to finish --- the grid is now in control of the point where free set "flips" to the next checkpoint.


src/vsr/superblock_free_set_encoded.zig
```diff
@@ -299,68 +305,64 @@ pub fn FreeSetEncodedType(comptime Storage: type) type {
             defer assert(set.callback == .none);
 
             assert(set.block_index == 0);
-            assert(!set.grid.?.superblock.free_set.opened);
+            assert(!set.grid.?.free_set.opened);
```
matklad, line 303: These path changes are annoying (it seems like free set actually used to be in the grid?) and make me with we had a dependency-injection-like setup where logical things like Grid, SuperBlock, Manifest, etc, are just separately allocated in the heap somewhere, and, whenever an X needs access to Y, we just have `y: *Y` field in X, which is set up early during startup, allowing circular references and what not. Perhaps this is a wish better left unfulfilled :-) At minimum, it'll break `@fieldParentPtr` pattern, requiring more explicit contexts for callbacks.

src/vsr/superblock_free_set_encoded.zig
```diff
@@ -159,61 +157,69 @@ pub fn FreeSetEncodedType(comptime Storage: type) type {
...
         ) void {
             set.grid = grid;
-            assert(!set.grid.?.superblock.free_set.opened);
+            assert(!set.grid.?.free_set.opened);
 
             assert(set.callback == .none);
             defer assert(set.callback == .open);
 
-            assert(reference.size % @sizeOf(FreeSet.Word) == 0);
+            assert(reference.free_set_size % @sizeOf(FreeSet.Word) == 0);
```
matklad, line 214: We don't verify `reference.storage_size` at all during the `open` process, but the caller (replica) verifies it after the fact by checking that the current `.checkpoint_reference()` is meta.eql to the starting point from the manifest.

src/vsr/superblock_free_set_encoded.zig
```diff
@@ -159,61 +157,69 @@ pub fn FreeSetEncodedType(comptime Storage: type) type {
             }
             set.* = .{
                 .write_block = set.write_block,
-                .checksum = vsr.checksum(&.{}),
```
matklad, line 162: This was only required for one of the minor fuzzers which faked empty free set. We no longer need that, so I changed to `0`, to serve as "unitialized" state and to avoid confusing empty and not-yet-open free set.

src/vsr/grid.zig
```diff
@@ -549,13 +584,13 @@ pub fn GridType(comptime Storage: type) type {
...
             assert(grid.superblock.opened);
             assert(grid.canceling == null);
-            assert(grid.checkpointing == null);
+            maybe(grid.checkpointing != null); // FreeSetEncoded writes a block.
             assert(grid.writing(block_header.address, block.*) == .not_writing);
             assert(!grid.blocks_missing.repair_waiting(
                 block_header.address,
                 block_header.checksum,
             ));
-            assert(!grid.superblock.free_set.is_free(block_header.address));
```
matklad, line 558: This assert is what motivated lifting `FreeSet.checkpoint()` call from `FreeSetEncoded` to the grid itself.

matklad (commented):
Ah, I also need to rename files & fuzzers...

matklad:
Note: this is the last PR in the free set series

src/vsr/grid.zig
```diff
@@ -248,26 +267,31 @@ pub fn GridType(comptime Storage: type) type {
...
 
-        /// Invoke the callback after all pending repair-writes to blocks that are about to be
-        /// released have completed. This guarantees that there are no outstanding writes to
-        /// (now-)free blocks when we enter the new checkpoint.
+        /// Checkpointing the grid:
+        /// 1. Writes the free set to disk.
+        /// 2. Awaits all pending repair-writes to blocks that were just freed. This guarantees that
+        ///    there are no outstanding writes to (now-)free blocks when we enter the new checkpoint.
+        /// 3. Marks blocks released during the current checkpoint as free and eligible for
+        ///    acquisition in the next checkpoint.
```
sentientwaffle, line 280:
```suggestion
        ///    acquisition in the next checkpoint.
        /// 4. Marks the free set's own blocks as released (but not yet free).
```

src/vsr/superblock_free_set.zig
```diff
@@ -448,6 +448,14 @@ pub const FreeSet = struct {
...
         // also linear.
         set.verify_index();
+
+        var address_previous: u64 = 0;
+        for (free_set_encoded_blocks) |address| {
+            assert(address > 0);
+            assert(address > address_previous);
+            address_previous = address;
+            set.release(address);
+        }
```
sentientwaffle, line 458: Ah, so we can't use `.mark_released()` here because, unlike in `open()`, the bits are already set in `blocks`.

src/vsr/grid.zig
```diff
@@ -287,17 +316,23 @@ pub fn GridType(comptime Storage: type) type {
...
-                assert(!grid.superblock.free_set.is_released(iop.write.address));
+                assert(!grid.free_set.is_free(iop.write.address));
+                assert(!grid.free_set.is_released(iop.write.address));
             }
 
+            // Now that there are no writes to released blocks, we can safely mark them as free.
+            // This concludes grid checkpointing.
+            grid.free_set.checkpoint(
+                grid.free_set_encoded.block_addresses[0..grid.free_set_encoded.block_count],
+            );
```
sentientwaffle, line 334: Maybe bring in (from replica, where we also assert this):

```
            assert(grid.free_set.count_released() == grid.free_set_encoded.block_count);
```

src/vsr/replica.zig
```diff
@@ -682,19 +681,23 @@ pub fn ReplicaType(
...
         fn free_set_open_callback(free_set_encoded: *FreeSetEncoded) void {
-            const superblock = @fieldParentPtr(SuperBlock, "free_set_encoded", free_set_encoded);
-            const self = @fieldParentPtr(Self, "superblock", superblock);
+            const grid = @fieldParentPtr(Grid, "free_set_encoded", free_set_encoded);
+            const self = @fieldParentPtr(Self, "grid", grid);
             assert(!self.state_machine_opened);
             assert(self.commit_stage == .idle);
             assert(self.syncing == .idle);
             assert(self.sync_tables == null);
             assert(self.grid_repair_tables.executing() == 0);
```
sentientwaffle, line 690:
```suggestion
            assert(self.grid_repair_tables.executing() == 0);
            assert(self.grid.free_set.count_released() == self.grid.free_set_encoded.block_count);
```

This serves as a nice hint that our state here matches the state immediately after the same checkpoint.

src/vsr/superblock_free_set_encoded.zig
sentientwaffle: Would `undefined` work, or is it expected to be `0`?

src/vsr/superblock.zig
```diff
@@ -521,10 +522,11 @@ pub const FreeSetReference = struct {
     checksum: u128,
     last_block_address: u64,
     last_block_checksum: u128,
-    size: u32,
+    free_set_size: u32,
+    storage_size: u64,
 
     pub fn empty(reference: *const FreeSetReference) bool {
```
sentientwaffle, line 528: Assert `storage_size >= data_file_size_min`.
And in the `else` branch below, we can assert that it is strictly greater.

src/vsr/superblock_free_set_encoded.zig
matklad: Technically, it will work! But I am wondering if I should use it here... If I set `checksum` to undefined, then, by the same token, I need to `undefined` all other fields, like `size`, and `block_count`, and what not. The init just allocates memory for all the fields, and we only learn the proper values of the fields in `open` and `checkpoint`.

But setting most of the things to `undefined` makes me uneasy --- I _think_ the behavior we'll get there would be implementation defined. Eg, one version of Zig sets it to `0xaa`, but another one  decides on `0x92`. 

So, if there _is_ a bug, and we read "unitialized" value in prod on some codepath, I think I'd prefer it to be deterministically zero, and rely on our asserts to catch us out. 

But I don't really have consensus with myself here, I can go either way! I guess the main conflict for me is that:

* `undefined` signals the right intention, and I'd love to use it for code clarity reasons.
* the runtime behavior of undefined is ... not sufficiently defined for my tastes!

src/vsr/superblock_free_set.zig
matklad: Yeah, I also don't particularly like this semi-duplication between `open -> mark_release` and `checkpoint`, but I couldn't find an obvious way to eliminate this. 

The current implementation also relies heavily on `include_staging / exuclude_staging` hack, which I am also not a fan of.

_But_ I actually like how we atomically checkpoint free set at the end in this PR, such that all prior grid operations (repairs&creation of free set own blocks) operate in the old checkpoint.

src/vsr/superblock.zig
matklad: Whoa, this is dead code! I definitely should be using this function though, to centralize assertions!

build.zig
```diff
@@ -597,8 +597,8 @@ pub fn build(b: *std.Build) !void {
             .description = "Fuzz the SuperBlock. Args: [--seed <seed>] [--events-max <count>]",
         },
         .{
-            .name = "fuzz_vsr_superblock_free_set",
-            .file = "src/vsr/superblock_free_set_fuzz.zig",
+            .name = "fuzz_vsr_free_set",
+            .file = "src/vsr/free_set_fuzz.zig",
             .description = "Fuzz the SuperBlock FreeSet. Args: [--seed <seed>]",
```
sentientwaffle:
```suggestion
            .description = "Fuzz the FreeSet. Args: [--seed <seed>]",
```

src/vsr/grid.zig
```diff
@@ -549,13 +586,13 @@ pub fn GridType(comptime Storage: type) type {
             const block_header = schema.header_from_block(block.*);
             assert(grid.superblock.opened);
             assert(grid.canceling == null);
-            assert(grid.checkpointing == null);
+            maybe(grid.checkpointing != null); // FreeSetEncoded writes a block.
```
sentientwaffle: We could tighten this by looking at the block type, i.e.

```
assert((grid.checkpointing != null) == (block type == .free_set))
```

src/vsr/grid.zig
```diff
@@ -248,26 +267,32 @@ pub fn GridType(comptime Storage: type) type {
...
-            grid.blocks_missing.checkpoint_commence(&grid.superblock.free_set);
-
             grid.checkpointing = .{ .callback = callback };
-            grid.on_next_tick(checkpoint_tick_callback, &grid.checkpointing_tick_context);
+            grid.blocks_missing.checkpoint_commence(&grid.free_set);
+            grid.free_set_encoded.checkpoint(checkpoint_free_set_callback);
         }
 
-        fn checkpoint_tick_callback(next_tick: *Grid.NextTick) void {
-            const grid = @fieldParentPtr(Grid, "checkpointing_tick_context", next_tick);
```
sentientwaffle, line 270: The `checkpointing_tick_context` field can be removed from `Grid` now.

src/vsr/free_set_encoded.zig
```diff
@@ -101,7 +98,7 @@ pub fn FreeSetEncodedType(comptime Storage: type) type {
         size_transferred: u32 = 0,
 
         // Checksum covering the entire encoded free set.
-        checksum: u128,
+        checksum: u128 = 0,
```
sentientwaffle, line 101: `init()` sets `checksum` to `.checksum = vsr.checksum(&.{}),`, but `reset` sets it to the default `0`.

src/vsr/grid.zig
matklad: Ah, this even found a bug in the tree fuzzer (it was still checkpointing free set directly, not the grid)!

sentientwaffle (approved):

PR #1333, opened 2023-11-29
Found ancient relics from 2021

batiati:
As Indiana Jones would say, it belongs in a museum!

This piece of code is a stub for RangeQueries, a long-desired feature that is coming soon!


jorangreef:
You know, when we started on this code, I was thinking: “One day we'll look back and the storage engine will be there.”

jorangreef (approved):
You call this archeology?! It belongs in a museum! 😉

PR #1334, opened 2023-11-30
vsr: make `comptime checksum(&.{})` available everywhere

matklad:
I have a bunch of asserts on the empty checksum in free set, and writing `assert(checksum = vsr.checksum(&.{}))` makes me hesitate, as that involves non-trivial runtime computation. So let's make

    comptime vsr.checksum(&.{})

JustWork, so that's a no-brainer for the future.

### DIFF
--- a/src/vsr/checksum.zig
+++ b/src/vsr/checksum.zig
@@ -48,9 +48,20 @@ fn seed_init() void {
 // Lazily initialize the Aegis State instead of recomputing it on each call to checksum().
 // Then, make a copy of the state and use that to hash the source input bytes.
 pub fn checksum(source: []const u8) u128 {
+    if (@inComptime()) {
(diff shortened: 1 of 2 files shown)


src/vsr/checksum.zig
```diff
@@ -48,11 +48,22 @@ fn seed_init() void {
...
+    }
     var stream = ChecksumStream.init();
     stream.add(source);
     return stream.checksum();
 }
 
+test "checksum empty" {
+    var stream = ChecksumStream.init();
+    stream.add(&.{});
+    assert(stream.checksum() == comptime checksum(&.{}));
```
batiati: Use `expectEqual` here

matklad:
I think this is an improvement as-is, I'll update free-set parts separately, when #1332 merges.

batiati (approved):

PR #1338, opened 2023-12-04
vsr: tighten checkpoint validity assert

matklad:
One checkpoint spans multiple lsm_batch_multiples

### DIFF
--- a/src/vsr.zig
+++ b/src/vsr.zig
@@ -1397,6 +1397,10 @@ pub const Checkpoint = struct {
     }
 
     pub fn valid(op: u64) bool {
+        // Divide by `lsm_batch_multiple` instead of `journal_slot_count - lsm_batch_multiple`:
+        // although today in practice checkpoints are evenly spaced, the LSM layer doesn't assume
+        // that. LSM allows any bar boundary to become a checkpoint which happens, e.g., in the tree
+        // fuzzer.
         return op == 0 or (op + 1) % constants.lsm_batch_multiple == 0;
     }
 };


src/vsr.zig
```diff
@@ -1399,7 +1399,8 @@ pub const Checkpoint = struct {
     }
 
     pub fn valid(op: u64) bool {
-        return op == 0 or (op + 1) % constants.lsm_batch_multiple == 0;
+        return op == 0 or
+            (op + 1) % (constants.journal_slot_count - constants.lsm_batch_multiple) == 0;
```
matklad: Is there a good name for `constants.journal_slot_count - constants.lsm_batch_multiple`? I'd love to extract this into something like `constants.vsr_checkpoint_op_count`, as this constant captures an important conceptual thing, but I am not 100% positive on the name...

sentientwaffle: I don't think we have a term for it right now. 

Brainstorming:
- "checkpoint span"
- "checkpoint period"

But I think both of those could be confused with the distance between op_checkpoint and the following checkpoint trigger, (even though that is equal to `journal_slot_count`. :thinking:

matklad:
Ah, so the tree fuzzer fails, because the code _doesn't_ assume evenly-spaced checkpoints, and allows any bar boundary to be a checkpoint, giving elastic checkpoints. 

I am now curious what's our intended design here:

* do we want to allow checkpointing at any bar boundary, with the evenly spaced checkpoints being an implementation details?
* or maybe we want to hard-code evenly spaced checkpoints? 

In other words, if today replicas start to vary the interval between checkpoints, would anything break?

sentientwaffle:
> do we want to allow checkpointing at any bar boundary, with the evenly spaced checkpoints being an implementation details?

There are cases where this could be useful, e.g. force a checkpoint immediately before an upgrade, to minimize the recovery time when the cluster restarts. Though you could also emulate that behavior by just committing a bunch of noop prepares until the normal checkpoint boundary.

> In other words, if today replicas start to vary the interval between checkpoints, would anything break?

At the LSM level, we don't care about the checkpoint interval (just the bar boundaries). Imo this simplifies the mental model -- LSM shouldn't need to know/care about the WAL's size.

But at the replica/VSR level, we currently do care about the interval (though I don't think this is technically necessary). We use `op_checkpoint_next_trigger` for example, to determine the absolute highest op that we could possibly have ever written to our log.

matklad:
reframed as a comment explaining the current code!

sentientwaffle (approved):

PR #1339, opened 2023-12-04
VSR: Add checkpoint_id to Header.Prepare 

sentientwaffle:
## Context

`Header.PrepareOk` includes a `checkpoint_id`, but `Header.Prepare` does not.
Replicas ignore `PrepareOk`s from checkpoint ids that do not match their own.
This _mostly_ works, but not 100%.

As mentioned in the ["Next Steps" of the first state sync PR](https://github.com/tigerbeetle/tigerbeetle/pull/834):

> - 256-byte message headers: Make room for the checkpoint identifier in prepare messages. This will let us guarantee that a replica's history never diverges from the canonical history by more than one checkpoint.

Without `checkpoint_id` in prepares, it is technically possible (albeit very unlikely!) for a replica to diverge from the canonical history by more than one checkpoint.
(If a replica partitioned from receiving `command=commit` or `command=ping` messages, but is able to receive `command=prepare` messages, then it will keep preparing + committing messages without ever learning whether it has diverged. So it can diverge by an arbitrary number of checkpoints from the canonical history.)

## Solution

- Add `checkpoint_id` to `Header.Prepare`.
- Replicas ignore `Prepare`s from checkpoint ids that do not match their own.
- `on_prepare_ok` can now assert that the checkpoint id matches the respective `Prepare`'s.

`on_repair()` doesn't need to test the `prepare.checkpoint_id` – it can assert it.
(`on_repair` requires an intact hash chain to the head, and the head's checkpoint id _was_ tested (via `on_prepare`).)

## Also included:

- `checkpoint_id` u256 → u128: `checkpoint_id` doesn't refer to any (future) AEAD tag (i.e. message checksum), so it can just be a `u128` hash (even when 256-bit encryption is in use). This also means that `Header.Prepare` now has enough reserved bytes for the checkpoint id.
- Rearrange `CheckpointState` fields for future u256 alignment: Right now we store "message" "checksums" as two `u128`s (the latter is an always-0 padding). In the future, those will be combined into a single `u256`. Shift the actual `u128`s later in the struct so that all of the "u256"s are 32-byte aligned.

### DIFF
--- a/src/vsr/replica.zig
+++ b/src/vsr/replica.zig
@@ -6232,7 +6232,7 @@ pub fn ReplicaType(
                     @bitCast(Header.PrepareOk{
                         .command = .prepare_ok,
                         .checkpoint_id = checkpoint_id,
-                        // .parent = header.parent,
+                        .parent = header.parent,
                         .client = header.client,
                         .prepare_checksum = header.checksum,
                         .request = header.request,
@@ -9369,7 +9369,7 @@ const PipelineQueue = struct {
             ok.header.prepare_checksum,
         ) orelse return null;
         assert(prepare.message.header.command == .prepare);
-        // assert(prepare.message.header.parent == ok.header.parent);
+        assert(prepare.message.header.parent == ok.header.parent);
         assert(prepare.message.header.client == ok.header.client);
         assert(prepare.message.header.request == ok.header.request);
         assert(prepare.message.header.cluster == ok.header.cluster);
@@ -9380,6 +9380,7 @@ const PipelineQueue = struct {
         assert(prepare.message.header.commit == ok.header.commit);
         assert(prepare.message.header.timestamp == ok.header.timestamp);
         assert(prepare.message.header.operation == ok.header.operation);
+        // TODO compare checkpoint id with prepare's
 
         return prepare;
     }
--- a/src/vsr/replica.zig
+++ b/src/vsr/replica.zig
@@ -1341,6 +1341,34 @@ pub fn ReplicaType(
                 return;
             }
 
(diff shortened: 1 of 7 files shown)


src/vsr/replica.zig
```diff
@@ -5597,6 +5612,12 @@ pub fn ReplicaType(
                 assert(self.journal.has(header));
             }
 
+            if (self.checkpoint_id_for_op(header.op)) |checkpoint_id| {
+                assert(header.checkpoint_id == checkpoint_id);
+            } else {
+                assert(header.op < self.op_repair_min());
```
matklad: I don't get this assert in the `else` branch, I think it's tautologically false here? 

```
            if (header.op < self.op_repair_min()) {
                // Slots too far back belong to the next wrap of the log.
                log.debug("{}: repair_header: false (precedes op_repair_min={})", .{
                    self.replica,
                    self.op_repair_min(),
                });
                return false;
            }

            if (self.checkpoint_id_for_op(header.op)) |checkpoint_id| {
                assert(header.checkpoint_id == checkpoint_id);
            } else {
                assert(header.op < self.op_repair_min());
            }
```

matklad:
>Rearrange CheckpointState fields for future u256 alignment:

Let's write a comptime assert that checks that:

- any `u128` field is aligned to `@sizeOf(u128)`
- any `u128` field which hash a `_padding` counterpart is aligned to  `@sizeOf(u256)`

src/vsr/replica.zig
sentientwaffle: Good point, I forgot about that earlier branch! I'll change the assert to:

```
            assert(header.checkpoint_id == self.checkpoint_id_for_op(header.op).?);
```

matklad (approved):

PR #1343, opened 2023-12-06
node: add unit tests to ci

kprotty:
They were written but not hooked up to CI to actually run.


src/clients/node/scripts/test.sh
batiati, line 1: Whenever possible, we're moving from .sh scripts to `./zig/zig build run_with_tb -- node $(pwd)/myscript.js` (in some cases, the test runner still manages the TB instance)

kprotty: Should we be using `shell.zig("build run_with_tb ...")` or `TmpTigerBeetle` in `ci.zig` code?

batiati: I guess `TmpTigerBeetle` to avoid "zig build inception" 😉

src/clients/node/ci.zig
```diff
@@ -11,13 +11,15 @@ const TmpTigerBeetle = @import("../../testing/tmp_tigerbeetle.zig");
...
-    // We have some unit-tests for node, but they are likely bitrotted, as they are not run on CI.
-
     // Integration tests.
 
     // We need to build the tigerbeetle-node library manually for samples to be able to pick it up.
     try shell.exec("npm install", .{});
 
+    // Run node client unit tests and benchmark after installing tigerbeetle-node
+    try shell.exec("npm run test", .{});
+    try shell.exec("npm run benchmark", .{});
```
matklad: How long does it take to run the benches? These tests gate the main branch, so they must be very fast (granted, currently our overall test suite isn't nearly fast enough, but the direction we want to go is "faster tests"). 

Usually, the benchmarks are quite slow, so they are removed from the CI, _or_ are being run with artificially low iteration counts, so that we don't _actually_ benchmark anything, but just smoke test that there are no errors. 

That is to say: if `nmp run benchmark` is significatnly slower than just `npm run test`, we need to do something here. If it's fast enough as is, than that's fine already!

kprotty: `test` and `benchmark` are effectively instant (the latter only makes `10` transfers with batch size `5120`, so acts more like a smoke-test on throughput)

src/clients/node/src/benchmark.ts
```diff
@@ -4,19 +4,20 @@ import {
...
+
+const getEnvVar = (name: string): string => {
+  if (!process.env[name]) throw new Error(`environment variable ${name} required`)
+  return process.env[name] as string
+}
 
 const client = createClient({
   cluster_id: 0n,
-  replica_addresses: ['3001']
+  replica_addresses: [getEnvVar('TB_ADDRESS')]
```
batiati: nit: can we default to 3000 if no environment variable was found?
The other clients do that.

kprotty: Was thinking about it, but presumed it could hamper debugging. If other clients already do so, it should be fine.

batiati (approved):

PR #1344, opened 2023-12-07
cli: add --in-memory flag

cb22:
Allow running with an in-memory data file, which is formatted automatically.

### DIFF
--- a/src/io/darwin.zig
+++ b/src/io/darwin.zig
@@ -662,7 +662,23 @@ pub const IO = struct {
 
     pub const INVALID_FILE: os.fd_t = -1;
 
-    /// Opens or creates a journal file:
+    /// Creates a virtual file backed by memory.
+    pub fn open_memory_file() !os.fd_t {
+        const flags: u32 = os.O.RDWR | os.O.EXCL | os.O.CREAT;
+
+        // Unlike on Linux where the filename passed in to memfd_create doesn't matter, shm_open
+        // names are shared between all processes. Immediately unlink the filename after creating
+        // it, and we open it with O_EXCL and O_CREAT to ensure that if a previous process somehow
+        // didn't unlink it, we won't clobber anything. We could add a random suffix if it's a
+        // concern.
+        const fd: os.fd_t = @intCast(os.darwin.shm_open("tigerbeetle-memory-file", flags, 0o600));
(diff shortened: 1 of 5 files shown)


src/io/darwin.zig
```diff
@@ -662,7 +662,23 @@ pub const IO = struct {
 
     pub const INVALID_FILE: os.fd_t = -1;
 
-    /// Opens or creates a journal file:
+    /// Creates a virtual file backed by memory.
+    pub fn open_memory_file() !os.fd_t {
```
cb22, line 666: Need to actually check this works on a Mac!

src/tigerbeetle/cli.zig
```diff
@@ -258,6 +260,9 @@ pub fn parse_args(allocator: std.mem.Allocator) !Command {
                     .{ storage_size_limit, constants.sector_size },
                 );
             }
+            if (start.in_memory and !std.mem.eql(u8, start.positional.path, "in-memory")) {
```
cb22, line 263: Felt like adding optional positional arguments for this one use case would be a bit much, so instead you have to run:

```
tigerbeetle start --addresses=127.0.0.1:3000 --in-memory in-memory
```

which I think is OK for a hidden dev only option.

matklad: Actually, let me take my :+1: back here -- I think we would want our users to use this option in their tests, no? So this would be our official surface area.

Perhaps this wants to be a separate command even, as this is actually `format-and-start`?

jorangreef: As a separate command is even better: `tigerbeetle temporary`

matklad: Thinking about this: can `--in-memory` also imply "kill oneself once stdin closes"? That'll get rid of a bunch of tricky code in tmp_tigerbeetle and language-specific equivalents? 

And maybe also: `--in-memory` makes us print our port to stdout, and then close it. This would allow removing _a lot_ of tricky code to asynchronously read stderr to sniff the port out.

That is, we could think about this as a general `tigerbeetle testing` which does everything testing-related, rather than just in-memory. 

But probably makes sense to start small!

matklad:
Let's close for now --- I don't think we have a burning desire for this, and, with --development flag and white-magic optimizations of our memory usage, I think this is not that high priority.

PR #1346, opened 2023-12-07
vsr: increase prepare durability near checkpoint boundary

matklad:
This fixes the following simulator failure:

$ git reset --hard 87cace1
$ ./zig/zig build simulator_run -Dsimulator-state-machine=testing -- 9854840972719737644

This is what happening:

- two replicas committed checkpoint trigger (op=59) and prepared the next op=60.
- two replicas are far behind, want to state sync, and need a canonical checkpoint.
- one replica has op=28 corrupted in the log and wants to repair it

The key issue is that preparing op=60 already pushes op=28 out from the log, but doesn't yet make the checkpoint canonical (as op=60 isn't committed). As a result:

- the two lagging replicas can't sync, because there's no canonical checkpoint.
- the repairing replica can't catch up, because corrupted op is missing from the cluster.
- the two up-to-date replicas can't commit op=60, because there isn't a prepare quorum with a fresh checkpoint.

Taking a step back, the core issue here is that we don't have enough _physical_ durability for op=28. _Logically_, it is present on three replicas, once in a WAL (where it got corrupted) and twice as data in the checkpoint. However, as these are two different representations, the cluster is not always capable of reconciling them.

It is useful to consider a different scenario of corruptions here: suppose that a single grid block is corrupted on both checkpoints, and some prepare  is corrupted in the log of a catching up replica. There are three corruptions, but as only two are correlated, the cluster should be live. However in this situation effectively any corruption in the grid "intersects" with any corruption in the WAL.

The solution here is to _extend_ durabilities of prepares, such that we only start overwriting prepares from the old checkpoint once the new one is confirmed canonical. Before, the WAL looked like this:

```
    |<-checkpoint->|<-lsm_bar->|
                               ^
                            trigger
```

We shorten the checkpoint part to make sure we can fit an extra pipeline worth of messages in:

```
    |<-checkpoint->|<-lsm_bar->|<-pipeline->|
                               ^
                            trigger
```

This gives us the following invariant:

A prepare can only be evicted from the WAL once there's a _quorum_ of checkpoints covering this prepare.

### DIFF
--- a/src/simulator.zig
+++ b/src/simulator.zig
@@ -632,30 +632,50 @@ pub const Simulator = struct {
         }
     }
 
-    // Check if the core is stuck in a view change because replicas can't converge to a canonical
-    // checkpoint. This can happen in two scenarios:
+    // Check if the core is stuck because replicas can't converge to a canonical checkpoint. This
+    // can happen in two scenarios:
     //
-    // - There is a canonical checkpoint, but replicas can't learn about because commit messages
-    //   don't flow during view change.
-    // - There is no canonical checkpoint at all --- one replica is ahead of others by a checkpoint,
-    //   but there are no commits on top of that checkpoint.
+    // A) There is no canonical checkpoint at all --- one replica is ahead of others by a
+    //    checkpoint, but there are no commits on top of that checkpoint.
+    // B) During view change, there is a canonical checkpoint, but replicas can't learn that it is
+    //    in fact canonical, because commit messages don't flow, so commit_max is not advanced,
+    //    and there are not enough replicas at a checkpoint for the quorum condition.
     //
     // TODO: both of the above genuine liveness issues to be fixed with async checkpoints, which
     // would allow to sync to the _previous_ checkpoint, guaranteed to be canonical.
     fn core_missing_checkpoint(simulator: *const Simulator) bool {
+        var core_replicas_total: u8 = 0;
+        var core_replicas_normal: u8 = 0;
+        var commit_max: u64 = 0;
+        var checkpoint_max: u64 = 0;
         var core_checkpoints = stdx.BoundedArray(u128, constants.replicas_max){};
         for (simulator.cluster.replicas) |replica| {
             if (!simulator.core.isSet(replica.replica)) continue;
             if (replica.standby()) continue;
-
-            if (replica.status == .normal) return false;
+            core_replicas_total += 1;
+            core_replicas_normal += @intFromBool(replica.status == .normal);
+            commit_max = @max(commit_max, replica.commit_max);
+            checkpoint_max = @max(
+                checkpoint_max,
+                replica.superblock.working.vsr_state.checkpoint.commit_min,
+            );
             core_checkpoints.append_assume_capacity(replica.superblock.working.checkpoint_id());
         }
+
+        assert(core_replicas_normal == core_replicas_total or core_replicas_normal == 0);
 
         for (0..core_checkpoints.count()) |i| {
             for (0..i) |j| {
                 if (core_checkpoints.get(i) != core_checkpoints.get(j)) {
-                    return true;
+                    if (core_replicas_normal == core_replicas_total) {
+                        // Case A) all replicas are normal, but commit_max is exactly checkpoint
+                        // trigger.
+                        return vsr.Checkpoint.trigger_for_checkpoint(checkpoint_max) == commit_max;
+                    } else {
+                        // Case B) all replicas are in a view change or recovering head, they
+                        // can't learn which checkpoint is canonical without commit messages.
+                        return true;
+                    }
                 }
             }
         }
--- a/src/vsr.zig
+++ b/src/vsr.zig
@@ -1289,7 +1289,8 @@ const ViewChangeHeadersArray = struct {
         assert(headers.command == .start_view);
         // This function is only called by a replica that is lagging behind the primary's
         // checkpoint, so the start_view has a full suffix of headers.
-        assert(headers.array.get(0).op >= constants.journal_slot_count);
+        assert(headers.array.get(0).op >=
+            constants.vsr_checkpoint_interval + constants.lsm_batch_multiple);
         assert(headers.array.count() >= constants.view_change_headers_suffix_max);
         assert(headers.array.count() >= constants.pipeline_prepare_queue_max + 1);
 
@@ -1337,7 +1338,8 @@ const ViewChangeHeadersArray = struct {
     }
 };
 
-/// For a replica with journal_slot_count=8 and lsm_batch_multiple=2:
+/// For a replica with journal_slot_count=9, lsm_batch_multiple=2, pipeline_prepare_queue_max=1, and
+/// checkpoint_interval = journal_slot_count - (lsm_batch_multiple + pipeline_prepare_queue_max) = 6
 ///
 ///   checkpoint() call           0   1   2   3
 ///   op_checkpoint               0   5  11  17
@@ -1345,11 +1347,11 @@ const ViewChangeHeadersArray = struct {
 ///   op_checkpoint_next_trigger  7  13  19  25
 ///
 ///     commit log (ops)           │ write-ahead log (slots)
-///     0   4   8   2   6   0   4  │ 0---4---
-///   0 ─────✓·%                   │ 01234✓6%   initial log fill
-///   1 ───────────✓·%             │ 890✓2%45   first wrap of log
-///   2 ─────────────────✓·%       │ 6✓8%0123   second wrap of log
-///   3 ───────────────────────✓·% │ 4%67890✓   third wrap of log
+///     0   4   8   2   6   0   4  │  0  -  -  -  4  -  -  -  8
+///   0 ─────✓·%                   │[ 0  1  2  3  4  ✓] 6  %  R
+///   1 ───────────✓·%             │  9 10  ✓]12  %  5[ 6  7  8
+///   2 ─────────────────✓·%       │ 18  % 11[12 13 14 15 16  ✓]
+///   3 ───────────────────────✓·% │[18 19 20 21 22  ✓]24  % 17
 ///
 /// Legend:
 ///
@@ -1357,7 +1359,8 @@ const ViewChangeHeadersArray = struct {
 ///   ·/%  op in memory at checkpoint
 ///     ✓  op_checkpoint
 ///     %  op_checkpoint's trigger
-///
+///     R  slot reserved in WAL
+///    [ ] range of ops from a checkpoint
 pub const Checkpoint = struct {
     comptime {
         assert(constants.journal_slot_count > constants.lsm_batch_multiple);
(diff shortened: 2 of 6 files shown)


src/simulator.zig
```diff
@@ -632,30 +632,50 @@ pub const Simulator = struct {
...
         }
 
         for (0..core_checkpoints.count()) |i| {
             for (0..i) |j| {
                 if (core_checkpoints.get(i) != core_checkpoints.get(j)) {
-                    return true;
+                    if (normal) {
+                        // Case A) all replicas are normal, but commit_max is exactly checkpoint
+                        // trigger.
+                        return vsr.Checkpoint.trigger_for_checkpoint(checkpoint_max) == commit_max;
```
matklad, line 673: I spent an embarassing amount of time trying to find a bug in my changes that fails

```
zig/zig build simulator_run -Dsimulator-state-machine=testing -Dsimulator-log=short -Doptimize=ReleaseSafe -- 15383775152417957975
```

only to realize that this is a pre-existing failure! 

Before, we were only checking for missing checkpoint in view-change, but it could be the case that in normal one replica is ahead, and is at trigger.

src/simulator.zig
```diff
@@ -632,30 +632,50 @@ pub const Simulator = struct {
...
+    // can happen in two scenarios:
     //
-    // - There is a canonical checkpoint, but replicas can't learn about because commit messages
-    //   don't flow during view change.
-    // - There is no canonical checkpoint at all --- one replica is ahead of others by a checkpoint,
-    //   but there are no commits on top of that checkpoint.
+    // A) There is no canonical checkpoint at all --- one replica is ahead of others by a
+    //    checkpoint, but there are no commits on top of that checkpoint.
+    // B) During view change, there is a canonical checkpoint, but replicas can't learn about
+    //    because commit messages don't flow during view change.
```
```suggestion
    // B) During view change, there is a canonical checkpoint, but replicas can't learn about it
    //    because commit messages don't flow during view change.
```

Typo. But also -- is this correct? Checkpoint ids are on ping messages too.

src/simulator.zig
```diff
@@ -632,30 +632,50 @@ pub const Simulator = struct {
...
+        var commit_max: u64 = 0;
+        var checkpoint_max: u64 = 0;
         var core_checkpoints = stdx.BoundedArray(u128, constants.replicas_max){};
         for (simulator.cluster.replicas) |replica| {
             if (!simulator.core.isSet(replica.replica)) continue;
             if (replica.standby()) continue;
 
-            if (replica.status == .normal) return false;
+            if (replica.status == .normal) {
+                normal = true;
```
sentientwaffle: If we are checking that _all_ replicas are normal (or none are), then we need an assert here too (e.g. by initializing `normal` to `?bool = null` & asserting it is null before transitioning to true.

src/vsr/replica.zig
```diff
@@ -1849,7 +1849,7 @@ pub fn ReplicaType(
             assert(message.header.replica == self.primary_index(message.header.view));
             assert(message.header.commit >= message.header.checkpoint_op);
             assert(message.header.commit - message.header.checkpoint_op <=
-                constants.journal_slot_count);
+                constants.journal_slot_count); // FIXME:
```
sentientwaffle: Did you mean to add a note here?

src/vsr/replica.zig
```diff
@@ -4015,10 +4015,8 @@ pub fn ReplicaType(
             // The SV includes headers corresponding to the triggers for preceding
             // checkpoints (as many as we have and can help repair, which is at most 2).
             for ([_]u64{
-                self.op_checkpoint_next_trigger() -|
-                    (constants.journal_slot_count - constants.lsm_batch_multiple),
-                self.op_checkpoint_next_trigger() -|
-                    (constants.journal_slot_count - constants.lsm_batch_multiple) * 2,
+                self.op_checkpoint_next_trigger() -| constants.vsr_checkpoint_interval,
+                self.op_checkpoint_next_trigger() -| constants.vsr_checkpoint_interval * 2,
```
sentientwaffle, line 4050: So much more readable! :heart:

src/vsr.zig
```diff
@@ -1374,11 +1375,11 @@ pub const Checkpoint = struct {
         const result = op: {
             if (checkpoint == 0) {
                 // First wrap: op_checkpoint_next = 8-2-1 = 5
-                break :op constants.journal_slot_count - constants.lsm_batch_multiple - 1;
+                break :op constants.vsr_checkpoint_interval - 1;
             } else {
                 // Second wrap: op_checkpoint_next = 5+8-2 = 11
                 // Third wrap: op_checkpoint_next = 11+8-2 = 17
```
sentientwaffle: Can you update these "First/Second/Third wrap" notes to include a pipeline?
Also the diagram above `pub const Checkpoint`

src/vsr.zig
```diff
@@ -1293,7 +1293,8 @@ const ViewChangeHeadersArray = struct {
         assert(headers.command == .start_view);
         // This function is only called by a replica that is lagging behind the primary's
         // checkpoint, so the start_view has a full suffix of headers.
-        assert(headers.array.get(0).op >= constants.journal_slot_count);
```
matklad, line 1292: TBH, I don't fully understand this assert. Is it checking that, if the primary is on the  "next" checkpoint, then we have at least checkpoint+trigger worth of messages? 

But is this actually correct? The first checkpoint is a bit shorter, due to ` - 1`, no?

src/simulator.zig
matklad: Replicas learn checkpoint it itself, but they can't learn that it is canonical, because they don't know `commit_max`.

sentientwaffle: If there are enough replicas on the new checkpoint though, then we don't need to know `commit_max`:

```
                if (candidates_matching >= candidates_threshold) {
                    // Fallback case:
                    // In an idle cluster, where the last commit landed on a checkpoint boundary,
                    // we still want to be able to sync a lagging replica.
                    log.debug("{}: on_{s}: jump_sync_target: " ++
                        "candidate checkpoint is canonical (quorum: {}/{})", .{
                        self.replica,
                        @tagName(header.command),
                        candidates_matching,
                        candidates_threshold,
                    });
                    break :canonical true;
                }
```

src/vsr.zig
sentientwaffle: Thinking out loud:

`start_view_into_do_view_change` is called by a replica that is transitioning from normal -> view_change, but is lagging behind the primary by at least a checkpoint, and the SV it has from the old view is from that newer checkpoint.

Using the checkpoint diagram for the purpose of mental model...:

```
///     commit log (ops)           │ write-ahead log (slots)
///     0   4   8   2   6   0   4  │ 0---4---
///   0 ─────✓·%                   │ 01234✓6%   initial log fill
///   1 ───────────✓·%             │ 890✓2%45   first wrap of log
///   2 ─────────────────✓·%       │ 6✓8%0123   second wrap of log
///   3 ───────────────────────✓·% │ 4%67890✓   third wrap of log
```

- In the first wrap, that means that our `self.op < 8`, so `headers.array.get(0).op >= 8`. `8 - 1` is our trigger, so it is the highest op that could possibly be in our log.
- So I think that the old assert was correct

src/simulator.zig
matklad: Adjusted comment:

* there _aren't_ enough replicas in core at a given checkpoint, so they can't prove checkpoint canonical from pings
* and they also can't learn about canonicity from commit messages, because there aren't any

src/vsr/replica.zig
```diff
@@ -5051,19 +5049,40 @@ pub fn ReplicaType(
...
+                        // Case 2: op is from a too distant past for us to know its checkpoint id.
+                        return null;
+                    }
+                    // Case 3: op is from the previous checkpoint whose id we still remember.
                     return self.superblock.working.vsr_state.checkpoint.previous_checkpoint_id;
                 }
+
+                assert(op + constants.vsr_checkpoint_interval > self.op_checkpoint_next_trigger());
             }
-            return self.superblock.working.checkpoint_id();
```
matklad, line 5066: I think this wasn't entirely correct actually! For ops far in the future, we'd use current checkpoint ID, rather than `null`.

src/vsr.zig
matklad: Got it! That `-1` in the computation of the frist checkpoint threw me into a loop. I was thinking we have some fancy logic where `op=0` doesn't actually go into the loog, because it's a root op or something (so, the first checkpoint is physically shorter than the rest). But that `-1` is just count <-> index conversion --- the first checkpoint has exactly `checkpoint_interval` ops (like all the other checkpoints), the first op number is 0, the last op number is `len - 1`

src/vsr.zig
```diff
@@ -1337,27 +1338,29 @@ const ViewChangeHeadersArray = struct {
...
 ///
 ///   checkpoint() call           0   1   2   3
 ///   op_checkpoint               0   5  11  17
 ///   op_checkpoint_next          5  11  17  23
 ///   op_checkpoint_next_trigger  7  13  19  25
 ///
 ///     commit log (ops)           │ write-ahead log (slots)
-///     0   4   8   2   6   0   4  │ 0---4---
-///   0 ─────✓·%                   │ 01234✓6%   initial log fill
-///   1 ───────────✓·%             │ 890✓2%45   first wrap of log
```
matklad, line 1350: It was a bit confusing how, on the right, we show full wrap of the log (up to op=15), while on the left we are up to 13. In the new version, I made left and right to show the same data

src/vsr.zig
```diff
@@ -1369,12 +1372,13 @@ pub const Checkpoint = struct {
...
         const result = op: {
             if (checkpoint == 0) {
-                // First wrap: op_checkpoint_next = 8-2-1 = 5
-                break :op constants.journal_slot_count - constants.lsm_batch_multiple - 1;
+                // First wrap: op_checkpoint_next = 6-1 = 5
+                // -1: vsr_checkpoint_interval is a count, result is an inclusive index.
+                break :op constants.vsr_checkpoint_interval - 1;
             } else {
-                // Second wrap: op_checkpoint_next = 5+8-2 = 11
-                // Third wrap: op_checkpoint_next = 11+8-2 = 17
```
matklad, line 1376: I figured it would be less confusing to not decompose checkpoint interval. After all, it is morally a parameter, and can be set arbitrary, as long as it allows at least a trigger plus pipeline of slack.

src/vsr/replica.zig
matklad: Meant to revisit before merging. That's just wrong, it wants to check for checkpoint trigger, not for journal lenght.

src/vsr.zig
```diff
@@ -1337,27 +1338,29 @@ const ViewChangeHeadersArray = struct {
...
-///     0   4   8   2   6   0   4  │ 0---4---
-///   0 ─────✓·%                   │ 01234✓6%   initial log fill
-///   1 ───────────✓·%             │ 890✓2%45   first wrap of log
-///   2 ─────────────────✓·%       │ 6✓8%0123   second wrap of log
-///   3 ───────────────────────✓·% │ 4%67890✓   third wrap of log
+///     0   4   8   2   6   0   4  │ 0 - - - 4 - - - 8
+///   0 ─────✓·%                   │[0 1 2 3 4 ✓]6 % R
+///   1 ───────────✓·%             │ 9 0 ✓]2 % 5[6 7 8
+///   2 ─────────────────✓·%       │ 8 % 1[2 3 4 5 6 ✓]
+///   3 ───────────────────────✓·% │[8 9 0 1 2 ✓]4 % 7
```
sentientwaffle: Numbering + brackets are off, I think?:
```suggestion
///     0   4   8   2   6   0   4  │ 0 - - - 4 - - - 8
///   0 ─────✓·%                   │ 0 1 2 3 4 ✓]6 % R
///   1 ───────────✓·%             │ 9 0 ✓]2 % 4 5 6 ✓]
///   2 ─────────────────✓·%       │ 8 % 0 1 2 ✓]4 % 6
///   3 ───────────────────────✓·% │ 7 8 ✓]0 % 2 3 4 ✓]
```

matklad: I _think_ it is correct, but there's high chance we might be reading the chart differently, as I find the original version confusing.

Eg, in the second row/checkpoint, when I write `9 0 ✓]2 % 5[6 7 8`, that `5` is correct --- it is a 5 (not 15) from the _first_ checkpoint which we haven't yet overwritten. And the checkpoint is `[6 7 8 9 10 11]`.


I _think_ the original version of the  chart on the right showed _full_ log wraps, with all entries overwritten, but that I think is confusing --- we have full wraps on the right, but full _checkpoints_ on the left. 

Now that I think about it, theer's no reason to compress `10` to just `0` on the right --- there's space to write full ops, ptal!

sentientwaffle: Ahh I understand now, thank you! Yes, we were reading the chart differently, but your version is much more useful than mine :tada:

sentientwaffle (approved):

sentientwaffle (approved):

PR #1347, opened 2023-12-11
Client: Java ID generator helper 

kprotty:
Creates a static helper function in the java client for generating thread-safe, unique, and monotonic `u128` IDs to use for Accounts and Transfers.

### DIFF
--- a/src/clients/java/src/main/java/com/tigerbeetle/UInt128.java
+++ b/src/clients/java/src/main/java/com/tigerbeetle/UInt128.java
@@ -175,10 +175,8 @@ public enum UInt128 {
     }
 
     private static long ulidLastTimestamp = 0L;
-    private static final byte[] ulidLastEntropy = new byte[80];
+    private static final byte[] ulidLastRandom = new byte[80];
     private static final SecureRandom ulidSecureRandom = new SecureRandom();
-    private static long ulidMonotonicCounterLo = 0;
-    private static int ulidMonotonicCounterHi = 0;
 
     /**
      * Generates a Universally Unique Lexicographically Sortable Identifier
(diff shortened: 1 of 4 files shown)


src/clients/java/src/main/java/com/tigerbeetle/UInt128.java
```diff
@@ -172,5 +174,53 @@ public static byte[] asBytes(final BigInteger value) {
...
+                ulidSecureRandom.nextBytes(ulidLastRandom);
+            }
+
+            var random = ByteBuffer.wrap(ulidLastRandom).order(ByteOrder.LITTLE_ENDIAN);
+            randomLo = random.getLong();
+            randomHi = random.getShort();
+
+            // Increment lo. If that overflows, increment hi. If that overflows, throw error.
+            // ULID spec says to throw error upon 80-bit overflow (not relative to initial random).
+            if (randomLo++ == Long.MAX_VALUE && randomHi++ == 0xffff) {
```
batiati: Zig has spoiled me a lot 😇

- Can we drop the `++` operator in favor of `+=1`?

- It's a signed integer, did you really mean `Long.MAX_VALUE` here, or should it be `0xFFFFFFFFFFFFFFFF` ?

- It's safe to use `+ 1` with signed integers since Java allows arithmetic operations to overflow. But it's worthy of a comment.

src/clients/java/src/test/java/com/tigerbeetle/UInt128Test.java
```diff
@@ -165,4 +169,89 @@ public void testAsBigIntegerZero() {
...
+
+                    // Same as serial test above, but with smaller bounds.
+                    var ulidA = new DecodedULID(UInt128.ULID());
+                    for (int j = 0; j < 10_000; j++) {
+                        if (j % 1000 == 0) {
+                            Thread.sleep(1);
+                        }
+
+                        var ulidB = new DecodedULID(UInt128.ULID());
+                        ulidB.assertMonotonicSince(ulidA);
```
batiati: Can we also use `UInt128.asBigInteger` to compare both ULIDs?
This will assert that the binary representation is really what we expect at Zig's side (a monotonically increasing `u128` little-endian).

```java
var bigIntA = UInt128.asBigInteger(...);
var bigIntB = UInt128.asBigInteger(...);
assertTrue(bigIntB.compareTo(bigIntA) > 0); 
```

batiati (commented):
Nice and elegant implementation!!!

Can we push the scope of this PR a little further and review the samples and docs to mention/use ULID where it's applicable?

src/clients/java/src/main/java/com/tigerbeetle/UInt128.java
```diff
@@ -172,5 +174,53 @@ public static byte[] asBytes(final BigInteger value) {
...
 
+    /**
+     * Generates a Universally Unique Lexicographically Sortable Identifier
+     * (<a href="https://github.com/ulid/spec">ULID</a>) as 16 bytes of a 128-bit value.
+     *
+     * The ULID returned always increases monotonically and is thread-safe.
+     *
+     * @return an array of 16 bytes representing the 128-bit value.
+     */
+    public static byte[] ULID() throws ArithmeticException {
```
batiati: By convention unchecked exceptions should not be declared in the `throws` clause, but only documented.

```suggestion
    /**
     * Generates a Universally Unique Lexicographically Sortable Identifier
     * (<a href="https://github.com/ulid/spec">ULID</a>) as 16 bytes of a 128-bit value.
     *
     * The ULID returned always increases monotonically and is thread-safe.
     *
     * @throws ArithmeticException as per ULID specs, if the 80 bits of randomness overflows.
     * @return an array of 16 bytes representing the 128-bit value.
     */
    public static byte[] ULID() {
```

src/clients/java/src/test/java/com/tigerbeetle/UInt128Test.java
kprotty: Sort of: `asBigInteger` interprets the bytes as 2 little-endian u64 values of a u128. ULID however stores timestamp as `{u32, u16}` and random as u80, both big-endian. We can instead interpret the bytes directly via `new BigInteger(ULID())` since it assumes big-endian. This also means Zig side might not see it as monotonically increasing little-endian.

kprotty: `ULID` has been renamed to `ID` and it stores the individual fields in little-endian which allows use of `asBigInteger`.

src/clients/java/src/main/java/com/tigerbeetle/UInt128.java
```diff
@@ -172,5 +174,60 @@ public static byte[] asBytes(final BigInteger value) {
...
+            }
+
+            var random = ByteBuffer.wrap(idLastRandom).order(ByteOrder.nativeOrder());
+            randomLo = random.getLong();
+            randomHi = random.getShort();
+
+            // If randomLo will overflow from increment, then increment randomHi as carry.
+            // If randomHi will overflow on increment, throw error on 80-bit random overflow.
+            if (randomLo == 0xFFFFFFFFFFFFFFFFL) {
+                if (randomHi == 0xffff) {
```
batiati: Let's standardize hex in upper case.

```suggestion
                if (randomHi == 0xFFFF) {
```

src/clients/java/src/main/java/com/tigerbeetle/UInt128.java
```diff
@@ -172,5 +174,60 @@ public static byte[] asBytes(final BigInteger value) {
...
     }
 
+    private static long idLastTimestamp = 0L;
+    private static final byte[] idLastRandom = new byte[80];
+    private static final SecureRandom idSecureRandom = new SecureRandom();
 
+    /**
+     * Generates a Universally Unique Sortable Identifier as 16 bytes of a 128-bit value.
+     *
+     * The ID() function is thread-safe, the bytes returned are formatted in little endian, and the
```
batiati: I think it's worth mentioning that this ID is based on the UILD spec, except it is little-endian, so it is binary sortable and not lexicographically sortable by the string representation. In fact, our version has no concept of string representation.

```suggestion
     * The ID() function is thread-safe, the bytes returned are stored in little endian, and the
```

batiati (changes requested):
Just two minor comments, but LGTM other than that!

batiati (dismissed):
Thanks King ✌️

batiati (approved):

PR #1351, opened 2023-12-11
VSR: Generalize trailer

sentientwaffle:
In preparation for moving client sessions data from the superblock trailer into the grid, generalize `FreeSetEncoded` into `CheckpointTrailer`.

Note that this commit doesn't touch the `schema.FreeSetNode` block type; that will be part of the next PR.


src/vsr/checkpoint_trailer.zig
sentientwaffle, line 1: This was a `git mv` from `free_set_encoded.zig`; not sure why it is showing it as a new file. :disappointed:

matklad: As I've recently learned, `git mv` is literally `mv && git add`, it doesn't actually record the fact that this was a move anywhere, move vs insert+delete is still based on diff heuristic. I think the only reliable way to record moves in git is to _commit_ just moves separately, and to fixup in a different commit on-top.

sentientwaffle: :facepalm: Ah, that explains it, thanks.

sentientwaffle: I'll split the commit up, but will hold off on pushing that until I implement other review comments, to avoid messing with any PR comments that you've already written.

matklad: not much written yet, please push as it is indeed a bit annoying to review without finer-grained diff :)

sentientwaffle: Done! https://github.com/tigerbeetle/tigerbeetle/pull/1351/commits/281360c5ff81df58114d431d11ed271e9e304a87

src/vsr/checkpoint_trailer.zig
```diff
@@ -140,269 +143,230 @@ pub fn FreeSetEncodedType(comptime Storage: type) type {
...
-        ///   1. Encode free set.
-        ///   2. Derive the number of blocks required to store the encoding.
-        ///   3. Allocate blocks for the encoding (in the old checkpoint).
-        ///   4. Write the blocks to disk.
-        ///   --------------------------------------------------------------
-        ///   5. Mark currently released blocks as free.
-        ///   6. Release the freshly acquired blocks in the new checkpoint.
-        ///
-        /// This function handles steps 1-4. The caller is responsible for calling
-        /// FreeSet.checkpoint which handles 5 and 6.
```
sentientwaffle: This description has moved to `grid.zig`.

src/vsr/checkpoint_trailer.zig
```diff
@@ -70,18 +74,18 @@ pub fn FreeSetEncodedType(comptime Storage: type) type {
         };
 
         // Reference to the grid is late-initialized in the open, because the free set is part of
-        // the superblock, which doesn't have access to grid. It is set to null by reset, to verify
-        // that the free set is not used before it is opened during sync.
+        // the grid, which doesn't have access to a stable grid pointer. It is set to null by
+        // `reset`, to verify that the free set is not used before it is opened during sync.
```
matklad, line 78: some residual free sets in the comment here

src/vsr/checkpoint_trailer.zig
```diff
@@ -140,269 +143,230 @@ pub fn FreeSetEncodedType(comptime Storage: type) type {
...
         }
 
-        fn open_read_next(set: *Self, address: u64, checksum: u128) void {
-            assert(set.callback == .open);
-            assert(set.size > 0);
+        fn open_read_next(trailer: *Self, address: u64, checksum: u128) void {
+            assert(trailer.callback == .open);
+            assert(trailer.size > 0);
+            assert((trailer.size_transferred == 0) ==
+                (trailer.block_index == trailer.block_count()));
```
matklad, line 236: :+1: for grouping related assert together. Do we have a rule of thumb which asserts go first, those that check the state of `self`, or those that check validity of parameters?

src/vsr/checkpoint_trailer.zig
```diff
@@ -140,269 +143,230 @@ pub fn FreeSetEncodedType(comptime Storage: type) type {
...
-                set.checksum = vsr.checksum(set.buffer[0..set.size]);
-            }
+            trailer.size_transferred = 0;
+            trailer.checksum = vsr.checksum(trailer.buffer[0..trailer.size]);
 
             {
-                assert(set.grid.?.free_set.count_reservations() == 0);
-                const reservation = set.grid.?.free_set.reserve(set.block_count()).?;
-                defer set.grid.?.free_set.forfeit(reservation);
+                assert(trailer.grid.?.free_set.count_reservations() == 0);
```
matklad, line 310: I wonder whether this is necessary? I think it would be correct to checkpoint a trailer concurrently with something else? Though, it's fine to leave assert as is, just to solidify the current implementation. 

However, let's add the same assert to:

* that place in `grid.zig` where we do "include_staging()" trick. _There_ it is crucialt that there are no outstanding reservations
* inside `include_staging` itself

src/vsr/grid.zig
```diff
@@ -266,45 +270,91 @@ pub fn GridType(comptime Storage: type) type {
...
+
+            grid.callback = .{ .open = callback };
+            grid.free_set_checkpoint.open(
+                grid,
+                grid.superblock.working.free_set_reference(),
+                free_set_open_callback,
+            );
+        }
+
+        fn free_set_open_callback(free_set_checkpoint: *CheckpointTrailer) void {
```
matklad: maybe `open_free_set_callback`, so that is lexicographically nested with `open`? I think we have both patterns, and I somewhat consciously chose `open_xxx_callback` naming for free set

src/vsr/grid.zig
```diff
@@ -266,45 +270,91 @@ pub fn GridType(comptime Storage: type) type {
...
+
+            {
+                assert(!grid.free_set.opened);
+                defer assert(grid.free_set.opened);
+
+                grid.free_set.open(.{
+                    .encoded = free_set_checkpoint.buffer[0..free_set_checkpoint.size],
+                    .block_addresses = free_set_checkpoint.block_addresses[0..free_set_checkpoint.block_count()],
+                });
+                assert((free_set_checkpoint.size > 0) == (grid.free_set.count_acquired() > 0));
```
matklad: I wonder if we should add `free_set.pristine()`? I worry a little that `count_acquired` is O(N), while its enough to just find a single set bit (which probably going to be the first one even)

src/vsr/grid.zig
matklad: And we should totally be consitent with `checkpoint_free_set_callback` in the same file

src/vsr/grid.zig
```diff
@@ -266,45 +270,91 @@ pub fn GridType(comptime Storage: type) type {
...
+        ///      checkpoint.
+        ///   7. Mark the free set's own blocks as released (but not yet free).
+        ///
+        /// This function handles step 1 and 5.
+        /// This function calls `free_set_checkpoint.checkpoint`, which handles steps 2-4.
+        /// The caller is responsible for calling FreeSet.checkpoint which handles 6 and 7.
         pub fn checkpoint(grid: *Grid, callback: *const fn (*Grid) void) void {
-            assert(grid.checkpointing == null);
-            assert(grid.canceling == null);
+            assert(grid.callback == .none);
```
matklad, line 327: I was thinking about a pattern like

```
assert(grid.callback == .none);
defere assert(grid.callback == .checkpoint);
```

to underscore that this is a state transition function. But that's probably more noise than helpful.

src/vsr/grid.zig
```diff
@@ -330,20 +380,20 @@ pub fn GridType(comptime Storage: type) type {
...
+            const callback = grid.callback.checkpoint;
+            grid.callback = .none;
             callback(grid);
         }
 
         pub fn cancel(grid: *Grid, callback: *const fn (*Grid) void) void {
-            assert(grid.canceling == null);
-            assert(grid.checkpointing == null);
+            // grid.open() is cancellable the same way that read_block()/write_block() are.
+            assert(grid.callback == .none or grid.callback == .open);
```
matklad: maybe switch for exhaustiveness here? feels useful to decect missing _new_ cancelation cases at compiletime.

src/vsr/superblock.zig
```diff
@@ -517,24 +515,21 @@ pub const ManifestReferences = struct {
...
-pub const FreeSetReference = struct {
-    /// Checksum over the entire encoded free set.
+pub const TrailerReference = struct {
+    /// Checksum over the entire encoded trailer.
     checksum: u128,
     last_block_address: u64,
     last_block_checksum: u128,
-    free_set_size: u32,
-    storage_size: u64,
+    trailer_size: u32,
```
matklad: For generic trailer, let's future-proof this to u64?

src/vsr/checkpoint_trailer.zig
sentientwaffle: I think in this particular case the comment is correct -- the reason for lazy-initializing the grid pointer is specifically due to the free-set/grid relationship, not client sessions.

src/vsr/checkpoint_trailer.zig
sentientwaffle: My personal rule of thumb is to put `self` assertions first, since `self` is the first parameter.

matklad: :facepalm: that's genius!

src/vsr/grid.zig
sentientwaffle: Maybe `.empty()`, since we have `open_empty()`?
(And to check that I'm on the same page: `pristine`/`empty` would check `count_acquired() == 0`, but is implemented with an early-exit when the first acquired address is found.)

src/vsr/grid.zig
sentientwaffle: I do like that pattern when there are multiple branches / return points, though!

src/vsr/superblock.zig
sentientwaffle: Should `free_set_size`/`client_sessions_size` in the `CheckpointState` be `u64` too?

matklad: I’d say yes!

src/vsr/grid.zig
matklad: I was a bit hesitant about empty, as I think we might want to check all of blocks, staging, and reservations. But I am probably overthinking this. And yes, the idea is to do an early return. Though I guess if we are checking staging, we’d still be O(N)?

sentientwaffle: :thinking: I think I'd rather assert each of those separately rather than bundling them up into a single function, to make it obvious what is being checked.
That information is already exposed via `free_set.count_*`.

src/vsr/grid.zig
```diff
@@ -391,7 +394,12 @@ pub fn GridType(comptime Storage: type) type {
 
         pub fn cancel(grid: *Grid, callback: *const fn (*Grid) void) void {
             // grid.open() is cancellable the same way that read_block()/write_block() are.
-            assert(grid.callback == .none or grid.callback == .open);
+            switch (grid.callback) {
+                .none => {},
+                .open => {},
+                .checkpoint => unreachable,
+                .cancel => unreachable,
+            }
```
matklad, line 402: Any specific reason to avoid 

```zig
.none, .open => {}
.checkpoint, .cancel => unreachable,
```

grouping here?

matklad: :thinking: I think we have one extra invariant here that we aren't actually checking perhaps? For states which are not cancelable (so, checkpoint and cancel), we must also ensure that we don't _read_ any blocks, as that could deadlock. 

In particular, we shouldn't be reading blocks during checkpoint. So, `assert(grid.callback != .cancel);` in `read_block` should also check `callback != chcekpoint`, and, really, it should be the same as current assert. 

so maybe add `gird.callback.can_cancel()` helper for assertions here and on the read path?

sentientwaffle: In `grid.read_block` we already have:

```
            switch (callback) {
                .from_local_storage => {
                    maybe(grid.callback == .checkpoint);
                    ...
                },
                .from_local_or_global_storage => {
                    assert(grid.callback != .checkpoint);
                    ...
                },
            }
```

Which I think is correct -- reads to satisfy grid repair requests are safe.

sentientwaffle: > Any specific reason to avoid

Just personal preference -- I like all of the variants aligned, so if I was grouping them I would do

```
.none,
.open,
=> {},
.checkpoint,
.cancel,
=> unreachable,
```

But in this case that is longer than just repeating the rhs.

matklad (approved):

PR #1352, opened 2023-12-12
lsm: resurrect secondary indexes gravestone optimization

matklad:
For secondary indexes, we don't need to compact tombstones all the way
to the last layer, and can immediately cancel out an insert and a
tombstone. What's more, both `put then remove` and `remove than put` are
cancelable, the first case just resurrects a previously inserted value.

### DIFF
--- a/src/lsm/compaction.zig
+++ b/src/lsm/compaction.zig
@@ -28,7 +28,7 @@
 //! † When A's value is a tombstone, there is a special case for garbage collection. When either:
 //! * level B is the final level, or
 //! * A's key does not exist in B or any deeper level,
-//! then the tombstone is omitted from the compacted output (see: `compaction_must_drop_tombstones`).
+//! then the tombstone is omitted from the compacted output, see: `compaction_must_drop_tombstones`.
 //!
 const std = @import("std");
 const mem = std.mem;
@@ -48,7 +48,8 @@ const TableInfoType = @import("manifest.zig").TreeTableInfoType;
 const ManifestType = @import("manifest.zig").ManifestType;
 const schema = @import("schema.zig");
 const TableDataIteratorType = @import("table_data_iterator.zig").TableDataIteratorType;
-const LevelTableValueBlockIteratorType = @import("level_data_iterator.zig").LevelTableValueBlockIteratorType;
+const LevelTableValueBlockIteratorType =
+    @import("level_data_iterator.zig").LevelTableValueBlockIteratorType;
 
 pub fn CompactionType(
     comptime Table: type,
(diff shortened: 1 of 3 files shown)


src/lsm/compaction.zig
```diff
@@ -482,8 +484,10 @@ pub fn CompactionType(
         // either using a HashMap or having to do extra sorting. They are eventually filtered out
         // by our compaction merge() however.
```
cb22: Maybe worth updating the TODO comment here?

src/lsm/compaction.zig
```diff
@@ -522,13 +538,15 @@ pub fn CompactionType(
...
             const source_count = source_index;
             const target_count = target_index;
-
             assert(target_count <= source_count);
-            assert(target_count > 0);
-
             compaction.context.table_info_a.immutable =
                 compaction.context.table_info_a.immutable[source_count..];
 
+            if (target_count == 0) {
```
cb22, line 541: I know we mention we can't have empty tables in schema.zig, just checking that this won't generate them?

src/lsm/table.zig
```diff
@@ -21,8 +21,13 @@ pub const TableUsage = enum {
     /// If your usage fits this pattern:
     /// * Only put keys which are not present.
     /// * Only remove keys which are present.
-    /// * TableKey == TableValue (modulo padding, eg CompositeKey)
-    /// Then we can unlock additional optimizations.
+    /// * TableKey == TableValue (modulo padding, eg CompositeKey).
+    /// Then we can unlock additional optimizations:
+    /// * Immediately cancel out tombstone  and the corresponding insert, without waiting for the
```
cb22:
```suggestion
    /// * Immediately cancel out the tombstone and the corresponding insert, without waiting for the
```

src/lsm/table.zig
```diff
@@ -21,8 +21,13 @@ pub const TableUsage = enum {
...
+    ///   tombstone to sink to the bottom of the LSM true: absence of upserts guarantees that
```
cb22: I don't quite follow the "absence of upserts guarantees" bit? Do you mean because we explicitly know if we're inserting vs updating we can guarantee that if we have a pair that cancels, it couldn't possibly have existed on a lower level?

src/tidy.zig
```diff
@@ -191,7 +191,6 @@ const naughty_list = [_][]const u8{
     "io/test.zig",
     "io/windows.zig",
     "lsm/binary_search.zig",
-    "lsm/compaction.zig",
```
cb22, line 194: :tada:

src/lsm/compaction.zig
matklad: It wont, for two reasons:

* this hits the same code path as `compaction.context.table_info_a.immutable.len == 0` case, with the `assert(compaction.values_in[0].len == 0)`.
* that case works correctly, it'll hit `if (values_in[0].len == 0 and values_in[1].len == 0)` condition if level b is also empty

src/lsm/table.zig
matklad: Good point, I think this should say just "updates" --- what we don't have here is to consequitive inserts, two inserts are always separated by a tombstone.

src/lsm/table.zig
matklad: I hate articles, we do just fine without them in Russian! I think it should be _a_ tombstone and _the_ corresponding insert though? Also sprinkled some more articles for the second bullet point.

cb22 (approved):
LGTM! Going to kick off a scale test with this shortly.

PR #1353, opened 2023-12-13
Allow git commit hash to be passed at build time.

DanielSidhion:
When building with nix, usually we don't fetch the git repository if it can be avoided (and instead use the source archive generated by GitHub). This PR allows the git commit revision to be passed at build time rather than be determined by running `git`, which makes the build play nicer with nix.

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -61,7 +61,7 @@ pub fn build(b: *std.Build) !void {
     defer shell.destroy();
 
     // The "tigerbeetle version" command includes the build-time commit hash.
-    options.addOption([]const u8, "git_commit", try shell.git_commit());
+    options.addOption([]const u8, "git_commit", b.option([]const u8, "git-commit", "The git commit revision of the source code.") orelse try shell.git_commit());
     options.addOption(
         []const u8,
         "version",


DanielSidhion:
Looking at how the commit hash is used, it seems it's only used when running `tigerbeetle version --verbose`. Is the commit hash really required in this case? Usually it can be inferred through the version itself, so IMO it doesn't make much sense to include it.

As an alternative way of making tigerbeetle build nicer with nix, I'd propose to just drop the git commit hash thing entirely.

matklad (approved):

matklad:
Thanks, this totally makes sense! I've actually already added the "pass hash externally" infrastructure to our release process for much the same reason, but looks like I've forgot to actually thread it all the way to the actual build process!

https://github.com/tigerbeetle/tigerbeetle/blob/main/.github/workflows/release.yml#L45

Thanks a bunch!

matklad:
Forgot to actually write down why we don't want to drop the hash entirely --- having a hash in the binary is too damn convenient for debugging! It _can_ be inferred from the version, but that's an easy to mess up step.

PR #1354, opened 2023-12-13
release: don't rely on git to get the commit hash

matklad:
The plan with `zig build release` all along was to just pass the hash in from the outside. Turns out, while the hash is passed, I actually forgot to thread it all the way through to build.zig!

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -61,7 +61,13 @@ pub fn build(b: *std.Build) !void {
     defer shell.destroy();
 
     // The "tigerbeetle version" command includes the build-time commit hash.
-    options.addOption([]const u8, "git_commit", b.option([]const u8, "git-commit", "The git commit revision of the source code.") orelse try shell.git_commit());
+    const git_commit = b.option(
+        []const u8,
+        "git-commit",
+        "The git commit revision of the source code.",
+    ) orelse try shell.git_commit();
+    options.addOption([]const u8, "git_commit", git_commit);
+
     options.addOption(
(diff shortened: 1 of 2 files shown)


matklad:
cc @DanielSidhion, this `release.zig` is our "official" release process which builds all the binaries the way they are intended to be used in production. In general, our binary releases are the "public interface": while we don't do this yet, we will also add some last minute testing of the actual binary artifacts, so that we have better assuarance and clarity on what _exactly_ we are shipping to users. For nix, I think you want to build from source, but you could look at the real release process if you want to match that closer. 

At some point, we probably should make a guarantee that `./zig/zig build install -Drelease` is how you get a production binary for your platform, but at the moment the build interface is a bit of implementation detail --- free free to use it, just don't expect it to never change!

DanielSidhion:
@matklad thank you, I read through what `release.zig` is doing, and it appears that for linux it just boils down to calling `zig build install`, so I'll stick to that for nix builds (I'm focused on packaging tigerbeetle for linux initially - macos is a different beast).

Nix has a specific target already specified when building, so it doesn't make sense to go through the entire build process (which builds many types of binaries for multiple targets all at once) to get just the single binary we're interested in. I'll keep an eye on changes to the build interface, I'm fine with having to keep up with any changes there. I'm assuming the interface won't change too much, so I might change my opinion if/when this assumption breaks :smile:

cb22 (approved):

PR #1356, opened 2023-12-13
VSR: Move client sessions from superblock to grid

sentientwaffle:
(Follow-up for https://github.com/tigerbeetle/tigerbeetle/pull/1351).

No more trailers in the superblock!

Client sessions is encoded with `CheckpointTrailer`, like the free set. Unlike the free set, client sessions encoding/decoding is handled directly by `replica.zig`, not `grid.zig`.

This simplifies the superblock checkpoint/sync control flow, which now just needs to care about the superblock headers. This simplifies state sync, since the superblock sync is now just a single `sync_checkpoint` message.


src/vsr/superblock.zig
```diff
@@ -545,30 +545,13 @@ comptime {
 /// The size of the entire superblock storage zone.
 pub const superblock_zone_size = superblock_copy_size * constants.superblock_copies;
 
-/// The size of an individual superblock including trailer.
-pub const superblock_copy_size = @sizeOf(SuperBlockHeader) + superblock_trailer_size_max;
+/// The size of an individual superblock including padding.
+/// TODO Add some padding between copies and include that here.
+pub const superblock_copy_size = @sizeOf(SuperBlockHeader);
```
sentientwaffle, line 550: A few reasons to include padding in this number:
- we probably want to reserve a bit of extra space as a just-in-case, since reformatting a whole data file isn't really an option
- can add padding here without requiring an additional zone (`grid_padding`)
- separate physical block addresses (in SSD terms, not grid blocks) of the superblock headers (probably doesn't matter with SSDs :shrug:)

src/testing/cluster/storage_checker.zig
```diff
@@ -3,14 +3,8 @@
 //! At each replica compact and checkpoint, check that storage is byte-for-byte identical across
 //! replicas.
 //!
-//! Areas verified at compaction (half-measure):
-//! - Acquired Grid blocks (ignores skipped recovery compactions)
-//!  TODO Because ManifestLog acquires blocks potentially several beats prior to actually writing
-//!  the block, this check will need to be removed or use a different strategy.
```
sentientwaffle, line 9: Outdated comment. Originally the storage checker was going to verify storage determinism at half-bar boundaries, but that turns out not to work (due to the manifest log). Checkpoint is more comprehensive anyway.

src/vsr/replica_test.zig
```diff
@@ -872,17 +872,17 @@ test "Cluster: sync: sync, bump target, sync" {
     t.replica(.R2).drop_all(.R_, .bidirectional);
     try c.request(checkpoint_2_trigger, checkpoint_2_trigger);
 
-    // Allow R2 to complete SyncStage.requesting_trailers, but get stuck
-    // during SyncStage.requesting_trailers.
+    // Allow R2 to complete SyncStage.requesting_target, but get stuck
```
sentientwaffle, line 875: This was a typo, I think.

src/vsr/replica.zig
```diff
@@ -7918,36 +7911,29 @@ pub fn ReplicaType(
...
 
-            inline for (comptime std.enums.values(vsr.SuperBlockTrailer)) |trailer| {
-                const trailer_size = stage.trailers.get(trailer).final.?.size;
-                @field(self.superblock, @tagName(trailer)).reset();
-                @field(self.superblock, @tagName(trailer)).decode(
-                    self.superblock.trailer_buffer(trailer)[0..trailer_size],
-                );
-            }
-            // Faulty bits will be set in state_machine_open_callback().
+            // Faulty bits will be set in sync_content().
```
sentientwaffle, line 7936: This comment was out of date. I think `state_machine_open_callback` did originally handle this logic, but it was moved to `sync_content` as soon as client-replies + table sync was added.

src/testing/cluster/storage_checker.zig
```diff
@@ -73,23 +69,34 @@ pub const StorageChecker = struct {
...
+        var client_sessions = try vsr.ClientSessions.init(allocator);
+        errdefer client_sessions.deinit(allocator);
+
         var free_set_buffer = try allocator.alignedAlloc(
             u8,
             @alignOf(u64),
             vsr.FreeSet.encode_size_max(Storage.grid_blocks_max),
         );
-        errdefer allocator.free(free_set);
+        errdefer allocator.free(free_set_buffer);
```
sentientwaffle, line 80: Typo. Not found by the compiler (prior to adding `client_sessions_buffer`) since the `errdefer` was not previously reachable (nothing could possibly error before here and the end of the function).

src/lsm/schema.zig
```diff
@@ -343,13 +344,12 @@ pub const TableData = struct {
...
-    pub const Word = u64;
-
+/// A TrailerNode is either a `BlockType.free_set` or `BlockType.client_sessions`.
+pub const TrailerNode = struct {
     pub const Metadata = extern struct {
-        previous_free_set_block_checksum: u128,
-        previous_free_set_block_checksum_padding: u128 = 0,
-        previous_free_set_block_address: u64,
+        previous_trailer_block_checksum: u128,
+        previous_trailer_block_checksum_padding: u128 = 0,
```
matklad, line 351: Wait, do we need this padding? I think we switched all pure checksumms to be unpadded?

src/testing/cluster/storage_checker.zig
matklad, line 80: Oups! Good catch! I've been thinking recently that maybe we should flat out `@panic` on OOM
errors, so that we could get rid of `errdefer`s like these...

src/testing/cluster/storage_checker.zig
```diff
@@ -195,26 +194,44 @@
...
+            }
+            assert(client_sessions_cursor == 0);
+        }
+        assert(vsr.checksum(checker.client_sessions_buffer[0..client_sessions_size]) ==
+            superblock.working.vsr_state.checkpoint.client_sessions_checksum);
+
+        checker.client_sessions.decode(checker.client_sessions_buffer[0..client_sessions_size]);
+        defer checker.client_sessions.reset();
 
         var checksum: u128 = 0;
```
matklad: Maybe use steraming checksum instead of `^` here to add dependency on the _order_ of replies and
not just their content?

src/vsr/checkpoint_trailer.zig
```diff
@@ -116,12 +116,7 @@ pub fn CheckpointTrailerType(comptime Storage: type) type {
             checkpoint: *const fn (trailer: *Self) void,
         } = .none,
 
-        comptime {
-            assert(FreeSet.Word == schema.FreeSetNode.Word);
-            assert(chunk_size_max % @sizeOf(FreeSet.Word) == 0);
-        }
-
-        pub fn init(allocator: mem.Allocator, buffer_size: usize) !Self {
+        pub fn init(allocator: mem.Allocator, buffer_size: usize, trailer_type: TrailerType) !Self {
```
matklad: Femto-nit, but mabye put type before the buffer? As it is the most important argument that
determines the rest.

src/vsr/client_replies.zig
```diff
@@ -308,7 +308,7 @@ pub fn ClientRepliesType(comptime Storage: type) type {
             maybe(client_replies.writing.isSet(slot.index));
             assert(message.header.command == .reply);
             // There is never any need to write a body-less message, since the header is
-            // stored safely in the client sessions superblock trailer.
+            // stored safely in the superblock header's `client_sessions`.
```
matklad: This makes it sound like sessions are stored in the superblock header, rather than in a checkpoint
trailer.

src/vsr/replica.zig
```diff
@@ -689,6 +695,49 @@ pub fn ReplicaType(
...
+            }
+
+            const trailer_buffer = self.client_sessions_checkpoint.buffer;
+            const trailer_size = self.client_sessions_checkpoint.size;
+
+            self.client_sessions.reset();
+            if (self.superblock.working.client_sessions_reference().empty()) {
+                assert(trailer_size == 0);
+            } else {
+                assert(trailer_size == ClientSessions.encode_size_max);
```
matklad: Maybe drop `_max` from `encode_size_max`? Otherwise, this line begs the question of whether it
was meant to be `<=` rather than `==`?

src/vsr/replica.zig
```diff
@@ -3177,6 +3158,11 @@
                 .checkpoint_client_replies => {
                     self.client_replies.checkpoint(commit_op_checkpoint_client_replies_callback);
                 },
+                .checkpoint_client_sessions => {
+                    self.client_sessions_checkpoint.size =
+                        self.client_sessions.encode(self.client_sessions_checkpoint.buffer);
```
matklad, line 3163: I am wondering if perhaps there's some way to assert that _both_ size and buffer are correctly set
at the same time? But I don't see any.

src/vsr/superblock.zig
```diff
@@ -177,6 +167,10 @@ pub const SuperBlockHeader = extern struct {
                     .free_set_last_block_checksum = 0,
                     .free_set_last_block_address = 0,
                     .free_set_size = 0,
+                    .client_sessions_checksum = vsr.checksum(&.{}),
```
matklad: While we are here, might do `comptime vsr.checksum(&.{})`. Not that it _really_ matters when we
compute the empty hash here, but computing it at runtime generally irks me a bit, as that's not a
trivial operation.

src/vsr/replica.zig
sentientwaffle: Yeah, the API is a little awkward. :thinking:

src/lsm/schema.zig
sentientwaffle: I don't think this actually is a pure checksum -- it is referring to a block, which would be encrypted, so the "checksum" would be the AEAD tag.

src/vsr/replica.zig
```diff
@@ -7977,6 +7963,7 @@ pub fn ReplicaType(
 
             self.grid.free_set.reset();
             self.grid.free_set_checkpoint.reset();
+            self.client_sessions_checkpoint.reset();
```
matklad, line 7966: Feels odd that we are not reseting client_sessions here, but rather do it in `client_sessions_open_callback`. 

I think resetting right before decoding makes sense to me from the perspective of defensive programming the decode code doesn't have to worry whether the object was properly reset. But maybe it makes sense to go on an offensive here? Reset the sessions here, so that, between this moment, and the moment we actually decode them, the sessions are in a well-defined reset state? In decoding, we can then _assert_ that the sessions are reset.

matklad (approved):

PR #1357, opened 2023-12-14
Fix a JVM 21 panic during jni tests

batiati:
JNI unit test causes the JVM to panic on Java 21.

Issue was due to checking if the local reference created in a popped frame was in fact invalidated, akin to scoped variable after exiting the block. Modified the `test "JNI: LocalFrame"` to handle changed behavior in `env.get_object_ref_type`, which now panics instead of returning `.invalid`.

The JNI documentation says:
> GetObjectRefType cannot be used on deleted references. Since references are typically implemented as pointers to memory data structures that can potentially be reused by any of the reference allocation services in the VM, once deleted, it is not specified what value the GetObjectRefType will return.

- The issue affected only the unit test (which tests all JNI functions), we **don't use** this function in TB Java Client.
- General clean up of unnecessary `var` declarations on `jni_test.zig`.

### Question

Should we add Java 21|latest to our CI? Currently we are running on Java 11 only.

### DIFF
--- a/src/clients/java/src/jni_tests.zig
+++ b/src/clients/java/src/jni_tests.zig
@@ -8,7 +8,7 @@ const JavaVM = jni.JavaVM;
 const JNIEnv = jni.JNIEnv;
 
 test "JNI: check jvm" {
-    var env: *JNIEnv = get_testing_env();
+    const env: *JNIEnv = get_testing_env();
 
     var jvm: *JavaVM = undefined;
     const get_java_vm_result = env.get_java_vm(&jvm);
@@ -28,14 +28,14 @@ test "JNI: check jvm" {
 }
 
 test "JNI: GetVersion" {
-    var env: *JNIEnv = get_testing_env();
+    const env: *JNIEnv = get_testing_env();
 
     const version = env.get_version();
     try testing.expect(version >= jni.jni_version_10);
 }
 
 test "JNI: FindClass" {
-    var env: *JNIEnv = get_testing_env();
+    const env: *JNIEnv = get_testing_env();
 
     const object_class = env.find_class("java/lang/Object");
     try testing.expect(object_class != null);
@@ -49,7 +49,7 @@ test "JNI: FindClass" {
 }
 
 test "JNI: GetSuperclass" {
-    var env: *JNIEnv = get_testing_env();
+    const env: *JNIEnv = get_testing_env();
 
     const object_class = env.find_class("java/lang/Object");
     try testing.expect(object_class != null);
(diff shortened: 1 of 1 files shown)


src/clients/java/src/jni_tests.zig
```diff
@@ -281,28 +281,30 @@ test "JNI: LocalFrame" {
...
+    // except by the frame result.
+    const pop_local_frame_result = env.pop_local_frame(local_ref);
+    try testing.expect(pop_local_frame_result != null);
+    defer env.delete_local_ref(pop_local_frame_result);
 
-    var dead_reference = env.get_object_ref_type(obj);
-    try testing.expect(dead_reference == .invalid);
+    const valid_reference = env.get_object_ref_type(pop_local_frame_result);
+    try testing.expect(valid_reference == .local);
+    try testing.expect(pop_local_frame_result != local_ref);
```
batiati, line 295: The new assertion can't guarantee that the reference was dropped. But it's a JVM detail, our goal is just to assert that the function we called was in fact executed.

matklad (approved):

matklad:
>Should we add Java 21|latest to our CI? Currently we are running on Java 11 only.

That's a good question! We actually do test "latest" versions as a part of our release validation workflow: 

https://github.com/tigerbeetle/tigerbeetle/blob/21d0b774e21c4cdc90e8d59575bf1a601f4a99fa/.github/workflows/release_validate.yml#L35

But that rightfully doesn't run these unit tests. 

I _am_ a bit hesitant with adding an extra version to our not-rocket-science --- adding everything could quickly lead to quite slow CI. So, in the ideal world I think the setup should be:

- for main, we gate on tests passing on the oldest supported version
- we rely on platform backwards comparability guarantees to not gate on the latest version
- we verify those guarantees by chron-checking relases (so, when the next java version comes out that breaks our stuff, we learn about that from cron, and not when we try to merge a next PR)

Which is basically the setup we have today (modulo the fact that not all setup-$lang actions support 'latest' version)! 

But it clearly fails here, because:

- our code was incorrect, it relied on something that was not actually covered by backwards compatability guarantees
- and it was a bug in our test code, not in the code which is covered by release testing

:thinking: 

I guess on balance it makes sense to just add 21 to our CI, work on brining the CI time down otherwise (eg, re-using beetles for integration tests, or pushing `--in-memory` flag to complition), and re-visit pruning the matrix otherwise. 

I explicitly _don't_ want to add checks that run on `main`, but not on PRs: if we care about something enough to test on every commit, it should be guarded by rockets.

PR #1358, opened 2023-12-14
vsr: use correct naming convention for a runtime limit

matklad:
`_max` is a hard limit --- something which can't be exceeded due to the physics of the database.

`_limit` is a soft limit --- something which we don't exceeded because we chose so.

storage_size_max used to be a hard limit, as it was hard-coded into the superblock. But today it is just a `limit`, affecting only the RAM consumption.


jorangreef:
Love the distinction between the two.

What about `limit` (soft) and `limit_max` (hard)?

src/config.zig
```diff
@@ -77,7 +77,7 @@ const ConfigProcess = struct {
     verify: bool,
     port: u16 = 3001,
     address: []const u8 = "127.0.0.1",
-    storage_size_max: u64 = 16 * 1024 * 1024 * 1024 * 1024,
+    storage_size_limit: u64 = 16 * 1024 * 1024 * 1024 * 1024,
```
sentientwaffle: Maybe this should be `storage_size_limit_default`, to disambiguate it from the "real" runtime storage limit (via CLI arg)?

matklad: I think it is `limit_max` rather than `limit_default`. Or rather, it is both default and max, but the max is more salient of the two

sentientwaffle: Agreed -- I didn't realize we used that value as the maximum! :sweat_smile:

sentientwaffle: Should this not also be named `storage_size_limit_max` then (like `constants` is)?

sentientwaffle (approved):

PR #1359, opened 2023-12-15
docs: establish release rotation

matklad:


docs/internals/releases.md
```diff
@@ -42,9 +42,34 @@ synchronization:
 
 All publishing keys are stored as GitHub Actions in the `release` environment.
 
-## Triggering a Release
+## Versioning
+
+Because releases are frequent, we avoid specifying the version in the source code. Instead, the
+version is setup automatically by the release process itself, using sequential build numbers as the
+patch version. It might a case that a particular version is skipped (e.g., if a release partially
```
cb22:
```suggestion
patch version. It might be the case that a particular version is skipped (e.g., if a release partially
```

docs/internals/releases.md
```diff
@@ -42,9 +42,34 @@ synchronization:
...
+failed). It is also a possibility that two releases are functionally identical (this means that no
+pull requests were merged in a week).
+
+## Release Process
 
-Releases are triggered manually, on Monday, by pushing a commit from `origin/main` to
+Releases are triggered manually, on Monday. Default release rotation is stateless and determined by
```
cb22, line 55: > stateless

Implicit time dependency :wink:

docs/internals/releases.md
```diff
@@ -42,9 +42,34 @@ synchronization:
...
+the following one-liner:
+
+```console
+$ python3 -c "import datetime; print((sorted(['batiati', 'cb22', 'kprotty', 'matklad', 'sentientwaffle']) * 142)[datetime.datetime.utcnow().isocalendar().week:][:3])"
+['batiati', 'cb22', 'kprotty']
+```
+The middle name is the default release manager for the _current_ week. After the release, please
```
cb22, line 67: I got a bit confused by this. Is it the middle name on the Monday that the release is meant to be done? Ie, it's @kprotty today?

matklad: Yes, this is correct! The idea behind three names is to avoid confusion when the script is run on the weekend --- you can sanity-check against who was the last release manager!

batiati (commented):
Two considerations to include alternative flows:

a) When "Monday" is a holiday, like the next one will be December 25th.

b)  If the next release manager is unavailable (local holiday, vacations, traveling, etc).
We could just ask for a volunteer instead of adding complexity to the one-liner.

matklad:
Both good points, added!

batiati (approved):
🤝

PR #1360, opened 2023-12-15
VSR: Allow checkpoint of empty trailer

sentientwaffle:
(The TODO was a note-to-self to check this that I didn't realize I committed.)

If the freeset shrinks down such that all blocks are freed (which admittedly isn't actually possible with our state machine) then it would encode to `0` bytes. `free_set.reserve()` asserts that the `block_count` to reserve is nonzero, so we must guard against that.


matklad (approved):
:+1: :facepalm: I meant to add a review comment on this TODO, but probably lost it in the GitHub's UI

matklad:
https://github.com/tigerbeetle/tigerbeetle/pull/1361 would allow using `// FIXME:` for this cases, as a separate mechanism distinct from `// TODO:`

PR #1364, opened 2023-12-19
ci: allow to manually trigger a release

matklad:
In case a release fails transiently, it might be useful to manually re-run it, before addressing the underlying issue.

Just clicking the re-run button doesn't work --- re-runs get same run_number, so they'll try to publish the _same_ version, which, rightfully, doesn't work.

To allow re-runnning the workflow on the same commit but with a different version, add a manual trigger.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,6 +1,7 @@
 name: Release
 
 on:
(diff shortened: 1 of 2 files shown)


matklad:
Actually, wait, let me add timeout in this PR as well, so that we can debug immediately.

cb22 (approved):

PR #1368, opened 2023-12-19
Bump concurrency_max

batiati:
To fully take advantage of [client batching](https://github.com/tigerbeetle/tigerbeetle/pull/1257), the `concurrency_max` parameter needs to be set high enough for concurrent threads/tasks to be able to fill a batch completely.

This PR:

- Bumps the hard limit from `4096` (arbitrary) to `8190` (calculated from the max number of elements in a batch).
- Bumps the default runtime limit from `32` to `256` (still arbitrary).

### DIFF
--- a/src/clients/go/docs.zig
+++ b/src/clients/go/docs.zig
@@ -163,7 +163,7 @@ pub const GoDocs = Docs{
 
     .client_object_documentation =
     \\The third argument to `NewClient` is a `uint` max concurrency
-    \\setting. `256` is a good default and can increase to `8190`
+    \\setting. `256` is a good default and can increase to `8192`
     \\as you need increased throughput.
     ,
 
--- a/src/clients/go/docs.zig
+++ b/src/clients/go/docs.zig
@@ -153,7 +153,7 @@ pub const GoDocs = Docs{
     \\if len(tbAddress) == 0 {
     \\  tbAddress = "3000"
     \\}
-    \\client, err := NewClient(ToUint128(0), []string{tbAddress}, 32)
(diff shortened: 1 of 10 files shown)


src/clients/go/docs.zig
```diff
@@ -163,7 +163,7 @@ pub const GoDocs = Docs{
 
     .client_object_documentation =
     \\The third argument to `NewClient` is a `uint` max concurrency
-    \\setting. `32` is a good default and can increase to `4096`
+    \\setting. `256` is a good default and can increase to `8190`
```
kprotty: why 8190 (over 8192)?

batiati: It's from here:

https://github.com/tigerbeetle/tigerbeetle/blob/9f11a15ca2610181a3c3f0b79ebd4315a69de9e6/src/clients/c/tb_client/context.zig#L121-L130

I hate exposing internal constants like this, but we already have this in other places until the client exposes it properly.

kprotty: Would it make sense for the Context-API change to remove the `concurrency_max` limit?

batiati: Yes, I think we should let the target runtime be completely in charge of that :+1: 
But for now, it's a good measure to just bump the old default `32` which is too low to take real advantage of batching.

batiati: @kprotty 
As a second thought, it should not be related to `batch_max`:

- The `concurrency_max` doesn't mean that all concurrent tasks are submitting the same event type!
- We can swap StateMachines (in fact we do it for the EchoTest).

Changed it to `8192` an arbitrary value, just double of the previous one 4096.

kprotty (approved):

PR #1373, opened 2023-12-20
LSM: Add value count to TableInfo

cb22:
Main use case is for paced compaction - with the value count available in TableInfo, we can more accurately spread the work since we'll now account for partially empty tables. (and lower our grid reservations per beat, too!)

The issue on the roadmap goes on to say:
> And possibly value-block count, since compression will decouple the ratio between the two

I'm not sure what our plans for compression are here - so I haven't done the value block just yet...

### DIFF
--- a/src/lsm/manifest.zig
+++ b/src/lsm/manifest.zig
@@ -45,6 +45,10 @@ pub fn TreeTableInfoType(comptime Table: type) type {
         key_min: Key, // Inclusive.
         key_max: Key, // Inclusive.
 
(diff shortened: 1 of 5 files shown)


sentientwaffle (approved):
LGTM. During paced compaction, can we verify that input tables do indeed contain exactly the `value_count` that their `TableInfo` says they should?

cb22:
Good idea - I've added a `FIXME` note so I don't forget!

PR #1374, opened 2023-12-20
ci: avoid unbounded loops

matklad:
As per second of the NASA's rules:

2. All loops must have fixed bounds. This prevents runaway code.

This is a better fix than https://github.com/tigerbeetle/tigerbeetle/pull/1367!


batiati (approved):
I love `for ... else` loops!

PR #1380, opened 2023-12-20
Client: Intrusive Requests

kprotty:
Allows for language clients to manage their own `Packet` memory, removing the need for `tb_client` to do so and thus removing the concepts of `acquire/release_packet` and `concurrency_max`. 

Also properly defines `deinit` semantics to cancel any pending packets before returning, even if inflight. Additionally requires that `submit` (or any other client call) must happen-before `deinit` and may not race with it.


src/clients/dotnet/TigerBeetle/NativeClient.cs
```diff
@@ -113,45 +109,36 @@ public void ReleasePacket(Packet packet)
...
-                    AssertTrue(packet != null);
-                    return new Packet(packet);
-                case PacketAcquireStatus.ConcurrencyMaxExceeded:
-                    throw new ConcurrencyExceededException();
-                case PacketAcquireStatus.Shutdown:
-                    throw new ObjectDisposedException("Client is closing");
-                default:
-                    throw new NotImplementedException();
-            }
+            return new Packet((TBPacket*)Marshal.AllocCoTaskMem(sizeof(TBPacket)));
```
batiati: Modern Dotnet has an interesting `MemoryPool` class that reuses buffers instead of allocating/freeing memory from the heap every time.

```C#
var pool = System.Buffers.MemoryPool<TBPacket>.Shared;
var owner = pool.Rent(1);
Memory<TBPacket> packet = owner.Memory;
var memoryHandle = packet.Pin();
//tb_submit(memoryHandle.Pointer);
            
// on_callback:
owner.Dispose();
```

Also, I’m not sure how `Marshal.AllocCoTaskMem` works on non-Windows platforms. It may just be a synonym for `AllocHGlobal`.

src/clients/java/src/main/java/com/tigerbeetle/NativeClient.java
```diff
@@ -18,57 +18,55 @@ private static final class NativeHandle implements Runnable {
...
-                atomicHandleReferences.decrementAndGet();
+                // After accessing the handle, remove our access from the ref.
+                // Observing REF_CLOSED bit set means close() happened during the submit()
+                // so remove an access from ClosePending. First thread to make ClosePending
+                // equal to zero knows no other thread is accessing the handle and can free it.
+                if ((atomicHandleReferences.addAndGet(-REF_ACCESS) & REF_CLOSED) != 0) {
+                    if (atomicHandleClosePending.addAndGet(-REF_ACCESS) == 0) {
+                        clientDeinit(handle);
+                    }
+                }
```
batiati: Neat! :heart:

src/clients/java/src/main/java/com/tigerbeetle/Request.java
```diff
@@ -74,11 +77,15 @@ protected Request(final NativeClient nativeClient, final Operations operation,
         this.sendBufferLen = batch.getBufferLen();
         this.replyBuffer = null;
 
+        // sizeof(tb_packet_t) + alignof(tb_packet_t).
```
batiati: Can we render those constants from `java_bindings.zig`?

src/clients/java/src/main/java/com/tigerbeetle/Client.java
```diff
@@ -99,12 +68,9 @@ public byte[] getClusterID() {
      * @throws RequestException refer to {@link PacketStatus} for more details.
      * @throws IllegalArgumentException if {@code batch} is empty.
      * @throws NullPointerException if {@code batch} is null.
-     * @throws ConcurrencyExceededException if there are more concurrent requests than defined at
-     *         {@link #Client(byte[], String[], int) concurrencyMax} parameter.
      * @throws IllegalStateException if this client is closed.
      */
-    public CreateAccountResultBatch createAccounts(final AccountBatch batch)
-            throws ConcurrencyExceededException {
+    public CreateAccountResultBatch createAccounts(final AccountBatch batch) throws Exception {
```
batiati: Since only unchecked exceptions are expected, I think we can remove `throws Exception` from all methods that used to throw `ConcurrencyExceededException`.

src/clients/java/src/client.zig
```diff
@@ -111,27 +109,26 @@ const NativeClient = struct {
...
-        }
-
-        return acquire_status;
+        // Compute a valid, aligned Packet pointer inside packet_buffer.
+        const packet_buf: []u8 = ReflectionHelper.get_packet_buffer_slice(env, request_obj) orelse
+            @panic("Request.packetBuffer somehow became null or invalid");
+        const packet_align_ptr =
+            std.mem.alignForward(usize, @intFromPtr(packet_buf.ptr), @alignOf(tb.tb_packet_t));
+        assert(packet_align_ptr + @sizeOf(tb.tb_packet_t) <=
+            @intFromPtr(packet_buf.ptr + packet_buf.len));
```
batiati: Sorry for going back and forth with this, but it seems too complicated (and adds alignment overhead!).
We could use [`java.lang.foreing`](https://docs.oracle.com/en%2Fjava%2Fjavase%2F22%2Fdocs%2Fapi%2F%2F/java.base/java/lang/foreign/package-summary.html) instead, but it's only available on Java 22.

I wonder if a plain `allocator.create(tb.tb_packet_t)` would be just simpler here (and with the same effect allocating off-heap memory) :smiling_face_with_tear:

batiati (commented):
I went halfway through the review today!
Great job @kprotty

src/clients/dotnet/TigerBeetle/NativeClient.cs
kprotty: Read that they were similar, but anything better is welcome. Was initially planning to use GCHandle.Alloc() but it takes an object rather than the struct. Will try out the MemoryPool approach.

src/clients/java/src/main/java/com/tigerbeetle/Client.java
kprotty: Removing `throws Exception` generates `unreported exception java.lang.Exception; must be caught or declared to be thrown`.

src/clients/dotnet/TigerBeetle/NativeClient.cs
kprotty: Any way to instantiate a `MemoryOwner<T>` from a `T*` or `IntPtr`? Current hurdle for using `MemoryPool`

batiati: I think you should keep the `IMemoryOwner` reference inside the `Request`, so you can call `Dispose()` on it when the callback arrives.

kprotty: Went with this approach for now.

src/clients/dotnet/TigerBeetle/Request.cs
```diff
@@ -1,151 +1,143 @@
...
 using System.Runtime.InteropServices;
 using System.Threading;
 using System.Threading.Tasks;
 using static TigerBeetle.AssertionException;
 
 namespace TigerBeetle;
 
-internal struct Packet
+
+internal abstract class NativeRequest
```
batiati, line 11: I liked the abstract class here instead of an `interface` :+1:

src/clients/dotnet/TigerBeetle/Request.cs
```diff
@@ -1,151 +1,143 @@
...
 {
-    public readonly unsafe TBPacket* Pointer;
+    private Tuple<IMemoryOwner<TBPacket>, MemoryHandle>? packetMemory = null;
```
batiati: We should reduce the GC work here by using `ValueTuple` instead.

But I think that using `GCHandle.Alloc(new TBPacket, GCHandleType.Pinned)` is even better, replacing the tuple, the `IMemoryOwner`, and the `MemoryHandle` with a single GCHandle. (The MemoryPool may be a good fit for allocating large buffers, however for small memory allocations like the `TBPacket` the runtime should be able to optimize it just fine).

src/clients/dotnet/TigerBeetle/Request.cs
```diff
@@ -1,151 +1,143 @@
...
 
-    public unsafe Packet(TBPacket* pointer)
+    protected unsafe void Submit(NativeClient nativeClient, TBOperation operation, void* data, int len)
     {
-        Pointer = pointer;
+        AssertTrue(packetMemory == null);
```
batiati: I am not sure if we can throw errors like this :thinking:.
Since this method is called directly by the unmanaged callback, any exception should be handled by the caller thread when we call `request.SetException(...)`.

src/clients/c/tb_client/context.zig
```diff
@@ -218,18 +218,21 @@ pub fn ContextType(
         }
 
         pub fn deinit(self: *Context) void {
-            const is_shutdown = self.shutdown.swap(true, .monotonic);
-            if (!is_shutdown) {
-                self.thread.join();
-                self.signal.deinit();
+            // Only one thread calls deinit() and it's UB for any further Context interaction.
+            assert(!self.shutdown.swap(true, .release));
```
batiati: Avoid `assert` with collateral effects, prefer assigning it to a `const` before the assertion.
Here and `on_submit`.

src/clients/c/tb_client/context.zig
```diff
@@ -263,6 +261,18 @@ pub fn ContextType(
                     @panic("IO.run() failed");
                 };
             }
+
+            // Cancel the request_inflight packet if any.
+            if (self.client.request_inflight) |*inflight| {
+                if (inflight.message.header.operation != .register) {
+                    const packet = @as(UserData, @bitCast(inflight.user_data)).packet;
+                    self.cancel(packet);
```
batiati, line 275: Add a comment here, maybe a `//TODO` explaining that in the future, we may want to signalize the cancelation of the inflight request with a different result code, so the application can handle it differently (i.e. retry the inflight request, and just error and abandon the queued ones).

src/clients/c/tb_client/context.zig
```diff
@@ -263,6 +261,18 @@ pub fn ContextType(
                     @panic("IO.run() failed");
                 };
             }
+
+            // Cancel the request_inflight packet if any.
+            if (self.client.request_inflight) |*inflight| {
+                if (inflight.message.header.operation != .register) {
+                    const packet = @as(UserData, @bitCast(inflight.user_data)).packet;
```
batiati, line 273: Can we assert `packet.next == null`? Here and `on_result`.

src/clients/java/src/main/java/com/tigerbeetle/NativeClient.java
```diff
@@ -18,57 +18,57 @@ private static final class NativeHandle implements Runnable {
...
         public NativeHandle(long handle) {
-            this.atomicHandle = new AtomicLong(handle);
+            this.handle = handle;
             this.atomicHandleReferences = new AtomicLong(0);
+            this.atomicHandleClosePending = new AtomicLong(0);
         }
 
-        public void submit(final Request<?> request) throws ConcurrencyExceededException {
+        // Silence PMD hint on `if (x) { if (y) {} }`. It is cleaner than `if (x && y)` here.
+        @SuppressWarnings("PMD.CollapsibleIfStatements")
```
batiati: We handle PMD exceptions at the file `exclude-pmd.properties`.

src/clients/java/src/main/java/com/tigerbeetle/Request.java
```diff
@@ -78,7 +78,7 @@ protected Request(final NativeClient nativeClient, final Operations operation,
             throw new IllegalArgumentException("Empty batch");
     }
 
-    public void beginRequest() throws ConcurrencyExceededException {
+    public void beginRequest() throws Exception {
```
batiati: If you remove this `throws Exception` here, all of them in `Client.java` can go away too.
Since `nativeClient.submit` does not throw checked exceptions, it does not need to include the throws clause.

I think some of the samples can get rid of the `throws Exception` as well.

src/clients/c/tb_client_header.zig
```diff
@@ -198,41 +197,44 @@ pub fn main() !void {
...
+        \\// completes submitted packets by invoking the callback with the given context.
         \\TB_STATUS tb_client_init(
         \\    tb_client_t* out_client,
         \\    tb_uint128_t cluster_id,
         \\    const char* address_ptr,
         \\    uint32_t address_len,
-        \\    uint32_t packets_count,
         \\    uintptr_t on_completion_ctx,
-        \\    void (*on_completion_fn)(uintptr_t, tb_client_t, tb_packet_t*, const uint8_t*, uint32_t)
+        \\    void (*on_completion)(uintptr_t, tb_client_t, tb_packet_t*, const uint8_t*, uint32_t)
```
batiati, line 208: :heart:

src/clients/go/tb_client.go
```diff
@@ -177,57 +176,52 @@ func (c *c_client) doRequest(
 		return 0, errors.ErrEmptyBatch{}
 	}
 
-	if c.tb_client == nil {
+	if c.ref_count.Add(2) & 1 != 0 {
```
batiati: Can we declare the constants `REF_ACCESS` and `REF_CLOSED` for this?

src/clients/go/tb_client.go
```diff
@@ -177,57 +176,52 @@ func (c *c_client) doRequest(
...
-	req.packet.data_size = C.uint32_t(count * int(getEventSize(op)))
-	req.packet.data = data
-
-	// Set where to write the result bytes.
-	req.result = result
-
-	// Submit the request.
-	C.tb_client_submit(c.tb_client, req.packet)
+	pinner.Pin(packet)
+	defer pinner.Unpin()
```
batiati: Should we move the `defer` to right after the declaration?

src/node.zig
```diff
@@ -54,14 +54,13 @@ fn init(env: c.napi_env, info: c.napi_callback_info) callconv(.C) c.napi_value {
...
         env,
         args[0],
         "replica_addresses",
     ) catch return null;
 
-    return create(env, cluster, concurrency, addresses) catch null;
+    return create(env, cluster, addresses) catch null;
 }
 
 fn deinit(env: c.napi_env, info: c.napi_callback_info) callconv(.C) c.napi_value {
```
batiati, line 66: Add a comment on `destroy` explaining that the single-thread runtime does not require synchronization on closing the client.

src/clients/dotnet/TigerBeetle/NativeClient.cs
```diff
@@ -98,60 +94,35 @@ private static NativeClient CallInit(InitFunction initFunction, UInt128Extension
...
-                default:
-                    throw new NotImplementedException();
+                lock (this)
+                {
+                    if (client != IntPtr.Zero)
+                    {
+                        tb_client_submit(client, packet);
+                        return;
+                    }
+                }
```
batiati: Replace the `lock` with the same pattern for synchronization used by the other clients.
In C# we can use `Interlocked` for atomic operations.

batiati (changes requested):

src/clients/dotnet/TigerBeetle/NativeClient.cs
kprotty: Thoughts on using a mutex for Go and `synchronized` for Java instead (replacing atomics)?

src/clients/dotnet/TigerBeetle/Request.cs
kprotty: Believe it should have at least one assert for `var request = GCHandle.FromIntPtr(packet->userData).Target ?? throw Exception`. A handle with a null Target seems to only happen for [Weak handles](<https://learn.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.gchandletype?view=net-8.0>) (otherwise its kept alive until Free) or for [already freed/invalid handles](<https://learn.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.gchandle.target?view=net-8.0#exceptions>)

src/clients/go/tb_client.go
kprotty: See above re: thoughts on `sync.Mutex`.

src/clients/dotnet/TigerBeetle/Request.cs
batiati: > Believe it should have at least one assert 

I agree, but I'm afraid the exception won't be caught by the caller, and possibly will only prevent the callback to complete...

In fact it should panic instead!
Maybe using Environment.FailFast:

http://msdn.microsoft.com/en-us/library/ms131100%28v=vs.110%29.aspx

src/clients/dotnet/TigerBeetle/NativeClient.cs
batiati: Usually we would avoid locking a mutex in the hot-path like `submit`, but I'm not sure if it would be called with a high enough frequency to make any noticeable difference.

Anyway, in [packet.zig](https://github.com/tigerbeetle/tigerbeetle/blob/e35fc23877aef917850c9af9b5591c8f0fb8da87/src/clients/c/tb_client/packet.zig#L57-L79) we already have a concurrent list guarded by a mutex, maybe we could get rid of this one and synchronize from the language side only ...

src/clients/java/src/main/java/com/tigerbeetle/NativeClient.java
```diff
@@ -18,57 +18,57 @@ private static final class NativeHandle implements Runnable {
...
-                            // Thread::onSpinWait method to give JVM a hint that the following code
-                            // is in a spin loop. This has no side-effect and only provides a hint
-                            // to optimize spin loops in a processor specific manner.
-                            Thread.onSpinWait();
-                        }
-
-                        // This function waits until all submitted requests are completed, and no
-                        // more packets can be acquired after that.
-                        clientDeinit(handle);
-                    }
```
batiati, line 70: Look @matklad, no spin-wait inside a synchronized block 😂😜

src/clients/dotnet/TigerBeetle/NativeClient.cs
kprotty: The mutex inside tb_client_submit syncs lang thread producer and client thread consumer (still needed even in JS which is single threaded).
The mutex here synchronizes lang thread submitters and the lang thread closer(s). Meaning langs without threading (JS) wouldnt need it.

Both mutexes could be replaced with lock-free idioms IIUC (As of posting this, noticed that the Mutex based submit queue in `packet.zig` doesn't seem to be used 🤔). Think it then comes down to a question of simplicity vs practical scalability. Submitting packets likely wont be the bottleneck of a client for some time.

batiati: Never mind, we don't need a `ConcurrentStack` anymore. It only held free packets (I thought it held queued requests instead).

So, please drop the type in `packet.zig` then! :tada:

batiati: > Think it then comes down to a question of simplicity vs practical scalability. Submitting packets likely wont be the bottleneck of a client for some time.

Agreed, no premature optimizations! :+1:

src/clients/java/src/main/java/com/tigerbeetle/NativeClient.java
```diff
@@ -15,61 +14,28 @@ final class NativeClient implements AutoCloseable {
...
-            this.atomicHandleClosePending = new AtomicLong(0);
         }
 
-        // Silence PMD hint on `if (x) { if (y) {} }`. It is cleaner than `if (x && y)` here.
-        @SuppressWarnings("PMD.CollapsibleIfStatements")
-        public void submit(final Request<?> request) throws Exception {
-            // Bump refs to access. Bail if REF_CLOSED bit was set.
-            // After observing REF_CLOSED, any other modifications which dont unset it don't matter.
-            if ((atomicHandleReferences.addAndGet(REF_ACCESS) & REF_CLOSED) != 0) {
+        public synchronized void submit(final Request<?> request) throws Exception {
```
batiati: Some considerations here:

1. I think that [double-checked locking](https://en.wikipedia.org/wiki/Double-checked_locking) will not save us much in this case. So, the C# implementation can be changed to match the other clients.

2. As library authors, we should not use `synchronized` methods in Java. A misused/malicious `lock` on the client instance could interfere with the intended behavior.

Instead use a private instance as a lock:

```java
private final Object lock = new Object();

public void submit() {
    synchronized(lock) {
        ...
    }
}
```

The C# implementation can use the same technique, but in C#, this pattern is called `syncRoot`.

```c#
private readonly object syncRoot = new object();

public void Submit()
{
    lock(syncRoot )
    {
         ...
    }
}
```

src/clients/java/src/main/java/com/tigerbeetle/NativeClient.java
```diff
@@ -15,27 +15,32 @@ final class NativeClient implements AutoCloseable {
...
+                if (handle == 0) {
+                    throw new IllegalStateException("Client is closed");
+                }
 
-            NativeClient.submit(handle, request);
+                NativeClient.submit(handle, request);
+            }
         }
 
         public synchronized void close() {
```
batiati, line 36:
```diff
- public synchronized void close() {
+ public void close() {
```

src/clients/dotnet/TigerBeetle/NativeClient.cs
```diff
@@ -144,8 +144,17 @@ public void Dispose()
...
-        OnComplete(packet, result, resultLen);
+        try
+        {
+            AssertTrue(ctx == IntPtr.Zero);
+            OnComplete(packet, result, resultLen);
+        }
+        catch (Exception e)
+        {
+            // The caller is unmanaged code, so if an exception occurs here we should force panic.
+            Console.WriteLine(e);
```
batiati: Let's drop the `Console` here since FailFast already dumps the exception into the proper logging channels.

src/clients/dotnet/TigerBeetle/Request.cs
```diff
@@ -10,48 +10,45 @@ namespace TigerBeetle;
...
     }
 
     public static unsafe void OnComplete(TBPacket* packet, ReadOnlySpan<byte> result)
     {
+        // Extract info from the packet before freeing it.
         var status = packet->status;
         var operation = packet->operation;
         var requestHandle = GCHandle.FromIntPtr(packet->userData);
 
-        AssertTrue(requestHandle.IsAllocated && requestHandle.Target != null, "Invalid GCHandle given to NativeRequest.Complete packet");
```
batiati: Why did you drop this assertion? Now that we are properly panicking the process, I think we should keep it.

src/clients/dotnet/TigerBeetle/Request.cs
```diff
@@ -10,48 +10,45 @@ namespace TigerBeetle;
...
 
-        AssertTrue(self.packetMemory != null, "NativeRequest completed without pinned packet memory");
-        (var packetOwner, var packetHandle) = self.packetMemory!;
-        self.packetMemory = null;
-
-        AssertTrue(packet == (TBPacket*)packetHandle.Pointer, "Mismatching pointer given to NativeRequest.Complete handler");
-        packetHandle.Dispose();
-        packetOwner.Dispose();
+        // Free the packet.
+        AssertTrue((IntPtr)packet == (request.packetHandle!).Value.AddrOfPinnedObject());
```
batiati:
```suggestion
        // Free the packet.
        AssertTrue(request.packetHandle != null);
        AssertTrue((IntPtr)packet == request.packetHandle!.Value.AddrOfPinnedObject());
```

src/clients/dotnet/TigerBeetle/NativeClient.cs
```diff
@@ -11,6 +11,7 @@ namespace TigerBeetle;
 internal sealed class NativeClient : IDisposable
 {
     private volatile IntPtr client;
+    private readonly object syncRoot = new object();
```
batiati: >The use of the NativeClient object as the mutex means a library
user could accidently lock the NativeClient and interfere with
tb_client synchronization. Instead, NativeClient now holds its
own internal object to acquire locks on.

Great write up King! After reading it, immediately I realized that the `NativeClient` instance is already private and never exposed to the final user (both in Java and C#) :facepalm:

So, there is no need for such pattern. You may completely ignore what I said before about using a private object as a lock, using `this` is just fine!

batiati:
Just some minor comments and we are good to go! Also, the CI is failing due to a `zig fmt` check.

batiati (changes requested):

src/clients/c/tb_client/context.zig
```diff
@@ -241,26 +214,26 @@ pub fn ContextType(
...
-
-                self.allocator.free(self.addresses);
-                self.allocator.free(self.packets);
-                self.allocator.destroy(self);
-            }
+            // Only one thread calls deinit() and it's UB for any further Context interaction.
+            const already_shutdown = self.shutdown.swap(true, .release);
+            assert(!already_shutdown);
+
+            // Wake up the run() thread for it to observe shutdowm=true, cancel inflight/pending
```
batiati:
```suggestion
            // Wake up the run() thread for it to observe shutdown=true, cancel inflight/pending
```

batiati (approved):
:tada:

PR #1388, opened 2023-12-21
Benchmark: use a histogram to store batch timings

cb22:
Previously, we required as much memory as transfers to store the timing of each one - to log out the deciles at the end. Switch this to using a histogram, with an upper limit of 10000ms. Now the benchmark uses a fixed amount of memory regardless of how many transfers we do.

If a percentile takes longer than 10000ms, we print a warning with it. There are [more advanced ways](https://www.timescale.com/blog/how-percentile-approximation-works-and-why-its-more-useful-than-averages/) to do this, but this shouldn't be a problem for now.

(This was causing annoyances when running larger scale tests.)

Also adds a missing deinit(), removes transfer timings (no one was looking at them :eyes:) and adds p95 and p99.

### DIFF
--- a/src/benchmark.zig
+++ b/src/benchmark.zig
@@ -127,18 +127,22 @@ pub fn main() !void {
         },
     );
 
-    const batch_accounts =
+    var batch_accounts =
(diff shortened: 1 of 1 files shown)


brson (approved):
It looks right to me.

PR #1389, opened 2023-12-21
Fix java context handle

batiati:
## The problem
The `contextHandle` can be in use by `submit` while the `deinit` function is called.
We synchronize `acquire_packet` to prevent new packets from being acquired during shutdown, but we didn't guard the `contextHandle` itself, causing the JVM to crash when `acquire_packet` was called with a disposed context.

## Solution
Guard the contextHandle and a reference counter behind atomics. It's cheaper than a lock.

Undo the temp fix introduced by https://github.com/tigerbeetle/tigerbeetle/pull/1375

### DIFF
--- a/src/clients/java/src/main/java/com/tigerbeetle/NativeClient.java
+++ b/src/clients/java/src/main/java/com/tigerbeetle/NativeClient.java
@@ -1,13 +1,18 @@
 package com.tigerbeetle;
 
 import static com.tigerbeetle.AssertionError.assertTrue;
+import java.util.concurrent.atomic.AtomicLong;
 
 final class NativeClient {
     static {
         JNILoader.loadFromJar();
     }
 
-    private volatile long contextHandle;
+    // Keeping the contextHandle and a reference counter guarded by
+    // atomics in order to prevent the client from using a disposed
+    // context during `close()`.
+    private final AtomicLong contextHandle;
+    private final AtomicLong contextHandleReferences;
 
     public static NativeClient init(final byte[] clusterID, final String addresses,
             final int concurrencyMax) {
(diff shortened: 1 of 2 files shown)


src/clients/java/src/main/java/com/tigerbeetle/NativeClient.java
```diff
@@ -31,31 +36,51 @@ private static void assertArgs(final byte[] clusterID, final String addresses,
...
+                    while (contextHandleReferences.getAcquire() > 0L) {
+                        // Thread::onSpinWait method to give JVM a hint that the following code is
+                        // in a spin loop. This has no side-effect and only provides a hint to
+                        // optimize spin loops in a processor specific manner.
+                        Thread.onSpinWait();
+                    }
+
+                    // This function waits until all submited requests are completed, and no more
+                    // packets can be acquired after that.
+                    clientDeinit(handle);
```
matklad, line 83: Heh, overall, I can't say I am too happy about this:

- abstractly, nesting synchronized and a spinloop (well, and spinloop by itself https://matklad.github.io/2020/01/02/spinlocks-considered-harmful.html :D) doesn't feel right. Ideally, I'd love to see something more principled here
- concretely, I think a flaw in this implementation is that, with many concurrent calls to close, only _one_ would actually wait for the stuff to get close. That is, it can be the case that a `.close()` call returns, but there's still some outstanding work. 

However, if I got this on the sync correctly, we are going to nix this code anyway soon, so I guess that's fine! Certainly better than an outright race we have at the moment!

matklad (approved):

src/clients/java/src/main/java/com/tigerbeetle/NativeClient.java
batiati: Yeah, my first thought was to encapsulate the context on a thread-safe class controlling all aspects more elegantly, but as you said, we are going to nuke this entire logic very soon.

About the `close` call, I don't think it can have side effects; the semantics are the same: only the first call to `close` will wait, no matter if concurrent or sequential calls.  If we want to solve that, we can just remove the first verification before the lock.

PR #1390, opened 2023-12-21
vsr: allow cluster of two primaries to viewchange

matklad:
Most of the time, passively abdicating is enough for the rest of the
cluster to trigger a view change. But there is an edge case --- if "the
rest of the cluster" is a single replica that believes itself to be a
primary (in an older view), and that replica doesn't have locally
uncommitted prepares, it will not itself try to abdicate, and it will
not learn that a newer view has started.

A _tactical_ fix here would be for the abdicating primary to sending an
SVC message, but let's use a brute force solution and just include view
information with pings. That way, we get a very robust way to
communicate the current view across the cluster.

We actually used to include view numbers with pings, but then got rid of
them, as:

- pings also carry clock synchronization information,
- we have a logic for dropping all messages whose view is > replica.view_durable()

The new insight here is that we can just include `view_durable` with
pings and pongs!

SEED: 9762771235456011173

### DIFF
--- a/src/vsr/replica.zig
+++ b/src/vsr/replica.zig
@@ -1237,6 +1237,7 @@ pub fn ReplicaType(
                 .command = .pong,
                 .cluster = self.cluster,
                 .replica = self.replica,
+                .view = self.view_durable(), // Don't drop pongs while the view is being updated.
                 // Copy the ping's monotonic timestamp to our pong and add our wall clock sample:
                 .ping_timestamp_monotonic = message.header.ping_timestamp_monotonic,
                 .pong_timestamp_wall = @as(u64, @bitCast(self.clock.realtime())),
@@ -2543,11 +2544,12 @@ pub fn ReplicaType(
                 .command = .ping,
                 .cluster = self.cluster,
                 .replica = self.replica,
-                .ping_timestamp_monotonic = self.clock.monotonic(),
-                // Checkpoint information:
+                .view = self.view_durable(), // Don't drop pings while the view is being updated.
                 .checkpoint_id = self.superblock.working.checkpoint_id(),
                 .checkpoint_op = self.op_checkpoint(),
+                .ping_timestamp_monotonic = self.clock.monotonic(),
             };
+            assert(ping.view <= self.view);
 
             self.send_header_to_other_replicas_and_standbys(ping.frame_const().*);
         }
@@ -8312,7 +8314,14 @@ pub fn ReplicaType(
                 .prepare, .commit => .normal,
                 // When we are recovering_head we can't participate in a view-change anyway.
                 // But there is a chance that the primary is actually running, despite the DVC/SVC.
-                .do_view_change, .start_view_change => if (self.status == .recovering_head) Status.normal else .view_change,
+                .do_view_change,
+                .start_view_change,
+                // For pings, we don't actually know where the new view is started or not.
+                // Conservatively transition to view change: at worst, we'll send a larger DVC
+                // instead of a RSV.
+                .ping,
+                .pong,
+                => if (self.status == .recovering_head) Status.normal else .view_change,
                 // on_start_view() handles the (possible) transition to view-change manually, before
                 // transitioning to normal.
                 .start_view => return,
--- a/src/vsr/replica_test.zig
+++ b/src/vsr/replica_test.zig
@@ -835,6 +835,57 @@ test "Cluster: view-change: DVC, 2/3 faulty header stall" {
     try expectEqual(t.replica(.R_).status(), .view_change);
 }
 
+test "Cluster: view-change: duel of the primaries" {
+    // In a cluster of 3, one replica gets partitioned away, and the remaining two _both_ become
+    // primaries (for different views). Additionally, the primary from the  higher view is
+    // abdicating. The primaries should figure out that they need to view-change to a higher view.
+    const t = try TestContext.init(.{ .replica_count = 3 });
+    defer t.deinit();
+
+    var c = t.clients(0, t.cluster.clients.len);
+    try c.request(20, 20);
+    try expectEqual(t.replica(.R_).commit(), 20);
+
+    try expectEqual(t.replica(.R_).view(), 1);
+    try expectEqual(t.replica(.R1).role(), .primary);
+
+    t.replica(.R2).drop_all(.R_, .bidirectional);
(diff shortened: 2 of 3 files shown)


src/vsr/replica.zig
```diff
@@ -2683,6 +2683,13 @@ pub fn ReplicaType(
                     self.replica,
                     self.view,
                 });
+                // Most of the time, passively abdicating is enough for the rest of the cluster to
+                // trigger a view change. But there is an edge case --- if "the rest of the cluster"
+                // is a single replica that believes itself to be a primary (in an older view), and
+                // that replica doesn't have locally uncommitted prepares, it will not itself try to
+                // abdicate, and it will not learn that a newer view has started. Sending an SVC
+                // here will have an effect of broadcasting our view number.
+                self.send_start_view_change();
```
matklad: Two things I am wondering here:

- should we perhaps be sending an SVC here for reasons unrelated to this particular VOPR failure? Like, we need a quorum of SVC, and it feels like if there's a primary abdicating, it should be able to contribute to that quorum? 
- if there are _no_ other good reasons to SVC here, then sending SVC _just_ to broadcast our view feels hacky. It feels like we shuold use pings to keep cluster in sync regarding the current view. I know we moved back from that, as replicas aren't allowed to announce their view number until that view number hit the disk in the superblock, and just delaying ping would wreck clock synchronisation.

  But I am wondering... Can replicas just attach `replica.superblock.working.vsr_state.view` to ping messages? For _pings_, we _dont_ care, I think, if the view is older, as long as it is not newer?

matklad:
(there's a replica test failure here which I need to look into (looks like views accidentally shifting in an unrelated test), but I want to get feedback on the idea of sending SVC here)

sentientwaffle:
```
no liveness, final cluster state (core=101):
 0   / .      114V  95/101/101C  71:102Jo  0/_0J!  71:102Wo <__0:__0>   48960Gf  0G!   0G?   1/4Pp  0/3Rq
 1   \  .     114V  95/101/101C  71:102Jo  0/_0J!  71:102Wo <__0:__0>   48960Gf  0G!   0G?  
 2   /   .    113V  95/102/102C  71:102Jo  0/_0J!  71:102Wo <__0:__0>   48938Gf  0G!   0G?   0/4Pp  0/3Rq
```

I was initially confused why R0 is behind R2 by a commit:
- I see R0 sending the `op=102` prepare to R2.
- R2 dutifully replies with a `prepare_ok`...
- ...to _itself_, since R2 thinks R2 is the primary! :laughing: 

(And the prepare in question was from the old view, so R2 doesn't learn the new view this way.)

src/vsr/replica.zig
sentientwaffle: > For pings, we dont care, I think, if the view is older, as long as it is not newer?

Agreed; I think this is safe.

I lean towards putting more on ping, to avoid the "hacky" solution.

As a third option though, in the past we have considered folding SVC into ping (i.e. SVC is just a flag on ping, not a separate message). If we go that route, then these two solutions converge.

matklad: On a morning walk, I realized that we actually need this SVC! Consider a Murphy's law scenario, where:

- only view-change quorum of replicas are alive
- moreover, they are connected as a star
- and the current primary is not in the center of the star
 
In this situation, we'd want for the center to become a new primary, but the center needs a quorum of SVCs, _including_ the one from the abdicating primary. But this is deeply into the SCC connectivity which we are not handling all that way today anyway.

I'll go with the ping solution --- it feels far superior as bullet-proof way to disseminate current view information across the cluster. Not sure why we didn't just code it that way back then, seems obvious in retrospect.

src/vsr/replica_test.zig
```diff
@@ -835,6 +835,57 @@ test "Cluster: view-change: DVC, 2/3 faulty header stall" {
...
+
+    // The stage is set: we have two primaries in different views, R2 is about to abdicate.
+    try expectEqual(t.replica(.R1).view(), 1);
+    try expectEqual(t.replica(.R1).role(), .primary);
+    try expectEqual(t.replica(.R1).commit(), 21);
+    try expectEqual(t.replica(.R2).role(), .primary);
+    try expectEqual(t.replica(.R2).commit(), 20);
+
+    t.replica(.R0).drop_all(.R_, .bidirectional);
+    t.run();
```
sentientwaffle: What is the reason for this `t.run()`?  :thinking:

matklad: Ahh, good catch, this fragment was actually just me debugging the test originally!

src/vsr/replica_test.zig
```diff
@@ -835,6 +835,51 @@ test "Cluster: view-change: DVC, 2/3 faulty header stall" {
...
+    t.replica(.R2).drop(.R0, .bidirectional, .prepare_ok);
+    t.replica(.R2).drop(.R0, .outgoing, .do_view_change);
+    t.run();
+
+    // The stage is set: we have two primaries in different views, R2 is about to abdicate.
+    try expectEqual(t.replica(.R1).view(), 1);
+    try expectEqual(t.replica(.R1).role(), .primary);
+    try expectEqual(t.replica(.R1).commit(), 21);
+    try expectEqual(t.replica(.R2).role(), .primary);
+    try expectEqual(t.replica(.R2).commit(), 20);
```
```suggestion
    // The stage is set: we have two primaries in different views, R2 is about to abdicate.
    try expectEqual(t.replica(.R1).view(), 1);
    try expectEqual(t.replica(.R1).status(), .normal);
    try expectEqual(t.replica(.R1).role(), .primary);
    try expectEqual(t.replica(.R1).commit(), 21);
    try expectEqual(t.replica(.R2).status(), .normal);
    try expectEqual(t.replica(.R2).role(), .primary);
    try expectEqual(t.replica(.R2).commit(), 20);
    try expectEqual(t.replica(.R2).op_head(), 21);
```

The `status=normal` and `op_head > commit_min` assertion emphasizes/checks that R2 will indeed abdicate.

matklad: Ah, right, `.role() == .primary` doesn't imply `.normal`, unlike the `replica.primary()` function!

sentientwaffle (dismissed):

src/vsr/replica_test.zig
matklad: And I also forgot to assert R2's view!

sentientwaffle (approved):

PR #1392, opened 2023-12-22
Benchmark: tighten up variable names

cb22:


matklad (approved):
Wanted to do a quick approve here, but bogged down trying to understand the code, and wrote https://github.com/tigerbeetle/tigerbeetle/pull/1444 as a result :-) 

I _think_ we could do a bit better here:

* The `_count` naming is confusing here --- this _isn't_ `count/index` pair, that's the sum of values in the histogram
* there's no need to pass this count it --- given that we are going to do a bunch of runs over the buckets anyway, we might just as well compute it here

PR #1393, opened 2023-12-22
ci: ensure all scripts are compiling

matklad:
Unify all script binaries (ci&release) under the same umbrella `zig build scripts --`. That way:

- they are always compiled and can't bitrot,
- we only pay linking cost once,
- build.zig shrinks in size.

Also unify `flags.parse_command` & `flags.parse_flags` into a single `flags.parse`. Not sure why I didn't do that from the start, seems obviously better to have just one entry point?

### DIFF
--- a/.github/workflows/clients.yml
+++ b/.github/workflows/clients.yml
@@ -29,7 +29,7 @@ jobs:
             7.0.x
 
       - run: ./scripts/install_zig.${{ matrix.os == 'windows-latest-large' && 'bat' || 'sh' }}
-      - run: ./zig/zig build ci -- --language=dotnet
+      - run: ./zig/zig build scripts -- ci --language=dotnet
       # - run: ./zig/zig build client_docs -- --language=dotnet
       # - if: matrix.os == 'ubuntu-latest-large'
         # run: ./.github/ci/fail_on_diff.sh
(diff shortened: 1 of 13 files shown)


.github/workflows/clients.yml
```diff
@@ -29,7 +29,7 @@ jobs:
             7.0.x
 
       - run: ./scripts/install_zig.${{ matrix.os == 'windows-latest-large' && 'bat' || 'sh' }}
-      - run: ./zig/zig build ci -- --language=dotnet
+      - run: ./zig/zig build scripts -- ci --language=dotnet
```
matklad, line 32: That's a bit of a clunky entry point if you invoke it from your shell, but I think we can live with that.

I am _tempted_ to write some hack so that you get an `./x` binary in the root of the project, so that you can do

```
./x ci
./x simulate
```

and such. Resisting the temptation for now, one step at a time. `-Drelease` was already a big win in that direction!

cb22:
@matklad looks like there are a few more call sites.

matklad:
Ah, should have grepped for `flags.parse_` rather than for `flags.parse`, thought that that one was already upgraded on the first inspection.

cb22 (approved):

PR #1395, opened 2023-12-22
build: simplify building fuzzers:

matklad:
Rule-of-thumb: size of build.zig should be O(1). That is, number of lines in the build script should be independent of the size of the project. Fuzzers currently violate that --- each fuzzer requires a dedicated target in build.zig.

Rectify that by building a single fat fuzzy executable which dispatches on the CLI arg to pick the fuzzer.

We also had ideas in the past about fuzzers being just `test` tests, and there's perhaps some merit to that, and I'd love to think about that more, but the current step already feels like an improvement.


src/fuzz_tests.zig
```diff
@@ -0,0 +1,55 @@
...
+const flags = @import("./flags.zig");
+const fuzz = @import("./testing/fuzz.zig");
+const fatal = flags.fatal;
+
+const log = std.log.scoped(.fuzz);
+
+// NB: this changes values in `constants.zig`!
+pub const tigerbeetle_config = @import("config.zig").configs.test_min;
+
+const Fuzzers = .{
```
cb22, line 13: Looks like the forest fuzz is missing here? (and `manifest_level_fuzz.zig` but IIRC that was broken? Perhaps best to include it as a comment...)

Or maybe we can build this list from searching for `_fuzz.zig` or still keep it static but assert its equal? (of course that's easier when all fuzzers build!)

.github/workflows/linux.yml
```diff
@@ -165,21 +165,21 @@ jobs:
...
-      - run: zig/zig build build_fuzz_vsr_free_set
-      - run: zig/zig build build_fuzz_lsm_forest
+      - run: zig/zig build fuzz -- ewah                   --seed=123
+      - run: zig/zig build fuzz -- lsm_segmented_array    --seed=123
+      - run: zig/zig build fuzz -- lsm_tree               --seed=123 --events-max=400
+      - run: zig/zig build fuzz -- vsr_journal_format     --seed=123
+      - run: zig/zig build fuzz -- vsr_superblock         --seed=123 --events-max=3
+      - run: zig/zig build fuzz -- vsr_superblock_quorums --seed=123
+
+      # These are to slow to run on CI.
```
cb22: Heh - so with `events-max` these aren't slow to run at all. Except for `vsr_free_set` which seems to ignore it.

.github/workflows/linux.yml
```diff
@@ -165,21 +165,21 @@ jobs:
...
-      - run: zig/zig build fuzz_lsm_tree -- --seed 123 --events-max 400
-      - run: zig/zig build fuzz_vsr_journal_format -- --seed 123
-      - run: zig/zig build fuzz_vsr_superblock -- --seed 123 --events-max 3
-      - run: zig/zig build fuzz_vsr_superblock_quorums -- --seed 123
-
-      # Build the fuzzers, but don't run them, since they're slow.
-      - run: zig/zig build build_fuzz_lsm_manifest_log
-      - run: zig/zig build build_fuzz_vsr_free_set
-      - run: zig/zig build build_fuzz_lsm_forest
+      - run: zig/zig build fuzz -- ewah                   --seed=123
```
cb22: Thinking about it; the point of these is to do a smoke test on all of our fuzzers to make sure they build and at least pass at a very basic level. 

Given that, would be handy if we could just do `zig/zig build fuzz -- all --seed=123 --events-max=10` here and keep our yaml O(1) :wink:.

scripts/fuzz_loop.sh
```diff
@@ -9,7 +9,7 @@ FUZZ_COMMAND=$1
 
 while true; do
```
cb22: I wonder if anyone uses these scripts...

src/fuzz_tests.zig
matklad: :fearful: `manifest_level_fuzz` doesn't compile, and `forest_fuzz` crashes with null pointer! Will deal with that in a separate PR. 

I don't think we need to specifically protect from missing fuzzers --- given that this is now the single entry point, I think this mistake will be unlikely for any new fuzzers. 

What would be cool though, is to check that any `.zig` file we have is used by _something_, not sure how to do that.

.github/workflows/linux.yml
matklad: Yeah, I'll try to re-enable them in a follow up, when I also deal with lsm_level not compiling and forest crashing.

.github/workflows/linux.yml
matklad: Most excellent suggestion! It also allows us to take advantave of exhaustiveness checking, which immediately points out that there's also `cache_map` fuzz, which is conspiciously absent form our CI setup! Will run that in a follow up!

scripts/fuzz_loop.sh
matklad: The easiest way to find out would be to kill them :D

.github/workflows/linux.yml
```diff
@@ -158,28 +158,12 @@ jobs:
...
-      - run: zig/zig build fuzz_vsr_superblock_quorums -- --seed 123
-
-      # Build the fuzzers, but don't run them, since they're slow.
-      - run: zig/zig build build_fuzz_lsm_manifest_log
-      - run: zig/zig build build_fuzz_vsr_free_set
-      - run: zig/zig build build_fuzz_lsm_forest
-
-      # # This both checks that the hash_log builds and acts as a regression test for
-      # # https://github.com/tigerbeetle/tigerbeetle/issues/404
-      # - run: zig/zig build fuzz_lsm_forest -Dhash-log-mode=create -Drelease -- --seed 16319736705930193193 --events-max 10000 && zig/zig build fuzz_lsm_forest -Dhash-log-mode=check -- --seed 16319736705930193193 --events-max 10000
```
matklad, line 182: This one (comparing that .release & .debug work the same) is killed without replacement. That's a useful idea, but, given that this is commented out, I think it's enough if we just remember this as an idea, to be used one we need it.

.github/workflows/linux.yml
```diff
@@ -158,28 +158,12 @@ jobs:
...
+      - run: zig/zig build fuzz -- smoke
```
cb22, line 166: :tada:

cb22 (approved):
:rocket:

src/fuzz_tests.zig
matklad: >What would be cool though, is to check that any .zig file we have is used by something, 

https://github.com/tigerbeetle/tigerbeetle/pull/1404

PR #1397, opened 2023-12-29
Don't exit repl on 'help'

brson:
One of the first things I did with tigerbeetle, per the docs, was run the repl, and the first thing I did in the repl was type 'help', which immediately exited the repl.

I thought this behavior was surprising (e.g. the python repl doesn't exit on help), and I couldn't find an obvious reason for it in the code or the commit log.

This change just removes the call to `exit`, so the repl continues to run after displaying the help message.

### DIFF
--- a/src/tigerbeetle/repl.zig
+++ b/src/tigerbeetle/repl.zig
@@ -448,7 +448,6 @@ pub fn ReplType(comptime MessageBus: type) type {
                 },
                 .help => {
                     try repl.display_help();
-                    std.os.exit(0);
                 },
 
                 .create_accounts,


brson:
CI failure looks unrelated to me.

matklad:
Yeah, makes sense! In general, our repl exists, but is not good, there should be plenty of low-hanging fruit there (and one rather hard to crack coconut as well: on the one hand, we'd love to have readline conveniences, on the other hand, we are not thrilled about linking to readline or implementing the required functionality ourselves).

Will look into CI failure, it is indeed unrelated: we try to test that our client library is installable on different distros, but that turns into an exercise of installing node itself, and apparently something broke there?

matklad (approved):

matklad:
See https://github.com/tigerbeetle/tigerbeetle/pull/1398 for CI failure fix.

PR #1398, opened 2023-12-30
node: prune down CI

matklad:
For NodeJS, we used to have a hefty matrix of different distributions, to make sure our stuff works in all possible contexts. However, it looks like the main thing we are testing here is installing node itself --- we used something called `nodesource` which provides third party packages for various distros, and requires add-hoc scripts for installation. This is also something that was changed recently (they used to have a single script, which is now deprecated). And it looks like it broke again:

https://github.com/tigerbeetle/tigerbeetle/actions/runs/7359371489/job/20034038580?pr=1397#step:5:573

It feels like here we are taking onto us some else's maintainance burden. I'd say that, if official node does not provide one-liner installation for these distros, we shouldn't invent our own wheel.

Let's just check that the two official images from NodeJS work, it feels to be a better balance of coverage vs maintainance work!

### DIFF
--- a/src/clients/node/ci.zig
+++ b/src/clients/node/ci.zig
@@ -39,47 +39,7 @@ pub fn tests(shell: *Shell, gpa: std.mem.Allocator) !void {
     // Container smoke tests.
     if (builtin.target.os.tag == .linux) {
         try shell.exec("npm pack --quiet", .{});
-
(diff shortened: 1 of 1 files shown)


batiati (approved):
> if official node does not provide one-liner installation for these distros, we shouldn't invent our own wheel.

Agreed! 💯

PR #1399, opened 2023-12-30
repl: avoid build.zig special case

matklad:
As a general rule of thumb, adding new functionality to a project should not require adding stuff to build.zig --- the more different things we need to build, the higher our maintenance burden down the line.

We violate this rule-of-thumb for repl due to a combination of two factors:

- repl's code is a part of binary module, it's not part of the vsr library module
- repl has tests

But our unit tests live inside the vsr module, they `@import("./vsr.zig")`, rather than `@import("vsr")`.

That's why we compile repl tests as a completely separate binary.

This feels like needless complexity --- let's just move repl stuff inside vsr?


matklad:
Do we remeber _why_ we didn't do this to begin this? Was there a strong reason for repl.zig to _not_ be inside vsr?

sentientwaffle:
> Do we remeber why we didn't do this to begin this? Was there a strong reason for repl.zig to not be inside vsr?

Idk; a couple of guesses though:
- Maybe the repl was originally going to be a separate binary?
- Maybe it is b/c repl is accounting-statemachine-specific, whereas vsr is statemachine-agnostic?

sentientwaffle (approved):

PR #1400, opened 2023-12-30
Drop support for .Net Standard 2.1

batiati:
In practice, drops the support for .Net Core 3.1 (already in [EOL](https://dotnet.microsoft.com/en-us/platform/support/policy/dotnet-core)) and .Net <= 6 (in [EOL](https://dotnet.microsoft.com/en-us/platform/support/policy/dotnet-core) in Dec 2024). The minimum supported version now is .Net 7.0; Recommended version is .Net 8.0 LTS.

- Simplifies the code, removing the custom `UInt128` type, and version-specific conditionals (i.e `#if NETSTANDARD` and `#if !NET7_0_OR_GREATER`).

- Removes the burden of publishing dual-version artifacts on Nuget.

- Speeds up the CI by not having to run the same tests for each supported version.

- Bonus: Introduces ~top-level~ file-scoped namespaces, saving one level of indentation.

### DIFF
--- a/src/clients/dotnet/TigerBeetle.Tests/ExceptionTests.cs
+++ b/src/clients/dotnet/TigerBeetle.Tests/ExceptionTests.cs
@@ -9,16 +9,17 @@ namespace TigerBeetle.Tests
         [TestMethod]
         public void AssertTrue()
         {
-            AssertionException.AssertTrue(1 == 1);
-            Assert.IsTrue(true);
+            // Should not throw an exception
+            // when the condition is true.
+            AssertionException.AssertTrue(true);
         }
 
         [TestMethod]
         [ExpectedException(typeof(AssertionException))]
         public void AssertFalse()
         {
-            AssertionException.AssertTrue(1 == 0);
-            Assert.IsTrue(true);
+            // Expected AssertionException.
+            AssertionException.AssertTrue(false);
         }
 
         [TestMethod]
@@ -26,7 +27,9 @@ namespace TigerBeetle.Tests
         {
             try
             {
-                AssertionException.AssertTrue(1 == 0, "hello {0}", "world");
+                AssertionException.AssertTrue(false, "hello {0}", "world");
+
+                // Should not be reachable:
                 Assert.IsTrue(false);
             }
             catch (AssertionException exception)
@@ -38,8 +41,7 @@ namespace TigerBeetle.Tests
         [TestMethod]
         public void AssertTrueWithMessage()
         {
-            AssertionException.AssertTrue(1 == 1, "unreachable");
-            Assert.IsTrue(true);
+            AssertionException.AssertTrue(true, "unreachable");
         }
 
         [TestMethod]
--- a/src/clients/dotnet/TigerBeetle.Tests/ExceptionTests.cs
+++ b/src/clients/dotnet/TigerBeetle.Tests/ExceptionTests.cs
@@ -1,82 +1,81 @@
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 using System;
 
-namespace TigerBeetle.Tests
+namespace TigerBeetle.Tests;
+
+[TestClass]
+public class ExceptionTests
 {
-    [TestClass]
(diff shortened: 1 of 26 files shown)


matklad:
> Bonus: Introduces top-level namespaces, saving one level of indentation.

Next time, would be great to file indentation/formatting changes as a separate commit, 3k diff is a bit intimidating ;-)

src/clients/dotnet/TigerBeetle.Tests/ExceptionTests.cs
```diff
@@ -1,81 +1,81 @@
...
+
+[TestClass]
+public class ExceptionTests
 {
-    [TestClass]
-    public class ExceptionTests
+    [TestMethod]
+    public void AssertTrue()
+    {
+        AssertionException.AssertTrue(1 == 1);
```
matklad: In a separate PR, could you add a documentation for `AssertionException`? This class, and these tests in particular, look quiet peculiar, as they seem tautological --- we we aren't just using `Assert`, why we need our own `AssertionException`? 

My _guess_ here is that the built-in assertion mechanism in .Net is off by default, so we need our own asserts that are guaranteed to always through, but I don't really know. Would be great to have that documented for future readers!

src/clients/dotnet/TigerBeetle.Tests/UInt128Tests.cs
```diff
@@ -4,149 +4,102 @@
...
+    [ExpectedException(typeof(ArgumentException))]
+    public void InvalidArrayConvertion()
+    {
+        byte[] array = new byte[17];
+        _ = array.ToUInt128();
+    }
+
+
+    [TestMethod]
+    public void BigIntegerConvertion()
```
matklad, line 70: In a separate PR, let's also add tests for small numbers, like `0`, and also a test for a _negative_ BigInteger (which we expect to throw)

matklad (dismissed):
LGTM!

I haven't reviewed every line, and I don't know .Net enough to review for unknown-unknowns anyway, but this all reads like a clear improvement, thanks batiati!

batiati:
> would be great to file indentation/formatting changes as a separate commit, 3k diff is a bit intimidating ;-)

It's the last day of the year! Let's do it right now 😂
Also, the review's comments were addressed in this PR.

src/clients/dotnet/TigerBeetle/AssertionException.cs
```diff
@@ -2,6 +2,12 @@
 
 namespace TigerBeetle
 {
+    /// <summary>
+    /// The exception that is thrown when an assertion used for correctness checks is triggered.
+    /// Correctness check assertions differ from other assertions offered by the .Net ecosystem
+    /// i.e. Tracer.Assert and Debug.Assert, because they are not meant to be disabled or re-routed.
+    /// It's recommended that the application handle AssertionException as unrecoverable fatal errors.
+    /// </remarks>
```
matklad: Opening tag (summary) and closing tag (remarks) do not match?

batiati: Nice catch, surely they don't match 😳

Important thing, this exposed that we are not generating the XML documentation file. Otherwise, it should be an error (or maybe a warning) during the build.

matklad (approved):

src/clients/dotnet/TigerBeetle/TigerBeetle.csproj
```diff
@@ -9,6 +9,9 @@
     <AssemblyName>TigerBeetle</AssemblyName>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
     <RollForward>LatestMajor</RollForward>
+    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+    <GenerateDocumentationFile>true</GenerateDocumentationFile>
+    <NoWarn>1591</NoWarn> <!-- Disables the warning CS1591: Missing XML comment for publicly visible type or member -->
```
matklad, line 14: :heart:

PR #1402, opened 2024-01-02
workbench: initial

matklad:
## What & Why

Add a workbench --- a simple HTML page, maintained as a part of
TigerBeetle monorepo, whose purpose is to serve any kind of developer
related information in a low-effort way. Basically, this is GUI for all
our automation.

Currently, workbench displays two pieces of information:

1. Release rotation (subsuming the python script)
2. Results of macro benchmarks across commits (binary size & build time)

To enable 2, this also adds some continuous benchmarking (CB)
infrastructure.

 ## How: Benchmark

CB is triggered by an GH action on the main branch. This action
intentionally isn't a CI gate (yet?). The idea is that we want to track
long-term trends in numbers, and find obvious regressions within a week.

CB is orechstrated by the `scripts/workbench.zig` script. There's no
specific rules what goes there, but, for starters, track just two
numbers:

- the size of a release binary,
- the time to build the software.

These probably aren't the most impactful things to track, but the idea
here is to establish a process of tracking. Adding _more_ things should
be relatively trivial, once the base infrastructure is in place.

Benchmark results across many runs are stored as a JSON file in the
<https://github.com/tigerbeetle/workbenchdb> repository, using `git` as
a distributed database of sorts.

Specifically:

- `workbench/data.json` is the file (the directory to allow "log
  rotation" once a year)
- it is a newline delimited JSON --- each line is a JSON object
- each objects records a "run" --- a set of measurements at a particular
  commit and a timestamp
- a measurement has a label, a value, and a unit (bytes, ms, etc)
- aggregating measurements with the same label across runs gives a time
  series
- the set of measurements can change over time --- different runs might
  have different sets of measurements
- unit is convenient both for humans and for the front-end to display
  results without hard-coding a list of measurements.
- as a TODO, it would perhaps be nice to track hosts cpu, memory and an
  OS in the run.

It is assumed that a single workbench is run at a time, so no provision
for merge conflicts is made.

  ## How: Workbench

A GitHub pages without any build step --- just some static files in
`src/workbench` directory. If this grows big, we'll probably want to
switch to TypeScript at least, but maybe it needn't grow big.

To display benchmark results, the data is `fetch`ed from
`raw.githubusercontent` and displayed using <https://apexcharts.com/>. No
affiliation with this particular library, it's just something I saw on
<https://deno.com/benchmarks> and I quite like that page.

### DIFF
--- a/src/workbench/index.html
+++ b/src/workbench/index.html
@@ -1,0 +1,35 @@
+<!DOCTYPE html>
+<html lang="en">
+
+<head>
+  <meta charset="UTF-8">
+  <meta name="viewport" content="width=device-width, initial-scale=1.0">
+  <title>TigerBeetle Dev Workbench</title>
+  <script src="./workbench.js"></script>
+  <script src="https://cdnjs.cloudflare.com/ajax/libs/apexcharts/3.45.2/apexcharts.min.js"
+    integrity="sha512-vIqZt7ReO939RQssENNbZ+Iu3j0CSsgk41nP3AYabLiIFajyebORlk7rKPjGddmO1FQkbuOb2EVK6rJkiHsmag=="
+    crossorigin="anonymous" referrerpolicy="no-referrer"></script>
+</head>
+
+<body>
+  <h1>TigerBeetle Dev Workbench</h1>
+
+  <section>
+    <h2>Release</h2>
+
+    Release Manager for:
+    <ul>
+      <li>last week: <span id="release-previous">N/A</span> </li>
+      <li><strong>this week: <span id="release-current">N/A</span></strong></li>
+      <li>next week: <span id="release-next">N/A</span></li>
+    </ul>
+  </section>
+
+  <section>
+    <h2>Release Metrics</h2>
+    <div id="charts" style="display: flex; flex-direction: column;">
+    </div>
+  </section>
+</body>
+
+</html>
--- a/src/workbench/workbench.js
+++ b/src/workbench/workbench.js
@@ -1,0 +1,206 @@
+// Code powering "developer dashboard" aka workbench, at <https://tigerbeetle.github.io>.
+//
+// At the moment, it isn't clear what's the right style for this kind of non-Zig developer facing
+// code, so the following is somewhat arbitrary:
+//
+// - camelCase naming
+// - `deno fmt` for style
+// - no TypeScript, no build step
+
+async function main() {
+  const runs = await fetchData();
+  const series = runsToSeries(runs);
+  plotSeries(series, document.querySelector("#charts"));
+
+  const releaseManager = getReleaseManager();
+  for (const week of ["previous", "current", "next"]) {
+    document.querySelector(`#release-${week}`).textContent =
+      releaseManager[week];
+  }
+}
+
+window.onload = main;
+
+function assert(condition) {
+  if (!condition) {
+    alert("Assertion failed");
+    throw "Assertion failed";
+  }
+}
+
+function getReleaseManager() {
+  const week = getWeek(new Date());
+  const candidates = [
+    "batiati",
+    "cb22",
+    "kprotty",
+    "matklad",
+    "sentientwaffle",
+  ];
+  candidates.sort();
+
+  return {
+    previous: candidates[week % candidates.length],
+    current: candidates[(week + 1) % candidates.length],
+    next: candidates[(week + 2) % candidates.length],
+  };
+}
+
+const dataUrl =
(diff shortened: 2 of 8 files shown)


matklad:
Deployed here: https://matklad.github.io/tigerbeetle/

Not much to see there without much data, so use your imagination ;)

matklad:
image

I am pretty sure that we'll need quite a few iterations on the UI side of things, but:

* I am pretty confident about underlying metrics data model:

```zig
const Measurement = struct {
    label: []const u8,
    value: u64,
    unit: []const u8,
};

const Run = struct {
    timestamp: i64,
    revision: []const u8,
    measurements: []const Measurement,
};
```

I'd say we want to collect at least some measurements of this shape and then think about how to display them best. 

* I am somewhat confident about the overall shape of things here: a zig script that is triggered when new code gets into main branch, which pushes the results to some git repo, and which results are subsequently fetched and visualized by client-side javascript

src/workbench/index.html
```diff
@@ -0,0 +1,33 @@
+<!DOCTYPE html>
+<html lang="en">
+
+<head>
+  <meta charset="UTF-8">
+  <meta name="viewport" content="width=device-width, initial-scale=1.0">
+  <title>TigerBeetle Dev Workbench</title>
+  <script src="./workbench.js"></script>
+  <script src="https://cdn.jsdelivr.net/npm/apexcharts"></script>
```
cb22: Perhaps worth pinning this and using SRI since it's easy enough...

```suggestion
  <script src="https://cdnjs.cloudflare.com/ajax/libs/apexcharts/3.45.2/apexcharts.min.js" integrity="sha512-vIqZt7ReO939RQssENNbZ+Iu3j0CSsgk41nP3AYabLiIFajyebORlk7rKPjGddmO1FQkbuOb2EVK6rJkiHsmag==" crossorigin="anonymous" referrerpolicy="no-referrer"></script>
```

src/workbench/workbench.js
```diff
@@ -0,0 +1,210 @@
...
+
+// The input data is array of runs, where a single run contains many measurements (eg, file size,
+// build time).
+//
+// This function "transposes" the data, such that measurements with identical labels are merged to
+// form a single array which is what we want to plot.
+//
+// This doesn't depend on particular plotting library though.
+function runsToSeries(runs) {
+  let result = new Map();
```
cb22: Nit, this can be `const`

src/workbench/workbench.js
```diff
@@ -0,0 +1,210 @@
...
+        name: series.label,
+        data: series.timestamp.map((t, i) => [t * 1000, series.value[i]]),
+      }],
+      xaxis: {
+        type: "datetime",
+      },
+    };
+
+    if (series.unit === "bytes") {
+      options = Object.assign(options, {
```
cb22: I don't really mind `Object.assign` but usually I'd go with either the spread operator

```
options = {...options, yaxis: {...}}
```

or just setting `yaxis` directly

```
options.yaxis = {...}
```

but up to you!

src/workbench/workbench.js
```diff
@@ -0,0 +1,210 @@
...
+        yaxis: {
+          labels: {
+            formatter: formatDuration,
+          },
+        },
+      });
+    }
+
+    let div = document.createElement("div");
```
cb22: both this `let` and the `var` below can be `const`

src/workbench/workbench.js
```diff
@@ -0,0 +1,210 @@
...
+    parts.push(`${milliseconds}ms`);
+  }
+
+  return parts.join(" ");
+}
+
+// Returns the ISO week of the date.
+//
+// Source: https://weeknumber.com/how-to/javascript
+Date.prototype.getWeek = function () {
```
cb22: Perhaps a standalone fn rather than extending the built in prototype?

src/workbench/index.html
matklad: Thanks, SRI is not in my active vocabulary ...

src/workbench/workbench.js
matklad: ... and `var` apparently still is :facepalm:

src/workbench/workbench.js
matklad: Right, not sure what I was thinking here... And thanks for the tip with spread --- didn't know that it works with objects as well!

src/workbench/workbench.js
matklad: Yeah...

sentientwaffle (approved):

PR #1405, opened 2024-01-08
Add example systemd unit and documentation on using it

### LINKED ISSUE #1089: systemd (and other service managers?) example files
One supported way to run TigerBeetle is directly, using systemd or similar. We should provide a `.service` file with our recommended configuration.

Perhaps for some other common service managers too. Mildly related to a larger project of providing Helm charts for TigerBeetle.

DanielSidhion:
Closes #1089.

This PR adds a systemd unit as example of how one would run TigerBeetle on a systemd system. Instructions on adapting the example to different use cases are included. An additional entry in the docs was created to bring awareness that this unit exists.

The values in the systemd unit are similar to what I'm using for a NixOS service in https://github.com/NixOS/nixpkgs/pull/279511.

### DIFF
--- a/docs/deploy/hardware.md
+++ b/docs/deploy/hardware.md
@@ -1,5 +1,5 @@
 ---
-sidebar_position: 4
+sidebar_position: 1
 ---
 
 # Hardware
--- a/docs/deploy/linux.md
+++ b/docs/deploy/linux.md
@@ -1,0 +1,11 @@
+---
+sidebar_position: 2
+---
+
+# Deploying on Linux
+
+## systemd
+
+TigerBeetle includes an example [systemd unit](https://github.com/tigerbeetle/tigerbeetle/tree/main/tools/systemd/) for use with Linux systems that use systemd.
+The unit is configured to start a [single-node cluster](../quick-start/single-binary.md), so you may need to adjust it for other cluster configurations.
+Instructions to do so are in our GitHub repository.
--- a/scripts/.cspell.json
+++ b/scripts/.cspell.json
@@ -83,6 +83,8 @@
 	"myscript",
 	"nacks",
 	"Nearline",
+	"networkd",
+	"nounset",
 	"otherstuff",
 	"paxos",
 	"POCPOU",
--- a/tools/systemd/README.md
+++ b/tools/systemd/README.md
@@ -1,0 +1,66 @@
+# systemd Service Unit
+
+This is an example of a systemd service that runs TigerBeetle.
+You will likely have to adjust it if you're deploying TigerBeetle.
+
+## Adjusting
+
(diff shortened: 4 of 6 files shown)


deploy/README.md
```diff
@@ -0,0 +1,3 @@
+# Tools for deployment and distributions
```
cb22: For now, can we move the `linux-systemd` folder to `tools/systemd` and get rid of the top level `deploy/` directory?

deploy/linux-systemd/tigerbeetle-pre-start
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
```
cb22: I don't think we need any bash specific features here - could we change to `/bin/sh`?

Also, if we rename this as `.sh`, we'll get our shellcheck repo linting to apply to it too.

deploy/linux-systemd/tigerbeetle-pre-start
```diff
@@ -0,0 +1,7 @@
...
+set -e
+set -o nounset
+
+if ! test -e ${TIGERBEETLE_DATA_FILE}; then
```
cb22: Probably best to quote all of these, in case there's a space in any of them (most likely the data file).

deploy/linux-systemd/tigerbeetle.service
```diff
@@ -0,0 +1,26 @@
+[Unit]
+Description=TigerBeetle server
```
cb22:
```suggestion
Description=TigerBeetle Replica
```

deploy/linux-systemd/tigerbeetle.service
```diff
@@ -0,0 +1,26 @@
...
+Documentation=https://docs.tigerbeetle.com/
+After=network-online.target
+Wants=network-online.target systemd-networkd-wait-online.service
+
+[Service]
```
cb22: The hardening here is great - are there any other parameters we might want to set that you know of off hand? The more locked down the better :smile:

deploy/linux-systemd/tigerbeetle-pre-start
DanielSidhion: Thanks for catching this!

deploy/linux-systemd/tigerbeetle.service
DanielSidhion: Browsing through the systemd docs I can see a few others that might be helpful, but I haven't had the time to try them yet. I think the current hardening configurations are a pretty good start, and as I come across more things, I plan on making a PR to keep this repo in sync with the NixOS service config.

cb22 (approved):

PR #1406, opened 2024-01-09
fuzz: cleanup smoke fuzzing

matklad:
- deduplicate the code
- run vsr_superblock, instead of running lsm_tree twice
- add some timing information for fuzzers

I don't know what's a reasonable time bound for smoke fuzzing. On the one hand, I want to keep it as fast as test (so, circa 300ms per fuzzer). On the one hand, the existing segmented array fuzzer takes 8 seconds not because the input is long, but because it explores hard-coded combinatorial matrix. And we'll probably be running smoke fuzzing as a separate job on CI, so spending half an hour there doesn't seem too wrong.

I decided to use 60seconds as a bond for now, and revise it upwards or downwards as necessary.

### DIFF
--- a/src/fuzz_tests.zig
+++ b/src/fuzz_tests.zig
@@ -10,6 +10,10 @@ const log = std.log.scoped(.fuzz);
 // NB: this changes values in `constants.zig`!
 pub const tigerbeetle_config = @import("config.zig").configs.test_min;
 
+pub const std_options = struct {
+    pub const log_level: std.log.Level = .info;
+};
+
 const Fuzzers = .{
     .ewah = @import("./ewah_fuzz.zig"),
     .lsm_cache_map = @import("./lsm/cache_map_fuzz.zig"),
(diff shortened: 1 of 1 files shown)


cb22:
> And we'll probably be running smoke fuzzing as a separate job on CI, so spending half an hour there doesn't seem too wrong

Smoke fuzzing, or doing some actual fuzzing? IMO smoking should be as quick as possible (agreed that 1 minute is a good starting point!) and actual fuzzing could be as long as the build process takes...

(Smoke fuzzing - I understand to mean this, we're testing with a fixed seed, limited events and the aim is to make sure there are no obvious breaks in the fuzzer / code.)

cb22 (approved):

matklad:
Smoke fuzzing! When starting to work on this PR, I believed that it indeed "should be as quick as possible" (so, 10 seconds _total_). 

But after looking at the segmented array case, I now think that maybe "as quick as possible" is not exactly right, and rather it should be "reasonably fast" (so, 10 minutes total would be OK).

I guess, ideally, we'd even have three modes:

* super quick 5 second run to check that stuff compiles locally
* medium-length 10 minutes run on CI, to find any easy, but non-trivial problems
* unbounded-in-length "I am going to fuzz you until either I or you give up"

Given that 3 is in a bit of a disarray now, I'd pick 2 over 1 (and over implementing both 1&2)

PR #1408, opened 2024-01-09
fuzz: use correct config

matklad:
We should be using fuzz_min rather than test_min here! Luckily, this was
caught by the forest fuzzer, which simply fails with too small storage
size.

With the correct config, we can now re-enable that fuzzer.

### DIFF
--- a/src/fuzz_tests.zig
+++ b/src/fuzz_tests.zig
@@ -63,10 +63,14 @@ pub fn main() !void {
             assert(cli_args.events_max == null);
             try smoke();
         },
-        inline else => |fuzzer| try @field(Fuzzers, @tagName(fuzzer)).main(.{
-            .seed = seed,
-            .events_max = cli_args.events_max,
-        }),
+        inline else => |fuzzer| {
+            var timer = try std.time.Timer.start();
+            try @field(Fuzzers, @tagName(fuzzer)).main(.{
+                .seed = seed,
+                .events_max = cli_args.events_max,
+            });
+            log.info("done in {}", .{std.fmt.fmtDuration(timer.lap())});
+        },
     }
 }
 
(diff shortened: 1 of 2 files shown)


src/fuzz_tests.zig
```diff
@@ -63,38 +67,42 @@ pub fn main() !void {
...
             .lsm_segmented_array,
             .vsr_journal_format,
             .vsr_superblock_quorums,
-            => .{ .seed = 123, .events_max = null },
+            => null,
         };
 
         var timer_single = try std.time.Timer.start();
-        try @field(Fuzzers, @tagName(fuzzer)).main(fuzz_args);
+        try @field(Fuzzers, @tagName(fuzzer)).main(.{ .seed = 123, .events_max = events_max });
```
cb22, line 89: Since we override the seed here, maybe best to move the printing of "Fuzz seed:" into the non-smoke case?

matklad: Excellent catch, done!

This reminds me that we should use that trick with "use git hash as a seed" here, will do that further up the stack of PRs!

src/fuzz_tests.zig
```diff
@@ -103,5 +95,26 @@ fn smoke() !void {
...
+        var seed_random: u64 = undefined;
+        try std.os.getrandom(std.mem.asBytes(&seed_random));
+        break :seed seed_random;
+    };
+    log.info("Fuzz seed = {}", .{seed});
+
+    var timer = try std.time.Timer.start();
+    switch (cli_args.positional.fuzzer) {
+        .smoke => unreachable,
+        inline else => |fuzzer| try @field(Fuzzers, @tagName(fuzzer)).main(
```
matklad, line 115: I tried `const fuzz_main: fn(fuzz.FuzzArgs) !void = switch(...)` version, but that doesn't work: an `!void` is an _inferred_ error union, it can't be used with function pointers.

cb22 (approved):

PR #1410, opened 2024-01-09
fuzz: smoke fuzz free_set

matklad:
NB: with 123 seed, we are hitting the 0 branch at

```
.release = if (random.boolean()) 0 else 500 * random.float(f64),
```

I am not going to fix this here, but perhaps long term smoke fuzzing
should run for a _range_ of seeds?

### DIFF
--- a/src/fuzz_tests.zig
+++ b/src/fuzz_tests.zig
@@ -74,11 +74,11 @@ fn main_smoke() !void {
             // to run them fast enough?
             .lsm_cache_map,
             .lsm_manifest_log,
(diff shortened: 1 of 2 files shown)


matklad:
>I am not going to fix this here, but perhaps long term smoke fuzzing
should run for a range of seeds?

This shold be fixed by replacing `.seed = 123` with `.seed = $(git rev-parse HEAD)`

matklad:
(in a follow-up that is)

cb22 (approved):

PR #1413, opened 2024-01-09
ci: introduce integration tests

matklad:
Integration tests for TigerBeetle. Although the term is not particularly well-defined, here
it means a specific thing:
  * the test binary itself doesn't contain any code from TigerBeetle,
  * but it has access to a pre-build `./tigerbeetle` binary.
All the testing is done through interacting with a separate tigerbeetle process.

---

* re-cast repl_integration as general integration tests
* tether them to `zig build test`
* rewrite repl test on top of TmpTigerbeetle and snaptest to avoid
  any dependency on the shell
* implement "redactions" for snaptest, to ignore parts of the output.
  For repl tests, we want to ignore timestamps

```
λ t ./zig/zig build --summary all test:integration
Build Summary: 8/8 steps succeeded; 12/12 tests passed
test:integration success
└─ run test 12 passed 7s MaxRSS:3G
   ├─ zig test Debug native-native success 3s MaxRSS:253M
   └─ install success
      ├─ install tigerbeetle success
      │  └─ zig build-exe tigerbeetle Debug native-native success 12s MaxRSS:798M
      │     ├─ options success
      │     └─ options (reused)
      └─ install generated to ../../tigerbeetle success
         ├─ zig build-exe tigerbeetle Debug native-native (+2 more reused dependencies)
         └─ zig build-exe tigerbeetle Debug native-native (+2 more reused dependencies)

real 25.37s
cpu  26.64s (23.39s user + 3.26s sys)
rss  3553.96mb
```


matklad:
>run test 12 passed 7s MaxRSS:3G

That's 12 test because it also runs the tests for stdx, snaptest, and everything else that's included. But those tests are fast, so I don't think we should try to dedupe this.

src/testing/snaptest.zig
```diff
@@ -176,6 +192,64 @@ pub const Snap = struct {
...
+            snapshot_cut_next.prefix
+        else
+            snapshot_rest;
+        assert(next_match.len > 0);
+        snapshot_rest = stdx.cut(snapshot_rest, next_match).?.suffix;
+
+        const got_cut_next = stdx.cut(got_rest, next_match) orelse return false;
+        const ignored = got_cut_next.prefix;
+        // If <snap:ignore> matched an empty string, report it as an error.
+        if (ignored.len == 0) return false;
```
sentientwaffle, line 223: Maybe also error if the ignored string contains a newline? Since that would probably be confusing and/or a mistake.

matklad: Good idea! I had the same one, but forgot to act on it :rofl:

sentientwaffle (approved):

PR #1414, opened 2024-01-09
LSM: Cleanup of binary_search_values

sm2n:
Switch the return type to an optional, so that we can centralize all the asserts in one place. Previously we had different asserts at different callsites for no real reason.

This is less expressive, but it turns out we were not actaully using the index in the inexact case at any callsites.

`?*const Value` and `BinarySearchResult` have the same size so there shouldn't be performance issues down the line.


matklad (approved):
Makes sense to me!

I think we need to re-visit this todo once we reach consensuse on the semantics of constants.verify, but let's do that once joran return from his sweet summer break!

PR #1415, opened 2024-01-10
io/linux: Add support for using block devices as datafiles

sm2n:
Tested on my local machine.

Currently, this does not check that `tigerbeetle format` is not overwriting an existing datafile.

I'm not entirely sure what the semantics for `tigerbeetle format` on block devices should be.

Some thoughts:
- a confirmation dialog could be good? but not good for scripting, though the standard solution there is to add another command line argument (but then we have to rework the interface which is maybe worth it)
- it's hard to detect whether we're overwriting stuff at all, since I don't think we have a tigerbeetle datafile magic

### DIFF
--- a/src/io/linux.zig
+++ b/src/io/linux.zig
@@ -998,12 +998,39 @@ pub const IO = struct {
     ) !os.fd_t {
         assert(relative_path.len > 0);
         assert(size % constants.sector_size == 0);
-
-        // TODO Use O_EXCL when opening as a block device to obtain a mandatory exclusive lock.
-        // This is much stronger than an advisory exclusive lock, and is required on some platforms.
+        // Be careful with openat(2): "If pathname is absolute, then dirfd is ignored." (man page)
+        assert(!std.fs.path.isAbsolute(relative_path));
 
         var flags: u32 = os.O.CLOEXEC | os.O.RDWR | os.O.DSYNC;
         var mode: os.mode_t = 0;
+
+        const kind: enum { file, block_device } = blk: {
+            const stat = os.fstatat(
+                dir_fd,
+                relative_path,
+                0,
+            ) catch |err| switch (err) {
+                error.FileNotFound => {
+                    if (method == .create or method == .create_or_open) {
+                        // It's impossible to distinguish creating a new file and opening a new block
+                        // device with the current API. So if it's possible that we should create a file
+                        // we try that instead of failing here.
+                        break :blk .file;
+                    } else {
+                        @panic("Block device not found.");
+                    }
+                },
+                else => |err_| return err_,
+            };
+            if (os.S.ISBLK(stat.mode)) {
+                break :blk .block_device;
+            } else {
+                if (!os.S.ISREG(stat.mode)) {
+                    @panic("Datafile path does not point to block device or regular file.");
+                }
+                break :blk .file;
+            }
+        };
 
         // This is not strictly necessary on 64bit systems but it's harmless.
         // This will avoid errors with handling large files on certain configurations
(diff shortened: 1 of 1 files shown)


cb22:
> it's hard to detect whether we're overwriting stuff at all, since I don't think we have a tigerbeetle datafile magic

This worries me a bit - things like someone accidentally overwriting their filesystem (or TigerBeetle datafile) by specifying the wrong data path.

Maybe a good enough solution would be to check that the first X MB are zero'd? That way, we don't need our own magic, and we don't need an override (if they want to script it, they can use `dd` as part of the same script too).

cb22:
@sm2n after chatting a bit more with Joran, if we're on a block device, we can check that the first `superblock_zone_size` of a device is zero'd:

- It'll ensure that the block device is not directly TigerBeetle.
- It'll be very likely to catch any cases where there's an existing other filesystem.
- In the case of there being a partition table (eg, two partitions, one starting at 0MB, one at 1024MB) and the operator tries to format the raw disk (/dev/sda) while a partition later is TigerBeetle (/dev/sda2) it'll be blocked by the MBR/GPT existing.

src/io/linux.zig
```diff
@@ -992,69 +992,117 @@ pub const IO = struct {
...
 
+        const file_kind: enum { file, block_device } = blk: {
+            const stat = os.fstatat(
+                dir_fd,
+                relative_path,
+                0,
+            ) catch |err| switch (err) {
+                error.FileNotFound => {
+                    assert(method == .create or method == .create_or_open);
+                    // We must be trying to create a new file, which is not possible for block devices
```
jorangreef: It seems like this classifier is flaky—if the path to the block device is not found, would this error not also be raised?

src/io/linux.zig
```diff
@@ -992,69 +992,117 @@ pub const IO = struct {
...
+                    break :blk .file;
+                },
+                else => |err_| return err_,
+            };
+            if (os.S.ISBLK(stat.mode)) {
+                break :blk .block_device;
+            } else {
```
jorangreef, line 1027: What else do we expect here? What would be reasonable characteristics? What would be unreasonable or surprising? Is there anything we can assert if it's a file?

src/io/linux.zig
```diff
@@ -992,69 +992,117 @@ pub const IO = struct {
...
-
-        // TODO Use O_EXCL when opening as a block device to obtain a mandatory exclusive lock.
-        // This is much stronger than an advisory exclusive lock, and is required on some platforms.
+        // Be careful with openat(2): "If pathname is absolute, then dirfd is ignored." (man page)
+        assert(!std.fs.path.isAbsolute(relative_path));
 
         var flags: u32 = os.O.CLOEXEC | os.O.RDWR | os.O.DSYNC;
         var mode: os.mode_t = 0;
 
+        const file_kind: enum { file, block_device } = blk: {
```
jorangreef:
```suggestion
        const kind: enum { file, block_device } = blk: {
```

src/io/linux.zig
```diff
@@ -992,69 +992,117 @@ pub const IO = struct {
...
-        var direct_io_supported = false;
-        var dir_on_tmpfs = try fs_is_tmpfs(dir_fd);
-
-        if (dir_on_tmpfs) {
-            log.warn("tmpfs is not durable, and your data will be lost on reboot", .{});
-        }
+        switch (file_kind) {
+            .block_device => {
+                if (constants.direct_io) {
+                    // Block devices should always support direct io
```
jorangreef: Per TigerStyle, comments are sentences with punctuation and proper capitalization.
```suggestion
                    // Block devices should always support Direct IO.
```

src/io/linux.zig
```diff
@@ -992,69 +992,117 @@ pub const IO = struct {
...
+                    flags |= os.O.DIRECT;
+                    // Use O_EXCL when opening as a block device to obtain an advisory exclusive lock.
+                    // Normally you can't do this for files you don't create, but for block devices this:
```
jorangreef:
```suggestion
                    // Normally, you can't do this for files you don't create, but for block devices this:
```

src/io/linux.zig
```diff
@@ -992,69 +992,117 @@ pub const IO = struct {
...
+                    //     - guarantees there are no mounts using this block device
```
jorangreef:
```suggestion
                    //     - guarantees that there are no mounts using this block device
```

Consistent use of "that" after "guarantees".

jorangreef: Let's move "guarantees" up as a shared qualifier before the opening ":".

And then move the "However," and "This" into a separate paragraph, since those bullets don't actually read correctly with the opening "for block devices this:".

src/io/linux.zig
```diff
@@ -992,69 +992,117 @@ pub const IO = struct {
...
+                    //     - guarantees that no new mounts can use this block device while we have it open
+                    //     - however it doesn't prevent other processes with root from opening without
+                    //       O_EXCL and writing (mount is just a special case that always checks O_EXCL)
+                    //     - this should be stronger than flock(2) locks, which work on a separate system
+                    // The relevant kernel code is here:
+                    // <https://github.com/torvalds/linux/blob/7da71072e1d6967c0482abcbb5991ffb5953fdf2/block/bdev.c#L932>
```
jorangreef: Nice. Add the kernel version to the comment as well, so that people in future know how dated the git link is.

src/io/linux.zig
```diff
@@ -992,69 +992,117 @@ pub const IO = struct {
...
-        // Be careful with openat(2): "If pathname is absolute, then dirfd is ignored." (man page)
-        assert(!std.fs.path.isAbsolute(relative_path));
         const fd = try os.openat(dir_fd, relative_path, flags, mode);
         // TODO Return a proper error message when the path exists or does not exist (init/start).
         errdefer os.close(fd);
 
-        // TODO Check that the file is actually a file.
+        {
+            const stat = try os.fstat(fd);
+            // the fd should either refer to a regular file or a block device
```
jorangreef:
```suggestion
            // The file descriptor should refer to a regular file or block device.
```

src/io/linux.zig
```diff
@@ -1098,8 +1146,39 @@ pub const IO = struct {
         // We always do this when opening because we don't know if this was done before crashing.
         try os.fsync(dir_fd);
 
-        const stat = try os.fstat(fd);
-        if (stat.size < size) @panic("data file inode size was truncated or corrupted");
+        switch (file_kind) {
+            .file => {
+                const stat = try os.fstat(fd);
+                if (stat.size < size) @panic("data file inode size was truncated or corrupted");
```
jorangreef: Can the `os.fstat` be inlined (to get rid of introducing the `stat` local variable, which is only used once)?

src/io/linux.zig
```diff
@@ -1098,8 +1146,39 @@ pub const IO = struct {
...
+            },
+            .block_device => {
+                const BLKGETSIZE64 = os.linux.IOCTL.IOR(0x12, 114, usize);
+                var block_device_bytes_size: usize = 0;
```
jorangreef:
```suggestion
                var block_device_size: usize = 0;
```

In TigerBeetle, the unit for `size` should always be bytes.

src/io/linux.zig
```diff
@@ -1098,8 +1146,39 @@ pub const IO = struct {
...
+                switch (std.os.errno(os.linux.ioctl(
+                    fd,
+                    BLKGETSIZE64,
+                    @intFromPtr(&block_device_bytes_size),
+                ))) {
+                    .SUCCESS => {},
+
+                    .FAULT => unreachable,
+                    .BADF => unreachable,
+                    .NOTTY => unreachable,
```
jorangreef: It seems like these should be explicit panics (with a user friendly error message).

I know Zig stdlib sometimes tries to "unreachable" away kernel errors but it's possible for the kernel to overload these in future. For example, IIRC Zig used to use unreachable for EINVAL sometimes, but this can be returned where a system doesn't support Direct I/O.

What other errors should be handled?

src/io/linux.zig
```diff
@@ -1098,8 +1146,39 @@ pub const IO = struct {
...
+                    else => |err| return os.unexpectedErrno(err),
+                }
+
+                if (block_device_bytes_size < size) {
+                    std.debug.panic(
+                        "Block device too small! Wanted at least {}, got {}",
```
jorangreef: This error message can be improved. For example, to be phrased as a single complete sentence, also with full stop, no exclamation, and using parentheses to describe the two difference sizes.

jorangreef (changes requested):
Great to see you adding this, Mukarram.

src/io/linux.zig
```diff
@@ -992,69 +992,117 @@ pub const IO = struct {
     ) !os.fd_t {
         assert(relative_path.len > 0);
         assert(size % constants.sector_size == 0);
```
jorangreef, line 1000: Can we assert that `size` is never 0?

src/io/linux.zig
sm2n: What do you think of the current one?

src/io/linux.zig
sm2n: I changed these to return the error. All of the possible errors from this ioctl are for checking if the arguments given are valid (I checked the kernel docs and also the source).

In this case, we're providing the arguments directly, without passing in any outside input, so it should always be valid. So it should be treated as an assertion failure if there is an error returned. I think it's reasonable to return an error return trace in that case, what do you think?

src/io/linux.zig
sm2n: I added an assert that the stat gives a regular file if it isn't a block device. I think that's about all we can do here.

src/io/linux.zig
sm2n: That's a good catch, I added a comment and a branch for this, let me know what you think.

sm2n:
I like the solution of checking if the superblock is empty. I ended up pasting @cb22's summary of the solution into the code, since I thought it was well written.

Now the code does actually check if the superblock is empty.

sm2n:
One thing I was wondering is about the dependency of `IO` on `VSR`, I think pulling in `superblock_zone_size` here is basically the only logical dependency. Do we want the code to be organized that way? I think the alternative would be putting the check logic in `main.zig`, which seems worse to me, but I want to raise the issue.

src/io/linux.zig
sm2n: It appears we can't, since the AOF uses a size of 0.

sm2n: hmmm, that means that the file might not even be `superblock_zone_size` large...

sm2n: Oh, the real issue here is that the AOF doesn't support block devices.

src/io/linux.zig
matklad: might use git tag instead of commit hash even, to also make the url shorter at the same time.

matklad:
@sm2n we've just rewritten our git history:

https://github.com/tigerbeetle/tigerbeetle/commit/d7c3f4654ea7c65b6d141be33dadd29e869c3984

Please rebase the changes on top of new main, and sorry for the inconvenience!

sm2n:
No big deal, 30s in interactive rebase did it.

src/io/linux.zig
```diff
@@ -1098,8 +1157,66 @@ pub const IO = struct {
...
+                if (method == .create or method == .create_or_open) {
+                    // Check that the first superblock_zone_size bytes are 0.
+                    // - It'll ensure that the block device is not directly TigerBeetle.
+                    // - It'll be very likely to catch any cases where there's an existing
+                    //   other filesystem.
+                    // - In the case of there being a partition table (eg, two partitions,
+                    //   one starting at 0MB, one at 1024MB) and the operator tries to format
+                    //   the raw disk (/dev/sda) while a partition later is
+                    //   TigerBeetle (/dev/sda2) it'll be blocked by the MBR/GPT existing.
+                    const superblock_zone_size = @import("../vsr/superblock.zig").superblock_zone_size;
```
cb22: Nit: line length (and the below `@panic`)

cb22: (this file is currently excluded from `tidy.zig` which enforces that check)

src/io/linux.zig
```diff
@@ -1098,8 +1157,66 @@ pub const IO = struct {
...
+                    var read_buf: [superblock_zone_size]u8 align(constants.sector_size) = undefined;
+
+                    // We can do this without worrying about retrying partial reads because on linux,
+                    // read(2) on block devices can not be interrupted by signals. See signal(7).
+                    assert(superblock_zone_size == try os.read(fd, &read_buf));
+                    if (!std.mem.allEqual(u8, &read_buf, 0)) {
+                        @panic(
+                            "Superblock on block device not empty. If this is the correct block device to use, please format it first.",
```
cb22: Could we make this print out the `dd` command to zero the device? Alternatively, if we don't want to risk that, perhaps change the wording to something like

```suggestion
                            "Superblock on block device not empty. If this is the correct block device to use, please zero the first {}MB using a tool like dd.",
```

(since format might be confusing since the user is running `tigerbeetle format`, and formatting it with eg `mkfs` won't work either)

cb22 (approved):

jorangreef (approved):

matklad:
Let’s merge it, to gauge our benchmark variance before dropping balance indexes!

PR #1417, opened 2024-01-10
clients: overhaul docs generation

matklad:
TL;DR: turn the logic for generating client readmes upside down --- rather than assambling a readme sample out of code fragments from docs.zig, write the entire readme sample as a proper sample in the samples directory, and inject fragments of that sample into the readme.

Motivation: right now, we have two completely different code paths for integration-testing our client samples:

* In the [ci scripts](https://github.com/tigerbeetle/tigerbeetle/blob/0f700cb0f82d607ea98dfa1fd381417a17cf5cde/src/clients/java/ci.zig#L21-L32), we run every sample for every client. This is triggered by `clients.yaml` on CI.
* In [docs_generate](https://github.com/tigerbeetle/tigerbeetle/blob/0f700cb0f82d607ea98dfa1fd381417a17cf5cde/src/clients/docs_generate.zig#L870-L887) we asseble a sample from our readme fragments and test that. This **used to be** tested by our CI, but got semi-intentionally dropped during the latest overhaul.

This PR reconciles the two approaches, by making sure that the code from README is just a sample. This also has auxilary benefits:

* we now can reformat the code from the samples. This is not automated, but I did a run through with a formatter manually.
* Now the code from readme can be conveniently debugged, as it exist on the file system
* Future work: with some effort, I think the walkthrough sample might become a reasonable sample for new users to directly look at
* Future work: now it is possible to add more thourghou checks to the sample, which are not necessary shown in the readme, but could test that the behavior is what we expect (as written, our samples return a ton of errors like "account id should not be zero").

I tried to change as little as possible, as the diff here is large anyway. I do plan to send a significant cleanup PR as a follow up: for example, I believe the entirety of shutil.zig and run_with_tb.zig is now dead code. 

That said, there's a bunch of user-visible changes here already:

* Code in readme is now formatted! E.g., .Net now puts `{` on a line by itself. 
* From java, and .Net, I dropped "don't forget to close the client" in favor of just showning the `try-with-resources` version, I don't think we should advertise bad practicies, and it was simpler to just drop this altogether, rather than add it to walkthroughs. 
* Before, we were also validating that "shell" bits of readme are correct. Thigs like `npm install tigerbeetle-node` and `node main.js`. Now, these are not tested. I think this is fine --- these are upstream stable interfaces which are not going to change. We _might_ have some bugs there, but the cost of such bug feels small, and they are unlikecly to reappear after they are repaired.


src/clients/go/samples/walkthrough/README.md
```diff
@@ -0,0 +1 @@
+Code from README collected into a single runnable project.
```
sentientwaffle: Maybe link to the README that is being referred to (since this file is _also_ a README).

matklad:
The end state I am aiming at with this stack of PRs:

https://github.com/tigerbeetle/tigerbeetle/pull/1422

src/clients/docs_generate.zig
```diff
@@ -409,6 +360,47 @@ const Generator = struct {
...
     }
 
+    fn code_section(self: Generator, mw: *MarkdownWriter, section_name: []const u8) void {
+        var code = std.ArrayList(u8).init(self.arena.allocator());
+
+        const section_start = self.sprintf("section:{s}\n", .{section_name});
+        const section_end = self.sprintf("endsection:{s}\n", .{section_name});
+
+        var text = self.walkthrough;
+        for (0..10) |_| {
```
sentientwaffle, line 375: If I'm reading this right, if there are multiple sections with the same name, then  this concatenates them all together, is that right?

src/clients/go/samples/walkthrough/README.md
matklad: >since this file is also a README

:facepalm:

src/clients/docs_generate.zig
matklad: Correct! This is used for java, which, for the first, imports example, needs to concatenate `public static void main(String[] args) {` with the `}` at the end of file. 

added a comment!

sentientwaffle (approved):

matklad:
:weary: docker:

>Unable to find image 'alpine:latest' locally
docker: Error response from daemon: Head "https://registry-1.docker.io/v2/library/alpine/manifests/latest": Get "https://auth.docker.io/token?account=githubactions&scope=repository%3Alibrary%2Falpine%3Apull&service=registry.docker.io": EOF.
See 'docker run --help'.

Amusingly, googling this error turns up some results from January _last_ year!

matklad:
Ok, for some reason, re-running the job didn't work, _but_ rebasing the whole branch did the trick.

matklad:
Wait, somehow, after a rebase, the review is not dismissed? :thinking: this seems like a new behavior...

sentientwaffle:
> Wait, somehow, after a rebase, the review is not dismissed? 🤔 this seems like a new behavior...

I noticed that too! I wonder what the heuristic is for whether or not the approval is retained.
https://github.com/orgs/community/discussions/12876#discussioncomment-7445850 is the only info I've found.

PR #1422, opened 2024-01-12
 clients: bring back testing of READMEs on CI

matklad:
Move logic from docs_generate utility into `scripts -- ci` workflow, so
that it gets executed by CI naturally. Along the way, cleanup code a
bit:

* use `shell.zig` rather than `shutil.zig`, remove the latter.
* merge `Generator` and `Markdown` writer into a single `Context`
  struct. There might be some finer grained factoring here, but we don't
  use markdown-writing capabilities elsewhere, so a single grab bag of
  stuff feels simpler.
* pull the overall control flow for reading&updating files into a
  top-level function.
* remove helper functions for generating common parts between main &
  sample readmes --- the "common" parts differ more than they are the
  same, it's more readable to copy the relevant paragraphs twice
* use code blocks to delimit markdown sections. This surfaces one wrong
  header even!


src/scripts/ci.zig
```diff
@@ -31,10 +33,22 @@ pub fn main(shell: *Shell, gpa: std.mem.Allocator, cli_args: CliArgs) !void {
     if (cli_args.validate_release) {
```
sentientwaffle, line 33: Out of scope for this PR, but right now we have a `--language` filter -- I wonder if it would make more sense to have a `--task=<tests|readmes>` (or replace `--language` with a more general `--filter`) so that you can generate the READMEs without running all of the client tests.

sentientwaffle (approved):

src/scripts/ci.zig
matklad: Yeah, the current filtering is a bit odd indeed. But my general rule of thumb for these sorts of dev-tools is to do pain-driven development --- send a specific PR for a specific feature once you need it for a particular workflow. These things _shouldn't_ be costly to change, so they could use less up-front design. 

I often find that just commenting specific stuff out manually works better than trying to recall elaborate filtering syntax >_<

PR #1423, opened 2024-01-12
java: don't use checked exceptions for our own programming errors 

matklad:
RequestException is thrown when the underlying C library detects that
the request is invalid --- the op or the size of message body is wrong.

However, the public API of of our Java client ensures that these are
always correct: it is impossible to trigger these errors unless:

* there's a bug in our C or Java code
* there's a version mismatch between our Java and C code (which is
  unlikely, as they come from the same jar)

So, there's absolutely no reason to tax the caller with declaring and
handling this exception.

### DIFF
--- a/src/clients/java/src/main/java/com/tigerbeetle/BlockingRequest.java
+++ b/src/clients/java/src/main/java/com/tigerbeetle/BlockingRequest.java
@@ -85,7 +85,7 @@ final class BlockingRequest<TResponse extends Batch> extends Request<TResponse> 
         return result != null || exception != null;
     }
 
-    public TResponse waitForResult() throws RequestException {
+    public TResponse waitForResult() {
 
         waitForCompletionUninterruptibly();
         return getResult();
@@ -143,7 +143,7 @@ final class BlockingRequest<TResponse extends Batch> extends Request<TResponse> 
         }
     }
 
-    TResponse getResult() throws RequestException {
+    TResponse getResult() {
 
         assertTrue(result != null || exception != null, "Unexpected request result: result=null");
 
--- a/src/clients/java/src/main/java/com/tigerbeetle/RequestException.java
+++ b/src/clients/java/src/main/java/com/tigerbeetle/RequestException.java
@@ -1,6 +1,12 @@
 package com.tigerbeetle;
 
-public final class RequestException extends Exception {
+/**
+ * RequestException is thrown when the internal invariants of the underlying native library are
+ * violated, which is expected to never happen. If tis exception is thrown, then either there is a
+ * programming error in the tigerbeetle-java library itself, or there is a version mismatch between
+ * the java code and the underlying native library.
+ */
+public final class RequestException extends RuntimeException {
 
     private final byte status;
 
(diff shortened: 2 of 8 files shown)


src/clients/java/src/main/java/com/tigerbeetle/RequestException.java
```diff
@@ -22,9 +28,9 @@ public String toString() {
         if (status == PacketStatus.TooMuchData.value)
             return "Too much data provided on this batch.";
         else if (status == PacketStatus.InvalidOperation.value)
-            return "Invalid operation. Check if this client is compatible with the server's version.";
+            return "Invalid operation.";
         else if (status == PacketStatus.InvalidDataSize.value)
-            return "Invalid data size. Check if this client is compatible with the server's version.";
```
matklad, line 27: As far as I understand, this is just wrong.

src/clients/java/src/main/java/com/tigerbeetle/BlockingRequest.java
```diff
@@ -143,7 +143,7 @@ private void waitForCompletionUninterruptibly() {
         }
     }
 
-    TResponse getResult() throws RequestException {
+    TResponse getResult() {
 
         assertTrue(result != null || exception != null, "Unexpected request result: result=null");
```
matklad, line 148: I am wondering... Do we actually need this exception as a separate class even? From that

>Check if this client is compatible with the server's version.

It feels like the original intention behind the exception was to signal error response from the server, and then it mutated to signal local programming errors. In this new role, perhaps we can just throw an `AssertionError` for non-ok packet statuses?

batiati: The original intention was to map result codes into Java exceptions: `InitializationException` for `tb_client_init`, `RequestException` for `packet.status`, and `ConcurrencyExceededException` for `tb_client_acquire_packet`.

> Check if this client is compatible with the server's version.

This message is indeed wrong, we never planned to return error response from the server at the protocol level (e.g. invalid size or operation), I think the C# client uses the same string.

(...)

I completely agree with `RequestException extends RuntimeException`, but I don't have a strong opinion on using `AssertionError` for it (feels like we are losing information if we reduce the packet status to an assertion error message).

batiati (approved):

matklad:
>I don't have a strong opinion on using AssertionError for it (feels like we are losing information if we reduce the packet status to an assertion error message).

Yeah, same here. The argument "for" removing the dedicated exception is reducing the public API surface area by one class.

I guess:

* it's a small mistake either way
* while things are still in flux, let's use a dedicated exception
* when we start giving backwards compatability guarantees about the API, I'd maybe feel moderately strongly about not exposing this exception, but we are not at that point yet.

batiati:
I agree. Let's wait for the client refactoring King is working on. I think `ConcurrencyExceededException` will be dropped as well.

> Yeah, same here. The argument "for" removing the dedicated exception is reducing the public API surface area by one class.

Additionally:

- If we were implementing the Java Client 100% in pure Java, `RequestException` wouldn't exist.

- `AssertionError` is used when the Java side has a bug and any other result from the native library is mapped by a specialized exception. In this case, the kind of results `RequestException` maps aren't expected to happen unless there's a bug on the Java side.

PR #1427, opened 2024-01-15
ci: actually use release branch for release validation

matklad:
Commits are unrelated, the first fixes rev->ref typo, the second one updates version of the action.

### DIFF
--- a/.github/workflows/clients.yml
+++ b/.github/workflows/clients.yml
@@ -22,7 +22,7 @@ jobs:
       # could be done.
       contents: read
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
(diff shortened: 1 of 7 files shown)


cb22:
(the rebase was to get actions to run, after adding v4 to our allowlist).

@matklad - once this has merged, could you give the list at https://github.com/tigerbeetle/tigerbeetle/settings/actions a once over and remove anything we no longer use?

cb22 (approved):

matklad:
I've changed that to just:

```
actions/*
```

* we already use GitHub extensively, so it makes sense to just blanket trust all actions maintained by GitHub itself
* we don't use, and probably shoundn't use, any third party actions, to avoid increasing trust boundary, and to reduce vendor lock in on GHA.

PR #1435, opened 2024-01-17
ci: no large blobs are recorded by the Git history

matklad:


src/tidy.zig
```diff
@@ -213,6 +215,50 @@ test "tidy naughty list" {
...
+    while (lines.next()) |line| {
+        // Parsing lines like
+        //     blob 1032 client/package.json
+        var blob = stdx.cut_prefix(line, "blob ") orelse continue;
+
+        var cut = stdx.cut(blob, " ").?;
+        const size = try std.fmt.parseInt(u64, cut.prefix, 10);
+        const path = cut.suffix;
+
+        if (std.mem.eql(u8, path, "src/vsr/replica.zig")) continue; // :-)
```
cb22, line 252: I'm not sure I quite follow why replica.zig and package-lock.json are here - aren't they both less than 4MB?

matklad: `@divExact(MiB, 4)` is a quarter of, not four megs

cb22: Ahhh. My brain just did not want to parse that!

cb22 (approved):

PR #1438, opened 2024-01-17
VSR: CheckpointTrailer in grid blocks

sentientwaffle:
Instead of encoding/decoding the trailer into/from a `buffer`, encode/decode it directly to/from grid blocks.

Benefits:
- One less copy (of all trailer data) during every checkpoint.
- Once we have a grid pool, blocks can be acquired lazily, and when not used (which is most of the time, since the free set is _usually_ very small) they contribute to the grid cache.

To keep implementation simple (avoiding modifications to `ClientSessions.{encode,decode}`), one assumption this makes is that the `ClientSessions` trailer fits in a single grid block. (The ClientSessions trailer is fixed-size.) This was already true for all of our configurations.

### DIFF
--- a/src/ewah.zig
+++ b/src/ewah.zig
@@ -61,98 +61,149 @@ pub fn ewah(comptime Word: type) type {
             return @as(Word, @bitCast(mark));
         }
 
+        pub const Decoder = struct {
+            target_words: []Word,
+            target_index: usize = 0,
+
+            /// Returns the number of *words* written to `target_words` by this invocation.
+            // TODO Refactor to return an error when `source` is invalid,
+            // so that we can test invalid encodings.
+            pub fn decode(decoder_: *Decoder, source: []align(@alignOf(Word)) const u8) usize {
+                assert(source.len % @sizeOf(Word) == 0);
+
+                const target_words = decoder_.target_words;
+                assert(disjoint_slices(u8, Word, source, target_words));
+
+                const source_words = mem.bytesAsSlice(Word, source);
+                var source_index: usize = 0;
+                var target_index: usize = decoder_.target_index;
+                while (source_index < source_words.len) {
+                    const marker: *const Marker = @ptrCast(&source_words[source_index]);
+                    source_index += 1;
+                    @memset(
+                        target_words[target_index..][0..marker.uniform_word_count],
+                        if (marker.uniform_bit == 1) ~@as(Word, 0) else 0,
+                    );
+                    target_index += marker.uniform_word_count;
+                    stdx.copy_disjoint(
+                        .exact,
+                        Word,
+                        target_words[target_index..][0..marker.literal_word_count],
+                        source_words[source_index..][0..marker.literal_word_count],
+                    );
+                    source_index += marker.literal_word_count;
+                    target_index += marker.literal_word_count;
+                }
+                assert(source_index == source_words.len);
+                assert(target_index <= target_words.len);
+
+                defer decoder_.target_index = target_index;
+                return target_index - decoder_.target_index;
+            }
+        };
+
+        pub fn decoder(target_words: []Word) Decoder {
+            return .{ .target_words = target_words };
+        }
+
+        // (This is a helper for testing only.)
         /// Decodes the compressed bitset in `source` into `target_words`.
         /// Returns the number of *words* written to `target_words`.
-        // TODO Refactor to return an error when `source` is invalid,
-        // so that we can test invalid encodings.
         pub fn decode(source: []align(@alignOf(Word)) const u8, target_words: []Word) usize {
             assert(source.len % @sizeOf(Word) == 0);
             assert(disjoint_slices(u8, Word, source, target_words));
 
-            const source_words = mem.bytesAsSlice(Word, source);
-            var source_index: usize = 0;
-            var target_index: usize = 0;
-            while (source_index < source_words.len) {
-                const marker: *const Marker = @ptrCast(&source_words[source_index]);
-                source_index += 1;
-                @memset(
-                    target_words[target_index..][0..marker.uniform_word_count],
-                    if (marker.uniform_bit == 1) ~@as(Word, 0) else 0,
-                );
-                target_index += marker.uniform_word_count;
-                stdx.copy_disjoint(
-                    .exact,
-                    Word,
-                    target_words[target_index..][0..marker.literal_word_count],
-                    source_words[source_index..][0..marker.literal_word_count],
-                );
-                source_index += marker.literal_word_count;
-                target_index += marker.literal_word_count;
-            }
-            assert(source_index == source_words.len);
-            assert(target_index <= target_words.len);
-            return target_index;
+            var decoder_ = decoder(target_words);
+            return decoder_.decode(source);
         }
 
+        pub const Encoder = struct {
+            source_words: []const Word,
+            source_index: usize = 0,
+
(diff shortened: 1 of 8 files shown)


src/ewah.zig
```diff
@@ -61,98 +64,208 @@ pub fn ewah(comptime Word: type) type {
...
+
+        pub const Encoder = struct {
+            source_words: []const Word,
+            source_index: usize = 0,
+            /// The number of literals left over from the previous encode() call that still need to
+            /// be copied.
+            literal_word_count: usize = 0,
+
+            /// Returns the number of bytes written to `target_chunk` by this invocation.
+            pub fn encode(encoder_: *Encoder, target_chunk: []align(@alignOf(Word)) u8) usize {
```
sentientwaffle: The reason to have `encode` take a single chunk -- rather than an array of chunks that can be filled in sequence by a single function call -- is that it will let the `CheckpointTrailer` acquire blocks from the grid pool as-needed rather than acquiring the worst-case number. (Leading to fewer evictions from the grid cache.)

src/ewah.zig
```diff
@@ -61,98 +64,208 @@ pub fn ewah(comptime Word: type) type {
             return @as(Word, @bitCast(mark));
         }
 
+        pub const Decoder = struct {
+            target_words: []Word,
+            target_index: usize = 0,
+            source_literal_words: usize = 0,
```
matklad, line 73: One thing that gives me pause here is that there isn't any safety checks to ensure that the entire
thing was decoded. Basically, there's no difference between a Decoder which is driven to completion,
and a Decoder that missed some chunks (or processed a chunk twice).

Perhaps it makes sense to add expected total length and a `.finish()` method just for assertions?

Ah, curious how that problem doesn't exist for Encoder! Yeah, I think we should add `Decoder.done`,
so that the caller can `assert(decoder.done())`.

src/ewah.zig
```diff
@@ -61,98 +64,208 @@
...
+
+            /// Returns the number of *words* written to `target_words` by this invocation.
+            // TODO Refactor to return an error when `source_chunk` is invalid,
+            // so that we can test invalid encodings.
+            pub fn decode(
+                decoder_: *Decoder,
```
matklad: Let's try harder to avoid `thing_` names, as that goes against our intended style. Here, I thnik we
could have

```zig
Ewah.decode_all
Ewah.decode_chunks
```

to simultaneously free up decoder name, and also emphasize the connection between two decoding
methods.

And maybe call _this_ method `fn decode_chunk`?

src/ewah.zig
```diff
@@ -61,98 +64,208 @@
...
+                    );
+                    source_index += literal_word_count_chunk;
+                    target_index += literal_word_count_chunk;
+                    decoder_.source_literal_words =
+                        marker.literal_word_count - literal_word_count_chunk;
+                }
+                assert(source_index <= source_words.len);
+                assert(target_index <= target_words.len);
+
+                defer decoder_.target_index = target_index;
```
matklad: Nifty! But maybe move this defer right to the `var target_index` declaration, so that the reader
immediately knows the final destination of that local variable?

src/vsr/checkpoint_trailer.zig
```diff
@@ -134,18 +129,19 @@ pub fn CheckpointTrailerType(comptime Storage: type) type {
...
-                .buffer = buffer,
+                .blocks = blocks,
+                .block_bodies = block_bodies,
                 .block_addresses = block_addresses,
                 .block_checksums = block_checksums,
             };
         }
 
         pub fn deinit(trailer: *Self, allocator: mem.Allocator) void {
+            for (trailer.blocks) |block| allocator.free(block);
```
matklad: Move this right before `allocator.free(trailer.blocks);` to keep reverse allocation order.

src/vsr/checkpoint_trailer.zig
```diff
@@ -167,6 +163,24 @@
...
+            for (
+                trailer.block_bodies[0..chunk_count],
+                trailer.blocks[0..chunk_count],
+                0..,
+            ) |*block_body, block, block_index| {
+                const chunk_size = Chunk.size(.{
+                    .block_index = @intCast(block_index),
+                    .block_count = chunk_count,
+                    .trailer_size = trailer_size,
+                });
```
matklad, line 187: I am wondering whether it makes sense to try to slice this by `chunk_size`, rather than by
statically known `chunk_size_max` --- the resulting chunks are only "true" during decoding. When
encoding, we conservatively size chunks, and the last one might not be full.

No strong feelings either way!

Another thing here is that, ideally, we'd compute `StreamChecksum` as we are filling in the chunks,
rather than as a separate path afterwards, so that we don't do two passes over the same data.
Though, given that we have one chunk anyway, not that it matters :)

src/vsr/grid.zig
```diff
@@ -338,8 +338,21 @@ pub fn GridType(comptime Storage: type) type {
...
+                );
+
+                grid.free_set_checkpoint.size = 0;
+                for (free_set_chunks) |chunk| {
+                    grid.free_set_checkpoint.size +=
+                        @as(u32, @intCast(free_set_encoder.encode(chunk)));
+
+                    if (free_set_encoder.done()) break;
+                }
+                assert(free_set_encoder.done());
```
matklad: Alternatively, `else unreachage`, as we have a defer above anyway

src/vsr/checkpoint_trailer.zig
sentientwaffle: > I am wondering whether it makes sense to try to slice this by chunk_size, rather than by
> statically known chunk_size_max --- the resulting chunks are only "true" during decoding. When
> encoding, we conservatively size chunks, and the last one might not be full.

I don't understand what you mean by this. :thinking:

matklad: Yeah, sorry, let me rephrase!

What I mean is that the code _looks_ like it slices the blocks into the chunks we actually are going to use. So, for every block, we return not just `block[@sizeOf(vsr.Header)..block_size_max]`, but cap that to the exact size computed by the `Chunk`.


But that's not actually true during encoding! Chunks are allocated conservatively, so encoder might not actually be able to use up all of the chunks. So, for the encoding step, `block_body.* = block[@sizeOf(vsr.Header)..][0..chunk_size];` is actually dead code --- the encoder disregards `chunk_size` info, and instead stops when it itself thinks it should to stop. 

This seems a bit confusing to me. Specifically, the fact that, after encoding, some memory in `chunks` might actually be uninitialized. 

The alternative here is to say that `chunks` doesn't even attempt to compute which fraction of each block is used, and returns a `[]const [chunk_size_max]u8`, leaving it up to the caller to track the actual usage per chunk.

sentientwaffle: Aha, that makes sense, thank you!

I think I'll split `.chunks()` into two different functions, `encode_chunks` and `decode_chunks`. Then neither need a `size` passed in -- `encode_chunks` returns an array of all max-size chunks, and `decode_chunks` uses `trailer.size` to truncate the last chunk. And `decode_chunks` is only used for reading existing chunks, so the returned slice can be `const`.

matklad: Ah, brilliant, I was only noticing something fishy here, but indeed it looks like what we have here is just two different function!

matklad (approved):

PR #1439, opened 2024-01-18
storage: avoid buffer bleed when reading past the end of the file

matklad:
`bytes_read == 0` is an EOF condition. If we failed to read anything, better to zero out the buffer explicitly, like we do for other failure cases, so that the caller doesn't see "garbage" data that was there before.

This _shouldn't_ matter in practice, as the caller always verifies the checksum anyway, but still feels like a good defense in depth.

### DIFF
--- a/src/storage.zig
+++ b/src/storage.zig
@@ -188,6 +188,7 @@ pub const Storage = struct {
 
     fn start_read(self: *Storage, read: *Storage.Read, bytes_read: ?usize) void {
         assert(read.offset % constants.sector_size == 0);
+        maybe(bytes_read == 0); // Retrying erroneous read at the same offset with smaller window.
 
         const bytes = bytes_read orelse 0;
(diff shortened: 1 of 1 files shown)


matklad:
Found this one live at https://youtu.be/THWDx_RyZ6A!

src/storage.zig
```diff
@@ -299,7 +300,8 @@ pub const Storage = struct {
         // - Another replica requested a block, but we are lagging far behind, and the block address
         //   requested is beyond the end of our data file.
         if (bytes_read == 0) {
-            read.callback(read);
+            @memset(read.buffer, 0);
+            self.start_read(read, read.buffer.len);
```
matklad, line 304: Going through `start_read` rather then directly invoking callback to have "a single return" from the async loop.

sentientwaffle (approved):

PR #1440, opened 2024-01-18
VSR/Sync: Remove jump_sync_target() kludge for R=2

sentientwaffle:
Thanks to https://github.com/tigerbeetle/tigerbeetle/pull/1346, there is no risk in a R=2 cluster of the leading replica overwriting a prepare required by the lagging replica. This workaround is not needed anymore!

(I removed the `"Cluster: sync: checkpoint diverges, sync (primary diverges)"` test in a separate commit to emphasize that the test still passes, it just isn't really needed anymore since it doesn't exercise anything interesting anymore.)

### DIFF
--- a/src/vsr/replica_test.zig
+++ b/src/vsr/replica_test.zig
@@ -964,8 +964,7 @@ test "Cluster: sync: sync, bump target, sync" {
     try TestReplicas.expect_sync_done(t.replica(.R_));
 }
 
-test "Cluster: sync: R=2" {
-    // TODO explicit code coverage marks: "candidate is canonical (R=2)"
+test "Cluster: repair: R=2 (primary checkpoints, but backup lags behind)" {
     const t = try TestContext.init(.{ .replica_count = 2 });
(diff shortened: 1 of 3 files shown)


src/vsr/replica_test.zig
```diff
@@ -964,49 +964,6 @@ test "Cluster: sync: sync, bump target, sync" {
     try TestReplicas.expect_sync_done(t.replica(.R_));
 }
 
-test "Cluster: sync: R=2" {
```
matklad, line 967: Hm, can we keep the test in? Even if there's no longer any special code needed, this is still an interesting case to test

sentientwaffle: Sure! I made a couple tweaks to it:
- The test doesn't actually trigger state sync anymore. (R=2 clusters _can't_ trigger state sync anymore, not counting standbys.)
- Prepare more ops into the new checkpoint, to emphasize the pipeline.

matklad (approved):

PR #1443, opened 2024-01-20
[Dotnet client] Switch from Arrays to Slices

### LINKED ISSUE #1442: Batching example for C# seems wrong?
We have the following as an example of batching in C#:

```cs
var BATCH_SIZE = 8190;
for (int i = 0; i < transfers.Length; i += BATCH_SIZE)
{
    var batchSize = BATCH_SIZE;
    if (i + BATCH_SIZE > transfers.Length)
    {
        batchSize = transfers.Length - i;
    }
    var segment = new ArraySegment<Transfer>(transfers, i, batchSize);
    createTransfersError = client.CreateTransfers(segment.Array);
    // error handling omitted
}
```

This seems wrong? If I am reading the docs correctly, `segment.Array` returns the entire array, not a subslice.

Which makes me think, should we perhaps be using spans in our API?

batiati:
closes #1442

### DIFF
--- a/src/clients/dotnet/README.md
+++ b/src/clients/dotnet/README.md
@@ -270,8 +270,7 @@ for (int i = 0; i < transfers.Length; i += BATCH_SIZE)
     {
         batchSize = transfers.Length - i;
     }
-    var segment = new ArraySegment<Transfer>(transfers, i, batchSize);
-    createTransfersError = client.CreateTransfers(segment.Array);
+    createTransfersError = client.CreateTransfers(transfers[i..batchSize]);
     // error handling omitted
 }
 ```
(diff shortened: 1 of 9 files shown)


matklad (approved):

src/clients/dotnet/README.md
```diff
@@ -270,8 +270,7 @@ for (int i = 0; i < transfers.Length; i += BATCH_SIZE)
     {
         batchSize = transfers.Length - i;
     }
-    var segment = new ArraySegment<Transfer>(transfers, i, batchSize);
-    createTransfersError = client.CreateTransfers(segment.Array);
+    createTransfersError = client.CreateTransfers(transfers[i..batchSize]);
```
tenatus, line 273: This isn't correct--the range syntax in C# is of the form [inclusive_start..exclusive_end] (while in the ArraySegment ctor of the previous version the second int is a count).

The line should be written: `createTransfersError = client.CreateTransfers(transfers[i..(i + batchSize)]);`

batiati: Thanks @tenatus!
Ref https://github.com/tigerbeetle/tigerbeetle/pull/2458

Issue #259, opened 2022-11-18
Roadmap


sentientwaffle:
The tasks listed are not guaranteed, and they are not ordered.
The intent is to give a sense of the project's direction.

- :red_circle:: Must be done for the production release.
- :yellow_circle:: Nice-to-have for the production release.

### Stability: Storage

- [x] VSR: Manifest free-set. :red_circle:
- [x] VSR: 256-byte headers. :red_circle:
  - [x] VSR: State machine version in headers. :red_circle:
- [x] VSR: Async checkpoints :red_circle:
- [x] LSM: Remove filter blocks. :red_circle:
- [x] LSM: Re-implement secondary index tombstone optimization. :yellow_circle: [#1352](https://github.com/tigerbeetle/tigerbeetle/pull/1352)
- [ ] [LSM: Compaction optimizations](https://github.com/tigerbeetledb/tigerbeetle/issues/274)
  - [ ] Coalesce small adjacent tables (context: https://github.com/tigerbeetledb/tigerbeetle/pull/463) :red_circle:
  - [ ] "Move-data-block" (more granular than "move table")
  - [ ] Start the next round of compaction reads _before_ starting the merge (cpu) work.
  - [ ] Last-level of each tree should have double-size tables, but half as many. :red_circle:
- [x] VSR: Size manifest trailer, and pace manifest compaction, to guarantee capacity. :red_circle:
- [ ] ~VSR: Encode configuration data into superblock.~ :red_circle:
- [x] VSR: Reserve more space in `SuperBlock.VSRState` for future use. :red_circle:
- [ ] Guard against running a binary against a data file that was created with a different configuration. :yellow_circle:
- [x] LSM: Add value count to `TableInfo`. (And possibly value-block count, since compression will decouple the ratio between the two.) :red_circle:
- [x] Redo snapshots. :red_circle: 
  - [ ] Snapshots should be relative to the op that "creates" them, not the op that compacts them.
  - [ ] Maybe use timestamps instead of ops as snapshot ids.
  - [ ] Store snapshot in manifest block header (like we do for all other blocks).
- [ ] Reserve a some extra space in the superblock for future use, just in case? (Since "growing" the superblock is not possible once a replica is formatted.)
- [x] VSR: Align grid zone start to grid block size. :red_circle:
- [x] VSR: Remove superblock trailers. :red_circle:
  - [x] Encode the client sessions trailer into the grid.
[...]

PR #1447, opened 2024-01-22
Allow the node pool size to be set on the command line

brson:
This partially addresses the [roadmap](https://github.com/tigerbeetle/tigerbeetle/issues/259) item

> LSM: Runtime-configurable NodePool size, with a sane default. (Currently it is constant and too small.)

It adds a CLI flag, `--lsm-manifest-memory`, that says how many bytes to allocate for the node pool. It does not address the matter of fixing the default size of the node pool. If this PR is merged I will update the roadmap to split these items in two and indicate one is not complete.

I'm open to other names for this flag. I chose this one since it seems relatively understandable, though it requires understanding the concept of a "manifest". This argument is similar in function to `--grid-cache` though, and the name of the flag doesn't follow the same format, e.g. it could be called `--lsm-manifest`.

I am unsure about the help text

> "When the number or size of LSM-trees would become too large for their 
    manifests to fit into memory, such requests are rejected."

In particular: "number or size" - is this true? Additionally, from comments I think it is not the case that such requests are _currently_ rejected, but that is the intent.

The value of this flag must be a large multiple of 16KB (the default is 64MB), which is perhaps not the easiest for users to reason about without help, and the error text currently displays this 16KB multiplier in bytes, which reads poorly.

The minimum value of this flag is equal to one node, which in practice would seemingly be a poor choice. The maximum value of this flag is the largest multiple of the node size which won't cause overflows elsewhere (slightly less than `maxInt(u32)`).

To match the grid cache, I have added this value to the initialization log sequence, splitting the current line about allocations into two, as one line was looking long to me on my console at 107 characters.

### DIFF
--- a/src/constants.zig
+++ b/src/constants.zig
@@ -555,6 +555,19 @@ pub const lsm_table_data_blocks_max = table_blocks_max: {
     );
 };
 
+/// The default size in bytes of the NodePool used for the LSM forest's manifests.
+pub const lsm_manifest_memory_size_default = lsm_manifest_memory: {
+    // TODO Tune this better.
+    const lsm_forest_node_count: u32 = 4096;
+    break :lsm_manifest_memory lsm_forest_node_count * lsm_manifest_node_size;
+};
+
+/// The maximum size in bytes of the NodePool used for the LSM forest's manifests.
+pub const lsm_manifest_memory_size_max = lsm_manifest_memory_max: {
+    const u32_max = std.math.maxInt(u32);
+    break: lsm_manifest_memory_max u32_max - (u32_max % lsm_manifest_node_size);
+};
+
 /// The number of milliseconds between each replica tick, the basic unit of time in TigerBeetle.
 /// Used to regulate heartbeats, retries and timeouts, all specified as multiples of a tick.
 pub const tick_ms = config.process.tick_ms;
--- a/src/constants.zig
+++ b/src/constants.zig
@@ -565,7 +565,7 @@ pub const lsm_manifest_memory_size_default = lsm_manifest_memory: {
 /// The maximum size in bytes of the NodePool used for the LSM forest's manifests.
 pub const lsm_manifest_memory_size_max = lsm_manifest_memory_max: {
     const u32_max = std.math.maxInt(u32);
-    break: lsm_manifest_memory_max u32_max - (u32_max % lsm_manifest_node_size);
+    break :lsm_manifest_memory_max u32_max - (u32_max % lsm_manifest_node_size);
 };
 
 /// The number of milliseconds between each replica tick, the basic unit of time in TigerBeetle.
--- a/src/constants.zig
+++ b/src/constants.zig
@@ -563,10 +563,8 @@ pub const lsm_manifest_memory_size_default = lsm_manifest_memory: {
 };
 
 /// The maximum size in bytes of the NodePool used for the LSM forest's manifests.
-pub const lsm_manifest_memory_size_max = lsm_manifest_memory_max: {
-    const u32_max = std.math.maxInt(u32);
-    break :lsm_manifest_memory_max u32_max - (u32_max % lsm_manifest_node_size);
-};
+pub const lsm_manifest_memory_size_max =
+    @divFloor(std.math.maxInt(u32), lsm_manifest_node_size) * lsm_manifest_node_size;
 
 /// The number of milliseconds between each replica tick, the basic unit of time in TigerBeetle.
 /// Used to regulate heartbeats, retries and timeouts, all specified as multiples of a tick.
--- a/src/constants.zig
+++ b/src/constants.zig
@@ -564,7 +564,21 @@ pub const lsm_manifest_memory_size_default = lsm_manifest_memory: {
 
 /// The maximum size in bytes of the NodePool used for the LSM forest's manifests.
 pub const lsm_manifest_memory_size_max =
-    @divFloor(std.math.maxInt(u32), lsm_manifest_node_size) * lsm_manifest_node_size;
+    @divFloor(std.math.maxInt(u32), lsm_manifest_memory_size_multiplier) *
+    lsm_manifest_memory_size_multiplier;
+
+/// The minimum size in bytes of the NodePool used for the LSM forest's manifests.
+pub const lsm_manifest_memory_size_min = lsm_manifest_memory_size_multiplier;
+
+/// The lsm memory size must be a multiple of this value.
+///
+/// While technically this could be equal to lsm_manifest_node_size, we set it
+/// to 1MB so it is a more obvious increment for users.
+pub const lsm_manifest_memory_size_multiplier = lsm_manifest_memory_multiplier: {
+    const lsm_manifest_memory_multiplier = 64 * lsm_manifest_node_size;
+    assert(lsm_manifest_memory_multiplier == 1024 * 1024);
+    break :lsm_manifest_memory_multiplier lsm_manifest_memory_multiplier;
+};
 
 /// The number of milliseconds between each replica tick, the basic unit of time in TigerBeetle.
 /// Used to regulate heartbeats, retries and timeouts, all specified as multiples of a tick.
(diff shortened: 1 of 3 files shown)


brson:
I'm not quite seeing offhand what the lint error is and how to fix it. Can someone explain? Is it the long lines errors?

matklad:
The actual error here is just automated formatter:

```
zig fmt --check: error: the following command exited with error code 1:
/home/runner/work/tigerbeetle/tigerbeetle/zig/zig fmt --check 
```

So, running `zig fmt .` should fix it.

matklad:
The bulk of the output is from out “lint” script, which _tries_ to check for things like line lengths and assertion density, but doesn’t actually _enforce_ anything, predictably leading to bad signal-to-noise ratio. Perhaps it might be best to remove that from our CI.

We have a second iteration on this infra in out tidy.zig, which is implemented as real tests that prevent code from being merged, and which we are incrementally enabling for more and more files:

https://github.com/tigerbeetle/tigerbeetle/blob/992e1a535c4ceb6fcf1d406b75937da4f62d4646/src/tidy.zig#L40

EDIT: lint's dead: https://github.com/tigerbeetle/tigerbeetle/pull/1448

sentientwaffle:
> "When the number or size of LSM-trees would become too large for their manifests to fit into memory, such requests are rejected."

Rejecting requests was our intent at one point, but iirc we changed tack and are now planning to just panic (with an informative error for the operator) in this case.

jorangreef:
Great to see this, @brson!

> The value of this flag must be a large multiple of 16KB (the default is 64MB), which is perhaps not the easiest for users to reason about without help, and the error text currently displays this 16KB multiplier in bytes, which reads poorly.

We could be "over strict" here (one can always relax a minimal surface area, but going the other way is harder) and enforce that the value of this flag must be a multiple of 1 MiB (64 x 16 KiB nodes) to make it easier for users to reason about.

jorangreef:
Regarding consistency of naming across CLI flags, we don't have to solve it in this PR.

But, I think we should aim for putting `--memory` first as the MSB, with `-cache-grid` (we'll likely have more than one `cache` so `grid` here is the qualifier after `cache`) and `-lsm-manifest` as qualifiers.

Eventually, the goal is to have a single `--memory` argument, from which TB itself can derive and work out an optimal mix for all the others (these then become more advanced flags).

src/tigerbeetle/cli.zig
```diff
@@ -259,6 +269,33 @@ pub fn parse_args(allocator: std.mem.Allocator) !Command {
...
+            }
+            if (lsm_manifest_memory % constants.lsm_manifest_node_size != 0) {
+                flags.fatal(
+                    "--lsm-memory: size {} must be a multiple of node size ({})",
+                    .{ lsm_manifest_memory, constants.lsm_manifest_node_size },
+                );
+            }
+
+            const lsm_forest_node_count_u64 =
+                lsm_manifest_memory / constants.lsm_manifest_node_size;
```
sentientwaffle: Prefer being explicit with `@divExact()`/`@divFloor()` to `/` (`@divExact()`, in this case).

src/tigerbeetle/cli.zig
```diff
@@ -259,6 +269,33 @@ pub fn parse_args(allocator: std.mem.Allocator) !Command {
                 );
             }
 
+            const lsm_manifest_memory = start.lsm_manifest_memory.bytes;
+            const lsm_manifest_memory_max = constants.lsm_manifest_memory_size_max;
+            const lsm_manifest_memory_min = constants.lsm_manifest_node_size;
```
sentientwaffle: :thinking: Maybe the minimum should be `lsm_manifest_node_size * number of trees`.

src/tigerbeetle/cli.zig
```diff
@@ -107,6 +110,12 @@ const CliArgs = union(enum) {
         \\        (Total RAM) - 3GB (TigerBeetle) - 1GB (System), eg 12GB for a 16GB machine.
         \\        Defaults to 1GB.
         \\
+        \\  --lsm-manifest-memory=<size><KB|MB|GB>
+        \\        Sets the amount of memory allocated for LSM-tree manifests. When the
+        \\        number or size of LSM-trees would become too large for their manifests to fit
+        \\        into memory, such requests are rejected.
+        \\        Defaults to 64MB.
```
sentientwaffle, line 116: For this (and the `--cache-grid` default) it would be nice if we actually formatted the default value from `constants` into `comptimePrint`, like we do for the default address/port.

src/constants.zig
```diff
@@ -555,6 +555,19 @@ pub const lsm_table_data_blocks_max = table_blocks_max: {
...
+pub const lsm_manifest_memory_size_default = lsm_manifest_memory: {
+    // TODO Tune this better.
+    const lsm_forest_node_count: u32 = 4096;
+    break :lsm_manifest_memory lsm_forest_node_count * lsm_manifest_node_size;
+};
+
+/// The maximum size in bytes of the NodePool used for the LSM forest's manifests.
+pub const lsm_manifest_memory_size_max = lsm_manifest_memory_max: {
+    const u32_max = std.math.maxInt(u32);
+    break: lsm_manifest_memory_max u32_max - (u32_max % lsm_manifest_node_size);
```
sentientwaffle:
```suggestion
    break :lsm_manifest_memory_max u32_max - (u32_max % lsm_manifest_node_size);
```

TIL that `break: symbol` actually parses, though!

sentientwaffle: Also I wonder if `@divFloor(std.math.maxInt(u32), lsm_manifest_node_size) * lsm_manifest_node_size` would be more clear? For me at least it is clearer idiom, and it removes the need for the extracted `u32_max` constant.

src/tigerbeetle/cli.zig
```diff
@@ -259,6 +269,33 @@ pub fn parse_args(allocator: std.mem.Allocator) !Command {
...
+            if (lsm_manifest_memory % constants.lsm_manifest_node_size != 0) {
+                flags.fatal(
+                    "--lsm-memory: size {} must be a multiple of node size ({})",
+                    .{ lsm_manifest_memory, constants.lsm_manifest_node_size },
+                );
+            }
+
+            const lsm_forest_node_count_u64 =
+                lsm_manifest_memory / constants.lsm_manifest_node_size;
+            assert(lsm_forest_node_count_u64 <= std.math.maxInt(u32));
```
sentientwaffle: This assertion is unnecessary; it will be handled by `@intCast()`.

src/tigerbeetle/cli.zig
brson: @sentientwaffle could you say a bit more about how grooves, forests, ledgers and trees all relate, and how to sum them all up into a total number of trees?

I think a single groove corresponds to a single ledger, and that probably there are multiple trees per ledger, but beyond that I don't have the full picture.

Is there a fixed limit to trees per groove and grooves per forest?

src/constants.zig
brson: Fixed by zig fmt. Now I know to use it.

src/tigerbeetle/cli.zig
sentientwaffle: The number of trees is known at comptime.
The state machine has exactly one forest.
Each forest has one or more grooves.
Each groove has one or more trees.

...But now I realize that `constants` (intentionally) doesn't have access to the `StateMachine`, so maybe the current minimum value is fine. Presumably this is a setting primarily for experts anyway.

brson: If I take @jorangreef's suggestion to require this value be a 1MB multiple, then that will make the minimum 1MB, which may be larger than the suggested calculation here.

brson:
I've made most of the needed revisions here, but still thinking about it, not quite done.

I took @jorangreef's suggestion to make the minimum size 1MB.

src/tigerbeetle/cli.zig
brson: Since this is a preexisting issue I'd prefer to fix them both in a separate PR and have filed this issue about it https://github.com/tigerbeetle/tigerbeetle/issues/1466

brson:
I think I've addressed all reviews and am ready for another look here @sentientwaffle. Some minor preexisting issues I prefer to address as separate PRs and have filed issues for them, linked above.

sentientwaffle (approved):

matklad:
@brson do you have GitHub permissions to click merge-when-ready here?

In our current practice, the “final call” to merge a PR is author’s. It is also common to hit “merge when ready” _before_ the review is done to avoid and extra round trip.

brson:
@matklad yes I can click it. Thanks.

PR #1449, opened 2024-01-22
ci: drop dead code from yml

matklad:
Triggers for release.yml are specified in the workflow itself:

```
name: Release

on:
  workflow_dispatch: # Manual triggering to re-run a release with a different run_number
  push:
    branches:
      - release
```

given that it misses `workflow_call`, I am actually surprise that we don't get any kind of error for the `pipeline-main.yml`...

### DIFF
--- a/.github/workflows/pipeline-main.yml
+++ b/.github/workflows/pipeline-main.yml
@@ -10,22 +10,6 @@ jobs:
     uses: ./.github/workflows/pipeline-core.yml
   other:
     uses: ./.github/workflows/pipeline-other.yml
-  release:
-    needs:
-      - clients
-      - core
-      - other
-    uses: ./.github/workflows/release.yml
-    secrets:
-      TIGERBEETLE_NODE_PUBLISH_KEY: ${{secrets.TIGERBEETLE_NODE_PUBLISH_KEY}}
-      NUGET_KEY: ${{secrets.NUGET_KEY}}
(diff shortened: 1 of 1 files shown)


cb22:
> given that it misses workflow_call, I am actually surprise that we don't get any kind of error for the pipeline-main.yml...

Ha - https://github.com/tigerbeetle/tigerbeetle/actions/workflows/pipeline-main.yml

So it turns out the answer is, `pipeline-main.yml` is not running at all, and the checks we see on `main` are from the merge group :exploding_head:

matklad:
@cb22 so merging this as is should fix it, right? Or is there something you want me to do here?

cb22:
> @cb22 so merging this as is should fix it, right? Or is there something you want me to do here?

It will - what I meant to add though, was since it's been broken all this time and no one has noticed, do we need it at all?

matklad:
I think we do, for two reasons:

* "new code that passed not-rocket-science" is a good trigger to have, and exactly the thing we need for workbench
* running a _second_ round of CI is helpful to fish out any residual flakes

The non-discoverability is a bummer, but I think it'll get fixed soon:

* the workbench should let us notice that something's off with our chars, so if things do not run at all, we'll notice
* once the workflow parses, hopefully we'll also get a :x: in github's UI if any particular job fails

sentientwaffle (approved):

matklad:
>hopefully we'll also get a ❌ in github's UI if any particular job fails

Red crosses confirmed!

PR #1450, opened 2024-01-23
tidy: ewah_benchmark

matklad:


src/ewah_benchmark.zig
```diff
@@ -43,7 +43,11 @@ pub fn main() !void {
         var bitset_lengths: [samples]usize = undefined;
         while (i < samples) : (i += 1) {
             bitsets[i] = try make_bitset(allocator, config);
-            bitsets_encoded[i] = try allocator.alignedAlloc(u8, @alignOf(usize), ewah.encode_size_max(bitsets[0].len));
+            bitsets_encoded[i] = try allocator.alignedAlloc(
+                u8,
+                @alignOf(usize),
+                ewah.encode_size_max(bitsets[i].len),
```
matklad, line 49: NB: intentionally changed `[0]` to `[i]` here -- they both end up being `config.words`, but `i` seems less confusing and more future proof

sentientwaffle (approved):

Issue #1376, opened 2023-12-20
Allow recovering head replicas to contribute to DVC


matklad:
We are a little less tolerant of storage faults if they cause a replica to enter into recovering head --- there, corruptions in _unrelated_ disk sectors might make the cluster available. This seems fixabe -- recovering head replicas can still contribute some nacks to DVC quorum, they just need to be extra careful

See comment at https://github.com/tigerbeetle/tigerbeetle/issues/1283#issuecomment-1864856466 for details. 

The TL;DR there is that it seems like CTRL protocol is actually a strict superset of view-change protocol. We use the fact that a replica is _not_ recovering its head to nack a whole bunch of slots implicitly, but we could make this nacking explicit.

matklad:
Partially fixed by not getting replicas into recovering head in many cases!

PR #1452, opened 2024-01-23
Test: Add "failing" test for sync-recovering-head issue

sentientwaffle:
Replicas in `status=recovering_head` cannot (currently) participate in view-change, even when they arrived at recovering_head via state sync, not corruption+crash. As a result, it is possible for a 2/3 cluster to get stuck without any corruptions or crashes.

See: https://github.com/tigerbeetle/tigerbeetle/pull/933#discussion_r1245440623 and `Simulator.core_missing_quorum()`.


matklad:
And also https://github.com/tigerbeetle/tigerbeetle/issues/1376

matklad (approved):
:heart: love negative tests!

PR #1453, opened 2024-01-23
ci: fix command name in yml

matklad:


matklad:
Name defined here:

https://github.com/tigerbeetle/tigerbeetle/blob/138e7be559e604d54913c2dccc31385be6530089/src/scripts/main.zig#L27

sentientwaffle (approved):

PR #1454, opened 2024-01-23
lsm: remove redundant stash

matklad:
At the moment, CacheMap handles lookups for both mutable and immutable
tables. However, immutable table is already sorted, so we can just look
into it directly, getting rid of the second stash.

This cuts self-reported memory usage with --cache-grid=512MB:
from 5043MB to 3085MB

And the RSS goes down:
from 3271MB to 2875MB


src/lsm/tree.zig
```diff
@@ -319,6 +319,10 @@ pub fn TreeType(comptime TreeTable: type, comptime Storage: type) type {
         /// ascertain the existence of the key are not in the Grid cache, it bails out.
         /// The returned `.positive` Value pointer is only valid synchronously.
         pub fn lookup_from_levels_cache(tree: *Tree, snapshot: u64, key: Key) LookupMemoryResult {
+            if (tree.table_immutable.get(key)) |value| {
+                return .{ .positive = value };
+            }
```
matklad, line 324: that's hte single addition in this PR

batiati:
Closed by #1584

PR #1457, opened 2024-01-23
Docs: Document prepare.checkpoint_id and constants.checkpoint_interval

sentientwaffle:


src/constants.zig
```diff
@@ -42,6 +42,15 @@ comptime {
...
+///   checkpoint trigger is not overwritten by the following checkpoint's entries. This final
+///   batch's updates were not persisted as part of the former checkpoint – they are only in memory
+///   until they are compacted by the *next* batch of commits (i.e. the first batch of the following
+///   checkpoint).
+/// - `pipeline_prepare_queue_max` (rounded up to the nearest batch multiple): This margin ensures
+///   that the entries prepared immediately following a checkpoint trigger never overwrite an entry
+///   from the previous WAL wrap until a quorum of replicas has reached that checkpoint.
 pub const vsr_checkpoint_interval = journal_slot_count -
     lsm_batch_multiple -
     lsm_batch_multiple * stdx.div_ceil(pipeline_prepare_queue_max, lsm_batch_multiple);
```
matklad: Let's maybe move the two points down to `assert` bellow, and here instead say that we pick the largest checkpoint interval that satisfies the assert. 

Basically, while the assert below and the initialization here looks redundant, they are not: we can pick _any_ value between lsm_batch_multiple and the upper bound, it's just that picking exactly the upper bound is most optimal for performance.

sentientwaffle: Excellent point!

matklad (approved):

PR #1462, opened 2024-01-24
repl: don't leak CLI arguments

matklad:
The first commit corrects a couple of minor error with repl's arguments. The two following comments refacro related code somewhat to hopefully make memory management somewhat easier.

### DIFF
--- a/src/flags.zig
+++ b/src/flags.zig
@@ -72,6 +72,8 @@ pub fn fatal(comptime fmt_string: []const u8, args: anytype) noreturn {
 ///
 /// If `pub const help` declaration is present, it is used to implement `-h/--help` argument.
 pub fn parse(args: *std.process.ArgIterator, comptime CliArgs: type) CliArgs {
+    assert(args.skip()); // Discard executable name.
+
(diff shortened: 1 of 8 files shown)


src/flags.zig
```diff
@@ -72,6 +72,8 @@ pub fn fatal(comptime fmt_string: []const u8, args: anytype) noreturn {
 ///
 /// If `pub const help` declaration is present, it is used to implement `-h/--help` argument.
 pub fn parse(args: *std.process.ArgIterator, comptime CliArgs: type) CliArgs {
+    assert(args.skip()); // Discard executable name.
```
kprotty, line 75: Should the `pub fn main()` in this file also go through `parse()` (it duplicates the skip logic).

matklad: Excellent catch @kprotty! Looks like we were skipping these tests! Should be fixed in the latest commit.

matklad:
Upstream issue: https://github.com/ziglang/zig/issues/18672

kprotty (approved):

PR #1463, opened 2024-01-24
cli: use port=0 to make testing with tigerbeetle easier

matklad:
It is possible to start tigerbetele passing `0` as an address:
    $ tigerbeetle start --addresses=0 0_0.tigrebeetle
This enables a couple of special behaviors, useful in tests:
- The operating system picks a free port, avoiding "address already in use" errors.
- The port, and only the port, is printed to the stdout, so that the parent process
  can learn it.
- tigerbeetle process exits when its stdin gets closed.

Practically, this allows us to remove a bunch of tricky asynchronous programming from all our integration tests:

- To get the port, just read a line of stdout. No need to worry about process deadlocking due to filling the stderr pipe
- To ensure that the process dies properly, pipe its stdin. Even if driver process is killed, the child will die, a pattern from https://matklad.github.io/2023/10/11/unix-structured-concurrency.html

We've discussed this couple of times already, but what I've realized today is that we can just do this when the port is zero. This is a (fair) bit of magical behavior, of course, but I think this is justified --- if you specify `--addresses=0`, you already need something extra to get the actual port anyway, and here we fully formalize this pattern. 

The backstory here is that, for `tigerbeetle benchmark`, I need to spawn a temporary tigrebeetle once again. I don't want to port the existing logic:

https://github.com/tigerbeetle/tigerbeetle/blob/30cfbfa2eca94b8cd0b1d2a8ce41c7e7720128f0/scripts/benchmark.sh#L12-L17

But I also don't want to depend on `tmp_tigerbeetle.zig` or `shell.zig` --- those things are for "test" and "scripts", and `tigrebeetle benchmark` is real production code. So I'd rather stick to raw `std.ChildProcess`, but doing the "let's stream stderr until this special line" dance again is also something I'd love to avoid.

### DIFF
--- a/src/clients/java/src/test/java/com/tigerbeetle/IntegrationTest.java
+++ b/src/clients/java/src/test/java/com/tigerbeetle/IntegrationTest.java
@@ -1546,43 +1546,13 @@ public class IntegrationTest {
                 throw new Exception("Format failed. " + error);
             }
 
-            this.process = new ProcessBuilder().command(
-                    new String[] {exe, "start", "--addresses=0", TB_FILE, "--cache-grid=512MB"})
-                    .redirectError(Redirect.PIPE).start();
+            this.process = new ProcessBuilder()
+                    .command(new String[] {exe, "start", "--addresses=0", TB_FILE,
+                            "--cache-grid=512MB"})
+                    .redirectOutput(Redirect.PIPE).redirectError(Redirect.DISCARD).start();
 
-            if (process.waitFor(100, TimeUnit.MILLISECONDS))
-                throw new Exception("Start server failed");
-
-            final var addressLoggedEvent = new CountDownLatch(1);
-            final var stderrLogged = new StringBuilder();
-            final var stderrReader = new Thread(() -> {
-                final var stderr = process.getErrorStream();
-                final var listening = "listening on ";
-                new BufferedReader(new InputStreamReader(stderr)).lines().forEach(line -> {
-                    stderrLogged.append(line);
-                    stderrLogged.append("\n");
-                    final var found = line.indexOf(listening);
-                    if (found != -1) {
-                        address = line.substring(found + listening.length()).trim();
-                        addressLoggedEvent.countDown();
-                    }
-                });
-            });
-            stderrReader.start();
-            process.onExit().whenCompleteAsync((process, exception) -> {
-                // Diagnose cases when the server crashes.
-                System.out.println(stderrLogged);
-
-                System.out.printf("TigerBeetle server exited with code %d\n\n",
-                        process.exitValue());
-            });
-
-            if (!addressLoggedEvent.await(60L, TimeUnit.SECONDS)) {
-                process.destroy();
-                stderrReader.join();
-                throw new Exception("failed to read the port, exitValue=" + process.exitValue()
-                        + " stderr:\n" + stderrLogged.toString());
-            }
+            final var stdout = process.getInputStream();
+            this.address = new BufferedReader(new InputStreamReader(stdout)).readLine().trim();
         }
 
         public String getAddress() {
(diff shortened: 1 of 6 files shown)


src/testing/tmp_tigerbeetle.zig
```diff
@@ -19,17 +19,7 @@ port_str: stdx.BoundedArray(u8, 8),
 
 tmp_dir: std.testing.TmpDir,
 
-// Lifetimes are tricky here! The order is:
```
matklad, line 22: Lifetimes are no longer tricky!

src/tigerbeetle/main.zig
```diff
@@ -227,6 +227,34 @@ const Command = struct {
...
 
+        // It is possible to start tigerbetele passing `0` as an address:
+        //     $ tigerbeetle start --addresses=0 0_0.tigrebeetle
+        // This enables a couple of special behaviors, useful in tests:
+        // - The operating system picks a free port, avoiding "address already in use" errors.
+        // - The port, and only the port, is printed to the stdout, so that the parent process
+        //   can learn it.
+        // - tigerbeetle process exits when its stdin gets closed.
+        const port_zero = std.net.Address.parseIp(constants.address, 0) catch unreachable;
+        if (args.addresses[replica.replica].eql(port_zero)) {
```
cb22: Maybe we should make this a bit more explicit?

If `constants.address` is eg `127.0.0.1`, then:

`--addresses=0` will trigger this mode, as will `--addresses=127.0.0.1:0` but not `--addresses=127.0.0.2:0`

src/tigerbeetle/main.zig
```diff
@@ -227,6 +227,34 @@ const Command = struct {
                 "if it's unexpected, please recompile TigerBeetle with -Dconfig-aof-recovery=false.", .{replica.replica});
         }
 
+        // It is possible to start tigerbetele passing `0` as an address:
+        //     $ tigerbeetle start --addresses=0 0_0.tigrebeetle
+        // This enables a couple of special behaviors, useful in tests:
+        // - The operating system picks a free port, avoiding "address already in use" errors.
+        // - The port, and only the port, is printed to the stdout, so that the parent process
+        //   can learn it.
+        // - tigerbeetle process exits when its stdin gets closed.
```
```suggestion
        // It is possible to start tigerbeetle passing `0` as an address:
        //     $ tigerbeetle start --addresses=0 0_0.tigerbeetle
        // This enables a couple of special behaviors, useful in tests:
        // - The operating system picks a free port, avoiding "address already in use" errors.
        // - The port, and only the port, is printed to the stdout, so that the parent process
        //   can learn it.
        // - tigerbeetle process exits when its stdin gets closed.
```

src/tigerbeetle/main.zig
matklad: Changed to check for literal `"0"` to lock this down. Enabling magic-zero mode for clusters of more than one machine seems not that useful, and, indeed, it's better to lock down anything we are not specifcially designing for!

src/clients/java/src/test/java/com/tigerbeetle/IntegrationTest.java
```diff
@@ -1546,43 +1546,13 @@ public Server() throws IOException, Exception, InterruptedException {
...
-
-            if (!addressLoggedEvent.await(60L, TimeUnit.SECONDS)) {
-                process.destroy();
-                stderrReader.join();
-                throw new Exception("failed to read the port, exitValue=" + process.exitValue()
-                        + " stderr:\n" + stderrLogged.toString());
-            }
+            this.process = new ProcessBuilder()
+                    .command(new String[] {exe, "start", "--addresses=0", TB_FILE,
+                            "--cache-grid=512MB"})
```
sentientwaffle, line 1551: TIL TB's positional arguments can be interleaved with flags!

sentientwaffle (approved):

src/clients/java/src/test/java/com/tigerbeetle/IntegrationTest.java
matklad: Probably should forbid this though...

matklad: https://github.com/tigerbeetle/tigerbeetle/pull/1478

PR #1470, opened 2024-01-25
Timestamp range in GetAccountTransfers filter

batiati:
The motivation is to provide the ability to filter by a time window, and not only to use the timestamp for pagination.
Currently, the [timestamp parameter](https://docs.tigerbeetle.com/reference/operations/get_account_transfers/#timestamp) is "exclusive" and behaves as `>` or `<` depending on the sort order defined by `flags.reversed`.

The reason behind this design is to provide a simplified API for pagination, i.e., for fetching the next page, the application can just use the timestamp of the last retrieved transfer like `filter.timestamp = last_timestamp_seen` no matter the sort order.

New filter signature:
```zig
pub const GetAccountTransfers = extern struct {
    /// The account id.
    account_id: u128,

    /// The initial timestamp (inclusive).
    /// Use ZERO for no filter.
    timestamp_min: u64,

    /// The final timestamp (inclusive).
    /// Use ZERO for no filter.
    timestamp_max: u64,

    /// Maximum number of transfers that can be returned by this query.
    limit: u32,

    /// Query flags.
    flags: GetAccountTransfersFlags,

    reserved: [24]u8 = [_]u8{0} ** 24,
}
```

With this change, both `timestamp_min` and `timestamp_max` will be inclusive (the SQL equivalent of `WHERE transfer.timestamp >= timestamp_min AND transfer.timestamp <= timestamp_max`).

For pagination purposes, the API gets more verbose and both parameters should be used according to the sort order, for example:
Set `filter.timestamp_min = last_timestamp_seen + 1` if `flags.reversed == false`
Set `filter.timestamp_max = last_timestamp_seen - 1` if `flags.reversed == true`


docs/reference/operations/get_account_transfers.md
```diff
@@ -15,18 +15,20 @@ Constraints:
 * Type is 128-bit unsigned integer (16 bytes)
 * Must not be zero or `2^128 - 1`
 
-### `timestamp`
+### `timestamp_min`
 
-The [`Transfer.timestamp`](../transfers.md#timestamp) from which results will be returned, exclusive range
-applied as:
+The minimum [`Transfer.timestamp`](../transfers.md#timestamp) from which results will be returned, inclusive range.
+Optional, use zero for no lower bound filter by timestamp.
```
sentientwaffle:
```suggestion
Optional; set to zero to disable the lower-bound filter.
```

docs/reference/operations/get_account_transfers.md
sentientwaffle, line 1: Hmm, what if both `timestamp_min` and `timestamp_max` were _exclusive_ instead of inclusive?
That way pagination still works as before.

...I think (more generally) I am confused whether it is good/safe for an application to use tigerbeetle's hybrid logical clock as a wall-clock timestamp, as opposed to recommending that applications use one of the user_data fields as an application-friendly timestamp.

(That said, I'm very open to using `timestamp` as an application timestamp -- user_data is a limited resource!)

jorangreef: TigerBeetle's cluster time HLC is definitely meant to be used as the authoritative wall-clock timestamp by the application.

It's good/safe, and also better than the API layer using different clocks in each stateless API instance, since TB can detect clock faults and clock skew and guarantee that a quorum of clocks feed into what is "true time" ("cluster time"), whereas stateless API instances might have unsynced clocks (or worse, might be partitioned asymmetrically from any clock sync protocol, with no way to detect that).

Another way of thinking about this, if TB's timestamps were not more trustworthy, then TB shouldn't be timestamping transactions, but instead each (unsynchronized) stateless API instance would be trusted to give its own view of time, and to timestamp transactions. But this is dangerous, exposing pending transfers to clock jumps if they happen in any instance in the API layer. And fixing this in the API layer would put more burden on what should be a simple stateless layer—designing the API layer to higher fault tolerances (i.e. implementing clock sync, as well as clock sync accuracy verification). Better for the smaller, centralized stateful component to take these burdens, than to place them on the many stateless components.

Finally, yet another way to see this, the timestamps here are used against recorded, historical, authoritative data from the financial system of record (TB, not stateless API layer). If the API layer instance's own clock seems to be behind/ahead, then that is simply the "clock fault model" playing out, and the API layer instance should realize it has strayed from the authority in the system.

batiati: > Hmm, what if both timestamp_min and timestamp_max were exclusive instead of inclusive?
That way pagination still works as before.

The Query API will probably allow the user to choose the operator (e.g., `>` or `>=`), but for now, we need to make this decision.

I think inclusive ranges are more natural when dealing with time periods. Even with exclusive ranges, it would still require the user to set the correct field `timestamp_min` or `timestamp_max` for pagination depending on the sort order, so it's not a big win.

sentientwaffle (dismissed):

sentientwaffle (approved):

PR #1474, opened 2024-01-25
Improve integration tests

batiati:
While investigating this [Rafik issue](https://github.com/interledger/rafiki/actions/runs/7653128123/job/20854361776?pr=2364), I noticed that our Node.js CI passes even if tests fail.

- Ensure that the Node.js test runner returns a non-zero exit code when tests fail.
- Add Node, Java, and Dotnet tests for `pending_transfer_expired`.

### DIFF
--- a/src/clients/java/src/test/java/com/tigerbeetle/IntegrationTest.java
+++ b/src/clients/java/src/test/java/com/tigerbeetle/IntegrationTest.java
@@ -72,11 +72,7 @@ public class IntegrationTest {
 
     @Test(expected = NullPointerException.class)
     public void testConstructorNullReplicaAddresses() throws Throwable {
-
         try (var client = new Client(clusterId, null)) {
-
-        } catch (Throwable any) {
-            throw any;
         }
     }
 
@@ -85,20 +81,13 @@ public class IntegrationTest {
 
         var replicaAddresses = new String[] {"3001", null};
         try (var client = new Client(clusterId, replicaAddresses)) {
-
-        } catch (Throwable any) {
-            throw any;
         }
     }
 
     @Test(expected = IllegalArgumentException.class)
     public void testConstructorEmptyReplicaAddresses() throws Throwable {
-
         var replicaAddresses = new String[0];
         try (var client = new Client(clusterId, replicaAddresses)) {
-
-        } catch (Throwable any) {
-            throw any;
         }
     }
 
@@ -129,7 +118,6 @@ public class IntegrationTest {
 
     @Test
     public void testConstructorInvalidReplicaAddresses() throws Throwable {
-
         var replicaAddresses = new String[] {"127.0.0.1:99999"};
         try (var client = new Client(clusterId, replicaAddresses)) {
             assert false;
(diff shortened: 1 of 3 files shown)


src/clients/java/src/test/java/com/tigerbeetle/IntegrationTest.java
```diff
@@ -699,6 +699,103 @@ public void testCreatePendingTransfersAndVoid() throws Throwable {
         }
     }
 
+    @Test
+    public void testCreatePendingTransfersAndVoidExpired() throws Throwable {
+
+        try (var server = new Server()) {
```
matklad: Drive-by, but can’t we take advantage of using declaration here and save an indent?

matklad: 🤦  this is Java, not C#

matklad: Though, we should be able to put both the server and the client into a single try?

```
try (FileReader fr = new FileReader(path);
	         BufferedReader br = new BufferedReader(fr)) {
	        return br.readLine();
	    }
```

matklad: And I think we don’t need catch here?

src/clients/node/src/test.ts
```diff
@@ -562,20 +562,24 @@ test('can get account transfers', async (): Promise<void> => {
...
   } finally {
     await client.destroy()
   }
+
+  const end = new Date().getTime()
+  console.log('Time taken (s):', (end - start)/1000)  
+  process.exit(fail_count);
 }
 
 main().catch((error: AssertionError) => {
```
matklad, line 581: Please rebase away changes to the lockfile, we intend to keep it frozen until we remove docusaurus altogether. 

(I tried commented on the lockfile diff, but my browser crashed)

matklad, line 581: Let’s rather exit process in the .catch to main here, to have a guarantee that any exception results in abnormal termination

src/clients/java/src/test/java/com/tigerbeetle/IntegrationTest.java
batiati: TIL it's possible to have more than one object inside a Java try-resource block!!

> And I think we don’t need catch here?

Yes, since we don't throw on `close` anymore, we can remove this catch block!

matklad (dismissed):

matklad:
Failure seems legit:

```
    Failed CreatePendingTransfersAndVoidExpiredAsync [6 s]
    Error Message:
     Assert.IsTrue failed. 
    Stack Trace:
       at TigerBeetle.Tests.IntegrationTests.CreatePendingTransfersAndVoidExpiredAsync() in /home/runner/work/tigerbeetle/tigerbeetle/src/clients/dotnet/TigerBeetle.Tests/IntegrationTests.cs:line 723
```

batiati:
> Failure seems legit:

Looks like `Task.Delay` does not have accurate resolution and sometimes finishes sooner than the timeout 😨.
Kind of the same problem here: https://github.com/zkSNACKs/WalletWasabi/issues/6166

matklad (approved):

PR #1476, opened 2024-01-26
vsr: liveness when SV strands checkpoint boundary in an idle cluster

matklad:
An SV might include messages from the current and the next checkpoint.

Installing headers from the next checkpoint would be unsafe, so replica effectively accepts only _part_ of an SV message if it hasn't checkpointed yet.

What makes this tricky is that "skipped" headers are from the previous view.

So, if a a replica sees such a header/prepare after it checkpoints, it will _still_ ignore it: this header is from the previous view, so it could have been truncated.

For prepares from older views, a replica needs primary to positively tell it that, yes, that prepare is part of the canonical history.

To fix this, detect if our view headers are ahead of our op during repair, and request a start view in that case.

SEED: 15556801325249033142

### DIFF
--- a/src/vsr/replica.zig
+++ b/src/vsr/replica.zig
@@ -5336,23 +5336,28 @@ pub fn ReplicaType(
 
             self.sync_reclaim_tables();
 
-            // Request outstanding committed headers to advance our op number:
+            // Request outstanding possibly committed headers to advance our op number:
             // This handles the case of an idle cluster, where a backup will not otherwise advance.
             // This is not required for correctness, but for durability.
-            if (self.op < self.op_repair_max()) {
+            if (self.op < self.op_repair_max() or
+                (self.status == .normal and self.op < self.view_headers.array.get(0).op))
+            {
                 assert(!self.solo());
                 assert(self.replica != self.primary_index(self.view));
 
                 log.debug(
-                    "{}: repair: break: view={} break={}..{} (commit={}..{} op={})",
+                    "{}: repair: break: view={} break={}..{} " ++
+                        "(commit={}..{} op={} view_headers_op={})",
                     .{
                         self.replica,
                         self.view,
                         self.op + 1,
                         self.op_repair_max(),
+
                         self.commit_min,
                         self.commit_max,
                         self.op,
+                        self.view_headers.array.get(0).op,
                     },
                 );
 
@@ -5366,8 +5371,9 @@ pub fn ReplicaType(
                         .nonce = self.nonce,
                     }),
                 );
-                return;
             }
+
+            if (self.op < self.op_repair_max()) return;
 
             // Request any missing or disconnected headers:
             if (self.journal.find_latest_headers_break_between(
--- a/src/vsr/replica_test.zig
+++ b/src/vsr/replica_test.zig
@@ -498,8 +498,10 @@ test "Cluster: repair: view-change, new-primary lagging behind checkpoint, forfe
     // Thanks to the new primary, the lagging backup is able to catch up to the latest
     // checkpoint/commit.
     try expectEqual(b1.role(), .backup);
-    try expectEqual(b1.commit(), checkpoint_1_trigger);
+    try expectEqual(b1.commit(), checkpoint_1_trigger + 1);
     try expectEqual(b1.op_checkpoint(), checkpoint_1);
+
+    try expectEqual(t.replica(.R_).commit(), checkpoint_1_trigger + 1);
 }
 
 // TODO: Re-enable when the op_checkpoint hack is removed from ignore_request_message().


src/vsr/replica_test.zig
```diff
@@ -498,8 +498,10 @@ test "Cluster: repair: view-change, new-primary lagging behind checkpoint, forfe
     // Thanks to the new primary, the lagging backup is able to catch up to the latest
     // checkpoint/commit.
     try expectEqual(b1.role(), .backup);
-    try expectEqual(b1.commit(), checkpoint_1_trigger);
+    try expectEqual(b1.commit(), checkpoint_1_trigger + 1);
```
matklad, line 501: Heh, turns out we _were_ testing this already!

src/vsr/replica.zig
```diff
@@ -5336,10 +5336,12 @@ pub fn ReplicaType(
 
             self.sync_reclaim_tables();
 
-            // Request outstanding committed headers to advance our op number:
+            // Request outstanding possibly committed headers to advance our op number:
             // This handles the case of an idle cluster, where a backup will not otherwise advance.
             // This is not required for correctness, but for durability.
-            if (self.op < self.op_repair_max()) {
+            if (self.op < self.op_repair_max() or
+                (self.status == .normal and self.op < self.view_headers.array.get(0).op))
```
matklad, line 5343: `self.status == .normal and` is needed because, during view change, primary truncates its op&log before it updates its view headers.

src/vsr/replica.zig
```diff
@@ -5336,10 +5336,12 @@ pub fn ReplicaType(
...
+            {
                 assert(!self.solo());
                 assert(self.replica != self.primary_index(self.view));
```
sentientwaffle, line 5347: Can you add `self.view_headers.array.get(0).op` to the log line below? Right now it just has `self.op_repair_max()`.

src/vsr/replica.zig
sentientwaffle: To check my understanding here:

- The prepares coming in while we are in normal mode include a view number, but that is the view that they were _originally_ prepared in, which might be behind the cluster's current view.
- Since they are behind the current view (but ahead of our `self.op`) the lagging (normal) replica (which is in the same view as the primary) has no way of knowing that they are indeed part of its current view, so it drops them.
- When we receive an SV in normal mode (for the same view that we are currently in) it is allowed to repair our headers and advance our op.

Is it possible for the lagging replica to get stuck at this step though? That is, it wants to request a SV to advance `self.op`, but if it is still in the old checkpoint (for example, because it needs to `repair_prepare()` some corruption before it can commit), then it can't actually install any headers from the SV yet. But we still "block" `repair()` from calling `repair_prepare()` in that case. Should their be an additional `and self.op_checkpoint_trigger() < self.op` clause?

matklad: >Is it possible for the lagging replica to get stuck at this step though?

Yes! And that's exactly the issue I hit when I implemented the fix initially. In this second attempt, I am also moving the `return` _out_ of this if, so that we return _only_ if we send RSV due to the first condition of the branch. That is, for the new case we specifically avoid blocking the repair. 

To be fair, I am not quite happy with the conditions here --- they look somewhat adhoc, and it might be hard to reverse-engineer the logic behind them without `git blame`

src/vsr/replica_test.zig
matklad: I am wondering if its worth to add something like

```zig
    try t.expect(snap(@src(),
        \\ 0   / .       3V  23/_27/_27C   0:_28Jo  0/_0J!   0:_31Wo <__0:__0>   49119Gf  0G!   0G?   1/4Pp  0/3Rq
        \\ 1   /  .      1V  23/_28/_28C   0:_28Jo  0/_0J!   0:_31Wo <__0:__0>   49118Gf  0G!   0G?   0/4Pp  0/3Rq
        \\ 2   \   .     3V  23/_27/_27C   0:_27Jo  0/_0J!   0:_31Wo <__0:__0>   49119Gf  0G!   0G?
        \\
    ));
```

to our replica tests, so that surprising cluster states pop out more..

src/vsr/replica.zig
sentientwaffle: Ahh, that makes sense. Did you consider, as an alternative, using the local view headers directly (instead of doings so via the SV)? e.g.

```
            for (view_headers.slice) |*header| {
                if (header.op <= self.op_checkpoint_next_trigger()) {
                    self.replace_header(header);
                }
            }
```

matklad: Yeah, I actually started with that, and than switched to the current approach:

* re-using existing headers requires some code duplication, as there are essentially two places in the code we want to trigger that. Triggering that from _repair_ also feels a bit iffy, as advancing op is not, strictly speaking, repair (the main thing we want to do here is not as much replacing headers per se, as bumping our op). Doing from checkpoint callback felt kinda random. _And_ I'd love to minimize the number of places in code where we bump the `op`, as that's super sensitive. 
* if we hit this case, we are lagging behind the cluster somewhat. So, it is plausible that something else was prepared in the view while we were checkpointing. So asking the primary to just give us the latest and greatest prepares feels like it might be more optimal, as we'll have t query those headers anyway.

sentientwaffle: That makes perfect sense; thanks!

src/vsr/replica_test.zig
sentientwaffle: :thinking: The VOPR output format prints so much information that it may be hard to spot which parts are important/relevant to the test. Maybe we could wildcard `____` components that the test is unconcerned with, though then it wouldn't be a snap test.

sentientwaffle (approved):

src/vsr/replica_test.zig
matklad: >though then it wouldn't be a snap test.

It would! snaps's support ignores:

https://github.com/tigerbeetle/tigerbeetle/blob/6c2cfd7262dba32f2d1f95e1d29a6e3abb6fdc8b/src/integration_tests.zig#L101

(should have called those `snip`s though)

PR #1478, opened 2024-01-26
cli: enforce that options go before positionals

matklad:


matklad:
Looks like historically we weren't actually frobidding this:

https://github.com/tigerbeetle/tigerbeetle/blob/661d2f108aaf43eb90c59019ed4898c9b32d4963/src/tigerbeetle/cli.zig#L153-L187

but I bet it's just because forbidding that with our manual loop was a pain.

sentientwaffle (approved):

PR #1479, opened 2024-01-26
dotnet: suppress null warning

matklad:


matklad:
@batiati this was caught by github: 

https://github.com/tigerbeetle/tigerbeetle/pull/1463/files#annotation_17317769065

is there some easy way to make our build outright fail if there are such warnings?

batiati (changes requested):
> is there some easy way to make our build outright fail if there are such warnings?

We just need to add `<TreatWarningsAsErrors>true</TreatWarningsAsErrors>` in the `TigerBeetle.Tests.csproj`

src/clients/dotnet/TigerBeetle.Benchmarks/TigerBeetle.Benchmarks.csproj
```diff
@@ -5,6 +5,7 @@
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
     <UseCurrentRuntimeIdentifier>true</UseCurrentRuntimeIdentifier>
     <RollForward>LatestMajor</RollForward>
+    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
```
batiati, line 8:
```suggestion
    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
    <Nullable>enable</Nullable>    
```

I think we can also remove `<AllowUnsafeBlocks>true</AllowUnsafeBlocks>` for tests and benchmarks; only the main library needs it set as `true`.

Btw, it's another yak asking to be shaved: the clients' benchmark projects 🦬

batiati: > I think we can also remove `<AllowUnsafeBlocks>true</AllowUnsafeBlocks>` for tests and benchmarks; only the main library needs it set as `true`.

No 😆, the test project needs unsafe code for testing the packet logic.
But we can remove it from the benchmark project for sure.

matklad: It was worth trying!

batiati (approved):

PR #1481, opened 2024-01-26
VSR: Remove unused primary-start-view truncate-after-checkpoint logic

sentientwaffle:
Previously there was an "implicit" nack that happened around checkpoint boundaries -- we inferred, based on a DVC's checkpoint op, headers that it could not possibly have prepared.

Thanks to nack improvments (https://github.com/tigerbeetle/tigerbeetle/pull/939 I think) this is handled explicitly via DVC nacks, so this logic is unreachable.

(Also: If I am wrong and we _do_ need to use the checkpoint boundary to nack, then the right place to do that is the DVC's nack bitset, not a special case on the new primary.)

---

Remove an old test from `replica_test.zig` (after verifying that it is passing).
The test's job was to test a hard-to-hit branch of `on_do_view_change()`... but that branch no longer exists.

### DIFF
--- a/src/vsr/replica.zig
+++ b/src/vsr/replica.zig
@@ -1856,25 +1856,10 @@ pub fn ReplicaType(
                 // identify an unambiguous set of canonical headers.
                 self.log_view = self.view;
 
+                assert(self.op == op_head);
                 assert(self.op >= self.commit_max);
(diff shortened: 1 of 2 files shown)


sentientwaffle:
(Rebased)

matklad (approved):

PR #1482, opened 2024-01-26
ci: bring back windows tests for .Net client

matklad:
We disabled them, because we were apparently running out of RAM. Since then GitHub upped specs of public runners 2X, so we should fit now.

### DIFF
--- a/src/testing/tmp_tigerbeetle.zig
+++ b/src/testing/tmp_tigerbeetle.zig
@@ -71,7 +71,12 @@ pub fn init(
 
     // Pass `--addresses=0` to let the OS pick a port for us.
     var process = try shell.spawn_options(
-        .{ .stdin_behavior = .Pipe, .stderr_behavior = .Ignore },
+        .{
+            .stdin_behavior = .Pipe,
+            // TODO(Zig): ignoring stderr is broken in 0.11, fixed in 0.12:
+            //     https://github.com/ziglang/zig/pull/15565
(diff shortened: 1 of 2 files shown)


src/testing/tmp_tigerbeetle.zig
```diff
@@ -71,7 +71,12 @@ pub fn init(
 
     // Pass `--addresses=0` to let the OS pick a port for us.
     var process = try shell.spawn_options(
-        .{ .stdin_behavior = .Pipe, .stderr_behavior = .Ignore },
+        .{
+            .stdin_behavior = .Pipe,
+            // TODO(Zig): ignoring stderr is broken in 0.11, fixed in 0.12:
+            //     https://github.com/ziglang/zig/pull/15565
```
batiati, line 77: Did you try to indent the URL to match the `TODO:` colon alignment ?!?! 🤯

 ```suggestion
            // TODO(Zig): ignoring stderr is broken in 0.11, fixed in 0.12:
            // https://github.com/ziglang/zig/pull/15565
```

batiati (approved):
🎉

src/testing/tmp_tigerbeetle.zig
matklad: No, I just like to put URLs on a dedicated line and “align: center” them.

sort of how complicated and scary math equations are pulled in the center:

https://www.overleaf.com/learn/latex/Display_style_in_math_mode

PR #1483, opened 2024-01-26
cli: tigerbeetle benchmark

matklad:
`tigerbeetle benchmark` runs a standard benchmark on the current
hardware. It is built into the production binary, and is intended to be
used "in the field".

At the same time, this an internal command --- we won't give any
stability guarantees with respect to its output format. Similarly, we'll
adjust benchmarking options liberally as we go. The options aren't
documented yet for this reason.

The benchmark runs in two "modes":

- if the `--adresses` argument is passed, than the benchmark issues load
  against the specified cluster.
- without `--adresses` argument, the benchmark spins up a temporary
  single-node cluster on the current machine.


build.zig
```diff
@@ -128,7 +128,7 @@ pub fn build(b: *std.Build) !void {
         .optimize = mode,
     });
     if (mode == .ReleaseSafe) {
-        tigerbeetle.strip = true;
+        // tigerbeetle.strip = true;
```
cb22: Think this slipped in?

src/tigerbeetle/benchmark_driver.zig
```diff
@@ -0,0 +1,126 @@
...
+const ChildProcess = std.ChildProcess;
+
+const cli = @import("./cli.zig");
+const benchmark_load = @import("./benchmark_load.zig");
+
+const log = std.log;
+
+// Note: we intentionally don't use a temporary directory for this data file, and instead just put
+// it into CWD, as performance of TigerBeetle very much depends on a specific file system.
+const data_file = "0_0.tigerbeetle.benchmark";
```
cb22, line 26: Could we `rm` this file if we created it when we shutdown, so we could run `tigerbeetle benchmark` multiple times without intervention?

cb22: I guess we could actually just unlink it straight after spinning up the single node cluster? (although maybe that won't work on Windows)

src/tigerbeetle/benchmark_load.zig
```diff
@@ -57,41 +52,23 @@ const transfer_count_per_batch = @divExact(
...
+    addresses: []const std.net.Address,
+    cli_args: *const cli.Command.Benchmark,
+) !void {
     const stderr = std.io.getStdErr().writer();
 
     if (builtin.mode != .ReleaseSafe and builtin.mode != .ReleaseFast) {
-        try stderr.print("Benchmark must be built as ReleaseSafe for reasonable results.\n", .{});
+        try stderr.print("Benchmark must be built with '-Drelease' for reasonable results.\n", .{});
+    }
+    if (build_options.config_base != .production) {
```
cb22, line 65: :tada:

src/tigerbeetle/benchmark_driver.zig
matklad: I think we do remove this file?

```
    var data_file_created = false;
    defer {
        if (data_file_created) {
            std.fs.cwd().deleteFile(data_file) catch {};
        }
    }
```

The logic here is guarded by the if because we don't want to remove the file if the user passed in the `--addresses` cli flag.

This of course doesn't work if the benchmark is killed. Unliking file immediately after format is smart, but it's not _guaranteed_ to work (we could be killed _right_ after format), and won't work on windows.

Overall, I am not yet sure what's the best way to deal with this file! I am torn between:

- just adding a random suffix to the file, so, even if the file is not deleted, it doesn't prevent concurrent benchmarks from running
- in the opposite direction, requiring the user to pass in a path to a datafile, because _where_ the data file is certainly affects performance. We, eg, don't want to put it into `/tmp`, for example.

So, I figured I preserve existing logic, and then iterate separately.

build.zig
matklad: Good catch!

src/tigerbeetle/benchmark_driver.zig
cb22: Ah - I skipped over that! The reason I made the comment was because I was Ctrl-C'ing which I guess counts as being killed...

It'll be fairly common I think, and I don't think we need a 100% solution there - but happy to leave it up to a future iteration.

(I was partial to the random suffix in same folder idea, too)

cb22 (approved):

PR #1485, opened 2024-01-27
Revisiting the code!

batiati:
A collection of unrelated changes that were in my stash for a while, and I decided to submit them as a single PR.
To simplify the review process, each change has its own commit:

- Removed (almost) all uses of unnecessary `@as(...)` where:
a) The compiler can infer the type, such as return values, variables/fields assignment, and function parameters.
b) Expressions such as `const a = @as(T, ...);` could be refactored to `const a: T = ...;`

- Changed the immutable variables declared with `var` to `const`.
With a little help from the Zig compiler 0.12-dev;

- Upgraded the code for derived indexes to use Zig 0.11.0.
This code was never referenced, so it wasn't migrated during the last upgrade.

-  Added some misc `assert`s.

### DIFF
--- a/src/state_machine.zig
+++ b/src/state_machine.zig
@@ -872,7 +872,7 @@ pub fn StateMachineType(
                             var chain_index = chain_start_index;
                             while (chain_index < index) : (chain_index += 1) {
                                 results[count] = .{
-                                    .index = @as(u32, @intCast(chain_index)),
+                                    .index = @intCast(chain_index),
                                     .result = .linked_event_failed,
                                 };
                                 count += 1;
@@ -881,7 +881,7 @@ pub fn StateMachineType(
                             assert(result == .linked_event_failed or result == .linked_event_chain_open);
                         }
                     }
-                    results[count] = .{ .index = @as(u32, @intCast(index)), .result = result };
+                    results[count] = .{ .index = @intCast(index), .result = result };
                     count += 1;
                 }
                 if (chain != null and (!event.flags.linked or result == .linked_event_chain_open)) {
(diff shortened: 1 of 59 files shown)


src/testing/cluster.zig
```diff
@@ -358,10 +358,10 @@ pub fn ClusterType(comptime StateMachineType: anytype) type {
         pub fn restart_replica(cluster: *Self, replica_index: u8) !void {
             assert(cluster.replica_health[replica_index] == .down);
 
-            var nonce = cluster.replicas[replica_index].nonce + 1;
+            const nonce = cluster.replicas[replica_index].nonce + 1;
             // Pass the old replica's Time through to the new replica. It will continue to tick while
             // the replica is crashed, to ensure the clocks don't desynchronize too far to recover.
-            var time = cluster.replicas[replica_index].time;
+            const time = cluster.replicas[replica_index].time;
```
matklad: This one is intentionally `var`, to force the compiler to make a copy here and not do a pass-by-reference

batiati: We need to express this in the code (with a dummy mutation) otherwise, the compiler will catch it as an error when we upgrade to Zig 0.12

src/state_machine.zig
```diff
@@ -819,7 +820,7 @@ pub fn StateMachineType(
             input: []align(16) const u8,
             output: *align(16) [constants.message_body_size_max]u8,
         ) usize {
-            comptime assert(operation != .lookup_accounts and operation != .lookup_transfers);
+            comptime assert(operation == .create_accounts or operation == .create_transfers);
```
matklad, line 823: :heart:

src/tidy.zig
```diff
@@ -408,7 +408,6 @@ const naughty_list = [_][]const u8{
     "state_machine/auditor.zig",
     "state_machine/workload.zig",
     "testing/aof.zig",
-    "testing/cluster.zig",
```
matklad, line 411: :heart: :heart: :heart:

matklad (approved):

Issue #1484, opened 2024-01-27
Liveness: 15422833536424415004


tigerbeetle-vopr:
<strong>Commit:</strong> ```7b3a79c502070c6088cb55c3e537e7f894c1293c``` <br><br><strong>Branches:</strong> dj-vsr-view-change-truncate-after-checkpoint<br><br><strong>Duration to run seed in ReleaseSafe mode:</strong> 26s<br><br><strong>Debug Logs:</strong><br><details><summary>Tail</summary><br>
```debug: Replaying seed 15422833536424415004 in Debug mode with full debug logging enabled...


SEED=15422833536424415004

replicas=3
standbys=1
clients=4
request_probability=38%
idle_on_probability=16%
idle_off_probability=12%
one_way_delay_mean=3 ticks
one_way_delay_min=2 ticks
packet_loss_probability=26%
path_maximum_capacity=8 messages
path_clog_duration_mean=210 ticks
path_clog_probability=1%
packet_replay_probability=41%
partition_mode=testing.packet_simulator.PartitionMode.uniform_partition
partition_symmetry=testing.packet_simulator.PartitionSymmetry.asymmetric
partition_probability=1%
unpartition_probability=3%
partition_stability=125 ticks
unpartition_stability=14 ticks
read_latency_min=0
read_latency_mean=11
write_latency_min=0
write_latency_mean=64
read_fault_probability=5%
write_fault_probability=6%
crash_probability=0.00002%
crash_stability=558 ticks
restart_probability=0.0002%
restart_stability=40 ticks
 1/.1V0/__1/__1C0:__2Jo0/_0J!0:_31Wo <__0:__0>0Ga0G!0G?2/4Pp0/3Rq
 0\ .1V0/__1/__1C0:__1Jo0/_0J!0:_31Wo <__0:__0>0Ga0G!0G?
 2\.1V0/__1/__1C0:__1Jo0/_0J!0:_31Wo <__0:__0>0Ga0G!0G?
 3|.1V0/__1/__1C0:__2Jo0/_0J!0:_31Wo <__0:__0>0Ga0G!0G?
 1/.1V0/__2/__2C0:__4Jo0/_1J!0:_31Wo <__0:__0>0Ga0G!0G?3/4Pp0/3Rq
 0\ .1V0/__2/__2C0:__3Jo0/_0J!0:_31Wo <__0:__0>0Ga0G!0G?
[...]

Issue #1378, opened 2023-12-20
Allow primaries with a dirty journal


matklad:
Before state sync, we had an invariant that only a replica with a pristine log can become primary:

https://github.com/tigerbeetle/tigerbeetle/blob/b4caaa510ec37f6e9ad9227bd557dc6c59a008e2/src/vsr/replica.zig#L5480

After the state sync, it might be the case that it's impossible for the primary to repair its entire log, as the cluster state beyond the checkpoint is not available. We need to fix it:

* add replica_test tests around this area
* unify the logic between the primary and the backup and just never repair beyond checkpoint:
   https://github.com/tigerbeetle/tigerbeetle/blob/b4caaa510ec37f6e9ad9227bd557dc6c59a008e2/src/vsr/replica.zig#L5122
* allow primary to start view even if it has some dirty entries in the journal. commit_min=commit_max+pristine pipeline should be enough

sentientwaffle:
There is also "repair-min" logic in `Replica.on_repair_sync_timeout()` that will probably need a similar update.

A backup tries to infer the _primary's_ `op_repair_min` via `commit_max` and the currently known highest-canonical-checkpoint. Based on this repair-min, it decides whether or not it is "stuck" -- i.e., will be able to repair its WAL (this is ideal), or should it fall back to state sync.

PR #1493, opened 2024-01-29
Test: Primary with dirty log

sentientwaffle:
A primary with a dirty log cannot (currently) start a view with a dirty log. This is a liveness bug that requires no crashes, only partitions.

This is a "failing" test for https://github.com/tigerbeetle/tigerbeetle/issues/1378

See also https://github.com/tigerbeetle/tigerbeetle/issues/1484

### DIFF
--- a/src/vsr/replica_test.zig
+++ b/src/vsr/replica_test.zig
@@ -893,6 +893,54 @@ test "Cluster: view-change: duel of the primaries" {
     try expectEqual(t.replica(.R2).commit(), 21);
 }
 
+test "Cluster: view-change: primary with dirty log" {
+    const t = try TestContext.init(.{ .replica_count = 3 });
+    defer t.deinit();
+
+    var c = t.clients(0, t.cluster.clients.len);
+    try c.request(16, 16);
+    try expectEqual(t.replica(.R_).commit(), 16);
+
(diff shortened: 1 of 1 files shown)


src/vsr/replica_test.zig
```diff
@@ -893,6 +893,52 @@ test "Cluster: view-change: duel of the primaries" {
...
+    // B2 tries to become primary. (Don't let B1 become primary – it would not realize its
+    // checkpoint entry is corrupt, which would defeat the purpose of this test).
+    // B2 tries to repair (request_prepare) this corrupt op, even though it is before its
+    // checkpoint. B1 discovers that this op is corrupt, and marks it as faulty.
+    b1.corrupt(.{ .wal_prepare = checkpoint_2 % slot_count });
+    t.run();
+
+    try expectEqual(b1.status(), .view_change);
+    try expectEqual(b2.status(), .view_change);
+    // try expectEqual(b1.status(), .normal);
```
matklad, line 940: Let's add a comment pointing to the issue here: failing tests are _great_ imo, but they _should_ be clearly marked as "hey, that's not right!", in case they slip through the cracks and stay unfixed longer than what we envision.

matklad (approved):

PR #1497, opened 2024-01-30
Improve windows compat

matklad:
I feel like https://github.com/tigerbeetle/tigerbeetle/pull/1482 would take some time for me to debug properly, so I want to land these two improvements separately from that. Can be reviewed per commit. 

First commit is definitely required.

For the second one, I am not _entirely_ sure it is needed, as the CI still fails even with it applied, but using random file feels cleaner either way.


src/clients/dotnet/TigerBeetle.Tests/IntegrationTests.cs
```diff
@@ -1509,21 +1509,20 @@ internal class TBServer : IDisposable
     private const string TB_EXE = "tigerbeetle";
     private const string TB_FILE = "dotnet-tests.tigerbeetle";
```
batiati: Can we delete this `TB_FILE` constant?

matklad: Ah, indeed we should, thanks!

batiati (approved):

PR #1499, opened 2024-01-30
tidy: io/linux.zig

matklad:


src/tidy.zig
```diff
@@ -352,12 +350,9 @@ fn find_long_line(file_text: []const u8) !?usize {
...
 
-/// Heuristically checks if a `line` is a comment with an URL.
-fn is_url(line: []const u8) bool {
-    const cut = stdx.cut(line, "// https://") orelse return false;
-    for (cut.prefix) |c| if (!(c == ' ' or c == '/')) return false;
-    for (cut.suffix) |c| if (c == ' ') return false;
-    return true;
+/// Heuristically checks if a `line` contains an URL.
+fn has_link(line: []const u8) bool {
+    return std.mem.indexOf(u8, line, "https://") != null;
```
matklad, line 355: Intentionally _not_ checking for `http` here, as it seems a good idea to stick with https if possible.

kprotty, line 355: random nit: might want to add `or contains("http://")`

kprotty (approved):

PR #1500, opened 2024-01-30
Test: Simplify "forfeit" replica_test

sentientwaffle:
I'm not sure why I included the extra steps when originally writing this test!

I can see _what_ I was doing: past-me apparently wanted the checkpoint to be committed atop, but not by either of the replicas in the second phase of the test. But that doesn't seem to be relevant to the on-do-view-change/forfeit logic. As far as present-me can tell, the important thing is only that the checkpoint is prepared atop, but that the cluster is not so far ahead of the lagging replica as to necessitate state sync.

I verified that (both before and after this change) the `"on_do_view_change: lagging primary; forfeiting"` log line is hit.

---

(Also add a new rule to the test, to be extra sure that this test doesn't cheat by using state sync.)

### DIFF
--- a/src/vsr/replica_test.zig
+++ b/src/vsr/replica_test.zig
@@ -467,20 +467,13 @@ test "Cluster: repair: view-change, new-primary lagging behind checkpoint, forfe
 
     b1.drop_all(.__, .bidirectional);
 
-    try c.request(checkpoint_1_trigger, checkpoint_1_trigger);
-    try expectEqual(a0.commit(), checkpoint_1_trigger);
-    try expectEqual(b1.commit(), 20);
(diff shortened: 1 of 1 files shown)


sentientwaffle:
As an aside: The reason I've been focusing on this test is that with prepare-beyond-checkpoint, this test will need to prepare to "border + 1" instead of "trigger + 1" to trigger the forfeit.

Since prepare headers carry `commit` numbers, `Ensure that B2 does not commit op=trigger+1.` isn't testable (with our current replica_test helpers, that is). (But as this PR shows, it doesn't need to.)

matklad (approved):

PR #1501, opened 2024-01-30
VSR: Use previous `checkpoint_id` in prepares' headers

sentientwaffle:
### Background

Right now, a replica will not prepare (or even queue) requests whose op would extend beyond their next checkpoint's trigger op.

This is problematic for performance (the "checkpoint latency spike"), since the replica enters the new checkpoint with nothing to commit.
And the "latency spike" is farther exacerbated by the fact that clients will back off retrying their requests.

Originally we didn't start preparing past the checkpoint to guard against overwriting WAL entries before they will definitely not be needed again. But thanks to `vsr_checkpoint_interval`, that reason does not apply.

However, prepare headers include a `checkpoint_id` ([motivation here](https://github.com/tigerbeetle/tigerbeetle/blob/30cfbfa2eca94b8cd0b1d2a8ce41c7e7720128f0/src/vsr/message_header.zig#L531-L539)). That checkpoint id isn't available until the checkpoint trigger prepare commits.

### Fix

~The first `constants.pipeline_prepare_queue_max` prepares that follow a checkpoint trigger op will now contain the _previous_ checkpoint's id, since that is available both before and after that checkpoint trigger commits.~

~These are called the "border" prepares – they may be prepared during either checkpoint. (This part is not implemented in this commit, though!)~

All prepares within a checkpoint include the `checkpoint_id` of the _previous_ checkpoint. The transition occurs at the checkpoint op.

### DIFF
--- a/src/vsr/replica.zig
+++ b/src/vsr/replica.zig
@@ -5021,9 +5021,10 @@ pub fn ReplicaType(
 
         /// Returns checkpoint id associated with the op.
         ///
-        /// Normally, this is just the id of the op's previous checkpoint. However, ops
-        /// between a checkpoint and its trigger can't know checkpoint's id yet, and instead use
-        /// the id of the grandparent checkpoint.
+        /// Specifically, returns the checkpoint id corresponding to the checkpoint with:
+        ///
+        ///   prepare.op > checkpoint_op
+        ///   prepare.op ≤ checkpoint_after(checkpoint_op)
         ///
         /// Returns `null` for ops which are too far in the past/future to know their checkpoint
         /// ids.
--- a/src/vsr/replica.zig
+++ b/src/vsr/replica.zig
@@ -1408,7 +1408,7 @@ pub fn ReplicaType(
             }
 
             if (message.header.checkpoint_id != self.superblock.working.checkpoint_id() and
-                message.header.checkpoint_id != self.superblock.working.vsr_state.checkpoint.previous_checkpoint_id)
+                message.header.checkpoint_id != self.superblock.working.vsr_state.checkpoint.parent_checkpoint_id)
             {
                 // Panic on encountering a prepare which does not match our own checkpoint id.
                 // (Or the previous checkpoint id, for border prepares.)
@@ -5042,7 +5042,7 @@ pub fn ReplicaType(
                         return null;
                     }
                     // Case 3: op is from the previous checkpoint whose id we still remember.
-                    return self.superblock.working.vsr_state.checkpoint.previous_checkpoint_id;
+                    return self.superblock.working.vsr_state.checkpoint.parent_checkpoint_id;
                 }
 
                 assert(op + constants.vsr_checkpoint_interval > self.op_checkpoint_next_border());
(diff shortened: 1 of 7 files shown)


matklad:
Joran floated an interesting idea during our 1-1 today: what if we don't do border prepares, and instead include checkpoint id of the previous checkpoint for all prepares, so that we don't have a special case?

I think this not only removes special case, but simplifies the mental model overall somewhat: prepares before the first trigger are not special. All prepares in a single checkpoint have the same checkpoint id (that of the previous checkpoint).

src/vsr/superblock.zig
```diff
@@ -812,9 +818,15 @@ pub fn SuperBlockType(comptime Storage: type) type {
             assert((update.storage_size == data_file_size_min) ==
                 update.free_set_reference.empty());
 
+            // NOTE: Within the vsr_state.checkpoint assignment below, do not read from vsr_state
+            // directly. A miscompilation bug (as of Zig 0.11.0) causes fields to receive the
+            // incorrect values.
```
sentientwaffle, line 823: (Not sure if this is technically a 'miscompilation' or just underspecification, but it is surprising and worth a note either way.)

src/vsr/replica.zig
```diff
@@ -5013,35 +5014,41 @@ pub fn ReplicaType(
...
+        /// Returns the highest op that this replica can safely prepare to its WAL.
+        fn op_checkpoint_next_border(self: *const Self) u64 {
+            return vsr.Checkpoint.border_for_checkpoint(self.op_checkpoint_next()).?;
+        }
+
         /// Returns checkpoint id associated with the op.
         ///
-        /// Normally, this is just the id of the checkpoint the op builds on top. However, ops
+        /// Normally, this is just the id of the op's previous checkpoint. However, ops
         /// between a checkpoint and its trigger can't know checkpoint's id yet, and instead use
```
matklad: We don't have trigger special case anymore.

src/vsr/replica.zig
```diff
@@ -5013,35 +5014,41 @@ pub fn ReplicaType(
             return vsr.Checkpoint.trigger_for_checkpoint(self.op_checkpoint_next()).?;
         }
 
+        /// Returns the highest op that this replica can safely prepare to its WAL.
+        fn op_checkpoint_next_border(self: *const Self) u64 {
```
matklad, line 5018: Now that we dont' have "border prepares", the "border" name doesn't seem really great, though I can't suggest a better alternative...

Maybe:

- trigger --- the `op` when the checkpoints starts
- due --- the `op` by which the checkpoint should have completed

Or maybe even _not_ introduce any name here? 


```
self.op_checkpoint_next_trigger() + constants.pipeline_prepare_queue_max
```

actually reads great to me: trigger, plus whatever we can pipeline over the trigger.

sentientwaffle: In the next PR in the sequence (which actually enables "prepare beyond the checkpoint") `op_checkpoint_next_border` is used _very_ frequently, so imo having it in a helper function rather than inlining `self.op_checkpoint_next_trigger() + constants.pipeline_prepare_queue_max` is quite useful.

It is used in ways that "trigger" is often used right now -- "the highest op that we will accept in our WAL". So I don't think "trigger" or "due" make sense. Maybe "limit" or "ceiling"?

matklad: :thinking: yeah, `due` has inverted semantics here -- checkpoint is indeed due to be done by the time we accept that `op` into our WAL, _but_, in code, we don't block until checkpoint is ready, we  rather drop prepares. 

Can't say that `limit` or `ceiling` is much better than the border. Hm, maybe

```
op_checkpoint_pipeline_limit
```

? I think "pipeline" is key here. Maybe even just 


```
op_pipeline_limit
```

? The fact that it depends on the checkpoint is sort of an implementation detail I think?

Though, no strong opinion either way, we can live with `border_prepare` as well .

matklad (approved):

src/vsr/replica.zig
sentientwaffle: :+1: I'm going to leave it as "border" at least for now. The fact that it is a term that is not used elsewhere in the code (so can't be confused with anything else) makes up for the fact that it is not self-explaining.

PR #1504, opened 2024-01-31
ci: retry downloads of .Net docker images

matklad:
Downloading these images fails fairly frequently on GitHub actions, let's retry for a bit.


batiati (approved):
What would it cost for `docker pull` to have a `--timeout` parameter 🤔

PR #1506, opened 2024-02-01
Extend tests to mutable ring buffer

sm2n:


src/ring_buffer.zig
```diff
@@ -276,18 +273,31 @@ const testing = std.testing;
...
     const ring_index = ring.index;
 
-    var loops: usize = 0;
-    while (loops < 2) : (loops += 1) {
-        var iterator = ring.iterator();
-        var index: usize = 0;
-        while (iterator.next()) |item| {
-            try testing.expectEqual(values[index], item);
-            index += 1;
+    inline for (.{ .immutable, .mutable }) |mutability| {
```
matklad, line 276: For the mutable case, let's maybe also mutate through the pointer, and check that iterating the second time around reflects the mutations.

matklad: Although, this _is_ a clear improvement, so I am fine with landing this as is, and testing updates separately.

matklad (dismissed):

src/ring_buffer.zig
sm2n: Done!

src/ring_buffer.zig
```diff
@@ -276,18 +273,46 @@ const testing = std.testing;
...
+        var loops: usize = 0;
+        while (loops < 2) : (loops += 1) {
```
matklad: While we are at it, let's also use a `for` loop here. And, given that we already do _two_ passes over the data, can we re-use this loopig for mutability testing? That is, to do something along the lines of 

```zig
                .immutable => while (iterator.next()) |item| {
                    try testing.expectEqual(values[index], if (loop == 0) values[index] else values[index] +% permutation);
                    index += 1;
                    item.* +%= permutation.
                },
```

sm2n: If we did that, it would break the other tests, since the mutable iterator tests would have different observable behavior than the immutable ones. Good catch on the for loop, will change.

src/ring_buffer.zig
```diff
@@ -276,18 +273,45 @@ const testing = std.testing;
...
 
-    var loops: usize = 0;
-    while (loops < 2) : (loops += 1) {
-        var iterator = ring.iterator();
-        var index: usize = 0;
-        while (iterator.next()) |item| {
-            try testing.expectEqual(values[index], item);
-            index += 1;
+    inline for (.{ .immutable, .mutable }) |mutability| {
+        for (0..1) |_| {
```
matklad: This should be `0..2`, no?

sm2n: yeah..., whoops

matklad (approved):

PR #1508, opened 2024-02-01
VSR: Prepare beyond checkpoint

sentientwaffle:
Prepare (or queue) requests past checkpoint boundaries.

### Background

Quoting from the [previous border-prepare PR](https://github.com/tigerbeetle/tigerbeetle/pull/1501):

> Right now, a replica will not prepare (or even queue) requests whose op would extend beyond their next checkpoint's trigger op.
> 
> This is problematic for performance (the "checkpoint latency spike"), since the replica enters the new checkpoint with nothing to commit.
> And the "latency spike" is farther exacerbated by the fact that clients will back off retrying their requests.
> 
> Originally we didn't start preparing past the checkpoint to guard against overwriting WAL entries before they will definitely not be needed again. But thanks to `vsr_checkpoint_interval`, that reason does not apply.

### Fix

The replica is now able to:
- begin preparing (at most `pipeline_prepare_queue_max`) requests, and
- queue requests (if its prepare queue is full)

...before a checkpoint begins, or while it is in progress.

### Performance

Note that because the benchmark runs with only a single client, this PR's performance improvement will not be visible.
(Even if it had multiple clients, it is unlikely to be visible because the benchmark checkpoints very infrequently).

### DIFF
--- a/src/vsr/replica.zig
+++ b/src/vsr/replica.zig
@@ -274,7 +274,7 @@ pub fn ReplicaType(
         /// * `replica.op - replica.commit_min    ≤ journal_slot_count`
         /// * `replica.op - replica.op_checkpoint ≤ journal_slot_count`
         ///   It is safe to overwrite `op_checkpoint` itself.
-        /// * `replica.op ≤ replica.op_checkpoint_next_border`:
+        /// * `replica.op ≤ replica.op_prepare_max`:
         ///   Don't wrap the WAL until we are sure that the overwritten entry will not be required
         ///   for recovery.
         op: u64,
@@ -533,7 +533,7 @@ pub fn ReplicaType(
             const vsr_headers = self.superblock.working.vsr_headers();
             // If we were a lagging backup that installed an SV but didn't finish fast-forwarding,
             // the vsr_headers head op may be part of the checkpoint after this one.
-            maybe(vsr_headers.slice[0].op > self.op_checkpoint_next_border());
+            maybe(vsr_headers.slice[0].op > self.op_prepare_max());
 
             // Given on-disk state, try to recover the head op after a restart.
             //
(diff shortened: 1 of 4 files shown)


matklad:
Another idea for `op_checkpoint_next_border`: `op_prepare_max`, analogous to `op_repair_max`.

matklad:
I _think_ we should also add an assert relating `checkpoint_interval` and `pipeline_prepare_queue_max` to constants.zig. In particular, we should statically forbid pipelines longer than checkpoint intervals, so that, in the `else` branch of 

```zig
            const checkpoint_id = checkpoint_id: {
                if (self.op + 1 <= self.op_checkpoint_next()) {
                    break :checkpoint_id self.superblock.working.vsr_state.checkpoint.parent_checkpoint_id;
                } else {
                    break :checkpoint_id self.superblock.working.checkpoint_id();
                }
            };
```

we indeed pick the correct checkpoint. I wonder if that code should call to `checkpoint_id_for_op().?` even, to avoid duplicating the logic? Although, duplicating might actually be desirable from the perspective of pairing asserts, specifically, paring assignment of checkpoint id by the primary with verification of checkpoint id by a backup.

src/vsr/replica.zig
```diff
@@ -1815,7 +1815,7 @@ pub fn ReplicaType(
             const op_checkpoint_max =
                 DVCQuorum.op_checkpoint_max(self.do_view_change_from_all_replicas);
             if (op_checkpoint_max > self.op_checkpoint() and
-                op_head > self.op_checkpoint_next_trigger())
+                op_head > self.op_checkpoint_next_border())
```
matklad: comment bellow still talks about the trigger.

sentientwaffle:
> I wonder if that code should call to checkpoint_id_for_op().? even, to avoid duplicating the logic? Although, duplicating might actually be desirable from the perspective of pairing asserts, specifically, paring assignment of checkpoint id by the primary with verification of checkpoint id by a backup.

Yeah, imo computing it separately is actually preferable for that reason.

matklad (approved):

PR #1509, opened 2024-02-02
VSR: Perform checkpoint steps concurrently

sentientwaffle:
Instead of performing the checkpoint steps sequentially, perform them concurrently:

- `checkpoint_state_machine`
- `checkpoint_client_replies`
- `checkpoint_client_sessions`

This PR doesn't make `checkpoint_grid` concurrent -- that will be in a follow-up PR since it is slightly more involved.

### DIFF
--- a/src/lsm/forest.zig
+++ b/src/lsm/forest.zig
@@ -438,7 +438,6 @@ pub fn ForestType(comptime _Storage: type, comptime groove_cfg: anytype) type {
             const forest = @fieldParentPtr(Forest, "manifest_log", manifest_log);
             assert(forest.progress.? == .checkpoint);
             assert(forest.manifest_log_progress == .idle);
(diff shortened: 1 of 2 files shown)


src/lsm/forest.zig
```diff
@@ -438,7 +438,6 @@ pub fn ForestType(comptime _Storage: type, comptime groove_cfg: anytype) type {
             const forest = @fieldParentPtr(Forest, "manifest_log", manifest_log);
             assert(forest.progress.? == .checkpoint);
             assert(forest.manifest_log_progress == .idle);
-            forest.grid.assert_only_repairing();
```
sentientwaffle, line 441: There may (now) be non-repair grid-writes while we are flushing the manifest log blocks – we are concurrently flushing the trailers at the same time.

matklad (approved):

PR #1512, opened 2024-02-02
Grid: Zero block padding, but not the entire block

sentientwaffle:
### Background

Previously when starting with a "fresh" grid block (e.g. before constructing a new block for writing), the fresh block was always zeroed.

This is nice in that it guards against buffer-bleeds. But zeroing the entire block (in `grid.zig`) is costly in terms of memory bandwidth.

### Fix

- In `grid.zig` zero the block's header (to guard against accidental reuse)...
- ...But don't zero the _entire_ block buffer.
- Before writing a block, zero its last sector's padding (since that last sector will reach the disk). (The `StorageChecker` verifies that the last sector's padding is zeroed.)

### DIFF
--- a/src/lsm/schema.zig
+++ b/src/lsm/schema.zig
@@ -164,11 +164,16 @@ pub const TableIndex = struct {
         assert(header.snapshot > 0);
 
         const header_metadata = metadata(index_block);
-
-        return TableIndex.init(.{
+        const index = TableIndex.init(.{
             .key_size = header_metadata.key_size,
             .data_block_count_max = header_metadata.data_block_count_max,
         });
+
+        for (index.padding(index_block)) |padding_area| {
+            assert(stdx.zeroed(index_block[padding_area.start..padding_area.end]));
+        }
+
+        return index;
     }
 
     pub fn metadata(index_block: BlockPtrConst) *const Metadata {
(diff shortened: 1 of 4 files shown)


src/lsm/schema.zig
```diff
@@ -164,11 +164,16 @@ pub const TableIndex = struct {
...
-
-        return TableIndex.init(.{
+        const index = TableIndex.init(.{
             .key_size = header_metadata.key_size,
             .data_block_count_max = header_metadata.data_block_count_max,
         });
+
+        for (index.padding(index_block)) |padding_area| {
+            assert(stdx.zeroed(index_block[padding_area.start..padding_area.end]));
+        }
```
sentientwaffle, line 174: Afaict we weren't asserting that these padding's were zero anywere!

src/vsr/grid.zig
```diff
@@ -685,6 +685,9 @@ pub fn GridType(comptime Storage: type) type {
                 }
             }
 
+            // Zero sector padding.
+            @memset(block.*[header.size..vsr.sector_ceil(header.size)], 0);
```
matklad, line 689: Maybe also assert, in `write_block_with`, that the padding is zeroed?

matklad:
Excellent trick with zeroing just the header btw! Gives defense-in-depth without affecting performance!

matklad (approved):

PR #1514, opened 2024-02-03
[Experimental] Add support to INCLUDE indexes

batiati:
## EXPERIMENTAL - It's not supposed to be merged

**Background:**

While prototyping the support for **historical account balances**, the idea of having _INCLUDED VALUES_ into the index was quite appealing:

- They're orthogonal with existing indexes. We can still use indexes to find the Groove Object, whether or not they have an included part. With the additional capacity of indexes having extra values that can save further lookups.

- They add less size to the Tree since they piggyback on the existing indexes (in comparison with creating a new Tree containing the key + fields)

- It's such an elegant solution to storing the history of account balances as included values for the existing indexes `Transfer.debit_account_id` and `Transfer.credit_account_id`.
  With the big advantage of sharing the same logic for searching all transfers or all balances for a given account).

- It's a nice primitive that can be useful for future use cases.

## So what went wrong?

Once we define that an index has an INCLUDE part, it changes the `Tree.Value` in compile-time.

It means that we can't have the account history as an **opt-in feature**.
If we store it as included values in the `debit_account_id` and `credit_account_id`, we're always paying for it!

The alternative would be, instead of using the existing indexes, creating two new indexes + include that are only inserted if the history is enabled for the account.

It would work fine, but we also would lose half of the benefits listed above!!!

So, let's this **PR archived for reference** while we try other approaches.
Benchmark shows the cost of having the history always enabled:

<table>
<tr>
<th>Before</th>
<th>After</th>
</tr>
<tr><td>
<code>1252 batches in 40.29 s
load offered = 1000000 tx/s
load accepted = 248204 tx/s
batch latency p1 = 0 ms
batch latency p10 = 6 ms
batch latency p20 = 8 ms
batch latency p30 = 8 ms
batch latency p40 = 9 ms
batch latency p50 = 15 ms
batch latency p60 = 20 ms
batch latency p70 = 24 ms
batch latency p80 = 32 ms
batch latency p90 = 38 ms
batch latency p95 = 48 ms
batch latency p99 = 501 ms
batch latency p100 = 597 ms
</code>
 </td>
<td>
<code>1255 batches in 64.96 s
load offered = 1000000 tx/s
load accepted = 153929 tx/s
batch latency p1 = 0 ms
batch latency p10 = 8 ms
batch latency p20 = 9 ms
batch latency p30 = 11 ms
batch latency p40 = 13 ms
batch latency p50 = 16 ms
batch latency p60 = 20 ms
batch latency p70 = 24 ms
batch latency p80 = 29 ms
batch latency p90 = 36 ms
batch latency p95 = 57 ms
batch latency p99 = 1123 ms
batch latency p100 = 1623 ms
</code>
 </td> </tr>
</table>

### DIFF
--- a/src/lsm/groove.zig
+++ b/src/lsm/groove.zig
@@ -11,6 +11,8 @@ const TableType = @import("table.zig").TableType;
 const TreeType = @import("tree.zig").TreeType;
 const GridType = @import("../vsr/grid.zig").GridType;
 const CompositeKeyType = @import("composite_key.zig").CompositeKeyType;
+const is_composite_key = @import("composite_key.zig").is_composite_key;
+
 const NodePool = @import("node_pool.zig").NodePool(constants.lsm_manifest_node_size, 16);
 const CacheMapType = @import("cache_map.zig").CacheMapType;
 const ScopeCloseMode = @import("tree.zig").ScopeCloseMode;
@@ -133,6 +135,69 @@ fn IndexTreeType(
     return TreeType(Table, Storage);
 }
 
+// FIXME Maybe having just one type for both IndexTree is better,
+// if `Include == void` it's the same cost as having `Key == Value`.
+fn IndexWithIncludeTreeType(
+    comptime Storage: type,
+    comptime CompositeKey_: type,
+    comptime Include: type,
+    comptime value_count_max: usize,
+) type {
+    // The schema requires the size to be a power of two,
+    // so we may need to add a pad to the next power of two greater than the current size.
+    const pad = pad: {
+        const value_size = @sizeOf(CompositeKey_) + @sizeOf(Include);
+        const expected_size = 1 << std.math.log2_int_ceil(usize, value_size);
+        assert(expected_size >= value_size);
+
+        const pad_size = expected_size - value_size;
+        break :pad [_]u8{0} ** pad_size;
+    };
+
+    const TableValue = extern struct {
+        const TableValue = @This();
+        pub const CompositeKey = CompositeKey_;
+
+        key: CompositeKey,
+        include: Include,
+        reserved: @TypeOf(pad) = pad,
+
+        inline fn key_from_value(self: *const TableValue) CompositeKey.Key {
+            return self.key.key_from_value();
+        }
+
+        inline fn tombstone(self: *const TableValue) bool {
+            return self.key.tombstone();
+        }
+
+        inline fn tombstone_from_key(key: CompositeKey.Key) TableValue {
+            return .{
+                .key = CompositeKey.tombstone_from_key(key),
+                .include = std.mem.zeroInit(Include, .{}),
+            };
+        }
+
+        comptime {
+            assert(stdx.no_padding(TableValue));
+            assert(std.math.isPowerOfTwo(@sizeOf(TableValue)));
+            assert(@alignOf(TableValue) == 16);
+        }
+    };
+
+    const Table = TableType(
+        CompositeKey_.Key,
+        TableValue,
+        TableValue.key_from_value,
+        CompositeKey_.sentinel_key,
+        TableValue.tombstone,
+        TableValue.tombstone_from_key,
+        value_count_max,
+        .include_index,
+    );
+
+    return TreeType(Table, Storage);
+}
+
 /// A Groove is a collection of LSM trees auto generated for fields on a struct type
 /// as well as custom derived fields from said struct type.
 pub fn GrooveType(
@@ -154,6 +219,10 @@ pub fn GrooveType(
     /// - derived: { .field = *const fn (*const Object) DerivedType }:
     ///     An anonymous struct which contain fields that don't exist on the Object
     ///     but can be derived from an Object instance using the field's corresponding function.
+    ///
+    /// - include: { .field = *const fn (context: ?*const anyopaque, *const Object) IncludeType }:
+    ///     An anonymous struct which contain fields or derived fields and provide additional values
+    ///     to be included alongside the index.
     comptime groove_options: anytype,
 ) type {
     @setEvalBranchQuota(64000);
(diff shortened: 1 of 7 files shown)


batiati:
> So, let's this PR archived for reference while we try other approaches.

**TLDR**, go direct to the item 4 😊

## 1. Include indexes

Resuming the strategy previously proposed to use include indexes, keeping it as an opt-in feature:
> The alternative would be, instead of using the existing indexes, creating two new indexes + include that are only inserted if the history is enabled for the account.

Please refer to the [top comment](https://github.com/tigerbeetle/tigerbeetle/pull/1514#issue-2116637241) for background.

Usage: https://github.com/tigerbeetle/tigerbeetle/blob/b4586e0d7d04beab1a345455bdd3664a32840998/src/state_machine.zig#L234-L300

Full code [here](https://github.com/tigerbeetle/tigerbeetle/compare/main...batiati-include)

**Pros:**

- It requires almost no changes to the Scan API.

**Cons:**

- The machinery for creating include types is too complex.
- Poor performance.

**Performance:**

```
1241 batches in 70.51 s
load offered = 1000000 tx/s
load accepted = 141833 tx/s
batch latency p1 = 2 ms
batch latency p10 = 8 ms
batch latency p20 = 8 ms
batch latency p30 = 11 ms
batch latency p40 = 16 ms
batch latency p50 = 20 ms
batch latency p60 = 23 ms
batch latency p70 = 27 ms
batch latency p80 = 31 ms
batch latency p90 = 37 ms
batch latency p95 = 56 ms
batch latency p99 = 1199 ms
batch latency p100 = 1934 ms
```

Data size after the benchmark: 30,7GiB

## 2. Allow objects keyed by the composite key `parent_timestamp + timestamp`

The initial idea was to allow objects keyed by arbitrary user-defined fields, but it would require the StateMachine to deal with complex range queries for 1:N scenarios (e.g. correctly extracting the key prefix). Instead, we can allow only the combination of parent_timestamp + timestamp as an alternative key, providing helper functions for querying by the parent's timestamp.

Usage: https://github.com/tigerbeetle/tigerbeetle/blob/0bb5b4531d45e5342bf88e002af6a7a5c9ad42de/src/state_machine.zig#L1122-L1134

Full code [here](https://github.com/tigerbeetle/tigerbeetle/compare/main...batiati-composite-keys)

**Pros:**

- It's easy to understand, we can insert both the credit and the debit `AccountBalance` into the same Tree.

- It's simple to query by the parent timestamp, requiring a few changes to the ScanAPI.

**Cons:**

- We can't directly query by only the debit or the credit side.
  It would require extra steps to match 1:1 the `get_account_transfers` API.

- Poor performance

**Performance:**

```
1229 batches in 69.74 s
load offered = 1000000 tx/s
load accepted = 143388 tx/s
batch latency p1 = 7 ms
batch latency p10 = 9 ms
batch latency p20 = 12 ms
batch latency p30 = 13 ms
batch latency p40 = 15 ms
batch latency p50 = 19 ms
batch latency p60 = 22 ms
batch latency p70 = 25 ms
batch latency p80 = 29 ms
batch latency p90 = 34 ms
batch latency p95 = 43 ms
batch latency p99 = 1278 ms
batch latency p100 = 1359 ms
```

Data file size: 30.9GiB

## 3. Two separate `Groove`s for the debit and the credit side, keyed by the Transfer's timestamp

The idea was to implement the account history using only the features we already have in the Groove type.
It's the fastest solution because it's only keyed by the timestamp, inserts the history only for accounts with the flag set, and does not add any new index.

For queries, we need to scan the existing indexes on `Transfer.debit_account_id` and `Transfer.credit_account_id` and then perform a lookup in each groove by the transfer's timestamp. It's challenging because we were not planning to add support to cross-grooves queries anytime soon, and this specific kind of query feels too much "ad-hoc" to justify such complex feature.

Usage: https://github.com/tigerbeetle/tigerbeetle/blob/97b1a55b5da041121075007f1ce3bc32ec70fca3/src/state_machine.zig#L1143-L1161

Full code [here](https://github.com/tigerbeetle/tigerbeetle/compare/main...batiati-history-as-two-grooves)

**Pros:**

- Requires no code changes to the Groove.
- Good performance

**Cons:**

- Complex to query.

**Performance:**

```
1247 batches in 41.82 s
load offered = 1000000 tx/s
load accepted = 239100 tx/s
batch latency p1 = 1 ms
batch latency p10 = 7 ms
batch latency p20 = 12 ms
batch latency p30 = 12 ms
batch latency p40 = 13 ms
batch latency p50 = 17 ms
batch latency p60 = 20 ms
batch latency p70 = 24 ms
batch latency p80 = 30 ms
batch latency p90 = 42 ms
batch latency p95 = 49 ms
batch latency p99 = 493 ms
batch latency p100 = 534 ms
```

Data file: 17GiB

## 4. A single `Groove` for both the debit and the credit side, keyed by the Transfer's timestamp

Like the previous one, this solution uses only the features we already have in the Groove type, storing both the debit and credit account balance into the same Groove, keyed by the timestamp.
For queries, we also rely on the existing indexes `Transfer.debit_account_id` and `Transfer.credit_account_id`.

Usage: https://github.com/tigerbeetle/tigerbeetle/blob/9ded136b631193100a273a741dfe4de67dd0a39f/src/state_machine.zig#L1144-L1166

Full code [here](https://github.com/tigerbeetle/tigerbeetle/compare/main...batiati-account-history)

**Pros:**

- Simple design
- Easy to query
- Good performance

**Cons:**

- We need to store zeroed fields for the account (dr or cr) without the flag set.
- The Groove object is 256 bytes long.
- We need to process the object before returning the results to the user.

**Performance:**

```
1248 batches in 42.22 s
load offered = 1000000 tx/s
load accepted = 236871 tx/s
batch latency p1 = 1 ms
batch latency p10 = 7 ms
batch latency p20 = 11 ms
batch latency p30 = 12 ms
batch latency p40 = 15 ms
batch latency p50 = 18 ms
batch latency p60 = 22 ms
batch latency p70 = 25 ms
batch latency p80 = 29 ms
batch latency p90 = 37 ms
batch latency p95 = 43 ms
batch latency p99 = 506 ms
batch latency p100 = 625 ms
```

Data file: 17GiB

batiati:
Closed in favor of #1526

PR #1519, opened 2024-02-05
VSR: Perform checkpoint-grid step concurrently 

sentientwaffle:
Perform the grid checkpoint (write freeset trailer + wait for repairs to finish) step concurrently with:

- `checkpoint_state_machine`
- `checkpoint_client_replies`
- `checkpoint_client_sessions`

...instead of sequentially.

The benchmark checkpoints so infrequently that I do not expect any performance improvement to be visible in the TPS.

---

Also note the second commit in this PR -- the first commit shuffled the results of a race such that the test fails, so the second commit fixes the race.

### DIFF
--- a/src/vsr/replica.zig
+++ b/src/vsr/replica.zig
@@ -59,7 +59,6 @@ const CommitStage = union(enum) {
     setup_client_replies,
     compact_state_machine,
     checkpoint_data: CheckpointDataProgress,
-    checkpoint_grid,
     checkpoint_superblock,
     /// A commit just finished. Clean up before proceeding to the next.
     cleanup,
@@ -69,6 +68,7 @@ const CheckpointData = enum {
     state_machine,
     client_replies,
     client_sessions,
+    grid,
 };
 
 const CheckpointDataProgress = std.enums.EnumSet(CheckpointData);
@@ -2793,7 +2793,7 @@ pub fn ReplicaType(
             // we would need to wait for it before sync starts anyhow, and the newer
             // checkpoint might sidestep the need for sync anyhow.
             if (self.commit_stage == .checkpoint_superblock) return;
-            if (self.commit_stage == .checkpoint_grid) return;
+            if (self.commit_stage == .checkpoint_data) return;
 
             // TODO Test connectivity to cluster to rule out a network partition.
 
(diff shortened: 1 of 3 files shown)


src/vsr/replica.zig
```diff
@@ -3148,8 +3148,10 @@ pub fn ReplicaType(
                     self.state_machine.checkpoint(commit_op_checkpoint_state_machine_callback);
                     self.client_sessions_checkpoint.checkpoint(commit_op_checkpoint_client_sessions_callback);
                     self.client_replies.checkpoint(commit_op_checkpoint_client_replies_callback);
+                    // The grid checkpoint must begin after the manifest/trailers have acquired all
+                    // their blocks, since it encodes the free set:
+                    self.grid.checkpoint(commit_op_checkpoint_grid_callback);
```
matklad, line 3153: If we do this _first_, is there any assert that can catch us? Basically, that no reservations can be created while the free set is being checkpointed?

I wonder if something like 

```diff
diff --git a/src/vsr/grid.zig b/src/vsr/grid.zig
index d86d33bf..3bc8660e 100644
--- a/src/vsr/grid.zig
+++ b/src/vsr/grid.zig
@@ -334,6 +334,7 @@ pub fn GridType(comptime Storage: type) type {
 
             {
                 assert(grid.free_set.count_reservations() == 0);
+                grid.free_set.opened = false;
 
                 grid.free_set.include_staging();
                 defer grid.free_set.exclude_staging();
@@ -361,6 +362,7 @@ pub fn GridType(comptime Storage: type) type {
         fn checkpoint_free_set_callback(set: *CheckpointTrailer) void {
             const grid = @fieldParentPtr(Grid, "free_set_checkpoint", set);
             assert(grid.callback == .checkpoint);
+            grid.free_set.opened = true;
             grid.checkpoint_join();
         }
 ```

could do the trick?

sentientwaffle: This is already guarded against via the `grid.callback == .none` assertions in `Grid.{reserve,acquire}`.

matklad: Ah, right, I was looking at `FreeSet.{reserve,acquire}`!

matklad (approved):

PR #1520, opened 2024-02-05
Do not insert null indexes

batiati:
We were inserting indexes even for zeroed fields. It adds considerable extra work and storage space for optional fields (for example, the `user_data_*` ones).

This PR also removes the support for nullable types from the derived indexes, unifying the ZERO == null semantics for all the secondary indexes.

**EDIT:** The benchmark shows only ~5% performance gain because effectively only `pending_transfer_id` is no longer inserted for the benchmark workload (we use random `user_data_*`).

### DIFF
--- a/src/lsm/groove.zig
+++ b/src/lsm/groove.zig
@@ -151,7 +151,7 @@ pub fn GrooveType(
     /// - ignored: [][]const u8:
     ///     An array of fields on the Object type that should not be given index trees
     ///
-    /// - derived: { .field = *const fn (*const Object) ?DerivedType }:
+    /// - derived: { .field = *const fn (*const Object) DerivedType }:
     ///     An anonymous struct which contain fields that don't exist on the Object
     ///     but can be derived from an Object instance using the field's corresponding function.
     comptime groove_options: anytype,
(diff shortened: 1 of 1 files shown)


src/lsm/groove.zig
```diff
@@ -926,9 +924,13 @@ pub fn GrooveType(
             inline for (std.meta.fields(IndexTrees)) |field| {
```
sentientwaffle, line 924: I was originally thinking that we now need to assert that `id != 0` at the LSM layer, but the id tree is handled separately from the secondary indexes, so nevermind!

sentientwaffle (approved):
Nice! Not sure if it is worth a TODO, but when we eventually support scanning by `user_id` we will need to remember to document the caveat that `user_id=0` objects are not scan-able.

jorangreef:
Nice! What is the performance difference?

batiati:
> Nice! What is the performance difference?

The benchmark had a modest ~5% of performance improvement because only `pending_transfer_id` and `timeout` are no longer inserted. The real improvement depends on each use case.

If we changed the workload to submit the `user_data_*` fields as ZERO, we would reach **>350K TPS**, which means **~25%** more (280K TPS) if compared with the previous code running the same workload (submitting `user_data` zeroed as well) and **~40%** more (250K TPS) if compared with the current workload submitting random numbers.

The difference between 25% and 40% also shows how fast sequential indexes are (e.g. prefix always the same ZERO + sequential timestamp) in comparison with no index at all 🤔.

jorangreef:
This is great @batiati.

PR #1522, opened 2024-02-06
VSR: Assert request header fields

sentientwaffle:
A request's index number and message checksum are both also included in the corresponding reply message.

Verify that they match!


src/vsr/message_header.zig
```diff
@@ -738,6 +738,7 @@ pub const Header = extern struct {
         replica: u8,
         reserved_frame: [16]u8 = [_]u8{0} ** 16,
 
+        /// The checksum of the corresponding Request.
         parent: u128,
```
matklad: Might call it request_checksum?

matklad (approved):

PR #1523, opened 2024-02-06
Client: add ULID helper functions

kprotty:
Taking inspiration from the java client, add little-endian ID generation helpers (based on [ULID](https://github.com/ulid/spec)) to the language clients.

This also re-enables the unit tests for our Go `pkg/types` module which was previously dead code.

### DIFF
--- a/src/clients/dotnet/TigerBeetle/UInt128Extensions.cs
+++ b/src/clients/dotnet/TigerBeetle/UInt128Extensions.cs
@@ -104,7 +104,8 @@ public static class UInt128Extensions
 }
 
 /// <summary>
-/// Safe generation of Universally Unique and Byte-Sortable Identifiers as UInt128s.
+/// Universally Unique and Binary-Sortable Identifiers as UInt128s based on
+/// <a href="https://github.com/ulid/spec">ULID</a>
 /// </summary>
 public static class ID
 {
@@ -113,8 +114,8 @@ public static class ID
 
     /// <summary>
     /// Generates a universally unique identifier as a UInt128.
-    /// Each call generates a monotonically increasing value to last under sequential consistency.
-    /// This function is thread-safe.
+    /// IDs are guaranteed to be monotonically increasing from the last.
+    /// This function is thread-safe and monotonicity is sequentially consistent.
     /// </summary>
     public static UInt128 Create()
     {
--- a/src/clients/dotnet/TigerBeetle/UInt128Extensions.cs
+++ b/src/clients/dotnet/TigerBeetle/UInt128Extensions.cs
@@ -109,7 +109,7 @@ public static class UInt128Extensions
 public static class ID
 {
     private static long idLastTimestamp = 0L;
-    private static readonly byte[] idLastRandom = new byte[12];
+    private static readonly byte[] idLastRandom = new byte[10];
 
     /// <summary>
     /// Generates a universally unique identifier as a UInt128.
@@ -118,8 +118,7 @@ public static class ID
     /// </summary>
     public static UInt128 Create()
     {
-        // Safe to cast from long to ulong as, in C#, this
-        var timestamp = DateTimeOffset.UtcNow.ToUnixTimeMilliseconds();
+        long timestamp = DateTimeOffset.UtcNow.ToUnixTimeMilliseconds();
         ulong randomLo;
         ushort randomHi;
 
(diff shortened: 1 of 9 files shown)


src/clients/java/src/main/java/com/tigerbeetle/UInt128.java
```diff
@@ -186,7 +186,7 @@ public static byte[] asBytes(final BigInteger value) {
      * <a href="https://github.com/ulid/spec">ULID</a> but is adjusted for u128-LE interpretation.
      *
      * @throws ArithmeticException if the random monotonic bits in the same millisecond overflows.
-     * @return an array of 16 bytes representing an unsigned 128-bit value in little endian.
+     * @return An array of 16 bytes representing an unsigned 128-bit value in little endian.
      */
     public static byte[] ID() {
```
batiati: This one slipped out from the previous review ... I think we should rename `ID()` to `createID()`, following the camelCase naming convention.

src/clients/dotnet/TigerBeetle/UInt128Extensions.cs
```diff
@@ -101,3 +102,65 @@ public static UInt128 ToUInt128(this BigInteger value)
...
+    private static readonly byte[] idLastRandom = new byte[12];
+
+    /// <summary>
+    /// Generates a universally unique identifier as a UInt128.
+    /// Each call generates a monotonically increasing value to last under sequential consistency.
+    /// This function is thread-safe.
+    /// </summary>
+    public static UInt128 Create()
+    {
+        // Safe to cast from long to ulong as, in C#, this
```
batiati: The comment is missing a part and the full stop.

src/clients/dotnet/TigerBeetle/UInt128Extensions.cs
```diff
@@ -101,3 +102,65 @@ public static UInt128 ToUInt128(this BigInteger value)
...
+{
+    private static long idLastTimestamp = 0L;
+    private static readonly byte[] idLastRandom = new byte[12];
+
+    /// <summary>
+    /// Generates a universally unique identifier as a UInt128.
+    /// Each call generates a monotonically increasing value to last under sequential consistency.
+    /// This function is thread-safe.
+    /// </summary>
+    public static UInt128 Create()
```
batiati, line 120: Let's call it `CreateID()` instead.

src/clients/dotnet/TigerBeetle/UInt128Extensions.cs
```diff
@@ -101,3 +102,65 @@ public static UInt128 ToUInt128(this BigInteger value)
...
+            randomLo = BitConverter.ToUInt64(lastRandom.Slice(0));
+            randomHi = BitConverter.ToUInt16(lastRandom.Slice(8));
+
+            // Increment the u80 stored in lastRandom using a u64 increment then u16 increment.
+            // Throws an exception if the entire u80 represented with both overflows.
+            // In C#, unsigned arithmetic wraps around on overflow by default so check for zero.
+            randomLo += 1;
+            if (randomLo == 0)
+            {
+                randomHi += 1;
```
batiati: Although the default behavior in C# is unchecked arithmetic, it can be changed by a compiler flag.
Let's make it explicit that we rely on overflow here.

```suggestion
                unchecked
                {
                    randomHi += 1;
                }
```

src/clients/go/pkg/types/main.go
```diff
@@ -73,12 +83,73 @@ func HexStringToUint128(value string) (Uint128, error) {
...
 	values := [2]uint64{value, 0}
 	return *(*Uint128)(unsafe.Pointer(&values[0]))
 }
+
+var idLastTimestamp int64
+var idLastRandom [10]byte
+var idMutex sync.Mutex
+
+// Generates a Universally Unique and Sortable Identifier encoded as a Uint128. 
+func ID() Uint128 {
```
batiati: Let's call it `CreateID()` and standardize the same docs comments across all languages.
It's nice to link the ULID spec as well.

src/clients/dotnet/TigerBeetle/UInt128Extensions.cs
```diff
@@ -101,3 +102,65 @@ public static UInt128 ToUInt128(this BigInteger value)
...
     }
 }
+
+/// <summary>
+/// Safe generation of Universally Unique and Byte-Sortable Identifiers as UInt128s.
+/// </summary>
+public static class ID
+{
+    private static long idLastTimestamp = 0L;
+    private static readonly byte[] idLastRandom = new byte[12];
```
batiati: Should it be 10 bytes?
```suggestion
    private static readonly byte[] idLastRandom = new byte[10];
```

src/clients/dotnet/TigerBeetle/UInt128Extensions.cs
```diff
@@ -101,3 +102,65 @@ public static UInt128 ToUInt128(this BigInteger value)
...
+            // Increment the u80 stored in lastRandom using a u64 increment then u16 increment.
+            // Throws an exception if the entire u80 represented with both overflows.
+            // In C#, unsigned arithmetic wraps around on overflow by default so check for zero.
+            randomLo += 1;
+            if (randomLo == 0)
+            {
+                randomHi += 1;
+                if (randomHi == 0)
+                {
+                    throw new OverflowException("random bits overflow on monotonic increment");
```
batiati:
```suggestion
                    throw new OverflowException("Random bits overflow on monotonic increment");
```

src/clients/dotnet/TigerBeetle/UInt128Extensions.cs
kprotty: TIL about the compiler flag. Just to _check_, `unchecked` around both increments right? (as they rely on overflow)

batiati (dismissed):
Thanks King!

batiati (approved):

PR #1525, opened 2024-02-08
Update slack invite in readme

brson:
The invite in the readme is expired. This one is a permanent invite URL, the same as on the website.


batiati (approved):
🙌

Issue #1115, opened 2023-08-11
Fast historical balance queries


bergkvist:
It seems like tigerbeetle is optimised for inserts rather than analytical queries.

I believe for smaller scale hobby projects or startups, the 1M TPS is not going to matter much, and the latency of analytical queries would be much more important than insert throughput. At a very small scale, running a separate analytical database might also not make sense from a cost and complexity perspective.

In particular - allowing a user to view historical account balances at different resolutions is something I imagine would be a very typical use case for investment banks, funds etc. (so that it can be plotted on a web frontend)

jorangreef:
Thanks @bergkvist — agreed on all these points, and for the same reasons.

We're already working on the LSM Scans infrastructure to support fast historical balance queries. Watch this space!

Let me know also if you'd like to bounce ideas on a call on this and other queries. Feel free to DM me in our community Slack.

matklad:
Closed by https://github.com/tigerbeetle/tigerbeetle/pull/1526 I would thing

PR #1526, opened 2024-02-08
Account historical balances

batiati:
#### Introduces the new operation `get_account_history` for querying the historical balances of a given account.

- This is an opt-in feature. Only accounts with the flag `history` set will retain the historical balance.

- Common filter type `AccountFilter` used for both `get_account_transfers` and `get_account_history`; Both operations share the same logic and return exactly the same timestamps when called with the same filter, which allows the results to be joined by the application layer.

- All clients and integration tests updated.

- Docs updated.

Related to #1115


docs/design/client-requests.md
```diff
@@ -22,12 +22,14 @@ The maximum number of events per batch depends on the maximum message size
...
+| Operation               | Batch Size |
+| ----------------------- | ---------: |
+| `lookup_accounts`       | 8190       |
+| `lookup_transfers`      | 8190       |
+| `create_accounts`       | 8190       |
+| `create_transfers`      | 8190       |
+| `get_account_transfers` | 8190       |
+| `get_account_history`   | 8190       |
 
 Presently the client application is responsible for batching events, but only as a stopgap
```
batiati, line 34: Something to change in another PR, we already have client batching.

src/lsm/scan_lookup.zig
```diff
@@ -11,12 +11,14 @@ const GridType = @import("../vsr/grid.zig").GridType;
 const ScanType = @import("scan_builder.zig").ScanType;
 
 /// Implements the lookup logic for loading objects from scans.
-pub fn ScanLookupType(comptime Groove: type, comptime Storage: type) type {
+pub fn ScanLookupType(
+    comptime Groove: type,
+    comptime Scan: type,
+    comptime Storage: type,
+) type {
```
batiati, line 18: We allow the `ScanLookup` to use indexes from one Groove, but perform the actual lookup in another one, since their `timestamp`s match.

src/tigerbeetle/cli.zig
```diff
@@ -54,11 +54,12 @@ const CliArgs = union(enum) {
 
     benchmark: struct {
         account_count: usize = 10_000,
+        account_history: bool = false,
         transfer_count: usize = 10_000_000,
         query_count: usize = 100,
         transfer_count_per_second: usize = 1_000_000,
         print_batch_timings: bool = false,
-        id_order: Command.Benchmark.IdOrder = .reversed,
+        id_order: Command.Benchmark.IdOrder = .sequential,
```
batiati, line 64: Unrelated, but it just feels like a better default.

docs/reference/account_balances.md
```diff
@@ -0,0 +1,65 @@
...
+sidebar_position: 1
+---
+
+# Account Balances
+
+An `AccountBalance` is a record storing the [`Account`](./accounts.md)'s balance
+at a given point in time.
+
+Only Accounts with the flag [`history`](./accounts.md#flagshistory) set retain the history of balances.
+
```
sentientwaffle, line 11: Be explicit here whether the balance fields refer to the balance before or after the transfer at that timestamp executes. (Maybe mention it in the "Account Filter"'s docs.)

docs/reference/account_filter.md
```diff
@@ -0,0 +1,82 @@
...
+
+An `AccountFilter` is a record containing the filter parameters for querying
+the [account transfers](operations/get_account_transfers.md)
+and the [account historical balances](operations/get_account_history.md).
+
+## Fields
+
+### `account_id`
+
+The unique identifier of the account for which the results will be retrieved.
```
sentientwaffle: Link "identifier" to `id` in the `references/accounts.md`.

src/scripts/client_readmes.zig
```diff
@@ -373,6 +373,21 @@ fn readme_root(ctx: *Context) !void {
...
+        ctx.header(2, "Get Account History");
+        ctx.paragraph(
+            \\NOTE: This is a preview API that is subject to breaking changes once we have
+            \\a stable querying API.
+            \\
+            \\Fetches the point-in-time balances of a given account, allowing basic filter and pagination
+            \\capabilities.
+            \\
+            \\The balances in the response are sorted by `timestamp` in chronological or
+            \\reverse-chronological order.
```
sentientwaffle: Mention that this can only be used for accounts that were created with the `history` flag set.

src/tigerbeetle/benchmark_driver.zig
```diff
@@ -116,7 +116,7 @@ fn start(allocator: std.mem.Allocator, options: struct {
     child.request_resource_usage_statistics = true;
     child.stdin_behavior = .Pipe;
     child.stdout_behavior = .Pipe;
-    child.stderr_behavior = .Ignore;
+    child.stderr_behavior = .Inherit;
```
sentientwaffle, line 119: :heart:

src/state_machine.zig
```diff
@@ -1353,6 +1616,14 @@ pub fn StateMachineType(
                     .tree_options_id = {},
                     .tree_options_index = .{},
                 },
+                .account_history = .{
+                    .prefetch_entries_for_read_max = 0,
+                    .prefetch_entries_for_update_max = batch_transfers_max,
+                    .cache_entries_max = options.cache_entries_posted,
```
sentientwaffle: Split this out into a separate option (e.g. `cache_entries_account_history`).

src/state_machine.zig
```diff
@@ -296,6 +363,36 @@ pub fn StateMachineType(
...
+                completion: *FieldType(field),
+            ) *StateMachine {
+                comptime assert(field != .null);
+
+                const context = @fieldParentPtr(ScanLookup, @tagName(field), completion);
+                return @fieldParentPtr(StateMachine, "scan_lookup", context);
+            }
+
+            pub fn get(self: *ScanLookup, comptime field: Field) *FieldType(field) {
+                assert(self.* == .null);
```
sentientwaffle, line 392: `assert(field != .null);`

src/state_machine.zig
```diff
@@ -592,9 +687,131 @@ pub fn StateMachineType(
...
+        fn prefetch_get_account_history_scan(self: *StateMachine, filter: AccountFilter) void {
             assert(self.scan_result_count == 0);
 
+            if (self.get_scan_from_filter(filter)) |scan| {
+                if (self.forest.grooves.accounts.get(filter.account_id)) |account| {
+                    if (account.flags.history) {
+                        var scan_buffer = std.mem.bytesAsSlice(
+                            AccountHistoryGrooveValue,
+                            self.scan_buffer,
+                        );
```
sentientwaffle: :thinking: depending on the multiple of `get_account_transfers`, is it possible that `scan_buffer`'s length is not an even multiple of `@sizeOf(AccountHistoryGrooveValue)`?

src/state_machine.zig
```diff
@@ -932,17 +1126,62 @@ pub fn StateMachineType(
...
             );
 
             return result_size;
         }
 
+        fn execute_get_account_history(
+            self: *StateMachine,
+            input: []const u8,
+            output: *align(16) [constants.message_body_size_max]u8,
+        ) usize {
```
sentientwaffle, line 1153: `assert(self.scan_result_count <= constants.batch_max.get_account_history)`

src/state_machine.zig
```diff
@@ -932,17 +1126,62 @@ pub fn StateMachineType(
             defer self.scan_result_count = 0;
```
sentientwaffle, line 1136: `assert(self.scan_result_count <= constants.batch_max.get_account_transfers`

src/state_machine.zig
```diff
@@ -716,6 +908,7 @@ pub fn StateMachineType(
                 .lookup_accounts => self.execute_lookup_accounts(input, output),
                 .lookup_transfers => self.execute_lookup_transfers(input, output),
                 .get_account_transfers => self.execute_get_account_transfers(input, output),
+                .get_account_history => self.execute_get_account_history(input, output),
             };
```
sentientwaffle, line 920: `assert(self.scan_result_count == 0);`

src/state_machine.zig
```diff
@@ -932,17 +1126,62 @@ pub fn StateMachineType(
...
+
+            const scan_results: []const AccountHistoryGrooveValue = mem.bytesAsSlice(
+                AccountHistoryGrooveValue,
+                self.scan_buffer[0 .. self.scan_result_count * @sizeOf(AccountHistoryGrooveValue)],
+            );
+
+            const output_slice: []AccountBalance = mem.bytesAsSlice(AccountBalance, output);
+            var output_count: usize = 0;
+
+            for (scan_results) |*result| {
```
sentientwaffle, line 1171: We could `assert((filter.account_id == result.dr_account_id) != (filter.account_id == result.cr_account_id))`.. to verify that _exactly_ one of the account ids matches. (Right now the unreachable checks that _at least_ one of the ids matches.)

Then again we have a pretty strong guarantee about this on the "create accounts" side so maybe its not helpful.

src/config.zig
```diff
@@ -231,6 +233,7 @@ pub const configs = struct {
             .cache_accounts_size_default = @sizeOf(vsr.tigerbeetle.Account) * 1024 * 1024,
             .cache_transfers_size_default = @sizeOf(vsr.tigerbeetle.Transfer) * 2048,
             .cache_transfers_posted_size_default = @sizeOf(u256) * 256 * 1024,
+            .cache_account_history_size_default = 2 * 256 * 1024,
```
batiati, line 236: Is there a better way to do that?
The ideal would be to define ZERO for the minimal possible cache 🤔

sentientwaffle: Ah right, the `SetAssociativeCache` doesn't allow for an empty capacity?
Maybe just add a TODO here that it can be zero in the future.

src/scripts/client_readmes.zig
```diff
@@ -379,8 +379,12 @@ fn readme_root(ctx: *Context) !void {
             \\NOTE: This is a preview API that is subject to breaking changes once we have
             \\a stable querying API.
             \\
-            \\Fetches the point-in-time balances of a given account, allowing basic filter and pagination
-            \\capabilities.
+            \\Fetches the point-in-time balances of a given account, allowing basic filter and
+            \\pagination capabilities.
+            \\
+            \\Only accounts created with the flag [history]
+            \\(https://docs.tigerbeetle.com/reference/accounts#flagshistory) set retain the
```
sentientwaffle: At least in github's markdown preview, this doesn't get parsed as a link (https://github.com/tigerbeetle/tigerbeetle/blob/2e4fa951427ba94697db99a083d1aee13fdf55ee/src/clients/go/README.md#get-account-history).

sentientwaffle (commented):
Nit and then LGTM

sentientwaffle (approved):

batiati:
Thanks DJ!

PR #1527, opened 2024-02-08
Benchmark: Forward replica process stderr

sentientwaffle:
Forward the replica child process' stderr to the parent process.

(Sometimes I like to add timers to specific parts of the code before running the full benchmark – this way those results are easily available.)

Also if there were any non-fatal issues with the replica, this would help to surface them.

### DIFF
--- a/src/tigerbeetle/benchmark_driver.zig
+++ b/src/tigerbeetle/benchmark_driver.zig
@@ -116,7 +116,7 @@ fn start(allocator: std.mem.Allocator, options: struct {
     child.request_resource_usage_statistics = true;
     child.stdin_behavior = .Pipe;
     child.stdout_behavior = .Pipe;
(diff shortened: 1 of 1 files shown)


sentientwaffle:
Nevermind, @batiati made this exact same change as part of https://github.com/tigerbeetle/tigerbeetle/pull/1526! :joy:

batiati:
> Nevermind, @batiati made this exact same change as part of #1526! 😂

Looks like we had the same problem with the replica crashing during the benchmark !!!
You should keep it as a separate PR, it's well documented.

sentientwaffle:
@batiati Thanks! In my case it wasn't actually crashing, I just wanted to get some debug logging from the replica during the benchmark :grin:

batiati (approved):

PR #1537, opened 2024-02-11
vsr: add failing test for root op corruption.

matklad:
Because op_checkpoint() == 0, backups don't try to repair op=0. But, without op=0, neither backup can become a primary.

Note: while we corrupt 2 copies of the prepare here, the corruptions are not simultaneous, and should be repaired.


matklad:
At this point, I don't understand why this is any different from the existing one for root op corruption, and it passes anyway.

PR #1548, opened 2024-02-12
clients/go: use big.Int in its pointer form

lucasew:
Using ToBigInt behaves weird in it's current form, that's why most of `big.Int` expects a pointer.

This PR solves this problem. I edited the golang cache files manually to test it lol.

### DIFF
--- a/src/clients/go/pkg/types/main.go
+++ b/src/clients/go/pkg/types/main.go
@@ -42,12 +42,12 @@ func (value Uint128) String() string {
 	return s[lastNonZero:]
 }
 
-func (value Uint128) BigInt() big.Int {
+func (value Uint128) BigInt() *big.Int {
 	// big.Int uses bytes in big-endian but Uint128 stores bytes in little endian, so reverse it.
 	bytes := value.Bytes()
 	swapEndian(bytes[:])
 
-	ret := big.Int{}
+	ret := &big.Int{}
 	ret.SetBytes(bytes[:])
 	return ret
 }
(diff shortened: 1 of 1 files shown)


src/clients/go/pkg/types/main.go
```diff
@@ -42,12 +42,12 @@ func (value Uint128) String() string {
...
-func (value Uint128) BigInt() big.Int {
+func (value Uint128) BigInt() *big.Int {
 	// big.Int uses bytes in big-endian but Uint128 stores bytes in little endian, so reverse it.
 	bytes := value.Bytes()
 	swapEndian(bytes[:])
 
-	ret := big.Int{}
+	ret := &big.Int{}
 	ret.SetBytes(bytes[:])
 	return ret
```
kprotty, line 52: The code here is being GC-conscious by avoiding the `big.Int` itself being heap allocated. In this case I think users could do `&u128.BigInt()` instead.

src/clients/go/pkg/types/main.go
```diff
@@ -82,7 +82,7 @@ func HexStringToUint128(value string) (Uint128, error) {
 }
 
 // BigIntToUint128 converts a [math/big.Int] to a Uint128.
-func BigIntToUint128(value big.Int) Uint128 {
+func BigIntToUint128(value *big.Int) Uint128 {
```
kprotty, line 85: Interacting with `*big.Int` pointers is indeed more convenient, so this change is nice to keep.

src/clients/go/pkg/types/main.go
lucasew: #TIL thank you

src/clients/go/pkg/types/main.go
lucasew: If bytes works the same I think in this case would be basically the same usability wise.

PR #1551, opened 2024-02-13
devhub: switch to a line graph and other misc enhancements

cb22:
I'm not sure if there are some ardent bar graph supporters here, but I think it's easier to see the change over time on a line graph... The y-axis range becomes relative now too.

Additionally, open tooltips to the commit rather than the range and show the date on the x-axis as it has more meaning than the commit sha (still as a category, though, to keep the points equally spaced).

Switch the date formatting to use .toLocaleString() for a nicer formatted date (/ time).

Before:
image

After:
image

### DIFF
--- a/src/devhub/devhub.js
+++ b/src/devhub/devhub.js
@@ -102,31 +102,43 @@ function plotSeries(seriesList, rootNode) {
         text: series.label,
       },
       chart: {
-        type: "bar",
+        type: "line",
         height: "400px",
         events: {
           dataPointSelection: (event, chartContext, { dataPointIndex }) => {
             window.open(
-              "https://github.com/tigerbeetle/tigerbeetle/commits/" +
+              "https://github.com/tigerbeetle/tigerbeetle/commit/" +
                 series.revision[dataPointIndex],
             );
           },
         },
+      },
+      markers: {
+        size: 4,
       },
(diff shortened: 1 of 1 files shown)


matklad:
Couple of thoughts:

* Can we use UTC for displaying time, to increase determinism
* I think I like bar charts and absolute scale a bit more:
    * line chart is a lie --- we have a discreet set of events, it doesn't make sense to interpolate between the points. It is a pretty-looking lie though!
    * relative numbers feel confusing in general, and, for high variance measurements, they don't allow to gauge variance at a glance

Don't feel strongly about either of those points though!

cb22:
> Can we use UTC for displaying time, to increase determinism

We can - but I went with local time because that's what `git log` / GitHub will display by default too - so it makes them line up exactly.

> line chart is a lie --- we have a discreet set of events, it doesn't make sense to interpolate between the points. It is a pretty-looking lie though!

Yes indeed :smile: however, I think the linear interpolation gives a very nice indication of the rate of change - the line could be made smaller to de-emphasize it further.

I think more subjectively, I just feel like I'm getting a clearer picture of the trend when looking at the lines, whereas the bars throw me off with all the busyness.

matklad (approved):

Issue #1510, opened 2024-02-02
Liveness: 15677965580699030051


tigerbeetle-vopr:
<strong>Commit:</strong> ```7f92126c74c715c295503e7b41f56cef552a550c``` <br><br><strong>Branches:</strong> main<br><br><strong>Duration to run seed in ReleaseSafe mode:</strong> 2m22s<br><br><strong>Debug Logs:</strong><br><details><summary>Tail</summary><br>
```debug: Replaying seed 15677965580699030051 in Debug mode with full debug logging enabled...


SEED=15677965580699030051

replicas=3
standbys=1
clients=7
request_probability=5%
idle_on_probability=14%
idle_off_probability=19%
one_way_delay_mean=10 ticks
one_way_delay_min=0 ticks
packet_loss_probability=23%
path_maximum_capacity=10 messages
path_clog_duration_mean=464 ticks
path_clog_probability=1%
packet_replay_probability=47%
partition_mode=testing.packet_simulator.PartitionMode.isolate_single
partition_symmetry=testing.packet_simulator.PartitionSymmetry.symmetric
partition_probability=2%
unpartition_probability=3%
partition_stability=100 ticks
unpartition_stability=8 ticks
read_latency_min=1
read_latency_mean=8
write_latency_min=0
write_latency_mean=81
read_fault_probability=4%
write_fault_probability=4%
crash_probability=0.00002%
crash_stability=873 ticks
restart_probability=0.0002%
restart_stability=946 ticks
 1/.1V0/__1/__1C0:__4Jo0/_0J!0:_31Wo <__0:__0>0Ga0G!0G?4/4Pp2/3Rq
 2\.1V0/__1/__1C0:__5Jo0/_3J!0:_31Wo <__0:__0>0Ga0G!0G?
 3|.1V0/__1/__1C0:__4Jo0/_0J!0:_31Wo <__0:__0>0Ga0G!0G?
 1/.1V0/__2/__2C0:__5Jo0/_1J!0:_31Wo <__0:__0>0Ga0G!0G?4/4Pp1/3Rq
 2\.1V0/__2/__2C0:__6Jo0/_2J!0:_31Wo <__0:__0>0Ga0G!0G?
 3|.1V0/__2/__2C0:__4Jo0/_0J!0:_31Wo <__0:__0>0Ga0G!0G?
 0\ .1V0/__1/__2C0:__4Jo0/_0J!0:_31Wo <__0:__0>0Ga0G!0G?
[...]

PR #1552, opened 2024-02-13
vsr: don't lose SV headres during view change

matklad:
This commit is a stepping stone to fixing #1510.

It deals with the following effect: a replica receiving an SV might not install all headers from the SV. Specifically, headers from the next log wrap are not added to the journal. As a result, such headers don't necessary survive into a DVC for a subsequent view change.

We special case (via `start_view_into_do_view_change`) scenario where the head op gets lost this way. However, ops other than the head might get lost as well:

- prepare_max=31
- replica receives SV with ops 31..32
- replica does not install op=32
- replica receives a prepare with op=33
- during view change, head op=33 is larger that view headers op=32, so SV_into_DVC logic doesn't get triggered.
- as a result, op=32 is absent from the DVC

This commit makes SV_into_DVC logic more precise: rather than constructing DVC from SV or from journal, the DVC is constructed from the union of both.

Right now, the only effect of this change should be that more headers are present in DVC, and, as a result, a would-be primary should require less repair.

However, this change also establishes a new, powerful invariant:

A DVC with log_view=v is now guaranteed to include all headers from previous views, that survived into view=v. That is, if a primary decides that a particular header should not be truncated, that fact is guaranteed to be present on _any_ replica which starts view=v. This in turn will allow us to strengthen the nack protocol!

Consider the following DVC quorum:

dvc replica=2 log_view=2
    op=12 view=2
    -
    -

dvc replica=1 log_view=1
    op=12 view=1
    op=11 view=1
    op=10 view=1

Here, `op=10` can be truncated --- although we don't the canonical header for this op, we know that it can't be op=10 view=1. Because view=1 < log_view=2, if the header is canonical, it must be a part of replica=2's dvc.

### DIFF
--- a/src/vsr/replica.zig
+++ b/src/vsr/replica.zig
@@ -7549,62 +7549,11 @@ pub fn ReplicaType(
                 Status.view_change,
             });
 
-            const status_before = self.status;
-            self.status = .view_change;
-
-            if (status_before == .normal and self.view_headers.array.get(0).op > self.op) {
-                // Transition from normal status, but the SV headers were part of the next wrap,
-                // so we didn't install them to our journal, and we didn't catch up.
-                // We will reuse the SV headers as our DVC headers to ensure that participating in
-                // another view-change won't allow the op to backtrack.
-                assert(self.log_view == self.view);
-                assert(self.view_headers.command == .start_view);
-
-                log.debug("{}: transition_to_view_change_status: " ++
-                    "(op_head={} view_headers_head={})", .{
-                    self.replica,
-                    self.op,
-                    self.view_headers.array.get(0).op,
-                });
-
-                self.view_headers.start_view_into_do_view_change();
-            } else if (status_before == .normal or
-                (status_before == .recovering and self.log_view == self.view) or
-                (status_before == .view_change and self.log_view == self.view))
+            if (self.status == .normal or
+                (self.status == .recovering and self.log_view == self.view) or
+                (self.status == .view_change and self.log_view == self.view))
             {
-                // Either:
-                // - Transition from normal status.
-                // - Recovering from normal status.
-                // - Retired primary that didn't finish repair.
-                assert(self.view == self.log_view);
-                assert(self.view < view_new);
-
-                self.view_headers.command = .do_view_change;
-                self.view_headers.array.clear();
-                self.view_headers.append(self.journal.header_with_op(self.op).?);
-
-                // The DVC headers include:
-                // - all available cluster-uncommitted ops, and
-                // - the highest cluster-committed op (if available).
-                // We cannot safely go beyond that in all cases:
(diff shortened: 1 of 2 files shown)


src/vsr/replica.zig
```diff
@@ -7659,6 +7610,92 @@ pub fn ReplicaType(
...
+            // Either:
+            // - Transition from normal status.
+            // - Recovering from normal status.
+            // - Retired primary that didn't finish repair.
+            assert(self.status == .normal or
+                (self.status == .recovering and self.log_view == self.view) or
+                (self.status == .view_change and self.log_view == self.view));
+
+            const primary_repairing =
+                self.status == .view_change and self.log_view == self.view;
```
matklad, line 7623: Don't really love this special casing. I tried to make the primary to update its view headers right after it receives dvc quorum, but that doesn't seem to work --- the repairing primary might still have gaps in headers, which prevents it from generating a valid SV.

matklad:
This is sort-of ready for review, but I wouldn't be surprise if I wake up to a vopr failure here tomorrow!

src/vsr/replica.zig
```diff
@@ -7659,6 +7610,92 @@ pub fn ReplicaType(
...
+                    };
+                };
+
+                if (header_journal != null and header_view != null) {
+                    assert(header_journal.?.op == header_view.?.op);
+                    assert(header_journal.?.view == header_view.?.view);
+                    assert(header_journal.?.checksum == header_view.?.checksum);
+                }
+
+                if (header_journal == null and header_view == null) {
```
sentientwaffle, line 7673: - `assert(view_headers_updated.array.len > 0);`: The new "op head" is never a blank.
- `assert(op != self.op);`: `self.op`'s header always exists in journal.

src/vsr/replica.zig
```diff
@@ -7659,6 +7610,92 @@ pub fn ReplicaType(
...
+
+                if (op <= self.commit_max) break;
+                op -= 1;
+            } else unreachable;
+
+            assert(op <= self.commit_max);
+            assert(op == self.commit_max or self.commit_max > self.op);
+
+            self.view_headers = view_headers_updated;
+            self.view_headers.verify();
```
sentientwaffle, line 7698: Alternatively, this function could just return `view_headers_updated` and let `transition_to_view_change_status` perform the actual update, so that it is more clear at the call site that that is what is happening.

matklad: I'd rather keep the signature consistent with `primary_update_view_headers`.

sentientwaffle (approved):

matklad:
CI failuer unrelated, though quiet mysterious! Filed https://github.com/tigerbeetle/tigerbeetle/pull/1553 to get a better error message next time

PR #1553, opened 2024-02-14
node: more helpful errors on assertion failure

matklad:
Instead of `0 != 1` we'll get the actual errors


src/clients/node/src/test.ts
```diff
@@ -602,7 +602,7 @@ test('can get account transfers', async (): Promise<void> => {
     timestamp = transfer.timestamp;
 
     assert.ok(history[i].timestamp == transfer.timestamp);
-    i++;    
+    i++;
```
matklad, line 605: @batiati if you use vscode, you might want to enable

```
"files.trimTrailingWhitespace": true,
```

setting ;-)

batiati (approved):

PR #1555, opened 2024-02-14
[Go] Add client batching integration tests

batiati:


src/clients/go/tb_client_test.go
```diff
@@ -14,10 +15,11 @@ import (
...
 const (
-	TIGERBEETLE_PORT                 = "3000"
-	TIGERBEETLE_CLUSTER_ID    uint64 = 0
-	TIGERBEETLE_REPLICA_ID    uint32 = 0
-	TIGERBEETLE_REPLICA_COUNT uint32 = 1
+	TIGERBEETLE_PORT                   = "3000"
+	TIGERBEETLE_CLUSTER_ID      uint64 = 0
+	TIGERBEETLE_REPLICA_ID      uint32 = 0
+	TIGERBEETLE_REPLICA_COUNT   uint32 = 1
+	TIGERBEETLE_CONCURRENCY_MAX uint   = 4096
```
kprotty, line 23: Should we bump this to 8192 to reflect the value in `tb_client/context.zig`?

src/clients/go/tb_client_test.go
```diff
@@ -230,6 +231,60 @@ func doTestClient(s *testing.T, client Client) {
 		assert.Equal(t, types.ToUint128(0), accountB.DebitsPending)
 	})
 
+	s.Run("can create concurrent transfers", func(t *testing.T) {
+		const TRANSFERS_MAX = 1_000_000
+		concurrencyMax := make(chan struct{}, TIGERBEETLE_CONCURRENCY_MAX)
+		finished := make(chan struct{}, TRANSFERS_MAX)
```
kprotty: Might make more sense to use a `sync.WaitGroup` here over a channel.

kprotty (commented):
Good to have this test in. Hopefully it shouldn't take too long to execute on CI.

src/clients/go/tb_client_test.go
batiati: Looks like the way we handle `TooManyOutstanding` isn't quite right. This test exposes a bug when running with `concurrency_max` higher than the size of the `DemuxPool` (which is 4096).

batiati:
> Good to have this test in. Hopefully it shouldn't take too long to execute on CI.

Yes, it took just a few seconds (< 10s) to run.

src/clients/go/tb_client_test.go
kprotty: Ah I see. tb_client wasn't adjusted properly after DemuxPool was introduced. It should be addressed by the intrusive packets PR at least.

kprotty (approved):

PR #1557, opened 2024-02-14
Calculate default values in CLI help message.

### LINKED ISSUE #1466: In CLI help message, use comptimePrint to print calculated default values
The CLI help message prints default values for some arguments. These are hardcoded strings but would ideally be the calculated values.

Example:

```
        \\  --cache-grid=<size><KB|MB|GB>
        \\        Set the grid cache size. The grid cache acts like a page cache for TigerBeetle,
        \\        and should be set as large as possible.
        \\        On a machine running only TigerBeetle, this is somewhere around
        \\        (Total RAM) - 3GB (TigerBeetle) - 1GB (System), eg 12GB for a 16GB machine.
        \\        Defaults to 1GB.
```

cc https://github.com/tigerbeetle/tigerbeetle/pull/1447#discussion_r1463602503

brson:
Closes #1466

### DIFF
--- a/src/tigerbeetle/cli.zig
+++ b/src/tigerbeetle/cli.zig
@@ -126,13 +126,13 @@ const CliArgs = union(enum) {
         \\        and should be set as large as possible.
         \\        On a machine running only TigerBeetle, this is somewhere around
         \\        (Total RAM) - 3GB (TigerBeetle) - 1GB (System), eg 12GB for a 16GB machine.
-        \\        Defaults to {[default_cache_grid]d}GB.
+        \\        Defaults to {[default_cache_grid_gb]d}GB.
         \\
         \\  --memory-lsm-manifest=<size><KB|MB|GB>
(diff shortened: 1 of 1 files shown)


src/tigerbeetle/cli.zig
```diff
@@ -158,6 +158,14 @@ const CliArgs = union(enum) {
     , .{
         .default_address = constants.address,
         .default_port = constants.port,
+        .default_cache_grid = @divExact(
+            constants.grid_cache_size_default,
+            1024 * 1024 * 1024,
+        ),
+        .default_memory_lsm_manifest = @divExact(
```
matklad: nit, but maybe use `default_cache_grid_gb` and `default_memory_lsm_manifest_mb` to connect the calculation and the actual help message?

brson: Done.

matklad (approved):

PR #1559, opened 2024-02-15
dotnet: simplify tests

matklad:
One issue I've fond when running tests on widows is that the data file isn't actually deleted, with an error message saying that another process still uses the data file.

Let's try:

- to always wait for the process
- and also don't swallow the error

### DIFF
--- a/src/clients/dotnet/TigerBeetle.Tests/IntegrationTests.cs
+++ b/src/clients/dotnet/TigerBeetle.Tests/IntegrationTests.cs
@@ -1595,7 +1595,7 @@ internal class TBServer : IDisposable
     private const string TB_SERVER = TB_PATH + "/" + TB_EXE;
 
     private readonly Process process;
-    private readonly String dataFile;
+    private readonly string dataFile;
 
     public string Address { get; }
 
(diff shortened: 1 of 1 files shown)


src/clients/dotnet/TigerBeetle.Tests/IntegrationTests.cs
```diff
@@ -1626,16 +1626,9 @@ public TBServer()
 
     public void Dispose()
     {
-        try
-        {
-            if (process != null && !process.HasExited)
```
matklad, line 1631: @batiati process can't be null here, right? If we are running Dispose, constructor has finished?

Or is there some special C# semantics when Dispose might get called if you only managed to run through half of the constructor's code?

batiati: Nice catch! Since `process` is readonly, it cannot be null after the constructor
https://github.com/tigerbeetle/tigerbeetle/blob/2cdabe44b44a54d279e2866da6812e1201506eb5/src/clients/dotnet/TigerBeetle.Tests/IntegrationTests.cs#L1624

And objects that are not fully constructed don't run the finalizer/destructor/dispose.

I suppose the original intent was to prevent double `Dispose(),` but it's completely out of scope to cover this edge case in a test runner!

batiati (commented):
Just a nit, change the type from `String` to `string` here:

https://github.com/tigerbeetle/tigerbeetle/blob/2cdabe44b44a54d279e2866da6812e1201506eb5/src/clients/dotnet/TigerBeetle.Tests/IntegrationTests.cs#L1625-L1626

batiati (approved):

PR #1561, opened 2024-02-16
tmp_tigerbeetle: better error reporting on process crash

matklad:
Ideally, we'd also fetch stderr, but that's tricky --- stderr is potentially infinite, so we need to allocate to rule-out child process deadlocking...

### DIFF
--- a/src/testing/tmp_tigerbeetle.zig
+++ b/src/testing/tmp_tigerbeetle.zig
@@ -80,10 +80,20 @@ pub fn init(
     }
 
     const port = port: {
-        errdefer log.err("failed to read port number from tigerbeetle process", .{});
+        var exit_status: ?std.ChildProcess.Term = null;
+        errdefer log.err(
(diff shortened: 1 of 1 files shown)


batiati (approved):
> Ideally, we'd also fetch stderr, but that's tricky

Doesn't [setting stderr as inherited](https://github.com/tigerbeetle/tigerbeetle/pull/1527) solve it?

matklad:
I think it’s important that there’s no output, unless there is failure. In my experience, too much unrelated debug output makes debugging harder, as it makes workflow of adding ad hoc printlns less useful.

So, I’d say ability to manually flip this to inherited and recompile is perhaps better than just doing that by default.

one extra considiration is that inheriting stderr makes the particular windows bug I am chasing at the moment go away :-)

matklad:
The irony! Looks like ignoring stderr is **exactly** the thing that breaks on windows:

https://github.com/ziglang/zig/pull/15565

PR #1562, opened 2024-02-16
[Dotnet] Starts a single server instance for integration tests.

batiati:
Changes the integration tests logic to share the same server instance for running all tests instead of a fresh instance per test.

It also removes the `[DoNotParallelize]` attribute.

Before
```
Passed!  - Failed:     0, Passed:    72, Skipped:     0, Total:    72, Duration: 2 m 45 s
```

After
```
Passed!  - Failed:     0, Passed:    72, Skipped:     0, Total:    72, Duration: 1 m 28 s
```

### DIFF
--- a/src/clients/dotnet/TigerBeetle.Tests/IntegrationTests.cs
+++ b/src/clients/dotnet/TigerBeetle.Tests/IntegrationTests.cs
@@ -13,31 +13,47 @@ namespace TigerBeetle.Tests;
 [TestClass]
 public class IntegrationTests
 {
-    private static Client GetClient(string address, int concurrencyMax = 32) => new(0, new string[] { address }, concurrencyMax);
-
-    private static readonly Account[] accounts = new[]
+    private static Account[] GenerateAccounts() => new[]
     {
             new Account
             {
-                Id = 100,
+                Id = ID.Create(),
                 UserData128 = 1000,
                 UserData64 = 1001,
                 UserData32 = 1002,
-                Flags = AccountFlags.History,
+                Flags = AccountFlags.None,
                 Ledger = 1,
                 Code = 1,
             },
             new Account
             {
-                Id = 101,
+                Id = ID.Create(),
                 UserData128 = 1000,
                 UserData64 = 1001,
                 UserData32 = 1002,
                 Flags = AccountFlags.History,
                 Ledger = 1,
                 Code = 2,
-            }
-        };
+            },
+    };
+
+    // Created by the test initializer:
+    private static TBServer server = null!;
+    private static Client client = null!;
+
+    [ClassInitialize]
+    public static void Initialize(TestContext _)
+    {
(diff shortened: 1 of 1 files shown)


matklad:
🥳

src/clients/dotnet/TigerBeetle.Tests/IntegrationTests.cs
```diff
@@ -142,94 +163,98 @@ public void ConstructorAndFinalizer()
...
     }
 
     [TestMethod]
-    [DoNotParallelize]
     public void CreateTransfers()
     {
-        using var server = new TBServer();
-        using var client = GetClient(server.Address);
+        Assert.IsNotNull(server);
+        Assert.IsNotNull(client);
```
matklad: Tbh, I’d drop these extra asserts from tests: this is something forced upon us by the testing framework, but we are using an idiomatic testing framework, and so can accept slight degradation of safety (in tests !!), so as not to swim upstream against C# here.

src/clients/dotnet/TigerBeetle.Tests/IntegrationTests.cs
```diff
@@ -1540,38 +1563,38 @@ private void ConcurrentTasksDispose(bool isAsync)
...
         };
 
-        _ = client.CreateTransfers(new Transfer[] { transfer });
+        _ = client.CreateTransfers(new[] { transfer });
         Assert.Fail();
     }
 
 
-    private static void AssertAccounts(Account[] lookupAccounts)
+    private static void AssertAccounts(Account[] expected, Account[] actual)
```
matklad, line 1520: Can we use this thing here? 

https://stackoverflow.com/a/11055686

hopefully it’ll print readable diff on mismatxhn

src/clients/dotnet/TigerBeetle.Tests/IntegrationTests.cs
batiati: It's due to the compiler null checks. I'd rather use those `Assert.IsNotNull(server)` than the null-forgiving syntax `server!` everywhere.

src/clients/dotnet/TigerBeetle.Tests/IntegrationTests.cs
batiati: We can't compare all fields; for example, the balances can change between creation and lookup.

src/clients/dotnet/TigerBeetle.Tests/IntegrationTests.cs
matklad: Ahh, that makes sense. Seems like C# has an escape hatch here?

https://stackoverflow.com/a/73167373

Could we declare the relevant fields as non-null and init then with `null!`?

matklad (approved):

PR #1563, opened 2024-02-17
[Java] Starts a single server instance for integration tests. 

batiati:
Changes the integration tests logic to share the same server instance for running all tests instead of a fresh instance per test.


src/clients/dotnet/TigerBeetle.Tests/IntegrationTests.cs
```diff
@@ -1582,7 +1579,7 @@ public TBServer()
         dataFile = Path.GetRandomFileName();
 
         {
-            var format = new Process();
```
matklad, line 1585: :heart:

matklad (approved):
:tada:

PR #1565, opened 2024-02-19
Improvements to CLI errors

### LINKED ISSUE #1465: Error messages for `--limit-storage` etc should use appropriate byte units
The error messages for `--storage-size-limit` print information about the low/high/multiples allowed for the argument. These are printed in bytes (but not labeled as bytes). These arguments accept a variety of units, eg "1024", "1KB" ,"1MB", and it is usually best to provide the argument in such units, not bytes.

Example:

```
$ ./tigerbeetle start --addresses=3000 --limit-storage=0 0_0.tigerbeetle
error: --storage-size-limit: size 0 is below minimum: 1108344832
```

These messages should print their numbers in a more readable unit, probably MB. The `--cache-grid` argument already does so.

This may soon be true for `--memory-lsm-manifest` as well cc https://github.com/tigerbeetle/tigerbeetle/pull/1447

brson:
These changes are primarily concerned with displaying human-readable units for byte-oriented command line arguments.

Firstly, it prints constant limits in whatever unit is appropriate, GB, MB, KB.

Secondly, when it prints the user-supplied value, instead of printing the calculated bytes, it prints the string representation of the value as originally written on the command line.

Instead of

```
$ ./tigerbeetle start --addresses=1001 --limit-storage=1KB 0_0.tigerbeetle
error: --storage-size-limit: size 1024 is below minimum: 1108344832
```

we see

```
$ ./tigerbeetle start --addresses=1001 --limit-storage=1KB 0_0.tigerbeetle
error: --limit-storage: size 1KB is below minimum: 1057MB
```

It also fixes the name of `--limit-storage` in the error messages.

Closes https://github.com/tigerbeetle/tigerbeetle/issues/1465

### DIFF
--- a/src/flags.zig
+++ b/src/flags.zig
@@ -311,8 +311,12 @@ fn parse_value_size(flag: []const u8, value: []const u8) ByteSize {
                 "{s}: value exceeds 64-bit unsigned integer: '{s}'",
                 .{ flag, value },
             ),
-            error.InvalidCharacter => fatal(
-                "{s}: expected a size, but found '{s}' (invalid digit or suffix)",
+            error.MissingDigits => fatal(
+                "{s}: need digits in size '{s}'",
+                .{ flag, value },
+            ),
+            error.InvalidUnit => fatal(
+                "{s}: invalid unit in size '{s}', (needed KiB, MiB, GiB or TiB)",
                 .{ flag, value },
             ),
             error.BytesOverflow => fatal(
@@ -333,7 +337,8 @@ pub const ByteUnit = enum(u64) {
 
 const ByteSizeParseError = error{
     ParseOverflow,
-    InvalidCharacter,
+    MissingDigits,
+    InvalidUnit,
     BytesOverflow,
 };
 
@@ -342,6 +347,8 @@ pub const ByteSize = struct {
     unit: ByteUnit = .bytes,
 
     fn parse(value: []const u8) ByteSizeParseError!ByteSize {
+        assert(value.len != 0);
+
         const units = .{
             .{ &[_][]const u8{ "TiB", "tib" }, ByteUnit.tib },
             .{ &[_][]const u8{ "GiB", "gib" }, ByteUnit.gib },
@@ -349,38 +356,55 @@ pub const ByteSize = struct {
             .{ &[_][]const u8{ "KiB", "kib" }, ByteUnit.kib },
         };
 
-        const unit: struct {
+        const split: struct {
             value_input: []const u8,
-            unit: ByteUnit,
-        } = unit: inline for (units) |unit| {
-            const suffixes = unit[0];
-            const unit_kind = unit[1];
-            for (suffixes) |suffix_| {
-                if (std.mem.endsWith(u8, value, suffix_)) {
-                    break :unit .{
-                        .value_input = value[0 .. value.len - suffix_.len],
-                        .unit = unit_kind,
-                    };
-                }
+            unit_input: []const u8,
+        } = split: for (0..value.len) |i| {
+            if (!std.ascii.isDigit(value[i])) {
+                break :split .{
+                    .value_input = value[0..i],
+                    .unit_input = value[i..],
+                };
             }
-        } else break :unit .{ .value_input = value, .unit = ByteUnit.bytes };
+        } else {
+            break :split .{
+                .value_input = value,
+                .unit_input = "",
+            };
+        };
 
-        const amount = std.fmt.parseUnsigned(u64, unit.value_input, 10) catch |err| {
+        const amount = std.fmt.parseUnsigned(u64, split.value_input, 10) catch |err| {
             switch (err) {
                 error.Overflow => {
                     return ByteSizeParseError.ParseOverflow;
                 },
                 error.InvalidCharacter => {
-                    return ByteSizeParseError.InvalidCharacter;
+                    // The only case this can happen is for the empty string
+                    return ByteSizeParseError.MissingDigits;
                 },
             }
         };
 
-        _ = std.math.mul(u64, amount, @intFromEnum(unit.unit)) catch {
+        const unit = if (split.unit_input.len > 0)
+            unit: inline for (units) |unit_| {
+                const suffixes = unit_[0];
+                const unit_kind = unit_[1];
+                for (suffixes) |suffix_| {
+                    if (std.mem.endsWith(u8, split.unit_input, suffix_)) {
+                        break :unit unit_kind;
+                    }
+                }
+            } else {
+                return ByteSizeParseError.InvalidUnit;
+            }
+        else
+            ByteUnit.bytes;
+
+        _ = std.math.mul(u64, amount, @intFromEnum(unit)) catch {
             return ByteSizeParseError.BytesOverflow;
         };
 
-        return ByteSize{ .value = amount, .unit = unit.unit };
+        return ByteSize{ .value = amount, .unit = unit };
     }
 
     pub fn bytes(size: *const ByteSize) u64 {
@@ -1023,7 +1047,7 @@ test "flags" {
     try t.check(&.{ "values", "--size=3Mib" }, snap(@src(),
         \\status: 1
         \\stderr:
-        \\error: --size: expected a size, but found '3Mib' (invalid digit or suffix)
+        \\error: --size: invalid unit in size '3Mib', (needed KiB, MiB, GiB or TiB)
         \\
     ));
 
--- a/src/flags.zig
+++ b/src/flags.zig
@@ -1042,14 +1042,10 @@ test "flags" {
         \\
     ));
 
-    try t.check(&.{ "values", "--size=3Mib" }, snap(@src(),
-        \\stdout:
-        \\int: 0
-        \\size: 3145728
-        \\boolean: false
-        \\path: not-set
-        \\optional: null
-        \\choice: marlowe
+    try t.check(&.{ "values", "--size=3bogus" }, snap(@src(),
+        \\status: 1
+        \\stderr:
+        \\error: --size: invalid unit in size '3bogus', (needed KiB, MiB, GiB or TiB)
         \\
     ));
 
--- a/src/flags.zig
+++ b/src/flags.zig
@@ -349,13 +349,6 @@ pub const ByteSize = struct {
     fn parse(value: []const u8) ByteSizeParseError!ByteSize {
         assert(value.len != 0);
 
-        const units = .{
-            .{ "TiB", ByteUnit.tib },
-            .{ "GiB", ByteUnit.gib },
-            .{ "MiB", ByteUnit.mib },
-            .{ "KiB", ByteUnit.kib },
-        };
-
         const split: struct {
             value_input: []const u8,
             unit_input: []const u8,
@@ -386,11 +379,9 @@ pub const ByteSize = struct {
         };
 
         const unit = if (split.unit_input.len > 0)
-            unit: inline for (units) |unit_| {
-                const suffix_ = unit_[0];
-                const unit_kind = unit_[1];
-                if (std.ascii.eqlIgnoreCase(split.unit_input, suffix_)) {
-                    break :unit unit_kind;
+            unit: inline for (comptime std.enums.values(ByteUnit)) |tag| {
+                if (std.ascii.eqlIgnoreCase(split.unit_input, @tagName(tag))) {
+                    break :unit tag;
                 }
             } else {
                 return ByteSizeParseError.InvalidUnit;
--- a/src/flags.zig
+++ b/src/flags.zig
@@ -302,33 +302,12 @@ fn parse_value(comptime T: type, flag: []const u8, value: [:0]const u8) T {
     comptime unreachable;
 }
 
-pub const ByteSize = struct { bytes: u64 };
 fn parse_value_size(flag: []const u8, value: []const u8) ByteSize {
     assert((flag[0] == '-' and flag[1] == '-') or flag[0] == '<');
 
-    const units = .{
-        .{ &[_][]const u8{ "TiB", "tib" }, 1024 * 1024 * 1024 * 1024 },
-        .{ &[_][]const u8{ "GiB", "gib" }, 1024 * 1024 * 1024 },
-        .{ &[_][]const u8{ "MiB", "mib" }, 1024 * 1024 },
-        .{ &[_][]const u8{ "KiB", "kib" }, 1024 },
-    };
-
-    const unit: struct { suffix: []const u8, scale: u64 } = unit: inline for (units) |unit| {
-        const suffixes = unit[0];
-        const scale = unit[1];
-        for (suffixes) |suffix| {
-            if (std.mem.endsWith(u8, value, suffix)) {
-                break :unit .{ .suffix = suffix, .scale = scale };
-            }
-        }
-    } else break :unit .{ .suffix = "", .scale = 1 };
-
-    assert(std.mem.endsWith(u8, value, unit.suffix));
-    const value_numeric = value[0 .. value.len - unit.suffix.len];
-
-    const amount = std.fmt.parseUnsigned(u64, value_numeric, 10) catch |err| {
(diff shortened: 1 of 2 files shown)


src/flags.zig
```diff
@@ -302,7 +302,18 @@ fn parse_value(comptime T: type, flag: []const u8, value: [:0]const u8) T {
...
+    original: []const u8,
+
+    pub fn default(value: u64) ByteSize {
+        return .{
+            .bytes = value,
+            .original = "error", // this shouldn't ever be visible in the UI
+        };
+    }
+};
+
```
batiati: I think we can be more precise here, making this type responsible for parsing the user input, so we don't need to store the original string or handle special cases for constants.

For example:

```zig
pub const ByteUnit = enum(u32) {
    bytes = 1,
    kb = 1024,
    mb = .kb * 1024,
    gb = .mb * 1024,
    tb = .gb * 1024,
};

pub const ByteSize = struct {
    value: u64,
    unit: ByteUnit,

    pub fn parse(input: []const u8) ByteSize {
        // Move the "parse_value_size" logic here.
        return .{
            .value = parsed_value,
            .unit = parsed_unit,
        };
    }

    pub fn bytes(size: *const ByteSize) u64 {
        return size.value * @intFromEnum(size.unit);
    }

    pub fn suffix(size: *const ByteSize) []const u8 {
        return switch (size.unit) {
            .bytes => "",
            .kb => "KB",
            .mb => "MB",
            .gb => "GB",
            .tb => "TB",
        };
    }
};

// Printing an error message:
flags.fatal("--limit-storage: size {}{s} is below minimum: {}MB", .{
                    start.limit_storage.value,
                    start.limit_storage.suffix(),
                    @divExact(storage_size_limit_min, 1024 * 1024),
                });
```

batiati (commented):
For this PR I think we can focus on only improving the error message.
However, should we support decimal inputs? Like `1.5GB`?

src/flags.zig
brson: I have made this change and pushed it. Some of my changes may not be idiomatic so please give it another look.

e.g. I wasn't able to write a self-referential enum like the one you suggested (`ByteUnit`). I also used an underscore to disambiguate a name.

brson:
> For this PR I think we can focus on only improving the error message. However, should we support decimal inputs? Like `1.5GB`?

Seems like something people will expect to work eventually. Maybe it can wait until we receive requests for it.

src/flags.zig
```diff
@@ -296,101 +296,135 @@ fn parse_value(comptime T: type, flag: []const u8, value: [:0]const u8) T {
...
+                ),
+            }
+        };
 
-    assert(std.mem.endsWith(u8, value, unit.suffix));
-    const value_numeric = value[0 .. value.len - unit.suffix.len];
+        _ = std.math.mul(u64, amount, @intFromEnum(unit.unit)) catch fatal(
+            "{s}: size in bytes exceeds 64-bit unsigned integer: '{s}'",
+            .{ flag, value },
+        );
```
brson: Is there a prettier way to write this overflow check that discards the actual result?

src/flags.zig
```diff
@@ -296,90 +296,135 @@ fn parse_value(comptime T: type, flag: []const u8, value: [:0]const u8) T {
...
-            if (std.mem.endsWith(u8, value, suffix)) {
-                break :unit .{ .suffix = suffix, .scale = scale };
+    pub fn default(value: u64) ByteSize {
+        return .{
+            .value = value,
+            .unit = ByteUnit.bytes,
+        };
+    }
+
+    fn parse(flag: []const u8, value: []const u8) ByteSize {
```
batiati: Since `flag` is only used in the error message, wouldn't it be simpler to remove this parameter and return `!ByteSize` from `parse`?

I think it's ok if the call site handles the error by printing the message with the raw input string.

src/flags.zig
```diff
@@ -296,90 +296,135 @@ fn parse_value(comptime T: type, flag: []const u8, value: [:0]const u8) T {
...
-        const scale = unit[1];
-        for (suffixes) |suffix| {
-            if (std.mem.endsWith(u8, value, suffix)) {
-                break :unit .{ .suffix = suffix, .scale = scale };
+    pub fn default(value: u64) ByteSize {
+        return .{
+            .value = value,
+            .unit = ByteUnit.bytes,
+        };
+    }
```
batiati: Let's remove the `default` function, declaring the field `unit: ByteUnit = .bytes`.
It simplifies the initialization e.g. `size: ByteSize = .{ .value = 0 };`

src/flags.zig
batiati: The built-in `@mulWithOverflow` returns the result and the overflow bit. The function `std.math.mul()` is just a wrapper to make it idiomatic with errors;

I think we need to discard the result anyway!

src/flags.zig
```diff
@@ -296,90 +296,135 @@ fn parse_value(comptime T: type, flag: []const u8, value: [:0]const u8) T {
...
+                if (std.mem.endsWith(u8, value, suffix_)) {
+                    break :unit .{ .suffix = suffix_, .unit = unitkind };
+                }
             }
-        }
-    } else break :unit .{ .suffix = "", .scale = 1 };
+        } else break :unit .{ .suffix = "", .unit = ByteUnit.bytes };
+
+        assert(std.mem.endsWith(u8, value, unit.suffix));
+        const value_numeric = value[0 .. value.len - unit.suffix.len];
```
batiati: It can be simplified, returning the input value as string instead of the suffix.

```zig
        const unit: struct {
            value_input: []const u8,
            unit: ByteUnit,
        } = unit: inline for (units) |unit| {
            const suffixes = unit[0];
            const unit_kind = unit[1];
            for (suffixes) |suffix_| {
                if (std.mem.endsWith(u8, value, suffix_)) {
                    break :unit .{ .value_input = value[0 .. value.len - suffix_.len], .unit = unit_kind };
                }
            }
        } else break :unit .{ .value_input = value, .unit = ByteUnit.bytes };
```

src/flags.zig
batiati: My bad! TIL that [zig does not allow self-dependency on enums](https://github.com/ziglang/zig/issues/5850), which was surprising since it's not an autonumerated enum.

batiati (commented):
Awesome! Just a few comments and we are good to merge!

insalt-glitch:
I know that I have not worked on this code, by no means I intend to stall this, but one thing that I found when looking through the CLI-parsing code: The current parse for byte units does not distinguish between `GB` or `KB` and `GiB` or `KiB`. Usually, `GB=1000^3 bytes` and `GiB=1024^3 bytes` for example, i.e. the difference between metric and binary systems. Maybe this was intended? The current system definitely keeps things simple. However, I think a warning might be appropriate?

src/flags.zig
brson: It might be, but there's a preexisting error handling pattern in this file where every function calls `fatal` directly on errors instead of propagating them; and several parsing functions take this same `flag` parameter as part of that pattern.

If the error handling is not correct here I think it should be addressed for the entire file at once. Can I (should I) file a followup issue about it instead of changing it here?

src/flags.zig
brson: I have made this change.

brson:
> I know that I have not worked on this code, by no means I intend to stall this, but one thing that I found when looking through the CLI-parsing code: The current parse for byte units does not distinguish between `GB` or `KB` and `GiB` or `KiB`. Usually, `GB=1000^3 bytes` and `GiB=1024^3 bytes` for example, i.e. the difference between metric and binary systems. Maybe this was intended? The current system definitely keeps things simple. However, I think a warning might be appropriate?

I was also a little surprised - seeing the support for these units made me think the code would be very pedantic about units, but it is not. I think the current behavior is defensible in that arguably most software people are intending to use base-2 units - I at least am always intending base-2 units even if I write "MB"; but I don't have a strong opinion.

src/flags.zig
brson: I have made this change.

brson:
@batiati I have addressed the review items and this is ready for another look.

src/flags.zig
batiati: I don't think it is necessary. Just that it feels odd when we have encapsulated “parse” as a type function that should know nothing about the error handling. I’d suggest you to keep the error handling function symmetrical to the other ones, but calling the “try ByteSize.parse” that we decoupled. Or, if you find it easier/clearer, you can revert it to a single function outside BiteSize, it is your call. I think the most important is to be able to precisely store and retrieve units either from constants or from the user input.

batiati:
> I know that I have not worked on this code, by no means I intend to stall this, but one thing that I found when looking through the CLI-parsing code: The current parse for byte units does not distinguish between `GB` or `KB` and `GiB` or `KiB`. Usually, `GB=1000^3 bytes` and `GiB=1024^3 bytes` for example, i.e. the difference between metric and binary systems. Maybe this was intended? The current system definitely keeps things simple. 

TigerBeetle aligns everything as power-of-two multiples, so there isn't much practical use for decimal units (sure the user can input 1000000 bytes instead of 1MB, but it will be allocated as the closest power-of-two value anyway).
That said, we have the option of either accepting both MB and MiB as the same thing or being pedantic and not accepting decimal units at all, so for that reason, we choose the best experience over nomenclature.

As my personal opinion, `KB`, `MB`, `GB` and `TB` shouldn't exist as a unit for bytes, and should only be considered as a mistyped version of the correct ones `KiB`, `MiB` , `GiB` and `TiB` 😂

image

batiati (changes requested):
> @batiati I have addressed the review items and this is ready for another look.

Great! Please just settle [the `ByteSize.parse` issue](https://github.com/tigerbeetle/tigerbeetle/pull/1565#discussion_r1503257403) before we merge.

I think we can also rename all occurrences (at least in this file) of inappropriate byte units to the correct form `KiB`, `MiB`, `GiB`, `TiB` (in strings, enums and variable names). But it's your call if you want to do that now or in a follow-up PR.

brson:
@batiati I have made the error-handling changes, reintroducing the free function "parse_value_size" and doing the error reporting there, while "ByteSize.parse" returns an error set.

I've filed two followup issues around unit handling:

- https://github.com/tigerbeetle/tigerbeetle/issues/1672
- https://github.com/tigerbeetle/tigerbeetle/issues/1673

brson:
https://github.com/tigerbeetle/tigerbeetle/pull/1674 will likely conflict with this. I can wait until that lands before rebasing this one.

brson:
@batiati This is rebased over the conflicting PRs and I think all review requests are resolved.

src/flags.zig
```diff
@@ -302,84 +302,136 @@ fn parse_value(comptime T: type, flag: []const u8, value: [:0]const u8) T {
...
+                }
+            }
+        } else break :unit .{ .value_input = value, .unit = ByteUnit.bytes };
+
+        const amount = std.fmt.parseUnsigned(u64, unit.value_input, 10) catch |err| {
+            switch (err) {
+                error.Overflow => {
+                    return ByteSizeParseError.ParseOverflow;
+                },
+                error.InvalidCharacter => {
```
batiati, line 374: As a colateral effect of not accepting MB or GB anymore, this error feels terribly cryptic now (imagine asking for 1gb and seeing this 😬)

Could we include MB/GB/TB only for the purpose of returning an specific error telling the user to use MiB/GiB/TiB instead?

batiati: Or maybe a simpler solution: just replace the error message with something like "invalid unit, it should be MiB, GiB or TiB"

brson: I've addressed this now. I restructured this parser a bit to distinguish between incorrect units and incorrect digits.

src/flags.zig
```diff
@@ -333,7 +337,8 @@ pub const ByteUnit = enum(u64) {
 
 const ByteSizeParseError = error{
     ParseOverflow,
-    InvalidCharacter,
+    MissingDigits,
```
batiati: It's great to handle  `InvalidUnit` separately!

Just a nit: rename `MissingDigits` with something prefixed with `Invalid*` to align with `InvalidUnit` ... for example, `InvalidSize` or `InvalidAmount`.

Also, I think the old message was more informative for this case `"{s}: expected a size, but found '{s}'"`

src/flags.zig
```diff
@@ -302,84 +302,160 @@ fn parse_value(comptime T: type, flag: []const u8, value: [:0]const u8) T {
...
+
+    fn parse(value: []const u8) ByteSizeParseError!ByteSize {
+        assert(value.len != 0);
+
+        const units = .{
+            .{ &[_][]const u8{ "TiB", "tib" }, ByteUnit.tib },
+            .{ &[_][]const u8{ "GiB", "gib" }, ByteUnit.gib },
+            .{ &[_][]const u8{ "MiB", "mib" }, ByteUnit.mib },
+            .{ &[_][]const u8{ "KiB", "kib" }, ByteUnit.kib },
+        };
```
batiati: Now that we have split the unit into a dedicated `[]const u8` slice, what do you think about using `std.ascii.endsWithIgnoreCase` and eliminating the special cases (i.e. "TiB" and "tib" )? 

I think it's okay if someone inputs "1GIB" or "1Gib" instead of "1GiB."

batiati (commented):
Sorry for too much back-and-forth, but I'm dogfooding this change and noticing things that may improve my own experience 😉

brson:
> Sorry for too much back-and-forth, but I'm dogfooding this change and noticing things that may improve my own experience 😉

No worries. I'm happy to do as much rework as needed.

src/flags.zig
brson: Done.

src/flags.zig
brson: Done.

brson:
Ah, don't approve this yet. My last patch introduced a new bug.

brson:
Ok, now it's ready for another review - with the split digits/units the unit comparison needs to be for equality, nod just "ends with".

src/flags.zig
```diff
@@ -969,9 +1043,13 @@ test "flags" {
...
-        \\status: 1
-        \\stderr:
-        \\error: --size: expected a size, but found '3Mib' (invalid digit or suffix)
+        \\stdout:
+        \\int: 0
+        \\size: 3145728
+        \\boolean: false
+        \\path: not-set
+        \\optional: null
+        \\choice: marlowe
```
batiati: Shouldn't we preserve the test intent here and use an invalid size for the flag `--size` instead?
We could also add checks to test both `InvalidSize` and `InvalidUnit`.

src/flags.zig
```diff
@@ -302,84 +302,158 @@ fn parse_value(comptime T: type, flag: []const u8, value: [:0]const u8) T {
...
+                const suffix_ = unit_[0];
+                const unit_kind = unit_[1];
+                if (std.ascii.eqlIgnoreCase(split.unit_input, suffix_)) {
+                    break :unit unit_kind;
+                }
+            } else {
+                return ByteSizeParseError.InvalidUnit;
+            }
+        else
+            ByteUnit.bytes;
```
batiati: Even better, we don't need the `units` tuples mapping text to enums anymore! 

```suggestion
        const unit = if (split.unit_input.len > 0)
            unit: inline for (std.enums.values(ByteUnit)) |tag| {
                if (std.ascii.eqlIgnoreCase(split.unit_input, @tagName(tag))) {
                    break :unit tag;
                }
            } else {
                return ByteSizeParseError.InvalidUnit;
            };
        else
            ByteUnit.bytes;
```

src/flags.zig
brson: You are right. I was thinking we needed a test for unusual spellings of "MiB", but there are existing tests that cover that case now, and I lost this case completely.

batiati: Maybe just a test for the deprecated form "--size=1MB", and for something completely wrong like `--size=INVALID`

brson: Done.

src/flags.zig
brson: Done.

batiati (approved):

PR #1568, opened 2024-02-19
ci: don't special case benchmark testing

matklad:
As per O(1) build files, benchmark is not special enough to warrant a dedicate space in the yamls. Move it to the integration tests. Even in debug, it's fast enough to run.

The release mode is exercised by the devhub.

### DIFF
--- a/src/shell.zig
+++ b/src/shell.zig
@@ -388,7 +388,7 @@ pub fn exec_options(
     }
 }
 
-/// Returns `true` if the command executed successfully with a zero exit code and an empty stderr.
+/// Returns `true` if the command executed successfully with a zero exit code.
 ///
 /// One intended use-case is sanity-checking that an executable is present, by running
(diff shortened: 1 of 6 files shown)


src/shell.zig
```diff
@@ -388,7 +388,7 @@ pub fn exec_options(
     }
 }
 
-/// Returns `true` if the command executed successfully with a zero exit code and an empty stderr.
+/// Returns `true` if the command executed successfully with a zero exit code.
```
matklad, line 391: This is the second time the stderr check gets in the way (as benchmark logs to stderr). Before that, I get a false error somewhere where I was running `tool --version`, and was getting print out to stderr about the tool being installed transparently by some sort of version manager. So I figured it's better to keep it simple and just not check for stderr.

cb22 (approved):

PR #1572, opened 2024-02-19
java: fix memory leak in tests

matklad:
NativeClient holds onto native memory, it should be deleted explicitly!

### DIFF
--- a/src/clients/java/src/test/java/com/tigerbeetle/AsyncRequestTest.java
+++ b/src/clients/java/src/test/java/com/tigerbeetle/AsyncRequestTest.java
@@ -20,51 +20,50 @@ public class AsyncRequestTest {
 
     @Test
     public void testCreateAccountsRequestConstructor() {
+        try (var client = getDummyClient()) {
+            var batch = new AccountBatch(1);
+            batch.add();
 
-        var client = getDummyClient();
-        var batch = new AccountBatch(1);
-        batch.add();
-
-        var request = AsyncRequest.createAccounts(client, batch);
-        assert request != null;
+            var request = AsyncRequest.createAccounts(client, batch);
+            assert request != null;
+        }
     }
 
     @Test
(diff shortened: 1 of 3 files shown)


src/clients/java/src/test/java/com/tigerbeetle/AsyncRequestTest.java
```diff
@@ -74,473 +73,479 @@ public void testConstructorWithClientNull() {
...
     @Test
     public void testEndRequestWithNullBuffer() throws Throwable {
+        try (var client = getDummyClient()) {
+            var batch = new TransferBatch(1);
+            batch.add();
 
-        var client = getDummyClient();
-        var batch = new TransferBatch(1);
-        batch.add();
+            var callback = new CallbackSimulator<CreateTransferResultBatch>(
```
matklad, line 164: Syntax for C#'s `using` is absolutely genius!

src/clients/java/src/test/java/com/tigerbeetle/AsyncRequestTest.java
```diff
@@ -20,51 +20,50 @@ public class AsyncRequestTest {
 
     @Test
     public void testCreateAccountsRequestConstructor() {
+        try (var client = getDummyClient()) {
```
batiati, line 23: Why don't we use a single static client here, like we did for the Integration tests?

batiati (commented):
🙌

src/clients/java/src/test/java/com/tigerbeetle/AsyncRequestTest.java
matklad: We _don't_ need single client here, as echoClient is cheap and we don't need to amortise. And with many clients, we exercise open/close code path more, which seems like a good thing. With a single client we probably wouldn't have noticed the leak here at all.

batiati (approved):

PR #1573, opened 2024-02-19
[Java] Re-enable Windows tests

batiati:


src/clients/java/src/test/java/com/tigerbeetle/AsyncRequestTest.java
```diff
@@ -24,7 +25,7 @@ public void testCreateAccountsRequestConstructor() {
             var batch = new AccountBatch(1);
             batch.add();
 
-            var request = AsyncRequest.createAccounts(client, batch);
+            var request = AsyncRequest.createAccounts(client, batch, null);
```
matklad: Looking at the CompletableFuture API, it seems like it would be idiomatic to use overloading here, rather than a nullable parameter?

batiati: Good point, we're using overloaded methods for the public API.
This case is just an internal class accepting a nullable parameter for the sake of simplicity.

However, even the public API can be considered non-idiomatic: the right call would be validating if the object is not null, and only accepting the default executor from the overloaded method which does not accept it as a parameter.

https://github.com/tigerbeetle/tigerbeetle/blob/c23944e7eb1671d70837c4a559bedc303bd3d74f/src/clients/java/src/main/java/com/tigerbeetle/Client.java#L138-L140

src/clients/java/src/main/java/com/tigerbeetle/AsyncRequest.java
```diff
@@ -81,7 +85,7 @@ protected void setResult(final TResponse result) {
 
         // To prevent the completion to run in the callback thread
         // we must call "completeAsync" instead of "complete".
-        future.completeAsync(() -> result);
+        future.completeAsync(() -> result, executor != null ? executor : future.defaultExecutor());
```
matklad: > Returns the default Executor used for async methods that do not specify an Executor. This class uses the [ForkJoinPool.commonPool()](https://docs.oracle.com/en%2Fjava%2Fjavase%2F11%2Fdocs%2Fapi%2F%2F/java.base/java/util/concurrent/ForkJoinPool.html#commonPool()) if it supports more than one parallel thread, or else an Executor using one thread per async task. This method may be overridden in subclasses to return an Executor that provides at least one independent thread.

So looks like this code will still trip thread-per-complete logic.


This makes me think:

_We_ create this future object, so we don’t actually make a meaningful use of virtually dispatch instance method to get the executor —- we know what the executor is going to be from the start.

*And* it looks like we actually want a different logic here —- if the user haven’t supplied the executor, we want to just use the common pool, even if it is single threaded (because we know we are a different thread. My *guess* is that JDK needs that hack to prevent submitting a future from a thread to the same thread).

So, what if we just *require* an executor at this level, and explicitly pass commonPoll at the API layer? That way, we’ll also avoid duplication at the API level, as methods wihout executor will delegate to methods with one.

batiati: >// To prevent the completion to run in the callback thread
>// we must call "completeAsync" instead of "complete"

Actually, this comment was wrong! We already have the result from the native thread, we don't need to defer it to an async executor. In fact, we already do the same with `completeExceptionally`.
This way, the API stays simple and the user can provide the executor for the completion (if they want to use `thenApply` and friends).

batiati: I can even remember when/why I wrote this comment, it was shortly after I found an issue in the C# client, where the async continuation would run on the native thread _if_ not stated the opposite:

https://github.com/tigerbeetle/tigerbeetle/blob/280faa17d491a01812321fd6571b7f2f0a03ec64/src/clients/dotnet/TigerBeetle/Request.cs#L145-L146

For some reason, I mixed both languages and applied the same "fix" for Java 🤦‍♂️

src/clients/java/src/test/java/com/tigerbeetle/IntegrationTest.java
```diff
@@ -1134,10 +1142,12 @@ public void testAsyncTasks() throws Throwable {
                 tasks[i] = client.createTransfersAsync(transfers).thenApplyAsync((result) -> {
                     semaphore.release();
                     return result;
-                });
+                }, executor);
             }
 
             // Wait for all threads.
+            CompletableFuture.allOf(tasks).join();
```
matklad:
```suggestion
            // Wait for all tasks.
            CompletableFuture.allOf(tasks).join();
```

matklad (approved):

matklad:
Ah wait, this seems buggy, but, as Java throws an exception in a different thread, we actually don't get an error here:

```
  Error: Exception in thread "Thread-4" com.tigerbeetle.AssertionError: CompletableFuture already completed
  	at com.tigerbeetle/com.tigerbeetle.AssertionError.assertTrue(AssertionError.java:14)
  	at com.tigerbeetle/com.tigerbeetle.AsyncRequest.setResult(AsyncRequest.java:82)
  	at com.tigerbeetle/com.tigerbeetle.Request.endRequest(Request.java:162)
  	at com.tigerbeetle/com.tigerbeetle.AsyncRequestTest$CallbackSimulator.run(AsyncRequestTest.java:581)
```

batiati:
> Ah wait, this seems buggy, but, as Java throws an exception in a different thread, we actually don't get an error here:

Yes, we can't do much from the tb_client’s thread, since we aready store any exception that can happen to be handled by the completion. But if the completion is already completed, the assertion is more for logging.

matklad:
That was a genuine bug I think, fixed in https://github.com/tigerbeetle/tigerbeetle/pull/1575

PR #1574, opened 2024-02-20
ci: remove coverage related dead code

matklad:
We haven't got to reviving coverage yet, and now GitHub complains that upload-artifact action needs to be updated to use newer node.

Let's remove the code --- it will be there in git history anyway.

Perhaps coverage shouldn't even be a part of CI gate, and instead run as a part of devhub benchmarking.


matklad:
Should fix these warnings: 

image

cb22 (approved):

PR #1576, opened 2024-02-20
vsr: synchronize op and commit_max updates on_prepare

matklad:
Before, op was updated first, and commit_max at the end, breaking

```
assert(self.commit_max >= self.op -| constants.pipeline_prepare_queue_max);
```

invariant. Broken invariant was observed by, eg, op_repair_min (see the first commit).

To fix this:

* rearrange the code a bit, to separate logical update to commit_max from attempting to commit stuff
* move logical update forward in on_prepare flow

### DIFF
--- a/src/vsr/replica.zig
+++ b/src/vsr/replica.zig
@@ -5074,6 +5074,7 @@ pub fn ReplicaType(
             if (self.status == .recovering) assert(self.solo());
             assert(self.op >= self.op_checkpoint());
             assert(self.op <= self.op_prepare_max());
+            assert(self.commit_max >= self.op -| constants.pipeline_prepare_queue_max);
 
             const op = op: {
                 if (self.primary_index(self.view) == self.replica) {
--- a/src/vsr/replica.zig
+++ b/src/vsr/replica.zig
@@ -1394,14 +1394,11 @@ pub fn ReplicaType(
             assert(message.header.op > self.op);
             assert(message.header.op > self.commit_min);
 
-            defer {
-                if (self.backup()) {
-                    // A prepare may already be committed if requested by repair() so take the max:
-                    self.advance_commit_max(message.header.commit, @src().fn_name);
-                    self.commit_journal();
-                    assert(self.commit_max >= message.header.commit);
-                }
+            if (self.backup()) {
+                self.advance_commit_max(message.header.commit, @src().fn_name);
+                assert(self.commit_max >= message.header.commit);
             }
+            defer if (self.backup()) self.commit_journal();
 
             // Verify that the new request will fit in the WAL.
             if (message.header.op > self.op_prepare_max()) {
@@ -3030,7 +3027,6 @@ pub fn ReplicaType(
                     assert(self.commit_max == self.commit_min);
                     assert(self.commit_max == self.op - self.pipeline.queue.prepare_queue.count);
                 }
-
             }
 
             if (commit > self.commit_max) {
--- a/src/vsr/replica.zig
+++ b/src/vsr/replica.zig
@@ -784,7 +784,8 @@ pub fn ReplicaType(
 
             if (self.solo()) {
                 if (self.commit_min < self.op) {
-                    self.commit_journal(self.op);
+                    self.advance_commit_max(self.op, @src().fn_name);
+                    self.commit_journal();
 
                     // Recovery will complete when commit_journal finishes.
                     assert(self.status == .recovering);
@@ -798,7 +799,7 @@ pub fn ReplicaType(
                     }
                 } else {
                     if (self.status != .recovering_head) {
-                        self.commit_journal(self.commit_max);
+                        self.commit_journal();
                     }
                 }
             }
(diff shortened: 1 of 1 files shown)


src/vsr/replica.zig
```diff
@@ -3777,14 +3786,7 @@ pub fn ReplicaType(
...
             assert(self.commit_min == prepare.header.op);
-            if (self.commit_min > self.commit_max) {
-                log.debug("{}: commit_op: advancing commit_max={}..{}", .{
-                    self.replica,
-                    self.commit_max,
-                    self.commit_min,
-                });
-                self.commit_max = self.commit_min;
-            }
+            self.advance_commit_max(self.commit_min, @src().fn_name);
```
matklad, line 3792: This feels something suspicious. I think it would be better to update `commit_max` _before_ the primary starts committing, not after it is done. Though I guess for solo mode updating later is actually desirable?

src/vsr/replica.zig
```diff
@@ -1393,14 +1393,11 @@ pub fn ReplicaType(
             assert(message.header.op > self.op);
             assert(message.header.op > self.commit_min);
```
sentientwaffle, line 1395: Maybe add a `assert(message.header.op > message.header.commit);` to emphasize that the `self.advance_commit_max(message.header.commit, @src().fn_name);` below is ok even though we won't have loaded `message.header` into our journal yet.

src/vsr/replica.zig
```diff
@@ -309,7 +309,6 @@ pub fn ReplicaType(
         /// * never decreases.
         /// Invariants (status=normal primary):
         /// * `replica.commit_max = replica.commit_min`.
-        /// * `replica.commit_max = replica.op - pipeline.queue.prepare_queue.count`.
```
sentientwaffle: Where does this not hold?

matklad: Ah, forgot to mention that I was confused by this. This almost never holds? For example, in an idle cluster `replica.commit_max = replica.op`. Or am I misreading this?

src/vsr/replica.zig
matklad: advancing commit_max should be ok even if we don't have a header? We allow `maybe(self.op < self.commit_max)`

src/vsr/replica.zig
sentientwaffle: `pipeline.queue.prepare_queue.count` is the number of items currently in the pipeline, not the capacity. So in an idle cluster `pipeline.queue.prepare_queue.count == 0`.

src/vsr/replica.zig
```diff
@@ -784,7 +783,8 @@ pub fn ReplicaType(
 
             if (self.solo()) {
                 if (self.commit_min < self.op) {
-                    self.commit_journal(self.op);
+                    self.advance_commit_max(self.op, @src().fn_name);
```
matklad, line 787: not sure whether I like `@src().fn_name`. On the one hand, it is cryptic and too clever. On the other hand, we did have bugs where we moved code around but didn't update call-sites. 

If we double-down here, we might pass the whole `@src()` in even

src/vsr/replica.zig
sentientwaffle: Ah, good point!

src/vsr/replica.zig
matklad: Ah, indeed, I misread that as `pipeline_prepare_queue_max` :facepalm:

src/vsr/replica.zig
sentientwaffle: :thinking: I like the "on primary, commit_min == commit_max invariant" that we have.

src/vsr/replica.zig
matklad: Ok, now advance_commit_max actually checks that invariant (had to move pipeline prepare popping to commit_op for that)

src/vsr/replica.zig
matklad: Ahh, this feels extra tricky --- it seems like we could be advancing `commit_max` here even if we are no longer primary, if view change happens while we are committing.

matklad:
Apparently, it's better not to forec-push VOPR-labeled pull requests!

src/vsr/replica.zig
matklad: :thinking: if we allow primary's `commit_max` to run ahead of `commit_min`, that might increase latency tolerance. If the primary is slow to commit, clients could get replies from backups which commit ahead of the primary.

src/vsr/replica.zig
```diff
@@ -3436,17 +3447,6 @@ pub fn ReplicaType(
...
-                    const prepare = self.pipeline.queue.pop_prepare().?;
-                    defer self.message_bus.unref(prepare.message);
-
-                    assert(prepare.message.header.command == .prepare);
-                    assert(prepare.message.header.checksum == self.commit_prepare.?.header.checksum);
-                    assert(prepare.message.header.op == self.commit_min);
-                    assert(prepare.message.header.op == self.commit_max);
-                    assert(prepare.ok_quorum_received);
-                }
-
```
sentientwaffle, line 3449: I did like how "pop prepare" was immediately followed by "pop request -> push prepare to pipeline". But `commit_op` is already pretty gigantic :sweat:

sentientwaffle (approved):

PR #1579, opened 2024-02-20
vsr: don't repair potentially unavailable ops

### LINKED ISSUE #1378: Allow primaries with a dirty journal

matklad:
op_repair_min tries to be smart and optimistic and repair _more_, which
sometimes lead to unavailability, due to two issues:

- primary tries to repair more than backups, but because backups repair
  less, those prepares might be unavailable! That is, we might suffer
  _more_ than `f` failures  for particular prepare over time, because it
  is not continuously repaired on the backups
- after state sync, replicas simply do not have any prepares beyond
  checkpoint.

Simplify op_repair_min logic to always repair from the latest
"confirmed" (committed on top of) checkpoint.

That is, replicas _right_ at the trigger repair the entire log wrap, in
addition to having the checkpoint. This ensures smooth hand off of
durability: prepares stop being repaired only when there's a quorum of
replicas at a checkpoint including the prepare.

Closes: https://github.com/tigerbeetle/tigerbeetle/issues/1378

### DIFF
--- a/src/vsr/replica.zig
+++ b/src/vsr/replica.zig
@@ -2823,7 +2823,7 @@ pub fn ReplicaType(
             // - we have not made any progress committing for some interval.
             // It is possible that we could in fact repair and progress anyway, but this is not
             // guaranteed: the rest of the cluster no longer repairs the relevant prepares.
-            // We  might be stuck waiting for a prepare or block which will never arrive.
+            // We might be stuck waiting for a prepare or block which will never arrive.
 
             log.warn("{}: on_repair_sync_timeout: start sync; lagging behind cluster " ++
                 "(op_head={} commit_min={} commit_max={} commit_stage={s})", .{
@@ -5048,13 +5048,13 @@ pub fn ReplicaType(
 
         /// Returns the oldest op that the replica must/(is permitted to) repair.
         ///
-        /// Safety condition: repairing an old op must not overwrite a newer op from the next  wrap.
+        /// Safety condition: repairing an old op must not overwrite a newer op from the next wrap.
         ///
         /// Availability condition: each committed op must be present either in a quorum of WALs or
         /// it in a quorum of checkpoints.
         ///
-        /// If op=trigger+1 is committed, the corresponding checkpoint is confirmed to be present on
-        /// a quorum of replicas. Repairig all ops since the latest confirmed checkpoint satisfies
+        /// If op=trigger+1 is committed, the corresponding checkpoint is durably present on
+        /// a quorum of replicas. Repairing all ops since the latest durable checkpoint satisfies
         /// both conditions.
         ///
         /// When called from status=recovering_head or status=recovering, the caller is responsible
--- a/src/vsr/replica.zig
+++ b/src/vsr/replica.zig
@@ -1458,6 +1458,7 @@ pub fn ReplicaType(
             });
             assert(message.header.op == self.op + 1);
             assert(message.header.op <= self.op_prepare_max());
+            assert(message.header.op - self.op_repair_min() < constants.journal_slot_count);
             self.op = message.header.op;
             self.journal.set_header_as_dirty(message.header);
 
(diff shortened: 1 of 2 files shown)


src/vsr/replica.zig
```diff
@@ -5083,44 +5065,18 @@ pub fn ReplicaType(
...
-                        // Don't allow "op_repair_min > op_head".
-                        break :op self.op_checkpoint();
-                    } else {
-                        // +1 because the primary (and other replicas) may overwrite op_checkpoint.
-                        // And we don't need op_checkpoint's prepare/header in our log if we crash.
-                        break :op self.op_checkpoint() + 1;
-                    }
+                    break :op (self.op_checkpoint() + 1) -| constants.vsr_checkpoint_interval;
                 }
             };
```
matklad, line 5083: Tried to add an extra assert here:

```
assert(self.prepare_max() - op < constants.journal_slot_count);
```

but that doesn't actually work for a funny reason:

* it can be violated if your `op_repair_min` still uses the previous checkpoint, and `prepare_max` already picks the next one (that is, commit_max is right on the trigger).
* but it actually isn't really violated --- when you receives that `op_prepare_max`, you advance your `commit_max`, which causes you to move op_repair_min by a checkpoint forward as well.

matklad: Ah, but wait, I _can_ assert that `on_prepare` instead!

src/vsr/replica.zig
```diff
@@ -2830,11 +2819,11 @@ pub fn ReplicaType(
...
             // At this point:
-            // - we know we are lagging behind the cluster by at least a checkpoint, and
-            // - our next commit may not be repairable, and
+            // - we know that the cluster committed atop of a future a checkpoint, and
             // - we have not made any progress committing for some interval.
-            // It is possible that we could in fact repair and progress anyway, but usually
-            // we are stuck waiting for a prepare or block which will never arrive.
+            // It is possible that we could in fact repair and progress anyway, but this is not
+            // guaranteed: the rest of the cluster no longer repairs the relevant prepares.
+            // We  might be stuck waiting for a prepare or block which will never arrive.
```
```suggestion
            // We might be stuck waiting for a prepare or block which will never arrive.
```

(Extra space.)

src/vsr/replica.zig
```diff
@@ -5058,21 +5047,15 @@ pub fn ReplicaType(
...
-        ///   - backups don't repair checkpointed ops because:
-        ///     - repairing an op from the previous checkpoint might "clean" a slot which belongs
-        ///       to a corrupt entry from the current WAL-wrap, leading to the backup incorrectly
-        ///       nacking the latter entry.
-        ///     - there is no guarantee that they will ever be available (if our head is behind),
-        ///       and we don't want to stall the new view startup.
-        ///   - primaries do repair checkpointed ops so that they can help lagging backups catch up,
-        ///     as long as the op is new enough to be present in the WAL of some other replica and
-        ///     is from the previous checkpoint.
+        /// Safety condition: repairing an old op must not overwrite a newer op from the next  wrap.
```
```suggestion
        /// Safety condition: repairing an old op must not overwrite a newer op from the next wrap.
```

Extra space

src/vsr/replica.zig
```diff
@@ -5058,21 +5047,15 @@ pub fn ReplicaType(
...
+        ///
+        /// Availability condition: each committed op must be present either in a quorum of WALs or
+        /// it in a quorum of checkpoints.
+        ///
+        /// If op=trigger+1 is committed, the corresponding checkpoint is confirmed to be present on
+        /// a quorum of replicas. Repairig all ops since the latest confirmed checkpoint satisfies
```
```suggestion
        /// a quorum of replicas. Repairing all ops since the latest confirmed checkpoint satisfies
```

sentientwaffle (commented):
The code all LGTM!

sentientwaffle (approved):

src/vsr/replica.zig
```diff
@@ -5058,21 +5047,15 @@ pub fn ReplicaType(
...
-        ///       and we don't want to stall the new view startup.
-        ///   - primaries do repair checkpointed ops so that they can help lagging backups catch up,
-        ///     as long as the op is new enough to be present in the WAL of some other replica and
-        ///     is from the previous checkpoint.
+        /// Safety condition: repairing an old op must not overwrite a newer op from the next wrap.
+        ///
+        /// Availability condition: each committed op must be present either in a quorum of WALs or
+        /// it in a quorum of checkpoints.
+        ///
+        /// If op=trigger+1 is committed, the corresponding checkpoint is durably present on
```
matklad, line 5056: Ah,  but I think this is wrong after https://github.com/tigerbeetle/tigerbeetle/pull/1508!

The condition is now `op=prepare_max + 1`, not `op=trigger + 1` I think!

PR #1580, opened 2024-02-21
vsr: harmonize and automate method logging pattern

matklad:
Instead of manually passing in method name, pass `@src()`. This is shorter and more resilient to refactors, but does look a bit cryptic.

On the other hand, once you've learned the pattern, it becomes instantly recognizable.

### DIFF
--- a/src/vsr/replica.zig
+++ b/src/vsr/replica.zig
@@ -2,6 +2,7 @@ const std = @import("std");
 const Allocator = std.mem.Allocator;
 const assert = std.debug.assert;
 const maybe = stdx.maybe;
+const SourceLocation = std.builtin.SourceLocation;
 
 const constants = @import("../constants.zig");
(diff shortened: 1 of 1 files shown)


src/vsr/replica.zig
```diff
@@ -7114,7 +7115,7 @@ pub fn ReplicaType(
                 // "`replica.op` exists" invariant may be broken briefly between
                 // set_op_and_commit_max() and replace_header().
 
-                self.set_op_and_commit_max(header_head.op, commit_max, "on_do_view_change");
+                self.set_op_and_commit_max(header_head.op, commit_max, @src());
```
matklad, line 7118: This was `on_do_view_change`, now it is `primary_set_log_from_do_view_change_messages`

sentientwaffle (approved):

PR #1581, opened 2024-02-21
nyrkiö

matklad:
https://nyrk.io/ is a cloud CI benchmark tracking service.

It is very similar to what devhub is doing, except that it actually is a
real project, and not just me copy-pasting JavaScript from
https://deno.com/benchmarks.

Otherwise, it is very similar --- their data model is isomorphic to what
we already use.

In this PR, we duplicate data to both devhubdb and nyrk. I think we'll
continue in this mode for some time, but I can see us switching to nyrk
100% in the future.

In a follow up, I do want to switch to nyrk-shaped json for devhubdb as
well, as I like their version more (esp grouping attributes _separately_
from the timestamp, which is the primary key), but I don't want to
change stuff on our end right away.

As nyrk is a 3rd party service, I don't fail CI job if we fail to
upload.

### DIFF
--- a/src/scripts/devhub.zig
+++ b/src/scripts/devhub.zig
@@ -13,6 +13,8 @@ const std = @import("std");
 const stdx = @import("../stdx.zig");
 const Shell = @import("../shell.zig");
 
+const log = std.log;
+
 pub const CliArgs = struct {
     sha: []const u8,
 };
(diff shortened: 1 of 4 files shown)


src/scripts/devhub.zig
```diff
@@ -94,17 +130,33 @@ fn upload_run(shell: *Shell, run: Run) !void {
...
+        branch: []const u8,
+        git_commit: []const u8,
+    },
+};
 
-    return try std.fmt.parseInt(u64, cut.prefix, 10);
+fn upload_nyrkio(shell: *Shell, run: NyrkioRun) !void {
+    const token = try shell.env_get("NYRKIO_TOKEN");
+    const payload = try std.json.stringifyAlloc(shell.arena.allocator(), [_]NyrkioRun{run}, .{});
+    try shell.exec(
```
cb22, line 151: No love for `std.http.Client`? :wink:

cb22 (approved):

src/scripts/devhub.zig
matklad: Wait, we _already_ have http.Client? Didn't realize that, thought that's the next release. But, given https://github.com/ziglang/zig/pull/18955, I'd probably let it bake a bit more until we depend on it

matklad:
Failure feels like something's wrong with actions/checkpout?

PR #1583, opened 2024-02-21
go: re-enable windows tests

matklad:


matklad:
perhaps could land directly with https://github.com/tigerbeetle/tigerbeetle/pull/1804

PR #1584, opened 2024-02-21
LSM: Make the object cache optional and remove redundant stash.

batiati:
Related to #1454, this PR also makes the `SetAssociativeCache` nullable.
See the commit messages for more details.

Default cache vs zero cache for `Transfer`, `Posted` and `AccountHistory`:

<table>
<tr>
<th>Before</th>
<th>After</th>
</tr>
<tr><td>
<code>1266 batches in 34.26 s
load offered = 1000000 tx/s
load accepted = 291921 tx/s
batch latency p1 = 0 ms
batch latency p10 = 5 ms
batch latency p20 = 7 ms
batch latency p30 = 8 ms
batch latency p40 = 9 ms
batch latency p50 = 12 ms
batch latency p60 = 14 ms
batch latency p70 = 18 ms
batch latency p80 = 23 ms
batch latency p90 = 31 ms
batch latency p95 = 36 ms
batch latency p99 = 490 ms
batch latency p100 = 548 ms
rss = 4152885248 bytes
</code> </td>
<td>
<code>1240 batches in 34.45 s
load offered = 1000000 tx/s
load accepted = 290241 tx/s
batch latency p1 = 2 ms
batch latency p10 = 6 ms
batch latency p20 = 7 ms
batch latency p30 = 7 ms
batch latency p40 = 9 ms
batch latency p50 = 12 ms
batch latency p60 = 15 ms
batch latency p70 = 19 ms
batch latency p80 = 24 ms
batch latency p90 = 30 ms
batch latency p95 = 36 ms
batch latency p99 = 483 ms
batch latency p100 = 536 ms
rss = 3652321280 bytes
</code> </td> </tr>
</table>

**EDIT**

**Bonus:** Running the benchmark with `--cache-accounts=0` (zero cache for all objects)  🤯

```
1251 batches in 34.43 s
load offered = 1000000 tx/s
load accepted = 290407 tx/s
batch latency p1 = 1 ms
batch latency p10 = 5 ms
batch latency p20 = 6 ms
batch latency p30 = 6 ms
batch latency p40 = 8 ms
batch latency p50 = 11 ms
batch latency p60 = 14 ms
batch latency p70 = 17 ms
batch latency p80 = 22 ms
batch latency p90 = 28 ms
batch latency p95 = 33 ms
batch latency p99 = 504 ms
batch latency p100 = 547 ms

rss = 3516792832 bytes
```


src/lsm/cache_map.zig
```diff
@@ -31,39 +31,33 @@ pub fn CacheMapType(
...
+            Key,
+            Value,
+            key_from_value,
+            hash_from_key,
+            .{},
+        );
+
+        pub const Map = std.HashMapUnmanaged(
+            Key,
+            Value,
```
matklad, line 48: Hm, I am rather sure that this should remain `Value, void,` --- key is a subset of value, so this effectively 2xes memory usage, mostly negating the removal of the second stash? 

I also wouldn't worry about copying in `HashMapContextValue.eql` --- the function is marked as inline, so there isn't any parameter passing at all there.

src/lsm/cache_map.zig
```diff
@@ -99,95 +92,84 @@ pub fn CacheMapType(
...
-                        // critical. Since we use HashMaps with no Value, putAssumeCapacity _will
-                        // not_ clobber the existing value.
-                        const gop = self.stash_1.getOrPutAssumeCapacity(evicted);
-                        gop.key_ptr.* = evicted;
+                        assert(evicted_key != key_from_value(value));
+                        self.stash.putAssumeCapacity(evicted_key, evicted.*);
+                    },
+                    .update => {
+                        // The old version was evicted.
+                        assert(evicted_key == key_from_value(value));
```
matklad: :heart:

src/lsm/cache_map.zig
```diff
@@ -203,41 +185,44 @@ pub fn CacheMapType(
         //       to the scope rollback log.
         //    b. If the item doesn't exist in the stash, it was an insert. Append a tombstone to
         //       the scope rollback log.
-        fn upsert_scope_opened(self: *Self, value: *const Value) void {
+        fn upsert_scope_opened(self: *CacheMap, value: *const Value) void {
```
matklad: :+1:

src/lsm/tree.zig
```diff
@@ -319,6 +319,17 @@ pub fn TreeType(comptime TreeTable: type, comptime Storage: type) type {
         /// ascertain the existence of the key are not in the Grid cache, it bails out.
         /// The returned `.positive` Value pointer is only valid synchronously.
         pub fn lookup_from_levels_cache(tree: *Tree, snapshot: u64, key: Key) LookupMemoryResult {
+            if (constants.verify) {
+                // The object cache should be used instead.
+                tree.table_mutable.sort();
+                const not_found = tree.table_mutable.get(key) == null;
+                assert(not_found);
+            }
```
matklad: Ahh, this one is tantalizing. I see what you are trying to assert here and I agree that that would be massively useful. 

I don't htink we can do this though, unfortunatly. 

The big problem with the present approach is that it changes semantics with `constants.verify` active --- we actually go and _do_ sort objects in place, which has downstream implication (by the time we hit compation, the stuff will be mostly sorted).

Even without that, this is O(1) -> O(N) check under verify in data plane, which I think we shouldn't be doing: https://github.com/tigerbeetle/tigerbeetle/pull/1585


Also, is this correct? I imagine if we have `put` and then `remove`, we won't have the key in object cache, but would still have two entries in table mutable (sorting the table doesn't cancel out tombstones)

src/tigerbeetle/benchmark_driver.zig
```diff
@@ -108,16 +112,65 @@ const TigerBeetleProcess = struct {
...
-    );
+    var arena = std.heap.ArenaAllocator.init(allocator);
+    defer arena.deinit();
+
+    var start_args = std.ArrayListUnmanaged([]const u8){};
+    try start_args.append(arena.allocator(), options.tigerbeetle);
+    try start_args.append(arena.allocator(), "start");
+    try start_args.append(arena.allocator(), "--addresses=0");
+
+    // Repassing the cache options to the tigerbeetle process:
```
matklad: s/Repassing/Forward/

src/tigerbeetle/benchmark_driver.zig
```diff
@@ -108,16 +112,65 @@ const TigerBeetleProcess = struct {
...
+    }
+
+    if (options.args.cache_grid) |cache_grid| {
+        try start_args.append(
+            arena.allocator(),
+            try std.fmt.allocPrint(arena.allocator(), "--cache-grid={s}", .{
+                cache_grid,
+            }),
+        );
+    }
```
matklad, line 144: Maybe inline loop here? Looks a bit too repetetive

matklad: Alternatively, maybe just allow forwarding arbitrary arguments? Though it feels like hard-codig specific oness is enough for now

src/lsm/cache_map.zig
batiati: It's not about the inline functions `eql` and `hash`, but all other places where we need to copy `value` instead of just passing the `key` (e.g., `get,` `getOrPut,` `remove`).

Since objects are usually much larger than the key, using K/V instead of Value/Void increased ~56MiB, which may be a valid trade-off in the hot path, although the benchmark is too noisy to catch any real difference.

matklad: Hm, this still doesn't feel quite right: we do duplicate data in memory. 56 megs is indeed not much in term of RAM, _but_ it is quite a bit in terms of CPU cache, and we already have values in CPU cache. 

Let me take closer look at the get and remove changes here...

matklad: That is, my current understndign that storing both Value and a Key (which is a subset of Value) should actually decrease perf, if we think from first principles. But indeed maybe some things get faster due to artifiacts of Zig current implementation...

batiati: Since the benchmark didn't catch any meaningful difference, let's stick with the memory we saved with Value/Void (the last commit can be reverted to compare both cases).

matklad: :+1: I mostly want to make sure that I am _not_ missing anything, and that storing value is indeed redundant, abstractly speakin!

batiati: FYI, [historically](https://github.com/tigerbeetle/tigerbeetle/commit/9c8d816577184558db7ad25eaa03dad5861948a9#diff-be4726c0a7b96915c03bad6692df47aa2c623e8066d22367a5e9c1755857dfe8R363) we always used Value/Void

```zig
 const PrefetchObjects = std.HashMapUnmanaged(Object, void, PrefetchObjectsContext, 70);
```

matklad: Yeah, I am even wonding if we should API-wrap an `std.HashMapUnmanaged` to have a signature identical to that of SetAssociativeCache (where you pass _both_ key and value, but where the key isn't actually stored, and only used for key_from_value function). Probably not worth doing just for the sake of more proper feng-shui API.

batiati: I'd argue that `CacheMap` _is_ that  common API wrapper 🤔

src/lsm/cache_map.zig
```diff
@@ -203,73 +194,96 @@ pub fn CacheMapType(
...
-                    },
-                    .insert => {
-                        // Case 2.
-                        const gop = self.stash_1.getOrPutAssumeCapacity(evicted);
-                        gop.key_ptr.* = evicted;
-
-                        // Case 3 below handles appending into the rollback log if needed.
-                    },
+            const key = key_from_value(value);
+            if (self.cache) |*cache| {
```
matklad, line 170: Not a fan of duplication all over the place. I wonder, would it be beneficial to introduce

```
fn cache_get(cache_map: *CacheMap)
fn cache_upsert(cache_map: *CacheMap)
fn cache_remove(cache_map: *CacheMap)
```

which encapsulate null checking?

I _think_ null cache behaves as infinitely-sized cache --- it never evicts anything, and always reports that a new value was inserted? 

Ie, it's hard to verify at the glance that the logic in two branches are compatible here:


```
            if (self.cache) |*cache| {
                const maybe_removed = cache.remove(key);

                if (self.scope_is_active) {
                    // TODO: Actually, does the fuzz catch this...
                    self.scope_rollback_log.appendAssumeCapacity(
                        maybe_removed orelse
                            self.stash.getKey(tombstone_from_key(key)) orelse return,
                    );
                }

                // We always need to try remove from the stash; since it could have a stale value.
                // The early return above is OK - if it doesn't exist, there's nothing to remove.
                _ = self.stash.remove(tombstone_from_key(key));
            } else {
                const maybe_removed = self.stash.fetchRemove(tombstone_from_key(key));
                if (self.scope_is_active) {
                    if (maybe_removed) |kv| {
                        // TODO: Actually, does the fuzz catch this...
                        self.scope_rollback_log.appendAssumeCapacity(kv.key);
                    }
                }
            }

```

src/lsm/cache_map.zig
```diff
@@ -203,73 +194,96 @@ pub fn CacheMapType(
...
+                    );
+                }
 
-            // We always need to try remove from the stash; since it could have a stale value.
-            // The early return above is OK - if it doesn't exist, there's nothing to remove.
-            _ = self.stash_1.remove(tombstone_from_key(key));
-            _ = self.stash_2.remove(tombstone_from_key(key));
+                // We always need to try remove from the stash; since it could have a stale value.
+                // The early return above is OK - if it doesn't exist, there's nothing to remove.
+                _ = self.stash.remove(tombstone_from_key(key));
```
matklad: Drive by, but I think we are doing extra hash map lookup here!

```
self.stash.getKey(tombstone_from_key(key))
self.stash.remove(tombstone_from_key(key));
```

these two will needlessly do the same work twice. If we are removig from stash unconditonally, mabybe pull `fetchRemove` to the top once, like we do in the other branch?

src/lsm/cache_map.zig
batiati: I simplified the code by decoupling the upsert from the rollback logic.

batiati:
As an experiment, I tried to get rid of both stashes.
Making `table_memory` a hybrid of an array + map[key, index] to keep it optimized to the already sorted workloads.

This impl is not memory optimized (not every `table_memory` needs a map), but it's complete enough to measure a performance hit in order of **7%** slower. Thanks, but no thanks!

https://github.com/tigerbeetle/tigerbeetle/blob/a39d05f816ba6dbce3afd33b488450703ec72746/src/lsm/table_memory.zig#L109-L157

matklad (approved):

PR #1585, opened 2024-02-23
docs: document how constants.verify should be used

matklad:
Forgot to uplift this from slack a while ago.


sentientwaffle (approved):
Perfect!

PR #1596, opened 2024-02-26
[Java] Register a Cleaner in addition to AutoCloseable

batiati:
Also reverts #1572 as a strategy for executing the `cleaner` during tests.

### DIFF
--- a/src/clients/java/src/main/java/com/tigerbeetle/NativeClient.java
+++ b/src/clients/java/src/main/java/com/tigerbeetle/NativeClient.java
@@ -10,10 +10,11 @@ final class NativeClient implements AutoCloseable {
 
     /*
      * Holds the `handle` in an object instance detached from `NativeClient` to provide state for
-     * the cleaner to dispose native memory when the `Client` instance is GCed.
+     * the cleaner to dispose native memory when the `Client` instance is GCed. Also implements
+     * `Runnable` to be usable as the cleaner action.
      * https://docs.oracle.com/javase%2F9%2Fdocs%2Fapi%2F%2F/java/lang/ref/Cleaner.html
      */
-    private static final class NativeHandle {
+    private static final class NativeHandle implements Runnable {
         // Keeping the contextHandle and a reference counter guarded by
         // atomics in order to prevent the client from using a disposed
         // context during `close()`.
@@ -70,6 +71,11 @@ final class NativeClient implements AutoCloseable {
                 }
             }
         }
+
+        @Override
+        public void run() {
+            close();
+        }
     }
 
     static {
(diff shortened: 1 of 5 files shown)


src/clients/java/src/main/java/com/tigerbeetle/NativeClient.java
```diff
@@ -36,55 +102,29 @@ private static void assertArgs(final byte[] clusterID, final String addresses,
     }
 
     private NativeClient(final long contextHandle) {
-        this.contextHandle = new AtomicLong(contextHandle);
-        this.contextHandleReferences = new AtomicLong(0L);
+        try {
+            this.handle = new NativeHandle(contextHandle);
+            this.cleanable = cleaner.register(this, () -> handle.close());
```
matklad: >The cleaning action could be a lambda but all too easily will capture the object reference, by referring to fields of the object being cleaned, preventing the object from becoming phantom reachable. Using a static nested class, as above, will avoid accidentally retaining the object reference.

https://docs.oracle.com/javase%2F9%2Fdocs%2Fapi%2F%2F/java/lang/ref/Cleaner.html

Isn't this _exactly_ the problem we have here? `() -> handle.close()` is `() -> this.handle.close()`, so it does close over the entire `NativeClient`, preventing it from becoming unreachable?

batiati: You're right, I hoped that `() -> handle.close()` only captured the `handle` instance, not `this`, but I'd need a scoped variable for that.

Lambda captures are so subtle that I'm not sure if this shortcut is worthy, I'll use the verbose equivalent with a `Runnable` action ... 

```
private static class CleanerAction implements Runnable {
    private NativeHandle handle;

    pub void run() { 
        // ...
    }
}
```

src/clients/java/src/main/java/com/tigerbeetle/NativeClient.java
```diff
@@ -36,55 +108,29 @@ private static void assertArgs(final byte[] clusterID, final String addresses,
...
     private NativeClient(final long contextHandle) {
-        this.contextHandle = new AtomicLong(contextHandle);
-        this.contextHandleReferences = new AtomicLong(0L);
+        try {
+            this.handle = new NativeHandle(contextHandle);
+            this.cleanable = cleaner.register(this, handle);
+        } catch (Throwable forward) {
+            clientDeinit(contextHandle);
+            throw forward;
+        }
```
matklad, line 117: Isn't this try-finally ?

batiati: I don't think so, `clientDeinit(contextHandle)` is only supposed to run as a defense-in-depth if something weird happens during the constructor.

matklad: Ahh, it's errdefer, not defer :facepalm:

batiati: I can't live without `errdefer` anymore ❤️

matklad (approved):
This is unrelated, but I must mention that I still find our synchronized-then-spin-wait loop deeply unsettling. I wish we did this in a more straightforward way

PR #1606, opened 2024-02-27
node: update node.zig

eltociear:
occured -> occurred


batiati (approved):
Thank you @eltociear!

PR #1607, opened 2024-02-27
tests: introduce explicit coverage marks

matklad:
This file piggy-backs on the logging infrastructure to implement explicit coverage marks:
     <https://ferrous-systems.com/blog/coverage-marks/>

See the docs and tutorial test in `src/testing/covered.zig` for API details

This PR doesn't actually _use_ coverage marks much, as I want to nail down the API first.

Design notes:

- gate all the logic with `comptime if (builtin.test)`
- assume single-threaded test execution. It looks like, if Zig will run tests in parallel, it'll use separate processes https://github.com/ziglang/zig/issues/15953. And if it does multithreading, replacing this with thread-locals won't be hard (and we have a plan to thread our own logger instance anyway)
- intercept not all `log.debug` calls, but only explicitly marked ones: `log.covered.debug`. This makes prod code a touch noisier, but I think it's worth it, as it makes is easy to go to the relevant test
- match on `comptime fmt: []const u8` rather than on the actual formatted log line. This is less flexible, but faster and allows for grepping string literals. 
- do not `defer` checks that the mark was hit, require an explicit `try mark.expect_hit()` calls. I am ambivalent here, but feels like defers might introduce extra scopes where non would be necessary otherwise. Each mark does assert that it starts in a pristine state, so forgetting to assert the marks should be noticiable via an assertion failure in a subesquent test.
- this doesn't address the problem of "orphaned" marks. That's best addressed through a tidy checks that collects all allegedly `.covered` marks and cross-checks that set against what we actually `covered.mark()`. This is left as a future exercise.

### DIFF
--- a/src/testing/covered.zig
+++ b/src/testing/covered.zig
@@ -1,0 +1,120 @@
+//! We'll get you covered!
+//!
+//! This file piggy-backs on the logging infrastructure to implement explicit coverage marks:
+//!     <https://ferrous-systems.com/blog/coverage-marks/>
+//!
+//! In production code, you can mark certain log lines as "this should be covered by a test".
+//! In test code, you can then assert that a _specific_ test covers a specific log line. The two
+//! benefits are:
+//! - tests are more resilient to refactors
+//! - production code is more readable (you can immediately jump to a specific test)
+test "tutorial" {
+    // Import by a qualified name.
+    const covered = @import("./covered.zig");
+
+    const production_code = struct {
+        // In production code, wrap the logger.
+        const log = covered.wrap_log(std.log.scoped(.my_module));
+
+        fn function_under_test(x: u32) void {
+            if (x % 2 == 0) {
+                // Both `log.info` and log.covered.info` are available.
+                // Only second version records coverage.
+                log.covered.info("x is even (x={})", .{x});
+            }
+        }
+    };
+
+    // Create a mark with the `mark` function...
+    const m = covered.mark("x is even");
+    production_code.function_under_test(92);
+    try m.expect_hit(); // ... and don't forget to assert at the end!
+}
+
+const std = @import("std");
+const assert = std.debug.assert;
+const builtin = @import("builtin");
+
+const GlobalStateType = if (builtin.is_test) struct {
+    mark_name: ?[]const u8 = null,
+    mark_hit_count: u32 = 0,
+} else void;
+
+/// Stores the currently active mark and its hit count. State is not synchronized and assumes
+/// single threaded execution.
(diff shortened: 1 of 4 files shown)


src/testing/covered.zig
```diff
@@ -0,0 +1,124 @@
...
+//!
+//! This file piggy-backs on the logging infrastructure to implement explicit coverage marks:
+//!     <https://ferrous-systems.com/blog/coverage-marks/>
+//!
+//! In production code, you can mark certain log lines as "this should be covered by a test".
+//! In test code, you can then assert that a _specific_ test covers a specific log line. The two
+//! benefits are:
+//! - tests are more resilient to refactors
+//! - production code is more readable (you can immediately jump to a specific test)
+test "tutorial" {
```
matklad, line 11: Was inspired by https://github.com/ziglang/zig/pull/18955/commits/00acf8a66dec78743d26d62e86391b8794a1c627 to just put a "tutorial" style test right after the doc comment, instead of doing inline examples.

src/testing/covered.zig
```diff
@@ -0,0 +1,124 @@
...
+    name: []const u8,
+
+    pub fn expect_hit(m: Mark) !void {
+        comptime assert(builtin.is_test);
+        assert(global_state.mark_name.?.ptr == m.name.ptr);
+
+        defer {
+            global_state.mark_name = null;
+            global_state.mark_hit_count = 0;
+        }
```
sentientwaffle: Or `defer global_state = .{};`?

sentientwaffle (dismissed):
Excellent!

I considered whether this should be attached to logging -- e.g. we could have a separate + explicit `coverage.add(enum);` instead of piggybacking on log lines. But let's just keep things simple!

sentientwaffle (approved):

matklad:
>I considered whether this should be attached to logging -- e.g. we could have a separate + explicit coverage.add(enum); instead of piggybacking on log lines

One principle we have is that we optimize for "production" code --- we want the logic to read as its going to be executed, we don't want `if BUGGIFY` or comments with machine-checkable proofs or anything else that distracts from what's actually going on.

And, from that perspective, fully explicit coverage marks are definietly noisy. 

But I realized that `log.covered.debug` strikes the perfect balance here:

* it is still explicit, so you could then grep for this log line in tests
* and yet it sufficiently unobtrusive, we don't really add anything to production code.

jorangreef:
Love this.

> But I realized that log.covered.debug strikes the perfect balance here:

I would only suggest that we flip it around. `log.debug` should always be covered, and then opt-out if you don't want coverage. This cleans up the control flow a little more.

PR #1609, opened 2024-02-27
Test/Network: transition_to_liveness_mode + link_drop_packet_fn

sentientwaffle:
`transition_to_liveness_mode()` + `link_drop_packet_fn()` are never uses simultaneously.

If they were, then we would need to `null`-ify `link_drop_packet_fn` at the same time as the `link_filter`.


matklad (approved):
:heart:

PR #1618, opened 2024-02-28
lsm: snapshot_min is inclusive

cb22:
Other places (and the code) is correct, it was just this one comment.

### DIFF
--- a/src/lsm/manifest.zig
+++ b/src/lsm/manifest.zig
@@ -32,7 +32,7 @@ pub fn TreeTableInfoType(comptime Table: type) type {
         /// Address of the table's index block.
         address: u64,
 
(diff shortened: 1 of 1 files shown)


cb22:
There are actually quite a few things already out of date with the other comments there and in `lsm.md`. I'm going to leave them for the compaction refactor rather, but this is fine to go as-is.

batiati (approved):
The comment was indeed incorrect, but it looks like the real meaning of "exclusive" in that context has to do with the range between tables (e.g  `snapshot_min = compaction.snapshot + 1`.)

:+1: 

https://github.com/tigerbeetle/tigerbeetle/blob/5ae6bcdfd6c9e7cef399dc105ce2a0066e7235d1/src/lsm/manifest.zig#L58-L62

PR #1620, opened 2024-02-28
tests: terminology change covered -> marks

matklad:
After chatting today with Joran I realized that "coverage" is the wrong word here. This is a very poor implementation of coverage indeed, as it requires you to manually annotate the code, instead of relying on the computer to do the job.

But coverage isn't the point. What we actually do here is traceability --- linking two artifacts, source code and tests, together, to make sure that they motivate each over.

Coverage answers the question "Is this tested?". Marks answer "Why this code needs to exist?"

### DIFF
--- a/src/testing/marks.zig
+++ b/src/testing/marks.zig
@@ -1,5 +1,3 @@
-//! We'll get you covered!
-//!
 //! This file piggy-backs on the logging infrastructure to implement explicit coverage marks:
 //!     <https://ferrous-systems.com/blog/coverage-marks/>
 //!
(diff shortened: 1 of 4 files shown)


src/testing/marks.zig
```diff
@@ -8,27 +6,35 @@
 //! benefits are:
 //! - tests are more resilient to refactors
 //! - production code is more readable (you can immediately jump to a specific test)
+//!
+//! At the surface level, this resembles usual code coverage, but the idea is closer to traceability
+//! from safety-critical systems:
+//!     <https://en.wikipedia.org/wiki/Requirements_traceability>
```
matklad, line 12: Hat tip to @jamesmunns for teaching me about traceability back at Ferrous!

sentientwaffle (approved):

src/testing/marks.zig
jamesmunns: 🫡

PR #1621, opened 2024-02-28
docs: complex journal entries recipe

emschwartz:


docs/recipes/complex-journal-entries.md
```diff
@@ -0,0 +1,61 @@
+# Complex (Multi-Debit, Multi-Credit) Journal Entries
+
+TigerBeetle is designed for maximum performance. In order to keep it lean, the database only supports simple transfers with a single debit and a single credit.
+
+However, you'll probably run into cases where you want transactions with multiple debits and/or credits. If so, this page is for you!
+
+## 1-to-N
```
sentientwaffle: Maybe "1-to-many" or "one-to-many", since the other section is titled "many-to-many".

docs/recipes/complex-journal-entries.md
```diff
@@ -0,0 +1,61 @@
...
+
+## Many-to-Many Transactions
+
+Transactions with multiple debits and multiple credits are a bit more tricky.
+
+To simplify the math, we can use a temporary "control account" as an intermediary.
+
+In this example, we'll use the following accounts:
+
+- Three _source accounts_ `A` and `B` on the `USD` ledger.
```
sentientwaffle:
```suggestion
- Two _source accounts_ `A` and `B` on the `USD` ledger.
```

docs/recipes/complex-journal-entries.md
```diff
@@ -0,0 +1,61 @@
+# Complex (Multi-Debit, Multi-Credit) Journal Entries
```
sentientwaffle: What about "Multi-Debit/Credit Journal Entries"? Or s/journal entries/transfers? (I'm just trying to shorten the title, since it is extracted into the sidebar navigation on https://docs.tigerbeetle.com/!)

docs/recipes/complex-journal-entries.md
```diff
@@ -0,0 +1,61 @@
...
+
+Transactions with multiple debits and multiple credits are a bit more tricky.
+
+To simplify the math, we can use a temporary "control account" as an intermediary.
+
+In this example, we'll use the following accounts:
+
+- Three _source accounts_ `A` and `B` on the `USD` ledger.
+- Three _destination accounts_ `X`, `Y`, and `Z`, on the `USD` ledger.
+- A _compound entry control account_ `Control` on the `USD` ledger.
```
sentientwaffle: Maybe mention that `Control` can have `flags.debits_must_not_exceed_credits` set? Though it isn't necessary, so maybe that is an unnecessary distraction.

emschwartz: Happy to include it or leave it out. It seems like if this is something we're going to really recommend to people, it might be worth having some account flag like `net_balance_must_be_zero` or _maybe_ allowing `debits_must_not_exceed_credits` and `credits_must_not_exceed_debits` to be set at the same time (although that would technically allow situations where both the debits and credits are increased at the same time, which we wouldn't want in this case).

sentientwaffle (dismissed):

sentientwaffle (approved):
LGTM.

For future reference, if you need to update the branch against `main`, prefer rebase over merge to keep the git history tidy. (It is usually only necessary if there is a conflict.)

PR #1623, opened 2024-02-29
vsr: apply marks to existing tests

matklad:


sentientwaffle (approved):
Excellent! And the tests still covered the parts of the code they were supposed to be covering! :grin:

PR #1626, opened 2024-02-29
vopr: correct false assertion in liveness checker

matklad:
It might be the case that:

* cluster is stuck due to a missing prepare
* and yet there's a single replica in recovering head

Seed: 17563231628283278428
Context: https://github.com/tigerbeetle/tigerbeetle/pull/933

### DIFF
--- a/src/simulator.zig
+++ b/src/simulator.zig
@@ -521,8 +521,7 @@ pub const Simulator = struct {
         if (core_recovering_head == 0) return false;
 
         const quorums = vsr.quorums(simulator.options.cluster.replica_count);
-        assert(quorums.view_change > core_replicas - core_recovering_head);
-        return true;
+        return quorums.view_change > core_replicas - core_recovering_head;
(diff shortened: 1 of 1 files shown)


sentientwaffle (approved):

src/simulator.zig
```diff
@@ -521,8 +521,7 @@ pub const Simulator = struct {
         if (core_recovering_head == 0) return false;
 
         const quorums = vsr.quorums(simulator.options.cluster.replica_count);
-        assert(quorums.view_change > core_replicas - core_recovering_head);
-        return true;
+        return quorums.view_change > core_replicas - core_recovering_head;
```
sentientwaffle, line 524: Explaining this to myself: In `Simulator.tick_crash` we do try to make sure that not too many replicas end up in `recovering_head`. But if the core excludes too many healthy replicas, then that `recovering_head` can't recover!

PR #1633, opened 2024-02-29
docs: small edits to multi-debit/credit recipe

emschwartz:
based on feedback from @jorangreef

### DIFF
--- a/docs/recipes/multi-debit-credit-transfers.md
+++ b/docs/recipes/multi-debit-credit-transfers.md
@@ -44,7 +44,7 @@ This example debits multiple accounts and credits a single account. It uses the 
 
 Transactions with multiple debits and multiple credits are a bit more involved (but you got this!).
 
-This is where the accounting concept of a Control Account comes in handy. We can use this as a temporary intermediary account, as illustrated below.
+This is where the accounting concept of a Control Account comes in handy. We can use this as an intermediary account, as illustrated below.
 
 In this example, we'll use the following accounts:
 
--- a/docs/recipes/multi-debit-credit-transfers.md
+++ b/docs/recipes/multi-debit-credit-transfers.md
@@ -62,7 +62,7 @@ In this example, we'll use the following accounts:
 
 Here, we use two transfers to debit accounts `A` and `B` and credit the `Control` account, and another three transfers to credit accounts `X`, `Y`, and `Z`.
 
-If you look closely at this example, you may notice that we could technically have simply debited `B` and credited `Z` directly because the amounts happen to line up. That is true!
+If you looked closely at this example, you may have noticed that we could have debited `B` and credited `Z` directly because the amounts happened to line up. That is true!
 
 For performance reasons, you _might_ want to implement logic to circumvent the `Control` account and reduce the number of transfer to implement a certain compound journal entry.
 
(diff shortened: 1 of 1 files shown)


docs/recipes/multi-debit-credit-transfers.md
```diff
@@ -42,9 +42,9 @@ This example debits multiple accounts and credits a single account. It uses the
 
 ## Many-to-Many Transfers
 
-Transactions with multiple debits and multiple credits are a bit more tricky.
+Transactions with multiple debits and multiple credits are a bit more involved (but you got this!).
 
-To simplify the math, we can use a temporary "control account" as an intermediary.
+This is where the accounting concept of a Control Account comes in handy. We can use this as a temporary intermediary account, as illustrated below.
```
jorangreef:
```suggestion
This is where the accounting concept of a Control Account comes in handy. We can use this as an intermediary account, as illustrated below.
```

To remove "temporary", since the control account remains. It's more that the financial state transition goes in/out or transitions through immediately.

docs/recipes/multi-debit-credit-transfers.md
```diff
@@ -62,4 +62,8 @@ In this example, we'll use the following accounts:
 
 Here, we use two transfers to debit accounts `A` and `B` and credit the `Control` account, and another three transfers to credit accounts `X`, `Y`, and `Z`.
 
-If you look closely at this example, you may notice that we could technically have simply debited `B` and credited `Z` directly because the amounts happen to line up. That is true! However, since you probably don't want to program a lot of heuristics to figure out which debits and credits exactly match up, you'll probably find it easier to set up these complex journal entries using this type of control account.
+If you look closely at this example, you may notice that we could technically have simply debited `B` and credited `Z` directly because the amounts happen to line up. That is true!
```
jorangreef:
```suggestion
If you looked closely at this example, you may have noticed that we could have debited `B` and credited `Z` directly because the amounts happened to line up. That is true!
```

Putting it in the past tense ("If you looked closely") can be a way to surprise people, to get them to pay more attention next time, as if it's a game (like we might quiz them afterwards) compared to asking them to pay more attention now.

jorangreef: Removed "technically" and "simply" to make the language itself simpler. Less adverbs, adjectives. Just the raw verbs and nouns. We could still drop "directly" except that illustrates the point.

docs/recipes/multi-debit-credit-transfers.md
```diff
@@ -62,4 +62,8 @@ In this example, we'll use the following accounts:
...
+
+For performance reasons, you _might_ want to implement logic to circumvent the `Control` account and reduce the number of transfer to implement a certain compound journal entry.
+
+However, you also might want to try to avoid premature optimizations (we've all been there!). You'll probably find it easier to program these compound journal entries using this type of control account -- and you can always come back to squeeze this extra bit of performance out later!
```
jorangreef:
```suggestion
For a little more extreme performance, you _might_ consider implementing logic to circumvent the control account where possible, to reduce the number of transfers to implement a compound journal entry.

However, if you're just getting started, you can avoid premature optimizations (we've all been there!). You may find it easier to program these compound journal entries _always_ using a control account -- and you can then come back to squeeze this performance out later!
```

jorangreef (dismissed):

jorangreef (approved):

PR #1637, opened 2024-03-01
docs: clarify language in accounts, transfers sections

emschwartz:
- **docs: clarify how to store association between linked transfers/accounts**
- **docs: link to linked_event_chain_open error**
- **docs: mention that accounts and transfer may have the same ID**
- **docs: clarify language about using same data structures internally/externally**
- **docs: mention why timeout is an interval in seconds**

### DIFF
--- a/docs/reference/accounts.md
+++ b/docs/reference/accounts.md
@@ -166,7 +166,8 @@ account in the batch, to create a chain of accounts, of arbitrary
 length, which all succeed or fail in creation together. The tail of a
 chain is denoted by the first account without this flag. The last
 account in a batch may therefore never have `flags.linked` set as
-this would leave a chain open-ended (see `linked_event_chain_open`).
+this would leave a chain open-ended (see
+[`linked_event_chain_open`](./operations/create_accounts.md#linked_event_chain_open)).
 
 Multiple chains or individual accounts may coexist within a batch to
 succeed or fail independently. Accounts within a chain are executed
--- a/docs/reference/accounts.md
+++ b/docs/reference/accounts.md
@@ -27,7 +27,11 @@ Constraints:
 
 * Type is 128-bit unsigned integer (16 bytes)
 * Must not be zero or `2^128 - 1` (the highest 128-bit unsigned integer)
-* Must not conflict with another account
+* Must not conflict with another account (note that an account MAY have the same `id` as a transfer,
+but we would recommend avoiding this)
+
+See the [`id` section in the data modeling doc](../design/data-modeling.md#id) for more
+recommendations on choosing an ID scheme.
 
 ### `debits_pending`
 
--- a/docs/reference/accounts.md
+++ b/docs/reference/accounts.md
@@ -7,9 +7,11 @@ sidebar_position: 1
 An `Account` is a record storing the cumulative effect of committed
 [transfers](./transfers.md).
 
-TigerBeetle uses the same data structures internally and
-externally. This means that sometimes you need to set temporary values
-for fields that TigerBeetle, not you (the user), are responsible.
+TigerBeetle uses the same data structures internally and externally. This means
+that when creating accounts, you will need to set temporary values for certain
+fields that TigerBeetle is responsible for. For example, you will set the
+[`timestamp`](#timestamp) field to `0` when creating an account and then
+TigerBeetle will set the field to the timestamp when the operation was executed.
 
 ### Updates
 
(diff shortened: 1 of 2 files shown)


sentientwaffle (approved):

sentientwaffle (approved):

docs/reference/accounts.md
```diff
@@ -7,9 +7,11 @@ sidebar_position: 1
...
 [transfers](./transfers.md).
 
-TigerBeetle uses the same data structures internally and
-externally. This means that sometimes you need to set temporary values
-for fields that TigerBeetle, not you (the user), are responsible.
+TigerBeetle uses the same data structures internally and externally. This means
+that when creating accounts, you will need to set temporary values for certain
+fields that TigerBeetle is responsible for. For example, you will set the
+[`timestamp`](#timestamp) field to `0` when creating an account and then
+TigerBeetle will set the field to the timestamp when the operation was executed.
```
jorangreef, line 14: Rather than mention "temporary values" (because there's no valid range of values except 0)—let's say that these fields are set by TigerBeetle, and therefore reserved on the way in.

jorangreef: We should motivate (or provide a link to such motivation) why TigerBeetle (and not the application) is responsible for timestamping.

docs/reference/accounts.md
```diff
@@ -27,7 +29,11 @@ Constraints:
 
 * Type is 128-bit unsigned integer (16 bytes)
 * Must not be zero or `2^128 - 1` (the highest 128-bit unsigned integer)
-* Must not conflict with another account
+* Must not conflict with another account (note that an account MAY have the same `id` as a transfer,
+but we would recommend avoiding this)
```
jorangreef, line 33: We should always explain why. i.e. Why we recommend avoiding ID collisions across namespaces. For example, other systems may not allow this, therefore being stricter in what one sends to TigerBeetle improves interoperability.

At this point, the question will arise, well... why does TigerBeetle allow it then? And the answer is that if one is migrating from SQL to TigerBeetle, then I think there may be Account and Transfer tables with monotonically increasing integer IDs that collide. Of course, the user could fix this up by hashing these to derive UUIDs that are unique across namespaces.

docs/reference/accounts.md
```diff
@@ -178,9 +185,13 @@ unique error result. Other accounts in the chain will have their error
...
-After the link has executed, the association of each event is lost.
-To save the association, it must be
-[encoded into the data model](../design/data-modeling.md).
+
+After the chain of account operations has executed, the fact that they were
+linked will not be saved by default.
+To save the association between accounts, it must be
+[encoded into the data model](../design/data-modeling.md), for example by
+adding an ID to one of the [user data](../design/data-modeling.md#user_data)
+fields.
```
jorangreef, line 194: Good to say this positively like this, because "association of each event" being lost was not really what linked transfers were intended for in the first place.

jorangreef: To clarify a little more (and forestall any question of, well how could we save the link as a non-default)... it's not that the link is not "saved by default" but rather that it's "not saved" (at all).

docs/reference/transfers.md
```diff
@@ -39,9 +39,11 @@ Fields used by each mode of transfer:
...
 | `timestamp`                   | none         | none     | none         | none         |
 
-TigerBeetle uses the same data structures internally and
-externally. This means that sometimes you need to set temporary values
-for fields that TigerBeetle, not you (the user), are responsible.
+TigerBeetle uses the same data structures internally and externally. This means 
+that when creating transfers, you will need to set temporary values for certain 
+fields that TigerBeetle is responsible for. For example, you will set the 
+[`timestamp`](#timestamp) field to `0` when creating a transfer and then
+TigerBeetle will set the field to the timestamp when the transfer was executed.
```
jorangreef, line 46: Can we DRY this up in one place?

jorangreef: Same for other sections below?

emschwartz: Does it even make sense to include this note? This feels like something that could just be documented in the client libraries, or the client libraries could hide this fact from users

docs/reference/accounts.md
emschwartz: I'm going to remove this note and just link to the `id` section on the data modeling page, because that already includes a mention of this

docs/reference/accounts.md
emschwartz: Good idea. I wrote up this page to do that https://github.com/tigerbeetle/tigerbeetle/pull/1655

PR #1640, opened 2024-03-01
Client: move `request_queue` outside `vsr.Client`

kprotty:
First step to making tb_client packets intrusive is having each packet map 1:1 to a `vsr.Client.Request`. This was awkward to do as `vsr.Client` is specific to `StateMachine` and holds a callback but `Packet` is type-erased with `tb_client_t` holding the callback..

An observation was made that most uses of `vsr.Client` are sequential and don't actually need a queue of concurrent requests (aof, benchmark, demo, repl). The only uses which queue are `tb_client` requiring batching/demuxing and `Cluster/Simulator` requiring specific queue semantics. So lets make `vsr.Client` only handle one message at a time, delegating queueing to its caller. Apart from simplifying internals, this means Packet (which `tb_client` already queues up outside VSR) now implicitly maps to a VSR request to be intrusive later.

Batching/demuxing was moved directly into `tb_client`, necessitating `Packet` to be a bit thicker. I bumped its size from 32b to 64b with an extra `reserved: [2]usize` to keep the cache-line idea from TIGER_STYLE. The thread-safe submission stack was also moved into `Context` so that requeuing could push to an unsynchronized FIFO instead and not call `signal.notify()` spuriously. This improved batched request throughput in language client tests.


src/testing/cluster.zig
```diff
@@ -61,6 +62,10 @@ pub fn ClusterType(comptime StateMachineType: anytype) type {
         pub const StateChecker = StateCheckerType(Client, Replica);
         pub const ManifestChecker = ManifestCheckerType(StateMachine.Forest);
 
+        const ClientQueue = RingBuffer(*Message.Request, .{
+            .array = constants.client_request_queue_max,
+        });
```
sentientwaffle: Is this just so that `Cluster.request()`'s API stays the same? It might make more sense to modify the state machine workload so that at most one request per client is ever in flight.

src/vsr/client.zig
```diff
@@ -170,12 +101,11 @@ pub fn Client(comptime StateMachine_: type, comptime MessageBus: type) type {
...
 
-        /// Pool of nodes which represent batched Infos on a given Request.
-        demux_pool: DemuxPool,
+        /// Tracks the currently processing register message if any.
+        register_inflight: ?*Message.Request = null,
 
-        /// A client is allowed at most one inflight request at a time at the protocol layer.
-        /// We therefore queue any further concurrent requests made by the application layer.
-        request_queue: RequestQueue = RequestQueue.init(),
+        /// Tracks a currently processing request message submitted by `raw_request`.
```
sentientwaffle:
```suggestion
        /// Tracks a currently processing (non-register) request message submitted by `raw_request`.
```

src/vsr/client.zig
```diff
@@ -30,98 +29,30 @@ pub fn Client(comptime StateMachine_: type, comptime MessageBus: type) type {
...
-                    } else {
-                        assert(batch.request.demux_queue.contains(batch.demux));
-                    }
-                }
-
-                const message = batch.request.message;
-                assert(message.header.command == .request);
+        /// Custom Demuxer which passes through to StateMachine.Demuxer.
+        pub fn DemuxerType(comptime operation: StateMachine.Operation) type {
+            return struct {
```
sentientwaffle: Should this definition move nearer to where it is used?

src/vsr/client.zig
```diff
@@ -296,208 +221,81 @@ pub fn Client(comptime StateMachine_: type, comptime MessageBus: type) type {
...
-            const demux = self.demux_pool.acquire() orelse unreachable;
-            errdefer self.demux_pool.release(demux);
-
-            demux.* = .{
-                .user_data = user_data,
-                .callback = callback,
-                .event_count = @intCast(@divExact(message.body().len, event_size)),
-                .event_offset = 0,
-            };
+            assert(!message.header.operation.vsr_reserved());
```
sentientwaffle: Does this assertion need to move to `request()`? `raw_request` replays prepares during AOF recovery, which can include reserved operations. (I realize that this is not a new assertion).

src/vsr/client.zig
```diff
@@ -587,42 +385,48 @@ pub fn Client(comptime StateMachine_: type, comptime MessageBus: type) type {
...
-                    });
-                    return;
-                }
-                assert(reply.header.request_checksum == inflight.message.header.checksum);
-            } else {
+            var inflight = if (self.register_inflight) |message|
+                Request{ .message = message, .user_data = 0, .callback = undefined }
+            else if (self.request_inflight) |inflight|
+                inflight
+            else {
```
sentientwaffle, line 388: I think we can add
```suggestion
            else {
                assert(reply.header.request < self.request_number);
```

src/vsr/client.zig
```diff
@@ -587,42 +385,48 @@ pub fn Client(comptime StateMachine_: type, comptime MessageBus: type) type {
...
                 log.debug("{}: on_reply: ignoring (no inflight request)", .{self.id});
                 return;
+            };
+
+            if (reply.header.request < inflight.message.header.request) {
```
sentientwaffle, line 394:
```suggestion
            if (reply.header.request < inflight.message.header.request) {
                assert(self.register_inflight == null);
```

src/vsr/client.zig
```diff
@@ -587,42 +385,48 @@ pub fn Client(comptime StateMachine_: type, comptime MessageBus: type) type {
...
+                    self.id,
+                    reply.header.request,
+                    inflight.message.header.request,
+                });
+                return;
             }
 
-            var inflight = self.request_queue.pop().?;
-            const inflight_request = inflight.message.header.request;
+            assert(reply.header.request_checksum == inflight.message.header.checksum);
```
sentientwaffle, line 405:
```suggestion
            assert(reply.header.request == inflight.message.header.request);
            assert(reply.header.request_checksum == inflight.message.header.checksum);
```

src/vsr/client.zig
```diff
@@ -587,42 +385,48 @@ pub fn Client(comptime StateMachine_: type, comptime MessageBus: type) type {
...
-            // Eagerly release request message, to ensure that user's callback can submit a new
-            // request.
+            // Eagerly release request message, to ensure that user's callback can submit a new one.
             self.release_message(inflight.message.base());
             assert(self.messages_available > 0);
-
-            // Even though we release our reference to the message, we might have another one
-            // retained by the send queue in case of timeout.
-            maybe(inflight.message.references > 0);
             inflight.message = undefined;
```
sentientwaffle: Could be be a little less eager? e.g. move this directly before the `if (inflight_vsr_operation == .register) {` branch that is responsible for invoking the callbacks. That minimize the duration that the message is released but not nulled.

src/vsr/client.zig
```diff
@@ -783,14 +549,9 @@ pub fn Client(comptime StateMachine_: type, comptime MessageBus: type) type {
...
-            assert(self.request_queue.empty());
-            self.request_queue.push_assume_capacity(.{
-                .message = message,
-            });
-
-            const request = self.request_queue.head_ptr().?;
-            assert(self.request_queue.tail_ptr() == request);
-            request.demux_queue.push(demux);
+            assert(self.request_inflight == null);
+            assert(self.register_inflight == null);
```
sentientwaffle: Can these move up to the top of the function (after `if (self.request_number > 0) return;`) to serve as preconditions?

src/vsr/client.zig
```diff
@@ -835,7 +596,13 @@ pub fn Client(comptime StateMachine_: type, comptime MessageBus: type) type {
         }
 
         fn send_request_for_the_first_time(self: *Self, message: *Message.Request) void {
-            assert(self.request_queue.head_ptr().?.message == message);
+            if (self.register_inflight) |inflight| {
+                maybe(self.request_inflight != null);
+                assert(inflight == message);
```
sentientwaffle: Swap the order of these asserts so that they mirror the `else` branch's asserts.

src/testing/cluster/state_checker.zig
```diff
@@ -178,17 +178,17 @@ pub fn StateCheckerType(comptime Client: type, comptime Replica: type) type {
...
-                assert(request.message.header.command == .request);
-                assert(request.message.header.operation == header_b.?.operation);
-                assert(request.message.header.size == header_b.?.size);
+                const message = client.register_inflight orelse client.request_inflight.?.message;
+                assert(message.header.client == header_b.?.client);
+                assert(message.header.checksum == header_b.?.request_checksum);
+                assert(message.header.request == header_b.?.request);
+                assert(message.header.command == .request);
+                assert(message.header.operation == header_b.?.operation);
+                assert(message.header.size == header_b.?.size);
```
sentientwaffle: I think `request` is a better name for this, to distinguish it from the prepare(s).

src/simulator.zig
```diff
@@ -817,13 +818,13 @@ pub const Simulator = struct {
             // since they may start one at any time.
             reserved += @intFromBool(c.session == 0);
             // Count the number of requests queued.
```
```suggestion
            // Count the number of non-register requests queued.
```

src/simulator.zig
```diff
@@ -817,13 +818,13 @@ pub const Simulator = struct {
...
-            reserved += c.request_queue.count;
+            reserved += @intFromBool(c.request_inflight != null);
         }
         // +1 for the potential request — is there room in the sequencer's queue?
         if (reserved + 1 > simulator.reply_sequence.free()) return;
 
-        // Make sure that there is capacity in the client's request queue.
-        if (client.messages_available == 0) return;
+        // Make sure that the client is free to send a new request
```
```suggestion
        // Make sure that the client is ready to send a new request
```
nit

src/fifo.zig
```diff
@@ -14,7 +14,7 @@ pub fn FIFO(comptime T: type) type {
         out: ?*T = null,
         count: u64 = 0,
         // This should only be null if you're sure we'll never want to monitor `count`.
-        name: ?[]const u8,
+        name: ?[*:0]const u8,
```
sentientwaffle: Is this change necessary for type reasons?

src/clients/c/tb_client/context.zig
```diff
@@ -243,52 +267,152 @@ pub fn ContextType(
...
+
+            // Otherwise, try to batch the packet with another already in self.pending.
+            if (StateMachine.batch_logical_allowed.get(@enumFromInt(packet.operation))) {
+                var it = self.pending.peek();
+                while (it) |root| {
+                    it = root.next;
+
+                    // Check for pending packets of the same operation which can be batched.
+                    if (root.operation != packet.operation) continue;
+                    if (@intFromPtr(root.batch_data) + packet.data_size > constants.message_body_size_max) continue;
```
sentientwaffle: I'm confused... Why use `batch_data` (which is a pointer) to store the length?

src/clients/c/tb_client/context.zig
```diff
@@ -243,52 +267,152 @@ pub fn ContextType(
...
-            stdx.copy_disjoint(.exact, u8, batch.slice(), readable);
+                const event_data = @as([*]const u8, @ptrCast(batched.data.?))[0..batched.data_size];
+                stdx.copy_disjoint(.inexact, u8, message.body()[offset..], event_data);
+                offset += @intCast(event_data.len);
+            }
 
-            // Submit the message for processing:
-            self.client.batch_submit(
+            assert(offset == batch_size);
+            self.client.raw_request(
```
sentientwaffle, line 353: Hmm, so we are stuck using `raw_request` b/c we need to be the ones to assemble the batch. That is unfortunate (iirc `raw_request` was originally just for AOF recovery) but I don't see a nice way around it. I guess we could change `Client.request()`'s API to take a `message` too, but (unlike `raw_request`) it could ignore the header.

src/clients/c/tb_client/context.zig
```diff
@@ -243,52 +267,152 @@ pub fn ContextType(
...
+            op: StateMachine.Operation,
+            reply: []const u8,
         ) void {
             const user_data: UserData = @bitCast(raw_user_data);
             const self = user_data.self;
             const packet = user_data.packet;
 
-            assert(packet.operation == @intFromEnum(op));
-            self.on_complete(packet, results);
+            // Submit the next pending packet after having VSR completed this one.
```
sentientwaffle:
```suggestion
            // Submit the next pending packet now that VSR has completed this one.
```

src/clients/c/tb_client/context.zig
```diff
@@ -243,52 +267,152 @@ pub fn ContextType(
...
+            if (self.pending.pop()) |packet_next| {
+                self.submit(packet_next);
+            }
+
+            switch (op) {
+                inline else => |operation| {
+                    // Demuxer expects []u8 but VSR callback provides []const u8.
+                    // The bytes are known to come from a Message body that will be soon discarded
+                    // therefor it's safe to @constCast and potentially modify the data in-place.
+                    var demuxer = Client.DemuxerType(operation).init(@constCast(reply));
```
sentientwaffle, line 388: Where do we modify the data? Or is it just the hypothetical -- i.e., that we could modify it if we wanted to.

src/clients/node/src/node.zig
```diff
@@ -234,14 +234,12 @@ fn request(
...
-        .status = .ok,
-        .data_size = @intCast(packet_data.len),
-        .data = packet_data.ptr,
-    };
+    packet.operation = @intFromEnum(operation);
+    packet.user_data = callback_ref;
+    packet.data = packet_data.ptr;
+    packet.data_size = @intCast(packet_data.len);
+    packet.next = null;
+    packet.status = .ok;
```
sentientwaffle: I prefer assigning with `packet.* = ...` even if some fields need to be repeated, since it makes it obvious that all of the fields have a value assigned (or explicitly set to `undefined`, as in the case of `reserved`).

src/vsr/client.zig
sentientwaffle, line 1: Since there is no more client request queue, we can reduce the number of messages that the client allocates. In `message_pool.zig`, replace:

```
    sum += constants.client_request_queue_max; // Client.request_queue
```

with:
```
    sum += 1; // Client.register_inflight
    sum += 1; // Client.request_inflight
```

Relatedly, are `Client.messages_available` and `constants.client_request_queue_max` still needed for anything, or can they be removed?

sentientwaffle (commented):
Mostly just minor questions or suggestions... this is an excellent simplification!

src/clients/c/tb_client/context.zig
kprotty: It's mostly to save space in packet, but will add the extra field to make it simpler.

src/clients/c/tb_client/context.zig
kprotty: It's modified in the main [StateMachine's Demuxer](<https://github.com/tigerbeetle/tigerbeetle/blob/main/src/state_machine.zig#L154-L159>) to update the `CreateTError.index`s to be packet-relative after having batched them. Should vsr.Client.Request.Callback just switch to `[]u8` instead?

src/fifo.zig
kprotty: Tried to put a FIFO in Packet for batching but `[]const u8` wasn't compatible with `extern struct`. Removed the FIFO as not enough space so this is indeed a rogue change.

src/vsr/client.zig
kprotty: It's used outside of `vsr.Client` (in the unit test below and tb_client Context).

src/vsr/client.zig
kprotty: `request()` already knows the operation is a StateMachine / not vsr reserved, so I think this assert can just be deleted for AOF. 

But in `Client.on_reply` it either expects a `.register` or a `StateMachine.Operation` to pass to the request callback. Would need to know how AOF intends to produce/handle this.

src/vsr/client.zig
kprotty: `messages_available` can be easily removed, but would need to double check on removing `client_request_queue_max` correctly. Will do so in its own PR.

src/vsr/client.zig
sentientwaffle: So why not move this (and the corresponding unit test) move to tb_client Context's file? It seems weird to have it here since `vsr.Client` is now unaware of batching.

kprotty: It's due to `EchoClient`: its Demuxer must return Events back but StateMachine.Demuxer adjusts & returns Results. Creating a stub StateMachine just for EchoClient, with most of the public decls from the original but a custom Demuxer, seemed more weird. So instead, `vsr.Client` and `EchoClient` each have their own Demuxer wrapper and corresponding unit test.

src/vsr/client.zig
```diff
@@ -296,210 +210,79 @@ pub fn Client(comptime StateMachine_: type, comptime MessageBus: type) type {
...
-                    @divExact(message.body().len, event_size),
-                ),
-                .event_offset = 0,
-            };
+            assert(message.header.client == self.id);
+            assert(message.header.release.value == self.release.value);
+            assert(message.header.cluster == self.cluster);
+            assert(message.header.command == .request);
+            assert(message.header.size >= @sizeOf(Header));
+            assert(message.header.size <= constants.message_size_max);
```
sentientwaffle, line 268:
```suggestion
            assert(message.header.size <= constants.message_size_max);
            assert(message.header.operation.valid(StateMachine));
```

src/clients/c/tb_client/context.zig
```diff
@@ -243,52 +267,144 @@ pub fn ContextType(
...
+
+            const operation: StateMachine.Operation = @enumFromInt(packet.operation);
+            message.header.* = .{
+                .release = self.client.release,
+                .client = self.client.id,
+                .request = undefined, // Set by client.raw_request.
+                .cluster = self.client.cluster,
+                .command = .request,
+                .operation = vsr.Operation.from(StateMachine, operation),
+                .size = @intCast(@sizeOf(vsr.Header) + packet.batch_size),
```
sentientwaffle: With `batch_size` a `u32`, is this `@intCast` needed?

src/clients/c/tb_client/context.zig
sentientwaffle: Aha, thanks! Yeah, if we are treating it as mutable then we probably should change the type. And ideally we would be able to assert `message.references == 1`, though I'm not positive if that would hold.

src/vsr/client.zig
```diff
@@ -296,210 +210,80 @@ pub fn Client(comptime StateMachine_: type, comptime MessageBus: type) type {
...
+            assert(message.header.operation.valid(StateMachine));
```
sentientwaffle, line 269: Can we assert that `message.header.timestamp == 0 or constants.aof_recovery);`?

sentientwaffle, line 269: Also assert that `view`/`parent`/`session`/`request` are all zero.

src/vsr/client.zig
```diff
@@ -589,42 +372,46 @@ pub fn Client(comptime StateMachine_: type, comptime MessageBus: type) type {
...
-                        inflight.message.header.request,
-                    });
-                    return;
-                }
-                assert(reply.header.request_checksum == inflight.message.header.checksum);
-            } else {
+            var inflight = if (self.register_inflight) |message|
+                Request{ .message = message, .user_data = 0, .callback = undefined }
+            else if (self.request_inflight) |inflight|
+                inflight
```
batiati, line 387: Is it worth documenting/asserting that `register_inflight`, when != null, is the equivalent of the old request_queue's head?

src/vsr/client.zig
kprotty: This [was reached](<https://github.com/tigerbeetle/tigerbeetle/actions/runs/8364883560/job/22901324354?pr=1640>) under the existing `test_aof` step so I'll comment it out for now.

src/clients/c/tb_client/echo_client.zig
```diff
@@ -17,30 +18,43 @@ pub fn EchoClient(comptime StateMachine_: type, comptime MessageBus: type) type
...
-            break :blk struct {
-                pub const StateMachine = Client.StateMachine;
-                pub const Error = Client.Error;
-                pub const Request = Client.Request;
-                pub const BatchError = Client.BatchError;
+        const VSRClient = @import("../../../vsr/client.zig").Client(StateMachine_, MessageBus);
+        pub const StateMachine = VSRClient.StateMachine;
+        pub const Request = VSRClient.Request;
+
+        /// Custom Demuxer which treats Event(operation)'s and results and echoes them back.
```
sentientwaffle: Does this mean to say s/"and results"/"as results"?

sentientwaffle (approved):

PR #1647, opened 2024-03-04
docs: quickstart tweaks

emschwartz:
- **docs: tidy up quickstart links**
- **docs: point users to data modeling after repl quickstart**
- **docs: link to internals doc about data file**

### DIFF
--- a/docs/quick-start/cli-repl.md
+++ b/docs/quick-start/cli-repl.md
@@ -1,8 +1,8 @@
 ---
-sidebar_position: 4
+sidebar_position: 5
 ---
 
-# Creating accounts and transfers with the CLI repl
+# Use the CLI REPL to Create Accounts and Transfers
 
 Once you've got some TigerBeetle replicas running, let's connect to the
 replicas and do some accounting!
--- a/docs/quick-start/cli-repl.md
+++ b/docs/quick-start/cli-repl.md
@@ -73,3 +73,9 @@ lookup_accounts id=1, id=2;
 And indeed you can see that account `1` has `debits_posted` as `10`
 and account `2` has `credits_posted` as `10`. The `10` amount is fully
 accounted for!
+
+You can take a look at the [`Accounts` reference](../reference/accounts.md) to understand all of the fields on the accounts.
+
+## Next: Designing for TigerBeetle
+
+Now that you've created some accounts and transfers, you may want to dig into the [data model](../design/data-modeling.md) to learn more about how to build with TigerBeetle.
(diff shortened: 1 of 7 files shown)


docs/quick-start/cli-repl.md
```diff
@@ -1,8 +1,8 @@
 ---
-sidebar_position: 4
```
matklad, line 2: excellent catch!

docs/quick-start/cli-repl.md
```diff
@@ -73,3 +73,9 @@ lookup_accounts id=1, id=2;
 And indeed you can see that account `1` has `debits_posted` as `10`
 and account `2` has `credits_posted` as `10`. The `10` amount is fully
 accounted for!
+
+You can take a look at the [`Accounts` reference](../reference/accounts.md) to understand all of the fields on the accounts.
+
+## Next: Designing for TigerBeetle
+
+Now that you've created some accounts and transfers, you may want to dig into the [data model](../design/data-modeling.md) to learn more about how to build with TigerBeetle.
```
matklad: Please hard-wrap lines at 100 characters.

We also want our docs to be readable as a raw markdown text, and there hard-wrapping helps. I don't know which editor you use, but I think more or less anything can nowdays wrap markdown properly.

docs/quick-start/single-binary-three.md
```diff
@@ -22,7 +22,7 @@ Want to build from source locally? Add `-build` as an argument to the bootstrap
 
 ## Running TigerBeetle
 
-Now create the TigerBeetle data file for each replica
+Now create the TigerBeetle [data file](../internals/data_file.md) for each replica:
```
matklad, line 25: I am not sure whether we should be linking internal docs from public docs --- they aren't really written for normal users to understand. 

I think it's fine to do that, _but_ we need to make sure that any internal doc that we link in this way starts with the disclaimer that "you don't need to read this to use TB". 

I guess the first-best solution would be to create some kind of glossary entry for the "data file" in the public docs, and, from that entry, to link to internals. 

No need to do anything for this PR though, I think linking is better that not linking at all!

docs/quick-start/with-docker.md
```diff
@@ -15,7 +15,7 @@ info(io): creating "0_0.tigerbeetle"...
 info(io): allocating 660.140625MiB...
 ```
 
-Then run the server.
+Then run the server:
```
matklad, line 19: :heart: love consistency here!

matklad (commented):
Please wrap lines, otherwise looks excellent to me!

docs/quick-start/single-binary-three.md
emschwartz: Instead of linking like this, we could have an extra sentence that says "(If you are curious, you can learn more about the contents of the data file in the internals doc)"

matklad:
Seems to still have some broken links left

matklad (approved):

PR #1648, opened 2024-03-04
Test/Marks: Log err vs warn

sentientwaffle:
https://github.com/tigerbeetle/tigerbeetle/pull/1607 changed `replica.zig`'s logger from

    const log = stdx.log.scoped(.replica);

to

    const log = covered.wrap_log(std.log.scoped(.replica));

`stdx.log`'s job was to downgrade `log.err()` to `log.warn()` during tests, since the former will trigger a test failure. (See `test_runner.zig` `log_err_count`). `marks.wrap_log()` handled this properly for `wrap_log().err`, but not `wrap_log().mark.err()`.

...But it (unexpectedly) works anyway (in Zig 0.11). That is, the test (`"Cluster: view-change: DVC, 2/3 faulty header stall"`) hit the `log.err()` (`"quorum received, deadlocked"`), and the test runner prints the messages as errors, but then it doesn't exit with code 1. I don't understand this!

Zig `0.12.0-dev.575+fb6fff256` failed as I would expect, but I don't see any differences in the test runner code that would account for this.

### DIFF
--- a/src/testing/marks.zig
+++ b/src/testing/marks.zig
@@ -77,10 +77,7 @@ pub fn check(name: []const u8) Mark {
 pub fn wrap_log(comptime base: anytype) type {
     if (builtin.is_test) {
         return struct {
-            pub const err = warn;
-            pub const warn = base.warn;
-            pub const info = base.info;
-            pub const debug = base.debug;
+            pub usingnamespace base;
 
             pub const mark = struct {
                 pub fn err(comptime fmt: []const u8, args: anytype) void {
@@ -106,11 +103,7 @@ pub fn wrap_log(comptime base: anytype) type {
         };
     } else {
         return struct {
-            pub const err = warn;
-            pub const warn = base.warn;
-            pub const info = base.info;
-            pub const debug = base.debug;
-
+            pub usingnamespace base;
             pub const mark = base;
         };
     }
--- a/src/vsr/replica.zig
+++ b/src/vsr/replica.zig
@@ -30,7 +30,7 @@ const SyncStage = vsr.SyncStage;
 const SyncTarget = vsr.SyncTarget;
 const ClientSessions = vsr.ClientSessions;
 
-const log = marks.wrap_log(std.log.scoped(.replica));
+const log = marks.wrap_log(stdx.log.scoped(.replica));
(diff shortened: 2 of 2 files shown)


src/testing/marks.zig
```diff
@@ -106,7 +108,7 @@ pub fn wrap_log(comptime base: anytype) type {
         };
     } else {
         return struct {
-            pub const err = warn;
+            pub const err = base.err;
```
sentientwaffle: Was downgrading `err` to `warn` outside of `builtin.is_test` deliberate?

src/testing/marks.zig
```diff
@@ -85,7 +87,7 @@ pub fn wrap_log(comptime base: anytype) type {
             pub const mark = struct {
                 pub fn err(comptime fmt: []const u8, args: anytype) void {
                     record(fmt);
-                    base.err(fmt, args);
+                    base.warn(fmt, args);
```
sentientwaffle: An alternate fix would be to pass `stdx.log.scoped(...)` in as the base logger to `wrap_log()`.

matklad:
Oh, dear, I indeed put _two_ bugs here which cancelled each other in testing :facepalm: 

In replica.zig, we should have

```
const log = marks.wrap_log(stdx.log.scoped(.replica));
```

with `stdx`

And then in marks.zig we of course need just

```
pub usingnamespace base;
```

rather than manually overriding each method.

matklad:
>But it (unexpectedly) works anyway (in Zig 0.11). That is, the test ("Cluster: view-change: DVC, 2/3 faulty header stall") hit the log.err() ("quorum received, deadlocked"), and the test runner prints the messages as errors, but then it doesn't exit with code 1. I don't understand this!

Ok, this was puzzling! This was a regression in Zig: in 0.11 (or was it 0.10?) they added two test modes:

* terminal mode, where the test binary just runs all tests sequentially
* server mode, where the test binary spans a server which accepts commands like "list tests" and "run a single test", where the server is driven by the code in build.zig

And then newly introduced server mode wrongly didn't consider log_err_count.

This was since fixed in https://github.com/ziglang/zig/pull/16958

matklad (approved):

PR #1649, opened 2024-03-05
vsr: never ignore SV

matklad:
I am mostly curious what breaks if we do this. We _already_ allow our view headers to be in the next wrap (through SV -> DVC conversion), so perhaps we can just relax this?


matklad:
Ok, year, this doesn't work: 

- even if `.recovering_head` replica doesn't update `.commit_max` immediately, it'll update it on the next `commit` message.

PR #1650, opened 2024-03-05
vsr: recovering head replicas don't get stuck in an old view

matklad:
There's a catch 22 between joining a view via an SV message and jumping several log wraps forward when state sync.

To state sync, we require that our durable view is not less than the view of the message that informed us about the checkpoint. If we _don't_ do that, we might observe a chain break between the log and the checkpoint.

To act on SV, we need to install at least one header from it (to ensure that our local prepare hashchain to the true cluster state at the given view). But to install the header it must be in-bounds for the current checkpoint.

In most of the cases this _doesn't_ lead to a deadlock because, even if a replica doesn't append anything from SV to its log, it at least switches to view_change status for that view, updating view durable.

The exception is `.recovering_head` state, where we bump our in-memory view number, but don't touch view durable: to update view durable we need to update our view headers, but that's not safe to do in general in recovering_head state.

This creates a deadlock, where .recovering_head replica can't state sync because it's view durable is too low, but also can't update its view durable, because it's too far behind the cluster.

But, if the SV _is_ far in the future, it's actually is possible to get out of `.recovering_head` without installing any headers:

* If the SV is for the same log_view we are currently in, we know we don't need to truncate anything from our log
* If the SV is for a view larger than our log_view, we can transition to view_change. Our DVC might incorrectly nack some prepares, but that is irrelevant, as the cluster is on the entire different checkpoint already.

SEED: 13825956399249152350

### DIFF
--- a/src/vsr/replica.zig
+++ b/src/vsr/replica.zig
@@ -1972,6 +1972,19 @@ pub fn ReplicaType(
                     });
                     self.sync_start_from_committing();
                 }
+
+                if (self.status == .recovering_head and self.op >= self.commit_min) {
+                    // All the ops that could have been in this replica's journal are already
+                    // committed by the cluster, it is safe to transition away from recovering_head.
+                    if (message.header.view == self.log_view) {
+                        self.transition_to_normal_from_recovering_head_status(message.header.view);
+                        self.commit_journal();
+                        self.repair();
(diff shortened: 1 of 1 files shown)


matklad:
I am going with a different approach here

PR #1653, opened 2024-03-05
Test/StateChecker: Assert that committed headers match

sentientwaffle:
When a replica commits, but it is not the first replica to commit that op, verify that they committed the same header.

(Before I thought that we were already checking this via `assert((commit_a == commit_b) == (checksum_a == checksum_b));`. But that is checking a different property -- that a replica does not disagree with its past self.)


sentientwaffle:
Nevermind; I think we do check this properly after all.

sentientwaffle:
(We check it on next tick (when our commits match), but that is probably enough.)

PR #1655, opened 2024-03-05
docs: add page about Time

emschwartz:
I basically tried to capture the content from @jorangreef's [talk](https://www.youtube.com/watch?v=C3oWbZ4tQPM). Please feel free to add any other details or bits of explanation you think would be helpful.

### DIFF
--- a/docs/design/time.md
+++ b/docs/design/time.md
@@ -1,0 +1,77 @@
+---
+sidebar_position: 6
+---
+
+# Time
+
+Time is a critical component of all distributed systems and databases. Within TigerBeetle, we keep
+track of two types of time: logical time and physical time.
+
+## Logical Time
+
+TigerBeetle uses [Viewstamped replication](https://pmg.csail.mit.edu/papers/vr-revisited.pdf) for
+its consensus protocol. This ensures [strict
+serializability](http://www.bailis.org/blog/linearizability-versus-serializability/) for all
+operations.
+
+## Physical Time
+
+TigerBeetle uses physical time in addition to the logical time provided by the consensus algorithm.
+Financial transactions require physical time for multiple reasons, including:
+
+- **Liquidity** - TigerBeetle supports [Two-Phase Transfers](./two-phase-transfers.md) that reserve funds
+  and hold them in a pending state until they are posted, voided, or the transfer times out. A
+  timeout is useful to ensure that the reserved funds are not held in limbo indefinitely.
+- **Compliance and Auditing** - For regulatory and security purposes, it is useful to have a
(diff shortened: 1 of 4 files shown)


docs/design/time.md
```diff
@@ -0,0 +1,77 @@
...
+
+Primary nodes monitor their clock skew with respect to the other replicas and may abdicate their
+role as primary if they appear to be far off the rest of the cluster.
+
+This is why the `timestamp` field must be set to `0` when operations are submitted, and it is then
+set by the Primary.
+
+Similarly, the [`Transfer.timeout`](../reference/transfers.md#timeout) is given as an interval in
+seconds, rather than as an absolute timestamp, because it is also managed by the Primary. The
+`timeout` is calculated relative to the `timestamp` when the operation arrives at the Primary.
```
sentientwaffle: s/Primary/primary

(I don't feel strongly about which is better, but we don't uppercase it elsewhere in the docs.)

docs/design/time.md
```diff
@@ -0,0 +1,77 @@
...
+
+### Timestamps are Totally Ordered
+
+All `timestamp`s within TigerBeetle are unique, immutable and [totally
+ordered](http://book.mixu.net/distsys/time.html). A transfer that is created before another transfer
+is guaranteed to have an earlier `timestamp` (even if they were created by the same Operation).
```
sentientwaffle: s/Operation/request or batch

"same operation" implies both were created by `create_transfers` (which is also true, but is less meaningful).

docs/design/time.md
```diff
@@ -0,0 +1,77 @@
...
+TigerBeetle clusters are distributed systems, so we cannot rely on all replicas in a cluster keeping
+their system clocks perfectly in sync. Because of this, we "trust but verify".
+
+In the TigerBeetle consensus protocol messages, replicas exchange timestamp information and monitor
+the skew of their clock with respect to the other replicas. The system ensures that the clock skew
+is within an acceptable window and may choose to stop processing transactions if the clocks are far
+out of sync.
+
+Importantly, the goal is not to reimplement clock synchronization protocols, but to verify that the
+cluster is operating within acceptable error bounds.
```
sentientwaffle:
```suggestion
Importantly, the goal is not to reimplement or replace clock synchronization protocols, but to verify that the
cluster is operating within acceptable error bounds.
```

to clarify that operators should still run NTP on replica servers, not just rely on TB to figure it out. (We could spell that out explicitly too, but that belongs more in a deployment guide than here.)

sentientwaffle (approved):

PR #1656, opened 2024-03-05
VSR/Eviction: Add Header.Eviction.reason

sentientwaffle:
Allow the cluster to signal to the client why their request is being rejected.

Right now there is only `Reason.no_session`, but as part of upgrades the cluster will use eviction messages to inform clients that they are running an incompatible version.

### DIFF
--- a/src/vsr/message_header.zig
+++ b/src/vsr/message_header.zig
@@ -1107,7 +1107,8 @@ pub const Header = extern struct {
         reserved_frame: [16]u8 = [_]u8{0} ** 16,
 
         client: u128,
-        reserved: [112]u8 = [_]u8{0} ** 112,
+        reserved: [111]u8 = [_]u8{0} ** 111,
+        reason: Reason,
 
         fn invalid_header(self: *const @This()) ?[]const u8 {
             assert(self.command == .eviction);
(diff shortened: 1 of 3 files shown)


src/vsr/message_header.zig
```diff
@@ -1107,16 +1107,36 @@ pub const Header = extern struct {
...
             if (self.checksum_body != checksum_body_empty) return "checksum_body != expected";
             if (self.client == 0) return "client == 0";
             if (!stdx.zeroed(&self.reserved)) return "reserved != 0";
+
+            const reasons = comptime std.enums.values(Reason);
+            inline for (reasons) |reason| {
+                if (@intFromEnum(self.reason) == @intFromEnum(reason)) {
+                    return "reason invalid";
+                }
+            }
```
matklad: It feels like a negation is missing here somewhere? This code returns `"reason invalid"` if the reason is either `0` or `1`, but I think the intention is to check the case where it's _neither_ 0 _nor_ 1?

Could you also double check that this works correctly in release safe mode and isn't optimized to true? If it does, then perhaps we can replace the 

> // The `Command` enum is exhaustive

comment in this file with a similar intFromEnum loop, just to make it a bit more explicit?

matklad: > comment in this file with a similar intFromEnum loop, just to make it a bit more explicit?

To clarify, we _still_ crash on unknown commands, but we have an explict `unreachable` in the source code for that, rather then relying on safety-check for switch statement.

sentientwaffle: Ah good catch, I got the condition backwards. :facepalm:

matklad (approved):

src/vsr/message_header.zig
sentientwaffle: (Also, I verified with a quick test that a bad `Reason` is detected now.):

```zig
    test "FIXME" {
        const v: u8 = 2;
        var x = Header.Eviction{
            .command = .eviction,
            .cluster = 0,
            .replica = 0,
            .view = 0,
            .client = 1,
            .reason = std.mem.bytesAsValue(Eviction.Reason, std.mem.asBytes(&v)).*,
        };
        x.set_checksum_body(&[_]u8{});
        x.set_checksum();
        if (x.invalid_header()) |e| {
            std.debug.print("INVALID: {s}\n", .{e});
        } else {
            @panic("fail");
        }
    }
```

PR #1657, opened 2024-03-06
Expiration of pending transfers

batiati:
## Background

  After bumping the state machine `prepare_timestamp` every 100ms triggered by a `vsr.Timeout`, the primary calls `pulse,` which checks if it intersects with the expiration timestamp of any pending transfer. If the verification returns positive the primary sends a `request` to itself with `command == .pulse`.
 
  The entire process is agnostic of the state machine, the replica just asks the state machine if it needs a `pulse`, which can be any logic implemented by the state machine, in this case, the  `expire_pending_transfers`.
  
  At the state machine level, the only thing "special" is that `pulse` operations cannot accept input or produce replies to clients. They are not intended to be called by clients, but there's no special limitation around this except by not exposing functions for calling them (in fact the AOF logic calls it).
 
  The state machine does the verification in memory, storing a `u64` referring to the next timestamp that a transfer will expire as a flag to check if an `expire_pending_transfers` is needed.
  * If we don't have enough information (i.e. first time we run), we must execute the pulse.
  * Every time a transfer is inserted/posted/voided, the state machine can update the in-memory expiration timestamp.
  * During the expiration, we need to scan for the next pending transfer and update the flag. If there are more transfers to expire than a single batch, the flag is adjusted to resume in the last processed one instead.

The next `pulse` can only run when the current one is committed and the in-memory flag is updated.

## Design decisions

- The derived index `Transfer.expires_at` is inserted for each pending transfer with `timeout > 0`.
  When a pending transfer is posted/voided/expired, this index is **removed**, so we can simply scan by an open interval (e.g. `WHERE expires_at <= prepare_timestamp`) to retrieve all **expired** transfers that need to be processed.
    
- The derived index `Transfer.pending_status` is inserted for all pending transfers and is updated during the transfer lifecycle from `pending` to `posted`|`voided`|`expired`.
  
  This index enables queries such as "give me all pending transfers", and "give me all posted/voided/expired transfers".
  
  Interesting point to think about, we can maybe get rid of the `PostedGroove`, although they have different index directions. 

- The expired transfer changes the account balance, but unlike posted/voided transfers, there is no corresponding transfer that reverts the balance.
  This would break the rule that every entry in `get_account_balances` must have a corresponding transfer with the same timestamp. As a result, **we don't register the account history** when expiring transfers!

## Tests

- Vopr and unit tests covering the expiration logic.


src/vsr/replica.zig
```diff
@@ -8466,7 +8587,11 @@ pub fn ReplicaType(
                 @panic("checkpoint diverged");
             }
 
-            if (candidate.view > self.view_durable()) {
+            // TODO: This block introduced by the PR #1598 causes a liveness issue
+            // when a replica cannot start the sync.
+            // SEED 10266230273529621812, 8842160722467235252
+            const fix_me = true;
+            if (fix_me and candidate.view > self.view_durable()) {
```
batiati: I still need to investigate this, unsure if it is caused by the `pulse` operations or if they just happened to expose the existing issue.

src/vsr/replica.zig
```diff
@@ -5319,6 +5390,46 @@ pub fn ReplicaType(
...
+            assert(self.primary());
+
+            const message = self.message_bus.get_message(.prepare);
+            defer self.message_bus.unref(message);
+
+            const prepare_timestamp = self.state_machine.prepare_timestamp;
+            const latest_entry = self.journal.header_with_op(self.op).?;
+            message.header.* = Header.Prepare{
+                .cluster = self.cluster,
+                .size = @intCast(@sizeOf(Header)),
```
sentientwaffle: `@intCast` shouldn't be necessary here.

src/simulator.zig
```diff
@@ -658,44 +659,58 @@ pub const Simulator = struct {
...
-        reply: *Message.Reply,
+        reply: ClusterReply,
     ) void {
         // TODO(Zig) Use @returnAddress to initialzie the cluster, then this can just use @fieldParentPtr().
         const simulator: *Simulator = @ptrCast(@alignCast(cluster.context.?));
-        simulator.reply_sequence.insert(reply_client, request, reply);
+        const inserted = simulator.reply_sequence.insert(reply);
+        if (!inserted) return;
 
         while (simulator.reply_sequence.peek()) |commit| {
```
sentientwaffle: Ah, we independently solved this problem in different ways (https://github.com/tigerbeetle/tigerbeetle/pull/1682/commits/09b304bab09797e0ba8f555323a19174c7805505). :facepalm: 

Any preference between the two approaches? I think mine might be a little less intrusive, but not sure yet.

src/vsr/replica.zig
```diff
@@ -9284,16 +9411,27 @@ const PipelineQueue = struct {
...
             assert(pipeline.request_queue.empty() or
-                pipeline.prepare_queue.count + 1 == constants.pipeline_prepare_queue_max);
+                pipeline.prepare_queue.count + 1 == constants.pipeline_prepare_queue_max or
+                pipeline.pulse_in_progress());
             return false;
         }
     }
 
+    fn pulse_in_progress(pipeline: *PipelineQueue) bool {
+        return inline for (0..constants.pipeline_prepare_queue_max) |index| {
```
sentientwaffle: Why inline-for?

src/vsr/replica.zig
```diff
@@ -9284,16 +9411,27 @@ const PipelineQueue = struct {
...
         } else {
             assert(pipeline.request_queue.empty() or
-                pipeline.prepare_queue.count + 1 == constants.pipeline_prepare_queue_max);
+                pipeline.prepare_queue.count + 1 == constants.pipeline_prepare_queue_max or
+                pipeline.pulse_in_progress());
             return false;
         }
     }
 
+    fn pulse_in_progress(pipeline: *PipelineQueue) bool {
```
sentientwaffle: You might want to take a look at `PipelineQueue.message_by_operation` (in https://github.com/tigerbeetle/tigerbeetle/pull/1682) which solves this a little more generally.

src/vsr/replica.zig
```diff
@@ -9284,16 +9411,27 @@ const PipelineQueue = struct {
         }
     }
 
-    fn full(pipeline: PipelineQueue) bool {
+    fn full(pipeline: *PipelineQueue) bool {
```
```suggestion
    fn full(pipeline: *const PipelineQueue) bool {
```
?

src/vsr/replica.zig
```diff
@@ -9256,12 +9381,14 @@ const PipelineQueue = struct {
         while (pipeline.prepare_queue.pop()) |p| message_pool.unref(p.message);
     }
 
-    fn verify(pipeline: PipelineQueue) void {
+    fn verify(pipeline: *PipelineQueue) void {
```
```suggestion
    fn verify(pipeline: *const PipelineQueue) void {
```

src/vsr/replica.zig
```diff
@@ -5319,6 +5390,46 @@ pub fn ReplicaType(
...
+                .commit = self.commit_max,
+                .timestamp = prepare_timestamp,
+                .operation = vsr.Operation.from(StateMachine, operation),
+                .client = 0,
+                .request = 0,
+            };
+
+            assert(message.body().len == 0);
+            const sector_ceil = vsr.sector_ceil(message.header.size);
+            if (message.header.size != sector_ceil) {
```
sentientwaffle: Why make this a branch, since it is always true?

src/vsr/replica.zig
```diff
@@ -5319,6 +5390,46 @@ pub fn ReplicaType(
             assert(reconfiguration_request.result != .reserved);
         }
 
+        fn pulse_inject(self: *Self, operation: StateMachine.Operation) void {
+            assert(self.status == .normal);
+            assert(self.primary());
```
sentientwaffle: Assert that pipeline prepare queue is non-full.
Assert that pipeline queue does not contain a pulse.

src/vsr/replica.zig
```diff
@@ -1071,6 +1075,11 @@ pub fn ReplicaType(
                     .id = replica_index,
                     .after = 50,
                 },
+                .pulse_timeout = Timeout{
+                    .name = "pulse_timeout",
+                    .id = replica_index,
+                    .after = 1 * (std.time.ms_per_s / constants.tick_ms), // 1s
```
```suggestion
                    .after = 1 * (std.time.ms_per_s / constants.tick_ms), // 1 ticks/s
```
The timeout triggers once per second, but the `1s` confused me at first since `after`'s units are not seconds.

src/lsm/scan_lookup.zig
```diff
@@ -23,6 +23,14 @@ pub fn ScanLookupType(
 
         pub const Callback = *const fn (*ScanLookup) void;
 
+        pub const Status = enum {
+            idle,
+            scan,
+            lookup,
+            buffer_finished,
+            scan_finished,
```
sentientwaffle, line 31: Can you help me understand what motivated splitting these two states?

docs/reference/transfers.md
```diff
@@ -185,10 +185,11 @@ Constraints:
...
 [`pending`](#flagspending) transfer's [arrival at the cluster](#timestamp)
-that it may be posted or voided. Zero denotes absence of timeout.
+that it may be [posted](#flagspost_pending_transfer) or [voided](#flagsvoid_pending_transfer).
+Zero denotes absence of timeout.
 
 If the timeout expires and the pending transfer has not already been
-posted or voided, the pending transfer is voided automatically.
+posted or voided, the pending balance is removed automatically.
 
 Non-pending transfers cannot have a timeout.
```
sentientwaffle, line 192: Can you specify the answer to: If multiple transfers' expiry is at the identical nanosecond, in what order are they expired? (It is in chronological (creation) order iiuc.)

src/state_machine.zig
```diff
@@ -210,12 +218,29 @@ pub fn StateMachineType(
                     .user_data_128 = config.lsm_batch_multiple * constants.batch_max.create_transfers,
                     .user_data_64 = config.lsm_batch_multiple * constants.batch_max.create_transfers,
                     .user_data_32 = config.lsm_batch_multiple * constants.batch_max.create_transfers,
-                    .timeout = config.lsm_batch_multiple * constants.batch_max.create_transfers,
                     .ledger = config.lsm_batch_multiple * constants.batch_max.create_transfers,
                     .code = config.lsm_batch_multiple * constants.batch_max.create_transfers,
+                    .expires_at = config.lsm_batch_multiple * constants.batch_max.create_transfers,
+                    // Updated when the transfer is posted/voided/expired.
+                    .pending_status = 2 * config.lsm_batch_multiple *
+                        constants.batch_max.create_transfers,
```
sentientwaffle, line 227: :thinking: Where does the 2x factor come from?

src/state_machine.zig
```diff
@@ -891,6 +943,211 @@ pub fn StateMachineType(
             self.prefetch_finish();
         }
 
+        fn prefetch_expire_pending_transfers(self: *StateMachine) void {
+            assert(self.scan_result_count == 0);
+            assert(self.prefetch_timestamp >= TimestampRange.timestamp_min);
+            assert(self.prefetch_timestamp != TimestampRange.timestamp_max);
```
sentientwaffle, line 957: Can we assert `self.expire_pending_transfers_pulse_timestamp <= self.prepare_timestamp`?

docs/reference/transfers.md
sentientwaffle, line 1: In the account history reference page, document the caveat about how timeouts interact with balance history.

src/state_machine.zig
```diff
@@ -891,6 +943,211 @@ pub fn StateMachineType(
...
+                // If there is no transfer to expire, set the value as `timestamp_max`.
+                self.expire_pending_transfers_pulse_timestamp = TimestampRange.timestamp_max;
+            } else {
+                assert(results.len == 1);
+
+                const t = &results[0];
+                assert(t.flags.pending);
+                assert(t.timeout > 0);
+
+                const expires_at = t.timestamp + t.timeout_ns();
```
sentientwaffle, line 1014: What can we assert about this (with respect to `commit_timestamp` maybe)?

src/state_machine.zig
```diff
@@ -891,6 +943,211 @@ pub fn StateMachineType(
...
+            // We only need the first result.
+            assert(scan_buffer.len == 1);
+
+            const transfers_groove: *TransfersGroove = &self.forest.grooves.transfers;
+            const scan_builder: *TransfersGroove.ScanBuilder = &transfers_groove.scan_builder;
+
+            // WHERE `expires_at > prefetch_timestamp`
+            // Scanning `Transfers` that _will_ expire, so we can determine the _next_ timestamp
+            // it will require to execute `expire_pending_transfers` again.
+            // We must do it before scanning for expired transfers to reuse the same buffer.
```
sentientwaffle, line 972: I am still early in the review, but to help me orient can you explain why we need the `prefetch_expire_pending_transfers` step instead of just jumping right into `prefetch_expire_pending_transfers_scan`?

Naively, I would imagine that we could do something like:
- Fetch the next 8190 transfers to expire.
- If we encounter one of those that shouldn't expire yet, save that timestamp so that we don't pulse again until then.
- If all 8190 transfers need to expire, then we will pulse again.

src/state_machine.zig
```diff
@@ -1599,6 +1890,86 @@ pub fn StateMachineType(
...
+            assert(status != .none and status != .pending);
+
+            self.forest.grooves.transfers.indexes.pending_status.remove(&.{
+                .timestamp = timestamp,
+                .field = @intFromEnum(TransferPendingStatus.pending),
+            });
+            self.forest.grooves.transfers.indexes.pending_status.put(&.{
+                .timestamp = timestamp,
+                .field = @intFromEnum(status),
+            });
```
sentientwaffle, line 1924: Ah, so we remove-put instead of .update since we don't prefetch the pending_status?

src/state_machine.zig
```diff
@@ -2504,8 +2905,58 @@ test "create/lookup 2-phase transfers" {
...
+        \\ transfer   T4 A1 A2   13   _  _  _  _    3 L1 C1   _ PEN   _   _   _   _  _ _ ok
+        \\ commit create_transfers
+
+        // Check balances before expiration.
+        \\ lookup_account A1 46  0  0  0
+        \\ lookup_account A2  0  0 46  0
+        \\ commit lookup_accounts
+
+        // Check balances after 1s.
+        \\ tick 1000000000
```
sentientwaffle: Nit, but I'd suggest modifying `tick` to take a unit enum, so you can instead write `tick 1 seconds`

src/vsr/replica.zig
```diff
@@ -2905,6 +2916,32 @@ pub fn ReplicaType(
             }
         }
 
+        fn on_pulse_timeout(self: *Self) void {
+            if (constants.aof_recovery) unreachable;
```
sentientwaffle: Why not `comptime assert(!constants.aof_recovery);`?

sentientwaffle (commented):
Very rough review, haven't gotten very far yet but I want to post my questions

src/vsr/replica.zig
```diff
@@ -3820,7 +3858,14 @@ pub fn ReplicaType(
             assert(self.commit_min == prepare.header.op);
             self.advance_commit_max(self.commit_min, @src());
 
-            if (self.event_callback) |hook| hook(self, .committed);
+            if (self.event_callback) |hook| hook(self, .{ .committed = prepare });
+
+            if (prepare.header.client == 0) {
+                // This prepare was sent by the primary, tere's no reply to the client.
```
```suggestion
                // This prepare was sent by the primary, there's no reply to the client.
```

src/vsr/replica.zig
```diff
@@ -8466,7 +8587,8 @@ pub fn ReplicaType(
                 @panic("checkpoint diverged");
             }
 
-            if (candidate.view > self.view_durable()) {
+            // Ignoring newer view, we need to change view first when receiving a SV.
+            if (candidate.view > self.view_durable() and self.view < candidate.view) {
```
batiati: Fix for the seeds `10266230273529621812` and `8842160722467235252`.
The replica's `view_durable` lags behind, but its `view` is already in the current view, so it ignores `SV`.

src/vsr/replica.zig
```diff
@@ -8522,7 +8644,9 @@ pub fn ReplicaType(
                 .checkpoint_op = candidate.checkpoint_op,
             };
 
-            if (self.syncing != .idle) {
+            if (self.syncing != .idle and
+                self.grid_repair_tables.executing() == 0)
+            {
```
batiati: Fix for the seed `16070513861840054826`.
The replica is already repairing tables (it's only possible after applying the fix above).

src/vsr/replica.zig
batiati: I wanted it to be a runtime panic; otherwise, I would need to comptime disable the timer logic as well.

src/vsr/replica.zig
batiati: To avoid a runtime loop, since the queue size is comptime-known, the generated code looks like an `if-else-if` chain.

src/lsm/scan_lookup.zig
batiati: When a scan finishes, it was because there are no more results (`scan_finished`) or because the entire buffer was filled (`buffer_finished`).

In this case, we need to know if we still need to process the same expiry timestamp again in the next pulse (e.g., there are more than 8190 transfers to expire).

In general, the QueryAPI will need this to indicate the pagination status.

src/state_machine.zig
batiati: The prefetch work is split into:

1. Scans for a single transfer that will expire next but has not expired yet (e.g., `WHERE expires_at > prefetch_timestamp ORDER BY ASC LIMIT 1`). This will be used to determine when a `pulse` is needed again.

    We do this step first to reuse the same buffer in the next step.

2. Scans for all transfers that have expired (e.g. `WHERE expires_at <= prefetch_timestamp ORDER BY ASC`).

3. Lookup the Debit and Credit accounts for each transfer.

4. Lookup the Posted for each transfer if `constants.verify`, to assert if the transfer wasn't already posted.

> Naively, I would imagine that we could do something like

I haven't considered unbounded fetching the next 8190 transfer to expire ( `WHERE expires_at > 0 ORDER BY ASC`) because we would do unnecessary work, since we need to always "materialize" all 8190 transfers and then when iterating maybe just a handful of them would need to expire.

But it's a nice idea if the Scan accepts a function pointer so the caller can tell when to stop (the first which is `expires_at > prefetch_timestamp`).
So we could do in a single scan the equivalent of what we are currently doing in two separated ones:

```SQL
WHERE expires_at <= prefetch_timestamp ORDER BY ASC
UNION
WHERE expires_at > prefetch_timestamp ORDER BY ASC LIMIT 1
```

src/state_machine.zig
batiati: I wish 😅! Intuitively, I would say "yes", and indeed, I tried it.

But when running the Vopr, it quickly found many situations where the prefetch could run out of sync with the variable. I concluded that we couldn't rely on this in-memory state during `prefetch` or `execute,` that's why we have the `pulse_reset` function every time a new primary starts a view.

src/state_machine.zig
batiati: The Vopr found this one! The index is inserted and then updated when the pending transfer is fulfilled.

docs/reference/transfers.md
batiati: The scan order is by the `expires_at` index, sorted by `{ timestamp + timeout_ns, timestamp }`.

src/state_machine.zig
batiati: Not sure if I understood what you mean ... but `PendingStatus` is an index, not an ObjectTree.
To update an index, you remove and reinsert it (we do the same in `groove.zig`).

src/vsr/replica.zig
batiati: See https://github.com/tigerbeetle/tigerbeetle/pull/1657#discussion_r1523023985

src/simulator.zig
batiati: Yeah, certainly I want to get rid of the `stalled_op` as you did, but I still need to store the `.no_reply` prepares so the simulator can run and verify the expiration logic.

I'm not happy with the way I solved it using the hook fired on `commit`:
- Each replica commits the prepare, the simulator needs to dedup them and process the event only once.
- We always assume that `pulse`s are messages from `client == 0 and request == 0`.

https://github.com/tigerbeetle/tigerbeetle/blob/21a6ee2c02595a5eabdb5cdd269f359d9f924121/src/vsr/replica.zig#L3861-L3868

src/message_pool.zig
```diff
@@ -28,6 +28,7 @@ pub const messages_max_replica = messages_max: {
     sum += constants.pipeline_prepare_queue_max; // Replica.Pipeline{Queue|Cache}
     sum += constants.pipeline_request_queue_max; // Replica.Pipeline{Queue|Cache}
     sum += 1; // Replica.commit_prepare
+    sum += 1; // The `StateMachine.pulse()` message
```
sentientwaffle: Is this referring to the message created by `Replica.pulse_inject()`? If so, that should already be covered by `sum += constants.pipeline_prepare_queue_max;`.

docs/reference/transfers.md
sentientwaffle: That's was my guess! But we should document it anyhow. (I think of our "reference" docs as a spec for our API -- even though an expiry collision is a (very rare) edge case, it is worth a brief mention.)

src/state_machine.zig
sentientwaffle: Ah right, and we update via a remove (insert tombstone) and an insert.

src/state_machine.zig
sentientwaffle: > But it's a nice idea if the Scan accepts a function pointer so the caller can tell when to stop

Yeah, that sounds like it would be perfect! The two queries are adjacent, so the second one can (usually) benefit from the first's cached blocks, but having a single query would be simpler too.

docs/reference/transfers.md
batiati: It's also relevant to document that we can only expire 8190 transfers per `pulse`, so the user might still see pending balances for expired transfers if their request is pipelined together with a pulse in such condition.

sentientwaffle: > so the user might still see pending balances for expired transfers if their request is pipelined together with a pulse in such condition.

Ahh I hadn't figured that out yet, but that makes perfect sense! The pulse is injected into the _end_ of the pipeline, so by the time it commits, there may be `pipeline_prepare_queue_max - 1` prepares queued after it, so if there are a lot of expired transfers then the pulse might not hit all of them. Or it does hit all of them, but some more happen to expire while those trailing prepares are being processed, before the next pulse can expire them.

src/ring_buffer.zig
```diff
@@ -111,6 +111,17 @@ pub fn RingBuffer(
             }
         }
 
+        pub inline fn get_const_ptr(self: *const Self, index: usize) ?*const T {
```
sentientwaffle: s/`get_const_ptr`/`get_ptr_const`, to match `next_tail_ptr_const`

docs/reference/transfers.md
batiati: Also, because we can only determine if we need the next pulse after the current one has been prefetched, that's why we support only one pulse at a time in the pipeline.

We can improve this logic in the future, replacing `expire_pending_transfers_pulse_timestamp` with some kind of "histogram" mapping multiple timestamps with a resolution of seconds to determine if we need to inject multiple pulses in the same pipeline, solving this issue.

src/simulator.zig
sentientwaffle: What if we use the upgrade PR's [implementation](https://github.com/tigerbeetle/tigerbeetle/blob/0b4da64429c3e0f7b78c7d778b4488d05f5d6bac/src/simulator.zig), but modified:

```zig
        while (!simulator.reply_sequence.empty()) {
            const op = simulator.reply_op_next;
            const prepare_header = simulator.cluster.state_checker.commits.items[op].header;
            assert(prepare_header.op == op);

            if (simulator.reply_sequence.peek(op)) |commit| {
                defer simulator.reply_sequence.next();

                assert(prepare_header.client != 0);
                simulator.reply_op_next += 1;

                // ...

                if (!commit.request.header.operation.vsr_reserved()) {
                    simulator.requests_replied += 1;
                    simulator.workload.on_reply(
                        commit.client_index,
                        commit.reply.header.operation.cast(StateMachine),
                        commit.reply.header.timestamp,
                        commit.request.body(),
                        commit.reply.body(),
                    );
                }
            } else {
                if (prepare_header.client == 0) {
                    if (prepare_header.operation == .pulse) { // <---------------- new code --------
                        simulator.workload.on_pulse(
                            prepare.header.operation.cast(StateMachine),
                            prepare.header.timestamp,
                        );
                    }
                    assert(prepare_header.operation.vsr_reserved());
                    // We don't receive replies for requests that originated at the replicas.
                    simulator.reply_op_next += 1;
                } else {
                    break;
                }
            }
```

That would avoid at least the `ReplySequence` side of the changes, though it wouldn't help with the `on_client_reply`-to-`on_cluster_reply`.

For that, what if the `while (!simulator.reply_sequence.empty())` was split out into a `Simulator.flush_workload` function, called by both `on_client_reply` and maybe in `done` or `tick` to flush any "trailing" pulses (pulses which were not succeeded by a non-pulse operation).

src/state_machine/workload.zig
```diff
@@ -364,6 +349,22 @@ pub fn WorkloadType(comptime AccountingStateMachine: type) type {
...
+            self: *Self,
+            operation: AccountingStateMachine.Operation,
+            timestamp: u64,
+        ) void {
+            assert(timestamp != 0);
+
+            switch (operation) {
+                .expire_pending_transfers => self.auditor.expire_pending_transfers(timestamp),
+                else => unreachable,
             }
```
sentientwaffle: Nit, but this is more succinct:
```suggestion
            assert(timestamp != 0);
            assert(operation == .expire_pending_transfers);

            self.auditor.expire_pending_transfers(timestamp);
```

src/state_machine.zig
```diff
@@ -508,9 +542,24 @@ pub fn StateMachineType(
                 .lookup_transfers => 0,
                 .get_account_transfers => 0,
                 .get_account_history => 0,
+                .expire_pending_transfers => 0,
```
sentientwaffle: I think this can actually be
```suggestion
                .expire_pending_transfers => unreachable,
```
since `pulse_inject` bypasses `state_machine.prepare()`.

src/state_machine.zig
```diff
@@ -555,6 +604,9 @@ pub fn StateMachineType(
                 .get_account_history => {
                     self.prefetch_get_account_history(parse_filter_from_input(input));
                 },
+                .expire_pending_transfers => {
```
sentientwaffle: Assert that `input.len == 0`.

src/state_machine.zig
```diff
@@ -508,9 +542,24 @@ pub fn StateMachineType(
...
                 .get_account_history => 0,
+                .expire_pending_transfers => 0,
             };
         }
 
+        pub fn pulse_reset(self: *StateMachine) void {
+            self.expire_pending_transfers_pulse_timestamp = TimestampRange.timestamp_min;
+        }
+
+        pub fn pulse_operation(self: *StateMachine) ?Operation {
```
```suggestion
        pub fn pulse_operation(self: *const StateMachine) ?Operation {
```

src/vsr/replica.zig
```diff
@@ -5319,6 +5390,46 @@ pub fn ReplicaType(
...
+            assert(message.body().len == 0);
+            const sector_ceil = vsr.sector_ceil(message.header.size);
+            assert(message.header.size < sector_ceil);
+            @memset(message.buffer[message.header.size..sector_ceil], 0);
+
+            message.header.set_checksum_body(message.body());
+            message.header.set_checksum();
+
+            self.push_prepare(message);
+        }
```
sentientwaffle: This duplicated logic makes me nervous; I'm wondering how we can share more with the standard prepare code path (`Replica.primary_pipeline_prepare()`).

What if instead of generating a `prepare`, we generated a `request` (which is simpler), then pass that request to `primary_pipeline_prepare`? That would also remove the `self.state_machine.prepare_timestamp` update from `on_pulse_timeout()`, and remove the need for the `checkpoint_id_now` helper.

src/vsr/replica.zig
```diff
@@ -457,6 +457,10 @@ pub fn ReplicaType(
         /// (syncing≠idle)
         sync_message_timeout: Timeout,
 
+        /// The number of ticks on an idle cluester before injecting a `pulse` operation.
+        /// (status=normal and primary)
```
```suggestion
        /// (status=normal and primary and !constants.aof_recovery)
```

src/state_machine.zig
batiati: We can't bc the aof recovery calls it.

src/vsr/replica.zig
batiati: The main reason was to avoid recursion calling `primary_pipeline_prepare`, and to avoid skip calling `pulse_operation` for those requests. I'll refactor that.

docs/reference/transfers.md
```diff
@@ -185,10 +185,20 @@ Constraints:
...
+posted or voided, the pending balance is removed automatically.
+
+Transfers expire in chronological order by their expiration time ([`timestamp`](#timestamp)
+_plus_ `timeout` converted in nanoseconds). If multiple transfers expire at the same time, then
+the creation [`timestamp`](#timestamp) is used.
+
+TigerBeetle can atomically expire [one batch](../design/client-requests.md#batching-events) of
+transfers at a time. It means that client requests executed immediately after the expiration _might_
+still see pending balances for expired transfers if there where more transfers to expire than it
+can process in a single batch.
```
sentientwaffle:
```suggestion
transfers at a time. This means that client requests executed immediately after the expiration _might_
still see pending balances for expired transfers if there were more transfers to expire than it
can process in a single batch.
```

src/simulator.zig
```diff
@@ -666,36 +667,62 @@ pub const Simulator = struct {
...
+                assert(prepare_header.client != 0);
+                simulator.reply_op_next += 1;
+
+                const commit_client = simulator.cluster.clients[commit.client_index];
+                assert(commit.reply.references == 1);
+                assert(commit.reply.header.op == op);
+                assert(commit.reply.header.command == .reply);
+                assert(commit.reply.header.client == commit_client.id);
+                assert(commit.reply.header.request == commit.request.header.request);
+                assert(commit.reply.header.operation == commit.request.header.operation);
```
sentientwaffle, line 747: Let's assert that `operation != .pulse`, to emphasize that we would never skip over a pulse due to the `if (!commit.request.header.operation.vsr_reserved()) {` branch below.

src/testing/cluster.zig
sentientwaffle, line 1: I think you can reset these changes, since the new `ReplySequence` means they aren't needed anymore.

src/testing/cluster/state_checker.zig
sentientwaffle, line 1: Fyi I think when you rebase, this will conflict with my recent upgrade PR (sorry!) but you should be able to just use the version from `main` and it will work with pulses.

src/vsr/replica_test.zig
```diff
@@ -15,6 +15,7 @@ const parse_table = @import("../testing/table.zig").parse;
 const marks = @import("../testing/marks.zig");
 const StateMachineType = @import("../testing/state_machine.zig").StateMachineType;
 const Cluster = @import("../testing/cluster.zig").ClusterType(StateMachineType);
+const ClusterReply = @import("../testing/cluster.zig").ClusterReply;
```
```suggestion
```

src/vsr/replica.zig
```diff
@@ -2905,6 +2916,32 @@ pub fn ReplicaType(
...
+            if (self.pipeline.queue.prepare_queue.full()) {
+                self.pulse_timeout.backoff(self.prng.random());
+                return;
+            }
+
+            const realtime = self.clock.realtime();
+            self.state_machine.prepare_timestamp = @max(
+                self.state_machine.prepare_timestamp,
+                @as(u64, @intCast(realtime)),
+            );
```
sentientwaffle, line 3047: Ah, I'd hoped we could remove this, but `state_machine.pulse` needs an up-to-date `prepare_timestamp`. I think it's worth a code comment explaining this, to guard against future mistakes.

src/vsr/replica.zig
```diff
@@ -2905,6 +2916,32 @@ pub fn ReplicaType(
...
 
+        fn on_pulse_timeout(self: *Self) void {
+            if (constants.aof_recovery) unreachable;
+
+            assert(self.status == .normal);
+            assert(self.primary());
+            assert(self.pulse_timeout.ticking);
+
+            if (self.pipeline.queue.prepare_queue.full()) {
+                self.pulse_timeout.backoff(self.prng.random());
```
sentientwaffle: I think, since we only send (at most) 1 pulse/second, and checking whether we are _able_ to send a pulse is inexpensive (`self.pipeline.queue.prepare_queue.full()`), it would be simpler to just move `self.pulse_timeout.reset();` up to execute it unconditionally, rather than using random backoff here.

src/vsr/replica.zig
```diff
@@ -5276,6 +5351,10 @@ pub fn ReplicaType(
             message.header.set_checksum_body(message.body());
             message.header.set_checksum();
 
+            self.push_prepare(message);
+        }
+
+        fn push_prepare(self: *Self, message: *Message.Prepare) void {
```
sentientwaffle: `push_prepare` can be inlined again.

src/vsr/replica.zig
```diff
@@ -5222,19 +5279,35 @@ pub fn ReplicaType(
...
+                            // correctly assure that time-dependant tasks (e.g. expiring transfers)
+                            // will never intersect with a request batch.
+                            if (self.state_machine.pulse()) {
+                                if (!self.pipeline.queue.contains_operation(.pulse)) {
+                                    self.send_request_pulse_to_self();
+                                    self.pipeline.queue.delay_request(request);
+                                    return;
+                                }
+                            }
+                        }
```
sentientwaffle: :thinking: I'm confused about a few things here:

Should 

```
                    self.state_machine.prepare(
                        request.message.header.operation.cast(StateMachine),
                        request.message.body(),
                    );
```

be _below_ this block, so that we don't call `state_machine.prepare` something if we aren't actually about to prepare it?

Next: `delay_request()` invokes `pipeline.request_queue.push_head_assume_capacity()`, so the delayed request is "next in line" of the requests. That makes sense, but then we use `send_request_pulse_to_self`, which goes to `on_message`, which will queue the pulse at the _back_ of the queue -- behind the request it is supposed to precede. (Or am I missing something here?)

Lastly: The pulse's `on_request()` may not even create the pulse at all -- incoming requests can be dropped by
```
            const realtime = self.clock.realtime_synchronized() orelse {
                log.err("{}: on_request: dropping (clock not synchronized)", .{self.replica});
                return;
            };
```
in `on_request()`. Though I suppose if that happens, we will just hit this same code path again next time around, so maybe that is fine. Then again, this feels like some hidden recursion, since `on_request` calls `primary_pipeline_prepare` if the clock is synchronized.

sentientwaffle: Assuming I'm not misunderstanding the situation, some possible solutions:

Add `primary_pipeline_prepare_or_pulse`, which does something like:

```
self.state_machine.prepare_timestamp = ...

if (!request.operation.vsr_reserved() and !constants.aof_recovery) {
    // refactor send_request_to_self into this, so that send_upgrade_to_self uses make_... as well:
    const pulse_request = self.make_request_to_self(...);
    defer self.message_pool.unref(pulse_request);
    self.pipeline.queue.delay_request(request);
    // faking realtime to avoid the clock-not-synchronized branch:
    self.primary_pipeline_prepare(.{ .message = pulse_request, .realtime = request.realtime });
} else {
    self.primary_pipeline_prepare(request)
}
```

src/state_machine.zig
```diff
@@ -1599,6 +1898,86 @@ pub fn StateMachineType(
...
+            const transfers: []const Transfer = std.mem.bytesAsSlice(
+                Transfer,
+                self.scan_buffer[0 .. self.scan_result_count * @sizeOf(Transfer)],
+            );
+
+            log.debug("expire_pending_transfers: len={}", .{transfers.len});
+
+            for (transfers) |expired| {
+                assert(expired.flags.pending);
+                assert(expired.timeout > 0);
```
sentientwaffle, line 1943:
```suggestion
                assert(expired.timeout > 0);
                assert(expired.amount > 0);
```

src/state_machine.zig
```diff
@@ -1599,6 +1898,86 @@ pub fn StateMachineType(
...
+                cr_account_new.credits_pending -= expired.amount;
+
+                grooves.accounts.update(.{ .old = dr_account, .new = &dr_account_new });
+                grooves.accounts.update(.{ .old = cr_account, .new = &cr_account_new });
+
+                // Expire the transfer.
+                grooves.posted.insert(&.{
+                    .timestamp = expired.timestamp,
+                    .fulfillment = .expired,
+                });
```
sentientwaffle: Could the `grooves.posted.insert` be moved into `self.transfer_update_pending_status`?

src/vsr/replica.zig
batiati: >which will queue the pulse at the back of the queue -- behind the request it is supposed to precede. (Or am I missing something here?)

1. `send_request_to_self` will always insert the request into the `loopback_queue` and call `flush_loopback_queue()`.
2. `delay_request`, by its turn, will always return the request to the request queue's head.

Does it make sense?  I was trying something overcomplicated when I figured it out.

src/vsr.zig
```diff
@@ -222,6 +222,8 @@ pub const Operation = enum(u8) {
     reconfigure = 3,
     /// The value 4 is reserved for pulse request.
     pulse = 4,
+    /// The value 4 is is reserved for release-upgrade requests.
```
sentientwaffle:
```suggestion
    /// The value 5 is is reserved for release-upgrade requests.
```
(Or swap them)

src/vsr.zig
```diff
@@ -220,23 +220,30 @@ pub const Operation = enum(u8) {
...
     reconfigure = 3,
+    /// The value 4 is reserved for pulse request.
+    pulse = 4,
     /// The value 4 is is reserved for release-upgrade requests.
-    upgrade = 4,
+    upgrade = 5,
 
     /// Operations <vsr_operations_reserved are reserved for the control plane.
     /// Operations ≥vsr_operations_reserved are available for the state machine.
     _,
```
sentientwaffle, line 230: (Feel free to do this in a follow-up PR, but) Can you add a `comptime` block which loops over the vsr-reserved operations, and verifies that the only one which `StateMachine.operation_from_vsr` can return non-`null` for is `pulse`?

src/vsr/replica.zig
```diff
@@ -3011,6 +3027,35 @@ pub fn ReplicaType(
...
 
+        fn on_pulse_timeout(self: *Self) void {
+            if (constants.aof_recovery) unreachable;
+            assert(self.status == .normal);
+            assert(self.primary());
+            assert(self.pulse_timeout.ticking);
+
+            defer self.pulse_timeout.reset();
+            if (self.pipeline.queue.prepare_queue.full()) return;
+            // Solo replicas only change views immediately when they start up.
```
```suggestion
            // Solo replicas only change views immediately when they start up,
            // and during that time they do not accept requests.
            // See Replica.open() for more detail.
```

src/vsr/replica.zig
```diff
@@ -9947,13 +10051,28 @@ const PipelineQueue = struct {
...
     fn push_request(pipeline: *PipelineQueue, request: Request) void {
         assert(request.message.header.command == .request);
+        pipeline.assert_request_queue(request);
+
+        pipeline.request_queue.push_assume_capacity(request);
+        if (constants.verify) pipeline.verify();
+    }
+
+    fn delay_request(pipeline: *PipelineQueue, request: Request) void {
+        assert(request.message.header.command == .request);
```
sentientwaffle: Can we assert that, when this is called, the pipeline prepare queue is full?

src/vsr/replica.zig
```diff
@@ -1454,10 +1465,15 @@ pub fn ReplicaType(
...
             assert(message.header.view <= self.view); // The client's view may be behind ours.
 
-            const realtime = self.clock.realtime_synchronized() orelse {
-                log.err("{}: on_request: dropping (clock not synchronized)", .{self.replica});
-                return;
-            };
+            // Messages with `client == 0` are sent from itself,
+            // it's not needed to get a different timestamp.
+            const realtime: i64 = if (message.header.client == 0)
+                @intCast(message.header.timestamp)
```
sentientwaffle, line 1473: Checking my understanding -- the header timestamp here is only non-zero for AOF recovery, right?

src/vsr/replica.zig
```diff
@@ -3011,6 +3027,25 @@ pub fn ReplicaType(
             }
         }
 
+        fn on_pulse_timeout(self: *Self) void {
+            assert(!constants.aof_recovery);
+            assert(self.status == .normal);
+            assert(self.primary());
+            assert(self.pulse_timeout.ticking);
+
+            defer self.pulse_timeout.reset();
```
sentientwaffle: No need to defer this.

src/vsr/replica.zig
batiati: Yes, it relies on this code to use the prepare timestamp (as desired)
However, the AOF already has a specific block to this.

https://github.com/tigerbeetle/tigerbeetle/blob/4fa955a01e66eda74ddec42375cf5911e9ba2728/src/vsr/replica.zig#L5592-L5600

src/vsr/replica.zig
```diff
@@ -3656,6 +3696,10 @@ pub fn ReplicaType(
                     self.primary_pipeline_prepare(request);
                 }
 
+                if (self.pulse_needed()) {
+                    self.send_request_pulse_to_self();
+                }
```
```suggestion
                if (self.pulse_needed()) {
                    assert(self.upgrade_release == null);
                    self.send_request_pulse_to_self();
                }
```
(To be symmetrical :smile:)

docs/reference/operations/get_account_history.md
```diff
@@ -11,18 +11,21 @@ Fetch the historical [`AccountBalance`](../account_balances.md)s of a given [`Ac
...
 
 - If the account has the flag [`history`](../accounts.md#flagshistory) set and any matching
-  balances exist, return an array of [`AccountBalance`](../account_balances.md)s.  
+  balances exist, return an array of [`AccountBalance`](../account_balances.md)s.
 - If the account does not have the flag [`history`](../accounts.md#flagshistory) set,
-  return nothing.  
-- If no matching balance exist, return nothing.  
-- If any constraint is violated, return nothing. 
+  return nothing.
+- If no matching balance exist, return nothing.
```
sentientwaffle:
```suggestion
- If no matching balance exists, return nothing.
```
Or change "balance" to plural, which we do in a prior bullet point

sentientwaffle (approved):

PR #1659, opened 2024-03-06
VSR: Header.release

sentientwaffle:
### Rename `Header.version` to `Header.protocol`

To disambiguate from the "release" version.

The difference is:

- (Wire) protocol version changes very infrequently.
- Release version changes with every (weekly) release, and is independent per state-machine/release-infrastructure.

### Add `Header.release`

Not all header types set a (nonzero) `release`. Similar to `Header.view`, there are enough headers that _do_ require a `release` that we include it in the frame anyway.

Rationale for which messages have `release == 0` vs `release != 0`:

- `Request.release` is the client's `release`.
- `Prepare.release` and ~`PrepareOk.release`~ `Reply.release` are _the same as_ `Request.release`. This is _not necessarily_ the same as the release that the cluster is running as it commits this op.
- Clients inform the cluster of their release.
- The cluster informs the clients of their release. (So this can be escalated to a upgrade warning/error if necessary.)

Nonzero `Header.release` fields are currently set to a placeholder (`1`).

### Ease protocol restriction for grid blocks

Even if we make a breaking protocol change, the legacy grid blocks must be handled indefinitely, since blocks are sent unwrapped over the network.

### DIFF
--- a/src/vsr/message_header.zig
+++ b/src/vsr/message_header.zig
@@ -48,7 +48,7 @@ pub const Header = extern struct {
     view: u32,
 
     /// The version of the protocol implementation that originated this message.
-    version: u16,
+    protocol: u16,
 
     /// The Viewstamped Replication protocol command for this message.
     command: Command,
@@ -162,7 +162,7 @@ pub const Header = extern struct {
         if (self.checksum_padding != 0) return "checksum_padding != 0";
         if (self.checksum_body_padding != 0) return "checksum_body_padding != 0";
         if (self.nonce_reserved != 0) return "nonce_reserved != 0";
-        if (self.version != vsr.Version) return "version != Version";
+        if (self.protocol != vsr.Version) return "protocol != Version";
         if (self.size < @sizeOf(Header)) return "size < @sizeOf(Header)";
         if (self.epoch != 0) return "epoch != 0";
         if (!stdx.zeroed(&self.reserved_frame)) return "reserved_frame != 0";
@@ -259,7 +259,7 @@ pub const Header = extern struct {
         size: u32,
         epoch: u32 = 0,
         view: u32 = 0,
-        version: u16 = vsr.Version,
+        protocol: u16 = vsr.Version,
         command: Command,
         replica: u8 = 0,
         reserved_frame: [16]u8,
@@ -286,7 +286,7 @@ pub const Header = extern struct {
         // NB: unlike every other message, pings and pongs use on disk view, rather than in-memory
         // view, to avoid disrupting clock synchronization while the view is being updated.
         view: u32,
-        version: u16 = vsr.Version,
+        protocol: u16 = vsr.Version,
         command: Command,
         replica: u8,
         reserved_frame: [16]u8 = [_]u8{0} ** 16,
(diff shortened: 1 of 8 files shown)


matklad:
> Prepare.release and PrepareOk.release are the same as Request.release. This is not necessarily the same as the release that the cluster is running as it commits this op.

This is a bit surprising, naively I'd expect prepare to use primary's version. Could you add a code comment explaining why we do it this way? Is this to leave a possibility of branching on the client version in the state machine?

src/vsr/message_header.zig
```diff
@@ -590,6 +612,7 @@ pub const Header = extern struct {
                     if (self.request != 0) return "root: request != 0";
                 },
                 else => {
+                    if (self.release == 0) return "release == 0";
```
matklad, line 617: Assert that release is zero for root and reserved.

src/vsr/message_header.zig
```diff
@@ -663,10 +688,11 @@ pub const Header = extern struct {
         size: u32 = @sizeOf(Header),
         epoch: u32 = 0,
         view: u32,
-        version: u16 = vsr.Version,
+        protocol: u16 = vsr.Version,
+        release: u16 = 0,
```
matklad: Comment that it is release of prepare and request

src/vsr/message_header.zig
```diff
@@ -1101,10 +1149,11 @@ pub const Header = extern struct {
         size: u32 = @sizeOf(Header),
         epoch: u32 = 0,
         view: u32,
-        version: u16 = vsr.Version,
+        protocol: u16 = vsr.Version,
+        release: u16 = 0, // Always 0.
```
matklad: Are we going to evict clients based on release? If we are, might be helpful to include replcia's release here?

src/vsr/message_header.zig
sentientwaffle: Ahh, we _could_ set it to that, but this was actually a typo in my PR description; I intended to set this to zero.

src/vsr/message_header.zig
sentientwaffle: Great idea! Yes, we will evict based on release.

sentientwaffle:
Note that I should have said that `Prepare.release` and `Reply.release` match `Request.release`; `PrepareOk.release` is zero. This was a typo in the original PR description. (Fixed now though.)

> Is this to leave a possibility of branching on the client version in the state machine?

That is the idea, though I'm open to switching it. If we switch it, there is still the possibility of a fallback in the form of conditionally (branched on the client version) rewriting the request body when it is converted to a prepare. (I think that the latter fallback is a little weaker than the former though.)

Another reason for setting `Reply.release == Request.release` is that it emphasizes that the format of the reply body _must_ be compatible with the client's (potentially older) version, even if the cluster is ahead.

In either case, each replica already knows which release should commit each prepare based on the combination of "what checkpoint is this op within" and "what is the release version at that checkpoint" (where the checkpoint's release version will be modified by committing `operation=upgrade` messages).

matklad (approved):

PR #1660, opened 2024-03-06
LSM: Pacing, Pipelining & Performance

cb22:
Turn back, traveler. Here be :dragon:s.

(Full write-up, rebase and production ready version coming soon™)

### DIFF
--- a/src/lsm/compaction.zig
+++ b/src/lsm/compaction.zig
@@ -1,6 +1,6 @@
-//! Compaction moves or merges a table's values into the next level.
+//! Compaction moves or merges a table's values from the previous level.
 //!
-//! Each Compaction is paced to run in one half-bar.
+//! Each Compaction is paced to run in an arbitrary amount of beats, by the forest.
 //!
 //!
 //! Compaction overview:
--- a/src/lsm/compaction.zig
+++ b/src/lsm/compaction.zig
@@ -42,6 +42,7 @@ const tracer = @import("../tracer.zig");
 const constants = @import("../constants.zig");
 
 const stdx = @import("../stdx.zig");
+const FIFO = @import("../fifo.zig").FIFO;
 const GridType = @import("../vsr/grid.zig").GridType;
 const BlockPtr = @import("../vsr/grid.zig").BlockPtr;
 const BlockPtrConst = @import("../vsr/grid.zig").BlockPtrConst;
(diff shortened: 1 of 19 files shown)


src/lsm/forest.zig
```diff
@@ -162,22 +179,506 @@ pub fn ForestType(comptime _Storage: type, comptime groove_cfg: anytype) type {
...
+                    .grid = grid,
+                };
+            }
+
+            pub fn deinit(self: *CompactionPipeline, allocator: mem.Allocator) void {
+                for (self.compaction_blocks) |block| allocator.free(block.block);
+                allocator.free(self.compaction_blocks);
+            }
+
+            pub fn reset(self: *CompactionPipeline) void {
```
matklad, line 260: Any way we can assert here that there are no in-progress compactions during reset? 

I think we can at leas check that all next pointers are null?

src/lsm/forest.zig
```diff
@@ -162,22 +179,506 @@ pub fn ForestType(comptime _Storage: type, comptime groove_cfg: anytype) type {
...
+            /// |-----------------------------|-----------------------------|
+            /// | Table A     | Table B       | Table A     | Table B       |
+            /// -------------------------------------------------------------
+            fn divide_blocks(self: *CompactionPipeline) CompactionHelper.CompactionBlocks {
+                // Some blocks only need to be valid for a beat, after which they're used for
+                // the next compaction.
+                const minimum_block_count_beat: u64 = blk: {
+                    var minimum_block_count: u64 = 0;
+
+                    // We need a minimum of 2 source value blocks; one from each table.
```
matklad, line 280: Nit, but I think we usually call those "data blocks" rather than "value blocks"?

src/lsm/forest.zig
```diff
@@ -162,22 +179,506 @@ pub fn ForestType(comptime _Storage: type, comptime groove_cfg: anytype) type {
...
+                const target_value_blocks = blocks[2 + equal_split_count + equal_split_count ..][0..equal_split_count];
+
+                std.log.info("divide_blocks: block_count_bar={} source_index_blocks.len={} source_value_level_a.len={} source_value_level_b.len={} target_value_blocks.len={}", .{
+                    block_count_bar,
+                    source_index_blocks.len,
+                    source_value_level_a.len,
+                    source_value_level_b.len,
+                    target_value_blocks.len,
+                });
+
```
matklad, line 353: Assert lower bounds on `source_index_blocks`, `source_value_level_a`, `source_value_level_b` and `target_value_blocks` `.len`s. 

I think: 

* `index_blocks.len == 2`
* `source_value_level_a,_b >= 2` (the code looks like we need at least two, though I don't understnd this fully yet)
* `target_value_blocks` >= 1

matklad: Assert that all subslices are disjoint:

```zig
                {
                    const subdivision: []const []CompactionHelper.CompactionBlock = &.{
                        source_index_blocks,
                        source_value_level_a,
                        source_value_level_b,
                        target_value_blocks,
                    };
                    for (subdivision, 0..) |slice_a, i| {
                        for (subdivision[0..i]) |slice_b| {
                            assert(stdx.disjoint_slices(
                                CompactionHelper.CompactionBlock,
                                CompactionHelper.CompactionBlock,
                                slice_a,
                                slice_b,
                            ));
                        }
                    }
                }
```

src/lsm/forest.zig
```diff
@@ -162,22 +179,506 @@ pub fn ForestType(comptime _Storage: type, comptime groove_cfg: anytype) type {
...
+                    break :blk block_count;
+                };
+
+                assert(self.compaction_blocks.len >= block_count_bar + minimum_block_count_beat);
+
+                // Reserve the top of the block address space for the bar scoped blocks.
+                const blocks = self.compaction_blocks[0 .. self.compaction_blocks.len - block_count_bar];
+                assert(blocks.len >= minimum_block_count_beat);
+
+                for (blocks) |*block| block.next = null;
```
matklad: Rather than nulling pointers here, _assert_ they are null, and null instead when the blocks are returned to the pool, to assert that the blocks have correct lifetimes.

src/lsm/compaction.zig
```diff
@@ -42,48 +42,313 @@ const tracer = @import("../tracer.zig");
...
+            // CompactionBlocks are stored in buffers where we need a pointer to get back to our
+            // parent.
+            target: ?*anyopaque = null,
+
+            // TODO: This could be a union to save a bit of memory and add a bit of safety.
+            read: Grid.Read = undefined,
+            write: Grid.Write = undefined,
+
+            next: ?*CompactionBlock = null,
+        };
```
matklad, line 119: Consider adding

```zig
stage: enum { free, pending, ready, ioing }
```

field to `CompactionBlock`  and asserting that the block is at a correct stage on use.

src/lsm/forest.zig
```diff
@@ -162,22 +179,506 @@ pub fn ForestType(comptime _Storage: type, comptime groove_cfg: anytype) type {
...
+                        // that all our blocks are distinct would be great. Both here, and for the beat blocks set up
+                        // by divide_blocks...
+                        const immutable_table_a_block = compaction_blocks[compaction_blocks.len - 1].block;
+                        log.info("Immutable dummy block is: [{}]", .{compaction_blocks.len - 1});
+
+                        const blocks = compaction_blocks[compaction_blocks.len - 3 .. compaction_blocks.len - 1];
+                        log.info("Index block range is: [{}..{}]", .{ compaction_blocks.len - 3, compaction_blocks.len - 1 });
+                        for (blocks) |*block| block.next = null;
+
+                        compaction_blocks.len -= 3;
```
matklad, line 415: Oh dear, didn't realize one can just assign to a slice length this way :fearful: 

Here be :dragon: :dragon:  indeed!

src/lsm/forest.zig
```diff
@@ -162,22 +179,506 @@ pub fn ForestType(comptime _Storage: type, comptime groove_cfg: anytype) type {
...
+                    // FIXME: Resetting these below variables like this sucks.
+                    self.beat_active.unset(self.slots[0].?.compaction_index);
+
+                    self.slot_filled_count = 0;
+                    assert(self.slot_running_count == 0);
+                    self.state = .filling;
+                    self.slots = .{ null, null, null };
+
+                    return self.advance_pipeline();
+                }
```
matklad:
```suggestion
                } else unreachable;
```

matklad:
Some high level asserts I'd love to see:

* to make sure we don't alias pooled blocks, it would help if each `CompactionBlock` stored its current state: which compaction owns the block, and which state of the pipeline owns the block. Then, in pipeline stages callback, asssert thate the owner and the stage are still valid
* To make sure that we didn't skip over any work: let compaction maintain separate counters for
  * `values_read_count` (incremented by the read stage based on data block size)
  * `values_merged_input` / `values_merged_output` incremented by the merge step by comparing compaction state before/after `merge_values` (we could also maintain the counter interanlly in `merge_values`, but that's a hot loop, so I'd rather not burn extra registers there, and I am fairly certain that merge is corret)
  * `values_write_count` incrementdeb by the write step based on the size of the data block
  * at the end of compaction, assert that the values agree with each other, and match the manifest
  * If input table total value count = total values read by read stage = total values read by the merge stage, I think we guarantee that no water was spilled from the interconnected bathtubs! 
* no good ideas how to check that we don't have races! I am fairly certain that we can't race on acquiring reservations, but it's not obvious that we can't race on _using_ reservations...

src/lsm/forest.zig
cb22: We're trying to move away from calling them data blocks and towards value blocks.

src/lsm/forest.zig
cb22: > the code looks like we need at least two, though I don't understnd this fully yet

We only need one, now! It used to be that we needed 2 (for `target_value_blocks` too), when we had a hard memory split between pipelines. The way the code started out was that blocks were divided into 2 splits, and there was a `current_split` that would be alternated between:

```
source_value_level_a = [[block, ...], [block, ...]]

read(0) // Reads blocks from disk into source_value_level_a[0]
merge(0) // Consumes blocks from source_value_level_a[0] (and source_value_level_b[0]) and writes them into target_value_blocks[0]
write(0) // Writes blocks to disk from target_value_blocks[0]
```

This was to allow for pipeline stages running concurrently:

```
read(0)   |    merge(0)    |    write(0)
          |    read(1)     |    merge(1)
          |                |    read(0)
```

The write stage was fine, but the logic around managing the state of the read blocks was tricky and didn't work out nicely, hence the switch to the current style.

src/lsm/forest.zig
cb22: Heh. I took the `A slice is a pointer and a length` from Zig's docs to heart.

I thought it read clearer and would be less error prone than doing the closed form way based on `i` but I'm happy to change it if you think otherwise...

cb22:
> Some high level asserts I'd love to see:

Those are great, I'll get on adding them!

> no good ideas how to check that we don't have races! I am fairly certain that we can't race on acquiring reservations, 

Yes, so the path to `beat_grid_reserve` is entirely sync (after being invoked by `state_machine.zig`'s `compact()`)

> but it's not obvious that we can't race on _using_ reservations...

This is trickier indeed. So walking backwards...

* The decision to finish a block is based on if a block is full or if the work is done for the bar.
* The blocks are acquired from the reservation in `check_and_finish_blocks`. 
* `check_and_finish_blocks` is called at the end of every `blip_merge`. 
* There's a barrier on the pipeline, so only one merge stage (potentially with a read and write stage) can be running at a time.
* Blips are memory independent, that is, we'll repeat a sequence of `blip_read` `blip_merge` `blip_write` until we process the same number of target values.

Not obvious indeed - I'll think how we can maybe add some asserts around it. Empirically, I have checked it with multiple varying memory allocations and ensured that blocks with the same checksums in the same addresses are generated.

src/lsm/compaction.zig
```diff
@@ -92,396 +377,1142 @@ pub fn CompactionType(
...
+                    tree.config.name,
+                    tree.table_immutable.mutability.immutable.snapshot_min,
+                    op,
+                    range_b.tables.count() + 1,
+                    values_count,
+                });
+
+                compaction_tables_value_count += values_count;
+                for (range_b.tables.const_slice()) |*table| {
+                    compaction_tables_value_count += table.table_info.value_count;
```
matklad, line 695: would be great to assert that `compaction_tables_value_count` is not to big, using Table.value_count_max

src/lsm/forest.zig
cb22: I'm very rusty with the reset code - can't it be called at any arbitrary moment?

matklad: Ah, you are actually correct! We might call reset with in-progress compaction, and that's actually the reason why we need reset at all!

The situation is like this:

* a replica compacts
* it gets stuck reading a block due to local corruption
* and the block is missing from the entire cluster, because the rest of the replicas moved over to the next checkpoint. 

in his situation:

* replica gets a newer checkpoint
* replica _cancels_ all outstanding grid reads and writes (dropping callbacks on the floor)
* replica installs the new superblock
* replica resets memory of all pending data structurues

So, yeah, this absolutely can get reset while some compaction is stuck on reading

matklad (approved):
Let's merge this!

I have a bunch of nitpicky comments which I have strategically held off, but I wasn't able to find any bugs here, the added assertions are good, and it looks like VOPR also thinks that this is fine. 

I'd love to iterate on this for a couple of weeks post merge though :rofl:

cb22:
> I'd love to iterate on this for a couple of weeks post merge though 🤣

:100: - so many performance (and other!) improvements possible!

PR #1661, opened 2024-03-06
docs: clarify language in Client Requests & Sessions

emschwartz:
- **docs: clarify language in Client Requests doc**
- **docs: clarify language in Client Sessions doc**

### DIFF
--- a/docs/design/client-requests.md
+++ b/docs/design/client-requests.md
@@ -26,8 +26,8 @@ operation events in each request.
 For the best performance, each request should batch as many events as possible. Typically this means
 funneling events through fewer client instances (e.g. a single client instance per process).
 
-The maximum number of events per batch depends on the maximum message size
-(`config.message_size_max`, specified in bytes) and the operation type.
+The maximum number of events per batch depends on the maximum message size and the operation type.
+(TODO: Expose each operation's batch sizes in the client).
 
 In the default configuration, the batch sizes are:
 
--- a/docs/design/client-requests.md
+++ b/docs/design/client-requests.md
@@ -32,18 +32,18 @@ the client instead).
 
 In the default configuration, the batch sizes are:
 
-| Operation               | Batch Size (Bytes) |
-| ----------------------- | -----------------: |
-| `lookup_accounts`       |               8190 |
-| `lookup_transfers`      |               8190 |
-| `create_accounts`       |               8190 |
-| `create_transfers`      |               8190 |
-| `get_account_transfers` |               8190 |
-| `get_account_history`   |               8190 |
+| Operation               | Request Batch Size (Events) | Reply Batch Size (Events) |
+| ----------------------- | --------------------------: | ------------------------: |
+| `lookup_accounts`       |                        8190 |                      8190 |
+| `lookup_transfers`      |                        8190 |                      8190 |
+| `create_accounts`       |                        8190 |                      8190 |
+| `create_transfers`      |                        8190 |                      8190 |
+| `get_account_transfers` |                           1 |                      8190 |
+| `get_account_history`   |                           1 |                      8190 |
 
-The application layer can batch events by itself, however, when the same client instance is shared
-by multiple threads/tasks, it is able to transparently batch requests of the same operation type in
-a single message.
+You can design your application to batch events manually. However, client instances automatically
+batch requests of the same operation type. Therefore, sharing the same client instance between
+multiple threads or tasks enables events to be batched transparently.
 
 - [Node](/src/clients/node/README.md#batching)
 - [Go](/src/clients/go/README.md#batching)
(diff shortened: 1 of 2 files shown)


docs/design/client-requests.md
```diff
@@ -4,61 +4,73 @@ sidebar_position: 3
...
 
-| Operation               | Batch Size |
-| ----------------------- | ---------: |
-| `lookup_accounts`       | 8190       |
-| `lookup_transfers`      | 8190       |
-| `create_accounts`       | 8190       |
-| `create_transfers`      | 8190       |
-| `get_account_transfers` | 8190       |
-| `get_account_history`   | 8190       |
+| Operation               | Batch Size (Bytes) |
```
sentientwaffle: This table counts events, not bytes.

Also, maybe we should split up the "request batch" and "reply batch".
`get_account_transfers` and `get_account_history` both take exactly 1 `AccountFilter`, but return (potentially) many more results.

docs/design/client-sessions.md
```diff
@@ -4,18 +4,18 @@ sidebar_position: 4
 
 # Client Sessions
 
-A _client session_ is a sequence of alternating [requests](./client-requests.md) and replies between a
-client and a cluster.
+A _client session_ is a series [requests](./client-requests.md) and replies sent between a client
```
sentientwaffle: :thinking: I think "sequence" is more correct here? At least in my head it has a stronger denotation of order.

docs/design/client-requests.md
```diff
@@ -4,61 +4,73 @@ sidebar_position: 3
...
 
-In some application architectures, the number of services that need to query TigerBeetle may:
 - [exceed `config.clients_max`](./client-sessions.md#eviction), or
-- may require additional [batching](#batching-events) to optimize throughput.
+- require additional [batching](#batching-events) to optimize throughput.
+
+Rather than each service connecting to TigerBeetle directly, you can set up your application
+services to forward their requests to a pool of intermediate services, as illustrated below. This
+API layer can coalesce events from many application services into requests, and forward back the
+respective replies.
```
sentientwaffle: We should probably clarify that TigerBeetle does not provide this intermediate service or a client to pool/connect to them.

docs/design/client-requests.md
```diff
@@ -4,61 +4,73 @@ sidebar_position: 3
...
+| `create_accounts`       |               8190 |
+| `create_transfers`      |               8190 |
+| `get_account_transfers` |               8190 |
+| `get_account_history`   |               8190 |
 
 The application layer can batch events by itself, however, when the same client instance is shared
-by multiple threads/tasks, it is able to transparently batch requests of the same operation type
-in a single message.
+by multiple threads/tasks, it is able to transparently batch requests of the same operation type in
+a single message.
```
sentientwaffle: Maybe split into two sentences, there are a lot of commas here.

sentientwaffle (commented):
For future reference, it would be convenient for reviewing if re-wrapping/formatting was in separate commits from the actual edits. :sweat_smile:

docs/design/client-requests.md
```diff
@@ -4,61 +4,76 @@ sidebar_position: 3
...
 
-| Operation               | Batch Size |
-| ----------------------- | ---------: |
-| `lookup_accounts`       | 8190       |
-| `lookup_transfers`      | 8190       |
-| `create_accounts`       | 8190       |
-| `create_transfers`      | 8190       |
-| `get_account_transfers` | 8190       |
-| `get_account_history`   | 8190       |
+| Operation               | Request Batch Size (Events) | Reply Batch Size (Events) |
```
sentientwaffle:
```suggestion
| Operation               | Request Batch Size (Events) | Reply Batch Size (Results) |
```

docs/design/client-requests.md
```diff
@@ -4,61 +4,76 @@ sidebar_position: 3
...
+operation events in each request. 
+
+For the best performance, each request should batch as many events as possible. Typically this means
+funneling events through fewer client instances (e.g. a single client instance per process).
 
 The maximum number of events per batch depends on the maximum message size
-(`config.message_size_max`) and the operation type.
-(TODO: Expose batch size in the client instead).
+(`config.message_size_max`, specified in bytes) and the operation type. (TODO: Expose batch size in
+the client instead).
```
sentientwaffle: :thinking: The TODO still applies, but I don't think we should mention `config.message_size_max` -- that is an internal detail, and possibly misleading. We don't expect users to modify/reconfigure the message size.

docs/design/client-requests.md
```diff
@@ -27,19 +27,18 @@ For the best performance, each request should batch as many events as possible.
 funneling events through fewer client instances (e.g. a single client instance per process).
 
 The maximum number of events per batch depends on the maximum message size
-(`config.message_size_max`, specified in bytes) and the operation type. (TODO: Expose batch size in
-the client instead).
+(`config.message_size_max`, specified in bytes) and the operation type.
```
sentientwaffle: I meant more like

```
The maximum number of events per batch depends on the maximum message size and the operation type. 
(TODO: Expose each operation's batch sizes in the client).
```

sentientwaffle (approved):

PR #1667, opened 2024-03-08
VSR: Add `release` to grid block header

sentientwaffle:
Each block's `header.release` is set to the release that is running when the block is generated.

The purpose of including `release` on blocks is for assertions and debugging/forensics. It doesn't impact any logic.

### DIFF
--- a/src/vsr/grid.zig
+++ b/src/vsr/grid.zig
@@ -669,6 +669,7 @@ pub fn GridType(comptime Storage: type) type {
         ) void {
             const header = schema.header_from_block(block.*);
             assert(header.cluster == grid.superblock.working.cluster);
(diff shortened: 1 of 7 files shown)


src/vsr/grid.zig
```diff
@@ -669,6 +669,7 @@ pub fn GridType(comptime Storage: type) type {
         ) void {
             const header = schema.header_from_block(block.*);
             assert(header.cluster == grid.superblock.working.cluster);
+            assert(header.release <= grid.superblock.working.vsr_state.checkpoint.release);
```
matklad, line 673: Might want the same on read_block?

matklad (approved):

PR #1669, opened 2024-03-08
Remove the java benchmark code

### LINKED ISSUE #1658: Java client jar includes Benchmark code
The java client includes the Benchmark code as part of the downloaded jar. 
While it may be useful for someone testing out the client jar to have the code immediately available, it does mean that the complete classpath of the application would include more than one `public static void main` method (one in the client jar and another in the client application). A compiled app would likely ignore the second psvm method, but it seems weird that it is there. 
Perhaps the Benchmark code could be moved to the src/test folder?

ddebree:
This resolves #1658


matklad:
Let's merge it as per https://github.com/tigerbeetle/tigerbeetle/issues/1658#issuecomment-1985554683 to avoid conflits with #1679

matklad (approved):

PR #1670, opened 2024-03-08
VSR: Ping releases

sentientwaffle:
### VSR: Add Replica.releases_bundled

This is presently stubbed out.
- For real (multiversion) binaries, this will be set to a list of releases compiled into the binary.
- For testing, we will manually set it to a list of releases (which may change between restarts).

### VSR: Ping bundled releases

To determine when the cluster can upgrade, each replica advertises which releases it is able/willing to upgrade to.

Note that pings will include all bundled releases -- newer, older, and currently-running. For the purposes of upgrades, only the newer versions actually matter, but the list isn't that large, so there isn't really a benefit to discarding/filtering the information out.

### VSR: Replica.upgrade_targets

Each replica tracks the releases that every other replica is advertising.

Note that only release versions are tracked -- if the protocol version changes, that will necessarily change the release version too.


src/config.zig
```diff
@@ -142,6 +142,7 @@ const ConfigCluster = struct {
     lsm_batch_multiple: comptime_int = 32,
     lsm_snapshots_max: usize = 32,
     lsm_manifest_compact_extra_blocks: comptime_int = 1,
+    vsr_releases_max: usize = 64,
```
sentientwaffle, line 145: `vsr_releases_max` is a new config/constant which defines the maximum number of releases that can be advertised by a replica in a `ping` message. (So this is effectively the maximum number of releases that can be compiled into a multiversion binary).

I set it to `64` right now, which I chose somewhat arbitrarily: We release on a weekly cadence, there are 52 weeks/year, rounded up to the nearest power-of-2 is 64.

src/vsr.zig
```diff
@@ -1044,6 +1047,18 @@ pub fn member_index(members: *const Members, replica_id: u128) ?u8 {
...
 
+pub fn verify_release_list(releases: []const u16) void {
+    assert(releases.len >= 1);
+    assert(releases.len <= constants.vsr_releases_max);
+
+    for (
+        releases[0 .. releases.len - 1],
+        releases[1..],
+    ) |release_a, release_b| {
+        assert(release_a > release_b);
```
matklad: This checks for the descending list, right? That is, that's the opposite of natural sort order? What's the reason for deviation?

src/vsr/replica.zig
```diff
@@ -487,6 +495,7 @@ pub fn ReplicaType(
             grid_cache_blocks_count: u32 = Grid.Cache.value_count_max_multiple,
             release: u16,
             release_client_min: u16,
+            releases_bundled: vsr.ReleaseList,
```
matklad: nit, but I'v bound bounded array to be clunky to work with (esp for vsr headers). Given that `vsr.ReleaseList` doesn't actually enforce validity in types (we need to call `verify_release_list`), then maybe just pass a slice of u16s here?

That is, I think the two stationary design points are:

* fully enforce semantic validity in types, such that, if the caller has `vsr.ReleaseList`, that's already a proof that the list is valid 
* use types to specify structural representation, and rely on assertions for invariance checking. That is, use simple types like `u16`, `[]u128` etc, instead of `vsr.Relaese`, `vsr.Checksum` etc.

Here we are sort-of in the middle, where the interface requires a `vsr.ReleaseList`, but that doesn't actually check most of invariants itself.

src/vsr/message_header.zig
```diff
@@ -308,16 +308,23 @@ pub const Header = extern struct {
...
         fn invalid_header(self: *const @This()) ?[]const u8 {
             assert(self.command == .ping);
-            if (self.size != @sizeOf(Header)) return "size != @sizeOf(Header)";
-            if (self.checksum_body != checksum_body_empty) return "checksum_body != expected";
+            if (self.size <= @sizeOf(Header)) return "size <= @sizeOf(Header)";
+            if (self.size > @sizeOf(Header) + @sizeOf(u16) * constants.vsr_releases_max) {
+                return "size > limit";
+            }
+            if (self.size != @sizeOf(Header) + self.release_count * @sizeOf(u16)) {
+                return "size != @sizeOf(Header) + release_count * @sizeOf(u16)";
```
matklad: Maybe add a separate check that `release_count <= vsr_releases_max`, so that we don't have to worry about overflow? 

I think the code is correct, but if, eg the `release_count` were an `u64` then it would be possible to get an overflow here. Seems like its more robust to check count limits individually before multiplying to check size limits.

src/vsr/replica.zig
```diff
@@ -2589,19 +2589,30 @@ pub fn ReplicaType(
...
                 .command = .ping,
+                .size = @intCast(@sizeOf(Header) + @sizeOf(u16) * self.releases_bundled.count()),
                 .cluster = self.cluster,
                 .replica = self.replica,
                 .view = self.view_durable(), // Don't drop pings while the view is being updated.
                 .release = self.release,
                 .checkpoint_id = self.superblock.working.checkpoint_id(),
                 .checkpoint_op = self.op_checkpoint(),
                 .ping_timestamp_monotonic = self.clock.monotonic(),
+                .release_count = @intCast(self.releases_bundled.count()),
```
matklad: Use `self.releases_bundled.count_as(u16)` to move the range check to compile time

src/vsr/replica.zig
```diff
@@ -2589,19 +2589,30 @@ pub fn ReplicaType(
         fn on_ping_timeout(self: *Self) void {
             self.ping_timeout.reset();
 
-            var ping = Header.Ping{
+            const message = self.message_bus.pool.get_message(.ping);
+            defer self.message_bus.unref(message);
+
+            message.header.* = Header.Ping{
                 .command = .ping,
+                .size = @intCast(@sizeOf(Header) + @sizeOf(u16) * self.releases_bundled.count()),
```
matklad: I think here we could also `.count_as(u32)` here to aviod intCast?

src/vsr/replica.zig
```diff
@@ -6614,6 +6625,15 @@ pub fn ReplicaType(
             self.send_message_to_other_replicas_base(message.base());
         }
 
+        fn send_message_to_other_replicas_and_standbys(self: *Self, message: *Message) void {
```
matklad, line 6662: I wonder why `send_header` functions don't delegate to `send_message` ones?

src/vsr/message_header.zig
```diff
@@ -308,16 +308,23 @@ pub const Header = extern struct {
...
+        release_count: u16,
 
-        reserved: [96]u8 = [_]u8{0} ** 96,
+        reserved: [94]u8 = [_]u8{0} ** 94,
 
         fn invalid_header(self: *const @This()) ?[]const u8 {
             assert(self.command == .ping);
-            if (self.size != @sizeOf(Header)) return "size != @sizeOf(Header)";
-            if (self.checksum_body != checksum_body_empty) return "checksum_body != expected";
+            if (self.size <= @sizeOf(Header)) return "size <= @sizeOf(Header)";
```
matklad: I think the code to validate the body of a ping is missing? I.e, checking that the releases are strictly sorted.

src/vsr/replica.zig
```diff
@@ -1314,6 +1325,30 @@ pub fn ReplicaType(
                 .ping_timestamp_monotonic = message.header.ping_timestamp_monotonic,
                 .pong_timestamp_wall = @bitCast(self.clock.realtime()),
             }));
+
+            if (message.header.replica < self.replica_count) {
+                const upgrade_targets = &self.upgrade_targets[message.header.replica];
+                if (upgrade_targets.* == null or
+                    upgrade_targets.*.?.timestamp > message.header.ping_timestamp_monotonic)
```
matklad: Huh, I think this is incorrect --- monotinic timestamps are not guaranteed to be monotinic across restarts. The doom scenario here would be a buggy replica advedtising a very high monotonic timestamp, then crashing, then restarting with a much smaller monotonic timestamp and not being able to sync versions. 

Not sure what's the best solution here, options I have in mind:

* use checkpoint_op/view for monotonicity (doesn't work for idle cluster)
* use op for monotonicity (better then checkpoint, but still doesnt' work for idle cluster)
* use real time --- that _should_ be stable across restarts, but is not monotinic overwise
* just always update the upgrade_target, making an assumption that all old messages in the network eventually die? 

I'd probably go for this combination:

* if the current target has higher view or checkpoint, skip the new target
* otherwise, just replace the target, ignoring the timestamp

I do not feel great about timestamps, far-future timestamp locking us up from an upgrade feels like a nasty failure mode.

src/vsr/replica.zig
```diff
@@ -1314,6 +1325,30 @@ pub fn ReplicaType(
...
+                {
+                    upgrade_targets.* = .{
+                        .checkpoint = message.header.checkpoint_op,
+                        .timestamp = message.header.ping_timestamp_monotonic,
+                        .releases = .{},
+                    };
+
+                    const releases = std.mem.bytesAsSlice(u16, message.body());
```
matklad: assert that releases are sorted.

matklad (commented):
The thing about time feels important, as well as validation of releases bundled with pings. Everything else are just nits!

src/vsr.zig
sentientwaffle: It was to emphasize that the highest versions are higher priorities... in retrospect, probably an unnecessary complication.

src/vsr/message_header.zig
sentientwaffle: This function doesn't have access to the message body. We do assert the order in `on_ping`, though.

src/vsr/replica.zig
sentientwaffle: I think it is due to `Message` pointers vs `MessageType` pointers, to keep the caller from needing to do any conversion.

src/vsr/replica.zig
```diff
@@ -1295,6 +1323,29 @@ pub fn ReplicaType(
...
+                        .view = message.header.view,
+                        .releases = .{},
+                    };
+
+                    const releases = std.mem.bytesAsSlice(u16, message.body());
+                    assert(releases.len == message.header.release_count);
+                    vsr.verify_release_list(releases);
+                    for (releases) |release| {
+                        if (release > self.release) {
+                            upgrade_targets.*.?.releases.append_assume_capacity(release);
```
matklad, line 1345: >append_assume_capacity

I think this can fire? We have a bound for a _single_ message, but different ping messages might have originated from different binaries, so their union might actually exceed what we have here. This is of course and edge case, but there's a related realistic problem: imagine if you run a binary with a new release on one replica, but then retire it in favor of the old binary. The knowledge that that replica has a new release persists until the viewchange/checkpoint boundary, although the actual binary does not support that version...

It feels like just replacing the whole set of targets might be safer? But then again, that assumes that the network drops all old messages eventually. 

Hm, I think I really don't like the union behavior because it is guaranteed to be wrong: if we have two pings with different release sets, then what we support is one release set or the other release set. When we union the two, we are guaranteed to have an unsupported release in there?

sentientwaffle: We replace the entire list every time, there is no union:

```
+                    upgrade_targets.* = .{
+                        .checkpoint = message.header.checkpoint_op,
+                        .view = message.header.view,
+                        .releases = .{},
+                    };
```
above resets the list to empty before we append anything.

matklad: :facepalm: gotta grab one more pu-erh cup as I don't seem to be sufficiently woken up, sorry!

matklad (dismissed):

matklad (approved):

PR #1671, opened 2024-03-08
docs: clarify more language in the Design section

emschwartz:
- **docs: move details about linked events into Client Requests doc**
- **docs: clarify language in Consistency doc**
- **docs: rewrap consistency doc at 100 chars**
- **docs: flesh out debits and credits section**
- **docs: mention clients' id() function**
- **docs: reword user_data section**
- **docs: remove advanced ID examples**
- **docs: rewrap at 100 chars**

### DIFF
--- a/docs/design/client-requests.md
+++ b/docs/design/client-requests.md
@@ -31,7 +31,7 @@ outcome of the next one in the request. These chains of events will all succeed 
 
 The last Transfer or Account in a batch may never have the `flags.linked` set, as it would leave a
 chain open-ended. Attempting to do so will result in the
-[`linked_event_chain_open`](./operations/create_transfers.md#linked_event_chain_open) error.
+[`linked_event_chain_open`](../reference/operations/create_transfers.md#linked_event_chain_open) error.
 
 Multiple chains of events may coexist within a batch to succeed or fail independently. 
 
@@ -39,7 +39,7 @@ Events within a chain are executed in order, or are rolled back on error, so tha
 event in the chain is visible to the next. Each chain is either visible or invisible as a unit to
 subsequent transfers after the chain. The event that was the first to fail within a chain will have
 a unique error result. Other events in the chain will have their error result set to
-[`linked_event_failed`](./operations/create_transfers.md#linked_event_failed).
+[`linked_event_failed`](../reference/operations/create_transfers.md#linked_event_failed).
 
 Consider this set of Transfers as part of a batch:
 
@@ -52,7 +52,7 @@ Consider this set of Transfers as part of a batch:
 | `E`      | `4`                | `false`      |
 
 If any of transfers `B`, `C`, or `D` fail (for example, due to
-[`exceeds_credits`](./operations/create_transfers.md#exceeds_credits), then `B`, `C`, and `D` will
+[`exceeds_credits`](../reference/operations/create_transfers.md#exceeds_credits), then `B`, `C`, and `D` will
 all fail. They are linked.
 
 Transfers `A` and `E` fail or succeed independently of `B`, `C`, `D`, and each other.
--- a/docs/design/client-requests.md
+++ b/docs/design/client-requests.md
@@ -7,16 +7,60 @@ sidebar_position: 3
 A _request_ is a [batch](#batching-events) of one or more [operation
 events](../reference/operations/index.md) sent to the cluster in a single message.
 
-- All events within a request batch must be of the same
-  [operation](../reference/operations/index.md) type. You cannot, for example, create accounts and
-  transfers in the same request.
-- The cluster commits an entire batch at once. Events are applied in series, such that successive
-  events observe the effects of previous ones and event timestamps are [totally
-  ordered](time.md#timestamps-are-totally-ordered).
-- Events within a request [succeed or fail](../reference/operations/create_transfers.md#result)
-  independently unless they are explicitly [linked](../reference/transfers.md#flagslinked).
-- The cluster returns a single reply for each unique request it commits. The reply contains a
-  [result](../reference/operations/create_transfers.md#result) for each event in the request.
+All events within a request batch must be of the same [operation](../reference/operations/index.md)
+type. You cannot, for example, create accounts and transfers in the same request.
+
+The cluster commits an entire batch at once. Events are applied in series, such that successive
+events observe the effects of previous ones and event timestamps are [totally
+ordered](time.md#timestamps-are-totally-ordered).
+
+The cluster returns a single reply for each unique request it commits. The reply contains a
+[result](../reference/operations/create_transfers.md#result) for each event in the request.
+
+## Linked Events
+
+Events within a request [succeed or fail](../reference/operations/create_transfers.md#result)
+independently unless they are explicitly linked using the `flags.linked`
+([`Account.flags.linked`](../reference/accounts.md#flagslinked) or
+[`Transfer.flags.linked`](../reference/transfers.md#flagslinked)).
+
+When the `linked` flag is specified, it links the outcome of a Transfer or Account creation with the
+outcome of the next one in the request. These chains of events will all succeed or fail together.
+
+**The last event in a chain is denoted by the first Transfer or Account without this flag.**
+
+The last Transfer or Account in a batch may never have the `flags.linked` set, as it would leave a
+chain open-ended. Attempting to do so will result in the
+[`linked_event_chain_open`](./operations/create_transfers.md#linked_event_chain_open) error.
+
+Multiple chains of events may coexist within a batch to succeed or fail independently. 
+
+Events within a chain are executed in order, or are rolled back on error, so that the effect of each
+event in the chain is visible to the next. Each chain is either visible or invisible as a unit to
+subsequent transfers after the chain. The event that was the first to fail within a chain will have
+a unique error result. Other events in the chain will have their error result set to
+[`linked_event_failed`](./operations/create_transfers.md#linked_event_failed).
+
+Consider this set of Transfers as part of a batch:
+
+| Transfer | Index within batch | flags.linked |
+|----------|--------------------|--------------|
+| `A`      | `0`                | `false`      |
+| `B`      | `1`                | `true`       |
+| `C`      | `2`                | `true`       |
+| `D`      | `3`                | `false`      |
+| `E`      | `4`                | `false`      |
+
+If any of transfers `B`, `C`, or `D` fail (for example, due to
+[`exceeds_credits`](./operations/create_transfers.md#exceeds_credits), then `B`, `C`, and `D` will
+all fail. They are linked.
+
+Transfers `A` and `E` fail or succeed independently of `B`, `C`, `D`, and each other.
+
+After the chain of linked events has executed, the fact that they were linked will not be saved. To
+save the association between Transfers or Accounts, it must be [encoded into the data
+model](../design/data-modeling.md), for example by adding an ID to one of the [user
+data](../design/data-modeling.md#user_data) fields.
 
 ## Batching Events
 
--- a/docs/design/client-requests.md
+++ b/docs/design/client-requests.md
@@ -52,7 +52,7 @@ Consider this set of Transfers as part of a batch:
 | `E`      | `4`                | `false`      |
 
 If any of transfers `B`, `C`, or `D` fail (for example, due to
-[`exceeds_credits`](../reference/operations/create_transfers.md#exceeds_credits), then `B`, `C`, and `D` will
+[`exceeds_credits`](../reference/operations/create_transfers.md#exceeds_credits)), then `B`, `C`, and `D` will
 all fail. They are linked.
 
 Transfers `A` and `E` fail or succeed independently of `B`, `C`, `D`, and each other.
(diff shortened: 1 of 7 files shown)


docs/design/consistency.md
```diff
@@ -34,53 +34,53 @@ What specific guarantees does TigerBeetle provide to applications?
 - A client session never observes a broken invariant (e.g.
   [`flags.credits_must_not_exceed_debits`](../reference/accounts.md#flagscredits_must_not_exceed_debits)
   or [`flags.linked`](../reference/transfers.md#flagslinked)).
-- Multiple client sessions may receive replies [out of order](#reply-order) relative to one another.
+- Multiple client sessions may receive replies [out of order](#reply-order) relative to one another. For example, if two clients submit requests around the same time, the client whose request is commmitted first might receive the reply later.
 - A client session can consider a request executed when it receives a reply for the request.
 
-#### Requests
+#### [Requests](./client-requests.md)
+> Note that a request refers to a batch, rather than a single operation.
```
sentientwaffle: I think "operation" might be a confusing term to use here, since each batch/request has exactly one operation. Maybe "event" instead?

docs/design/consistency.md
```diff
@@ -34,53 +34,53 @@ What specific guarantees does TigerBeetle provide to applications?
...
 
 - A request executes within the cluster at most once.
-- Requests do not [time out](#retries) — a timeout typically implies failure, which cannot be
-  conclusively determined in the context of network faults.
+- Requests do not [time out](#retries) (a timeout typically implies failure, which cannot be
```
sentientwaffle: These should be separate clauses:

 - Requests do not [time out](#retries). (A timeout typically implies failure, which cannot be

docs/design/consistency.md
```diff
@@ -194,17 +194,16 @@ But suppose the application crashes and restarts immediately after sending its r
   2. _Application_: Send "create account" request `A₁` to the cluster.
   3. _Application_: Crash. Restart.
   4. _Cluster_: Create `A₁`; reply `ok` — but the application session has reset,
-    so this reply never reaches the application).
-  5. _Application_: Send "create account" request `A₁` to the cluster.
+    so this reply never reaches the application.
+  5. _Application_: Send "create account" request `A₁` to the cluster with the same ID as in step 1.
```
sentientwaffle: Can you add a comma before "with the same ID"? On my first read-through I thought you were referring to the cluster id, not the account id.

docs/design/consistency.md
```diff
@@ -34,53 +34,53 @@ What specific guarantees does TigerBeetle provide to applications?
...
 - Once committed, an event will always be committed — the cluster's state never backtracks.
-- Within a cluster, object timestamps are unique.
-  For all objects `A` and `B` belonging to the same cluster, `A.timestamp ≠ B.timestamp`.
-- Within a cluster, object timestamps are strictly increasing.
-  For all objects `A` and `B` belonging to the same cluster, if `A.timestamp < B.timestamp`,
-  then `A` was committed earlier than `B`.
-- If a client session is terminated and restarts, it is guaranteed to see updates for which the
+- Within a cluster, object [timestamps are unique and strictly increasing](./time.md#timestamps-are-totally-ordered).
+  For all objects `A` and `B` belonging to the same cluster, `A.timestamp ≠ B.timestamp` and
+  if `A.timestamp < B.timestamp`, then `A` was committed earlier than `B`.
```
sentientwaffle: I think I slightly prefer these as separate points. :sweat_smile: 
But if they are joined, can you add a comma after `A.timestamp ≠ B.timestamp`? I think it reads a little easier that way.

docs/design/data-modeling.md
```diff
@@ -22,17 +22,33 @@ In double-entry accounting, an account balance is the difference between the two
...
+
+### Debit Balances
+
+`balance = debits - credits`
+
+By convention, debit balances are used to represent:
+- Operator's Assets
+- Operator's Expenses
+
+To limit the balance for this type of account, use
```
sentientwaffle: Maybe "To enforce a positive debit balance, use"
(and likewise for credit balances).

docs/design/data-modeling.md
```diff
@@ -109,17 +109,20 @@ When selecting an `id` scheme:
 ### Examples (Recommended)
 #### Time-Based Identifiers
 
-A time-based identifier (such as ULID or UUIDv7) are recommended for most applications.
+A time-based identifier (such as ULID or UUIDv7) is recommended for most applications.
```
sentientwaffle: We should remove mention of the ULID specs.. those have the wrong endian-ness to take advantage of the LSM optimization.

sentientwaffle: This section needs amending -- ULID/UUIDv7 are actually _anti_-recommended. They encode bytes in the wrong order to take advantage of the LSM optimizations.

We should describe our own id scheme, and be very clear that it is not the same as (though is inspired by) ULIDv7. I'm not sure how we should refer to our scheme, though..

docs/design/data-modeling.md
```diff
@@ -109,17 +109,20 @@ When selecting an `id` scheme:
...
+
+TigerBeetle clients include an `id()` function to generate IDs using these recommendations.
```
sentientwaffle: > generate IDs using these recommendations

What recommendations?

docs/design/data-modeling.md
```diff
@@ -109,17 +109,20 @@ When selecting an `id` scheme:
...
 
-**Important**: When creating multiple objects during the same millisecond, increment the random bytes
-(instead of generating new random bytes). Make sure to also store random bytes first, timestamp bytes 
-second, and both in little-endian. These details ensure that a sequence of objects have strictly 
-increasing ids according to the server. (Such ids are amenable to LSM optimizations, 
+**Important**: If you generate IDs yourself instead of using the `id()` function provided by the
+client libraries, **increment the random bytes** when creating multiple objects during the same
+millisecond (instead of generating new random bytes). Make sure to also store random bytes first,
+timestamp bytes second, and both in little-endian. These details ensure that a sequence of objects
+have strictly increasing ids according to the server. (Such ids are amenable to LSM optimizations,
```
sentientwaffle: I think "first"/"second" is potentially confusing in this context, can we use "low-order" and "high order" instead?

docs/design/data-modeling.md
```diff
@@ -109,17 +109,20 @@ When selecting an `id` scheme:
 ### Examples (Recommended)
 #### Time-Based Identifiers
 
-A time-based identifier (such as ULID or UUIDv7) are recommended for most applications.
+A time-based identifier (such as ULID or UUIDv7) is recommended for most applications.
+
+TigerBeetle clients include an `id()` function to generate IDs using these recommendations.
 
 A ULID ("Universally Unique Lexicographically Sortable identifier") consists of:
```
sentientwaffle: Again, we should make it clear that we are describing our own ID scheme, not any spec.

docs/design/consistency.md
emschwartz: I think I don't totally understand what this means:
> A timeout typically implies failure, which cannot be conclusively determined in the context of network faults.

sentientwaffle: Consider case A:

1. Client connects to cluster.
2. Client sends request X.
3. (There is a network partition between the client and the cluster -- the cluster never receives request X.)

And case B:

1. Client connects to cluster.
2. Client sends request X.
3. Cluster receives request X.
4. Cluster sends reply X to client.
5. (There is a network partition between the client and the cluster -- the client never receives reply X.)

From the client's point of view, these are indistinguishable!

If it were somehow possible to distinguish cases A and B at the client, then it would be safe for `client.request()` to throw a "not connected" error after some timeout in case A. But if the client threw a "not connected" error in case B, 
then the client/application would assume that X was not executed, which is incorrect!

As such, we handle these cases the same way -- the client keeps retrying a request until it receives a reply, at which point is is certain that the cluster has processed it.

The (big) downside here is that network (and configuration) issues exist -- right now we hide all issues, which can make troubleshooting difficult. In the future I expect we'll expose out-of-band warnings if we _suspect_ that the cluster is inaccessible.

docs/design/data-modeling.md
emschwartz: TigerBeetle Time-Based IDs? TBTBIDs?

docs/design/consistency.md
```diff
@@ -4,83 +4,96 @@ sidebar_position: 5
...
+  [timeouts](../reference/transfers.md#timeout).
 - Requests retried by their original client session receive identical replies.
 - Requests retried by a different client (same request body, different session) may receive
   [different replies](#consistency-with-foreign-databases).
-- Events within a request are executed in sequential order.
-- Events within a request do not interleave with events from other requests.
-  (TODO: Can timeouts interleave batches, or should we shift the batch so that timeouts land
-  entirely before/after?)
-- All events within a request batch are committed, or none are.
+- Events within a request are executed in sequential order. The effects of a given event will be
```
sentientwaffle: Is "sequential" redundant here? Or we could just say "executed in sequence".

sentientwaffle: Also, s/'will be'/'are' might be clearer. "will be observable when the next event within that request is applied" sounds ambiguous, as if even "X" might be observable only _after_ "X+1" is applied.

docs/design/consistency.md
```diff
@@ -4,83 +4,96 @@ sidebar_position: 5
...
+  observable when the next event within that request is applied.
+- Events within a request do not interleave with events from other requests. (TODO: Can timeouts
+  interleave batches, or should we shift the batch so that timeouts land entirely before/after?)
```
sentientwaffle: You can remove this TODO. (The answer to the question it asks turns out to be quite complicated; see `references/transfers.md#timeout`).

docs/design/consistency.md
```diff
@@ -4,83 +4,96 @@ sidebar_position: 5
...
+- All events within a request batch are committed, or none are. Note that this does not mean that
+  all of the events in a batch will either succeed or fail. Events succeed or fail independently
+  unless they are explicitly [linked](./client-requests.md#linked-events)
```
sentientwaffle: "All of the events in a batch will either succeed or fail" sounds true to me. Can we reword?
```suggestion
- All events within a request batch are committed, or none are. Note that this does not mean that
  all of the events in a batch will succeed, or that all will fail. Events succeed or fail independently
  unless they are explicitly [linked](./client-requests.md#linked-events)
```

docs/design/data-modeling.md
```diff
@@ -49,73 +66,93 @@ ledger might look something like this:
...
 - 80 bits of randomness (low-order bits)
 
-**Important**: When creating multiple objects during the same millisecond, increment the random bytes
-(instead of generating new random bytes). Make sure to also store random bytes first, timestamp bytes 
-second, and both in little-endian. These details ensure that a sequence of objects have strictly 
-increasing ids according to the server. (Such ids are amenable to LSM optimizations, 
-leading to higher database throughput).
+When creating multiple objects during the same millisecond, we increment the random bytes rather
+than generating new random bytes. Furthermore, it is important that IDs are stored in little-endian
+with the random bytes as the lower-order bits and the timestamp as the higher-order bits.  These
```
sentientwaffle:
```suggestion
with the random bytes as the lower-order bits and the timestamp as the higher-order bits. These
```

docs/design/two-phase-transfers.md
```diff
@@ -7,42 +7,45 @@ sidebar_position: 2
 A two-phase transfer moves funds in stages:
 
 1. Reserve funds ([pending](#reserve-funds-pending-transfer))
-2. Resolve funds ([post](#post-pending-transfer), [void](#void-pending-transfer), or [timeout](#timeout))
+2. Resolve funds ([post](#post-pending-transfer), [void](#void-pending-transfer), or
+   [timeout](#timeout))
```
sentientwaffle: Let's call the last one "expire"/"expire-pending-transfer" instead of "timeout", to disambiguate from timeout-the-field/noun. That will also match how we refer to it in the code.

sentientwaffle (dismissed):

docs/design/client-requests.md
```diff
@@ -7,16 +7,60 @@ sidebar_position: 3
...
+| Transfer | Index within batch | flags.linked |
+|----------|--------------------|--------------|
+| `A`      | `0`                | `false`      |
+| `B`      | `1`                | `true`       |
+| `C`      | `2`                | `true`       |
+| `D`      | `3`                | `false`      |
+| `E`      | `4`                | `false`      |
+
+If any of transfers `B`, `C`, or `D` fail (for example, due to
+[`exceeds_credits`](../reference/operations/create_transfers.md#exceeds_credits), then `B`, `C`, and `D` will
```
batiati: Missing the example's closing parentheses:

```suggestion
If any of transfers `B`, `C`, or `D` fail (for example, due to
[`exceeds_credits`](../reference/operations/create_transfers.md#exceeds_credits)), then `B`, `C`, and `D` will
```

sentientwaffle (approved):

docs/design/two-phase-transfers.md
```diff
@@ -51,34 +54,36 @@ value of the pending transfer's field:
...
 
-### Timeout
+### Expire Pending Transfer
 
-A pending transfer may optionally be created with a
-[timeout](../reference/transfers.md#timeout). If the timeout interval passes
-before the transfer is either posted or voided, the full amount will be voided.
+A pending transfer may optionally be created with a [timeout](../reference/transfers.md#timeout). If
+the timeout interval passes before the transfer is either posted or voided, the transfer expires and
+the full amount is returned to the original account.
```
batiati, line 78: > full amount is returned to the original account.

I think it's more correct to say "the pending balance is removed" instead.

PR #1678, opened 2024-03-12
cli: add explicit `--standby` argument

matklad:
Before:

   * `--replica=1 --replica_count=2` creates an active replica with index=1
   * `--replica=2 --replica_count=2` creates a standby with index=2

Now:

   * `--replica=1 --replica_count=2` creates an active replica with index=1
   * `--standby=2 --replica_count=2` creates a standby with index=2
   * `--replica=2 --replica_count=2` is an error

That is, the semantics of the argument is the same, its replica index stored in the superblock, but the name must be `--replica` for active replicas and `--standby` for standbys.

This binds intention into the CLI and makes it more self explanatory and less error prone.

Additionally, we have an easier time "hiding" `--standby` feature until we need it for reconfig, backups, cdc, or canaries.

### DIFF
--- a/src/tigerbeetle/cli.zig
+++ b/src/tigerbeetle/cli.zig
@@ -1,3 +1,16 @@
+//! Parse and validate command-line arguments for the tigerbeetle binary.
+//!
+//! Everything that can be validated without reading the data file must be validated here.
+//! Caller must additionally assert validity of arguments as a defense in depth.
+//!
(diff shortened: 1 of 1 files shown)


matklad:
Well, let me expand this to a large experimental CLI idea.

EDIT: Done.

sentientwaffle (approved):

PR #1679, opened 2024-03-12
Ensure `-Dconfig=production` is set when `-Drelease` is...

cb22:
... ideally, we'd get rid of `-Dconfig=production` entirely - it's a footgun requiring two different unrelated flags to get a production build. Also add a warning to `main.zig` when started with it.

We could:
- Gate `constants.verify` can be on `-Drelease` and have its own flag to override
- Have `-Dconfig=production` be the default on `-Drelease`

I favor the first one - currently the only difference between the two configurations is verify and Direct IO being required...


matklad:
Java benchmarks are gone in https://github.com/tigerbeetle/tigerbeetle/pull/1669 --- they are dead code inherited since the time Java was in a separate repo. To benchark clients, we'll probably want to have a common IPC based harness. 

So let's maybe remove .Net bencharking code here as well, rather than update it?

matklad:
Not sure what's the general fix here: to me, `-Drelease` and `-Dconfig` do seem orthogonal. In particular, there's a strong expectation that the only difference between `-Drelease` and `-Drelease=false` is the level of optimiztion. When debugging, I'd love to be able to manipulate the level of debug info without affecting semantics of code otherwise. 


:thinking: ok, I think I understand what should be happening. I don't think this is about `-Drelease=true|false`, rather, it is about _what_ we are building. 

My understanding is that there's actually very little reason to build _tigerbeetle_ binary and client libraries with configs other then production? So maybe just build those always for production? Or maybe for production by default? 

Basically, I see this matrix:

| Artifact | Desired config |
|-------------|-------------|
|`tigerbeetle`| always prod (maybe toggle verify for scale tests)|  
|client libraries| always prod |
|aof| always prod |
|`simulator`| always test_min|
|`unit-tests`| always test_min|
|`fuzz-tests`| always fuzz_min|

feels like our current interface prioritize customizing config, while it's almost always clear which one to use (and, most of the time, there's really only one option)

matklad (approved):

PR #1681, opened 2024-03-12
vsr: add entire header to checkpoint state 

matklad:
To function, a replica needs a checkpoint and a header in WAL. A header
is required to make "head op is in journal" invariant work.

Currently after state sync a replica has only the checkpoint, with the
headers absent. We paper over this by transitioning the replica to
`.recovering_head`, but that creates availability issues, as in

    test "Cluster: sync: view-change with lagging replica in recovering_head"

To make the cluster more available, we include an entire header with the
checkpoint, and not just op/checksum pair.

This PR is just storage stability part of that change --- adding a header to the checkpoint state on disk and to the sync_checkpoint_target message, without any functional changes otherwise. 

In particular, the header _isn't_ used for determining the head op upon restart: doing that now will trip the invariant that DVC includes the head op.

One alternative design would be to ensure that view headers always include some header from the current checkpoint, but that would require updating the headers during checkpoint. It seems better to keep vsr and checkpoint headers separate and verify them against each other.

### DIFF
--- a/src/vsr/replica.zig
+++ b/src/vsr/replica.zig
@@ -3720,8 +3720,7 @@ pub fn ReplicaType(
                 commit_op_checkpoint_superblock_callback,
                 &self.superblock_context,
                 .{
-                    .commit_min_checksum = self.journal.header_with_op(vsr_state_commit_min).?.checksum,
-                    .commit_min = vsr_state_commit_min,
+                    .header = self.journal.header_with_op(vsr_state_commit_min).?.*,
                     .commit_max = self.commit_max,
                     .sync_op_min = vsr_state_sync_op.min,
                     .sync_op_max = vsr_state_sync_op.max,
--- a/src/vsr/replica.zig
+++ b/src/vsr/replica.zig
@@ -2613,10 +2613,15 @@ pub fn ReplicaType(
                 stage.target.checkpoint_id,
             });
 
-            self.sync_requesting_checkpoint_callback(std.mem.bytesAsValue(
+            assert(message.body().len == @sizeOf(vsr.CheckpointState));
+            const checkpoint_state = std.mem.bytesAsValue(
                 vsr.CheckpointState,
                 message.body()[0..@sizeOf(vsr.CheckpointState)],
-            ));
+            );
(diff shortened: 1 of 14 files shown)


src/vsr/superblock.zig
```diff
@@ -819,10 +812,10 @@ pub fn SuperBlockType(comptime Storage: type) type {
             update: UpdateCheckpoint,
         ) void {
             assert(superblock.opened);
-            assert(update.commit_min <= update.commit_max);
-            assert(update.commit_min > superblock.staging.vsr_state.checkpoint.commit_min);
-            assert(update.commit_min_checksum !=
-                superblock.staging.vsr_state.checkpoint.commit_min_checksum);
```
sentientwaffle, line 825: I would add

```
assert(update.header.valid_checksum());
assert(update.header.command == .prepare);
```

matklad: Good call, added this as well as `.command` check to the `on_sync_checkpoint`

src/vsr/replica.zig
```diff
@@ -2613,10 +2613,15 @@ pub fn ReplicaType(
...
-            self.sync_requesting_checkpoint_callback(std.mem.bytesAsValue(
+            assert(message.body().len == @sizeOf(vsr.CheckpointState));
+            const checkpoint_state = std.mem.bytesAsValue(
                 vsr.CheckpointState,
                 message.body()[0..@sizeOf(vsr.CheckpointState)],
-            ));
+            );
+            assert(checkpoint_state.header.invalid() == null);
+            assert(checkpoint_state.header.valid_checksum());
+            assert(checkpoint_state.header.command == .prepare);
```
sentientwaffle: I think we should verify the checksum before `.invalid()`? If the header is corrupt, then it will likely be invalid.
I would probably check `command` before `invalid` too, since `invalid` branches on `command`.

matklad: >I think we should verify the checksum before .invalid()

Ah, thanks, that's a very good rule of thumb indeed!

sentientwaffle (approved):

PR #1682, opened 2024-03-12
VSR: In-place upgrades

sentientwaffle:
To ensure storage determinism over state machine + LSM modifications, code upgrades are coordinated (via consensus) across the cluster. Multiple versions of the code will be compiled into a single `tigerbeetle` binary so that the replica itself can decide which version should be running, and deploy its own upgrades. This approach optimizes for safely (eliminating a large category of storage nondeterminism) and operational simplicity.

1. Replicas share which releases they are willing to upgrade into over `ping` messages.
2. When a `status=normal` primary detects a majority cluster that is willing to upgrade to a particular release, it triggers an upgrade...
3. Primary sends a `command=request operation=upgrade client=0` to itself, bearing the target release `UpgradeRequest { release: u16 }`.
4. When `operation=upgrade` commits, the primary sends itself another, until both:
  - `commit_min == op_checkpoint_next_trigger`, AND
  - All messages in the last bar are `operation=upgrade`.
5. At this point, the replica checkpoints, increasing its `CheckpointState.release` to the new value.
6. When the checkpoint completes, the replicas sees that it is running an older release (`self.release < self.superblock.working.vsr_state.checkpoint.release`) and attempts to execute the new release.

Other edge cases to keep in mind:
- If the above steps are interrupted by a view change: The first `commit_upgrade` sets `self.upgrade_release`, so that the new primary resumes sending the padding `operation=upgrade`s. (This is triggered by `on_upgrade_timeout()`.)
- If we send a padding upgrade request to ourself, then immediately drop it due to clock-not-synchronized, `on_upgrade_timeout` will keep trying.
- When a replica finishes state sync, it checks whether it needs to transition to a new release.

`ignore_request()` does not allow any request to queue behind an `operation=upgrade`:
- This simplifies the mental model + invariants of upgrades.
- This ensures that we will reach the "last bar is all checkpoints" soon after an upgrade is triggered.

Note that all `operation=upgrade` messages set `client=0`, which means they cannot displace any real clients' sessions from the sessions table.

Clients also have a release number, which they include in their message headers.
The replica has a minimum client release which they accept requests from. Messages from clients running a release which is less-than that minimum, or greater than the cluster's current release, will be rejected by a `command=eviction` message.


src/simulator.zig
```diff
@@ -666,36 +727,55 @@ pub const Simulator = struct {
...
-            if (!commit.request.header.operation.vsr_reserved()) {
-                simulator.requests_replied += 1;
-                simulator.workload.on_reply(
-                    commit.client_index,
-                    commit.reply.header.operation.cast(StateMachine),
-                    commit.reply.header.timestamp,
-                    commit.request.body(),
-                    commit.reply.body(),
-                );
+        while (!simulator.reply_sequence.empty()) {
```
sentientwaffle, line 730: The `Simulator`'s `ReplySequence` assembles replies (from the cluster to all clients) into commit order.
(Replies may arrive out of order, but the workload checker needs to process them in commit order so that the workload can correctly reproduce the cluster's sequence of events.)

Right now `ReplySequence` has a comprehensive view of replies: currently all `command=request`s originate from a client (`client_id ≠ 0`).

But upgrades originate from the primary itself (`client = 0`)! (Expiry messages for driving transfer timeouts will too.)

As such, this commit updates `ReplySequence` so that we will be able to skip past upgrade/expiry operations that the clients will never receive a reply for.

src/simulator.zig
```diff
@@ -167,6 +175,8 @@ pub fn main() !void {
         .replica_crash_stability = random.uintLessThan(u32, 1_000),
         .replica_restart_probability = 0.0002,
         .replica_restart_stability = random.uintLessThan(u32, 1_000),
+        .replica_release_advance_probability = 0.0001,
+        .replica_release_catchup_probability = 0.001,
```
sentientwaffle, line 179: These are separate because we want "probability that the cluster decides to upgrade" to be lower than "probability that a replica can catch up to an already-upgraded cluster".

src/vsr.zig
```diff
@@ -220,6 +220,8 @@ pub const Operation = enum(u8) {
     register = 2,
     /// The value 3 is reserved for reconfiguration request.
     reconfigure = 3,
+    /// The value 4 is is reserved for release-upgrade requests.
+    upgrade = 4,
```
sentientwaffle, line 224: I was tempted to name this "release" to keep up the "re*" trend, but "upgrade" is more precise and less overloaded. :stuck_out_tongue:

src/vsr/replica.zig
```diff
@@ -869,6 +896,8 @@ pub fn ReplicaType(
             release: u16,
             release_client_min: u16,
             releases_bundled: []const u16,
+            release_execute: *const fn (replica: *Self, release: u16) void,
+            test_context: ?*anyopaque,
```
sentientwaffle, line 900: Replicas will need to transition ("execute") to a different release from within `Replica.open()` (synchronously). For tests, this callback is implemented by `testing/cluster.zig`, and needs the `test_context` so that it can retrieve the `cluster` reference.

src/vsr/replica.zig
```diff
@@ -2969,6 +2997,72 @@ pub fn ReplicaType(
...
+                                break;
+                            }
+                        }
+                    }
+
+                    // A majority quorum (i.e. `max(quorum_commit, quorum_view_change)`) is required
+                    // to ensure that the upgraded cluster can both commit and view-change.
+                    if (release_replicas >= vsr.quorums(self.replica_count).majority) {
+                        release_target = release;
+                    }
```
sentientwaffle: Replicas advertise (in `ping` messages) the list of release versions their binary supports. When the primary sees that enough replicas have a newer version available, it kicks off the upgrade process.

How many is "enough"? The lower bound is `max(quorum_commit, quorum_view_change) == quorum_majority`, since we want the quorum in the new version to be able to both commit and change views.

I am tempted to set the threshold to `self.replica_count`. In practice, not upgrading all replicas together would probably be an operator error which leads to lagging replicas that then need to state sync.

...But I'm worried about the case of: A replica is broken/recovering, and the operator may _need_ to upgrade anyway. (e.g. to roll out a hotfix quickly)

src/vsr/replica.zig
```diff
@@ -3970,6 +4101,49 @@ pub fn ReplicaType(
...
+                prepare.body()[0..@sizeOf(vsr.UpgradeRequest)],
+            );
+            assert(request.release >= self.release);
+
+            if (request.release == self.release) {
+                // We are replaying this upgrade request after restarting into the new version.
+                assert(self.upgrade_release == null);
+            } else {
+                if (self.upgrade_release) |upgrade_release| {
+                    assert(upgrade_release == request.release);
```
sentientwaffle, line 4139: Once an "upgrade to release X" is committed, the replica cannot change its mind and decide to commit "upgrade to release Y" (for any Y≠X) until release X has checkpointed and the replica is running release X.

(We could ease this restriction later if necessary, but I think it is a helpful simplification for my mental model if nothing else.)

src/vsr/replica.zig
```diff
@@ -4482,14 +4711,14 @@ pub fn ReplicaType(
...
             assert(message.header.command == .request);
-            assert(message.header.client > 0);
             assert(message.header.view <= self.view); // See ignore_request_message_backup().
             assert(message.header.session == 0 or message.header.operation != .register);
             assert(message.header.request == 0 or message.header.operation != .register);
 
             if (self.client_sessions.get(message.header.client)) |entry| {
                 assert(entry.header.command == .reply);
                 assert(entry.header.client == message.header.client);
+                assert(entry.header.client != 0);
```
sentientwaffle, line 4732: `client=0` doesn't have a session.

src/vsr/replica.zig
```diff
@@ -4687,12 +4925,12 @@ pub fn ReplicaType(
             assert(self.primary());
 
             assert(message.header.command == .request);
-            assert(message.header.client > 0);
             assert(message.header.view <= self.view); // See ignore_request_message_backup().
 
             if (self.pipeline.queue.message_by_client(message.header.client)) |pipeline_message| {
                 assert(pipeline_message.header.command == .request or
                     pipeline_message.header.command == .prepare);
+                assert(message.header.client != 0);
```
sentientwaffle, line 4944: Since `client=0` has no session, the `pipeline.queue.message_by_client` case is already covered for client 0 in `ignore_request_message_duplicate()`.

src/vsr/replica.zig
```diff
@@ -8761,6 +9092,88 @@ pub fn ReplicaType(
...
+            maybe(self.pipeline.queue.message_by_operation(.upgrade) != null);
+
+            const upgrade = vsr.UpgradeRequest{ .release = self.upgrade_release.? };
+            self.send_request_to_self(.upgrade, std.mem.asBytes(&upgrade));
+
+            // If our clock is not synchronized, then the request-to-self might have been dropped.
+            maybe(self.pipeline.queue.message_by_operation(.upgrade) != null);
+        }
+
+        fn send_request_to_self(self: *Self, operation: vsr.Operation, body: []const u8) void {
```
sentientwaffle, line 9157: I wrote this thinking that the [pulse messages](https://github.com/tigerbeetle/tigerbeetle/pull/1657) might be able to use this same function, but it looks like pulses are injected in a very different way.

src/vsr/replica.zig
```diff
@@ -5260,8 +5498,6 @@ pub fn ReplicaType(
             assert(!self.pipeline.queue.prepare_queue.full());
             self.pipeline.queue.verify();
 
-            assert(!self.ignore_request_message(request.message));
```
sentientwaffle, line 5267: This assertion doesn't play nicely with upgrades. Consider:

1. Queue request X, operation=whatever
2. Queue request X+1, operation=upgrade
3. A commit finishes, pops request X from the queue and tries to prepare it...

But it hits `on_request: ignoring (upgrade queued)` here, even though the request precedes the upgrade so it should work.

src/vsr/replica.zig
matklad: Maybe `self.replica_count - 1`? In any case, I suggest writing 


```
vsr.quorums(self.replica_count).upgrade
```

even if that ends up being equal to majority.

src/vsr/replica.zig
```diff
@@ -9434,6 +9843,20 @@ const PipelineQueue = struct {
         return message;
     }
 
+    fn message_by_operation(pipeline: PipelineQueue, operation: vsr.Operation) ?*const Message {
```
matklad: This feels like an error-prone API, because it is not specified _which_ message is returned in case several match. 

Maybe `fn operation_pipelined(pipeline: PipelineQueue, operation: vsr.Operation) bool`?

Internally, it'll still compute the actual message and assert that we do have a message with the specifed op, but the call site won't be able to rely on impl detail of which message is returned.

src/vsr/replica.zig
```diff
@@ -8313,6 +8566,80 @@ pub fn ReplicaType(
...
+                    // preceding the checkpoint trigger consists of operation=upgrade.
+                    //
+                    // Otherwise we would risk the following:
+                    // 1. Execute op=X in the state machine on version v1.
+                    // 2. Upgrade, checkpoint, restart.
+                    // 3. Replay op=X when recovering from checkpoint on v2.
+                    // If v1 and v2 produce different results when executing op=X, then an assertion
+                    // will trip (as v2's reply doesn't match v1's in the client sessions).
+                    assert(found_upgrade < constants.lsm_batch_multiple);
+                    maybe(found_upgrade > 0);
```
matklad: To explain this maybe: if a cluster is idle, the op is between a checkpoint and its trigger, and the cluster receives an upgrade request, then:

* we want to append the request to log immediately
* but we want to pulse a checkpoint worth of upgrades up to the _next_ checkpoint, because the first critical bar of the current checkpoint is already poisoned by a non-upgrade request.

matklad: Wait, no something doesn't add up here? If there is an upgrade committed, anything on top of it should also be an upgrade? 

That is, across the _entire_ loop, `found_upgrade` could be greater than 0 but less than `lsm_batch_multiple`. However, if there's a non-upgrade op in this batch, it has to be the first op, no? 

Am I misunderstanding the big picture here?

src/vsr.zig
```diff
@@ -543,6 +545,15 @@ test "ReconfigurationRequest" {
...
 }
 
+pub const UpgradeRequest = extern struct {
+    release: u16,
+
+    comptime {
+        assert(@sizeOf(UpgradeRequest) == 2);
+        assert(stdx.no_padding(UpgradeRequest));
+    }
+};
```
matklad: Hm, won't this hit the same alignment issue as pings?

src/vsr/replica.zig
```diff
@@ -8173,6 +8419,13 @@ pub fn ReplicaType(
             ));
 
             self.commit_min = self.superblock.working.vsr_state.checkpoint.commit_min;
+            self.upgrade_release = null;
```
matklad: Maybe _not_ null this out? My understanding is that, when we hit this case, the following code is essentially `noreturn`. So, if we _don't_ null out here, we'll trip assertions if `release_transition` is buggy and doesn't actually reset the entire state!

src/vsr/replica.zig
```diff
@@ -8761,6 +9088,88 @@ pub fn ReplicaType(
...
+                .view = self.view,
+                .commit = self.commit_max,
+                .commit_checksum = latest_committed_entry,
+                .timestamp_monotonic = self.clock.monotonic(),
+                .checkpoint_op = self.superblock.working.vsr_state.checkpoint.commit_min,
+                .checkpoint_id = self.superblock.working.checkpoint_id(),
+            }));
+        }
+
+        fn send_request_upgrade_to_self(self: *Self) void {
```
matklad, line 9144: Do we guard anywhere that nobody else can send us an upgrade request? I think we can check that in `on_message_from_bus`?

sentientwaffle: We do check that `operation == upgrade` implies `client == 0`, but verifying in `on_message_from_bus` is a good idea!

matklad: Yeah, I was thinking about an evil client sending a message with `client = 0` "manually"

src/vsr/replica.zig
sentientwaffle: Ah, you're right, this should be
```
assert(found_upgrade == 0);
```

This is a relic from an earlier iteration which allowed non-upgrades to queue up behind the upgrades if no upgrade had actually committed yet.

sentientwaffle:
(Rebased atop main to resolve a few git conflicts.)

src/tigerbeetle/main.zig
```diff
@@ -325,6 +326,21 @@ const Command = struct {
     }
 };
 
+fn replica_release_execute(replica: *Replica, release: u16) void {
```
matklad:
```suggestion
fn replica_release_execute(replica: *Replica, release: u16) noreturn {
```

src/vsr/replica.zig
```diff
@@ -1339,7 +1380,9 @@ pub fn ReplicaType(
                     const releases_all = std.mem.bytesAsSlice(u16, message.body());
                     const releases = releases_all[0..message.header.release_count];
```
matklad, line 1381: should assert that releases _past_ `release_count` are all zeroes

src/vsr/replica.zig
```diff
@@ -3974,6 +4106,50 @@ pub fn ReplicaType(
...
+            assert(stdx.zeroed(&request.reserved));
+
+            if (request.release == self.release) {
+                // We are replaying this upgrade request after restarting into the new version.
+                assert(self.upgrade_release == null);
+            } else {
+                if (self.upgrade_release) |upgrade_release| {
+                    assert(upgrade_release == request.release);
+                } else {
+                    self.upgrade_release = request.release;
```
matklad, line 4141: Assert that anything that's pipelined at this point is an upgrade. 

The subtlty here is that, when a view change happens during upgrade, the new primary can't prepare anything until it commits backlog. 

And maybe assert in `PipelineQueue.verify` that nothing can be pielined past an upgrade?

src/vsr/replica_test.zig
```diff
@@ -1240,6 +1247,108 @@ test "Cluster: prepare beyond checkpoint trigger" {
...
+        t.replica(.R_).drop(.__, .bidirectional, .prepare_ok);
+        t.run();
+        t.run();
+        try expectEqual(t.replica(.R_).op_checkpoint(), 0);
+        try expectEqual(t.replica(.R_).release(), 1);
+
+        t.replica(.R_).pass(.__, .bidirectional, .prepare_ok);
+        t.run();
+        t.run();
+        t.run();
```
matklad: Do we need all three for the release to propagate?

src/simulator.zig
```diff
@@ -797,12 +890,21 @@ pub const Simulator = struct {
...
+                        simulator.replica_releases[replica.replica] <
+                        simulator.replica_releases_limit and
+                        (simulator.core.isSet(replica.replica) or chance_f64(
                         simulator.random,
-                        simulator.options.replica_restart_probability,
-                    )) {
-                        continue;
-                    }
+                        simulator.options.replica_release_catchup_probability,
+                    ));
```
matklad: Not the prettiest expression! 

Can't suggest any local improvements, perhaps its time to introuce `tick_crash_up` and `tick_crash_down` functions to save on indentation here

matklad (commented):
More or less finished review! Looks good to me module a couple of missing asserts!

src/vsr/replica_test.zig
sentientwaffle: Ah, oops! It used to need those b/c I originally set the `upgrade_timeout` much higher.

src/vsr/replica.zig
sentientwaffle: Ah.. earlier in the function I had the assert:

```
            if (self.pipeline == .queue) {
                assert(self.pipeline.queue.prepare_queue.count == 1);
                assert(self.pipeline.queue.request_queue.empty());
            }
```

But this is not actually correct... If we are a new primary (still in status=view_change) committing the ops in our pipeline, we may have multiple upgrades!

src/simulator.zig
sentientwaffle: Good idea. I also modified `tick_crash_down` so that we recompute `recoverable_count` before restarting a replica, rather than tracking it incrementally.

matklad:
zig fmt fails!

matklad (approved):

PR #1683, opened 2024-03-12
Remove the install scripts from the java client.

ddebree:
They are not used anywhere and appear to be left over from the git submodules.


matklad (approved):

matklad:
Thanks @ddebree!

PR #1686, opened 2024-03-13
vsr: fix liveness issue when recovering head replica can't state sync

matklad:
To recap, the issue:

* we can't persist checkpoint until our view is up to date, otherwise we
  observe a header break between the log and and checkpoint
* so we wait until our view_durable is up-to-date
* but that doesn't work for recovering head replicas, as it doesn't update its view durable
* and we _can't_ just excempt recevering head replica from this logic,
  as it might crash and restart into .normal due to transient failure

The new insight is that we can actually fix the latter, and make
`.recovering_head` case sticky across restarts! If a replica reastarts
and notices that it's checkpoint.header view is greater than view/view
durable, that necessary means that it crashed in recovering head state,
and needs to continue to recover its head!

### DIFF
--- a/src/vsr/replica.zig
+++ b/src/vsr/replica.zig
@@ -690,6 +690,7 @@ pub fn ReplicaType(
                 }
             } else {
                 if (self.log_view < self.superblock.working.vsr_state.checkpoint.header.view) {
+                    // During state sync, the replica installed CheckpointState from a future view.
                     self.transition_to_recovering_head();
                 } else {
                     // Even if op_head_certain() returns false, a DVC always has a certain head op.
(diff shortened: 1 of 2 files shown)


matklad:
This probably could benefit from more code commets, but I'll leave that for tomorrow (and also let this simulating overnight on my laptop)

matklad:
Actually, with the checkpoint header in the superblock, we can simply the whole thing to

* always accept checkpoint
* and just go to removering head if either checkpoints' op or checkpoint's view is in the future!

src/vsr/replica.zig
```diff
@@ -7384,16 +7392,21 @@ pub fn ReplicaType(
             assert(self.commit_stage == .idle);
```
sentientwaffle, line 7398: Can `transition_to_normal_*` and `transition_to_view_change` assert that `self.log_view >= self.superblock.working.vsr_state.checkpoint.header.view`?

src/vsr/replica.zig
```diff
@@ -4797,6 +4801,10 @@ pub fn ReplicaType(
                     }
 
                     if (self.status == .recovering_head) {
+                        if (message.header.view < self.superblock.working.vsr_state.checkpoint.header.view) {
```
sentientwaffle: Add a `log.debug("... ignoring (...)")` line. (The "ignoring" debug lines are some of my favorites for troubleshooting in the vopr :smile:).

src/vsr/replica.zig
```diff
@@ -7507,6 +7528,7 @@ pub fn ReplicaType(
             assert(self.journal.header_with_op(self.op) != null);
             assert(self.pipeline == .cache);
             assert(self.view_headers.command == .start_view);
+            defer assert(self.log_view >= self.superblock.working.vsr_state.checkpoint.header.view);
```
sentientwaffle, line 7531: :heart:

sentientwaffle (approved):

PR #1692, opened 2024-03-14
test: fix fragile tests

matklad:
When isolating replicas from each other, the test unintentionally breaks clock synchronization as well. The test currently pass, but slight variation might make it fail with b1 rejecting requests due to unsynchronized clocks.


matklad:
Notably, this fails in #1660

sentientwaffle (approved):

PR #1698, opened 2024-03-14
vopr: cluster_log: print max, rather than min release

matklad:


sentientwaffle (approved):
Ah right, I swapped the order on that list. Thanks!

PR #1700, opened 2024-03-14
VSR: SuperBlock reserved space; RegisterResult

sentientwaffle:
- Increase reserved bytes in `CheckpointState`
- Increase reserved bytes in `VSRState`
- Reserve space after each superblock copy
- Add `RegisterResult`

Is there anything else we should reserve space for in the superblock zone/copies?

### DIFF
--- a/src/vsr/replica.zig
+++ b/src/vsr/replica.zig
@@ -3982,7 +3982,7 @@ pub fn ReplicaType(
 
             const reply_body_size = switch (prepare.header.operation) {
                 .reserved, .root => unreachable,
-                .register => 0,
+                .register => self.commit_register(prepare, reply.buffer[@sizeOf(Header)..]),
                 .reconfigure => self.commit_reconfiguration(
                     prepare,
                     reply.buffer[@sizeOf(Header)..],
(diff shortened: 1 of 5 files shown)


matklad:
* we _might_ reserve space for extra superblock copies, but I don't think it's useful. I think if we change the number of copies, that'll only applly to new data files?
* we _might_ want to reserve space for extra superblock zones? Basically increasing `size_grid_padding` to be able to stuff something else between the superrblock and the grid? Though, I don't think that would be a good idea --- if we need an entire new zone, it'll probably be too large for a reasonable amount of padding we can reserve...

src/vsr/replica.zig
```diff
@@ -4081,6 +4081,28 @@ pub fn ReplicaType(
...
+            assert(prepare.header.size == @sizeOf(vsr.Header));
+            assert(prepare.header.op == self.commit_min + 1);
+            assert(prepare.header.op <= self.op);
+
+            const result = std.mem.bytesAsValue(
+                vsr.RegisterResult,
+                output_buffer[0..@sizeOf(vsr.RegisterResult)],
+            );
+
+            result.* = .{};
```
matklad, line 4102: Client should assert that the result is zeroed when receiving a register reply

matklad:
Let's lend this though to avoid conflicts though, superblock size changes are annoying!

matklad (approved):

src/vsr/replica.zig
sentientwaffle: :thinking: I actually left that out deliberately.. My thinking was that at some point we will change the register result to be non-zero, and we will want to do that without crashing old clients.

PR #1709, opened 2024-03-15
docs: emphasize that IDs are globally unique

emschwartz:
point to user_data for grouping,

### DIFF
--- a/docs/design/data-modeling.md
+++ b/docs/design/data-modeling.md
@@ -110,9 +110,11 @@ The primary purpose of an `id` is to serve as an "idempotency key" — to avoid
 executing an event twice. For example, if a client creates a transfer but the server's reply is
 lost, the client (or application) will retry — the database must not transfer the money twice.
 
-Note that `id`s are unique per cluster -- not per ledger. To store a relationship between `Account`s
-or `Transfer`s on different ledgers, you should store another identifier in one of the
-[`user_data`](#user_data) fields.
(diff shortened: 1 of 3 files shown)


docs/design/data-modeling.md
```diff
@@ -103,11 +103,17 @@ Example uses:
...
-[transfers](../reference/transfers.md#id)) is to serve as an "idempotency key" — to avoid
+The `id` field uniquely identifies each [`Account`](../reference/accounts.md#id) and
+[`Transfer`](../reference/transfers.md#id) within the cluster.
+
+The primary purpose of an `id` is to serve as an "idempotency key" — to avoid
 executing an event twice. For example, if a client creates a transfer but the server's reply is
 lost, the client (or application) will retry — the database must not transfer the money twice.
 
+Note that `id`s are unique per cluster -- not per ledger. To store a relationship between `Account`s
+or `Transfer`s on different ledgers, you should store another identifier in one of the
```
matklad: I think the `To store a relationship` would not make sense to the reader who haven't yet tried this. That is, this phrase says "X won't work, do Y", but it doesn't actually explain "X" (using the same id on two ledgers), so, unless you already know about X, this won't make much sense

matklad (approved):

matklad:
@emschwartz you might want to avoid @mentioning folks in commit messages --- git has this annoying bug that, every time a PR with a mention is rebased, the recepient gets pinged, and that this also works across the forks.

Mentions in PR descriptions are fine!

PR #1717, opened 2024-03-15
node: use id() function throughout examples

emschwartz:
Part of https://github.com/tigerbeetle/tigerbeetle/issues/1668

### DIFF
--- a/src/simulator.zig
+++ b/src/simulator.zig
@@ -600,7 +600,7 @@ pub const Simulator = struct {
 
         var missing_op: ?u64 = null;
         for (simulator.cluster.replicas) |replica| {
-            if (simulator.core.isSet(replica.replica)) {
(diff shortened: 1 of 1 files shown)


batiati (changes requested):
Docs are generated during CI using snippets extracted from the sample program code. We don't need change the markdown file manually.

You may want to look at `/src/scripts/client_readmes.zig` and `/src/clients/node/docs.zig`. 

To generate, run `./zig/zig build scripts -- ci --language=node`, if it returns `return error.DocsUpdated;` so your .md file was overwritten.

batiati:
Hi @emschwartz, please take a look at this example I did, you can add language-specific texts as needed.
https://github.com/tigerbeetle/tigerbeetle/compare/batiati-id-examples

But you'll need to add samples for all clients at once (since they all share the same script for generating the readme file).
Please rebase or reset your branch.

PR #1718, opened 2024-03-15
fix - the const keyword over var

gunnarahlberg:
fix - const instead of var

Changed the instances that could be const. These changes showed as warning using the zig version 0.12.0-dev.2330+9e684e8d1. It was discussed here https://ziggit.dev/t/error-local-variable-is-never-mutated/2238

This is my first PR


matklad (approved):
Thanks!

PR #1719, opened 2024-03-15
VSR: Don't send_request_upgrade_to_self while checkpointing upgrade

sentientwaffle:


matklad:
Assert in pipeline_prepare that we do not in fact prepare superfluous upgrades (if op is past the trigger and is an upgrade, then checkpoint op must not be upgrade)


And, I gusss, for good measure in commit as well? If we commit upgrade for the already active realse, the op must belong to first lsm batch multiple

sentientwaffle:
> Assert in pipeline_prepare that we do not in fact prepare superfluous upgrades (if op is past the trigger and is an upgrade, then checkpoint op must not be upgrade)

I asserted something similar, but not exactly that. I think the checkpoint op _may_ be an upgrade if we performed multiple upgrades in sequence.

matklad (dismissed):

sentientwaffle:
Rebased due to git conflict.
(Previous merge failed due to github actions issue.)

matklad (approved):

PR #1721, opened 2024-03-15
VSR: Bump protocol version to 1

sentientwaffle:
(Same reasoning as https://github.com/tigerbeetle/tigerbeetle/pull/1706)


sentientwaffle:
Nevermind, the reasoning from https://github.com/tigerbeetle/tigerbeetle/pull/1706 doesn't apply -- when we increment the protocol, we will still need to run + test both protocols, even in dev builds.

PR #1722, opened 2024-03-15
VSR: Release triple

sentientwaffle:
Use a release major/minor/patch triple.

The `Release` is `struct { value: u32 }`, and it can be cast to a `ReleaseTriple`. Reasoning:

- The most common operation with releases is comparison, so that should be simple. (Hence the `Release.value` backing integer.)
- But when logged, it should be formatted human-readably, e.g. `v1.2.3`. (Thus, a custom `format()` method.)

---

Also see https://github.com/tigerbeetle/tigerbeetle/commit/c7b3e0c396d4e33a414f668cdcb5184720bf89fe, which uses `-Dversion` to inject the release version into the binary. This is currently somewhat awkward/fragile: we inject the git tag as a human-readable version (for `tigerbeetle version`), but we don't actually want to use that version as the release number _unless_ it is a real production release build. Very open to suggestions for how to improve this (though possibly should be a follow-up).


sentientwaffle:
After more thought, what if we just had a special `-Dproduction` (or similar) flag that we pass to zig build only from `release.zig` when building a release. We could have a flag in each superblock header that says whether it should be used with production or development binaries. Then we can leave release version alone (rather than always using 0.0.1).

matklad:
>Then we can leave release version alone (rather than always using 0.0.1).

We still need _some_ release version, and I believe that source code should not try to mention a _real_ version (as it's hard to update). 

So, I'd say we should be using `0.0.1` as a dev version anyway, and it might double as "non-production" flag.

Though, while thinking about this, it might be good to stamp the superblock, on format, with _both_ superblock version and release version. Release version will always change, and it might help debugging the cases where we forgot to bump superblock version. That is, we should never _use_ release from the superblock, but we might find ourselves in a situation where that's useful.

matklad:
LGTM, but conflicts need resolving!

sentientwaffle:
> We still need some release version, and I believe that source code should not try to mention a real version (as it's hard to update).

Including it within the code would be tricky anyhow... it would have to be within the state machine, since different state machines would have different release versions. I think that we will need _some_ release in the state machine, if only to declare the minimum compatible client version.

> So, I'd say we should be using 0.0.1 as a dev version anyway, and it might double as "non-production" flag.

Suppose we are making a data format change (e.g. changing the format of a table block, or changing the statemachine protocol) and want to verify (during development) that the upgrade from release `X-1` to `X` handles it properly. How do we do that if `X-1 == X == 0.0.1`?

> Though, while thinking about this, it might be good to stamp the superblock, on format, with both superblock version and release version.

Good idea!

matklad (approved):

matklad:
>Suppose we are making a data format change (e.g. changing the format of a table block, or changing the statemachine protocol) and want to verify (during development) that the upgrade from release X-1 to X handles it properly. How do we do that if X-1 == X == 0.0.1?

If this is not an integration test (so, no actual `tigerbeetle` binary is involved) the code can set internal release version to whatever, pretty much how the simulator does that today. 

For integration tests, when we want to build two "real" `tigerbeetle` binaries, I guess we'll see once we get there! I'd imagine we'll be carefully building them in a special way, and just checking that they won't leak by accident. I suppose we would want to use "real" X and X+1 versions to test real data files?

PR #1723, opened 2024-03-15
Do not pulse when pipeline queue is full

batiati:
Crash

```
/home/batiati/tigerbeetle/src/vsr/replica.zig:9323:19: 0x471135 in send_request_pulse_to_self (simulator)
            assert(!self.pipeline.queue.full());
                  ^
/home/batiati/tigerbeetle/src/vsr/replica.zig:3055:44: 0x42603c in on_pulse_timeout (simulator)
                                           ^
/home/batiati/tigerbeetle/src/vsr/replica.zig:1257:66: 0x3e08f6 in tick (simulator)
            if (self.pulse_timeout.fired()) self.on_pulse_timeout();
```

SEED 7668991848053039090

### DIFF
--- a/src/vsr/replica.zig
+++ b/src/vsr/replica.zig
@@ -3043,8 +3043,8 @@ pub fn ReplicaType(
             assert(self.pulse_timeout.ticking);
 
             self.pulse_timeout.reset();
-            if (!self.pulse_needed()) return;
             if (self.pipeline.queue.full()) return;
+            if (!self.pulse_needed()) return;
 
             // To decide whether or not to `pulse` a time-dependant
             // operation, the State Machine needs an updated `prepare_timestamp`.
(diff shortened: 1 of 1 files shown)


src/vsr/replica.zig
```diff
@@ -3044,6 +3044,7 @@ pub fn ReplicaType(
 
             self.pulse_timeout.reset();
             if (!self.pulse_needed()) return;
+            if (self.pipeline.queue.full()) return;
```
sentientwaffle: If you swap the order of these two checks:
```suggestion
            if (self.pipeline.queue.full()) return;
            if (!self.pulse_needed()) return;
```

Then `pulse_needed` could assert that `!self.pipeline.queue.full()`. Which in turn "documents" the reason that this `full()` check is here, rather than inside `pulse_needed` -- `pulse_needed()`s other call sites are guaranteed to have a non-full pipeline.

sentientwaffle (approved):

PR #1724, opened 2024-03-16
Drops the "not for production use" notice.

batiati:
🥳


matklad:
Seems overdue, really: although we do have some residual availability bugs, I don't remeber us hitting a correctness bug for quite some time!

matklad (approved):

PR #1732, opened 2024-03-17
vopr: false positive during upgrades

matklad:
Normally, executing upgrade is a noreturn function. In the simulator, it
is not, and instead the replica is restarted on the next tick.

The problem with this is that there might be some callbacks scheduled on
the storage, which fire before the actual upgrade happens, and observe
an inconsistent state (as, aftre transitioning to a new release replica
doesn't bother to repair local invariants).

Cancel the storage immediately upon scheduling upgrades to prevent those
callbacks from firing.

Seed: 2843879175163466061

### DIFF
--- a/src/testing/cluster.zig
+++ b/src/testing/cluster.zig
@@ -542,6 +542,7 @@ pub fn ClusterType(comptime StateMachineType: anytype) type {
                 maybe(replica.release.value < release.value);
             }
 
+            cluster.storages[replica.replica].reset();
             cluster.replica_upgrades[replica.replica] = release;
         }
 


src/vsr/replica.zig
```diff
@@ -7764,7 +7764,7 @@ pub fn ReplicaType(
             self.send_prepare_oks_after_view_change();
         }
 
-        fn transition_to_recovering_head(self: *Self) void {
+        pub fn transition_to_recovering_head(self: *Self) void {
```
sentientwaffle: Was this included by mistake?

sentientwaffle:
Another possibility is to reset the replica's storage during `replica_release_execute_soon`.

```
            cluster.storages[replica.replica].reset();
```

(It will be reset when we crash the replica on the next tick anyways. The reason we don't just crash the replica on the same tick as the release transition is that it would require extra branching when we are transitioning during `Replica.open`, so this was simpler.)

matklad:
>Another possibility

Yup, that seems much better, thanks!

matklad:
Let's get that in to clear popular VOPR failures.

cb22 (approved):

PR #1733, opened 2024-03-17
Rename `get_account_history` to `get_account_balances`

batiati:


sentientwaffle (dismissed):
LGTM, though fwiw I think "account historical balances" would be more self-explanatory. (Albeit quite verbose.)

batiati:
> LGTM

Thanks!

>  though fwiw I think "account historical balances" would be more self-explanatory. (Albeit quite verbose.)

I feel the same, so I called it "historical balances" everywhere in the docs. However, I think `get_account_historical_balances` is too verbose for a function name! I also considered `get_historical_balances`, but it's nice to have both function names aligned under the same prefix (e.g `get_account_{transfers,balances}` ).

One thing that we could still change here is `Account.flags.history` to `Account.flags.historical_balances`, wdyt?

sentientwaffle:
> I feel the same, so I called it "historical balances" everywhere in the docs.

:heart: Yeah, this will help a lot!

> However, I think get_account_historical_balances is too verbose for a function name! I also considered get_historical_balances, but it's nice to have both function names aligned under the same prefix (e.g get_account_{transfers,balances} ).

I agree!

> One thing that we could still change here is Account.flags.history to Account.flags.historical_balances, wdyt?

Hmm, I could go either way on this; I don't have a preference.

sentientwaffle (approved):

PR #1734, opened 2024-03-17
Fixes a `scan_buffer_pool` leak

batiati:
If `get_account_balances` was called with an invalid account or an account without `flags.history == true`, then the `scan_buffer_pool` would not be reset for the next call with a valid filter.

It also writes unit tests that reproduce the bug.

### DIFF
--- a/src/state_machine.zig
+++ b/src/state_machine.zig
@@ -746,8 +746,11 @@ pub fn StateMachineType(
 
         fn prefetch_get_account_transfers(self: *StateMachine, filter: AccountFilter) void {
             assert(self.scan_result_count == 0);
+            assert(self.forest.scan_buffer_pool.scan_buffer_used == 0);
 
             if (self.get_scan_from_filter(filter)) |scan| {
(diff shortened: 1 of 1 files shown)


src/state_machine.zig
```diff
@@ -835,14 +843,14 @@ pub fn StateMachineType(
...
-                // TODO(batiati): Improve the way we do validations on the state machine.
-                log.info(
-                    "invalid filter for get_account_balances: {any}",
-                    .{filter},
-                );
             }
 
+            // TODO(batiati): Improve the way we do validations on the state machine.
+            log.info(
+                "invalid filter for get_account_balances: {any}",
```
sentientwaffle, line 850: I think this error message is not necessarily correct anymore. Unless querying an account that doesn't exist or doesn't set `history` qualifies as an "invalid filter"? But we can fix this later, seeing as this is just placeholder error handling anyhow.

sentientwaffle (approved):

PR #1736, opened 2024-03-18
vsr: commit_dispatch doesn't cancel cleanup

matklad:
commit_dispatch starts off as

    fn commit_dispatch(self: *Self, stage_new: CommitStage) void {
      assert(self.commit_min <= self.op);
      if (self.syncing == .canceling_commit) {
          return self.sync_cancel_commit_callback();
      }

That is, every commit stage is a cancellation point. Notably, this includes the `.cleanup` stage, which:

* releases the commit_prepare
* transitions to the next release

Not rleasing prepare is ok --- sync_dispatch would handle that in cancelling grid stage

But not doing release transition is bad, as we rely on our release and superblock release being in sync all the time.

To fix this, just make the cleanup stage of commit flow sync.

Seed: 14169780878022532988

### DIFF
--- a/src/vsr/replica.zig
+++ b/src/vsr/replica.zig
@@ -3366,10 +3366,9 @@ pub fn ReplicaType(
             assert(self.commit_min <= self.commit_max);
             assert(self.commit_min <= self.op);
 
-            if (self.syncing == .canceling_commit) {
+            if (self.syncing == .canceling_commit and stage_new != .cleanup) {
                 return self.sync_cancel_commit_callback();
             }
-            assert(self.syncing == .idle);
 
             const stage_old = self.commit_stage;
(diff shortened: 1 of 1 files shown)


matklad:
closes https://github.com/tigerbeetle/tigerbeetle/issues/1735

sentientwaffle (dismissed):

matklad:
wait, no assert on the very next line could trip

matklad:
PTAL @sentientwaffle!

(I origially missed that the assert fires with this very seed it attempts to fix because my first implementation just directly called `commit_op_cleanup` without going through dispatch, and I refactored it to just check for `.cleanup` explicitly at the last minute, without re-running the seed)

sentientwaffle (approved):

PR #1737, opened 2024-03-18
vsr: commit cancelation cleanup happens in commit_dispatch 

matklad:
Follow up to https://github.com/tigerbeetle/tigerbeetle/pull/1736

The current code where commit_stage is reset in `sync_cancel_grid_callback` looks very puzzling to me, feels _much_ more natural to do that immediately upon the cancellation. 

Am I missing the reason while we need the commit prepare to live while the grid is being canceled?

### DIFF
--- a/src/vsr/replica.zig
+++ b/src/vsr/replica.zig
@@ -3367,9 +3367,11 @@ pub fn ReplicaType(
             assert(self.commit_min <= self.op);
 
             if (self.syncing == .canceling_commit and stage_new != .cleanup) {
+                if (self.commit_prepare) |prepare| self.message_bus.unref(prepare);
+                self.commit_prepare = null;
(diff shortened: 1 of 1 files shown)


matklad:
Fails with seed `402fe5de1626ba12a6f7287dbd90b54ce006baca`, because `sync_cancel_grid_callback` observes non-idle commit stage, though I don't yet got it how that's possible.

sentientwaffle:
I think the issue is that `prefetch_state_machine` and `compact_state_machine` are not interruptible, so we jump directly to `canceling_grid`. (i.e., if we set them to `canceling_commit`, there's no guarantee that they would ever dispatch again, since they might be waiting for a block that will never arrive.)

matklad:
Ahh, that makes sense!

PR #1742, opened 2024-03-18
vsr: correct assert after #1736

matklad:


matklad:
Missed another assert! This _doesn't_ trigger if cleanup goes into `release_transition`, but totally trips for normal cleanups

sentientwaffle (approved):

PR #1743, opened 2024-03-18
vopr: false assert

matklad:
Even if replica.release == core_release_max, it might still be in the process of upgrading to a release originating from outside of core.


sentientwaffle (approved):
Ah, I had though that wouldn't be possible since during liveness mode the simulator is prompt about upgrading replicas to the highest version. But I guess there is a tick where it isn't quite upgraded yet.

PR #1749, opened 2024-03-18
VOPR: Use replica_releases_limit for release_max

sentientwaffle:
This is an alternate fix for https://github.com/tigerbeetle/tigerbeetle/issues/1740 which retains the assertion.

The original fix was https://github.com/tigerbeetle/tigerbeetle/pull/1743

### DIFF
--- a/src/simulator.zig
+++ b/src/simulator.zig
@@ -451,13 +451,14 @@ pub const Simulator = struct {
         }
 
         // Even though there are no client requests in progress, the cluster may be upgrading.
-        const release_max = simulator.core_release_max();
+        const release_max = releases[simulator.replica_releases_limit - 1].release;
         for (simulator.cluster.replicas) |*replica| {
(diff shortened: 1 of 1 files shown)


src/simulator.zig
```diff
@@ -451,13 +451,14 @@ pub const Simulator = struct {
         }
 
         // Even though there are no client requests in progress, the cluster may be upgrading.
-        const release_max = simulator.core_release_max();
```
matklad, line 454: Will this work? Given that core replicas don't restart normally, I think they aren't guaranteed to progress to `replica_releases_limit - 1`?

sentientwaffle: Ahh you're right, I forgot that `transition_to_liveness_mode` zeroes the `replica_release_advance_probability`!
So it is possible that 1/6 replicas (outside the core) has version 3, and all other replicas have version 2. That would get stuck in this case.

PR #1756, opened 2024-03-18
build: propagate version to clients

matklad:
Inject version information through an env var, rather than an argument.

This is super hacky, but makes sure that the version propagates to all clients.

### DIFF
--- a/src/scripts/release.zig
+++ b/src/scripts/release.zig
@@ -65,6 +65,8 @@ pub fn main(shell: *Shell, gpa: std.mem.Allocator, cli_args: CliArgs) !void {
     };
     log.info("version={s} sha={s}", .{ version_info.version, version_info.sha });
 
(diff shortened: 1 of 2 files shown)


src/scripts/release.zig
```diff
@@ -159,7 +161,6 @@ fn build_tigerbeetle(shell: *Shell, info: VersionInfo, dist_dir: std.fs.Dir) !vo
                 \\    -Dtarget={target}
                 \\    -Drelease={release}
                 \\    -Dgit-commit={commit}
-                \\    -Dversion={version}
             , .{
                 .target = target,
                 .release = if (debug) "false" else "true",
```
sentientwaffle, line 166: `version` (a few lines down) can be removed

matklad: Even must be, shell checks for unusead args!

matklad: removed

sentientwaffle (approved):

PR #1765, opened 2024-03-19
docs: getting started section

emschwartz:
Rename Quickstart to Getting Started, and move Installation page inside it


docs/getting-started/download.md
```diff
@@ -1,8 +1,30 @@
...
+
+### Windows
+
+```powershell
+git clone https://github.com/tigerbeetle/tigerbeetle && cd tigerbeetle && .\bootstrap.ps1
+```
+
+### Build from Source
+
+To build TigerBeetle from source, add `-build` as an argument to the bootstrap script.
```
batiati, line 27: It's worth mentioning here that the user may want to build from sources when the target CPU is too old and lacks some of the instructions the prebuilt binary requires (i.e. aes).

batiati (approved):

PR #1766, opened 2024-03-20
Reject invalid requests

batiati:
Implement StateMachine-aware validation for the requests, ignoring messages when the input length does not match the expected operation.

### DIFF
--- a/src/state_machine.zig
+++ b/src/state_machine.zig
@@ -552,7 +552,7 @@ pub fn StateMachineType(
                     comptime assert(event_size > 0);
                     comptime assert(batch_max > 0);
 
-                    // TODO(batiati) Clients do not validate batch size == 0,
+                    // Clients do not validate batch size == 0,
                     // and even the simulator can generate requests with no events.
                     maybe(input.len == 0);
 
--- a/src/state_machine.zig
+++ b/src/state_machine.zig
@@ -3495,54 +3495,58 @@ test "StateMachine: input_validate" {
     defer allocator.free(input);
 
     const Event = struct { operation: Operation, min: usize, max: usize, size: usize };
-    const events: []const Event = &.{
-        .{
-            .operation = .pulse,
-            .min = 0,
-            .max = 0,
-            .size = 0,
-        },
-        .{
-            .operation = .create_accounts,
-            .min = 0,
-            .max = @divExact(global_constants.message_body_size_max, @sizeOf(Account)),
-            .size = @sizeOf(Account),
-        },
-        .{
-            .operation = .create_transfers,
-            .min = 0,
-            .max = @divExact(global_constants.message_body_size_max, @sizeOf(Transfer)),
-            .size = @sizeOf(Transfer),
-        },
-        .{
-            .operation = .lookup_accounts,
-            .min = 0,
-            .max = @divExact(global_constants.message_body_size_max, @sizeOf(Account)),
-            .size = @sizeOf(u128),
-        },
-        .{
-            .operation = .lookup_transfers,
-            .min = 0,
-            .max = @divExact(global_constants.message_body_size_max, @sizeOf(Transfer)),
-            .size = @sizeOf(u128),
-        },
-        .{
-            .operation = .get_account_transfers,
-            .min = 1,
-            .max = 1,
-            .size = @sizeOf(AccountFilter),
-        },
-        .{
-            .operation = .get_account_balances,
-            .min = 1,
-            .max = 1,
-            .size = @sizeOf(AccountFilter),
-        },
+    const events = comptime events: {
+        var array: []const Event = &.{};
+        for (std.enums.values(Operation)) |operation| {
+            array = switch (operation) {
+                .pulse => array ++ [_]Event{.{
+                    .operation = operation,
+                    .min = 0,
+                    .max = 0,
+                    .size = 0,
+                }},
+                .create_accounts => array ++ [_]Event{.{
+                    .operation = operation,
+                    .min = 0,
+                    .max = @divExact(global_constants.message_body_size_max, @sizeOf(Account)),
+                    .size = @sizeOf(Account),
+                }},
+                .create_transfers => array ++ [_]Event{.{
+                    .operation = operation,
+                    .min = 0,
+                    .max = @divExact(global_constants.message_body_size_max, @sizeOf(Transfer)),
+                    .size = @sizeOf(Transfer),
+                }},
+                .lookup_accounts => array ++ [_]Event{.{
+                    .operation = operation,
+                    .min = 0,
+                    .max = @divExact(global_constants.message_body_size_max, @sizeOf(Account)),
+                    .size = @sizeOf(u128),
+                }},
+                .lookup_transfers => array ++ [_]Event{.{
+                    .operation = operation,
+                    .min = 0,
+                    .max = @divExact(global_constants.message_body_size_max, @sizeOf(Transfer)),
+                    .size = @sizeOf(u128),
+                }},
+                .get_account_transfers => array ++ [_]Event{.{
+                    .operation = operation,
+                    .min = 1,
+                    .max = 1,
+                    .size = @sizeOf(AccountFilter),
+                }},
+                .get_account_balances => array ++ [_]Event{.{
+                    .operation = operation,
+                    .min = 1,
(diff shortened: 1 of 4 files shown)


src/state_machine.zig
```diff
@@ -533,8 +533,38 @@ pub fn StateMachineType(
...
+                    if (input.len > batch_max * event_size) return false;
+                },
+            }
+
+            return true;
+        }
+
         /// Updates `prepare_timestamp` to the highest timestamp of the response.
-        pub fn prepare(self: *StateMachine, operation: Operation, input: []align(16) u8) void {
+        pub fn prepare(self: *StateMachine, operation: Operation, input: []align(16) const u8) void {
```
batiati, line 568: Is there a reason for it to be `input: []align(16) u8` instead of `input: []align(16) const u8` ?

src/state_machine.zig
```diff
@@ -533,8 +533,41 @@ pub fn StateMachineType(
...
+                    const batch_max = comptime constants.batch_max.operation_batch_max(
+                        comptime_operation,
+                    );
+
+                    comptime assert(event_size > 0);
+                    comptime assert(batch_max > 0);
+
+                    // TODO(batiati) Clients do not validate batch size == 0,
+                    // and even the simulator can generate requests with no events.
+                    maybe(input.len == 0);
```
batiati: Should we validate zero-sized `.create_{accounts,transfers}` and `lookup_{accounts,transfers}` ?

src/state_machine.zig
sentientwaffle: IIRC: Even though our state machine does not need to modify the input, it is in principle safe for the replica to do so. But other state machines might need/want to modify a request when it arrives.

src/state_machine.zig
```diff
@@ -3444,6 +3483,96 @@ test "get_account_balances: invalid filter" {
...
+        .{
+            .operation = .get_account_balances,
+            .min = 1,
+            .max = 1,
+            .size = @sizeOf(AccountFilter),
+        },
+    };
+
+    for (std.enums.values(Operation), events) |operation, event| {
+        assert(event.operation == operation);
```
sentientwaffle: Is the loop over `values(Operation)` to verify that all operations are covered?

src/vsr/replica.zig
```diff
@@ -4767,6 +4767,22 @@ pub fn ReplicaType(
...
+            if (StateMachine.operation_from_vsr(message.header.operation)) |operation| {
+                if (!StateMachine.input_valid(operation, message.body())) {
+                    log.debug(
+                        "{}: on_request: ignoring invalid request (operation={s}, input_len={})",
+                        .{
+                            self.replica,
+                            @tagName(operation),
+                            message.body().len,
+                        },
+                    );
```
sentientwaffle: A client that sent an invalid request would just endlessly retry. Maybe we could add a new eviction reason? Then at least the error surfaces. Likewise, this should probably be at least a `log.info` or maybe `log.warn`, since it is an exceptional case.

src/vsr/message_header.zig
```diff
@@ -167,6 +167,7 @@ pub const Header = extern struct {
         if (self.checksum_body_padding != 0) return "checksum_body_padding != 0";
         if (self.nonce_reserved != 0) return "nonce_reserved != 0";
         if (self.size < @sizeOf(Header)) return "size < @sizeOf(Header)";
+        if (self.size > constants.message_size_max) return "size > message_size_max";
```
sentientwaffle, line 170: :heart:

src/state_machine.zig
batiati: The way comptime "interfaces" work, each state machine can define whether it requires a mutable slice or not, without changing the contract with the replica.

src/state_machine.zig
```diff
@@ -533,8 +533,41 @@ pub fn StateMachineType(
             callback(self);
         }
 
+        pub fn input_valid(operation: Operation, input: []align(16) const u8) bool {
```
sentientwaffle, line 536: Could we implement this validation without increasing the surface area of the state machine? e.g.:

- Add `get_account_transfers`, `get_account_balances` to `batch_max` (hard coded to 1 each).
- Replica uses `constants.batch_max` and `@sizeOf(Event)` to determine valid sizes.

src/state_machine.zig
sentientwaffle: I'm leaning towards it not being an error, since if it was an error that would implicitly increase the surface area of the client interface. e.g. I'm imagining someone writing client code like this:

```
var accounts_to_lookup = []
for (whatever)  {
  if (something) accounts_to_lookup.append(account);
}
client.lookup_accounts(accounts)
```

which is simpler than
```
var accounts_to_lookup = []
for (whatever)  {
  if (something) accounts_to_lookup.append(account);
}
if (accounts.len > 0) client.lookup_accounts(accounts)
```

src/vsr/message_header.zig
matklad: I've was looking at this code today, and it seems that it might be beneficial to straight up assert here? 

checksum&header define framing, so they _must_ be validated by the caller. MessageBus validates them:

```
                    if (header.size < @sizeOf(Header) or header.size > constants.message_size_max) {
                        log.err("header with invalid size {d} received from peer {}", .{
                            header.size,
                            connection.peer,
                        });
                        connection.terminate(bus, .shutdown);
                        return null;
                    }
```

but this code isn't vopered, so it would be great to assert that on the replica's side for defense in depth.

src/state_machine.zig
batiati: I went back and forth deciding whether the validation should be done by the replica or the state machine.
My reasons why I implemented at the state machine level:
- We would need to split `batch_max` into `event_max` and `result_max` (which I think is a good change).
- The state machine may need to perform something more complex than simply validating the length.
- We can add `assert(input_valid(...))` everywhere in the state machine code :smile:.

src/vsr/replica.zig
batiati: I think it must behave similarly when the client sends an invalid operation.

- Let's log as `err` in both
- Should we evict the client in both cases??

```zig
            if (!message.header.operation.valid(StateMachine)) {
                // Some possible causes:
                // - client bug
                // - client memory corruption
                // - client/replica version mismatch
                log.err("{}: on_request: ignoring invalid operation (client={} operation={})", .{
                    self.replica,
                    message.header.client,
                    @intFromEnum(message.header.operation),
                });
                return true;
            }
```

sentientwaffle: Yes, I think we should evict the client in both cases, since this is a severe/unexpected issue (probably a bug) and the session is unrecoverable. I would use a different eviction `Reason` for each.

src/state_machine.zig
sentientwaffle: I thought I posted a response to this already but I can't find it, so apologies if this is duplicated...

> The state machine may need to perform something more complex than simply validating the length.

Good point. These are "hard" errors, though -- they terminate the session and the operator must restart the client.
Maybe that's ok though... in the future, `input_valid()` could return an error message, and we could include that error message in the `Eviction` body to bubble it back up to the client/operator.

(Currently error reporting is not very transparent -- the two error behaviors we have are "client retries endlessly but server keeps ignoring it" and "the server sends an eviction to the client and the client panics".)

src/vsr/replica.zig
```diff
@@ -4743,16 +4743,36 @@ pub fn ReplicaType(
                 return true;
             }
 
-            if (!message.header.operation.valid(StateMachine)) {
```
sentientwaffle, line 4746: I think we still need to check this first. `StateMachine.operation_from_vsr` will panic (by way of `vsr.Operation.to(StateMachine, operation)` -> `assert(operation.valid(StateMachine));`), so the `ignoring invalid operation` is unreachable.

src/vsr/message_header.zig
batiati: > it seems that it might be beneficial to straight up assert here?

Maybe it's too intrusive ... For example, AOF recovery and other tools may need to deal with malformed messages, so `header.invalid()` should not panic, and the `replica` instead calls `assert(header.invalid() == null)` when it needs to enforce the invariant.

sentientwaffle (approved):

PR #1768, opened 2024-03-20
bus: alignment of header is not assumed

### LINKED ISSUE #1518: MessageBus should verify message aligment
https://github.com/tigerbeetle/tigerbeetle/blob/78be5cb1ab0e48a0d1a791ba922b79a8e8396469/src/message_bus.zig#L712

As far as I can tell, this `@alignCast` can fire: we never check that `header.size` is a multiple of `@sizeOf(u256)`.

matklad:
This `@alignCast` could fire. There are several options to fix it:

1. Require that message size is a multiple of header's alignment.
2. At the transport layer, artificially pad messages to alignment.
3. Copy the data to fix the alignment.

I think 3 makes most sense:

1 doesn't really work, as state_machine replies are not aligned. A replica is a sequence of (u32 index, u32 error code) tuples, and they are 8-byte and not 16-byte aligned.

2 feels hacky in general and changes our network protocol. Additionally, there could be all kinds of encryption&compression layers in the transport, and it feels odd to try to maintain physical alignment through all that.

3 initially felt wrong to me, but I've grown to like it. The thing is, we _already_ have a memcpy here. If a replica gets two messages in a single `.read` call, the second one would be copied over to a new message anyway.

Although the current implementation copies the entire header for simplicity, we actually don't need to do that. Message bus uses header to calculate checksum and to look at a couple of fields (cluster, client, replica, version). It should be perfectly fine to do all that on an unaligned data.

Closes: #1518

### DIFF
--- a/src/message_bus.zig
+++ b/src/message_bus.zig
@@ -709,7 +709,12 @@ fn MessageBusType(comptime process_type: vsr.ProcessType) type {
                     return null;
                 }
 
-                const header: *const Header = @alignCast(mem.bytesAsValue(Header, data[0..@sizeOf(Header)]));
+                // If the message bus receives more than one message at a time, subsequent messages
+                // might not be aligned to 16. These messages would be copied over to a fresh
+                // `Message` anyway, fixing the alignment issue, but care must be taken to
+                // ensure header alignment before that.
+                var header: Header = undefined;
+                @memcpy(mem.asBytes(&header), data[0..@sizeOf(Header)]);
 
                 if (!connection.recv_checked_header) {
                     if (!header.valid_checksum()) {
@@ -739,7 +744,7 @@ fn MessageBusType(comptime process_type: vsr.ProcessType) type {
                         // This has the same effect as an asymmetric network where, for a short time
                         // bounded by the time it takes to ping, we can hear from a peer before we
                         // can send back to them.
-                        .replica => if (!connection.set_and_verify_peer(bus, header)) {
+                        .replica => if (!connection.set_and_verify_peer(bus, &header)) {
                             log.err(
                                 "message from unexpected peer: peer={} header={}",
                                 .{ connection.peer, header },


src/message_bus.zig
```diff
@@ -709,7 +709,18 @@ fn MessageBusType(comptime process_type: vsr.ProcessType) type {
...
                 }
 
-                const header: *const Header = @alignCast(mem.bytesAsValue(Header, data[0..@sizeOf(Header)]));
+                // If the message bus receives more than one message at a time, subsequent messages
+                // might not be aligned to 16. These messages would be copied over to a fresh
+                // `Message` anyway, fixing the alignment issue, but care must be taken when
+                // working with the header before that.
+                var header_aligned: Header = undefined;
+                const header: *const Header = // if (connection.recv_parsed % @alignOf(Header) == 0)
+                    if (false)
```
matklad: Will remove before the merge, but want to run our test-suite with the copying code to smoke test it

matklad:
By the way, that's how option 1 could look: https://github.com/tigerbeetle/tigerbeetle/commit/954aaf4a8f7d2cda862971f41bd6492f4c9852ae. Coded that up before reading the summary of the issue and learning that the approach doesn't work due to replies

src/message_bus.zig
```diff
@@ -709,7 +709,18 @@ fn MessageBusType(comptime process_type: vsr.ProcessType) type {
                     return null;
                 }
 
-                const header: *const Header = @alignCast(mem.bytesAsValue(Header, data[0..@sizeOf(Header)]));
+                // If the message bus receives more than one message at a time, subsequent messages
+                // might not be aligned to 16. These messages would be copied over to a fresh
+                // `Message` anyway, fixing the alignment issue, but care must be taken when
+                // working with the header before that.
+                var header_aligned: Header = undefined;
+                const header: *const Header = // if (connection.recv_parsed % @alignOf(Header) == 0)
```
sentientwaffle: I wonder if we should just copy it onto the stack header unconditionally... i.e. the cost of the branch even in the "happy case" might be more than the time added by the copy. Probably not measurable difference either way, though.

matklad: Yeah, good call, let's keep it simple. If we are optimizing this, we really should do the "checksum unalinged data" approach, not much sense to implement half-optimisation.

sentientwaffle (approved):

PR #1773, opened 2024-03-20
lsm: remove compaction interface type

matklad:
Use (tree_id, level_b) index pair for "dynamic" dispatch instead

### DIFF
--- a/src/lsm/forest.zig
+++ b/src/lsm/forest.zig
@@ -16,7 +16,7 @@ const allocate_block = @import("../vsr/grid.zig").allocate_block;
 const NodePool = @import("node_pool.zig").NodePool(constants.lsm_manifest_node_size, 16);
 const ManifestLogType = @import("manifest_log.zig").ManifestLogType;
 const ScanBufferPool = @import("scan_buffer.zig").ScanBufferPool;
-const CompactionInterfaceType = @import("compaction.zig").CompactionInterfaceType;
+const CompactionInfo = @import("compaction.zig").CompactionInfo;
 const CompactionHelperType = @import("compaction.zig").CompactionHelperType;
 const BlipStage = @import("compaction.zig").BlipStage;
 const Exhausted = @import("compaction.zig").Exhausted;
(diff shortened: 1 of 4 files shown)


cb22:
@matklad not sure if you saw the failure here - the fuzz needs to be changed too.

src/lsm/forest.zig
```diff
@@ -921,6 +956,9 @@ fn CompactionPipelineType(comptime Forest: type, comptime Grid: type) type {
             const first_beat = compaction_beat == 0;
             const half_beat = compaction_beat == @divExact(constants.lsm_batch_multiple, 2);
 
+            if (self.forest == null) self.forest = forest;
```
matklad, line 959: Forest does `bar_apply_to_manifest` from `compact_callback`, so I figured it'd be simpler to just always have forest set. We still have `callback` to assert the proper lifecycle. 

Though, I do not like re-entrancy of `Forest.compact_callback`, would be good to fix that, but out of scope for this PR.

src/lsm/forest.zig
```diff
@@ -1060,35 +1102,18 @@ fn CompactionPipelineType(comptime Forest: type, comptime Grid: type) type {
...
-            self.forest = null;
 
             callback(forest);
         }
 
         // TODO: It would be great to get rid of *anyopaque here. Batiati's scan approach
         // wouldn't compile for some reason.
         fn blip_callback(
-            compaction_interface_opaque: *anyopaque,
+            slot_opaque: *anyopaque,
```
matklad, line 1112: Looks like just passing a pointer to a slot works here?

src/testing/cluster/manifest_checker.zig
```diff
@@ -54,15 +54,16 @@ pub fn ManifestCheckerType(comptime Forest: type) type {
             for (0..constants.lsm_levels) |level| {
                 checksum_stream.add(std.mem.asBytes(&level));
 
-                inline for (Forest.tree_id_range.min..Forest.tree_id_range.max + 1) |tree_id| {
+                inline for (Forest.tree_id_range.min..Forest.tree_id_range.max + 1) |tree_id_u16| {
```
matklad, line 57: Intentionally avoiding `enum.values` here such that the verification code doesn't rely on the order of filds in an enum, and does a stupid numeric loop.

matklad:
ready for review @cb22 !

src/lsm/forest.zig
cb22: > Though, I do not like re-entrancy of Forest.compact_callback, would be good to fix that, but out of scope for this PR.

Yes - that was originally to get things working with the existing manifest compaction without having to touch that code too!

src/lsm/forest.zig
```diff
@@ -1268,11 +1332,26 @@ fn CompactionPipelineType(comptime Forest: type, comptime Grid: type) type {
 
                 // TODO: CompactionInterface internally stores a pointer to the real
```
cb22: Dead comment here

cb22 (commented):
LGTM - just a dangling old comment!

src/lsm/forest.zig
matklad: Thanks, removed!

cb22 (approved):

PR #1774, opened 2024-03-21
[REPL] Implement `get_account_transfers` and `get_account_balances`

batiati:


src/repl.zig
```diff
@@ -387,10 +407,16 @@ pub const Parser = struct {
             parser.offset = after_whitespace;
             try parser.print_current_position();
             try parser.printer.print_error(
-                \\Operation must be help, create_accounts, lookup_accounts,
-                \\create_transfers, or lookup_transfers. Got: '{s}'.
-                \\
-            ,
+                "Operation must be " ++
```
matklad, line 410:
```suggestion
                "Operation must be one of " ++
```

we are dropping `or`, so let's add `one of` so that it still reads naturally. Or maybe do

```
|operation, index|

const separator = if (index == 0) " " else if (index == Operation.len - 1) "or " else ","
```

src/repl.zig
```diff
@@ -832,9 +862,41 @@ pub fn ReplType(comptime MessageBus: type) type {
                         }
                     }
                 },
+                .get_account_transfers => {
+                    const get_account_transfers_results = std.mem.bytesAsSlice(
+                        tb.Transfer,
+                        result,
+                    );
+
+                    if (get_account_transfers_results.len == 0) {
```
matklad, line 873: Hm, but we _should_ get no results if the account wasn't transacted against?

batiati: likewise in lookup_accounts and lookup_transfers! I'm going to drop the word "Failed" then.

src/repl.zig
batiati: Using [Oxford commas](https://qph.cf2.quoracdn.net/main-qimg-790a26b47136cf4e21994999e1e32b95.webp) :smile_cat:

src/repl.zig
```diff
@@ -410,10 +410,12 @@ pub const Parser = struct {
...
                     var names: []const u8 = "";
-                    for (std.enums.values(Operation)) |operation| {
+                    for (std.enums.values(Operation), 0..) |operation, index| {
                         if (operation == .none) continue;
-                        if (names.len > 0) names = names ++ ", ";
-                        names = names ++ @tagName(operation);
+                        names = names ++
+                            (if (names.len > 0) ", " else "") ++
+                            (if (index == std.enums.values(Operation).len - 1) "or " else "") ++
+                            @tagName(operation);
```
matklad, line 418: What a neat way to write this!

matklad (approved):

PR #1775, opened 2024-03-21
lsm: CompactionPipeline extracted to a dedicated function

matklad:
This saves a level of indent, and separates meta-programming heavy
Forest code from the actual pipelining algorithm.

The only non-trivial motion here is that "Setup loop" code is moved from
Forest.compact into Pipeline.beat, but that actually feels more logical,
as the pipeline code already has first-beat setup.


matklad:
Backstory: tidy complained about line lengths in #1773, so I tried to find a way to make code less indented overall, and seems like moving pipeline stuff to a separate section makes sense overall? Not only due to indent, but also because it's just helpful to not have 500 lines of pipelines through the middle of your forest!

cb22 (approved):

PR #1785, opened 2024-03-22
vsr: assert that SV doesn't go past op_repiar_min

matklad:
We require that the primary always includes as much headers as fit into an SV. That is, the header can be omitted _only_ if we get all the way to op=0.

But primary doesn't have literally _all_ messages, its log cuts off at op_repair_min.

This all works out: if primary is at a checkpoint, it has at least lsm_batch_multiple headers on top of it, and that's larger than `view_change_headers_suffix_max`. But we should assert this!

### DIFF
--- a/src/constants.zig
+++ b/src/constants.zig
@@ -252,7 +252,7 @@ comptime {
 }
 
 /// Maximum number of headers from the WAL suffix to include in an SV message.
-/// Must at least cover the full pipeline.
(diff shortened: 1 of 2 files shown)


matklad:
Status: still need to push over the finish line, but that's not a high priority as it only affects the test config.

matklad:
Ok, apparently it's not important enough!

PR #1788, opened 2024-03-22
Client: cleanup Demuxer wrapper

kprotty:
The vsr.Client wrapper simply took/gave bytes over types. So instead the StateMachine's Demuxer API is changed and vsr.Client simply re-exports it. 

Additionally, Demuxer takes a mutable slice to fixup `Create{T}Result.index` but vsr.Client returned an immutable slice. So Client was also changed to return a mutable slice instead of Demuxer having to `@constCast`.

### DIFF
--- a/src/vsr/client.zig
+++ b/src/vsr/client.zig
@@ -440,6 +440,9 @@ pub fn Client(comptime StateMachine_: type, comptime MessageBus: type) type {
                 }
             } else {
                 // The message is the result of raw_request(), so invoke the user callback.
+                // NOTE: the callback is allowed to mutate `reply.body()` so ensure the message has
+                // no other references or uses before this callback.
+                assert(reply.references == 1);
                 inflight.callback(
                     inflight.user_data,
                     inflight_vsr_operation.cast(StateMachine),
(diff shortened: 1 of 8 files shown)


src/vsr/client.zig
```diff
@@ -21,45 +21,19 @@ pub fn Client(comptime StateMachine_: type, comptime MessageBus: type) type {
...
 
         pub const StateMachine = StateMachine_;
-
+        pub const DemuxerType = StateMachine.DemuxerType;
         pub const Request = struct {
             pub const Callback = *const fn (
                 user_data: u128,
                 operation: StateMachine.Operation,
-                results: []const u8,
+                results: []u8,
```
sentientwaffle, line 29: In `vsr.Client.on_reply`, can we assert `reply.references == 1` just before 

```
                // The message is the result of raw_request(), so invoke the user callback.
                inflight.callback(
                    inflight.user_data,
                    inflight_vsr_operation.cast(StateMachine),
                    reply.body(),
                );
```

? I'm not sure if it will work... there might be an extra reference in the message bus, but it is worth a try.

And either way, I would add a comment to that callback invocation warning that the callback is allowed to mutate the reply body in-place.

kprotty: To guarantee no other references, should vsr copy the reply body to its own internal buffer and pass that to the callback?

sentientwaffle: No, I was just wondering if we already had a guarantee of no other references. I don't think it is worth the overhead of copying it.

sentientwaffle (approved):

PR #1794, opened 2024-03-25
release 2024-03-25 (placeholder -- not a real release)

sentientwaffle:


matklad (approved):
No song for us until we do an actual release seems fair!

sentientwaffle:
Yeah, that was what I was thinking.

PR #1796, opened 2024-03-25
lsm: don't duplicate the entire CompactionInfo 

matklad:
CompactionInfo is stored by value by both pipeline slots and
compactions.

That is confusing! Given that the slot already has an index, it can use
that to fetch the info. But that's verbose, so let's duplicate _just_
compaction coordinates.


matklad:
rebased&ready for review @cb22 !

cb22 (approved):
LGTM!

PR #1797, opened 2024-03-26
compaction: use FIFO as a block_pool instead of trying to slice arrays

cb22:
Previously, we tried to slice up an array of compaction blocks and hand them out. Due to an error in the calculation of block_count_bar, not enough blocks were reserved, so beat scoped blocks could alias higher bar scoped blocks.

This was caught by asserts, when it did eventually happen, but rather than just fixing the math let's make it much harder for the situation to arise.

Instead of an array, blocks are now popped off a simple FIFO, and pushed back on when they're done. Additional asserts have been added, ensuring values match what we expect, and all the setting of state (eg, `block.stage = .free`, `block.next = null`) have been removed and is instead replaced with a proper block lifecycle.

This also (as a side effect) improves memory efficiency. We now use all of the available blocks per beat, rather than reserving an amount for bar work which isn't running.

### DIFF
--- a/src/lsm/compaction.zig
+++ b/src/lsm/compaction.zig
@@ -93,9 +93,10 @@ pub const BlipStage = enum { read, merge, write, drained };
 pub fn CompactionHelperType(comptime Grid: type) type {
     return struct {
         pub const CompactionBlocks = struct {
-            /// Index blocks are global, and shared between blips. The index reads happen
-            /// as a mini-stage before reads kick off.
-            source_index_blocks: []CompactionBlock,
+            // Index blocks are global, and shared between blips. The index reads happen
+            // as a mini-stage before reads kick off.
+            source_index_block_a: *CompactionBlock,
+            source_index_block_b: *CompactionBlock,
 
             /// For each source level, we have a buffer of CompactionBlocks.
             source_value_blocks: [2]BlockFIFO,
@@ -119,26 +120,28 @@ pub fn CompactionHelperType(comptime Grid: type) type {
             next: ?*CompactionBlock = null,
         };
 
-        pub const BlockFIFO = struct {
-            const Inner = FIFO(CompactionBlock);
+        pub const CompactionBlockFIFO = FIFO(CompactionBlock);
 
-            free: Inner,
-            pending: Inner,
-            ready: Inner,
-            ioing: Inner,
+        pub const BlockFIFO = struct {
+            free: CompactionBlockFIFO,
+            pending: CompactionBlockFIFO,
+            ready: CompactionBlockFIFO,
+            ioing: CompactionBlockFIFO,
 
             count: usize,
 
             /// All blocks start in free.
-            pub fn init(blocks: []CompactionBlock) BlockFIFO {
-                assert(blocks.len % 2 == 0);
+            pub fn init(block_pool: *CompactionBlockFIFO, count: usize) BlockFIFO {
+                assert(count % 2 == 0);
+                assert(block_pool.count >= count);
 
-                var free: Inner = .{
+                var free: CompactionBlockFIFO = .{
                     .name = "free",
                     .verify_push = false,
                 };
 
-                for (blocks) |*block| {
+                for (0..count) |_| {
+                    const block = block_pool.pop().?;
                     assert(block.stage == .free);
                     free.push(block);
                 }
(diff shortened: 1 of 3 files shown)


src/lsm/forest.zig
```diff
@@ -1002,49 +1010,51 @@ fn CompactionPipelineType(comptime Forest: type, comptime Grid: type) type {
...
-                    // the same.
-                    const len = compaction_blocks.len; // Line length limits...
-                    const immutable_table_a_block = compaction_blocks[len - 1].block;
-                    const blocks = compaction_blocks[len - 3 .. len - 1];
-                    for (blocks) |*block| {
-                        block.next = null;
-                        block.stage = .free;
-                    }
-
-                    compaction_blocks.len -= 3;
```
cb22, line 1025: @matklad RIP :(

src/lsm/compaction.zig
```diff
@@ -148,10 +151,21 @@ pub fn CompactionHelperType(comptime Grid: type) type {
...
                     .ready = .{ .name = "ready", .verify_push = false },
                     .ioing = .{ .name = "ioing", .verify_push = false },
-                    .count = blocks.len,
+                    .count = count,
                 };
             }
 
+            /// Return all the blocks to the pool they came from.
+            pub fn deinit(self: *BlockFIFO, block_pool: *CompactionBlockFIFO) void {
+                assert(self.count == self.free.count);
```
sentientwaffle: Can we also assert that 

```
            assert(self.pending.count == 0);
            assert(self.ready.count == 0);
            assert(self.ioing.count == 0);
```

(As an aside, why is `ioing` not called `writing`?)

src/lsm/compaction.zig
```diff
@@ -842,18 +853,16 @@ pub fn CompactionType(
 
             // TODO: We only support 2 index blocks (1 for table a, 1 for table b) and don't
             // support spanning value reads across them for now.
-            assert(blocks.source_index_blocks.len == 2);
-
-            // TODO: index_block_a will always point to source_index_blocks[0] (even though if our
+            // TODO: index_block_a will always point to source_index_block_a (even though if our
             // source is immutable this isn't needed! Future optimization)...
             // index_block_b will be the index block of the table we're currently merging with.
             var read_target: usize = 0;
```
sentientwaffle: This variable can be removed now.

src/lsm/compaction.zig
```diff
@@ -1776,6 +1787,34 @@ pub fn CompactionType(
             beat.deactivate_assert_and_callback(.write, null);
         }
 
+        /// Return our blocks to the block pool.
+        pub fn beat_blocks_release(
```
sentientwaffle: This name confused me at first; I was expecting it to be responsible for `grid.release()`.
What about `beat_blocks_unassign`?

Likewise for `bar_blocks_release()`.

Or maybe `beat_teardown`/`bar_teardown`, and rename `beat_blocks_assign` to `beat_setup`? That would match better with `bar_setup`.

src/lsm/forest.zig
```diff
@@ -797,36 +851,57 @@ fn CompactionPipelineType(comptime Forest: type, comptime Grid: type) type {
...
         pub fn reset(self: *CompactionPipeline) void {
+            var block_pool: CompactionBlockFIFO = .{
+                .name = "block_pool",
+                .verify_push = false,
+            };
+
+            for (self.block_pool_raw) |*compaction_block| {
+                compaction_block.stage = .free;
+                compaction_block.next = null;
+
```
sentientwaffle:
```suggestion
                compaction_block.* = .{ .block = compaction_block.block };
```

src/lsm/forest.zig
```diff
@@ -838,111 +913,44 @@ fn CompactionPipelineType(comptime Forest: type, comptime Grid: type) type {
...
-                "source_value_level_a.len={} source_value_level_b.len={} " ++
-                " target_value_blocks.len={}", .{
-                block_count_bar,
-                source_index_blocks.len,
-                source_value_level_a.len,
-                source_value_level_b.len,
-                target_value_blocks.len,
-            });
+            // -2 for the source_index_block_{a,b}.
+            var equal_split_count = @divFloor(self.block_pool.count - 2, 3);
```
sentientwaffle: Rather than needing this explicit `- 2` (and an explanation), you could pop them beforehand, e.g.:

```
const source_index_block_a = self.block_pool.pop().?;
const source_index_block_b = self.block_pool.pop().?;
```

src/lsm/compaction.zig
cb22: Ah yes, `release` is used in Grid terminology. `unassign` is good.

`bar_setup` is the odd one out; I was trying to go with more descriptive naming, but if you think it's clearer... (especially since some of the things have been overloaded, like `bar_setup_budget` sets the budget and some bar blocks)

sentientwaffle: I don't feel strongly about the bar/beat terminology corresponding. But I think the before/after terminology should match in each case, e.g. assign/unassign, setup/teardown.

src/lsm/compaction.zig
cb22: > (As an aside, why is ioing not called writing?)

It's currently only used by writing, but I think there was a use case (equally possible I made it up) where reading blocks would transition through the same state in future.

sentientwaffle (approved):

PR #1799, opened 2024-03-26
clients: move native library building to zig

matklad:


src/clients/dotnet/TigerBeetle/TigerBeetle.csproj
```diff
@@ -15,23 +15,18 @@
   <PropertyGroup>
     <OS Condition="$([MSBuild]::IsOSPlatform('Windows'))">Windows</OS>
   </PropertyGroup>
```
batiati: We can get rid of that too!!

```suggestion
```

src/clients/java/pom.xml
```diff
@@ -233,23 +233,6 @@
...
-            </goals>
-            <configuration>
-              <arguments>
-                <argument>build</argument>
-                <argument>java_client</argument>
-                <argument>-Drelease</argument>
-                <argument>-Dconfig=production</argument>
-              </arguments>
-            </configuration>
-          </execution>
```
batiati, line 252: Wouldn't it be better to remove this entire `<plugin>` section and run the JNI tests from the CI as well?

matklad: Ahh, that `test:jni` did look suspicious to me, but I failed to parse the whole thing as `zig build test:jni` invocation. XML is verbose!

batiati (dismissed):
Nice! The [HACKING.md](https://github.com/tigerbeetle/tigerbeetle/blob/main/docs/HACKING.md#hacking-on-clients) file already points to the CI script, so we don't need to change any instructions for building the clients!

batiati:
>To complicate things, some languages trigger the native build automatically, while others require manually running ./zig/zig build lang_client.

Although it could be changed here ....

batiati (approved):
> Although it could be changed here ....

Thanks, and sorry for the last-minute change :sweat_smile:!!!

matklad:
No worries, I can fixup&rebase&force-push at the speed of light!

matklad:
(and sometimes I even do that correctly!)

PR #1800, opened 2024-03-26
benchmark: allow setting seed on the CLI

cb22:
For now, it's still fixed. We might want to randomize this though with some future thought (or perhaps only combined with --verify in future).

Additionally, sets the account user_data_* fields to random values too.

### DIFF
--- a/src/tigerbeetle/benchmark_load.zig
+++ b/src/tigerbeetle/benchmark_load.zig
@@ -129,7 +129,11 @@ pub fn main(
         );
     }
 
-    var rng = std.rand.DefaultPrng.init(42);
+    // If no seed was given, use a default seed for reproducibility.
+    const seed: usize = cli_args.seed orelse 42;
+    log.info("Benchmark seed = {}", .{seed});
+
+    var rng = std.rand.DefaultPrng.init(seed);
     const random = rng.random();
     const account_id_permutation: IdPermutation = switch (cli_args.id_order) {
         .sequential => .{ .identity = {} },
--- a/src/tigerbeetle/benchmark_load.zig
+++ b/src/tigerbeetle/benchmark_load.zig
@@ -214,6 +214,8 @@ const Benchmark = struct {
             return;
         }
 
+        const random = b.rng.random();
+
         // Reset batch.
         b.batch_accounts.clearRetainingCapacity();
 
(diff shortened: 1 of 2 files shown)


matklad:
Hm, I'd say for benchmarks we do want the seed to be the same by default? It's easier to compare benchmarks if you know that the code is exactly the same. Opt-in randomization might be a useful thing though

cb22:
> Hm, I'd say for benchmarks we do want the seed to be the same by default? It's easier to compare benchmarks if you know that the code is exactly the same. Opt-in randomization might be a useful thing though

Yeah it's a valid point, I'm kinda blending this as part of making the benchmark do end-to-end verification, and there it's useful to have unique values each time - I might just swap the default here.

I've got another follow-up PR that actually allows you to disable the attempt at targeting a specific rate, to get even more deterministic batches. Currently, it's actually pretty random what happens in the warm-up phase of the benchmark.

As a counterpoint to the first point though, I would - at least glancing at the code - not expect the seed to affect the results that much (barring pending transfers - but that's disabled by default) since the shape of the data is still the same, ~but that's _not_ the case - I've got 2 seeds that have a ~50k TPS delta between them which is far more than expected.~ (maybe not...)

src/tigerbeetle/benchmark_load.zig
```diff
@@ -223,9 +233,9 @@ const Benchmark = struct {
         {
             b.batch_accounts.appendAssumeCapacity(.{
                 .id = b.account_id_permutation.encode(b.account_index + 1),
-                .user_data_128 = 0,
-                .user_data_64 = 0,
-                .user_data_32 = 0,
+                .user_data_128 = random.int(u128),
+                .user_data_64 = random.int(u64),
+                .user_data_32 = random.int(u32),
```
batiati, line 234: ❤️

src/tigerbeetle/benchmark_load.zig
```diff
@@ -129,7 +129,15 @@ pub fn main(
         );
     }
 
-    var rng = std.rand.DefaultPrng.init(42);
+    const seed: usize = cli_args.seed orelse seed: {
+        // If no seed was given, use a random seed instead.
+        var seed_random: u64 = undefined;
+        try std.os.getrandom(std.mem.asBytes(&seed_random));
+        break :seed seed_random;
+    };
```
batiati: Do you want to do it here or in another PR?

```suggestion
    // If no seed was given, use a default seed for reproducibility.
    const seed: usize = cli_args.seed orelse 42;
```

cb22: I'll do it here - I'll commit and rebase this once I'm finished digging into my strange results!

cb22: @batiati done!

batiati (approved):

PR #1801, opened 2024-03-26
build: remove TIGERBEETLE_RELEASE env var hack

matklad:
Overhaul version handling:

* `tigerbeetle --version` uses proper semver syntax of `x.y.z+sha`
  rather than `git describe` default format.
* default to version 0.0.1 for dev builds (the commit is still included
  as build info in semver syntax).
* add version with git commit to `constants`, so that there's no need to
  import build_config directly to get at it
* use `-D` option rather the an env var to thread release info

The latter became possible after #1799

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -65,11 +65,11 @@ pub fn build(b: *std.Build) !void {
         []const u8,
         "git-commit",
         "The git commit revision of the source code.",
-    ) orelse try shell.git_commit();
-    options.addOption([]const u8, "git_commit", git_commit);
+    ) orelse std.mem.trimRight(u8, b.exec(&.{ "git", "rev-parse", "--verify", "HEAD" }), "\n");
+    assert(git_commit.len == 40);
+    options.addOption(?[40]u8, "git_commit", git_commit[0..40].*);
 
     const release_version_string = b.env_map.get("TIGERBEETLE_RELEASE");
-    options.addOption([]const u8, "version", release_version_string orelse try shell.git_tag());
     options.addOption([]const u8, "release", release_version_string orelse "0.0.1");
 
     options.addOption(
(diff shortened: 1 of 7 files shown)


src/config.zig
```diff
@@ -18,6 +18,7 @@ const BuildOptions = struct {
     config_log_level: std.log.Level,
     tracer_backend: TracerBackend,
     hash_log_mode: HashLogMode,
+    git_commit: ?[40]u8,
```
matklad, line 21: Although we always pass a git commit in, I think it makes more sense to keep it optional at this layer, as, in general, code might be build outside of the repo.

build.zig
```diff
@@ -65,12 +65,15 @@ pub fn build(b: *std.Build) !void {
...
+    options.addOption(?[40]u8, "git_commit", git_commit[0..40].*);
 
-    const release_version_string = b.env_map.get("TIGERBEETLE_RELEASE");
-    options.addOption([]const u8, "version", release_version_string orelse try shell.git_tag());
-    options.addOption([]const u8, "release", release_version_string orelse "0.0.1");
+    options.addOption(
+        []const u8,
+        "release",
+        b.option([]const u8, "tigerbeetle-release", "Release triple.") orelse "0.0.1",
+    );
```
sentientwaffle, line 76: Aren't all of these options "tigerbeetle" options? Maybe we should just call this `"version"`, to avoid confusion with `-Drelease`/release-builds.

matklad: Yeah, I am undecided here. We use `release` internally, but `-Drelease` is already occupied by, well, `-Drelease`.

I settled on `tigerbeetle-release` because its a long ugly hyphen-word --- this should only be set by the release infra, so verbose is good I guess. 

No strong opinion either way.

sentientwaffle (approved):

build.zig
sentientwaffle: Just thought of another option: `-Dconfig-release` (to match `-Dconfig-log-level`, `-Dconfig-aof-record` etc).

matklad: Excellent! https://github.com/tigerbeetle/tigerbeetle/pull/1811

PR #1802, opened 2024-03-26
docs: About section

emschwartz:
Moves the content from FAQs.md, HISTORY.md, and DESIGN.md into pages in the About section.
Also moved Internals to be under About to reduce the number of items in the docs sidebar.

- **docs: break FAQs into separate pages in About section**
- **docs: rewrap about section to 100 chars**
- **docs: copy performance section of design doc into about section**
- **docs: move content from DESIGN.md into About section**
- **docs: move content from HISTORY.md into About section**
- **docs: move Tiger Style into About section**
- **docs: rewrap Tiger Style to 100 chars**
- **docs: link to About section from main README**
- **docs: move internals into About**
- **docs: reorder pages in About section**
- **docs: leave placeholder markdown files**


emschwartz:
Note that this PR doesn't add any new content, only moves content around and reformats it. I think it would be useful to add another page or a section on VSR, but I left that for another PR.

docs/TIGER_STYLE.md
sentientwaffle, line 1: Could we use a symlink instead? After https://github.com/tigerbeetle/tigerbeetle this is the most visited page of the repo, so "breaking" the link (even if it isn't really broken) isn't ideal.

docs/about/internals/_category_.json
sentientwaffle, line 1: If `internals` is nested within `about`, then the "position" in this file should be updated

emschwartz: Funny enough, the position happens to be the same 😅

docs/TIGER_STYLE.md
emschwartz: Great point!

sentientwaffle, line 1: TIL github doesn't actually render symlinks: https://github.com/tigerbeetle/tigerbeetle/blob/2d82ca902066cb87403d3d8558e9c23f364b0d63/docs/TIGER_STYLE.md.
Maybe swap around which file is the symlink, so that `docs/TIGER_STYLE.md` renders?

sentientwaffle (approved):

PR #1803, opened 2024-03-27
build: dev builds use superblock version 0

matklad:
We bumped this to 1 for the release, but what we actually want is this logic:

* release builds use the real version which we bump manually.
* dev builds always use 0, so that they are easily distinguishable from prod builds.

Use `-Dtigerbeetle-release` to distinguish which case we are in.

### DIFF
--- a/src/vsr/superblock.zig
+++ b/src/vsr/superblock.zig
@@ -45,7 +45,9 @@ pub const Quorums = @import("superblock_quorums.zig").QuorumsType(.{
     .superblock_copies = constants.superblock_copies,
 });
 
-pub const SuperBlockVersion: u16 = 1;
+pub const SuperBlockVersion: u16 =
+    // Make sure that data files created by development builds are distinguished through version.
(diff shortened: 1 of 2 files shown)


src/vsr/superblock.zig
```diff
@@ -45,7 +45,9 @@ pub const Quorums = @import("superblock_quorums.zig").QuorumsType(.{
     .superblock_copies = constants.superblock_copies,
 });
 
-pub const SuperBlockVersion: u16 = 1;
+pub const SuperBlockVersion: u16 =
+    // Make sure that data files created by development builds are distinguished through version.
+    if (constants.config.process.release.value == vsr.Release.minimum) 0 else 1;
```
sentientwaffle:
```suggestion
    if (constants.config.process.release.value == vsr.Release.minimum.value) 0 else 1;
```

sentientwaffle (approved):

sentientwaffle:
Eventually we will have a data file inspector tool, and then as part of our release pipeline we should verify that data files created by release binaries have non-zero superblock version.

PR #1808, opened 2024-03-27
docs: reliable transaction submission

emschwartz:
- **client: remove vsr.Client Demuxer wrapper**
- **client: annotate message body lifetime**
- **client: remove message.ref assertion**
- **cli: use proper semver in tigerbeetle --version**
- **build: remove TIGERBEETLE_RELEASE env var hack**
- **benchmark: set random user_data for accounts**
- **benchmark: use random seed each time, allow setting it**
- **compaction: use FIFO as a block_pool instead of trying to slice arrays**
- **docs: add page on reliable transaction submission**

### DIFF
--- a/docs/develop/reliable-transaction-submission.md
+++ b/docs/develop/reliable-transaction-submission.md
@@ -16,7 +16,8 @@ especially transfers -- reliably.
 `id` field that is used as an idempotency key to ensure the same object is not created twice.
 
 **The client software, such as your app or web page, that the user interacts with should generate the
-`id` (not your API):**
+`id` (not your API). This `id` should be persisted locally before submission, and the same `id`
+should be used for subsequent retries.**
 
 1. User initiates a transfer.
 2. Client software (app, web page, etc) [generates the transfer `id`](./data-modeling.md#id).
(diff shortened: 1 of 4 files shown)


docs/develop/reliable-transaction-submission.md
```diff
@@ -0,0 +1,56 @@
...
+
+[`Transfer`s](../reference/transfers.md#id) and [`Account`s](../reference/accounts.md#id) carry an
+`id` field that is used as an idempotency key to ensure the same object is not created twice.
+
+**The client software, such as your app or web page, that the user interacts with should generate the
+`id` (not your API):**
+
+1. User initiates a transfer.
+2. Client software (app, web page, etc) [generates the transfer `id`](./data-modeling.md#id).
+3. Client software **persists the `id` in the app or browser local storage.**
```
sentientwaffle, line 24: We should also describe what to do with that persisted id (since just persisting it does nothing if we don't use it for anything if the page reloads).

emschwartz: Good point! I added some more language about this. Lmk what you think

sentientwaffle (approved):

PR #1809, opened 2024-03-27
CLI: Log version early in `tigerbeetle start`

sentientwaffle:
Print critical information like the binary's release before the superblock is even opened.

My objective here is to minimize round trips with developers when troubleshooting. (If someone runs into a bug, they will send us the replica logs. If there is anything we would ask them immediately after that (e.g. "What release are you running?") then we should just include that information in the default logs.)

The new logs look something like this:

    info(main): version=0.0.1+f28ec34
    info(main): release=0.0.1
    info(main): git_commit=f28ec344691748d60824d3becfa21d9d1d1c0bf7

### DIFF
--- a/src/tigerbeetle/main.zig
+++ b/src/tigerbeetle/main.zig
@@ -184,13 +184,22 @@ const Command = struct {
         const nonce = std.crypto.random.int(u128);
         assert(nonce != 0); // Broken CSPRNG is the likeliest explanation for zero.
 
+        // TODO Where should this be set?
+        const release_client_min = config.process.release;
+        const releases_bundled = &[_]vsr.Release{config.process.release};
+
+        log_main.info("version={}", .{constants.semver});
+        log_main.info("release={}", .{config.process.release});
+        log_main.info("release_client_min={}", .{release_client_min});
+        log_main.info("releases_bundled={any}", .{releases_bundled.*});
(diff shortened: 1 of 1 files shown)


src/tigerbeetle/main.zig
```diff
@@ -184,13 +184,22 @@ const Command = struct {
         const nonce = std.crypto.random.int(u128);
         assert(nonce != 0); // Broken CSPRNG is the likeliest explanation for zero.
 
+        // TODO Where should this be set?
+        const release_client_min = config.process.release;
+        const releases_bundled = &[_]vsr.Release{config.process.release};
+
+        log_main.info("version={}", .{constants.semver});
+        log_main.info("release={}", .{config.process.release});
```
matklad, line 191: Do we need to print release? It is the tripple part of version, so maybe just print version until the two do diverge in practice? 

Alternatively, don't print version and print only release and git_commit.

I agree that we should print crucial info early, but I'd also love us to be as succinct as possible otherwise, as redundant and useless info in logs generally makes debugging harder.

sentientwaffle: Good point! I kept `release` but removed `version`. (Reasoning: The other numbers we are printing are also releases. And the version's git commit is already covered a few lines down.)

matklad (approved):

PR #1806, opened 2024-03-27
LSM/Compaction: Pacing review comments/assertions/refactors

sentientwaffle:
This is a bunch of miscellaneous:
- documentation "improvements" (hopefully)
- assertions
- small/local refactors

as a follow-up "review" for https://github.com/tigerbeetle/tigerbeetle/pull/1660.

PR #1810, opened 2024-03-28
compaction: recalculate pacing on each beat

cb22:
Originally, the pacing code would calculate a desired `value_count_per_beat` once, at the start of the bar. This would then be the goal number of source values to process, for each beat.

However, since we pace off _source_ values, we'd often process quite a few more values than desired. This is normal and expected - we have to round up to full target value blocks, filling from the immutable table can consume a lot of values with the removal of duplicates.

Now, recalculate value_count_per_beat each beat. This self adjusts for those effects and helps to smooth out latencies better over the entire bar.

Additionally, add in a few asserts and make sure we don't exceed `beats_max` (our beat budget).

## Before
1241 batches in 52.53 s
load offered = 1000000 tx/s
load accepted = 190363 tx/s
batch latency p1 = 1 ms
batch latency p10 = 4 ms
batch latency p20 = 4 ms
batch latency p30 = 6 ms
batch latency p40 = 18 ms
batch latency p50 = 28 ms
batch latency p60 = 37 ms
batch latency p70 = 47 ms
batch latency p80 = 58 ms
batch latency p90 = 68 ms
batch latency p95 = 79 ms
batch latency p99 = 429 ms
batch latency p100 = 458 ms

## After
1247 batches in 50.53 s
load offered = 1000000 tx/s
load accepted = 197887 tx/s
batch latency p1 = 1 ms
batch latency p10 = 4 ms
batch latency p20 = 4 ms
batch latency p30 = 14 ms
batch latency p40 = 20 ms
batch latency p50 = 27 ms
batch latency p60 = 38 ms
batch latency p70 = 47 ms
batch latency p80 = 55 ms
batch latency p90 = 64 ms
batch latency p95 = 75 ms
batch latency p99 = 361 ms
batch latency p100 = 381 ms

(the effect grows even larger the more compactions there are going on!)

### DIFF
--- a/src/lsm/compaction.zig
+++ b/src/lsm/compaction.zig
@@ -315,10 +315,10 @@ pub fn CompactionType(
             /// Number of beats we should aim to finish this compaction in. It might be fewer, but
             /// it'll never be more.
             beats_max: ?u64,
+            beats_finished: u64 = 0,
             /// The total number of source values for this compaction.
             /// This is fixed for the duration of the compaction.
             compaction_tables_value_count: u64,
-            value_count_per_beat: u64 = 0,
 
             // The total number of source values processed by this compaction across the bar. Must
             // equal compaction_tables_value_count by bar_apply_to_manifest(). Tracked
@@ -396,6 +396,7 @@ pub fn CompactionType(
             };
 
             grid_reservation: Grid.Reservation,
+            value_count_per_beat: u64,
 
             // TODO: This is now always 0 / 1 so get rid of it and just use index_read_done?
             index_blocks_read_b: usize = 0,
@@ -725,14 +726,9 @@ pub fn CompactionType(
             const bar = &compaction.bar.?;
             assert(!bar.move_table);
 
-            assert(bar.value_count_per_beat == 0);
+            assert(bar.beats_max == null);
 
-            // TODO: Move this calculation into beat_grid_reserve, and subtract the values we've
-            // already done from it.
-            // This way we self correct our pacing and better spread the work out....
-            bar.value_count_per_beat = stdx.div_ceil(bar.compaction_tables_value_count, beats_max);
-            assert(bar.value_count_per_beat > 0);
-
+            bar.beats_max = beats_max;
             bar.target_index_blocks = target_index_blocks;
             assert(target_index_blocks.count > 0);
 
@@ -740,11 +736,9 @@ pub fn CompactionType(
             // be null!
             bar.source_a_immutable_block = source_a_immutable_block;
 
-            log.debug("bar_setup_budget({s}): bar.compaction_tables_value_count={} " ++
-                "bar.value_count_per_beat={}", .{
+            log.debug("bar_setup_budget({s}): bar.compaction_tables_value_count={}", .{
                 compaction.tree_config.name,
                 bar.compaction_tables_value_count,
-                bar.value_count_per_beat,
             });
         }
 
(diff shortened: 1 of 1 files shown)


cb22:
(cc @sentientwaffle - in #1806 you removed `beats_max`, but it's used now)

src/lsm/compaction.zig
```diff
@@ -714,25 +715,19 @@ pub fn CompactionType(
...
 
             assert(bar.value_count_per_beat == 0);
+            assert(bar.beats_max == null);
 
-            // TODO: Move this calculation into beat_grid_reserve, and subtract the values we've
-            // already done from it.
-            // This way we self correct our pacing and better spread the work out....
-            bar.value_count_per_beat = stdx.div_ceil(bar.compaction_tables_value_count, beats_max);
-            assert(bar.value_count_per_beat > 0);
-
```
matklad, line 735: We no need to change the name of this function and the comment, as it no longer in fact is in charge of setting up the budget.

src/lsm/compaction.zig
```diff
@@ -754,7 +749,18 @@ pub fn CompactionType(
...
-            assert(bar.value_count_per_beat > 0);
+            assert(bar.beats_max != null);
+
+            // Calculate how many values we have to compact each beat, to self-correct our pacing.
+            // Pacing will have imperfections due to rounding up to fill target value blocks and
+            // immutable table filtering duplicate values.
+            assert(bar.compaction_tables_value_count > bar.source_values_merge_count);
+            const beats_remaining = bar.beats_max.? - bar.beats_finished;
+            assert(beats_remaining > 0);
+            bar.value_count_per_beat = stdx.div_ceil(
```
matklad: This now probably needs to move to `beat` state?

src/lsm/compaction.zig
```diff
@@ -1895,6 +1905,11 @@ pub fn CompactionType(
             // TODO: Can we assert target_values_merge_count > 0 here?
             assert(bar.target_values_merge_count == bar.target_values_write_count);
 
+            // Assert we've finished within the number of beats we were allocated.
+            // TODO(metric): Track the delta between target and actual.
+            if (!bar.move_table)
+                assert(bar.beats_finished <= bar.beats_max.?);
```
matklad, line 1927: My reading is that our style consensus is mildly in favor of using braces here ^^

src/lsm/compaction.zig
```diff
@@ -754,7 +749,18 @@ pub fn CompactionType(
...
+            // Calculate how many values we have to compact each beat, to self-correct our pacing.
+            // Pacing will have imperfections due to rounding up to fill target value blocks and
+            // immutable table filtering duplicate values.
+            assert(bar.compaction_tables_value_count > bar.source_values_merge_count);
+            const beats_remaining = bar.beats_max.? - bar.beats_finished;
+            assert(beats_remaining > 0);
+            bar.value_count_per_beat = stdx.div_ceil(
+                bar.compaction_tables_value_count - bar.source_values_merge_count,
+                beats_remaining,
+            );
```
matklad, line 772: Let's to a post assert here

```
assert(bar.source_values_merge_count + value_count_per_beat * beats_remaining > bar.compaction_tables_value_count)
```

matklad: And, if we _do_ store this data in the bar, let's assert that `value_count_per_beat` is non-incerasing across beats

src/lsm/compaction.zig
cb22: D'oh! :smile:

src/lsm/compaction.zig
cb22: In my head, `beats_max` was always the budget, and `value_count_per_beat` was just a derivation from that. That said, I'm not particularly tied to the function name - especially since it does other stuff too.

matklad: Ah, I see this now!

matklad (approved):

PR #1812, opened 2024-03-28
go: work around windows CI failure

matklad:
go run on windows _occasionally_ fails with

$ go run main.go
> fork/exec C:\Users\RUNNER~1\AppData\Local\Temp\go-build1166602505\b001\exe\main.exe: The request is not supported.

I don't understand what could be happening there. It seems that `go run` itself fails when it tries to spawn a subprocess? But why it does that only occasionally?

Try working around that by splitting `go run` into explicit build&run.


sentientwaffle (dismissed):

sentientwaffle (approved):
LGTM++

Issue #136, opened 2022-08-19
Better docker shutdown behaviour


matdehaast:
Currently when shutting down  tigerbeetle in docker is a slow process. I think TB needs to better handle sigint/sigterm which  could correspond to better shutdown modes. Here is an comment from the postgres docker image

> We set the default STOPSIGNAL to SIGINT, which corresponds to what PostgreSQL
calls "Fast Shutdown mode" wherein new connections are disallowed and any
in-progress transactions are aborted, allowing PostgreSQL to stop cleanly and
flush tables to disk, which is the best compromise available to avoid data
corruption.
Users who know their applications do not keep open long-lived idle connections
may way to use a value of SIGTERM instead, which corresponds to "Smart
Shutdown mode" in which any existing sessions are allowed to finish and the
server stops when all sessions are terminated.
See https://www.postgresql.org/docs/12/server-shutdown.html for more details
about available PostgreSQL server shutdown signals.
See also https://www.postgresql.org/docs/12/server-start.html for further
justification of this as the default value, namely that the example (and
shipped) systemd service files use the "Fast Shutdown mode" for service
termination.

jorangreef:
Thanks @matdehaast 

Can you provide more context?

It looks like Postgres uses `SIGINT` to “stop cleanly and flush tables to disk [..] to avoid data corruption”, which does not apply to TB. You can crash TB safely at any time, as quickly as you like. You shouldn't need to be waiting on anything?

jorangreef:
TB also does not need Postgres' “Smart Shutdown mode”, which Postgres uses to delay shutdown until existing sessions are finished—since the new primary will simply complete the work that the terminating primary was doing, and the new primary will reply directly back to those existing sessions. So again, shutdown with TB should always be instantaneous.

I don't know if Docker is doing something to delay shutdown?

matdehaast:
Thanks @jorangreef. So from my current understand of using TB, when I try kill a docker tigerbeetle it doesn't actually crash. It hits the docker timeout before hard killing it. Which is 20s. It would be nice for it to exit quicker if I kill the docker instance. It may be just the way that is configured to be passed to the TB process? Not sure

jorangreef:
Do you know what signal Docker is sending the TB process? It's probably sending us something very polite :)

Update: Also, confirming that you are using `docker stop ...`? You should probably see no delay if you use `docker kill ...`?

jorangreef:
[...]

PR #1813, opened 2024-03-28
Tools/docker: tini

sentientwaffle:
Docker's default stop-signal is SIGTERM.
TigerBeetle does not have signal handlers, so "docker stop ..." with SIGINT hangs for several seconds. Setting the signal to SIGKILL instead ensures that "docker stop" doesn't delay.

This is a partial fix for https://github.com/tigerbeetle/tigerbeetle/issues/136. (The more complete fix would be adding a `SIGTERM` signal handler.)

### DIFF
--- a/docs/getting-started/with-docker-compose.md
+++ b/docs/getting-started/with-docker-compose.md
@@ -7,9 +7,9 @@ sidebar_position: 5
 First, provision the [data file](../internals/data_file.md) for each node:
 
 ```console
-docker run --init -v $(pwd)/data:/data ghcr.io/tigerbeetle/tigerbeetle format --cluster=0 --replica=0 --replica-count=3 /data/0_0.tigerbeetle
-docker run --init -v $(pwd)/data:/data ghcr.io/tigerbeetle/tigerbeetle format --cluster=0 --replica=1 --replica-count=3 /data/0_1.tigerbeetle
-docker run --init -v $(pwd)/data:/data ghcr.io/tigerbeetle/tigerbeetle format --cluster=0 --replica=2 --replica-count=3 /data/0_2.tigerbeetle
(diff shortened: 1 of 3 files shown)


matklad:
Hm, if we _don't_ have signal handlers, shouldn't SIGTERM kill us outright?

sentientwaffle:
Ok, this is more complicated than I expected!
If you send `SIGTERM` to a non-docker tigerbeetle replica, it _does_ exit immediately (as anticipated). 

The pause seems to be a common docker issue:

- https://github.com/docker-library/postgres/issues/184
- https://github.com/docker-library/redis/issues/386

Possible explanation is that in docker, the replica is PID 1, which has special behavior around signals: https://github.com/rustdesk/rustdesk-server/issues/36#issuecomment-1159504574

matklad:
How idiomatic it is to _not_ run as pid1 in Docker? I've read somewhere a very reasonable argument that one should have a proper pid1 in Docker, but I am not sure if that's something people actually do?

cb22:
> How idiomatic it is to _not_ run as pid1 in Docker? I've read somewhere a very reasonable argument that one should have a proper pid1 in Docker, but I am not sure if that's something people actually do?

It's quite idiomatic. Last I looked, it was even recommend for eg Node, and they bundle Tini - a small init system for this in later Docker releases.

matklad:
Yeah, let's maybe do the same then? I am sure there's more weirdness to pid1

sentientwaffle:
~Ok, this ends up being a documentation update.~

There are a couple ways tini can be used:
1. `docker run --init ...` (and an equivalent config for docker-compose).
2. Install it into the container.

~Number 2 ends up being surprisingly complicated and finicky (...alpine), so I went with 1.~

Nevermind, 1 works, it was just a DNS issue with my local docker.

sentientwaffle:
Alpine version bump is not related, but may as well keep it fresh!

cb22 (approved):
Agreed that (2) is better in any case! Older versions of Docker won't have `--init`, etc.

PR #1814, opened 2024-03-28
lsm: assert that runtime block metadata matches comptime one

matklad:
Blocks on disk are type-erased. They store tree id, value/block count max and key/value size in the block header, such that the blocks can be processed, in an "untyped" manner.

When we access the blocks however, we statically know the corresponding values from the comptime type of the tree. Verify that the two match.

Noticed this today on Twitch!


sentientwaffle (approved):
LGTM! If we ever want to change the count parameters in the future, we can ease those assertions.

PR #1815, opened 2024-03-28
Remove approximate pacing TODO

batiati:
Ref #1240

### DIFF
--- a/src/vsr/grid.zig
+++ b/src/vsr/grid.zig
@@ -155,10 +155,6 @@ pub fn GridType(comptime Storage: type) type {
             cache_interface.hash_address,
             .{
                 .ways = set_associative_cache_ways,
-                // TODO: This is temporary, until compaction pacing is merged. Use a sub-optimal
-                // cache line size, since combined with the large block size of 1MiB, we would
-                // require a 2GiB grid cache minimum otherwise.
-                .cache_line_size = 16,
                 .value_alignment = @alignOf(u64),
             },
         );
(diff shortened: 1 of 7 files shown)


src/vsr/grid.zig
```diff
@@ -155,10 +155,6 @@ pub fn GridType(comptime Storage: type) type {
             cache_interface.hash_address,
             .{
                 .ways = set_associative_cache_ways,
-                // TODO: This is temporary, until compaction pacing is merged. Use a sub-optimal
-                // cache line size, since combined with the large block size of 1MiB, we would
-                // require a 2GiB grid cache minimum otherwise.
-                .cache_line_size = 16,
```
batiati, line 161: Now that we settled with 512KiB grid blocks, what's the requirement of minimum grid cache:

1. Default `cache_line_size = 64` and minimum grid cache of 1GB ?
2. Keep the minimum grid cache at 512MiB and increase `cache_line_size = 32` ?
3. Keep `cache_line_size = 16`, remove the TODO and document the new minimum grid cache to 256MiB?

cb22: Hmmm. This is worth thinking about - we do want to allow the smallest possible memory footprint out the box.

How about, if `--cache-grid` < 1024MB, `cache_line_size` is 16, otherwise it's 64? That'll allow low values to work, while still being optimal for higher ones. It would be nice to be able to go even lower though... Maybe even zero grid cache.

cb22: @batiati let me know if you'd like to do the above, otherwise I'm happy to take it on.

batiati: Feel free to tackle it, I would love to, but I won't be able to work on it until next week or so.

PR #1818, opened 2024-03-29
docs: add two recent talks

emschwartz:


matklad:
Oh, might add https://www.youtube.com/watch?v=eFTQzhfO6nc as well. Something like

> In this impromptu talk, matklad discusses build and release process of TigerBeetle. 

I am OoO until the next week, so might not be fast with reviews!

matklad (approved):

PR #1823, opened 2024-04-02
Webapi

lil5:
A begin to adding a web api to tigerbeetle - in tigerbeetle itself

- [x] add `tigerbeetle api` cli command
- [x] add zap package
- [x] use zap in cli
- [ ] implement repl operations in web
  - [ ] pulse
  - [ ] create_accounts
  - [ ] create_transfers
  - [ ] lookup_accounts
  - [ ] lookup_transfers
  - [ ] get_account_transfers
  - [ ] get_account_balances

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -111,12 +111,10 @@ pub fn build(b: *std.Build) !void {
     const vsr_options_module = options.createModule();
     const vsr_module = b.addModule("vsr", .{
         .source_file = .{ .path = "src/vsr.zig" },
-        .dependencies = &.{
-            .{
-                .name = "vsr_options",
-                .module = vsr_options_module,
-            },
-        },
+        .dependencies = &.{.{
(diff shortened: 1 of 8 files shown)


matklad:
My understanding is that we most likely don't want this in the current form, as this violates a couple of major invariants for us:

* we don't depend on third party code (using external libraries is a pain in an unstable language, and opens up an avenue for supply chain attack)
* the code is simple and mechanically sympathetic (and HTTP implementation, even if implemented in the main code base, would not be simple or mechanically sympathetic)

We probably do want to have some sort of rest API on top of tigerbeetle, but I think that's most likely is going to be implemented as an extranl process. IE, something like Rust will be talking to web using well-established HTTP libraries, and then rely the info to the tigerbeetle using the existing extern struct based protocol.

matklad:
Closing as not planned in the current form!

PR #1827, opened 2024-04-05
compaction: add standalone stage to blocks

cb22:
There are some use cases where we use blocks without wrapping them in a CompactionBlockFIFO.

Previously, they just always stayed in the `.free` stage so their lifetimes were never asserted.

Add an extra stage, `.standalone`, so we can assert these are in the correct stage now too.

### DIFF
--- a/src/lsm/compaction.zig
+++ b/src/lsm/compaction.zig
@@ -122,7 +122,7 @@ pub fn CompactionHelperType(comptime Grid: type) type {
             read: Grid.Read = undefined,
             write: Grid.Write = undefined,
 
-            stage: enum { free, pending, ready, ioing } = .free,
+            stage: enum { free, pending, ready, ioing, standalone } = .free,
             next: ?*CompactionBlock = null,
         };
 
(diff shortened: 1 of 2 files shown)


src/lsm/compaction.zig
```diff
@@ -810,6 +810,9 @@ pub fn CompactionType(
             assert(blocks.source_value_blocks[1].count > 0);
             assert(blocks.target_value_blocks.count > 0);
 
+            assert(blocks.source_index_block_a.stage == .standalone);
+            assert(blocks.source_index_block_b.stage == .standalone);
```
sentientwaffle: :thinking: Maybe to be symmetrical with `beat_blocks_unassign`, we could assert that these are `.free` and transition them to `standalone` here?

src/lsm/tree_fuzz.zig
```diff
@@ -324,11 +324,18 @@ fn EnvironmentType(comptime table_usage: TableUsage) type {
                     continue;
                 }
                 const source_a_immutable_block = env.block_pool.pop().?;
+                source_a_immutable_block.stage = .standalone;
```
sentientwaffle: `bar_setup_budget` should assert the `source_a_immutable_block`'s `stage` as well.

sentientwaffle (approved):

PR #1830, opened 2024-04-08
release: fix missing .zip suffix

cb22:
`if (debug) "-debug" else "" ++ ".zip"` is read as `if (debug) "-debug" else ("" ++ ".zip")` and not
`(if (debug) "-debug" else "") ++ ".zip`.

There's no actual bug here, because `zip` adds `.zip` for you automatically. But still, make it explicit.


matklad (approved):
Makes sense! 

The code to check that this still works correctly would be our release validation.

PR #1831, opened 2024-04-08
docs: clarify sorting for get_account_balances

emschwartz:


docs/reference/account_filter.md
```diff
@@ -70,7 +70,7 @@ matches the parameter [`account_id`](#account_id).
 
 #### `flags.reversed`
 
-Whether the results are sorted by timestamp in chronological or reverse-chronological order.
+Whether the results are sorted by timestamp in chronological or reverse-chronological order. If this flag is set, the results will be sorted in reverse-chronological order.
```
matklad: Might be more clear to expand the definition of "chronological": when the flag is set, transfers with larger timestamps go first. Not sure how to best word this though, but I fear that the negations here might be confusing, and that it's better to say positively what happens.

docs/reference/operations/get_account_transfers.md
```diff
@@ -8,15 +8,17 @@ The query filter. See [`AccountFilter`](../account_filter.md) for constraints.
 
 ## Result
 
-- If any matching transfers exist, return an array of [`Transfer`](../transfers.md)s.  
-- If no matching transfers exist, return nothing.  
-- If any constraint is violated, return nothing. 
+- If any matching transfers exist, return an array of [`Transfer`](../transfers.md)s.
+- By default, `Transfer`s are sorted chronologically by `timestamp`. You can use the [`reversed`](../account_filter.md#flagsreversed) to change this.
+- The result is always limited in size. If there are more results, you need to page through them using the `AccountFilter`'s [`timestamp_min`](../account_filter.md#timestamp_min) and/or [`timestamp_max`](../account_filter.md#timestamp_max).
+- If no matching `Transfer`s exist, return nothing.
```
matklad: 1st and 4th bullet are connected and should  be sequential. But I actually don't like "return nothing", as it's not specific. Surely, _something_ is returned to signify the absence of results? 

Maybe use the following instead of both:

* Returns a (possibly empty) array of matching transfers
* If any constraint is violated, return an empty array.
* By default, `Transfer`s are sorted chronologically by `timestamp`...
* The result is always limited in size ....

matklad (approved):

PR #1832, opened 2024-04-08
Tidy: Fix multiline string parsing

sentientwaffle:
The string `"\\"` is a single (escaped) backslash. But we want to match a double-backslash, which is `"\\\\"`.

### DIFF
--- a/src/tidy.zig
+++ b/src/tidy.zig
@@ -350,7 +350,7 @@ fn has_link(line: []const u8) bool {
 
 /// If a line is a `\\` string literal, extract its value.
 fn parse_multiline_string(line: []const u8) ?[]const u8 {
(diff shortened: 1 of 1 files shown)


src/tidy.zig
```diff
@@ -350,7 +350,7 @@ fn has_link(line: []const u8) bool {
 
 /// If a line is a `\\` string literal, extract its value.
 fn parse_multiline_string(line: []const u8) ?[]const u8 {
-    const cut = stdx.cut(line, "\\") orelse return null;
+    const cut = stdx.cut(line, "\\\\") orelse return null;
```
matklad, line 353: :rofl: thought that this is ugly, and that we need to use a raw literal here...

which would look like

```
\\\\
```

anyway, but for a different reason

matklad (approved):

PR #1833, opened 2024-04-08
Test/StorageChecker: Remove infinite loop

sentientwaffle:


matklad (approved):
LGTM, but perhaps even better would be to use a counting loop here (we know the total size, so then number of blocks as well), and assert that the linked list has a correct length.

PR #1836, opened 2024-04-09
cfo: add continuous fuzzing orchestrator

matklad:
Entry Point:

```
./zig/zig build scripts -- cfo --concurrency=8 --budget-minutes=10 --hang-minutes=30
```

We have a number of machines which run

    git clone https://github.com/tigerbeetle/tigerbeetle && cd tigerbeetle
    while True:
        git fetch origin && git reset --hard origin/main
        ./scripts/install_zig.sh
        ./zig/zig build scripts -- cfo

By modifying this script, we can make those machines do interesting things.

The primary use-case is fuzzing: `cfo` runs a random fuzzer, and, if it finds a failure, it is recorded in devhubdb.

Specifically:

CFO keeps `args.concurrency` fuzzes running at the same time. For simplicity, it polls currently running fuzzers for completion every second in a fuzzing loop. A fuzzer fails if it returns non-zero error code.

The fuzzing loops runs for `args.budget_minutes`. To detect hangs, if any fuzzer is still running after additional `args.hang_minutes`, it is killed (thus returning non-zero status and recording a failure).

After the fuzzing loop, CFO collects a list of seeds, some of which are failing. Next, it merges, this list into previous set of seeds (persisting seeds is to be implemented, at the moment the old list is always empty).

Rules for merging:

- Keep seeds for at most `commit_count_max` distinct commits.
- Prefer fresher commits (based on commit time stamp).
- For each commit and fuzzer combination, keep at most `seed_count_max` seeds.
- Prefer failing seeds to successful seeds.
- Prefer older seeds.

These rules ensure that in the steady state (assuming fuzzer's clock doesn't fail) the set of seeds is stable. If the clock goes backwards, there might be churn in the set of a seed, but the number of failing seeds will never decrease.

TODO:

- rather than fuzzing the same repository, checkout a fresh copy into a separate directory. This would be required to fuzz the release branch.
- actually plug this into some continuously running machines
- upload fuzzing results somewhere

---

While the above is required to make this practically useful, it seems worthwhile to submit existing code implementing  the main driving&merging loop as is!

### DIFF
--- a/src/scripts/cfo.zig
+++ b/src/scripts/cfo.zig
@@ -1,0 +1,492 @@
+//! Continuous Fuzzing Orchestrator.
+//!
+//! We have a number of machines which run
+//!
+//!     git clone https://github.com/tigerbeetle/tigerbeetle && cd tigerbeetle
+//!     while True:
+//!         git fetch origin && git reset --hard origin/main
+//!         ./scripts/install_zig.sh
+//!         ./zig/zig build scripts -- cfo
+//!
+//! By modifying this script, we can make those machines do interesting things.
+//!
+//! The primary use-case is fuzzing: `cfo` runs a random fuzzer, and, if it finds a failure, it is
+//! recorded in devhubdb.
+//!
+//! Specifically:
+//!
+//! CFO keeps `args.concurrency` fuzzes running at the same time. For simplicity, it polls currently
+//! running fuzzers for completion every second in a fuzzing loop. A fuzzer fails if it returns
+//! non-zero error code.
+//!
+//! The fuzzing loops runs for `args.budget_minutes`. To detect hangs, if any fuzzer is still
+//! running after additional `args.hang_minutes`, it is killed (thus returning non-zero status and
+//! recording a failure).
+//!
+//! It is important that the caller (systemd typically) arranges for CFO to be a process group
+//! leader. It is not possible to reliably wait for (grand) children with POSIX, so its on the
+//! call-site to cleanup any run-away subprocesses
+//!
+//! After the fuzzing loop, CFO collects a list of seeds, some of which are failing. Next, it
+//! merges, this list into previous set of seeds (persisting seeds is to be implemented, at the
+//! moment the old list is always empty).
+//!
+//! Rules for merging:
+//!
+//! - Keep seeds for at most `commit_count_max` distinct commits.
+//! - Prefer fresher commits (based on commit time stamp).
+//! - For each commit and fuzzer combination, keep at most `seed_count_max` seeds.
+//! - Prefer failing seeds to successful seeds.
+//! - Prefer older seeds.
+//!
+//! These rules ensure that in the steady state (assuming fuzzer's clock doesn't fail) the set of
+//! seeds is stable. If the clock goes backwards, there might be churn in the set of a seed, but the
+//! number of failing seeds will never decrease.
+
+const std = @import("std");
+const builtin = @import("builtin");
+const log = std.log;
+const assert = std.debug.assert;
+
+const stdx = @import("../stdx.zig");
+const flags = @import("../flags.zig");
+const fatal = flags.fatal;
+const Shell = @import("../shell.zig");
+
+pub const CliArgs = struct {
+    budget_minutes: u64 = 10,
+    hang_minutes: u64 = 30,
+    concurrency: ?u32 = null,
+    // Personal Access Token for <https://github.com/tigerbeetle/devhubdb>.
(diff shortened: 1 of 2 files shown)


matklad:
>While the above is required to make this practically useful, it seems worthwhile to submit existing code implementing the main driving&merging loop as is!

Oh, I'd say it's also in a good shape to put it on a server somewhere, with the relevant systemd glue. It won't be immediately useful, but then it already will be changeable without extra redeploy

src/scripts/cfo.zig
```diff
@@ -0,0 +1,484 @@
...
+const fatal = flags.fatal;
+const Shell = @import("../shell.zig");
+
+pub const CliArgs = struct {
+    budget_minutes: u64 = 10,
+    hang_minutes: u64 = 30,
+    concurrency: u32 = 16,
+};
+
+const Fuzzer = enum {
```
cb22, line 64: I was wondering why it seemed obsessed with the cache_map and ewah :rofl:

src/scripts/cfo.zig
```diff
@@ -0,0 +1,484 @@
...
+
+const stdx = @import("../stdx.zig");
+const flags = @import("../flags.zig");
+const fatal = flags.fatal;
+const Shell = @import("../shell.zig");
+
+pub const CliArgs = struct {
+    budget_minutes: u64 = 10,
+    hang_minutes: u64 = 30,
+    concurrency: u32 = 16,
```
cb22: Might be nicer to default this to num_cpu_cores in Zig than in the calling script?

src/scripts/cfo.zig
```diff
@@ -0,0 +1,484 @@
...
+                    const seed_record = SeedRecord{
+                        .commit_timestamp = commit_timestamp,
+                        .commit_sha = commit_sha,
+                        .fuzzer = fuzzer,
+                        .ok = false,
+                        .seed_timestamp = @as(u64, @intCast(std.time.timestamp())),
+                        .seed = seed,
+                        .command = command,
+                    };
+                    log.debug("will start '{s}'", .{command});
```
cb22, line 136: Maybe cleanup of these log messages (the corresponding `will reap` one below is more confusing, it makes it sound like something odd is happening when its just the fuzzer being done)

src/scripts/cfo.zig
```diff
@@ -0,0 +1,484 @@
...
+//! recorded in devhubdb.
+//!
+//! Specifically:
+//!
+//! CFO keeps `args.concurrency` fuzzes running at the same time. For simplicity, it polls currently
+//! running fuzzers for completion every second in a fuzzing loop. A fuzzer fails if it returns
+//! non-zero error code.
+//!
+//! The fuzzing loops runs for `args.budget_minutes`. To detect hangs, if any fuzzer is still
+//! running after additional `args.hang_minutes`, it is killed (thus returning non-zero status and
```
cb22, line 23: This is a neat way to handle both keeping the concurrency and hanging!

src/scripts/cfo.zig
matklad: I will probably remove cache map —- it takes very long on some of the seeds

src/scripts/cfo.zig
cb22: Although - it might be too conservative currently. Managed to hit it on my laptop which is pretty fast.

matklad: I’d say that cache map might be a too slow fuzzer under some configurations? I’d say half an hour per fuzzer is what we’d want here

cb22: We can probably limit its fuzz ops quite dramatically, I would be surprised if it weren't one of the quicker fuzzers on a per op basis...

src/scripts/cfo.zig
matklad: Clarity is in the eye of the beholder! To me, these do sound very clear:

* "will" comes from how VS Code names its events. It is always `WillDoSomething` or `DidDoSomething`, which makes it clear when something happens relative to the callback
* "reap" is I think is a term of art? We know that the process is done already. Calling `wait` to retrieve its status is called reaping


These are intentionally `debug` level logs --- I won't expect them to see during "normal" operation, but its helpful to debug the code.

cb22: > comes from how VS Code

No wonder I found it confusing! :laughing: 

(It might be from how I normally use  `will` when I write my log comments)

cb22 (approved):
Let's get it merged and running!

mattnite:
Stealing

PR #1841, opened 2024-04-11
cfo: integration with devhub

matklad:
* after cfo finishes, it uploades data to devhubdb
* the failing seeds for the last commits are displayed on the hub

With this change, differenc computers might race to update devhubdb. For
this reason, add a retry loop around two places where we do the upload.

### DIFF
--- a/src/scripts/cfo.zig
+++ b/src/scripts/cfo.zig
@@ -57,11 +57,10 @@ pub const CliArgs = struct {
     budget_minutes: u64 = 10,
     hang_minutes: u64 = 30,
     concurrency: ?u32 = null,
-    // Personal Access Token for <https://github.com/tigerbeetle/devhubdb>.
-    devhgubdb_pat: ?[]const u8 = null,
 };
 
 const Fuzzer = enum {
+    canary,
     ewah,
     vsr_free_set,
     vsr_superblock,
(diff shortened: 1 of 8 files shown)


matklad:
follow up to https://github.com/tigerbeetle/tigerbeetle/pull/1836

src/scripts/cfo.zig
```diff
@@ -179,24 +210,83 @@ pub fn main(shell: *Shell, gpa: std.mem.Allocator, cli_args: CliArgs) !void {
...
+    log.info("uploading {} seeds", .{seeds_new.len});
 
-    const json = try std.json.stringifyAlloc(
-        shell.arena.allocator(),
-        merged,
-        .{ .whitespace = .indent_2 },
-    );
-    std.debug.print("{s}\n", .{json});
+    // Personal Access Token for <https://github.com/tigerbeetle/devhubdb>.
+    const token = try shell.env_get("DEVHUBDB_PAT");
```
matklad, line 228: @cb22 if we spin this somewhere, that's the env var to use to setup personall access token for `https://github.com/tigerbeetle/devhubdb>`. I suggest you generate your own token, so that at least two people know the current process (we use the same tokens for devhub metrics, go repo and the docs website)

This is an env var and not a cli argument to be consistent with `scripts/devhub.zig`

src/scripts/cfo.zig
```diff
@@ -72,34 +71,67 @@ pub fn main(shell: *Shell, gpa: std.mem.Allocator, cli_args: CliArgs) !void {
...
+    seeds: *std.ArrayList(SeedRecord),
+    options: struct {
+        commit: [40]u8,
+        concurrency: usize,
+        budget_seconds: u64,
+        hang_seconds: u64,
+    },
+) !void {
+    // Fuzz an independent clone of the repository, so that CFO and the fuzzer could be on
+    // different (to fuzz PRs).
```
cb22:
```suggestion
    // different branches (to fuzz PRs).
```

cb22 (dismissed):
LGTM, just a small comment nit.

cb22 (approved):

Issue #1835, opened 2024-04-09
thread 19356 panic: file system does not support Direct I/O


lil5:
Error:
```
$ ./tigerbeetle format --cluster=0 --replica=0 --replica-count=1 0_0.tigerbeetle
thread 19356 panic: file system does not support Direct I/O
```

I built from source on a linux workstation, the home user directory is encrypted with ecryptfs.

sentientwaffle:
Hi @lil5!

Did you build with `-Dconfig=production`? If so, try building again without that flag.

We enforce Direct IO for the production config since it is critical for performance.
But I think ecryptfs is not compatible with direct I/O (since ecryptfs is a stacked filesystem that rewrites data).

(And by the way, the data files between `-Dconfig=production` and the default config are identical.)

lil5:
I built with only the -build flag.

I'll close this issue, I'm moving to a MacBook Air & my vps works great with tigerbeetle ❤️. So I'm not going to put too much into this issue. 

Thanks, I advise any future viewers to just run tigerbeetle from an external ssd if you have the same issue.

sentientwaffle:
Inspired by this issue, I implemented https://github.com/tigerbeetle/tigerbeetle/pull/1843 -- once that gets released, this should be simpler. You'll be able to just pass `--development` at runtime to a production binary to make it run without Direct IO.

PR #1843, opened 2024-04-11
CLI: Add `--development` runtime argument to gate Direct IO panic

sentientwaffle:
### Context

- Production Linux binaries (`-Dconfig=production`) use `constants.direct_io_required == true` to require that Direct IO is available. (Excluding data files created on tmpfs.) If Direct IO is not available, they will panic.
- Non-production Linux binaries (`-Dconfig=development`) allow non-direct-IO, and just print a warning.

### Considerations

- Direct IO is critical to safety, to ensure the data is properly durable when the write completes.
- But we can't guarantee that a user's development environment has access to Direct IO. For example, some [file systems](https://github.com/tigerbeetle/tigerbeetle/issues/1835) inherently don't support it.
- We don't want to have multiple release binaries for Linux (i.e. one for production and one for development).
- When safety and user experience are in conflict, we prioritize safety.

### Solution

Add a new CLI flag to `tigerbeetle format` and `tigerbeetle start`. When `--development` is set, the absence of Direct IO is downgraded from a panic to a warning.

When `--development` is *not* set, print an informative message explaining all of this.

I considered several other alternate names, e.g. `--allow-buffered-io`. But while that is more precise, it is less clear to the end user, who is not as familiar with TigerBeetle's internals.


cb22 (approved):
LGTM!

A future consideration: will we want to stack other things on top of `--development` in future? If so - it might make sense to _keep_ the `tmpfs` checking logic. But if not, then it could be removed and if you want to run on `tmpfs`, you'll need to specify `--development` too.

sentientwaffle:
> A future consideration: will we want to stack other things on top of --development in future?

I'm open to this, though I haven't thought of any specific use cases yet.

PR #1848, opened 2024-04-16
Query engine 1/N

batiati:
This PR is the first part of a multi-PR effort to implement the QueryAPI.
Evolving the existing Scans, adding the required building blocks for queries with the minimal amount of work that makes sense to be reviewed together:

### 1. ScanBuilder sanitization:

Implements support for `id` and `timestamp` (previously, only secondary indexes were supported);
Removes the support for range queries over secondary indexes; only equality by the prefix is supported now.

With that, we removed the **unsupported use case** of merging scans not sorted by timestamp from this API.

### 2. ScanRange type:

Range queries (e.g. >, <, >=, <=) are still possible without merging scans, just not supported by the `ScanBuilder` anymore.

The new type `ScanRange` is non-composable and acts like an adapter between the raw `ScanTree` (which returns the `CompositeKey`) and the `ScanLookup` (which expects a timestamp).

The implementation of the expiry of pending transfers was changed to use a `ScanRange`, taking advantage of the new API that allows the caller to execute inline filters during the matching. 
Instead of two queries, now a single scan will suffice to fulfill the pending transfer logic. See [comments](https://github.com/tigerbeetle/tigerbeetle/pull/1657#discussion_r1523105225) for more details. No noticeable performance gain, but the code was greatly simplified.

### 3. Zig-zag merge join:

Implements the zig-zag merge join algorithm and all the machinery required to probe scans during the "zag" phase.

This refactors the existing K-way merge implementation to accommodate the required API for probing streams.
The most notable change is that the iterator is now initialized only during the first read, allowing "resets" when the underlying streams change due to a "zag".

There are unit and fuzz tests for the Zig-zag merge implementation.

### Pending work and next steps:

- Merge differences (not implemented yet).
- Fuzzer for `ScanBuilder ` and `ScanMerge` (most of the code present in this PR **is not reachable in production**. During development, I did tests and benchmarks using the existing `get_account_transfers` operation with a provisory modified state machine, which performed more complex AND clauses).
- Query API wire protocol (payload, results, errors, pagination, batching, etc);
- Client APIs (for each target language);


src/lsm/zig_zag_merge.zig
```diff
@@ -0,0 +1,665 @@
...
+                    // The stream must remain drained after probed.
+                    assert(stream_peek(it.context, @intCast(stream_index)) == error.Drained);
+                }
+                return error.Drained;
+            }
+
+            return probe_key;
+        }
+
+        fn peek_key_classic(it: *ZigZagMergeIterator) error{Drained}!?Key {
```
batiati: `peek_keys_classic` is the reference implementation of the ZigZag over 2 streams (the "seed" and the "probe") . It's here only for the purpose of the review, and will be removed.

This implementation is less efficient since it can unnecessarily probe streams with a seed that may be behind the other streams not included in the current iteration. For example, it zig-zags only between `A` and `B` first and then uses the matching key as the seed to zig-zag into `C`, only to discover that `C` might be far ahead from the seed, restarting the process with `C` and `A` .

The optimized  `peek_keys` version, reads all streams in one pass, probing those that don't match at once.
For example, it peeks at `A`,`B`, and `C` for the most advanced key. If `C` produces the key, it probes `A` and `B` and repeats the process, probing the remaining until all streams match.

src/lsm/zig_zag_merge.zig
```diff
@@ -0,0 +1,664 @@
...
+        pub fn pop(it: *ZigZagMergeIterator) error{Drained}!?Value {
+            while (try it.peek_key()) |key| {
+                const value = stream_pop(it.context, 0);
+                assert(key_from_value(&value) == key);
+                for (1..it.streams_count) |stream_index| {
+                    const value_other = stream_pop(it.context, @intCast(stream_index));
+                    assert(key_from_value(&value_other) == key);
+
+                    if (constants.verify) {
+                        // It's assumed that streams will produce the same value.
```
sentientwaffle: Wait, why are they all same value _type_?

(Updated later:) Ok, it is because all of the streams produce only timestamps, not objects or composite keys. Will we later use zig-zag merge iterator with `Key != Value`?

src/lsm/zig_zag_merge.zig
```diff
@@ -0,0 +1,664 @@
...
+                        assert(stdx.equal_bytes(Value, &value, &value_other));
+                    }
+                }
+
+                if (it.previous_key_popped) |previous| {
+                    switch (std.math.order(previous, key)) {
+                        .lt => assert(it.direction == .ascending),
+                        // Discard duplicate values.
+                        .eq => continue,
```
sentientwaffle: In what situation could there be a duplicate value?

src/lsm/zig_zag_merge.zig
```diff
@@ -0,0 +1,664 @@
...
+                // Looking into all non-drained streams for a match, while accumulating
+                // the most ahead key to probe the streams behind.
+                probing = BitSet.initEmpty();
+                for (0..it.streams_count) |stream_index| {
+                    if (drained.isSet(stream_index)) continue;
+
+                    const key = stream_peek(it.context, @intCast(stream_index)) catch |err| {
+                        switch (err) {
+                            // Return immediately on empty streams.
+                            error.Empty => return null,
```
sentientwaffle: Elaborating a little, since I had to think this through:

```suggestion
                            // Return immediately on empty streams.
                            // If any one stream is empty, then there can be no value remaining in the intersection.
                            error.Empty => return null,
```

src/lsm/zig_zag_merge.zig
```diff
@@ -0,0 +1,664 @@
...
+                        .ascending => key > probe_key,
+                        .descending => key < probe_key,
+                    }) {
+                        // The stream is ahead, it will be the probe key,
+                        // meaning all streams before must be probed.
+                        probe_key = key;
+
+                        // Setting all previous streams as `true` except the drained ones.
+                        probing.setRangeValue(.{ .start = 0, .end = stream_index }, true);
+                        probing.setIntersection(drained.complement());
```
sentientwaffle, line 159: Can we assert now that `stream_index` is not set? (To make it clear that the `end` of `setRangeValue` is exclusive.)

Though, this might be more clearly represented as a `probe_stream` usize/index rather than a `probing` bitset. Then just before the `probing_iterator` loop we could set `probing` all at once, as the complement of `drained` minus the `probe_stream`. That would allow us to remove the subsequent `else if (switch (it.direction) ...` branch. And we could remove the `probing = BitSet.initEmpty();` at the start of the loop.

src/lsm/zig_zag_merge.zig
```diff
@@ -0,0 +1,664 @@
...
+                        assert(switch (it.direction) {
+                            .ascending => key > probe_key,
+                            .descending => key < probe_key,
+                        });
+                    }
+                }
+            }
+
+            if (drained.count() == it.streams_count) {
+                //Can't probe if all streams are drained.
```
sentientwaffle:
```suggestion
                // Can't probe if all streams are drained.
```

src/lsm/zig_zag_merge.zig
```diff
@@ -0,0 +1,664 @@
...
+                while (drained_iterator.next()) |stream_index| {
+                    stream_probe(it.context, @intCast(stream_index), probe_key);
+
+                    // The stream must remain drained after probed.
+                    assert(stream_peek(it.context, @intCast(stream_index)) == error.Drained);
+                }
+                return error.Drained;
+            }
+
+            return probe_key;
```
sentientwaffle: Maybe expand this a little to make it clearer what the final state is here? I had to reread the above nested loops a few times to (hopefully) grok it. Instead of `while (drained_iterator.next())`, something like:

```suggestion
            // At this point, all the buffered streams have produced a match
            // that will be used to probe the drained streams.
            if (drained.count() > 0) {
                for (0..stream_count) |stream| {
                    if (drained.isSet(stream)) {
                        stream_probe(it.context, @intCast(stream_index), probe_key);
                
                        // The stream must remain drained after probed.
                        assert(stream_peek(it.context, @intCast(stream_index)) == error.Drained);
                    } else {
                        assert(stream_peek(it.context, @intCast(stream_index)).? == probe_key);
                    }
                }
                return error.Drained;
            } else {
                for (0..stream_count) |stream| {
                    assert(stream_peek(it.context, @intCast(stream_index)).? == probe_key);
                }
                return probe_key;
            }
```

src/lsm/zig_zag_merge.zig
```diff
@@ -0,0 +1,664 @@
...
+
+fn TestContext(comptime streams_max: u32) type {
+    const testing = std.testing;
+
+    return struct {
+        const ZigZagMergeIterator = @This();
+
+        const log = false;
+
+        //
```
sentientwaffle:
```suggestion
```

src/lsm/k_way_merge.zig
```diff
@@ -38,58 +41,75 @@ pub fn KWayMergeIteratorType(
...
-        k: u32,
+        k: u32 = 0,
 
-        direction: Direction,
         previous_key_popped: ?Key = null,
 
-        /// This function may create an Iterator with k being less than stream_count_max if
-        /// stream_peek() for one of the streams immediately returns null.
-        pub fn init(context: *Context, stream_count_max: u32, direction: Direction) Self {
-            // TODO Do we ever expect stream_count_max to be 0?
```
sentientwaffle, line 57: Is it possible for `streams_count` to be zero? If so, let's add a `maybe()` with an explanation, because this is surprising to me.

src/lsm/zig_zag_merge.zig
batiati: Well, I can't foresee any concrete case, even something like `(A OR B) AND (A OR C)` would not produce a duplicated value.

However, since our binary search supports duplicated values, it would be reasonable to assume that `{1,1} ∩ {1,1} = {1}`. But I agree it's better to `unreachable` to avoid masking unintended behavior.

src/lsm/scan_tree.zig
```diff
@@ -581,6 +701,56 @@ fn ScanTreeLevelType(comptime ScanTree: type, comptime Storage: type) type {
...
+
+            if (self.values == .fetching) {
+                // The key couldn't be found in the buffered data.
+                // The level iterator must read the index block again from the new key range.
+                //
+                // TODO: We may use the already buffered `index_block` to check if the key
+                // is present in other value blocks within the same table, advancing the level
+                // iterator instead of calling `table_next(.begin)`.
+                // However, it's most likely the index block is still in the grid cache, so this
+                // may not represent any real improvement.
```
sentientwaffle, line 763: Is it possible for `self.manifest` to equal `.finished` at this point? Or is that what this TODO is referencing, that it might be `finished` but then we re-fetch the last table once more?

src/lsm/scan_builder.zig
```diff
@@ -420,26 +452,67 @@ pub fn ScanType(
...
+                inline else => |*scan_impl, tag| {
+                    const ScanTree = @TypeOf(scan_impl.*);
+                    const Value = ScanTree.Tree.Table.Value;
+
+                    if (comptime is_composite_key(Value)) {
+                        const prefix = prefix: {
+                            const prefix_min = Value.key_prefix(scan_impl.key_min);
+                            const prefix_max = Value.key_prefix(scan_impl.key_max);
+                            if (prefix_min == prefix_max) break :prefix prefix_min;
+                            unreachable;
```
sentientwaffle: Why not
```suggestion
                            assert(prefix_min == prefix_max);
                            break :prefix prefix_min;
```

src/lsm/scan_builder.zig
```diff
@@ -452,3 +525,100 @@ pub fn ScanType(
...
+        value: *const Tree.Table.Value,
+    ) callconv(.Inline) EvaluateNext,
+    /// Extracts the ObjectTree's timestamp from the table value.
+    comptime timestamp_from_value: fn (
+        context: EvaluatorContext,
+        value: *const Tree.Table.Value,
+    ) callconv(.Inline) u64,
+) type {
+    return struct {
+        const Scan = @This();
```
sentientwaffle: Could we call this `ScanRange` instead? Since there is also a `ScanType`/`Scan` in the same file.

src/lsm/scan_builder.zig
```diff
@@ -452,3 +525,100 @@ pub fn ScanType(
...
     };
 }
+
+pub const EvaluateNext = enum {
+    include_and_continue,
+    include_and_stop,
+    exclude_and_continue,
+    exclude_and_stop,
+};
+
```
sentientwaffle: Add a high-level explanation, since all of these different Scans can be hard for me to keep track of :stuck_out_tongue:  

```suggestion

/// Apply a custom filter and/or stop-condition when scanning a range of values.
```

src/lsm/zig_zag_merge.zig
batiati: >Though, this might be more clearly represented as a probe_stream usize/index rather than a probing bitset.

We may have more than one stream as the "seed".

Example:
```
A.peek() == 2
B.peek() == 2
C.peek() == 1
```

The probing bitset will contain just `C`, with `probe_key=2`.

sentientwaffle: Ahh right, nevermind!

src/lsm/zig_zag_merge.zig
batiati: > Will we later use zig-zag merge iterator with Key != Value?

Yes, scans produce only timestamps, and since the composite key prefix is always fixed (equality), they also take only the timestamp as the probe key, so `Key == Value` for merging scans.

I don't think we should enforce `Key == Value`, the intersection logic can be reused as a primitive. 

EDIT: I just realized that this comment could be clarified.

src/lsm/k_way_merge.zig
batiati: It might be useful to represent empty sets, but I don't know what use case was in mind when this comment was originally written.

src/lsm/scan_tree.zig
batiati: Yes, either `.finished` or `.iterating_blocks`, it will need to read the index block again to determine if the probe_key can be in a value block ahead of the current one. The `TODO` comment applies to both cases.

src/lsm/zig_zag_merge.zig
```diff
@@ -0,0 +1,546 @@
...
+                .context = context,
+                .streams_count = streams_count,
+                .direction = direction,
+            };
+        }
+
+        // Resets the iterator when the underlying streams are moved.
+        // It's not necessary for ZigZagMerge, but it follows the same API for all MergeIterators.
+        pub fn reset(it: *ZigZagMergeIterator) void {
+            _ = it;
```
sentientwaffle, line 69: Is this ever called, or could it be `unreachable`?

src/state_machine.zig
```diff
@@ -2086,6 +1980,136 @@ fn sum_overflows(comptime Int: type, a: Int, b: Int) bool {
...
+            assert(self.phase == .idle);
+            self.* = .{};
+        }
+
+        fn scan(
+            self: *ExpirePendingTransfers,
+            tree: *Tree,
+            buffer: *const ScanBuffer,
+            snapshot: u64,
+            timestamp: u64,
```
sentientwaffle: Can you fold these two parameters into a struct? Since they are both `u64`s, they would be very easy to mix up at the call site.

Also, maybe rename `timestamp` to be more descriptive, to avoid confusing it with the `pulse_timestamp`. (Likewise in `running.timestamp`). e.g. `prepare_timestamp`?

src/state_machine.zig
```diff
@@ -2086,6 +1980,136 @@ fn sum_overflows(comptime Int: type, a: Int, b: Int) bool {
...
+        const ScanLookupStatus = @import("lsm/scan_lookup.zig").ScanLookupStatus;
+
+        const Tree = std.meta.FieldType(TransfersGroove.IndexTrees, .expires_at);
+        const Key = Tree.Table.Key;
+        const Value = Tree.Table.Value;
+
+        // TODO(zig) Context should be `*ExpirePendingTransfers`,
+        // but its a dependency loop.
+        const Context = struct {};
+        const Scan = ScanRangeType(
```
sentientwaffle: Call this `ScanRange`... I am easily mixed-up by all of the "Scans" floating around. :sweat_smile:

src/lsm/scan_range.zig
```diff
@@ -0,0 +1,103 @@
...
+            const on_read_callback = struct {
+                fn callback(ctx: *Context, ptr: *ScanTree) void {
+                    const parent: *ScanRange = @fieldParentPtr(ScanRange, "scan_tree", ptr);
+                    ctx.callback(ctx, parent);
+                }
+            }.callback;
+
+            scan.scan_tree.read(context, on_read_callback);
+        }
+
```
sentientwaffle: I think our convention is to extract the callback, so that reading top-to-bottom flows nicely. e.g.:

```
pub fn read(scan: *ScanRange, context: *Context) void {
...
}

fn read_callback(context: *Context, scan_tree: *ScanTree) void {
    const parent: *ScanRange = @fieldParentPtr(ScanRange, "scan_tree", scan_tree);
    context.callback(context, parent);
}
```

src/state_machine.zig
```diff
@@ -1091,56 +1010,31 @@ pub fn StateMachineType(
...
         }
 
-        fn prefetch_expire_pending_transfers_scan_callback(scan_lookup: *TransfersScanLookup) void {
-            const self: *StateMachine = ScanLookup.parent(.transfer, scan_lookup);
+        fn prefetch_expire_pending_transfers_scan_callback(
+            scan_lookup: *ExpirePendingTransfers.ScanLookup,
+        ) void {
+            const self: *StateMachine = ScanLookup.parent(.expire_pending_transfers, scan_lookup);
+            self.expire_pending_transfers.finish(scan_lookup.state);
             self.scan_result_count = @intCast(scan_lookup.slice().len);
```
sentientwaffle: Can we assert `self.scan_result_count == 0` before we overwrite it?

src/state_machine.zig
```diff
@@ -997,91 +1001,6 @@ pub fn StateMachineType(
             assert(self.prefetch_timestamp >= TimestampRange.timestamp_min);
             assert(self.prefetch_timestamp != TimestampRange.timestamp_max);
```
sentientwaffle, line 1016: Since we are about to start overwriting `scan_buffer`, can we assert that `scan_result_count == 0`?

src/state_machine.zig
sentientwaffle, line 1: In `fn commit()`, can we add

```zig
maybe(self.scan_result_count == 0);
defer assert(self.scan_result_count == 0);
```

Though now that I think about it, maybe we should change `scan_result_count` from a `u32` to a `?u32`, so we can be clearer in the code about the difference between "no results" and "no results _yet_".

sentientwaffle, line 1: `StateMachine` has a `scan_buffer` which is (afaict?) distinct from the `ScanBuffer`/`ScanBufferPool` types.

Could we call it `scan_lookup_buffer` instead? And correspondingly, s/`scan_result_count`/`scan_lookup_buffer_result_count` (I know it is long, but I want to emphasize that those two fields correspond).

src/state_machine.zig
```diff
@@ -1091,56 +1010,31 @@ pub fn StateMachineType(
...
             scan_lookup.read(
                 scan_buffer,
                 &prefetch_expire_pending_transfers_scan_callback,
             );
         }
 
-        fn prefetch_expire_pending_transfers_scan_callback(scan_lookup: *TransfersScanLookup) void {
-            const self: *StateMachine = ScanLookup.parent(.transfer, scan_lookup);
+        fn prefetch_expire_pending_transfers_scan_callback(
+            scan_lookup: *ExpirePendingTransfers.ScanLookup,
```
sentientwaffle, line 1048: Maybe `scan_lookup.read`'s callback should pass `ScanLookup.slice()` as a second argument, to denote that the ownership is transferring back to the state machine? Then `ScanLookup.lookup_worker_finished` could set `self.buffer = null` at the same time it sets `self.callback = null`.

src/state_machine.zig
```diff
@@ -2086,6 +1980,136 @@ fn sum_overflows(comptime Int: type, a: Int, b: Int) bool {
...
+        inline fn value_next(context: *Context, value: *const Value) EvaluateNext {
+            const self: *ExpirePendingTransfers = @fieldParentPtr(
+                ExpirePendingTransfers,
+                "context",
+                context,
+            );
+            assert(self.phase == .running);
+
+            const expires_at: u64 = value.field;
+            self.pulse_timestamp = expires_at;
```
sentientwaffle: What can we assert about `pulse_timestamp` before the update? Something like `assert(pulse_timestamp == TimestampRange.timestamp_max or pulse_timestamp < expires_at)`?

src/state_machine.zig
```diff
@@ -2086,6 +1980,136 @@ fn sum_overflows(comptime Int: type, a: Int, b: Int) bool {
...
+        fn scan(
+            self: *ExpirePendingTransfers,
+            tree: *Tree,
+            buffer: *const ScanBuffer,
+            snapshot: u64,
+            timestamp: u64,
+        ) *Scan {
+            assert(self.phase == .idle);
+            assert(timestamp >= TimestampRange.timestamp_min and
+                timestamp <= TimestampRange.timestamp_max);
```
sentientwaffle: Can we assert that `pulse_timestamp <= timestamp`?

src/state_machine.zig
```diff
@@ -1784,8 +1678,8 @@ pub fn StateMachineType(
 
                 // In case the pending transfer's timeout is exactly the one we are using
                 // as flag, we need zero the valut to run the next `pulse`.
```
```suggestion
                // as flag, we need zero the value to run the next `pulse`.
```

src/state_machine.zig
```diff
@@ -2086,6 +1980,136 @@ fn sum_overflows(comptime Int: type, a: Int, b: Int) bool {
...
+                } },
+            };
+            return &self.phase.running.scan;
+        }
+
+        fn finish(self: *ExpirePendingTransfers, status: ScanLookupStatus) void {
+            assert(self.phase == .running);
+            switch (status) {
+                .scan_finished => if (self.pulse_timestamp <= self.phase.running.timestamp) {
+                    // There are no more transfers to expire in the next pulse.
```
sentientwaffle: Clarifying (both for me and the comment): in this case, there are no unexpired pending transfers left at all, right?

src/state_machine.zig
```diff
@@ -2086,6 +1980,136 @@ fn sum_overflows(comptime Int: type, a: Int, b: Int) bool {
...
+
+        context: Context = undefined,
+        phase: union(enum) {
+            idle,
+            running: struct {
+                scan: Scan,
+                timestamp: u64,
+            },
+        } = .idle,
+        pulse_timestamp: u64 = TimestampRange.timestamp_min,
```
sentientwaffle: Not sure if these descriptions are correct; please correct me if necessary:

```suggestion
        phase: union(enum) {
            idle,
            running: struct {
                scan: Scan,
                /// Collect all transfers with an expiry less-than-or-equal-to this timestamp.
                timestamp: u64,
            },
        } = .idle,
        /// When ==timestamp_min, there may be pending transfers to expire, but we need to scan to check.
        /// When ==timestamp_max, there are no pending transfers to expire.
        /// Otherwise, this is the timestamp of the next pending transfer expiry.
        pulse_timestamp: u64 = TimestampRange.timestamp_min,
```

src/lsm/scan_tree.zig
```diff
@@ -271,6 +288,73 @@ pub fn ScanTreeType(
                 },
                 .needs_data => return error.ReadAgain,
                 .buffering => unreachable,
+                .aborted => return null,
+            }
+        }
+
+        /// Probing a scan modifies the key_min/key_max range, either moving the scan if the next
+        /// key is already buffered, or causing it to read again from the new key range.
+        pub fn probe(self: *ScanTree, key: Key) void {
```
sentientwaffle: Clarify that in both cases the outcome of `probe()` is that the next key from the `ScanTree` will be either probe-key, or (if probe-key doesn't exist in the tree) the next key after probe-key in the index.

sentientwaffle: Also maybe rename `key` to `probe_key`. "key" is a very generic term, so being more specific can hint at its purpose.

src/lsm/scan_tree.zig
```diff
@@ -271,6 +288,73 @@ pub fn ScanTreeType(
...
+                return;
+            } else if (self.key_max < key) {
+                if (self.direction == .ascending) self.abort();
+                return;
+            }
+
+            // Updates the scan range depending on the direction.
+            switch (self.direction) {
+                .ascending => {
+                    if (self.key_min == key) return; // No need to move
```
sentientwaffle: Doesn't the `ZigZagMergeIterator` avoid calling `probe` on streams that already have the right key?
Could we assert that `key_min != key` here instead of returning? (Likewise for the `.descending` branch.)

src/lsm/zig_zag_merge.zig
```diff
@@ -0,0 +1,546 @@
...
+            assert(it.streams_count <= streams_max);
+            assert(it.streams_count > 1);
+
+            const key_min: Key = switch (it.direction) {
+                .ascending => 0,
+                .descending => std.math.maxInt(Key),
+            };
+
+            var drained: BitSet = BitSet.initEmpty();
+            var probe_key: Key = key_min;
```
sentientwaffle, line 114: :thinking: If we persisted this on `ZigZagMergeIterator`, then we could verify that it never regresses, i.e. verify the correct implementation of `stream_probe` across iterations, by checking that `stream_peek`'s return value is after-or-equal-to the latest probe key. (e.g. Right now if `ScanTree` had a bug in its probe implementation, I think it would be possible for zig-zag to endlessly retry probe-drained-probe-drained and never advance.)

src/lsm/scan_builder.zig
```diff
@@ -18,14 +19,27 @@ const TimestampRange = @import("timestamp_range.zig").TimestampRange;
 
 const Error = @import("scan_buffer.zig").Error;
 
-/// ScanBuilder is a helper to create and combine scans using
-/// any of the Groove's secondary indexes.
 pub fn ScanBuilderType(
```
sentientwaffle: Maybe just "ScanBuilder is a helper to create and combine scans using any of the Groove's indexes."?

src/lsm/zig_zag_merge.zig
batiati: It's called, so `ScanMerge` doesn't need to know which type of `merge_iterator` needs a reset.

src/lsm/scan_range.zig
batiati: Nice catch, just realized that I copied this code from `ScanBuilder` where the inner type depends on the dispatcher type :man_facepalming:.

src/lsm/scan_builder.zig
batiati: :+1:  I removed the comment while I was thinking in the `TODO` next line, since it won't be single-groove soon.

src/lsm/scan_tree.zig
batiati: We avoid probing a scan already in the right key, but it's possible to probe the same scan twice, first as buffered and then as drained. 

This `if` does not check the current key, it only allows double-probing, so we don't have to track it during the iteration.

https://github.com/tigerbeetle/tigerbeetle/blob/7a6a44176cf9930033d690f23cd2e8dc3dd343ce/src/lsm/zig_zag_merge.zig#L162-L176

src/lsm/zig_zag_merge.zig
```diff
@@ -207,8 +205,22 @@ pub fn ZigZagMergeIteratorType(
...
+            assert(it.probe_key_previous == null or
+                probe_key == it.probe_key_previous.? or
+                it.key_ahead(probe_key, it.probe_key_previous.?));
+
+            it.probe_key_previous = probe_key;
             return if (drained.count() == 0) probe_key else error.Drained;
         }
+
+        /// Depending on the direction, returns true if key `a` is ahead of key `b`.
+        inline fn key_ahead(it: *const ZigZagMergeIterator, a: Key, b: Key) bool {
```
sentientwaffle: Nit, but maybe change this to take a `struct { key_before, key_after }`, since I think the argument order is confusing otherwise.

src/lsm/scan_range.zig
```diff
@@ -66,16 +66,14 @@ pub fn ScanRangeType(
...
-                fn callback(ctx: *Context, ptr: *ScanTree) void {
-                    const parent: *ScanRange = @fieldParentPtr(ScanRange, "scan_tree", ptr);
-                    ctx.callback(ctx, parent);
-                }
-            }.callback;
-
             scan.scan_tree.read(context, on_read_callback);
         }
 
+        fn on_read_callback(ctx: *Context, ptr: *ScanTree) void {
```
sentientwaffle: Nit: s/ctx/context to match the other usage.

src/lsm/scan_tree.zig
```diff
@@ -292,44 +292,42 @@ pub fn ScanTreeType(
             }
         }
 
-        /// Probing a scan modifies the key_min/key_max range, either moving the scan if the next
-        /// key is already buffered, or causing it to read again from the new key range.
-        pub fn probe(self: *ScanTree, key: Key) void {
+        /// Modifies the key_min/key_max range and moves the scan to the next value
+        /// such as `>=probe_key` or `<=probe_key` depending on the direction.
```
sentientwaffle: Confused by the wording, what about:

```
        /// such that `value.key >= probe_key` (ascending) or `value.key <= probe_key` (descending).
```

sentientwaffle:
Aside from those last 3 nits, this LGTM!

sentientwaffle (approved):

PR #1850, opened 2024-04-16
Fix panic when the replica receives a reserved command.

batiati:
Fixes a panic when a replica receives a `command == .reserved` from the client  (occurs when a client < 0.15.3 connects to the cluster).

https://github.com/tigerbeetle/tigerbeetle/blob/f7c9a8735417a07ce0fc5addbf41b82300068a94/src/vsr/message_header.zig#L192-L216

### DIFF
--- a/src/message_bus.zig
+++ b/src/message_bus.zig
@@ -745,6 +745,12 @@ fn MessageBusType(comptime process_type: vsr.ProcessType) type {
                         return null;
                     }
 
+                    if (header.command == .reserved) {
+                        log.err("reserved command received from {}", .{connection.peer});
+                        connection.terminate(bus, .shutdown);
+                        return null;
+                    }
+
                     if (header.cluster != bus.cluster) {
                         log.err("message addressed to the wrong cluster: {}", .{header.cluster});
                         connection.terminate(bus, .shutdown);
@@ -837,6 +843,8 @@ fn MessageBusType(comptime process_type: vsr.ProcessType) type {
                 header: *const Header,
             ) bool {
                 comptime assert(process_type == .replica);
+
+                assert(header.command != .reserved);
 
                 assert(bus.cluster == header.cluster);
                 assert(bus.connections_used > 0);


src/message_bus.zig
```diff
@@ -730,6 +730,12 @@ fn MessageBusType(comptime process_type: vsr.ProcessType) type {
                 @memcpy(mem.asBytes(&header), data[0..@sizeOf(Header)]);
 
                 if (!connection.recv_checked_header) {
+                    if (header.command == .reserved) {
+                        log.err("reserved command received from {}", .{connection.peer});
+                        connection.terminate(bus, .shutdown);
+                        return null;
+                    }
+
```
matklad: We should check this _after_ we validate the checksum, to make sure it is really reserved, rather than just a packet corruption

matklad:
> Fixes a panic when a replica receives a command == .reserved from the client (occurs when a client < 0.15.3 connects to the cluster).

Hm, but even old clients _shouldn't_ be sending `.reserved`, no?

batiati:
> Hm, but even old clients _shouldn't_ be sending `.reserved`, no?

I wondered the same, and vaguely assumed that it would be due to the offset of the `command` field being changed by the new `release` and `protocol` fields. The checksum is still valid, but the fields are interpreted differently.

matklad:
:thinking: hm, so I think what should happen _logically_ in this case is that we should reject the message because of the version mismatch, and we already have code for that here: 

https://github.com/tigerbeetle/tigerbeetle/blob/81806334726ef7438c74d5f630b126b32431663a/src/vsr/replica.zig#L4724-L4744

This logic doesn't trigger, because we get a framing error and misinterpret fields.

I think it is a correct behavior for us to crash here --- we generally do assume that the is trusted, and, if something is so wrong that we see a message which we can not understand at all, we should crash.

Hopefully we won't have any more of this bugs in the future!

So let's maybe add a stability test instead? For `Header.Request` and for `Header.Prepare`, let's have a test that casts a raw hard-coded byte slice, and asserts that the resulting header is `.valid()`, and that all fields have expected value. That way, we freeze specific meaning of bytes in time.

batiati:
> I think it is a correct behavior for us to crash here

The fact that a malformed request can crash the server makes me nervous, even in a trusted network :fearful: 

What if we included some information about the header layout in the hash, so the expected checksum would always change when we added/removed fields? For example, using a comptime salt based on the off-set of each field :thinking:

matklad:
Summarising slack discussion: we _do_ want to drop connection, rather than panic in this case, but this entire header parsing logic should be moved over to replica, such that:

* replica is in control of message validity invariants
* the code is vopred

stretch goal is to add a dedicated fuzzer here.

PR #1851, opened 2024-04-16
VSR: Grid Scrubber

sentientwaffle:
## Context

A "data scrubber" is a background task that gradually/incrementally reads the disk and validates what it finds. Its purpose is to discover faults proactively – as early as possibly – rather than waiting for them to be discovered by normal database operation (e.g. during compaction).

["An Analysis of Latent Sector Errors in Disk Drives" (2007)](https://research.cs.wisc.edu/adsl/Publications/latent-sigmetrics07.pdf) found that >60% of latent sector errors (the most common type of disk fault) were discovered by a scrubber that cycles every 2 weeks. (Note that this data was collected on HDDs, not SSDs.)

Finding and repairing errors proactively minimizes the risk of cluster-wide data loss due to multiple intersecting faults (analogous to a "double-fault") – a scenario where we fail to read a block, and try to repair that block from another replica, only to discover that the copy of the block on the remote replica's disk is *also* faulty.

## Grid Scrubber

- Only acquired blocks are scrubbed.
- Scrubbing is driven by a replica timer (`grid_scrub_timeout`).
- The replica adjusts the scrubber's timer interval at runtime, based on:
  - The number of grid blocks.
  - The `grid_scrubber_cycle_ms` constant.
- `grid_scrubber_cycle_ms` is the (approximate) goal for how long it takes to scrub the entire grid.
  - Currently set to a default of 90 days: 90 days ≈ 2 weeks * 6 replicas.
  - We can scrub a 16TiB data file every 90 days via ~2.16 MiB/s of reads.
  - (We can set it higher, but another goal is to keep the disk-throughput overhead of the scrubber very low.)
  - See `grid_scrubber_cycle_ticks` in `constants.zig` for more detail.
- The grid scrubber tours across all block types:
  - Scrub table index blocks via `ForestTableIterator`.
  - Immediately after scrubbing a table index block, scrub each of its data blocks in sequence.
  - Scrub flushed manifest blocks.
  - Scrub the freeset and client-sessions trailers from the latest checkpoint.
  - `replica_test.zig` has a test which verifies that all blocks (and thus all block types) are covered by the scrubber.
- When the scrubber encounters a fault (i.e. we find an unexpected or corrupt block), it reports it to the replica, which queues the block in `GridBlocksMissing` for repair.

### Scrubber Invariant

Scrubber cycles have the invariant:

- On an idle replica (i.e. not committing), a full tour from "init" to "done" scrubs every acquired block in the grid.
- On a non-idle replica, a full tour from "init" to "done" scrubs all blocks that survived the entire span of the tour, but may not scrub blocks that were added during the tour.

### DIFF
--- a/src/constants.zig
+++ b/src/constants.zig
@@ -356,6 +356,39 @@ comptime {
     assert(grid_missing_tables_max > 0);
 }
 
+/// The maximum number of concurrent scrubber reads.
+///
+/// Unless the scrubber cycle is extremely short and the data file very large there is no need to
+/// set this higher than 1.
+pub const grid_scrubber_reads_max = config.process.grid_scrubber_reads_max;
+
+/// `grid_scrubber_cycle_ms` is the (approximate, target) total milliseconds per scrub of each
+/// replica's entire grid. Scrubbing work is spread evenly across this duration.
+///
+/// Napkin math for the "worst case" scrubber read overhead as a function of cycle duration
+/// (assuming a fully-loaded data file – maximum size and 100% acquired):
+///
+///   storage_size_limit_max      = 16TiB
+///   grid_scrubber_cycle_seconds = 90 days * 24 hr/day * 60 min/hr * 60 s/min (1 cycle/year)
+///   read_bytes_per_second       = storage_size_max / grid_scrubber_cycle_seconds ≈ 2.16 MiB/s
+///
+pub const grid_scrubber_cycle_ticks = config.process.grid_scrubber_cycle_ms / tick_ms;
+
+/// Accelerate/throttle scrubber reads if they are less/more frequent than this range.
+/// (This is to keep the timeouts from being too extreme when the grid is tiny or huge.)
+pub const grid_scrubber_interval_ticks_min = config.process.grid_scrubber_interval_ms_min / tick_ms;
+pub const grid_scrubber_interval_ticks_max = config.process.grid_scrubber_interval_ms_max / tick_ms;
+
+comptime {
+    assert(grid_scrubber_reads_max > 0);
+    assert(grid_scrubber_reads_max <= grid_iops_read_max);
+    assert(grid_scrubber_cycle_ticks > 0);
+    assert(grid_scrubber_cycle_ticks > @divFloor(std.time.ms_per_min, tick_ms)); // Sanity-check.
+    assert(grid_scrubber_interval_ticks_min > 0);
+    assert(grid_scrubber_interval_ticks_min <= grid_scrubber_interval_ticks_max);
+    assert(grid_scrubber_interval_ticks_max > 0);
+}
+
 /// The minimum and maximum amount of time in milliseconds to wait before initiating a connection.
 /// Exponential backoff and jitter are applied within this range.
 pub const connection_delay_min_ms = config.process.connection_delay_min_ms;
--- a/src/vsr/grid_scrubber.zig
+++ b/src/vsr/grid_scrubber.zig
@@ -20,6 +20,7 @@
 //!
 //! TODO Start replicas scrubbing from distinct/random offsets in the tour to farther minimize risk
 //! of cluster data loss. (Right now replicas will often have identical scrubbing schedules.)
+//! TODO Accelerate scrubbing rate (at runtime) if faults are detected frequently.
 const std = @import("std");
 const assert = std.debug.assert;
 const maybe = stdx.maybe;
--- a/src/vsr/grid_scrubber.zig
+++ b/src/vsr/grid_scrubber.zig
@@ -169,7 +169,9 @@ pub fn GridScrubberType(comptime Forest: type) type {
                 assert(!scrubber.forest.grid.free_set.is_free(index_address));
 
                 if (scrubber.forest.grid.free_set.is_released(index_address)) {
-                    scrubber.tour_skip_table_data();
+                    // Skip scrubbing the table data, since the table is about to be released.
+                    const tables = scrubber.tour.table_data.tables;
+                    scrubber.tour = .{ .table_index = .{ .tables = tables } };
                 }
             }
         }
@@ -246,7 +248,19 @@ pub fn GridScrubberType(comptime Forest: type) type {
                     stdx.copy_disjoint(.inexact, u8, scrubber.tour_index_block, result.valid);
                     scrubber.tour.table_data.index_block = scrubber.tour_index_block;
                 } else {
-                    scrubber.tour_skip_table_data();
+                    // The scrubber can't scrub the table data blocks until it has the corresponding
+                    // index block. We will wait for the index block, and keep re-scrubbing it until
+                    // it is repaired (or until the block is released by a checkpoint).
+                    //
+                    // (Alternatively, we could just skip past the table data blocks, and we will
+                    // come across them again during the next cycle. But waiting for them makes for
+                    // nicer invariants + tests.)
+                    log.debug("{}: read_next_callback: waiting for index repair " ++
+                        "(address={} checksum={x:0>32})", .{
+                        scrubber.superblock.replica_index.?,
+                        read.read.address,
+                        read.read.checksum,
+                    });
                 }
             }
 
@@ -392,13 +406,6 @@ pub fn GridScrubberType(comptime Forest: type) type {
             tour.* = .init;
 
             return null;
-        }
-
-        fn tour_skip_table_data(scrubber: *GridScrubber) void {
-            assert(scrubber.tour == .table_data);
-
-            const tables = scrubber.tour.table_data.tables;
-            scrubber.tour = .{ .table_index = .{ .tables = tables } };
         }
     };
 }
--- a/src/vsr/grid_scrubber.zig
+++ b/src/vsr/grid_scrubber.zig
@@ -94,9 +94,8 @@ pub fn GridScrubberType(comptime Forest: type) type {
         tour: union(enum) {
             init,
             done,
-            table_index: struct { tables: ForestTableIterator },
+            table_index,
             table_data: struct {
-                tables: ForestTableIterator,
                 index_checksum: u128,
                 index_address: u64,
                 /// Points to `tour_index_block` once the index block has been read.
@@ -109,6 +108,10 @@ pub fn GridScrubberType(comptime Forest: type) type {
             free_set: struct { index: usize = 0 },
             client_sessions: struct { index: usize = 0 },
         },
+
+        /// When tour == .init, tour_tables == .{}
+        /// When tour == .done, tour_tables.next() == null.
+        tour_tables: ForestTableIterator,
 
         /// Contains a table index block when tour=table_data.
         tour_index_block: BlockPtr,
@@ -126,6 +129,7 @@ pub fn GridScrubberType(comptime Forest: type) type {
                 .forest = forest,
                 .client_sessions_checkpoint = client_sessions_checkpoint,
                 .tour = .init,
+                .tour_tables = .{},
                 .tour_index_block = tour_index_block,
             };
         }
@@ -145,6 +149,7 @@ pub fn GridScrubberType(comptime Forest: type) type {
             }
 
             scrubber.tour = .init;
+            scrubber.tour_tables = .{};
         }
 
         /// Cancel queued reads to blocks that will be released by the imminent checkpoint.
@@ -174,8 +179,7 @@ pub fn GridScrubberType(comptime Forest: type) type {
 
                 if (scrubber.forest.grid.free_set.is_released(index_address)) {
                     // Skip scrubbing the table data, since the table is about to be released.
-                    const tables = scrubber.tour.table_data.tables;
-                    scrubber.tour = .{ .table_index = .{ .tables = tables } };
+                    scrubber.tour = .table_index;
                 }
             }
         }
@@ -297,7 +301,7 @@ pub fn GridScrubberType(comptime Forest: type) type {
         fn tour_next(scrubber: *GridScrubber) ?BlockId {
             const tour = &scrubber.tour;
             if (tour.* == .init) {
-                tour.* = .{ .table_index = .{ .tables = ForestTableIterator{} } };
+                tour.* = .table_index;
             }
 
             if (tour.* == .table_data) {
@@ -331,13 +335,12 @@ pub fn GridScrubberType(comptime Forest: type) type {
                         .block_type = .data,
                     };
                 } else {
-                    const tables = tour.table_data.tables;
-                    tour.* = .{ .table_index = .{ .tables = tables } };
+                    tour.* = .table_index;
                 }
             }
 
             if (tour.* == .table_index) {
-                if (tour.table_index.tables.next(scrubber.forest)) |table_info| {
+                if (scrubber.tour_tables.next(scrubber.forest)) |table_info| {
                     if (Forest.Storage == TestStorage) {
                         scrubber.superblock.storage.verify_table(
                             table_info.address,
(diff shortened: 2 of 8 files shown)


matklad:
> (Note that this data was collected on HDDs, not SSDs.)

Would be great to refresh that reference. "Why you motivate storage fault model by referencing to papers about spinning rust nobody is using anymore" is a very reasonable question to ask :rofl:

src/constants.zig
```diff
@@ -356,6 +356,39 @@ comptime {
...
+pub const grid_scrubber_cycle_ticks = config.process.grid_scrubber_cycle_ms / tick_ms;
+
+/// Accelerate/throttle scrubber reads if they are less/more frequent than this range.
+/// (This is to keep the timeouts from being too extreme when the grid is tiny or huge.)
+pub const grid_scrubber_interval_ticks_min = config.process.grid_scrubber_interval_ms_min / tick_ms;
+pub const grid_scrubber_interval_ticks_max = config.process.grid_scrubber_interval_ms_max / tick_ms;
+
+comptime {
+    assert(grid_scrubber_reads_max > 0);
+    assert(grid_scrubber_reads_max <= grid_iops_read_max);
```
matklad, line 384: Explaining to oneself: although surprising, I think we do want to non-strict inequality here, so that TigerBeetle _could_ work if `grid_iops_read_max` is `1`.

src/vsr/grid_scrubber.zig
```diff
@@ -0,0 +1,481 @@
...
+                index_address: u64,
+                /// Points to `tour_index_block` once the index block has been read.
+                index_block: ?BlockPtr = null,
+                data_block_index: u32 = 0,
+            },
+            /// The manifest log tour iterates manifest blocks in reverse order.
+            /// (To ensure that manifest compaction doesn't lead to missed blocks.)
+            manifest_log: struct { iterator: ManifestBlockIterator = .init },
+            free_set: struct { index: usize = 0 },
+            client_sessions: struct { index: usize = 0 },
```
matklad, line 121: I am wondering, hypothetically, can we scrub not by explicitly iterating all the data structures in a well-typed manner, but rather on the raw block level?:

* maintain a list of block address/checksum to read
* which is initially populated from pointers from the supeblock
* on each step:
  * pop the block pointer from the list
  * read the block
  * match on block's type
  * add any pointers from this block to the list


This feels potentially simpler, as there are no specialized iterators. But maybe it is not? Things like manifest ending abruptly (not with a null, but with a dangling pointer) make this strategy hard.

src/vsr/grid_scrubber.zig
```diff
@@ -0,0 +1,481 @@
...
+        ///   every acquired block in the grid.
+        /// - On a non-idle replica, a full tour from "init" to "done" scrubs all blocks that
+        ///   survived the entire span of the tour, but may not scrub blocks that were added during
+        ///   the tour.
+        tour: union(enum) {
+            init,
+            done,
+            table_index: struct { tables: ForestTableIterator },
+            table_data: struct {
+                tables: ForestTableIterator,
```
matklad: Not sure about storing `ForestTableIterator` as a part of this enum. It's nice that we "have" the iterator only when we are in the corrcet state, but I think the iterator is a rathre sizable transfer, and, I near, moving it between enum variants would add a lot of memcpys. 

Maybe store all the iterators directly as fields? With the invariant that they are exhausted when we are not reading from them.

src/vsr/grid_scrubber.zig
```diff
@@ -0,0 +1,481 @@
...
+                    read.ignore = true;
+                }
+            }
+
+            scrubber.tour = .init;
+        }
+
+        /// Cancel queued reads to blocks that will be released by the imminent checkpoint.
+        /// (The read still runs, but the results will be ignored.)
+        pub fn checkpoint(scrubber: *GridScrubber) void {
```
matklad, line 174: Ah, that's nifty! I wonder how the scrubber detects when a block gets freed, this is really elegant way to go about it!

src/vsr/replica.zig
```diff
@@ -269,6 +270,8 @@ pub fn ReplicaType(
         grid_repair_tables: IOPS(RepairTable, constants.grid_missing_tables_max) = .{},
         grid_repair_writes: IOPS(BlockWrite, constants.grid_repair_writes_max) = .{},
         grid_repair_write_blocks: [constants.grid_repair_writes_max]BlockPtr,
+        grid_scrubber: GridScrubber,
+        grid_scrubber_reads: IOPS(GridScrubber.Read, constants.grid_scrubber_reads_max) = .{},
```
matklad: It it feels a bit odd that Replica owns these reads, rather than the scrubber directly.

src/vsr/replica.zig
```diff
@@ -3021,6 +3043,51 @@ pub fn ReplicaType(
...
+                    });
+
+                    self.grid.blocks_missing.enqueue_block(
+                        fault.block_address,
+                        fault.block_checksum,
+                    );
+                }
+            }
+
+            if (self.grid_scrubber_reads.acquire()) |read| {
```
matklad: If we allow several scrubber reads, shouldn't this be a `while`? That is, try to enque many concurrent reads if we can?

src/vsr/grid_scrubber.zig
```diff
@@ -0,0 +1,481 @@
...
+            read: Grid.Read = undefined,
+            block_type: ?schema.BlockType,
+
+            /// Set to true before the read completes if the scrub result should be ignored:
+            /// - The scrub succeeded – no need for repair.
+            /// - The block was freed by a checkpoint in the time that the read was in progress.
+            ///   (At checkpoint, the FreeSet frees blocks released during the preceding checkpoint.
+            ///   We can scrub released blocks, but not free blocks. Setting this flag ensures that
+            ///   GridScrubber doesn't require a read-barrier at checkpoint.)
+            ignore: bool,
```
matklad: To make this a bit more self-explanatory and greppable, might use a `status: enum { ok, canceled, overwritten,, ... }` here.

src/vsr/grid_scrubber.zig
```diff
@@ -0,0 +1,481 @@
...
+                        index_schema.data_addresses_used(scrubber.tour_index_block);
+                    const data_block_checksums =
+                        index_schema.data_checksums_used(scrubber.tour_index_block);
+                    return .{
+                        .block_checksum = data_block_checksums[data_block_index].value,
+                        .block_address = data_block_addresses[data_block_index],
+                        .block_type = .data,
+                    };
+                } else {
+                    const tables = tour.table_data.tables;
```
matklad: assert that `data_block_index == index_schema.data_blocks_used(scrubber.tour_index_block)`

src/vsr/grid_scrubber.zig
```diff
@@ -0,0 +1,481 @@
...
+                if (tour.free_set.index < free_set_trailer.block_count()) {
+                    const index = tour.free_set.index;
+                    tour.free_set.index += 1;
+                    return .{
+                        .block_checksum = free_set_trailer.block_checksums[index],
+                        .block_address = free_set_trailer.block_addresses[index],
+                        .block_type = .free_set,
+                    };
+                } else {
+                    tour.* = .{ .client_sessions = .{} };
```
matklad, line 420: ditto, assert index==count

src/vsr/grid_scrubber.zig
```diff
@@ -0,0 +1,481 @@
...
+                if (tour.client_sessions.index < client_sessions.block_count()) {
+                    const index = tour.client_sessions.index;
+                    tour.client_sessions.index += 1;
+                    return .{
+                        .block_checksum = client_sessions.block_checksums[index],
+                        .block_address = client_sessions.block_addresses[index],
+                        .block_type = .client_sessions,
+                    };
+                } else {
+                    tour.* = .done;
```
matklad, line 439: and here

src/vsr/grid_scrubber.zig
```diff
@@ -0,0 +1,481 @@
...
+                }
+            }
+
+            // Wrap around to the next cycle.
+            assert(tour.* == .done);
+            tour.* = .init;
```
matklad, line 452: maybe `log.info` here, so the operator can double-check that the scrubber scrubs.

src/vsr/replica_test.zig
```diff
@@ -1355,6 +1355,68 @@ test "Cluster: upgrade: state-sync to new release" {
...
+    }
+
+    // Disable new read/write faults so that we can use `storage.faults` to track repairs.
+    // (That is, as the scrubber runs, the number of faults will monotonically decrease.)
+    b2_storage.options.read_fault_probability = 0;
+    b2_storage.options.write_fault_probability = 0;
+
+    // Tick until B2's grid repair stops making progress.
+    // (This may take multiple scrubber cycles, since when a table index is corrupt, we skip over
+    // scrubbing its data for that cycle.)
```
matklad: I think this is not true in the current implementation? We wait for the index to get repaired before proceeding with scrubbing?

src/constants.zig
sentientwaffle: Good point! I was actually thinking about it in terms of higher `grid_iops_read_max > 1`, if we (for some reason) wanted to scrub very aggressively on an otherwise-idle cluster.

src/vsr/replica_test.zig
```diff
@@ -1355,6 +1355,68 @@ test "Cluster: upgrade: state-sync to new release" {
...
+            assert(b2_free_set.is_free(address));
+            assert(b2_storage.area_faulty(.{ .grid = .{ .address = address } }));
+        } else {
+            assert(!b2_free_set.is_free(address));
+            assert(!b2_storage.area_faulty(.{ .grid = .{ .address = address } }));
+        }
+    }
+
+    try TestReplicas.expect_equal_grid(a0, b2);
+    try TestReplicas.expect_equal_grid(b2, b2);
```
matklad: Is comparing b2 with itself intentional?

src/vsr/grid_scrubber.zig
sentientwaffle: (Per W+T discussion, this doesn't work b/c manifest blocks hold references to index blocks that may have been released and then reused by a _different_ table.)

matklad:
Finished first round of review, looks awesome!

matklad:
Ah, consider if we can assert that #blocks scrubbed = free_set_size - number of blocks freed

src/vsr/grid_scrubber.zig
sentientwaffle: Good idea -- `@sizeOf(ForestTableIterator) == 2776`, btw. I hadn't realized it was so large.

src/vsr/replica.zig
sentientwaffle: :+1: I'll change it so that `read_next()` returns a bool (whether a new read was started), and change `read_reclaim` to `read_fault() -> ?BlockReference`.

src/vsr/replica.zig
sentientwaffle: :+1: As part of this the timeout interval calculation needs to account for the number of concurrent scrub reads:

```
            self.grid_scrub_timeout.after = std.math.clamp(
                @divFloor(
                    constants.grid_scrubber_cycle_ticks,
                    @max(1, self.grid.free_set.count_acquired()),
                ) * constants.grid_scrubber_reads_max,
                constants.grid_scrubber_interval_ticks_min,
                constants.grid_scrubber_interval_ticks_max,
            );
```

src/vsr/grid_scrubber.zig
sentientwaffle: A checkpoint can change the number of trailer blocks.

So we can only assert that the `count == index` here (and for `client_sessions` below) if during `GridScrubber.checkpoint()` we skip past free_set/client_sessions in the tour. Which since it is the last state of the tour, means there would be two different code paths than can "end" a cycle.

src/vsr/grid_scrubber.zig
sentientwaffle: See https://github.com/tigerbeetle/tigerbeetle/pull/1851#discussion_r1571153653

src/vsr/replica_test.zig
sentientwaffle: You're right! The "wait for index before continue" was a last minute addition; I forgot to remove this note.

src/vsr/replica_test.zig
sentientwaffle: Nope; typo.

sentientwaffle:
> Ah, consider if we can assert that #blocks scrubbed = free_set_size - number of blocks freed

Ahh I think this doesn't work after all :disappointed: 

I was imagining something like this at the end of a cycle:

```zig
assert(scrubber.tour_blocks_scrubbed_count + scrubber.tour_blocks_released_count >=
    scrubber.forest.grid.free_set.count_acquired());
```

(where `tour_blocks_released_count` counts the cumulative number of blocks released over the course of the tour).

But the replica may both:
- acquire an arbitrary number of blocks "behind" us in the cycle (which the current cycle does not cover)
- release an arbitrary number of blocks "ahead" of us in the cycle.

src/vsr/grid_scrubber.zig
matklad: :exploding_head: indeed! That's subtle, worth a `maybe` maybe?

matklad:
>I was imagining something like this at the end of a cycle:

We could save `scrubber.forest.grid.free_set.count_acquired()` at the _start_ of the tour, rather than look at it at the end. It won't then count blocks, acquired "behind", which I think would make the assert work?

matklad:
To double check: we don't persist scrubber progress to disk, so, if replica restarts, it starts scrubbing from the beginning. So, with 90day scrubbing interval, and, say, weekly, updgrades, replicas won't actually scrub to the end. But that would be fixed by TODO: if a replica picks a random scrubbing offeset, then even with frequent restarts it'll scrub everything.

sentientwaffle:
> To double check: we don't persist scrubber progress to disk, so, if replica restarts, it starts scrubbing from the beginning. So, with 90day scrubbing interval, and, say, weekly, updgrades, replicas won't actually scrub to the end. But that would be fixed by TODO: if a replica picks a random scrubbing offeset, then even with frequent restarts it'll scrub everything.

Yes, that's correct.

sentientwaffle:
> Would be great to refresh that reference. "Why you motivate storage fault model by referencing to papers about spinning rust nobody is using anymore" is a very reasonable question to ask

Yes, unfortunately I used that reference because I couldn't find any comparable studies for SSDs.

I would like to derive the cycle interval more intelligently but I'll save that for a follow-up PR.

src/vsr/replica.zig
```diff
@@ -3078,8 +3078,10 @@ pub fn ReplicaType(
                 );
             }
 
-            const scrub_next = self.grid_scrubber.read_next();
-            maybe(scrub_next);
+            for (0..constants.grid_scrubber_reads_max + 1) |_| {
+                const scrub_next = self.grid_scrubber.read_next();
+                if (!scrub_next) break;
+            } else unreachable;
```
matklad, line 3084: Really loves this pattern with a bounded loop with an unreachable! Which, well, is what our TigerStyle tells us to use, but its only recently that I came to appreciate this fully!

src/vsr/grid_scrubber.zig
```diff
@@ -127,6 +127,9 @@ pub fn GridScrubberType(comptime Forest: type) type {
         /// Contains a table index block when tour=table_data.
         tour_index_block: BlockPtr,
 
+        /// These counters reset after every tour cycle.
+        tour_blocks_scrubbed_count: usize,
```
matklad, line 132: micronit, but let's use `u64`. Number of blocks scrubbed is a property independent of host architecture

matklad (approved):

PR #1852, opened 2024-04-18
docs: move Cloud page to Internals section

emschwartz:
- **docs: rewrap cloud doc to 100 chars**
- **docs: move cloud doc to deploy section**
- **docs: rewrap managed service doc to 100 chars**
- **docs: link to performance page from cloud page**


emschwartz:
This doc seems a little draft-y to me, but tbh I'm not sure what more we'd want to say about TigerBeetle in the cloud...

matklad:
Hm, yeah, the audience of this document are developers of TigerBeetle, not necessary the users. So I think it'd make more sense to move this to intenals?

That is, we probably do want a "cloud" doc in the deploy section, but its content should be completely different.

docs/about/internals/cloud.md
```diff
@@ -0,0 +1,59 @@
+---
+sidebar_position: 7
+---
+
+# Cloud
+
+Tigerbeetle is [optimized for performance](../about/performance.md), exploiting "close to the metal"
```
matklad: link needs adjustment

docs/about/internals/cloud.md
```diff
@@ -0,0 +1,59 @@
...
+
+We are intentionally designing TigerBeetle to repair these local storage failures automatically on a
+fine-grained basis using cluster redundancy. That’s also another reason why we’re not using Raft as
+consensus protocol, because Raft can only elect a node as leader if its entire journal is fault
+free, it cannot leverage full cluster redundancy to repair the leader’s journal (and then the rest
+of the cluster). It’s possible to have cluster failure with Raft if all nodes only have a single
+disk block fault somewhere on disk, even if they’re all at different locations. On the other hand,
+VR’s view change protocol is entirely in-memory, and doesn’t require any disk persistence for the
+protocol to be correct (as Raft’s protocol does), again much cleaner and safer in the event of disk
+faults during leader election.
```
sentientwaffle: Maybe we should delete most of this paragraph? (The first sentence is great though.)

> That’s also another reason why we’re not using Raft as consensus protocol, because Raft can only elect a node as leader if its entire journal is fault free, it cannot leverage full cluster redundancy to repair the leader’s journal (and then the rest of the cluster). It’s possible to have cluster failure with Raft if all nodes only have a single disk block fault somewhere on disk, even if they’re all at different locations.

VRR doesn't discuss repair at all iirc... And isn't the repair strategy is consensus-algorithm-agnostic anyway? I can't see any reason why you couldn't apply it to Raft just as well as VSR.

> On the other hand, VR’s view change protocol is entirely in-memory, and doesn’t require any disk persistence for the protocol to be correct (as Raft’s protocol does), again much cleaner and safer in the event of disk
faults during leader election.

Our implementation does rely on disk persistence for correctness. We update the superblock before participating in a view change to ensure that our advertised view never regresses.

docs/about/internals/cloud.md
```diff
@@ -0,0 +1,59 @@
...
+large scale server fleets, which are typically cloud native, and there’s already been [work
+done](https://www.phoronix.com/scan.php?page=news_item&px=KVM-IO-uring-Passthrough-LF2020) to share
+the host’s io_uring queues with virtualized guests.
+
+Our testing is only getting started though, we’re still building out the system end to end, so it
+will be great to benchmark more performance numbers in various environments (and across cloud
+providers) and share these as we go.
+
+Credit to @tdaly61 from the Mojaloop community for prompting us with some great questions about
+Tigerbeetle in the cloud.
```
sentientwaffle, line 52: Could also link to https://tigerbeetle.com/blog/a-friendly-abstraction-over-iouring-and-kqueue for more detail.

matklad:
@emschwartz what's the status here? Do you want to apply dj's suggesions, or is it ready to merge already? (not sure if I broke anything with my force-push last week, so checking just in case!)

sentientwaffle (approved):

PR #1854, opened 2024-04-19
cfo: add orchestrator script

matklad:
cfo.zig takes care of running the fuzzers, but someone needs to run cfo!

Long term, this _probably_ should be integrated into some kind of infrastructure as code setup, but, to start small, we'll be running cfo in a persistent tmux session. Rather then relying on systemd or some such, we'll write our own tiny "process manager" --- the only thing we really need is:

* running in a loop
* ensuring that there are no orphaned processes

and both are easy to achieve with just sh. Note that it has to be sh, rather than Zig, as this is bootstrap code that needs to be run with what we have on a machine built-in (hence, POSIX sh and not bash)

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -537,8 +537,9 @@ pub fn build(b: *std.Build) !void {
         const scripts_exe = b.addExecutable(.{
             .name = "scripts",
             .root_source_file = .{ .path = "src/scripts/main.zig" },
-            .target = target,
             .main_pkg_path = .{ .path = "src" },
+            .target = target,
+            .optimize = mode,
         });
         const scripts_run = b.addRunArtifact(scripts_exe);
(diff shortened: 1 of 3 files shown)


build.zig
```diff
@@ -537,8 +537,9 @@ pub fn build(b: *std.Build) !void {
         const scripts_exe = b.addExecutable(.{
             .name = "scripts",
             .root_source_file = .{ .path = "src/scripts/main.zig" },
-            .target = target,
             .main_pkg_path = .{ .path = "src" },
+            .target = target,
+            .optimize = mode,
```
matklad, line 542: This is just so that `-Drelease` disables debug-level logs

cb22 (approved):

PR #1855, opened 2024-04-19
devhub: show seed duration & freshness

matklad:
Should help debugging dead or slow fuzzers


matklad:
Screenshot_20240419_121653

Note: fuzzer seeds are recorded with a second resolution. I probably should have used millis, but this doesn't seem to important to fix.

cb22 (approved):

PR #1856, opened 2024-04-19
cfo: ensure seed stability

matklad:
We don't want CFO to churn through different failing seeds. To make
seeds stable, compare by (real-time) seed timestamp first.

That is, prefer older seeds, rather than numerically smaller seeds.

This was the intention from the begging, and it's just a bug that the
order is different.

Example of seed churn this should prevent: https://github.com/tigerbeetle/devhubdb/commit/f261d938b161b6d59852e2ca334ef5ee9927efdc

### DIFF
--- a/src/scripts/cfo.zig
+++ b/src/scripts/cfo.zig
@@ -301,9 +301,9 @@ const SeedRecord = struct {
     commit_sha: [40]u8,
     fuzzer: Fuzzer,
     ok: bool,
-    seed: u64,
     seed_timestamp_start: u64,
     seed_timestamp_end: u64,
+    seed: u64,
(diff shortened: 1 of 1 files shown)


src/scripts/cfo.zig
```diff
@@ -502,39 +502,39 @@ test "cfo: SeedRecord.merge" {
             \\    "commit_sha": "2222222222222222222222222222222222222222",
```
matklad, line 502: Curious that the timestapm/seed order is correct in the input for the tests. I guess I coded correct order first, and then just sorted the fields out of a habit  without noticing :rofl:

cb22 (approved):

PR #1858, opened 2024-04-20
Minor Update to java README.md

balayanv:
i believe batching example intended to actually batch not send every transfer as a batch of 1


matklad:
Oh dear, that's a good one, thanks! These docs are generated, let me fix this at the source!

matklad:
Ah, no, wait the current docs are corerct. What you are editing is the negative example of what _not_ do do. As the doc says, this'll be much slower than the batching example, which immediately follows. 

That said, the docs are worded in a confusing way, and we actually have automated batching implemented. We'll try to make this clearer soon, thanks for pointing out how confusing it is!

balayanv:
ah seems i was skimming trough it too fast )

PR #1849, opened 2024-04-16
docs: header and footer tweaks

emschwartz:
- **docs: rename Twitter links to 𝕏**
- **docs: use white logo for dark mode**
- **docs: make title use italics**
- **docs: add logo to footer with link to homepage**
- **docs: use single row of footer links**

PR #1859, opened 2024-04-21
Coalesce tables instead of compacting when possible

brson:
This patch adds table coalescing to the LSM. Prior to compacting level A of a tree into level B, it instead attempts to coalesce tables in level A such that one table in level A will be completely freed. This is assumed to be more efficient than doing the compaction.

As far as I can tell this patch is fully functional and passes all tests that I am aware of, but it has plenty of deficiencies I will describe. I think though that it is time to start getting some review.

## How the search works

The search is implemented in the second commit and is well documented. It is part of `manifest_level.zig`, with the entry point being `find_table_coalesce_window`.

This searches through windows of visible tables, the windows up to size `lsm_growth_factor`; with a limit on the maximum number of tables searched, both visible and invisible. This limit is set by `lsm_manifest_coalesce_max_tables_to_search` to 1024, but this number is not determined experimentally - it is intended to be "reasonably fast" to search (<1ms?), but I have not done any timing testing. The search never short circuits - it either searches all tables in the level (for small levels), or 1024 tables (for large levels).

On levels with more than 1024 tables it uses a deterministic random number to decide at what index to start the search; this is to spread the search out across table ranges, across compactions, but without the added bookkeeping of remembering the previously searched indexes. @jorangreef previously suggested to use a hash instead of a rng here, but I decided to stick with an rng because the rng has built-in facilities for picking a number in range with the correct statistical distribution, where a hash does not. I suspect the performance of a fast hash and a fast rng is similar, but have not done any testing. The RNG here is PCG, which is fast with good randomness.

Windows are chosen that can free up exactly one table - it never tries to find windows that free more than one table. Of those windows the best is chosen by the `choose_coalesce_window_candidate` function. This function currently just picks the window with the fewest values.

I don't really understand the I/O costs involved so am not confident that optimizing for fewest values is correct or not. We could alternately take table numbers into account, or short-circuit the search as soon as _any_ viable window is found.

The search logic is very straightforward, but it depends on a set of iterators for which the logic is somewhat complex. This is an intentional decision for the sake of making the search readable.

The `CoalasceTableIterator` and `CoalesceWindowIterator` are decently documented, and are designed to account for the cost of visiting invisible tables when limiting the amount of work done per search, and avoid revisiting invisible tables many times in pathological cases. The complexity of these iterators is unfortunate for both readability and performance, and I was reluctant to do it this way, but ultimately decided it was the cleanest way to correctly account for invisible tables. Even today I was tweaking the logic in these iterators so they deserve a close look.

Note that because iterating invisible tables counts toward the limits of the search, levels with many invisible tables will have their coalesce-window search significantly weakened. As I understand it snapshots are a theoretical feature right now, so this is just a consideration for the future.

## How integration with compaction works

It's plugged into `CompactionPipeline.beat`.

Compaction of levels is striped across half-bars. Prior to attempting to compact level b (a into b) with `bar_setup`, this function first "backs up" and checks whether it can instead coalesce level a.

When this happens it sets level "a"s `Compaction` object for coalescing (and from the view of `Compaction` and the rest of the forest code this level "a" will be seen as a level "b", with no level "a" compacting into it).

When coalescing happens, an "odd" level gets compacted in the same half-bar as other "even" levels (or vice-versa). e.g. if our half-bar is compacting levels (0, 2, 4), we might actually end up compacting (0, 1, 4) if it was determined that coalescing level 1 was better than compacting level 1 into level 2.

This mixing of odd and even compaction levels within a single half bar is the reason for the first commit of this patch that removes most of the checks for which level is allowed to be compacting during which half-bar.

This adds a new `coalesce` variant to `TableInfoA` that indicates that compaction is doing a coalesce and there is no table A.

## Limitations

I have not yet been able to test on large levels - I have only tested up to level 3. This means I haven't ever run the code path where the search randomly selects the table index to begin searching from. I don't even know how to generate a workload that builds deep enough levels to test this - I am running out of memory on large forest_fuzz workloads. I could test on a large-memory cloud VM if that's the right way to do it.

There are no new tests here. I think there should be some unit tests for the coalesce search but haven't tried to write them yet. I think there should also be an integration test that coalescing limits write amplification on small-batch workloads, but don't know how to write it. Also a smoke test that just confirms coalescing is on and doing something.

I don't know how long it takes to search 1024 tables. I think we should have some idea of the CPU time of this operation.

I don't know how often this optimization is triggered on real workloads, but suspect that there are workloads where it is ineffective, and that it may be more or less effective on different trees. It may be useful to completely avoid this search in some situations, and I have indicated this with a todo, but don't have enough evidence to know when or how to do it yet.

(The benchmark workload is ineffective at triggering this optimization, but the forest_fuzz triggers it at all levels).

Coalesced compaction objects never set `drop_tombstones` to `true`. I am not sure the impact of this. I note that coalescing always follows the "copy" path in the compactor, never merge, move, or, "copy_drop_tombstones".

Coalesce is never performed for the last level. This is mostly just because the structure of the compaction setup code doesn't lend itself easily to coalescing the last level.

This adds a new constant, `value_count_max_actual`, which contains the maximum number of values allowed in a particular table. Contrast this with the existing `value_count_max` constant, which is sometimes a lower value. I don't really understand the logic around these two constants but the dual names are quite confusing and I'm open to suggestions about how to rename one of them.

### DIFF
--- a/src/lsm/compaction.zig
+++ b/src/lsm/compaction.zig
@@ -70,7 +70,7 @@ const half_bar_beat_count = @divExact(constants.lsm_batch_multiple, 2);
 pub const CompactionInfo = struct {
     /// How many values, across all input tables, need to be processed.
     /// This is the total – it is fixed for the duration of the compaction.
-    compaction_tables_value_count: usize,
+    compaction_tables_value_count: u64,
 
     // Keys are integers in TigerBeetle, with a maximum size of u256. Store these
     // here, instead of Key, to keep this unspecialized.
@@ -569,7 +569,7 @@ pub fn CompactionType(
                 .tables = tables.tables,
             };
 
-            var compaction_tables_value_count: usize = 0;
+            var compaction_tables_value_count: u64 = 0;
             for (range_b.tables.const_slice()) |*table| {
                 compaction_tables_value_count += table.table_info.value_count;
             }
@@ -644,7 +644,7 @@ pub fn CompactionType(
                     table_immutable_values_count,
                 });
 
-                var compaction_tables_value_count: usize = table_immutable_values_count;
+                var compaction_tables_value_count: u64 = table_immutable_values_count;
                 for (range_b.tables.const_slice()) |*table| {
                     compaction_tables_value_count += table.table_info.value_count;
                 }
@@ -687,7 +687,7 @@ pub fn CompactionType(
                     compaction.level_b,
                 });
 
-                var compaction_tables_value_count: usize = table_a.value_count;
+                var compaction_tables_value_count: u64 = table_a.value_count;
                 for (range_b.tables.const_slice()) |*table| {
                     compaction_tables_value_count += table.table_info.value_count;
                 }
--- a/src/lsm/compaction.zig
+++ b/src/lsm/compaction.zig
@@ -2079,6 +2079,11 @@ pub fn CompactionType(
                 );
             }
 
+            // Coalescing should always result in a free table.
+            if (bar.table_info_a == .coalesce) {
+                manifest.assert_level_free_table(level_b);
+            }
+
             // Our bar is done!
             compaction.bar = null;
         }
--- a/src/lsm/compaction.zig
+++ b/src/lsm/compaction.zig
@@ -265,6 +265,7 @@ pub fn CompactionType(
         const tombstone = Table.tombstone;
 
         const TableInfoA = union(enum) {
+            coalesce,
             immutable: []Value,
             disk: TableInfoReference,
         };
@@ -551,6 +552,57 @@ pub fn CompactionType(
         pub fn assert_between_bars(compaction: *const Compaction) void {
             assert(compaction.bar == null);
             assert(compaction.beat == null);
+        }
+
+        pub fn maybe_coalesce(compaction: *Compaction, tree: *Tree, op: u64) ?CompactionInfo {
+            const level = compaction.level_b;
+
+            const prng_seed = tree.compaction_op.?;
+            const manifest_level = tree.manifest.levels[level];
+            const tables = manifest_level.find_table_coalesce_window(
+                prng_seed,
+            ) orelse return null;
+
+            const range_b = Manifest.CompactionRange{
+                .key_min = tables.tables.get(0).table_info.key_min,
+                .key_max = tables.tables.get(tables.tables.count() - 1).table_info.key_max,
+                .tables = tables.tables,
+            };
+
+            var compaction_tables_value_count: usize = 0;
+            for (range_b.tables.const_slice()) |*table| {
+                compaction_tables_value_count += table.table_info.value_count;
+            }
+
+            std.log.debug("found coalesce window in {s}, level {}, tables {}/{}, values {}", .{
+                tree.config.name,              level,
+                tables.tables.count(),         tree.manifest.levels[level].tables.len(),
+                compaction_tables_value_count,
+            });
+
+            compaction.bar = .{
+                .tree = tree,
+                .op_min = compaction_op_min(op),
+
+                .move_table = false,
+                .table_info_a = .coalesce,
+                .range_b = range_b,
+                .drop_tombstones = false,
+
+                .compaction_tables_value_count = compaction_tables_value_count,
+
+                .target_index_blocks = null,
+                .beats_max = null,
+            };
+
+            return .{
+                .compaction_tables_value_count = compaction_tables_value_count,
+                .target_key_min = compaction.bar.?.range_b.key_min,
+                .target_key_max = compaction.bar.?.range_b.key_max,
+                .move_table = compaction.bar.?.move_table,
+                .tree_id = tree.config.id,
+                .level_b = level,
+            };
         }
 
         /// Perform the bar-wise setup, and returns the compaction work that needs to be done for
@@ -901,6 +953,9 @@ pub fn CompactionType(
                 },
                 .immutable => {
                     // Immutable values come from the in memory immutable table - no need to read.
+                },
+                .coalesce => {
+                    // Coalescing doesn't require reading from level_a.
                 },
             }
 
@@ -1314,7 +1369,9 @@ pub fn CompactionType(
                 // Sanity check our primary condition.
                 assert(bar.source_values_merge_count == bar.compaction_tables_value_count);
 
-                if (bar.table_info_a == .immutable) {
+                if (bar.table_info_a == .coalesce) {
+                    // Nothing to do
+                } else if (bar.table_info_a == .immutable) {
                     assert(bar.table_info_a.immutable.len == 0);
                 } else {
                     assert(blocks.source_value_blocks[0].ready.count == 0);
@@ -1336,7 +1393,10 @@ pub fn CompactionType(
             const bar = &compaction.bar.?;
             const beat = &compaction.beat.?;
 
-            if (bar.table_info_a == .immutable) {
+            if (bar.table_info_a == .coalesce) {
+                beat.source_a_values = null;
+                return .exhausted;
+            } else if (bar.table_info_a == .immutable) {
                 // Immutable table can never .need_read, since all its values come from memory.
                 assert(bar.source_a_immutable_block != null);
                 stdx.maybe(bar.source_a_immutable_values == null);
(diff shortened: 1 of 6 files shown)


brson:
I pushed an additional commit that asserts that coalescing resulted in a free table, and I'm interested in other assertions I should be making.

src/lsm/manifest_level.zig
```diff
@@ -729,6 +734,336 @@ pub fn ManifestLevelType(
...
+        ///
+        /// For large levels, the index to begin iteration is chosen
+        /// pseudo-randomly but deterministically, seeded by the caller with a
+        /// number that is unique per compaction and deterministic across
+        /// replicas.
+        pub fn find_table_coalesce_window(
+            level: *const Self,
+            prng_seed: u64,
+        ) ?TableCoalesceWindow {
+            const num_tables = level.tables.len();
```
matklad: Our standard naming here is `table_count`

* `count` is the dimensionless number of things
* `index` is dimensionless index. `count` is inclusive, `index` is exclusive, `index` and `count` are a name-pair
* `size` is the the size in bytes, `count` multiplied by `@sizeOf`
* we generally don't use `len` or `length`

src/lsm/compaction.zig
```diff
@@ -553,6 +554,57 @@ pub fn CompactionType(
...
+
+            var compaction_tables_value_count: usize = 0;
+            for (range_b.tables.const_slice()) |*table| {
+                compaction_tables_value_count += table.table_info.value_count;
+            }
+
+            std.log.debug("found coalesce window in {s}, level {}, tables {}/{}, values {}", .{
+                tree.config.name,              level,
+                tables.tables.count(),         tree.manifest.levels[level].tables.len(),
+                compaction_tables_value_count,
```
matklad, line 580: :exploding_head: never realized that this is an option for formatting arguments to printf, thanks!

src/lsm/compaction.zig
```diff
@@ -553,6 +554,57 @@ pub fn CompactionType(
...
+                prng_seed,
+            ) orelse return null;
+
+            const range_b = Manifest.CompactionRange{
+                .key_min = tables.tables.get(0).table_info.key_min,
+                .key_max = tables.tables.get(tables.tables.count() - 1).table_info.key_max,
+                .tables = tables.tables,
+            };
+
+            var compaction_tables_value_count: usize = 0;
```
matklad: Maybe `u64` rather than `usize`? The numbe of values in the table is a quantity independent from the underlying architecture. 

In general, because everything is explicitly limited, very few things in TigerBeetle need to be an `usize`. I.e., while in Rust you might use `usize` for something indexing an array, for us the size of array is known, and usually we can say that `u32` would be enough (and Zig allows you to index with `u32`).

Also, a bit of trivia in case you've missed it: unlike Rust and like C, in Zig `usize` is an alias of `u32` or `u64`, rather than a distinct type.

src/lsm/forest.zig
```diff
@@ -993,20 +992,37 @@ fn CompactionPipelineType(comptime Forest: type, comptime Grid: type) type {
...
+                                        level_b,
+                                        tree.config.name,
+                                        op,
+                                    });
+                                    do_compaction = false;
+                                }
+                            }
+
+                            if (do_compaction) {
+                                var compaction = &tree.compactions[level_b];
```
matklad: probably needs to be `const`? You are taking a pointer here, so mutaiblity of the var itself doesn't matter (`let mut compation: &mut _`). 

We also in general try to avoid creating local aliases, and would prefer to inline `tree.compactions[level_b]` I think, but that's a super fuzzy rule at this point.

src/lsm/manifest_level.zig
```diff
@@ -729,6 +734,336 @@ pub fn ManifestLevelType(
...
             return range;
         }
+
+        const TableCoalesceWindow = struct {
+            tables: stdx.BoundedArray(TableInfoReference, max_coalesce_window_size),
+        };
+
+        const TableCoalesceWindowCandidate = struct {
+            tables: stdx.BoundedArray(*const TableInfo, max_coalesce_window_size),
+            total_values: u64,
```
matklad: `value_count_total`. We use un-English like big endian naming scheme, where we just slap extra qualifiers to the end. Each `TableInfo` has `value_count`, so this would be `value_cont_total`.

src/lsm/manifest_level.zig
```diff
@@ -729,6 +734,336 @@ pub fn ManifestLevelType(
...
+                }
+            }
+
+            // todo - this is a hot function and might profit from memcpy optimization
+            fn pop_buffer_queue(it: *CoalesceTableIterator) ?*const TableInfo {
+                if (it.buffer.empty()) {
+                    return null;
+                }
+
+                const retval = it.buffer.get(0);
```
matklad: We use `result` to name return variables

src/lsm/manifest_level.zig
```diff
@@ -729,6 +734,336 @@ pub fn ManifestLevelType(
...
+
+                var index: usize = 0;
+                while (index < it.buffer.count() - 1) {
+                    it.buffer.slice()[index] = it.buffer.get(index + 1);
+                    index += 1;
+                }
```
matklad: This could be a for loop unless I am missing something?

```
for (it.buffer.slice()) |*element, index|
```

But, yeah, it feels like this could use `std.mem.copyLeft` or something

matklad (commented):
Did a surface style review to align naming and such!

cb22:
I'm still looking through this properly, but comparing with the benchmark client set to send in batches of 10 for 100k transactions, shows a massive space amp win from 8709MiB to 2332MiB!

sentientwaffle:
The value for `lsm_manifest_coalesce_max_tables_to_search` should originate at `ConfigCluster` since (iiuc) its choice ultimately impacts data layout on disk.

> I have not yet been able to test on large levels - I have only tested up to level 3. This means I haven't ever run the code path where the search randomly selects the table index to begin searching from.

...Plus then you can pick a smaller value for the `test_min` config, which means the fuzzer will be able to hit interesting cases in a reasonable about of time/RAM.

---

> (The benchmark workload is ineffective at triggering this optimization, but the forest_fuzz triggers it at all levels).

Is this because the benchmark creates full batches, which never need to be coalesced?

src/lsm/compaction.zig
brson: seems to just be what zig fmt wants to do in this situation

brson:
> > (The benchmark workload is ineffective at triggering this optimization, but the forest_fuzz triggers it at all levels).
> 
> Is this because the benchmark creates full batches, which never need to be coalesced?

Yes.

src/lsm/compaction.zig
matklad: No, you asked it to do this! :) It's the same trick as with trailing comma to get multi-line layout, see:

https://github.com/tigerbeetle/tigerbeetle/pull/1849#issuecomment-2060818331

src/lsm/compaction.zig
brson: I appear to have copied this code from similar code in bar_setup that also uses a `usize`. Should I change both?

brson: It's also a `usize` in _one_ of the structs in this file, but a `u64` in another struct.

In general, I am finding that zig's integer promotion, while safe, makes it hard for me to know what type any particular integer is, and what type I should declare various ints that flow into other ints.

src/lsm/forest.zig
brson: Seems like it should be. Interesting the compiler doesn't have an opinion about this.

src/lsm/manifest_level.zig
brson: Done.

src/lsm/forest.zig
brson: Done.

src/lsm/manifest_level.zig
brson: Done.

src/lsm/manifest_level.zig
brson: I did not know `for` could provide an index, though in this case it needs to look more like

```
                for (it.buffer.slice()[0 .. it.buffer.count() - 1], 0..) |*element, index| {
                    element.* = it.buffer.get(index + 1);
                }
```

Since I'm not iterating all elements, and the index range needs to supplied to `for`.

I've made this change.

src/lsm/manifest_level.zig
brson: Done.

src/lsm/compaction.zig
brson: I made this variable a u64 in four places where it was previously usize.

brson:
I've resolved @matklad's reviews. Thanks! Have not resolved others yet.

brson:
We're going to try another approach to solving this problem.

PR #1864, opened 2024-04-22
devhub: show one failing seed per fuzzer

matklad:
Currently, CFO doesn't have the logic to carry fuzz failures from old commits to new commits automatically. So, rather than showing all failures for the latest commit, lets show the first (latest) failure for each fuzzer.

As a nice side-effect, that will give us only one entry for canary.


matklad:
Screenshot_20240422_160311

cb22 (approved):

PR #1865, opened 2024-04-22
ewah: fix off-by-one in fuzzer's code

matklad:
If decoded_size is `0`, we try to generate then 0 bits, which panics. Furthermore, our encoding/decoding assumes that the size is non zero (a zero-sized disk write is slower than no write at all).

Make sure that the input buffer is at least one byte long. Note that the number of bits actually set is allowed to be zero: ewah is symmetric between 0 and one, and all bits set is a reasonable configuration.

Seed: ./zig/zig build -Drelease fuzz -- --seed=1459624380881812643 ewah

This is the first CFO trophy

### DIFF
--- a/src/ewah_fuzz.zig
+++ b/src/ewah_fuzz.zig
@@ -15,19 +15,19 @@ pub fn main(args: fuzz.FuzzArgs) !void {
         const random = prng.random();
 
         const decoded_size_max = @divExact(1024 * 1024, @sizeOf(Word));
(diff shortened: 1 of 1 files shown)


matklad:
This is how the failure would look on devhub, [once it is patched to actually show it](https://github.com/tigerbeetle/tigerbeetle/pull/1864#issuecomment-2069797493) :sweat_smile: 

image

sentientwaffle (approved):

PR #1874, opened 2024-04-24
tidy: blob check vs shallow clone

matklad:
We want to avoid large files in the repository. Naively listing just the files won't work: if the file is added in one commit and removed in a subsequent one, it will be absent from the working tree, but will still be perpetually present in the Git history.

For this reason, tidy is smart --- it asks git itself to list large git objects.

But GitHub outsmarts tidy, as it does a shallow clone by default.

Fix this treachery within treachery within treachery with a proper clone

See https://stackoverflow.com/a/6802238/1936422 for why this specific depth.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -18,6 +18,9 @@ jobs:
       # https://github.com/actions/checkout/issues/1169
       - run: git config --system --add safe.directory '*'
       - uses: actions/checkout@v4
+        # Fetch the entire history of the commit in question for tidy.
+        with:
+          fetch-depth: 2147483647
       - run: ./scripts/install_zig.sh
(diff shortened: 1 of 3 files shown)


sentientwaffle (dismissed):

matklad:
Double-checked that it actually works:

https://github.com/tigerbeetle/tigerbeetle/actions/runs/8821492655/job/24217470840

sentientwaffle (approved):
LGTM! Apologies for the delay; I didn't notice it needed to be refreshed.

matklad:
If you hadn't delayed the review, I wouldn't have realized that I can actually check this by adding large file and reverting it in this branch. So, no, you failed to delay unproductively!

PR #1875, opened 2024-04-24
lsm: speed up tree fuzzer

matklad:
Currently it timeouts on CFO. I am not sure this'll be enough to get it
to pass, but it seems a nice change regradless, as it speeds up the
fuzzing 2X for me

./zig/zig build -Drelease fuzz --
    --seed=3167000064919571166 --events-max=100000 lsm_tree

info(fuzz): done in 16.707s

VS

info(fuzz): done in 35.953s

The problematic path here is scans, where our data structure under test
is an LSM tree, and our model is a hash map. Hash maps can't do scans!

The proper fix here would be to use a BTree as a model, but Zig's stdlib
doesn't have a BTree.

So instead, we optimize the existing code, making use of the following
observation:

Our result set is limited, so most of the keys from the model do not
belong to the result set (even if they belong to the original scan range).

So, instead of binary searching every time, do a quick check for the
value being to the left/to the right of the result set, and fallback to
binary search only when the value is in bounds.

### DIFF
--- a/src/lsm/tree_fuzz.zig
+++ b/src/lsm/tree_fuzz.zig
@@ -68,6 +68,12 @@ const Value = packed struct(u128) {
 
 const FuzzOpTag = std.meta.Tag(FuzzOp);
 const FuzzOp = union(enum) {
+    const Scan = struct {
+        min: u64,
+        max: u64,
+        direction: Direction,
+    };
+
     compact: struct {
         op: u64,
         checkpoint: bool,
@@ -75,11 +81,7 @@ const FuzzOp = union(enum) {
     put: Value,
     remove: Value,
     get: u64,
-    scan: struct {
-        min: u64,
-        max: u64,
-        direction: Direction,
-    },
+    scan: Scan,
 };
 
 const batch_size_max = constants.message_size_max - @sizeOf(vsr.Header);
--- a/src/lsm/tree_fuzz.zig
+++ b/src/lsm/tree_fuzz.zig
@@ -642,56 +642,48 @@ fn EnvironmentType(comptime table_usage: TableUsage) type {
 
             // Asserting the negative space:
             // All keys existing in our model must be checked against the scan range.
-            if (scan_range.direction == .descending) std.mem.sort(
+            if (scan_range.direction == .descending) std.mem.reverse(
                 Value,
                 tree_values,
-                {},
-                struct {
-                    fn sort(_: void, a: Value, b: Value) bool {
-                        return Table.key_from_value(&a) <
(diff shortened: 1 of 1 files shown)


matklad:
Actually, it might as well fix that CFO failure!

```
λ ./zig/zig build -Drelease fuzz -- --seed=3167000064919571166 lsm_tree
info(fuzz): Fuzz seed = 3167000064919571166
info(lsm_tree_fuzz): table_usage=lsm.table.TableUsage.secondary_index
info(lsm_tree_fuzz): fuzz_op_count = 1023796
info(lsm_tree_fuzz): fuzz_op_distribution = enums.EnumFieldStruct(@typeInfo(lsm.tree_fuzz.FuzzOp).Union.tag_type.?,f64,null){ .compact = 0.00e+00, .put = 8.00e+00, .remove = 0.00e+00, .get = 0.00e+00, .scan = 4.00e+00 }
info(lsm_tree_fuzz): puts_since_compact_max = 120
info(lsm_tree_fuzz): compacts_per_checkpoint = 8
info(lsm_tree_fuzz): Passed!
info(fuzz): done in 12m9.417s
```

That's within our 30min budget!

src/lsm/tree_fuzz.zig
```diff
@@ -581,111 +585,105 @@ fn EnvironmentType(comptime table_usage: TableUsage) type {
...
+                        },
+                    }
+                } else {
+                    assert(Table.tombstone(&tree_value));
+                }
+            }
+
+            // Asserting the negative space:
+            // All keys existing in our model must be checked against the scan range.
+            if (scan_range.direction == .descending) std.mem.reverse(
```
matklad, line 645: Swaped a simple reverse for sorting

src/lsm/tree_fuzz.zig
```diff
@@ -581,111 +585,105 @@ fn EnvironmentType(comptime table_usage: TableUsage) type {
...
-                                            model_value_key <
-                                                Table.key_from_value(&tree_values[0]),
-                                        ),
-                                    }
-                                }
-                            } else {
-                                // Must not be found:
-                                if (value_maybe) |value| {
-                                    // Or it's a tombstone.
-                                    assert(Table.tombstone(value));
```
matklad, line 684: I _think_ this is dead code? if the value is outside of range, it can't be found even as a tombsotne

batiati (approved):
That's great!

matklad:
>Actually, it might as well fix that CFO failure!

No worries, CFO promptly found a seed that gives 4x more op count :rofl:

PR #1878, opened 2024-04-25
Update Zig from 0.11.0 to 0.13.0 

kprotty:
Upgrading over an upgrade.


matklad:
@kprotty you might want to rebase your commits on too of the main branch —- I had to force push a couple of comments on the tip to get rid of a large binary file accidentally committed!

src/lsm/compaction.zig
```diff
@@ -1813,9 +1813,9 @@ pub fn CompactionType(
 
         fn blip_write_next_tick(next_tick: *Grid.NextTick) void {
             const write: *?Beat.Write = @ptrCast(
-                @fieldParentPtr(Beat.Write, "next_tick", next_tick),
+                @as(*Beat.Write, @fieldParentPtr("next_tick", next_tick)),
```
matklad, line 1827: :thinking: @cb22 is the original code correct? It casts `*Write` to `*?Write` for the purpose of `fieldParentPtr`, but I think that cast might be invalid? In general, sizeOf(Write) might be unequal to `sizeOf(?Write)`

cb22: @matklad ha, this was on my mind too recently with Rafael's post in #zig.

It took me a second to figure out WTF was going on here! If we try out a simple test, it panics as you'd expect, because with the current Optional implementation ([not defined!](https://github.com/ziglang/zig/issues/104#issuecomment-824429303)) casting a `*Example` to a `*?Example` and trying to use it results in:

```zig
var example = Example{ .foo = 32 };
var maybe_example: ?Example = example;
var example_pointer: *Example = &example;
var example_maybe_bad_cast: *?Example = @ptrCast(example_pointer);

std.log.info("example={any} maybe_example={any} example_pointer_deref={any} example_maybe_bad_cast_deref={any}", .{
    std.mem.asBytes(&example).*,
    std.mem.asBytes(&maybe_example).*,
    std.mem.asBytes(example_pointer).*,
    std.mem.asBytes(example_maybe_bad_cast).*,
});

var val = example_maybe_bad_cast.*.?;
```

```
info: example={ 32, 0, 0, 0 } maybe_example={ 32, 0, 0, 0, 1, 0, 0, 0 } example_pointer_deref={ 32, 0, 0, 0 } example_maybe_bad_cast_deref={ 32, 0, 0, 0, 0, 0, 0, 0 }

thread 941320 panic: attempt to use null value
/tmp/test.zig:39:39: 0x21ea26 in main (test)
    var val = example_maybe_bad_cast.*.?;
```

But this code is in a hot path and does work. It's because in the parent struct its an optional, so here when its cast back, the bits happen to align because in the current implementation the T is first, and it works:

```zig
const Parent = struct {
    example: ?Example,
};

var parent = Parent{ .example = example };
var parent_example_cast: *?Example = @ptrCast(&parent.example.?);

std.log.info("parent={any} parent_example_cast_deref={any}", .{
    std.mem.asBytes(&parent).*,
    std.mem.asBytes(parent_example_cast).*,
});

var val = parent_example_cast.*.?;
```

```
info: parent={ 32, 0, 0, 0, 1, 0, 0, 0 } parent_example_cast_deref={ 32, 0, 0, 0, 1, 0, 0, 0 }
```

(all a very long way of saying it works, but we should definitely change it)

batiati: The "clean" solution is replacing the optional with an union, keeping the same functionality since Zig supports fieldParentPtr in union fields.

I think we should mark the code with a TODO(zig) because it's likely that optionals would be supported in fieldParentPtr some day.

matklad: I wonder if there's some clever way to refactor this whole thing away? One idea I was toying this is for compaction pipeline, instead of immediately poking the forest on completion on every blip, is to just enqueue completed blips to some sort of a queue of completed things.

That way, the code in forest could do:

```
whlile (get_next_completed_blip) |blip| {
    schedule more work
}
```

and that way we don't even need `on_next_tick` (any work scheduled during the loop gets run on a subsequent iteration of the loop).

We'll still need a callback after read/write to make the forest to kick off new jobs, but that callback can be siplimfied to "check if there are any completions" instead of the current "this specific thing was completed".

src/io/darwin.zig
```diff
@@ -704,7 +704,7 @@ pub const IO = struct {
                 log.info("creating \"{s}\"...", .{relative_path});
             },
             .create_or_open => {
-                flags.CREATE = true;
+                flags.CREAT = true;
```
matklad, line 708: :rofl:

kprotty: posix moment

src/vsr/superblock.zig
```diff
@@ -1090,7 +1090,7 @@ pub fn SuperBlockType(comptime Storage: type) type {
         }
 
         fn write_header_callback(write: *Storage.Write) void {
-            const context: *Context = @fieldParentPtr("write", write);
+            const context: *Context = @alignCast(@fieldParentPtr("write", write));
```
matklad, line 1093: Explaning this change for myself: in Zig 0.12, Zig no longer assumes that the alignment would magically work out.

That is, if you use `fieldParentPtr` wrong, you of course can get a pointer to garbage. However, it is also possible to get a pointer with a wrong alignment, which is somewhat more severe (it makes the pointer itself invalid). So, Zig needs some sort of safety check here, and this `@alignCast` is where this safety check happens.

src/vsr/free_set.zig
```diff
@@ -835,7 +835,7 @@ test "FreeSet encode, decode, encode" {
 
     // Random.
     var seed: u64 = undefined;
-    try std.os.getrandom(mem.asBytes(&seed));
```
matklad, line 838: In the follow up, we'll need to change this to crypto.Random

src/io/linux.zig
```diff
@@ -1438,7 +1434,7 @@ pub const IO = struct {
 
         while (true) {
             const res = stdx.fstatfs(dir_fd, &statfs);
-            switch (os.linux.getErrno(res)) {
```
matklad, line 1435: I belive this is wrong. The posix `errno` either looks at the libc threadlocal, or does the normal integer logic. If this code ends up being used when we also link libc, I think what would happen is that it'll use thread-local errno, which won't actually be set as we are doing syscalls directly here. 

I think this needs to be `os.linux.E.init`

Also: libc errno is just such a terrible interface, especially on Linux. Syscall returns you a perfectly fine error code, and then you go and stuff those exquisite bits into a thread local! Barbarity! 

Luckily, we don't use libc :o)

matklad: I am not sure where we should be using `posix.errno` on Darwin. I think yes, because we do link libc on Darwin because there's no other way?

src/time.zig
```diff
@@ -69,8 +69,8 @@ pub const Time = struct {
         // https://developer.apple.com/documentation/kernel/1646199-mach_continuous_time
         // https://opensource.apple.com/source/Libc/Libc-1158.1.2/gen/clock_gettime.c.auto.html
         const darwin = struct {
-            const mach_timebase_info_t = os.darwin.mach_timebase_info_data;
-            extern "c" fn mach_timebase_info(info: *mach_timebase_info_t) os.darwin.kern_return_t;
+            const mach_timebase_info_t = std.posix.system.mach_timebase_info_data;
```
matklad: maybe just `std.system.darwing`? Feels weired to get this out of posix

src/clients/c/tb_client_header.zig
```diff
@@ -64,13 +64,13 @@ fn resolve_c_type(comptime Type: type) []const u8 {
     }
 }
 
-fn to_uppercase(comptime input: []const u8) []const u8 {
+fn to_uppercase(comptime input: []const u8) [input.len]u8 {
```
matklad, line 67: :muscle:

src/stdx.zig
```diff
@@ -733,3 +733,9 @@ pub fn EnumUnionType(
         .tag_type = Enum,
     } });
 }
+
+/// Creates a slice to a comptime slice without triggering
+/// `error: runtime value contains reference to comptime var`
+pub fn comptime_slice(comptime slice: anytype, comptime len: usize) []const @TypeOf(slice[0]) {
```
matklad, line 739: :muscle: :muscle: :muscle:

src/io/linux.zig
```diff
@@ -437,7 +436,7 @@ pub const IO = struct {
                                 .OPNOTSUPP => error.FileLocksNotSupported,
                                 .AGAIN => error.WouldBlock,
                                 .TXTBSY => error.FileBusy,
-                                else => |errno| os.unexpectedErrno(errno),
+                                else => |errno| posix.unexpectedErrno(errno),
```
matklad, line 438: This one is fine, as is, it doesn't use errno

build.zig
```diff
@@ -119,7 +119,7 @@ pub fn build(b: *std.Build) !void {
 
     const vsr_options_module = options.createModule();
     const vsr_module = b.addModule("vsr", .{
-        .root_source_file = .{ .path = "src/vsr.zig" },
+        .root_source_file = .{ .src_path = .{ .owner = b, .sub_path = "src/vsr.zig" } },
```
matklad: IN a fololw up, we'll need to convert these to just `b.path()`

src/tigerbeetle/main.zig
```diff
@@ -37,17 +37,16 @@ pub const std_options = .{
...
-    if (builtin.target.os.tag == .macos) {
-        std.posix.setrlimit(std.posix.rlimit_resource.STACK, .{
-            .cur = 16 * 1024 * 1024,
-            .max = 16 * 1024 * 1024,
-        }) catch @panic("unable to adjust stack limit");
-    }
+    // in build.zig doesn't work. setrlimit with 16MB errors with EINVAL on macOS 13 as well.
+    // So instead, we spawn a thread with the desired stack size instead.
+    const thread = try std.Thread.spawn(.{ .stack_size = 16 * 1024 * 1024 }, main_real, .{});
+    thread.join();
```
matklad, line 45: I guess the real fix would be to finally fix that tail recursion in prefetch :rofl:

matklad (commented):
Code changes look good to me with the exception of std.posix.errno which I think is wrong. I'll still have to do the runtime testing here, but I'll aim to land this on monday.

Thanks a lot for doing this upgrade, much needed work!

build.zig
matklad: Ah, this actually happens in a latter commit, :+1:

src/tigerbeetle/main.zig
batiati: I think we already fixed it :thinking: 

https://github.com/tigerbeetle/tigerbeetle/pull/1054#discussion_r1430322968

matklad: :thinking: Ok, let's leave it as is, but then check if we could just remove this code after the upgrade is done.

In my mental model,  we _shouldn't_ be using all that much of a stack, so the stack size _shouldn't_ matter, but maybe I am wrong!

matklad:
Checked that, for the same seed, this version produces byte-for-byte identical data file. This required a bit of patching of std.random to get back the old behavior (the algoritm for random numbes in ranged changed for small-sized ints like u8), but after that it is the same :tada:

matklad (approved):
:shipit:

PR #1879, opened 2024-04-25
MessagePool: Remove old assert

sentientwaffle:
`vsr.Client` doesn't depend on `constants.client_request_queue_max`.


src/message_pool.zig
```diff
@@ -60,7 +60,6 @@ pub const messages_max_client = messages_max: {
 comptime {
     // These conditions are necessary (but not sufficient) to prevent deadlocks.
     assert(messages_max_replica > constants.replicas_max);
-    assert(messages_max_client > constants.client_request_queue_max);
```
kprotty, line 63: Should this assert be `> 1` instead given there can be one inflight?

kprotty (approved):

PR #1880, opened 2024-04-25
Constants: Reduce connections_max by 1

sentientwaffle:
Since `MessagePool` never has a connection to itself.

Saves a whopping 5MiB of RAM :stuck_out_tongue: 
(This is a "bug" I found while working on a more impactful change -- to not allocate messages for extra connections when `replica_count < replica_count_max`.)


matklad (approved):
:smile:

PR #1882, opened 2024-04-26
CLI: Print memory usage more accurately

sentientwaffle:
More accurately print memory usage at startup.

Before:

```
info(main): 0: Allocated 4255MiB in 16 regions during replica init
```

After:

```
info(main): 0: Allocated 3404MiB in 16 regions during replica init
```

To be very clear, the actual memory allocated has not changed. Before we were effectively printing the virtual memory – over-allocation by the arena allocator which will never be page-faulted in. The new number is _closer_ to RSS (though still overestimates it – arena doesn't store enough info for us to be more accurate).

### DIFF
--- a/src/counting_allocator.zig
+++ b/src/counting_allocator.zig
@@ -1,0 +1,44 @@
+const std = @import("std");
+const assert = std.debug.assert;
+
+const Self = @This();
+
+parent_allocator: std.mem.Allocator,
+size: usize = 0,
+
+pub fn init(parent_allocator: std.mem.Allocator) Self {
+    return .{ .parent_allocator = parent_allocator };
+}
+
+pub fn deinit(self: *Self) void {
+    self.* = undefined;
+}
+
+pub fn allocator(self: *Self) std.mem.Allocator {
+    return .{
+        .ptr = self,
+        .vtable = &.{
+            .alloc = alloc,
+            .resize = resize,
+            .free = free,
+        },
+    };
+}
+
+fn alloc(ctx: *anyopaque, len: usize, ptr_align: u8, ret_addr: usize) ?[*]u8 {
+    const self: *Self = @alignCast(@ptrCast(ctx));
+    self.size += len;
+    return self.parent_allocator.rawAlloc(len, ptr_align, ret_addr);
+}
+
+fn resize(ctx: *anyopaque, buf: []u8, buf_align: u8, new_len: usize, ret_addr: usize) bool {
+    const self: *Self = @alignCast(@ptrCast(ctx));
+    self.size = (self.size - buf.len) + new_len;
+    return self.parent_allocator.rawResize(buf, buf_align, new_len, ret_addr);
+}
+
+fn free(ctx: *anyopaque, buf: []u8, buf_align: u8, ret_addr: usize) void {
+    const self: *Self = @alignCast(@ptrCast(ctx));
+    self.size -= buf.len;
+    return self.parent_allocator.rawFree(buf, buf_align, ret_addr);
+}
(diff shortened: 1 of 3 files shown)


matklad:
Alternatively, maybe write a small allocator wrapper that tracks the total number of bytes allocated? 

This exact approach feels like a half-measure -- an improvement, but not a principled solution. 

Tracking allocated bytes directly is of course incorrect (as what you are interested in are pages), but you can get the pages by looking at the RSS anyway, and having both numbers, external and internal memory usage, should give the most info.

src/counting_allocator.zig
```diff
@@ -0,0 +1,43 @@
...
+
+fn alloc(ctx: *anyopaque, len: usize, ptr_align: u8, ret_addr: usize) ?[*]u8 {
+    const self: *Self = @alignCast(@ptrCast(ctx));
+    self.size += len;
+    return self.parent_allocator.rawAlloc(len, ptr_align, ret_addr);
+}
+
+fn resize(ctx: *anyopaque, buf: []u8, buf_align: u8, new_len: usize, ret_addr: usize) bool {
+    const self: *Self = @alignCast(@ptrCast(ctx));
+    self.size += new_len - buf.len;
```
matklad: `self.size = (self.size - buf.len) + new_len`

resize can either grow or shrink

src/counting_allocator.zig
```diff
@@ -0,0 +1,43 @@
...
+    return self.parent_allocator.rawResize(buf, buf_align, new_len, ret_addr);
+}
+
+fn free(ctx: *anyopaque, buf: []u8, buf_align: u8, ret_addr: usize) void {
+    const self: *Self = @alignCast(@ptrCast(ctx));
```
matklad, line 41: sWe _probably_ want to `-=` here? I am no sure if some of our internal structs don't do internal reallocations?

src/tigerbeetle/main.zig
```diff
@@ -163,10 +163,12 @@ const Command = struct {
     }
 
     pub fn start(arena: *std.heap.ArenaAllocator, args: *const cli.Command.Start) !void {
```
matklad: Might change `ArenaAllocator` just for `Allocator` here? We needed arena to print stats, but we now print stats without arena. And I think we can drop "in N regions", as that's about the impl detail of the arena, rather than about our own code.

src/counting_allocator.zig
sentientwaffle: Ah, great point! I forgot the resize could shrink.

src/tigerbeetle/main.zig
```diff
@@ -162,11 +162,13 @@ const Command = struct {
...
-    pub fn start(arena: *std.heap.ArenaAllocator, args: *const cli.Command.Start) !void {
+    pub fn start(base_allocator: std.mem.Allocator, args: *const cli.Command.Start) !void {
+        var counting_allocator = vsr.CountingAllocator.init(base_allocator);
+
         var traced_allocator = if (constants.tracer_backend == .tracy)
-            tracer.TracyAllocator("tracy").init(arena.allocator())
+            tracer.TracyAllocator("tracy").init(counting_allocator.allocator())
         else
-            arena;
+            &counting_allocator;
```
matklad, line 171: I think the types are different now? One is concreet allocator, the other is allocator vtable

matklad: I mean, due to comptime it doesn't matter that the types are different, but this should break downstream code? Although tests pass, so I must be wrong. Could you double check that it still compiles with tracy?

matklad: ah, no, just can't read as usual!

matklad (approved):

PR #1886, opened 2024-04-29
lsm: fix false positive fuzz failure

### LINKED ISSUE #1876: LSM segmented array fails the fuzz
```
$ ./zig/zig build -Drelease fuzz -- --seed=13934190539457735373 lsm_segmented_array

info(fuzz): Fuzz seed = 13934190539457735373
expected u32@7f99ec14e0bc, found null
error: TestExpectedEqual
```

Haven't looking into this yet

matklad:
For this seed, the array ends up containing maxInt purely by accident.

Seed: ./zig/zig build -Drelease fuzz -- --seed=13934190539457735373 lsm_segmented_array

Closes #1876

### DIFF
--- a/src/lsm/segmented_array.zig
+++ b/src/lsm/segmented_array.zig
@@ -1255,6 +1255,21 @@ fn FuzzContextType(
                 );
                 try testing.expectEqual(iterator_end.next(), null);
             }
+
+            {
+                // 0 is not symmetric with maxInt, because `array.search` doesn't take direction
+                // into account.
+                var iterator_start = context.array.iterator_from_cursor(
+                    context.array.search(0),
+                    .descending,
+                );
+                if (context.reference.items.len == 0) {
+                    try testing.expectEqual(iterator_start.next(), null);
+                } else {
+                    try testing.expect(iterator_start.next() != null);
+                    try testing.expectEqual(iterator_start.next(), null);
+                }
+            }
         }
 
         fn reference_index(context: *const Self, key: Key) u32 {
--- a/src/lsm/manifest_level.zig
+++ b/src/lsm/manifest_level.zig
@@ -464,8 +464,8 @@ pub fn ManifestLevelType(
 
             if (level.keys.len() == 0) return null;
 
-            // Ascending:  Find the first table where table.key_max ≤ iterator.key_min.
-            // Descending: Find the first table where table.key_max ≤ iterator.key_max.
+            // Ascending:  Find the first table where table.key_max ≥ iterator.key_min.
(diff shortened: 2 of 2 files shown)


src/lsm/segmented_array.zig
```diff
@@ -1253,7 +1253,11 @@ fn FuzzContextType(
                     context.array.search(math.maxInt(Key)),
                     .ascending,
                 );
-                try testing.expectEqual(iterator_end.next(), null);
+                try testing.expectEqual(
```
matklad: I wanted to add symmetric case for 0, but it looks like that actually reveals a bug in our code :fearful: 

https://github.com/tigerbeetle/tigerbeetle/pull/1887

matklad: Ah, no, this is surprising, but I think correct. The second commit adds this to a fuzzer in the right way.

src/lsm/segmented_array.zig
```diff
@@ -1255,6 +1255,21 @@ fn FuzzContextType(
                 );
                 try testing.expectEqual(iterator_end.next(), null);
             }
+
+            {
+                // 0 is not symmetric with maxInt, because `array.search` doesn't take direction
+                // into account.
```
sentientwaffle, line 1261: :exploding_head: Surprising!

I double-checked the call sites to see if they were buggy:

`ForestTableIterator.next()` is fine, it only ever iterates ascending.

`ManifestLevel.iterator_start()` is more confusing. I _think_ the code is correct thanks to `iterator_start_boundary()`, but this might be accidental.

It has a comment which notes this behavior:

```
            // Ascending:  Find the first table where table.key_max ≤ iterator.key_min.
            // Descending: Find the first table where table.key_max ≤ iterator.key_max.
            const target = level.keys.search(switch (direction) {
                .ascending => key_min,
                .descending => key_max,
            });
```

...And I think in the comment, the inequalities are backwards?

matklad: My reading that it is correct rather because of this logic

https://github.com/tigerbeetle/tigerbeetle/pull/1886/commits/6555946ad2431385145ae9a858447c145994f929

But year, the inequalities are indeed backwards, will fix

matklad: Yes, I think `iterator_start_boundary` deals with equal keys, but that case actually works in a normal way here. what needsspecial handling is when the smallest key is greater than what we have

sentientwaffle (dismissed):
Aside from the (preexisting) backwards comment in `ManifestLevel` this LGTM.

sentientwaffle (approved):

PR #1887, opened 2024-04-29
lsm: add failing fuzz test

matklad:


matklad:
no, this is working as intended actually!

PR #1888, opened 2024-04-29
style: prefer explicitly-sized integers

matklad:
Note that this might be more practically relevant than it seems, as compiling TigerBeetle for 32-bit WASM is not unreasonable.


src/vsr/grid_scrubber.zig
```diff
@@ -117,8 +117,8 @@ pub fn GridScrubberType(comptime Forest: type) type {
             /// The manifest log tour iterates manifest blocks in reverse order.
             /// (To ensure that manifest compaction doesn't lead to missed blocks.)
             manifest_log: struct { iterator: ManifestBlockIterator = .init },
-            free_set: struct { index: usize = 0 },
-            client_sessions: struct { index: usize = 0 },
+            free_set: struct { index: u32 = 0 },
+            client_sessions: struct { index: u32 = 0 },
```
matklad, line 121: CheckpointTrailer.block_count is u32

src/vsr/grid_scrubber.zig
```diff
@@ -481,10 +481,10 @@ fn ManifestBlockIteratorType(comptime ManifestLog: type) type {
             manifest_log: *const ManifestLog,
         ) ?vsr.BlockReference {
             // Don't scrub the trailing `blocks_closed`; they are not yet flushed to disk.
-            const log_block_count =
-                manifest_log.log_block_addresses.count - manifest_log.blocks_closed;
+            const log_block_count: u32 =
+                @intCast(manifest_log.log_block_addresses.count - manifest_log.blocks_closed);
```
matklad, line 485: This cast is justified because `ManifestLog.checkpoint_references` returns `u32`.

Introducing an `fn ManifestLog.block_count() u32` would have been a more _clean_ solution, but it is perhaps not worth it to add this API, given that the field is going to be publicly accessible anyway.

src/vsr/grid_scrubber.zig
matklad: I originaly hesitated whether i should touch these, but then I noticed that `data_block_index` is u32, so this also turns out a conssitency win if nothing else!

sentientwaffle (approved):

Issue #1839, opened 2024-04-10
Panicked on Windows


azam:
Trying to run the prebuilt binaries panicked with the following message.

```
PS D:\data\tigerbeetle> .\tigerbeetle.exe format --cluster=0 --replica=0 --replica-count=1 0_0.tigerbeetle
Illegal instruction at address 0x7ff76276a6c5
Unable to dump stack trace: debug info stripped
```

Tried building per instruction on `README.md` but same result (instruction address is different).

```
PS D:\data\tigerbeetle> .\tigerbeetle.exe format --cluster=0 --replica=0 --replica-count=1 0_0.tigerbeetle
Illegal instruction at address 0x7ff6b982b695
Unable to dump stack trace: debug info stripped
PS D:\data\tigerbeetle>
```

Here is my environment:

OS: Windows 10 Pro
Build: 22H2 (19045)
CPU: Intel(R) Core(TM) i7-2600K CPU @ 3.40GHz

cb22:
Hey @azam! You're likely running in to https://github.com/tigerbeetle/tigerbeetle/issues/1592#issuecomment-1964865411.

Could you let us know if the workaround there (patching the build.zig file) works for you?

azam:
@cb22 
Thank you, the patch works!
I hope there will be a mention on compiling for old CPU targets somewhere in the docs.

Issue #1592, opened 2024-02-25
Illegal instruction (core dumped)


RealZimboGuy:
attempting to run 

     ./tigerbeetle format --cluster=0 --replica=0 --replica-count=1 0_0.tigerbeetle

dont see any other info, maybe if you have a docker image it could be a nice easy way to spin up an instance ?

 cat /etc/os-release

        NAME="Linux Mint"
        VERSION="20.3 (Una)"
        ID=linuxmint
        ID_LIKE=ubuntu
        PRETTY_NAME="Linux Mint 20.3"
        VERSION_ID="20.3"
        HOME_URL="https://www.linuxmint.com/"
        SUPPORT_URL="https://forums.linuxmint.com/"
        BUG_REPORT_URL="http://linuxmint-troubleshooting-guide.readthedocs.io/en/latest/"
        PRIVACY_POLICY_URL="https://www.linuxmint.com/"
        VERSION_CODENAME=una
        UBUNTU_CODENAME=focal

sentientwaffle:
Hi @RealZimboGuy,

We do have a docker image; see: https://docs.tigerbeetle.com/quick-start/with-docker for details.

Can you paste the output of:

- `./tigerbeetle version --verbose`
- `sha256sum ./tigerbeetle`
- `cat /proc/cpuinfo | head -n28`

Thanks!

RealZimboGuy:
./tigerbeetle version --verbose
        Illegal instruction (core dumped)
 
        sha256sum ./tigerbeetle
        f692ca17fb97f235d6db61b20ce7a0b6ae3aaee9c7d6f341eabfcdb2d0fcb0b0  ./tigerbeetle

        cat /proc/cpuinfo | head -n28
        processor	: 0
        vendor_id	: GenuineIntel
        cpu family	: 6
        model		: 42
        model name	: Intel(R) Core(TM) i7-2670QM CPU @ 2.20GHz
        stepping	: 7
        microcode	: 0x2f
        cpu MHz		: 900.000
        cache size	: 6144 KB
        physical id	: 0
        siblings	: 8
        core id		: 0
        cpu cores	: 4
        apicid		: 0
        initial apicid	: 0
        fpu		: yes
        fpu_exception	: yes
        cpuid level	: 13
        wp		: yes
[...]

PR #1890, opened 2024-04-29
Add SIGILL handler, to give context to the crashes on older CPUs

cb22:
A [common](https://github.com/tigerbeetle/tigerbeetle/issues/1592) [problem](https://github.com/tigerbeetle/tigerbeetle/issues/1839) is the cryptic "Illegal Instruction" message on older CPUs, caused by targeting modern CPU instruction sets and additions like AES-NI and AVX2.

This PR adds a signal handler (and Windows equivalent) to print out a friendly message. I went for the signal handler route as it's the most robust; trying to check features at runtime can invoke code that uses autovectorization and thus panics itself.

Modifying `build.zig` to use unsupported features for my CPU (`x86_64_v3+aes+avx512cd`):

```
zig build install && ./tigerbeetle start --addresses="127.0.0.1:3001" /tmp/0_0.tigerbeetle
error:
error: TigerBeetle's binary releases are compiled targeting modern CPU
error: instructions such as NEON / AES on ARM and x86_64_v3 / AES-NI on
error: x86-64.
error:
error: These instructions can be unsupported on older processors, leading to
error: "illegal instruction" panics.
error:
error: If you'd like to try TigerBeetle on an older processor, you can
error: compile from source with the changes detailed at
error: https://github.com/tigerbeetle/tigerbeetle/issues/1592
error:
Illegal instruction at address 0x465840
fish: Job 1, './tigerbeetle start --addresses…' terminated by signal SIGILL (Illegal instruction)
```

Tested on Windows using Wine, but untested on macOS currently.

### DIFF
--- a/src/tigerbeetle/main.zig
+++ b/src/tigerbeetle/main.zig
@@ -37,6 +37,8 @@ pub const std_options = struct {
 };
 
 pub fn main() !void {
+    try SigIllHandler.register();
+
     // TODO(zig): Zig defaults to 16MB stack size on Linux, but not yet on mac as of 0.11.
     // Override it here, so it can have the same stack size. Trying to set `tigerbeetle.stack_size`
     // in build.zig doesn't work.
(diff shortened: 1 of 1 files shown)


matklad:
:thinking: I am wondering if we should try to explicitly cpuid for these features at startup?

Benefits:

- overall simpler than signal handler
- we have an explicit list of what we actually depend on
- guaranteed to fail fast (and not when we actually reach the code triggering SIGILL)

Drawbacks: 

- might be hard to make sure that the set of checks is precise? Is  there some Zig function perhaps that just looks at `builtin.Target` and compares that with what's actually running?

cb22:
> 🤔 I am wondering if we should try to explicitly cpuid for these features at startup?
> 
> Benefits:
> 
>     * overall simpler than signal handler
> 
>     * we have an explicit list of what we actually depend on
> 
>     * guaranteed to fail fast (and not when we actually reach the code triggering SIGILL)
> 
> 
> Drawbacks:
> 
>     * might be hard to make sure that the set of checks is precise? Is  there some Zig function perhaps that just looks at `builtin.Target` and compares that with what's actually running?

I went down this route initially, and it was a biiig rabbit hole. My processor is an Alder Lake core, but Zig 0.11 doesn't yet have Alder Lake, so its builtin CPU detection picks it up as a generic `x86_64` processor. (It's 'fixed' in 0.12, but it would have the same issue going forward)

Since some of its feature flags are based on the CPU model and not detection (eg `fast_variable_crosslane_shuffle`) - checking your built features are a subset of your detected features doesn't work in this case, unless you also figure out what's significant in the set (which is then brittle).

I also ran into `std.zig.system.NativeTargetInfo.detect(.{});` itself causing an illegal instruction due to autovectorization that I mentioned above - so we'd have to do some reimplementation of the feature detection for both ARM and x86, then we could test explicitly for things like AES, AVX2 / NEON.

matklad (approved):
I must confess I am still uneasy about signal handlers, but this is definitely an improvement over the status quo!

PR #1891, opened 2024-04-29
docs: use illustration for system architecture diagram

emschwartz:


emschwartz:
Actually we're going to use this one 
data_plane

matklad (approved):
At some point, we should move this image to our website

matklad:
One thing I've missed is taht as is the image weighs 2 megs. This feels a bit too heavy for mostly while image with few colors! Would be good to optimize it down at some point.

PR #1892, opened 2024-04-29
ci: unspecial-case minor benchmark

matklad:
We have three "minor", "unit" benchmarks for specific data structures. These benchmarks are not being run continuously, and are an odd addition to our build infrastructure.

I don't think they would provide a lot of value even if we track them --- the specific code is changed rarely, and if it does change, we'll hopefully catch any big regressions in our macro benchmarks.

These benchmarks could still be valueable for someone explicitly hacking on the data structures though.

So, in this PR then benchmark are instead recast as smoke tests. This way:

- they are run by our CI infrastructure and are protected from bitrot
- they don't need any special handling in our build and CI infrastructure
- they still can be used for ad-hoc benchmarking (by manually adjusting the corresponding limits)

### DIFF
--- a/src/lsm/binary_search_benchmark.zig
+++ b/src/lsm/binary_search_benchmark.zig
@@ -26,9 +26,9 @@ const body_fmt = "K={:_>2}B V={:_>3}B N={:_>4} {s}{s}: WT={:_>6}ns UT={:_>6}ns" 
     " CY={:_>6} IN={:_>6} CR={:_>5} CM={:_>5} BM={}";
 
 test "benchmark: binary search" {
-    if (builtin.os.tag != .windows) {
+    if (builtin.os.tag != .linux) {
         // Benchmark uses perf_event_open and only runs on Linux.
-        // return;
+        return;
     }
 
     log.info("Samples: {}", .{searches});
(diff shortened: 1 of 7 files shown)


cb22:
@matklad seems like these benchmarks aren't happy on macOS.

matklad:
Not surprising given that I typoed `linux` as `windows`

matklad:
With a heavy heart ended up removing perf support. I _really_ love the interface here, but I just don't think its the right time to maintain it. We want to minizie deps on the OS layer, and perf is a heavy inteface dep!

src/lsm/binary_search_benchmark.zig
```diff
@@ -16,18 +22,17 @@ const kv_types = .{
...
+
+test "benchmark: binary search" {
+    log.info("Samples: {}", .{searches});
+    log.info("WT: Wall time/search", .{});
+    log.info("UT: utime time/search", .{});
+    log.info("CY: CPU cycles/search", .{});
+    log.info("IN: instructions/search", .{});
+    log.info("CR: cache references/search", .{});
+    log.info("CM: cache misses/search", .{});
+    log.info("BM: branch misses/search", .{});
```
cb22: Perhaps we should remove these logs too?

cb22 (approved):

PR #1895, opened 2024-04-29
MessagePool: Compute `messages_max` at runtime

sentientwaffle:
#### Lift `messages_max_{replica,client}` calculation to runtime.

To pave the way for upcoming memory optimizations.

#### Lift `connections_max` to runtime

`connections_max` accounts for all possible replicas (6) and standbys (6).
But at runtime, we know how many replicas and standbys exist (potentially <6).

This means that we can reduce the number of messages allocated based on how many replicas/standbys there actually are.

In the case of a single-replica cluster, this save 60 messages (60MiB):

- `5` from `Replica.do_view_change_from_all_replicas`
- `11` from `Connection.recv_message`
- `11 * 4` from `Connect.send_queue`

Weirdly, this causes the virtual memory to _increase_ (presumably due to the changed access pattern to the arena allocator).

### DIFF
--- a/src/message_bus.zig
+++ b/src/message_bus.zig
@@ -106,12 +106,14 @@ fn MessageBusType(comptime process_type: vsr.ProcessType) type {
             on_message_callback: *const fn (message_bus: *Self, message: *Message) void,
             options: Options,
         ) !Self {
-            // There must be enough connections for all replicas and at least one client.
-            // -1 since we don't need a connection to ourself.
-            assert(constants.connections_max > options.configuration.len - 1);
             assert(@as(vsr.ProcessType, process_id) == process_type);
 
-            const connections = try allocator.alloc(Connection, constants.connections_max);
+            // The maximum number of connections that can be held open by the server at any time.
+            // -1 since we don't need a connection to ourself.
+            const connections_max: u32 =
+                @intCast(options.configuration.len - 1 + constants.clients_max);
+
+            const connections = try allocator.alloc(Connection, connections_max);
             errdefer allocator.free(connections);
             @memset(connections, .{});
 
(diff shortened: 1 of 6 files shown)


src/message_bus.zig
```diff
@@ -106,12 +106,16 @@ fn MessageBusType(comptime process_type: vsr.ProcessType) type {
...
-            assert(constants.connections_max > options.configuration.len - 1);
             assert(@as(vsr.ProcessType, process_id) == process_type);
 
-            const connections = try allocator.alloc(Connection, constants.connections_max);
+            const connections_max: u32 = switch (process_type) {
+                // The maximum number of connections that can be held open by the server at any
+                // time. -1 since we don't need a connection to ourself.
+                .replica => @intCast(options.configuration.len - 1 + constants.clients_max),
+                .client => @intCast(options.configuration.len),
+            };
```
sentientwaffle, line 116: Clients only connect to replicas, not to other clients.

src/message_pool.zig
```diff
@@ -14,55 +14,77 @@ comptime {
...
+                // yet, so we may guess high. (We can't differentiate between replicas and
+                // standbys.)
+                sum += @min(replica.members_count, constants.replicas_max);
+                sum += connections_max; // Connection.recv_message
+                // Connection.send_queue:
+                sum += connections_max * constants.connection_send_queue_max_replica;
+                sum += 1; // Handle bursts (e.g. Connection.parse_message)
+                // Handle Replica.commit_op's reply:
+                // (This is separate from the burst +1 because they may occur concurrently).
+                sum += 1;
```
sentientwaffle, line 78: For clarity, see https://github.com/tigerbeetle/tigerbeetle/pull/1895/commits/2ba19dadde6197239f6685281caa07c7ee51f418 for the parts of this that weren't simply moved.

matklad (approved):
Excellent!

PR #1897, opened 2024-04-30
CLI: Small caches for `tigerbeetle start --development`

sentientwaffle:
When `--development` is passed to `tigerbeetle start`, use minimal cache sizes.

This reduces the RSS of a single-replica with `--development` from ~3.30GiB to ~2.41GiB.

I left `limit_storage` and `memory_lsm_manifest` untouched since reducing them could prevent you from opening a data file that works without `--development`.

### DIFF
--- a/src/tigerbeetle/cli.zig
+++ b/src/tigerbeetle/cli.zig
@@ -160,7 +160,7 @@ const CliArgs = union(enum) {
         \\  --development
         \\        Allow the replica to format/start even when Direct IO is unavailable.
         \\        For safety, production replicas should always enforce Direct IO -- this flag should only be
-        \\        used for testing and development.
+        \\        used for testing and development. It should not be used for production or benchmarks.
         \\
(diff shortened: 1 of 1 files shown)


matklad:
A wild idea, but what if, instead of making caches smaller, we remove MMAP_POPULATE in development? So that the memory is paged-in lazily?

matklad:
To clarify, I see two benefits here:

* presumably, this should make our memory usage lower, without changing API surface
* it'll also make things faster in development

sentientwaffle:
:thinking: The RSS savings would only be temporarily though -- if they use it for a while it would go up as the cache is filled. I think staying predictable here is better. And the memory usage is going to come down quite a bit for `--development`.

matklad:
Ah, right, even in dev you'll probably actually use the entire cache eventually.

matklad (approved):

PR #1898, opened 2024-04-30
VSR: Add `pipeline_limit`/`--limit-pipeline`

sentientwaffle:
Add a runtime `pipeline_limit`. Context: Our default pipeline size is `clients_max`, which is 64 for the prod/dev config. This is much higher than even production clusters are expected to need (as a margin of safety). We can trim the value down at runtime to save some memory.

Constraints:

1. `pipeline_limit ≥ constants.pipeline_prepare_queue_max`: We don't guarantee that all replicas in a cluster run with the same `pipeline_limit`.
2. `pipeline_limit ≤ constants.pipeline_prepare_queue_max + constants.pipeline_request_queue_max = clients_max`: There is no reason to have a pipeline larger than the `clients_max`.

Set `pipeline_limit` via `--limit-pipeline`, a CLI flag to `tigerbeetle start`. (`--limit-pipeline` is deliberately omitted from `tigerbeetle --help`.)

- When the `--development` flag is set, use the minimum `pipeline_limit` by default.
- When the `--development` flag is not set, use the maximum `pipeline_limit` by default.

For `--development` replicas, this reduces RSS by 336MiB as fewer messages are allocated.

> One man's constant is another man's variable.
> – Alan Perlis

### DIFF
--- a/src/tigerbeetle/cli.zig
+++ b/src/tigerbeetle/cli.zig
@@ -206,7 +206,7 @@ const StartDefaults = struct {
 };
 
 const start_defaults_production = StartDefaults{
-    .limit_pipeline = constants.pipeline_prepare_queue_max + constants.pipeline_request_queue_max,
+    .limit_pipeline = constants.pipeline_request_queue_max,
     .cache_accounts = .{ .value = constants.cache_accounts_size_default },
     .cache_transfers = .{ .value = constants.cache_transfers_size_default },
     .cache_transfers_pending = .{ .value = constants.cache_transfers_pending_size_default },
@@ -215,7 +215,7 @@ const start_defaults_production = StartDefaults{
 };
 
 const start_defaults_development = StartDefaults{
-    .limit_pipeline = constants.pipeline_prepare_queue_max,
+    .limit_pipeline = 0,
     .cache_accounts = .{ .value = 0 },
     .cache_transfers = .{ .value = 0 },
     .cache_transfers_pending = .{ .value = 0 },
@@ -243,7 +243,7 @@ pub const Command = union(enum) {
         cache_transfers_pending: u32,
         cache_account_balances: u32,
         storage_size_limit: u64,
-        pipeline_limit: u32,
+        pipeline_requests_limit: u32,
         cache_grid_blocks: u32,
         lsm_forest_node_count: u32,
         development: bool,
@@ -411,7 +411,7 @@ pub fn parse_args(allocator: std.mem.Allocator, args_iterator: *std.process.ArgI
             }
 
             const pipeline_limit = start.limit_pipeline orelse defaults.limit_pipeline;
-            const pipeline_limit_min = constants.pipeline_prepare_queue_max;
+            const pipeline_limit_min = 0;
             const pipeline_limit_max =
                 constants.pipeline_prepare_queue_max + constants.pipeline_request_queue_max;
             if (pipeline_limit > pipeline_limit_max) {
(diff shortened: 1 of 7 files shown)


src/vsr/replica.zig
```diff
@@ -179,6 +179,16 @@ pub fn ReplicaType(
         /// Invariant: replica < node_count
         replica: u8,
 
+        /// Runtime upper-bound number of prepares/requests in the pipeline.
+        /// Does not change after initialization.
+        /// Invariants:
+        /// - pipeline_limit ≥ pipeline_prepare_queue_max
+        /// - pipeline_limit ≤ pipeline_prepare_queue_max + pipeline_request_queue_max
```
matklad: :thinking: could we perhaps make only `pipeline_request_queue_max` part configurable?

I think our pipeline is actually two different numbers:

* how many prepares to pipeline (this is a protocol-defining parameter, as replica routinely make judgement that for op=N op=N-pipeline *must* be comitted)
* how many requests to pipeline on top of that

and only the second one can very

In the implementation, the two are mixed together, because we subdivide the same memory between:

* prepare cache
* prepare pipeline
* request pipeline

but I don't think it makes sense to leak this all the way to the CLI?


In the alterantive I am imagining, the client specifies the length of requests pipeline, and then we add that to `pipeline_prepare_queue_max` to get the actual size?

Do we even need `pipeline_request_queue_max` as a compile-time const? Could we use runtime-allocated slice for pipeline queue/cache instaed of an array?

sentientwaffle: > In the alterantive I am imagining, the client specifies the length of requests pipeline, and then we add that to pipeline_prepare_queue_max to get the actual size?

:+1: Yeah, that works! Correspondingly, I renamed the flag to `--limit-pipeline-requests`.

> Do we even need pipeline_request_queue_max as a compile-time const?

I think it is still a convenient alias for `constants.clients_max - constants.pipeline_prepare_queue_max` (i.e. the upper-bound request queue size).

> Could we use runtime-allocated slice for pipeline queue/cache instaed of an array?

We _could_, but it would be an extensive enough refactor that it isn't worth it imo. The trouble is that we swap `Replica.pipeline` back-and-forth between two different representations:

```
        /// The pipeline is a queue for a replica which is the primary and in status=normal.
        /// At all other times the pipeline is a cache.
        pipeline: union(enum) {
            /// The primary's pipeline of inflight prepares waiting to commit in FIFO order,
            /// with a tail of pending requests which have not begun to prepare.
            /// This allows us to pipeline without the complexity of out-of-order commits.
            queue: PipelineQueue,
            /// Prepares in the cache may be committed or uncommitted, and may not belong to the
            /// current view.
            cache: PipelineCache,
        },
```

Internally, the `cache` uses a single array, while `queue` splits into two different arrays.

src/tigerbeetle/cli.zig
```diff
@@ -404,6 +410,24 @@ pub fn parse_args(allocator: std.mem.Allocator, args_iterator: *std.process.ArgI
                 );
             }
 
+            const pipeline_limit =
+                start.limit_pipeline_requests orelse defaults.limit_pipeline_requests;
+            const pipeline_limit_min = 0;
+            const pipeline_limit_max =
+                constants.pipeline_prepare_queue_max + constants.pipeline_request_queue_max;
```
matklad:
```suggestion
                constants.pipeline_request_queue_max;
```

src/vsr/replica.zig
```diff
@@ -179,6 +179,16 @@ pub fn ReplicaType(
...
         replica: u8,
 
+        /// Runtime upper-bound number of prepares/requests in the pipeline.
+        /// Does not change after initialization.
+        /// Invariants:
+        /// - pipeline_request_queue_limit ≥ 0
+        /// - pipeline_request_queue_limit ≤ pipeline_request_queue_max
+        ///
+        /// The former invariant is critical since we don't guarantee that all replicas in a cluster
+        /// are started with the same `pipeline_request_queue_limit`.
```
matklad: Might adjust comment, as the first invariant is now just a type

matklad (commented):
LGTM except for the check in the CLI which I thing is wrong under the new semantics?

matklad (approved):

PR #1899, opened 2024-04-30
lsm_fuzz: speed it up

matklad:
The core reason why it is slow is that it is using a HashMap as a model,
but, since then introduction of scans, we really need a data structure
supporting range queries.

While Zig's stdlib doesn't have such a datastructure, we do! We can use
SortedSegmentedArray.

The following seed now finishes in under 5 minutes on my machine:

Seed: ./zig/zig build -Drelease fuzz -- --seed=17470417632183688204 lsm_tree

I have a suspicion that this change might still not be enough, and that
we should change our event generation to just make scans more _costly_
(one scan is equivalent to scan_result_max puts), but lets postpone that
until CFO finds a failing seed.

### DIFF
--- a/src/lsm/tree_fuzz.zig
+++ b/src/lsm/tree_fuzz.zig
@@ -34,6 +34,7 @@ const SuperBlock = vsr.SuperBlockType(Storage);
 const ScanBuffer = @import("scan_buffer.zig").ScanBuffer;
 const ScanTreeType = @import("scan_tree.zig").ScanTreeType;
 const FreeSetEncoded = vsr.FreeSetEncodedType(Storage);
+const SortedSegmentedArray = @import("./segmented_array.zig").SortedSegmentedArray;
 
 const CompactionHelperType = @import("compaction.zig").CompactionHelperType;
 const CompactionHelper = CompactionHelperType(Grid);
@@ -95,6 +96,7 @@ const replica = 4;
 const replica_count = 6;
 const node_count = 1024;
 const scan_results_max = 4096;
+const events_max = 10_000_000;
 const tree_options = .{
     // This is the smallest size that set_associative_cache will allow us.
     .cache_entries_max = 2048,
@@ -159,6 +161,7 @@ fn EnvironmentType(comptime table_usage: TableUsage) type {
         scan_buffer: ScanBuffer,
         scan_results: []Value,
         scan_results_count: u32,
+        scan_results_model: []Value,
         compaction_exhausted: bool = false,
 
         block_pool: CompactionHelper.CompactionBlockFIFO,
@@ -201,6 +204,9 @@ fn EnvironmentType(comptime table_usage: TableUsage) type {
             env.scan_results = try allocator.alloc(Value, scan_results_max);
             env.scan_results_count = 0;
             defer allocator.free(env.scan_results);
+
+            env.scan_results_model = try allocator.alloc(Value, scan_results_max);
+            defer allocator.free(env.scan_results_model);
 
             // TODO: Pull out these constants. 3 is block_count_bar_single, 8 is
             // minimum_block_count_beat.
@@ -514,7 +520,7 @@ fn EnvironmentType(comptime table_usage: TableUsage) type {
         }
 
         pub fn apply(env: *Environment, fuzz_ops: []const FuzzOp) !void {
-            var model = Model.init(table_usage);
+            var model = try Model.init(table_usage);
             defer model.deinit();
 
             for (fuzz_ops, 0..) |fuzz_op, fuzz_op_index| {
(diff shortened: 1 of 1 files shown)


src/lsm/tree_fuzz.zig
```diff
@@ -599,96 +601,152 @@ fn EnvironmentType(comptime table_usage: TableUsage) type {
...
-            // Asserting the positive space:
-            // all keys found by the scan must exist in our model.
-            var tree_value_last: ?Value = null;
-            for (tree_values) |tree_value| {
+            const model_values = model.scan(
+                scan_range.min,
+                scan_range.max,
+                scan_range.direction,
+                env.scan_results_model,
+            );
```
sentientwaffle, line 602: Great symmetry with `tree_values`!

src/lsm/tree_fuzz.zig
```diff
@@ -599,96 +601,152 @@ fn EnvironmentType(comptime table_usage: TableUsage) type {
...
             }
 
-            // Asserting the negative space:
-            // All keys existing in our model must be checked against the scan range.
-            if (scan_range.direction == .descending) std.mem.reverse(
+            try std.testing.expectEqualSlices(
                 Value,
-                tree_values,
+                tree_values[0 .. tree_values.len - tombstone_count],
+                model_values[0 .. tree_values.len - tombstone_count],
```
sentientwaffle: Checking my understanding: Are we are deliberately ignoring any extra values from `model_values`, because they didn't fit in the tree's scan due to tombstones?

src/lsm/tree_fuzz.zig
```diff
@@ -599,96 +601,152 @@ fn EnvironmentType(comptime table_usage: TableUsage) type {
...
             );
+        }
+    };
+}
+
+// A tree is a sorted set. The ideal model would have been an in-memory B-tree, but there isn't
+// one in Zig's standard library. Use a SortedSegmentedArray instead which is essentially a stunted
+// B-tree on-level deep.
```
sentientwaffle:
```suggestion
// B-tree one-level deep.
```

src/lsm/tree_fuzz.zig
```diff
@@ -815,7 +873,7 @@ pub fn main(fuzz_args: fuzz.FuzzArgs) !void {
     };
 
     const fuzz_op_count = @min(
-        fuzz_args.events_max orelse @as(usize, 1E7),
+        fuzz_args.events_max orelse events_max,
```
sentientwaffle, line 871: Maybe `events_max_default`?

src/lsm/tree_fuzz.zig
```diff
@@ -599,96 +601,152 @@ fn EnvironmentType(comptime table_usage: TableUsage) type {
...
-            }
+            },
         }
-    };
-}
+        return result[0..result_count];
+    }
+
+    fn put(model: *Model, value: *const Value) !void {
+        model.remove(value);
```
sentientwaffle, line 720: Ah right, `SegmentedArray` allows duplicates.

matklad:
Ah, interesting, I think as written the check isn’t quiet correct: if the
tree returns an empty array, we’ll be comparing empty arrays, regardless of
what the model returns, I think I need to remove (model_values.len +
tombstone_count) -| scan_results_max values from the model instead!

src/lsm/tree_fuzz.zig
matklad: That was the idea, but the implementation was incorrect (it wrongly trusted the length of tree_values). Fixed the impl and added an assert that would have caught that bug.

src/lsm/tree_fuzz.zig
matklad: went back and forth here, but I think events_max is better.

This is both: 

* a compile-time limit on the number of events (so, `events_max`)
* a default runtime-limit for the actual number of events (so, `events_limit_default`)


comptime semantics is what I am relying on here

src/lsm/tree_fuzz.zig
```diff
@@ -541,21 +543,30 @@ fn EnvironmentType(comptime table_usage: TableUsage) type {
...
-                                env.tree.remove(&old_value);
+                            // Secondary index requires that the key implies the value (typically
+                            // key ≡ value), and that there are no updates.
+                            const canonical_value: Value = .{
+                                .id = value.id,
+                                .value = 0,
+                                .tombstone = value.tombstone,
+                            };
+                            if (model.contains(Value.key_from_value(&canonical_value))) {
+                                env.tree.remove(&canonical_value);
```
matklad, line 554: Missed `// secondary_index only preserves keys - may return old values` subtelty during the initial impl, and decided that it's better to account for then when generating input data.

I am also wondering if we should add asserts to make the old code blow up at runtime?

Here: https://github.com/tigerbeetle/tigerbeetle/blob/fbda8466dcff2c64c7b7ee23d670183d7a9c72e3/src/lsm/compaction.zig#L1581-L1582

we could assert that tombstones are different, but everything _but_ tombstones is the same.

matklad: :thinking: probably isn't worth it: in our real usage, Key=Value, so it's actually impossible to get divergance.

src/lsm/tree_fuzz.zig
matklad: Yeah, that's the whole idea with extracting a dedicated `Model`, to make sure that the interface is mechanically identical.

sentientwaffle (approved):

PR #1904, opened 2024-05-01
cache_map: speed up fuzzing

matklad:
HashMap model doesn't support scopes in an efficient manner. Use HashMap + UndoLog to fix this.

The following seed now finishes in 5 seconds in debug

Seed: ./zig/zig build fuzz --  --seed=5311370176449483187 lsm_cache_map

### DIFF
--- a/src/lsm/cache_map_fuzz.zig
+++ b/src/lsm/cache_map_fuzz.zig
@@ -29,44 +29,24 @@ const FuzzOp = union(enum) {
 
 const Environment = struct {
     cache_map: TestCacheMap,
-    model: std.hash_map.AutoHashMap(Key, OpValue),
-
-    // For Scope support in the model.
-    scope_open: bool = false,
-    scope_model: std.hash_map.AutoHashMap(Key, OpValue),
-
-    // For compact support in the model.
-    compacts: u32 = 0,
+    model: Model,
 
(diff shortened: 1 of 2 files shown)


src/lsm/cache_map_fuzz.zig
```diff
@@ -219,6 +186,100 @@ const Environment = struct {
...
+        }
+    }
+
+    fn copy_hash_map(dst: *Map, src: *Map) !void {
+        dst.clearRetainingCapacity();
+        var it = src.iterator();
+        while (it.next()) |kv| {
+            try dst.putNoClobber(kv.key_ptr.*, kv.value_ptr.*);
+        }
+    }
```
cb22: Unneeded now, no?

cb22:
_Slightly_ better than my copy-the-entire hash map :smile:

src/lsm/cache_map_fuzz.zig
matklad: Indeed, fixed, and also sneaked in some zebra stripes

matklad: Screenshot_20240502_110903

just so the latest commit is clearly visible

cb22 (approved):

PR #1906, opened 2024-05-02
tree_fuzz: fix false positive

matklad:
1_000 was an out of thin air constant I used for testing, lets use the correct value here

Seed: ./zig/zig build -Drelease fuzz -- --seed=8888456137543307989 lsm_tree

### DIFF
--- a/src/lsm/tree_fuzz.zig
+++ b/src/lsm/tree_fuzz.zig
@@ -645,9 +645,14 @@ const Model = struct {
     table_usage: TableUsage,
 
     fn init(table_usage: TableUsage) !Model {
+        const model_node_count = std.math.divCeil(
+            u32,
+            events_max * @sizeOf(Value),
+            NodePool.node_size,
+        ) catch unreachable;
         return .{
             .array = try Array.init(allocator),
-            .node_pool = try NodePool.init(allocator, 1_000),
+            .node_pool = try NodePool.init(allocator, model_node_count),
             .table_usage = table_usage,
         };
     }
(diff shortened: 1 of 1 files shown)


src/lsm/tree_fuzz.zig
```diff
@@ -645,9 +645,14 @@ const Model = struct {
     table_usage: TableUsage,
 
     fn init(table_usage: TableUsage) !Model {
+        const model_node_count = std.math.divCeil(
+            u32,
+            events_max * @sizeOf(Value),
+            NodePool.node_size,
+        ) catch unreachable;
```
sentientwaffle: Fyi we have a `stdx.div_ceil` that doesn't return an error. (It asserts that the denominator is nonzero.)

sentientwaffle: Nodes don't store values, they store `TableInfo`. Should we be dividing by table's value capacity?

matklad: These nodes do store values though? That's a different segmented array

sentientwaffle: :facepalm: Ahh I was still thinking about manifestlevels, which isn't what this is for.

sentientwaffle (dismissed):

src/lsm/tree_fuzz.zig
matklad: Ahhhh, I had a vague recollection of that, but grepped for `divCeil` which didtt' help :rofl:

sentientwaffle (approved):

PR #1909, opened 2024-05-03
Scripts: Use ZIG_EXE environment variable

sentientwaffle:
I use (many) git worktrees, and rather than giving each their own zig binary, I install it once in my PATH, so `./zig/zig` doesn't exist.
`ZIG_EXE` is compatible with both workflows.

(`ZIG_EXE` is set automatically for `zig run ...`  – but not `zig build ...` with Run steps, so it needs to be explicitly set in `build.zig`.)

### DIFF
--- a/src/scripts/cfo.zig
+++ b/src/scripts/cfo.zig
@@ -118,7 +118,6 @@ fn run_fuzzers(
         hang_seconds: u64,
     },
 ) !void {
-    const zig_exe = try shell.zig_exe_alloc(shell.arena.allocator());
     // Fuzz an independent clone of the repository, so that CFO and the fuzzer could be on
     // different branches (to fuzz PRs and releases).
     shell.project_root.deleteTree("working") catch {};
@@ -189,7 +188,7 @@ fn run_fuzzers(
                         .{ .stdin_behavior = .Pipe },
                         "{zig} build -Drelease fuzz -- --seed={seed} {fuzzer}",
                         .{
-                            .zig = zig_exe,
+                            .zig = shell.zig_exe.?,
                             .seed = try shell.print("{d}", .{seed}),
                             .fuzzer = @tagName(fuzzer),
                         },
--- a/src/scripts/cfo.zig
+++ b/src/scripts/cfo.zig
@@ -118,6 +118,7 @@ fn run_fuzzers(
         hang_seconds: u64,
     },
 ) !void {
+    const zig_exe = try shell.zig_exe_alloc(shell.arena.allocator());
     // Fuzz an independent clone of the repository, so that CFO and the fuzzer could be on
     // different branches (to fuzz PRs and releases).
     shell.project_root.deleteTree("working") catch {};
@@ -143,7 +144,7 @@ fn run_fuzzers(
 
     const random = std.crypto.random;
 
-    try shell.exec("../zig/zig build -Drelease build_fuzz", .{});
+    try shell.zig("build -Drelease build_fuzz", .{});
 
     const FuzzerChild = struct {
         child: std.ChildProcess,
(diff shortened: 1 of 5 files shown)


src/shell.zig
```diff
@@ -596,12 +596,33 @@ pub fn spawn_options(
     return child;
 }
 
+pub fn zig_exe_alloc(shell: Shell, allocator: std.mem.Allocator) ![]const u8 {
```
matklad: For simplicity, let's rather do this during `Shell.create` and store the result as a field, like how we do for `project_root`. 

I debated whether we should keep shell.zig as an independent re-usable utility, or whether to allow dependency on TigerBeetle repo specifically, and I figured that a bounded amount of dependency is probably better.k

As a bonus, this'll tell us eagarly if we miss the environmental variable in our build.zig!

src/scripts/cfo.zig
```diff
@@ -143,7 +144,7 @@ fn run_fuzzers(
 
     const random = std.crypto.random;
 
-    try shell.exec("../zig/zig build -Drelease build_fuzz", .{});
+    try shell.zig("build -Drelease build_fuzz", .{});
```
matklad, line 146: :+1: I actually forgot that we have `shell.zig`!

src/scripts/cfo.zig
```diff
@@ -186,8 +187,12 @@ fn run_fuzzers(
                     log.debug("will start '{s}'", .{command});
                     const child = try shell.spawn_options(
```
matklad, line 187: but, sadly, we can't use it here, because we need a child...

It's on my todo list to re-visint shell.zig and try to come up with somewhat more orthogonal set of APIs. I like _using_ shell.zig, but the implementaiton is almost 1000 lines of mostly copy-paste :(

src/scripts/ci.zig
```diff
@@ -97,7 +97,7 @@ fn validate_release(shell: *Shell, gpa: std.mem.Allocator, language_requested: ?
         .{},
     );
     const tag = stdx.cut(release_info, "\t").?.prefix;
-    log.info("validateing release {s}", .{tag});
```
matklad, line 100: :heart:

src/shell.zig
```diff
@@ -596,12 +596,33 @@ pub fn spawn_options(
...
+    _ = shell;
+
+    // ZIG_EXE is set in `build.zig`, so it is available when using the run step for `zig build
+    // scripts`.
+    // ZIG_EXE is already an absolute path, but zig/zig is not.
+    const zig_exe_default = comptime "zig/zig" ++ builtin.target.exeFileExt();
```
matklad: So, let's assume that it's just always set!

matklad (commented):
LGTM!

In general, I am a firm believer that it's better to say "**this** is the single configuration our project uses" rather than to try to support many different ways to do things. (and, for my worktrees, I actuall have a `./zig/zig` symlink in each one of them).

_However_ we don't actually make this more configurable really, we are just extracting a hard-coded constant into a variable, which is an improvement just along the readability axis!

LGTM with `zig_exe_alloc` replaced with a field --- I don't think we should have any "caller manages the memory" functions in shell.zig, it should figure out memory itself

src/shell.zig
sentientwaffle: Ah, that is much better! We even already have the `EnvMap` in `Shell.create`, which simplifies the error handling.

Though, we do miss the environment variable often -- e.g. `build.zig` _itself_ uses `Shell`. It just doesn't end up being needed in those cases.

src/shell.zig
sentientwaffle: As a bonus to this -- `b.zig_exe` (in `build.zig`) is already an absolute path!

matklad (approved):

PR #1912, opened 2024-05-06
docs: move Requests and Sessions into Reference

emschwartz:
- **docs: rename Reference to API Reference**
- **docs: use code formatting in titles**
- **docs: move requests and sessions docs into api reference**
- **docs: rename Operations to Requests and merge in Client Requests page**
- **docs: update sessions page**
- **docs: move guarantees from consistency page to api reference pages**


docs/api-reference/account_balances.md
```diff
@@ -2,7 +2,7 @@
 sidebar_position: 3
 ---
 
-# Account Balances
+# `AccountBalance`s
```
matklad: I'd simplify just to `AccountBalance`. I _think_ `# titles` also generate fragment links in URL, and I just `#accountbalance` works better than `#AccountBalances` (but we need to check what our doc generator actually does

matklad: Yeah, and in the TOC for API reference I'd also expect to see `Account`, as it names a specific "class" rather than accounts in general.

matklad: and I guess as an url, `http://localhost:3000/api-reference/account` reads better.

docs/about/internals/sync.md
```diff
@@ -50,49 +54,62 @@ Checkpoints:
...
+9. Repair [replies](./vsr.md#protocol-sync-client-replies),
+   [free set, client sessions, and manifest blocks](./vsr.md#protocol-repair-grid), and
+   [table blocks](./vsr.md#protocol-sync-forest) that were created within the `sync_op_{min,max}`
+   range.
 10. Update the superblock with:
     - Set `vsr_state.sync_op_min = 0`
     - Set `vsr_state.sync_op_max = 0`
 
-If a newer sync target is discovered during steps *5*-*6* or *9*, go to step *4*.
+If a newer sync target is discovered during steps _5_-_6_ or _9_, go to step _4_.
```
matklad, line 73: BTW, would be great to document the style we use somewhere, just so that we don't flip back and forth between * and _

docs/getting-started/cli-repl.md
```diff
@@ -0,0 +1,87 @@
+---
+sidebar_position: 6
```
matklad: I think this was added by accident? We killed this folder on main, no?

docs/api-reference/account_filter.md
```diff
@@ -2,7 +2,7 @@
 sidebar_position: 4
 ---
 
-# Account Filter
+# `AccountFilter`
```
matklad: The url for this page is 

```
http://localhost:3000/api-reference/account_filter
```

underscores in URLS are a poor taste, because the urls are styled with an underline anyway. 

So, we need to change _something_ here. 

I don't know what naming scheme would be best. 

```
http://localhost:3000/api-reference/account-filter
```

sounds ok to me

docs/api-reference/requests/README.md
```diff
@@ -0,0 +1,112 @@
+# Requests
```
matklad: Here, I think plural works, because there indeed many different kinds of requests. There are subpages to this page

docs/develop/consistency.md
```diff
@@ -1,224 +0,0 @@
...
-conventional database.
-
-Strict consistency guarantees (at the database level) simplify 1) application logic and 2) error
-handling farther up the stack.
-
-## Guarantees
-
-TigerBeetle provides strict serializability
-([serializability + linearizability](http://www.bailis.org/blog/linearizability-versus-serializability/))
-to each [client session](./client-sessions.md).
```
matklad: Would be good to move these two intro paragrpahs to the `Concepts` section, once we have it. It's important to mention "strict serializability" _before_ we get to the reference

docs/api-reference/accounts.md
```diff
@@ -21,6 +21,17 @@ Accounts **cannot be deleted** after creation. This provides a strong guarantee
...
+
+- Accounts are immutable. They are never modified once they are successfully created (excluding
+  balance fields, which are modified by transfers).
+- There is at most one `Account` with a particular [`id`](#id).
+- The sum of all accounts' [`debits_pending`](#debits_pending) equals
+  the sum of all accounts' [`credits_pending`](#credits_pending).
+- The sum of all accounts' [`debits_posted`](#debits_posted) equals the
+  sum of all accounts' [`credits_posted`](#credits_posted).
+
+
```
matklad: These are much more at the right place here, thanks!

docs/api-reference/_category_.json
sentientwaffle, line 1: Since this section is also going to include (e.g.) Limits, Client Sessions, etc -- maybe "reference" is a better name still?

docs/about/internals/sync.md
emschwartz: I'm not sure how folks would feel about introducing a tool that does this but I think it would make sense to use a formatting tool like Prettier, include the config file, and include a step in the build process that checks that it matches the given format

emschwartz:
Sorry, that last commit changed the format on too many things. Going to remove it and do that again

docs/about/internals/sync.md
matklad: This mostly depends on the tool! I think our decision matrix is

```
good tool > no tool > bad tool
```

For prettier in particular, I think it might get disqualified because our long term plan is "npm is banned". We don't want to depend on JS ecosystem as it is tends to be brittle, slow, and churn-prone.

docs/api-reference/_category_.json
emschwartz: 🤷 I don't feel strongly about this...

docs/about/internals/sync.md
emschwartz: That's fair. I looked around and there doesn't seem to be a good option for markdown formatting in a different language that's well supported.

I don't love this option but we could document the style we use + include a prettier config file in the repo, but not enforce its use with a build step...

docs/develop/consistency.md
emschwartz: Agreed!

matklad (dismissed):

docs/api-reference/_category_.json
emschwartz: Joran agreed with just calling it Reference

matklad (dismissed):

matklad (approved):

PR #1913, opened 2024-05-06
cfo: implement PR fuzzing 

matklad:
When a PR is labeled with `fuzz fuzzer_name`, CFO will fuzz that PR
using the specified fuzzers (you can apply multiple labels to a PR).

The time is split 50:50 between all of the PRs and the main branch.

Implementation:

* Use `gh` utility to talk to GitHub about open PRs
* Add `branch` field to SeedRecord, to separate "things that fail on
  main" from "things that fail in the PR"
* On the "front-end", show seeds from all branches (that is, "one
  failing seed per fuzzer _per branch_").

### DIFF
--- a/src/scripts/cfo.zig
+++ b/src/scripts/cfo.zig
@@ -81,20 +81,25 @@ pub fn main(shell: *Shell, gpa: std.mem.Allocator, cli_args: CliArgs) !void {
     }
 
     // Personal Access Token for <https://github.com/tigerbeetle/devhubdb>.
-    const token_option = shell.env_get_option("DEVHUBDB_PAT");
-    if (token_option == null) {
+    const devhub_token_option = shell.env_get_option("DEVHUBDB_PAT");
+    if (devhub_token_option == null) {
         log.err("'DEVHUB_PAT' environmental variable is not set, will not upload results", .{});
+    }
+
+    const gh_token_option = shell.env_get_option("GH_TOKEN");
+    if (gh_token_option == null) {
+        log.err("'GH_TOKEN' environmental variable is not set, will not fetch pull requests", .{});
     }
 
     assert(try shell.exec_status_ok("git --version", .{}));
 
     var seeds = std.ArrayList(SeedRecord).init(shell.arena.allocator());
-    try run_fuzzers(shell, &seeds, .{
+    try run_fuzzers(shell, &seeds, gh_token_option, .{
         .concurrency = cli_args.concurrency orelse try std.Thread.getCpuCount(),
         .budget_seconds = cli_args.budget_minutes * std.time.s_per_min,
         .hang_seconds = cli_args.hang_minutes * std.time.s_per_min,
     });
-    if (token_option) |token| {
+    if (devhub_token_option) |token| {
         try upload_results(shell, gpa, token, seeds.items);
     } else {
         log.info("skipping upload, no token", .{});
@@ -104,12 +109,15 @@ pub fn main(shell: *Shell, gpa: std.mem.Allocator, cli_args: CliArgs) !void {
 fn run_fuzzers(
     shell: *Shell,
     seeds: *std.ArrayList(SeedRecord),
+    gh_token: ?[]const u8,
     options: struct {
         concurrency: usize,
         budget_seconds: u64,
         hang_seconds: u64,
     },
 ) !void {
+    _ = gh_token;
+
     const commit_sha: [40]u8 = commit_sha: {
         const commit_str = try shell.exec_stdout("git rev-parse HEAD", .{});
         assert(commit_str.len == 40);
--- a/src/scripts/cfo.zig
+++ b/src/scripts/cfo.zig
@@ -230,8 +230,6 @@ fn run_fuzzers_prepare_tasks(shell: *Shell, gh_token: ?[]const u8) !struct {
     seed_record: []SeedRecord,
     weight: []u32,
 } {
-    _ = gh_token;
-
     var working_directory = std.ArrayList([]const u8).init(shell.arena.allocator());
     var seed_record = std.ArrayList(SeedRecord).init(shell.arena.allocator());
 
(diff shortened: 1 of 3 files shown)


matklad:
Tentatively assigning for a review as this is _mostly_ ready, though I haven't done a self-review here yet.

matklad:
image

src/scripts/cfo.zig
```diff
@@ -240,6 +225,166 @@ fn run_fuzzers(
...
+    { // Main branch fuzzing.
+        const commit = try run_fuzzers_commit_info(shell);
+        log.info("fuzzing commit={s} timestamp={d}", .{ commit.sha, commit.timestamp });
+
+        try shell.exec("git clone https://github.com/tigerbeetle/tigerbeetle working/main", .{});
+        try shell.pushd("./working/main");
+        defer shell.popd();
+
+        try shell.exec("git switch --detach {commit}", .{ .commit = @as([]const u8, &commit.sha) });
+        shell.exec("../zig/zig build -Drelease build_fuzz", .{}) catch |err| {
```
cb22: `shell.zig()`?

src/scripts/cfo.zig
```diff
@@ -320,15 +465,19 @@ const SeedRecord = struct {
...
-    seed: u64,
-    command: []const u8, // excluded from comparison
+    ok: bool = false,
+    seed_timestamp_start: u64 = 0,
+    seed_timestamp_end: u64 = 0,
+    seed: u64 = 0,
+    // The following fields are excluded from comparison:
+    command: []const u8 = "",
+    // TODO: when all seeds in the db have a branch specified, remove this default.
+    branch: []const u8 = "",
```
cb22, line 475: Is `branch` meant to be the name of the branch (eg, `main`) or a link to it on GitHub (`.branch = "https://github.com/tigerbeetle/tigerbeetle",`)?

src/scripts/cfo.zig
```diff
@@ -240,6 +225,166 @@ fn run_fuzzers(
...
+                color: []const u8,
+            };
+            number: u32,
+            labels: []Label,
+        };
+
+        const pr_list_text = try shell.exec_stdout(
+            "gh pr list --state open --json number,labels",
+            .{},
+        );
```
sentientwaffle, line 281: TIL non-maintainers can't add labels to PRs! Which is important, since we don't want to let anyone run arbitrary code on the fuzzer machines.

src/scripts/cfo.zig
```diff
@@ -240,6 +225,166 @@ fn run_fuzzers(
...
+        weight_main.* = @max(task_pr_count, 1);
+        weight_main_total += weight_main.*;
+    }
+    for (weight[task_main_count..]) |*weight_pr| {
+        weight_pr.* = @max(task_main_count, 1);
+        weight_pr_total += weight_pr.*;
+    }
+    if (weight_main_total > 0 and weight_pr_total > 0) {
+        assert(weight_main_total == weight_pr_total);
+    }
```
sentientwaffle, line 360: Out of scope for this PR, but (within `main`) would it also make sense to weight different fuzzers differently? e.g. `SegmentedArray` is quite static, whereas `Forest` covers a lot of tricky stuff.

src/scripts/cfo.zig
matklad: The URL. It is used for:

* UI-level grouping of failing seeds (so that seeds from main don't "shadow" seeds from PRs when displaying)
* UI links to the code

so, anything that's sufficiently unique and sufficiently clickable works.

src/scripts/cfo.zig
matklad: Perhaps! I actual implemented the weights thing even before thinking about how exactly I am going to use them, as that seems flexible enough. 

But my ground guess is that picking fuzzers uniformly would be good enough, and it has the benefit of being pretty robust.

In general, we severely lack any means of checking whether we are fuzzing efficiently at all, and I'd personally would like to get at least basic coverage going before trying to guess what we should spend more time fuzzing.

src/scripts/cfo.zig
matklad: >since we don't want to let anyone run arbitrary code on the fuzzer machines.

Long term, I'd actually go in the opposite direciton and change the thread model to allow arbitrary code there, and rather tightening the scope of what fuzzing machines can do.

I wouldn't worry about someone stealing our machines to mine bitcoint or sending spam --- we have just a single machine and are not a scalable target.

I do worry about someone deliberately attacking our infra, but fuzzers should not be able to do a lot of damage. Right now, they have write acccess to devhubdb, and that's it (*). So, they could wipe our historical data, and they can XSS devhub page (which is written in a very xss-friendly way), but neither seems like a big problem to me.

(*) the  asterisk here is that the machine is stateful, and who knows which secrets might be in its bash history! If we _do_ say our fuzzers are not trusted, we should make sure they are deployed in a stateless and human-less way.

src/scripts/cfo.zig
```diff
@@ -240,6 +225,166 @@ fn run_fuzzers(
...
+                .{ .pr_number = pr.number },
+            );
+            try shell.exec("git switch --detach FETCH_HEAD", .{});
+
+            const commit = try run_fuzzers_commit_info(shell);
+            log.info("fuzzing commit={s} timestamp={d}", .{ commit.sha, commit.timestamp });
+
+            shell.zig("build -Drelease build_fuzz", .{}) catch |err| {
+                log.warn("failed to build: {}", .{err});
+            };
```
sentientwaffle, line 315: To check my understanding -- is the `zig build build_fuzz` here (and above) to prevent all of the child fuzz processes from trying to compile the same code concurrently?

sentientwaffle (approved):

src/scripts/cfo.zig
matklad: Not exactly: it wouldn’t be wrong to just remove these.

But it is helpful to:

* get more precise times for seed (otherwise, the first seed would seem longer than it is)
* log “this failed to compile” for cfo itself. CFO doesn’t really try to understand *why* things fail, but compilation failures feel worthwhile to log, just in cas something abut out build process itself changes.

matklad:
Merged as is, but added some more explanatory comments and cleanups in https://github.com/tigerbeetle/tigerbeetle/pull/1916

PR #1917, opened 2024-05-07
Multiversion: make release_client_min a parameter, set by release.zig

cb22:
There was a bit of back and forth thinking as to where it should be set. 

Figured release.zig was the most natural place, although it does feel a bit strange to have a hard-coded constant there, we need it _somewhere_ and tying it in to the code directly felt worse off.

### DIFF
--- a/src/config.zig
+++ b/src/config.zig
@@ -19,7 +19,8 @@ const BuildOptions = struct {
     tracer_backend: TracerBackend,
     hash_log_mode: HashLogMode,
     git_commit: ?[40]u8,
-    release: []const u8,
+    release: ?[]const u8,
+    release_client_min: ?[]const u8,
     config_aof_record: bool,
     config_aof_recovery: bool,
 };
@@ -47,6 +48,7 @@ const build_options: BuildOptions = blk: {
 fn launder_type(comptime T: type, comptime value: anytype) T {
     if (T == bool or
         T == []const u8 or
+        T == ?[]const u8 or
         T == ?[40]u8)
     {
         return value;
@@ -83,6 +85,7 @@ const ConfigProcess = struct {
     hash_log_mode: HashLogMode = .none,
     verify: bool,
     release: vsr.Release = vsr.Release.minimum,
+    release_client_min: vsr.Release = vsr.Release.minimum,
     git_commit: ?[40]u8 = null,
     port: u16 = 3001,
     address: []const u8 = "127.0.0.1",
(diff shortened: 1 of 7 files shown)


src/vsr.zig
```diff
@@ -270,6 +270,9 @@ pub const BlockReference = struct {
 
 /// Viewstamped Replication protocol commands:
 pub const Command = enum(u8) {
+    // Looking to make backwards incompatible changes here? Make sure to check release.zig for
```
cb22, line 273: I could make a comptime check that calculates the hash of all the fields and values - so if you update something, you have to update manually. Think it's worth it?

sentientwaffle: I think it's _probably_ not worth it:

- It could provide a false sense of assurance -- not all upgrade-worthy changes would trigger the warning.
- (And also, it isn't necessarily required to upgrade in all such cases. For example, if you were adding a `command`, but it is safe for old versions to ignore the command.)

src/config.zig
```diff
@@ -309,15 +312,26 @@ pub const configs = struct {
             @compileError("invalid release version");
         };
 
+        const release_client_min_raw = if (build_options.release_client_min) |release_client_min|
+            release_client_min
+        else
+            build_options.release;
+        const release_client_min_triple = vsr.ReleaseTriple.parse(release_client_min_raw) catch {
+            @compileError("invalid release_client_min version");
+        };
```
sentientwaffle: Can you additionally assert:

```
assert((release_triple.value == vsr.ReleaseTriple.minimum.value) ==
    (release_client_min_triple.value == vsr.ReleaseTriple.minimum.value));
```

i.e.: If `release_triple` is _not_ the minimum (0.0.1) then `release_client_min_triple` is _also_ not the minimum.

sentientwaffle: And imo we should pass 0.0.1 as the default in `build.zig`, rather than using `build_options.release` as the default. That way we can't accidentally publish 0.15.x with a minimum client version _equal to_ 0.15.x if we miss-set the parameter.

cb22: Hmmm - I think the behavior we want here is for it to be an error to specify one of `release` / `release_client_min` without the other too?

Eg,
`-Dconfig-release=...` -> error
`-Dconfig-release=... -Dconfig-release-client-min=...` -> good
`-Dconfig-release-client-min=...` -> error
` ` -> good (both are `0.0.1`)

sentientwaffle: Yeah, that works too!

sentientwaffle (approved):

PR #1918, opened 2024-05-07
cfo: enable VOPR

matklad:
I believe after this PR we have achieved feature parity with the old VOPR Hub and can retire it. But that's left for a follow up.

### DIFF
--- a/src/scripts/cfo.zig
+++ b/src/scripts/cfo.zig
@@ -73,6 +73,48 @@ const Fuzzer = enum {
     vsr_journal_format,
     vsr_superblock_quorums,
     vsr_superblock,
+    vopr,
+    vopr_testing,
+
+    fn print_command(fuzzer: Fuzzer, shell: *Shell, seed: u64) ![]const u8 {
+        if (fuzzer == .vopr or fuzzer == .vopr_testing) {
+            const state_machine: []const u8 =
+                if (fuzzer == .vopr) "" else " -Dsimulator-state-machine=testing";
+            return try shell.print(
+                "./zig/zig build -Drelease{s} simulator_run -- {d}",
+                .{ state_machine, seed },
+            );
+        }
+        return try shell.print(
+            "./zig/zig build -Drelease fuzz -- --seed={d} {s}",
+            .{ seed, @tagName(fuzzer) },
+        );
+    }
+
+    fn spawn_command(fuzzer: Fuzzer, shell: *Shell, seed: u64) !std.ChildProcess {
+        if (fuzzer == .vopr or fuzzer == .vopr_testing) {
+            const state_machine: []const []const u8 =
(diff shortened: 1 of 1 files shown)


src/scripts/cfo.zig
```diff
@@ -73,6 +73,48 @@ const Fuzzer = enum {
...
+                .{
+                    .zig = shell.zig_exe.?,
+                    .state_machine = state_machine,
+                    .seed = seed,
+                },
+            );
+        }
+        return try shell.spawn_options(
+            .{ .stdin_behavior = .Pipe },
+            "{zig} build -Drelease fuzz -- --seed={seed} {fuzzer}",
```
matklad, line 110: I don't love how I have to repeat essentially the same thing again and again and again and again here. I'll solve this separately, by harmonizing vopr and minor fuzzer's CLI interface (I think I'll take VOPR's approach of passing the seed in as a positonal)

src/scripts/cfo.zig
```diff
@@ -383,12 +417,6 @@ fn run_fuzzers_prepare_repository(shell: *Shell, target: union(enum) {
             break :commit try run_fuzzers_commit_info(shell);
         },
     };
-    log.info("fuzzing commit={s} timestamp={d}", .{ commit.sha, commit.timestamp });
-    // Strictly speaking, we don't need to build the code here, but doing this makes timing
-    // information on seeds more precise.
-    shell.zig("build -Drelease build_fuzz", .{}) catch |err| {
-        log.warn("failed to build: {}", .{err});
-    };
```
matklad, line 391: Well, "lets build beforehand" doesn't really work given that we have _two_ build configs for vopr, which we might or might not need (depending on PR labels). I _could_ solve this by carefully compiling before starting the timer, but I am not sure it is worth it, and I want to tidy CLI interface and build.zig before that, so that I don't need even more duplication here.

src/scripts/cfo.zig
```diff
@@ -73,6 +73,48 @@ const Fuzzer = enum {
...
     vsr_superblock,
+    vopr,
+    vopr_testing,
+
+    fn print_command(fuzzer: Fuzzer, shell: *Shell, seed: u64) ![]const u8 {
+        if (fuzzer == .vopr or fuzzer == .vopr_testing) {
+            const state_machine: []const u8 =
+                if (fuzzer == .vopr) "" else "-Dsimulator-state-machine=testing";
+            return try shell.print(
+                "./zig/zig build -Drelease{s} simulator_run -- {d}",
```
sentientwaffle, line 84:
```suggestion
                "./zig/zig build -Drelease {s} simulator_run -- {d}",
```

matklad: Ah, good catch! This is intentionally without a space, but I am missing one above in `-Dsimulator-state-machine`!

sentientwaffle (approved):
:rocket:

PR #1920, opened 2024-05-07
cfo: build all fuzzers first

matklad:
For some reason, all VOPR runs timed out on CFO:

https://github.com/tigerbeetle/devhubdb/blob/9ee578f1f39530573dafcf5dfa1c856f312ef1bc/fuzzing/data.json#L530-L617

(note `"ok": false` and `"seed_timestamp_end": 1715094977`) for each one of the seeds.

I wasn't able to reproduce this locally or on the CFO machine, but my hypothesis here is that the build system deadlocks with a lot of concurrency.

Restore the behavior of building all code _once_ before fuzzing, even if we don't need some of the fuzzers.

Note that I've unintentionally didn't remove building of minor fuzzers, so those _dont_ suffer concurrent build issues, and perhaps are not timing out for this reason.

### DIFF
--- a/src/scripts/cfo.zig
+++ b/src/scripts/cfo.zig
@@ -173,8 +173,6 @@ fn run_fuzzers(
 
     const random = std.crypto.random;
 
-    try shell.zig("build -Drelease build_fuzz", .{});
-
     const FuzzerChild = struct {
         child: std.ChildProcess,
         seed: SeedRecord,
@@ -242,7 +240,10 @@ fn run_fuzzers(
                 };
 
                 if (fuzzer_done or last_iteration) {
-                    log.debug("will reap '{s}'", .{fuzzer.seed.command});
+                    log.debug(
+                        "will reap '{s}'{s}",
+                        .{ fuzzer.seed.command, if (!fuzzer_done) " (timeout)" else "" },
+                    );
                     const term = try if (fuzzer_done) fuzzer.child.wait() else fuzzer.child.kill();
                     var seed_record = fuzzer.seed;
                     seed_record.ok = std.meta.eql(term, .{ .Exited = 0 });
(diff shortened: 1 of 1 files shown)


matklad:
Sorry for flipping my mind back and forth here @sentientwaffle :rofl:

src/scripts/cfo.zig
```diff
@@ -436,6 +440,18 @@ fn run_fuzzers_commit_info(shell: *Shell) !Commit {
     return .{ .sha = commit_sha, .timestamp = commit_timestamp };
 }
 
+fn run_fuzzers_build_all(shell: *Shell) void {
+    try shell.zig("build -Drelease build_fuzz", .{}) catch |err| {
+        log.err("'build -Drelease build_fuzz failed': {}", .{err});
```
sentientwaffle: s/failed// to be consistent with the other error messages?

sentientwaffle (dismissed):

src/scripts/cfo.zig
```diff
@@ -242,7 +240,11 @@ fn run_fuzzers(
                 };
 
                 if (fuzzer_done or last_iteration) {
-                    log.debug("will reap '{s}'", .{fuzzer.seed.command});
+                    log.debug(
+                        "will reap '{s}'{s}",
+                        .{fuzzer.seed.command},
+                        if (!fuzzer_done) " (timeout)" else "",
```
sentientwaffle: Ah last string, this needs to be in the arguments `.{}` list.

sentientwaffle (approved):

src/scripts/cfo.zig
matklad: one of these days, I'll either hook all "does it compiles/is it formatted?" checks to `./zig/zig build check`, OR I make `./zig/zig build test` to do all that plus finish in seconds!

Right now, I routinely slip many trivial mistakes in PRs simply because its easier to `git push` then to think about which quick command I need to run to check if my code compiles :)

PR #1922, opened 2024-05-08
cfo: vopr no longer hangs

matklad:
Currently vopr hangs in the CFO. This due to the classical bug of systems programming: VOPR logs to the stderr, but no one reads the stderr, so it fills up and then the process deadlocks.

Good news is that we have a fix for these classes of programs: shell.zig intends to wrap raw child process spawning API in a set of well-defined high-level operations, which are protected from these kinds of footguns.

The bad news is that the actual API there is buggy, and defaults to piping by default. Fix this and make each caller that wants to interract with the streams explicitly request that.


matklad:
This is super funny! I literately had two bugs in mind when creating shell.zig:

* failing to check process status after `try wait`
* failing to concurrently drain both stdout and stderr when spawning an interactive program.

One does not simply not step on a rake!

sentientwaffle (approved):

PR #1923, opened 2024-05-08
Build: remove `main_pkg_path = src/` early for 0.12

kprotty:
0.12 is [removing `main_pkg_path`](<https://ziggit.dev/t/suggested-project-layout-for-multiple-entry-point-for-zig-0-12/4219>) in favor of import modules. `vsr_module` already acts as our reference point for `src/` imports, so this extends its usage to everywhere `.main_pkg_path = "src/"` was used previously.

This is a change we can do in 0.11 already, reducing the large review scope of the [0.12 migration](#1878).


build.zig
```diff
@@ -231,9 +231,9 @@ pub fn build(b: *std.Build) !void {
             .name = "tb_client_header",
             .root_source_file = .{ .path = "src/clients/c/tb_client_header.zig" },
             .target = target,
-            .main_pkg_path = .{ .path = "src" },
+            // .main_pkg_path = .{ .path = "src" },
```
matklad: Let's remove this for good rather than leaving commeted-out?

matklad (dismissed):
:thinking: 

I am still thinking what's the best layout for us long term, but this is indeed the minimal change that unblocks the migration!

src/lsm/binary_search_benchmark.zig
```diff
@@ -142,24 +142,49 @@ const BenchmarkResult = struct {
...
-        const rusage = std.os.getrusage(0);
+        const utime_now = utime_nanos();
         return BenchmarkResult{
             .wall_time = self.timer.read() / samples,
-            .utime = (timeval_to_ns(rusage.utime) - timeval_to_ns(self.rusage.utime)) / samples,
+            .utime = @intCast((utime_now - self.utime_ns) / samples),
         };
     }
+
+    fn utime_nanos() u128 {
```
matklad, line 163: When we touch this code next time, I think we could remove this whole rusage tracking thing in favor of timer, following the rationale similar to per removal. 

https://github.com/tigerbeetle/tigerbeetle/commit/9702791c6008f4f69829debf935b6ac3a7cd5422

matklad (approved):

PR #1924, opened 2024-05-08
vorp: add poor man's minimization 

matklad:
By passing `--lite` flag to VOPR, it will only look at correctness
issues in a small cluster.

Best reviewed per commit!

### DIFF
--- a/src/scripts/cfo.zig
+++ b/src/scripts/cfo.zig
@@ -75,14 +75,24 @@ const Fuzzer = enum {
     vsr_superblock,
     vopr,
     vopr_testing,
+    vopr_lite,
+    vopr_testing_lite,
 
     fn print_command(fuzzer: Fuzzer, shell: *Shell, seed: u64) ![]const u8 {
-        if (fuzzer == .vopr or fuzzer == .vopr_testing) {
-            const state_machine: []const u8 =
-                if (fuzzer == .vopr) "" else " -Dsimulator-state-machine=testing";
+        if (fuzzer == .vopr or fuzzer == .vopr_testing or
+            fuzzer == .vopr_lite or fuzzer == .vopr_testing_lite)
+        {
+            const state_machine: []const u8 = if (fuzzer == .vopr or fuzzer == .vopr_lite)
+                ""
+            else
+                " -Dsimulator-state-machine=testing";
+            const lite: []const u8 = if (fuzzer == .vopr or fuzzer == .vopr_testing)
+                ""
+            else
+                " --lite";
             return try shell.print(
-                "./zig/zig build -Drelease{s} simulator_run -- {d}",
-                .{ state_machine, seed },
+                "./zig/zig build -Drelease{s} simulator_run --{s} {d}",
+                .{ state_machine, lite, seed },
             );
         }
         return try shell.print(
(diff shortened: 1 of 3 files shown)


src/scripts/cfo.zig
```diff
@@ -202,7 +226,7 @@ fn run_fuzzers(
                     try shell.pushd(working_directory);
                     defer shell.popd();
 
-                    assert(try shell.dir_exists(".git"));
+                    assert(try shell.dir_exists(".git") or shell.file_exists(".git"));
```
matklad, line 224: for the curious reader, in my repo

```
λ cat .git
gitdir: /home/matklad/p/tb/.bare/worktrees/work
```

src/scripts/cfo.zig
```diff
@@ -92,15 +102,24 @@ const Fuzzer = enum {
...
+            fuzzer == .vopr_lite or fuzzer == .vopr_testing_lite)
+        {
+            const state_machine: []const []const u8 = if (fuzzer == .vopr or fuzzer == .vopr_lite)
+                &.{}
+            else
+                &.{"-Dsimulator-state-machine=testing"};
+            const lite: []const []const u8 = if (fuzzer == .vopr or fuzzer == .vopr_testing)
+                &.{}
+            else
+                &.{"--lite"};
```
sentientwaffle, line 115: Why `.{"string"}` instead of `"string"`?

src/scripts/cfo.zig
```diff
@@ -58,6 +58,9 @@ pub const CliArgs = struct {
     budget_minutes: u64 = 10,
     hang_minutes: u64 = 30,
     concurrency: ?u32 = null,
+    // Don't checkout a separate repository, and instead fuzz in-place. Useful for debugging
+    // CFO itself.
+    no_checkout: bool = false,
```
sentientwaffle: Alternatively, we could let `gh_token == null` imply `no_checkout`.
Since if `gh_token == null` we aren't downloading extra branches, either.

src/scripts/cfo.zig
matklad: This interpolates not into a string, but into an array of CLI arguments. The shell syntax supports splats: in `"foo {bar}"`, `bar` can be either a string, or a list of strings, and it'll do the right thing in both cases. This gives raise to the idiom of optional arguments `bar` can be a singleton or an empty list.

We could also add explicit support for optionals to shell, so that you can intrepolate `?[]const u8`.

Somewhat orthogonally, but perhaps we should split `{bar}` sytnax into `{bar}`, `{bar...}` and `{bar?}` to sintactically signify the artity of interpolation (the arity is "obvious" from the types, but an extra mark could perhaps make the code a bit more readable)

src/scripts/cfo.zig
matklad: Excellent suggestion, thanks!

src/scripts/cfo.zig
matklad: Again, hope to refactor this whole thing away!

sentientwaffle: Ah ok, I was just confused why you didn't use `""` for the "no argument" case.

sentientwaffle (approved):

sentientwaffle:
Another idea for vopr-lite -- since vopr-lite ignores liveness issues anyhow, we could greatly increase the `requests_max` (e.g. instead of `constants.journal_slot_count * 3`, we could set `constants.journal_slot_count * 30`).

matklad:
Wouldn't that make it heavier though?

src/scripts/cfo.zig
matklad: I fully admit that this particular part of the file is write-only :sob:

sentientwaffle:
Yeah, that's true. Maybe once the liveness issues are solved we can add a vopr-heavy which has higher requests + max ticks...

PR #1925, opened 2024-05-08
io: fix path handling on Windows by switching to NtCreateFile

cb22:
Previously we were using `kernel32.CreateFileW` which had 2 problems:
* No way to provide a dir_fd, so we were always opening `basename(path)`, causing us to ignore paths!
* Didn't check if the file exists (so `tigerbeetle format` would overwrite files).

Switch to vendoring Zig's OpenFile, which uses `ntdll.NtCreateFile` and has neither of these issues. Vendoring was required, since we need to pass in file attributes, which the Zig implementation didn't allow for by default.

Before:
```
~/g/t/tigerbeetle-7> ./tigerbeetle.exe format --replica-count=1 --cluster=0 --replica=0
info(io): creating "foo"...
info(io): allocating 1.06298828125GiB...
info(main): 0: formatted: cluster=0 replica_count=1

~/g/t/tigerbeetle-7> ./tigerbeetle.exe format --replica-count=1 --cluster=0 --replica=0
info(io): creating "foo"...
info(io): allocating 1.06298828125GiB...
info(main): 0: formatted: cluster=0 replica_count=1

~/g/t/tigerbeetle-7> rm foo

~/g/t/tigerbeetle-7> ./tigerbeetle.exe format --replica-count=1 --cluster=0 --replica=0 /tmp/foo
info(io): creating "foo"...
info(io): allocating 1.06298828125GiB...
info(main): 0: formatted: cluster=0 replica_count=1

~/g/t/tigerbeetle-7> ls -l foo
.rw-r--r-- 1.1G federico  8 May 20:43 foo
```

After:
```
~/g/t/tigerbeetle-7> ./tigerbeetle.exe format --replica-count=1 --cluster=0 --replica=0 /tmp/foo
info(io): creating "foo"...
info(io): allocating 1.06298828125GiB...
info(main): 0: formatted: cluster=0 replica_count=1

~/g/t/tigerbeetle-7> ./tigerbeetle.exe format --replica-count=1 --cluster=0 --replica=0 /tmp/foo
info(io): creating "foo"...
error: PathAlreadyExists

~/g/t/tigerbeetle-7> ls -l /tmp/foo
.rw-r--r-- 1.1G federico  8 May 20:44 foo
```

(the above samples are running through Wine)

### DIFF
--- a/src/io/windows.zig
+++ b/src/io/windows.zig
@@ -957,7 +957,11 @@ pub const IO = struct {
 
     pub const INVALID_FILE = os.windows.INVALID_HANDLE_VALUE;
 
-    fn open_file_handle(relative_path: []const u8, method: enum { create, open }) !os.fd_t {
+    fn open_file_handle(
+        dir_handle: os.fd_t,
+        relative_path: []const u8,
+        method: enum { create, open },
+    ) !os.fd_t {
         const path_w = try os.windows.sliceToPrefixedFileW(relative_path);
 
         // FILE_CREATE = O_CREAT | O_EXCL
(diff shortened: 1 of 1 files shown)


cb22:
Hm, this seems to fail on real Windows systems. I'll double check what's going on tomorrow.

cb22:
> Hm, this seems to fail on real Windows systems. I'll double check what's going on tomorrow.

Turns out a few things:

- I was originally passing the flags in the wrong place to `NtCreateFile`; they go with the existing flags like `os.windows.FILE_NON_DIRECTORY_FILE`, not where `FILE_ATTRIBUTE_NORMAL` goes.
- Zig's `GENERIC_READ` seems to be only the read permission, not the full table described [here](https://learn.microsoft.com/en-us/windows/win32/api/winternl/nf-winternl-ntcreatefile?redirectedfrom=MSDN).
- When using ntdll directly, the flags are different: `FILE_NO_INTERMEDIATE_BUFFERING` and `FILE_WRITE_THROUGH` compared to `FILE_FLAG_NO_BUFFERING` and `FILE_FLAG_WRITE_THROUGH` respectively.

Somewhat amusing that it worked on Wine, but tested on a real Windows system (and CI!) now too.

src/io/windows.zig
```diff
@@ -977,31 +981,37 @@ pub const IO = struct {
         const shared_mode: os.windows.DWORD = 0;
 
         // O_RDWR
+        // Zig's mask seems wonky; according to
+        // https://learn.microsoft.com/en-us/windows/win32/api/winternl/nf-winternl-ntcreatefile
+        // FILE_GENERIC_READ should include SYNCHRONIZE but it does not.
```
matklad, line 986: Could you file an upstream issue here?

matklad (approved):
I haven't reviewed the intricacies of windows APIs here, but, given that this was quiet broken before, and now it works in tests, this should be an improvement!

src/io/windows.zig
cb22: Dug a bit further, it's not an upstream issue, but rather confusing naming.

`GENERIC_READ`, which you can use for files too, is different from `FILE_GENERIC_READ` which includes `SYNCHRONIZE`.

I'll put up a small PR to update the comment.

PR #1927, opened 2024-05-08
cfo: implement simple minimization and prioritization

matklad:
Fast failures are likely simple. Moreover, any correctness (as opposed
to liveness) failure is likely fast. So let's just order seeds that fail fast before those that fail slower


matklad:
I have a feeling that this might actually work quite great, despite being "they used a pencil" kind of solution!

sentientwaffle (approved):

PR #1931, opened 2024-05-10
go: run tests in parallel 

matklad:
We guarantee that the same client is safe to use concurrently from many
goroutines. While we do have a dedicated concurrency max tests in place
here, let's also run all the other tests in parallel, just so we can
exercise more weird states.


matklad:
To be merged after https://github.com/tigerbeetle/tigerbeetle/pull/1932

kprotty (approved):

PR #1935, opened 2024-05-10
Multiversion: backport execing and hardcode 0.15.3 as having 0.15.4 available

cb22:
0.15.4 will be the first version with the ability to read the multiversion metadata embedded in TigerBeetle. This presents a bit of a bootstrapping problem:

* Operator replaces 0.15.3 with 0.15.4,
* 0.15.4 starts up, re-execs into 0.15.3,
* 0.15.3 knows nothing about 0.15.4 or how to check it's available, so we just hang.

Work around this by hardcoding that if 0.15.3 is present in a pack, 0.15.4 must be too.

Additionally, fix up setting `CLOEXEC` on the listening sockets. We were only setting it correctly on the accepted sockets.

### DIFF
--- a/src/tigerbeetle/main.zig
+++ b/src/tigerbeetle/main.zig
@@ -33,6 +33,8 @@ const SuperBlock = vsr.SuperBlockType(Storage);
 const superblock_zone_size = vsr.superblock.superblock_zone_size;
 const data_file_size_min = vsr.superblock.data_file_size_min;
 
+var replica_release_execute_args: ?std.process.ArgIterator = null;
+
 pub const std_options = struct {
     pub const log_level: std.log.Level = constants.log_level;
     pub const logFn = constants.log;
@@ -58,6 +60,11 @@ pub fn main() !void {
 
     var arg_iterator = try std.process.argsWithAllocator(allocator);
     defer arg_iterator.deinit();
+
+    // Store a copy of the arg iterator, so it can be used later to determine the
+    // binary path. Only strictly needed on Windows.
+    replica_release_execute_args = arg_iterator;
+    defer replica_release_execute_args = null;
 
     var command = try cli.parse_args(allocator, &arg_iterator);
     defer command.deinit(allocator);
@@ -186,13 +193,26 @@ const Command = struct {
         const nonce = std.crypto.random.int(u128);
         assert(nonce != 0); // Broken CSPRNG is the likeliest explanation for zero.
 
+        // Ensure that this code is only ever built as minimum or 0.15.3.
+        comptime assert(
+            config.process.release.value == vsr.Release.minimum.value or
+                config.process.release.value == vsr.Release.from(.{
+                .major = 0,
+                .minor = 15,
+                .patch = 3,
+            }).value,
+        );
+
         var replica: Replica = undefined;
         replica.open(allocator, .{
             .node_count = @intCast(args.addresses.len),
             .release = config.process.release,
             // TODO Where should this be set?
             .release_client_min = config.process.release,
-            .releases_bundled = &[_]vsr.Release{config.process.release},
+            .releases_bundled = &[_]vsr.Release{
+                config.process.release,
+                vsr.Release{ .value = config.process.release.value + 1 },
+            },
             .release_execute = replica_release_execute,
             .storage_size_limit = args.storage_size_limit,
             .storage = &command.storage,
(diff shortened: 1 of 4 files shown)


cb22:
@sentientwaffle this is ready for review; but leaving as a draft PR since the merging / build process will be... Interesting.

src/tigerbeetle/main.zig
```diff
@@ -348,8 +385,72 @@ fn replica_release_execute(replica: *Replica, release: vsr.Release) noreturn {
...
+            const cast_args: [*:null]const ?[*:0]const u8 = @ptrCast(std.os.argv.ptr);
+            const cast_envp: [*:null]const ?[*:0]const u8 = @ptrCast(std.os.environ.ptr);
+
+            std.log.info("replica_release_execute: executing '{s}'...\n", .{binary_path});
+            std.os.execveZ(binary_path, cast_args, cast_envp) catch @panic("execveZ failed");
+            unreachable;
+        },
+        .windows => {
+            var binary_path = replica_release_execute_args.?.next().?;
+            assert(std.fs.path.isAbsolute(binary_path));
```
sentientwaffle: Is the global `replica_release_execute_args` needed by "real" multiversioning too?

If so, maybe we should just use that instead of `std.os.argv` for macos/linux too, to avoid some of the branching on `.windows` at the top-level and in `main()`.

...Then again, which how much type casting is going on, maybe it wouldn't actually simplify anything.

src/tigerbeetle/main.zig
```diff
@@ -348,8 +385,72 @@ fn replica_release_execute(replica: *Replica, release: vsr.Release) noreturn {
...
+
+            // We can pass through our env and args as-is to exec. We have to manipulate the types
+            // here somewhat: they're cast in start.zig and we can't access `argc_argv_ptr`
+            // directly.
+            // process.zig does the same trick in execve().
+            const cast_args: [*:null]const ?[*:0]const u8 = @ptrCast(std.os.argv.ptr);
+            const cast_envp: [*:null]const ?[*:0]const u8 = @ptrCast(std.os.environ.ptr);
+
+            std.log.info("replica_release_execute: executing '{s}'...\n", .{binary_path});
+            std.os.execveZ(binary_path, cast_args, cast_envp) catch @panic("execveZ failed");
```
sentientwaffle: If `execveZ` fails, the error would be nice to have. Maybe replace the panic with a `|err| log.fatal(..., .{err})`? (Likewise for windows.)

src/io/windows.zig
```diff
@@ -956,7 +956,7 @@ pub const IO = struct {
 
     pub const INVALID_FILE = os.windows.INVALID_HANDLE_VALUE;
 
-    fn open_file_handle(relative_path: []const u8, method: enum { create, open }) !os.fd_t {
+    fn open_file_handle(relative_path: []const u8, method: enum { create, open, open_wait }) !os.fd_t {
```
sentientwaffle: Just curious, but why not `.deinit()` the `IO` to unlock the data file before starting the new process?

cb22: No reason - that's a great idea! I think I should be able to deinit the replica and storage; afaik it's the port and the data file that we need to ensure are free.

(I was using the file as a proxy for both, here)

src/tigerbeetle/main.zig
cb22: > Is the global replica_release_execute_args needed by "real" multiversioning too?

Nope, there it's handled inside `multiversioning.zig`.

I've removed the branch for now, at the cost of a little extra memory on Linux / macOS, but since this is a transitional release only I think that's OK!

src/tigerbeetle/main.zig
```diff
@@ -348,8 +368,84 @@ fn replica_release_execute(replica: *Replica, release: vsr.Release) noreturn {
...
+    switch (builtin.os.tag) {
+        .macos, .linux => {
+            // We can pass through our env and args as-is to exec. We have to manipulate the types
+            // here somewhat: they're cast in start.zig and we can't access `argc_argv_ptr`
+            // directly.
+            // process.zig does the same trick in execve().
+            const cast_args: [*:null]const ?[*:0]const u8 = @ptrCast(std.os.argv.ptr);
+            const cast_envp: [*:null]const ?[*:0]const u8 = @ptrCast(std.os.environ.ptr);
+
+            std.log.info("replica_release_execute: executing '{s}'...\n", .{binary_path});
```
sentientwaffle, line 386: Nit, but you could use `log_main` for the scoped logger.

Also, `log.info()` doesn't need an explicit `\n`.

sentientwaffle (approved):

src/tigerbeetle/main.zig
cb22: > Nit, but you could use log_main for the scoped logger.

Ah yes!

> Also, log.info() doesn't need an explicit `\n`.

Indeed - this is to make the logging output when its exec'ing through multiple versions easier to read - that'll be the last line printed before the new process takes over.

cb22:
Closing this PR - I've pushed it as a tag, https://github.com/tigerbeetle/tigerbeetle/releases/tag/0.15.3-multiversion-1, and that tag will be referenced in `build.zig` for 0.15.4.

PR #1938, opened 2024-05-10
state_machine: allow useless pulses

matklad:
Primary decides whether to pulse or not, based on it's know (or unknown) next transfer to expire. But in general different replicas might have different state knowledge about pending expiring transfer.

So, a primary might pulse because it _doesn't_ know, but then the backup would know _not_ to pulse, and yet it sill must process the pulse.

So we definietly should remove the assert, and instead:

* either don't do the scan at all
* or assert that it returned zero results

This PR take the second option as it makes assertions denser.

Seed:  ./zig/zig build  simulator_run -- --lite 11775367263850621025

### DIFF
--- a/src/state_machine.zig
+++ b/src/state_machine.zig
@@ -1053,7 +1053,7 @@ pub fn StateMachineType(
             const self: *StateMachine = ScanLookup.parent(.expire_pending_transfers, scan_lookup);
             assert(self.scan_lookup_result_count == null);
 
-            self.expire_pending_transfers.finish(scan_lookup.state);
+            self.expire_pending_transfers.finish(scan_lookup.state, results);
             self.scan_lookup_result_count = @intCast(results.len);
 
             self.scan_lookup = .null;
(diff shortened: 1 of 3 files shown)


matklad:
One alternative here would be to reset `expire_pending_transfers` at a checkpoint boundary, but I think it's simpler (and faster) to just not try to force more determinism here than we have.

matklad:
Current failure:

```
./zig/zig build simulator_run -- --lite 8112484246554900713
```

matklad:
That failure goes away if I revert the last commit (disabling reset on view change).

I don't understand why though. 

On that seed, auditor expires transfer `340282366920938463463374607431768211090`, while the replicas expire `340282366920938463463374607431768211085` instead

matklad:
Oh, this is curious, both of these two transfers expire at exactly identical nanosecond, at `1619300000005`.

matklad:
Ahh, so this is at least a bug in auditor:

```
                        self.pending_expiries.add(.{
                            .transfer = transfer.id,
                            .timestamp = transfer_timestamp + transfer.timeout_ns(),
                        }) catch unreachable;
```

`pending_expiries` is a priority Q which is ordered by the (expiery) `.timestamp`, but as we have a collition here, the order of items isn't actually well-defined!

matklad:
Yeah, I belive this is unrelated to the change actually, fixed in https://github.com/tigerbeetle/tigerbeetle/pull/1957

sentientwaffle (approved):
Makes sense! I like `value_next_expired_at` -- avoiding updates to `pulse_next_timestamp` from "inside" the scan is easier to follow.

PR #1939, opened 2024-05-10
flags: fix parsing of multiple positional fields

matklad:
It never actually worked, we had incorrect gold values in tests! We only found this out now, when `fuzz_tests` started to use several arguments!

### DIFF
--- a/src/flags.zig
+++ b/src/flags.zig
@@ -207,21 +207,23 @@ fn parse_flags(args: *std.process.ArgIterator, comptime Flags: type) Flags {
 
         if (@hasField(Flags, "positional")) {
             counts.positional += 1;
-            inline for (positional_fields, 0..) |positional_field, positional_index| {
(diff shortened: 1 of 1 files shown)


src/flags.zig
```diff
@@ -851,7 +853,7 @@ test "flags" {
 
     try t.check(&.{ "pos", "x", "y" }, snap(@src(),
         \\stdout:
-        \\p1: y
+        \\p1: x
```
matklad, line 856: We had this particular test since forever. This is the dark side of snaptests --- might be too easy to gloss over an incorrect result!

matklad: But (all) fuzz tests caught us here :rofl: 

image

sentientwaffle (approved):
Ahh, that'll teach me to not skim past snapshot diffs. :facepalm:

PR #1946, opened 2024-05-13
VSR/Client: Add `vsr.Client.register()`

sentientwaffle:
### Context

Clients (at the level above `vsr.Client`) are responsible for automatically coalescing user operations into batches. The maximum batch size is currently `message_body_size_max`.

In order to minimize replica memory usage when maximum throughput is not needed (e.g. during development), the replica is going to support running with a lower, _runtime-defined_ batch size limit. Since this batch size limit impacts the clients, the replica must communicate it to clients. This will occur via the result of the `register` request. (Note that this is not implemented in this PR – this PR is a prerequisite.)

Previously, `vsr.Client` would send its `command=request operation=register` message when the first of the following occurred:

  1. `vsr.Client.request()` is called, or
  2. the client receives a `pong_client` message.

But that means that the first batch created by the client would potentially precede the client actually knowing the batch limit.

### Solution

Add `vsr.Client.register(callback, user_data)`. `vsr.Client.request()` cannot be called until `.register()` has invoked its callback.

`tb_client` invoked `vsr.Client.register` automatically during its setup – no waiting until `request()`/`on_pong()`. It will refrain from generating/sending any batches until the register reply arrives.

### Future Work (Maybe?)

The above opens up some nice possibilities for UX of the client API:

- If `.register` takes too long, we could bubble up an exception/warning to the application suggesting that they might have misconfigured TB (e.g. wrong host/port/cluster_id). (Right now the client just hangs silently forever, which is not a great developer experience.)
- We could expose `.register()`, to allow developers to (optionally) wait until the connection is established before they start generating work – since any generated work would just be buffered in memory until `register` arrives.

---

~Note that CI (`replica_test.zig`) is failing on this branch. This is semi-unrelated; it is addressed in a [separate PR](https://github.com/tigerbeetle/tigerbeetle/pull/1947).~

### DIFF
--- a/src/message_pool.zig
+++ b/src/message_pool.zig
@@ -30,7 +30,6 @@ pub const Options = union(vsr.ProcessType) {
                 sum += constants.replicas_max; // Connection.recv_message
                 // Connection.send_queue:
                 sum += constants.replicas_max * constants.connection_send_queue_max_client;
-                sum += 1; // Client.register_inflight
                 sum += 1; // Client.request_inflight
                 // Handle bursts.
                 // (e.g. Connection.parse_message(), or sending a ping when the send queue is full).
--- a/src/aof.zig
+++ b/src/aof.zig
@@ -323,6 +323,12 @@ pub const AOFReplayClient = struct {
         );
         errdefer client.deinit(allocator);
 
+        client.register(register_callback, undefined);
+        while (client.request_inflight != null) {
+            client.tick();
+            try io.run_for_ns(constants.tick_ms * std.time.ns_per_ms);
+        }
+
         return Self{
             .io = io,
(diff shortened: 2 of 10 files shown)


src/message_pool.zig
```diff
@@ -30,7 +30,6 @@ pub const Options = union(vsr.ProcessType) {
                 sum += constants.replicas_max; // Connection.recv_message
                 // Connection.send_queue:
                 sum += constants.replicas_max * constants.connection_send_queue_max_client;
-                sum += 1; // Client.register_inflight
```
batiati, line 33: :heart:

batiati (approved):

PR #1949, opened 2024-05-13
docs: remove needless scripts

matklad:
* there's no need for `install.sh` when it's just `./scripts/install_zig.sh && ./zig/zig build -Drelease -Dconfig=production`
* there's no need to call `zig build install`, install is the default step
* git hook uses the wrong zig, and in general hooks are very developer-specific


README.md
```diff
@@ -179,7 +179,8 @@ First grab the sources and run the setup script:
 ```console
 git clone https://github.com/tigerbeetle/tigerbeetle.git
 cd tigerbeetle
-scripts/install.sh
+./scripts/install_zig.sh
```
cb22: Ah! Let's point the user to the quick start instructions above? They no longer need source to benchmark.

.github/ci/benchmark/tigerbeetle.hcl
```diff
@@ -99,7 +99,7 @@ git clone ${var.git_url}
 cd tigerbeetle
 git checkout ${var.git_ref}
 ./scripts/install_zig.sh
-./zig/zig build install -Drelease -Dconfig=production
+./zig/zig build -Drelease -Dconfig=production
```
cb22: We can remove this whole folder too, unless you think it's useful to point people to... It's not actively used.

cb22:
(the macOS yaml still references install.sh)

.github/ci/benchmark/tigerbeetle.hcl
matklad: :+1:

README.md
matklad: I removed this section altogether. I don't think it's possible to meaningfully use these benchmarks unless you know quite a bit more than we are saying here.

cb22 (approved):

PR #1951, opened 2024-05-14
vsr: new state sync protocol 

matklad:
This commit implements the new state sync protocol, which fixes a couple
of long-standing liveness issues. As an unintended bonus, the new
implementation ends up being simpler, at least if measured in lines of
code!

 ## Old Protocol Problems

The original thinking was to implement state-sync as a separate
protocol, mostly orthogonal to the rest of VSR. VSR maintains the log,
state sync maintains the checkpoint. Checkpoints are committed, so
there's no doubt regarding _what_ to sync, so a separate protocol sounds
easy enough, right?

Well, turns out its' not that simple, as the two issues were discovered:

_First_, the log and the checkpoint might become physically disjoint
(checkpoint ahead of the log). The implementation fixed this case by
maneuvering the replica into recovering_head, but that sadly created a
liveness issue --- recovering_head replicas don't participate in
consensus, so it is possible for cluster to get stuck (with most of the
cluster in recovering_head) even without any storage faults!

_Second_, the log and the checkpoint might become logically inconsistent
(the message in the log for the checkpoint's slot might be from the
different view). Although checkpoint is committed, that's not true about
the log! This issue manifested itself as assertion failures in
`valid_hash_chain`. There were several band-aid fixes for the failures,
which boiled down to maneuvering replica into recovering_head if
checkpoint's view and log view are different. But we weren't able to
plug all the wholes, as it slowly became apparent that fixing everything
would require duplicating VSR-specfic SV handling logic when dealing
with checkpoints.

 ## The New Protocol

The idea is rather simple --- add primary's checkpoint to SV and handle
state sync in `on_start_view`. This makes it so that the checkpoint and
log are consistent by construction. This also yields an unexpected
benefit: a replica no longer ignores SV's which are too far in the
future, it is always possible to install a header from SV (although you
might need to update the checkpoint as well).

The implementation is somewhat tricky --- we update checkpoint
asynchronously, but we update replica state immediately, so there's a
window while our `op` is consistent with in-memory checkpoint, but not
on-disc checkpoint. This signalled by

    self.syncing == .updating_superblock

state. While in this state, the replica mostly just waits for the
superblock update to finish and doesn't do much else.

Additionally, the implementation right now is optimized for smaller
diff, rather than for path dependent perfection. For example, we still
call it `Superblock.view_change`, even though now this function can be
called _only_ to advance the checkpoint state. I want to do a follow up
cleanup PR. There's perhaps more code/asserts that needs to be adjusted
now that `.recovering_head` is again possible only after `.recovering`.

The equivalent code have been running in VOPR for quite some time, and I
think any sync-consistency related issues are fixed. The one current
outstading VOPR bug (with several seeds) is an orthogonal issue where we
allow to sync to a checkpoint before it's sync_content is done.

This particular commit has some extra assertions and cleanups, so they
might still blow up, but hopefully would be trivial to fix.

### DIFF
--- a/src/constants.zig
+++ b/src/constants.zig
@@ -322,7 +322,10 @@ comptime {
     assert(view_change_headers_max > 0);
     assert(view_change_headers_max >= pipeline_prepare_queue_max + 3);
     assert(view_change_headers_max <= journal_slot_count);
-    assert(view_change_headers_max <= @divFloor(message_body_size_max, @sizeOf(vsr.Header)));
+    assert(view_change_headers_max <= @divFloor(
+        message_body_size_max - @sizeOf(vsr.CheckpointState),
+        @sizeOf(vsr.Header),
+    ));
     assert(view_change_headers_max > view_change_headers_suffix_max);
 }
 
--- a/src/testing/cluster.zig
+++ b/src/testing/cluster.zig
@@ -796,10 +796,6 @@ pub fn ClusterType(comptime StateMachineType: anytype) type {
                     };
                 },
                 .sync_stage_changed => switch (replica.syncing) {
-                    .requesting_checkpoint => cluster.log_replica(
-                        .sync_commenced,
-                        replica.replica,
-                    ),
                     .idle => cluster.log_replica(.sync_completed, replica.replica),
                     else => {},
                 },
--- a/src/testing/cluster.zig
+++ b/src/testing/cluster.zig
@@ -825,7 +825,6 @@ pub fn ClusterType(comptime StateMachineType: anytype) type {
                 commit = ' ',
                 checkpoint_commenced = '[',
                 checkpoint_completed = ']',
-                sync_commenced = '<',
                 sync_completed = '>',
             },
             replica_index: u8,
--- a/src/vsr.zig
+++ b/src/vsr.zig
@@ -203,7 +203,7 @@ pub const Command = enum(u8) {
 
     start_view_change = 10,
     do_view_change = 11,
-    start_view = 12,
+    start_view_deprecated = 12,
 
     request_start_view = 13,
     request_headers = 14,
@@ -222,7 +222,7 @@ pub const Command = enum(u8) {
     // A reserved command for the new state sync protocol. At the moment, replica ignores this
     // command (as opposed to panicking on an unknown command), to allow the next release to send
     // both versions of StartView.
-    start_view2 = 23,
+    start_view = 23,
 
     comptime {
         for (std.enums.values(Command), 0..) |result, index| {
--- a/src/vsr.zig
+++ b/src/vsr.zig
@@ -203,6 +203,8 @@ pub const Command = enum(u8) {
 
     start_view_change = 10,
     do_view_change = 11,
+    // Historical version of SV, without checkpoint. Currently  both `start_view_deprecated` and
+    // `start_view` are handled. Next release will ignore `start_view_deprecated`.
     start_view_deprecated = 12,
 
     request_start_view = 13,
@@ -219,9 +221,6 @@ pub const Command = enum(u8) {
     request_sync_checkpoint = 21,
     sync_checkpoint = 22,
 
-    // A reserved command for the new state sync protocol. At the moment, replica ignores this
-    // command (as opposed to panicking on an unknown command), to allow the next release to send
-    // both versions of StartView.
     start_view = 23,
 
     comptime {
--- a/src/vsr/replica.zig
+++ b/src/vsr/replica.zig
@@ -4638,7 +4638,7 @@ pub fn ReplicaType(
         /// Construct a SV message, including attached headers from the current log_view.
         /// The caller owns the returned message, if any, which has exactly 1 reference.
         fn create_start_view_message(self: *Self, nonce: u128) struct {
-            current: *Message.StartView,
+            current: *Message.StartViewDeprecated,
             next: *Message.StartView,
         } {
             assert(self.status == .normal);
@@ -4689,6 +4689,7 @@ pub fn ReplicaType(
             message.header.set_checksum_body(message.body());
             message.header.set_checksum();
 
+            comptime assert(Message.StartView == Message.StartViewDeprecated);
             const message_copy = self.message_bus.get_message(.start_view);
             defer self.message_bus.unref(message_copy);
 
--- a/src/vsr/replica.zig
+++ b/src/vsr/replica.zig
@@ -2225,11 +2225,7 @@ pub fn ReplicaType(
             }
             assert(self.view == message.header.view);
 
-            const view_checkpoint = std.mem.bytesAsValue(
-                vsr.CheckpointState,
-                message.body()[message.body().len -
-                    @sizeOf(vsr.CheckpointState) ..][0..@sizeOf(vsr.CheckpointState)],
-            );
+            const view_checkpoint = start_view_message_checkpoint(message);
             if (vsr.Checkpoint.trigger_for_checkpoint(view_checkpoint.header.op)) |trigger| {
                 assert(message.header.commit >= trigger);
             }
@@ -2239,10 +2235,9 @@ pub fn ReplicaType(
                 ).?,
             );
 
-            const view_headers = message_body_as_view_headers(message.base_const());
-            assert(view_headers.command == .start_view);
-            assert(view_headers.slice[0].op == message.header.op);
-            assert(view_headers.slice[0].op >= view_headers.slice[view_headers.slice.len - 1].op);
+            const view_headers = start_view_message_headers(message);
+            assert(view_headers[0].op == message.header.op);
+            assert(view_headers[0].op >= view_headers[view_headers.len - 1].op);
 
             if (message.header.commit > self.op_prepare_max() and (
             //  Cluster is far ahead, replica's WAL no longer intersects primary's WAL.
@@ -2312,10 +2307,10 @@ pub fn ReplicaType(
                 // Replace our log with the suffix from SV. Transition to sync above guarantees
                 // that there's at least one message that fits the effective checkpoint, but some
                 // messages might be beyond its prepare_max.
-                maybe(view_headers.slice[0].op > self.op_prepare_max_sync());
+                maybe(view_headers[0].op > self.op_prepare_max_sync());
 
                 // Find the first message that fits, make it our new head.
-                for (view_headers.slice) |*header| {
+                for (view_headers) |*header| {
                     assert(header.commit <= message.header.commit);
 
                     if (header.op <= self.op_prepare_max_sync()) {
@@ -2330,7 +2325,7 @@ pub fn ReplicaType(
                     }
                 } else unreachable;
 
-                for (view_headers.slice) |*header| {
+                for (view_headers) |*header| {
                     if (header.op <= self.op_prepare_max_sync()) {
                         self.replace_header(header);
                     }
@@ -2343,7 +2338,7 @@ pub fn ReplicaType(
                 self.commit_min = self.syncing.updating_superblock.checkpoint_state.header.op;
             }
 
-            self.view_headers.replace(.start_view, view_headers.slice);
+            self.view_headers.replace(.start_view, view_headers);
             assert(self.view_headers.array.get(0).view <= self.view);
             assert(self.view_headers.array.get(0).op == message.header.op);
             maybe(self.view_headers.array.get(0).op > self.op_prepare_max_sync());
@@ -4616,6 +4611,7 @@ pub fn ReplicaType(
         /// The caller owns the returned message, if any, which has exactly 1 reference.
         fn create_start_view_message(self: *Self, nonce: u128) *Message.StartView {
             assert(self.status == .normal);
+            assert(self.syncing != .updating_superblock);
             assert(self.replica == self.primary_index(self.view));
             assert(self.commit_min == self.commit_max);
             assert(self.commit_min <= self.op);
@@ -4632,8 +4628,8 @@ pub fn ReplicaType(
             defer self.message_bus.unref(message);
 
             message.header.* = .{
-                .size = @sizeOf(Header) * (1 + self.view_headers.array.count_as(u32)) +
-                    @sizeOf(vsr.CheckpointState),
+                .size = @sizeOf(Header) + @sizeOf(vsr.CheckpointState) +
+                    @sizeOf(Header) * self.view_headers.array.count_as(u32),
                 .command = .start_view,
                 .cluster = self.cluster,
                 .replica = self.replica,
@@ -4644,20 +4640,18 @@ pub fn ReplicaType(
                 .nonce = nonce,
             };
 
-            const body_headers_size = self.view_headers.array.count_as(usize) * @sizeOf(vsr.Header);
-            const body_headers = message.body()[0..body_headers_size];
-            const body_checkpoint = message.body()[body_headers_size..];
-            stdx.copy_disjoint(
-                .exact,
-                Header.Prepare,
-                std.mem.bytesAsSlice(Header.Prepare, body_headers),
-                self.view_headers.array.const_slice(),
-            );
             stdx.copy_disjoint(
                 .exact,
                 u8,
-                body_checkpoint,
+                message.body()[0..@sizeOf(vsr.CheckpointState)],
                 std.mem.asBytes(&self.superblock.working.vsr_state.checkpoint),
+            );
+            comptime assert(@sizeOf(vsr.CheckpointState) % @sizeOf(Header) == 0);
+            stdx.copy_disjoint(
+                .exact,
+                u8,
+                message.body()[@sizeOf(vsr.CheckpointState)..],
+                std.mem.sliceAsBytes(self.view_headers.array.const_slice()),
             );
             message.header.set_checksum_body(message.body());
             message.header.set_checksum();
@@ -10208,13 +10202,11 @@ const DVCQuorum = struct {
 
 fn message_body_as_view_headers(message: *const Message) vsr.Headers.ViewChangeSlice {
     assert(message.header.size > @sizeOf(Header)); // Body must contain at least one header.
-    assert(message.header.command == .do_view_change or
-        message.header.command == .start_view);
+    assert(message.header.command == .do_view_change);
 
     return vsr.Headers.ViewChangeSlice.init(
         switch (message.header.command) {
             .do_view_change => .do_view_change,
-            .start_view => .start_view,
             else => unreachable,
         },
         message_body_as_headers_unchecked(message),
@@ -10225,8 +10217,7 @@ fn message_body_as_view_headers(message: *const Message) vsr.Headers.ViewChangeS
 /// The headers may contain gaps and/or breaks.
 fn message_body_as_prepare_headers(message: *const Message) []const Header.Prepare {
     assert(message.header.size > @sizeOf(Header)); // Body must contain at least one header.
-    assert(message.header.command == .start_view or
-        message.header.command == .headers);
+    assert(message.header.command == .headers);
 
     const headers = message_body_as_headers_unchecked(message);
     var child: ?*const Header.Prepare = null;
@@ -10250,18 +10241,40 @@ fn message_body_as_prepare_headers(message: *const Message) []const Header.Prepa
 fn message_body_as_headers_unchecked(message: *const Message) []const Header.Prepare {
     assert(message.header.size > @sizeOf(Header)); // Body must contain at least one header.
     assert(message.header.command == .do_view_change or
-        message.header.command == .start_view or
         message.header.command == .headers);
-
-    const headers_end_offset = switch (message.header.command) {
-        .start_view => message.header.size - @sizeOf(vsr.CheckpointState),
-        else => message.header.size,
-    };
 
     return std.mem.bytesAsSlice(
         Header.Prepare,
-        message.buffer[@sizeOf(Header)..headers_end_offset],
+        message.body(),
     );
+}
+
+fn start_view_message_checkpoint(message: *const Message.StartView) *const vsr.CheckpointState {
+    assert(message.header.command == .start_view);
+    const checkpoint = std.mem.bytesAsValue(
+        vsr.CheckpointState,
+        message.body()[0..@sizeOf(vsr.CheckpointState)],
+    );
+    assert(checkpoint.header.valid_checksum());
+    assert(stdx.zeroed(&checkpoint.reserved));
+    return checkpoint;
+}
+
+fn start_view_message_headers(message: *const Message.StartView) []const Header.Prepare {
+    assert(message.header.command == .start_view);
+    // Body must contain at least one header.
+    assert(message.header.size > @sizeOf(Header) + @sizeOf(vsr.CheckpointState));
+
+    const headers = std.mem.bytesAsSlice(
+        Header.Prepare,
+        message.body()[@sizeOf(vsr.CheckpointState)..],
+    );
+    // To run verification.
+    _ = vsr.Headers.ViewChangeSlice.init(.start_view, headers);
+    if (constants.verify) {
+        for (headers) |header| assert(header.valid_checksum());
+    }
+    return headers;
 }
 
 /// The PipelineQueue belongs to a normal-status primary. It consists of two queues:
--- a/src/vsr/replica.zig
+++ b/src/vsr/replica.zig
@@ -302,9 +302,6 @@ pub fn ReplicaType(
         opened: bool,
 
         syncing: SyncStage = .idle,
-        /// The latest discovered checkpoint that would be safe to sync to.
-        /// Kept up-to-date during every status, while syncing or healthy.
-        sync_target_max: ?SyncTarget = null,
         /// Invariants:
         /// - If syncing≠idle then sync_tables=null.
         sync_tables: ?ForestTableIterator = null,
@@ -507,7 +504,7 @@ pub fn ReplicaType(
         /// (status=normal or (status=view-change and primary))
         repair_timeout: Timeout,
 
-        /// The number of ticks before triggering state sync.
+        /// The number of ticks before checking whether state sync should be requested.
         /// This allows the replica to attempt WAL/grid repair before falling back, even if it
         /// is lagging behind the primary, to try to avoid unnecessary state sync.
         ///
@@ -521,10 +518,6 @@ pub fn ReplicaType(
 
         /// (always running)
         grid_scrub_timeout: Timeout,
-
-        /// The number of ticks before sending a sync message (message types depend on syncing).
-        /// (syncing≠idle)
-        sync_message_timeout: Timeout,
 
         /// The number of ticks on an idle cluster before injecting a `pulse` operation.
         /// (status=normal and primary and !constants.aof_recovery)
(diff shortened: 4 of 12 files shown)


src/constants.zig
```diff
@@ -322,7 +322,10 @@ comptime {
     assert(view_change_headers_max > 0);
     assert(view_change_headers_max >= pipeline_prepare_queue_max + 3);
     assert(view_change_headers_max <= journal_slot_count);
-    assert(view_change_headers_max <= @divFloor(message_body_size_max, @sizeOf(vsr.Header)));
+    assert(view_change_headers_max <= @divFloor(
+        message_body_size_max - @sizeOf(vsr.CheckpointState),
```
matklad, line 326: I randomly stuffed `vsr.CheckpointState` at the end, after the headers. Not sure where it should go at the front or at the back

src/vsr/replica.zig
```diff
@@ -517,10 +514,6 @@ pub fn ReplicaType(
         /// (always running)
         grid_scrub_timeout: Timeout,
 
-        /// The number of ticks before sending a sync message (message types depend on syncing).
-        /// (syncing≠idle)
-        sync_message_timeout: Timeout,
```
matklad, line 527: `repair_sync_timeout` now directly sends the requesting message (which is just request_start_view)

src/vsr/replica.zig
```diff
@@ -766,27 +759,22 @@ pub fn ReplicaType(
                     self.transition_to_normal_from_recovering_status();
                 }
             } else {
-                if (self.log_view < self.superblock.working.vsr_state.sync_view) {
-                    // During state sync, the replica installed CheckpointState from a future view.
-                    self.transition_to_recovering_head();
```
matklad, line 776: transition_to_recovering_head now only happens from `.recovering` right after open

src/vsr/replica.zig
```diff
@@ -1423,6 +1402,13 @@ pub fn ReplicaType(
                     });
                     return;
                 },
+                .request_sync_checkpoint, .sync_checkpoint => {
+                    log.warn("{}: on_message: ignoring old protocol ({s})", .{
+                        self.replica,
+                        @tagName(message.header.command),
+                    });
+                    return;
+                },
```
matklad, line 1418: Left the actual messages for now. I must say I am unsure how this whole things works with uprgades =/

src/vsr/replica.zig
```diff
@@ -1602,6 +1588,11 @@ pub fn ReplicaType(
             assert(message.header.replica < self.replica_count);
             assert(message.header.operation != .reserved);
 
+            if (self.syncing == .updating_superblock) {
+                log.debug("{}: on_prepare: ignoring (sync)", .{self.replica});
+                return;
+            }
```
matklad, line 1601: `self.syncing == .updating_superblock` is a weird state where we can't do anything useful, so I added a bunch of `if (`self.syncing == .updating_superblock`) return` where the vopr told me to!

src/vsr/replica.zig
```diff
@@ -2276,18 +2345,26 @@ pub fn ReplicaType(
...
+                    if (self.syncing == .updating_superblock) {
+                        self.view_durable_update();
+                    }
                     self.commit_journal();
                 },
-                .normal => {},
+                .normal => {
+                    if (self.syncing == .updating_superblock) {
+                        self.view_durable_update();
+                    }
```
matklad, line 2371: still need to come with a better name for `view_durable_update` as it no longer necessary updates the view. But I leave this for a follow up

src/vsr/replica.zig
```diff
@@ -2220,54 +2212,131 @@ pub fn ReplicaType(
...
+                // Replace our log with the suffix from SV. Transition to sync above guarantees
+                // that there's at least one message that fits the effective checkpoint, but some
+                // messages might be beyond its prepare_max.
+                maybe(view_headers.slice[0].op > self.op_prepare_max_sync());
+
+                // Find the first message that fits, make it our new head.
+                for (view_headers.slice) |*header| {
+                    assert(header.commit <= message.header.commit);
+
+                    if (header.op <= self.op_prepare_max_sync()) {
```
matklad, line 2316: NB: now use the `_sync` variant!

src/vsr/replica.zig
```diff
@@ -3075,57 +3111,28 @@ pub fn ReplicaType(
...
+                    "(op_head={} commit_min={} commit_max={} commit_stage={s})", .{
+                    self.replica,
+                    self.op,
+                    self.commit_min,
+                    self.commit_max,
+                    @tagName(self.commit_stage),
+                });
+                self.send_header_to_replica(
+                    self.primary_index(self.view),
+                    @bitCast(Header.RequestStartView{
```
matklad, line 3163: Unlike the old sync flow, we now don't preserve the state which tells us "we were stuck when the timeout fired" anymore. Rather, the state is preseved implictly by the network, when we send request start view, and then, when SV arrives to us, we just go to state sync.

In other words, it is possible that the priary decides to send and SV to us out of the blue, and, if we receive it and _at that point_ `repair_stuck` returns true, we'll go into the state sync, even though we weren't a lot of time in the repair_stuck state.

But I don't think we are sending SVs at random, and, if we receive one, it is _most_ likely due to us sending a request beforehand

src/vsr/replica.zig
```diff
@@ -7840,6 +7824,15 @@ pub fn ReplicaType(
             assert(self.log_view_durable() <= self.view_durable());
             assert(self.log_view_durable() <= self.log_view);
 
+            switch (self.syncing) {
+                .updating_superblock => |stage| {
+                    if (stage.checkpoint_state.header.op == self.op_checkpoint()) {
```
matklad, line 7995: There might be several updates queued, so we check that the correct one finished

src/vsr/replica_test.zig
```diff
@@ -966,39 +969,6 @@ test "Cluster: sync: partition, lag, sync (transition from idle)" {
...
-    t.run();
-
-    try expectEqual(t.replica(.R_).status(), .normal);
-    try expectEqual(t.replica(.R_).commit(), checkpoint_3_trigger);
-    try expectEqual(t.replica(.R_).sync_status(), .idle);
-
-    t.run(); // (Wait for grid sync to finish.)
-    try TestReplicas.expect_sync_done(t.replica(.R_));
-}
-
```
matklad, line 1000: We no longer have this bump sync target case at all, updates are _mostly_ atomic.

src/vsr/replica_test.zig
```diff
@@ -1074,55 +1044,47 @@ test "Cluster: sync: R=4, 2/4 ahead + idle, 2/4 lagging, sync" {
     try TestReplicas.expect_sync_done(t.replica(.R_));
 }
 
-// TODO: Replicas in recovering_head cannot (currently) participate in view-change, even when
-// they arrived at recovering_head via state sync, not corruption+crash. As a result, it is possible
-// for a 2/3 cluster to get stuck without any corruptions or crashes.
-// See: https://github.com/tigerbeetle/tigerbeetle/pull/933#discussion_r1245440623,
-// https://github.com/tigerbeetle/tigerbeetle/issues/1376, and `Simulator.core_missing_quorum()`.
-test "Cluster: sync: view-change with lagging replica in recovering_head" {
```
matklad, line 1081: This is the test that motivated this whole ordeal!

src/vsr/replica_test.zig
```diff
@@ -2038,7 +1993,7 @@ const TestReplicas = struct {
 
         for (t.replicas.const_slice()) |replica_index| {
             const replica: *const Cluster.Replica = &t.cluster.replicas[replica_index];
-            assert(replica.sync_content_done());
+            if (!replica.sync_content_done()) return error.SyncContentPending;
```
matklad, line 2091: just to get nicer error message when a test fails

matklad:
The level of VOPR scrutiny this survives: 

image

It _will_ fail later with a mostly unrelated bug with shallowly-durable checkpoints, but perhaps the sync+view change parts actually work :sweat_smile:

src/vsr/replica.zig
```diff
@@ -2220,54 +2212,131 @@ pub fn ReplicaType(
...
+            //  Cluster is far ahead, replica's WAL no longer intersects primary's WAL.
+                message.header.commit > self.op_prepare_max() + constants.vsr_checkpoint_interval or
+                // Cluster is a bit ahead. The replica might still be able to repair WAL, but that
+                // is not certain.
+                (self.syncing == .idle and self.repair_stuck()) or
+                // Completing previously starting state sync.
+                self.syncing == .awaiting_checkpoint))
+            {
+                // State sync: at this point, the we know we want to replace our checkpoint
+                // with the one from SV.
```
```suggestion
                // State sync: at this point, we know we want to replace our checkpoint
                // with the one from this SV.
```

src/vsr/replica.zig
```diff
@@ -2220,54 +2212,131 @@ pub fn ReplicaType(
...
-                        "(op_prepare_max={} quorum_head={})", .{
-                        self.replica,
-                        self.op_prepare_max(),
-                        view_headers.slice[0].op,
-                    });
+                    // If we are already checkpointing, let that finish first --- perhaps we won't
+                    // need state sync after all.
+                    if (self.commit_stage == .checkpoint_superblock) return;
+                    if (self.commit_stage == .checkpoint_data) return;
+                    // Otherwise, cancel in progress commit and prepare to sync
```
```suggestion
                    // Otherwise, cancel in progress commit and prepare to sync.
```

src/vsr/replica.zig
```diff
@@ -2220,54 +2212,131 @@ pub fn ReplicaType(
...
                 }
-                return;
+                switch (self.syncing) {
+                    .idle => unreachable,
+                    .canceling_commit,
+                    .canceling_grid,
+                    .updating_superblock,
+                    => {
+                        log.debug(
+                            \\{}: on_start_view sync {s} view={} op_checkpoint={} op_checkpoint_new={}
```
```suggestion
                            \\{}: on_start_view: sync {s} view={} op_checkpoint={} op_checkpoint_new={}
```

Nit, but we use this format pretty much everywhere. (I use it for syntax highlighting vopr logs!)

src/vsr/replica.zig
```diff
@@ -2220,54 +2212,131 @@ pub fn ReplicaType(
...
+                            self.op_checkpoint(),
+                            view_checkpoint.header.op,
+                        });
+                        return;
+                    },
+                    .awaiting_checkpoint => {},
+                }
+
+                log.debug(
+                    \\{}: on_start_view sync started view={} op_checkpoint={} op_checkpoint_new={}
```
```suggestion
                    \\{}: on_start_view: sync started view={} op_checkpoint={} op_checkpoint_new={}
```

src/vsr/replica.zig
```diff
@@ -2220,54 +2212,131 @@ pub fn ReplicaType(
...
+                    view_checkpoint.header.op,
+                });
+
+                self.sync_dispatch(.{ .updating_superblock = .{
+                    .checkpoint_state = view_checkpoint.*,
+                } });
+
+                // The new checkpoint would we written to the superblock asynchronously.
+                // From this point on, we are in a delicate state where we must be using this
+                // in-memory checkpoint to check validity of log messages.
```
```suggestion
                // The new checkpoint will we written to the superblock asynchronously.
                // From this point on, we are in a delicate state where we must be using this
                // in-memory checkpoint to check validity of log messages.
```

src/vsr/replica.zig
```diff
@@ -2220,54 +2212,131 @@ pub fn ReplicaType(
...
+                assert(!self.state_machine_opened);
             }
 
-            for (view_headers.slice) |*header| {
-                if (header.op <= self.op_prepare_max()) {
-                    self.replace_header(header);
+            {
+                // Replace our log with the suffix from SV. Transition to sync above guarantees
+                // that there's at least one message that fits the effective checkpoint, but some
+                // messages might be beyond its prepare_max.
```
sentientwaffle, line 2309: "Replace our log with the suffix from SV." is a pun, sort of: 

- You can interpret the "suffix" as the suffix of the primary's log, i.e. the SV headers, or
- you can interpret the "suffix" as the suffix of the SV headers themselves, after we discard whichever leading headers are greater than `op_prepare_max_sync()`. :smile:

src/vsr/replica.zig
```diff
@@ -7813,6 +7765,38 @@ pub fn ReplicaType(
...
+                        .updating_superblock => |_| sync_op_min: {
+                            const syncing_already =
+                                self.superblock.staging.vsr_state.sync_op_max > 0;
+                            const sync_min_old = self.superblock.staging.vsr_state.sync_op_min;
+
+                            const sync_min_new = if (vsr.Checkpoint.trigger_for_checkpoint(
+                                self.op_checkpoint(),
+                            )) |trigger|
+                                // +1 because sync_op_min is inclusive, but (when !syncing_already)
+                                // `vsr_state.commit_min` itself does not need to be synced.
```
```suggestion
                                // `vsr_state.checkpoint.commit_min` itself does not need to be synced.
```

src/vsr/replica.zig
```diff
@@ -7840,6 +7824,15 @@ pub fn ReplicaType(
...
+                .updating_superblock => |stage| {
+                    if (stage.checkpoint_state.header.op == self.op_checkpoint()) {
+                        self.sync_superblock_update_finish();
+                    }
+                },
+                else => {},
+            }
+
             // recovering_head here is triggered by state sync.
             if (self.status == .recovering_head) return;
```
sentientwaffle: Can this be removed?

src/testing/cluster.zig
```diff
@@ -729,10 +729,6 @@ pub fn ClusterType(comptime StateMachineType: anytype) type {
                     };
                 },
                 .sync_stage_changed => switch (replica.syncing) {
-                    .requesting_checkpoint => cluster.log_replica(
-                        .sync_commenced,
```
sentientwaffle, line 800: You can remove `sync_commenced` from `log_replica`'s enum. And update `docs/about/internals/testing.md` accordingly.

And if it isn't in 2 parts, maybe we should change the symbol.

src/vsr/replica_test.zig
```diff
@@ -649,10 +650,12 @@ test "Cluster: repair: ack committed prepare" {
...
     t.replica(.R_).pass(.R_, .bidirectional, .start_view_change);
+    t.replica(.R_).pass(.R_, .bidirectional, .start_view);
     t.replica(.R_).pass(.R_, .bidirectional, .do_view_change);
     p.drop(.__, .bidirectional, .prepare);
     p.drop(.__, .bidirectional, .do_view_change);
     p.drop(.__, .bidirectional, .start_view_change);
+    p.drop(.__, .outgoing, .start_view);
     t.run();
     try expectEqual(b1.commit(), 2);
     try expectEqual(b2.commit(), 2);
```
sentientwaffle: Before this block,

```
    try expectEqual(p.role(), .primary);
```

And after:

```
    try expectEqual(p.role(), .backup);
```

src/vsr/replica_test.zig
```diff
@@ -649,10 +650,12 @@ test "Cluster: repair: ack committed prepare" {
 
     // Change views. B1/B2 participate. Don't allow B2 to repair op=3.
     t.replica(.R_).pass(.R_, .bidirectional, .start_view_change);
+    t.replica(.R_).pass(.R_, .bidirectional, .start_view);
     t.replica(.R_).pass(.R_, .bidirectional, .do_view_change);
     p.drop(.__, .bidirectional, .prepare);
     p.drop(.__, .bidirectional, .do_view_change);
     p.drop(.__, .bidirectional, .start_view_change);
+    p.drop(.__, .outgoing, .start_view);
```
sentientwaffle: :thinking: Why are these changes necessary?

src/vsr/replica_test.zig
```diff
@@ -1074,55 +1044,47 @@ test "Cluster: sync: R=4, 2/4 ahead + idle, 2/4 lagging, sync" {
...
+    // Let the cluster run for some time without B2 state syncing.
+    b2.drop(.R_, .bidirectional, .start_view);
     t.run();
+    try expectEqual(b2.op_checkpoint(), 0);
 
-    // try expectEqual(b1.role(), .primary);
-    try expectEqual(b1.status(), .normal);
-    try expectEqual(b2.status(), .recovering_head);
-    // try expectEqual(t.replica(.R_).status(), .normal);
+    // Let B2 state sync. This unblocks the cluster.
```
sentientwaffle, line 1076: Prior to this, we could add

```
try c.request(checkpoint_2_trigger + 1, checkpoint_2_trigger);
```

to demonstrate that the cluster is blocked.

src/vsr/replica_test.zig
```diff
@@ -1074,55 +1044,47 @@ test "Cluster: sync: R=4, 2/4 ahead + idle, 2/4 lagging, sync" {
...
-    //   view).
-    // For this test, we want the former to occur before the latter (since the latter would always
-    // work).
-    b2.drop(.R_, .bidirectional, .start_view_change);
-    t.run();
-    b2.pass(.R_, .bidirectional, .start_view_change);
+    // Let the cluster run for some time without B2 state syncing.
+    b2.drop(.R_, .bidirectional, .start_view);
     t.run();
+    try expectEqual(b2.op_checkpoint(), 0);
```
sentientwaffle, line 1073: We can also test `b2.status() == .view_change`

src/vsr/replica_test.zig
```diff
@@ -1207,18 +1169,11 @@ test "Cluster: sync: checkpoint from a newer view" {
...
         b1.drop(.R_, .incoming, .pong);
 
-        const b1_view_before = b1.view();
         try c.request(checkpoint_2_trigger - 1, checkpoint_2_trigger - 1);
-        try expectEqual(b1_view_before, b1.view());
-        try expectEqual(b1.op_checkpoint(), checkpoint_1);
-        try expectEqual(b1.status(), .recovering_head);
 
         b1.stop();
         try b1.open();
```
sentientwaffle, line 1176: Without the asserts before/after b1 is restarted, its not clear to me why we restart it -- maybe we can remove the restart?

src/vsr/replica.zig
sentientwaffle: Likewise `superblock.view_change`. Maybe combine `superblock.view_change()` and `superblock.checkpoint()` into `superblock.update(union(enum) { checkpoint, view_change, sync })`?

src/vsr/superblock.zig
```diff
@@ -913,14 +916,16 @@ pub fn SuperBlockType(comptime Storage: type) type {
             assert(superblock.staging.vsr_state.commit_max <= update.commit_max);
```
sentientwaffle, line 921: This function's comment is out of date.

src/vsr/sync.zig
```diff
@@ -18,56 +18,31 @@ pub const Stage = union(enum) {
     /// Waiting for `Grid.cancel()`.
     canceling_grid,
 
-    /// We need to sync, but are waiting for a usable `sync_target_max`.
-    requesting_target,
-
-    requesting_checkpoint: RequestingCheckpoint,
+    /// We received an SV, decided to sync, but were committing at that point. So instead we
+    /// requested cancelletion of the commit process and entered `awaiting_checkpoint` to get
```
sentientwaffle:
```suggestion
    /// requested cancellation of the commit process and entered `awaiting_checkpoint` to get
```

src/vsr/replica.zig
```diff
@@ -8932,10 +8819,9 @@ pub fn ReplicaType(
             } });
         }
 
-        fn sync_superblock_update(self: *Self) void {
+        fn sync_superblock_update_start(self: *Self) void {
```
sentientwaffle, line 8987: Other preconditions:
- `commit_stage == .idle` (likewise for `sync_superblock_update_finish`)
- Sanity-check `self.syncing.updating_superblock`'s checkpoint op against our current checkpoint op.
- `maybe(self.view_durable_updating());`

src/vsr/replica.zig
```diff
@@ -1602,6 +1588,11 @@ pub fn ReplicaType(
...
             assert(message.header.operation != .reserved);
 
+            if (self.syncing == .updating_superblock) {
+                log.debug("{}: on_prepare: ignoring (sync)", .{self.replica});
+                return;
+            }
+
             if (self.is_repair(message)) {
                 log.debug("{}: on_prepare: ignoring (repair)", .{self.replica});
                 self.on_repair(message);
```
sentientwaffle: In `on_repair`, add `assert(self.syncing != .updating_superblock);`.
(The ordering of the checks in `on_prepare` is important and subtle -- we check `self.syncing == .updating_superblock` before checking `is_repair`.)

src/constants.zig
matklad: Thinking more about this, stuffing fixed-sized checkpoint after the variable number of headers is just weird, so I swapped the order.

src/vsr/replica.zig
```diff
@@ -10365,13 +10125,40 @@ fn message_body_as_prepare_headers(message: *const Message) []const Header.Prepa
...
+    const headers = std.mem.bytesAsSlice(
+        Header.Prepare,
+        message.body()[@sizeOf(vsr.CheckpointState)..],
+    );
+    // To run verification.
+    _ = vsr.Headers.ViewChangeSlice.init(.start_view, headers);
+    if (constants.verify) {
+        for (headers) |header| assert(header.valid_checksum());
+    }
+    return headers;
```
matklad, line 10366: As a follow up, looks like our header parsign fucntions are no longer polymorphic, so we could perhaps could have `Message.StartView.checkpoint`, `Message.Headers.headers`, etc.

src/constants.zig
sentientwaffle: :+1: I was going to suggest the same thing! For that reason, and also b/c that way it matches the order in the superblock header as well (where `CheckpointState` also precedes the view headers).

src/vsr/replica.zig
```diff
@@ -9011,15 +8847,15 @@ pub fn ReplicaType(
 
             const stage: *const SyncStage.UpdatingSuperBlock = &self.syncing.updating_superblock;
 
-            assert(self.superblock.working.checkpoint_id() == stage.target.checkpoint_id);
+            assert(self.superblock.working.vsr_state.checkpoint.header.checksum ==
+                stage.checkpoint_state.header.checksum);
```
sentientwaffle, line 9027: Additionally, we could assert

```
            assert(self.superblock.staging.vsr_state.checkpoint.header.checksum ==
                stage.checkpoint_state.header.checksum);
```

to verify that even if there is another `view_durable_update` running, it isn't a state-sync update.

src/vsr/replica.zig
```diff
@@ -2220,54 +2212,126 @@ pub fn ReplicaType(
...
 
-            for (view_headers.slice) |*header| {
-                assert(header.commit <= message.header.commit);
+            const view_headers = start_view_message_headers(message);
+            assert(view_headers[0].op == message.header.op);
+            assert(view_headers[0].op >= view_headers[view_headers.len - 1].op);
+
+            if (message.header.commit > self.op_prepare_max() and (
+            //  Cluster is far ahead, replica's WAL no longer intersects primary's WAL.
+                message.header.commit > self.op_prepare_max() + constants.vsr_checkpoint_interval or
```
sentientwaffle, line 2243: :thinking: Shouldn't this be
```suggestion
                message.header.commit > self.op_prepare_max() + constants.journal_slot_count or
```
instead?

src/vsr/replica.zig
matklad: (got the plan here now, would add some more commits to implement it)

src/vsr/replica.zig
matklad: Maybe even something like

```
fn update(view: ?ViewUpdate, checkpont: ?CheckpointUpdate) {
    assert(!(view == null and checkpoint == null));
}

```

src/vsr/replica.zig
matklad: That's not a nit at all, that's really important to make our log messages consistent, thanks!

src/vsr/replica.zig
matklad: Good catch! At some point I should go over the original sync PR and see if there are more recovoinrg_head logic which can be removed, or asserts tightened.

src/testing/cluster.zig
matklad: Removed the enum, but I'll leave further adjustments for later. It feels like maybe the entire `sync_dispatch` machinery no longer pulls its weight.

src/vsr/replica_test.zig
matklad: Good catch, they aren't, backed them out! 

As why I have I added them, I sadly don't remeber. The only thing I recall about fixing this test is how the comment speaks about `A0` but we use `p` as a variable named

src/vsr/replica_test.zig
matklad: Added the following comment:

```
// Wipe B1 in-memory state and check that it ends up in a consistent state after restart. 
```

What we are testing here is that b1 doesn't trip `assert(valid_hash_chain)` assertion, so there isn't anything specific to test for here.

src/vsr/replica.zig
matklad: Oh, this is excellent!

src/vsr/replica.zig
matklad: Good question. I think either version is correct, and, moreover, we could completely remove this condition (we'll just hit `repair_stuck` a bit later in the future).

I'll leave this as is, though, because it seems like a nice heuristic for when to switch to state-sync proactively, and because this allows to comment on the subtelty here (which the original comment missed):

The overall condition checkifs that SV's commit is ahead of our current checkpont's prepare_max. 

**If** SV's checkpoint is our `op_checkpoint_next`, then `message.header.commit > self.op_prepare_max()` guarantees that that checkpoint is durable and is safe to sync to. 

**However**, if SV's checkpoint is further in the future, it might be the case that it is not durable. Eg, we are a severely lagging replica, primary just triggered a checkpoint and tells this fresh checkpoint to us, but the rest of the cluster is still slightly behind the cluster. But in this case it is still save to sync to that checkpoint, because our own prepares are no longer relevant. 

---

I guess if we want to tighten the heuristic where to skip the state sync, we should check which prepare we are actually missing, and sync when that gets evicted from primary's log.

sentientwaffle (dismissed):
LGTM! (Though I assume this shouldn't be merged as-is, depending on what the transition/upgrade strategy will look like.)

matklad:
>Though I assume this shouldn't be merged as-is, depending on what the transition/upgrade strategy will look like.

Indeed! I've now added the upgrade strategy as the last commit, which I _think_ should work.

Not ready to land that right away ether though:

* I want to test in in vopr (in fact, I already have a failing seed on hand)
* The VOPR tests (in https://github.com/tigerbeetle/tigerbeetle/pull/2176) are somewhat hacky, as they make `vsr.Version` overridable at runtime. I am not fully sure if I want to land that approach (it adds a bunch of vopr-only code), but, right now, I am more curious whether the actual upgrade path will work.

matklad:
Ok, I realized that I got myself off-by-oned here! What https://github.com/tigerbeetle/tigerbeetle/pull/2176) aims to test is the the release after the next one, when we switch to only having version 1.

I _think_ to the extent that the switch logic could be tested for the next release, it _is_ tested in this PR as is: because the replicas send both SV0 and SV1, VOPR dropping `SV1` message simulates a cluster with some old replicas. 

What is _not_ tested by the VOPR is that the old memebers can state sync (the latest commit adds back the code to respond to `fn request_sync_target`, but no one actually requests it yet). To test _that_ in VOPR, I'll need to somehow bring back the old state sync logic, and that seems not really feasible to do without adding a bunch of effectively vopr-only code

src/vsr.zig
```diff
@@ -219,6 +219,11 @@ pub const Command = enum(u8) {
     request_sync_checkpoint = 21,
     sync_checkpoint = 22,
 
+    // A reserved command for the new state sync protocol. At the moment, replica ignores this
+    // command (as opposed to panicking on an unknown command), to allow the next release to send
+    // both versions of StartView.
```
sentientwaffle: The comment is out of date.

src/vsr/replica.zig
```diff
@@ -4598,7 +4631,10 @@ pub fn ReplicaType(
 
         /// Construct a SV message, including attached headers from the current log_view.
         /// The caller owns the returned message, if any, which has exactly 1 reference.
-        fn create_start_view_message(self: *Self, nonce: u128) *Message.StartView {
+        fn create_start_view_message(self: *Self, nonce: u128) struct {
+            current: *Message.StartView,
```
```suggestion
            current: *Message.StartViewDeprecated,
```

(I know they are technically identical.)

src/vsr/replica.zig
```diff
@@ -4645,7 +4683,27 @@ pub fn ReplicaType(
             message.header.set_checksum_body(message.body());
             message.header.set_checksum();
 
-            return message.ref();
+            const message_copy = self.message_bus.get_message(.start_view);
```
sentientwaffle, line 4693: If we are using `.get_message(.start_view)`, could we comptime-assert `Header.StartView == Header.StartViewDeprecated` here?

sentientwaffle:
Other than those nits :arrow_up: this LGTM.

matklad:
Going to leave this waiting until we make a release with https://github.com/tigerbeetle/tigerbeetle/pull/2181. After that, fix nits, rebase on main, and should be good to go!

sentientwaffle (approved):
LGTM to merge after the next release!

PR #1932, opened 2024-05-10
Reproduce go memory corruption

matklad:
```
$ CC="/home/matklad/p/tb/work/zig/zig cc" go test

0xc000149e40 free  unmarked
0xc000149e80 free  unmarked
0xc000149ec0 free  unmarked
0xc000149f00 free  unmarked
0xc000149f40 free  unmarked
0xc000149f80 free  unmarked
0xc000149fc0 free  unmarked
fatal error: found pointer to free object

runtime stack:
runtime.throw({0x24ade8?, 0xc0001480c0?})
	/nix/store/9kriq85qac7phcxgpdqbqr25vlr61ifw-go-1.22.2/share/go/src/runtime/panic.go:1023 +0x5c fp=0x7faaa9ffb9d0 sp=0x7faaa9ffb9a0 pc=0x398e9c
runtime.(*mspan).reportZombies(0x7faab33965e8)
	/nix/store/9kriq85qac7phcxgpdqbqr25vlr61ifw-go-1.22.2/share/go/src/runtime/mgcsweep.go:872 +0x2ea fp=0x7faaa9ffba50 sp=0x7faaa9ffb9d0 pc=0x38794a
runtime.(*sweepLocked).sweep(0x0?, 0x0)
	/nix/store/9kriq85qac7phcxgpdqbqr25vlr61ifw-go-1.22.2/share/go/src/runtime/mgcsweep.go:638 +0x3c6 fp=0x7faaa9ffbb68 sp=0x7faaa9ffba50 pc=0x386946
runtime.(*mcentral).uncacheSpan(0x0?, 0x0?)
	/nix/store/9kriq85qac7phcxgpdqbqr25vlr61ifw-go-1.22.2/share/go/src/runtime/mcentral.go:236 +0x98 fp=0x7faaa9ffbb90 sp=0x7faaa9ffbb68 pc=0x377178
runtime.(*mcache).releaseAll(0x7faafb629108)
	/nix/store/9kriq85qac7phcxgpdqbqr25vlr61ifw-go-1.22.2/share/go/src/runtime/mcache.go:291 +0x13e fp=0x7faaa9ffbbf8 sp=0x7faaa9ffbb90 pc=0x376abe
runtime.(*mcache).prepareForSweep(0x7faafb629108)
[...]

PR #1952, opened 2024-05-14
go: pin pointers crossing FFI boundary

matklad:
References:

* https://pkg.go.dev/runtime#Pinner
* https://pkg.go.dev/cmd/cgo#hdr-Passing_pointers

Command I used for debugging:

GOEXPERIMENT=cgocheck2 CC="/home/matklad/p/tb/work/zig/zig cc" go test -gcflags=-d=checkptr

Bug repro:

https://github.com/tigerbeetle/tigerbeetle/pull/1932

### DIFF
--- a/src/clients/go/tb_client.go
+++ b/src/clients/go/tb_client.go
@@ -29,6 +29,7 @@ extern __declspec(dllexport) void onGoPacketCompletion(
 import "C"
 import (
 	e "errors"
+	"runtime"
 	"strings"
 	"unsafe"
 
@@ -189,6 +190,11 @@ func (c *c_client) doRequest(
 	// Release the packet for other goroutines to use.
 	defer C.tb_client_release_packet(c.tb_client, req.packet)
 
+	pinner := new(runtime.Pinner)
+	pinner.Pin(&req)
+	pinner.Pin(data)
+	pinner.Pin(result)
+
 	req.packet.user_data = unsafe.Pointer(&req)
 	req.packet.operation = C.uint8_t(op)
 	req.packet.status = C.TB_PACKET_OK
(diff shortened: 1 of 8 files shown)


src/clients/go/tb_client.go
```diff
@@ -203,6 +209,7 @@ func (c *c_client) doRequest(
 
 	// Wait for the request to complete.
 	<-req.ready
+	pinner.Unpin()
```
batiati, line 212: I have faced this question before, and my reasoning was that we didn't need to pin memory because the reference is held during the duration of the gorountine, that is, it is not a callback in the sense of two independent calls, but just a “wait for the channel signal”.

So, why this assumption is not true and the GC is removing the reference? Is there some kind of compiler state machine under the hood, like in async functions?

matklad: From the language-lawyer perspective, the cgo docs are pretty clear that:

* all things must be pinned
* arguments are pinned implicitly only during a single FFI call

while this  ref is scoped to a go function call, it isn't scoped to a single FFI call, so we are in violation here. And running `cgocheck2` tells pretty much this:

```
write of unpinned Go pointer 0xc000012ab0 to non-Go memory 0x7f7a69b1bfd8
fatal error: unpinned Go pointer stored into non-Go memory
```

as to why this actually blows up at runtime (https://github.com/tigerbeetle/tigerbeetle/pull/1932) I am actually not sure!

One explanation I have:

* the `req` object is allocated on the stack of the goroutine
* when we do `<-`, the goroutine is suspended
* when the goroutine is suspending, the GC moves its stack (I think _stacks_ in Go are movable)
* the pointer now points to "fun"

batiati (changes requested):
LGTM, can you also bump the Go version and regenerate the docs?

https://github.com/tigerbeetle/tigerbeetle/blob/eaa4f595970fdd27bef5389737df9d6b70904bee/src/clients/go/docs.zig#L25-L31

matklad:
Good point!

batiati (approved):

PR #1955, opened 2024-05-15
vopr: tics_max are runtime-overridable

matklad:
Wanted to do this since forever, as bisecting on the ticks_max is the first step to debug a liveness issue. Now that we use flags in the simulator, we can just go and do it.

This is a small step -- a more principled approach would be perhaps to stuff the entire config in the CLI, so that everything is overidable. However, I rarely need to override anything but the ticks, so lets start small!

### DIFF
--- a/src/simulator.zig
+++ b/src/simulator.zig
@@ -70,6 +70,8 @@ const cluster_id = 0;
 const CliArgs = struct {
     // "lite" mode runs a small cluster and only looks for crashes.
     lite: bool = false,
+    ticks_max_requests: u32 = 40_000_000,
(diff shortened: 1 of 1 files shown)


src/simulator.zig
```diff
@@ -312,10 +313,8 @@ pub fn main() !void {
     simulator.transition_to_liveness_mode();
 
     // Liveness: a core set of replicas is up and fully connected. The rest of replicas might be
-    // crashed or partitioned permanently. The core should converge to the same state.
```
sentientwaffle: The second line of the comment got trimmed by accident.

sentientwaffle (approved):

PR #1957, opened 2024-05-15
auditor: fix non-determinism 

matklad:
Expiration timestamp might match between two transfers, so the order of
items in the priority queue is non-deterministic. The order is
important, because we expire a batchfull of transfers at a time at most.

Order the transfers by their (unique) timestamp, just like the state
machine!

This the second priority queue bug in a row!

### DIFF
--- a/src/state_machine/auditor.zig
+++ b/src/state_machine/auditor.zig
@@ -44,14 +44,14 @@ const PendingTransfer = struct {
 };
 
 const PendingExpiry = struct {
-    transfer: u128,
-    timestamp: u64,
+    transfer_id: u128,
+    expires_at: u64,
 };
 
 const PendingExpiryQueue = PriorityQueue(PendingExpiry, void, struct {
     /// Order by ascending timestamp.
     fn compare(_: void, a: PendingExpiry, b: PendingExpiry) std.math.Order {
-        return std.math.order(a.timestamp, b.timestamp);
+        return std.math.order(a.expires_at, b.expires_at);
     }
 }.compare);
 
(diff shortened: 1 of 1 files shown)


matklad:
(seed in https://github.com/tigerbeetle/tigerbeetle/pull/1938#issuecomment-2112560940)

src/state_machine/auditor.zig
```diff
@@ -44,14 +44,20 @@ const PendingTransfer = struct {
...
 
 const PendingExpiryQueue = PriorityQueue(PendingExpiry, void, struct {
-    /// Order by ascending timestamp.
+    /// Order by ascending expiration date and then by transfer's timestamp.
     fn compare(_: void, a: PendingExpiry, b: PendingExpiry) std.math.Order {
-        return std.math.order(a.timestamp, b.timestamp);
+        const order = switch (std.math.order(a.expires_at, b.expires_at)) {
+            .eq => std.math.order(a.transfer_timestamp, b.transfer_timestamp),
+            else => |order| order,
+        };
```
matklad, line 58: Is there some Zig idiom that does essentially https://doc.rust-lang.org/std/cmp/enum.Ordering.html#method.then_with?

https://github.com/tigerbeetle/tigerbeetle/blob/main/src/scripts/cfo.zig#L555-L563 looks to ad-hoc to my taste to promote that to stdx...

sentientwaffle: Not that I know of!

sentientwaffle (approved):
LGTM! Fwiw when I wrote this originally, I was imagining that we would ensure that the state machine would ensure that all expiry times are unique timestamps. :facepalm:

matklad:
> machine would ensure that all expiry times are unique timestamps. 

Yeah, I don’t think that’s possible

PR #1959, opened 2024-05-16
io: remove dead benchmark

matklad:
This benchmark doesn't compile, and it looks like it wansn't even ported to Zig 0.11. As we don't actively need it, remove it


src/tidy.zig
```diff
@@ -147,7 +147,6 @@ const DeadDetector = struct {
 
     fn is_entry_point(file: FileName) bool {
         const entry_points: []const []const u8 = &.{
-            "benchmark.zig",
```
matklad, line 150: Funny how this tidy check was explicitly designed to catch dead files like this, but this particular one dodged the check, because I was lazy and implemnted the check based solely on the file name, and not on the entire path (and we used to have benchmark entry points)

kprotty (approved):
I believe it was originally written to ensure the IO backends were maximizing network throughput during development. Ok to be removed now.

PR #1962, opened 2024-05-16
docs: add Concepts section

emschwartz:
- **docs: add Concepts page about OLTP**
- **docs: only use OLTP in the sidebar**
- **docs: remove link to next page (until it exists)**
- **docs: add Concepts page about Debit / Credit**


matklad:
Took us _a while_, but we've finally landed the concepts section, thanks for this intial work!

PR #1964, opened 2024-05-18
fix: broken link in `account-balance.md`

lewisdaly:
I noticed this link was broken when I was looking through the nodejs docs


emschwartz (approved):

emschwartz:
Thanks for noticing this and submitting the fix!

PR #1966, opened 2024-05-20
release: 2024-05-20

sentientwaffle:

PR #1967, opened 2024-05-20
Scripts/Changelog: Allow nonexistent ./zig-cache

sentientwaffle:
Don't fail to build the changelog if the `./zig-cache` directory doesn't exist.

I ran into this as I was building the [changelog](https://github.com/tigerbeetle/tigerbeetle/pull/1966) from within a brand-new worktree, so `zig-cache` didn't exist yet.

### DIFF
--- a/src/scripts/changelog.zig
+++ b/src/scripts/changelog.zig
@@ -17,6 +17,7 @@ pub fn main(shell: *Shell, gpa: std.mem.Allocator) !void {
     const merges = try shell.exec_stdout(
         \\git log --merges --first-parent origin/release..origin/main
     , .{});
+    try shell.project_root.makePath("./zig-cache");
     try shell.project_root.writeFile("./zig-cache/merges.txt", merges);
     log.info("merged PRs: ./zig-cache/merges.txt", .{});
 


matklad:
There should be some helper in shell.zig which writes the file creating the required directories automatically. I think it’s ensure file contents?

sentientwaffle:
`shell.file_ensure_content` doesn't create directories, it just doesn't fail hard if the directory doesn't exist.

matklad:
Ok, then it’s “should be” in a sense that we need to write one!

that’s another specific thing I had in mind with the whole shell infra: for scripting, function to write a file should just go and do it!

🤔 i am surprised that we didn’t hit this before, will take a look tomorrow, I am still 50% sure we’ve solved this once somewhere already…

src/scripts/changelog.zig
```diff
@@ -17,8 +17,13 @@ pub fn main(shell: *Shell, gpa: std.mem.Allocator) !void {
...
     , .{});
-    try shell.project_root.writeFile("./zig-cache/merges.txt", merges);
-    log.info("merged PRs: ./zig-cache/merges.txt", .{});
+
+    if (try shell.dir_exists("./zig-cache")) {
+        try shell.project_root.writeFile("./zig-cache/merges.txt", merges);
+        log.info("merged PRs: ./zig-cache/merges.txt", .{});
+    } else {
+        log.warn("merged PRs: ./zig-cache does not exist; not writing merges.txt", .{});
+    }
```
matklad: Took a closer look at what we have and recalled why I am conflicted about having `shell.justWriteTheFilePlease` helper! The "capability oriented" `std` API, `dir.writeFile(text)` is just nicely looking syntactically, with `dir` as a method receiver. 

So I think we shouldn't wrap it just yet, so let's change the code to

```suggestion
    try shell.project_root.makePath("./zig-cache");
    try shell.project_root.writeFile("./zig-cache/merges.txt", merges);
    log.info("merged PRs: ./zig-cache/merges.txt", .{});
```

I think the semantics of just creating a directory is the right one, and I think it's a fair game that we re-use `zig-cache` directory here. Semantically, it is conventient to use project-local temporary directory, and, while we can create our own, I think just re-using existing one is simpler.

matklad (approved):

PR #1970, opened 2024-05-22
benchmark: the devhub file size line was not flat

cb22:
It is now flat.

...

Previously, the benchmark would calculate a target TPS and busy-wait to try and achieve it.

The problem with that approach is it leads to unpredictable and random batch sizing. Instead, remove the `transfer_count_per_second‎` flag entirely and replace it with two new options:

* `transfer_batch_size` - sets the size of a transfer batch,
* `transfer_batch_delay_us` - sets the inter-batch delay.

These two flags allow obtaining the same behaviour, in a more predictable and application style way, rather than trying to target an arbitrary TPS. They default to 'as fast as possible'.

The random batch sizing lead to the datafile size varying with each run.

Also, fix that `account_balances` CLI flag wasn't being passed through because we were setting defaults twice!


src/flags.zig
```diff
@@ -108,6 +108,8 @@ fn parse_commands(args: *std.process.ArgIterator, comptime Commands: type) Comma
 }
 
 fn parse_flags(args: *std.process.ArgIterator, comptime Flags: type) Flags {
+    @setEvalBranchQuota(5000);
```
matklad, line 111: attonit, but `5_000`

matklad (approved):

src/flags.zig
cb22: I'll fix that in an attopr to test this is working as expected!

PR #1973, opened 2024-05-23
Fix go memory corruption

jokruger:
This commit addresses "bad pointer" errors by using runtime.Pinner to pin Go objects in memory during C function calls, preventing the GC from releasing them prematurely. It also ensures pointers are reset to nil after the C function completes to avoid GC checks on nonexistent pointers.

### DIFF
--- a/src/clients/go/tb_client.go
+++ b/src/clients/go/tb_client.go
@@ -29,6 +29,7 @@ extern __declspec(dllexport) void onGoPacketCompletion(
 import "C"
 import (
 	e "errors"
+	"runtime"
 	"strings"
 	"unsafe"
 
@@ -175,6 +176,13 @@ func (c *c_client) doRequest(
 		ready:  make(chan struct{}),
 	}
 
+	pinner := &runtime.Pinner{}
+	defer pinner.Unpin()
+
+	pinner.Pin(&req)
+	pinner.Pin(data)
+	pinner.Pin(result)
+
 	switch acquire_status := C.tb_client_acquire_packet(c.tb_client, &req.packet); acquire_status {
 	case C.TB_PACKET_ACQUIRE_CONCURRENCY_MAX_EXCEEDED:
 		return 0, errors.ErrConcurrencyExceeded{}
@@ -185,9 +193,6 @@ func (c *c_client) doRequest(
 			panic("tb_client_acquire_packet(): returned null packet")
 		}
 	}
-
-	// Release the packet for other goroutines to use.
-	defer C.tb_client_release_packet(c.tb_client, req.packet)
 
 	req.packet.user_data = unsafe.Pointer(&req)
 	req.packet.operation = C.uint8_t(op)
@@ -205,6 +210,13 @@ func (c *c_client) doRequest(
 	<-req.ready
 	status := C.TB_PACKET_STATUS(req.packet.status)
 	wrote := int(req.packet.data_size)
+
+	// Release the packet for other goroutines to use.
+	p := req.packet
+	req.packet.user_data = nil
+	req.packet.data = nil
+	req.packet = nil
+	C.tb_client_release_packet(c.tb_client, p)
 
 	// Handle packet error
 	if status != C.TB_PACKET_OK {


arijoon:
is this a duplicate of https://github.com/tigerbeetle/tigerbeetle/pull/1952 ?

jokruger:
> is this a duplicate of #1952 ?

No (but related), #1952 does not solve the problem (I've tried it in my case but crash remained). Pin/unpin guarantee the object will not be GC-ed while C (zig) function is still running. But the main cause of problem is that some of the request and packet stored pointers to data, allocated and deallocated in C (zig), was not erased after deallocation, so GC reports it as "bad pointer" or "pointer to freed memory". So all you need is to set req.packet.user_data, req.packet.data and req.packet to nil and then deallocate what was the req.packet. Pin/unpin probably also better to keep.

batiati:
> https://github.com/tigerbeetle/tigerbeetle/pull/1952 does not solve the problem (I've tried it in my case but crash remained).

Hey @jokruger, can you provide some insight on how to reproduce the failing test after applying #1952?

> the main cause of problem is that some of the request and packet stored pointers to data, allocated and deallocated in C (zig), was not erased after deallocation, so GC reports it as "bad pointer" or "pointer to freed memory". So all you need is to set req.packet.user_data, req.packet.data and req.packet to nil and then deallocate what was the req.packet. Pin/unpin probably also better to keep.


Considering that the Zig side `release_packet()` doesn't free the `req.packet` memory, it just pushes it onto a stack for later, so the GC is unable to track any invalid reference at this point. Even if not, It should be fine to hold invalid C pointers in Go. Otherwise, the GC could error out even before you set them to null.

My guess is that if it's still failing, it's probably a flaky test, and maybe we haven't addressed the real cause yet!

jokruger:
> > #1952 does not solve the problem (I've tried it in my case but crash remained).
> 
> Hey @jokruger, can you provide some insight on how to reproduce the failing test after applying #1952?

You need to do a lot of additional mem related work in Go, i.e. create and destroy a lot of objects (probably this is triggering GC). I tried to isolate it in a small test, but each time I reduce the code to simple one, it starts working ok :(

> Considering that the Zig side `release_packet()` doesn't free the `req.packet` memory, it just pushes it onto a stack for later, so the GC is unable to track any invalid reference at this point.

GC will have info only on pointers allocated in Go - i.e. it is not able to check if pointer is actually valid or not (i.e. was released in Zig or not) - it able only check if it is valid from Go point of view. Imho when GC sees pointer which was not allocated in Go it treat it as "bad pointer".

> Even if not, It should be fine to hold invalid C pointers in Go. Otherwise, the GC could error out even before you set them to null.
> My guess is that if it's still failing, it's probably a flaky test, and maybe we haven't addressed the real cause yet!

Maybe this is the reason?

```
	req := request{
		packet: nil,
		ready:  make(chan struct{}),
	}

	switch acquire_status := C.tb_client_acquire_packet(c.tb_client, &req.packet); acquire_status {
```

From Go's point of view, the req.packet after creation is a nil pointer (so Go knows it is a pointer). Then, C.tb_client_acquire_packet sets it to non-nil value, but it is done by Zig, not by Go, so if we will try to release req (actually to force GC it), it should fail because we trying to GC a structure with a pointer GC has no info about.

When we call C.tb_client_release_packet, it takes ownership of packet pointer but does not clears corresponding variable in Go, so Go still has a structure with unknown pointer. So when we exit function, req starts waiting for GC and when it happens it fails with bad pointer.

Not sure if same is happening to req.packet.user_data and req.packet.data, but it is definitely safe to set all to nil (i.e. to a state it was before we start unsafe pointer manipulations) before GC

batiati:
Well, it makes sense, although it creeps me out with all this GC esoterism :fearful: !
CCing @kprotty to help us with this.

In the meantime, it would be great if you could help us to test your hypothesis by just changing this:

```go
acquire_status := C.tb_client_acquire_packet(c.tb_client, &req.packet); acquire_status {
```

with a temp local pointer before initializing the `request`, like this:

```go
var p *C.tb_packet_t
switch acquire_status := C.tb_client_acquire_packet(c.tb_client, &p); acquire_status {
case C.TB_PACKET_ACQUIRE_CONCURRENCY_MAX_EXCEEDED:
	return 0, errors.ErrConcurrencyExceeded{}
case C.TB_PACKET_ACQUIRE_SHUTDOWN:
	return 0, errors.ErrClientClosed{}
default:
	if p == nil {
		panic("tb_client_acquire_packet(): returned null packet")
	}
}

req := request{
	packet: p,
	ready:  make(chan struct{}),
}
```

jokruger:
> Well, it makes sense, although it creeps me out with all this GC esoterism 😨 ! CCing @kprotty to help us with this.
> 
> In the meantime, it would be great if you could help us to test your hypothesis by just changing this:
> 
> ```go
> acquire_status := C.tb_client_acquire_packet(c.tb_client, &req.packet); acquire_status {
> ```
> 
> with a temp local pointer before initializing the `request`, like this:
> 
> ```go
> var p *C.tb_packet_t
> switch acquire_status := C.tb_client_acquire_packet(c.tb_client, &p); acquire_status {
> case C.TB_PACKET_ACQUIRE_CONCURRENCY_MAX_EXCEEDED:
> 	return 0, errors.ErrConcurrencyExceeded{}
> case C.TB_PACKET_ACQUIRE_SHUTDOWN:
> 	return 0, errors.ErrClientClosed{}
> default:
> 	if p == nil {
> 		panic("tb_client_acquire_packet(): returned null packet")
> 	}
> }
> 
> req := request{
> 	packet: p,
> 	ready:  make(chan struct{}),
> }
> ```

```
...
if c.tb_client == nil {
return 0, errors.ErrClientClosed{}
}

var p *C.tb_packet_t
switch acquire_status := C.tb_client_acquire_packet(c.tb_client, &p); acquire_status {
case C.TB_PACKET_ACQUIRE_CONCURRENCY_MAX_EXCEEDED:
    return 0, errors.ErrConcurrencyExceeded{}
case C.TB_PACKET_ACQUIRE_SHUTDOWN:
    return 0, errors.ErrClientClosed{}
default:
    if p == nil {
        panic("tb_client_acquire_packet(): returned null packet")
    }                                                                
}                                                                    
                                                                     
// Release the packet for other goroutines to use.           
defer C.tb_client_release_packet(c.tb_client, p)             
                                                                     
req := request{                                                      
    packet: p,                                                         
    ready:  make(chan struct{}),
}
...
```
results in
```
runtime: pointer 0xc0000da000 to unallocated span span.base()=0xc0000d8000 span.limit=0xc0000e0000 span.state=0
fatal error: found bad pointer in Go heap (incorrect use of unsafe or cgo?)
```

batiati:
Thanks so much, I feel better already :sweat_smile: 
Let us try to reproduce it in the simplest possible form, so we can figure out what's actually happening.

jokruger:
Another possible reason - data and user_data unsafe pointers. Same mechanics - if we not set them to nil it may lead to situation when req contains pointer to accounts[0] (from LookupAccounts) or transferIDs[0] (from LookupTransfers) while accounts or transferIDs already gone.

I've tried to add
```
req.packet.data = nil
req.packet.user_data = nil
```

right before `if status != C.TB_PACKET_OK {` and it works ok.

batiati:
Closing, as it was fixed by #1380.
Specifically by this code:
https://github.com/tigerbeetle/tigerbeetle/blob/2e008091365e53e7b0c1e4d5774ba367d4715433/src/clients/go/tb_client.go#L184-L186

PR #1974, opened 2024-05-23
benchmark/devhub: track empty datafile size and devhub improvements

cb22:
Track the size of a freshly formatted datafile. Additionally, make the devhub line things across graphs and speed up rendering.

### DIFF
--- a/src/devhub/devhub.js
+++ b/src/devhub/devhub.js
@@ -51,11 +51,15 @@ async function mainMetrics() {
   const dataUrl =
     "https://raw.githubusercontent.com/tigerbeetle/devhubdb/main/devhub/data.json";
   const data = await (await fetch(dataUrl)).text();
+  const maxBatches = 200;
   const batches = data.split("\n")
     .filter((it) => it.length > 0)
-    .map((it) => JSON.parse(it));
+    .map((it) => JSON.parse(it))
+    .slice(-1 * maxBatches)
+    .reverse();
+
   const series = batchesToSeries(batches);
-  plotSeries(series, document.querySelector("#charts"));
+  plotSeries(series, document.querySelector("#charts"), batches.length);
 }
 
 async function mainSeeds() {
@@ -150,11 +154,11 @@ function pullRequestNumber(record) {
 //
 // This doesn't depend on particular plotting library though.
 function batchesToSeries(batches) {
-  const result = new Map();
-  for (const batch of batches) {
+  const results = new Map();
+  for (const [index, batch] of batches.entries()) {
     for (const metric of batch.metrics) {
-      if (!result.has(metric.name)) {
-        result.set(metric.name, {
+      if (!results.has(metric.name)) {
+        results.set(metric.name, {
           name: metric.name,
           unit: undefined,
           value: [],
@@ -163,7 +167,7 @@ function batchesToSeries(batches) {
         });
       }
 
-      const series = result.get(metric.name);
+      const series = results.get(metric.name);
       assert(series.name == metric.name);
 
       if (series.unit) {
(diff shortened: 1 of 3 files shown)


src/devhub/devhub.js
```diff
@@ -51,11 +51,14 @@ async function mainMetrics() {
   const dataUrl =
     "https://raw.githubusercontent.com/tigerbeetle/devhubdb/main/devhub/data.json";
   const data = await (await fetch(dataUrl)).text();
+  const maxBatches = 200;
   const batches = data.split("\n")
     .filter((it) => it.length > 0)
-    .map((it) => JSON.parse(it));
+    .map((it) => JSON.parse(it))
+    .slice(-1 * maxBatches);
```
matklad: Nice! This’ll also naturally evict old timelines we no longer update!

the long term plan is also to log-rotate the actual JSON file manually once a year or so

src/devhub/devhub.js
```diff
@@ -172,24 +175,45 @@ function batchesToSeries(batches) {
...
+      // group according to time. Apex charts is much quicker when given an x value, even though it
+      // isn't strictly needed.
+      series.value.push([index, metric.value]);
       series.git_commit.push(batch.attributes.git_commit);
       series.timestamp.push(batch.timestamp);
     }
   }
-  return result.values();
+
+  // Sort the results by batches that have the latest entries first, so metrics that have stopped
```
matklad: Let’s do a little bit more here: sort series’s by their order of appearance in data.

that way, if we change the order of things in zig code for the latest batch, it’ll affect the ordering of the graphs.

this’ll naturally sort shorter series to the bottom.

impl wise, I’d imagine that we can do this online, without explicit sorting: when adding a new series, append its name to the list of series.

so, after processing the first batch we sort all current series

src/devhub/devhub.js
```diff
@@ -172,24 +175,45 @@ function batchesToSeries(batches) {
...
+  for (const series of seriesList.values()) {
     let options = {
       title: {
         text: series.name,
       },
       chart: {
         type: "line",
         height: "400px",
+        animations: {
+          enabled: false,
```
matklad, line 202: ❤️

matklad (commented):
I am debating whether we should just slap an iframe with nyrkio, but I think it does make sense to do everything we are doing here!

src/devhub/devhub.js
matklad: Hm, do we even need any code here? JavaScript honors insertion order, no?

cb22: > Hm, do we even need any code here? JavaScript honors insertion order, no?

It does, but because of how the data is structured it means early metrics that now disappeared will still be in their original position high up.

I'll make it so that it'll be ordered first by current batch metric ordering, then by appearance for old batches.

matklad: Hm, I think the following diff would do the trick:

```diff
diff --git a/src/devhub/devhub.js b/src/devhub/devhub.js
index 2a521a262..8a2af7700 100644
--- a/src/devhub/devhub.js
+++ b/src/devhub/devhub.js
@@ -56,6 +56,7 @@ async function mainMetrics() {
     .filter((it) => it.length > 0)
     .map((it) => JSON.parse(it))
     .slice(-1 * maxBatches);
+  batches.reverse() // Latest batches first.
 
   const series = batchesToSeries(batches);
   plotSeries(series, document.querySelector("#charts"), batches.length);
@@ -184,20 +185,7 @@ function batchesToSeries(batches) {
     }
   }
 
-  // Sort the results by batches that have the latest entries first, so metrics that have stopped
-  // drop to the bottom.
-  const sortedResults = new Map();
-  for (const metric of batches[batches.length - 1].metrics) {
-    sortedResults.set(metric.name, results.get(metric.name));
-  }
-
-  for (const [metric, result] of results.entries()) {
-    if (!sortedResults.has(metric)) {
-      sortedResults.set(metric, result);
-    }
-  }
-
-  return sortedResults;
+  return results;
 }
 
 // Plot time series using <https://apexcharts.com>.
```

cb22: That reverses the data too, so it needs [some tweaking](https://github.com/tigerbeetle/tigerbeetle/compare/71f753c98ecd74eaeada53f72eb4d92f7c3d2c86..087606f18d5578def44a7dc595d05ce3d18bee19) to render correctly.

matklad (approved):

PR #1975, opened 2024-05-26
dotnet: Build for .NET 8, the current LTS version

woksin:
Simply update the .NET client project to build for net8.0 alongside the previous version, 7.0 and change the GitHub Workflow to use dotnet tool version 8.0.x instead ofo 7.0.x

.NET 7 is already out of support and .NET 8 is the current LTS version
https://dotnet.microsoft.com/en-us/platform/support/policy/dotnet-core#lifecycle

### DIFF
--- a/.github/workflows/pipeline-core.yml
+++ b/.github/workflows/pipeline-core.yml
@@ -34,14 +34,14 @@ jobs:
     strategy:
       matrix:
         include:
-          - { os: 'ubuntu-latest',  language: 'dotnet',  language_version: '7.0.x' }
+          - { os: 'ubuntu-latest',  language: 'dotnet',  language_version: '8.0.x' }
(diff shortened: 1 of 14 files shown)


batiati:
Thanks so much for the heads-up @woksin, we missed the .net 7.0 EOS notice in May 14th!
I think we can abandon .net7.0, so I also updated the related projects, docs and CI scripts.

batiati (approved):

woksin:
> Thanks so much for the heads-up @woksin, we missed the .net 7.0 EOS notice in May 14th!
> 
> I think we can abandon .net7.0, so I also updated the related projects, docs and CI scripts.

Thanks for cleaning up my cleanup, @batiati. I could also help with making the .net client more idiomatic .NET, at least in terms of setup and configuration, if you would want and appreciate that 😃

PR #1977, opened 2024-05-29
Fix uuid/guid layout issues in dotnet

jkulubya:
The GUID type byte layout in dotnet is
- 4 byte int (typically little endian but can be big endian depending on cpu)
-  2 byte short (typically little endian but can be big endian depending on cpu)
-  2 byte short (typically little endian but can be big endian depending on cpu)
-  8 bytes stored as-is

i.e. the raw bytes you provide in a constructor might be rearranged, and reading the raw bytes of a GUID out on the other side might not give you what you put in.

Secondly, even though the bytes might be rearranged internally, externally, the GUID behaves as if the bytes weren't rearranged i.e., Guid.ToString will print the bytes as you provided them, and not as they are laid out in memory.

What this means is that when treating a GUID as a dumb container for 16 bytes, you must rearrange the bytes going in and out yourself, otherwise the meaning of the GUID changes between the time it's written and when it's read. Guid.ToString and libraries like EF Core and Npgsql take care of this for you.

Take an example of this guid -> e6a48153-cb00-8fab-aa68-fc4387bf8f01.

| Current/New behaviour  | GUID.ToString() (sent to e.g. Npgsql/EF Core) | Guid in-memory layout | Guid.ToUInt128() | UInt128 in-memory layout (sent to TB) |
 |------------|-----------------------------------------------|-----------------------|----------------------------|---------------------------|
New | e6a48153-cb00-8fab-aa68-fc4387bf8f01          | 53-81-A4-E6-00-CB-AB-8F-AA-68-FC-43-87-BF-8F-01 | 2075611103632272357516034594534696166 | E6-A4-81-53-CB-00-8F-AB-AA-68-FC-43-87-BF-8F-01 |
Current | e6a48153-cb00-8fab-aa68-fc4387bf8f01          | 53-81-A4-E6-00-CB-AB-8F-AA-68-FC-43-87-BF-8F-01 | 2075611103632272355506525592271225171 | 53-81-A4-E6-00-CB-AB-8F-AA-68-FC-43-87-BF-8F-01

Notice how with the current behaviour the last column does not match the second, while it does with the new - which meant that queries via the tigerbeetle repl using uuids/guids that were saved in a separate database were failing with a NotFound error


batiati:
Hello @jkulubya, thanks for unveiling this endianness issue!

Do you have a simple failing test case for the current implementation? It would be great if we had a test simulating where/when the internal representation could possibly change (e.g. serialization, external database, etc.), leading to the incorrect end-to-end conversion.

Otherwise, since the TB client is only supported on LittleEndian platforms, it seems more appropriate to use the constructor [`Guid(int a, short b, short c, byte d, byte e, byte f, byte g, byte h, byte i, byte j, byte k)`](https://github.com/microsoft/referencesource/blob/51cf7850defa8a17d815b4700b67116e3fa283c2/mscorlib/system/guid.cs#L112-L128) instead of conditionally swapping the byte order directly into the array.

Another concern is how this GUID representation would interact with other languages, for example, when using the UUID implementation from Java, where the binary representation should be the source of truth on both sides.

jkulubya:
> Do you have a simple failing test case for the current implementation? It would be great if we had a test simulating where/when the internal representation could possibly change (e.g. serialization, external database, etc.), leading to the incorrect end-to-end conversion.

I'm struggling to come up with a useful example but the idea is illustrated below:

```
 [TestMethod]
    public void Failing()
    {
        var str = "00000001-0001-4000-AA00-000000000000";
        var bytes = new byte[] {0x00, 0x00, 0x00, 0x01, 0x00, 0x01, 0x40, 0x00, 0xAA, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00};
        
        var guid1 = Guid.Parse(str);

        var uint128 = BinaryPrimitives.ReadUInt128LittleEndian(bytes);
        var guid2 = uint128.ToGuid();
        
        Assert.AreEqual(guid1, guid2);
    }
```

`bytes` is a direct translation of the uuid/guid in `str`. It would be the typical binary representation of a uuid/guid used in other languages (just 16 bytes in a row), as well as internally in tigerbeetle.

(A) Parsing the string directly, and  (B) dumping the byte array into a Uint128 and then into the Guid constructor will produce different Guids, yet they are supposed to represent the same value.

The problem is that (A) is (basically) how dotnet libraries such as Entity Framework and npgsql read guids, while (B) is how guids would flow into the application from tigerbeetle. (Same problem in reverse on the write side)

> Otherwise, since the TB client is only supported on LittleEndian platforms, it seems more appropriate to use the constructor [Guid(int a, short b, short c, byte d, byte e, byte f, byte g, byte h, byte i, byte j, byte k)](https://github.com/microsoft/referencesource/blob/51cf7850defa8a17d815b4700b67116e3fa283c2/mscorlib/system/guid.cs#L112-L128) instead of conditionally swapping the byte order directly into the array.

`Guid(ReadOnlySpan<byte> b, bool bigEndian)` is probably a better bet than that, but it looks like it also conditionally swaps bytes. The ints and shorts are the root of the issue because that's where the endianness comes into play.

> Another concern is how this GUID representation would interact with other languages, for example, when using the UUID implementation from Java, where the binary representation should be the source of truth on both sides.

As long as the developer takes care to sort out the endianness at the boundary where a uuid/guid value is leaving/entering dotnet land, then other languages/DBs are don't care (In this case the boundary is between guid and uint128). That's what `Guid.ToString` and ORMs do - treat the byte layout of guid as an implementation detail and only ever write out/read in the uuid type we're used to.

Guids/uuids in dotnet shouldn't be this hard :/ - [https://github.com/dotnet/runtime/issues/86084](https://github.com/dotnet/runtime/issues/86084)

jkulubya:
I've updated it to use a `Guid` constructor that will do the swaps internally, and added a test that fails using the old logic but passes now

src/clients/dotnet/TigerBeetle.Tests/UInt128Tests.cs
```diff
@@ -109,14 +106,40 @@ public void BigIntegerExceedU128()
...
+        var roundTripped = parsed.ToGuid().ToUInt128();
+
+        // Send bytes to TB (TB client send raw bytes of dotnet types over the wire) 
+        var roundTrippedBytes = new Span<byte>(&roundTripped, 16);
+
+        var externalIdUInt128 = idSourcedExternally.ToUInt128();
+        var externalIdBytes = new Span<byte>(&externalIdUInt128, 16);
+
+        Assert.IsTrue(roundTrippedBytes.SequenceEqual(bytesStoredInTB));
+        Assert.IsTrue(externalIdBytes.SequenceEqual(bytesStoredInTB));
```
batiati, line 140: I'm sorry, but I'm still failing to see how the boundary between the bytes and the string representation of a Guid would be mixed in a "round trip".

How exactly the guid `00000001-0001-4000-AA00-000000000000` from another database would be stored in TigerBeetle as `{0x00, 0x00, 0x00, 0x01, 0x00, 0x01, 0x40, 0x00, 0xAA, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00}`, since all conversion methods represent it as `{ 0x01, 0x00, 0x00, 0x00, 0x01, 0x00, 0x00, 0x40, 0xAA, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00 }` instead?

Considering this test below (passing in the current implementation), the correct behavior seems to be `bigEndian = false`, the same logic for `BigInteger` as well.

Probably I am missing something here :thinking: 

```c#
    [TestMethod]
    public void TestEndianness()
        // Little-endian bytes of `@as(u128, 0x123456)`:
        var expectedBytes = new byte[16] { 86, 52, 18, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0 };

        // Guid string representation of the same binary value:
        var expectedGuid = Guid.Parse("00123456-0000-0000-0000-000000000000");

        // Literal representation in hex:
        UInt128 expectedU128 = 0x123456;

        // BigInt representation.
        var expectedBigInt = BigInteger.Parse("123456", NumberStyles.HexNumber);

        // Asserting our .Net understanding of Guids:
        // "bigEndian: false" doesn't change the bytes
        bool le = expectedBytes.SequenceEqual(new Guid(expectedBytes, bigEndian: false).ToByteArray());
        Assert.IsTrue(le);
        // "bigEndian: true" changes the bytes
        bool be = expectedBytes.SequenceEqual(new Guid(expectedBytes, bigEndian: true).ToByteArray());
        Assert.IsFalse(be);


        // Asserting our .Net understanding of BigIntegers:
        //  BigInteger.ToByteArray returns the fewest number of bytes possible.

        // "isBigEndian: false" doesn't change the bytes
        var bigIntFewestBytesLE = expectedBigInt.ToByteArray(isUnsigned: true, isBigEndian: false);
        Assert.IsTrue(expectedBytes.Take(bigIntFewestBytesLE.Length).SequenceEqual(bigIntFewestBytesLE));
        Assert.IsTrue(expectedBytes.Skip(bigIntFewestBytesLE.Length).All(b => b == 0));
        // "isBigEndian: true" changes the bytes
        var bigIntFewestBytesBE = expectedBigInt.ToByteArray(isUnsigned: true, isBigEndian: true);
        Assert.IsFalse(expectedBytes.Take(bigIntFewestBytesBE.Length).SequenceEqual(bigIntFewestBytesBE));

        // Binary representation
        Assert.IsTrue(expectedBytes.SequenceEqual(expectedGuid.ToByteArray()));
        Assert.IsTrue(expectedBytes.SequenceEqual(expectedU128.ToArray()));

        // U128 representation
        Assert.IsTrue(expectedU128 == 0x123456UL);
        Assert.IsTrue(expectedU128 == expectedBytes.ToUInt128());
        Assert.IsTrue(expectedU128 == expectedGuid.ToUInt128());
        Assert.IsTrue(expectedU128 == expectedBigInt.ToUInt128());
        Assert.IsTrue(expectedU128 == (ulong)expectedBigInt);

        // Guid representation
        Assert.IsTrue(expectedU128.ToGuid() == expectedGuid);
        Assert.IsTrue(expectedU128 == expectedGuid.ToUInt128());
        Assert.IsTrue(expectedU128.ToGuid().ToString() == expectedGuid.ToString());

        // BigInt representation
        Assert.IsTrue(expectedU128 == expectedBigInt.ToUInt128());
        Assert.IsTrue(expectedU128.ToBigInteger() == expectedBigInt);
        Assert.IsTrue(expectedBytes.SequenceEqual(expectedBigInt.ToUInt128().ToArray()));
    }
```

jkulubya: This is the crux of the issue
> How exactly the guid 00000001-0001-4000-AA00-000000000000 from another database would be stored in TigerBeetle as {0x00, 0x00, 0x00, 0x01, 0x00, 0x01, 0x40, 0x00, 0xAA, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00}, since all conversion methods represent it as { 0x01, 0x00, 0x00, 0x00, 0x01, 0x00, 0x00, 0x40, 0xAA, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00 } instead?

You're correct - Tigerbeetle client reads/writes the bytes as they are laid out in the guid type, but everything else in dotnet doesn't. See [mysql](https://github.com/mysql-net/MySqlConnector/blob/5a377426aa049f90689e3fffa9894ecd7e2f8e7b/src/MySqlConnector/ColumnReaders/GuidBinary16ColumnReader.cs) and [postgres](https://github.com/npgsql/npgsql/blob/main/src/Npgsql/Internal/Converters/Primitive/GuidUuidConverter.cs)

The reason they swap the bytes around is that an "abstract" guid/uuid value of say  `00000001-0001-4000-AA00-000000000000`  is laid out internally as `{ 0x01, 0x00, 0x00, 0x00, 0x01, 0x00, 0x00, 0x40, 0xAA, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00 }` in the ms guid type. The problem is that the rest of the world outside of microsoft  understands that series of bytes to represent a uuid value of `01000000-0100-0040-AA00-000000000000` and NOT `00000001-0001-4000-AA00-000000000000`.

Looking at your test:

<blockquote>

```
var expectedBytes = new byte[16] { 86, 52, 18, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0 };

// Guid string representation of the same binary value:
var expectedGuid = Guid.Parse("00123456-0000-0000-0000-000000000000");
```

</blockquote>

Dotnet would say that the string UUID corresponding to those bytes is `00123456-0000-0000-0000-000000000000`, so your test passes, and those bytes are written to TB. But javascript and java would say it's `56341200-0000-0000-0000-000000000000`.

I ran into the issue with the following steps -
- Generate a random guid to use as account ID
- Save account metadata to postgres
- Save account to TB
- Read GUID from postgres as uint128
- Query TB using the cli for account with that uint128 -> this fails with a NotFound error
- Query TB using the guid in postgres having swapped the bytes -> succeeds

..because the same guid  bytes were written to postgres in the uuid big-endian format, and written to tb in the ms guid format.

jkulubya: My cli is now saying `message queue for peer message_bus.MessageBusType(.client).Connection.Peer{ .replica = 0 } full, dropping request message` but I was going to test adding an account with ID=123456 via cli and then querying using these 2 guids ->

- Guid.Parse("00123456-0000-0000-0000-000000000000")
- Guid.Parse("56341200-0000-0000-0000-000000000000")

Which one should hit?

I'll take a look next week

batiati: Sorry for going back and forth with this, but I'm trying to determine where this boundary is crossed and how transparently or explicitly we should handle this transition.

> ..because the same guid bytes were written to postgres in the uuid big-endian format, and written to tb in the ms guid format.

Let me check my understanding:

When generating the GUID "00123456-0000-0000-0000-000000000000" in .Net and saving it to PG/MySQL, one could read it back as "00123456-0000-0000-0000-000000000000" in C#, but "56341200-0000-0000-0000-000000000000" in Java or JS, right? Nevertheless, only the string representation diverges, the binary representation should still be consistent across different platforms, is it correct?

When you say "Read GUID from Postgres as uint128", is that the database performing the cast?
I mean, since it was a GUID originally, shouldn't you need to read it as a GUID and then call `.ToUInt128()` in C# to perform the proper conversion*?

_*Thanks for spotting the wrong spelling in conver**t**ion :smile:_

batiati: Something to add, from windows-rs

https://github.com/microsoft/windows-rs/blob/15947886be041bebd0fe670fd4e73f18d95100d0/crates/libs/core/src/guid.rs#L50-L65

batiati: > I was going to test adding an account with ID=123456 via cli and then querying using these 2 guids ->
> 
> * Guid.Parse("00123456-0000-0000-0000-000000000000")
> * Guid.Parse("56341200-0000-0000-0000-000000000000")
> 
> Which one should hit?

In C#, it should hit `Guid.Parse("00123456-0000-0000-0000-000000000000")`, but I think the string representation is the less important thing here (since it is an app abstraction to deal with `uint128` though a `[16]u8`). The important thing is that conversions must be coherent with the binary representation in all platforms (like the hex literal "0x123456" is converted to the same BigInt in C#, GO, Java, and JS).

Speaking of hex, you must create the account with ID=1193046, which is the decimal representation of 0x123456.

For the record, I'm convinced that we should fix it at the C# client, since a SELECT into PG would render this UUID as a different string ... It's a huge footgun.

I just need to setup an environment to test all the variants first, in order to avoid introducing another footgun elsewhere!

> I'll take a look next week

Likewise, thanks!

jkulubya:
**A brain dump of my current understanding of this issue:**

Let's take the guid/uuid represented by the string "00123456-0000-0000-0000-000000000000".

If you do a sql insert in postgres with that exact string into a uuid column, it will be stored as this byte sequence `00-12-34-56-00-00-00-00-00-00-00-00-00-00-00-00`.

If you do a `var x = Guid.Parse("00123456-0000-0000-0000-000000000000")` in dotnet, the bytes at `&x` will be `56-34-12-00-00-00-00-00-00-00-00-00-00-00-00-00`.

If you `insert` Guid `x` using a postgres driver in dotnet, it's not a problem, because the driver will rearrange the bytes from `56-34-12-00-00-00-00-00-00-00-00-00-00-00-00-00` to `00-12-34-56-00-00-00-00-00-00-00-00-00-00-00-00` and when you later `select` that same value the driver will do the inverse. The bytes in postgres and dotnet will be ordered differently, but they will have the same "meaning" i.e. calling ".ToString()" in dotnet and the equivalent in postgres will both return "00123456-0000-0000-0000-000000000000".

The issue starts when I try to save `x` to TB by calling `x.ToUInt128()`. `ToUInt128` performs the conversion by just reading the bytes at `&x` without reordering them, i.e. take 16 bytes starting at `&x`. This is where the value in TB falls out of sync with what exists in PG, and by extension, apps in other languages that interface with the same PG instance. TB will have `56-34-12-00-00-00-00-00-00-00-00-00-00-00-00-00` instead of the `00-12-34-56-00-00-00-00-00-00-00-00-00-00-00-00` in PG.

 Then if you `select` the guid using the pg driver in java, and then try to query TB with it, it'll fail with `NotFound`. if you convert the bytes stored in PG to a (little-endian) Uint128 ("1446253056") and query using the TB CLI, it'll also fail because the equivalent little-endian UInt128 stored in TB is instead "1193046".

So that leaves us with a question for TB:
Is the byte representation of this language-agnostic uuid string "00123456-0000-0000-0000-000000000000":

- a) `56-34-12-00-00-00-00-00-00-00-00-00-00-00-00-00` (current behaviour) or
- b) `00-12-34-56-00-00-00-00-00-00-00-00-00-00-00-00` (this PR)?

I vote option B, because:
- Querying will be consistent across all language interfaces to TB
- Option A (for our purposes) is an internal implementation detail which ignores that UUIDs have a well-defined structure 
- Consistency with other database drivers in dotnet
- The RFC for UUIDs states that the uuid represented by string "f81d4fae-7dec-11d0-a765-00a0c91e6bf6" can be represented as binary `11111000 00011101 01001111 10101110 01111101 11101100 00010001 11010000 10100111 01100101 00000000 10100000 11001001 00011110 01101011 111101101` (`F8-1D-4F-AE-7D-EC-11-D0-A7-65-00-A0-C9-1E-6B-F6`) or as big-endian integer `329800735698586629295641978511506172918`, all of which tally with what option B is proposing.

src/clients/dotnet/TigerBeetle.Tests/UInt128Tests.cs
jkulubya: > In C#, it should hit Guid.Parse("00123456-0000-0000-0000-000000000000"), but I think the string representation is the less important thing here (since it is an app abstraction to deal with uint128 though a [16]u8). The important thing is that conversions must be coherent with the binary representation in all platforms (like the hex literal "0x123456" is converted to the same BigInt in C#, GO, Java, and JS).

The binary representation is not consistent from dotnet to other languages. I'd generally agree that the string representation isn't important, but it happens to also be the binary representation in other languages, so it's a nice way to quickly verify

> When generating the GUID "00123456-0000-0000-0000-000000000000" in .Net and saving it to PG/MySQL, one could read it back as "00123456-0000-0000-0000-000000000000" in C#, but "56341200-0000-0000-0000-000000000000" in Java or JS, right? Nevertheless, only the string representation diverges, the binary representation should still be consistent across different platforms, is it correct?
 
`Guid.Parse("00123456-0000-0000-0000-000000000000")` will stored in PG as `00 12 34 56 00...`, but stored in TB as `56 34 12 00...`.

> When you say "Read GUID from Postgres as uint128", is that the database performing the cast?

I mean construct a little-endian UInt128 "by hand" (really I used a c# repl to run `BinaryPrimitives.ReadUInt128LittleEndian(Guid.Parse("uuid-string-copy-pasted-from-db").ToArray(bigEndian:true))` but you can use your language of choice here to construct a little-endian UInt128 from 128 bits).

> I mean, since it was a GUID originally, shouldn't you need to read it as a GUID and then call .ToUInt128() in C# to perform the proper conversion*?

The key here was to avoid using dotnet. I have 128 bits in PG, and I want to use those bits to query tb from another language for what the dotnet app had written earlier. My understanding is that inside of tigerbeetle, at the end of the day an ID is just 128 bits, what they mean or how they are handled (guid/uint128/byte array) exactly is up to the developer. The cli handles the 128 bits as a little endian UInt128 so that's the "other language" that I was using.

jkulubya:
Hi @batiati, I'm still keen on getting this pr over the line. Let me know if you need any more info

batiati:
> I'm still keen on getting this pr over the line. Let me know if you need any more info

Apologies @jkulubya, I owe you feedback!
We put it on hold for a bit until we stabilize our weekly release cadence so we can more easily manage this kind of change (not a breaking API, but a behavior change).

I think we will be able to tackle this during the following weeks.

batiati:
Hello @jkulubya!
Please take a look at PR #2224 where we worked out this issue with the dotnet string representation of GUIDs, and we also implemented consistency tests across all languages.

The fix was quite similar to your PR, however, I noticed that while the blocks were swapped depending on the endianness, it wasn't enough to just set `bigEndian:true` as it would not reinterpret each block correctly, thus not passing the cross-language test.

jkulubya:
I'll leave a few questions on the other PR if you don't mind.

batiati:
> I'll leave a few questions on the other PR if you don't mind.

Thanks @jkulubya!
I'm closing this one, let's continue the thread on #2224!

PR #1979, opened 2024-05-30
Fix go client two phase many example comments

ljnsn:
Just a minor thing I noticed while working on the Python client.


ljnsn:
Seems like the readme for this example is also generated incorrectly, it says debits and credits posted should be 1500 in the end, but if we post and void alternating it should be 900 in the end (as is also asserted in the example), shouldn't it?

matklad:
Both are good catches, thanks!

matklad (approved):

matklad:
https://github.com/tigerbeetle/tigerbeetle/pull/1983 fixes the 900

PR #1981, opened 2024-05-31
VSR/LSM/CLI: Add `--limit-request`

sentientwaffle:
### Overview

Add a runtime request size limit, passed to `tigerbeetle start` and shared back to clients' automatic batching.

This dramatically reduces the replica's lower-bound memory usage (i.e. for `tigerbeetle start --development`).

### StateMachine,LSM: Add `batch_size_limit`, `batch_value_count_limit`

`batch_size_limit` is the runtime-maximum message body size for `command=request`.
The `batch_value_count_limit`s are derived from the `batch_size_limit`.

LSM still uses `value_count_max` for determining data layout (e.g. table schema).

But for runtime allocations, instead use the (potentially smaller) `batch_value_count_limit`.
This can reduce the memory allocated by:
- `CacheMap` (the stash and the `scope_rollback_log`)
- Mutable table
- Immutable table

In `StateMachine`, the value-limit computation is extracted into `batch_value_counts_limit`, so that the logic can be shared between:
- Computing the `batch_value_count_max` of each tree at comptime.
- Computing the `batch_value_count_limit` of each tree at runtime (`StateMachine.init()`).

(Sharing this logic is useful for ensuring that the two computations can't ever diverge.)

### VSR: Add `Register{Request,Reply}.batch_size_limit`

Flow:

1. Replicas have a runtime `batch_size_limit` (which is upper-bounded by `constants.message_body_size_max`).
    - Replica's don't communicate this number to each other!
2. A `vsr.Client` sends a `command=register` to the cluster.
    - This request includes an _empty_ `RegisterRequest` (`RegisterRequest.batch_size_limit = 0`).
    - (In the future, we might want clients to include other fields on the `RegisterRequest`, which is why the client sends one at all.)
3. The primary converts the `request` to a `prepare`. In the process, it overwrites `RegisterRequest.batch_size_limit` with its own `batch_size_limit`.
    - (Stamping the prepare's `RegisterRequest` with the value ensures that every replica will create an identical reply.)
4. The primary relays the `prepare` to the next replica in the chain.
5. When the `operation=register` commits, the reply (`RegisterReply`) includes the `batch_size_limit` from the `RegisterRequest`.
6. The `vsr.Client` stores the `batch_size_limit`.
    - `vsr.Client` panics if it is given a request larger than `batch_size_limit`.
    - The `tb_client` `Context` uses `batch_size_limit` to coalesce batches.
7. If a primary receives a request that is larger than its batch limit, it will evict that client.
8. If any replica tries to prepare or commit a prepare that is larger than its batch limit, it will **panic**.
    - My assumption here is: This is effectively an operator/configuration error, so we want to fail loudly.

Note that this does *not* impact the allocation sizes of:
- Message buffers or,
- Block buffers

(This PR was preceded by https://github.com/tigerbeetle/tigerbeetle/pull/1946.)

### CLI: Add `--limit-request`

Note that this flag is the *total request size*, not the *message body size*.
I wanted the default value to be a round number (`1MiB`) rather than `1048448B` (`1MiB - 128B`).

When `--development` is set, the default `--limit-request` is `32KiB`.
When `--development` is _not_ set, the default is the full message size (`1MiB`).

`--development`'s choice is somewhat arbitrary, but my considerations were:

1. It brings a single-replica cluster with `--development` down below `1GiB` RSS.
2. Subsequent decreases have diminishing returns reducing RSS:
    - `--development --limit-request=64KiB` allocates `1029MiB`
    - `--development --limit-request=32KiB` allocates `994MiB`
    - `--development --limit-request=16KiB` allocates `977MiB`
    - `--development --limit-request=8KiB` allocates `968MiB`


jorangreef:
Appreciate the beautiful description, and all the details around the flow.

> If any replica tries to prepare or commit a prepare that is larger than its batch limit, it will panic.
> My assumption here is: This is effectively an operator/configuration error, so we want to fail loudly.

Great idea to fail loudly. Would we also want a replica to panic if it sees the primary stamping the `RegisterRequest` with a limit different (not only larger) compared to its own?

> vsr.Client panics if it is given a request larger than batch_size_limit.

Related to the previous question, would this mean that it's possible for two different replicas to be started at runtime, each with a different`batch_size_limit`, and then for one primary to replace the other and panic the client?

sentientwaffle:
> Would we also want a replica to panic if it sees the primary stamping the RegisterRequest with a limit different (not only larger) compared to its own?

I think that would make transitioning from a lower-to-higher batch size limit complicated. (i.e. transitioning a development cluster to production. Right now this "just works" with no special code.)

> Related to the previous question, would this mean that it's possible for two different replicas to be started at runtime, each with a different `batch_size_limit`, and then for one primary to replace the other and panic the client?

If the primary changes from low-to-high batch limit, then everything is fine.

If the primary changes from high-to-low batch limit (operator error!), and the client already has a session on the high limit, then the replica will panic if it ever tries to commit a request that doesn't fit in its low limit:

```
@panic("Cannot commit prepare; batch limit too low.");
```

src/clients/dotnet/TigerBeetle.Tests/IntegrationTests.cs
sentientwaffle, line 1: Dotnet integration tests use `--development` replicas, so the query limits are smaller.

src/clients/java/src/test/java/com/tigerbeetle/IntegrationTest.java
sentientwaffle, line 1: Java integration tests use `--development` replicas, so the query limits are smaller.

src/clients/c/tb_client/context.zig
sentientwaffle, line 1: If the application starts sending accounts/transfers/etc, don't assemble them into batches until `.register()` is finished -- i.e. until we know what the batch size is.

jorangreef:
> I think that would make transitioning from a lower-to-higher batch size limit complicated.

Makes sense. Do we document how to transition from low-to-high ("just do it, one by one, rolling") or from high-to-low ("shutdown everything, then apply the same lower limit or you'll get a panic")?

I'm wondering if a fairly common case might also be from high-to-low, if someone spins up a cluster without the development or limit flag, and then adds it gradually.

src/vsr.zig
```diff
@@ -409,6 +409,16 @@ pub const Operation = enum(u8) {
     }
 };
 
+pub const RegisterRequest = extern struct {
+    reserved: [64]u8 = [_]u8{0} ** 64,
```
matklad: Maybe 256, so that it's exactly the same size as the header? No particular reason here, given that we don't otherwise enforce message length alignment, but "two headers" seems like a nice round size.

src/vsr/message_header.zig
```diff
@@ -501,11 +501,11 @@ pub const Header = extern struct {
...
                     if (self.session != 0) return "register: session != 0";
                     if (self.request != 0) return "register: request != 0";
-                    // The .register operation carries no payload:
-                    if (self.checksum_body != checksum_body_empty) {
-                        return "register: checksum_body != expected";
+                    if (self.size != @sizeOf(Header) and
+                        self.size != @sizeOf(Header) + @sizeOf(vsr.RegisterRequest))
+                    {
+                        return "register: size != @sizeOf(Header) [+ @sizeOf(vsr.RegisterRequest)]";
                     }
```
matklad: I wonder what's our long term plan for evolution here. In ten years, do we expect to see the same code, or only the new version?

I am not sure what's best, I see two stable solutions:

* we just always support old clients. We keep such ifs indefinitely and don't have any specific infra in code for them besides the tests
* we aim to move to only newer code eventually, and so we need some source-code sigils to mark "remove this code in the future".

src/clients/dotnet/TigerBeetle.Tests/IntegrationTests.cs
```diff
@@ -1618,7 +1618,7 @@ public TBServer()
 
         process = new Process();
         process.StartInfo.FileName = TB_SERVER;
-        process.StartInfo.Arguments = $"start --addresses=0 --development ./{dataFile}";
+        process.StartInfo.Arguments = $"start --addresses=0 ./{dataFile}";
```
matklad: Hm, why are we dropping development here?

matklad: And we perhaps should add `--development` to Java's format as well (my understanding is that `format --development` affects only direct IO requirements)

src/lsm/groove.zig
```diff
@@ -485,16 +486,17 @@ pub fn GrooveType(
             var objects_cache = try ObjectsCache.init(allocator, .{
                 .cache_value_count_max = options.cache_entries_max,
 
-                // In the worst case, each Map must be able to store the value_count_max (to
+                // In the worst case, each Map must be able to store the batch_value_count_limit (to
```
matklad: comment needs adjustment

src/lsm/groove.zig
```diff
@@ -485,16 +486,17 @@ pub fn GrooveType(
...
-                // value_count_max.
-                .map_value_count_max = @as(u32, ObjectTree.Table.value_count_max) +
-                    (options.prefetch_entries_for_read_max * constants.lsm_batch_multiple),
+                // batch_value_count_limit.
+                .map_value_count_max = constants.lsm_batch_multiple *
+                    (options.tree_options_object.batch_value_count_limit +
+                    options.prefetch_entries_for_read_max),
 
                 // Scopes are limited to a single beat, so the maximum number of entries in a
                 // single scope is value_count_max / constants.lsm_batch_multiple.
```
matklad: Comment needs adjustment

src/vsr/client.zig
```diff
@@ -196,6 +200,48 @@ pub fn Client(comptime StateMachine_: type, comptime MessageBus: type) type {
...
+        pub fn register(self: *Self, callback: Request.RegisterCallback, user_data: u128) void {
+            assert(self.request_inflight == null);
+            assert(self.request_number == 0);
+
+            const message = self.get_message().build(.request);
+            errdefer self.release_message(message.base());
+
+            // We will set parent, session, view and checksums only when sending for the first time:
+            message.header.* = .{
+                .size = @sizeOf(Header) + @sizeOf(vsr.RegisterRequest),
```
matklad, line 213: Do we have some smoke test that sending just `@sizeOf(Header)` bytes works?

I guess its a medium-size architecture question of how we could add such a test, as, presumably, we don't want to support it explicitly in client's code, but we still want to use client in the simulator...

src/vsr/message_header.zig
sentientwaffle: May as well add a sigil in any case -- even if we did always support old clients, having a hint that a particular code path is specifically for old clients could be useful for readability anyhow.

What do you think of `TODO(Deprecated:v0.15.3)`?

src/clients/dotnet/TigerBeetle.Tests/IntegrationTests.cs
sentientwaffle: Ah, dropping it was mistakenly left in from when I was tinkering with the integration tests!

src/vsr/client.zig
sentientwaffle: Yeah, I think this comes down to how we plan to test multi-version compatibility (for both server+client). :thinking:

src/vsr/message_header.zig
matklad: To spell this out, the version use is the latest released version at the time of writing. So, anything up to and including the version _doesn't_ support the new way. At the same time, there's no strict guarantee that the next version( eg, `v0.15.4`) supports the new way, as we don't want to time code to land in a particular version. 

The intention being is that, when we do drop the old code, its old enough that plus or minus version doesn't really matter.

---

`TODO(Deprecated:v0.15.3)?` sounds too wordy to me, and it isn't really a TODO I think, as we don't necessary want to fix this proactively. Maybe `//COMPAT(v0.15.3)`?

matklad: Alternatives: `//Compat(v0.15.3)` or `//TB(v0.15.3)`, as even `COMPAT` feels to shouty to me. `TB` sounds nice, rhymes with `NB` and version makes is visually distinct. If we go with `TB`, we should document it in some dev-docs though, as it isn't exactly self-documenting.

sentientwaffle:
> Makes sense. Do we document how to transition from low-to-high ("just do it, one by one, rolling") or from high-to-low ("shutdown everything, then apply the same lower limit or you'll get a panic")?
> I'm wondering if a fairly common case might also be from high-to-low, if someone spins up a cluster without the development or limit flag, and then adds it gradually.

I'll add some notes to the CLI help text of `--development` explaining how the transitions work.

But imo we should just say that high-to-low  (i.e. transitioning a production replica to `--development`) is not recommended -- I think its edge cases are sufficiently complicated as to fall under our "unstable" interface.

I expect that high-to-low will not be a common case, but if it is, then we can streamline it.

In particular, for high-to-low the operator would need to:
1. ensure that all "recent" requests fit within the new batch size (i.e. requests that are in the WAL)
2. shut down all replicas
3. shut down all clients
4. restart all replicas with `--development`
5. restart all clients

src/state_machine.zig
```diff
@@ -552,19 +537,22 @@ pub fn StateMachineType(
...
+                    );
+                    assert(batch_limit > 0);
 
                     // Clients do not validate batch size == 0,
                     // and even the simulator can generate requests with no events.
                     maybe(input.len == 0);
 
                     if (input.len % event_size != 0) return false;
-                    if (input.len > batch_max * event_size) return false;
+                    if (input.len > batch_limit * event_size) return false;
```
matklad: Can we simplify this to just `input.len > self.batch_size_limit`?

matklad: Ah, no, because we also look at the result set. 

:thinking: so this is actually tricky, this means that the client generates a request, they need to concern themselves not only with the maximum size of input, but also with the maximum size of output. 

Lets maybe simplify this and:

* add comptime check that sizeOf(Result) <= sizeOf(Event)
* add a runtime check in commit that the result is not larger than we expect

src/state_machine.zig
```diff
@@ -1115,8 +1100,9 @@ pub fn StateMachineType(
         ) usize {
             _ = client;
             assert(op != 0);
-            assert(input_valid(operation, input));
+            assert(self.input_valid(operation, input));
             assert(timestamp > self.commit_timestamp or global_constants.aof_recovery);
+            assert(input.len <= self.batch_size_limit);
```
matklad, line 1098: Assert postcondition, that result <= self.batch_size_limit

sentientwaffle: The result is allowed to exceed the `batch_size_limit` -- the batch limit is only for requests, not replies.

matklad: Hm, then why do we compute it as 


```
                    const batch_limit = @min(
                        @divFloor(self.batch_size_limit, event_size),
                        @divFloor(self.batch_size_limit, result_size),
                    );
```

in `operation_batch_max`?

sentientwaffle: Actually now I'm confused.. I think this behavior changed when I wasn't paying attention :sweat_smile:

matklad:
btw, might be good to slap `fuzz vopr` on this one!

sentientwaffle:
(Rebased to resolve a conflict with `main`. The only "new" commit is https://github.com/tigerbeetle/tigerbeetle/pull/1981/commits/8a69afb8a5d77fef91b5493ac3cfb76afcde3248).

src/clients/c/tb_client/context.zig
```diff
@@ -308,10 +332,19 @@ pub fn ContextType(
...
-            if (events.len > constants.message_body_size_max) {
+            // Make sure the packet.data wouldn't overflow a request, and that the corresponding
+            // results won't overflow a reply.
+            const events_batch_max = switch (operation) {
+                .pulse => unreachable,
+                inline else => |operation_comptime| StateMachine.operation_batch_max(
+                    operation_comptime,
+                    self.batch_size_limit.?,
+                ),
+            };
```
matklad, line 343: Do we need code like this in client.zig (that is, non-c version)?

sentientwaffle:
Ahh right, `input_valid` requires a `StateMachine` reference... I'll give some more thought on the best way to organize this.

sentientwaffle:
> Do we need code like this in client.zig (that is, non-c version)?

:thinking: In the interest of not growing the statemachine interface even more, I think I'll just let tb_client handle this validation. If `client.zig` is given something that is not valid, it will send it and end up being evicted.

sentientwaffle:
Rebased b/c cfo was hitting a bug that is now fixed on `main`.

src/vsr/replica.zig
```diff
@@ -4260,7 +4295,28 @@ pub fn ReplicaType(
...
+                };
+            } else {
+                assert(prepare.header.size == @sizeOf(vsr.Header) + @sizeOf(vsr.RegisterRequest));
+
+                const register_request = std.mem.bytesAsValue(
+                    vsr.RegisterRequest,
+                    prepare.body()[0..@sizeOf(vsr.RegisterRequest)],
+                );
+                assert(register_request.batch_size_limit > 0);
+                assert(register_request.batch_size_limit <= constants.message_body_size_max);
```
matklad, line 4313: Can we `assert` that the limit is `<=` our current limit? (it could be `<` if we restart with a larger limit)

src/vsr/replica.zig
```diff
@@ -5871,6 +5945,34 @@ pub fn ReplicaType(
...
+            assert(request.header.operation == .register);
+            assert(request.header.request == 0);
+
+            if (request.header.size == @sizeOf(vsr.Header)) {
+                // Old clients don't send a RegisterRequest.
+            } else {
+                assert(request.header.size == @sizeOf(vsr.Header) + @sizeOf(vsr.RegisterRequest));
+
+                const batch_size_limit = self.request_size_limit - @sizeOf(vsr.Header);
+                assert(batch_size_limit > 0);
```
matklad, line 5963: If we assert that it is `>0` (rather than `>=`) here, let's also add an assert to the same effect to `init`:

```
            assert(request_size_limit <= constants.message_size_max);
            assert(request_size_limit > @sizeOf(Header));

```

to not postepone the crash until register.

src/state_machine.zig
```diff
@@ -552,19 +537,18 @@ pub fn StateMachineType(
...
 
                     // Clients do not validate batch size == 0,
                     // and even the simulator can generate requests with no events.
                     maybe(input.len == 0);
 
                     if (input.len % event_size != 0) return false;
-                    if (input.len > batch_max * event_size) return false;
+                    if (input.len > batch_limit * event_size) return false;
                 },
             }
```
matklad, line 544: assert `input.len < self.batch_size_limit`, as different switch branches use diferent logic to enforce that invariant.

matklad: Ah, I see that we do that at the call-site, but it still feels worthwhile to do this inside, so that one can AND reasoning from each branch.

src/state_machine.zig
```diff
@@ -1015,11 +998,9 @@ pub fn StateMachineType(
 
             var scan_lookup_buffer = std.mem.bytesAsSlice(
                 Transfer,
-                self.scan_lookup_buffer[0 .. @sizeOf(Transfer) *
-                    // We must be constrained to the same limit as `create_transfers`.
-                    constants.batch_max.create_transfers],
+                // We must be constrained to the same limit as `create_transfers`.
+                self.scan_lookup_buffer[0..self.scan_buffer_size_create_transfers],
```
matklad: Hm, what's the reason for this constrained? If we have a small `request_size_limit`, what would break if a scan returns a large batch? Is it just the memory for scan buffers themselves?

src/state_machine.zig
```diff
@@ -1929,73 +1911,178 @@ pub fn StateMachineType(
...
+                        .debit_account_id = .{ .batch_value_count_limit = batch_values_limit.transfers.debit_account_id },
+                        .credit_account_id = .{ .batch_value_count_limit = batch_values_limit.transfers.credit_account_id },
+                        .user_data_128 = .{ .batch_value_count_limit = batch_values_limit.transfers.user_data_128 },
+                        .user_data_64 = .{ .batch_value_count_limit = batch_values_limit.transfers.user_data_64 },
+                        .user_data_32 = .{ .batch_value_count_limit = batch_values_limit.transfers.user_data_32 },
+                        .pending_id = .{ .batch_value_count_limit = batch_values_limit.transfers.pending_id },
+                        .ledger = .{ .batch_value_count_limit = batch_values_limit.transfers.ledger },
+                        .code = .{ .batch_value_count_limit = batch_values_limit.transfers.code },
+                        .amount = .{ .batch_value_count_limit = batch_values_limit.transfers.amount },
+                        .expires_at = .{ .batch_value_count_limit = batch_values_limit.transfers.expires_at },
```
matklad, line 1969: Feels like it's dangerously easy to make a typo here. Though I guess let's worry about that if we have to change the code

src/lsm/groove.zig
```diff
@@ -485,16 +486,17 @@ pub fn GrooveType(
             var objects_cache = try ObjectsCache.init(allocator, .{
```
matklad, line 489: I think here we might want to check that runtime options are compatible with comptime ones?  That limit <= max

src/testing/state_machine.zig
```diff
@@ -323,7 +327,7 @@ fn WorkloadType(comptime StateMachine: type) type {
             workload.requests_sent += 1;
 
             // +1 for inclusive limit.
-            const size = workload.random.uintLessThan(usize, constants.message_body_size_max + 1);
+            const size = workload.random.uintAtMost(usize, workload.options.batch_size_limit);
```
matklad, line 330: :heart:

src/clients/c/tb_client/context.zig
```diff
@@ -308,10 +332,19 @@ pub fn ContextType(
...
+            if (@divExact(events.len, event_size) > events_batch_max) {
                 return self.on_complete(packet, error.TooMuchData);
```
matklad, line 345: This is unrelated to the PR, but I am wondering, with auto-batching, do we have a TOCTOU here, where we check size in `request`, but use later, after some user's code has been run, in `submit`?

We are definitely missing assertions there!

matklad: cc @batiati

src/simulator.zig
```diff
@@ -120,6 +120,20 @@ pub fn main() !void {
...
+            const event_size = @sizeOf(StateMachine.Event(operation));
+            event_size_max = @max(event_size_max, event_size);
+        }
+        break :batch_size_limit_min event_size_max;
+    };
+    const batch_size_limit: u32 = if (random.boolean())
+        constants.message_body_size_max
+    else
+        batch_size_limit_min +
+            random.uintAtMost(u32, constants.message_body_size_max - batch_size_limit_min);
```
matklad, line 135: Ideally, we'd check that a cluster with _different_ batch sizes behaves as expected. 

We actually can do it the same way we do with upgrades? Ratcheting batch_size up over time?

src/vsr/replica_test.zig
```diff
@@ -1536,6 +1578,11 @@ const TestContext = struct {
         };
     }
 
+    pub fn client_bus(t: *TestContext, client_index: usize) !*TestClientBus {
+        // Reuse one of `Cluster.clients`' ids since the Network preallocated links for it.
+        return TestClientBus.init(t, t.cluster.clients[client_index].id);
```
matklad, line 1583: :exploding_head: glorious! Really like how we just put a separate bus alongside what we already have!

matklad (commented):
Just couple of more asserts we could add, but otherwise this looks great. 

Also, it's surprising how little code needs to actually _change_, the bulk of diff is just threading the values, but then, it's like one-line change to CacheMap parameter, and a really tiny diff to table_memory!

matklad:
Oh, a potentially bigger thing. Right now, `--limit-request` is an experimental flag (I think we got that infra merged _after_ this PR was open). But, at the same time, the actual logic here affects our actual wire protocol. So perhaps this should _not_ be an experimental flag?

sentientwaffle:
> Oh, a potentially bigger thing. Right now, --limit-request is an experimental flag (I think we got that infra merged after this PR was open). But, at the same time, the actual logic here affects our actual wire protocol. So perhaps this should not be an experimental flag?

And also responding to https://github.com/tigerbeetle/tigerbeetle/pull/1981#discussion_r1635185342

---

"Experimental" isn't the best label for `--limit-request`... "advanced" would be better, but I think the semantics of `--experimental` are a good fit. (The functionality is hidden but still available.)

Mixed-batch-size clusters are tricky: Replica's don't actually agree to batch sizes like they do for e.g. upgrades. If a "high" batch size is the primary, then it can cause the "low"s to panic. So I don't think there' a nice way to test this in the simulator. (And I don't think it is something that we "support".) The CLI help describes the situation -- which, since we only expose this functionality via `--development`, is relatively uncomplicated. (Not counting the part hidden behind `--experimental`).

```
        \\        Since this shrinks the batch size, note that:
        \\        * All replicas should use the same batch size. That is, if any replica in the cluster has
        \\          "--development", then all replicas should have "--development".
        \\        * It is always possible to increase the batch size by restarting without "--development".
        \\        * Shrinking the batch size of an existing cluster is possible, but not recommended.
```

The assumption is that a cluster transitioning from development to production is not _already_ in production, so it is okay to restart the entire cluster without worrying about those crashes.

src/state_machine.zig
sentientwaffle: Great catch; it doesn't need to be constrained! I was getting this mixed up with `lookup_{accounts,transfers}`, where the output size corresponds to the input size.

src/simulator.zig
sentientwaffle: (see https://github.com/tigerbeetle/tigerbeetle/pull/1981#issuecomment-2161295191)

src/state_machine.zig
sentientwaffle: Ah, wait, pulses do need to be constrained! See: https://github.com/tigerbeetle/tigerbeetle/pull/1981/commits/2d06376abebf42deb87fbd06e41777f512e31bf2

sentientwaffle:
Rebased to resolve a conflict with main.

https://github.com/tigerbeetle/tigerbeetle/pull/1981/commits/f96a1d4a5728f8f851131007f22291b86e7b34d1 and https://github.com/tigerbeetle/tigerbeetle/pull/1981/commits/fccdbf5490b252ef5deada00602c793f01e235df are the only new commits since last review.

matklad (dismissed):

matklad:
VOPR failure introduced by the last commit

```
λ git rev-parse HEAD 
fccdbf5490b252ef5deada00602c793f01e235df

λ ./zig/zig build  simulator_run -- --lite 11697763002021897417
thread 26152 panic: on_lookup_accounts: account data mismatch
/home/matklad/p/tb/sim/src/state_machine/auditor.zig:426:21: 0x3dd6d8 in on_lookup_accounts (simulator)
                    @panic("on_lookup_accounts: account data mismatch");
                    ^
/home/matklad/p/tb/sim/src/state_machine/workload.zig:338:68: 0x38019e in on_reply (simulator)
                .lookup_accounts => self.auditor.on_lookup_accounts(
                                                                   ^
/home/matklad/p/tb/sim/src/simulator.zig:827:48: 0x371b22 in on_cluster_reply (simulator)
                    simulator.workload.on_reply(
                                               ^
/home/matklad/p/tb/sim/src/testing/cluster.zig:680:37: 0x36e934 in client_on_reply (simulator)
            cluster.on_cluster_reply(cluster, client_index, request_message, reply_message);
                                    ^
/home/matklad/p/tb/sim/src/vsr/client.zig:439:34: 0x388c44 in on_reply (simulator)
                on_reply_callback(self, inflight.message, reply);
                                 ^
/home/matklad/p/tb/sim/src/vsr/client.zig:179:44: 0x32f96e in on_message (simulator)
                .reply => |m| self.on_reply(m),
                                           ^
/home/matklad/p/tb/sim/src/testing/cluster/network.zig:317:39: 0x4fe42d in deliver_message (simulator)
        target_bus.on_message_callback(target_bus, target_message);
                                      ^
/home/matklad/p/tb/sim/src/testing/packet_simulator.zig:403:45: 0x446e04 in tick (simulator)
                        link_packet.callback(link_packet.packet, path);
                                            ^
/home/matklad/p/tb/sim/src/testing/cluster/network.zig:136:38: 0x3fe30c in tick (simulator)
                                     ^
/home/matklad/p/tb/sim/src/testing/cluster.zig:401:33: 0x3a8d16 in tick (simulator)
            cluster.network.tick();
                                ^
/home/matklad/p/tb/sim/src/simulator.zig:552:31: 0x3767c9 in tick (simulator)
        simulator.cluster.tick();
                              ^
/home/matklad/p/tb/sim/src/simulator.zig:306:23: 0x374317 in main (simulator)
        simulator.tick();
                      ^
/home/matklad/p/tb/main/zig/lib/std/start.zig:574:37: 0x32a37e in posixCallMainAndExit (simulator)
                                    ^
/home/matklad/p/tb/main/zig/lib/std/start.zig:243:5: 0x329e61 in _start (simulator)
```

matklad (approved):

PR #1982, opened 2024-06-03
multiversioning/flags: gate most flags behind --experimental

cb22:
This makes it required to specify `--experimental` when starting TigerBeetle if you are using flags that aren't explicitly documented in the help message.

Stable flags for `start` are specified in an allowlist - everything else is considered experimental otherwise.

The main reason behind this is multiversion upgrades: by starting with a small and intentionally limited set of flags, we can limit the surface area of what we need to keep stable to ensure different versions keep compatibility with the same CLI.

### DIFF
--- a/src/tigerbeetle/cli.zig
+++ b/src/tigerbeetle/cli.zig
@@ -42,21 +42,27 @@ const CliArgs = union(enum) {
     },
 
     start: struct {
+        // Stable CLI arguments.
         addresses: []const u8,
-        limit_storage: flags.ByteSize = .{ .value = constants.storage_size_limit_max },
+        cache_grid: ?flags.ByteSize = null,
+        development: bool = false,
+        positional: struct {
+            path: [:0]const u8,
+        },
+
+        // Everything below here is considered experimental, and requires `--experimental` to be
+        // set. Experimental flags disable automatic upgrades with multiversion binaries; each
+        // replica has to be manually restarted.
+        // Experimental flags must default to null.
+        experimental: bool = false,
+
+        limit_storage: ?flags.ByteSize = null,
         limit_pipeline_requests: ?u32 = null,
         cache_accounts: ?flags.ByteSize = null,
         cache_transfers: ?flags.ByteSize = null,
         cache_transfers_pending: ?flags.ByteSize = null,
         cache_account_balances: ?flags.ByteSize = null,
-        cache_grid: ?flags.ByteSize = null,
-        memory_lsm_manifest: flags.ByteSize =
-            .{ .value = constants.lsm_manifest_memory_size_default },
-        development: bool = false,
-
-        positional: struct {
-            path: [:0]const u8,
-        },
+        memory_lsm_manifest: ?flags.ByteSize = null,
     },
 
     version: struct {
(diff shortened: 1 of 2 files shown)


src/tigerbeetle/cli.zig
```diff
@@ -42,21 +42,28 @@ const CliArgs = union(enum) {
...
     start: struct {
+        // Stable CLI interface.
         addresses: []const u8,
-        limit_storage: flags.ByteSize = .{ .value = constants.storage_size_limit_max },
+        cache_grid: ?flags.ByteSize = null,
+        positional: struct {
+            path: [:0]const u8,
+        },
+
+        development: bool = false,
```
matklad: Move `development` before `positional`? Or is there some reason why this one is somewhat separate?

src/tigerbeetle/cli.zig
```diff
@@ -377,6 +375,34 @@ pub fn parse_args(allocator: std.mem.Allocator, args_iterator: *std.process.ArgI
...
+                "addresses",   "cache_grid",   "positional",
+                "development", "experimental",
+            };
+            inline for (std.meta.fields(@TypeOf(start))) |field| {
+                comptime var stable_field = false;
+                inline for (stable_args) |stable_arg| {
+                    if (comptime std.mem.eql(u8, field.name, stable_arg)) {
+                        stable_field = true;
+                    }
+                }
```
matklad: maybe 
```suggestion
                const stable_field = comptime for (stable_args) |stable_arg| {
                    if (comptime std.mem.eql(u8, field.name, stable_arg)) {
                        break true;
                    }
                } else false;
                if (stable_field) continue;
```

src/tigerbeetle/cli.zig
```diff
@@ -377,6 +375,34 @@ pub fn parse_args(allocator: std.mem.Allocator, args_iterator: *std.process.ArgI
...
+
+                if (comptime !stable_field) {
+                    const flag_name = flags.flag_name(field);
+
+                    // If you've added a flag and get a comptime error here, it's likely because
+                    // we require experimental flags to default to null.
+                    if (@field(start, field.name) != null and !start.experimental) {
```
matklad: comptime check that field's default is `null`, to protect against `foo: ?u32 = 0`.

src/tigerbeetle/cli.zig
cb22: No strong reason, when I was writing it out I was thinking in terms of what disables multiversion upgrading.

src/tigerbeetle/cli.zig
```diff
@@ -377,6 +374,36 @@ pub fn parse_args(allocator: std.mem.Allocator, args_iterator: *std.process.ArgI
...
+            };
+            inline for (std.meta.fields(@TypeOf(start))) |field| {
+                const stable_field = comptime for (stable_args) |stable_arg| {
+                    if (std.mem.eql(u8, field.name, stable_arg)) {
+                        break true;
+                    }
+                } else false;
+                if (stable_field) continue;
+
+                if (comptime !stable_field) {
```
sentientwaffle: Isn't the `if (stable_field) continue;` unnecessary, since it is followed by `if (comptime !stable_field) {`?

src/tigerbeetle/cli.zig
```diff
@@ -377,6 +374,36 @@ pub fn parse_args(allocator: std.mem.Allocator, args_iterator: *std.process.ArgI
...
         },
         .start => |start| {
+            // Allowlist of stable flags. --development will disable automatic multiversion
+            // upgrades too, but the flag itself is stable.
+            const stable_args = .{
+                "addresses",   "cache_grid",   "positional",
+                "development", "experimental",
+            };
+            inline for (std.meta.fields(@TypeOf(start))) |field| {
+                const stable_field = comptime for (stable_args) |stable_arg| {
```
sentientwaffle, line 384: Assert that there's no typos in `stable_args`:

```suggestion
                const stable_field = comptime for (stable_args) |stable_arg| {
                    assert(std.meta.fieldIndex(@TypeOf(start), stable_arg) != null);
```

src/tigerbeetle/cli.zig
```diff
@@ -42,21 +42,27 @@ const CliArgs = union(enum) {
     },
 
     start: struct {
+        // Stable CLI interface.
```
sentientwaffle: Would "arguments" make more sense than "interface" here?

src/tigerbeetle/cli.zig
cb22: Yep!

sentientwaffle (approved):

PR #1985, opened 2024-06-03
Grid/Scrubber: Random tour origin

sentientwaffle:
(Follow-up for https://github.com/tigerbeetle/tigerbeetle/pull/1851)

Start iterating tables for scrubbing from a random level/tree.

From the code:

    // Each replica's scrub origin is chosen independently.
    // This reduces the chance that the same block across multiple replicas can bitrot
    // without being discovered and repaired by a scrubber.
    //
    // To accomplish this, try to select an origin uniformly across all blocks:
    // - Bias towards levels with more tables.
    // - Bias towards trees with larger Values (since their tables have more blocks).
    // - (Though, for ease of implementation, the origin is always at the beginning of a
    //   tree's level, never in the middle.)

---

Also: Don't reset the tour position when state sync is triggered.

If we reset the tour, then we effectively bias scrubbing towards the `tour_tables_origin`. (Uniform scrubbing is preferable.)

### DIFF
--- a/src/vsr/grid_scrubber.zig
+++ b/src/vsr/grid_scrubber.zig
@@ -232,10 +232,9 @@ pub fn GridScrubberType(comptime Forest: type) type {
                 }
             }
 
-            scrubber.tour = .init;
-            scrubber.tour_blocks_scrubbed_count = 0;
-            if (scrubber.tour_tables_origin) |tour_tables_origin| {
-                scrubber.tour_tables = WrappingForestTableIterator.init(tour_tables_origin);
+            if (scrubber.tour == .table_data) {
+                // Skip scrubbing the table data; the table may not exist when state sync finishes.
+                scrubber.tour = .table_index;
             }
         }
 
--- a/src/vsr/grid_scrubber.zig
+++ b/src/vsr/grid_scrubber.zig
@@ -161,7 +161,7 @@ pub fn GridScrubberType(comptime Forest: type) type {
 
         pub fn open(scrubber: *GridScrubber, random: std.rand.Random) void {
             // Compute the tour origin exactly once.
-            if (scrubber.tour_tables_origin) |_| {
+            if (scrubber.tour_tables_origin != null) {
                 return;
             }
 
@@ -171,49 +171,35 @@ pub fn GridScrubberType(comptime Forest: type) type {
             //
             // To accomplish this, try to select an origin uniformly across all blocks:
             // - Bias towards levels with more tables.
-            // - Bias towards trees with larger Values (since their tables have more blocks).
+            // - Bias towards trees with more blocks per table.
             // - (Though, for ease of implementation, the origin is always at the beginning of a
             //   tree's level, never in the middle.)
             assert(scrubber.tour == .init);
 
-            const tree_count = Forest.tree_infos.len;
+            scrubber.tour_tables_origin = .{
+                .level = 0,
+                .tree_id = Forest.tree_infos[0].tree_id,
+            };
 
-            var weights: [tree_count][constants.lsm_levels]u64 = undefined;
             var weights_sum: u64 = 0;
             for (0..constants.lsm_levels) |level| {
-                inline for (Forest.tree_infos, 0..) |tree_info, tree_index| {
+                inline for (Forest.tree_infos) |tree_info| {
                     const tree_id = comptime Forest.tree_id_cast(tree_info.tree_id);
                     const tree = scrubber.forest.tree_for_id_const(tree_id);
                     const levels = &tree.manifest.levels;
-                    weights[tree_index][level] =
-                        levels[level].tables.len() * @sizeOf(tree_info.Tree.Table.Value);
-                    weights_sum += levels[level].tables.len();
-                }
-            }
-
-            scrubber.tour_tables_origin = origin: {
-                if (weights_sum == 0) {
-                    break :origin .{
-                        .level = 0,
-                        .tree_id = Forest.tree_infos[0].tree_id,
-                    };
-                }
-
-                var origin_offset = random.uintLessThan(u64, weights_sum);
-                for (0..constants.lsm_levels) |level| {
-                    inline for (Forest.tree_infos, 0..) |tree_info, tree_index| {
-                        if (origin_offset >= weights[tree_index][level]) {
-                            origin_offset -= weights[tree_index][level];
-                        } else {
-                            break :origin .{
+                    const tree_level_weight = @as(u64, levels[level].tables.len()) *
+                        tree_info.Tree.Table.index.data_block_count_max;
+                    if (tree_level_weight > 0) {
+                        weights_sum += tree_level_weight;
+                        if (random.uintLessThan(u64, weights_sum) < tree_level_weight) {
+                            scrubber.tour_tables_origin = .{
                                 .level = @intCast(level),
                                 .tree_id = tree_info.tree_id,
                             };
                         }
                     }
                 }
-                unreachable;
-            };
+            }
 
             scrubber.tour_tables = WrappingForestTableIterator.init(scrubber.tour_tables_origin.?);
 
--- a/src/vsr/grid_scrubber.zig
+++ b/src/vsr/grid_scrubber.zig
@@ -18,8 +18,6 @@
 //! and try to repair the block from another replica, only to discover that the copy of the block on
 //! the remote replica's disk is *also* faulty.
 //!
-//! TODO Start replicas scrubbing from distinct/random offsets in the tour to farther minimize risk
-//! of cluster data loss. (Right now replicas will often have identical scrubbing schedules.)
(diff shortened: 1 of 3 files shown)


src/vsr/grid_scrubber.zig
```diff
@@ -156,6 +159,71 @@ pub fn GridScrubberType(comptime Forest: type) type {
             scrubber.* = undefined;
         }
 
+        pub fn open(scrubber: *GridScrubber, random: std.rand.Random) void {
+            // Compute the tour origin exactly once.
+            if (scrubber.tour_tables_origin) |_| {
```
matklad:
```suggestion
            if (scrubber.tour_tables_origin != null) {
```

this looks slighly less surprising to me, though either way works

src/vsr/grid_scrubber.zig
```diff
@@ -156,6 +159,71 @@ pub fn GridScrubberType(comptime Forest: type) type {
...
+            var weights: [tree_count][constants.lsm_levels]u64 = undefined;
+            var weights_sum: u64 = 0;
+            for (0..constants.lsm_levels) |level| {
+                inline for (Forest.tree_infos, 0..) |tree_info, tree_index| {
+                    const tree_id = comptime Forest.tree_id_cast(tree_info.tree_id);
+                    const tree = scrubber.forest.tree_for_id_const(tree_id);
+                    const levels = &tree.manifest.levels;
+                    weights[tree_index][level] =
+                        levels[level].tables.len() * @sizeOf(tree_info.Tree.Table.Value);
+                    weights_sum += levels[level].tables.len();
```
matklad: Hm, this is not the sum of weights, no? I'd expect to see

```
weights_sum += weights[tree_index][level]
```

src/vsr/grid_scrubber.zig
```diff
@@ -156,6 +159,71 @@ pub fn GridScrubberType(comptime Forest: type) type {
...
+
+            var weights: [tree_count][constants.lsm_levels]u64 = undefined;
+            var weights_sum: u64 = 0;
+            for (0..constants.lsm_levels) |level| {
+                inline for (Forest.tree_infos, 0..) |tree_info, tree_index| {
+                    const tree_id = comptime Forest.tree_id_cast(tree_info.tree_id);
+                    const tree = scrubber.forest.tree_for_id_const(tree_id);
+                    const levels = &tree.manifest.levels;
+                    weights[tree_index][level] =
+                        levels[level].tables.len() * @sizeOf(tree_info.Tree.Table.Value);
```
matklad: shouldn't be this additionally weighted by table_value_count_max?

matklad: Id also maybe add an explicit `@as(u64)` here somewhere. I _think_ as stated the computation is carried over in `usize`, and we want u64 here.

src/vsr/grid_scrubber.zig
```diff
@@ -156,6 +159,71 @@ pub fn GridScrubberType(comptime Forest: type) type {
...
+                            origin_offset -= weights[tree_index][level];
+                        } else {
+                            break :origin .{
+                                .level = @intCast(level),
+                                .tree_id = tree_info.tree_id,
+                            };
+                        }
+                    }
+                }
+                unreachable;
```
matklad: Took me a second to convince myself that this is indeed unreacheable. I wonder if the code would end up simpler if we do it in one pass, without computing the array of weights. On a loop iteration, you have `weight_running_sum` and `weight_this_tree`, and toss a coin to decide whether the origin should be updated to point to the current tree.  So, on each iteration you pick a random tree from the set of trees considered so far:

```zig
            scrubber.tour_tables_origin = .{
                .level = 0,
                .tree_id = Forest.tree_infos[0].tree_id,
            };

            var weights_sum: u64 = 0;
            for (0..constants.lsm_levels) |level| {
                inline for (Forest.tree_infos) |tree_info| {
                    const tree_id = comptime Forest.tree_id_cast(tree_info.tree_id);
                    const tree = scrubber.forest.tree_for_id_const(tree_id);
                    const tree_level_weight =
                        tree.manifest.levels[level].tables.len() * @sizeOf(tree_info.Tree.Table.Value);
                    if (tree_level_weight > 0) {
                        weights_sum += tree_level_weight;
                        if (random.uintLessThan(u64, weights_sum) < tree_level_weight) {
                            scrubber.tour_tables_origin = .{
                                .level = level,
                                .tree_id = tree_info.tree_id,
                            };
                        }
                    }
                }
            }

```

src/vsr/grid_scrubber.zig
sentientwaffle: Good catch!

src/vsr/grid_scrubber.zig
sentientwaffle: It is much simpler, but I'm having a hard time proving to myself that it is not biased towards the later levels/trees. :thinking:

src/vsr/grid_scrubber.zig
sentientwaffle: I'm going to combine `@sizeOf(Value)` and `table_value_count_max` into:

```
                    const tree_level_weight =
                        @as(u64, levels[level].tables.len()) *
                        tree_info.Tree.Table.index.data_block_count_max;
```

since that is more directly what we care about.

src/vsr/grid_scrubber.zig
matklad: The loop invariant here is that we pick tree "uniformly" across the the trees considered so far. So the induction step is that we picked tree at random among the first n trees, and we need to decide whether to keep our choice, or to pick the n+1 tree. And we know the weight of this new tree, as well as the weights of trees so far.

I think this is called reservoir sampling: https://en.wikipedia.org/wiki/Reservoir_sampling

matklad: Though, again, I am not sure what's simpler two loops which manage state, or one loop which tries to be smart.

sentientwaffle: I like the reservoir sampling sampling version -- much easier to prove that it won't crash.

src/vsr/grid_scrubber.zig
```diff
@@ -156,6 +159,58 @@ pub fn GridScrubberType(comptime Forest: type) type {
             scrubber.* = undefined;
         }
 
+        pub fn open(scrubber: *GridScrubber, random: std.rand.Random) void {
```
matklad, line 162: I am wondering if, for simplicity, scrubber should own a separate copy of Random? Random is a pointer, so here we are relying on the fact that the replica is immovable. 

In Rust terms, Random actually holds a lifetime which infects the scrubber, and it seems that holding an independent PRNG by value could be simpler

sentientwaffle: > we are relying on the fact that the replica is immovable.
> Random actually holds a lifetime which infects the scrubber

How so? The scrubber doesn't actually store a reference to `Random`, it just uses it within the scope of this function.

matklad: IIRC random internally is two pointers: pointer to table and type-erased data. The same pattern as Allocator.

sentientwaffle: Right, but `scrubber` is only borrowing those pointers for the duration of `open()`, it doesn't store them on `GridScrubber`.

matklad: Ahhh, right, of course!

matklad (dismissed):
LGTM, though I haven’t looked at cfo results

sentientwaffle:
:facepalm: I forgot to add it to the CFO. I'll add it for a bit before I merge!

sentientwaffle:
CFO found a bug; will fix tomorrow!

sentientwaffle:
Turns out to be a subtler bug than expected! Preexisting too, I think. https://github.com/tigerbeetle/tigerbeetle/pull/1985/commits/187712b4d9681506af7bca67235c9511e80a8265

matklad:
Might want to include

```
Seed: ./zig/zig build -Drelease simulator_run -- --lite 13919299316979954169	
```

into the last commit message. 

I don't think I've ever used a seed from a commit message, but I think it's still useful to record in history, just in case we want to draw some vopr graphs in the future, or run some ML

matklad:
This is extreme git pedantry, but I suggest the exact form `Seed: ....`. That way it becomes a git trailer, and can get extracted by automation:

```
λ git log --format=%B -n 1 a33279526228764139d52798d3254fc092d3d333 | git interpret-trailers --parse
Seed: ./zig/zig build -Dsimulator-state-machine=testing simulator_run -- 11290157931574340033
```

Again, this is of questionable direct usefulness, but, if there's a standard, we might as well use it!

matklad (approved):
Wow, that last bug indeed is subtle!

matklad (approved):

sentientwaffle:
Updated! I didn't know about git trailers, that is neat.

PR #1987, opened 2024-06-05
vopr: fix WALInvalid crash in the simulator

matklad:
If the entire Zone.wal_headers is corrupted, the replica becomes permanently unavailable (returns `WALInvalid` from `open`). In the simulator, there are only two WAL sectors, which could both get corrupted when a replica crashes while writing them simultaneously. To keep the replica operational, arbitrarily repair one of the sectors.

In production `journal_iops_write_max < header_sector_count`, which makes is impossible to get torn writes for all journal header sectors at the same time.

Seed: ./zig/zig build -Dsimulator-state-machine=testing simulator_run -- 11290157931574340033


sentientwaffle (approved):
`todos -= 1;`

Issue #1968, opened 2024-05-21
Guidance Needed for Migrating Historical Transactions with Timestamps to TigerBeetle


breakds:
Hello,

We are currently in the process of migrating our accounting system to TigerBeetle and have encountered a challenge with preserving the original timestamps of our historical transactions. Our current system contains three years' worth of transaction data that we need to import into TigerBeetle.

## Issue:

Based on the documentation, we understand that TigerBeetle uses logical time for transactions, and timestamps cannot be updated after being set. This constraint poses a challenge for us as we need to maintain the integrity and sequence of our historical transactions.

## Specific Questions:

1. What is the recommended approach for migrating historical transactions with their original timestamps to TigerBeetle, considering the logical time constraints?
2. Are there any best practices or tools within TigerBeetle that can assist with this type of data migration?
3. A slightly 

## Related Question:

If the TigerBeetle servers go down (e.g., due to a power failure) and it takes 2 hours to restore, will transactions that occur during that downtime, once backfilled, be timestamped 2 hours later than they actually occurred? This discrepancy could affect hourly balance reports. What is the recommended approach to handle this scenario to ensure accurate reporting?

Any guidance or recommendations you can provide would be greatly appreciated. Thank you for your support!

matklad:
Thanks for reaching out @breakds!

This particular area is a work-in-progress, so sadly there's no a ready-made solution yet (or rather, we have some code in place to do this, but it isn't a production-ready feature yet).

The overall shape of the solution is likely going to be a separate request or a flag to allow the timestamps to be filed by user for some initial portion of cluster's livetime.

So I actually have two questions here:

* do you have a guarantee that all your existing timestamps are unique? 
* during migration, can you guarantee that all transfers will be submitted in order of increasing timestamps?

If the answer to both questions is yes, then I think we should be able to injest these data

As for the related question:

From the perspective of TigerBeetle's data model, there's only a single transaction timestamp --- that when the transaction was processed by the cluster. This timestamp is assigned by the cluster. So, in the described scenario, all txes will be timestamped with the later time, when they are executed.

You can however use one of the user_data fields to record time-of-submission for transactions. Though, for reporting purposes you probably want to use execution time (or our exisint historical account balances feature), as that's the timestamp where the effect of TX actually takes place.

batiati:
To complement matklad’s answer, we also started discussing the design of a similar feature in the issue below; please take a look and give your feedback!

[...]

Issue #1893, opened 2024-04-29
Tracking Issue: Backfilling of transfer data


jamalzkhan:
Hi TB team (adding this here for tracking)

If there is a TigerBeetle client 🐯 and it wants to add historical data for example from the past 2 years from the OLGP (keeping timestamp consistency), it would be useful to have a 'backfill' mode or short-term tool available where timestamp fields can be passed.

If it's a flag then maybe it can have a --backfill mode and when the cluster is restarted (without flag) then this functionality can no longer happen and traffic flows normally (without the ability for timestamp thrashing).

This would be useful for:

1) Clients who want to use TigerBeetle as a single source of truth for ledger (by backfilling their data)

2) Queries based on timestamp range (windows) to work seamlessly as opposed to using user_data_x which may not support range queries for sometime.

Thank you 🙏🏼

matklad:
In general, this might be a tad complicated, because timestamps in tigerbeetle are server-generated and server as the (internal) primary key. So just ramming in arbitrary timestamps won't work.

That said, we actually have something like this implemented! We support backfilling TigerBeetle cluster from a different tigerbeetle cluster, through AOF file.

This isn't very well documented yet, and I don't think we tried using that for importing data from different systems, but this _should_ be possible!

jamalzkhan:
Hi @matklad - Thanks for getting back on this.

I think the AOF file could work perfectly as this is a one-off task by clients usually.

If there is any documentation on this or steps when the team gets time it would be great to have so it can be tested?

A sample AOF file would also be ace to try and generate.

I saw this PR and but couldn't find the sample test.aof - https://github.com/tigerbeetle/tigerbeetle/pull/355

cb22:
Hey @jamalzkhan - a good starting point is the [test_aof.sh](https://github.com/tigerbeetle/tigerbeetle/blob/main/.github/ci/test_aof.sh) file which runs tests for it in CI.

Feel free to ask if you have any questions about what it's doing!

jamalzkhan:
Hi @cb22 - Thank you, tried to run the test_aof.sh but when I tried to open the AOF file to try and add my own data I can't find a reader / writer for it - I think it's probably an internal binary format.

Any pointers as to how I can construct an AOF file using some simple syntax like a CSV or JSON format and pipe this into some tool ?

batiati:
Follow up with what we discussed in our design meeting:

[...]

PR #1989, opened 2024-06-05
New operations import_{accounts,transfers}

batiati:
## New operations import_{accounts,transfers} 

Allows the ingestion of historical accounts and transfers with their original `timestamp` into TigerBeetle.

The new operations `import_accounts` and `import_transfers` behave exactly like the existing `create_accounts` and `create_transfers,` except for some validations regarding the `timestamp` field.

For that reason, **the same events and results are shared between the `create_*` and `import_*` operations**.
It was considered to create at least new result enums, such as `ImportAccountResult` and `ImportTransferResult`, but the benefits of handling unified error codes with shared enums outweighed the decision.
 
From the docs:

https://github.com/tigerbeetle/tigerbeetle/blob/f629c7b93b0a4dfe75a482451967b4b5b696b286/docs/reference/requests/import_transfers.md?plain=1#L10-L45

## This PR

**Best reviewed per commit.**

- State machine logic and unit tests.
- REPL.
- Clients (including README and examples).
- Documentation.

**TODO** 

- VOPR needs changes to support synchronizing clients, otherwise we will hit `import_timestamp_must_not_regress` every time.

Releated #1893 and #1968


docs/reference/account.md
```diff
@@ -226,7 +226,8 @@ You can read more about [Time in TigerBeetle](../develop/time.md).
 Constraints:
 
 - Type is 64-bit unsigned integer (8 bytes)
-- Must be set to `0` by the user when the `Account` is created
+- Must be set to `0` by the user when the `Account` is created or
+  with a user-defined unique timestamp when [importing accounts](./requests/import_accounts.md).
```
sentientwaffle: To clarify that these are two different cases, maybe split this into:

```
- Must be set to `0` for accounts in [`create_accounts`](./requests/create_accounts.md).
- Must be set to a user-defined unique timestamp for accounts in [`import_accounts`](./requests/import_accounts.md).
```

And maybe remove or modify the "It is set by TigerBeetle to the moment the account arrives at the cluster." since that is not true for import_accounts.

src/tigerbeetle.zig
```diff
@@ -257,6 +261,11 @@ pub const CreateTransferResult = enum(u32) {
     exceeds_credits = 54,
     exceeds_debits = 55,
 
+    import_timestamp_must_not_be_zero = 56,
+    import_regressed_timestamp = 57,
+    import_future_timestamp = 58,
+    import_timeout_must_be_zero = 59,
```
sentientwaffle: Can you change the order of these 4 to match their precedence relative to each other? Or is there a specific reason they are in this order?

docs/reference/requests/create_transfers.md
```diff
@@ -399,6 +401,36 @@ Otherwise, the [credit account](../transfer.md#credit_account_id) has
...
+The user-defined [`Transfer.timestamp`](../transfer.md#timestamp) is greater than the current [cluster time](../../coding/time.md),
+but it must be a past timestamp.
+
+### `import_timeout_must_be_zero`
+
+This result only applies to the [import_transfers](./import_transfers.md) operation.
+
+The transfer was not created.
+[`Transfer.timeout`](../transfer.md#timeout) is nonzero, but it must be zero.
+
```
sentientwaffle: Unfortunately, I think that we shouldn't leave these grouped together like this in the docs, due to

> Results are listed in this section in order of descending precedence — that is, if more than one error is applicable to the transfer being created, only the result listed first is returned.

(Likewise for `create_accounts`.)

docs/reference/account.md
```diff
@@ -219,14 +219,14 @@ accounts with this flag set.
...
 
-It is set by TigerBeetle to the moment the account arrives at the cluster.
-
 You can read more about [Time in TigerBeetle](../coding/time.md).
 
 Constraints:
 
 - Type is 64-bit unsigned integer (8 bytes)
-- Must be set to `0` by the user when the `Account` is created
+- Must be set to `0` for accounts in [`create_accounts`](./requests/create_accounts.md).
```
sentientwaffle:
```suggestion
- Must be set to `0` for accounts in [`create_accounts`](./requests/create_accounts.md).
  It will be set by TigerBeetle to the moment the account arrives at the cluster.
```

src/tigerbeetle.zig
sentientwaffle, line 1: I don't think we can(/should) renumber the existing result codes; it breaks compatibility.

For https://github.com/tigerbeetle/tigerbeetle/pull/1989#discussion_r1674649420 I just thought that the new codes should be _organized_ by priority.

docs/reference/requests/create_accounts.md
```diff
@@ -31,8 +31,25 @@ to the next operation).
 
 ### `timestamp_must_be_zero`
 
-The account was not created. The [`Account.timestamp`](../account.md#timestamp) is nonzero, but must
-be zero. The cluster is responsible for setting this field.
+The account was not created.
+The [`Account.timestamp`](../account.md#timestamp) is nonzero, but must be zero.
+The cluster is responsible for setting this field when
+[creating a new `Account`](create_accounts.md).
```
sentientwaffle, line 37: Can we copy the same language from the other errors, e.g.:

```
This result only applies to the [create_accounts](./create_accounts.md) operation.
```

src/scripts/client_readmes.zig
```diff
@@ -442,6 +442,19 @@ fn readme_root(ctx: *Context) !void {
         ctx.code_section("linked-events");
     }
 
+    {
+        ctx.header(2, "Importing historical events");
+        ctx.paragraph(
+            \\Allows the ingestion of historical accounts and transfers with their original
+            \\timestamp into TigerBeetle.
```
sentientwaffle, line 449:
```suggestion
            \\Ingest historical accounts or transfers with their original
            \\timestamp into TigerBeetle.
```

src/tigerbeetle.zig
```diff
@@ -148,29 +148,34 @@ pub const CreateAccountResult = enum(u32) {
     linked_event_chain_open = 2,
     timestamp_must_be_zero = 3,
 
-    reserved_field = 4,
-    reserved_flag = 5,
+    import_timestamp_must_not_be_zero = 4,
+    import_timestamp_must_not_be_in_the_future = 5,
```
batiati, line 152: After #2117, I wonder if we need a `import_timestamp_must_not_be_int_max`...

src/state_machine.zig
```diff
@@ -1466,15 +1481,31 @@ pub fn StateMachineType(
...
+                            if (event.timestamp != 0) break :blk .timestamp_must_be_zero;
+                            event.timestamp = timestamp - events.len + index + 1;
+                        },
+                        .import_accounts, .import_transfers => {
+                            if (event.timestamp == 0) {
+                                break :blk .import_timestamp_must_not_be_zero;
+                            }
+                            if (event.timestamp >= self.commit_timestamp) {
+                                break :blk .import_timestamp_must_not_be_in_the_future;
+                            }
```
sentientwaffle, line 1495: This is a determinism bug! `commit_timestamp` is usually the timestamp of the last commit, but if we are replaying the first op from the the WAL immediately after a restart it might be zero instead.

`timestamp` is consistent, though (i.e. the timestamp that is stamped on the `prepare` header) -- would that work instead?

docs/reference/requests/import_accounts.md
```diff
@@ -0,0 +1,61 @@
...
+
+  The user-defined timestamp must be **unique** across the entire database, and no two objects can
+  have the same timestamp, even different objects like an `Account` and a `Transfer` cannot share
+  the same timestamp.
+
+- No timestamp regression
+
+  Timestamps must always be increasing. Even user-defined timestamps that are required to be past
+  dates need to be at least one nanosecond ahead of the timestamp of the last object committed by
+  the cluster.
```
sentientwaffle, line 31: TODO

src/state_machine.zig
```diff
@@ -1527,6 +1558,55 @@ pub fn StateMachineType(
...
+                            key_range.key_max
+                        else
+                            0,
+                        if (self.forest.grooves.transfers.objects.key_range) |*key_range|
+                            key_range.key_max
+                        else
+                            0,
+                    );
+
+                    assert(event.timestamp < self.commit_timestamp);
```
sentientwaffle, line 1603: I think this isn't guaranteed; `commit_timestamp` might be `0` after a restart + replay.

src/state_machine.zig
```diff
@@ -1527,6 +1558,55 @@ pub fn StateMachineType(
...
+                    const timestamp_max = @max(
+                        if (self.forest.grooves.accounts.objects.key_range) |*key_range|
+                            key_range.key_max
+                        else
+                            0,
+                        if (self.forest.grooves.transfers.objects.key_range) |*key_range|
+                            key_range.key_max
+                        else
+                            0,
+                    );
```
sentientwaffle, line 1601: :thinking: The natural way to write a migration script is something like this:

```
for (account in old database):
    client.import_accounts([account])
for (transfer in old database):
    client.import_transfers([transfer])
```

But this doesn't work b/c accounts/transfers have interleaved timestamps.
The interleaved version would be more complicated to write. (And made even more complicated by automatic client batching.)

Interleaving is also less efficient, since you can't maximize batching -- that is, batches will often end prematurely to allow for interleaving, since a single batch can't create both accounts and transfers.

Could we treat the `timestamp_max`'s as independent by object type?

src/state_machine.zig
```diff
@@ -1789,6 +1883,20 @@ pub fn StateMachineType(
...
+
+            // Imported transfers with past timestamp cannot set a timeout for expiration.
+            if (operation == .import_transfers and t.timeout != 0) {
+                assert(t.flags.pending);
+                return .import_timeout_must_be_zero;
+            }
+
+            assert(t.timestamp > dr_account.timestamp);
+            assert(t.timestamp > cr_account.timestamp);
+
```
sentientwaffle, line 1899: I think these asserts can go above the `if (operation == .import_transfers and t.timeout != 0) {` condition?

src/state_machine.zig
```diff
@@ -1861,45 +1969,53 @@ pub fn StateMachineType(
             if (dr_account.debits_exceed_credits(amount)) return .exceeds_credits;
             if (cr_account.credits_exceed_debits(amount)) return .exceeds_debits;
 
-            // After this point, the transfer must succeed.
-            defer assert(self.commit_timestamp == t.timestamp);
+            // After this point, the transfer must succeed atomically.
+            var result: CreateTransferResult = undefined;
+            defer assert(result == .ok);
+            defer if (operation == .create_transfers) assert(self.commit_timestamp == t.timestamp);
+            result = completed_atomically: {
```
sentientwaffle, line 1976: What is the reason for this refactor?

src/state_machine.zig
```diff
@@ -3254,7 +3388,11 @@ fn check(test_table: []const u8) !void {
                 try reply.appendSlice(std.mem.asBytes(&transfers.get(id).?));
             },
             .commit => |commit_operation| {
-                assert(operation == null or operation.? == commit_operation);
+                assert(operation == null or switch (commit_operation) {
+                    .create_accounts, .import_accounts => operation.? == .create_accounts,
+                    .create_transfers, .import_transfers => operation.? == .create_transfers,
+                    else => operation.? == commit_operation,
```
sentientwaffle, line 3394: A little simpler:
```suggestion
                    .import_accounts => operation.? == .create_accounts,
                    .import_transfers => operation.? == .create_transfers,
                    else => operation.? == commit_operation,
```

src/state_machine.zig
batiati: > Could we treat the timestamp_max's as independent by object type?

I don't think we should do that, unless we want to relax the invariants  `t.timestamp > dr_account.timestamp and t.timestamp > cr_account.timestamp`.

src/state_machine.zig
batiati:
```diff
- defer assert(self.commit_timestamp == t.timestamp);
```

We don't update `self.commit_timestamp` for import_* operations. Since we are dealing with past timestamps, the commit_timestamp is kept, like in lookups and queries.

src/tigerbeetle.zig
batiati: :man_facepalming:

docs/reference/requests/import_accounts.md
sentientwaffle: Ah this TODO was intended for me, I wanted to re-read this before posting my review but I forgot. :man_facepalming:

src/state_machine.zig
sentientwaffle: :thinking: I don't understand how that is connected to the `completed_atomically` blocks.

Also, I think that due to https://github.com/tigerbeetle/tigerbeetle/pull/1989#discussion_r1683206981 you may as well update `commit_timestamp` anyway.

src/state_machine.zig
sentientwaffle: Ah, that's a good point.

Could we add a new error code for `import_transfers` to protect against that case? It might be worth it to make migration scripts simpler (unless I'm misunderstanding how they would be implemented?).

batiati:
Closed in favor or #2171

PR #1990, opened 2024-06-05
Grid/Scrubber: Model probability of data loss in terms of cycle interval

sentientwaffle:
(Follow-up for https://github.com/tigerbeetle/tigerbeetle/pull/1851.)

Model the probability that the cluster experiences eventual data loss due to bitrot.

More specifically, the probability that *every* copy of *any* grid block is corrupted before the scrubber can repair.

The probability of every single copy of the same block being corrupted across all replicas is quite small, even without scrubbing. This model is mostly just to "prove" that, and show how scrubbing shrinks that probability even farther.

Aside: Though this calculation uses `f64` for ergonomics/clarity, my [original implementation](https://gist.github.com/sentientwaffle/c90df9e9ff49ef5f970e66d32c7e071c) used arbitrary-precision numbers, and the loss of precision due to floats is not significant.

---

At the same time, bump the default scrubber cycle interval from 90 days to 180 days... which is still higher than strictly needed.


src/vsr/grid_scrubber.zig
```diff
@@ -636,3 +636,133 @@ fn ManifestBlockIteratorType(comptime ManifestLog: type) type {
...
+    const storage_blocks = @divExact(storage_size, block_size);
+    const test_duration_days = test_duration_years * 365;
+    const test_duration_cycles = stdx.div_ceil(test_duration_days, cycle_interval_days);
+    const sector_faults_per_cycle =
+        stdx.div_ceil(sector_faults_per_year * cycle_interval_days, 365);
+
+    // P(a specific sector is uncorrupted for an entire cycle)
+    //
+    // If we inject a fault randomly into S sectors, then the probability that the fault misses one
+    // sector in specific is `(S-1)/S`. If we then repeat this N times, then the chance that it
```
matklad: "one sector in specific" doesn't parse grammatically for me.

src/vsr/grid_scrubber.zig
```diff
@@ -636,3 +636,133 @@ fn ManifestBlockIteratorType(comptime ManifestLog: type) type {
...
+    const p_sector_healthy_per_cycle = std.math.pow(
+        f64,
+        @as(f64, @floatFromInt(storage_sectors - 1)) / @as(f64, @floatFromInt(storage_sectors)),
+        @as(f64, @floatFromInt(sector_faults_per_cycle)),
+    );
+
+    // P(a specific block is uncorrupted for an entire cycle)
+    // If any of the block's sectors is corrupted, then the whole block is corrupted.
+    const p_block_healthy_per_cycle =
+        std.math.pow(f64, p_sector_healthy_per_cycle, @as(f64, @floatFromInt(block_sectors)));
```
matklad: I think this is not pedantically correct. A _different_ way to compute this is to say that a probability that a single fault misses the block is `(S - block_sectors) / S`, and hence the probability that the all faults miss the block is  `((S - block_sectors) / S) ^ sector_faults_per_cycle`, which is a different number, so there must be a difference here!

I think the bug is in the formula in the source code, as you can't multiply probabilities for different sectors, as they are not independent --- if one sector is know to not be faulty then other sectors are more likely to be faulty, assuming we have a constant error rate. 

But it another way, if we expect ten faults per cycle, and our block size is equal to the data size, then we know that the block is correpted, and that the probability of it being healthy is zero. However, probability of every particular sector to be healthy is non-zero, so, if we multiply them all, we'll get something bigger than zero. 

Now, I am not sure if this is actually the problem in the formula, or the problem in how we state the error rates. I guess, the problem is that we say that the faults are uncorrelated, but then use a model where a bag of faults is distributed across the disk, which makes them correlated!

src/constants.zig
```diff
@@ -366,8 +366,8 @@ pub const grid_scrubber_reads_max = config.process.grid_scrubber_reads_max;
 /// (assuming a fully-loaded data file – maximum size and 100% acquired):
 ///
 ///   storage_size_limit_max      = 16TiB
-///   grid_scrubber_cycle_seconds = 90 days * 24 hr/day * 60 min/hr * 60 s/min (1 cycle/year)
-///   read_bytes_per_second       = storage_size_max / grid_scrubber_cycle_seconds ≈ 2.16 MiB/s
+///   grid_scrubber_cycle_seconds = 180 days * 24 hr/day * 60 min/hr * 60 s/min (1 cycle/year)
```
matklad: 2 cycles per year (and it used to be four?)

matklad (approved):

PR #1991, opened 2024-06-06
IO: Add async openat()

cb22:
Adds an async `openat()` implementation, for Linux's IO implementation. Will be used by multiversioning: we'll be doing `open()`s on a regular basis internally while serving requests, so we can't afford for `open()` to block.

Windows and macOS are not production targets; so it's OK if for now they still use a standard `open()` or equivalent.

### DIFF
--- a/src/io/test.zig
+++ b/src/io/test.zig
@@ -13,8 +13,8 @@ test "write/read/close" {
 
         io: IO,
         done: bool = false,
-        fd: os.fd_t,
 
+        fd: ?os.fd_t = null,
         write_buf: [20]u8 = [_]u8{97} ** 20,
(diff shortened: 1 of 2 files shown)


src/io/test.zig
```diff
@@ -23,31 +23,67 @@ test "write/read/close" {
...
+
+                self.io.write(
+                    *Context,
+                    &self,
+                    write_callback,
+                    &completion,
+                    self.fd.?,
+                    &self.write_buf,
+                    10,
+                );
```
kprotty: Would it be nicer to do
```suggestion
                self.openat_callback(&completion, file.handle)
```

cb22: Much!

kprotty (approved):

PR #1993, opened 2024-06-06
docs: note that lookup_accounts shouldn't be used before creating transfers

emschwartz:


docs/reference/requests/lookup_accounts.md
```diff
@@ -2,6 +2,12 @@
 
 Fetch one or more accounts by their `id`s.
 
+⚠️ Note that you **should not** check an account's balance using this request before creating a
+transfer. That would not be atomic and the balance could change in between the check and the
+transfer. Instead, you should use a balance limit on the account or a
+[balance-conditional transfer](../../develop/recipes/balance-conditional-transfers.md) to ensure the
+balance check is done atomically with the transfer.
```
matklad: Hm, let's phrase this in a more generalized way, something along the lines of:

> Instead, set the `debits_not_exceed_credits` flag on the corresponding accounts. More complex conditional transfres can be expressed using [/balance-conditional-transfers.md]

That is, the current phrasing is very specific to one problem, and, if you don't have that problem, you might miss the general principle. 

So we should:

* point the general principle (don't move data to code)
* point the basic mechanism through which the principle is achieved (debits_don't_exceed_credits flag on an account)
* point to the more general mechnism (linked transfers)

matklad (approved):

PR #1997, opened 2024-06-10
Replace non-mutated `var`s with `const`

kprotty:
In preparation for the Zig 0.13 upgrade as the change is 1) compatible with 0.11 and 2) allows more careful review of `var`s which *should've* been mutated.


matklad:
LGTM!

It's interesting that a bunch of these are for pointers: that is, we tend to write

```zig
var foo: *Foo = allocator.create(Foo)
```

when we intend to write `Foo`, but this needs to be `const`, because we are not going to mutate the pointer itself!

matklad (approved):

PR #1999, opened 2024-06-11
IO: Add async statx()

cb22:


src/io/test.zig
```diff
@@ -99,9 +106,34 @@ test "write/read/close" {
...
+                    statx_callback,
+                    completion,
+                    os.AT.FDCWD,
+                    self.path,
+                    0,
+                    os.linux.STATX_BASIC_STATS,
+                    &self.statx,
+                );
+            } else {
+                self.done = true;
```
kprotty, line 124: Maybe same trick again?
```suggestion
                self.statx_callback(completion, {});
```

cb22: Ah yes

cb22: Ah no - Darwin / Win don't have the IO.StatxError type.

kprotty (approved):

PR #2000, opened 2024-06-11
Fix incorrect early return

batiati:
The transfer was partially inserted before the validation `pending_transfer_expired`, leaving the system inconsistent with multiple transfers for the same pending_id.

Unit tests do not catch this since the test executor always runs the expiry logic between commits. In the real world, due to the pulse mechanics, a request can fail with `pending_transfer_expired` before the expiry pulse.

### DIFF
--- a/src/state_machine.zig
+++ b/src/state_machine.zig
@@ -1546,6 +1546,9 @@ pub fn StateMachineType(
             if (dr_account.debits_exceed_credits(amount)) return .exceeds_credits;
             if (cr_account.credits_exceed_debits(amount)) return .exceeds_debits;
 
+            // After this point, the transfer must succeed.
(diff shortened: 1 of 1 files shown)


src/state_machine.zig
```diff
@@ -1546,6 +1546,9 @@ pub fn StateMachineType(
             if (dr_account.debits_exceed_credits(amount)) return .exceeds_credits;
             if (cr_account.credits_exceed_debits(amount)) return .exceeds_debits;
 
+            // After this point, the transfer must succeed.
+            defer assert(self.commit_timestamp == t.timestamp);
```
matklad, line 1550: :100:

matklad (approved):
Oh, this was a nasty one, great catch!

PR #2003, opened 2024-06-11
CLI: Add data file inspector tool

sentientwaffle:
~Add a new `zig build scripts -- inspect ...` subcommand for inspecting a data file.~

Add a new `tigerbeetle inspect ...` subcommand for inspecting a data file.

Principles:
- `inspect` never modifies the data file.
- When the data file is corrupt, decode as much as possible. (This is somewhat aspirational).
- Outside of the "summary" commands, don't discard potentially useful information.

GitHub PR descriptions have a 65536 character limit (TIL!), so please check out this [gist](https://gist.github.com/sentientwaffle/b4615156c4c7000205b434a941d26d48) instead, for samples of all of the subcommands.


cb22:
@sentientwaffle what are your thoughts on promoting it to a full subcommand? eg, `tigerbeetle inspect superblock 0_0.tigerbeetle`

sentientwaffle:
I have no objections -- that would be more convenient than going through `scripts`!
I just wasn't sure whether it warranted increasing the size of our binary.

It would be super useful, though, when helping developers troubleshoot their local issues... (rather than trying to figure out how to share multi-GiB data files! :sweat_smile:)

matklad:
> what are your thoughts on promoting it to a full subcommand? eg, tigerbeetle inspect superblock 0_0.tigerbeetle

Wanted to make this exact comment! I’d say this would be incredibly useful as a build-in command, as I’d rather work on reducing binary size elsewhere :-)

jorangreef:
> Wanted to make this exact comment! I’d say this would be incredibly useful as a build-in command, as I’d rather work on reducing binary size elsewhere :-)

And I wanted to say the same!

In my mind I think of `fsck`. TB is now officially the filesystem it always wanted to be.

src/tigerbeetle/inspect.zig
```diff
@@ -0,0 +1,1100 @@
...
+const Grid = vsr.GridType(Storage);
+const BlockPtr = vsr.grid.BlockPtr;
+const BlockPtrConst = vsr.grid.BlockPtrConst;
+const allocate_block = vsr.grid.allocate_block;
+const is_composite_key = vsr.lsm.composite_key.is_composite_key;
+
+pub fn main(allocator: std.mem.Allocator, cli_args: *const cli.Command.Inspect) !void {
+    var stdout_buffer = std.io.bufferedWriter(std.io.getStdOut().writer());
+
+    const stdout = stdout_buffer.writer();
```
matklad: You could do `stdout.any()` and pass `AnyWriter`. That is a dynamically-dispatched type, like `mem.Allocator`, and I think that's a better default patter for writers, as it doesn't require monomorphisation, like anytype. 

That's a new zig 0.12 thing.

src/tigerbeetle/inspect.zig
```diff
@@ -0,0 +1,1100 @@
...
+        const dirname = std.fs.path.dirname(path) orelse ".";
+        inspector.dir_fd = try vsr.io.IO.open_dir(dirname);
+        errdefer std.posix.close(inspector.dir_fd);
+
+        const basename = std.fs.path.basename(path);
+        inspector.fd = try vsr.io.IO.open_file(
+            inspector.dir_fd,
+            basename,
+            vsr.superblock.data_file_size_min,
+            .open,
```
matklad: This opens the file in read-write mode. I feel strongly that, for defense in depth, we should open in read-only mode, such that we dont' have a "radioactive" file descriptor in the address space.

src/tigerbeetle/cli.zig
```diff
@@ -109,6 +109,102 @@ const CliArgs = union(enum) {
...
+        manifest: struct {
+            superblock_copy: u8 = 0,
+            positional: struct { path: []const u8 },
+        },
+        tables: struct {
+            superblock_copy: u8 = 0,
+            tree: []const u8,
+            level: ?u6 = null,
+            positional: struct { path: []const u8 },
+        },
```
matklad: Nit, but I'd probably go for `inspect --superblock <path>` rather than subcommands. It's unclear how exactly this'll will evolve, and a single command with a bunch of flags is more flexible.

Eg, I can imagine that at some point we'll want to add a `--format=json` flag, and adding that too all subcommands would be a chore. 

It also feels like

```
$ insperct --superblock --wal <datafile>
```

could be a thing.

src/tigerbeetle/cli.zig
```diff
@@ -109,6 +109,102 @@ const CliArgs = union(enum) {
...
+            \\  tigerbeetle inspect grid [--block=<address>] <path>
+            \\
+            \\  tigerbeetle inspect manifest <path>
+            \\
+            \\  tigerbeetle inspect tables --tree=<name|id> [--level=<integer>] <path>
+            \\
+            \\Options:
+            \\
+            \\  When `--superblock-copy` is set, use the trailer referenced by that superblock copy.
+            \\  Otherwise, copy=0 will be used by default.
```
matklad: Hm, I'd expect that, by default, we read all copies and do the quorum

src/tigerbeetle/cli.zig
```diff
@@ -109,6 +109,102 @@ const CliArgs = union(enum) {
...
+            \\        In the left column of the output, "|" denotes which copies have a particular value.
+            \\        "||||" means that all four superblock copies are in agreement.
+            \\        "|_|_" means that the value matches in copies 0/2, but differs from copies 1/3.
+            \\
+            \\  wal
+            \\        Inspect the WAL headers and prepares.
+            \\        In the left column of the output, "|" denotes which set of headers has each value.
+            \\        "||" denotes that the prepare and the redundant header match.
+            \\        "|_" is the redundant header.
+            \\        "_|" is the prepare's header.
```
matklad: Perhaps better to move the legend to the output itself?

src/tigerbeetle/inspect.zig
```diff
@@ -0,0 +1,1100 @@
+//! Decode a TigerBeetle data file without running a replica or modifying the data file.
+//! This tool adheres to the "be liberal in what you accept" side of Postel's Law.
```
matklad: I'd maybe expand a bit on the why of the tool here. In parcicular, worth mentionning wherether the intended audience are the operators, or the developers of TigerBeetle.

In general, I feel like we could on average a couple of pararaphs of prose at the start of each file, as that's a very good way to "orient"  the reader in the codebase!

src/tigerbeetle/inspect.zig
```diff
@@ -0,0 +1,1100 @@
...
+                }
+                try inspector.inspect_replies_slot(stdout, args.superblock_copy, slot);
+            } else {
+                try inspector.inspect_replies(stdout, args.superblock_copy);
+            }
+        },
+        .grid => |args| {
+            if (args.superblock_copy >= constants.superblock_copies) {
+                return fatal(
+                    "--superblock-copy: copy exceeds {}\n",
```
matklad: We _shouldn't_ need a `\n` here?

src/tigerbeetle/inspect.zig
```diff
@@ -0,0 +1,1100 @@
...
+            if (args.block) |address| {
+                try inspector.inspect_grid_block(stdout, address);
+            } else {
+                try inspector.inspect_grid(stdout, args.superblock_copy);
+            }
+        },
+        .manifest => |args| {
+            if (args.superblock_copy >= constants.superblock_copies) {
+                return fatal(
+                    "--superblock-copy: copy exceeds {}\n",
```
matklad: ditto

src/tigerbeetle/inspect.zig
```diff
@@ -0,0 +1,1100 @@
...
+                    .{constants.superblock_copies - 1},
+                );
+            }
+
+            try inspector.inspect_manifest(stdout, args.superblock_copy);
+        },
+        .tables => |args| {
+            const tree_id = parse_tree_id(args.tree) orelse {
+                return fatal("--tree: invalid tree name/id: {s}\n", .{args.tree});
```
matklad: ditto

src/tigerbeetle/inspect.zig
```diff
@@ -0,0 +1,1100 @@
...
+    allocator: std.mem.Allocator,
+    dir_fd: std.posix.fd_t,
+    fd: std.posix.fd_t,
+    io: vsr.io.IO,
+    storage: Storage,
+
+    busy: bool = false,
+    read: Storage.Read = undefined,
+
+    fn init(allocator: std.mem.Allocator, path: []const u8) !*Inspector {
```
matklad: As this returns a pointer, this could use `create/destroy` rather than `init/deinit`. Not sure if this naming convention is wortth pursuing though!

src/tigerbeetle/inspect.zig
```diff
@@ -0,0 +1,1100 @@
...
+
+        inspector.storage = try Storage.init(&inspector.io, inspector.fd);
+        errdefer inspector.storage.deinit();
+
+        return inspector;
+    }
+
+    fn deinit(inspector: *Inspector) void {
+        inspector.storage.deinit();
+        std.posix.close(inspector.fd);
```
matklad, line 189: `inspector.io.deinit();` missing

src/tigerbeetle/inspect.zig
```diff
@@ -0,0 +1,1100 @@
...
+        std.posix.close(inspector.dir_fd);
+        inspector.allocator.destroy(inspector);
+    }
```
matklad, line 193: `inspector.* = undefined`

matklad (commented):
Let's add a smoke test to our integration_tests here. I think we can tack it onto the end of `benchmark smoke` test, to minimize the runtime.

What we _should_ test is that the functionality works (so, just non-zero exit status).

What we _could_ also test is specific output format, just so that it's easier to review in the future if we change it. But i am not sure how easy would that be. 

There's some support for redacitons in our snaptests, but I am not sure if that'l be enough:

```
try Snap.snap(@src(),
    \\it's <snap:ignore>ms
).diff(time);
```

src/tigerbeetle/inspect.zig
matklad: IN particular, the "principles" section 

```
Principles:

inspect never modifies the data file.
When the data file is corrupt, decode as much as possible. (This is somewhat aspirational).
Outside of the "summary" commands, don't discard potentially useful information.
```

definitely would be helpful for anyone trying to make sense of this code in the future !

src/tigerbeetle/inspect.zig
```diff
@@ -0,0 +1,1100 @@
...
+                );
+            }
+
+            try inspector.inspect_manifest(stdout, args.superblock_copy);
+        },
+        .tables => |args| {
+            const tree_id = parse_tree_id(args.tree) orelse {
+                return fatal("--tree: invalid tree name/id: {s}\n", .{args.tree});
+            };
+            try inspector.inspect_tables(stdout, args.superblock_copy, .{
```
matklad, line 108: Missing range check for `args.superblock_copy` for this case

src/tigerbeetle/inspect.zig
```diff
@@ -0,0 +1,1100 @@
...
+        }
+        return buffer;
+    }
+
+    const ClientSessions = extern struct {
+        headers: [constants.clients_max]vsr.Header.Reply,
+        sessions: [constants.clients_max]u64,
+    };
+
+    fn read_client_sessions(inspector: *Inspector, superblock_copy: u8) !ClientSessions {
```
matklad: Feels like this could stack overflow? SizeOf ClientSessions is non-trivial! I guess we could allocate an instance of it as a part of overall Inspector? Though, if it doesn't actually crash, then no need to fix it.

Or, we could heap allocate it. I actually forgot that that's an option :rofl:

src/tigerbeetle/inspect.zig
```diff
@@ -0,0 +1,1100 @@
...
+                "read_block: invalid block address={} checksum={?x:0>32} (bad checksum)",
+                .{ address, checksum },
+            );
+            return error.InvalidChecksum;
+        }
+
+        if (!header.valid_checksum_body(buffer[@sizeOf(vsr.Header)..header.size])) {
+            log.err(
+                "read_block: invalid block address={} checksum={?x:0>32} (bad checksum_body)",
+                .{ address, checksum },
```
matklad: Feels worthwhile to print the checksum the block actually has as well, if only to make the output unambigious.

src/tigerbeetle/cli.zig
sentientwaffle: :thinking: I like subcommands b/c each subcommand can have an independent set of flags for validation. e.g. `wal` has `--slot`, but `superblock` doesn't.

For something like `--format=json`, we could handle that the same way as `path`:
```
const path = switch (command) {
    inline else => |args| args.positional.path,
};
```

src/tigerbeetle/inspect.zig
sentientwaffle: Oops; those were converted from `writeAll` and I forgot to remove the newline.

sentientwaffle:
> Let's add a smoke test to our integration_tests here.

:+1: 

> I think we can tack it onto the end of benchmark smoke test, to minimize the runtime.

It'll be a separate test, since `tigerbeetle benchmark` auto-deletes its file.

matklad:
It might be useful to teach the benchmark to not delete file!

src/tigerbeetle/inspect.zig
```diff
@@ -25,7 +25,7 @@ const is_composite_key = vsr.lsm.composite_key.is_composite_key;
 pub fn main(allocator: std.mem.Allocator, cli_args: *const cli.Command.Inspect) !void {
     var stdout_buffer = std.io.bufferedWriter(std.io.getStdOut().writer());
 
-    const stdout = stdout_buffer.writer();
+    const stdout = stdout_buffer.writer().any();
```
matklad: :thinking: does this actually work as intended? I think we'll need one extra temporary here: `.writer()` stores the pointer to `stdout_buffer`, but `.any()` stores the pointer to `.writer()` (so, two start pointer). But `.writer` is a temporary, and it should die after the expression is  evaluated.

sentientwaffle: Weirdly it does work, but I will update it b/c I agree that it shouldn't...

matklad (approved):
:rocket:

PR #2004, opened 2024-06-12
Simulator coverage for get_account_{transfers,balances}

batiati:
Implemented assertions based on the expected number of results in variations of filters such as `debits | credits` and timestamp range.

It's not able to fine-check intermediate states such as the expected `account_balance` between transfers, although the path covered is good enough to check the assertions already in place along the code.

### DIFF
--- a/src/state_machine/workload.zig
+++ b/src/state_machine/workload.zig
@@ -171,6 +171,8 @@ pub fn WorkloadType(comptime AccountingStateMachine: type) type {
         create_transfers = @intFromEnum(Operation.create_transfers),
         lookup_accounts = @intFromEnum(Operation.lookup_accounts),
         lookup_transfers = @intFromEnum(Operation.lookup_transfers),
+        get_account_transfers = @intFromEnum(Operation.get_account_transfers),
+        get_account_balances = @intFromEnum(Operation.get_account_balances),
     };
 
     return struct {
@@ -212,6 +214,7 @@ pub fn WorkloadType(comptime AccountingStateMachine: type) type {
             assert(options.lookup_account_invalid_probability <= 100);
 
             assert(options.account_limit_probability <= 100);
+            assert(options.account_history_probability <= 100);
             assert(options.linked_valid_probability <= 100);
             assert(options.linked_invalid_probability <= 100);
 
@@ -243,6 +246,8 @@ pub fn WorkloadType(comptime AccountingStateMachine: type) type {
                     account.flags.debits_must_not_exceed_credits = b;
                     account.flags.credits_must_not_exceed_debits = !b;
                 }
+
+                account.flags.history = chance(random, options.account_history_probability);
             }
 
             return Self{
@@ -288,6 +293,8 @@ pub fn WorkloadType(comptime AccountingStateMachine: type) type {
                     .create_transfers => Action.create_transfers,
                     .lookup_accounts => Action.lookup_accounts,
                     .lookup_transfers => Action.lookup_transfers,
+                    .get_account_transfers => Action.get_account_transfers,
+                    .get_account_balances => Action.get_account_balances,
                 };
             };
 
@@ -300,6 +307,8 @@ pub fn WorkloadType(comptime AccountingStateMachine: type) type {
                     self.build_lookup_accounts(self.batch(u128, action, body)),
                 .lookup_transfers => @sizeOf(u128) *
                     self.build_lookup_transfers(self.batch(u128, action, body)),
+                .get_account_transfers, .get_account_balances => @sizeOf(tb.AccountFilter) *
+                    self.build_get_account_filter(self.batch(tb.AccountFilter, action, body)),
             };
             assert(size <= body.len);
 
@@ -347,8 +356,18 @@ pub fn WorkloadType(comptime AccountingStateMachine: type) type {
                     std.mem.bytesAsSlice(u128, request_body),
                     std.mem.bytesAsSlice(tb.Transfer, reply_body),
                 ),
-                //TODO: implement query.
-                .get_account_transfers, .get_account_balances => unreachable,
+                .get_account_transfers => self.on_get_account_transfers(
+                    client_index,
+                    timestamp,
+                    std.mem.bytesAsSlice(tb.AccountFilter, request_body),
+                    std.mem.bytesAsSlice(tb.Transfer, reply_body),
+                ),
+                .get_account_balances => self.on_get_account_balances(
+                    client_index,
+                    timestamp,
+                    std.mem.bytesAsSlice(tb.AccountFilter, request_body),
+                    std.mem.bytesAsSlice(tb.AccountBalance, reply_body),
+                ),
                 //Not handled by the client.
                 .pulse => unreachable,
             }
@@ -382,7 +401,7 @@ pub fn WorkloadType(comptime AccountingStateMachine: type) type {
                     account.ledger = 0;
                     results[i].insert(.ledger_must_not_be_zero);
                 } else {
-                    if (!self.auditor.accounts_created[account_index]) {
+                    if (!self.auditor.accounts_state[account_index].created) {
                         results[i].insert(.ok);
                     }
                     // Even if the account doesn't exist yet, we may race another request.
@@ -503,6 +522,73 @@ pub fn WorkloadType(comptime AccountingStateMachine: type) type {
             return lookup_ids.len;
         }
 
+        fn build_get_account_filter(self: *const Self, body: []tb.AccountFilter) usize {
+            assert(body.len == 1);
+            const account_filter = &body[0];
+            account_filter.* = tb.AccountFilter{
+                .account_id = 0,
+                .limit = 0,
+                .flags = .{
+                    .credits = false,
+                    .debits = false,
+                    .reversed = false,
+                },
+                .timestamp_min = 0,
+                .timestamp_max = 0,
+            };
+
+            account_filter.account_id = if (self.auditor.pick_account(.{
+                .created = null,
+                .debits_must_not_exceed_credits = null,
+                .credits_must_not_exceed_debits = null,
+            })) |account| account.id else
+            // Pick an account with valid index (rather than "random.int(u128)") because the
+            // Auditor must decode the id to check for a matching account.
+            self.auditor.account_index_to_id(self.random.int(usize));
+
+            switch (self.random.enumValue(enum { none, debits, credits, all })) {
+                .none => {}, // Testing invalid flags.
+                .debits => account_filter.flags.debits = true,
+                .credits => account_filter.flags.credits = true,
+                .all => {
+                    account_filter.flags.debits = true;
+                    account_filter.flags.credits = true;
+                },
+            }
+            account_filter.flags.reversed = self.random.boolean();
+
+            const batch_size = AccountingStateMachine.constants.batch_max.get_account_transfers;
+            account_filter.limit = switch (self.random.enumValue(enum { none, one, batch, max })) {
+                .none => 0, // Testing invalid limit.
+                .one => 1,
+                .batch => batch_size,
+                .max => std.math.maxInt(u32),
+            };
+
+            // The filter by timestamp range is restrictive and applied only when both the debit and
+            // credit sides are included, so we can assert the result count minus one.
+            if (account_filter.flags.debits and account_filter.flags.credits and
+                account_filter.limit >= batch_size and
+                chance(self.random, self.options.account_filter_timestamp_range_probability))
+            {
+                if (self.auditor.get_account_state(
+                    account_filter.account_id,
+                )) |account_state| {
+                    // Exclude the first or the last result depending on the sort order.
+                    if (account_state.transfer_timestamp_last -
+                        account_state.transfer_timestamp_first > 1)
+                    {
+                        account_filter.timestamp_min = account_state.transfer_timestamp_first +
+                            @intFromBool(!account_filter.flags.reversed);
+                        account_filter.timestamp_max = account_state.transfer_timestamp_last -
+                            @intFromBool(account_filter.flags.reversed);
+                    }
+                }
+            }
+
+            return 1;
+        }
+
         /// The transfer built is guaranteed to match the TransferPlan's outcome.
         /// The transfer built is _not_ guaranteed to match the TransferPlan's method.
         ///
@@ -541,12 +627,14 @@ pub fn WorkloadType(comptime AccountingStateMachine: type) type {
 
             const limit_debits = transfer_plan.limit and self.random.boolean();
             const limit_credits = transfer_plan.limit and (self.random.boolean() or !limit_debits);
+            assert(transfer_plan.limit == (limit_debits or limit_credits));
 
             const debit_account = self.auditor.pick_account(.{
                 .created = true,
                 .debits_must_not_exceed_credits = limit_debits,
                 .credits_must_not_exceed_debits = null,
             }) orelse return null;
+            assert(!limit_debits or debit_account.flags.debits_must_not_exceed_credits);
 
             const credit_account = self.auditor.pick_account(.{
                 .created = true,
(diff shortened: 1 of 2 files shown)


src/state_machine/workload.zig
```diff
@@ -503,6 +522,78 @@ pub fn WorkloadType(comptime AccountingStateMachine: type) type {
...
+                    0;
+                break :transfer_count dr_transfer_count + cr_transfer_count;
+            } else 0;
+
+            switch (self.random.enumValue(enum { none, one, all, max })) {
+                .none => account_filter.limit = 0, // Testing invalid limit.
+                .one => account_filter.limit = 1,
+                .all => account_filter.limit = transfer_count,
+                .max => account_filter.limit = batch_size,
+            }
```
matklad: maybe `account_filter.limit = switch()`? And maybe add `math.intMax(u32)` case just in case?

src/state_machine/workload.zig
```diff
@@ -503,6 +522,78 @@ pub fn WorkloadType(comptime AccountingStateMachine: type) type {
...
+                const dr_transfer_count: u32 = if (account_filter.flags.debits)
+                    account_state.dr_transfer_count
+                else
+                    0;
+                const cr_transfer_count: u32 = if (account_filter.flags.credits)
+                    account_state.cr_transfer_count
+                else
+                    0;
+                break :transfer_count dr_transfer_count + cr_transfer_count;
+            } else 0;
```
matklad: Might be more readable in imperative form:

```zig
            var transfer_count: u32 = 0;
            if (self.auditor.get_account_state(account_filter.account_id)) |account_state| {
                if (account_filter.flags.debits) {
                    transfer_count += account_state.dr_transfer_count;
                }
                if (account_filter.flags.credits) {
                    transfer_count += account_state.cr_transfer_count;
                }
            }
```

src/state_machine/workload.zig
```diff
@@ -503,6 +522,78 @@ pub fn WorkloadType(comptime AccountingStateMachine: type) type {
...
+
+            switch (self.random.enumValue(enum { none, one, all, max })) {
+                .none => account_filter.limit = 0, // Testing invalid limit.
+                .one => account_filter.limit = 1,
+                .all => account_filter.limit = transfer_count,
+                .max => account_filter.limit = batch_size,
+            }
+
+            // Apply filter by timestamp only when we the prefix or the suffix of the expected
```
matklad: "when we the prefix" doesn't parse for me

src/state_machine/workload.zig
```diff
@@ -503,6 +522,78 @@ pub fn WorkloadType(comptime AccountingStateMachine: type) type {
             return lookup_ids.len;
         }
 
+        fn build_get_account_filter(self: *const Self, body: []tb.AccountFilter) usize {
+            assert(body.len == 1);
+            const account_filter = &body[0];
+            account_filter.* = std.mem.zeroInit(tb.AccountFilter, .{});
```
matklad: zeroInit is a bit of a footgun, as it's not actually zeroInit, but rather default init. Using `mem.zero` is somewhat more reliable. 

Though, in this case I think the best is to go with explicit init:

```zig
account_filter.* = .{ /* code here */ }
```

to make sure we didn't forget any fields

src/state_machine/workload.zig
```diff
@@ -503,6 +522,78 @@ pub fn WorkloadType(comptime AccountingStateMachine: type) type {
...
+
+            // Apply filter by timestamp only when we the prefix or the suffix of the expected
+            // result can be asserted. The filter range is always restrictive.
+            if (account_filter.flags.debits and account_filter.flags.credits and
+                account_filter.limit > 1 and transfer_count < batch_size and
+                chance(self.random, self.options.account_filter_timestamp_range_probability))
+            {
+                if (self.auditor.get_account_state(
+                    account_filter.account_id,
+                )) |account_state| {
```
matklad: In this branch, we _should_ have account state, right? THen maybe `.?` instead of an if here, to make it fail loudly in case of some future unrelated change?

src/state_machine/workload.zig
```diff
@@ -503,6 +522,78 @@ pub fn WorkloadType(comptime AccountingStateMachine: type) type {
...
+                    // We want to exclude at least one result:
```
matklad: Hm, the code looks like we are trying to _keep_ at least one result?

src/state_machine/workload.zig
```diff
@@ -503,6 +522,78 @@ pub fn WorkloadType(comptime AccountingStateMachine: type) type {
...
+        fn build_get_account_filter(self: *const Self, body: []tb.AccountFilter) usize {
+            assert(body.len == 1);
+            const account_filter = &body[0];
+            account_filter.* = std.mem.zeroInit(tb.AccountFilter, .{});
+
+            account_filter.account_id = if (self.auditor.pick_account(.{
+                .created = true,
+                .debits_must_not_exceed_credits = null,
+                .credits_must_not_exceed_debits = null,
+                .has_transfers = true,
```
matklad: maybe `has_transfers = null`? Feels like a good case to cover.

matklad: Or maybe just pick account at random? (so, .created = null as well)

src/state_machine/workload.zig
```diff
@@ -773,6 +861,204 @@ pub fn WorkloadType(comptime AccountingStateMachine: type) type {
...
+            if (self.auditor.get_account_state(account_filter.account_id)) |account_state| {
+                const transfer_count: u32 = transfer_count: {
+                    const dr_transfer_count: u32 = if (account_filter.flags.debits)
+                        account_state.dr_transfer_count
+                    else
+                        0;
+                    const cr_transfer_count: u32 = if (account_filter.flags.credits)
+                        account_state.cr_transfer_count
+                    else
+                        0;
```
matklad: Consider imperative style, should more concise

src/state_machine/workload.zig
```diff
@@ -773,6 +861,204 @@ pub fn WorkloadType(comptime AccountingStateMachine: type) type {
...
+
+                    break :transfer_count dr_transfer_count + cr_transfer_count;
+                };
+
+                const expected_count = @min(account_filter.limit, transfer_count, batch_size);
+                if (account_filter.timestamp_min == 0 and account_filter.timestamp_max == 0) {
+                    assert(result_count == expected_count);
```
matklad: Hm, how does this work? It seems that new transfers might be created concurrently to our requests? The fact that the account has this many transfers right now doesn't generally mean that it had this many tranfers at the time the lookup was executed. Am I missing some implicit synchronisation logic here here?

src/testing/storage.zig
```diff
@@ -303,8 +303,10 @@ pub const Storage = struct {
...
         }
 
-        // Process the queues in a single loop, since their callbacks may append to each other.
-        while (storage.next_tick_queue.pop()) |next_tick| {
+        // Avoid chained completions from extending how long it's popped.
+        var q = storage.next_tick_queue;
+        storage.next_tick_queue.reset();
+        while (q.pop()) |next_tick| {
             next_tick.callback(next_tick);
         }
```
matklad: Could you expand the commit explaining a bit _why_ we do it this way? 

It looks like the comment (and correspondingly the logic) were changed to do _exactly_ the thing that the previous comment said should not be done, so I am wondering why :rofl:  (and also why this specific change uncovered this effect)

src/state_machine/workload.zig
```diff
@@ -773,6 +861,204 @@ pub fn WorkloadType(comptime AccountingStateMachine: type) type {
...
+                account_state.transfer_timestamp_last +| 1
+            else
+                account_state.transfer_timestamp_first -| 1;
+
+            for (results) |*transfer| {
+                assert(if (account_filter.flags.reversed)
+                    transfer.timestamp < timestamp_last
+                else
+                    transfer.timestamp > timestamp_last);
+                timestamp_last = transfer.timestamp;
```
matklad: I think in loops we usually call such a tracking variable as `_last`

src/state_machine/workload.zig
```diff
@@ -773,6 +861,204 @@ pub fn WorkloadType(comptime AccountingStateMachine: type) type {
...
+            var timestamp_last: u64 = if (account_filter.flags.reversed)
+                account_state.transfer_timestamp_last +| 1
+            else
+                account_state.transfer_timestamp_first -| 1;
+
+            for (results) |*transfer| {
+                assert(if (account_filter.flags.reversed)
+                    transfer.timestamp < timestamp_last
+                else
+                    transfer.timestamp > timestamp_last);
```
matklad: I'd pull `if` outsied of assert (so, duplciating the assert), as it'd be more readable that way. Also, should the assert fail, with two asserts the failure would immediately point a _which_ of the two cases failed. That's basically the same thing why

```
assert(foo)
assert(bar)
```

is superior to

```
assert(foo && bar)
```

matklad (commented):
Realized that I don't understand one big picture aspect of this: in `validate_account_filter_result_count`, we compare the number of transfers the account has at a given moment with the amount of results. But why than works? More transfers could get created in the meantime, no?

matklad:
removed vopr label until this is rebased to 0.13 :)

src/testing/storage.zig
batiati: This commit fixed the seed `13070689082331695065`, which causes the simulator to loop indefinitely (not during the liveness check), because we finished the prefetch of invalid filters using `next_tick`.

I want it to be exhaustively fuzzed by the CFO because I'm not 100% confident that this fix reproduces exactly the same behavior of the real Storage `on_next_tick`, but apparently, the previous code didn't either. 🤣

src/state_machine/workload.zig
batiati: The validation runs _just after_  the query is executed, so the `transfer_count` is up to date. Whatever the number of transfers it has inserted, the number of matches must be one of the three possible values: `@min(account_filter.limit, transfer_count, batch_size)`.

Synchronization will be a problem when implementing `import_{accounts,transfers}` though.

src/state_machine/workload.zig
batiati: We are excluding the first or the last result (if ascending or descending), filtering by timestamp.

```zig
// We want to exclude at least one result:
```

This comment is indeed outdated :+1: 
It says "at least one", but that's no longer true since we can filter by timestamp only when both flags `debits` and `credits` are set, excluding _exactly one_ result.

src/state_machine/workload.zig
batiati: It can be an invalid account.

src/testing/storage.zig
batiati: Reverted the code since it crashed [another test](https://github.com/tigerbeetle/tigerbeetle/actions/runs/9585722422/job/26432244243).
This seed doesn't fail anymore, so let it fuzz for some time!

src/state_machine/workload.zig
batiati: Reworked the logic, now we only rely on `transfer_count` when asserting the result, never when populating `FuzzOp`.

src/state_machine/workload.zig
batiati: The name _last is because it's the last timestamp seen, can be the max or min depending on the sorting order.

src/state_machine/workload.zig
matklad: >The validation runs just after the query is executed,

Hm, does it? If I am reading the code correctly, this logic is triggered when the client receives a reply, and a reply message might be delayed by the network.

src/state_machine/workload.zig
matklad: Urgh, sorry, I managed to write exactly the opposite of what I wanted to write. We typically call such variables

```
timestamp_previous
```

not last --- "last" is the absolute, last of all there is. previous refers to the latest iteration of the loop.

matklad:
vopr failure:

```
./zig/zig build  simulator_run -- --lite 3734369335599541400
thread 6028 panic: reached unreachable code
/home/matklad/p/tb/review/zig/lib/std/debug.zig:412:14: 0x1120acd in assert (simulator)
             ^
/home/matklad/p/tb/review/src/state_machine/workload.zig:1032:27: 0x12085fb in validate_account_filter_result_count (simulator)
                    assert(result_count == @min(account_filter.limit, batch_size));
                          ^
/home/matklad/p/tb/review/src/state_machine/workload.zig:990:54: 0x11d22e2 in on_get_account_balances (simulator)
            self.validate_account_filter_result_count(account_filter, results.len);
                                                     ^
/home/matklad/p/tb/review/src/state_machine/workload.zig:365:70: 0x1172136 in on_reply (simulator)
                .get_account_balances => self.on_get_account_balances(
                                                                     ^
/home/matklad/p/tb/review/src/simulator.zig:829:48: 0x1165aca in on_cluster_reply (simulator)
                    simulator.workload.on_reply(
                                               ^
/home/matklad/p/tb/review/src/testing/cluster.zig:679:37: 0x1162019 in client_on_reply (simulator)
            cluster.on_cluster_reply(cluster, client_index, request_message, reply_message);
                                    ^
/home/matklad/p/tb/review/src/vsr/client.zig:439:34: 0x117ad4d in on_reply (simulator)
                on_reply_callback(self, inflight.message, reply);
                                 ^
/home/matklad/p/tb/review/src/vsr/client.zig:179:44: 0x11254c7 in on_message (simulator)
                .reply => |m| self.on_reply(m),
                                           ^
/home/matklad/p/tb/review/src/testing/cluster/network.zig:317:39: 0x12d475e in deliver_message (simulator)
        target_bus.on_message_callback(target_bus, target_message);
                                      ^
/home/matklad/p/tb/review/src/testing/packet_simulator.zig:403:45: 0x1236d78 in tick (simulator)
                        link_packet.callback(link_packet.packet, path);
                                            ^
/home/matklad/p/tb/review/src/testing/cluster/network.zig:136:38: 0x11f1c0c in tick (simulator)
```

src/state_machine/workload.zig
matklad: :thinking: I think if you need to guarantee timing, you could use replica's on commit hook (in replica.zig):

```
            if (self.event_callback) |hook| hook(self, .committed);
```

Though, I'd say its more valuable to try to keep workfload/validator asynchronous, such that it also works for out-of-process scenarious without much pain.

batiati: > his logic is triggered when the client receives a reply, and a reply message might be delayed by the network.

I'm missing something here. When the `Auditor` validates the balance of an account, if new transfers could change that before the validation, wouldn't it be the same problem?

matklad: Excellent point! I am clearly missing something here as well, let me re-read this whole thing, because this couldn't possible work, and yet evidently it does :rofl:

matklad:
Another one!

```
./zig/zig build -Drelease simulator_run -- --lite 1871810428293586998
thread 235499 panic: reached unreachable code
/home/matklad/p/tb/review/zig/lib/std/debug.zig:412:14: 0x10a598e in validate_account_filter_result_count (simulator)
             ^
/home/matklad/p/tb/review/src/state_machine/workload.zig:897:49: 0x106f3bc in on_reply (simulator)
            validate_account_filter_result_count(
                                                ^
/home/matklad/p/tb/review/src/simulator.zig:829:48: 0x106009f in on_cluster_reply (simulator)
                    simulator.workload.on_reply(
                                               ^
/home/matklad/p/tb/review/src/vsr/client.zig:439:34: 0x1037db8 in on_message (simulator)
                on_reply_callback(self, inflight.message, reply);
                                 ^
/home/matklad/p/tb/review/src/testing/cluster/network.zig:317:39: 0x10e561d in deliver_message (simulator)
        target_bus.on_message_callback(target_bus, target_message);
                                      ^
/home/matklad/p/tb/review/src/testing/packet_simulator.zig:403:45: 0x106894c in tick (simulator)
                        link_packet.callback(link_packet.packet, path);
                                            ^
/home/matklad/p/tb/review/src/simulator.zig:308:23: 0x1066962 in main (simulator)
        simulator.tick();
                      ^
/home/matklad/p/tb/review/zig/lib/std/start.zig:524:37: 0x10360e5 in posixCallMainAndExit (simulator)
                                    ^
/home/matklad/p/tb/review/zig/lib/std/start.zig:266:5: 0x1035b11 in _start (simulator)
```

src/state_machine/workload.zig
matklad: Ok, yeah, i indeed missed big chunk of logic! Simulator makes use of `reply_sequence: ReplySequence,` to reorder replies in their _logical_ order before calling callbacks!

matklad (approved):

PR #2005, opened 2024-06-12
Run alpine container with --security-opt seccomp=unconfined

batiati:
Related https://github.com/tigerbeetle/tigerbeetle/actions/runs/9481890435/job/26125998852?pr=2004

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -12,7 +12,9 @@ jobs:
 
   test_on_alpine:
     runs-on: ubuntu-latest
-    container: alpine
+    container:
+      image: alpine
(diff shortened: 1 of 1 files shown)


.github/workflows/linux.yml
```diff
@@ -12,7 +12,9 @@ jobs:
 
   test_on_alpine:
     runs-on: ubuntu-latest
-    container: alpine
+    container:
+      image: alpine
+      options: --security-opt seccomp=unconfined
```
matklad, line 18: Add an `# comment` explainig why we need this, otherwise this will be a very puzzling line for someone unfamiliar with the context :rofl:

matklad (approved):

PR #2008, opened 2024-06-13
Test/Replica: Shrink `TestClients.request()` when possible

sentientwaffle:
Now that `TestClients` can defer client registration until we intentionally trigger a request, we don't need the `c.request(20, 20)` boilerplate. (Previously, that boilerplate was to guard against "surprise" requests being triggered by `pong_client` messages.)


matklad (approved):
Much better now, thanks a lot for this dj!

PR #2009, opened 2024-06-13
CLI/LSM: Tune compaction memory

sentientwaffle:
### CLI: Add `--memory-lsm-compaction` flag

Add a (deliberate undocumented) `--memory-lsm-compaction` flag to `tigerbeetle start`.

This overrides the `Forest`'s `compaction_blocks` count at runtime.

When `--development` is set, use the minimum possible value.
This reduces the RAM of `tigerbeetle start --development` by about 120MiB.

---

### LSM: Reduce default compaction_block_count

Most of the compaction blocks were not being used.

This reduces the RAM requirement of a production (non `--development`) replica by 112MiB.

### DIFF
--- a/src/constants.zig
+++ b/src/constants.zig
@@ -732,11 +732,12 @@ pub const aof_record = config.process.aof_record;
 /// replay our AOF.
 pub const aof_recovery = config.process.aof_recovery;
 
-/// The amount of memory allocated for compactions. Compactions will be deterministic regardless
-/// of how much memory you give them, but will run in fewer steps with more memory.
-// TODO: Expose this as a CLI flag, to allow tuning of compaction rate.
-// (And use this current value as the default.)
-pub const compaction_block_memory = config.process.compaction_block_memory;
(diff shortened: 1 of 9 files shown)


src/constants.zig
```diff
@@ -732,11 +732,8 @@ pub const aof_record = config.process.aof_record;
 /// replay our AOF.
 pub const aof_recovery = config.process.aof_recovery;
 
-/// The amount of memory allocated for compactions. Compactions will be deterministic regardless
-/// of how much memory you give them, but will run in fewer steps with more memory.
-// TODO: Expose this as a CLI flag, to allow tuning of compaction rate.
-// (And use this current value as the default.)
-pub const compaction_block_memory = config.process.compaction_block_memory;
+/// The maximum size in bytes of the NodePool used for the LSM forest's manifests.
```
cb22: Misdirected comment?

cb22 (approved):

PR #2011, opened 2024-06-13
LSM/SetAssociativeCache: Ease clocks alignment to shrink lower-bound size

sentientwaffle:
Ease clock hand alignment restrictions to:
- Shrink cache lower-bound.
- Allow more granular tuning of the cache size.

Each clock hand is (still) guaranteed (by `comptime` asserts) to not span multiple cache lines.
But in order to shrink the lower-bound cache size, remove the requirement that `clocks` itself is a multiple of the cache line size.

Old `value_count_max_multiple`:
- Grid cache: 512 values.
- Object caches: 2048 values.

New `value_count_max_multiple`:
- Grid cache: 64 values.
- Object caches: 256 values.

64 values is a minimum grid cache size of 32MiB, down from 256MiB.


cb22:
LGTM - putting it through the cache map fuzz for a bit too.

cb22 (approved):

matklad:
We merged zig 0.13, so this might want a rebase to unbreak the fuzzer

cb22:
> so this might want a rebase to unbreak the fuzzer

Is the fuzzer broken? :thinking: The GH rebased it automatically / I tested with the cache map fuzz locally manually rebased and it seemed ok.

cb22:
Ah or do you mean those seeds that were showing up for this on devhub?

sentientwaffle:
I'm guessing CFO doesn't rebase automatically, so it was trying to run the 0.11 code with zig 0.13?

matklad:
> I'm guessing CFO doesn't rebase automatically, so it was trying to run the 0.11 code with zig 0.13?


this: cfo doesn’t rebase, and it uses the same zig version for all fuzzing

PR #2014, opened 2024-06-17
devhub: make it harder to miss failures due to visualization bugs

matklad:
Our dev hub erronusly filters out some failing seeds! :fearful: 

- fix the filtering logic
- add summary states for the total number of seeds, which would tell you if the filtering logic is buggy


matklad:
image

sentientwaffle (approved):

PR #2018, opened 2024-06-17
cfo: avoid useless compilations

matklad:
Before this commit, CFO used to proactively compile all fuzzers for all branches. After this change, only the required fuzzers are compiled.

Originally, the code was written with a clean split between "here we build all the stuff we need" and "here we run all the stuff we need". But, as the "stuff" became more dynamic (specifically, when we start fuzzing labled PRs), the building logic got less precise: it was still building everything for every branch, even if only a single fuzzer is needed.

So instead run a separate build command for a specific fuzzer just before we run it. While we could just run the fuzzer directly and rely on zig to do the build automatically, this'll give us less precise timings.

While we are at it, add a somewhat more clear logic about the fuzzing plan for a parictular run of the CFO.

### DIFF
--- a/src/scripts/cfo.zig
+++ b/src/scripts/cfo.zig
@@ -78,7 +78,19 @@ const Fuzzer = enum {
     vopr_lite,
     vopr_testing_lite,
 
-    fn fill_args(fuzzer: Fuzzer, accumulator: *std.ArrayList([]const u8)) !void {
+    fn fill_args_build(fuzzer: Fuzzer, accumulator: *std.ArrayList([]const u8)) !void {
+        switch (fuzzer) {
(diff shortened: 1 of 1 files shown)


matklad:
Backstory: I am trying to understand why we have this seed from a week ago:

[b17a2f2](https://github.com/tigerbeetle/tigerbeetle/commit/b17a2f255651732931dcd0ee9fbb52cc7cdd7d50)		[vopr](https://tigerbeetle.github.io/tigerbeetle/?fuzzer=vopr&commit=b17a2f255651732931dcd0ee9fbb52cc7cdd7d50)	./zig/zig build -Drelease simulator_run -- 6246930092533231845	23s	8d 4h 56m ago

and why it wasn't "refreshed" since that time

cb22 (approved):

PR #2023, opened 2024-06-18
vopr: fix Zig TODO

matklad:
It now can generate random u128

Drive by change when reviewing #2004


batiati (approved):
Reinforcing the importance of marking code like this with `\\ TODO(zig)` ... we should have been able to track this one since we updated it to 0.11!

matklad:
I think this actually requries Zig 0.12! And it's correct that we fix this separately, rather than in that grand upgrade, to keep PRs smaller.

PR #2024, opened 2024-06-18
build: simplify linking tracy

matklad:
With Zig 0.13, it is now possible to automatically propagte C
dependencies between modules. So, it becomes sufficient to configury
tracy only for the vsr_module, and everything else gets the appropriate
linker flags for free!

We can also get rid of our bespoke GitCloneStep along the way for an even nicer simplification!

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -62,9 +62,6 @@ pub fn build(b: *std.Build) !void {
 
     const options = b.addOptions();
 
-    var shell = Shell.create(b.allocator) catch unreachable;
-    defer shell.destroy();
-
     // The "tigerbeetle version" command includes the build-time commit hash.
     const git_commit = b.option(
         []const u8,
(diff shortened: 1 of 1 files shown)


build.zig
```diff
@@ -127,6 +119,48 @@ pub fn build(b: *std.Build) !void {
...
+                    "-D_WIN32_WINNT=0x601",
+                }
+            else
+                &[_][]const u8{
+                    "-DTRACY_ENABLE=1",
+                    "-DTRACY_FIBERS=1",
+                    "-fno-sanitize=undefined",
+                };
+
+            const tracy = b.addSystemCommand(&.{
```
kprotty, line 143: neat

kprotty (approved):

build.zig
matklad: I have a slight suspicion that something like this was actually possible back in 0.9 days, and that my shell.zig based GitCloneStep was a pile of neeedless accidental complexity from the start, but I didn't pursue that inquiry in order to remain optimistic :)

PR #2027, opened 2024-06-19
vsr: unify checking for production config

matklad:
"Is this a production config?" is a somewhat ill-formed question, as, physically, a config is just a set of many _specific_ options rather than a broad toggle. But there are a couple of places where we need an answer to that question to enable better sanity checking! This PR introduces a specific function for this check, but it _should not_ be used in "real" code --- instead, check for specific properties like `config.verify`.

The main motivation here is preventing benchmark from poking into `vsr_options`, which would allow us to simplify build.zig

### DIFF
--- a/src/config.zig
+++ b/src/config.zig
@@ -69,6 +69,14 @@ pub const Config = struct {
 
     cluster: ConfigCluster,
     process: ConfigProcess,
+
+    /// Returns true if the configuration is intended for "production".
+    /// Intended solely for extra sanity-checks: all meaningful decisions should be driven by
+    /// specific fields of the config.
+    pub fn is_production(config: *const Config) bool {
+        _ = config;
+        return build_options.config_base == .production;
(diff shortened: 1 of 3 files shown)


src/config.zig
```diff
@@ -69,6 +69,14 @@ pub const Config = struct {
 
     cluster: ConfigCluster,
     process: ConfigProcess,
+
+    /// Returns true if the configuration is intended for "production".
+    /// Intended solely for extra sanity-checks: all meaningful decisions should be driven by
+    /// specific fields of the config.
+    pub fn is_production(config: *const Config) bool {
+        _ = config;
+        return build_options.config_base == .production;
```
matklad, line 78: I originally did `config.process.direct_io` thing here, but then realized I can actually cheat and look at the build options directly!

sentientwaffle (approved):

PR #2028, opened 2024-06-19
Scan fuzz testing

batiati:
This fuzzer generates random arbitrary complex query conditions such as `(a OR b) AND (c OR d OR (e AND f AND g))`.

It also includes an array of at least one object that matches the condition. Those objects are used as templates to populate the database in such a way that the results retrieved by the query can be asserted.

The idea behind it was _vaguely_ inspired by [this paper](https://xblk-ecnu-edu-cn.translate.goog/article/2021/1000-5641/2021-5-74.shtml?_x_tr_sl=auto&_x_tr_tl=en&_x_tr_hl=en&_x_tr_pto=wapp)

> [...] all non-primary key attributes are deterministically calculated based on the primary key through the attribute generation function.
The load generator also uses the results of the Schema generator. In the load generator, a complex load template is first generated according to the number of defined connection operators. The load template generated by the load generator ensures that the syntax and semantics are correct.

### DIFF
--- a/src/config.zig
+++ b/src/config.zig
@@ -167,10 +167,10 @@ const ConfigCluster = struct {
     lsm_manifest_compact_extra_blocks: comptime_int = 1,
     vsr_releases_max: usize = 64,
 
-    /// Enough for scanning the three user_data fields plus code and ledger.
+    /// Minimal value.
     // TODO(batiati): Maybe this constant should be derived from `grid_iops_read_max`,
     // since each scan can read from `lsm_levels` in parallel.
-    lsm_scans_max: comptime_int = 5,
+    lsm_scans_max: comptime_int = 2,
 
     /// The WAL requires at least two sectors of redundant headers — otherwise we could lose them
     /// all to a single torn write. A replica needs at least one valid redundant header to
@@ -289,6 +289,8 @@ pub const configs = struct {
             .lsm_growth_factor = 4,
             // (This is higher than the production default value because the block size is smaller.)
             .lsm_manifest_compact_extra_blocks = 5,
+            // (We need to fuzz more scans merge than in production.)
+            .lsm_scans_max = 12,
         },
     };
 
--- a/src/config.zig
+++ b/src/config.zig
@@ -170,7 +170,7 @@ const ConfigCluster = struct {
     /// Minimal value.
     // TODO(batiati): Maybe this constant should be derived from `grid_iops_read_max`,
     // since each scan can read from `lsm_levels` in parallel.
-    lsm_scans_max: comptime_int = 2,
+    lsm_scans_max: comptime_int = 5,
 
     /// The WAL requires at least two sectors of redundant headers — otherwise we could lose them
     /// all to a single torn write. A replica needs at least one valid redundant header to
--- a/src/config.zig
+++ b/src/config.zig
@@ -167,7 +167,7 @@ const ConfigCluster = struct {
     lsm_manifest_compact_extra_blocks: comptime_int = 1,
     vsr_releases_max: usize = 64,
 
-    // Minimal value.
+    /// Enough for scanning the three user_data fields plus code and ledger.
     // TODO(batiati): Maybe this constant should be derived from `grid_iops_read_max`,
     // since each scan can read from `lsm_levels` in parallel.
     lsm_scans_max: comptime_int = 5,
--- a/src/lsm/scan_fuzz.zig
+++ b/src/lsm/scan_fuzz.zig
@@ -582,8 +582,9 @@ const QuerySpecFuzzer = struct {
                     stack.append_assume_capacity(templates);
                 },
                 .merge => |merge| {
-                    const templates_to_merge = stack.slice()[stack.count() - merge.operand_count ..];
                     var templates: Templates = .{};
+                    const templates_to_merge =
+                        stack.slice()[stack.count() - merge.operand_count ..];
 
                     switch (merge.operator) {
                         .union_set => {
@@ -608,7 +609,8 @@ const QuerySpecFuzzer = struct {
                                 for (0..merge.operand_count) |i| {
                                     const reverse_index = merge.operand_count - 1 - i;
                                     if (reverse_index == 0 or
-                                        matrix_indexes[reverse_index] < templates_to_merge[reverse_index].count() - 1)
+                                        matrix_indexes[reverse_index] <
+                                        templates_to_merge[reverse_index].count() - 1)
                                     {
                                         matrix_indexes[reverse_index] += 1;
                                     } else {
--- a/src/lsm/scan_fuzz.zig
+++ b/src/lsm/scan_fuzz.zig
@@ -251,7 +251,7 @@ const QuerySpec = struct {
 };
 
 /// This fuzzer generates random arbitrary complex query conditions such as
-/// `(a OR b) and (c OR d OR (e AND f AND g))`.
+/// `(a OR b) AND (c OR d OR (e AND f AND g))`.
 /// It also includes an array of at least one object that matches the condition.
 /// Those objects are used as template to populate the database in such a way
 /// that the results retrieved by the query can be asserted.
@@ -566,8 +566,12 @@ const Environment = struct {
     pub fn run(
         storage: *Storage,
         random: std.rand.Random,
+        /// Repeating multiple times is valuable since it populates
+        /// more data and scans again after compaction.
         repeat: u32,
     ) !void {
+        assert(repeat > 0);
+
         var env: Environment = undefined;
         try env.init(storage, random);
         defer env.deinit();
@@ -592,45 +596,40 @@ const Environment = struct {
             random,
         );
 
-        try env.apply(query_specs, repeat);
+        for (0..repeat) |_| {
+            try env.apply(query_specs);
+        }
     }
 
     fn apply(
         env: *Environment,
         query_specs: []QuerySpec,
-        /// Repeating the same test multiple times is valuable since
-        /// it scans the same values after compaction.
-        repeat: u32,
     ) !void {
-        assert(repeat > 0);
         assert(env.state == .fuzzing);
 
-        for (0..repeat) |step| {
-            // Inserting one batch for each query spec.
-            for (query_specs) |*query_spec| {
-                try env.populate_things(query_spec);
-            }
+        // Inserting one batch for each query spec.
+        for (query_specs) |*query_spec| {
+            try env.populate_things(query_spec);
+        }
 
-            // Executing each query spec.
-            for (query_specs) |*query_spec| {
-                std.log.info(
-                    \\step: {}
-                    \\query {}:
-                    \\expected_results: {}
-                    \\reversed: {}
-                    \\condition: {}
-                    \\
-                    \\
-                , .{
-                    step,
-                    query_spec.prefix,
-                    query_spec.expected_results,
-                    query_spec.reversed,
-                    query_spec.condition,
-                });
+        // Executing each query spec.
+        for (query_specs) |*query_spec| {
+            log.debug(
+                \\prefix: {}
+                \\object_count: {}
+                \\expected_results: {}
+                \\reversed: {}
+                \\condition: {}
+                \\
+            , .{
+                query_spec.prefix,
+                env.object_count,
+                query_spec.expected_results,
+                query_spec.reversed,
+                query_spec.condition,
+            });
 
-                try env.query(query_spec);
-            }
+            try env.query(query_spec);
         }
     }
 
@@ -976,6 +975,8 @@ pub fn main(fuzz_args: fuzz.FuzzArgs) !void {
 }
 
 fn prefix_combine(prefix: u32, suffix: u32) u64 {
+    assert(prefix != 0);
+    assert(suffix != 0);
     return @as(u64, @intCast(prefix)) << 32 |
         @as(u64, @intCast(suffix));
 }
--- a/src/lsm/scan_fuzz.zig
+++ b/src/lsm/scan_fuzz.zig
@@ -30,7 +30,8 @@ const batch_max: u32 = @divFloor(
 /// The testing object.
 const Thing = extern struct {
     id: u128,
-    // All indexes must be `u64` to avoid conflicting matching values:
+    // All indexes must be `u64` to avoid conflicting matching values,
+    // the most significant bits are a seed used for assertions.
     index_1: u64,
     index_2: u64,
     index_3: u64,
@@ -43,7 +44,7 @@ const Thing = extern struct {
     index_10: u64,
     index_11: u64,
     index_12: u64,
-    reserved: u64,
+    checksum: u64,
     timestamp: u64,
 
     comptime {
@@ -52,6 +53,7 @@ const Thing = extern struct {
         assert(@alignOf(Thing) == 16);
     }
 
+    /// Initializes a struct with all fields zeroed.
     fn zeroed() Thing {
         return .{
             .id = 0,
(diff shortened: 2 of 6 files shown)


batiati:
First seed found `16495112179316261289` :partying_face: 

It looks like our assertions about how scans get drained cannot hold with this kind of query:
```SQL
WHERE ((index_11=4294967297 OR index_3=4294967298 OR index_9=4294967299 OR index_10=4294967300) AND index_6=4294967301 AND (((index_12=4294967302 OR index_2=4294967303))))
```

src/lsm/scan_fuzz.zig
```diff
@@ -0,0 +1,991 @@
...
+                            .union_set => try writer.print(" OR ", .{}),
+                            .intersection_set => try writer.print(" AND ", .{}),
+                        }
+                    }
+                }
+                try writer.print(")", .{});
+            },
+        }
+    }
+};
```
matklad: So this gives me a pause. My nit comment is that, with many types flowing around, it's a bit hard to see the overall recursive structure of of types here. So perhaps some judicial usage of nesting would help to illumintate that?

Something like

```zig
const QueryCondition = union(enum) {
    const Field = struct {}
    const ParaenthesisCondition = struct { /* uses QueryCondition here*/ }

    field: Field,
    parenthesis_condition: ParaenthesisCondition,
}
```

just so that you can immediately see "ah, those things are circular, so what I am looking at is a tree". 

I guess another nit is that `ParenthesisCondition` is perhaps not the best name, as its simultaneously long and is not very illuminating. 

But my bigger worry is that such recursive code feels un-TigerStyle. It says **Do not use recursion**, but `scan_from_condition` is recursive. While this recursive AST structure is near and dear to my heart of recovering compiler writer, I am wondering if there's some more TigerStyle representation here? 

For example, I think reverse polish notation might be a more conductive to simple loops?

```zig
const Query = BoundedArray(QueryPart, query_len_max);
const QueryPart = union(enum) {
    field: struct {index: Index, value: u64 }
    merge: struct { operator: QueryOperator, operand_count: u8 } // NB: non-recursive
}
```

then, `scan_from_condition` can go non-recursively as:

```zig
const stack = BoundedArray(*Scan, query_len_max);
for (query.items) |query_part| {
    switch(query_part) {
        .field => stack.push(scan_builder.scan_prefix())
        .merge => |merge| {
              const scan = scan_builder.merge(stack.items[stack.items.len - merge.operand_count..]);
              stack.pop_many(merge.operand_count);
              stack.push(scan);
        }
    }
}
assert(stack.len == 1)
```

matklad: To clarify, I think its fine for fuzzers to be less TigerStyle than the main code. Still, perhaps it'd be interesting to use the fuzzer as a test bed to figure out patters for working with "trees" in TigerStyle. Maybe recursion is the best answer after all!

src/config.zig
```diff
@@ -296,6 +296,7 @@ pub const configs = struct {
     pub const fuzz_min = config: {
         var base = test_min;
         base.process.storage_size_limit_max = 1 * 1024 * 1024 * 1024;
+        base.cluster.lsm_scans_max = 12;
```
matklad: Ouch, I didn't realize we are adjusting fuzz_min here, I have just removed it on main :crying_cat_face:

src/lsm/scan_fuzz.zig
```diff
@@ -0,0 +1,1170 @@
...
+            for (1..n + 1) |i| {
+                result *= i;
+            }
+            return result;
+        }
+    }.factorial;
+
+    break :template_max factorial(constants.lsm_scans_max) /
+        (2 * factorial(constants.lsm_scans_max - 2));
+};
```
matklad: No strong opition, but I'd maybe directly computed `C(N, 2)` as `N*(N-1)/2`, just for simplicty. If we compute the fully genreal version, I'd go for something like https://stackoverflow.com/a/15302448, as more computation-friendly method.

matklad: I guelss the leass nitpicky comment here is that I don't immediately understand _why_ C(N, 2) is the right answer here.

src/lsm/scan_fuzz.zig
```diff
@@ -0,0 +1,1170 @@
...
+        const query_spec_count = random.intRangeAtMostBiased(
+            usize,
+            2,
+            query_specs.inner.capacity(),
+        );
+
+        for (0..query_spec_count) |prefix| {
+            var fuzzer = QuerySpecFuzzer{
+                .random = random,
+                .prefix = @truncate(prefix + 1),
```
matklad: This should be `@intCast`, right? Like with `/` vs `@divExact`, we prefer throwing alternatives.

src/config.zig
batiati: We can live with the production value, the only problem is that right now it's too low to stress any meaningful condition; However it soon will be increased by another PR.

I just need to adjust the code to consider the minimum between scans_max and the fields count.

src/lsm/scan_fuzz.zig
```diff
@@ -0,0 +1,1170 @@
...
+                    stack.truncate(stack.count() - merge.operand_count);
+                    stack.append_assume_capacity(templates);
+                },
+            }
+        }
+
+        assert(stack.count() == 1);
+        return stack.get(0);
+    }
+};
```
matklad, line 548: Hm, statically pre-generating all templates feels very complicated! Can we get by without templates? I think we could generate a random value from the query diretly perhaps? Somethign like this:

```diff
diff --git a/src/lsm/scan_fuzz.zig b/src/lsm/scan_fuzz.zig
index e8bc5929f..ffe3870f9 100644
--- a/src/lsm/scan_fuzz.zig
+++ b/src/lsm/scan_fuzz.zig
@@ -107,6 +107,12 @@ const Thing = extern struct {
         }
     }
 
+    fn merge_all(things: []const Thing) Thing {
+        var result = Thing.zeroed();
+        for (things) |thing| result.merge(thing);
+        return result;
+    }
+
     /// Creates a struct from a template, setting all zeroed fields with random values and
     /// calculating the checksum of the resulting struct.
     fn from_template(template: Thing, random: std.rand.Random, id: u128, timestamp: u64) Thing {
@@ -207,34 +213,12 @@ const Scan = ThingsGroove.ScanBuilder.Scan;
 /// If `lsm_scans_max == x`, then we can have at most x fields and x - 1 merge operations.
 const query_part_max = (constants.lsm_scans_max * 2) - 1;
 
-/// The max number of combinations all indexes can produce.
-/// Combinations Formula: `C(n,r) = n! / (r! * (n - r)!)`,
-/// where `n == lsm_scans_max` and `r == 2` (either assigned or zeroed).
-const template_max: u32 = template_max: {
-    assert(constants.lsm_scans_max >= 2);
-
-    const factorial = struct {
-        fn factorial(comptime n: comptime_int) comptime_int {
-            var result = 1;
-            for (1..n + 1) |i| {
-                result *= i;
-            }
-            return result;
-        }
-    }.factorial;
-
-    break :template_max factorial(constants.lsm_scans_max) /
-        (2 * factorial(constants.lsm_scans_max - 2));
-};
-
 /// The max number of query specs generated per run.
 /// Always generate more than one query spec, since multiple queries can
 /// test both the positive space (results must match the query) and the
 /// negative space (results from other queries must not match).
 const query_spec_max = 8;
 
-const Templates = stdx.BoundedArray(Thing, template_max);
-
 const QueryPart = union(enum) {
     const Field = struct { index: Index, value: u64 };
     const Merge = struct { operator: QueryOperator, operand_count: u8 };
@@ -269,9 +253,6 @@ const QuerySpec = struct {
     query: Query,
     // Ascending or descending.
     reversed: bool,
-    // Matching templates to populate the database.
-    // Zeroed fields are not part of the condition and can be filled with random values.
-    templates: Templates,
     // Number of expected results.
     expected_results: u32,
 
@@ -388,12 +369,10 @@ const QuerySpecFuzzer = struct {
     ) QuerySpec {
         const field_max = self.random.intRangeAtMostBiased(u32, 1, constants.lsm_scans_max);
         const query = self.generate_query(field_max);
-        const templates = get_templates(query.const_slice());
 
         return QuerySpec{
             .prefix = self.prefix,
             .query = query,
-            .templates = templates,
             .reversed = self.random.boolean(),
             .expected_results = 0,
         };
@@ -558,87 +537,6 @@ const QuerySpecFuzzer = struct {
             return index;
         }
     }
-
-    /// Templates are objects that match the condition, such as:
-    /// - The condition `a=1` (from now on represented only as `a` for brevity)
-    ///   generates a template with the field `a` set to `1`, so when inserting
-    ///   objects based on this template, they're expected to match.
-    ///
-    /// - The condition `(a OR b)` generates two templates, one with the
-    ///   field `a` and another one with the field `b` set, so both of them
-    ///   can satisfy the OR clause.
-    ///
-    /// - The condition `(a AND b)` generates only a single template that
-    ///   matches the criteria with both fields `a` and `b` set to the
-    ///   corresponding value.
-    ///
-    /// More complex queries like `(a OR b) AND (c OR d)` need to combine the
-    /// templates generated by each individual condition in such a way any of
-    /// them can satisfy the OR and AND clauses: {a,c},{a,d},{b,c}, and {b,d}.
-    fn get_templates(
-        query_parts: []const QueryPart,
-    ) Templates {
-        var stack = stdx.BoundedArray(Templates, query_part_max){};
-        for (query_parts) |query_part| {
-            switch (query_part) {
-                .field => |field| {
-                    var template = Thing.zeroed();
-                    template.set_index(field.index, field.value);
-
-                    var templates: Templates = .{};
-                    templates.append_assume_capacity(template);
-                    stack.append_assume_capacity(templates);
-                },
-                .merge => |merge| {
-                    var templates: Templates = .{};
-                    const templates_to_merge =
-                        stack.slice()[stack.count() - merge.operand_count ..];
-
-                    switch (merge.operator) {
-                        .union_set => {
-                            // Templates should be added on unions.
-                            for (templates_to_merge) |operand_templates| {
-                                for (operand_templates.const_slice()) |template| {
-                                    templates.append_assume_capacity(template);
-                                }
-                            }
-                        },
-                        .intersection_set => {
-                            // Templates should be merged on intersections,
-                            // see the documentation above for more details.
-                            var matrix_indexes = [_]u32{0} ** query_part_max;
-                            while (matrix_indexes[0] < templates_to_merge[0].count()) {
-                                var template = Thing.zeroed();
-                                for (templates_to_merge, 0..) |template_to_merge, i| {
-                                    template.merge(template_to_merge.get(matrix_indexes[i]));
-                                }
-                                templates.append_assume_capacity(template);
-
-                                for (0..merge.operand_count) |i| {
-                                    const reverse_index = merge.operand_count - 1 - i;
-                                    if (reverse_index == 0 or
-                                        matrix_indexes[reverse_index] <
-                                        templates_to_merge[reverse_index].count() - 1)
-                                    {
-                                        matrix_indexes[reverse_index] += 1;
-                                    } else {
-                                        matrix_indexes[reverse_index] = 0;
-                                        matrix_indexes[reverse_index - 1] += 1;
-                                    }
-                                }
-                            }
-                        },
-                    }
-
-                    stack.truncate(stack.count() - merge.operand_count);
-                    stack.append_assume_capacity(templates);
-                },
-            }
-        }
-
-        assert(stack.count() == 1);
-        return stack.get(0);
-    }
 };
 
 const Environment = struct {
@@ -807,8 +705,6 @@ const Environment = struct {
 
     // TODO: sometimes update and delete things.
     fn populate_things(env: *Environment, query_spec: *QuerySpec) !void {
-        assert(query_spec.templates.count() > 0);
-
         for (0..batch_max) |_| {
             // Total number of objects inserted.
             env.object_count += 1;
@@ -826,28 +722,49 @@ const Environment = struct {
                 continue;
             }
 
-            const template = query_spec.templates.slice()[
-                env.random.intRangeAtMostBiased(
-                    usize,
-                    0,
-                    query_spec.templates.count() - 1,
-                )
-            ];
-
+            const template = env.random_thing_matching_query(query_spec);
             query_spec.expected_results += 1; // Expected objects that match the spec.
-            env.forest.grooves.things.insert(&template.from_template(
+            const thing = template.from_template(
                 env.random,
                 prefix_combine(
                     query_spec.prefix,
                     query_spec.expected_results,
                 ),
                 env.object_count,
-            ));
+            );
+            env.forest.grooves.things.insert(&thing);
         }
 
         try env.commit();
     }
 
+    fn random_thing_matching_query(env: *Environment, query_spec: *QuerySpec) Thing {
+        var stack: [query_part_max]Thing = undefined;
+        var stack_top: usize = 0;
+        for (query_spec.query.slice()) |query_part| {
+            switch (query_part) {
+                .field => |field| {
+                    var thing = Thing.zeroed();
+                    thing.set_index(field.index, field.value);
+                    stack[stack_top] = thing;
+                    stack_top += 1;
+                },
+                .merge => |merge| {
+                    const operands = stack[stack_top - merge.operand_count .. stack_top];
+                    const result = switch (merge.operator) {
+                        .union_set => operands[env.random.uintLessThan(usize, operands.len)],
+                        .intersection_set => Thing.merge_all(operands),
+                    };
+                    stack_top -= merge.operand_count;
+                    stack[stack_top] = result;
+                    stack_top += 1;
+                },
+            }
+        }
+        assert(stack_top == 1);
+        return stack[0];
+    }
+
     fn run_query(env: *Environment, query_spec: *QuerySpec) !void {
         assert(query_spec.expected_results > 0);
 
```

src/lsm/scan_fuzz.zig
```diff
@@ -0,0 +1,1170 @@
...
+            if (value != 0) {
+                assert(template.get_index(index) == 0);
+                template.set_index(index, value);
+            }
+        }
+    }
+
+    /// Creates a struct from a template, setting all zeroed fields with random values and
+    /// calculating the checksum of the resulting struct.
+    fn from_template(template: Thing, random: std.rand.Random, id: u128, timestamp: u64) Thing {
```
matklad: usually, when we have  several numeric arguments, we pack them into an options struct:

```
options: struct { id: u128, timestamp: u64}
```

that way its harder to mix them up. In this case, I think it'll also help call-site readability, as it is non-trivial observaion that we use object count as our timestamp.

src/lsm/scan_fuzz.zig
batiati: > I guelss the leass nitpicky comment here is that I don't immediately understand _why_ C(N, 2) is the right answer here. 

We could narrow this number down, if we observe those premises:

- A merge involving N fields using "AND" would generate a single template; The "OR" merge with N fields would generate N templates;
- Joining two merges together generates N1 + N2 templates if "OR", or N1 * N2 if "AND".
- Each parent merge aggregates the templates of the children merges.
- A query with N fields can have N-1 merges in total.

For example a query with N fields, such as (a or b) and (c or d) would generate at the worst case (N/2)^2 templates

I cannot mathematically prove it 😬, so C(N,2) appeared to be a safer upper limit, since we don't repeat the same index in a query, and each index would be either in use or blank; So even the worst case would not produce more templates than all combinations of fields involved.

Maybe we could let the fuzzer prove it instead 🤓

batiati: Well, I can prove _by example_ that Cr(12,2) = 66, isn't the right answer either :sweat_smile: 
```
(1 or 2 or 3 or 4 or 5 or 6 or 7 or 8 or 9 or 10 or 11 or 12) = 12 templates
(1 or 2 or 3 or 4 or 5 or 6) and (7 or 8 or 9 or 10 or 11 or 12) = 36 templates
(1 or 2 or 3 or 4 or 5) and (6 or 7 or 8 or 9 or 10) and (11 or 12) = 50 templates
(1 or 2 or 3 or 4 or 5) and (6 or 7 or 8 or 9) and (10 or 11 or 12) = 60 templates
(1 or 2 or 3 or 4) and (5 or 6 or 7 or 8) and (9 or 10 or 11 or 12) = 64 templates
(1 or 2) and (3 or 4) and (5 or 6) and (7 or 8) and (9 or 10) and (11 or 12) = 64 templates
(1 or 2 or 3 or 4) and (5 or 6 or 7) and (8 or 9 or 10) and (11 or 12) = 72 templates 💩
(1 or 2 or 3) and (4 or 5 or 6) and (7 or 8 or 9) and (10 or 11 or 12) = 81 templates 💩
```

src/lsm/scan_fuzz.zig
```diff
@@ -850,12 +743,70 @@ const Environment = struct {
...
+    ///   can satisfy the OR clause.
+    ///
+    /// - The condition `(a AND b)` generates only a single template that
+    ///   matches the criteria with both fields `a` and `b` set to the
+    ///   corresponding value.
+    ///
+    /// More complex queries like `(a OR b) AND (c OR d)` need to combine the
+    /// templates generated by each individual condition in such a way any of
+    /// them can satisfy the OR and AND clauses: {a,c},{a,d},{b,c}, and {b,d}.
+    fn template_matching_query(env: *const Environment, query_spec: *const QuerySpec) Thing {
```
matklad, line 769: I've realized that the thing we are doing here has a fancy name. I _think_ this is [abstract interpretation](https://en.wikipedia.org/wiki/Abstract_interpretation) --- what we are doing here is essentially evaluating the query, except that we substitude a different domain of random templates instead of actual sets of objects

matklad (approved):
:+1: 

One thing we _don't_ do here is interspersing scans and mutation of the tree, and that feels like important case, esp given that scans do `self.groove.objects.table_mutable.sort();`.

But that should be covered by VOPR I think, now that it does lookup_accounts and such.

PR #2029, opened 2024-06-19
build: drop shellcheck 

matklad:
We currently have the following shellchecked scripts:

scripts/install_zig.sh

src/scripts/cfo_supervisor.sh
src/docs_website/scripts/build.sh
.github/ci/docs_check.sh
.github/ci/test_aof.sh

All of them except `install_zig` are run only our our machines. We don't
need to lint those, "works on my machine" is the only thing we need from
those scripts.

install_zig is run on the user's machine, and we do need to make it a
good script, but I think we can manage one decent script without a
linter, and this script hasn't seen much changes lately anyway.

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -236,20 +236,6 @@ pub fn build(b: *std.Build) !void {
         run_step.dependOn(&run_cmd.step);
     }
 
-    // Linting targets
-    // We currently have: lint_zig_fmt.
-    // The meta-target lint runs them all
-    {
-        // lint_zig_fmt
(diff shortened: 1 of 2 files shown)


build.zig
```diff
@@ -5,7 +5,6 @@ const CrossTarget = std.zig.CrossTarget;
 const Mode = std.builtin.Mode;
 
 const config = @import("./src/config.zig");
-const Shell = @import("./src/shell.zig");
```
matklad, line 8: Like this one especially!

sentientwaffle (approved):

PR #2030, opened 2024-06-19
zig: macOS no longer stack overflows

kprotty:
The previous recursiveness in Groove no longer stack overflows (including on macOS 14.5)


matklad:
Excellent, thanks!

Crazy idea: in `constants.verify=true`, hack `log.debug`, such that it also records current stack depth and panics if its too deep.

matklad (approved):

PR #2032, opened 2024-06-19
General cleanup of C header generation

matklad:
* move the code for generating a header from a dedicated binary to a test, to simplify build.zig
* add missing items to the header tests
* make sure that adding items to the test manually  is not required

Best reviewed per-commit!


build.zig
```diff
@@ -321,7 +308,7 @@ pub fn build(b: *std.Build) !void {
         go_client(
             b,
             mode,
-            &.{ &install_step.step, &tb_client_header_generate.step },
+            &.{&install_step.step},
```
matklad: We perhaps no longer need an array here, but I am basically rewriting our build.zig anyway, so will probably fix that at that point.

src/clients/c/tb_client_header_generate.zig
```diff
@@ -247,5 +250,19 @@ pub fn main() !void {
...
-    try std.fs.cwd().writeFile(.{ .sub_path = "src/clients/c/tb_client.h", .data = buffer.items });
+
+    const file_path = "src/clients/c/tb_client.h";
+
+    const contents_new = buffer.items;
+    const contents_old = std.fs.cwd().readFileAlloc(allocator, file_path, 1024 * 1024) catch
+        try allocator.alloc(u8, 0);
+    defer allocator.free(contents_old);
+
+    if (!std.mem.eql(u8, contents_new, contents_old)) {
```
matklad: We have `Shell.file_ensure_contents` for exactly this logic, but i don't want to pull `shell.zig` here as our overall module setup is a bit janky, and it would be simpler if this _doesn't_ depend on stdx and such.

src/clients/c/tb_client_header_generate.zig
```diff
@@ -137,7 +138,9 @@ fn emit_struct(
     try buffer.writer().print("}} {s};\n\n", .{c_name});
 }
 
-pub fn main() !void {
+// This tests generates the `tb_client.h` file. If you make changes to the C ABI of the client, run
+// `./zig/zig build test -- "tb_client.h: generate"` to re-generate the header file.
+test "tb_client.h: generate" {
```
matklad: I am slightly concerned here that the same CU both generates the header file, and also `@cImports` it, so there _might_ be some unfortunate cases where we cant' compile tests because the header is broken, but we can't renegenerate header because we can't compile tests. I reasoned that that's OK, as you can always comment out the offending code to unblock re-generation, which is, yes, non-ideal, but I think is fine for the rare use-case.

But then I recalled that Zig is lazy, and that, if you pass in `./zig/zig build test -- "tb_client.h: generate"`, then _only_ this specific test will get compiled. So we are fine here actually!

src/clients/c/tb_client_header_test.zig
```diff
@@ -45,6 +45,9 @@ test "tb_client.h: verify" {
         .{ tb.CreateTransferResult, "TB_CREATE_TRANSFER_RESULT" },
         .{ tb.CreateAccountsResult, "tb_create_accounts_result_t" },
         .{ tb.CreateTransfersResult, "tb_create_transfers_result_t" },
+        .{ tb.AccountFilter, "tb_account_filter_t" },
+        .{ tb.AccountFilterFlags, "TB_ACCOUNT_FILTER_FLAGS" },
+        .{ tb.AccountBalance, "tb_account_balance_t" },
```
batiati, line 50: :heart:

build.zig
batiati: I'm worried if it would impact the development cycle for the Go client, since now we need to regenerate the header in a separate step before building the Go library.

Couldn't we run the header test during `./zig/zig build scripts -- ci --language=go` as we do with the JNI test for the Java client?

matklad: I don't think this would affect work on the go client much. tb_client.h is commited to the main branch, and, thanks to the test, every commit guarantees that it actually matches our C client. 

So, if you want to hack on go client in isolation, there's no need to manually trigger the rebuild, beacause the file is guaranteed to be fresh.

What might get harder are changes where you want to atomically change both the C client, and also the Go client. In this situation, you'd have to manually kick `./zig/zig build test -- "tb_client.h: generate"`. But this feels ok: it makes sense that, before hacking on clients, you want to make sure that the tests for TigerBeetle itself pass.

batiati: > What might get harder are changes where you want to atomically change both the C client, and also the Go client.

Precisely, that's the most common workflow when implementing new features.

What I mean is that none of the other clients require an updated `tb_client.h` to work, only the binding code generated during CI is sufficient to make any change in the tb_client lib visible to them. Except for the GO client, in which the `tb_client.h` file is part of the binding code required by CGO.

If we don't want to _proactively_ generate the header during `zig build scripts -- ci—-language=go`, we should at least _preventively_ check if the header is up to date, warning the user to regenerate it, avoiding footguns. Still, I'm in favor of reducing the number of things one must remember to perform a task.

matklad: Ah, I'd expect us to typically _not_ need atomic changes, so that "this is how we change C code, and that's the correspending header changes" and "here's how we adapt Go code to the new headers" are in separate commits, but it makes sense that sometimes its easier to do this in lock-step. 

I've added this test as a pre-requisite for go client build. This is a bit hacky, I think ideally we'd use the LazyPath infrastructure here, but I am not 100% sure where will this end up at the end of the larger build.zig refactor, so went with the simpler  solution for now!

matklad: Ah, that doesn't trivially work, as our tests are incompatible with `-D=production`, need to think about how to do this in a nicer way...

matklad:
@batiati dropped the generator->test change, while leaving the rest of the cleanups. While I do want to ultimately move this to a test (because the property that our Zig code is synchronized with non-Zig code in the repo is something we _should_ be testing), I realized that:

* I also want to do this for other clients, not only c
* I need to find a way to avoid devx regression here --- this test should be run automatically when you build a client. I'll want to revisit this, but _after_ the big build.zig refactor!

Sadly, this means that I had to drop the "synchronize once and for all" change, as having an extra compilation unit here (that is, both tests and the genrating binary) hits "file is imported by two modules" case in Zig...

Thanks for flagging the devx issues batiati , I wouldn't have noticed that myself!

batiati (approved):

PR #2033, opened 2024-06-19
build: orthonormalize the interface 

### LINKED ISSUE #1280: Rework build.zig and shorten our CI script
Our build.zig and .yml workflows accrued quite a bit of cruft. I haven't yet reached the critical mass of annoyance  to go and clean this all up, but I'd love to document some ideas of what could be improved:

1. Centralize few top-level steps for common workflows:

   * `zig build test` for unit-tests
   * `zig build check` for quick pre-commit checks (compile all binaries (without codegen if possible), check formatting, run tidy).
   * `zig build fuzz` as a unified frontend for fuzzing
   * `zig build benchmark` ditto for benches
   * `zig build ci` to run everything that CI does (so, all of the above), use that from .yml

2. Cut down on the number of build.zig targets altogether. Most notably, don't do a separate binary for each unit-fuzzer. Things like `repl_integration` also feel like they should be moved elsewhere, so that each new `xxxx_integration` does not require changing build.zig or linux.yml.
3. Consistently name targets: `simulator`, `simulator:build`, `fuzz`, `fuzz:build`
4. Just in general, shorten build.zig. We shouldn't need 1kloc of logic to build tigerbeetle, it is supposed to be simple&small! 
5. Revive mac and windows tests
6. Try to get rid of `builder_with_isolated_cache`
7. Use `-Drelease` / `-Ddebug` instead of `-Doptimize`
8. Get rid of `ldd: warning: you do not have execution permission for` warning (but, ideally, find non-hacky way to build java)

One important overarching goal: adding new things to TigerBeetle should not require changing build.zig or .yml. Those should be fixed, and new additions should only touch zig files.

matklad:
Refactor build.zig to make the interface easier to remember and the
implementation somewhat easier to follow.

Many step names changed to follow `step`/ `step:substep` pattern, with
top-level `step` being the primary interface.

Main steps:

* `zig build` builds TigerBeetle to `./src/tigerbeetle[.exe]`, no
  changes here.
* `zig build check` quickly check for compilation errors. No changes
  here, but now we can introduce `zig build vopr:check` in a clean way.
* `zig build run` run TigerBeetle, no changes here.

* `zig build scripts` -- free-form release automation, no changes here.

* `zig build test`, `zig build test:unit`, `zig build test:integration`,
  `zig build test:fmt` runs various tests.
* `zig build fuzz`, `zig build fuzz:build` run or just build minor fuzzers
* `zig build vopr`, `zig build vopr:build` run or just build the simulator

* `zig build client:{c,node,java,dotnet,go}` -- build corresponding
  native libraries.

Odd steps which I'd love to get rid off, but which don't really fit
anywhere else:

* `zig build aof` for running the aof utility
* `zig build clients:c:sample` for checking C sample
* `zig build test:jni` for running JNI tests. These have a dependency on
  JAVA_HOME, so they are _not_ being run by default.

The main implementation change is that all build steps and options are
pre-declared at the top of the file, so you can see the entire interface
at a glance (got the idea from https://ziggit.dev/t/build-system-tricks/3531)

Closes: https://github.com/tigerbeetle/tigerbeetle/issues/1280


matklad:
Closes the remaining 80% of https://github.com/tigerbeetle/tigerbeetle/issues/1280

build.zig
```diff
@@ -40,128 +40,151 @@ pub fn build(b: *std.Build) !void {
...
+        .aof = b.step("aof", "Run TigerBeetle AOF Utility"),
+        .check = b.step("check", "Check if TigerBeetle compiles"),
+        .clients_c = b.step("clients:c", "Build C client library"),
+        .clients_c_sample = b.step("clients:c:sample", "Build C client sample"),
+        .clients_dotnet = b.step("clients:dotnet", "Build dotnet client shared library"),
+        .clients_go = b.step("clients:go", "Build Go client shared library"),
+        .clients_java = b.step("clients:java", "Build Java client shared library"),
+        .clients_node = b.step("clients:node", "Build Node client shared library"),
+        .fuzz = b.step("fuzz", "Run non-VOPR fuzzers"),
+        .fuzz_build = b.step("fuzz:build", "Run non-VOPR fuzzers"),
```
batiati: s/run/build

batiati (approved):

PR #2034, opened 2024-06-20
fuzz: use cross-platform abstraction to get a random seed

matklad:


src/fuzz_tests.zig
```diff
@@ -104,12 +104,7 @@ fn main_smoke() !void {
 fn main_single(cli_args: CliArgs) !void {
     assert(cli_args.positional.fuzzer != .smoke);
 
-    const seed: usize = cli_args.positional.seed orelse seed: {
-        // If no seed was given, use a random seed instead.
-        var seed_random: u64 = undefined;
-        try std.posix.getrandom(std.mem.asBytes(&seed_random));
-        break :seed seed_random;
-    };
+    const seed = cli_args.positional.seed orelse std.crypto.random.int(u64);
```
matklad, line 107: love how there's no try here anymore: that's the right API for _secure_ random numbers, where, on failure, you just hang/crash the process.

kprotty (approved):

PR #2036, opened 2024-06-20
Fix build in test_min config

brson:
This fixes three compile-time divExact failures in the test_min configuration (`zig/zig build check -Dtest_min`).

Two of them are fixed with an ugly comptime conditional to change cli error messages from `MiB` to `KiB`.

The other is a change to the `clients_max` value from `4 + 3` to `4 + 4`. This prevents the following failure:

```zig
src/tigerbeetle/cli.zig:223:32: error: exact division produced remainder
    .limit_pipeline_requests = @divExact(constants.clients_max, 2) -
```

I don't know the significance of `4 + 3` being split into two values in this expression or whether 8 is an appropriate constant here.

### DIFF
--- a/src/config.zig
+++ b/src/config.zig
@@ -277,7 +277,7 @@ pub const configs = struct {
             .verify = true,
         },
         .cluster = .{
-            .clients_max = 4 + 4,
+            .clients_max = 4 + 3,
             .pipeline_prepare_queue_max = 4,
             .view_change_headers_suffix_max = 4 + 1,
             .journal_slot_count = Config.Cluster.journal_slot_count_min,
--- a/src/config.zig
+++ b/src/config.zig
@@ -277,7 +277,7 @@ pub const configs = struct {
             .verify = true,
         },
         .cluster = .{
-            .clients_max = 4 + 3,
+            .clients_max = 4 + 4,
             .pipeline_prepare_queue_max = 4,
             .view_change_headers_suffix_max = 4 + 1,
             .journal_slot_count = Config.Cluster.journal_slot_count_min,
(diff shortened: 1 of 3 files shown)


src/config.zig
```diff
@@ -277,7 +277,7 @@ pub const configs = struct {
             .verify = true,
         },
         .cluster = .{
-            .clients_max = 4 + 3,
+            .clients_max = 4 + 4,
```
sentientwaffle: :thinking: There's no particular reason to require the number of clients to be even... Instead, in `cli.zig`:

```
    .limit_pipeline_requests = vsr.stdx.div_ceil(constants.clients_max, 2) -|
        constants.pipeline_prepare_queue_max,
```

That will also fix the replica test failure.

brson:
Test failures don't look related.

src/tigerbeetle/cli.zig
```diff
@@ -496,11 +496,19 @@ pub fn parse_args(allocator: std.mem.Allocator, args_iterator: *std.process.ArgI
...
+                        request_size_limit.suffix(),
+                        @divExact(request_size_limit_max, 1024 * 1024),
+                    });
+                } else {
+                    flags.fatal("--limit-request: size {}{s} exceeds maximum: {}KiB", .{
+                        request_size_limit.value,
+                        request_size_limit.suffix(),
+                        @divExact(request_size_limit_max, 1024),
+                    });
+                }
```
sentientwaffle, line 511: Maybe this pattern warrants a formatter? Could be used elsewhere in this file for flexibility. Similar to `std.fmt.fmtIntSizeBin` except without the inexplicable floating-point conversion. (Using `@divExact`, no decimals).

src/config.zig
brson: I've done this.

sentientwaffle (approved):

PR #2037, opened 2024-06-22
Balancing transfers `amount` validation

batiati:
Removes the `amount` field from the idempotency check for balancing transfers with `amount=0`. It allows resubmitted balancing transfers to fail with `.exists` instead of `exists_with_different_amount`.

Also adds unit tests asserting the behavior of zeroed fields for balancing and two-phase transfers.

### DIFF
--- a/docs/reference/requests/create_transfers.md
+++ b/docs/reference/requests/create_transfers.md
@@ -284,8 +284,9 @@ A transfer with the same `id` already exists, but with a different
 A transfer with the same `id` already exists, but with a different
 [`amount`](../transfer.md#amount).
 
-If the transfer has `flags.balancing_debit` or `flags.balancing_credit` set, this error refers to
-the actual amount transferred, not the original (possibly higher) balancing amount.
+If the transfer has [`flags.balancing_debit`](../transfer.md#flagsbalancing_debit) or
+[`flags.balancing_debit`](../transfer.md#flagsbalancing_debit) set, then the actual amount
+transferred exceeded the maximum amount of the transfer in the request.
 
 ### `exists_with_different_pending_id`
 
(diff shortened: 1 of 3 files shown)


src/state_machine.zig
sentientwaffle, line 1: This new behavior should be documented in https://docs.tigerbeetle.com/reference/requests/create_transfers/#exists.

src/state_machine.zig
```diff
@@ -1572,20 +1572,32 @@ pub fn StateMachineType(
...
-            assert(t.pending_id == 0 and e.pending_id == 0); // We know that the flags are the same.
+            // If the accounts are the same, the ledger must be the same.
+            assert(t.ledger == e.ledger);
+
+            // Balancing transfers with `amount=0` must not raise `exists_with_different_amount`.
+            const balancing_transfer = t.amount == 0 and
+                (t.flags.balancing_debit or t.flags.balancing_credit);
+            if (t.amount != e.amount and !balancing_transfer) {
+                return .exists_with_different_amount;
+            }
```
sentientwaffle: Balancing transfers support nonzero amounts as well: a balancing transfer with nonzero amount `A` will move the largest possible amount `X` such that `0 < X ≤ A`. To be consistent, we should consider that case too.

What do you think of this approach?:
If the transfer is created with amount `X` (that is, it actually move `X` -- `A` may have been zero), and then retried with amount `Y`, then the retry should return:

- if `Y=0`: `exists`
- if `Y≥X`: `exists`
- if `Y<X`: `exists_with_different_amount`

docs/reference/requests/create_transfers.md
```diff
@@ -318,7 +319,14 @@ A transfer with the same `id` already exists, but with a different [`code`](../t
 
 ### `exists`
 
-A transfer with the same `id` already exists, and is identical to the transfer in the request.
+A transfer with the same `id` already exists.
+
+If the transfer has [`flags.balancing_debit`](../transfer.md#flagsbalancing_debit) or
+[`flags.balancing_debit`](../transfer.md#flagsbalancing_debit) set, then the existing
```
```suggestion
If the transfer has [`flags.balancing_debit`](../transfer.md#flagsbalancing_debit) or
[`flags.balancing_credit`](../transfer.md#flagsbalancing_credit) set, then the existing
```

docs/reference/requests/create_transfers.md
```diff
@@ -284,8 +284,9 @@ A transfer with the same `id` already exists, but with a different
 A transfer with the same `id` already exists, but with a different
 [`amount`](../transfer.md#amount).
 
-If the transfer has `flags.balancing_debit` or `flags.balancing_credit` set, this error refers to
-the actual amount transferred, not the original (possibly higher) balancing amount.
+If the transfer has [`flags.balancing_debit`](../transfer.md#flagsbalancing_debit) or
+[`flags.balancing_debit`](../transfer.md#flagsbalancing_debit) set, then the actual amount
+transferred exceeded the maximum amount of the transfer in the request.
```
```suggestion
transferred exceeds this failed transfer's `amount`.
```

sentientwaffle (commented):
A couple clarifications, but otherwise LGTM

sentientwaffle (approved):

PR #2041, opened 2024-06-24
Bring back shellcheck for `install_zig.sh`

cb22:
@matklad here's my delayed comment, in code form!

I don't think we should remove shellcheck for `install_zig.sh`: it might run on weird who-knows-what systems that we can't control, so we want to avoid something like a Steam `rm -rf` and shellcheck is an easy way to help us towards that goal. (the fact that we use it in CI ourselves gives me good enough confidence that it's tested!)

But, this file is really outside our build system entirely, and I'm in favour of removing it from _there_ - so I think it makes sense if the CI checks it directly.

While we're here, we might as well ban `.sh` entirely too :shrug:.

### DIFF
--- a/src/tidy.zig
+++ b/src/tidy.zig
@@ -262,16 +262,22 @@ test "tidy extensions" {
     const allowed_extensions = std.StaticStringMap(void).initComptime(.{
         .{".bat"}, .{".c"},     .{".cs"},   .{".csproj"},  .{".css"},  .{".go"},
         .{".h"},   .{".hcl"},   .{".java"}, .{".js"},      .{".json"}, .{".md"},
-        .{".mod"}, .{".props"}, .{".ps1"},  .{".service"}, .{".sh"},   .{".sln"},
-        .{".sum"}, .{".ts"},    .{".txt"},  .{".xml"},     .{".yml"},  .{".zig"},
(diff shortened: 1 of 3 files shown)


src/tidy.zig
```diff
@@ -262,16 +262,22 @@ test "tidy extensions" {
...
+        .{".gitignore"},                             .{".nojekyll"},
+        .{"CNAME"},                                  .{"Dockerfile"},
+        .{"exclude-pmd.properties"},                 .{"favicon.ico"},
+        .{"favicon.png"},                            .{"LICENSE"},
+        .{"module-info.test"},                       .{"index.html"},
+        .{"logo.svg"},                               .{"logo-white.svg"},
+        .{"logo-with-text-white.svg"},               .{"scripts/install_zig.sh"},
+        .{"src/scripts/cfo_supervisor.sh"},          .{"src/docs_website/scripts/build.sh"},
+        .{".github/ci/docs_check.sh"},               .{".github/ci/test_aof.sh"},
+        .{"tools/systemd/tigerbeetle-pre-start.sh"}, .{"tools/vscode/format_debug_server.sh"},
```
matklad, line 280: :heart:

matklad (approved):

PR #2044, opened 2024-06-25
build: use LazyPath for tb_client.h

marler8997:
Leverages LazyPath to propagate build dependencies for generating tb_client.h.  The advantage of using LazyPath is it automatically manages build step dependencies that use the lazy path.

In additon, this change adds a new InstallFile step based on the one from std.  InstallFile from std is only meant to install files to the install prefix, however, we want to install files to the repository itself, which our new step can do.  We use this to "install" the generated tb_client.h file to various locations.

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -236,8 +236,7 @@ pub fn build(b: *std.Build) !void {
         run_step.dependOn(&run_cmd.step);
     }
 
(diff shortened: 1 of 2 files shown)


matklad:
Thanks so much for this marler8997, both changes are great!

matklad (approved):

PR #2047, opened 2024-06-26
ci: simplify down to a single .yml

matklad:
This PR merges linux.yml, macos.yml, windows.yml, pipeline-main.yml and
pipeline-core.yml down to just ci.yml. With the result file being under
200 lines, it's much easier to understand what actually gets run when if
its all in one place!

I've also merged smaller steps like docs_check into larger
steps. This doesn't affect our overall CI times, as the
slowest part is client testing at the moment:

image

image


servusdei2018 (approved):

.github/workflows/ci.yml
```diff
@@ -0,0 +1,167 @@
+name: CI
+
+concurrency:
+  group: core-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+on:
+  merge_group:
+  pull_request:
+  workflow_call:
```
cb22, line 10: I think we can drop this

cb22 (approved):
One nit, otherwise LGTM! I'm approving in the mean time, so you can merge and confirm either way.

.github/workflows/ci.yml
matklad: I think it’s useful to have that in just in case you want to kick CI on main branch manually.

cb22: I think you'll want `workflow_dispatch` for that - `workflow_call` is to allow this to be called from other pipelines.

matklad: 🤦

PR #2048, opened 2024-06-26
enforce generated file synchronization

marler8997:
This adds an extra check "git diff --exit-code" to various workflow jobs that will cause them to fail if they cause any files tracked by git to be modified.  This will help detect and reject any commits that failed to synchronize the generated files commited to the repo.

Here's an example of a bad CI run that would result from forgetting to update the generated header files with this change: https://github.com/tigerbeetle/tigerbeetle/actions/runs/9683291713/job/26718376550?pr=2048

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -26,6 +26,7 @@ jobs:
           fetch-depth: 2147483647
       - run: ./scripts/install_zig.sh
       - run: zig/zig build test
+      - run: git diff --exit-code
 
   test_on_ubuntu:
     runs-on: ubuntu-latest
@@ -36,6 +37,7 @@ jobs:
       - run: ./scripts/install_zig.sh
       - run: zig/zig build test
       - run: zig/zig build -Dtracer-backend=tracy
+      - run: git diff --exit-code
 
   aof:
     runs-on: ubuntu-latest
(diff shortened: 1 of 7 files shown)


matklad:
If we do this, this should be part of build.zig: anything that's in yaml is next-to impossible to run outside of GHA, and we don't want to be tied to GHA. As a principle, we are moving logic out of yaml and into zig. 

And, even for build.zig, I'd love to find a solution that's independent of git, just to minimize dependencies. But maybe not? Currently some of our steps do create temporary files outside of .zig-cache, and, ideally, that shouldnt' be the case.

marler8997:
I agree with and would also encourage the points you made:

- put logic in build.zig, not yml
- decouple checks from git
- run checks locally, not just on the CI

That being said there's 2 points that might warrant an exception in this case.  First is that you can view the git commit itself as something to be tested.  If so, then it makes sense to leverage git to verify this independent asset.  I do agree that having a check independent of git would also be nice, and I'd advocate for having both.  The advantage of the git check is it would catch any bugs with tigerbeetle's implementation and the advantage of the tigerbeetle check is it can trivially be invoked without git, in particular before comitting your changes.

The second point is that this git diff check is only meant to be run "post commit".  Typically developers are expected to make changes, run tests, then commit.  If you have a check that only works "post commit", then you'd either have to change this workflow order, or duplicate testing, i.e.  make changes, commit, test   OR   make changes, test, commit, test again. In both workflows if the post commit tests fail, you need to roll back or amend the commit which is "awkward".

matklad:
We pickde this approach instead: https://github.com/tigerbeetle/tigerbeetle/pull/2052

PR #2052, opened 2024-06-26
build: add -Denforce-no-changes option

marler8997:
This option modifies the behavior of the build to enforce that any generated files committed to the repo are not changed.

This is another idea to catch bad commits that failed to update generated files committed to the repo (orignal idea: https://github.com/tigerbeetle/tigerbeetle/pull/2048).  The difference is that this one doesn't rely on git.  Instead, this one adds the option `-Denforce-no-changes` to `build.zig` to have it verify any generated files are already up-to-date instead of generating them.

Note that both solutions could also be used together.  This solution works without depending on git but also won't catch as many errors.  The git solution will catch any and all possible violations but also only works post commit.  I'd advocate for having both checks in place.

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -58,6 +58,19 @@ pub fn build(b: *std.Build) !void {
 
     const mode = b.standardOptimizeOption(.{ .preferred_optimize_mode = .ReleaseSafe });
     const emit_llvm_ir = b.option(bool, "emit-llvm-ir", "Emit LLVM IR (.ll file)") orelse false;
+    const enforce_no_changes = blk: {
+        if (b.option(
+            bool,
+            "enforce-no-changes",
+            "Enforce that generated files are not changed",
+        )) |v| break :blk v;
+        if (b.graph.env_map.get("CI")) |ci_env| {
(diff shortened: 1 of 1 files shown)


marler8997:
With this change, here's what a CI failure would look like if you forgot to update the generated header files:

https://github.com/tigerbeetle/tigerbeetle/actions/runs/9684269284/job/26721620818?pr=2052

image

matklad:
I like this approach! It should be easy to extend to our other genreated files, like dotnet_bindings and such. 

But I would suggest one change --- instead of a specific `-D` flag, that needs to be explicitly passed, lets look for a `CI` env variable. I find it a generally useful pattern to hide some checks behind "is_ci", and it helps when you don't need to specify this flag explicitly everywhere.

marler8997:
That's a great idea.  Looks like GitHub actions will set `CI=true`, so I've updated the PR to check for this and set this option to `true` when it's present.  Now we don't need to modify any of the `yml` files so I've removed those changes as well.

matklad:
LGTM, thanks!

matklad (approved):

PR #2056, opened 2024-06-28
test: make sure that TmpTigerBeetle prints stderr in case of an error

matklad:


src/testing/tmp_tigerbeetle.zig
```diff
@@ -106,12 +118,58 @@ pub fn init(
...
         .process = process,
     };
 }
 
-pub fn deinit(tb: *TmpTigerBeetle, gpa: std.mem.Allocator) void {
-    _ = gpa;
-    _ = tb.process.kill() catch unreachable;
+pub fn deinit(tb: *TmpTigerBeetle) void {
+    if (tb.stderr_reader) |reader| {
+        _ = tb.process.kill() catch unreachable;
```
sentientwaffle: :thinking: Why is this inside the branch?

sentientwaffle (commented):
I tested this by adding a bogus panic in `replica.zig`'s `on_request` handler. Then when I ran a CI (`zig build scripts -- ci --language=go`) it just hangs -- the client's message bus endlessly tries to reconnect since the replica has crashed. (Though there is no direct indication that the replication has failed in the logs, either.)

I think the issue is that we are waiting for `try shell.exec("go build main.go", .{});` to finish, but that has no idea that the replica failed already.

src/testing/tmp_tigerbeetle.zig
matklad: Due to the pattern of 

```
defer deinit()
errdefer log_stderr()
```

both functions can be called at the same time. To log stderr detreminstically, we need to kill the process and join the thread (an alternative would be to add a mutex and log "what we have so far", but that would give non-deterministic result). And it looks like
while `process.kill` is idempotent, `thread.join` is not (its UB to cal it twice). 

So I have to add my own idempotency on top, by using `tb.stderr_reader != null` condition to check the joining was done already. 


In genreal, the synchronization here is very tricky! I don't love this code, but I don't see anything substantionally simpler (what I'd _love_ to do is to just dump stderr to file, but Zig doesn't allowing setting child's stderr to a file descriptor).

sentientwaffle: Ahh I see. `tb.stderr_reader` is always non-null until either `deinit()` or `log_stderr` is called -- whichever one executes first also nullifies `stderr_reader` so that we don't try to kill the thread twice.

matklad:
>I tested this by adding a bogus panic in replica.zig's on_request handler. Then when I ran a CI (zig build scripts -- ci --language=go) it just hangs -- the client's message bus endlessly tries to reconnect since the replica has crashed. (Though there is no direct indication that the replication has failed in the logs, either.)

Yeah, there's this thing as well, which makes this significantly less useful. :thinking: I think the proper solution here would be async&cancelation, so that the user's code runs tmp_tigerbeetle and the actual test concurrently, and we use async magic to propagate tmp_beetle failure laterally to the testing "thread". Right now, the API tries to hide all concurrency from the user, but that means that it can't report such an asycnhroneous failure

src/testing/tmp_tigerbeetle.zig
```diff
@@ -106,12 +118,58 @@ pub fn init(
...
+        self.thread.join();
+        const buffer = self.buffer;
+        buffer.allocator.destroy(self);
+        return buffer;
+    }
+
+    fn thread_main(reader: *StreamReader) void {
+        // NB: don't use `readAllAlloc` to get partial output in case of errors.
+        reader.file.reader().readAllArrayList(&reader.buffer, 100 * 1024 * 1024) catch {};
+    }
```
sentientwaffle: What about something like:

```
    fn thread_main(reader: *StreamReader) void {
        var buf: [1024]u8 = undefined;
        while (true) {
            const line = reader.file.reader().readUntilDelimiter(&buf, '\n') catch break;
            if (!std.mem.startsWith(u8, line, "info") and
                !std.mem.startsWith(u8, line, "warning"))
            {
                std.debug.print("tmp_tigerbeetle: {s}\n", .{line});
            }
        }
    }
```

That way if tmp_tigerbeetle's `tigerbeetle start` panics, that isn't invisible when the test hangs trying to connect.

matklad:
:fearful: realized that this has a pretty bad bug --- I am moving the allocator to a separate thread, but Zig's allocators are not thread safe! This would be horrible to debug! But I think your suggestion actually simplifies everything here massively, let me try this

matklad:
Ok, seems to work sometimes:

```
λ ./zig/zig build scripts -- ci --language=node
$ /home/matklad/p/tb/work/zig/zig build node_client -Drelease -Dconfig=production

> tigerbeetle-node@0.12.0 prepare
> tsc


up to date in 3s
$ npm install

> tigerbeetle-node@0.12.0 prepare
> tsc


up to date, audited 4 packages in 2s

found 0 vulnerabilities
info: testing tests
$ node ./dist/test
error(tmptigerbeetle): tigerbeetle stderr:
++++
info(io): opening "0_0.tigerbeetle"...
warning(main): Grid cache size of 32MiB is small. See --cache-grid
info(main): release=0.0.1
info(main): git_commit=413fee8056fa83c2d349bae1537a60678a360c73
info(main): 0: Allocated 652MiB during replica init
info(main): 0: Grid cache: 32MiB, LSM-tree manifests: 128MiB
info(main): 0: cluster=0: listening on 127.0.0.1:33999
warning(main): 0: started with constants.verify - expect reduced performance. Recompile with -Dconfig=production if unexpected.
thread 128007 panic: oppp
/home/matklad/p/tb/work/src/tigerbeetle/main.zig:383:23: 0x124d1f3 in start (tigerbeetle)
            if (true) @panic("oppp");
                      ^
/home/matklad/p/tb/work/src/tigerbeetle/main.zig:63:44: 0x126b7e1 in main (tigerbeetle)
        .start => |*args| try Command.start(arena.allocator(), args),
                                           ^
/home/matklad/p/tb/work/zig/lib/std/start.zig:524:37: 0x1206e65 in posixCallMainAndExit (tigerbeetle)
                                    ^
/home/matklad/p/tb/work/zig/lib/std/start.zig:266:5: 0x1206981 in _start (tigerbeetle)
    ^
???:?:?: 0x4 in ??? (???)


++++
error(message_bus): error receiving from message_bus.MessageBusType(.client).Connection.Peer{ .replica = 0 }: error.ConnectionResetByPeer
error(message_bus): error connecting to replica 0: error.ConnectionRefused
error(message_bus): error connecting to replica 0: error.ConnectionRefused
error(message_bus): error connecting to replica 0: error.ConnectionRefused
```

But only sometimes --- this doesn't work when clients spawn a TigerBeetle process themselves, or when `zig test` helpful hides the output until the tests completes

sentientwaffle:
> this doesn't work when clients spawn a TigerBeetle process themselves

We are inconsistent about who is responsible for setting up TB. i.e. the node tests (iirc) are handled by tmp_tigerbeetle, whereas the golang tests set up their own TB.

> when zig test helpful hides the output until the tests completes

Yeah, not ideal... Though this shouldn't be an issue for most uses of tmp_tigerbeetle (the various CI's), right? It only affects the repl + benchmark-smoke tests.

matklad:
Right!

sentientwaffle (dismissed):

matklad:
Argh, there's another subtle bug here --- there's a race on stderr file descriptor!

Which I discovered the last time I tried to write this code and then forgot about it :facepalm: 

https://github.com/ziglang/zig/issues/16820

matklad:
Ready for review once again!

src/testing/tmp_tigerbeetle.zig
```diff
@@ -19,6 +19,12 @@ port_str: stdx.BoundedArray(u8, 8),
 
 tmp_dir: std.testing.TmpDir,
 
+// A separate thread for reading process stderr without blocking it. The process must be terminated
+// before stopping the StreamReader.
+//
+// StreamReader echoes process' stderr unless explicitly instructed otherwise. Because
```
sentientwaffle: "Because" is missing the explanation :wink:

matklad: Too bad I forgot it, as I no longer remeber what I wanted to write there, just removed the `Because`

sentientwaffle (approved):

PR #2058, opened 2024-06-28
Benchmark: Support "hot" accounts

sentientwaffle:
Add two new flags to the benchmark:

- `--account-count-hot` (default: `0`)
- `--transfer-hot-percent` (default: `100`)

So, the benchmark's default behavior is unchanged (since there are no hot accounts). But if `--account-count-hot=N`, then `N` of all of the (10k) accounts are "hot" -- `--transfer-hot-percent` (100% by default) of transfers pick one of the hot accounts as their debit account.

Note that this won't have much impact on throughput with only 10k accounts total, since a single table can fit almost 260k accounts.


src/tigerbeetle/benchmark_load.zig
```diff
@@ -276,11 +292,23 @@ const Benchmark = struct {
...
-                credit_account_index = (credit_account_index + 1) % b.account_count;
-            }
+            const account_count_cold = b.account_count - b.account_count_hot;
+
+            const debit_account_hot = b.account_count_hot > 0 and
+                random.uintLessThan(u64, 100) < b.transfer_hot_percent;
+            const debit_account_index = if (debit_account_hot)
+                random.uintLessThan(u64, b.account_count_hot)
+            else
+                random.uintLessThan(u64, account_count_cold) + b.account_count_hot;
```
matklad:
```suggestion
            const debit_account_index = if (debit_account_hot)
                random.uintLessThan(u64, b.account_count_hot)
            else
                random.uintLessThan(u64, account_count);
```

Otherwise it'll have a counter-intuitive behavior where cold accounts are actually hot ones, when `b.account_count_hot` is "almost all accounts" and `b.transfer_hot_percent` is "very low". 

It also feels more intuitive that non-hot transfer selects account at random.

batiati (approved):

PR #2060, opened 2024-07-01
README.md: Remove reference to --development flag

lilic:
This flag is not present in the latest release, which can be confusing if following the README quickstart section.

---
I could also understand the argument for keeping it in the README since its in the main branch. However, the quickstart section mentions fetching the latest released binary, which is what I did and how I got into the error:

```
error: unexpected argument: '--development'
```


sentientwaffle (approved):
Thanks!

PR #2068, opened 2024-07-02
test: fix flaky replica test

matklad:
In this test, the replica can either state-sync or fast-forward the log.
Right now, the choice between the two depends on timing:
`on_repair_sync` timeout might force the replica to state sync.

Fix this by completely stopping the replica, instead of just
partitioning it.

One alternative I've tried is preventing the replica from learning about
the new checkpoint by dropping commits and pings, but that _also_
prevents log forwariding from working.

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -286,6 +286,9 @@ pub fn build(b: *std.Build) !void {
 
         const run_unit_tests = b.addRunArtifact(unit_tests);
         run_unit_tests.setEnvironmentVariable("ZIG_EXE", b.graph.zig_exe);
+        if (b.args != null) { // Don't cache test results if running a specific test.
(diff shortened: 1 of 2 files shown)


build.zig
```diff
@@ -286,6 +286,9 @@ pub fn build(b: *std.Build) !void {
 
         const run_unit_tests = b.addRunArtifact(unit_tests);
         run_unit_tests.setEnvironmentVariable("ZIG_EXE", b.graph.zig_exe);
+        if (b.args != null) { // Don't cache test results if running a specific test.
+            run_unit_tests.has_side_effects = true;
+        }
```
sentientwaffle, line 291: :heart:

sentientwaffle (approved):

PR #2071, opened 2024-07-04
cfo: make cfo supervisor robust to network errors

matklad:
git clone and install_zig touch the network and can fail transitively. This should not kill the supervisor!

### DIFF
--- a/src/scripts/cfo_supervisor.sh
+++ b/src/scripts/cfo_supervisor.sh
@@ -11,17 +11,12 @@ git --version
 while true
 do
     rm -rf ./tigerbeetle
-    git clone https://github.com/tigerbeetle/tigerbeetle tigerbeetle
-    cd tigerbeetle
-
-    ./scripts/install_zig.sh
-
-    # `unshare --pid` ensures that, if the parent process dies, all children die as well.
-    # `unshare --user` is needed to make `--pid` work without root.
-    # `|| true` because we want to be resilient to bugs in `cfo` itself.
-    unshare --user -f --pid \
-        ./zig/zig build -Drelease scripts -- cfo \
-        || true
-
-    cd ../
+    (
+        git clone https://github.com/tigerbeetle/tigerbeetle tigerbeetle
+        cd tigerbeetle
+        ./scripts/install_zig.sh
+        # `unshare --pid` ensures that, if the parent process dies, all children die as well.
+        # `unshare --user` is needed to make `--pid` work without root.
+        unshare --user -f --pid ./zig/zig build -Drelease scripts -- cfo
+    ) || sleep 10 # Be resilient to cfo bugs and network errors, but avoid busy-loop retries.
 done


src/scripts/cfo_supervisor.sh
```diff
@@ -11,10 +11,10 @@ git --version
 while true
 do
     rm -rf ./tigerbeetle
-    git clone https://github.com/tigerbeetle/tigerbeetle tigerbeetle
+    git clone https://github.com/tigerbeetle/tigerbeetle tigerbeetle || true
```
sentientwaffle: If `git clone`  fails, won't `cd tigerbeetle` also fail?

cb22: Good point. Maybe we should only `set -u`? If we're adding `|| true` everywhere it's the same!

sentientwaffle: Is it possible that `cd tigerbeetle` would fail, and then `cd ../` would succeed, though?

cb22: (and replace the relative `cd` with absolute paths)

We could also keep it top level, then `set +e` at the start of the loop and `set -e` at the end (I think :D).

And perhaps a `sleep 1`...

cb22: Although. There's no reason we can't `scp` a zig binary to the machine :upside_down_face:

matklad: Ah, good point, that cd was a bad idea from the start, no need to mutate shell's global state in the control loop

matklad: Added a sleep 1 for a good measure --- indeed seems nice to protect from runaway CPU loops.

I'd rather keep `-e` and stick with `|| true` though, I do like ignoring only _specific_ errors, even if it's most of them

matklad: >Although. There's no reason we can't scp a zig binary to the machine 🙃

That's a very subversive thought, I'll entertain it for a while! One benefit of the script is that it's easier to vi it in place just to try stuff out...

But, yeah, if at any point we move to a proper infra-as-code here, I agree that the boostrap seed should be a Zig binary.

matklad: swapped `|| sleep 10` for  `|| true` which does a larger timeout, but only when we need one!

matklad: Ok, so it seems that `( loop body ) || sleep` is actually the good shape for an sh control loop. 

I _think_ I am done with forecpushes here, but let me `sleep 600` to make sure :rofl:

src/scripts/cfo_supervisor.sh
```diff
@@ -11,17 +11,13 @@ git --version
...
-        || true
-
-    cd ../
+    (
+        git clone https://github.com/tigerbeetle/tigerbeetle tigerbeetle
+        cd tigerbeetle
+        ./scripts/install_zig.sh
+        # `unshare --pid` ensures that, if the parent process dies, all children die as well.
+        # `unshare --user` is needed to make `--pid` work without root.
+        # `|| true` because we want to be resilient to bugs in `cfo` itself.
```
cb22: One more force push :smile: - needs to be updated to reflect the `sleep 10` and maybe pulled to the top (or just before the sleep)

cb22 (approved):

PR #2072, opened 2024-07-04
test: fix deterministic flakes 

matklad:
Update itself might need to pump unknown number of ops, so we can move
past the second checkpoint he

### DIFF
--- a/src/vsr/replica_test.zig
+++ b/src/vsr/replica_test.zig
@@ -1297,7 +1297,7 @@ test "Cluster: upgrade: R=1" {
 }
 
 test "Cluster: upgrade: state-sync to new release" {
-    const t = try TestContext.init(.{ .replica_count = 3, .seed = 2 });
+    const t = try TestContext.init(.{ .replica_count = 3 });
     defer t.deinit();
 
     var c = t.clients(0, t.cluster.clients.len);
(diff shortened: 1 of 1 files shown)


src/vsr/replica_test.zig
```diff
@@ -3,7 +3,7 @@ const assert = std.debug.assert;
 const maybe = stdx.maybe;
 const log = std.log.scoped(.test_replica);
 const expectEqual = std.testing.expectEqual;
-const expect = std.testing.expectEqual;
+const expect = std.testing.expect;
```
matklad, line 6: This one made me soo confused. Reminded me of this today's post https://avva.livejournal.com/3669085.html (might need google translate, but the code is self-explanatory)

src/vsr/replica_test.zig
```diff
@@ -1297,15 +1297,18 @@ test "Cluster: upgrade: R=1" {
...
 
     var c = t.clients(0, t.cluster.clients.len);
 
     t.replica(.R_).stop();
     try t.replica(.R0).open_upgrade(&[_]u8{ 1, 2 });
     try t.replica(.R1).open_upgrade(&[_]u8{ 1, 2 });
-    try c.request(checkpoint_2_trigger, checkpoint_2_trigger);
+    t.run();
+    try expectEqual(t.replica(.R0).commit(), checkpoint_1_trigger);
+    try c.request(constants.vsr_checkpoint_interval, constants.vsr_checkpoint_interval);
```
matklad, line 1310: I am wondering whether we should change the semantic of `request` to "advancing to op n" rather than "doing n requests"?

My gut feeling is yes, but curious what you'd think here.

sentientwaffle: :thinking: Overall, I think that makes sense -- that is, it fits how we actually write tests.

The original idea of `var c = t.clients(0, t.cluster.clients.len);` was that there could be multiple "pools" of clients which have their requests managed independently. This approach makes more sense with "do n requests" rather than "advance to op n"... But we don't ever use multiple pools right now. (Though maybe we will soon for eviction tests? But probably not worth designing our API around them.)

sentientwaffle (approved):

src/vsr/replica_test.zig
matklad: I think I’ll keep this on the stove for now, and revisit the next time I hit the problem!

PR #2073, opened 2024-07-04
Tidy two lsm files

brson:
Mostly straighforward except the first commit: generate_fuzz_op contained a long line that was seemingly not possible to shorten locally in a way that read better than the existing expression, so I extracted the generate_fuzz_op_action function.

The resulting code is perhaps better for having shorter functions, but the extracted function signature is pretty ugly: worst is that I couldn't figure out how to name the HashMap type this function uses so I just provided it as _some_ comptime type and called the hashmap methods on it.

### DIFF
--- a/src/lsm/forest_fuzz.zig
+++ b/src/lsm/forest_fuzz.zig
@@ -792,60 +792,24 @@ pub fn generate_fuzz_ops(random: std.rand.Random, fuzz_op_count: usize) ![]const
             // Otherwise pick a random FuzzOp.
             fuzz.random_enum(random, FuzzOpActionTag, action_distribution);
         const action = switch (action_tag) {
-            .compact => compact: {
-                const compact_op = op;
+            .compact => action: {
+                const action = generate_compact(random, .{
+                    .op = op,
+                    .persisted_op = persisted_op,
+                });
                 op += 1;
-                const checkpoint =
-                    // Can only checkpoint on the last beat of the bar.
-                    compact_op % constants.lsm_batch_multiple == constants.lsm_batch_multiple - 1 and
-                    compact_op > constants.lsm_batch_multiple and
-                    // Never checkpoint at the same op twice
-                    compact_op > persisted_op + constants.lsm_batch_multiple and
-                    // Checkpoint at roughly the same rate as log wraparound.
-                    random.uintLessThan(usize, Environment.compacts_per_checkpoint) == 0;
-                if (checkpoint) {
+                if (action.compact.checkpoint) {
                     persisted_op = op - constants.lsm_batch_multiple;
                 }
-                break :compact FuzzOpAction{
-                    .compact = .{
-                        .op = compact_op,
-                        .checkpoint = checkpoint,
-                    },
-                };
+                break :action action;
             },
-            .put_account => put_account: {
-                const id = random_id(random, u128);
-                var account = id_to_account.get(id) orelse Account{
-                    .id = id,
-                    // `timestamp` must be unique.
-                    .timestamp = fuzz_op_index,
-                    .user_data_128 = random_id(random, u128),
-                    .user_data_64 = random_id(random, u64),
-                    .user_data_32 = random_id(random, u32),
-                    .reserved = 0,
-                    .ledger = random_id(random, u32),
-                    .code = random_id(random, u16),
-                    .flags = .{
-                        .debits_must_not_exceed_credits = random.boolean(),
-                        .credits_must_not_exceed_debits = random.boolean(),
-                    },
-                    .debits_pending = 0,
-                    .debits_posted = 0,
-                    .credits_pending = 0,
-                    .credits_posted = 0,
-                };
-
-                // These are the only fields we are allowed to change on existing accounts.
-                account.debits_pending = random.int(u64);
-                account.debits_posted = random.int(u64);
-                account.credits_pending = random.int(u64);
-                account.credits_posted = random.int(u64);
-
-                try id_to_account.put(account.id, account);
-                break :put_account FuzzOpAction{ .put_account = .{
+            .put_account => action: {
+                const action = generate_put_account(random, &id_to_account, .{
                     .op = op,
-                    .account = account,
-                } };
+                    .timestamp = fuzz_op_index,
+                });
+                try id_to_account.put(action.put_account.account.id, action.put_account.account);
+                break :action action;
             },
             .get_account => FuzzOpAction{ .get_account = random_id(random, u128) },
             .scan_account => blk: {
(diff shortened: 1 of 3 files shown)


src/lsm/forest_fuzz.zig
```diff
@@ -772,99 +784,22 @@ pub fn generate_fuzz_ops(random: std.rand.Random, fuzz_op_count: usize) ![]const
...
-        };
+        const action = try generate_fuzz_op_action(
+            random,
+            std.hash_map.AutoHashMap(u128, Account),
+            &id_to_account,
+            action_tag,
+            fuzz_op_index,
+            &op,
+            &persisted_op,
+        );
```
matklad: Looking at the wider context, it indeed seems like the code is rather firmly telling us that we should split it up into more functions already 

At the same time, I think mechanically extracting the whole `generate_fuzz_op_action` is not perhaps the best factoring here. 

I have these personal rules of thumbs that might be helpful here:

* when splitting code, keep the core control flow in the root function, and streamline the leaves. So the split is that caller cares _only_ about control flow, and callee cares about everything else (aka, [push ifs up](https://matklad.github.io/2023/11/15/push-ifs-up-and-fors-down.html))
* Similarly, keep mutation isolated to the root function, and keep leaves pure. Zig is much worse than Rust at keeping track of pointers and aliasing, so we in general try to avoid using them. And here, because we extract mutation, we are forced into passing `*u64` in, which might be easy to get wrong without borrow checker.

I guess in this case we get an extra smell: we extract the switch, so we have to use the _union_ of what each separate prong needs as function parameters, so we have a whole bunch of them, but we only use a few at a time. 

Maybe something like this would work better here?

```diff
diff --git a/src/lsm/forest_fuzz.zig b/src/lsm/forest_fuzz.zig
index 3a9a8bd15..316c7a4ee 100644
--- a/src/lsm/forest_fuzz.zig
+++ b/src/lsm/forest_fuzz.zig
@@ -779,60 +779,24 @@ pub fn generate_fuzz_ops(random: std.rand.Random, fuzz_op_count: usize) ![]const
             // Otherwise pick a random FuzzOp.
             fuzz.random_enum(random, FuzzOpActionTag, action_distribution);
         const action = switch (action_tag) {
-            .compact => compact: {
-                const compact_op = op;
+            .compact => action: {
+                const action = generate_compact(random, .{
+                    .op = op,
+                    .persisted_op = persisted_op,
+                });
                 op += 1;
-                const checkpoint =
-                    // Can only checkpoint on the last beat of the bar.
-                    compact_op % constants.lsm_batch_multiple == constants.lsm_batch_multiple - 1 and
-                    compact_op > constants.lsm_batch_multiple and
-                    // Never checkpoint at the same op twice
-                    compact_op > persisted_op + constants.lsm_batch_multiple and
-                    // Checkpoint at roughly the same rate as log wraparound.
-                    random.uintLessThan(usize, Environment.compacts_per_checkpoint) == 0;
-                if (checkpoint) {
+                if (action.compact.checkpoint) {
                     persisted_op = op - constants.lsm_batch_multiple;
                 }
-                break :compact FuzzOpAction{
-                    .compact = .{
-                        .op = compact_op,
-                        .checkpoint = checkpoint,
-                    },
-                };
+                break :action action;
             },
-            .put_account => put_account: {
-                const id = random_id(random, u128);
-                var account = id_to_account.get(id) orelse Account{
-                    .id = id,
-                    // `timestamp` must be unique.
-                    .timestamp = fuzz_op_index,
-                    .user_data_128 = random_id(random, u128),
-                    .user_data_64 = random_id(random, u64),
-                    .user_data_32 = random_id(random, u32),
-                    .reserved = 0,
-                    .ledger = random_id(random, u32),
-                    .code = random_id(random, u16),
-                    .flags = .{
-                        .debits_must_not_exceed_credits = random.boolean(),
-                        .credits_must_not_exceed_debits = random.boolean(),
-                    },
-                    .debits_pending = 0,
-                    .debits_posted = 0,
-                    .credits_pending = 0,
-                    .credits_posted = 0,
-                };
-
-                // These are the only fields we are allowed to change on existing accounts.
-                account.debits_pending = random.int(u64);
-                account.debits_posted = random.int(u64);
-                account.credits_pending = random.int(u64);
-                account.credits_posted = random.int(u64);
-
-                try id_to_account.put(account.id, account);
-                break :put_account FuzzOpAction{ .put_account = .{
+            .put_account => action: {
+                const action = generate_put_account(random, &id_to_account, .{
                     .op = op,
-                    .account = account,
-                } };
+                    .timestamp = fuzz_op_index,
+                });
+                try id_to_account.put(action.put_account.account.id, action.put_account.account);
+                break :action action;
             },
             .get_account => FuzzOpAction{ .get_account = random_id(random, u128) },
             .scan_account => blk: {
@@ -907,6 +871,61 @@ pub fn generate_fuzz_ops(random: std.rand.Random, fuzz_op_count: usize) ![]const
     return fuzz_ops;
 }
 
+fn generate_compact(
+    random: std.rand.Random,
+    options: struct { op: u64, persisted_op: u64 },
+) FuzzOpAction {
+    const checkpoint =
+        // Can only checkpoint on the last beat of the bar.
+        options.op % constants.lsm_batch_multiple == constants.lsm_batch_multiple - 1 and
+        options.op > constants.lsm_batch_multiple and
+        // Never checkpoint at the same op twice
+        options.op > options.persisted_op + constants.lsm_batch_multiple and
+        // Checkpoint at roughly the same rate as log wraparound.
+        random.uintLessThan(usize, Environment.compacts_per_checkpoint) == 0;
+    return FuzzOpAction{ .compact = .{
+        .op = options.op,
+        .checkpoint = checkpoint,
+    } };
+}
+
+fn generate_put_account(
+    random: std.rand.Random,
+    id_to_account: *const std.AutoHashMap(u128, Account),
+    options: struct { op: u64, timestamp: u64 },
+) FuzzOpAction {
+    const id = random_id(random, u128);
+    var account = id_to_account.get(id) orelse Account{
+        .id = id,
+        // `timestamp` must be unique.
+        .timestamp = options.timestamp,
+        .user_data_128 = random_id(random, u128),
+        .user_data_64 = random_id(random, u64),
+        .user_data_32 = random_id(random, u32),
+        .reserved = 0,
+        .ledger = random_id(random, u32),
+        .code = random_id(random, u16),
+        .flags = .{
+            .debits_must_not_exceed_credits = random.boolean(),
+            .credits_must_not_exceed_debits = random.boolean(),
+        },
+        .debits_pending = 0,
+        .debits_posted = 0,
+        .credits_pending = 0,
+        .credits_posted = 0,
+    };
+
+    // These are the only fields we are allowed to change on existing accounts.
+    account.debits_pending = random.int(u64);
+    account.debits_posted = random.int(u64);
+    account.credits_pending = random.int(u64);
+    account.credits_posted = random.int(u64);
+    return FuzzOpAction{ .put_account = .{
+        .op = options.op,
+        .account = account,
+    } };
+}
+
 const io_latency_mean = 20;
 
 pub fn main(fuzz_args: fuzz.FuzzArgs) !void {
```


keep the switch and state manipulation in this function, but move the actual data generation out of the way?

src/lsm/groove.zig
```diff
@@ -174,8 +174,11 @@ pub fn GrooveType(
         // See if we should ignore this field from the options.
         //
         // By default, we ignore the "timestamp" field since it's a special identifier.
-        // Since the "timestamp" is ignored by default, it shouldn't be provided in groove_options.ignored.
-        comptime var ignored = mem.eql(u8, field.name, "timestamp") or mem.eql(u8, field.name, "id");
+        // Since the "timestamp" is ignored by default, it shouldn't be provided
+        // in groove_options.ignored.
+        comptime var ignored = mem.eql(u8, field.name, "timestamp") or brk: {
+            break :brk mem.eql(u8, field.name, "id");
+        };
```
matklad: I'd maybe just 

```
        comptime var ignored =
            mem.eql(u8, field.name, "timestamp") or mem.eql(u8, field.name, "id");
```

brson: Thanks, I didn't know fmt would preserve a line break there.

src/lsm/forest_fuzz.zig
brson: I applied this patch as written. Thanks!

matklad (approved):

PR #2077, opened 2024-07-05
vsr: ensure flexible quorums actually work

matklad:
Change the Clock to use replication_quorum, rather than simple majority.

Clocks were implemented before flexible quorum, and, with simple
majority, they'll actually prevent a flexible quorum from committing!

Thanks to dj for writing the test here!


matklad:
Note: making sure that this is discoverable by VOPR liveness mode is TODO

sentientwaffle (approved):

PR #1821, opened 2024-03-29
VSR/Journal: Recover to decision=fix for some ops before checkpoint

sentientwaffle:
Fixes https://github.com/tigerbeetle/tigerbeetle/issues/1787.

## Background

In https://github.com/tigerbeetle/tigerbeetle/issues/1787, `Simulator.restart_replica` crashes when a replica recovers into `status=recovering_head` unexpectedly:

1. Replica 6 state syncs a new superblock.
2. Replica 6 crashes. (And is down for a long time, until eventually...)
3. ...`Simulator.transition_to_liveness_mode()`: replica 6 restarts with faults disabled.
4. Replica 6 recovers its journal. Slot 2 recovers to case `@L` (`decision=vsr`), which is: both prepare and header are valid and have the same `op`, but they have different view numbers.
5. `!op_head_certain` → `status=recovering_head`.

Both versions of the `@L` header in the WAL are for the same op number, and that op precedes the replica's `op_checkpoint`.

Which means:
- the op is definitely not `op_head`, and
- the replica will never need to replay that op.

## Fix

Rename WAL recovery case `@M` to `@N`.

Split WAL recovery case `@L` into `@L` and `@M`:

- `@L` is the same as before, with the added condition that `header.op >= op_checkpoint`. `decision=vsr`.
- `@N` is the same as `@L`, except with the condition `header.op < op_checkpoint`. `decision=fix`.

PR #2078, opened 2024-07-05
vsr: don't crash in op_head_certain after state sync

matklad:
The seed here crashes at

```
// Immediately after recovery, any faulty non-reserved prepares must be within our 
// current checkpoint. See recovery case `@M`.
```

What happens here is that the statement is true after we recover the journal from disk, but not after we used our vsr headers to further path the journal.

Specifically, journal marks op=2 (which is outside of the checkpoint) as case `@E`, and reserves it. But then we go and repair a vsr_headers header over this slot. So it is:

* faulty
* dirty
* non-reserved
* in the past checkpoint

I think this can only happen after state sync, when our op lags behind checkpoint.

Seed: ./zig/zig build  vopr -- --lite 16548263080520932250

### DIFF
--- a/src/vsr/replica.zig
+++ b/src/vsr/replica.zig
@@ -639,6 +639,11 @@ pub fn ReplicaType(
             self.opened = false;
             self.journal.recover(journal_recover_callback);
             while (!self.opened) self.superblock.storage.tick();
+            for (self.journal.headers, 0..constants.journal_slot_count) |*header, slot| {
+                if (self.journal.faulty.bit(.{ .index = slot })) {
+                    assert(header.operation == .reserved);
+                }
+            }
 
             // Abort if all slots are faulty, since something is very wrong.
             if (self.journal.faulty.count == constants.journal_slot_count) return error.WALInvalid;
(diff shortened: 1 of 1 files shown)


matklad:
>I think this can only happen after state sync, when our op lags behind checkpoint.

Will revisit this the next week. While this diff does fix the seed, I am not entirely sure that this is sufficient. 

Could it be the case that `op_head > checkpoint`, but there's still a header we install from `vsr_headers`? But if that is the case, I think it could _also_ be the case that that faulty header is actually newer than the one from vsr_headers, in which case overwriting it might be incorrect?

matklad:
@sentientwaffle I realize that I no longer understand the assert in question:

https://github.com/tigerbeetle/tigerbeetle/pull/1821/commits/8acda4e94e493c3964bd4bd3b55f703c8ed8323e

Here's what I understand:

In `recover_journal`, the replica looks at the prepares and headers in the journal, and decides on the logical state of the journal, given the fault model. Here, the replica compares prepare's op to the checkpoint, to disambiguate between `@L` and `@M`. Replica also overwrites any faulty headers with `.reserved`. 

Subsequently, in `open`, replica adds new heads to the journal using vsr headers. Here, we don't look at the checkpoint. 

And then we assert that 

>any faulty non-reserved prepares must be within our current checkpoint

Which doesn't make sense to me. After `recovery_journal`, a stronger statement is true: all faulty prepares are reserved. 

After we additionally install VSR headers, we might get some faulty non-resreved ones, but this is orthogonal to checkpointing. 

What would makes sense is the following pair of asserts:

* In `recover_slots`, check that `.vsr` is only for the current checkpoint
* In `open`, before installing vsr headers, verify that all faultys are reserved. 

But maybe I am missing what the original assert was trying to enforce?

sentientwaffle (approved):
> But maybe I am missing what the original assert was trying to enforce?

My first thought was that maybe this assert predates [prepare beyond checkpoint](https://github.com/tigerbeetle/tigerbeetle/pull/1508), but that is not the case.

Agreed that this version makes more sense (and is simpler).

matklad:
>In open, before installing vsr headers, verify that all faultys are reserved.

Did this

>In recover_slots, check that .vsr is only for the current checkpoint

Didn't actually do this, because there are non-@M cases where this can be false. To choose the case, we look at prepare.header.op, but in `recover_slot` we have `reduntant_header.op`, which might actually be past the checkpoint. I _think_ asserting that propertly would amount to duplciating the actual recovering logic, so that's why I have left only the first assert.

PR #2080, opened 2024-07-05
New operations query_{accounts, transfers}

batiati:
## New operations `query_{accounts, transfers}`

Implements basic query capabilities for fetching `Accounts` and `Transfers` by some secondary indexes and timestamp range.

### The secondary indexes queryable by this API:

- `user_data_128`
- `user_data_64`
- `user_data_32`
- `code`
- `ledger`

Note that indexes are optional and orthogonal:
- When the query specifies no index, all objects are returned.
- When the query specifies one or multiple indexes, only objects that match all fields (intersection) are returned.

Those fields are common between `Account`s and `Transfer`s and strongly associated with external entities, making them highly relevant indexes in user-defined queries.

Also, since the current API is limited to returning the intersection of indexes, it's better to keep it simple and not to include indexes that may require more complex semantics to query (for example `{debit,credit}_account_id` may require `OR` instead of `AND`, and `pending_id` may require `IS NULL`/`IS NOT NULL`). 
**More complex queries will be allowed in the future by the proper query API.** 

### The timestamp range

- `timestamp_min`
- `timestamp_max`

The timestamp range is optional and orthogonal with indexes.
Combined with the `limit` field, it can be used as a pagination mechanism similar to `get_account_transfers` and `get_account_balances`.

One of the goals of this API was to provide a `SELECT *` like capability, allowing the user to export data out of TigerBeetle based on the last object's timestamp that was already exported. 

Example in pseudo-SQL:
`SELECT * FROM Transfers WHERE timestamp > $last_timestamp_exported LIMIT $batch_size`

## This PR

**Best reviewed per commit.**

- State machine logic and unit tests.
- REPL.
- VOPR.
- Clients (including README and examples).
- Documentation.

Depends on scans being fuzzed by #2028


src/state_machine.zig
```diff
@@ -62,6 +63,8 @@ pub fn StateMachineType(
                 pub const lookup_transfers = operation_batch_max(.lookup_transfers, config.message_body_size_max);
                 pub const get_account_transfers = operation_batch_max(.get_account_transfers, config.message_body_size_max);
                 pub const get_account_balances = operation_batch_max(.get_account_balances, config.message_body_size_max);
+                pub const query_accounts = operation_batch_max(.query_accounts, config.message_body_size_max);
+                pub const query_transfers = operation_batch_max(.query_transfers, config.message_body_size_max);
```
sentientwaffle: :thinking: I think `operation_batch_max` is only accidentally correct for the `get_account_*` and `query_*` operations, since their `@sizeOf(Event) <= @sizeOf(Result)`:

```
        pub fn operation_batch_max(comptime operation: Operation, batch_size_limit: u32) u32 {
            assert(batch_size_limit <= constants.message_body_size_max);

            const event_size = @sizeOf(Event(operation));
            const result_size = @sizeOf(Result(operation));
            comptime assert(event_size > 0);
            comptime assert(result_size > 0);

            return @min(
                @divFloor(batch_size_limit, event_size),
                @divFloor(constants.message_body_size_max, result_size),
            );
        }
```

Since `get_account_transfers`/`get_account_balances`/`query_accounts`/`query_transfers` don't have automatic batching, we don't actually care about `@divFloor(batch_size_limit, event_size)`.

Probably `operation_batch_max()` should take `batch_logical_allowed` into account?

src/state_machine.zig
```diff
@@ -985,6 +1020,154 @@ pub fn StateMachineType(
...
+                .min = if (filter.timestamp_min == 0)
+                    TimestampRange.timestamp_min
+                else
+                    filter.timestamp_min,
+
+                .max = if (filter.timestamp_max == 0)
+                    TimestampRange.timestamp_max
+                else
+                    filter.timestamp_max,
+            };
```
sentientwaffle, line 1176: `assert(timestamp_range.min <= timestamp_range.max);`

src/state_machine.zig
```diff
@@ -985,6 +1020,154 @@ pub fn StateMachineType(
...
+                .user_data_64,
+                .user_data_32,
+                .ledger,
+                .code,
+            };
+            comptime assert(indexes.len <= global_constants.lsm_scans_max);
+
+            var scan_conditions: stdx.BoundedArray(*Groove.ScanBuilder.Scan, indexes.len) = .{};
+            inline for (indexes) |index| {
+                if (@field(filter, @tagName(index)) != 0) {
```
sentientwaffle, line 1190: In `docs/reference/transfers.md`/`accounts.md` we should document that only nonzero `user_data_*` values can be used as query filters.

src/state_machine.zig
```diff
@@ -1113,9 +1133,19 @@ pub fn StateMachineType(
             comptime Groove: type,
             groove: *Groove,
             filter: QueryFilter,
-        ) *Groove.ScanBuilder.Scan {
+        ) ?*Groove.ScanBuilder.Scan {
             assert(self.forest.scan_buffer_pool.scan_buffer_used == 0);
 
+            const filter_valid =
+                filter.timestamp_min != std.math.maxInt(u64) and
+                filter.timestamp_max != std.math.maxInt(u64) and
```
sentientwaffle, line 1157: Can you remind me why we don't allow `maxInt(u64)`? (For query and accountfilter)

src/clients/go/tb_client.go
```diff
@@ -392,6 +405,48 @@ func (c *c_client) GetAccountBalances(filter types.AccountFilter) ([]types.Accou
 	return results[0:resultCount], nil
 }
 
+func (c *c_client) QueryAccounts(filter types.QueryFilter) ([]types.Account, error) {
+	//TODO(batiati): we need to expose the max message size to the client.
```
sentientwaffle, line 409: Or rather, the batch limit size (as returned on the register reply).
The batch limit may be less than the maximum message size.

src/state_machine/auditor.zig
```diff
@@ -179,6 +203,21 @@ pub const AccountingAuditor = struct {
...
+        const query_intersections = try allocator.alloc(
+            QueryIntersection,
+            options.accounts_max / 2,
+        );
+        errdefer allocator.free(query_intersections);
+        for (query_intersections, 1..) |*query_intersection, index| {
+            query_intersection.* = .{
+                .user_data_64 = @truncate(index * 1_000_000),
+                .user_data_32 = @truncate(index * 1_000),
+                .code = @truncate(index), // will be used to recover the index
```
sentientwaffle: Should this be `@intCast`? Otherwise if we are truncating bits, how would we recover the index?

src/state_machine/workload.zig
```diff
@@ -616,25 +636,79 @@ pub fn WorkloadType(comptime AccountingStateMachine: type) type {
...
+                    // Excluding the first or last object:
+                    if (query_filter.flags.reversed) {
+                        query_filter.timestamp_min = state.timestamp_min;
+                        query_filter.timestamp_max = state.timestamp_max - 1;
+                    } else {
+                        query_filter.timestamp_min = state.timestamp_min + 1;
+                        query_filter.timestamp_max = state.timestamp_max;
+                    }
+                    // Later we can assert that results.len == count - 1:
+                    query_filter.limit = state.count;
```
sentientwaffle, line 726: Clever!

src/state_machine.zig
batiati: Now that you mentioned it, I think we have discussed it before, it should be `maxInt(u63)` instead to not conflict with the tombstones.

I guess `maxInt(64) - 1` was wrongly based on `snapshot_latest`:
https://github.com/tigerbeetle/tigerbeetle/blob/f8cfbd5b527195432ab7992b50dfc63682b99463/src/lsm/tree.zig#L28-L33

sentientwaffle: Ah right, the tombstone bit is or'd into the timestamp of composite keys.

batiati: I'll change them to `maxInt(u63)` in a follow-up PR.

src/state_machine.zig
batiati: That's true, however, can we handle that in another PR?

I have some ideas on how to properly handle asymmetrical batches (where the size of the events doesn't influence the size of the results); Also it touches on the need to expose the batch size to the language client (we do that in the Go and the [Node](https://github.com/tigerbeetle/tigerbeetle/blob/cbfc1635a51a6b14085b3bd8d769b5b3cc131ded/src/node.zig#L549-L559) clients - as you spotted in the [comment bellow](https://github.com/tigerbeetle/tigerbeetle/pull/2080#discussion_r1671204708))

sentientwaffle (approved):

PR #2081, opened 2024-07-08
ci: sanity-check level code coverage

matklad:
It's easy to just run kcov, so let's just do that without thinking to
much about how it's best to integrate this into our overall dev process.

For now, let's just run a handful of test on main post merge, and upload
the results as a GitHub artifact (so, you can download the HTML report
manually).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -149,6 +149,20 @@ jobs:
           path: ./src/devhub
       - uses: actions/deploy-pages@v1
 
+  kcov:
+    if: github.ref == 'refs/heads/main'
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 2147483647
+      - run: sudo apt install kcov
+      - run: ./scripts/install_zig.sh
+      - run: ./zig/zig build scripts -- kcov
+      - uses: actions/upload-artifact@v4
+        with:
+          name: code-coverage-report
+          path: zig-out/kcov
 
   # Work around GitHub considering Skipped jobs success for "Require status checks before merging"
   # See also:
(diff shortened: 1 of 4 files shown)


matklad:
Example report: https://github.com/tigerbeetle/tigerbeetle/actions/runs/9842057297/artifacts/1678280132

matklad:
>Example report:

It honestly looks *much* better than what I'd naively expect from the project which doesn't track coverage. Two obvious blind spots are:

* client eviciton
* scans (but there are couple in-progress PRs for this already)

batiati (approved):
Amazing!

What do you think about adding it as a new metric to the devhub?
It can be extracted from the JSON ouput:
```sh
❯ jq '.percent_covered' ./zig-out/kcov/vopr/coverage.json
"71.56"
```

matklad:
I am still thinking about the best way to integrate it with dev workflow, and I don’t know good answers! The two principle problems are:

* getting useful signal
* fuzzer randomization 


For getting useful signal out of it, don’t think coverage percentage would actually be useful. Once you get past, say, 80 coverage, I don’t think you’ll get a visceral feel for that the next 5% means.  And we are pretty well covered already.

What you are interested in are rather uncovered “hot spots”, and not the overall percentage. So, I looked at the data and it pointed at client eviction and scans as problematic spots, but it isn’t clear how to extract that in robust, automatic way.

🤔 i wonder of what we need is the list of uncovered non-test functions? So, not the overall percentage, but “this function is uncovered”? Ah, and we can sort that by the size of the function! Yeah, “largest least covered function” is perhaps what we should put out there to devhub.

The second problem for us is fuzzer randomization. Ideally, we’d run the whole thing multiple times and merge results, and integrate that in cfo, but it’s unclear how to do that exactly in a way that’s robust and fast.

In other words, I am still thinking here: I want to avoid doing just “something”, as that risks giving us false sense of usefulness.

PR #2087, opened 2024-07-09
tidy: remove exceptions from Java

matklad:
We already have carve out for `https://` URLS in tidy, so URLs per se won't be problematic.

What remains is:

* One instance where we legitimately have an overly-long doc line, just shorten it.
* pom.xml, which uses `http://` URLs. We intentionally don't allow non-S http in tidy! But, luckily, we can add some newlines and change one url to https to fix this. I am using https://github.com/quarkusio/quarkus/blob/0be9c09688329a32524526ff6b0819bdc83c0ca8/pom.xml#L2-L5 as the evidence that s is indeed valid there

### DIFF
--- a/src/tidy.zig
+++ b/src/tidy.zig
@@ -98,7 +98,6 @@ const DeadDetector = struct {
     }
 
     fn deinit(detector: *DeadDetector) void {
-        assert(detector.files.count() == 0); // Sanity-check that `.finish` was called.
         detector.files.deinit();
     }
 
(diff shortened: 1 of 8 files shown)


batiati (dismissed):

src/tidy.zig
```diff
@@ -98,7 +98,6 @@ const DeadDetector = struct {
     }
 
     fn deinit(detector: *DeadDetector) void {
-        assert(detector.files.count() == 0); // Sanity-check that `.finish` was called.
```
matklad, line 101: Forgot to why this: this creates extra nosie if there are other tidy errors. ideally, `deinit` would take `?anyerror` and assert only if there's an error, but there's no appropriate form of `defer` for this pattern.

batiati (approved):

PR #2088, opened 2024-07-09
Verify benchmark state after execution, fix benchmark, tests on Windows

SuperAuguste:
- [x] Fix `AccessDenied` issue on Windows in benchmark
- [x] Fix `std.process.Child` issue in tests, enable them in Windows CI
- [x] Verify that all accounts were created properly
- [x] Verify that all transfers were created properly


matklad:
If you are looking at out windows build, would be great to add zig build test over here!

https://github.com/tigerbeetle/tigerbeetle/blob/c922683ab91af2e06e48009d1aed5d3fdeed2c7b/.github/workflows/ci.yml#L72

SuperAuguste:
Looked like too easy of a win so I tested it locally and it looks like there's at least one test failing due to ~~Windows filesystem shenanigans~~ - will fix and enable in CI.

EDIT: Turns out ChildProcess's `cwd` is cross-platform whereas `cwd_dir` is not!

SuperAuguste:
Enabled Windows on CI because it worked on my machine but now it's failing in CI. :/

Will keep working on benchmark for now and see if I can figure this out later.

matklad:
> Will keep working on benchmark for now and see if I can figure this out later.


You’ll need that clone depth parameter other ci jobs are using I thibk

SuperAuguste:
Oof I totally missed that, thanks Alex! :)

SuperAuguste:
bruh moment: `zig fmt --check .` was failing on Windows CI. First hypothesis was `*.zig` files being auto-CRLF-ed but then I saw the `.gitattributes` in `src` and my fears were assuaged. But of course there's one `.zig` file outside of `src`: the `build.zig`. Will modify the `.gitattributes` in repo root and we should be good. 😅

.gitattributes
```diff
@@ -1,2 +1,3 @@
 *.md text=auto eol=lf
-*.sh text=auto eol=lf
\ No newline at end of file
+*.sh text=auto eol=lf
+build.zig text=auto eol=lf
```
matklad, line 3: out of scope for this PR (because it might have unknown unknowns), but it would be great to simplify this to just

```
* text=auto eol=lf
```

matklad:
Do we need

```diff
diff --git a/build.zig b/build.zig
index e68dc5708..9dc374e99 100644
--- a/build.zig
+++ b/build.zig
@@ -301,7 +301,7 @@ pub fn build(b: *std.Build) !void {
         run_integration_tests.step.dependOn(b.getInstallStep());
         build_steps.test_integration.dependOn(&run_integration_tests.step);
 
-        const run_fmt = b.addFmt(.{ .paths = &.{"."}, .check = true });
+        const run_fmt = b.addFmt(.{ .paths = &.{ ".", "build.zig" }, .check = true });
         build_steps.test_fmt.dependOn(&run_fmt.step);
 
         build_steps.@"test".dependOn(&run_unit_tests.step);
```

?

EDIT: no, `.` includes `build.zig`, I am confusing this with something else I thinkg

src/tigerbeetle/benchmark_driver.zig
```diff
@@ -66,6 +60,13 @@ pub fn main(allocator: std.mem.Allocator, args: *const cli.Command.Benchmark) !v
     const addresses = args.addresses orelse &.{tigerbeetle_process.?.address};
     try benchmark_load.main(allocator, addresses, args);
 
+    if (tigerbeetle_process) |*p| {
+        const rusage = p.deinit();
```
matklad, line 67: :thinking: looks like we won't deinit `TigerBeetle` process if `        const me = try std.fs.selfExePathAlloc(allocator);` throws above. 

What's the motivation for moving this code out of a defer? (and, if we do move it out, then we can `s/catch {}/try` I guess?)

src/tigerbeetle/benchmark_load.zig
```diff
@@ -483,6 +514,110 @@ const Benchmark = struct {
...
+            std.mem.sliceAsBytes(b.batch_account_ids.items),
+        );
+    }
+
+    fn validate_accounts_finish(
+        b: *Benchmark,
+        operation: StateMachine.Operation,
+        result: []const u8,
+    ) void {
+        std.debug.assert(operation == .lookup_accounts);
```
cb22:
```suggestion
        assert(operation == .lookup_accounts);
```

(and below too)

src/tigerbeetle/benchmark_load.zig
```diff
@@ -483,6 +514,110 @@ const Benchmark = struct {
...
+
+        const accounts = std.mem.bytesAsSlice(tb.Account, result);
+
+        for (b.batch_accounts.items, accounts) |expected, actual| {
+            assert(expected.id == actual.id);
```
cb22, line 557: Can we compare the entire object, as bytes? (if so, we've got `stdx.equal_bytes` as a helper). Ditto for the transfers.

src/tigerbeetle/benchmark_load.zig
```diff
@@ -172,6 +182,18 @@ pub fn main(
 
     benchmark.create_accounts();
 
+    while (!benchmark.done) {
```
cb22, line 186: Let's make these conditional on new CLI args - something like `--transfer-validate` and `--account-validate`.

src/tigerbeetle/benchmark_load.zig
cb22: Right, of course we can't do that as easily here, because the accounts returned from the replica will have their `{debits,credits}_{pending,posted}` set.

It would be nice to track the balances in a model here, and compare them directly, but I think that'll get a bit messy with the pending / timeout logic. We could zero out those fields on the `actual` before comparison, though.

src/tigerbeetle/benchmark_driver.zig
SuperAuguste: Good point! The `try` over `catch {}` is the reason why. To keep this code working on Windows, I'd have to move the `if (data_file_created) {` logic into the same `defer`. I'd be fine with that.

src/tigerbeetle/benchmark_load.zig
```diff
@@ -483,6 +514,110 @@ const Benchmark = struct {
...
+
+        // Submit batch.
+        b.send(
+            validate_transfers_finish,
+            .lookup_transfers,
+            std.mem.sliceAsBytes(b.batch_transfer_ids.items),
+        );
+    }
+
+    fn validate_transfers_finish(
```
cb22, line 598: Some logging output would be great. (as another FYI, anything in the format of `{key} = {value}` can be easily slurped up by the [devhub](https://tigerbeetle.github.io/tigerbeetle/) if we like, too).

I'm not sure though that there are any metrics we want to track here, since we already do the query tests. Perhaps a simple
```
validated {x} accounts, {y} transfers
```

src/tigerbeetle/benchmark_driver.zig
matklad: :thinking: Ah, I guess the original smell here is that we are doing too much in the defer. I think we should change the code to

```
var tigerbeetle_process: ?TigerBeetleProcess = null;
defer if (tigerbeetle_process) |*p| _ = p.deinit();

...
...

const rusage = p.deinit();
p = null;
```

src/tigerbeetle/benchmark_load.zig
SuperAuguste: I'll just create a single `validate` flag as it must be all or nothing to recreate the same RNG state.

src/tigerbeetle/benchmark_load.zig
SuperAuguste: I thought about zeroing out the fields and then doing an `eql` but the issue is that `result` is a `[]const u8` so we can't do that.

cb22 (approved):

PR #2090, opened 2024-07-10
darwin: call fs_sync (F_FULLFSYNC) after each write()

cb22:
Turns out, while we were handling macOS's fsync quirks correctly for file creation, we weren't when actually writing to the file.

This does kill performance on macOS, but that performance was never ours to begin with :smile: 

main: load accepted = 259922 tx/s
this pr: load accepted = 52038 tx/s

There's probably a bit that can be done to speed this up, but let's get correct first.


matklad (approved):
Note to self: once we do "whole process testing" thingy, check that it can catch this bug.

PR #2092, opened 2024-07-10
tidy: state machine

matklad:


src/state_machine.zig
```diff
@@ -1971,15 +2034,16 @@ pub fn StateMachineType(
...
-                        .user_data_32 = .{ .batch_value_count_limit = batch_values_limit.accounts.user_data_32 },
-                        .ledger = .{ .batch_value_count_limit = batch_values_limit.accounts.ledger },
-                        .code = .{ .batch_value_count_limit = batch_values_limit.accounts.code },
+                    .tree_options_object = .{
+                        .batch_value_count_limit = batch_values_limit.accounts.timestamp,
+                    },
+                    .tree_options_id = .{
+                        .batch_value_count_limit = batch_values_limit.accounts.id,
                     },
+                    .tree_options_index = index_tree_options(
```
matklad, line 2043: I am not very happy with this brute-force solution, but I think it's better than what we had here before --- it was very hard to spot a typo in the wall of text we had here before.

src/tidy.zig
```diff
@@ -361,7 +361,16 @@ fn find_long_line(file_text: []const u8) !?usize {
             if (parse_multiline_string(line)) |string_value| {
                 const string_value_length = try std.unicode.utf8CountCodepoints(string_value);
                 if (string_value_length <= 100) continue;
+
+                if (std.mem.startsWith(u8, string_value, " account A") or
+                    std.mem.startsWith(u8, string_value, " transfer T") or
+                    std.mem.startsWith(u8, string_value, " transfer   "))
+                {
+                    // Table tests from state_machine.zig. They are intentionally wide.
```
matklad, line 369: The general philosophy is that we don't have a fine grained opt-out like `fmt:off`, and instead add extremely well motivated exceptions to the tidy itself. 

I think our state machine tests pass the threshold of well-motivatedness.

src/state_machine.zig
```diff
@@ -2362,7 +2438,8 @@ const TestContext = struct {
 
         ctx.state_machine = try StateMachine.init(allocator, &ctx.grid, .{
             .batch_size_limit = message_body_size_max,
-            .lsm_forest_compaction_block_count = StateMachine.Forest.Options.compaction_block_count_min,
+            .lsm_forest_compaction_block_count = StateMachine.Forest.Options
+                .compaction_block_count_min,
```
sentientwaffle, line 2442: TIL `zig fmt` allows this!

sentientwaffle (approved):

PR #2095, opened 2024-07-10
stdx: add custom formatter for displaying units in error messages

### LINKED ISSUE #2050: Use custom formatter for displaying units in error messages
re https://github.com/tigerbeetle/tigerbeetle/pull/2036

In many error messages we display byte units (MiB, KiB, etc.). Sometimes these units are displayed differently based on the compile-time configuration (see https://github.com/tigerbeetle/tigerbeetle/pull/2036/files#diff-104744d347a9d7ea7a7e958f254327603a52f6e79517adbdba9950e8af779c97R499).

This code is becoming unwieldy. Per the above PR it may make sense to use custom formatters to automatically pick the best unit to display in these situations.

Ours would be similar to `std.fmt.fmtIntSizeBin` except that we want to use `@divExact`, expecting our number to always be evenly divisible by the displayed unit. We also want to continue to use comptime to ensure these numbers are evenly divisible during the build, and not have unexpected divExact failures at runtime.

tensorush:
Closes https://github.com/tigerbeetle/tigerbeetle/issues/2050

Implemented custom formatter `fmtIntSizeBinExact` and incorporated it into `cli.zig`.

### DIFF
--- a/src/stdx.zig
+++ b/src/stdx.zig
@@ -749,12 +749,12 @@ pub fn comptime_slice(comptime slice: anytype, comptime len: usize) []const @Typ
 /// Return a Formatter for a u64 value representing a file size.
 /// This formatter statically checks that the number is a multiple of 1024,
 /// and represents it using the IEC measurement units (KiB, MiB, GiB, ...).
-pub fn fmtIntSizeBinExact(comptime value: u64) std.fmt.Formatter(formatIntSizeBinExact) {
+pub fn fmt_int_size_bin_exact(comptime value: u64) std.fmt.Formatter(format_int_size_bin_exact) {
     comptime assert(value % 1024 == 0);
     return .{ .data = value };
 }
 
-fn formatIntSizeBinExact(
+fn format_int_size_bin_exact(
     value: u64,
     comptime fmt: []const u8,
     options: std.fmt.FormatOptions,
@@ -771,9 +771,9 @@ fn formatIntSizeBinExact(
     var buf: [7]u8 = undefined;
 
     const mags_iec = " KMGTPEZY";
-    const magnitude = @min(std.math.log2(value) / 10, mags_iec.len - 1);
+    const magnitude = @min(@divFloor(std.math.log2(value), 10), mags_iec.len - 1);
     const suffix = mags_iec[magnitude];
-    const new_value = value / std.math.pow(u64, 1024, magnitude);
+    const new_value = @divExact(value, std.math.pow(u64, 1024, magnitude));
 
     const i = std.fmt.formatIntBuf(&buf, new_value, 10, .lower, .{});
     buf[i..][0..3].* = [_]u8{ suffix, 'i', 'B' };
@@ -781,9 +781,9 @@ fn formatIntSizeBinExact(
     return std.fmt.formatBuf(buf[0 .. i + 3], options, writer);
 }
 
-test fmtIntSizeBinExact {
-    try std.testing.expectFmt("0B", "{}", .{fmtIntSizeBinExact(0)});
-    try std.testing.expectFmt("8KiB", "{}", .{fmtIntSizeBinExact(8 * 1024)});
-    try std.testing.expectFmt("42MiB", "{}", .{fmtIntSizeBinExact(42 * 1024 * 1024)});
-    try std.testing.expectFmt("999GiB", "{}", .{fmtIntSizeBinExact(999 * 1024 * 1024 * 1024)});
+test fmt_int_size_bin_exact {
+    try std.testing.expectFmt("0B", "{}", .{fmt_int_size_bin_exact(0)});
+    try std.testing.expectFmt("8KiB", "{}", .{fmt_int_size_bin_exact(8 * 1024)});
+    try std.testing.expectFmt("42MiB", "{}", .{fmt_int_size_bin_exact(42 * 1024 * 1024)});
+    try std.testing.expectFmt("999GiB", "{}", .{fmt_int_size_bin_exact(999 * 1024 * 1024 * 1024)});
 }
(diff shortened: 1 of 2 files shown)


src/stdx.zig
```diff
@@ -745,3 +745,45 @@ pub fn EnumUnionType(
 pub fn comptime_slice(comptime slice: anytype, comptime len: usize) []const @TypeOf(slice[0]) {
     return &@as([len]@TypeOf(slice[0]), slice[0..len].*);
 }
+
+/// Return a Formatter for a u64 value representing a file size.
+/// This formatter statically checks that the number is a multiple of 1024,
+/// and represents it using the IEC measurement units (KiB, MiB, GiB, ...).
+pub fn fmtIntSizeBinExact(comptime value: u64) std.fmt.Formatter(formatIntSizeBinExact) {
```
matklad: We use snake case for function names in TigerBeetle

src/stdx.zig
```diff
@@ -745,3 +745,45 @@ pub fn EnumUnionType(
...
+        return std.fmt.formatBuf("0B", options, writer);
+    }
+
+    // The worst case in terms of space needed is 4 bytes,
+    // since 1023 is the highest possible number,
+    // + 3 bytes for the unit suffix.
+    var buf: [7]u8 = undefined;
+
+    const mags_iec = " KMGTPEZY";
+    const magnitude = @min(std.math.log2(value) / 10, mags_iec.len - 1);
```
matklad: use `@divFloor` here, in general we avoid `/`

src/stdx.zig
```diff
@@ -745,3 +745,45 @@ pub fn EnumUnionType(
...
+    const suffix = mags_iec[magnitude];
+    const new_value = value / std.math.pow(u64, 1024, magnitude);
```
matklad: and this could be `@divExact`

src/stdx.zig
```diff
@@ -745,3 +745,45 @@ pub fn EnumUnionType(
...
+
+    const i = std.fmt.formatIntBuf(&buf, new_value, 10, .lower, .{});
+    buf[i..][0..3].* = [_]u8{ suffix, 'i', 'B' };
+
+    return std.fmt.formatBuf(buf[0 .. i + 3], options, writer);
+}
+
+test fmtIntSizeBinExact {
+    try std.testing.expectFmt("0B", "{}", .{fmtIntSizeBinExact(0)});
+    try std.testing.expectFmt("8KiB", "{}", .{fmtIntSizeBinExact(8 * 1024)});
```
matklad: Could you add a case for `1025 * 1024`, it isn't entirely obvious that this'll work correctly.

src/tigerbeetle/cli.zig
```diff
@@ -456,26 +456,25 @@ pub fn parse_args(allocator: std.mem.Allocator, args_iterator: *std.process.ArgI
...
+                flags.fatal("--limit-storage: size {}{s} exceeds maximum: {}", .{
                     start_limit_storage.value,
                     start_limit_storage.suffix(),
-                    @divExact(storage_size_limit_max, 1024 * 1024),
+                    vsr.stdx.fmtIntSizeBinExact(storage_size_limit_max),
                 });
-            }
-            if (storage_size_limit < storage_size_limit_min) {
-                flags.fatal("--limit-storage: size {}{s} is below minimum: {}KiB", .{
+            } else if (storage_size_limit < storage_size_limit_min) {
```
matklad: `fatal` is `noreturn`, so I'd rather skip the elses, as that gives simpler structure.

src/stdx.zig
tensorush: You're right, it doesn't. Hmm, rn can't think of a way to change that `comptime assert` to account for that case as well.

tensorush: Sorry, I misunderstood the requirements a bit, please see the new `format_int_size_bin_exact`.

src/stdx.zig
```diff
@@ -745,3 +745,50 @@ pub fn EnumUnionType(
...
+    _ = fmt;
+    if (value == 0) {
+        return std.fmt.formatBuf("0B", options, writer);
+    }
+
+    // The worst case in terms of space needed is 20 bytes,
+    // since `maxInt(u64)` is the highest number,
+    // + 3 bytes for the measurement units suffix.
+    var buf: [23]u8 = undefined;
+
```
matklad, line 773: could you add `comptime assert` that comptime printing `maxInt(u64) GiB` would be not longer than 23?

src/stdx.zig
```diff
@@ -745,3 +745,50 @@ pub fn EnumUnionType(
...
+    return std.fmt.formatBuf(buf[0 .. i + 3], options, writer);
+}
+
+test fmt_int_size_bin_exact {
+    try std.testing.expectFmt("0B", "{}", .{fmt_int_size_bin_exact(0)});
+    try std.testing.expectFmt("8KiB", "{}", .{fmt_int_size_bin_exact(8 * 1024)});
+    try std.testing.expectFmt("1025KiB", "{}", .{fmt_int_size_bin_exact(1025 * 1024)});
+    try std.testing.expectFmt("12345KiB", "{}", .{fmt_int_size_bin_exact(12345 * 1024)});
+    try std.testing.expectFmt("42MiB", "{}", .{fmt_int_size_bin_exact(42 * 1024 * 1024)});
```
matklad, line 794: Let's test the extreme value of `maxInt(u64)` here as well

src/tigerbeetle/cli.zig
```diff
@@ -493,7 +493,7 @@ pub fn parse_args(allocator: std.mem.Allocator, args_iterator: *std.process.ArgI
             if (pipeline_limit < pipeline_limit_min) {
                 flags.fatal("--limit-pipeline-requests: count {} is below minimum: {}", .{
                     pipeline_limit,
-                    pipeline_limit_min,
+                    vsr.stdx.fmt_int_size_bin_exact(pipeline_limit_min),
```
matklad: This is a count of message, not a size in bytes, so I think should be left as is

(this might be a dulicate comment, GitHub is acting weirdly for me here)

src/stdx.zig
tensorush: Tested the extreme value of `maxInt(u64) - 1023`, satisfying the exact division by 1024

matklad (approved):
Thanks!

PR #2096, opened 2024-07-10
ci: run the benchmark, with --validate, in CI

cb22:
With #2088, we can now run the benchmark each time in CI and have it validate its results. It takes around ~3.5 minutes end to end and runs in parallel with the other (longer) steps so there's no slowdown.

I've made it use the git SHA like the vopr does, as the seed for the benchmark. But I'm not sure if maybe a fixed seed is better :thinking: 

It would also be nice to kill the replica, and restart it (to ensure that caches are flushed) before validating, but that can be a future TODO.

### DIFF
--- a/src/integration_tests.zig
+++ b/src/integration_tests.zig
@@ -150,7 +150,7 @@ test "benchmark smoke" {
 
     const tigerbeetle = try tigerbeetle_exe(shell);
     const status_ok = try shell.exec_status_ok(
-        "{tigerbeetle} benchmark --transfer-count=4000",
+        "{tigerbeetle} benchmark --transfer-count=10_000 --transfer-batch-size=10 --validate",
         .{ .tigerbeetle = tigerbeetle },
     );
     try std.testing.expect(status_ok);
--- a/src/testing/fuzz.zig
+++ b/src/testing/fuzz.zig
@@ -71,3 +71,23 @@ pub const FuzzArgs = struct {
     seed: u64,
     events_max: ?usize,
 };
+
+pub fn parse_seed(bytes: []const u8) u64 {
+    if (bytes.len == 40) {
+        // Normally, a seed is specified as a base-10 integer. However, as a special case, we allow
+        // using a Git hash (a hex string 40 character long). This is used by our CI, which passes
+        // current commit hash as a seed --- that way, we run simulator on CI, we run it with
+        // different, "random" seeds, but the failures remain reproducible just from the commit
+        // hash!
+        const commit_hash = std.fmt.parseUnsigned(u160, bytes, 16) catch |err| switch (err) {
+            error.Overflow => unreachable,
+            error.InvalidCharacter => @panic("commit hash seed contains an invalid character"),
+        };
+        return @truncate(commit_hash);
+    }
+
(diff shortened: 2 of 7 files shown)


matklad:
> I've made it use the git SHA like the vopr does, as the seed for the benchmark. But I'm not sure if maybe a fixed seed is better 🤔

Let’s got for fixed seed here, the less variance we introduce, the better

matklad:
We probably want to add validate to 

https://github.com/tigerbeetle/tigerbeetle/blob/aafcf0f2016c7029b1ae1e3e9d7c348bb0195e16/src/scripts/devhub.zig#L35

as well? Ah, and we also have an integration test that might also want to do this:

https://github.com/tigerbeetle/tigerbeetle/blob/aafcf0f2016c7029b1ae1e3e9d7c348bb0195e16/src/integration_tests.zig#L152



I am also wondering if doing this *just* for devhub (so, post merge) would be better? yeah, I think between a small integration test, and a big devhub benchmark, we perhaps can avoid a new CI job here?

cb22:
Ah, I forgot about the integration test benchmark. I think that's OK for now, until we come up with something for scale tests etc too.

(My main issue with it only being in the devhub is that's only run after code is already in `main`.)

matklad:
>(My main issue with it only being in the devhub is that's only run after code is already in main.)

Yeah, that's why we have integration test for the benchmark, it does gate PR merging.

src/integration_tests.zig
```diff
@@ -150,7 +150,7 @@ test "benchmark smoke" {
 
     const tigerbeetle = try tigerbeetle_exe(shell);
     const status_ok = try shell.exec_status_ok(
-        "{tigerbeetle} benchmark --transfer-count=4000",
+        "{tigerbeetle} benchmark --transfer-count=10_000 --transfer-batch-size=10 --validate",
```
cb22, line 153: Bumped the `transfer-count` up and the `transfer-batch-size` down to squeeze a bit more surface area from this run, while still being fast.

src/testing/fuzz.zig
```diff
@@ -71,3 +71,23 @@ pub const FuzzArgs = struct {
     seed: u64,
     events_max: ?usize,
 };
+
+pub fn parse_seed(bytes: []const u8) u64 {
```
cb22, line 75: Figured this belongs here anyhow :smile:

src/tigerbeetle/benchmark_load.zig
```diff
@@ -126,7 +126,11 @@ pub fn main(
     }
 
     // If no seed was given, use a default seed for reproducibility.
-    const seed: usize = cli_args.seed orelse 42;
+    const seed = seed_from_arg: {
+        const seed_argument = cli_args.seed orelse break :seed_from_arg 42;
+        break :seed_from_arg vsr.testing.parse_seed(seed_argument);
```
matklad, line 131: Hm, using vsr.testing namespace here feels wrong, as this is code from the main binary, so it _shouldn't_ include that namespace. Sort-of the same reason why `benchamrk_driver` doesn't use `shell.zig` and spanwns the process by hand. 

But we already use a couple of other functions here, so this is per-existing and doesn't need fixing in this PR

matklad (approved):

PR #2101, opened 2024-07-11
lsm: don't store data twice

matklad:
value_count_limit is equal to values.len, better to avoid duplicating this info!


cb22 (approved):
The title is trollface.jpg :rofl:

PR #2103, opened 2024-07-12
vopr: fix false positive 

matklad:
The seed in question triggers assertion failure in the VOPR itself.
This commit relaxes the assertion.

Background:

Historically, VOPR tried to maintain the invariant that at least a
quorum of reiplcas are available. In particular, when restarting a
replica VOPR tried to repair torn writes, to prevent too many replicas
from getting a recovering head.

Since then, we've changed VOPR mode of operation: instead of trying to
_strictly_ stay withing acceptable error budget, it now sometimes
injects slightly more errors and, if the cluster gets stuck, we
carefully check that the cluster is stuck for a good reason.

As this seed shows, there are cases where we can't trivially repair a
torn write (the prepare and the header are both valid, but from
different views). But, as the overall invariant was relaxed, we can just
ignore this case as well.

Relevant crash logs:

```
...
[warn] (journal): 0: recover_slot: recovered slot=0003 label=@l decision=vsr operation=vsr.Operation.reserved op=3
...
[debug] (journal): 0: recover_slots: dirty=3 faulty=1
...
[warn] (replica): 0: op_head_certain: faulty slots count=1
...
[warn] (replica): 0: transition_to_recovering_head: op_checkpoint=47 commit_min=47 op_head=79 log_view=71 view=71
[info] (cluster):  0 ^ \ h      71V  47/_47/_75C  46:_79Jo  1/_1J!  46:_79Wo <__0:__0> v1:2  nullGa  0G!   0G?
thread 296191 panic: reached unreachable code
/home/matklad/p/tb/work/zig/lib/std/debug.zig:412:14: 0x11505fd in assert (vopr)
             ^
/home/matklad/p/tb/work/src/vopr.zig:1071:19: 0x11b4a45 in restart_replica (vopr)
            assert(fault or

```

Seed: ./zig/zig build -Dvopr-state-machine=testing vopr -- --lite 2546999548770977086

### DIFF
--- a/src/vopr.zig
+++ b/src/vopr.zig
@@ -989,8 +989,11 @@ pub const Simulator = struct {
                 r.syncing == .idle);
         }
 
+        // To improve VOPR utilization, try to prevent the replica from going into
+        // `.recovering_head` state if the replica is needed to form a quorum.
         const fault = recoverable_count >= recoverable_count_min or replica.standby();
         simulator.restart_replica(replica.replica, fault);
+        maybe(!fault and replica.status == .recovering_head);
     }
 
     fn restart_replica(simulator: *Simulator, replica_index: u8, fault: bool) void {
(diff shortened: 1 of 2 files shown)


sentientwaffle (approved):
Makes sense!

PR #2104, opened 2024-07-12
vsr: advance op_prepare_max when checkpoint is durable on a commit-quorum of replicas

chaitanyabhandari:
### Problem

Currently, we allow overwriting a prepare that a replica has already checkpointed. This builds on the understanding that the prepares that have already checkpointed are *durable* in the LSM tree. The key problem is that *op_prepare_max* is maintained *independently* on each replica according to their *op_checkpoint*. However, this can compromise availability in the presence of storage faults. Consider the following scenario in a 3-replica cluster:

A prepare `P` is replicated to R1 & R2, but not to R3. R1 decides to evict and overwrite `P` since it is in R1's checkpoint. However, R2 hasn't reached this checkpoint yet, so it still has `P` in its WAL.  Further, assume a pair of uncorrelated storage faults - a storage corruption of `P` in R2's WAL, combined with a storage corruption in R1's LSM. This can render the cluster unavailable. The core issue here is that even though even have *logical* durability of data across a commit-quorum of replicas (`P` is durable in R1's checkpoint in the LSM tree and in R2's WAL), the *physical* durability of the data is compromised, since the data is present across a commit-quorum of replicas in *different* forms.

### Solution

The solution to this problem is to allow the overwrite of a prepare only if a *commit-quorum* of replicas has already checkpointed that prepare. Specifically, The *primary* only prepares past *op_prepare_max* when it hears from at least a commit quorum of replicas at the current checkpoint.

### Example
Assume:
* journal_slot_count=10
* lsm_batch_multiple=2
* pipeline_prepare_queue_max=2
*  checkpoint_interval = journal_slot_count - (lsm_batch_multiple + pipeline_prepare_queue_max) = 6

Based on the aforementioned configuration we can compute the following:
```
      checkpoint() call           0   1   2   3
      op_checkpoint               0   5  11  17
      op_checkpoint_next          5  11  17  23
      op_checkpoint_next_trigger  7  13  19  25
```

Now, let's assume a 3-replica cluster. Here, *replication quorum* is *2*. All replicas start with op_checkpoint as 0.

Let's say R1 is the primary. R1 commits *op=7* (*op_checkpoint_next_trigger* for op_checkpoint=5) and updates its *op_checkpoint* to *5*. Now, R1 does not immediately prepare to past the current *op_prepare_max* (9). Instead, it waits till either R2 or R3 also update their *op_checkpoint* to *5* (via ping/commit messages). Once R1 hears from a replication quorum of replicas on op_checkpoint=5, it sets its op_prepare_max to 15.

### DIFF
--- a/src/vsr.zig
+++ b/src/vsr.zig
@@ -1480,7 +1480,7 @@ pub const Checkpoint = struct {
         assert(constants.journal_slot_count > constants.lsm_batch_multiple);
         assert(constants.journal_slot_count % constants.lsm_batch_multiple == 0);
     }
-    /// Checkpoint identifier.
+    /// Checksum of the CheckpointState in the superblock.
     id: u128,
     /// The op_checkpoint() that corresponds to the checkpoint id.
     op: u64,
@@ -1502,9 +1502,8 @@ pub const Checkpoint = struct {
                 }
 
                 // Ignore repeat candidate.
-                if (checkpoint.op == checkpoint_existing.op and
-                    checkpoint.id == checkpoint_existing.id)
-                {
+                if (checkpoint.op == checkpoint_existing.op) {
+                    assert(checkpoint.id == checkpoint_existing.id);
                     return false;
                 }
             }
@@ -1515,10 +1514,8 @@ pub const Checkpoint = struct {
         pub fn count(quorum: *const Quorum, checkpoint: *const Checkpoint) usize {
             var matching: usize = 0;
             for (quorum.checkpoints) |checkpoint_existing| {
-                if (checkpoint_existing != null and
-                    checkpoint_existing.?.op == checkpoint.op and
-                    checkpoint_existing.?.id == checkpoint.id)
-                {
+                if (checkpoint_existing != null and checkpoint_existing.?.op == checkpoint.op) {
+                    assert(checkpoint_existing.?.id == checkpoint.id);
                     assert(std.meta.eql(checkpoint.*, checkpoint_existing.?));
                     matching += 1;
                 }
--- a/src/vsr.zig
+++ b/src/vsr.zig
@@ -1480,24 +1480,48 @@ pub const Checkpoint = struct {
         assert(constants.journal_slot_count > constants.lsm_batch_multiple);
         assert(constants.journal_slot_count % constants.lsm_batch_multiple == 0);
     }
-    pub const Quorum = struct {
-        /// The latest known checkpoint operation from every *other* replica.
-        checkpoints: [constants.replicas_max]u128 = [_]u128{0} ** constants.replicas_max,
+    /// Checkpoint identifier.
+    id: u128,
+    /// The op_checkpoint() that corresponds to the checkpoint id.
+    op: u64,
 
-        pub fn replace(quorum: *Quorum, replica: u8, checkpoint_op: u128) bool {
-            const checkpoint_op_existing: u128 = quorum.checkpoints[replica];
-            // Ignore old and repeat checkpoints.
-            if (checkpoint_op < checkpoint_op_existing or checkpoint_op == checkpoint_op_existing) {
-                return false;
+    pub const Quorum = struct {
+        /// The latest known checkpoint identifier from every *other* replica.
+        checkpoints: [constants.replicas_max]?Checkpoint =
+            [_]?Checkpoint{null} ** constants.replicas_max,
+
+        pub fn replace(
+            quorum: *Quorum,
+            replica: u8,
+            checkpoint: *const Checkpoint,
+        ) bool {
+            if (quorum.checkpoints[replica]) |checkpoint_existing| {
+                // Ignore old candidate.
+                if (checkpoint.op < checkpoint_existing.op) {
+                    return false;
+                }
+
+                // Ignore repeat candidate.
+                if (checkpoint.op == checkpoint_existing.op and
+                    checkpoint.id == checkpoint_existing.id)
+                {
+                    return false;
+                }
             }
-            quorum.checkpoints[replica] = checkpoint_op;
+            quorum.checkpoints[replica] = checkpoint.*;
             return true;
         }
 
-        pub fn count(quorum: *const Quorum, checkpoint_op: u128) u8 {
-            var matching: u8 = 0;
-            for (quorum.checkpoints) |checkpoint_op_existing| {
-                if (checkpoint_op_existing == checkpoint_op) matching += 1;
+        pub fn count(quorum: *const Quorum, checkpoint: *const Checkpoint) usize {
+            var matching: usize = 0;
+            for (quorum.checkpoints) |checkpoint_existing| {
+                if (checkpoint_existing != null and
+                    checkpoint_existing.?.op == checkpoint.op and
+                    checkpoint_existing.?.id == checkpoint.id)
+                {
+                    assert(std.meta.eql(checkpoint.*, checkpoint_existing.?));
+                    matching += 1;
+                }
             }
             return matching;
         }
--- a/src/vsr.zig
+++ b/src/vsr.zig
@@ -1452,7 +1452,7 @@ const ViewChangeHeadersArray = struct {
     }
 };
 
-/// For a replica with journal_slot_count=9, lsm_batch_multiple=2, pipeline_prepare_queue_max=1, and
+/// For a replica with journal_slot_count=10, lsm_batch_multiple=2, pipeline_prepare_queue_max=2,
 /// checkpoint_interval = journal_slot_count - (lsm_batch_multiple + pipeline_prepare_queue_max) = 6
 ///
 ///   checkpoint() call           0   1   2   3
@@ -1461,11 +1461,11 @@ const ViewChangeHeadersArray = struct {
 ///   op_checkpoint_next_trigger  7  13  19  25
 ///
 ///     commit log (ops)           │ write-ahead log (slots)
-///     0   4   8   2   6   0   4  │  0  -  -  -  4  -  -  -  8
-///   0 ─────✓·%                   │[ 0  1  2  3  4  ✓] 6  %  R
-///   1 ───────────✓·%             │  9 10  ✓]12  %  5[ 6  7  8
-///   2 ─────────────────✓·%       │ 18  % 11[12 13 14 15 16  ✓]
-///   3 ───────────────────────✓·% │[18 19 20 21 22  ✓]24  % 17
+///     0   4   8   2   6   0   4  │  0  -  -  -  4  -  -  -  -  9
+///   0 ─────✓·%                   │[ 0  1  2  3  4  ✓] 6  %  R  R
+///   1 ───────────✓·%             │  10 ✓] 12 %  4  5[ 6  7  8  9
+///   2 ─────────────────✓·%       │  10 11[12 13 14 15 16 ✓] 18 %
+///   3 ───────────────────────✓·% │  20 21 22 ✓] 24 %  16 17 [18 19
 ///
 /// Legend:
 ///
@@ -1480,6 +1480,52 @@ pub const Checkpoint = struct {
         assert(constants.journal_slot_count > constants.lsm_batch_multiple);
         assert(constants.journal_slot_count % constants.lsm_batch_multiple == 0);
     }
+    /// Checkpoint identifier.
+    id: u128,
+    /// The op_checkpoint() that corresponds to the checkpoint id.
+    op: u64,
+
+    pub const Quorum = struct {
+        /// The latest known checkpoint identifier from every *other* replica.
+        checkpoints: [constants.replicas_max]?Checkpoint =
+            [_]?Checkpoint{null} ** constants.replicas_max,
+
+        pub fn replace(
+            quorum: *Quorum,
+            replica: u8,
+            checkpoint: *const Checkpoint,
+        ) bool {
+            if (quorum.checkpoints[replica]) |checkpoint_existing| {
+                // Ignore old candidate.
+                if (checkpoint.op < checkpoint_existing.op) {
+                    return false;
+                }
+
+                // Ignore repeat candidate.
+                if (checkpoint.op == checkpoint_existing.op and
+                    checkpoint.id == checkpoint_existing.id)
+                {
+                    return false;
+                }
+            }
+            quorum.checkpoints[replica] = checkpoint.*;
+            return true;
+        }
+
+        pub fn count(quorum: *const Quorum, checkpoint: *const Checkpoint) usize {
+            var matching: usize = 0;
+            for (quorum.checkpoints) |checkpoint_existing| {
+                if (checkpoint_existing != null and
+                    checkpoint_existing.?.op == checkpoint.op and
+                    checkpoint_existing.?.id == checkpoint.id)
+                {
+                    assert(std.meta.eql(checkpoint.*, checkpoint_existing.?));
+                    matching += 1;
+                }
+            }
+            return matching;
+        }
+    };
 
     pub fn checkpoint_after(checkpoint: u64) u64 {
         assert(valid(checkpoint));
--- a/src/vsr.zig
+++ b/src/vsr.zig
@@ -1480,48 +1480,24 @@ pub const Checkpoint = struct {
         assert(constants.journal_slot_count > constants.lsm_batch_multiple);
         assert(constants.journal_slot_count % constants.lsm_batch_multiple == 0);
     }
-    /// Checkpoint identifier.
-    id: u128,
-    /// The op_checkpoint() that corresponds to the checkpoint id.
-    op: u64,
-
     pub const Quorum = struct {
-        /// The latest known checkpoint identifier from every *other* replica.
-        checkpoints: [constants.replicas_max]?Checkpoint =
-            [_]?Checkpoint{null} ** constants.replicas_max,
+        /// The latest known checkpoint operation from every *other* replica.
+        checkpoints: [constants.replicas_max]u128 = [_]u128{0} ** constants.replicas_max,
 
-        pub fn replace(
-            quorum: *Quorum,
-            replica: u8,
-            checkpoint: *const Checkpoint,
-        ) bool {
-            if (quorum.checkpoints[replica]) |checkpoint_existing| {
-                // Ignore old candidate.
-                if (checkpoint.op < checkpoint_existing.op) {
-                    return false;
-                }
-
-                // Ignore repeat candidate.
-                if (checkpoint.op == checkpoint_existing.op and
-                    checkpoint.id == checkpoint_existing.id)
-                {
-                    return false;
-                }
+        pub fn replace(quorum: *Quorum, replica: u8, checkpoint_op: u128) bool {
+            const checkpoint_op_existing: u128 = quorum.checkpoints[replica];
+            // Ignore old and repeat checkpoints.
+            if (checkpoint_op < checkpoint_op_existing or checkpoint_op == checkpoint_op_existing) {
+                return false;
             }
-            quorum.checkpoints[replica] = checkpoint.*;
+            quorum.checkpoints[replica] = checkpoint_op;
             return true;
         }
 
-        pub fn count(quorum: *const Quorum, checkpoint: *const Checkpoint) usize {
-            var matching: usize = 0;
-            for (quorum.checkpoints) |checkpoint_existing| {
-                if (checkpoint_existing != null and
-                    checkpoint_existing.?.op == checkpoint.op and
-                    checkpoint_existing.?.id == checkpoint.id)
-                {
-                    assert(std.meta.eql(checkpoint.*, checkpoint_existing.?));
-                    matching += 1;
-                }
+        pub fn count(quorum: *const Quorum, checkpoint_op: u128) u8 {
+            var matching: u8 = 0;
+            for (quorum.checkpoints) |checkpoint_op_existing| {
+                if (checkpoint_op_existing == checkpoint_op) matching += 1;
             }
             return matching;
         }
--- a/src/vsr/replica.zig
+++ b/src/vsr/replica.zig
@@ -3242,6 +3242,7 @@ pub fn ReplicaType(
                 // Already upgrading.
                 // Normally we chain send-upgrade-to-self via the commit chain.
                 // But there are a couple special cases where we need to restart the chain:
+                // - The request-to-self might have been dropped due to lack of space in the WAL.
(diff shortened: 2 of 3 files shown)


src/vsr.zig
```diff
@@ -1571,6 +1571,52 @@ pub const Checkpoint = struct {
...
+                // Ignore old candidate.
+                if (checkpoint.op < checkpoint_existing.op) {
+                    return false;
+                }
+
+                // Ignore repeat candidate.
+                if (checkpoint.op == checkpoint_existing.op and
+                    checkpoint.id == checkpoint_existing.id)
+                {
+                    return false;
```
matklad:
```suggestion
                if (checkpoint.op == checkpoint_existing.op)
                {
                    assert(checkpoint.id == checkpoint_existing.id);
                    return false;
```

src/vsr.zig
```diff
@@ -1571,6 +1571,52 @@ pub const Checkpoint = struct {
...
+            for (quorum.checkpoints) |checkpoint_existing| {
+                if (checkpoint_existing != null and
+                    checkpoint_existing.?.op == checkpoint.op and
+                    checkpoint_existing.?.id == checkpoint.id)
+                {
+                    assert(std.meta.eql(checkpoint.*, checkpoint_existing.?));
+                    matching += 1;
+                }
+            }
+            return matching;
```
matklad, line 1523: Should be be able to assert that `matching > 0`?

src/vsr.zig
```diff
@@ -1571,6 +1571,52 @@ pub const Checkpoint = struct {
...
+                    checkpoint.id == checkpoint_existing.id)
+                {
+                    return false;
+                }
+            }
+            quorum.checkpoints[replica] = checkpoint.*;
+            return true;
+        }
+
+        pub fn count(quorum: *const Quorum, checkpoint: *const Checkpoint) usize {
```
matklad, line 1514: As per TigerStyle, avoid `usize` and use specific numeric types. Here, `u8` would work

src/vsr/replica.zig
```diff
@@ -646,9 +650,21 @@ pub fn ReplicaType(
             self.opened = false;
             self.journal.recover(journal_recover_callback);
             while (!self.opened) self.superblock.storage.tick();
+
+            // A replication quorum of replicas have committed atop op_prepare_max, op_checkpoint
+            // is guaranteed to be durable on a commit quorum of replicas.
+            if (self.op_checkpoint() == 0 or self.solo() or
+                self.commit_max > self.op_prepare_max())
+            {
+                self.checkpoint_quorum = true;
```
matklad: It is confusing that we set `self.checkpoint_quorum` based on `op_prepare_max`, but that itself uses `self.checkpoint_quorum` inside! Perhpaps better use `prepare_max_for_checkpoint` directly here.

src/vsr.zig
chaitanyabhandari: Requiring `matching > 0` requires us to use [replace()](https://github.com/tigerbeetle/tigerbeetle/pull/2104/files#diff-e95e4dc6c1aae349e798cfeed22ad01f830cce0e654e6b4b765b3e7b0c677b6aR9511-R9514) in multiple places (commit_op_checkpoint_superblock_callback, sync_superblock_update_callback, open) to make sure that we set our own checkpoint in checkpoint_from_all_replicas every time it changes. That seemed a little messy and error prone so I decided to use [if (matching >= self.quorum_replication - 1)](https://github.com/tigerbeetle/tigerbeetle/pull/2104/files#diff-e95e4dc6c1aae349e798cfeed22ad01f830cce0e654e6b4b765b3e7b0c677b6aR9522-R9523) instead, assuming we already have op_checkpoint() durable.

Thoughts?

src/vsr.zig
```diff
@@ -1575,6 +1575,28 @@ pub const Checkpoint = struct {
...
+            const checkpoint_op_existing: u128 = quorum.checkpoints[replica];
+            // Ignore old and repeat checkpoints.
+            if (checkpoint_op < checkpoint_op_existing or checkpoint_op == checkpoint_op_existing) {
+                return false;
+            }
+            quorum.checkpoints[replica] = checkpoint_op;
+            return true;
+        }
+
+        pub fn count(quorum: *const Quorum, checkpoint_op: u128) u8 {
```
matklad: Something's wrong with types here. checkpoint_id is u128, but checkpoint_op is u64. Strincty speaking, we only need to track the op, but I'd track both op and id, if possible, just to have tighter verification.

src/vsr/replica.zig
```diff
@@ -445,6 +446,9 @@ pub fn ReplicaType(
         /// protocol.
         do_view_change_quorum: bool = false,
 
+        checkpoint_from_all_replicas: CheckpointQuorum = .{},
+        checkpoint_quorum: bool = false,
```
matklad: Document the menaning of the field. While `checkpoint_from_all_replicas` is somewhat obivous, this bool isn't at all.

src/vsr/replica.zig
```diff
@@ -1550,6 +1566,17 @@ pub fn ReplicaType(
...
+            // requests in the request_queue to decide whether to accept this request.
+            if (self.op + self.pipeline.queue.request_queue.count + 1 > self.op_prepare_max()) {
+                log.debug("{}: on_request: ignoring op={} (too far ahead, prepare_max={}).", .{
+                    self.replica,
+                    self.op + self.pipeline.queue.request_queue.count + 1,
+                    self.op_prepare_max(),
+                });
+                defer self.message_bus.unref(request.message);
+                return;
+            }
```
matklad, line 1580: Hm, this looks a bit strange: my mental model is that request queue is a just a buffer where  we can/should add a new request if we can't otherwise execute it. 

That is, the case where we get a request, but we can't immediately convert it to prepare, is I think _exactly_ the case where we let the request sit in the request queue.

src/vsr/replica.zig
matklad: I am actually wondering, do we need to maintain it as a field? Can we make this a function that looks at self.checkpoint_from_all_replicas, self.solo() and self.op() ?

src/vsr/replica.zig
```diff
@@ -5798,7 +5840,14 @@ pub fn ReplicaType(
...
         /// committed and checkpointed.
         fn op_prepare_max(self: *const Self) u64 {
-            return vsr.Checkpoint.prepare_max_for_checkpoint(self.op_checkpoint_next()).?;
+            const checkpoint = checkpoint: {
+                if (self.checkpoint_quorum) {
+                    break :checkpoint self.op_checkpoint_next();
+                } else {
+                    break :checkpoint self.op_checkpoint();
+                }
+            };
```
matklad: Don't need a block here:

```suggestion
            const checkpoint = if (self.checkpoint_quorum) 
                    self.op_checkpoint_next()
                else 
                    self.op_checkpoint();
```

src/vsr/replica.zig
chaitanyabhandari: My understanding about the request queue is similar, I understood at it as an overflow for the prepare queue! However, the reason why we can't admit requests into the prepare/queue freely in _on_request_ now is:
 
By virtue of this change, replicas that are past `vsr.Checkpoint.trigger_for_checkpoint(C)` (and have set `self.op_checkpoint() = C`) could still have their `self.op_prepare_max` at `vsr.Checkpoint.prepare_max_for_checkpoint(C-1)`, which is merely a pipeline_prepare_queue_max away. Freely admitting requests into the prepare/request queue could get to a primary to a point where it hits [this assert](https://github.com/tigerbeetle/tigerbeetle/blob/main/src/vsr/replica.zig#L1645-L1655) in _on_prepare_, which we're _aren't_ supposed to hit as a primary.

Here's a trace from failed VOPR run (without the admission check in on_request) that shows the issue more clearly. I've logged the sizes of the prepare & request queues, self.op, and op_prepare_max:
```
[debug] (replica): 1:: commit_op: popped prepare with type get_account_transfers op 26:: prepare queue 3/4, request queue 2/3, self.op 29, op_prepare_max 31
[debug] (replica): 1:: on_request:: popped request query_accounts: prepare queue 3/4, request queue 1/3, self.op 29, op_prepare_max 31
[debug] (replica): 1:: primary_repair_prepare:: pushed prepare query_accounts: prepare queue 4/4, request queue 1/3, self.op 29, op_prepare_max 31
[debug] (replica): 1:: commit_op: popped prepare with type create_transfers op 27:: prepare queue 3/4, request queue 1/3, self.op 30, op_prepare_max 31
[debug] (replica): 1:: on_request:: popped request lookup_transfers: prepare queue 3/4, request queue 0/3, self.op 30, op_prepare_max 31
[debug] (replica): 1:: primary_repair_prepare:: pushed prepare lookup_transfers: prepare queue 4/4, request queue 0/3, self.op 30, op_prepare_max 31
[debug] (replica): 1:: on_request:: pushed request create_transfers: prepare queue 4/4, request queue 1/3, self.op 31, op_prepare_max 31
[debug] (replica): 1:: on_request:: pushed request create_transfers: prepare queue 4/4, request queue 2/3, self.op 31, op_prepare_max 31
[debug] (replica): 1:: commit_op: popped prepare with type create_transfers op 28:: prepare queue 3/4, request queue 2/3, self.op 31, op_prepare_max 31
[debug] (replica): 1:: on_request:: popped request create_transfers: prepare queue 3/4, request queue 1/3, self.op 31, op_prepare_max 31
[debug] (replica): 1:: primary_repair_prepare:: pushed prepare create_transfers: prepare queue 4/4, request queue 1/3, self.op 31, op_prepare_max 31
```

We can see from the logs that even after we've committed op=27 (which is the checkpoint trigger) and op=28, our op_prepare_max is still at 31. When we reach self.op=31, we have capacity for two more requests in the request queue, which we would not be able to accommodate till we receive a replication quorum for the checkpoint and advance our op_prepare_max.

src/vsr/replica.zig
chaitanyabhandari: Would be quite clean to make it a function of these, but I would have to give it some deeper thought if this is possible...

As of now, whether we mark self.checkpoint_quorum as True seems quite dynamic; we try to use incoming pings, commits, and DVC messages (in case we haven't received enough pings/commits and we don't want to wait to hear about a checkpoint quorum through them).

src/vsr/replica.zig
matklad: >Freely admitting requests into the prepare/request queue could get to a primary to a point where it hits [this assert](https://github.com/tigerbeetle/tigerbeetle/blob/main/src/vsr/replica.zig#L1645-L1655) in on_prepare, which we're aren't supposed to hit as a primary.

I'd say that the problem here is not that we admit a request to request queu, but rather than we promote it to a prepare

Eg, here:

```
            if (self.status == .normal and self.primary()) { to 
                if (self.pipeline.queue.pop_request()) |request| {
                    // Start preparing the next request in the queue (if any).
                }
```

I think we should add an if to check if we _can_ start preparing that op, and, correspondingly, add a call to `pop_request` to the place where we advance `op_prepare_max` on_ping

chaitanyabhandari: Agreed, I had explored the option of changing this in [commit_op_client_replies_ready_callback](https://github.com/tigerbeetle/tigerbeetle/blob/main/src/vsr/replica.zig#L3870-L3873), but realised that there are also certain requests that are implicitly sent ([send_pulse_to_self](https://github.com/tigerbeetle/tigerbeetle/blob/main/src/vsr/replica.zig#L9711), [send_request_upgrade_to_sef](https://github.com/tigerbeetle/tigerbeetle/blob/main/src/vsr/replica.zig#L9722)) which currently assume that there is always enough space in the prepare_queue (and consequently the WAL). This would spread out the code to check if there's enough space in the WAL over multiple functions, which is why I thought that containing it in _on_request_ be cleaner in comparison. Thoughts?

chaitanyabhandari: Also, I recall trying the approach of not popping the request from the request queue if we don't have enough space in the WAL (in [commit_op_client_replies_ready_callback](https://github.com/tigerbeetle/tigerbeetle/blob/main/src/vsr/replica.zig#L3870-L3873)). I believe we then end up violating [this](https://github.com/tigerbeetle/tigerbeetle/blob/main/src/vsr/replica.zig#L10319-L10321) invariant, since we pop a request from the prepare queue but we may not move one over from the request queue promptly.
```C
/// - If request_queue is not empty, then prepare_queue is full OR 1-less than full.
///   (The caller is responsible for maintaining this invariant. If the caller removes an entry
///   from `prepare_queue`, an entry from request_queue should be moved over promptly.)
```

I believe if we want to go down that route (while maintaining the invariant that the request queue is an overflow for the prepare queue), we would have to _empty_ the request queue if we find that we don't yet have the space in our WAL to prepare those requests?

src/vsr/replica.zig
```diff
@@ -5798,7 +5840,11 @@ pub fn ReplicaType(
         /// doing so would overwrite a message (or the slot of a message) that has not yet been
         /// committed and checkpointed.
         fn op_prepare_max(self: *const Self) u64 {
-            return vsr.Checkpoint.prepare_max_for_checkpoint(self.op_checkpoint_next()).?;
+            const checkpoint = if (self.checkpoint_quorum)
+                self.op_checkpoint_next()
+            else
+                self.op_checkpoint();
+            return vsr.Checkpoint.prepare_max_for_checkpoint(checkpoint).?;
         }
```
matklad, line 5881: Oh, I've realized something important: we only need to change the definition of `op_prepare_max` on the primary.

If you are a backup, and you receive a prepare, you don't actually have to check whether you know a quorum of replicas at the corresponding checkpoint. 

That is because, if you got the prepare, than it was prepared on the primary. But that means that the primary knows that a quorum of checkpoints exist!

So I think we could actually make this a smaller change: let's keep the definiton of `prepare_max` as is, meaning "the highest thig we can prepare without evicing current checkpoint's ops from the log", and instead just add an extra logic to on_request for the primary only. 

I like this approach more, as it keeps the defintion of prepare_max static.

src/vsr/replica.zig
matklad: So, in terms of what should be happening physically, I am rather sure that the request should be queued up, if we have available space. 

Dropping request is pretty bad --- that means that the client would have to do a round-trip!

So, the direct answer is that we should relax that invariant, and allow partially full prepare queue. 

That being said, that's going to be a larger refactor, so I am fine landing this if there's a TODO and a corresponding issue.

_That_ being said, even that would be unsatisfactory! Normaly, you can have both the prepare queue and request queue full --- that's the _depth_ of your pipeline preventing clients from retransmitting requests. 

If we implement this preepare_max limiting directly, we'll be in a situation where sometimes our `pipeline_queue` is not full. So, we actually _have_ RAM to  accept an extra request, but choose not to. 

I think the _perfect_ implementation would be to _add_ this overflowing request to a prepare queue, and instead just skip the part where we append it to our log and broadcast. That is, ideally, after we get the checkpoint confirmation, we should be able to, in parallel, append&broadcast a whole bunch of prepares at ones. 

And the existance of this alterantive solution actually warms me up quite a bit towards the simple approach we do in this PR for now!

chaitanyabhandari: @matklad beautiful! I had pondered on some of these ideas while reasoning about the semantics of the prepare queue & request queue as well (how we'd impose an extra RTT on the client even though we have space in our queues, relaxing the prepare queue invariant to avoid this), but I held off on proposing those because it was increasing the scope of this change a fair bit. Note to self - _be bolder next time_! 😎 

> And the existance of this alterantive solution actually warms me up quite a bit towards the simple approach we do in this PR for now!

Also, I agree with the _perfect_ implementation that you proposed! However, I have a small confusion about the "simple approach" that you talked about. You are referring to https://github.com/tigerbeetle/tigerbeetle/pull/2104#discussion_r1686483466, right?

src/vsr/replica.zig
chaitanyabhandari: That's a great observation, it would make the solution so clean and elegant! Rushing to try out this approach. 🏃🏼

src/vsr/replica.zig
matklad: No, the "simple approach" is "reject request before adding it to request queue", as opposed to "add request to request queue, but don't convert it to prepare until the checkpoint is confirmed"

chaitanyabhandari: Ah got it, thank you for clarifying! 👍🏼

src/vsr/replica.zig
chaitanyabhandari: Took a stab at this as part of the latest commit: https://github.com/tigerbeetle/tigerbeetle/pull/2104/files#diff-e95e4dc6c1aae349e798cfeed22ad01f830cce0e654e6b4b765b3e7b0c677b6aR5856 😄

src/vsr/replica.zig
```diff
@@ -9477,7 +9548,7 @@ pub fn ReplicaType(
             // could lead to permanent unavailability.
```
matklad, line 9543: The comment needs adjustment, as we now use "prepared atop" rather than "committed atop", though perhaps there's a better phrasing here.

src/vsr/replica.zig
```diff
@@ -5879,15 +5950,9 @@ pub fn ReplicaType(
                     break :repair_min 0;
                 }
 
-                const op_checkpoint_trigger =
```
matklad, line 5882: Hm, why are we looking at `checkpoint_quorum` here?  Shouldn't we just look at our `op`?

If op is beyond `op_prepare_max`, we can drop the old checkpoint. Otherwise, can keep it. 

That way, only the primary would call `checkpoint_quorum`, and we will just (implicitly, by observring a prepare beyond prepare_max) re-use that information.

chaitanyabhandari: I thought we could use *checkpoint_quorum* here because it nicely mirrors our comment about which ops satisfy both our safety & availability conditions:
```
        /// Repairing all ops since the latest checkpoint durable on a commit-quorum of replicas
        /// satisfies both conditions.
```

However, simply using our op & adding a comment should serve the same purpose as well!

src/vsr/replica.zig
chaitanyabhandari: Good catch!

chaitanyabhandari: https://github.com/tigerbeetle/tigerbeetle/pull/2104/files#diff-e95e4dc6c1aae349e798cfeed22ad01f830cce0e654e6b4b765b3e7b0c677b6aR9525-R9532

src/vsr/replica.zig
chaitanyabhandari: https://github.com/tigerbeetle/tigerbeetle/pull/2104/commits/3d24e4a6d05ec1ba516fa568ef9e4f5ebf721fd4 simplifies checkpoint_quorum() such that it is only invoked from the primary!

src/vsr.zig
```diff
@@ -1480,6 +1480,52 @@ pub const Checkpoint = struct {
         assert(constants.journal_slot_count > constants.lsm_batch_multiple);
         assert(constants.journal_slot_count % constants.lsm_batch_multiple == 0);
     }
+    /// Checkpoint identifier.
```
matklad: Useless comment!

Would be good to explain _what_ it is (checksum of the CheckpointState in the superblock).

src/vsr.zig
```diff
@@ -1480,6 +1480,52 @@ pub const Checkpoint = struct {
...
+            quorum.checkpoints[replica] = checkpoint.*;
+            return true;
+        }
+
+        pub fn count(quorum: *const Quorum, checkpoint: *const Checkpoint) usize {
+            var matching: usize = 0;
+            for (quorum.checkpoints) |checkpoint_existing| {
+                if (checkpoint_existing != null and
+                    checkpoint_existing.?.op == checkpoint.op and
+                    checkpoint_existing.?.id == checkpoint.id)
```
matklad: > `checkpoint_existing.?.id == checkpoint.id`

This should be an assert I think

src/vsr.zig
matklad: Yeah, it seems that its easy to _not_ include us here (but than, at the call-site, we should assert that our own checkpoint is `null`.

src/vsr/replica.zig
```diff
@@ -9744,7 +9810,9 @@ pub fn ReplicaType(
 
             const upgrade = vsr.UpgradeRequest{ .release = self.upgrade_release.? };
             self.send_request_to_self(.upgrade, std.mem.asBytes(&upgrade));
-            assert(self.pipeline.queue.contains_operation(.upgrade));
+
+            // This request may be rejected in on_request due to lack of space in the WAL.
+            maybe(self.pipeline.queue.contains_operation(.upgrade));
```
matklad, line 9817: Might want to add this case to `// Normally we chain send-upgrade-to-self via the commit chain.` comment

matklad:
This basically looks good to me now, all the remaining comments are nits.

Future work (be sure not to forget about this, or file the issues):

* think about not dropping the request, but somehow adding it to the pipeline, such that no round-trip is needed around checkpoints
* we should prevent the replica from advertising its checkpoint until it finished state sync (sync_content_done)
* that semi-related VOPR failure about journal truncation

chaitanyabhandari:
Thanks a lot for the thorough review, @matklad! I've addressed the remainder of the comments in https://github.com/tigerbeetle/tigerbeetle/pull/2104/commits/08eaaf0aceb00f18951a40b99182db2a2fd93f93.

Ack, taking note of the the future work! Thank you for 2nd point there, I hadn't thought of that one. 😄

src/vsr.zig
matklad: this one still stands!

chaitanyabhandari: Oops, had misinterpreted this the first time you posted it and then it got lost! Added in https://github.com/tigerbeetle/tigerbeetle/pull/2104/commits/f0221b6dc846c5dfbd0fc4f61b6fbcc1857183c8. 😄

matklad:
@chaitanyabhandari shell we close this PR?

chaitanyabhandari:
In this PR, we implemented an explicit synchronization point for the primary at op_prepare_max, wherein it would have to wait for a commit quorum of replicas to reach the former checkpoint before preparing past it.  We discovered (in [this](https://tigerbeetleinc.slack.com/archives/C03VBG9D2F7/p1721816599554649) Slack discussion) that there was a way to avoid this synchronization point by increasing the size of the journal (while consequently decreasing the checkpoint interval).

Closing this PR since we implemented the simpler fix in https://github.com/tigerbeetle/tigerbeetle/pull/2137.

PR #2031, opened 2024-06-19
Tidy: Ban std.mem.copy/copyForwards/copyBackwards

sentientwaffle:
Ban `std.mem`'s copy functions, in favor of the corresponding `stdx` implementations. The `stdx` implementations have extra assertions.

- `std.mem.copy` → `stdx.copy_disjoint`
- `std.mem.copyForwards` → `stdx.copy_left`
- `std.mem.copyBackwards` → `stdx.copy_right`

PR #2109, opened 2024-07-15
Fuzz `storage.zig` with mocked IO

SuperAuguste:
`storage.zig` is a crucial foundation of the whole TigerBeetle database, so it's mocked for testing in a lot of places. This leaves the real `storage.zig` code untested, until now!

This implements a mocked `testing/io.zig` which can emit faults to test `storage.zig`'s error recovery code. Note that no changes were made to the `storage.zig` logic itself (Git has trouble diffing changes in that file meaningfully I guess).


src/testing/io.zig
```diff
@@ -0,0 +1,229 @@
...
+const assert = std.debug.assert;
+const log = std.log.scoped(.io);
+
+const constants = @import("../constants.zig");
+const FIFO = @import("../fifo.zig").FIFO;
+const Time = @import("../time.zig").Time;
+const buffer_limit = @import("../io.zig").buffer_limit;
+const DirectIO = @import("../io.zig").DirectIO;
+
+/// Very simple mock IO implementation that only implements what is needed to test Storage
```
cb22: Nit: for all comments we like to use full sentences, with full grammar, so ending with a full stop here and below!

src/testing/io.zig
```diff
@@ -0,0 +1,229 @@
...
+        /// Chance out of 100 that a read larger than a logical sector
+        /// will return an error.InputOutput.
+        larger_than_logical_sector_read_fault_probability: u8 = 0,
+    };
+
+    files: []const []u8,
+    options: Options,
+    prng: std.rand.DefaultPrng,
+
+    time: Time = .{},
```
cb22: I think `time` is unused?

src/testing/io.zig
```diff
@@ -0,0 +1,229 @@
...
+        },
+        write: struct {
+            fd: posix.fd_t,
+            buf: [*]const u8,
+            len: u32,
+            offset: u64,
+        },
+    };
+
+    /// Returned memory can be invalidated upon call to open_in_memory_file
```
cb22: `open_in_memory_file`?

src/testing/io.zig
```diff
@@ -0,0 +1,229 @@
...
+    fn get_file_data(io: *IO, fd: posix.fd_t) []u8 {
+        return io.files[
+            switch (posix.fd_t) {
```
cb22: Could we use `fd: u32` (eg) and avoid this switch?

src/testing/io.zig
```diff
@@ -0,0 +1,229 @@
...
+    fn submit(
+        self: *IO,
+        context: anytype,
+        comptime callback: anytype,
+        completion: *Completion,
+        comptime operation_tag: std.meta.Tag(Operation),
+        operation_data: anytype,
+        comptime OperationImpl: type,
+    ) void {
+        const onCompleteFn = struct {
```
cb22: I know existing IO implementations are bad at it, but prefer `snake_case` for function names - even when vendoring code! (this one surprised me too, but see eg stdx/bounded_array.zig)

src/testing/io.zig
cb22: Or perhaps, we'd want our mocked `IO` to own the memory for its storage. So it could take in an allocator and sizes rather than buffers. Up to you, though!

src/testing/io.zig
```diff
@@ -0,0 +1,229 @@
...
+            fn onComplete(io: *IO, _completion: *Completion) void {
+                // Perform the actual operation
+                const op_data = &@field(_completion.operation, @tagName(operation_tag));
+                const result = OperationImpl.do_operation(io, op_data);
+
+                // Complete the Completion
+
```
cb22: Extra newline.

src/testing/io.zig
```diff
@@ -0,0 +1,229 @@
...
+                fn do_operation(io: *IO, op: anytype) ReadError!usize {
+                    if (io.x_in_100(io.options.read_fault_probability) or
+                        (op.len > constants.sector_size and
+                        io.x_in_100(io.options.larger_than_logical_sector_read_fault_probability)))
+                    {
+                        return error.InputOutput;
+                    }
+
+                    const data = io.get_file_data(op.fd);
+                    @memcpy(op.buf[0..op.len], data[op.offset..][0..op.len]);
```
cb22: Prefer [stdx.copy_disjoint](https://github.com/tigerbeetle/tigerbeetle/pull/2031) (I may have been the reason for DJ writing that PR :joy:).

src/testing/io.zig
```diff
@@ -0,0 +1,229 @@
...
+                .fd = fd,
+                .buf = buffer.ptr,
+                .len = @as(u32, @intCast(buffer_limit(buffer.len))),
+                .offset = offset,
+            },
+            struct {
+                fn do_operation(io: *IO, op: anytype) WriteError!usize {
+                    const data = io.get_file_data(op.fd);
+                    if (op.offset + op.len >= data.len) {
+                        @panic("TODO");
```
cb22:
```suggestion
                        @panic("write beyond simulated file size");
```

or similar.

src/storage.zig
```diff
@@ -7,400 +7,555 @@ const log = std.log.scoped(.storage);
...
-    }
+            assert(buffer.len > 0);
+        }
+    };
+}
 
-    fn start_read(self: *Storage, read: *Storage.Read, bytes_read: ?usize) void {
-        assert(read.offset % constants.sector_size == 0);
-        maybe(bytes_read == 0); // Retrying erroneous read at the same offset with smaller window.
+test "storage with mocked io layer" {
```
cb22: So this test seemed to run quite quickly - do you think it has enough coverage as is? Or would it benefit from being a standalone fuzz (like all the other _fuzz.zig) fuzzers?

(I ask, since if it would benefit from more time, then we likely want to split it out to keep our unit tests quick)

src/storage.zig
```diff
@@ -7,400 +7,555 @@ const log = std.log.scoped(.storage);
...
-            // Resolving the read inline means start_read() must not have been called from
-            // read_sectors(). If it was, this is a synchronous callback resolution and should
-            // be reported.
-            assert(bytes_read != null);
+    const storage_size = constants.sector_size * 64;
 
-            read.callback(read);
-            return;
-        }
+    var written_storage_data: [storage_size]u8 align(constants.sector_size) = undefined;
```
cb22: We prefer big endian naming (`storage_data_written`, `storage_data_stored` etc) so that the prefixes can line up!

src/storage.zig
```diff
@@ -7,400 +7,555 @@ const log = std.log.scoped(.storage);
...
-            error.Unexpected,
-            => {
-                log.err(
-                    "impossible read: offset={} buffer.len={} error={s}",
-                    .{ read.offset, read.buffer.len, @errorName(err) },
-                );
-                @panic("impossible read");
-            },
-        };
+        const write_buf = written_storage_data[zone.start() + offset_in_zone ..][0..len];
```
cb22: [Try to avoid abbreviations](https://github.com/tigerbeetle/tigerbeetle/blob/main/docs/TIGER_STYLE.md#developer-experience) where possible, so `write_buffer` and `length`.

src/storage.zig
```diff
@@ -7,400 +7,555 @@ const log = std.log.scoped(.storage);
...
+        // "larger than a single sector" case.
+        .read_fault_probability = 0,
+        .larger_than_logical_sector_read_fault_probability = 25,
+    });
+
+    var storage = try StorageWithMockedIO.init(&io, io.file_at_index(0));
+
+    var data_prng = std.rand.DefaultPrng.init(seed);
+
+    const n = 10_000;
```
cb22:
```suggestion
    const iterations = 10_000;
```

src/testing/io.zig
SuperAuguste: We can use `fd: u32` but it would require modifying all `posix.fd_t`s in the codebase to some sort of `IO.FileDescriptor` or the like. I think long-term it would make sense to do that (and maybe make all the IO in TigerBeetle use `IO` rather than the `posix` functions directly?), but not sure if it should be done in this PR.

The original implementation had an allocator and sizes but I found that this allocator-less implementation is a good simplification for our usecase that also enables us to use static allocation.

src/storage.zig
SuperAuguste: Very good point. I think turning it into a standalone fuzz would make the most sense. Will address everything else first and then take a go at that. :)

src/testing/io.zig
SuperAuguste: Oops this is from the original allocator-backed implementation I wrote 😅

src/testing/io.zig
cb22: Ah yes, I forgot that `Storage.init` takes in a `os.fd_t` and that's platform dependent :/.

Although, as you mention, I think we could get away with re-exporting `fd_t` within IO, so the testing one would have `pub const fd_t = u32;` and the others could `pub const fd_t = os.fd_t;`?

SuperAuguste: Oh great idea, I'll do that right now then!

src/fuzz_tests.zig
```diff
@@ -87,6 +88,7 @@ fn main_smoke() !void {
             .vsr_journal_format,
             .vsr_superblock_quorums,
             .lsm_scan,
+            .storage,
```
cb22, line 91: :+1: you'll want to add it to `cfo.zig` too.

src/storage_fuzz.zig
```diff
@@ -0,0 +1,135 @@
...
+        // "larger than a single sector" case.
+        .read_fault_probability = 0,
+        .larger_than_logical_sector_read_fault_probability = 25,
+    });
+
+    var storage = try Storage.init(&io, 0);
+
+    var data_prng = std.rand.DefaultPrng.init(args.seed);
+
+    const iterations = 10_000;
```
cb22:
```suggestion
    const iterations = args.events_max orelse 10_000;
```

src/testing/io.zig
```diff
@@ -0,0 +1,209 @@
...
+const os = std.os;
+const posix = std.posix;
+const mem = std.mem;
+const assert = std.debug.assert;
+const log = std.log.scoped(.io);
+
+const stdx = @import("../stdx.zig");
+const constants = @import("../constants.zig");
+const FIFO = @import("../fifo.zig").FIFO;
+const Time = @import("../time.zig").Time;
```
cb22: I think this import is unused.

src/testing/io.zig
```diff
@@ -0,0 +1,209 @@
...
+        larger_than_logical_sector_read_fault_probability: u8 = 0,
+    };
+
+    files: []const []u8,
+    options: Options,
+    prng: std.rand.DefaultPrng,
+
+    completed: FIFO(Completion) = .{ .name = "io_completed" },
+
+    pub fn init(files: []const []u8, options: Options) IO {
```
cb22: Can we assert files.len > 0?

src/testing/io.zig
```diff
@@ -0,0 +1,209 @@
...
+        self: *IO,
+        context: anytype,
+        comptime callback: anytype,
+        completion: *Completion,
+        comptime operation_tag: std.meta.Tag(Operation),
+        operation_data: anytype,
+        comptime OperationImpl: type,
+    ) void {
+        const on_complete_fn = struct {
+            fn onComplete(io: *IO, _completion: *Completion) void {
```
cb22:
```suggestion
            fn on_complete(io: *IO, _completion: *Completion) void {
```

cb22 (commented):
Looking good! A few minor nits and then let's get it merged!

src/testing/io.zig
SuperAuguste: Having zero files is perfectly fine and supported, although it would be weird. I'll add an assertion in `read` and `write` to ensure that the `fd_t` is within the length of `files`, though.

SuperAuguste:
Pushed the new `fault_map` based fault injection but still thinking of a nice way to generate clustered maps. Will push that as soon as I come up with something.

src/storage_fuzz.zig
```diff
@@ -0,0 +1,164 @@
...
+            &write_completion,
+            storage_data_written[zone.start()..][0..zone.size().?],
+            zone,
+            0,
+        );
+
+        storage.tick();
+    }
+
+    for (0..iterations) |_| {
```
cb22: Let's move the fault map generation and writes from above to within this loop, so the fuzzer tests different fault maps for each iteration.

src/storage_fuzz.zig
```diff
@@ -0,0 +1,164 @@
...
+        .client_replies,
+    };
+
+    const sector_count = 64;
+    const storage_size = sector_count * constants.sector_size;
+    const iterations = args.events_max orelse 10_000;
+
+    const failed_sector_cluster_count = 7;
+    const failed_sector_cluster_minimum_length = 1;
+    const failed_sector_cluster_maximum_length = 4;
```
cb22: We can also generate these randomly (along with sector count) from within a range, to maximize our fuzz surface area!

cb22 (approved):
:rocket:

PR #2116, opened 2024-07-16
cfo: don't crash if the set of fuzzers change

matklad:
Before this commit, if a fuzzer is removed from the `Fuzzer` enum, the cfo would crash when attempting to deserialize old seed records. Fix this, by treating a fuzzer as `[]const u8` for the purposes of deserialization.

### DIFF
--- a/src/scripts/cfo.zig
+++ b/src/scripts/cfo.zig
@@ -165,7 +165,7 @@ fn run_fuzzers(
         log.info("fuzzing commit={s} timestamp={} fuzzer={s} branch='{s}' weight={}", .{
             seed_record.commit_sha[0..7],
             seed_record.commit_timestamp,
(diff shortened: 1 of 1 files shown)


src/scripts/cfo.zig
```diff
@@ -998,4 +1005,60 @@ test "cfo: SeedRecord.merge" {
...
+                .seed = 1,
+                .command = "fuzz ewah",
+                .branch = "main",
+            },
+        },
+        &.{
+            .{
+                .commit_timestamp = 1,
+                .commit_sha = .{'1'} ** 40,
+                .fuzzer = "American Fuzzy Lop",
```
batiati, line 1028: 😜

batiati (approved):

PR #2117, opened 2024-07-17
Fix incorrect `timestamp_max` max value.

batiati:
Since the most significant bit of the timestamp is the tombstone flag, we need to fix the `timestamp_max` as a `maxInt(u63)`.

### DIFF
--- a/docs/reference/account-filter.md
+++ b/docs/reference/account-filter.md
@@ -27,7 +27,7 @@ Optional; set to zero to disable the lower-bound filter.
 Constraints:
 
 - Type is 64-bit unsigned integer (8 bytes)
-- Must not be `2^64 - 1`
+- Must not overflow a 64-bit signed integer (`2^63 - 1`).
 
 ### `timestamp_max`
 
@@ -37,7 +37,7 @@ Optional; set to zero to disable the upper-bound filter.
 Constraints:
 
 - Type is 64-bit unsigned integer (8 bytes)
-- Must not be `2^64 - 1`
+- Must not overflow a 64-bit signed integer (`2^63 - 1`).
 
 ### `limit`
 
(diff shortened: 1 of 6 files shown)


docs/reference/account-filter.md
```diff
@@ -27,7 +27,7 @@ Optional; set to zero to disable the lower-bound filter.
 Constraints:
 
 - Type is 64-bit unsigned integer (8 bytes)
-- Must not be `2^64 - 1`
+- Must not overflow a 64-bit signed integer (`2^63 - 1`).
```
sentientwaffle: I don't think we should mention signed ints here... `-1` doesn't overflow a 64-bit signed int, but isn't a valid value here. Specifying a maximum value is simpler.

```suggestion
- Must be less than `2^63`.
```

(Likewise for `create_transfers.md`.)

src/lsm/timestamp_range.zig
```diff
@@ -2,7 +2,7 @@ const std = @import("std");
 
 pub const TimestampRange = struct {
     pub const timestamp_min = 1;
-    pub const timestamp_max = std.math.maxInt(u64) - 1;
+    pub const timestamp_max = std.math.maxInt(u63); // The last bit is the tombstone flag.
```
sentientwaffle: Maybe add a note that the maximum is inclusive.

src/state_machine.zig
```diff
@@ -1840,9 +1842,9 @@ pub fn StateMachineType(
             }
 
             if (sum_overflows(
-                u64,
-                t.timestamp,
-                @as(u64, t.timeout) * std.time.ns_per_s,
+                u63,
+                @intCast(t.timestamp),
+                @as(u63, t.timeout) * std.time.ns_per_s,
```
sentientwaffle, line 1856: Couldn't this multiplication overflow too? (Before the sum.)

sentientwaffle: Ahh right, `timeout` is only a `u32`!

sentientwaffle (dismissed):

batiati:
Sorry @sentientwaffle I added a couple more asserts after your approval!

sentientwaffle (dismissed):

sentientwaffle (approved):

PR #2120, opened 2024-07-19
Fix `pulse` in idle clusters.

### LINKED ISSUE #2118: Inconsistent reporting of account balance by lookup_accounts for expired pending transfers
I stumbled upon what seems to me as a bug. I verified it with latest master build.

So far my intuition about pending transfers in TBDB was that when such transfer expires, the subsequent call to `lookup_accounts` will return updated (voided) pending balances, but this is not the observed behaviour.

Here's a demonstrator code:

```java
public class ReserveFunds {

    private static final Client client = new Client(UInt128.asBytes(0), new String[]{"10.8.0.1:3000"});

    public static void main(String[] args) throws InterruptedException {
        int ledgerId = 61;
        UUID debitAccountId = UUID.fromString("be3999e7-70ed-430c-ab38-b8527ecdf52d");
        UUID creditAccountId = UUID.fromString("07c9ff85-2ca8-489c-8053-02ab2c09b0f1");

        System.out.println("Balance before: " + getAccountBalance(debitAccountId)); // initial balance is 10
        reserve(UUID.randomUUID(), ledgerId, debitAccountId, creditAccountId, 10, 5); // debit 10, expires in 5 seconds
        Thread.sleep(10 * 1000); // wait 10 seconds
        System.out.println("Balance after: " + getAccountBalance(debitAccountId)); // 5 seconds after expiry, the balance is still 0
    }

    public static void reserve(UUID reserveTransferId,
                               int ledgerId,
                               UUID debitAccountId,
                               UUID creditAccountId,
                               long amount,
                               int timeout) {
        TransferBatch transferBatch = new TransferBatch(1);
        transferBatch.add();
        transferBatch.setId(UInt128.asBytes(reserveTransferId));
        transferBatch.setFlags(TransferFlags.PENDING);
        transferBatch.setLedger(ledgerId);
        transferBatch.setCode(3); // code=3 is for reserve
        transferBatch.setDebitAccountId(UInt128.asBytes(debitAccountId));
        transferBatch.setCreditAccountId(UInt128.asBytes(creditAccountId));
        transferBatch.setAmount(amount);
        transferBatch.setTimeout(timeout);

        try {
            CreateTransferResultBatch resultBatch = client.createTransfers(transferBatch);
            if (resultBatch.next()) {
                throw new RuntimeException("CreateTransferResult=" + resultBatch.getResult());
            }
        } catch (ConcurrencyExceededException e) {
            throw new RuntimeException(e);
        }
    }

    public static long getAccountBalance(UUID accountId) {
        try {
            IdBatch idBatch = new IdBatch(UInt128.asBytes(accountId));
            AccountBatch accountBatch = client.lookupAccounts(idBatch);
            if (!accountBatch.next()) {
                throw new RuntimeException("Account Not Found");
            }

            long credits = accountBatch.getCreditsPosted(UInt128.LeastSignificant) +
                    accountBatch.getCreditsPending(UInt128.LeastSignificant);
            long debits = accountBatch.getDebitsPosted(UInt128.LeastSignificant) +
                    accountBatch.getDebitsPending(UInt128.LeastSignificant);

            return credits - debits;
        } catch (ConcurrencyExceededException e) {
            throw new RuntimeException(e);
        }
    }

}
```

I would expect to get balance=10 after the sleep (5 seconds since the transfer expired), but I still get balance=0.
But, when I re-run this code immediately after, then the first getAccountBalance properly returns balance=10, as I would expect.

What is more interesting - while the original process sleeps in the `Thread.sleep(10 * 1000)`, should there be another client connection (in different process) that requests `lookup_accounts` (not necessarily for accounts involved in the ongoing pending transfer) then once `Thread.sleep(10 * 1000)` is done the `getAccountBalance(debitAccountId)` reports proper balance=10.

What is going on here?

batiati:
The state machine depended on `prepare_timestamp` to evaluate `pulse()`, but in an idle cluster, `prepare_timestamp` would only be set if `pulse` returned true!

Added unity tests for clients (except Go which does not have integration tests for any of the two-phase transfer scenarios)

Closes #2118

### DIFF
--- a/src/vsr/replica.zig
+++ b/src/vsr/replica.zig
@@ -9718,6 +9718,8 @@ pub fn ReplicaType(
             assert(self.primary());
             assert(!self.pipeline.queue.full());
             assert(!self.pipeline.queue.contains_operation(.pulse));
+            assert(self.pulse_enabled());
+            assert(self.state_machine.pulse_needed(self.state_machine.prepare_timestamp));
 
             self.send_request_to_self(.pulse, &.{});
             assert(self.pipeline.queue.contains_operation(.pulse));
--- a/src/vsr/replica.zig
+++ b/src/vsr/replica.zig
@@ -3178,16 +3178,20 @@ pub fn ReplicaType(
 
             self.pulse_timeout.reset();
             if (self.pipeline.queue.full()) return;
-            if (!self.pulse_needed()) return;
+            if (!self.pulse_enabled()) return;
 
             // To decide whether or not to `pulse` a time-dependant
             // operation, the State Machine needs an updated `prepare_timestamp`.
             const realtime = self.clock.realtime();
-            self.state_machine.prepare_timestamp = @max(
+            const timestamp = @max(
                 self.state_machine.prepare_timestamp,
                 @as(u64, @intCast(realtime)),
             );
-            self.send_request_pulse_to_self();
+
+            if (self.state_machine.pulse_needed(timestamp)) {
+                self.state_machine.prepare_timestamp = timestamp;
+                self.send_request_pulse_to_self();
+            }
         }
 
         fn on_upgrade_timeout(self: *Self) void {
@@ -3872,7 +3876,9 @@ pub fn ReplicaType(
                     self.primary_pipeline_prepare(request);
                 }
 
-                if (self.pulse_needed()) {
+                if (self.pulse_enabled() and
+                    self.state_machine.pulse_needed(self.state_machine.prepare_timestamp))
+                {
                     assert(self.upgrade_release == null);
                     self.send_request_pulse_to_self();
                 }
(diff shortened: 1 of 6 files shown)


src/vsr/replica.zig
```diff
@@ -3872,7 +3876,9 @@ pub fn ReplicaType(
                     self.primary_pipeline_prepare(request);
                 }
 
-                if (self.pulse_needed()) {
+                if (self.pulse_enabled() and
+                    self.state_machine.pulse_needed(self.state_machine.prepare_timestamp))
+                {
                     assert(self.upgrade_release == null);
                     self.send_request_pulse_to_self();
```
sentientwaffle, line 3883: Can `send_request_pulse_to_self` assert

```
assert(self.pulse_enabled());
assert(self.state_machine.pulse_needed(self.state_machine.prepare_timestamp));
```

sentientwaffle (dismissed):
Aside from a couple asserts, this LGTM, but I think we should leave it running in the CFO over the weekend.

sentientwaffle (approved):

PR #2122, opened 2024-07-21
Added error handling for limits of byte-oriented command line args

### LINKED ISSUE #1464: Add error handling for limits of byte-oriented command line args
Some command line arguments have error handling for cases where the arguments are outside of reasonable limits. Several do not. Here are some that I see offhand:

- `--cache-grid` - Extremely large values hit an `intCast` assertion. Unreasonably large values are accepted (1000000GB).
- `--cache-accounts` - small values cause "unreachable" assertions, large values `intCast`
- `--cache-transfers` - ditto

dadepo:
Closes #1464


src/vsr/grid.zig
```diff
@@ -26,7 +26,7 @@ pub const BlockPtrConst = *align(constants.sector_size) const [constants.block_s
 pub fn allocate_block(
     allocator: mem.Allocator,
 ) error{OutOfMemory}!*align(constants.sector_size) [constants.block_size]u8 {
-    const block = try allocator.alignedAlloc(u8, constants.sector_size, constants.block_size);
+    const block = allocator.alignedAlloc(u8, constants.sector_size, constants.block_size) catch @panic("OOM");
```
dadepo: Not sure what value of `cache-grid` that can be considered to be too unreasonably large to reject during cli parsing. But it seems the effect of a large value is an OOM at this point.

src/tigerbeetle/cli.zig
```diff
@@ -856,7 +866,11 @@ fn parse_cache_size_to_count(
         value_count_max_multiple,
     ) * value_count_max_multiple;
 
-    const result: u32 = @intCast(count_rounded); // TODO: better error message on overflow
```
matklad, line 897: Rather than propagating the error upwards, let's call `.fatal` here like we do in `parse_addresses`. This will probably require adding the name of the CLI flag as a parameter to produce a nice error message.

src/vsr/grid.zig
matklad: I think this code doesn't need changes: propagating `OOM` as an error up the stack is the correct behavior here.

matklad (approved):

matklad (changes requested):
Sorry, didn't mean to approve just yet, see comments in the previous reviews!

src/tigerbeetle/cli.zig
dadepo: > This will probably require adding the name of the CLI flag as a parameter to produce a nice error message.

In my limited knowledge of the code base, won't extending `parse_cache_size_to_count` to include the name of the CLI flag feels a bit tacky? But I am fine adding it, if it fits with the style of the codebase.

The other way could be to do some comptime mapping of first argument to the CLI flag. ie. `tigerbeetle.Account => cache-accounts`, `tigerbeetle.Transfer => cache-transfers` etc.

Let me know which approach is the more preferred :)

matklad: Adding a string argument is the way to go, as that's the simplest solution. 

Returning an error via `!` is problematic, because it gives the caller a choice between propagating the error further up the stack or handling it, but, in this case, we don't need this choice, because we already know the right answer --- exit with fatal error.

matklad:
Fails CI at the moment!

dadepo:
> Fails CI at the moment!

...right about the same time I pushed the fix :)

dadepo:
Now just need to resolve the conflict...

src/tigerbeetle/cli.zig
```diff
@@ -885,6 +890,7 @@ fn parse_cache_size_to_count(
     comptime T: type,
     comptime SetAssociativeCache: type,
     size: flags.ByteSize,
+    cliFlag: []const u8,
```
matklad: We use `snake_case`

matklad (approved):
Thanks!

PR #2013, opened 2024-06-14
multiversioning: bring up multiversion binaries and build process for linux

cb22:
## Overview
This PR adds the platform specific support for multiversion binaries on Linux! 

The idea behind multiversion binaries is to give operators a great experience when upgrading TigerBeetle clusters. Upgrades should be simple, involve minimal downtime and be robust, while not requiring external coordination. Multiple versions in a single binary are required for two reasons:
* It allows a replica to crash after the binary has been upgraded, and still come back online.
	* It also allows for deployments, like Docker, where the binary is immutable and the process has to be terminated to learn about new versions from itself.
* It allows for migrations over a range to happen easily without having to manually jump from version to version.

The upgrade instructions look something like:

```
# SSH to each replica, in no particular order:
cd /tmp
wget https://github.com/tigerbeetle/tigerbeetle/releases/download/0.15.4/tigerbeetle-x86_64-linux.zip
unzip tigerbeetle-x86_64-linux.zip
rm /usr/bin/tigerbeetle
mv tigerbeetle /usr/bin/tigerbeetle
```

When the primary determines enough replicas have the new binary, it'll coordinate the upgrade (#1670). There are three main parts to multiversion binaries: building, monitoring and executing, with platform specific parts in each.

## Building
Physically, multiversion binaries are regular TigerBeetle ELF files that have two extra sections embedded into them - marked as noload so that they're not memory mapped:
* `.tbmvm` or TigerBeetleMultiVersionMetadata - a metadata struct containing information on past versions embedded as well as offsets, sizes, checksums and the like.
* `.tbmvp` or TigerBeetleMultiVersionPack - a concatenated pack of binaries. The offsets in `.tvmvm` refer into here.

(the short names are for compatibility with Windows / PE when that gets added.)

These are added by an explicit objcopy step in the release process, _after_ the regular build is done. After the epoch, the build process only needs to pull the last TigerBeetle release from GitHub, to access its embedded pack to build its own.

## Monitoring
On a 1 second timer, TigerBeetle `stat`s its binary file, looking for changes. Should anything differ (besides `atime`) it'll re-read the binary into memory, verify checksums and metadata, and start advertising new versions without requiring a restart. 

This optimization allows skipping a potentially expensive WAL replay when upgrading: the previous version is what will checkpoint to the new version, at which point the exec happens.

## Executing
[...]

PR #2123, opened 2024-07-22
devhub/checksum: enforce hardware acceleration

cb22:
Since the Zig version bump, when being built with an explicit target (`zig/zig build -Dtarget=x86_64-linux` - like the release script does) `build.zig` doesn't override the CPU flags properly so the resulting binary is compiled without AES-NI / AVX2, and uses a software AES implementation for AEGIS.

This didn't show up previously in the devhub results, because that builds its own binary without an explicit target. The performance different is big - 99k TPS with how the release script builds for 1M transfers, vs 246k TPS with the proper flags.

We haven't actually had a release since this version bump, so this isn't affecting anyone yet! This PR does a few things:

- Enforces at `comptime` that hardware acceleration is available. You'll want to disable this if you're building TigerBeetle targeting an older platform.
  - Currently this is failing, because clients are explicitly built as baseline...
- Adds checksum performance data to the benchmark and devhub, for `constants.message_size_max`.
  - Additionally, there was seemingly a ~3-4x slowdown from Zig 0.11 to Zig 0.13 for Aegis checksums (38MB, 90ms 0.13 vs 25ms 0.11).
- Uses the "real" output of `zig build scripts -- release` as the input to the devhub benchmark, and tests this build command on `main` as opposed to only on the `release` branch.
- Enables hardware acceleration for client builds instead of baseline. 

Notably, this PR does not actually fix the `build.zig` issue yet :smile:.


matklad (approved):
LGTM as is, but there are a couple things I'd love to do here and/or in a follow up:

* call `scripts -- release` from devhub, rather than from yaml, and use that as well for measuring binary size and build time (this might require adding a `--platform` flag to release infra to build a single platform only)
* when we benchmark checksum, it'd be good to assert that the result is what we expect, to prevent compiler from optimizing it away and a sanity check that it is correct (we can do that if the seed is the default one at least).

I'd also like to revisit our build.zig logic --- it should be fixed there obviously!

cb22:
> this might require adding a --platform flag to release infra to build a single platform only

I know @sentientwaffle [would like that too](https://github.com/tigerbeetle/tigerbeetle/pull/2013#discussion_r1684806220) :smile:.

PR #2124, opened 2024-07-22
cfo: improve canary efficiency

matklad:
One of the CFO entries right now is

    afd9c34|canary|./zig/zig build -Drelease fuzz -- canary 10948499344078146000|1s|22h 44m 1s ago

With the current logic, we know that we fuzzed that commit 22h ago, but we can't say whether we are still fuzzing it or not, because the _oldest_ canary seed is preferred by the seed merging logic.

Flip that to the newest seed, so that we clearly see that a canary is alive.

For all other seeds, preserving the oldest is the right call, as it adds stability.


chaitanyabhandari (approved):
First look at our snapshot testing library and it's so cool! 🚀

PR #2125, opened 2024-07-22
build: unbreak CPU feature selection

matklad:
This fixes the issue when, after Zig 0.13, TigerBeetle would sometimes fall back to software implementaiton of AEGIS.

What we need to do is to just enable a couple of CPU features. Previously, the logic worked by first performing normal target resolution in build.zig, and then patching the target to what we need.

However, after 0.13 the patching no longer works: the features is enabled in build.zig, but not in the actual binary. This is _probably_ due to discrepancy between Target.Query and ResolvedTarget, but I wasn't able to trace this to the root.

Instead, the entire logic is re-written to just pick one of the set of hard-coded targets.

As a bonus, `scripts -- release` now accepts optional `--target` and `--mode` flags, to compile only subset of release binaries.

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -7,11 +7,10 @@ const Mode = std.builtin.Mode;
 const config = @import("./src/config.zig");
 
 // TigerBeetle binary requires certain CPU feature and supports a closed set of CPUs. Here, we
-// specify exactly which features the binary needs. Client shared libraries might be more lax with
-// CPU features required.
-const supported_targets: []const CrossTarget = supported_targets: {
-    @setEvalBranchQuota(100_000);
-    var result: []const CrossTarget = &.{};
+// specify exactly which features the binary needs.
+fn resolve_target(b: *std.Build, target_requested: ?[]const u8) !std.Build.ResolvedTarget {
+    const target_host = @tagName(builtin.target.cpu.arch) ++ "-" ++ @tagName(builtin.target.os.tag);
+    const target = target_requested orelse target_host;
     const triples = .{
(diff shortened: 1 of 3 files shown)


cb22:
Seems some stray rebase markers made it in!

matklad:
Oh, dear, I don't know what it means to build for a single platform with multiversioning :rofl: 

Backed off the release.zig changes for now!

build.zig
```diff
@@ -864,6 +870,7 @@ fn set_cpu_features(
     };
     target_requested.result.cpu.features.addFeatureSet(target_supported.cpu_features_add);
     target_requested.result.cpu.features.removeFeatureSet(target_supported.cpu_features_sub);
+    std.debug.print("{}\n", .{target_requested.result.cpu.features});
```
cb22: Are this and the above `print` meant to be here?

src/scripts/devhub.zig
```diff
@@ -34,6 +34,11 @@ pub fn main(shell: *Shell, gpa: std.mem.Allocator, cli_args: CliArgs) !void {
     const executable_size_bytes = (try shell.cwd.statFile("tigerbeetle")).size;
     try shell.project_root.deleteFile("tigerbeetle");
 
+    try shell.zig(
+        \\build scripts -- release --build --run-number=200 --sha={sha}
```
cb22: Can we change this to 189 - the current release.zig asserts that it's being called with `0.15.4` as a target.

cb22 (approved):

PR #2127, opened 2024-07-23
ci: check x86_64 macos and clients on mac

cb22:
(This is on top of #2125)

We support x86_64 macOS, but don't run any tests on it. This runs our tests on `macos-13`, which is the last free GitHub runner to be on x86_64.

Both the test suite, and the go and Node clients (but also for macos-latest). Limited to those two, since concurrency limits mean that adding in more in the matrix will block waiting for runners...

This feels a bit icky, but we do need to:
* run our test suite on macOS,
* test macOS x86_64, if we support it. (or we drop it)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -75,7 +75,13 @@ jobs:
       - run: .\zig\zig build clients:c:sample
 
   test_macos:
-    runs-on: macos-latest
+    strategy:
+      matrix:
+        include:
+          - { os: 'macos-latest' }
+          - { os: 'macos-13' }
+
+    runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v4
         with:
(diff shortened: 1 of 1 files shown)


matklad:
I suggest dropping support for x86_64 macs

.github/workflows/ci.yml
```diff
@@ -103,6 +109,13 @@ jobs:
           - { os: 'windows-latest', language: 'node',    language_version: '18.x'  }
           - { os: 'windows-latest', language: 'node',    language_version: '20.x'  }
 
+          # Limited matrix for macOS - runners are concurrency limited.
+          - { os: 'macos-latest',   language: 'go',      language_version: '1.21'  }
+          - { os: 'macos-latest',   language: 'node',    language_version: '20.x'  }
+
+          - { os: 'macos-13',       language: 'go',      language_version: '1.21'  }
+          - { os: 'macos-13',       language: 'node',    language_version: '20.x'  }
+
```
matklad, line 118: Wait, we _weren't_ testing clients on macs at all? :facepalm:

matklad (dismissed):
Last commit LGTM, but would need a rebase!

matklad (approved):

PR #2128, opened 2024-07-23
devhub: Add GITHUB_TOKEN

cb22:
Fixes https://github.com/tigerbeetle/tigerbeetle/actions/runs/10057090484/job/27797326947

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -147,6 +147,7 @@ jobs:
         env:
           DEVHUBDB_PAT: ${{ secrets.DEVHUBDB_PAT }}
           NYRKIO_TOKEN: ${{ secrets.NYRKIO_TOKEN }}
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
       - uses: actions/upload-pages-artifact@v3
         with:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -10,6 +10,9 @@ jobs:
   release:
     runs-on: ubuntu-latest
     environment: release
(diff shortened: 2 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -147,6 +147,7 @@ jobs:
         env:
           DEVHUBDB_PAT: ${{ secrets.DEVHUBDB_PAT }}
           NYRKIO_TOKEN: ${{ secrets.NYRKIO_TOKEN }}
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
matklad, line 150: Hm, shouldn't this literally be ` GH_TOKEN: ${{ github.token }}`, such that its a read-only token?

The `secrets.GITHUB_TOKEN` can public a new release, and that's not something we want to be able do from within devhub

cb22: So `github.token` and `secrets.GITHUB_TOKEN` are [identical](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#about-the-github_token-secret) according to GitHub.

I think we should change `GITHUB_TOKEN` to be [default restrictive](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/enabling-features-for-your-repository/managing-github-actions-settings-for-a-repository#setting-the-permissions-of-the-github_token-for-your-repository), and then add back permissions for publish.

matklad: Oh, excellent catch, thanks!

matklad (approved):

cb22:
I've changed the GitHub permissions setting:
image

PR #2131, opened 2024-07-23
cfo: maintain the count of successful seeds

matklad:
Add a `count: u32` field to seeds. When merging two seeds, sum up the counts.

As a result, for successful seeds, the VOPR would now maintain the total number of fuzzer runs.

This changes the big design here somewhat: the original idea was to keep the set of seeds stable, such that the data in the shared git repository updated only when there are new commits or new failing seeds.

But it seems that the count info is sufficiently useful to justify hammering github after every CFO round (the round is about 30 minutes, so its not too high load).

Note: this PR doesn't yet do any visualization of counts


matklad:
Praise the mere queue! This PR had a semantic merge conflict, such that tests passed on the branch, the code merges cleanly, but would have failed CI if merged into main!

kprotty (approved):

PR #2133, opened 2024-07-24
cli: add napking benchmark to print static dimensions of TigerBeetle

matklad:


matklad:
Not sure if I want to actually merge this, but I've been thinking about something like this for a long time.

Would be a good place to, eg, show relative contribution of various trees to our memory usage and such.

matklad:
Example outout:

```
λ ./zig/zig build run -Dconfig=production -- benchmark --napkin

# Journal
journal_slot_count               = 1024
checkpoint_interval              = 960
lsm_batch_multiple               = 32
pipeline_prepare_queue_max       = 8

# State Machine
prepare_body_size                = 1048320 bytes (~1023 KiB)
prepare_body_size_batch          = 33546240 bytes (~31 MiB)
transfers_max                    = 8190
accounts_max                     = 8190
transfers_max_batch              = 262080
accounts_max_batch               = 262080
```

PR #2134, opened 2024-07-24
Test/Integration: Smoke test CLI help/version commands

sentientwaffle:


src/integration_tests.zig
```diff
@@ -269,3 +269,22 @@ test "benchmark/inspect smoke" {
...
+        "{tigerbeetle} --help",
+        "{tigerbeetle} inspect --help",
+        "{tigerbeetle} version",
+        // TODO(Multiversioning): Uncomment once this is fixed.
+        // "{tigerbeetle} version --verbose",
+    }) |command| {
+        const status_ok =
+            try shell.exec_status_ok(command, .{ .tigerbeetle = tigerbeetle });
+        try std.testing.expect(status_ok);
+    }
```
matklad, line 291: Let's also smoke test that the output contains some string, such that it is stable, but we still catch the issue where, eg, we use a buffered writer and forget to flush it.

Maybe also check that the output is not too short or something

cb22 (approved):

PR #2136, opened 2024-07-24
multiversioning: bring up mac and windows support

cb22:
Following on from #2013, this brings multiversion support to Windows and macOS. Multiversion, just like running a replica on these platforms, is considered a development tool and not suitable for production.

In particular, there's no monitoring and automatic restarting if the binary changes, and more of a best effort to not panic in the `parse_{pe,macho}` than an exhaustive attempt like `parse_elf`. These platforms also use physical files on disk, so they might be prone to more strange side-effects than Linux, which uses a memfd. Indeed, they also use sync IO instead of an event loop implementation for `openat` too.

Some known caveats:
* Windows terminal interfacing is strange, the process retains stdout/err when it reexecutes (CreateProcess) but it doesn't seem like it's possible to get it to keep stdin. Spawning a whole new terminal window is another option.

Perhaps one other notable change - in the build process, everything is collected under `multiversion-build/` and that folder is deleted at the end. It makes the cleanup code much easier, as well as making it easier to manually inspect things by disabling the defer.

### DIFF
--- a/src/config.zig
+++ b/src/config.zig
@@ -141,7 +141,13 @@ const ConfigProcess = struct {
     grid_scrubber_interval_ms_min: usize = std.time.ms_per_s / 20,
     grid_scrubber_interval_ms_max: usize = std.time.ms_per_s * 10,
     aof_recovery: bool = false,
-    multiversion_binary_size_max: u64 = 64 * 1024 * 1024,
+    multiversion_binary_platform_size_max: u64 = blk: {
+        // {Linux, Windows} get 64MB. macOS gets 128MB since it has universal binaries. Both of
+        // these are doubled in debug.
+        var size_max_mb = if (builtin.target.os.tag == .macos) 128 else 64;
+        if (builtin.mode != .ReleaseSafe) size_max_mb *= 2;
+        break :blk size_max_mb;
+    } * 1024 * 1024,
     multiversion_poll_interval_ms: u64 = 1000,
 };
 
--- a/src/config.zig
+++ b/src/config.zig
@@ -141,13 +141,7 @@ const ConfigProcess = struct {
     grid_scrubber_interval_ms_min: usize = std.time.ms_per_s / 20,
     grid_scrubber_interval_ms_max: usize = std.time.ms_per_s * 10,
     aof_recovery: bool = false,
-    multiversion_binary_platform_size_max: u64 = blk: {
-        // {Linux, Windows} get 64MB. macOS gets 128MB since it has universal binaries. Both of
-        // these are doubled in debug.
-        var size_max_mb = if (builtin.target.os.tag == .macos) 128 else 64;
-        if (builtin.mode != .ReleaseSafe) size_max_mb *= 2;
-        break :blk size_max_mb;
-    } * 1024 * 1024,
+    multiversion_binary_platform_size_max: u64 = 64 * 1024 * 1024,
     multiversion_poll_interval_ms: u64 = 1000,
 };
 
--- a/src/multiversioning.zig
+++ b/src/multiversioning.zig
@@ -513,25 +513,29 @@ test "MultiversionHeader.advertisable" {
     }
 }
 
-const memfd_path = "tigerbeetle-multiversion";
+const multiversion_uuid = "tigerbeetle-multiversion-1768a738-ef69-4605-8b5c-c6e63580e345";
 
 pub const Multiversion = struct {
-    const ArgsEnvp = struct {
+    const ArgsEnvp = if (builtin.target.os.tag == .windows) void else struct {
         // Coerces to [*:null]const ?[*:0]const u8 but lets us keep information to free the memory
         // later.
         args: [:null]?[*:0]const u8,
         envp: [*:null]const ?[*:0]const u8,
     };
 
+    const ExePathFormat = enum { elf, pe, macho, auto };
+
     io: *IO,
 
     exe_path: [:0]const u8,
+    exe_path_format: ExePathFormat,
     args_envp: ArgsEnvp,
 
     source_buffer: []u8,
     source_fd: ?posix.fd_t = null,
 
     target_fd: posix.fd_t,
+    target_path: [:0]const u8,
     target_body_offset: ?u32 = null,
     target_header: ?MultiversionHeader = null,
     /// This list is referenced by `Replica.releases_bundled`.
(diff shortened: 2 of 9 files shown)


src/multiversioning.zig
```diff
@@ -665,6 +737,9 @@ pub const Multiversion = struct {
     fn timeout_start(self: *Multiversion) void {
         if (!self.timeout_start_enabled) return;
 
+        // This is tested elsewhere, but needed to not codegen.
+        if (builtin.target.os.tag != .linux) return;
```
sentientwaffle: If the condition is actually unreachable, then:

```suggestion
        if (builtin.target.os.tag != .linux) unreachable;
```

Or if that doesn't make codegen happy, then we could assert immediately before the `if`.

src/tigerbeetle/main.zig
```diff
@@ -516,6 +511,22 @@ fn replica_release_execute(replica: *Replica, release: vsr.Release) noreturn {
         @panic("release not available");
     }
 
+    if (builtin.os.tag == .windows) {
+        // Unlike on Linux / macOS which use `execve{at,z}` for multiversion binaries,
+        // Windows has to use CreateProcess. This is a problem, because it's a race between
+        // the parent process exiting and the new process starting. Work around this by
+        // deinit'ing Replica and storage before continuing.
+        // We don't need to clean up all resources here, since the process will be terminated
+        // in any case; only the ones that would block a new process from starting up.
```
```suggestion
        // in any case; only the resources that would block a new process from starting up.
```

nit, b/c I had trouble parsing this at first.

src/multiversioning.zig
```diff
@@ -1278,9 +1556,11 @@ test "parse_elf" {
...
 }
 
-pub fn validate(allocator: std.mem.Allocator, exe_path: [:0]const u8) !void {
-    if (builtin.target.os.tag != .linux) @panic("only linux is supported for validate()");
-
+pub fn print_information(
+    allocator: std.mem.Allocator,
+    exe_path: [:0]const u8,
+    stdout: anytype,
+) !void {
```
sentientwaffle, line 1577: Can `tigerbeetle version --verbose` be enabled in the [smoke tests](https://github.com/tigerbeetle/tigerbeetle/pull/2134)?

src/multiversioning.zig
```diff
@@ -1294,44 +1574,19 @@ pub fn validate(allocator: std.mem.Allocator, exe_path: [:0]const u8) !void {
...
     assert(multiversion.stage == .ready);
-    multiversion.timeout_start_enabled = false;
-
-    if (multiversion.stage == .err) {
-        try std.fmt.format(stdout, "multiversioning not enabled: {}\n", .{multiversion.stage.err});
 
-        return;
-    }
+    try std.fmt.format(stdout, "multiversioning.exe_path={s}\n", .{exe_path});
+    try std.fmt.format(stdout, "multiversioning.absolute_exe_path={s}\n", .{absolute_exe_path});
```
sentientwaffle: FYI this can also be written as `stdout.print("...", .{...});`.

src/multiversioning.zig
```diff
@@ -1278,9 +1556,11 @@ test "parse_elf" {
     }
 }
 
-pub fn validate(allocator: std.mem.Allocator, exe_path: [:0]const u8) !void {
-    if (builtin.target.os.tag != .linux) @panic("only linux is supported for validate()");
-
+pub fn print_information(
+    allocator: std.mem.Allocator,
+    exe_path: [:0]const u8,
+    stdout: anytype,
```
sentientwaffle: Check out the `AnyWriter` pattern (for example in `inspect.zig`). It's nice both b/c the type signature is more explicit, and avoids generics.

src/multiversioning.zig
```diff
@@ -1294,44 +1574,19 @@ pub fn validate(allocator: std.mem.Allocator, exe_path: [:0]const u8) !void {
...
-    defer allocator.free(absolute_exe_path);
-
-    var multiversion = try Multiversion.init(
-        allocator,
-        &io,
-        absolute_exe_path,
-    );
-    defer multiversion.deinit(allocator);
+    multiversion.open_sync() catch |err| {
+        try std.fmt.format(stdout, "multiversioning not enabled: {}\n", .{multiversion.stage.err});
```
sentientwaffle: Can this be:
```suggestion
        try std.fmt.format(stdout, "multiversioning not enabled: {}\n", .{err});
```

..afaict they should be the same.

src/multiversioning.zig
```diff
@@ -1294,44 +1574,19 @@ pub fn validate(allocator: std.mem.Allocator, exe_path: [:0]const u8) !void {
...
-    if (multiversion.stage == .err) {
-        try std.fmt.format(stdout, "multiversioning not enabled: {}\n", .{multiversion.stage.err});
 
-        return;
-    }
+    try std.fmt.format(stdout, "multiversioning.exe_path={s}\n", .{exe_path});
+    try std.fmt.format(stdout, "multiversioning.absolute_exe_path={s}\n", .{absolute_exe_path});
 
     const header = multiversion.target_header.?;
     try std.fmt.format(
```
sentientwaffle: For the header fields (below, outside of the diff) can you indicate in the label that they are part of the header? e.g.

```
            try std.fmt.format(stdout, "multiversioning.header.{s}={any}\n", .{
```

Likewise for `multiversioning.header.past`.

src/multiversioning.zig
```diff
@@ -985,10 +1096,62 @@ pub const Multiversion = struct {
...
+                // Includes the null byte, that utf8ToUtf16LeWithNull needs.
+                var buffer: [std.fs.MAX_PATH_BYTES]u8 = undefined;
+                var fixed_allocator = std.heap.FixedBufferAllocator.init(&buffer);
+                const allocator = fixed_allocator.allocator();
+
+                const target_path_w = std.unicode.utf8ToUtf16LeWithNull(
+                    allocator,
+                    self.target_path,
+                ) catch
+                    unreachable;
```
sentientwaffle:
```suggestion
                ) catch unreachable;
```

src/multiversioning.zig
```diff
@@ -557,72 +561,140 @@ pub const Multiversion = struct {
         err: anyerror,
     } = .init,
 
-    pub fn init(allocator: std.mem.Allocator, io: *IO, exe_path: [:0]const u8) !Multiversion {
-        assert(builtin.target.os.tag == .linux);
+    pub fn init(
+        allocator: std.mem.Allocator,
+        io: *IO,
+        exe_path: [:0]const u8,
+        exe_path_format: enum { auto, native },
```
sentientwaffle: Maybe s/auto/detect? More descriptive (and the lengths match!).

src/multiversioning.zig
```diff
@@ -789,34 +872,70 @@ pub const Multiversion = struct {
...
+        // `init_from_bytes` validates the header checksum internally.
         const source_buffer_header = source_buffer[offsets.header..][0..@sizeOf(
             MultiversionHeader,
         )];
         const header = try MultiversionHeader.init_from_bytes(source_buffer_header);
+        var header_inactive_platform: ?MultiversionHeader = undefined;
+
+        // MachO's checksum_binary_without_header works slightly differently since there are
+        // actually two headers, once for x86_64 and one for aarch64. It zeros them both.
+        if (offsets.header_inactive_platform) |header_inactive_platform_offset| {
```
sentientwaffle, line 908: Assert `offsets.format == .macho`.

src/multiversioning.zig
```diff
@@ -999,30 +1162,49 @@ pub fn self_exe_path(allocator: std.mem.Allocator) ![:0]const u8 {
...
+        // Technically, "/memfd:tigerbeetle-multiversion-... (deleted)" is a valid path at which you
         // could place your binary - please don't!
         assert(std.fs.cwd().statFile(native_self_exe_path) catch null == null);
 
         // Running from a memfd already; the real path is argv[0].
         const path = try allocator.dupeZ(u8, std.mem.span(os.argv[0]));
         assert(std.fs.path.isAbsolute(path));
 
+        return path;
+    } else if (std.mem.indexOf(u8, native_self_exe_path, multiversion_uuid) != null) {
```
sentientwaffle, line 1189: Maybe assert that `os == windows or os == macos`? For "documentation" if nothing else.

src/config.zig
```diff
@@ -141,7 +141,13 @@ const ConfigProcess = struct {
...
     grid_scrubber_interval_ms_max: usize = std.time.ms_per_s * 10,
     aof_recovery: bool = false,
-    multiversion_binary_size_max: u64 = 64 * 1024 * 1024,
+    multiversion_binary_platform_size_max: u64 = blk: {
+        // {Linux, Windows} get 64MB. macOS gets 128MB since it has universal binaries. Both of
+        // these are doubled in debug.
+        var size_max_mb = if (builtin.target.os.tag == .macos) 128 else 64;
+        if (builtin.mode != .ReleaseSafe) size_max_mb *= 2;
+        break :blk size_max_mb;
+    } * 1024 * 1024,
```
sentientwaffle: Maybe just make the config value a "base" (64MiB), and then account fo Debug/Macos by multipling it in `constants.zig` on the RHS of `const multiversion_binary_platform_size_max =`. Then `const multiversion_binary_size_max = ...` can be computed too, to avoid assigning it to a magic number.

src/scripts/release.zig
```diff
@@ -58,6 +59,10 @@ const multiversion_epoch = "0.15.3";
 /// This commit references https://github.com/tigerbeetle/tigerbeetle/pull/1935.
 const multiversion_epoch_commit = "035c895bf85f5106d94f08cac49719994344880e";
 
+/// As does this tag. Have it as an explicit tag, so it's not just a random commit SHA that's
```
sentientwaffle: I'd just copy the `This commit references https://github.com/tigerbeetle/tigerbeetle/pull/1935.` comment down, to be more robust to future shuffling of these lines.

src/scripts/release.zig
```diff
@@ -382,9 +380,102 @@ fn build_tigerbeetle(shell: *Shell, info: VersionInfo, dist_dir: std.fs.Dir) !vo
...
+                },
+                .{
+                    .cpu_type = 0x00000004, // NS32032 == .tb_mvb for x86_64
+                    .cpu_subtype = 0x00000000,
+                    .path = "multiversion-build/multiversion-x86_64-macos" ++ debug_suffix ++
+                        ".body",
+                },
+                .{
+                    .cpu_type = 0x00000005, // NS32332 == .tb_mvh for x86_64
+                    .cpu_subtype = 0x00000000,
```
sentientwaffle: What is the meaning of these "cpu types"? Does MachO just not care what the values are, so they are arbitrary constants from tigerbeetle's point-of-view?

Could we define them as an `enum(c_int) { aarch64_tb_mvh = 1, aarch64_tb_mvb = 2, ... }`, since they are used in multiple places?

Also, maybe in `parse_macho` we could use a `std.enums.EnumSet` to verify that we find (exactly one of) each section in the header.

sentientwaffle (commented):
Not quite done reviewing, but posting what I have so far.

src/scripts/release.zig
cb22: > What is the meaning of these "cpu types"? Does MachO just not care what the values are, so they are arbitrary constants from tigerbeetle's point-of-view?

Exactly that, they are valid (in the spec) but ancient (macOS has never run on anything other than x86_64 / arm64) platforms. They were chosen so that it wouldn't be a random value, but also wouldn't be something that we could realistically encounter.

src/multiversioning.zig
```diff
@@ -985,10 +1096,62 @@ pub const Multiversion = struct {
...
+            .macos => {
+                std.posix.execveZ(self.target_path, self.args_envp.args, self.args_envp.envp) catch
+                    return error.ExecveZFailed;
 
-        unreachable;
+                unreachable;
+            },
+            .windows => {
+                // Includes the null byte, that utf8ToUtf16LeWithNull needs.
+                var buffer: [std.fs.MAX_PATH_BYTES]u8 = undefined;
```
sentientwaffle, line 1121: From `std/fs.zig`:
```
/// Deprecated: use `max_path_bytes`.
pub const MAX_PATH_BYTES = max_path_bytes;
```

src/scripts/release.zig
cb22: > Also, maybe in parse_macho we could use a std.enums.EnumSet to verify that we find (exactly one of) each section in the header.

I think we might go the opposite way in future: allow non-universal binaries, so not too sure we want to enforce this.

sentientwaffle (approved):

src/multiversioning.zig
cb22: There are a few instances scattered around the repo, I'll do a follow-up with all of them.

PR #2138, opened 2024-07-25
ci: disable flaky Go tests on arm mac

matklad:


matklad:
Would be better to fix it of course, but CI must be green first!

cb22 (approved):

PR #2139, opened 2024-07-25
vsr: don't block commit on repairing ops  from the past checkpoint

matklad:
Before this commit, we waiting until all repairable headers are fixed before trying to commits prepares. This is suboptimal: if we have all the header after the checkpoint, then we can start committing even if some headers before checkpoint are missing. This PR does just that.

That seems like a good idea in general, to maximize availability, and to not waste durability, but this also fixes a VOPR failure:

* Cluster of A, B, C, with C partitioned
* A and B move ahead
* B gets partitioned
* C joins A and state syncs to its checkpoint
* A gets corruption in the log in the previous checkpoint

This gets A into an interesting position --- in theory, it could commit everything, because it has the checkpoing and everything after that. But it gets stuck in repair, trying to fix items from the last checkpoint! Formally speaking, the cluster is allowed to be unavailable here: with C partitioned and A corrupted, we have two faults.  However, currently this doesn't trigger "core missing prepare" case in the VOPR, because the missing stuff is before A's commit_min.

No seed, as that was found on one of my temporary branches with semi-random changes on top.

### DIFF
--- a/src/vsr/replica.zig
+++ b/src/vsr/replica.zig
@@ -6219,11 +6219,13 @@ pub fn ReplicaType(
 
             if (self.op < self.op_repair_max()) return;
 
-            // Request any missing or disconnected headers:
-            if (self.journal.find_latest_headers_break_between(
+            const header_break = self.journal.find_latest_headers_break_between(
                 self.op_repair_min(),
                 self.op,
-            )) |range| {
+            );
+
+            // Request any missing or disconnected headers:
+            if (header_break) |range| {
                 assert(!self.solo());
                 assert(range.op_min >= self.op_repair_min());
                 assert(range.op_max < self.op);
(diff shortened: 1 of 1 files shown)


src/vsr/replica.zig
```diff
@@ -6251,15 +6253,40 @@ pub fn ReplicaType(
...
+            assert(self.valid_hash_chain_between(
+                @min(self.op_checkpoint() + 1, self.op),
+                self.op,
+            ));
 
             if (self.journal.dirty.count > 0) {
                 // Request and repair any dirty or faulty prepares.
-                self.repair_prepares();
-            } else if (self.client_replies.faulty.findFirstSet()) |slot| {
-                // After we have all prepares, repair replies.
```
matklad, line 6262: My understanding is that there's actually no reason why we mush repair prepares before proceeding to repair replies, but I am not totally sure.

matklad:
Ah, actually got a fresh seed which shows this effect

[56b5b05 ](https://github.com/tigerbeetle/tigerbeetle/commit/56b5b0572f05b9f9cb795d3c49b55807884eee16)[#1951](https://github.com/tigerbeetle/tigerbeetle/pull/1951)	matklad	[vopr](https://tigerbeetle.github.io/tigerbeetle/?fuzzer=vopr&commit=56b5b0572f05b9f9cb795d3c49b55807884eee16)	./zig/zig build -Drelease vopr -- 14210728117710881556

src/vsr/replica.zig
```diff
@@ -6251,15 +6253,40 @@ pub fn ReplicaType(
...
+                    // Otherwise, concurrently repair headers before checkpoint, while trying
+                    // to commit everything after the checkpoint.
+                    return;
+                }
             }
-            assert(self.valid_hash_chain_between(self.op_repair_min(), self.op));
+            assert(self.valid_hash_chain_between(
+                @min(self.op_checkpoint() + 1, self.op),
+                self.op,
+            ));
```
sentientwaffle, line 6267: Nice! This will still test the `valid_hash_chain_between`'s assert that

```
            // The op immediately after the checkpoint always connects to the checkpoint.
            if (op_min <= self.op_checkpoint() + 1 and op_max > self.op_checkpoint()) {
                assert(self.superblock.working.vsr_state.checkpoint.header.checksum ==
                    self.journal.header_with_op(self.op_checkpoint() + 1).?.parent);
```

We could lift that assert here to be explicit, but `commit_op` will check it as well, so maybe it would be redundant.

src/vsr/replica.zig
sentientwaffle: Yep, no reason.

sentientwaffle (approved):

src/vsr/replica.zig
matklad: Oh dear, I specifically thought that I'll need to split the original hash-chain verification into two parts, and then forgot it...

But yeah, now that I remeber to think about it, I think relying on commit_op is fine here!

PR #2140, opened 2024-07-25
VSR: Header.Block.peer_type=unknown

sentientwaffle:
This is a serious bug! `MessageBus` sees `block` messages (due to state sync or repair) and `peer_type` says they are always from replica 0 (since `Header.Block.replica == 0` always). So if they are being sent by a non-R0 replica, message bus drops the messages  with `message from unexpected peer`.

Follow-up PR will actually add `peer_type` coverage to the VOPR, but I wanted to get the fix in soonest.

### DIFF
--- a/src/vsr/message_header.zig
+++ b/src/vsr/message_header.zig
@@ -211,8 +211,27 @@ pub const Header = extern struct {
             .block => return .unknown,
             // These messages identify the peer as either a replica or a client:
             .ping_client => |ping| return .{ .client = ping.client },
+
             // All other messages identify the peer as a replica:
-            else => return .{ .replica = self.replica },
+            .ping,
+            .pong,
+            .pong_client,
+            .prepare_ok,
+            .reply,
+            .commit,
+            .start_view_change,
+            .do_view_change,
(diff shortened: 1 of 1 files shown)


src/vsr/message_header.zig
```diff
@@ -208,6 +208,7 @@ pub const Header = extern struct {
                 }
             },
             .prepare => return .unknown,
+            .block => return .unknown,
             // These messages identify the peer as either a replica or a client:
             .ping_client => |ping| return .{ .client = ping.client },
             // All other messages identify the peer as a replica:
```
matklad, line 215: To not have this bug in the future, let's replace else with explicit listing of all messges?

matklad:
>Follow-up PR will actually add peer_type coverage to the VOPR, but I wanted to get the fix in soonest.

Related https://github.com/tigerbeetle/tigerbeetle/pull/1850#issuecomment-2066399042

I think Replica should expose interface that accepts a byte buffer, and converts that to a message, so that all raw parsing logic is in replica

matklad (approved):

PR #2141, opened 2024-07-25
cli: clarify that the order of the addresses is significant

matklad:


cb22 (approved):

jorangreef:
It would be good to motivate why the order is significant, i.e. the order enables a replica or client to know how to reach a given replica.

matklad:
We sort-of already do that in the docs for `--replica`, where we say its an index into the `--addresses`. In general, I'd push the why out of `--help` and keep that concise. 

And, really, we should just fixed that :P

jorangreef:
Agreed that we should design this away in the first place.

But till then, we still need to say _why_ it's significant, to give a rational reason.

We can't only say it's important but not give the explanation. (And here I should also motivate why 😄) and that's because if we don't say _why_, then the reader will not be motivated to follow the recommendation, or will have to spend time digging to understand.

It's a psychological thing that was found in the design of chemical factories (I wish I could find the reference), that if engineers did not motivate _why_, then accidents would happen. In fact, there was a manufacturing firm that got really good at designing these extremely dangerous factories, because they had a strict standard for always motivating the _why_, no matter how small the request.

I actually put this in TigerStyle for the same reason (but I find it personally hard as a recommendation always to follow!):

> Always motivate, always say why. Never forget to say why. Because if you explain the rationale for a decision, it not only increases the hearer's understanding, and makes them more likely to adhere or comply, but it also shares criteria with them with which to evaluate the decision and its importance.

We owe it to our readers, to give clear logic justifying significance.

PR #2143, opened 2024-07-25
vsr: decoupling replication and appending to WAL

chaitanyabhandari:
Currently, while handling a prepare message in [on_prepare](https://github.com/tigerbeetle/tigerbeetle/blob/9a2c665bdbfa81b060c31a4be9cdd1c5f37b9819/src/vsr/replica.zig#L1592), [replication](https://github.com/tigerbeetle/tigerbeetle/blob/9a2c665bdbfa81b060c31a4be9cdd1c5f37b9819/src/vsr/replica.zig#L6949) of the prepare to the next replica in the configuration chain is subject to the same restrictions as [appending](https://github.com/tigerbeetle/tigerbeetle/blob/9a2c665bdbfa81b060c31a4be9cdd1c5f37b9819/src/vsr/replica.zig#L1726) the prepare to the replica's WAL. While these [asserts](https://github.com/tigerbeetle/tigerbeetle/blob/9a2c665bdbfa81b060c31a4be9cdd1c5f37b9819/src/vsr/replica.zig#L1597-L1696) (whether there is space in the WAL, whether the prepare is from a future view, whether the replica is in normal status) are critical invariants for appending the prepare to the replica's WAL, forwarding it to another replica even if the sending replica doesn't satisfy these invariants should be a safe operation.

This change decouples replication and appending to the WAL on backups, such that a replica _always_ forwards a prepare to the next replica, even if it can't process that prepare locally.

### DIFF
--- a/src/vsr/replica.zig
+++ b/src/vsr/replica.zig
@@ -1600,13 +1600,15 @@ pub fn ReplicaType(
                 return;
             }
 
-            if (self.status != .normal) {
-                log.debug("{}: on_prepare: ignoring ({})", .{ self.replica, self.status });
+            if (message.header.view < self.view) {
+                log.debug("{}: on_prepare: ignoring (older view)", .{self.replica});
                 return;
             }
 
-            if (message.header.view < self.view) {
-                log.debug("{}: on_prepare: ignoring (older view)", .{self.replica});
+            self.replicate(message);
+
+            if (self.status != .normal) {
+                log.debug("{}: on_prepare: ignoring ({})", .{ self.replica, self.status });
                 return;
             }
 
@@ -1722,7 +1724,6 @@ pub fn ReplicaType(
             self.op = message.header.op;
             self.journal.set_header_as_dirty(message.header);
 
-            self.replicate(message);
             self.append(message);
         }
 
@@ -6969,10 +6970,8 @@ pub fn ReplicaType(
         /// Replication to standbys works similarly, jumping off the replica just before primary.
         /// TODO Use recent heartbeat data for next replica to leapfrog if faulty (optimization).
         fn replicate(self: *Self, message: *Message.Prepare) void {
-            assert(self.status == .normal);
             assert(message.header.command == .prepare);
-            assert(message.header.view == self.view);
-            assert(message.header.op == self.op);
+            assert(message.header.view >= self.view);
 
             if (message.header.op <= self.commit_max) {
                 log.debug("{}: replicate: not replicating (committed)", .{self.replica});
@@ -7484,7 +7483,11 @@ pub fn ReplicaType(
             }
 
             assert(message.header.cluster == self.cluster);
-            assert(message.header.release.value <= self.release.value);
+
+            // Compare the release in this message to ours only if we authored the message.
+            if (message.header.replica == self.replica) {
+                assert(message.header.release.value <= self.release.value);
+            }
 
             if (message.header.command == .block) {
                 assert(message.header.protocol <= vsr.Version);
@@ -7504,13 +7507,7 @@ pub fn ReplicaType(
                 .prepare => |header| {
                     maybe(self.standby());
                     assert(self.replica != replica);
-                    // Do not assert message.header.replica because we forward .prepare messages.
-                    switch (self.status) {
-                        .normal => assert(message.header.view <= self.view),
-                        .view_change => assert(message.header.view < self.view),
-                        // These are replies to a request_prepare:
-                        else => assert(message.header.view <= self.view),
-                    }
+                    if (header.replica == self.replica) assert(message.header.view <= self.view);
                     assert(header.operation != .reserved);
                 },
                 .prepare_ok => |header| {
@@ -7638,7 +7635,6 @@ pub fn ReplicaType(
                     assert(message.header.replica != replica);
                 },
             }
-
             if (replica != self.replica) {
                 // Critical: Do not advertise a view/log_view before it is durable.
                 // See view_durable()/log_view_durable().
--- a/src/vsr/replica.zig
+++ b/src/vsr/replica.zig
@@ -6956,6 +6956,10 @@ pub fn ReplicaType(
         fn replicate(self: *Self, message: *Message.Prepare) void {
             assert(message.header.command == .prepare);
             assert(message.header.view >= self.view);
+            // We may replicate older prepares from either a primary of the current or future view
+            // whose start view we're waiting on (to truncate our log and set our self.op
+            // accordingly).
+            maybe(message.header.op < self.op);
 
             if (message.header.op <= self.commit_max) {
                 log.debug("{}: replicate: not replicating (committed)", .{self.replica});
@@ -7491,6 +7495,7 @@ pub fn ReplicaType(
                 .prepare => |header| {
                     maybe(self.standby());
                     assert(self.replica != replica);
+                    // Do not assert message.header.replica because we forward .prepare messages.
                     if (header.replica == self.replica) assert(message.header.view <= self.view);
                     assert(header.operation != .reserved);
                 },
--- a/src/vsr/replica.zig
+++ b/src/vsr/replica.zig
@@ -7624,9 +7624,11 @@ pub fn ReplicaType(
                     assert(message.header.replica != replica);
                 },
             }
-            if (replica != self.replica) {
-                // Critical: Do not advertise a view/log_view before it is durable.
-                // See view_durable()/log_view_durable().
+            // Critical:
+            // Do not advertise a view/log_view before it is durable. We only need perform these
+            // checks if we authored the message, not if we're simply forwarding a message along.
+            // See view_durable()/log_view_durable().
+            if (replica != self.replica and message.header.replica == self.replica) {
                 if (message.header.view > self.view_durable() and
                     message.header.command != .request_start_view and
                     !(message.header.command == .ping and
--- a/src/vsr/replica.zig
+++ b/src/vsr/replica.zig
@@ -1600,11 +1600,6 @@ pub fn ReplicaType(
                 return;
             }
 
-            if (message.header.view < self.view) {
-                log.debug("{}: on_prepare: ignoring (older view)", .{self.replica});
-                return;
-            }
-
             self.replicate(message);
 
             if (self.status != .normal) {
@@ -5624,12 +5619,10 @@ pub fn ReplicaType(
         fn is_repair(self: *const Self, message: *const Message.Prepare) bool {
             assert(message.header.command == .prepare);
 
+            if (message.header.view < self.view) return true;
+
             if (self.status == .normal) {
-                if (message.header.view < self.view) return true;
                 if (message.header.view == self.view and message.header.op <= self.op) return true;
-            } else if (self.status == .view_change) {
-                if (message.header.view < self.view) return true;
-                // The view has already started or is newer.
             }
 
             return false;
--- a/src/vsr/replica.zig
+++ b/src/vsr/replica.zig
@@ -1594,7 +1594,10 @@ pub fn ReplicaType(
             assert(message.header.replica < self.replica_count);
             assert(message.header.operation != .reserved);
 
-            if (self.is_repair(message)) {
+            if (message.header.view < self.view or
+                (self.status == .normal and
+                message.header.view == self.view and message.header.op <= self.op))
+            {
                 log.debug("{}: on_prepare: ignoring (repair)", .{self.replica});
                 self.on_repair(message);
                 return;
@@ -5613,18 +5616,6 @@ pub fn ReplicaType(
                 });
                 return true;
             }
-            return false;
-        }
-
-        fn is_repair(self: *const Self, message: *const Message.Prepare) bool {
-            assert(message.header.command == .prepare);
-
-            if (message.header.view < self.view) return true;
-
-            if (self.status == .normal) {
-                if (message.header.view == self.view and message.header.op <= self.op) return true;
-            }
-
             return false;
         }
 


src/vsr/replica.zig
```diff
@@ -1594,6 +1594,8 @@ pub fn ReplicaType(
             assert(message.header.replica < self.replica_count);
             assert(message.header.operation != .reserved);
 
+            self.replicate(message);
```
matklad: This is probably too early: we shouldn't replicate messages which are not the tip of the log. So at least we should move this beyond `is_repair`. And perhaps "older view" case also shouldn't replicate? Though I don't have a clear mental model of that case, so would be good to double-check when that happens.

chaitanyabhandari: Ah, indeed. This should at least be moved beyond `on_repair` given we'd needlessly be replicating the prepares that we _requested_ from another replica for _repair_, which is just plain wrong! Thanks a lot for spotting it and pointing it out. 😄 

Regarding all of the other checks, my understanding was that replication entails simply _forwarding_ the prepare to the next replica in the chain; whether that prepare should and can be processed at a particular backup (and appended to the WAL) is governed (and protected) by all the checks that we have in place in on_prepare. For example, even if we receive a prepare from the future, we can forward it to the next replica in the chain instead of dropping it entirely, because the next replica _may_ be in a view that can accept the prepare (and we are going to drop it locally anyway, thanks to our checks in on_prepare). Thoughts?

chaitanyabhandari:
The VOPR has found a liveness bug (`./zig/zig build -Drelease vopr -- 10549347189305880271`) on 2c2ef0de0f6f9d775b25535354075f19abd080ab, wherein two replicas get stuck in view change. However, based on my analysis, it seems like it's unrelated to this change. Rather, it seems like an liveness failure!

```C
[info] (cluster): no liveness, final cluster state (core=11):
[info] (cluster):  0   \ v      15512V 119/124/143C  98:143Jo  0/_1J!  98:129Wo <__0:__0> v1:3    51Ga  0G!   0G?
[info] (cluster):  1   \  v     15512V 139/139/143C 112:143Jo  1/_1J! 112:143Wo <__0:__0> v3:3    50Ga  0G!   0G?
[info] (cluster):  2   #   #
[info] (cluster):  3   |    .      4V 139/143/143C 112:143Jo  0/_0J! 112:143Wo <__0:__0> v3:3    50Ga  0G!   0G?
[info] (cluster):  4   |     .     4V 139/143/143C 112:143Jo  0/_0J! 112:143Wo <__0:__0> v3:3    50Ga  0G!   0G?
[err] (cluster): you can reproduce this failure with seed=10549347189305880271
[err] (cluster): no state convergence
vopr
+- run vopr failure
```

Here, the final state is that R0 & R1 are stuck in view change, and R2 is down. R2 is stopped for an upgrade but isn't restarted because the upgrade version isn't available:
```C
[debug] (cluster): 2: release_execute_soon: release=0.0.1..0.0.3
[debug] (cluster): 2: release_execute: release=0.0.1..0.0.3
[debug] (cluster): 2: release_execute: target version not available
```


Now, as to why R0 and R1 are stuck in view change, here are 3 DVC quorums for reference (for view=6,7,9):
```C
[debug] (replica): 0: on_do_view_change: view=6 quorum received
[debug] (replica): 0: on_do_view_change: dvc: replica=0 log_view=2 op=129 commit_min=124 checkpoint=119
[debug] (replica): 0: on_do_view_change: dvc: header: replica=0 op=129 checksum=320142782286098618624460214495175585374 nack=false present=true type=valid
[debug] (replica): 0: on_do_view_change: dvc: replica=1 log_view=4 op=143 commit_min=139 checkpoint=139
[debug] (replica): 0: on_do_view_change: dvc: header: replica=1 op=143 checksum=143069130741402480840817816795632497981 nack=false present=true type=valid
[debug] (journal): 0: remove_entries_from: op_min=144
[debug] (replica): 0: primary_set_log_from_do_view_change_messages: view=6 op=129..143 commit=129..143

[debug] (replica): 1: on_do_view_change: view=7 quorum received
debug] (replica): 1: on_do_view_change: dvc: replica=0 log_view=6 op=143 commit_min=124 checkpoint=119
[debug] (replica): 1: on_do_view_change: dvc: header: replica=0 op=143 checksum=143069130741402480840817816795632497981 nack=true present=false type=valid
[debug] (replica): 1: on_do_view_change: dvc: replica=1 log_view=4 op=143 commit_min=139 checkpoint=139
[debug] (replica): 1: on_do_view_change: dvc: header: replica=1 op=143 checksum=143069130741402480840817816795632497981 nack=false present=true type=valid

debug] (replica): 0: on_do_view_change: view=9 quorum received
[debug] (replica): 0: on_do_view_change: dvc: replica=0 log_view=6 op=143 commit_min=124 checkpoint=119
[debug] (replica): 0: on_do_view_change: dvc: header: replica=0 op=143 checksum=143069130741402480840817816795632497981 nack=true present=false type=valid
[debug] (replica): 0: on_do_view_change: dvc: replica=1 log_view=7 op=143 commit_min=139 checkpoint=139
[debug] (replica): 0: on_do_view_change: dvc: header: replica=1 op=143 checksum=143069130741402480840817816795632497981 nack=false present=true type=valid
```

These are the ops that each of the replicas wants to repair in order to become primary:
```C
[debug] (replica): 0: repair: break: view=6 break=130..142 (commit=124..143 op=143)
[debug] (replica): 1: repair: break: view=7 break=142..142 (commit=139..143 op=143)
```
However, we can see from the DVC for view=6 that the head op for R0 was op=129, but it was incremented to op=143 after it received DVC quorum. So, R0 doesn't have ops from 129 → 142. Moreover, slots=14 & 13 in the WAL prepare region, and slots 0 → 16 in the WAL headers region are corrupted on R1:
```C
[debug] (storage): 1: corrupting sector at zone=vsr.Zone.wal_headers offset=0 slots=0...16
[debug] (storage): 1: corrupting sector at zone=vsr.Zone.wal_prepares offset=57344 slot=14
[debug] (storage): 1: corrupting sector at zone=vsr.Zone.wal_prepares offset=53248 slot=13
```

Therefore, R1 can't help R0 recover op=142 due to a storage fault, and R0 can't help R1 recover op=142 because it never had that op at all!
```C
[debug] (journal): 1: copy_latest_headers_between: op_min=130 op_max=142 dest.len=13 copied=12

[debug] (replica): 0: on_request_headers: ignoring (op=142..142, no headers)
```

chaitanyabhandari:
More about the liveness bug that I analyzed in https://github.com/tigerbeetle/tigerbeetle/pull/2143#issuecomment-2251549864. I instrumented [core_missing_prepare](https://github.com/tigerbeetle/tigerbeetle/blob/9a2c665bdbfa81b060c31a4be9cdd1c5f37b9819/src/vopr.zig#L652), which detects the first missing op between commit_min and the head op as `op=130 on replica 0`. However, this op is present on replica 0 so it rightfully isn't regarded as a missing prepare. But I believe the logic in this function (which I haven't given a close look as of now) should be fixed to regard `op=142 on replica 1` as a missing prepare?

```C
[debug] (simulator): 0: head op=143 commit_min=124 commit_max=143
[debug] (simulator): 0: found non-clean op=130
[debug] (simulator): 0: found lowest missing op=130
[debug] (simulator): 0: found non-clean op=131
[debug] (simulator): 0: found non-clean op=132
[debug] (simulator): 0: found non-clean op=133
[debug] (simulator): 0: found non-clean op=134
[debug] (simulator): 0: found non-clean op=135
[debug] (simulator): 0: found non-clean op=136
[debug] (simulator): 0: found non-clean op=137
[debug] (simulator): 0: found non-clean op=138
[debug] (simulator): 0: found non-clean op=139
[debug] (simulator): 0: found non-clean op=140
[debug] (simulator): 0: found non-clean op=141
[debug] (simulator): 0: found non-clean op=142
[debug] (simulator): 0: found non-clean op=143
[debug] (simulator): 1: head op=143 commit_min=139 commit_max=143
[debug] (simulator): 1: found non-clean op=142
````

Plan is to give the logic of [core_missing_prepare](https://github.com/tigerbeetle/tigerbeetle/blob/9a2c665bdbfa81b060c31a4be9cdd1c5f37b9819/src/vopr.zig#L652) a closer look first thing tomorrow.

File: src/vsr/replica.zig [hunk 2]
```diff
@@ -6947,10 +6948,7 @@ pub fn ReplicaType(
         /// Replication to standbys works similarly, jumping off the replica just before primary.
         /// TODO Use recent heartbeat data for next replica to leapfrog if faulty (optimization).
         fn replicate(self: *Self, message: *Message.Prepare) void {
-            assert(self.status == .normal);
             assert(message.header.command == .prepare);
-            assert(message.header.view == self.view);
-            assert(message.header.op == self.op);
```
matklad, line 6975: can we assert `header.op >= self.op`?

src/vsr/replica.zig
```diff
@@ -1600,6 +1600,8 @@ pub fn ReplicaType(
                 return;
             }
 
+            self.replicate(message);
+
             if (self.status != .normal) {
                 log.debug("{}: on_prepare: ignoring ({})", .{ self.replica, self.status });
                 return;
```
matklad, line 1610: Ok, the `(message.header.view < self.view)` condition looks sus to me!

If I am reading `is_repair` correctly ,we are only ending up in that branch when we are recovering head. If we are normal or view change, that we consider this case a repair.

But this means, if we are forwarding this, we either forward it to a recovering head, which will further forward, or we are forwarding to normal or start_view, in which case we are forwarding a repair, which we shouldn't do. 

_And_ in on repair, we already have 

            if (self.status != .normal and self.status != .view_change) {
                log.debug("{}: on_repair: ignoring ({})", .{ self.replica, self.status });
                return;
            }


So perhaps we should just always consider `view < view` case as repair, even in recovering head, and here assert that the view is >=?

If the view is >, that's the case where we _correctly_ should forward a prepare, because we won't be jump viewing immediately, and instead waiting for an SV from the primary

src/vsr/replica.zig
matklad: And I guess `header.view >= self.view`

src/vsr/replica.zig
chaitanyabhandari: Isn't the opposite true for _is_repair_, wherein we are only entering the branch if we are in *normal/view_change*? 

```C
        fn is_repair(self: *const Self, message: *const Message.Prepare) bool {
            assert(message.header.command == .prepare);

            if (self.status == .normal) {
                if (message.header.view < self.view) return true;
                if (message.header.view == self.view and message.header.op <= self.op) return true;
            } else if (self.status == .view_change) {
                if (message.header.view < self.view) return true;
                // The view has already started or is newer.
            }

            return false;
        }
```

Are you talking about a case where we receive a prepare as a *recovering_head* replica? In that case, wouldn't we just be forwarding the prepare without doing anything with it locally (which *seems* harmless to me)? 🤔

matklad: I mean that in `on_repair` we havndle the case where we neither normal, nor view_change already. So in this sense `on_repair` and `is_repair` are self-inconsistent. I suggest treating `on_repair` as a source of truth, and always rout prepares from older views as repairs, which is a simpler mental model.

src/vsr/replica.zig
chaitanyabhandari: Re `header.view >= self.view` I agree that dropping messages with `header.op < self.view` should not be forwarded, since prepares from an older primary can never receive enough prepare_oks to get committed (and also for normal, view_change they are repairs, which shouldn't be forwarded).

chaitanyabhandari: Re `header.op >= self.op`, I believe this would only be valid if we also check if status is normal _before_ replicating a prepare. However, I wast thinking that may it be safe, and rather beneficial to allow forwarding messages with `header.op < self.op` in some cases...

Consider a case in a 3-node cluster R0, R1, & R2, wherein R1 was primary, it prepared a bunch of ops but couldn't commit the last few. Then, R2 becomes primary and R0 has received its start_view and transitioned to normal, but R1 hasn't. Now, R2 sends a prepare to R1 (who is in view_change still), but it doesn't forward this prepare because it is in view_change status. In this case, it _could_ have been better for R1 to forward it along the chain, in the hope that the subsequent replicas are write it to their WAL (in this case R2, which is in normal and can process it) and prepare_ok it? 

Or was there a specific where you think it would be unsafe to do so?

matklad: Good point re op! I guess there's also a case where in the new view we _trucated_ the ops, so even if our op is larger, its actually truncated, so it still non-repair.

src/vsr/replica.zig
```diff
@@ -1600,13 +1600,15 @@ pub fn ReplicaType(
...
             }
 
-            if (self.status != .normal) {
-                log.debug("{}: on_prepare: ignoring ({})", .{ self.replica, self.status });
+            if (message.header.view < self.view) {
+                log.debug("{}: on_prepare: ignoring (older view)", .{self.replica});
                 return;
             }
```
matklad: Hm, I'd still just make `is_repair` to return `true` for this use case, this _should_ work and would lead to simpler code, no?

chaitanyabhandari: Oops, I hadn't _completely_ understood your suggestion earlier then. Indeed it would simplify things, let me do this!

matklad: Yeah, I formulated it in a quiet confusing way, sorry about that! It's much easier to think than to communicate :rofl:

src/vsr/replica.zig
```diff
@@ -5623,12 +5619,10 @@ pub fn ReplicaType(
         fn is_repair(self: *const Self, message: *const Message.Prepare) bool {
             assert(message.header.command == .prepare);
 
+            if (message.header.view < self.view) return true;
+
```
matklad: Would it make sense to further compress this down to a single expression? View is less or view is equal, satisfy is normal, and op is less?

matklad: This=this entire function

chaitanyabhandari: Ah yes, no other usages of this function anyway! Let me compress it down to a single expression and and use it directly in _on_prepare_.

matklad: Oh, removing the function is even better!

chaitanyabhandari: https://github.com/tigerbeetle/tigerbeetle/pull/2143/commits/b4bac8022d9cfdac3bc8807459da8d932e6d3529

src/vsr/replica.zig
```diff
@@ -7504,13 +7491,7 @@ pub fn ReplicaType(
                 .prepare => |header| {
                     maybe(self.standby());
                     assert(self.replica != replica);
-                    // Do not assert message.header.replica because we forward .prepare messages.
```
matklad, line 7507: The comment is still useful!

src/vsr/replica.zig
```diff
@@ -6969,10 +6954,8 @@ pub fn ReplicaType(
         /// Replication to standbys works similarly, jumping off the replica just before primary.
         /// TODO Use recent heartbeat data for next replica to leapfrog if faulty (optimization).
         fn replicate(self: *Self, message: *Message.Prepare) void {
-            assert(self.status == .normal);
             assert(message.header.command == .prepare);
-            assert(message.header.view == self.view);
-            assert(message.header.op == self.op);
+            assert(message.header.view >= self.view);
```
matklad, line 6958: Let's add a `maybe(message.header.op < self.op)`. Hm, actually, maybe we could assert implication here?

```zig
if (message.header.op < self.op) assert(message.header.view < self.view)
```

In general, it's always usetful to think which asserts are missing!

chaitanyabhandari: Ack, makes sense! Adding `maybe(message.header.op < self.op)` since by this point we've already dropped older prepares so we can't `assert(message.header.view < self.view)`. The older ops that we forward may be either from a primary of a newer view the same view whose start_view we're waiting on (https://github.com/tigerbeetle/tigerbeetle/pull/2143#discussion_r1693251200).

chaitanyabhandari: https://github.com/tigerbeetle/tigerbeetle/pull/2143/commits/f6895cea32ffe0e02e969be767bc8f9057d2e1c3

matklad (approved):

matklad (approved):

PR #2147, opened 2024-07-25
ci: skip merge group checks on main

cb22:
Jotting this down as an idea:

Maybe we don't want to do this at all, or maybe we want to do it a different way, but I _think_ GitHub guarantees that the commit that passes the `merge_group` is the commit that will be merged onto main. So, we can skip running all our CI on an identical commit.

(If we split the files, so we have a ci.yml that has the merge_group and pull_request triggers, and then a main.yml that has the devhub and kcov job with a push trigger, it'll get rid of the skipped jobs on both the PR check page and the main commit status page)

(Unfortunately the merge_group and the PR have different commit hashes, even though they have the same contents. So to make those conditional if they're identical would be a bit harder)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,6 +12,7 @@ on:
 
 jobs:
   smoke:
+    if: github.ref != 'refs/heads/main'
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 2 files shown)


matklad:
We intentionally run the checks in main to catch residual flaky tests! But would be good to `# why` it in yaml!

PR #2148, opened 2024-07-26
tidy: add long function logic

SuperAuguste:
Will check if functions are longer than `maximum_allowable_function_line_count`.

If functions are longer than that constant, `tidy` will be very disappointed. :(

`tidy` will also complain if the constant is not exactly identical to our current longest function. This ensures we keep lowering this constant as we make our functions less lengthy.

### DIFF
--- a/src/tidy.zig
+++ b/src/tidy.zig
@@ -25,6 +25,8 @@ test "tidy" {
     var dead_detector = DeadDetector.init(allocator);
     defer dead_detector.deinit();
 
+    var function_line_count_longest: usize = 0;
+
     // NB: all checks are intentionally implemented in a streaming fashion, such that we only need
     // to read the files once.
     while (try walker.next()) |entry| {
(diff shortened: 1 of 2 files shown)


src/tidy.zig
```diff
@@ -69,6 +81,117 @@ fn tidy_long_line(file: SourceFile) !void {
...
+        first_token_location: std.zig.Ast.Location,
+        last_token_location: std.zig.Ast.Location,
+        /// Functions that are not "innermost," meaning that they have other functions
+        /// inside of them (such as functions that return `type`s) are not checked as
+        /// it is normal for them to be very lengthy.
+        is_innermost: bool,
+
+        fn is_parent_of(a: @This(), b: @This()) bool {
+            return a.first_token_location.line_start < b.first_token_location.line_start and
+                a.last_token_location.line_end > b.last_token_location.line_end;
```
matklad, line 110: Neat way to check if functions are nested, I was imagining something way more complex here!

src/tidy.zig
```diff
@@ -69,6 +81,117 @@ fn tidy_long_line(file: SourceFile) !void {
     }
 }
 
+/// `tigerbeetle/cli.zig`'s `parse_args` holds the function line coiunt record as of July 26, 2024
+/// with 366 lines. As we trim our functions, make sure to update this constant; tidy will
+/// error if you do not.
+const maximum_allowable_function_line_count = 366;
```
matklad: I'd say `function_line_count_max` would be a more TigerStyle name.

src/tidy.zig
```diff
@@ -69,6 +81,117 @@ fn tidy_long_line(file: SourceFile) !void {
...
+
+fn tidy_long_functions(
+    file: SourceFile,
+    maximum_allowable_function_line_count_needs_updating: *bool,
```
matklad: We generally try to avoid this pattern of passing out parameters for bools and such, as it makes following the control flow harder, and makes it easier to accidentally step into some aliasing nastiness. 

I'd maybe do something like this:

```
fn longest_function(file: SourceFile) struct { line_count: u32, first_line } {
}
```

and then the caller can do control flow based on `maximum_allowable_function_line_count_needs_updating` (which would get a reduced scope, and could become a shorter name because of that)

src/tidy.zig
matklad: >Add units or qualifiers to variable names, and put the units or qualifiers last, sorted by descending significance, so that the variable starts with the most significant word, and ends with the least significant word. For example, latency_ms_max rather than max_latency_ms. This will then line up nicely when latency_ms_min is added, as well as group all variables that relate to latency.

https://github.com/tigerbeetle/tigerbeetle/blob/main/docs/TIGER_STYLE.md#naming-things

src/tidy.zig
SuperAuguste: Thanks! This technique's straight from ZLS. :P

src/tidy.zig
SuperAuguste: Thanks, this also made the logic way clearer!

matklad: It's useful to reflect _why_ it became clearer. I think the main principle here is "centralize control flow": https://github.com/tigerbeetle/tigerbeetle/commit/47af705b7aebdb6361ae1a15092131fb93bba322

matklad (approved):
Thanks! 

Now we have another metric to game :partying_face:

PR #2150, opened 2024-07-26
Update quick-start.md

PakhomovAlexander:
When I click "copy" console command I get 
three commands (macOs, Linux, Windows) 
in my clipboard. This does not make any sense.

I've separated these commands so the  one
can click "copy" on the command they want
to execute. The UIX is better in this case.


matklad (approved):
:heart:

PR #2153, opened 2024-07-26
VSR: Fix stuck-client bug; Add eviction replica_tests (part 1/2)

sentientwaffle:
## Summary

Fix a bug in which a client could get stuck -- indefinitely unable to make requests, despite having a session.
(Thankfully it isn't an easy bug to hit -- triggering the bug requires more than `clients_max` clients.)

Additionally:

- Add replica tests for client eviction and the aforementioned bug.
    - Add an optional `eviction_callback` hook to `vsr.Client.Options`. When set, don't panic on eviction. (Language clients don't pass this flag -- they still panic, like before.)
- The follow-up PR will add eviction to the VOPR as well. (It involves some refactoring that isn't needed for this PR.)

### Scenario

1. Client `A` sends an `operation=register` to a fresh cluster. (request `A₁`)
2. The cluster prepares and commits `A₁`, and sends the reply to `A`.
3. (The reply doesn't arrive immediately. Client `A` keeps re-sending `A₁`.)
4. Client `A` receives the reply to `A₁`, and issues a second request. (request `A₂`)
5. `clients_max` other clients register, evicting `A`'s session.
6. An old retry (or replay) of `A₁` arrives at the cluster.
7. Request `A₁` is committed (for a second time, as a different op -- and thus a different session number -- evicting one of the other clients).
8. Client `A` is stuck. It keeps retrying its second request (`A₂`), but `A` has the session number from the first time `A₁` was committed.

(To be clear, enrolling so many clients is an operator error. But we should handle it more gracefully nonetheless!)

In this case, (prior to this PR,) the client `A` would be stuck forever (or at least until it is evicted for unrelated reasons, due to more new clients registering).

### Fix

When the cluster observes a client sending requests with an old session number, send an eviction message (`reason=session_too_low`), so that the error can be bubbled up to the operator.

### DIFF
--- a/src/testing/cluster.zig
+++ b/src/testing/cluster.zig
@@ -267,11 +267,13 @@ pub fn ClusterType(comptime StateMachineType: anytype) type {
                 errdefer for (clients[0..i]) |*c| c.deinit(allocator);
                 client.* = try Client.init(
                     allocator,
-                    client_id_permutation.encode(i + client_id_permutation_shift),
-                    options.cluster_id,
-                    options.replica_count,
-                    &client_pools[i],
-                    .{ .network = network },
+                    .{
+                        .id = client_id_permutation.encode(i + client_id_permutation_shift),
+                        .cluster = options.cluster_id,
+                        .replica_count = options.replica_count,
+                        .message_pool = &client_pools[i],
+                        .message_bus_options = .{ .network = network },
+                    },
                 );
                 client.release = options.releases[0].release;
             }
--- a/src/testing/cluster.zig
+++ b/src/testing/cluster.zig
@@ -222,6 +222,7 @@ pub fn ClusterType(comptime StateMachineType: anytype) type {
             var replica_pools = try allocator.alloc(MessagePool, node_count);
             errdefer allocator.free(replica_pools);
 
+            // There may be more clients than `clients_max` (to test session eviction).
             const pipeline_requests_limit =
                 @min(options.client_count, constants.clients_max) -|
                 constants.pipeline_prepare_queue_max;
--- a/src/testing/cluster.zig
+++ b/src/testing/cluster.zig
@@ -689,9 +689,9 @@ pub fn ClusterType(comptime StateMachineType: anytype) type {
             assert(reply_message.header.command == .reply);
             assert(reply_message.header.operation == request_message.header.operation);
 
-            const client_index = for (cluster.clients, 0..) |*c, i| {
-                if (client == c) break i;
-            } else unreachable;
+            const client_index =
+                cluster.client_id_permutation.decode(client.id) - client_id_permutation_shift;
+            assert(&cluster.clients[client_index] == client);
 
             cluster.on_cluster_reply(cluster, client_index, request_message, reply_message);
         }
--- a/src/testing/cluster.zig
+++ b/src/testing/cluster.zig
@@ -116,6 +116,9 @@ pub fn ClusterType(comptime StateMachineType: anytype) type {
 
         clients: []Client,
         client_pools: []MessagePool,
+        /// Updated when the *client* is informed of the eviction.
+        /// (Which may be some time after the client is actually evicted by the cluster.)
+        client_eviction_reasons: []?vsr.Header.Eviction.Reason,
         client_id_permutation: IdPermutation,
 
         state_checker: StateChecker,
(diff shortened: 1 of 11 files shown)


src/vsr/client.zig
```diff
@@ -111,42 +111,44 @@ pub fn Client(comptime StateMachine_: type, comptime MessageBus: type) type {
 
         pub fn init(
             allocator: mem.Allocator,
-            id: u128,
-            cluster: u128,
-            replica_count: u8,
-            message_pool: *MessagePool,
-            message_bus_options: MessageBus.Options,
+            options: struct {
```
matklad, line 120: Long overdue, :+1:

src/vsr/client.zig
```diff
@@ -117,6 +123,10 @@ pub fn Client(comptime StateMachine_: type, comptime MessageBus: type) type {
                 replica_count: u8,
                 message_pool: *MessagePool,
                 message_bus_options: MessageBus.Options,
+                eviction_callback: ?*const fn (
+                    client: *Self,
+                    eviction: *const Message.Eviction,
+                ) void = null,
```
matklad, line 133: Could you write some docs what should be done after callback? I see that basically every method has `assert(!evicted)`, so presumably the client shuold not be used afterwards. Should the caller then call deinit?

src/testing/cluster.zig
```diff
@@ -219,7 +220,8 @@ pub fn ClusterType(comptime StateMachineType: anytype) type {
             errdefer allocator.free(replica_pools);
 
             const pipeline_requests_limit =
-                options.client_count -| constants.pipeline_prepare_queue_max;
+                @min(options.client_count, constants.clients_max) -|
```
matklad, line 227: I don't get this `@min`, shouldn't we just assert that `client_count` is not greater than clients_max?

EDIT: got it: clients_max is how many clients can be registered with a cluster at the same time. But of course there could be arbitrary number of clients! 

Worth a comment in the code though!

src/vsr/client.zig
```diff
@@ -13,7 +13,7 @@ const Message = @import("../message_pool.zig").MessagePool.Message;
 const IOPS = @import("../iops.zig").IOPS;
 const FIFO = @import("../fifo.zig").FIFO;
 
-const log = std.log.scoped(.client);
+const log = stdx.log.scoped(.client);
```
matklad, line 16: Did you mean to use `log.mark` somewhere?

src/testing/cluster.zig
```diff
@@ -713,6 +713,21 @@ pub fn ClusterType(comptime StateMachineType: anytype) type {
...
+            // 1. Client `A` sends an `operation=register` to a fresh cluster. (`A₁`)
+            // 2. Cluster prepares + commits `A₁`, and sends the reply to `A`.
+            // 4. `A` receives the reply to `A₁`, and issues a second request (`A₂`).
+            // 5. `clients_max` other clients register, evicting `A`'s session.
+            // 6. An old retry (or replay) of `A₁` arrives at the cluster.
+            // 7. `A₁` is committed (for a second time, as a different op).
+            //    If `StateChecker` were to check `Client.request_inflight`, it would see that `A₁`
+            //    is not actually in-flight, despite being committed for the "first time" by a
+            //    replica.
+            cluster.state_checker.clients_exhaustive = false;
```
matklad, line 729: Might want to update comment in https://github.com/tigerbeetle/tigerbeetle/blob/1955acc454d2e1ef655bc416cbed30cd8417ffc3/src/testing/cluster/state_checker.zig#L201-L203

src/vsr/replica.zig
```diff
@@ -5123,7 +5123,19 @@ pub fn ReplicaType(
...
+                    // 1. Client `A` sends an `operation=register` to a fresh cluster. (`A₁`)
+                    // 2. Cluster prepares + commits `A₁`, and sends the reply to `A`.
+                    // 4. `A` receives the reply to `A₁`, and issues a second request (`A₂`).
+                    // 5. `clients_max` other clients register, evicting `A`'s session.
+                    // 6. An old retry (or replay) of `A₁` arrives at the cluster.
+                    // 7. `A₁` is committed (for a second time, as a different op, evicting one of
+                    //    the other clients).
+                    // 8. `A` sends a second request (`A₂`), but `A` has the session number from the
+                    //    first time `A₁` was committed.
                     log.err("{}: on_request: ignoring older session (client bug)", .{self.replica});
```
matklad: Might remove `client bug`? And maybe even `log.mark` this?

matklad (commented):
Only nits!

Love git history on this one, a joy to review!

src/vsr/client.zig
sentientwaffle: Nope, this was just to prevent tests from failing due when a `log.err` is hit.

```
pub const log = if (builtin.is_test)
    // Downgrade `err` to `warn` for tests.
    // Zig fails any test that does `log.err`, but we want to test those code paths here.
```

matklad: Ahh, right!

matklad (approved):

PR #2154, opened 2024-07-26
VSR: `Header.Reply.peer_type=unknown`; test `peer_type`

sentientwaffle:
Replies are tagged with the `replica` of their corresponding prepare message, which is not necessarily the same replica that transmits the reply to the client. Replicas can also send each other replies (for repair).

Follow-up for https://github.com/tigerbeetle/tigerbeetle/pull/2140.

### DIFF
--- a/src/vsr/message_header.zig
+++ b/src/vsr/message_header.zig
@@ -209,6 +209,7 @@ pub const Header = extern struct {
             },
             .prepare => return .unknown,
             .block => return .unknown,
+            .reply => return .unknown,
             // These messages identify the peer as either a replica or a client:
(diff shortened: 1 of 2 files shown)


src/vsr/message_header.zig
```diff
@@ -217,7 +218,6 @@ pub const Header = extern struct {
             .pong,
             .pong_client,
             .prepare_ok,
-            .reply,
```
matklad, line 220: Well, even spelling out the else explicitly doesn't prevent all bugs =/

matklad (approved):
:+1:

Issue #1182, opened 2023-09-11
REPL: Add Tab Completion


cb22:
It would be great if TigerBeetle's REPL had tab completion on static tokens, in the same way you'd expect from a shell etc.

Eg:

```                                                                                                                                                                                                                                                                                                                                            ~/g/t/tigerbeetle [130] > ./tigerbeetle client --addresses=3001 --cluster=0                                                                                                                                                                                                                                                                            
TigerBeetle CLI Client 0.13.123-81-g1a952461
  Hit enter after a semicolon to run a command.

Examples:
  create_accounts id=1 code=10 ledger=700,
                  id=2 code=10 ledger=700;
  create_transfers id=1 debit_account_id=1 credit_account_id=2 amount=10 ledger=700 code=10;
  lookup_accounts id=1;
  lookup_accounts id=1, id=2;

> crea<tab>
> create_accounts led<tab>
> create_accounts ledger=
```

siddhantsharma301:
hey! i'd like to give this issue a shot if possible

cb22:
Hey @siddhantsharma301 - I realized some people may not get notifications for reactions - do go ahead! :rocket:

jalotra:
hey @siddhantsharma301 
did you took this up, any reference draft PR raised for this ?

jalotra:
hey @cb22  can I take this up ?

hp77-creator:
Hi @cb22 , I went through the code, currently for checking the input, following is being used: 
`stdin_stream.readUntilDelimiterOrEofAlloc`. To have tab completion we will have to change it to a character by character check. Is this feasible? I am not much familiar with `zig` so I am thinking of iterating over the input and suggesting the related commands when last character matches with a tab character.

raghunandanbhat:
Started working on this. Changes will be done on top of #2155, which handles raw input. Planning to open a PR soon.

PR #2155, opened 2024-07-29
Interactive REPL raw input mode + enhancements

### LINKED ISSUE #1183: REPL: Command history / Arrow Key navigation
Pressing left / right / up / down in the REPL currently just prints out the control codes. 

Making left / right move the cursor, and up / down scroll through history (which can just be in-memory for now, and later stored on disk) would be a nice improvement. This might get a bit complex to do cross-platform between *nix and Windows without using an external library.

wpaulino:
Closes https://github.com/tigerbeetle/tigerbeetle/issues/1183.


kjloveless (commented):
No relation to tigerbeetle but I think this looks like a great start!

I was reviewing some open tasks and saw #1182 
I too figured to handle this properly dropping into raw mode would make the most sense.

The only thing I think would make this cleaner is if the Terminal stuff got moved into their own file. You could reference rockorager/libvaxis and how they init the TTY in main.zig.

Good work!

raghunandanbhat:
@wpaulino great work on this PR. 
I noticed a potential improvement to avoid adding duplicate commands to history if the same command is entered twice in a row. This can keep the history a little bit cleaner. 

here's what I would add to make it happen -

```zig 
if (input.len > 0) {
    if (repl.history.items.len == 0 or !std.mem.eql(u8, repl.history.items[repl.history.items.len - 1], input)) {
        const history_item_next = try repl.history.allocator.alloc(u8, input.len);
        @memcpy(history_item_next, input);
        repl.history.append(history_item_next) catch |err| {
            repl.event_loop_done = true;
            return err;
        };
    }
}

**Comment by cb22** [comment]
Hey @wpaulino - this looks awesome :smile:!

I'll be reviewing it fully in the next few days, but as a head start, you'll want to rename all functions to [snake_case](https://github.com/tigerbeetle/tigerbeetle/blob/main/docs/TIGER_STYLE.md#naming-things).

**Comment by nithinkjoy-tech** [comment]
Hey @wpaulino - wow, amazing work. 🚀

**Comment by brson** [comment]
I'm excited for the improvements here.

I've only just skimmed the code, and intend to review in depth this week.

I have tested this a little on windows and the repl is crashing on startup. I am building under cygwin, and am testing under both the cygwin terminal (mtty i think) and the standard cmd.exe.

Running under cygwin I see this:

```
Examples:
  create_accounts id=1 code=10 ledger=700 flags=linked|history,
                  id=2 code=10 ledger=700;
  lookup_accounts id=1;
  lookup_accounts id=1, id=2;
  get_account_transfers account_id=1 flags=debits|credits;
  get_account_balances account_id=1 flags=debits|credits;

error.Unexpected: GetLastError(87): The parameter is incorrect.

C:\cygwin64\home\Brian\tigerbeetle\src\repl.zig:1207:51: 0x7a370c in promptModeSet (tigerbeetle.exe.obj)
                    return windows.unexpectedError(windows.kernel32.GetLastError());
                                                  ^
C:\cygwin64\home\Brian\tigerbeetle\src\repl.zig:532:35: 0x7a17ce in readUntilNewlineOrEofAlloc (tigerbeetle.exe.obj)
            try repl.promptModeSet();
                                  ^
C:\cygwin64\home\Brian\tigerbeetle\src\repl.zig:699:58: 0x7a4991 in do_repl (tigerbeetle.exe.obj)
            const input = repl.readUntilNewlineOrEofAlloc(
                                                         ^
...
```

Running under cmd.exe I see:

```
Examples:
  create_accounts id=1 code=10 ledger=700 flags=linked|history,
                  id=2 code=10 ledger=700;
  lookup_accounts id=1;
  lookup_accounts id=1, id=2;
  get_account_transfers account_id=1 flags=debits|credits;
  get_account_balances account_id=1 flags=debits|credits;

error.Unexpected: GetLastError(87): The parameter is incorrect.

???:?:?: 0x58370c in ??? (tigerbeetle.exe)
???:?:?: 0x5817ce in ??? (tigerbeetle.exe)
???:?:?: 0x584991 in ??? (tigerbeetle.exe)
```

I would expect that running under cmd.exe would work; and I would not be surprised if some additional check is needed to be compatible with the cygwin terminal environment.

It may be possible that _building_ under cygwin is causing the toolchain to build in a different way than building under the regular windows command line, but I don't know anything about how zig treats cygwin.

Any advice about how you are configuring your windows build environment @wpaulino is welcome.

Under cygwin `uname` prints `CYGWIN_NT-10.0-19045` and I haven't updated cygwin in some time, so that may be relevant.

**Comment by wpaulino** [comment]
Thanks for testing @brson. I somehow wrote down the wrong value for one of the output flags after I had successfully tested it last time. You should be able to run it in Windows now with the latest update.

**Comment by brson** [comment]
@wpaulino this is a great foundation for adding more advanced repl behavior. Very neat.
I have done some more reading of the code and more testing on both Linux and Windows,
though I need to circle back and do more still.
I suspect it will take a few rounds to land since this is some fiddly multi-platform code.

Below I have made various observations about what I'm seeing in the code and the behavior,
and there should be enough here for you to start revising.

---

Like @kjloveless, I think there is too much platform-specific terminal-handling code
in repl.zig, and would like to see it moved to its own file, perhaps terminal.zig,
leaving repl.zig responsible for the logic of managing the repl loop, history, etc.

This includes:

- the platform-specific setup code inside of `run`
- `get_cursor_position`, `get_terminal`, `prompt_mode_set`,
  `prompt_mode_unset`
- probably the `WindowsConsoleMode`, `Terminal`, and `UserInput` types.
- probably `read_until_newline_or_eof_alloc`
- perhaps the `starting_terminal_mode` field

I'm inclined to do that in this PR so we don't have to revisit later.
Can you think about how to split the code up along these lines?

---

Under cmd.exe, backspace behavior is incorrect,
changing the cursor line at the wrong time, not erasing previous characters.

Under cygwin's mintty, backspace is mostly the same as Linux,
except that the last character on the line is not erased when backspacing across
a newline. This only seems to happen when backspace from the end of the input -
if I press the back arrow key and backspace from the middle of the input,
it does not leave a "stray" character sitting in the last column of the terminal.

---

If I resize the terminal the repl's size calculations are wrong,
and e.g. backspacing produces strange results.
If I change the code to call `get_terminal` immediately after reading a byte,
the size calculations behave better:

```
                const user_input = try UserInput.read(reader) orelse return null;
                var terminal = try repl.get_terminal(allocator, reader);
                switch (user_input) {
```

---

If `repl.interactive` is false, the `starting_terminal_mode` field
is never changed from undefined.

Although the field is never touched again in this case, I am still
uncomfortable leaving it undefined and would prefer it to be initialized
to some sentinel value that represents non-interactive mode,
perhaps just a `null`.

Then in every function that uses this field I'd like to see an assertion
that it has been initialized correctly.

It might be best to bury this unsafe detail in terminal.zig
so it's not exposed to the rest of the repl.

---

The parsing behavior across multiple lines has changed:

On the main branch, if I press enter I will continue writing a command across multiple lines,
until I end the line with ";". On this branch the command is immediately parsed
when I press enter, regardless of whether I've entered a terminating ";".

The new code does not require a terminating ";", but will parse a command with or without.

It's not necessarily wrong to do it this way,
but I am interested to hear a rationale for this change.
Is there a precedent either way in other repls?

My intuition is that it is nice to be able to split input across newlines
e.g. if I am entering a large command. On the other hand,
it is also nice to not have to terminate every command with ";".

---

**Comment by wpaulino** [comment]
> If I resize the terminal the repl's size calculations are wrong,
and e.g. backspacing produces strange results.

Yeah this is expected for the time being. I chose not to call `get_terminal` after every read because it's possible we consume some of the user's input as part of reading the cursor position. The best way to handle this would be via a SIGWINCH handler (on POSIX, I haven't searched for a Windows equivalent yet) that redraws the current prompt buffer. I considered this a good candidate for follow-up work in a different PR. For what it's worth, the Python REPL seems to share the same behavior -- if you resize the terminal, the dimensions won't be updated until a Ctrl-C/newline.

> The parsing behavior across multiple lines has changed:
> On the main branch, if I press enter I will continue writing a command across multiple lines,
until I end the line with ";". On this branch the command is immediately parsed
when I press enter, regardless of whether I've entered a terminating ";".

I wasn't aware that was an explicit goal before. I can certainly amend this to retain that behavior. Drawing a comparison to the Python REPL again, multi-line input is supported and often used for defining functions. The only question I have if we decide to proceed is whether we should preserve the full input buffer in the history (i.e., until the terminating `;`), or split it up per line (on every `\n`) as done in the Python REPL.

In the meantime, I'll start addressing the remainder of your comments.

**Comment by brson** [comment]
Not sure if you are looking for more review here yet @wpaulino, but it seemed like I should circle back here in case you are wanting more feedback.

I see that the platform-specific code is now in terminal.zig and that is a nice improvement, makes repl.zig much more readable.

I ran through some basic tests again on linux, cygwin and cmd.exe, and still see a variety of artifacts with characters appearing where they shouldn't: e.g. back-arrowing into the middle of the command and inserting new chars seems to behave strangely; the repl under cmd.exe doesn't work.

On cygwin I no longer see the same artifacts when backspacing across newlines, but there is new incorrect behavior: when i backspace across a newline, then backspace all the way back to the beginning of command I cannot backspace past the first char; there is always one char stuck at the beginning of the command.

> Yeah this is expected for the time being. I chose not to call `get_terminal` after every read because it's possible we consume some of the user's input as part of reading the cursor position. The best way to handle this would be via a SIGWINCH handler (on POSIX, I haven't searched for a Windows equivalent yet) that redraws the current prompt buffer. I considered this a good candidate for follow-up work in a different PR. For what it's worth, the Python REPL seems to share the same behavior -- if you resize the terminal, the dimensions won't be updated until a Ctrl-C/newline.

Acknowledged and agreed. Python also doesn't seem to handle well terminal resizing in the middle of command entry.

**Comment by wpaulino** [comment]
Thanks @brson, I am looking for more feedback/review :)

All the issues you mentioned should be addressed now. I confirmed in all three OSs, including cygwin and cmd.exe on Windows, but would appreciate if you could double check.

> the repl under cmd.exe doesn't work

What exactly doesn't work? I've had no trouble testing it in cmd.exe on a Windows 11 build.

**Comment by brson** [comment]
> Thanks @brson, I am looking for more feedback/review :)
> 
> All the issues you mentioned should be addressed now. I confirmed in all three OSs, including cygwin and cmd.exe on Windows, but would appreciate if you could double check.
> 
> > the repl under cmd.exe doesn't work
> 
> What exactly doesn't work? I've had no trouble testing it in cmd.exe on a Windows 11 build.

Thanks for your patience @wpaulino and sorry for the long delay. I am giving this a fresh look now.

**Comment by brson** [comment]
Ok, I retested this in several environments and they all work consistently now:

- Linux (WSL) in the Windows 11 Terminal,
- Windows Command Prompt in the Windows 11 Termal,
- Windows Powershell in the Windows 11 Terminal,
- conhost.exe (used to be cmd.exe) in Windows 11

I don't see any obvious visual glitches in any of these now. Since my previous review I went from Windows 10 to Windows 11. I'm feeling confident about everything working as expected now.

I am still reading the code closely and see a few things we might tweak, but so far it looks great. I need to take a little more time still reading the code. Thanks for your patience @wpaulino.

**Comment by raghunandanbhat** [comment]
hey @wpaulino, I see some strange behavior on macos terminal- iTerm when backspacing. 
- go to middle of the line using left-arrow.
- backspace to delete mid-buffer, then I see cursor at the end of the current line.
- if I type any `printable` character, it prints everything from where I pressed backspace till the end of buffer.

Screenshot 2024-10-18 at 3 44 55 PM

in the above screenshot, I changed `id` to 16, immediately cursor moves to end. Typing any `printable` character prints `code=10 ledger=7000;` and same will get printed multiple times as I type.

File: src/terminal.zig [hunk 1]
```diff
@@ -0,0 +1,399 @@
...
+
+    pub fn read_user_input(self: *Terminal) !?UserInput {
+        assert(self.mode_start != null);
+        const stdin = self.stdin.reader().any();
+
+        switch (try stdin.readByte()) {
+            std.ascii.control_code.eot => return null,
+            std.ascii.control_code.etx => return .ctrlc,
+            std.ascii.control_code.cr, std.ascii.control_code.lf => return .newline,
+            std.ascii.control_code.bs, std.ascii.control_code.del => return .backspace,
```
brson, line 86: This code indicates `control_code.del` causes a delete, but the delete _key_ on my keyboard does not seem to produce this control code (it causes a `~` to be printed in the repl). In what scenario is this `del` code produced?

brson (changes requested):
OK @wpaulino , I have read through everything in detail and done some more testing.

I basically think this is mostly good as-is and we should get it landed so we can continue forward on more repl improvements.

I have one change request: the `terminal.print` calls that use control codes are quite impenetrable and everyone who looks closely at this code will need to relearn what they do. We don't need to completely document these control codes, but I would like to have some comment before each of them indicating what they do. Some of them already have a comment, but not all.

Please also rebase to fix the merge conflict.

I see one strange behavior still: if i enter a command that is longer than the entire screen, then backspace all the way back to the beginning of the screen, the backscrolling behavior is doesn't look quite right: the lines that previously went offscreen are not restored to the top line, and instead the cursor is backspacing through an empty line where characters should be. I think this is not important enough to hold up the PR though, but maybe something to look at in the future.

---

I do have a concern about how this code handles resources, but again I think it probably doesn't need to hold up this PR, since the new code is following patterns that already existed in the repl:

The repl uses dynamic allocation (unlike the server) but never frees its allocations (nor calls deinit methods...) - it relies on arenas to free memory in batches. After understanding how arenas are used here I don't _think_ there are any leaks or related problems in this code, but ISTM the potential for future mistakes is high, particularly since there are two separate arenas used in this code with two different scopes. (cc @matklad I wonder what you think about this re dynamic memory in the repl)

brson:
I pushed a merge commit to resolve the conflict here.

brson:
After some internal discussion, I understand we need to not introduce any new dynamic allocation to the repl (we can allocate during init, but not during execution - that's tigerstyle), so I am attempting to remove the dynamic allocation from this PR. I know the repl already contained some dynamic allocation, and that is something we also need to look at later.

My work on that is https://github.com/brson/tigerbeetle/tree/interactive-repl-improvements-static. I think I can submit it today, and it will require another round of review from another tigerbeetle team member.

brson:
This has finally been merged. Thanks for your patience @wpaulino.

kjloveless:
@wpaulino excellent work! keep crushing it!

PR #2160, opened 2024-07-29
VSR: Info-log sync start/end

sentientwaffle:
Operators should know when state sync is running.


src/vsr/replica.zig
```diff
@@ -8968,6 +8972,8 @@ pub fn ReplicaType(
             };
             const sync_view = stage.target.view;
 
+            log.info("{}: sync: ops={}..{}", .{ self.replica, sync_op_min, sync_op_max });
```
matklad, line 8983: Could you check that:

* we log the same on `open`
* it also uses `sync:` prefix?

matklad (approved):

PR #2162, opened 2024-07-30
multiversioning: validate inner past checksums on print_information()

cb22:
It's theoretically possible for the release script to generate an invalid checksum for a past release, embed that as `past.checksums[...]` and then compute a correct outer checksum.

When executing, the binary will see a correct outer checksum, continue, try to extract the release and verify its checksum and then panic.

Make it so that `print_information()` verifies these internal checksums too.

### DIFF
--- a/src/multiversioning.zig
+++ b/src/multiversioning.zig
@@ -565,6 +565,7 @@ pub const Multiversion = struct {
     target_fd: posix.fd_t,
     target_path: [:0]const u8,
     target_body_offset: ?u32 = null,
+    target_body_size: ?u32 = null,
     target_header: ?MultiversionHeader = null,
     /// This list is referenced by `Replica.releases_bundled`.
     releases_bundled: ReleaseList = .{},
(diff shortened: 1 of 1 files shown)


src/multiversioning.zig
```diff
@@ -1603,6 +1619,13 @@ pub fn print_information(
     try output.print("multiversioning.absolute_exe_path={s}\n", .{absolute_exe_path});
 
     const header = multiversion.target_header.?;
+
+    // `source_buffer` contains the same data as `target_file` - this code doesn't update anything
+    // after the initial open_sync().
+    try header.past.verify_checksums(
+        multiversion.source_buffer[multiversion.target_body_offset.?..],
```
sentientwaffle: In the elf/etc parser functions should we also return body size to use? (I think we already parse it out.)

src/multiversioning.zig
```diff
@@ -1226,9 +1228,10 @@ pub fn self_exe_path(allocator: std.mem.Allocator) ![:0]const u8 {
 }
 
 const HeaderBodyOffsets = struct {
-    header: u32,
-    header_inactive_platform: ?u32 = null,
-    body: u32,
+    header_offset: u32,
+    header_offset_inactive_platform: ?u32 = null,
+    body_offset: u32,
+    body_size: u32,
```
sentientwaffle, line 1234: Much clearer!

sentientwaffle (approved):

PR #2163, opened 2024-07-30
tidy: don't check client_readmes.zig for function length

matklad:
We have some giant functions in that file, but they are essentially
templated markdown documents, so they are expected to be large.

I've tried extracting each individual section to a separate function,
and, while it is not too ugly and gets past the tidy, I wouldn't
consider that to be an actual improvement!

So let's opt out here!

### DIFF
--- a/src/tidy.zig
+++ b/src/tidy.zig
@@ -84,9 +84,8 @@ fn tidy_long_line(file: SourceFile) !void {
     }
 }
 
-/// `scripts/client_readmes.zig`'s `readme_root` holds the function line count record.
 /// As we trim our functions, make sure to update this constant; tidy will error if you do not.
-const function_line_count_max = 364;
+const function_line_count_max = 361; // build_tigerbeetle
 
 fn tidy_long_functions(
     file: SourceFile,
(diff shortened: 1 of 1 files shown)


matklad:
assigning to a random online person!

src/tidy.zig
```diff
@@ -84,9 +84,8 @@ fn tidy_long_line(file: SourceFile) !void {
     }
 }
 
-/// `scripts/client_readmes.zig`'s `readme_root` holds the function line count record.
 /// As we trim our functions, make sure to update this constant; tidy will error if you do not.
-const function_line_count_max = 364;
+const function_line_count_max = 361; // build_tigerbeetle
```
matklad, line 88: I'll definitely pick also the next one, build_tigerbeetle --- will force me to learn about multiversioning stuff!

matklad: And, if someone wants to prefetch, https://github.com/tigerbeetle/tigerbeetle/blob/bf81db05635a0da0664a099c275958f2b51d2b90/src/state_machine.zig#L3000 is the one after that

sentientwaffle: re:`build_tigerbeetle` you might conflict w/ @cb22, who is working on non-epoch multiversioning.

sentientwaffle (approved):

PR #2164, opened 2024-07-30
cli: don't allocate when parsing CLI arguments 

matklad:
No _strong_ reason to do this, but the fact that we have to deal with
allocating and defalcating twelve medium-sized integers irked me
incessantly for many years!

### DIFF
--- a/src/clients/c/tb_client/context.zig
+++ b/src/clients/c/tb_client/context.zig
@@ -104,7 +104,7 @@ pub fn ContextType(
         allocator: std.mem.Allocator,
         client_id: u128,
 
-        addresses: []const std.net.Address,
+        addresses: stdx.BoundedArray(std.net.Address, constants.replicas_max),
         io: IO,
         message_pool: MessagePool,
         client: Client,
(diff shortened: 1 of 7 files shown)


src/clients/c/tb_client/context.zig
```diff
@@ -134,15 +134,22 @@ pub fn ContextType(
...
+                context.addresses.unused_capacity_slice(),
             ) catch |err| return switch (err) {
                 error.AddressLimitExceeded => error.AddressLimitExceeded,
-                else => error.AddressInvalid,
+                error.AddressHasMoreThanOneColon,
+                error.AddressHasTrailingComma,
+                error.AddressInvalid,
+                error.PortInvalid,
+                error.PortOverflow,
+                => error.AddressInvalid,
```
matklad, line 148: We were eating up an allocation failure here before, so I figured I'd list them all explicitly!

matklad:
Conflicts resolved, should be good to go!

kprotty (approved):

PR #2165, opened 2024-07-31
multiversioning/release: bring up build for non-epoch versions

cb22:
The previous code could only handle one case: building the 0.15.4 release. This extends it to be more general going forward - it downloads and bundles past releases from the previous release.

### DIFF
--- a/src/constants.zig
+++ b/src/constants.zig
@@ -105,20 +105,24 @@ pub const vsr_releases_max = config.cluster.vsr_releases_max;
 
 /// The maximum cumulative size of a final TigerBeetle output binary - including potential past
 /// releases and metadata.
-pub const multiversion_binary_platform_size_max = blk: {
+pub fn multiversion_binary_platform_size_max(options: struct { macos: bool, debug: bool }) u64 {
     // {Linux, Windows} get the base value. macOS gets 2x since it has universal binaries. All cases
     // get a further 2x in debug.
     var size_max = config.process.multiversion_binary_platform_size_max;
-    if (builtin.target.os.tag == .macos) size_max *= 2;
-    if (builtin.mode != .ReleaseSafe) size_max *= 2;
-    break :blk size_max;
-};
+    if (options.macos) size_max *= 2;
+    if (options.debug) size_max *= 2;
+
+    return size_max;
+}
 
 /// The maximum size, like above, but for any platform.
 pub const multiversion_binary_size_max =
     config.process.multiversion_binary_platform_size_max * 2 * 2;
 comptime {
-    assert(multiversion_binary_platform_size_max <= multiversion_binary_size_max);
+    assert(multiversion_binary_platform_size_max(.{
+        .macos = true,
+        .debug = true,
+    }) <= multiversion_binary_size_max);
 }
 
 pub const multiversion_poll_interval_ms = config.process.multiversion_poll_interval_ms;
--- a/src/multiversioning.zig
+++ b/src/multiversioning.zig
@@ -13,6 +13,13 @@ const elf = std.elf;
 // Re-export to make release code easier.
 pub const checksum = @import("vsr/checksum.zig");
 pub const multiversion_binary_size_max = constants.multiversion_binary_size_max;
+pub const multiversion_binary_platform_size_max = constants.multiversion_binary_platform_size_max;
+
+// Useful for test code, or constructing releases in release.zig.
+pub const ListU32 = stdx.BoundedArray(u32, constants.vsr_releases_max);
+pub const ListU128 = stdx.BoundedArray(u128, constants.vsr_releases_max);
+pub const ListGitCommit = stdx.BoundedArray([20]u8, constants.vsr_releases_max);
+pub const ListFlag = stdx.BoundedArray(MultiversionHeader.Flags, constants.vsr_releases_max);
 
 /// In order to embed multiversion headers and bodies inside a universal binary, we repurpose some
 /// old CPU Type IDs.
@@ -169,7 +176,7 @@ test "ReleaseTriple.parse" {
 }
 
 pub const MultiversionHeader = extern struct {
-    const Flags = packed struct {
+    pub const Flags = packed struct {
         /// Normally release upgrades are allowed to skip to the latest. If a corresponding release
         /// is set to true here, it must be visited on the way to the newest release.
         visit: bool,
@@ -365,11 +372,14 @@ pub const MultiversionHeader = extern struct {
     past: PastReleases = .{},
     past_padding: [16]u8 = std.mem.zeroes([16]u8),
 
+    current_git_commit: [20]u8,
+    current_release_client_min: u32,
+
     /// Reserved space for future use. This is special: unlike the rest of the *_padding fields,
     /// which are required to be zeroed, this is not. This allows adding whole new fields in a
     /// backwards compatible way, while preventing the temptation of changing the meaning of
     /// existing fields without bumping the schema version entirely.
-    reserved: [4768]u8 = std.mem.zeroes([4768]u8),
+    reserved: [4744]u8 = std.mem.zeroes([4744]u8),
 
     /// Parses an instance from a slice of bytes and validates its checksum. Returns a copy.
     pub fn init_from_bytes(bytes: *const [@sizeOf(MultiversionHeader)]u8) !MultiversionHeader {
@@ -379,7 +389,7 @@ pub const MultiversionHeader = extern struct {
         return self;
     }
 
-    fn verify(self: *const MultiversionHeader) !void {
+    pub fn verify(self: *const MultiversionHeader) !void {
         const checksum_calculated = self.calculate_header_checksum();
 
         if (checksum_calculated != self.checksum_header) return error.ChecksumMismatch;
@@ -389,6 +399,15 @@ pub const MultiversionHeader = extern struct {
         if (!stdx.zeroed(&self.current_flags_padding)) return error.InvalidCurrentFlags;
         if (!self.current_flags.visit) return error.InvalidCurrentFlags;
         if (self.current_release == 0) return error.InvalidCurrentRelease;
+
+        // current_git_commit and current_release_client_min were added after 0.15.4.
+        if (self.current_release > (try Release.parse("0.15.4")).value) {
+            if (stdx.zeroed(&self.current_git_commit)) return error.InvalidCurrentRelease;
+            if (self.current_release_client_min == 0) return error.InvalidCurrentRelease;
+        } else {
+            if (!stdx.zeroed(&self.current_git_commit)) return error.InvalidCurrentRelease;
+            if (self.current_release_client_min != 0) return error.InvalidCurrentRelease;
+        }
 
         stdx.maybe(stdx.zeroed(&self.reserved));
 
@@ -460,10 +479,6 @@ pub const MultiversionHeader = extern struct {
 };
 
 test "MultiversionHeader.advertisable" {
-    const ListU32 = stdx.BoundedArray(u32, constants.vsr_releases_max);
-    const ListU128 = stdx.BoundedArray(u128, constants.vsr_releases_max);
-    const ListGitCommit = stdx.BoundedArray([20]u8, constants.vsr_releases_max);
-
     const tests = [_]struct {
         releases: []const u32,
         flags: []const MultiversionHeader.Flags,
@@ -523,6 +538,8 @@ test "MultiversionHeader.advertisable" {
             .current_checksum = 0,
             .current_flags = .{ .visit = true, .debug = false },
             .checksum_binary_without_header = 1,
+            .current_git_commit = std.mem.zeroes([20]u8),
+            .current_release_client_min = 0,
         };
         header.checksum_header = header.calculate_header_checksum();
 
@@ -600,7 +617,10 @@ pub const Multiversion = struct {
 
         const multiversion_binary_size_max_by_format = switch (exe_path_format) {
             .detect => constants.multiversion_binary_size_max,
-            .native => constants.multiversion_binary_platform_size_max,
+            .native => constants.multiversion_binary_platform_size_max(.{
+                .macos = builtin.target.os.tag == .macos,
+                .debug = builtin.mode != .ReleaseSafe,
+            }),
         };
 
         // To keep the invariant that whatever has been advertised can be executed, while allowing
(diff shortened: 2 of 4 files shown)


src/constants.zig
```diff
@@ -105,20 +105,24 @@ pub const vsr_releases_max = config.cluster.vsr_releases_max;
 
 /// The maximum cumulative size of a final TigerBeetle output binary - including potential past
 /// releases and metadata.
-pub const multiversion_binary_platform_size_max = blk: {
+pub fn multiversion_binary_platform_size_max(options: struct { macos: bool, debug: bool }) u64 {
```
cb22, line 108: Open to other ways of doing this; not sure if we want a `fn` in constants.zig.

src/multiversioning.zig
```diff
@@ -365,11 +372,14 @@ pub const MultiversionHeader = extern struct {
     past: PastReleases = .{},
     past_padding: [16]u8 = std.mem.zeroes([16]u8),
 
+    current_git_commit: [20]u8,
```
cb22, line 375: I'm going to claim I left these out on purpose, to test adding new fields :smile:.

src/multiversioning.zig
```diff
@@ -390,6 +400,15 @@ pub const MultiversionHeader = extern struct {
         if (!self.current_flags.visit) return error.InvalidCurrentFlags;
         if (self.current_release == 0) return error.InvalidCurrentRelease;
 
+        // current_git_commit and current_release_client_min were added after 0.15.4.
+        if (self.current_release > (try Release.parse("0.15.4")).value) {
```
cb22, line 404: If we keep adding things like this, this'll be a whole bunch of if statements. Maybe that's ok?

src/scripts/release.zig
```diff
@@ -530,106 +534,286 @@ fn build_tigerbeetle(shell: *Shell, info: VersionInfo, dist_dir: std.fs.Dir) !vo
...
+        old_current_release_output_name,
+        multiversion_binary_size_max,
+    ));
+
+    const old_current_release_size: u32 = @intCast((try (try shell.cwd.openFile(
+        old_current_release_output_name,
+        .{ .mode = .read_only },
+    )).stat()).size);
+
+    // FIXME: This can vary by platform; so we could create a binary for Linux that has 10 past
```
cb22: An unsolved problem here. We could:
* make it assert the counts are identical, and fix it down the line,
* limit to the lessor of all platforms,
* align the limits so that the same number of binaries can fit,
* ???

The debug builds are very large - 20MB a pop on Linux. With the doubled max of 128MB, that's 6 releases. For production, the 64MB max could handle 23 releases at ~2.7MB each.

matklad:
>error: 'tidy.test.tidy' failed: scripts/release.zig: banned: FIXME comments must be addressed before getting to main

Ah, I see that we actually do want to discuss what to do here before merge, sorry, was to fast to look at the CI before diff!

src/constants.zig
matklad: Hm, I need to look how this whole thing is setup, but what gives me a pause is the larger issue: `release.zig` is build orchestration, it _shouldn't_ (transitively) depend on anything in the main binary. 

So, what's strange is not that we add something to constants, but that release.zig pulls in multiversioning.zig which pulls in IO and constants.

This is not necessary bad for any particular reason, but still makes me uneasy.

src/scripts/release.zig
matklad: Hm, this entire logic seems backwards (modulo my current _very_ foggy understanding of the logic, might be talking extreme amount of nonsense here)

It looks like what we do here is that we have a certain size budget, and try to fit as many versions as possible. 

I'd expect this to go the other way around: just _declare_ that "I want exactly X versions", pack the X, and then asser that it fits the size budget?

src/constants.zig
cb22: > So, what's strange is not that we add something to constants, but that release.zig pulls in multiversioning.zig which pulls in IO and constants.

Right, so the release logic uses the `Header` struct, some misc `Release` structs and the `parse_*`. Those mostly have no dependencies, except for `constants.vsr_releases_max`.

Could pull them out into a standalone file, and include that both from `release.zig` and `multiversioning.zig`.

src/scripts/release.zig
cb22: > I'd expect this to go the other way around: just declare that "I want exactly X versions", pack the X, and then asser that it fits the size budget?

We have two limits: a logical number on the past releases `past_releases_max` (which is `constants.vsr_releases_max - 1`, ie 64) and the physical binary size we're willing to put out (because we have to allocate for it, it's not free to increase. These numbers were mostly plucked from "how big of a binary do we feel comfortable shipping". 64MB on Windows / Linux, x2 for Mac, further x2 for debug).

This leads to an implicit number of releases, but that was by design because the size of a releases is not fixed (and we might introduce compression). In general, we want to pack as many as possible in.

That said, this is the release logic, so this is all easy to change. We could define targets, assuming a certain size per build, and assert that those are under the maximum size. If we ever hit that assert in a build, we deal with it then by lowering the target, reducing binary size, etc.

matklad: >This leads to an implicit number of releases, but that was by design because the size of a releases is not fixed (and we might introduce compression). In general, we want to pack as many as possible in

Hm, this does look surprising. If I am a user, I probably have some sort of automation running to pick new releases automatically, and that automation probably would want to depend on specific number of releases available. 

I am worrying about the scenarios where we always packed 8 releases, but then one day we blew up the code somewhere, packed 7 by accident, and somoene eles rollout blew up.

cb22: > Hm, this does look surprising. If I am a user, I probably have some sort of automation running to pick new releases automatically, and that automation probably would want to depend on specific number of releases available.

Hmm - I'm not sure that's going to be common :thinking:. We could perhaps drop releases for other reasons, and I think we would expect users to have fixed versions installed and not have new releases automatically installed.

The release notes would described the minimum supported versions for both the replica and client_release_min.

src/scripts/release.zig
```diff
@@ -530,106 +534,286 @@ fn build_tigerbeetle(shell: *Shell, info: VersionInfo, dist_dir: std.fs.Dir) !vo
...
+    const tigerbeetle_name_zip = if (macos)
+        "tigerbeetle-universal-macos" ++ debug_suffix
+    else
+        tigerbeetle_name_binary;
 
-    const past_binary = try shell.project_root
-        .openFile("./tigerbeetle-epoch/dist/extracted/" ++ exe_name, .{ .mode = .read_only });
+    // macOS gets called twice, once for aarch64 and once for x86_64, but both of these map to
+    // tigerbeetle-universal-macos{-debug}.zip. There's no need to download and extract again.
+    if (!(macos and shell.cwd.statFile(tigerbeetle_name_zip) catch null != null)) {
```
matklad: Could use `shell.file_exists` here

src/scripts/release.zig
```diff
@@ -530,106 +534,286 @@ fn build_tigerbeetle(shell: *Shell, info: VersionInfo, dist_dir: std.fs.Dir) !vo
...
+        try shell.exec("gh release download -p {tigerbeetle_name_zip}.zip", .{
+            .tigerbeetle_name_zip = tigerbeetle_name_zip,
+        });
+        try shell.exec("unzip {tigerbeetle_name_zip}.zip", .{
+            .tigerbeetle_name_zip = tigerbeetle_name_zip,
+        });
+        try shell.exec("mv {exe_name} {tigerbeetle_name_zip}", .{
```
matklad: Maybe `shell.cwd.rename`, to avoid gratuitous dependency on mv?

src/scripts/release.zig
```diff
@@ -530,106 +534,286 @@ fn build_tigerbeetle(shell: *Shell, info: VersionInfo, dist_dir: std.fs.Dir) !vo
...
+        "tigerbeetle-universal-macos" ++ debug_suffix
+    else
+        tigerbeetle_name_binary;
 
-    const past_binary = try shell.project_root
-        .openFile("./tigerbeetle-epoch/dist/extracted/" ++ exe_name, .{ .mode = .read_only });
+    // macOS gets called twice, once for aarch64 and once for x86_64, but both of these map to
+    // tigerbeetle-universal-macos{-debug}.zip. There's no need to download and extract again.
+    if (!(macos and shell.cwd.statFile(tigerbeetle_name_zip) catch null != null)) {
+        try shell.exec("gh release download -p {tigerbeetle_name_zip}.zip", .{
```
matklad, line 576: Would would go wring if `gh release download` gets us the wrong release? I guess we could accidentally skip over a `visit = true` version that way? 

I wonder if there's some neat way to assert that there aren't any in-between versions?

src/scripts/release.zig
```diff
@@ -343,8 +319,19 @@ fn build_tigerbeetle(shell: *Shell, info: VersionInfo, dist_dir: std.fs.Dir) !vo
...
+
+                // It's expected that the first iteration of the outer loop is debug, x86_64-linux.
+                try shell.exec_options(
+                    .{ .echo = false },
+                    "multiversion-build/past/" ++
+                        "tigerbeetle-x86_64-linux-debug-{release} multiversion {exe_name}",
+                    .{ .release = release, .exe_name = exe_name },
+                );
+            }
+
```
matklad, line 329: This'll bump build_tigerbeetle over function line length limit!

```
scripts/release.zig:158 error: above function line count max with 377 lines
```

I  think for this PR we should just add thisas an excteption to tidy. 

However, this should be refactored once we actually do do the release. As it stands, the code's indeed quite impenetrable :D 

I am even wondering if prepping the multi-release should be lifted up to the build system, such that release.zig does less building and focuses more on packaging&uploading. 

Probably still more convenient to do this whole thing imperatively via shell.zig

src/scripts/release.zig
matklad: >The release notes would described the minimum supported versions for both the replica and client_release_min.

Hm, one complication here would be that we won't know these versions until after the release is done? We write notes first ,and then make a release, and only during release we'd figure out how much of the past releases we actually were able to fit?

src/scripts/release.zig
```diff
@@ -530,106 +534,286 @@ fn build_tigerbeetle(shell: *Shell, info: VersionInfo, dist_dir: std.fs.Dir) !vo
...
+    const old_current_release_size: u32 = @intCast((try (try shell.cwd.openFile(
+        old_current_release_output_name,
+        .{ .mode = .read_only },
+    )).stat()).size);
+
+    // FIXME: This can vary by platform; so we could create a binary for Linux that has 10 past
+    // versions and one for Windows that only has 8. What do we want to do here?
+    const past_count = blk: {
+        const max_size = multiversion_binary_platform_size_max(.{
```
matklad: Hm, this seems wrong, shouldn't we also count the _current_ size towards the limit?

src/scripts/release.zig
matklad: :thinking: yeah, I am not sure what we want to do here. I think I am still more worried about the count being different between two subsequent linux releases, then between a linux and a windows release. 

My instinct here is just say that we always pack just `4` releases, and then adjust as we go. 

In particular, it seems to me that we won't actually want to pack consequtive releases, if we have `visit` infra? I'd expect us to do some kind of binary climbing, where you have

* the latest release
* the release before that
* the previous release that's divisible by 4
* the previous release that's divisible by 8
* the previous release that's divisible by 16

such that you can jump over large slices of time...

cb22: > 🤔 yeah, I am not sure what we want to do here. I think I am still more worried about the count being different between two subsequent linux releases, then between a linux and a windows release.

Discussed this today, one idea was to flip it around:
* Binaries get a fixed size cap, eg 4MB for release, 20MB for debug
* The number we pack in gets a fixed multiple, eg 10 binaries
* That determines our memory limit: 40MB release, 200MB debug

Another idea was to remove some of the TOCTOU proofing, and get rid of the double buffering and exec'ing the latest binary from memfd rather than disk, so we'd only need to statically allocate the memory for 1 release rather than the whole binary size.

> My instinct here is just say that we always pack just 4 releases, and then adjust as we go.

I think for this version, let's do exactly that.

> In particular, it seems to me that we won't actually want to pack consequtive releases, if we have visit infra? I'd expect us to do some kind of binary climbing, where you have

That's clever, but you'd have to couple that with an intermediate to first get you to that previous release divisible by x number (the binary on disk must contain the running version, even if it's not marked as visit).

cb22: > Hm, one complication here would be that we won't know these versions until after the release is done? We write notes first ,and then make a release, and only during release we'd figure out how much of the past releases we actually were able to fit?

Not quite, I was thinking [here](https://github.com/tigerbeetle/tigerbeetle/releases/tag/0.15.4) which is generated after the final binaries have been generated, and could be autopopulated.

src/scripts/release.zig
cb22: Yes - good catch!

matklad:
CI failure seems legit?

> src\scripts\release.zig:725:14: error: type 'u0' cannot represent integer value '511'

src/scripts/release.zig
```diff
@@ -530,106 +534,283 @@ fn build_tigerbeetle(shell: *Shell, info: VersionInfo, dist_dir: std.fs.Dir) !vo
...
+    var offsets = multiversioning.ListU32{};
+    var sizes = multiversioning.ListU32{};
+    var flags_ = multiversioning.ListFlag{};
+    var git_commits = multiversioning.ListGitCommit{};
+    var release_client_mins = multiversioning.ListU32{};
+
+    // Extract the old current release - this is the release that was the current release, and not
+    // embedded in the past pack.
+    const old_current_release = header.current_release;
+    const old_current_release_output_name = try std.fmt.allocPrint(
```
matklad: There's `shell.fmt` for this

cb22:
> CI failure seems legit?
> 
> > src\scripts\release.zig:725:14: error: type 'u0' cannot represent integer value '511'

Ah yes. It's the mode of createFile failing on Windows builds (even though it's not called!). Will sort it out shortly.

matklad (approved):
LGTM!

In particular, I've spend today just runing upgrade from 0.15.4 to the present version, and it works (canary: I did find one bug, but that was a bug in my test!)

I am not quiet happy with the code in release.zig, but I want to refactor that anyway and to just fold it into normal `zig build`, so that one wouldn't have to touch release infra to create a multiversion binary!

PR #2167, opened 2024-07-31
Switch AOF to being a runtime option

cb22:
Depends on #1989 

Removes the AOF code from the build system, and makes it a runtime option instead. Importing an AOF is also now handled with the `import_` operations instead of a custom build time flag.

Additionally, only log certain state machine operations. These are controlled by the `aof_record` EnumArray inside the state machine. This is useful to avoid logging of read only operations.

(still WIP: doesn't yet do the imports, and hash chaining is broken with only recording certain operations.)


cb22:
Closing this in favor of doing it piece by piece.

PR #2168, opened 2024-08-01
multiversion: fix UAF/double free

matklad:
The `path` here is "owned" by args_iterator:

* on windows, it really is owned, so, after return it points to garbage
* on nix, it actually points to static readonly memory, so we can't free it at the end


cb22 (approved):
Good catch!

PR #2171, opened 2024-08-01
Import historical events

batiati:
## Import historical events

Allows the ingestion of historical accounts and transfers with their original `timestamp` into TigerBeetle.

The new `flag.imported` allows the creation of transfers and accounts using a user-defined timestamp representing _when_ the event happened outside TigerBeetle.

Refer to the docs for more details about the validations introduced, in a nutshell:

- Timestamps must be monotonically increasing. Even when importing events with a past date, the user-defined timestamp must be greater than the last timestamp committed for the same kind of object.

That is, it's possible for the user to write a migration script like this:

```
for (account in old_database): 
    client.create_accounts([account])
for (transfer in old_database):
    client.create_transfers([transfer])
```

- It's possible for an `Account` and a `Transfer` to have the same timestamp. However, it's not allowed for a transfer to refer to an account with a more advanced timestamp, such as `transfer.timestamp < debit_account.timestamp` or `transfer.timestamp < credit_account.timestamp`.

- The `imported` flag is orthogonal to all other flags, except that it does not allow `timeout > 0` for pending transfers.

**Best reviewed per commit.**

Included:
- Unit tests and client integration tests.
- Docs and client README.

Releated #1893 and #1968
Replaces #1989


src/tigerbeetle.zig
```diff
@@ -146,8 +148,15 @@ pub const CreateAccountResult = enum(u32) {
     ok = 0,
     linked_event_failed = 1,
     linked_event_chain_open = 2,
+
+    imported_event_expected = 22,
+    imported_event_not_expected = 23,
```
batiati, line 153: Named them `imported_event_*` to align with `linked_event_*`.

docs/reference/account.md
```diff
@@ -215,6 +215,41 @@ When set, the account will retain the history of balances at each transfer.
 Note that the [`get_account_balances`](./requests/get_account_balances.md) operation only works for
 accounts with this flag set.
 
+#### `flags.imported`
+
+When set, allows importing historical `Account`s with their original [`timestamp`](#timestamp).
+
+TigerBeetle will not use the [real-time clock](../coding/time.md) to assign the timestamp, allowing
+the user to define it, expressing _when_ the account was effectively created by an external
+event.
```
sentientwaffle:
```suggestion
TigerBeetle will not use the [real-time clock](../coding/time.md) to assign the timestamp, instead allowing
the user to define it, expressing _when_ the account was effectively created by an external
event.
```

Also maybe call it the "cluster's clock" or "wall clock" instead of "real-time clock"? Isn't "real-time clock" mostly used in the context of hardware?

docs/reference/account.md
```diff
@@ -215,6 +215,41 @@ When set, the account will retain the history of balances at each transfer.
...
+- It is not allowed to mix events with the `imported` flag set and _not_ set in the same batch.
+  The application must submit batches of imported events separately.
+
+- User-defined timestamps must be **unique** and expressed as nanoseconds since the UNIX epoch.
+  No two accounts can share the same timestamp.
+
+- User-defined timestamps must be a past date, never ahead of the cluster real-time clock at the
+  time the request arrives.
+
+- Timestamps must be monotonically increasing.
```
sentientwaffle:
```suggestion
- Timestamps must be strictly increasing.
```

"monotonically increasing" is true, but allows for repeated values.

docs/reference/account.md
```diff
@@ -215,6 +215,41 @@ When set, the account will retain the history of balances at each transfer.
...
+
+  Since the timestamp cannot regress, importing past events can be naturally restrictive without
+  coordination, as the last timestamp can be updated using the real-time clock during regular
+  cluster activity. Instead, it's recommended to import events only on a fresh cluster or
+  during a scheduled maintenance window.
+
+  Additionally, the entire batch can be submitted as a [linked chain](#flagslinked), ensuring that
+  if any event fails, none of them are committed, preserving the last timestamp unchanged.
+  This approach gives the application a chance to correct failed imported events, re-submitting
+  the batch again with the same user-defined timestamps.
```
sentientwaffle: Maybe make this a bit stronger? e.g. instead of "can be" we could say that it is recommended.

docs/reference/account.md
```diff
@@ -226,7 +261,9 @@ You can read more about [Time in TigerBeetle](../coding/time.md).
 Constraints:
 
 - Type is 64-bit unsigned integer (8 bytes)
-- Must be set to `0` by the user when the `Account` is created
+- Must be `0` when the `Account` is created
```
sentientwaffle:
```suggestion
- Must be `0` when the `Account` is created with
  [`flags.imported`](#flagsimported) unset
```

docs/reference/requests/create_accounts.md
```diff
@@ -29,10 +29,36 @@ The account was not created. The [`Account.flags.linked`](../account.md#flagslin
 the last event in the batch, which is not legal. (`flags.linked` indicates that the chain continues
 to the next operation).
 
+### `imported_event_expected`
+
+The account was not created. The [`Account.flags.imported`](../account.md#flagsimported) was
+expected to be set, as it's not allowed to mix accounts with different `imported` flag in the
+same batch. The first account determines the entire operation.
+
```
sentientwaffle: Maybe clearer:

> The account was not created. The [`Account.flags.imported`](../account.md#flagsimported) was set on the first object of the batch, but not all accounts in the batch. Batches cannot mix import accounts with non-import accounts.

(ditto for `imported_event_not_expected`)

docs/reference/requests/create_accounts.md
```diff
@@ -29,10 +29,36 @@ The account was not created. The [`Account.flags.linked`](../account.md#flagslin
...
+The account was not created. The [`Account.flags.imported`](../account.md#flagsimported) was
+expected to _not_ be set, as it's not allowed to mix accounts with different `imported` flag
+in the same batch. The first account determines the entire operation.
+
 ### `timestamp_must_be_zero`
 
-The account was not created. The [`Account.timestamp`](../account.md#timestamp) is nonzero, but must
-be zero. The cluster is responsible for setting this field.
+The account was not created. The [`Account.timestamp`](../account.md#timestamp) is nonzero, but
+must be zero. The cluster is responsible for setting this field.
```
sentientwaffle, line 49:
```suggestion
must be zero. `flags.imported` is not set, so the cluster is responsible for setting this field.
```

Worth mentioning imports since a dev may run into this when they are intending to import but forget to set the flag.

docs/reference/transfer.md
```diff
@@ -253,6 +263,7 @@ Constraints:
 
 - Type is 32-bit unsigned integer (4 bytes)
 - Must be zero if `flags.pending` is _not_ set
+- Must be zero if `flags.imported` is set.
```
sentientwaffle, line 268: Up above where we say "Non-pending transfers cannot have a timeout." we could also add that "Imported transfers cannot have a timeout." as emphasis.

docs/reference/transfer.md
```diff
@@ -389,6 +400,41 @@ pending transfer will never exceed/overflow either account's limits.
...
+  Since the timestamp cannot regress, importing past events can be naturally restrictive without
+  coordination, as the last timestamp can be updated using the real-time clock during regular
+  cluster activity. Instead, it's recommended to import events only on a fresh cluster or
+  during a scheduled maintenance window.
+
+  Additionally, the entire batch can be submitted as a [linked chain](#flagslinked), ensuring that
+  if any event fails, none of them are committed, preserving the last timestamp unchanged.
+  This approach gives the application a chance to correct failed imported events, re-submitting
+  the batch again with the same user-defined timestamps.
+
```
sentientwaffle, line 439: Add a bullet mentioning the incompatibility with transfer timeouts.

src/clients/dotnet/README.md
```diff
@@ -531,3 +531,50 @@ batch.Add(new Transfer { Id = 4, /* ... rest of transfer ... */ });
...
 // error handling omitted
 ```
+
+## Imported Events
+
+When the `imported` flag is specified for an account when creating accounts or
+a transfer when creating transfers, it allows importing historical events with
+a user-defined timestamp.
+
+The entire batch of events must be set with the flag `imported`.
```
sentientwaffle, line 539: Can we recommend `linked` (and use it in the examples)? Iiuc an import script that doesn't use linked chains for all batches is probably not correct.

src/repl.zig
sentientwaffle, line 1: Ah, and since `imported` is just another flag, the repl automatically supports imports!

src/state_machine.zig
```diff
@@ -1713,6 +1735,18 @@ pub fn StateMachineType(
                 return create_account_exists(a, e);
             }
 
+            if (a.flags.imported) {
+                // Allows past timestamp, but validates whether it regressed from the last
+                // inserted event.
```
```suggestion
                // inserted account.
```

(To clarify, since we are only comparing the account groove range, not the transfers groove).

src/tigerbeetle.zig
```diff
@@ -257,8 +282,26 @@ pub const CreateTransferResult = enum(u32) {
...
+    //
+    // Workaround:
+    // https://github.com/ziglang/zig/blob/66b71273a2555da23f6d706c22e3d85f43fe602b/lib/std/enums.zig#L1278-L1280
+    imported_event_expected = 56,
+    imported_event_not_expected = 57,
+    imported_event_timestamp_must_not_be_zero = 58,
+    imported_event_timestamp_must_not_advance = 59,
+    imported_event_debit_account_must_not_advance = 60,
+    imported_event_credit_account_must_not_advance = 61,
+    imported_event_timestamp_must_not_regress = 62,
```
sentientwaffle: In the state machine, `imported_event_timestamp_must_not_regress` is higher precedence than `imported_event_{debit,credit}_account_must_not_advance`.

src/tigerbeetle.zig
```diff
@@ -257,8 +282,26 @@ pub const CreateTransferResult = enum(u32) {
...
+    // note: use @setEvalBranchQuota() to raise the branch limit from 1000.
+    //
+    // Workaround:
+    // https://github.com/ziglang/zig/blob/66b71273a2555da23f6d706c22e3d85f43fe602b/lib/std/enums.zig#L1278-L1280
+    imported_event_expected = 56,
+    imported_event_not_expected = 57,
+    imported_event_timestamp_must_not_be_zero = 58,
+    imported_event_timestamp_must_not_advance = 59,
+    imported_event_debit_account_must_not_advance = 60,
+    imported_event_credit_account_must_not_advance = 61,
```
sentientwaffle: :thinking: I'm not sure I understand what "account must not advance" means. I understand what the code is checking, I just don't see how it corresponds to that label.

src/state_machine.zig
```diff
@@ -1788,6 +1822,30 @@ pub fn StateMachineType(
...
+                if (t.timestamp < cr_account.timestamp) {
+                    return .imported_event_credit_account_must_not_advance;
+                }
+                if (t.timeout != 0) {
+                    assert(t.flags.pending);
+                    return .imported_event_timeout_must_be_zero;
+                }
+            }
+            assert(t.timestamp >= dr_account.timestamp);
+            assert(t.timestamp >= cr_account.timestamp);
```
sentientwaffle: Should this instead be:

```suggestion
                if (t.timestamp <= dr_account.timestamp) {
                    return .imported_event_debit_account_must_not_advance;
                }
                if (t.timestamp <= cr_account.timestamp) {
                    return .imported_event_credit_account_must_not_advance;
                }
                if (t.timeout != 0) {
                    assert(t.flags.pending);
                    return .imported_event_timeout_must_be_zero;
                }
            }
            assert(t.timestamp > dr_account.timestamp);
            assert(t.timestamp > cr_account.timestamp);
```

? In the absence of imports, there is a guarantee that timestamps are not shared between accounts and transfers. Why discard that invariant?

docs/reference/requests/create_transfers.md
sentientwaffle, line 1: Also update `flags_are_mutually_exclusive`.

src/state_machine.zig
```diff
@@ -1890,14 +1948,16 @@ pub fn StateMachineType(
...
+                assert(t2.flags.pending);
+                assert(!t2.flags.imported);
+                const expires_at = t2.timestamp + t2.timeout_ns();
                 if (expires_at < self.expire_pending_transfers.pulse_next_timestamp) {
                     self.expire_pending_transfers.pulse_next_timestamp = expires_at;
                 }
             }
 
-            self.commit_timestamp = t.timestamp;
+            self.commit_timestamp = t2.timestamp;
```
sentientwaffle, line 2005: Do these change from `t` to `t2` mean something? Don't they match except for the `amount`? :confused:

src/state_machine.zig
```diff
@@ -2000,6 +2060,18 @@ pub fn StateMachineType(
 
             if (self.get_transfer(t.id)) |e| return post_or_void_pending_transfer_exists(t, e, p);
 
+            if (t.flags.imported) {
+                // Allows past timestamp, but validates whether it regressed from the last
+                // inserted event.
```
```suggestion
                // inserted transfer.
```

src/state_machine.zig
```diff
@@ -2088,7 +2162,7 @@ pub fn StateMachineType(
                 .cr_account = &cr_account_new,
             });
 
-            self.commit_timestamp = t.timestamp;
+            self.commit_timestamp = t2.timestamp;
```
sentientwaffle, line 2215: Why did these `t`'s change to `t2`'s?

src/state_machine.zig
```diff
@@ -2949,54 +3028,6 @@ const TestGetAccountTransfers = TestAccountFilter;
...
-                .post_pending_transfer = t.flags_post_pending_transfer != null,
-                .void_pending_transfer = t.flags_void_pending_transfer != null,
-                .balancing_debit = t.flags_balancing_debit != null,
-                .balancing_credit = t.flags_balancing_credit != null,
-                .padding = t.flags_padding,
-            },
-            .timestamp = timestamp orelse t.timestamp,
-        };
-    }
-};
```
sentientwaffle, line 2998: Nice improvement! Just checking the id is much simpler.

src/state_machine.zig
batiati: Yes, this is just a minor refactor. Currently only `amount` differs, but since we are inserting `t2`, I think it makes sense to place all validations there, avoiding possible misbehavior if something changes other fields in the future.

src/state_machine.zig
batiati: Same https://github.com/tigerbeetle/tigerbeetle/pull/2171#discussion_r1704722150

src/state_machine.zig
batiati: > ? In the absence of imports, there is a guarantee that timestamps are not shared between accounts and transfers. Why discard that invariant?

Sure! Let's keep it for non-imported transfers. 

However, it's an inconvenient special case: Why shouldn't we allow an imported Transfer whose timestamp is equal to one of the imported Accounts? I'm not saying we should allow it either!!!

sentientwaffle: > Why shouldn't we allow an imported Transfer whose timestamp is equal to one of the imported Accounts?

Imo if we disallow it now, we can trivially allow it later if it turns out to be valuable.
But going the other direction would be much harder.

jorangreef: Indeed, this is important: “to minimize surface area” (here meaning preserving stricter invariants)

jorangreef: Sometimes with these things also, we might not yet know why we _should_ be strict, but later we may be grateful that we were.... when it eventually becomes clear. i.e. Erring on the side of being strict closes potential semantic gaps down the line, and also “preserves optionality” (since we have more flexibility, we can decide to relax strict invariants, but one can't go the other way once relaxed).

src/state_machine.zig
sentientwaffle: Makes sense; thanks!

src/clients/go/tb_client_test.go
```diff
@@ -44,7 +46,7 @@ func WithClient(t testing.TB, withClient func(Client)) {
 	replicaCountArg := fmt.Sprintf("--replica-count=%d", TIGERBEETLE_REPLICA_COUNT)
 	clusterArg := fmt.Sprintf("--cluster=%d", TIGERBEETLE_CLUSTER_ID)
 
-	fileName := fmt.Sprintf("./%d_%d.tigerbeetle", TIGERBEETLE_CLUSTER_ID, TIGERBEETLE_REPLICA_ID)
+	fileName := fmt.Sprintf("./%d_%d_%d.tigerbeetle", TIGERBEETLE_CLUSTER_ID, TIGERBEETLE_REPLICA_ID, rand.Int())
 	_ = os.Remove(fileName)
```
sentientwaffle: Should we move
```
	t.Cleanup(func() {
		_ = os.Remove(fileName)
	})
```
up to here in the code? Otherwise it looks like it is possible for the `format` step to fail and the file never gets cleaned up.

Since the file name is randomized it is extra important that we tidy up properly :smile:

docs/reference/requests/create_transfers.md
```diff
@@ -214,20 +214,20 @@ refer to an existing `Account`.
...
 [`Transfer.timestamp`](../transfer.md#timestamp).
 
-### `imported_event_credit_account_must_not_advance`
+### `imported_event_timestamp_must_postdate_credit_account`
 
 This result only applies when [Transfer.flags.imported](../transfer.md#flagsimported) is set.
 
 The transfer was not created. [`Transfer.credit_account_id`](../transfer.md#credit_account_id) must
-refer to an `Account` which [`timestamp`](../account.md#timestamp) is less than or equal to the
+refer to an `Account` which [`timestamp`](../account.md#timestamp) is less than the
```
sentientwaffle: s/which/whose or "with a (timestamp) that is less than ..."

docs/reference/requests/create_transfers.md
```diff
@@ -172,6 +214,22 @@ refer to an existing `Account`.
...
+[`Transfer.timestamp`](../transfer.md#timestamp).
+
+### `imported_event_timestamp_must_postdate_credit_account`
+
+This result only applies when [Transfer.flags.imported](../transfer.md#flagsimported) is set.
+
+The transfer was not created. [`Transfer.credit_account_id`](../transfer.md#credit_account_id) must
+refer to an `Account` which [`timestamp`](../account.md#timestamp) is less than the
+[`Transfer.timestamp`](../transfer.md#timestamp).
```
sentientwaffle: Move these down below `imported_event_timestamp_must_not_regress` to match precedence.

src/state_machine.zig
```diff
@@ -3954,89 +3991,213 @@ test "create_transfers: balancing_debit/balancing_credit + pending" {
...
+
+test "imported events: timestamp" {
+    try check(
+        \\ tick 10 nanoseconds
+        \\
+        \\ account A1  0  0  0  0  _  _  _ _ L1 C1   _    _  _  _ IMP _  0 imported_event_timestamp_must_not_be_zero
+        \\ account A1  0  0  0  0  _  _  _ _ L1 C1   _    _  _  _ IMP _ 99 imported_event_timestamp_must_not_advance
+        \\ account A1  0  0  0  0  _  _  _ _ L1 C1   _    _  _  _ IMP _  2 ok
+        \\ account A2  0  0  0  0  _  _  _ _ L1 C1   _    _  _  _ IMP _  1 imported_event_timestamp_must_not_regress
+        \\ account A2  0  0  0  0  _  _  _ _ L1 C1   _    _  _  _ IMP _  3 ok
```
sentientwaffle, line 4076: Create one last account `A3` (which the transfers don't need to touch) with a high timestamp, to demonstrate that `create_transfers` isn't consulting the account groove key range.

src/clients/node/samples/walkthrough/main.js
```diff
@@ -446,6 +446,54 @@ async function main() {
...
+
+        // External source of time.
+        let historicalTimestamp = 0n
+        const historicalAccounts = [];
+        const historicalTransfers = [];
+
+        // section:imported-events
+        // First, load and import all accounts with their timestamps from the historical source.
+        const accountsBatch = [];
+        for (let index = 0; i < historicalAccounts.length ; i++) {
```
sentientwaffle:
```suggestion
        for (let index = 0; i < historicalAccounts.length; i++) {
```

src/clients/node/samples/walkthrough/main.js
```diff
@@ -446,6 +446,54 @@ async function main() {
...
+          let account = historicalAccounts[i];
+          // Set a unique and strictly increasing timestamp.
+          account.timestamp = historicalTimestamp + index;
+          // Set the account as `imported`.
+          account.flags = AccountFlags.imported;
+          // To ensure atomicity, the entire batch (except the last event in the chain)
+          // must be `linked`.
+          if (index < historicalAccounts.length - 1) {
+            account.flags = AccountFlags.linked;
```
sentientwaffle:
```suggestion
            account.flags |= AccountFlags.linked;
```

src/clients/node/samples/walkthrough/main.js
```diff
@@ -446,6 +446,54 @@ async function main() {
...
+        for (let index = 0; i < historicalTransfers.length ; i++) {
+          let transfer = historicalTransfers[i];
+          // Set a unique and strictly increasing timestamp.
+          transfer.timestamp = historicalTimestamp + index;
+          // Set the account as `imported`.
+          transfer.flags = TransferFlags.imported;
+          // To ensure atomicity, the entire batch (except the last event in the chain)
+          // must be `linked`.
+          if (index < historicalTransfers.length - 1) {
+            transfer.flags = TransferFlags.linked;
```
sentientwaffle:
```suggestion
            transfer.flags |= TransferFlags.linked;
```

src/clients/node/samples/walkthrough/main.js
```diff
@@ -446,6 +446,54 @@ async function main() {
...
+          }
+
+          accountsBatch.push(account);
+        }
+        accountErrors = await client.createAccounts(accountsBatch);
+
+        // Error handling omitted.
+        // Then, load and import all transfers with their timestamps from the historical source.
+        const transfersBatch = [];
+        for (let index = 0; i < historicalTransfers.length ; i++) {
```
sentientwaffle:
```suggestion
        for (let index = 0; i < historicalTransfers.length; i++) {
```

src/clients/go/samples/walkthrough/main.go
```diff
@@ -298,6 +298,50 @@ func main() {
...
+
+		accountsBatch = append(accountsBatch, account)
+	}
+	accountsRes, err = client.CreateAccounts(accountsBatch)
+
+	// Then, load and import all transfers with their timestamps from the historical source.
+	transfersBatch := []Transfer{}
+	for index, transfer := range historicalTransfers {
+		// Set a unique and strictly increasing timestamp.
+		transfer.Timestamp = historicalTimestamp + uint64(index)
```
sentientwaffle: Note that these transfers' timestamps would collide with the ones created by the accounts. (`historicalTimestamp + 0,1,...`).

sentientwaffle:
> > Why shouldn't we allow an imported Transfer whose timestamp is equal to one of the imported Accounts?
 
> Imo if we disallow it now, we can trivially allow it later if it turns out to be valuable.
But going the other direction would be much harder.

This isn't as simple as I thought! Still doable though.

On create_transfers, we need to prefetch an account with that transfer's import timestamp, to verify that it doesn't exist.
(We'll need to add a Groove.prefetch_enqueue_timestamp function.)

batiati:
> This isn't as simple as I thought! Still doable though.

Prefetching by timestamp is fine (we already have the machinery in place). The tricky part is how we can efficiently index the `object_cache` by both `.id` and `.timestamp`, also, it would require more memory for an extra account per transfer.

We can optimize for memory,  instead of needing a prefetch cache of 3x Accounts per Transfer, we could first try to prefetch the Account by timestamp, and then abort prefetching the `{dr,cr}_account` by id if we found a conflicting one (creating accounts isn't a problem because we can repurpose the Transfer cache).
Although it would make the prefetch more complex and involve more steps, we pay the price only when `imported` is enabled.

Do you think the idea is worth exploring?

sentientwaffle:
It is just ~1MiB of extra memory, right? Or does it get multiplied by the batch multiple somehow?

Also based on
```
                    .prefetch_entries_for_read_max = batch_accounts_limit,
                    .prefetch_entries_for_update_max = @max(
                        batch_accounts_limit, // create_account()
                        2 * batch_transfers_limit, // create_transfer(), debit and credit accounts
                    ),
```

it seems like we might not actually need to increase the prefetch size, since `prefetch_entries_for_read_max` is "unused" by `create_transfers`?

I'm not sure how to handle indexing the cachemap, though. :thinking:

batiati:
> it seems like we might not actually need to increase the prefetch size, since prefetch_entries_for_read_max is "unused" by create_transfers?

You're right, we already have enough space!

> I'm not sure how to handle indexing the cachemap, though. 🤔

Quick (and dirty) PoC here: https://github.com/tigerbeetle/tigerbeetle/commit/7bd446a63521790769303d0dbad81ddee150de25

Please, tell me what do you think (unit tests are failing with `timestamp_must_not_regress` as expected).
I'm using an extra `HashMap(u64, *Value)` keyed by timestamp with a pointer from the cache/stash (but the pointers may not be stable, even trying to keep them in sync during update).
Maybe we can implement a two-hop lookup timestamp->id->Value. Or maybe we don't even need a real lookup by timestamp, but just a Set to check whether the timestamp exists or not :thinking:.

sentientwaffle:
Ahh, I was wrong about the 1MiB memory -- that number actually does gets multiplied by the batch multiple:

```
                .map_value_count_max = constants.lsm_batch_multiple *
                    (options.tree_options_object.batch_value_count_limit +
                    options.prefetch_entries_for_read_max),
```

As you say though, we are already allocating enough space... except that we need to allocate a new hash map anyway :facepalm: 

A simple set of "timestamp-exists" would be smaller, at least... Maybe something like `prefetch_enqueue_exists(timestamp)`? If it used a separate `prefetch_keys_exists` hashmap on the groove, we could reuse that in-place as the "exists" set.

batiati:
> A simple set of "timestamp-exists" would be smaller, at least... Maybe something like `prefetch_enqueue_exists(timestamp)`? If it used a separate `prefetch_keys_exists` hashmap on the groove, we could reuse that in-place as the "exists" set.

Implemented a `TimestampSet` that "survives" only during a single batch, and can assert that `exists` is only called for timestamps that were prefetched by `prefetch_exists_enqueue`.

The benchmark now can run `imported` workloads using the cli arg `--flag_imported`.

!! Still need more tests to explore all paths.

docs/reference/requests/create_accounts.md
```diff
@@ -29,10 +29,41 @@ The account was not created. The [`Account.flags.linked`](../account.md#flagslin
...
+### `imported_event_timestamp_must_not_be_zero`
+
+This result only applies when [Account.flags.imported](../account.md#flagsimported) is set.
+
+The account was not created. The [`Account.timestamp`](../account.md#timestamp) is zero, but must
+be a user-defined timestamp.
+
+### `imported_event_timestamp_must_not_advance`
+
+This result only applies when [Account.flags.imported](../account.md#flagsimported) is set.
```
sentientwaffle:
```suggestion
This result only applies when [`Account.flags.imported`](../account.md#flagsimported) is set.

The account was not created. The [`Account.timestamp`](../account.md#timestamp) is zero, but must
be a user-defined timestamp.

### `imported_event_timestamp_must_not_advance`

This result only applies when [`Account.flags.imported`](../account.md#flagsimported) is set.
```

docs/reference/requests/create_transfers.md
```diff
@@ -31,11 +31,42 @@ The transfer was not created. The [`Transfer.flags.linked`](../transfer.md#flags
...
+### `imported_event_timestamp_must_not_be_zero`
+
+This result only applies when [Transfer.flags.imported](../transfer.md#flagsimported) is set.
+
+The transfer was not created. The [`Transfer.timestamp`](../transfer.md#timestamp) is zero, but must
+be a user-defined timestamp.
+
+### `imported_event_timestamp_must_not_advance`
+
+This result only applies when [Transfer.flags.imported](../transfer.md#flagsimported) is set.
```
```suggestion
The [`Transfer.timestamp`](../transfer.md#timestamp) can only be assigned when creating transfers
with [`Transfer.flags.imported`](../transfer.md#flagsimported) set.

### `imported_event_timestamp_must_not_be_zero`

This result only applies when [`Transfer.flags.imported`](../transfer.md#flagsimported) is set.

The transfer was not created. The [`Transfer.timestamp`](../transfer.md#timestamp) is zero, but must
be a user-defined timestamp.

### `imported_event_timestamp_must_not_advance`

This result only applies when [`Transfer.flags.imported`](../transfer.md#flagsimported) is set.
```

docs/reference/requests/create_transfers.md
```diff
@@ -326,11 +368,48 @@ If the transfer has [`flags.balancing_debit`](../transfer.md#flagsbalancing_debi
...
+The transfer was not created. [`Transfer.credit_account_id`](../transfer.md#credit_account_id) must
+refer to an `Account` whose [`timestamp`](../account.md#timestamp) is less than the
+[`Transfer.timestamp`](../transfer.md#timestamp).
+
+### `imported_event_timeout_must_be_zero`
+
+This result only applies when [Transfer.flags.imported](../transfer.md#flagsimported) is set.
+
+The transfer was not created. The [`Transfer.timeout`](../transfer.md#timeout) is nonzero, but
+must be zero.
```
```suggestion

### `imported_event_timestamp_must_not_regress`

This result only applies when [`Transfer.flags.imported`](../transfer.md#flagsimported) is set.

The transfer was not created. The user-defined [`Transfer.timestamp`](../transfer.md#timestamp)
regressed, but it must be greater than the last timestamp assigned to any `Transfer` in the cluster and cannot be equal to the timestamp of any existing [`Account`](../account.md).

### `imported_event_timestamp_must_postdate_debit_account`

This result only applies when [`Transfer.flags.imported`](../transfer.md#flagsimported) is set.

The transfer was not created. [`Transfer.debit_account_id`](../transfer.md#debit_account_id) must
refer to an `Account` whose [`timestamp`](../account.md#timestamp) is less than the
[`Transfer.timestamp`](../transfer.md#timestamp).

### `imported_event_timestamp_must_postdate_credit_account`

This result only applies when [`Transfer.flags.imported`](../transfer.md#flagsimported) is set.

The transfer was not created. [`Transfer.credit_account_id`](../transfer.md#credit_account_id) must
refer to an `Account` whose [`timestamp`](../account.md#timestamp) is less than the
[`Transfer.timestamp`](../transfer.md#timestamp).

### `imported_event_timeout_must_be_zero`

This result only applies when [`Transfer.flags.imported`](../transfer.md#flagsimported) is set.

The transfer was not created. The [`Transfer.timeout`](../transfer.md#timeout) is nonzero, but
must be zero.
```

src/lsm/groove.zig
```diff
@@ -409,6 +409,59 @@ pub fn GrooveType(
         ObjectsCacheHelpers.tombstone,
     );
 
+    const TimestampSet = struct {
+        const map_load_percentage_max = 50;
```
sentientwaffle: (Unused)

src/lsm/groove.zig
sentientwaffle, line 1: Don't abbreviate "destin", per tigerstyle

src/lsm/groove.zig
```diff
@@ -409,6 +409,59 @@ pub fn GrooveType(
         ObjectsCacheHelpers.tombstone,
     );
 
+    const TimestampSet = struct {
```
sentientwaffle, line 411: Would `comptime assert(has_id);` work, or does this get compiled even when `!has_id`?

src/lsm/groove.zig
```diff
@@ -901,11 +1041,25 @@ pub fn GrooveType(
                 const worker = LookupContext.parent(.object, completion);
                 worker.lookup = undefined;
 
+                const destin = destin: {
+                    assert(worker.destin != null);
+                    defer worker.destin = null;
+                    break :destin worker.destin.?;
+                };
```
sentientwaffle: This is equivalent to the more succinct:

```
const destination = worker.destination.?;
worker.destination = null;
```

src/lsm/groove.zig
```diff
@@ -640,67 +719,120 @@ pub fn GrooveType(
...
+            // The mutable table needs to be sorted to enable searching by timestamp.
+            // The immutable table will be searched by `prefetch_from_memory_by_timestamp`.
+            groove.objects.table_mutable.sort();
+            if (groove.objects.table_mutable.get(timestamp)) |object| {
+                assert(object.timestamp == timestamp);
+                groove.timestamps.found(timestamp);
+                return;
+            }
+
+            groove.prefetch_from_memory_by_timestamp(timestamp, .timestamps);
```
sentientwaffle: :thinking: What do you think of:

- Replace `TimestampSet.enroll/found/exists` with `TimestampSet.has/get/set` (more familiar "Set" terminology -- I was initially confused by the semantics of `enroll`/`found`).
- Then `prefetch_exists_enqueue` looks like:

```
if (timestamps.has(timestamp)) return;

if (table_mutable.get(...)) {
    timestamps.insert(timestamp, true);
    return;
}
```

`insert` can assert `!has`.

I think this approach is a bit stricter -- with the current implementation, if due to some bug a key was missed by the async prefetch (or inserted into the wrong destination), `exists()` would "succeed" + return false, but with the above approach it would panic instead.

src/state_machine.zig
sentientwaffle, line 1: I guess we are implicitly safe from the high-order bit being set on a user-provided timestamp (reserved for the tombstone) since that would be ahead of the cluster clock. (Do we assert this somewhere though?)

src/state_machine.zig
```diff
@@ -744,6 +768,11 @@ pub fn StateMachineType(
                     self.forest.grooves.accounts.prefetch_enqueue(t.debit_account_id);
                     self.forest.grooves.accounts.prefetch_enqueue(t.credit_account_id);
                 }
+
+                if (t.flags.imported) {
+                    // Looking for accounts with the same timestamp.
+                    self.forest.grooves.accounts.prefetch_exists_enqueue(t.timestamp);
+                }
             }
```
sentientwaffle: To avoid rechecking the condition for every transfer, you could lift it out of the loop:

```
if (transfers.len > 0 and
    transfers[0].flags.imported)
{
    for (...)
}
```

src/state_machine.zig
```diff
@@ -2001,6 +2097,21 @@ pub fn StateMachineType(
...
+                // can still handle `exists` results when importing.
+                if (self.forest.grooves.transfers.objects.key_range) |*key_range| {
+                    if (t.timestamp <= key_range.key_max) {
+                        return .imported_event_timestamp_must_not_regress;
+                    }
+                }
+                if (self.forest.grooves.accounts.exists(t.timestamp)) {
+                    return .imported_event_timestamp_must_not_regress;
+                }
+            }
```
sentientwaffle, line 2121: Ah, so we don't need to check that this is newer than the credit/debit accounts since we verified that this is newer than all transfers (and therefore newer than the pending transfer, which is newer than the accounts)!

Can you add:

```
assert(t.timestamp > dr_account.timestamp);
assert(t.timestamp > cr_account.timestamp);
```

src/state_machine.zig
batiati: Good catch, we're now explicitly returning `imported_event_timestamp_out_of_range` it not `timestamp_min <= event.timestamp <= timestamp_max`.

src/lsm/groove.zig
batiati: > if due to some bug a key was missed by the async prefetch (or inserted into the wrong destination), exists() would "succeed" + return false, but with the above approach it would panic instead

Instead of a Set, the TimestampSet it's backed by a `HashMap(u64,bool)` precisely to assert that we can only call `exists` for timestamps that were enrolled before.

But I agree enroll/found/exists isn't a great API, changed it to enqueue/set/exists instead.

sentientwaffle: > Instead of a Set, the TimestampSet it's backed by a HashMap(u64,bool) precisely to assert that we can only call exists for timestamps that were enrolled before.

Right now we enroll/enqueue the timestamp immediately, before we know whether or not it exists. I was suggesting that we only insert the timestamp (along with `true` or `false`) once we know for sure.

With the existing code, if the prefetch result ends up in the map instead of the set, `.exists()` will return `false` instead of panicking, since we enrolled it immediately in `prefetch_exists_enqueue`.

src/lsm/groove.zig
```diff
@@ -410,7 +410,8 @@ pub fn GrooveType(
 
     const TimestampSet = struct {
         const TimestampSet = @This();
-        const Map = std.AutoHashMapUnmanaged(u64, bool);
+        const Found = enum { found, not_found };
```
sentientwaffle, line 413: Great idea to use an enum here so that it can't get confused with `has()`!

src/lsm/groove.zig
batiati: Done! We should probably do something similar with `prefetch_enqueue`, asserting that the key passed in `groove.get` was previously enqueued for prefetching.
I'll send it in a follow-up PR.

sentientwaffle (approved):

PR #2173, opened 2024-08-01
Test/VSR: Checkpoint op diagram

sentientwaffle:
I found this useful for visualizing for visualizing the checkpoint schedule (for "realistic" configuration values).


matklad (approved):
:tada: always wanted it as well.

This _might_ be subsumed by https://github.com/tigerbeetle/tigerbeetle/pull/2133, but only when I actually get to finishing that!

PR #2174, opened 2024-08-02
Test/VOPR: Test client eviction in the VOPR (2/2)

sentientwaffle:
Test client eviction in the VOPR.

This requires some shuffling of  `ReplySequence`, `Cluster`,and `Simulator`:
- We need to load replies into `ReplySequence` earlier -- when the reply is generated by the replica, rather than when the reply actually reaches the client (since the client may be evicted before a reply arrives).
- As such, `Cluster` now distinguishes between `on_cluster_reply` and `on_client_reply`.
- As `Simulator` ignores inflight requests by evicted clients for the purposes of `Simulator.done()`.

Note that with this code, the VOPR would be capable of finding the bug fixed by https://github.com/tigerbeetle/tigerbeetle/pull/2153!

Part 1: https://github.com/tigerbeetle/tigerbeetle/pull/2153
Part 1.5: https://github.com/tigerbeetle/tigerbeetle/pull/2170

### DIFF
--- a/src/vopr.zig
+++ b/src/vopr.zig
@@ -203,6 +203,7 @@ pub fn main() !void {
                 .cache_entries_account_balances = 256,
             },
         },
+        .on_cluster_reply = Simulator.on_cluster_reply,
     };
 
     const workload_options = StateMachine.Workload.Options.generate(random, .{
@@ -453,7 +454,7 @@ pub const Simulator = struct {
         assert(options.request_idle_off_probability > 0);
         assert(options.request_idle_off_probability <= 100);
 
-        var cluster = try Cluster.init(allocator, on_cluster_reply, options.cluster);
+        var cluster = try Cluster.init(allocator, options.cluster);
         errdefer cluster.deinit();
 
         var workload = try StateMachine.Workload.init(allocator, random, options.workload);
@@ -801,8 +802,8 @@ pub const Simulator = struct {
     fn on_cluster_reply(
         cluster: *Cluster,
         reply_client: usize,
-        request: *Message.Request,
-        reply: *Message.Reply,
+        request: *const Message.Request,
+        reply: *const Message.Reply,
     ) void {
         const simulator: *Simulator = @ptrCast(@alignCast(cluster.context.?));
         simulator.reply_sequence.insert(reply_client, request, reply);
--- a/src/vopr.zig
+++ b/src/vopr.zig
@@ -203,7 +203,8 @@ pub fn main() !void {
                 .cache_entries_account_balances = 256,
             },
         },
-        .on_client_reply = Simulator.on_cluster_reply,
+        .on_cluster_reply = Simulator.on_cluster_reply,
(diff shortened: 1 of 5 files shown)


sentientwaffle:
(Rebased to resolve conflict.)

src/vopr.zig
```diff
@@ -118,7 +118,10 @@ pub fn main() !void {
     const standby_count =
         if (cli_args.lite) 0 else random.uintAtMost(u8, constants.standbys_max);
     const node_count = replica_count + standby_count;
-    const client_count = 1 + random.uintLessThan(u8, constants.clients_max);
+    // -1 since otherwise it is possible that all clients will evict each other.
+    // (Due to retried register messages from the first set of evicted clients.
+    // See the "Cluster: eviction: session_too_low" replica test for a related scenario.)
+    const client_count = @max(1, random.uintAtMost(u8, constants.clients_max * 2 - 1));
```
matklad, line 124: :tada:

src/vopr.zig
```diff
@@ -436,6 +439,8 @@ pub const Simulator = struct {
     /// Total number of requests sent, including those that have not been delivered.
     /// Does not include `register` messages.
     requests_sent: usize = 0,
+    /// Total number of replies received by non-evicted clients.
+    /// Does not include `register` messages.
     requests_replied: usize = 0,
```
matklad, line 444: Hm, do we now count `requests_replied` correctly? We update it `on_client_reply`, not `on_cluster_reply`, and it is at least non obvious why a reply won't get lost. I am thinking about the following scenario:

* client sends a request
* cluster processes it, sends a reply
* reply is lost
* a different client is registered, so the first one is evicted
* original client retries the request
* it gets an eviction message

matklad: Note: vopr_lite considers this a success, as that's a liveness issue

sentientwaffle: :thinking: `requests_replied` indeed only counts replies which actually reach a (non-evicted) client.

But in turn we subtract `simulator.requests_cancelled()` from `requests_sent`.
When `tick_requests()` runs:

```
        if (simulator.requests_sent - simulator.requests_cancelled() ==
            simulator.options.requests_max) return;
```

`requests_cancelled` means we don't count that request as "sent", since the client will never receive a reply:

```
    fn requests_cancelled(simulator: *const Simulator) u32 {
        var count: u32 = 0;
        for (
            simulator.cluster.clients,
            simulator.cluster.client_eviction_reasons,
        ) |*client, reason| {
            count += @intFromBool(reason != null and
                client.request_inflight != null and
                client.request_inflight.?.message.header.operation != .register);
        }
        return count;
    }
```

So `tick_requests` will trigger an additional request.

matklad: Hm, yeah, it actually is pretty obvious! I somehow managed to convince myself that it s not enough, but, no, this should be enough!

matklad (approved):

PR #2175, opened 2024-08-02
StateMachine: Fix balancing transfers with large balances

sentientwaffle:
### Context

When a balancing transfer is created with `amount = 0`, the transfer should move as much money as possible (within the constraints of the respective accounts).

### Bug

Balancing transfers used on accounts with net balances greater than `2^64 - 1` would not move as much money as they should.

(There was a leftover `u64` from before amounts/balances were changed to `u128`.)

### DIFF
--- a/src/state_machine.zig
+++ b/src/state_machine.zig
@@ -1791,7 +1791,8 @@ pub fn StateMachineType(
             const amount = amount: {
                 var amount = t.amount;
                 if (t.flags.balancing_debit or t.flags.balancing_credit) {
-                    if (amount == 0) amount = std.math.maxInt(u64);
+                    comptime assert(@TypeOf(amount) == u128);
+                    if (amount == 0) amount = std.math.maxInt(u128);
                 } else {
                     assert(amount != 0);
                 }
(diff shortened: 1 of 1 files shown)


src/state_machine.zig
```diff
@@ -3866,6 +3866,33 @@ test "create_transfers: balancing_debit | balancing_credit (amount=0)" {
...
+        \\ account A2  0  0  0  0  _  _  _ _ L1 C1   _ D<C   _ _ _ _ ok
+        \\ account A3  0  0  0  0  _  _  _ _ L1 C1   _ D<C   _ _ _ _ ok
+        \\ account A4  0  0  0  0  _  _  _ _ L1 C1   _ D<C   _ _ _ _ ok
+        \\ commit create_accounts
+        \\
+        \\ setup A1 0  0 0 -1
+        \\ setup A4 0 -1 0  0
+        \\
+        \\ transfer   T1 A1 A2    0   _  _  _  _    _ L1 C1   _   _   _   _ BDR   _  _ _ ok
+        //\\ transfer   T2 A3 A4    0   _  _  _  _    _ L1 C1   _   _   _ BDR   _   _  _ _ ok
```
batiati: Is it supposed to be commented?

batiati (commented):
Nice catch!! It seems like it's been around since 128-bit balances were introduced. Glad you spotted it!

src/state_machine.zig
sentientwaffle: :facepalm: No, I meant to add that back once I got the first part working, but forgot. Thanks!

batiati (approved):

PR #2176, opened 2024-08-02
vopr: test protocol upgrades

matklad:


matklad:
As per https://github.com/tigerbeetle/tigerbeetle/pull/1951#issuecomment-2265856660, this acutally tests way more than what I intend to test!

PR #2178, opened 2024-08-04
Minor improvements to zig install scripts

brson:
Three minor changes:

- Set the mode on install_zig.bat to executable. This makes it executable by cygwin.
- Modify install_zig.sh to work with cygwin.
- Modify install_zig.sh not to display stray "/usr/bin/wget" output when probing for wget.

I have not tested on macos, but it still looks correct to me.

### DIFF
--- a/zig/download.sh
+++ b/zig/download.sh
@@ -26,11 +26,21 @@ else
 fi
 
 # Determine the operating system:
-if [ "$(uname)" = "Linux" ]; then
-    ZIG_OS="linux"
-else
-    ZIG_OS="macos"
-fi
+case "$(uname)" in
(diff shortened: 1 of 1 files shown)


jorangreef:
@brson my apologies! I only saw your work here after conflicting nicely with #2180, would you please layer on after that lands?

brson:
I've rebased this and retested. I removed the commit to set the executable bit on download.bat, as I was finding it confusing on linux - autocomplete was suggesting it, and executing it would fail. With install.sh being cygwin compatible, that is a reasonable choice for running under cygwin anyway.

cb22 (approved):

PR #2179, opened 2024-08-04
Refactor Scan iterators.

batiati:
Refactor (and rename) iterators that used to be shared between compaction and scans, simplifying them to serve only Scans (since compaction doesn't use them anymore).

Consolidates the logic from `level_data_iterator.zig` into `ScanTreeLevelType`.

Also, rename some occurrences of `data_block` to `value_block`.

*Best reviewed per commit


src/lsm/table_value_iterator.zig
```diff
@@ -82,14 +54,14 @@ pub fn TableDataIteratorType(comptime Storage: type) type {
...
-        pub fn empty(it: *const TableDataIterator) bool {
+        pub fn empty(it: *const TableValueIterator) bool {
             assert(it.context.addresses.len == it.context.checksums.len);
             return it.context.addresses.len == 0;
         }
 
         /// Calls `callback` with either the next data block or null.
         /// The block is only valid for the duration of the callback.
-        pub fn next(it: *TableDataIterator, callback: Callback) void {
+        pub fn next(it: *TableValueIterator, callback: Callback) void {
```
matklad: WDYT about renaming this `next` to `next_value_table`? The main reason is that it's a bit hard to find call-sites of this function by just grepping the code, as `next` is not particularly unique.

matklad: Another thing that's a bit awkward about this API is the next tick, as we can trivially know whether we need a next tick or not. So perhaps we could _start_ this function with

```
assert(!it.empty())
```

and then let the caller handles synchroneous path explicitly? It feel like that'll actually _simplify_ `value_block_callback` as it won't need to handle the empty case

src/lsm/table_value_iterator.zig
```diff
@@ -110,13 +82,16 @@ pub fn TableDataIteratorType(comptime Storage: type) type {
...
             } else {
+                assert(it.context.addresses.len == 0);
+                assert(it.context.checksums.len == 0);
+
                 it.callback = .{ .next_tick = callback };
                 it.context.grid.on_next_tick(on_next_tick, &it.next_tick);
             }
         }
 
         fn on_read(read: *Grid.Read, block: BlockPtrConst) void {
```
matklad: In on read, feels like we don't need `constants.verify`, as verification is `O(1)` per block, so shouldn't be visible in profile

src/lsm/scan_tree.zig
```diff
@@ -760,72 +817,108 @@ fn ScanTreeLevelType(comptime ScanTree: type, comptime Storage: type) type {
...
-            // the key depending on the `key_min`.
-            const end = end: {
-                const keys_min = Table.index_data_keys_used(
-                    iterator.context.index_block,
-                    .key_min,
-                );
-                break :end indexes.end + @intFromBool(
-                    indexes.end < keys.len and keys_min[indexes.end] <= scan.key_max,
+            const Range = struct { start: u32, end: u32 };
+            const range_found: ?Range = range: {
```
matklad, line 848: Hm, how come this could be null?

Manifest remembers both `key_min` and `key_max`, so we should be able to rule this out without IO?

Ahhh, no: we could have two keys in the table, `1` and `9999`, and, if we are looking for range `[10, 20]` then we have to load its index block from disk.

I wonder if we could add some asserts about table_info.key_range vs scan.key_range? It seems like the above case the is the only one? That is, that

```
scan.min < table.min < table.max < scan.max
```

should hold? Otherwise at least `table.key_min` or `table.key_max` should get in

matklad: Ah, we actually have a similar asser for data blocks:

```
                    assert(key_min < scan.key_min and
                        scan.key_max < key_max);

```

so we could do it for entire tables as well

src/lsm/scan_tree.zig
```diff
@@ -568,92 +605,109 @@ fn ScanTreeLevelType(comptime ScanTree: type, comptime Storage: type) type {
...
-            self.iterator.next(.{
-                .on_index = index_block_callback,
-                .on_data = data_block_callback,
-            });
+            switch (self.state) {
+                .iterating_manifest => unreachable,
+                .loading_index => |*loading_index| {
+                    assert(self.values == .fetching);
+                    // Reading the index blocks:
+                    const grid: *Grid = self.scan.tree.grid;
```
matklad: I think we generally try to avoid aliases? So

```
                    self.scan.tree.grid.read_block(
                        .{ .from_local_or_global_storage = index_block_callback },
                        &loading_index.read,
                        loading_index.address,
                        loading_index.checksum,
                        .{ .cache_read = true, .cache_write = true },
                    );
```

src/lsm/scan_tree.zig
```diff
@@ -568,92 +605,109 @@ fn ScanTreeLevelType(comptime ScanTree: type, comptime Storage: type) type {
...
+                    grid.on_next_tick(
+                        finished_callback,
+                        &finished.next_tick,
+                    );
+                },
+            }
         }
 
         pub fn peek(self: *const ScanTreeLevel) error{ Drained, Empty }!Key {
```
matklad, line 637: Out out scope of this PR, but one thing I am finding slightly unsatisfactory about the scans is that the implementation is very scalar, it doesn't batch. 

So for example, kway merge operates on a value at a time: in the inner loop for the kway merge there's a break edge for each iteration.

At the same time, our data is fundmentally batched --- a value block contains a whole bunch of values. Table memory is also an array. 

It seems like it would be more in-line with our batching philosophy to do a batching merge, pushing fors down:


```
fn kmerge(
    inputs: [][]const Value,
    output: []Value
) struct {
   values_written: usize, 
   input_exhausted: usize, // index into inputs
}
```

That is, `kmerge` takes entire _slices_, and, in the inner loop, merges slices until one of them is exhausted. The exhausted slice is communicated by the caller (the amount of values consumed is communicated by mutating `inputs`). The caller then does some asynchronous work to re-fill the exhausted batch, and then re-runs `kmerge`. 

One obstacle here is that our inputs are not acteally slices. Or, rather, they are slices for levels, but they have some duplicates for table_memory case. But it _should_ be possible to dedup when we sorting, right?

matklad (commented):
LGTM with:

* needless `const grid: *Grid = self.scan.tree.grid;` removed
* extra assert for `range = null` case

I think it'd be helpful to try to get rid of on_next_tick by pushing the synchroneous case up to the caller, but that can be done in a follow up (or maybe that's not really possible).

I am really intrigued by trying to do a "batching" kway merge, as it seems that could yield a much tighter inner loop, but that's definitely a larger mission!

src/lsm/scan_tree.zig
batiati: > It seems like it would be more in-line with our batching philosophy to do a batching merge

I think we would do more CPU work than it's needed.
- It requires a buffer to copy the kway merged output.
- The query could have a limit=X, so we don't need to merge all buffered keys;
- The query can combine multiple scans in a UNION/INTERSECTION, so each scan moves (and consumes the buffer) in a different frequency (we apply the limit for the query, not for each scan);

We might want to take a different approach and batch queries instead of merges.
- Reserve grid blocks (so they don't evict from the cache).
  It would allow us to iterate directly from the values returned from the Grid without copying to a buffer.
- Without extra buffers we increase the limit of `lsm_scans_max` relative to the size of the Grid cache / lsm_levels.
- With more scans, we can batch _multiple queries_ in the same request:
  Other than the query initialization (when all LSM levels need to be loaded from storage), during the execution, each level fetches data independently, not using much of the available IO. It's a great opportunity to batch them together.

matklad: Ahhhh, right! I was thinking that the `output` would be just the final message where we are copying the value into (this would also handle the limit case, as the merge could stop if the _output_ is exhausted as well), but, indeed, when we combine several scans that doesn't work.

batiati:
> I think it'd be helpful to try to get rid of on_next_tick by pushing the synchroneous case up to the caller, but that can be done in a follow up (or maybe that's not really possible).

Yep, done! It simplified the internal state even further!  🧑‍🍳😙🤌 
Let's let it vopr'ing for a while before we merge.

src/lsm/scan_tree.zig
```diff
@@ -829,39 +861,42 @@ fn ScanTreeLevelType(comptime ScanTree: type, comptime Storage: type) type {
             const loading_index: *LoadingIndex = @fieldParentPtr("read", read);
             const state: *State = @fieldParentPtr("loading_index", loading_index);
             const self: *ScanTreeLevel = @fieldParentPtr("state", state);
-            const scan: *const ScanTree = self.scan;
```
matklad, line 771: :+1:

src/lsm/scan_tree.zig
```diff
@@ -821,6 +763,96 @@ fn ScanTreeLevelType(comptime ScanTree: type, comptime Storage: type) type {
...
+                        // Keep iterating to the next `value_block`.
+                        self.values = .fetching;
+                    }
+                },
+                .finished => unreachable,
+            }
+        }
+
+        /// Moves the iterator to the next `table_info` that might contain the key range.
+        fn move_manifest_table_next(
```
matklad: I'd call it `move_next_manifest_table`: the base here is `move_next`, within this larger `move_next` we have `move_next_manifest_table` suboperation. In this case, the qualifier is not "next", but rather "manifest table".

A general rule of thumb: if you have function `foo` and it calls function `bar`, and no-one else calls `bar`, then `bar` probably should be called `foo_bar` --- _likely_ it'll follow MSB semantics, _and_ it'll sort nicely in the source file!

matklad (dismissed):

matklad (approved):

PR #2180, opened 2024-08-05
Replace scripts directory with zig download scripts

jorangreef:
The goal here is to reduce the number of files and directories in the root directory, so as to minimize cognitive overload when viewing the root directory on GitHub or in the terminal.

Starting with removing the `scripts` directory...

There are a few problems with `scripts/install_zig.sh` (and `.bat`):

* They download to the local `zig` directory and don't actually install anything.
* They are hidden within the `scripts` directory and are not discoverable.
* In other words, if someone wants to start playing with Zig it's not clear how to do that.
* They are the sole reason for the `scripts` directory to exist.
* They contain a few too many characters to type.
* They don't follow TigerStyle MSB (most significant byte) naming convention.
* When Zig is downloaded, this adds yet another directory (!) to the root.

While we could move these scripts to the root directory, this would remove one directory at the cost of adding two files.

Instead, rename these scripts to `zig/download.sh` (and `.bat`).


.github/ci/docs_check.sh
```diff
@@ -24,9 +24,9 @@ cd /build
 # Validate dictionary is formatted correctly
 curl -L -o /tmp/jq https://github.com/stedolan/jq/releases/download/jq-1.6/jq-linux64
 chmod +x /tmp/jq
-cat scripts/.cspell.json | /tmp/jq empty
+cat docs/.cspell.json | /tmp/jq empty
```
cb22, line 27: :tada:

.gitignore
```diff
@@ -7,7 +7,6 @@ simulator
 /tigerbeetle.exe
 *_*.tigerbeetle.*
 *_*.tigerbeetle
-zig/
 zig-cache/
```
cb22: Do we need `zig-cache/` anymore, now that 0.13 has moved to `.zig-cache/`? Unrelated :smile:

zig/.gitignore
```diff
@@ -0,0 +1,6 @@
+LICENSE
+README.md
+doc
+lib
+zig
+zig.exe
```
cb22: Do we want to be explicit about what's in the `zig` tarball? If not, we could do

```suggestion
*
!download.bat
!download.sh
```

Even if we do, we might want to still do this and then keep the manual `mv`'ing to not have 2 places to update.

zig/download.bat
```diff
@@ -60,33 +60,11 @@ if exist %ZIG_TARBALL% (
 echo Downloading %ZIG_URL%...
 curl --silent --progress-bar --output %ZIG_TARBALL% %ZIG_URL%
 if not exist  %ZIG_TARBALL% (
-  echo Failed to download Zig zip file.
+  echo Failed to download zip file.
   exit 1
 )
 
-:: Replace any existing Zig installation so that we can install or upgrade:
```
cb22, line 67: Out of scope for this PR, but it would be great to get rid of this whole bat script and replace it with a powershell one.

zig/download.sh
```diff
@@ -76,17 +76,22 @@ else
...
-rm -rf zig
-mv "$ZIG_DIRECTORY" zig
+# Replace these existing directories and files so that we can install or upgrade:
+rm -rf zig/doc
+rm -rf zig/lib
+mv "$ZIG_DIRECTORY/LICENSE" zig
+mv "$ZIG_DIRECTORY/README.md" zig
+mv "$ZIG_DIRECTORY/doc" zig
+mv "$ZIG_DIRECTORY/lib" zig
+mv "$ZIG_DIRECTORY/zig" zig
```
cb22: A nit, and the old script didn't do this, but using a trailing slash (`zig/`) to indicate the target is a directory is nice.

zig/.gitignore
cb22: Also, no trailing newline.

zig/download.bat
```diff
@@ -96,18 +74,38 @@ if not exist %ZIG_TARBALL% (
...
-:: Removes the zip file
 del /q %ZIG_TARBALL%
 if exist %ZIG_TARBALL% (
-  echo Failed to delete %ZIG_TARBALL% file.
+  echo The %ZIG_TARBALL% file could not be deleted.
   exit 1
 )
 
-echo "Congratulations, you have successfully installed Zig version %ZIG_RELEASE%. Enjoy!"
+echo "Downloading completed (%ZIG_DIRECTORY%\zig.exe)! Enjoy!"
```
cb22, line 111: Ditto the trailing newline

zig/download.bat
jorangreef: Indeed!

cb22 (approved):

PR #2181, opened 2024-08-05
vsr: rollout new sync protocol, stage 0 

matklad:
This is a pre-requisite for
https://github.com/tigerbeetle/tigerbeetle/pull/1951

To avoid needlessly disturbing the clients and to simplify upgrade rollout procedure, we want to avoid bumping VSR Protocol version and instead introduce a new command for the new SV.

Then, for rollout, replicas will be sending _both_ versions of SV. To make this work though, it is important that a replica on the old version ignores SV from the new version. Right now replicas are super-strict around receiving unknown message commands, and intentionally panic.

This commit implements the opt-in ignoring for new SVs.

I choose the somewhat awkward StartView2 name:

- while we perhaps can invent a new name here, I don't want to rename (or add a separate command for) `request_start_view`
- once the upgrade completes, StartView2 will get renamed back to StartView

### DIFF
--- a/src/vsr.zig
+++ b/src/vsr.zig
@@ -219,6 +219,11 @@ pub const Command = enum(u8) {
     request_sync_checkpoint = 21,
     sync_checkpoint = 22,
 
+    // A reserved command for the new state sync protocol. At the moment, replica ignores this
+    // command (as opposed to panicking on an unknown command), to allow the next release to send
+    // both versions of StartView.
+    start_view2 = 23,
+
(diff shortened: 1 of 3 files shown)


matklad:
https://github.com/tigerbeetle/tigerbeetle/pull/1951 was updated to include the next stage of rollout. Makes sense to take a look at that to make sure that it all makes sense!

sentientwaffle (approved):
LGTM!

PR #2182, opened 2024-08-05
client: print only active addresses

Rexicon226:
I was playing around with the client and noticed it hit an `unreachable` while printing the parsed addresses. It seems like `BoundedArray`'s `{any}` format prints the entire inner buffer, even the uninitialized ones. 

I changed the print to use `.const_slice()` which returns the initialized addresses.


kprotty (approved):

matklad:
Good catch, and mea culpa!

PR #2184, opened 2024-08-06
multiversioning: relax assert in exec_latest

cb22:
Fixes an overly strict assertion in `exec_latest`. `target_fd` is double buffered, and only ever updated in `target_update`, so it's safe to call `exec_latest` as long as:
* `stage` is not `.target_update` (should never happen, as it's a sync fn),
* `target_fd` has been populated (checked through `self.target_header`).

~Marked as draft because:~
~* I haven't tested this yet,~
~* Not yet sure if we'll backport etc.~

### DIFF
--- a/src/multiversioning.zig
+++ b/src/multiversioning.zig
@@ -1054,7 +1054,13 @@ pub const Multiversion = struct {
     }
 
     pub fn exec_latest(self: *Multiversion) !noreturn {
-        assert(self.stage == .ready);
+        // target_fd is only modified in target_update() which happens synchronously.
+        assert(self.stage != .target_update);
+
+        // Ensure that target_update() has been called at least once, and thus target_fd is
+        // populated by checking that target_header has been set.
+        assert(self.target_header != null);
+
         log.info("re-executing {s}...\n", .{self.exe_path});
         try self.exec_target_fd();
     }
--- a/src/multiversioning.zig
+++ b/src/multiversioning.zig
@@ -356,7 +356,7 @@ pub const MultiversionHeader = extern struct {
     vsr_releases_max: u32 = constants.vsr_releases_max,
 
     /// The current release is executed differently to past releases embedded in the body, so store
-    /// it separately. See exec_latest vs exec_release.
+    /// it separately. See exec_current vs exec_release.
     current_release: u32,
 
     current_flags: Flags,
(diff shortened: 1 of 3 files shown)


src/multiversioning.zig
```diff
@@ -1054,7 +1054,13 @@ pub const Multiversion = struct {
...
     pub fn exec_latest(self: *Multiversion) !noreturn {
-        assert(self.stage == .ready);
+        // target_fd is only modified in target_update() which happens synchronously.
+        assert(self.stage != .target_update);
+
+        // Ensure that target_update() has been called at least once, and thus target_fd is
+        // populated by checking that target_header has been set.
+        assert(self.target_header != null);
+
         log.info("re-executing {s}...\n", .{self.exe_path});
```
matklad: Is `\n`  intentional here?

src/multiversioning.zig
```diff
@@ -1054,7 +1054,13 @@ pub const Multiversion = struct {
     }
 
     pub fn exec_latest(self: *Multiversion) !noreturn {
-        assert(self.stage == .ready);
+        // target_fd is only modified in target_update() which happens synchronously.
+        assert(self.stage != .target_update);
```
matklad, line 1058: Do we want `!= .err`?

matklad: Ah, ok, no, we _don't_ need `.err` --- even if we failed to update our `target_fd`, the old version is still fine!

Let's though pass in the version we are intended to switch to as an argument, so that we can check, at the last moment, that `target_fd` indeed contains the version we want to execute. 

We have that check in main as a condition:

```
    for (replica.releases_bundled.const_slice()) |release_bundled| {
        if (release_bundled.value == release.value) break;
    } else {
        log.err("{}: release_execute: release {} is not available;" ++
            "upgrade (or downgrade) the binary", .{
            replica.replica,
            release,
        });
        @panic("release not available");
    }
```

But we should also do an equivalenet at the latest possible moment as an assert.

src/multiversioning.zig
cb22: Yep - it gives a bit of visual separation in the logs from one version to the next.

src/multiversioning.zig
cb22: > Ah, ok, no, we don't need .err --- even if we failed to update our target_fd, the old version is still fine!

Yep!

> Let's though pass in the version we are intended to switch to as an argument, so that we can check, at the last moment, that target_fd indeed contains the version we want to execute.

Good idea, will do.

src/multiversioning.zig
matklad: Might worth a comment. As it stands, it _will_ get cleanupd at some point by accident

matklad (approved):

PR #2186, opened 2024-08-06
ci: fix release validation

matklad:
It runs on the release branch and doesn't have zig/download yet

### DIFF
--- a/.github/workflows/release_validate.yml
+++ b/.github/workflows/release_validate.yml
@@ -38,7 +38,8 @@ jobs:
         with:
           node-version: 'latest'
 
-      - run: ./zig/download.sh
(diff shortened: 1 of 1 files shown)


.github/workflows/release_validate.yml
```diff
@@ -38,7 +38,7 @@ jobs:
         with:
           node-version: 'latest'
 
-      - run: ./zig/download.sh
+      - run: ./zig/download.sh | ./scripts/install_zig.sh
```
batiati, line 42: _When_ will we be able to remove it? Is it worth a comment?

matklad: Good call, added a comment!

batiati (approved):

PR #2187, opened 2024-08-06
test: smoke test for upgrades

matklad:
This is in no way integrated with our CI story, but, having seen the upgrade procedure with my own eyes when running manually against 0.15.3 and 0.15.4, I am so much confident now!

```
$ gh release download 0.15.3 --pattern tigerbeetle-x86_64-linux.zip
$ unzip tigerbeetle-x86_64-linux.zip && mv tigerbeetle tigerbeetle-0.15.3

$ gh release download 0.15.4 --pattern tigerbeetle-x86_64-linux.zip --clobber
$ unzip tigerbeetle-x86_64-linux.zip && mv tigerbeetle tigerbeetle-0.15.4

$ ./zig/zig build scripts -- upgrader --old=./tigerbeetle-0.15.3 --new=./tigerbeetle-0.15.4
```

### DIFF
--- a/src/scripts/upgrader.zig
+++ b/src/scripts/upgrader.zig
@@ -1,0 +1,141 @@
+/// Smoke test for upgrade procedure:
+/// - given two multiversion tigerbeetle binaries, old and new
+/// - format three data files using old
+/// - spawn a benchmark load using old against a local cluster
+/// - spawn three replicas at old
+/// - for some time in a loop, crash, restart, or upgrade a random replica; then
+/// - crash all replicas and restart them at the new version
+/// - join the benchmark load, to make sure it exits with zero.
+const std = @import("std");
+const builtin = @import("builtin");
+const log = std.log;
+const assert = std.debug.assert;
+
+const stdx = @import("../stdx.zig");
+const flags = @import("../flags.zig");
+const fatal = flags.fatal;
+const Shell = @import("../shell.zig");
+
+pub const CliArgs = struct {
+    old: []const u8,
+    new: []const u8,
+};
+
+const replica_count = 3;
+
+pub fn main(shell: *Shell, gpa: std.mem.Allocator, cli_args: CliArgs) !void {
+    _ = gpa;
+
+    const seed = std.crypto.random.int(u64);
+    log.info("seed = {}", .{seed});
+    var rng = std.rand.DefaultPrng.init(seed);
+    const random = rng.random();
+
+    try shell.exec("{tigerbeetle} version", .{ .tigerbeetle = cli_args.old });
+    try shell.exec("{tigerbeetle} version", .{ .tigerbeetle = cli_args.new });
+
+    shell.project_root.deleteTree(".zig-cache/upgrader") catch {};
+    try shell.project_root.makePath(".zig-cache/upgrader");
+
+    for (0..replica_count) |replica_index| {
+        try shell.exec(
+            \\{tigerbeetle} format --cluster=0 --replica={replica} --replica-count=3
+            \\    .zig-cache/upgrader/0_{replica}.tigerbeetle
+        , .{
+            .tigerbeetle = cli_args.old,
+            .replica = replica_index,
+        });
+    }
+
+    log.info("cluster at --addresses=127.0.0.1:7001,127.0.0.1:7002,127.0.0.1:7003", .{});
+
+    var load = try shell.spawn(.{ .stderr_behavior = .Inherit },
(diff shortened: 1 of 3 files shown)


matklad:
I intend to make this into a proper test of course, but I first want to wait until the dust settles in our release.zig!

src/scripts/upgrader.zig
```diff
@@ -0,0 +1,141 @@
+/// Smoke test for upgrade procedure:
+/// - given two tigerbeetle binaries, old and new
```
cb22: These would need to be two releases (ie, multiversion binaries) and not just the output from `zig build`, correct?

src/scripts/upgrader.zig
```diff
@@ -0,0 +1,141 @@
...
+/// - format three data files using old
+/// - spawn a benchmark load using old against a local cluster
+/// - spawn three replicas at old
+/// - for some time in a loop, crash, restart, or upgrade a random replica; then
+/// - crash all replicas and restart them  at the new version
```
cb22:
```suggestion
/// - crash all replicas and restart them at the new version
```

src/scripts/upgrader.zig
```diff
@@ -0,0 +1,141 @@
...
+/// - joint the benchmark load, to make sure it exits with zero
```
cb22:
```suggestion
/// - join the benchmark load, to make sure it exits with zero
```
?

src/scripts/upgrader.zig
```diff
@@ -0,0 +1,141 @@
...
+                replicas[replica_index] = try spawn(shell, .{
+                    .tigerbeetle = if (replicas_upgraded[replica_index])
+                        cli_args.new
+                    else
+                        cli_args.old,
+                    .replica = replica_index,
+                });
+            }
+        } else {
+            if (crash) {
```
cb22, line 95: Before killing / upgrading, can we assert the `load` is still running?

src/tigerbeetle/benchmark_load.zig
```diff
@@ -423,7 +423,7 @@ const Benchmark = struct {
         const ms_time = @divTrunc(batch_end_ns - b.batch_start_ns, std.time.ns_per_ms);
 
         if (b.print_batch_timings) {
-            log.info("batch {}: {} tx in {} ms\n", .{
+            log.info("batch {}: {} tx in {} ms", .{
```
cb22, line 426: Finally! :rofl:

src/scripts/upgrader.zig
matklad: Correct (though I am leaning that we should move multiversioning to `zig build` so that `zig build -Dmultiversion` produces one)! 

Clarified

src/scripts/upgrader.zig
matklad: https://type.method.ac

src/scripts/upgrader.zig
matklad: Not really I think, for two reasons:

* there isn't an API in Zig to check if the process is still running, without actually waiting for it (that's why we have that funky "make stdout non-blocking" code in devhub)
* I don't think it is _guranteed_ that the benchmark won't finish earlier. I did time the iteration number and transfer count such that benchmark usually outlives the chaotic period

Ideally, we should let then benchmark just run forever, and then tell it to stop after some time, but that needs non-trivial changes in the benchmark (we should catch SIGUSR1 or something?), and, worse, it needs changes in benchmark in 0.15.3, as it is important for the test that we use the benchmark with old client release

src/tigerbeetle/benchmark_load.zig
matklad: Too bad we aren't actually runnig  _this_ code, and instead use the one from downloaded release!

src/scripts/upgrader.zig
cb22: Re `zig build -Dmultiversion` - see also [this thread](https://tigerbeetle.slack.com/archives/C04375JUAMS/p1722902746155869) about packaging all this for Nix.

matklad:
Fixed typos and also tweaked the limits a bit to make it finish faster.

Oh, one note: this test manually restarts replicas, and doesn't check the "replace binary on disk" flow. For two reasons:

* 0.15.3 doesn't have that anyway
* I consider that feature to be a nice to have, the baseline we should take care of first is that restarting manually gets the upgrade going

cb22 (approved):

PR #2188, opened 2024-08-07
Add more systemd hardening params and update systemd readme

DanielSidhion:
This PR adds a bunch more systemd hardening params to lock the service even further.

A section on the systemd readme was added with instructions for development environments (someone said on Slack they had to figure this out by themselves, so might as well have it in the readme).

The new systemd service was tested on a single replica cluster and a 3-replica cluster. Upgrades weren't tested, but I hope to do that once a version greater than 0.15.4 lands due to the unique upgrade process from 0.15.3 to 0.15.4.

I read through TB's source code for any mentions of `posix` and `linux` to check the involved syscalls and other things the TB code was relying on, and tried to lock down almost everything else that I could understand from systemd's docs. The testing I did wasn't extensive, so I appreciate any further testing from others.


cb22:
Thanks @DanielSidhion!

Looks like the [spell check](https://github.com/tigerbeetle/tigerbeetle/blob/deeb547da5297c9e8f834409e7aa10e868efe824/.github/ci/docs_check.sh#L30) is trying to check the syste*md* unit file: https://github.com/tigerbeetle/tigerbeetle/actions/runs/10278086700/job/28441172616 :smile:

DanielSidhion:
@cb22 I added more words to the spellcheck dictionary, looks like the CI job for that succeeds now. Thanks for calling it out :)

cb22 (approved):

PR #2189, opened 2024-08-07
tidy: use tidy instead of gitattributes

matklad:
Let's leave it up to the user to instruct their git to checkout line
endings as is, and instead enforce UNIX line endings with tidy.

The reason why `.gitattributes` is needed is that git on windows has a pretty bad default of `autocrlf = true`. The correct fix is to configure your git not to eat your data!

As a side-effect, we are no longer mixing tabs and spaces!


src/tidy.zig
```diff
@@ -62,30 +80,104 @@ test "tidy" {
...
+    if (mem.indexOfScalar(u8, file.text, '\r') != null) {
+        if (std.mem.endsWith(u8, file.path, ".bat")) return null;
+        return '\r';
+    }
+
+    // Learning the best from UNIX, Visual Studio, like make, insists on tabs.
+    if (std.mem.endsWith(u8, file.path, ".sln")) return null;
+    // Go code uses tabs.
+    if (std.mem.endsWith(u8, file.path, ".go") or
+        (std.mem.endsWith(u8, file.path, ".md") and mem.indexOf(u8, file.text, "```go") != null))
```
batiati, line 168: Since all .md with GO code should be from auto-generated client readmes, WDYT of replacing tabs directly at `client_readmes.read_section()`?

batiati (approved):

src/tidy.zig
matklad: I assume we actually want to have tabs, to make the code copy-pastable?

batiati: > to make the code copy-pastable?

Not a big deal I think.

Tabs are just idiomatic for Go and enforced by `go fmt`. The code might still be valid, differently from Zig, which rejects tabs at the compiler level! :laughing: 

Anyway, it's good to have 100% idiomatic samples there.

PR #2190, opened 2024-08-07
VSR: Evict clients that try to change releases mid-session

sentientwaffle:
Evict clients that try to change releases mid-session. (This would indicate a client bug.)


matklad (approved):
🤯 excellent catch!

PR #2193, opened 2024-08-08
build: refactor client bindings generation

matklad:
Our binding generated code survived a number of code motions, and
accumulated a bunch of accidental complexity. This new implementation
aims at simplifying the things overall.

Bindings are source files in target language (.java, .go, etc), which
are generated from our tigerbeetle.zig code. We want the generated files
to change automatically when the sources change. At the same time, we
want to commit the generated files to our repository. We expect this
files to be human readable. Users our our libraries should read
generated bindings in their language, rather then the generating code in
Zig.

This is achieved by generating the bindings in build.zig, with a twist:
if a CI env variable is set, instead of overwriting the files, they are
checked for freshness.

If the generator generates a single file, it should write it to stdout.

If it generates multiple files, it should generate them to the output
directory, which is specified as the sole CLI argument.

Implementation-wise, `Generated` struct handles all the details.

### DIFF
--- a/src/clients/dotnet/dotnet_bindings.zig
+++ b/src/clients/dotnet/dotnet_bindings.zig
@@ -1,13 +1,10 @@
 const std = @import("std");
-const assert = std.debug.assert;
+const vsr = @import("vsr");
 
-// TODO: Move this back to src/clients/dotnet when there's a better solution for main_pkg_path=src/
-const vsr = @import("vsr.zig");
+const assert = std.debug.assert;
 const stdx = vsr.stdx;
 const tb = vsr.tigerbeetle;
 const tb_client = vsr.tb_client;
-
-const output_file = "src/clients/dotnet/TigerBeetle/Bindings.cs";
 
 const TypeMapping = struct {
     name: []const u8,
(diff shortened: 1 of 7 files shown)


src/clients/dotnet/dotnet_bindings.zig
```diff
@@ -515,23 +512,5 @@ pub fn main() !void {
     var buffer = std.ArrayList(u8).init(allocator);
     try generate_bindings(&buffer);
 
-    try std.fs.cwd().writeFile(.{ .sub_path = output_file, .data = buffer.items });
-}
-
-const testing = std.testing;
-
-test "bindings dotnet" {
```
kprotty, line 523: Should we no run the bindings through `zig build test`? Makes sense given the CI already calls `zig build clients:X` which runs them.

IIRC, them being imported to `unit_tests.zig` was the issue the `vsr` import couldn't work. So it's nice that it does now.

matklad: Yeah, that's the major pivot: instead of veryfing freshness through _tests_, we now special-case CI, and just fail the build there if the files are not fresh. The relevant part is here: 

https://github.com/tigerbeetle/tigerbeetle/pull/2193/files#diff-f87bb3596894756629bc39d595fb18d479dc4edf168d93a911cadcb060f10fccR985

And yeah, that's exactly what allows us to get rid of the tests, add a proper vsr module dependency, and move the files back into `client/$lang` folders!

kprotty (approved):

PR #2195, opened 2024-08-08
Update currency-exchange.md

onyno:
Swap the accounts A1 and L1 for the spread example (to match the prior example) with the addition of the fee transfer


matklad:
Good catch, thanks!

matklad (approved):

PR #2198, opened 2024-08-09
multiversioning: don't timeout_start() twice

cb22:
If `statx` returns an error, it would set some state, call `timeout_start()` and then call `handle_error()`. Except, `handle_error()` itself calls `timeout_start()` which leads to an assertion failure.

Practically, this means that upgrading 0.15.4 by deleting the binary and replacing it was broken; overwriting in place for upgrades still works (or anything that wouldn't cause the stat to fail).


matklad (approved):
Yeah, that's the kind of thing that makes `.tick`-style API for timeouts easier to use!

PR #2199, opened 2024-08-09
devhub: bump run number

cb22:
Remembering to keep this in sync with the minimum requirement in release.zig is a bit janky. Bump it manually for now, to get main green again.


matklad:
Yeah, I think that the way we map run ids to release numbers is now my top annoyance! Thinking about what we could do better here…

matklad (approved):

PR #2200, opened 2024-08-09
LSM: Don't call `iterator.empty()` of undefined iterator

sentientwaffle:
Don't call `iterator.empty()` of undefined iterator.

`scan_tree.zig`'s `index_block_callback` was resetting `iterator` to `undefined`:
```
            self.state = .{
                .iterating_values = .{
                    .key_exclusive_next = self.state.loading_index.key_exclusive_next,
                    .iterator = undefined,
                },
            };
```

Then when `range_found` is `null`, it winds up panicking at this (apparently unrelated) assert, when it calls `empty` and then eventually `next_value_block` on the undefined iterator:

```
        pub fn next_value_block(it: *TableValueIterator, callback: Callback) void {
            assert(it.callback == null);
```

Aside: My ideal solution here would be to make `iterator` itself optional, but `@fieldParentPtr` doesn't support that afaict.

Seed: ./zig/zig build -Drelease fuzz -- lsm_tree 7115158452533090
Seed: ./zig/zig build -Drelease fuzz -- lsm_tree 4216256162519329136

### DIFF
--- a/src/lsm/scan_tree.zig
+++ b/src/lsm/scan_tree.zig
@@ -257,14 +257,14 @@ pub fn ScanTreeType(
                     .fetching => {
                         assert(level.state == .iterating_manifest or
                             level.state == .loading_index or
-                            level.state == .iterating_values);
+                            level.state == .iterating);
 
                         if (level.state == .iterating_manifest) level.move_next();
                         self.state.buffering.pending_count += 1;
                         level.fetch();
                     },
                     .buffered => {
-                        assert(level.state == .iterating_values);
+                        assert(level.state == .iterating);
                         assert(state_before == .needs_data);
                     },
                     .finished => {
@@ -572,9 +572,12 @@ fn ScanTreeLevelType(comptime ScanTree: type, comptime Storage: type) type {
                 checksum: u128,
                 read: Grid.Read = undefined,
             },
-            iterating_values: struct {
+            iterating: struct {
                 key_exclusive_next: Key,
-                iterator: TableValueIterator,
+                values: union(enum) {
+                    none,
+                    iterator: TableValueIterator,
+                },
             },
             finished: struct {
                 next_tick: Grid.NextTick = undefined,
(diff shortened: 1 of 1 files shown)


batiati (commented):
> Aside: My ideal solution here would be to make iterator itself optional, but @fieldParentPtr doesn't support that afaict.

How about a `union(enum)`?

```zig
state: union(enum) {
    idle,
    iterating: *TableValueIterator
},
```

sentientwaffle:
> How about a union(enum)?

I somehow missed that zig supported `@fieldParentPtr` of tagged unions!

src/lsm/scan_tree.zig
```diff
@@ -572,9 +572,12 @@ fn ScanTreeLevelType(comptime ScanTree: type, comptime Storage: type) type {
                 checksum: u128,
                 read: Grid.Read = undefined,
             },
-            iterating_values: struct {
+            iterating: struct {
```
batiati, line 575: Now we've dropped `_values`, we could also rename `iterating_manifest`.

:thinking:  "iterating manifest" isn't accurate for what we are doing, maybe `loading_manifest`, `moving_manifest` , or just `loading`.

src/lsm/scan_tree.zig
```diff
@@ -619,10 +622,10 @@ fn ScanTreeLevelType(comptime ScanTree: type, comptime Storage: type) type {
                         .{ .cache_read = true, .cache_write = true },
                     );
                 },
-                .iterating_values => |*iterating_values| {
+                .iterating => |*iterating| {
                     assert(self.values == .fetching);
-                    assert(!iterating_values.iterator.empty());
-                    iterating_values.iterator.next_value_block(value_block_callback);
+                    assert(!iterating.values.iterator.empty());
+                    iterating.values.iterator.next_value_block(value_block_callback);
```
batiati, line 628:
```suggestion
                    assert(iterating.* == .values);
                    assert(!iterating.values.iterator.empty());
                    iterating.values.iterator.next_value_block(value_block_callback);
```

matklad:
I am going to merge this as is to get this into a release! Our general policy is that if something happens, we skip a release week, but this seems rather low-risk to just get in!

Sorry for not noticing the VOPR failure when merging the original PR! The new `count` (which shows the total number of fuzzings) and `freshness` (which, for canaries, now show the most recent run) tell how well a PR was fuzzed, but of course one needs to actually look at them :facepalm:

matklad (approved):

PR #2201, opened 2024-08-12
cfo: log start and stop times

matklad:
CFO runs in a loop, so noting times is useful to gauge if a particular run is  stuck or not.

### DIFF
--- a/src/stdx.zig
+++ b/src/stdx.zig
@@ -796,3 +796,56 @@ test fmt_int_size_bin_exact {
         fmt_int_size_bin_exact(std.math.maxInt(u64) - 1023),
     });
 }
+
+// DateTime in UTC, intended primarily for logging.
+//
+// NB: this is a pure function of a timestamp. To convert timestamp to UTC, no knowledge of
+// timezones or leap seconds is necessary.
+pub const DateTimeUTC = struct {
+    year: u16,
+    month: u8,
+    day: u8,
+    hour: u8,
+    minute: u8,
+    second: u8,
+
+    pub fn now() DateTimeUTC {
+        const timestamp_seconds = std.time.timestamp();
+        assert(timestamp_seconds > 0);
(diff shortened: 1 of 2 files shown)


src/stdx.zig
```diff
@@ -796,3 +796,56 @@ test fmt_int_size_bin_exact {
...
+    second: u8,
+
+    pub fn now() DateTimeUTC {
+        const timestamp_seconds = std.time.timestamp();
+        assert(timestamp_seconds > 0);
+        return DateTimeUTC.from_timestamp(@intCast(timestamp_seconds));
+    }
+
+    pub fn from_timestamp(timestamp: u64) DateTimeUTC {
+        var ts = std.time.epoch.EpochSeconds{ .secs = timestamp };
```
cb22:
```suggestion
        var epoch_seconds = std.time.epoch.EpochSeconds{ .secs = timestamp };
```

maybe

src/stdx.zig
```diff
@@ -796,3 +796,56 @@ test fmt_int_size_bin_exact {
...
+        const year_day = ts.getEpochDay().calculateYearDay();
+        const month_day = year_day.calculateMonthDay();
```
cb22: I know what these are doing, but the naming messes with my head!

src/stdx.zig
matklad: Right, and doesn't need to be a var

src/stdx.zig
matklad: Yeah, one of the more confusing modules in the standard library.  That's why I wrapped that all into `DateTimeUTC` to have a sane API :rofl: The result seems correct to me:

```
14:53:51|~/p/tb/work|matklad/cfo-date✓
λ ./zig/zig build scripts -- cfo
info: start 2024-08-12 13:54:32 UTC
...
λ date --utc
Mon 12 Aug 2024 13:54:35 UTC
```

cb22 (approved):

PR #2206, opened 2024-08-12
Composite key refactor

batiati:
- Remove Zig workaround.
- Standardize `tombstone_bit` declaration with other sites.
- Replace `Self` with the type name.
- Add assertions.

### DIFF
--- a/src/lsm/composite_key.zig
+++ b/src/lsm/composite_key.zig
@@ -6,23 +6,21 @@ const stdx = @import("../stdx.zig");
 
 pub fn CompositeKeyType(comptime Field: type) type {
     // The type if zeroed padding is needed.
-    const pad = switch (Field) {
-        u128 => @as(u64, 0),
-        // [0]u8 as zero-sized-type workaround for https://github.com/ziglang/zig/issues/16394.
-        u64 => [0]u8{},
+    const Pad = switch (Field) {
+        u128 => u64,
+        u64 => u0,
         else => @compileError("invalid Field for CompositeKey: " ++ @typeName(Field)),
     };
-    const Pad = @TypeOf(pad);
 
     return extern struct {
-        const Self = @This();
+        const CompositeKey = @This();
 
         pub const sentinel_key: Key = key_from_value(&.{
             .field = math.maxInt(Field),
             .timestamp = math.maxInt(u64),
         });
 
-        const tombstone_bit: u64 = 1 << 63;
+        const tombstone_bit: u64 = 1 << (64 - 1);
 
         // u128 may be aligned to 8 instead of the expected 16.
         const field_bitsize_alignment = @divExact(@bitSizeOf(Field), 8);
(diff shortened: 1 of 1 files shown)


src/lsm/composite_key.zig
```diff
@@ -35,29 +33,31 @@ pub fn CompositeKeyType(comptime Field: type) type {
...
+            assert(@sizeOf(CompositeKey) == @sizeOf(Field) * 2);
+            assert(@sizeOf(CompositeKey) == @sizeOf(Key));
+            assert(@alignOf(CompositeKey) >= @alignOf(Field));
+            assert(@alignOf(CompositeKey) == field_bitsize_alignment);
+            assert(stdx.no_padding(CompositeKey));
         }
 
-        pub inline fn key_from_value(value: *const Self) Key {
+        pub inline fn key_from_value(value: *const CompositeKey) Key {
+            assert(value.padding == 0);
```
matklad: My gut feeling is to guard these asserts with `if constants.verify`: they are going to be in the innermost loop possible, and feels like they _could_ affect codegen adversely. Eg, if you have a loop, the compiler might be able to vectorize it. But if there's a possibility of an early exit from the loop, vectorization might be incorrect (as the code won't actually get to the end of the loop), so compiler might be forced to pick a scalar implementation.

But maybe this assert doesn't make optimization situation worse? We are already running with overflow checks, so an extra `noreturn` possibility _shouldn't_ make the code slower? Hard to tell though! 

I guess my decision tree is this:

* if we have a set of benchmarks that can detect realatively fine-grained performance effects, then we should run those benches. 
* if not, I'd lean towards guarding by `if constants.verify`. 
* Perhaps it would also be interesting to look at the actual optimized wasm of say, binary search, to understand how different it will be with tombstones?

matklad (commented):
Looks awesome overall, but I have some misgivings about assert in the data-plannest part of our codebase...

src/lsm/composite_key.zig
batiati: > * if we have a set of benchmarks that can detect realatively fine-grained performance effects, then we should run those benches.

The closest thing to a micro-benchmark is the `lsm_segmented_array` fuzzer, that heavily stresses the CompositeKey.
I could not detect any measurable difference with asserts enabled. 

> * if not, I'd lean towards guarding by `if constants.verify`.

I agree.

> * Perhaps it would also be interesting to look at the actual optimized wasm of say, binary search, to understand how different it will be with tombstones?

I didn't get it. What relation do tombstones have with the assert on `CompositeKey.padding == 0`?

matklad: >I didn't get it. What relation do tombstones have with the assert on CompositeKey.padding == 0?

Sorry, I think it was a data race between me writing this comment, and me concurently continuing reading the code, where we _do_ have a pre-existing assert in `tombstone_from_key`.

What I meant to say is that looking at the asm forthe inner loop of the binary search over composite keys with and without the assert could be interesting

matklad (approved):

PR #2208, opened 2024-08-13
devhub: work around run number requirements

cb22:
(the expectation is this code will change / go away in the near future, before we hit ~300 more releases :smile:)


matklad (approved):

matklad:
I have a horrible dj-inspired idea to make our changelog the source of truth for release numbers. So, release.zig would go and parse the changelog, and would use the latest release number from it.

So, bumping release number _would_ become a toil, but it would be the part of overall changelog writing (and our `scripts changelog` would take care of that).

PR #2209, opened 2024-08-13
devhub: run-number is somewhat bounded because it's cast into a u8

cb22:


matklad (approved):

matklad:
Ok, this places an upper bound on when this needs to be fixed properly :rofl:

PR #2211, opened 2024-08-13
vsr: ignore messages related to the old state sync protocol

matklad:
The current release (0.15.5) uses both protocols simultaneously. In the next release, after this commit, we will stop sending and processing messages related to the old protocol. However, we will still be _ignoring_ (as opposed to asserting the absence of) old messages.

In the release after that, deprecated message commands would be fully removed, so that the replica would panic if it receives one.


matklad:
To be merged ~~next~~ this week!

matklad:
VOPR green at 12584 runs!

sentientwaffle (approved):

PR #2212, opened 2024-08-13
vsr: remove old state sync dead code

matklad:
Missed these in #1951!

### DIFF
--- a/src/vsr/replica.zig
+++ b/src/vsr/replica.zig
@@ -8948,57 +8948,6 @@ pub fn ReplicaType(
             self.sync_dispatch(.awaiting_checkpoint);
         }
 
-        fn sync_requesting_checkpoint_callback(
-            self: *Self,
-            checkpoint_state: *const vsr.CheckpointState,
-        ) void {
-            assert(!self.solo());
-            assert(self.syncing == .requesting_checkpoint);
-            assert(!self.superblock.updating(.checkpoint));
-            assert(self.commit_stage == .idle);
-            assert(self.commit_prepare == null);
-            assert(self.sync_tables == null);
-            assert(self.grid.read_global_queue.empty());
-            assert(self.grid.write_queue.empty());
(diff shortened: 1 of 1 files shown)


src/vsr/replica.zig
```diff
@@ -879,9 +879,10 @@ pub fn ReplicaType(
             if (self.superblock.working.vsr_state.sync_op_max > 0) {
                 maybe(self.client_replies.writing.count() > 0);
                 for (0..constants.clients_max) |entry_slot| {
-                    if (self.client_sessions.entries_free.isSet(entry_slot)) {
-                        assert(!self.client_replies.faulty.isSet(entry_slot));
-                    } else {
+                    const slot_faulty = self.client_replies.faulty.isSet(entry_slot);
+                    const slot_free = self.client_sessions.entries_free.isSet(entry_slot);
+                    assert(!slot_faulty);
```
matklad, line 884: Added this assert to catch the missing unset code. Before this commit, the code looks like it says that it actually `maybe(slot_faulty)` here, but it seems like that's not true?

We enter here after open or after state sync, and, in both cases, the faulty bits are reset. And the sessions are reset as well, so nothing should be able to _set_ this bit sneakily, even if we are reading some of the replies asynchronously?

src/vsr/replica.zig
```diff
@@ -9021,6 +8971,10 @@ pub fn ReplicaType(
             self.grid.free_set_checkpoint.reset();
             self.client_sessions_checkpoint.reset();
             self.client_sessions.reset();
+            // Faulty bits will be set in sync_content().
+            while (self.client_replies.faulty.findFirstSet()) |slot| {
+                self.client_replies.faulty.unset(slot);
+            }
```
matklad, line 8977: This is from sync_requesting_checkpoint_callback! And this _should not_ be dead code, we should clear faulty status of client replies!

sentientwaffle (approved):

PR #2213, opened 2024-08-13
multiversioning: fix aarch64 macos build

cb22:
At first glance, you'd think this would have made the 0.15.5 release broken on macOS - but it turns out not!

0.15.5 pulled in the current release for 0.15.4 as a past release. That doesn't rely on offsets within a pack, so it worked.

0.15.3 was pulled from the pack. This would be affected, however the 0.15.3 embedded had one special quality: it was the universal binary itself for both platforms! This meant it was identical in size, offset and checksum, so using the x86_64 offsets and sizes and checksums for arm worked.

### DIFF
--- a/src/multiversioning.zig
+++ b/src/multiversioning.zig
@@ -941,7 +941,7 @@ pub const Multiversion = struct {
             MultiversionHeader,
         )];
         const header = try MultiversionHeader.init_from_bytes(source_buffer_header);
-        var header_inactive_platform: ?MultiversionHeader = undefined;
+        var header_inactive_platform: ?MultiversionHeader = null;
 
         // MachO's checksum_binary_without_header works slightly differently since there are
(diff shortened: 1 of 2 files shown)


cb22:
I would like to additionally assert the inactive header value checksums in `print_information` which is used by the CLI verification of the final built binaries - since currently it's a blind spot - but I don't think it's right to add `*_inactive` to the multiversioning struct itself. Maybe a `exe_path_arch` in the same vein as `exe_path_format`.

Verifying the release on macos might be good too.

matklad:
>I would like to additionally assert the inactive header value checksums in print_information which is used by the CLI verification of the final built binaries - since currently it's a blind spot

It seems like it might be better to just _run_ verification on both flavors of mac, so that you _only_ verify the host arch, but we actually make sure to verify on all relevant hosts?

matklad (approved):

PR #2214, opened 2024-08-13
build: shorten one giant function

matklad:
Our build.zig is mostly just one giant function that does everything. This is hard to read. In particular, it is somewhat hard to navigate to any particular place inside this function.

To make it easier to navigate, split it into a bunch of functions, where each funciton is responsible for a small number of steps. Pass in all the dependencies and parametres explicitly: while verbose, and labor-prone to write the frist time, it seems like this should be easier to maintain long term.

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -6,6 +6,9 @@ const Mode = std.builtin.Mode;
 
 const config = @import("./src/config.zig");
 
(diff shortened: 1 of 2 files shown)


batiati (approved):
Awesome, indeed it will be much easier to maintain. :heart:

PR #2215, opened 2024-08-13
vsr: refactor replica <-> multiversion API

matklad:
Slight refactors to bring multiversion in line with pre-existing code patterns. Namely:

* use `anyopaque` for release_execute context, so that replica knows absolutely nothing about `Multiversion`, no even its type
* remove `Multiversion`'s usage of `io.timeout` in favor of `fn tick`


matklad:
One thing I am feeling uneasy about here: I haven't tested this code! I don't think we even have any non manual way to test this. Which I think means I need to teach `zig build` to do multiversions! And then write an integration test for this!

matklad:
One more thing I wanted to do here is remove the aliasing of `releases_bundled` in favor of explicitly calling `replica.set_releases` or some such, but I've run out of steam in the middle!

src/multiversioning.zig
```diff
@@ -852,14 +854,13 @@ pub const Multiversion = struct {
...
         } else {
             self.stage = .init;
-            self.timeout_start();
         }
 
         self.timeout_statx_previous = .{ .previous = self.timeout_statx };
     }
 
     fn binary_open(self: *Multiversion) void {
-        assert(self.stage == .init or self.stage == .ready or self.stage == .err);
```
matklad, line 862: It see like we are at init at all of the call-sites?

src/multiversioning.zig
```diff
@@ -748,13 +754,11 @@ pub const Multiversion = struct {
             allocator.free(std.mem.span(self.args_envp.args[0].?));
             allocator.free(self.args_envp.args);
         }
-
-        self.timeout_start_enabled = false;
```
cb22, line 752: Do we need to `self.timeout.stop()`?

matklad: We don't, but I've `self.* = undefined` for good measure to make sure this isn't used after deinit by mistake.

cb22:
> One thing I am feeling uneasy about here: I haven't tested this code! I don't think we even have any non manual way to test this. Which I think means I need to teach zig build to do multiversions! And then write an integration test for this!

I've tested it manually, FWIW, and it all seems to work as expected (the timeout, but I didn't test that the on_timeout handling if something is in progress is correct, but it looks OK to me). But agreed re integration tests!

cb22 (approved):
https://tenor.com/view/perfectly-balanced-infinity-war-thanos-balance-all-things-should-be-balanced-gif-16273067

PR #2216, opened 2024-08-13
vopr: fix false positive when getting to recovering_head after version upgrade

matklad:
We have special code to disable storage faults when restarting a replica after a crash in the vopr. We also have similar code when restarting replica during upgrades but it doesn't quiet handle all special cases (in particular, it doesn't repair the WAL). Extract this special code into a special function to make sure that more of the logic is shared!

In particular, in this seed we have a cluster of two, where one replica gets `.recovering_head` in the release upgrade path

Seed: ./zig/zig build -Drelease vopr -- 11029999072009206202

### DIFF
--- a/src/vopr.zig
+++ b/src/vopr.zig
@@ -1061,28 +1061,9 @@ pub const Simulator = struct {
             }
         }
 
-        var header_prepare_view_mismatch: bool = false;
+        var wal_fixed: bool = true;
         if (!fault) {
-            // The journal writes redundant headers of faulty ops as zeroes to ensure
-            // that they remain faulty after a crash/recover. Since that fault cannot
-            // be disabled by `storage.faulty`, we must manually repair it here to
-            // ensure a cluster cannot become stuck in status=recovering_head.
-            // See recover_slots() for more detail.
-            const headers_offset = vsr.Zone.wal_headers.offset(0);
-            const headers_size = vsr.Zone.wal_headers.size().?;
-            const headers_bytes = replica_storage.memory[headers_offset..][0..headers_size];
-            for (
-                mem.bytesAsSlice(vsr.Header.Prepare, headers_bytes),
-                replica_storage.wal_prepares(),
-            ) |*wal_header, *wal_prepare| {
-                if (wal_header.checksum == 0) {
-                    wal_header.* = wal_prepare.header;
-                } else {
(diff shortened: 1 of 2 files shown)


src/vopr.zig
```diff
@@ -1104,13 +1085,9 @@ pub const Simulator = struct {
         ) catch unreachable;
 
         if (replica.status == .recovering_head) {
-            // Even with faults disabled, a replica that was syncing before it crashed
-            // (or just recently finished syncing before it crashed) may wind up in
-            // status=recovering_head.
-            assert(fault or
-                replica.op < replica.op_checkpoint() or
-                replica.log_view < replica.superblock.working.vsr_state.sync_view or
```
matklad, line 1112: Another artifact of the old state sync (or rather, my initial attempt at fixing it)

matklad:
Kinda unhappy about this fix to be honest! This gets the seed going, and I think it _is_ an improvement, but I think it's a matter of time until VOPR finds a seed that works-around our repair attempts here because it _is_ possible that wal is un-repairable.

I am wondering if I should be scrapping this whole "try not to fault" logic here, and instead do a "core_missing_head" thing?

matklad:
Requesting review mostly to brainstorm the proper solution here!

sentientwaffle:
:thinking: My main issue is that this extra repair code shouldn't run during the `replica_test`s. (Which for me hints that `Cluster` is the wrong place for it).

> I am wondering if I should be scrapping this whole "try not to fault" logic here, and instead do a "core_missing_head" thing?

Yeah, I'm starting to lean that way too. My only worry is that it could end up "hiding" a real liveness bug if it was too broad -- i.e. if over time it began to cover scenarios that we weren't actually aware of --  but that seems manageable by making the verification strict.

matklad:
I sort of convinced myself that the right fix here is to go in the opposite direction, and to “just” get rid of recovering head. That is, use that idea that replicas would _explocitly_ send nacks for the entire journal in a DVC. So that a recovering head replica is just a normal replica that simply dosnt nack and prepares past the head.

I worry that with our fault disabling logic we already might be masking a real liveness issue, as vopr seems to eager with repairing replicas if there’s a chance of one more recovering head

matklad:
Discussed this during walk&talk, decided to let it be for the moment and revisit once this bites us again!

PR #2217, opened 2024-08-14
vopr: test that we always survive torn writes only scenario

matklad:
I have a hunch that the vopr might be overeager with disabling faults around recovering heads, lets check that we can survive torn writes only!


chaitanyabhandari:
Closing this since https://github.com/tigerbeetle/tigerbeetle/pull/2345 got merged!

PR #2218, opened 2024-08-14
AOF: make aof record a flag

cb22:
Partly supersedes #2167.

Removes the AOF record code from the build system, and makes it a runtime option instead: `--experimental --aof`.

Additionally, only replay certain state machine operations. These are controlled by `AOF.replay_message`, and are needed in the case of replaying an AOF that has a large number of lookups that otherwise significantly slow down importing.

~Proper support for expiring pending transfers is still TBC and TODO.~

### DIFF
--- a/src/tigerbeetle/main.zig
+++ b/src/tigerbeetle/main.zig
@@ -26,8 +26,7 @@ const MessagePool = vsr.message_pool.MessagePool;
 const StateMachine = vsr.state_machine.StateMachineType(Storage, constants.state_machine_config);
 const Grid = vsr.GridType(Storage);
 
-const AOFType = if (constants.aof_record) AOF else void;
-const Replica = vsr.ReplicaType(StateMachine, MessageBus, Storage, Time, AOFType);
+const Replica = vsr.ReplicaType(StateMachine, MessageBus, Storage, Time, AOF);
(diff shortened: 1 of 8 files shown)


src/tigerbeetle/main.zig
```diff
@@ -260,13 +259,12 @@ const Command = struct {
...
-        var aof: AOFType = undefined;
-        if (constants.aof_record) {
+        var aof: ?AOF = if (args.aof) blk: {
             const aof_path = try std.fmt.allocPrint(allocator, "{s}.aof", .{args.path});
             defer allocator.free(aof_path);
 
-            aof = try AOF.from_absolute_path(aof_path);
-        }
+            break :blk try AOF.from_absolute_path(aof_path);
+        } else null;
```
matklad:
```suggestion
        } else null;
        defer if (aof != null) aof.?.deinit();
```

cb22: Good catch!

cb22: Actually maybe not :laughing: .

The AOF doesn't have a `deinit()`. It does have `close()` which closes the file, which is now called. (_should_ that be deinit()? :thinking:)

matklad: Close actually make more sense!

matklad (approved):

PR #2219, opened 2024-08-14
Remove double dash from version flag

ippsav:


sentientwaffle (approved):
Good catch; thanks!

ippsav:
@sentientwaffle seem to be that there is no longer a step called `simulator_run` as mentioned in `HACKING.md` i guess that was replaced at some point maybe `./tigerbeetle benchmark` is the way to do it now ? not sure

sentientwaffle:
Ah, thanks! That should be `./tigerbeetle vopr` now -- I'll update it.

PR #2220, opened 2024-08-14
StateMachine: Allow zero-amount transfers

sentientwaffle:
### Overview

Zero-amount transfers were originally forbidden or used as sentinels (depending on flags).
We have since encountered a couple of scenarios where zero-amount transfers would be idiomatic or convenient.

As of this PR, zero-amount transfers are permitted, and `amount=0` is no longer used as a sentinel value.

### Upgrading

The release which includes this PR should be tagged `0.16.0`.

This is a *breaking API change* which is gated in the state machine by the *client's* release.
Developers who continue using a `0.15.x` client can continue with the old behavior (until we eventually bump the `client_release_min`, forcing them to upgrade).

My plan is that we will eventually bump `client_release_min` (to `0.16.0`) as part of `0.17.0`. At that time we can also remove the legacy logic from the state machine.

#### Single-Phase transfers

- *Old behavior*: Creating a transfer with `amount = 0` returns the [`amount_must_not_be_zero`](https://docs.tigerbeetle.com/reference/requests/create_transfers#amount_must_not_be_zero) error.

- *New behavior*: Creating a transfer with `amount = 0` returns `ok` and creates the transfer (assuming there are no other errors). (The account balances do not change.)

#### Balancing transfers

- *Old behavior*: Creating a balancing transfer with `amount = 0` is shorthand for creating a transfer with `amount = AMOUNT_MAX` (i.e. `amount = 2 ** 128 - 1`). That is, it transfers as much money as possible between the accounts.

- *New behavior*: Creating a balancing transfer with `amount = 0` creates the transfer with `amount = 0` -- the transfer is created, but it transfers no money.

#### Post-pending transfers

- *Old behavior*: Creating a post-pending transfer with `amount = 0` posts the entire pending amount.

- *New behavior*: Creating a post-pending transfer with `amount = 0` posts zero, *voiding* the entire pending amount.

#### Void-pending transfers

- *Not changed*: Creating a void-pending transfer with `amount = 0` is still a sentinel for "void the entire pending amount".

### Motivation

- *Single-phase transfers*: (Allowing zero-amount single-phase transfers is just to be uniform with the other two changes.)
- *Balancing transfers*: When a balancing transfer is part of a linked chain, allowing successful zero-amount transfers prevents surprising failures when the accounts are already net-zero-balanced.
- *Post-pending transfers*: Applications may create the posting transfer with an amount that is the result of a computation -- if that result is zero, then posting zero of the pending balance (rather than all of it) is less likely to result in a subtle bug.


src/clients/java/java_bindings.zig
```diff
@@ -762,48 +772,48 @@ fn emit_u128_batch_accessors(
...
+        \\        putUInt128(at(Struct.{[property]s}), {[param_name]s});
+        \\    }}
+        \\
+        \\
+    , .{
+        .visibility = if (is_private or is_read_only) "" else "public ",
+        .property = to_case(field.name, .pascal),
+        .param_name = to_case(field.name, .camel),
+    });
+
```
sentientwaffle: This diff looks more complicated than it actually is due to indentation -- it just pulls the `set[property](byte[])` implementation out of the `else` branch.

src/clients/java/src/main/java/com/tigerbeetle/AccountBalanceBatch.java
```diff
@@ -75,6 +75,17 @@ void setDebitsPending(final BigInteger debitsPending) {
...
+    /**
+     * @param debitsPending an array of 16 bytes representing the 128-bit value.
+     * @throws IllegalArgumentException if {@code debitsPending} is not 16 bytes long.
+     * @throws IllegalStateException if not at a {@link #isValidPosition valid position}.
+     * @throws IllegalStateException if a {@link #isReadOnly() read-only} batch.
+     * @see <a href="https://docs.tigerbeetle.com/reference/account-balances#debits_pending">debits_pending</a>
+     */
+    void setDebitsPending(final byte[] debitsPending) {
+        putUInt128(at(Struct.DebitsPending), debitsPending);
+    }
```
sentientwaffle: I don't think we actually need these implementations, but not sure if it is worth specifically omitting them either.

src/testing/state_machine.zig
```diff
@@ -225,24 +225,32 @@ pub fn StateMachineType(
...
+                    value.add(std.mem.asBytes(&operation));
+                    value.add(std.mem.asBytes(&client_release));
+                    value.add(input);
+
                     state_machine.forest.grooves.things.insert(&.{
                         .timestamp = timestamp,
                         .id = op,
-                        .value = @as(u64, @truncate(vsr.checksum(input))),
+                        .value = @as(u64, @truncate(value.checksum())),
                     });
```
sentientwaffle, line 254: By encoding these fields  (`client`, `op`, etc) into the LSM state, we implicitly check (via the statechecker) that they are all consistent across the cluster.

src/clients/dotnet/dotnet_bindings.zig
```diff
@@ -263,12 +270,13 @@ fn emit_struct(
         \\{{
         \\    public const int SIZE = {};
         \\
-        \\
+        \\{s}
```
batiati, line 272: Let's add a new line here just to separate constants from field declarations.

```suggestion
        \\{s}
        \\
```

src/clients/java/java_bindings.zig
```diff
@@ -99,12 +100,20 @@ const type_mappings = .{
...
-    } },
+    .{
+        tb.Transfer, TypeMapping{
+            .name = "TransferBatch",
+            .private_fields = &.{"reserved"},
+            .readonly_fields = &.{"timestamp"},
+            .docs_link = "reference/transfer#",
+            .constants =
+            \\    public static byte[] getAmountMax() {
+            \\        return UInt128.asBytes(Long.MIN_VALUE, Long.MIN_VALUE);
```
batiati: Add a comment here that we're dealing with signed integers, that's why "MIN_VALUE" is actually what we want.

src/clients/java/java_bindings.zig
```diff
@@ -99,12 +100,20 @@ const type_mappings = .{
...
-        .docs_link = "reference/transfer#",
-    } },
+    .{
+        tb.Transfer, TypeMapping{
+            .name = "TransferBatch",
+            .private_fields = &.{"reserved"},
+            .readonly_fields = &.{"timestamp"},
+            .docs_link = "reference/transfer#",
+            .constants =
+            \\    public static byte[] getAmountMax() {
```
batiati: Suggestion: use `BigInteger` instead of an array of bytes.

```java
public static final BigInteger MAX_AMOUNT = UInt128.asBigInteger(Long.MIN_VALUE, Long.MIN_VALUE);
```

Advantages:

We can remove the getter method `getAmountMax()` (which allocates a new instance each time). Unlike arrays, `BigInteger` is immutable and could be safely declared a constant.

We already expose the API accepting `BigInteger` for fields that deal with amounts (there is no need to introduce the overload with `byte[]`).

src/clients/java/src/main/java/com/tigerbeetle/AccountBalanceBatch.java
batiati: Those private setters exist solely for testing (i.e. setting something to assert if the public getter is correct).
However, they decrease the test coverage ratio if they're not used. :wink: 

_Related to the comment about making MAX_AMOUNT a BigInteger_

src/clients/java/java_bindings.zig
```diff
@@ -393,7 +402,7 @@ fn emit_batch(
         \\{[big_integer_import]s}
         \\
         \\{[visibility]s}final class {[name]s} extends Batch {{
-        \\
+        \\{[constants]s}
```
batiati, line 415: Add a new line:

```suggestion
        \\
        \\{[constants]s}
```

src/state_machine.zig
```diff
@@ -2471,6 +2554,21 @@ pub fn StateMachineType(
     };
 }
 
+// TODO(client_release_min): When client_release_min is bumped, remove this function and the
+// legacy code it gates.
```
batiati: > When client_release_min is bumped, remove this function and the legacy code it gates

Can we add an `assert` to ensure that we won't forget to remove this function?

```zig
fn forbid_zero_amounts(client_release: vsr.Release) bool {
    const release_min_inclusive = vsr.Release.from(.{ .major = 0, .minor = 15, .patch = 3 });
    const release_max_exclusive = vsr.Release.from(.{ .major = 0, .minor = 16, .patch = 0 });
    
    assert(client_release.value < release_max_exclusive.value);
    return client_release.value >= release_min_inclusive.value;
}
```

src/state_machine.zig
```diff
@@ -2471,6 +2554,21 @@ pub fn StateMachineType(
...
+//
+// Specifically, when forbid_zero_amounts() is true:
+// - Zero-amount transfers are forbidden (`amount_must_not_be_zero`).
+// - Post-pending-transfer uses amount=0 as a sentinel for "post full amount".
+// - Balancing transfers use amount=0 as a sentinel for `maxInt(u128)`.
+fn forbid_zero_amounts(client_release: vsr.Release) bool {
```
batiati: Suggestion:
I think we could build a pattern for future cases and scope this kind of function into a namespace (or at least some common prefix).

Example:
`compatibility.forbid_zero_amounts(...)`
`legacy.allow_zero_amounts(...)`

docs/coding/two-phase-transfers.md
```diff
@@ -31,12 +31,22 @@ A post-pending transfer, denoted by
 pending transfer to "post", transferring some or all of the pending transfer's reserved amount to
 its destination.
 
+- If the posted [`amount`](../reference/transfer.md#amount) is greater than or equal to the pending
+  transfer's amount, the full pending transfer's amount is posted.
```
batiati, line 35: Explain that if the amount is greater than the amount of the pending transfer, the value recorded is set to the original amount.

src/clients/node/src/index.ts
```diff
@@ -73,6 +73,8 @@ export type Event = Account | Transfer | AccountID | TransferID | AccountFilter
 export type Result = CreateAccountsError | CreateTransfersError | Account | Transfer | AccountBalance
 export type ResultCallback = (error: Error | null, results: Result[] | null) => void
 
+export const amount_max: bigint = 340282366920938463463374607431768211455n // 2**128 - 1
```
batiati:
```suggestion
export const amount_max: bigint = (2n ** 128n) - 1n;
```

src/clients/java/java_bindings.zig
sentientwaffle: :+1: I assume `MAX_AMOUNT` is the idiomatic name in java -- for client APIs, do we favor tigerstyle naming conventions (`AMOUNT_MAX`) or language conventions?

batiati: Yeah, let's use TigerStyle whenever it does not violate Java's idiom.
So we can do TigerStyle's msb naming convention with Java's upper case: `AMOUNT_MAX`.

src/state_machine.zig
sentientwaffle: Good idea! Though it would actually be something like:

```
fn forbid_zero_amounts(client_release: vsr.Release) bool {
    const release_min_inclusive = vsr.Release.from(.{ .major = 0, .minor = 15, .patch = 3 });
    const release_max_exclusive = vsr.Release.from(.{ .major = 0, .minor = 16, .patch = 0 });
    const release_max_transition = vsr.Release.from(.{ .major = 0, .minor = 17, .patch = 0 });

    assert(client_release.value < release_max_transition.value);
    return client_release.value >= release_min_inclusive.value and
        client_release.value < release_max_exclusive.value;
}
```

batiati: Can we comptime this assert somehow? :thinking:

src/state_machine.zig
sentientwaffle: I don't think it needs to be a namespace -- these functions would only ever live in `state_machine.zig`.
Though, if we end up adding more than one we can reconsider.

batiati (approved):

PR #2221, opened 2024-08-14
vsr: syncing replicas shouldn't falsely contribute to checkpoint durability

chaitanyabhandari:
### Problem

Currently, replicas that have freshly synced to a new checkpoint but not finished syncing the tables for that checkpoint freely send prepare_ok for all new prepares that they append to their WAL (up to op_prepare_max). However, in doing so, they may  be falsely and prematurely contributing to the durability of that checkpoint, since they haven't yet finished making all the grid changes from that checkpoint durable. This could lead to liveness issues in the presence of storage faults.

For instance, consider this series of steps in a 3-replica cluster:

* R0, R1, R2 are at op_checkpoint=0
* R1 and R2 move to op_checkpoint=19
* R2 moves to op_checkpoint=39. R1 necessarily appends all prepares up to that as well, but doesn't actually commit anything (perhaps its disk is very slow) and crashes
* One block of op_checkpoint=39 gets corrupted on R2 (the single corruption)
* R0 syncs to op_checkpoint=39
* R0 together with R2 move all the way up to op_checkpoint=59. They are able to do so because the corrupted block isn't actually necessary for that strings of prepares. 
* R1 restarts, and notices that the cluster is at op_checkpoint=59 and syncs to that
* But now that corrupted block becomes permanently lost, as R1 no longer has a chance to repair it.

### Solution

To avoid falsely contributing to the durability of the current checkpoint, replicas syncing table blocks should not send prepare_oks for prepares past the current checkpoint's prepare_max. However, ops past prepare_max + pipeline_prepare_queue_max can be safely prepare_ok'd. This is because if the primary prepared these ops, prepare_max + 1 is committed and the checkpoint is durable on a commit quorum! Concretely, if a replica syncs to checkpoint Cn, it withholds prepares between ops `[Cn + trigger + pipeline, Cn + trigger + 2 * pipeline]`.

Withholding prepares between [Cn + trigger + pipeline, Cn + trigger + 2 * pipeline]) fixes the aforementioned scenario by ensuring R1 and R2 get to op_checkpoint=39 but not beyond. As a result, when R0 restarts, it *won't* state sync; instead, it would do WAL repair and just re-create the corrupted block.

### DIFF
--- a/src/vsr/replica.zig
+++ b/src/vsr/replica.zig
@@ -7229,6 +7229,7 @@ pub fn ReplicaType(
             assert(header.replica == self.primary_index(header.view));
             assert(header.view <= self.view);
             assert(header.op <= self.op or header.view < self.view);
+            maybe(!self.sync_content_done());
 
             if (self.status != .normal) {
                 log.debug("{}: send_prepare_ok: not sending ({})", .{ self.replica, self.status });
@@ -7246,6 +7247,18 @@ pub fn ReplicaType(
                 log.debug("{}: send_prepare_ok: not sending (sync_status={s})", .{
                     self.replica,
                     @tagName(self.syncing),
+                });
+                return;
+            }
+            // To avoid falsely contributing to the durability of the current checkpoint, replicas
+            // syncing table blocks should not send prepare_oks for prepares past (and including)
+            // the next checkpoint's trigger.
+            if (!self.sync_content_done() and self.op_checkpoint() > 0 and
+                header.op >= vsr.Checkpoint.prepare_max_for_checkpoint(self.op_checkpoint()).?)
+            {
+                log.debug("{}: send_prepare_ok: not sending (sync_content_done={})", .{
+                    self.replica,
+                    self.sync_content_done(),
                 });
                 return;
             }
--- a/src/vsr/replica.zig
+++ b/src/vsr/replica.zig
@@ -7250,11 +7250,17 @@ pub fn ReplicaType(
                 });
                 return;
             }
+
             // To avoid falsely contributing to the durability of the current checkpoint, replicas
-            // syncing table blocks should not send prepare_oks for prepares past (and including)
-            // the next checkpoint's trigger.
-            if (!self.sync_content_done() and self.op_checkpoint() > 0 and
-                header.op >= vsr.Checkpoint.prepare_max_for_checkpoint(self.op_checkpoint()).?)
+            // syncing table blocks should not send prepare_oks for prepares past the current
+            // checkpoint's prepare_max. However, ops past prepare_max + pipeline_prepare_queue_max
+            // can be safely prepare_ok'd. This is because if the primary prepared these ops,
+            // prepare_max + 1 is committed and the checkpoint is durable on a commit quorum!
+            const prepare_max_current_checkpoint =
+                vsr.Checkpoint.prepare_max_for_checkpoint(self.op_checkpoint());
+            if (!self.sync_content_done() and header.op > prepare_max_current_checkpoint.? and
+                (header.op <= prepare_max_current_checkpoint.? +
+                constants.pipeline_prepare_queue_max))
             {
                 log.debug("{}: send_prepare_ok: not sending (sync_content_done={})", .{
                     self.replica,
--- a/src/vsr/replica.zig
+++ b/src/vsr/replica.zig
@@ -9097,6 +9097,14 @@ pub fn ReplicaType(
             if (self.superblock.staging.vsr_state.sync_op_max == 0) {
                 return true;
             } else {
+                // Trailers/manifest haven't yet been synced.
+                if (self.client_sessions_checkpoint.size_transferred <
+                    self.superblock.working.client_sessions_reference().trailer_size or
+                    !self.grid.free_set.opened or !self.state_machine_opened)
+                {
+                    return false;
+                }
+
                 for (0..constants.clients_max) |entry_slot| {
                     if (self.client_sessions.entries_free.isSet(entry_slot)) continue;
 
--- a/src/vsr/replica.zig
+++ b/src/vsr/replica.zig
@@ -7253,20 +7253,19 @@ pub fn ReplicaType(
 
             // To avoid falsely contributing to the durability of the current checkpoint, replicas
             // syncing table blocks should not send prepare_oks for prepares past the current
-            // checkpoint's prepare_max. However, ops past prepare_max + pipeline_prepare_queue_max
-            // can be safely prepare_ok'd. This is because if the primary prepared these ops,
-            // prepare_max + 1 is committed and the checkpoint is durable on a commit quorum!
-            const prepare_max_current_checkpoint =
-                vsr.Checkpoint.prepare_max_for_checkpoint(self.op_checkpoint());
-            if (!self.sync_content_done() and header.op > prepare_max_current_checkpoint.? and
-                (header.op <= prepare_max_current_checkpoint.? +
-                constants.pipeline_prepare_queue_max))
-            {
-                log.debug("{}: send_prepare_ok: not sending (sync_content_done={})", .{
-                    self.replica,
-                    self.sync_content_done(),
-                });
-                return;
+            // checkpoint's prepare_max.
+            if (!self.sync_content_done()) {
+                const prepare_max_current_checkpoint =
+                    vsr.Checkpoint.prepare_max_for_checkpoint(self.op_checkpoint()).?;
+                if (self.commit_max <= prepare_max_current_checkpoint and
+                    header.op > prepare_max_current_checkpoint)
+                {
+                    log.debug("{}: send_prepare_ok: not sending (sync_content_done={})", .{
+                        self.replica,
+                        self.sync_content_done(),
+                    });
+                    return;
+                }
             }
 
             assert(self.status == .normal);
@@ -9098,12 +9097,7 @@ pub fn ReplicaType(
                 return true;
             } else {
                 // Trailers/manifest haven't yet been synced.
-                if (self.client_sessions_checkpoint.size_transferred <
-                    self.superblock.working.client_sessions_reference().trailer_size or
-                    !self.grid.free_set.opened or !self.state_machine_opened)
-                {
-                    return false;
-                }
+                if (!self.state_machine_opened) return false;
 
                 for (0..constants.clients_max) |entry_slot| {
                     if (self.client_sessions.entries_free.isSet(entry_slot)) continue;


matklad:
Would it be feasible to write a replica test here?

chaitanyabhandari:
> Would it be feasible to write a replica test here?

@matklad ah yes, would certainly be useful to have one! I'll work on it. :D

src/vsr/replica.zig
```diff
@@ -7245,6 +7246,16 @@ pub fn ReplicaType(
                 });
                 return;
             }
+            // To avoid falsely contributing to the durability of the current checkpoint, replicas
+            // syncing table blocks should not send prepare_oks for prepares past (and including)
+            // the next checkpoint's trigger.
+            if (!self.sync_content_done() and header.op >= self.op_checkpoint_next_trigger()) {
```
matklad: Hm, I need to revisit this once I stop paniking over my currently deeply in progress slides, but one subtelty which is I am not sure if taker care of here is that the current checkpoint changes.

Imagine a scenario where:

* we sync to checkpoint X
* we avoid externalizing prepare_ok which signal that our _current_ checkpoint is durable
* non-the-less, we continue commiting normaly, and move to checkpoint `X + checkpoint_inteval`
* this _doesnt_ mean that our state sync finished! Even though we have enough of `X` to commit up to the next checkpoint, there might be some data which simply wasn't acceded yet. 
* and yet, because we moved to the next checkpoint, _now_ will be sending prepare oks that we were witholding previously

My original thinking here is that, while the sync isn't done, we'll have to withold prepare oks around _every_ checkpoint trigger. 

EDIT: but maybe this works? If we progressed to the next checkpoint, then the primary thinks that the current checkpoint is fine...

---

Another surprising thing is that we use `op_checkpoint_next_trigger` here. I'd expect us using

```
                const prepare_max_for_current_checkpoint =
                    vsr.Checkpoint.prepare_max_for_checkpoint(self.op_checkpoint()).?;
```

to mirror the logic we have in op_repair_min

But overall super low-confidence on both point: I'll need to actually think about this to be sure, and I am avoiding that, otherwise I won't do my slides on time :stuck_out_tongue:

chaitanyabhandari: I will ponder on these while you work on your slides! 🤔 🕵🏼

src/vsr/replica.zig
```diff
@@ -2136,15 +2136,40 @@ pub fn ReplicaType(
...
+                if (self.op_checkpoint_next() == op_checkpoint_max and
+                    self.commit_min < self.op_checkpoint_next_trigger() and
+                    self.op >= self.op_checkpoint_next_trigger())
+                {
+                    const commit_max = @max(
+                        self.commit_max,
+                        DVCQuorum.commit_max(self.do_view_change_from_all_replicas),
+                    );
+
+                    self.commit_max = commit_max;
```
matklad: Let's use `advance_commit_max` for this purpose --- a function makes it easier to audit all places where we advance commit max.

matklad: Hm, I am not sure this would be correct: you can only advance commit max if you are certain about your head, and we aren't currently. 

That is, I think what could happen here is that we advance commit_max, but the op correponding to commit_max in our log is wrong (the cluster truncated it, but we didn't as we missed one of the previous views)

matklad: The actual problem here seems real enough though! 

I wonder if we could perhaps solve it by relaxing the condition of forfeiting? It seems like we should only forefeit if it is true that **both**:

* cluster is ahead of us by a checkpoint
* the checkponit is confirmed

matklad: >Hm, I am not sure this would be correct: 

Though the fact that vopr doesn't find any issue here perhaps means that I am wrong about this one?

chaitanyabhandari: Ah, advancing commit_max without being certain about whether the op we have in our log is correct or not is something that I didn't consider! Great point, thanks a lot for pointing that out.

The correct solution would perhaps involve also invoking `primary_set_log_from_do_view_change_messages`, since we would then keep our headers, head op & commit max up-to-date across view changes... Basically, we're only invoking commit_journal here, when what we *should* be doing is everything (updating log_view, invoking primary_set_log_from_do_view_change_messages and then repair which in turn should perform advance the commit pipeline)?

chaitanyabhandari: > Though the fact that vopr doesn't find any issue here perhaps means that I am wrong about this one?

This has changed since, we have a failing seed! So I definitely need to think deeper about this change. :)

Also you talk about the checkpoint being confirmed, are you referring to it being durable on a commit-quorum of replicas?

matklad: > Also you talk about the checkpoint being confirmed, are you referring to it being durable on a commit-quorum of replicas?

Yep! Which I think is “commit max is beyond prepare max for prev checkpoint”

chaitanyabhandari: Ah, I _think_ I've got it! Thinking back to your earlier comment https://github.com/tigerbeetle/tigerbeetle/pull/2221#discussion_r1717477390, I realized that we require altering _both_ of these solutions to effectively and correctly address this issue... 

Syncing replicas not sending a prepare_ok for any ops past op_prepare_max (if tables haven't yet been synced) ensures that a syncing replica doesn't falsely contribute to the durability of a checkpoint. Specifically, this ensures that no replica would be able to overwrite ops from the previous checkpoint. This dovetails nicely with our discussion in this comment! If a replica is lagging by only a checkpoint and it finds that commit_max < op_prepare_max, it doesn't need to forfeit the view change. It can simpy use WAL repair to catch up since no ops from the lagging replica's checkpoint would have been overwritten yet (since we don't have durability for the *next* checkpoint).

chaitanyabhandari: Although, I believe the latter change *would* require some changes to [primary_set_log_from_do_view_change_messages](https://github.com/tigerbeetle/tigerbeetle/blob/880d406c08c1a7354a50d536cca2cacfa4e04e1b/src/vsr/replica.zig#L8149), which currently assumes that the header of the canonical DVC is within our op_prepare_max (`assert(header_head.op <= self.op_prepare_max());`) and replaces the replica's head op with the canonical DVC's head op (`self.set_op_and_commit_max(header_head.op, commit_max, @src());`).  Specifically, [primary_set_log_from_do_view_change_messages](https://github.com/tigerbeetle/tigerbeetle/blob/880d406c08c1a7354a50d536cca2cacfa4e04e1b/src/vsr/replica.zig#L8149) would likely have to change the head op to `@min(self.op_prepare_max(), header_head.op)`. Also, we would have to add checks around using the DVC headers to replace our own headers, such that only headers up to op_prepare_max are replaced.

I still haven't thought this through properly though, so this may not be true. 😄

chaitanyabhandari: Ah yes, quick experiment with running the VOPR after making the above changes *minus* the changes to [primary_set_log_from_do_view_change_messages](https://github.com/tigerbeetle/tigerbeetle/blob/880d406c08c1a7354a50d536cca2cacfa4e04e1b/src/vsr/replica.zig#L8149) that I'd anticipated showed that we would indeed need to make those changes:
```C
[debug] (replica): 1: on_do_view_change: view=25 quorum received
[debug] (replica): 1: on_do_view_change: dvc: replica=0 log_view=24 op=71 commit_min=67 checkpoint=59
[debug] (replica): 1: on_do_view_change: dvc: header: replica=0 op=71 checksum=208545351406523309574206797536401977373 nack=false present=true type=valid
[debug] (replica): 1: on_do_view_change: dvc: header: replica=0 op=70 checksum=60171286467781141794056692787268822268 nack=false present=true type=valid
[debug] (replica): 1: on_do_view_change: dvc: header: replica=0 op=69 checksum=168193058532343774027948307381785896824 nack=false present=true type=valid
[debug] (replica): 1: on_do_view_change: dvc: header: replica=0 op=68 checksum=223364609765652245498886553981331736644 nack=false present=true type=valid
[debug] (replica): 1: on_do_view_change: dvc: header: replica=0 op=67 checksum=299943675278850709487137762353369328066 nack=false present=true type=valid
[debug] (replica): 1: on_do_view_change: dvc: replica=1 log_view=23 op=67 commit_min=62 checkpoint=39
[debug] (replica): 1: on_do_view_change: dvc: header: replica=1 op=67 checksum=299943675278850709487137762353369328066 nack=false present=true type=valid
[debug] (replica): 1: on_do_view_change: dvc: header: replica=1 op=66 checksum=325996494377482587983529677958291707278 nack=false present=true type=valid
thread 19040384 panic: reached unreachable code
/Users/chaitanyabhandari/Desktop/tigerbeetle/tigerbeetle/zig/lib/std/debug.zig:412:14: 0x100ba170f in assert (vopr)
             ^
/Users/chaitanyabhandari/Desktop/tigerbeetle/tigerbeetle/src/vsr/replica.zig:8207:19: 0x100ca0a43 in primary_set_log_from_do_view_change_messages (vopr)
            assert(header_head.op <= self.op_prepare_max());
                  ^
```

matklad: > which currently assumes that the header of the canonical DVC is within our op_prepare_max (assert(header_head.op <= self.op_prepare_max());)

Ohh, this is fascinating! I’ll double check this tomorrow, but so far it indeed looks like it could be the case that, during view change, there is a checkpoint ahead of yours, and even the head op comes from this next checkpoint, and yet you might need to become a primary, because the checkpoint is borked, and only wal repair would work.

One question though: is this forfeiting change actually repeated to “not sending prepare ok”s one? It looks like these might be two unrelated issues? If that is the case, than perhaps we  can split this into two prs.

Intuitively it seems that “don’t falsely confirm checkpoint as durable if you are still state syncing” is much easier than “don’t forfeit if checkpoint might not be durable”.

chaitanyabhandari: Agreed, these are two separate issues and makes sense to tackle them separately!

Also, there's a failing liveness issue on main right now. I will investigate it tomorrow, hopefully it is related to the latter issue, i.e. “don’t forfeit if checkpoint might not be durable”. 🤞🏼

src/vsr/replica.zig
```diff
@@ -7252,6 +7253,18 @@ pub fn ReplicaType(
                 });
                 return;
             }
+            // To avoid falsely contributing to the durability of the current checkpoint, replicas
+            // syncing table blocks should not send prepare_oks for prepares past (and including)
+            // the next checkpoint's trigger.
+            if (!self.sync_content_done() and self.op_checkpoint() > 0 and
```
matklad: Why do we need `self.op_checkpoint() > 0`? If `!self.sync_content_done()` holds than we _must_ be on some non-zero checkpoint, no?

src/vsr/replica.zig
```diff
@@ -7252,6 +7253,18 @@ pub fn ReplicaType(
...
+                header.op >= vsr.Checkpoint.prepare_max_for_checkpoint(self.op_checkpoint()).?)
```
matklad: Hm, this feels overly strict to me. There are only a pipeline worth of preapres for which `header.op <= vsr.Checkpoint.prepare_max_for_checkpoint(self.op_checkpoint()).?` is true. 

So, I think practically this condition is equivalent to "don't ever ok prepares if you are syncing content": there are about 1000 prepares in a checkpoint, and we are allowed to ack only 8 of them. 

I _think_ this could be relaxed quite a bit. Assume that we are syncing content, and we are at checkpoint `C`. Now, anything between `C` and `C + trigger + pipeline` is safe to ack. 

`C + trigger` is trivially safe, as those things are even committed. 

`trigger <-> pipeline`is ok because that's before `prepare_max(C)` and doesn't signal that we have reached `C` (in other words, _even_ if we were on `C - checkpoint_interval`, we could still acked those prepares).

Now, something like `C + trigger + pipeline + 1`  is unsafe to ack -- the primary can look at that, and decide that we have checkpoint `C`, and decide that `C` is durable on the cluster, and then evict some log message, compormisign durability.

But I think anything beyond `C + trigger + pipeline + pipeline` is safe to ack again? Because if we see something in that range, we know that `C + trigger + pipeline + 1` is already committed, and that means that the checkpoint `C` _is_ actually durable?

chaitanyabhandari: > anything beyond C + trigger + pipeline + pipeline is safe to ack again

Beautiful! Indeed this can be relaxed by a *LOT*, since if we see an op beyond C + trigger + pipeline + pipeline, it is guaranteed that (C + trigger + pipeline + 1) is committed. Updating this condition right away.

matklad: Would be great to double-check the logic with replica test ^^

chaitanyabhandari: Writing a replica test to reproduce this issue unveiled that this logic that we discussed is correct but doesn't address the issue at hand! I have pushed the replica test as part of https://github.com/tigerbeetle/tigerbeetle/pull/2221/commits/f7a82d22e3adab532ea4472890ed1fbcd3195a7f for you to peruse, although it is a little janky right now as one test that fails on main and passes on this branch right now (branched by a hardcoded flag). 

While it is correct to withhold prepares between _[C<sub>n</sub> + trigger + pipeline, C<sub>n</sub> + trigger + 2 * pipeline]_ if we're not sure about durability of C<sub>n</sub>, the main triggering condition in this issue is that the syncing replica allows the primary to move on to the *next* checkpoint. Specifically, with the help of the syncing replica R0, R2 advances it's checkpoint to C<sub>n+1</sub>. Consequently, R2 is not able to help R1 with a corruption in its checkpoint (since it has a newer version of that block). R0 isn't able to help R1 with the corruption since it hasn't finished syncing its tables.

I think it makes sense to revive the logic that we started off with, wherein a syncing replica withholds prepare_oks for prepares past the trigger of the _next_ checkpoint. I've implemented both of these changes together in https://github.com/tigerbeetle/tigerbeetle/pull/2221/commits/6b589c58607f27aac4e167fd59bb7a665b3d3b2a, what do you think?

matklad: Had to diagram this:

image

And I think what your are saying makes sense to me!

Basically, there are two availability risks if a syncing replica participates in consensus:

1. First, the primary may falsely believe that its current checkpoint is durable, which would lead to eviction of old prepares. Which might be incorrect, because if only the previous checkpoint is durably, the preapres are needed to run wal repair. 
2. Second, even if the primary correctly believes that its current checkpoint is durable, by advancing to the _next_ checkpoint it makes the now past checkponit non-durable. 

So we either wrongly reduce durability of prepares, or durability of checkpoints. 

And yet, I am not _quiet_ convinced that 2 is a real problem --- it seems like this scenario requires one too many faults? At least in the diagram, we actually have two faults --- a grid corruption and one recovering head replica. So perhaps this is a false postive in the vopr? Though, I think even if we somehow unstuck R2 from recovering head, the cluster would still be stuck? Even if R2 heals and becomes a priamry, it won't be able to help either R0 or R1 to repair a checkpoint?

It would be very useful to encode a clear-cut one-fault unavailability as a replcia test for the second scenario!

For the first scenario, one fault would look like this:

* R0, R1, R2 are at op_checkpoint=0
* R1 and R2 move to op_checkpoint=19
* R2 moves to op_checkpoint=39. R1 necessarily appends all prepares up to that as well, but doesn't actually commit anything (perhaps its disk is very slow) and crashes
* One block of op_checkpoint=39 gets corrupted on R2 (the single corruption)
* R0 syncs to op_checkpoint=39
* R0 together with R2 move all the way up to op_checkpoint=79. They are able to do so because the corrupted block isn't actually necessary for that strings of prepares. 
* R1 restarts, and notices that the cluster is at op_checkpoint=79 and syncs to that
* But now that corrupted block becomes permanently lost, as R1 no longer has a chance to repair it.

The first mitigation (withholding prepares between [Cn + trigger + pipeline, Cn + trigger + 2 * pipeline]) fixes this scenario I think. R1 and R2 would get to op_checkpoint=59, but not beyond. And R0, when it restarts, _wont_ state sync, so it'll do wal repair and would just re-create the corrupted block.

Ahhh, ok, now I see one fault scenario where the second mitigation, preventing advancing the checkpoint, is required! Basically, in the scenario above what kills us is a fault in the new, unconfirmed checkpoint. The scenario I am going to describe now faults a durable checkpoint:

* R0, R1, R2 are at op_checkpoint=0
* R1 and R2 move to op_checkpoint=19
* R1 crashes and gets a single corruption in its checkpoing
* R0 syncs to op_checkpoint=19
* R0 and R2 commit all the way up to `op_checkpoint=39`
* At this point, R2 evicts the block from op_checkpoint=19 from its grid
* R1 restarts, and, assuming the withholding prepares mitigation is in place, it _wont_ try to state sync, and would try to repair the wal, which won't work, because a grid block is missing in the cluster.

:thinking: hmmmm, wait, I think it could get worse? It seems we don't even need state sync here, and the same issue affects normal processing:

* R0 is permanently partitioned, it never participates in this scenario at all
* R1 and R2 move to op_checkpoint=19
* R1 and R2 append prepares to the log all the way up to op_prepare_max
* R2 moves to op_checkpoint=39, and commits compaction_ops prepares on top of it, evicting some block from the previous op_checkpoint=19
* This same block gets corrupted on R1
* R1 crashes and restarts, and now needs that block

matklad: In terms of this pr: let's land the frist part of witholding prepares, as the first scenario above is a convincing demonstartion why we need it.

For the second part, the _last_ scenario I feel shows that we'll need to do more work here

chaitanyabhandari: @matklad thank you for meticulously sketching out all those scenarios!

One question though, in the first scenario, wouldn't withholding prepare_oks between [Cn + trigger + pipeline, Cn + trigger + 2 * pipeline] keep R1 & R2 at op_checkpoint=39 (as opposed to op_checkpoint=59)? Since R1 isn't able to move from op_checkpoint=19 → 39, the cluster would only be able to commit up till op_prepare_max(19)=47, and then when R0 syncs to op_checkpoint=39, it withholds prepare_oks between [48, 51].

chaitanyabhandari: Also, the last scenario indeed seems like something that we would have to discuss more elaborately to come up with a general solution from the problem.

However @matklad, in the meantime, what are your thoughts about pushing the current solution (syncing replicas don't contribute prepare_oks for checkpoint advancement) as a temporary mitigation? It has been VOPR green for 12223 runs till now, and it does address the failing seed that this PR aims to solve (though it indeed doesn't solve the *more general* problem that can be encountered even without state sync).

matklad: >One question though, in the first scenario, wouldn't withholding prepare_oks between [Cn + trigger + pipeline, Cn + trigger + 2 * pipeline] keep R1 & R2 at op_checkpoint=39

Ah, yeah, you are right!

>what are your thoughts about pushing the current solution (syncing replicas don't contribute prepare_oks for checkpoint advancement) as a temporary mitigation?

I'd rather not to! As I learned from my failed attempts to fix the old state-sync protocol, incremental "this fixes this particular case" adjustments are not necessary a part of the final solution, and might actually create extra complexity when you'd have to back them out.

chaitanyabhandari: > might actually create extra complexity when you'd have to back them out.

Ack, that does make sense. Let's solve it once and for all, without any mitigations on the way. :)

I will do away with the "syncing replicas don't contribute prepare_oks for checkpoint advancement" logic then! Also will change the description for this PR and to reflect what it is _actually_ trying to solve, i.e. scenario 1 from https://github.com/tigerbeetle/tigerbeetle/pull/2221#discussion_r1723213288 (with small corrections).

chaitanyabhandari:
The motivation behind https://github.com/tigerbeetle/tigerbeetle/pull/2221/commits/4802bef59beb48b7b882cc24fceabc1ec1aa8fd8 is to make _sync_content_done_ account for scenarios where the trailer & manifest haven't yet been synced. 

This was discovered in a replica test wherein I dropped all incoming *.block* messages on a syncing replica. In this scenario, the syncing replica was not able to recover trailer & manifest logs (the callback for [grid.open](https://github.com/tigerbeetle/tigerbeetle/blob/56193862533052b686396f521298920812f75d9d/src/vsr/replica.zig#L9050) was never called because the free set couldn't be synced), but _sync_content_done_ erroneously returned *True*(the syncing replica hasn't even *begun* to sync its content yet).

src/vsr/replica.zig
```diff
@@ -9078,6 +9097,14 @@ pub fn ReplicaType(
             if (self.superblock.staging.vsr_state.sync_op_max == 0) {
                 return true;
             } else {
+                // Trailers/manifest haven't yet been synced.
+                if (self.client_sessions_checkpoint.size_transferred <
+                    self.superblock.working.client_sessions_reference().trailer_size or
+                    !self.grid.free_set.opened or !self.state_machine_opened)
```
matklad: It seems like this is _just_ `!self.state_machine_opened`? Unopened client sessions should imply unopened state machine I think

src/vsr/replica.zig
```diff
@@ -7250,6 +7251,24 @@ pub fn ReplicaType(
...
+            // To avoid falsely contributing to the durability of the current checkpoint, replicas
+            // syncing table blocks should not send prepare_oks for prepares past the current
+            // checkpoint's prepare_max. However, ops past prepare_max + pipeline_prepare_queue_max
+            // can be safely prepare_ok'd. This is because if the primary prepared these ops,
+            // prepare_max + 1 is committed and the checkpoint is durable on a commit quorum!
+            const prepare_max_current_checkpoint =
+                vsr.Checkpoint.prepare_max_for_checkpoint(self.op_checkpoint());
+            if (!self.sync_content_done() and header.op > prepare_max_current_checkpoint.? and
+                (header.op <= prepare_max_current_checkpoint.? +
+                constants.pipeline_prepare_queue_max))
```
matklad: This is ugly fomatting! Not a problem per se, but usually a sign that there's some out of the box simplification possible. 

The trivial improvement here would be to split the if into the principal and helper condition, and to push `.?` up:

```zig
            if (!self.sync_content_done()) {
                // To avoid falsely contributing to the durability of the current checkpoint,
                // replicas syncing table blocks should not send prepare_oks for prepares past the
                // current checkpoint's prepare_max. However, ops past prepare_max +
                // pipeline_prepare_queue_max can be safely prepare_ok'd. This is because if the
                // primary prepared these ops, prepare_max + 1 is committed and the checkpoint is
                // durable on a commit quorum!
                const prepare_max_current_checkpoint =
                    vsr.Checkpoint.prepare_max_for_checkpoint(self.op_checkpoint()).?;

                if (header.op > prepare_max_current_checkpoint and
                    (header.op <= prepare_max_current_checkpoint + constants.pipeline_prepare_queue_max))
                {
                    log.debug("{}: send_prepare_ok: not sending (sync_content_done={})", .{
                        self.replica,
                        self.sync_content_done(),
                    });
                    return;
                }
            }
```

This is more readable, but actually makes line length problem worse. 

If we look at the offending condition about the pipeline, it becomes clear that something else is wrong here --- the usage of pipeline length constant is somewhat arbitrarty here. Like, if you think about it, it becomes clear why this implies what you think it should imply, but it requires some indirect reasons. I think a more direct way to say this would be 

```zig
                if (header.op > prepare_max_current_checkpoint and
                    (self.commit_max <= prepare_max_current_checkpoint))
```

?

That way it's exactly the condition from op_repair_min:

```
     const prepare_max_for_current_checkpoint =
                    vsr.Checkpoint.prepare_max_for_checkpoint(self.op_checkpoint()).?;
                // After state sync, commit_max might lag behind checkpoint_op.
                maybe(self.commit_max < prepare_max_for_current_checkpoint);
                if (self.commit_max > prepare_max_for_current_checkpoint) {
           
```

src/vsr/replica.zig
chaitanyabhandari: Yes, just `!self.state_machine_opened` would suffice since these steps happen sequentially! That is precisely what I'd done before but I saw this comment and thought it could be a good idea to separate them out?
https://github.com/tigerbeetle/tigerbeetle/blob/56193862533052b686396f521298920812f75d9d/src/vsr/replica.zig#L829-L835

We could also take this up when we _actually_ start doing these concurrently. I'm okay doing either!

src/vsr/replica.zig
chaitanyabhandari: I _had_ considered just using `self.commit_max <= prepare_max_for_current_checkpoint` here (mirroring logic in op_repair_min), but the thought got lost somewhere in a sea of ideas. 🤦🏼 


I agree, this certainly isn't the most aesthetically pleasing way to do this. I will clean it up, thanks a lot for the suggestions! I like the first suggestion of separating the conditions into principal and helper. Indeed much more readable that way, and also makes it much clearer what we mean - _if the checkpoint isn't durable, don't prepare_ok prepares that may falsely indicate that it is durable_.

chaitanyabhandari: https://github.com/tigerbeetle/tigerbeetle/pull/2221/commits/f6fa8466f242f544725e4f8941de44811f1c91e1

src/vsr/replica.zig
chaitanyabhandari: Went with just simplifying it down to `!self.state_machine_opened`: https://github.com/tigerbeetle/tigerbeetle/pull/2221/commits/f6fa8466f242f544725e4f8941de44811f1c91e1

chaitanyabhandari:
It seems like the VOPR is omniscient! The issue that it found on my branch (`./zig/zig build -Drelease vopr -- 10842232175341922821` on https://github.com/tigerbeetle/tigerbeetle/commit/f6fa8466f242f544725e4f8941de44811f1c91e1) is similar to what we were discussing earlier today (re the last scenario in https://github.com/tigerbeetle/tigerbeetle/pull/2221#discussion_r1723213288).
```C
[info] (cluster): no liveness, final cluster state (requests_max=96 requests_replied=89):
[info] (cluster):  0   \ v    5126V 139/147/147C 117:147Jo  0/_0J! 117:148Wo <__0:__0> v1:3   223Ga  1G!   3G?
[info] (cluster):  1   /  v   5128V 119/131/147C 116:147Jo  0/_0J! 116:147Wo <__0:__0> v1:3   258Ga  1G!   2G?
[info] (cluster):  2   \   v  5128V  19/_26/_28C   1:_32Jo  0/_0J!   1:_32Wo <__0:__0> v1:3    51Ga  1G!   5G?
[err] (cluster): you can reproduce this failure with seed=10842232175341922821
```

Here's a summary of happened here:

1. R2 crashes
2. R0 & R1 commit ops till op=147
3. R0 moves to op_checkpoint=139
4. R1 gets a corruption in op_checkpoint=119 but isn't able to recover it since R0 has recreated that block in op_checkpoint=139. Therefore, R1 isn't able to commit up till the trigger for op_checkpoint=139 and remains stuck at op_checkpoint=119.
5. R0 crashes and restarts with a corruption in op_checkpoint=139. It isn't able to recover the block since R1 has a block from the previous checkpoint.
6. R0 and R1 remain stuck in view change.

matklad (approved):
Looks good to me now!

PR #2224, opened 2024-08-15
Ensure u128 consistency across languages.

batiati:
Ensure u128 consistency across languages.

- Fix Dotnet GUID representation that was not compatible with Java.
- Implement tests to assert that the same binary will be interpreted equally as UInt128, BigInteger, and GUID (when applicable) for all languages (except Node since it's limited to the native big integer type only).

Related: #1977

### DIFF
--- a/src/clients/dotnet/TigerBeetle.Tests/UInt128Tests.cs
+++ b/src/clients/dotnet/TigerBeetle.Tests/UInt128Tests.cs
@@ -11,6 +11,34 @@ namespace TigerBeetle.Tests;
 [TestClass]
 public class UInt128Tests
 {
+    /// <summary>
+    /// Consistency of U128 across Zig and the language clients.
+    /// It must be kept in sync with all platforms.
+    /// </summary>
+    [TestMethod]
+    public void ConsistencyTest() {
+        // Decimal representation:
+        ulong upper = 11647051514084770242;
+        ulong lower = 15119395263638463974;
+        var u128 =  new UInt128(upper, lower);
+        Assert.AreEqual("214850178493633095719753766415838275046", u128.ToString());
+
+        // Binary representation:
+        byte[] binary = new byte[] {
+            0xe6, 0xe5, 0xe4, 0xe3, 0xe2, 0xe1,
+            0xd2, 0xd1,
+            0xc2, 0xc1,
+            0xb2, 0xb1,
+            0xa4, 0xa3, 0xa2, 0xa1,
+        };
+        Assert.IsTrue(binary.SequenceEqual(u128.ToArray()));
+
+        // GUID representation:
+        var guid = Guid.Parse("a1a2a3a4-b1b2-c1c2-d1d2-e1e2e3e4e5e6");
+        Assert.AreEqual(guid, u128.ToGuid());
+        Assert.AreEqual(u128, guid.ToUInt128());
+    }
+
     [TestMethod]
     public void GuidConversion()
     {
--- a/src/clients/dotnet/TigerBeetle.Tests/UInt128Tests.cs
+++ b/src/clients/dotnet/TigerBeetle.Tests/UInt128Tests.cs
@@ -16,11 +16,12 @@ public class UInt128Tests
     /// It must be kept in sync with all platforms.
     /// </summary>
     [TestMethod]
-    public void ConsistencyTest() {
+    public void ConsistencyTest()
+    {
         // Decimal representation:
         ulong upper = 11647051514084770242;
         ulong lower = 15119395263638463974;
-        var u128 =  new UInt128(upper, lower);
+        var u128 = new UInt128(upper, lower);
         Assert.AreEqual("214850178493633095719753766415838275046", u128.ToString());
 
         // Binary representation:
--- a/src/clients/dotnet/TigerBeetle.Tests/UInt128Tests.cs
+++ b/src/clients/dotnet/TigerBeetle.Tests/UInt128Tests.cs
@@ -138,12 +138,19 @@ public class UInt128Tests
     [TestMethod]
     public void LittleEndian()
     {
-        var expected = new byte[16] { 86, 52, 18, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0 };
+        byte[] bytes_expected = new byte[16] { 86, 52, 18, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0 };
+        UInt128 decimal_expected = new UInt128(0, 0x123456);
+        BigInteger bigint_expected = BigInteger.Parse("123456", NumberStyles.HexNumber);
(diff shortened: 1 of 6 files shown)


sentientwaffle (approved):

src/clients/dotnet/TigerBeetle.Tests/UInt128Tests.cs
```diff
@@ -11,6 +11,35 @@ namespace TigerBeetle.Tests;
...
+    public void ConsistencyTest()
+    {
+        // Decimal representation:
+        ulong upper = 11647051514084770242;
+        ulong lower = 15119395263638463974;
+        var u128 = new UInt128(upper, lower);
+        Assert.AreEqual("214850178493633095719753766415838275046", u128.ToString());
+
+        // Binary representation:
+        byte[] binary = new byte[] {
```
jkulubya, line 28: @batiati These bytes are exactly what's stored on disk in tb, correct? And they have no meaning, i.e. they're an opaque identifier (client provides this sequence of bytes, client gets back account with matching ID)?

src/clients/dotnet/TigerBeetle.Tests/UInt128Tests.cs
```diff
@@ -11,6 +11,35 @@ namespace TigerBeetle.Tests;
...
+            0xe6, 0xe5, 0xe4, 0xe3, 0xe2, 0xe1,
+            0xd2, 0xd1,
+            0xc2, 0xc1,
+            0xb2, 0xb1,
+            0xa4, 0xa3, 0xa2, 0xa1,
+        };
+        Assert.IsTrue(binary.SequenceEqual(u128.ToArray()));
+
+        // GUID representation:
+        var guid = Guid.Parse("a1a2a3a4-b1b2-c1c2-d1d2-e1e2e3e4e5e6");
```
jkulubya, line 38: If the understanding above is correct, why should the expected string representation of a guid container be reversed here?

src/clients/dotnet/TigerBeetle/UInt128Extensions.cs
```diff
@@ -31,24 +32,34 @@ internal unsafe struct UnsafeU128
...
     {
-        unsafe
-        {
-            return new Guid(new ReadOnlySpan<byte>(&value, SIZE));
-        }
+        Span<byte> data = stackalloc byte[SIZE];
+        MemoryMarshal.Write(data, in value);
+
+        // The GUID layout is big endian.
+        // This is important to preserve the string representation.
```
jkulubya, line 39: It's unclear which string representation is referred to here but most languages represent a byte[16]/uuid as the same 16 bytes concat-ed together with no special processing (except c#, which triggered the original pr).

{0xe6, 0xe5, 0xe4, 0xe3, 0xe2, 0xe1, 0xd2, 0xd1, 0xc2, 0xc1, 0xb2, 0xb1, 0xa4, 0xa3, 0xa2, 0xa1} would equal e6e5e4e3-e2e1-d2d1-c2c1-b2b1a4a3a2a1.

The problem with the approach here is if a uuid comes in from elsewhere (e.g. we generate our own ulid like type, but little-endian to optimise for tigerbeetle), then it's going to be reversed here and break the optimisation. I think what I'm advocating for is to treat the dotnet guid type as dumb as possible.

I don't know enough Java to understand why the string representation of {0xe6, 0xe5, 0xe4, 0xe3, 0xe2, 0xe1, 0xd2, 0xd1, 0xc2, 0xc1, 0xb2, 0xb1, 0xa4, 0xa3, 0xa2, 0xa1} isn't e6e5e4e3-e2e1-d2d1-c2c1-b2b1a4a3a2a1, but I suspect that's in tb Java client code.

batiati:
Hey @jkulubya, great to hear from you!

The main goal of this PR is to assert that the memory/wire representation of the _little-endian unsigned 128-bit integer_ we use natively is consistent across platforms that do not provide a native 128-bit type:
- in Java, we deal with a pair of signed 64-bit integers -- or -- an array of 16 bytes.
- in Go, we use a struct containing a pair of unsigned 64-bit integers.
- in C#, we use the UInt128 type introduced in .NET 8.0. However, it's not a first-class citizen in C# since it lacks a literal representation and native p/invoke support.

Secondary goals:
- Assert that the decimal representation through BigInteger matches all platforms.
- Assert that the GUID/UUID string representation also matches Java and C#.

For Java, we are using the most undoubtedly possible API:
[`new UUID(long mostSigBits, long leastSigBits)`](https://docs.oracle.com/javase/8/docs/api/java/util/UUID.html#UUID-long-long-)

For C# we reproduce the same logic, consistently keeping the same representation when a GUID is persisted in one platform and restored in another.

We could argue that the same binary transformation also happens with BigIntegers, which are big-endian in C#, Java, and Go. The difference is that with BigIntegers, there are no questions about what unsigned decimal number they were supposed to represent.

We can conclude that what really matters are two aspects:
- a: the binary representation while it's a u128 or [16]byte.
- b: the string representation while it's a GUID or a BitInteger.
Any conversion between (a) and (b) must produce equivalent results, in any platform.

(...)

>I don't know enough Java to understand why the string representation of {0xe6, 0xe5, 0xe4, 0xe3, 0xe2, 0xe1, 0xd2, 0xd1, 0xc2, 0xc1, 0xb2, 0xb1, 0xa4, 0xa3, 0xa2, 0xa1} isn't e6e5e4e3-e2e1-d2d1-c2c1-b2b1a4a3a2a1, but I suspect that's in tb Java client code.

That said, we have two hypotheses here to explore:

1. The source of truth **is** the decimal `214850178493633095719753766415838275046`.
    The byte representation just happens to appear swapped when converting it to a GUID, because GUIDs are big 
    endian.
    That is fine, since we are not supposed to take the raw bytes and cast it as another type, and here's an example that reinforces this idea:
https://github.com/tigerbeetle/tigerbeetle/blob/0c429d854714659f95ed9c6a024ce7789446949c/src/clients/dotnet/TigerBeetle.Tests/UInt128Tests.cs#L139-L154

    Which also seems to be the case here:
https://github.com/microsoft/windows-rs/blob/f19edde93252381b7a1789bf856a3a67df23f6db/crates/tests/core/tests/guid.rs#L25-L31

or

2. In order to convert a U128 to UUID/GUID, we should not simply extract the `mostSigBits` and the `leastSigBits`, but convert it to big-endian before, so the byte order matches the expectation that "most languages represent a byte[16]/uuid as the same 16 bytes concat-ed together with no special processing".

**EDIT:**

As a follow-up draft, how it would look like if we adopted option 2:
https://github.com/tigerbeetle/tigerbeetle/compare/batiati-guid-big-endian

src/clients/dotnet/TigerBeetle/UInt128Extensions.cs
jkulubya: Or put another way, if the guid type was just a dumb container that gave you the same bytes you put in the same exact order you put them in, would you still need to reverse? If not, then using the bigEndian=true parameter is enough.

PR #2230, opened 2024-08-22
client: move batching from StateMachine to VSR

kprotty:
Currently, batching of requests with similar operations into the same message only occurs for `create_accounts`/`create_transers`. This is partially due to their returned results containing an `index` field which can be fixed up (through a StateMachine-specific type called `Demuxer`) and used to link each result to its corresponding request. Therefor other operations like `lookup_*` and queries aren't batched.

The goal here is to move the batching logic outside StateMachine and into the messaging/protocol layer so that it works with any operation. This initial version adds a 256-byte header at the start of message body which contains `[128]u16`. The first value contains the number of requests in the message (N). The next N contain the number of items in the message belonging to each request. For example, requests having 2, 7, and 4 items would have a batch header of `[N=3, 2, 7, 4, 0....]`.

A format which keeps the items together in the message body is needed for alignment + processing purposes; All batched items are prefetched by the StateMachine together, but they're committed separately so the response message body can fill in the appropriate item counts.

TODO: support a variable amount of headers. For request item counts of 1 (i.e. `[N, 1, 1, 1...]`) a message is limited to 128 as opposed to the higher `8191` or `batch_limit` as reported by vsr.register.

### DIFF
--- a/src/clients/c/tb_client.h
+++ b/src/clients/c/tb_client.h
@@ -255,9 +255,10 @@ typedef struct tb_packet_t {
     void* data;
     struct tb_packet_t* batch_next;
     struct tb_packet_t* batch_tail;
-    uint32_t batch_size;
-    uint8_t batch_allowed;
-    uint8_t reserved[7];
+    uint16_t batch_count_packets;
+    uint16_t batch_count_events;
+    uint16_t batch_count_results;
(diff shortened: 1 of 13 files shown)


cb22:
(superseded by #2617)

PR #2233, opened 2024-08-22
build: multiversion build v2

matklad:
This moves multiversion from release.zig to build.zig, so that you can
just build a multiversion binary normally, eg:

```
$ ./zig/zig build -Dmultiversion=0.15.6 -Dconfig-release=0.15.7 -Dconfig-release-client-min=0.15.6 -Dtarget=x86_64-linux
$ ./tigerbeetle multiversion ./tigerbeetle
multiversioning.exe_path=./tigerbeetle
multiversioning.absolute_exe_path=/home/matklad/p/tb/work/tigerbeetle
multiversioning.releases_bundled={ 0.15.3, 0.15.4, 0.15.5, 0.15.6, 0.15.7 }
...
```

You can pass `-Dmultiversion=latest` to fetch the latest release from
GitHub.

Note that `-Dtarget=aarch64-macos` will build a _universal_ macos image
(this _is_ confusing flag name, but I don't see a  simple non-ambigious
solution here).

As a bonus, multiversion build now should work on any host, and not on
x86_64 linux only.

Implementation wise, the bulk of work happens in
`build_multiversion.zig` program, which is invoked from build.zig as a
custom Run step. Why not encode all that logic direcctly into build.zig?

The main reason is adrewrk, the most famous Zig troll. His favorite way
of trolling Zig developers is by regularly breaking the API of build.zig:

https://github.com/ziglang/zig/pull/14498

The next planned trolling is to separate "configure" and "make" phases
of `build.zig` into physically separate processes, and remove support
for custom steps:

https://github.com/ziglang/zig/issues/20981

In this PR, we are going to troll adrewrk _back_ by fixing our code even
before it gets broken, and by making our custom step a Run step instead.

That being said, it I have split `gh release download` into a separate
step, so that is properly cached. That is, re-build multiversion binary
does not re-download the release.


src/build_multiversion.zig
```diff
@@ -0,0 +1,839 @@
...
+    llvm_objcopy: []const u8,
+    source: []const u8,
+    output: []const u8,
+}) !u128 {
+    // Copy the object using llvm-objcopy before taking our hash. This is to ensure we're
+    // round trip deterministic between adding and removing sections:
+    // `llvm-objcopy --add-section ... src dst_added` followed by
+    // `llvm-objcopy --remove-section ... dst_added src_back` means
+    // checksum(src) == checksum(src_back)
+    // Note: actually don't think this is needed, we could assert it?
```
cb22, line 408: To expand on this - this no-op copy was needed when I was initially using GNU objcopy, but I don't think it's needed for LLVM objcopy. We could instead actually add and remove a section here, and check the result is deterministic - which would assert the behaviour we actually want!

matklad: It seems like the result isn't actually deterministic?

https://github.com/tigerbeetle/tigerbeetle/commit/ec90ba43aea7425239d5c436eb94667d57614c49

I kept the existing behavior here

cb22: Hm - can you try with the dummy objcopy? Maybe it still is needed! (The end result should definitely be deterministic, since we verify that modified binary against the `current_checksum` which was taken before any other operations, but after the dummy objcopy). 

It could also be that the dummy objcopy was needed only for debug, or something along those lines.

matklad: Checked debug and release manually, in both cases the results looks like this: 

```
0468184187d6c475de4245d7bb05c23cd87cc54b20a165c55931e9b40dfc7665  tigerbeetle
fa4f97f9d9c62ed5f686e760232b2c6c7a5806a0f538bbe898668f9f0156527b  tigerbeetle-add-section
046bf770b824c0ec7faf6b8bdec03cb3c015311e2026650875787335ce905905  tigerbeetle-add-section-remove-section
046bf770b824c0ec7faf6b8bdec03cb3c015311e2026650875787335ce905905  tigerbeetle-noop
```

That is, no-op copy is equal to adding&removing the section, but is different from tigerbeetle we get directly out of the build.

src/build_multiversion.zig
```diff
@@ -0,0 +1,848 @@
...
+    tmp_path: []const u8,
+    target: Target,
+    arch: Target.Arch,
+    tigerbeetle_past: []const u8,
+    output: []const u8,
+}) !struct {
+    past_releases: MultiversionHeader.PastReleases,
+    unpacked: []const []const u8,
+} {
+    // TODO: parse_macho needlessly depends on builtin.target.cpu.arch
```
matklad, line 436: TODO fixed separately in https://github.com/tigerbeetle/tigerbeetle/pull/2235

build.zig
```diff
@@ -359,9 +365,109 @@ fn build_tigerbeetle(
...
 ) void {
+    const out_filename = if (options.target.result.os.tag == .windows)
+        "tigerbeetle.exe"
+    else
+        "tigerbeetle";
+
+    const tigerbeetle_bin = if (options.multiversion) |version| bin: {
+        assert(!options.emit_llvm_ir);
+        const build_mutliversion_exe = b.addExecutable(.{
+            .name = "build_mutliversion",
```
sentientwaffle: `%s/mutliversion/multiversion`

matklad: :heart: 

I would like in general to brainstorm some better name for the concept! `multiversion`  is somewhat hard to mentally parse for me, is long, and we use both `multiversion` and `multiversioning`...

src/build_multiversion.zig
```diff
@@ -0,0 +1,848 @@
...
+    tigerbeetle_current_x86_64: ?[]const u8 = null, // NB: Will be x86-64 on the CLI!
+    tigerbeetle_current_aarch64: ?[]const u8 = null,
+    tigerbeetle_past: []const u8,
+    output: []const u8,
+    tmp: []const u8,
+};
+
+// These are the options for cli_args.tigerbeetle_current. Ideally, they should be passed at
+// runtime, but passing them at comptime is more convenient.
+const vrs_options = @import("vsr_options");
```
sentientwaffle: s/vrs/vsr

build.zig
cb22: It's an oddity too, because I actually settled on calling everything else `release` internally. So `multirelease` would technically make more sense for the concept name, but it felt even worse...

build.zig
```diff
@@ -359,9 +365,109 @@ fn build_tigerbeetle(
...
+            "--tigerbeetle-past=",
+            download_release(b, version, options.target, options.mode),
+        );
+        build_multiversion.addArg(b.fmt(
+            "--tmp={s}",
+            .{b.cache_root.join(b.allocator, &.{"tmp"}) catch @panic("OOM")},
+        ));
+        break :bin build_multiversion.addPrefixedOutputFileArg("--output=", out_filename);
+    } else bin: {
+        const tigerbeetle_exe = build_tigerbeetle_executable(b, .{
```
batiati, line 383: Can we do the same thing at the first branch of this `if` and extract the logic into a function?

```zig
if (options.multiversion) |version| bin: {
  assert(!options.emit_llvm_ir);
  const build_multiversion_exe = build_tigerbeetle_multiversion_executable(b, .{ ... });
  break :bin build_multiversion_exe;
} else bin: {
  const tigerbeetle_exe = build_tigerbeetle_executable(b, .{ ... });
  // ...
  break :bin tigerbeetle_exe.getEmittedBin();
}
```

matklad: Excellent call, I forgot that we don't do large functions anymore :rofl:

batiati (approved):
It was the most cartoonish trolling-back move I've seen in a code base, like building a brick house to avoid being blown down by the Zig wolf!

PR #2234, opened 2024-08-23
Closing accounts

batiati:
### Closing accounts

- Introduces a new flag, `Account.flags.closed`, which will cause the account to reject any further transfers, except for voiding two-phase transfers that are still pending.

- Expired pending balances will be reversed automatically according to the `timeout` in closed accounts.

- Index on `Accounts.flags.closed`.

- The account flag can be set during creation or through a `Transfer` with `Transfer.flags.closing_debit_account` or `Transfer.flags.closing_credit_account` set (aka the "closing transfer").
 
- Closing transfers are required to be a two-phase pending transfer.

- The closed account can be re-opened by voiding or expiring the closing transfer.

- There's no "permanently closed" status, since the account does not allow posting any transfer, not even the pending closing transfer.


docs/reference/requests/create_transfers.md
```diff
@@ -442,6 +470,16 @@ automatic expiration.
...
 
+### `debit_account_closed`
+
+The transfer was not created. [`Transfer.debit_account_id`](../transfer.md#debit_account_id) must
+refer to an `Account` whose [`Account.flags.closed`](../account.md#flagsclosed) is not set.
+
+### `debit_account_closed`
+
+The transfer was not created. [`Transfer.credit_account_id`](../transfer.md#credit_account_id) must
+refer to an `Account` whose [`Account.flags.closed`](../account.md#flagsclosed) is not set.
```
```suggestion
The transfer was not created. [`Transfer.debit_account_id`](../transfer.md#debit_account_id) must
refer to an `Account` whose [`Account.flags.closed`](../account.md#flagsclosed) is not already set.

### `debit_account_closed`

The transfer was not created. [`Transfer.credit_account_id`](../transfer.md#credit_account_id) must
refer to an `Account` whose [`Account.flags.closed`](../account.md#flagsclosed) is not already set.
```

(...Just to be a little clearer about the order of events.)

docs/reference/transfer.md
sentientwaffle, line 1: Additionally, we should extend the `docs/recipes/close-account.md` with this flag.
And maybe add an extra example to it, with pending/post close? (Or that could be a separate recipe).

src/tigerbeetle.zig
```diff
@@ -298,6 +301,9 @@ pub const CreateTransferResult = enum(u32) {
     imported_event_timestamp_must_postdate_credit_account = 62,
     imported_event_timeout_must_be_zero = 63,
 
+    debit_account_closed = 64,
+    credit_account_closed = 65,
```
sentientwaffle: What do you think of changing the names to:

```suggestion
    debit_account_already_closed = 64,
    credit_account_already_closed = 65,
```


"Account closed" is ambiguous -- "closed" could be interpreted as an adjective or a verb:
1. Adjective: The current state of the debit account is that it is closed.
2. Verb: The account has just been closed.

src/state_machine.zig
```diff
@@ -2037,15 +2047,16 @@ pub fn StateMachineType(
...
+            assert(!cr_account_new.flags.closed);
+            if (t2.flags.closing_debit_account) dr_account_new.flags.closed = true;
+            if (t2.flags.closing_credit_account) cr_account_new.flags.closed = true;
 
+            if (!stdx.equal_bytes(Account, dr_account, &dr_account_new)) {
                 self.forest.grooves.accounts.update(.{ .old = dr_account, .new = &dr_account_new });
+            }
+            if (!stdx.equal_bytes(Account, cr_account, &cr_account_new)) {
                 self.forest.grooves.accounts.update(.{ .old = cr_account, .new = &cr_account_new });
             }
```
sentientwaffle: I'm undecided on whether checking `equal_bytes()` is the right approach here (likewise in `post_or_void_pending_transfer`) or if we should use a more granular condition like:

```suggestion
            if (amount > 0 or t2.flags.closing_debit_account) {
                self.forest.grooves.accounts.update(.{ .old = dr_account, .new = &dr_account_new });
            }
            if (amount > 0 or t2.flags.closing_credit_account) {
                self.forest.grooves.accounts.update(.{ .old = cr_account, .new = &cr_account_new });
            }
```

We assert `equal_bytes` in `groove.update()`'s implementation (gated by `constants.verify`), so by checking it in a different way here we get a bit of defense-in-depth. :thinking: 

What do you think?

src/state_machine.zig
```diff
@@ -2234,6 +2248,15 @@ pub fn StateMachineType(
...
             };
 
+            // The only movement allowed in a closed account is
+            // posting or voiding the closing transfer:
+            if (dr_account.flags.closed and !p.flags.closing_debit_account) {
+                return .debit_account_closed;
+            }
+            if (cr_account.flags.closed and !p.flags.closing_credit_account) {
+                return .credit_account_closed;
+            }
```
sentientwaffle: Suppose:

1. Create pending transfer X between accounts A and B.
2. Create (pending?) transfer Y between accounts A and B, closing A.

What is supposed to happen at this point if:
- X is posted?
- X is voided?
- X expires?

src/tigerbeetle.zig
```diff
@@ -133,7 +134,9 @@ pub const TransferFlags = packed struct(u16) {
     balancing_debit: bool = false,
     balancing_credit: bool = false,
     imported: bool = false,
-    padding: u9 = 0,
+    closing_debit_account: bool = false,
+    closing_credit_account: bool = false,
```
sentientwaffle: If we are including this in 0.16, we could reorder these to:
```
    balancing_debit: bool = false,
    balancing_credit: bool = false,
    closing_debit_account: bool = false,
    closing_credit_account: bool = false,
    imported: bool = false,
    padding: u9 = 0,
```

... which admittedly serves no real purpose except aesthetics.

sentientwaffle: Should the name be `closing_debit` to match `balancing_debit`? Or is the different intentional?

src/state_machine.zig
```diff
@@ -3563,32 +3617,32 @@ fn check(test_table: []const u8) !void {
 
 test "create_accounts" {
```
sentientwaffle, line 3653: TODO for me: add support for some kind of "bitset" to `src/testing/table.zig` so that adding a new flag isn't `O(test cases)` of updates.

src/state_machine.zig
```diff
@@ -4438,16 +4492,90 @@ test "imported events: linked chain" {
...
+        \\ account A1  0  0  0  0  _  _  _ _ L1 C1   _  _  _ _ _ _ _ _ ok
+        \\ account A2  0  0  0  0  _  _  _ _ L1 C1   _  _  _ _ _ _ _ _ ok
+        \\ account A3  0  0  0  0  _  _  _ _ L1 C1   _  _  _ _ _ _ _ _ ok
+        \\ commit create_accounts
+        \\
+        // Temporarily closing the debit account.
+        \\ transfer   T1 A1 A2   15   _  _   _  _    _ L1 C1   _   _   _   _   _   _  _  _   _   _ _ ok
+        \\ transfer   T2 A1 A2    0   _  _   _  _    0 L1 C1   _   PEN _   _   _   _  _  CDR _   _ _ ok
+        \\ transfer   T3 A1 A2    5   _  _   _  _    _ L1 C1   _   _   _   _   _   _  _  _   _   _ _ debit_account_closed
+        \\ transfer   T3 A2 A1    5   _  _   _  _    _ L1 C1   _   _   _   _   _   _  _  _   _   _ _ credit_account_closed
```
sentientwaffle: Add another two cases to also verify that it can't be re-closed while it is already closed.

src/tigerbeetle.zig
batiati: The prefix `closing_` is to match `balancing_`. I thought an extra suffix `_account` would align better with `{debit,credit}_account_id`.

Glad to reduce it to just `closing_{debit,credit}` if you think it's clear enough.

src/state_machine.zig
batiati: Ok, I'm also concerned by the "footguns" that can be caused by closing an account with balance (especially pending balance). But the orthogonality of primitives seems to outweigh this risk. (e.g. it's the application policy to balance the account before closing it, we provide both primitives).

>X is posted?
  X is voided?

A closed account will not accept any transfer except by posting/voiding the same transfer that set the closed flag. In this case, the pending balance cannot be posted or voided.

> X expires?

Expiration reverts the balance without a transfer, so it will succeed even for closed accounts.

(...)

Any thoughts on this?

First, I had the idea of forbidding an account from being permanently closed if it had any pending balance. Requiring the closing transfer to be "pending", and the user should re-open, void/post the pending balance and then close permanently. But it seemed too complicated to explain.

sentientwaffle:
Can you apply

```
diff --git a/src/state_machine.zig b/src/state_machine.zig
index 817a03ca3..35c7c5cb5 100644
--- a/src/state_machine.zig
+++ b/src/state_machine.zig
@@ -4525,6 +4525,7 @@ test "create_transfers: closing accounts" {
         // Temporarily closing the debit account.
         \\ transfer   T1 A1 A2   15   _  _   _  _    _ L1 C1   _   _   _   _   _   _  _  _   _   _ _ ok
         \\ transfer   T2 A1 A2    0   _  _   _  _    0 L1 C1   _   PEN _   _   _   _  _  CDR _   _ _ ok
+        \\ transfer   T2 A1 A2    0   _  _   _  _    0 L1 C1   _   PEN _   _   _   _  _  CDR _   _ _ exists
         \\ transfer   T3 A1 A2    5   _  _   _  _    _ L1 C1   _   _   _   _   _   _  _  _   _   _ _ debit_account_already_closed
         \\ transfer   T3 A2 A1    5   _  _   _  _    _ L1 C1   _   _   _   _   _   _  _  _   _   _ _ credit_account_already_closed
         \\ commit create_transfers
@@ -4545,6 +4546,7 @@ test "create_transfers: closing accounts" {
         // No other pending transfer can be posted or voided in a closed account.
         \\ transfer   T5 A1 A2   10   _  _   _  _    1 L1 C1   _   PEN _   _   _   _  _  _   _   _ _ ok
         \\ transfer   T6 A1 A2    0   _  _   _  _    2 L1 C1   _   PEN _   _   _   _  _  _   CCR _ _ ok // Temporarly closing the credit account.
+        \\ transfer   T6 A1 A2    0   _  _   _  _    2 L1 C1   _   PEN _   _   _   _  _  _   CCR _ _ exists
         \\ transfer   T7 A1 A2   10   T5 _   _  _    _ L1 C1   _   _   POS _   _   _  _  _   _   _ _ credit_account_already_closed
         \\ transfer   T7 A1 A2   10   T5 _   _  _    _ L1 C1   _   _   _   VOI _   _  _  _   _   _ _ credit_account_already_closed
         \\ commit create_transfers
@@ -4567,6 +4569,7 @@ test "create_transfers: closing accounts" {
         \\
         // Closing both accounts permanently.
         \\ transfer   T7 A1 A2    0   _  _   _  _    0 L1 C1   _   _   _   _   _   _  _  CDR CCR _ _ ok
+        \\ transfer   T7 A1 A2    0   _  _   _  _    0 L1 C1   _   _   _   _   _   _  _  CDR CCR _ _ exists
         \\ transfer   T8 A1 A3    1   _  _   _  _    _ L1 C1   _   _   _   _   _   _  _  _   _   _ _ debit_account_already_closed
         \\ transfer   T8 A3 A2    1   _  _   _  _    _ L1 C1   _   _   _   _   _   _  _  _   _   _ _ credit_account_already_closed
         \\ commit create_transfers
```

to verify that `exists` is checked correctly relative to already_closed?

src/state_machine.zig
```diff
@@ -2135,6 +2154,8 @@ pub fn StateMachineType(
             if (t.flags.pending) return .flags_are_mutually_exclusive;
             if (t.flags.balancing_debit) return .flags_are_mutually_exclusive;
             if (t.flags.balancing_credit) return .flags_are_mutually_exclusive;
+            if (t.flags.closing_credit) return .flags_are_mutually_exclusive;
+            if (t.flags.closing_debit) return .flags_are_mutually_exclusive;
```
sentientwaffle: Nit, but swap these so debit is first, as it is for the other checks.

src/state_machine.zig
```diff
@@ -2178,6 +2199,7 @@ pub fn StateMachineType(
                     }
                 }
             };
+            assert((p.amount > 0 and amount > 0) or !forbid_zero_amounts(client_release));
```
sentientwaffle: I think this is not guaranteed -- a zero-amount pending transfer might be created by a 0.16.0 client, and then posted/voided by a 0.15.x client.

docs/reference/requests/create_transfers.md
```diff
@@ -215,6 +215,15 @@ transfer.
 The transfer was not created. [`Transfer.timeout`](../transfer.md#timeout) is nonzero, but only
 [pending](../transfer.md#flagspending) transfers have nonzero timeouts.
 
+### `closing_transfer_must_be_pending`
+
+The transfer was not created. [`Transfer.flags.pending`](../transfer.md#flagspending) is not set,
+but it should be `true`.
```
jorangreef: > but it should be `true`.

We should always explain _why_. For example, why `pending` should be true. This helps the reader understand the spirit/motivation behind the feature, and empathize with why the design makes sense.

jorangreef: At present, we're only stating _must_ but without giving the rational reason behind that _must_. So the reader has no way to understand (at least in the context of reading this here), and yet this understanding is more important than the _must/must not_ because understanding _motivates_.

jorangreef: i.e. _must_ is good but _why_ is even better.

jorangreef (commented):
Awesome to be getting this in, @batiati !

src/state_machine.zig
batiati: Good catch, I think this pre-existing `assert` was too tight then.
If `p` was created on `0.16.0` but `t`  on `0.15.x` with `t.amount = 0` to post the totality of `p`'s amount.

https://github.com/tigerbeetle/tigerbeetle/blob/9b8e34b1aa0c8be2a80e42b6dfdbd0a712138e52/src/state_machine.zig#L2299-L2303

src/clients/java/ci.zig
```diff
@@ -13,7 +13,7 @@ pub fn tests(shell: *Shell, gpa: std.mem.Allocator) !void {
     try shell.zig("build clients:java -Drelease -Dconfig=production", .{});
     try shell.zig("build -Drelease -Dconfig=production", .{});
 
-    try shell.zig("build test:jni", .{});
+    //try shell.zig("build test:jni", .{});
```
sentientwaffle: Did you mean to commit this?

batiati: 😱

docs/reference/requests/create_transfers.md
```diff
@@ -187,6 +215,15 @@ transfer.
...
 [pending](../transfer.md#flagspending) transfers have nonzero timeouts.
 
+### `closing_transfer_must_be_pending`
+
+The transfer was not created. [`Transfer.flags.pending`](../transfer.md#flagspending) is not set,
+but closing transfers must be two-phase pending transfers.
+
+If either [`Transfer.flags.closing_debit`](../transfer.md#flagsclosing_debit) or
+[`Transfer.flags.closing_credit`](../transfer.md#flagsclosing_credit) is set,
+[`Transfer.flags.pending`](../transfer.md#flagspending) must also be set.
```
sentientwaffle, line 225: More explanation:

```suggestion
[`Transfer.flags.pending`](../transfer.md#flagspending) must also be set.
This ensures that closing transfers are reversible (by voiding the pending transfer),
and requires that the reversal operation reference the corresponding closing transfer,
to guard against unexpected interleaving of close/unclose operations.
```

sentientwaffle (approved):

PR #2236, opened 2024-08-23
vsr: op_prepare_max too conservative; decouple into op_prepare_max & op_prepare_ok_max

chaitanyabhandari:
### Problem 

Our journal slot count is set to at least `vsr_checkpoint_ops + lsm_compaction_interval + 2 * pipeline_prepare_queue_max` ops. This is a carefully computed constant that ensures that the cluster doesn't overwrite old ops from the WAL till it is safe to do so.

This is currently achieved by allowing each replica to write a maximum of `op_prepare_max` (set to `op_checkpoint_next + lsm_compaction_interval + pipeline_prepare_queue_max` _before_ this PR) ops to their WAL. This in turn ensures that  ops from the previous wrap aren't evicted from the journal till a commit quorum of replicas reach a particular checkpoint. 

The crux of the problem is that op_prepare_max is  _correct but conservative_, as our journal contains an extra `pipeline_prepare_queue_max` (observe the 2*pipeline_prepare_queue_max in the journal slot count) worth of slots before it wraps around, which are currently underutilized. 

### Solution

This PR allows replicas to fully utilize all slots in the WAL while still being safe and ensuring replicas don't evict ops from the previous wrap till it is safe to do so. This is achieved by decoupling `op_prepare_max` into `op_prepare_max` & `op_prepare_ok_max`. 

Specifically, `op_prepare_max` is now set to `op_checkpoint_next + lsm_compaction_interval + 2 * pipeline_prepare_queue_max` to ensure all WAL slots can be used to _prepare_ ops, before the journal wraps around and ops from the previous wrap are overwritten. 

Additionally, we introduce `op_prepare_ok_max`, which governs whether a replica can _prepare_ok_ a particular op. This is set to  `op_checkpoint_next + lsm_compaction_interval +  pipeline_prepare_queue_max`, such that a replica only prepare_oks a particular if it has a sufficiently fresh checkpoint. This is crucial to ensure that an op from the previous wrap is overwritten only if a commit quorum of replicas have reached a particular checkpoint.

### Orthogonal benefits

Additionally, this PR _**improves VSR availability**_ by allowing potential primaries lagging by a pipeline worth of ops to advance checkpoint during view change. 

Before this PR, a lagging primary used to forfeit view change if the DVC quorum contained headers past the lagging primary's op_prepare_max:
https://github.com/tigerbeetle/tigerbeetle/blob/241c20fc3c3716ef694532568ddc64694e6e76e8/src/vsr/replica.zig#L2097 

Now, we allow lagging primaries to advance their checkpoint even if the DVC quorum contains some operations past its prepare_max (provided the maximum checkpoint in the cluster is not durable, i.e. `commit_max <= op_prepare_max`), since our decoupling of `op_prepare_max` into `op_prepare_max` & `op_prepare_ok_max` allows lagging replicas to prepare these ops and write them to their journal.


chaitanyabhandari:
The failing VOPR seed (`
./zig/zig build -Drelease vopr -- 11017357084674077446`) that CFO found on my branch is failing on `main` (3792676416b3fb33feadee6720dd795cf181a83d) as well. Will take a closer look tomorrow but seems like the VOPR is injecting too many faults?
```C
[debug] (journal): 2: recover_slots: dirty=32 faulty=32
thread 22970946 panic: attempt to unwrap error: WALInvalid
/Users/chaitanyabhandari/Desktop/tigerbeetle/tigerbeetle/src/vsr/replica.zig:657:76: 0x102fe9d57 in open (vopr)
            if (self.journal.faulty.count == constants.journal_slot_count) return error.WALInvalid;
                                                                           ^
/Users/chaitanyabhandari/Desktop/tigerbeetle/tigerbeetle/src/testing/cluster.zig:538:13: 0x102fef68f in replica_open (vopr)
            try replica.open(
            ^
/Users/chaitanyabhandari/Desktop/tigerbeetle/tigerbeetle/src/testing/cluster.zig:478:13: 0x103091243 in restart_replica (vopr)
            try cluster.replica_open(replica_index, .{
            ^
/Users/chaitanyabhandari/Desktop/tigerbeetle/tigerbeetle/src/vopr.zig:1104:17: 0x1030395df in restart_replica (vopr)
        ) catch unreachable;
                ^
```

src/vsr/replica.zig
```diff
@@ -2094,19 +2094,14 @@ pub fn ReplicaType(
...
-                //   headers as a DVC since it has not completed repair.)
-                //
+            const commit_max = @max(
+                self.commit_max,
+                DVCQuorum.commit_max(self.do_view_change_from_all_replicas),
+            );
+            // We are lagging from the cluster by at least a checkpoint, and that checkpoint is
+            // durable on a commit quorum of replicas. Forfeit the view change and prefer the
+            // replica with the durable checkpoint for primary.
+            if (commit_max > self.op_prepare_max()) {
```
matklad: I guess we could look only at `DVCQuorum.commit_max(self.do_view_change_from_all_replicas),` here, we don't need to `@max(self.commit_max`?

actually, could we asser that `DVCQUorum.commit_max >= self.commit_max`? Our DVC should also be a part of the quorum I think?

src/vsr/replica.zig
```diff
@@ -6569,11 +6588,12 @@ pub fn ReplicaType(
             assert(self.pipeline == .cache);
 
             var op = self.commit_max + 1;
-            while (op <= self.op) : (op += 1) {
-                const op_header = self.journal.header_with_op(op).?;
-                if (!self.pipeline.cache.contains_header(op_header)) {
-                    return op;
-                }
+            const view_headers_op_max = self.view_headers.array.get(0).op;
```
matklad: Assert that self.view_headers.command is an SV

matklad: I like how we now don't look at the journal and use an SV as a source of truth. THis is correct, because SV _should_ include the piepeline. 

But can we additionally cross-assert that what's in our SV is consistent with what's in our journal?

src/vsr/replica.zig
```diff
@@ -6508,6 +6514,19 @@ pub fn ReplicaType(
...
+                assert(view_headers_op_max <= self.op_prepare_max());
+                self.op = view_headers_op_max;
+
+                var op = self.commit_max + 1;
+                while (op <= view_headers_op_max) : (op += 1) {
+                    const op_header =
+                        &self.view_headers.array.const_slice()[view_headers_op_max - op];
+                    self.replace_header(op_header);
+                }
+            }
```
matklad: This block is worth a comment!

src/vsr/replica.zig
```diff
@@ -6508,6 +6514,19 @@ pub fn ReplicaType(
                 return .busy;
             }
 
+            const view_headers_op_max = self.view_headers.array.get(0).op;
+            if (view_headers_op_max > self.op) {
+                assert(view_headers_op_max <= self.op_prepare_max());
+                self.op = view_headers_op_max;
```
matklad: Let's call `set_op_and_commit_max` here, so that its easier to find places where we update op, and also to have a trace in the log where this update happens.

matklad (commented):
Overall, looks reasonable, but I'd try folding SV update _on disk_ into this PR as well, as discussed!

src/vsr/replica.zig
matklad: >actually, could we asser that DVCQUorum.commit_max >= self.commit_max? Our DVC should also be a part of the quorum I think?

RElevant code:

```
            // When computing the new commit_max, we cannot simply rely on the fact that our own
            // commit_min is attached to our own DVC in the quorum.
            // Consider the case:
            // 1. Start committing op=N…M.
            // 2. Send `do_view_change` to self.
            // 3. Finish committing op=N…M.
            // 4. Remaining `do_view_change` messages arrive, completing the quorum.
            // In this scenario, our own DVC's commit is `N-1`, but `commit_min=M`.
            // Don't let the commit backtrack.
            const commit_max = @max(
                self.commit_min,
                DVCQuorum.commit_max(self.do_view_change_from_all_replicas),
            );
```

So it looks like we should advance to `commit_min`,if we want to have this assert. But, as we don't have to advance, I think we could just get without this max at all!

chaitanyabhandari: This PR also removes the block of code you've pasted (https://github.com/tigerbeetle/tigerbeetle/pull/2236/files#diff-e95e4dc6c1aae349e798cfeed22ad01f830cce0e654e6b4b765b3e7b0c677b6aL8057-L8069), because we anyway take the max with self.commit_max in set_op_and_commit_max, and _self.commit_max >= self.commit_min_!

https://github.com/tigerbeetle/tigerbeetle/blob/241c20fc3c3716ef694532568ddc64694e6e76e8/src/vsr/replica.zig#L7973-L7975

src/vsr/replica.zig
chaitanyabhandari: https://github.com/tigerbeetle/tigerbeetle/pull/2236/commits/36fef729e334b6fefa0e6ad20fe7f68c0367b457#diff-e95e4dc6c1aae349e798cfeed22ad01f830cce0e654e6b4b765b3e7b0c677b6aR6522-R6524

src/vsr/replica.zig
chaitanyabhandari: https://github.com/tigerbeetle/tigerbeetle/pull/2236/commits/36fef729e334b6fefa0e6ad20fe7f68c0367b457#diff-e95e4dc6c1aae349e798cfeed22ad01f830cce0e654e6b4b765b3e7b0c677b6aR6605-R6612

chaitanyabhandari:
The VOPR issue found on this branch (described in https://github.com/tigerbeetle/tigerbeetle/pull/2236#issuecomment-2309247263) has been fixed by https://github.com/tigerbeetle/tigerbeetle/pull/2248.

chaitanyabhandari:
@matklad since you reviewed this last, 9c59b59502aa038a450a2e6eb3e47f3f6761109b & d5047ce84d61c70a5276a9509422cedc7fc48ef2 changed the definition of op_prepare_max to be `trigger + 2 * pipeline` instead of `trigger + pipeline` and added logic to appropriately withhold prepare_oks such that replicas don't falsely contribute to the _next_ checkpoint's durability. They were VOPR green for ~12 hours before I pushed the latest commit (https://github.com/tigerbeetle/tigerbeetle/pull/2236/commits/e0905d46dd5dcfe4aaf7770624a6cd58db3cc6d3), so I believe this PR is now ready for review again!

https://github.com/tigerbeetle/tigerbeetle/pull/2236/commits/e0905d46dd5dcfe4aaf7770624a6cd58db3cc6d3 only adds `ReplicaType.op_prepare_ok_max` & `Checkpoint.durable` utilities and makes no logical changes to the older commits  (9c59b59502aa038a450a2e6eb3e47f3f6761109b & d5047ce84d61c70a5276a9509422cedc7fc48ef2). 


Screenshot 2024-08-29 at 10 59 27 AM

matklad:
Will take a look tomorrow, have an IronBeetle rn!

chaitanyabhandari:
This has been simmering on the stove nicely; VOPR green for ~20k runs!

Screenshot 2024-08-30 at 9 09 58 AM

src/vsr/replica.zig
```diff
@@ -8055,31 +8065,20 @@ pub fn ReplicaType(
...
-            // 1. Start committing op=N…M.
-            // 2. Send `do_view_change` to self.
-            // 3. Finish committing op=N…M.
-            // 4. Remaining `do_view_change` messages arrive, completing the quorum.
-            // In this scenario, our own DVC's commit is `N-1`, but `commit_min=M`.
-            // Don't let the commit backtrack.
-            const commit_max = @max(
-                self.commit_min,
-                DVCQuorum.commit_max(self.do_view_change_from_all_replicas),
-            );
```
matklad, line 8070: Hm, wait, why is this no longer needed? If this is true, could you leave

`assert(self.commit_min < DVCQuorum.commit_max(self.do_view_change_from_all_replicas))`

?

src/vsr/replica.zig
```diff
@@ -5763,6 +5760,12 @@ pub fn ReplicaType(
             return vsr.Checkpoint.trigger_for_checkpoint(self.op_checkpoint_next()).?;
         }
 
+        /// Returns the op that triggered the current checkpoint.
+        fn op_checkpoint_trigger(self: *const Self) u64 {
```
matklad: I'd rather we not have this _particualr_ helper: we already have `op_checkpoint_next_trigger`, and it feels like confusing the two would be _very_ easy (mentally, I call "checkpoint trigger" the trigger that is in the future)

src/vsr/replica.zig
```diff
@@ -2094,19 +2093,15 @@ pub fn ReplicaType(
...
-                // - If 1 and ¬2, then the ops in the next checkpoint are uncommitted and our
-                //   current checkpoint is sufficient.
-                // - If 2 and ¬1, then the ops past the checkpoint's op-prepare-max must be
-                //   uncommitted because no replica in the DVC quorum even prepared them.
-                //   (They are present anyway because the origin of the DVC is reusing its SV
-                //   headers as a DVC since it has not completed repair.)
-                //
+            const commit_max = @max(
+                self.commit_max,
+                DVCQuorum.commit_max(self.do_view_change_from_all_replicas),
```
matklad, line 2103: explain why we can't rely on our commit_max being a part of DVCQUorum (and using our commit_min, imo, makes this a bit clearer)

src/vsr/replica.zig
```diff
@@ -5783,6 +5786,27 @@ pub fn ReplicaType(
...
+        /// Sending prepare_ok for a op higher than `op_prepare_ok_max` causes a replica to falsely
+        /// contribute to the durability of a checkpoint:
+        /// * A replica could sync to a checkpoint that is not yet durable on a quorum of
+        ///   replicas. To avoid falsely contributing to checkpoint durability, syncing replicas
+        ///   must withhold some prepare_oks till they haven't synced all tables.
+        /// * Replicas can safely write two pipeline_prepare_queue_max worth of ops past
+        ///   the next checkpoint's trigger to their WAL (see `op_prepare_max`). To avoid falsely
+        ///   contributing to the next checkpoint's durability, they must not send prepare_oks for
+        ///   ops past the first pipeline_prepare_queue_max.
+        fn op_prepare_ok_max(self: *const Self) u64 {
```
matklad: Hm, this comment is a bit confusing to me. It makes sense, given that I know the context, but I fear that, when I forget the context, the comment' won't help me much!

Let's maybe re-word it to state invariants positively and front-load the "why", as per tiger style?

Something along the lines of (needs further rewording and grammar work):

* sending `prepare_ok` for `op=N` signifies that a replica has a sufficiently fresh checkponit. Namely, if replcia's current checkpoint is `C`, it won't send prepare OK for ops larger than `C + checkpoint_op + compaction_interval + piepeline_queue`
* this is used to determine when it is safe to overwrite old prepares from the journal (see op_repair_min)
* maybe explain why constants work out the way they do. 

THen, the part about _syncing_ replcia could actually be moved into the body of the function, as it is _implementaiton_ of the contract

src/vsr.zig
```diff
@@ -1520,12 +1520,22 @@ pub const Checkpoint = struct {
         assert(valid(checkpoint));
 
         if (trigger_for_checkpoint(checkpoint)) |trigger| {
-            return trigger + constants.pipeline_prepare_queue_max;
+            return trigger + (2 * constants.pipeline_prepare_queue_max);
```
matklad, line 1523: Now that we can prepare, but not prepare_ok a prepare, let's maybe introduce a `fn send_prepare_ok_after_checkponit`, which we call after checkpoint to send oks for prepares we withheld?

src/vsr/replica.zig
chaitanyabhandari: Oops, I think you missed my last comment from https://github.com/tigerbeetle/tigerbeetle/pull/2236#discussion_r1731384408! Since `set_op_and_commit_max` takes the max with _self.commit_max_, and _self.commit_max >= self.commit_min_, we should be able to remove this right? Or could _self.commit_min be > self.commit_max_ in some case that I'm missing? 🤯 

https://github.com/tigerbeetle/tigerbeetle/blob/56193862533052b686396f521298920812f75d9d/src/vsr/replica.zig#L8107-L8109

src/vsr/replica.zig
chaitanyabhandari: Ah, good point... I haven't been able to find a crisp terminology for this yet as well! When I'm thinking to myself, I usually call `self.op_checkpoint()` as the *current checkpoint* and  `self.op_checkpoint()` as the *next checkpoint*, but I do find myself getting confused a fair few times if I'm not super attentive. 😅

src/vsr.zig
chaitanyabhandari: Ah, good catch, I thought that we could just rely on the primary's _on_prepare_timeout_ such that it requests the prepare_oks that it needs. It would be much better for replicas proactively send prepare_oks after they are done writing checkpoint to superblock!

src/vsr/replica.zig
matklad: Indeed, missed that, thanks! I'd maybe still compute the correct version here, to make it more obvious (or at least add a `maybe(self.commit_max > commit_max)` after we call `set_op_and_commit_max`). 

I am wonder if we should replace that `@max` inside that `set_op_and_commit_max` with an assert, and fix the callsites? But that's for a different PR!

chaitanyabhandari: Ack, would be good to do that in a separate PR! On that note, I'm wondering if we should come up with a way to keep track of TODOs/ideas from review comments (or if we already have one that I'm not aware of), because it seems like it's easy for them to get lost/forgotten... 

Currently, I'm just keeping track of them in my own Slack chat. For accountability I was wondering if we should keep a master GitHub issue/Slack channel where could dump such TODOs as a team?

matklad: Yeah, I've been wondering the same re tracking todos...

src/vsr/replica.zig
chaitanyabhandari: Agreed, I wasn't very happy with this comment as well, it definitely needs a lot of context to understand properly... I haven't spent too much time trying to formalize this yet, it would be great to do that now. Thank you for the direction!

src/vsr/replica.zig
chaitanyabhandari: I would love to create a GitHub issue and assign it to myself immediately, but such ideas/TODOs are usually not concrete enough to create a GitHub issue immediately, and this would also lead to too many open issues... 

A single GitHub issue where we go and dump links to conversations from reviews comments doesn't sound too bad to me though!

src/vsr/replica.zig
chaitanyabhandari: https://github.com/tigerbeetle/tigerbeetle/pull/2236/commits/cecee4531bb3588085d35ecf1caf80d17da0995a

src/vsr.zig
chaitanyabhandari: cecee4531bb3588085d35ecf1caf80d17da0995a

src/vsr/replica.zig
chaitanyabhandari: https://github.com/tigerbeetle/tigerbeetle/pull/2236/commits/161ffd0652c247f3d5c4cbe651123a7b03e56251: Added some assertions and comments while computing commit_max for potential primary, including `assert(self.commit_min <= DVCQuorum.commit_max(self.do_view_change_from_all_replicas));` as you suggested in https://github.com/tigerbeetle/tigerbeetle/pull/2236#discussion_r1738868671.

Thoughts?

src/vsr/replica.zig
chaitanyabhandari: > add a maybe(self.commit_max > commit_max) after we call set_op_and_commit_max

https://github.com/tigerbeetle/tigerbeetle/commit/161ffd0652c247f3d5c4cbe651123a7b03e56251 addresses this.

src/vsr/replica.zig
chaitanyabhandari: Ah, I certainly needed to add this _wrong_ assert to understand this better! It turns out commit_min _can_ be greater than commit_max derived from the DVCQuorum. Relaxed our asserts in https://github.com/tigerbeetle/tigerbeetle/pull/2236/commits/fbfcbffbe0e6e19672361a59b5718d6f9a6180b7, hopefully it is clearer now.

src/vsr/replica.zig
chaitanyabhandari: Took a stab at it in https://github.com/tigerbeetle/tigerbeetle/pull/2236/commits/33c264b3adfc07c29ec0181c7d5afca590ed08cb & 730d4ffce8ba54352854715437d4f0ab6d76806f! Do let me know if we can make it clearer or improve it further. 😄

src/vsr/replica.zig
```diff
@@ -5783,6 +5788,40 @@ pub fn ReplicaType(
             ).?;
         }
 
+        /// Returns the highest op that this replica can safely prepare_ok.
+        ///
+        /// Sending prepare_ok for a particular op signifies that a replica has a sufficiently fresh
```
matklad, line 5795: `s/witth/withh/g`  (two instances)

src/vsr/replica.zig
```diff
@@ -5783,6 +5788,40 @@ pub fn ReplicaType(
...
+        /// checkpoint. Specifically, if a replica is at checkpoint Cₙ, it wittholds prepare_oks for
+        /// ops larger than Cₙ + checkpoint_ops + compaction_interval + pipeline_prepare_queue_max.
+        /// Committing past this op would allow a primary at checkpoint Cₙ₊₁ to overwrite ops from
+        /// the previous wrap, which is safe to do only if a commit quorom of replicas are on Cₙ₊₁.
```
matklad: `s/quorom`/quorum`

src/vsr/replica.zig
```diff
@@ -7173,6 +7201,27 @@ pub fn ReplicaType(
...
+                // We may have breaks or stale headers in our uncommitted chain here. However:
+                // * being able to send what we have will allow the pipeline to commit earlier, and
+                // * the primary will drop any prepare_ok for a prepare not in the pipeline.
+                // This is safe only because the primary can verify against the prepare checksum.
+                if (self.journal.header_with_op(op)) |header| {
+                    self.send_prepare_ok(header);
+                    defer self.flush_loopback_queue();
+                }
+            }
+        }
```
matklad, line 7222: Extract common code between this function and the one bellow into

```
fn send_prepare_oks_from(self: *Self, op: u64)
```

I am wondering if for `send_prepare_oks_after_checkpoint` we could start from `@max(commit_max, trigger + pipeline + 1)`? That is, if we already know that the stuff is committed, no need to re-ack it.

src/vsr/replica_test.zig
```diff
@@ -833,6 +836,74 @@ test "Cluster: view-change: duel of the primaries" {
     try expectEqual(t.replica(.R2).commit(), 3);
 }
 
+test "Cluster: view_change: lagging replica advances checkpoint during view change" {
```
matklad, line 839: Consider writing a short paragraph at the begining of the test to explain the high-level picture for _why_ we need this behavior. See, eg, `"Cluster: sync: slightly lagging replica"`.

Here, I think it would be something like this:

> It could be the case that the replica with the most advanced checkpoint has its checkpoint corrupted. In this case, a replica with a slightly older checkpoint must step up as a primary

matklad (commented):
The logic looks perfect to me now, just some typos/minor code duplication! d+ otherwise!

src/vsr/replica.zig
chaitanyabhandari: Great to get your green light on this, I was _about_ to ask you about this deduplication but decided against it because the two scenarios have distinct things we can assert! Let me lift those up to the callsite if they don't exist already.

chaitanyabhandari: Also, nice observation re sending duplicate prepare_oks. I've incorporated both comments in https://github.com/tigerbeetle/tigerbeetle/pull/2236/commits/d77bf8d17a44d74addde2fdd224981a5c000f36a!

matklad (approved):
LGTM, but let's merge after today's release to avoid races on the changelog!

chaitanyabhandari:
> LGTM, but let's merge after today's release to avoid races on the changelog!

Ack, and thanks a lot for the review and discussions @matklad! We somehow end up finding very interesting side quests & alternative solutions in the changes that I take up. 🚀

I'm also going to change the description of this PR to reflect the solution that we finally implemented, as it _actually_ is a broader improvement that just happens to solve the issue that we set out to solve in the first place.

PR #2238, opened 2024-08-23
LSM: Optionally index zero values

sentientwaffle:
As of https://github.com/tigerbeetle/tigerbeetle/pull/2220, zero-amount transfers are allowed. Since we have an index on `Transfer.amount`, we need to ensure zero-amount transfers are indexed too!

Note that we don't support querying by amount yet, so this isn't actually tested.

This commit should be merged before `0.16.0` is released.

### DIFF
--- a/src/lsm/groove.zig
+++ b/src/lsm/groove.zig
@@ -152,6 +152,10 @@ pub fn GrooveType(
     /// - ignored: [][]const u8:
     ///     An array of fields on the Object type that should not be given index trees
     ///
+    /// - optional: [][]const u8:
+    ///     An array of index names (object fields and derived indexes) that should *not* index
+    ///     zero values.
+    ///
     /// - derived: { .field = *const fn (*const Object) DerivedType }:
     ///     An anonymous struct which contain fields that don't exist on the Object
     ///     but can be derived from an Object instance using the field's corresponding function.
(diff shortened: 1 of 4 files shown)


src/lsm/groove.zig
```diff
@@ -1288,6 +1300,13 @@ pub fn GrooveType(
             };
         }
 
+        fn index_zeroes(comptime field: std.builtin.Type.StructField) bool {
+            for (groove_options.optional) |field_name| {
```
batiati, line 1304: Can we enforce comptime evaluation here with a `comptime for`?

sentientwaffle: I think this is comptime evaluation already, since the only argument is comptime, but I made all invocations `comptime index_zeroes()` so that it is unambiguous.

sentientwaffle:
As part of the review, please sanity check my choice of `optional` fields! (Both the positive and negative space.)

src/state_machine.zig
```diff
@@ -231,6 +243,7 @@ pub fn StateMachineType(
                 .ids = constants.tree_ids.transfers_pending,
                 .batch_value_count_max = batch_value_count_max.transfers_pending,
                 .ignored = &[_][]const u8{"padding"},
+                .optional = &[_][]const u8{"status"},
```
batiati, line 246: Zero means `TransferPendingStatus.none` (not pending), so it's not expected to have any object with `status == 0`.

I don't have a strong opinion here, the index could be considered either optional or required to be non-zero.

batiati (approved):

src/state_machine.zig
sentientwaffle: Ah, I didn't realize that. I guess it doesn't matter in either case then.

PR #2240, opened 2024-08-24
LSM: Secondary indexes on flags.

batiati:
### LSM: Secondary indexes on flags.

- Allow secondary indexes to be a `CompositeKey(void)` for minimal overhead when indexing flags (see commit messages).
- StateMachine: Index `Accounts.flags.imported` and `Transfer.flags.imported`.
- Fuzz: scan indexed flags.

Note: Since this PR adds a new index over a flag introduced by #2171, it should be merged preferentially before releasing `0.16.0`.

EDIT:
The two new indexes add ~24MiB of RSS.
Running the benchmark with `--flag-imported` increases the data file only by ~90MiB, with no performance impact.

### DIFF
--- a/src/lsm/composite_key.zig
+++ b/src/lsm/composite_key.zig
@@ -5,11 +5,17 @@ const math = std.math;
 const stdx = @import("../stdx.zig");
 const constants = @import("../constants.zig");
 
+/// Combines a field (the key prefix) with a timestamp (the primary key).
+/// To keep alignment, it supports either `u64` or `u128` prefixes (which can be truncated
+/// to smaller types to fit the correct field data type).
+/// It also supports composite keys without a prefix (`Field == void`), which is useful for
+/// indexing flags that are only checked with "exists".
 pub fn CompositeKeyType(comptime Field: type) type {
     // The type if zeroed padding is needed.
     const Pad = switch (Field) {
-        u128 => u64,
+        void => u0,
         u64 => u0,
+        u128 => u64,
         else => @compileError("invalid Field for CompositeKey: " ++ @typeName(Field)),
     };
 
@@ -17,14 +23,17 @@ pub fn CompositeKeyType(comptime Field: type) type {
         const CompositeKey = @This();
 
         pub const sentinel_key: Key = key_from_value(&.{
-            .field = math.maxInt(Field),
+            .field = if (Field == void) {} else math.maxInt(Field),
             .timestamp = math.maxInt(u64),
         });
 
         const tombstone_bit: u64 = 1 << (64 - 1);
 
         // u128 may be aligned to 8 instead of the expected 16.
-        const field_bitsize_alignment = @divExact(@bitSizeOf(Field), 8);
+        const field_bitsize_alignment = @max(
+            @divExact(@bitSizeOf(Field), 8),
+            @divExact(@bitSizeOf(u64), 8),
+        );
 
         pub const Key = std.meta.Int(
             .unsigned,
@@ -37,8 +46,13 @@ pub fn CompositeKeyType(comptime Field: type) type {
         padding: Pad = 0,
 
         comptime {
-            assert(@sizeOf(CompositeKey) == @sizeOf(Field) * 2);
             assert(@sizeOf(CompositeKey) == @sizeOf(Key));
+            assert(@sizeOf(CompositeKey) == switch (Field) {
+                void => @sizeOf(u64),
+                u64 => @sizeOf(u128),
+                u128 => @sizeOf(u256),
+                else => unreachable,
+            });
             assert(@alignOf(CompositeKey) >= @alignOf(Field));
             assert(@alignOf(CompositeKey) == field_bitsize_alignment);
             assert(stdx.no_padding(CompositeKey));
@@ -46,11 +60,17 @@ pub fn CompositeKeyType(comptime Field: type) type {
 
         pub inline fn key_from_value(value: *const CompositeKey) Key {
             if (constants.verify) assert(value.padding == 0);
-            return @as(Key, value.timestamp & ~tombstone_bit) | (@as(Key, value.field) << 64);
+            if (Field == void) {
+                comptime assert(Key == u64);
+                return value.timestamp & ~tombstone_bit;
+            } else {
+                comptime assert(@sizeOf(Key) == @sizeOf(Field) * 2);
+                return @as(Key, value.timestamp & ~tombstone_bit) | (@as(Key, value.field) << 64);
+            }
         }
 
         pub inline fn key_prefix(key: Key) Field {
-            return @truncate(key >> 64);
+            return if (Field == void) {} else @truncate(key >> 64);
         }
 
         pub inline fn tombstone(value: *const CompositeKey) bool {
@@ -60,11 +80,10 @@ pub fn CompositeKeyType(comptime Field: type) type {
 
         pub inline fn tombstone_from_key(key: Key) CompositeKey {
             const timestamp: u64 = @truncate(key);
-            const field: Field = @truncate(key >> 64);
             assert(timestamp & tombstone_bit == 0);
 
             return .{
-                .field = field,
+                .field = key_prefix(key),
                 .timestamp = timestamp | tombstone_bit,
             };
         }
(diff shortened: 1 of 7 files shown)


src/lsm/composite_key.zig
```diff
@@ -5,26 +5,35 @@ const math = std.math;
 const stdx = @import("../stdx.zig");
 const constants = @import("../constants.zig");
 
+/// Combines a field (the key prefix) with a timestamp (the primary key).
+/// To keep alignment, it supports either `u64` or `u128` prefixes (which can be truncated
+/// to smaller types to fit the correct field data type).
+/// It also supports composite keys without a prefix (`Field == void`), which is useful for
+/// indexing flags that are only checked with "exists".
```
sentientwaffle:
```suggestion
/// Combines a field (the key prefix) with a timestamp (the primary key).
/// - To keep alignment, it supports either `u64` or `u128` prefixes (which can be truncated
///   to smaller types to fit the correct field data type).
/// - "Deleted" values are denoted by a tombstone bit in the timestamp.
/// - It also supports composite keys without a prefix (`Field == void`), which is useful for
///   indexing flags that are only checked with "exists".
```

src/lsm/composite_key.zig
```diff
@@ -78,48 +97,76 @@ pub fn is_composite_key(comptime Value: type) bool {
...
+                .timestamp = 100,
+            });
+            try std.testing.expect(a == b);
+        }
+
+        {
+            const key = CompositeKey.key_from_value(&.{ .field = 1, .timestamp = 100 });
+            const value = CompositeKey.tombstone_from_key(key);
+            try std.testing.expect(CompositeKey.tombstone(&value));
+            try std.testing.expect(value.timestamp == @as(u64, 100) | CompositeKey.tombstone_bit);
```
sentientwaffle, line 146: Can you also add a negative test for `CompositeKey.tombstone`? Right now we only test the positive case. (Likewise for `CompositeKeyType(void)`.)

src/lsm/forest_fuzz.zig
```diff
@@ -748,11 +758,22 @@ const Environment = struct {
...
                 // Asserting the positive space:
                 // all objects found by the scan must exist in our model.
-                for (accounts) |account| {
+                for (accounts) |*account| {
                     const prefix_current: u128 = switch (params.index) {
-                        inline else => |field| @field(account, @tagName(field)),
+                        .imported => index: {
+                            assert(params.min == 0 and params.max == 0);
+                            assert(account.flags.imported);
+                            break :index 0;
```
sentientwaffle: Maybe s/0/undefined, to emphasize that this value isn't ever used? (Assuming I'm reading the code right!)

And down below, in the `else` branch of `if (params.min == params.max) {`, you could `assert(params.index != .imported);`.

src/lsm/forest_fuzz.zig
```diff
@@ -748,11 +758,22 @@ const Environment = struct {
 
                 // Asserting the positive space:
                 // all objects found by the scan must exist in our model.
-                for (accounts) |account| {
+                for (accounts) |*account| {
                     const prefix_current: u128 = switch (params.index) {
-                        inline else => |field| @field(account, @tagName(field)),
+                        .imported => index: {
+                            assert(params.min == 0 and params.max == 0);
```
sentientwaffle: Can we also assert `prefix_last == null`?

src/lsm/forest_fuzz.zig
```diff
@@ -762,7 +783,11 @@ const Environment = struct {
                     assert(model_account.timestamp == account.timestamp);
                     assert(model_account.ledger == account.ledger);
                     assert(model_account.code == account.code);
-                    assert(std.meta.eql(model_account.flags, account.flags));
+                    assert(stdx.equal_bytes(
+                        tb.AccountFlags,
+                        &model_account.flags,
+                        &account.flags,
+                    ));
```
sentientwaffle, line 792: Is this a cleanup/performance change, or something else?

src/lsm/forest_fuzz.zig
```diff
@@ -748,11 +758,22 @@ const Environment = struct {
...
+                            assert(params.min == 0 and params.max == 0);
+                            assert(account.flags.imported);
+                            break :index 0;
+                        },
+                        inline else => |field| index: {
+                            const Helper = GrooveAccounts.IndexTreeFieldHelperType(@tagName(field));
+                            comptime assert(Helper.Index != void);
+
+                            const value = Helper.derive_index(account).?;
+                            assert(value >= params.min and value <= params.max);
```
sentientwaffle, line 775: Nit, but I think splitting `assert(... and ...);` into `assert(...); assert(...);` is a good habit -- it is a little cleaner, and more precise in the case of failure.

src/lsm/groove.zig
```diff
@@ -79,6 +79,7 @@ const IdTreeValue = extern struct {
 /// Normalizes index tree field types into either u64 or u128 for CompositeKey
 fn IndexCompositeKeyType(comptime Field: type) type {
     switch (@typeInfo(Field)) {
+        .Void => return void,
         .Enum => |e| {
```
sentientwaffle, line 83: May as well add a "test" to the `comptime {` block below: `assert(IndexCompositeKeyType(void) == void);`.

src/lsm/groove.zig
sentientwaffle, line 1: In the `GrooveType` comment:

Change
```
    /// - derived: { .field = *const fn (*const Object) DerivedType }:
```
to:
```
    /// - derived: { .field = *const fn (*const Object) ?DerivedType }:
```

src/lsm/groove.zig
```diff
@@ -343,37 +348,47 @@ pub fn GrooveType(
...
-            return @typeInfo(derived_fn).Fn.return_type.?;
+            return @TypeOf(@field(@as(Object, undefined), field_name));
         }
 
         fn HelperType(comptime field_name: []const u8) type {
             return struct {
-                const Index = IndexType(field_name);
-                const IndexInteger = switch (@typeInfo(Index)) {
+                pub const Index = IndexType(field_name);
+                pub const IndexPrefix = switch (@typeInfo(Index)) {
```
sentientwaffle: :heart: "Prefix" is a much better name than "Integer"!

src/lsm/groove.zig
```diff
@@ -1119,13 +1137,14 @@ pub fn GrooveType(
...
-                        .field = index,
-                    });
+                if (Helper.derive_index(object)) |value| {
+                    const allow_zero = comptime index_zeroes(field);
+                    if (allow_zero or value != 0) {
+                        @field(groove.indexes, field.name).put(&.{
+                            .timestamp = object.timestamp,
+                            .field = value,
+                        });
+                    }
```
sentientwaffle: :thinking: It is a little weird that derived indexes now have 2 different ways to be optional.
Maybe we should _require_ that derived indexes allow zero values?

(I recognize that for `expires_at` and `imported` it doesn't make a difference one way or the other, since they are nonzero.)

batiati: Since `derive_index` is the common abstraction for dealing with fields and functions, we could unify the way to represent optional indexes always as `null`.

What do you think if the function `derive_index` (maybe renamed as `index_from_object`) returned `null` instead of zero for optional fields (and the "optional" list would be only applicable to fields, not to derived indexes)?

src/lsm/forest_fuzz.zig
batiati: Nothing specific, but I guess `stdx.equal_bytes` is preferable over `std.meta.eql`.

src/lsm/forest_fuzz.zig
batiati: I'll take a look, but it's because the fuzzer uses a `u128` as a type-erased prefix (so it's no aware of the field's type).
In fact, not mixing zero with "nothing" was the main reason I chose to use `void` instead of `u0` for the `CompositeKey` prefix.

src/lsm/groove.zig
sentientwaffle: :+1: Sounds good to me! It might be a bit more verbose (due to all of the extra function definitions in state machine), but if there are many of them we could add a helper function to generate the `index_from_object` functions.

sentientwaffle (approved):

PR #2242, opened 2024-08-25
[PoC] Stack allocation.

batiati:
PoC: Just for running GHA

- Added 8 dummy indexes to force [stack overflow](https://github.com/tigerbeetle/tigerbeetle/actions/runs/10547365124/job/29219958195) (the `Forest` object is ~608KiB and the `StateMachine` 717KiB).
  At first, I tried to handle the stack allocation directly on the root cause, heap allocating `Replica` on `main.zig`.
  It didn't work. I suspect the problem is due to stack allocations of locals during the various `init` functions.

- The proposed solution moved large declarations that don't need a "relative" pointer for @fieldParentPtr to the heap.
  Objects containing large arrays are now created on the heap and then initialized.

  -  The `Groove`'s trees (particularly the large struct `IndexTrees`) are now pointers to heap-allocated object trees. 
  -  Similarly, `Manifest` and `Compaction` are also heap-allocated.
  
The Forest object is now approximately 1.5KiB in size.
Interestingly, this change reduced the total RSS by 4KiB!

### DIFF
--- a/src/lsm/forest.zig
+++ b/src/lsm/forest.zig
@@ -37,10 +37,10 @@ pub fn ForestType(comptime _Storage: type, comptime groove_cfg: anytype) type {
         groove_fields = groove_fields ++ [_]std.builtin.Type.StructField{
             .{
                 .name = field.name,
-                .type = Groove,
+                .type = *Groove,
                 .default_value = null,
                 .is_comptime = false,
-                .alignment = @alignOf(Groove),
+                .alignment = @alignOf(*Groove),
             },
         };
 
@@ -78,7 +78,7 @@ pub fn ForestType(comptime _Storage: type, comptime groove_cfg: anytype) type {
         comptime var ids: []const u16 = &.{};
 
         inline for (std.meta.fields(_Grooves)) |groove_field| {
-            const Groove = groove_field.type;
+            const Groove = std.meta.Child(groove_field.type);
 
             for (std.meta.fields(@TypeOf(Groove.config.ids))) |field| {
                 const id = @field(Groove.config.ids, field.name);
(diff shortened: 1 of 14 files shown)


src/lsm/forest.zig
```diff
@@ -271,16 +272,23 @@ pub fn ForestType(comptime _Storage: type, comptime groove_cfg: anytype) type {
...
-            var compaction_pipeline =
-                try CompactionPipeline.init(allocator, grid, options.compaction_block_count);
+            const compaction_pipeline = try CompactionPipeline.init(
+                allocator,
+                grid,
+                options.compaction_block_count,
+            );
             errdefer compaction_pipeline.deinit(allocator);
 
             const scan_buffer_pool = try ScanBufferPool.init(allocator);
```
matklad, line 294: Heh, incidently I was playing with copyhound on the weekend, and one thing I've noticed is that here and elsewhere we do

```
return Forest{
   ...
}
```

But, acccoding to the table at https://ziglang.org/documentation/master/#Result-Locations, that is going to be a stack-to-stack copy of the entire forest. 

If we do

```
return .{
   ...
};
```

then we'll get RLS, and the thing would be in-place constructed, so that'll actually prevent the copies. 

It _seems_ that the semantics we want here is to heap-allocate the entire replica, and then in-place construct the forest. 

Though, not sure if that's workable with error handling...

matklad: Asked the hive mind here: https://ziggit.dev/t/result-location-semantics-and-error-handling/5767

batiati: > Asked the hive mind here: https://ziggit.dev/t/result-location-semantics-and-error-handling/5767
:popcorn: 

I think we would need to reverse the logic, passing the result location to `init` in order to initialize it in place, like #2243.

PR #2244, opened 2024-08-26
vsr: embrace more concurrency while repairing

matklad:
Try to repair (but not commit) prepares, even if we don't have all the headers between checkpoint and head.

There are two degrees of validity of hash chain:

A) hashchain chains onto our op_head
B) hashchain chains onto both op_head _and_ op_checkpoint

Right now, we use A) in `on_prepare / on_repair` path and use B) in `on_repair_timeout` and `commit_journal`.

This commit relaxes the `on_repair` condition to B) -- if we'd accept prepare if it just somehow ended up in our inbox, I think it is natural to go and request it. Doing things differently on normal and on repair paths is just inconsistent.

Maaaybe we should strengthen the `on_prepare / on_repair` to also require that we have full hashchain, but that seems like it'll significantly reduce the benefit of our out-of-order prepare infrastructure, so let's just double down on concurrency here. The key invariant is we only start _committing_ when the chain is safely anchored at both ends, and that is still true.

The VOPR seed here is _indirectly_ related. The seed is actually a false positive due to `core_missing_prepare` only looking at the oldest prepare. That is, under this seed the cluster _should_ remain unavailable.

This change "fixes" the VOPR seed, presumably because the replicas repair more aggressively and manage to keep the cluster available despite the faults. Though, there's a possibility that what fixes the seed is just an unrelated perturbation of messages in the network.

Seed: ./zig/zig build -Drelease -Dvopr-state-machine=testing vopr -- 11711437087271027568

### DIFF
--- a/src/vsr/replica.zig
+++ b/src/vsr/replica.zig
@@ -6225,18 +6225,7 @@ pub fn ReplicaType(
                         .op_max = range.op_max,
                     }),
                 );
-
-                if (range.op_max > self.op_checkpoint()) {
-                    // If there's a header break after checkpoint, wait until it is repaired.
-                    // Otherwise, concurrently repair headers before checkpoint, while trying
-                    // to commit everything after the checkpoint.
-                    return;
-                }
             }
-            assert(self.valid_hash_chain_between(
-                @min(self.op_checkpoint() + 1, self.op),
-                self.op,
-            ));
 
             if (self.journal.dirty.count > 0) {
                 // Request and repair any dirty or faulty prepares.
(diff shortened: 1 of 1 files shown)


matklad:
Embracing concurrency also in the reviews :D

src/vsr/replica.zig
```diff
@@ -9129,7 +9130,10 @@ pub fn ReplicaType(
         }
 
         /// Whether it is safe to commit or send prepare_ok messages.
-        /// Returns true if the hash chain is valid and up to date for the current view.
+        /// Returns true if the hash chain is valid:
+        ///   - connects to the checkpoint checkpoint
```
```suggestion
        ///   - connects to the checkpoint
```

chaitanyabhandari (approved):

Issue #2241, opened 2024-08-24
Error SystemOutdated when starting cluster Ubuntu 24.04 s390x, when all updates are already installed


xet7:
Running at IBM LinuxOne Community Cloud, with debug version:
```
linux1@ubu2404s390x:~$ uname -a
Linux ubu2404s390x 6.8.0-41-generic #41-Ubuntu SMP Fri Aug  2 19:51:49 UTC 2024 s390x s390x s390x GNU/Linux
linux1@ubu2404s390x:~$ cat /etc/issue
Ubuntu 24.04 LTS \n \l
linux1@ubu2404s390x:~$ free -h
               total        used        free      shared  buff/cache   available
Mem:           7.8Gi       1.4Gi       4.7Gi        40Mi       1.9Gi       6.4Gi
Swap:             0B          0B          0B
linux1@ubu2404s390x:~/repos$ git clone https://github.com/tigerbeetle/tigerbeetle && cd tigerbeetle
./zig/download.sh # or .bat if you're on Windows.
./zig/zig build
./tigerbeetle version
Cloning into 'tigerbeetle'...
remote: Enumerating objects: 63964, done.
remote: Counting objects: 100% (787/787), done.
remote: Compressing objects: 100% (333/333), done.
remote: Total 63964 (delta 518), reused 613 (delta 421), pack-reused 63177 (from 1)
Receiving objects: 100% (63964/63964), 41.82 MiB | 14.53 MiB/s, done.
Resolving deltas: 100% (42744/42744), done.
Downloading Zig 0.13.0 release build...
Downloading https://ziglang.org/download/0.13.0/zig-linux-x86_64-0.13.0.tar.xz...
Extracting zig-linux-x86_64-0.13.0.tar.xz...
Downloading completed (/home/linux1/repos/tigerbeetle/zig/zig)! Enjoy!
TigerBeetle version 0.0.1+3792676
linux1@ubu2404s390x:~/repos/tigerbeetle$ ./tigerbeetle format --cluster=0 --replica=0 --replica-count=1 --development 0_0.tigerbeetle
info(io): creating "0_0.tigerbeetle"...
info(io): allocating 1.06298828125GiB...
error: SystemOutdated
/home/linux1/repos/tigerbeetle/zig/lib/std/os/linux/IoUring.zig:61:19: 0x13b0635 in init_params (tigerbeetle)
        .NOSYS => return error.SystemOutdated,
                  ^
[...]

PR #2246, opened 2024-08-26
io: print less confusing error message if IO_uring is not available

matklad:
When IO_uring is disabled (eg, by default in docker), kernel returns `ENOSYS`, and Zig converts that to `SystemOutdated`. This is misleading --- the system might be perfectly up-to-date, and the io_uring just disabled by _some_ security policy.

As we have an explicit version check above, it is most likely that `SystemOutdated` means security stuff, so print an appropriate error message.

I wish I could point the user at a _specific_ security policy, but it looks like there are at least two ways to disable io_uring:

* via seccomp
* via systcl https://lwn.net/Articles/937013/

related: #2241

### DIFF
--- a/src/io/linux.zig
+++ b/src/io/linux.zig
@@ -40,6 +40,19 @@ pub const IO = struct {
             @panic("Linux kernel 5.5 or greater is required for io_uring OP_ACCEPT");
         }
 
+        errdefer |err| switch (err) {
+            error.SystemOutdated => {
+                log.err("io_uring is not available", .{});
+                log.err("likely cause: the syscall is disabled by seccomp", .{});
+            },
+            error.PermissionDenied => {
+                log.err("io_uring is not available", .{});
+                log.err("likely cause: the syscall is disabled by sysctl, " ++
+                    "try 'sysctl -w kernel.io_uring_disabled=0'", .{});
+            },
+            else => {},
+        };
+
         return IO{ .ring = try IO_Uring.init(entries, flags) };
     }
 


src/io/linux.zig
```diff
@@ -40,6 +40,13 @@ pub const IO = struct {
             @panic("Linux kernel 5.5 or greater is required for io_uring OP_ACCEPT");
         }
 
+        errdefer |err| {
```
cb22: :exploding_head:

cb22:
> I wish I could point the user at a specific security policy, but it looks like there are at least two ways to disable io_uring:

I think we can! If io_uring is disabled by the sysctl (`kernel.io_uring_disabled`) it [actually returns](https://lore.kernel.org/io-uring/168910866387.217585.15007808058378960214.b4-ty@kernel.dk/T/) `EPERM`; so, we can handle that and direct the user to `sysctl -w kernel.io_uring_disabled=0` or similar.

matklad:
Aha, so if it is _not_ permission denied, then it must be seccomp! Added that case!

cb22 (approved):
Let's get this in, because it's a good improvement over what we do (nothing) at the moment!

But, in general, I wonder how helpful these error messages should become in future. Should we link to a page on the docs? For example, if an operator runs the sysctl, it'll work _now_ but if they reboot the machine it won't have been persisted.

matklad:
Yeah, linking to docs sounds like a fantastic rule of thumb for cases like these!

PR #2248, opened 2024-08-26
vopr: heal *both* wal header sectors before replica startup

chaitanyabhandari:
Solves failing VOPR seed (`./zig/zig build -Drelease vopr -- 11017357084674077446`) on main (https://github.com/tigerbeetle/tigerbeetle/commit/3792676416b3fb33feadee6720dd795cf181a83d). 

Our [current logic](https://github.com/tigerbeetle/tigerbeetle/blob/56193862533052b686396f521298920812f75d9d/src/vopr.zig#L1049-L1061) arbitrarily repairs one WAL header sectors if _both_ sectors are corrupt. This doesn't account for the following cases that could also lead to `open` returning `WALInvalid`:
* Only one of the WAL header sectors is corrupt and the other one is corrupted on startup (what happened in the current failing seed)
* The sector that is arbitrarily repaired is corrupted again on startup

Instead of arbitrarily repairing one sector, we now repair _both_ sectors so that even if one of them corrupted on startup, the replica still remains operational.

### DIFF
--- a/src/vopr.zig
+++ b/src/vopr.zig
@@ -1041,8 +1041,8 @@ pub const Simulator = struct {
             // If the entire Zone.wal_headers is corrupted, the replica becomes permanently
             // unavailable (returns `WALInvalid` from `open`). In the simulator, there are only two
             // WAL sectors, which could both get corrupted when a replica crashes while writing them
-            // simultaneously. To keep the replica operational, arbitrarily repair one of the
-            // sectors.
+            // simultaneously. Repair both sectors so that even if one of them becomes corrupted on
+            // startup, the replica still remains operational.
             //
             // In production `journal_iops_write_max < header_sector_count`, which makes is
             // impossible to get torn writes for all journal header sectors at the same time.
@@ -1050,14 +1050,8 @@ pub const Simulator = struct {
                 @divExact(vsr.Zone.wal_headers.start(), constants.sector_size);
             const header_sector_count =
                 @divExact(constants.journal_size_headers, constants.sector_size);
-            var header_sector_count_faulty: u32 = 0;
             for (0..header_sector_count) |header_sector_index| {
-                header_sector_count_faulty += @intFromBool(replica_storage.faults.isSet(
-                    header_sector_offset + header_sector_index,
-                ));
-            }
-            if (header_sector_count_faulty == header_sector_count) {
-                replica_storage.faults.unset(header_sector_offset);
+                replica_storage.faults.unset(header_sector_offset + header_sector_index);
             }
         }
 


src/vopr.zig
```diff
@@ -1041,23 +1041,17 @@ pub const Simulator = struct {
             // If the entire Zone.wal_headers is corrupted, the replica becomes permanently
             // unavailable (returns `WALInvalid` from `open`). In the simulator, there are only two
             // WAL sectors, which could both get corrupted when a replica crashes while writing them
-            // simultaneously. To keep the replica operational, arbitrarily repair one of the
-            // sectors.
+            // simultaneously. Repair both sectors so that even if one of them corrupted on startup,
+            // the replica still remains operational.
```
```suggestion
            // simultaneously. Repair both sectors so that even if one of them becomes corrupted on startup,
            // the replica still remains operational.
```

sentientwaffle:
:thinking: Could we fix this from the fault injection side instead? i.e. prevent `Storage.reset()` from injecting a fault when doing so would result in 100% of the WAL header sectors being corrupt.

As this patch is right now, we wouldn't be testing recovery with corrupt WAL header sectors, which is not ideal.

chaitanyabhandari:
> 🤔 Could we fix this from the fault injection side instead? i.e. prevent `Storage.reset()` from injecting a fault when doing so would result in 100% of the WAL header sectors being corrupt.
> 
> As this patch is right now, we wouldn't be testing recovery with corrupt WAL header sectors, which is not ideal.

Mhmmmm I thought that we would still be testing recovery with corrupt WAL header sectors, because we are still marking _replica_storage.faulty_ as _True_ if [recoverable_count >= recoverable_count_min or replica.standby()](https://github.com/tigerbeetle/tigerbeetle/blob/56193862533052b686396f521298920812f75d9d/src/vopr.zig#L1029), and so we _could_ corrupt WAL headers while [reading](https://github.com/tigerbeetle/tigerbeetle/blob/56193862533052b686396f521298920812f75d9d/src/testing/storage.zig#L397-L411) them as long as we have non-zero _read_fault_probability_... In fact, I think that is precisely what was happening in this failed seed, we corrupted a WAL header sector on read, which we could still end up doing?  🤔

```C
[debug] (journal): 2: recover: recovering
[debug] (journal): 2: recover_headers: offset=0 size=4096 recovering
[debug] (journal): 2: recover_headers: offset=4096 size=4096 recovering
[debug] (journal): 2: recover_headers: offset=4096 size=4096 recovered
[debug] (storage): 2: corrupting sector at zone=vsr.Zone.wal_headers offset=0 slots=0...16
[debug] (journal): 2: recover_headers: offset=0 size=4096 recovered
[debug] (journal): 2: recover_headers: complete
```
 

Let me double check this though!

sentientwaffle:
Ah, that is a good point!

sentientwaffle (approved):

PR #2249, opened 2024-08-27
ci: remove write privelegies from GITHUB_TOKEN

matklad:
Hat tip to sobolevn for https://github.com/ziglang/zig/pull/21217

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,17 +1,20 @@
 name: CI
 
-concurrency:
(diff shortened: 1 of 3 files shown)


cb22:
Ah - so we have [the setting](https://github.com/tigerbeetle/tigerbeetle/settings/actions) on GH enabled so that the default permissions for the action token is only read, unless explicitly specified otherwise.

matklad:
Ahhh, indeed, I remember now! Yeah, than we don't need this!

PR #2252, opened 2024-08-27
Change `init` signature

batiati:
## Change init signature

Change `fn init(allocator: Allocator) !MyType` to `fn init(self: *MyType, allocator: Allocator) !void` allowing in-place initialization.

The primary motivation for this PR was to address the silent stack growth caused by intermediate copy/move allocations during the initialization of large objects. Specifically, the `Forest` struct can grow indefinitely depending on the number of `Grooves`/`IndexTrees` needed to support the StateMachine's custom logic, causing TigerBeetle to crash during startup due to stack-overflow.

Also changes TigerStyle to recommend the adoption of this pattern.

Best reviewed per commit.

Related: #2242 and #2243.


matklad:
Would be good to work-in the "why?" into commit message (preferable), or PR description!

I think the primary motivation here is to reduce stack space utilization, but this isn't obvious from the relevant commit messages

src/lsm/node_pool.zig
```diff
@@ -4,9 +4,9 @@ const math = std.math;
 const mem = std.mem;
 const meta = std.meta;
 
-pub fn NodePool(comptime _node_size: u32, comptime _node_alignment: u13) type {
```
matklad, line 7: :heart: thanks!

src/lsm/forest.zig
```diff
@@ -236,76 +236,71 @@ pub fn ForestType(comptime _Storage: type, comptime groove_cfg: anytype) type {
...
 
-            var grooves: Grooves = undefined;
             var grooves_initialized: usize = 0;
-
             errdefer inline for (std.meta.fields(Grooves), 0..) |field, field_index| {
                 if (grooves_initialized >= field_index + 1) {
-                    @field(grooves, field.name).deinit(allocator);
+                    const Groove = field.type;
+                    const groove: *Groove = &@field(forest.grooves, field.name);
+                    groove.deinit(allocator);
```
matklad, line 275: Hm, wouldn't `@field(forest.grooves, field.name).deinit(allocator)` work? Feels simpler that way!

matklad: If this is for consistency with init path, I'd use 

```
try @field(forest.grooves, field.name)init(allocator, &forest.node_pool, gird, groove_options,)
```

there as well!

matklad: Otoh, `groove.deinit` is certainly more greppable than `@field().deinit`, so perhaps this new form is more readable? I guess I am find either way!

src/lsm/manifest_level.zig
```diff
@@ -815,28 +827,30 @@ pub fn TestContext(
         inserts: u64 = 0,
         removes: u64 = 0,
 
-        fn init(random: std.rand.Random) !Self {
-            var pool = try TestPool.init(
+        fn init(random: std.rand.Random) !ManifestLevel {
```
matklad: Could we make this in-place as well?

I don't feel good about this-pattern, where we delegate from by-value init to in-place init. 

This is because with in-place, I think it would also be reasonable to assume that you aren't going to get moved, and to assume pointer stability. But if your caller moves you, there's no pointer stability! 

So the risk here is that at some point in the future, we'll notice that, for example, `ManifestLevelType` uses in-place init, use that to observe the pointer, and than the fuzz test would _continue_ to work because Zig optimizes the code to remove copies, so we don't notice the mistake right way. And then, during a Zig upgrade, its optimization change, it introduces the copy, and we'd have to chase that for a long time.

Should we maybe mention this in TigerStyle?

1. To construct large size-of objects, pass an out pointer
2. out-pointer inits can assume pointer stability and immovable types
3. but make sure that in-place ints are viral --- if your field is in-place initialized, you should be in-place as well!

src/lsm/segmented_array.zig
```diff
@@ -1078,7 +1079,8 @@ fn FuzzContextType(
         removes: u64 = 0,
 
         fn init(allocator: std.mem.Allocator, random: std.rand.Random) !Self {
-            var pool = try TestPool.init(allocator, TestArray.node_count_max);
```
matklad, line 1081: Same here, we are in-place initing pool, so the entire FuzzContext has to be in-place

src/vsr/replica.zig
```diff
@@ -1078,7 +1078,7 @@ pub fn ReplicaType(
             }
             errdefer for (self.grid_repair_write_blocks) |b| allocator.free(b);
 
-            self.state_machine = try StateMachine.init(
```
matklad, line 1081: :tada:

build.zig
```diff
@@ -434,8 +434,6 @@ fn build_tigerbeetle_executable(b: *std.Build, options: struct {
     }
     // Ensure that we get stack traces even in release builds.
     tigerbeetle.root_module.omit_frame_pointer = false;
-    // TODO: investigate why do we need to grow the stack when adding new indexes.
-    tigerbeetle.stack_size = 18 * 1024 * 1024;
```
matklad, line 438: :tada: :tada: :tada:

src/lsm/scan_builder.zig
```diff
@@ -46,43 +46,38 @@ pub fn ScanBuilderType(
...
-        // however, merging scans does not require resources.
-        // E.g, if `lsm_scans_max = 4` we can have at most 3 merge operations:
-        // M₁(M₂(C₁, C₂), M₃(C₃, C₄))
-        const ScanSlots = stdx.BoundedArray(Scan, constants.lsm_scans_max);
-        const MergeSlots = stdx.BoundedArray(Scan, constants.lsm_scans_max - 1);
-
-        scan_slots: *ScanSlots,
-        merge_slots: *MergeSlots,
+        scan_slots: Slots,
+        merge_slots: Slots,
```
matklad: Again, would be great to "why" in the commit message!

And, as usual, every time I see bounded array I am tempted by something simple like

```
scans: *[constants.lsm_scans_max]Scan = undefined;
scan_count: u32 = 0;
```

matklad (commented):
Excellent, thanks, we should have done this from the start!

One blocker though is that I think we should ensure that in-place init is always 100% viral, to avoid ambiguities about pointer stability. I think this is violated only in tests a couple of times, but would be great to double check this!

src/lsm/forest.zig
batiati: Reading `@field(...).function(...)` isn't that bad. The main reason for refactoring (here and in `Forest` as well) is to make clear each field's type since it's not trivial to understand what's going on inside a loop with too many metaprogramming instructions.

matklad (approved):

PR #2253, opened 2024-08-27
vsr: primary rejects prepares if log_view is not durable

chaitanyabhandari:
### Problem
Solves a failed VOPR seed (./zig/zig build -Drelease -Dvopr-log=full vopr -- --lite 9087327468587795673) on main(140c7a03d9b6e6ad0a85c46f0b93c597e98f3a95) wherein a primary accepting prepares before making its log_view durable exposes a break in its hash chain.

 A _failing_ replica test that reproduces this issue on a more recent commit on main (aea9cbf1391768c78f35fe31988187119383903e) is in the following GitHub gist: https://gist.github.com/chaitanyabhandari/4a64e74cd90d778d1560d3d327ac4f64

### Solution

The solution entails rejecting prepares on the primary if its view isn't durable. Turns out this piece of code already exists for solo clusters, this PR enables it for primaries in all cases. 
 

### Triggering conditions

1. R1 becomes primary in view=1, initiates writing log_view=1 to the superblock
2. R1 prepares op=1,2,3, and 4.
3. R1 crashes before log_view becomes durable.
4. R1 restarts with self.log_view=0 and truncates op=1,2,3 and 4 as per [case_cut_view_range](https://github.com/tigerbeetle/tigerbeetle/blob/main/src/vsr/journal.zig#L1335-L1345) in journal recovery. R1 sets self.op=0.
5. R1 becomes primary for view=1, successfully writes log_view=1 to the superblock.
6. R1 prepares op=1, replacing the old op=1 in its journal.
7. R1 crashes and restarts with self.log_view=1 and treats op=2,3, and 4 as case [@G ](https://github.com/tigerbeetle/tigerbeetle/blob/main/src/vsr/journal.zig#L2192-L2197)in journal recovery.
8. R1 loads ops=1,2,3, and 4 from the journal into its view headers in [view_change_update_view_headers](https://github.com/tigerbeetle/tigerbeetle/blob/main/src/vsr/replica.zig#L8503)

This exposes a break in the hash chain, since op=2 points to the older op=1 that is prepared, while the journal contains a new op=1 with a different checksum.

### DIFF
--- a/src/vsr/replica.zig
+++ b/src/vsr/replica.zig
@@ -4726,42 +4726,6 @@ pub fn ReplicaType(
             return message.ref();
         }
 
-        /// Discards uncommitted ops during a view change from after and including `op`.
-        /// This is required to maximize availability in the presence of storage faults.
-        /// Refer to the CTRL protocol from Protocol-Aware Recovery for Consensus-Based Storage.
-        fn primary_discard_uncommitted_ops_from(self: *Self, op: u64, checksum: u128) void {
-            assert(self.status == .view_change);
-            assert(self.primary_index(self.view) == self.replica);
-            assert(self.repairs_allowed());
-
-            assert(self.valid_hash_chain(@src()));
-
-            const slot = self.journal.slot_with_op(op).?;
-            assert(op > self.commit_max);
-            assert(op <= self.op);
-            assert(self.journal.header_with_op_and_checksum(op, checksum) != null);
-            assert(self.journal.dirty.bit(slot));
-
-            log.debug("{}: primary_discard_uncommitted_ops_from: ops={}..{} view={}", .{
-                self.replica,
-                op,
-                self.op,
-                self.view,
-            });
-
-            self.op = op - 1;
-            self.journal.remove_entries_from(op);
-
-            assert(self.journal.header_for_op(op) == null);
-            assert(!self.journal.dirty.bit(slot));
-            assert(!self.journal.faulty.bit(slot));
-
-            // We require that `self.op` always exists. Rewinding `self.op` could change that.
-            // However, we do this only as the primary within a view change, with all headers
-            // intact.
-            assert(self.journal.header_with_op(self.op) != null);
-        }
-
         fn flush_loopback_queue(self: *Self) void {
             // There are five cases where a replica will send a message to itself:
             // However, of these five cases, all but one call send_message_to_replica().
--- a/src/vsr/replica.zig
+++ b/src/vsr/replica.zig
@@ -3901,7 +3901,9 @@ pub fn ReplicaType(
             assert(self.commit_min == self.commit_prepare.?.header.op);
             assert(self.commit_min <= self.commit_max);
 
-            if (self.status == .normal and self.primary() and !self.view_durable_updating()) {
+            if (self.status == .normal and self.primary()) {
+                assert(!self.view_durable_updating());
+
                 if (self.pipeline.queue.pop_request()) |request| {
                     // Start preparing the next request in the queue (if any).
                     self.primary_pipeline_prepare(request);
--- a/src/vsr/replica.zig
+++ b/src/vsr/replica.zig
@@ -5055,11 +5055,9 @@ pub fn ReplicaType(
                 }
             }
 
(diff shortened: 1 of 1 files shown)


src/vsr/replica.zig
```diff
@@ -4726,42 +4726,6 @@ pub fn ReplicaType(
             return message.ref();
         }
 
-        /// Discards uncommitted ops during a view change from after and including `op`.
-        /// This is required to maximize availability in the presence of storage faults.
-        /// Refer to the CTRL protocol from Protocol-Aware Recovery for Consensus-Based Storage.
-        fn primary_discard_uncommitted_ops_from(self: *Self, op: u64, checksum: u128) void {
```
matklad, line 4732: wait wat, wasn't this supposed to be used somewhere? 

Ah, ok, this was (not) removed in f74eda450b0a0fa687395e1a5cd97810af0b6345, `git log -S` is everuseful!

src/vsr/replica.zig
```diff
@@ -9516,7 +9478,9 @@ pub fn ReplicaType(
             assert(self.state_machine.pulse_needed(self.state_machine.prepare_timestamp));
 
             self.send_request_to_self(.pulse, &.{});
-            assert(self.pipeline.queue.contains_operation(.pulse));
+            if (!self.pipeline.queue.contains_operation(.pulse)) {
+                assert(self.view_durable_updating());
+            }
```
matklad: As a rule of thumb, ff something might or might not happen, it is genreally better to assert that it does, and to push the condition where it doesn't up to the caller:

https://matklad.github.io/2023/11/15/push-ifs-up-and-fors-down.html

In this case, we actually already do that for upgrade requests!

src/vsr/replica.zig
chaitanyabhandari: Ah, TIL about `git log -S`, I'm going to use this one in the future! :D

src/vsr/replica.zig
chaitanyabhandari: Ah, this a _great_ rule of thumb! It actually helped us  `assert(!self.view_durable_updating());` in primary_pipeline_prepare, send_request_pulse_to_self, _and_ send_request_upgrade_to_self. Added in https://github.com/tigerbeetle/tigerbeetle/pull/2253/commits/f0055581452f8c0ad86d7f2503316e650d863279.

src/vsr/replica.zig
```diff
@@ -3897,7 +3901,7 @@ pub fn ReplicaType(
             assert(self.commit_min == self.commit_prepare.?.header.op);
             assert(self.commit_min <= self.commit_max);
 
-            if (self.status == .normal and self.primary()) {
+            if (self.status == .normal and self.primary() and !self.view_durable_updating()) {
```
matklad: Hm, this feels wrong: if we bail out of this with `self.view_durable_updating()`, then at least we should chain-call this back once view_durable_update finishes, otherwise we won't re-start committing the pipeline. 

But i am wondreifg if we could just `assert` that we are not updating the view here? In this code ,we are primary, in normal, and we are already committing, so the view update must have happened already, no?

chaitanyabhandari: Mhmmmm great catch _"we won't re-start committing the pipeline"_, I think I indeed got overly cautious here. I was worrying about a case wherein a primary reaches here without making its view durable. But that is not possible at all, since a primary starts off with nothing to commit (commit_min == commit_max), and now doesn't accept any new prepares till the view is durable. So if we've reached this point, we _should_ be guaranteed to have made our view durable!

chaitanyabhandari: https://github.com/tigerbeetle/tigerbeetle/pull/2253/commits/3af24ebbffa0c5e6617865a83ed8c08ab9681dee

matklad (approved):
LGTM, but let's bake this in VOPR for some times, I am curious about:

```
            if (self.status == .normal and self.primary()) {
                assert(!self.view_durable_updating());
``` 

in particular. SHould be fine, but only VOPR knows!

chaitanyabhandari:
This is VOPR green after ~9 hours of CFO testing, merging it!

PR #2254, opened 2024-08-28
vsr: complete rollout of the new state sync protocol

matklad:
The final piece of the puzzle -- complete removal the old messages. Now the replicas would panic if they see any of the old messages, treating them exactly as any other kind of an unknown message.

### DIFF
--- a/src/vsr.zig
+++ b/src/vsr.zig
@@ -203,9 +203,6 @@ pub const Command = enum(u8) {
 
     start_view_change = 10,
     do_view_change = 11,
-    // Historical version of SV, without checkpoint. Currently  both `start_view_deprecated` and
-    // `start_view` are handled. Next release will ignore `start_view_deprecated`.
-    start_view_deprecated = 12,
 
     request_start_view = 13,
     request_headers = 14,
@@ -218,14 +215,29 @@ pub const Command = enum(u8) {
     request_blocks = 19,
     block = 20,
 
-    request_sync_checkpoint = 21,
-    sync_checkpoint = 22,
-
     start_view = 23,
 
+    // If a command is removed from the protocol, its ordinal is added here and can't be re-used.
+    const gaps = .{
+        12, // start_view without checkpoint
+        21, // request_sync_checkpoint
+        22, // sync_checkpoint
+    };
+
     comptime {
-        for (std.enums.values(Command), 0..) |result, index| {
-            assert(@intFromEnum(result) == index);
+        var value_previous: ?u8 = null;
+        for (std.enums.values(Command)) |command| {
+            const value_current = @intFromEnum(command);
+            assert(std.mem.indexOfScalar(u8, &gaps, value_current) == null);
+            if (value_previous == null) {
+                assert(value_current == 0);
+            } else {
+                assert(value_previous.? < value_current);
+                for (value_previous.? + 1..value_current) |value_gap| {
+                    assert(std.mem.indexOfScalar(u8, &gaps, value_gap) != null);
+                }
+            }
+            value_previous = value_current;
         }
     }
 };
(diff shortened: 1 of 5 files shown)


matklad:
To be merged after the next release!

src/vsr/replica_test.zig
```diff
@@ -511,8 +511,7 @@ test "Cluster: repair: view-change, new-primary lagging behind checkpoint, forfe
     b2.pass_all(.__, .bidirectional);
     b1.pass_all(.__, .bidirectional);
     a0.drop_all(.__, .bidirectional);
-    // Block state sync to prove that B1 recovers via WAL repair.
-    b1.drop(.__, .bidirectional, .sync_checkpoint);
+    // TODO: make sure that B1 uses WAL repair rather than state sync here.
     const mark = marks.check("on_do_view_change: lagging primary; forfeiting");
```
matklad, line 518: It seems like with both sync protocol, if a replica is partitinoned from the cluster and then rejoins, it will be biased towards state sync. This feels like a bug, ideally, the replica should be able to tell apart "I cant' commit because I cant talk to the cluster at all" from "I cant' commit because some of the prepares/blocks are no longer available", but I don't see a straightforward way to do that just yet, so I'd rather punt on that until we have general partial connectivity tolerance.

src/vsr.zig
```diff
@@ -218,14 +215,29 @@ pub const Command = enum(u8) {
...
-    sync_checkpoint = 22,
-
     start_view = 23,
 
+    // If a command is removed from the protocol, its ordinal is added here and can't be re-used.
+    const gaps = .{
+        12, // start_view without checkpoint
+        21, // request_sync_checkpoint
+        22, // sync_checkpoint
+    };
```
matklad, line 225: The principle I am following here is that, ideally, our production code doesn't have "extras" --- there's no buggify, there's no compatability hacks (eventually), etc. 

So, for deprecated messages, I want to handle them exactly as the unknown messages, rather than, say, ignoring and logging.

matklad:
One thing this PR doesn't do is that it doesn't flip `visit` back to `false`. If I am reading the code correctly, visit affects the _previous_ release, as the current one is always `visit=true`. So it's the next week we could flip visit back to false (but I am tempted to refactor that logic maybe to remove this shift by one)

sentientwaffle (approved):
LGTM, though btw there are some lingering references to `sync_checkpoint` in documentation:

```
$ gg sync_checkpoint
docs/about/internals/sync.md:107:The syncing replica sends `command=request_sync_checkpoint` messages (with the sync target
docs/about/internals/sync.md:108:identifier attached to each) until it receives a `command=sync_checkpoint` with a matching
docs/about/internals/sync.md:167:- `command=request_sync_checkpoint`: Requested checkpoint identifier.
docs/about/internals/sync.md:168:- `command=sync_checkpoint`: Current checkpoint identifier of sender.
docs/about/internals/vsr.md:54:| `request_sync_checkpoint` | replica |      replica | [Sync Superblock](#protocol-sync-superblock)                                                                                               |
docs/about/internals/vsr.md:55:|         `sync_checkpoint` | replica |      replica | [Sync Superblock](#protocol-sync-superblock)                                                                                               |
src/vsr.zig:223:        21, // request_sync_checkpoint
src/vsr.zig:224:        22, // sync_checkpoint
src/vsr/superblock.zig:307:    /// This struct is sent in a `sync_checkpoint` message from a healthy replica to a syncing
```

matklad:
Yeah, the state sync docs are in general out of sync, but that's tackled in #2269 

Curious though, what is `gg`? (my gg is https://github.com/matklad/config/blob/master/gg/src/main.rs)

sentientwaffle:
```
alias gg='git grep --line-number'
```

PR #2255, opened 2024-08-28
vsr: prefer wal repair over state sync

matklad:
New state sync protocol regressed the behavior where the replica would try to repair WAL switching to state sync, and this commit puts the old behavior back in.

Good thing that we have a test here! Sadly, the test was silently regressed as it didn't actually check that the state sync didn't happen. Mark the relevant logs to pin down the behavior!

As for the actual fix, it centers on the `repair_stuck` function. Before, it was a stateless function. My original thinking was that we will only get an SV message in response to our own RSV, and, given that we have a repair_sync_timeout when requesting an SV, there should be natural debouncing here.

What I didn't consider are the two other paths to send an RSV:

* when receiving a message from a newer view
* in `.repair()` when our head is < commit_max

So what actually happens is that a lagging replcia joining the cluster would pretty much spam the primary with RSVs anyway, which in turn, would immediately flip the laggard into the state sync.

So, in this PR `repair_stuck` tracks a bit of state. Namely, once every repair_sync_timeout we note our commit_min, and note whether it is _different_ from the one we had before. commit_min being the same across two timeouts becomes another ingredient to decide if the commit is stuck.

This state is tracked in `sync_wal_repair_progress` which is intentionally never reset and only ever updated by the timeout.

### DIFF
--- a/src/vsr/replica.zig
+++ b/src/vsr/replica.zig
@@ -302,6 +302,12 @@ pub fn ReplicaType(
         /// Invariants:
         /// - If syncing≠idle then sync_tables=null.
         sync_tables: ?ForestTableIterator = null,
+        /// Tracks wal repair progress to decide when to switch to state sync.
+        /// Updated on repair_sync_timeout.
+        sync_wal_repair_progress: struct {
+            commit_min: u64 = 0,
+            advanced: bool = true,
+        } = .{},
 
         /// The release we are currently upgrading towards.
         ///
@@ -367,9 +373,9 @@ pub fn ReplicaType(
         ///
         /// Invariants (not applicable during status=recovering):
         /// * `replica.commit_min` exists in the Journal OR `replica.commit_min == op_checkpoint`.
-        /// * `replica.commit_min ≤ replica.op`
+        /// * `replica.commit_min ≤ replica.op`.
         /// * `replica.commit_min ≥ replica.op_checkpoint`.
-        /// * never decreases while the replica is alive
+        /// * never decreases while the replica is alive and not state-syncing.
         commit_min: u64,
 
         /// The op number of the latest committed operation (according to the cluster).
@@ -2277,7 +2283,7 @@ pub fn ReplicaType(
                     .awaiting_checkpoint => {},
                 }
 
-                log.debug(
+                log.mark.debug(
                     \\{}: on_start_view: sync started view={} op_checkpoint={} op_checkpoint_new={}
                 , .{
                     self.replica,
@@ -2327,9 +2333,17 @@ pub fn ReplicaType(
             }
 
             if (self.syncing == .updating_superblock) {
+                // State sync can "truncate" the first batch of committed ops!
                 maybe(self.commit_min >
                     self.syncing.updating_superblock.checkpoint_state.header.op);
+                assert(self.commit_min < constants.lsm_compaction_ops +
+                    self.syncing.updating_superblock.checkpoint_state.header.op);
+
                 self.commit_min = self.syncing.updating_superblock.checkpoint_state.header.op;
+                self.sync_wal_repair_progress = .{
+                    .commit_min = self.commit_min,
+                    .advanced = true,
+                };
             }
 
             self.view_headers.replace(.start_view, view_headers);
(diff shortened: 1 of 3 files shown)


src/vsr/replica_test.zig
```diff
@@ -471,9 +471,10 @@ test "Cluster: repair: partition 2-1, then backup fast-forward 1 checkpoint" {
     try expectEqual(r_lag.status(), .normal);
     try expectEqual(r_lag.op_checkpoint(), 0);
 
-    // Allow repair, but ensure that state sync doesn't run.
-    r_lag.drop(.__, .bidirectional, .sync_checkpoint);
```
matklad, line 475: Caught the bug when removing `.sync_checkponit` for https://github.com/tigerbeetle/tigerbeetle/pull/2254!

src/vsr/replica.zig
```diff
@@ -3123,6 +3129,14 @@ pub fn ReplicaType(
             assert(self.repair_sync_timeout.ticking);
             self.repair_sync_timeout.reset();
 
+            const commit_min_previous = self.sync_wal_repair_progress.commit_min;
+            assert(commit_min_previous <= self.commit_min);
+            self.sync_wal_repair_progress.commit_min = self.commit_min;
```
sentientwaffle: :confused: Can this line be removed? It looks redundant since we immediately set `self.sync_wal_repair_progress = .{ ...`.

matklad: Ah yeah, that was the first version of the logic!

src/vsr/replica.zig
```diff
@@ -2327,9 +2333,17 @@ pub fn ReplicaType(
             }
 
             if (self.syncing == .updating_superblock) {
+                // State sync can "truncate" the first batch of committed ops!
                 maybe(self.commit_min >
```
matklad, line 2337: TIL that commit_min can actually regress during state sync! It somehow didn't register with me before, but looks correct: you might get stuck compacting the last batch of checkpoint's prepares

sentientwaffle: :laughing: I was surprised by the same thing a while back -- state_checker has a workaround for it!: https://github.com/tigerbeetle/tigerbeetle/blob/aea9cbf1391768c78f35fe31988187119383903e/src/testing/cluster/state_checker.zig#L101-L107

src/vsr/replica.zig
```diff
@@ -302,6 +302,12 @@ pub fn ReplicaType(
         /// Invariants:
         /// - If syncing≠idle then sync_tables=null.
         sync_tables: ?ForestTableIterator = null,
+        /// Tracks wal repair progress to decide when to switch to state sync.
+        /// Updated on repair_sync_timeout.
+        sync_wal_repair_progress: struct {
+            commit_min: u64 = 0,
+            unchanged: bool = false,
+        } = .{},
```
sentientwaffle: Could we get by with just storing `commit_min`, and computing `unchanged` via `self.sync_wal_repair_progress.commit_min == self.commit_min`?

matklad: I think that would be prone to false positives. It could be the case that you invoke `repair_stuck` just after the timeout triggered and updated `commit_min`. 

But I think I see the bug in the opposite direction. If at the point of `repair_stuck` we see `.unchanged`, but our commit_min advanced, than we are not, in fact, stuck. So we kinda need both  I think!

matklad: Ah, but let me remove negation here, `!unchanged` is unnecessary baroque!

sentientwaffle (approved):

src/vsr/replica.zig
```diff
@@ -3123,6 +3137,13 @@ pub fn ReplicaType(
             assert(self.repair_sync_timeout.ticking);
             self.repair_sync_timeout.reset();
 
+            const commit_min_previous = self.sync_wal_repair_progress.commit_min;
+            assert(commit_min_previous <= self.commit_min);
+            self.sync_wal_repair_progress = .{
+                .commit_min = self.commit_min,
+                .advanced = commit_min_previous < self.commit_min,
```
sentientwaffle, line 3144: Maybe add a comment here explaining that this is so that a commit_min regressing due to state sync doesn't count as "advancing".

sentientwaffle: Wait, nevermind, we take care of that in `on_start_view`. So I'm back to not understanding why `advanced` is separate :thinking:

matklad: Ahhh, I think I get what’s confusing here! There two calls to repair stuck: one here, and another in on_start_view. For _this_ call we indeed don’t need advanced! But it is required for the second call site.

sentientwaffle: Aha, I understand now; thank you!

PR #2256, opened 2024-08-28
Build: Check zig version

sentientwaffle:
Note that this check isn't very useful right now -- If you try to compile TB with the wrong zig version, it will most likely fail before this check runs (e.g. while typechecking `build.zig`) due to zig-build instability.

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -42,6 +42,20 @@ fn resolve_target(b: *std.Build, target_requested: ?[]const u8) !std.Build.Resol
     return b.resolveTargetQuery(query);
 }
 
+const zig_version = std.SemanticVersion{
+    .major = 0,
+    .minor = 13,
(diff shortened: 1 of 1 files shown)


matklad:
I am wondering if we should be setting https://github.com/ziglang/zig/blob/master/doc/build.zig.zon.md#minimum_zig_version?

That in theory should be relatively more stable across zig versions! But is it worth adding one more top-level file?

sentientwaffle:
I think we want to be more strict than just a minimum version -- we probably want to ensure that the tigerbeetle you are building was compiled with the exact same version that CFO is fuzzing.

matklad (approved):
Let's see how it goes!

I am still somewhat skeptical here overall, though its a bit hard to articulate why. So let's just see how this works out in practice!

PR #2257, opened 2024-08-29
vsr: correct assert

matklad:
It can be the case that:

* commit_min = checkpoint_trigger
* op_checkpoint = _pervious_ checkpoint
* commit_stage = .idle

This happens if we executed checkpoint trigger, advanced our commit_min, but then got stuck in the subsequent compaction.

Seed: ./zig/zig build  vopr -- --lite 2464843278033237343


sentientwaffle (approved):
Makes sense!

PR #2260, opened 2024-08-29
test: add integration tests for upgrades

matklad:
Adds a real multiversion binary to our integration tests, and implements a simplistic upgrade test. 

Ideally, this should be moved to our upcoming system-tests, but let's have at least a smoke test in the meantime: https://github.com/tigerbeetle/tigerbeetle/pull/2270


matklad:
Seems not too fail locally!

I _think_ this'll make our CI quite a bit slower, but I'll like to ignore that for now, and revisit some time later  hollistically, with the goal of making the overall CI faster.

.github/workflows/ci.yml
```diff
@@ -31,7 +32,7 @@ jobs:
       # Docker restricts io_uring by default, running unconfined then.
       options: --security-opt seccomp=unconfined
     steps:
-      - run: apk add -U git
+      - run: apk add -U git -U github-cli
```
cb22: I think the second `-U` is redundant.

build.zig
```diff
@@ -630,28 +649,67 @@ fn build_test(
...
+        .git_commit = "bee71e0000000000000000000000000000bee71e".*, // Beetle-hash!
+        .config_verify = true,
+        .config_release = "0.16.99",
+        .config_release_client_min = "0.15.3",
+        .config_log_level = .info,
+        .config_aof_recovery = false,
+        .tracer_backend = .none,
+        .hash_log_mode = .none,
+    });
+    const tigerbeetle_past = download_release(b, "latest", options.target, options.mode);
```
cb22: We do use `tigerbeetle_past` above for the same thing, but for multiversioning naming, I think `previous` works better when referring to a previous release. 

`past` internally refers to the [non-current releases](https://github.com/tigerbeetle/tigerbeetle/blob/main/src/multiversioning.zig#L373) within a multiversion release.

src/build_multiversion.zig
```diff
@@ -79,14 +79,9 @@ pub fn main() !void {
...
-        // When we fetch llvm-objcopy in build.zig, there isn't an easy way to mark it as
-        // executable, so do it here.
-        const fd = try shell.cwd.openFile(cli_args.llvm_objcopy, .{ .mode = .read_write });
-        defer fd.close();
-
-        try fd.chmod(0o777);
-    }
+    // When we fetch llvm-objcopy in build.zig, there isn't an easy way to mark it as
+    // executable, so do it here.
+    try shell.file_make_executable(cli_args.llvm_objcopy);
```
cb22, line 84: Could this be done with a system command `chmod +x` in `build_tigerbeetle_executable_get_objcopy`? I'm not sure if that's better... But it feels weird to have a build step that gets you the binary and you then need to make sure it's executable outside of that.

src/integration_tests.zig
```diff
@@ -289,3 +281,208 @@ test "help/version smoke" {
...
 }
+
+test "in-place upgrade" {
+    // Smoke test that in-place upgrades work.
+    //
+    // Stats a cluster of three replicas using the previous release of TigerBeetle and then replaces
+    // the binaries on disk with a new version.
+    //
+    // Against this upgrading cluster, we are running a benchmark load and check that it finishes
+    // with zero status.
```
cb22:
```suggestion
    // Against this upgrading cluster, we are running a benchmark load and checking that it finishes
    // with a zero status.
```

src/integration_tests.zig
```diff
@@ -289,3 +281,208 @@ test "help/version smoke" {
...
+        }
+
+        fn deinit(context: *Context) void {
+            for (&context.replicas) |*replica| {
+                if (replica.*) |*alive| {
+                    _ = alive.kill() catch {};
+                }
+            }
+
+            // context.shell.cwd.deleteTree(context.tmp) catch {};
```
cb22: Do we want to clear our tmp dir?

build.zig
matklad: ah, I got confused thinking that `_past` is the right name here, thanks!

src/build_multiversion.zig
matklad: I think the current solution is better than `chmod`, because system dependencies feel like a worse problem to me. But yeah, this _is_ pretty hacky.

The proper way to do that would be to write a custom `chmod`ing step, but that seems relatively large amount of code for little gain, given that the end-game is to not use vendored objcopy at all.

I'll write the custom step if you think thats worthwhile though!

src/integration_tests.zig
matklad: :facepalm:

cb22:
Looks like some merge conflicts with `tracer_backend`

src/build_multiversion.zig
cb22: > I'll write the custom step if you think thats worthwhile though!

I think it's OK as-is for now! We do want to get rid of that step anyhow at some point. (and if someone else tries to use it in the meantime, figuring out the error will hopefully discourage them :smile:.)

matklad:
should be good now!

cb22 (approved):

PR #2263, opened 2024-08-30
build: vendor llvm-objcopy

matklad:
We now depend on llvm-objcopy to make our multiversion binaries. Long term, we should probably upstream the capabilities we need to `zig objcopy` but for now "vendor" a specific copy of llvm-objcopy we use.

Vendoring is useful primarily because there isn't an easy way to install the tool in a cross-platform way! As we want to make multiversion a normal build, it is important that the build remains seemless and that the user doesn't have to manually install anything.

The solution is to upload specific versions of llvm-objectopy as a release on <https://github.com/tigerbeetle/dependencies> and use that.

Sadly, you have to trust me that the binaries I uploaded there are good, but the mac and windows versions are straight from LLVM upstream release, and the build of statically-linked linux version seems to be reproduciable.

Assuming the specific binaries are fine, any further risks are locked down by veryfying the checksum upon the download.

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -471,6 +471,10 @@ fn build_tigerbeetle_executable_multiversion(b: *std.Build, options: struct {
     build_multiversion_exe.root_module.addOptions("vsr_options", options.vsr_options);
 
     const build_multiversion = b.addRunArtifact(build_multiversion_exe);
+    build_multiversion.addPrefixedFileArg(
+        "--llvm-objcopy=",
+        build_tigerbeetle_executable_get_objcopy(b),
+    );
     if (options.target.result.os.tag == .macos) {
         build_multiversion.addArg("--target=macos");
         inline for (.{ "x86_64", "aarch64" }, .{ "x86-64", "aarch64" }) |arch, flag| {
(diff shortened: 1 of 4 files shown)


build.zig
```diff
@@ -521,6 +538,41 @@ fn build_tigerbeetle_executable_multiversion(b: *std.Build, options: struct {
     return build_multiversion.addPrefixedOutputFileArg("--output=", basename);
 }
 
+// Downloads a pre-build llvm-objcopy from <https://github.com/tigerbeetle/dependencies>.
+fn build_tigerbeetle_executable_get_objcopy(b: *std.Build) std.Build.LazyPath {
+    const llvm_objcopy_artifact: struct { name: []const u8, checksum: u256 } =
```
cb22, line 543: Hm - x86_64 macos and aarch64 Linux?

build.zig
```diff
@@ -521,6 +538,41 @@ fn build_tigerbeetle_executable_multiversion(b: *std.Build, options: struct {
...
+        switch (b.graph.host.result.os.tag) {
+        .linux => .{
+            .name = "llvm-objcopy-x86_64-linux",
+            .checksum = 0x3e2fe8f359c63eb62069e322f9bc079b2876301510afb15f70d117b30a2eea36,
```
cb22, line 547: I did get a different checksum when building for Linux with your instructions [here](https://github.com/tigerbeetle/dependencies), but it seemed to be repeatable on my machine :thinking:.

Can you try build again, and see what checksum you get?

(mine was `925d140798f45f53878d47c8d2e9548893b71ac57edbe28f43c1ffa2aa0dc7d3  llvm-objcopy`)

build.zig
matklad: Intentionally not including those. You can build multiversion TigerBeetle _for_ these tripples, but you have to build _from_ one of our supported three host triples

matklad: (to clarify, won't be a huge problem to add those, but I'd rather do that lazily, once we need. There's a chance that we'll just rewrite-it-in-zig faster)

build.zig
matklad: Huh, 

```
λ sha256sum llvm-objcopy
3e2fe8f359c63eb62069e322f9bc079b2876301510afb15f70d117b30a2eea36  llvm-objcopy
```

Let me try re-building this after removing my local containers, maybe docker re-uses old ubuntu:latest?

matklad: Nope, still 3e2fe8f359c63eb62069e322f9bc079b2876301510afb15f70d117b30a2eea36

matklad: Trying this command to pin the image now:

```
λ docker run -it --name llvm_build ubuntu:noble-20240801 /bin/bash -c "
      apt-get update && apt-get install -y git cmake ninja-build clang
      git clone --depth 1 --branch release/18.x https://github.com/llvm/llvm-project.git
      mkdir llvm-project/build && cd llvm-project/build
      cmake -G Ninja -DLLVM_ENABLE_PROJECTS=clang -DCMAKE_BUILD_TYPE=Release -DLLVM_TARGETS_TO_BUILD=X86 -DLLVM_BUILD_STATIC=ON ../llvm
      ninja llvm-objcopy
  "

  docker cp llvm_build:/llvm-project/build/bin/llvm-objcopy .
  docker rm llvm_build
```

matklad: Huh, I get that hash with `ubuntu:noble-20240801` and with `ubuntu:noble-20240530`!

matklad: @cb22 tried this on a different laptop, and I still get 

```
λ sha256sum llvm-objcopy
3e2fe8f359c63eb62069e322f9bc079b2876301510afb15f70d117b30a2eea36  llvm-objcopy
```

cb22: It looks very likely it was my `ubuntu:latest` that was very out of date! Could have sworn I saw it pulling, though! Will confirm output from the pinned version in a second.

cb22: Woo

```
/tmp> sha256sum llvm-objcopy                                                      539.147s 15:59
3e2fe8f359c63eb62069e322f9bc079b2876301510afb15f70d117b30a2eea36  llvm-objcopy
```

cb22 (approved):

build.zig
matklad: Pinned the version upstream:

https://github.com/tigerbeetle/dependencies/commit/983da8decafdb1616dc3695490dcc6d5939d9eb2

PR #2266, opened 2024-08-30
Lightweight Tracing

sentientwaffle:
### Overview

Log IO/CPU event spans for analysis/visualization.

Unlike tracy, this approach doesn't require linking libc/libcpp or any third-party libraries.
So we can compile tracing in by default, and optionally enable it at runtime -- even for production release builds.

The trace JSON output is compatible with:

- <https://ui.perfetto.dev/>
- <https://gravitymoth.com/spall/spall.html>
- <chrome://tracing/>

### Usage

```
$ ./tigerbeetle start --experimental --trace=trace.json
```

Then upload the `trace.json` file to one of the trace visualization tools.

### Examples

Here is an [example trace](https://gist.github.com/sentientwaffle/c734797cfa0dbcb0aa6f1c6759699376), captured while running the benchmark:

**Perfetto:**

perfetto

**Spall:**

spall

### Notes

- When enabled, traces are written to the specified file as JSON.
  (In contrast to `log` statements, which are written to stderr).
- Event pairing (`trace.start`/`trace.stop` (or `trace.start`/`trace.reset`)) is asserted at runtime.
- `trace.start()/.stop()/.reset()` will `log.debug()` regardless of whether tracing is enabled.
- Eventually, we could choose parse traces from structured logs and convert them to JSON with a separate tool, but for the time being this is simpler.

<details>
<summary>API Design notes</summary>

One goal was to keep the integration API lightweight.

The original implementation was a global, but I ended up threading a single trace 
- Threading this is necessary since `Trace` tracks the enabled/disabled state of each event. This in turns allows the fuzzers to assert that tracing is used correctly.
- In the future, threading might come in handy for the structured logger, as we could attach fields like `replica_index` and `tree_name` so that every log statement doesn't need to repeat them.

Unlike most tracing APIs, events aren't ever "stacked". (See https://gravitymoth.com/spall/spall.html for an example).
Rather, events are independent -- they can be interleaved.

This keeps the API simple. Stacks are useful when doing fine-grained analysis (like CPU profiling) -- but for that, linux perf would suffice. Tracing is useful for more coarse-grained analysis (async-IO + CPU rather than just CPU).

The first parameter of each trace function is typed (an enum rather than a string) so that we can match start/end events together. I don't think registering events like this is onerous -- I don't expect us to end up with more than a couple dozen unique event types.

</details>

<details>
<summary>Tool overview</summary>

[Spall](https://gravitymoth.com/spall/spall.html) has excellent mouse navigation, but no keybindings.

[Perfetto](https://ui.perfetto.dev/) has keybindings and extensive (albeit opaque) support for querying events. (e.g. "how much cumulative time is spent in event X?").

I didn't check Chrome other than to test that it worked with our format -- I think it is like Perfetto but with fewer features.

Additionally, Perfetto + Chrome support string labels for threads, which would be very nice to use (since thread numbers are meaningless), but Spall errors out on string thread ids.

</details>

### DIFF
--- a/src/state_machine.zig
+++ b/src/state_machine.zig
@@ -3026,6 +3026,7 @@ const TestContext = struct {
     const message_body_size_max = 64 * @max(@sizeOf(Account), @sizeOf(Transfer));
 
     storage: Storage,
+    trace: vsr.trace.Tracer,
     superblock: SuperBlock,
     grid: Grid,
     state_machine: StateMachine,
@@ -3044,6 +3045,9 @@ const TestContext = struct {
         );
         errdefer ctx.storage.deinit(allocator);
 
+        ctx.trace = try vsr.trace.Tracer.init(allocator, 0, .{});
+        errdefer ctx.trace.deinit(allocator);
+
         ctx.superblock = try SuperBlock.init(allocator, .{
             .storage = &ctx.storage,
             .storage_size_limit = data_file_size_min,
@@ -3056,6 +3060,7 @@ const TestContext = struct {
 
         ctx.grid = try Grid.init(allocator, .{
             .superblock = &ctx.superblock,
+            .trace = &ctx.trace,
             .missing_blocks_max = 0,
             .missing_tables_max = 0,
         });
@@ -3075,10 +3080,11 @@ const TestContext = struct {
     }
 
     fn deinit(ctx: *TestContext, allocator: mem.Allocator) void {
-        ctx.storage.deinit(allocator);
-        ctx.superblock.deinit(allocator);
-        ctx.grid.deinit(allocator);
         ctx.state_machine.deinit(allocator);
+        ctx.grid.deinit(allocator);
+        ctx.superblock.deinit(allocator);
+        ctx.trace.deinit(allocator);
+        ctx.storage.deinit(allocator);
         ctx.* = undefined;
     }
 
--- a/src/tigerbeetle/main.zig
+++ b/src/tigerbeetle/main.zig
@@ -349,6 +349,8 @@ const Command = struct {
 
         const clients_limit = constants.pipeline_prepare_queue_max + args.pipeline_requests_limit;
 
+        var stdout_writer = std.io.getStdOut().writer();
+
         var replica: Replica = undefined;
         replica.open(allocator, .{
             .node_count = args.addresses.count_as(u8),
@@ -379,9 +381,7 @@ const Command = struct {
                 .clients_limit = clients_limit,
             },
             .grid_cache_blocks_count = args.cache_grid_blocks,
-            .tracer_options = .{
-                .writer = if (args.trace) std.io.getStdOut().writer().any() else null,
-            },
+            .tracer_options = .{ .writer = if (args.trace) stdout_writer.any() else null },
         }) catch |err| switch (err) {
             error.NoAddress => fatal("all --addresses must be provided", .{}),
             else => |e| return e,
--- a/src/tigerbeetle/main.zig
+++ b/src/tigerbeetle/main.zig
@@ -349,7 +349,15 @@ const Command = struct {
 
         const clients_limit = constants.pipeline_prepare_queue_max + args.pipeline_requests_limit;
 
-        var stdout_writer = std.io.getStdOut().writer();
+        const trace_file = if (args.trace) |trace_path|
+            std.fs.cwd().createFile(trace_path, .{ .exclusive = true }) catch |err| {
+                std.debug.panic("error creating trace file: {}", .{err});
+            }
+        else
+            null;
+        defer if (trace_file) |file| file.close();
+
+        const trace_writer = if (trace_file) |file| file.writer() else null;
 
         var replica: Replica = undefined;
         replica.open(allocator, .{
@@ -381,7 +389,7 @@ const Command = struct {
                 .clients_limit = clients_limit,
             },
             .grid_cache_blocks_count = args.cache_grid_blocks,
-            .tracer_options = .{ .writer = if (args.trace) stdout_writer.any() else null },
+            .tracer_options = .{ .writer = if (trace_writer) |writer| writer.any() else null },
         }) catch |err| switch (err) {
             error.NoAddress => fatal("all --addresses must be provided", .{}),
             else => |e| return e,
--- a/src/tigerbeetle/main.zig
+++ b/src/tigerbeetle/main.zig
@@ -379,6 +379,9 @@ const Command = struct {
                 .clients_limit = clients_limit,
             },
             .grid_cache_blocks_count = args.cache_grid_blocks,
+            .tracer_options = .{
+                .writer = if (args.trace) std.io.getStdOut().writer().any() else null,
+            },
         }) catch |err| switch (err) {
             error.NoAddress => fatal("all --addresses must be provided", .{}),
             else => |e| return e,
--- a/src/trace.zig
+++ b/src/trace.zig
@@ -93,6 +93,10 @@ const std = @import("std");
 const assert = std.debug.assert;
 const log = std.log.scoped(.trace);
 
+const constants = @import("constants.zig");
+
+const trace_span_size_max = 1024;
+
 // TODO This could be changed to a union(enum) if variable-cardinality events are needed (for
 // example, an event per grid IOP). (`stack()` would need to be updated as well).
 pub const Event = enum {
@@ -115,8 +119,9 @@ pub const Event = enum {
 pub const Tracer = struct {
     replica_index: u8,
     options: Options,
+    buffer: []u8,
 
-    events_enabled: [Event.stack_count]bool = [_]bool{false} ** Event.stack_count,
+    events_enabled: [Event.stack_count]bool = .{false} ** Event.stack_count,
     time_start: std.time.Instant,
 
     pub const Options = struct {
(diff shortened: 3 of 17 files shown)


src/trace.zig
```diff
@@ -0,0 +1,338 @@
...
+//!         // events usefully.)
+//!         "name": "replica_commit stage='next_pipeline'",
+//!
+//!         // Extra event arguments. (Encoded from the second argument to `trace.start()`).
+//!         "args": {
+//!             "stage": "next_pipeline",
+//!             "op": 1
+//!         },
+//!     },
+//!
```
matklad, line 91: Huge fan of top-of-the file comments, thanks!

src/trace.zig
```diff
@@ -0,0 +1,338 @@
...
+    fn stack(event: *const Event) u32 {
+        return @intFromEnum(event.*);
+    }
+};
+
+pub const Tracer = struct {
+    replica_index: u8,
+    options: Options,
+
+    events_enabled: [Event.stack_count]bool = [_]bool{false} ** Event.stack_count,
```
matklad: Just `.{false} ** Event.stack_count` should work I think?

src/trace.zig
```diff
@@ -0,0 +1,338 @@
...
+    pub const Options = struct {
+        /// The tracer still validates start/stop state even when writer=null.
+        writer: ?std.io.AnyWriter = null,
+    };
+
+    pub fn init(allocator: std.mem.Allocator, replica_index: u8, options: Options) !Tracer {
+        _ = allocator;
+
+        if (options.writer) |writer| {
+            try writer.writeAll("[\n");
```
matklad, line 134: This is to make the result a valid JSON document? Is this required by any of the tools consuming the result?

From the perspective of overall simplicity, I much prefer just json-value-per-line format:

https://jsonlines.org

Parsing that is _much_ easier, as it doesn't require a streaming JSON parser, but just the ability to frame by `\n`, and the normal one-shot JSON parser. 

This of course puts a constraint on the consumer to disallow interior newlines, but that's what we want to do anyway!

src/trace.zig
```diff
@@ -0,0 +1,338 @@
...
+    fn start_write_data(writer: std.io.AnyWriter, data: anytype) !void {
+        comptime assert(@typeInfo(@TypeOf(data)) == .Struct);
+
+        inline for (std.meta.fields(@TypeOf(data))) |data_field| {
+            const data_field_value = @field(data, data_field.name);
+            try writer.writeByte(' ');
+            try writer.writeAll(data_field.name);
+            try writer.writeByte('=');
+            if (data_field.type == []const u8) {
+                try writer.print("'{s}'", .{data_field_value});
```
matklad: Do we need to assert that the output doesn't contain any funnybytes and is not to long? 

I'd maybe do at least a length check here

src/trace.zig
matklad: And is it intentional that we never write `]`? i think it is, but worth a comment!

src/trace.zig
```diff
@@ -0,0 +1,338 @@
...
+            .category = @tagName(event),
+            .event = 'B',
+            .timestamp = time_elapsed_us,
+            .name = @tagName(event),
+            .data = struct_format(data, .sparse),
+            .args = std.json.Formatter(@TypeOf(data)){ .value = data, .options = .{} },
+        }) catch unreachable;
+    }
+
+    fn start_write_data(writer: std.io.AnyWriter, data: anytype) !void {
```
matklad: This seems to never be called?

src/trace.zig
```diff
@@ -0,0 +1,338 @@
...
+            "}},\n", .{
+            .process_id = tracer.replica_index,
+            .thread_id = event.stack(),
+            .category = @tagName(event),
+            .event = 'B',
+            .timestamp = time_elapsed_us,
+            .name = @tagName(event),
+            .data = struct_format(data, .sparse),
+            .args = std.json.Formatter(@TypeOf(data)){ .value = data, .options = .{} },
+        }) catch unreachable;
```
matklad, line 192: Do we want some `constants.verify` checks that each event is short enough?

src/tigerbeetle/main.zig
```diff
@@ -379,6 +379,9 @@ const Command = struct {
                 .clients_limit = clients_limit,
             },
             .grid_cache_blocks_count = args.cache_grid_blocks,
+            .tracer_options = .{
+                .writer = if (args.trace) std.io.getStdOut().writer().any() else null,
```
matklad: * aren't we closing over a pointer to a temporary here again? 
* the stdout _is_ buffered, right?

src/state_machine.zig
```diff
@@ -3075,10 +3080,11 @@ const TestContext = struct {
...
 
     fn deinit(ctx: *TestContext, allocator: mem.Allocator) void {
-        ctx.storage.deinit(allocator);
-        ctx.superblock.deinit(allocator);
-        ctx.grid.deinit(allocator);
         ctx.state_machine.deinit(allocator);
+        ctx.grid.deinit(allocator);
+        ctx.superblock.deinit(allocator);
+        ctx.trace.deinit(allocator);
+        ctx.storage.deinit(allocator);
```
matklad, line 3087: :heart:

src/trace.zig
```diff
@@ -0,0 +1,338 @@
...
+                    if (data_field.type != bool and
+                        data_field.type != u8 and
+                        data_field.type != []const u8 and
+                        data_field.type != [:0]const u8 and
+                        @typeInfo(data_field.type) != .Enum and
+                        @typeInfo(data_field.type) != .Union)
+                    {
+                        break;
+                    }
+                }
```
matklad, line 285: Wow, that's a smart way to pick fields to include with the name automatically!

src/trace.zig
sentientwaffle: Unfortunately, the open bracket is required for the tools to parse the JSON. The close bracket is not required, which is nice since we don't need to emit it before shutting down. There is comment at the start of the file:

```
//! - The JSON output is a "[" followed by a comma-separated list of JSON objects. The JSON array is
//!   never closed with a "]", but Chrome, Spall, and Perfetto all handle this.
```

src/trace.zig
sentientwaffle: I'm not sure what you mean by an event being short enough.

matklad: That ` .args = std.json.Formatter(@TypeOf(data)){ .value = data, .options = .{} },` does't result in too large JSON output. 

That is to prevent outright mistakes like accidentally including the entire block into a message, but also to prevent boiling frog type of situation, where, for some event, we just add more and more fields, until it becomes hard to analyse. 

I guess, one background thought here is that perhaps we could format the even into a fixed-sized internal buffer first, and then dump it to write as one single slice? Potentially this could be a bit faster I think.

sentientwaffle: Ah that makes sense, thanks!

> we could format the even into a fixed-sized internal buffer first, and then dump it to write as one single slice? Potentially this could be a bit faster I think.

I think this is what `BufferedWriter` does, which I need to add anyhow.

matklad: I was goign to say that we can avoid BufferedWriter and do a single `fmt.bufPrint`, but it uses the writer internally anyway: 

```
pub fn bufPrint(buf: []u8, comptime fmt: []const u8, args: anytype) BufPrintError![]u8 {
    var fbs = std.io.fixedBufferStream(buf);
    format(fbs.writer().any(), fmt, args) catch |err| switch (err) {
        error.NoSpaceLeft => return error.NoSpaceLeft,
        else => unreachable,
    };
    return fbs.getWritten();
}
```

I wonder if at some point we should do `fmt` which also tracks the worst-case size of the output, so that we statically check that the size of buffer is big enough? Sounds like a bigger adventure though!

src/tigerbeetle/main.zig
```diff
@@ -379,6 +381,7 @@ const Command = struct {
                 .clients_limit = clients_limit,
             },
             .grid_cache_blocks_count = args.cache_grid_blocks,
+            .tracer_options = .{ .writer = if (args.trace) stdout_writer.any() else null },
```
batiati: Suggestion: Redirecting the log to `stdout` is a great default, but I think we could also support a file path.

For instance: 
- Currently, we cannot redirect the benchmark's stdout to a file because it is already used to capture the database port.
- Also, the process may not run inside a shell that conveniently offers the ability to pipe the stdout to a file. 
- Or even for enabling trace for client-side events.

src/tigerbeetle/main.zig
```diff
@@ -349,6 +349,16 @@ const Command = struct {
 
         const clients_limit = constants.pipeline_prepare_queue_max + args.pipeline_requests_limit;
 
+        const trace_file = if (args.trace) |trace_path|
+            std.fs.cwd().createFile(trace_path, .{ .exclusive = true }) catch |err| {
+                std.debug.panic("error creating trace file: {}", .{err});
+            }
+        else
+            null;
+        errdefer if (trace_file) |file| file.close();
```
batiati: Should we always close this file at the end? (although we don't have a graceful shutdown to exit `main` :thinking:)
```suggestion
        defer if (trace_file) |file| file.close();
```

src/trace.zig
cb22: > I guess, one background thought here is that perhaps we could format the even into a fixed-sized internal buffer first, and then dump it to write as one single slice? Potentially this could be a bit faster I think.

One thought with https://github.com/tigerbeetle/tigerbeetle/pull/2266/commits/edcc13f93b2ff9b7625e94ebf2c35861579b3282 is to make sure that we're not affected by the `write()` to the file blocking. Probably good enough for now to use a path on tmpfs as the target.

sentientwaffle: > Probably good enough for now to use a path on tmpfs as the target.

That, and tracing is behind `--experimental`. Long-term we should probably use io_uring for logging (which will include tracing).

batiati (approved):

PR #2268, opened 2024-09-02
changelog: reference test and make `Released:` formatting consistent

cb22:
Also, fixes the case where during the release process itself, `--no-changelog` shouldn't be passed in.

### DIFF
--- a/src/scripts/changelog.zig
+++ b/src/scripts/changelog.zig
@@ -5,6 +5,7 @@ const stdx = @import("../stdx.zig");
 const Shell = @import("../shell.zig");
 
 const Release = @import("../multiversioning.zig").Release;
+const ReleaseTriple = @import("../multiversioning.zig").ReleaseTriple;
 
 const log = std.log;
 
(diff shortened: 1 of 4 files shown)


src/scripts/changelog.zig
```diff
@@ -268,6 +272,7 @@ test ChangelogIterator {
         \\
     );
     try std.testing.expectEqualStrings(entry.text_body,
+        \\
```
cb22, line 275: We probably don't want this leading newline?

src/scripts/devhub.zig
```diff
@@ -34,10 +36,41 @@ pub fn main(shell: *Shell, gpa: std.mem.Allocator, cli_args: CliArgs) !void {
...
+        try shell.zig(
+            \\build scripts -- release --build --no-changelog --sha={sha}
+            \\    --language=zig
+        , .{ .sha = cli_args.sha });
+    } else {
+        try shell.zig(
+            \\build scripts -- release --build --sha={sha}
+            \\    --language=zig
+        , .{ .sha = cli_args.sha });
+    }
```
cb22, line 73: Unfortunately passing in `` as a value for a `{no_changelog_flag}` doesn't work :smile:

batiati: Passing `--no-changelog=true|false` didn't work?

src/scripts/changelog.zig
batiati: I guess it's there just to stress the parser

batiati (approved):

src/scripts/devhub.zig
cb22: Nope:

```
error: --no-changelog: argument does not require a value in '--no-changelog=false'
```

PR #2269, opened 2024-09-03
vsr: sync sync docs with the new sync protocol

matklad:


chaitanyabhandari (approved):
LGTM; we went from 10 → 5 steps in the state sync algorithm, this simplification is ❤️!

PR #1187, opened 2023-09-11
Antithesis: Pull in from old repo

cb22:
This PR pulls in and updates our Antithesis support from the previous repo, as well as hooks it up to CI.

PR #2270, opened 2024-09-03
test: System test with Java client

owickstrom:
This adds a new `systest`, which is comprised of a workload generating commands that are sent to a cluster of replicas. It's based on #1187 but focuses on the Java client and a simpler architecture. It's not currently integrated with CI or anything like that. We're trigger it manually while we're getting the Antithesis setup in order.


src/scripts/antithesis.zig
```diff
@@ -0,0 +1,241 @@
+//! Deployment script for our systest (src/testing/systest).
+//!
+//! * Builds the Java client and the associated workload using Maven
+//! * Builds Docker images for the workload, replicas, and config
+//! * Optionally pushes the images to the Antithesis registry
+//!
+//! Currently there's no support for triggering tests with this script.
```
matklad, line 7: :+1: top-of-file context-explaining comments are golden!

src/scripts/antithesis.zig
```diff
@@ -0,0 +1,241 @@
...
+};
+
+const Image = enum { config, workload, replica };
+
+pub fn main(shell: *Shell, gpa: std.mem.Allocator, cli_args: CLIArgs) !void {
+    assert(try shell.exec_status_ok("docker --version", .{}));
+
+    // Docker tag to build and push
+    const tag = std.mem.trim(u8, cli_args.tag, &std.ascii.whitespace);
+    assert(tag.len > 0);
```
matklad: I'd say rather than trimming, it would be more tiger-style to assert that there aren't whitespace.

src/scripts/antithesis.zig
```diff
@@ -0,0 +1,241 @@
...
+    push: bool = false,
+};
+
+const Image = enum { config, workload, replica };
+
+pub fn main(shell: *Shell, gpa: std.mem.Allocator, cli_args: CLIArgs) !void {
+    assert(try shell.exec_status_ok("docker --version", .{}));
+
+    // Docker tag to build and push
+    const tag = std.mem.trim(u8, cli_args.tag, &std.ascii.whitespace);
```
matklad: :exploding_head: I didn't know that `std.ascii.whitespace` is a thing, thanks!

src/scripts/antithesis.zig
```diff
@@ -0,0 +1,241 @@
...
+
+fn build_image(
+    shell: *Shell,
+    gpa: std.mem.Allocator,
+    comptime image: Image,
+    tag: []const u8,
+) !void {
+    switch (image) {
+        .config => {
+            const image_dir = try shell.exec_stdout("mktemp -d", .{});
```
matklad: Rather than shelling out to mktemp, let's lift this piece of code to shell.zig:

https://github.com/tigerbeetle/tigerbeetle/pull/2260/files#diff-dee49dfa908b58ba4278b1d022ce0726c79da00b3a3a87d93fa0619e7836d46bR321


There's also `std.testing.TmpDir`, but it's awkwrard, as it doesn't expose the full path.

src/scripts/antithesis.zig
```diff
@@ -0,0 +1,241 @@
...
+
+            try shell.pushd(image_dir);
+            defer shell.popd();
+
+            // TODO(owickstrom): remove the need for .env file by rendering docker-compose.yaml
+            // with the correct tag?
+            const env_file = try std.fs.path.join(gpa, &.{ image_dir, ".env" });
+            defer gpa.free(env_file);
```
matklad: You can use `shell.fmt` to avoid the need to free individual strings. I am very distrustfull of code that manages individual strings, it always leaks eventually.

matklad: I am on the fence on whether we need `shell.join` as well -- the `/` separator _does_ work on windows, so I don't see problems with raw string formatting personally.

src/scripts/antithesis.zig
```diff
@@ -0,0 +1,241 @@
...
+
+            const env_file_contents = try std.fmt.allocPrint(gpa, "TAG={s}", .{tag});
```
matklad: shell.fmt

matklad (commented):
First batch of comments, will finish review later today!

src/scripts/antithesis.zig
owickstrom: OK, let's go with `shell.fmt` there too, then.

src/scripts/antithesis.zig
```diff
@@ -0,0 +1,236 @@
...
+    ,
+
+    .replica =
+    \\FROM debian:stable-slim
+    \\WORKDIR /opt/tigerbeetle
+    \\
+    \\COPY zig-out/bin/tigerbeetle ./tigerbeetle
+    \\COPY src/testing/systest/scripts/run.sh ./run.sh
+    \\
+    \\ENTRYPOINT ["./run.sh"]
```
matklad: Can we inline the script as well?

src/scripts/antithesis.zig
```diff
@@ -0,0 +1,236 @@
...
+
+            const env_file_contents = try shell.fmt("TAG={s}", .{tag});
+            _ = try shell.file_ensure_content(env_file, env_file_contents);
+
+            const docker_compose_file = try shell.fmt("{s}/docker-compose.yaml", .{
+                image_dir,
+            });
+            _ = try shell.file_ensure_content(docker_compose_file, docker_compose_contents);
+
+            try shell.exec("mkdir -p volumes/database", .{});
```
matklad: there's `Dir.makePath`

In general, don't gratitiously shell out to posix utilities, minimize OS dependencies. Zig stdlib is our interface to OS facilities.

matklad (commented):
LGTM overall, let's just avoid introducing dependency on coreutils here!

src/scripts/antithesis.zig
owickstrom: It's a bit messy, but if I can embed it only in the Zig source, not the Docker source inside the Zig source, it might be justifiable. I'll give it a go.

src/shell.zig
```diff
@@ -275,6 +275,21 @@ pub fn file_ensure_content(
     return .updated;
 }
 
+/// Creates a new temporary directory (in the project-level .zig-cache) and returns the
+/// absolute path.
+pub fn create_tmp_dir(
```
matklad:
```
/// caller is responsible for cleaning the directory up
```

and `defer shell.cwd.deleteTree(image_dir)`

src/scripts/antithesis.zig
```diff
@@ -0,0 +1,295 @@
...
+    comptime image: Image,
+    tag: []const u8,
+) !void {
+    switch (image) {
+        .config => {
+            // This is the temporary directory where we'll assemble the config docker image.
+            // By design it's not automatically deleted with a defer, because the caller of
+            // the script might want to debug the docker compose setup locally.
+            const image_dir = try shell.create_tmp_dir();
+
```
matklad, line 72: defer deleteTree missing I think?

src/tidy.zig
```diff
@@ -519,7 +519,7 @@ test "tidy extensions" {
         .{"zig/download.sh"},                     .{"src/scripts/cfo_supervisor.sh"},
         .{"src/docs_website/scripts/build.sh"},   .{".github/ci/docs_check.sh"},
         .{".github/ci/test_aof.sh"},              .{"tools/systemd/tigerbeetle-pre-start.sh"},
-        .{"tools/vscode/format_debug_server.sh"},
```
matklad: can be reverted now I think

src/scripts/antithesis.zig
owickstrom: Added a comment above about why we keep it. Maybe it should be a CLI flag?

matklad: Ah, I think it's ok to not clean up in that case, but let's log the path so that the user knows that to debug. Basically, move parts of the comment into a log message!

matklad (commented):
LGTM, but needs a rebase!

src/scripts/antithesis.zig
owickstrom: Done!

owickstrom:
@matklad thanks! Final touches applied, I think, and just waiting for CI.

matklad (approved):
:rocket: 

(as a reminder, its on PR's author to click "merge when ready" after review is green (but you don't need to wait for CI to be green, merge queue handles that))

Issue #2102, opened 2024-07-11
Prefetch accounts directly in `prefetch_create_transfers`


brson:
Per @matklad on ironbeetle.

In `prefetch_create_transfers` we prefetch a bunch of transfers, then later in `prefetch_create_transfers_callback_transfers` prefetch their associated accounts. It should be possible though to prefetch these accounts directly in `prefetch_create_transfers` instead of waiting for the transfer prefetch callback.

wpaulino:
I'm interested in tackling this. This can also be the case with `prefetch_create_accounts`. I had a look through the prefetching logic and noticed the reason for this is because, at the moment, the state machine only handles prefetching from one groove at a time. I'd like to explore an approach where we maintain a single `PrefetchContext` (and all other associated contexts) as we do today (so all of the existing concurrency limits remain) that supports parallel multi-groove lookups. Are there any preliminary objections/concerns I should be aware of before going down this path?

PR #2271, opened 2024-09-03
Concurrent prefetch

batiati:
Allows multiple grooves to prefetch simultaneously.

While each groove's prefetch already tries to use all available IO (by employing multiple workers), some situations may benefit from more concurrency, especially when creating transfers, where, in most cases, we can prefetch transfers and accounts in one go.

Related https://github.com/tigerbeetle/tigerbeetle/issues/2102

### DIFF
--- a/src/lsm/groove.zig
+++ b/src/lsm/groove.zig
@@ -472,7 +472,7 @@ pub fn GrooveType(
             self.* = undefined;
         }
 
-        fn reset(self: *TimestampSet) void {
+        pub fn reset(self: *TimestampSet) void {
             self.map.clearRetainingCapacity();
         }
 
@@ -757,8 +757,6 @@ pub fn GrooveType(
 
             groove.prefetch_snapshot = snapshot_target;
             assert(groove.prefetch_keys.count() == 0);
-
-            if (has_id) groove.timestamps.reset();
         }
 
         /// This must be called by the state machine for every key to be prefetched.
--- a/src/state_machine.zig
+++ b/src/state_machine.zig
@@ -393,35 +393,43 @@ pub fn StateMachineType(
             cache_entries_account_balances: u32,
         };
 
-        /// Since prefetch contexts are used one at a time, it's safe to access
-        /// the union's fields and reuse the same memory for all context instances.
-        const PrefetchContext = union(enum) {
-            null,
-            accounts: AccountsGroove.PrefetchContext,
-            transfers: TransfersGroove.PrefetchContext,
-            transfers_pending: TransfersPendingGroove.PrefetchContext,
+        const PrefetchContext = struct {
+            const Fields = struct {
+                accounts: AccountsGroove.PrefetchContext,
+                transfers: TransfersGroove.PrefetchContext,
+                transfers_pending: TransfersPendingGroove.PrefetchContext,
+            };
+            const Tag = std.meta.FieldEnum(Fields);
+            const EnumSet = std.EnumSet(Tag);
 
-            pub const Field = std.meta.FieldEnum(PrefetchContext);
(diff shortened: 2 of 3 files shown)


src/lsm/groove.zig
```diff
@@ -472,7 +472,7 @@ pub fn GrooveType(
             self.* = undefined;
         }
 
-        fn reset(self: *TimestampSet) void {
+        pub fn reset(self: *TimestampSet) void {
```
matklad, line 475: Oh, I missed the introduciton of `TimestampSet`. Could you add a comment on TimestampSet type or `timestamps` field explaining why we need it? 

And maybe call this function `clear` rather than `reset`? I _think_ we consistently use `reset` on the cancellation path, with the semantics that it could be called in the midst of things happening, and should clear any inprogerss work.

The way we use this function though, it seems like we always call it in well-defined place, to just clear stuff out, not to cancel any in-progress prefetches?

src/state_machine.zig
```diff
@@ -393,35 +393,43 @@ pub fn StateMachineType(
...
-            pub fn FieldType(comptime field: Field) type {
-                return std.meta.fieldInfo(PrefetchContext, field).type;
+        const PrefetchContext = struct {
+            const Fields = struct {
+                accounts: AccountsGroove.PrefetchContext,
+                transfers: TransfersGroove.PrefetchContext,
+                transfers_pending: TransfersPendingGroove.PrefetchContext,
+            };
+            const Tag = std.meta.FieldEnum(Fields);
+            const EnumSet = std.EnumSet(Tag);
```
matklad, line 403: If we are introducing a name here, could we also up the abstraction layer here?

```
used: FieldSet
```

matklad: Hm, wait, I think there's a bigger picture here!

The smell right now is that the new PrefetchContext encapsulates "type dispatch" logic, where we do something meta to get at accounts/transfers/pending. But we already have this logic encoded in the Forest!

So can we sink `PrefetchContext` to be a field of `Groove` (I think we can statically `void` the fild if `prefetch_entries_for_read_max == 0`, like it is for accounts_balances groove)?

I think this _wasn't_ done before exactly to take advantage of this union memory sharding, but we can't do this anymore (and it seems like there isn't actually that much memory to share here?).

However, there's one thing which we _must_ share --- 

```
workers: [Grid.read_iops_max]PrefetchWorker = undefined,
```

These read_iops must be shared across all concurrent prefetches. 

Well, maybe not literally the workers, because workers are type-specific, but the underlying Grid.Reads definietly should be shared.

So it seems like we should keep reads at the state machine level (or maybe Forest levl, for better encapsulation), but sink the rest of prefetch context down to individual grooves?

matklad: (to clarify, the sinking types part here is feng-shui, and maybe we don't actually need that. But sharing the pool of reads across all prefetches is observrable behavior, and something I think we should tackle)

batiati: > Hm, wait, I think there's a bigger picture here!

Yes, I have plans to unify prefetch lookups and [ScanLookupType](https://github.com/tigerbeetle/tigerbeetle/blob/bc4a71bc5fa93de1f33403b0ea1d7a1f176e565e/src/lsm/scan_lookup.zig#L22), and the idea of a per-forest pool of resources is pretty much what I had in mind.

sentientwaffle:
Could you include a before/after benchmark for this?

batiati:
> Could you include a before/after benchmark for this?

No difference in before/after results, just noise. I suppose that even parallelizing we are still queuing read IOs (since workers are per groove)

batiati:
Closing it as "obsolete".

PR #2272, opened 2024-09-03
 vsr: remove obsolete sync_view field 

matklad:
This was a part of the old state sync protocol, where the role of the
field was to communicate in which view the sync target checkpoint was
committed.

Today, this information is tracked explicitly in an SV message which
allows us to atomically:

* advance our view
* install WAL consistent with the view
* install the checkpoint consistent with the WAL

So, we can remove `sync_view`. Sadly, that is a part of our superblock,
so we still need to have _something_ on disk there. My plan is to mark
this field `reserved` and assert that it is zeroed, but, before doing
that, lets explicitly zero it out. With all probability, it is already
zeroed everywhere in practice, but it is theoretically possible that a
replica went through several upgrades while not finishing state sync, so
let's lean on the safe side here.


src/vsr/superblock.zig
```diff
@@ -879,7 +874,7 @@ pub fn SuperBlockType(comptime Storage: type) type {
             vsr_state.commit_max = update.commit_max;
             vsr_state.sync_op_min = update.sync_op_min;
             vsr_state.sync_op_max = update.sync_op_max;
-            vsr_state.sync_view = update.sync_view;
```
matklad, line 882: Crucially, this is part of vsr_state, but not checkpoint_state. So, this is replica-local state, and we can change the rules for updating this without breaking compatibility across versions

sentientwaffle (approved):

matklad:
vopr's green, merging!

PR #2273, opened 2024-09-03
vsr/journal: remove machinery for explicit header sector locks, use lock_sectors instead

chaitanyabhandari:
Currently, while writing prepare headers, we rely on explicit header sector locking using [write_prepare_on_lock_header_sector](https://github.com/tigerbeetle/tigerbeetle/blob/bc4a71bc5fa93de1f33403b0ea1d7a1f176e565e/src/vsr/journal.zig#L1862) & [write_prepare_unlock_header_sector](https://github.com/tigerbeetle/tigerbeetle/blob/bc4a71bc5fa93de1f33403b0ea1d7a1f176e565e/src/vsr/journal.zig#L1931). This is independent from the machinery used to lock sectors while writing prepares, which uses [lock_sectors](https://github.com/tigerbeetle/tigerbeetle/blob/bc4a71bc5fa93de1f33403b0ea1d7a1f176e565e/src/vsr/journal.zig#L2014) &  [write_sectors_on_write](https://github.com/tigerbeetle/tigerbeetle/blob/bc4a71bc5fa93de1f33403b0ea1d7a1f176e565e/src/vsr/journal.zig#L2063), likely because of this optimization:
https://github.com/tigerbeetle/tigerbeetle/blob/bc4a71bc5fa93de1f33403b0ea1d7a1f176e565e/src/vsr/journal.zig#L1935-L1944

This PR aims to simplify the flow for locking prepare header sectors, by getting rid of [write_prepare_on_lock_header_sector](https://github.com/tigerbeetle/tigerbeetle/blob/bc4a71bc5fa93de1f33403b0ea1d7a1f176e565e/src/vsr/journal.zig#L1862) & [write_prepare_unlock_header_sector](https://github.com/tigerbeetle/tigerbeetle/blob/bc4a71bc5fa93de1f33403b0ea1d7a1f176e565e/src/vsr/journal.zig#L1931). In doing so, we are forgoing the optimization that explicit header sector locking implemented for simplicity. In the future, if we want to implement that optimization again, it can be implemented in [lock_sectors](https://github.com/tigerbeetle/tigerbeetle/blob/bc4a71bc5fa93de1f33403b0ea1d7a1f176e565e/src/vsr/journal.zig#L2014), by implementing a variant that can be called when the write already has a queue behind it, and is guaranteed not to collide with an in-flight write.

### DIFF
--- a/src/vsr/journal.zig
+++ b/src/vsr/journal.zig
@@ -1954,6 +1954,9 @@ pub fn JournalType(comptime Replica: type, comptime Storage: type) type {
                 if (!other.range.locked) continue;
 
                 if (other.range.overlaps(&write.range)) {
+                    assert(other.range.offset == write.range.offset);
+                    assert(other.range.buffer.len == write.range.buffer.len);
+
                     var tail = &other.range;
                     while (tail.next) |next| tail = next;
                     tail.next = &write.range;
--- a/src/vsr/journal.zig
+++ b/src/vsr/journal.zig
@@ -2016,13 +2016,6 @@ pub fn JournalType(comptime Replica: type, comptime Storage: type) type {
                 current = waiting.next;
                 waiting.next = null;
                 journal.lock_sectors(@as(*Journal.Write, @fieldParentPtr("range", waiting)));
-                // Header sector locks are always an exact match, so there's no need to re-check the
-                // the waiting writes against all other writes.
-                if (range.ring == .headers) {
-                    assert(waiting.locked);
-                    waiting.next = current;
-                    break;
-                }
             }
 
             range.callback(write);
--- a/src/vsr/journal.zig
+++ b/src/vsr/journal.zig
@@ -175,13 +175,6 @@ pub fn JournalType(comptime Replica: type, comptime Storage: type) type {
             message: *Message.Prepare,
             trigger: Trigger,
 
-            /// True if this Write has acquired a lock on a sector of headers.
-            /// This also means that the Write is currently writing sectors or queuing to do so.
-            header_sector_locked: bool = false,
-
-            /// Linked list of Writes waiting to acquire the same header sector as this Write.
-            header_sector_next: ?*Write = null,
-
             /// This is reset to undefined and reused for each Storage.write_sectors() call.
             range: Range,
         };
@@ -1834,42 +1827,12 @@ pub fn JournalType(comptime Replica: type, comptime Storage: type) type {
                 journal.write_prepare_release(write, null);
                 return;
             }
-
-            assert(!write.header_sector_locked);
-            assert(write.header_sector_next == null);
-
-            const write_offset = journal.offset_logical_in_headers_for_message(message);
-
-            var it = journal.writes.iterate();
-            while (it.next()) |other| {
-                if (other == write) continue;
-                if (!other.header_sector_locked) continue;
-
-                const other_offset = journal.offset_logical_in_headers_for_message(other.message);
-                if (other_offset == write_offset) {
-                    // The `other` and `write` target the same sector; append to the list.
-                    var tail = other;
-                    while (tail.header_sector_next) |next| tail = next;
-                    tail.header_sector_next = write;
-                    return;
-                }
-            }
-
-            write.header_sector_locked = true;
-            journal.write_prepare_on_lock_header_sector(write);
-        }
-
-        fn write_prepare_on_lock_header_sector(journal: *Journal, write: *Write) void {
-            assert(journal.status == .recovered);
-            assert(write.header_sector_locked);
-
             // TODO It's possible within this section that the header has since been replaced but we
             // continue writing, even when the dirty bit is no longer set. This is not a problem
             // but it would be good to stop writing as soon as we see we no longer need to.
             // For this, we'll need to have a way to tweak write_prepare_release() to release locks.
             // At present, we don't return early here simply because it doesn't yet do that.
 
-            const message = write.message;
             const slot_of_message = journal.slot_for_header(message.header);
             const offset = Ring.headers.offset(slot_of_message);
             assert(offset % constants.sector_size == 0);
@@ -1885,7 +1848,6 @@ pub fn JournalType(comptime Replica: type, comptime Storage: type) type {
                 offset,
                 offset + constants.sector_size,
             });
-
             // Memory must not be owned by journal.headers as these may be modified concurrently:
             assert(@intFromPtr(buffer.ptr) < @intFromPtr(journal.headers.ptr) or
                 @intFromPtr(buffer.ptr) > @intFromPtr(journal.headers.ptr) + headers_size);
@@ -1896,9 +1858,6 @@ pub fn JournalType(comptime Replica: type, comptime Storage: type) type {
         fn write_prepare_on_write_header(write: *Journal.Write) void {
             const journal = write.journal;
             const message = write.message;
-
-            assert(write.header_sector_locked);
-            journal.write_prepare_unlock_header_sector(write);
 
             if (!journal.has(message.header)) {
                 journal.write_prepare_debug(message.header, "entry changed while writing headers");
@@ -1924,25 +1883,6 @@ pub fn JournalType(comptime Replica: type, comptime Storage: type) type {
             journal.faulty.clear(slot);
 
             journal.write_prepare_release(write, message);
-        }
-
-        /// Release the lock held by a write on an in-memory header sector and pass
-        /// it to a waiting Write, if any.
-        fn write_prepare_unlock_header_sector(journal: *Journal, write: *Journal.Write) void {
-            assert(write.header_sector_locked);
-            write.header_sector_locked = false;
-
-            // Unlike the ranges of physical memory we lock when writing to disk,
-            // these header sector locks are always an exact match, so there's no
-            // need to re-check the waiting writes against all other writes.
-            if (write.header_sector_next) |waiting| {
-                write.header_sector_next = null;
-
-                assert(waiting.header_sector_locked == false);
-                waiting.header_sector_locked = true;
-                journal.write_prepare_on_lock_header_sector(waiting);
-            }
-            assert(write.header_sector_next == null);
         }
 
         fn write_prepare_release(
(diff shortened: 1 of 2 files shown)


chaitanyabhandari:
The following failing VOPR seeds found by the CFO are unrelated to this PR as they fail on `main` as well:
1. ./zig/zig build -Drelease vopr -- 12757902879161248377
2. ./zig/zig build -Drelease vopr -- --lite 11833167314180507701

As per the tail of the VOPR log, this is *likely* related to https://github.com/tigerbeetle/tigerbeetle/pull/2272/commits/71ad5d6f12959439c899b0d0415fc217d28a044c, which changed an assert from `assert(fault or replica.op < replica.op_checkpoint() or replica.log_view < replica.superblock.working.vsr_state.sync_view or header_prepare_view_mismatch);` → `assert(fault or header_prepare_view_mismatch)` since it removes the _sync_view_ field. 
```C
[warn] (replica): 2: transition_to_recovering_head: op_checkpoint=19 commit_min=19 op_head=16 log_view=7 view=14
[info] (cluster):  2 ^ /   h    14V  19/_19/_23C   0:_16Jo  0/_0J!   0:_31Wo <__0:__0> v2:3  nullGa  0G!   0G?
thread 4252714 panic: reached unreachable code
/Users/chaitanyabhandari/Desktop/tigerbeetle/tigerbeetle/zig/lib/std/debug.zig:412:14: 0x100e5970f in assert (vopr)
             ^
/Users/chaitanyabhandari/Desktop/tigerbeetle/tigerbeetle/src/vopr.zig:1102:19: 0x100edadeb in restart_replica (vopr)
            assert(fault or header_prepare_view_mismatch);
                  ^
/Users/chaitanyabhandari/Desktop/tigerbeetle/tigerbeetle/src/vopr.zig:1030:34: 0x100f319a3 in tick_crash_down (vopr)
        simulator.restart_replica(replica.replica, fault);
                                 ^
```

As per the failed seed `11833167314180507701` it is *indeed* possible for a `recovering_head` replica to satisfy `replica.op < replica.op_checkpoint()` condition. In this case, R2 appends ops 17 → 23 to its WAL in log_view=13 (in which it was a backup in .normal status). However, these ops are appended to its WAL while log_view_durable=7. When R2 restarts, its log_view is set 7 and ops 17 → 23 are truncated because they were appended in log_view=13. 

```
[warn] (journal): 2: recover_slot: recovered slot=0017 label=@TruncateViewRange decision=cut_view_range operation=vsr.Operation.reserved op=17
[warn] (journal): 2: recover_slot: recovered slot=0018 label=@TruncateViewRange decision=cut_view_range operation=vsr.Operation.reserved op=18
[warn] (journal): 2: recover_slot: recovered slot=0019 label=@TruncateViewRange decision=cut_view_range operation=vsr.Operation.reserved op=19
[warn] (journal): 2: recover_slot: recovered slot=0020 label=@TruncateViewRange decision=cut_view_range operation=vsr.Operation.reserved op=20
[warn] (journal): 2: recover_slot: recovered slot=0021 label=@TruncateViewRange decision=cut_view_range operation=vsr.Operation.reserved op=21
[warn] (journal): 2: recover_slot: recovered slot=0022 label=@TruncateViewRange decision=cut_view_range operation=vsr.Operation.reserved op=22
[warn] (journal): 2: recover_slot: recovered slot=0023 label=@TruncateViewRange decision=cut_view_range operation=vsr.Operation.reserved op=23
```


Since we have a conditon in `open` wherein we transition to `recovering_head` if `self.op_checkpoint() > self.op`,  I am adding back `replica.op < replica.op_checkpoint()` to the VOPR. However, I wonder if it would be beneficial if backups could *also* reject prepares if they haven't yet made their log_view durable (we implemented this for primaries in https://github.com/tigerbeetle/tigerbeetle/pull/2253). 🤔 
https://github.com/tigerbeetle/tigerbeetle/blob/bc4a71bc5fa93de1f33403b0ea1d7a1f176e565e/src/vsr/replica.zig#L771-L772

matklad:
>However, these ops are appended to its WAL while log_view_durable=7. When R2 restarts, its log_view is set 7 and ops 17 → 23 are truncated because they were appended in log_view=13.

Ah, so it looks like that assert actually did uncover a problem! If a replcia volonturely cuts its prepares due to `cut_view_range`, it shouldn't get into recovering head, as there isn't corruption (well, at least it _seems_ like just cut_view_range alone shouldn't be enough to trigger recovering head).

chaitanyabhandari:
> If a replcia volonturely cuts its prepares due to `cut_view_range`, it shouldn't get into recovering head, as there isn't corruption (well, at least it seems like just cut_view_range alone shouldn't be enough to trigger recovering head)

Mhmmmm this does make sense... however, in this case, `cut_view_range` caused self.op to regress below self.op_checkpoint(), and that does seem like a case where a replica _currently_ can't safely participate in consensus, right? So I _think_ .recovering_head is the only option here? 🤔

chaitanyabhandari:
On a different (but related) note, the logical change this PR makes seems to be doing well under VOPR scrutiny, so we _could_ get this in (without https://github.com/tigerbeetle/tigerbeetle/pull/2273/commits/a4cb4e87dc889a84eb7e7726a7738ca75e87f808 and post-review, of course 😄) and think about the correct approach to tackle https://github.com/tigerbeetle/tigerbeetle/pull/2273#issuecomment-2327821611 separately as well...
Screenshot 2024-09-04 at 10 32 30 AM

sentientwaffle:
> in this case, cut_view_range caused self.op to regress below self.op_checkpoint(), and that does seem like a case where a replica currently can't safely participate in consensus, right?

Could we repair our journal with `CheckpointState.header`? Iirc that `header.op` should equal the old checkpoint trigger, with is greater than `op_checkpoint()`.

> On a different (but related) note, the logical change this PR makes seems to be doing well under VOPR scrutiny, so we could get this in (without https://github.com/tigerbeetle/tigerbeetle/commit/a4cb4e87dc889a84eb7e7726a7738ca75e87f808 and post-review, of course 😄) and think about the correct approach to tackle https://github.com/tigerbeetle/tigerbeetle/pull/2273#issuecomment-2327821611 separately as well...

:+1:

chaitanyabhandari:
> > in this case, cut_view_range caused self.op to regress below self.op_checkpoint(), and that does seem like a case where a replica currently can't safely participate in consensus, right?
> 
> Could we repair our journal with `CheckpointState.header`? Iirc that `header.op` should equal the old checkpoint trigger, with is greater than `op_checkpoint()`.
> 

Ah great point, thanks dj!

I just checked, it's `op_checkpoint()`, not the trigger. But I believe this should _still_ suffice because with the help of `CheckpointState.header`, we would be able to make sure a replica *always* starts with a head op >= self.op_checkpoint. Let me try this on a separate branch right now.
https://github.com/tigerbeetle/tigerbeetle/blob/6092320f25b4e66b8f78f76cf2c20c622ca0041f/src/vsr/replica.zig#L4075-L4085

src/vsr/journal.zig
```diff
@@ -2083,6 +2016,13 @@ pub fn JournalType(comptime Replica: type, comptime Storage: type) type {
                 current = waiting.next;
                 waiting.next = null;
                 journal.lock_sectors(@as(*Journal.Write, @fieldParentPtr("range", waiting)));
+                // Header sector locks are always an exact match, so there's no need to re-check the
+                // the waiting writes against all other writes.
+                if (range.ring == .headers) {
+                    assert(waiting.locked);
+                    waiting.next = current;
+                    break;
+                }
```
sentientwaffle: > Header sector locks are always an exact match

This is a good point!

Instead of `Range.offset: u64` and `Range.overlaps()`, we could store `Range.slot: Slot` -- then prepares are an exact match too -- `.overlaps` would be replaced by `.equals() { compare ring and slot }`.

sentientwaffle: This approach of overwriting `waiting.next` (which I assume is guaranteed to be `null` still?) and then early-exiting from the loop makes me a little nervous. I think I'd rather just reuse the same loop code, even at the cost of the early-exist optimization.

Alternatively, you could refactor `lock_sectors` such that there is a variant that can be called when the write already has a queue behind it, and is guaranteed not to collide with an in-flight write.

chaitanyabhandari: Ah, yes! This is a great point, let me try this out.

chaitanyabhandari: Mhmmmmm, I *did* feel a bit awkward overwriting waiting.next as well. I had briefly tried branching out `lock_sectors` based on `ring == .headers` but it felt a bit unergonomic as well... Let me take another stab at this!

However, I don't think `waiting.next` would be guaranteed to be null, here? IIUC, if we were to remove this branch and just do the below (the older logic), the first header that is queued would be able to take the lock, and then the rest of the ones waiting would be queued on it again after this loop (which is the state that we arrive at by breaking out of the loop early as well).
```C
            var current = range.next;
            range.next = null;
            while (current) |waiting| {
                assert(waiting.locked == false);
                current = waiting.next;
                waiting.next = null;
                journal.lock_sectors(@as(*Journal.Write, @fieldParentPtr("range", waiting)));
            }

```

The only difference is that each queued header write would have to iterate through all of the write IOPs: https://github.com/tigerbeetle/tigerbeetle/blob/a4cb4e87dc889a84eb7e7726a7738ca75e87f808/src/vsr/journal.zig#L1951-L1962

sentientwaffle: `lock_sectors` asserts that `write.range.next == null`, and we set it to `null` immediately before calling `lock_sectors`:

```
                waiting.next = null;
                journal.lock_sectors(@as(*Journal.Write, @fieldParentPtr("range", waiting)));
```

And `lock_sectors` itself doesn't modify its parameter's `next` (though it may append it to a different write).

chaitanyabhandari: Took a stab at it in https://github.com/tigerbeetle/tigerbeetle/pull/2273/commits/97c50edf318c863c612641d2b206c54a2c082503.

chaitanyabhandari: Ah yes, makes sense! Decided to go for the simpler implementation here instead of implementing a variant of `lock_sectors`: https://github.com/tigerbeetle/tigerbeetle/pull/2273/commits/569a1e3f342c7c6ff8b243d5953df6b5641cd508.

chaitanyabhandari: We decided to revert this and instead just add asserts in the older version: https://github.com/tigerbeetle/tigerbeetle/pull/2273/commits/94de0abe43d6bedc4ae26e46973fb480ecdd36ec

chaitanyabhandari:
Also, given replicas can currently go to .recovering_head if `self.op < self.op_checkpoint()`, are we okay to retain & merge  https://github.com/tigerbeetle/tigerbeetle/pull/2273/commits/a4cb4e87dc889a84eb7e7726a7738ca75e87f808 as part of this PR as well? I could then change the condition from `assert(fault or replica.op < replica.op_checkpoint() or header_prepare_view_mismatch);` → `assert(fault or header_prepare_view_mismatch)` while I push a fix for this (as per https://github.com/tigerbeetle/tigerbeetle/pull/2273#issuecomment-2329706765).

sentientwaffle (approved):

PR #2274, opened 2024-09-03
Add zipfian-distributed random numbers to benchmark

brson:
This adds two generators of Zipfian random numbers, and modifies the benchmark to use them to select accounts for transfers and querying.

This immediately allows the benchmark to produce more realistic workloads than at present. The ultimate goal though is to approximate the five YCSB core workloads within our benchmark, which this PR does not do.

The comments in zipfian.zig are thorough and mostly not repeated in this PR description.

Some sources that may help understand this PR:

- https://en.wikipedia.org/wiki/Zipf's_law
- ["Quickly Generating Billion Record Synthetic Databases"](https://www.microsoft.com/en-us/research/publication/quickly-generating-billion-record-synthetic-databases/) - the basis for this technique, page 30. Note the code listing contains errors.
- [YCSB's ZipfianGenerator](https://github.com/brianfrankcooper/YCSB/blob/master/core/src/main/java/site/ycsb/generator/ZipfianGenerator.java) - for comparison, also derived from the paper, but I did not use it directly
- [YCSB paper](https://courses.cs.duke.edu/fall13/cps296.4/838-CloudPapers/ycsb.pdf) - core workloads described on page 6

There is an obvious unit test missing here: a 2-sample kolmogorov-smirnov test that the generated numbers are statistically likely to follow the zipfian distribution. I have done this test in python, which has the required statistical methods built-in, but not in zig. I am happy to write this test in zig if asked.

I have not paid close attention to overflow, nan, or type narrowing issues as this is test code and we always compile in ReleaseSafe mode, though I'm happy to address such issues if requested.

---

This modifies the benchmark in incompatible ways. It removes the `--account-count-hot` and `--transfer-hot-percent` flags, and adds the `--account-distribution` flag, with the possible values "zipfian", "latest", and "uniform".

"zipfian" is the default as it is probably most representative of real workloads, with hot accounts being shuffled randomly across the keyspace. Passing "--account-distribution=uniform" results in the same workloads as today, but without the concept of "hot accounts". "latest" is a zipfian distribution where the most recent accounts are the most accessed; used in the YCSB core workloads.

It is no longer possible to represent the previous `--account-count-hot` and `--transfer-hot-percent` behavior. These previously allowed for a rough approximation of the "80-20 rule" and similar distributions of work, where e.g. 80 percent of transfers would use 20 percent of accounts.

That type of behavior is the purpose of the Zipfian distribution, though at least in this PR it is not tunable the way the previous flags are. It would be desirable and feasible to have similar flags like `--account-hot-percent=20` and `--account-hot-probability=80`. As mentioned in the comments of zipfian.zig, I don't presently know how to tune the distribution to say "20 percent of items have 80 percent probability" when also the number of items selected from is growable. Tuning the distribution like this involves selecting the correct `theta` for the set size, and maintaining the same 80-20 ratio seemingly requires changing theta with the set size; and there is an expensive calculation involved in changing theta. I have not yet seen any literature about this topic and have not attempted to implement it. An important caveat here though is that __at present__ the benchmark _does not_ grow the set of accounts during the benchmark, which moots the issue. I do expect to modify the benchmark in the future though to insert new accounts over time, as needed by the YCSB core workloads.

Another change: where previously the "hot" accounts would be selected for only the debit side of a transfer, and credit accounts would be selected uniformly, now both sides of the transfer follow the selected distribution.

---

Perhaps surprisingly, tigerbeetle appears to perform worse with zipfian access patterns than uniform, and so after this PR that switches to zipfian by default benchmark results in the default configuration will appear to regress.

Here are three representative runs, using uniform, zipfian, and latest distributions on my laptop:

```
$ zig/zig build run -- benchmark --transfer-count=1000000 --account-distribution=uniform
info(io): opening "0_0-3e8c5e56.tigerbeetle.benchmark"...
info(main): multiversioning: disabled for development (0.0.1) release.
info(main): release=0.0.1
info(main): git_commit=df661c9c8e46a6b1b7dccdd2affa31628f9fd529
info(replica): superblock release=0.0.1
info(main): 0: Allocated 3070MiB during replica init
info(main): 0: Grid cache: 1024MiB, LSM-tree manifests: 128MiB
info(main): 0: cluster=0: listening on 127.0.0.1:45325
warning(main): 0: started with constants.verify - expect reduced performance. Recompile with -Dconfig=production if unexpected.
Benchmark must be built with '-Drelease' for reasonable results.
Benchmark must be built with '-Dconfig=production' for reasonable results.
info: Benchmark running against { 127.0.0.1:45325 }
info: Benchmark seed = 42
info: Account distribution = uniform
info(message_bus): connection from client 126812723862003569197151784371068247806
123 batches in 27.36 s
transfer batch delay = 0 us
load accepted = 36548 tx/s
batch latency p1 = 79 ms
batch latency p10 = 83 ms
batch latency p20 = 84 ms
batch latency p30 = 85 ms
batch latency p40 = 87 ms
batch latency p50 = 89 ms
batch latency p60 = 94 ms
batch latency p70 = 189 ms
batch latency p80 = 231 ms
batch latency p90 = 293 ms
batch latency p95 = 326 ms
batch latency p99 = 2876 ms
batch latency p100 = 3616 ms

100 queries in 8.80 s
query latency p1 = 6 ms
query latency p10 = 6 ms
query latency p20 = 7 ms
query latency p30 = 7 ms
query latency p40 = 7 ms
query latency p50 = 8 ms
query latency p60 = 8 ms
query latency p70 = 9 ms
query latency p80 = 10 ms
query latency p90 = 285 ms
query latency p95 = 316 ms
query latency p99 = 394 ms
query latency p100 = 3465 ms

rss = 3234099200 bytes

datafile empty = 1141374976 bytes
datafile = 1823477760 bytes
```

```
$ zig/zig build run -- benchmark --transfer-count=1000000 --account-distribution=zipfian
info(io): opening "0_0-f156d697.tigerbeetle.benchmark"...
info(main): multiversioning: disabled for development (0.0.1) release.
info(main): release=0.0.1
info(main): git_commit=df661c9c8e46a6b1b7dccdd2affa31628f9fd529
info(replica): superblock release=0.0.1
info(main): 0: Allocated 3070MiB during replica init
info(main): 0: Grid cache: 1024MiB, LSM-tree manifests: 128MiB
info(main): 0: cluster=0: listening on 127.0.0.1:35147
warning(main): 0: started with constants.verify - expect reduced performance. Recompile with -Dconfig=production if unexpected.
Benchmark must be built with '-Drelease' for reasonable results.
Benchmark must be built with '-Dconfig=production' for reasonable results.
info: Benchmark running against { 127.0.0.1:35147 }
info: Benchmark seed = 42
info: Account distribution = zipfian
info(message_bus): connection from client 180844811304677403733424898741826464730
info(message_bus): message queue for peer message_bus.MessageBusType(.replica).Connection.Peer{ .replica = 0 } full, dropping request message
info(message_bus): message queue for peer message_bus.MessageBusType(.replica).Connection.Peer{ .replica = 0 } full, dropping request message
123 batches in 43.87 s
transfer batch delay = 0 us
load accepted = 22794 tx/s
batch latency p1 = 125 ms
batch latency p10 = 141 ms
batch latency p20 = 146 ms
batch latency p30 = 175 ms
batch latency p40 = 182 ms
batch latency p50 = 190 ms
batch latency p60 = 207 ms
batch latency p70 = 242 ms
batch latency p80 = 276 ms
batch latency p90 = 360 ms
batch latency p95 = 397 ms
batch latency p99 = 4553 ms
batch latency p100 = 6304 ms

100 queries in 11.50 s
query latency p1 = 4 ms
query latency p10 = 12 ms
query latency p20 = 16 ms
query latency p30 = 21 ms
query latency p40 = 33 ms
query latency p50 = 61 ms
query latency p60 = 67 ms
query latency p70 = 73 ms
query latency p80 = 81 ms
query latency p90 = 294 ms
query latency p95 = 332 ms
query latency p99 = 494 ms
query latency p100 = 3144 ms

rss = 3234918400 bytes

datafile empty = 1141374976 bytes
datafile = 1823477760 bytes
```

```
$ zig/zig build run -- benchmark --transfer-count=1000000 --account-distribution=latest
info(io): opening "0_0-b7af7882.tigerbeetle.benchmark"...
info(main): multiversioning: disabled for development (0.0.1) release.
info(main): release=0.0.1
info(main): git_commit=df661c9c8e46a6b1b7dccdd2affa31628f9fd529
info(replica): superblock release=0.0.1
info(main): 0: Allocated 3070MiB during replica init
info(main): 0: Grid cache: 1024MiB, LSM-tree manifests: 128MiB
info(main): 0: cluster=0: listening on 127.0.0.1:43467
warning(main): 0: started with constants.verify - expect reduced performance. Recompile with -Dconfig=production if unexpected.
Benchmark must be built with '-Drelease' for reasonable results.
Benchmark must be built with '-Dconfig=production' for reasonable results.
info: Benchmark running against { 127.0.0.1:43467 }
info: Benchmark seed = 42
info: Account distribution = latest
info(message_bus): connection from client 193447347888948890384498736063778658676
123 batches in 31.64 s
transfer batch delay = 0 us
load accepted = 31602 tx/s
batch latency p1 = 92 ms
batch latency p10 = 98 ms
batch latency p20 = 103 ms
batch latency p30 = 112 ms
batch latency p40 = 119 ms
batch latency p50 = 128 ms
batch latency p60 = 142 ms
batch latency p70 = 203 ms
batch latency p80 = 284 ms
batch latency p90 = 308 ms
batch latency p95 = 352 ms
batch latency p99 = 3423 ms
batch latency p100 = 3738 ms

100 queries in 11.78 s
query latency p1 = 5 ms
query latency p10 = 6 ms
query latency p20 = 7 ms
query latency p30 = 10 ms
query latency p40 = 25 ms
query latency p50 = 50 ms
query latency p60 = 68 ms
query latency p70 = 72 ms
query latency p80 = 88 ms
query latency p90 = 282 ms
query latency p95 = 365 ms
query latency p99 = 457 ms
query latency p100 = 3541 ms

rss = 3232616448 bytes

datafile empty = 1141374976 bytes
datafile = 1823477760 bytes
```

### DIFF
--- a/src/stdx/zipfian.zig
+++ b/src/stdx/zipfian.zig
@@ -105,11 +105,13 @@ pub const ZipfianGenerator = struct {
         // Math voodoo, copied from the paper,
         // which doesn't explain it, but claims it is from Knuth volume 3.
 
-        const nf: f64 = @floatFromInt(self.n);
-
         // NB: These depend only on zetan and could be cached for a minor speedup.
         const alpha = 1.0 / (1.0 - self.theta);
-        const eta = (1.0 - math.pow(f64, 2.0 / nf, 1.0 - self.theta)) /
+        const eta = (1.0 - math.pow(
+            f64,
+            2.0 / @as(f64, @floatFromInt(self.n)),
+            1.0 - self.theta,
+        )) /
             (1.0 - zeta(2.0, self.theta) / self.zetan);
 
         const u = rng.float(f64);
@@ -123,7 +125,10 @@ pub const ZipfianGenerator = struct {
             return 1;
         }
 
-        return @as(u64, @intFromFloat(nf * math.pow(f64, (eta * u) - eta + 1.0, alpha)));
+        return @as(u64, @intFromFloat(
+            @as(f64, @floatFromInt(self.n)) *
+                math.pow(f64, (eta * u) - eta + 1.0, alpha),
+        ));
     }
 
     /// Grow the size of the random set.
@@ -140,7 +145,6 @@ pub const ZipfianGenerator = struct {
     /// The probability that an index will be chosen.
     fn probability(self: *const ZipfianGenerator, item: u64) f64 {
         assert(item < self.n);
-        const itemf: f64 = @floatFromInt(item);
 
         // Reference: https://en.wikipedia.org/wiki/Zipf's_law#Formal_definition
         //
@@ -150,7 +154,11 @@ pub const ZipfianGenerator = struct {
         //
         // zetan is the generalized harmonic number of order "s" (theta) for `n`.
         // We add 1 to `k` because our items are 0-based but the math is 1-based.
-        return (1.0 / self.zetan) * (1.0 / math.pow(f64, itemf + 1, self.theta));
+        return (1.0 / self.zetan) * (1.0 / math.pow(
+            f64,
+            @as(f64, @floatFromInt(item)) + 1,
+            self.theta,
+        ));
     }
 
     /// Returns the numbef of items at which the cumulative distribution function (CDF - the
@@ -180,8 +188,7 @@ fn zeta(n: u64, theta: f64) f64 {
     var i: u64 = 1;
     var zeta_sum: f64 = 0.0;
     while (i <= n) : (i += 1) {
-        const ifl: f64 = @floatFromInt(i);
-        zeta_sum += math.pow(f64, 1.0 / ifl, theta);
+        zeta_sum += math.pow(f64, 1.0 / @as(f64, @floatFromInt(i)), theta);
     }
     return zeta_sum;
 }
@@ -197,8 +204,7 @@ fn zeta_incremental(
     var i = n_previous + 1;
     var zeta_sum = zetan_previous;
     while (i <= n_new) : (i += 1) {
-        const ifl: f64 = @floatFromInt(i);
-        zeta_sum += math.pow(f64, 1.0 / ifl, theta);
+        zeta_sum += math.pow(f64, 1.0 / @as(f64, @floatFromInt(i)), theta);
     }
     return zeta_sum;
 }
--- a/src/stdx/zipfian.zig
+++ b/src/stdx/zipfian.zig
@@ -86,7 +86,7 @@ pub const ZipfianGenerator = struct {
     ///
     /// `theta` is the "skew" and is usually specified to be greater than 0 and less than 1,
     /// with YCSB using 0.99, though values greater than 1 also seem to generate reasonable
-    /// distributions.
+    /// distributions. `theta = 1` isn't allowed since it does not behave reasonably. 
     pub fn init_theta(items: u64, theta: f64) ZipfianGenerator {
         assert(theta > 0.0);
         assert(theta != 1.0); // 1.0 does not behave reasonably.
--- a/src/stdx/zipfian.zig
+++ b/src/stdx/zipfian.zig
@@ -38,11 +38,11 @@
 //! both of which generate random keys from 0 to a specified maximum.
 //! In the basic `ZipfianGenerator`, key 0 has the highest probability,
 //! 1 the next highest, etc.
-//! The `ShuffledZipfian` generator instead spreads the distribution out
+//! The `ZipfianShuffled` generator instead spreads the distribution out
 //! across the key space as if it were a shuffled deck.
 //!
 //! Both generators allow for the key space to grow (but not shrink),
-//! dynamically recomputing the distribution. When the `ShuffledZipfian`
+//! dynamically recomputing the distribution. When the `ZipfianShuffled`
 //! generator grows it acts as if each new key was inserted into the shuffled
 //! deck randomly, preserving the relative probability of existing keys.
 //!
@@ -50,21 +50,21 @@
 //! the shuffled generator does not because the tail distribution is fudged.
 
 const std = @import("std");
-const assert = @import("std").debug.assert;
-const Random = @import("std").Random;
-const math = @import("std").math;
-const BoundedArray = @import("./bounded_array.zig").BoundedArray;
+const stdx = @import("../stdx.zig");
+const assert = std.debug.assert;
+const Random = std.Random;
+const math = std.math;
+const BoundedArray = stdx.BoundedArray;
 
 /// The default "skew" of the distribution.
 const theta_default = 0.99; // per YCSB
 
 /// Generates Zipfian-distributed numbers from 0 to a specified maximum.
 ///
-/// The internal variables here are the same is in the paper;
-/// the external intended to be more understandable to the user.
+/// Many internal variables here are the same is in the paper, which I think
+/// should reduce confusion if this subject needs to be revisited; the external
+/// intended to be more understandable to the user and follow TigerStyle.
 pub const ZipfianGenerator = struct {
-    const Self = @This();
-
     theta: f64,
 
     /// The number of items in the set.
@@ -86,10 +86,10 @@ pub const ZipfianGenerator = struct {
     ///
     /// `theta` is the "skew" and is usually specified to be greater than 0 and less than 1,
     /// with YCSB using 0.99, though values greater than 1 also seem to generate reasonable
-    /// distributions. `theta = 1` isn't allowed since it does not behave reasonably. 
+    /// distributions. `theta = 1` isn't allowed since it does not behave reasonably.
     pub fn init_theta(items: u64, theta: f64) ZipfianGenerator {
         assert(theta > 0.0);
-        assert(theta != 1.0); // 1.0 does not behave reasonably.
+        assert(theta != 1.0);
         return ZipfianGenerator{
             .theta = theta,
             .n = items,
@@ -97,7 +97,9 @@ pub const ZipfianGenerator = struct {
         };
     }
 
-    pub fn next(self: *const Self, rng: Random) u64 {
+    /// Note that the variables in this function are mostly named
+    /// as in the reference paper and do not follow TigerStyle.
+    pub fn next(self: *const ZipfianGenerator, rng: Random) u64 {
         assert(self.n > 0);
 
         // Math voodoo, copied from the paper,
@@ -125,9 +127,9 @@ pub const ZipfianGenerator = struct {
     }
 
     /// Grow the size of the random set.
-    pub fn grow(self: *Self, new_items: u64) void {
+    pub fn grow(self: *ZipfianGenerator, new_items: u64) void {
         const items = self.n + new_items;
-        const zetan_new = zeta_incr(self.n, new_items, self.zetan, self.theta);
+        const zetan_new = zeta_incremental(self.n, new_items, self.zetan, self.theta);
         self.* = .{
             .theta = self.theta,
             .n = items,
@@ -136,7 +138,7 @@ pub const ZipfianGenerator = struct {
     }
 
     /// The probability that an index will be chosen.
-    fn probability(self: *const Self, item: u64) f64 {
+    fn probability(self: *const ZipfianGenerator, item: u64) f64 {
         assert(item < self.n);
         const itemf: f64 = @floatFromInt(item);
 
@@ -152,18 +154,19 @@ pub const ZipfianGenerator = struct {
     }
 
     /// Returns the numbef of items at which the cumulative distribution function (CDF - the
-    /// probability of some value less than x being generated) is greater or equal to `prob`.
+    /// probability of some value less than x being generated) is greater or equal to
+    /// `cdf_probability`.
     ///
     /// If there is no such value, returns the total number of items.
-    fn cumulative_distribution_items(self: *const Self, prob: f64) u64 {
-        assert(prob >= 0.0 and prob <= 1.0);
+    fn cumulative_distribution_items(self: *const ZipfianGenerator, cdf_probability: f64) u64 {
+        assert(cdf_probability >= 0.0 and cdf_probability <= 1.0);
 
-        var idx: u64 = 0;
-        var prob_sum: f64 = 0.0;
-        while (idx < self.n) : (idx += 1) {
-            prob_sum += self.probability(idx);
-            if (prob_sum >= prob) {
-                return idx + 1;
+        var index: u64 = 0;
+        var probability_sum: f64 = 0.0;
+        while (index < self.n) : (index += 1) {
+            probability_sum += self.probability(index);
+            if (probability_sum >= cdf_probability) {
+                return index + 1;
             }
         }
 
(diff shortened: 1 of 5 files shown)


sentientwaffle:
> Another change: where previously the "hot" accounts would be selected for only the debit side of a transfer, and credit accounts would be selected uniformly, now both sides of the transfer follow the selected distribution.

Do you mean that it is likely that both sides of the transfer are hot accounts? I think that is not as realistic for the use case that this was trying to test -- that of a hot, low-cardinality operator account transacting with a high-cardinality user account.

brson:
> > Another change: where previously the "hot" accounts would be selected for only the debit side of a transfer, and credit accounts would be selected uniformly, now both sides of the transfer follow the selected distribution.
> 
> Do you mean that it is likely that both sides of the transfer are hot accounts? I think that is not as realistic for the use case that this was trying to test -- that of a hot, low-cardinality operator account transacting with a high-cardinality user account.

Yeah, that's what I mean. I'll revisit and try to preserve both capabilities.

src/stdx/zipfian.zig
```diff
@@ -0,0 +1,490 @@
...
+
+/// The default "skew" of the distribution.
+const theta_default = 0.99; // per YCSB
+
+/// Generates Zipfian-distributed numbers from 0 to a specified maximum.
+///
+/// The internal variables here are the same is in the paper;
+/// the external intended to be more understandable to the user.
+pub const ZipfianGenerator = struct {
+    const Self = @This();
```
batiati:
```suggestion
```
There's no need to alias `const Self = @This()`, it's only required for generic types.

We also avoid `Self` in favor of a more descriptive type name (although `self` as the argument name can be used).
Example:
```zig
pub fn next(self: *const ZipfianGenerator, rng: Random) u64
```

src/stdx/zipfian.zig
batiati, line 1: General adjustments regarding the TigerStyle:

We should avoid using abbreviations such as `idx` for `index`, `prob` for `probability`, or `incr` for `increment`.

In some cases, it might be how the variable is known in a paper or reference implementation—which (I guess) might be the case of `u`, `uz`, `eta`, `ans`, etc. So it's fine to keep the reference name, but not without a comment explaining the meaning of each new term introduced in the scope.

Following this logic, should `zetan` be `zeta_n`? 

Naming variables such as `ifl`, `itemf` and `nf` meaning "_the integer value as a float_" is discouraged as well.

Also, observe the MSB order when naming things: i.e. `prev_zetan` -> `zetan_previous`, `ShuffledZipfian` -> `ZipfianShuffled`.

src/stdx/zipfian.zig
```diff
@@ -0,0 +1,490 @@
...
+
+        if (uz < 1.0) {
+            return 0;
+        }
+
+        if (uz < 1.0 + math.pow(f64, 0.5, self.theta)) {
+            return 1;
+        }
+
+        return @as(u64, @intFromFloat(nf * math.pow(f64, eta * u - eta + 1.0, alpha)));
```
batiati: Make the precedence explicit:

```suggestion
        return @as(u64, @intFromFloat(nf * math.pow(f64, (eta * u) - eta + 1.0, alpha)));
```

src/stdx/zipfian.zig
```diff
@@ -0,0 +1,490 @@
...
+//! deck randomly, preserving the relative probability of existing keys.
+//!
+//! The non-shuffled generator should pass a 2-sample Kolmogorov–Smirnov test;
+//! the shuffled generator does not because the tail distribution is fudged.
+
+const std = @import("std");
+const assert = @import("std").debug.assert;
+const Random = @import("std").Random;
+const math = @import("std").math;
+const BoundedArray = @import("./bounded_array.zig").BoundedArray;
```
batiati: NIT: It's a common idiom importing the topmost container and derive the other imports from there.

Examples:

Use `const Random = std.Random` instead of `@import("std").Random`.

 Import `const stdx = @import("stdx.zig")` and then use `stdx.BoundedArray` instead of `@import("./bounded_array.zig").BoundedArray`.

src/stdx/zipfian.zig
```diff
@@ -0,0 +1,490 @@
...
+    /// Create a generator from `[0, items)` with `theta` equal to 0.99.
+    pub fn init(items: u64) ZipfianGenerator {
+        return ZipfianGenerator.init_theta(items, theta_default);
+    }
+
+    /// Create a generator from `[0, items)` with given `theta`.
+    ///
+    /// `theta` is the "skew" and is usually specified to be greater than 0 and less than 1,
+    /// with YCSB using 0.99, though values greater than 1 also seem to generate reasonable
+    /// distributions.
```
batiati: Move the explanation of why `assert(theta != 1.0)` to the function's documentation.

```suggestion
    /// `theta` is the "skew" and is usually specified to be greater than 0 and less than 1,
    /// with YCSB using 0.99, though values greater than 1 also seem to generate reasonable
    /// distributions. `theta = 1` isn't allowed since it does not behave reasonably. 
```

src/stdx/zipfian.zig
```diff
@@ -0,0 +1,490 @@
...
+            .theta = theta,
+            .n = items,
+            .zetan = zeta(items, theta),
+        };
+    }
+
+    pub fn next(self: *const Self, rng: Random) u64 {
+        assert(self.n > 0);
+
+        // Math voodoo, copied from the paper,
```
batiati, line 105: :rofl:

src/stdx/zipfian.zig
brson: I have partially resolved this: spelled out abbreviations, switched word orders, but have some questions and reservations:



> In some cases, it might be how the variable is known in a paper or reference implementation—which (I guess) might be the case of `u`, `uz`, `eta`, `ans`, etc. So it's fine to keep the reference name, but not without a comment explaining the meaning of each new term introduced in the scope.

> Following this logic, should `zetan` be `zeta_n`?

Indeed these terse names, as well as the member variables `theta`, `n`, and `thetan`, and the names in the `zeta` function, are written as in the reference paper.

My inclination is to leave most/all of them as in the paper  - during research I found it confusing that different sources used different names for the same things and my _feeling_ is that the next person comes along and tries to understand this will have an easier time if the names are the same.

The `zeta` function is pretty straightforward and not specific to this algorithm, and the `zeta_incremental` function is my own, so I have tried to tigerstyle those functions more, including renaming `ans` to `zeta_sum`.

Regarding explaining `alpha`, `eta`, `u` and `uz`, I do not currently understand why they are named this way, but I have a copy of the knuth book on the way and will try to understand the algorithm more deeply. For now I have added comments on the `next` function explaining that the variable names are intentionally as in the paper and not tigerstyle.




> Naming variables such as `ifl`, `itemf` and `nf` meaning "_the integer value as a float_" is discouraged as well.

Can you suggest a replacement naming scheme for this pattern?

brson:
Thanks for the reviews @batiati.

@sentientwaffle I have not yet addressed your comment regarding being able to preserve different distributions for the credit/debit side of transfers, and specifically the use case of a single account sending transfers to every other account; but what I am thinking is:

There are three different uses of account distributions in the benchmark: for the credit and debit side of transfers, and for queries; so I add `--account-credit-distribution`, `--account-debit-distribution`, and `--account-query-distribution`. So the debitor can be zipfian and creditor uniform, etc.

Because setting all three is a pretty big hassle I would still keep the `--account-distribution` flag that sets them all at once.

This scheme doesn't strictly allow one to model the case where there is literally 1 central bank account sending transactions to every other account, but I would suggest the default zipfian skew is good enough for that scenerio for now.

I would like in the future to re-add the ability to more precisely control the proportion and probability of hot accounts, but in a way that works more generally with the concept of swappable probability distributions than the two flags I have removed (again I am thinking something like `--account-hot-percent=20` and `--account-hot-probability=80`). And if we needed to restore the ability to say "literally 1 account is the creditor on every transaction", we might create a special distribution for that called e.g. "single" (or we might support both `--account-hot-percent` and `--account-hot-count` and figure out how to translate between the two for every distribution).

src/stdx/zipfian.zig
batiati: > Can you suggest a replacement naming scheme for this pattern?

In most cases, just not taking an alias seems better than any name we can come up with, for example:
```zig
 zeta_sum += math.pow(f64, 1.0 / @as(f64, @floatFromInt(i)), theta);
```
Otherwise, we can follow the same non-abbreviation pattern: 
```zig
const whatever_float: f64 = @floatFromInt(whatever);
```

batiati:
> @sentientwaffle I have not yet addressed your comment regarding being able to preserve different distributions for the credit/debit side of transfers, and specifically the use case of a single account sending transfers to every other account; but what I am thinking is:

We already have two CLI args `account_count` and `account_count_hot`.
I don't think we need different distributions. If the user asks for `--account_count=1_000_000`  and `--account_count_hot=10` it would work pretty well if we could make the Zipfian distribution select the debit and credit accounts over the two distinct universes.

brson:
I have made a few changes:

- inlined all the float conversions so they don't need to be named
- switch the default distribution back to uniform for now - since i intend to make more significant changes to the benchmark in the future, i can rethink what the right defaults are later. so for now the default benchmark behavior is unchanged
- added separate distributions for debit/credit/query (prior to reading @batiati's latest comments)

> We already have two CLI args `account_count` and `account_count_hot`.
> I don't think we need different distributions. If the user asks for `--account_count=1_000_000` and `--account_count_hot=10` it would work pretty well if we could make the Zipfian distribution select the debit and credit accounts over the two distinct universes.

I need to think more about how to incorporate this feedback, and how it impacts my own intentions for future work. I don't personally need three distinct distributions, but the `--account-count-hot` flag is awkward for me for reasons I can articulate after more pondering. The idea of splitting the keyspace into two "universes" is intriguing though.

It also gives me an idea for how to naively/cleverly implement a zipfian distribution that maintains the 80/20 rule without excessive recomputation.

brson:
> > 
> 
> We already have two CLI args `account_count` and `account_count_hot`. I don't think we need different distributions. If the user asks for `--account_count=1_000_000` and `--account_count_hot=10` it would work pretty well if we could make the Zipfian distribution select the debit and credit accounts over the two distinct universes.

I'll prepare a patch the preserves basically the same previous logic for `--account-count-hot` and `--transfer-hot-percent`, while incorporating the zipfian distribution along the lines you describe, and removing the separate credit/debit/query distributions, and we can see what it looks like.

brson:
> I'll prepare a patch the preserves basically the same previous logic for `--account-count-hot` and `--transfer-hot-percent`, while incorporating the zipfian distribution along the lines you describe, and removing the separate credit/debit/query distributions, and we can see what it looks like.

@batiati I have pushed a commit that makes this change. I am ok with it if it is your preference.

batiati (commented):
Thanks Brian!
Could you please rebase this branch to fix the failing CI job?

When benchmarking queries, it's worth noting that the query's execution time is directly proportional to the number of matches. This explains the immense variation: querying hot accounts will likely find a full batch, while cold accounts may quickly return nothing or just a few transfers (to workaround this, we even changed DevHub to track P100 instead).

Maybe we could change the query benchmark to select only from hot accounts, if we want to benchmark small queries as well we could use `filter.limit = X`  for this purpose.
WDYT?

brson:
> Thanks Brian! Could you please rebase this branch to fix the failing CI job?

Done.

> 
> When benchmarking queries, it's worth noting that the query's execution time is directly proportional to the number of matches. This explains the immense variation: querying hot accounts will likely find a full batch, while cold accounts may quickly return nothing or just a few transfers (to workaround this, we even changed DevHub to track P100 instead).
> 
> Maybe we could change the query benchmark to select only from hot accounts, if we want to benchmark small queries as well we could use `filter.limit = X` for this purpose. WDYT?

Seems reasonable to me. Should I do that here or in a separate PR?

batiati (approved):

PR #2275, opened 2024-09-04
vsr: sync cleanups

matklad:
The goal of that PR was to unify `superblock.checkpoint` and `superblock.view_change` functions into a single

```
update(options: struct {
    view: ?ViewUpdate,
    checkpoint: ?Checkpoint,
})
```

That goal was not achieved! I did make a couple of refactors that make the code clearer along the way though.

The reason why I wasn't able unify the two functions is that they actually update checkpoint differently. `fn checkpoint` advances the checkpoint logically -- it materializes `CheckpointState` out of passed arguments. In contrast, `view_change` jumps the checkpoint physically --- it gets the _bytes_ of `CheckpointState` as an argument.

So instead:

* group checkpoint-related fields of `view_change` into an optional struct, to make it clear that the checkpoint is optionally updated
* add assertions that that checkpoint is connected to ours via parent_id/grandparent_id
* rename and peel a layer of `syncing.updating_superblock`.

I am still looking for a better name for `superblock.view_change` / `replica.view_durable_update`.

### DIFF
--- a/src/vsr/replica.zig
+++ b/src/vsr/replica.zig
@@ -284,9 +284,9 @@ pub fn ReplicaType(
         /// Durably store VSR state, the "root" of the LSM tree, and other replica metadata.
         superblock: SuperBlock,
 
-        /// Context for SuperBlock.open(), .checkpoint(), and .sync().
+        /// Context for SuperBlock.open() and .checkpoint().
         superblock_context: SuperBlock.Context = undefined,
-        /// Context for SuperBlock.view_change().
+        /// Context for SuperBlock.view_change(), which can happen concurrently to .checkpoint().
         superblock_context_view_change: SuperBlock.Context = undefined,
 
         grid: Grid,
@@ -1590,7 +1590,7 @@ pub fn ReplicaType(
             assert(message.header.replica < self.replica_count);
             assert(message.header.operation != .reserved);
 
-            if (self.syncing == .updating_superblock) {
+            if (self.syncing == .updating_checkpoint) {
                 log.debug("{}: on_prepare: ignoring (sync)", .{self.replica});
                 return;
             }
@@ -1903,7 +1903,7 @@ pub fn ReplicaType(
 
         fn on_repair(self: *Self, message: *Message.Prepare) void {
             assert(message.header.command == .prepare);
-            assert(self.syncing != .updating_superblock);
+            assert(self.syncing != .updating_checkpoint);
 
             if (self.status != .normal and self.status != .view_change) {
                 log.debug("{}: on_repair: ignoring ({})", .{ self.replica, self.status });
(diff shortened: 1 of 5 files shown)


src/vsr/replica.zig
```diff
@@ -7811,38 +7809,34 @@ pub fn ReplicaType(
...
+
+                                break :sync_op_min if (syncing_already)
+                                    @min(sync_min_old, sync_min_new)
+                                else
+                                    sync_min_new;
+                            },
+                            .sync_op_max = vsr.Checkpoint.trigger_for_checkpoint(
+                                checkpoint_state.header.op,
+                            ).?,
                         },
```
sentientwaffle, line 7838: This is an nice example of "push ifs up"!

src/vsr/superblock.zig
```diff
@@ -934,9 +934,26 @@ pub fn SuperBlockType(comptime Storage: type) type {
...
-            vsr_state.checkpoint = update.checkpoint.*;
-            vsr_state.sync_op_min = update.sync_op_min;
-            vsr_state.sync_op_max = update.sync_op_max;
+            if (update.sync_checkpoint) |*sync_checkpoint| {
+                assert(superblock.staging.vsr_state.checkpoint.header.op <
+                    sync_checkpoint.checkpoint.header.op);
+
+                const checkpoint_next = superblock.staging.vsr_state.checkpoint.header.op +
+                    constants.vsr_checkpoint_ops;
+                const checkpoint_next_next = checkpoint_next + constants.vsr_checkpoint_ops;
```
sentientwaffle: Maybe use `vsr.Checkpoint.checkpoint_after`?

sentientwaffle (commented):
LGTM, though a fuzzer needs updating before CI will pass.

src/vsr/superblock.zig
matklad: Ah good point!

I've been recently thinking that maybe we should actually remove next/after methods. I _think_ they predate the introduction of `vsr_checkpont_ops`, and, to me, something like

```
self.op_checkpoint() + constants.vsr_checkpoint_ops
```

actually reads better than

```
vsr.Checkpoint.checkpoint_after(self.op_checkpoint())
```

It is _the same_ name: `+` is after, ` constants.vsr_checkpoint_ops` is `vsr.Checkpoint.checkpoint`. And it is also polymorphic, in that you could do `+ 2 *` or `-`without inventing new names. 

So, we'd have just one _basis_: checkpoint, trigger, prepare_max, prepare_ok_max, and then open-code stepping by checkpoint_ops if needed. 

But we are not doing that right now, so better to stick to using helpers.

sentientwaffle: I think the only gotcha is the first checkpoint:

```
            if (checkpoint == 0) {
                // First wrap: op_checkpoint_next = 6-1 = 5
                // -1: vsr_checkpoint_ops is a count, result is an inclusive index.
                break :op constants.vsr_checkpoint_ops - 1;
            } else {
```

sentientwaffle (approved):

PR #2278, opened 2024-09-05
cli: tone down constants.verify warning

matklad:
We intentionally run our releases with verification on


matklad:
Added a second commit which just gets us rid of non-production build mode!

cb22 (approved):
> Added a second commit which just gets us rid of non-production build mode!

I think we should go one further, and remove `-Dconfig=...` :smile: but LGTM!

matklad:
>I think we should go one further, and remove -Dconfig=... 😄 but LGTM!

Huh,but I think we could, let me tackle this!

matklad:
https://github.com/tigerbeetle/tigerbeetle/pull/2292

PR #2279, opened 2024-09-05
build: remove tools/docker/Dockerfile

matklad:
That's not really a "TigerBeetle Dockerfile", just something we need to push out our container to the world. It is less confusing to just inline this at the point of use.


matklad:
Assigning to the next release manager!

chaitanyabhandari (approved):

PR #2280, opened 2024-09-05
ci: put coverage on devhub 

matklad:
We already record coverage, but, for some reason, we store it as GitHub
artifact. I don't understand why we do that, pushing coverage staright
to devhub seems much more straightforward and useful.


sentientwaffle (approved):

matklad:
Heh, didn't work out. It _looks_ like tar is confused by symlinks or something, will take a look tomorrow!

PR #2288, opened 2024-09-06
vsr: fix VOPR false positive

matklad:
This is a bug specific to handling releases in VOPR, where release_execute isn't no-return, and so we end up trying to update the checkpoint twice.

### DIFF
--- a/src/vsr/replica.zig
+++ b/src/vsr/replica.zig
@@ -7887,6 +7887,7 @@ pub fn ReplicaType(
                 .updating_checkpoint => |checkpoint| {
                     if (checkpoint.header.op == self.op_checkpoint()) {
                         self.sync_superblock_update_finish();
+                        assert(self.syncing == .idle);
                     }
                 },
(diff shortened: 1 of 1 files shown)


src/vsr/replica.zig
```diff
@@ -7887,6 +7887,7 @@ pub fn ReplicaType(
                 .updating_checkpoint => |checkpoint| {
                     if (checkpoint.header.op == self.op_checkpoint()) {
                         self.sync_superblock_update_finish();
+                        assert(self.syncing == .idle);
```
matklad, line 7890: "Surely this is a waste of time and can't fire" thought I when adding this assert during debugging!

chaitanyabhandari (approved):

matklad:
Forgot to include the seed into the commit message:

```
Seed: ./zig/zig build -Drelease -Dvopr-state-machine=testing vopr -- --lite 918954104495348913
```

PR #2290, opened 2024-09-06
VOPR branch: Testing if this code is dead

matklad:

PR #2291, opened 2024-09-06
lsm: refactor Forest.compact

matklad:
- make the fork/join structure more obvious and symmetric
- remove "The manifest log compaction completed between compaction beats" dead code. It seems like that was a part of incomplete manifest pacing, which didn't make in. We do run manifest compaction in a single beat today.

### DIFF
--- a/src/lsm/forest.zig
+++ b/src/lsm/forest.zig
@@ -223,13 +223,19 @@ pub fn ForestType(comptime _Storage: type, comptime groove_cfg: anytype) type {
             },
         } = null,
 
-        compaction_progress: enum { idle, trees_or_manifest, trees_and_manifest } = .idle,
+        compaction_progress: ?struct {
+            trees_done: bool,
+            manifest_log_done: bool,
+
+            fn all_done(compaction_progress: @This()) bool {
+                return compaction_progress.trees_done and compaction_progress.manifest_log_done;
+            }
+        } = null,
 
         grid: *Grid,
         grooves: Grooves,
         node_pool: NodePool,
         manifest_log: ManifestLog,
-        manifest_log_progress: enum { idle, compacting, done, skip } = .idle,
 
         compaction_pipeline: CompactionPipeline,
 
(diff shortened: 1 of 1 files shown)


src/lsm/forest.zig
```diff
@@ -514,47 +540,24 @@ pub fn ForestType(comptime _Storage: type, comptime groove_cfg: anytype) type {
...
-            assert(forest.manifest_log_progress == .compacting);
-
-            forest.manifest_log_progress = .done;
-
-            if (forest.progress) |progress| {
-                assert(progress == .compact);
-
-                forest.compact_callback();
-            } else {
-                // The manifest log compaction completed between compaction beats.
```
matklad, line 546: This branch is now an assert. VOPRing that specifically at https://github.com/tigerbeetle/tigerbeetle/pull/2290

src/lsm/forest.zig
```diff
@@ -425,41 +437,52 @@ pub fn ForestType(comptime _Storage: type, comptime groove_cfg: anytype) type {
...
+            if (forest.compaction_progress.?.all_done()) {
+                forest.compact_finish();
+            }
+        }
+
+        fn compact_finish(forest: *Forest) void {
+            assert(forest.progress.? == .compact);
+            assert(forest.compaction_progress != null);
+            assert(forest.compaction_progress.?.trees_done == true);
+            assert(forest.compaction_progress.?.manifest_log_done == true);
```
matklad: I had `defer assert(forest.compaction_progress == null)` here, which was failing in VOPR, and which motivated #2294.

But I think now that this assert is  simply incorrect, and changes in #2294 are not actuall relevant. 

What happens is this:

* some grid completes asynchronously and invokes its callback
* it turns out to be the last write of a compaction, so `compact_finish` gets called
* that in turn invokes replica-level callback at the end of commit loop
* the loop goes to the next iteration and starts committing next prepare
* there's nothing to exectute for the next prepare, so we go straight to compacting.
* so, on the callstack, we have something like

```
grid_write_callback
compact_finish
commit_dispatch
compact
```

This is weird, but I don't think it is wrong.

src/lsm/forest.zig
```diff
@@ -425,40 +437,50 @@ pub fn ForestType(comptime _Storage: type, comptime groove_cfg: anytype) type {
...
```
sentientwaffle: Nit, but this could just be
```suggestion
            assert(forest.compaction_progress.?.trees_done);
            assert(forest.compaction_progress.?.manifest_log_done);
```

sentientwaffle (dismissed):

sentientwaffle (approved):

PR #2292, opened 2024-09-06
vsr: remove -Dconfig

matklad:
There's no need to specify config on the CLI --- production config is all you need!

We still need a way to override config for tests and fuzzing, but we do that from-within the source.


cb22 (approved):
:rocket:

PR #2294, opened 2024-09-07
vsr: refactor commit_dispatch loop 

matklad:
Rewrite commit_dispatch to more closely resemble straight line  code.

 # Motivation

While I like the general idea of commit_dispatch, it feels like our current implementation isn't quite as obvious as it needs to be:

* Each stage is responsible for calling the next stage (you pass an argument to .commit_dispatch), which makes it hard to know in which order things actually happen. They sort-of happen in the order of the switch prongs in the dispatch, but it's not immediately clear if we are not being clever and skipping some of the stages sometimes.
* I always have trouble finding the place where we _actually_ execute things. The entry point is `setup_client_reply_callback`, which is the least likely springboard for the actual execution!
* There's some confusion about "sometimes asynchronous" stages, up to the point where we get unexpected-reentracy _despite_ having an explicit comment in the source code about _not_ doing that.

 # Implementation

The new approach is an attempt to do the most direct thing possible. If programming were actually easy, we'd write our commit path as, roughly

```
loop {
  prefetch()
  execute()
  compact()
}
```

The problem is that programming is, in fact, not easy, and the principle difficulty here is that most of parts of the loop are asynchronous. Were we using a language with async-await, we would have written:

```
loop {
  prefetch().await
  execute().await
  compact().await
}
```

We don't have async-await, but we can express the same logic manually. `prefetch` being async means that `prefetch` might not actually complete immediately, and instead just schedule some background work. Which means that we need to "pause" our commit work and return to it later.

Pausing is easily achieved by early returning:

```
loop {
  if (prefetch() == .pending) return
  if (execute() == .pending) return
  if (compact() == .pending) return
}
```

Resumption is trickier, as we should be able to restart the loop from the middle. If we were writing in C, we could have done something like this:

```
switch(state) {
  loop {
case prefetch:
    if (prefetch() == .pending) return
case execute:
    if (execute() == .pending) return
case compact:
    if (compact() == .pending) return
  }
}
```

Luckily, Zig doesn't allow this kind of convoluted code (though I remain intrigued by the possibilities opened by the recently implemented https://github.com/ziglang/zig/issues/8220).

But we can simulate it easily enough by guarding each prong of the switch with an if:

```
loop {
  if (state == .init) {
    state == .prefetch
    if (prefetch() == .pending) return
  }
  if (state == .prefetch) {
    state == .execute
    if (execute() == .pending) return
  }
  if (state == .execute) {
    state == .compact
    if (compact() == .pending) return
  }
}
```

Which is exactly what are we doing here!

Specifically:

* `commit_dispatch` is the main commit loop. It is an asynchronous loop committing everything we can commit.
* `commit_dispatch_enter` enters the loop from the outside.
* if the loop schedules an asynchronous operation, it's callback should invoke `commit_dispatch_resume` without arguments, which will pick up the loop where we've left.
* All the "parts" of the `commit_dispatch` loop are right after it in the source code order and are named `commit_stage`. All callbacks are named `commit_stage_callback`

One interesting side-effect here is that the new structure will be able to accommodate our "optimistic prefetch" idea, where we _assume_ that everything is in cache and start commit _without_ prefetch, bailing out if we have a missing entry. The new loop works even if the entire commit path turns out to be synchronous!

### DIFF
--- a/src/vsr/replica.zig
+++ b/src/vsr/replica.zig
@@ -3681,35 +3681,71 @@ pub fn ReplicaType(
 
                 if (self.commit_stage == .check_prepare) {
                     self.commit_stage = .prefetch;
+                    self.trace.start(.replica_commit, .{
+                        .stage = @tagName(self.commit_stage),
+                        .op = self.commit_prepare.?.header.op,
+                    });
                     if (self.commit_prefetch() == .pending) return;
                 }
 
                 if (self.commit_stage == .prefetch) {
+                    self.trace.stop(.replica_commit, .{ .stage = @tagName(self.commit_stage) });
                     self.commit_stage = .reply_setup;
+                    self.trace.start(.replica_commit, .{
+                        .stage = @tagName(self.commit_stage),
+                        .op = self.commit_prepare.?.header.op,
+                    });
                     if (self.commit_reply_setup() == .pending) return;
                 }
 
                 if (self.commit_stage == .reply_setup) {
+                    self.trace.stop(.replica_commit, .{ .stage = @tagName(self.commit_stage) });
                     self.commit_stage = .execute;
+                    self.trace.start(.replica_commit, .{
+                        .stage = @tagName(self.commit_stage),
+                        .op = self.commit_prepare.?.header.op,
+                    });
+
                     self.commit_execute();
                 }
 
                 if (self.commit_stage == .execute) {
+                    self.trace.stop(.replica_commit, .{ .stage = @tagName(self.commit_stage) });
                     self.commit_stage = .compact;
+                    self.trace.start(.replica_commit, .{
+                        .stage = @tagName(self.commit_stage),
+                        .op = self.commit_prepare.?.header.op,
+                    });
+
                     if (self.commit_compact() == .pending) return;
                 }
 
                 if (self.commit_stage == .compact) {
+                    self.trace.stop(.replica_commit, .{ .stage = @tagName(self.commit_stage) });
                     self.commit_stage = .{ .checkpoint_data = .{} };
+                    self.trace.start(.replica_commit, .{
+                        .stage = @tagName(self.commit_stage),
+                        .op = self.commit_prepare.?.header.op,
+                    });
+
                     if (self.commit_checkpoint_data() == .pending) return;
                 }
 
                 if (self.commit_stage == .checkpoint_data) {
+                    self.trace.stop(.replica_commit, .{ .stage = @tagName(self.commit_stage) });
                     self.commit_stage = .checkpoint_superblock;
+                    self.trace.start(.replica_commit, .{
+                        .stage = @tagName(self.commit_stage),
+                        .op = self.commit_prepare.?.header.op,
+                    });
+
                     if (self.commit_checkpoint_superblock() == .pending) return;
                 }
 
                 if (self.commit_stage == .checkpoint_superblock) {
+                    self.trace.stop(.replica_commit, .{
+                        .stage = @tagName(self.commit_stage),
+                    });
                     self.commit_stage = .idle;
                     self.commit_finish();
 
--- a/src/vsr/replica.zig
+++ b/src/vsr/replica.zig
@@ -48,32 +48,6 @@ pub const Status = enum {
     recovering_head,
 };
 
-const CommitStage = union(enum) {
-    /// Not committing.
-    idle,
-    /// About to start committing.
-    next,
-    next_journal,
-    next_pipeline,
-    prefetch_state_machine,
-    /// Ensure that the ClientReplies has at least one Write available.
-    setup_client_replies,
-    compact_state_machine,
-    checkpoint_data: CheckpointDataProgress,
-    checkpoint_superblock,
-    /// A commit just finished. Clean up before proceeding to the next.
-    cleanup,
-};
-
-const CheckpointData = enum {
-    state_machine,
-    client_replies,
-    client_sessions,
-    grid,
-};
-
-const CheckpointDataProgress = std.enums.EnumSet(CheckpointData);
-
 pub const ReplicaEvent = union(enum) {
     message_sent: *const Message,
     state_machine_opened,
@@ -396,6 +370,7 @@ pub fn ReplicaType(
 
         /// Guards against concurrent commits, and tracks the commit progress.
         commit_stage: CommitStage = .idle,
+        commit_dispatch_entered: bool = false,
 
         /// Whether we are reading a prepare from storage to construct the pipeline.
         pipeline_repairing: bool = false,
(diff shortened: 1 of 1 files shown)


matklad:
Note: this is a draft. The code compiles, and passes _one_ vopr seed, but fails on others (for example, the re-entrancy check fires if journal fetches (absent) prepare synchroneously). I'd love to get some feedback on the idea before I charge ahead with adding missing asserts and ~removing~ fixing the failing ones!

src/vsr/replica.zig
```diff
@@ -3691,11 +3573,215 @@ pub fn ReplicaType(
...
+            assert(!self.commit_dispatch_entered);
+            self.commit_dispatch_entered = true;
+            defer self.commit_dispatch_entered = false;
+
+            if (self.syncing == .canceling_commit) {
+                self.sync_dispatch(.canceling_grid);
+            }
+
+            for (0..constants.vsr_checkpoint_ops) |_| {
+                if (self.commit_stage.transition(.start)) {
```
sentientwaffle: Maybe inline `transition`? I don't see a reason to abstract it out. 

e.g.

```
                if (self.commit_stage == .idle) {
                    self.commit_stage = .start;
                    assert(self.commit_prepare == null);
                    if (self.commit_start() == .pending) return;
                }
```

src/vsr/replica.zig
```diff
@@ -3691,11 +3573,215 @@ pub fn ReplicaType(
...
+        /// The code bellow is a manual desugaring of loop with the following structure:
+        /// loop {
+        ///     prefetch().await;
+        ///     execute().await;
+        ///     compact().await;
+        /// }
+        fn commit_dispatch(self: *Self) void {
+            assert(!self.commit_dispatch_entered);
+            self.commit_dispatch_entered = true;
+            defer self.commit_dispatch_entered = false;
```
sentientwaffle: Could this be set to `false` by `commit_dispatch_resume`, so that it continues to guard this function even while e.g. an async prefetch is in flight? (Or is there some other assert here that would guard against starting prefetch multiple times?)

sentientwaffle (commented):
A couple of minor suggestions, but overall this looks great!

src/vsr/replica.zig
matklad: Indeed!

src/vsr/replica.zig
matklad: Yeah, `commit_stage != .idle` is equivalent to your suggested `commit_dispatch_entered`. 

Though, I did follow your suggestion, because _not_ doing that actually results in failing asserts, because we do have re-entrancy here. When we fetch prepare from the journal, the callback can be called synchroneously if preapare doesn't have a body, or if it is missing. 

So both in the new code and in the old code, we _could_ enter `commit_dispatch` while already having a a running `commit_dispatch` on the call stack. I also want to fix that, but I am not sure I want to fix that in this PR

src/vsr/replica.zig
```diff
@@ -3691,11 +3601,224 @@ pub fn ReplicaType(
...
+            /// Get the next prepare to commit from the journal or pipeline and...
+            start,
+            /// ...if there isn't there, break out of commit loop.
+            check_prepare,
+            /// Load required data from LSM tree on disk into memory.
+            prefetch,
+            /// Ensure that the ClientReplies has at least one Write available.
+            reply_ready,
+            /// Execute state machine logic.
+            execute,
```
sentientwaffle, line 3622: :tada: Splitting out the synchronous execution step is great! (Before it was run during `setup_client_replies`, which confused me when I was checking traces.)

src/vsr/replica.zig
```diff
@@ -3691,11 +3601,224 @@ pub fn ReplicaType(
...
+        fn commit_dispatch_resume(self: *Self) void {
+            assert(self.commit_stage != .idle);
+            assert(self.commit_dispatch_entered);
+            self.commit_dispatch_entered = false;
+            self.commit_dispatch();
+        }
+
+        fn commit_dispatch_cancel(self: *Self) void {
+            assert(self.commit_stage != .idle);
+            self.commit_dispatch_entered = false;
```
```suggestion
```

(This is also the last line of the function.)

src/vsr/replica.zig
```diff
@@ -3691,11 +3601,224 @@ pub fn ReplicaType(
...
+                        self.commit_dispatch_entered = false;
+                        return;
+                    }
+                }
+                assert(self.commit_prepare == null);
+                assert(self.commit_stage == .idle);
+            } else unreachable;
+
+            assert(self.commit_stage == .check_prepare);
+            assert(self.commit_prepare == null);
```
sentientwaffle, line 3769:
```suggestion
            assert(self.commit_prepare == null);
            maybe(self.commit_dispatch_entered);
```

src/vsr/replica.zig
```diff
@@ -3691,11 +3601,224 @@ pub fn ReplicaType(
...
+            if (self.syncing == .canceling_commit) {
+                switch (self.commit_stage) {
+                    .start, .reply_ready, .checkpoint_data, .checkpoint_superblock => {
+                        self.sync_dispatch(.canceling_grid);
+                        return;
+                    },
+                    else => unreachable,
+                }
+            }
+
```
sentientwaffle, line 3666: Could you add back the tracing?

I'm not sure what the best way to do so is, since `commit_dispatch` is potentially recursive.

Which means something like this doesn't work:

```zig
            const stage_before: std.meta.Tag(CommitStage) = self.commit_stage;
            defer if (self.commit_stage != stage_before) {
                if (stage_before != .idle) {
                    self.trace.stop(.replica_commit, .{ .stage = @tagName(stage_before) });
                }
                    self.trace.start(.replica_commit, .{
                        .stage = @tagName(self.commit_stage),
                        .op = self.commit_min,
                    });
                }
            };
```

src/vsr/replica.zig
```diff
@@ -3691,11 +3601,224 @@ pub fn ReplicaType(
...
+            /// Not committing.
+            idle,
+            /// Get the next prepare to commit from the journal or pipeline and...
+            start,
+            /// ...if there isn't there, break out of commit loop.
+            check_prepare,
+            /// Load required data from LSM tree on disk into memory.
+            prefetch,
+            /// Ensure that the ClientReplies has at least one Write available.
+            reply_ready,
```
sentientwaffle: Or `reply_setup`? `reply_ready` sounds like the reply is _already_ available.

src/vsr/replica.zig
```diff
@@ -3559,99 +3534,34 @@ pub fn ReplicaType(
             return other_replica;
         }
 
-        fn commit_dispatch(self: *Self, stage_new: CommitStage) void {
-            assert(self.commit_min <= self.commit_max);
-            assert(self.commit_min <= self.op);
-
-            if (self.syncing == .canceling_commit and stage_new != .cleanup) {
-                return self.sync_cancel_commit_callback();
```
sentientwaffle, line 3561: `sync_cancel_commit_callback` is now dead code.

src/vsr/replica.zig
```diff
@@ -3691,11 +3601,224 @@ pub fn ReplicaType(
...
-            assert(self.commit_stage == .next_journal);
+        fn commit_dispatch_resume(self: *Self) void {
+            assert(self.commit_stage != .idle);
+            assert(self.commit_dispatch_entered);
+            self.commit_dispatch_entered = false;
+            self.commit_dispatch();
+        }
+
+        fn commit_dispatch_cancel(self: *Self) void {
+            assert(self.commit_stage != .idle);
```
sentientwaffle, line 3800: ~More strictly, can we assert that the `commit_stage` is either `.prefetch` or `.compact` ?~

Nevermind; we can't, because in `sync_dispatch` we jump from other (uninterruptible) states directly to `self.sync_dispatch(.canceling_grid);` without changing our `commit_stage`:

```zig
            if (self.syncing == .canceling_commit) {
                switch (self.commit_stage) {
                    .start, .reply_ready, .checkpoint_data, .checkpoint_superblock => {
                        self.sync_dispatch(.canceling_grid);
                        return;
                    },
                    else => unreachable,
                }
            }
```

src/vsr/replica.zig
matklad: I think this should be assert actually

src/vsr/replica.zig
matklad: Good suggestion, `_ready` also clashes with the `return .ready`

src/vsr/replica.zig
matklad: :+1: 

we need a tidy rule that checks that all non-pub decls are used within a file

src/vsr/replica.zig
matklad: Went with just manually opening&closing every span. We could add something like `commit_dispatch_transition` back to DRY this out, but I have not yet reached activation energy for that.

src/vsr/replica.zig
sentientwaffle: I can't remember why I thought it was a `maybe()` yesterday, but if the VOPR doesn't hit it then I was wrong anyhow. :shrug:

src/vsr/replica.zig
```diff
@@ -3685,11 +3595,260 @@ pub fn ReplicaType(
...
+
+            if (self.commit_prepare) |prepare| self.message_bus.unref(prepare);
+            self.trace.reset(.replica_commit);
+            self.commit_prepare = null;
+            self.commit_stage = .idle;
+            self.commit_dispatch_entered = false;
```
sentientwaffle, line 3807: Can we assert that `self.commit_dispatch_entered` is true before this?

src/vsr/replica.zig
sentientwaffle: Ahh, I was thinking: when `compact` calls its callback synchronously (and thus `commit_dispatch` is recursive), then when that unwinds, we the inner `commit_dispatch` would have already set `commit_dispatch_entered = false`. But that is fine, b/c in that case we wouldn't `break` the loop:

```
                if (self.commit_stage == .execute) {
                    self.commit_stage = .compact;

                    if (self.commit_compact() == .pending) return;
                }
```

So `assert` makes sense.

sentientwaffle (approved):

src/vsr/replica.zig
sentientwaffle: As an alternative to `commit_dispatch_transition`, we could just add `self.commit_trace_start();`:

```
                fn commit_trace_start(...) void {
                    self.trace.start(.replica_commit, .{
                        .stage = @tagName(self.commit_stage),
                        .op = self.commit_prepare.?.header.op,
                    });
                }
```

to DRY out `commit_dispatch`.

PR #2300, opened 2024-09-09
`AccountFilter` fields

batiati:
Expand AccountFilter fields.

- Add `user_data_{128,64,32}` and `code` to `AccountFilter` enabling filtering transfers and historical balances by these fields when using `get_account_transfers` and `get_account_balances`.


src/tigerbeetle.zig
sentientwaffle, line 1: What is the plan for upgrading/transitioning? It looks like right now this upgrade would require that `client_release_min == release`, i.e. simultaneous upgrade of cluster + clients.

batiati: Good point (I keep forgetting about upgrades :cry: )

Option 1. We can keep the first 64 bytes of `AccountFilter` unchanged and add the new fields after this offset, so old clients would be compatible with the new filter (requiring just a minor change on how we read the filter from the message body).

Option 2. Intermediate release capable of reading the old version of `AccountFilter` from the message body depending on the `client_release`, and converting it to the actual filter. We can drop this logic once we bump `client_release_min`.

src/tigerbeetle.zig
```diff
@@ -385,6 +385,20 @@ pub const QueryFilterFlags = packed struct(u32) {
...
     account_id: u128,
+    /// Filter by the `user_data_128` index.
+    /// Use zero for no filter.
+    user_data_128: u128,
+    /// Filter by the `user_data_64` index.
+    /// Use zero for no filter.
+    user_data_64: u64,
+    /// Filter by the `user_data_32` index.
+    /// Use zero for no filter.
+    user_data_32: u32,
```
sentientwaffle, line 396: Why not add `ledger` too?

src/tigerbeetle.zig
sentientwaffle: The latter seems like a bit of extra work, but will end up cleaner after a few release cycles. I don't have a strong preference, though.

src/tigerbeetle.zig
batiati: Because it’s already filtered by the account_id.

sentientwaffle: Ah right, thanks!

src/state_machine.zig
```diff
@@ -663,6 +667,7 @@ pub fn StateMachineType(
 
         pub fn prefetch(
             self: *StateMachine,
+            client_release: vsr.Release,
             callback: *const fn (*StateMachine) void,
```
sentientwaffle: Imo `callback` should be the first argument (by convention), then `client_release`.

sentientwaffle (approved):

PR #2304, opened 2024-09-11
Add Scan and Lookup Traces.

batiati:
Add new Traces:

- Lookup/LookupWorker (covers both prefetching and scan lookups)
- ScanTree/ScanTreeLevel

Note 1: [Spall](https://gravitymoth.com/spall/spall.html) sorts the "Thread ID" alphabetically, so it fails to render a usable visualization of this trace, as `ScanTree` and `ScanTreeLevel` are nested events (e.g., each `ScanTree` has up to `lsm_levels` `ScanTreeLevel`) and stack number (displayed as "Thread ID: ##") is the only thing that groups them together.

Note 2: For a more "illustrated" trace sample, try running the benchmark with the flags `--id-order=random` to skip the negative lookup optimization and `--transfer-pending` to scan by expired transfers.

```bash
./tigerbeetle benchmark --trace=trace.json --id-order=random --transfer-pending  
```

image

### DIFF
--- a/src/lsm/scan_buffer.zig
+++ b/src/lsm/scan_buffer.zig
@@ -39,10 +39,18 @@ pub const ScanBuffer = struct {
         }
     };
 
+    index: u8,
     levels: [constants.lsm_levels]LevelBuffer,
 
-    pub fn init(self: *ScanBuffer, allocator: Allocator) !void {
+    pub fn init(
+        self: *ScanBuffer,
+        allocator: Allocator,
+        options: struct {
+            index: u8,
+        },
+    ) !void {
         self.* = .{
+            .index = options.index,
             .levels = undefined,
         };
         for (&self.levels, 0..) |*level, i| {
@@ -73,9 +81,12 @@ pub const ScanBufferPool = struct {
             .scan_buffer_used = 0,
         };
 
-        for (&self.scan_buffers, 0..) |*scan_buffer, i| {
-            errdefer for (self.scan_buffers[0..i]) |*buffer| buffer.deinit(allocator);
-            try scan_buffer.init(allocator);
+        for (&self.scan_buffers, 0..) |*scan_buffer, index| {
+            errdefer for (self.scan_buffers[0..index]) |*buffer| buffer.deinit(allocator);
+            try scan_buffer.init(
+                allocator,
+                .{ .index = @intCast(index) },
+            );
         }
         errdefer for (&self.scan_buffers) |*buffer| buffer.deinit(allocator);
     }
--- a/src/lsm/scan_lookup.zig
+++ b/src/lsm/scan_lookup.zig
@@ -32,27 +32,11 @@ pub fn ScanLookupType(
         pub const Callback = *const fn (*ScanLookup, []const Object) void;
 
         const LookupWorker = struct {
+            index: u8,
             scan_lookup: *ScanLookup,
             lookup_context: Groove.ObjectTree.LookupContext = undefined,
             index_produced: ?usize = null,
         };
-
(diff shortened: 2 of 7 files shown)


src/lsm/scan_lookup.zig
```diff
@@ -32,28 +32,12 @@ pub fn ScanLookupType(
...
-        /// Reasoning: the larger the block size is, higher is the probability of multiple
-        /// lookups hitting the same grid block:
-        const lookup_workers_max = @max(
-            stdx.div_ceil(
-                @divFloor(constants.block_size, @sizeOf(Object)),
-                Grid.read_iops_max,
-            ),
-            Grid.read_iops_max,
-        );
-
```
batiati, line 56: This logic was removed:
- It simplified the mental model of always handling the max of `Grid.read_iops_max` parallel lookups.
- Also allowed using the same `trace` event for lookups from prefetch and scans.
- In practice, the benchmark didn't show any performance improvement.
- Reducing the number of `[max]LookupWorker` in the array also **reduced the size of the binary** :exploding_head:

src/trace.zig
```diff
@@ -161,14 +171,40 @@ pub const Event = union(enum) {
...
+            },
+            inline .grid_read, .grid_write => |data| {
+                assert(data.iop < event_stack_cardinality.get(event.*));
+                const stack_base = event_stack_base.get(event.*);
+                return stack_base + @as(u32, @intCast(data.iop));
+            },
+            inline else => |data| {
+                comptime assert(@TypeOf(data) == void);
+                return event_stack_base.get(event.*);
+            },
```
sentientwaffle: How about this:

```suggestion
            inline else => |data, event_tag| {
                comptime assert(@TypeOf(data) == void);
                return comptime event_stack_base.get(event_tag);
            },
```

Iiuc we use `inline else` rather than `else` so that we can assert `@TypeOf(data) == void`.

So since we are generating code for each variant anyway, may as well inline the stack base as a constant rather than needing to fetch it at runtime from the `EnumArray`.

src/trace.zig
```diff
@@ -161,14 +171,40 @@ pub const Event = union(enum) {
...
+                const scan_tree_level_offset = data.level + 1;
+                return stack_base + scan_tree_offset + scan_tree_level_offset;
+            },
+            inline .grid_read, .grid_write => |data| {
+                assert(data.iop < event_stack_cardinality.get(event.*));
+                const stack_base = event_stack_base.get(event.*);
+                return stack_base + @as(u32, @intCast(data.iop));
+            },
+            inline else => |data| {
+                comptime assert(@TypeOf(data) == void);
```
sentientwaffle, line 210: Also add `comptime assert(event_stack_cardinality.get(event_tag) == 1);`.

src/lsm/scan_buffer.zig
```diff
@@ -39,10 +39,12 @@ pub const ScanBuffer = struct {
         }
     };
 
+    index: u8,
     levels: [constants.lsm_levels]LevelBuffer,
 
-    pub fn init(self: *ScanBuffer, allocator: Allocator) !void {
```
sentientwaffle, line 44: Maybe pull `index` into an `options` struct, so that the call site is labeled?

sentientwaffle (commented):
Just a couple nits, otherwise looks great!

sentientwaffle (commented):
CFO found a failing seed due to tracing start/stop asserts: `./zig/zig build -Drelease vopr -- --lite 15404620186638443928`. I haven't looked into the details but I'm guessing it is missing a `.reset()` somewhere.

sentientwaffle (approved):

PR #2305, opened 2024-09-12
Fix documentation bug for Query Transfers

jorispz:
Documentation used AccountFilter flags instead of QueryFilter


matklad:
Well spotted @jorispz thanks! 

There's actually a bug in our sample code as well, I added a fix for that on top of your commit in https://github.com/tigerbeetle/tigerbeetle/pull/2306.

I am closing this PR as 2306 builds directly on it!

PR #2306, opened 2024-09-12
node: fix type error in the sample

matklad:


matklad:
Ideally, we'd make type-script detect this error, but I don't see a simple way to do that

matklad:
I wonder if we somehow could make this into a real type or runtime error? Our flags in TS are ints, and it feels exceedingly easy to confuse different ints together.

maybe we should be saying 

```ts
flags: {
    reversed: true
}
```

instead?

This is JS, so we are going to do marshalling anyway, so theer's no harm in using proper objects for flags

batiati:
> maybe we should be saying
> 
> ```ts
> flags: {
>     reversed: true
> }
> ```
> 
> instead?

:+1: Actually there's an open activity or this in the [RoadMap](https://github.com/tigerbeetle/tigerbeetle/issues/259)
image

batiati (approved):

PR #2307, opened 2024-09-12
LSM: Human-readable tree names

sentientwaffle:
Tree names are used for logging and tracing -- no need for super-verbose fully-qualified labels.

So for example, instead of a tree name like:

    state_machine.StateMachineType(
      storage.Storage(io.linux.IO),
      .{
        .release = .{ ... },
        .message_body_size_max = 1048320,
        .lsm_compaction_ops = 32,
        .vsr_operations_reserved = 128,
      },
    ).AccountBalancesGrooveValue

abbreviate to:

    AccountBalancesGrooveValue

### DIFF
--- a/src/lsm/groove.zig
+++ b/src/lsm/groove.zig
@@ -615,7 +615,7 @@ pub fn GrooveType(
                 // single scope is batch_value_count_limit (total – not per beat).
                 .scope_value_count_max = options.tree_options_object.batch_value_count_limit,
 
-                .name = @typeName(Object),
+                .name = comptime tree_name(Object),
             });
             errdefer groove.objects_cache.deinit(allocator);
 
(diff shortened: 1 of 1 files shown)


sentientwaffle:
Ah, just noticed https://github.com/tigerbeetle/tigerbeetle/pull/2304/commits/714014763b7594bf189e2021e9c46a5518018975 already solves this :facepalm:

batiati:
> Ah, just noticed [7140147](https://github.com/tigerbeetle/tigerbeetle/commit/714014763b7594bf189e2021e9c46a5518018975) already solves this 🤦

Your solution is also great because it removes the noisy namespace from the name.

sentientwaffle:
Ah, good point!

batiati (approved):

PR #2311, opened 2024-09-13
vsr: primary abdicates if its clock is out of sync

matklad:
This existing behavior is wrong:

If primary clock is out of sync (for example, all pongs are dropped), it drops all incoming requests, but otherwise appears fully functional to the backups (commits still go through). This way, a primary with broken clock can deadlock the cluster.

To fix this, let the primary abdicate if it is unable to convert a request to prepare.

Re-use existing abdication infrastructure for this purpose, relaxing the asserts. I went back-and-forth on the implementation strategy. The alternative is to introduce a separate `primary_abdicate_clock` timeout, but this _also_ requires a separate `self.abdicating_clock` boolean field (otherwise it is not reset correctly), which ends up duplicating a whole bunch of asserts for no big gain.

So, in the current implementation, the timeout is started when either we pipeline a prepare or we drop a request, and rest on every prepare_ok.

Resetting clock sync condition on the prepare_ok feels a bit odd, but I think that is fine, as we are still making progress.

### DIFF
--- a/src/vsr/replica.zig
+++ b/src/vsr/replica.zig
@@ -411,7 +411,8 @@ pub fn ReplicaType(
         /// Used when the primary believes that it is partitioned and needs to step down.
         /// In particular, guards against a deadlock in the case where small messages (e.g.
         /// heartbeats, pings/pongs) succeed, but large messages (e.g. prepares) fail.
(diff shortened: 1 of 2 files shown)


matklad:
Kudos to @owickstrom for pointing out clock as an area of interest!

sentientwaffle (approved):

PR #2295, opened 2024-09-08
vopr: non faulty replica shouldn't enter recovering_head with op < op_checkpoint

chaitanyabhandari:
Solves a failing seed (`./zig/zig build -Drelease vopr -- --lite 4766636330770358721`) found on this branch (since it involves changing a VOPR condition) wherein a replica enters a recovering_head post startup even if faults are disabled.

#### Problem

Imagine a potential primary handling a do_view_change. Currently, the potential primary replica first updates its in-memory log_view and initiates repair. During repair, the replica could repair some ops around the checkpoint and its trigger and advance its checkpoint. If such a replica crashes before making its log_view durable, it would restart with a regressed log_view and the TruncateViewRange condition would truncate all ops that were repaired (since their view # is greater than the log_view the replica restarts with). This truncation could regress self.op below self.op_checkpoint(), which could lead to a non faulty replica [starting up in the recovering_head state](https://github.com/tigerbeetle/tigerbeetle/blob/bc4a71bc5fa93de1f33403b0ea1d7a1f176e565e/src/vsr/replica.zig#L770-L787).

#### Solution

Potential primaries now initiate making their view_headers durable *before* they start repair. These headers are computed from the journal *after* the potential primary writes the headers from the DVC quorum to its journal using [self.primary_set_log_from_do_view_change_messages()](https://github.com/tigerbeetle/tigerbeetle/blob/bc4a71bc5fa93de1f33403b0ea1d7a1f176e565e/src/vsr/replica.zig#L2153). Therefore, if the potential primary advances its checkpoint during repair, the checkpoint is guaranteed to be written to the superblock *after* the view_headers (since writes to the superblock are serialized). 

After this change, we can never encounter a scenario on startup where `self.log_view < self.view` and `self.op < self.op_checkpoint()`.

PR #2313, opened 2024-09-13
vsr: non faulty replica shouldn't enter recovering_head with op < op_checkpoint

chaitanyabhandari:
### Problem

See https://github.com/tigerbeetle/tigerbeetle/pull/2295 for broader context.

Imagine a potential primary handling a do_view_change. Currently, the potential primary replica first updates its in-memory log_view and initiates repair. During repair, the replica could repair some ops around the checkpoint and its trigger and advance its checkpoint. If such a replica crashes before making its log_view durable, it would restart with a regressed log_view and the TruncateViewRange condition would truncate all ops that were repaired (since their view # is greater than the log_view the replica restarts with). This truncation could regress self.op below self.op_checkpoint().

Backups may be vulnerable to the same state as well. Specifically, imagine a case where a backup in view V is handling a start_view message for view V+1. In this case, the backup first transitions to view_change, invokes view_durable_update to make view=V +1 durable. Then, the backup transitions to normal, invokes view_durable_update to make log_view=V+1 and start_view headers durable. However, this invocation of view_durable_update ends up being a no-op if the earlier view_durable_update is still underway. Finally, if the backup initiates repair and advances to the next checkpoint _before_ log_view=V+1 is made durable, it could restart into a state wherein self.op regresses below self.op_checkpoint(). 

These cases where self.op regresses below self.op_checkpoint() could lead to a non faulty replica [starting up in the recovering_head state](https://github.com/tigerbeetle/tigerbeetle/blob/bc4a71bc5fa93de1f33403b0ea1d7a1f176e565e/src/vsr/replica.zig#L770-L787). 

### Solution

We solve this problem by making view_headers, view & log_view durable during `superblock.checkpoint`. Before this PR, they were only made durable during `superblock.view_change`. This entails regenerating start_view headers _before_ `superblock.checkpoint` is invoked, to ensure that durable view_headers *always* contain on op from the checkpoint that a replica starts with.

### DIFF
--- a/src/vsr/replica.zig
+++ b/src/vsr/replica.zig
@@ -4849,7 +4849,10 @@ pub fn ReplicaType(
             // Determine the consecutive extent of the log that we can help recover.
             // This may precede op_repair_min if we haven't had a view-change recently.
             const range_min = (op_hash_chain_verified + 1) -| constants.journal_slot_count;
-            const range = self.journal.find_latest_headers_break_between(range_min, op_hash_chain_verified);
+            const range = self.journal.find_latest_headers_break_between(
+                range_min,
+                op_hash_chain_verified,
+            );
             const op_min = if (range) |r| r.op_max + 1 else range_min;
             assert(op_min <= op);
             assert(op_min <= self.op_repair_min());
--- a/src/vsr/replica.zig
+++ b/src/vsr/replica.zig
@@ -708,7 +708,7 @@ pub fn ReplicaType(
                 }
             } else {
                 // This case can occur if we loaded an SV for its hook header but never finished
-                // that SV to a DVC (dropping the hooks), but never finished the view change.
+                // that SV to a DVC (dropping the hooks), and never finished the view change.
                 if (op_head == null) {
                     assert(self.view > self.log_view);
                     op_head = self.journal.op_maximum();
@@ -744,21 +744,21 @@ pub fn ReplicaType(
                     self.transition_to_normal_from_recovering_status();
                 }
             } else {
-                // Even if op_head_certain() returns false, a DVC always has a certain head op.
-                if (self.log_view == self.view and !self.op_head_certain()) {
-                    self.transition_to_recovering_head_from_recovering_status();
-                } else {
-                    if (self.log_view == self.view) {
+                if (self.log_view == self.view) {
+                    if (self.op_head_certain()) {
                         if (self.primary_index(self.view) == self.replica) {
                             self.transition_to_view_change_status(self.view + 1);
                         } else {
                             self.transition_to_normal_from_recovering_status();
                         }
                     } else {
-                        assert(self.view > self.log_view);
-                        assert(self.op >= self.op_checkpoint());
-                        self.transition_to_view_change_status(self.view);
+                        self.transition_to_recovering_head_from_recovering_status();
                     }
+                } else {
+                    // Even if op_head_certain() returns false, a DVC always has a certain head op.
+                    assert(self.view > self.log_view);
+                    assert(self.op >= self.op_checkpoint());
+                    self.transition_to_view_change_status(self.view);
                 }
             }
 
@@ -4257,23 +4257,28 @@ pub fn ReplicaType(
             // The only exception to this is a replica that arrived at a checkpoint via state
             // sync.
             if (self.view == self.log_view) {
-                // Unconditionally convert potential primary's DVC headers -> SV headers; they
-                // may contain truncated ops. For all other cases, update SV headers only
-                // if they aren't already up to date.
+                // Unconditionally convert potential primary's DVC headers → SV headers; the DVC
+                // headers may contain truncated ops which should not be made durable. For all other
+                // cases, update SV headers only if they aren't already up to date.
                 if ((self.status == .view_change and
                     self.primary_index(self.view) == self.replica) or
-                    self.view_headers.array.get(0).op < self.op)
+                    self.view_headers.array.get(0).op < self.op_checkpoint_next_trigger())
                 {
                     self.view_headers.command = .start_view;
                     self.update_start_view_headers();
-                    assert(self.view_headers.array.get(0).op == self.op);
                 }
+                assert(self.view_headers.command == .start_view);
+                assert(self.view_headers.array.get(0).op >= self.op_checkpoint_next_trigger());
+            } else {
+                // Replica moved to a future view before checkpoint completed; no need to update its
+                // headers, it would've stored its head op in its DVC headers while moving to
+                // view_change.
+                assert(self.view_headers.command == .do_view_change);
+                assert(self.view_headers.array.get(0).op >= self.op);
             }
 
-            assert(self.view_headers.array.get(0).op >= self.op);
-
             log.debug("{}: commit_checkpoint_superblock: checkpoint_superblock start " ++
-                "(op={} current_checkpoint={} next_checkpoint={} view_durable={}..{} " ++
+                "(op={} checkpoint={}..{} view_durable={}..{} " ++
                 "log_view_durable={}..{})", .{
                 self.replica,
                 self.op,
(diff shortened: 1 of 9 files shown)


src/vsr/replica.zig
```diff
@@ -707,9 +707,8 @@ pub fn ReplicaType(
                     break;
                 }
             } else {
-                // This case can only occur if we loaded an SV for its hook header, then converted
-                // that SV to a DVC (dropping the hooks; see start_view_into_do_view_change()),
-                // but never finished the view change.
+                // This case can occur if we loaded an SV for its hook header but never finished
+                // that SV to a DVC (dropping the hooks), but never finished the view change.
```
```suggestion
                // This case can occur if we loaded an SV for its hook header but never finished
                // that SV to a DVC (dropping the hooks), and never finished the view change.
```

"but" twice reads awkwardly.

src/vsr/replica.zig
```diff
@@ -757,10 +756,9 @@ pub fn ReplicaType(
                         }
                     } else {
                         assert(self.view > self.log_view);
+                        assert(self.op >= self.op_checkpoint());
```
sentientwaffle: I think this can be refactored to be clearer:

```
                if (self.log_view == self.view) {
                    if (self.op_head_certain()) {
                        if (self.primary_index(self.view) == self.replica) {
                            self.transition_to_view_change_status(self.view + 1);
                        } else {
                            self.transition_to_normal_from_recovering_status();
                        }
                    } else {
                        self.transition_to_recovering_head_from_recovering_status();
                    }
                } else {
                    assert(self.view > self.log_view);
                    assert(self.op >= self.op_checkpoint());
                    self.transition_to_view_change_status(self.view);
                }
```
This way the `self.log_view == self.view` comparison is in one place, and it avoids compound conditions.

And maybe move the `// Even if op_head_certain() returns false, a DVC always has a certain head op.` comment down into that last `else` branch?

src/vsr/replica.zig
```diff
@@ -4254,11 +4252,55 @@ pub fn ReplicaType(
...
+                    self.view_headers.command = .start_view;
+                    self.update_start_view_headers();
+                    assert(self.view_headers.array.get(0).op == self.op);
+                }
+            }
+
+            assert(self.view_headers.array.get(0).op >= self.op);
+
+            log.debug("{}: commit_checkpoint_superblock: checkpoint_superblock start " ++
+                "(op={} current_checkpoint={} next_checkpoint={} view_durable={}..{} " ++
```
```suggestion
                "(op={} checkpoint={}..{} view_durable={}..{} " ++
```

src/vsr/replica.zig
```diff
@@ -4254,11 +4252,55 @@ pub fn ReplicaType(
...
             }
 
+            // Update view_headers to include at least one op from the future checkpoint. This
+            // ensures a replica never starts with its head op less than self.op_checkpoint().
+            // The only exception to this is a replica that arrived at a checkpoint via state
+            // sync.
+            if (self.view == self.log_view) {
+                // Unconditionally convert potential primary's DVC headers -> SV headers; they
+                // may contain truncated ops. For all other cases, update SV headers only
+                // if they aren't already up to date.
```
sentientwaffle: > ; they may contain truncated ops.

This feels like an incomplete thought to me -- maybe follow with why that is okay?

src/vsr/replica.zig
```diff
@@ -4796,28 +4837,35 @@ pub fn ReplicaType(
             return message.ref();
         }
 
-        fn primary_update_view_headers(self: *Self) void {
-            assert(self.status != .recovering_head);
```
sentientwaffle, line 4800: :thinking: Does this assert no longer apply?

src/vsr/replica.zig
```diff
@@ -4796,28 +4837,35 @@ pub fn ReplicaType(
...
-            assert(self.replica == self.primary_index(self.view));
-            assert(self.view == self.log_view);
```
sentientwaffle, line 4802: It looks like this assert still applies.

src/vsr/replica.zig
```diff
@@ -8319,16 +8383,21 @@ pub fn ReplicaType(
...
             assert(self.primary());
 
-            // SVs will be sent out after the view_durable update completes.
-            assert(self.view_durable_updating());
-            assert(self.log_view > self.log_view_durable());
+            // SVs will be sent out after the view_durable update completes. view-change headers
+            // and log_view may already be durable if checkpoint was advanced during repair.
+            if (self.log_view > self.log_view_durable()) {
+                assert(self.view_durable_updating());
+                assert(self.log_view > self.log_view_durable());
```
sentientwaffle: This assert is redundant; it is in a `if (self.log_view > self.log_view_durable()) {` branch.

src/vsr/replica.zig
```diff
@@ -7979,6 +8036,19 @@ pub fn ReplicaType(
             assert(self.view_durable_updating());
         }
 
+        fn primary_send_start_view_after_view_change(self: *Self) void {
+            assert(self.primary());
```
sentientwaffle, line 8053:
```suggestion
            assert(self.status == .normal);
            assert(self.primary());
            assert(self.log_view == self.view);
```

`.primary()` also asserts `.status == .normal` so that is redundant, but imo it is nice for readability to assert it separately.

src/vsr/replica.zig
chaitanyabhandari: Issue with the way GitHub has organized these changes, perhaps... We still have this assert: https://github.com/tigerbeetle/tigerbeetle/pull/2313/files#diff-e95e4dc6c1aae349e798cfeed22ad01f830cce0e654e6b4b765b3e7b0c677b6aR4887-R4892

src/vsr/replica.zig
chaitanyabhandari: https://github.com/tigerbeetle/tigerbeetle/pull/2313/files#r1765584128

src/vsr/replica.zig
chaitanyabhandari: Awesome, this indeed is a much cleaner version!

src/vsr/replica.zig
sentientwaffle: Right, I meant as part of `update_start_view_headers` though. Likewise for https://github.com/tigerbeetle/tigerbeetle/pull/2313/files#r1765584792

chaitanyabhandari: Ah, misunderstood! Makes sense to add them to `update_start_view_headers ` too, will do. :D

src/vsr/superblock.zig
```diff
@@ -812,6 +812,11 @@ pub fn SuperBlockType(comptime Storage: type) type {
...
         const UpdateCheckpoint = struct {
             header: vsr.Header.Prepare,
+            view_attributes: ?struct {
+                log_view: u32,
+                view: u32,
+                headers: *const vsr.Headers.ViewChangeArray,
+            },
             commit_max: u64,
             sync_op_min: u64,
             sync_op_max: u64,
```
sentientwaffle, line 822: In `fn checkpoint`, if `view_attributes` is passed:
- `assert(log_view <= view);`
- `headers.verify()`

src/vsr/replica.zig
```diff
@@ -4254,11 +4252,60 @@ pub fn ReplicaType(
...
+                        // view_headers for solo replicas do not include ops that are not durable in
+                        // their journal.
+                        break :view_attributes if (self.solo())
+                            null
+                        else
+                            .{
+                                .headers = &self.view_headers,
+                                .view = self.view,
+                                .log_view = self.log_view,
+                            };
```
sentientwaffle, line 4307: For `solo()` replicas, even if this update isn't strictly necessary, could we allow it anyway, to avoid the added dimensionality?

src/vsr/replica.zig
```diff
@@ -8319,16 +8391,20 @@ pub fn ReplicaType(
 
             self.view_headers.command = .start_view;
             self.primary_update_view_headers();
-            self.view_headers.verify();
```
sentientwaffle, line 8322: Does this assertion no longer pass?

chaitanyabhandari: I moved this assertion into `update_start_view_headers` as its called from multiple places!

src/vsr/replica.zig
chaitanyabhandari: I added this after the VOPR pointed me towards a failed seed , seems like it _is_ required for `solo()` replicas... What happened there was that a solo replica's durable headers contained a header whose prepare was not yet durable, and it hit `@panic("cannot recover corrupt prepare");` while committing on startup:
```
            if (prepare == null) {
                log.debug("{}: commit_start_journal_callback: prepare == null", .{self.replica});
                if (self.solo()) @panic("cannot recover corrupt prepare");
            }
```

sentientwaffle: That makes sense, thanks!

src/vsr/superblock.zig
chaitanyabhandari: Took care of this in https://github.com/tigerbeetle/tigerbeetle/pull/2313/commits/982b84985c985e9b17fc060ef8c3cf135d6af774!

sentientwaffle (approved):

PR #2314, opened 2024-09-13
Reduce client libraries verbosity on stderr

batiati:
The main goal is to set the client libraries `log_level` as `.err` to reduce noise in user space: only fatal logs should be logged.

To achieve this goal, also we had to change `log.err` to `log.warn` in many places where non-fatal messages were logged (even outside client code, to standardize this pattern).

The behavior running with `log_level = .info` is unchanged.

From std.log docs:
warn:
 "Log a warning message. This log level is intended to be used if it is uncertain whether something has gone wrong or not, but the circumstances would be worth investigating."

 err:
 "Log an error message. This log level is intended to be used when something has gone wrong. This might be recoverable or might be followed by the program exiting."

### DIFF
--- a/src/multiversioning.zig
+++ b/src/multiversioning.zig
@@ -1047,7 +1047,7 @@ pub const Multiversion = struct {
         // While these look like blocking IO operations, on a memfd they're memory manipulation.
         // TODO: Would panic'ing be a better option? On Linux, these should never fail. On other
         // platforms where target_fd might be backed by a file, they could...
-        errdefer log.err("target binary update failed - " ++
+        errdefer log.warn("target binary update failed - " ++
             "this replica might fail to automatically restart!", .{});
 
         const target_file = std.fs.File{ .handle = self.target_fd };
@@ -1063,7 +1063,7 @@ pub const Multiversion = struct {
     fn handle_error(self: *Multiversion, result: anyerror) void {
         assert(self.stage != .init);
 
-        log.err("binary does not contain valid multiversion data: {}", .{result});
+        log.warn("binary does not contain valid multiversion data: {}", .{result});
 
         self.stage = .{ .err = result };
     }
--- a/src/multiversioning.zig
+++ b/src/multiversioning.zig
@@ -1063,7 +1063,7 @@ pub const Multiversion = struct {
     fn handle_error(self: *Multiversion, result: anyerror) void {
         assert(self.stage != .init);
 
-        log.warn("binary does not contain valid multiversion data: {}", .{result});
+        log.err("binary does not contain valid multiversion data: {}", .{result});
 
         self.stage = .{ .err = result };
     }
--- a/src/shell.zig
+++ b/src/shell.zig
@@ -522,9 +522,9 @@ pub fn exec_stdout_options(
     try child.collectOutput(&stdout, &stderr, options.output_bytes_max);
     const term = try child.wait();
     errdefer {
-        log.warn("command failed", .{});
+        log.err("command failed", .{});
         echo_command(argv.slice());
-        log.warn("stdout:\n{s}\nstderr:\n{s}", .{ stdout.items, stderr.items });
+        log.err("stdout:\n{s}\nstderr:\n{s}", .{ stdout.items, stderr.items });
     }
 
     switch (term) {
--- a/src/shell.zig
+++ b/src/shell.zig
@@ -193,7 +193,7 @@ pub fn env_get_option(shell: *Shell, var_name: []const u8) ?[]const u8 {
 
 pub fn env_get(shell: *Shell, var_name: []const u8) ![]const u8 {
     errdefer {
-        log.err("environment variable '{s}' not defined", .{var_name});
+        log.warn("environment variable '{s}' not defined", .{var_name});
     }
 
     return try std.process.getEnvVarOwned(shell.arena.allocator(), var_name);
@@ -372,7 +372,7 @@ pub fn copy_path(
     dst_path: []const u8,
 ) !void {
     errdefer {
-        log.err("failed to copy {s} to {s}", .{ src_path, dst_path });
+        log.warn("failed to copy {s} to {s}", .{ src_path, dst_path });
     }
     if (std.fs.path.dirname(dst_path)) |dir| {
         try dst_dir.makePath(dir);
@@ -522,9 +522,9 @@ pub fn exec_stdout_options(
     try child.collectOutput(&stdout, &stderr, options.output_bytes_max);
     const term = try child.wait();
     errdefer {
-        log.err("command failed", .{});
+        log.warn("command failed", .{});
         echo_command(argv.slice());
-        log.err("stdout:\n{s}\nstderr:\n{s}", .{ stdout.items, stderr.items });
+        log.warn("stdout:\n{s}\nstderr:\n{s}", .{ stdout.items, stderr.items });
     }
 
     switch (term) {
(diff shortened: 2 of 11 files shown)


src/vsr/replica.zig
```diff
@@ -1393,7 +1393,7 @@ pub fn ReplicaType(
             }
 
             if (self.loopback_queue) |loopback_message| {
-                log.err("{}: on_message: on_{s}() queued a {s} loopback message with no flush", .{
```
matklad, line 1396: This one sounds like it should be an assert? Could you file this separately & vopr tag it?

src/multiversioning.zig
```diff
@@ -1063,7 +1063,7 @@ pub const Multiversion = struct {
     fn handle_error(self: *Multiversion, result: anyerror) void {
         assert(self.stage != .init);
 
-        log.err("binary does not contain valid multiversion data: {}", .{result});
+        log.warn("binary does not contain valid multiversion data: {}", .{result});
```
matklad: I'd maybe leave both of these as errors --- during upgrades, it feels better to ping proactively

src/shell.zig
```diff
@@ -522,9 +522,9 @@ pub fn exec_stdout_options(
     try child.collectOutput(&stdout, &stderr, options.output_bytes_max);
     const term = try child.wait();
     errdefer {
-        log.err("command failed", .{});
+        log.warn("command failed", .{});
         echo_command(argv.slice());
-        log.err("stdout:\n{s}\nstderr:\n{s}", .{ stdout.items, stderr.items });
+        log.warn("stdout:\n{s}\nstderr:\n{s}", .{ stdout.items, stderr.items });
     }
```
matklad: These all are inside `errdefers`, so we are already propagating process-terminating error upwards here and, as such, `err` feels more appropriate.

This is in contrast to message bus where, although the source of errors is _also_ syscalls returning error code, there's a natural retry loop around the whole thing, which makes an individual error non-fatal.

No strong feelings here though!

src/storage.zig
```diff
@@ -268,7 +268,7 @@ pub fn Storage(comptime IO: type) type {
                         return;
                     } else {
                         // We tried to read at (or less than) logical sector granularity and failed.
-                        log.err("latent sector error: offset={}, zeroing sector...", .{
+                        log.warn("latent sector error: offset={}, zeroing sector...", .{
                             read.offset,
```
matklad, line 272: I am unsure about this one --- we expect disk errors to be _very_ rare, so perhaps it's worth keeping it as an `err`?

I guess there's also an angle that it would be _extremely_ interesting to learn how frequently do we get this, so artificially increasing loudness here might be a good thing?

As in, every time someone is paged over "latent sector error", they would feel relieved that they are using TigerBeetle :rofl:

src/vsr/clock.zig
```diff
@@ -516,7 +519,7 @@ pub fn ClockType(comptime Time: type) type {
                         fmt.fmtDurationSigned(delta),
                     });
                 } else {
-                    log.err("{}: system time is {} ahead, clamping system time to cluster time", .{
+                    log.warn("{}: system time is {} ahead, clamping system time to cluster time", .{
```
matklad, line 522: These _definitely_ should be warns imo, we are hitting them in antithesis!

src/vsr/grid.zig
```diff
@@ -1024,7 +1024,7 @@ pub fn GridType(comptime Storage: type) type {
             if (result != .valid) {
                 const header =
                     mem.bytesAsValue(vsr.Header.Block, block.*[0..@sizeOf(vsr.Header)]);
-                log.err(
```
matklad, line 1027: same comment as for latent sector errors -- although technically this is a recoverable failure that warrants a warn, this is _very_ curious, and perhaps is worth an .err just for that!

src/vsr/replica.zig
```diff
@@ -5269,7 +5269,10 @@ pub fn ReplicaType(
                 } else if (entry.session < message.header.session) {
                     // This cannot be because of a partition since we check the client's view
                     // number.
-                    log.err("{}: on_request: ignoring newer session (client bug)", .{self.replica});
+                    log.warn(
```
matklad: This feel like it warrants an `.err` --- if this _is_ a bug, we'd want our users to report this case (but this probably shouldn't be an outright assert, as we dno't want to trust clients __too__ much. Server is running with ECC RAM, but clients dont)

matklad: this applies to other "bug" cases

src/vsr/replica.zig
```diff
@@ -7630,7 +7633,7 @@ pub fn ReplicaType(
             }
 
             if (message.header.invalid()) |reason| {
-                log.err("{}: send_message_to_replica: invalid ({s})", .{ self.replica, reason });
```
matklad, line 7633: we are panicking just after, so .err feels appropriate!

matklad (commented):
:fearful: the fact that we print _anything_ to stderr from embeddable library is pretty much not good. As Andrew says, stdout, stderr, and stdin are owned by `fn main`. If you didn't write `fn main`, it's not your business to mess up with standard streams.

I wonder if we should override client logging to just not print anything at all, ever? That feels like a proper solution here.

---

Regardless, the changes to log levels look good to me in general. We are over-using errors. I am not a fan of how Zig docs describe the useage though:

If you describe log levels in terms of what they are _for_, you are bound to get philosophical discussing the levels of errorness.

My preferred way is to describe then in terms of what they _do_:

* `error` sends an alert out immediately
* `warn` sends an alert if it repeats several times in a short timespan

batiati:
@matklad any clue about why is it failing on CI ~trying to call `gh`~?
https://github.com/tigerbeetle/tigerbeetle/actions/runs/10856639050/job/30131676058?pr=2314

batiati:
> I wonder if we should override client logging to just not print anything at all, ever? That feels like a proper solution here.

Yes we should, but I don't see a problem with logging fatal errors anyway (since libraries _should not_ panic either, but we do in very unlikely scenarios)

matklad:
>Yes we should, but I don't see a problem with logging fatal errors anyway (since libraries should not panic either, but we do in very unlikely scenarios)

There might be a couple:

* stderr might be closed
* or the stderr might be buffered, so our output would corrupt whatever is the main process printing
* or stderr might be expected to be well-formed JSON or something like that
* or stderr might be /dev/null, and the proper "output stream" for diagnostic messages should be something else.
* or there might be other concurrent threads writing to stderr at this point (typically stderr is mutex protected, but Zig, and, say, Java, would use different mutexes)

In terms of theoretical "how it should be", just using global file descriptor 2 feels very wrong to me.

I admit that the _practical_ impact here is low, and might even be positive :-)

matklad:
>https://github.com/tigerbeetle/tigerbeetle/actions/runs/10856639050/job/30131676058?pr=2314

Huh, I've just seen integration tests failing at https://github.com/tigerbeetle/tigerbeetle/actions/runs/10856143117/job/30130219645

Not sure what's going on here, and if the two are related

batiati:
> There might be a couple:

A "couple" of five bullet points :rofl: 

That's fair, I think Java and Node that encapsulates the client inside the VM process we could wrap the log function to call the proper environment function (i.e. [env.fatal_error](https://github.com/tigerbeetle/tigerbeetle/blob/89761e4e13ff97494f085668979fd10690ab3aa7/src/clients/java/src/client.zig#L596)). For Go and C# we can just dump logs to a no-op function.

matklad (approved):

PR #2316, opened 2024-09-14
shell: make exec_ family of APIs more robust

matklad:
They now will always reliably print failing command, its stderr and stdout.

The key insight here is that Zig supports pollTimeout, so even if a process hangs, we could still kill it reliably and extract the output so far.

So, there's no reason to inherit the streams.

### DIFF
--- a/src/build_multiversion.zig
+++ b/src/build_multiversion.zig
@@ -786,16 +786,14 @@ fn self_check_enabled(target: Target) bool {
 
 fn self_check(shell: *Shell, multiversion: []const u8, past_releases: []const []const u8) !void {
     assert(past_releases.len > 0);
-    try shell.exec_options(
-        .{ .echo = false },
+    try shell.exec(
         "{multiversion} multiversion {multiversion}",
         .{ .multiversion = multiversion },
     );
     for (past_releases) |past_release| {
         // 0.15.3 didn't have the multiversion subcommand since it was the epoch.
         if (std.mem.indexOf(u8, past_release, "0.15.3") != null) continue;
-        try shell.exec_options(
-            .{ .echo = false },
+        try shell.exec(
             "{past_release} multiversion {multiversion}",
             .{ .multiversion = multiversion, .past_release = past_release },
         );
--- a/src/clients/dotnet/ci.zig
+++ b/src/clients/dotnet/ci.zig
@@ -11,8 +11,8 @@ const TmpTigerBeetle = @import("../../testing/tmp_tigerbeetle.zig");
 pub fn tests(shell: *Shell, gpa: std.mem.Allocator) !void {
     assert(shell.file_exists("TigerBeetle.sln"));
 
-    try shell.zig("build clients:dotnet -Drelease", .{});
-    try shell.zig("build -Drelease", .{});
+    try shell.exec_zig("build clients:dotnet -Drelease", .{});
+    try shell.exec_zig("build -Drelease", .{});
 
     try shell.exec("dotnet restore", .{});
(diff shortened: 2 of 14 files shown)


matklad:
Example output:

```
node ci
  error: process failed with ExecNonZeroExitStatus: npm lol
  error: stdout:
  ++++
  Unknown command: "lol"
  
  
  Did you mean this?
    npm ll # List installed packages
  To see a list of supported npm commands, run:
    npm help
  ++++
```

src/build_multiversion.zig
```diff
@@ -786,16 +786,14 @@ fn self_check_enabled(target: Target) bool {
 
 fn self_check(shell: *Shell, multiversion: []const u8, past_releases: []const []const u8) !void {
     assert(past_releases.len > 0);
-    try shell.exec_options(
-        .{ .echo = false },
```
matklad, line 790: I think there's no need to proactively echo things anymore, the command is printed iff it errors/hangs

src/clients/dotnet/ci.zig
```diff
@@ -11,8 +11,8 @@ const TmpTigerBeetle = @import("../../testing/tmp_tigerbeetle.zig");
 pub fn tests(shell: *Shell, gpa: std.mem.Allocator) !void {
     assert(shell.file_exists("TigerBeetle.sln"));
 
-    try shell.zig("build clients:dotnet -Drelease", .{});
-    try shell.zig("build -Drelease", .{});
+    try shell.exec_zig("build clients:dotnet -Drelease", .{});
```
matklad, line 14: added `exec_` prefix for consistency.

src/integration_tests.zig
```diff
@@ -240,12 +240,11 @@ test "benchmark/inspect smoke" {
     const shell = try Shell.create(std.testing.allocator);
     defer shell.destroy();
 
-    const status_ok_benchmark = try shell.exec_status_ok(
```
matklad, line 243: No need for separate `status` API, it is just `exec` now!

src/scripts/antithesis.zig
```diff
@@ -95,7 +95,7 @@ fn build_image(
                 \\
                 \\    cd {s} && TAG={s} docker compose up
                 \\
-                \\This temporary directory is not automatically deleted. You may do so
```
matklad, line 98: https://github.com/ziglang/zig/pull/20575 :-)

src/shell.zig
```diff
@@ -392,152 +392,160 @@ pub fn copy_path(
...
+    shell: *Shell,
+    argv: []const []const u8,
+    options: struct {
+        stdin_slice: ?[]const u8 = null,
+        capture_stdout: ?*[]const u8 = null, // Optional out parameter.
+        expect_zero_status: bool = true,
+        output_limit_bytes: usize = 128 * 1024 * 1024,
+        timeout_ns: u64 = 10 * std.time.ns_per_min,
+    },
+) !void {
```
matklad, line 469: This is a sad do-everything-all-at-once kind of function, but at least it is the _only_ non-trivial function which is left, and it is no longer monomorphised for different commands separately!

src/shell.zig
```diff
@@ -392,152 +392,160 @@ pub fn copy_path(
...
+    const argv_formatted = try std.mem.join(shell.gpa, " ", argv);
```
matklad, line 470: One can write a lot more code to format comand line in a nice way, but I think this will be sufficiently unambigious for our use-case!

batiati (approved):

PR #2321, opened 2024-09-16
Add missing `trace.stop` when TableInfo does not contain the key range.

batiati:


matklad:
Could you add `Seed: ./zig/zig build -Drelease fuzz -- lsm_tree 6723122779978547475` to the commit message?

matklad (approved):

PR #2323, opened 2024-09-16
docs: move `tools/systemd` to `docs/operating/linux.md` and remove `tools` entirely

fabioarnold:
One less item in the root directory.


docs/operating/linux.md
```diff
@@ -6,11 +6,129 @@ sidebar_position: 3
...
+
+#### **tigerbeetle-pre-start.sh**
+```bash
+#!/bin/sh
+set -eu
+
+if ! test -e "${TIGERBEETLE_DATA_FILE}"; then
+  /usr/local/bin/tigerbeetle format --cluster="${TIGERBEETLE_CLUSTER_ID}" --replica="${TIGERBEETLE_REPLICA_INDEX}" --replica-count="${TIGERBEETLE_REPLICA_COUNT}" "${TIGERBEETLE_DATA_FILE}"
+fi
+```
```
batiati: We could remove `ExecStartPre=...` from the example and move this snippet to the "Pre-start script" section, making it sound more like a suggestion than the recommended way to create the replica data file.

IMO, this pre-start script is very opinionated.
The cluster operator might want to handle the storage provision and file creation in a different step.

fabioarnold: OK, I moved the script to the "Pre-start script" section.

batiati (approved):
:100:

PR #2324, opened 2024-09-16
devhub: allow filtering by either commit or fuzzer

matklad:
I regularly patch query string only to realize that this doesn't actually work as I expect it to work, so let's fix this


batiati (approved):
:tada:

PR #2326, opened 2024-09-17
Per transfer *_must_not_exceed_* flags

batiati:
Allows the application to enforce "credits must not exceed debits" and "debits must not exceed credits" for accounts not marked with these flags, but at the transfer level.

Unlike Accounts, where the `credits_must_not_exceed_debits` and `debits_must_not_exceed_credits` flags are mutually exclusive, Transfers allow both flags to be used together, as they apply to the corresponding debit or credit account.


batiati:
Closed by #2327

PR #2328, opened 2024-09-18
Docs: Void balance-conditional control transfer

sentientwaffle:
For balance-conditional transfers, the "threshold check" transfer was previously reverted by a transfer for the same amount in the opposite direction.

~Instead, recommend using pending+void, to avoid creating transfers unnecessarily.~ It still creates transfers.

### DIFF
--- a/docs/coding/recipes/balance-conditional-transfers.md
+++ b/docs/coding/recipes/balance-conditional-transfers.md
@@ -46,19 +46,19 @@ We will refer to two amounts:
 
 ### If the Source Account Has a Credit Balance
 
-| Transfer | Debit Account | Credit Account | Amount    | Flags                                                          |
-| -------- | ------------- | -------------- | --------- | -------------------------------------------------------------- |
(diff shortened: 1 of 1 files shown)


docs/coding/recipes/balance-conditional-transfers.md
```diff
@@ -70,9 +70,9 @@ flag would ensure that the first transfer fails. If the first transfer fails, th
 transfers would also fail.
 
 If the first transfer succeeds, it means that the source account did have the threshold balance. In
-this case, the second transfer will atomically transfer the threshold amount back to the source
-account from the control account. Then, the third transfer would execute the desired transfer to the
-ultimate destination account.
+this case, the second transfer cancel the first transfer (returning the threshold amount to the
```
batiati:
```suggestion
this case, the second transfer cancels the first transfer (returning the threshold amount to the
```

batiati (approved):

PR #2333, opened 2024-09-20
vsr: crash cleanly if the data file is too large

### LINKED ISSUE #340: Check storage size vs `storage_size_limit`
In `main.zig`:

        // TODO Panic if the data file's size is larger that args.storage_size_limit.

(Should this be part of `main.zig` (which has access to the `fd`), or `Replica.open()` (which has the `storage`)?)

matklad:
If, at startup, data file size is larger than what our CLI prepared us for, fatal-error cleanly.

Note that this _doesn't_ correctly handle the case where the data file grows when the replica is already up. That'll assert on `free_set.reserve().?` currently which is not ideal.

Closes: https://github.com/tigerbeetle/tigerbeetle/issues/340

### DIFF
--- a/src/lsm/node_pool.zig
+++ b/src/lsm/node_pool.zig
@@ -65,7 +65,7 @@ pub fn NodePoolType(comptime _node_size: u32, comptime _node_alignment: u13) typ
             const node_index = pool.free.findFirstSet() orelse vsr.fatal(
                 .manifest_node_pool_exhausted,
                 "out of memory for manifest, " ++
-                    "restart the replica with larger '--memory-lsm-manifest' argument",
+                    "restart the replica increasing '--memory-lsm-manifest'",
                 .{},
             );
             assert(pool.free.isSet(node_index));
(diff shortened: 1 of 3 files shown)


batiati (approved):
I wonder if we should proactively redefine the limit once the replica starts with a file larger than the default value, printing a warning message as well.

We prefer to fail loudly and give full control of the resources to the operator, however, being able to keep the cluster alive after the first crash without human intervention seems very attractive. :thinking:

cb22:
> We prefer to fail loudly and give full control of the resources to the operator, however, being able to keep the cluster alive after the first crash without human intervention seems very attractive.

I'm less concerned about crashes that are explicit limits (ie, this `.storage_size_exceeds_limit`) and more about ones where the limit is implicit (`.manifest_node_pool_exhausted`, what is enough? How much space does 128MB allow me to reference?)

I think in the real world, we'd want to expose metrics on both the value _and_ the limits, so it'll be very easy for operators to keep track. Additionally, we could probably do some warn-level logging once we start getting near said limits.

matklad:
Oops, forgot about this PR!

matklad:
>I wonder if we should proactively redefine the limit once the replica starts with a file larger than the default value, printing a warning message as well.

I'd maybe not do this --- it is not given that TB is the most important thing running on a machine. For example, if there are _several_ tigerbeetles running on the same host, restarting with larger ram could actually kill the others! Or could make the machine run out of disk space.

PR #2334, opened 2024-09-20
vopr: disable new storage faults before transitioning to liveness mode

chaitanyabhandari:
Resolves the failed VOPR seed (./zig/zig build -Drelease -Dvopr-state-machine=testing vopr -- 12865272390949771184) on main (5378e181bcd5030ceac012ae6cecd428c6b0d3d2). The seed is a false positive, VOPR panics with `@panic("block found in core");` since we disable network & process crash faults while transitioning to liveness mode in the VOPR, but _storage faults are kept enabled_. This PR disables storage faults as well.

This is solved by zeroing storage fault probabilities before transitioning to liveness mode (similar to what we do for network & process faults: https://github.com/tigerbeetle/tigerbeetle/blob/34e85703da17cb9b890c43855453580dfcd74c06/src/vopr.zig#L602-L606)

### DIFF
--- a/src/testing/storage.zig
+++ b/src/testing/storage.zig
@@ -715,6 +715,12 @@ pub const Storage = struct {
             assert(data_block_header.block_type == .data);
         }
     }
+
+    pub fn transition_to_liveness_mode(storage: *Storage) void {
+        storage.options.read_fault_probability = 0;
+        storage.options.write_fault_probability = 0;
+        storage.options.crash_fault_probability = 0;
+    }
 };
 
 pub const Area = union(enum) {
--- a/src/vopr.zig
+++ b/src/vopr.zig
@@ -583,6 +583,15 @@ pub const Simulator = struct {
         simulator.tick_crash();
     }
 
+    /// Executes the following:
+    /// * Pick a quorum of replicas to be fully available (the core)
+    /// * Restart any core replicas that are down at the moment
+    /// * Heal all network partitions between core replicas
+    /// * Disable storage faults on the core replicas
+    /// * For all failures involving non-core replicas, make those failures permanent.
+    ///
+    /// See https://tigerbeetle.com/blog/2023-07-06-simulation-testing-for-liveness for broader
+    /// context.
     pub fn transition_to_liveness_mode(simulator: *Simulator) void {
         simulator.core = random_core(
             simulator.random,
--- a/src/vopr.zig
+++ b/src/vopr.zig
@@ -597,6 +597,7 @@ pub const Simulator = struct {
             if (simulator.cluster.replica_health[replica_index] == .down) {
                 simulator.restart_replica(@intCast(replica_index), fault);
             }
+            simulator.cluster.storages[replica_index].transition_to_liveness_mode();
         }
 
         simulator.cluster.network.transition_to_liveness_mode(simulator.core);


matklad:
CI failure mystifies me completely. It's clearly unrelated to the changes though

src/vopr.zig
```diff
@@ -597,9 +597,14 @@ pub const Simulator = struct {
...
                 simulator.restart_replica(@intCast(replica_index), fault);
             }
+            // Disable storage faults during liveness mode.
+            simulator.cluster.storages[replica_index].transition_to_liveness_mode();
         }
 
+        // Disable network faults during liveness mode.
         simulator.cluster.network.transition_to_liveness_mode(simulator.core);
+
+        // Disable process crash/restart faults during liveness mode.
```
matklad: These comments feel like they don't actually add anything to the code

src/testing/storage.zig
```diff
@@ -715,6 +715,13 @@ pub const Storage = struct {
             assert(data_block_header.block_type == .data);
         }
     }
+
+    pub fn transition_to_liveness_mode(storage: *const Storage) void {
+        const storage_options: *Options = @constCast(&storage.options);
```
matklad: That's not good, that's illegal behavior I think. `const` is not a lint, compiler can rely on stuff not changing in the presence of `const`. Pass `*Storage` here?

src/vopr.zig
```diff
@@ -1101,7 +1107,6 @@ pub const Simulator = struct {
             assert(fault or header_prepare_view_mismatch);
         }
 
-        replica_storage.faulty = true;
```
matklad: I think we can leave this as is? Seems __much__ more reabale to restore `faulty` in this function, no?

src/vopr.zig
```diff
@@ -1028,6 +1033,7 @@ pub const Simulator = struct {
         const fault = recoverable_count >= recoverable_count_min or replica.standby();
         simulator.restart_replica(replica.replica, fault);
         maybe(!fault and replica.status == .recovering_head);
+        simulator.cluster.storages[replica.replica].faulty = true;
```
sentientwaffle: > Whether storage faults should be enabled/disabled post restart should be at the callers' discretion.

:thinking: `restart_replica` modifies `replica_storage.faulty` already (to potentially disable faults to prevent excessive `recovering_head` replicas) -- so I would expect it to restore that, rather than expecting the caller to remember to do so.

src/vopr.zig
chaitanyabhandari: Mhmmmm, I had added them to save a level of indirection. When I first skimmed through this code, I wasn't able to understand the overall picture of liveness mode without peeking into the `network.transition_to_liveness_mode` function. Perhaps better to document this above the function definition instead, then?

matklad: This probably means that we are missing the overall comment on the `transition_to_liveness_mode`, rather the comment on the impl details. `network.transition_to_liveness_mode` is somewhat intricate, as it doesn't _just_ disables faults, but actually "freezes" existing partitions.

Worth linking https://tigerbeetle.com/blog/2023-07-06-simulation-testing-for-liveness at least!

src/vopr.zig
chaitanyabhandari: Mhmmmmm after all of our discussions, I _think_ I've finally understood the semantics of the `faulty` flag. I thought that setting `faulty` as `False` was something that we might want to do for the lifetime of the replica. Therefore, it felt clearer that with `restart_replica(replica, false)`, our intent was to start up a replica with storage faults disabled (and they stay disabled). 

But from the documentation as well, it's now clear we mean for this to be set transiently, only during startup.
https://github.com/tigerbeetle/tigerbeetle/blob/34e85703da17cb9b890c43855453580dfcd74c06/src/testing/storage.zig#L154-L156

I'll revert the commit that changes this behaviour.

src/vopr.zig
chaitanyabhandari: https://github.com/tigerbeetle/tigerbeetle/pull/2334#discussion_r1773458812

src/vopr.zig
chaitanyabhandari: https://github.com/tigerbeetle/tigerbeetle/pull/2334/commits/c8fb7969dd83219cd83cd1366c5e09af543feafc

src/testing/storage.zig
chaitanyabhandari: https://github.com/tigerbeetle/tigerbeetle/pull/2334/commits/7f7e1fda8341c4a494e13ee70d5f072d7e309919

src/vopr.zig
chaitanyabhandari: https://github.com/tigerbeetle/tigerbeetle/pull/2334/commits/7f7e1fda8341c4a494e13ee70d5f072d7e309919

src/vopr.zig
chaitanyabhandari: https://github.com/tigerbeetle/tigerbeetle/pull/2334/commits/7f7e1fda8341c4a494e13ee70d5f072d7e309919

matklad (approved):

Issue #2231, opened 2024-08-22
Tracking Issue: Breaking API Changes


sentientwaffle:
**Subscribe to this issue to be notified of breaking API/behavior changes that are visible to the client.**

Note that the conversation is locked to reduce notification amplification! Feel free to open follow up issues if needed!

sentientwaffle:
## 0.16.0

### Zero-amount transfers

Zero-amount transfers are now permitted.

This is a _breaking API change_ which is gated in the state machine by the _client's_ release.

As of `0.16.0`:
- Clients no longer return `amount_must_not_be_zero` to `create_transfers`.
- Balancing transfers no longer use `amount=0` as a sentinel value to represent "transfer as much as possible". For that purpose, use `AMOUNT_MAX` (exact constant name depends on client implementation) instead.
- Post-pending transfers no longer use `amount=0` as a sentinel value to represent "transfer full pending amount". For that purpose, use `AMOUNT_MAX` instead. In `0.16.0`, sending a post-pending transfer with `amount=0` will successfully post `amount=0`, voiding the remainder (i.e. voiding the whole pending amount).

batiati:
# 0.16.4

### Transient Failures

Clients have the strong guarantee that a `Transfer.id` that has once failed due to a transient error code will never succeed again if retried.

Transient failures occur when semantically valid transfers fail due to reasons that depend exclusively on the database status. For example, when an `account` does not have enough credits to fulfill the `transfer`.

Please refer to the [documentation](https://docs.tigerbeetle.com/reference/requests/create_transfers) for a complete list of all transient failures.

As of `0.16.4`:

- New precedence order for the `create_transfers` and `create_accounts` error codes.  
  The idempotency checks (e.g., all `exists_*` cases) are now validated prior to the semantical checks.

  For example, when submitting a `transfer` with the `id` of an existing one, clients might receive `exists_with_different_<field_name>` instead of `<field_name>_must_not_be_zero`.

  _This is **not** a breaking API change and affects all supported client versions._

- `create_transfers` now returns `exists_with_different_ledger`
  Reflecting the new precedence order change.

  _This is a breaking API change which is gated in the state machine by the client's release._  
  _Clients of previous versions will receive `transfer_must_have_the_same_ledger_as_accounts` instead._

- Added a new result code [`id_already_exists`](https://docs.tigerbeetle.com/reference/requests/create_transfers#id_already_exists).
[...]

PR #2335, opened 2024-09-20
Transient failure

batiati:
This PR introduces the new concept of "transient failure", which are events that fail due to the system's status rather than the request data. The list of transient errors is as follows:

create_accounts: none
create_transfers: `debit_account_not_found`, `credit_account_not_found`, `pending_transfer_not_found`, `exceeds_credits`, `exceeds_debits`, `debit_account_already_closed`, and `credit_account_already_closed`.

With this, we introduce the invariant: **Once a transfer fails with a transient error code, it must always fail with `id_already_failed` when retried with the same ID.**
This ensures that a user will never encounter a "transient error code" twice for the same ID, providing the application with stronger guarantees that the outcome of a transfer will remain consistent across retries.

Additionally, this changes the precedence of the idempotency checks (i.e., all `exists_*` checks) to the beginning, right after `id_must_not_be_zero` and `id_must_not_be_int_max`.

The objectless `id` is stored in the `IdTree` with a zeroed `timestamp`.

The Groove is not aware of "transient failures" but instead it has a configuration that enables "orphaned IDs", which are IDs stored without a corresponding object.

We are using the ObjectCache (with an empty object) to cache such IDs, which can be beneficial for pathological workloads that retry the same failed event in a loop.


matklad:
>Transient failure 

Upon receiving an email, got spooked that our CI went flaky!

src/tigerbeetle.zig
```diff
@@ -286,31 +298,65 @@ pub const CreateTransferResult = enum(u32) {
...
+    pub const Ordered = type: {
+        const values = std.enums.values(CreateTransferResult);
+        var fields: [values.len]std.builtin.Type.EnumField = undefined;
+        for (0..values.len) |index| {
+            const result: CreateTransferResult = @enumFromInt(index);
+            fields[index] = .{
+                .name = @tagName(result),
+                .value = index,
+            };
+        }
```
batiati: It started to become a pain to keep the enum artificially ordered. Now, we declare the values in precedence order and generate an ordered enum at compile time to use in the `EnumSet` as a workaround.

sentientwaffle: Could this instead `sort` the original enum's fields, so the enum values still correspond to the original enum's? Since the precedence ordering is really just "documentation", we could even do that for the original enum, e.g.

```
const CreateTransferResult = EnumSorted(enum { ... })
```

src/tigerbeetle.zig
```diff
@@ -286,31 +298,65 @@ pub const CreateTransferResult = enum(u32) {
     exceeds_credits = 54,
     exceeds_debits = 55,
 
-    // TODO(zig): This enum should be ordered by precedence, but it crashes
+    debit_account_already_closed = 65,
+    credit_account_already_closed = 66,
+
+    // Last item: id_already_failed = 68.
```
sentientwaffle, line 304: :thinking: Why is this listed here? I thought that its position after `exists_*` was correct.

src/state_machine.zig
```diff
@@ -1681,6 +1687,51 @@ pub fn StateMachineType(
...
+                        .pending_transfer_not_found,
+
+                        .exceeds_credits,
+                        .exceeds_debits,
+
+                        .debit_account_already_closed,
+                        .credit_account_already_closed,
+                        => true,
+                        else => false,
+                    };
```
sentientwaffle: What do you think of defining a new `CreateTransferResult` function `fn transient(CreateTransferResult) bool`, which uses an exhaustive switch? Listing all of the results is quite verbose, but it would make it hard to add a new result code without thinking through whether it should be transient.

sentientwaffle (commented):
I like this approach! Handling `exists` as early as possible makes a lot of sense even in ignoring `id_already_failed`!

src/tigerbeetle.zig
batiati: That's a nice idea.

> Since the precedence ordering is really just "documentation"

One side effect of this change that I liked was that the client bindings are now also generated according to precedence. It can be quite helpful when reading the code (in other languages) because it matches the docs.

sentientwaffle: Good point; I completely forgot about how it would impact the client implementation!

src/tigerbeetle.zig
batiati: This comment was intended to be a hint of the last element.
Just to help the next person who will add a new element to follow the numeric order without having to manually browse all values.

src/lsm/groove.zig
```diff
@@ -820,10 +843,25 @@ pub fn GrooveType(
...
                 .positive => |id_tree_value| {
                     if (IdTreeValue.tombstone(id_tree_value)) return;
-                    groove.prefetch_from_memory_by_timestamp(
-                        id_tree_value.timestamp,
-                        .objects_cache,
-                    );
+
+                    if (groove_options.orphaned_ids and
+                        id_tree_value.timestamp == 0)
+                    {
```
sentientwaffle: Could this be:
```suggestion
                    if (id_tree_value.timestamp == 0) {
                        assert(groove_options.orphaned_ids);
```
(And likewise in `lookup_id_callback` too.)

src/tigerbeetle.zig
```diff
@@ -286,31 +300,143 @@ pub const CreateTransferResult = enum(u32) {
...
+            assert(!set.isSet(index));
+            set.set(index);
+        }
+
+        // It's a non-ordered enum, we need to ensure
+        // there are no gaps in the numbering of the values.
+        assert(set.count() == set.capacity());
+    }
+
+    /// TODO(zig): This enum is be ordered by precedence, but it crashes
```
sentientwaffle: Typo "is be"

src/state_machine.zig
```diff
@@ -1855,6 +1890,12 @@ pub fn StateMachineType(
             if (a.id == 0) return .id_must_not_be_zero;
             if (a.id == math.maxInt(u128)) return .id_must_not_be_int_max;
 
+            switch (self.forest.grooves.accounts.get(a.id)) {
+                .found => |e| return create_account_exists(a, e),
+                .orphaned_id => unreachable,
+                .not_found => {},
+            }
```
sentientwaffle: This is equivalent to
```
            if (self.forest.grooves.accounts.get(a.id)) |e| {
                return create_account_exists(a, e);
            }
```

src/state_machine.zig
```diff
@@ -4535,9 +4667,16 @@ test "create_transfers: per-transfer balance invariant" {
         \\ transfer T3 A2 A3  0  T2  _  _  _    0 L1 C1   _   _ _ VOI   _   _ _ _ _ _ _ linked_event_failed
         \\ commit create_transfers
         \\
-        \\ transfer T1 A1 A2 40   _  _  _  _    0 L1 C1 LNK   _ _   _   _   _ _ _ _ _ _ ok
-        \\ transfer T2 A2 A3  1   _  _  _  _    0 L1 C1 LNK PEN _   _ BDR   _ _ _ _ _ _ ok
-        \\ transfer T3 A2 A3  0  T2  _  _  _    0 L1 C1   _   _ _ VOI   _   _ _ _ _ _ _ ok
+        // Ids failed in a linked chain can be reused, buy
```
```suggestion
        // Ids failed in a linked chain can be reused, but
```

docs/reference/requests/create_transfers.md
```diff
@@ -259,11 +381,21 @@ The transfer was not created. [`Transfer.code`](../transfer.md#code) is zero, bu
 The transfer was not created. [`Transfer.debit_account_id`](../transfer.md#debit_account_id) must
 refer to an existing `Account`.
 
+This is a [`transient error`](#id_already_failed).
```
sentientwaffle: I don't think "transient error" should be in backticks, since it isn't an identifier.

src/lsm/scan_builder.zig
```diff
@@ -441,6 +441,14 @@ pub fn ScanType(
                             continue;
                         }
 
+                        if (index == .id) {
+                            // When iterating over `IdTree` it can return a timestamp zero, which
+                            // indicates an orphaned id.
+                            if (Groove.config.orphaned_ids and value.timestamp == 0) continue;
```
sentientwaffle: Could this be:
```suggestion
                            if (value.timestamp == 0) {
                                assert(Groove.config.orphaned_ids);
                                continue;
                            }
```

src/state_machine/workload.zig
```diff
@@ -948,6 +1004,28 @@ pub fn WorkloadType(comptime AccountingStateMachine: type) type {
...
+
+            // Enqueue the `id`s of all transient failures to be retried in the next request.
+            for (results_sparse) |item| {
+                assert(item.result != .ok);
+                assert(item.result != .id_already_failed);
+                if (item.result.transient()) {
+                    self.transient_failures.putNoClobber(
+                        transfers[item.index].id,
+                        {},
+                    ) catch unreachable;
```
sentientwaffle: Alternatively, if the hash map had some (arbitrary, fixed, constant) capacity, then this could insert iff it is not full.

sentientwaffle (commented):
Note that cfo found a forest fuzzer failure too.

src/state_machine.zig
batiati: Although it does not apply to Accounts, I still prefer to explicitly handle `orphaned_id => unreachable` both for clarity and symmetry with the idempotency validation in `create_transfer`.

The helper functions `self.get_transfer(id)` and `self.get_account(id)` should be used in situations where the "orphaned_id" case is not a concern and can even be interpreted as "not found" (e.g. for pending transfers).

src/tigerbeetle.zig
```diff
@@ -400,8 +400,8 @@ pub const CreateTransferResult = enum(u32) {
         assert(set.count() == set.capacity());
     }
 
-    /// TODO(zig): This enum is be ordered by precedence, but it crashes
```
sentientwaffle: s/"This enum"/CreateTransferResult to clarify that we aren't referring to `Ordered`.

src/state_machine.zig
```diff
@@ -2524,7 +2599,18 @@ pub fn StateMachineType(
...
+                .found => |t| t,
+                .not_found, .orphaned_id => null,
+            };
+        }
+
+        fn get_account(self: *const StateMachine, id: u128) ?*const Account {
+            return switch (self.forest.grooves.accounts.get(id)) {
+                .found => |a| a,
+                .not_found => null,
+                .orphaned_id => unreachable,
```
sentientwaffle: Alternatively,
```suggestion
                .found_object => |a| a,
                .found_orphaned_id => unreachable,
                .not_found => null,
```

batiati:
I almost forgot to mention: Imported events are **not** handled as a special case.
After all, it feels like transient errors should take precedence over importing legacy events, as they are part of the financial consistency checks.

What are your thoughts @sentientwaffle?

sentientwaffle:
> After all, it feels like transient errors should take precedence over importing legacy events, as they are part of the financial consistency checks.

I don't think we need to special-case transient errors + import events -- I think that the import script should use a single tigerbeetle client, not a stateless tigerbeetle api server, so transient errors shouldn't happen during imports anyway.

src/state_machine/workload.zig
```diff
@@ -453,11 +475,18 @@ pub fn WorkloadType(comptime AccountingStateMachine: type) type {
...
             client_index: usize,
             transfers: []tb.Transfer,
         ) usize {
             const results = self.auditor.expect_create_transfers(client_index);
+
+            const failed_ids_count = self.build_create_transfers_failed_ids(
+                transfers,
+                results,
+            );
+            if (failed_ids_count > 0) return failed_ids_count;
```
sentientwaffle: :thinking: I'd rather we allowed mixed failed-ids-and-regular-transfer batches. Otherwise I'm worried that every other batch will be a failed transfers batch, which would change the coverage of LSM (i.e. every other batch wouldn't be inserting objects).

Instead of removing the id (`self.transient_failures.popOrNull()`), change `transient_failures` to an arraylist (don't worry about duplicates) and select a random element (but don't remove it). Then when we get back `id_already_failed` we can verify that it is in the list.

sentientwaffle:
When (or before) 0.16.4 is released, can you add a comment to our [breaking changes tracking issue](https://github.com/tigerbeetle/tigerbeetle/issues/2231) explaining transient errors?

sentientwaffle (approved):

PR #2337, opened 2024-09-23
[docs] Use `const` where appropriate in JS snippets

michielbdejong:
My TypeScript compiler complained about the use of `let` in these snippets, saying e.g.:
```
 error  'transferErrors' is never reassigned. Use 'const' instead  prefer-const
```
This PR attempts to fix that in the JS docs

### DIFF
--- a/src/clients/node/README.md
+++ b/src/clients/node/README.md
@@ -94,7 +94,7 @@ See details for account fields in the [Accounts
 reference](https://docs.tigerbeetle.com/reference/account).
 
 ```javascript
-let account = {
+const account = {
   id: 137n,
   debits_pending: 0n,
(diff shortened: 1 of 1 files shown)


batiati:
Hi @michielbdejong!

Thanks for pointing it out!
Since our documentation is generated from the [walkthrough sample](https://github.com/tigerbeetle/tigerbeetle/blob/main/src/clients/node/samples/walkthrough/main.js), the correction would involve more than just changing the markdown file.
I think we should scope each example inside a block and not share declarations between snippets (basically, that's why it's a `let` instead of a `const`).

batiati:
Closed by #2384

PR #2338, opened 2024-09-23
Add note about search key docusaurus.config.js

fabioarnold:
The key is only used for search and would be easily intercepted when using the search.


matklad (approved):
:heart:

PR #2342, opened 2024-09-23
release: validate changelog in tests

matklad:


src/scripts/changelog.zig
```diff
@@ -9,6 +9,8 @@ const ReleaseTriple = @import("../multiversioning.zig").ReleaseTriple;
 
 const log = std.log;
 
+const const_max_bytes = 10 * 1024 * 1024;
```
sentientwaffle: `const_bytes_max`?

sentientwaffle: or `const_size_max`?

sentientwaffle: Wait, why is `const_` the prefix?

matklad: :fearful: not sure what happened here, definitely should be `changelog_bytes_max`

sentientwaffle (approved):

PR #2343, opened 2024-09-23
client: halt VM on completion exception

owickstrom:
* Any double-completion results in Runtime.halt
* Same behavior for async and blocking requests
* Use internal sync/async-specific methods for testing

### DIFF
--- a/src/clients/java/src/main/java/com/tigerbeetle/BlockingRequest.java
+++ b/src/clients/java/src/main/java/com/tigerbeetle/BlockingRequest.java
@@ -115,7 +115,7 @@ final class BlockingRequest<TResponse extends Batch> extends Request<TResponse> 
         synchronized (this) {
 
             if (isDone()) {
-                throw new IllegalStateException("This request has already been completed");
+                throw new IllegalStateException("Request has already been completed");
             } else {
                 this.result = result;
                 this.exception = null;
@@ -132,7 +132,7 @@ final class BlockingRequest<TResponse extends Batch> extends Request<TResponse> 
         synchronized (this) {
 
             if (isDone()) {
-                throw new IllegalStateException("This request has already been completed");
+                throw new IllegalStateException("Request has already been completed");
             } else {
                 this.result = null;
                 this.exception = exception;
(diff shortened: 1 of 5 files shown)


src/clients/java/src/test/java/com/tigerbeetle/AsyncRequestTest.java
```diff
@@ -526,6 +526,22 @@ public void testFailedFutureWithTimeout() throws InterruptedException, TimeoutEx
...
+    public void testFailedFutureCompletedTwice() {
+        var client = getDummyClient();
+        var batch = new IdBatch(1);
+        batch.add();
+
+        var request = AsyncRequest.lookupTransfers(client, batch);
+        var status = PacketStatus.TooMuchData.value;
+
+        // First completion is OK, registering the exception in the CompletableFuture.
+        request.setException(new RequestException(status));
```
batiati: Can we assert the exception is thrown only by the second attempt to set the result?

```suggestion
        try {
          // First completion is OK, registering the exception in the CompletableFuture.
          request.setException(new RequestException(status));
        } catch {
          // No exception is expected in the first call.
          assert false;
        }
```

src/clients/java/src/main/java/com/tigerbeetle/BlockingRequest.java
```diff
@@ -115,13 +115,8 @@ protected void setResult(final TResponse result) {
         synchronized (this) {
 
             if (isDone()) {
-
-                this.result = null;
-                this.exception = new AssertionError(this.exception,
-                        "This request has already been completed");
-
+                throw new IllegalStateException("This request has already been completed");
```
batiati: Can we standardize the messages as "Request already completed" like the async case?

src/clients/java/src/main/java/com/tigerbeetle/Request.java
```diff
@@ -177,10 +177,17 @@ void endRequest(final byte receivedOperation, final byte status) {
...
-            setException(exception);
+        try {
+            if (exception == null) {
+                setResult((TResponse) result);
+            } else {
+                setException(exception);
+            }
+        } catch (Throwable any) {
+            System.err.println("Completion of request failed!\n"
+                    + "This is a bug in TigerBeetle. Please report it at https://github.com/tigerbeetle/tigerbeetle.\n"
```
batiati, line 188: :heart:

batiati (approved):

PR #2344, opened 2024-09-23
node+tidy: tidy up file locations and tidy.zig itself

cb22:
`node.zig` can be moved to `src/clients/node` (but getting `tb_client_exports.zig` out of `src/` is a bit more involved). Additionally, prune the tidy filename exception list, and while we're there, validate all exceptions are actually exceptions.

### DIFF
--- a/src/tidy.zig
+++ b/src/tidy.zig
@@ -600,17 +600,15 @@ test "tidy extensions" {
     });
 
     const exceptions = std.StaticStringMap(void).initComptime(.{
-        .{".editorconfig"},                       .{".gitignore"},
-        .{".nojekyll"},                           .{"CNAME"},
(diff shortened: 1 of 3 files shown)


src/tidy.zig
```diff
@@ -600,24 +600,37 @@ test "tidy extensions" {
...
     });
 
     const allocator = std.testing.allocator;
     const shell = try Shell.create(allocator);
     defer shell.destroy();
 
     const paths = try list_file_paths(shell);
+
+    for (exceptions.keys()) |exception| {
+        for (paths) |path| {
```
cb22, line 621: The `O(mn)` can serve as motivation to keep our exception list below a few million :smile:.

matklad (approved):
:heart:

PR #2345, opened 2024-09-23
vsr: handle multiple torn writes at the end of the WAL

chaitanyabhandari:
Currently, our logic for handling torn prepares (prepares that were being made durable before a replica crash, but weren't written out completely) only handles _one_ torn prepare: https://github.com/tigerbeetle/tigerbeetle/blob/577b638e52725f5ff128822e4c75ccc4fb2ca5f2/src/vsr/journal.zig#L1369

This is incorrect because there could be _multiple_ prepares concurrently being written before a replica crash:
https://github.com/tigerbeetle/tigerbeetle/blob/577b638e52725f5ff128822e4c75ccc4fb2ca5f2/src/config.zig#L114

This PR correctly detects and truncates multiple torn writes in the WAL, improving _availability_! Before this change, a replica with multiple torn writes at the end of the WAL would transition into .recovering_head status on startup as those writes would be conservatively treated as corruptions.

### DIFF
--- a/src/vsr/journal.zig
+++ b/src/vsr/journal.zig
@@ -1296,10 +1296,9 @@ pub fn JournalType(comptime Replica: type, comptime Storage: type) type {
             assert(journal.headers.len == cases.len);
 
             // Refine cases @B and @C: Repair (truncate) a prepare if it was torn during a crash.
-            if (journal.recover_torn_prepare(&cases)) |torn_slot| {
+            for (journal.torn_prepares(&cases)) |torn_slot| {
                 assert(cases[torn_slot.index].decision(replica.solo()) == .vsr);
                 cases[torn_slot.index] = &case_cut_torn;
-
                 log.warn("{}: recover_slots: torn prepare in slot={}", .{
                     journal.replica,
                     torn_slot.index,
@@ -1315,7 +1314,7 @@ pub fn JournalType(comptime Replica: type, comptime Storage: type) type {
             // - after prepare_op_max is computed,
             // - after the case decisions are made (to avoid @H:vsr arising from an
             //   artificially reserved prepare),
-            // - after recover_torn_prepare(), which computes its own max ops.
+            // - after torn_prepares(), which computes its own max ops.
             // - before we repair the 'fix' cases.
             //
             // (These headers can originate if we join a view, write some prepares from the new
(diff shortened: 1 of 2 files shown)


src/vsr/journal.zig
```diff
@@ -1356,93 +1355,141 @@ pub fn JournalType(comptime Replica: type, comptime Storage: type) type {
...
                     }
                 }
             }
-
-            // The prepare is torn.
-            assert(!journal.prepare_inhabited[torn_slot.index]);
-            assert(!torn_prepare_untrusted.valid_checksum());
-            assert(cases[torn_slot.index].decision(replica.solo()) == .vsr);
-            return torn_slot;
+            return torn_slots[0..torn_slot_count];
```
sentientwaffle: Since this array is on our stack, I think returning a slice of it is UB.
Maybe use a `BoundedArray` instead?

src/vsr/journal.zig
```diff
@@ -1356,93 +1355,141 @@ pub fn JournalType(comptime Replica: type, comptime Storage: type) type {
...
+                    } else {
+                        break :prepare_max op_prepare_max_slot;
+                    }
+                },
             };
-
-            // We must be certain that the torn prepare really was being appended to the WAL.
-            // Return null if any faults do not lie between the checkpoint and the torn prepare,
-            // such as:
+            var torn_slot_count: u8 = 0;
```
sentientwaffle:
```suggestion
            var torn_slot_count: u32 = 0;
```

`journal_iops_write_max < maxInt(u8)` and `journal_slot_count < maxInt(u8)`, but that isn't an invariant we depend on elsewhere afaik.

chaitanyabhandari:
I was using https://github.com/tigerbeetle/tigerbeetle/pull/2350 to test the new torn prepares logic against only torn prepare faults. We have disabled all read & write storage faults on that branch, and enabled only [_non .wal_header_ crash storage faults](https://github.com/tigerbeetle/tigerbeetle/pull/2350/files#diff-531b6ad03ce7fc1d083c746e3bea5d7403c7a65b4ad70ce01bc07e3138da0595R225-R228), and we're [asserting](https://github.com/tigerbeetle/tigerbeetle/pull/2350/files#diff-ff61239ba80a7cf2991f89b9394ed1d1eeb75198961805f094cf6c11dbd9576aR1099) that the replica doesn't start up into .recovering_head. We're skipping faults to .wal_headers since faults in the redundant headers *along with* faults in the prepares *can* cause a replica to restart into .recovering_head. 

Despite this, that test is still imperfect because we can get into recovering_head due to a fault in a prepare *along with* a read to a reserved redundant header. This could happen if:
 1. Slot was found corrupt on a previous startup, which set the header
    to reserved in memory.
 2. Replica crashes *before* the corrupt slot was repaired, but
    *after* the reserved header was written to disk with a write to
    a nearby header (there are multiple headers in a single sector)

This is the only root cause that I've observed amongst all failures on that branch. I'm also thinking about refining the  [assert condition](https://github.com/tigerbeetle/tigerbeetle/pull/2350/files#diff-ff61239ba80a7cf2991f89b9394ed1d1eeb75198961805f094cf6c11dbd9576aR1099) in a hacky way (since the VOPR code there isn't meant for merging anyway), so we don't see these false positives.

src/vsr/journal.zig
```diff
@@ -1356,93 +1355,141 @@ pub fn JournalType(comptime Replica: type, comptime Storage: type) type {
...
+
+                        // Header is valid and from a previous wrap.
+                        assert(header_redundant_ok != null);
+                        assert(header_redundant_ok.?.op < op_max);
+
+                        assert(!header_prepare_untrusted.valid_checksum());
+                        assert(!journal.prepare_inhabited[index]);
+
+                        torn_slots[torn_slot_count] = slot;
+                        torn_slot_count += 1;
```
sentientwaffle: I don't see anything preventing more than `journal_iops_write_max` "torn" writes from occurring, which would cause this to panic. (If we do get more tears than we expect, maybe we should `log.warn` + then not truncate any of them? Not sure.)

src/vsr/journal.zig
```diff
@@ -1356,93 +1355,141 @@ pub fn JournalType(comptime Replica: type, comptime Storage: type) type {
...
+                    // We only need to check the slot for op_prepare_max. Checked separately as
+                    // SlotRange.contains doesn't handle empty ranges.
+                    if (op_max == op_prepare_max - 1 and index != op_prepare_max_slot.index) {
+                        assert(op_max_to_op_prepare_max.head.index == op_prepare_max_slot.index);
+                        assert(op_max_to_op_prepare_max.tail.index == op_prepare_max_slot.index);
+                        continue;
+                    }
+
+                    if ((op_max == op_prepare_max - 1 and index == op_prepare_max_slot.index) or
+                        op_max_to_op_prepare_max.contains(slot))
```
sentientwaffle: Extract `op_max == op_prepare_max - 1` to a `range_empty` variable, for clarity. (Also, should it use `op_max_to_op_prepare_max.tail` instead of `op_prepare_max - 1`?)

src/vsr/journal.zig
```diff
@@ -1356,93 +1355,141 @@ pub fn JournalType(comptime Replica: type, comptime Storage: type) type {
...
+                        // 3. Misdirected read for a prepare header.
+                        if (header_prepare_untrusted.valid_checksum()) {
+                            // This is a faulty slot and header is valid, prepare must be corrupt.
+                            assert(header_prepare_ok == null);
+                            return torn_slots[0..0];
+                        }
+
+                        // Header is valid and from a previous wrap.
+                        assert(header_redundant_ok != null);
+                        assert(header_redundant_ok.?.op < op_max);
```
sentientwaffle, line 1478: > Header is valid and from a previous wrap.

Can we also assert that the op is less than `op_checkpoint`?

sentientwaffle (commented):
WIP review

src/vsr/journal.zig
chaitanyabhandari: Oops, this was a bad miss on my part! Thank you for pointing it out.

src/vsr/journal.zig
chaitanyabhandari: Mhmmmmm, great catch! To clarify, you're referring to cases where > `journal_iops_write_max` old prepares after our head get corrupted, right?

If so, we _could_ actually correctly truncate all of them; instead of returning a bounded array we could keep a bitset of size `journal_slot_count` which we could use to directly set `cases[torn_slot.index] = &case_cut_torn;` within `recover_torn_prepares` itself! Thoughts?

sentientwaffle: My thinking was that if there are that many corruptions, then this might be something more serious than torn writes, and we should err on the side of caution. Marking the slot as faulty is always safe, whereas truncating incorrectly could lead to an incorrect nack.

chaitanyabhandari: Ah, that makes a lot of sense; we _should_ bound it then. I will take your original suggestion: bounding torn prepares to `journal_iops_write_max `, and if we see > `journal_iops_write_max `, we err on the side of caution and don't truncate anything.

src/vsr/journal.zig
chaitanyabhandari: Took care of this in https://github.com/tigerbeetle/tigerbeetle/pull/2345/commits/8c0600bcefe1bab79a1df21d142fbe2c3d4a10d3.

src/vsr/journal.zig
chaitanyabhandari: Returning bounded array obviates the need for this variable.

src/vsr/journal.zig
chaitanyabhandari: Took care of this in https://github.com/tigerbeetle/tigerbeetle/pull/2345/commits/8c0600bcefe1bab79a1df21d142fbe2c3d4a10d3.

src/vsr/journal.zig
chaitanyabhandari: Took care of this in https://github.com/tigerbeetle/tigerbeetle/pull/2345/commits/8c0600bcefe1bab79a1df21d142fbe2c3d4a10d3.

src/vsr/journal.zig
chaitanyabhandari: op wouldn't be strictly less than `op_checkpoint`; our range's tail is op_prepare_max's slot, which is the same as the current op_checkpoint's slot!

Added a <= assert in https://github.com/tigerbeetle/tigerbeetle/pull/2345/commits/8c0600bcefe1bab79a1df21d142fbe2c3d4a10d3.

src/vsr/journal.zig
```diff
@@ -1356,93 +1355,154 @@ pub fn JournalType(comptime Replica: type, comptime Storage: type) type {
...
-            // The prepare is at least corrupt, possibly torn, but not valid and simply misdirected.
+            // Nothing to truncate - head op is not certain as it must be >= op_checkpoint.
+            if (op_max < replica.op_checkpoint()) return .{};
 
-            const header_untrusted = &journal.headers_redundant[torn_slot.index];
-            const header = header_ok(replica.cluster, torn_slot, header_untrusted) orelse
-                return null;
-            // The redundant header is valid, also for the correct cluster and not misdirected.
+            // Nothing to truncate - replica could not have prepared anything beyond prepare_max.
+            if (op_max == replica.op_prepare_max()) return .{};
```
sentientwaffle, line 1394: :thinking: Can you expand on why this condition is necessary? Pipelined prepares can come out of order.

e.g. just b/c we have `op_prepare_max`, doesn't mean that `op_prepare_max - 1` was prepared.

src/vsr/journal.zig
```diff
@@ -1356,93 +1355,154 @@ pub fn JournalType(comptime Replica: type, comptime Storage: type) type {
...
+                        op_max_to_op_prepare_max.tail.index;
+                    if (range_empty) {
+                        assert(op_max_to_op_prepare_max.head.index == op_prepare_max_slot.index);
+                        assert(op_max_to_op_prepare_max.tail.index == op_prepare_max_slot.index);
+                        if (index != op_prepare_max_slot.index) continue;
+                    }
+
+                    if ((range_empty and index == op_prepare_max_slot.index) or
+                        op_max_to_op_prepare_max.contains(slot))
+                    {
```
sentientwaffle: Could this be simplified to:
```suggestion
                    const range_empty = op_max_to_op_prepare_max.head.index ==
                        op_max_to_op_prepare_max.tail.index;
                    if ((range_empty and index == op_prepare_max_slot.index) or
                        (!range_empty and op_max_to_op_prepare_max.contains(slot)))
                    {
```

src/vsr/journal.zig
```diff
@@ -1356,93 +1355,154 @@ pub fn JournalType(comptime Replica: type, comptime Storage: type) type {
...
+                        //       to reserved in memory.
+                        //    ii. Replica crashes *before* the corrupt slot was repaired, but
+                        //       *after* the reserved header was written to disk with a write to
+                        //       a nearby header (there are multiple headers in a single sector)
+                        if (header_redundant_ok.?.operation == .reserved) return .{};
+
+                        // 3. Misdirected read for a prepare header.
+                        if (header_prepare_untrusted.valid_checksum()) {
+                            // This is a faulty slot and header is valid, prepare must be corrupt.
+                            assert(header_prepare_ok == null);
```
sentientwaffle: This can panic even if `header_prepare_untrusted` has a valid checksum -- e.g. if it is a valid prepare from the wrong cluster and/or slot (due to a misdirected write).

src/vsr/journal.zig
```diff
@@ -1296,10 +1296,9 @@ pub fn JournalType(comptime Replica: type, comptime Storage: type) type {
             assert(journal.headers.len == cases.len);
 
             // Refine cases @B and @C: Repair (truncate) a prepare if it was torn during a crash.
-            if (journal.recover_torn_prepare(&cases)) |torn_slot| {
+            for (journal.torn_prepares(&cases).const_slice()) |torn_slot| {
```
sentientwaffle: I genuinely don't know whether this code is safe or not, so just to be careful let's extract the array:
```suggestion
            const torn_slots = journal.torn_prepares(&cases);
            for (torn_slots.const_slice()) |torn_slot| {
```

src/vsr/journal.zig
chaitanyabhandari: My reasoning here was: our objective here is to truncate torn prepares that were being appended to the _end_ of the WAL, if we are able to discern a certain _end/head_. In this case, since we've found `op_prepare_max` in our journal, it is guaranteed to be certain because this replica couldn't have prepared any op greater than that. Also, since the replica couldn't have been appending any op > op_prepare_max to the end of the WAL, there is nothing to truncate. 

As for the ops _before_ `op_prepare_max`, if they were being prepared before we crashed, they would be categorized as `@B/@C` and fixed using our regular repair.

Thoughts?

chaitanyabhandari: Mhmmmm, but I get what you're thinking now: we _could_ be repairing old prepares in slots past `op_prepare_max` and could crash while writing those. If we don't truncate those here, that would needlessly land us in `recovering_head`... This means torn prepares don't necessarily have to be "potentially greater" prepares, they could be _any_ prepares after our head. 🤔

sentientwaffle: > we could be repairing old prepares in slots past op_prepare_max and could crash while writing those.

Actually that's not quite what I was thinking.

Suppose two prepares are being prepared concurrently:
- op=X, which is 1MiB, and
- op=X+1, which is 1KiB.

Op=X+1 finishes successfully, but before op=X is done writing we crash, tearing the prepare write.

chaitanyabhandari: Mhmmmm, but in this case couldn't we just let op=X be fixed using our regular repair protocol (since our head op is going to be set to op=X+1 anyway, if there are no other potential heads in the journal of course)?

sentientwaffle: Ahh, I see, recover-torn-prepares job is just to prevent `recovering_head`! That makes sense. Can you update `torn_prepares`'s comment to explain that?

chaitanyabhandari: Yes, let me update the comment! 

Also, wanted to get your input about https://github.com/tigerbeetle/tigerbeetle/pull/2345#discussion_r1777409243 as well, which _seems_ like a valid concern to me. If we are repairing some old ops to the right of op_prepare_max and they are torn, we it seems like that could currently land us in .recovering_head since we bail out...

I would have to think about this in terms of a concrete example though, with op_repair_min and op_repair_max, to understand this scenario more clearly. 😄

sentientwaffle: > we could be repairing old prepares in slots past op_prepare_max and could crash while writing those. If we don't truncate those here, that would needlessly land us in recovering_head... This means torn prepares don't necessarily have to be "potentially greater" prepares, they could be any prepares after our head.

Checking my understanding: Would this only impact slots that are also to the left of `op_checkpoint`?

src/vsr/journal.zig
chaitanyabhandari: Mhmmmm great catch; we indeed can't `assert(header_prepare_ok == null);` as there are many cases where a fault is categorized as `vsr` but both header & prepare are valid.

src/vsr/journal.zig
chaitanyabhandari: Thought about this a bit more, seems like the case where our head op turns out to `op_prepare_max` is a special where we need to scan the recovery cases of the _entire journal_, since we could be repairing any op from the current wrap.

For example, if `op_checkpoint=19`, `op_prepare_max=51` (which is in the same slot as op_checkpoint=19), and our `op_max=51`, then we _could_ have been repairing anything starting between `[20, 50]` (op_repair_min and op_repair_max respectively). Therefore, to be able to avoid .recovering_head in this scenario, we'd have to scan all of the cases (except for op_max=op_prepare_max). 

What are your thoughts @sentientwaffle?

sentientwaffle: Isn't it the other way around? If `op_head == op_prepare_max`, then we _definitely_ know the head op, since we couldn't possibly have prepared anything beyond `op_prepare_max`. (Since `op_prepare_max` is derived from our superblock checkpoint).

chaitanyabhandari: But could we still be repairing old ops between [op_checkpoint, op_prepare_max]? 🤔

sentientwaffle: Right, but how would that lead to `recovering_head`?

chaitanyabhandari: Mhmmmm, I was thinking that if the prepares being repaired are torn, they'd be marked as faulty and our `op_head_certain` logic would cause us to go into recovering_head... https://github.com/tigerbeetle/tigerbeetle/blob/f95789c065fd37897b65472b9cf57836771f710f/src/vsr/replica.zig#L5893-L5906

chaitanyabhandari: Oh wait, `slot_known_range` actually turns out to be an empty range in that case, would `slot_known_range.contains` actually panic here if we have faulty slots or am I missing something obvious? 😮

chaitanyabhandari: Oh no, scratch that we have the following logic that would prevent that.
https://github.com/tigerbeetle/tigerbeetle/blob/f95789c065fd37897b65472b9cf57836771f710f/src/vsr/replica.zig#L5859-L5869

chaitanyabhandari: But, this logic _would_ lead us into recovering_head in case there are any faults, right? 🤔

sentientwaffle: Maybe `op_head_certain` should take `op_prepare_max` into account? e.g.:
- Assert that `op_head <= op_prepare_max`, and
- `slot_known_range = slot_op_checkpoint .. slot_op_prepare_max`

chaitanyabhandari: Mhmmmm you're right, `op_head_certain` could be a good place to handle this! If our head op is op_prepare_max, then we could bail out early since it's guaranteed to be certain (faults in the journal can't affect the certainty in this case)...

src/vsr/journal.zig
chaitanyabhandari: Took care of this in https://github.com/tigerbeetle/tigerbeetle/pull/2345/commits/4a8eb61f8001cc24f89b665c5af17b7011e9eb74.

src/vsr/journal.zig
chaitanyabhandari: Took care of this in https://github.com/tigerbeetle/tigerbeetle/commit/4a8eb61f8001cc24f89b665c5af17b7011e9eb74.

src/vsr/journal.zig
chaitanyabhandari: Took care of this in https://github.com/tigerbeetle/tigerbeetle/commit/4a8eb61f8001cc24f89b665c5af17b7011e9eb74.

src/vsr/journal.zig
chaitanyabhandari: Took care of this in https://github.com/tigerbeetle/tigerbeetle/pull/2345/commits/95cc35ab166ff11f8a3061a1b7585ffe7522c483. It has been passing continuously running VOPR runs that I've started locally, but I definitely want to let this soak on CFO for a while...

src/vsr/replica.zig
```diff
@@ -722,6 +722,10 @@ pub fn ReplicaType(
                 self.op -| constants.pipeline_prepare_queue_max,
             );
 
+            // Guaranteed since our durable view_headers always contain an op from the current
+            // checkpoint (see `commit_checkpoint_superblock`).
+            assert(self.op >= self.op_checkpoint());
```
sentientwaffle: :tada: Excellent that this assert can be lifted up out of the branch!

If we move it up a few more lines, it can be paired with the upper-bound assertion, i.e.:

```zig
// Our durable view_headers always contain an op from the current
// checkpoint (see `commit_checkpoint_superblock`).
assert(op_head.? >= self.op_checkpoint());
assert(op_head.? <= self.op_prepare_max());
```

src/vsr/journal.zig
chaitanyabhandari: Okay, happy about this commit now, ~2.1k passed seeds on my local VOPR run and doing well on the CFO as well!

src/vsr/replica.zig
```diff
@@ -5891,13 +5879,13 @@ pub fn ReplicaType(
...
-                .head = slot_op_checkpoint,
-                .tail = slot_op_head,
+                .head = self.journal.slot_for_op(self.op + 1),
+                .tail = self.journal.slot_for_op(self.op_prepare_max()),
             };
 
             var iterator = self.journal.faulty.bits.iterator(.{ .kind = .set });
             while (iterator.next()) |slot| {
-                if (!slot_known_range.contains(.{ .index = slot })) {
+                if (slot_known_range.contains(.{ .index = slot })) {
```
sentientwaffle: What guards against the "empty range panic" issue when `op_head+1 == op_prepare_max`?

src/vsr/replica.zig
```diff
@@ -5891,13 +5879,13 @@ pub fn ReplicaType(
             }
```
sentientwaffle, line 5880: :thinking: Nit, I think the above condition could be clearer if we were checking `slot_op_prepare_max` instead of `slot_op_checkpoint` -- even if they are the same, the former is what we actually care about.

src/vsr/replica.zig
chaitanyabhandari: Oops, forgot about that one here! Great catch, thanks a lot for pointing it out. :D

chaitanyabhandari:
Agree with https://github.com/tigerbeetle/tigerbeetle/pull/2345#discussion_r1777722649! Addressed this and other commends in https://github.com/tigerbeetle/tigerbeetle/pull/2345/commits/2832f8e94a5ced5d14127e05e6d3ba84d5258f8d.

src/vsr/replica.zig
```diff
@@ -5882,23 +5870,28 @@ pub fn ReplicaType(
             // If faulty, this slot may hold either:
             // - op=op_checkpoint, or
             // - op=op_prepare_max
-            if (self.journal.faulty.bit(slot_op_checkpoint)) {
+            if (self.journal.faulty.bit(slot_prepare_max)) {
                 log.warn("{}: op_head_certain: faulty checkpoint slot={}", .{
```
```suggestion
                log.warn("{}: op_head_certain: faulty prepare_max slot={}", .{
```

src/vsr/replica.zig
```diff
@@ -757,7 +761,6 @@ pub fn ReplicaType(
                 } else {
                     // Even if op_head_certain() returns false, a DVC always has a certain head op.
```
sentientwaffle: I think you can remove this comment; the above "Guaranteed since our durable..." comment refers to the same thing.

sentientwaffle (commented):
Just a couple last nits, but otherwise LGTM!

src/vsr/replica.zig
chaitanyabhandari: Mhmmm, isn't this comment different? I always thought it meant that we don't need to call `op_head_certain` for the case where `log_view < view` because we rely _only_ on DVC (without looking at the journal) to infer head in this case. 🤔

chaitanyabhandari: Actually, let me just update that in the comment for clarity!

sentientwaffle: Ah right, thank you!

chaitanyabhandari:
Final set of comments addressed in https://github.com/tigerbeetle/tigerbeetle/pull/2345/commits/26ff2ce2a5e86ecf8315bf40356ed4dfc11bba79!

sentientwaffle (approved):

PR #2346, opened 2024-09-24
ci: trigger systests on main

owickstrom:
Trigger test on each push to `main`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -189,3 +189,35 @@ jobs:
         run: exit 0
       - if: ${{  (contains(needs.*.result, 'failure') || contains(needs.*.result, 'cancelled')) }}
         run: exit 1
+
+  antithesis:
+    if: github.ref == 'refs/heads/main'
+    runs-on: ubuntu-latest
+    environment: antithesis
+
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 2147483647
+      - name: Install dependencies
+        run: sudo apt install -y jq curl
+      - uses: actions/setup-java@v4
+        with:
+          distribution: "temurin"
+          java-version: "21"
+      - name: Login to Container Registry
+        run: docker login -u "$ANTITHESIS_DOCKER_USER" -p "$ANTITHESIS_DOCKER_PASS" us-central1-docker.pkg.dev
+        env:
+          ANTITHESIS_DOCKER_USER: ${{ secrets.ANTITHESIS_DOCKER_USER }} 
+          ANTITHESIS_DOCKER_PASS: ${{ secrets.ANTITHESIS_DOCKER_PASS }} 
+      - name: Download Zig
+        run: ./zig/download.sh
+      - name: Build and push
+        run: |
+          ./zig/zig build scripts -- antithesis \
+            --tag='antithesis-latest' \
+            --push --trigger-test \
+            --antithesis-user="$ANTITHESIS_USER" --antithesis-password="$ANTITHESIS_PASS"
+        env:
+          ANTITHESIS_USER: ${{ secrets.ANTITHESIS_USER }}
+          ANTITHESIS_PASS: ${{ secrets.ANTITHESIS_PASS }}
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -189,3 +189,40 @@ jobs:
...
+      #       '{params: {"custom.duration":"0.5", "antithesis.images": $images, "antithesis.config_image": $config_image, "antithesis.description": $git_commit, "antithesis.report.recipients":"oskar@tigerbeetle.com"}}')
+      #
+      #     curl --fail \
+      #       -u "$ANTITHESIS_USER:ANTITHESIS_PASS" \
+      #       -X POST https://tigerbeetle.antithesis.com/api/v1/launch_experiment/tigerbeetle \
+      #       -d "$body"
+      #   env:
+      #     ANTITHESIS_USER: ${{ secrets.ANTITHESIS_USER }}
+      #     ANTITHESIS_PASS: ${{ secrets.ANTITHESIS_PASS }}
+      #
```
matklad: Up to you, but consider, instead of introducing a separate CI job here, to chain-load this onto the existing `devhub` one. 

The principle here is 

https://matklad.github.io/2023/12/31/O(1)-build-file.html

Rather than adding new build hooks, add stuff to an existing hook. For antithesis, the proper hook seems to be "new code is pushed to `main`", which is our `devhub` job (maybe its worth to rename `devhub` job to "run on main" or something)

owickstrom: Ah, yeah I was thinking about it. But I got stuck on the idea of a "debounced" trigger, which seemed to require a separate job. If we're OK with it running on every merge (possibly consuming our budget faster), I'll put it in the devhub script (perhaps renamed as you say).

src/scripts/antithesis.zig
```diff
@@ -159,6 +180,68 @@ fn push_image(shell: *Shell, image: Image, tag: []const u8) !void {
...
+        .params = .{
+            .@"custom.duration" = "0.5",
+            .@"antithesis.images" = try shell.fmt("workload:{s};replica:{s}", .{ tag, tag }),
+            .@"antithesis.config_image" = try shell.fmt("config:{s}", .{tag}),
+            .@"antithesis.description" = commit,
+            .@"antithesis.report.recipients" = "oskar@tigerbeetle.com",
+        },
+    }, .{}, body.writer());
+
+    var client = std.http.Client{ .allocator = shell.gpa };
```
matklad, line 208: wow, we have a usable http client in stdlib! I don't think we've used it before, I am wondering if we should replace some of our shelling out with this stuff...

src/scripts/antithesis.zig
```diff
@@ -159,6 +180,68 @@ fn push_image(shell: *Shell, image: Image, tag: []const u8) !void {
...
+    defer client.deinit();
+
+    const user_pass = try shell.fmt("{s}:{s}", .{ antithesis_user, antithesis_password });
+    const b64 = std.base64.url_safe.Encoder;
+
+    const auth = try shell.gpa.alloc(u8, b64.calcSize(user_pass.len));
```
matklad: I'd rather stack-allocate a kilobyte and use that. If someone tries to use War and Peace as their password, we should fail _before_ we get to doing an HTTP request

src/scripts/antithesis.zig
```diff
@@ -159,6 +180,68 @@ fn push_image(shell: *Shell, image: Image, tag: []const u8) !void {
...
+        .payload = body.items,
+        .response_storage = .{ .dynamic = &response },
+    });
+
+    if (result.status != std.http.Status.ok) {
+        std.debug.print(
+            "Trigger test failed (code {}): {s}\n",
+            .{ result.status, response.items },
+        );
+        unreachable;
```
matklad: `log.err` rather than `debug.print`

matklad: and maybe `return error.BadStatus` rather than unreacheable?

src/scripts/antithesis.zig
owickstrom: Yeah I was surprised it was in there! First time I used it.

src/scripts/antithesis.zig
matklad: Actually, the same goes for the json above I think? Would be good to impose static limits on the amount of data we send over the wire

matklad (dismissed):

src/scripts/antithesis.zig
owickstrom: Done :+1:

matklad (dismissed):

matklad (approved):

PR #2347, opened 2024-09-24
vsr: fix typo

matklad:


cb22 (approved):
I think I liked this idea so much during the original review that my brain skipped the typo! :rofl:

PR #2349, opened 2024-09-24
ci: limit len of http response being read

owickstrom:
Let's not read an unbounded response over HTTP just for an error message.


matklad (approved):
🤌

PR #2353, opened 2024-09-25
lsm: simplify compaciton dependency injection

matklad:
Now that we have pointer stability, we no longer need to inject forest just-in-time during compaction.


batiati (approved):
❤️

matklad:
:heart: *you* for giving us pointer stability !

PR #2356, opened 2024-09-27
test: full-system integration test with Vortex

owickstrom:
Introducing _Vortex_, our new full-system integration test. This features a supervisor, which:

* starts a cluster of 3 replicas, running them as subprocesses
* starts the Vortex _workload_ as a separate process, which communicates with the cluster through a _driver_
* injects basic faults

The workload generates and submits commands and queries that check the consistency of results against a simple model.

Right now there's only a Zig (tb_client) driver available. The idea is that we can write drivers for each language client we support. Third-party clients might also be able to use Vortex to test their correctness.

Vortex can be run with a configurable duration. Our aim is to run it continuously on a VM later on. For now, it's integrated into CI as a smoke test. This can be run locally (Linux-only for now) with:

```shell
zig build test:integration -- "systest smoke"
```

You can also run it directly:

```shell
zig build vortex -Drelease && unshare -nfr zig-out/bin/vortex supervisor --tigerbeetle-executable=./tigerbeetle --test-duration-minutes=1
```

It uses [unshare(1)](https://man7.org/linux/man-pages/man1/unshare.1.html) to create a new namespace, with isolated network emulation and process tree. This enables us to run the supervisor without affecting the systems network overall, and we can run multiple instances of Vortex using the same (isolated) ports for replicas.

If you have permissions troubles with Ubuntu, see: https://github.com/YoYoGames/GameMaker-Bugs/issues/6015#issuecomment-2135552784

Future ideas for this work:

* Use unshare with a syscall in the supervisor, rather than through the shell
* More advanced fault injection (full partitions, packet reordering, filesystem corruption and read/write faults, clocks, etc)
* Upgrades (clients and replicas)
* Liveness checks
* Multiple drivers in a single test


src/testing/systest/supervisor/systest.zig
```diff
@@ -0,0 +1,248 @@
...
+
+        var process = try LoggedProcess.init(allocator, name, argv, .{});
+        errdefer process.deinit();
+
+        replicas[i] = .{ .name = name, .port = replica_ports[i], .process = process };
+        try process.start();
+    }
+
+    // let the replicas start
+    std.time.sleep(5 * std.time.ns_per_s);
```
owickstrom: Maybe we shouldn't?

owickstrom: Doesn't do anything useful, I'm removing it.

.github/workflows/ci.yml
```diff
@@ -48,6 +48,11 @@ jobs:
       - uses: actions/checkout@v4
         with:
           fetch-depth: 2147483647
+      - uses: actions/setup-java@v4
+        with:
+          distribution: "temurin"
+          java-version: "21"
+
```
matklad: Needing java to test our database is a rather significant regression. This is an intentional regression, so that we can iterate on systest infra quickly _and_ also get some coverage for the java client for free, but it's a regression non the less.

owickstrom it's on you to keep the tabs here and make sure we fix this eventually!

build.zig
```diff
@@ -641,8 +641,10 @@ fn build_test_integration(b: *std.Build, step_test_integration: *std.Build.Step,
         .optimize = options.mode,
         .filters = b.args orelse &.{},
     });
+    integration_tests.root_module.addOptions("vsr_options", vsr_options);
     integration_tests.root_module.addOptions("test_options", integration_tests_options);
     const run_integration_tests = b.addRunArtifact(integration_tests);
+    run_integration_tests.setEnvironmentVariable("ZIG_EXE", b.graph.zig_exe);
```
matklad: Ditto here, ideally we shouldn't inject neither ver_options not ZIG_EXE here. Fine for now, but needs fixing

src/testing/systest/supervisor/nemesis.zig
```diff
@@ -0,0 +1,245 @@
+//! The nemesis injects faults into a running cluster, to test for fault tolerance. It's inspired
+//! by Jepsen, but is not so advanced.
+const std = @import("std");
+const Shell = @import("../../../shell.zig");
+const Replica = @import("./replica.zig");
+const LoggedProcess = @import("./logged_process.zig");
+
+const assert = std.debug.assert;
+const log = @import("log.zig");
```
matklad:
```suggestion
const log = std.log.scoped(.nemesis);
```

I _think_ we don't need the log.zig abstraction we are building here?

src/testing/systest/supervisor/log.zig
```diff
@@ -0,0 +1,63 @@
...
+    log_with_level(.info, scope, format, args);
+}
+
+pub fn err(
+    scope: []const u8,
+    comptime format: []const u8,
+    args: anytype,
+) void {
+    log_with_level(.info, scope, format, args);
+}
```
matklad: See comment in nemesis.zig --- I think we shouldn't be needing this abstraction.

src/testing/systest/supervisor/logged_process.zig
```diff
@@ -0,0 +1,280 @@
...
+child: ?std.process.Child = null,
+stdin_thread: ?std.Thread = null,
+stderr_thread: ?std.Thread = null,
+current_state: AtomicState,
+
+pub fn state(self: *Self) State {
+    return self.current_state.load(.seq_cst);
+}
+
+pub fn init(
```
matklad: init usually goes first

src/testing/systest/supervisor/logged_process.zig
```diff
@@ -0,0 +1,280 @@
...
+    allocator: std.mem.Allocator,
+    name: []const u8,
+    argv: []const []const u8,
+    options: Options,
+) !*Self {
+    var arena = std.heap.ArenaAllocator.init(allocator);
+
+    const cwd = try std.process.getCwdAlloc(arena.allocator());
```
matklad: Memory leak here if process creating fails. You want to add

```zig
errdefer arena.deinit()
```

above

src/testing/systest/supervisor/logged_process.zig
matklad: and, as this is returning an allocated pointer, let's call these `create/destroy` rather than `inint/deinit`

src/testing/systest/supervisor/logged_process.zig
```diff
@@ -0,0 +1,280 @@
...
+    allocator.destroy(self);
+}
+
+pub fn start(
+    self: *Self,
+) !void {
+    self.expect_state_in(.{ .initial, .terminated, .completed });
+    defer self.expect_state_in(.{.running});
+
+    var child = std.process.Child.init(self.argv, self.allocator);
```
matklad: Same here --- any try below can fail, in which case this child process is getting leaked.

src/testing/systest/supervisor/logged_process.zig
```diff
@@ -0,0 +1,280 @@
...
+                    if (line_opt) |line| {
+                        log.info(process.name, "{s}", .{line});
+                    } else {
+                        break;
+                    }
+                }
+            }
+        }.log_stderr,
+        .{ child.stderr.?, self },
+    );
```
matklad: Let's avoid this thread for know and just inherit stderr. I am 0.8 sure that, given that we are at least somewhat cooperative, we could push the requirement to print unambiguous logs down to each child. And if we do do that, the system as a whole becomes significantly simpler.

Mabybe I am wrong, and we do need log wrapping, but I'd start with the assumption that we don't, and see how far we could push that

src/testing/systest/supervisor/logged_process.zig
```diff
@@ -0,0 +1,280 @@
...
+    self.arena.deinit();
+    allocator.destroy(self);
+}
+
+pub fn start(
+    self: *Self,
+) !void {
+    self.expect_state_in(.{ .initial, .terminated, .completed });
+    defer self.expect_state_in(.{.running});
+
```
matklad: `assert(self.child == null)`

src/testing/systest/supervisor/logged_process.zig
```diff
@@ -0,0 +1,280 @@
...
+}
+
+test "LoggedProcess: starts and stops" {
+    var gpa = std.heap.GeneralPurposeAllocator(.{}){};
+    const allocator = gpa.allocator();
+    defer assert(gpa.deinit() == .ok);
+
+    const argv: []const []const u8 = &.{
+        "awk",
+        \\ BEGIN { for (i = 0; i < 10; i++) { print i > "/dev/fd/2"; } }
```
matklad: Let's remove `awk` dependency here

src/testing/systest/supervisor/systest.zig
```diff
@@ -0,0 +1,245 @@
...
+//! * filesystem fault injection?
+//! * cluster membership changes?
+
+const std = @import("std");
+const builtin = @import("builtin");
+const Shell = @import("../../../shell.zig");
+const LoggedProcess = @import("./logged_process.zig");
+const Replica = @import("./replica.zig");
+const Nemesis = @import("./nemesis.zig");
+const log = @import("log.zig");
```
matklad:
```suggestion
const log = std.log.scoped(.systest);
```

src/testing/systest/supervisor/replica.zig
```diff
@@ -0,0 +1,5 @@
+const LoggedProcess = @import("./logged_process.zig");
+
+name: []const u8,
+port: u16,
+process: *LoggedProcess,
```
matklad: Let's avoid adding a second `replica.zig` file, this is tiny enough to just inline at call-site

src/testing/systest/supervisor/nemesis.zig
```diff
@@ -0,0 +1,245 @@
...
+replicas: []const Replica,
+netem_rules: NetemRules,
+
+pub fn init(
+    shell: *Shell,
+    allocator: std.mem.Allocator,
+    random: std.rand.Random,
+    replicas: []const Replica,
+) !*Self {
+    const nemesis = try allocator.create(Self);
```
matklad: add `errdefer allocator.destroy(nemesis)`.

We don't have trys here yet, so this is dead code, but if we add a try below, the code would be incorrect. I guess you could also `errdefer comptime unreachable` if you really depend on anything below not being failable

src/testing/systest/supervisor/nemesis.zig
```diff
@@ -0,0 +1,245 @@
...
+const NetemRules = std.AutoHashMap(netem.Op, netem.Args);
+
+shell: *Shell,
+allocator: std.mem.Allocator,
+random: std.rand.Random,
+replicas: []const Replica,
+netem_rules: NetemRules,
+
+pub fn init(
+    shell: *Shell,
```
matklad: Nemesis is a long-running event loop, so using shell (arena allocation) here is not right. 

Let's push as much as possible to static allocation here. For example, replica's slice should never be larger than, say, 16. That means that things like `matching` shouldn't be array lists and could be bounded arrays instead.

Similarly, I bet we could give a bound on the number of rules we need here, etc.

src/testing/systest/supervisor/nemesis.zig
```diff
@@ -0,0 +1,245 @@
...
+        },
+        .network_loss_remove => return try self.netem_rules_delete(.loss),
+        .sleep => {
+            std.time.sleep(3 * std.time.ns_per_s);
+            return true;
+        },
+    }
+}
+
+fn random_replica_in_state(self: *Self, state: LoggedProcess.State) !?Replica {
```
matklad: There's no reason for this function to ever return an error

src/testing/systest/supervisor/nemesis.zig
```diff
@@ -0,0 +1,245 @@
...
+    };
+    const Args = []const []const u8;
+};
+
+fn netem_rules_add(self: *Self, op: netem.Op, args: netem.Args) !bool {
+    if (self.netem_rules.contains(op)) {
+        return false;
+    }
+    try self.netem_rules.put(op, try self.allocator.dupe([]const u8, args));
+    assert(try self.netem_sync());
```
matklad: dupe above leaks if this try returns an erro

matklad: I guess, either one dose not `try`, or one does not `alloc`. There's no `alloc`&`try` at the same time, that's too hard to get right!

matklad (commented):
Ok, so I am very satisfied with the big picture here, but I have doubts about the tactics-level implementation! I guess I have two kinds of concerns here:

* some things feel clearly wrong, but clearly temporary (us adding java dependency to our integration tests)
* but some things feel just wrong. For example, nemesis and systest are written in "let's allocate left and right and leak memory on errors". This is ok for provisional and temporary code, but these two things are going to be a core part of our very long-term testing strategy, and we should get this right. 
* and the meta concern here is that it's not super clear _which_ parts are "this is intentionally sloppy as we'll rewrite next week", and _which_ "this actually should be quite a bit more robust"

So, I think we have quite a bit to iterate on here still!

.github/workflows/ci.yml
owickstrom: Yeah, we should prioritize moving the workload to Zig and have a Zig driver as well (REPL or just direct client usage).

src/testing/systest/supervisor/nemesis.zig
owickstrom: You're right, it was mostly a convenience, but I'll scrap it.

src/testing/systest/supervisor/log.zig
owickstrom: Ah, now I remember, it was because I wanted to use `std.log.scoped(...)` but couldn't make it work with the generated names replica1, replica2, etc. But we should make do without the fancy prefixed logs, it introduces too much extra cruft.

src/testing/systest/supervisor/logged_process.zig
owickstrom: Hmm, wouldn't concurrent logs jumble together if we don't use `std.log` (which has some lock around stderr)?

src/testing/systest/supervisor/logged_process.zig
owickstrom: Rewrote it with `yes`, is that better? I'm having a hard time finding something better that can be spawned for this test. Suggestions welcome!

src/testing/systest/supervisor/nemesis.zig
owickstrom: Fixed

src/testing/systest/supervisor/logged_process.zig
owickstrom: Removing the arena entirely instead.

src/testing/systest/supervisor/logged_process.zig
owickstrom: I don't think `Child.init` heap-allocates? It's just a stack-allocated struct holding the passed-in allocator.

matklad: after `.spawn`, child owns a file descriptor. This fd leaks if, eg, try format_argv fails (together with leaking the entire process)

src/testing/systest/supervisor/logged_process.zig
owickstrom: Fixed :+1:

src/testing/systest/supervisor/logged_process.zig
owickstrom: Ah, fixed now.

owickstrom:
OK, I think all easy fixes (sloppy memory management and such) are done now. Two things remaining essentially:

1. Log-wrapping of stderr, what should be done there
1. Rewriting workload in Zig
    1. Only using a Zig driver for the integration test in CI, and
    1. keeping the Java thing separate from our regular CI.
    1. This also, I believe, will remove the need for passing in ZIG_EXE and vsr_options into the integration test

src/integration_tests.zig
```diff
@@ -484,3 +486,54 @@ test "in-place upgrade" {
...
+
+    // Build workload
+    {
+        try shell.pushd("./src/testing/systest/workload");
+        defer shell.popd();
+
+        try shell.exec("mvn clean install --batch-mode --quiet", .{});
+    }
+
+    log.info("Running 1m systest...", .{});
```
matklad: We _generally_ expect log messages to be prefixed, so we don't really start them with a capital letter.

src/testing/systest/supervisor/systest.zig
```diff
@@ -0,0 +1,241 @@
...
+const replica_ports = [replica_count]u16{ 3000, 3001, 3002 };
+
+pub const CLIArgs = struct {
+    tigerbeetle_executable: []const u8,
+    test_duration_minutes: u16 = 10,
+};
+
+pub fn main(shell: *Shell, allocator: std.mem.Allocator, args: CLIArgs) !void {
+    if (builtin.os.tag == .windows) {
+        log.err("supervisor: systest is not supported for Windows", .{});
```
matklad:
```suggestion
        log.err("systest is not supported for Windows", .{});
```

We don't put general area prefixes here. We sometimes put the name of the function there.

src/testing/systest/supervisor/systest.zig
```diff
@@ -0,0 +1,241 @@
...
+        return error.NotSupported;
+    }
+
+    const tmp_dir = try shell.create_tmp_dir();
+    defer shell.cwd.deleteDir(tmp_dir) catch {};
+
+    // Check that we are running as root
+    if (!std.mem.eql(u8, try shell.exec_stdout("id -u", .{}), "0")) {
+        log.err(
+            "This script needs to be run in a separate namespace using `unshare -nfr`",
```
matklad:
```suggestion
            "this script needs to be run in a separate namespace using 'unshare -nfr'",
```

capitalization, and we conventionally use single quotes in log messages (not sure if that's the right convention, but that's, e.g, what our cli.zig uses)

src/testing/systest/supervisor/systest.zig
```diff
@@ -0,0 +1,241 @@
...
+pub fn main(shell: *Shell, allocator: std.mem.Allocator, args: CLIArgs) !void {
+    if (builtin.os.tag == .windows) {
+        log.err("supervisor: systest is not supported for Windows", .{});
+        return error.NotSupported;
+    }
+
+    const tmp_dir = try shell.create_tmp_dir();
+    defer shell.cwd.deleteDir(tmp_dir) catch {};
+
+    // Check that we are running as root
```
matklad:
```suggestion
    // Check that we are running as root.
```

Comments are proper sentences. This includes punctuation. 

https://github.com/tigerbeetle/tigerbeetle/blob/cd8f4e556c0cdb9726e838f21a73bc5a078fcb71/docs/TIGER_STYLE.md#L326-L327

src/testing/systest/supervisor/systest.zig
```diff
@@ -0,0 +1,241 @@
...
+            .{},
+        );
+        std.process.exit(1);
+    }
+
+    // Ensure loopback can be used
+    try shell.exec("ip link set up dev lo", .{});
+
+    log.info(
+        "supervisor: starting test with target runtime of {d}m",
```
matklad:
```suggestion
        "starting test with target runtime of {d}m",
```

src/testing/systest/supervisor/systest.zig
```diff
@@ -0,0 +1,241 @@
...
+    );
+    const test_duration_ns = @as(u64, @intCast(args.test_duration_minutes)) * std.time.ns_per_min;
+    const time_start = std.time.nanoTimestamp();
+
+    var replicas: [replica_count]*LoggedProcess = undefined;
+    for (0..replica_count) |i| {
+        const name = try shell.fmt("replica{d}", .{i});
+        const datafile = try shell.fmt("{s}/1_{d}.tigerbeetle", .{ tmp_dir, i });
+
+        // Format datafile
```
matklad:
```suggestion
        // Format datafile.
```

src/testing/systest/supervisor/systest.zig
```diff
@@ -0,0 +1,241 @@
...
+const builtin = @import("builtin");
+const Shell = @import("../../../shell.zig");
+const LoggedProcess = @import("./logged_process.zig");
+const Nemesis = @import("./nemesis.zig");
+const log = std.log.scoped(.systest);
+
+const assert = std.debug.assert;
+
+const replica_count = 3;
+const replica_ports = [replica_count]u16{ 3000, 3001, 3002 };
```
matklad: maybe just

```
const cluster_address = "3000,3001,3002";
comptime {
    assert(std.mem.count(cluster_addresses, ",") == replica_count - 1)
}
```

?

That'll avoid leaky comma_separated_ports (it currently incorrectly handles allocation failures)

src/testing/systest/supervisor/systest.zig
```diff
@@ -0,0 +1,241 @@
...
+    const nemesis = try Nemesis.init(allocator, prng.random(), &replicas);
+    defer nemesis.deinit();
+
+    // Let the workload finish by itself, or kill it after we've run for the required duration.
+    // Note that the nemesis is blocking in this loop.
+    const workload_result = term: {
+        while (std.time.nanoTimestamp() - time_start < test_duration_ns) {
+            // Try to do something funky in the nemesis, and if it fails, wait for a while
+            // before trying again.
+            if (!try nemesis.wreak_havoc()) {
```
matklad: It's unclear what the boolean here means, and the comment doesn't really explain it. "fails" is  `try` part, but we also have a `bool`?

src/testing/systest/supervisor/systest.zig
```diff
@@ -0,0 +1,241 @@
...
+
+    // Set up nemesis (fault injector)
+    var prng = std.rand.DefaultPrng.init(0);
+    const nemesis = try Nemesis.init(allocator, prng.random(), &replicas);
+    defer nemesis.deinit();
+
+    // Let the workload finish by itself, or kill it after we've run for the required duration.
+    // Note that the nemesis is blocking in this loop.
+    const workload_result = term: {
+        while (std.time.nanoTimestamp() - time_start < test_duration_ns) {
```
matklad: I personally prefer writing these kinds of loops in the deadline form:

```zig
while (std.time.nanoTimestamp() < deadline)
```

With time, I always confuse if it is `a-b` or `b-a`, and the deadline form avoids that. 

Though, I am also known for _always_ messing up any retry-timeout related code, so don't listen to me on this one 🤣

src/testing/systest/supervisor/systest.zig
```diff
@@ -0,0 +1,241 @@
...
+    errdefer workload.destroy();
+
+    // Set up nemesis (fault injector)
+    var prng = std.rand.DefaultPrng.init(0);
+    const nemesis = try Nemesis.init(allocator, prng.random(), &replicas);
+    defer nemesis.deinit();
+
+    // Let the workload finish by itself, or kill it after we've run for the required duration.
+    // Note that the nemesis is blocking in this loop.
+    const workload_result = term: {
```
matklad: don't need a block here, zig is python enough:

```
const result = while(...) {
   break x;
} else y;
```

src/testing/systest/supervisor/systest.zig
matklad: I am not sure if it makes sense to separate nemesis in this way. `nemesis` also manages the processes, so both this `main` function and `nemesis` are effectively vying for control over the replicas slice. So maybe factor it slightly differently?

Have a single thing that maintains the slice of replicas, the workload, and implements the event loop. Then, separate `netmem` in a separate file, which _doesn't_ have access to the rest of the state. 

So one file is all the state and event-loop, and the other is mostly stateless poking at OS interfaces to week havoc?

src/testing/systest/supervisor/systest.zig
```diff
@@ -0,0 +1,241 @@
...
+        class_path,
+        "Main",
+    });
+
+    var env = try std.process.getEnvMap(shell.arena.allocator());
+    try env.put("REPLICAS", try comma_separate_ports(shell.arena.allocator(), &replica_ports));
+
+    var process = try LoggedProcess.create(allocator, name, argv, .{ .env = &env });
+    try process.start();
+    return process;
```
matklad: I am not a fan of such two-step create/start split in the LoggerProcess API. Or rather, I am a fan of what you are getting at conceptually here, but I think this split is not quite right. The way I see it, we have two things here:

* a single instance of process with a specific pid
* a _program_ which we kill&restart, and which gives rise to many instances of processes

And LoggerProcess tires to model the second concept. But I think you actually can't model that cleanly, given our upgrade story. 

During upgrades, you might actually have _two_ tigerbeetle binaries, and you need to check both flows:

* replace the binary on disk, and check that the process notices that and re-execs
* kill the old process, start the new process using a new binary in the different location. 

So it feels like it is perhaps best to _not_ try to model this, and leave mapping of executables on disk to processes to the main event loop, and instead narrowly scope `LoggerProcess` to be a one-shot process/pid?

src/testing/systest/supervisor/systest.zig
```diff
@@ -0,0 +1,241 @@
...
+        if (replica.state() == .running) {
+            _ = try replica.terminate();
+        }
+        replica.destroy();
+    }
+
+    switch (workload_result) {
+        .Exited => |code| {
+            if (code == 128 + std.posix.SIG.KILL) {
+                log.info("supervisor: workload terminated (SIGKILL) as requested", .{});
```
matklad: Hm, this is not right I think? 128+signal is the shell thing, but we don't have a shell here, so we have proper .Exited/.Signal thing here.

src/testing/systest/supervisor/logged_process.zig
```diff
@@ -0,0 +1,329 @@
+//! Runs a subprocess and logs its stderr output. The processes can terminate by its own or be
+//! actively terminated with a SIGKILL, and later be restarted again.
+//!
+//! We use this to run a cluster of replicas (and the workload), and terminate and restart replicas
+//! to test fault tolerance.
+//!
+//! NOTE: some duplication exists between this and `tmp_tigerbeetle.zig` that could perhaps be
+//! unified.
```
matklad: As per comment in main, consider making this thing one-shot, and model restart by the caller creating a new logged_process

src/testing/systest/supervisor/logged_process.zig
```diff
@@ -0,0 +1,329 @@
...
+                        continue;
+                    };
+                    if (line_opt) |line| {
+                        log.info("{s}: {s}", .{ process.name, line });
+                    } else {
+                        break;
+                    }
+                }
+            }
+        }.log_stderr,
```
matklad: Good point about stderr not having a global lock, but I still don't feel good about this code. I must confess, at this point, this is 80% gut feeling, and I am totally ready to eventually admit that I was wrong here, but just not right away :-) 

I think we need to do _something_ with logs and error streams, but I am very not sure that a separate stderr reading thread is the correct something _here_, and I fear that implementing a wrong solution would disincentivize us from finding the right, simple solution here. 

So, let's "do nothing" here, let's just inherit stderr stream, make a mental note to "fix this later", and revisit it later. 

I have a suspicion that this _might_ actually be fine, provided that the child processes obey three rules:

* Each process log is self-identifying 
* Each log message fits into 4k limit of kernel's atomic writes
* Each log message is a result of a single `.write` syscall

If all the above holds, then I thing the log would _not_ be messed up. And, given that we write the child processes, we have the power to make the above points true, and they _seems_ like good ideas nonetheless.

src/testing/systest/supervisor/nemesis.zig
```diff
@@ -0,0 +1,290 @@
...
+        .sleep => {
+            std.time.sleep(3 * std.time.ns_per_s);
+            return true;
+        },
+    }
+}
+
+fn random_replica_in_state(self: *Self, state: LoggedProcess.State) ?*LoggedProcess {
+    var matching: [replicas_count_max]*LoggedProcess = undefined;
+    var pos: u8 = 0;
```
matklad: 1. We don't generally contract words, so, `position` (curiously, I used to be a `let mut res` person before joining TB, but I am in the `var result` camp now!)
2. We don't usually use `position`, it is rather `index` or `count`

src/testing/systest/supervisor/nemesis.zig
```diff
@@ -0,0 +1,290 @@
...
+        if (replica.state() == state) {
+            matching[pos] = replica;
+            pos += 1;
+        }
+    }
+    if (pos == 0) {
+        return null;
+    }
+    std.rand.shuffle(self.random, *LoggedProcess, matching[0..pos]);
+    return matching[0];
```
matklad: Why shuffle when you can select a random index? Alternatively, could go all the way to the reservoir sampling here (I _think_ Knuth shuffle is reservoir sampling in disguise anyway?)

src/testing/systest/supervisor/systest.zig
owickstrom: Fixed `comma_separated_ports` instead. :) I have a strong suspicion we will want those port values. Maybe premature optimization now, but the fix made more sense to me.

src/testing/systest/supervisor/nemesis.zig
owickstrom: I actually don't know. Probably had some other idea at first and it got left over! Fixed.

src/testing/systest/supervisor/systest.zig
owickstrom: When we used SIGTERM it needed to be handled this way, because the JVM exited with (128 + signal). See http://journal.thobe.org/2013/02/jvms-and-kill-signals.html.

But I think now that we use SIGKILL it might not be needed anymore. I'll try dropping it.

src/testing/systest/supervisor/systest.zig
owickstrom: Let's go with that!

src/testing/systest/supervisor/logged_process.zig
owickstrom: Time for curveball: I think we should keep this, because I want to catch `err` logs from workload/replica and fail the systest on that. In the workload case, there shouldn't be error logs, just crashes, but for TB replicas we saw error logs without crashes. I'd rather be defensive here than relying on convention across all our future drivers and changes in replicas.

owickstrom: And maybe there are reasonable cases where a replica should err-log but not crash?

matklad: Hm, I'd maybe avoid parsing the logs. This is again, gut feeling, but for cfo, things definitely improved once we stopped trying to collect the logs, and instead went for binary "does it exit with zero". I think it makes more sense to treat this as a black box --- there's a certain contract that the cluster should fulfill, and it doesn't _really_ matter whether replica logs something or not.

src/testing/systest/supervisor/systest.zig
owickstrom: OK, I've made LoggedProcess a one-shot process, and have the restart logic in a new Replica struct that maintains that lifecycle. That's where we can put the upgrade logic and such later. I'm also considering move the datafile formatting there, but I'm not sure.

Please let me know what you think about this new split.

src/testing/systest/supervisor/systest.zig
owickstrom: Removed :+1:

src/testing/systest/supervisor/logged_process.zig
owickstrom: So, I tried, at it seems to hold so far. Let's give it a chance.

src/testing/systest/supervisor/systest.zig
owickstrom: OK, I have moved some responsibility from nemesis to the systest main control flow. Not the full slice of replicas and the termination action, though. It seems reasonable to keep these "bad" actions and their probabilities together in the replica, so I refrained from moving `terminate_replica` to systest.

owickstrom: After sleeping on it, I'm gonna try moving more control to main. :)

owickstrom: It's much simpler now. :+1:

src/testing/systest/supervisor/logged_process.zig
```diff
@@ -0,0 +1,299 @@
...
+const assert = std.debug.assert;
+
+const Self = @This();
+pub const State = enum(u8) { running, terminated, completed };
+const AtomicState = std.atomic.Value(State);
+const Options = struct { env: ?*const std.process.EnvMap = null };
+
+// Passed in to init
+allocator: std.mem.Allocator,
+argv: []const []const u8,
```
matklad: Let's avoid storing argv. We aren't actually using it anywhere, and, if we don't store it, then the caller wouldn't have to dupe it!

src/testing/systest/supervisor/logged_process.zig
```diff
@@ -0,0 +1,299 @@
...
+
+// Lifecycle state
+current_state: AtomicState,
+
+pub fn spawn(
+    allocator: std.mem.Allocator,
+    argv: []const []const u8,
+    options: Options,
+) !*Self {
+    const cwd = try std.process.getCwdAlloc(allocator);
```
matklad: Let's add a sys test-wide invariant that everything that runs, runs with systest-own cwd. That way, we could completely remove any mention of cwd from everywhere. 

It would be up to sestet's caller to place it in the correct directory.

src/testing/systest/supervisor/logged_process.zig
```diff
@@ -0,0 +1,299 @@
...
+        .argv = argv,
+        .options = options,
+        .current_state = AtomicState.init(.running),
+        .child = std.process.Child.init(argv, allocator),
+        .stdin_thread = undefined,
+        .stderr_thread = undefined,
+    };
+
+    self.child.cwd = cwd;
+    self.child.env_map = options.env;
```
matklad: Same for the environment --- I think we can add a constraint that we don' use environmental variables and rely only on the explicit arguments. That way, we won't have to mess up with env. 

In general, we control all the processes that we've spawn, so we should totally add as much constraints on what the children can do to simplify the whole.

src/testing/systest/supervisor/logged_process.zig
```diff
@@ -0,0 +1,299 @@
...
+
+const assert = std.debug.assert;
+
+const Self = @This();
+pub const State = enum(u8) { running, terminated, completed };
+const AtomicState = std.atomic.Value(State);
+const Options = struct { env: ?*const std.process.EnvMap = null };
+
+// Passed in to init
+allocator: std.mem.Allocator,
```
matklad: Let's not store allocator here. Shell stores an allocator as an exception, but production code should take it as an explicit parameter in init/deinit. See

* https://zsf.zulipchat.com/#narrow/stream/454371-std/topic/Xunmanaged.20-.3E.20X
* https://matklad.github.io/2020/12/28/csdi.html

for some background info on the pattern

src/testing/systest/supervisor/logged_process.zig
```diff
@@ -0,0 +1,299 @@
...
+    // See: https://github.com/ziglang/zig/issues/16820
+    _ = kill: {
+        if (builtin.os.tag == .windows) {
+            const exit_code = 1;
+            break :kill std.os.windows.TerminateProcess(self.child.id, exit_code);
+        } else {
+            break :kill std.posix.kill(self.child.id, std.posix.SIG.KILL);
+        }
+    } catch |err| {
+        std.debug.print(
```
matklad: log.err

In general, never use `std.debug.print` (I used it a bunch in various auxiliary support files, and I now regret that 🤣 )

src/testing/systest/supervisor/logged_process.zig
```diff
@@ -0,0 +1,299 @@
...
+    }
+
+    log.err("expected state in {any} but actual state is {s}", .{
+        valid_states,
+        @tagName(actual_state),
+    });
+    unreachable;
+}
+
+fn format_argv(allocator: std.mem.Allocator, argv: []const []const u8) ![]const u8 {
```
matklad: Feels like its a dead code now?

src/testing/systest/supervisor/network_faults.zig
```diff
@@ -0,0 +1,173 @@
...
+            .{loss.loss_pct},
+        ));
+        args.appendAssumeCapacity(try std.fmt.bufPrint(
+            args_loss_correlation[0..],
+            "{d}%",
+            .{loss.correlation_pct},
+        ));
+    }
+
+    // Everything stack-allocated and simple up to here. Now this feels like a shame just for
```
matklad: It feels indeed! Would be great if this were a thing:

```zig
log.info("syncing netem: {}", .{stdx.fmt_join("{}", " ", args.items)});

// in stdx.zig
fn fmt_join(comptime item_fmt: []const u8, sep: []const u8, items: anytype)
```

I _think_ this should be possible, although I don't remember the relevant fmt APIs off the top of my head my head. Could you them up and implement this(maybe as a separate PR to stdx even)?  I remember wanting this a couple of time and chickening out of actually doing the work :-)

matklad: https://github.com/ziglang/zig/blob/ea527f7a850f0200681630d8f36131eca31ef48b/lib/std/fmt.zig#L833-L840 is the pattern to follow

src/testing/systest/driver.zig
```diff
@@ -0,0 +1,147 @@
...
+) !void {
+    assert(events.len <= events_count_max);
+
+    const writer = self.child.stdin.?.writer().any();
+
+    try writer.writeInt(u8, @intFromEnum(request_type), .little);
+    try writer.writeInt(u32, @intCast(events.len), .little);
+
+    const bytes = std.mem.sliceAsBytes(events);
+    const size_written = try writer.write(bytes);
```
matklad: This wants to be a `writeAll` I think?

owickstrom: Good catch! That API difference is subtle...

.github/workflows/ci.yml
owickstrom: Fixed!

build.zig
owickstrom: Fixed!

owickstrom:
> OK, I think all easy fixes (sloppy memory management and such) are done now. Two things remaining essentially:
>
> *  Log-wrapping of stderr, what should be done there
>  *  Rewriting workload in Zig 

Both fixed now.

Things I'll do in a separate PR:

* Write a Java driver, so that we can remove the old systest workload
* Implement retries in the Zig workload

build.zig
```diff
@@ -622,16 +633,30 @@ fn build_test_integration(b: *std.Build, step_test_integration: *std.Build.Step,
...
+    const step_vortex = b.step(
+        "vortex_integration_test",
+        "special build of vortex for multiversion integration test",
+    );
+    const vortex_exe = build_vortex(b, step_vortex, .{
+        .tb_client_header = options.tb_client_header,
+        .vsr_options = vsr_options,
+        .target = options.target,
+        .mode = options.mode,
+    });
```
matklad, line 659: Let's maybe pass vortex lazy_path through `options`? Right now, the `build voretx` and the vortex used by integration tests are different, so you'll get unnecessary rebuild.

The fact that integration tests build a separate copy of tiger beetle is an exception --- we intentionally use different option from the main tiger beetle build. Here, we could re-use vortex as is.

build.zig
```diff
@@ -622,16 +633,30 @@ fn build_test_integration(b: *std.Build, step_test_integration: *std.Build.Step,
...
     integration_tests_options.addOptionPath("tigerbeetle_exe_past", tigerbeetle_previous);
+    integration_tests_options.addOptionPath("vortex_exe", vortex_exe);
     const integration_tests = b.addTest(.{
         .root_source_file = b.path("src/integration_tests.zig"),
         .target = options.target,
         .optimize = options.mode,
         .filters = b.args orelse &.{},
     });
     integration_tests.root_module.addOptions("test_options", integration_tests_options);
+    integration_tests.step.dependOn(step_vortex);
```
matklad: You shouldn't need an explicit dependency. The thing about the `LazyPath` is that it auto-magically tracks deps. So, just `integration_tests_options.addOptionPath("vortex_exe", vortex_exe);` already adds the required dep!

build.zig
```diff
@@ -798,6 +823,33 @@ fn build_scripts(
...
+) std.Build.LazyPath {
+    const vortex = b.addExecutable(.{
+        .name = "vortex",
+        .root_source_file = b.path("src/vortex.zig"),
+        .target = options.target,
+        .optimize = options.mode,
+    });
+
+    vortex.root_module.addOptions("vsr_options", options.vsr_options);
+    vortex.linkLibC();
```
matklad, line 857: Why do we need `linkLibC`? libc is not nodejs-level banned, but is definitely not encouraged, a kernel is all you need ;)

matklad: Ahhh, that is because we use c client in the default driver! Smart! And we need libc for the C client for a bad reason (current version of Zig requires it for cImport).

src/testing/vortex/supervisor.zig
```diff
@@ -0,0 +1,336 @@
...
+        out = out ++ std.fmt.comptimePrint(",{d}", .{port});
+    }
+
+    assert(std.mem.count(u8, out, ",") == ports.len - 1);
+    return out;
+}
+
+// Create a new Vortex-specific temporary directory. Caller owns the returned memory.
+fn create_tmp_dir(allocator: std.mem.Allocator) ![]const u8 {
+    const root = process.getEnvVarOwned(allocator, "TMPDIR") catch try allocator.dupe(u8, "/tmp");
```
matklad: Let's use `.zig-cache/tmp` for our temp dir. I think we can assume that supervisor is being run with cwd pointing at the repository root? 

That way, we don't depend on ambient environment (TMPDIR) and actually keep stuff scoped to the project. And other things already use zig-cache/tmp, so we also gain uniformity here.

src/testing/vortex/supervisor.zig
```diff
@@ -0,0 +1,336 @@
...
+    if (builtin.os.tag == .windows) {
+        log.err("vortex is not supported for Windows", .{});
+        return error.NotSupported;
+    }
+
+    const tmp_dir = try create_tmp_dir(allocator);
+    defer allocator.free(tmp_dir);
+    defer std.fs.cwd().deleteTree(tmp_dir) catch |err| {
+        log.err("failed deleting temporary directory ({s}): {any}", .{ tmp_dir, err });
+    };
```
matklad: Let's do a single

```
defer {

};
```

here to avoid thinking about the order of evaluation.

src/testing/vortex/supervisor.zig
```diff
@@ -0,0 +1,336 @@
...
+    log.info(
+        "starting test with target runtime of {d}m",
+        .{args.test_duration_minutes},
+    );
+    const test_duration_ns = @as(u64, @intCast(args.test_duration_minutes)) * std.time.ns_per_min;
+    const test_deadline = std.time.nanoTimestamp() + test_duration_ns;
+
+    var replicas: [replica_count]*Replica = undefined;
+    inline for (0..replica_count) |i| {
+        var datafile_buf: [1024]u8 = undefined;
```
matklad: Use `std.fs.max_path_bytes` here

src/testing/vortex/supervisor.zig
```diff
@@ -0,0 +1,336 @@
...
+        const result = try process.Child.run(.{ .allocator = allocator, .argv = &.{
+            args.tigerbeetle_executable,
+            "format",
+            std.fmt.comptimePrint("--cluster={d}", .{cluster_id}),
+            std.fmt.comptimePrint("--replica={d}", .{i}),
+            std.fmt.comptimePrint("--replica-count={d}", .{replica_count}),
+            datafile,
+        } });
+        defer allocator.free(result.stderr);
+        defer allocator.free(result.stdout);
```
matklad: Same here, merge defers. Defer can not fail, so it is always correct to merge them.

src/testing/vortex/supervisor.zig
```diff
@@ -0,0 +1,336 @@
...
+                if (code != 0) {
+                    log.err("failed formatting datafile: {s}", .{result.stderr});
+                    process.exit(1);
+                }
+            },
+            else => {
+                log.err("failed formatting datafile: {s}", .{result.stderr});
+                process.exit(1);
+            },
+        }
```
matklad: I think you can do

```zig
        if (std.meta.eql(result.term, .{ .Exited = 0})) {
            
        }
```

here?

src/testing/vortex/supervisor.zig
matklad: Missing 

```
        errdefer for (replicas[0..i]) |replica| replica.destroy();
```

to free stuff if the middle of the loop blows up

matklad: also maybe `s/i/replica_index`?

src/testing/vortex/logged_process.zig
```diff
@@ -0,0 +1,259 @@
...
+    allocator.destroy(self);
+}
+
+pub fn state(self: *Self) State {
+    return self.current_state.load(.seq_cst);
+}
+
+pub fn terminate(
+    self: *Self,
+    signal: comptime_int,
```
matklad: Why pass the signal here? We should only ever support abrupt, ungraceful shutdown, so this always will be a sigkill.

src/testing/vortex/replica.zig
matklad, line 1: let's avoid the second `replica.zig` file in the codebase. `replica.zig` is the main file of the entire thing, it is valuable that you can unambiguously go to definition to it.

src/testing/vortex/replica.zig
```diff
@@ -0,0 +1,96 @@
...
+
+pub fn state(self: *Self) State {
+    if (self.process) |process| {
+        switch (process.state()) {
+            .running => return State.running,
+            .terminated => return State.terminated,
+            .completed => return State.completed,
+        }
+    } else return State.initial;
+}
```
matklad:
```suggestion
    if (self.process) |process| {
        switch (process.state()) {
            .running => return .running,
            .terminated => return .terminated,
            .completed => return .completed,
        }
    } else return .initial;
}
```

src/testing/vortex/replica.zig
```diff
@@ -0,0 +1,96 @@
...
+    if (self.process) |process| {
+        process.destroy(self.allocator);
+    }
+
+    const addresses = try std.fmt.allocPrint(
+        self.allocator,
+        "--addresses={s}",
+        .{self.replica_addresses},
+    );
+    defer self.allocator.free(addresses);
```
matklad: let's pre-compute this in init to avoid a random allocation here.

src/testing/vortex/workload.zig
```diff
@@ -0,0 +1,407 @@
...
+const pending_transfers_count_max = 1024;
+const accounts_count_max = 128;
+
+pub const CLIArgs = struct {
+    @"cluster-id": u128,
+    addresses: []const u8,
+    @"driver-command": []const u8,
+};
+
+pub fn main(allocator: std.mem.Allocator, args: CLIArgs) !void {
```
matklad: Let's make the interface more crisp here. Instead of workload spawning the driver, let's just pass in driver-s pipes:

```
fn main(
  allocator, 
  args, 
  pipes: struct {in : std.fs.File, out: std.fs.File }

)
```

and let the superviser to spawn the driver.

build.zig
owickstrom: I think we can't, actually, because the versions from `vsr_options` have to match, otherwise the client can't talk to the replicas (because it would be on version 0.0.1).

build.zig
owickstrom: Nice, trying that.

build.zig
owickstrom: That's it!

src/testing/vortex/supervisor.zig
owickstrom: It feels a little off. I don't have super strong opinions about it but it would preclude us from:

1. compiling Vortex, shipping the executable to some VM, running it there (might not be a use case, ever)
2. lettings others use Vortex outside or repo (this seems like a better reason to make it cwd-portable)

owickstrom: We could do a fallback solution? Check if `.zig-cache/tmp` exists and use that, or `TMPDIR`, or `/tmp`. But it's feels like bad magic.

src/testing/vortex/logged_process.zig
owickstrom: Leftover, I did an experiment with SIGTERM that didn't pan out. Removing again.

src/testing/vortex/workload.zig
owickstrom: I'm not sure what you mean here. The workload runs in a separate process, so we'd have to pass in file descriptors? Or am I missing something?

owickstrom: e.g. `/proc/<PID>/fd/0`, with the PID of the externally spawned driver?

build.zig
owickstrom: Sadly didn't work.

src/testing/vortex/supervisor.zig
owickstrom: Had a `defer` for replicas below that I'm moving up.

src/testing/vortex/replica.zig
owickstrom: Just rename, or do you think I should put it in `supervisor.zig`?

owickstrom: It's only about 100 LOC.

src/testing/vortex/supervisor.zig
matklad: I think if someone uses vortex outside of the repo, they could live with `.zig-cache` created in the cwd?

owickstrom: I guess? It's a little weird but probably not a problem.

build.zig
owickstrom: Fixed!

src/testing/vortex/supervisor.zig
owickstrom: Fixed.

src/testing/vortex/replica.zig
owickstrom: Done!

build.zig
owickstrom: Closing this, feel free to reopen.

matklad: Yeah, you are right, I missed that we use a different version here!

src/testing/vortex/workload.zig
owickstrom: OK, interface is now a lot crisper, and that let me delete a good amount of code as well!

src/testing/vortex/workload.zig
```diff
@@ -0,0 +1,432 @@
+//! This workload runs in an loop, generating and executing operations on a cluster through a
+//! _driver_.
+//!
+//! Any sucessful operations are reconciled with a model, tracking what accounts exist. Future
```
matklad:
```suggestion
//! Any successful operations are reconciled with a model, tracking what accounts exist. Future
```

src/testing/vortex/workload.zig
```diff
@@ -0,0 +1,432 @@
...
+const pending_transfers_count_max = 1024;
+const accounts_count_max = 128;
+
+const DriverStdio = struct { input: std.fs.File, output: std.fs.File };
+
+pub fn main(
+    allocator: std.mem.Allocator,
+    driver: *const DriverStdio,
+) !void {
+    var model = Model{
```
matklad, line 50: It's OK to use dynamic allocation here for now, but I'd seriously consider refactoring
this to static alloc, to make code _easier_ to understand. Static alloc is harder to code for
the first time, as you need to actually _understand_ how many accounts/pending transfers you
could have, but, once understood and encoded in the source code, this understanding can be
re-used for free by future readers.

In terms of code, that's s/AutoHashMap/AutoHashMapUnmanaged

src/testing/vortex/workload.zig
```diff
@@ -0,0 +1,432 @@
...
+
+        const query = lookup_all_accounts(&model);
+        const query_result = try execute(query, driver) orelse break;
+        try reconcile(query_result, &query, &model);
+
+        log.info("accounts created = {d}, commands run = {d}", .{ model.accounts.items.len, i });
+    }
+}
+
+var accounts_buffer = std.mem.zeroes([accounts_count_max]tb.Account);
```
matklad: Normally, I would advice against using global variables in such a way, but it seems like the workload is a genuine exception, where the globals make sense!

Though, we _do_ pass model as a parameter explicitly everywhere, so maybe just pull this over to model?

src/testing/vortex/workload.zig
```diff
@@ -0,0 +1,432 @@
...
+const Result = union(enum) {
+    create_accounts: []tb.CreateAccountsResult,
+    create_transfers: []tb.CreateTransfersResult,
+    lookup_all_accounts: []tb.Account,
+};
+const ResultBuffers = FixedSizeBuffers(Result);
+var result_buffers: ResultBuffers = std.mem.zeroes(ResultBuffers);
+
+fn execute(command: Command, driver: *const DriverStdio) !?Result {
+    switch (command) {
```
matklad, line 92: Could this be `inline else => |tag|` to avoid code duplication?

src/testing/vortex/workload.zig
```diff
@@ -0,0 +1,432 @@
...
+            var accounts_created: [accounts_count_max]struct {
+                tb.CreateAccountResult,
+                tb.Account,
+            } = undefined;
+            for (
+                accounts_new,
+                accounts_created[0..accounts_new.len],
+            ) |account_new, *account_created| {
+                account_created[0] = .ok;
+                account_created[1] = account_new;
```
matklad: Feels like a useless copy here? Can you use accounts_new as is below? That is,
have two parallels arrays rather than an array of pairs?

src/testing/vortex/workload.zig
```diff
@@ -0,0 +1,432 @@
...
+
+                // No further validation needed for failed transfers.
+                if (entry.result != .ok) {
+                    continue;
+                }
+
+                if (transfer.flags.pending) {
+                    try testing.expect(!model.pending_transfers.contains(transfer.id));
+                    try model.pending_transfers.put(transfer.id, {});
+                    // We can't let this grow unboundedly:
```
matklad: Ah, so if we assert this here, there's no reason _not_ to use putAssumingCapacity!

src/testing/vortex/workload.zig
```diff
@@ -0,0 +1,432 @@
...
+    }
+}
+
+/// Tracks information about the accounts and transfers created by the workload.
+const Model = struct {
+    accounts: std.ArrayListUnmanaged(tb.Account),
+    pending_transfers: std.AutoHashMap(u128, void),
+
+    // O(n) lookup, but it's limited by `accounts_count_max`, so it's OK for this test.
+    fn account_exists(self: @This(), id: u128) bool {
```
matklad: Maybe model rather than self? We try to use self less, to keep the naming unfimorm between methods and functions.

src/testing/vortex/workload.zig
```diff
@@ -0,0 +1,432 @@
...
+        .lookup_all_accounts = 0,
+    }).?;
+    switch (command_tag) {
+        .create_accounts => return random_create_accounts(random, model),
+        .create_transfers => return random_create_transfers(random, model),
+        .lookup_all_accounts => unreachable,
+    }
+}
+
+fn random_create_accounts(random: std.Random, model: *const Model) !Command {
```
matklad: No need for `!` here?

src/testing/vortex/workload.zig
```diff
@@ -0,0 +1,432 @@
...
+            .ledger = 1,
+            .code = random.intRangeAtMost(u16, 1, 100),
+            .flags = flags,
+        });
+    }
+
+    return .{ .create_accounts = events[0..events_count] };
+}
+
+fn random_create_transfers(random: std.Random, model: *const Model) !Command {
```
matklad: and here?

src/testing/vortex/zig_driver.zig
```diff
@@ -0,0 +1,128 @@
...
+) !void {
+    switch (operation) {
+        inline else => |comptime_operation| {
+            const size = @sizeOf(StateMachine.Result(comptime_operation));
+            if (size > 0) {
+                const count = @divExact(result.len, size);
+                try writer.writeInt(u32, @intCast(count), .little);
+                try writer.writeAll(result);
+            } else {
+                log.err(
```
matklad, line 138: Do we want to crash here? Otherwise, the workload might hang I think?

src/testing/vortex/zig_driver.zig
```diff
@@ -0,0 +1,128 @@
...
+            assert(count <= events_count_max);
+            const buffer_size_max =
+                @sizeOf(StateMachine.Event(comptime_operation)) * events_count_max;
+
+            var buffer: [buffer_size_max]u8 = undefined;
+
+            const buffer_size = @sizeOf(StateMachine.Event(comptime_operation)) * count;
+            assert(try reader.readAtLeast(buffer[0..buffer_size], buffer_size) == buffer_size);
+
+            return .{ comptime_operation, buffer[0..buffer_size] };
```
matklad: Variable `buffer` does not live long enough ;-)

src/testing/vortex/workload.zig
owickstrom: Makes sense. I did look around for how to do that, but might need some guidance. Let's revisit in a future PR!

src/testing/vortex/workload.zig
owickstrom: I see your point, but model would be a strange place to keep it. It's really only a concern for the workload itself.

owickstrom: I can, however, lift the `accounts_buffer` into `main`.

src/testing/vortex/workload.zig
owickstrom: Not quite, because the `Command` tag enum isn't the same as `StateMachine.Operation`. I think we'll need to live with the somewhat duplicated code, because it would create a weird coupling if we tried to unify them.

src/testing/vortex/workload.zig
owickstrom: Fixed!

src/testing/vortex/workload.zig
owickstrom: OK, fixed!

src/testing/vortex/workload.zig
owickstrom: Fixed now instead. :)

src/testing/vortex/zig_driver.zig
owickstrom: Fixing this in https://github.com/tigerbeetle/tigerbeetle/pull/2422. Merge into this PR first, or into main afterwards?

src/testing/vortex/zig_driver.zig
owickstrom: Done.

src/testing/vortex/zig_driver.zig
owickstrom: Merged into this PR now.

src/testing/vortex/workload.zig
matklad: Hm, the following works, but, yeah, its a bit on the ugly side:

```zig
fn execute(command: Command, driver: *const DriverStdio) !?Result {
    switch (command) {
        inline else => |events, tag| return execute_regular(
            tag,
            if (tag == .lookup_all_accounts)
                .lookup_accounts
            else
                std.enums.nameCast(StateMachine.Operation, tag),
            events,
            driver,
        ),
    }
}
```

src/testing/vortex/zig_driver.zig
matklad: Looks like it still returns a slice pointing to stack-allocated data, no?

src/testing/vortex/workload.zig
owickstrom: I'll see if I can clean it up.

src/testing/vortex/zig_driver.zig
owickstrom: Argh, not sure what happened here. Fixing tomorrow.

owickstrom: Fixed.

src/testing/vortex/workload.zig
owickstrom: Fixed! Found a nicer way to slice it, avoiding the `execute_regular` helper.

matklad (approved):

Issue #2357, opened 2024-09-27
REPL: Support easier way of specifying full amount for post_pending_transfer


tKe:
It seems the REPL doesn't support (that I can see) a way to specify `amount_max` on when creating a `post_pending_transfer` transfer - the only way to post the full pending amount currently seems to be:
```bash
tb "create_transfers id=24002 pending_id=24000 flags=post_pending_transfer amount=340282366920938463463374607431768211455;"
```

Attempting to use `amount=-0` results in:
> Expected value after equal sign in key-value pair. e.g. `id=1`.

Ideally either `amount=-0` (documented) or `amount=max` should be supported in the REPL, if only for the `amount` field.

tKe:
as an aside this will also impact `exercises/024_two_phase_transfers.sh` in the tigerlings

batiati:
Thanks @tKe!
Supporting `-0` is a great suggestion!

tKe:
I suggested `-0` primarily as it mirrors the tigerbeetle test syntax, however it may be worth considering supporting  `amount=amount_max` for consistency with the client constants as ultimately that's how most consumers will end up thinking of it.

batiati:
Closed by #2358

PR #2358, opened 2024-09-27
Update REPL

batiati:
- Support representing negative inputs as `maxInt - input`, useful for sentinels such as `AMOUNT_MAX`.
- Allow inputs to specify base prefixes `0x`, `0o`, or `0b`, useful for GUIDs and UUIDs.
- Remove the restriction on handling `timestamp`, allowing the REPL to handle `imported` events.
- Add unit tests.

Related: #2357

### DIFF
--- a/src/repl.zig
+++ b/src/repl.zig
@@ -1681,6 +1681,14 @@ test "repl.zig: Handle parsing errors" {
             .in = "create_transfers id=abcd",
             .err = error.BadKeyValuePair,
         },
+        .{
+            .in = "create_transfers amount=0y1234",
+            .err = error.BadKeyValuePair,
+        },
+        .{
+            .in = "create_transfers amount=--0",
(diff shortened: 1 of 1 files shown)


src/repl.zig
```diff
@@ -1018,6 +1055,106 @@ test "repl.zig: Parser single transfer successfully" {
...
+                .timeout = 0,
+                .ledger = 0,
+                .code = 0,
+                .flags = .{},
+                .timestamp = 0,
+            },
+        },
+        .{
+            .in =
+            \\create_transfers amount=0xbee71e
```
matklad, line 1100: 🪲

matklad (dismissed):
Consider adding some negative tests here, like `--92` and `0y90`, but, given this is a REPL, I think we are fine without them.

matklad (approved):

PR #2359, opened 2024-09-27
LSM: Incrementally sort mutable table

sentientwaffle:
### Overview

Previously, we sorted the mutable table updates at the end of each bar, immediately prior to converting into the immutable table. This causes a large latency spike every 64 ops (>250ms).

Instead, spread the sort work across the whole bar. At the end of each beat, sort the suffix of value updates that were just appended during that beat. This spreads out the last beat's work across the bar. The final sort at the end of the last beat merges the sorted runs together -- but many sort algorithms (including zig's "block" sort `std.mem.sort`) optimize the case where the data being sorted is a sequence of sorted sub-arrays, so the final sort is now more efficient.

Also included:
- Move the mutable table sort/deduplication logic from `compaction.zig` to `table_memory.zig`.
- Remove an extra copy of the mutable table data that was happening in `compaction.zig`, and acquire one fewer block per compaction.

### Benchmark

This change doesn't appear to impact overall TPS, but it does cut the transaction p100 by more than half.

<details>
<summary>Before</summary>

```
1222 batches in 29.69 s
transfer batch delay = 0 us
load accepted = 336842 tx/s
batch latency p1 = 5 ms
batch latency p10 = 5 ms
batch latency p20 = 6 ms
batch latency p30 = 10 ms
batch latency p40 = 13 ms
batch latency p50 = 15 ms
batch latency p60 = 20 ms
batch latency p70 = 23 ms
batch latency p80 = 25 ms
batch latency p90 = 27 ms
batch latency p95 = 29 ms
batch latency p99 = 278 ms
batch latency p100 = 282 ms

100 queries in 4.28 s
query latency p1 = 26 ms
query latency p10 = 29 ms
query latency p20 = 30 ms
query latency p30 = 31 ms
query latency p40 = 35 ms
query latency p50 = 38 ms
query latency p60 = 45 ms
query latency p70 = 48 ms
query latency p80 = 53 ms
query latency p90 = 57 ms
query latency p95 = 60 ms
query latency p99 = 105 ms
query latency p100 = 125 ms

rss = 3231879168 bytes

datafile empty = 1141374976 bytes
datafile = 14825820160 bytes
```
</details>

<details>
<summary>After</summary>

```
1222 batches in 29.53 s
transfer batch delay = 0 us
load accepted = 338663 tx/s
batch latency p1 = 7 ms
batch latency p10 = 11 ms
batch latency p20 = 12 ms
batch latency p30 = 15 ms
batch latency p40 = 18 ms
batch latency p50 = 21 ms
batch latency p60 = 25 ms
batch latency p70 = 28 ms
batch latency p80 = 30 ms
batch latency p90 = 32 ms
batch latency p95 = 34 ms
batch latency p99 = 108 ms
batch latency p100 = 113 ms

100 queries in 4.21 s
query latency p1 = 26 ms
query latency p10 = 29 ms
query latency p20 = 30 ms
query latency p30 = 31 ms
query latency p40 = 35 ms
query latency p50 = 40 ms
query latency p60 = 46 ms
query latency p70 = 48 ms
query latency p80 = 51 ms
query latency p90 = 56 ms
query latency p95 = 58 ms
query latency p99 = 61 ms
query latency p100 = 116 ms

rss = 3177619456 bytes

datafile empty = 1141374976 bytes
datafile = 14825820160 bytes
```
</details>

### DIFF
--- a/src/lsm/compaction.zig
+++ b/src/lsm/compaction.zig
@@ -1320,7 +1320,7 @@ pub fn CompactionType(
                 source_exhausted_beat = beat.source_values_processed >= beat.value_count_per_beat;
 
                 log.debug("blip_merge({s}): beat.source_values_processed={} " ++
-                    "beat.value_count_per_beat={}. (source_exhausted_bar={}, " ++
+                    "beat.value_count_per_beat={} (source_exhausted_bar={}, " ++
                     "source_exhausted_beat={})", .{
                     compaction.tree_config.name,
                     beat.source_values_processed,
@@ -1381,15 +1381,18 @@ pub fn CompactionType(
                     bar.source_a_immutable_values.?.len == 0)
                 {
                     if (bar.table_info_a.immutable.len > 0) {
-                        const values = Table.data_block_values(
-                            bar.source_a_immutable_block.?.block,
+                        // Only consume one block at a time so that `blip_merge` never goes over its
+                        // target by more than 1 value block.
+                        const filled = @min(
+                            Table.data.value_count_max,
+                            bar.table_info_a.immutable.len,
                         );
-                        const immutable_remaining_before_fill = bar.table_info_a.immutable.len;
-                        const filled = compaction.fill_immutable_values(values);
-                        bar.source_a_values_consumed_for_fill =
-                            immutable_remaining_before_fill - bar.table_info_a.immutable.len;
+
+                        bar.source_a_values_consumed_for_fill = filled;
+                        bar.source_a_immutable_values = bar.table_info_a.immutable[0..filled];
+                        bar.table_info_a.immutable = bar.table_info_a.immutable[filled..];
+
                         updated_fill_count = true;
-                        bar.source_a_immutable_values = values[0..filled];
                         log.debug("set_source_a({s}): refilled immutable block. {} values out, " ++
                             "{} values consumed", .{
                             compaction.tree_config.name,
@@ -1571,36 +1574,6 @@ pub fn CompactionType(
                 }
             }
             return beat.source_b_len_after_set;
-        }
-
-        /// Copies values to `target` from our immutable table input.
-        /// Return the number of values written to the target and updates immutable table slice to
-        /// the non-processed remainder.
-        fn fill_immutable_values(compaction: *Compaction, target: []Value) usize {
-            const bar = &compaction.bar.?;
-
-            var source = bar.table_info_a.immutable;
-            assert(source.len > 0);
-            assert(target.len > 0);
-
-            // TODO Don't copy this, just use directly.
-            const count = @min(target.len, source.len);
-            assert(count > 0);
-
-            stdx.copy_disjoint(.exact, Value, target[0..count], source[0..count]);
-            bar.table_info_a.immutable = bar.table_info_a.immutable[count..];
-
-            if (constants.verify) {
-                // The immutable table's keys must be strictly increasing.
-                // (A table with a single value is always strictly increasing.)
-                for (
-                    target[0 .. count - 1],
-                    target[1 .. count],
-                ) |*value, *value_next| {
-                    assert(key_from_value(value_next) > key_from_value(value));
-                }
-            }
-            return count;
         }
 
         fn check_and_finish_blocks(compaction: *Compaction, force_flush: bool) enum {
--- a/src/lsm/compaction.zig
+++ b/src/lsm/compaction.zig
@@ -1384,11 +1384,9 @@ pub fn CompactionType(
                         bar.table_info_a.immutable = bar.table_info_a.immutable[filled..];
 
                         updated_fill_count = true;
-                        log.debug("set_source_a({s}): refilled immutable block. {} values out, " ++
-                            "{} values consumed", .{
+                        log.debug("set_source_a({s}): refilled immutable values (filled={})", .{
                             compaction.tree_config.name,
                             filled,
-                            bar.source_a_values_consumed_for_fill,
                         });
                     }
                 }
--- a/src/lsm/compaction.zig
+++ b/src/lsm/compaction.zig
@@ -1573,8 +1573,7 @@ pub fn CompactionType(
             return beat.source_b_len_after_set;
         }
 
-        /// Copies values to `target` from our immutable table input. In the process, merge values
-        /// with identical keys (last one wins) and collapse tombstones for secondary indexes.
+        /// Copies values to `target` from our immutable table input.
         /// Return the number of values written to the target and updates immutable table slice to
         /// the non-processed remainder.
         fn fill_immutable_values(compaction: *Compaction, target: []Value) usize {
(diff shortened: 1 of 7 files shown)


src/lsm/table_memory.zig
```diff
@@ -136,23 +148,102 @@ pub fn TableMemoryType(comptime Table: type) type {
...
+                    if (Table.usage == .secondary_index) {
+                        // Secondary index optimization --- cancel out put and remove.
+                        // NB: while this prevents redundant tombstones from getting to disk, we
+                        // still spend some extra CPU work to sort the entries in memory. Ideally,
+                        // we annihilate tombstones immediately, before sorting, but that's tricky
+                        // to do with scopes.
+                        assert(Table.tombstone(&table.values[source_index]) !=
+                            Table.tombstone(&table.values[source_index + 1]));
+                        source_index += 2;
+                        target_index += 0;
```
matklad, line 209: We now run this logic only for chunks. That is, if there's an in-del pair that strides a chunk boundary, we won't collapse it. I _think_ it is correct, but could you add an assert that the value at `offset-1` is not collapsible

src/lsm/table_memory.zig
```diff
@@ -136,23 +148,102 @@ pub fn TableMemoryType(comptime Table: type) type {
...
+                sort_values_by_key_in_ascending_order,
+            );
+
+            // Merge values with identical keys (last one wins) and collapse tombstones for
+            // secondary indexes.
+            const source_count: u32 = table.count();
+            var source_index: u32 = offset;
+            var target_index: u32 = offset;
+            while (source_index < source_count) {
+                table.values[target_index] = table.values[source_index];
```
matklad: This reassignment bugs me a bit, for the case where target_index == source_index. I don't know if the semantics here is of `memcpy` or of `copyForward`?

Not sure if an extra `if source_index == target_index` is warranted here.

But maybe we can do better than an if? WDYT about adding fast-path for finding the _first_ index we need to skip? So that, by the time you enter this loop, you already have `source_index < target_index`? 

```
const first_mismatch = for(source_index..source_count) {
   // just read array elements here
} orelse return; // fast path, nothing to collapse

const source_index = first_mismatch + 1;
const target_index = first_mismatch;
assert(target_index < source_index);
```

For cases where there's nothing to collapse (eg, object trees), I _think_ the fast path could actually be faster, as it won't memcpy stuff around.

src/lsm/table_memory.zig
cb22: > We now run this logic only for chunks.

It runs for chunks initially, but then `make_immutable()` calls `.sort()` which in turn calls `sort_suffix_from_offset(0);`.

src/lsm/table_memory.zig
```diff
@@ -136,23 +148,102 @@ pub fn TableMemoryType(comptime Table: type) type {
             table.* = .{
                 .values = table.values,
                 .value_context = .{},
-                .mutability = .mutable,
+                .mutability = .{ .mutable = .{} },
                 .name = table.name,
             };
         }
 
         pub fn sort(table: *TableMemory) void {
```
cb22, line 154: There's now an out of date comment in `make_immutable()` about not being incremental :smile:.

src/lsm/forest.zig
```diff
@@ -516,6 +516,11 @@ pub fn ForestType(comptime _Storage: type, comptime groove_cfg: anytype) type {
                     forest.compaction_pipeline.block_pool_raw.len);
             }
 
+            // Groove sync compaction - must be done after all async work for the beat completes???
```
cb22:
```suggestion
            // Groove sync compaction - must be done after all async work for the beat completes.
```
(I'm not sure what I was so inquisitive about here...)

src/lsm/table_memory.zig
sentientwaffle: For simplicity (to avoid duplicating tombstone coalescing logic) I instead went with:

```zig
                const value = table.values[source_index];
                table.values[target_index] = value;
```

The `first_mismatch` approach wouldn't help performance for the immutable object tree case b/c that case never even calls `sort_suffix_from_offset()`. Instead we track that the update are all already sorted:

```zig
        pub fn sort(table: *TableMemory) void {
            assert(table.mutability == .mutable);

            if (!table.value_context.sorted) {
                table.sort_suffix_from_offset(0);
                table.value_context.sorted = true;
            }
        }
```
and
```zig
        pub fn put(table: *TableMemory, value: *const Value) void {
            assert(table.mutability == .mutable);
            assert(table.value_context.count < table.values.len);
            if (table.value_context.sorted) {
                table.value_context.sorted = table.value_context.count == 0 or
                    key_from_value(&table.values[table.value_context.count - 1]) <
                    key_from_value(value);
            } else {
                assert(table.value_context.count > 0);
            }

            table.values[table.value_context.count] = value.*;
            table.value_context.count += 1;
        }
```

src/lsm/compaction.zig
```diff
@@ -337,7 +337,6 @@ pub fn CompactionType(
             /// * It uses an iterator interface, as opposed to raw blocks like the rest.
             /// * It is responsible for keeping track of its own position, across beats.
             /// * It encompasses all possible values, so we don't need to worry about reading more.
-            source_a_immutable_block: ?*Helpers.CompactionBlock = null,
```
cb22, line 340: This is great! Over the forest, since this was a bar scoped block, this saves 512KB * 26 trees * 4 levels = 52MB!

cb22:
Hm, seems like the in-place upgrade integration test is failing here.

cb22:
I haven't dug into it too much further, but I can confirm it passes fine on main, and hangs on this branch (rebased on top of main) on my laptop.

sentientwaffle:
The integration test failure is also on main; see https://github.com/tigerbeetle/tigerbeetle/pull/2369 and https://github.com/tigerbeetle/tigerbeetle/pull/2370.

There is also a `fuzz_lsm_tree` fuzzer failure that the CFO found, but that seed also fails on `main`, so it shouldn't block this PR:

```
$ ./zig/zig build -Drelease fuzz -- lsm_tree 2665580190874383179
info(fuzz): Fuzz seed = 2665580190874383179
info(lsm_tree_fuzz): table_usage=lsm.table.TableUsage.secondary_index
info(lsm_tree_fuzz): fuzz_op_count = 9320872
info(lsm_tree_fuzz): fuzz_op_distribution = enums.EnumFieldStruct(@typeInfo(lsm.tree_fuzz.FuzzOp).Union.tag_type.?,f64,null){ .compact = 1.00e0, .put = 8.00e0, .remove = 0.00e0, .get = 0.00e0, .scan = 0.00e0 }
info(lsm_tree_fuzz): puts_since_compact_max = 120
info(lsm_tree_fuzz): compacts_per_checkpoint = 8
thread 9048 panic: reached unreachable code
/var/home/djg/C/bin/zig-linux-x86_64-0.13.0/lib/std/debug.zig:412:14: 0x117f2c3 in assert_level_table_counts (fuzz)
             ^
/var/home/djg/C/t/db/review/src/lsm/tree_fuzz.zig:397:60: 0x10cb30e in compact (fuzz)
                env.tree.manifest.assert_level_table_counts();
                                                           ^
/var/home/djg/C/t/db/review/src/lsm/tree_fuzz.zig:544:36: 0x10cc7a1 in open_then_apply (fuzz)
                        env.compact(c.op);
                                   ^
/var/home/djg/C/t/db/review/src/lsm/tree_fuzz.zig:227:36: 0x10d012b in main (fuzz)
            try env.open_then_apply(fuzz_ops);
                                   ^
/var/home/djg/C/t/db/review/src/fuzz_tests.zig:122:75: 0x10369b2 in posixCallMainAndExit (fuzz)
        inline else => |fuzzer| try @field(Fuzzers, @tagName(fuzzer)).main(
                                                                          ^
/var/home/djg/C/bin/zig-linux-x86_64-0.13.0/lib/std/start.zig:266:5: 0x10350b1 in _start (fuzz)
    ^
???:?:?: 0x2 in ??? (???)
Unwind information for `???:0x2` was not available, trace may be incomplete
```

cb22 (approved):
:rocket:

PR #2360, opened 2024-09-28
lsm: refactor compaction

matklad:
Preface: 

I've been hacking on this code for a past couple of days, and, while I am not yet bold enough to hit `./zig/zig build check` button, I am at the point where I desperately need feedback!

I _think_ the overall shape of code should be somewhat clear at this point, even if it is missing a detail or two (I see that I've removed `release_table_blocks`, I probably should be calling that _somewhere_).

Right now, I am at the stage where I am pretty convinced in this new approach, but that's probably simply because I've been staring at this code for quite some time and it makes _too_ much sense to me. So curious WDYT here!

---

Refactor compaction to:
- make the code easier to follow,
- make it less unobvious how the compactions are actually schedules (that is, what work can be in-flight at any given point in time),
- make scheduling a bit more aggressive inside a single compaction (e.g. start merge when the first two value blocks are in, separate IO and block resources),
- make it easier to run several compactions at the same time.

Code Changes
============

* Fewer types
* All tricky logic is centralized in `compaction_dispatch`
* Many FIFOs are replaced with fewer RingBuffers
* Flatten Beat and Bar into compaction

CompactionPipeline no longer implements pipelining. Now, it only deals with scheduling compactions themselves. Pipelining is the job of each individual compaction. That is, CompactonPipeline kicks off a particualar compaciton, and recieves the control back only when that compaciton's beat is done.

CompactionPipeline should be rename to CompactionSchedule, or maybe just merged with a forest.

A new type, CompactionPool, is introduced. This is a bag of all resources shared by compactions. Previously, blocks, reads and writes were lumped together. Now, CompactionPool holds separate IOPS for reads, writes, blocks, and CPUS.

CompactionPool is a forest-level object, but it _doesn't_ depend on the type of the forest. So, all compactions share the same CompactionPool _type_, which allows them to share the actual object.

The actual pipelining logic is moved to `Compaction.compaction_dispatch` function. This function no longer tries to _explicitly_ encode a pipeline. Rather, it is just a loop that spawns a bunch of jobs, and the pipelining happens as a side-effect of that. The pseudo-code for compaction_dispatch is:

    for each job:
        if have resources for the job AND
           job's dependencies are fulfilled AND
           it makes sense to start the job:

           start the job

Here, the jobs are reading the index blocks, reading the value blocks, merging/copying, writing the data blocks, writing the index blocks. Resources are blocks (memory) and grid reads&writes. Dependencies are blocks read previously by other jobs (reading a value block needs an index block). The "makes sense" part coves:
- spreading resources across jobs to avoid deadlocks,
- avoiding overreads when finishing a beat-worth of work.

There _isn't_ some kind of comptime meta programming framework for defining jobs, it's just a good old pile of ifs (pushed up to the `compaction_dispatch` cental location).

Dependency tracking is done via RingBuffers. For example, there's a RingBuffer of level_b value blocks. Blocks are added to this queue in order when we _start_ the corrsponding read. When we finish the read, we flip the status of block (which is in the queue at this point) to `read_done`.

So, when we think about launching a merge, we check that:

- there's a head value block in the queue
- the status of this block is `read_done`

If the merge ends up fully consuming the block, it pops it off the queue and adds back to the shared pool.

This setup is similar to a bunch of fifos we used before, but:

* With ring-buffers, so that it's statically obvious what's the upper bound for the size of the queue.
* Queues are matched to block kind (value block/index block/output block) rather than to it's lifetime stage (pending/reading etc).
* There are just fewer queues (I think, didn't actually count).

Behavior Changes
================

* Separate limits for memory, CPU, and IO resources.
* Aggressive scheduling -- start merge as soon as the first two blocks are in
* Running two compactions just work, given enough resources.
* A different "winding down" approach.

Previously, allocating a block for compaction also allocated a grid read and a grid write. In the new approach, those are allocated separately. Additionally, besides available resources there are static limits on the amount of outstanding work. That is, RingBuffers for blocks are statically sized, so you can't read more than X blocks at the same time even if you have spare blocks.

For simplicity, it is assumed that there's actually enough reads and writes to fill the queues fully, but the approach should work even if you restrict _all_ compactions to just a single read and a single write.

The blocks are distributed in a priority round-robin fashion. That is, we first give block to the table builder, and then alternate between level b and level a. This way:

* We avoid deadlocks: if the number of blocks is larger than a minimum, than the jobs will be scheduled in the right order to avoid running out of blocks completely
* We give equal resources to level a and level b (the approach should be extensible to fair allocation, proportional to level "size")
* The schedule ends up being maximally pipelined (the first four reads are: level_a index, level_b index, level_b value, level_a value, which allows starting the merge ASAP).

I _think_ this schedule allows even for a simplistic strategy for running several compactions: if you have N*min blocks, it would be correct to start N compactions at the same time. At least one of them will be guaranteed to be able to proceed. This scheme is simplistic, because:
- a single compaction can monopolize an unfair share of resources. Compactions are greedy.
- there's no code path for "if compaction A frees resource X, wake up compaction B waiting on X".

Both points should be fixable I think though!

Another difference (I think) happens in the "winding down" phase. That is, in what happens when the compaction consumed the required quota of inputs, but haven't yet produced a full value table of output. This is a tricky situation, because, depending on the tombstones, you might have to read quite a bit more tables still. The worst case here is that you have a full table on level A, a full table on level B, but their merge is an empty set.

In the previous approach, due to explicit pipelining, we'll just loop over:

    loop {
        read fixed number of blocks from level
        run merge
    }

In the new approach, this is handled by `AND it makes sense to start the job` part of the pseudocode --- if we _are_ in the winding-down phase, we just artificially limit the depth of our RingBuffers.

Such overreads are tracked via `values_wasted_count`.

### DIFF
--- a/src/lsm/compaction.zig
+++ b/src/lsm/compaction.zig
@@ -2,7 +2,6 @@
 //!
 //! Each Compaction is paced to run in an arbitrary amount of beats, by the forest.
 //!
-//!
 //! Compaction overview:
 //!
 //! 1. Given:
@@ -41,6 +40,7 @@ const log = std.log.scoped(.compaction);
 const constants = @import("../constants.zig");
 
 const stdx = @import("../stdx.zig");
+const maybe = stdx.maybe;
 const trace = @import("../trace.zig");
 const FIFO = @import("../fifo.zig").FIFO;
 const GridType = @import("../vsr/grid.zig").GridType;
@@ -50,6 +50,8 @@ const allocate_block = @import("../vsr/grid.zig").allocate_block;
 const TableInfoType = @import("manifest.zig").TreeTableInfoType;
 const ManifestType = @import("manifest.zig").ManifestType;
 const schema = @import("schema.zig");
+const RingBuffer = @import("../ring_buffer.zig").RingBuffer;
+const IOPS = @import("../iops.zig").IOPS;
 
 /// The upper-bound count of input tables to a single tree's compaction.
 ///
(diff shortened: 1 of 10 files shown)


matklad:
cc @cb22 @sentientwaffle @batiati

matklad:
It is getting there!

This now basically works, and passes non-trivial amount of VOPR seeds.

I am yet to restore logging&tracing though...

src/lsm/compaction.zig
```diff
@@ -281,253 +277,140 @@ pub fn CompactionType(
...
-        grid: *Grid,
-
-        // Populated by {bar,beat}_setup.
-        bar: ?Bar,
-        beat: ?Beat,
-
-        pub fn init(tree_config: Tree.Config, grid: *Grid, level_b: u8) Compaction {
+        } = .{},
+
+        // Position points at the next vale from the layer that should be feed into the merge
```
```suggestion
        // Position points at the next value from the layer that should be feed into the merge
```

src/lsm/compaction.zig
```diff
@@ -65,180 +67,175 @@ pub const compaction_tables_output_max = compaction_tables_input_max;
...
+                pool.blocks.push(block);
             }
+        }
 
-            pub fn ready_to_ioing(self: *BlockFIFO) ?*CompactionBlock {
-                const value = self.ready.pop() orelse return null;
-                assert(value.stage == .ready);
-                value.stage = .ioing;
-                self.ioing.push(value);
+        pub fn idle(pool: *ResourcePool) bool {
```
sentientwaffle, line 220: When `idle` is true, can we assert that `pool.blocks.count == pool.blocks_backing_storage.len`?
(We should assert it between beats, to make sure that we don't leak blocks.)

src/lsm/compaction.zig
```diff
@@ -751,1373 +671,1140 @@ pub fn CompactionType(
...
-            assert(bar.table_builder.value_count < Table.layout.block_value_count_max);
-
-            var values_out_index = bar.table_builder.value_count;
-
+        const CopyDropTombstonesResult = struct {
+            consumed: u32,
+            dropped: u32,
+            produced: u32,
+        };
+        /// Copy values from vales_in to values_out, dropping tombstones as we go.
```
```suggestion
        /// Copy values from values_in to values_out, dropping tombstones as we go.
```

src/lsm/compaction.zig
```diff
@@ -751,1373 +671,1140 @@ pub fn CompactionType(
...
 
-            // Copy variables locally to ensure a tight loop - TODO: Actually benchmark this.
-            const source_local = if (source == .a)
-                beat.source_a_values.?
-            else
-                beat.source_b_values.?;
-            const values_out = bar.table_builder.data_block_values();
-            const values_out_index = bar.table_builder.value_count;
+        // The three functions below are hot CPU loops doing the actual merging, TigerBeetle's data
+        // plane. To reduce the probability of the optimizer getting confused over pointes, don't
```
```suggestion
        // plane. To reduce the probability of the optimizer getting confused over pointers, don't
```

sentientwaffle (commented):
Just a few nits while reading through `compaction.zig`.

src/lsm/compaction.zig
matklad: That's what idle used to do, before I "realized" (vopr) that this is wrong! The output index blocks survive between the beats. So `.idle` _only_ assets "IO", and between the bars we separately assert `blocks_acquired() == 0`

sentientwaffle: Ah, good point -- so we can only assert it between bars!

src/lsm/compaction.zig
```diff
@@ -281,1843 +302,1607 @@ pub fn CompactionType(
...
-
-                .bar = null,
-                .beat = null,
             };
         }
 
-        pub fn deinit(compaction: *Compaction) void {
-            _ = compaction;
-            // TODO: Assert things here - compaction doesn't own anything that needs to be
-            // deallocated.
```
matklad, line 537: vopr calls deinit on crash, so there isn't anything useful we can assert here.

src/lsm/compaction.zig
```diff
@@ -281,1843 +302,1607 @@ pub fn CompactionType(
...
             // level_a comes from the immutable table. This means that blip_read will be a partial,
             // no-op, and that the minimum input blocks are lowered by one.
             if (compaction.level_b == 0) {
                 // Do not start compaction if the immutable table does not require compaction.
-                if (tree.table_immutable.mutability.immutable.flushed) {
-                    return null;
+                if (compaction.tree.table_immutable.mutability.immutable.flushed) {
+                    assert(compaction.quotas.bar == 0);
+                    assert(compaction.quotas.bar_done());
+                    log.debug("{s}:{}: bar_commence: immutable table flushed", .{
```
matklad, line 538: I am using `$name:$level: $method` naming scheme for logs

src/trace.zig
```diff
@@ -106,9 +106,8 @@ const trace_span_size_max = 1024;
 pub const Event = union(enum) {
     replica_commit,
 
-    compact_blip_read,
-    compact_blip_merge,
```
matklad, line 110: It's not clear how to trace read and writes --- they aren't really spans, it's more like we have `n` concurrent IOPS in flight at any given moment.

src/vsr/replica_test.zig
```diff
@@ -59,6 +59,15 @@ comptime {
...
 }
 
+test "Cluster: smoke" {
+    const t = try TestContext.init(.{ .replica_count = 1 });
+    defer t.deinit();
+
+    var c = t.clients(0, t.cluster.clients.len);
+    try c.request(checkpoint_2_trigger, checkpoint_2_trigger);
+    try expectEqual(t.replica(.R_).commit(), checkpoint_2_trigger);
+}
```
matklad, line 69: I personally love to leave such smoke tests in, as they make it easier to debug stuff during larger refactors!

src/vsr/replica.zig
```diff
@@ -7463,9 +7463,7 @@ pub fn ReplicaType(
                 }
 
                 // Case 2: We have a _different_ prepare — safe to nack even if it is faulty.
-                if (journal_header != null and vsr.Headers.dvc_header_type(header) == .valid and
-                    journal_header.?.checksum != header.checksum)
-                {
+                if (journal_header != null and journal_header.?.checksum != header.checksum) {
```
matklad: That's a revert of #2385 which fixes correctness issue at the cost of adding a liveness one. Will back out before merging.

matklad:
Was pretty green before the last round of cleanups, 

Screenshot 2024-10-08 at 21 30 08

the benchmark results are not terrible (though I haven't properly benchmarked this yet), and I _think_ I did everything I wanted to in the scope of this PR!

Ready for review!

src/vsr/replica.zig
chaitanyabhandari: And the liveness issue is being addressed in https://github.com/tigerbeetle/tigerbeetle/pull/2399!

src/lsm/forest.zig
```diff
@@ -756,635 +738,169 @@ pub fn ForestType(comptime _Storage: type, comptime groove_cfg: anytype) type {
...
+        pool: ResourcePool,
         next_tick: Grid.NextTick = undefined,
-
         callback: ?*const fn (*Forest) void = null,
 
+        const CompactionSchedule = @This();
+        const ResourcePool = ResourcePoolType(Grid);
+
+        // The minimum number of blocks required to complete all compactions. This is the sum of
+        // blocks reserved between the bars plus enough blocks to complete a single bar.
```
sentientwaffle: This comment doesn't refer to `forest_compactions_concurrent_max`. (I think it is referring to `forest_block_count_min`.)

src/lsm/forest.zig
```diff
@@ -756,635 +738,169 @@ pub fn ForestType(comptime _Storage: type, comptime groove_cfg: anytype) type {
...
-            {
-                if (first_beat) {
-                    assert(self.compactions.count() == 0);
-                }
+            const last_beat = compaction_beat == constants.lsm_compaction_ops - 1;
+            const last_half_beat = compaction_beat == half_bar - 1;
 
-                // Iterate by levels first, then trees, as we expect similar levels to have similar
-                // time-of-death for writes. This helps internal SSD GC.
+            if (first_beat or half_beat) {
```
sentientwaffle, line 817: Can we assert `pool.blocks_acquired() == 0` here?

src/lsm/forest.zig
```diff
@@ -428,16 +427,7 @@ pub fn ForestType(comptime _Storage: type, comptime groove_cfg: anytype) type {
...
-            // bar.
-            assert(op >= constants.lsm_compaction_ops or
-                forest.compaction_pipeline.compactions.count() == 0);
-
-            forest.compaction_pipeline.beat(op, compact_trees_callback);
-            if (forest.grid.superblock.working.vsr_state.op_compacted(op)) {
-                assert(forest.compaction_pipeline.compactions.count() == 0);
-                assert(forest.compaction_pipeline.bar_active.count() == 0);
-            }
+            forest.compaction_schedule.beat_start(op, compact_trees_callback);
```
sentientwaffle: Nit, but `manifest_log.compact()` takes `(callback, op)`, so maybe swap the order here to match?

src/lsm/forest.zig
```diff
@@ -756,635 +738,169 @@ pub fn ForestType(comptime _Storage: type, comptime groove_cfg: anytype) type {
...
+            (compaction_block_count_beat_min - compaction_block_count_bar_max);
+
         pub fn init(
-            self: *CompactionPipeline,
+            self: *CompactionSchedule,
             allocator: mem.Allocator,
             grid: *Grid,
             forest: *Forest,
             block_count: u32,
         ) !void {
```
sentientwaffle, line 768: Out of scope, but right now VOPR randomizes the number of compaction blocks, but uses the same value for all members of the cluster:

```
            .accounting => .{
                .batch_size_limit = batch_size_limit,
                .lsm_forest_compaction_block_count = random.uintAtMost(u32, 256) +
                    StateMachine.Forest.Options.compaction_block_count_min,
                .lsm_forest_node_count = 4096,
                .cache_entries_accounts = 256,
                .cache_entries_transfers = 256,
                .cache_entries_posted = 256,
                .cache_entries_account_balances = 256,
            },
```

If we instead randomized that per-replica, the storage checker would verify that the output of compaction is independent of the compaction block count.

src/lsm/forest.zig
```diff
@@ -756,635 +738,169 @@ pub fn ForestType(comptime _Storage: type, comptime groove_cfg: anytype) type {
...
-            // way. Maybe move to a union enum for state.
-            for (self.slots) |slot| assert(slot == null);
-            assert(self.callback == null);
+            self.beat_resume();
+        }
 
-            for (self.compactions.slice(), 0..) |*compaction, i| {
-                if (!self.bar_active.isSet(i)) continue;
-                if (compaction.move_table) continue;
+        fn beat_resume(self: *CompactionSchedule) void {
```
sentientwaffle, line 848:
```suggestion
        fn beat_resume(self: *CompactionSchedule) void {
            assert(self.callback != null);

```

src/lsm/compaction.zig
```diff
@@ -63,182 +65,202 @@ pub const compaction_tables_input_max = 1 + constants.lsm_growth_factor;
...
+                // Next stage is either free or loops back to read_value_block_done.
+                merge,
+
+                // Block is in the write queue. Goes directly to free from here.
+                write_value_block,
+                write_index_block,
+            },
+            // The index block is freed immediately after the read for the last value block is
+            // submitted so, by the time merge consumes a value block, the index block is gone
+            // already. Use this field to mark the spots when the compaciton position should advance
```
```suggestion
            // already. Use this field to mark the spots when the compaction position should advance
```

... But "compaciton" is a perfect/terrible name for the smallest possible unit of compaction. :laughing:

src/lsm/compaction.zig
```diff
@@ -281,1843 +302,1617 @@ pub fn CompactionType(
...
+        /// - check if compaction is needed at all (if the level_a is full),
+        /// - find a table on level_a and the corresponding range on level_b that should be
+        ///   compacted,
+        /// - compute the bar quota (just the total number of values in all input tables),
+        /// - execute move table optimization if range_b turns out to be empty.
+        pub fn bar_commence(compaction: *Compaction, op: u64) void {
+            assert(compaction.idle());
+            assert(compaction.stage == .inactive);
+            compaction.stage = .paused;
+            compaction.op_min = compaction_op_min(op);
```
sentientwaffle: Since this is being called only at the beginning of each (half-)bar, I think we can assert:
```
assert(op == compaction_op_min(op));
```

src/lsm/compaction.zig
```diff
@@ -281,1843 +302,1617 @@ pub fn CompactionType(
...
+        // When a block is read from disk, it is added to the tail of the corresponding queue. When
+        // the head block from both level a and level b queues is in the .read_value_block_done
+        // state, the two blocks are popped off the queues and passed down to the merge. At this
+        // point, any number of the blocks still in the queues can continue their read operations.
+        //
+        // For index blocks, queues of length one are used. Because an index block is freed as soon
+        // as the read for the last value block is scheduled, the pipeline should not dry out even
+        // when switching between the tables.
+        //
+        // Note that level_L_position field tracks the logical, deterministic progression of
```
```suggestion
        // Note that level_L_position fields track the logical, deterministic progression of
```

Or maybe

```suggestion
        // Note that level_{a,b}_position fields track the logical, deterministic progression of
```

src/lsm/compaction.zig
```diff
@@ -281,1843 +302,1617 @@ pub fn CompactionType(
...
 
-                    .move_table = range_b.tables.empty(),
-                    .table_info_a = .{ .disk = table_range.table_a },
-                    .range_b = range_b,
-                    .drop_tombstones = tree.manifest.compaction_must_drop_tombstones(
-                        compaction.level_b,
-                        range_b,
-                    ),
+            // Append the entries to the manifest update queue here and now if we're doing
+            // move table. They'll be applied later by bar_apply_to_manifest.
```
```suggestion
            // move table. They'll be applied later by bar_complete().
```

src/lsm/compaction.zig
```diff
@@ -281,1843 +302,1617 @@ pub fn CompactionType(
...
-                    compaction.bar.?.source_values_read_count =
-                        compaction.bar.?.compaction_tables_value_count;
-                    compaction.bar.?.source_values_merge_count =
-                        compaction.bar.?.compaction_tables_value_count;
-                    compaction.bar.?.target_values_merge_count =
-                        compaction.bar.?.compaction_tables_value_count;
-                    compaction.bar.?.target_values_write_count =
-                        compaction.bar.?.compaction_tables_value_count;
+            if (compaction.table_info_a == null) {
+                assert(compaction.range_b == null);
```
sentientwaffle, line 663:
```suggestion
                assert(compaction.range_b == null);
                assert(compaction.manifest_entries.count == 0);
```

src/lsm/compaction.zig
```diff
@@ -281,1843 +302,1617 @@ pub fn CompactionType(
...
-            compaction: *Compaction,
-            beats_max: u64,
-            target_index_blocks: Helpers.BlockFIFO,
-        ) void {
-            // Limited to half bars for now.
-            assert(beats_max <= @divExact(constants.lsm_compaction_ops, 2));
-            assert(beats_max > 0);
+            // Mark the immutable table as flushed, if we were compacting into level 0.
+            if (compaction.level_b == 0) {
+                compaction.tree.table_immutable.mutability.immutable.flushed = true;
```
sentientwaffle, line 703:
```suggestion
                assert(!compaction.tree.table_immutable.mutability.immutable.flushed);
                compaction.tree.table_immutable.mutability.immutable.flushed = true;
```

src/lsm/compaction.zig
```diff
@@ -281,1843 +302,1617 @@ pub fn CompactionType(
...
+            }
+
+            if (compaction.quotas.beat_done()) {
+                assert(compaction.stage == .paused);
+                log.debug("{s}:{}: beat_commence: quota fulfilled", .{
+                    compaction.tree.config.name,
+                    compaction.level_b,
+                });
+                return;
+            }
```
sentientwaffle, line 814:
```suggestion
            }
            assert(!compaction.move_table);
```

src/lsm/compaction.zig
```diff
@@ -281,1843 +302,1617 @@ pub fn CompactionType(
...
-                assert(blocks.source_value_blocks[1].ready.count == 0);
-                // table_b's release_table_blocks gets called in update_position_b.
-                // TODO: Maybe we should move table_a's there too...?
             }
-
-            beat.deactivate_assert_and_callback(.merge, .{
-                .bar = source_exhausted_bar,
-                .beat = source_exhausted_bar or source_exhausted_beat,
-            });
+            assert(!progressed);
```
sentientwaffle, line 1190: So since the safety counter has a `+1`, if the safety counter is exhausted this `!progressed` assert is guaranteed to succeed?

src/lsm/forest.zig
matklad: We've decided that callbacks go last:

>- Callbacks go last in the list of parameters. This mirrors control flow: callbacks are also
  _invoked_ last.

I will actually swap it here for consistency, and then send a follow up to correct all the callbacks I could find, as per our zero tech-debt rule which we are violating here (had we fixed callbacks before, we wouldn't have to exchange the comments here (TigerStyle is _so_ good, we should try to follow it better :-) )).

src/lsm/compaction.zig
matklad: Ah, this op/op_min stuff did look odd to me, but I haven't looked deeper, you are of course correct! 

We probably should _also_ update the diagram above `compaction_op_min`, and the rest of the docs, but I've postponed docs refreshing for now.

src/lsm/forest.zig
sentientwaffle: > go last in the list of parameters

Oops; I thought this was the other way around in tigerstyle!

src/lsm/compaction.zig
```diff
@@ -281,1843 +302,1623 @@ pub fn CompactionType(
...
+            // index block can get re-read across the bar, the same table can be released twice.
+            if (compaction.grid.free_set.is_released(index_block_address)) {
+                for (data_block_addresses) |address| {
+                    assert(compaction.grid.free_set.is_released(address));
+                }
+            } else {
+                for (data_block_addresses) |address| {
+                    compaction.grid.free_set.release(address);
+                }
+                compaction.grid.free_set.release(index_block_address);
```
sentientwaffle: Ah, so we use `free_set.release()` instead of `grid.release()` b/c we don't want to demote the blocks from the cache, as they may be re-read again during the next beat(s).

Let's assert in `bar_commence()` for each (level A and B) index block address: 

```
assert(!compaction.grid.free_set.is_released(address));
```

src/lsm/compaction.zig
```diff
@@ -281,1843 +302,1623 @@ pub fn CompactionType(
...
+                } else {
+                    break :values null;
+                }
+            };
+            assert(!(level_a_values_used == null and level_b_values_used == null));
+
+            const level_a_values = if (level_a_values_used) |values_used| values: {
+                const values_remaining = values_used[compaction.level_a_position.value..];
+                // Only consume one block at a time so that `blip_merge` never goes over its
+                // target by more than 1 value block.
```
sentientwaffle: `blip_merge` is gone:

```suggestion
                // Only consume one block at a time so that a beat never outputs past its quota
                // by more than 1 value block.
```

src/lsm/compaction.zig
matklad: Ah, good catch, I am missing `else unreachable` here! The `assert(!progressed)` here is mostly just for readability, it is an `endwhile` of sorts that reminds the reader, by the time they get to this line, that there's this whole `for` loop thing going on, which started 200 lines before.

I am also torn on what we should do if the safety counter is exhausted: crash with `unreachable`, or `log.error` and soldier on? In this case, the second option would be correct. But I figured that crashing is probably better as a principle.

src/lsm/compaction.zig
matklad: >Ah, so we use free_set.release() instead of grid.release() b/c we don't want to demote the blocks from the cache, as they may be re-read again during the next beat(s).

No, that is a genuine refactoring bug! I didn't intend to fix this.

That being said, this feels a bit backwards --- it seems that, rather than demoting tables after the fact, the grid should perhaps read them without bumping the cache in the first place? 

>Let's assert in bar_commence() for each (level A and B) index block address:

Ah, this is great!

src/lsm/compaction.zig
```diff
@@ -281,1843 +302,1660 @@ pub fn CompactionType(
...
+            read.compaction = compaction;
+            compaction.grid.read_block(
+                .{ .from_local_or_global_storage = read_value_block_callback },
+                &read.grid_read,
+                value_block_address,
+                value_block_checksum.value,
+                .{ .cache_read = true, .cache_write = true },
+            );
+            const index_block_address = Table.block_address(index_block.ptr);
+            if (index_block_address == 74) {
```
sentientwaffle: FIXME :smile: 
(line 1422 also)

src/lsm/compaction.zig
```diff
@@ -281,1843 +302,1660 @@ pub fn CompactionType(
...
-                        value_block_addresses[i],
-                        value_block_checksums[i].value,
-                        .{ .cache_read = true, .cache_write = true },
-                    );
-
-                    read.pending_reads_data += 1;
-                    i += 1;
+                // Read level A value block.
+                if (compaction.table_info_a.? == .immutable) {
+                    // The whole table is in memory, no need to read anything.
```
sentientwaffle, line 1065:
```suggestion
                    // The whole table is in memory, no need to read anything.
                    assert(compaction.level_a_index_block.count == 0);
```

(I found this assert useful when considering the "priority" calculations for reading value blocks.)

src/lsm/compaction.zig
matklad: oh dear, I guess one more tidy check is coming soon!

matklad:
Rebased to pick up the VOPR fix for "all clients evicted" issue

src/lsm/compaction.zig
```diff
@@ -281,1843 +302,1648 @@ pub fn CompactionType(
...
+                => unreachable,
+            }
 
-        /// Perform read IO to fill our source_index_block_{a,b} and source_value_blocks with as
-        /// many blocks as we can, given their sizes, and where we are in the amount of work we
-        /// need to do this beat.
-        pub fn blip_read(compaction: *Compaction, callback: BlipCallback, ptr: *anyopaque) void {
-            log.debug("blip_read({s}): scheduling read IO", .{compaction.tree_config.name});
+            // The loop below runs while (!progressed) and, every time progressed is set to true,
+            // one of the safety_counter resources is acquired.
```
sentientwaffle: Isn't it actually

```suggestion
            // The loop below runs while (progressed) and, every time progressed is set to true,
            // one of the safety_counter resources is acquired.
```

Or it could be

```suggestion
            // The loop below runs until (!progressed) and, every time progressed is set to true,
            // one of the safety_counter resources is acquired.
```

src/lsm/compaction.zig
```diff
@@ -281,1843 +302,1648 @@ pub fn CompactionType(
...
+            var progressed = true;
+            const safety_counter =
+                compaction.pool.?.reads.total() +
+                compaction.pool.?.writes.total() +
+                compaction.pool.?.cpus.total() + 1;
```
sentientwaffle: Or maybe even
```suggestion
                compaction.pool.?.reads.available() +
                compaction.pool.?.writes.available() +
                compaction.pool.?.cpus.available() + 1;
```

matklad: Hm, I was mentally trying to convince you that `total` is better, but ended up convincing myself that `available` is the way to go 🤣 

I can't put my finger on the exact guideline here, it seems the two things are in tension here:

* we want to be our safety counters to be simple, so it is obvious that they are correct.
* but we also want them to be tight, such that, if they are _not_ correct, the VOPR tells us this. 

Here, total is less tight, but simpler, but I think it's not _that_ simpler than `.available`, and that better captures the reason _why_ this is valid safety counter.

src/lsm/compaction.zig
```diff
@@ -281,1843 +302,1648 @@ pub fn CompactionType(
...
-
-            assert(blocks.source_index_block_a.stage == .standalone);
-            assert(blocks.source_index_block_b.stage == .standalone);
-
-            blocks.source_index_block_a.stage = .free;
-            blocks.source_index_block_b.stage = .free;
-
-            block_pool.push(blocks.source_index_block_a);
-            block_pool.push(blocks.source_index_block_b);
+        fn merge_inputs(compaction: *Compaction) struct { ?[]const Value, ?[]const Value } {
```
```suggestion
        fn merge_inputs(compaction: *const Compaction) struct { ?[]const Value, ?[]const Value } {
```

sentientwaffle (dismissed):
LGTM!

src/lsm/compaction.zig
matklad: Fixed&squashed all review commits down to one!

sentientwaffle (approved):

PR #2362, opened 2024-09-29
vopr: core_missing_prepare detects replicas stuck repairing headers

chaitanyabhandari:
Fixes the failed seed `./zig/zig build -Drelease vopr -- 7036583507627146912` on main ([4c4e536](https://github.com/tigerbeetle/tigerbeetle/commit/4c4e53625ae97b0157c7ddd5c57c75bf7afa781f)). This PR adds `core_missing_header`, which finds the _highest header op_ (since headers are repaired from highest → lowest missing op) that is also missing on all replicas in the core.

### Detailed root cause analysis
In this VOPR seed, R0 and R2 are found stuck in view change wherein:
* R2 (self.op=123, op_checkpoint=119) has a corrupt _prepare_ for op=121
* R0 (self.op=122, op_checkpoint=99) is missing _headers_ from 110 → 121 (it performed state sync to reach op_checkpoint=99)
* R0 can't repair headers beyond op=122 since prepare for op=121 is corrupted on R2
* R2 can't commit past op 120 since prepare for op=121 is corrupted
* Neither R0 nor R2 can commit all the ops in their journal to step up as primary

The issue here is that `core_missing_prepare` looks for the smallest missing _prepare_ (likely assuming that the header chain is intact and the commit pipeline is stuck because of a missing prepare), and checks if it is missing on the core. https://github.com/tigerbeetle/tigerbeetle/blob/4a8eb61f8001cc24f89b665c5af17b7011e9eb74/src/vopr.zig#L690-L691

However, a stuck commit pipeline could also be caused by missing _headers_, as we don't start committing till we've repaired all of our headers up till our head op.
https://github.com/tigerbeetle/tigerbeetle/blob/4a8eb61f8001cc24f89b665c5af17b7011e9eb74/src/vsr/replica.zig#L6442-L6448

### DIFF
--- a/src/vopr.zig
+++ b/src/vopr.zig
@@ -362,6 +362,8 @@ pub fn main() !void {
             stdx.unimplemented("repair requires reachable primary");
         } else if (simulator.core_missing_quorum()) {
             output.warn("no liveness, core replicas cannot view-change", .{});
+        } else if (simulator.core_missing_header()) |header| {
+            output.warn("no liveness, op={} is not available in core", .{header.op});
         } else if (simulator.core_missing_prepare()) |header| {
             output.warn("no liveness, op={} is not available in core", .{header.op});
         } else if (try simulator.core_missing_blocks(allocator)) |blocks| {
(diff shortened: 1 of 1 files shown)


matklad (commented):
The logic looks right to me! I am not quite happy with the code --- missing_header and missing_prepare are almost, but not quite, identical. 

At minimum, we should add a comment explaining the difference between the to (looking for smallest vs largest op). But is there perhaps a way to reduce the amount of duplication here?

chaitanyabhandari:
Mhmmmmmmm yes, not 100% happy with this distinction as well... I was first thinking about merging the `core_missing_header` logic into `core_missing_prepare`, but it started looking a bit convoluted.

Re the comment about why we find the largest header:
https://github.com/tigerbeetle/tigerbeetle/pull/2362/files#diff-ff61239ba80a7cf2991f89b9394ed1d1eeb75198961805f094cf6c11dbd9576aR692, I could add another such comment in _core_missing_prepare_.


Alternatively, I could take another stab at merging the functions wherein we:  
* find _smallest missing prepare_ and the *highest missing header* in the core
* if the *highest missing header* is:
  * non null: we ignore the *smallest missing prepare* and check if the  *highest missing header* is present on the core
  * null: we check if the _smallest missing prepare_ is present on the core
  
Let me also push a commit with the functions merged so we can take a call accordingly!

chaitanyabhandari:
Took a stab at merging the two in https://github.com/tigerbeetle/tigerbeetle/pull/2362/commits/cbf7b34d905f7cf10c5ce7307cb754eb31722034!

matklad (approved):
LGTM now! Thanks for putting in the effort to get a little bit of extra refactoring&compression here!

PR #2363, opened 2024-09-29
vsr: potential primaries updating their SV may have gaps in journal

chaitanyabhandari:
This PR fixes the failed seed `./zig/zig build vopr -- --lite 17467401530694012113` on main ([4c4e536](https://github.com/tigerbeetle/tigerbeetle/commit/4c4e53625ae97b0157c7ddd5c57c75bf7afa781f)). In this seed, a potential primary attempted to update its start_view headers (assuming it had no gaps from commit_min →  self.op) _before_ it finished repair (before invoking `superblock.checkpoint`).

### Solution

We now make do_view_change headers durable for potential primaries, instead of start_view headers. This is because a potential primary performing `superblock.checkpoint` is not guaranteed to have no gaps from commit_min →  self.op.
 
However, to achieve this, we now have to loosen verification while loading view_headers stored in the superblock. This is because we may end up finding _either_ do_view_change/start_view headers in our superblock even when `log_view == view`. 
https://github.com/tigerbeetle/tigerbeetle/blob/4a8eb61f8001cc24f89b665c5af17b7011e9eb74/src/vsr/superblock.zig#L454-L462

### Detailed Backstory

When we introduced the change (https://github.com/tigerbeetle/tigerbeetle/pull/2313) for making view_headers durable during `superblock.checkpoint`, we were under the impression that a potential primary (replica in .view_change that is repairing to become primary) is guaranteed to have no gaps from _commit_min →  self.op_ while creating its start_view headers. 
https://github.com/tigerbeetle/tigerbeetle/blob/4a8eb61f8001cc24f89b665c5af17b7011e9eb74/src/vsr/replica.zig#L4855-L4857

This understanding hinges on the assumption that a potential primary only starts committing (and therefore checkpointing) _after_ repairing its header hash chain up till self.op.
https://github.com/tigerbeetle/tigerbeetle/blob/4a8eb61f8001cc24f89b665c5af17b7011e9eb74/src/vsr/replica.zig#L6442-L6448

While this understanding is _correct_, it is *imprecise*. It does not cater to the case where a replica starts committing the checkpoint_trigger op in a *previous log_view*, but reaches the `checkpoint_superblock` state in a log_view in which it is a potential primary. 
https://github.com/tigerbeetle/tigerbeetle/blob/4a8eb61f8001cc24f89b665c5af17b7011e9eb74/src/vsr/replica.zig#L3628-L3629

### DIFF
--- a/src/vsr.zig
+++ b/src/vsr.zig
@@ -1293,13 +1293,6 @@ const ViewChangeHeadersSlice = struct {
         // A SV never includes gaps or faulty headers.
         assert(Headers.dvc_header_type(head) == .valid);
 
-        if (headers.command == .start_view) {
-            assert(headers.slice.len >= @min(
-                constants.view_change_headers_suffix_max,
-                head.op + 1, // +1 to include the head itself.
-            ));
-        }
-
         var child = head;
         for (headers.slice[1..], 0..) |*header, i| {
             const index = i + 1;
@@ -1318,7 +1311,14 @@ const ViewChangeHeadersSlice = struct {
 
             switch (Headers.dvc_header_type(header)) {
                 .blank => {
-                    assert(headers.command == .do_view_change);
+                    // We can't verify that SV headers contain no gaps headers here:
+                    // superblock.checkpoint could make .do_view_change headers durable instead of
+                    // .start_view headers when view == log_view (see `commit_checkpoint_superblock`
+                    // in `replica.zig`). When these headers are loaded from the superblock on
+                    // startup, they are considered to be .start_view headers (see `vsr_headers` in
+                    // `superblock.zig`).
+                    maybe(headers.command == .do_view_change);
+                    maybe(headers.command == .start_view);
                     continue; // Don't update "child".
                 },
                 .valid => {
--- a/src/vsr/replica.zig
+++ b/src/vsr/replica.zig
@@ -4256,30 +4256,24 @@ pub fn ReplicaType(
                 log.info("{}: sync: done", .{self.replica});
             }
 
-            // Update view_headers to include at least one op from the future checkpoint. This
-            // ensures a replica never starts with its head op less than self.op_checkpoint().
-            // The only exception to this is a replica that arrived at a checkpoint via state
-            // sync.
-            if (self.view == self.log_view) {
-                // Unconditionally convert potential primary's DVC headers → SV headers; the DVC
-                // headers may contain truncated ops which should not be made durable. For all other
-                // cases, update SV headers only if they aren't already up to date.
-                if ((self.status == .view_change and
-                    self.primary_index(self.view) == self.replica) or
-                    self.view_headers.array.get(0).op < self.op_checkpoint_next_trigger())
-                {
-                    self.view_headers.command = .start_view;
+            if (self.status == .view_change and self.view == self.log_view) {
+                // Unconditionally update a potential primary's DVC headers; current headers may
+                // contain truncated ops that must not be made durable. We can't update SV
+                // headers for a potential primary because we could arrive here while the potential
+                // primary is still repairing (and thus may still have gaps in its journal).
+                assert(self.do_view_change_quorum);
+                assert(self.view_headers.command == .do_view_change);
+                self.update_do_view_change_headers();
+            } else {
+                // Update view_headers to include at least one op from the future checkpoint. This
+                // ensures a replica never starts with its head op less than self.op_checkpoint().
+                if (self.view_headers.array.get(0).op < self.op_checkpoint_next_trigger()) {
+                    assert(self.status == .normal);
                     self.update_start_view_headers();
                 }
-                assert(self.view_headers.command == .start_view);
-                assert(self.view_headers.array.get(0).op >= self.op_checkpoint_next_trigger());
-            } else {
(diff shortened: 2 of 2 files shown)


chaitanyabhandari:
This PR in its current form is unsafe. We are currently thinking of safe solutions.

Specifically, if we start committing the checkpoint trigger in a previous log_view, and receive a callback for checkpoint_superblock in the current log_view where we are potential primary, we could have breaks in our journal (since we could still be mid repair). With this solution, we would create a full SV starting from self.commit_min (which is the checkpoint trigger), we would make the SV durable during the checkpoint (excluding the new head that we've received).

Now, if we were to then crash & restart after making the checkpoint durable, we would find ourself in a state where view == log_view, which could make us the canonical log_view for a new view change post startup. This could lead to a safety violation: we don't have the head op that we received before crashing view == log_view in our SV headers, and the head op isn't guaranteed to exist in our journal, we could end up truncating a committed head op during view change.

chaitanyabhandari:
At first glance, the failed VOPR seed (./zig/zig build -Drelease vopr -- 2093263118278902398) is a panic in the VOPR itself, and it _seems_ to be unrelated to this change. Planning on digging deeper to confirm this.

This run has 11 clients, _all_ of which get evicted:
```C
[debug] (simulator): tick_requests: client=8 is evicted, reason=vsr.message_header.Header.Eviction.Reason.session_too_low
[debug] (simulator): tick_requests: client=9 is evicted, reason=vsr.message_header.Header.Eviction.Reason.session_too_low
[debug] (simulator): tick_requests: client=10 is evicted, reason=vsr.message_header.Header.Eviction.Reason.session_too_low
[debug] (simulator): tick_requests: client=0 is evicted, reason=vsr.message_header.Header.Eviction.Reason.session_too_low
[debug] (simulator): tick_requests: client=1 is evicted, reason=vsr.message_header.Header.Eviction.Reason.session_too_low
[debug] (simulator): tick_requests: client=2 is evicted, reason=vsr.message_header.Header.Eviction.Reason.no_session
[debug] (simulator): tick_requests: client=3 is evicted, reason=vsr.message_header.Header.Eviction.Reason.no_session
[debug] (simulator): tick_requests: client=4 is evicted, reason=vsr.message_header.Header.Eviction.Reason.session_too_low
[debug] (simulator): tick_requests: client=5 is evicted, reason=vsr.message_header.Header.Eviction.Reason.session_too_low
[debug] (simulator): tick_requests: client=6 is evicted, reason=vsr.message_header.Header.Eviction.Reason.no_session
[debug] (simulator): tick_requests: client=7 is evicted, reason=vsr.message_header.Header.Eviction.Reason.session_too_low
thread 31907068 panic: reached unreachable code
/Users/chaitanyabhandari/Desktop/tigerbeetle/tigerbeetle/src/vopr.zig:932:17: 0x1041139f7 in tick_requests (vopr)
                unreachable;
                ^
```
The VOPR expects at least one client to be available, so it hits an `unreachable` here:
https://github.com/tigerbeetle/tigerbeetle/blob/4a8eb61f8001cc24f89b665c5af17b7011e9eb74/src/vopr.zig#L917-L929

src/vsr.zig
```diff
@@ -1280,7 +1280,6 @@ const ViewChangeHeadersSlice = struct {
             .command = command,
             .slice = slice,
         };
-        headers.verify();
```
sentientwaffle: Rather than making `verify()` not always safe to call, lets change `verify` to allow gaps in `start_view` headers, and then add a no-gap-verification specifically to the start-view headers that are sent in `start_view` messages.

src/vsr/replica.zig
```diff
@@ -4256,31 +4256,25 @@ pub fn ReplicaType(
...
-                // cases, update SV headers only if they aren't already up to date.
-                if ((self.status == .view_change and
-                    self.primary_index(self.view) == self.replica) or
-                    self.view_headers.array.get(0).op < self.op_checkpoint_next_trigger())
-                {
-                    self.view_headers.command = .start_view;
+            // Unconditionally update a potential primary's DVC headers; current headers may contain
+            // truncated ops which should not be made durable. We can't update SV headers for a
+            // potential primary because we could arrive here while the potential primary is
+            // still repairing (and thus may still have gaps in its journal).
```
sentientwaffle: Let's move this comment into the branch, to clarify that it doesn't apply to the `else` branch.

src/vsr.zig
chaitanyabhandari: Great idea, it is indeed looking much cleaner this way! Sending a commit your way soon after I clean it up.

chaitanyabhandari: Also, turns out we already have the no gaps verification in `start_view_message_headers`, it was a remnant from an old code where I was trying to loosen verifications around SV: https://github.com/tigerbeetle/tigerbeetle/blob/4a8eb61f8001cc24f89b665c5af17b7011e9eb74/src/vsr/replica.zig#L10282-L10287

chaitanyabhandari:
Comments addressed in https://github.com/tigerbeetle/tigerbeetle/pull/2363/commits/ed0a1c73a24b2caa9fbcecf54b52dfbb27ded83d!

sentientwaffle (approved):
LGTM -- a very elegant fix, too!

PR #2365, opened 2024-10-01
ci: try to extract better logging out of upgrade test

matklad:
Before this PR, an upgrade test could succeed, fail immediately, or hang. After this PR, the possible outcomes are a success, or a failure after 10 minutes. That is, we eliminate hangs at the cost of making over failures slower. I think the trade is perhaps worth it!

This is the 100500 instance where we are in need of async process management API, where we want to run a bunch of processes, kill them on timeout, react to processes crashing themselves, and doing general evented IPC.

As usual, we solve this in an ad-hoc way, by combining thread-spawning with hacky APIs and suboptimal behavior.

I am _this_ close to adding process management syscalls to out async IO library and rewriting shell.zig on top of that :-)

### DIFF
--- a/src/integration_tests.zig
+++ b/src/integration_tests.zig
@@ -379,18 +379,31 @@ test "in-place upgrade" {
             const tigerbeetle_load = try context.shell.fmt("{s}/tigerbeetle-load", .{context.tmp});
             try context.install_replica(tigerbeetle_load, .past);
 
-            var load = try context.shell.spawn(.{},
-                \\{tigerbeetle} benchmark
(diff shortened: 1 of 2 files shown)


owickstrom:
Maybe the time is now, @matklad .

sentientwaffle (approved):

matklad:
Let's try merging this, but I guess we have to actually fix the underlying bug first 🤣

sentientwaffle:
This should be mergeable once it is rebased on `main`.

PR #2367, opened 2024-10-02
test: run concurrent workloads with separate ledgers

owickstrom:
Restructures the systest workload (in Java) to run multiple workloads in separate threads, isolated by using unique ledgers. This let's us verify the thread safety claim of the Java client, which is shared across the threads.

The verified properties after querying known accounts are similar to before, because we already grouped accounts by ledger and checked consistency for each group.

### DIFF
--- a/src/testing/systest/workload/src/main/java/Main.java
+++ b/src/testing/systest/workload/src/main/java/Main.java
@@ -2,6 +2,8 @@ import java.security.SecureRandom;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Random;
+import java.util.concurrent.ExecutorCompletionService;
+import java.util.concurrent.Executors;
 import com.tigerbeetle.Client;
 import com.tigerbeetle.UInt128;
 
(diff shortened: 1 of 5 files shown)


src/testing/systest/workload/src/main/java/Main.java
```diff
@@ -17,12 +19,22 @@ public static void main(String[] args) throws Exception {
...
+          "REPLICAS must list at least one address (comma-separated)");
+    }
+
     Random random = new SecureRandom();
     byte[] clusterID = UInt128.asBytes(Long.parseLong(env.getOrDefault("CLUSTER", "1")));
 
     try (var client = new Client(clusterID, replicaAddresses)) {
-      System.err.println("starting workload");
-      new Workload(random, client).run();
+      try (var scope = new StructuredTaskScope.ShutdownOnFailure()) {
```
matklad: StructuredTaskScope? Wow, modern Java is wild!

Will this actually run on our CI thogh?

owickstrom: Yeah, stealing good ideas from Kotlin, I'm guessing. :+1: 

In PR builds, no, not until #2356 is merged. But it will run on `main` since #2346 .

matklad: What I mean is do we use java new enough for this? I see this is some kind of preview feature

owickstrom: Ah. Yeah it worked, but I'm actually surprised why I didn't need any special flag to accept experimental features. I'll look into it. Can solve it with some older API otherwise.

src/testing/systest/workload/src/main/java/Main.java
```diff
@@ -17,12 +19,22 @@ public static void main(String[] args) throws Exception {
...
+        System.err.println("starting %d workload(s)".formatted(workloadCount));
+        IntStream.rangeClosed(1, workloadCount)
+          .forEach(ledger -> scope.fork(new Workload(random, client, ledger)));
+        scope.join();
```
matklad: Do you need join? I'd imagine STS is AutoClosable and would work with try with resources?

matklad (dismissed):

src/testing/systest/workload/src/main/java/Main.java
owickstrom: Looks like I could use [throwIfFailed](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/concurrent/StructuredTaskScope.ShutdownOnFailure.html#throwIfFailed()) instead. The idea is that the first exception thrown by one of the threads is rethrown in main.

owickstrom: Not sure if that's the default behavior for the autoclosable, checking.

owickstrom: Looks like it just "shuts down" by default, not rethrowing exceptions.

src/testing/systest/workload/src/main/java/Main.java
owickstrom: After a lot of messing around, it turns out that it _doesn't_ work. It was hidden by Maven not realizing it needed to recompile things.

matklad (dismissed):

owickstrom:
Minor statistics printing thing added.

Example stderr output:

    starting 32 workload(s)
    285310 requests in total, 37% successful, throughput of 28530 req/s

matklad (approved):

PR #827, opened 2023-05-29
IO: `DSYNC` WAL writes and `fdatasync` the rest on checkpoint. 

kprotty:
We currently open the database file with `O_DSYNC` (along with `O_DIRECT`) which makes all writes ensure the data (not file metadata) is flushed to disk before completing. While this is important for the Write-Ahead-Log (WAL), it may not be necessary for other writes given we have an explicit `checkpoint()` to flush changes to `Storage`.

@cb22 made the observation that we could apply `O_DSYNC` to individual writes using [pwritev2](https://manpages.debian.org/testing/manpages-dev/preadv2.2.en.html) and RWF_DSYNC. `io_uring` supports the RWF flags using the standard `IORING_OP_WRITEV` using `sqe.rw_flags` so it was doable.

This PR makes WAL writes use `DSYNC` for its data integrity and uses `fdatasync` after a checkpoint (but before its callback) for data integrity of other writes. The f(data)sync portion is only implemented for linux ATM as this is currently exploratory work.

## Pre-merge checklist

Performance:

* [x] Compare `zig build benchmark` on linux before and after this pr.
    ``` sh
    # benchmark results before
    1224 batches in 84.66 s
    load offered = 1000000 tx/s
    load accepted = 118113 tx/s
    batch latency p00 = 0 ms
    batch latency p10 = 23 ms
    batch latency p20 = 24 ms
    batch latency p30 = 25 ms
    batch latency p40 = 26 ms
    batch latency p50 = 28 ms
    batch latency p60 = 28 ms
    batch latency p70 = 29 ms
    batch latency p80 = 30 ms
    batch latency p90 = 31 ms
    batch latency p100 = 4305 ms
    transfer latency p00 = 0 ms
    transfer latency p10 = 3644 ms
    transfer latency p20 = 7668 ms
    transfer latency p30 = 14368 ms
    transfer latency p40 = 22154 ms
    transfer latency p50 = 31379 ms
    transfer latency p60 = 40727 ms
    transfer latency p70 = 50935 ms
    transfer latency p80 = 60691 ms
    transfer latency p90 = 68430 ms
    transfer latency p100 = 74678 ms
    
    # benchmark results after
    1225 batches in 57.92 s
    load offered = 1000000 tx/s
    load accepted = 172637 tx/s
    batch latency p00 = 0 ms
    batch latency p10 = 23 ms
    batch latency p20 = 23 ms
    batch latency p30 = 24 ms
    batch latency p40 = 25 ms
[...]

PR #2368, opened 2024-10-02
io/storage: add the ability to do per io dsync, and explicit fsync

cb22:
Specifically, we handle all non-grid writes with DSYNC, much in the same way as they were before. Grid writes are flushed after each beat on compaction with an explicit fsync. (technically we request the entire file to be flushed, and Linux will request the entire NVMe namespace to be flushed...)

This supplants #827 with a few differences:
* Due to the new compaction code, there's a natural place to put a call to fsync: after the work for a beat has finished. This fsync is called there as a barrier, so it happens much more frequently than checkpoint, so we won't have a large periodic spike.
* Darwin support. After adding `F_FULLSYNC` after each write, the performance on macOS suffered quite significantly. This brings it back up, with a roughly 4x (!) improvement from 44k TPS to 186k TPS.

In terms of Linux performance, this takes the benchmark from ~308k to ~365k for 10 million transfers. The gap widens with more compactions - at 50 million transfers, it's ~176k vs ~314k.

One source of this performance is that our compaction queue utilization isn't perfect. By allowing the drive's cache to absorb some of the writes, then flush them, we're cheating around this. Additionally, these numbers are from consumer hardware - it's likely enterprise SSDs with battery backed cache won't see much, if any, change.

However, even with perfect queue depth, there's another hidden advantage: when growing the data file, because the inode size metadata has to be updated and journaled, [Linux falls back to using FLUSH as opposed to FUA](https://www.uwsg.indiana.edu/hypermail/linux/kernel/1904.1/03845.html). Now, this happens once per beat, instead of once per compaction write! 

(an alternative solution to the above is preallocating and growing the data file in chunks).

## Tracing NVMe requests
If you're curious, you can directly trace the ops issued to the NVMe drive on Linux with:

```
echo 1 > /sys/kernel/debug/tracing/events/nvme/enable
cat /sys/kernel/debug/tracing/trace_pipe  | grep nvme_setup_cmd | grep 'nvme_cmd_flush'
```

In this case, for the 10 million transfer benchmark, total flushes go from ~3000 to ~2300.

### DIFF
--- a/src/io/linux.zig
+++ b/src/io/linux.zig
@@ -442,12 +442,16 @@ pub const IO = struct {
                 .timeout => |*op| {
                     sqe.prep_timeout(&op.timespec, 0, 0);
                 },
-                .write => |op| {
-                    sqe.prep_write(
+                .writev2 => |*op| {
+                    // By setting rw_flags below, writev becomes writev2 - this is what liburing
+                    // does.
+                    sqe.prep_writev(
                         op.fd,
-                        op.buffer[0..buffer_limit(op.buffer.len)],
+                        @as(*const [1]posix.iovec_const, &op.iovec),
                         op.offset,
                     );
+
+                    sqe.rw_flags = op.rw_flags;
                 },
             }
             sqe.user_data = @intFromPtr(completion);
@@ -753,7 +757,7 @@ pub const IO = struct {
                     const result: TimeoutError!void = err;
                     completion.callback(completion.context, completion, &result);
                 },
-                .write => {
+                .writev2 => {
                     const result: WriteError!usize = blk: {
                         if (completion.result < 0) {
                             const err = switch (@as(posix.E, @enumFromInt(-completion.result))) {
@@ -838,10 +842,11 @@ pub const IO = struct {
         timeout: struct {
             timespec: os.linux.kernel_timespec,
         },
-        write: struct {
+        writev2: struct {
             fd: fd_t,
-            buffer: []const u8,
+            iovec: posix.iovec_const,
             offset: u64,
+            rw_flags: u32,
         },
     };
 
@@ -1268,16 +1273,24 @@ pub const IO = struct {
         fd: fd_t,
         buffer: []const u8,
         offset: u64,
+        options: struct { dsync: bool },
     ) void {
         completion.* = .{
             .io = self,
             .context = context,
             .callback = erase_types(Context, WriteError!usize, callback),
             .operation = .{
-                .write = .{
+                // Internally, use the writev operation: it allows per IO control of fsync. This is
+                // only available on the vectorized write op, so use that, but limit it to a single
+                // iovec.
+                .writev2 = .{
                     .fd = fd,
-                    .buffer = buffer,
+                    .iovec = .{
+                        .base = buffer.ptr,
+                        .len = buffer_limit(buffer.len),
+                    },
                     .offset = offset,
+                    .rw_flags = if (options.dsync) os.linux.RWF.DSYNC else 0,
                 },
             },
         };
@@ -1436,7 +1449,11 @@ pub const IO = struct {
         var flags: posix.O = .{
             .CLOEXEC = true,
             .ACCMODE = if (method == .open_read_only) .RDONLY else .RDWR,
-            .DSYNC = true,
+
+            // Even though DSYNC false is the default, spell it out here explicitly. Non-grid writes
+            // are written on a per IO basis with RWF_DSYNC. Grid writes aren't, and are flushed
+            // before returning from compaction after each beat.
+            .DSYNC = false,
         };
         var mode: posix.mode_t = 0;
 
@@ -1548,8 +1565,8 @@ pub const IO = struct {
             },
         }
 
-        // This is critical as we rely on O_DSYNC for fsync() whenever we write to the file:
-        assert(flags.DSYNC);
+        // Potentially surprising: see the explanation in `var flags`.
+        assert(!flags.DSYNC);
 
         const fd = try posix.openat(dir_fd, relative_path, flags, mode);
         // TODO Return a proper error message when the path exists or does not exist (init/start).
--- a/src/lsm/forest.zig
+++ b/src/lsm/forest.zig
@@ -247,6 +247,7 @@ pub fn ForestType(comptime _Storage: type, comptime groove_cfg: anytype) type {
         compaction_schedule: CompactionSchedule,
 
         scan_buffer_pool: ScanBufferPool,
+        flush: Grid.Flush = undefined,
 
         pub fn init(
             forest: *Forest,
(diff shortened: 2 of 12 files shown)


src/storage.zig
```diff
@@ -383,6 +393,13 @@ pub fn StorageType(comptime IO: type) type {
             assert(write.offset % constants.sector_size == 0);
             self.assert_bounds(write.buffer, write.offset);
 
+            const dsync = write.zone.dsync();
```
cb22, line 396: The first version of this exposed `dsync` as an option on `write_sectors`. I thought moving it inside `start_write`, and relying on the `zone` already present would be less prone to error.

src/vsr.zig
```diff
@@ -163,6 +163,22 @@ pub const Zone = enum {
         };
     }
 
+    pub fn dsync(zone: Zone) bool {
+        return switch (zone) {
+            .grid => false,
+            else => true,
+        };
```
sentientwaffle, line 170: Is the `switch` statement (rather than `return zone != .grid;`) to be more explicit?

src/testing/storage.zig
```diff
@@ -662,9 +667,23 @@ pub const Storage = struct {
...
     }
 
+    pub fn flush_sectors(
+        storage: *Storage,
+        callback: *const fn (flush: *Storage.Flush) void,
+        flush: *Storage.Flush,
+    ) void {
+        storage.unflushed = 0;
+        flush.callback = callback;
+        flush.callback(flush);
```
sentientwaffle, line 684: I think we need to properly simulate loss of unflushed writes, and also `flush_sectors` being properly asynchronous.

The simplest approach (for the former) might be to add a new `memory_unsynced: []align(constants.sector_size) u8` for the whole storage, though that nearly doubles VOPR's memory requirements, which is a bit much.

Maybe instead:
- Add new `writes_unsynced`/`writes_unsynced_buffers` arrays  (similar to `overlays`/`overlay_buffers`, except in reverse -- they store the old data, while the new data is in `memory`.
- On `reset()` unsynced writes can be randomly reverted.

src/vsr.zig
```diff
@@ -163,6 +163,22 @@ pub const Zone = enum {
...
+    }
+
+    // Ensuring only the grid is written without dsync is _critical_ to safety!
```
sentientwaffle, line 173: :thinking: I think client replies would also work? Similarly to grid blocks, they are reconstructed when we replay the WAL during recovery.

src/lsm/forest.zig
```diff
@@ -540,6 +541,34 @@ pub fn ForestType(comptime _Storage: type, comptime groove_cfg: anytype) type {
...
+            // On the last beat, wait until everything has been persisted before calling the
+            // callback.
+            if (last_beat) {
+                forest.grid.superblock.storage.flush_sectors(flush_callback, &forest.flush);
+            } else {
+                forest.compact_finish_dispatch_callback();
+            }
+        }
+
+        pub fn flush_callback(flush: *Grid.Flush) void {
```
sentientwaffle, line 553:
```suggestion
        fn flush_callback(flush: *Grid.Flush) void {
```

src/lsm/forest.zig
```diff
@@ -540,6 +541,34 @@ pub fn ForestType(comptime _Storage: type, comptime groove_cfg: anytype) type {
...
+            const forest: *Forest = @alignCast(@fieldParentPtr("flush", flush));
+            forest.compact_finish_dispatch_callback();
+        }
+
+        pub fn compact_finish_dispatch_callback(forest: *Forest) void {
```
sentientwaffle, line 558:
```suggestion
        fn compact_finish_dispatch_callback(forest: *Forest) void {
```

src/lsm/forest.zig
```diff
@@ -540,6 +541,34 @@ pub fn ForestType(comptime _Storage: type, comptime groove_cfg: anytype) type {
...
+            assert(forest.compaction_progress != null);
+            assert(forest.compaction_progress.?.trees_done);
+            assert(forest.compaction_progress.?.manifest_log_done);
+
+            const op = forest.progress.?.compact.op;
+            const compaction_beat = op % constants.lsm_compaction_ops;
+            const last_beat = compaction_beat == constants.lsm_compaction_ops - 1;
+
+            if (last_beat) {
+                assert(forest.grid.superblock.storage.unflushed == 0);
```
sentientwaffle, line 569: Can we also add this same assertion to `SuperBlock.checkpoint()`?

src/io/linux.zig
```diff
@@ -1436,7 +1449,11 @@ pub const IO = struct {
         var flags: posix.O = .{
             .CLOEXEC = true,
             .ACCMODE = if (method == .open_read_only) .RDONLY else .RDWR,
-            .DSYNC = true,
+
+            // Even though DSYNC false is the default, spell it out here explicitly. Non-grid writes
+            // are written on a per IO basis with RWF_DSYNC. Grid writes aren't, and are flushed
+            // before returning from compaction after each beat.
+            .DSYNC = false,
```
sentientwaffle, line 1456: In this function's comment:
```
    /// - Ensures that the file data (and file inode in the parent directory) is durable on disk.
    ///   The caller is responsible for ensuring that the parent directory inode is durable.
```
I think that is referring to `DSYNC = true`, so should be updated.

src/storage.zig
```diff
@@ -446,6 +464,48 @@ pub fn StorageType(comptime IO: type) type {
...
+                .start = self.tracer.time.monotonic_instant(),
+            };
+
+            // Windows doesn't have per IO fsync support yet - it runs with FILE_WRITE_THROUGH which
+            // is the same as O_DSYNC.
+            if (builtin.os.tag == .windows) {
+                self.unflushed = 0;
+
+                return self.on_fsync(&flush.completion, {});
+            }
```
sentientwaffle, line 484: Should this use `next_tick` to avoid calling the callback synchronously? Maybe we could add a `next_tick: NextTick` to `Storage.Flush`.

src/lsm/forest.zig
```diff
@@ -540,6 +541,34 @@ pub fn ForestType(comptime _Storage: type, comptime groove_cfg: anytype) type {
...
             }
 
+            // On the last beat, wait until everything has been persisted before calling the
+            // callback.
+            if (last_beat) {
+                forest.grid.superblock.storage.flush_sectors(flush_callback, &forest.flush);
+            } else {
+                forest.compact_finish_dispatch_callback();
+            }
+        }
```
sentientwaffle, line 551: Just curious, but am I correct that if we queued `{write, fsync}` then the fsync would not necessarily sync that write, since the write hadn't completed (cqe) by the time the fsync was processed?

(That is not relevant here of course, since at this point all of this bar's compaction writes have already "completed").

sentientwaffle (commented):
Best branch name yet :laughing:

cb22:
We've decided to not go ahead with this for now, and instead:
* implement fallocate / growing of the data file as needed,
* optimize scheduling of compaction work, and,
* figure something out for macOS separately.

Maybe, once all those are done, we'll revisit this (or not!).

PR #2369, opened 2024-10-02
Multiversioning: Read the entire binary

sentientwaffle:
The `"in-place upgrade"` integration test went from passing consistently to failing _most_ (though not all) of the time.

In the upgrade integration test, after we replace the binary, the running replica detects that the binary changed.
Then it reads the new binary.

When the whole binary fits in a single read, the test passes.
When the whole binary doesn't fit in a single read, it fails to parse (`error.InvalidELF`).

(I'm guessing that this started failing as the binary has grown larger (as more versions are bundled each week).)

Fix: ~Use the size from `statx()`, and keep reading from increasing offsets until the entire binary is read.~
Read until one of the reads returns zero bytes. (Since `statx()` is only available on linux).

### DIFF
--- a/src/multiversioning.zig
+++ b/src/multiversioning.zig
@@ -2,6 +2,7 @@ const builtin = @import("builtin");
 const std = @import("std");
 const stdx = @import("stdx.zig");
 const assert = std.debug.assert;
+const maybe = stdx.maybe;
 const os = std.os;
 const posix = std.posix;
 const constants = @import("constants.zig");
@@ -578,6 +579,8 @@ pub const Multiversion = struct {
 
     source_buffer: []align(8) u8,
     source_fd: ?posix.fd_t = null,
+    source_offset: ?u64 = null,
+    source_size: ?u64 = null,
 
     target_fd: posix.fd_t,
     target_path: [:0]const u8,
@@ -760,7 +763,7 @@ pub const Multiversion = struct {
         assert(self.stage == .init);
         assert(!self.timeout.ticking);
 
-        self.binary_open();
+        self.binary_statx();
 
         assert(self.stage != .init);
 
@@ -802,9 +805,6 @@ pub const Multiversion = struct {
     fn on_timeout(self: *Multiversion) void {
         self.timeout.reset();
 
-        assert(builtin.target.os.tag == .linux);
-        if (comptime builtin.target.os.tag != .linux) return; // Prevent codegen.
-
         switch (self.stage) {
             .source_stat,
             .source_open,
@@ -814,6 +814,14 @@ pub const Multiversion = struct {
 
             .init, .ready, .err => {},
         }
+        self.binary_statx();
+    }
+
+    fn binary_statx(self: *Multiversion) void {
+        assert(self.stage == .init or self.stage == .ready or self.stage == .err);
+        assert(builtin.target.os.tag == .linux);
+        if (comptime builtin.target.os.tag != .linux) return; // Prevent codegen.
+
         self.stage = .source_stat;
         self.io.statx(
             *Multiversion,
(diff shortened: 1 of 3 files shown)


src/multiversioning.zig
```diff
@@ -578,6 +578,8 @@ pub const Multiversion = struct {
 
     source_buffer: []align(8) u8,
     source_fd: ?posix.fd_t = null,
+    source_offset: ?u64 = null,
+    source_size: ?u64 = null,
```
matklad: Can pack these fields into `stage.source_read`?

matklad: Ahh, we need `size` also for `stage.source_open`, then, yeah, perhaps simpler to just keep them as nullable fields

src/multiversioning.zig
```diff
@@ -842,26 +849,31 @@ pub const Multiversion = struct {
...
             &self.timeout_statx,
         ))) {
-            log.info("binary change detected: {s}", .{self.exe_path});
+            if (self.timeout_statx_previous != .none) {
+                log.info("binary change detected: {s}", .{self.exe_path});
+            }
 
             self.stage = .init;
-            self.binary_open();
+            self.binary_open(self.timeout_statx.size);
```
matklad: check that statx.size is not greater than the size of the buffer we are going to read stuff into

src/multiversioning.zig
```diff
@@ -890,18 +902,35 @@ pub const Multiversion = struct {
...
         self.source_fd = fd;
+        self.binary_read();
+    }
+
+    fn binary_read(self: *Multiversion) void {
+        assert(self.stage == .source_read);
+        assert(self.source_fd != null);
+        assert(self.source_offset != null);
+        assert(self.source_size != null);
+        assert(self.source_offset.? < self.source_size.?);
```
matklad: assert source_offset < self.source_buffer.len

src/multiversioning.zig
```diff
@@ -911,16 +940,46 @@ pub const Multiversion = struct {
...
-        self.target_update(source_buffer) catch |e| return self.handle_error(e);
-        assert(self.stage == .ready);
+        self.source_offset.? += bytes_read;
+        assert(self.source_offset.? <= self.source_buffer.len);
+
+        // The file is smaller than we expected.
+        // It was likely modified in the time since our statx().
+        if (bytes_read == 0) return self.handle_error(error.FileModifiedDuringRead);
+
+        switch (std.math.order(self.source_offset.?, self.source_size.?)) {
```
matklad: An alternative here is not to bother with `statx.size`, and just read until read returns `0`. But, given we _have_ `statx`, we might as well use it.

sentientwaffle: Ah, actually I'll go with your suggestion -- we only have statx on Linux.

sentientwaffle:
This PR fixes one bug, revealing another bug with identical symptoms hiding underneath.

There is a [config](https://github.com/tigerbeetle/tigerbeetle/blob/f66289bd9c7357546b379dadbffb38e91a8eb11c/src/config.zig#L138) which defines the "base" max binary size (set to 64MiB). In `constants.zig` we multiple that by 2 for debug binaries, so it is 128MiB.

`0.16.4`'s debug build is already 126MiB. So when the integration test tries to use a new multiversion binary past that, it gets stuck. And unfortunately, fixing this isn't as simple as just bumping the multiplier up (it should be more like 8x or 10x). Because that would still break auto-upgrades for anyone upgrade from a release `<=0.16.4`:
- When you swap in the larger binary for auto-upgrade, the running (`0.16.4`) replica can't read it (too big), so it assumes it is invalid and won't auto upgrade.
- And even worse, if you manually restart into the new binary, it execs into the old one, and then can't upgrade since it can't read its own multiversion metadata.

src/multiversioning.zig
```diff
@@ -911,16 +930,34 @@ pub const Multiversion = struct {
...
+                posix.close(self.source_fd.?);
+                self.source_offset = null;
+                self.source_fd = null;
+            }
+        }
 
         const bytes_read = result catch |e| return self.handle_error(e);
-        const source_buffer = self.source_buffer[0..bytes_read];
+        self.source_offset.? += bytes_read;
+        assert(self.source_offset.? <= self.source_buffer.len);
```
matklad, line 951: Hm, so what happens here if we read a binary that's larger than our buffer?

It seems that:

* we'll read right to the end of the buffer
* and then we'll do another `read` syscall passing-in an empty slice
* and that'll return 0 bytes read
* so we'll interpret that as a file of just the right size

Should we perhaps explicitly treat that as an error?

But I think I also have an argument for why this is OK: we don't have a guarantee that we read the entire binary anyway. Like, if someone `cp`s binary by doing writes, rather than an atomic rename, we totally can get truncation. So we have to deal with truncated binaries anyway, and we do, by verifying checksum:

https://github.com/tigerbeetle/tigerbeetle/blob/cd8f4e556c0cdb9726e838f21a73bc5a078fcb71/src/multiversioning.zig#L966-L971

So I'd say the logic here is correct, but is correct in an un-obvious way, and it's worth adding something like

```
// This could be a truncated file, but it'll get rejected 
// when we verify the checksum.
maybe(self.source_offset.? == self.source_buffer.len);
```

sentientwaffle:
Note that the multiversion integration test for this PR will continue to fail until it is rebased atop https://github.com/tigerbeetle/tigerbeetle/pull/2370.

matklad (approved):

PR #2370, opened 2024-10-03
build: verify multiversion header during build

matklad:
Right now, it is possible to build a multiversion binary where last release is not the newest of the pack. Luckily, we validate that when _loading_ the binary, but we definitely should have validated that during the build as well!

What's worse, it seems like we don't have any binary size validation whatsoever, ~~and that, in fact, our latest debug release blew past the limit already!~~ Assert&fix that as well. It seems like we've caught that just in time --- the _next_ release would've blew past the limit!

This does require cutting the number of releases we include from 6 to just 5.

### DIFF
--- a/src/build_multiversion.zig
+++ b/src/build_multiversion.zig
@@ -104,6 +104,13 @@ pub fn main() !void {
             .output = cli_args.output,
         }),
     }
+
+    const stat = try shell.cwd.statFile(cli_args.output);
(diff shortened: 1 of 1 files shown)


matklad:
I also considered reducing the number of included binaries only for _debug_, but I think it's useful to have release&debug to be identical. 

We should probably actually go and look into this whole separate debug info story...

cb22 (approved):

PR #2375, opened 2024-10-03
Config: Increase size limit of multiversion debug binaries

sentientwaffle:
`Debug` builds are more than 2x larger than `ReleaseSafe` builds.

### DIFF
--- a/src/constants.zig
+++ b/src/constants.zig
@@ -112,18 +112,18 @@ pub const vsr_releases_max = config.cluster.vsr_releases_max;
 /// The maximum cumulative size of a final TigerBeetle output binary - including potential past
 /// releases and metadata.
 pub fn multiversion_binary_platform_size_max(options: struct { macos: bool, debug: bool }) u64 {
-    // {Linux, Windows} get the base value. macOS gets 2x since it has universal binaries. All cases
-    // get a further 2x in debug.
+    // {Linux, Windows} get the base value. macOS gets 2x since it has universal binaries.
+    // All cases get a further 8x in debug.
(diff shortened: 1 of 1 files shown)


matklad:
If we do this, we should also bump `const past_count: u32 = @min(3, header.past.count);` back to `4`.

sentientwaffle:
Yes, though not immediately, otherwise we will break auto-upgrading.

matklad:
Ahhh, right, we should do that once the next version is the oldest! Actually, even one version _after_ that, as it will cease to be the oldest if we increase the limit 🤣

matklad:
Hm, I am not feeling great about this. With our current parameters, this'll mean that `debug` binaries would allocate an extra half a gig right of the bat for multi versioning (as opposed to current 128). This feels like a bit too much.

That is, while the multiplier of `8` does feel true, it seems like we should play with some parameters _elsewhere_ to afford it.

cb22:
If we relax the latest-version-always-execs-first rule, we could lower the buffer required to the size of a single release's maximum (plus metadata).

We could handle on-disk format changes (say, we introduced compression) with a _visits_ release.

sentientwaffle:
We could only use the debug binary for the latest release, and for all of the past versions use the `ReleaseSafe` build.

cb22:
I'm going to close this for now: the plan is to skip copying the _whole_ binary into memory, and only copying a particular release into memory (on Linux, it's on disk on Mac / Windows) when needed to execute it.

That is, our memory usage will be bounded by the size of an individual release, not the binary as a whole.

PR #2376, opened 2024-10-03
ci: avoid dependency on gh

matklad:
We use `gh` to download releases, but that is suboptimal as:

* this is an extra dependency,
* it requires auth.

But zig is a perfectly fine `curl`, via `zig fetch`, so let's use that.

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -497,37 +497,27 @@ fn build_tigerbeetle_executable_multiversion(b: *std.Build, options: struct {
 
 // Downloads a pre-build llvm-objcopy from <https://github.com/tigerbeetle/dependencies>.
 fn build_tigerbeetle_executable_get_objcopy(b: *std.Build) std.Build.LazyPath {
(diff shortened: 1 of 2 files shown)


build.zig
```diff
@@ -1477,99 +1467,96 @@ fn download_release(
...
-        if (hash != verify_checksum.checksum) {
-            std.log.err("checksum mismatch, specified '{x}', got '{x}'", .{
-                verify_checksum.checksum,
-                hash,
-            });
-            return error.ChecksumMismatch;
-        }
-        step.result_cached = true;
-        verify_checksum.generated_file.path = source_path;
+// Use 'zig fetch' to download and unpack the specified URL, optionally verifying the checksum.
```
cb22, line 1509: Does `zig fetch` extract zips?

matklad: It does! And it also chmods things!

cb22 (approved):
The embedded source does give me that early PHP feeling, but I'm not sure if having it standalone would be better either. Either way, this is definitely an improvement!

PR #2377, opened 2024-10-04
state_machine: fix cross-release bug, tests, rename FormerAccountFilter

cb22:
After adding new fields to the `AccountFilter` in 0.16.2, a `FormerAccountFilter` was added for compatibility with older clients.

Its usage was determined by `account_filter_from_input()`, which switches appropriately depending on the filter type, but this fn wasn't called in the execute path for `execute_get_account_balances`.

Instead, it always tried to interprete the filter as the new type, leading to a panic when executing `get_account_balances` on a replica running 0.16.2 or newer from an older client.

This commit fixes that bug, adds a few tests to catch it, and adds the ability to `setup_client_release` in state_machine unit tests for testing cross-release things like this. All it does currently is set the client_release passed in to things like `input_valid` and `prefetch`.

While we're here, `FormerAccountFilter` -> `AccountFilterFormer`.

### DIFF
--- a/src/state_machine.zig
+++ b/src/state_machine.zig
@@ -598,8 +598,8 @@ pub fn StateMachineType(
                     if (input.len != 0) return false;
                 },
                 .get_account_transfers, .get_account_balances => {
-                    const event_size: u32 = if (FormerAccountFilter.required(client_release))
-                        @sizeOf(FormerAccountFilter)
+                    const event_size: u32 = if (AccountFilterFormer.required(client_release))
+                        @sizeOf(AccountFilterFormer)
                     else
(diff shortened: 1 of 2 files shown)


src/state_machine.zig
```diff
@@ -3182,7 +3179,8 @@ const TestContext = struct {
     superblock: SuperBlock,
     grid: Grid,
     state_machine: StateMachine,
-    busy: bool = false,
+    busy: bool,
```
cb22, line 3182: TestContext is initialized as 

```
var test_context: TestContext = undefined;
test_context.init(...);
```

So default values are ignored!

batiati (approved):

PR #2378, opened 2024-10-04
[FIX] Client compatibility if a transient error is solved and retried.

batiati:
Assert clients < 0.16.4 are compatible with scenarios where the new results `id_already_failed` and `exists_with_different_ledger` would be returned.

Fix an overtight assertion that would cause the **replica to crash** if a transient error is solved.

### DIFF
--- a/src/lsm/groove.zig
+++ b/src/lsm/groove.zig
@@ -1214,8 +1214,12 @@ pub fn GrooveType(
         pub fn insert(groove: *Groove, object: *const Object) void {
             assert(object.timestamp >= TimestampRange.timestamp_min);
             assert(object.timestamp <= TimestampRange.timestamp_max);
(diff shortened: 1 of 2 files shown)


src/lsm/groove.zig
```diff
@@ -1214,8 +1214,12 @@ pub fn GrooveType(
         pub fn insert(groove: *Groove, object: *const Object) void {
             assert(object.timestamp >= TimestampRange.timestamp_min);
             assert(object.timestamp <= TimestampRange.timestamp_max);
-            if (constants.verify) {
```
batiati, line 1217: Removed `constants.verify` here for symmetry with `insert_orphaned_ids`.

cb22 (approved):

PR #2379, opened 2024-10-04
Add missing Query API in docs and C header

batiati:
- Update "requests" page with `query_*` operations.
- Update C header with query filter/flags

### DIFF
--- a/docs/reference/requests/README.md
+++ b/docs/reference/requests/README.md
@@ -23,6 +23,8 @@ event in the request.
   `credit_account_id`
 - [`get_account_balances`](./get_account_balances.md): fetch the historical account balance by the
   `Account`'s `id`.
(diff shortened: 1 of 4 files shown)


docs/README.md
```diff
@@ -59,6 +59,6 @@ Learn more about TigerBeetle's [mission, history](./about/README.md),
 
 - [𝕏](https://twitter.com/tigerbeetledb)
 - [GitHub](https://github.com/tigerbeetle/tigerbeetle)
-- [Slack](https://slack.tigerbeetle.com/invite)
```
cb22: Hmm - I think the idea was that we could keep this constant and update it there without spamming git?

batiati: That's true, I didn't notice it was a `tigerbeetle.com` url. :+1:

cb22 (approved):

PR #2383, opened 2024-10-05
stdx: port Rust's dbg! to Zig

matklad:
Utility for printf debugging, a-la Rust's dbg! macro.

dbg prints the value with the prefix, while also returning the value, which makes it convenient to drop it in the middle of a complex expression.

### DIFF
--- a/src/tidy.zig
+++ b/src/tidy.zig
@@ -134,6 +134,12 @@ fn tidy_banned(source: []const u8) ?[]const u8 {
         return "FIX" ++ "ME comments must be addressed before getting to main";
     }
 
+    if (std.mem.indexOf(u8, source, "dbg(") != null) {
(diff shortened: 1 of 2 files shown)


matklad:
Now trying to debug compaction 😅

src/tidy.zig
```diff
@@ -134,6 +134,12 @@ fn tidy_banned(source: []const u8) ?[]const u8 {
         return "FIX" ++ "ME comments must be addressed before getting to main";
     }
 
+    if (std.mem.indexOf(u8, source, "dbg(") != null) {
+        if (std.mem.indexOf(u8, source, "fn dbg(") == null) {
+            return "forgot to remove dbg?";
```
cb22:
```suggestion
            return "dbg() must be removed before getting to main";
```

cb22 (approved):

PR #2384, opened 2024-10-06
[Docs] Revamp Client Documentation

batiati:
- Rewrite the "walkthrough" sample project to ensure each snippet is self-contained (i.e., no undeclared variables and use `const` instead of mutable variables).
- Standardize all examples across all languages by using consistent IDs, codes, etc.
- Mention the recommended time-based `ID()` function in the examples.

Related: https://github.com/tigerbeetle/tigerbeetle/pull/2337

### DIFF
--- a/src/clients/java/README.md
+++ b/src/clients/java/README.md
@@ -83,8 +83,6 @@ mvn install
 Now, create `src/main/java/Main.java` and copy this into it:
 
 ```java
-package com.tigerbeetle.samples;
-
 import com.tigerbeetle.*;
 
 public final class Main {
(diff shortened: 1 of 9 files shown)


src/clients/java/README.md
```diff
@@ -652,53 +781,84 @@ This approach gives the application a chance to correct failed imported events,
...
-// First, load and import all accounts with their timestamps from the historical source.
-accounts = new AccountBatch(historicalAccounts.length);
-for(int index = 0; index < historicalAccounts.length; index += 1) {
-    accounts.add();
+// External source of time
+long historicalTimestamp = 0L;
+ResultSet historicalAccounts = null; // Loaded from an external source;
+ResultSet historicalTransfers = null ; // Loaded from an external source.
 
+var BATCH_SIZE = 8190;
```
sentientwaffle, line 789: Once client exposes the runtime batch size we should use that instead, so that this example works as expected on replicas started with `--development`.

sentientwaffle (approved):

PR #2387, opened 2024-10-07
cfo: ability to use hostname as git committer

cb22:


src/shell.zig
```diff
@@ -609,10 +609,19 @@ pub fn spawn(
...
+        const hostname = try std.posix.gethostname(&hostname_buffer);
+
+        try shell.env.put("GIT_AUTHOR_NAME", hostname);
+        try shell.env.put("GIT_COMMITTER_NAME", hostname);
+    } else {
+        try shell.env.put("GIT_AUTHOR_NAME", "TigerBeetle Bot");
+        try shell.env.put("GIT_COMMITTER_NAME", "TigerBeetle Bot");
+    }
+
     try shell.env.put("GIT_AUTHOR_EMAIL", "bot@tigerbeetle.com");
```
cb22, line 627: I don't think it's needed, but could make the address `bot@{hostname}` too?

matklad: Let's keep it simple!

matklad (approved):

PR #2391, opened 2024-10-07
vopr: pass correct op_min to find_latest_headers_break_between in core_missing_prepare

chaitanyabhandari:
Fixes failed VOPR seed `./zig/zig build -Drelease vopr -- 6316830026767099413` on main (a8f5b2b4845fdae6e64a2b5b20b6563195c76f66). 

Currently, `core_missing_prepare` can pass incorrect arguments to `find_latest_headers_break_between`: op_min=op_checkpoint, op_max=op_prepare_max, which triggers the below assert. Even though it's perfectly reasonable for a replica to have _self.commit_min=op_checkpoint_, and _self.op=op_prepare_max_, this state indicates that the op_prepare_max has already replaced op_checkpoint in the journal! 
https://github.com/tigerbeetle/tigerbeetle/blob/ed0a1c73a24b2caa9fbcecf54b52dfbb27ded83d/src/vsr/journal.zig#L627

As `core_missing_prepare` aims to find out why a replica isn't committing _beyond_ commit_min, _op_min_ passed to `find_latest_headers_break_between`  must be _commit_min + 1_.

### DIFF
--- a/src/vopr.zig
+++ b/src/vopr.zig
@@ -692,7 +692,7 @@ pub const Simulator = struct {
                     // Check if replica was stuck while repairing headers. Find largest missing
                     // header as we repair headers from high -> low ops.
                     if (replica.journal.find_latest_headers_break_between(
-                        replica.commit_min,
+                        replica.commit_min + 1,
                         @min(replica.op, commit_max),
                     )) |range| {
                         if (missing_header_op == null or missing_header_op.? < range.op_max) {
--- a/src/vopr.zig
+++ b/src/vopr.zig
@@ -680,7 +680,7 @@ pub const Simulator = struct {
 
         // Don't check for missing uncommitted ops (since the StateChecker does not record them).
         // There may be uncommitted ops due to pulses/upgrades sent during liveness mode.
-        const commit_max: u64 = simulator.cluster.state_checker.commits.items.len - 1;
+        const cluster_commit_max: u64 = simulator.cluster.state_checker.commits.items.len - 1;
 
         var missing_header_op: ?u64 = null;
         var missing_prepare_op: ?u64 = null;
@@ -688,12 +688,14 @@ pub const Simulator = struct {
         for (simulator.cluster.replicas) |replica| {
             if (simulator.core.isSet(replica.replica) and !replica.standby()) {
                 assert(simulator.cluster.replica_health[replica.replica] == .up);
-                if (replica.op > replica.commit_min) {
+                const commit_max = @min(replica.op, cluster_commit_max);
+                const commit_min = replica.commit_min;
+                if (commit_min < commit_max) {
                     // Check if replica was stuck while repairing headers. Find largest missing
                     // header as we repair headers from high -> low ops.
                     if (replica.journal.find_latest_headers_break_between(
-                        replica.commit_min + 1,
-                        @min(replica.op, commit_max),
+                        commit_min + 1,
+                        commit_max,
                     )) |range| {
                         if (missing_header_op == null or missing_header_op.? < range.op_max) {
                             missing_header_op = range.op_max;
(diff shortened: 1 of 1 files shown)


src/vsr/journal.zig
```diff
@@ -624,7 +624,7 @@ pub fn JournalType(comptime Replica: type, comptime Storage: type) type {
             assert(journal.status == .recovered);
             assert(journal.header_with_op(op_max) != null);
             assert(op_max >= op_min);
-            assert(op_max - op_min + 1 <= slot_count);
+            assert(op_max - op_min <= slot_count);
```
matklad: I am wondering if we should fix it here, or at the call-site? 

Suppose that we are in a situation where op_min is the checkpoint, and op_max is op_prepare_max. 

I _think_ this means that `op_min` is actually outside our journal --- it is checkpointed, so the journal needs only to include ops higher than that. 

In _this_ function, I think then `const A = journal.header_with_op(op_min);` would return `null`, and we'll return `op_min` as a part of the header break, while it's not the case? 

IOW, I _think_ the caller should be able to assert that any prepare from the resulting range can actually be appended to the journal safely (between op_repair_min and op_prepare_max (🤔 is this a sign that we should rename op_repair_min to op_prepare_min?)). But if we allow op_min=op_checkpoint here, that would be violated?

chaitanyabhandari: Mhmmmm, thank you for bringing this up, I _think_ you're right about fixing it at the call site... I certainly hadn't thought this one through this clearly and was waiting on VOPR1000 to give me some feedback! Let me read through this function very carefully and dry run this case. 

> But if we allow op_min=op_checkpoint here, that would be violated?

Also let me think about this case as well. 🤔

src/vopr.zig
```diff
@@ -692,7 +692,7 @@ pub const Simulator = struct {
                     // Check if replica was stuck while repairing headers. Find largest missing
                     // header as we repair headers from high -> low ops.
                     if (replica.journal.find_latest_headers_break_between(
-                        replica.commit_min,
+                        replica.commit_min + 1,
                         @min(replica.op, commit_max),
```
matklad: will break if commit_min == commit_max?

chaitanyabhandari: Ah, I thought that since `core_missing_prepare` is supposed to check for cases where we `commit_min < commit_max`, we wouldn't hit that case at all. 🤔

chaitanyabhandari: Mhmmmmm but `pending` checks for so many more conditions apart from just that. So we _could_ encounter a case where `commit_min == commit_max` but the cluster still hasn't converged in liveness mode...
https://github.com/tigerbeetle/tigerbeetle/blob/cbf7b34d905f7cf10c5ce7307cb754eb31722034/src/vopr.zig#L504

src/vsr/journal.zig
chaitanyabhandari: >  I think the caller should be able to assert that any prepare from the resulting range can actually be appended to the journal safely (between op_repair_min and op_prepare_max

Also, this is such a great point, I hadn't thought about it this way at all! Improved my mental model. 😄

src/vopr.zig
chaitanyabhandari: Surprising to me that VOPR1000 hasn't yet been able to find a case `core_missing_prepare` breaks due to `commit_min == commit_max`. I went through all of the conditions in `pending` and it _does_ seem to me like `commit_min < commit_max` would be a symptom for all of the conditions that `pending` checks... 

Either way, I revisited that logic in `core_missing_prepare` and clarified it a bit as a part of https://github.com/tigerbeetle/tigerbeetle/pull/2391/commits/741610cef1de769360409ee9e86b6c2482472f6f.

matklad (approved):

PR #2392, opened 2024-10-07
test: Create more valid transfers in Java workload

owickstrom:
This PR generates more valid transfers by tracking pending IDs, and in general making sure the flags follow the rules. There are still quite a few failed transfers (~25%) with this workload, but that's OK. We want to test failure paths too.

Also, we now check that linked accounts and transfers succeed or fail together.

### DIFF
--- a/src/testing/systest/workload/src/main/java/Command.java
+++ b/src/testing/systest/workload/src/main/java/Command.java
@@ -1,8 +1,10 @@
 import java.math.BigInteger;
 import java.util.ArrayList;
 import java.util.BitSet;
+import java.util.Optional;
 import java.util.stream.Collectors;
 import com.tigerbeetle.AccountBatch;
+import com.tigerbeetle.AccountFlags;
 import com.tigerbeetle.Client;
 import com.tigerbeetle.CreateAccountResultBatch;
 import com.tigerbeetle.IdBatch;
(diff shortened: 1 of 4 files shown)


src/testing/systest/workload/src/main/java/Command.java
```diff
@@ -120,36 +135,59 @@ public CreateTransfersResult execute(Client client) {
...
-      var creditAccount = model.accounts.get(transfer.creditAccountId());
-      assert debitAccount != null;
-      assert creditAccount != null;
-      assert debitAccount.ledger() == creditAccount.ledger();
+    Optional<ResultEntry<NewTransfer>> previousEntry = Optional.empty();
+
+    for (var current : entries) {
+      var transfer = current.value();
+
+      // Check that linked transfers succeed or fail together.
```
batiati, line 162: Suggestion for a future PR, could we assert flag-dependent results?

Some examples:
`LinkedEventFailed` is expected only when `current.flags.linked or previous.flags.linked`.
`Imported*` is expected only when `flags.imported`.
`PendingIdMustNotBeZero` only when `flags.pending` .

Actually, I'm aiming at asserting that `result[i]` is in fact related to `transfers[i]` and not mixed up across concurrent batches.

batiati (approved):

src/testing/systest/workload/src/main/java/Command.java
owickstrom: Yes, I've been thinking about the same! But I'll do that in the Zig port once that's in.

PR #2395, opened 2024-10-08
docs: Styling of `if`s and `for`s in sentence in docs

LiamSwayne:
If and for are keywords, so they have been highlighted as such in the markdown, which can provide some additional clarity in the affected sentence.


matklad (approved):
Thanks!

PR #2398, opened 2024-10-08
fix: correct AMOUNT_MAX value

### LINKED ISSUE #2396: Java client TransferBatch.AMOUNT_MAX incorrect
The current client is generated with the `AMOUNT_MAX` constant equal to the value of `170141183460469231740910675752738881536` (`0x00000000_00000080_00000000_00000080`) rather than the expected value of `340282366920938463463374607431768211455` (`0xffffffffffffffffffffffffffffffff`)

Should the constant instead be defined as `UInt128.asBigInteger(-1L, -1L);`?

https://github.com/tigerbeetle/tigerbeetle/blob/9cd85e3700715ee469797012c2fbd90ad896ddce/src/clients/java/java_bindings.zig#L110

tKe:
Correct value for the AMOUNT_MAX sentinel value definition in the java client.

fixes #2396

### DIFF
--- a/src/clients/java/src/test/java/com/tigerbeetle/TransferTest.java
+++ b/src/clients/java/src/test/java/com/tigerbeetle/TransferTest.java
@@ -209,6 +209,11 @@ public class TransferTest {
     }
 
     @Test
+    public void testAmountMax() {
+        assertEquals(BigInteger.TWO.pow(128).subtract(BigInteger.ONE), TransferBatch.AMOUNT_MAX);
+    }
+
+    @Test
     public void testPendingId() {
         var transfers = new TransferBatch(1);
         transfers.add();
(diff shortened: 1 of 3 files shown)


src/clients/java/src/test/java/com/tigerbeetle/TransferTest.java
```diff
@@ -208,6 +208,11 @@ public void testAmountLong() {
         assertEquals(1L, transfers.getAmount(UInt128.MostSignificant));
     }
 
+    @Test
+    public void testAmountMax() {
+        assertEquals(BigInteger.TWO.pow(128).subtract(BigInteger.ONE), TransferBatch.AMOUNT_MAX);
+    }
+
```
batiati, line 215: chef's kiss :pinched_fingers:

batiati (approved):
Thank you for spotting this!
I've seen off-by-one errors before, but off-by-2¹²⁸ was the first time :roll_eyes: :rofl:

PR #2385, opened 2024-10-06
vsr: nack different prepares b/w DVC & journal iff DVC is non-blank

chaitanyabhandari:
This PR fixes the failed seed `./zig/zig build  vopr -- --lite 364110200398911730` on main (https://github.com/tigerbeetle/tigerbeetle/commit/5159e60472d154dfcf43f44d5bc36d2f8186913f). 

### Problem

In this seed, we observe a replica R whose durable DVC headers contain blanks (as they were first populated using a journal that had corrupt slots). These DVC headers cause R to not contribute to the nack quorum for the blank ops (they are marked as `nack=false present=false`), causing view changes to hit the `quorum received, awaiting repair` case. However, when R crashes and restarts with a clean journal and the same DVC headers with blanks, it nacks the blank ops (they are marked as `nack=true present=false`), triggering the following assertion: https://github.com/tigerbeetle/tigerbeetle/blob/ed0a1c73a24b2caa9fbcecf54b52dfbb27ded83d/src/vsr/replica.zig#L10140

The root cause here is that `send_do_view_change` nacks an op if it is non-null in the journal, and its checksum in the journal doesn't match its checksum in the DVC. However, this logic does not correctly honor _blank_ DVCs. In other words, if an op in a DVC is blank, it is *always* nacked, regardless of whether its present in the journal or not.
https://github.com/tigerbeetle/tigerbeetle/blob/ed0a1c73a24b2caa9fbcecf54b52dfbb27ded83d/src/vsr/replica.zig#L7458-L7461

### Solution

The solution entails comparing the checksum between a header in the journal and the DVC *iff* the header is non-blank in the DVC. This ensures that if the DVC with blanks was populated using a corrupt journal (perhaps due to misdirected reads), and the replica starts up with the *correct* ops in the journal (no misdirected reads), the blank ops aren't nacked while broadcasting the DVC.

### Detailed root-cause analysis

<details>
<summary>Detailed root-cause analysis for 364110200398911730</summary>

<br>
R1 starts up with log_view=view=130 with a certain head op=81, and corruptions in slots 0 → 15. While transitioning to view_change for view=131 on startup, R1 uses a corrupt journal to populate its DVC headers, leading to some blanks in its headers.

```C
[debug] (superblock): 1: view_change: vsr_header: op=81 checksum=46810032428715586571191746510006090604
[debug] (superblock): 1: view_change: vsr_header: op=80 checksum=128466774560797227729232685875003653228
[...]

PR #2399, opened 2024-10-08
vsr: DVC header is blank and journal has valid header from old wrap

chaitanyabhandari:
Fixes `./zig/zig build -Drelease -Dvopr-log=full vopr -- --ticks-max-requests=3000000 7746807953021648557` on the matklad/c2 branch (4ee9445f69a4eaf976e795a3d8f4c1214d05fc57). There is no failing seed on main yet but main is vulnerable to this issue as well.

This is a regression due to https://github.com/tigerbeetle/tigerbeetle/pull/2385. While https://github.com/tigerbeetle/tigerbeetle/pull/2385 *correctly* solves a safety issue, it misses out on correctly handling  a *blank DVC header from the current wrap with a valid journal header from an old wrap*. We must nack such headers as the op from the current wrap is *guaranteed* to not exist in the journal.

This is fixed by using `header_for_op` → `header_with_op` while populating the nack & present bitsets in [send_do_view_change](https://github.com/tigerbeetle/tigerbeetle/blob/ed0a1c73a24b2caa9fbcecf54b52dfbb27ded83d/src/vsr/replica.zig#L7417). While both these functions fetch a *valid* header from the slot where an op resides, there exists a subtle difference between the two:
* `header_for_op`:  valid header could be from an old wrap
* `header_with_op`: valid header corresponds to the specified op

Using `header_with_op`, nacking a *blank DVC header from the current wrap with a valid journal header from an old wrap* is handled by the following pre-existing condition:
https://github.com/tigerbeetle/tigerbeetle/blob/ed0a1c73a24b2caa9fbcecf54b52dfbb27ded83d/src/vsr/replica.zig#L7463-L7466

### DIFF
--- a/src/vsr/replica.zig
+++ b/src/vsr/replica.zig
@@ -7451,7 +7451,7 @@ pub fn ReplicaType(
             var present = BitSet.initEmpty();
             for (self.view_headers.array.const_slice(), 0..) |*header, i| {
                 const slot = self.journal.slot_for_op(header.op);
-                const journal_header = self.journal.header_for_op(header.op);
+                const journal_header = self.journal.header_with_op(header.op);
                 const dirty = self.journal.dirty.bit(slot);
                 const faulty = self.journal.faulty.bit(slot);
 
--- a/src/vsr/replica.zig
+++ b/src/vsr/replica.zig
@@ -7455,41 +7455,44 @@ pub fn ReplicaType(
                 const dirty = self.journal.dirty.bit(slot);
                 const faulty = self.journal.faulty.bit(slot);
 
-                // Case 1: We have this header in memory, but haven't persisted it to disk yet.
-                if (journal_header != null and journal_header.?.checksum == header.checksum and
-                    dirty and !faulty)
-                {
-                    nacks.set(i);
-                }
-
-                // Case 2: We have a _different_ prepare — safe to nack even if it is faulty.
-                if (journal_header != null and vsr.Headers.dvc_header_type(header) == .valid and
-                    journal_header.?.checksum != header.checksum)
-                {
-                    nacks.set(i);
-                }
-
-                // Case 3: We don't have a prepare at all, and that's not due to a fault.
+                // Nack bit case 1: We don't have a prepare at all, and that's not due to a fault.
                 if (journal_header == null and !faulty) {
                     nacks.set(i);
                 }
 
-                // Presence bit: the prepare is on disk, is being written to disk, or is cached
-                // in memory. These conditions mirror logic in `on_request_prepare` and imply
-                // that we can help the new primary to repair this prepare.
-                if ((self.journal.prepare_inhabited[slot.index] and
-                    self.journal.prepare_checksums[slot.index] == header.checksum) or
-                    self.journal.writing(header.op, header.checksum) or
-                    self.pipeline_prepare_by_op_and_checksum(header.op, header.checksum) != null)
-                {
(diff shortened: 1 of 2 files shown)


matklad:
Let's also add a replica test for the case that is fixed by https://github.com/tigerbeetle/tigerbeetle/pull/2385!

chaitanyabhandari:
@matklad: Bumping this in case you missed seeing it, I have added the replica test for the case fixed by https://github.com/tigerbeetle/tigerbeetle/pull/2385 in https://github.com/tigerbeetle/tigerbeetle/pull/2399/commits/1a6024703708279f5379f037dc9f9f241eaf9589! 😄

matklad:
Yeah, I am reviewing this right now actually, sorry for the delay!

matklad (approved):
This looks great to me, but let's double-check what's up with 

```
./zig/zig build -Drelease vopr -- 13843591167547372328
```

chaitanyabhandari:
@matklad Alas, buggy `core_missing_prepare` strikes back in `./zig/zig build -Drelease vopr -- 13843591167547372328`!

Here, op=123 is missing on R0 and R1, but `core_missing_prepare` isn't able to correctly detect that. The reason for that is that there is _also_ a missing header op=105 on R0, but that header _is_ present on R1. 
```C
[debug] (simulator): 0: commit_min=79, commit_max=106
[debug] (simulator): 0: found missing header op=105
[debug] (simulator): 1: commit_min=122, commit_max=123
[debug] (simulator): 1: checking prepare op=123
[debug] (simulator): 1: found dirty/fault prepare op=123, current missing prepare op=null
[info] (cluster): no liveness, final cluster state (core=11100011):
[info] (cluster):  0   \ v         9356V  79/_79/102C  71:106Jo  0/_0J!  71:106Wo <__0:__0> v1:2    39Ga  0G!   0G?
[info] (cluster):  1   \  v        9356V 119/122/123C  92:123Jo  1/_1J!  92:123Wo <__0:__0> v2:2    39Ga  0G!   0G?
```
In other words, we incorrectly assume in `core_missing_prepare` that if a header is missing, it must be the root cause for the cluster being stuck. However, as we observed above, we could have it so that a header *and* a prepare are missing, but the cluster is actually stuck because of the *prepare*, not the header. In the above case, R0, the replica with the missing header was lagging and waiting to state sync, so it wasn't even _trying_ to fix its headers. So perhaps, instead of giving preference to header over the prepare, we should check *both* for whether they're missing? 🤔 

https://github.com/tigerbeetle/tigerbeetle/blob/ed0a1c73a24b2caa9fbcecf54b52dfbb27ded83d/src/vopr.zig#L717-L720

chaitanyabhandari:
Found a seed on main as well, for the symptom described in https://github.com/tigerbeetle/tigerbeetle/pull/2399#issuecomment-2407528193: 

```
./zig/zig build -Drelease vopr -- 14126642652659963795
```

matklad:
I think it's safe to say that we should be landing this then!

matklad:
>Here, op=123 is missing on R0 and R1, but core_missing_prepare isn't able to correctly detect that. The reason for that is that there is also a missing header op=105 on R0, but that header is present on R1.

Ah, I think the original assumption there was that you _should_ be able to repair the header in this case, but this seems to be not true: to repair headers, you need to be in .normal state, and that's exactly what we can't do here. 

So, yeah, checking for either prepare or header missing feels right here.

PR #2400, opened 2024-10-09
vopr: adjust log_replica hieroglyphs for new state sync

matklad:
Sync is now almost instantaneous, so we need to switch from switch_commence/switch_complete to something else. Mnemonics:

- $ S Sync
- ! bang Crash


sentientwaffle (approved):
Great mnemonic!

(If you were wondering, `^`/`$` for recovery/crash were originally chosen due to vim's begin/end-of-line motions.)

PR #2402, opened 2024-10-09
vsr: fix asserts for start_view update with missing root op

chaitanyabhandari:
Fixes `./zig/zig build -Drelease vopr -- --lite 471198279191464987` on the matklad/c2 branch (34b067681de03e7cf4c34c0426ac9c728bf59b3c). There is no failing seed on main yet but main is vulnerable to this issue as well.

In this seed, a repairing backup with a corrupted root op (op=0) attempts to update its SV headers while advancing its op_checkpoint from 0 → 19. `repair` allows the repairing backup to commit and advance its checkpoint since the header break is *after* the checkpoint (op_checkpoint=0).
https://github.com/tigerbeetle/tigerbeetle/blob/ed0a1c73a24b2caa9fbcecf54b52dfbb27ded83d/src/vsr/replica.zig#L6431-L6441

While updating its start_view headers, the repairing backup trips `assert(op_min <= self.op_repair_min());`, as op_min=1 and op_repair_min=0.
https://github.com/tigerbeetle/tigerbeetle/blob/ed0a1c73a24b2caa9fbcecf54b52dfbb27ded83d/src/vsr/replica.zig#L4879-L4881

This PR tightens asserts in `update_start_view_headers` to handle this case.

### DIFF
--- a/src/vsr/replica.zig
+++ b/src/vsr/replica.zig
@@ -4857,9 +4857,6 @@ pub fn ReplicaType(
             // All replicas are guaranteed to have no gaps in their headers between op_checkpoint
             // and commit_min. Between commit_min and self.op:
             // * The primary is guaranteed to have no gaps.
-            // * A potential primary in .view_change status is guaranteed to have no gaps, since we
-            //   only start committing *after* we have the hash chain intact up to self.op
-            //   (see `repair`).
             // * Backups are not guaranteed to have no gaps, they may have not received some
             //   prepares yet.
             const op_head_no_gaps = if (self.primary_index(self.view) == self.replica)
@@ -4884,15 +4881,19 @@ pub fn ReplicaType(
                 op_head_no_gaps,
             );
             const op_min = if (range) |r| r.op_max + 1 else range_min;
-
             assert(op_min <= op);
+
             if (self.op_checkpoint() == 0 and range != null) {
-                // We may get here with a missing root op if a we're a backup that is advancing its
-                // checkpoint mid repair.
-                assert(self.status == .normal and self.backup());
+                // We get here only if we are a backup with a missing/corrupt root op, advancing our
+                // checkpoint mid-repair. Primaries can never have a corrupt root op as repair
+                // ensures a primary's journal is clean before it transitions to .normal status.
+                assert(self.status == .normal);
+                assert(self.backup());
                 assert(self.commit_min == self.op_checkpoint_next_trigger());
+                assert(op_min == 1);
                 assert(range.?.op_max == 0);
                 assert(range.?.op_min == 0);
+                assert(self.journal.faulty.bit(.{ .index = 0 }));
             } else {
                 assert(op_min <= self.op_repair_min());
             }
--- a/src/vsr/replica.zig
+++ b/src/vsr/replica.zig
@@ -4884,8 +4884,18 @@ pub fn ReplicaType(
                 op_head_no_gaps,
             );
             const op_min = if (range) |r| r.op_max + 1 else range_min;
+
             assert(op_min <= op);
-            assert(op_min <= self.op_repair_min());
+            if (self.op_checkpoint() == 0 and range != null) {
+                // We may get here with a missing root op if a we're a backup that is advancing its
+                // checkpoint mid repair.
+                assert(self.status == .normal and self.backup());
+                assert(self.commit_min == self.op_checkpoint_next_trigger());
+                assert(range.?.op_max == 0);
+                assert(range.?.op_min == 0);
+            } else {
+                assert(op_min <= self.op_repair_min());
+            }
 
             // The SV includes headers corresponding to the op_prepare_max for preceding
             // checkpoints (as many as we have and can help repair, which is at most 2).


chaitanyabhandari:
Spreading the load since I already have an open PR (https://github.com/tigerbeetle/tigerbeetle/pull/2399) assigned to Alex, for which there is prior context as well. This one, OTOH, is a fresh issue! :)

src/vsr/replica.zig
```diff
@@ -4884,8 +4884,18 @@ pub fn ReplicaType(
                 op_head_no_gaps,
             );
             const op_min = if (range) |r| r.op_max + 1 else range_min;
+
             assert(op_min <= op);
```
sentientwaffle, line 4884: Nit, but I would shift this assert up one line, so that it isn't separated from the `const op_min = ...` assignment.

src/vsr/replica.zig
```diff
@@ -4884,8 +4884,18 @@ pub fn ReplicaType(
...
-            assert(op_min <= self.op_repair_min());
+            if (self.op_checkpoint() == 0 and range != null) {
+                // We may get here with a missing root op if a we're a backup that is advancing its
+                // checkpoint mid repair.
+                assert(self.status == .normal and self.backup());
```
sentientwaffle: Per tigerstyle, split `assert(... and ...)`:
```suggestion
                assert(self.status == .normal);
                assert(self.backup());
```

src/vsr/replica.zig
```diff
@@ -4884,8 +4884,18 @@ pub fn ReplicaType(
                 op_head_no_gaps,
             );
             const op_min = if (range) |r| r.op_max + 1 else range_min;
+
             assert(op_min <= op);
-            assert(op_min <= self.op_repair_min());
+            if (self.op_checkpoint() == 0 and range != null) {
+                // We may get here with a missing root op if a we're a backup that is advancing its
```
```suggestion
                // We may get here with a missing root op if we're a backup that is advancing its
```

src/vsr/replica.zig
```diff
@@ -4884,8 +4884,18 @@ pub fn ReplicaType(
...
+                // checkpoint mid repair.
+                assert(self.status == .normal and self.backup());
+                assert(self.commit_min == self.op_checkpoint_next_trigger());
+                assert(range.?.op_max == 0);
+                assert(range.?.op_min == 0);
```
sentientwaffle, line 4895: Since `range.op_max == 0` then can we assert that `op_min == 1`?

Can we also assert that `journal.faulty.bit(.{ .index = 0 })`?

Though, I'm missing how we know that `range.min == range.max == 0` here.

src/vsr/replica.zig
```diff
@@ -4884,8 +4884,18 @@ pub fn ReplicaType(
                 op_head_no_gaps,
             );
             const op_min = if (range) |r| r.op_max + 1 else range_min;
+
             assert(op_min <= op);
-            assert(op_min <= self.op_repair_min());
+            if (self.op_checkpoint() == 0 and range != null) {
+                // We may get here with a missing root op if a we're a backup that is advancing its
+                // checkpoint mid repair.
```
sentientwaffle: I would reword this to be imperative -- this comment is within the branch, so we are explaining what _is happening_, not what _could happen_.

src/vsr/replica.zig
chaitanyabhandari: > Though, I'm missing how we know that range.min == range.max == 0 here.

@sentientwaffle Mhmmmm, this flow _is_ getting a bit confusing. 😕 

`update_start_view_headers` is called by either:
1. primary about to perform superblock.checkpoint
2. primary refreshing its SV headers to send out start_view
3. backup about to perform superblock.checkpoint

For cases 1 & 2, the primary is _guaranteed_ to have a clean op=0 while invoking `update_start_view_headers` (repair ensures that). Even if op=0 is corrupted during normal processing, it can't be marked as faulty as it is never read from disk at runtime (we start committing from op=1).

For case 3, however, a backup could be repairing and advancing its checkpoint while it has a corrupt op=0 (if it is read as corrupted on startup), since the header break is considered be at <= self.op_checkpoint.
https://github.com/tigerbeetle/tigerbeetle/blob/ed0a1c73a24b2caa9fbcecf54b52dfbb27ded83d/src/vsr/replica.zig#L6431-L6441

chaitanyabhandari: @sentientwaffle: Although, your comment pushed me to think harder about this entire situation and now I feel that this solution is **_VERY_** wrong and could certainly have gaps other than in slot=0 here... 

We can only guarantee here we have no breaks between op_checkpoint → op_head_no_gaps. Here, since `range_min=op_head_no_gaps + 1) -| constants.journal_slot_count;`, it could actually _precede_ op_checkpoint! And we actually have no guarantees about the journal being clean from range_min → op_checkpoint.
https://github.com/tigerbeetle/tigerbeetle/blob/ed0a1c73a24b2caa9fbcecf54b52dfbb27ded83d/src/vsr/replica.zig#L4872-L4878

I am also very surprised at the fact that VOPR1000 hasn't been able to find a failing seed related to this. In fact, the PR is doing quite well on CFO. 😅

chaitanyabhandari:
All comments addressed in https://github.com/tigerbeetle/tigerbeetle/pull/2402/commits/4c5a86b81affd948aa7cfd1f67f83e406365a9da! 

Also, I sketched this out properly and realized that this solution _is_ correct. Although all of the things that I have mentioned [here](https://github.com/tigerbeetle/tigerbeetle/pull/2402#discussion_r1795829627) _are_ correct, they only apply to cases where `op_checkpoint != 0`, and therefore are all subsumed within the else branch:
```C
            if (self.op_checkpoint() == 0 and range != null) {
                // We get here only if we are a backup with a missing/corrupt root op, advancing our
                // checkpoint mid-repair. Primaries can never have a corrupt root op as repair
                // ensures a primary's journal is clean before it transitions to .normal status.
                assert(self.status == .normal);
                assert(self.backup());
                assert(self.commit_min == self.op_checkpoint_next_trigger());
                assert(op_min == 1);
                assert(range.?.op_max == 0);
                assert(range.?.op_min == 0);
                assert(self.journal.faulty.bit(.{ .index = 0 }));
            } else {
                assert(op_min <= self.op_repair_min());
            }
```

When we are a repairing backup transitioning from op_checkpoint=0 → 19, we can say with certainty that there should be no breaks between ops=1 → 19. Therefore, op=0 is the only possible broken header in this situation!

sentientwaffle (approved):

PR #2404, opened 2024-10-11
devhub: Apply a bit of styling

fabioarnold:
Some tweaks to fonts, padding, and colors. I added the fonts we use for the website (Inter and Recursive).

Before:

devhub-before-light

devhub-before-dark

After:

devhub-after-light

devhub-after-dark


cb22:
@fabioarnold you'll need to add exceptions for [those files](https://github.com/tigerbeetle/tigerbeetle/actions/runs/11290391068/job/31402157256?pr=2404) to tidy.zig

cb22:
Hmm - we also probably don't want to have those binary woffs in this repo in any case - could we maybe load them cross-origin from tigerbeetle.com? (I recall you said Google Fonts is out because of GDPR)

src/devhub/style/Inter-Bold.woff2
matklad, line 1: Adding binary fonts to the database repo sounds like a bridge too far to my taste! I'd maybe use built-in fonts, or otherwise loaded them from some kind of CDN with subresource integrity. This is an internal tool, so CDN feels acceptable! (otherwise, I wonder if we could load fonts from our main website? )

fabioarnold:
I got rid of the fonts and exported the logo from Figma to get the text just right (also skewed at the correct angle).

cb22 (approved):

PR #2405, opened 2024-10-11
[Benchmark] Fix Uniform Distribution with Hot Accounts

batiati:
Benchmark fails with `credit_account_not_found` if `--account-count-hot` is used with uniform distribution.


brson (approved):
LGTM. Thanks for the fix!

PR #2409, opened 2024-10-14
vopr: core_missing_prepare only checks replicas that have attempted to repair

chaitanyabhandari:
### Problem

This PR fixes `./zig/zig build -Drelease vopr -- 14126642652659963795` (and also most recently `15696783805748628886`) on main (https://github.com/tigerbeetle/tigerbeetle/commit/dad59e92884fa9d0654c8eb40e3d1314845489a5). This is a false positive in the VOPR, wherein `core_missing_prepare` does not distinguish between:
* a replica that is not able to initiate repair as it cannot step up as potential primary.
* a replica that is able to initiate repair but not able to complete it due to missing prepares/headers. 

A replica can only initiate repair its WAL in certain states - if it is a potential primary in view_change status, or if it is in normal status. Lagging replicas cannot step up as potential primaries and do not initiate repair for their WAL.
https://github.com/tigerbeetle/tigerbeetle/blob/ed0a1c73a24b2caa9fbcecf54b52dfbb27ded83d/src/vsr/replica.zig#L7070-L7083

The key problem is that `core_missing_prepare` treats all the replicas in the core as the same and assumes that all of them should be able to converge to `commit_min == commit_max`. Specifically, it does not distinguish between a replica that is not able to initiate repair (lagging replicas) from a replica that is able to initiate repair but not able to complete it due to missing prepares/headers. 
https://github.com/tigerbeetle/tigerbeetle/blob/ee59910771501fe1158969c869a23b642af9f2cd/src/vopr.zig#L688-L693

### Solution 

`core_missing_prepare` skips checking lagging replicas in view_change status, since it is correct for them to not initiate repair. For example, in the above VOPR seed, we can see that R2 is lagging (its op_checkpoint is 99 while the maximum op_checkpoint is 139). Consequently, it hasn't been able to step up as a potential primary and hasn't initiated repair.

```C
[debug] (replica): 2: on_do_view_change: view=19526 quorum received
[debug] (replica): 2: on_do_view_change: dvc: replica=1 log_view=19525 op=143 commit_min=143 checkpoint=139
[debug] (replica): 2: on_do_view_change: dvc: header: replica=1 op=143 checksum=166389484745224136312941382982924441585 nack=false present=true type=valid
[debug] (replica): 2: on_do_view_change: dvc: replica=2 log_view=501 op=143 commit_min=106 checkpoint=99
[debug] (replica): 2: on_do_view_change: dvc: header: replica=2 op=143 checksum=166389484745224136312941382982924441585 nack=true present=false type=valid
[debug] (replica): 2: on_do_view_change: lagging primary; forfeiting (view=19526..19528 checkpoint=99..139)
```

### DIFF
--- a/src/vopr.zig
+++ b/src/vopr.zig
@@ -681,6 +681,15 @@ pub const Simulator = struct {
         // Don't check for missing uncommitted ops (since the StateChecker does not record them).
         // There may be uncommitted ops due to pulses/upgrades sent during liveness mode.
         const cluster_commit_max: u64 = simulator.cluster.state_checker.commits.items.len - 1;
+        var cluster_op_checkpoint: u64 = 0;
+        for (simulator.cluster.replicas) |*replica| {
+            if (simulator.core.isSet(replica.replica)) {
+                cluster_op_checkpoint = @max(
+                    cluster_op_checkpoint,
+                    replica.superblock.working.vsr_state.checkpoint.header.op,
+                );
+            }
+        }
 
         var missing_header_op: ?u64 = null;
         var missing_prepare_op: ?u64 = null;
@@ -688,6 +697,11 @@ pub const Simulator = struct {
         for (simulator.cluster.replicas) |replica| {
             if (simulator.core.isSet(replica.replica) and !replica.standby()) {
                 assert(simulator.cluster.replica_health[replica.replica] == .up);
+
+                // Lagging replicas do not initiate WAL repair during view change.
+                if (replica.status == .view_change and
+                    replica.op_checkpoint() < cluster_op_checkpoint) continue;
+
                 const commit_max = @min(replica.op, cluster_commit_max);
                 const commit_min = replica.commit_min;
                 if (commit_min < commit_max) {
--- a/src/vopr.zig
+++ b/src/vopr.zig
@@ -703,26 +703,25 @@ pub const Simulator = struct {
                     replica.op_checkpoint() < cluster_op_checkpoint) continue;
 
                 const commit_max = @min(replica.op, cluster_commit_max);
-                const commit_min = replica.commit_min;
-                if (commit_min < commit_max) {
-                    // Check if replica was stuck while repairing headers. Find largest missing
-                    // header as we repair headers from high -> low ops.
-                    if (replica.journal.find_latest_headers_break_between(
-                        commit_min + 1,
-                        commit_max,
-                    )) |range| {
-                        if (missing_header_op == null or missing_header_op.? < range.op_max) {
-                            missing_header_op = range.op_max;
-                        }
-                    } else {
-                        // Check if replica was stuck while repairing prepares. Find smallest
-                        // missing prepare as we repair prepares from low -> high ops.
-                        for (commit_min + 1..commit_max + 1) |op| {
-                            const header = simulator.cluster.state_checker.header_with_op(op);
-                            if (!replica.journal.has_clean(&header)) {
-                                if (missing_prepare_op == null or missing_prepare_op.? > op) {
-                                    missing_prepare_op = op;
-                                }
+                const op_repair_min = replica.op_repair_min();
+
+                // Check if replica's commit pipeline was stuck due to missing headers.
+                // Find largest missing header as we repair headers from high -> low ops.
+                if (replica.journal.find_latest_headers_break_between(
+                    op_repair_min,
+                    commit_max,
+                )) |range| {
+                    if (missing_header_op == null or missing_header_op.? < range.op_max) {
+                        missing_header_op = range.op_max;
+                    }
+                } else {
+                    // Check if replica's commit pipeline was stuck due to missing prepares.
+                    // Find smallest missing prepare as we repair prepares from low -> high ops.
+                    for (op_repair_min..commit_max + 1) |op| {
+                        const header = simulator.cluster.state_checker.header_with_op(op);
+                        if (!replica.journal.has_clean(&header)) {
+                            if (missing_prepare_op == null or missing_prepare_op.? > op) {
+                                missing_prepare_op = op;
                             }
                         }
                     }
--- a/src/vopr.zig
+++ b/src/vopr.zig
@@ -706,7 +706,7 @@ pub const Simulator = struct {
                 const op_repair_min = replica.op_repair_min();
 
                 // Check if replica's commit pipeline was stuck due to missing headers.
-                // Find largest missing header as we repair headers from high -> low ops.
+                // Find latest missing header as we repair headers from high -> low ops.
                 if (replica.journal.find_latest_headers_break_between(
                     op_repair_min,
                     commit_max,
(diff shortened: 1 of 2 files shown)


src/vopr.zig
```diff
@@ -690,7 +690,13 @@ pub const Simulator = struct {
...
                 const commit_max = @min(replica.op, cluster_commit_max);
                 const commit_min = replica.commit_min;
-                if (commit_min < commit_max) {
+
+                // WAL repair can only be performed by replicas in normal state or potential
+                // primaries (see repairs_allowed in replica.zig). Use the difference between view
+                // and log_view as a heuristic for whether a replica has attempted repair or not.
+                if (commit_min < commit_max and
+                    replica.view - replica.log_view <= simulator.cluster.replica_count)
+                {
```
sentientwaffle: Rather than using a heuristic, why not compare the replica's `op_checkpoint` with the cluster's `op_checkpoint`?

chaitanyabhandari: Mhmmmm good point, we could use `op_checkpoint` to figure this out as well!

However, I want to press pause on this PR till our discussion tomorrow, as it triggered an entire train of thought that I wanted to discuss with you! It is related to lagging replicas also initiating `repair` during view change, so they can commit as much as they can (and even step up as potential primaries if need be). 😄

src/vopr.zig
```diff
@@ -681,13 +681,27 @@ pub const Simulator = struct {
...
+        }
 
         var missing_header_op: ?u64 = null;
         var missing_prepare_op: ?u64 = null;
 
         for (simulator.cluster.replicas) |replica| {
             if (simulator.core.isSet(replica.replica) and !replica.standby()) {
                 assert(simulator.cluster.replica_health[replica.replica] == .up);
+
+                // Lagging replicas do not initate WAL repair during view change.
```
```suggestion
                // Lagging replicas do not initiate WAL repair during view change.
```

sentientwaffle (commented):
Besides the typo this LGTM, assuming the vopr agrees!

chaitanyabhandari:
@sentientwaffle great call regarding letting this soak in the CFO, turns out our solution was not complete after all! Specifically, we didn't have logic for the case where a potential primary has been able to commit up to `commit_max`, but it hasn't yet been able to repair _all_ faulty prepares in the journal. I have added that in https://github.com/tigerbeetle/tigerbeetle/pull/2409/commits/a452108873f0fe2ea0792b8ab06d7eeceae3e589.


However, our discussion yesterday makes we wonder if committed prepares be repaired *after* a replica has transitioned to primary... Specifically, should a replica be able step up as primary as long as it has:
* committed up to commit_max
* has repaired all uncommitted prepares that it needs to load into its prepare pipeline

I will try this out along with the *early start_view* idea!

chaitanyabhandari:
Giving the ideas from https://github.com/tigerbeetle/tigerbeetle/pull/2409#issuecomment-2419011461 a shot in https://github.com/tigerbeetle/tigerbeetle/pull/2414.

chaitanyabhandari:
Based on the ideas in https://github.com/tigerbeetle/tigerbeetle/pull/2409#issuecomment-2419011461,  https://github.com/tigerbeetle/tigerbeetle/pull/2414 ensures a replica doesn't wait for _all_ prepares in its journal to be repaired before stepping up as primary. Instead, it only waits for ops between op_repair_min → self.op. In https://github.com/tigerbeetle/tigerbeetle/pull/2409/commits/f4a3ea38f13a63c9f3f505e8fe7145d2bc363f8a, I've added the logic to `core_missing_prepare` as well!

@sentientwaffle with that, after tons of back and forth, this PR is finally ready for review. 😄

src/vopr.zig
```diff
@@ -713,6 +728,30 @@ pub const Simulator = struct {
...
+                    const header = replica.journal.header_for_op(index);
+                    const slot_in_repair_range =
+                        (!range_empty and slots.contains(.{ .index = index })) or
+                        (range_empty and index == op_repair_min_slot.index);
+                    if (header != null and slot_in_repair_range) {
+                        if (missing_prepare_op == null or missing_prepare_op.? > header.?.op) {
+                            missing_prepare_op = header.?.op;
+                        }
+                    }
+                }
```
sentientwaffle: Is it possible that there is a slot in the range with `!dirty` but with the wrong/missing header? (e.g. an op from a previous wrap.)

Also, should this all be conditional on the replica being a potential primary?

chaitanyabhandari: > Is it possible that there is a slot in the range with !dirty but with the wrong/missing header? (e.g. an op from a previous wrap.)

Ah yes, it should be only for a potential primary. This code is becoming so branchy though... 😞

> Is it possible that there is a slot in the range with !dirty but with the wrong/missing header? (e.g. an op from a previous wrap.)

Mhmmmm, that is a brilliant point! I had completely missed that, since we're actually repairing all _ops_ from op_repair_min → op, we would actually have to iterate on ops here, and use `header_with_op`. I will fix this.

chaitanyabhandari: This comment also fixes the same bug in https://github.com/tigerbeetle/tigerbeetle/pull/2414. I've actually been feeling a bit awkward about replica.zig code being copied into test code, but haven't been able to find a better way to do this in vopr.zig. 🤔

chaitanyabhandari:
Thought about this for a bit and realized that `core_missing_prepare` _could_ just check for missing headers and prepares between `op_repair_min` → `commit_max`, instead of `commit_min` → `commit_max`, as we don't step as primary till we've repaired everything between  `op_repair_min` → `commit_max`! It simplifies the logic a fair bit as well.

Thoughts, @sentientwaffle?

src/vopr.zig
```diff
@@ -681,34 +681,47 @@ pub const Simulator = struct {
...
+                if (replica.journal.find_latest_headers_break_between(
+                    op_repair_min,
+                    commit_max,
+                )) |range| {
+                    if (missing_header_op == null or missing_header_op.? < range.op_max) {
+                        missing_header_op = range.op_max;
+                    }
+                } else {
+                    // Check if replica's commit pipeline was stuck due to missing prepares.
+                    // Find smallest missing prepare as we repair prepares from low -> high ops.
```
sentientwaffle: Nit, but I would clarify s/smallest/earliest, since the size of the prepare is irrelevant.

sentientwaffle (commented):
Much simpler!

This LGTM, but we can let it bake in the vopr to see if anything else crops up.

src/vopr.zig
chaitanyabhandari: Ah, nice catch there! I did mean earliest & latest, but used smallest & largest; was perhaps thinking in terms of smallest index & largest index. 😄 

Addressed in https://github.com/tigerbeetle/tigerbeetle/pull/2409/commits/2fc1f816e89365b164f37ce384c115474a4dd867.

sentientwaffle (approved):

PR #2412, opened 2024-10-16
Test: Test stdlib stable sort

sentientwaffle:
Stable sort is critical for compaction correctness.

Zig stdlib does have a sort fuzz test, but it ~doesn't verify stability~, doesn't cover the presorted subarray case, and doesn't check arrays much larger than the sort algorithm's on-stack cache.

### DIFF
--- a/src/stdx/sort_test.zig
+++ b/src/stdx/sort_test.zig
@@ -1,0 +1,116 @@
+const std = @import("std");
+const assert = std.debug.assert;
+
+const stdx = @import("../stdx.zig");
+
+test "sort_stable" {
+    const Value = struct {
+        const Value = @This();
+
+        x: u32, // x determines the order of the values.
+        y: u32, // y ensures that values are distinct for the purpose of checking stability.
+
+        fn compare_x_ascending(_: void, a: Value, b: Value) bool {
+            return a.x < b.x;
+        }
+
+        fn compare_xy_ascending(_: void, a: Value, b: Value) bool {
+            if (a.x < b.x) return true;
+            if (a.x > b.x) return false;
+            return a.y < b.y;
+        }
+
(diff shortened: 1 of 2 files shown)


src/stdx/sort_test.zig
```diff
@@ -0,0 +1,116 @@
...
+            if (a.x < b.x) return true;
+            if (a.x > b.x) return false;
+            return a.y < b.y;
+        }
+
+        fn compare_xy_descending(_: void, a: Value, b: Value) bool {
+            if (a.x > b.x) return true;
+            if (a.x < b.x) return false;
+            return a.y < b.y;
+        }
```
matklad: Is it intentional that this is descending in x, but ascending in y? The name makes it feel like y should also be descending

matklad: Ah, I see that this is intentional. Maybe it'd be simpler to set `.y = 0` initially, and then set `.y = @intCast(i)` after we partitioned things?

matklad (commented):
I _think_ std actually does check stability, and does verify arrays larger than the stack size:

https://github.com/ziglang/zig/pull/20284

But, yeah, it's a good idea to double-check that on our side as well, and injecting sorted-runs is centrally a good idea, given how the algorithm works internally. 

This looks good to me, but maybe a comment on `compare_xy_descending` explaining that y is sorted ascended is worthwhile.

sentientwaffle:
> I think std actually does check stability, and does verify arrays larger than the stack size:

Thanks; I missed that! I had checked my local (`0.13.0`) version of `sort.zig`, not `main`'s.

> Maybe it'd be simpler to set `.y = 0` initially, and then set `.y = @intCast(i)` after we partitioned things?

Great idea!

matklad (approved):

PR #2414, opened 2024-10-17
vsr: replica needn't repair its *entire* journal to become primary

chaitanyabhandari:
Currently, a replica only steps up as primary once it has no header breaks from *op_repair_min → head op*, as well as no corrupt/faulty prepares *in the entirety of its journal*. 

This is overly restrictive; a replica needn't have a _completely_ pristine log to step up as primary. It can transition to .normal status as primary once:
* **commit_min == commit_max**: Requires non faulty headers and prepares between op_checkpoint → commit_max
* **Prepares loaded into its in-memory pipeline**: Requires non faulty headers and prepares for all uncommitted ops, between commit_max → head op
* **No broken headers & faulty prepares between op_repair_min → op_checkpoint**

In other words, a replica should always be able to recover back till op_repair_min, but faults *before* op_repair_min should not cause a replica to not step up as primary.

### DIFF
--- a/src/vsr/replica.zig
+++ b/src/vsr/replica.zig
@@ -6488,9 +6488,9 @@ pub fn ReplicaType(
                 );
             }
 
-            if (header_break != null or self.journal.dirty.count > 0) return;
-
             if (self.status == .view_change and self.primary_index(self.view) == self.replica) {
+                if (!self.primary_journal_repaired()) return;
+
                 if (self.commit_min == self.commit_max) {
                     if (self.commit_stage != .idle) {
                         // If we still have a commit running, we started it the last time we were
@@ -6706,14 +6706,35 @@ pub fn ReplicaType(
             return true;
         }
 
+        /// Primary must have no missing headers and no faulty prepares between op_repair_min
+        /// and self.op, to maintain the invariant that a replica can repair everything back
+        /// till op_repair_min
+        fn primary_journal_repaired(self: *Self) bool {
+            assert(self.status == .normal or self.status == .view_change);
+            assert(self.primary_index(self.view) == self.replica);
+            assert(self.view == self.log_view);
+
+            if (!self.valid_hash_chain_between(self.op_repair_min(), self.op)) return false;
+
+            for (self.op_repair_min()..self.op + 1) |op| {
+                const header = self.journal.header_with_op(op);
+                assert(header != null);
+                if (self.journal.dirty.bits.isSet(self.journal.slot_for_header(header.?).index)) {
+                    return false;
+                }
+            }
+
+            return true;
+        }
+
         /// Reads prepares into the pipeline (before we start the view as the new primary).
         fn primary_repair_pipeline(self: *Self) enum { done, busy } {
             assert(self.status == .view_change);
             assert(self.primary_index(self.view) == self.replica);
             assert(self.commit_max == self.commit_min);
             assert(self.commit_max <= self.op);
-            assert(self.journal.dirty.count == 0);
             assert(self.pipeline == .cache);
+            assert(self.primary_journal_repaired());
 
             if (self.pipeline_repairing) {
                 log.debug("{}: primary_repair_pipeline: already repairing...", .{self.replica});
@@ -6738,11 +6759,10 @@ pub fn ReplicaType(
             assert(self.commit_max == self.commit_min);
             assert(self.commit_max <= self.op);
             assert(self.commit_max >= self.op -| constants.pipeline_prepare_queue_max);
-            assert(self.journal.dirty.count == 0);
-            assert(self.valid_hash_chain_between(self.op_repair_min(), self.op));
             assert(self.pipeline == .cache);
             assert(!self.pipeline_repairing);
             assert(self.primary_repair_pipeline() == .done);
+            assert(self.primary_journal_repaired());
 
             var pipeline_queue = PipelineQueue{
                 .pipeline_request_queue_limit = self.pipeline_request_queue_limit,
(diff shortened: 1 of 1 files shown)


src/vsr/replica.zig
```diff
@@ -2312,7 +2312,7 @@ pub fn ReplicaType(
                             break;
                         }
                     }
-                } else unreachable;
+                } else return;
```
chaitanyabhandari: No header from the start_view headers is within our prepare_max, we bail out and wait for a more up-to-date start_view message. Earlier, our start_view headers were guaranteed to contain at least one hook op (op_prepare_max from previous checkpoints), but now that is not guaranteed as a primary only maintains a pristine journal from op_checkpoint → self.op.

src/vsr/replica.zig
```diff
@@ -4883,21 +4883,6 @@ pub fn ReplicaType(
...
-                // ensures a primary's journal is clean before it transitions to .normal status.
-                assert(self.status == .normal);
-                assert(self.backup());
-                assert(self.commit_min == self.op_checkpoint_next_trigger());
-                assert(op_min == 1);
-                assert(range.?.op_max == 0);
-                assert(range.?.op_min == 0);
-                assert(self.journal.faulty.bit(.{ .index = 0 }));
-            } else {
-                assert(op_min <= self.op_repair_min());
```
chaitanyabhandari: Earlier, we guaranteed a clean, non-faulty log from op_repair_min → self.op, while now we only guarantee clean log from op_checkpoint → self.op. Therefore, we can't assert `assert(op_min <= self.op_repair_min());`.
https://github.com/tigerbeetle/tigerbeetle/blob/ed0a1c73a24b2caa9fbcecf54b52dfbb27ded83d/src/vsr/replica.zig#L8357

chaitanyabhandari:
Seems to be doing well under fuzzing, marking this as ready for review!
Screenshot 2024-10-17 at 10 49 58 PM

chaitanyabhandari:
Spreading the load and assigning this to @matklad, as I realized that I have another upcoming PR that I have to assign to dj. I don't want to overwhelm him with too many of my PRs. 😄

src/vsr/replica.zig
```diff
@@ -8628,8 +8635,6 @@ pub fn ReplicaType(
                 assert(self.view == view_new);
                 assert(self.log_view == view_new);
                 assert(self.commit_min == self.commit_max);
-                assert(self.journal.dirty.count == 0);
-                assert(self.journal.faulty.count == 0);
```
matklad, line 8632: Hm, it is a bit sad to see these asserts go. Let's maybe extract "are there faulty slots since op_repair_min" function and assert this here and elsewhere. 

Is there some other way we can strengthen, rather than relax assertions in this PR?

chaitanyabhandari: Mhmmmm, we could perhaps add this:
```C
        fn primary_assert_journal_repaired(self: *Self) void {
            assert(self.valid_hash_chain_between(self.op_repair_min(), self.op));

            const slots = vsr.SlotRange{
                .head = self.journal.slot_for_op(self.op_repair_min()),
                .tail = self.journal.slot_with_op(self.op).?,
            };
            var iterator = self.journal.dirty.bits.iterator(.{ .kind = .set });
            while (iterator.next()) |index| {
                assert(!slots.contains(.{ .index = index }));
            }
        }
```

chaitanyabhandari: Or actually, _this_ so we can call this from `repair` too!
```C
        fn primary_journal_repaired(self: *Self) bool {
            if (!self.valid_hash_chain_between(self.op_repair_min(), self.op)) return false;

            const slots = vsr.SlotRange{
                .head = self.journal.slot_for_op(self.op_repair_min()),
                .tail = self.journal.slot_with_op(self.op).?,
            };
            var iterator = self.journal.dirty.bits.iterator(.{ .kind = .set });
            while (iterator.next()) |index| {
                if (slots.contains(.{ .index = index })) return false;
            }

            return true;
        }
```

chaitanyabhandari: Done in https://github.com/tigerbeetle/tigerbeetle/pull/2414/commits/eff5ae741845ea75bc8e38f5dab9bd3a8c4ed970!

src/vsr/replica.zig
```diff
@@ -6706,14 +6706,36 @@ pub fn ReplicaType(
...
 
+        /// Primary must have no missing headers and no faulty prepares between op_repair_min
+        /// and self.op, to maintain the invariant that a replica can repair everything back
+        /// till op_repair_min
+        fn primary_journal_repaired(self: *Self) bool {
+            assert(self.status == .normal or self.status == .view_change);
+            assert(self.primary_index(self.view) == self.replica);
+            assert(self.view == self.log_view);
+
+            if (!self.valid_hash_chain_between(self.op_repair_min(), self.op)) return false;
```
matklad, line 6717: I am _slightly_ concerned with performance here, as this is O(journal_slot_count), but I think we are not calling this on any hot paths, so should be fine.

src/vsr/replica.zig
```diff
@@ -6706,14 +6706,36 @@ pub fn ReplicaType(
...
+
+            for (self.op_repair_min()..self.op + 1) |op| {
+                const header = self.journal.header_with_op(op);
+                if (header == null or
```
matklad: Hm, seems rather we can `assert(header != null)` here? Otherwise, we'd have bailed due to `!valid_hash_chain` before, no?

chaitanyabhandari: Good point, addressed this in https://github.com/tigerbeetle/tigerbeetle/pull/2414/commits/e5126b07e9a1311da5669022ce020e7c6956951c!

src/vsr/replica.zig
```diff
@@ -8377,11 +8397,10 @@ pub fn ReplicaType(
             assert(self.do_view_change_quorum);
             assert(!self.pipeline_repairing);
             assert(self.primary_repair_pipeline() == .done);
+            assert(self.primary_journal_repaired());
 
             assert(self.commit_min == self.commit_max);
             assert(self.commit_max <= self.op);
-            assert(self.journal.dirty.count == 0);
-            assert(self.journal.faulty.count == 0);
             assert(self.valid_hash_chain_between(self.op_repair_min(), self.op));
```
matklad: Can drop this assert as that is a part of primary_journal_repaired now?

chaitanyabhandari: Ah, good catch! Removed it as a part of https://github.com/tigerbeetle/tigerbeetle/pull/2414/commits/9da6c0396c43b037c37d7f3067d64341734ac5de. 😄

matklad (approved):

PR #2415, opened 2024-10-17
Devhub: Fix invisible text for dark-mode chart labels

sentientwaffle:
(Tested in firefox and chromium.)

### DIFF
--- a/src/devhub/style.css
+++ b/src/devhub/style.css
@@ -127,6 +127,7 @@ td {
     }
 
     #charts {
+        color: black;
         filter: invert(100%);
     }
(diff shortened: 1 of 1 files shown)


fabioarnold (approved):
Also works in Safari. :)

I briefly looked into how to handle dark mode with ApexCharts. And the default already is:
```js
  window.Apex = {
    theme: {
      mode: "auto"
    }
  };
```
But some colors in dark mode are still too bright. So you have to provide a color palette.

Anyway, the current solution is simpler and works well enough. Thank you!

PR #2416, opened 2024-10-18
vsr: send start_view message as soon as you can construct start_view headers

chaitanyabhandari:
Currently, a primary only sends a start_view message to the backups after it has done the following:
* Repaired header hash chain _and_ prepares between op_repair_min → head op
* Committed up to commit_max

This is overly restrictive. A replica should be able to send a start_view message to the backups _as soon as_ it has repaired the header hash chain between op_repair_min → head op. In other words, once a replica has repaired headers and prepares, it can concurrently send start_view message to the backups and attempt to commit up to commit_max.

This is good for both _performance_ and _availability_:
* Performance: During view change, as backups can now repair concurrently while a potential primary attempts to commit up to commit_max
* Availability: This allows lagging replicas to repair, commit, advance their checkpoint *and* step up as primary, in case the most up-to-date replica isn't able to commit up to commit_max (due to a corruption in its grid, for instance)

### DIFF
--- a/src/vsr/replica.zig
+++ b/src/vsr/replica.zig
@@ -2179,14 +2179,14 @@ pub fn ReplicaType(
             assert(message.header.view >= self.view);
             assert(message.header.replica != self.replica);
             assert(message.header.replica == self.primary_index(message.header.view));
-            assert(message.header.commit >= message.header.checkpoint_op);
-            assert(message.header.op >= message.header.commit);
-            assert(message.header.op - message.header.commit <=
+            assert(message.header.commit_max >= message.header.checkpoint_op);
+            assert(message.header.op >= message.header.commit_max);
+            assert(message.header.op - message.header.commit_max <=
                 constants.pipeline_prepare_queue_max);
 
             // The SV message may be from a primary that hasn't yet committed up to its commit_max,
             // and the commit_max may be from the primary's *next* checkpoint.
-            maybe(message.header.commit - message.header.checkpoint_op >
+            maybe(message.header.commit_max - message.header.checkpoint_op >
                 constants.vsr_checkpoint_ops + constants.lsm_compaction_ops);
 
             if (message.header.view == self.log_view and message.header.op < self.op) {
@@ -2217,10 +2217,15 @@ pub fn ReplicaType(
 
             const view_checkpoint = start_view_message_checkpoint(message);
             if (vsr.Checkpoint.trigger_for_checkpoint(view_checkpoint.header.op)) |trigger| {
-                assert(message.header.commit >= trigger);
+                assert(message.header.commit_max >= trigger);
             }
             assert(
                 message.header.op <= vsr.Checkpoint.prepare_max_for_checkpoint(
+                    vsr.Checkpoint.checkpoint_after(view_checkpoint.header.op),
+                ).?,
+            );
+            assert(
+                message.header.commit_max <= vsr.Checkpoint.prepare_max_for_checkpoint(
                     vsr.Checkpoint.checkpoint_after(view_checkpoint.header.op),
                 ).?,
             );
@@ -2229,12 +2234,12 @@ pub fn ReplicaType(
             assert(view_headers[0].op == message.header.op);
             assert(view_headers[0].op >= view_headers[view_headers.len - 1].op);
 
-            if (vsr.Checkpoint.durable(self.op_checkpoint_next(), message.header.commit) and (
+            if (vsr.Checkpoint.durable(self.op_checkpoint_next(), message.header.commit_max) and (
             //  Cluster is at least two checkpoints ahead. Although SV's checkpoint is not
             //  guaranteed to be durable on a quorum of replicas, it is safe to sync to it, because
             //  prepares in this replica's WAL are no longer needed.
                 vsr.Checkpoint.durable(self.op_checkpoint_next() +
-                constants.vsr_checkpoint_ops, message.header.commit) or
+                constants.vsr_checkpoint_ops, message.header.commit_max) or
                 // Cluster is on the next checkpoint, and that checkpoint is durable and is safe to
                 // sync to. Try to optimistically avoid state sync and prefer WAL repair, unless
                 // there's evidence that the repair can't be completed.
@@ -2245,7 +2250,7 @@ pub fn ReplicaType(
                 // State sync: at this point, we know we want to replace our checkpoint
                 // with the one from this SV.
 
-                assert(message.header.commit > self.op_checkpoint_next_trigger());
+                assert(message.header.commit_max > self.op_checkpoint_next_trigger());
                 assert(view_checkpoint.header.op > self.op_checkpoint());
 
                 if (self.syncing == .idle) {
@@ -2303,15 +2308,19 @@ pub fn ReplicaType(
 
                 // Find the first message that fits, make it our new head.
                 for (view_headers) |*header| {
-                    assert(header.commit <= message.header.commit);
+                    assert(header.commit <= message.header.commit_max);
 
                     if (header.op <= self.op_prepare_max_sync()) {
                         if (self.log_view < self.view or
                             (self.log_view == self.view and header.op >= self.op))
                         {
-                            self.set_op_and_commit_max(header.op, message.header.commit, @src());
+                            self.set_op_and_commit_max(
+                                header.op,
+                                message.header.commit_max,
+                                @src(),
+                            );
                             assert(self.op == header.op);
-                            assert(self.commit_max >= message.header.commit);
+                            assert(self.commit_max >= message.header.commit_max);
                             break;
                         }
                     }
@@ -2395,7 +2404,7 @@ pub fn ReplicaType(
             assert(start_view_message.references == 1);
             assert(start_view_message.header.view == self.view);
             assert(start_view_message.header.op == self.op);
-            assert(start_view_message.header.commit == self.commit_max);
+            assert(start_view_message.header.commit_max == self.commit_max);
             assert(start_view_message.header.nonce == message.header.nonce);
 
             self.send_message_to_replica(message.header.replica, start_view_message);
@@ -4802,21 +4811,21 @@ pub fn ReplicaType(
         /// The caller owns the returned message, if any, which has exactly 1 reference.
         fn create_start_view_message(self: *Self, nonce: u128) *Message.StartView {
             assert(self.status == .normal or self.status == .view_change);
-            if (self.status == .normal) {
-                assert(self.commit_min == self.commit_max);
-            } else {
-                // Potential primaries may send a SV message before committing up to commit_max.
-                // (see `repair`).
-                assert(self.status == .view_change);
-                maybe(self.commit_min < self.commit_max);
-            }
-
             assert(self.syncing != .updating_checkpoint);
             assert(self.replica == self.primary_index(self.view));
             assert(self.commit_min <= self.op);
             assert(self.view >= self.view_durable());
             assert(self.log_view >= self.log_view_durable());
             assert(self.log_view == self.view);
+            if (self.status == .normal) {
+                assert(self.commit_min == self.commit_max);
+            } else {
+                // Potential primaries may send a SV message before committing up to commit_max.
+                // (see `repair`).
+                assert(self.status == .view_change);
+                assert(self.do_view_change_quorum);
+                maybe(self.commit_min < self.commit_max);
+            }
 
             self.primary_update_view_headers();
             assert(self.view_headers.command == .start_view);
@@ -4834,7 +4843,7 @@ pub fn ReplicaType(
                 .view = self.view,
                 .checkpoint_op = self.op_checkpoint(),
                 .op = self.op,
-                .commit = self.commit_max,
+                .commit_max = self.commit_max,
                 .nonce = nonce,
             };
 
@@ -5718,7 +5727,7 @@ pub fn ReplicaType(
 
                     // Syncing replicas must be careful about receiving SV messages, since they
                     // may have fast-forwarded their commit_max via their checkpoint target.
-                    if (message_header.commit < self.op_checkpoint()) {
+                    if (message_header.commit_max < self.op_checkpoint()) {
                         log.debug("{}: on_{s}: ignoring (older checkpoint)", .{
                             self.replica,
                             command,
@@ -6502,23 +6511,15 @@ pub fn ReplicaType(
             if (self.status == .view_change and self.primary_index(self.view) == self.replica) {
                 if (!self.primary_journal_repaired()) return;
 
-                self.primary_update_view_headers();
-
                 // Sending start_view messages to backups and committing up to commit_max can be
-                // performed concurrently. This is good for performance *and* availability, as it
-                // allows lagging backups to repair *while* the potential primary
-                // commits.
-                if (self.log_view == self.log_view_durable()) {
-                    // view-change headers and log_view may already be durable if checkpoint was
-                    // advanced during repair.
-                    self.primary_send_start_view_after_view_change();
-                } else {
-                    // Check staging as superblock.checkpoint() may currently be making log_view
-                    // durable.
-                    if (self.log_view > self.superblock.staging.vsr_state.log_view) {
-                        self.view_durable_update();
-                    }
-                    assert(self.checkpoint_updating() or self.view_durable_updating());
+                // performed concurrently. This is good for performance *and* availability, as
+                // it allows lagging backups to repair while the potential primary commits.
+                self.send_start_view();
+
+                // Check staging as superblock.checkpoint() may currently be updating view or
+                // log_view.
+                if (self.log_view > self.superblock.staging.vsr_state.log_view) {
+                    self.view_durable_update();
                 }
 
                 if (self.commit_min == self.commit_max) {
@@ -6743,6 +6744,7 @@ pub fn ReplicaType(
             assert(self.status == .normal or self.status == .view_change);
             assert(self.primary_index(self.view) == self.replica);
             assert(self.view == self.log_view);
+            if (self.status == .view_change) assert(self.do_view_change_quorum);
 
             if (!self.valid_hash_chain_between(self.op_repair_min(), self.op)) return false;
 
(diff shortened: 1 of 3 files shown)


src/vsr/replica.zig
sentientwaffle, line 1: In `message_header.zig`, this comment is no longer accurate:

```
    pub const StartView = extern struct {
        ...
        /// Set to `commit_min`/`commit_max` (they are the same).
        commit: u64,
        ...
    };
```

Maybe we should even rename this SV header field to `commit_max`? (Accompanied by a comment that it is often -- but not always -- also equal to `commit_min`.)

src/vsr/replica.zig
```diff
@@ -4269,7 +4272,6 @@ pub fn ReplicaType(
                 // headers for a potential primary because we could arrive here while the potential
                 // primary is still repairing (and thus may still have gaps in its journal).
                 assert(self.do_view_change_quorum);
-                assert(self.view_headers.command == .do_view_change);
```
sentientwaffle, line 4272: Why remove this assert? Afaict it is still accurate.

src/vsr/replica.zig
```diff
@@ -4799,10 +4801,18 @@ pub fn ReplicaType(
...
+            assert(self.status == .normal or self.status == .view_change);
+            if (self.status == .normal) {
+                assert(self.commit_min == self.commit_max);
+            } else {
+                // Potential primaries may send a SV message before committing up to commit_max.
+                // (see `repair`).
+                assert(self.status == .view_change);
+                maybe(self.commit_min < self.commit_max);
+            }
+
```
sentientwaffle: Could you reorder the new conditional asserts to shift them after the unconditional asserts? e.g.:

```zig
            assert(self.syncing != .updating_checkpoint);
            assert(self.replica == self.primary_index(self.view));
            assert(self.commit_min <= self.op);
            assert(self.view >= self.view_durable());
            assert(self.log_view >= self.log_view_durable());
            assert(self.log_view == self.view);
            if (self.status == .normal) {
                assert(self.commit_min == self.commit_max);
            } else {
                // Potential primaries may send a SV message before committing up to commit_max.
                // (see `repair`).
                assert(self.status == .view_change);
                maybe(self.commit_min < self.commit_max);
            }
```

(I just find it slightly more readable.)

src/vsr/replica.zig
```diff
@@ -4799,10 +4801,18 @@ pub fn ReplicaType(
...
         fn create_start_view_message(self: *Self, nonce: u128) *Message.StartView {
-            assert(self.status == .normal);
+            assert(self.status == .normal or self.status == .view_change);
+            if (self.status == .normal) {
+                assert(self.commit_min == self.commit_max);
+            } else {
+                // Potential primaries may send a SV message before committing up to commit_max.
+                // (see `repair`).
+                assert(self.status == .view_change);
+                maybe(self.commit_min < self.commit_max);
```
```suggestion
                assert(self.commit_min >= self.commit_max -| constants.pipeline_prepare_queue_max);
                assert(self.commit_min <= self.commit_max);
                maybe(self.commit_min < self.commit_max);
```

src/vsr/replica.zig
```diff
@@ -2180,12 +2180,15 @@ pub fn ReplicaType(
...
-            assert(message.header.commit - message.header.checkpoint_op <=
-                constants.vsr_checkpoint_ops + constants.lsm_compaction_ops);
             assert(message.header.op >= message.header.commit);
             assert(message.header.op - message.header.commit <=
                 constants.pipeline_prepare_queue_max);
 
+            // The SV message may be from a primary that hasn't yet committed up to its commit_max,
+            // and the commit_max may be from the primary's *next* checkpoint.
+            maybe(message.header.commit - message.header.checkpoint_op >
+                constants.vsr_checkpoint_ops + constants.lsm_compaction_ops);
```
sentientwaffle: Can we assert that `message.header.commit` does not exceed `prepare_max_for_checkpoint(message.header.checkpoint_op)`?

src/vsr/replica.zig
```diff
@@ -6491,6 +6502,25 @@ pub fn ReplicaType(
...
+                    // advanced during repair.
+                    self.primary_send_start_view_after_view_change();
+                } else {
+                    // Check staging as superblock.checkpoint() may currently be making log_view
+                    // durable.
+                    if (self.log_view > self.superblock.staging.vsr_state.log_view) {
+                        self.view_durable_update();
+                    }
+                    assert(self.checkpoint_updating() or self.view_durable_updating());
+                }
```
sentientwaffle: Maybe rework this to be less branchy? Iirc `primary_send_start_view_after_view_change`'s SV message will be dropped by `send_message if `self.log_view != self.log_view_durable()`, so we don't need to check that.

```zig
                // Sending start_view messages to backups and committing up to commit_max can be
                // performed concurrently. This is good for performance *and* availability, as it
                // allows lagging backups to repair *while* the potential primary commits.
                maybe(self.commit_min < self.commit_max);
                self.primary_send_start_view_after_view_change();
                // Check staging as superblock.checkpoint() may currently be making log_view
                // durable.
                if (self.log_view > self.superblock.staging.vsr_state.log_view) {
                    self.view_durable_update();
                }
```

Though that `self.commit_min < self.commit_max` makes me thing that we could move all of this into the `else` branch of the `if (self.commit_min == self.commit_max) {` check, e.g.:

```zig
                if (self.commit_min == self.commit_max) {
                        // If we still have a commit running, we started it the last time we were
                        // primary, and its still running. Wait for it to finish before repairing
                        // the pipeline so that it doesn't wind up in the new pipeline.
                        assert(self.commit_prepare.?.header.op >= self.commit_min);
                        assert(self.commit_prepare.?.header.op <= self.commit_min + 1);
                        assert(self.commit_prepare.?.header.view < self.view);
                        return;
                    }

                    // Repair the pipeline, which may discover faulty prepares and drive more
                    // repairs.
                    switch (self.primary_repair_pipeline()) {
                        // primary_repair_pipeline() is already working.
                        .busy => {},
                        .done => self.primary_start_view_as_the_new_primary(),
                    }
                } else {
                    // Sending start_view messages to backups and committing up to commit_max can be
                    // performed concurrently. This is good for performance *and* availability, as it
                    // allows lagging backups to repair *while* the potential primary commits.
                    self.primary_update_view_headers();
                    self.primary_send_start_view_after_view_change();

                    // Check staging as superblock.checkpoint() may currently be making log_view
                    // durable.
                    if (self.log_view > self.superblock.staging.vsr_state.log_view) {
                        self.view_durable_update();
                    }
                }
```

sentientwaffle: And not exactly related, but is the call to `self.primary_update_view_headers();` necessary? Doesn't `primary_send_start_view_after_view_change` call that internally via `create_start_view_message()`?

src/vsr/replica.zig
```diff
@@ -8028,10 +8061,6 @@ pub fn ReplicaType(
                     self.view > self.view_durable() or
                     self.syncing == .updating_checkpoint,
             );
-            // The primary must only persist the SV headers after repairs are done.
-            // Otherwise headers could be nacked, truncated, then restored after a crash.
```
sentientwaffle, line 8032: :thinking: I think I originally wrote this comment... but I don't recall what scenario I was referring to here:

> Otherwise headers could be nacked, truncated, then restored after a crash.

Any ideas?

src/vsr/replica.zig
```diff
@@ -8089,8 +8118,8 @@ pub fn ReplicaType(
         }
 
         fn primary_send_start_view_after_view_change(self: *Self) void {
-            assert(self.status == .normal);
-            assert(self.primary());
+            assert(self.status == .normal or self.status == .view_change);
+            assert(self.replica == self.primary_index(self.view));
```
sentientwaffle, line 8136: In the case that we are in `status == .view_change`, can we additionally assert `primary_journal_repaired`?

sentientwaffle: Also, in `primary_journal_repaired` itself can we assert that if `status == .view_change` that we have completed our DVC quorum?

src/vsr/replica.zig
```diff
@@ -4799,10 +4801,18 @@ pub fn ReplicaType(
...
         /// The caller owns the returned message, if any, which has exactly 1 reference.
         fn create_start_view_message(self: *Self, nonce: u128) *Message.StartView {
-            assert(self.status == .normal);
+            assert(self.status == .normal or self.status == .view_change);
+            if (self.status == .normal) {
+                assert(self.commit_min == self.commit_max);
+            } else {
+                // Potential primaries may send a SV message before committing up to commit_max.
+                // (see `repair`).
+                assert(self.status == .view_change);
```
sentientwaffle, line 4825: Can we assert `do_view_change_quorum`?

src/vsr/replica.zig
```diff
@@ -8105,8 +8134,7 @@ pub fn ReplicaType(
         fn view_durable_update_callback(context: *SuperBlock.Context) void {
             const self: *Self = @fieldParentPtr("superblock_context_view_change", context);
             assert(self.status == .normal or self.status == .view_change or
-                (self.status == .recovering and self.solo()) or
-                self.status == .recovering_head);
```
sentientwaffle, line 8109: Is this change related somehow?

src/vsr/replica.zig
```diff
@@ -8135,8 +8163,7 @@ pub fn ReplicaType(
             }
 
             // The view/log_view incremented while the previous view-change update was being saved.
-            // We check staging superblock here because a superblock.checkpoint() call may currently
-            // be making view & log_view durable.
+            // Check staging as superblock.checkpoint() may currently be making log_view durable.
```
```suggestion
            // Check staging as superblock.checkpoint() may be updating view/log_view.
```

src/vsr/replica.zig
```diff
@@ -8648,11 +8677,6 @@ pub fn ReplicaType(
                 assert(self.commit_min == self.commit_max);
                 assert(self.primary_journal_repaired());
 
-                // Now that the primary is repaired and in status=normal, it can update its
-                // view-change headers. view-change headers and log_view may already be durable if
-                // checkpoint was advanced during repair.
-                if (self.log_view > self.log_view_durable()) self.view_durable_update();
```
sentientwaffle, line 8654: Instead, assert that either:

- `self.log_view > self.log_view_durable()`, or
- There is a superblock update in progress, and `self.log_view == self.superblock.staging.vsr_state.log_view`. (Maybe this condition could be extracted into a function that `repair()` could assert as well.)

src/vsr/replica.zig
chaitanyabhandari: This is unrelated; I realized that we can't get here if we're in `recovering_head` status so I changed the assert!

src/vsr/replica.zig
chaitanyabhandari: Mhmmmm, I believe `commit_checkpoint_superblock` can now be called when view_headers.command is either .do_view_change _or_ .start_view, since we may call this callback after we've converted our headers to .start_view, but _before_ we finish committing and transition to .normal status.

src/vsr/replica.zig
chaitanyabhandari: Good catch about the invocation to `self.primary_update_view_headers()`, it is indeed extraneous as we invoke it in `create_start_view_message` as well.

Also, great point re the SV getting dropped if log_view_durable != log_view,  I was not satisfied with the branches here as well, and dropping the branch would make this cleaner!

src/vsr/replica.zig
chaitanyabhandari: Indeed, more readable!

src/vsr/replica.zig
chaitanyabhandari: Mhmmmm, I think this comment _still_ holds true, as we never persist any view headers that contain truncated ops. If we persist view headers headers during checkpoint, that ensures our original DVC headers with possibly truncated ops, and we only persist view headers during repair when we have completely repaired our journal from op_repair_min → head op.

It's the assert which doesn't hold true anymore, since we're now updating durable view for a potential primary _before_ it transitions to normal, so it _could_ be the case that `self.log_view == self.view and self.replica == self.primary_index(self.view) and self.status == .view_change` .

chaitanyabhandari:
Thanks for the review, @sentientwaffle! Addressed the comments in https://github.com/tigerbeetle/tigerbeetle/pull/2416/commits/3095ebe3f4019eb4b4e2fbdb88acce10fbb10497.

Turns out we can't add the asserts from https://github.com/tigerbeetle/tigerbeetle/pull/2416/files#r1813060483 & https://github.com/tigerbeetle/tigerbeetle/pull/2416/files#r1812905538:
* https://github.com/tigerbeetle/tigerbeetle/pull/2416/files#r1813060483: `primary_journal_repaired` may hold when we first update our view headers from DVC → SV in `repair`, but then may stop holding if the WAL is corrupted before a potential primary transitions to normal
* https://github.com/tigerbeetle/tigerbeetle/pull/2416/files#r1812905538: `commit_min` is not guaranteed to be within `constants.pipeline_prepare_queue_max` of `commit_max`, it may be lagging by more

src/vsr/replica.zig
```diff
@@ -4799,14 +4810,22 @@ pub fn ReplicaType(
...
             assert(self.log_view >= self.log_view_durable());
             assert(self.log_view == self.view);
+            if (self.status == .normal) {
+                assert(self.commit_min == self.commit_max);
+            } else {
+                // Potential primaries may send a SV message before committing up to commit_max.
+                // (see `repair`).
+                assert(self.status == .view_change);
+                assert(self.do_view_change_quorum);
+                maybe(self.commit_min < self.commit_max);
```
```suggestion
                assert(self.commit_min <= self.commit_max);
```

src/vsr/replica.zig
sentientwaffle: I think these comments might have been missed in the last round?

src/vsr/replica.zig
```diff
@@ -8088,9 +8115,10 @@ pub fn ReplicaType(
             assert(self.view_durable_updating());
         }
 
-        fn primary_send_start_view_after_view_change(self: *Self) void {
-            assert(self.status == .normal);
-            assert(self.primary());
+        fn send_start_view(self: *Self) void {
```
sentientwaffle: Imo it is ok to call this `primary_send_start_view`, even though the replica might not be in normal status.

src/vsr/replica.zig
chaitanyabhandari: This is first comment I was talking about in https://github.com/tigerbeetle/tigerbeetle/pull/2416#issuecomment-2435684626! We can't assert `primary_journal_repaired` as some prepares *might* get corrupted after the DVC headers were first converted into SV headers. In https://github.com/tigerbeetle/tigerbeetle/commit/3095ebe3f4019eb4b4e2fbdb88acce10fbb10497, instead of adding an assert, I added a check for this:

https://github.com/tigerbeetle/tigerbeetle/blob/3095ebe3f4019eb4b4e2fbdb88acce10fbb10497/src/vsr/replica.zig#L8196-L8202

Actually, we don't need all prepares to be present before sending SV. We only need the _headers_ to be intact from op_repair_min → self.op. I had decided to not separate out `primary_journal_repaired` into `primary_journal_headers_repaired` and `primary_journal_prepares_repaired` for the sake of simplicity, since this is the only place where the distinction is required. 🤔

chaitanyabhandari: I took a stab at separating them out in https://github.com/tigerbeetle/tigerbeetle/pull/2416/commits/df6caa5fb234c94f47702aab8af8e01a088ddd37, so we can refine the condition for sending SVs a bit more such that we send SVs as soon as the _header chain_ is repaired.

There is *some* code duplication across `primary_journal_repaired `, `primary_journal_headers_repaired ` and `primary_journal_prepares_repaired ` (specifically the asserts), but I think its still cleaner to retain  `primary_journal_repaired` since the separation between the two currently only matters during repair. Thoughts?

src/vsr/replica.zig
```diff
@@ -4799,14 +4810,22 @@ pub fn ReplicaType(
...
             assert(self.log_view >= self.log_view_durable());
             assert(self.log_view == self.view);
+            if (self.status == .normal) {
+                assert(self.commit_min == self.commit_max);
+            } else {
+                // Potential primaries may send a SV message before committing up to commit_max.
+                // (see `repair`).
+                assert(self.status == .view_change);
+                assert(self.do_view_change_quorum);
+                assert(self.commit_max >= self.commit_min);
```
sentientwaffle: Nit, but I would swap this to 
```suggestion
                assert(self.commit_min <= self.commit_max);
```
so that its for mirrors the similar assert in the preceding `status == .normal` branch.

src/vsr/replica.zig
sentientwaffle: Ah, thanks for the clarifications! Sorry, I overlooked that comment originally. :facepalm:

src/vsr/replica.zig
```diff
@@ -6489,7 +6509,21 @@ pub fn ReplicaType(
             }
 
             if (self.status == .view_change and self.primary_index(self.view) == self.replica) {
-                if (!self.primary_journal_repaired()) return;
+                if (!self.primary_journal_headers_repaired()) {
+                    return;
+                } else {
```
```suggestion
                if (!self.primary_journal_headers_repaired()) return;

```

src/vsr/replica.zig
```diff
@@ -8153,18 +8199,33 @@ pub fn ReplicaType(
...
+                },
+                .view_change => {
+                    if (self.log_view < self.view) {
+                        if (!self.do_view_change_quorum) self.send_do_view_change();
+                    } else {
+                        assert(self.log_view == self.view);
+                        // Potential primaries that have updated their SV headers can send out SV
+                        // messages (see `repair`). Check primary_journal_repaired again as our
+                        // WAL may have been corrupted after SV headers were updated for a potential
+                        // primary.
```
sentientwaffle: Comment out of date? We aren't checking `primary_journal_repaired` anymore.

chaitanyabhandari:
Addressed comments in https://github.com/tigerbeetle/tigerbeetle/pull/2416/commits/c6861678885081c7d6d80a5b7222116719b0475c!

src/vsr/replica.zig
chaitanyabhandari: No problem dj! It turned out for the best as it made me rethink this logic and refine the conditions under which we send SV. 😄

sentientwaffle (approved):

PR #2417, opened 2024-10-18
build: fix c client cross-compilation on macs

matklad:
tb_client_header is a code-generating program, it runs on host and should be build for host.


matklad:
(side effect of vortex review!)

owickstrom (approved):

PR #2419, opened 2024-10-21
Reformat tidy for fewer merge conflicts

brson:
I find that these lines are prone to merge conflicts, and I hypothesize that formatting these 1-per line will make the diffs simpler to read when fixing (though I have not tested that this is true).


matklad (approved):
Makes sense!

PR #2423, opened 2024-10-22
Tab completion in REPL

### LINKED ISSUE #1182: REPL: Add Tab Completion

raghunandanbhat:
Closes #1182. This is based on the changes in PR #2155.

### DIFF
--- a/src/completion.zig
+++ b/src/completion.zig
@@ -37,7 +37,7 @@ pub const Completion = struct {
     match_index: usize = 0,
 
     pub fn init(self: *Completion, allocator: std.mem.Allocator) !void {
-        self.* = .{
+        self.* = Completion{
             .matches = try CompletionList.init(allocator, completion_entry_bytes),
             .prefix = BoundedArray(u8, completion_entry_bytes){},
             .suffix = BoundedArray(u8, completion_entry_bytes){},
--- a/src/completion.zig
+++ b/src/completion.zig
@@ -5,8 +5,6 @@ const assert = std.debug.assert;
 const RingBuffer = @import("ring_buffer.zig").RingBufferType;
 const BoundedArray = stdx.BoundedArrayType;
 
-const completion_entries = 256;
-const completion_entry_bytes = 512;
 const keywords = [_][]const u8{
     "create_accounts",
     "create_transfers",
@@ -26,9 +24,11 @@ const keywords = [_][]const u8{
     "credit_account_id",
     "amount",
 };
+const completion_entries = keywords.len;
+const completion_entry_bytes = 512;
 
 pub const Completion = struct {
-    const CompletionList = RingBuffer([completion_entries]u8, .slice);
+    const CompletionList = RingBuffer([completion_entry_bytes]u8, .{ .array = completion_entries });
 
     matches: CompletionList,
     prefix: BoundedArray(u8, completion_entry_bytes),
@@ -36,9 +36,9 @@ pub const Completion = struct {
     query: BoundedArray(u8, completion_entry_bytes),
     match_index: usize = 0,
 
-    pub fn init(self: *Completion, allocator: std.mem.Allocator) !void {
+    pub fn init(self: *Completion) !void {
         self.* = Completion{
-            .matches = try CompletionList.init(allocator, completion_entry_bytes),
+            .matches = CompletionList.init(),
             .prefix = BoundedArray(u8, completion_entry_bytes){},
             .suffix = BoundedArray(u8, completion_entry_bytes){},
             .query = BoundedArray(u8, completion_entry_bytes){},
@@ -77,9 +77,9 @@ pub const Completion = struct {
     /// returns the query.
     pub fn get_next_completion(self: *Completion) ![]const u8 {
         if (self.matches.count > 0) {
-            const macth_ptr = self.matches.get_ptr(self.match_index).?;
+            const match_ptr = self.matches.get_ptr(self.match_index).?;
             self.match_index = (self.match_index + 1) % self.matches.count;
-            return std.mem.sliceTo(macth_ptr, '\x00');
+            return std.mem.sliceTo(match_ptr, '\x00');
         } else {
             return self.query.const_slice();
         }
--- a/src/completion.zig
+++ b/src/completion.zig
@@ -1,0 +1,201 @@
+const std = @import("std");
+const stdx = @import("stdx.zig");
+const assert = std.debug.assert;
+
+const RingBuffer = @import("ring_buffer.zig").RingBufferType;
+const BoundedArray = stdx.BoundedArrayType;
+
+const completion_entries = 256;
+const completion_entry_bytes = 512;
+const keywords = [_][]const u8{
+    "create_accounts",
+    "create_transfers",
+    "lookup_accounts",
+    "lookup_transfers",
+    "get_account_transfers",
+    "get_account_balances",
+    "query_accounts",
+    "query_transfers",
+    "help",
+    "id",
+    "code",
+    "ledger",
+    "flags",
+    "account_id",
+    "debit_account_id",
+    "credit_account_id",
+    "amount",
+};
+
+pub const Completion = struct {
+    const CompletionList = RingBuffer([completion_entries]u8, .slice);
+
+    matches: CompletionList,
+    prefix: BoundedArray(u8, completion_entry_bytes),
+    suffix: BoundedArray(u8, completion_entry_bytes),
+    query: BoundedArray(u8, completion_entry_bytes),
+    match_index: usize = 0,
+
+    pub fn init(self: *Completion, allocator: std.mem.Allocator) !void {
+        self.* = .{
+            .matches = try CompletionList.init(allocator, completion_entry_bytes),
+            .prefix = BoundedArray(u8, completion_entry_bytes){},
+            .suffix = BoundedArray(u8, completion_entry_bytes){},
+            .query = BoundedArray(u8, completion_entry_bytes){},
+        };
+    }
+
+    // Split the buffer into 3 parts -
+    // prefix: Everything before the word being completed.
+    // query : Partial word that needs completion, from last whitespace to buffer index.
+    // suffix: Everything after the cursor position.
+
+    /// Split the input buffer into prefix, query and suffix. Then tries to find the completions
+    /// matching the current query.
+    pub fn split_and_complete(self: *Completion, buffer: []const u8, buffer_index: usize) !void {
+        var query_start_index = buffer_index;
+
+        while (query_start_index > 0 and !std.ascii.isWhitespace(buffer[query_start_index - 1])) {
+            query_start_index -= 1;
+        }
+
+        self.prefix.clear();
+        self.prefix.append_slice_assume_capacity(buffer[0..query_start_index]);
+
+        self.suffix.clear();
+        self.suffix.append_slice_assume_capacity(buffer[buffer_index..]);
+
+        if (self.matches.count == 0) {
+            self.query.clear();
+            self.query.append_slice_assume_capacity(buffer[query_start_index..buffer_index]);
+
+            try self.get_completions(self.query.const_slice());
+        }
+    }
+
+    /// Returns the next match in the completion list. If the completion list is empty,
+    /// returns the query.
+    pub fn get_next_completion(self: *Completion) ![]const u8 {
+        if (self.matches.count > 0) {
+            const macth_ptr = self.matches.get_ptr(self.match_index).?;
+            self.match_index = (self.match_index + 1) % self.matches.count;
+            return std.mem.sliceTo(macth_ptr, '\x00');
+        } else {
+            return self.query.const_slice();
+        }
+    }
+
+    /// Reset the completion list.
+    pub fn clear(self: *Completion) void {
+        self.matches.clear();
+        self.prefix.clear();
+        self.suffix.clear();
+        self.query.clear();
+        self.match_index = 0;
+    }
+
+    /// Returns number of items in the completion list.
+    pub fn count(self: *Completion) usize {
+        return self.matches.count;
+    }
+
+    fn get_completions(self: *Completion, target: []const u8) !void {
+        if (std.mem.eql(u8, target, "")) {
+            return;
+        }
+
+        for (keywords) |kw| {
+            if (std.mem.startsWith(u8, kw, target)) {
+                assert(kw.len < completion_entry_bytes - 1);
+
+                if (self.matches.full()) {
+                    self.matches.advance_head();
+                }
+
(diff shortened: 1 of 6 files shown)


raghunandanbhat:
@brson, please review this PR when you have a chance. thanks!

brson:
@raghunandanbhat Looking at it now.

brson:
FYI we need to not introduce any new dynamic allocation to the repl (we can allocate during init, but not during execution - that's tigerstyle), so I am attempting to (quickly...) remove the dynamic allocation from https://github.com/tigerbeetle/tigerbeetle/pull/2155 to get that landed and unblock you @raghunandanbhat. It will almost certainly be a merge conflict for you.

My work on that is https://github.com/brson/tigerbeetle/tree/interactive-repl-improvements-static. I think I can submit it today, and it will require another round of review from another tigerbeetle team member.

In the meantime we can still make progress on the review here.

brson (changes requested):
@raghunandanbhat I have finally given this a close read and some testing. Happy in general. I'll make some overall observations and requests here, then leave some minor comments inline.

I am very glad you documented what your ANSI escape codes do, at least in most cases. I think this is something that could be very tough for future maintainers to get a grasp on, so anything we can do to make it clear what all this near-gibberish does is great. I don't want to make any changes to how the escape codes are documented now, but maybe we can think about how to ensure every use of ANSI codes is understandable in future work.

Recall that we need to not create dynamic allocations in new repl code. Here are there are quite a few but the transformation to make them non-dynamic should be straightforward.

I previously linked you to my branch that removes allocations from @wpaulino's prerequisite PR: https://github.com/brson/tigerbeetle/tree/interactive-repl-improvements-static. At this point that branch is essentially ready to go, and I'll submit it as its own PR later tonight. You can probably start working to rebase off of that, because it will definitely require some changes.

In general, to convert the dynamic ArrayList allocations to static, we can just convert those ArrayLists to fields on the repl and allocate their maximum capacity in the repl's `run` function (normally in tigerbeetle we do this in `init`, but the repl's type is different - it's created in `run`; then when they appended to we need to use `appendCapacity` etc. so they don't grow. Picking the size of these in most cases should not be too hard - in my branch their are constants defining the maximum command size (`repl_history_entry_bytes_without_nul` I think), so your three buffers - `match_prefix`, `match_query`, `match_suffix` - could potentially all have a capacity of `repl_history_entry_bytes_without_nul`. These dynamic allocations are something we have to fix before landing.

This PR adds new local variables to the `read_until_newline_or_eof_alloc` function, which is already probably the densest function in the repl, and after removing dynamic allocations, those ArrayLists will end up being members of `ReplType`. It seems like we might be able to encapsulate all the completion state into another struct, perhaps moving some of the completion logic out of the big `read_until_newline...` function into methods on a completion type. Please give that some thought and see if it makes sense to you. I'm not determined to make this change now, but I think it could be a readability improvement.

In testing, I do see some bugs in how things are displayed in the terminal, particularly in corner cases where the input spans multiple lines. Unfortunately, I'm not prepared to describe them in detail right now, but I definitely will.

src/repl.zig
```diff
@@ -583,6 +622,122 @@ pub fn ReplType(comptime MessageBus: type) type {
...
+                            // \x1b[0m - Reset the style property of text. In this case, resets
+                            //           the effect of \x1b[7m - color inversion.
+                            if (std.mem.eql(u8, match, current_match)) {
+                                try repl.terminal.print("\x20\x1b[7m{s}\x1b[0m\x20", .{match});
+                            } else {
+                                try repl.terminal.print("\x20{s}\x20", .{match});
+                            }
+                        }
+
+                        // If the completion menu overflow past the last row of terminal screen,
```
brson: `overflow` -> `overflows`

raghunandanbhat:
Thanks for the feedback @brson. 
I have already made some progress on making the code to use static allocations, based on the changes in your branch. I will have it ready for review once your PR lands.

---

> This PR adds new local variables to the `read_until_newline_or_eof_alloc` function, which is already probably the densest function in the repl, and after removing dynamic allocations, those ArrayLists will end up being members of `ReplType`. It seems like we might be able to encapsulate all the completion state into another struct, perhaps moving some of the completion logic out of the big `read_until_newline...` function into methods on a completion type. Please give that some thought and see if it makes sense to you. I'm not determined to make this change now, but I think it could be a readability improvement.
> 

I’ll look into moving the completion logic into a separate struct. My main concern is that much of the code is focused on cursor position adjustments, so I’m unsure how much readability will improve. I’ll see if there’s a way to balance clarity without adding extra back-and-forth for readers trying to follow the logic.

> In testing, I do see some bugs in how things are displayed in the terminal, particularly in corner cases where the input spans multiple lines. Unfortunately, I'm not prepared to describe them in detail right now, but I definitely will.

I have seen this bug when the input overflows to next line and I think I have fixed it. Would still like to hear your feedback once you have the chance to test it.

src/repl.zig
raghunandanbhat: this comment is fixed. if you see anything else, please let me know.

raghunandanbhat:
@brson - I made changes after your review, please review and let me know your feedback.

I moved completion state to a new Completion struct with static allocations - 
- a `RingBuffer` to hold  `256` completion entries, each entry of length `512`. 
- `stdx.BoundedArray` for the three buffers- `prefix`, `query` and `suffix`.

>In testing, I do see some bugs in how things are displayed in the terminal, particularly in corner cases where the input spans multiple lines. Unfortunately, I'm not prepared to describe them in detail right now, but I definitely will.

I've fixed this bug. Please retest this part and let me know if you see any issues. 

Also, if you have any suggestion to make the ANSI escape codes more readable, please let me know.

brson:
The prerequisite pr has now been merged: https://github.com/tigerbeetle/tigerbeetle/pull/2464

Thanks for your patience.

raghunandanbhat:
@brson please review the changes when you have a chance. thank you.

src/repl.zig
```diff
@@ -770,10 +729,10 @@ pub fn ReplType(comptime MessageBus: type) type {
                         // Re-fill the buffer with prefix, current match and suffix. Set the
                         // buffer index where the current selected match ends.
                         repl.buffer.clearRetainingCapacity();
-                        try repl.buffer.appendSlice(match_prefix.items);
-                        try repl.buffer.appendSlice(current_match);
-                        try repl.buffer.appendSlice(match_suffix.items);
-                        buffer_index = repl.buffer.items.len - match_suffix.items.len;
+                        try repl.buffer.appendSlice(repl.completion.prefix.slice());
+                        try repl.buffer.appendSlice(current_completion);
+                        try repl.buffer.appendSlice(repl.completion.suffix.slice());
```
brson: These three `appendSlice` calls need to be `appendSliceAssumeCapacity` so that we guarantee not to reallocate `buffer`.

That will panic if we do push too many items, so we need some kind of guarantee here and assertion, which is a bit tricky.

Today we attempt to guarantee not to overflow this buffer at a singe location in the `.printable` branch:

```
                        if (repl.buffer.items.len == repl.buffer.capacity) {
                            continue;
                        }
```

but here we need to reason about how to keep the length of the prefix, current_completion, and suffix all within a fixed buffer size.

Today `buffer` has size `repl_history_entry_bytes_without_nul` since that's the maximum size of a history entry, and if we receive a printable character from input we just refuse to accept it. After this patch the user could backspace into the middle of a line, press tab, and instantly insert `completion_bytes_max - 1` bytes.

I cannot think of a foolproof way to ensure there is always room for `completion_bytes_max - 1` bytes available in this buffer, so maybe we just have to accept that sometimes we can not do the completion.

A simple thing to do here would be to sum the length of the prefix, completion, and suffix, and if that is longer than `repl_history_entry_bytes_without_nul` just quietly don't use the completion and instead just keep using the original query string. It's a surprising result for the user but it's a corner case involving large inputs so hopefully won't be hit much.

Since the terminal updates in this `.tab` switch arm all depend on the length of `current_completion`, we need to do this overflow update early, and can probably just call `continue` to quit the `.tab` case early.

brson:
Some of the long-line artifacts are resolved now, but I still find one obvious one when the terminal width is smaller than the completion menu:

tb-completion-bug

brson:
I am finding that putting the completion state in its own type is making the main logic of the complex `read_until_newline_or_eof` method easier to follow, but have not given completion.zig a thorough read yet.

raghunandanbhat:
> Some of the long-line artifacts are resolved now, but I still find one obvious one when the terminal width is smaller than the completion menu:

could you please tell me if you're resizing the terminal window in the middle of command entry while window size is smaller than completion menu. If so, this is expected because terminal size is refreshed only after ctrl+c/newline, not for every keystroke. This was also mentioned in the [REPL enhancement PR](https://github.com/tigerbeetle/tigerbeetle/pull/2155#issuecomment-2318876636) and Python repl also behaves the same.

If you resize the window, then hit `\n` and then completions menu toggle should work. Please provide steps to reproduce this if you still see this issue.

brson:
> could you please tell me if you're resizing the terminal window in the middle of command entry while window size is smaller than completion menu. If so, this is expected because terminal size is refreshed only after ctrl+c/newline, not for every keystroke. This was also mentioned in the [REPL enhancement PR](https://github.com/tigerbeetle/tigerbeetle/pull/2155#issuecomment-2318876636) and Python repl also behaves the same.
> 
> If you resize the window, then hit `\n` and then completions menu toggle should work. Please provide steps to reproduce this if you still see this issue.

ok, yes it is as you describe - from an empty prompt i am resizing the window, then typing "c-tab-tab-tab" etc produces artifacts. surprising, but this is because the window size is calculated when the prompt is first displayed, then i resized the window.

So I agree there's nothing to fix about that, at least for now, since python etc. has similar behavior wrt window resizing.

src/completion.zig
```diff
@@ -0,0 +1,201 @@
...
+pub const Completion = struct {
+    const CompletionList = RingBuffer([completion_entries]u8, .slice);
+
+    matches: CompletionList,
+    prefix: BoundedArray(u8, completion_entry_bytes),
+    suffix: BoundedArray(u8, completion_entry_bytes),
+    query: BoundedArray(u8, completion_entry_bytes),
+    match_index: usize = 0,
+
+    pub fn init(self: *Completion, allocator: std.mem.Allocator) !void {
```
brson: Please add a corresponding deinit method to free the completion list, and call it from the repl (in a `defer` in `run` I think).

In https://github.com/tigerbeetle/tigerbeetle/pull/2542 the repl is being changed to be less sloppy about memory management, and we need to start explicitly freeing.

You might want to give that PR a look because it will probably cause conflicts with this one.

src/completion.zig
```diff
@@ -0,0 +1,201 @@
...
+
+            try self.get_completions(self.query.const_slice());
+        }
+    }
+
+    /// Returns the next match in the completion list. If the completion list is empty,
+    /// returns the query.
+    pub fn get_next_completion(self: *Completion) ![]const u8 {
+        if (self.matches.count > 0) {
+            const macth_ptr = self.matches.get_ptr(self.match_index).?;
```
brson: macth_ptr -> match_ptr

src/completion.zig
```diff
@@ -0,0 +1,201 @@
...
+    "ledger",
+    "flags",
+    "account_id",
+    "debit_account_id",
+    "credit_account_id",
+    "amount",
+};
+
+pub const Completion = struct {
+    const CompletionList = RingBuffer([completion_entries]u8, .slice);
```
brson: I _think_ this should be `RingBuffer([completion_entry_bytes]u8 ...`, not `[completion_entries]u8` - the type of element in the ringbuffer is a completion entry.

Likewise, I _think_ the wrong constant is used for `CompletionList.init` in `Completion.init` - there we should be asking for the number of elements - `completion_entries`.

src/completion.zig
```diff
@@ -0,0 +1,201 @@
...
+
+    matches: CompletionList,
+    prefix: BoundedArray(u8, completion_entry_bytes),
+    suffix: BoundedArray(u8, completion_entry_bytes),
+    query: BoundedArray(u8, completion_entry_bytes),
+    match_index: usize = 0,
+
+    pub fn init(self: *Completion, allocator: std.mem.Allocator) !void {
+        self.* = Completion{
+            .matches = try CompletionList.init(allocator, completion_entry_bytes),
```
brson: As mentioned above, I think this constant should be `completion_entries`, not `completion_entry_bytes`.

src/completion.zig
```diff
@@ -0,0 +1,201 @@
+const std = @import("std");
+const stdx = @import("stdx.zig");
+const assert = std.debug.assert;
+
+const RingBuffer = @import("ring_buffer.zig").RingBufferType;
+const BoundedArray = stdx.BoundedArrayType;
+
+const completion_entries = 256;
```
brson: This constant seems high to me - isn't the maximum completions, at least currently, limited by the number of keywords? Does it make sense to set this exactly to `keywords.len`?

raghunandanbhat: yes, 256 is high. setting it to `completion_entries = keywords.len` makes sense.

src/completion.zig
raghunandanbhat: fixed

src/completion.zig
raghunandanbhat: modified RingBuffer into this-
```zig
const CompletionList = RingBuffer([completion_entry_bytes]u8, .{ .array = completion_entries });


pub fn init(self: *Completion) !void {
        self.* = Completion{
            .matches = CompletionList.init(),
           }
    }
```
this way memory is allocated statically and no need to call deinit.

src/completion.zig
raghunandanbhat: fixed

src/repl.zig
raghunandanbhat: Thanks for pointing out this issue. I have added a check early in the tab completion code to skip the current completion if adding it would overflow the buffer.

raghunandanbhat:
@brson, I have addressed all of your review comments and made the necessary changes. please review.

brson (dismissed):

brson:
Looks great. Thanks @raghunandanbhat!

brson (approved):

Issue #2418, opened 2024-10-19
Running with --development doesnt print error message on panic


saltedpotatos:
I was updating my binary from 0.15.5 to 0.16.9 and re launched the binary  to see a panic

`./tigerbeetle start --addresses=3000 --development 1_1.tigerbeetle`
```sh
info(io): opening "1_1.tigerbeetle"...
warning(main): Grid cache size of 32MiB is small. See --cache-grid
info(main): multiversioning: disabled due to --development.
info(main): release=0.16.9
info(main): release_client_min=0.15.3
info(main): releases_bundled={ 0.16.9 }
info(main): git_commit=a082ff0237d5083f35e70d56c550ce271b6e4bf7
info(replica): superblock release=0.15.5
info(replica): 0: release_transition: release=0.16.9..0.15.5 (reason=open)
thread 3511012 panic: replica_release_execute unsupported
Unable to dump stack trace: debug info stripped
[1]    21282 abort      ./tigerbeetle start --addresses=3000 --development 1_1.tigerbeetle
```

but starting without the --development flag prints out an error message

`./tigerbeetle start --addresses=3000 1_1.tigerbeetle`
```sh
info(io): opening "1_1.tigerbeetle"...
info(main): release=0.16.9
info(main): release_client_min=0.15.3
info(main): releases_bundled={ 0.16.2, 0.16.3, 0.16.4, 0.16.8, 0.16.9 }
info(main): git_commit=a082ff0237d5083f35e70d56c550ce271b6e4bf7
info(replica): superblock release=0.15.5
info(replica): 0: release_transition: release=0.16.9..0.15.5 (reason=open)
error(main): 0: release_execute: release 0.15.5 is not available; upgrade (or downgrade) the binary
thread 3514491 panic: release not available
Unable to dump stack trace: debug info stripped
[1]    21540 abort      ./tigerbeetle start --addresses=3000 1_1.tigerbeetle
```

I haven't looked at the source yet, but I'm assuming the error message should also show when started with --development?

sentientwaffle:
The actual error that occurs in this case differs depending on whether `--development` is set.

## With `--development`

When `--development` is set, multiversion binary auto-upgrades are disabled:

```
info(main): multiversioning: disabled due to --development.
...
info(replica): 0: release_transition: release=0.16.9..0.15.5 (reason=open)
thread 3511012 panic: replica_release_execute unsupported
```

...So to upgrade you would need to:

1. Stop the old binary.
[...]

PR #2425, opened 2024-10-25
Multiversion: Simplify upgrade with dev flags

sentientwaffle:
## Overview

When `--development` or `--experimental` is passed to `tigerbeetle start`, upgrading is more complicated than usual:

1. Stop the replica (old binary).
2. Start the replica (new binary) without the offending flag.
3. Wait for the upgrade to complete.
4. Stop the replica (new binary).
5. Start the replica (new binary) with the extra flags.

In this PR, simplify it to:

1. Stop the replica (old binary).
2. Start the replica (new binary) with the extra flags.

... i.e. when `--development`/`--experimental` are set, we still don't monitor the binary for changes, but we _do_ support upgrades.

Also in this commit, clarify some of the `log.info` messages.

## Details

Note that the previous logic:

```zig
if (args.development) {
    log.info("multiversioning: disabled due to --development.", .{});
    break :blk null;
}
if (args.experimental) {
    log.info("multiversioning: disabled due to --experimental.", .{});
    break :blk null;
}
```

... still exists in the bundled versions, so upgrades with `--development`/`--experiment` will only work as those bundled versions get phased out, replaced by binaries with the new logic.

To test this locally, I first built an "old" version with the new logic, then bundled that "old" version with the "new" binary, and tested an upgrade. (See the change in `build.zig`, which simplifies this process by adding a `-Dmultiversion-file` flag).

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -90,7 +90,12 @@ pub fn build(b: *std.Build) !void {
         .multiversion = b.option(
             []const u8,
             "multiversion",
-            "Past version to include for upgrades",
+            "Past version to include for upgrades (\"latest\" or \"x.y.z\")",
+        ),
+        .multiversion_file = b.option(
+            []const u8,
+            "multiversion-file",
+            "Past version to include for upgrades (local binary file)",
         ),
         .config_verify = b.option(bool, "config_verify", "Enable extra assertions.") orelse true,
         .config_aof_recovery = b.option(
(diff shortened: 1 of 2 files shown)


sentientwaffle:
This fix was inspired by https://github.com/tigerbeetle/tigerbeetle/issues/2418

cb22 (approved):

PR #2427, opened 2024-10-26
ci: allow git tag and config.process.release to differ

matklad:
This is useful for rare hot-fix releases, where we need to update the version of the code, but we don't want to run the upgrade through the protocol!

As a bonus, the `--build` part of the release script now works on macOS.

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -1515,6 +1515,7 @@ fn fetch(b: *std.Build, options: struct {
     const copy_from_cache = b.addRunArtifact(b.addExecutable(.{
         .name = "copy-from-cache",
         .root_source_file = b.addWriteFiles().add("main.zig",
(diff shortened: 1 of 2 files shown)


build.zig
```diff
@@ -1541,7 +1542,8 @@ fn fetch(b: *std.Build, options: struct {
             \\        source_path,
             \\        std.fs.cwd(),
             \\        args[4],
-            \\        .{},
+            \\        // TODO(Zig): https://github.com/ziglang/zig/pull/21555
```
cb22, line 1545: :smile:

cb22 (approved):
LGTM - but perhaps we should do a trial run of this somehow once merged?

matklad:
Yeah, I don’t see a good way to test this :(

PR #2428, opened 2024-10-26
grid: make reserve() panic rather than returning null

cb22:
Currently all uses of `grid.reserve()` unwrap the result. Cut out the middleman, and panic directly in `reserve()` with a nice error message.

### DIFF
--- a/src/vsr/grid.zig
+++ b/src/vsr/grid.zig
@@ -465,10 +465,21 @@ pub fn GridType(comptime Storage: type) type {
             grid.superblock.storage.on_next_tick(.lsm, callback, next_tick);
         }
 
-        /// Returning null indicates that there are not enough free blocks to fill the reservation.
-        pub fn reserve(grid: *Grid, blocks_count: usize) ?Reservation {
+        /// Aborts if there are not enough free blocks to fill the reservation.
+        /// Should a use case arise where a null return would be preferred, this can be split
+        /// into panicking and non-panicking versions.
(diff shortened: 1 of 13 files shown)


src/vsr/superblock.zig
```diff
@@ -1276,7 +1276,7 @@ pub fn SuperBlockType(comptime Storage: type) type {
                     vsr.fatal(
                         .storage_size_exceeds_limit,
                         "data file too large size={} > limit={}, " ++
-                            "restart the replica increasing '--storage_size_limit'",
+                            "restart the replica increasing '--storage-size-limit'",
```
matklad, line 1279: Good catch on the `-` vs `_`!

src/vsr/grid.zig
```diff
@@ -465,10 +465,21 @@ pub fn GridType(comptime Storage: type) type {
             grid.superblock.storage.on_next_tick(.lsm, callback, next_tick);
         }
 
-        /// Returning null indicates that there are not enough free blocks to fill the reservation.
-        pub fn reserve(grid: *Grid, blocks_count: usize) ?Reservation {
+        /// Panics if there are not enough free blocks to fill the reservation.
```
matklad:
```suggestion
        /// Aborts if there are not enough free blocks to fill the reservation.
```

we are not actually using Zig's panic mechanism here

matklad (dismissed):
This looks good to me, but I have a nagging feeling that maybe I am missing something about our overall desired behavior here? Started a Slack discussion about that.

matklad (approved):

PR #2430, opened 2024-10-28
IO/Linux: Cancel IO before freeing memory

sentientwaffle:
### IO

Cancel in-flight IO before freeing memory.

More specifically: If IO has been submitted to io_uring, we must be careful not to free any buffers referenced by that IO until it is complete. Otherwise we might write to that memory after it has been freed.

`IO` implements this by maintaining a new doubly-linked list of in-flight IO. When deinitializing IO, we cancel those IO's one at a time, and wait for each to complete (where "complete" might mean failing with a `ECANCELED` result).

```
// Before freeing any memory, wait for in-flight IO to finish:
try io.cancel();

// Now that there is no running IO, and no more IO is allowed to start running, we can safely free
// all of our buffers:
message_bus.deinit();
io.deinit();
```

If in the future we bump our minimum supported Linux kernel version to at least 5.19, then this logic can be massively simplified by taking advantage of the `IORING_ASYNC_CANCEL_ALL` and `IORING_ASYNC_CANCEL_ANY` flags to cancel all IO with a single operation. And if we bump to 6.0, we can additionally use `io_uring`, simplifying our `IO.cancel()` implementation even farther.

<details>
<summary>Other approaches</summary>

Another approach I tried was adding a `io.cancel(completion)` function which consumers of IO were expected to call during their own `deinit()`.

But this had several drawbacks:

- `io.cancel` has to wait for IO to finish, which means it needs to `io.tick`. And `io.tick()` can fail, which means many `deinit()` functions can return an error.
- We need to increase the surface area of `Storage` since consumers of `Storage` don't know about `IO`, e.g. `storage.cancel_read()`, `storage.cancel_write()`.
- While canceling an operation, we tick IO, which can make progress on other IO, calling their callbacks, which is hard to reason about. In the implemented version, cancellation is all-or-nothing -- once we start cancellation, we can just not ever invoke a completion callback again.

</details>

---

### MessageBus

Fix `MessageBus.deinit()` assertion failure caused by invoking `connection.terminate()` twice (for the same connection).

Stack trace of assert (via Vortex!):

```

thread 17 panic: reached unreachable code
/home/owi/projects/tigerbeetle/tigerbeetle/src/message_bus.zig:0:0: 0x10796a0 in terminate (vortex)
/home/owi/projects/tigerbeetle/tigerbeetle/src/message_bus.zig:182:37: 0x106bed5 in deinit (vortex)
                connection.terminate(bus, .shutdown);
                                    ^
/home/owi/projects/tigerbeetle/tigerbeetle/src/clients/c/tb_client/context.zig:231:31: 0x106b920 in on_deinit (vortex)
            self.client.deinit(self.allocator);
                              ^
```

### DIFF
--- a/src/io/linux.zig
+++ b/src/io/linux.zig
@@ -27,8 +27,8 @@ pub const IO = struct {
     completed: FIFO(Completion) = .{ .name = "io_completed" },
 
     // TODO Track these as metrics:
-    ios_queued: u64 = 0,
-    ios_in_kernel: u64 = 0,
+    ios_queued: u32 = 0,
+    ios_in_kernel: u32 = 0,
 
     /// The head of a doubly-linked list of all operations that are:
     /// - in the submission queue, or
@@ -38,7 +38,7 @@ pub const IO = struct {
     awaiting: ?*Completion = null,
     /// The number of Completions in the `awaiting` doubly-linked list.
     /// This is only used for safety checks.
-    awaiting_count: u64 = 0,
+    awaiting_count: u32 = 0,
 
     // This is the completion that performs the cancellation.
     // This is *not* the completion that is being canceled.
@@ -153,7 +153,7 @@ pub const IO = struct {
     }
 
     fn flush(self: *IO, wait_nr: u32, timeouts: *usize, etime: *bool) !void {
-        if (constants.verify) self.awaiting_verify();
+        if (constants.verify) self.verify_awaiting(null);
 
         // Flush any queued SQEs and reuse the same syscall to wait for completions if required:
         try self.flush_submissions(wait_nr, timeouts, etime);
@@ -284,7 +284,7 @@ pub const IO = struct {
     }
 
     fn enqueue(self: *IO, completion: *Completion) void {
-        if (constants.verify) self.awaiting_verify();
+        if (constants.verify) self.verify_awaiting(completion);
 
         switch (self.cancel_status) {
             .inactive => {},
@@ -333,11 +333,11 @@ pub const IO = struct {
     ///   cancellation stage.
     /// - Linux kernel ≥6.0 supports `io_uring_register_sync_cancel` which would remove the `queued`
     ///   cancellation stage.
-    pub fn cancel(self: *IO) !void {
+    pub fn cancel_all(self: *IO) void {
         assert(self.cancel_status == .inactive);
-        self.awaiting_verify();
+        self.verify_awaiting(null);
 
-        // Ensure that even if we return early due to an io_uring that IO won't accept more IO.
+        // Even if we return early due to an io_uring error, IO won't allow more operations.
         defer self.cancel_status = .done;
 
         self.cancel_status = .next;
@@ -346,34 +346,17 @@ pub const IO = struct {
         while (self.unqueued.pop()) |_| {}
 
         while (self.awaiting) |target| {
+            assert(self.cancel_status == .next);
             assert(self.awaiting_count > 0);
             assert(target.operation != .cancel);
 
-            if (self.cancel_status == .next) {
-                self.cancel_completion = .{
-                    .io = self,
-                    .context = self,
-                    .callback = struct {
-                        fn wrapper(
-                            ctx: ?*anyopaque,
-                            comp: *Completion,
-                            res: *const anyopaque,
-                        ) void {
-                            const io: *IO = @ptrCast(@alignCast(ctx.?));
-                            const result =
-                                @as(*const CancelError!void, @ptrCast(@alignCast(res))).*;
-                            io.cancel_callback(comp, result);
-                        }
-                    }.wrapper,
-                    .operation = .{ .cancel = .{ .target = target } },
-                };
-
-                self.cancel_status = .{ .queued = .{ .target = target } };
-                self.enqueue(&self.cancel_completion);
-            }
+            self.cancel(target);
+            assert(self.cancel_status == .queued);
 
             while (self.cancel_status == .queued or self.cancel_status == .wait) {
-                try self.run_for_ns(constants.tick_ms * std.time.ns_per_ms);
+                self.run_for_ns(constants.tick_ms * std.time.ns_per_ms) catch |err| {
+                    std.debug.panic("IO.cancel_all: run_for_ns error: {}", .{err});
+                };
             }
             assert(self.cancel_status == .next);
         }
@@ -382,10 +365,33 @@ pub const IO = struct {
         assert(self.ios_in_kernel == 0);
     }
 
+    fn cancel(self: *IO, target: *Completion) void {
+        self.cancel_completion = .{
+            .io = self,
+            .context = self,
+            .callback = struct {
+                fn wrapper(
+                    ctx: ?*anyopaque,
+                    comp: *Completion,
+                    res: *const anyopaque,
+                ) void {
+                    const io: *IO = @ptrCast(@alignCast(ctx.?));
+                    const result =
+                        @as(*const CancelError!void, @ptrCast(@alignCast(res))).*;
+                    io.cancel_callback(comp, result);
+                }
+            }.wrapper,
+            .operation = .{ .cancel = .{ .target = target } },
+        };
+
+        self.cancel_status = .{ .queued = .{ .target = target } };
+        self.enqueue(&self.cancel_completion);
+    }
+
     const CancelError = error{
         NotRunning,
         NotInterruptable,
-    };
+    } || posix.UnexpectedError;
 
     fn cancel_callback(self: *IO, completion: *Completion, result: CancelError!void) void {
         assert(self.cancel_status == .queued);
@@ -397,23 +403,26 @@ pub const IO = struct {
             result catch |err| switch (err) {
                 error.NotRunning => break :status .next,
                 error.NotInterruptable => {},
+                error.Unexpected => unreachable,
             };
             // Wait for the target operation to complete or abort.
             break :status .{ .wait = .{ .target = self.cancel_status.queued.target } };
         };
     }
 
-    fn awaiting_verify(self: *const IO) void {
+    fn verify_awaiting(self: *const IO, exclude: ?*const Completion) void {
         var awaiting_count: u32 = 0;
         var awaiting = self.awaiting;
 
         if (awaiting) |completion| {
             assert(completion.awaiting_previous == null);
+            if (exclude) |c| assert(completion != c);
         }
 
         while (awaiting) |completion| {
             const awaiting_next = completion.awaiting_next;
             if (awaiting_next) |next| {
+                if (exclude) |c| assert(next != c);
                 assert(next.awaiting_previous == completion);
             }
             awaiting_count += 1;
@@ -509,7 +518,7 @@ pub const IO = struct {
         fn complete(completion: *Completion) void {
             switch (completion.operation) {
                 .cancel => {
-                    const result: anyerror!void = result: {
+                    const result: CancelError!void = result: {
                         if (completion.result < 0) {
                             break :result switch (@as(posix.E, @enumFromInt(-completion.result))) {
                                 // No operation matching the completion is queued, so there is
@@ -527,7 +536,7 @@ pub const IO = struct {
                     completion.callback(completion.context, completion, &result);
                 },
                 .accept => {
-                    const result: anyerror!posix.socket_t = blk: {
+                    const result: AcceptError!posix.socket_t = blk: {
                         if (completion.result < 0) {
                             const err = switch (@as(posix.E, @enumFromInt(-completion.result))) {
                                 .INTR => {
@@ -536,7 +545,6 @@ pub const IO = struct {
                                 },
                                 .AGAIN => error.WouldBlock,
                                 .BADF => error.FileDescriptorInvalid,
-                                .CANCELED => error.Canceled,
                                 .CONNABORTED => error.ConnectionAborted,
                                 .FAULT => unreachable,
                                 .INVAL => error.SocketNotListening,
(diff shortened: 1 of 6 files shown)


src/io/test.zig
```diff
@@ -739,3 +739,101 @@ test "pipe data over socket" {
...
+            for (&read_completions, read_buffers) |*completion, buffer| {
+                context.io.read(*Context, &context, read_callback, completion, file, buffer, 0);
+            }
+            try context.io.tick();
+
+            // Set to true *before* calling cancel() to ensure that any farther callbacks from IO
+            // completion will panic.
+            context.canceled = true;
+
+            try context.io.cancel();
```
sentientwaffle: Note that if this line is removed, the test fails on a checksum mismatch (as expected!).

src/io/linux.zig
```diff
@@ -398,6 +403,25 @@ pub const IO = struct {
...
+        var awaiting_count: u32 = 0;
+        var awaiting = self.awaiting;
+
+        if (awaiting) |completion| {
+            assert(completion.awaiting_previous == null);
+        }
+
+        while (awaiting) |completion| {
+            const awaiting_next = completion.awaiting_next;
+            if (awaiting_next) |next| {
```
kprotty, line 424: Future note: would be nice to share this doubly-linked-list logic in fifo.zig or similar for `io/darwin.zig` and `io/windows.zig`

src/io/linux.zig
```diff
@@ -29,6 +30,33 @@ pub const IO = struct {
     ios_queued: u64 = 0,
     ios_in_kernel: u64 = 0,
 
+    /// The head of a doubly-linked list of all operations that are:
+    /// - in the submission queue, or
+    /// - in the kernel, or
+    /// - in the completion queue, or
+    /// - in the `completed` list (excluding zero-duration timeouts)
+    awaiting: ?*Completion = null,
```
kprotty, line 38: Is it simpler to have awaiting track only the second point (in the kernel)? Meaning 1) on SQE acquire in `enqueue` then insert to list 2) on CQE recv in `flush_completions` then remove from list.

src/io/linux.zig
```diff
@@ -215,9 +300,128 @@ pub const IO = struct {
...
+    /// - Or, an io_uring error occurs.
+    ///
+    /// TODO(Linux):
+    /// - Linux kernel ≥5.19 supports the IORING_ASYNC_CANCEL_ALL and IORING_ASYNC_CANCEL_ANY flags,
+    ///   which would allow all events to be cancelled simultaneously with a single "cancel"
+    ///   operation, without IO needing to maintain the `awaiting` doubly-linked list and the `next`
+    ///   cancellation stage.
+    /// - Linux kernel ≥6.0 supports `io_uring_register_sync_cancel` which would remove the `queued`
+    ///   cancellation stage.
+    pub fn cancel(self: *IO) !void {
```
kprotty: Would it make sense to have this implicitly in `deinit()` instead? Need to change the test case if so.

If not, think `main.zig`/`repl.zig`/other uses of `IO.init` need to cancel as well. Or at least assert `awaiting_count == 0` in `IO.deinit()`.

sentientwaffle: > Would it make sense to have this implicitly in `deinit()` instead? Need to change the test case if so.

By the time `io.deinit()` is being called, the buffers have already been freed, so it is too late.

> If not, think main.zig/repl.zig/other uses of IO.init need to cancel as well. Or at least assert awaiting_count == 0 in IO.deinit().

Since `main.zig` and `repl.zig` only exit when the process exits, I don't think we actually need to cancel IO in those cases.

sentientwaffle:
The integration tests had a flake in CI but I haven't managed reproduce it locally -- not sure if it is a new flake introduced by this PR or just chance.

src/io/linux.zig
kprotty: Ah, the buffer lifetimes are beginning to click. Thanks.

src/io/linux.zig
```diff
@@ -215,9 +300,128 @@ pub const IO = struct {
...
+        assert(self.cancel_status == .inactive);
+        self.awaiting_verify();
+
+        // Ensure that even if we return early due to an io_uring that IO won't accept more IO.
```
matklad: It feels like this misses a word? I parse this as `(if we return early (due to an io_uring))`

src/io/linux.zig
```diff
@@ -207,6 +284,14 @@ pub const IO = struct {
     }
 
     fn enqueue(self: *IO, completion: *Completion) void {
+        if (constants.verify) self.awaiting_verify();
```
matklad: Can we pass `completion` as an argument to `awaiting_verify` to make sure it is not enqueued already?

src/io/linux.zig
```diff
@@ -207,6 +284,14 @@ pub const IO = struct {
...
+
+        switch (self.cancel_status) {
+            .inactive => {},
+            .queued => assert(completion.operation == .cancel),
+            else => unreachable,
+        }
```
matklad, line 293: Excellent assert!

src/io/linux.zig
```diff
@@ -215,9 +300,128 @@ pub const IO = struct {
...
+        assert(self.cancel_status == .inactive);
+        self.awaiting_verify();
+
+        // Ensure that even if we return early due to an io_uring that IO won't accept more IO.
+        defer self.cancel_status = .done;
+
+        self.cancel_status = .next;
+
+        // Discard any operations that haven't started yet.
+        while (self.unqueued.pop()) |_| {}
```
matklad, line 346: Should we call the callbacks with `-ECANCEL` as the error code? That way, we guarantee that a callback is _always_ called, as opposed to only _sometimes_ getting called?

matklad: Ahh, I see it now, we actually arrange it so that the callback is _not_ called on cancellation. That is, if we call cancel, and at that time there's _already_ a completion in the completion queue (i.e., the operation actually fully finished before cancelation), the callback for that is not called. 

This seems reasonable, I think, to make sure that individual callbacks won't have to worry about the cancelation! And, if the callback holds some closure on the client side, I assume it's up to the client to properly clean it up!

src/io/linux.zig
```diff
@@ -215,9 +300,128 @@ pub const IO = struct {
...
+                            const result =
+                                @as(*const CancelError!void, @ptrCast(@alignCast(res))).*;
+                            io.cancel_callback(comp, result);
+                        }
+                    }.wrapper,
+                    .operation = .{ .cancel = .{ .target = target } },
+                };
+
+                self.cancel_status = .{ .queued = .{ .target = target } };
+                self.enqueue(&self.cancel_completion);
```
matklad: maybe extract this code into (non-pub for now) `fn cancel`, and rename current `cancel` to `cancel_all`?

src/io/linux.zig
```diff
@@ -215,9 +300,128 @@ pub const IO = struct {
...
+            }
+
+            while (self.cancel_status == .queued or self.cancel_status == .wait) {
+                try self.run_for_ns(constants.tick_ms * std.time.ns_per_ms);
```
matklad: This `try` is suspicious. If we bail out with `try` here, we actually could than get UB? Can we `catch unreachable` here? It seems that, for cancelation, we actually shouldn't be able to error here perhaps? Or maybe not --- there _could_ be some not-yet-submitted IOPS in front of our cancel request, and those can error on submission? 

But perhaps we can guarantee that we just never fail on submission?

src/io/linux.zig
matklad, line 38: Consider extracting the intrusive doubly-linked list into a separate data structure, a-la `FIFO`

matklad, line 38: Consider extracting the intrusive doubly-linked list into a separate data structure, a-la `FIFO`

src/io/linux.zig
```diff
@@ -215,9 +300,128 @@ pub const IO = struct {
...
+
+        while (awaiting) |completion| {
+            const awaiting_next = completion.awaiting_next;
+            if (awaiting_next) |next| {
+                assert(next.awaiting_previous == completion);
+            }
+            awaiting_count += 1;
+            awaiting = awaiting_next;
+        }
+        assert(awaiting_count == self.awaiting_count);
```
matklad, line 431: this compares `u32` to `u64`. Let's use `u32` everywhere?

src/io/linux.zig
```diff
@@ -503,6 +738,7 @@ pub const IO = struct {
                                 .OPNOTSUPP => error.OperationNotSupported,
                                 .PIPE => error.BrokenPipe,
                                 .TIMEDOUT => error.ConnectionTimedOut,
+                                .CANCELED => error.Canceled,
```
matklad: Wait, shouldn't this _remain_ unexpected? I think we are guaranteed to _not_ complete canceled completions? 

Ah, and there's also a clear bug here: we _don't_ add `Canceled` to `pub const ReadError = error{` and friends.  That is, what happens here is that we essentially downcast an `anyerror` to a well-typed error (by casting the function pointer), and that would be an invalid cast. 

🤔 


perhaps we should't type `const result: anyerror!usize` and rather use `const result: SendError!usize` here?

src/io/windows.zig
```diff
@@ -283,6 +283,11 @@ pub const IO = struct {
         }
     }
 
+    pub fn cancel(self: *IO) !void {
```
matklad:
```suggestion
    pub fn cancel(_: *IO) !void {
```

src/io/test.zig
```diff
@@ -739,3 +739,105 @@ test "pipe data over socket" {
...
+        fn run_test() !void {
+            defer std.fs.cwd().deleteFile(file_path) catch {};
+
+            var context: Context = .{ .io = try IO.init(32, 0) };
+            defer context.io.deinit();
+
+            {
+                // Initialize the file and fill it with test data.
+                const file = try std.fs.cwd().createFile(file_path, .{
+                    .read = true,
```
matklad: meant to be `.write = true`? And this could use `std.fs.cwd().writeFile` perhaps? 

As a rule of thumb, it's better to use bulk file operations if possible, where you don't create a file descriptor yourself and let the stdlib to handle  it for you.

src/io/linux.zig
sentientwaffle: Exactly! Once cancellation starts, we don't call any IO callbacks. This reduces the surface area added by cancellation -- message_bus/storage don't really need to concern themselves with it. This "all-or-nothing" approach to cancellation simplifies the mental model, no need to worry about callbacks starting up additional IO work.

src/io/linux.zig
sentientwaffle: There are a wide variety of reasons that `run_for_ns` could fail here, including:

```
        // The kernel was unable to allocate memory or ran out of resources for the request.
        // The application should wait for some completions and try again:
        .AGAIN => return error.SystemResources,
        // The SQE `fd` is invalid, or IOSQE_FIXED_FILE was set but no files were registered:
        .BADF => return error.FileDescriptorInvalid,
        // The file descriptor is valid, but the ring is not in the right state.
        // See io_uring_register(2) for how to enable the ring.
        .BADFD => return error.FileDescriptorInBadState,
        // The application attempted to overcommit the number of requests it can have pending.
        // The application should wait for some completions and try again:
        .BUSY => return error.CompletionQueueOvercommitted,
        // The SQE is invalid, or valid but the ring was setup with IORING_SETUP_IOPOLL:
        .INVAL => return error.SubmissionQueueEntryInvalid,
        // The buffer is outside the process' accessible address space, or IORING_OP_READ_FIXED
        // or IORING_OP_WRITE_FIXED was specified but no buffers were registered, or the range
        // described by `addr` and `len` is not within the buffer registered at `buf_index`:
        .FAULT => return error.BufferInvalid,
        .NXIO => return error.RingShuttingDown,
        // The kernel believes our `self.fd` does not refer to an io_uring instance,
        // or the opcode is valid but not supported by this kernel (more likely):
        .OPNOTSUPP => return error.OpcodeNotSupported,
        // The operation was interrupted by a delivery of a signal before it could complete.
        // This can happen while waiting for events with IORING_ENTER_GETEVENTS:
        .INTR => return error.SignalInterrupt,
        else => |errno| return posix.unexpectedErrno(errno),
```

... so the safest thing to do if we hit one is just to crash. That's what the replica does if it hits an error during its normal IO-tick-loop anyhow.

src/io/linux.zig
sentientwaffle: > Wait, shouldn't this remain unexpected? I think we are guaranteed to not complete canceled completions?

Good point!

> perhaps we should't type const result: anyerror!usize and rather use const result: SendError!usize here?

This found ~a bug~ several bugs!

src/io/linux.zig
sentientwaffle: I will extract it in a follow-up PR, but I don't want to delay this PR.

src/io/linux.zig
```diff
@@ -346,34 +346,17 @@ pub const IO = struct {
...
-                self.cancel_status = .{ .queued = .{ .target = target } };
-                self.enqueue(&self.cancel_completion);
-            }
+            self.cancel(target);
+            assert(self.cancel_status == .queued);
 
             while (self.cancel_status == .queued or self.cancel_status == .wait) {
-                try self.run_for_ns(constants.tick_ms * std.time.ns_per_ms);
+                self.run_for_ns(constants.tick_ms * std.time.ns_per_ms) catch |err| {
+                    std.debug.panic("IO.cancel_all: run_for_ns error: {}", err);
```
matklad: .{err}

sentientwaffle: Ah, thanks :facepalm:

matklad (approved):

PR #2432, opened 2024-10-29
vsr: assert free set consistency

chaitanyabhandari:
_On startup and after checkpoint_, we now assert that the free set is consistent. Specifically, we assert that the count of acquired blocks in the free set is the sum of:
1. Index blocks across all tables in the forest
2. Value blocks across all tables in the forest
3. ManifestLog blocks
4. CheckpointTrailer blocks (client sessions & free set)

### DIFF
--- a/src/lsm/manifest_log.zig
+++ b/src/lsm/manifest_log.zig
@@ -860,8 +860,8 @@ pub fn ManifestLogType(comptime Storage: type) type {
                 manifest_log.close_block();
                 assert(manifest_log.entry_count == 0);
                 assert(manifest_log.blocks_closed > 0);
-                assert(manifest_log.blocks_closed == manifest_log.blocks.count);
             }
+            assert(manifest_log.blocks_closed == manifest_log.blocks.count);
 
             manifest_log.flush(callback);
         }
--- a/src/lsm/tree.zig
+++ b/src/lsm/tree.zig
@@ -287,6 +287,11 @@ pub fn TreeType(comptime TreeTable: type, comptime Storage: type) type {
             return .negative;
         }
 
+        pub fn block_value_count_max(tree: *Tree) u32 {
+            _ = tree; // autofix
+            return Table.layout.block_value_count_max;
+        }
+
         fn cached_data_block_search(
             tree: *Tree,
             address: u64,
--- a/src/lsm/tree.zig
+++ b/src/lsm/tree.zig
@@ -287,8 +287,7 @@ pub fn TreeType(comptime TreeTable: type, comptime Storage: type) type {
             return .negative;
         }
 
-        pub fn block_value_count_max(tree: *Tree) u32 {
-            _ = tree; // autofix
+        pub fn block_value_count_max(_: *const Tree) u32 {
             return Table.layout.block_value_count_max;
         }
 
--- a/src/testing/cluster/storage_checker.zig
+++ b/src/testing/cluster/storage_checker.zig
@@ -164,12 +164,17 @@ pub const StorageChecker = struct {
         }
     }
 
-    pub fn replica_checkpoint(checker: *StorageChecker, superblock: *const SuperBlock) !void {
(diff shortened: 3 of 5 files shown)


src/lsm/tree.zig
```diff
@@ -287,6 +287,11 @@ pub fn TreeType(comptime TreeTable: type, comptime Storage: type) type {
             return .negative;
         }
 
+        pub fn block_value_count_max(tree: *Tree) u32 {
+            _ = tree; // autofix
```
matklad:
```suggestion
            _ = tree;
```

Or I think even `_: *Tree` could work?

src/lsm/manifest_log.zig
```diff
@@ -860,8 +860,8 @@ pub fn ManifestLogType(comptime Storage: type) type {
                 manifest_log.close_block();
                 assert(manifest_log.entry_count == 0);
                 assert(manifest_log.blocks_closed > 0);
-                assert(manifest_log.blocks_closed == manifest_log.blocks.count);
             }
+            assert(manifest_log.blocks_closed == manifest_log.blocks.count);
```
matklad, line 864: ❤️

src/testing/cluster/storage_checker.zig
```diff
@@ -164,12 +164,17 @@ pub const StorageChecker = struct {
         }
     }
 
-    pub fn replica_checkpoint(checker: *StorageChecker, superblock: *const SuperBlock) !void {
-        const syncing = superblock.working.vsr_state.sync_op_max > 0;
+    pub fn replica_checkpoint(
+        checker: *StorageChecker,
+        comptime Replica: type,
```
matklad, line 169: Hm, I think it was a nice property that storage checker _didn't_ know the type of the replica beforehand. This ensures that all verification we do here is independent of replica's code. 

Given that replica itself already runs this same code, let's maybe drop it from storage checker? 

If we do do this in the checker, we should have an independent implementation, which works off manifest log in the grid. I'd maybe say that on-line verification is the 90% here.

src/vsr/replica.zig
```diff
@@ -9795,6 +9796,42 @@ pub fn ReplicaType(
...
+            assert(self.grid.free_set.count_released() ==
+                self.grid.free_set_checkpoint.block_count() +
+                self.client_sessions_checkpoint.block_count());
+
+            // Must be invoked either on startup, or after checkpoint completes.
+            assert(!self.state_machine_opened or self.commit_stage == .checkpoint_superblock);
+
+            var forest_tables_iterator = ForestTableIterator{};
+            var tables_index_block_count: u32 = 0;
+            var tables_value_block_count: u32 = 0;
```
matklad: Go for `u64`, to be consistent with our `block_address`. In general, as a rule of thumb, anything storage-related should be counted in `u64`s.

src/testing/cluster/storage_checker.zig
chaitanyabhandari: Mhmmmmm I had initially kept this all within replica.zig, but was concerned that it may be too costly to do this at every checkpoint in production, especially since the number of tables in the forest increase exponentially in across LSM levels... Thoughts? Shall we only keep this on startup in replica.zig?

src/lsm/tree.zig
chaitanyabhandari: Ah, `_: *Tree` is great!

src/testing/cluster/storage_checker.zig
matklad: Ahhh, sorry, I missed the fact that we _don't_ check this in prod at checkpoint. Yeah, I think your are correct here, so lets leave this as is!

chaitanyabhandari:
Comments addressed in https://github.com/tigerbeetle/tigerbeetle/pull/2432/commits/737ded756a6f973f7ba793233e8da0b1d3b46fbe!

matklad (approved):

PR #2434, opened 2024-10-29
Vortex: check timestamp order for accounts and transfers

owickstrom:
By collecting at most 8190 of the prior successful transfers, possibly created in different requests, we can query those and check that their timestamps are indeed monotonically increasing.

Similarly, we check account timestamps when reconciling the existing query.

### DIFF
--- a/src/testing/vortex/workload.zig
+++ b/src/testing/vortex/workload.zig
@@ -30,6 +30,7 @@ const arbitrary = @import("./arbitrary.zig");
 const tb = @import("../../tigerbeetle.zig");
 const constants = @import("../../constants.zig");
 const StateMachineType = @import("../../state_machine.zig").StateMachineType;
+const RingBufferType = @import("../../ring_buffer.zig").RingBufferType;
 
 const log = std.log.scoped(.workload);
 const assert = std.debug.assert;
@@ -206,7 +207,12 @@ fn reconcile(result: Result, command: *const Command, model: *Model) !void {
                 try testing.expect(model.account_exists(transfer.credit_account_id));
             }
 
-            model.latest_transfers.append(successful_transfer_ids.const_slice());
+            // Drop the oldest transfer IDs and add new ones.
+            for (0..@min(successful_transfer_ids.count(), model.latest_transfers.count)) |_| {
+                model.latest_transfers.retreat_tail();
+            }
+            try model.latest_transfers.push_slice(successful_transfer_ids.const_slice());
+
             model.transfers_created += transfers.len;
         },
         .lookup_all_accounts => |accounts_found| {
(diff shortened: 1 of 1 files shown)


matklad:
Might want to do that after https://github.com/tigerbeetle/tigerbeetle/pull/2440 and related work? Then you'll be able to get timestamps from replies directly

owickstrom:
> Might want to do that after #2440 and related work? Then you'll be able to get timestamps from replies directly

It's OK to add it independently, because Vortex currently does the request-response cycle in sequence with a single driver. But when we do add multiple concurrent drivers, we need to sort replies for this to work, so it _will_ be useful.

src/testing/vortex/workload.zig
```diff
@@ -152,25 +157,42 @@ fn reconcile(result: Result, command: *const Command, model: *Model) !void {
...
                 // Check that linked transfers fail together.
-                if (entry.index > 0) {
-                    const preceeding_transfer = transfers[entry.index - 1];
+                if (transfer_index > 0) {
+                    const preceeding_transfer = transfers[transfer_index - 1];
                     if (preceeding_transfer.flags.linked) {
-                        try testing.expect(e > 0);
-                        const preceeding_entry = entries[e - 1];
-                        try testing.expect(preceeding_entry.index == entry.index - 1);
+                        try testing.expect(transfer_index > 0);
```
matklad: This is already inside `if (tranfesr_index > 0)`

src/testing/vortex/workload.zig
```diff
@@ -352,6 +448,15 @@ fn lookup_all_accounts(model: *const Model) Command {
     return .{ .lookup_all_accounts = buffer[0..events_count] };
 }
 
+fn lookup_latest_transfers(model: *const Model) Command {
+    const ids = model.latest_transfers.slice();
+    // While the model's array should live long enough for the command to execute, we keep it
+    // uniform here and use the command buffer as usual.
+    const buffer = command_buffers.lookup_latest_transfers[0..ids.len];
+    @memcpy(buffer, ids);
```
matklad: Use `stdx.copyDisjoint` as that is a bit more precise about intention

src/testing/vortex/workload.zig
```diff
@@ -184,31 +206,104 @@ fn reconcile(result: Result, command: *const Command, model: *Model) !void {
...
+                }
+                timestamp_max = transfer.timestamp;
+            }
+        },
     }
 }
 
+/// A bounded array of transfer IDs, which appends at the end, dropping items at the start when
+/// necessary. Like a constrained /deque/.
+const LatestsTransfers = struct {
```
matklad: Can we use RingBuffer from `ring_buffer.zig` here? That won't give you a `.slice`, but I an iterator would work?

owickstrom: Didn't know we had one! Was surfing around Zig std for a ring buffer but couldn't find a polymorphic one.

It's refactored now. Felt a bit clumsy to loop in order to drop N values from the tail. What do you think?

src/testing/vortex/workload.zig
```diff
@@ -184,31 +206,69 @@ fn reconcile(result: Result, command: *const Command, model: *Model) !void {
                 try testing.expect(model.account_exists(transfer.debit_account_id));
                 try testing.expect(model.account_exists(transfer.credit_account_id));
             }
+
+            // Drop the oldest transfer IDs and add new ones.
+            for (0..@min(successful_transfer_ids.count(), model.latest_transfers.count)) |_| {
+                model.latest_transfers.retreat_tail();
```
matklad, line 212: I think this is fine for now! If we need this in a couple of more places, we can add a helper to do this!

matklad (approved):

PR #2435, opened 2024-10-30
java: add a test for concurrent close

matklad:


matklad:
This currently crashes the JVM with

```
thread 1063421 panic: reached unreachable code
/Users/matklad/p/tb/work/zig/lib/std/debug.zig:412:14: 0x10ed3cf1f in assert (tb_jniclient)
             ^
/Users/matklad/p/tb/work/src/clients/c/tb_client/context.zig:529:23: 0x10edc5dc3 in cancel (tb_jniclient)
                assert(batched.batch_next == null or batched.batch_allowed);
                      ^
/Users/matklad/p/tb/work/src/clients/c/tb_client/context.zig:282:62: 0x10eddc387 in run (tb_jniclient)
            while (self.submitted.pop()) |packet| self.cancel(packet);
                                                             ^
/Users/matklad/p/tb/work/zig/lib/std/Thread.zig:408:13: 0x10edc7913 in callFn__anon_11237 (tb_jniclient)
            @call(.auto, f, args);
            ^
/Users/matklad/p/tb/work/zig/lib/std/Thread.zig:674:30: 0x10ed9867f in entryFn (tb_jniclient)
                return callFn(f, args_ptr.*);
                             ^
???:?:?: 0x19b48b2e3 in ??? (libsystem_pthread.dylib)
???:?:?: 0x19b4860fb in ??? (libsystem_pthread.dylib)
```

matklad:
cc @kprotty, @batiati could you take a look at what's going on here?

matklad:
Debugged:

* https://github.com/tigerbeetle/tigerbeetle/blob/0b764aa8f543631655ecbc82659b4a0a105c430a/src/clients/java/src/client.zig#L127-L133 -- here we don't initialized a bunch of fields, like `batch_next` and such

* https://github.com/tigerbeetle/tigerbeetle/blob/e6ede94ceb48a7ea781fb72ab6d769c2ce8ce6c7/src/clients/c/tb_client/context.zig#L421-L425 -- here we send this partially-initialized packet to the cancelation routine
* https://github.com/tigerbeetle/tigerbeetle/blob/e6ede94ceb48a7ea781fb72ab6d769c2ce8ce6c7/src/clients/c/tb_client/context.zig#L530 -- here we dereferencing wild pointers!

kprotty:
packets come in through `self.submitted` and either go to `request` or `cancel` directly through the while loop on shutdown. The `request` path is normally the one to initialize the `packet.batch*` fields before potentially calling `cancel` but the while-loop path doesnt.

I think the while-loop path should go through `request` which set the right fields, sees shutdown before reaching `vsr.Client`, and eventually `cancel`s it.

src/clients/c/tb_client/context.zig
```diff
@@ -560,6 +560,22 @@ pub fn ContextType(
...
+                .next = null,
+                .user_data = packet.user_data,
+                .operation = packet.operation,
+                .status = .ok,
+                .data_size = packet.data_size,
+                .data = packet.data,
+                .batch_next = null,
+                .batch_tail = null,
+                .batch_size = 0,
+                .batch_allowed = false,
```
batiati, line 578: We could simplify this API by moving these fields to a separate struct (opaque to the client libraries).

Example:
```zig
// Exposed to the client.
pub const PacketData = extern struct {
    user_data: ?*anyopaque,
    operation: u8,
    status: Packet.Status,
    data_size: u32,
    data: ?*anyopaque,
    reserved: [42]u8 = [_]u8{0} ** 42, // Internal fields are just opaque.

    comptime {
        assert(@sizeOf(PacketData) == @sizeOf(Packet));
        assert(@alignOf(PacketData) == @alignOf(Packet));
    }
};

// Internal implementation.
pub const Packet = extern struct {
    user_data: ?*anyopaque,
    operation: u8,
    status: Status,
    data_size: u32,
    data: ?*anyopaque,
    next: ?*Packet,
    batch_next: ?*Packet,
    batch_tail: ?*Packet,
    batch_size: u32,
    batch_allowed: bool,
    reserved: [7]u8 = [_]u8{0} ** 7,

    comptime {
        assert(@sizeOf(Packet) == 64);
        assert(@alignOf(Packet) == 8);
    }
}
```

matklad:
Ok, so I tried follow batiati's suggestion to refactor the API, but then figured out that that's going to be a large change, and that I don't want to mix bug-fix and a refactor in the same PR. I still want to do a refactor, but I want to more carefully survey our entire C ABI layer before that :P 

So I decided to go with kprotty's suggestion instead, and to make sure that the cancellation path also goes through `.request`. This indeed feels like the right thing to do as, it reduces dimensionality (a packet always goes through on_submit -> request -> cancel OR complete). But that, sadly, didn't work --- request has "registration was completed" as a pre-condition, but, I think, we could get canceled before registration finishes. I added an assert to that end. 

So I think I still want to merge what we have here now, to fix the bug in the most straightforward way, and then revisit the wider context later!

src/clients/c/tb_client/context.zig
```diff
@@ -560,6 +562,22 @@ pub fn ContextType(
...
 
         fn on_submit(implementation: *ContextImplementation, packet: *Packet) void {
+            // Packet is caller-allocated to enable elastic intrusive-link-list-based memory
+            // management. However, some of Packet's fields are essentially private. Initialize
+            // them here to avoid threading default fields through FFI boundary.
+            packet.* = .{
+                .next = null,
+                .user_data = packet.user_data,
+                .operation = packet.operation,
+                .status = .ok,
```
batiati, line 572: Note when refactoring `tb_packet_t`:

`status` is the only field that is both exposed to the client and updated by `tb_client`. We should change this API so that the status is instead returned alongside `results`.

batiati (approved):

PR #2436, opened 2024-10-30
bus: reject connections from unknown replicas

matklad:
This could happen if, eg, the operator passes too few addresses to a
client.

### DIFF
--- a/src/message_bus.zig
+++ b/src/message_bus.zig
@@ -879,16 +879,16 @@ fn MessageBusType(comptime process_type: vsr.ProcessType) type {
 
                 connection.peer = header_peer;
                 switch (connection.peer) {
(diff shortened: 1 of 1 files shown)


src/message_bus.zig
```diff
@@ -879,16 +879,20 @@ fn MessageBusType(comptime process_type: vsr.ProcessType) type {
 
                 connection.peer = header_peer;
                 switch (connection.peer) {
-                    .replica => {
+                    .replica => |replica_index| {
+                        if (replica_index >= bus.configuration.len) {
+                            return false;
```
matklad, line 884: We have `log.warn` at the call-site of this function.

sentientwaffle (approved):

PR #2437, opened 2024-10-31
grid: divide request buffer between the read_global_queue & faulty_blocks queues

chaitanyabhandari:
This PR resolves a VOPR false positive on `main` (`./zig/zig build -Drelease vopr -- 435571418441293928`). 

### Problem

After the VOPR liveness mode, if the cluster is found to not have liveness, `core_missing_blocks` checks if all corrupt blocks that can be repaired (are available on the core) should be repaired. However, our grid repair protocol does not guarantee this. Specifically, while choosing what blocks to send in `.request_blocks`, we *first* check `grid.read_global_queue`, and if there is any more capacity in the request buffer, we check `grid.blocks_missing.faulty_blocks`. This could lead to the blocks in `grid.blocks_missing.faulty_blocks` getting starved.

In this VOPR seed, there were 3 faulty blocks in `grid.read_global_queue` (that were not available on the core), leaving space for 1 block from `grid.blocks_missing.faulty_blocks` (`grid_repair_request_max` is 4). That 1 block was not available in the core, so we *never* try to repair any other blocks from `grid.blocks_missing.faulty_blocks`. 

This is because our logic for *cycling* through `grid.blocks_missing.faulty_blocks` is buggy (it requires a + 1 so we start from the *next* faulty block on subsequent invocations).
https://github.com/tigerbeetle/tigerbeetle/blob/5a5b605ebbdfa5e912c25b39e510485d1c8c493a/src/vsr/grid_blocks_missing.zig#L176-L177

### Solution

Now, our grid repair protocol guarantees that all corrupt blocks that can be repaired (are available on the core) should be repaired. To achieve this, we now divide up the buffer for `.request_blocks` between `grid.blocks_missing.faulty_blocks` & `grid.read_global_queue`, such that we always request blocks from both.

Additionally, we fix our buggy logic for cycling through `grid.blocks_missing.faulty_blocks` .

### DIFF
--- a/src/vsr/grid.zig
+++ b/src/vsr/grid.zig
@@ -1177,10 +1177,16 @@ pub fn GridType(comptime Storage: type) type {
         pub fn next_batch_of_block_requests(grid: *Grid, requests: []vsr.BlockRequest) usize {
             assert(grid.callback != .cancel);
             assert(requests.len > 0);
+            assert(requests.len == constants.grid_repair_reads_max);
 
             // Prioritize requests for blocks with stalled Grid reads, so that commit/compaction can
-            // continue.
-            const request_faults_count = @min(grid.read_global_queue.count, requests.len);
+            // continue. We now divide the buffer up between `read_global_queue` and
+            // `blocks_missing.faulty_blocks` so that we always request blocks from both queues.
+            const request_faults_count = @min(
+                grid.read_global_queue.count,
+                // Surplus from `blocks_missing.faulty_blocks` may be used by `read_global_queue`.
+                requests.len / 2 + (requests.len / 2 -| grid.blocks_missing.faulty_blocks.count()),
+            );
             // (Note that many – but not all – of these blocks are also in the GridBlocksMissing.
             // The `read_global_queue` is a FIFO, whereas the GridBlocksMissing has a fixed
             // capacity.)
--- a/src/vsr/grid.zig
+++ b/src/vsr/grid.zig
@@ -1180,12 +1180,18 @@ pub fn GridType(comptime Storage: type) type {
             assert(requests.len == constants.grid_repair_reads_max);
 
             // Prioritize requests for blocks with stalled Grid reads, so that commit/compaction can
-            // continue. We now divide the buffer up between `read_global_queue` and
+            // continue. We divide the buffer up between `read_global_queue` and
             // `blocks_missing.faulty_blocks` so that we always request blocks from both queues.
+            // Surplus from `blocks_missing.faulty_blocks` may be used by `read_global_queue`.
+            const request_faults_count_max = requests.len -
+                @min(@divFloor(requests.len, 2), grid.blocks_missing.faulty_blocks.count());
+            assert(request_faults_count_max > 0);
+            assert(request_faults_count_max <= requests.len);
+            assert(request_faults_count_max >= @divFloor(requests.len, 2));
+
             const request_faults_count = @min(
                 grid.read_global_queue.count,
-                // Surplus from `blocks_missing.faulty_blocks` may be used by `read_global_queue`.
-                requests.len / 2 + (requests.len / 2 -| grid.blocks_missing.faulty_blocks.count()),
+                request_faults_count_max,
             );
             // (Note that many – but not all – of these blocks are also in the GridBlocksMissing.
             // The `read_global_queue` is a FIFO, whereas the GridBlocksMissing has a fixed
--- a/src/vsr/grid_blocks_missing.zig
+++ b/src/vsr/grid_blocks_missing.zig
@@ -173,8 +173,10 @@ pub const GridBlocksMissing = struct {
             }
         }
 
+        // +1 so we start from the next faulty block in the subsequent function invocation, thereby
+        // cycling through the faulty blocks in the queue.
         queue.faulty_blocks_repair_index =
-            (queue.faulty_blocks_repair_index + fault_offset) % faults_total;
+            (queue.faulty_blocks_repair_index + fault_offset + 1) % faults_total;
 
         assert(requests_count <= requests.len);
         assert(requests_count <= faults_total);
--- a/src/vopr.zig
+++ b/src/vopr.zig
@@ -683,7 +683,7 @@ pub const Simulator = struct {
         const cluster_commit_max: u64 = simulator.cluster.state_checker.commits.items.len - 1;
         var cluster_op_checkpoint: u64 = 0;
         for (simulator.cluster.replicas) |*replica| {
-            if (simulator.core.isSet(replica.replica)) {
+            if (simulator.core.isSet(replica.replica) and !replica.standby()) {
                 cluster_op_checkpoint = @max(
                     cluster_op_checkpoint,
                     replica.superblock.working.vsr_state.checkpoint.header.op,


matklad:
Maybe I am missing some context here, but why don't we want to just shuffle all the blocks to make sure everything that is possible to repair, is repaired?

chaitanyabhandari:
@matklad I briefly discussed this with dj yesterday (https://tigerbeetleinc.slack.com/archives/C042KLT9R8F/p1730320994636999), I believe we decided not to do the shuffling and account for our current behavior in the VOPR, as it actually doesn't solve the entire problem of starvation (it merely makes it _somewhat_ better). 

Specifically, there was also the case where the blocks in the `read_global_queue` are absent on the core and can't be repaired, but are enough to fully fill up the request buffer for .request_blocks. Additionally, there are some blocks in `faulty_blocks` that *can* be repaired. With our current logic, we will never get to the blocks from faulty_blocks...  My proposed solution to alleviate this was: we could perhaps also divide up the buffer between faulty_blocks & read_global_queue, such that we always request blocks from both?

Thoughts?

sentientwaffle:
> My proposed solution to alleviate this was: we could perhaps also divide up the buffer between faulty_blocks & read_global_queue, such that we always request blocks from both?

Let's go for it! Best to maximize repair, even if we are otherwise stuck on an unrepairable block.

chaitanyabhandari:
> > My proposed solution to alleviate this was: we could perhaps also divide up the buffer between faulty_blocks & read_global_queue, such that we always request blocks from both?
> 
> Let's go for it! Best to maximize repair, even if we are otherwise stuck on an unrepairable block.

Implemented as part of the new commits!

chaitanyabhandari:
https://github.com/tigerbeetle/tigerbeetle/pull/2437/commits/52ba13067d02528596452b05d72ab0f96364628c is a fix for a VOPR false positive (`./zig/zig build -Drelease vopr -- 10196760699601082042`) on this branch (dae675ea161fd8a396e41b97ee72293870d92a74). In this seed, R0 & R1 have both corrupted op=135 in their WAL, so neither of them can step up as primary. R3 (a standby) is more up to date (op_checkpoint=139), *does* have op=135 intact in its WAL, but can't help R0 & R1 since replicas don't repair prepares from standby nodes.

```C
[info] (cluster): no liveness, final cluster state (core=1011):
[info] (cluster):  0   \ v     10673V 119/134/143C 112:143Jo  1/_1J! 112:143Wo <__0:__0> v1:3   274Ga  0G!   0G?
[info] (cluster):  1   \  v    10673V 119/134/143C  16:143Jo  0/_1J!   7:143Wo <_24:123> v1:3   274Ga  0G!   0G?
[info] (cluster):  2   #   #   
[info] (cluster):  3   |    .     5V 139/143/143C 112:143Jo  0/_0J! 112:143Wo <__0:__0> v3:3   154Ga  0G!   0G?
```

The issue here is that `core_missing_prepare` does not filter out standby nodes while calculating the maximum checkpoint in the cluster. This leads to R0 & R1 being erroneously being considered as *lagging* replicas, even though R3 is a more-up-to-date *standby* that can't participate in consensus.

However, this makes me wonder - *Why can't replicas repair prepares/headers from standby nodes?* I am planning on exploring this in a separate PR.

src/vsr/grid_blocks_missing.zig
```diff
@@ -173,8 +173,10 @@ pub const GridBlocksMissing = struct {
             }
         }
 
+        // +1 so we start from the next faulty block in the subsequent function invocation, thereby
+        // cycling through the faulty blocks in the queue.
         queue.faulty_blocks_repair_index =
-            (queue.faulty_blocks_repair_index + fault_offset) % faults_total;
+            (queue.faulty_blocks_repair_index + fault_offset + 1) % faults_total;
```
sentientwaffle, line 179: Great catch!

src/vsr/grid.zig
```diff
@@ -1177,10 +1177,16 @@ pub fn GridType(comptime Storage: type) type {
...
 
             // Prioritize requests for blocks with stalled Grid reads, so that commit/compaction can
-            // continue.
-            const request_faults_count = @min(grid.read_global_queue.count, requests.len);
+            // continue. We now divide the buffer up between `read_global_queue` and
+            // `blocks_missing.faulty_blocks` so that we always request blocks from both queues.
+            const request_faults_count = @min(
+                grid.read_global_queue.count,
+                // Surplus from `blocks_missing.faulty_blocks` may be used by `read_global_queue`.
+                requests.len / 2 + (requests.len / 2 -| grid.blocks_missing.faulty_blocks.count()),
```
sentientwaffle: Extract to:

```zig
            const request_faults_count_max =
                @divFloor(requests.len, 2) +
                (@divFloor(requests.len, 2) -| grid.blocks_missing.faulty_blocks.count());
            assert(request_faults_count_max > 0);
            assert(request_faults_count_max < requests.len);
```

and prefer `@divFloor()` to `/` for division, since it is more explicit.

chaitanyabhandari:
Addressed comments in https://github.com/tigerbeetle/tigerbeetle/pull/2437/commits/bb3e3fe97f7f358377af17c8576af67067c2e133!

src/vsr/grid.zig
```diff
@@ -1177,10 +1177,22 @@ pub fn GridType(comptime Storage: type) type {
...
             // Prioritize requests for blocks with stalled Grid reads, so that commit/compaction can
-            // continue.
-            const request_faults_count = @min(grid.read_global_queue.count, requests.len);
+            // continue. We divide the buffer up between `read_global_queue` and
+            // `blocks_missing.faulty_blocks` so that we always request blocks from both queues.
+            // Surplus from `blocks_missing.faulty_blocks` may be used by `read_global_queue`.
+            const request_faults_count_max =
+                @divFloor(requests.len, 2) +
+                (@divFloor(requests.len, 2) -| grid.blocks_missing.faulty_blocks.count());
+            assert(request_faults_count_max > 0);
```
sentientwaffle, line 1188:
```suggestion
            assert(request_faults_count_max > 0);
            assert(request_faults_count_max >= @divFloor(requests.len, 2));
```

src/vsr/grid.zig
```diff
@@ -1177,10 +1177,22 @@ pub fn GridType(comptime Storage: type) type {
...
 
             // Prioritize requests for blocks with stalled Grid reads, so that commit/compaction can
-            // continue.
-            const request_faults_count = @min(grid.read_global_queue.count, requests.len);
+            // continue. We divide the buffer up between `read_global_queue` and
+            // `blocks_missing.faulty_blocks` so that we always request blocks from both queues.
+            // Surplus from `blocks_missing.faulty_blocks` may be used by `read_global_queue`.
+            const request_faults_count_max =
+                @divFloor(requests.len, 2) +
+                (@divFloor(requests.len, 2) -| grid.blocks_missing.faulty_blocks.count());
```
sentientwaffle: Ah, if `requests.len` is an odd number, then `request_faults_count_max` will at most be `requests.len - 1`!
We could switch one of the `@divFloor`'s to a `div_ceil`, but the intent might be clearer with e.g.:

```
requests.len - @min(
  @divFloor(requests.len, 2),
  grid.blocks_missing.faulty_blocks.count(),
);
```

chaitanyabhandari: Ah, that is a great point regarding odd-numbered requests.len, especially because there's no asserts stopping us from setting `grid_repair_reads_max` to an odd number!

chaitanyabhandari:
Addressed comments in https://github.com/tigerbeetle/tigerbeetle/commit/85878c390afb910256465c2637005d0984219261! 

Also, the CFO found a VSR crash in a 1-node cluster (`./zig/zig build -Drelease vopr -- 13768260193285844595`) on this branch, it also fails on main. I will inspect it tomorrow!
```C
/Users/chaitanyabhandari/Desktop/tigerbeetle/tigerbeetle/src/vsr/replica.zig:1884:21: 0x104ab9547 in on_commit (vopr)
                    @panic("commit checksum verification failed");
```

sentientwaffle (approved):

PR #2439, opened 2024-10-31
test: Java workload sleep

owickstrom:
This is to increase chance of cluster healing, and to not overwhelm the Antithesis environment with disk space usage.


matklad:
We still need to do this, but no one is actively working on it at the moment.

PR #2441, opened 2024-10-31
devhub: Replace curl shell invocation with Zig's http client

fabioarnold:


src/shell.zig
```diff
@@ -857,3 +857,56 @@ fn discover_project_root() !std.fs.Dir {
...
+
+/// Issues an HTTP POST request with the given `body` to the given `url`.
+///
+/// If the response is not 200 OK, the response body is logged and an error is returned.
+pub fn http_post(shell: *Shell, url: []const u8, body: []const u8, options: HttpOptions) !void {
+    errdefer |err| log.err("failed to HTTP post to \"{s}\": {s}", .{ url, @errorName(err) });
+
+    const allocator = shell.arena.allocator();
+    var client = std.http.Client{ .allocator = allocator };
+    defer client.deinit();
```
matklad: Let's use `shell.gpa` here --- arena is for caller, such that they don't have to think about one-off memory management. But _inside_ shell we are writing re-usable functions, so it's relatively easier to get memory management right.

So, we could be using `gpa` here and be careful around defers. 


If, at some point, we would love to _return_ some data, then the returned data (and only it) would be allocated in the arena.

src/shell.zig
```diff
@@ -857,3 +857,56 @@ fn discover_project_root() !std.fs.Dir {
...
+
+    const uri = try std.Uri.parse(url);
+
+    var header_buffer: [4 << 10]u8 = undefined;
+    var request = try client.open(.POST, uri, .{ .server_header_buffer = &header_buffer });
+    defer request.deinit();
+    if (options.content_type) |content_type| {
```
matklad: We add a blank line after defer, so that it stands out more

matklad (approved):

PR #2442, opened 2024-10-31
IO: stdx.unexpectedErrno 

sentientwaffle:
If we hit `unexpectedErrno` that indicates a bug in our IO error handling -- we should dump useful information to aid troubleshooting.

`std.posix.unexpectedErrno()` only logs the error in `mode=Debug`.

Add a `stdx.unexpectedErrno()` with useful logging in `ReleaseSafe` mode too.

### DIFF
--- a/src/stdx.zig
+++ b/src/stdx.zig
@@ -870,7 +870,7 @@ pub const DateTimeUTC = struct {
 
 /// Like std.posix's `unexpectedErrno()` but log unconditionally, not just when mode=Debug.
 /// The added `label` argument works around the absence of stack traces in ReleaseSafe builds.
-pub fn unexpectedErrno(label: []const u8, err: std.posix.system.E) std.posix.UnexpectedError {
+pub fn unexpected_errno(label: []const u8, err: std.posix.system.E) std.posix.UnexpectedError {
     log.scoped(.stdx).err("unexpected errno: {s}: code={d} name={?s}", .{
         label,
         @intFromEnum(err),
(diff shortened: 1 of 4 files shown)


src/stdx.zig
```diff
@@ -867,3 +867,18 @@ pub const DateTimeUTC = struct {
         });
     }
 };
+
+/// Like std.posix's `unexpectedErrno()` but log unconditionally, not just when mode=Debug.
+/// The added `label` argument works around the absence of stack traces in ReleaseSafe builds.
+pub fn unexpectedErrno(label: []const u8, err: std.posix.system.E) std.posix.UnexpectedError {
+    log.scoped(.stdx).err("unexpected errno: {s}: code={d} name={?s}\n", .{
```
matklad: why `\n`?

src/stdx.zig
```diff
@@ -867,3 +867,18 @@ pub const DateTimeUTC = struct {
...
+        label,
+        @intFromEnum(err),
+        std.enums.tagName(std.posix.system.E, err),
+    });
+
+    if (builtin.mode == .Debug) {
+        std.debug.dumpCurrentStackTrace(null);
```
matklad, line 881: 🤯 ❤️

src/stdx.zig
sentientwaffle: Good catch; this was modified from std's implementation, which used `std.debug.print()`!

matklad (dismissed):

sentientwaffle:
Prior merge failed due to a typo in the last commit. Not sure why it failed in the _merge queue_ rather than the _PR's CI_, though -- maybe a merge queue race. :fearful:

matklad (approved):

PR #2443, opened 2024-10-31
lsm: compaction comments and assertions

matklad:
Small things I've noticed on the stream!

### DIFF
--- a/src/lsm/compaction.zig
+++ b/src/lsm/compaction.zig
@@ -133,7 +133,7 @@ pub fn ResourcePoolType(comptime Grid: type) type {
                 build_index_block,
                 build_value_block,
 
-                // Block is in the read queue
+                // Block is in the read queue.
                 read_index_block,
                 read_index_block_done,
                 read_value_block,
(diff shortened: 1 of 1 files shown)


src/lsm/compaction.zig
```diff
@@ -1561,13 +1562,22 @@ pub fn CompactionType(
...
+
             if (compaction.table_info_a.? == .immutable) {
                 compaction.counters.in += merge_result.consumed_a;
             }
             compaction.counters.dropped += merge_result.dropped;
 
             compaction.quotas.done += merge_result.consumed_a + merge_result.consumed_b;
-            assert(compaction.table_builder.value_count <= Table.data.value_count_max);
+            assert(compaction.quotas.done <= compaction.quotas.bar);
+            assert(compaction.quotas.done <= compaction.quotas.beat + Table.data.value_count_max);
```
matklad: This one is quite load beading actually! It proves our grid reservation calculation in beat_commence to be correct!

matklad: Ah, no, this is incorrect! We could arbitrary overshoot our beat quota in reality! This happens, for example, if level an and level b perfectly annihilate each other through tombstones, producing no output. In this case we'd have to run the entire companion in a single beat. 

This doesn't violate our gird reservation assumption --- although we consume _more_ values then expected, we don't produce more!

sentientwaffle (approved):

PR #1489, opened 2024-01-29
vsr: remove false assertion when reading client replies

matklad:
SEED: 15814337035080855944
Closes: #1486

PR #2444, opened 2024-11-01
vsr: handle inflight write for old reply post state sync

chaitanyabhandari:
This PR resolves a failed VOPR seed (`./zig/zig build -Drelease vopr -- 435571418441293928`) on main (0b764aa8f543631655ecbc82659b4a0a105c430a), wherein a syncing replica crashes while handling a `RequestReply` message:
```C
thread 56286805 panic: attempt to use null value
/Users/chaitanyabhandari/Desktop/tigerbeetle/tigerbeetle/src/vsr/client_replies.zig:169:29: 0x104743c37 in read_reply_sync (vopr)
            if (write_latest.?.message.header.checksum != session.header.checksum) {
                            ^
/Users/chaitanyabhandari/Desktop/tigerbeetle/tigerbeetle/src/vsr/replica.zig:2585:52: 0x10470c593 in on_request_reply (vopr)
            if (self.client_replies.read_reply_sync(slot, entry)) |reply| {
                                                   ^
/Users/chaitanyabhandari/Desktop/tigerbeetle/tigerbeetle/src/vsr/replica.zig:1390:60: 0x1046ed337 in on_message (vopr)
                .request_reply => |m| self.on_request_reply(m),
                                                           ^
```

`read_reply_sync` checks inflight writes to the slot with the ClientReply that we want to read, to synchronously serve a more up-to-date reply to the client. The logic assumes that if there is an inflight write to the slot that we want to read, it _must_ be a reply to the client whose reply is being requested via `.request_reply` (hence the unchecked usage of `write_latest.?`. However, this assumption may not hold true post state sync! 

https://github.com/tigerbeetle/tigerbeetle/blob/5a5b605ebbdfa5e912c25b39e510485d1c8c493a/src/vsr/client_replies.zig#L152-L162 

This is precisely what we observed in the failed seed. Let R be a lagging replica in the cluster:
* R initiates the write for an *old request* with op=O1, client=C1.
* R receives a more-up-to-date superblock, syncs to it, and updates its in-memory `client_sessions` from the superblock. However, state sync does not cancel in-flight writes for client replies, so the old write remains inflight.
* R it receives a `RequestReply` for op=O2, client=C2.
* R has the requested reply in memory, proceeds to process the request, and crashes due to the bug in `read_reply_sync`.

### DIFF
--- a/src/vsr/client_replies.zig
+++ b/src/vsr/client_replies.zig
@@ -159,7 +159,13 @@ pub fn ClientRepliesType(comptime Storage: type) type {
                 }
             }
 
-            if (write_latest.?.message.header.checksum != session.header.checksum) {
+            // The reply being written to the target slot may not be for the client that we're
+            // looking for. For example, it may be an old reply for a different client.
+            maybe(write_latest == null);
+
+            if (write_latest == null or
+                write_latest.?.message.header.checksum != session.header.checksum)
+            {
                 // We are writing a reply, but that's a wrong reply according to `client_sessions`.
                 // This happens after state sync, where we update `client_sessions` without
                 // waiting for the in-flight write requests to complete.
@@ -444,8 +450,8 @@ pub fn ClientRepliesType(comptime Storage: type) type {
                 message.header.request,
             });
 
-            // Release the write *before* invoking the callback, so that if the callback
-            // checks .writes.available() we doesn't appear busy when we're not.
+            // Release the write *before* invoking the callback, so that if the callback checks
+            // .writes.available() we don't erroneously appear busy.
             client_replies.writing.unset(write.slot.index);
             client_replies.writes.release(write);
 


matklad:
Was wondering why don't we just not cancel in-flight reply writes, git-blamed the relevant code a bit and arrived at https://github.com/tigerbeetle/tigerbeetle/pull/1489/commits/4b64528e3770542b973a7ffab8e6bf87eb2c049e. I have no memory of writing that! But the logic there seems sound, and, indeed, the present PR fixes a bug in that very commit!

matklad:
Could you reword the commit message itself to include our CFO trailer?

```
Seed: ./zig/zig build -Drelease vopr -- 435571418441293928
```

Otherwise, LGTM and a great catch!

chaitanyabhandari:
Altered the commit message in https://github.com/tigerbeetle/tigerbeetle/pull/2444/commits/72408055e95acb4d66327a706f046300f359c291!

matklad (approved):

PR #2447, opened 2024-11-04
vsr: checkpoint durability check while handling start_view must not be flaky

chaitanyabhandari:
This PR resolves a failed VOPR seed (./zig/zig build -Drelease vopr -- 13226853613775535404) on main (https://github.com/tigerbeetle/tigerbeetle/commit/3cd7cf4799f65c81ffbc36eda788b8b26fd7da72), wherein a cluster gets stuck due to a syncing replica that isn't able to transition out of `awaiting_checkpoint` status.

Currently, we use the following check to decide whether a replica should use WAL repair or state sync to make progress. If a replica gets into state sync  upon receiving a start_view message, and then transitions to `awaiting_checkpoint`, the only state transition allowed for it is to `updating_superblock`. This transition requires the below check to return *True* again. In other words, if the below check is flaky, the replica could get stuck in `awaiting_checkpoint` status and not participate in consensus, which in turn could lead to the cluster getting stuck.
https://github.com/tigerbeetle/tigerbeetle/blob/5a5b605ebbdfa5e912c25b39e510485d1c8c493a/src/vsr/replica.zig#L2238-L2249

This is precisely what we observed in the above failing seed. Let's assume a 3-replica cluster with R0, R1, & R2. While processing a start_view message from R1, `vsr.Checkpoint.durable` returns `True`, leading to R2 transitioning into `canceling_commit` →  `canceling_grid`  → `awaiting_checkpoint`. However, while processing a start_view message from R2, `vsr.Checkpoint.durable` returns `False`, as its commit_max is smaller. R0 gets stuck in `awaiting_checkpoint` status, stops responding to to `request_start_view` and stops send `prepare_ok`. This leads to the R0 and R2 getting stuck.

This change ensures that the checkpoint durability check is done using the most up-to-date commit_max that the replica knows of, to make sure the result of `vsr.Checkpoint.durable` isn't flaky (between two invocations while a replica is alive).

### DIFF
--- a/src/vsr/replica.zig
+++ b/src/vsr/replica.zig
@@ -2235,12 +2235,19 @@ pub fn ReplicaType(
             assert(view_headers[0].op == message.header.op);
             assert(view_headers[0].op >= view_headers[view_headers.len - 1].op);
 
-            if (vsr.Checkpoint.durable(self.op_checkpoint_next(), message.header.commit_max) and (
+            // Use the maximum commit_max the replica knows of to gauge checkpoint durability.
+            // This is important so that the result of `vsr.Checkpoint.durable` isn't flaky.
+            // Otherwise, a replica could transition to state sync using a start_view from one
+            // replica, but may never be able to transition out of it if subsequent start_view
+            // messages from another replica have a smaller commit_max.
+            const commit_max = @max(self.commit_max, message.header.commit_max);
+
+            if (vsr.Checkpoint.durable(self.op_checkpoint_next(), commit_max) and (
             //  Cluster is at least two checkpoints ahead. Although SV's checkpoint is not
             //  guaranteed to be durable on a quorum of replicas, it is safe to sync to it, because
             //  prepares in this replica's WAL are no longer needed.
                 vsr.Checkpoint.durable(self.op_checkpoint_next() +
-                constants.vsr_checkpoint_ops, message.header.commit_max) or
+                constants.vsr_checkpoint_ops, commit_max) or
                 // Cluster is on the next checkpoint, and that checkpoint is durable and is safe to
                 // sync to. Try to optimistically avoid state sync and prefer WAL repair, unless
                 // there's evidence that the repair can't be completed.
@@ -2251,7 +2258,7 @@ pub fn ReplicaType(
                 // State sync: at this point, we know we want to replace our checkpoint
                 // with the one from this SV.
 
-                assert(message.header.commit_max > self.op_checkpoint_next_trigger());
+                assert(commit_max > self.op_checkpoint_next_trigger());
                 assert(view_checkpoint.header.op > self.op_checkpoint());
 
                 if (self.syncing == .idle) {
@@ -2317,11 +2324,13 @@ pub fn ReplicaType(
                         {
                             self.set_op_and_commit_max(
                                 header.op,
-                                message.header.commit_max,
+                                commit_max,
                                 @src(),
                             );
                             assert(self.op == header.op);
                             assert(self.commit_max >= message.header.commit_max);
+                            assert(self.commit_max == commit_max);
+
                             break;
                         }
                     }


chaitanyabhandari:
An alternative solution that I could see for this issue was to lean into the fact that the below check could be flaky, and give a way for the replica to transition from `awaiting_checkpoint → idle`. That is, if a replica transitions from `canceling_commit → canceling_grid → awaiting_checkpoint` and _then_ finds that it can recover using WAL repair (the below check returns *False*), we transition from `awaiting_checkpoint` → `idle`. In fact, we could allow this transition for the `canceling_commit` & `awaiting_checkpoint` state, since they're both states where where a replica is simply *waiting* (unlike `canceling_grid` and `updating_superblock`), so we should safely be able to transition to `idle` for these cases.
https://github.com/tigerbeetle/tigerbeetle/blob/5a5b605ebbdfa5e912c25b39e510485d1c8c493a/src/vsr/replica.zig#L2238-L2249 

Thoughts, @matklad?

matklad:
Some context here:

The whole `.awaiting_checkpoint` thing was done for simplicity. There's a corresponding TODO in the code to get rid of it:

https://github.com/tigerbeetle/tigerbeetle/blob/6ff7a8e2a65cf0ddf163c52aea8b195e206532d4/src/vsr/sync.zig#L24-L27

I actually set out to do that at one point, but I hit some snag along the way (which one I don't remember). 

So I think the most promising path here is to bite the bullet and remove `.awaiting_checkpoint` completely, by just saving an `SV` and re-triggering it once commit-canceling flow is done. 

I am perhaps the best person to look into that, as there's a bunch of other refactors around that I had in mind? So let me make a weak promise to look into this until the end of the week!

---

That being said: 


> This leads to the R0 and R2 getting stuck.

My mental model here is that R2 should become a primary, and should send a fresher SV which would get us unstuck. Why isn't this happening?

matklad:
EDIT: recall the snag now! I tried to use `loopback_queue` interface to re-send an SV to the replica itself, but the had a bunch of assertions that only replica's own messages could be re-sent.

chaitanyabhandari:
> My mental model here is that R2 should become a primary, and should send a fresher SV which would get us unstuck. Why isn't this happening?


R2 does become primary in this case, but it only knows up to commit_max=107 (checkpoint trigger), so even with the most-up-to-date SV it can send, it can't get R0 unstuck (that requires commit_max > 107). Also, R2 tries to get a prepare_ok for op=108, but R0 doesn't respond since it is in `awaiting_checkpoint` status. 

Another interesting thing here is that R0's commit_max=108, but it isn't able to contribute that information in its DVC as its head op=106. In the below snippet 0 → R2, 2→ R0:
```C
257278-[debug] (replica): 0: on_do_view_change: view=9 quorum received
257279-[debug] (replica): 0: on_do_view_change: dvc: replica=0 log_view=7 op=108 commit_min=107 checkpoint=99
257280-[debug] (replica): 0: on_do_view_change: dvc: header: replica=0 op=108 checksum=312459197167059754128816670263654750072 nack=false present=true type=valid
257281-[debug] (replica): 0: on_do_view_change: dvc: header: replica=0 op=107 checksum=247868208235864206315964918732757375681 nack=false present=true type=valid
257282-[debug] (replica): 0: on_do_view_change: dvc: replica=2 log_view=7 op=106 commit_min=102 checkpoint=79
257283-[debug] (replica): 0: on_do_view_change: dvc: header: replica=2 op=106 checksum=227632590180233009894702014628389866256 nack=false present=true type=valid
```

This makes me wonder if we should also send `commit_max` separately in the DVC instead of encoding it as the last DVC header, for cases like these where a replica's commit_max is > its head op? 🤔

chaitanyabhandari:
> I am perhaps the best person to look into that, as there's a bunch of other refactors around that I had in mind? So let me make a weak promise to look into this until the end of the week!

Ack, sounds great to me!

PR #2453, opened 2024-11-05
io: fix race condition in fs_supports_direct_io

matklad:
This fixes (at least one issue) flaky upgrade test. There, several
replicas are being run in the same directory, and they all try to create
this `fs_supports_direct_io` probe file, which mostly succeeds, but
sometimes can fail!

Add a random cookie to prevent this.

And, as this is the 1274278th time I need to print a number to a
stack-allocated buffer, also add an `stdx.array_print` utility to check
buffer's size automatically.

### DIFF
--- a/src/io/linux.zig
+++ b/src/io/linux.zig
@@ -1662,7 +1662,10 @@ pub const IO = struct {
     fn fs_supports_direct_io(dir_fd: fd_t) !bool {
         if (!@hasField(posix.O, "DIRECT")) return false;
 
-        const path = "fs_supports_direct_io";
+        var cookie: [16]u8 = .{'0'} ** 16;
+        _ = stdx.array_print(16, &cookie, "{0x}", .{std.crypto.random.int(u64)});
+
+        const path: [:0]const u8 = "fs_supports_direct_io-" ++ cookie ++ "";
         const dir = std.fs.Dir{ .fd = dir_fd };
         const flags: posix.O = .{ .CLOEXEC = true, .CREAT = true, .TRUNC = true };
         const fd = try posix.openatZ(dir_fd, path, flags, 0o666);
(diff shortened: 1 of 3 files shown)


src/stdx.zig
```diff
@@ -815,6 +815,36 @@ test fmt_int_size_bin_exact {
     });
 }
 
+
+/// Like std.fmt.bufPrint, but checks, at compile time, that the buffer is sufficiently large.
+pub fn array_print(
+    comptime n: usize,
+    buffer: *[n]u8,
```
matklad, line 821: Could pass an `anytype` here,  but figured that with `*[n]` the signature would be more understandable. 

Alternatively, instead of taking an array as an argument, this could _return_ a bounded array of the precise size, but I think it's actually beneficial to see a hard-coded constant at the call-site, to serve as a sanity check that we aren't accidentally allocating 4megs on the stack.

src/io/linux.zig
```diff
@@ -1662,7 +1662,10 @@ pub const IO = struct {
     fn fs_supports_direct_io(dir_fd: fd_t) !bool {
         if (!@hasField(posix.O, "DIRECT")) return false;
 
-        const path = "fs_supports_direct_io";
+        var cookie: [16]u8 = .{'0'} ** 16;
+        _ = stdx.array_print(16, &cookie, "{0x}", .{std.crypto.random.int(u64)});
+
+        const path: [:0]const u8 = "fs_supports_direct_io-" ++ cookie ++ "";
```
sentientwaffle, line 1668: TIL you can use `++` at runtime with arrays!

sentientwaffle (approved):

src/io/linux.zig
matklad: Yup! One of the coolest zig features!

PR #2455, opened 2024-11-06
style: emphasize human reasoning

matklad:
Recently, VOPR became too good, and it becomes _really_ tempting to switch to an _empirical_ mode of coding: write some code and let the VOPR figure out whether it is correct or not.

This is suboptimal, for two reasons.

_Firstly_, VOPR, as any testing methodology, is one-sided --- if it finds a bug, there's definitely a bug, but silence of the VOPR doesn't guarantee total absence of bugs! Something might be _really_ unlikely, or completely uncovered by the VOPR.

_Secondly_, safety comes in layers and cross checks. Just formal or informal reasoning is not enough. And just simulation is not enough. Simulation AND reasoning is still not enough, but is _much_ better, because a bug now has to slip trough two cracks in two different surfaces! Minas Tirith is a seventh-fold walled fortress for a reason!

So instead of using VOPR to run experiments, like a physicist would do, we should be more like a mathematician --- derive correctness of code from the first principles, and use VOPR to debug our understanding instead of debugging our code.

### DIFF
--- a/docs/TIGER_STYLE.md
+++ b/docs/TIGER_STYLE.md
@@ -141,6 +141,15 @@ Code](https://spinroot.com/gerard/pdf/P10.pdf) will change the way you code fore
     also why **tests must test exhaustively**, not only with valid data but also with invalid data,
     and as valid data becomes invalid.
 
+  - Assertions are a safety net, not a substitute for human understanding. With simulation testing,
+    there is the temptation to trust the fuzzer. But a fuzzer can prove only the presence of bugs,
+    not their absence. Therefore:
+    - Build a precise mental model of the code first,
+    - encode your understanding in the form of assertions,
+    - write the code and comments to explain and justify the mental model to your reviewer,
+    - and use VOPR as the final line of defense, to find bugs in your and reviewer's understanding
+      of code.
+
 - All memory must be statically allocated at startup. **No memory may be dynamically allocated (or
   freed and reallocated) after initialization.** This avoids unpredictable behavior that can
   significantly affect performance, and avoids use-after-free. As a second-order effect, it is our


sentientwaffle (dismissed):

docs/TIGER_STYLE.md
```diff
@@ -141,6 +141,11 @@ Code](https://spinroot.com/gerard/pdf/P10.pdf) will change the way you code fore
     also why **tests must test exhaustively**, not only with valid data but also with invalid data,
     and as valid data becomes invalid.
 
+  - Asserts are a safety net, not a substitute for understanding. With simulation testing, there is
+    a temptation to trust the fuzzer. But a fuzzer can only demonstrate presence of bugs, it can't
+    prove their absence. Assertions must be justified through informal reasoning and understanding.
+    The understanding then is the subject of fuzzing, not the code itself.
```
jorangreef:
```suggestion
  - Assertions are a safety net, not a substitute for human understanding. With simulation testing, there is
    the temptation to trust the fuzzer. But a fuzzer can prove only the presence of bugs, not
    their absence. Assertions must therefore be justified first through reasoning, which understanding then is the subject of fuzzing, not the code itself.
```

jorangreef: Can we bring more defense in depth into this as well? i.e. Human programmer encodes their understanding of the positive and negative space into code and assertions. Human reviewer attempts to verify this understanding. And then simulation attempts to falsifies this understanding.

jorangreef (dismissed):

docs/TIGER_STYLE.md
jorangreef: i.e. We don't want to skip right to only attempting to autonomously falsifying code. We also want two humans to apply their reasoning of the domain as a first and second line of defense.

sentientwaffle (approved):

g-berthiaume:
Small nitpick, if I may.

`VOPR` is not defined anywhere in `docs/TIGER_STYLE.md`. 
As I like to share this document with people who are unfamiliar with tiger beetles, it would be great if it could be self-contained.

Best !

PR #2460, opened 2024-11-07
Message: rename `.body()` to `.body_used()`, and use it for `state_machine.commit()`

cb22:
This matches the convention inside the LSM, and makes it more clear at a glance that you're getting the portion of the body bounded by the header's size.

This was prompted by seeing that `replica.zig` calls the state machine with a manual `prepare.buffer[@sizeOf(Header)..prepare.header.size]` and wondering why that didn't use the original `.body()`. Turns out, there doesn't seem to be a reason, so change that to use `.body_used()` too.

### DIFF
--- a/src/vsr/replica.zig
+++ b/src/vsr/replica.zig
@@ -1451,7 +1451,7 @@ pub fn ReplicaType(
                         .releases = .{},
                     };
 
-                    const releases_all = std.mem.bytesAsSlice(vsr.Release, message.body());
+                    const releases_all = std.mem.bytesAsSlice(vsr.Release, message.body_used());
                     const releases = releases_all[0..message.header.release_count];
                     assert(releases.len == message.header.release_count);
                     vsr.verify_release_list(releases, message.header.release);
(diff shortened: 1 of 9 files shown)


src/vsr/replica.zig
```diff
@@ -4463,7 +4463,7 @@ pub fn ReplicaType(
                     prepare.header.op,
                     prepare.header.timestamp,
                     prepare.header.operation.cast(StateMachine),
-                    prepare.buffer[@sizeOf(Header)..prepare.header.size],
```
cb22, line 4466: The only change that's not a rename.

src/vsr/replica.zig
```diff
@@ -4589,7 +4589,7 @@ pub fn ReplicaType(
 
                 const register_request = std.mem.bytesAsValue(
                     vsr.RegisterRequest,
-                    prepare.body()[0..@sizeOf(vsr.RegisterRequest)],
+                    prepare.body_used()[0..@sizeOf(vsr.RegisterRequest)],
```
cb22, line 4592: I didn't look at these; is there a reason we further slice by `sizeOf`? (on all the below too)

sentientwaffle: No, probably just by habit.

sentientwaffle (approved):

matklad:
Haven't reached the activation energy here, but I've long low-key wanted to change the LSM naming convention to be `values / values_buffer` rather than `values_used / values`. Anyway, the fact that it is consistent now should simplify the change!

Issue #696, opened 2023-04-27
VOPR: sometimes run with an unrestricted amount of faults


matklad:
Today, we carefully limit the number of faults a VOPR can inject for a particular prepare, to make sure that simulator's cluster can repair and remains live. We should add a mode where VOPR injects arbitrary many faults, and just skip liveness checks for this scenarios, still verifying correctness.

matklad:
cc @chaitanyabhandari I think one of vopr false positive (replica entering recovering head during upgrade) would be fixed as a result of this. THis is a big change  somewhat.

PR #2463, opened 2024-11-08
VOPR: Test misdirected writes

sentientwaffle:
(Not ready for review yet.)

Test misdirected writes in the VOPR.

A _misdirected write_ is a disk fault where:

1. the data is not written to the intended location. (i.e. a "lost write"), and
2. the data _is_ written to a different, mistaken location.

Currently:

- We allow for (at most) one misdirect fault per `Storage` (i.e. per replica) for the time being, for simplicity and because double-faults are not covered by our fault model. This will hopefully match physical disks – misdirected faults are an order of magnitude less frequent than bit rot, which in turn is an order of magnitude less frequent than LSEs.
- In order to keep things interesting:
  - misdirections are always within the same zone,
  - the entire write is misdirected (rather than only some of the sectors), and
  - the misdirected write lands on a convenient offset.
  
  Thanks to rigorous checksums, misdirections that break these rules just manifest as corruptions, and corruption is already well-tested. The goal here is to test how TigerBeetle handles well-formed but incorrectly-located data.

The misdirection is implemented with "overlays" over the `Storage.memory` buffer, so that `Storage.memory` is always pristine. That is, `Storage.memory` is what the disk would look like if no faults had ever occurred.

---

Also in this PR:
- Refactoring `Storage`.
- Enable read/write faults in the `client_replies` zone. Previously we were only injecting faults due to crashes ("torn writes").
- With these new faults, the VOPR found:
  - a couple of minor incorrect asserts in the [journal](https://github.com/tigerbeetle/tigerbeetle/pull/2463/commits/176d716f2bb4fa413ac05b04a0cc43a5087097e0).
  - a [correctness bug](https://github.com/tigerbeetle/tigerbeetle/pull/2463/commits/3f309bf1a9fcaf3eb518ab1b99dcf9b53a6bc562) in journal recovery!
  - a bug in the [client replies](https://github.com/tigerbeetle/tigerbeetle/pull/2463/commits/3e9906731f356336a793cd6b7b0f6df9a89d0919).
  - (Fixes in https://github.com/tigerbeetle/tigerbeetle/pull/2506).

### DIFF
--- a/src/iops.zig
+++ b/src/iops.zig
@@ -25,7 +25,7 @@ pub fn IOPSType(comptime T: type, comptime size: u6) type {
         }
 
         pub fn index(self: *IOPS, item: *T) usize {
-            const i = (@intFromPtr(item) - @intFromPtr(&self.items)) / @sizeOf(T);
+            const i = @divExact((@intFromPtr(item) - @intFromPtr(&self.items)), @sizeOf(T));
             assert(i < size);
             return i;
         }
--- a/src/lsm/tree_fuzz.zig
+++ b/src/lsm/tree_fuzz.zig
@@ -854,13 +854,14 @@ pub fn main(fuzz_args: fuzz.FuzzArgs) !void {
     const table_usage = random.enumValue(TableUsage);
     log.info("table_usage={}", .{table_usage});
 
-    const storage_fault_atlas = ClusterFaultAtlas.init(3, random, .{
+    var storage_fault_atlas = try ClusterFaultAtlas.init(allocator, 3, random, .{
         .faulty_superblock = false,
         .faulty_wal_headers = false,
         .faulty_wal_prepares = false,
         .faulty_client_replies = false,
         .faulty_grid = true,
     });
+    defer storage_fault_atlas.deinit(allocator);
 
     const storage_options = .{
         .seed = random.int(u64),
--- a/src/testing/cluster.zig
+++ b/src/testing/cluster.zig
@@ -182,11 +182,13 @@ pub fn ClusterType(comptime StateMachineType: anytype) type {
             const storage_fault_atlas = try allocator.create(StorageFaultAtlas);
             errdefer allocator.destroy(storage_fault_atlas);
 
-            storage_fault_atlas.* = StorageFaultAtlas.init(
+            storage_fault_atlas.* = try StorageFaultAtlas.init(
+                allocator,
                 options.replica_count,
                 random,
                 options.storage_fault_atlas,
             );
+            errdefer storage_fault_atlas.deinit(allocator);
 
             var grid_checker = try allocator.create(GridChecker);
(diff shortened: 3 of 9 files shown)


chaitanyabhandari:
I was _very_ curious to see what kinds of bugs would be unveiled by this PR, especially because I had thought about misdirected IO a fair bit while implementing https://github.com/tigerbeetle/tigerbeetle/pull/2345, but it was never tested against misdirected IO. Surprisingly, the first one I saw (`./zig/zig build -Drelease vopr -- 17637614406271072333`) is due to torn prepares! 

The failure is due to a *double storage fault*, i.e. a misdirected write to the redundant headers (for op=110), followed by a corruption in the slot for op=110 that is discovered on startup. This causes op=110 to get truncated as a torn prepare (as this double fault manifests as a corrupted prepare + a *valid* redundant header from the previous wrap), which triggers the following assert (because head op is inferred as op=109 on startup, even though op=110 was prepare_ok'd): https://github.com/tigerbeetle/tigerbeetle/blob/5a5b605ebbdfa5e912c25b39e510485d1c8c493a/src/testing/cluster/state_checker.zig#L116-L120

Here are the relevant logs stitched together: 

```C
[debug] (storage): 3: write_sectors_finish: misdirect zone=wal_headers offset=0->0 size=4096
[debug] (journal): 3: write_sectors: ring=vsr.journal.Ring.headers offset=0 len=4096 unlocked
[debug] (journal): 3: write: view=1 slot=14 op=110 len=320: 321777901078589245069031686077520390140 complete, marking clean
[debug] (replica): 3: send_prepare_ok: op=110 checksum=321777901078589245069031686077520390140

-------

[debug] (storage): 3: corrupting sector at zone=vsr.Zone.wal_prepares offset=57344 slot=14
[debug] (journal): 3: faulty slot=14, header_redundant_ok=false
[debug] (journal): 3: faulty slot=14, header_redundant_ok.operation=vsr.Operation(130)
[debug] (journal): 3: faulty slot=14, header_redundant_ok.operation=vsr.Operation(130), header_prepare_untrusted.valid_checksum()=false
[debug] (journal): 3: faulty slot=14, header_redundant_ok.?.op=78, header_prepare_untrusted.valid_checksum()=false
[warn] (journal): 3: recover_slot: recovered slot=0014 label=@TruncateTorn decision=cut_torn operation=vsr.Operation.reserved op=14
```
However, since our fault model does not include double storage faults, one option would be to be conservative in the VOPR while injecting storage faults... Makes me wonder though, given we eventually want to do https://github.com/tigerbeetle/tigerbeetle/issues/696 and we'd have to remove all of these places where we conservatively inject faults, perhaps adding a separate checker for this in the VOPR is the way to go? 🤔

sentientwaffle:
Superseded by https://github.com/tigerbeetle/tigerbeetle/pull/2677.

PR #2464, opened 2024-11-09
Interactive repl improvements + static allocation

brson:
This is a continuation of https://github.com/tigerbeetle/tigerbeetle/pull/2155, with additional commits that convert from dynamic allocation to static allocation.

It addresses the dynamic allocation introduced in that PR, but not the pre-existing dynamic allocation used by the repl parser, which is more complex and I intend to address later.

This is a prerequisite to landing the repl followup PR https://github.com/tigerbeetle/tigerbeetle/pull/2423

The approach I took here was to convert local buffers to member variables, mostly `ArrayList`s with pre-allocated capacity and `appendAssumeCapacity` calls. In the history ringbuffer though I used flat fixed-sized arrays because I found that easier to combine with the ringbuffer API.

I encourage reviewers to limit their reviews to my commits, as @wpaulino's are reviewed in https://github.com/tigerbeetle/tigerbeetle/pull/2155

cc @wpaulino @raghunandanbhat


src/terminal.zig
```diff
@@ -0,0 +1,430 @@
...
+const windows = std.os.windows;
+
+const builtin = @import("builtin");
+
+/// The terminal needs to read control codes, but the exact input capacity
+/// is unknown; this should be more than enough.
+const buffer_in_capacity = 256;
+
+pub const Terminal = struct {
+    mode_start: ?*const anyopaque,
```
batiati: We don't need to type-erase this:

```suggestion
    const ModeStart = switch (builtin.os.tag) {
        .linux, .macos => posix.termios,
        .windows => WindowsConsoleMode,
        else => unreachable,
    };

    mode_start: ?*const ModeStart,
```

I think `mode_start` doesn't need to be heap-allocated.

```
mode_start: ?ModeStart,
```

src/terminal.zig
```diff
@@ -0,0 +1,430 @@
...
+        interactive: bool,
+    ) !Terminal {
+        const stdout = std.io.getStdOut();
+        if (interactive and !stdout.getOrEnableAnsiEscapeSupport()) {
+            std.debug.print("ANSI escape sequences not supported.\n", .{});
+            std.process.exit(1);
+        }
+
+        const stdin = std.io.getStdIn();
+        var mode_start: ?*const anyopaque = null;
```
batiati:
```suggestion
        var mode_start: ?*const ModeStart = null;
```

src/terminal.zig
```diff
@@ -0,0 +1,430 @@
...
+        arguments: anytype,
+    ) !void {
+        if (self.stderr) |stderr| {
+            try stderr.print(format, arguments);
+        }
+    }
+
+    pub fn read_user_input(self: *Terminal) !?UserInput {
+        assert(self.mode_start != null);
+        const stdin = self.stdin.reader().any();
```
batiati: We can use the typed implementation instead:

```suggestion
        const stdin = self.stdin.reader();
```

src/terminal.zig
```diff
@@ -0,0 +1,430 @@
...
+        }
+    }
+
+    fn get_cursor_position(
+        self: *Terminal,
+    ) !struct { row: usize, column: usize } {
+        // Obtaining the cursor's position relies on sending a request payload to stdout. The
+        // response is read from stdin, but it may have been altered by user input, so we keep
+        // retrying until successful.
+        const stdin = self.stdin.reader().any();
```
batiati: We can use the typed implementation instead:

```suggestion
        const stdin = self.stdin.reader();
```

src/terminal.zig
```diff
@@ -0,0 +1,430 @@
...
+    // These are made optional so that printing on failure can be disabled in tests expecting them.
+    stdout: ?std.fs.File.Writer,
+    stderr: ?std.fs.File.Writer,
+
+    buffer_in: std.ArrayList(u8),
+
+    pub fn init(
+        allocator: std.mem.Allocator,
+        interactive: bool,
+    ) !Terminal {
```
batiati: `Terminal` is a large struct, we could do [in-place initialization](https://github.com/tigerbeetle/tigerbeetle/blob/main/docs/TIGER_STYLE.md#cache-invalidation) instead.
```suggestion
    pub fn init(
        self: *Terminal,
        allocator: std.mem.Allocator,
        interactive: bool,
    ) !void {
```

src/terminal.zig
batiati, line 1: We need a `deinit()` function to free `buffer_in` and `mode_start`.

src/repl.zig
```diff
@@ -538,22 +531,196 @@ pub fn ReplType(comptime MessageBus: type) type {
             }
         }
 
-        const single_repl_input_max = 10 * 4 * 1024;
+        const prompt = "> ";
+
+        fn read_until_newline_or_eof_alloc(
```
batiati: Since it now returns the memory owned by the buffer, we can drop the `_alloc` suffix.

src/repl.zig
```diff
@@ -538,22 +531,196 @@ pub fn ReplType(comptime MessageBus: type) type {
...
+            var history_index = repl.history.count;
+
+            while (true) {
+                const user_input = try repl.terminal.read_user_input() orelse return null;
+                switch (user_input) {
+                    .ctrlc => {
+                        // Erase everything below the current cursor's position in case Ctrl-C was
+                        // pressed somewhere inside the buffer.
+                        try repl.terminal.print("^C\x1b[J\n", .{});
+                        return &.{};
```
batiati, line 557: Funny question: how do we exit the repl?
Pressing Ctrl+C multiple times just cleans the prompt

batiati (commented):
Wow, using the REPL now is delightful!

src/repl.zig
brson: Ctrl+D

brson: Should probably respond to something like `exit` too, eventually.

src/terminal.zig
brson: Nothing in the repl does any cleaup or deallocation currently because it all allocates into an arena that is thrown away. I agree it's wrong, but would much prefer to address this issue later.

src/repl.zig
batiati: we should make it exit `vim` style :rofl:!

src/terminal.zig
brson: Ah, I'll just do it now.

batiati: > Nothing in the repl does any cleaup or deallocation currently because it all allocates into an arena that is thrown away. I agree it's wrong, but would much prefer to address this issue later.

That's fine then, just call the allocator `arena` and it would be clear enough.

src/repl.zig
batiati: Being pedantic, the correct action would be passing a pointer to a `*std.heap.ArenaAllocator` instead of an `Allocator`, but just the naming convention is good enough.

src/terminal.zig
batiati: Being pedantic, the correct action would be passing a pointer to a *std.heap.ArenaAllocator instead of an Allocator, but just the naming convention is good enough.

src/terminal.zig
brson: Done

src/terminal.zig
brson: ok, for now I've just made the arena type explicit in the Terminal.init type signature. I do think it makes sense to get rid of the arenas here and handle all the deinit correctly, but it'll take a careful readthrough to put in all the correct error handling etc.

brson:
@batiati Thanks for the careful review. I've addressed all comments.

brson:
I've retested this in various windows terminals and all seem to be working still.

I have not tested these patches at all in macos - just hoping they work the same as under linux.

brson:
I've tested on macos too.

batiati (commented):
LGTM!
Before we merge it, please print a notice instructing how to quit the REPL.

_Example:_
```
TigerBeetle CLI Client 0.0.1+a377697
  Crtl+D to exit.
```

**Suggestion:** We could handle the keys `home` and `end` as well, but I'll let up to you whether to include in this PR or in a follow-up.

brson:
Updated with the Ctrl+D message.

batiati (approved):

mrjbq7:
I just tried the repl and the parser doesn't support commands that span multiple lines.  I copied the first example verbatim and pasted and it doesn't work without being edited to be on one line:

```
$ ./tigerbeetle repl --cluster=0 --addresses=3000

TigerBeetle CLI Client 0.0.1+2d5501f
  Ctrl+D to exit.

Examples:
  create_accounts id=1 code=10 ledger=700 flags=linked|history,
                  id=2 code=10 ledger=700;
  lookup_accounts id=1;
  lookup_accounts id=1, id=2;
  get_account_transfers account_id=1 flags=debits|credits;
  get_account_balances account_id=1 flags=debits|credits;

>   create_accounts id=1 code=10 ledger=700 flags=linked|history,
Fail near line 1, column 63:

  create_accounts id=1 code=10 ledger=700 flags=linked|history,
                                                               ^ Near here.

Expected key starting key-value pair. e.g. `id=1`
> 
```

brson:
> I just tried the repl and the parser doesn't support commands that span multiple lines. I copied the first example verbatim and pasted and it doesn't work without being edited to be on one line:

Thanks @mrjbq7. Missed this at the time. Also reported in https://github.com/tigerbeetle/tigerbeetle/issues/2529.

PR #2465, opened 2024-11-10
state_machine: rename `execute()` to `execute_create()`

cb22:
At first glance, `.execute()` _sounds_ special, but it's really not: it's just `execute_create_accounts()` and `execute_create_transfers()` in a trenchcoat.

(The real entrypoint is `pub fn commit()`.)


batiati (approved):
🧥👍

PR #2471, opened 2024-11-11
Fix CI to  accept changelog entries with the "unreleased" tag.

batiati:


src/scripts/release.zig
```diff
@@ -87,7 +87,16 @@ pub fn main(shell: *Shell, gpa: std.mem.Allocator, cli_args: CLIArgs) !void {
...
+            if (changelog_current.release == null) {
+                @panic("The last changelog entry must have a release version.");
+            }
+            const changelog_previous =
+                changelog_previous: while (changelog_iteratator.next_changelog()) |entry|
+            {
+                // The release number can be null if it was tagged as "unreleased".
+                if (entry.release == null) continue;
+                break :changelog_previous entry;
+            } else unreachable;
```
kprotty: `while` loops act as their own block, so maybe a shorthand:
```suggestion
            const changelog_previous = 
                while (changelog_iterator.next_changelog()) |entry| {
                    // The release number can be null if it was tagged as "unreleased".
                    if (entry.release == null) continue;
                    break entry;
                } else unreachable;
```

kprotty (approved):

matklad:
The unreleased versions were intentionally not accepted. If an unreleased release happens, the _next_ release should include all the changes from the unreleased one. This is to make it easier for users of TigerBeetle to figure out what's actually new in a particular release. 

See https://github.com/tigerbeetle/tigerbeetle/blob/main/docs/about/internals/releases.md#skipping-release.

So let's perhaps keep rejecting unreleased, but emit a better error message in this case?

batiati:
Wouldn't it be better to merge all release notes then? So we don't miss the history (and of course the weekly TigerTrack!)

matklad:
Yeah, that's the process, you merge entries for two (or more) weeks into one!

PR #2472, opened 2024-11-11
clients: add application_logger and always log warn and err regardless

cb22:
This adds the scaffolding for log messages to be passed up into client languages and eventually handled by language-native logging libraries (eg, pino, log4j, etc).

No client library actually uses this yet, however (except for one heavily requested WIP client 🐍) - it's more to get the change out there and we can start adapting the clients.

As part of this, change the default back to logging .warn and .err messages from TigerBeetle. While it's not ideal for a library to be hijacking and printing things on stderr by itself, the current situation - where there's no logging at all - makes tracking down mistakes much harder.

This is hopefully a middle-ground until language clients all set their own log handlers.

### DIFF
--- a/src/clients/c/tb_client.zig
+++ b/src/clients/c/tb_client.zig
@@ -16,6 +16,11 @@ pub const tb_status_t = enum(c_int) {
     system_resources,
     network_subsystem,
 };
+pub const tb_register_log_callback_status_t = enum(c_int) {
+    success = 0,
+    already_registered,
+    not_registered,
+};
 
 pub const tb_operation_t = StateMachine.Operation;
 pub const tb_completion_t = *const fn (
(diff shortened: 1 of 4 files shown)


src/clients/c/tb_client.zig
```diff
@@ -49,6 +53,77 @@ const TestingContext = blk: {
...
+
+        callback(@intFromEnum(message_level), output.ptr, output.len);
+    }
+
+    callback: ?Callback = null,
+    mutex: std.Thread.Mutex = .{},
+    buffer: [log_line_max]u8 = undefined,
+};
+
+pub fn register_log_callback(
```
cb22, line 117: There's no way to unregister or register a new handler, currently.

src/clients/c/tb_client.zig
```diff
@@ -49,6 +53,77 @@ const TestingContext = blk: {
...
+        logging.mutex.lock();
+        defer logging.mutex.unlock();
+
+        const prefix = if (scope == .default) ": " else "(" ++ @tagName(scope) ++ "): ";
+        const output = std.fmt.bufPrint(
+            &logging.buffer,
+            prefix ++ format,
+            args,
+        ) catch |err| switch (err) {
+            error.NoSpaceLeft => blk: {
```
cb22, line 98: It's possible to do something more complex, like split the line ourselves, but 8192 chars should be enough for anybody.

src/clients/c/tb_client.zig
```diff
@@ -49,6 +53,77 @@ const TestingContext = blk: {
     break :blk ContextType(EchoClient);
 };
 
+/// Logging is global per application; it would be nice to be able to define a different logger for
+/// each client instance, though.
```
cb22, line 58: I don't think there's an easy way to do this, but I'd love to be wrong!

src/clients/c/tb_client.zig
```diff
@@ -49,6 +53,77 @@ const TestingContext = blk: {
...
+var logging: Logging = .{};
+
+pub const Logging = struct {
+    const Callback = *const fn (
+        message_level: u8,
+        message_ptr: [*]const u8,
+        message_len: usize,
+    ) callconv(.C) void;
```
batiati, line 66: Are we ok with the client library swallowing errors, or should this callback return `int` as an error code?

src/clients/c/tb_client.zig
batiati: Yes, Zig's logging infrastructure has no context, just a global function pointer. We could rely on the runtime specifics to store the context (such as napi/jni), but not for shared/static libs!

While it'd be really nice to isolate logging _per client_, the downside is that we wouldn't be able to log situations where the context isn't yet available (like, for example, creating the client itself).

So, I don't think it's worth trying, a global logger is good enough. Most importantly, when we wire this into the clients, we should consistently use this same logger to output any log emitted by the language side.

src/clients/c/tb_client.zig
```diff
@@ -49,6 +53,77 @@ const TestingContext = blk: {
...
+        comptime message_level: std.log.Level,
+        comptime scope: @Type(.EnumLiteral),
+        comptime format: []const u8,
+        args: anytype,
+    ) void {
+        // Messages are silently dropped if no logging callback is specified - unless they're warn
+        // or err. The value in having those for debugging is too high to silence them, even until
+        // client libraries catch up and implement a callback handler.
+        const callback = logging.callback orelse {
+            if (message_level == .warn or message_level == .err) {
```
batiati, line 81: I think we could let our `root` file specify the verbosity in all cases (so it's consistent with or without a custom handler).

After all, we do want to use `warn` only for cases that the operator must be aware of, so it should not be too verbose.

src/clients/c/tb_client.zig
batiati: We can get rid of the buffer if we wrap our callback in an `AnyWriter` interface, so the client library can write directly to its buffer.

```zig
// Signal the client that we are starting a new log entry (and lock its mutex if needed).
logging.log_entry_begin(message_level);
// Signal the client to flush the log entry (and release the mutex).
defer logging.log_entry_end();
// Use the client as a writer in the std.fmt.format API, so the client can manage its buffer.
std.fmt.format(logging.writer(), prefix ++ format, args);
```

src/clients/c/tb_client.zig
batiati: We could make the API simpler by allowing the user to replace the handler, or even setting it as `null` if they want to unregister.

src/clients/c/tb_client.zig
cb22: zig's `defaultLog` swallows writer errors, so I think it's ok for us to assume the client library will do the same (or deal with it internally) too.

src/clients/c/tb_client.zig
cb22: > I think we could let our root file specify the verbosity in all cases (so it's consistent with or without a custom handler).

Hmm - the downside is that it's then impossible to control the level at runtime, which is what we want most of the time. 

I actually have a WIP PR that does the opposite entirely - for replicas (and, it would be the same here) - making our default log level `.debug` and filtering at runtime (performance hit is negligible!) so that you can do something like `--log-debug` on the production binary, which is very useful.

Once this PR lands though and clients use it, I'd remove this custom handler when no callback is set entirely (or, make all messages go to stderr with a warning to fix your client). The client would potentially be able to enable `.debug` logs with an extra argument to `register_log_callback` (because the FFI call overhead could stack up) but if they wanted to filter .info or higher, that should be done with their logging library.

The only reason for this specific check for now is because we're being noisy on `stderr` which we really shouldn't be, so it's trying to minimize that.

src/clients/c/tb_client.zig
cb22: > We can get rid of the buffer if we wrap our callback in an AnyWriter interface, so the client library can write directly to its buffer.

That's pretty cool, but I think it might be better to buffer it in Zig explicitly and pass the final buffer - if we use a writer interface, we'll cross FFI for each format section as it writes it out.

(I think, if I'm understanding correctly!)

batiati: > if we use a writer interface, we'll cross FFI for each format section as it writes it out.

Yes, it might be a problem for interpreted languages, that we could mitigate by wrapping it into a `BufferedWriter` ... but in this case, it's just simpler to buffer the entire thing like we're doing!!!

> It's possible to do something more complex, like split the line ourselves, but 8192 chars should be enough for anybody.

If we ever want to address the buffer limit, that may be the way!

batiati (approved):
LGTM! :partying_face:

PR #2474, opened 2024-11-12
io: handle eagain by submitting again

cb22:
This fixes running TigerBeetle on XFS, which can return EAGAIN for reads. The exact source of the EAGAIN is... strange. Given that we don't open our file as O_NONBLOCK, nor use anything like RWF_NOWAIT.

It seems specific to XFS so far - but I haven't been able to track down exactly why it's happen yet. There seem to be enough places where EAGAIN is returned in XFS's code.

### DIFF
--- a/src/io/linux.zig
+++ b/src/io/linux.zig
@@ -607,11 +607,12 @@ pub const IO = struct {
                     const result: ReadError!usize = blk: {
                         if (completion.result < 0) {
                             const err = switch (@as(posix.E, @enumFromInt(-completion.result))) {
-                                .INTR => {
+                                .INTR, .AGAIN => {
+                                    // Some file systems, like XFS, can return EAGAIN even when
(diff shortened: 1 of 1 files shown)


sentientwaffle (approved):
LGTM!

Probably overkill, but we could also add a "safety counter" to `completion` to make sure we can't just endlessly retry if the os/fs is misbehaving.

cb22:
> Probably overkill, but we could also add a "safety counter" to completion to make sure we can't just endlessly retry if the os/fs is misbehaving.

Was thinking the same thing! Was also wondering if we might ever need this for `write()`... I haven't encountered it, but from the XFS source, it might indeed be possible to hit.

(but indeed, spinning on a retry that's always returning EAGAIN would be Very Bad™ - I'll add the counter and it for write() EAGAIN in a follow up)

PR #2475, opened 2024-11-12
[StateMachine] Save 256K of RAM by not having a prefetch cache for historical balances.

batiati:
The `prefetch_entries_for_update_max` parameter was set incorrectly, since we don't need to prefetch account balances as they are append-only.


sentientwaffle (approved):
Great catch!

PR #2478, opened 2024-11-13
mem: `mlockall` pages to prevent swap bypassing storage fault tolerance

kprotty:
Attempt to prevent swap from introducing disk errors into memory by locking all allocated pages before running the replica. On failure, hint to the operator that they should try to disable swap system-wide instead.

### DIFF
--- a/docs/operating/docker.md
+++ b/docs/operating/docker.md
@@ -170,7 +170,12 @@ docker run -p 3000:3000 -v $(pwd)/data:/data ghcr.io/tigerbeetle/tigerbeetle:deb
 #### `error: SystemResources`
 
 If you get `error: SystemResources` when running TigerBeetle in Docker on macOS,
-you will need to do one of the following:
+the container is blocking tigerbeetle from locking memory which is necessary for io_uring
+and optionally to avoid kernel swap bypassing TigerBeetle's storage fault tolerance.
+
+#### Allowing MEMLOCK 
+
+To raise the memory lock limits under Docker, execute one of the following:
 
 1. Run `docker run` with `--cap-add IPC_LOCK`
 2. Run `docker run` with `--ulimit memlock=-1:-1`
--- a/docs/operating/docker.md
+++ b/docs/operating/docker.md
@@ -170,8 +170,8 @@ docker run -p 3000:3000 -v $(pwd)/data:/data ghcr.io/tigerbeetle/tigerbeetle:deb
 #### `error: SystemResources`
 
 If you get `error: SystemResources` when running TigerBeetle in Docker on macOS,
-the container is blocking tigerbeetle from locking memory which is necessary for io_uring
-and optionally to avoid kernel swap bypassing TigerBeetle's storage fault tolerance.
+the container may be blocking TigerBeetle from locking memory, which is necessary both for io_uring
+and to prevent the kernel's use of swap from bypassing TigerBeetle's storage fault tolerance.
 
 #### Allowing MEMLOCK 
 
--- a/docs/operating/linux.md
+++ b/docs/operating/linux.md
@@ -141,8 +141,17 @@ In case you want to use this service for development as well, you may need to ad
 
 ### Memory Locking
 
-TigerBeetle relies on `RLIMIT_MEMLOCK` being adequate enough to initialize io_uring. It also locks all allocated memory to avoid the kernel swapping pages to disk and bypassing TigerBeetle's storage determinism. If enough memory cannot be locked, the environment can be appropriately modified either by disabling swap, adding the `CAP_IPC_LOCK` capability (provided by the systemd service), or raising the `memlock` value under `/etc/security/limits.conf`.
+TigerBeetle requires `RLIMIT_MEMLOCK` to be set high enough to:
 
-Memory locking is disabled for development environments using the `--development` flag.
+1. initialize io_uring, which requires memory shared with the kernel to be locked, as well as
+2. lock all allocated memory, and so prevent the kernel from swapping any pages to disk, which would not only affect performance but also bypass TigerBeetle's storage fault-tolerance.
 
-For linux running under Docker, refer to [Allowing MEMLOCK](docker.md#allowing-memlock).
+If the required memory cannot be locked, then the environment should be modified either by (in order of preference):
+
+1. giving the local `tigerbeetle` binary the `CAP_IPC_LOCK` capability, or
+2. raising the global `memlock` value under `/etc/security/limits.conf`, or else
+3. disabling swap (io_uring may still require an RLIMIT increase).
+
+Memory locking is disabled for development environments when using the `--development` flag.
+
+For Linux running under Docker, refer to [Allowing MEMLOCK](docker.md#allowing-memlock).
--- a/docs/operating/linux.md
+++ b/docs/operating/linux.md
@@ -148,7 +148,7 @@ TigerBeetle requires `RLIMIT_MEMLOCK` to be set high enough to:
 
 If the required memory cannot be locked, then the environment should be modified either by (in order of preference):
 
-1. giving the local `tigerbeetle` binary the `CAP_IPC_LOCK` capability, or
+1. giving the local `tigerbeetle` binary the `CAP_IPC_LOCK` capability (`sudo setcap "cap_ipc_lock=+ep" ./tigerbeetle`), or
 2. raising the global `memlock` value under `/etc/security/limits.conf`, or else
 3. disabling swap (io_uring may still require an RLIMIT increase).
 
--- a/docs/operating/linux.md
+++ b/docs/operating/linux.md
@@ -20,6 +20,8 @@ After=network-online.target
 Wants=network-online.target systemd-networkd-wait-online.service
 
 [Service]
+AmbientCapabilities=CAP_IPC_LOCK
+
 Environment=TIGERBEETLE_CACHE_GRID_SIZE=1GiB
 Environment=TIGERBEETLE_ADDRESSES=3001
 Environment=TIGERBEETLE_REPLICA_COUNT=1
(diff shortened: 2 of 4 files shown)


src/tigerbeetle/main.zig
```diff
@@ -446,6 +447,92 @@ const Command = struct {
...
+                    .NOMEM => log.warn(lock_mem_err, .{"total memory would exceed RLIMIT_MEMLOCK"}),
+                    .PERM => log.warn(lock_mem_err, .{"not enough privileges to lock memory"}),
+                    .INVAL => unreachable, // MCL_ONFAULT specified with MCL_CURRENT.
+                    else => |err| return stdx.unexpected_errno("mlockall", err),
+                }
+            },
+            .macos => {
+                // macOS has mlock() but not mlockall().
+            },
+            .windows => {
```
cb22: Does the Windows case need a call to [VirtualLock](https://learn.microsoft.com/en-us/windows/win32/api/memoryapi/nf-memoryapi-virtuallock)?

kprotty: `VirtualLock` takes in a range of memory, acting closer to `mlock` than `mlockall`. We could technically have our own `mmap/VirtualAlloc` bump allocator for main.start then mlock that allocated memory range. But this would exclude things like globals and stack so not as thorough as `SetProcessWorkingSetSize`.

docs/operating/docker.md
```diff
@@ -170,7 +170,12 @@ docker run -p 3000:3000 -v $(pwd)/data:/data ghcr.io/tigerbeetle/tigerbeetle:deb
 #### `error: SystemResources`
 
 If you get `error: SystemResources` when running TigerBeetle in Docker on macOS,
-you will need to do one of the following:
+the container is blocking tigerbeetle from locking memory which is necessary for io_uring
+and optionally to avoid kernel swap bypassing TigerBeetle's storage fault tolerance.
```
jorangreef:
```suggestion
the container may be blocking TigerBeetle from locking memory, which is necessary both for io_uring
and to prevent the kernel's use of swap from bypassing TigerBeetle's storage fault tolerance.
```

Capital "T". Capital "B".

docs/operating/linux.md
```diff
@@ -135,4 +137,12 @@ If you wish to change those, you are expected to understand those implications a
 
 The service was created assuming it'll be used in a production scenario.
 
-In case you want to use this service for development as well, you may need to adjust the `ExecStart` line to include the `--development` flag if your development environment doesn't support Direct IO, or if you require smaller cache sizes and/or batch sizes due to memory constraints.
\ No newline at end of file
+In case you want to use this service for development as well, you may need to adjust the `ExecStart` line to include the `--development` flag if your development environment doesn't support Direct IO, or if you require smaller cache sizes and/or batch sizes due to memory constraints.
+
+### Memory Locking
+
+TigerBeetle relies on `RLIMIT_MEMLOCK` being adequate enough to initialize io_uring. It also locks all allocated memory to avoid the kernel swapping pages to disk and bypassing TigerBeetle's storage determinism. If enough memory cannot be locked, the environment can be appropriately modified either by disabling swap, adding the `CAP_IPC_LOCK` capability (provided by the systemd service), or raising the `memlock` value under `/etc/security/limits.conf`.
```
jorangreef:
```suggestion
TigerBeetle requires `RLIMIT_MEMLOCK` to be set high enough to:

1. initialize io_uring, which requires memory shared with the kernel to be locked, as well as
2. lock all allocated memory, and so prevent the kernel from swapping any pages to disk, which would not only affect performance but also bypass TigerBeetle's storage fault-tolerance.

If the required memory cannot be locked, then the environment should be modified either by (in order of preference):

1. giving the local `tigerbeetle` binary the `CAP_IPC_LOCK` capability, or
2. raising the global `memlock` value under `/etc/security/limits.conf`, or else
3. disabling swap (io_uring may still require an RLIMIT increase).
```

It's not about storage determinism, but storage fault-tolerance. Let's use bullets to make these reasons and mitigations clearer. And let's rank the mitigations in the best order (disabling swap is more a last resort than a first resort?).

docs/operating/linux.md
```diff
@@ -135,4 +137,12 @@ If you wish to change those, you are expected to understand those implications a
...
+
+Memory locking is disabled for development environments using the `--development` flag.
```
jorangreef:
```suggestion
Memory locking is disabled for development environments when using the `--development` flag.
```

docs/operating/linux.md
```diff
@@ -135,4 +137,12 @@ If you wish to change those, you are expected to understand those implications a
...
+
+For linux running under Docker, refer to [Allowing MEMLOCK](docker.md#allowing-memlock).
```
jorangreef:
```suggestion
For Linux running under Docker, refer to [Allowing MEMLOCK](docker.md#allowing-memlock).
```

Linux is a proper noun.

src/tigerbeetle/main.zig
```diff
@@ -446,6 +447,110 @@ const Command = struct {
             watchdog.detach();
         }
 
+        if (!args.development) {
+            // Try to lock all memory in the process to avoid the kernel swapping pages to disk and
+            // potentially introducting undetectable disk corruption into memory.
+            // This is a best-effort attempt and not a hard rule as it may not cover all memory edge
+            // case. So warn on error to notify the operator to adjust conditions if possible.
+            const lock_mem_err = "Unable to lock pages in memory ({s})" ++
+                " - kernel swapping may bypass TigerBeetle's storage fault tolerance. ";
```
jorangreef:
```suggestion
                " - kernel swap would otherwise bypass TigerBeetle's storage fault tolerance. ";
```

src/tigerbeetle/main.zig
```diff
@@ -446,6 +447,110 @@ const Command = struct {
...
+
+            switch (builtin.os.tag) {
+                .linux => blk: {
+                    const MCL_CURRENT = 1; // Lock all currently mapped pages.
+                    const MCL_ONFAULT = 3; // Lock all pages faulted in (i.e. stack space).
+                    const rc = os.linux.syscall1(.mlockall, MCL_CURRENT | MCL_ONFAULT);
+                    switch (os.linux.E.init(rc)) {
```
jorangreef:
```suggestion
                    const result = os.linux.syscall1(.mlockall, MCL_CURRENT | MCL_ONFAULT);
                    switch (os.linux.E.init(result)) {
```

Per TigerStyle, don't abbreviate variable names.

src/tigerbeetle/main.zig
```diff
@@ -446,6 +447,110 @@ const Command = struct {
             watchdog.detach();
         }
 
+        if (!args.development) {
+            // Try to lock all memory in the process to avoid the kernel swapping pages to disk and
+            // potentially introducting undetectable disk corruption into memory.
+            // This is a best-effort attempt and not a hard rule as it may not cover all memory edge
+            // case. So warn on error to notify the operator to adjust conditions if possible.
+            const lock_mem_err = "Unable to lock pages in memory ({s})" ++
```
jorangreef:
```suggestion
            const mlockall_error = "Unable to lock pages in memory ({s})" ++
```

Per TigerStyle, don't abbreviate variable names. For example, write `error` not `err`.

Also, relate the error to the syscall, e.g. `mlockall`

src/tigerbeetle/main.zig
```diff
@@ -446,6 +447,110 @@ const Command = struct {
...
+                        .NOMEM => log.warn(lock_mem_err, .{"memory would exceed RLIMIT_MEMLOCK"}),
+                        .PERM => log.warn(lock_mem_err, .{"not enough privileges to lock memory"}),
+                        .INVAL => unreachable, // MCL_ONFAULT specified with MCL_CURRENT.
+                        else => |err| return stdx.unexpected_errno("mlockall", err),
+                    }
+                    log.warn(
+                        "If this is a production replica, consider either " ++
+                            "increasing the MEMLOCK process limit, " ++
+                            "running the replica with CAP_IPC_LOCK privilege, " ++
+                            "or disabling swap system-wide.",
```
jorangreef:
```suggestion
                        "If this is a production replica, consider either " ++
                            "running the replica with CAP_IPC_LOCK privilege, " ++
                            "increasing the MEMLOCK process limit, " ++
                            "or disabling swap system-wide.",
```

src/tigerbeetle/main.zig
```diff
@@ -446,6 +447,110 @@ const Command = struct {
...
+                        process_handle,
+                        working_set_min,
+                        working_set_max,
+                    ) == os.windows.FALSE) {
+                        // From std.os.windows.unexpectedError():
+                        const fmt_flags = os.windows.FORMAT_MESSAGE_FROM_SYSTEM |
+                            os.windows.FORMAT_MESSAGE_IGNORE_INSERTS;
+
+                        // 614 is the length of the longest windows error description
+                        var buf_wstr: [614:0]os.windows.WCHAR = undefined;
```
jorangreef: `buf_wstr` and `fmt_flags`

TigerStyle.

src/tigerbeetle/main.zig
```diff
@@ -446,6 +447,110 @@ const Command = struct {
...
+                        const len = os.windows.kernel32.FormatMessageW(
```
jorangreef:
```suggestion
                        const size = os.windows.kernel32.FormatMessageW(
```

Either `size` or `count` or `length`.

jorangreef (changes requested):

docs/operating/linux.md
```diff
@@ -135,4 +137,21 @@ If you wish to change those, you are expected to understand those implications a
...
+### Memory Locking
+
+TigerBeetle requires `RLIMIT_MEMLOCK` to be set high enough to:
+
+1. initialize io_uring, which requires memory shared with the kernel to be locked, as well as
+2. lock all allocated memory, and so prevent the kernel from swapping any pages to disk, which would not only affect performance but also bypass TigerBeetle's storage fault-tolerance.
+
+If the required memory cannot be locked, then the environment should be modified either by (in order of preference):
+
+1. giving the local `tigerbeetle` binary the `CAP_IPC_LOCK` capability, or
```
cb22: We could suggest the exact command: `sudo setcap "cap_ipc_lock=+ep" tigerbeetle`

cb22: (additionally, we could suggest it in the log warning message directly too.)

cb22 (commented):
Seems to work well on Linux and confirmed with `sudo cat /proc/{pid}/smaps`!

docs/operating/linux.md
kprotty: I think suggesting in docs is better as it also has the commands in Docker to raise rlimits too.

cb22:
@kprotty LGTM, just needs spell checking fixed.

cb22 (approved):

jorangreef (approved):

PR #2481, opened 2024-11-14
[Java] Expose the VSR timestamp to the client

batiati:
For diagnostics and debug purposes, this PR wires the VSR timestamp to all the language clients and exposes it in the Java Client API.

Reference: #2440

Usage:
```java
var results = client.createTransfers(batch);
long timestamp = results.getHeader().getTimestamp()
```


sentientwaffle (approved):

matklad:
Thanks for tackling this Rafael!

PR #2484, opened 2024-11-15
Client: Don't panic on eviction

sentientwaffle:
Language clients no longer panic when they receive an eviction message from the cluster.
Instead, any in-flight (or future) packets fail with `PacketStatus.ClientEvicted`.

Note that for simplicity the client eviction reason is not made available.

Rather than plumbing the eviction reason around (e.g. by turning `PacketStatus` into a `union(enum)`?), we can use [logging](https://github.com/tigerbeetle/tigerbeetle/pull/2472) to inform the operator about the details.


src/clients/java/src/main/java/com/tigerbeetle/RequestException.java
batiati, line 1: **Note to self:** `RequestException` does indicate user-facing exceptions such as `TooMuchData`, `ClientShutdown` and now `ClientEvicted`, on the contrary to what is stated in the comment "_thrown when the internal invariants of the underlying native library are violated, which is expected to never happen._"

This documentation must be updated, but maybe this exception should be a checked exception explicitly handled by the user, reverting https://github.com/tigerbeetle/tigerbeetle/commit/5c93c3c898e25b1817b0ae37431f7dc24242bc91. 



https://github.com/tigerbeetle/tigerbeetle/blob/5e9a0a201f80ed801523d2831e6b1ffc9d865435/src/clients/java/src/main/java/com/tigerbeetle/RequestException.java#L3-L9

src/clients/java/src/test/java/com/tigerbeetle/IntegrationTest.java
```diff
@@ -940,6 +940,37 @@ public void testCreateTransferTooMuchDataAsync() throws Throwable {
         }
     }
 
+    @Test
+    public void testClientEvicted() throws Throwable {
```
batiati, line 944: :heart:

batiati (approved):
>Note that for simplicity the client eviction reason is not made available.
>
>Rather than plumbing the eviction reason around (e.g. by turning PacketStatus into a union(enum)?), we can use https://github.com/tigerbeetle/tigerbeetle/pull/2472 to inform the operator about the details.

Alternatively, we could select a few eviction reasons that are worth exposing to the client and represent them as distinguished `PacketStatus` errors.
For example:

```zig
const PacketError = error{
    TooMuchData,
    ClientShutdown,
    ClientReleaseIncompatible, // Evicted due to incompatible releases, should upgrade.
    ClientEvicted, // The server kicked us, we may try connecting again.
    InvalidOperation,
    InvalidDataSize,
};
```

sentientwaffle:
> Alternatively, we could select a few eviction reasons that are worth exposing to the client and represent them as distinguished PacketStatus errors.

Great suggestion! I will implement that in a follow-up PR, so that this PR is included in monday's release.

sentientwaffle:
Looks like this PR introduced a flake to the integration tests, so it won't be in monday's release after all. I'll investigate after the weekend!

PR #2485, opened 2024-11-18
Vortex: custom network fault injection

owickstrom:
This puts a TCP proxy in front of each replica, which intercepts all communication and injects basic network faults (packet loss, corruption, and delays). By implementing our own network faults like this, we get rid of the dependency on Linux `tc`/`netem`, and technically even `unshare(1)`, although we might still run tests in unshare to clean up processes and avoid port collisions.

The proxy is currently not message-aware, meaning that we don't collect full messages before we forward them; any chunk of bytes recvd, whatever length, is sent immediately (in case of no fault injection).

Faults are configured from the outside by the Vortex supervisor. Also, the proxy uses async IO, so the supervisor now loops and ticks the IO (`run_for_ns`).

A new action `quiesce` is introduced, which is a longer sleep that also makes sure there are no ongoing faults. This lets the cluster recover before starting new rounds of faults.

Tests can now be run with:

```
zig build -Drelease install
zig build -Drelease vortex
unshare -nfrp bash -c 'ip link set up dev lo ; zig-out/bin/vortex supervisor --tigerbeetle-executable=./zig-out/bin/tigerbeetle --test-duration-minutes=10'
```

Future work:

* Duplicated writes
* Misdirected writes
* More interesting partitioning, instead of just global packet loss
* Run on macOS and Windows
* Possibly make it message-aware, to inject more interesting faults?


src/integration_tests.zig
```diff
@@ -510,15 +510,23 @@ test "vortex smoke" {
...
+        \\   {s} supervisor \
+        \\   --test-duration-minutes=1 \
+        \\   --tigerbeetle-executable={s}
     , .{
-        .vortex = vortex_exe,
-        .tigerbeetle = tigerbeetle,
+        vortex_exe,
+        tigerbeetle,
+    });
+    _ = try shell.file_ensure_content(script_path, script_contents, .{ .mode = 0o777 });
```
cb22: If you're writing directly, I think you can use `shell.cwd.writeFile` rather?

src/integration_tests.zig
```diff
@@ -510,15 +510,23 @@ test "vortex smoke" {
...
-        \\   {vortex} supervisor
-        \\      --test-duration-minutes=1
-        \\      --tigerbeetle-executable={tigerbeetle}
+    const script_path = try shell.fmt("./.zig-cache/tmp/{}.sh", .{
+        std.crypto.random.int(u64),
+    });
+    defer shell.cwd.deleteTree(script_path) catch {};
+
+    const script_contents = try shell.fmt(
+        \\ ip link set up dev lo && \
```
cb22, line 519: Since we're now doing the network fault injection, we don't need unshare with `--net` right? So we could remove this script and the network namespace entirely?

src/testing/vortex/faulty_network.zig
```diff
@@ -0,0 +1,583 @@
...
+//!
+//! A pipe runs, alternating recv and send, until it sees that the connection is no longer proxying
+//! or if there's an error or EOF, in which case it also closes the parent connection.
+//!
+//! When the `Faults` struct is populated with non-null configurations, the pipes inject
+//! corresponding faults according to the probabilities.
+//!
+//! NOTE: The pipe is not yet message-aware (and perhaps shouldn't be?), which means that we deal
+//! with whatever chunks of bytes we receieve immediately, instead of collecting a buffer with a
+//! full message before piping it through.
```
cb22, line 20: As a note, the sort of faulting we do here would appear very differently to what underlying network packet loss / corruption / latency would do to TCP, since we're simulating those things on top of TCP rather than below TCP (and seeing how the TCP state machine reacts too).

I don't think it's a problem, or that there's an easy solution, but something interesting to think about!

src/testing/vortex/faulty_network.zig
```diff
@@ -0,0 +1,583 @@
...
+        pipe.recv_inflight = true;
+
+        pipe.connection.io.recv(
+            *Pipe,
+            pipe,
+            on_recv,
+            &pipe.recv_completion,
+            pipe.input,
+            pipe.buffer[0..],
+        );
```
cb22, line 117: Ah, I guess we don't need to handle partial recv as anything special, because we are a proxy - if we recv partially and send all that on, that's OK. Can you make a comment about that? (I was looking for recv all logic)

src/testing/vortex/faulty_network.zig
```diff
@@ -0,0 +1,583 @@
...
+                pipe.connection.replica_index,
+                pipe.connection.connection_index,
+            });
+            pipe.io.timeout(
+                *Pipe,
+                pipe,
+                on_timeout,
+                &pipe.send_completion,
+                timeout_duration_ns,
+            );
```
cb22, line 198: Does Vortex pull in anything from vsr? Maybe consider making this use timeouts and calling them from `.tick()`, but it's probably fine as-is!

src/testing/vortex/faulty_network.zig
```diff
@@ -0,0 +1,583 @@
...
+    } = .free,
+
+    replica_index: usize,
+    connection_index: usize,
+
+    origin_fd: std.posix.socket_t = IO.INVALID_SOCKET,
+    remote_fd: std.posix.socket_t = IO.INVALID_SOCKET,
+
+    origin_to_remote_pipe: Pipe,
+    remote_to_origin_pipe: Pipe,
```
cb22, line 271: The symmetry 🤌

src/integration_tests.zig
owickstrom: We could, but I wanted to keep it to avoid port clashes. Perhaps we can use port=0 and get parse the assigned ports, but it would complicate the setup in supervisor a lot. Maybe for another PR?

owickstrom: We could also use some less likely bound ports (i.e. not 3000)?

src/testing/vortex/faulty_network.zig
owickstrom: Yeah. I haven't looked into how netem does it, but I'm assuming it's much more realistic than this TCP proxy-based take. Could the proxy be written for a lower layer perhaps?

src/testing/vortex/faulty_network.zig
owickstrom: Yeah, exactly. I'll add that.

src/testing/vortex/faulty_network.zig
owickstrom: I don't think so, no. This timeout is only to simulate random network delays, so it's not coupled with VSR in any sense.

src/testing/vortex/faulty_network.zig
owickstrom: Names were carefully selected.

src/integration_tests.zig
owickstrom: Good catch, this is probably a residue from less random times.

src/integration_tests.zig
cb22: Ah yes, the port clashes - think that's good for another PR then :+1:.

src/testing/vortex/faulty_network.zig
cb22: Hmm - not without getting crazy, I think :smile: perhaps with a custom TUN/TAP pair or something similar.

cb22 (approved):

src/testing/vortex/faulty_network.zig
owickstrom: A rabbit hole if there ever was one...

matklad: >I don't think it's a problem, or that there's an easy solution, but something interesting to think about!

I think we can just use a RAW socket to mess up at the IP layer? I think it'll require the faulty_network to understand some parts of the TCP state machine, but it shouldn't be anything crazy in terms of the kernel APIs used?

PR #2486, opened 2024-11-18
tidy: ban `!comptime ...`

cb22:
Sometimes, you need to reach for an explicit `comptime` in an if condition:
```
if (comptime foo) {
    std.log.info("foo!", .{});
}
```

and sometimes you need to negate that:

```
if (!comptime foo) {
    std.log.info("not foo!", .{});
}
```

all good so far. The problem is that this comptime expression spans across the entire condition, so:

```
if (!comptime foo and bar) {
    std.log.info("not foo, bar!", .{});
}
```

is really `!comptime (foo and bar)` and not `!(comptime (foo)) and bar`.

### DIFF
--- a/src/tidy.zig
+++ b/src/tidy.zig
@@ -160,6 +160,10 @@ fn tidy_banned(source: []const u8) ?[]const u8 {
         return "use a type name instead of Self";
     }
 
+    if (std.mem.indexOf(u8, source, "!" ++ "comptime") != null) {
+        return "use ! inside comptime";
+    }
+
     return null;
 }
 
--- a/src/tigerbeetle/inspect.zig
+++ b/src/tigerbeetle/inspect.zig
@@ -842,7 +842,7 @@ fn print_struct(
 
     const Type = @typeInfo(@TypeOf(value)).Pointer.child;
     // Print structs *without* a custom format() function.
-    if (@typeInfo(Type) == .Struct and !comptime std.meta.hasFn(Type, "format")) {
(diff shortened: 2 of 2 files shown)


src/tigerbeetle/inspect.zig
```diff
@@ -842,7 +842,7 @@ fn print_struct(
 
     const Type = @typeInfo(@TypeOf(value)).Pointer.child;
     // Print structs *without* a custom format() function.
-    if (@typeInfo(Type) == .Struct and !comptime std.meta.hasFn(Type, "format")) {
+    if (@typeInfo(Type) == .Struct and comptime !std.meta.hasFn(Type, "format")) {
```
matklad:
```suggestion
    if (comptime @typeInfo(Type) == .Struct and !std.meta.hasFn(Type, "format")) {
```

Seems more readable this way -- the whole condition is intended to be comptime!

src/tidy.zig
```diff
@@ -160,6 +160,10 @@ fn tidy_banned(source: []const u8) ?[]const u8 {
         return "use a type name instead of Self";
     }
 
+    if (std.mem.indexOf(u8, source, "!" ++ "comptime") != null) {
+        return "use ! inside comptime, or wrap expression in brackets";
```
matklad:
```suggestion
        return "use ! inside comptime";
```

Seems reasonable to require that comptime is  the topmost expression!

matklad (approved):

PR #2487, opened 2024-11-18
Python client

cb22:
As the title suggests, add a TigerBeetle Python client! 

The client supports both sync and async and is backed by `ctypes` with code generation from Zig. It has no runtime dependencies, and bundles in the C libraries - much like the Go client. It should support alternative Python runtimes, like PyPy, but these are currently untested.

`dataclasses` and `enum` are used internally, so you get pretty output and helpful bitflags and enums out the box. There are also type hints. It's possible to do a lot of magic with Python, but we went with an approach of mainly doing codegen for now.

The performance of the client can likely be improved significantly, especially by having a CPython extension to handle the marshalling between the TigerBeetle library and the Python objects. Currently all that work is done in Python itself. That said, some basic benchmarks with the echo client show the performance hit for a full batch to be in the low-ms range.

```python
from tigerbeetle_python import ClientSync, Account, QueryFilter

client = ClientSync(cluster_id=0, replica_addresses="3000")

client.create_accounts([Account(id=1, ledger=1, code=1)])
# []

client.create_accounts([Account(id=1, ledger=1, code=1)])
# [CreateAccountsResult(index=0, result=<CreateAccountResult.EXISTS: 21>)]

client.query_accounts(QueryFilter(limit=1))
# [Account(id=1, debits_pending=0, debits_posted=0, credits_pending=0, credits_posted=0, user_data_128=0, user_data_64=0, user_data_32=0, ledger=1, code=1, flags=<AccountFlags.NONE: 0>, timestamp=1731942817744061130)]
```


src/clients/python/tests/test_basic.py
```diff
@@ -0,0 +1,1348 @@
+import os
```
cb22, line 1: This whole file is basically a 1-to-1 port of the node client's test.ts

src/clients/python/tests/test_basic.py
```diff
@@ -0,0 +1,1405 @@
...
+
+def test_accept_zero_length_lookup_accounts(client):
+    accounts = client.lookup_accounts([])
+    assert accounts == []
+
+def test_accept_zero_length_lookup_transfers(client):
+    transfers = client.lookup_transfers([])
+    assert transfers == []
+
+def test_uint128(client):
```
cb22, line 1321: Except for this test!

src/clients/python/pyproject.toml
```diff
@@ -0,0 +1,39 @@
...
+
+[build-system]
+requires = ["hatchling"]
+build-backend = "hatchling.build"
+
+[tool.hatch.build.targets.sdist]
+only-include = ["tigerbeetle_python"]
+artifacts = ["*.so", "*.dylib", "*.dll"]
+
+[tool.hatch.build.targets.sdist.force-include]
```
cb22, line 30: This is used to convert our directory structure (README.md in root) to what Python wants for packagaing (README.md inside package).

src/clients/python/samples/basic/README.md
```diff
@@ -0,0 +1,61 @@
...
+* Python (or PyPy, etc) >= `3.7`
+
+## Setup
+
+First, clone this repo and `cd` into `tigerbeetle/src/clients/python/samples/basic`.
+
+Then, install the TigerBeetle client:
+
+```console
+pip install tigerbeetle_python
```
cb22: To confirm this the name we want to go with - not eg just `tigerbeetle`.

src/clients/python/src/tigerbeetle_python/client.py
```diff
@@ -0,0 +1,226 @@
...
+    response: Any
+    operation: bindings.Operation
+    c_event_type: Any
+    c_result_type: Any
+    on_completion: Callable | None
+    on_completion_context: CompletionContextSync | CompletionContextAsync | None
+
+
+
+def tbid() -> int:
```
cb22: I dropped the strictly monotonic increasing timestamp parts from the Node client... I'm not sure they're worth it? I can put them back.

src/clients/python/python_bindings.zig
```diff
@@ -0,0 +1,502 @@
...
+            // For ergonomics, the client allows calling things like .query_accounts(filter) even
+            // though the _submit function requires a list for everything. Wrap them here.
+            const event_name_or_list = comptime if (!StateMachine.event_is_slice(op))
+                "[" ++ StateMachine.event_name(op) ++ "]"
+            else
+                StateMachine.event_name(op);
+
+            // TODO: Pulse shouldn't be hardcoded.
+            if (op != .pulse) {
+                buffer.print(
```
cb22: This block is pretty dense and almost unreadable here. Between the high levels of indentation and parameters. Still trying to figure out something nicer.

src/clients/python/samples/basic/README.md
cb22: After chatting with Joran, we're going to go the opposite direction! All clients will just be `tigerbeetle` in their respective languages. I'll hold off renaming this one though, until you're done with review.

src/state_machine.zig
```diff
@@ -386,6 +386,66 @@ pub fn StateMachineType(
...
+                .create_transfers => "transfers",
+                .lookup_accounts => "accounts",
+                .lookup_transfers => "transfers",
+                .get_account_transfers => "filter",
+                .get_account_balances => "filter",
+                .query_accounts => "query_filter",
+                .query_transfers => "query_filter",
+            };
+        }
+
```
cb22, line 448: To add some context around these - the ideal goal would be for a state machine to define everything it exposes, in a way that's both checked at compile time (so you can't forget something off...) and makes it possible for a client bindings generator to generate bindings entirely from that description.

The current Python generator still relies on hardcoded types, `mappings_state_machine`, so it's not quite there yet.

(as to the how, we could hoist the execution logic of `commit()` up into VSR, so that calls into specific functions in the state machine, based on what it exposed. This would also give better type safety than `input: []align(16) const u8`)

cb22: But... Out of scope for this PR!

docs/COMMUNITY_PROJECTS.md
```diff
@@ -23,6 +23,6 @@ clients may lag behind.
 * [tigerbeetle-cpp](https://github.com/kassane/tigerbeetle-cpp): A C++ client
 * [tigerbeetle-rs](https://github.com/ZetaNumbers/tigerbeetle-rs): A Rust client
 
-Need a client for Go, Java, Node.js, or .NET? We've got you covered
+Need a client for Go, Java, Node.js, .NET or Python? We've got you covered
```
matklad:
```suggestion
Need a client for Go, Java, Node.js, .NET, or Python? We've got you covered
```

src/clients/python/pyproject.toml
```diff
@@ -0,0 +1,39 @@
...
+"src/tigerbeetle_python" = "tigerbeetle_python"
+"README.md" = "tigerbeetle_python/README.md"
+
+[tool.hatch.build.targets.wheel]
+only-include = ["tigerbeetle_python"]
+artifacts = ["*.so", "*.dylib", "*.dll"]
+
+[tool.ruff]
+line-length = 100
```
matklad, line 39: Is default 80?

If it is, I'd maybe suggest sticking to it, in explicit violation of TigerStyle --- our users might look at the library source code, so having it formatting according to pep8 would be beneficial

src/clients/python/samples/basic/main.py
```diff
@@ -0,0 +1,49 @@
+import os
+
+from tigerbeetle_python import Account, ClientSync, Transfer
```
matklad: I _think_ idiomatic Python uses qualified names? `import numpy as np`?

Maybe let's create a precedent here for `import tigerbeetle as tb`?

src/clients/python/samples/two-phase-many/main.py
```diff
@@ -0,0 +1,261 @@
...
+  id=6,
+  debit_account_id=1,
+  credit_account_id=2,
+  amount=100,
+  pending_id=1,
+  ledger=1,
+  code=1,
+  flags=TransferFlags.POST_PENDING_TRANSFER,
+)
+])
```
matklad: Hm, is this the right formatting? Seems like a level of indentation is lost here. Or is it indeed the true style?

src/clients/python/samples/walkthrough/main.py
```diff
@@ -0,0 +1,554 @@
...
+    # endsection:create-transfers-errors
+except:
+    raise
+
+try:
+    # section:no-batch
+    batch = [] # Array of transfer to create.
+    for transfer in batch:
+        transfer_errors = client.create_transfers(transfer)
+        # Error handling omitted.
```
matklad: list comprehension would be more idiomatic here

src/clients/python/src/tigerbeetle_python/bindings.py
```diff
@@ -0,0 +1,733 @@
...
+    INVALID_OPERATION = 4
+    INVALID_DATA_SIZE = 5
+
+
+class CPacket(ctypes.Structure):
+    @classmethod
+    def from_param(cls, obj):
+        if obj.operation > 2**8 - 1:
+            raise IntegerOverflowError("operation is too large to fit \
+in 8 bits (" + str(obj.operation) + ")")
```
matklad: could use an f-string

src/clients/python/src/tigerbeetle_python/bindings.py
```diff
@@ -0,0 +1,733 @@
...
+    CLIENT_EVICTED = 2
+    CLIENT_SHUTDOWN = 3
+    INVALID_OPERATION = 4
+    INVALID_DATA_SIZE = 5
+
+
+class CPacket(ctypes.Structure):
+    @classmethod
+    def from_param(cls, obj):
+        if obj.operation > 2**8 - 1:
```
matklad: Do we need to check that it is non-negative as well? If we do, then I think we could use a helper function for validation.

matklad: Ahhh, let's maybe do 

```
operation = ctypes.c_int8(obj.operation)
```

and rely on the build-in range checks?

I don't know what's the right way to handle duck typing here, but this way we offload that question to Python's stdlib!

src/clients/python/src/tigerbeetle_python/bindings.py
```diff
@@ -0,0 +1,733 @@
...
+    ("code", ctypes.c_uint16),
+    ("reserved", ctypes.c_uint8 * 6),
+    ("timestamp_min", ctypes.c_uint64),
+    ("timestamp_max", ctypes.c_uint64),
+    ("limit", ctypes.c_uint32),
+    ("flags", ctypes.c_uint32),
+]
+
+
+@dataclasses.dataclass
```
matklad: Let's maybe use a frozen instance here?

https://docs.python.org/3/library/dataclasses.html#frozen-instances

matklad: And I _think_ we might also want to pass `slots=true`? There's going to be a lot of these, so might as well optimize memory usage!

src/clients/python/src/tigerbeetle_python/client.py
```diff
@@ -0,0 +1,224 @@
...
+class PacketError(Exception):
+    pass
+
+
+class Client:
+    _clients: dict[int, Any] = {}
+    _counter = AtomicInteger()
+
+    def __init__(self, cluster_id: int, replica_addresses: str):
+        self._client_id = Client._counter.increment()
```
matklad: whoa, shouldn't this be a random number?

matklad: or is it a different kind of client id?

matklad: Ahhh, it _is_ different, perhaps call it `client_key` then?

src/clients/python/src/tigerbeetle_python/client.py
matklad: Maybe it is worth it? If time goes backwards due to ntp adjustment or what not, we will get a latency spike as we'd have to go all the way to the bottom of LSM to disconfirm transfer existence.

src/clients/python/src/tigerbeetle_python/client.py
```diff
@@ -0,0 +1,224 @@
...
+        packet_id = Client._counter.increment()
+
+        packet = bindings.CPacket()
+        packet.next = None
+        packet.user_data = packet_id
+        packet.operation = operation
+        packet.status = bindings.PacketStatus.OK
+
+        operations_array_type = c_event_type * len(operations)
+        operations_array = operations_array_type(*[c_event_type.from_param(operation) for operation in operations])
```
matklad:
```suggestion
        operations_array = operations_array_type(*(c_event_type.from_param(operation) for operation in operations))
```

to avoid materializing needless array

src/clients/python/src/tigerbeetle_python/client.py
```diff
@@ -0,0 +1,224 @@
...
+                        c_event_type: Any, c_result_type: Any) -> InflightPacket:
+        packet_id = Client._counter.increment()
+
+        packet = bindings.CPacket()
+        packet.next = None
+        packet.user_data = packet_id
+        packet.operation = operation
+        packet.status = bindings.PacketStatus.OK
+
+        operations_array_type = c_event_type * len(operations)
```
matklad: oh dear, Python is so Pythonic!

src/clients/python/src/tigerbeetle_python/client.py
matklad: or maybe a shorter `*map(c_event_type.from_param, operations)`

src/clients/python/src/tigerbeetle_python/lib.py
```diff
@@ -0,0 +1,68 @@
+import ctypes
+import platform
+from pathlib import Path
+
+
+class NativeError(Exception):
+    pass
+
+class IntegerOverflowError(Exception):
```
matklad:
```suggestion
class IntegerOverflowError(ValueError):
```

src/clients/python/src/tigerbeetle_python/lib.py
```diff
@@ -0,0 +1,68 @@
...
+        prefix = "lib"
+        system = "macos"
+        suffix = ".dylib"
+    elif platform.system() == "Windows":
+        system = "windows"
+        suffix = ".dll"
+    else:
+        raise NativeError("Unsupported system: " + platform.system())
+
+    source_path = Path(__file__).resolve()
```
matklad: Would it work without `.resovle()`? The rule of thumb for paths is to avoid canonicalization, and use them as is.

matklad (commented):
First pass!

src/clients/python/src/tigerbeetle_python/bindings.py
cb22: > Do we need to check that it is non-negative as well

Yes, good catch!

> and rely on the build-in range checks?

```
>>> ctypes.c_uint8(-10000000)
c_ubyte(128)
```

These are not the [built-in range checks](https://docs.python.org/3/library/ctypes.html#ctypes.c_byte) you're looking for.

src/clients/python/src/tigerbeetle_python/bindings.py
cb22: > Let's maybe use a frozen instance here?

Hmm, I'm torn on this - I feel like it might make things harder, especially things like building the object as you go which might have a valid use case.

> And I think we might also want to pass slots=true? There's going to be a lot of these, so might as well optimize memory usage!

Yeah I think it's worth it - `slots` was only added in 3.10, but we can add it if supported.

src/clients/python/src/tigerbeetle_python/client.py
cb22: Heh, I was extra confused here for a sec, because initially this was a random number. But yes, it is a key for looking up the client (a similar pattern is used for packets too) internally.

docs/COMMUNITY_PROJECTS.md
cb22: Is the Oxford comma part of TigerStyle? :laughing:

src/clients/python/pyproject.toml
cb22: > Is default 80?

PEP8 says 79. Black, ruff and a whole bunch of other tools default to 88. It is relatively common to see projects deviate from this style rule but keep the rest though.

Personally I like 100, but I'm happy to change it if you think so!

src/clients/python/samples/two-phase-many/main.py
cb22: Nope - not quite sure what happened here. The default linter / formatter for `hatch` should be covering this.

src/clients/python/samples/walkthrough/main.py
cb22: It would be normally, but I think the straight for loop is better for showing the "don't do this" nature of the example?

cb22: (although, that code won't actually work if batches is non-empty, because create_transfers expects a list!)

.github/workflows/ci.yml
```diff
@@ -147,6 +153,13 @@ jobs:
         with:
           node-version: ${{ matrix.language_version }}
 
+      - if: matrix.language == 'python'
+        uses: actions/setup-python@v4
+        with:
+          python-version: ${{ matrix.language_version }}
+      - if: matrix.language == 'python'
+        run: pip install build hatch pytest
```
matklad, line 163: Should we use `uv`?

Pros:

* I _think_ we'd only need to install uv on GHA, and all the other deps _should_ get fetched automatically
* uv genuinely feels like the last attempt to fix Python packaging

Cons:

* `uv` is a startup, so future is uncertain!
* outside view tells us that `uv` will not be the last attempt to fix Python packaging

I guess, I'd say let's stick to vanilla tools for now perhaps, and re-evaluate if we need more deps.

.github/workflows/ci.yml
```diff
@@ -105,12 +105,18 @@ jobs:
           - { os: 'ubuntu-latest',  language: 'node',    language_version: '18.x'  }
           - { os: 'ubuntu-latest',  language: 'node',    language_version: '20.x'  }
 
+          # Support Python 3.7 explicitly, even though it's EOL.
```
matklad, line 108: <3

I love how we generally try to support older stuff, that builds trust!

.github/workflows/ci.yml
matklad: And, to double check, most of these deps (build, hatch, twine) are pypa, so official. 

pytest is third party, but it is pytest, we _certainly_ should be using it, it's worth violating no deps rule for it.

build.zig
```diff
@@ -70,6 +70,7 @@ pub fn build(b: *std.Build) !void {
         .clients_go = b.step("clients:go", "Build Go client shared library"),
         .clients_java = b.step("clients:java", "Build Java client shared library"),
         .clients_node = b.step("clients:node", "Build Node client shared library"),
+        .clients_python = b.step("clients:python", "Build Python client library"),
```
matklad, line 73: As a general style note, our clients are always sorted lexicographically, so dotnet goes first, and python goes last.

src/scripts/release.zig
```diff
@@ -449,6 +456,87 @@ fn build_node(shell: *Shell, info: VersionInfo, dist_dir: std.fs.Dir) !void {
...
 
+fn build_python(shell: *Shell, info: VersionInfo, dist_dir: std.fs.Dir) !void {
+    var section = try shell.open_section("build python");
+    defer section.close();
+
+    try shell.pushd("./src/clients/python");
+    defer shell.popd();
+
+    const python_version = shell.exec_stdout("python --version", .{}) catch {
+        return error.NoNode;
```
matklad:
```suggestion
        return error.NoPython;
```

src/scripts/release.zig
```diff
@@ -449,6 +456,87 @@ fn build_node(shell: *Shell, info: VersionInfo, dist_dir: std.fs.Dir) !void {
...
+
+    try backup_create(shell.cwd, "pyproject.toml");
+    defer backup_restore(shell.cwd, "pyproject.toml");
+
+    const pyproject = try shell.cwd.readFileAlloc(
+        shell.arena.allocator(),
+        "pyproject.toml",
+        1024 * 1024,
+    );
+    const version_line = try std.fmt.allocPrint(
```
matklad: `shell.fmt`

src/scripts/release.zig
```diff
@@ -590,6 +678,7 @@ fn publish(
         // Our docs are build with node, so publish the docs together with the node package.
         try publish_docs(shell, info);
     }
+    if (languages.contains(.python)) try publish_python(shell, info);
```
matklad, line 652: Could you move docs publishing after the python, so that, if the docs fails, everything else is still released?

src/scripts/release.zig
```diff
@@ -734,6 +823,22 @@ fn publish_node(shell: *Shell, info: VersionInfo) !void {
...
+fn publish_python(shell: *Shell, info: VersionInfo) !void {
+    var section = try shell.open_section("publish python");
+    defer section.close();
+
+    assert(try shell.dir_exists("zig-out/dist/python"));
+
+    _ = try shell.env_get("TWINE_USERNAME");
+    _ = try shell.env_get("TWINE_PASSWORD");
+
+    try shell.exec("python3 -m twine upload {package}", .{
```
matklad, line 810: This is the only time we spell it as python3. Everywhere else we spell it python. Let's be consistent? And moderate preference for `python3`, as that's I think the standard (though, I have vague memories of arch _not_ having python3 alias?)

src/clients/python/ci.zig
```diff
@@ -0,0 +1,116 @@
...
+        });
+        copied_count = 0;
+        for (native_files) |native_file| {
+            try Shell.copy_path(shell.cwd, native_file, lib_dir, native_file);
+            copied_count += 1;
+        }
+        // 7 = 4 + 3
+        //     4 = x86_64 for mac, windows and {linux musl, linux glibc}
+        //         3 = aarch64 for mac and {linux musl, linux glibc}
+        assert(copied_count == 7);
```
matklad: Hm, I think all other languages do native library copying step as a part of `zig build client:$lang`, should we do the same here? You can use `b.addInstallFile` I think to copy stuff over.

matklad: Eg, GO:

```
        step_clients_go.dependOn(&b.addInstallFile(
            lib.getEmittedBin(),
            b.pathJoin(&.{ "../src/clients/go/pkg/native/", name, lib.out_filename }),
        ).step);

```

src/state_machine.zig
matklad: I wonder if we should extract meta information out of state machine, and include it in tigerbeetle.zig? So that python codegen and state machine depend on tigerbeetle.zig, but python doesn't depend on the state machine directly. 

The fact that we instantiate IO in our python bindings generator is a smell --- something's not factored quite right. But it's not clear _what_ is the right factoring, so definitely for the follow up!

src/clients/python/python_bindings.zig
```diff
@@ -0,0 +1,497 @@
...
+
+    pub fn print(self: *Buffer, comptime format: []const u8, args: anytype) void {
+        self.inner.writer().print(format, args) catch unreachable;
+    }
+};
+
+fn mapping_name_from_type(mappings: anytype, Type: type) ?[]const u8 {
+    inline for (mappings) |mapping| {
+        const ZigType = mapping[0];
+        const name = mapping[1];
```
matklad:
```suggestion
        const ZigType, const name = mapping;
```

matklad: and probably could be comptime for rather than inline? `comptime` is better as it is more specific (every comptime for can be an inline one, but not every inline for can be replaced with comptime for)

src/clients/python/python_bindings.zig
```diff
@@ -0,0 +1,497 @@
...
+                128 => "c_uint128",
+                else => @compileError("invalid int type"),
+            };
+        },
+        .Optional => |info| switch (@typeInfo(info.child)) {
+            .Pointer => return resolve_type_ctypes(info.child),
+            else => @compileError("Unsupported optional type: " ++ @typeName(Type)),
+        },
+        .Pointer => |info| {
+            assert(info.size != .Slice);
```
matklad: Let's state the invariant positively by switching on `info.size` and explicitly listing allowed variants. That way, if Zig adds a new variant to Size, we'll get notified instead of it being treated silently as non slice

src/clients/python/python_bindings.zig
```diff
@@ -0,0 +1,497 @@
...
+
+            if (Type == *anyopaque) {
+                return "ctypes.c_void_p";
+            }
+
+            return comptime "ctypes.POINTER(C" ++
+                mapping_name_from_type(mappings_all, info.child).? ++
+                ")";
+        },
+        .Void, .Opaque => return "None",
```
matklad: Opaque feels wrong to me, shouldn't it also be compileError?

src/clients/python/python_bindings.zig
```diff
@@ -0,0 +1,497 @@
...
+            return name;
+        }
+    }
+    return null;
+}
+
+/// Resolves a Zig Type into a string representing the name of a corresponding Python ctype. This
+/// resolves both VSR and state machine specific mappings, as both are needed when interfacing via
+/// FFI.
+fn resolve_type_ctypes(comptime Type: type) []const u8 {
```
matklad: alternative naming for consideration: `zig_to_ctype` / `zig_to_python`

src/clients/python/python_bindings.zig
```diff
@@ -0,0 +1,497 @@
...
+            return switch (info.bits) {
+                8 => "int",
+                16 => "int",
+                32 => "int",
+                64 => "int",
+                128 => "int",
+                else => @compileError("invalid int type"),
+            };
+        },
+        .Void, .Opaque => return "None",
```
matklad: Same here, unsure about Opaque!

src/clients/python/python_bindings.zig
```diff
@@ -0,0 +1,497 @@
...
+        \\from .lib import c_uint128, dataclass, tbclient, validate_uint
+        \\
+        \\
+        \\
+    , .{});
+
+    // Emit ctype struct and enum type declarations.
+    inline for (mappings_all) |type_mapping| {
+        const ZigType = type_mapping[0];
+        const python_name = type_mapping[1];
```
matklad:
```suggestion
        const ZigType, const python_name = type_mapping[0];
```

src/clients/python/python_bindings.zig
```diff
@@ -0,0 +1,497 @@
...
+
+        // VSR ctype structs don't have a corresponding Python dataclass - so don't generate the
+        // `def to_python(self):` method for them.
+        const generate_ctypes_to_python = if (mapping_name_from_type(
+            mappings_state_machine,
+            ZigType,
+        ) != null) true else false;
```
matklad:
```
        const generate_ctypes_to_python = mapping_name_from_type(
            mappings_state_machine,
            ZigType,
        ) != null;
```

=P

src/clients/python/python_bindings.zig
matklad: When scaling up, the best way to write such kinds of things I've found is to move the entire template _up_ and let the logic follow template, like here:

https://github.com/tigerbeetle/blog/commit/279d1d83912a5d0393c0014b34b24236eb01a4bc#diff-f2948449439cc27583df5f215cf1af9256df2de90031f03f3d2e0dc63cfcd4eeR138

Not sure if its worth inventing some DSL for python though!

The good news, it doesn't _really_ matter if this code is readable or not, the important thing to review is the generated code!

src/clients/python/python_bindings.zig
```diff
@@ -0,0 +1,497 @@
...
+    buffer: *Buffer,
+    comptime Type: type,
+    comptime type_info: anytype,
+    comptime python_name: []const u8,
+    comptime skip_fields: []const []const u8,
+) !void {
+    if (@typeInfo(Type) == .Enum) {
+        buffer.print("class {s}(enum.IntEnum):\n", .{python_name});
+    } else {
+        // Packed structs.
```
matklad, line 162: assert it is one

src/clients/python/python_bindings.zig
matklad: This could be marginally better:

```
fn emit_method(buffer: *Buffer, comptime operation: std.builtin.Type.EnumField, options: struct {
    is_async: bool,
}) void {
    const op: StateMachine.Operation = @enumFromInt(operation.value);

    const event_type = comptime if (StateMachine.event_is_slice(op))
        "list[" ++ resolve_type_dataclass(StateMachine.EventType(op)) ++ "]"
    else
        resolve_type_dataclass(StateMachine.EventType(op));

    const result_type =
        comptime "list[" ++ resolve_type_dataclass(StateMachine.ResultType(op)) ++ "]";

    // For ergonomics, the client allows calling things like .query_accounts(filter) even
    // though the _submit function requires a list for everything. Wrap them here.
    const event_name_or_list = comptime if (!StateMachine.event_is_slice(op))
        "[" ++ StateMachine.event_name(op) ++ "]"
    else
        StateMachine.event_name(op);

    buffer.print(
        \\    {[prefix_fn]s}def {[fn_name]s}(self, {[event_name]s}: {[event_type]s}) -> {[result_type]s}:
        \\        return {[prefix_call]s}self._submit(
        \\            Operation.{[uppercase_name]s},
        \\            {[event_name_or_list]s},
        \\            {[event_type_c]s},
        \\            {[result_type_c]s},
        \\        )
    ,
        .{
            .prefix_fn = if (options.is_async) "async " else "",
            .fn_name = operation.name,
            .event_name = StateMachine.event_name(op),
            .event_type = event_type,
            .result_type = result_type,
            .event_name_or_list = event_name_or_list,
            .prefix_call = if (options.is_async) "await " else "",
            .uppercase_name = to_uppercase(operation.name),
            .event_type_c = ctype_type_name(StateMachine.EventType(op)),
            .result_type_c = ctype_type_name(StateMachine.ResultType(op)),
        },
    );
}
````

* extracting stuff into a function gives you couple of more levels of indentation to work with
* for arguments, they don't fit on the single line
* with `\\` literal, that means that the output would also be split
* but the length of the output should be _roughly_ the same as the length of the input
* and, indeed, with the existing impl some of the lines are longer than 100! So I think it's actually good to to argument per line for the output as well
* and tidy is smart, it checks that what's inside a `\\` is limited by 100, it doesn't count indentation in this case.

**Review by matklad — COMMENTED** [review]
Second batch!

File: src/clients/python/python_bindings.zig (hunk 35 above)
Comment (cb22) [reply src/clients/python/python_bindings.zig]: Too much time staring at Python style `foo if bar else foz` :smile:

File: src/scripts/release.zig (hunk 27 above)
Comment (cb22) [reply src/scripts/release.zig:810]: > though, I have vague memories of arch not having python3 alias

I vaguely remember that too, but it seems as if currently it has `python3` (and `python`) pointing to `python3.12` as the real executable.

File: .github/workflows/ci.yml (hunk 21 above)
Comment (cb22) [reply .github/workflows/ci.yml:163]: > outside view tells us that uv will not be the last attempt to fix Python packaging

Heh!

I went with `hatch` for the build, purely because it's what's recommended by default in the packaging tutorial. Our building is extremely simple, and these are only build time deps (the resulting package doesn't depend on any of them) so it's going to be easy to change in future if we decide we need to.

File: src/clients/python/ci.zig (hunk 28 above)
Comment (cb22) [reply src/clients/python/ci.zig]: Yeah, I've duplicated building the shared library into `build_python_client` rather than depending on the `c_client` step... But I'm not sure on it: the existing clients build their own libraries whereas the Python one does really use the C client libraries directly.

The C sample builds the C library directly, so I took inspiration from that.

File: src/clients/python/python_bindings.zig [hunk 37]
```diff
@@ -0,0 +1,506 @@
...
+    buffer.print("class {s}:\n", .{c_name});
+
+    inline for (type_info.fields) |field| {
+        if (comptime !std.mem.eql(u8, field.name, "reserved")) {
+            buffer.print("    {[name]s}: {[python_type]s} = {[python_type]s}" ++
+                "({[python_default]s}) # noqa: RUF009\n", .{
+                .name = field.name,
+                .python_type = zig_to_python(field.type),
+                .python_default = "0",
+            });
```
matklad: Maybe something like

```
            const python_type = zig_to_python(field.type);
            buffer.print("    {[name]s}: {[python_type]s} = ", .{
                .name = field.name,
                .python_type = python_type,
            });
            if (std.mem.eql(u8, python_type, "int")) {
                buffer.print("0\n", .{});
            } else {
                buffer.print("{s}.NONE\n", .{python_type});
            }
```

to get rid of the warning, and also to just simplify the code?

src/clients/python/src/tigerbeetle_python/bindings.py
```diff
@@ -0,0 +1,756 @@
...
+    ("reserved", ctypes.c_uint8 * 6),
+    ("timestamp_min", ctypes.c_uint64),
+    ("timestamp_max", ctypes.c_uint64),
+    ("limit", ctypes.c_uint32),
+    ("flags", ctypes.c_uint32),
+]
+
+
+@dataclass
+class Account:
```
matklad: Maybe move Python data structures to be _before_ the C ones? So that it's easier to see what makes sense to read, and what's just C boilerplate?

src/clients/python/src/tigerbeetle_python/__init__.py
```diff
@@ -0,0 +1,3 @@
+from .bindings import * # noqa
```
matklad: Let's maybe list this out explicitly? As a user of the library, I'll look at the `init` first, and, if I see `*`, I wouldn't actually know what we can import?

And it feels like we _shouldn't_ be re-exporting everything here perhaps? Things like mixing are private to the library.

src/clients/python/src/tigerbeetle_python/lib.py
```diff
@@ -0,0 +1,84 @@
...
+
+def validate_uint(*, bits: int, name: str, number: int):
+    if number > 2**bits - 1:
+        raise IntegerOverflowError(f"{name}=={number} is too large to fit in {bits} bits")
+    if number < 0:
+        raise IntegerOverflowError(f"{name}=={number} cannot be negative")
+
+
+class c_uint128(ctypes.Structure): # noqa: N801
+    _fields_ = [("low", ctypes.c_uint64), ("high", ctypes.c_uint64)] # noqa: RUF012
```
matklad: Let's set `_align_` explicitly here?

matklad: I also worry about mutability here. Let's call them `_low` and `_high` at least, to make sure fields are not directly manipulated?

src/clients/python/src/tigerbeetle_python/lib.py
```diff
@@ -0,0 +1,84 @@
...
+
+    @classmethod
+    def from_param(cls, obj):
+        if obj > (2**128 - 1):
+            raise IntegerOverflowError("Integer Overflow: " + str(obj))
+
+        return cls(high=obj >> 64, low=obj & 0xffffffffffffffff)
+
+    def to_python(self):
+        return self.high << 64 | self.low
```
matklad: 🤔 does `<<` work?

```
>>> ctypes.c_uint64(0) << 64
Traceback (most recent call last):
  File "<python-input-6>", line 1, in <module>
    ctypes.c_uint64(0) << 64
    ~~~~~~~~~~~~~~~~~~~^^~~~
TypeError: unsupported operand type(s) for <<: 'c_ulong' and 'int'
```

Also, shouldn't our `to_python` functions in bindings.py convert from ctypes to Python's ints?

src/clients/python/src/tigerbeetle_python/lib.py
```diff
@@ -0,0 +1,84 @@
...
+    if number < 0:
+        raise IntegerOverflowError(f"{name}=={number} cannot be negative")
+
+
+class c_uint128(ctypes.Structure): # noqa: N801
+    _fields_ = [("low", ctypes.c_uint64), ("high", ctypes.c_uint64)] # noqa: RUF012
+
+    @classmethod
+    def from_param(cls, obj):
+        if obj > (2**128 - 1):
```
matklad: could `validate_uint` here?

matklad (commented):
Third batch!

src/clients/python/src/tigerbeetle_python/lib.py
cb22: > 🤔 does << work?

It does - ctypes structs play pretty fast and loose with integer types... You define the type as a `c_uint64`, but that's only for information when passed across the FFI boundary. Otherwise, it's always a Python int - both when being called from Python -> C and when being called from C -> Python) _unless_ you explicitly instantiate the type like you did there.

That's why the bindings don't convert from ctypes to Python ints too.

src/clients/python/samples/basic/main.py
```diff
@@ -0,0 +1,49 @@
+import os
+
+import tigerbeetle_python as tb
+
+client = tb.ClientSync(cluster_id=0, replica_addresses=os.getenv("TB_ADDRESS", "3000"))
```
matklad: I'd expect to see

```
with tb.ClientSync(cluster_id=0, replica_addresses=os.getenv("TB_ADDRESS", "3000")) as client:
   ...
```

src/clients/python/src/tigerbeetle_python/client.py
```diff
@@ -0,0 +1,229 @@
...
+    def _acquire_packet(self, operation: bindings.Operation, operations: Any,
+                        c_event_type: Any, c_result_type: Any) -> InflightPacket:
+        packet = bindings.CPacket()
+        packet.next = None
+        packet.user_data = Client._counter.increment()
+        packet.operation = operation
+        packet.status = bindings.PacketStatus.OK
+
+        operations_array_type = c_event_type * len(operations)
+        operations_array = operations_array_type(*map(c_event_type.from_param, operations))
```
matklad: Hm, I am not sure about the memory management story. Here, operations_array owns the memory we passed to C. When we call `ctypes.cast`, would that incerement the ref-count, such that, when `operations_array` goes out of scope, the underlying memory is still tracked by `packed` through `packet.data` field?

matklad: I guess, if it doesn't leak and doesn't crash that's fine, but still want to double-check.

src/clients/python/samples/basic/main.py
cb22: We could make the clients work as context managers too, but I don't think this is the behavior we want to encourage! We'd rather have the users create a single client somewhere and use that everywhere throughout their application.

Otherwise, they're very likely going to be creating / destroying clients per request. Destroying is not a problem, but the create would lead to a register and extra latency (which pending my message bus tweaks is _way_ higher than it should be!)

src/clients/python/src/tigerbeetle_python/__init__.py
cb22: Hmm - I do agree explicit imports is better; but what I was going for here was complete state machine independence - the hand written code knows nothing of the autogenerated bindings which define the state machine interface too.

I could autogenerate `__init__.py` too to keep that property, but then it's two autogenerated files.

src/clients/python/src/tigerbeetle_python/client.py
cb22: > When we call ctypes.cast, would that incerement the ref-count, such that, when operations_array goes out of scope, the underlying memory is still tracked by packed through packet.data field?

It does!

```python
import sys
from ctypes import *

operations_array_type = c_uint32 * 32
x = operations_array_type(*range(0, 32))

sys.getrefcount(x) # prints 2

foo = cast(x, c_void_p)
sys.getrefcount(x) # prints 3

```

src/clients/python/src/tigerbeetle_python/lib.py
cb22: > Let's set _align_ explicitly here?

`_align_` was only added in 3.13 - I want to add it, but I don't want it to give a false sense of confidence. I think I might set it and explicitly check alignment.

src/clients/python/src/tigerbeetle_python/lib.py
cb22: I've removed the validation here, and made it `validate_uint` in the `from_param` so it behaves like all the other ctypes integers.

src/clients/python/samples/basic/main.py
matklad: 🤔 but it feels like, if we have `deinit` method, we should encourage that it gets called, no? The fact that our own sample doesn't close the client is concerning (so, if you copy-paste this into one thousand tests, you'll leak one thousands C things, which is not good). 

I am not sure that `with` necessary encourages short scopes? I'd assume the user would have a `with tb.Client`  in main, and then pass the client everywhere as a parameter. 

Client is a resource. Idiomatic Python construct for dealing with resources, both short-lived and long-lived, is `with`. While it _could_ encourage some misuse, I think it's better to stick to language idioms, rather than preventively outthink the user?

We do use equivalent idioms (using in C#, try-with-resources in Java) for other languages.

src/clients/python/src/tigerbeetle_python/__init__.py
matklad: Let's keep it as is for now then!

src/clients/python/tests/test_basic.py
matklad: argh, this is unfortunate a bit, as node tests are not written in the correct style. Specifically, hard-coding account ids like we do here is not great, because this makes tests order-dependent. 

Both sets of test should have used `tbid` function to generate unique ids. But I guess we will then refactor both sets of tests later?

matklad: Or, I guess, we should be _generating_ the tests?

matklad (commented):
Finished review!

Looks good to me, modulo three things:

* rename `Client.deinit` to `Client.close`
* ensure that the sample code calls close
* consider providing context-manager interface

src/clients/python/tests/test_basic.py
cb22: > Or, I guess, we should be generating the tests?

Yes, definitely! Especially since we already have a suite of test vectors in a DSL lying around... I think that, coupled with some hand written language specific tests, coupled with Vortex for checking semantics (what you think is 0x123456789 is actually 0x123456789 to all the other clients) would be ideal.

cb22:
Addressed all 3 - ended up going with the `with ...` style in the samples in the end, and renamed the package to `tigerbeetle`. I'll rebase shortly! :smile:

cb22:
I've also set up a Pypi account and added the TWINE_* variables.

hookla:
Could u incorporate a kind of health check / stay alive so client logs when it loose connection to tb

matklad (approved):

PR #2492, opened 2024-11-20
docs: add note about atomic lookups being limited to batch size

cb22:
This mainly applies to accounts, because they appear as mutable...

### DIFF
--- a/docs/reference/requests/lookup_accounts.md
+++ b/docs/reference/requests/lookup_accounts.md
@@ -10,6 +10,11 @@ transfer. Instead, set the
 accounts to limit their account balances. More complex conditional transfers can be expressed using
 [balance-conditional transfers](../../coding/recipes/balance-conditional-transfers.md).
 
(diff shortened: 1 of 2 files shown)


docs/reference/requests/lookup_accounts.md
```diff
@@ -10,6 +10,10 @@ transfer. Instead, set the
 accounts to limit their account balances. More complex conditional transfers can be expressed using
 [balance-conditional transfers](../../coding/recipes/balance-conditional-transfers.md).
 
+⚠️ It is not possible currently to look up more than a full batch (8190) of accounts atomically.
+When issuing multiple `lookup_accounts` calls, it can happen that other operations will interleave
+between the calls leading to read skew.
```
sentientwaffle: Maybe reference the `history` flag? Seems like in a use case where you need >1 batch of atomic accounts that might be useful.

matklad (dismissed):

matklad (approved):

PR #2493, opened 2024-11-20
clock: change log severity levels

cb22:
Upgrade `no agreement on cluster time` to an error: it means requests can't be processed _right now_ and it's as good of a reason as any to page someone.

At the same time, change the synchronized message to an info: you'll want to be able to see that things have returned to normal.

### DIFF
--- a/src/vsr/clock.zig
+++ b/src/vsr/clock.zig
@@ -346,8 +346,8 @@ pub fn ClockType(comptime Time: type) type {
             return self.time.realtime();
         }
 
-        /// Called by `StateMachine.prepare_timestamp()` when the primary wants to timestamp a
-        /// batch. If the primary's clock is not synchronized with the cluster, it must wait until
+        /// Called by `Replica.on_request()` when the primary wants to timestamp a batch. If the
+        /// primary's clock is not synchronized with the cluster, it must wait until it is.
         /// it is.
         /// Returns the system time clamped to be within our synchronized lower and upper bounds.
         /// This is complementary to NTP and allows clusters with very accurate time to make use of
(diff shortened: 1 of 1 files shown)


src/vsr/clock.zig
```diff
@@ -480,7 +480,7 @@ pub fn ClockType(comptime Time: type) type {
         fn after_synchronization(self: *Clock) void {
             const new_interval = self.epoch.synchronized.?;
 
-            log.debug("{}: synchronized: truechimers={}/{} clock_offset={}..{} accuracy={}", .{
+            log.info("{}: synchronized: truechimers={}/{} clock_offset={}..{} accuracy={}", .{
```
cb22, line 483: Perhaps it's too much internal information for a log.info output?

matklad:
> error: it means requests can't be processed right now 

I don’t think we log.err in such cases though? Like, if a view change happens, I don’t think we emit an err, although the cluster is unable to process requests.

we perhaps _should_ be err logging such cases, but I am not sure: some amount of retries feels normal for the domain, and we should page only if we are unable to process request for some time. 

So, emitting a warn and paging on repeating warns feels appropriate?

but I totally trust your judgement here, I have little expertise with such things!

matklad (approved):
Approving! To me it still feels like warn is more appropriate here, but ill defer to you :)

matklad:
Though, I guess if we do err here, we maybe want to add some errs on viewchsnges and such for replica?

cb22:
> Though, I guess if we do err here, we maybe want to add some errs on viewchsnges and such for replica?

This is a good question! I guess the difference for me is that this specific clock message:
* only happens after warnings already (eg, `synchronization failed, partitioned`),
* indicates a level trigger vs an edge trigger (our clock and therefore cluster is now entering the state of down, until we say otherwise) and,
* will very likely require external or operator intervention to fix.

PR #2495, opened 2024-11-21
[Java] Remove `Batch.EMPTY` Shared Objects

batiati:
Since the introduction of `batch.getHeader()` by #2481, we must create a new instance per request instead of sharing the same `Batch.EMPTY` instance.
However, to reduce allocations, the underlying buffer is still a shared empty object.

This PR fixes a problem where concurrent requests with `reply.len == 0` were overriding each other's headers.

### DIFF
--- a/src/clients/java/src/test/java/com/tigerbeetle/IntegrationTest.java
+++ b/src/clients/java/src/test/java/com/tigerbeetle/IntegrationTest.java
@@ -24,6 +24,7 @@ import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertNotEquals;
 import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertNotSame;
 import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertTrue;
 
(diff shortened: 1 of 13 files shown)


sentientwaffle (dismissed):
LGTM -- thanks for tracking this down!

The original bug can be reproduced with the integration test:

```java
    @Test
    public void testRace() throws Throwable {
        try (final var client2 = new Client(clusterId, new String[] {server.getAddress()})) {
            for (int i = 0; i < 1000; i++) {
                var r1 = client.lookupAccountsAsync(new IdBatch(UInt128.id()));
                var r2 = client2.lookupAccountsAsync(new IdBatch(UInt128.id()));

                var t1 = r1.get().getHeader().getTimestamp();
                var t2 = r2.get().getHeader().getTimestamp();
                if (t1 == t2) {
                    throw new Exception("BUG: " + String.format("timestamp: %d = %d", t1, t2));
                }
            }
        }
    }
```

...Though that is niche enough that it probably isn't worth adding to the integration test suite.

src/clients/java/src/test/java/com/tigerbeetle/IntegrationTest.java
```diff
@@ -2186,6 +2187,33 @@ public void testImportedFlagAsync() throws Throwable {
         assertEquals(timestamp + accounts.getLength() + 1, lookupTransfers.getTimestamp());
     }
 
+    /**
+     * Asserts that empty replies are not shared.
+     * https://github.com/tigerbeetle/tigerbeetle/pull/2495
+     */
+    @Test
+    public void testEmptyReply() throws Throwable {
+        final int TASKS_COUNT = 1_000;
```
sentientwaffle: If we are going to add this to the integration test suite, let's change the task count to 4 or so... it didn't take that many repetitions to hit the bug, and I don't want to make the test suite slower :smile:

batiati: Cool, it's about 300ms now!

sentientwaffle (approved):

PR #2499, opened 2024-11-22
clock: don't let large RTT mask unsynchornized clocks

matklad:
Although we do binary search over the `tolerance` variable, the _actual_
error margin we use for Marzullo is `tolerace + RTT`, as seen in
`window_tuples`:

```
    .offset = sample.clock_offset -
        @as(i64, @intcast(sample.one_way_delay + tolerance)),
```

That is, if RTT is large, it would _easier_ for Marzullo to find a
quorum of clocks.

So, if pings take 30 seconds to round-trip, we might end up with the
interval 50 (= (10 + 30/2) * 2) seconds wide, larger than our max offset
tolerance of 10 seconds.

Fix this by capping the synchronized interval to the offset tolerance.


src/vsr/clock.zig
```diff
@@ -460,6 +460,11 @@ pub fn ClockType(comptime Time: type) type {
                 const interval = Marzullo.smallest_interval(self.window_tuples(tolerance));
                 if (interval.sources_true < self.quorum) break;
 
+                const accuracy = interval.upper_bound - interval.lower_bound;
+                if (accuracy > 2 * clock_offset_tolerance_max) {
+                    continue; // Interval too wide due to large RTT of samples.
+                }
+
```
jorangreef, line 467: Could this mean that we never synchronize the window? And thus deadlock the cluster?

src/vsr/clock.zig
```diff
@@ -473,9 +478,11 @@ pub fn ClockType(comptime Time: type) type {
             // for the operator, as the counterpoint to `no agreement on cluster time`.
             if (self.epoch.synchronized == null and self.window.synchronized != null) {
                 const new_interval = self.window.synchronized.?;
+                const accuracy = new_interval.upper_bound - new_interval.lower_bound;
```
jorangreef, line 481: Perhaps we could avoid introducing the non-big endian `new_interval` into the scope entirely, using `window.synchronized` inline instead.

jorangreef (commented):
A question and style nit from me, but otherwise I'll leave the correctness review here to @sentientwaffle to double check everything.

src/vsr/clock.zig
jorangreef: i.e. If samples have a large RTT, will we still make progress?

The intention is that this is more "a clock fault verifier"—if things look badly out, as if PTP is clearly definitely not working, then don't process transactions and alert the operator, but otherwise we can err on the side of more tolerance rather than less.

matklad: >Could this mean that we never synchronize the window? And thus deadlock the cluster?


Yes, that's the intention. I am personally torn a bit on what the behavior _should_ be. 

I assume that we _do_ want to halt the cluster if the clocks are clearly not synchronized. But then, it would be strange to _not_ halt if the clocks are as badly desynced, _and_ we have a horrible inter-node network latency on top of that. 

The behavior before this PR feels self-contradictory to me, where we allow two errors to cancel-each other. 

But now I am wondering if this is the right fix? Perhaps we should change `window_tuples` to _not_ add RTT to clock's error margin? This is the counter-intuitive thing --- if you increase individual clock's error margin, you make it easier for marzullo to pick some interval. 

Given that we binary search for the error margin anyway, it doesn't feel wrong to not add it.

matklad: >The behavior before this PR feels self-contradictory to me

Two explain this better, the two non-contradictory behaviors would be:

* to _always_ halt transaction processing if the clocks can't be proven to be in sync
* to _never_ halt transaction processing, and use clock sync as a defense in depth mechanism (so, we cap primary's time if we can, but use it as is if we can't check our's clocks against the peers)

matklad: That's the logic before PR, without RTT=0 (or very low) the clocks are considered unsynced, but with RTT=10s they counter-intuitively become synced!

image

jorangreef: A picture worth ten thousand milliseconds!

jorangreef: > always halt transaction processing if the clocks can't be proven to be in sync

IIRC we don't do this at present? That is, we're not trying to "prove that clocks are in sync" but rather we're trying to halt if the clocks are clearly way out of sync.

In other words, we "over-tolerate" at present, and err on the side of allowing the error margin also to include any uncertainty around RTT vs bufferbloat.

> Perhaps we should change window_tuples to not add RTT to clock's error margin?

Yes, I think this would be good to do.

jorangreef: (for anyone else following along, some of the background on the intention here: https://www.youtube.com/watch?v=C3oWbZ4tQPM)

matklad: >IIRC we don't do this at present?

No, we do halt transaction processing if we can't sync clocks. That is, the _first_ picture would be a stuck cluster both before and after this PR. 

https://github.com/tigerbeetle/tigerbeetle/pull/2493 logs this situation.

In terms of code:

* window.synchronized is null here: 
  https://github.com/tigerbeetle/tigerbeetle/blob/584610e006030812e1271163ce3d37652cdef645/src/vsr/clock.zig#L470
* so clock.realtime_synchronized propagates null:
  https://github.com/tigerbeetle/tigerbeetle/blob/584610e006030812e1271163ce3d37652cdef645/src/vsr/clock.zig#L367
* and we drop request here:
  https://github.com/tigerbeetle/tigerbeetle/blob/584610e006030812e1271163ce3d37652cdef645/src/vsr/replica.zig#L1541

jorangreef: To clarify, it's not about "if we can't sync clocks" (we leave it to PTP or NTP or something else to do that, while also clamping primary within reasonable bounds as defense in depth), but rather we're trying to observe them to see if they're beyond 10 seconds  + RTT tolerance.

If that's the case, if they're way out, then we halt processing.

jorangreef: At least, that's always been the intention here when I wrote it. We're only trying to verify. We trust the operator's clock sync protocol to sync. And then we just want to verify that the clocks are clearly not way out.

“Doveryai, no proveryai”

jorangreef: Agreed that we do halt processing, but it's only when we see that most clocks all don't agree... and they're way outside tolerance (even adding RTT).

jorangreef: The "adding RTT" part is us being over-tolerant, not necessarily a bug. But it will lead to situations where we don't detect faulty clocks as much as we could.

sentientwaffle: Maybe there is a missing dimension here... If RTT varies wildly from message to message, then we could increase the tolerance (or maybe stop accepting requests). But if the RTT is high (but _consistent_) then we could keep the tolerance low.

What if we tracked multiple samples per replica, and used the RTT variance when computing the clock tolerance?

matklad:
Thinking about this more, let's keep the current logic.

What I originally thought is that it is _incorrect_ that larger RTT might make the clocks appear to be synchronized, while they won't be in small RTT. But, while counter-intuitive, I _think_ it makes sense:

* with small RTT, you know more or less exactly where the clocks are, and you can clearly say "the clocks are unsynced"
* with large RTT, you are less certain about remote clocks: they might, or might not be synced. 

And than it becomes the question of which side of error do we pick: if the clocks look like they might, or might not be synced, do we:

* optimistically assume that they are synced, and remain available
* or pessimistically assume that they are not synced, and halt processing


And we do want to lean toward being available here --- even if the clocks are not synced, they still can't be _wildly_ different (or rtt probably is less than 10 minutes), and it's much better to be off by 10 minutes in the timestamp than to completely halt processing.

PR #2501, opened 2024-11-22
[Java] Call `DetachCurrentThread` when the Client is closed.

batiati:
The thread exit is detected by thread-local storage (TLS) destructors using `pthread_key_create` on POSIX systems and `FlsAlloc` on Windows.

### DIFF
--- a/src/clients/java/src/client.zig
+++ b/src/clients/java/src/client.zig
@@ -879,8 +879,9 @@ test "JNIThreadHelper:tls" {
         const TestContext = @This();
 
         var tls_key: ?tls.Key = null;
+        var event: std.Thread.ResetEvent = .{};
 
-        counter: u32,
+        counter: std.atomic.Value(u32),
 
         fn init() TestContext {
             if (tls_key == null) {
@@ -888,17 +889,18 @@ test "JNIThreadHelper:tls" {
             }
 
             return .{
-                .counter = 0,
+                .counter = std.atomic.Value(u32).init(0),
             };
         }
 
         fn thread_main(self: *TestContext) void {
             tls.set_key(tls_key.?, self);
+            event.wait();
         }
 
         fn destructor_callback(value: *anyopaque) callconv(.C) void {
             const self: *TestContext = @alignCast(@ptrCast(value));
-            self.counter += 1;
+            _ = self.counter.fetchAdd(1, .monotonic);
         }
     };
 
@@ -907,9 +909,16 @@ test "JNIThreadHelper:tls" {
     for (&threads) |*thread| {
         thread.* = try std.Thread.spawn(.{}, TestContext.thread_main, .{&context});
     }
+
+    // Assert that the callback only fires when the thread finishes.
+    try std.testing.expect(context.counter.load(.monotonic) == 0);
+
+    // Signal all threads to complete and wait for them.
+    TestContext.event.set();
     for (&threads) |*thread| {
         thread.join();
     }
 
-    try std.testing.expect(context.counter == threads.len);
+    // Assert that all callbacks have fired.
+    try std.testing.expect(context.counter.load(.monotonic) == threads.len);
 }
--- a/src/clients/java/src/client.zig
+++ b/src/clients/java/src/client.zig
@@ -773,9 +773,9 @@ const JNIThreadHelper = struct {
     }
 
     // Will be called by the OS with the JVM handler when the thread finalizes.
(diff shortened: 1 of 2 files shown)


src/clients/java/src/client.zig
```diff
@@ -586,6 +608,18 @@ const JNIHelper = struct {
         return env;
     }
 
+    pub inline fn dettach_current_thread(jvm: *jni.JavaVM) void {
```
sentientwaffle:
```suggestion
    pub inline fn detach_current_thread(jvm: *jni.JavaVM) void {
```

src/clients/java/src/client.zig
```diff
@@ -571,6 +576,23 @@ const JNIHelper = struct {
...
+        return switch (jni_result) {
+            .ok => env,
+            .thread_detached => null,
+            else => {
+                const message = "Unexpected result calling JavaVM.GetEnv";
+                log.err(
+                    message ++ "; Error = {} ({s})",
+                    .{ @intFromEnum(jni_result), @tagName(jni_result) },
+                );
+                @panic("JNI: " ++ message);
```
sentientwaffle, line 591: Maybe combine into  `std.debug.panic` -- at least right now, clients don't surface any log messages afaik. (Likewise in `create_key`/`set_key`).

src/clients/java/src/client.zig
```diff
@@ -708,3 +742,165 @@ const JNIHelper = struct {
...
+    pub fn attach_current_thread_with_cleanup(jvm: *jni.JavaVM) *jni.JNIEnv {
+        // Create the thread-local storage key and the corresponding destructor
+        // function pointer **once** per JVM (may be concurrent).
+        if (@atomicLoad(tls.Key, &tls_key, .acquire) == 0) {
+            mutex.lock();
+            defer mutex.unlock();
+
+            if (tls_key == 0) {
+                // The destructor function will be called before the thread exits.
+                const key = tls.create_key(&destructor_callback);
```
sentientwaffle: `assert(key != 0);`

batiati: Thanks for pointing it out! I just discovered that `zero` is indeed a valid tls key, so we cannot use it as the unitialized flag!

I did a nice refactor, now using `std.once`, and added tests for the initialization case.

src/clients/java/src/client.zig
batiati: Yes, moving to `std.debug.panic` is a great idea!

I'll let it "as is" for now just because it's standardized with the rest of the JNI client, and change all in a follow-up PR when the client logging is merged.

src/clients/java/src/client.zig
```diff
@@ -708,3 +742,174 @@ const JNIHelper = struct {
         return @ptrCast(address[0..@as(usize, @intCast(length))]);
     }
 };
+
+/// Helper for managing the `AttachCurrentThread`/`DetachCurrentThread` lifecycle
+/// when the JNI layer is unaware of when the native thread exits.
+/// https://developer.android.com/training/articles/perf-jni#threads
+const JNIThreadHelper = struct {
+    var tls_key: tls.Key = undefined;
```
sentientwaffle: Maybe an optional `?tls.Key` instead?

src/clients/java/src/client.zig
```diff
@@ -708,3 +742,174 @@ const JNIHelper = struct {
...
+    /// Note: We don't need to delete the key because the JNI module cannot be unloaded,
+    /// so it will always be available for the duration of the JVM process.
+    fn create_key() void {
+        tls_key = tls.create_key(&destructor_callback);
+    }
+
+    // Will be called by the OS with the JVM handler when the thread finalizes.
+    fn destructor_callback(value: *anyopaque) callconv(.C) void {
+        JNIHelper.detach_current_thread(@ptrCast(value));
+    }
```
sentientwaffle: Can we give this a more descriptive name (`jvm`)?

sentientwaffle (dismissed):
A couple of optional nits but otherwise LGTM.

src/clients/java/src/client.zig
batiati: It can't be `jvm` because the unit test uses another pointer type.
But I think we can call it `tls_value` since it's indexed by the `tls_key`.

sentientwaffle: The test uses its own `destructor_callback` implementation, so we could name it more descriptively here, right?

sentientwaffle (approved):

allysa08 (approved):

PR #2502, opened 2024-11-22
Vortex: Java driver

owickstrom:
This adds two things:

1. The ability to pass in a `--driver-command` to Vortex, which is used to spawn the driver process that the workload communicates with over stdio
2. A Java driver, which is the second driver to be added (the `tb_client`-based Zig driver is the default)
 
I've intentionally not added any scheduling or automation using this driver yet. Part of why I wanted to get this done in its current form is to prove the driver API works. We have other client languages to write drivers for, and this support is useful for both existing and new client languages.

If the Java driver shouldn't be merged without any automated usage, I can split the PR up and only merge the parameterized driver option. Let me know.

### DIFF
--- a/src/testing/vortex/java_driver/src/main/java/Main.java
+++ b/src/testing/vortex/java_driver/src/main/java/Main.java
@@ -1,0 +1,378 @@
+import java.io.IOException;
+import java.nio.ByteBuffer;
+import java.nio.ByteOrder;
+import java.nio.channels.Channels;
+import java.nio.channels.ReadableByteChannel;
+import java.nio.channels.WritableByteChannel;
+import java.util.HashMap;
+import java.util.Map;
+import com.tigerbeetle.AccountBatch;
+import com.tigerbeetle.Client;
+import com.tigerbeetle.IdBatch;
+import com.tigerbeetle.TransferBatch;
+import com.tigerbeetle.UInt128;
+
+/**
+ * A Vortex driver using the Java language client for TigerBeetle.
+ */
+public final class Main {
+  public static void main(String[] args) throws Exception {
+    assert args.length == 2 : "java driver requires two positional command-line arguments";
+
+    byte[] clusterID = UInt128.asBytes(Long.parseLong(args[0]));
+
+    var replicaAddressesArg = args[1];
+    String[] replicaAddresses = replicaAddressesArg.split(",");
+    if (replicaAddresses.length == 0) {
+      throw new IllegalArgumentException(
+          "REPLICAS must list at least one address (comma-separated)");
+    }
+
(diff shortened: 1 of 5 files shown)


src/testing/vortex/java_driver/src/main/java/Main.java
```diff
@@ -0,0 +1,378 @@
...
+import com.tigerbeetle.IdBatch;
+import com.tigerbeetle.TransferBatch;
+import com.tigerbeetle.UInt128;
+
+/**
+ * A Vortex driver using the Java language client for TigerBeetle.
+ */
+public final class Main {
+  public static void main(String[] args) throws Exception {
+    assert args.length == 2 : "java driver requires two positional command-line arguments";
```
matklad: I _think_ JVM disables asserts by default. So let's if and throw here? 

Or maybe not, if we care to pass the enable assertions flag when running this.

src/testing/vortex/java_driver/src/main/java/Main.java
```diff
@@ -0,0 +1,378 @@
...
+
+    byte[] clusterID = UInt128.asBytes(Long.parseLong(args[0]));
+
+    var replicaAddressesArg = args[1];
+    String[] replicaAddresses = replicaAddressesArg.split(",");
+    if (replicaAddresses.length == 0) {
+      throw new IllegalArgumentException(
```
matklad, line 30: but we definitely should use consistent error handling methods for then number of arguments and each specific argument!

src/testing/vortex/java_driver/src/main/java/Main.java
```diff
@@ -0,0 +1,378 @@
...
+      var writer = new Driver.Writer(Channels.newChannel(System.out));
+      var driver = new Driver(client, reader, writer);
+      while (true) {
+        driver.next();
+      }
+    }
+  }
+}
+
+record Driver(Client client, Reader reader, Writer writer) {
```
matklad, line 45: Java so new that GitHub even fails to highlight it! Who needs flying cars when we have pattern matching in Java!

matklad (commented):
LGTM with assert replaced with an illegal argument exception!

Really love how this ends up looking!

src/testing/vortex/java_driver/src/main/java/Main.java
owickstrom: Society if we used Java records more!

src/testing/vortex/java_driver/src/main/java/Main.java
owickstrom: Ah, yeah I'll rewrite to if/throw instead.

matklad (approved):

PR #2503, opened 2024-11-25
devhub: Tweaks

fabioarnold:
* "DevHub" part of logo should have the color `#CF0`
* Added tooltips for the "ok", "fail" and "canary" labels, and the issue triage link
* Display the number of untriaged issues
  * I use the `https://api.github.com/repos/tigerbeetle/tigerbeetle/issues` endpoint to fetch all open issues, which also includes open pull requests (GitHub quirk). This makes the previous pull request fetch redundant. However, I couldn't just replace it because the `url` is slightly different (`issues` vs `pulls`). Let me know if we should prioritize fewer fetches or simpler code.


src/devhub/index.html
```diff
@@ -39,8 +43,9 @@ <h3>Release manager</h3>
         <h3>Links</h3>
         <div>
           <p><a href="https://github.com/tigerbeetle/tigerbeetle/pulls/assigned/@me">My code review</a></p>
-          <p><a href="https://github.com/tigerbeetle/tigerbeetle/issues?q=is%3Aissue+is%3Aopen+-label%3Atriaged">Issue
-              triage</a></p>
+          <p><a href="https://github.com/tigerbeetle/tigerbeetle/issues?q=is%3Aissue+is%3Aopen+-label%3Atriaged"
+              title="We want to keep the number of untriaged issues to zero. Triaged means someone looked at the issue and responded.">Issue
```
jorangreef:
```suggestion
              title="Let's keep the number of untriaged issues at zero. Triaged means that we looked at the issue, responded and labeled the issue as triaged.">Issue
```

src/devhub/index.html
```diff
@@ -49,15 +54,15 @@ <h3>Links</h3>
         <h3>Main branch fuzzes</h3>
         <table>
           <tr>
-            <td class="ok" style="padding-right: 16px;">ok</td>
+            <td class="ok" style="padding-right: 16px;" title="This many fuzzes didn't have any failure.">ok</td>
```
jorangreef: Can we add some concept of time and also which branch(es) this relates to? Or remove confusion around these?

src/devhub/index.html
```diff
@@ -49,15 +54,15 @@ <h3>Links</h3>
...
             <td><span id="fuzz-ok">N/A</span></td>
           </tr>
           <tr>
-            <td class="fail" style="padding-right: 16px;">fail</td>
+            <td class="fail" style="padding-right: 16px;" title="This many fuzzes had a failure.">fail</td>
```
jorangreef: Perhaps, 

"The number of failed fuzz seeds on any branch in the last N days.”

src/devhub/index.html
```diff
@@ -49,15 +54,15 @@ <h3>Links</h3>
...
             <td><span id="fuzz-fail">N/A</span></td>
           </tr>
           <tr>
-            <td class="canary" style="padding-right: 16px;">canary</td>
+            <td class="canary" style="padding-right: 16px;" title="The canary fuzzer found this many failures.">canary</td>
```
jorangreef: Can we explain the concept of "canary" here? i.e. What should the viewer expect to be "good" here? And how should they interpret this if it's bad?

src/devhub/index.html
jorangreef: (Or I might be misunderstanding what this is!)

src/devhub/index.html
matklad: Branch is in the `h3`: `<h3>Main branch fuzzes</h3>`. 

For time, this is under-specified, but it's over our entire fuzzing database (our database is constant-size, so we trim old entries)

matklad: Thinking about this, let's maybe remove this entire `Main branch fuzzers` table? 

The _purpose_ of this table is to double-check the code used to render specific failing seeds: at one point, we had a bug where the failing seeds were _recorded_ in the devhub database, but were wrongly excluded from the presentation. 

But I personally don't actually look at this table all that much, so I don't think I'll spot the problem if the same bug gets re-introduced.

So perhaps let's try to move the logic as an assertion into the code? Just asserting that, if `mainBranchFail > 0`, then we have a row in the table marked as a failure?

matklad:
>Let me know if we should prioritize fewer fetches or simpler code.

We probably want fewer fetches to not run into API limits, but it doesn't matter _that_ much

matklad (dismissed):
I'd personally would be happy to have this as is, as the number next to untriaged issues is a big deal! But, yeah, I'd also replaced the table with an in-code assert, and tried to do less hits to the GitHub API

matklad (approved):

PR #2507, opened 2024-11-25
[StateMachine] Rework AccountBalances.

batiati:
- Rename `AccountBalances` to `AccountEvents`, inserting an object for every event that updates the balance (transfers and expires), regardless of `Account.flags.history`.

- Add indexes for the account's latest balance update, enabling direct "as-of" balance lookups by timestamp. This index is stored only for accounts with `Account.flags.history`.

- Add indexes for expiry events (dr/cr accounts, pending transfers, and ledgers), providing the same query capabilities for expires as for regular transfers.

- Add indexes for tracking `{post,void}_transfer`, since `Transfers.flags` isn´t indexed.

- Add index for "prunable" events (those not related to accounts marked with the `history` flag), that could be removed once CDC is done.

There are no plans to retrofit the indexes.


src/stdx.zig
```diff
@@ -785,22 +785,31 @@ fn format_int_size_bin_exact(
...
-    var val = value;
-    while (val % 1024 == 0) : (magnitude += 1) {
-        val = @divExact(val, 1024);
+    var value_unit = value;
+    while (value_unit % 1024 == 0) : (magnitude += 1) {
+        value_unit = @divExact(value_unit, 1024);
     }
 
-    const mags_iec = " KMGTPEZY";
+    const mags_iec = "BKMGTPEZY";
```
sentientwaffle:
```suggestion
    const magnitudes_iec = "BKMGTPEZY";
```

(It took me a moment to decode the old name!)

src/state_machine.zig
```diff
@@ -759,7 +889,7 @@ pub fn StateMachineType(
             assert(input.len <= self.batch_size_limit);
 
             self.prepare_timestamp += switch (operation) {
-                .pulse => 0,
+                .pulse => constants.batch_max.create_transfers, // Max transfers to expire.
```
sentientwaffle, line 909: ~Was this a bug before?~ Ah I see -- previously weren't creating any new objects, we were only updating objects in-place. But now we are inserting "account events" with their own timestamps, which means the prepare/commit timestamps are now advanced by transfer expiry!

src/state_machine.zig
```diff
@@ -2881,36 +3074,46 @@ pub fn StateMachineType(
...
+                    .event_timestamp = event_timestamp,
+                    .dr_account = &dr_account_new,
+                    .cr_account = &cr_account_new,
+                    .transfer_flags = null,
+                    .transfer_pending_status = .expired,
+                    .transfer_pending = p,
+                    .amount_requested = 0,
+                    .amount = p.amount,
+                });
+                self.commit_timestamp = event_timestamp;
```
sentientwaffle: Since this is safety-critical, we should assert that the value is being increased.

src/state_machine.zig
```diff
@@ -305,15 +320,102 @@ pub fn StateMachineType(
...
+                        fn prunable(object: *const AccountEvent) ?void {
+                            if (object.dr_account_flags.history or
+                                object.cr_account_flags.history)
+                            {
+                                return null;
+                            } else {
+                                return {};
+                            }
+                        }
+                    }.prunable,
```
sentientwaffle, line 417: :thinking: I was considering whether it makes sense to invert this index, since history=false is the case where we care more about performance anyway. But it is nice that this index exactly corresponds to the entries that can be deleted, so it is fine as-is.

src/state_machine.zig
```diff
@@ -288,12 +303,12 @@ pub fn StateMachineType(
             }
         };
 
-        const AccountBalancesGroove = GrooveType(
+        const AccountEventsGroove = GrooveType(
```
sentientwaffle, line 311: Since these objects are currently never queried, what can we do to verify that we are indeed inserting the data correctly?

src/state_machine.zig
batiati: I tried to balance the load, when `history == false`, we add this `prunnable` index, otherwise we add `account_timestamp`.

src/state_machine.zig
batiati: > Since these objects are currently never queried

We do query `AccountEvents` when calling `get_account_balances`. However, none of the new indexes or fields are used. 

> what can we do to verify that we are indeed inserting the data correctly?

We could assert the data on the write path while it's still in memory (although it's not usually what we do with indexes. There are other cases of never queried indexes for future use). Any thoughts?

src/state_machine.zig
sentientwaffle: Thinking about this some more:
- I expect most accounts to have `history == false`,
- which means that this index will be very "dense" -- i.e. it would not be uncommon to have long runs of timestamps that are all in this index,
- whereas if we stored the inverse condition, it would be much more sparse.

TB doesn't support compression right now -- but when it does, the density lends itself well to an inexpensive index (e.g. we could store the first integer in a "run" + "length of run"). So I am okay paying the more expensive cost until then.

src/state_machine.zig
sentientwaffle: Hmm, we are handicapped by the fact that the state machine workload doesn't keep a list of all the transfers it has created and instead tries to be clever with workload generation to avoid doing so. I mean to fix that but I doubt I'll get to it soon. :sweat: 

If we did have that, then we could add a temporary checker to the VOPR which randomly fires (perhaps triggered by the replica's `committed` hook)? It would just reach into the replica's forest and run queries and verify that the events match the workload's expectations.

:thinking: Weaker than a fuzz test, but maybe we could check with a state machine unit test? Add a new "action" to our state machine unit test DSL which, when executed, runs a query directly against the state machine's events groove?

sentientwaffle (approved):
LGTM!

PR #2510, opened 2024-11-25
Devhub: Highlight count of untriaged issues

sentientwaffle:
(Make untriaged issues jump out at a glance. Feel free to rework the exact style, though!)


fabioarnold (approved):
image

I mean if it looks like that you want it to be 0, right? 😄

PR #2508, opened 2024-11-25
VOPR: Bump vopr-lite `requests_max`

sentientwaffle:
Since vopr-lite is green, bump up the `requests_max` to see what happens.

PR #2513, opened 2024-11-26
LSM/CacheMap: Scoped insert may be evicted into stash

sentientwaffle:
This crash was originally found by a [VOPR running with extra requests](https://github.com/tigerbeetle/tigerbeetle/pull/2508).
The cache map fuzzer with this branch's [modifications](https://github.com/tigerbeetle/tigerbeetle/pull/2513/commits/95ca2f70d069f3e0d81f490f85890a723418b904) can find it now as well.

## Bug

Consider this scenario for the transfers `CacheMap`:

1. `cache_map.scope_open()` (A linked transfer begins.)
2. `cache_map.upsert(.{ .key = 123 })`. This is an insert (the key does not already exist in the database), so we also `self.scope_rollback_log.appendAssumeCapacity(tombstone_from_key(key));`.
3. `cache_map.upsert(.{ .key = 456 })`.
    - Suppose `456` collides with `123` in the `SetAssociativeCache`, and the `123` insert is evicted from the cache.
    - Then `123`'s value is moved into the stash.
4. `cache_map.scope_close(.discard)` (The linked transfer fails and is rolled back).

In `cache_map.scope_close()`, we iterate the rollback log...

```zig
// The scope_rollback_log stores the operations we need to reverse the changes a scope
// made. They get replayed in reverse order.
var i: usize = self.scope_rollback_log.items.len;
while (i > 0) {
    i -= 1;
```

... when we reach the tombstone for `123` in our rollback log...

```zig
    const rollback_value = &self.scope_rollback_log.items[i];
    if (tombstone(rollback_value)) {
        // Reverting an insert consists of a .remove call.
        // The value in here will be a tombstone indicating the original value didn't
        // exist.
        const key = key_from_value(rollback_value);

        // A tombstone in the rollback log can only occur when the value doesn't exist
        // in _both_ the cache and stash on insert.
        if (self.cache) |*cache| {
```

... we find that `123` isn't in the cache (`!cache_removed`), so we panic:

```zig
            // If we have cache enabled, it must be there.
            const cache_removed = cache.remove(key) != null;
            assert(cache_removed);
        }
```

It is actually in the stash instead:

```zig
        // It should be in the stash _iif_ we don't have cache enabled.
        const stash_removed = self.stash_remove(key) != null;
        assert(stash_removed == (self.cache == null));
    } else {
        // Reverting an update or delete consists of an insert of the original value.
        self.upsert(rollback_value);
    }
}
```

<details>
<summary>Stack trace</summary>

    /var/home/djg/C/t/db/review/src/lsm/cache_map.zig:293:31: 0x159e6f3 in scope_close (vopr)
                            assert(cache_removed);
                                  ^
    /var/home/djg/C/t/db/review/src/lsm/groove.zig:1399:45: 0x154d500 in scope_close (vopr)
                groove.objects_cache.scope_close(mode);
                                                ^
    /var/home/djg/C/t/db/review/src/state_machine.zig:1565:62: 0x1504fce in scope_close (vopr)
                        self.forest.grooves.transfers.scope_close(mode);
                                                                 ^
    /var/home/djg/C/t/db/review/src/state_machine.zig:1652:45: 0x14d1174 in execute_create__anon_29789 (vopr)
                                self.scope_close(operation, .discard);
                                                ^
    /var/home/djg/C/t/db/review/src/state_machine.zig:1490:57: 0x147abad in commit (vopr)

</details>

## Fix

A tombstone in the rollback log means that there is a value in the cache *xor* in the stash.

## Why didn't we find this bug sooner?

- **Why didn't the VOPR find this bug?** The bug requires a cache collision (between two values both inserted in the same scope) and a linked transfer rollback – each of these are rare events in the VOPR.
- **Why didn't forest or tree fuzzers find this bug?** They don't test scopes. (Added a TODO.)
- **Why didn't cache map fuzzer find this bug?** Scopes were limited to 32 updates and hash collisions are rare. The fuzzer is now updated to use larger scopes, so it is capable of finding this bug.

### DIFF
--- a/src/lsm/cache_map.zig
+++ b/src/lsm/cache_map.zig
@@ -157,7 +157,8 @@ pub fn CacheMapType(
                 } else {
                     // If it was an insert, append a tombstone to the scope rollback log.
                     const key = key_from_value(value);
-                    self.scope_rollback_log.appendAssumeCapacity(tombstone_from_key(key));
+                    const key_tombstone = tombstone_from_key(key);
+                    self.scope_rollback_log.appendAssumeCapacity(key_tombstone);
                 }
             }
         }
--- a/src/lsm/cache_map.zig
+++ b/src/lsm/cache_map.zig
@@ -157,9 +157,7 @@ pub fn CacheMapType(
                 } else {
                     // If it was an insert, append a tombstone to the scope rollback log.
                     const key = key_from_value(value);
-                    self.scope_rollback_log.appendAssumeCapacity(
-                        tombstone_from_key(key),
-                    );
+                    self.scope_rollback_log.appendAssumeCapacity(tombstone_from_key(key));
                 }
             }
         }
@@ -282,15 +280,12 @@ pub fn CacheMapType(
 
                     // A tombstone in the rollback log can only occur when the value doesn't exist
                     // in _both_ the cache and stash on insert.
-                    if (self.cache) |*cache| {
-                        // If we have cache enabled, it must be there.
-                        const cache_removed = cache.remove(key) != null;
-                        assert(cache_removed);
-                    }
+                    const cache_removed =
+                        if (self.cache) |*cache| cache.remove(key) != null else false;
 
-                    // It should be in the stash _iif_ we don't have cache enabled.
+                    // The key should be in the stash iff it wasn't in the cache.
                     const stash_removed = self.stash_remove(key) != null;
-                    assert(stash_removed == (self.cache == null));
+                    assert(stash_removed != cache_removed);
                 } else {
                     // Reverting an update or delete consists of an insert of the original value.
                     self.upsert(rollback_value);
--- a/src/lsm/cache_map_fuzz.zig
+++ b/src/lsm/cache_map_fuzz.zig
@@ -374,16 +374,13 @@ pub fn generate_fuzz_ops(random: std.rand.Random, fuzz_op_count: usize) ![]const
                 break :blk FuzzOp{ .compact = {} };
             },
             .scope => blk: {
-                if (!scope_is_open) {
-                    scope_is_open = true;
-                    operations_since_scope_open = 0;
+                operations_since_scope_open = 0;
+                defer scope_is_open = !scope_is_open;
 
-                    break :blk FuzzOp{ .scope = .open };
-                } else {
-                    scope_is_open = false;
-                    operations_since_scope_open = 0;
-
+                if (scope_is_open) {
                     break :blk FuzzOp{ .scope = if (random.boolean()) .persist else .discard };
+                } else {
+                    break :blk FuzzOp{ .scope = .open };
                 }
             },
         };
--- a/src/lsm/cache_map_fuzz.zig
+++ b/src/lsm/cache_map_fuzz.zig
@@ -13,6 +13,11 @@ const log = std.log.scoped(.lsm_cache_map_fuzz);
 const Key = TestTable.Key;
 const Value = TestTable.Value;
 
+const map_value_count_max = 1024;
+// Use a large scope (relative to map_value_count_max) to increase the chances of
+// (SetAssociativeCache) hash collisions.
+const scope_value_count_max = map_value_count_max;
+
 const OpValue = struct {
     op: u32,
     value: Value,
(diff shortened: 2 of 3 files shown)


sentientwaffle:
CFO is reporting a failing `lsm_cache_map` seed, but it passes for me locally on the same commit:

```
$ git show | head -n1
commit 38e916fdf61f7577e9bcf5482990eef15881aef0
$ zig build -Drelease fuzz -- lsm_cache_map 4077271093479445358
info(fuzz): Fuzz seed = 4077271093479445358
info(lsm_cache_map_fuzz): fuzz_op_count = 633357495
info(lsm_cache_map_fuzz): fuzz_op_distribution = enums.EnumFieldStruct(@typeInfo(lsm.cache_map_fuzz.FuzzOp).Union.tag_type.?,f64,null){ .compact = 0.00e0, .get = 4.00e0, .upsert = 8.00e0, .remove = 0.00e0, .scope = 1.00e0 }
info(lsm_cache_map_fuzz): Verified 4721123 items from model exist and match in cache_map.
info(lsm_cache_map_fuzz): Verified all items in the cache and stash exist and match the model.
info(lsm_cache_map_fuzz): Passed lsm.cache_map.CacheMapType(u128,lsm.scan_fuzz.Thing,(function 'key_from_value'),(function 'hash'),(function 'tombstone_from_key'),(function 'tombstone')).Options{ .cache_value_count_max = 256, .map_value_count_max = 1024, .scope_value_count_max = 1024, .name = { 102, 117, 122, 122, 32, 109, 97, 112 } }!
info(lsm_cache_map_fuzz): Verified 4721123 items from model exist and match in cache_map.
info(lsm_cache_map_fuzz): Verified all items in the cache and stash exist and match the model.
info(lsm_cache_map_fuzz): Passed lsm.cache_map.CacheMapType(u128,lsm.scan_fuzz.Thing,(function 'key_from_value'),(function 'hash'),(function 'tombstone_from_key'),(function 'tombstone')).Options{ .cache_value_count_max = 0, .map_value_count_max = 1024, .scope_value_count_max = 1024, .name = { 102, 117, 122, 122, 32, 109, 97, 112 } }!
info(fuzz): done in 2m84.991ms
```

src/lsm/cache_map.zig
```diff
@@ -157,9 +157,7 @@ pub fn CacheMapType(
                 } else {
                     // If it was an insert, append a tombstone to the scope rollback log.
                     const key = key_from_value(value);
-                    self.scope_rollback_log.appendAssumeCapacity(
-                        tombstone_from_key(key),
-                    );
+                    self.scope_rollback_log.appendAssumeCapacity(tombstone_from_key(key));
```
matklad: FWIW, I slightly prefer the original formatting. The rule of thumb here is that I don't like code of the shape of `f(g(x))`, the composition of functions is easy to miss.  Though, I usually like to format those as

```
const y = g(x)
f(y)
```

src/lsm/cache_map.zig
```diff
@@ -282,15 +280,12 @@ pub fn CacheMapType(
 
                     // A tombstone in the rollback log can only occur when the value doesn't exist
                     // in _both_ the cache and stash on insert.
-                    if (self.cache) |*cache| {
-                        // If we have cache enabled, it must be there.
-                        const cache_removed = cache.remove(key) != null;
-                        assert(cache_removed);
-                    }
```
matklad, line 289: Ahhh, so I actually had a wrong mental model for the cache map! Rollbacks _dont_ guarantee that it ends up in _physically_ the same state, only that it would be logically the same. 

After the insert, the value here was in cache. But, before the rollback, it is in stash, not in cache. 

🤔 

I am now wondering if we compute the size of the stash correctly? That is, when we roll-forward, we spill some values into the stash. But when we roll back, because we are doing upserts, we might spill _more_ values into the stash?

That is, inserting `N` values into cache map does `N` inserts into the stash, but subsequent roll-back can do N more. I think this should be fine, as the keys are the same, but we'll need to double-check this!

matklad (dismissed):
Another great catch dj!

This looks correct to me, but it was surprising to me that our roll-back doesn't restore the original state of the data structure. I wonder if there might be more bugs around that?

src/lsm/cache_map.zig
sentientwaffle: Stepping through the logic:

During scope rollback, each stash insert comes from (non-tombstone) values in the scope rollback log.

Each of those inserts in the rollback log corresponds to either:
- a stash remove during scope roll-forward (capacity for the insert in the stash is guaranteed by the remove), or
- a cache eviction during scope roll-forward (capacity for the insert in the stash is unused).

So I think the stash capacity is correct.

matklad: >a stash remove during scope roll-forward

Does that leave a (hashmap-level) tombstone in place which still uses up capacity?

sentientwaffle: Hashmap-level tombstones don't contribute to hashmap capacity, do they?

matklad: Ah, right, tombstones are always re-used. I thought that might not be the case, but, checking the code and writing the test confirms that this is in fact fine!

matklad (approved):

PR #2514, opened 2024-11-26
Vortex: support tracing and check liveness

owickstrom:
This PR adds two things:

1. the ability to trace Vortex test runs
2. a liveness check

Both additions are based on a new progress event sent by the workload on stdout. The supervisor reads those and tracks finished requests and event counts.

## Tracing

Tracing is done with Chrome's [Trace Event Format](https://docs.google.com/document/d/1CvAClvFfyA5R-PhYUmn5OOQtYMH4h6I0nSsKchNAySU/preview?tab=t.0#heading=h.yr4qxyxotyw), which can be viewed in Chromium/Chrome by visiting `chrome://tracing`. Use WASD keys to move around and the tools at hotkeys 1-4.

Traces are enabled by passing the `--trace=<output path>` option to Vortex.

Here's a screenshot of how a trace might look.

tracing

## Liveness Checking

The progress events, in combination with the state of process and network faults controlled by the supervisor, is used to implement the liveness check. During _acceptably_ faulty periods we require some requests to finish after a certain amount of time (120 seconds). If there are network faults, which are currently global, or if there are more than `f` process faults (based on the `2f+1` equation), the liveness check is disabled.

Here's a case where a request took over 2 minutes to complete (in an acceptably faulty state), failing the test:

image

_(this was screenshot before I fixed the recording of the last request, so there's a lengthy yellow request bar missing)_

### DIFF
--- a/src/testing/vortex/constants.zig
+++ b/src/testing/vortex/constants.zig
@@ -1,3 +1,4 @@
+const std = @import("std");
 const constants = @import("../../constants.zig");
 
 pub usingnamespace constants;
@@ -9,7 +10,7 @@ pub const connections_count_max = @divFloor(constants.clients_max, replica_count
 // We allow the cluster to not make progress processing requests for this amount of time. After
 // that it's considered a test failure.
 pub const liveness_requirement_seconds = 120;
-pub const liveness_requirement_micros = liveness_requirement_seconds * 1_000_000;
+pub const liveness_requirement_micros = liveness_requirement_seconds * std.time.us_per_s;
 
 // How many replicas can be faulty while still expecting the cluster to make progress (based on
 // 2f+1).
--- a/src/testing/vortex/constants.zig
+++ b/src/testing/vortex/constants.zig
@@ -5,3 +5,11 @@ pub usingnamespace constants;
 pub const cluster_id = 1;
 pub const replica_count = 3;
 pub const connections_count_max = @divFloor(constants.clients_max, replica_count);
+
+// We allow the cluster to not make progress processing requests for this amount of time. After
+// that it's considered a test failure.
+pub const liveness_requirement_seconds = 120;
+
+// How many replicas can be faulty while still expecting the cluster to make progress (based on
+// 2f+1).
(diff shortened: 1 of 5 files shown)


src/testing/vortex/constants.zig
```diff
@@ -5,3 +5,12 @@ pub usingnamespace constants;
 pub const cluster_id = 1;
 pub const replica_count = 3;
 pub const connections_count_max = @divFloor(constants.clients_max, replica_count);
+
+// We allow the cluster to not make progress processing requests for this amount of time. After
+// that it's considered a test failure.
+pub const liveness_requirement_seconds = 120;
+pub const liveness_requirement_micros = liveness_requirement_seconds * 1_000_000;
```
matklad: There's `std.time.us_per_s`

src/testing/vortex/supervisor.zig
```diff
@@ -98,12 +127,20 @@ pub fn main(allocator: std.mem.Allocator, args: CLIArgs) !void {
         allocator.free(tmp_dir);
     }
 
+    var trace = if (args.trace) |trace_file_path|
+        try TraceWriter.from_file(trace_file_path)
+    else
+        .noop;
+    defer trace.deinit();
```
matklad: Let's reduce dimensionality here and always generate the trace into `${output_diretory}/trace.json`?

src/testing/vortex/supervisor.zig
```diff
@@ -328,10 +328,19 @@ const Supervisor = struct {
...
                     std.time.ns_per_s;
-                if (now > deadline and
-                    supervisor.workload.event_count_total == 0)
-                {
-                    log.err("no successful requests after {d} seconds of {d} faulty replica(s)", .{
+                const no_finished_requests =
+                    supervisor.workload.request_duration_max_micros == null;
+                const too_slow_request =
+                    if (supervisor.workload.request_duration_max_micros) |duration|
+                    duration > (constants.liveness_requirement_seconds * 1_000_000)
```
matklad: call it `duration_micros` and use `std.time.us_per_second`

matklad (approved):

PR #2522, opened 2024-11-28
fix: enforce single-object restrictions for non-sliceable operations & apply TigerStyle

martijngonlag:
Previously, attempting to use multiple objects on operations that do not support them (e.g. `get_account_transfers`) would cause the REPL to panic.

The addition of the Python client exposed `StateMachine.event_is_slice(...)`, providing a timely solution to implement this check. The function reference can be found here: https://github.com/tigerbeetle/tigerbeetle/blob/c1c325e55d1015169aea491d94ff5a49d9b9ec6c/src/state_machine.zig#L417-L431

Instead of panicking when multiple objects are provided on operations that don't support it, the code now gracefully returns an appropriate error with feedback to the user. For example:

```s
Input:
> `get_account_transfers account_id=1, account_id=2`

Output:
> Fail near line 1, column 35:
>
> get_account_transfers account_id=1, account_id=2
>                                    ^ Near here.
> get_account_transfers expects a single account_filter but received multiple.
```

### DIFF
--- a/src/repl.zig
+++ b/src/repl.zig
@@ -23,10 +23,10 @@ pub const Parser = struct {
     terminal: *const Terminal,
 
     pub const Error = error{
-        KeyValuePairBad,
-        ValueBad,
-        OperationBad,
         IdentifierBad,
+        OperationBad,
+        ValueBad,
+        KeyValuePairBad,
         KeyValuePairEqualMissing,
         SyntaxMatchNone,
         SliceOperationUnsupported,
@@ -333,7 +333,7 @@ pub const Parser = struct {
                 const state_machine_op = std.meta.stringToEnum(
                     StateMachine.Operation,
                     @tagName(operation),
-                ) orelse return error.SliceOperationUnsupported;
+                ).?;
 
                 if (!StateMachine.event_is_slice(state_machine_op)) {
                     try parser.print_current_position();
--- a/src/repl.zig
+++ b/src/repl.zig
@@ -23,12 +23,13 @@ pub const Parser = struct {
     terminal: *const Terminal,
 
     pub const Error = error{
-        BadKeyValuePair,
-        BadValue,
-        BadOperation,
-        BadIdentifier,
-        MissingEqualBetweenKeyValuePair,
-        NoSyntaxMatch,
+        KeyValuePairBad,
+        ValueBad,
+        OperationBad,
+        IdentifierBad,
+        KeyValuePairEqualMissing,
+        SyntaxMatchNone,
+        SliceOperationUnsupported,
     };
 
     pub const Operation = enum {
@@ -131,7 +132,7 @@ pub const Parser = struct {
             return;
         }
 
-        return Error.NoSyntaxMatch;
+        return Error.SyntaxMatchNone;
     }
 
     fn parse_value(parser: *Parser) []const u8 {
@@ -315,13 +316,12 @@ pub const Parser = struct {
         var object_has_fields = false;
         while (parser.offset < parser.input.len) {
             parser.eat_whitespace();
-            // Always need to check `i` against length in case we've hit the end.
+            // Always need to check i against length in case we've hit the end.
             if (parser.offset >= parser.input.len or parser.input[parser.offset] == ';') {
                 break;
             }
 
             // Expect comma separating objects.
-            // TODO: Not all operations allow multiple objects, e.g. get_account_transfers.
             if (parser.offset < parser.input.len and parser.input[parser.offset] == ',') {
                 parser.offset += 1;
                 inline for (@typeInfo(ObjectSyntaxTree).Union.fields) |object_tree_field| {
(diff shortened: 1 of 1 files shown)


src/repl.zig
```diff
@@ -29,6 +29,7 @@ pub const Parser = struct {
         BadIdentifier,
         MissingEqualBetweenKeyValuePair,
         NoSyntaxMatch,
+        UnsupportedSliceOperation,
```
jorangreef:
```suggestion
        SliceOperationUnsupported,
```

Per TigerStyle, to put the noun first and the qualifier last. This way, if `SliceOperation` has other errors, they all share the same prefix.

Note also that other errors here, before this PR, don't follow TigerStyle. We could go through these and probably clean up the naming.

src/repl.zig
```diff
@@ -1904,3 +1919,40 @@ test "repl.zig: Handle parsing errors" {
...
+            .in = "query_accounts account_id=1, account_id=2",
+            .err = error.UnsupportedSliceOperation,
+        },
+        .{
+            .in = "query_transfers account_id=1, account_id=2",
+            .err = error.UnsupportedSliceOperation,
+        },
+    };
+
+    for (tests) |t| {
```
jorangreef:
```suggestion
    for (vectors) |vector| {
```

Per TigerStyle, don't abbreviate variable names.

Also a practice of the code base to try to use names that have a clear singular and plural. And these are more test vectors (within a test).

src/repl.zig
```diff
@@ -1904,3 +1919,40 @@ test "repl.zig: Handle parsing errors" {
         try std.testing.expectError(t.err, result);
     }
 }
+
+test "repl.zig: Parser fails for operations not supporting multiple objects" {
+    const tests = [_]struct {
+        in: []const u8,
+        err: anyerror,
```
jorangreef:
```suggestion
        string []const u8,
        result: anyerror,
```

Per TigerStyle, don't abbreviate.

Also per TigerStyle, where names go together in pairs like here, try to choose variable names with the same number of characters. This has the 2nd order effect of letting similar code all line up in the source. It's subtle but when you know to look for it, you'll see it all over the code base, and it's intentional.

jorangreef: For example, if you were writing a routine to copy from a to b, you wouldn't use names like `src` and `dest` which are common in C code, but rather `source` and `target` (in full, not abbreviated, and they line up symmetrically, which makes it easy on the eye and easy for the reader to compare similar blocks of code at a glance).

jorangreef (commented):
Nice work, @martijngonlag!

I'll leave the correctness review to @batiati but since this is one of your first PRs, I wanted to encourage you with some extra review.

src/repl.zig
```diff
@@ -330,6 +330,19 @@ pub const Parser = struct {
                         try arguments.appendSlice(std.mem.asBytes(&unwrapped_field));
                     }
                 }
+                const state_machine_op = std.meta.stringToEnum(
+                    StateMachine.Operation,
+                    @tagName(operation),
+                ) orelse return error.SliceOperationUnsupported;
```
batiati: In this case, I suppose `orelse unreachable` would be a better fit, since `operation` was already parsed and validated.

Note: You can use `.?`, since `optional.?` is the shorthand for `optional orelse unreachable` .

src/repl.zig
```diff
@@ -330,6 +330,19 @@ pub const Parser = struct {
...
+
+                if (!StateMachine.event_is_slice(state_machine_op)) {
+                    try parser.print_current_position();
+                    try parser.terminal.print_error(
+                        "{s} expects a single {s} but received multiple.\n",
+                        .{ @tagName(operation), @tagName(object) },
+                    );
+                    return error.SliceOperationUnsupported;
+                }
```
batiati, line 345: :heart:

src/repl.zig
```diff
@@ -23,12 +23,13 @@ pub const Parser = struct {
...
-        BadIdentifier,
-        MissingEqualBetweenKeyValuePair,
-        NoSyntaxMatch,
+        KeyValuePairBad,
+        ValueBad,
+        OperationBad,
+        IdentifierBad,
+        KeyValuePairEqualMissing,
+        SyntaxMatchNone,
+        SliceOperationUnsupported,
```
batiati: Just for better aligning, move `KeyValuePairEqualMissing` up, close to `KeyValuePairBad`.

**Short note on naming things (it's hard!):**

The TigerStyle tells us to put the most significant word first, so things such as "the maximum size of a batch" should be named `batch_size_max` instead of `max_batch_size`.

However, it's at our discretion to identify the most significant word for the context.
One great example, the results `exceeds_credits` and `exceeds_debits`, where "exceeds" is in fact the reason why the transfer failed. 

One could argue why "Bad" isn't the most important word in this particular case, to align `BadKeyValuePair`, `BadValue`, `BadOperation`: Since it's an enum for parser error, we expect that all values would be something related to "bad", "invalid", "missing" or whatever ... so those words add very little meaning in the name.

martijngonlag: Makes sense! They are already aligned elsewhere in the code (ex: https://github.com/tigerbeetle/tigerbeetle/blob/main/src/repl.zig#L903) so I will align it with those two occurrences.

batiati (approved):

Issue #2509, opened 2024-11-25
Tigerbeetle Repl Client cannot access Tigerbeetle Server through virtual networking


ashley-nygaard:
I'm running into an issue trying to access Tigerbeetle via the SDK or Repl when the server is not directly on the same host.
I've created a VM of Ubuntu 22.04.5 Live Server arm64.
I have confirmed that I can host a static website via nginx from the VM and access the site on the host machine using port forwarding configuration in VirtualBox.
That confirms that the host can 'see' inside the VM.
Theoretically, with similar forwarding configuration for Tigerbeetle, I should be able to use the Repl to connect to Tigerbeetle running inside the VM.

When using the tigerbeetle-go client or the REPL from the host machine, I cannot access the running Tigerbeetle cluster.

The errors I receive are:

```plaintext
info(message_bus): peer performed an orderly shutdown: message_bus.MessageBusType(.client).Connection.Peer{ .replica = 0 }
```

and

```plaintext
error(message_bus): error receiving from message_bus.MessageBusType(.replica).Connection.Peer{ .replica = 0 }: error.ConnectionResetByPeer` when the client tries to connect.
```

When using Tigerbeetle and the client (or Repl) on the same machine everything connects as expected.

Expected Outcome:

Tigerbeetle to allow connections when the client is not on the same machine.

- VM
  - ubuntu-22.04.5-live-server-arm64
- Host
  - M1 Sonoma 14.0

sentientwaffle:
Hi @ashley-nygaard -- can you share the snippet of code you used to initialize tigerbeetle-go?

Also, which versions of the replica and the client are you using? e.g. 0.16.x?

ashley-nygaard:
Hi @sentientwaffle I'm trying the examples in the Quick Start.

For the Repl I'm using [this example](https://docs.tigerbeetle.com/quick-start#4-connect-to-the-repl):

```bash
./tigerbeetle repl --cluster=0 --addresses=3000
```

Then under the `tigerbeetle-go` client, I'm just using the connection example [here](https://docs.tigerbeetle.com/quick-start#4-connect-to-the-repl).

Anything above `tigerbeetle-go@0.16.1` and I get the following error:

```plaintext
warning(replica): 0: on_request: ignoring unsupported client version; too high (client=244598801749443206418031865540777813779 version=0.16.14>0.16.1)
error(replica): 0: sending eviction message to client=244598801749443206418031865540777813779 reason=client_release_too_high
[...]

PR #2523, opened 2024-11-29
docs: Add note to docker server startup process

martijngonlag:
In #2509, @ashley-nygaard had an issue connecting a local TB client (running on macos) to a TB replica running in Docker. After some troubleshooting, we determined that the solution was to explicitly set the `--address` parameter to the IP address of the Docker container. Updating the documentation with this note to make future developers aware.

### DIFF
--- a/docs/operating/docker.md
+++ b/docs/operating/docker.md
@@ -33,6 +33,8 @@ info(io): allocating 660.140625MiB...
 
 ## Run the Server
 
+> **_NOTE:_** To allow external clients to connect, replace `0.0.0.0` with the container's IP address
+
 ```console
 docker run -it --security-opt seccomp=unconfined \
     -p 3000:3000 -v $(pwd)/data:/data ghcr.io/tigerbeetle/tigerbeetle \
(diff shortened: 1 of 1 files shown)


docs/operating/docker.md
```diff
@@ -33,6 +33,8 @@ info(io): allocating 660.140625MiB...
 
 ## Run the Server
 
+> **_NOTE:_** To allow external clients to connect, replace `0.0.0.0` with the container's IP address
```
sentientwaffle, line 36: Why doesn't `0.0.0.0` work -- that should listen on all available interfaces, right?

martijngonlag: TLDR; I made a poor assumption in troubleshooting #2509 and my previous experience with networking and Docker. The existing command works fine as-is; the "issue" was elsewhere, namely, when you start TB with `--addresses=3000` it defaults to `127.0.0.1:3000`, which on a VM, is a different localhost. If you start it with `--addresses=0.0.0.0:3000` it works with port-forwarding as expected.

PR #2534, opened 2024-12-02
ci: add arm64 runners & fix arm64 client ABI issues

cb22:
We consider arm64 Linux a supported platform, for development at least, but currently haven't been running any CI against it.

This PR adds arm64 runners, testing both the replica and clients, and fixes a few bugs along the way.

Multiversion tests on Linux arm64 were broken because they were trying to use x86_64 objcopy, so add the new objcopy platform to our [dependencies](https://github.com/tigerbeetle/dependencies) repo and use it where appropriate.

The dotnet and Python clients were affected by an ABI difference on arm64 Linux: a `struct {u64, u64}` does not have the same ABI as a `u128`, so trying to pass this by value (which is how they internally treat `u128`s) would lead to the function call to `tb_client_init` getting passed parameters incorrectly and segfaulting.

Fix this, by treating the `u128` as a `*const [16]u8` and copying and casting the raw bytes to avoid alignment and ABI issues.

### DIFF
--- a/src/tb_client_exports.zig
+++ b/src/tb_client_exports.zig
@@ -1,4 +1,5 @@
 const std = @import("std");
+const stdx = @import("stdx.zig");
 const builtin = @import("builtin");
 
 // TODO: Move this back to src/clients/c when there's a better solution for main_pkg_path=src/
@@ -28,13 +29,25 @@ comptime {
 
 fn init(
     out_client: *tb.tb_client_t,
-    cluster_id: u128,
+    cluster_id_ptr: *const [16]u8,
     addresses_ptr: [*:0]const u8,
     addresses_len: u32,
     on_completion_ctx: usize,
     on_completion_fn: tb.tb_completion_t,
 ) callconv(.C) tb.tb_status_t {
     const addresses = @as([*]const u8, @ptrCast(addresses_ptr))[0..addresses_len];
+
+    // Passing u128 by value is prone to ABI issues. Pass as a [16]u8, and explicitly copy into
+    // memory we know will be aligned correctly. Don't just use bytesToValue here, as that keeps
+    // pointer alignment, and will result in a potentially unaligned access of a
+    // `*align(1) const u128`.
+    const cluster_id: u128 = blk: {
+        var cluster_id: u128 = undefined;
+        stdx.copy_disjoint(.exact, u8, std.mem.asBytes(&cluster_id), cluster_id_ptr);
+
+        break :blk cluster_id;
+    };
+
     const client = tb.init(
         std.heap.c_allocator,
         cluster_id,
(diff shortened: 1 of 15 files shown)


src/clients/dotnet/TigerBeetle/UInt128Extensions.cs
```diff
@@ -16,7 +16,7 @@ public static class UInt128Extensions
     /// for P/Invove in methods marked with the [DllImport] attribute.
     /// It's necessary only because P/Invoke's marshaller does not support System.UInt128.
     /// </summary>
-    [StructLayout(LayoutKind.Explicit, Size = SIZE)]
+    [StructLayout(LayoutKind.Explicit, Pack = 16, Size = SIZE)]
```
batiati: I think we don't need `Pack=16` here.

src/tb_client_exports.zig
```diff
@@ -28,13 +29,23 @@ comptime {
...
     const addresses = @as([*]const u8, @ptrCast(addresses_ptr))[0..addresses_len];
+
+    // Passing u128 by value is prone to ABI issues. Pass as a [16]u8, and explicitly copy into
+    // memory we know will be aligned correctly.
+    const cluster_id: u128 = blk: {
+        var cluster_id: u128 = undefined;
+        stdx.copy_disjoint(.exact, u8, std.mem.asBytes(&cluster_id), cluster_id_ptr);
+
+        break :blk cluster_id;
+    };
```
batiati, line 49:
```suggestion
    const cluster_id = std.mem.bytesToValue(u128, cluster_id_ptr);
```

batiati (commented):
Can you please add a brief description to this PR?

cb22:
> Can you please add a brief description to this PR?

Whoops, thought I added it!

src/tb_client_exports.zig
cb22: I thought about `bytesToValue`, but isn't there potential for an alignment issue here? (I'm not sure!)

batiati: Looking at the std lib implementation, it just cast and dereference the pointer... I doubt it'll cause any issues with alignment (IIUC the issue was caused in the argument list when crossing the FFI). But also I'm not sure 😰

Is there a way to validate this? At least we could add a comment explaining why we avoid bytesToValue (maybe a good candidate for stdx).

cb22: > IIUC the issue was caused in the argument list when crossing the FFI

Yep, it was an assembly level / ABI issue.

> Is there a way to validate this? At least we could add a comment explaining why we avoid bytesToValue (maybe a good candidate for stdx).

So, `bytesToValue` calls and uses `bytesAsValue`, which does keep the pointer alignment, so it's accessing and copying from a `*align(1) const u128`.

I think this could cause unaligned accesses - which are seemingly handled both on x86_64 and arm64, but maybe it would be better to avoid it entirely...

batiati (dismissed):

cb22:
(added a comment explaining the non-use of `bytesToValue`)

batiati (approved):

PR #2535, opened 2024-12-03
cfo: simplify memory management

matklad:
This was _supposed_ to fix our CFO OOM issue, but I realized that I don't quite see a nice way to do that just yet (asked the community for inspiration at https://ziggit.dev/t/how-to-print-path-to-a-freshly-build-executable-in-zig-cache/7157), so here's just a small simplifying refactoring!


matklad:
OK, so looks like https://github.com/tigerbeetle/tigerbeetle/pull/2526 was just wrong --- it checks the status of `zig/zig build` process, but what actually gets sigkill delivered is the underling fuzz_test process!

cb22 (approved):

PR #2536, opened 2024-12-03
LSM: Clarify Tree.table_count_max vs Forest.table_count_max

cb22:
The two have the same values currently, but are different. This commit:
* makes it clearer which is being used where,
* and fixes a latent bug where manifest_log specific code uses the per tree value rather than the per forest value.

### DIFF
--- a/src/lsm/manifest.zig
+++ b/src/lsm/manifest.zig
@@ -9,7 +9,7 @@ const constants = @import("../constants.zig");
 const growth_factor = constants.lsm_growth_factor;
 
 const vsr = @import("../vsr.zig");
-const table_count_max = @import("tree.zig").table_count_max;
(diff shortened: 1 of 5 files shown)


src/lsm/manifest.zig
```diff
@@ -9,7 +9,7 @@ const constants = @import("../constants.zig");
 const growth_factor = constants.lsm_growth_factor;
 
 const vsr = @import("../vsr.zig");
-const table_count_max = @import("tree.zig").table_count_max;
+const table_count_max_tree = @import("tree.zig").table_count_max;
```
cb22, line 12: There is `forest_table_count_max` below, however. This code lines up nicely as `table_count_max_tree` and the below code as `forest_table_count_max` :sweat:.

sentientwaffle (approved):

PR #2537, opened 2024-12-04
LSM: fatal with an explicit error if tables are exhausted

cb22:
Also, fixes a latent integer overflow bug.

### DIFF
--- a/src/lsm/manifest_log.zig
+++ b/src/lsm/manifest_log.zig
@@ -213,7 +213,10 @@ pub fn ManifestLogType(comptime Storage: type) type {
             manifest_log.table_extents = TableExtents{};
             try manifest_log.table_extents.ensureTotalCapacity(
                 allocator,
-                options.forest_table_count_max,
+                // Allocate space for one additional table, so that the code can still use
+                // `getOrPutAssumeCapacity` while making it easier to check if the limit has been
+                // exceeded to error with a friendly message.
+                options.forest_table_count_max + 1,
             );
             errdefer manifest_log.table_extents.deinit(allocator);
 
@@ -398,6 +401,7 @@ pub fn ManifestLogType(comptime Storage: type) type {
                         const extent =
                             manifest_log.table_extents.getOrPutAssumeCapacity(table.address);
                         if (!extent.found_existing) {
+                            manifest_log.check_tables_count();
                             extent.value_ptr.* = .{ .block = block_address, .entry = entry };
                             manifest_log.open_event(manifest_log, table);
                         }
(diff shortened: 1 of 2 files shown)


src/lsm/manifest_log.zig
```diff
@@ -1239,6 +1239,14 @@ const Pace = struct {
...
         tables_count: u32,
     }) u32 {
+        if (options.tables_count > pace.tables_max) {
+            vsr.fatal(
+                .forest_tables_count_would_exceed_limit,
+                "forest_tables_count would exceed limit (tables_count={} tables_max={}) - " ++
+                    "please contact the team directly who will be able to assist",
+                .{ options.tables_count, pace.tables_max },
+            );
+        }
```
matklad: I'd say a good rule-of-thumb for this sort of checks is to detect the limit just before it is exceeded, rather than at some later point. Perhaps there are _other_ code paths that rely on the fact that table_max is not exceeded?

So, when there's an invariant guarding data D, check it when D is modified, not when it is read. 

Here, I think we should keep this assert as is, and rather add a check at `open_read_block_callback` and `append_internal`?

src/lsm/manifest_log.zig
```diff
@@ -1257,7 +1265,7 @@ const Pace = struct {
         // Our "target" block count extrapolates a log block count from our table count and the
         // log's maximum load factor.
         const log_blocks_target = @max(1, @divFloor(
-            pace.log_blocks_cycle_max * options.tables_count,
+            @as(u64, @intCast(pace.log_blocks_cycle_max)) * options.tables_count,
```
matklad: Good catch! It seems like at least `log_blocks_cycle_max` and `log_blocks_full_max` in Pace want to be u64s, so let's maybe make anything block-related in Pace to be u64, and leave only tables as `u32`?

src/lsm/manifest_log.zig
cb22: > So, when there's an invariant guarding data D, check it when D is modified, not when it is read.

Agreed!

It was a little tricky here, because in `append_internal` it uses `getOrPutAssumeCapacity`, and it is allowable for a `.insert` to overwrite a previous record due to compaction... So I went with a bit of a creative solution.

src/lsm/manifest_log.zig
```diff
@@ -536,10 +540,14 @@ pub fn ManifestLogType(comptime Storage: type) type {
                 .update,
                 => {
                     const extent = manifest_log.table_extents.getOrPutAssumeCapacity(table.address);
-                    if (extent.found_existing) {
-                        maybe(table.label.event == .insert); // (Compaction.)
-                    } else {
+                    if (!extent.found_existing) {
```
cb22, line 543: Swapped the `if` here to line up with the `getOrPutAssumeCapacity` in `open_read_block_callback`.

src/lsm/manifest_log.zig
```diff
@@ -213,7 +213,10 @@ pub fn ManifestLogType(comptime Storage: type) type {
             manifest_log.table_extents = TableExtents{};
             try manifest_log.table_extents.ensureTotalCapacity(
                 allocator,
-                options.forest_table_count_max,
+                // Allocate space for one additional table, so that the code can still use
+                // `getOrPutAssumeCapacity` while making it easier to check if the limit has been
+                // exceeded to error with a friendly message.
+                options.forest_table_count_max + 1,
```
matklad, line 219: 🤣 👍

matklad (approved):

PR #2539, opened 2024-12-04
log: change comptime log to .debug and add runtime filtering

cb22:
Previously, .debug log messages were filtered out at comptime. While nice from a binary size and performance (more on that later) point of view, this meant that a user couldn't specify `--log-debug` or similar and would need a new binary to get debug logging.

Make it so that we do runtime log filtering instead. The performance penalty for this was negligible in testing; both microbenchmarks and the actual benchmark.

The one downside is binary bloat: this increases the size of a release build by ~300KiB.

To use, specify `--log-debug --experimental` on the CLI when starting a replica. `--log-debug` instead of `--log=...` because I feel we don't want to allow levels quieter than `.info` in any case.

It's currently only supported for `start`. Scaffolding for clients is there too, with filtering of the messages before they cross FFI and requiring an additional flag to be set. Actually setting this flag from any client libraries is still unimplemented - but this makes it no harder than before to change client logging (edit code & recompile).

Things like the VOPR and fuzzers use their own way of setting logging, and are unaffected.

### DIFF
--- a/src/clients/c/tb_client.zig
+++ b/src/clients/c/tb_client.zig
@@ -75,15 +75,21 @@ pub const Logging = struct {
         comptime format: []const u8,
         args: anytype,
     ) void {
-        // Messages are silently dropped if no logging callback is specified - unless they're warn
-        // or err. The value in having those for debugging is too high to silence them, even until
-        // client libraries catch up and implement a callback handler.
-        const callback = logging.callback orelse {
-            if (message_level == .warn or message_level == .err) {
-                std.log.defaultLog(message_level, scope, format, args);
-            }
+        // Debug logs are dropped here unless debug is set, because of the potential penalty in
+        // crossing FFI to drop them.
+        if (message_level == .debug and !logging.debug) {
             return;
-        };
+        }
+
+        // Other messages are silently dropped if no logging callback is specified - unless they're
+        // warn or err. The value in having those for debugging is too high to silence them, even
+        // until client libraries catch up and implement a callback handler.
+        if (logging.callback == null and (message_level == .warn or message_level == .err)) {
+            std.log.defaultLog(message_level, scope, format, args);
+            return;
+        }
+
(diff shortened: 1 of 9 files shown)


src/clients/c/tb_client.zig
```diff
@@ -75,10 +75,14 @@ pub const Logging = struct {
...
-        // Messages are silently dropped if no logging callback is specified - unless they're warn
-        // or err. The value in having those for debugging is too high to silence them, even until
-        // client libraries catch up and implement a callback handler.
-        const callback = logging.callback orelse {
+        // Debug logs are dropped here unless debug is set, because of the potential penalty in
+        // crossing FFI to drop them.
+        // Other messages are silently dropped if no logging callback is specified - unless they're
+        // warn or err. The value in having those for debugging is too high to silence them, even
+        // until client libraries catch up and implement a callback handler.
+        const callback = if (message_level == .debug and logging.debug)
```
matklad: meant to say

```suggestion
        const callback = if (message_level == .debug and !logging.debug)
```

perhaps?

Could you also separate out this if to be a distinct statement? Right now, the if is quite busy and feels like it handles separate concerns? 

Maybe avoid the ternary at all?

```zig
if (message_level == .debug and !logging.debug) {
    return;
}

if (logging.callback == null and 
      (message_level == .warn or message_level === .err)) {
  std.log.defaultLog(message_level, scope, format, args);
  return;
}

const callback = logging.callback orelse return;
```

matklad (commented):
<3!

src/clients/c/tb_client.zig
cb22: Whoops! Good catch.

Yeah agreed, I thought originally there might be a tiny performance benefit from a ternary, but I'm probably wrong and client logging is called way less than replica logging so it's less relevant too.

matklad (approved):

PR #2540, opened 2024-12-05
cli: optionally generate random cluster id in format

cb22:
When formatting a TigerBeetle cluster, it's important to give it a random cluster ID to differentiate it from other clusters. Cluster IDs can't be changed after format, so it's important to get this right at the beginning.

However, currently you have to generate this externally which is cumbersome and leads to our docs and most people using `--cluster=0`, which is potentially unsafe.

This PR allows omitting the `--cluster` argument from format entirely, which will make it generate a random ID and print that out, so it can be used when formatting other replicas and for the client.

It also makes `--cluster=0` an explicit warning.

### DIFF
--- a/src/tigerbeetle/cli.zig
+++ b/src/tigerbeetle/cli.zig
@@ -31,7 +31,7 @@ const StateMachine = vsr.state_machine.StateMachineType(
 
 const CLIArgs = union(enum) {
     const Format = struct {
-        cluster: u128,
+        cluster: ?u128 = null,
         replica: ?u8 = null,
         // Experimental: standbys don't have a concrete practical use-case yet.
         standby: ?u8 = null,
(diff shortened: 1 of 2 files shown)


cb22:
This doesn't touch our docs yet, and I think the easiest way might be to not change that with the warning. 

Otherwise, we either need to indicate in the docs something like `--cluster=<use the value from above>`, make it a standalone subcommand and script syntax `cluster=$(tigerbeetle random)` `--cluster=$cluster`, or make it print only the cluster ID on stdout and use that.

batiati (commented):
That's a great idea!

Just one comment: How easily can we automate scripts for provisioning a cluster this way? For example, retrieving the generated cluster ID to store it in a configuration/secret? Otherwise, the `cluster=$(tigerbeetle random)` idea sounds better (and even more orthogonal).

cb22:
> Just one comment: How easily can we automate scripts for provisioning a cluster this way? For example, retrieving the generated cluster ID to store it in a configuration/secret? Otherwise, the cluster=$(tigerbeetle random) idea sounds better (and even more orthogonal).

Yeah - it's a tricky one. I'm not so sure we want to commit to having a dedicated subcommand just yet, so we might add that in future. Especially given the numerous different ways one could provision a cluster.

I do think it's important to have the default 'do the right thing', so that would be in addition to the current cluster ID change!

I've also added some extra logging, on actual replica startup (but it might be a bit alarming, so perhaps we should gate it behind a superblock format version?).

batiati (approved):
That makes sense! I think we can elegantly solve scripting by implementing a way to retrieve the cluster ID using the existing sub-command `./tigerbeetle inspect`!

PR #2541, opened 2024-12-05
test: fix possible resource leak

owickstrom:
There's a loop initializing the proxies used by Vortex's network fault injection, and in cases were some proxies failed to initialize, the successfully initialized ones wouldn't be properly cleaned up again.

### DIFF
--- a/src/testing/vortex/faulty_network.zig
+++ b/src/testing/vortex/faulty_network.zig
@@ -462,6 +462,11 @@ const Proxy = struct {
     accept_fd: std.posix.socket_t,
     mapping: Mapping,
     connections: [constants.connections_count_max]Connection,
+
+    fn deinit(proxy: *Proxy) void {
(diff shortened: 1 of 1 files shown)


src/testing/vortex/faulty_network.zig
```diff
@@ -462,6 +462,10 @@ const Proxy = struct {
     accept_fd: std.posix.socket_t,
     mapping: Mapping,
     connections: [constants.connections_count_max]Connection,
+
+    fn deinit(proxy: *Proxy) void {
+        proxy.io.close_socket(proxy.accept_fd);
```
matklad, line 467: Could you also add `proxy.* = undefined` to help making improper usage after deinit fail loudly?

owickstrom: Done :+1:

matklad (approved):

Issue #2515, opened 2024-11-26
Use StaticAllocator in Repl.run


brson:
As we add more features to the repl, we're having to be careful about how it handles dynamic allocation. Today it's easy to accidentally add new dynamic allocation because we aren't using the `StaticAllocator`, which can turn off allocation after init with the `transition_from_init_to_static` method.

We should use this allocator in the repl.

Today the repl uses two allocators: the "main" arena, in which we do all the upfront allocation, the `arena` argument to `Repl.run`; and the "execution arena" which is reallocated every time around the repl loop and is used only by the parser.

As a first step, we can transition the main arena to be wrapped by a `StaticAllocator` in the `Repl.run` method, similar to what `Replica.open` does; call `transition_from_init_to_static` appropriately.

Making the parser work without dynamic allocation can be a separate step.

cc @raghunandanbhat

brson:
A complication here is that the repl uses the Client, which does dynamic allocation. To make the rest of the repl use static allocation, the client will have to get its own allocator.

fidelicura:
Shouldn't this be closed? See merged https://github.com/tigerbeetle/tigerbeetle/pull/2542.

brson:
@fidelicura It looks to me like there's one more step to do here.

That PR restructured the allocation in the repl so that it is done upfront, has init/deinit methods. The repl though still doesn't actually use the `StaticAllocator` to enforce that no reallocation happens, in the way `replica.zig` does.

Probably in `Repl.init` or `Repl.run`, it needs to wrap the given arena allocator in `StaticAllocator`, do its allocation, then call `transition_from_init_to_static`.

PR #2542, opened 2024-12-05
REPL: Switch to static allocation

### LINKED ISSUE #2528: Add deallocation and deinitialization to the repl
The entire repl assumes it is using arena allocators and neither deallocates nor runs deinit functions. As the repl grows, this is looking wrong, and opens the door to future bugs.

We should add deinit methods, calls to deinit and free on errdefer, etc.

cb22:
This was bothering me when I was making the REPL evented (the REPL currently doesn't process any IO while waiting for user input...) :smile:.

Went with stack allocation (`BoundedArray`) for most small things, with an `ArrayListUnmanaged` for `arguments` which stores the events sent by the client; and an `init` / `deinit` that handles the setup and teardown for `run()`.

Also, remove the dangling `allocator` reference that was in client.zig.

Closes #2528 and #2515.

### DIFF
--- a/src/repl.zig
+++ b/src/repl.zig
@@ -263,7 +263,7 @@ pub const Parser = struct {
     fn parse_arguments(
         parser: *Parser,
         operation: Operation,
-        arguments: *std.ArrayList(u8),
+        arguments: *std.ArrayListUnmanaged(u8),
     ) !void {
         const default: ObjectSyntaxTree = switch (operation) {
             .help, .none => return,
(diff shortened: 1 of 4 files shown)


brson (approved):
Looks great.

The way the bounds of the arguments array is known statically, with the parser parsing directly into a flat message buffer, is pretty cool.

I do have some concern that it may be possible to overflow the capacity of `arguments` due to logic bugs in the parser, but can't prove it at the moment, and don't feel it needs to block the PR. (In particular, in testing it seems that fields can be repeated multiple times without producing an error, which maybe could lead to an overflow. But if so that's a preexisting logic bug. I'll make a note to investigate later.)

PR #2544, opened 2024-12-06
LSM/Scan: Fix missing intersection results bug 

sentientwaffle:
## Summary

Fix a correctness bug in which some queries erroneously end a scan too early, truncating the results.

This bug is only in the read-path -- the omitted objects are still safely durable in the LSM.

This bug can affect any scan which intersects multiple secondary index trees.
That is, depending on what filters are included, it can be hit by `get_account_balances`, `get_account_transfer`, `query_accounts`, and `query_transfers`.

This bug is deterministic -- when one replica hits this bug, all of them will, regardless of how their IO is reordered.
(So even with this bug, the checkpoint won't diverge due to distinct client replies.)

## Bug

A scan may intersect many `ScanTree`s together.
The intersection is implemented with zig-zag merge join.

When any of the intersected `ScanTree`s races ahead, the merge algorithm informs the lagging trees about the leading key (`ScanTree.probe()`).
This lets those other trees skip past ranges of keys that it now knows will definitely not intersect.

Suppose that `ScanTree.probe()` is called, which increases the `ScanTree`'s `key_min` (ascending) or decreases the `key_max` (descending).
Then this calls `ScanTreeLevel.probe()` for each level:

```zig
pub fn probe(self: *ScanTreeLevel, probe_key: Key) void {
    switch (self.values) {
        .fetching => {},
        .buffered => |buffer| {
```

If `ScanTreeLevel.probe()` has values already buffered in memory...

```zig
            const slice: []const Value = ScanTree.probe_values(
                self.scan.direction,
                buffer,
                probe_key,
            );

            if (slice.len == 0) {
                // ...
                self.move_next();
```

...and none of the buffered values intersect with the new shrunk range, then we need to get a new value block, so it calls `ScanTreeLevel.move_next()`.

```zig
fn move_next(self: *ScanTreeLevel) void {
    switch (self.state) {
        .loading_manifest => self.move_next_manifest_table(null),
        .loading_index => unreachable,
        .iterating => |*iterating| {
            if (iterating.values == .none or
                iterating.values.iterator.empty())
            {
```

If we are all out of value blocks for the current table, then we need a new table:

```zig
                const key_exclusive_next = iterating.key_exclusive_next;
                if (self.scan.key_min <= key_exclusive_next and
                    key_exclusive_next <= self.scan.key_max)
                {
                    // Load the next `table_info`.
                    self.state = .loading_manifest;
                    self.values = .fetching;
                    self.move_next_manifest_table(key_exclusive_next); // <---
                } else {
                    // The next `table_info` is out of the key range, so it's finished.
                    self.state = .{ .finished = .{} };
                    self.values = .finished;
                }
```

In the code above, `key_exclusive_next` tracks the `table_info.key_max` (ascending) or `table_info.key_min` (descending) of the most recent table we iterated, to ensure that we don't scan the same table twice.
Typically the condition would check whether we are done iterating tables for this level, because the extreme key of the most recent iterated table lies outside of our target range (`scan.key_kin`/`scan.key_max`).

But because the target range has shrunk, our exclusive key may now lie outside of that range from the wrong end, causing the `ScanTreeLevel` to incorrectly consider itself finished.

## Fix

Separate the conditions:

- The `key_exclusive_next` is out of bounds because we iterated all the tables we need to iterate: Stop scanning (same as before).
- The `key_exclusive_next` is out of bounds because our target range shrunk: Keep scanning, but ignore the `key_exclusive_next`.

### Why didn't we find this bug sooner?

There are four fuzzers that perform scans: `vopr`, `fuzz_lsm_scan`, `fuzz_lsm_forest`, and `fuzz_lsm_tree`:

- `fuzz_lsm_tree` only scans a single tree, so it wouldn't ever hit this bug.
- `fuzz_lsm_forest` also only scans a single tree -- it never intersects across multiple trees.
- `vopr` and `fuzz_lsm_scan` do intersect multiple trees, but they generate objects according to templates based on the queries they plan to run. Because of the common prefix for the query's target fields, the objects matching each query are consecutive in each index, so it never needs to "zig-zag". (See below for more detail.)

Also, coverage checking of fuzzing could help find issues like these earlier.
(We track coverage in the devhub, though since we only get coverage for a single seed per fuzzer, it underestimates fuzzer coverage.)

<details>
<summary>Why didn't the VOPR find this bug?</summary>

The generated queries intersect across multiple trees, but never in such a way that `ScanTreeLevel.probe()` is called.

The workload pre-generates its queries at startup:

```zig
for (query_intersections, 1..) |*query_intersection, index| {
    query_intersection.* = .{
        .user_data_64 = @intcast(index * 1_000_000),
        .user_data_32 = @intcast(index * 1_000),
        .code         = @intcast(index), // It will be used to recover the index.
    };
}
```

Each object is generated to match exactly one of these disjoint queries:

```zig
const query_intersection_index = self.random.uintLessThanBiased(usize, self.auditor.query_intersections.len);
const query_intersection = self.auditor.query_intersections[query_intersection_index];

transfer.* = .{
    ...
    .user_data_64 = query_intersection.user_data_64,
    .user_data_32 = query_intersection.user_data_32,
    .code         = query_intersection.code,
    ...
};
```

The `query_intersection` counts how many (created) objects match its filter:

```zig
const query_intersection_index = transfer.code - 1;
const query_intersection = &self.query_intersections[query_intersection_index];
assert(transfer.user_data_64 == query_intersection.user_data_64);
assert(transfer.user_data_32 == query_intersection.user_data_32);
assert(transfer.code == query_intersection.code);
query_intersection.transfers.count += 1;
if (query_intersection.transfers.timestamp_min == 0) {
    query_intersection.transfers.timestamp_min = transfer_timestamp;
}
query_intersection.transfers.timestamp_max = transfer_timestamp;
```

But the queries generated are such that `user_data_64 / 1e6 == user_data_u32 / 1e3 == code`, which means that the objects matching a given query are consecutive in each of those three indexes, so the merge-join never needs to discard values on probe.

</details>

<details>
<summary>Why didn't the scan-fuzzer find this bug?</summary>

`ScanTreeLevel.probe()` does get called, but never such that `ScanTreeLevel.move_next_manifest_table()` gets called.

Similar to the VOPR, the fuzzer carefully chooses the prefix of generated objects based on what query they should match.
So the matching objects of each query are consecutive in the respective indices.
It inserts "noise" objects that don't match any query, but they use `std.math.maxInt(u32)` as a prefix, so they don't interleave with the consecutive runs of target values.

</details>

### Improvements

The scan fuzzer has been rewritten (see https://github.com/tigerbeetle/tigerbeetle/commit/73f13ee2d092fe6998c315b0c4ae5fbc2983154a) and can now easily find this bug.

Additionally, greatly increase the number of queries run and the number of objects inserted per test run.

### DIFF
--- a/src/lsm/scan_merge.zig
+++ b/src/lsm/scan_merge.zig
@@ -1,5 +1,6 @@
 const std = @import("std");
 const assert = std.debug.assert;
+const maybe = stdx.maybe;
 
 const stdx = @import("../stdx.zig");
 const constants = @import("../constants.zig");
@@ -245,13 +246,31 @@ fn ScanMergeType(
                     }
 
                     if (self.merge_iterator) |*merge_iterator| {
-                        // It's not expected to probe a scan that already produced a key equals
-                        // or ahead the probe.
-                        assert(merge_iterator.key_popped == null or
+                        if (merge_iterator.key_popped) |key_popped| {
+                            // The new timestamp may lag behind the merge_iterator's latest key.
+                            //
+                            // Suppose there is a query:
+                            //   (index_1 AND (index_1 OR (index_2 AND index_3)))
+                            // with the listed timestamps in each index:
+                            //
+                            //   zig_zag_merge₁:     [              13, 13     ]
+                            //     tree₀:            [          12, 12, 13     ]
+                            //     k_way_merge₁:     [ 2, 2, 3,     12, 13, 14 ]
+                            //       tree₁:          [ 2, 2, 3,                ]
+                            //       zig_zag_merge₂: [              13, 13, 14 ]
+                            //         tree₂:        [ 2, 2, 3,     12, 13, 14 ]
+                            //         tree₃:        [              13, 13, 14 ]
+                            //
+                            // 1. While peeking the first key from zig_zag_merge₁, we peek 12 from
+                            //    tree₀ and 1 from k_way_merge₁. So we probe k_way_merge₁ with 12.
+                            // 2. k_way_merge₁ relays 11 to its streams (tree₁ + zig_zag_merge₂).
+                            // 3. Probing zig_zag_merge₂ with 12 trips the assert, because tree₃ has
+                            //    already produced a higher key (11 < 12).
                             switch (self.direction) {
-                            .ascending => merge_iterator.key_popped.? < timestamp,
-                            .descending => merge_iterator.key_popped.? > timestamp,
-                        });
+                                .ascending => maybe(key_popped < timestamp),
+                                .descending => maybe(key_popped > timestamp),
+                            }
+                        }
 
                         // Once the underlying streams have been changed, the merge iterator needs
                         // to reset its state, otherwise it may have dirty keys buffered.
--- a/src/lsm/scan_tree.zig
+++ b/src/lsm/scan_tree.zig
@@ -828,6 +828,7 @@ fn ScanTreeLevelType(comptime ScanTree: type, comptime Storage: type) type {
             assert(self.values == .fetching);
 
             assert(self.scan.state == .seeking or
+                self.scan.state == .needs_data or
                 self.scan.state == .buffering);
 
             const manifest: *Manifest = &self.scan.tree.manifest;
--- a/src/lsm/scan_tree.zig
+++ b/src/lsm/scan_tree.zig
@@ -787,17 +787,28 @@ fn ScanTreeLevelType(comptime ScanTree: type, comptime Storage: type) type {
                         // If the next key is out of the range,
                         // there are no more `table_info`s to scan next.
                         const key_exclusive_next = iterating.key_exclusive_next;
-                        if (self.scan.key_min <= key_exclusive_next and
-                            key_exclusive_next <= self.scan.key_max)
-                        {
-                            // Load the next `table_info`.
-                            self.state = .loading_manifest;
-                            self.values = .fetching;
-                            self.move_next_manifest_table(key_exclusive_next);
-                        } else {
+                        if (switch (self.scan.direction) {
+                            .ascending => key_exclusive_next > self.scan.key_max,
+                            .descending => key_exclusive_next < self.scan.key_min,
+                        }) {
                             // The next `table_info` is out of the key range, so it's finished.
                             self.state = .{ .finished = .{} };
                             self.values = .finished;
+                        } else {
+                            // Load the next `table_info`.
+                            self.state = .loading_manifest;
+                            self.values = .fetching;
+                            if (switch (self.scan.direction) {
+                                .ascending => key_exclusive_next < self.scan.key_min,
+                                .descending => key_exclusive_next > self.scan.key_max,
+                            }) {
+                                // A probe() skipped past the last table we iterated, so our
+                                // key_exclusive_next is now out of bounds, superseded by the
+                                // tightened key_min (ascending) or key_max (descending) bound.
+                                self.move_next_manifest_table(null);
+                            } else {
+                                self.move_next_manifest_table(key_exclusive_next);
+                            }
                         }
                     } else {
                         // Keep iterating to the next `value_block`.
--- a/src/state_machine.zig
+++ b/src/state_machine.zig
@@ -917,6 +917,11 @@ pub fn StateMachineType(
             assert(self.scan_lookup_result_count == null);
             assert(self.forest.scan_buffer_pool.scan_buffer_used == 0);
 
+            log.debug("{?}: get_account_transfers: {}", .{
+                self.forest.grid.superblock.replica_index,
+                filter,
+            });
+
             if (self.get_scan_from_account_filter(filter)) |scan| {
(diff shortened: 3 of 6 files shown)


src/lsm/scan_tree.zig
```diff
@@ -817,6 +828,7 @@ fn ScanTreeLevelType(comptime ScanTree: type, comptime Storage: type) type {
             assert(self.values == .fetching);
 
             assert(self.scan.state == .seeking or
+                self.scan.state == .needs_data or
                 self.scan.state == .buffering);
```
sentientwaffle, line 832: In `zig_zag_merge.zig`:

    for (0..it.streams_count) |stream_index| {
        if (drained.isSet(stream_index)) {
            // Probing the drained stream will update the key range for the next read.
            stream_probe(it.context, @intcast(stream_index), probe_key);
            // The stream must remain drained after probed.
            assert(stream_peek(it.context, @intcast(stream_index)) == error.Drained);
        } else {
            ...
        }
    }

...we are probing drained streams, which are precisely the streams with `needs_data`.

src/lsm/scan_merge.zig
```diff
@@ -245,14 +245,6 @@ fn ScanMergeType(
                     }
 
                     if (self.merge_iterator) |*merge_iterator| {
-                        // It's not expected to probe a scan that already produced a key equals
-                        // or ahead the probe.
-                        assert(merge_iterator.key_popped == null or
-                            switch (self.direction) {
-                            .ascending => merge_iterator.key_popped.? < timestamp,
-                            .descending => merge_iterator.key_popped.? > timestamp,
-                        });
```
sentientwaffle: This assert is hit by nested alternating intersection/unions.

Suppose there is a query `(index_0 AND (index_1 OR (index_2 AND index_3)))` with the listed timestamps in each index:

    zig_zag_merge₁:     [              12, 13     ]
      tree₀:            [          11, 12, 13     ]
      k_way_merge₁:     [ 1, 2, 3,     12, 13, 14 ]
        tree₁:          [ 1, 2, 3,                ]
        zig_zag_merge₂: [              12, 13, 14 ]
          tree₂:        [ 1, 2, 3,     12, 13, 14 ]
          tree₃:        [              12, 13, 14 ]

1. While peeking the first key from `zig_zag_merge₁`, we peek `11` from `tree₀` and `1` from `k_way_merge₁`.
2. So we probe `k_way_merge₁` with `11`. `k_way_merge₁` relays `11` to its streams (`tree₁` + `zig_zag_merge₂`).
3. Probing `zig_zag_merge₂` with `11` trips the assert, because `tree₃` has already produced a higher key (`11 < 12`).

(Note that we can't simply gate this assert on `merge == .merge_intersection`, since the inner intersection still fails.)

src/state_machine/auditor.zig
```diff
@@ -205,8 +205,8 @@ pub const AccountingAuditor = struct {
         errdefer allocator.free(accounts_state);
         @memset(accounts_state, AccountState{});
 
-        // The number of known intersection values ​​for the secondary indices is kept
-        // low enough to explore different cardinalities.
```
sentientwaffle, line 209: The wrapping here is incidental -- the first line was hiding a couple of unicode zero-width spaces. (`git grep "$(printf %b '\u200b')"`)

sentientwaffle:
I'm leaving auto-merge disabled to give this a chance to bake in the CFO for the weekend.

src/state_machine.zig
```diff
@@ -1013,16 +1023,23 @@ pub fn StateMachineType(
...
+                    log.info(
+                        "get_account_balances: cannot query account.id={}; flags.history=false",
+                        .{filter.account_id},
+                    );
                 }
+            } else {
+                log.info(
+                    "get_account_balances: cannot query account.id={}; account does not exist",
+                    .{filter.account_id},
+                );
```
batiati, line 1040: :heart:

src/lsm/scan_merge.zig
batiati: I think this example is so great that it must be converted to a code comment instead :smiley:

batiati (commented):
Great write-up @sentientwaffle!

There's one failing seed in the last commit, but I couldn't reproduce it (still it's worth checking).

https://devhub.tigerbeetle.com/?fuzzer=lsm_scan&commit=099c8d48d6fcf7275ddc69f9e39d64d14c553d2c

sentientwaffle:
> I think this example is so great that it must be converted to a code comment instead 😃

Updated! I included it in the same commit as that fix.

> There's one failing seed in the last commit, but I couldn't reproduce it (still it's worth checking).

It is a cfo false positive -- the seed passes locally on the same commit.

There were also several failures on an earlier commit prior to [shrinking events_max](https://github.com/tigerbeetle/tigerbeetle/pull/2544/commits/8fa9f6e15f2c342e3e7cb030f044c9986e5b383f) due to hitting the cfo timeout (i.e. the fuzzer took longer than 30 minutes to run). But those seeds pass locally as well, if run without a timeout.

batiati (approved):

PR #2550, opened 2024-12-09
node: fix panic in node client on eviction or shutdown

### LINKED ISSUE #2547: Node Client 0.16.16 'panic: attempt to use null value'
Hi there,

We're seeing an issue approximately once a day where the node client will panic and exit, also causing the node process to crash.
```
thread 25 panic: attempt to use null value
/home/runner/work/tigerbeetle/tigerbeetle/src/clients/node/node.zig:287:27: 0xffff800910db in on_completion (tb_nodeclient)
???:?:?: 0x1625651f in ??? (???)
Unwind information for `???:0x1625651f` was not available, trace may be incomplete
```
Unfortunately we don't seem to be able to get more information on the crash into our monitoring.
We're currently on 0.16.16 but have now noticed that this was occurring with similar frequency on 0.16.11 and potentially earlier. Node version is 22.11

cb22:
Inside `node.zig`, `.client_evicted` and `.client_shutdown` packet errors fall through to the general code, with the idea that those errors can be bubbled up as exceptions via `on_completion_js`.

However, when the packet has an error, `result_ptr` is null, so we hit a null pointer deference on `result_ptr.?[..]`. To fix it, shift the handling for a successful packet inside the switch, since `on_completion_js` already has the logic for freeing the packet and handling the exception.

Fixes #2547.

### DIFF
--- a/src/clients/node/node.zig
+++ b/src/clients/node/node.zig
@@ -266,34 +266,37 @@ fn on_completion(
     _ = timestamp;
 
     switch (packet.status) {
-        .ok => {},
+        .ok => {
+            switch (@as(Operation, @enumFromInt(packet.operation))) {
+                inline else => |op| {
+                    const event_count = @divExact(
+                        packet.data_size,
+                        @sizeOf(StateMachine.EventType(op)),
+                    );
+                    const buffer: BufferType(op) = .{
+                        .ptr = @ptrCast(packet),
+                        .count = event_count,
+                    };
+
+                    const Result = StateMachine.ResultType(op);
+                    const results: []const Result = @alignCast(std.mem.bytesAsSlice(
+                        Result,
+                        result_ptr.?[0..result_len],
+                    ));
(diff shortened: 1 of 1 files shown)


src/clients/node/node.zig
```diff
@@ -266,36 +266,39 @@ fn on_completion(
...
+                    @memcpy(buffer.results()[0..results.len], results);
+
+                    packet.data_size = @bitCast(BufferSize{
+                        .event_count = @intCast(event_count),
+                        .result_count = @intCast(results.len),
+                    });
+                },
+                .pulse => unreachable,
+            }
+        },
```
matklad, line 295: It isn't immediately obvious that `defer buffer.free();` remains correct in `on_completion_js`, but I think it does. We don't touch `event_count` part of `packet.data_size`, and that tells us what we should be freeing.

matklad (approved):

PR #2553, opened 2024-12-09
[Benchmark] Forward the `--experimental` flag when setting the object cache.

batiati:
Error when running the benchmark with any of the `--cache-*` options.
```
❯ ./tigerbeetle benchmark --cache-transfers=2GiB
2024-12-09 20:03:47.713Z error(vsr): --cache-transfers is marked experimental, add `--experimental` to continue.
thread 588607 panic: integer overflow
```

### DIFF
--- a/src/tigerbeetle/benchmark_driver.zig
+++ b/src/tigerbeetle/benchmark_driver.zig
@@ -59,8 +59,19 @@ pub fn main(allocator: std.mem.Allocator, args: *const cli.Command.Benchmark) !v
             .args = args,
         });
     } else {
-        if (args.trace) |_| {
-            vsr.fatal(.cli, "--trace: incompatible with --addresses", .{});
+        // Arguments forwarded to the replica cannot be used with a cluster started by the user.
+        inline for (.{
+            "cache_accounts",
+            "cache_transfers",
+            "cache_transfers_pending",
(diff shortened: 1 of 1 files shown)


cb22:
Ah, perhaps we should rather allow `--experimental` to be specified and passed through?

batiati:
> Ah, perhaps we should rather allow `--experimental` to be specified and passed through?

For context, initially, we had considered forwarding **all** command line options that are not from the benchmark to the TB process. But since we run a curated instance of TB, not all arguments make sense as being user-defined. Following this logic, we decided the benchmark should accept and forward only a subset of parameters that can affect the benchmark result; otherwise, the user should start a TB cluster by themselves and use the benchmark with the `--addresses` argument.

That said, for consistency, I don't think we should expose the `--experimental` flag as a benchmark thing, but I don't have a super strong opinion on that.

While writing this comment, I just realized that we also need to change the code that validates the incompatibility of forwarded arguments with clusters started by the user!

https://github.com/tigerbeetle/tigerbeetle/blob/c7145e6f382c3a4b5722fc16ff9a933ee27a0c0c/src/tigerbeetle/benchmark_driver.zig#L62-L64

cb22 (approved):

PR #2554, opened 2024-12-10
stdx: move memory lock to stdx from main

matklad:
Two motivations here:

- `pub fn main` is the "face" of our source code, and it is a bad style if we have a bunch of fiddly platform-specific code getting in a way of understanding what _actually_ goes on in our main.
- I want to re-use some of the logic to lock memory used by fuzzers

### DIFF
--- a/src/stdx/mlock.zig
+++ b/src/stdx/mlock.zig
@@ -1,0 +1,112 @@
+const builtin = @import("builtin");
+const std = @import("std");
+const os = std.os;
+
+const stdx = @import("../stdx.zig");
+
+const log = std.log.scoped(.mlock);
+
+const MemoryLockError = error{memory_not_locked} || std.posix.UnexpectedError;
+
+const mlockall_error = "Unable to lock pages in memory ({s})" ++
+    " - kernel swap would otherwise bypass TigerBeetle's storage fault tolerance. ";
+
+/// Pin virtual memory pages allocated so far to physical pages in RAM, preventing the pages from
+/// being swapped out and introducing storage error into memory, bypassing ECC RAM.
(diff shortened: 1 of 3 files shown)


src/stdx/mlock.zig
```diff
@@ -0,0 +1,112 @@
...
+    var working_set_max: os.windows.SIZE_T = 0;
+
+    if (get_process_working_set_size(
+        process_handle,
+        &working_set_min,
+        &working_set_max,
+    ) == os.windows.FALSE) {
+        working_set_min = allocated_size; // Count bytes allocated so far.
+        working_set_min += 64 * 1024 * 1024; // 64mb buffer room for stack/globals.
+        working_set_max = working_set_min * 2; // Buffer room for new page faults.
```
matklad, line 85: This feels suspicious --- does this doubles our memory size? And also, it feels rather wrong --- we _could_ have large fragmentation here, the relation between the size allocated and the number of pages we need is non-trivial...

But out of scope for this.

cb22:
@matklad needs a zig fmt on mlock.zig!

matklad:
Done!

cb22:
@matklad `src/stdx/mlock.zig:43:13: error: expected type 'error{Unexpected,memory_not_locked}!void', found '@TypeOf(.enum_literal)'` :smile:

src/stdx/mlock.zig
cb22: I don't believe so - https://learn.microsoft.com/en-us/windows/win32/api/memoryapi/nf-memoryapi-setprocessworkingsetsize - my understanding is Windows will keep at least the lower bound, to a max of the upper bound.

cb22 (approved):

PR #2556, opened 2024-12-10
tidy: remove unused imports

matklad:
I meant to do this when implementing tidy check for dead code, but the initial implementation missed the `const foo = std.foo` case (it considered it a usage of `foo`), so we actually missed a lot of unused things! 

I am _somewhat_ apprehensive about removing "I will need it later" things like `const log =` and `const assert =`, but, I think, on balance, it's better to stick to simple rule about no unused stuff.

### DIFF
--- a/src/clients/java/src/client.zig
+++ b/src/clients/java/src/client.zig
@@ -13,7 +13,8 @@ const builtin = @import("builtin");
 const jni = @import("jni.zig");
 
 // When referenced from unit_test.zig, there is no vsr import module. So use relative path instead.
-pub const vsr = if (builtin.is_test) @import("../../../vsr.zig") else @import("vsr");
+pub const vsr =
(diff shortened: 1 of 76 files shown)


src/clients/java/src/client.zig
```diff
@@ -13,7 +13,8 @@ const builtin = @import("builtin");
 const jni = @import("jni.zig");
 
 // When referenced from unit_test.zig, there is no vsr import module. So use relative path instead.
-pub const vsr = if (builtin.is_test) @import("../../../vsr.zig") else @import("vsr");
+pub const vsr =
+    if (builtin.is_test) @import("../../../vsr.zig") else @import("vsr");
```
matklad, line 17: Just reformatting, which I semi-accidentally left in: in one of the early attempts, this would have been a false negative, so I added a new line. It's not longer necessary with the final iml, but code reads a bit better like this I think.

sentientwaffle (approved):

PR #2557, opened 2024-12-10
VSR: Improve `prepare_timeout`/`replicate` replica selection

sentientwaffle:
# Bug

Consider the following scenario:

1. `view=2`, so `R2` is primary.
2. `R3` is killed.
3. `R2`: Primary receives a request.
4. `R2`: Primary replicates request to `R3`.
5. `R2`: `prepare_timeout` fires after 500ms, and we replicate to `R1`.
6. `R1`: Ack, but don't replicate, since the next node in the ring is `R2`, which is primary. (So `R1` assumes replication is done).
7. `R2`: After another >500ms (longer due to backoff), `prepare_timeout` fires again, this time replicating to `R4`.
8. `R4`: Ack, replicate to `R0`.
9. `R2`: Commit + reply.

So the cluster isn't deadlocked, just slow.

Steps 3-9 repeat for each prepare.

Due to how `prepare_timeout` selects its target replica, for this particular `view`/`replica_count` combination, `R2` always picks `R1` as its first retry.

# Fix

A complete solution for efficient ring replication in compromised clusters will involve tracking the health of all replicas.

But in the mean time, mitigate the issue by:
- In `prepare_timeout`, leapfrog the (most likely down) replica for the first retry. (That is, in the above scenario, the first retry would go to `R4` instead of `R3`).
- In `replicate`/`prepare_timeout`, use a counter-clockwise replication ring for odd ops.
- Reduce `prepare_timeout`'s timeout from 500ms to 250ms.

### DIFF
--- a/src/vsr/replica.zig
+++ b/src/vsr/replica.zig
@@ -2945,13 +2945,17 @@ pub fn ReplicaType(
                 return;
             }
 
+            // See replicate().
+            const ring_direction: i16 = if (prepare.message.header.op % 2 == 0) 1 else -1;
+
             // The list of remote replicas yet to send a prepare_ok:
             var waiting: [constants.replicas_max]u8 = undefined;
             var waiting_len: usize = 0;
             for (1..self.replica_count) |ring_index| {
-                const replica: u8 = @intCast(
-                    (@as(usize, self.replica) + ring_index) % self.replica_count,
-                );
+                const replica: u8 = @intCast(@mod(
+                    @as(i16, self.replica) + ring_direction * @as(i16, @intCast(ring_index)),
+                    self.replica_count,
+                ));
                 assert(replica != self.replica);
                 if (!prepare.ok_from_all_replicas.isSet(replica)) {
                     waiting[waiting_len] = replica;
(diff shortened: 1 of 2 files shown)


src/vsr/replica.zig
```diff
@@ -2945,28 +2945,22 @@ pub fn ReplicaType(
                 return;
             }
 
+            // Use the opposite direction from `replicate()`.
+            const ring_direction: i16 = if (prepare.message.header.op % 2 == 0) -1 else 1;
```
matklad: Hm, I think we want the _same_ direction here? You want to jump over the faulty replica to the first healthy one, to resume the ring. If you jump to the last replica, the ring will remain unclosed? 

It's only for _next_ op that it makes sense to switch direction: 

- backups switch direction as well, so it _will_ get replicated across the entire ring
- and it will additionally inform backups about the previous prepare.

matklad: If I am missing the reason why we do the reverse here, would be good to add that "why" to the comment!

sentientwaffle: :facepalm: Ah you're right -- and I actually changed it back to the same direction yesterday but forgot to force-push!

matklad (approved):

PR #2552, opened 2024-12-09
Client: Distinguish ClientReleaseToo{Low,High} errors

sentientwaffle:
Client eviction can have one of several causes.

Two of the most likely ones for a user to hit by accident are:
- `client_release_too_low`: The client's release is below the cluster's `client_release_min` -- upgrade the client.
- `client_release_too_high`: The client's release is above the cluster's current release -- downgrade the client (or upgrade the cluster).

(This is a follow-up for https://github.com/tigerbeetle/tigerbeetle/pull/2484).

PR #2558, opened 2024-12-11
misc: use --save-exact for node and better log on eviction

cb22:
Without `--save-exact`, node will record the package as eg `^0.16.16` which will bump patch versions. This is incompatible with how we handle versions between clients and replicas.

Additionally, note that down in the release notes, and include Python there which was missing.

Lastly, when logging the eviction, include the client release and hint on what to do, if it was due to mismatched versions.

### DIFF
--- a/src/vsr/client.zig
+++ b/src/vsr/client.zig
@@ -395,11 +395,23 @@ pub fn ClientType(comptime StateMachine_: type, comptime MessageBus: type) type 
             assert(eviction.header.view >= self.view);
 
             if (self.on_eviction_callback) |callback| {
-                log.err("{}: session evicted: reason={?s} (cluster_release={})", .{
-                    self.id,
-                    std.enums.tagName(vsr.Header.Eviction.Reason, eviction.header.reason),
-                    eviction.header.release,
-                });
+                const eviction_specific_log = switch (eviction.header.reason) {
+                    .client_release_too_low => " - your client is too old; upgrade to a version " ++
+                        "compatible with your cluster",
+                    .client_release_too_high => " - your client is too new; downgrade to the " ++
+                        "same version as your cluster",
+                    else => "",
+                };
+                log.err(
+                    "{}: session evicted: reason={?s} (cluster_release={}, client_release={}){s}",
+                    .{
+                        self.id,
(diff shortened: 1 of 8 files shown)


src/vsr/client.zig
```diff
@@ -395,11 +395,23 @@ pub fn ClientType(comptime StateMachine_: type, comptime MessageBus: type) type
...
-                log.err("{}: session evicted: reason={?s} (cluster_release={})", .{
-                    self.id,
-                    std.enums.tagName(vsr.Header.Eviction.Reason, eviction.header.reason),
-                    eviction.header.release,
-                });
+                const eviction_specific_log = switch (eviction.header.reason) {
+                    .client_release_too_low => " your client is too old; upgrade to a version " ++
+                        "compatible with your cluster",
+                    .client_release_too_high => " your client is too new; downgrade to the same " ++
+                        "version as your cluster",
```
matklad:
```suggestion
                    .client_release_too_low => " Your client is too old; upgrade to a version " ++
                        "compatible with your cluster",
                    .client_release_too_high => " Your client is too new; downgrade to the same " ++
                        "version as your cluster",
```

maybe? It's after `.`

cb22: It should have a trailing full stop then too, but it feels... weird to have mixed capitalization like that.

I was really looking for a way to distinguish the information from the suggestion. Think I'll use a `-` instead.

src/vsr/client.zig
```diff
@@ -395,11 +395,23 @@ pub fn ClientType(comptime StateMachine_: type, comptime MessageBus: type) type
...
+                log.err(
+                    "{}: session evicted: reason={?s} (cluster_release={}, client_release={}){s}",
+                    .{
+                        self.id,
+                        std.enums.tagName(vsr.Header.Eviction.Reason, eviction.header.reason),
+                        eviction.header.release,
+                        self.release,
+                        eviction_specific_log,
+                    },
+                );
```
sentientwaffle, line 414: Related: https://github.com/tigerbeetle/tigerbeetle/pull/2552

sentientwaffle (approved):

PR #2559, opened 2024-12-11
Vortex: parameterize output dir and don't delete it

owickstrom:
Previously the output directory was always in `.zig-cache/tmp` and it was also deleted in a defer statement. So if a test failed, the datafiles were gone. Not great.

Now the output directory is always kept. Additionally, the `--output-directory=<PATH>` option can be used to not generate one in `.zig-cache`.

I'll create another PR (if I have time!) to redirect all logs into files in the output directory too.

### DIFF
--- a/src/testing/vortex/supervisor.zig
+++ b/src/testing/vortex/supervisor.zig
@@ -104,6 +104,7 @@ pub const CLIArgs = struct {
     test_duration_minutes: u16 = 10,
     driver_command: ?[]const u8 = null,
     disable_faults: bool = false,
+    output_directory: ?[]const u8 = null,
     log_debug: bool = false,
 };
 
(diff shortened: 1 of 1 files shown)


src/testing/vortex/supervisor.zig
```diff
@@ -678,18 +676,18 @@ pub fn comma_separate_ports(comptime ports: []const u16) []const u8 {
     return out;
 }
 
-// Create a new Vortex-specific temporary directory. Caller owns the returned memory.
-fn create_tmp_dir(allocator: std.mem.Allocator) ![]const u8 {
+// Create a new Vortex-specific temporary directory.
+fn create_tmp_dir(absolute_path_buffer: []u8) ![]const u8 {
```
matklad, line 697: <3, good to get rid of an allocator here!

matklad (approved):

src/testing/vortex/supervisor.zig
owickstrom: Couldn't resist!

matklad (dismissed):

matklad:
tidy objects!

cb22 (approved):

PR #2560, opened 2024-12-11
devhub: run under sudo to capture memlock rss

cb22:


matklad:
I wonder if this might make git complain about "dubious ownership", but def worth a try!

matklad (approved):

PR #2561, opened 2024-12-12
[Java] Add tests asserting that the client can handle thread interruption.

batiati:
The Java client exposes _blocking_ methods (i.e., `createTransfers`, `createAccounts`, `lookupTransfers`, `lookupAccounts`, etc), which might be waiting while the running thread is interrupted. This PR adds unit tests to assert that the thread is correctly interrupted and no result is produced.

Currently, since we don't support cancelations, we assume the thread will never be interrupted by ourselves, swallowing the `InterruptedException` and throwing an `AssertionError` instead.
After reviewing some best practices [here](https://stackoverflow.com/questions/10868255/when-should-a-method-throw-interruptedexception-and-how-should-i-handle-one-tha) and [here](https://nurkiewicz.com/2014/05/interruptedexception-and-interrupting.html), I think we _should_ add `throws InterruptedException` to those blocking methods. The `AssertionError` isn't the right exception to signal that the operation could not be finished due to a `thread.interrupt()` provoked by the caller (however, the API using those methods will be broken).

### DIFF
--- a/src/clients/java/src/test/java/com/tigerbeetle/IntegrationTest.java
+++ b/src/clients/java/src/test/java/com/tigerbeetle/IntegrationTest.java
@@ -1366,6 +1366,41 @@ public class IntegrationTest {
         }
     }
 
+    /**
+     * This test asserts that the client can handle thread interruption.
+     */
+    @Test
+    public void testConcurrentInterruptedTasks() throws Throwable {
+        final int TASKS_COUNT = 256;
+        final var barrier = new CountDownLatch(TASKS_COUNT);
+        final var zeroedId = UInt128.asBytes(0L);
+
+        // Connect to an invalid cluster, so the calls never complete.
+        try (final var client = new Client(clusterId, new String[] {"0"})) {
+            final var tasks = new TransferTask[TASKS_COUNT];
+            for (int i = 0; i < TASKS_COUNT; i++) {
+                // Starting multiple threads.
+                tasks[i] = new TransferTask(client, zeroedId, zeroedId, TransferFlags.NONE, barrier,
+                        new CountDownLatch(0));
+                tasks[i].start();
+            }
+
+            // Waits until all threads start.
+            barrier.await();
+
+            // Interrupt all threads.
+            for (int i = 0; i < TASKS_COUNT; i++) {
+                tasks[i].interrupt();
+                tasks[i].join();
+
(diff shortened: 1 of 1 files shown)


matklad:
>After reviewing some best practices [here](https://stackoverflow.com/questions/10868255/when-should-a-method-throw-interruptedexception-and-how-should-i-handle-one-tha) and [here](https://nurkiewicz.com/2014/05/interruptedexception-and-interrupting.html), I think we should add throws InterruptedException to those blocking methods. The AssertionError isn't the right exception to signal that the operation could not be finished due to a thread.interrupt() provoked by the caller


Agree here!

src/clients/java/src/test/java/com/tigerbeetle/IntegrationTest.java
```diff
@@ -1366,6 +1366,41 @@ public void testAsyncTasks() throws Throwable {
...
+     * This test asserts that the client can handle thread interruption.
+     */
+    @Test
+    public void testConcurrentInterruptedTasks() throws Throwable {
+        final int TASKS_COUNT = 256;
+        final var barrier = new CountDownLatch(TASKS_COUNT);
+        final var zeroedId = UInt128.asBytes(0L);
+
+        // Connect to an invalid cluster, so the calls never complete.
+        try (final var client = new Client(clusterId, new String[] {"0"})) {
```
matklad, line 1379: Could you also manually smoke-test this with a real cluster, to make sure that interruption is correctly handled at any stage of request lifecycle? I don't think we  need a test here (though it won't harm), but manual smoke test feels in order!

matklad (approved):

src/clients/java/src/test/java/com/tigerbeetle/IntegrationTest.java
batiati: Done! It does work fine as well.

It's not worth adding a test because the thread interruption is checked only once when waiting for the reply, and by using an invalid cluster, we can guarantee that all threads will be stuck in the same point waiting (so the current test maximizes the coverage).

Another blocking method I considered adding tests is `client.close()`, but since all the blocking logic is done in native code outside the JVM, we never check for the thread interruption flag.
The cleaning logic must be uninterruptible, so `thread.interrupt()` does not affect a thread running `client.close()` as it always runs to completion.

PR #2562, opened 2024-12-12
devhub: pass env through sudo

cb22:


matklad (approved):

.github/workflows/ci.yml
```diff
@@ -193,7 +193,7 @@ jobs:
       - run: sudo apt install kcov
       - run: ./zig/download.sh
       # Run under sudo to enable memory locking for accurate RSS stats.
-      - run: sudo ./zig/zig build scripts -- devhub --sha=${{ github.sha }}
+      - run: sudo -E ./zig/zig build scripts -- devhub --sha=${{ github.sha }}
```
matklad, line 196: As a rule of thumb, in written scripts, prefer long-form of the arguments, like `sure --preserve-env`. That way they are usually a bit more self-explanatory for the readers. No big deal though!

PR #2563, opened 2024-12-12
vsr: pare op_repair_min back to 90

matklad:


matklad:
I must confess that I've lost all the context around this one!

PR #2565, opened 2024-12-12
VSR: Revert reverse ring replication

sentientwaffle:
This reverts commit da50a79cc0f527fc91a1bcadca6f2c431e2aa059.

I think this has some unexpected consequences around header repair which might make it counterproductive. I am investigating more, but don't want to release it in the mean time.


matklad (approved):
🤔

PR #2567, opened 2024-12-13
misc: add random_int_nonzero

cb22:
Quite a few places (and some that forget!) use the pattern of generating a random u128 then checking it's non-zero, in case there's a broken CSPRNG.

Factor that out into a function and use it everywhere `std.crypto.random.int(u128)` was used directly before.

### DIFF
--- a/src/stdx.zig
+++ b/src/stdx.zig
@@ -940,3 +940,13 @@ pub fn unexpected_errno(label: []const u8, err: std.posix.system.E) std.posix.Un
     }
     return error.Unexpected;
 }
+
+pub fn unique_u128() u128 {
+    const value = std.crypto.random.int(u128);
+
+    // Broken CSPRNG is the likeliest explanation for zero or all ones.
+    assert(value != 0);
+    assert(value != std.math.maxInt(u128));
+
+    return value;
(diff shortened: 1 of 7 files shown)


src/stdx.zig
```diff
@@ -940,3 +940,12 @@ pub fn unexpected_errno(label: []const u8, err: std.posix.system.E) std.posix.Un
...
 }
+
+pub fn random_int_nonzero(comptime T: type) T {
+    const value = std.crypto.random.int(T);
+
+    // Broken CSPRNG is the likeliest explanation for zero.
+    assert(value != 0);
+
+    return value;
+}
```
matklad: Let's maybe make it

```suggestion
pub fn unique_u128() u128 {
    const value = std.crypto.random.int(u128);

    // Broken CSPRNG is the likeliest explanation for zero.
    assert(value != 0);
    assert(value != ~0);

    return value;
}
```

?

I think `unique` captures what we want here more precisely

matklad: And we need to prevent calling this function with `u8`, which would be wrong!

matklad (approved):

PR #2568, opened 2024-12-13
devhub: track ci pipeline duration on the devhub

cb22:
Specifically, this tracks the merge_group duration for a specific commit, which is the time said commit spent in the merge queue running CI.

Since the tip of the merge queue becomes the tip on `main`, and this *must* happen before the devhub runs, we can reliably extract this info and publish it to our metrics.

### DIFF
--- a/src/scripts/devhub.zig
+++ b/src/scripts/devhub.zig
@@ -129,6 +129,21 @@ fn devhub_metrics(shell: *Shell, cli_args: CLIArgs) !void {
         "us",
     );
 
+    const ci_pipeline_duration_s = blk: {
+        const times_gh = try shell.exec_stdout("gh run list -c {sha} -e merge_group " ++
+            "--json startedAt,updatedAt -L 1 --template {template}", .{
+            .sha = cli_args.sha,
+            .template = "{{range .}}{{.startedAt}} {{.updatedAt}}{{end}}",
+        });
+        const iso8601_started_at, const iso8601_updated_at =
+            (stdx.cut(times_gh, " ") orelse break :blk null).unpack();
+
+        const epoch_started_at = try shell.iso8601_to_timestamp_seconds(iso8601_started_at);
+        const epoch_updated_at = try shell.iso8601_to_timestamp_seconds(iso8601_updated_at);
+
+        break :blk epoch_updated_at - epoch_started_at;
+    };
+
     const batch = MetricBatch{
         .timestamp = commit_timestamp,
         .attributes = .{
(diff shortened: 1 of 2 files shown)


src/scripts/devhub.zig
```diff
@@ -129,6 +129,28 @@ fn devhub_metrics(shell: *Shell, cli_args: CLIArgs) !void {
...
+        const iso8601_updated_at = times_iterator.next() orelse break :blk null;
+
+        const epoch_started_at = try std.fmt.parseInt(u64, try shell.exec_stdout(
+            "date -d {iso8601_started_at} +%s",
+            .{ .iso8601_started_at = iso8601_started_at },
+        ), 10);
+        const epoch_updated_at = try std.fmt.parseInt(u64, try shell.exec_stdout(
+            "date -d {iso8601_updated_at} +%s",
+            .{ .iso8601_updated_at = iso8601_updated_at },
+        ), 10);
```
cb22: I'll admit this code is a bit suspect, calling `date` to do ISO8601 to epoch conversion, but I didn't particular fancy writing a parser for it then dealing with `epoch.zig` :sweat_smile:.

matklad:
@cb22 you are on fire lately, I am absolutely in love with like half of your changes you made in the last two weeks!

src/scripts/devhub.zig
matklad: SGTM, but let's extract this into:

```zig
fn iso8601_to_timestamp_seconds(shell: *Shell, datetime: []const u8) !u64
```

so that it would be easier to write a parser in the future (I bet a sufficiently smart LLM can do it)

src/scripts/devhub.zig
```diff
@@ -129,6 +129,28 @@ fn devhub_metrics(shell: *Shell, cli_args: CLIArgs) !void {
...
 
+    const ci_pipeline_duration_s = blk: {
+        const times_gh = try shell.exec_stdout("gh run list -c {sha} -e merge_group " ++
+            "--json startedAt,updatedAt -L 1 --template {template}", .{
+            .sha = cli_args.sha,
+            .template = "{{range .}}{{.startedAt}} {{.updatedAt}}{{end}}",
+        });
+        var times_iterator = std.mem.split(u8, times_gh, " ");
+        const iso8601_started_at = times_iterator.next() orelse break :blk null;
+        const iso8601_updated_at = times_iterator.next() orelse break :blk null;
```
matklad:
```
        const iso8601_started_at, const iso8601_updated_at =
            (stdx.cut(times_gh, " ") orelse break :blk null).unpack();
```

src/scripts/devhub.zig
```diff
@@ -150,6 +172,7 @@ fn devhub_metrics(shell: *Shell, cli_args: CLIArgs) !void {
                 .value = checksum_message_size_max_us,
                 .unit = "us",
             },
+            .{ .name = "ci pipeline duration", .value = ci_pipeline_duration_s.?, .unit = "s" },
```
matklad: Let's swap it with the build one, the build was always ment to be a proxy for CI. I _think_ the order of graph will be determined by order in this slice.

src/scripts/devhub.zig
cb22: I always forget about our `stdx.*` utils, and that Zig now has destructuring. (perhaps suppressing JS memories...)

src/scripts/devhub.zig
cb22: By swap, I assume you meant keep both and swap the order - let me know if otherwise. I think it might be interesting to track debug build times too, but I'll look at that in a follow up.

matklad: Yeah, keep both, swap the order!

matklad (approved):

cb22:
(holding off merging this until after #2574)

PR #2569, opened 2024-12-13
vsr: remove dead code

matklad:
I _think_ this entire block became dead code after https://github.com/tigerbeetle/tigerbeetle/pull/1552, as now view headers are guaranteed to include an op from our log.

### DIFF
--- a/src/vsr/replica.zig
+++ b/src/vsr/replica.zig
@@ -708,27 +708,11 @@ pub fn ReplicaType(
                     }
                     break;
                 }
-            } else {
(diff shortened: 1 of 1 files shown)


matklad:
No, this is very much not dead, see the code around

```
                        // Transition from normal status, but the SV headers were part of the next
                        // wrap, so we didn't install them to our journal, and we didn't catch up.
                        // We will reuse the SV headers as our DVC headers to ensure that
                        // participating in another view-change won't allow the op to backtrack.
```

PR #2571, opened 2024-12-16
vsr: don't crash if a round of view changes happened while repairing …

matklad:
…pipeline

This fixes availability issue in VSR, where the prospective primary could crash during view change when a journal read initiated in the previous view change finishes. This is unlikely to be hit in practice, but was discovered by VOPR.

 # Scenario

The prospective primary needs to fully repair before starting a new view. The final part of the repair is the pipeline --- the prospective primary needs to make sure that all potentially committed ops are in memory. Specifically, they are to be stored in PipelineCache, which is to be flipped over to PipelineQueue proper when the view is started.

The prospective primary _gets_ the prepares using the normal repair protocol, but it could be the case that a prepare only exists on disk, and not in memory.

Fetching prepares from disk is exactly the job of
`primary_repair_pipeline` function.

Reading prepares is asynchronous, so it could be the case that, by the time the read finishes, it is no longer relevant. In the simple case, the view is advanced by one and the replica is no longer future primary.

The more complex case is when a view is advanced by a multiple of `replica_count`, so the replica is a prospective primary _again_.

In general, we can't say if the read we are dealing with was initiated for _this_ view change, but we don't need to. It's fine if the older reads _help_ a future view change, as long as any reads which are not helpful are ignored.

This latter case was buggy, we could have proceeded with repair_pipeline_callback even though we wouldn't have called repair_pipeline from repair yet.

This commits adds the missing condition of commit_min == commit_max

 # Why wasn't this discovered before

This is a pre-existing bug. I _think_ it was brought to light through pauses. CEO caught it where cfo missed :-)

Seed: ./zig/zig build vopr -- --lite 9406017288286641868

### DIFF
--- a/src/vsr/replica.zig
+++ b/src/vsr/replica.zig
@@ -6776,6 +6776,7 @@ pub fn ReplicaType(
         fn primary_repair_pipeline(self: *Replica) enum { done, busy } {
             assert(self.status == .view_change);
             assert(self.primary_index(self.view) == self.replica);
+            assert(self.commit_stage == .idle);
             assert(self.commit_max == self.commit_min);
             assert(self.commit_max <= self.op);
             assert(self.pipeline == .cache);
@@ -6841,6 +6842,7 @@ pub fn ReplicaType(
         fn primary_repair_pipeline_op(self: *const Replica) ?u64 {
             assert(self.status == .view_change);
             assert(self.primary_index(self.view) == self.replica);
+            assert(self.commit_stage == .idle);
             assert(self.commit_max == self.commit_min);
             assert(self.commit_max <= self.op);
             assert(self.pipeline == .cache);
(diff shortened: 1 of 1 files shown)


src/vsr/replica.zig
```diff
@@ -6903,7 +6904,20 @@ pub fn ReplicaType(
                 return;
             }
 
-            // We may even be several views ahead and may now have a completely different pipeline.
+            if (self.commit_min != self.commit_max or self.commit_stage != .idle) {
```
matklad, line 6909: commit_min != commit_max is the actually fuzz failure here, but I think we need commmit_stage as well for the same reason we check it in repair

sentientwaffle, line 6909: Can we also assert `commit_stage == .idle` in `primary_repair_pipeline_read()`, to mirror this condition?

matklad: Good suggestions, added!

sentientwaffle (approved):

PR #2578, opened 2024-12-18
io: add retry for flock

cb22:
On Linux, TigerBeetle takes out an `flock` over the data file when starting. This turns out to lead to interesting races in CI or tests, when killing and starting replicas in an automated fashion in quick succession.

`flock`s are cleaned up by the kernel when the file descriptor is closed, but since that file descriptor is used by io_uring, it actually outlives the process itself.

That means it's possible to kill a tigerbeetle replica, have it really be dead, start another one on the same data file and get a `another process holds the data file lock` error!

There's unfortunately no real satisfying way of fixing this nicely that I could find; so I thought the simplest, a tiny backoff loop, might be the best.

It is actually possible to mix in fcntl per-process locks, and check both: if the fcntl lock doesn't exist, but the flock does, you can then conditionally back off. But, this is a fair bit of extra complexity.

50ms chosen as a backoff value because in practice I've seen it be released within ~3ms relatively consistently.

### DIFF
--- a/src/io/linux.zig
+++ b/src/io/linux.zig
@@ -1477,10 +1477,30 @@ pub const IO = struct {
 
         // Obtain an advisory exclusive lock that works only if all processes actually use flock().
         // LOCK_NB means that we want to fail the lock without waiting if another process has it.
-        posix.flock(fd, posix.LOCK.EX | posix.LOCK.NB) catch |err| switch (err) {
-            error.WouldBlock => @panic("another process holds the data file lock"),
-            else => return err,
-        };
+        //
+        // This is wrapped inside a retry loop with a sleep because of the interaction between
+        // io_uring semantics and flock: flocks are held per fd, but io_uring will keep a reference
+        // to the fd alive even once a process has been terminated, until all async operations have
+        // been completed.
+        //
+        // This means that when killing and starting a tigerbeetle process in an automated way, you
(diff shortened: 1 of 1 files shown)


src/io/linux.zig
```diff
@@ -1477,10 +1477,27 @@ pub const IO = struct {
...
+        // to the fd alive even once a process has been terminated, until all async operations have
+        // been completed.
+        //
+        // This means that when killing and starting a tigerbeetle process in an automated way, you
+        // can see "another process holds the data file lock" errors, even though the process really
+        // has terminated.
+        for (0..2) |_| {
+            posix.flock(fd, posix.LOCK.EX | posix.LOCK.NB) catch |err| switch (err) {
+                error.WouldBlock => {
+                    std.time.sleep(50 * std.time.ns_per_ms);
```
matklad, line 1492: I am mildly annoyed that this'll do one extra needless sleep at the end, but also that there's no bullet-proof way to write a nice retry loop without this issue. 

Consider doing one last try in else:

https://github.com/tigerbeetle/tigerbeetle/blob/216778fdcc3705704a6ff6a5bfd1ce5f6840b0e1/src/clients/python/ci.zig#L73-L94

matklad (dismissed):

matklad (approved):

PR #2579, opened 2024-12-18
linux: fix flag value

matklad:
3 is CURRENT | FUTURE, not ONFAULT

This is not a big bug, FUTURE is no-op for us, and ONFAULT is only
relevant for stack space, which is unlikely to be swapped out, but,
still, this is a bug!


cb22:
Ah, great catch. I think this removes the need for #2573 - I was wondering why MCL_CURRENT | MCL_ONFAULT was behaving slightly differently from what the man page said :smile:.

cb22 (approved):

matklad:
I am no sure this obviates #2573, I would expect this RP to behave the same as main, give-or-take a couple of pages.

PR #2581, opened 2024-12-18
Move repl files to their own directory

brson:
We have accumulated several files that only support the repl, and I want the repl's parser to live in its own file as well. This PR puts `completion.zig`, `terminal.zig` in `repl/` and extracts `parser.zig`.


cb22 (approved):
Drive by approval!

PR #2582, opened 2024-12-19
VSR: Timeout backoff is too aggressive

sentientwaffle:
## Bug

For timeouts which use backoff, the backoff never gets "reset".
- The first time a timeout ever backs off, that timeout interval is forever increased to at least `700ms` (`constants.rtt_ms * constants.rtt_multiple + constants.backoff_min_ms`)
- (Added to this is a jitter, which varies. The jitter's contribution to the timeout is _not_ permanent.)

There are two timeouts which use backoff:
- The primary's `prepare_timeout` (for retrying prepare replication).
- The client's `request_timeout`.

## Fix

When we stop the timeout, restore its duration to the original value.

(The extra `t.run()` is to fix a test that broke due to the permutation.)

### DIFF
--- a/src/vsr.zig
+++ b/src/vsr.zig
@@ -673,6 +673,7 @@ pub const Timeout = struct {
     name: []const u8,
     id: u128,
     after: u64,
+    after_dynamic: ?u64 = null, // null iff !ticking
     attempts: u8 = 0,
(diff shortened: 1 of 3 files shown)


matklad (approved):
🤦 

I assumed that this is exactly how our timeouts actually worked!

PR #2583, opened 2024-12-19
VSR: Set timeout RTTs dynamically

sentientwaffle:
## Bug

From `constants.zig`:

    /// The conservative round-trip time at startup when there is no network knowledge.
    /// Adjusted dynamically thereafter for RTT-sensitive timeouts according to network congestion.
    /// This should be set higher rather than lower to avoid flooding the network at startup.
    pub const rtt_ticks = config.process.rtt_ms / tick_ms;

We weren't ever actually adjusting the RTT for timeouts.
So when computing backoff, we always used `600ms` as the RTT, which is typically too long.

Two timeouts use backoff:
- the replica's `prepare_timeout`, and
- the client's `request_timeout`.

## Fix

Estimate the RTT and use that to dynamically set the timeout. This is _very_ rough right now -- ideally we would be using more samples, and distinguishing between different replicas.

For the client:
- `vsr.ClientType()` now takes a `Time` type.
- `ping_client` messages now include the client's monotonic timestamp.
- `pong_client` echoes back the client's monotonic timestamp.
- Right now the client doesn't do any sort of averaging over RTTs, it just always uses the latest one from the *primary*.   (Since the primary is where requests are being sent.)

## Benchmark

The `prepare_timeout` fix is a huge improvement for throughput in a compromised cluster -- p100 is 8s instead of 27s. (...There is still a lot of room for improvement.)

To test that, run the benchmark against a cluster-of-3 that only has 2 replicas running:

```
$ ./tigerbeetle benchmark --clients=32 --transfer-batch-size=5 --transfer-count=2000
```

(The small batches are to focus the benchmark on consensus rather than compaction.)

<details>
<summary>Results before</summary>

```
400 batches in 329.70 s
transfer batch size = 5 txs
transfer batch delay = 0 us
load accepted = 6 tx/s
batch latency p1 = 3343 ms
batch latency p10 = 25688 ms
batch latency p20 = 25932 ms
batch latency p30 = 26034 ms
batch latency p40 = 26133 ms
batch latency p50 = 26255 ms
batch latency p60 = 26374 ms
batch latency p70 = 26554 ms
batch latency p80 = 26687 ms
batch latency p90 = 26832 ms
batch latency p95 = 27013 ms
batch latency p99 = 27192 ms
batch latency p100 = 27277 ms
```

</details>

<details>
<summary>Results after</summary>

```
400 batches in 91.80 s
transfer batch size = 5 txs
transfer batch delay = 0 us
load accepted = 21 tx/s
batch latency p1 = 1443 ms
batch latency p10 = 6718 ms
batch latency p20 = 6961 ms
batch latency p30 = 7054 ms
batch latency p40 = 7189 ms
batch latency p50 = 7289 ms
batch latency p60 = 7361 ms
batch latency p70 = 7443 ms
batch latency p80 = 7538 ms
batch latency p90 = 7640 ms
batch latency p95 = 7767 ms
batch latency p99 = 7920 ms
batch latency p100 = 8042 ms
```

</details>

### DIFF
--- a/src/repl.zig
+++ b/src/repl.zig
@@ -477,8 +477,8 @@ const repl_history_entry_bytes_without_nul = 511;
 
 const ReplBufferBoundedArray = stdx.BoundedArrayType(u8, repl_history_entry_bytes_without_nul);
 
-pub fn ReplType(comptime MessageBus: type) type {
-    const Client = vsr.ClientType(StateMachine, MessageBus);
+pub fn ReplType(comptime MessageBus: type, comptime Time: type) type {
+    const Client = vsr.ClientType(StateMachine, MessageBus, Time);
 
     // Requires 512 * 256 == 128KiB of stack space.
     const HistoryBuffer = RingBufferType(
@@ -959,9 +959,12 @@ pub fn ReplType(comptime MessageBus: type) type {
 
         pub fn init(
             allocator: std.mem.Allocator,
-            addresses: []const std.net.Address,
-            cluster_id: u128,
-            verbose: bool,
+            time: Time,
+            options: struct {
+                addresses: []const std.net.Address,
+                cluster_id: u128,
+                verbose: bool,
+            },
         ) !Repl {
             var arguments = try std.ArrayListUnmanaged(u8).initCapacity(
                 allocator,
@@ -986,11 +989,12 @@ pub fn ReplType(comptime MessageBus: type) type {
                 allocator,
                 .{
                     .id = client_id,
-                    .cluster = cluster_id,
-                    .replica_count = @intCast(addresses.len),
+                    .cluster = options.cluster_id,
+                    .replica_count = @intCast(options.addresses.len),
+                    .time = time,
                     .message_pool = message_pool,
                     .message_bus_options = .{
-                        .configuration = addresses,
+                        .configuration = options.addresses,
                         .io = io,
                     },
                 },
@@ -999,7 +1003,7 @@ pub fn ReplType(comptime MessageBus: type) type {
 
             return .{
                 .client = client,
-                .debug_logs = verbose,
+                .debug_logs = options.verbose,
                 .request_done = true,
                 .event_loop_done = false,
                 .interactive = false,
--- a/src/vsr.zig
+++ b/src/vsr.zig
@@ -720,7 +720,7 @@ pub const Timeout = struct {
     /// Sets the value of `after` as a function of `rtt` and `attempts`.
     /// Adds exponential backoff and jitter.
     /// May be called only after a timeout has been stopped or reset, to prevent backward jumps.
-    pub fn set_after_for_rtt_and_attempts(self: *Timeout, random: std.rand.Random) void {
+    fn set_after_for_rtt_and_attempts(self: *Timeout, random: std.rand.Random) void {
         // If `after` is reduced by this function to less than `ticks`, then `fired()` will panic:
         assert(self.ticks == 0);
         assert(self.rtt > 0);
--- a/src/vsr.zig
+++ b/src/vsr.zig
@@ -753,14 +753,16 @@ pub const Timeout = struct {
         assert(self.rtt > 0);
         assert(rtt_ticks > 0);
 
-        log.debug("{}: {s} rtt={}..{}", .{
-            self.id,
-            self.name,
-            self.rtt,
-            rtt_ticks,
-        });
+        if (self.rtt != rtt_ticks) {
+            log.debug("{}: {s} rtt={}..{}", .{
+                self.id,
+                self.name,
+                self.rtt,
+                rtt_ticks,
+            });
 
-        self.rtt = rtt_ticks;
+            self.rtt = rtt_ticks;
+        }
     }
 
     pub fn start(self: *Timeout) void {
--- a/src/vsr.zig
+++ b/src/vsr.zig
@@ -749,9 +749,11 @@ pub const Timeout = struct {
         assert(self.after_dynamic.? > 0);
     }
 
-    pub fn set_rtt(self: *Timeout, rtt_ticks: u64) void {
+    pub fn set_rtt_ns(self: *Timeout, rtt_ns: u64) void {
         assert(self.rtt > 0);
-        assert(rtt_ticks > 0);
+
+        const rtt_ms = @divFloor(rtt_ns, std.time.ns_per_ms);
+        const rtt_ticks = @max(1, @divFloor(rtt_ms, constants.tick_ms));
 
         if (self.rtt != rtt_ticks) {
             log.debug("{}: {s} rtt={}..{}", .{
--- a/src/vsr.zig
+++ b/src/vsr.zig
@@ -754,16 +754,17 @@ pub const Timeout = struct {
 
         const rtt_ms = @divFloor(rtt_ns, std.time.ns_per_ms);
         const rtt_ticks = @max(1, @divFloor(rtt_ms, constants.tick_ms));
+        const rtt_ticks_clamped = @min(rtt_ticks, constants.rtt_max_ticks);
 
-        if (self.rtt != rtt_ticks) {
+        if (self.rtt != rtt_ticks_clamped) {
             log.debug("{}: {s} rtt={}..{}", .{
                 self.id,
                 self.name,
                 self.rtt,
-                rtt_ticks,
+                rtt_ticks_clamped,
             });
 
-            self.rtt = rtt_ticks;
+            self.rtt = rtt_ticks_clamped;
         }
     }
 
--- a/src/vsr/client.zig
+++ b/src/vsr/client.zig
@@ -460,6 +460,7 @@ pub fn ClientType(
                     }
                 }
                 std.mem.sort(u64, round_trip_times_ns.slice(), {}, std.sort.asc(u64));
+                assert(round_trip_times_ns.count() > 0);
 
                 const rtt_median_ns =
                     round_trip_times_ns.get(@divFloor(round_trip_times_ns.count(), 2));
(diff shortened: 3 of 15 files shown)


src/vsr/replica.zig
```diff
@@ -1501,6 +1504,7 @@ pub fn ReplicaType(
                 .replica = self.replica,
                 .view = self.view,
                 .release = self.release,
+                .ping_timestamp_monotonic = message.header.ping_timestamp_monotonic,
```
sentientwaffle, line 1510: If this ping message is from an old client, then we are interpreting their reserved (zeroed) bytes as `ping_timestamp_monotonic`. The replica echos those zeroes back, and the client checks on the pong side that they are zeros (reserved).

I tested this locally between this a replica running branch (built as if it was release 0.16.18) and a client (the benchmark) running 0.16.17.

matklad: Ahhh, I missed that!

src/repl.zig
```diff
@@ -961,7 +961,10 @@ pub fn ReplType(comptime MessageBus: type) type {
             allocator: std.mem.Allocator,
             addresses: []const std.net.Address,
             cluster_id: u128,
-            verbose: bool,
+            options: struct {
+                verbose: bool,
+                time: Time,
+            },
         ) !Repl {
```
matklad: This _feels_ weird. I'd maybe do

```
        pub fn init(
            allocator: std.mem.Allocator,
            time: Time,
            options: struct {
                verbose: bool,
                addresses: []const std.net.Address,
                cluster_id: u128,
            },
        ) !Repl {
```

Dependencies are injected positionally, as their type pretty much uniquely identifies them. Actually parameters that vary are passer through options. 


No super strong feelings here!

src/vsr/clock.zig
```diff
@@ -368,6 +368,27 @@ pub fn ClockType(comptime Time: type) type {
...
+                }
+            }
+
+            if (one_way_delays.count() == 0) {
+                return null;
+            } else {
+                std.mem.sort(u64, one_way_delays.slice(), {}, std.sort.asc(u64));
+                const one_way_delay_median =
+                    one_way_delays.get(@divFloor(one_way_delays.count(), 2));
+                return one_way_delay_median * 2;
```
matklad, line 388: `assert(one_way_delay_median > 0)` perhaps?

src/vsr/client.zig
```diff
@@ -433,6 +446,32 @@ pub fn ClientType(comptime StateMachine_: type, comptime MessageBus: type) type
...
+                self.request_timeout.set_rtt_ns(rtt_median_ns);
+            } else {
+                log.debug("{}: on_pong: monotonic timestamp regressed {}..{} replica={}", .{
+                    self.id,
+                    ping_timestamp_monotonic,
+                    pong_timestamp_monotonic,
+                    pong.header.replica,
+                });
+            }
         }
```
matklad: We repeat this "get median of things" code two times, and it would be great to have `stdx.median` API here, but I don't see how that could work, sadly... The problem is that we need to filter-out null elements from the input slice first, and that's tricky to express without iterators. 

So perhaps just repeating it twice is best!

src/vsr/clock.zig
sentientwaffle: Is that a guarantee? I think the monotonic time is not strictly increasing.

src/vsr/client.zig
sentientwaffle: Yeah, also I think this is temporary -- as we add more "topology-awareness" I think we will come up with a better approach anyway.

src/vsr/clock.zig
matklad: Ah, right, not sure what I was thinking here!

src/vsr/clock.zig
```diff
@@ -368,6 +368,27 @@ pub fn ClockType(comptime Time: type) type {
...
+            var one_way_delays = stdx.BoundedArrayType(u64, constants.replicas_max + 1){};
+            for (self.window.sources, 0..) |source, replica_index| {
+                if (self.replica != replica_index) {
+                    if (source) |sampled| {
+                        one_way_delays.append_assume_capacity(sampled.one_way_delay);
+                    }
+                }
+            }
+
+            if (one_way_delays.count() == 0) {
```
matklad: maybe

```suggestion
            if (one_way_delays.count() < self.quorum) {
```

to prevent rtt from jerking abruptly when we switch the window?

src/vsr.zig
```diff
@@ -749,18 +749,22 @@ pub const Timeout = struct {
         assert(self.after_dynamic.? > 0);
     }
 
-    pub fn set_rtt(self: *Timeout, rtt_ticks: u64) void {
+    pub fn set_rtt_ns(self: *Timeout, rtt_ns: u64) void {
```
matklad, line 752: Should we hard-clamp `rtt` to be between, say, 4ms and 5min? As defense in depth, to make sure that we don't end up with a "3 and a half ice-ages" timeout due to a bug somewhere? 

🤔 So, 5m was about "how long would it take to send a 1MiB prepare over dial-up", and that made me realize that what we are measuring is indeed rtt via pings, but what we are actually interested in is the time to send a prepare/request, which actually _might_ be throughput-limited, for nearly-full messages. 

Let's maybe set the lower bound to something like 30ms then, to prevent cases where pings make it feel like the connection is really good, while in fact it is not that great. Our clock uses the min rtt over a window, so, even if some pings will be naturally delayed by the prepares in the queue in front of them, I think we might get a situation where a window has a bunch of lucky pings that depress our rtt and cause us to retry prepares, which just take a long time to flush down the narrow pipe.

src/vsr/client.zig
```diff
@@ -433,6 +446,32 @@ pub fn ClientType(comptime StateMachine_: type, comptime MessageBus: type) type
...
+
+                var round_trip_times_ns = stdx.BoundedArrayType(u64, constants.replicas_max){};
+                for (self.replica_round_trip_times_ns) |round_trip_time_ns| {
+                    if (round_trip_time_ns) |rtt_ns| {
+                        round_trip_times_ns.append_assume_capacity(rtt_ns);
+                    }
+                }
+                std.mem.sort(u64, round_trip_times_ns.slice(), {}, std.sort.asc(u64));
+
+                const rtt_median_ns =
```
matklad, line 465: assert that we have at least one timestamp to choose from

src/vsr/client.zig
```diff
@@ -93,6 +99,10 @@ pub fn ClientType(comptime StateMachine_: type, comptime MessageBus: type) type
         /// Used for end-to-end keepalive, and to discover a new primary between requests.
         ping_timeout: vsr.Timeout,
 
+        /// The round-trip time (estimated by the latest ping/pong pair) from each replica.
+        replica_round_trip_times_ns: [constants.replicas_max]?u64 =
+            .{null} ** constants.replicas_max,
```
matklad, line 104: Hm, so ideally we should clear this periodically. Imagine the case where we had a replica real close, and then that replica dies, and all our other pings are slower. The false fast ping would lead us to an incorrect estimation, until we hear from the replica agiain. 

I guess this is mostly mitigated by us taking the median, but what bothers me is just the fact that we use different logic between the replica and the client.

🤔 can we use the full `Clock` here? I _think_ client can be treated as a stand-by? They need the same logic, where they are aware of cluster time, but don't affect it.  That way, we'll even dedup the median-computing code.

src/vsr.zig
matklad: piggybacking, but `set_after_for_rtt_and_attempts` doesn't need to be `pub`.

sentientwaffle: Isn't 30ms on the high side? e.g. same-zone latency is going to be much lower than that.

src/vsr/client.zig
sentientwaffle: :thinking: `Clock` additionally demands the replica's wall time, which we don't include on pongs.

src/vsr.zig
matklad: Yeah, it is high, but intentionally so, and is 10x lower than what we have today anyway. I am thinking about 10mbs connection, where the rtt for ping could be quite fast, but sending a meg would take almost a second. Because we use timeouts for prepares and requests, we should estimate how long does it take to send a MiB.

Although, I guess maybe we don't need lower bound that much, because our delay will be at least `backoff_min_ms` which is 100ms anyway?

src/vsr/client.zig
matklad: Ah, right, then I think it is fine as is!

src/vsr.zig
sentientwaffle: Would the slow connection case already be covered by the message bus, via dropping messages from the send queue?

matklad: Hm, I don't quite understand what you mean here. As in, I think we will see dropped messages in the send queue, yeah, and that is bad. 

Basically, the doom and gloom scenario in my head is that the primary gets a fast rtt from small pings, which makes it pick a small delay for prepare timeout. This will cause the primary to re-send the prepare, which makes the situation worse. 

The right solution here would be for the primary to measure the rtt for the prepare. If it somehow were able to do that, it would be able to reason along the lines of 

> I see that it takes only 10ms to rtt a ping message, which is small. But I also see that I need 100ms to rtt a 1MiB prepare, so, if I am not seeing prepare_ok after 50 ms, that's probably because it just takes long to send a prepare. 

Now, the primary can't know the rtt for prepares, it knows only rtt for pings, which, in low-bandwidth scenarios, can be much lower than prepare rtt. So my proposed fix is to bound the rtt from below. 

---

Though, again, I didn't realize that we add at least 100ms to the retry delay on top of rtt delay. Given that, I think it's fine to not bound from bellow here.

sentientwaffle: Alright, I left it out for now -- I agree that the right solution is to track prepare latency independently.

matklad: (I think we do need a bound from above in any case, as I think the code as is would allow practically infinite timeouts)

sentientwaffle: Ah good point -- we have a clamp on the backoff amount but that doesn't cover the RTT itself.

matklad (approved):

PR #2584, opened 2024-12-19
vsr: throttle repair a bit

matklad:
Our repair can create a feedback loop. repair requests prepares and headers, but, upon receiving either back, we re-trigger repair, which could lead to duplicate repair work.

To avoid that, make sure that we are not sending more than one repair message per replica per our repair timeout.

### DIFF
--- a/src/constants.zig
+++ b/src/constants.zig
@@ -74,6 +74,14 @@ comptime {
     assert(vsr_checkpoint_ops % lsm_compaction_ops == 0);
 }
 
+pub const vsr_repair_message_budget_max = replicas_max * 2;
+pub const vsr_repair_message_budget_refill = @divFloor(vsr_repair_message_budget_max, 2);
+
+comptime {
+    assert(vsr_repair_message_budget_max > 0);
+    assert(vsr_repair_message_budget_refill > 0);
+}
+
 /// The maximum number of clients allowed per cluster, where each client has a unique 128-bit ID.
 /// This impacts the amount of memory allocated at initialization by the server.
 /// This determines the size of the VR client table used to cache replies to clients by client ID.
--- a/src/vsr/replica.zig
+++ b/src/vsr/replica.zig
@@ -420,6 +420,9 @@ pub fn ReplicaType(
         /// Unique do_view_change messages for the same view from ALL replicas (including ourself).
         do_view_change_from_all_replicas: DVCQuorumMessages = dvc_quorum_messages_null,
 
+        // +1 to make it easier to decrement budget with -|.
+        repair_messages_budget: u32 = constants.vsr_repair_message_budget_max + 1,
+
         /// Whether the primary has received a quorum of do_view_change messages for the view
         /// change. Determines whether the primary may effect repairs according to the CTRL
         /// protocol.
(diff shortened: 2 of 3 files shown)


matklad:
Test failure is the same one as in https://github.com/tigerbeetle/tigerbeetle/pull/2582

src/vsr/replica_test.zig
```diff
@@ -1785,6 +1785,7 @@ test "Cluster: view_change: lagging replica repairs WAL using start_view from po
     a0.pass_all(.R_, .bidirectional);
     b1.pass(.R_, .outgoing, .do_view_change);
 
+    t.run();
```
matklad, line 1788: sic, I needed to add one more `.run` on top of what https://github.com/tigerbeetle/tigerbeetle/pull/2582 did

src/vsr/replica.zig
```diff
@@ -3116,6 +3118,10 @@ pub fn ReplicaType(
 
         fn on_repair_timeout(self: *Replica) void {
             assert(self.status == .normal or self.status == .view_change);
```
sentientwaffle, line 3121:
```suggestion
            assert(self.status == .normal or self.status == .view_change);
            assert(self.repair_messages_budget <= constants.vsr_repair_message_budget_max);
```

src/constants.zig
```diff
@@ -42,6 +42,9 @@ pub const vsr_checkpoint_ops = journal_slot_count -
     lsm_compaction_ops -
     lsm_compaction_ops * stdx.div_ceil(pipeline_prepare_queue_max * 2, lsm_compaction_ops);
 
+pub const vsr_repair_message_budget_max = replicas_max;
```
matklad: On the fence on whether this should be replicas_max or replicas_max * 2

src/vsr/replica.zig
```diff
@@ -6405,17 +6411,19 @@ pub fn ReplicaType(
...
-                    @bitCast(Header.RequestStartView{
-                        .command = .request_start_view,
-                        .cluster = self.cluster,
-                        .replica = self.replica,
-                        .view = self.view,
-                        .nonce = self.nonce,
-                    }),
-                );
+                self.repair_messages_budget -|= 1;
+                if (self.repair_messages_budget > 0) {
```
sentientwaffle, line 6418: So if `repair_messages_budget` was originally 1, then we reduce it to `0` without sending anything?
That is somewhat surprising, but seems fine. Though it means we should comptime-assert that `vsr_repair_message_budget_max > 1`.

src/constants.zig
sentientwaffle: Or `(replicas_max - 1) * 2` since we aren't sending to ourselves?

matklad: repair timeout is 750, now this is one repair message every 250 ms, which actually feels too low, will bump to `*2`

src/vsr/replica.zig
matklad: Yeah, good point! Set the budget t one max + 1 with a comment explaining why.

sentientwaffle (approved):

PR #2585, opened 2024-12-19
Tracing: Log duration at end of event

sentientwaffle:
Log duration at end of event. (You could just calculate it by hand from the log timestamps, but that is inconvenient.)

One possible downside to this is that the debug logs generated by a VOPR run are no longer deterministic. But we don't ever rely on that anyway. (If it would be useful we can gate this feature.)


matklad (approved):
Rather than gating, we could also parametrize tracer by time, but I think it's fine as is!

PR #2587, opened 2024-12-19
Config: Temporarily decrease default RTT

sentientwaffle:
Temporarily increase default RTT to mitigate slow timeouts in compromised clusters. This impacts the `prepare_timeout` in particular.

(This will be reverted as part of https://github.com/tigerbeetle/tigerbeetle/pull/2583.)


src/config.zig
```diff
@@ -115,7 +115,7 @@ const ConfigProcess = struct {
     client_replies_iops_read_max: usize = 1,
     client_replies_iops_write_max: usize = 2,
     tick_ms: u63 = 10,
-    rtt_ms: u64 = 300,
+    rtt_ms: u64 = 50,
```
matklad, line 118: Is it intentional that we decrease Client request_timeout with this as well?

matklad: This line

https://github.com/tigerbeetle/tigerbeetle/blob/8fb15c76e3fe294d5846a95405c521174620b107/src/vsr/client.zig#L152-L153

sentientwaffle: Ah, partly. The request_timeout actually uses RTT twice.. so half of that was intentional :thinking:

matklad (dismissed):

matklad (approved):

matklad:
I think we no longer want to do this? Optimistically closing!

sentientwaffle:
Yep, this was superseded by https://github.com/tigerbeetle/tigerbeetle/pull/2583!

PR #2590, opened 2024-12-20
logging: expose more debug logging on cli and python client

cb22:
This PR adds `--log-debug` for the `format`, `benchmark`, `multiversion` and `repl` subcommands.

`benchmark` gets two parameters - `--log-debug`, which controls the debug logging of the benchmark load, and `--log-debug-replica` which controls the debug logging of the started replica.

Additionally, expose debug logging to the Python client, and have it register a log callback - making it the first language log integrated client!

### DIFF
--- a/src/clients/c/tb_client.zig
+++ b/src/clients/c/tb_client.zig
@@ -124,10 +124,12 @@ pub const Logging = struct {
 
 pub fn register_log_callback(
     callback_maybe: ?Logging.Callback,
+    debug: bool,
 ) callconv(.C) tb_register_log_callback_status_t {
     if (logging.callback == null) {
         if (callback_maybe) |callback| {
             logging.callback = callback;
+            logging.debug = debug;
             return .success;
         } else {
             return .not_registered;
@@ -135,6 +137,7 @@ pub fn register_log_callback(
     } else {
         if (callback_maybe == null) {
             logging.callback = null;
+            logging.debug = debug;
             return .success;
         } else {
             return .already_registered;
--- a/src/clients/python/tests/test_basic.py
+++ b/src/clients/python/tests/test_basic.py
@@ -5,7 +5,7 @@ from dataclasses import asdict
 import pytest
 
 import tigerbeetle as tb
-
+tb.configure_logging(debug=True)
 
 @pytest.fixture
 def client():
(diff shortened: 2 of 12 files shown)


src/clients/python/tests/test_basic.py
```diff
@@ -5,7 +5,7 @@
 import pytest
 
 import tigerbeetle as tb
-
+tb.configure_logging(debug=True)
```
cb22, line 8: Since logging is per application and not per client, I though this was the most suitable API. (`debug=True` when creating a client would be nice, but not really what's happening behind the scenes).

src/tigerbeetle/benchmark_load.zig
```diff
@@ -10,9 +10,6 @@ const builtin = @import("builtin");
 const assert = std.debug.assert;
 const panic = std.debug.panic;
 const log = std.log;
-pub const std_options = .{
```
cb22, line 13: (this hasn't been used for a while, since this file isn't an entry point!)

src/tigerbeetle/main.zig
```diff
@@ -65,19 +65,24 @@ pub fn main() !void {
...
 
+    switch (command) {
+        inline else => |*args| {
+            if (@hasField(@typeInfo(@TypeOf(args)).Pointer.child, "log_debug")) {
+                if (args.log_debug) {
+                    log_level_runtime = .debug;
+                }
+            }
+        },
+    }
```
sentientwaffle: I think this would be more explicit + readable:
```suggestion
    switch (command) {
        .version => {},
        inline else => |*args| {
            if (args.log_debug) {
                log_level_runtime = .debug;
            }
        },
    }
```

In particular, if we add a new subcommand, it means we would need to explicitly "opt out" of debug logs for it to compile.

src/clients/c/tb_client.zig
```diff
@@ -124,17 +124,20 @@ pub const Logging = struct {
 
 pub fn register_log_callback(
     callback_maybe: ?Logging.Callback,
+    debug: bool,
```
sentientwaffle, line 127: :thinking: I don't love bare boolean arguments, but since this is exposed via bindings I guess it is simpler than adding a new options struct.

cb22: Yeah - I'd like to keep the extern fn signatures as simple as possible! No structs (or u128s...).

sentientwaffle (approved):

Issue #2469, opened 2024-11-11
LSM performance poor when batches are small


gladkikhartem:
1. Setup Tigerbeetle cluster inside docker based on this Page: https://docs.tigerbeetle.com/operating/docker/  with a clean, fresh data storage.
2. Create 10k accounts using Golang client (v0.16.4)  (with flag **History:true**)
3. Create a for loop making 1 transaction to tigerbeetle at a time (i.e. batch size = 1, concurrency = 1)

After cluster is Initially created latency distribution is following (ranges in milliseconds, measurement sample size = 1000);
```
1-6.7      84.9%  ██████████▏  849
6.7-12.4   7.2%   ▉            72
12.4-18.1  3.6%   ▌            36
18.1-23.8  0.3%   ▏            3
23.8-29.5  0%     ▏            
29.5-35.2  1%     ▏            10
35.2-40.9  1%     ▏            10
40.9-46.6  1.3%   ▏            13
46.6-52.3  0.5%   ▏            5
52.3-58    0.2%   ▏            2
```

4. Create 1M transactions in TB via 1000 batches of 1000 transactions
5. Repeat the measurement (batch size = 1, concurrency = 1)
Now latency distribution looks like following (ranges in milliseconds, measurement sample size = 1000);
```
1-87.1       53.1%  ██████████▏  531
87.1-173.2   41.1%  ███████▊     411
173.2-259.3  4.1%   ▊            41
259.3-345.4  0.8%   ▏            8
345.4-431.5  0.2%   ▏            2
431.5-517.6  0.2%   ▏            2
517.6-603.7  0.1%   ▏            1
603.7-689.8  0%     ▏            
689.8-775.9  0.1%   ▏            1
775.9-862    0.3%   ▏            3
```

Docker stats:
```
fd603a23591d   wallet_tigerbeetle_1_1   16.67%    2.396GiB / 13.34GiB   17.96%    0B / 0B       3.4GB / 27.2GB    10
8f0ddc1dc408   wallet_tigerbeetle_2_1   16.59%    2.405GiB / 13.34GiB   18.02%    0B / 0B       3.41GB / 27.2GB   10
ebf04163809a   wallet_tigerbeetle_0_1   17.12%    2.388GiB / 13.34GiB   17.89%    0B / 0B       3.39GB / 27.2GB   11
```
As you see TigerBeetle has written 27GB! of data for 1 million transactions inserted (1000 batches of 1000).
[...]

PR #2070, opened 2024-07-03
LSM: Greedy level-0 table coalescing

sentientwaffle:
Currently trees with disjoint (or mostly-disjoint) tables have extremely high space amplification when `create_transfers` batches are not full (i.e. when `request.header.size << 1MiB`).

This PR is a stopgap fix, to mitigate space amp until a more considered/performant fix can land.
It patches the space amp issue, but at the cost of increasing write amp when batches are small.

#### Space Amplification

An immutable table is constructed from the updates that occur over a single bar.
The immutable table is then compacted down into level 0. (etc).

- Each table contains, at minimum, a single value. (1 index block -> 1 data block -> 1 value).
- All data blocks in a table except the final one are guaranteed to be full. (The final data block in a table is only guaranteed at least one value.)
- No matter how many values they hold, each data block occupies 512KiB.

This leads to the "space amplification" ratio -- the ratio of bytes actually occupied by values vs the total bytes (including the block padding up to 512KiB, since blocks are fixed-size). A high space amplification ratio means that there is a lot of space on disk wasted. Since TigerBeetle has a fixed maximum number of tables, "tiny tables" (i.e. tables with very few values / high space amplification) also effectively reduce the maximum number of accounts/transfers that the cluster can store.

For example, if a single transfer (128B) is created over 32 ops, the immutable table will be written to level 0, and will occupy 1MiB (512KiB index block + 512KiB data block).

#### Compaction

When tables are compacted, if the level-A input table has a disjoint key range the tables in level B, then:
- If level B is 0 (the level-A table is the immutable table), then just write the immutable table directly into level B -- no merging.
- If level B is >0, then "move-table": update the input table's level in the metadata (from A to B). (None of the values need to be copied. All of the value blocks are left untouched).

In the transfers object tree, these disjoint tables are guaranteed, since keys (timestamps) are strictly increasing and transfers are immutable. Other trees typically have (some degree of) overlapping tables, depending on workload.

Disjoint tables are the best case for compaction performance -- metadata updates are pretty much free -- _much_ cheaper than merge-compaction.
But the downside is that without merging, tiny tables are tiny forever.

#### Example

In the production config, with a 1MiB batch size and an `lsm_batch_multiple` of `32`, a transfers object table references as many as ~32MiB of transfers, split into 512KiB data blocks (modulo headers).

In the worst case, a table might hold only a single transfer, leading to a space amp ratio of 8192.

And unfortunately, this runaway-space-amp "bug" is extremely easy to hit:
[...]

PR #2592, opened 2024-12-20
LSM: Coalesce tables in memory

sentientwaffle:
## Context

Improve the throughput and write amplification of workloads with small batches.

### Coalescing

[Greedy table coalescing](https://github.com/tigerbeetle/tigerbeetle/pull/2070) reduced space amplification of inserting objects with partially-full batches. (Especially for disjoint tables.)
It accomplished this by aggressively compacting small tables together until they aren't small anymore. (Even if the tables are only adjacent without actual overlap.)

Unfortunately, this extra compaction adds a lot of extra CPU/IO (especially write amplification) work to the cluster. The smaller the batches, the more times we rewrite the table.

### Immutable table compaction

Right now:

- We convert the mutable table into the immutable table after the last beat of each bar. (Aka the last beat of each second-half-bar.)
- But we don't actually start compacting the immutable table into level 0 until the first beat of each second-half-bar (odd → even).

For clarity:

```
[          bar 1         ][          bar 2         ]
[half-bar 1A][half-bar 1B][half-bar 2A][half-bar 2B]
```

The immutable table created at the end of *bar 1* (at the end of *half-bar 1B*) is flushed during *half-bar 2B*.
The immutable table isn't used (except for reads) during *half-bar 2A*.

## Optimization

At the end of each first-half-bar (e.g. *2A*), if the sum of:

- the number of values in the immutable table (constructed at the end of *1B*),
- the number of values in the mutable table so far (during *2A*), and
- the worst-case number of values that may be inserted into the mutable table during the last half-bar (*2B*)

...fits within a single table, then _don't compact the immutable table_.

Instead, at the end of the last half-bar (*2B*):

1. append the mutable table's values to the immutable table,
2. sort+dedup the immutable table, and
3. reset the mutable table.

#### Caveats

- This optimization must be disabled during the bar immediately preceding each checkpoint trigger.
  - (Recovering from the checkpoint replays only the last bar of ops, and it must reconstruct the same immutable table.)
  - Because of this, LSM must now be aware of the checkpoint schedule. (Hence the forest/tree fuzzer modifications.)
- This optimization will need to be disabled when we are taking a persistent snapshot.
- Note that this doesn't provide a _guaranteed_ lower-bound table size.

## Performance

This optimization improves transactions-per-second and write amplification for workloads with many small batches.
The smaller the batch, the greater the (relative) improvement.

For example, with 256 transfers/batch the total bytes written is cut to 1/12th of the original, and transactions/second increases by 4.2x.

### Benchmarks

Benchmark notes:

- Benchmarked on tmpfs.
- "total bytes written" is included in the benchmark output but was collected with a kludgey modification to the replica, so is not committed.
- "Before" is `main` as of e16b2787a08d277a8de26fd46d486e1ea49aacc0.

<details>
<summary>256 transfers/batch</summary>

**Before**:

```
$ ./tigerbeetle benchmark --transfer-batch-size=256
...
39063 batches in 270.14 s
transfer batch size = 256 txs
transfer batch delay = 0 us
load accepted = 37018 tx/s
batch latency p1 = 0 ms
batch latency p10 = 0 ms
batch latency p20 = 0 ms
batch latency p30 = 0 ms
batch latency p40 = 1 ms
batch latency p50 = 3 ms
batch latency p60 = 11 ms
batch latency p70 = 12 ms
batch latency p80 = 13 ms
batch latency p90 = 14 ms
batch latency p95 = 15 ms
batch latency p99 = 17 ms
batch latency p100 = 26 ms
...
total bytes written = 32668221440
```

**After**:

```
$ ./tigerbeetle benchmark --transfer-batch-size=256
...
39063 batches in 63.57 s
transfer batch size = 256 txs
transfer batch delay = 0 us
load accepted = 157306 tx/s
batch latency p1 = 0 ms
batch latency p10 = 0 ms
batch latency p20 = 0 ms
batch latency p30 = 0 ms
batch latency p40 = 0 ms
batch latency p50 = 0 ms
batch latency p60 = 0 ms
batch latency p70 = 0 ms
batch latency p80 = 0 ms
batch latency p90 = 1 ms
batch latency p95 = 9 ms
batch latency p99 = 16 ms
batch latency p100 = 28 ms
...
total bytes written = 391991070720
```

~1/12 as many bytes written!

</details>

<details>
<summary>1024 transfers/batch</summary>

**Before**:

```
$ ./tigerbeetle benchmark --transfer-batch-size=1024
...
9766 batches in 86.72 s
transfer batch size = 1024 txs
transfer batch delay = 0 us
load accepted = 115311 tx/s
batch latency p1 = 1 ms
batch latency p10 = 1 ms
batch latency p20 = 2 ms
batch latency p30 = 2 ms
batch latency p40 = 2 ms
batch latency p50 = 6 ms
batch latency p60 = 12 ms
batch latency p70 = 14 ms
batch latency p80 = 15 ms
batch latency p90 = 17 ms
batch latency p95 = 19 ms
batch latency p99 = 22 ms
batch latency p100 = 31 ms
...
total bytes written = 32070647808
```

**After**:

```
$ ./tigerbeetle benchmark --transfer-batch-size=1024
...
9766 batches in 45.87 s
transfer batch size = 1024 txs
transfer batch delay = 0 us
load accepted = 218005 tx/s
batch latency p1 = 1 ms
batch latency p10 = 1 ms
batch latency p20 = 1 ms
batch latency p30 = 1 ms
batch latency p40 = 1 ms
batch latency p50 = 2 ms
batch latency p60 = 2 ms
batch latency p70 = 2 ms
batch latency p80 = 4 ms
batch latency p90 = 14 ms
batch latency p95 = 17 ms
batch latency p99 = 25 ms
batch latency p100 = 30 ms
...
total bytes written = 103231049728
```

~1/3 as many bytes written.

</details>

<details>
<summary>2048 transfers/batch</summary>

**Before**:

```
$ ./tigerbeetle benchmark --transfer-batch-size=2048
...
4883 batches in 55.90 s
transfer batch size = 2048 txs
transfer batch delay = 0 us
load accepted = 178887 tx/s
batch latency p1 = 3 ms
batch latency p10 = 3 ms
batch latency p20 = 4 ms
batch latency p30 = 4 ms
batch latency p40 = 5 ms
batch latency p50 = 9 ms
batch latency p60 = 14 ms
batch latency p70 = 16 ms
batch latency p80 = 17 ms
batch latency p90 = 19 ms
batch latency p95 = 23 ms
batch latency p99 = 35 ms
batch latency p100 = 42 ms
...
total bytes written = 29656117248
```

**After**:

```
$ ./tigerbeetle benchmark --transfer-batch-size=2048
...
...
4883 batches in 43.39 s
transfer batch size = 2048 txs
transfer batch delay = 0 us
load accepted = 230483 tx/s
batch latency p1 = 3 ms
batch latency p10 = 3 ms
batch latency p20 = 4 ms
batch latency p30 = 4 ms
batch latency p40 = 4 ms
batch latency p50 = 4 ms
batch latency p60 = 5 ms
batch latency p70 = 8 ms
batch latency p80 = 15 ms
batch latency p90 = 18 ms
batch latency p95 = 22 ms
batch latency p99 = 41 ms
batch latency p100 = 47 ms
total bytes written = 53931683840
```

~1/2 as many bytes written.

</details>

<details>
<summary>max transfers/batch</summary>

**Before**:

```
$ ./tigerbeetle benchmark
...
load accepted = 327555 tx/s
batch latency p1 = 10 ms
batch latency p10 = 14 ms
batch latency p20 = 14 ms
batch latency p30 = 16 ms
batch latency p40 = 18 ms
batch latency p50 = 21 ms
batch latency p60 = 25 ms
batch latency p70 = 28 ms
batch latency p80 = 30 ms
batch latency p90 = 32 ms
batch latency p95 = 34 ms
batch latency p99 = 110 ms
batch latency p100 = 122 ms
...
total bytes written = 18012184576
```

**After**:

```
$ ./tigerbeetle benchmark
...
1222 batches in 30.32 s
transfer batch delay = 0 us
load accepted = 329818 tx/s
batch latency p1 = 10 ms
batch latency p10 = 13 ms
batch latency p20 = 14 ms
batch latency p30 = 16 ms
batch latency p40 = 18 ms
batch latency p50 = 21 ms
batch latency p60 = 24 ms
batch latency p70 = 27 ms
batch latency p80 = 29 ms
batch latency p90 = 31 ms
batch latency p95 = 33 ms
batch latency p99 = 114 ms
batch latency p100 = 122 ms
...
total bytes written = 18055143424
```

Bytes written is unchanged.

</details>


sentientwaffle:
This fixes https://github.com/tigerbeetle/tigerbeetle/issues/2469.

matklad:
This looks awesome! Everything looks right after the cursory look, but I'll do a careful review tomorrow! If we are releasing on Monday, I'd love to get this in, as it can save a lot of trouble for some users (but, of course, the users probably should arrange for sending us bigger batches)!

src/lsm/compaction.zig
```diff
@@ -41,6 +41,7 @@ const constants = @import("../constants.zig");
 
 const stdx = @import("../stdx.zig");
 const maybe = stdx.maybe;
+const vsr = @import("../vsr.zig");
```
matklad, line 44: Hm, this, and, the corresponding change in the fuzzer in generate op_checkpoint is a bit sad --- I quite like the design where LSM doesn't know _when_ checkpoint happens.

I wonder if it would be worth retaining this design by pulling the decision about checkpoint up, and pass `flush_memory_table: bool` parameter to the LSM? It _feels_ like we might need something like that anyway once we get to snapshots?

src/lsm/compaction.zig
```diff
@@ -541,8 +543,51 @@ pub fn CompactionType(
                     return;
                 }
 
+                const table_value_count_limit =
+                    constants.lsm_compaction_ops * compaction.tree.options.batch_value_count_limit;
```
matklad: 🤔 is this correct? The `tree.options.batch_value_count_limit` is the runtime limit, right? So, it could vary across the replicas, and it can change over time.

src/lsm/tree.zig
```diff
@@ -513,24 +513,29 @@ pub fn TreeType(comptime TreeTable: type, comptime Storage: type) type {
...
+                // In addition, the immutable table is conceptually an output table of this
+                // compaction bar, and now its snapshot_min matches the snapshot_min of the
+                // Compactions' output tables.
+                tree.table_mutable.make_immutable(snapshot_min);
+                tree.table_immutable.make_mutable();
+                std.mem.swap(TableMemory, &tree.table_mutable, &tree.table_immutable);
+            } else {
+                // The immutable table wasn't flushed because there is enough room left over for the
+                // mutable table's values, allowing us to skip some compaction work.
+                tree.table_immutable.absorb(&tree.table_mutable, snapshot_min);
```
matklad, line 536: Maybe add a light-weight assert that `table_immutable.value_context.count + table_mutable.value_context.count <= table_immutable.values.len`?

We assert thoroughly at the definition site, but a rough sanity-check at the call-site could help readability. 

Offtopic, but `value_context` as a name bugs me. Maybe we should flatten it? 🤔

matklad: Ah, and then assert after the call that table_mutabe is empty!

With these two asserts, the reader would perfectly guess what absorb actually is doing!

src/lsm/table_memory.zig
```diff
@@ -29,8 +31,11 @@ pub fn TableMemoryType(comptime Table: type) type {
                 suffix_offset: u32 = 0,
             },
             immutable: struct {
-                /// An empty table has nothing to flush
+                /// An empty table has nothing to flush.
                 flushed: bool = true,
+                /// This field is only used for assertions, to verify that we don't absorb the
+                /// mutable table immediately prior to checkpoint.
+                absorbed: bool = false,
```
matklad, line 38: I wonder if `flushed` & `absorbed` could be merged into a single state enum, for tighter asserts? 

Previously, we used `flushed` to assert that we didn't forget to flush, but now we implicitly make decision that if not flushed, than we should absorb. 

Though, to be fair, I can't immediately come up with good names for the enum variants!

matklad (commented):
LGTM in general, but using `compaction.tree.options.batch_value_count_limit;` feels like a bug that can lead to non-determinism, unless I am missing something!

src/lsm/table_memory.zig
sentientwaffle: These flags are not actually orthogonal. (They _usually_ are, but not strictly.)

At the end of `TableMemory.absorb()`, when we set `absorbed` to `true`:

```zig
            table_immutable.mutability = .{ .immutable = .{
                .flushed = table_immutable.value_context.count == 0,
                .absorbed = true,
                .snapshot_min = snapshot_min,
            } };
```

...it is possible that we set `flushed` to `true` at the same time, if all of the keys in the mutable table annihilate all of the keys in the immutable table.

src/lsm/compaction.zig
sentientwaffle: Ah, excellent catch! 

I don't think we have any tests that would catch this... the VOPR does test different batch sizes, but it leaves them static over the whole run.

src/lsm/compaction.zig
sentientwaffle: Yeah, I 100% agree that it would be nice to keep LSM unaware of VSR's checkpoint schedule.

> I wonder if it would be worth retaining this design by pulling the decision about checkpoint up, and pass flush_memory_table: bool parameter to the LSM? It feels like we might need something like that anyway once we get to snapshots?

I assume you mean as a parameter for `compact()`? Since `.compact()` is called for every beat, not just every bar/half-bar, it might make more sense to have a separate `.checkpoint_soon()` function which is called exactly one bar before `.checkpoint()`.

...That said, I am going to punt this to a TODO b/c it is a weekend/holiday.

sentientwaffle:
Ah, the value-count-max vs batch-limit is a little trickier than I first thought. :thinking:

src/lsm/table_memory.zig
```diff
@@ -163,37 +209,42 @@ pub fn TableMemoryType(comptime Table: type) type {
...
-                sort_values_by_key_in_ascending_order,
-            );
+            const target_count = sort_suffix_from_offset(table.values_used(), offset);
+            table.value_context.count = target_count;
+            table.mutability = .{ .mutable = .{ .suffix_offset = target_count } };
+        }
+
+        /// Returns the new length of `values`. (Values are deduplicated after sorting, so the
+        /// returned count may be less than or equal to the original `values.len`.)
+        fn sort_suffix_from_offset(values: []Value, offset: u32) u32 {
```
matklad, line 229: 👍 on passing values and offset directly, rather than `self`. I think that for hot data-plane methods that's the right choice, as it gives much more information to the optimizer about what can't change (and, also, just usually makes the code easier to follow, if a bit more verbose on the call site)

matklad (approved):

PR #2593, opened 2024-12-20
vsr: repair headers&prepares even when lagging behind the primary

matklad:
Our repair procedure have a condition where we don't initiate repair until our head op is at least commit_max. This feels suspicious to me, as it feels like this _could_ live-lock a replica, if the commit max advances faster than the op (via commit messages).

I _think_ this can be safely relaxed to requiring that the current op is from the current view --- if we have _some_ op from a view, that gives us a strong cryptographic proof tha everything that hash-chains to it is, in fact, a part of the view in question.

### DIFF
--- a/src/vsr/replica.zig
+++ b/src/vsr/replica.zig
@@ -6435,7 +6435,18 @@ pub fn ReplicaType(
                 }
             }
 
-            if (self.op < self.op_repair_max()) return;
+            if (self.op < self.op_repair_max()) {
+                const op_header_view = self.journal.header_with_op(self.op).?.view;
+                assert(op_header_view <= self.view);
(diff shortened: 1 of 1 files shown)


matklad:
That condition was bugging me for a long time, and today, while already enjoying the Christmas break, it occurred to me that we can relax it. 

I am not sure if this was contributing to the surprising behavior when we change replication to flip the direction along the replication ring, but it also feels related to that.

sentientwaffle (approved):
I think we could use similar logic to start committing before repairing all of the headers between commit_min and op_head!

matklad:
Yeah, I feel we have at least n+1 more places where we are not as good with repair as we could be, but lets get this one it!

PR #2595, opened 2024-12-22
devhub: track format and startup time 

cb22:
Add tracking of format time and startup time (with an 8gb grid cache) to the devhub, so we can see it come down :smile:.

Additionally, defer MessageBus.init until after the grid cache is done.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -178,6 +178,23 @@ jobs:
       - run: ./zig/download.${{ matrix.os == 'windows-latest' && 'bat' || 'sh' }}
       - run: ./zig/zig build scripts -- ci --language=${{ matrix.language }}
 
+  devhub_dry_run:
+    if: github.ref != 'refs/heads/main'
+    runs-on: ubuntu-22.04
+
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 2147483647
+      - run: sudo apt-get update && sudo apt-get install -y kcov
+      - run: ./zig/download.sh
+      # Run under sudo to enable memory locking for accurate RSS stats.
+      - run: sudo -E ./zig/zig build scripts -- devhub --sha=${{ github.sha }}
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          # Not providing DEVHUBDB_PAT and NYRKIO_TOKEN stops devhub from uploading its results, but
+          # it still runs everything.
+
   devhub:
     if: github.ref == 'refs/heads/main'
     runs-on: ubuntu-22.04
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -178,27 +178,9 @@ jobs:
       - run: ./zig/download.${{ matrix.os == 'windows-latest' && 'bat' || 'sh' }}
       - run: ./zig/zig build scripts -- ci --language=${{ matrix.language }}
 
-  devhub_dry_run:
-    if: github.ref != 'refs/heads/main'
-    runs-on: ubuntu-22.04
-
-    steps:
-      - uses: actions/checkout@v4
-        with:
-          fetch-depth: 2147483647
-      - run: sudo apt-get update && sudo apt-get install -y kcov
-      - run: ./zig/download.sh
-      # Run under sudo to enable memory locking for accurate RSS stats.
-      - run: sudo -E ./zig/zig build scripts -- devhub --sha=${{ github.sha }}
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-          # Not providing DEVHUBDB_PAT and NYRKIO_TOKEN stops devhub from uploading its results, but
-          # it still runs everything.
-
   devhub:
-    if: github.ref == 'refs/heads/main'
     runs-on: ubuntu-22.04
-    environment: devhub
+    environment: ${{ github.ref == 'refs/heads/main' && 'devhub' || '' }}
     permissions:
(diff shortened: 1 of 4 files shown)


matklad:
>Additionally, defer MessageBus.init until after the grid cache is done.

I am missing the motivation here. Not saying that this _shouldn't be done, just that it feels arbitrary to me one way or another, so I am wondering what I might be missing!

cb22:
> > Additionally, defer MessageBus.init until after the grid cache is done.
> 
> I am missing the motivation here. Not saying that this _shouldn't be done, just that it feels arbitrary to me one way or another, so I am wondering what I might be missing!

Oops, I meant to! Originally I was going to use this to measure the startup latency, which was a bit silly considering the port 0 trick and reading stdout is significantly easier.

I still think it's worthwhile to leave in - it brings the time when the replica can be externally spoken to (ie, MessageBus will accept TCP connections) closer to the time when replica is actually listening for messages.

Practically, that means things like external simple health checks can be more accurate, as can logs that say things like "connected to replica 0" - only for replica 0 to still be unavailable for a minute while the grid cache faults in.

matklad:
Worth a comment in the source!

Or perhaps even refactoring `init_tcp` into a public method, so it's more explicit _when_ we are ready to accept!

cb22:
> Or perhaps even refactoring init_tcp into a public method, so it's more explicit when we are ready to accept!

Holding off on that for now, since there's a long list of MessageBus related things I want to look at!

src/scripts/devhub.zig
```diff
@@ -129,6 +129,35 @@ fn devhub_metrics(shell: *Shell, cli_args: CLIArgs) !void {
...
+        var process = try shell.spawn(
+            .{
+                .stdin_behavior = .Pipe,
+                .stdout_behavior = .Pipe,
+                .stderr_behavior = .Inherit,
+            },
+            "./tigerbeetle start --addresses=0 --cache-grid=8GiB datafile",
+            .{},
+        );
+        errdefer _ = process.kill() catch unreachable;
```
matklad: Should be defer? Or we should close `process.stdin` on the happy path.

src/scripts/devhub.zig
```diff
@@ -129,6 +129,35 @@ fn devhub_metrics(shell: *Shell, cli_args: CLIArgs) !void {
...
+
+        var port_buf: [std.fmt.count("{}\n", .{std.math.maxInt(u16)})]u8 = undefined;
+        _ = try process.stdout.?.readAll(&port_buf);
+
```
matklad: Maybe also open a TCP connection to the port and write a ping message manually, just to check that the thing really is up? 

I _think_ waiting for the port to be bound is morally equivalent and should be fine, but, as a matter of principle, it still feels better to measure _directly_, rather than by proxy, and a direct measurement here is after sending a message.

.github/workflows/ci.yml
```diff
@@ -192,17 +191,31 @@ jobs:
           fetch-depth: 2147483647
       - run: sudo apt-get update && sudo apt-get install -y kcov
       - run: ./zig/download.sh
+
+      # Dummy devhub run - checks that all the devhub tests pass in CI. They are run again, in main
+      # once merged. Kcov is skipped to avoid adding to the pipeline time.
+      - run: sudo -E ./zig/zig build scripts -- devhub --sha=${{ github.sha }} --skip-kcov
```
cb22, line 197: It would be nice to not need both the `--skip-kcov` and the `--devhub` flag to release, but without them, devhub takes ~20 minutes in CI which would add to the total pipeline time...

src/vsr/replica.zig
```diff
@@ -5315,6 +5321,36 @@ pub fn ReplicaType(
                 return true;
             }
 
+            if (message.header.release.value < self.release_client_min.value) {
```
cb22, line 5324: We weren't dropping pings from older clients before!

src/scripts/devhub.zig
matklad: Still there^

src/scripts/devhub.zig
```diff
@@ -129,8 +135,76 @@ fn devhub_metrics(shell: *Shell, cli_args: CLIArgs) !void {
...
+        var writer = stream.writer();
+        try writer.writeAll(std.mem.asBytes(&ping)[0..@sizeOf(Header)]);
+
+        const reader = stream.reader();
+        _ = try reader.readAll(std.mem.asBytes(&eviction)[0..@sizeOf(Header)]);
+
+        std.debug.assert(eviction.command == .eviction);
+        std.debug.assert(eviction.valid_checksum());
+        std.debug.assert(eviction.valid_checksum_body(&[0]u8{}));
```
matklad: Let's consistently use unqualified `assert`

src/scripts/devhub.zig
cb22: Whoops!

matklad: Yeah, this one was the only one I _wanted_ to fix before merging 😅

src/scripts/release.zig
```diff
@@ -216,78 +229,90 @@ fn build(shell: *Shell, languages: LanguageSet, info: VersionInfo) !void {
...
     const targets = .{
         "x86_64-linux",
         "x86_64-windows",
         "aarch64-linux",
         "aarch64-macos", // Will build a universal binary.
     };
 
+    inline for (.{ true, false }) |debug| {
+        inline for (targets) |target| {
+            try build_tigerbeetle_target(shell, info, dist_dir, debug, target);
```
matklad, line 241: great refactor!

matklad (approved):

PR #2598, opened 2024-12-23
clients/go: `t` is not thread-safe

AskAlexSharov:
```
golangci-lint run .
tb_client_test.go:426:6: testinggoroutine: call to (*testing.T).Fatal from a non-test goroutine (govet)
					t.Fatal(err)
					^
tb_client_test.go:482:6: testinggoroutine: call to (*testing.T).Fatal from a non-test goroutine (govet)
					t.Fatal(err)
					^
```


matklad:
Fixed by https://github.com/tigerbeetle/tigerbeetle/pull/2639, thanks Alex!

PR #2600, opened 2024-12-23
vsr: free grid blocks when checkpoint becomes durable

chaitanyabhandari:
### Problem

Currently, when a replica checkpoints, it immediately *frees* all blocks released in the previous checkpoint. Consequently, the replica could immediately reacquire and overwrite those blocks, say as part of a compaction in the new checkpoint. This behavior is incorrect could lead to the cluster becoming unavailable, as it can compromise durability of the blocks are reacquired and overwritten. 

Specifically, consider a 3-replica cluster (R0, R1, and R2) where R0 and R1 are up, and R2 is down. Now, consider the following sequence of events:
* R0 and R1 are on checkpoint C<sub>n</sub>. Block B with checksum=X exists on both R0 and R1.
* R1 compacts (B, X), marks it as released.
* R1 transitions from C<sub>n</sub> → C<sub>n+1</sub>, freeing all blocks it released during C<sub>n</sub>, including (B, X).
* R1 reacquires and writes content with checksum=Y to block B, during a compaction in C<sub>n+1</sub>
* R0 gets a corruption at (B, X). 
* R0 wants to compact block B during a compaction in C<sub>n</sub>.
* R0 requests (B, X) it from R1, but R1 contains version (B, Y). 
* Cluster becomes unavailable.

Technically have 2 faults here, which is beyond what a 3-replica cluster can tolerate, so the cluster can remain correctly unavailable. *But we can do better*!

We even saw the impact of this behavior in [practice](https://tigerbeetleinc.slack.com/archives/C04THA39JGZ/p1735456479938959). Here, an idle cluster that had overwritten a block from the previous checkpoint wasn't able to help a lagging replica repair that block. Since the cluster was idle while its checkpoint was not considered durable (it hadn't committed up to (pipeline + 1)<sup>st</sup> prepare after checkpoint trigger), the lagging replica got stuck in an indefinite state sync.

### Solution

When a replica checkpoints, it must not immediately free all blocks in the previous checkpoint. Instead, the replica must *wait for the checkpoint to be become durable on a commit-quorum of replicas*. 

In the example discussed above, this would ensure that a single corruption on R1 doesn't render the entire cluster unavailable. Now, R1 would wait for C<sub>n+1</sub> to become durable on *both* R0 and R1 before it frees, reacquires and rewrites blocks released in C<sub>n</sub>.  Therefore, it becomes impossible for R0 to corrupt a block that it can't request from R1.

### Implementation Details

An intuitive way to reason about these changes is to think about the FreeSet block-acquisition and release flows during these three phases:
* At Checkpoint
* Pre Checkpoint Durability
* At Checkpoint Durability

#### At Checkpoint

At checkpoint, replica invokes `Grid.checkpoint()`, which encodes the `FreeSet.blocks_acquired` and `FreeSet.blocks_released` bitsets.  Once, the  new checkpoint is written to the superblock, replica invokes `FreeSet.mark_checkpoint_not_durable()`. 

#### Pre Checkpoint Durability

While a checkpoint is not durable, `FreeSet.blocks_released` bitset remains unchanged, and all block releases (from LSM forest compaction, ManifestLog compaction, FreeSet & Client Sessions checkpoint) are absorbed in the `FreeSet.blocks_released_prior_checkpoint_durability` hash map.  This ensures that the released blocks from the previous checkpoint are not reacquired till a checkpoint becomes durable. 

#### At Checkpoint Durability

At checkpoint durability, replica calls `Grid.checkpoint_durable()`, which first awaits all repair writes to the blocks released in the *previous checkpoint* (i.e. blocks in `FreeSet.blocks_released` bitset). Once done, it first invokes `FreeSet.free_released_blocks()` which flips the `FreeSet.blocks_released` bitset (freeing all released blocks). Then, it invokes `FreeSet.mark_checkpoint_durable()`, which moves all released blocks from `FreeSet.blocks_released_prior_checkpoint_durability` → `FreeSet.blocks_released`. 

To ensure storage determinism, each replica calls `Grid.checkpoint_durable()` at the same time, when it is committing the (pipeline + 1)<sup>st</sup> prepare after checkpoint trigger.


chaitanyabhandari:
Had added the commits from https://github.com/tigerbeetle/tigerbeetle/pull/2637 (which solves a couple of open VOPR false positives) to this PR, and we've been VOPR green for 3 days, 17 hours with these changes! I am removing those commits now.
Screenshot 2025-01-13 at 8 59 21 AM

src/vsr/grid.zig
```diff
@@ -500,19 +579,22 @@ pub fn GridType(comptime Storage: type) type {
...
         /// released.
         ///
-        /// This does not remove the block from the cache — the block can be read until the next
+        /// This does not remove the blocks from the cache — the blocks can be read until the next
         /// checkpoint.
         ///
-        /// Asserts that the address is not currently being read from or written to.
-        pub fn release(grid: *Grid, address: u64) void {
+        /// Asserts that the addresses are not currently being read from or written to.
+        pub fn release(grid: *Grid, addresses: []const u64) void {
```
sentientwaffle, line 604: Checking my understanding -- Is this just a API convenience change, since we usually release many blocks at a time?

src/vsr/superblock.zig
```diff
@@ -43,7 +43,7 @@ pub const Quorums = @import("superblock_quorums.zig").QuorumsType(.{
 
 pub const SuperBlockVersion: u16 =
     // Make sure that data files created by development builds are distinguished through version.
-    if (constants.config.process.release.value == vsr.Release.minimum.value) 0 else 1;
+    if (constants.config.process.release.value == vsr.Release.minimum.value) 1 else 2;
```
sentientwaffle: Is there some reason to change the development build superblock version?

```suggestion
    if (constants.config.process.release.value == vsr.Release.minimum.value) 0 else 2;
```

src/vsr/superblock_fuzz.zig
```diff
@@ -439,12 +447,17 @@ const Environment = struct {
...
                 .last_block_address = 0,
                 .trailer_size = 0,
                 .checksum = vsr.checksum(&.{}),
-            },
+            }, .blocks_released = .{
+                .last_block_checksum = 0,
+                .last_block_address = 0,
+                .trailer_size = 0,
+                .checksum = vsr.checksum(&.{}),
+            } },
```
sentientwaffle: Formatting wise, I think our convention for this is:

```
.free_set_references = .{
    .blocks_acquired = .{
        ...
    },
    .blocks_released = .{
        ...
    },
},
```

src/vsr/free_set.zig
```diff
@@ -42,23 +43,37 @@ pub const Reservation = struct {
 ///
 pub const FreeSet = struct {
     pub const Word = u64;
+    pub const BitsetKinds = enum {
```
sentientwaffle:
```suggestion
    pub const BitsetKind = enum {
```

(e.g. we have `vsr.Operation` and `vsr.Command`, not `vsr.Operations` and `vsr.Commands`)

src/vsr/free_set.zig
```diff
@@ -93,85 +108,127 @@ pub const FreeSet = struct {
...
 
-    pub fn init(allocator: mem.Allocator, blocks_count: usize) !FreeSet {
-        assert(blocks_count % shard_bits == 0);
-        assert(blocks_count % @bitSizeOf(Word) == 0);
+    pub fn init(allocator: mem.Allocator, options: struct {
+        blocks_count: usize,
+        blocks_released_prior_checkpoint_durability_max: usize,
+    }) !FreeSet {
+        assert(options.blocks_count % shard_bits == 0);
+        assert(options.blocks_count % @bitSizeOf(Word) == 0);
```
sentientwaffle, line 116:
```suggestion
        assert(options.blocks_count % @bitSizeOf(Word) == 0);
        assert(options.blocks_count > 0);
        assert(options.blocks_released_prior_checkpoint_durability_max > 0);
        assert(options.blocks_released_prior_checkpoint_durability_max < blocks_count);
```

(Just as a sanity-check.)

src/vsr/free_set.zig
```diff
@@ -93,85 +108,127 @@ pub const FreeSet = struct {
...
-    }) void {
+    /// Block addresses themselves are not a part of the encoded bitset for acquired blocks,
+    /// see CheckpointTrailer for details.
+    pub fn open(set: *FreeSet, options: struct { encoded: struct {
+        blocks_acquired: []const []align(@alignOf(Word)) const u8,
+        blocks_released: []const []align(@alignOf(Word)) const u8,
+    }, block_addresses: struct {
+        blocks_acquired: []const u64,
+        blocks_released: []const u64,
+    } }) void {
```
sentientwaffle: Format into

```suggestion
    pub fn open(set: *FreeSet, options: struct {
        encoded: struct {
            blocks_acquired: []const []align(@alignOf(Word)) const u8,
            blocks_released: []const []align(@alignOf(Word)) const u8,
        },
        block_addresses: struct {
            blocks_acquired: []const u64,
            blocks_released: []const u64,
        },
    }) void {
```

src/vsr/free_set.zig
```diff
@@ -93,85 +108,127 @@ pub const FreeSet = struct {
...
-    pub fn open(set: *FreeSet, options: struct {
-        encoded: []const []align(@alignOf(Word)) const u8,
-        block_addresses: []const u64,
-    }) void {
+    /// Block addresses themselves are not a part of the encoded bitset for acquired blocks,
+    /// see CheckpointTrailer for details.
+    pub fn open(set: *FreeSet, options: struct { encoded: struct {
+        blocks_acquired: []const []align(@alignOf(Word)) const u8,
+        blocks_released: []const []align(@alignOf(Word)) const u8,
+    }, block_addresses: struct {
```
sentientwaffle: Maybe rename `block_addresses` to `free_set_block_addresses`, to be clear what blocks we are referring to here?

src/vsr/free_set.zig
```diff
@@ -93,85 +108,127 @@ pub const FreeSet = struct {
...
     }
 
     // A shortcut to initialize and open an empty free set for tests.
     pub fn open_empty(allocator: mem.Allocator, blocks_count: usize) !FreeSet {
-        var set = try init(allocator, blocks_count);
-        set.open(.{ .encoded = &.{}, .block_addresses = &.{} });
+        var set = try init(allocator, .{
+            .blocks_count = blocks_count,
+            .blocks_released_prior_checkpoint_durability_max = 0,
+        });
```
sentientwaffle, line 235: Even though it isn't strictly necessary, we should `errdefer .. deinit` here.

src/vsr/free_set.zig
```diff
@@ -370,10 +429,26 @@ pub const FreeSet = struct {
...
+    /// becomes durable (when checkpoint_durable is set to `true`).
+    ///
+    /// Calling this function is only valid while the current checkpoint is not durable. During this
+    /// period, blocks are marked as released in `blocks_released_prior_checkpoint_durability`;
+    /// `blocks_released` remains unchanged and contains blocks released during the previous
+    /// checkpoint interval.
+    pub fn to_be_freed_at_checkpoint_durability(set: *const FreeSet, address: u64) bool {
+        assert(set.opened);
+        assert(!set.checkpoint_durable);
+        maybe(set.blocks_released_prior_checkpoint_durability.count() > 0);
```
sentientwaffle: Is there anything else we can assert here?

- Is the `address` always acquired at this point?
- Is the `address` never in `blocks_released_prior_checkpoint_durability`?

Where we can't assert, add `maybe()`s to still "document" the invariants (or lack thereof).

src/vsr/free_set.zig
```diff
@@ -382,11 +457,21 @@ pub const FreeSet = struct {
...
+
+        // `blocks_released` remains unchanged while the current checkpoint is not durable,
+        // since it contains blocks released in the previous checkpoint. These blocks must not be
+        // freed till the current checkpoint is durable, so as to maintain the durability of these
+        // blocks on a commit quorum of replicas.
+        if (!set.checkpoint_durable) {
+            set.blocks_released_prior_checkpoint_durability.putAssumeCapacity(block, {});
+        } else {
+            set.blocks_released.set(block);
+        }
```
sentientwaffle: Slightly simpler, to avoid the negation:

```suggestion
        if (set.checkpoint_durable) {
            set.blocks_released.set(block);
        } else {
            set.blocks_released_prior_checkpoint_durability.putAssumeCapacity(block, {});
        }
```

src/vsr/free_set.zig
```diff
@@ -407,114 +492,121 @@ pub const FreeSet = struct {
...
-            // Update the index when every block in the shard is allocated.
+            // Update the index when every block in the shard is acquired.
             if (set.find_free_block_in_shard(shard) == null) set.index.set(shard);
-            set.staging.set(block);
+
+            if (set.checkpoint_durable) {
+                set.blocks_released.set(block);
+            } else {
+                set.blocks_released_prior_checkpoint_durability.putAssumeCapacity(block, {});
+            }
```
sentientwaffle: Can we `assert(set.checkpoint_durable)` at the start of this function, and remove one of these branches?

src/vsr/free_set.zig
```diff
@@ -407,114 +492,121 @@ pub const FreeSet = struct {
...
+        for (source_chunks) |source_chunk| {
+            words_decoded += decoder.decode_chunk(source_chunk);
+        }
+        assert(words_decoded * @bitSizeOf(MaskInt) <= set.blocks_acquired.bit_length);
+        assert(decoder.done());
     }
 
-    /// Decodes the compressed bitset chunks in `source_chunks` into `set`.
-    /// Panics if the `source_chunks` encoding is invalid.
     pub fn decode_chunks(
```
sentientwaffle, line 637: Can this be private?

src/vsr/free_set.zig
```diff
@@ -407,114 +492,121 @@ pub const FreeSet = struct {
...
         set: *FreeSet,
-        source_chunks: []const []align(@alignOf(Word)) const u8,
+        source_chunks_blocks_acquired: []const []align(@alignOf(Word)) const u8,
+        source_chunks_blocks_released: []const []align(@alignOf(Word)) const u8,
     ) void {
```
sentientwaffle, line 641: Can we assert `!opened` and `checkpoint_durable`?

src/vsr/free_set.zig
```diff
@@ -407,114 +492,121 @@ pub const FreeSet = struct {
...
-            set.release(address);
-        }
     }
 
-    /// Temporarily marks staged blocks as free.
-    /// Amortizes the cost of toggling staged blocks when encoding and getting the highest address.
-    /// Does not update the index and MUST therefore be paired immediately with exclude_staging().
-    pub fn include_staging(set: *FreeSet) void {
+    pub fn free_released_blocks(set: *FreeSet) void {
         assert(set.opened);
```
sentientwaffle: Not sure whether we expect `checkpoint_durable` or `!checkpoint_durable`, but it seems like we should have one or the other?

src/vsr/free_set.zig
```diff
@@ -407,114 +492,121 @@ pub const FreeSet = struct {
...
-        while (it.next()) |block| {
-            set.staging.unset(block);
-            const address = block + 1;
-            set.release_now(address);
+    /// Marks the current checkpoint as durable and moves all released blocks from
+    /// `blocks_released_prior_checkpoint_durability` → `blocks_released`.
+    ///
+    /// Must be called after all released blocks are freed at checkpoint durability, via
+    /// `free_released_blocks`.
+    pub fn mark_checkpoint_durable(set: *FreeSet) void {
```
sentientwaffle, line 586: :thinking: Why not inline `free_released_blocks` into this?

src/vsr/free_set.zig
```diff
@@ -407,114 +492,121 @@ pub const FreeSet = struct {
...
+        assert(!set.checkpoint_durable);
+        assert(set.blocks_released.count() == 0);
+
+        set.checkpoint_durable = true;
+
+        // Block releases from the current checkpoint that were temporarily recorded in
+        // blocks_released_prior_checkpoint_durability can now be moved to blocks_released.
+        while (set.blocks_released_prior_checkpoint_durability.count() > 0) {
+            const block = set.blocks_released_prior_checkpoint_durability.keys()[0];
+            set.blocks_released_prior_checkpoint_durability.swapRemoveAt(0);
```
sentientwaffle: Could this be simplified to:

```suggestion
        while (set.blocks_released_prior_checkpoint_durability.popOrNull()) |block_entry| {
            const block = block_entry.key;
```

src/vsr/free_set.zig
```diff
@@ -407,114 +492,121 @@ pub const FreeSet = struct {
...
+            set.blocks_released.set(block);
         }
-        assert(set.staging.count() == 0);
+        assert(set.blocks_released_prior_checkpoint_durability.count() == 0);
+
         // Index verification is O(blocks.bit_length) so do it only at checkpoint, which is
         // also linear.
         set.verify_index();
```
sentientwaffle: Ah, nice catch moving this to the end, so that it is a proper postcondition!

src/vsr/free_set.zig
```diff
@@ -407,114 +492,121 @@ pub const FreeSet = struct {
...
-    pub fn exclude_staging(set: *FreeSet) void {
-        assert(set.opened);
-        const free = set.count_free();
+    /// Decodes the compressed bitset chunks in `source_chunks` into `target_bitset`.
+    /// Panics if the `source_chunks` encoding is invalid.
+    pub fn decode(
+        set: *FreeSet,
+        target_bitset: FreeSet.BitsetKinds,
+        source_chunks: []const []align(@alignOf(Word)) const u8,
+    ) void {
```
sentientwaffle, line 616: Can we assert `!opened` and `checkpoint_durable`?

src/vsr/free_set.zig
```diff
@@ -407,114 +492,121 @@ pub const FreeSet = struct {
...
-        assert(set.count_free() == free + set.staging.count());
+        assert(set.blocks_released.count() == 0);
     }
 
-    pub fn exclude_staging(set: *FreeSet) void {
-        assert(set.opened);
-        const free = set.count_free();
+    /// Decodes the compressed bitset chunks in `source_chunks` into `target_bitset`.
+    /// Panics if the `source_chunks` encoding is invalid.
+    pub fn decode(
```
sentientwaffle: I haven't looked into this at all, but why does storage checker use `decode()` (the lower-level API) instead of `decode_chunks`?

src/vsr/free_set_fuzz.zig
sentientwaffle, line 1: Could this also test `mark_checkpoint_durable`/`mark_checkpoint_not_durable`?

src/vsr/free_set.zig
```diff
@@ -901,24 +1019,25 @@ fn test_encode(patterns: []const TestPattern) !void {
...
 
-    decoded_actual.decode(encoded[0..encoded_length]);
+    decoded_actual.decode_chunks(&.{encoded[0..encoded_length]}, &.{});
     try expect_free_set_equal(decoded_expect, decoded_actual);
 }
 
 fn expect_free_set_equal(a: FreeSet, b: FreeSet) !void {
-    try expect_bit_set_equal(a.blocks, b.blocks);
+    try expect_bit_set_equal(a.blocks_acquired, b.blocks_acquired);
+    try expect_bit_set_equal(a.blocks_released, b.blocks_released);
```
sentientwaffle, line 1093: Also compare `blocks_released_prior_checkpoint_durability`?

sentientwaffle (commented):
Partial review -- mostly looked at freeset so far.

src/vsr/grid.zig
chaitanyabhandari: That's right! I saw that we were iterating over an array of blocks at multiple call sites, so I thought it'd be nice to change the API itself.

src/vsr/superblock.zig
chaitanyabhandari: My understanding was that we may want upgrade between multiple development releases, the same as production releases, but there might be a gap there! Do we not allow building a multi-version build with multiple development builds (as in, TB built in development mode over different commits)?

src/vsr/free_set.zig
chaitanyabhandari: Regarding `assert(options.blocks_released_prior_checkpoint_durability_max > 0);`: in our current design, we _do_ allow `blocks_released_prior_checkpoint_durability_max` to be 0. We use this in all fuzzers, and the free set tests as well (see [forest_fuzz.zig](https://github.com/tigerbeetle/tigerbeetle/pull/2600/files#diff-b3d58cbc7044eb17a665c94405868876c74dd1a6a9e5a5b5b4e3dd6cafcd5a17) for reference). The rationale for this is that using the `blocks_released_prior_checkpoint_durability` hash map is only important in the case of a cluster with multiple replicas.  checkpoints are trivially durable in a single-replica cluster. So, we explicitly mark `checkpoint_durable` as `True` in all tests and fuzzers, which ensures that they only use the `blocks_acquired` & `blocks_released` and go through the checkpoint durability flows.

https://github.com/tigerbeetle/tigerbeetle/blob/26cce437f6654783cd0dfeb27cdaacd7b8372752/src/lsm/forest_fuzz.zig#L200-L204 

Alternatively, we _could_ define `blocks_released_prior_checkpoint_durability_max` as a common utility which takes as input `[]const TreeInfo` (to calculate the number of tree compaction blocks released), `*const SuperBlock` (to calculate the number of encoded FreeSet blocks released), and `*const Pace` (to calculate the number of manifest log compaction blocks released), and have each fuzzer invoke it. We would also have to add a stage to the fuzzer to mark checkpoint as durable. But perhaps its worth the extra coverage?

Thoughts?

src/vsr/free_set.zig
chaitanyabhandari: Good point; added!

src/vsr/free_set.zig
chaitanyabhandari: Great point, the function indeed reads much better with more invariants documented. :)

src/vsr/free_set.zig
chaitanyabhandari: Chasing this down made me realize that there's a bug in some of the storage checker changes in this PR... Not sure how this code will look when I fix that bug, but I'll keep this comment in mind! Thank you for the orthogonal nudge. :P

src/vsr/free_set.zig
chaitanyabhandari: Ah, it can indeed be simplified to `popOrNull`, great point!

I used `swapRemoveAt` because the documentation said that it was the O(1) alternative to remove keys from the hash map. Checked out `popOrNull`, it is just as efficient as `swapRemoveAt`, they both invoke `removeFromIndexByIndexGeneric`.

src/vsr/free_set.zig
chaitanyabhandari: Another option here is to call `mark_checkpoint_not_durable()` and then immediately call `mark_checkpoint_durable` to signal that for single-replica clusters (or in this case, fuzzers that run in a single process), checkpoints can be immediately marked durable. 

https://github.com/tigerbeetle/tigerbeetle/blob/26cce437f6654783cd0dfeb27cdaacd7b8372752/src/lsm/forest_fuzz.zig#L320

I had taken up this approach in an earlier draft of the PR, but moved away from it for the sake of simplicity of tests & fuzzers. However, it did come at a cost, I had to separate out `free_released_blocks` instead of inlining it in `mark_checkpoint_durable` (which you highlighted in https://github.com/tigerbeetle/tigerbeetle/pull/2600#discussion_r1915637648). 

One drawback of this approach over the one that I described in the above comment is that we wouldn't really be increasing coverage too much over the current version. While we would call `mark_checkpoint_not_durable`, it would work on an empty `blocks_released_prior_checkpoint_durability`, so essentially, all releases would be recorded in `blocks_released`.

I believe the version that we ultimately go for depends on whether we want the fuzzers to _exactly_ mirror the production flow, i.e. whether we want to go through the exact flow where `blocks_released_prior_checkpoint_durability` absorbs all releases from checkpoint trigger → (checkpoint_trigger + pipeline + 1), and _then_ we call `mark_checkpoint_durable`...

src/vsr/superblock.zig
sentientwaffle: Development releases are always built with `release == vsr.Release.minimum == 0.0.1`.  If you want to test multiversion/upgrades, then you need to build it with a real release number.

chaitanyabhandari: Ack, understood! Apologies for not spending a second before asking this question, I should've figured this out on my own. 🤦🏼

sentientwaffle: No worried; it is probably undocumented. :stuck_out_tongue:

src/vsr/checkpoint_trailer.zig
```diff
@@ -109,9 +109,9 @@ pub fn CheckpointTrailerType(comptime Storage: type) type {
             trailer_type: TrailerType,
             buffer_size: usize,
         ) !CheckpointTrailer {
-            const block_count_max = stdx.div_ceil(buffer_size, chunk_size_max);
+            const block_count_max_ = stdx.div_ceil(buffer_size, chunk_size_max);
```
sentientwaffle: This could invoke `block_count_max()`.

`CheckpointTrailer.block_count()` could call `block_count_max()` also.

src/vsr/free_set.zig
sentientwaffle: > So, we explicitly mark checkpoint_durable as True in all tests and fuzzers, which ensures that they only use the blocks_acquired & blocks_released and go through the checkpoint durability flows.

That makes sense, thank you!

> Alternatively, we could define blocks_released_prior_checkpoint_durability_max as a common utility which takes as input []const TreeInfo (to calculate the number of tree compaction blocks released), *const SuperBlock (to calculate the number of encoded FreeSet blocks released), and *const Pace (to calculate the number of manifest log compaction blocks released), and have each fuzzer invoke it. We would also have to add a stage to the fuzzer to mark checkpoint as durable. But perhaps its worth the extra coverage?

I think extracting the function is a good idea, whether or not we use it in the fuzzers.

- I'm not sure where the best place for it is, since it depends on everything! Maybe `vsr.zig`? It is most closely related to the freeset, but the freeset doesn't know about LSM, which is nice.
- Instead of taking `superblock`, it should take `storage_size_limit: u64`. No need to depend on all of superblock when that's the only field we need.

---

I think we can leave most fuzzers as-is (i.e. assuming that the checkpoint is always durable). So then the assert could be:

```
assert(options.blocks_released_prior_checkpoint_durability_max > 0 or
    constants.verify);
```

If possible, it would be great if the forest fuzzer simulated checkpoint durability.

Right now we don't really test the `blocks_released_prior_checkpoint_durability`'s capacity. (i.e. testing that we don't ever overflow the capacity.) IIUC the VOPR is the only test that puts any blocks in it, and right now my guess is the VOPR doesn't test LSM deeply enough to approach that limit. (We should ratchet that up in the future, but the forest fuzzer will always have an advantage for testing of large LSM's because it doesn't have to deal with replication overhead.)

src/lsm/tree.zig
```diff
@@ -285,6 +285,10 @@ pub fn TreeType(comptime TreeTable: type, comptime Storage: type) type {
             return Table.layout.block_value_count_max;
         }
 
+        pub fn data_block_count_max(_: *const Tree) u32 {
+            return Table.layout.data_block_count_max;
+        }
```
sentientwaffle: I think I'd just inline this at the callsite.

src/vsr/grid.zig
```diff
@@ -1231,8 +1324,33 @@ pub fn GridType(comptime Storage: type) type {
...
+                            // only *after* the checkpoint is written to the superblock.
+                            assert(grid.superblock.staging.parent_checkpoint_id() ==
+                                grid.superblock.working.checkpoint_id());
+                            assert(grid.free_set.checkpoint_durable);
+                            break :checkpoint_durable false;
+                        }
+                    },
+                    address,
+                    checksum,
+                );
```
sentientwaffle: :exploding_head: Excellent catch -- I hadn't considered this case! So a block might be coherent with the new checkpoint before it is durable, but then when the checkpoint transitions to durable that block is no longer allowed b/c it was freed (via `blocks_released_prior_checkpoint_durability`).

Could we additionally assert: If a block is coherent with the durable checkpoint, then the block is also coherent with the corresponding non-durable checkpoint? That would give the grid checker a better chance to catch violations during non-durable checkpoints, since checkpoints are non-durable for a relatively short time. (We could implement it in the grid checker itself.)

src/vsr.zig
```diff
@@ -68,7 +69,7 @@ pub const CountingAllocator = @import("counting_allocator.zig");
 
 /// The version of our Viewstamped Replication protocol in use, including customizations.
 /// For backwards compatibility through breaking changes (e.g. upgrading checksums/ciphers).
-pub const Version: u16 = 0;
```
sentientwaffle: Per discussion w/ @chaitanyabhandari, instead of bumping the protocol version, we will do a staged rollout in 3 releases:

1. Before releasing this PR, release a change with a new noop start view command.
2. Release this PR. (This PR will support both the old and new start view commands.)
3. Remove the old start view command.

Since the upgrades set `visit=true`, and the upgrade quorum is equal to `replica_count`, even clusters that are lagging behind multiple releases will still upgrade correctly.

src/vsr/grid.zig
```diff
@@ -208,22 +212,29 @@ pub fn GridType(comptime Storage: type) type {
...
+            errdefer free_set_checkpoint_blocks_acquired.deinit(allocator);
+
+            var free_set_checkpoint_blocks_released = try CheckpointTrailer.init(
                 allocator,
                 .free_set,
-                FreeSet.encode_size_max(block_count_limit),
+                FreeSet.encode_size_max(blocks_count),
             );
-            errdefer free_set_checkpoint.deinit(allocator);
+            errdefer free_set_checkpoint_blocks_released.deinit(allocator);
```
sentientwaffle: Maybe neater as:

```zig
            const free_size_size = FreeSet.encode_size_max(blocks_count);
            var free_set_checkpoint_blocks_acquired =
                try CheckpointTrailer.init(allocator, .free_set, free_set_size);
            errdefer free_set_checkpoint_blocks_acquired.deinit(allocator);

            var free_set_checkpoint_blocks_released =
                try CheckpointTrailer.init(allocator, .free_set, free_set_size);
            errdefer free_set_checkpoint_blocks_released.deinit(allocator);
```

src/vsr/grid.zig
```diff
@@ -282,32 +296,64 @@ pub fn GridType(comptime Storage: type) type {
...
+        }
+
+        fn open_free_set_callback_blocks_released(trailer: *CheckpointTrailer) void {
+            assert(trailer.callback == .none);
+            const grid: *Grid = @fieldParentPtr("free_set_checkpoint_blocks_released", trailer);
+            grid.open_free_set_callback();
+        }
+
+        fn open_free_set_callback(grid: *Grid) void {
             const callback = grid.callback.open;
```
sentientwaffle, line 322: Since `open_free_set_callback` was invoked, we know that at least one of the trailers is done -- we should assert that!

src/vsr/grid.zig
```diff
@@ -282,32 +296,64 @@ pub fn GridType(comptime Storage: type) type {
...
+                            .block_addresses[0..block_count_encoded_blocks_acquired],
+                        .blocks_released = grid.free_set_checkpoint_blocks_released
+                            .block_addresses[0..block_count_encoded_blocks_released],
+                    },
                 });
-                assert((grid.free_set.count_acquired() > 0) == (free_set_checkpoint.size > 0));
+                assert((grid.free_set.count_acquired() > 0) ==
+                    (grid.free_set_checkpoint_blocks_acquired.size > 0));
                 assert(grid.free_set.count_reservations() == 0);
-                assert(grid.free_set.count_released() == grid.free_set_checkpoint.block_count());
```
sentientwaffle, line 308: We can still assert:
```zig
assert(grid.free_set.count_released() >= grid.free_set_checkpoint.block_count());
```
(Note that `>=` rather than `==`.)

src/vsr/grid.zig
chaitanyabhandari: Indeed, _much_ neater! <3

src/vsr/free_set.zig
chaitanyabhandari: Discussed in https://github.com/tigerbeetle/tigerbeetle/pull/2600#discussion_r1915742483, we can consider inlining `free_released_blocks` into `mark_checkpoint_durable` but that would then entail calling `mark_checkpoint_not_durable`  & `mark_checkpoint_durable` in all fuzzers. I had gone down this route before, but it seemed simpler to separate these out, but I feel more and more now that my API design is being driven by how the tests consume it, which seems incorrect. 

I will push a version where `free_released_blocks` is inlined! It will also resolve https://github.com/tigerbeetle/tigerbeetle/pull/2600#discussion_r1915632495.

chaitanyabhandari:
Addressed _most_ comments in c5d530acfba41ac197d411faf9379880a52caefe. There are two open ones that I am yet to address:
* Testing the production checkpoint durability flow in the forest fuzzer (https://github.com/tigerbeetle/tigerbeetle/pull/2600#discussion_r1916997129)
* Additional coherency-related asserts (https://github.com/tigerbeetle/tigerbeetle/pull/2600#discussion_r1917336383)

src/vsr/free_set.zig
chaitanyabhandari: Addressed in b8ea99a0a1deb379a45c5f55cbccdc459a00d09b, storage checker uses `decode_chunks` instead of `decode`.

chaitanyabhandari:
Addressed _most_ comments in b8ea99a0a1deb379a45c5f55cbccdc459a00d09b. There are two open ones that I am yet to address:
* Testing the production checkpoint durability flow in the forest fuzzer (https://github.com/tigerbeetle/tigerbeetle/pull/2600#discussion_r1916997129)
* Additional coherency-related asserts (https://github.com/tigerbeetle/tigerbeetle/pull/2600#discussion_r1917336383)

src/lsm/manifest_log_fuzz.zig
```diff
@@ -153,7 +156,7 @@ fn generate_events(
         // All of the trees we are inserting/modifying have the same id (for simplicity), but we
         // want to perform more updates if there are more trees, to better simulate a real state
         // machine.
-        for (manifest_log_options.tree_id_min..manifest_log_options.tree_id_max + 1) |_| {
+        for (0..20) |_| {
```
sentientwaffle: Ensure that we don't specify this constant in multiple places, so that they can't accidentally mismatch. Either extract `const tree_count = 20` to a constant (and use it when initializing the `Pace`) or add `tree_count` as a field to the pace and reuse that here.

src/testing/cluster/storage_checker.zig
```diff
@@ -71,18 +72,33 @@ pub const StorageChecker = struct {
...
         errdefer checkpoints.deinit();
 
-        var free_set = try vsr.FreeSet.init(allocator, Storage.grid_blocks_max);
+        const free_set_encoded_blocks_max =
+            CheckpointTrailer.block_count_max(vsr.FreeSet.encode_size_max(Storage.grid_blocks_max));
+        var free_set = try vsr.FreeSet.init(
+            allocator,
+            .{
+                .blocks_count = Storage.grid_blocks_max,
+                .blocks_released_prior_checkpoint_durability_max = 2 * free_set_encoded_blocks_max,
```
sentientwaffle: I don't understand why the option is set to this value.

src/testing/cluster/storage_checker.zig
```diff
@@ -71,18 +72,33 @@ pub const StorageChecker = struct {
...
+        );
+        errdefer allocator.free(free_set_blocks_acquired_encoded);
+
+        const free_set_blocks_released_encoded = try allocator.alignedAlloc(
             u8,
             @alignOf(u64),
             vsr.FreeSet.encode_size_max(Storage.grid_blocks_max),
         );
-        errdefer allocator.free(free_set_buffer);
+        errdefer allocator.free(free_set_blocks_released_encoded);
```
sentientwaffle: Formatting nit:

```zig
        const free_set_size = vsr.FreeSet.encode_size_max(Storage.grid_blocks_max);
        ...
        const free_set_blocks_acquired_encoded =
            try allocator.alignedAlloc(u8, @alignOf(u64), free_set_size);
        errdefer allocator.free(free_set_blocks_acquired_encoded);

        const free_set_blocks_released_encoded =
            try allocator.alignedAlloc(u8, @alignOf(u64), free_set_size);
        errdefer allocator.free(free_set_buffer);
```
(`free_set_size` can be defined even earlier and reused when computing `free_set_encoded_blocks_max` too.)

src/testing/cluster/storage_checker.zig
```diff
@@ -203,9 +207,11 @@ pub const StorageChecker = struct {
     fn check(
         checker: *StorageChecker,
         caller: []const u8,
-        superblock: *const SuperBlock,
+        comptime Replica: type,
+        replica: *const Replica,
```
sentientwaffle: In the interest of "minimize dependencies", instead of depending on `replica`, could we narrow this to just `forest`? From the `forest` we can get both the `grid` and the `superblock`.

src/testing/cluster/storage_checker.zig
```diff
@@ -320,20 +325,26 @@ pub const StorageChecker = struct {
         return checksum.checksum();
     }
 
-    fn checksum_grid(
+    fn decode_free_set_from_superblock(
         checker: *StorageChecker,
+        bitset: vsr.FreeSet.BitsetKind,
         superblock: *const SuperBlock,
```
sentientwaffle: Nit, but I'd swap the parameter order here, since it is called twice with the same superblock.

src/testing/cluster/storage_checker.zig
```diff
@@ -358,41 +369,82 @@ pub const StorageChecker = struct {
...
 
-        checker.free_set.decode(checker.free_set_buffer[0..free_set_size]);
+        assert(vsr.checksum(free_set_buffer[0..free_set_size]) == free_set_checksum);
+    }
+
+    fn checksum_grid(
+        checker: *StorageChecker,
+        comptime Replica: type,
+        replica: *const Replica,
+        use_free_set_from_superblock: bool,
```
sentientwaffle: I'd change this to an inline `enum { free_set_from_disk, free_set_from_memory }` so that it is more descriptive at the call site.

src/tigerbeetle/inspect.zig
```diff
@@ -558,77 +558,157 @@ const Inspector = struct {
         try print_reply_body(output, reply);
     }
 
-    fn inspect_grid(inspector: *Inspector, output: std.io.AnyWriter, superblock_copy: ?u8) !void {
-        const superblock = try inspector.read_superblock(superblock_copy);
-
+    fn decode_free_set_from_superblock(
```
sentientwaffle: Since this isn't actually doing the free set decoding, I would rename it to `read_free_set_bitset`, and then move it down between `read_block` and `read_client_sessions`.

src/tigerbeetle/inspect.zig
```diff
@@ -558,77 +558,157 @@ const Inspector = struct {
...
+            superblock.vsr_state.checkpoint.free_set_blocks_acquired_size,
+        );
+        defer inspector.allocator.free(free_set_blocks_acquired_buffer);
+
+        var free_set_blocks_acquired_addresses =
+            try std.ArrayList(u64).initCapacity(inspector.allocator, stdx.div_ceil(
+            superblock.vsr_state.checkpoint.free_set_blocks_acquired_size,
+            constants.block_size - @sizeOf(vsr.Header),
+        ));
+        defer free_set_blocks_acquired_addresses.deinit();
```
sentientwaffle: This indentation (and for the corresponding "released" and "buffer" above/below) is confusing.. I think I'd expand it out to

```suggestion
        var free_set_blocks_acquired_addresses = try std.ArrayList(u64).initCapacity(
            inspector.allocator,
            stdx.div_ceil(
                superblock.vsr_state.checkpoint.free_set_blocks_acquired_size,
                constants.block_size - @sizeOf(vsr.Header),
            ),
        );
        defer free_set_blocks_acquired_addresses.deinit();
```

... even though that is more verbose. Or check if extracting `superblock.vsr_state.checkpoint.free_set_blocks_acquired_size` out to variables makes it more compact overall.

src/vsr/checkpoint_trailer.zig
```diff
@@ -163,7 +162,11 @@ pub fn CheckpointTrailerType(comptime Storage: type) type {
         }
 
         pub fn block_count(trailer: *const CheckpointTrailer) u32 {
-            return @intCast(stdx.div_ceil(trailer.size, chunk_size_max));
+            return block_count_max(trailer.size);
+        }
+
+        pub fn block_count_max(trailer_size: u64) u32 {
+            return @intCast(stdx.div_ceil(trailer_size, chunk_size_max));
         }
```
sentientwaffle: Hmm, actually maybe we are conflating "block count" and "blocks max". Maybe `block_count_max` should be named `block_count_for_trailer_size`, since it only actually computes the "blocks max" if it is passed the upper bound trailer size.

src/tigerbeetle/inspect.zig
```diff
@@ -644,17 +724,11 @@ const Inspector = struct {
                 free_set.count_acquired(),
                 free_set.count_released(),
                 free_set.highest_address_acquired(),
-                std.fmt.fmtIntSizeBin(superblock.vsr_state.checkpoint.free_set_size),
+                std.fmt.fmtIntSizeBin(superblock.vsr_state.checkpoint
+                    .free_set_blocks_acquired_size),
```
sentientwaffle, line 686: (As a separate field) we should also log the size and compression ratio of the "released" bitset. Rename the existing `size` and `compression_ratio` labels to `acquired_size` and `acquired_compression_ratio`.

src/vsr.zig
```diff
@@ -1693,3 +1695,63 @@ pub const Snapshot = struct {
...
+    const compaction_blocks_released_half_bar_max =
+        stdx.div_ceil(constants.lsm_levels, 2) *
+        compaction_input_blocks_forest_max;
+
+    const compaction_blocks_released_pipeline_max =
+        (pipeline_half_bars * compaction_blocks_released_half_bar_max) +
+        // Compaction is paced across all beats, so if a pipeline is less than half a bar,
+        // for simplicity, use the upper bound for a half a bar (treating pacing as
+        // imperfect).
+        @intFromBool(pipeline_half_bars == 0) * compaction_blocks_released_half_bar_max;
```
sentientwaffle: I think the first comment was wrong -- it said "blocks released within a pipeline" but that variable only looks at (one level of) a half-bar.

Also, even though multiplying by levels on the outside is equivalent, I think it would be clearer to multiple all at once, to give `compaction_input_blocks_forest_max` a more intuitive meaning. ("The maximum number of blocks released by running exactly one compaction per tree in the forest" is more confusing then "The maximum number of blocks released by all compaction during a half-bar.").

And a nit -- when `var` is needed to construct a constant, it is nice to hide it in a block it to scope the mutability.

```suggestion
    // Maximum number of blocks released within a single half-bar by compaction.
    const compaction_blocks_released_half_bar_max = blocks: {
        var blocks: usize = 0;
        inline for (Forest.tree_infos) |tree_info| {
            blocks +=
                stdx.div_ceil(constants.lsm_levels, 2) *
                compaction_input_tables_max *
                (1 + tree_info.Tree.Table.layout.data_block_count_max);
        }
        break :blocks blocks;
    };

    const compaction_blocks_released_pipeline_max =
        (pipeline_half_bars * compaction_blocks_released_half_bar_max) +
        // Compaction is paced across all beats, so if a pipeline is less than half a bar,
        // for simplicity, use the upper bound for a half a bar (treating pacing as
        // imperfect).
        @intFromBool(pipeline_half_bars == 0) * compaction_blocks_released_half_bar_max;
```

src/vsr.zig
```diff
@@ -1693,3 +1695,63 @@ pub const Snapshot = struct {
...
+    // Maximum number of blocks released within a pipeline by ManifestLog compactions.
+    const manifest_log_blocks_released_pipeline_max =
+        pipeline_half_bars * Forest.manifest_log_blocks_released_half_bar_max;
+
+    const block_count_max_grid = block_count_max(storage_size_limit);
+
+    return compaction_blocks_released_pipeline_max +
+        manifest_log_blocks_released_pipeline_max +
+        CheckpointTrailer.block_count_max(ClientSessions.encode_size) +
+        // We checkpoint both `blocks_acquired` and `blocks_released` bitsets in FreeSet.
```
sentientwaffle: I think this comment probably isn't required.

src/vsr/free_set.zig
```diff
@@ -93,85 +108,149 @@ pub const FreeSet = struct {
...
+    pub fn init(allocator: mem.Allocator, options: struct {
+        blocks_count: usize,
+        blocks_released_prior_checkpoint_durability_max: usize,
+    }) !FreeSet {
+        assert(options.blocks_count % shard_bits == 0);
+        assert(options.blocks_count % @bitSizeOf(Word) == 0);
+        assert(options.blocks_released_prior_checkpoint_durability_max > 0 or
+            constants.verify);
+        assert(options.blocks_released_prior_checkpoint_durability_max < options.blocks_count or
+            constants.verify);
```
sentientwaffle: Just curious, but what test winds up with `options.blocks_released_prior_checkpoint_durability_max >= options.blocks_count`?

sentientwaffle: I know I suggested these asserts, but now I think we should remove them. :sweat_smile: 
Asserts that are effectively gated on `constants.verify` _not_ being set are dangerous, since we always set `constants.verify` during tests. If we had a better way to gate on fuzzers in particular then we could do that instead, but I'm not aware of one.

src/vsr/free_set.zig
```diff
@@ -93,85 +108,149 @@ pub const FreeSet = struct {
...
+            options.encoded.blocks_acquired,
+            options.encoded.blocks_released,
+        );
+        set.mark_released(options.free_set_block_addresses.blocks_acquired);
+        set.mark_released(options.free_set_block_addresses.blocks_released);
         set.opened = true;
     }
 
+    // A shortcut to initialize an empty free set for tests.
+    pub fn init_empty(allocator: mem.Allocator, blocks_count: usize) !FreeSet {
```
sentientwaffle, line 216: Since this is only used in tests, we should `comptime assert(constants.verify);` (Its comptime, so that doesn't contradict my comment in `init()` :sweat_smile:).

(Likewise in `open_init()`.)

src/vsr/free_set.zig
```diff
@@ -407,114 +521,118 @@ pub const FreeSet = struct {
...
-    fn release_now(set: *FreeSet, address: u64) void {
+    /// Given the address, marks an acquired block as free.
+    fn free(set: *FreeSet, address: u64) void {
         assert(set.opened);
+
         const block = address - 1;
-        assert(set.blocks.isSet(block));
-        assert(!set.staging.isSet(block));
+        assert(set.blocks_acquired.isSet(block));
+        assert(set.blocks_released.isSet(block));
```
sentientwaffle, line 563: Assert that `!blocks_released_prior_checkpoint_durability.contains(block)`. (Unless we can assert that it is empty, in which case this would be redundant.)

src/vsr/free_set.zig
```diff
@@ -407,114 +521,118 @@ pub const FreeSet = struct {
...
+
+            set.blocks_released_prior_checkpoint_durability.putAssumeCapacity(block, {});
         }
     }
 
-    /// Given the address, marks an allocated block as free.
-    fn release_now(set: *FreeSet, address: u64) void {
+    /// Given the address, marks an acquired block as free.
+    fn free(set: *FreeSet, address: u64) void {
         assert(set.opened);
```
sentientwaffle, line 558: Can we assert that `checkpoint_durable` as well, or does that break too many tests?

As an alternative, could we assert that `blocks_released_prior_checkpoint_durability` is empty?

src/vsr/free_set.zig
```diff
@@ -407,114 +521,118 @@ pub const FreeSet = struct {
...
+        assert(set.index.count() == 0);
+        assert(set.blocks_acquired.count() == 0);
+        assert(set.blocks_released.count() == 0);
+        assert(set.blocks_released_prior_checkpoint_durability.count() == 0);
+
+        assert(set.reservation_count == 0);
+        assert(set.reservation_blocks == 0);
+
+        set.decode(.blocks_acquired, source_chunks_blocks_acquired);
+        set.decode(.blocks_released, source_chunks_blocks_released);
```
sentientwaffle, line 655: At the very end of this function, we should `set.verify_index();`.

src/vsr/grid.zig
```diff
@@ -317,63 +367,95 @@ pub fn GridType(comptime Storage: type) type {
...
-        ///   6. Mark currently released blocks as free and eligible for acquisition in the next
-        ///      checkpoint.
-        ///   7. Mark the free set's own blocks as released (but not yet free).
+        ///   5. Mark the free set's own blocks as released (but not yet free).
         ///
-        /// This function handles step 1 and 5.
-        /// This function calls `free_set_checkpoint.checkpoint`, which handles steps 2-4.
-        /// The caller is responsible for calling FreeSet.checkpoint which handles 6 and 7.
+        /// This function handles step 1, and calls `CheckpointTrailer.checkpoint`, which handles
+        /// 2-5.
```
sentientwaffle: 5 is the caller's responsibility, right? (via calling `mark_checkpoint_not_durable()`)

src/vsr/grid.zig
```diff
@@ -317,63 +367,95 @@ pub fn GridType(comptime Storage: type) type {
...
+
+            const callback = grid.callback.checkpoint;
+            grid.callback = .none;
+            callback(grid);
+        }
+
+        ///   Now that the checkpoint is durable on a commit quorum of replicas:
+        ///   1. Await all pending repair-writes to blocks that were just freed. This guarantees
+        ///      that there are no outstanding writes to (now-)free blocks.
+        ///   2. Mark currently released blocks as free and eligible for acquisition.
```
sentientwaffle:
```suggestion
        /// Now that the checkpoint is durable on a commit quorum of replicas:
        /// 1. Await all pending repair-writes to blocks that were just freed. This guarantees
        ///    that there are no outstanding writes to (now-)free blocks.
        /// 2. Mark currently released blocks as free and eligible for acquisition.
```

src/vsr/grid.zig
```diff
@@ -317,63 +367,95 @@ pub fn GridType(comptime Storage: type) type {
...
+            }
+            assert(grid.free_set_checkpoint_blocks_released.callback == .none);
+
+            const callback = grid.callback.checkpoint;
+            grid.callback = .none;
+            callback(grid);
+        }
+
+        ///   Now that the checkpoint is durable on a commit quorum of replicas:
+        ///   1. Await all pending repair-writes to blocks that were just freed. This guarantees
```
sentientwaffle: Instead of "just freed" are they "about to be freed"? Since we haven't called `mark_checkpoint_durable` yet.

src/vsr/grid.zig
```diff
@@ -1218,6 +1306,14 @@ pub fn GridType(comptime Storage: type) type {
             }
         }
 
+        pub fn block_count_max(superblock: *const SuperBlock) usize {
+            const shard_count_limit: usize = @intCast(@divFloor(
+                superblock.storage_size_limit - vsr.superblock.data_file_size_min,
+                constants.block_size * FreeSet.shard_bits,
+            ));
+            return shard_count_limit * FreeSet.shard_bits;
+        }
```
sentientwaffle: Can this be removed now in favor of `vsr.block_count_max()`?

src/vsr/free_set.zig
```diff
@@ -370,10 +450,32 @@ pub const FreeSet = struct {
...
+    pub fn to_be_freed_at_checkpoint_durability(set: *const FreeSet, address: u64) bool {
+        const block = address - 1;
+
+        assert(set.opened);
+        assert(!set.checkpoint_durable);
+
+        // Block address must be acquired, but is not necessarily released.
+        assert(set.blocks_acquired.isSet(block));
+        maybe(set.blocks_released.isSet(block));
+        maybe(set.blocks_released_prior_checkpoint_durability.contains(block));
```
sentientwaffle, line 490: Are these mutually exclusive? i.e. is this true:

```zig
        assert(!set.blocks_released.isSet(block) or
            !set.blocks_released_prior_checkpoint_durability.contains(block));
```

src/vsr/replica.zig
```diff
@@ -4272,11 +4273,41 @@ pub fn ReplicaType(
...
+                self.grid_scrubber.checkpoint_durable();
+                self.grid.checkpoint_durable(commit_checkpoint_durable_grid_callback);
+            } else {
+                self.state_machine.compact(
+                    commit_compact_callback,
+                    self.commit_prepare.?.header.op,
+                );
+            }
+
+            return .pending;
```
sentientwaffle: I think we should instead add a new commit dispatch stage `checkpoint_durable` which returns `.ready` most of the time, and returns `pending` when it actually runs. (Similar to how `commit_checkpoint_data` works.)

src/vsr/replica.zig
```diff
@@ -4303,11 +4334,14 @@ pub fn ReplicaType(
             assert(op <= self.op);
             assert((op + 1) % constants.lsm_compaction_ops == 0);
             log.info("{}: commit_checkpoint_data: checkpoint_data start " ++
-                "(op={} current_checkpoint={} next_checkpoint={})", .{
+                "(op={} current_checkpoint={} next_checkpoint={} free_set.acquired={} " ++
+                "free_set.released={}", .{
```
sentientwaffle: (Missing close-paren, and also may as well balance the lines a little.)
```suggestion
                "(op={} current_checkpoint={} next_checkpoint={} " ++
                "free_set.acquired={} free_set.released={})", .{
```

src/vsr/replica.zig
```diff
@@ -4556,14 +4578,34 @@ pub fn ReplicaType(
...
+                .block_addresses[0..self.grid.free_set_checkpoint_blocks_acquired.block_count()]);
+            self.grid.release(self.grid.free_set_checkpoint_blocks_released
+                .block_addresses[0..self.grid.free_set_checkpoint_blocks_released.block_count()]);
+
+            assert(self.grid.free_set.count_released() >=
+                self.grid.free_set_checkpoint_blocks_acquired.block_count() +
+                self.grid.free_set_checkpoint_blocks_released.block_count() +
+                self.client_sessions_checkpoint.block_count());
+
             // Send prepare_oks that may have been wittheld by virtue of `op_prepare_ok_max`.
```
```suggestion
            // Send prepare_oks that may have been withheld by virtue of `op_prepare_ok_max`.
```

src/vsr/replica_test.zig
```diff
@@ -1473,9 +1473,19 @@ test "Cluster: scrub: background scrubber, fully corrupt grid" {
         if (a0_free_set.is_free(address)) {
             assert(b2_free_set.is_free(address));
             assert(b2_storage.area_faulty(.{ .grid = .{ .address = address } }));
-        } else {
-            assert(!b2_free_set.is_free(address));
+        } else if (!a0_free_set.is_released(address)) {
+            // Acquired (but not released) blocks are guaranteed to be repaired by the scrubber.
+            assert(!b2_free_set.is_free(address) and !b2_free_set.is_released(address));
```
```suggestion
            assert(!b2_free_set.is_free(address));
            assert(!b2_free_set.is_released(address));
```

chaitanyabhandari:
https://github.com/tigerbeetle/tigerbeetle/pull/2600/commits/b47f04419ef92a53a2ed74e31958f5711a6336da contains some refactoring, and also adds logic for testing the production checkpoint durability flow in the forest fuzzer (https://github.com/tigerbeetle/tigerbeetle/pull/2600#discussion_r1916997129).

The only comment left to address now is the additional coherency-related asserts (https://github.com/tigerbeetle/tigerbeetle/pull/2600#discussion_r1917336383)!

sentientwaffle (commented):
First pass done!

This review raced with your last commit, so some of my comments point to code that has moved... not sure how github will handle that, let's find out!

src/testing/cluster/storage_checker.zig
chaitanyabhandari: Oops, this is a remnant from an older version of the code, it can just be reverted to `0`! Although, the latest commit does specify this value for the fuzzers, and explains why:
https://github.com/tigerbeetle/tigerbeetle/blob/b47f04419ef92a53a2ed74e31958f5711a6336da/src/lsm/manifest_log_fuzz.zig#L324-L327

src/vsr/free_set.zig
chaitanyabhandari: Its the StateMachine tests! They specify `.storage_size_limit = data_file_size_min` and `.blocks_released_prior_checkpoint_durability_max = 0`. The maximum # of grid blocks corresponding to `data_file_size_min` is 0.

https://github.com/tigerbeetle/tigerbeetle/blob/b47f04419ef92a53a2ed74e31958f5711a6336da/src/state_machine.zig#L3392-L3408

src/tigerbeetle/inspect.zig
chaitanyabhandari: Note to self: Don't blindly rely on `fmt`, take a second to pause and see what formatting _actually_ looks good. 😄

src/vsr/checkpoint_trailer.zig
chaitanyabhandari: Ah yes, agreed that we were conflating there! `block_count_for_trailer_size` is much better.

src/vsr/grid.zig
chaitanyabhandari: Correct; nice catch!

src/vsr/grid.zig
chaitanyabhandari: Indeed, should be "about to be freed"! Great catch.

src/vsr/grid.zig
chaitanyabhandari: The fact that I did this in https://github.com/tigerbeetle/tigerbeetle/pull/2600/commits/b47f04419ef92a53a2ed74e31958f5711a6336da without reading your comment makes me happier than it should... :joy:

src/vsr/free_set.zig
chaitanyabhandari: We certainly can, and without breaking _any_ tests! All these functions are in much better shape assertions-wise, since we took up a lot of good changes in the last few commits. Now, all fuzzers invoke `mark_checkpoint_not_durable` and `mark_checkpoint_durable` immediately after on _every_ checkpoint, as opposed to invoking `mark_checkpoint_not_durable` once at the beginning.

src/vsr.zig
chaitanyabhandari: Ah, I _really_ like this pattern to hide mutability! Also, I had intended to scope the entire calculation and then put "blocks released within a pipeline" at the top, but forgot to do that... Changed it to the version you have recommended, its very clean.

src/vsr/free_set.zig
chaitanyabhandari: Yes, it is! A block can't exist in both `blocks_released` and `blocks_released_prior_checkpoint_durability`.

src/vsr/free_set.zig
chaitanyabhandari: Mhmmmm, now that I think about it, `verify` is set to `true` both _both_  tests (`test_min`) and production (`default_production`). So IIUC, these asserts would've _always_ evaluated to `True` and we weren't actually checking anything at all? 😅

chaitanyabhandari: _Unless_ someone explicitly builds TB with `config_verify=false`...

sentientwaffle: I suspect we will turn it off for production builds eventually! But yeah, right now it is always on.

src/vsr/replica.zig
chaitanyabhandari: Ah, perfect! The flow is much, much cleaner now.

src/testing/cluster/storage_checker.zig
chaitanyabhandari: Mhmmmmm, I didn't quite understand this comment... Are you recommending parametrizing `StorageChecker` by `Forest`, similar to `ManifestCheckerType`, and then passing in `replica.state_machine.forest` to `StorageChecker.check` as well? 🤔

chaitanyabhandari:
Addressed comments in https://github.com/tigerbeetle/tigerbeetle/pull/2600/commits/be61420b2745638fc9ee3ec3bf4639a0e186521b! After _tons_ of refactoring, I will finally get around the thinking about the additional coherency-related asserts now. :)

src/testing/cluster/storage_checker.zig
sentientwaffle: > Are you recommending parametrizing StorageChecker by Forest, similar to ManifestCheckerType, and then passing in replica.state_machine.forest to StorageChecker.check as well?

No, don't parameterize the whole `StorageChecker`. The methods that previously took a `superblock`, and now take a `replica` instead (`check`, `checksum_grid`), should instead take `forest`, since that is the narrowest type (I think) that provides everything they need. (Now that I am looking at it, it makes sense to leave `replica_sync` with a `replica` parameter though, since the other public `StorageChecker` functions also need `replica`.

src/vsr/replica.zig
```diff
@@ -4289,6 +4306,29 @@ pub fn ReplicaType(
             return self.commit_dispatch_resume();
         }
 
+        fn commit_checkpoint_durable(self: *Replica) enum { ready, pending } {
+            assert(self.commit_stage == .checkpoint_durable);
+
+            if (!self.grid.free_set.checkpoint_durable and
+                vsr.Checkpoint.durable(self.op_checkpoint(), self.commit_min))
+            {
```
sentientwaffle: Nit, but now that the `else` branch is very simple, we can avoid using a compound condition by rewriting it to:

```suggestion
            if (self.grid.free_set.checkpoint_durable) return .ready;
            if (!vsr.Checkpoint.durable(self.op_checkpoint(), self.commit_min)) return .ready;
            
            ...
```

src/vsr/grid.zig
chaitanyabhandari: Thinking about this: with our current grid checker changes, i.e. tracking (checkpoint_id, block_address, checkpoint_durable) instead of (op_checkpoint, block_address), shouldn't we already be able to catch violations during non-durable checkpoints? Specifically, we would panic if we find a *different* checksum for a block_address, while checkpoint_id is not durable? Or did you have some other violations in mind here? 
https://github.com/tigerbeetle/tigerbeetle/blob/5ab07ff443b78f6eac13472832638145cff92b74/src/testing/cluster/grid_checker.zig#L28-L32

src/vsr/replica.zig
chaitanyabhandari: I really like this nit! Makes the code more beautiful. :)

src/vsr/grid.zig
chaitanyabhandari: >  hadn't considered this case! So a block might be coherent with the new checkpoint before it is durable, but then when the checkpoint transitions to durable that block is no longer allowed b/c it was freed

Also, this was one of my favourite findings from this work as well. I was 🤯!

sentientwaffle: > Thinking about this: with our current grid checker changes, i.e. tracking (checkpoint_id, block_address, checkpoint_durable) instead of (op_checkpoint, block_address), shouldn't we already be able to catch violations during non-durable checkpoints? Specifically, we would panic if we find a different checksum for a block_address, while checkpoint_id is not durable? Or did you have some other violations in mind here?

Yes, we already catch incoherent blocks, as long as either:
- both versions of the block were recorded while `checkpoint_durable`, or
- both versions of the block were recorded while `!checkpoint_durable`.

But `assert_coherent` isn't called exhaustively for each checkpoint -- rather, we call it opportunistically, whenever we happen  to read/write a block. Since we have `checkpoint_durable=false` for a small span of time (relative to `checkpoint_durable=true`) we will call `assert_coherent(... checkpoint_durable=false)` for fewer blocks. The more blocks we can check, the better our chances of finding a collision (if a collision exists).

So if we record a block when `checkpoint_durable`, could we simultaneously infer that the same block is part of (that same checkpoint's) `!checkpoint_durable` block set?

chaitanyabhandari: Mhmmmm if a block is recorded when `checkpoint_durable`, then either of these could be true:
* The block was freed and recreated after checkpoint durability, with a different version from what it was when `!checkpoint_durable`
* The block was not released in the previous checkpoint, so it should be the same version as it was when `!checkpoint_durable`

 Even if its the latter, it could be that we didn't perform any read/write to the block when `!checkpoint_durable`, so that block version wouldn't exist in the checkpoint's `!checkpoint_durable` block set...

Although, if we record a version V of a block when `!checkpoint_durable` for checkpoint C<sub>n</sub>, then we _can_ say that _if_ the block exists in C<sub>n-1</sub>'s `checkpoint_durable` block set, it _must_ be version V? Is this the assert that you were referring to?

chaitanyabhandari: Took a stab at it in https://github.com/tigerbeetle/tigerbeetle/pull/2600/commits/1e2333e43f794863932946d952957948c6446d80! But this doesn't quite solve the issue that you mentioned wherein we call `assert_coherent` for _fewer_ blocks during `!checkpoint_durable`, since it only adds another assert for the blocks that we _do_ check. 😕

sentientwaffle: > Although, if we record a version V of a block when !checkpoint_durable for checkpoint Cn, then we can say that if the block exists in Cn-1's checkpoint_durable block set, it must be version V? Is this the assert that you were referring to?

Ah ok, the invariant I was imagining doesn't actually hold! Thanks for the explanation.

Using the parent checksum is a clever workaround. And I guess we can't instead insert that block into the parent's `checker.blocks` because it might be newly-allocated during the non-durable checkpoint.

LGTM!

chaitanyabhandari:
@sentientwaffle changes to send both the old and new versions of the start_view message are in: 9db5eabb6f9d981dba26a54d222a76ab24fb86d1!

src/vsr/replica.zig
```diff
@@ -2361,7 +2354,12 @@ pub fn ReplicaType(
         }
 
         fn on_start_view_set_checkpoint(self: *Replica, message: *Message.StartView) bool {
+            // Accept SV messages for state sync only from replicas that are on the new
+            // CheckpointState.
```
chaitanyabhandari, line 2358: The rationale behind only accepting SV messages for state sync from _new_ replicas is that we would never find ourselves in a situation where a replica on a newer release is lagging behind a replica that is on an older release. This reasoning _is_ sound, no?

sentientwaffle: Makes sense to me, though I would add a `log.warn()` to that branch just in case.

src/vsr/replica.zig
```diff
@@ -2464,7 +2462,10 @@ pub fn ReplicaType(
                             break;
                         }
                     }
-                } else unreachable;
+                } else {
+                    assert(message.header.command == .start_view_deprecated);
+                    return;
```
sentientwaffle: :thinking: How does this get hit?

src/vsr/replica.zig
chaitanyabhandari: Added!

src/vsr/replica.zig
```diff
@@ -5086,7 +5104,10 @@ pub fn ReplicaType(
 
         /// Construct a SV message, including attached headers from the current log_view.
         /// The caller owns the returned message, if any, which has exactly 1 reference.
-        fn create_start_view_message(self: *Replica, nonce: u128) *Message.StartView {
+        fn create_start_view_message(self: *Replica, nonce: u128) struct {
+            current: *Message.StartView,
```
sentientwaffle: Maybe `s/current/deprecated`, since this release will actually prefer to receive `next`, not `current`?

src/vsr/replica.zig
```diff
@@ -5140,7 +5161,68 @@ pub fn ReplicaType(
             message.header.set_checksum_body(message.body_used());
             message.header.set_checksum();
 
-            return message.ref();
+            const message_copy = self.message_bus.get_message(.start_view_deprecated);
+            defer self.message_bus.unref(message_copy);
```
sentientwaffle: This name might be misleading... `message_copy` isn't a copy of `message`.

src/vsr/superblock.zig
```diff
@@ -1211,13 +1363,65 @@ pub fn SuperBlockType(comptime Storage: type) type {
                 superblock.staging.* = working.*;
 
                 const working_checkpoint = &superblock.working.vsr_state.checkpoint;
+                const staging_checkpoint = &superblock.staging.vsr_state.checkpoint;
+
+                // The SuperBlock on disk is an older version wherein @TypeOf(VSRState.checkpoint)
+                // is CheckpointStateOld. Translate CheckpointStateOld → CheckpointState (zeroing
+                // out new fields) and update the working and staging superblocks.
+                if (working.version < SuperBlockVersion) {
```
sentientwaffle: I think this should be stricter, i.e. instead of
```
if (working.version < SuperBlockVersion) {
```
we should have:
```
if (working.version == 1) {
    // ...
} else {
    assert(working.version == SuperBlockVersion);
}
```

And actually, just after we get the working superblock:

```
if (superblock.quorums.working(superblock.reading, threshold)) |quorum| {
```
we should test that the version is one that we can use, and panic informatively otherwise. Afaict we don't check this yet (even before this PR I mean), which is a latent bug!

src/vsr/replica.zig
chaitanyabhandari: Its because we call `on_start_view_set_journal` for `start_view_deprecated` messages that we received, even if they are too far ahead (which we never do for `start_view` messages). If a replica sends both `start_view` and `start_view_deprecated` to another lagging replica, it would hit this branch... 
https://github.com/tigerbeetle/tigerbeetle/blob/fa42fa9f3e457d2da579a01a8890e881b7af9ae1/src/vsr/replica.zig#L2337-L2353

src/vsr/replica.zig
chaitanyabhandari: Agreed, changed to `message_deprecated`! 👍🏼

src/vsr/replica.zig
sentientwaffle: Ah that makes sense, thanks!

src/vsr/superblock.zig
chaitanyabhandari: The latent bug is a great catch! Took care of it in https://github.com/tigerbeetle/tigerbeetle/pull/2600/commits/6531e747339f2e9f68ff76f2f70f94b278d42f58.

sentientwaffle (approved):
LGTM!

matklad:
FWIW, this is currently red on VOPR:

```
fccba09 #2600 ./zig/zig build -Drelease vopr -- 6175602010311320490
```

Not sure if this PR was supposed to fix _all_ outstanding issues!

chaitanyabhandari:
@matklad discussed this one with DJ last Friday, it is a false positive unrelated to these changes for which I'm raising a separate PR. 😄

matklad:
@chaitanyabhandari can we remove `CheckpointStateOld` now?

PR #2603, opened 2024-12-27
VSR: State sync logging/tracing/config

sentientwaffle:
- Add `--timeout-grid-repair-ms` (experimental) runtime flag for tuning state sync performance.
- Add `--timeout-prepare-ms` (experimental) runtime flag for tuning replication performance in compromised clusters.
- Log and trace state sync progress.

### DIFF
--- a/src/vsr/replica.zig
+++ b/src/vsr/replica.zig
@@ -541,6 +541,7 @@ pub fn ReplicaType(
             release_execute: *const fn (replica: *Replica, release: vsr.Release) void,
             release_execute_context: ?*anyopaque,
             test_context: ?*anyopaque = null,
+            timeout_grid_repair_message_ticks: ?u64 = null,
         };
 
         /// Initializes and opens the provided replica using the options.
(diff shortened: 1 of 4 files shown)


src/vsr/replica.zig
```diff
@@ -9308,8 +9330,21 @@ pub fn ReplicaType(
...
+                        queue_table.index_address,
+                        queue_table.index_checksum,
+                        queue_table.table_blocks_written,
+                        queue_table.table_blocks_total,
+                    },
+                );
+
                 const table: *RepairTable = @fieldParentPtr("table", queue_table);
                 self.grid_repair_tables.release(table);
+                self.trace.stop(.{ .replica_sync_table = .{
```
matklad, line 9345: Do we need reset for the case where we jump from sync to sync?

sentientwaffle: No, in that case we call `GridBlocksMissing.cancel()`:

```zig
    pub fn cancel(queue: *GridBlocksMissing) void {
        assert(queue.checkpointing == null);

        queue.faulty_blocks.clearRetainingCapacity();
        while (queue.faulty_tables.pop()) |table| {
            queue.faulty_tables_free.push(table);
        }

        queue.* = .{
            .options = queue.options,
            .faulty_blocks = queue.faulty_blocks,
            .faulty_tables_free = queue.faulty_tables_free,
        };
    }
```

which still results in the table being reclaimed.

matklad (approved):

PR #2605, opened 2024-12-27
vsr: tone down false warning in clock

matklad:
m0 (time of ping) can legitimately be prior to window start if the reset happened right between a ping and a pong, which is by far the likelier explanation than a restart.

On the other hand, m2 has to be within the window, as this compares two reads of the monotonic clock of the same process.

### DIFF
--- a/src/vsr/clock.zig
+++ b/src/vsr/clock.zig
@@ -262,24 +262,16 @@ pub fn ClockType(comptime Time: type) type {
                 return;
             }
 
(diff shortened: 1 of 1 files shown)


matklad:
Draft as I need to also _test_ that the assert won't fire :-) I believe this, but to err is human!

EDIT: run fuzzing with larger parameters manually, seems to work!

chaitanyabhandari (approved):

PR #2608, opened 2024-12-27
replica: broadcast small prepares

cb22:
(this currently breaks some replica tests...)

### DIFF
--- a/src/vsr/replica.zig
+++ b/src/vsr/replica.zig
@@ -7288,8 +7288,23 @@ pub fn ReplicaType(
             // accordingly).
             maybe(message.header.op < self.op);
 
+            // TODO: Prepares are normally ring replicated, but commits are broadcast. It's possible
+            // for a commit to arrive before prepares, because they are broadcast directly from the
+            // primary whereas the prepares have to go through the ring. This will result in that
+            // replica not replicating, even though it should!
             if (message.header.op <= self.commit_max) {
                 log.debug("{}: replicate: not replicating (committed)", .{self.replica});
+                return;
+            }
+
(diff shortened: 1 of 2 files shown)


src/vsr/replica.zig
```diff
@@ -7288,11 +7288,26 @@ pub fn ReplicaType(
...
                 log.debug("{}: replicate: not replicating (committed)", .{self.replica});
                 return;
             }
 
+            // Broadcasting uses 5x the bandwidth of ring topology, but is significantly faster for
+            // smaller messages. Only broadcast if the message size is 5x less than
+            // message_size_max, to keep the maximum total bandwidth usage around the same.
+            const broadcast_threshold = @divFloor(constants.message_size_max, 5);
+            if (self.status == .normal and self.primary() and
+                message.header.size < broadcast_threshold)
```
sentientwaffle, line 7305: Should we also disable backups' ring replication in this case, so that they don't just ring-replicate a message they receive via broadcast?

matklad: I think it's fine if they broadcast: we do have protections from this being quadratic (actually to strict I think), so the overall bandwidth will be the same as for the case of sending a big message once

matklad (approved):

src/vsr/replica.zig
sentientwaffle: > I think it's fine if they broadcast

Nit, but they wouldn't broadcast (since they probably aren't normal-primaries), they would ring-replicate (to replicas that likely already received the broadcast).

matklad: they = the primary :P

PR #2610, opened 2024-12-27
clients/go: use `AccountEventResult` instead of `TransferEventResult` for `resultsCount` calculation in `CreateAccounts`

GuilAFT:
Doesn't make that much of a difference now (as both types are the same size), but it might cause unexpected bugs in the future.


cb22:
Great catch @GuimilXD!

cb22 (approved):

PR #2611, opened 2024-12-27
[Client] Add Events for efficient cross-thread notification

batiati:
Removes the local TCP server hack being used in `tb_client.Signal` to facilitate cross-thread notification.

Instead, it uses the most efficient mechanism for waking the IO abstraction: `eventfd` on Linux, `EVFILT_USER` on macOS, and `PostQueuedCompletionStatus` on Windows.

Reviewed from previous work https://github.com/tigerbeetle/tigerbeetle/pull/622

### DIFF
--- a/src/clients/c/test.zig
+++ b/src/clients/c/test.zig
@@ -244,10 +244,6 @@ test "c_client tb_status" {
     // All other status are not testable.
 }
 
-test {
-    _ = @import("tb_client/signal.zig");
-}
-
 // Asserts the validation rules associated with the "TB_PACKET_STATUS" enum.
 test "c_client tb_packet_status" {
     const RequestContext = RequestContextType(constants.message_body_size_max);
--- a/src/clients/c/test.zig
+++ b/src/clients/c/test.zig
@@ -244,6 +244,10 @@ test "c_client tb_status" {
     // All other status are not testable.
 }
 
+test {
+    _ = @import("tb_client/signal.zig");
+}
+
 // Asserts the validation rules associated with the "TB_PACKET_STATUS" enum.
 test "c_client tb_packet_status" {
     const RequestContext = RequestContextType(constants.message_body_size_max);
--- a/src/io/linux.zig
+++ b/src/io/linux.zig
@@ -1367,6 +1367,76 @@ pub const IO = struct {
         self.enqueue(completion);
     }
 
+    pub const Event = posix.fd_t;
+    pub const INVALID_EVENT: Event = -1;
+
+    pub fn open_event(self: *IO) !Event {
+        _ = self;
+
+        // eventfd initialized with no (zero) previous write value.
+        const event_fd = posix.eventfd(0, linux.EFD.CLOEXEC) catch |err| switch (err) {
+            error.SystemResources,
+            error.SystemFdQuotaExceeded,
+            error.ProcessFdQuotaExceeded,
+            => return error.SystemResources,
+            error.Unexpected => return error.Unexpected,
+        };
+        assert(event_fd != INVALID_EVENT);
+        errdefer os.close(event_fd);
+
+        return event_fd;
+    }
+
+    pub fn event_listen(
+        self: *IO,
+        event: Event,
+        completion: *Completion,
+        comptime on_event: fn (*Completion) void,
+    ) void {
+        assert(event != INVALID_EVENT);
+        const Context = struct {
+            const Context = @This();
+            var buffer: u64 = undefined;
+
+            fn on_read(
+                _: *Context,
+                completion_inner: *Completion,
+                result: ReadError!usize,
+            ) void {
+                const bytes = result catch unreachable; // eventfd reads should not fail.
+                assert(bytes == @sizeOf(u64));
+                on_event(completion_inner);
+            }
+        };
+
+        self.read(
+            *Context,
+            undefined,
+            Context.on_read,
+            completion,
+            event,
+            std.mem.asBytes(&Context.buffer),
+            0, // eventfd reads must always start from 0 offset.
+        );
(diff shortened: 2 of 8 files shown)


src/io/linux.zig
```diff
@@ -1301,6 +1301,89 @@ pub const IO = struct {
...
+        assert(event_fd != INVALID_EVENT);
+        _ = completion;
+        _ = self;
+
+        const value: u64 = 1;
+        const bytes = posix.write(event_fd, std.mem.asBytes(&value)) catch unreachable;
+        assert(bytes == @sizeOf(u64));
+    }
+
+    pub fn close_event(self: *IO, event_fd: posix.fd_t, completion: *Completion) void {
```
matklad: Let's not pass `Completion`? We don't use it, and we don't pass it in close_socket either.

src/io/linux.zig
```diff
@@ -1301,6 +1301,89 @@ pub const IO = struct {
...
     }
 
+    pub const INVALID_EVENT = -1;
+
+    pub fn open_event(
+        self: *IO,
+        completion: *Completion,
+        comptime on_event: fn (*Completion) void,
+    ) !i32 {
+        // eventfd initialized with no (zero) previous write value,
```
matklad: s/,/. ?

src/io/linux.zig
matklad: Hm, what is the cancelation story here? At the moment we call `close`, there might be an outstanding `event_listen`. It might even have fired! So we could get a callback called after `.close`. 

Should `event_listen` be checking `cancel_status` perhaps?

matklad (commented):
I feel like some sort of cancelation story is needed for Linux&Windows? (completion-based darwin should be fine perhaps)

batiati:
> I feel like some sort of cancelation story is needed for Linux&Windows? (completion-based darwin should be fine perhaps)

Indeed, we didn't have a well-defined cancelation story at all!

Refactored the `event` so that it does not reattach itself: it's the caller's responsibility to call `event_listen`, like the caller would `recv` a socket again after reading from it.

`Signal` is now the source of truth for shutdowns, avoiding leaving unfinished events registered.

src/io/test.zig
```diff
@@ -332,6 +332,71 @@ test "timeout" {
...
+            const self: *Context = @fieldParentPtr("event_completion", completion);
+            assert(std.Thread.getCurrentId() == self.main_thread_id);
+
+            self.count += 1;
+            if (self.count == events_count) return;
+
+            // Reattaching the event.
+            self.io.event_listen(self.event, &self.event_completion, on_event);
+        }
+    }.run_test();
```
matklad, line 397: Nice pattern, will certainly steal it for myself!

src/io/test.zig
```diff
@@ -332,6 +332,71 @@ test "timeout" {
...
+    try struct {
+        const Context = @This();
+
+        io: IO,
+        count: u32 = 0,
+        main_thread_id: std.Thread.Id,
+        event: IO.Event = IO.INVALID_EVENT,
+        event_completion: IO.Completion = undefined,
+
+        const delay = 100 * std.time.ns_per_ms;
```
matklad: Shorten to 5 to make the test faster? 100 ms is a lot of time!

src/io/linux.zig
```diff
@@ -1367,6 +1367,76 @@ pub const IO = struct {
...
     }
 
+    pub const Event = posix.fd_t;
+    pub const INVALID_EVENT: Event = -1;
+
+    pub fn open_event(self: *IO) !Event {
+        _ = self;
+
+        // eventfd initialized with no (zero) previous write value.
+        const event_fd = posix.eventfd(0, linux.EFD.CLOEXEC) catch |err| switch (err) {
```
matklad, line 1377: Pre-existing, but we probably shouldn't go through `posix` here, for two reasons:

* even if libc is linked we'd rather do syscalls directly, to not mess up with errno
* we _know_ we are linux, we don't need cross-platform posix abstraction!

src/clients/c/test.zig
```diff
@@ -244,6 +244,10 @@ test "c_client tb_status" {
     // All other status are not testable.
 }
 
+test {
+    _ = @import("tb_client/signal.zig");
```
matklad: Include the at the top-level, in unit_tests.zig. I don't have a strong argument for one list versus
hierarchical tree of inclusions, but one list is what we do today!

src/io/linux.zig
batiati: The thing is, we'd need to handle `errno` for `linux.eventfd(...)` ourselves.
I can open a follow-up PR replacing all posix calls with native Linux ones.

matklad: >The thing is, we'd need to handle errno for linux.eventfd(...) ourselves.


I don't think so? The errno is the libc thing, not the syscall thing. sys calls just return an error code as a return value as any non-insane API would do, and then libc wrapper helpfully stuffs the error code into the `errno` thread local. 

So if we use `linux.xxx` API, we don't have to touch `errno`. 

Or are you talking specifically about converting an `u32` return number to an Error value? I _think_ `std.os.linux.E.init(code)` would do that?

matklad (approved):

src/io/linux.zig
batiati: > Or are you talking specifically about converting an u32 return number to an Error value? I think std.os.linux.E.init(code) would do that?

Yes, I mean making our `io/linux.zig` to switch over `E` instead of handling the `error` returned by the posix abstraction (not only for `eventfd`, but for all other places we use a posix call such as `write`, `open`, `close`, `fsync`, etc).

PR #2612, opened 2024-12-27
bus: adaptively batch messages in the send queue

matklad:
Before this PR, sending a message meant adding it as is to a message queue. Given that the queue is only 4 messages long, this could easily overflow the queue if the messages are small. In particular, if a replica accumulated a lot of bytes in its tcp receive buffer, then it'll receive a bunch of messages almost simultaneously, and can overflow the send queue without givin a chance for the kernel to even copy the bytes.

To combat this problem, introduce batching of messages, in an approach dual to that on the receive path:

- if the queue is empty, send a message even if it is tiny
- if there's already a queued message, try to batch with it
    - batching involves an extra memcpy, but this is fine, as we can't send the bytes over at the moment anyway

### DIFF
--- a/src/message_bus.zig
+++ b/src/message_bus.zig
@@ -23,7 +23,10 @@ pub const MessageBusReplica = MessageBusType(.replica);
 pub const MessageBusClient = MessageBusType(.client);
 
 fn MessageBusType(comptime process_type: vsr.ProcessType) type {
-    const SendQueue = RingBufferType(*Message, .{
+    // To avoid sending tinygrams, message bus concatenates small messages together.
+    // Invariant: message.header.size ≤ size
+    const MessageBatch = struct { message: *Message, size: usize };
+    const SendQueue = RingBufferType(MessageBatch, .{
         .array = switch (process_type) {
             .replica => constants.connection_send_queue_max_replica,
             // A client has at most 1 in-flight request, plus pings.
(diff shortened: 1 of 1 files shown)


src/message_bus.zig
```diff
@@ -669,6 +682,50 @@ fn MessageBusType(comptime process_type: vsr.ProcessType) type {
...
+                // The head message could be in the kernel already, can't batch with it.
+                if (connection.send_queue.count < 2) return false;
+
+                const batch: *MessageBatch = connection.send_queue.tail_ptr().?;
+                if (batch.size + message.header.size > constants.message_size_max) {
+                    return false; // To big to batch.
+                }
+
+                if (batch.message.header.size == batch.size) {
+                    // Get a fresh message from a message bus to avoid aliasing.
```
sentientwaffle, line 699: :exploding_head: Subtle! So initially we just use the exact message, but if anything gets batched with it we need to do a full copy of both messages -- the first message might be referenced multiple times by the send queue(s), so we can't safely mutate it.

Could we set a higher threshold for qualifying for batching? e.g. to guarantee that we don't copy a message that is >128KiB (number chosen at random).  It is common to mix large/small messages together. (For example, a primary might send a `prepare` that is 95% full, and then follow it with a `commit` or a `ping`, ) In that case we would still want to be zero-copy for the prepare.

Then we could also batch with any (non-head) message in our send queue, not just the last one. That way large-small-large-small is 3 batches, not 4.

All that said: I'm happy with merging this as-is since it is already a big improvement for the small-message workload.

sentientwaffle (approved):

cb22 (changes requested):
(blocking this from accidentally merging until we discuss the state sync result from it a bit more!)

matklad:
I think #2616 is a better (more mechanically sympathetic) implementation of this idea

PR #2613, opened 2024-12-28
vsr: replicate committed messages

matklad:
Previously, only messages advancing op and not known to be committed were replicated. This played badly with a pipeline: if two prepares get reordered on _one_ replica, it creates a bubble that rides until the end of the ring, triggering repairs on any replicas along the way.

With this commit, we replicate any prepare that we haven't seen before.

I originally considered explicitly amounting for a pipeline, and repairing after `op -| pipeline`, but I think that would be worse.

Consider a scenario where half of the cluster gets odd prepares, and half of the cluster gets even ones. Here, if a replica from the first half gets a lucky even prepare, it can be well-advised to forward it along.

Another case: consider replica re-joining the cluster after a delay. This replica will be requesting all prepares, and it, at the first sight, seems wasteful to replicate them all. But actually, as the replica were missing, it is likely that anyone in front of it is now missing these prepares either! So it actually is good if it can replay them.

### DIFF
--- a/src/vsr/replica.zig
+++ b/src/vsr/replica.zig
@@ -7320,14 +7320,16 @@ pub fn ReplicaType(
         /// TODO Use recent heartbeat data for next replica to leapfrog if faulty (optimization).
         fn replicate(self: *Replica, message: *Message.Prepare) void {
             assert(message.header.command == .prepare);
-            assert(message.header.view >= self.view);
-            // We may replicate older prepares from either a primary of the current or future view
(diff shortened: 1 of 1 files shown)


cb22 (dismissed):

cb22:
@matklad looks like the VOPR isn't happy!

matklad:
Oops, that's a trivial error in an assert of course, but it would have been pretty bad to ship it!

matklad:
Ok, this is vopr-lite green at 14 974

cb22 (approved):

PR #2614, opened 2024-12-28
[Client] Prevent IO thread starvation

batiati:
This is a "not react to stimuli" problem: The `tb_client` IO thread consumes packets produced by multiple user threads in a tight loop, potentially blocking the IO under heavy load for as long as packets are available to consume.

- Now `on_signal` only consumes the number of packets necessary to batch one more request into `pending`. Additionally, it enforces an upper limit to prevent too many invalid packets from blocking the IO thread.
- `SubmissionStack` was a first-in-last-out data structure, effectively reordering requests when moving them from `submitted` to `pending`. It has been replaced with a FIFO structure guarded by a mutex, which has minimal impact compared to the previous spin lock (in the past a Mutex was used to guard the pool of packets before packets implemented intrusive memory).
- ~The `notify()` signal does not use `timeout=0` anymore if fired while running, allowing other IO completions to run between notifications.~ (Superseded by #2611)

### DIFF
--- a/src/clients/c/tb_client/packet.zig
+++ b/src/clients/c/tb_client/packet.zig
@@ -1,6 +1,8 @@
 const std = @import("std");
+const builtin = @import("builtin");
 const assert = std.debug.assert;
-const Atomic = std.atomic.Value;
+
+const FIFOType = @import("../tb_client.zig").vsr.fifo.FIFOType;
 
 pub const Packet = extern struct {
     next: ?*Packet,
@@ -31,31 +33,33 @@ pub const Packet = extern struct {
         invalid_data_size,
     };
 
-    /// Thread-safe stack optimized for 1 consumer (io thread) and N producers (client threads),
-    /// `push` uses a spin lock, and `pop` is not thread-safe.
-    pub const SubmissionStack = struct {
-        pushed: Atomic(?*Packet) = Atomic(?*Packet).init(null),
-        popped: ?*Packet = null,
+    /// Thread-safe FIFO.
+    pub const SubmissionQueue = struct {
+        fifo: FIFOType(Packet) = .{
+            .name = null,
+            .verify_push = builtin.is_test,
+        },
+        mutex: std.Thread.Mutex = .{},
 
-        pub fn push(self: *SubmissionStack, packet: *Packet) void {
-            var pushed = self.pushed.load(.monotonic);
-            while (true) {
-                packet.next = pushed;
-                pushed = self.pushed.cmpxchgWeak(
-                    pushed,
-                    packet,
-                    .release,
-                    .monotonic,
-                ) orelse break;
-            }
+        pub fn push(self: *SubmissionQueue, packet: *Packet) void {
+            self.mutex.lock();
+            defer self.mutex.unlock();
+
+            self.fifo.push(packet);
         }
 
-        pub fn pop(self: *SubmissionStack) ?*Packet {
-            if (self.popped == null) self.popped = self.pushed.swap(null, .acquire);
-            const packet = self.popped orelse return null;
-            self.popped = packet.next;
-            packet.next = null;
-            return packet;
+        pub fn pop(self: *SubmissionQueue) ?*Packet {
+            self.mutex.lock();
+            defer self.mutex.unlock();
+
+            return self.fifo.pop();
+        }
+
+        pub fn empty(self: *SubmissionQueue) bool {
+            self.mutex.lock();
+            defer self.mutex.unlock();
+
+            return self.fifo.count == 0;
         }
     };
 };
--- a/src/clients/dotnet/TigerBeetle.Tests/IntegrationTests.cs
+++ b/src/clients/dotnet/TigerBeetle.Tests/IntegrationTests.cs
@@ -1939,27 +1939,14 @@ public class IntegrationTests
 
     private void ConcurrencyTest(bool isAsync)
     {
-        const int TASKS_QTY = 100_000;
-
         using var client = new Client(0, new[] { server.Address });
 
         var accounts = GenerateAccounts();
         var accountResults = client.CreateAccounts(accounts);
         Assert.IsTrue(accountResults.Length == 0);
 
-        var tasks = new Task<CreateTransferResult>[TASKS_QTY];
-
-        async Task<CreateTransferResult> asyncAction(Transfer transfer)
-        {
-            return await client.CreateTransferAsync(transfer);
-        }
-
-        CreateTransferResult syncAction(Transfer transfer)
-        {
-            return client.CreateTransfer(transfer);
-        }
-
-        for (int i = 0; i < TASKS_QTY; i++)
+        var tasks = new Task<CreateTransferResult>[isAsync ? 1_000_000 : 10_000];
+        for (int i = 0; i < tasks.Length; i++)
         {
             var transfer = new Transfer
             {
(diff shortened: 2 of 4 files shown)


src/clients/java/src/test/java/com/tigerbeetle/IntegrationTest.java
```diff
@@ -1292,15 +1292,11 @@ public void testCloseConcurrent() throws Throwable {
...
 
     /**
-     * This test asserts that async tasks will respect client's concurrencyMax.
+     * This test asserts that async calls will not block.
      */
     @Test
     public void testAsyncTasks() throws Throwable {
         final int TASKS_COUNT = 1_000_000;
-        final int CONCURRENCY_MAX = 8192;
-        final var semaphore = new Semaphore(CONCURRENCY_MAX);
```
batiati, line 1301: This is an old piece of code when we had a "concurrency max" limit.
Currently, this code only masked the high latency when submitting 1M async requests at once.

src/clients/dotnet/TigerBeetle.Tests/IntegrationTests.cs
```diff
@@ -1939,27 +1939,14 @@ private static ulong GetTimestampLast()
 
     private void ConcurrencyTest(bool isAsync)
     {
-        const int TASKS_QTY = 100_000;
```
batiati, line 1942: We can now run 1M async requests without penalizing the test runtime.

src/clients/c/tb_client/signal.zig
```diff
@@ -241,18 +238,16 @@ pub const Signal = struct {
...
-            .notified => {}, // A thread woke us up before we started waiting so reschedule below.
+            // We should be the only ones who could've started waiting.
+            .waiting => unreachable,
+            // Wait for a `notify`.
+            .running => {},
+            // A `notify` was already called in the meantime.
+            // Not react to stimuli: since `notify` is triggered by the user,
+            // avoid calling it again in a loop (e.g., via a zero-timeout IO operation)
+            // to prevent starving the IO thread and allow other pending IO operations to complete.
+            .notified => self.notify(),
```
matklad: Comment is a bit confusing, it says "notify is triggered by the user, avoid calling it again", where I understand `it` to refer to `notify`, but then we are calling notify! 

Perhaps this was meant to say something like:


We are already signaled! We _could_ invoke the user callback right here, but that might lead to busy loop, if that callback calls notify again, without giving other IO chance to run.

Instead, don't react to stimuli and repeat the notification asynchronously.

src/clients/dotnet/TigerBeetle.Tests/IntegrationTests.cs
matklad: 🎉

src/clients/c/tb_client/packet.zig
```diff
@@ -31,31 +33,31 @@ pub const Packet = extern struct {
...
 
-        pub fn pop(self: *SubmissionStack) ?*Packet {
-            if (self.popped == null) self.popped = self.pushed.swap(null, .acquire);
-            const packet = self.popped orelse return null;
-            self.popped = packet.next;
-            packet.next = null;
-            return packet;
+        /// Not thread safe, should be called only by the consumer thread.
+        pub fn empty(self: *const SubmissionQueue) bool {
+            return self.fifo.count == 0;
```
matklad, line 62: Hm, this is still UB? If producer thread calls `push` concurrently, then we have non-atomic write and non-atomic read racing. 

This'll probably work _fine_ on x86 where loads & stores are atomic, but formally this is nasal demons territory.

To make this work, we need to make this load, _and_ the corresponding store in `push` at least `.monotonic`

matklad: Hm, do we need it?

```diff
diff --git a/src/clients/c/tb_client/context.zig b/src/clients/c/tb_client/context.zig
index c273aebf0..1dd657ba3 100644
--- a/src/clients/c/tb_client/context.zig
+++ b/src/clients/c/tb_client/context.zig
@@ -327,12 +327,7 @@ pub fn ContextType(
                 // - If the packet can be batched with another previously enqueued packet.
                 if (self.pending.count > enqueued_count) break;
             }
-
-            // Defer this work to later,
-            // allowing the IO thread to remain free for processing completions.
-            if (!self.submitted.empty()) {
-                self.signal.notify();
-            }
+            self.signal.notify();
         }
 
         fn request(self: *Context, packet: *Packet) void {
diff --git a/src/clients/c/tb_client/packet.zig b/src/clients/c/tb_client/packet.zig
index 0ab9596ca..8ed65cb30 100644
--- a/src/clients/c/tb_client/packet.zig
+++ b/src/clients/c/tb_client/packet.zig
@@ -54,10 +54,5 @@ pub const Packet = extern struct {
 
             return self.fifo.pop();
         }
-
-        /// Not thread safe, should be called only by the consumer thread.
-        pub fn empty(self: *const SubmissionQueue) bool {
-            return self.fifo.count == 0;
-        }
     };
 };
```

I think that's correct diff? If we break out off the loop not via return, we know we have work to do?

matklad (commented):
Looks good overall, the `.empty` I think is UB, but it also feels like we don't need it at all actually?

src/clients/c/tb_client/packet.zig
```diff
@@ -31,31 +33,31 @@ pub const Packet = extern struct {
...
-                    .monotonic,
-                ) orelse break;
-            }
+    /// Thread-safe FIFO.
+    pub const SubmissionQueue = struct {
+        fifo: FIFOType(Packet) = .{
+            .name = null,
+            .verify_push = builtin.is_test,
+        },
+        mutex: std.Thread.Mutex = .{},
```
matklad, line 42: Good call to slap a mutex onto it! This is control plane, so we don't need lock-free here!

src/clients/c/tb_client/packet.zig
batiati: > Hm, this is still UB? If producer thread calls push concurrently

The rationale for it not being guarded by a mutex is that if the IO thread observes a dirty `fifo.count == 0`, the thread that pushed a new element will call `notify()` either way. Otherwise, if it observes a dirty `fifo.count > 0`, there's no side effect since the IO thread is the only one that can dequeue packets.

> I think that's correct diff? If we break out off the loop not via return, we know we have work to do?

It's possible that `if (self.pending.count > enqueued_count)` breaks the loop at the last element, so checking `!empty()` is necessary to avoid calling `notify()` on an empty queue.

src/clients/c/tb_client/packet.zig
batiati: Note: It feels suboptimal. We made `tb_client.submit` thread-safe, but the language clients still need to guard the `tb_client` handler with their own mutex to prevent calls to `submit()` racing against `close()` (which isn't thread-safe).

As a follow up PR I'm thinking going either way:
(A) Remove this `mutex` relying entirely on the language side synchronization.
(B) Make `tb_client` an intrusive data structure, so `close` will be responsible for stopping the thread and freeing resources but never dropping the handler itself, so the language client won't need to synchronize anymore.

src/clients/c/tb_client/packet.zig
matklad: >It's possible that if (self.pending.count > enqueued_count) breaks the loop at the last element, so checking !empty() is necessary to avoid calling notify() on an empty queue.

Ah, I see it now. It seems that we need to wrap `.empty` into a mutex then. As written, the code is UB, as there could be concurrent non-atomic writes to the same field by the producer and consumer. 

So, compiler is allowed to reason like this: "we are accessing `count` field outside of the mutex, so we must be the single thread allowed to write to this field, so I can optimize the loop above and hoist the load of count field out of the mutex". It of course won't do this, and I am 0.8 sure this won't break in practice, but I am also 0.9 sure that formally this is not kosher.

matklad (approved):

PR #2615, opened 2024-12-28
aof: improve efficiency

cb22:
This reworks the AOF to be significantly more efficient, both in space and performance, by:
* Relying on the WAL for durability between checkpoints.
* Switching to buffered read / writes and fsync.
* Removing unneeded metadata and  padding.

The unholy mix of sync buffered IO with async fsync is a stopgap measure, until a fully async AOF which will have a ton of other things as a precursor to DR comes along.

It also completely changes the on disk format: that's fine, past entries should be replayed with a matching version!

In terms of numbers:

## Before:
25000 batches in 52.22 s
transfer batch size = 2 txs
transfer batch delay = 0 us
load accepted = 957 tx/s

AOF size: 202 080KiB

## After:
25000 batches in 24.68 s
transfer batch size = 2 txs
transfer batch delay = 0 us
load accepted = 2025 tx/s

AOF size: 15 552KiB

### DIFF
--- a/src/aof.zig
+++ b/src/aof.zig
@@ -1,4 +1,5 @@
 const std = @import("std");
+const builtin = @import("builtin");
 const assert = std.debug.assert;
 
 const constants = @import("constants.zig");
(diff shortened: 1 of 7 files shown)


src/aof.zig
```diff
@@ -16,42 +17,24 @@ const Header = vsr.Header;
...
-    primary: u64,
-    replica: u64,
-    // Use large padding here to align the message itself to the sector boundary.
-    reserved: [4064]u8 = std.mem.zeroes([4064]u8),
-
-    comptime {
-        assert(stdx.no_padding(AOFEntryMetadata));
-    }
-};
+const magic_number: u128 = 980549605906083105481957035548671106;
```
matklad: nit, but might be better to spell it in hex so that its more googlable

src/aof.zig
```diff
@@ -242,43 +247,61 @@ pub const AOF = struct {
     }
 
     pub fn close(self: *AOF) void {
-        std.posix.close(self.fd);
+        self.file.close();
     }
```
matklad: `self.file = undefined`

src/aof.zig
```diff
@@ -16,42 +17,24 @@ const Header = vsr.Header;
...
 
-    /// Calculate the actual length of the AOFEntry that needs to be written to disk,
-    /// accounting for sector alignment.
+    /// Calculate the actual length of the AOFEntry that needs to be written to disk.
     pub fn calculate_disk_size(self: *AOFEntry) u64 {
-        const unaligned_size = @sizeOf(AOFEntry) - self.message.len + self.header().size;
-
-        return vsr.sector_ceil(unaligned_size);
+        return @sizeOf(AOFEntry) - self.message.len + self.header().size;
```
matklad, line 37: When we read an entry in the iterator in `next`, check that the original `bytes_read` is at least sizeOf(Header).

src/io/linux.zig
```diff
@@ -564,6 +567,26 @@ pub const IO = struct {
...
+                        if (completion.result < 0) {
+                            const err = switch (@as(posix.E, @enumFromInt(-completion.result))) {
+                                .INTR => {
+                                    completion.io.enqueue(completion);
+                                    return;
+                                },
+                                .BADF => error.FileDescriptorInvalid,
+                                .IO => error.InputOutput,
+                                .INVAL => unreachable,
+                                else => |errno| stdx.unexpected_errno("fsync", errno),
```
matklad, line 581: I am not sure we've got all interesting error codes here, but we are going to crash on sync failure anyway, so this is fine

src/aof.zig
```diff
@@ -242,43 +247,61 @@ pub const AOF = struct {
...
         entry.from_message(
             message,
-            .{ .replica = options.replica, .primary = options.primary },
             &self.last_checksum,
         );
 
         const disk_size = entry.calculate_disk_size();
 
         const bytes = std.mem.asBytes(&entry);
+        try self.file.writeAll(bytes[0..disk_size]);
```
matklad: Yeah, write here was bothering me _a lot_, thanks!

matklad (commented):
LGTM, modulo nits and what I think is a pre-existing bug in Iterator.next!

src/io/linux.zig
cb22: Those are all the ones explicitly defined for `fsync`, but yeah it can fail with any of the error conditions defined for `read` or `write` too.

src/aof.zig
cb22: Yeah, I think my original logic was very wrong, but it didn't manifest because it was also using direct IO!

matklad (approved):

PR #2616, opened 2024-12-28
bus:  batch small messages when sending over socket

matklad:
Before this PR, sending a message meant adding it as is to a message queue. Given that the queue is only 4 messages long, this could easily overflow the queue if the messages are small. In particular, if a replica accumulated a lot of bytes in its tcp receive buffer, then it'll receive a bunch of messages almost simultaneously, and can overflow the send queue without givin a chance for the kernel to even copy the bytes.

To combat this problem, try submitting messages to the kernel as soon as possible. That is, whenever we want to send a message, do a synchroneous, non-blocking write to the socket. Most likely, this'll just copy bytes over to the kernel. If it returns WouldBlock, then fall back to asynchronous send via io_uring.

This is a more mechanically sympathetic version of https://github.com/tigerbeetle/tigerbeetle/pull/2612

### DIFF
--- a/src/message_bus.zig
+++ b/src/message_bus.zig
@@ -1003,8 +1003,11 @@ fn MessageBusType(comptime process_type: vsr.ProcessType) type {
                 assert(connection.peer == .client or connection.peer == .replica);
                 assert(connection.state == .connected);
                 assert(connection.fd != IO.INVALID_SOCKET);
-                const message = connection.send_queue.head() orelse return;
                 assert(!connection.send_submitted);
+
+                connection.send_now(bus);
+
+                const message = connection.send_queue.head() orelse return;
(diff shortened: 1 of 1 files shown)


cb22:
Hmm - I don't think our accept or connect sockets are opened with NONBLOCK on Linux, though?

There is also `MSG_DONTWAIT` that can be set on a per send basis which might be safer?

matklad:
>Hmm - I don't think our accept or connect sockets are opened with NONBLOCK on Linux, though?

Whoa! It didn't occur to me to even think of that!

Switched to send&DONTBLOCK, though this now becomes linux specific.

cb22 (approved):

He-Pin:
batching batching, less system call

PR #2617, opened 2024-12-28
[VSR] Protocol batching

batiati:
### VSR Protocol Batching

This PR replaces the state machine batching mechanism (aka `Demuxer`) with protocol-integrated batching. Batching consists of the application submitting multiple independent units of work of the same `operation` (the batch payload) within a single VSR message. This amortizes network and consensus costs, enhancing performance in scenarios where highly concurrent user requests submit operations containing only a few events each, sharing the same physical request.

Previously, batching relied on the offset of each payload within the message, splitting the reply based on the `index` of each `Result` type. While this method already improved performance, it was only applicable to operations where the number of events directly correlated with the number of results. Each result had to be tagged with an `index` field indicating the corresponding event that generated it. In practice, batching was only applicable for the `create_accounts` and `create_transfers` operations.

The new batching mechanism is more versatile and can be applied to any state machine operation _capable of handling multiple events_ (queries are not supported yet, but lookups are). It eliminates the need for the state machine to be aware of batched requests by operating at the VSR protocol level, enabling custom state machines to take advantage of it as well.

#### VSR Changes

The VSR header for `request`, `prepare`, and `reply` messages now includes the `batch_count` field, which indicates whether a request is batched.

**`batch_count == 0`**: Treated as a non-batched request, skipping batching logic.
**`batch_count > 0`**: Indicates a batched request that should be split into multiple batch payloads.

Clients from older versions, which are unaware of the `batch_count` field, will always send non-batched requests (`batch_count == 0`), ensuring compatibility with the new protocol version.

**How it works:**

- The client (`tb_client` library) is responsible for accumulating concurrent requests, encoding them into batches, and sending them as a single VSR request.

- Upon receiving a batched request, the replica decodes the message body, splits it into multiple batch payloads based on `batch_count`, and feeds the state machine with independent units of work within the same VSR `op` number.

- Each batch may generate a number of `result`s unrelated to the number of `event`s (e.g., `lookup_accounts` and `lookup_transfers`), which are batched together and sent back to the client.

- The client decodes the reply's body into multiple batch payloads and calls the respective user callback for each one.

**Message layout:**

- Batched requests use a portion at the end of the message body (the batch `trailer`) to encode batch metadata, so batched requests can hold fewer events than non-batched ones.

- The `trailer` size is always a multiple of the `event`/`result` size to keep system invariants.

  - The batch `trailer` is an array of `u16` values representing the _number of events_ in each batch, as defined by the `batch_count` field in the VSR header.

  - The `trailer` has a variable length, depending on the number of batches (one `u16` per batch, in multiples of the `event`/`result` size).

  - To facilitate batch encoding, the trailer is written from the end of the message towards the beginning. For example, the last element of the array corresponds to the number of events in the first batch.

  - Unused elements in the trailer, required for padding, are filled with `maxInt(u16)`.

  Examples:
  For a request with `@sizeOf(Account) == 128` and  `batch_count == 100`, the trailer size will be **256 bytes**.
  For a reply with `@sizeOf(CreateAccountsResult) == 8` and `batch_count == 100`, the trailer size will be **200 bytes**.

```
size == 1280         message.body_used().len == 1024 bytes    
batch_count == 4     payload == 896 bytes                     trailer == 128 bytes 
┌──────────┐ ┌─────────────────────────────────────────────┐┌─────────────────────┐
│          │ │┌─────────┐┌───────────┐┌───────┐┌──────────┐││┌───────┐┌─┐┌─┐┌─┐┌─┐│
│VSR Header│ ││512 bytes││ 128 bytes ││0 bytes││ 256 bytes││││padding││2││0││1││4││
│          │ │└───▲─────┘└───▲───────┘└──▲────┘└───▲──────┘││└───────┘└┬┘└┬┘└┬┘└┬┘│
└──────────┘ └────┼──────────┼───────────┼─────────┼───────┘└──────────┼──┼──┼──┼─┘
                  │          │           │         │                   │  │  │  │  
                  │          │           │         └───────────────────┘  │  │  │  
                  │          │           │                                │  │  │  
                  │          │           └────────────────────────────────┘  │  │  
                  │          │                                               │  │  
                  │          └───────────────────────────────────────────────┘  │  
                  │                                                             │  
                  └─────────────────────────────────────────────────────────────┘  
```

- The `tb_client` avoids batching user requests if the combined payload length and trailer exceed the maximum message body size. Instead, it sends a non-batched request (`batch_count == 0`) to preserve maximum capacity (8190 events per message).

- The client can pack as many batches as the sum of payloads + trailer allows (e.g., up to 8064 single-event batches of `create_accounts` and `create_transfers`, that is VSR Header + 8064 events * 128 bytes + `[8064]u16` of `trailer` == 1MiB message).

### This PR

See the commit messages for the implementation details.
Include VOPR and Vortex (java driver only) tests.

### DIFF
--- a/src/clients/c/tb_client/context.zig
+++ b/src/clients/c/tb_client/context.zig
@@ -267,6 +267,11 @@ pub fn ContextType(
                     @panic("IO.run() failed");
                 };
             }
+
+            // If evicted, the inflight request was already canceled during eviction.
+            if (self.eviction_reason == null) {
+                self.cancel_request_inflight();
+            }
             self.cancel_queued_packets();
         }
 
@@ -280,6 +285,17 @@ pub fn ContextType(
             while (self.submitted.pop()) |packet| {
                 packet.assert_phase(.submitted);
                 self.packet_cancel(packet);
+            }
+        }
+
+        /// Cancel the current inflight packet, as it won't be replied anymore.
+        fn cancel_request_inflight(self: *Context) void {
+            if (self.client.request_inflight) |*inflight| {
+                if (inflight.message.header.operation != .register) {
+                    const packet = @as(UserData, @bitCast(inflight.user_data)).packet;
+                    packet.assert_phase(.sent);
+                    self.packet_cancel(packet);
+                }
             }
         }
 
@@ -579,14 +595,7 @@ pub fn ContextType(
 
             self.eviction_reason = eviction.header.reason;
 
-            // Cancel the current inflight packet, as it won't be replied anymore.
-            if (self.client.request_inflight) |*inflight| {
-                if (inflight.message.header.operation != .register) {
-                    const packet = @as(UserData, @bitCast(inflight.user_data)).packet;
-                    packet.assert_phase(.sent);
-                    self.packet_cancel(packet);
-                }
-            }
+            self.cancel_request_inflight();
             self.cancel_queued_packets();
         }
 
--- a/src/clients/c/tb_client/context.zig
+++ b/src/clients/c/tb_client/context.zig
@@ -634,7 +634,7 @@ pub fn ContextType(
                     result_size,
                     reply,
                     batch_count,
-                );
+                ) catch unreachable;
 
                 var it: ?*Packet = packet;
                 while (it) |batched| {
--- a/src/clients/c/tb_client/context.zig
+++ b/src/clients/c/tb_client/context.zig
@@ -15,6 +15,9 @@ const Header = vsr.Header;
 const IO = vsr.io.IO;
 const message_pool = vsr.message_pool;
 
+const BatchEncoder = vsr.BatchEncoder;
+const BatchDecoder = vsr.BatchDecoder;
+
 const MessagePool = message_pool.MessagePool;
 const Message = MessagePool.Message;
 const Packet = @import("packet.zig").Packet;
@@ -249,13 +252,6 @@ pub fn ContextType(
             self.allocator.destroy(self);
         }
 
-        pub fn tick(self: *Context) void {
-            // `Tick` is invoked by the result callback that may have arrived during shutdown.
-            if (self.state.load(.acquire) == .running) {
-                self.client.tick();
-            }
-        }
-
         pub fn io_thread(self: *Context) void {
             while (!self.signal.stop_requested()) {
                 self.client.tick();
@@ -268,40 +264,41 @@ pub fn ContextType(
                 };
             }
 
-            // Cancel the request_inflight packet if any.
-            if (self.client.request_inflight) |*inflight| {
-                if (inflight.message.header.operation != .register) {
-                    const user_data: UserData = @bitCast(inflight.user_data);
-                    self.cancel(user_data.packet);
+            // Cancelling any inflight or queued request.
+            var packet_list: ?*Packet = packet_list: {
+                if (self.client.request_inflight) |*inflight| {
+                    if (inflight.message.header.operation != .register) {
+                        const user_data: UserData = @bitCast(inflight.user_data);
+                        // Cancel the request_inflight packet if any.
+                        break :packet_list user_data.packet;
+                    }
                 }
-            }
 
-            while (self.submitted.pop()) |packet| {
+                break :packet_list self.submitted.flush();
+            };
+            while (packet_list) |packet| {
+                packet_list = packet.next orelse self.submitted.flush();
+                packet.next = null;
                 self.cancel(packet);
             }
         }
 
         /// Submit the enqueued requests through the vsr client.
         /// Always called by the io thread.
-        fn submit(self: *Context, packet: *Packet) void {
+        fn submit(self: *Context, packet_list: *Packet) void {
             assert(self.batch_size_limit != null);
             assert(self.client.request_inflight == null);
-            assert(packet.next == null);
 
-            const message = self.client.get_message().build(.request);
-            defer self.client.release_message(message.base());
-
-            // The batching logic will be inserted here:
-            const head: *Packet, const bytes_written: usize, const batch_count: u16 = blk: {
-                if (!self.packet_valid(packet)) {
-                    // Packet failed, nothing to send.
-                    self.signal.notify();
-                    return;
+            // Find a valid packet to start the batch.
+            const head: *Packet = head: {
+                var it: ?*Packet = packet_list;
+                while (it) |packet| {
+                    it = packet.next;
+                    if (self.packet_valid(packet)) break :head packet;
                 }
-
-                const event_data = packet.events();
-                stdx.copy_disjoint(.inexact, u8, message.buffer[@sizeOf(Header)..], event_data);
-                break :blk .{ packet, event_data.len, 0 };
+                // Nothing to send.
+                self.signal.notify();
+                return;
             };
 
             const operation: StateMachine.Operation = operation_from_int(head.operation).?;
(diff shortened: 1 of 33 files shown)


src/vsr/message_header.zig
```diff
@@ -530,23 +538,26 @@ pub const Header = extern struct {
...
+                    if (self.batch_count != 0) return "upgrade: batch_count != 0";
                 },
                 else => {
                     if (self.operation == .reconfigure) {
                         if (self.size != @sizeOf(Header) + @sizeOf(vsr.ReconfigurationRequest)) {
-                            return "size != @sizeOf(Header) + @sizeOf(ReconfigurationRequest)";
+                            return "reconfigure: size != @sizeOf(Header)" ++
+                                " + @sizeOf(ReconfigurationRequest)";
                         }
+                        if (self.batch_count != 0) return "reconfigure: upgrade: batch_count != 0";
```
sentientwaffle:
```suggestion
                        if (self.batch_count != 0) return "reconfigure: batch_count != 0";
```

src/state_machine.zig
```diff
@@ -606,12 +572,12 @@ pub fn StateMachineType(
...
         forest: Forest,
 
-        prefetch_input: ?[]align(16) const u8 = null,
+        prefetch_input: ?[]const u8 = null,
         prefetch_callback: ?*const fn (*StateMachine) void = null,
         prefetch_context: PrefetchContext = .null,
 
         scan_lookup: ScanLookup = .null,
-        scan_lookup_buffer: []align(16) u8,
+        scan_lookup_buffer: []u8,
```
sentientwaffle: If this doesn't need to be aligned, then we don't need to allocate it with `alignedAlloc()` (in `init()`).

Though it looks like we are `@alignCast`-ing it, so I'm confused why we remove the `align(16)`?

src/state_machine.zig
```diff
@@ -718,50 +684,74 @@ pub fn StateMachineType(
...
+                    }
                 },
                 .query_accounts, .query_transfers => {
                     if (input.len != @sizeOf(QueryFilter)) return false;
+                    if (!std.mem.isAligned(@intFromPtr(input.ptr), @alignOf(QueryFilter))) {
+                        log.info("Invalid input alignment for {s}", .{
+                            @tagName(operation),
+                        });
+                        return false;
+                    }
```
sentientwaffle, line 708: :thinking: Why can't we assert that it _is_ aligned?

src/state_machine.zig
```diff
@@ -1911,11 +1916,15 @@ pub fn StateMachineType(
...
+            output: []u8,
         ) usize {
-            const batch = mem.bytesAsSlice(u128, input);
-            const output_len = @divFloor(output.len, @sizeOf(Account)) * @sizeOf(Account);
-            const results = mem.bytesAsSlice(Account, output[0..output_len]);
+            const batch: []const u128 = @alignCast(mem.bytesAsSlice(u128, input));
+            const results: []Account = @alignCast(mem.bytesAsSlice(
+                Account,
+                output,
+            ));
```
```suggestion
            const results: []Account = @alignCast(mem.bytesAsSlice(Account, output));
```

src/state_machine.zig
```diff
@@ -1930,11 +1939,15 @@ pub fn StateMachineType(
...
+            output: []u8,
         ) usize {
-            const batch = mem.bytesAsSlice(u128, input);
-            const output_len = @divFloor(output.len, @sizeOf(Transfer)) * @sizeOf(Transfer);
-            const results = mem.bytesAsSlice(Transfer, output[0..output_len]);
+            const batch: []const u128 = @alignCast(mem.bytesAsSlice(u128, input));
+            const results: []Transfer = @alignCast(mem.bytesAsSlice(
+                Transfer,
+                output,
+            ));
```
```suggestion
            const results: []Transfer = @alignCast(mem.bytesAsSlice(Transfer, output));
```

src/clients/c/tb_client/context.zig
```diff
@@ -253,40 +258,9 @@ pub fn ContextType(
...
-                @intFromEnum(eviction.header.reason),
-            });
-
-            self.evicted = eviction.header.reason;
-            self.cancel_all();
-        }
-
-        pub fn tick(self: *Context) void {
-            if (self.evicted == null) {
+        pub fn io_thread(self: *Context) void {
```
sentientwaffle: Excellent function name!

Also, could this function be made private? I don't see it called except from our own `init()`.

src/clients/c/tb_client/context.zig
```diff
@@ -295,193 +269,254 @@ pub fn ContextType(
...
+                .pulse => unreachable,
+                inline else => |operation_comptime| @sizeOf(
+                    StateMachine.EventType(operation_comptime),
+                ),
+            };
+
             const message = self.client.get_message().build(.request);
-            errdefer self.client.release_message(message.base());
+            // Don't need to release the message, since this function cannot fail/return early.
+            defer packet.assert_phase(.sent);
```
sentientwaffle: For cases like this, our convention (in the replica) is to `defer self.client.release_message(...)` anyway, but pass `message.ref()` to `client.raw_request` instead of just `message`. That way it will still be correct if in the future we add an early return.

src/vsr/batch.zig
```diff
@@ -0,0 +1,547 @@
...
+
+        return .{
+            .element_size = element_size,
+            .buffer = buffer,
+            .bytes_written = 0,
+            .batch_count = 0,
+        };
+    }
+
+    pub fn writable(self: *const BatchEncoder) []u8 {
```
sentientwaffle, line 164: Would it make sense to combine this with `add()` (i.e. pass in the `bytes_written`) so that it isn't possible to forget it to call `add()`? Alternatively we could keep track with a boolean flag on the encoder to ensure the calls are alternating.

src/clients/c/tb_client/context.zig
```diff
@@ -295,193 +269,254 @@ pub fn ContextType(
...
+                }
+            };
+            assert(bytes_written % event_size == 0);
+            assert(bytes_written <= self.batch_size_limit.?);
 
-            const operation: StateMachine.Operation = @enumFromInt(packet.operation);
             message.header.* = .{
                 .release = self.client.release,
                 .client = self.client.id,
                 .request = 0, // Set by client.raw_request.
```
sentientwaffle, line 518: After calling `client.raw_request`, we can assert that `message.header.request != 0`, to verify that this comment is indeed correct.

src/clients/c/tb_client/context.zig
```diff
@@ -295,193 +269,254 @@ pub fn ContextType(
...
-                };
-
-                if (linked_chain_open) return false;
-            }
-
-            return true;
-        }
-
-        fn submit(self: *Context, packet: *Packet) void {
+        /// Sends the packet (the entire batched liked list of packets) through the vsr client.
```
sentientwaffle:
```suggestion
        /// Sends the packet (the entire batched linked list of packets) through the vsr client.
```

src/state_machine.zig
```diff
@@ -718,50 +684,74 @@ pub fn StateMachineType(
...
             }
 
             return true;
         }
 
-        /// Updates `prepare_timestamp` to the highest timestamp of the response.
-        pub fn prepare(
+        /// Returns the logical time increment (in nanoseconds) for the highest
+        /// timestamp of the response.
+        pub fn prepare_nanoseconds(
```
sentientwaffle, line 742: From the state machine's perspective, why does the `_nanoseconds` matter? Why does it care about the unit? Or is this just to hint that we are not actually modifying the timestamp in-place?

src/state_machine.zig
```diff
@@ -718,50 +684,74 @@ pub fn StateMachineType(
...
             self: *StateMachine,
```
```suggestion
            self: *const StateMachine,
```

src/vsr/replica.zig
```diff
@@ -4033,13 +4035,31 @@ pub fn ReplicaType(
             }
 
             if (StateMachine.operation_from_vsr(prepare.header.operation)) |prepare_operation| {
+                const body: []const u8 = body: {
+                    if (prepare.header.batch_count == 0) break :body prepare.body_used();
+
+                    // If it's a batched request, the message body must be decoded to
+                    // remove the trailer.
+                    // The `prefech` can process all batches at once.
```
```suggestion
                    // The `prefetch` can process all batches at once.
```

src/vsr/batch.zig
```diff
@@ -0,0 +1,547 @@
...
+
+        defer {
+            self.batch_events = self.batch_events[0 .. self.batch_events.len - 1];
+            self.payload = if (self.payload.len > batch_lenght)
+                self.payload[batch_lenght..]
+            else
+                &.{};
+        }
+
+        return self.payload[0..batch_lenght];
```
sentientwaffle:
```suggestion
        const batch_length = batch_length: {
            assert(self.batch_events.len > 0);
            maybe(self.payload.len == 0);

            // Batch metadata is written from the end of the message, so the last
            // element corresponds to the first batch.
            const batch_event_count = self.batch_events[self.batch_events.len - 1];
            assert(batch_event_count != std.math.maxInt(u16));
            maybe(batch_event_count == 0);

            const batch_length = batch_event_count * self.element_size;
            assert(batch_length <= self.payload.len);
            assert(batch_length % self.element_size == 0); // Must be aligned.

            break :batch_length batch_length;
        };

        defer {
            self.batch_events = self.batch_events[0 .. self.batch_events.len - 1];
            self.payload = if (self.payload.len > batch_length)
                self.payload[batch_length..]
            else
                &.{};
        }

        return self.payload[0..batch_length];
```

src/vsr/batch.zig
```diff
@@ -0,0 +1,547 @@
...
+
+    pub fn init(
+        /// The size of each element.
+        element_size: usize,
+        /// The message body used, including the trailer.
+        body: []const u8,
+        /// The number of batches encoded in this body.
+        batch_count: u16,
+    ) error{BatchInvalid}!BatchDecoder {
+        if (batch_count == 0) return error.BatchInvalid;
```
sentientwaffle, line 45: What do you think of allowing `batch_count == 0` in the decoder, so that we can push some branches out of `replica.zig` and into here?

And similarly for the encoder, though that seems less crucial. (Symmetry is nice, though.)

src/clients/c/tb_client/packet.zig
```diff
@@ -31,31 +20,152 @@ pub const Packet = extern struct {
...
         }
+
+        /// Not thread safe, should be called only by the consumer thread.
+        pub fn empty(self: *const SubmissionQueue) bool {
+            return self.fifo.count == 0;
+        }
+    };
+
+    const Phase = enum(u8) {
+        submitted = 0,
```
sentientwaffle, line 66: Why is the excplicit integer value needed?

src/clients/c/tb_client/context.zig
```diff
@@ -490,105 +525,189 @@ pub fn ContextType(
...
+            } else {
+                // Although the protocol allows `batch_count == 1`,
+                // the client never packages a single packet into a batched request.
+                assert(batch_count > 1);
+                assert(packet.batch_next != null);
+                var decoder = BatchDecoder.init(
+                    result_size,
+                    reply,
+                    batch_count,
+                ) catch unreachable;
```
sentientwaffle: Maybe change this `unreachable` into a `@panic()` that clarifies that the replica sent us bad data.
(e.g. as opposed to being `unreachable` b/c we already checked for an invalid batch, which is not the case.)

src/vsr/batch.zig
```diff
@@ -0,0 +1,547 @@
...
+        });
+
+        const buffer: []u8 = self.buffer.?;
+        return if (buffer.len >= self.bytes_written + trailer_size)
+            buffer[self.bytes_written .. buffer.len - trailer_size]
+        else
+            &.{};
+    }
+
+    pub fn can_add(self: *const BatchEncoder, bytes_written: usize) bool {
```
sentientwaffle: Is this function only used by tests? If so `assert(constants.verify);`.

Possibly I'm not grokking this function, but:
- Is `bytes_written` the right name for this parameter? If I am understanding right, it is passed bytes that are not yet in `writable` -- i.e. not yet written.
- Also, this function doesn't take into account extra trailer bytes that will need to be included.

src/clients/c/tb_client/packet.zig
batiati: This matches the initial state during the conversion from `Packet.Extern` to `Packet`, as we ensure that `reserved` is zeroed.

src/state_machine.zig
batiati: Only if we can ensure that malformed inputs cannot trip this assertion :thinking: :thinking:.

src/state_machine.zig
batiati: It hints that we are not actually modifying the timestamp and clarifies that we can `+=` the return value with a timestamp. Maybe `prepare_delta` would be a better name.

src/vsr/batch.zig
batiati: Usually, we only know the `bytes_written` after we wrote to the slice, such as:

```zig
const bytes_written = state_machine.commit(op, ..., batch.writable());
batch.add(bytes_written);
```

src/vsr/batch.zig
batiati: > Is this function only used by tests?
> Is bytes_written the right name for this parameter?

Good catch, I think I will remove this function and check `writable().len` on the callsite.

> Also, this function doesn't take into account extra trailer bytes that will need to be included.

Yes it does, `writable()` calculates the trailer size for `self.batch_count + 1`.
Just added a comment clarifying that.

batiati:
Ref: #2787

PR #2618, opened 2024-12-28
state_machine: add simple metrics

cb22:
This adds simple metrics to the state machine, emitted by logging at checkpoint. p0, p100, count, sum (and mean) and count_batch are gathered per operation, and additionally for StateMachine.compact and StateMachine.checkpoint:

```
2024-12-28 19:05:35.000Z info(state_machine): create_accounts: p0=8152us mean=18751us p100=29351us sum=37503us count=2 count_batch=10000
2024-12-28 19:05:35.000Z info(state_machine): create_transfers: p0=710us mean=1068us p100=14894us sum=1055393us count=988 count_batch=1976
2024-12-28 19:05:35.000Z info(state_machine): compact: p0=29us mean=857us p100=116716us sum=851132us count=992 count_batch=992
2024-12-28 19:05:35.000Z info(state_machine): checkpoint: p0=6736us mean=6736us p100=6736us sum=6736us count=1 count_batch=1
```

### DIFF
--- a/src/state_machine.zig
+++ b/src/state_machine.zig
@@ -521,6 +521,85 @@ pub fn StateMachineType(
             }
         };
 
+        const Metrics = struct {
+            const TimingSummary = struct {
+                duration_min_us: ?u64 = null,
+                duration_max_us: ?u64 = null,
+                duration_sum_us: u64 = 0,
+
+                count: u64 = 0,
+
+                /// If an operation supports batching (eg, lookup_accounts) this is a count of the
+                /// number of internal operations.
+                count_batch: u64 = 0,
+            };
+
+            /// Technically 'timer' can't be used, but that'll error out at comptime.
+            const MetricEnum = std.meta.FieldEnum(Metrics);
+
+            create_accounts: TimingSummary = .{},
+            create_transfers: TimingSummary = .{},
+            lookup_accounts: TimingSummary = .{},
+            lookup_transfers: TimingSummary = .{},
+            get_account_transfers: TimingSummary = .{},
+            get_account_balances: TimingSummary = .{},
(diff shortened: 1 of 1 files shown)


src/state_machine.zig
```diff
@@ -521,6 +521,85 @@ pub fn StateMachineType(
...
+                /// number of internal operations.
+                count_batch: u64 = 0,
+            };
+
+            /// Technically 'timer' can't be used, but that'll error out at comptime.
+            const MetricEnum = std.meta.FieldEnum(Metrics);
+
+            create_accounts: TimingSummary = .{},
+            create_transfers: TimingSummary = .{},
+            lookup_accounts: TimingSummary = .{},
```
chaitanyabhandari, line 542: IIUC, all of these are currently measuring the time taken across `prefetch`, `reply_setup`, and `commit`, where `reply_setup` is a VSR-level operation, not state machine level. 

Two questions here:
* Would it be better to remove `reply_setup` from the calculation since it is VSR level?
* Would it make more sense to measure timings across (`prefetch` → `prefetch_finish`), and within `commit` separately? Or do we want to club `prefetch` & `commit` because `commit` is purely in-memory?

cb22: Both of those are good points! 

I think adding more granularity (either in a followup, or for the real metrics work) will be the way to go.

chaitanyabhandari: Ack, sounds reasonable, we can take these up as enhancements in the current metrics in case we need more granularity. 👍🏼 

Those were the only questions I had, apart from the that this looks great to me. Approving!

chaitanyabhandari (approved):

PR #2620, opened 2024-12-28
replica: always replicate, even if in journal

cb22:
It could happen that while a primary is ring-replicating a prepare, a replica might receive a broadcasted message containing future knowledge, decide to repair a few headers, and race requesting the headers (and installing them in memory), before it receives the prepare for the first time from the primary, breaking ring replication to subsequent replicas.

### DIFF
--- a/src/vsr/replica.zig
+++ b/src/vsr/replica.zig
@@ -1604,7 +1604,13 @@ pub fn ReplicaType(
             // - replicate anything that the next replica is likely missing,
             // - avoid a feedback loop of cascading needless replication.
             // Replicate anything that we didn't previously had ourselves.
-            if (message.header.op > self.commit_min and !self.journal.has(message.header)) {
+            //
+            // Use `has_clean` (checks whether a replica has both the header and the corresponding
+            // prepare) instead of `has` (checks whether the replica has the header). The latter
+            // is prone to a race wherein a replica that receives a future header *before* the
+            // corresponding prepare (via `start_view` for instance) would end up not forwarding
+            // the prepare, thereby breaking the replication chain.
+            if (message.header.op > self.commit_min and !self.journal.has_clean(message.header)) {
                 self.replicate(message);
             }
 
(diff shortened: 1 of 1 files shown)


matklad:
Ahhh, good catch!

I _think_ this should use `has_clean` perhaps? That is, replica has bth the header and the prepare itself!

jorangreef:
I was also just thinking of has_clean. That could also race though
(receive header, repair, then receive prepare via ring).

chaitanyabhandari:
Mhmmmmm, I actually can't spot the potential race condition here with using `has_clean`.  Even if we receive some headers from outside the replication ring (for example using a `start_view` that we request during repair), and we repair the header in our in-memory headers, we _would_ set the dirty bit for the corresponding journal slot in `set_header_as_dirty`. We only clear the dirty bit in `write_prepare_on_write_header`, which is _after_ the corresponding prepare is written. So, in the interim, if we were to receive the prepare (via ring replication or if we request it using request_prepare), we are *guaranteed* to forward it along at least once (multiple times even, in case we receive it multiple times before we write it). 

https://github.com/tigerbeetle/tigerbeetle/blob/7140db636c3a4e5adfff963898e7221113c738e1/src/vsr/journal.zig#L1917-L1945

Is there any other case that you had in mind, @jorangreef?

jorangreef:
The race would be if we clear the dirty bit via request_prepare, before
receiving the prepare via ring replication—we would then break ring
replication for subsequent replicas.

So we shouldn't check has/has_clean here at all and simply err on the side
of not breaking ring replication.

There's then a small case where we over-replicate (a replica comes back
online and genuinely repairs its log), but that's a lesser weevil.

chaitanyabhandari:
Agree with erring on the side of safety here!

> The race would be if we clear the dirty bit via request_prepare, before
receiving the prepare via ring replication—we would then break ring
replication for subsequent replicas.

In this case though, we wouldn't "break" ring replication, no? If we receive a prepare via `request_prepare`, we would first forward it along the ring (and then _that_ prepare could continue down the ring) before we write it and clear the dirty bit. IIUC, it wouldn't matter if we're forwarding along a prepare that we receive via ring replication or request_prepare, as long as we _are_ forwarding it at least once? 🤔

jorangreef:
Ah, thanks, I see! I'm convinced.

Let's use has_clean!

chaitanyabhandari (approved):

PR #2621, opened 2024-12-29
VSR: Ratchet table state-sync progress

sentientwaffle:
## Context

If a replica is lagging behind and state syncs, it often won't finish syncing the tables before the cluster has checkpointed again. And that next checkpoint will kick off table sync all over again.

## Fix

When we kick off a new state sync, don't abandon the progress we have made so far syncing tables. Instead, continue with the old table sync, and then when we finish the old table sync, switch to the new table sync.

(Note that right now we (still) don't persist this incremental content-state sync progress -- if the replica restarts before sync completes, it needs to redo all sync work.)

<details>
<summary>Aside about iterator order</summary>

I was confused why `ForestTableIterator` iterates from level 0 to max rather than starting at the bottom levels, since the latter would prioritize syncing old tables.

Luckily there is already a comment explaining this:

> The iterator must traverse from the top (level 0) to the bottom of each tree to avoid skipping tables that are compacted with move-table.

</details>

### DIFF
--- a/src/lsm/forest.zig
+++ b/src/lsm/forest.zig
@@ -613,6 +613,28 @@ pub fn ForestType(comptime _Storage: type, comptime groove_cfg: anytype) type {
             }
         }
 
+        /// Returns whether the forest contains this table (ignoring snapshot_max) at any level.
+        pub fn contains_table(
+            forest: *const Forest,
+            table: *const schema.ManifestNode.TableInfo,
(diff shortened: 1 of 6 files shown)


sentientwaffle:
Closing this -- I have resumed work on it but I think I found a better approach to this so it will be in a different PR.

PR #2622, opened 2024-12-29
vsr: warn on dropped prepares

matklad:


cb22 (dismissed):

src/message_bus.zig
```diff
@@ -728,11 +728,16 @@ fn MessageBusType(comptime process_type: vsr.ProcessType) type {
...
                 assert(connection.fd != IO.INVALID_SOCKET);
 
+                var count: u32 = 0;
                 while (connection.parse_message(bus)) |message| {
                     defer bus.unref(message);
 
+                    count += 1;
                     connection.on_message(bus, message);
                 }
+                if (count > constants.bus_message_burst_warn_min) {
```
cb22, line 738: This might spam if it's a big burst

matklad: This is after the loop, so it's _once_ per burst. 

It might or might not be spammy depending on how often we get bursts, but that's exactly what I am curious about. If we do get large bursts often, then perhaps we need to finally add that "submit all IO __now__" API to IO and call that in-between the messages.

matklad:
Also added a warn when we drop a prepare without writing it to disk due to running out of IO, which is downstream from the incoming burst from the message bus, but is still worth capturing.

cb22 (approved):

PR #2623, opened 2024-12-29
vsr: optimize the order of repair of prepares

matklad:
Instead repairing strictly chronologically, start with commit_min+1, so that we can commit further.

This is helpful in the following scenario:

- the cluster is under high load,
- in particular, the volume of prepares a backup gets through the network overwhelms its disk,
- so a backup doesn't write _all_ incoming prepares to disk
- but still commits them, as they are added to the pipeline_cache

In other words, a replica can end up in a situation where it misses a lot of prepares prior to its own commit min! Before this commit, what would happen is that the replica would repair all those prepares before anything that it actually needs to commit further.

Now, we always try to repair `commit_min+1` first.

### DIFF
--- a/src/vsr/replica.zig
+++ b/src/vsr/replica.zig
@@ -7076,26 +7076,45 @@ pub fn ReplicaType(
                 return;
             }
 
-            // Repair prepares in chronological order. Older prepares will be overwritten by the
-            // cluster earlier, so we prioritize their repair. This also encourages concurrent
-            // commit/repair.
-            var op = op_min;
-            while (op <= self.op) : (op += 1) {
-                const slot = self.journal.slot_with_op(op).?;
-                if (self.journal.dirty.bit(slot)) {
-                    // Rebroadcast outstanding `request_prepare` every `repair_timeout` tick.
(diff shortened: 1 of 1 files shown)


src/vsr/replica.zig
```diff
@@ -7076,26 +7076,45 @@ pub fn ReplicaType(
...
-                        if (budget == 0) {
-                            log.debug("{}: repair_prepares: request budget used", .{self.replica});
-                            return;
+            // Iterate through op_min..=self.op, but make sure to iterate commit_min+1..=self.op
+            // sub range first:
+            // - our first priority is to commit further,
+            // - afterwards, repair commited prepares which are at risk of getting evicted from
+            //   the journal, to help repair any lagging replicas.
+            const repair_ranges_exclusive: [2]struct { u64, u64 } = .{
+                .{ @max(op_min, self.commit_min + 1), self.op + 1 },
```
chaitanyabhandari, line 7085: These checks are so *clean*!

chaitanyabhandari (approved):

PR #2625, opened 2024-12-30
vsr: always repair if header being repaired has the same view as op_head

chaitanyabhandari:
This PR adds an optimization in `repair_header`, wherein we always repair a header if it has the same view as the replica's head op.

Currently, in `repair_header`, _any_ header being repaired must have hash chain connectivity all the way up till the head op.
https://github.com/tigerbeetle/tigerbeetle/blob/7140db636c3a4e5adfff963898e7221113c738e1/src/vsr/replica.zig#L6743-L6753

If `repair_header` returns `False`, we skip repairing the corresponding prepare:
https://github.com/tigerbeetle/tigerbeetle/blob/7140db636c3a4e5adfff963898e7221113c738e1/src/vsr/replica.zig#L1980-L1985

While this check is critical for safety in cases where the cluster has removed certain prepares through view changes (so that we don't install an old, removed header in place of a potentially committed, fresh one), it is suboptimal for cases where the cluster hasn't gone through a view change recently. In these cases (where the header being repaired has the same view number as the head op), we can safely install the header without checking hash chain connectivity, as the primary of the view would've already done so, in `on_prepare`:
https://github.com/tigerbeetle/tigerbeetle/blob/7140db636c3a4e5adfff963898e7221113c738e1/src/vsr/replica.zig#L1709-L1713

This is especially useful in cases where a replica is trying to repair a prepare and has gaps in its log suffix (for example, if the replica misses receiving some prepares, or advances its head op using a start_view, leaving gaps in its journal). Currently, a replica with gaps in its journal would *first* have to repair all of its headers, and *then* its prepares. By that time, that replica may have already have already advanced it head, and may have new gaps in its journal! This could lead to a vicious state where such a replica is stuck committing an old missing prepare in the journal, as it can't be repaired.

### DIFF
--- a/src/vsr/journal.zig
+++ b/src/vsr/journal.zig
@@ -515,27 +515,19 @@ pub fn JournalType(comptime Replica: type, comptime Storage: type) type {
             return op;
         }
 
-        pub fn has(journal: *const Journal, header: *const Header.Prepare) bool {
+        pub fn has_header(journal: *const Journal, header: *const Header.Prepare) bool {
             assert(journal.status == .recovered);
             assert(header.command == .prepare);
             assert(header.operation != .reserved);
 
-            const slot = journal.slot_for_op(header.op);
-            const existing = &journal.headers[slot.index];
-            if (existing.operation == .reserved) {
-                return false;
+            if (journal.header_with_op_and_checksum(header.op, header.checksum)) |_| {
+                return true;
             } else {
-                if (existing.checksum == header.checksum) {
-                    assert(existing.checksum_body == header.checksum_body);
-                    assert(existing.op == header.op);
-                    return true;
-                } else {
-                    return false;
-                }
+                return false;
             }
         }
 
-        pub fn has_clean(journal: *const Journal, header: *const Header.Prepare) bool {
+        pub fn has_prepare(journal: *const Journal, header: *const Header.Prepare) bool {
             if (journal.slot_with_op_and_checksum(header.op, header.checksum)) |slot| {
                 if (!journal.dirty.bit(slot)) {
                     assert(journal.prepare_inhabited[slot.index]);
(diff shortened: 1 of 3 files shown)


src/vsr/journal.zig
```diff
@@ -515,27 +515,19 @@ pub fn JournalType(comptime Replica: type, comptime Storage: type) type {
...
             assert(journal.status == .recovered);
             assert(header.command == .prepare);
             assert(header.operation != .reserved);
 
-            const slot = journal.slot_for_op(header.op);
-            const existing = &journal.headers[slot.index];
-            if (existing.operation == .reserved) {
-                return false;
+            if (journal.header_with_op_and_checksum(header.op, header.checksum)) |_| {
+                return true;
```
chaitanyabhandari, line 524: Took the opportunity to also refactor `has_header`, I observed that its logic (barring the `assert(existing.checksum_body == header.checksum_body);`) was _exactly_ the same as `header_with_op_and_checksum`!

src/vsr/replica.zig
```diff
@@ -6763,7 +6763,12 @@ pub fn ReplicaType(
             assert(header.op < self.op or
                 self.journal.header_with_op(self.op).?.checksum == header.checksum);
 
-            if (!self.repair_header_would_connect_hash_chain(header)) {
+            if (self.journal.header_with_op(self.op).?.view == header.view) {
+                // Fast path for cases where the header being replaced is from the same view
+                // as the head. In this case, we can skip checking if our hash chain connects
+                // up till the head, as the primary for that view would have already done so in
+                // `on_prepare` (by invoking `panic_if_hash_chain_would_break_in_the_same_view`).
```
matklad, line 6770: Oh, this is even better than what I thought to do here! My idea was to check that `header.view == replica.view`, but comparing it with `header_head.view` is strictly better!

🥳

matklad:
VOPR light is green at 192 165, vopr hits pre-existing liveness

matklad (approved):

PR #2627, opened 2024-12-30
vsr: relax defense-in-depth for committing

matklad:
Instead of manually verifying the entire hash-chain ourselves (for which we must have all the prepares), first check if the views tell us that the primary already did the verification.

### DIFF
--- a/src/vsr/replica.zig
+++ b/src/vsr/replica.zig
@@ -5222,15 +5222,6 @@ pub fn ReplicaType(
                     log.debug("{}: on_{s}: ignoring (view change)", .{ self.replica, command });
                     return true;
                 },
-                .request_headers, .request_prepare, .request_reply => {
-                    if (self.primary_index(self.view) != message.header.replica) {
-                        log.debug("{}: on_{s}: ignoring (view change, requested by backup)", .{
-                            self.replica,
-                            command,
-                        });
-                        return true;
-                    }
-                },
                 .headers => {
                     if (self.primary_index(self.view) != self.replica) {
                         log.debug("{}: on_{s}: ignoring (view change, received by backup)", .{
@@ -5245,6 +5236,11 @@ pub fn ReplicaType(
                         });
                         return true;
                     }
+                },
+                .request_headers, .request_prepare, .request_reply => {
+                    // on_headers, on_prepare, and on_reply have the appropriate logic to handle
+                    // incorrect headers, prepares, and replies.
+                    return false;
                 },
                 else => unreachable,
             }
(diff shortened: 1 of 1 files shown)


src/vsr/replica.zig
```diff
@@ -3920,16 +3920,23 @@ pub fn ReplicaType(
...
             if (self.commit_min < self.commit_max and self.commit_min < self.op) {
                 const op = self.commit_min + 1;
-                const header = self.journal.header_with_op(op).?;
+                const header = self.journal.header_with_op(op) orelse return .ready;
+
+                // Assuming that the head op is correct, it is definitely safe to commit the next
+                // prepare if it is from the same view as the head --- the primary for that view
+                // made sure that the hash chain is valid. If it is from the different view, we
+                // additionally verify ourselves that the hash-chain is not broken
+                const safe_to_commit = self.valid_hash_chain(@src()) or (self.status == .normal and
```
jorangreef: Further restrict this to backups?

jorangreef:
```suggestion
                const valid_hash_chain_or_same_view = self.valid_hash_chain(@src()) or (self.status == .normal and
```

matklad: It is restricted to backups, we are under `assert(!(self.status == .normal and self.primary()));`

jorangreef: Yes, I spotted that, but figured it might be nice to make the local intention clearer, this should only be used by backups in normal, not a primary in normal (or we expect the hash chain to be valid).

src/vsr/replica_test.zig
```diff
@@ -1696,10 +1696,13 @@ test "Cluster: view_change: DVC header doesn't match current header in journal"
...
 }
 
-test "Cluster: view_change: lagging replica repairs WAL using start_view from potential primary" {
+test "Cluster: view_change: lagging replica state syncs using start_view from potential primary" {
     // It could be the case that the replica with the most advanced checkpoint has a corruption in
     // its grid. In this case, a replica on an older checkpoint can use a start_view message from
-    // the most up-to-date replica to repair its WAL, advance its checkpoint, and become primary.
+    // the most up-to-date replica to state sync, advance its checkpoint, and become primary. We
+    // make one core assumption about the corrupted grid block that is preventing the up-to-date
+    // replica from becoming primary -- it is a block that the lagging replica *already* has,
```
chaitanyabhandari: I am not very happy with this assumption... However, since we don't have any granular control over corrupting specific blocks that a replica needs for compaction, I wasn't able to come up with a better way to encode our expectation except through a comment. @sentientwaffle thoughts?

sentientwaffle (approved):

PR #2628, opened 2024-12-30
VSR: Request extra headers during repair

sentientwaffle:
When our journal has a break in the headers, don't repair the breaks serially. Instead request as many headers as possible, to hopefully repair multiple breaks.

Suppose a backup's journal header suffix has a pattern like: `1 _ 3 _ 5 _ 7` (where `_` indicates a gap where we don't know the header). Instead of requiring three separate `request_headers`/`headers` round-trips, we can do it in one.

This alternating header/gap pattern is common in a cluster of three when the ring replication direction alternates and one of the backups is unavailable.

### DIFF
--- a/src/vsr/replica.zig
+++ b/src/vsr/replica.zig
@@ -6540,7 +6540,10 @@ pub fn ReplicaType(
                             .command = .request_headers,
                             .cluster = self.cluster,
                             .replica = self.replica,
-                            .op_min = range.op_min,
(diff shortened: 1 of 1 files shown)


matklad (approved):
Makes sense to me!

A further improvement might be to get rid of the range encoding and just include the op numbers for all the headers we want explicitly, but that's even perhaps unnecessary --- likely, missing headers are somewhat near to the head op anyway

PR #2457, opened 2024-11-06
vsr: get rid of awaiting_checkpoint sync state

matklad:

PR #2629, opened 2024-12-31
VSR: get rid of awaiting_checkpoint sync state

sentientwaffle:
Building on https://github.com/tigerbeetle/tigerbeetle/pull/2457.

(The first commit in this PR is #2457's squashed and rebased commits.)

---

Previously, state sync required receiving _two_ SV messages:

- the first SV message triggers grid cancelation, which is asynchronous, and
- the second SV message, received after cancelation completes, does the
  actual sync work.

This adds considerable delay, especially since the replica doesn't sync tables while in `awaiting_checkpoint`.

To fix this, make `on_start_view` asynchronous: if a replica receives SV that requires state sync, it _stores_ this particular SV message, and resumes its processing after cancelation is done.

<details>
<summary>Trace</summary>

In this trace, the blue bars each ~5s of `awaiting_checkpoint`, on a replica that is falling behind a (rapidly committing/checkpointing) cluster:

chrome___tracing_

</details>

### DIFF
--- a/src/vsr/replica.zig
+++ b/src/vsr/replica.zig
@@ -1289,6 +1289,8 @@ pub fn ReplicaType(
                 self.commit_prepare = null;
             }
 
+            if (self.sync_start_view) |message| self.message_bus.unref(message);
+
             var grid_reads = self.grid_reads.iterate();
             while (grid_reads.next()) |read| self.message_bus.unref(read.message);
 
@@ -1399,9 +1401,7 @@ pub fn ReplicaType(
                     // Ignore further messages until finishing (asynchronous) processing of sync SV.
                     // Notably, this prevents our view from jumping ahead of SV.
                     assert(self.sync_start_view != null);
-                    log.debug("{}: on_message: ignoring (syncing)", .{
-                        self.replica,
-                    });
+                    log.warn("{}: on_message: ignoring (syncing)", .{self.replica});
                     return;
                 },
                 .updating_checkpoint => {},
@@ -2234,6 +2234,22 @@ pub fn ReplicaType(
             assert(message.header.command == .start_view);
             if (self.ignore_view_change_message(message.base_const())) return;
 
+            if (self.status == .recovering_head) {
+                if (message.header.view > self.view or
+                    message.header.op >= self.op_prepare_max() or
+                    message.header.nonce == self.nonce)
+                {
+                    // This SV is guaranteed to have originated after the replica crash,
+                    // it is safe to use to determine the head op.
+                } else {
+                    log.mark.debug(
+                        "{}: on_start_view: ignoring (recovering_head, nonce mismatch)",
+                        .{self.replica},
+                    );
+                    return;
+                }
+            }
+
             assert(self.status == .view_change or
                 self.status == .normal or
                 self.status == .recovering_head);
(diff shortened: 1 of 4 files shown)


src/vsr/replica.zig
```diff
@@ -2211,13 +2228,34 @@ pub fn ReplicaType(
...
+                    // This SV is guaranteed to have originated after the replica crash,
+                    // it is safe to use to determine the head op.
+                } else {
+                    log.mark.debug(
+                        "{}: on_start_view: ignoring (recovering_head, nonce mismatch)",
+                        .{self.replica},
+                    );
+                    return;
+                }
+            }
```
sentientwaffle, line 2251: I pulled the `nonce mismatch` check up out of `ignore_view_change_message()` because otherwise:

1. While in `status=recovering_head`, receive a SV message from a newer view (with nonce=0).
2. Since `message.header.view > self.view` we accept it and set `self.view = message.header.view`.
3. Then we call `on_start_view_set_journal(message)` (still with that same SV).
4. Only now the views match, so we require that the nonces match as well -- we panic on `assert(!self.ignore_view_change_message(message.base_const()))`.

src/vsr/replica.zig
```diff
@@ -2272,75 +2337,69 @@ pub fn ReplicaType(
...
-                // with the one from this SV.
+            // Cluster is at least two checkpoints ahead. Although SV's checkpoint is not guaranteed
+            // be durable on a quorum of replicas, it is safe to sync to it, because prepares in
+            // this replica's WAL are no longer needed.
+            const far_behind = vsr.Checkpoint.durable(self.op_checkpoint_next() +
+                constants.vsr_checkpoint_ops, message.header.commit_max);
+            // Cluster is on the next checkpoint, and that checkpoint is durable and is safe to
+            // sync to. Try to optimistically avoid state sync and prefer WAL repair, unless
+            // there's evidence that the repair can't be completed.
+            const likely_stuck = self.syncing == .idle and self.repair_stuck();
```
matklad, line 2352: This I think would be the right place to implement that other idea where a replica which syncs tables triggers state sync more aggressively, to avoid the situation where it waits for an entire `repair_sync_timeout` before triggering the sync, all while the cluster is making progress in the new checkpoint already.

sentientwaffle: It is the right place, but I think I will include the actual change in https://github.com/tigerbeetle/tigerbeetle/pull/2621 instead. The reason is that as of that PR, `sync_tables` will get set earlier (immediately after we sync the checkpoint, rather than after state-machine open), so checking `sync_tables != null` can be used to answer the "are we already busy syncing" question.

sentientwaffle:
https://github.com/tigerbeetle/tigerbeetle/pull/2629/commits/ac89339d5a0a104c15e938f8617bd6dcd2d26136 is (afaict) a latent journal bug. I'm not 100% clear why we didn't discover it before -- I guess this PR allows view changes to be faster in some circumstances.

Also interesting to note that vopr found it, but not vopr-lite, even after many seeds.

src/vsr/replica.zig
```diff
@@ -1388,6 +1395,18 @@ pub fn ReplicaType(
...
+                .idle => {},
+                .canceling_commit, .canceling_grid => {
+                    // Ignore further messages until finishing (asynchronous) processing of sync SV.
+                    // Notably, this prevents our view from jumping ahead of SV.
+                    assert(self.sync_start_view != null);
+                    log.warn("{}: on_message: ignoring (syncing)", .{self.replica});
+                    return;
+                },
+                .updating_checkpoint => {},
+            }
```
matklad, line 1408: TBH, this feels like a cludge to me, as it violates "embrace concurrency" principle. But also, a replica in the middle of `on_start_view` is in a rather weird state, and carefully handling that was hard.

matklad (approved):
LGTM-ish!

I am still not happy with the overall flow here, it feels complicated, but it's not I was able to come up with anything better while I was sitting on this code!

PR #1133, opened 2023-08-15
IO: Async disk I/O on Windows

kprotty:
Enables `ReadFile` and `WriteFile` to use Overlapped IO instead of being synchronous.
```sh
info(main): 0: Allocated 3304MiB during replica init
info(main): 0: Grid cache: 1024MiB, LSM-tree manifests: 128MiB
info(main): 0: cluster=0: listening on 127.0.0.1:65448
info: Benchmark running against { 127.0.0.1:65448 }
info: Benchmark seed = 42

# main
1260 batches in 58.23 s
load offered = 1000000 tx/s
load accepted = 171742 tx/s
batch latency p1 = 0 ms
batch latency p10 = 5 ms
batch latency p20 = 6 ms
batch latency p30 = 16 ms
batch latency p40 = 23 ms
batch latency p50 = 29 ms
batch latency p60 = 50 ms
batch latency p70 = 59 ms
batch latency p80 = 67 ms
batch latency p90 = 79 ms
batch latency p95 = 105 ms
batch latency p99 = 331 ms
batch latency p100 = 352 ms

# this PR
1269 batches in 36.72 s
load offered = 1000000 tx/s
load accepted = 272351 tx/s
batch latency p1 = 0 ms
batch latency p10 = 5 ms
batch latency p20 = 6 ms
batch latency p30 = 9 ms
batch latency p40 = 16 ms
batch latency p50 = 20 ms
batch latency p60 = 26 ms
batch latency p70 = 29 ms
batch latency p80 = 32 ms
batch latency p90 = 35 ms
batch latency p95 = 41 ms
batch latency p99 = 327 ms
batch latency p100 = 345 m
```

PR #2630, opened 2024-12-31
IO: Async disk I/O on Windows

cb22:
This replaces #1133, which did all the heavy lifting. It's a rebase, a small cleanup, and some modifications to make overlapped IO work with NtCreateFile instead of `kernel32.CreateFileW`.

There's also a bonus fix which makes tigerbeetle run under `wine` again!


batiati (approved):
> There's also a bonus fix which makes tigerbeetle run under wine again!

:heart:

PR #2635, opened 2025-01-06
replica: add in star replication and closed loop replication

cb22:
# Background
When running high control-plane load against a cluster (most easily simulated by tiny batches within the benchmark), it's possible for the replicas that don't form part of the replication quorum to start lagging behind and enter a state sync loop.

This is somewhat by design for failure tolerance: replication in TigerBeetle effectively consists of two phases:
1. a closed loop wait for a replication quorum and,
2. open loop replication to the remaining replicas.

Any latency in the first gets exposed directly to the client, while the second does not. This causes some problems here, however, especially when combined with a ring topology that effectively favors replicas 'closer' to the primary in the ring.

The open loop system is not nearly as stable as it should be; this can be tested by starting a benchmark like above and slightly perturbing (eg, `timeout 0.2 stress -d 10`) only one replica. 

What happens is that a combination of factors (#2626, #2627, dropping messages on the floor when no journal IOPS are available, slow repair vs prepare) lead to the replica falling behind and entering state sync when it shouldn't be, and since the cluster is loaded, it currently has no hope of exiting a state-sync -> repair -> state-sync loop. (#2621).

While that loop doesn't directly affect the cluster, the added IO load on what is already a loaded cluster, can increase latencies significantly.

# The Workaround
This PR adds two new options: `--replicate-star` which switches to a star topology, and `--replicate-closed-loop` which enforces closed loop replication on the entire cluster.

These options are *highly* experimental, designed to work around for a specific use case, and are going to be removed in future.

Switching to a star topology places additional network load on the primary, but mitigates the impact a single break in the ring might have, which is currently unoptimized. 

(Before #2627 it was also a critical piece of the puzzle as to why replicas could fall behind so easily: only prepares get ring replicated, so other replicas can learn about a newer committed op, then refuse to commit any prepares until they've received _all_ prepares up to the committed op they know about.)

Switching to closed loop makes the primary wait until _all_ replicas have acknowledged a prepare, with an escape hatch built into the `prepare_timeout`: if it takes more than the current 25 ticks / ~250ms, the whole pipeline will be unblocked if they have received at least a `quorum_replication` number of `prepare_ok`s.

This helps prevent cluster unavailability if a single replica goes down, at the cost of increased latency if it does.

### DIFF
--- a/src/vsr/replica.zig
+++ b/src/vsr/replica.zig
@@ -7451,6 +7451,14 @@ pub fn ReplicaType(
             assert(!self.journal.has_prepare(message.header));
             assert(message.header.op > self.commit_min);
 
+            if (self.replicate_options.star) {
+                if (self.status == .normal and self.primary()) {
+                    self.send_message_to_other_replicas_and_standbys(message.base());
+                }
+
+                return;
+            }
+
             const next = next: {
                 // Replication in the ring of active replicas.
                 if (!self.standby()) {
--- a/src/vsr/replica.zig
+++ b/src/vsr/replica.zig
@@ -114,6 +114,11 @@ pub fn ReplicaType(
         const Clock = vsr.ClockType(Time);
         const ForestTableIterator = ForestTableIteratorType(StateMachine.Forest);
 
+        const ReplicateOptions = struct {
+            closed_loop: bool = false,
+            star: bool = false,
+        };
+
         const BlockRead = struct {
             read: Grid.Read,
             replica: *Replica,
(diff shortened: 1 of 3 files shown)


src/vsr/replica.zig
```diff
@@ -1831,8 +1855,10 @@ pub fn ReplicaType(
             assert(self.prepare_timeout.ticking);
             assert(self.primary_abdicate_timeout.ticking);
             assert(!self.primary_abdicating);
-            if (self.primary_pipeline_pending()) |_| {
-                if (prepare_pending == prepare) self.prepare_timeout.reset();
+            if (self.primary_pipeline_pending()) |prepare_pending| {
+                if (prepare_pending == prepare) {
+                    self.prepare_timeout.reset();
+                }
```
cb22, line 1861: This logic should be unchanged, just a drive-by cleanup.

chaitanyabhandari (approved):

PR #2637, opened 2025-01-10
vopr: fix multiple false positives

chaitanyabhandari:
This PR fixes the following two liveness VOPR false positives.

---

The first one is a new one that showed after https://github.com/tigerbeetle/tigerbeetle/pull/2623. It manifested as a case where a replica had multiple missing prepares, some smaller than `commit_min` & some bigger than `commit_min`. As per our new repair logic, we first repair missing prepares *after* `commit_min`, and then we get to prepares *before* commit_min. However, as per our current `core_missing_prepare` logic, we iterate from op_repair_min → head_op, erroneously find a missing prepare before commit_min (which a replica may not get to repair due to the repair budget), and then bail out due to the below logic. 
https://github.com/tigerbeetle/tigerbeetle/blob/4b7740417cde0668f9563c65157ecd966b5626c4/src/vopr.zig#L749-L757

The solution here is for `core_missing_prepare` to mirror our repair logic, traversing (commit_min, head_op] first, and then [op_repair_min, commit_min].

---

The second one is a long standing issue caused by the replica upgrade logic in `src/testing/cluster.zig`. Specifically, `replica_release_execute` disables all existing storage faults in the WAL (by marking `faulty` as `False`), but doesn't fix the headers that we write as reserved (when a particular corrupt prepare is not repaired but a *nearby* corrupt prepare is repaired, a replica marks the header for corrupt prepare as reserved, so it is sticky across restarts). Due to this logic, too many replicas may transition into recovering_head, rendering the cluster *correctly* unavailable. However, this is not correctly detected by the VOPR as of now.
https://github.com/tigerbeetle/tigerbeetle/blob/4b7740417cde0668f9563c65157ecd966b5626c4/src/testing/cluster.zig#L644-L650

The solution here is simple, at the end of `safety` mode, we now check if more than a view change quorum of replicas are in .recovering_head status. If so, we consider the cluster to be correctly unavailable. 

https://github.com/tigerbeetle/tigerbeetle/pull/2216 was a previous attempt at solving this issue but we [closed it because](https://github.com/tigerbeetle/tigerbeetle/pull/2216#issuecomment-2287104782) we didn't want to add logic to the fix the WAL to run in replica tests.

### DIFF
--- a/src/vopr.zig
+++ b/src/vopr.zig
@@ -657,11 +657,7 @@ pub const Simulator = struct {
     }
 
     /// The core contains at least a view-change quorum of replicas. But if one or more of those
-    /// replicas are in status=recovering_head (due to corruption or state sync), then that may be
-    /// insufficient.
-    /// TODO: State sync can trigger recovering_head without any crashes, and we should be able to
-    /// recover in that case.
-    /// (See https://github.com/tigerbeetle/tigerbeetle/pull/933#discussion_r1245440623)
+    /// replicas are in status=recovering_head (due to corruption), then that may be insufficient.
     pub fn core_missing_quorum(simulator: *const Simulator) bool {
         assert(simulator.core.count() > 0);
 
@@ -678,6 +674,20 @@ pub const Simulator = struct {
 
         const quorums = vsr.quorums(simulator.options.cluster.replica_count);
         return quorums.view_change > core_replicas - core_recovering_head;
+    }
+
+    fn smallest_missing_prepare_between(
+        simulator: *const Simulator,
+        replica: *const Cluster.Replica,
+        op_min: u64, // Inclusive
+        op_max: u64, // Inclusive
+    ) ?u64 {
+        assert(op_min <= op_max);
+        for (op_min..op_max + 1) |op| {
+            const header = simulator.cluster.state_checker.header_with_op(op);
+            if (!replica.journal.has_prepare(&header)) return op;
+        }
+        return null;
     }
 
     // Returns a header for a prepare which can't be repaired by the core due to storage faults.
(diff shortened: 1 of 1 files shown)


chaitanyabhandari:
We had also discussed that https://github.com/tigerbeetle/tigerbeetle/issues/696 would resolve this issue. I do want to pick it up when my pipeline clears up a bit, but I believe there is still some value in getting the VOPR bugs down to zero (as long as this fix seems elegant enough, of course). For instance, "power of zero" would instill more confidence in the current state of the code. Also, it could save us quite a bit of effort debugging failed VOPR seeds to find out if its an existing one or a new one. 

FYI, with this and https://github.com/tigerbeetle/tigerbeetle/pull/2600, we've been VOPR green for over three days!
Screenshot 2025-01-13 at 12 10 49 AM

src/vopr.zig
```diff
@@ -680,6 +676,19 @@ pub const Simulator = struct {
         return quorums.view_change > core_replicas - core_recovering_head;
     }
 
+    fn smallest_missing_prepare_between(
+        simulator: *const Simulator,
+        replica: *const Cluster.Replica,
+        op_min: u64, // Inclusive
+        op_max: u64, // Inclusive
+    ) ?u64 {
+        for (op_min..op_max + 1) |op| {
```
matklad, line 703: Missing an obvious assert here: `assert(op_min <= op_max)`!

chaitanyabhandari: Addressed in fddd1a12e55fa5ec865b00f9005bd3c1e30dd57d!

matklad (approved):

PR #2639, opened 2025-01-11
[Go] Run lint during CI

batiati:
Run [golangci-lint](https://golangci-lint.run/) during CI.

Fixes the violation reported on #2598:  "_the goroutine calls T.Fatal, which must be called in the same goroutine as the test_" by standardizing error checking with `assert.True(t, err == null)` instead of `if err != null t.Fatal()` across all tests.

 Thanks, @AskAlexSharov, for pointing it out!

### DIFF
--- a/src/clients/go/ci.zig
+++ b/src/clients/go/ci.zig
@@ -15,6 +15,8 @@ pub fn tests(shell: *Shell, gpa: std.mem.Allocator) !void {
         return error.GoFmt;
     }
 
+    try shell.exec("go vet", .{});
+
     // `go build`  won't compile the native library automatically, we need to do that ourselves.
     try shell.exec_zig("build clients:go -Drelease", .{});
     try shell.exec_zig("build -Drelease", .{});
--- a/src/clients/go/tb_client_test.go
+++ b/src/clients/go/tb_client_test.go
@@ -423,9 +423,9 @@ func doTestClient(t *testing.T, client Client) {
 				})
 
 				if err != nil {
-					t.Fatal(err)
+					t.Error(err)
+					return
 				}
-
 				assert.Empty(t, results)
 			}(i)
 		}
@@ -479,7 +479,8 @@ func doTestClient(t *testing.T, client Client) {
 				})
 
 				if err != nil {
-					t.Fatal(err)
+					t.Error(err)
+					return
 				}
 
 				if i%10 == 0 {


src/clients/go/ci.zig
```diff
@@ -15,6 +15,27 @@ pub fn tests(shell: *Shell, gpa: std.mem.Allocator) !void {
...
     }
 
+    if (builtin.target.os.tag == .linux) {
+        log.info("golangci-lint run", .{});
+
+        const image = "golangci/golangci-lint";
+        const tag = "latest";
+        for (0..5) |attempt| {
+            if (attempt > 0) std.time.sleep(1 * std.time.ns_per_min);
+            if (shell.exec("docker image pull {image}:{tag}", .{ .image = image, .tag = tag })) {
```
matklad: Hm, go produces statically-liked binaries, so there isn't a material reason to use docker here, on top of our usual avoidance of docker. If we are to run golangci-lint, I'd rather we find some other way to install it.  I think it's fine to punt on github actions for this? They have some action for installing it. 


Though, I am not sure if we want `golangci-lint`, as this is third-party tooling. We _should_ be running `go vet` if we aren't already, but I'd suggest maybe holding of golangci-lint until it is clear that it is useful? 

The present error, while a true one, isn't really a big issue --- worst case, it'll lead to test failing incorrectly.

src/clients/go/tb_client_test.go
```diff
@@ -118,9 +118,7 @@ func doTestClient(t *testing.T, client Client) {
 			accountA,
 			accountB,
 		})
-		if err != nil {
-			t.Fatal(err)
-		}
+		assert.True(t, err == nil)
```
matklad: A minor note is that `assert.True(t, err == nil)` erases the information about _which_ error occurred. It'll tell you that `true != false`, not that you got some specific error. I don't think this is a big deal, but, if we go to the pain of using a third-party assertion library, we might as well actually use it? I think the right incarnation here would be

```
assert.Nil(err)
```

matklad: But I think I'd stick with `Fatal` here and elsewhere, except the specific cases. 

`Fatal` stops the test here and now, `assert.True`, `assert.Nil` will continue the test. But it doesn't make sense to continue the test if we weren't able to create accounts?

So I suggest maybe just keeping `t.Fatal`, and replacing it with `t.Error` when we spawn coroutines?

src/clients/go/ci.zig
batiati: TIL about `go vet`, and no, we don't run it already!
Switching to it since it was able to catch this present issue.

src/clients/go/tb_client_test.go
batiati: > But it doesn't make sense to continue the test if we weren't able to create accounts?

IMO it kind of make sense to `assert.True(err == nil)` instead of aborting the test, the same happens if `CreateAccounts` fails and the test continues regardless `assert.Empty`.

> So I suggest maybe just keeping t.Fatal, and replacing it with t.Error when we spawn coroutines?

Agreed, let's keep it simple.

matklad (approved):

src/clients/go/tb_client_test.go
matklad: Yeah, no strong opinion on whether we should abort the test or continue. I do have a strong opinion that `assert.Nil(err)` is better than `assert.True(err == nil)`, as that would be a strict improvement (Nil also continues).

PR #2641, opened 2025-01-13
cli: add tigerbeetle inspect constants command

matklad:
This command is intended to help with napkin math --- things like length of the journal, frequency of checkpoints, size of a sector are important to understand while investigating real-world behavior. While all them _can_ be derived manually from `constants.zig`, substituting the appropriate process config, this requires non-trivial and annoying mental effort.

The idea of `inspect constants` is to present the information in a convenient for a human developer or operator form.

Right now, to keep things simple, this only prints the layout of the data file. But, with subcommand in-place, activation energy to print more things is low, so I expect we'll add a bunch of stuff here soon.

In particular, this perhaps can subsume `Checkpoint ops diagram` test.

Present output:

```
λ ./zig/zig build run -- inspect constants
Data File Layout:
superblock          96KiB
  copy              24KiB x4

wal_headers         256KiB
  sector            4KiB x64
    header          256B x16

wal_prepares        1GiB
  prepare           1MiB x1024

client_replies      64MiB
  reply             1MiB x64

grid_padding        160KiB

grid                elastic
  block             512KiB
```

### DIFF
--- a/src/tigerbeetle/cli.zig
+++ b/src/tigerbeetle/cli.zig
@@ -140,6 +140,7 @@ const CLIArgs = union(enum) {
 
     // Experimental: the interface is subject to change.
     const Inspect = union(enum) {
+        constants,
         superblock: struct {
             positional: struct { path: []const u8 },
         },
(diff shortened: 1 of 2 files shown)


sentientwaffle (approved):
LGTM. Also note that we have `tigerbeetle version --verbose` which prints all comptime `config.process`/`config.cluster` parameters.

matklad:
Yeah, I considered whether this should be folded into `version --verbose`, but it feels better to have a separate command here. `version --verbose` lists the ground facts. `inspect constants` lists a tasteful selection of derived facts, it is intentionally biased for humans!

PR #2648, opened 2025-01-15
ci: run devhub in ubuntu-22.04

cb22:
kcov has been removed from Ubuntu 24.04, which is the new `ubuntu-latest`. This is a temporary workaround!


matklad (approved):

matklad:
Got a spare moment at the airport to reduce our dependency to just kcov, instead of the entirety of Ubuntu:

https://github.com/tigerbeetle/tigerbeetle/pull/2649

PR #2651, opened 2025-01-16
VSR: Assert constants.verify at comptime

sentientwaffle:
When we assert `constants.verify`, do so at comptime, not runtime. The intent of these asserts is to make sure that these functions are not even compiled into non-verify builds.

I left one assert unmodified (in `free_set.zig`) to avoid conflicting with https://github.com/tigerbeetle/tigerbeetle/pull/2600.

Tested with `zig build -Dconfig_verify=false`.


chaitanyabhandari (approved):
Thank you for avoiding the conflict! :D

PR #2653, opened 2025-01-17
Journal: Don't write invalid redundant headers during repair

sentientwaffle:
## Overview

To hit this bug:

1. Recover from a WAL with a faulty entry, then
2. repair that entry,
3. then crash and recover again before the WAL wraps and overwrites that entry.

If the bug is hit, the just-repaired slot is recovered as faulty.

## Bug

Suppose we are repairing a faulty prepare:
1. (The dirty/faulty bits are already set.)
2. We start by writing the prepare...
3. Then when the prepare is written, we write the redundant header sector...
4. Then when the redundant header sector is written, we clear the dirty/faulty bits.

The problem is that during step 3, since the faulty bit is set, the sector of redundant headers that we write contains garbage (i.e. "deliberately" invalid headers -- incorrectly so, in this case). This is the code (from `Journal.header_sector()`) which generates the redundant headers to write:

    var i: usize = 0;
    while (i < headers_per_sector) : (i += 1) {
        const slot = Slot{ .index = sector_slot.index + i };

        if (journal.faulty.bit(slot)) {
            // Redundant faulty headers are deliberately written as invalid.
            // This ensures that faulty headers are still faulty when they are read back
            // from disk during recovery. This prevents faulty entries from changing to
            // reserved (and clean) after a crash and restart (e.g. accidentally converting
            // a case `@D` to a `@I` after a restart).
            sector_headers[i] = Header.Prepare.reserved(replica.cluster, i);
            sector_headers[i].checksum = 0; // Invalidate the checksum.
            assert(!sector_headers[i].valid_checksum());
        } else {
            // Write headers from `headers_redundant` instead of `headers` — we need to
            // avoid writing (leaking) a redundant header before its corresponding prepare
            // is on disk.
            sector_headers[i] = journal.headers_redundant[slot.index];
        }
    }

This garbage means that after step 4, if we crash/recover, we can end up with the same slot as faulty, even though we just repaired it.

#### Why didn't the VOPR find this bug?

(Note that with the `JournalChecker` (added in this PR), the VOPR _does_ catch this bug now.)

Redundant headers only influence the test during WAL recovery.

The fault atlas mostly ensures that we don't inject a pattern of WAL faults that the cluster cannot repair.
The caveat of that "mostly" is that the fault atlas doesn't try to prevent `status=recovering_head`.

Instead, when the VOPR restarts a replica, it suppresses storage faults during recovery if the replica entering `status=recovering_head` would possibly deadlock the cluster:

    // To improve VOPR utilization, try to prevent the replica from going into
    // `.recovering_head` state if the replica is needed to form a quorum.
    const fault = recoverable_count >= recoverable_count_min or replica.standby();
    simulator.replica_restart(replica.replica, fault);

Because corrupt WAL prepares can leak into "corrupt" (zeroed) WAL headers, when `replica_restart()` is invoked with faults disabled, it actually repairs the zeroed WAL headers:

    var header_prepare_view_mismatch: bool = false;
    if (!fault) {
        // The journal writes redundant headers of faulty ops as zeroes to ensure
        // that they remain faulty after a crash/recover. Since that fault cannot
        // be disabled by `storage.faulty`, we must manually repair it here to
        // ensure a cluster cannot become stuck in status=recovering_head.
        // See recover_slots() for more detail.
        const headers_offset = vsr.Zone.wal_headers.offset(0);
        const headers_size = vsr.Zone.wal_headers.size().?;
        const headers_bytes = replica_storage.memory[headers_offset..][0..headers_size];
        for (
            mem.bytesAsSlice(vsr.Header.Prepare, headers_bytes),
            replica_storage.wal_prepares(),
        ) |*wal_header, *wal_prepare| {
            if (wal_header.checksum == 0) {
                wal_header.* = wal_prepare.header;
            } else {
                if (wal_header.view != wal_prepare.header.view) {
                    header_prepare_view_mismatch = true;
                }
            }
        }
    }

This bug was hit by the VOPR with [misdirected write injection](https://github.com/tigerbeetle/tigerbeetle/pull/2463) as part of a liveness bug. I ran into it because the restart in question was due to an upgrade (`release_execute()`), and upgrades don't use the above redundant-header-repair.

## Fix

- Rather than ignoring `headers_redundant` during `Journal.header_sector()` when `faulty.bit(slot)`, maintain it properly so `Journal.header_sector()` can just use it as-is. That is, `headers_redundant` always contains "what we would write to the redundant headers, is we were to write to it right now".
- When `set_header_as_dirty` replaces a header, set the corresponding `headers_redundant` to an invalid header.
- During journal recovery, initialize `headers_redundant` with invalid headers corresponding to faulty prepares.

(We still set `headers_redundant` to the proper value in `Journal.write_prepare_header()`, just before generating the sector of redundant headers.)

### DIFF
--- a/src/vsr/journal.zig
+++ b/src/vsr/journal.zig
@@ -236,7 +236,7 @@ pub fn JournalType(comptime Replica: type, comptime Storage: type) type {
 
         /// We copy-on-write to these buffers, as the in-memory headers may change while writing.
         /// The buffers belong to the IOP at the corresponding index in IOPS.
-        headers_iops: *align(constants.sector_size) [
+        write_headers_sectors: *align(constants.sector_size) [
             constants.journal_iops_write_max
         ][constants.sector_size]u8,
 
@@ -321,12 +321,12 @@ pub fn JournalType(comptime Replica: type, comptime Storage: type) type {
             errdefer allocator.free(prepare_inhabited);
             @memset(prepare_inhabited, false);
 
-            const headers_iops = (try allocator.alignedAlloc(
+            const write_headers_sectors = (try allocator.alignedAlloc(
                 [constants.sector_size]u8,
                 constants.sector_size,
                 constants.journal_iops_write_max,
             ))[0..constants.journal_iops_write_max];
-            errdefer allocator.free(headers_iops);
+            errdefer allocator.free(write_headers_sectors);
 
             log.debug("{}: slot_count={} size={} headers_size={} prepares_size={}", .{
                 replica,
@@ -345,7 +345,7 @@ pub fn JournalType(comptime Replica: type, comptime Storage: type) type {
                 .faulty = faulty,
                 .prepare_checksums = prepare_checksums,
                 .prepare_inhabited = prepare_inhabited,
-                .headers_iops = headers_iops,
+                .write_headers_sectors = write_headers_sectors,
             };
 
             assert(@mod(@intFromPtr(&journal.headers[0]), constants.sector_size) == 0);
@@ -369,7 +369,7 @@ pub fn JournalType(comptime Replica: type, comptime Storage: type) type {
             journal.faulty.deinit(allocator);
             allocator.free(journal.headers);
             allocator.free(journal.headers_redundant);
-            allocator.free(journal.headers_iops);
+            allocator.free(journal.write_headers_sectors);
             allocator.free(journal.prepare_checksums);
             allocator.free(journal.prepare_inhabited);
 
@@ -871,12 +871,6 @@ pub fn JournalType(comptime Replica: type, comptime Storage: type) type {
                 return;
             }
 
-            // Check that the `headers` slot belongs to the same op that it did when the read began.
-            // The slot may not match the Read's op/checksum due to either:
-            // * The in-memory header changed since the read began.
-            // * The in-memory header is reserved+faulty; the read was via `prepare_checksums`
-            const slot = journal.slot_with_op_and_checksum(op, checksum);
-
             const error_reason: ?[]const u8 = reason: {
                 if (!message.header.valid_checksum()) {
                     break :reason "corrupt header after read";
@@ -917,7 +911,11 @@ pub fn JournalType(comptime Replica: type, comptime Storage: type) type {
             };
 
             if (error_reason) |reason| {
-                if (slot) |s| {
+                // Check that the `headers` slot belongs to the same op that it did when the read
+                // began. The slot may not match the Read's op/checksum due to either:
+                // * The in-memory header changed since the read began.
+                // * The in-memory header is reserved+faulty; the read was via `prepare_checksums`
+                if (journal.slot_with_op_and_checksum(op, checksum)) |s| {
                     journal.faulty.set(s);
                     journal.dirty.set(s);
                 }
@@ -925,6 +923,8 @@ pub fn JournalType(comptime Replica: type, comptime Storage: type) type {
                 journal.read_prepare_log(op, checksum, reason);
                 callback(replica, null, null);
             } else {
+                assert(message.header.checksum == checksum);
+
                 callback(replica, message, destination_replica);
             }
         }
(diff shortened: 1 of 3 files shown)


src/vsr/journal.zig
```diff
@@ -877,78 +877,56 @@ pub fn JournalType(comptime Replica: type, comptime Storage: type) type {
...
                     journal.faulty.set(s);
                     journal.dirty.set(s);
                 }
 
-                journal.read_prepare_log(op, checksum, "corrupt body after read");
+                journal.read_prepare_log(op, checksum, reason);
                 callback(replica, null, null);
-                return;
+            } else {
+                callback(replica, message, destination_replica);
```
matklad, line 928: Might redundantly `assert(message.header.checksum == checksum)` here, as control-flow is not entirely trivial

src/vsr/journal.zig
```diff
@@ -877,78 +877,56 @@ pub fn JournalType(comptime Replica: type, comptime Storage: type) type {
...
-            }
+                if (!message.header.valid_checksum_body(message.body_used())) {
+                    break :reason "corrupt body after read";
+                }
+                break :reason null;
+            };
 
-            if (!message.header.valid_checksum_body(message.body_used())) {
+            if (error_reason) |reason| {
                 if (slot) |s| {
```
matklad: Maybe

```
if (journal.slot_with_op_and_checksum(op, checksum)) |slot|
```

to minimize the scope of slot?

src/vsr/journal.zig
```diff
@@ -1552,6 +1552,12 @@ pub fn JournalType(comptime Replica: type, comptime Storage: type) type {
                 },
             }
 
+            journal.headers_redundant[slot.index] = journal.headers[slot.index];
+            if (journal.faulty.bit(slot)) {
+                assert(journal.headers_redundant[slot.index].operation == .reserved);
+                journal.headers_redundant[slot.index].checksum = 0; // Invalidate the checksum.
+            }
+
```
matklad, line 1562: Can we 

```zig
assert(journal.faulty.bit(slot) == !journal.headers_redundant[slot.index].valid_checksum())
```

src/vsr/journal.zig
```diff
@@ -1724,6 +1730,7 @@ pub fn JournalType(comptime Replica: type, comptime Storage: type) type {
         }
 
         pub fn set_header_as_dirty(journal: *Journal, header: *const Header.Prepare) void {
+            const replica: *Replica = @alignCast(@fieldParentPtr("journal", journal));
```
matklad: Perhaps it would be better to pass `cluster` as a parameter to `set_header_as_dirty`, or even to journal's init? Reaching out to replica like this feels inelegant.

src/vsr/journal.zig
```diff
@@ -1908,6 +1908,23 @@ pub fn JournalType(comptime Replica: type, comptime Storage: type) type {
             }
 
             const slot = journal.slot_with_header(message.header).?;
+            if (journal.headers_redundant[slot.index].checksum != message.header.checksum) {
+                // Scenario:
+                // 1. write_prepare(h₁)
+                // 2. write_prepare_header(h₁)
+                // 3. remove_entry(h₁)
+                // 4. set_header_as_dirty(h₁)
```
matklad, line 1918: Can we assert that the slot in question is, in fact, dirty at this point?

src/vsr/journal.zig
```diff
@@ -2096,21 +2110,19 @@ pub fn JournalType(comptime Replica: type, comptime Storage: type) type {
...
-                    // Write headers from `headers_redundant` instead of `headers` — we need to
-                    // avoid writing (leaking) a redundant header before its corresponding prepare
-                    // is on disk.
-                    sector_headers[i] = journal.headers_redundant[slot.index];
+                    maybe(journal.faulty.bit(slot));
                 }
+
+                // Write headers from `headers_redundant` instead of `headers` — we need to avoid
+                // writing (leaking) a redundant header before its corresponding prepare is on disk.
+                sector_headers[i] = journal.headers_redundant[slot.index];
```
matklad: Consider doing a bulk copy_disjoint here, rather than doing it one header at a time, to make it clear that what we write is `headers_redundant` as is (but copied to avoid races).

Unrelated, but the `headers_iops` name is a bit confusing, though not sure what would be better. `write_headers_sector`?

matklad (commented):
LGTM overall, just some nits and a couple of missing assertions!

Excellent work, as usual!

src/vsr/journal.zig
sentientwaffle: :bulb:  Aha, we can just use `header.cluster`!

matklad (approved):

PR #2656, opened 2025-01-19
vsr: check some more invariants

matklad:
- add Replica.invariants method which is called on_tick and on_message, as a central spot to verify invariants
- add the basic invariant that the head op is always present
- separately from this infrastructure, assert that obviously correct prepares are added to the journal.

### DIFF
--- a/src/vsr/replica.zig
+++ b/src/vsr/replica.zig
@@ -1635,6 +1635,17 @@ pub fn ReplicaType(
             assert(message.header.replica < self.replica_count);
             assert(message.header.operation != .reserved);
 
+            // Sanity check --- if the prepare is definitely from the current log-wrap, it should be
+            // appended.
+            defer {
+                if (self.status == .normal and message.header.view == self.view and
+                    message.header.op > self.op_checkpoint() and
+                    message.header.op <= self.op_checkpoint_next_trigger())
+                {
+                    assert(self.journal.has_header(message.header));
+                }
+            }
+
             // Replication balances two goals:
             // - replicate anything that the next replica is likely missing,
             // - avoid a feedback loop of cascading needless replication.
--- a/src/vsr/replica.zig
+++ b/src/vsr/replica.zig
@@ -820,6 +820,7 @@ pub fn ReplicaType(
             // Asynchronously open the free set and then the (Forest inside) StateMachine so that we
             // can repair grid blocks if necessary:
             self.grid.open(grid_open_callback);
+            self.invariants();
         }
 
         fn superblock_open_callback(superblock_context: *SuperBlock.Context) void {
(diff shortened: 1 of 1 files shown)


matklad:
VOPR status before the label removed:

89f9003 #2656	matklad	vopr_lite	./zig/zig build -Drelease vopr -- --lite 8331395237796188405	1m 1s	42m 55s ago	17883

And I have another 5k runs locally

matklad:
Ok, it failed:

```
./zig/zig build -Drelease vopr -- --lite 15323227265840554788
```

matklad:
Ok, this is puzzling, the failure was actually a false positive! I think I perhaps ended up in a semantic conflict with #2680, rebased & restarted the fuzzers

sentientwaffle:
CFO found an actual failure: `./zig/zig build -Drelease vopr -- --lite 9260296628271813392`.

src/vsr/replica.zig
```diff
@@ -1312,6 +1313,10 @@ pub fn ReplicaType(
             }
         }
 
+        pub fn invariants(self: *const Replica) void {
```
sentientwaffle, line 1316: I think we typically call this sort of  function `verify`, but `invariants` is a good name too.

src/vsr/replica.zig
```diff
@@ -1635,6 +1642,18 @@ pub fn ReplicaType(
             assert(message.header.replica < self.replica_count);
             assert(message.header.operation != .reserved);
 
+            // Sanity check --- if the prepare is definitely from the current log-wrap, it should be
+            // appended.
+            defer {
+                if (self.status == .normal and self.syncing == .idle and
+                    message.header.view == self.view and
+                    message.header.op > self.op_checkpoint() and
+                    message.header.op <= self.op_checkpoint_next_trigger())
```
sentientwaffle, line 1651: Could this be increased to
```suggestion
                    message.header.op <= self.op_prepare_max())
```
instead?

sentientwaffle (approved):

src/vsr/replica.zig
```diff
@@ -1312,6 +1313,10 @@ pub fn ReplicaType(
             }
         }
 
+        pub fn invariants(self: *const Replica) void {
+            assert(self.journal.header_with_op(self.op) != null);
```
sentientwaffle, line 1317: Other ideas to verify:
- `self.log_view <= self.view`
- `log_view`/`view` are at least the superblock `vsr_state`'s views.
- maybe `op >= op_checkpoint` and `op <= op_prepare_max`?

src/vsr/replica.zig
matklad: I thought about that, but figured that by checking the trigger we are also indirectly checking that op_prepare_max itself is correct! It's a weaker invariant, but a separate code path.

src/vsr/replica.zig
matklad: Yeah, there's a whole bunch of stuff I want to check here, but, the more conditions I add, the harder is to debug vopr failures, so I decided to start small!

matklad: (what I _actually_ want to check specifically is that what Joran was talking about a while ago: that, if we have a prepare locally, we must be committing!)

PR #2658, opened 2025-01-20
release: 2025-01-20 (attempt 2)

sentientwaffle:
Due to CI flake: https://github.com/tigerbeetle/tigerbeetle/actions/runs/12871534132/job/35885017647


chaitanyabhandari (approved):

matklad:
@sentientwaffle for the future, I think we should _just_ bump the number in the changeling, without adding `(unreleased)` header! Or, if we _do_ add `(unreleased)`, we should adjust the docs to say so: https://github.com/tigerbeetle/tigerbeetle/blob/main/docs/about/internals/releases.md#error-handling

PR #2659, opened 2025-01-21
aof: reduce binary size

cb22:
In a surprising turn of events, it turns out the second big spike here:
image

was due to https://github.com/tigerbeetle/tigerbeetle/pull/2615! (the first was due to debug logging)

It looks like when using the struct initialization syntax

```zig
        self.* = AOFEntry{
            .message = undefined,
        };
```

Zig is embedding ~1MB of zeros in the binary itself, to later `memcpy` over. It seems this is somewhat optimization dependent, in the below minimal reproduction (in contrast to the real issue), it only happens in `Debug` but not `ReleaseSafe` or `ReleaseFast`. 

(The real `self.*` bloat doesn't go away even when coupled with `@setRuntimeSafety(false)`.)

<details>

```zig
const std = @import("std");

const Example = struct {
    message: [1024*1024]u8,
};

export fn tester() [*]u8 {
    var foo = Example { .message = undefined };
    const bar = std.mem.asBytes(&foo);

    return bar.ptr;
}
```
</details>

### DIFF
--- a/src/aof.zig
+++ b/src/aof.zig
@@ -79,10 +79,18 @@ pub const AOFEntry = extern struct {
         // object storage, this must be embedded in the filename or path.
         // Whether this replica is the primary can be determined by the view number from the
         // relevant op.
-        self.* = AOFEntry{
-            .message = undefined,
-        };
+        comptime {
+            const fields = std.meta.fieldNames(AOFEntry);
+            assert(fields.len == 2);
+            assert(std.mem.eql(u8, fields[0], "magic_number"));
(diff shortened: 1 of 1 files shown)


matklad:
Let’s file an issue against the Zig repository as well? Perhaps it’s simpler to fix upstream, and maybe we can get this into 0.14?

matklad (approved):

cb22:
Looks very similar to https://github.com/ziglang/zig/issues/21988, even though this isn't global constants.

PR #2661, opened 2025-01-21
CFO: Split budget/timeout/reload intervals

sentientwaffle:
## Bug

Previously, we only started new fuzzers during the first `budget_minutes` (10 minutes) of each CFO run.

At the end of that 10 minutes, we wait for up to another `hang_minutes` (30 minutes) to wait for any long running fuzzers to finish. But during this time we don't start any new fuzzers.

This that CFO cores are idle for up to 30 of every 40 minutes. This is also the rate at which the CFO pushes seeds to the devhub and refreshes its tasks.

## Fix

To fix:

- Remove `hang_minutes`.
- `budget_minutes` is now the total fuzz loop duration.
  - That is, it determines how often the CFO's code is refreshed.
  - (Approximately -- it doesn't count compile + upload times.)
  - This is now set to 40 minutes, to match our "real" budget from before. (I want it to be long enough to have a usefully-high `timeout_minutes`.)
- Upload seeds to the devhubdb and refresh our tasks every `refresh_minutes`. This is now set to 5 minutes.
- If a fuzzer runs for longer than `timeout_minutes`, it is killed and counts as a failure.

With this fix, CFO will use all cores all the time, and it will push updates every 5 minutes.
So when you add a fuzz label to a PR, the maximum-time-till-devhub-seeds is now 10 minutes instead of 80 minutes.

---

Unrelated to the above fix -- https://github.com/tigerbeetle/tigerbeetle/pull/2661/commits/020eaa7455c372dbafaee62cc85e9cb86a68bbbc should give us a better idea of how long the various fuzzers can take to run.

### DIFF
--- a/src/scripts/cfo.zig
+++ b/src/scripts/cfo.zig
@@ -23,6 +23,8 @@
 //! `args.timeout_minutes` is terminated and recorded as a failure. At the end of the fuzzing loop,
 //! any fuzzers that are still running are cancelled. Cancelled seeds are not recorded.
 //!
+//! Note that the budget/refresh timers do not count time spent cloning or compiling code.
+//!
 //! It is important that the caller (systemd typically) arranges for CFO to be a process group
 //! leader. It is not possible to reliably wait for (grand) children with POSIX, so its on the
 //! call-site to cleanup any run-away subprocesses. See `./cfo_supervisor.sh` for one way to
@@ -60,7 +62,7 @@ const Shell = @import("../shell.zig");
 
 pub const CLIArgs = struct {
     /// How long to run the cfo before exiting (so that cfo_supervisor can refresh our code).
-    budget_minutes: u64 = 40,
+    budget_minutes: u64 = 60,
     /// The interval for flushing accumulated seeds to the devhub.
     /// In addition to this interval, any remaining seeds will be uploaded at the end of the budget.
     refresh_minutes: u64 = 5,
@@ -142,6 +144,9 @@ pub fn main(shell: *Shell, gpa: std.mem.Allocator, cli_args: CLIArgs) !void {
     } else {
         try shell.exec("gh --version", .{});
     }
+
+    // Only garbage-collect the working directory once per run, so that we can reuse the Zig cache.
+    shell.project_root.deleteTree("working") catch {};
 
     try run_fuzzers(shell, gpa, gh_token_option, .{
         .concurrency = cli_args.concurrency orelse try std.Thread.getCpuCount(),
@@ -368,10 +373,6 @@ fn run_fuzzers_prepare_tasks(shell: *Shell, gh_token: ?[]const u8) !Tasks {
     var working_directory = std.ArrayList([]const u8).init(shell.arena.allocator());
     var seed_record = std.ArrayList(SeedRecord).init(shell.arena.allocator());
 
-    // Fuzz an independent clone of the repository, so that CFO and the fuzzer could be on
-    // different branches (to fuzz PRs and releases).
-    shell.project_root.deleteTree("working") catch {};
-
     { // Main branch fuzzing.
         const commit = if (gh_token == null)
             // Fuzz in-place when no token is specified, as a convenient shortcut for local
@@ -382,6 +383,8 @@ fn run_fuzzers_prepare_tasks(shell: *Shell, gh_token: ?[]const u8) !Tasks {
             try shell.pushd("./working/main");
             defer shell.popd();
 
+            // Fuzz an independent clone of the repository, so that CFO and the fuzzer could be on
+            // different branches (to fuzz PRs and releases).
             break :commit try run_fuzzers_prepare_repository(shell, .main_branch);
         };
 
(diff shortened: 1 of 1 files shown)


src/scripts/cfo.zig
```diff
@@ -181,76 +172,96 @@ fn run_fuzzers(
...
+    for (0..options.budget_seconds) |second| {
+        const iteration_last = second == options.budget_seconds - 1;
+        const iteration_pull = second % options.refresh_seconds == 0;
+        const iteration_push = (second % options.refresh_seconds == 0 and second > 0) or
+            iteration_last;
+
+        // Note that tasks are allocated by the arena, so they accumulate over the lifetime of CFO.
+        if (iteration_pull) tasks_cache = null;
+        const tasks = tasks_cache orelse tasks: {
+            const tasks = try run_fuzzers_prepare_tasks(shell, gh_token);
```
matklad, line 213: This is formally a race: `prepare_tasks` removes the `working` directory, which might still be `stat`ed by `zig build fuzz`, which I think can lead to spurious errors. Though, there's at least a second between us starting `build fuzz` and us deleting the directory, so this should be fine. 

I think the proper solution here would be to run the fuzz binary directly. I want to look into that soon, as it's also required for oom handling.

src/scripts/cfo.zig
```diff
@@ -293,19 +314,35 @@ fn run_fuzzers(
                 }
             }
         }
+        assert(running_count == options.concurrency);
 
-        if (second < options.budget_seconds) {
-            assert(running_count == options.concurrency);
+        if (iteration_push) {
+            if (options.devhub_token) |token| {
+                try upload_results(shell, gpa, token, seeds.items);
```
matklad, line 348: I was wondering whether we need to cache the devhub repo dir, now that we do many uploads, but that seems unnecessary, it is pretty snappy anyway.

src/scripts/cfo.zig
matklad: Could you also log `shell.arena.queryCapacity()` somewhere (either next to `fuzzing` or at the end of the cfo)? I am mildly concerned that we now add more loops around arenas. I think it _should_ be fine to not free all this memory, but would be good to keep it in check!

src/scripts/cfo.zig
```diff
@@ -54,8 +59,13 @@ const stdx = @import("../stdx.zig");
...
 pub const CLIArgs = struct {
-    budget_minutes: u64 = 10,
-    hang_minutes: u64 = 30,
+    /// How long to run the cfo before exiting (so that cfo_supervisor can refresh our code).
+    budget_minutes: u64 = 40,
+    /// The interval for flushing accumulated seeds to the devhub.
+    /// In addition to this interval, any remaining seeds will be uploaded at the end of the budget.
+    refresh_minutes: u64 = 5,
+    /// A fuzzer which takes longer than this timeout is killed and counts as a failure.
+    timeout_minutes: u64 = 30,
```
matklad, line 70: check that budget_minutes >= timeout_minutes. 

refresh_minutes I think we don't need to check, because we always have upload at the end. Though perhaps assert that it is at least one minute?

matklad (commented):
LGTM, only nits!

I am beginning to wonder if we need to statically allocate CFO?

src/scripts/cfo.zig
sentientwaffle: :+1: Good idea. When `budget_minutes < timeout_minutes` I will just print a warning, though -- for e.g. testing you almost always want to run with a small budget, but that shouldn't imply a seed taking >1m is a failed seed.

src/scripts/cfo.zig
sentientwaffle: > This is formally a race: prepare_tasks removes the working directory, which might still be stated by zig build fuzz, which I think can lead to spurious errors. Though, there's at least a second between us starting build fuzz and us deleting the directory, so this should be fine.

Good catch! I just added a TODO for now, but I might fix that in a follow up.

sentientwaffle:
> I am beginning to wonder if we need to statically allocate CFO?

Tricky due to the json parsing. We could certainly rely less on the arena, though.

sentientwaffle:
(Disabling auto-merge since I think there's a bug.)

sentientwaffle:
The last commit keeps us from spending so much time recompiling fuzzers.

We will still recompile every `budget_minutes` (rather than every `refresh_minutes`) -- which is now increased to 60.

We could avoid recompiling even then -- just don't delete the whole `working` tree. But the downsides are that it is less "clean" (reusing state between CFO runs) and also we still need to garbage-collect old PRs.

matklad:
Hm, this makes me think: do we give a separate .zig-cache for each PR? Would we get a compile time improvement if we set ZIG_LOCAL_CACHE_DIR env var to working/zig-cache?

sentientwaffle:
> Would we get a compile time improvement if we set ZIG_LOCAL_CACHE_DIR env var to working/zig-cache?

I'm not really clear on what zig caches globally (which would already be shared) vs per-project (which is not shared).

I will experiment tomorrow to see if using a shared `ZIG_LOCAL_CACHE_DIR` improves things! But I can post that as a follow-up PR if this has already merged by then.

matklad (approved):

sentientwaffle:
I tested this out locally -- unfortunately, using a shared `ZIG_LOCAL_CACHE_DIR` doesn't improve compile times.

PR #2662, opened 2025-01-22
vsr: rollout new CheckpointState format, stage 0

chaitanyabhandari:
This is a prerequisite for https://github.com/tigerbeetle/tigerbeetle/pull/2600, which ensures blocks released during a checkpoint are freed only when the next checkpoint is durable, solving a long-standing liveness issue. Since this involves changing the on-disk format of `CheckpointState`, which is sent as part of the `StartView` message, we need to safely upgrade the VSR protocol (preferably without bumping up the VSR protocol version, which would require more intrusive changes).

Inspired from https://github.com/tigerbeetle/tigerbeetle/pull/2181, this PR adds a no-op `start_view_new` command, which is ignored by replicas running on this version. Then, https://github.com/tigerbeetle/tigerbeetle/pull/2600 will take up the change to send _both_ `start_view` and `start_view_new` messages, so replicas running on a version that is aware of only the old CheckpointState can process a StartView message sent by a replica that is aware of the new CheckpointState  format. Finally, in the subsequent release, we can remove the `start_view` message and rename `start_view_new` → `start_view`. 

Multi-version binary upgrades ensure all releases bundled within a binary are visited while upgrading to the highest version in the binary. Additionally, the upgrade quorum is equal to replica_count. Therefore, even clusters that are lagging behind multiple releases will still upgrade correctly.

### DIFF
--- a/src/vsr/message_header.zig
+++ b/src/vsr/message_header.zig
@@ -89,6 +89,7 @@ pub const Header = extern struct {
             .start_view_change => StartViewChange,
             .do_view_change => DoViewChange,
             .start_view => StartView,
+            .start_view_new => StartView,
             .request_start_view => RequestStartView,
             .request_headers => RequestHeaders,
             .request_prepare => RequestPrepare,
@@ -210,7 +211,8 @@ pub const Header = extern struct {
             .reply => return .unknown,
             // These messages identify the peer as either a replica or a client:
             .ping_client => |ping| return .{ .client = ping.client },
-
+            // At the current version, start_view_new is completely ignored.
+            .start_view_new => return .unknown,
             // All other messages identify the peer as a replica:
             .ping,
             .pong,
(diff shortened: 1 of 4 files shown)


src/vsr/message_header.zig
```diff
@@ -210,7 +211,8 @@ pub const Header = extern struct {
             .reply => return .unknown,
             // These messages identify the peer as either a replica or a client:
             .ping_client => |ping| return .{ .client = ping.client },
-
+            // At the current version, is start_view_new is completely ignored.
```
sentientwaffle:
```suggestion
            // At the current version, start_view_new is completely ignored.
```

chaitanyabhandari: Thank you; fixed in [f8b222a](https://github.com/tigerbeetle/tigerbeetle/pull/2662/commits/f8b222a9b3ad965ec4d5ac08272f2a8c603a3987)!

sentientwaffle (approved):

PR #2663, opened 2025-01-24
docs: fix various issues

fabioarnold:
* About menu didn't automatically expand when clicking on it
* Clicking and accepting a search result wouldn't sync with the side navigation
* We only need to break anywhere when code appears in headings
* Tweaked the heading sizes on mobile
* Support en-dash and em-dash using Pandoc's smart extension
* You can now toggle the menu using the "M" key
* Fix external links linking to Markdown files
* Add 404 page
* Add missing requests index page
* Style footnotes


src/docs_website/assets/style/style.css
```diff
@@ -708,6 +708,7 @@ article {
 
         code {
           color: var(--gray-12);
+          word-break: break-all;
```
matklad, line 713: Whoa, I missed that we use CSS nesting (actually, I missed that css nesting exists!)! Is it a sufficiently supported feature? I guess we are aiming at mostly tech-literate people with modern browsers, so we should be fine!

src/docs_website/pandoc/markdown-links.lua
```diff
@@ -1,6 +1,8 @@
 is_readme = PANDOC_STATE.input_files[1]:sub(-9) == "README.md"
 
 function Link (link)
+  local is_external = link.target:sub(1, 8) == "https://"
```
matklad, line 4: In our link checker, once we do it, we should also verify that we never use http:// URLs

src/docs_website/src/docs.zig
```diff
@@ -367,7 +384,7 @@ const Page = struct {
     ) !void {
         const pandoc_step = std.Build.Step.Run.create(b, "run pandoc");
         pandoc_step.addFileArg(website.pandoc_bin);
-        pandoc_step.addArgs(&.{ "--from", "gfm", "--to", "html5" });
+        pandoc_step.addArgs(&.{ "--from", "gfm+smart", "--to", "html5" });
```
matklad, line 387: Please double-check that `+smart` doesn't cause any surprise rendering issues for us, by diffing HTML before and after.

src/docs_website/src/html/404.html
```diff
@@ -0,0 +1,225 @@
...
+      return hash32(seed) / 0xffffffff;
+    }
+
+    function isDarkmode() {
+      return window.matchMedia && window.matchMedia('(prefers-color-scheme: dark)').matches;
+    }
+  </script>
+</body>
+
+</html>
```
matklad: Add a newline, and, ideally, an automated check that we always have a newline for .html (docs.add sounds like a good place for the check?)

src/docs_website/src/html/page-script.js
```diff
@@ -78,14 +76,13 @@ function syncSideNavWithLocation() {
...
           }
         }
       }
     });
   }
-}
\ No newline at end of file
+}
+
+syncSideNavWithLocation();
```
matklad: ditto here. All our source-files should end in `\n`, for consistency and diff minimization.

src/docs_website/assets/style/style.css
fabioarnold: Works in all major browsers since Dec '23. https://caniuse.com/css-nesting

src/docs_website/src/html/404.html
fabioarnold: `docs` in this case is a `Build.Step.WriteFile`. So not easily extendable.

Instead I've added a newline check to `file_checker.zig`.

src/docs_website/src/docs.zig
fabioarnold: OK, I manually went through all of our output files and I think we're good. It's fancier quotes, ellipsis, and dashes. No surprises!

Here's a sample:
image

src/docs_website/src/file_checker.zig
```diff
@@ -37,15 +37,27 @@ fn check_file_sizes(arena: std.mem.Allocator, path: []const u8) !void {
...
             log.err("file '{s}' has unsupported type '{s}'", .{
                 try dir.realpathAlloc(arena, entry.path),
                 std.fs.path.extension(entry.path),
             });
             return error.UnsupportedFileType;
+        };
+
+        if (!file_type.is_binary) {
+            const file = try dir.openFile(entry.path, .{});
+            defer file.close();
```
matklad, line 52: We add a blank after defer, to make it stand out more

matklad (approved):

PR #2664, opened 2025-01-24
clients: switch to using gpa as backing allocator

cb22:
Currently, libtbclient (via `tb_client_exports.zig`) uses `std.heap.c_allocator` directly. Rather, wrap it inside a GeneralPurposeAllocator, so we can get access to all of the nice safety features and leak detection it provides.

### DIFF
--- a/src/clients/c/tb_client/context.zig
+++ b/src/clients/c/tb_client/context.zig
@@ -148,6 +148,9 @@ pub fn ContextType(
 ) type {
     return struct {
         const Context = @This();
+        const GPA = std.heap.GeneralPurposeAllocator(.{
+            .thread_safe = true,
+        });
 
         const StateMachine = Client.StateMachine;
         const allowed_operations = [_]StateMachine.Operation{
@@ -180,6 +183,7 @@ pub fn ContextType(
             InvalidDataSize,
         };
 
+        gpa: GPA,
         allocator: std.mem.Allocator,
         client_id: u128,
         cluster_id: u128,
@@ -203,17 +207,37 @@ pub fn ContextType(
         thread: std.Thread,
 
         pub fn init(
-            allocator: std.mem.Allocator,
+            root_allocator: std.mem.Allocator,
             client_out: *ClientInterface,
             cluster_id: u128,
             addresses: []const u8,
             completion_ctx: usize,
             completion_callback: CompletionCallback,
         ) InitError!void {
-            var context = try allocator.create(Context);
-            errdefer allocator.destroy(context);
+            var context: *Context = context: {
+                // Wrap the root allocator - usually heap.c_allocator when built as a library - in
+                // a GPA to keep maximum compatibility while gaining the extra safety. As a library,
+                // libtbclient is running inside another process's address space.
+                var gpa = GPA{
+                    .backing_allocator = root_allocator,
+                };
+                errdefer assert(gpa.deinit() == .ok);
 
-            context.allocator = allocator;
+                const context = try gpa.allocator().create(Context);
+
+                // Moving the GPA is safe, since we don't have any live reference to `allocator`.
+                context.gpa = gpa;
+
+                break :context context;
+            };
+
+            errdefer {
+                var gpa: GPA = context.gpa;
+                gpa.allocator().destroy(context);
+                assert(gpa.deinit() == .ok);
+            }
+
+            const allocator = context.gpa.allocator();
             context.client_id = stdx.unique_u128();
             context.cluster_id = cluster_id;
             context.addresses_copy = try allocator.dupe(u8, addresses);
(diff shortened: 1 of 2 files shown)


matklad:
>Rather, wrap it inside a GeneralPurposeAllocator, so we can get access to all of the nice safety features and leak detection it provides.

We should be calling .deinit to do this though, no? 

At some point, we should take `std.mem.Allocator` as a parameter of `tb_client.init`, such that the caller can pass us their own allocator. This would allow us to get rid of libc. But this is definitely future work

src/tb_client_exports.zig
```diff
@@ -11,6 +11,12 @@ pub const std_options = .{
     .logFn = tb.Logging.application_logger,
 };
 
+/// Static backing allocator for clients. Wraps heap.c_allocator to keep maximum compatibility since
+/// as a library, libtbclient is running inside another process's address space.
+var gpa = std.heap.GeneralPurposeAllocator(.{}){
+    .backing_allocator = std.heap.c_allocator,
+};
+
```
matklad: Maybe also `comptime assert(@sizeOf(gpa.mutex) > 0)` to double-check that it is thread-safe?

batiati:
> We should be calling .deinit to do this though, no?

Yes, I think it's fine to use `c_allocator` when calling `tb.init()` from the FFI exports. However, we should wrap it into a GeneralPurposeAllocator inside `context.zig`, so we can check for leaks during `tb.deinit()`.

https://github.com/tigerbeetle/tigerbeetle/blob/f81ff2009a4e504c085b631417668e8093e9696c/src/clients/c/tb_client/context.zig#L126

src/tb_client_exports.zig
cb22: I've gone with creating the GPA type with `.thread_safe = true`.

cb22:
>  We should be calling .deinit to do this though, no?

> However, we should wrap it into a GeneralPurposeAllocator inside `context.zig`, so we can check for leaks during `tb.deinit()`.

Yep! I blame the previous not working static hack on post-flight tiredness :)

Gone with exactly this approach now; the GPA is owned by the context (which is still allocated with the root allocator) and it allocates / deallocates and deinits now.

src/clients/c/tb_client/context.zig
```diff
@@ -114,16 +118,24 @@ pub fn ContextType(
...
+            errdefer root_allocator.destroy(context);
+
+            // Wrap the root allocator - presumed to be heap.c_allocator - in a GPA to keep maximum
+            // compatibility while gaining the extra safety. As a library, libtbclient is running
+            // inside another process's address space.
+            context.gpa = GPA{
+                .backing_allocator = root_allocator,
+            };
+            const allocator = context.gpa.allocator();
+            errdefer assert(context.gpa.deinit() == .ok);
```
batiati: We can safely _move_ the GPA since we don't keep live references to `gpa.allocator()`.
So we don't need to allocate `context` differently from `root_allocator`.

```suggestion
            const context: *Context = context: {
                // Wrap the root allocator - presumed to be heap.c_allocator - in a GPA to keep maximum
                // compatibility while gaining the extra safety. As a library, libtbclient is running
                // inside another process's address space.
                var gpa = GPA{
                    .backing_allocator = root_allocator,
                };
                errdefer assert(gpa.deinit() == .ok);

                const context = try gpa.allocator().create(Context);
                // Moving the GPA is safe, since we don't have any live reference to `allocator`.
                context.gpa = gpa;
                break :context context;
            };
            errdefer {
                var gpa: GPA = context.gpa;
                gpa.allocator().destroy(context);
            }
            const allocator = context.gpa.allocator();
```

src/clients/c/tb_client/context.zig
```diff
@@ -237,17 +249,18 @@ pub fn ContextType(
...
             self.signal.deinit();
-            self.client.deinit(self.allocator);
-            self.message_pool.deinit(self.allocator);
+            self.client.deinit(self.gpa.allocator());
+            self.message_pool.deinit(self.gpa.allocator());
             self.io.deinit();
 
-            self.allocator.destroy(self);
+            assert(self.gpa.deinit() == .ok);
+            root_allocator.destroy(self);
```
batiati: We also don't need to pass `root_allocator` to `deinit`, now we can use the GPA for dealocating the context (and assert it).

```suggestion
            var gpa: GPA = self.gpa;
            const allocator = gpa.allocator();
            allocator.destroy(self);
            assert(gpa.deinit() == .ok);
```

src/clients/c/tb_client/context.zig
cb22: Ah yes, wrapping it inside a block is clever! I had initially done it by moving the allocator externally, and the code wasn't easy to follow.

cb22:
@batiati rebased and refactored this on top of main!

src/clients/c/tb_client/context.zig
```diff
@@ -163,6 +166,7 @@ pub fn ContextType(
             InvalidDataSize,
         };
 
+        gpa: GPA,
```
batiati, line 186: Do we need both `gpa` and `allocator`?
I think we can store just the GPA reference here and use `self.gpa.allocator()`.

batiati (dismissed):

allysa08 (approved):

cb22:
@batiati had to rebase due to a conflict once #2665 merged.

batiati (approved):

PR #2665, opened 2025-01-24
clients: add tb_client_init_parameters and implement for python

cb22:
This would have caught #2652 - it adds a simple round trip method, tb_client_init_parameters, that allows retrieving the cluster_id and addresses passed in when creating a cluster.

They are stored on the client context, so it works with both the real and echo clients. There's a little bit of overhead doing it this way, but I think it's the most straighforward.

Addresses in particular is duplicated as a string - IPv6 addresses (or even v4) might canonicalize and thus differ from the value passed in if the net.Address were formatted instead.

It also adds a test for Python - which should be extended to all languages.

### DIFF
--- a/src/clients/c/tb_client/context.zig
+++ b/src/clients/c/tb_client/context.zig
@@ -23,6 +23,13 @@ const Message = MessagePool.Message;
 const Packet = @import("packet.zig").Packet;
 const Signal = @import("signal.zig").Signal;
 
+pub const InitParameters = extern struct {
+    cluster_id: u128,
+    client_id: u128,
+    addresses_ptr: [*]const u8,
+    addresses_len: u64,
+};
+
 /// Thread-safe client interface allocated by the user.
 /// Contains the `VTable` with function pointers to the StateMachine-specific implementation
 /// and the synchronization status.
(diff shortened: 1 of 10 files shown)


batiati:
Interesting! What do you think about adding the `client_id` we generate as well? It can be helpful when cross-comparing application and server logs, as we use it as the client session.

cb22:
Good idea, done!

src/clients/c/tb_client/context.zig
```diff
@@ -194,11 +198,18 @@ pub fn ContextType(
...
+
             context.completion_fn = completion_fn;
             context.implementation = .{
                 .completion_ctx = completion_ctx,
                 .submit_fn = Context.on_submit,
                 .deinit_fn = Context.on_deinit,
+
+                .cluster_id = cluster_id,
+                .client_id = context.client_id,
+                .addresses = addresses_copy,
```
batiati: The best way to handle this, would be adding a function pointer to the `ContextImplementation` interface to retrieve the fields, instead of declaring the fields inside the interface.

I wouldn't bother with this, but [I have plans](https://github.com/tigerbeetle/tigerbeetle/pull/2614#discussion_r1900454115) to move the `ContextImplementation` to intrusive memory, so it's very important to make that struct very simple (just a pointer + v_table).

```diff
pub const ContextImplementation = struct {
    completion_ctx: usize,
    submit_fn: *const fn (*ContextImplementation, *Packet) void,
    deinit_fn: *const fn (*ContextImplementation) void,
+   init_parameters_fn: *const fn(*ContextImplementation, out ...) void,
};
```

batiati: **Note:** PR [#2742](https://github.com/tigerbeetle/tigerbeetle/pull/2742) significantly changes this implementation, using VTables.

cb22:
@batiati rebased and refactored this on top of main!

batiati (commented):
LGTM, please add `tb_client_init_parameters` to `tb_client_header.zig` and update the `tb_client.h` file.

batiati (approved):

PR #2666, opened 2025-01-24
VSR/Journal: `writing() → enum{none,slot,exact}`

sentientwaffle:
## Bug

`Journal.writing()`'s implementation made the (incorrect) assumption that `journal.writes` were ordered chronologically. (This didn't manifest in the VOPR since incorrect results just lead to retries.)

Also, the previous version allowed hard-to-reason-about sequences of writes to the same slot.

## Fix

Since in practice contention of WAL writes at the same slot is rare (requiring either a view-change or a very weird checkpoint + WAL-repair scenario), just disallow it to simplify journal invariants.

`writing(op, checksum)` now returns either:
- `exact`: A message with the given `op`/`checksum` is being written.
- `slot`: A message with the given `op`/`checksum` is not being written, but a message _is_ being written to the same slot.
- `none`: None of the above.

(I'm not 100% happy with this solution, but the other things I tried were even more complicated.)

## Also

Rewrite `Journal.writing()` to take a header instead of `op`/`checksum`.

### DIFF
--- a/src/vsr/journal.zig
+++ b/src/vsr/journal.zig
@@ -1781,7 +1781,7 @@ pub fn JournalType(comptime Replica: type, comptime Storage: type) type {
             assert(message.header.size >= @sizeOf(Header));
             assert(message.header.size <= message.buffer.len);
             assert(journal.has_header(message.header));
-            assert(!journal.writing(message.header.op, message.header.checksum));
+            assert(journal.writing(message.header) == .none);
             if (replica.solo()) assert(journal.writes.executing() == 0);
 
             // The underlying header memory must be owned by the buffer and not by journal.headers:
@@ -1840,6 +1840,7 @@ pub fn JournalType(comptime Replica: type, comptime Storage: type) type {
             const journal = write.journal;
             const message = write.message;
             assert(journal.status == .recovered);
+            assert(journal.writing(message.header) == .exact);
 
             // `prepare_inhabited[slot.index]` is usually false here, but may be true if two
             // (or more) writes to the same slot were queued concurrently and this is not the
@@ -1952,6 +1953,8 @@ pub fn JournalType(comptime Replica: type, comptime Storage: type) type {
             // This allows us to enforce journal.writes.len≤1 when replica_count=1, because the
             // callback may immediately start the next write.
             journal.writes.release(write);
+            assert(journal.writing(write_message.header) == .none);
+
             write_callback(replica, wrote, write_trigger);
             replica.message_bus.unref(write_message);
         }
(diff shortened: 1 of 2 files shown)


src/vsr/replica.zig
```diff
@@ -9936,7 +9940,7 @@ pub fn ReplicaType(
                 return;
             }
 
-            if (self.journal.writing(message.header.op, message.header.checksum)) {
+            if (self.journal.writing(message.header) != .none) {
```
matklad: I'd `switch` here, to make the logic more obvious at the call-site. Maybe even adjust the log message to be different?

src/vsr/journal.zig
```diff
@@ -2142,25 +2145,34 @@ pub fn JournalType(comptime Replica: type, comptime Storage: type) type {
...
+                {
+                    assert(write_slot.index == slot.index);
+                    assert(found == .none);
+                    found = .exact;
+                } else {
+                    if (write_slot.index == slot.index) {
+                        assert(found == .none);
+                        found = .slot;
+                    }
                 }
```
matklad: Took me a moment to convince myself that the nested ifs don't miss a different prepare in the same slot. Perhaps we can flatten this?

```zig
                if (write_slot.index == slot.index) {
                    assert(found == .none);
                    found = .slot;
                }
                if (write.message.header.op == header.op and
                    write.message.header.checksum == header.checksum)
                {
                    assert(write_slot.index == slot.index);
                    assert(found == .slot);
                    found = .exact;
                }
```

Not entirely sure if that's better though!

src/vsr/journal.zig
```diff
@@ -2142,25 +2145,34 @@ pub fn JournalType(comptime Replica: type, comptime Storage: type) type {
...
-                    // write to the same slot, writing() must return false.
-                    found = false;
+        pub fn writing(journal: *Journal, header: *const Header.Prepare) Writing {
+            const slot = journal.slot_for_header(header);
+            var found: Writing = .none;
+            var writes = journal.writes.iterate();
+            while (writes.next()) |write| {
+                const write_slot = journal.slot_for_op(write.message.header.op);
+                if (write.message.header.op == header.op and
+                    write.message.header.checksum == header.checksum)
```
matklad: Might want to keep the comment explaining why we compare both op & checksum

matklad: Though, I am not 100% convinced that the reasoning is sound. We might as well branch after comparing the first 16 bytes of checksum! So another option is to compare just the sums, and assert op equality.

matklad (commented):
LGTM overal!

- On the `writes` field, add an `// Invariant:` comment explaining that slots are not written concurrently.
- Maybe check the invariant right before we do `storage.write_sectors`, not via calling `.writing`, but by directly comparing `slots`?

One observation: even if we don't write prepare due to `.slot`, we'd still save it in our pipeline queue/cache. As `write_preapre_callback` triggers `repair`, and that uses pipeline for repairs, we'll still effectively have the chaining behavior, see `// The message may be available in the local pipeline.`

Though:

- I am not 100% sure it _works_, our repair was prone to bailing out too early. We _hopefully_ fixed that, but I am still suspicions
- We might extend that comment to also mention `.writing(header) == slot` and `"waiting for IOP"` cases.

src/vsr/journal.zig
sentientwaffle: Replaced with:

```zig
                if (write_slot.index == slot.index) {
                    assert(found == .none);

                    if (write.message.header.checksum == header.checksum) {
                        assert(write.message.header.op == header.op);
                        found = .exact;
                    } else {
                        maybe(write.message.header.op == header.op);
                        found = .slot;
                    }
                } else {
                    assert(write.message.header.op != header.op);
                }
```

which hopefully makes it more clear!

src/vsr/replica.zig
```diff
@@ -9936,13 +9945,15 @@ pub fn ReplicaType(
...
-                    message.header.checksum,
-                });
-                return;
+            switch (self.journal.writing(message.header)) {
+                .none => {},
+                .slot, .exact => |reason| {
+                    log.debug(
+                        "{}: write_prepare: ignoring op={} checksum={} (already writing {})",
+                        .{ self.replica, message.header.op, message.header.checksum, reason },
+                    );
```
```suggestion
                    log.debug(
                        "{}: write_prepare: ignoring op={} checksum={} (already writing {s})",
                        .{ self.replica, message.header.op, message.header.checksum, @tagName(reason) },
                    );
```

I _think_ just printing enum would print `Reason.writing`?

sentientwaffle: Updated. I knew that but didn't want to go over 100 characters... :stuck_out_tongue:

matklad (approved):

PR #2671, opened 2025-01-25
I'm so sorry. I didn't mean to make all of these prs.

pseudogram:
---
[//]: # (BEGIN SAPLING FOOTER)
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/tigerbeetle/tigerbeetle/pull/2671).
* __->__ #2671
* #2669
* #2668
* #2667


matklad:
No worries @pseudogram!

By the way, have you tried ReviewStack on larger pull requests? A persistent problem for us is that, for a PR with thousands of line changed and hundreds of comments, GH review slows to a crawl. If ReviewStack is faster (not to mention stacked PRs), we might switch to it!

I actually need to look at ReviewStack more, I didn't realize it existed, so, actually, thanks for all those PRs, they _are_ helpful!

pseudogram:
TLDR; no.

I’ve been looking for a way to used stacked PRs effectively for a while now. 

graphite.dev looks cool but pricing and lack of on prem hosting rules it out.

Facebook open sourced [sapling](https://sapling-scm.com) a couple years ago. And, I just started testing it out. 

By default every commit becomes a new PR. So in your case with lots of comments, it _could_ speed up because your comments would be distributed across PRs.

PR #2674, opened 2025-01-27
[Go client] Don't pre-allocate the reply buffer.

batiati:
Rather than pre-allocating the reply buffer with the worst-case length, allocate it during the result callback when the actual reply size is known.

This more efficient approach is particularly important for query operations where the result size cannot be inferred from the input length, avoiding the need to allocate an oversized buffer with the maximum message size (1 MiB).

### DIFF
--- a/src/clients/go/tb_client.go
+++ b/src/clients/go/tb_client.go
@@ -236,7 +236,6 @@ func (c *c_client) doRequest(
 		}
 	}
 
-	// Return the amount of bytes written into result
 	return reply, nil
 }
 
@@ -277,7 +276,7 @@ func onGoPacketCompletion(
 			}
 		}
 
-		// Write the result data into the request's result.
+		// Copy the result data into a new buffer.
 		reply = make([]uint8, result_len)
 		C.memcpy(unsafe.Pointer(&reply[0]), unsafe.Pointer(result_ptr), C.size_t(result_len))
 	}
--- a/src/clients/go/tb_client.go
+++ b/src/clients/go/tb_client.go
@@ -61,8 +61,7 @@ type Client interface {
 }
 
 type request struct {
-	result unsafe.Pointer
-	ready  chan struct{}
+	ready chan []uint8
 }
 
 type c_client struct {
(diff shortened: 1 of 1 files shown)


src/clients/go/tb_client.go
```diff
@@ -244,7 +237,7 @@ func (c *c_client) doRequest(
 	}
 
 	// Return the amount of bytes written into result
```
matklad: Update comment.

src/clients/go/tb_client.go
```diff
@@ -286,211 +278,204 @@ func onGoPacketCompletion(
 		}
 
 		// Write the result data into the request's result.
```
matklad: Ditto.

src/clients/go/tb_client.go
```diff
@@ -286,211 +278,204 @@ func onGoPacketCompletion(
...
 	}
 
-	resultCount := wrote / int(unsafe.Sizeof(types.AccountEventResult{}))
-	return results[0:resultCount], nil
+	if reply == nil {
+		return make([]types.AccountEventResult, 0), nil
+	}
+
+	resultsCount := len(reply) / int(unsafe.Sizeof(types.AccountEventResult{}))
+	results := unsafe.Slice((*types.AccountEventResult)(unsafe.Pointer(&reply[0])), resultsCount)
```
matklad, line 312: Hm, is it valid to cast `[]byte` to `[]T` in Go that way? I can't find a definitive answer one way or another, but what gives me a pause is that this package:

https://github.com/alecthomas/unsafeslice/blob/48ea3b1ba1c124a29aa74f9762f984f74c1533a6/unsafeslice.go#L20-L31

does things quite differently (modifying the actual slice header). 

How can we verify that this cast is indeed valid, and won't confuse garbage collector or cause some other nastiness?

matklad: Ahh, apparently it _is_ the correct way to do this, it's just relatively new: https://stackoverflow.com/a/69676107

Ok, yeah, if this passes our tests without panicking runtime, I am reasonably sure it is correct!

matklad (dismissed):
Makes much sense this way to me, thanks!

src/clients/go/tb_client.go
batiati: Yes, the slice conversion is fine. However, I'm not ashamed to admit that I don't fully understand how Go's memory management works, and I'm always uncertain whether the GC will interpret our intent correctly, especially in edge cases like this one, where the buffer is allocated from an exported Go function called from Zig!

AFAIK, the GC will track the memory addresses, regardless of the type.
Like you said, "I am reasonably sure it is correct"!

matklad (approved):

PR #2506, opened 2024-11-25
VSR: Misc misdirected write + client reply fixes

sentientwaffle:
These are all fixes for bugs found by [VOPR misdirected write fault injection](https://github.com/tigerbeetle/tigerbeetle/pull/2463). (I am upstreaming the fixes in advance while still working out liveness issues with the fault injection.)

I recommended reviewing this PR commit-by-commit.

Note that the seeds mentioned in the respective commit messages won't work on this branch -- they require the fault injection of https://github.com/tigerbeetle/tigerbeetle/pull/2463.

The bugs are:
- [Journal: Fix assertion after misdirected read/write](https://github.com/tigerbeetle/tigerbeetle/commit/417a23f2b1304b8f5030ab7d3fb3a6332de4444c)
- [Journal: Don't cut_view_range for redundant headers](https://github.com/tigerbeetle/tigerbeetle/commit/cc4927e4748f872e979f918ef029b9018477aa8b): This is a correctness bug.
- [VSR: Fix ClientReplies write/checkpoint race](https://github.com/tigerbeetle/tigerbeetle/commit/c0ee15983a68baeebff1be48c7d3cc97396ce4f1): Unrelated to misdirected writes -- that same PR adds more corruption to the client replies zone, which revealed this race.

PR #2677, opened 2025-01-27
VOPR: Test misdirected writes

sentientwaffle:
(Supersedes https://github.com/tigerbeetle/tigerbeetle/pull/2463.)

Test misdirected writes in the VOPR.

A _misdirected write_ is a disk fault where:

1. the data is not written to the intended location. (i.e. a "lost write"), and
2. the data _is_ written to a different, mistaken location.

Currently:

- We allow for (at most) one misdirect fault per `Storage` (i.e. per replica) for the time being, for simplicity and because double-faults are not covered by our fault model. This will hopefully match physical disks – misdirected faults are an order of magnitude less frequent than bit rot, which in turn is an order of magnitude less frequent than LSEs.
- In order to keep things interesting:
  - misdirections are always within the same zone,
  - the entire write is misdirected (rather than only some of the sectors), and
  - the misdirected write lands on a convenient offset.
  
  Thanks to rigorous checksums, misdirections that break these rules just manifest as corruptions, and corruption is already well-tested. The goal here is to test how TigerBeetle handles well-formed but incorrectly-located data.

The misdirection is implemented with "overlays" over the `Storage.memory` buffer, so that `Storage.memory` is always pristine. That is, `Storage.memory` is what the disk would look like if no faults had ever occurred.

---

Also in this PR:
- Refactoring `Storage`.
- Enable read/write faults in the `client_replies` zone. Previously we were only injecting faults due to crashes ("torn writes").
- With these new faults, the VOPR found:
  - a couple of minor incorrect asserts in the [journal](https://github.com/tigerbeetle/tigerbeetle/pull/2463/commits/176d716f2bb4fa413ac05b04a0cc43a5087097e0).
  - a [correctness bug](https://github.com/tigerbeetle/tigerbeetle/pull/2463/commits/3f309bf1a9fcaf3eb518ab1b99dcf9b53a6bc562) in journal recovery!
  - a bug in the [client replies](https://github.com/tigerbeetle/tigerbeetle/pull/2463/commits/3e9906731f356336a793cd6b7b0f6df9a89d0919).
  - (Fixes in https://github.com/tigerbeetle/tigerbeetle/pull/2506, https://github.com/tigerbeetle/tigerbeetle/pull/2653 and https://github.com/tigerbeetle/tigerbeetle/pull/2711).

### DIFF
--- a/src/iops.zig
+++ b/src/iops.zig
@@ -25,7 +25,7 @@ pub fn IOPSType(comptime T: type, comptime size: u6) type {
         }
 
         pub fn index(self: *IOPS, item: *T) usize {
-            const i = (@intFromPtr(item) - @intFromPtr(&self.items)) / @sizeOf(T);
+            const i = @divExact((@intFromPtr(item) - @intFromPtr(&self.items)), @sizeOf(T));
             assert(i < size);
             return i;
         }
--- a/src/testing/storage.zig
+++ b/src/testing/storage.zig
@@ -30,6 +30,7 @@ const math = std.math;
 const mem = std.mem;
 
 const FIFOType = @import("../fifo.zig").FIFOType;
+const IOPSType = @import("../iops.zig").IOPSType;
 const constants = @import("../constants.zig");
 const vsr = @import("../vsr.zig");
 const superblock = @import("../vsr/superblock.zig");
@@ -68,6 +69,9 @@ pub const Storage = struct {
         /// Chance out of 100 that a write will corrupt a sector, if the target memory is within
         /// a faulty area of this replica.
         write_fault_probability: u8 = 0,
+        /// Chance out of 100 that a write will misdirect to the wrong sector, if the target memory
+        /// is within a faulty area of this replica.
+        write_misdirect_probability: u8 = 0,
         /// Chance out of 100 that a crash will corrupt a sector of a pending write's target,
         /// if the target memory is within a faulty area of this replica.
         crash_fault_probability: u8 = 0,
@@ -129,6 +133,9 @@ pub const Storage = struct {
 
     pub const NextTickSource = enum { lsm, vsr };
 
+    /// See `Storage.overlays`.
+    const overlays_count_max = 2;
+
     allocator: mem.Allocator,
 
     size: u64,
@@ -141,7 +148,36 @@ pub const Storage = struct {
     /// Set bits correspond to faulty sectors. The underlying sectors of `memory` is left clean.
     faults: std.DynamicBitSetUnmanaged,
 
-    /// Whether to enable faults (when false, this supersedes `faulty_wal_areas`).
+    /// Overlays take precedence over the (pristine) data in `memory`.
+    ///
+    /// Each misdirected write creates two overlays.
+    /// When a misdirected write is triggered:
+    /// - The intended target is overlaid with its old data.
+    /// - The intended target's `memory` is set to the `write.buffer` data.
+    /// - The mistaken target is overlaid with the `write.buffer` data.
+    /// - The mistaken target's `memory` is left untouched.
+    ///
+    /// The reason for all of this is:
+    /// - By keeping `memory` pristine, we can trivially disable both sides of the misdirected-write
+    ///   fault by flipping the `faulty` flag.
+    /// - By tracking the overlays separately, they can be repaired separately.
+    ///
+    /// Other notes:
+    /// - We allow for (at most) one misdirect fault per Storage for the time being, for simplicity
+    ///   and because double-faults are not covered by our fault model. This will hopefully match
+    ///   physical disks – misdirected faults are an order of magnitude less frequent than bit rot,
+    ///   which in turn is an order of magnitude less frequent than LSEs.
+    /// - In order to keep things interesting:
+    ///   - misdirections are always within the same zone,
+    ///   - the entire write is misdirected (rather than only some of the sectors), and
+    ///   - the misdirected write lands on a convenient offset.
+    ///   Thanks to rigorous checksums, misdirections that break these rules just manifest as
+    ///   corruptions, and corruption is already well-tested (see `faults`). The goal here is to
+    ///   test how TigerBeetle handles well-formed but incorrectly-located data.
+    overlays: IOPSType(struct { zone: vsr.Zone, offset: u64, size: u32 }, overlays_count_max) = .{},
+    overlay_buffers: [overlays_count_max][]align(constants.sector_size) u8,
+
+    /// Whether to enable faults (when false, this supersedes `faulty_wal_areas` &c).
     /// This is used to disable faults during the replica's first startup.
     faulty: bool = true,
 
@@ -157,6 +193,11 @@ pub const Storage = struct {
         assert(options.read_latency_mean >= options.read_latency_min);
         assert(options.fault_atlas == null or options.replica_index != null);
 
+        assert(options.read_fault_probability <= 100);
+        assert(options.write_fault_probability <= 100);
+        assert(options.write_misdirect_probability <= 100);
+        assert(options.crash_fault_probability <= 100);
+
         const prng = std.rand.DefaultPrng.init(options.seed);
         const sector_count = @divExact(size, constants.sector_size);
         const memory = try allocator.alignedAlloc(u8, constants.sector_size, size);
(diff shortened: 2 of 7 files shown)


matklad:
How do we want to sequence this with respect to https://github.com/tigerbeetle/tigerbeetle/pull/2708 @chaitanyabhandari ?

sentientwaffle:
Either way is fine with me -- not sure which would be easier overall.

chaitanyabhandari:
Same, no strong preferences there! I don't mind it if this one makes it in first, since DJ has been working on it for a while and he seems to have ironed most of the bugs here. This one has been VOPR green for a while too. 
Screenshot 2025-02-11 at 11 03 11 AM



It would be really interesting to see how boundless misdirected IO will look like. From a very cursory glance, I believe I would have get rid of the below, right @sentientwaffle? Or actually, I will have to think about whether we are guaranteed to be safe in the presence of double faults to the same slot like these at all (for example a double misdirected fault to both the header and the prepare would lead us to *forget* a prepare that we've acked, and there doesn't seem to be any way to detect that)... 🤔 

https://github.com/tigerbeetle/tigerbeetle/blob/fc72954ba4348f5c535e6423cbe93138cc36c3dc/src/testing/storage.zig#L499-L503

sentientwaffle:
Yeah, I'm not sure whether we _can_ get rid of `read_sectors_fault_eligible`, since it protects from correctness bugs via double-faults that are outside of our fault model.

src/iops.zig
```diff
@@ -25,7 +25,7 @@ pub fn IOPSType(comptime T: type, comptime size: u6) type {
         }
 
         pub fn index(self: *IOPS, item: *T) usize {
-            const i = (@intFromPtr(item) - @intFromPtr(&self.items)) / @sizeOf(T);
+            const i = @divExact((@intFromPtr(item) - @intFromPtr(&self.items)), @sizeOf(T));
```
matklad, line 28: ❤️

src/vopr.zig
```diff
@@ -898,6 +901,44 @@ pub const Simulator = struct {
...
+                        replies_latest[entry_slot] = entry.header;
+                    }
+                }
+            }
+        }
+
+        for (replies_latest, 0..) |reply_or_empty, reply_slot| {
+            const reply = reply_or_empty orelse continue;
+            var reply_in_core: bool = false;
+            for (simulator.cluster.replicas) |replica| {
```
matklad: Given that we don't do extra asserts in the loop, we can short circuit with

```rust
const reply_in_core = for () {
    break true;
} else false;
```

to make high-level logic more apparent.

src/testing/storage.zig
```diff
@@ -151,7 +148,36 @@ pub const Storage = struct {
     /// Set bits correspond to faulty sectors. The underlying sectors of `memory` is left clean.
     faults: std.DynamicBitSetUnmanaged,
 
-    /// Whether to enable faults (when false, this supersedes `faulty_wal_areas`).
+    /// Overlays take precedence over the (pristine) data in `memory`.
```
matklad, line 154: Could you add a comment on the `memory` field that it _always_ holds the data written, as if no errors occurred? That's the main idea, and everything else follows from it.

src/testing/storage.zig
```diff
@@ -151,7 +148,36 @@ pub const Storage = struct {
...
+    ///   fault by flipping the `faulty` flag.
+    /// - By tracking the overlays separately, they can be repaired separately.
+    ///
+    /// Other notes:
+    /// - We allow for (at most) one misdirect fault per Storage for the time being, for simplicity
+    ///   and because double-faults are not covered by our fault model. This will hopefully match
+    ///   physical disks – misdirected faults are an order of magnitude less frequent than bit rot,
+    ///   which in turn is an order of magnitude less frequent than LSEs.
+    /// - In order to keep things interesting:
+    ///   - misdirections are always within the same zone,
```
matklad, line 174: Hm, given that _everything_ is a message, I think it might be valuable to mix this up still? To make sure we do check header.command every time we read something from disk?

src/testing/storage.zig
```diff
@@ -178,6 +209,16 @@ pub const Storage = struct {
         var faults = try std.DynamicBitSetUnmanaged.initEmpty(allocator, sector_count);
         errdefer faults.deinit(allocator);
 
+        var overlay_buffers_allocated: u32 = 0;
+        var overlay_buffers: [overlays_count_max][]align(constants.sector_size) u8 = undefined;
+        errdefer for (overlay_buffers) |b| allocator.free(b);
```
matklad: free of a wild pointer! You want to slice by `overlay_buffers_allocated` I think? 

I don't understand how people write Zig _without_ static memory management, it's impossible to not make this kinds of errors consistently otherwise!

src/testing/storage.zig
```diff
@@ -151,7 +148,36 @@ pub const Storage = struct {
...
+    ///   which in turn is an order of magnitude less frequent than LSEs.
+    /// - In order to keep things interesting:
+    ///   - misdirections are always within the same zone,
+    ///   - the entire write is misdirected (rather than only some of the sectors), and
+    ///   - the misdirected write lands on a convenient offset.
+    ///   Thanks to rigorous checksums, misdirections that break these rules just manifest as
+    ///   corruptions, and corruption is already well-tested (see `faults`). The goal here is to
+    ///   test how TigerBeetle handles well-formed but incorrectly-located data.
+    overlays: IOPSType(struct { zone: vsr.Zone, offset: u64, size: u32 }, overlays_count_max) = .{},
+    overlay_buffers: [overlays_count_max][]align(constants.sector_size) u8,
```
matklad: Might plug a `message_size_max` here, then you'll be able to allocate the entire bunch in one go.

src/testing/storage.zig
```diff
@@ -195,12 +236,14 @@ pub const Storage = struct {
...
             .reads = reads,
             .writes = writes,
         };
     }
 
     pub fn deinit(storage: *Storage, allocator: mem.Allocator) void {
+        for (storage.overlay_buffers) |b| allocator.free(b);
         allocator.free(storage.memory);
         storage.memory_written.deinit(allocator);
         storage.faults.deinit(allocator);
```
matklad: Can you sort the entire deinint in reverse-init order, so that's its easier to match visually with init?

src/testing/storage.zig
```diff
@@ -424,9 +471,74 @@ pub const Storage = struct {
...
+
+            {
+                // Don't fault a WAL prepare if the corresponding WAL header write was misdirected,
+                // to avoid a double-fault which the journal interprets as a torn prepare.
+                var overlays_iterator = storage.overlays.iterate_const();
+                while (overlays_iterator.next()) |overlay| {
+                    if (overlay.zone == .wal_headers and overlay.offset == header_offset) {
+                        return .none;
+                    }
+                }
```
matklad: What about scenario where a header write gets misdirected due to a storage fault, and a prepare writes get torn due to a crash/the process being killed. 

That is, where we do have a double-fault, but where one of the faults is a simple tornwrite, and not a hardware error. 

Do we intend to handle these cases?

src/testing/storage.zig
```diff
@@ -461,8 +573,57 @@ pub const Storage = struct {
...
+            const overlay_intended_buffer = storage.overlay_buffers[overlay_intended_index];
+            const target_intended_buffer =
+                storage.memory[write.zone.offset(write.offset)..][0..write.buffer.len];
+
+            stdx.copy_disjoint(.inexact, u8, overlay_mistaken_buffer, write.buffer);
+            stdx.copy_disjoint(.inexact, u8, overlay_intended_buffer, target_intended_buffer);
+        }
+
         const offset_in_storage = write.zone.offset(write.offset);
         stdx.copy_disjoint(
```
matklad: Maybe move this copy to happen right before we return? So that we first handle two-faults, and setup pristine memory last.

matklad: Incidentally, `pristine` might be a better name for `memory` perhaps?

src/testing/storage.zig
matklad: Though, that makes the offsets tricky: now, if you misdirect a prepare over a grid, it can cover _two_ blocks, which breaks our direct offset comparison. But maybe it is fine? That is, it's fine if a misdirected prepare only ever shadows a single block, because the second block will either contain garbage, or misdirected header, and we'll check both cases either way.

sentientwaffle: I will add that as a TODO -- I think it will be easier after https://github.com/tigerbeetle/tigerbeetle/pull/2708 lands.

src/testing/storage.zig
sentientwaffle: I'm not sure what you mean by this

src/testing/storage.zig
sentientwaffle: Added a TODO:

```
                // TODO If in our fault tracking we distinguish between "torn writes" injected by
                // reset() and simulated LSE's/bitrot, then we could allow the former in this case.
```

src/testing/storage.zig
sentientwaffle: I slightly prefer `memory`, mostly because it pairs well with `memory_written`.

src/testing/storage.zig
matklad:
```
overlay_buffers: *align(constants.sector_size) [overlays_count_max][constats.message_size_max] u8,
```

sentientwaffle: Ah, that makes sense, thanks!

matklad (approved):

PR #2679, opened 2025-01-27
Add an initial Rust client

brson:
This is a Rust client for TigerBeetle that binds to the tb_client library written in Zig. In the future we may create a pure-Rust client.

This is not ready for production, but it is ready for early review, and suitable for experimentation and feedback.
The API is intended to be ergonomic, but the current implementation does a lot of allocations per request. Some of these can be eliminated with more or less effort, but perhaps not all can be eliminated. I intend that for future work.

For the most part this library is presenting the underlying tb_client C API as an idiomatic Rust API, without extra transformation and semantics added on top.

The big exception to that is related to point queries that accept multiple events and return multiple responses. In the underlying protocol, these queries can return "empty" responses for efficiency, and callers are required to correlate the responses with the requests. This API does that correlation for the user, so they get back 1:1 results for every event submitted. In the current implementation this requires allocating a new vector to hold these results. This could conceivably be optimized by returning an iterator instead.

These bindings follow a simple and obvious implementation strategy: the API is async, when requests are submitted to tb_client, ownership of allocations is transferred to tb_client; when tb_client calls our completion callaback (on another thread), the rust code immediately takes back ownership and sends all relevant data in a message back to the original rust async event loop thread. The Rust API caller can call multiple async requests concurrently - any resources passed in are not captured in the futures returned (though the `Client` `self` reference is captured to keep it alive for the duration of requests).

For now, I don't intend to publish this to crates.io, so someone who wants to use it must obtain the TigerBeetle source and link to the tigerbeetle crate through a "path" dependency. There are brief instructions in the README.md for this.

How the build integration works: the Rust build, when it detects it is building from the tigerbeetle source, runs `zig/zig build clients:c`, then moves the artifacts it needs into an appropriate place in the rust source directory to link to. When deployed to crates.io we'll have to build in a different mode, not yet implemented. I found no changes were needed to the tigerbeetle `build.zig` script to make this work; as such there is no `zig/zig build clients:rust` command.

This PR only includes cursory documentation. I intend to document as a followup. The README.md is _not_ hooked up the documentation generator, assuming we don't want this project documented on the website yet.

This PR only includes a single smoke test. I intend to write tests as a followup.

I have not attempted to set up CI for this project, though I think we should - please help me understand how to do that.


src/clients/rust/Cargo.toml
```diff
@@ -0,0 +1,18 @@
...
+edition = "2021"
+
+[dependencies]
+bitflags = "2.6.0"
+futures = "0.3.31"
+thiserror = "2.0.3"
+
+[build-dependencies]
+anyhow = "1.0.93"
+bindgen = "0.70.1"
```
matklad: This is a bit unfortunate --- bindgen is a big dependency, and requires quite a bit of compiling on the part of the user! It also arrives at the result in a round-about way: Zig source of truth ->(via tb_client_header.zig) C header -> (via bindgen) Rust code.


What we do for other clients is directly generating code in $LANG from Rust. See how `go_bindings.zig` and `java_bindings.zig` work.

src/clients/rust/Cargo.toml
```diff
@@ -0,0 +1,18 @@
+[package]
+name = "tigerbeetle"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+bitflags = "2.6.0"
+futures = "0.3.31"
+thiserror = "2.0.3"
```
matklad: This is reasonable I guess! The 90% here would be to get rid of `thiserror` and implement the relevant errors manually, just to get rid of syn in transitive deps, but this would be a stretch goal.

src/clients/rust/build.rs
```diff
@@ -0,0 +1,158 @@
...
+
+    for build_target in build_targets {
+        let mut cmd = std::process::Command::new(&zig_compiler);
+        cmd.args(["build", build_target]);
+        let res = cmd.status()?;
+
+        if !res.success() {
+            panic!("zig build failed");
+        }
+    }
```
matklad: For deployment to crates.io, we should package pre-build .so. That's not how Rust usually works, but, given that Zig itself is unstable, we should try to avoid running the Zig compiler on user's machine.

src/clients/rust/tests/tests.rs
```diff
@@ -0,0 +1,302 @@
...
+
+struct TestHarness {
+    tigerbeetle_bin: String,
+    temp_dir: TempDir,
+    port_number: u16,
+    server: Option<Child>,
+}
+
+impl TestHarness {
+    fn new(name: &str) -> AnyResult<TestHarness> {
```
matklad: I usually spell this as `anyhow::Result<TestHarness>`, to avoid aliases, though no strong opinion here.

src/clients/rust/tests/tests.rs
```diff
@@ -0,0 +1,302 @@
...
+        let manifest_dir = env!("CARGO_MANIFEST_DIR");
+
+        let tigerbeetle_bin = if cfg!(unix) {
```
matklad: could use https://doc.rust-lang.org/stable/std/env/consts/constant.EXE_SUFFIX.html to avoid cfgs?

src/clients/rust/tests/tests.rs
```diff
@@ -0,0 +1,302 @@
...
+            format!("{manifest_dir}/../../../tigerbeetle")
+        } else if cfg!(windows) {
+            format!("{manifest_dir}/../../../tigerbeetle.exe")
+        } else {
+            todo!()
+        };
+
+        let work_dir = format!("{manifest_dir}/work");
+        fs::create_dir_all(&work_dir)?;
+        let temp_dir = TempDir::with_prefix_in(name, &work_dir)?;
```
matklad: Could use `CARGO_TARGET_TMPDIR` here:

https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-crates

src/clients/rust/tests/tests.rs
```diff
@@ -0,0 +1,302 @@
...
+
+        let port_number = NEXT_PORT_NUMBER.fetch_add(1, Ordering::SeqCst);
```
matklad: We have a cute trick to avoid this port-searching! If you pass `--addresses=0` to tiger beetle, it'll pick an open port and print it to stdout:

```
λ ./zig/zig build run -Drelease -- start --addresses=0 ./0_0.tigerbeetle
2025-01-29 17:30:40.021Z info(io): opening "0_0.tigerbeetle"...
2025-01-29 17:30:40.038Z info(main): multiversioning: upgrades disabled for development (0.0.1) release.
2025-01-29 17:30:40.038Z info(main): release=0.0.1
2025-01-29 17:30:40.038Z info(main): release_client_min=0.0.1
2025-01-29 17:30:40.038Z info(main): releases_bundled={ 0.0.1 }
2025-01-29 17:30:40.038Z info(main): git_commit=0f3485f09795ca174631d0a3b615de710bb09fc9
2025-01-29 17:30:40.244Z info(replica): superblock release=0.0.1
2025-01-29 17:30:40.600Z info(replica): 0: transition_to_normal_from_recovering_status: view=1 primary
2025-01-29 17:30:40.600Z info(main): 0: Allocated 2977MiB during replica init
2025-01-29 17:30:40.600Z info(main): 0: Grid cache: 1024MiB, LSM-tree manifests: 128MiB
2025-01-29 17:30:40.600Z info(main): 0: cluster=1: listening on 127.0.0.1:52696
2025-01-29 17:30:40.600Z info(main): 0: started with extra verification checks
52696 # <- this is the port!
```

matklad: So, you want to spawn a process, and then read one line of output. 

Additionally, with magic port=0, tigerbeetle will exit automatically when its stdin is closed, which makes is impossible to leak a process in a test even if the tests itself is killed.

src/clients/rust/tests/tests.rs
```diff
@@ -0,0 +1,302 @@
...
+    let code = 1;
+
+    let account_id1 = 1;
+    let account_id2 = 2;
+    let transfer_id1 = 3;
+
+    let account_id2_user_data_32 = 4;
+    let transfer_id1_user_data_32 = 5;
+
+    block_on(async {
```
matklad, line 106: 👍 for just `block_on`ing here, instead of doing `#[tokio::test]` macro, keeps things simple

src/clients/rust/tests/tests.rs
```diff
@@ -0,0 +1,302 @@
...
+        let mut harness = TestHarness::new("smoke")?;
+        harness.prepare_database()?;
+        harness.serve()?;
+
+        let address = &format!("127.0.0.1:{}", harness.port_number);
+        let client = tb::Client::new(0, address)?;
+
+        {
+            let res = client
```
matklad: `let result`, we use fully-named variables! 

(FWIW, I used to be in the `res` camp before TigerBeetle :)

src/clients/rust/tests/tests.rs
```diff
@@ -0,0 +1,302 @@
...
+                    tb::Account {
+                        id: account_id1,
+                        debits_pending: 0,
+                        debits_posted: 0,
+                        credits_pending: 0,
+                        credits_posted: 0,
+                        user_data_128: 0,
+                        user_data_64: 0,
+                        user_data_32: 0,
+                        reserved: tb::Reserved::default(),
```
matklad, line 126: I wonder if we can make it just 

```
reserved: tb::Reserved
```

?

Probably too cute, I think I actually like default more!

src/clients/rust/tests/tests.rs
```diff
@@ -0,0 +1,302 @@
...
+        let mut harness = TestHarness::new("smoke")?;
+        harness.prepare_database()?;
+        harness.serve()?;
+
+        let address = &format!("127.0.0.1:{}", harness.port_number);
+        let client = tb::Client::new(0, address)?;
+
+        {
+            let res = client
+                .submit_create_accounts(&[
```
matklad: Let's remove `submit_` prefix to match other clients! I don't think it brings anything to the table?

src/clients/rust/README.md
```diff
@@ -0,0 +1,51 @@
+---
+title: Rust
+---
```
matklad:
```suggestion
```

We removed this stuff from our docs!

src/clients/rust/src/lib.rs
```diff
@@ -0,0 +1,844 @@
...
+    }
+}
+
+fn create_packet<RustEvent, CEvent>(
+    op: u32, // TB_OPERATION
+    events: &[RustEvent],
+) -> (Box<tbc::tb_packet_t>, Receiver<CompletionMessage<CEvent>>)
+where
+    RustEvent: Copy,
+    CEvent: From<RustEvent> + 'static,
```
matklad: I thinks for rust specifically, we should avoid any conversions here, and make sure that this crate's `pub` types have ABI compatible with Zig. 

It is fine to expose ABI over crate boundary, as our ABI is stable.

src/clients/rust/src/lib.rs
```diff
@@ -0,0 +1,844 @@
...
+    }
+
+    pub fn close(mut self) -> impl Future<Output = ()> {
+        struct SendClient(tbc::tb_client_t);
+        unsafe impl Send for SendClient {}
+
+        let client = std::mem::replace(&mut self.client, std::ptr::null_mut());
+        let client = SendClient(client);
+
+        let (tx, rx) = channel::<()>();
```
matklad: I usually use `channel::<std::convert::Infallible>` in situations like this --- the sender being close is already a good wake-up event.

matklad (commented):
First round of review in!

src/clients/rust/Cargo.toml
austenadler: I made a PR to @brson's repo to remove thiserror dependency: https://github.com/brson/tigerbeetle/pull/1/files

src/clients/rust/tests/tests.rs
schalk-b: I like `default` here more too. It's more explicit.

src/clients/rust/build.rs
brson: Agreed about packaging the libraries for crates.io. There are some `todo!` codepaths for this.

I planned to package static libs for this, but you are suggesting dynamic. Why would we want to use dynamic libs here? It will force users to copy them around.

src/clients/rust/Cargo.toml
brson: Merged! Thanks @austenadler

src/clients/rust/tests/tests.rs
brson: Done.

src/clients/rust/tests/tests.rs
brson: Done. Forgot this existed!

src/clients/rust/tests/tests.rs
brson: Done.

src/clients/rust/tests/tests.rs
brson: Nice. Done.

src/clients/rust/tests/tests.rs
brson: Done.

src/clients/rust/tests/tests.rs
brson: Done.

src/clients/rust/README.md
brson: Done.

src/clients/rust/build.rs
matklad: > I planned to package static libs for this, but you are suggesting dynamic. 

I think I am just wrong here, using static is the way to go!

matklad:
@brson are you looking to fix https://github.com/tigerbeetle/tigerbeetle/pull/2679#discussion_r1934335353 (avoid conversion between "Rust" types and "C" type and instead exposing just one type with a known public ABI in the API) in this PR?

src/clients/rust/src/lib.rs
```diff
@@ -0,0 +1,938 @@
...
+    ExistsWithDifferentTimeout,
+    ExistsWithDifferentDebitAccountId,
+    ExistsWithDifferentCreditAccountId,
+    ExistsWithDifferentAmount,
+    ExistsWithDifferentUserData128,
+    ExistsWithDifferentUserData64,
+    ExistsWithDifferentUserData32,
+    ExistsWithDifferentLedger,
+    ExistsWithDifferentCode,
+    Exists,
```
matklad, line 585: Ah, so this one's tricky, which I missed on the original read! `exists`, generally speaking, is not necessary an error: we use end-to-end idempotency (https://docs.tigerbeetle.com/coding/reliable-transaction-submission/), so it might be the case that the same transfer gets included into several batches due to restransmissions, and one batch will get an `Exists`. But in this case, `Exists` wouldn't be an error, as the transfer, in fact, was successfully submitted. 

Taking a step back here, I think we should not actually implement `std::error::Error` here at all? Things like `ExceedsCredits` are not actually errors, they are normal values which are part of the domain. In particular, we actually _don't_ want to have short-circuiting behavior of `?`.

🤔 

So, yeah:

* Definitely, let's remove `std::error::Error` implementation here. (but let's keep Display, it feels useful)
* I am torn on whether we should be using `Result<(), CreateTransferResult>`, or maybe just add an `Ok` variant to `CreateTransferResult`. But lets stick with Result for now.

src/clients/rust/src/lib.rs
```diff
@@ -0,0 +1,938 @@
...
+    Unexpected,
+    OutOfMemory,
+    AddressInvalid,
+    AddressLimitExceeded,
+    SystemResources,
+    NetworkSubsystem,
+    Unknown(u32),
+}
+
+impl std::error::Error for Status {}
```
matklad: In contrast, this thing is 100% an error that should be `?`-out.

brson:
> @brson are you looking to fix [#2679 (comment)](https://github.com/tigerbeetle/tigerbeetle/pull/2679#discussion_r1934335353) (avoid conversion between "Rust" types and "C" type and instead exposing just one type with a known public ABI in the API) in this PR?

Yes. Slow going but it's coming!

brson:
@matklad I pushed a commit that replaces bindgen. It has some todos still at the top that I'll address before finishing this pr. At the moment it exactly reproduces bindgen's output, but there are some changes we might make yet, based on some of your other requests, in particular regarding surfacing the C types directly in the Rust API.

brson:
I pushed another commit that fixes incorrect types for enums and packed structs in the bindings generator. This bug exists in the C generator as well, though seem innocuous in the short term. Could be worth revisiting in the future (note to self).

src/clients/rust/src/lib.rs
brson: Done.

src/clients/rust/src/lib.rs
brson: I pushed a commit that makes these types repr(C), with the same representation as the Zig types. As of now they are still two distinct types, one emitted by the code generator, and one written by hand. The handwritten ones are distinguished by some extra rustisms: a `Reserved` helper type and `bitflags!` fields.

So the conversions are still _there_ but they are transmutes now.

It doesn't seem fully satisfying, and I'm not sure if I've accomplished what you are hoping for. In particular there are still `From` impls, and those huge generic helper functions haven't changed at all. Can you give it another look?

(Still working on the error types)

matklad: I had two goals here:

- Expose the "standard" TigerBeetle ABI in the public API, to not unnecessary hide a detail with is in fact public ABI
- Remove two copies of the types and associated generic infrastructure. 

The first one is more important as that concerns the public interface, the second one is more of an implementation detail, so I am happy that we've got the first one. 

But I am wondering if hitting the second target as well would be easy now? Can we instruct the generator to not generates the types, and instead use hand-written ones somehow?

brson: Ah, great suggestions. I have changed the code to just use the hand-written Rust definitions for Account, Transfer, *Filter, AccountBalance. It lead to some good simplifications. I did though leave the codegen for the C types so I could make some assertions about the correctness of the Rust types. See what you think.

brson:
Still a little left to do here, including a merge with main.

src/clients/rust/src/lib.rs
brson: > I am torn on whether we should be using Result<(), CreateTransferResult>, or maybe just add an Ok variant to CreateTransferResult. But lets stick with Result for now.

Well I addressed this issue generally but I kanda flubbed this point: I removed the outer `Result`, just returning `CreateTransferResult`, so the user needs to match this and handle it explicitly. It was kinda accidental - I didn't realize quite what you were asking for, but I feel ok about it - I don't love putting types in the `Result::Err` position that don't implement `Error` anyway. See what you think.

matklad: Yeah, I like the current state!

src/clients/rust/src/lib.rs
matklad: Yup, this looks great! Let's also commit the generated `tp_client.rs` to git. This file is host & target independent, so there's no real reason to run build.zig every time someone builds the client. And it's useful to look at this file for code review, to make sure it looks reasonably. 

We use this pattern for committed generated code for other clients, and so far it works great:

https://github.com/tigerbeetle/tigerbeetle/blob/main/src/clients/dotnet/TigerBeetle/Bindings.cs

src/clients/rust/Cargo.toml
```diff
@@ -0,0 +1,16 @@
+[package]
+name = "tigerbeetle"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+bitflags = "2.6.0"
+futures = "0.3.31"
```
matklad: Just to double check, `futures` is our private dependency, and the public API is phrased strictly in terms of `async fn`, right? 


I think we could even depend just on `futures-channel` here?

src/clients/rust/tests/tests.rs
```diff
@@ -0,0 +1,293 @@
+use futures::executor::block_on;
+use std::env;
+use std::env::consts::EXE_SUFFIX;
+use std::io::{BufRead as _, BufReader};
+use std::mem;
+use std::process::{Child, Command, Stdio};
+use tempfile::TempDir;
```
matklad: Let's maybe open-code our own TempDir, using `CARGO_TARGET_TMPDIR` and

```rust
fn random_seed() -> u64 {
    std::hash::Hasher::finish(&std::hash::BuildHasher::build_hasher(
        &std::collections::hash_map::RandomState::new(),
    ))
}
```

to get a random number (or just use an global atomic counter).

We don't really need all of

```
dependencies = [
 "cfg-if",
 "fastrand",
 "getrandom",
 "once_cell",
 "rustix",
 "windows-sys",
]
```

to create a temporary directory for tests (creating secure temp dir in production is a different issue, as it is much harder, but we don't have to solve that problem luckily).

brson:
I've merged with main. Next I'll see what's wrong with CI, but this is almost done.

src/clients/rust/src/lib.rs
brson: I've committed tb_client.rs.

src/clients/rust/Cargo.toml
brson: Yes it is an internal implementation detail. I have changed the dependency to futures-channel; we still depend on futures as a dev-dependency (for block_on).

brson:
Almost done. Need to address the TempDir issue and probably add CI for rust.

brson:
**Oh note to self: I have forgotten to handle all the new status return values for tb_client API calls**

src/clients/rust/tests/tests.rs
brson: Done.

brson:
I have also pushed a commit to check all the status codes.

brson:
@matklad I think all rework requested is complete, and I've merged with main.

The only thing missing here is CI for the rust client. I do think we should land that here but have only just looked at what's involved, so maybe a few more days until this is ready.

zetanumbers:
Interesting to see official rust bindings finally being written. As an author of [unofficial rust bindings](https://github.com/tigerbeetle-rust/tigerbeetle-unofficial) I would like to check if safety caveats are all being followed.

brson:
@matklad do you think we should add an extra `-sys` crate here to link to `tb_client`. I hadn't even thought about it until just now. Can we punt that to followup work?

brson:
> Interesting to see official rust bindings finally being written. As an author of [unofficial rust bindings](https://github.com/tigerbeetle-rust/tigerbeetle-unofficial) I would like to check if safety caveats are all being followed.

Thanks for the link @zetanumbers. I don't think I've seen this before.

matklad:
>@matklad do you think we should add an extra -sys crate here to link to tb_client. I hadn't even thought about it until just now. Can we punt that to followup work?

No, let's keep this as a single crate for simplicity. We aren't really bindings to a system library, and I don't think we should support use-case of alternative wrappers for our -sys library. We can always split later, but I'd start with a single crate.

matklad:
> have only just looked at what's involved, so maybe a few more days until this is ready.

This shouldn't be too hard. You should add entries to https://github.com/tigerbeetle/tigerbeetle/blob/d848392be563ee6cf3569d6e1e956b92a7c36cd2/.github/workflows/ci.yml#L56-L90. I think it makes sense to test all three OSes, but only one Rust version for now. 

And then over here:

https://github.com/tigerbeetle/tigerbeetle/blob/d848392be563ee6cf3569d6e1e956b92a7c36cd2/src/scripts/ci.zig#L22

You should add


```
 .rust = @import("../clients/rust/ci.zig"),
```

where `rust/ci.zig` is a script to basically run `cargo test`. 

The principle here is that we avoid encoding testing logic into YAMLs, and orchestrate testing from Zig.

brson:
Ok, I've filled out the CI script for Rust. It runs `test`, `fmt` and `clippy`.

One caveat: I have only stubbed out the code for filling in the readme - it's a big task and I don't think we need it yet. Let's just get this in tree and tested first.

.github/workflows/ci.yml
```diff
@@ -102,6 +103,11 @@ jobs:
         uses: actions/setup-go@v5
         with: { go-version: "${{ matrix.language_version }}" }
 
+      - if: matrix.language == 'rust'
+        uses: dtolnay/rust-toolchain@master
```
matklad: This _should_ fail, as I think we deny all non-github actions on our CI, for supply chain reasons. Even if the action is compromised, our CI _shouldn't_ be able to do anything bad, as we only have sensitive permissions on the main branch, but, still. 


The good news is that GitHub runners come with rustup pre-installed, so we can use it directly here. I am a huge fan of `rustup` by the way, great tool ;-)

So I think we can do

```
- run: rustup default ${{ matrix.language_version }} && rustup component add clippy
```

or whatever is the right syntax.

matklad (commented):
Almost good to go!

.github/workflows/ci.yml
brson: Done.

matklad (dismissed):
🥳

brson:
How do I get the CI checks to complete? I haven't seen any CI results on this pr yet.

matklad:
[Invalid workflow file: .github/workflows/ci.yml#L107](https://github.com/tigerbeetle/tigerbeetle/actions/runs/14756809268/workflow)
You have an error in your yaml syntax on line 107

matklad:
Ah, GitHub as usual keeps super quiet about errors in workflows.

matklad (dismissed):

brson:
Something is going wrong with the windows node configuration. I'll investigate later.

Edit: seems to have been a transient failure

brson:
@matklad it looks like the ci is passing, but the pr needs another approving review

matklad (approved):

zetanumbers (approved):
You cannot have safe async requests which borrow indeed.

PR #2680, opened 2025-01-28
cfo: refactor and hopefully handle OOM correctly

matklad:
Best reviewed per-commit. The first few in the series are NFCs, the last one actually fixes the OOM catching:

CFO already includes code to handle OOM:

    if (std.meta.eql(term, .{ .Signal = std.posix.SIG.KILL })) {
        // Something killed the fuzzer. This is likely OOM, so count this seed
        // neither as a success, nor as a failure.
        log.info("ignored SIGKILL for '{s}'", .{fuzzer.seed.command});

Sadly, this code is effectively a no-op. The problem here is that CFO
runs fuzzers via `zig build run`, an intermediate process. So, when oom
killer terminates the fuzzing process, the build_runner process notices
that and exits with non-zero status.

To fix this, arrange it so that we run the fuzzer process directly. To
that end, add `-Dprint-exe` option to build.zig to know _what_ to run
(we can't run the binaries from `./zig-out/bin`, because, eg, `vopr` and
`vopr -Dstate-machine=testing` alias).

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -146,6 +146,11 @@ pub fn build(b: *std.Build) !void {
             "llvm-objcopy",
             "Use this llvm-objcopy instead of downloading one",
         ),
+        .print_exe = b.option(
+            bool,
+            "print-exe",
+            "Build tasks print the path of the executable",
+        ) orelse false,
     };
     const target = try resolve_target(b, build_options.target);
     const mode = b.standardOptimizeOption(.{ .preferred_optimize_mode = .ReleaseSafe });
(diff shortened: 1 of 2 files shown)


sentientwaffle (approved):
Overall LGTM. Before we merge we should either:

- remove the fuzz labels from the PR's being currently fuzzed, until they can be rebased, or
- split the `build.zig` change into a separate PR and rebase those branches on top of it before merging the CFO fix.

(Probably the first since its simpler!)

Otherwise those branches will have tons of failures as `-Dprint-exe` doesn't exist on them yet.

matklad:
I'll remove the fuzz labels!

PR #2681, opened 2025-01-28
VSR/Grid: Remove invalid zero-padding assert 

sentientwaffle:
In the grid, after we read a block, we:
1. verify that the block has a valid header, then
2. verify that the block has a valid body, then
3. verify that it is the block we were expecting to find, then
4. assert that the block's sector padding is zeroed.

The last step can fail due to disk corruption -- sector padding isn't covered by any checksum.

Despite injecting many storage faults the VOPR (previously) couldn't detect this.
This is because when the VOPR injects a storage fault, it corrupts the whole sector -- if the padding is corrupt, then the header or body (or both) is also corrupt.

We missed this bug because we injected "too many" faults!

Accordingly, this PR adds single-bit fault storage injection.

---

Edit: VOPR found a similar bug [in the superblock header padding](https://github.com/tigerbeetle/tigerbeetle/pull/2681/commits/cac302db920e6f7a80174aedb70e87eb4e939a62) and [journal prepare padding](https://github.com/tigerbeetle/tigerbeetle/pull/2681/commits/1ea02eb5fcc600d853126a275d5ee647e0f425f1).

---

And a bug in the superblock `copy` index handling: https://github.com/tigerbeetle/tigerbeetle/pull/2681/commits/f38f2921435c71ad1eb20369bc52674edb1dcfe6

Each copy of the superblock includes a copy index (0...3) so that we can detect misdirects.

We want the superblock header checksums to match even though the copies are different, so the copy index is not covered by the checksum. We read in a superblock quorum, pick arbitrarily which of the identical headers to load into memory (well, near-identical -- they have distinct copy indexes). Then later when we are about to write our superblock, when we go to compute the superblock's checksum, we assert that (`superblock.copy < constants.superblock_copies`).

But if the copy of the superblock we loaded was corrupt, this may not be true!

### DIFF
--- a/src/vsr/grid.zig
+++ b/src/vsr/grid.zig
@@ -1130,7 +1130,7 @@ pub fn GridType(comptime Storage: type) type {
             // Remove the "root" read so that the address is no longer actively reading / locked.
             grid.read_queue.remove(read);
 
-            const result = read_block_validate(block.*, .{
+            const result = grid.read_block_validate(block.*, .{
                 .address = read.address,
                 .checksum = read.checksum,
             });
@@ -1160,7 +1160,7 @@ pub fn GridType(comptime Storage: type) type {
             grid.read_block_resolve(read, result);
         }
 
-        fn read_block_validate(block: BlockPtrConst, expect: struct {
+        fn read_block_validate(grid: *const Grid, block: BlockPtrConst, expect: struct {
             address: u64,
             checksum: u128,
         }) ReadBlockResult {
@@ -1180,9 +1180,12 @@ pub fn GridType(comptime Storage: type) type {
             if (header.checksum != expect.checksum) return .unexpected_checksum;
 
             if (constants.verify) {
-                // We wrote the padding as zeroes, but it may have been corrupted, and the padding
-                // is not covered by any checksums.
-                maybe(stdx.zeroed(block[header.size..vsr.sector_ceil(header.size)]));
+                if (!stdx.zeroed(block[header.size..vsr.sector_ceil(header.size)])) {
+                    log.warn("{}: read_block_validate: found corrupted padding for address={}", .{
+                        grid.superblock.replica_index.?,
+                        expect.address,
+                    });
+                }
             }
 
             assert(header.address == expect.address);
--- a/src/vsr/grid.zig
+++ b/src/vsr/grid.zig
@@ -1179,10 +1179,6 @@ pub fn GridType(comptime Storage: type) type {
 
             if (header.checksum != expect.checksum) return .unexpected_checksum;
 
-            if (constants.verify) {
-                assert(stdx.zeroed(block[header.size..vsr.sector_ceil(header.size)]));
-            }
-
             assert(header.address == expect.address);
             return .{ .valid = block };
         }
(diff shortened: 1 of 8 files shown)


src/vsr/grid.zig
```diff
@@ -1179,10 +1179,6 @@ pub fn GridType(comptime Storage: type) type {
 
             if (header.checksum != expect.checksum) return .unexpected_checksum;
 
-            if (constants.verify) {
-                assert(stdx.zeroed(block[header.size..vsr.sector_ceil(header.size)]));
-            }
```
matklad: Should we just forcefully zero-this out then, to reduce dimensionality?

sentientwaffle: :thinking: I think that relying even on zeroed padding is a bug, so I'd rather _not_ zero the padding, to avoid potentially concealing bugs. I guess we could zero only when `!constants.verify`... but I'd rather avoid that kind of divergence between test/non-test.

sentientwaffle:
Disabling auto-merge -- the new corruption injection found another bug!

src/vsr/grid.zig
matklad: Ok, makes sense to go with offensive programming here! Worth capturing the reasoning in a comment here though!

sentientwaffle:
And another bug: https://github.com/tigerbeetle/tigerbeetle/pull/2681/commits/010a30b31696e92e4b109a021d243f815237428f

For this one I did `@memset` the padding because we assert zeroed padding already in a bunch of places, and both message buses already had `@memset`s. I'm open to copying the grid's approach of just ignoring padding instead, though.

matklad:
>For this one I did @memset the padding because we assert zeroed padding already in a bunch of places, and both message buses already had @memsets. 

Hm, yeah, this makes me think that maybe grid approach is not good? If we can't tell right off the bat whether padding to sector-size matters or not, we _will_ write some code where it matters in the future. 

Both ignoring and zeroing don't sit right with me.o

But what about this alternative:

If there's an error in the padding we've read from disk, we consider the entire block/prepare to be invalid, even if checksum matches. 

Rationale:

* Simple mental model
* Removes rare, special code path for rare condition
* If we _did_ get an error somewhere in sector, it is probably worth it to write again, so flagging block/prepare is corrupted feels desirable here. 

---

Still not sure what's the best approach here, but it definitely must be consistent between the grid and the journal.

sentientwaffle:
> If there's an error in the padding we've read from disk, we consider the entire block/prepare to be invalid, even if checksum matches.

:thinking: This would simplify the `TestStorage` changes... we could remove the sector/byte fault distinction and just always do byte faults.

Though, I don't love the fact that this approach doesn't maximize our ability to repair.

> Removes rare, special code path for rare condition

Doesn't it just replace it with a new special/rare code path? i.e. instead of `if (nonzero padding) { memset(padding, 0) }`, we have `if (nonzero padding) { message is corrupt }`.

matklad:
>Doesn't it just replace it with a new special/rare code path? i.e. instead of if (nonzero padding) { memset(padding, 0) }, we have if (nonzero padding) { message is corrupt }.

This feels a bit less special, than having the padding _most_ of the time being zero, unless there's a corruption. 


>Though, I don't love the fact that this approach doesn't maximize our ability to repair.

Yeah, my gut feeling is that:

* byte error in the padding seems a relatively unlikely 
* and if it occurs, it probably is better to repair the entire sector, than to carry on.

matklad (approved):

PR #2682, opened 2025-01-28
[VSR] Protocol batching (multi-operation)

batiati:
### VSR Protocol Batching - Multi-operation

This PR enhances the protocol batching proposal introduced in [PR #2617](https://github.com/tigerbeetle/tigerbeetle/pull/2617), following the same guidelines but allowing clients to submit batches containing different `operations` within the same request.

For more details about the motivation behind shifting the responsibility of batching from the state machine (aka `Demuxer`) to the VSR protocol, please refer to [PR #2617](https://github.com/tigerbeetle/tigerbeetle/pull/2617).

#### Motivation

Previously in #2617, we assumed that the batched request payload would always consist of a single operation. However, since the goal of protocol batching is to abstract the payload encoding from the state machine, enabling multi-operation payloads became the natural choice.

The major change lies in how the state machine process multiple cycles of prefetches for each operation, transitioning from `prefetch -> execute -> compact` to `prefetch¹ -> execute¹ -> prefetch² -> execute² -> prefetchⁿ -> executeⁿ -> compact` (where _n_ is the number of batches).

Interestingly, this change enabled batching queries at no additional cost. The primary reason queries were not batchable in #2617 was the hard limit on LSM scans, which could only run a single query during prefetching. By splitting the work into multiple prefetches, multiple queries are now allowed, up to the `limit` of results that can fit into a reply message.

To reduce the cost of context switching between prefetch and execute, adjacent batches of the same operation can, in some cases, be prefetched together in one go (for example, `create_{accounts,transfers}` and `lookup_{accounts,transfers}`), in the following sequence: `prefetch¹⁺²⁺ⁿ -> execute¹ -> execute² -> executeⁿ -> compact` (where _n_ is the number of adjacent batches).

#### VSR Changes

A new reserved VSR operation, `batched`, has been introduced to indicate whether a request is batched. Clients from older versions, which are unaware of this new operation, will always send non-batched requests, ensuring compatibility with the new protocol version.

**How it works:**

- The client (`tb_client` library) is responsible for accumulating concurrent requests, encoding them into batches, and sending them as a single VSR request with the reserved operation `batched`. The underlying state-machine operation for each batch is encoded as part of the batch trailer.

- Upon receiving a batched request, the replica decodes the message body, splits it into multiple batch payloads, and feeds the state machine with independent units of work under the same VSR `op` number.

- Each batch may generate a number of `result`s unrelated to the number of `event`s (e.g., `query_transfers` and `lookup_transfers`). These results are encoded into a batched reply and sent back to the client.

- The client decodes the reply's body into multiple payloads and invokes the respective user callback for each one.

**Encoding layout:**

- Batched requests use a portion at the end of the message body (the batch `trailer`) to encode batch metadata. As a result, batched requests can hold fewer events than non-batched ones.

- The `trailer` size is determined by the number of batches, requiring 4 bytes per batch (a 24-bit integer for the size and an 8-bit integer for the operation), plus a _postamble_ at the very end of the trailer containing the `batch_count`.

- To minimize the number of copies during encoding, the trailer is written from the end of the message towards the beginning. For example, the last element of the array corresponds to the number of events in the first batch (as in a stack).

- Batches of different operations may require padding between them to maintain alignment. Padding bytes are always zeroed. The padding between batches is not stored in the encoding and should be calculated based on the size/alignment expected by the state machine (e.g., by calling `std.mem.alignForward`). The `vsr.batch` implementation abstracts this complexity.

### Example

Consider a **reply** message containing 4 different operations:

| # | operation            | operation code | result size |
|---|----------------------|----------------|-------------|
| 1 |`lookup_accounts`     | 131            | 128 bytes   |
| 2 |`create_accounts`     | 129            |   8 bytes   |
| 3 |`lookup_transfers`    | 132            | 128 bytes   |
| 4 |`get_account_balances`| 134            | 128 bytes   |
-----------------------------------------------------------

```
size == 1076   message.body_used().len == 820 bytes    trailer == 20 bytes
               payload == 800 bytes                    batches (4 bytes each)          postamble (4 bytes)   
┌──────┐ ┌───────────────────────────────────────────┐┌─────────────────────────────────────┐
│      │ │┌─────────┐┌─────────┐┌────────┐┌─────────┐││┌───┐┌───┐┌───┐┌───┐            ┌───┐│
│ VSR  │ ││512 bytes││ 24 bytes││ 0 bytes││264 bytes││││264││  0││ 24││512│← size      │ 4 ││← batch_count
|Header| |│4 results││3 results││0 result││2 results││││134||132││129││131│← operation │ 0 ││← reserved
│      │ │└───▲─────┘└───▲─────┘└────▲───┘└──▲──────┘││└─┬─┘└─┬─┘└─┬─┘└─┬─┘            └───┘│
└──────┘ └────┼──────────┼───────────┼───────┼───────┘└──┼────┼────┼────┼───────────────────┘
              │          │           │       │  8 bytes  │    │    │    │  
              │          │           │       └──padding──┘    │    │    │  
              │          │           │                        │    │    │  
              │          │           └────────────────────────┘    │    │  
              │          │                                         │    │  
              │          └─────────────────────────────────────────┘    │  
              │                                                         │  
              └─────────────────────────────────────────────────────────┘  
```

- The `tb_client` avoids batching user requests if the combined payload length and trailer exceed the maximum message body size. Instead, it sends a non-batched request to preserve maximum capacity (up to 8190 events per message).

- The `tb_client` attempts to pack batches of the same operation sequentially, reducing the need for multiple prefetch cycles.

- The client can pack as many batches as the sum of payloads and trailer allows.

### This PR

See the commit messages for implementation details (most of which are the same commits from PR #2617).

This PR includes VOPR and Vortex (Java driver only) tests. However, Vortex can only process a single operation at a time, so multi-operation batches are not explored.

### DIFF
--- a/src/clients/c/tb_client/context.zig
+++ b/src/clients/c/tb_client/context.zig
@@ -267,6 +267,11 @@ pub fn ContextType(
                     @panic("IO.run() failed");
                 };
             }
+
+            // If evicted, the inflight request was already canceled during eviction.
+            if (self.eviction_reason == null) {
+                self.cancel_request_inflight();
+            }
             self.cancel_queued_packets();
         }
 
@@ -280,6 +285,17 @@ pub fn ContextType(
             while (self.submitted.pop()) |packet| {
                 packet.assert_phase(.submitted);
                 self.packet_cancel(packet);
+            }
+        }
+
+        /// Cancel the current inflight packet, as it won't be replied anymore.
+        fn cancel_request_inflight(self: *Context) void {
+            if (self.client.request_inflight) |*inflight| {
+                if (inflight.message.header.operation != .register) {
+                    const packet = @as(UserData, @bitCast(inflight.user_data)).packet;
+                    packet.assert_phase(.sent);
+                    self.packet_cancel(packet);
+                }
             }
         }
 
@@ -579,14 +595,7 @@ pub fn ContextType(
 
             self.eviction_reason = eviction.header.reason;
 
-            // Cancel the current inflight packet, as it won't be replied anymore.
-            if (self.client.request_inflight) |*inflight| {
-                if (inflight.message.header.operation != .register) {
-                    const packet = @as(UserData, @bitCast(inflight.user_data)).packet;
-                    packet.assert_phase(.sent);
-                    self.packet_cancel(packet);
-                }
-            }
+            self.cancel_request_inflight();
             self.cancel_queued_packets();
         }
 
--- a/src/clients/c/tb_client/context.zig
+++ b/src/clients/c/tb_client/context.zig
@@ -634,7 +634,7 @@ pub fn ContextType(
                     result_size,
                     reply,
                     batch_count,
-                );
+                ) catch unreachable;
 
                 var it: ?*Packet = packet;
                 while (it) |batched| {
--- a/src/clients/c/tb_client/context.zig
+++ b/src/clients/c/tb_client/context.zig
@@ -15,6 +15,9 @@ const Header = vsr.Header;
 const IO = vsr.io.IO;
 const message_pool = vsr.message_pool;
 
+const BatchEncoder = vsr.BatchEncoder;
+const BatchDecoder = vsr.BatchDecoder;
+
 const MessagePool = message_pool.MessagePool;
 const Message = MessagePool.Message;
 const Packet = @import("packet.zig").Packet;
@@ -249,13 +252,6 @@ pub fn ContextType(
             self.allocator.destroy(self);
         }
 
-        pub fn tick(self: *Context) void {
-            // `Tick` is invoked by the result callback that may have arrived during shutdown.
-            if (self.state.load(.acquire) == .running) {
-                self.client.tick();
-            }
-        }
-
         pub fn io_thread(self: *Context) void {
             while (!self.signal.stop_requested()) {
                 self.client.tick();
@@ -268,40 +264,41 @@ pub fn ContextType(
                 };
             }
 
-            // Cancel the request_inflight packet if any.
-            if (self.client.request_inflight) |*inflight| {
-                if (inflight.message.header.operation != .register) {
-                    const user_data: UserData = @bitCast(inflight.user_data);
-                    self.cancel(user_data.packet);
+            // Cancelling any inflight or queued request.
+            var packet_list: ?*Packet = packet_list: {
+                if (self.client.request_inflight) |*inflight| {
+                    if (inflight.message.header.operation != .register) {
+                        const user_data: UserData = @bitCast(inflight.user_data);
+                        // Cancel the request_inflight packet if any.
+                        break :packet_list user_data.packet;
+                    }
                 }
-            }
 
-            while (self.submitted.pop()) |packet| {
+                break :packet_list self.submitted.flush();
+            };
+            while (packet_list) |packet| {
+                packet_list = packet.next orelse self.submitted.flush();
+                packet.next = null;
                 self.cancel(packet);
             }
         }
 
         /// Submit the enqueued requests through the vsr client.
         /// Always called by the io thread.
-        fn submit(self: *Context, packet: *Packet) void {
+        fn submit(self: *Context, packet_list: *Packet) void {
             assert(self.batch_size_limit != null);
             assert(self.client.request_inflight == null);
-            assert(packet.next == null);
 
-            const message = self.client.get_message().build(.request);
-            defer self.client.release_message(message.base());
-
-            // The batching logic will be inserted here:
-            const head: *Packet, const bytes_written: usize, const batch_count: u16 = blk: {
-                if (!self.packet_valid(packet)) {
-                    // Packet failed, nothing to send.
-                    self.signal.notify();
-                    return;
+            // Find a valid packet to start the batch.
+            const head: *Packet = head: {
+                var it: ?*Packet = packet_list;
+                while (it) |packet| {
+                    it = packet.next;
+                    if (self.packet_valid(packet)) break :head packet;
                 }
-
-                const event_data = packet.events();
-                stdx.copy_disjoint(.inexact, u8, message.buffer[@sizeOf(Header)..], event_data);
-                break :blk .{ packet, event_data.len, 0 };
+                // Nothing to send.
+                self.signal.notify();
+                return;
             };
 
             const operation: StateMachine.Operation = operation_from_int(head.operation).?;
@@ -310,14 +307,89 @@ pub fn ContextType(
                 .pulse => unreachable,
                 inline else => |operation_comptime| blk: {
                     const event_size = @sizeOf(StateMachine.EventType(operation_comptime));
-                    const events_batch_max = StateMachine.operation_batch_max(
-                        operation_comptime,
-                        self.batch_size_limit.?,
-                    );
+                    const events_batch_max = if (StateMachine.event_is_slice(operation))
+                        StateMachine.operation_batch_max(
+                            operation_comptime,
+                            self.batch_size_limit.?,
+                        )
+                    else
+                        1;
 
                     break :blk .{ event_size, events_batch_max };
                 },
             };
+            assert(self.batch_size_limit.? >= event_size * events_batch_max);
+
+            const message = self.client.get_message().build(.request);
+            defer self.client.release_message(message.base());
+
+            var encoder = BatchEncoder.init(
+                event_size,
+                message.buffer[@sizeOf(Header)..][0 .. event_size * events_batch_max],
+            );
+
+            // Whether to batch this packet with others in the same request,
+            // or send it on it's own message.
+            const is_batchable = head.next != null and
+                StateMachine.event_is_slice(operation) and
+                encoder.can_add(head.events().len);
+
+            const bytes_written: usize, const batch_count: u16 = if (is_batchable) blk: {
+
+                // ********************************************************************************
+                // NOTE: This algorithm iterates over the linked list in O(n) time for each request
+                // and is susceptible to starvation, as it reorders packets when returning them to
+                // the queue. It's merely an attempt to remove the batch-specific linked list
+                // fields from `Packet`.
+                // ********************************************************************************
+
+                // Encoding the first packet into the batch:
+                stdx.copy_disjoint(.inexact, u8, encoder.writable(), head.events());
+                encoder.add(head.events().len);
+
+                // Iterating for other packets with the same `operation` that might be sent together
+                // in the same message.
+                var it: ?*Packet = head.next;
+                var packet_tail = head;
+                packet_tail.next = null;
+                while (it) |batch| {
+                    it = batch.next;
+                    if (!self.packet_valid(batch)) continue;
+
+                    if (batch.operation == head.operation and
+                        encoder.can_add(batch.events().len))
+                    {
+                        const event_data = batch.events();
+                        stdx.copy_disjoint(.inexact, u8, encoder.writable(), event_data);
+                        encoder.add(event_data.len);
+
+                        packet_tail.next = batch;
+                        batch.next = null;
+                        packet_tail = batch;
+                    } else {
+                        // There's no room for this packet, returning it to the queue,
+                        // it will be sent next time.
+                        self.submitted.push(batch);
+                    }
+                }
+                encoder.finish();
+
+                break :blk .{ encoder.bytes_written, encoder.batch_count };
+            } else blk: {
+                var it: ?*Packet = head.next;
+                head.next = null;
+                while (it) |batch| {
+                    it = batch.next;
+                    if (self.packet_valid(batch)) {
+                        self.submitted.push(batch);
+                    }
+                }
+
+                const event_data = head.events();
+                stdx.copy_disjoint(.inexact, u8, message.buffer[@sizeOf(Header)..], event_data);
+                break :blk .{ event_data.len, 0 };
+            };
+            assert(batch_count == 0 or is_batchable);
             assert(bytes_written % event_size == 0);
             assert(bytes_written <= events_batch_max * event_size);
 
@@ -346,10 +418,10 @@ pub fn ContextType(
         /// When the result is `false` then `packet: *Packet` isn't valid anymore.
         fn packet_valid(self: *Context, packet: *Packet) bool {
             assert(packet.status == .ok);
-            assert(packet.user_data != null);
             assert(packet.data != null or packet.data_size == 0);
             assert(stdx.zeroed(&packet.reserved));
-            assert(packet.next == null);
+            maybe(packet.next == null);
+            maybe(packet.user_data == null);
 
             const operation: StateMachine.Operation = operation_from_int(packet.operation) orelse {
                 self.notify_completion(packet, error.InvalidOperation);
@@ -465,8 +537,8 @@ pub fn ContextType(
 
             // Submit the next pending packet (if any) now that VSR has completed this one.
             assert(self.client.request_inflight == null);
-            if (self.submitted.pop()) |packet_next| {
-                self.submit(packet_next);
+            if (self.submitted.flush()) |packet_list| {
+                self.submit(packet_list);
             }
 
             // `StateMachine.result_size_bytes` is intended for use on the replica side,
@@ -479,14 +551,31 @@ pub fn ContextType(
             } else unreachable;
 
             assert(result_size % result_size == 0);
-            // Implement batching decode here, using
-            // the linked list and batch_count.
-            assert(packet.next == null);
-            assert(batch_count == 0);
-            self.notify_completion(packet, .{
-                .timestamp = timestamp,
-                .reply = reply,
-            });
+            if (batch_count == 0) {
+                assert(packet.next == null);
+                self.notify_completion(packet, .{
+                    .timestamp = timestamp,
+                    .reply = reply,
+                });
+            } else {
+                assert(batch_count > 1 or packet.next == null);
+                var decoder = BatchDecoder.init(
+                    result_size,
+                    reply,
+                    batch_count,
+                );
+                var it: ?*Packet = packet;
+                while (it) |batch| {
+                    it = batch.next;
+                    batch.next = null;
+
+                    self.notify_completion(batch, .{
+                        .timestamp = timestamp,
+                        .reply = decoder.next().?,
+                    });
+                }
+                assert(decoder.next() == null);
+            }
         }
 
         fn signal_notify_callback(signal: *Signal) void {
@@ -501,8 +590,8 @@ pub fn ContextType(
 
             switch (self.state.load(.acquire)) {
                 .running => if (self.client.request_inflight == null) {
-                    if (self.submitted.pop()) |packet| {
-                        self.submit(packet);
+                    if (self.submitted.flush()) |packet_list| {
+                        self.submit(packet_list);
                     }
                 },
                 .evicted => return,
@@ -554,8 +643,6 @@ pub fn ContextType(
                 reply: []const u8,
             },
         ) void {
-            assert(packet.next == null);
-
             const completion_ctx = self.implementation.completion_ctx;
             const tb_client = api.context_to_client(&self.implementation);
             const result = completion catch |err| {
--- a/src/clients/c/tb_client/context.zig
+++ b/src/clients/c/tb_client/context.zig
@@ -468,6 +468,7 @@ pub fn ContextType(
                 .command = .request,
                 .operation = vsr.Operation.from(StateMachine, operation),
                 .size = @sizeOf(vsr.Header) + packet.batch_size,
+                .batch_count = 0,
             };
 
             // Copy all batched packet event data into the message.
--- a/src/clients/c/tb_client/context.zig
+++ b/src/clients/c/tb_client/context.zig
@@ -429,7 +429,7 @@ pub fn ContextType(
             self.pending.push(packet);
         }
 
-        /// Sends the packet (the entire batched liked list of packets) through the vsr client.
+        /// Sends the packet (the entire batched linked list of packets) through the vsr client.
         /// Always called by the io thread.
         fn packet_send(self: *Context, packet: *Packet) void {
             assert(self.batch_size_limit != null);
@@ -450,8 +450,10 @@ pub fn ContextType(
             };
 
             const message = self.client.get_message().build(.request);
-            // Don't need to release the message, since this function cannot fail/return early.
-            defer packet.assert_phase(.sent);
+            defer {
+                self.client.release_message(message.base());
+                packet.assert_phase(.sent);
+            }
 
             const bytes_written: usize = bytes_written: {
                 if (packet.batch_count == 0) {
@@ -512,8 +514,9 @@ pub fn ContextType(
                     .self = self,
                     .packet = packet,
                 }),
-                message,
+                message.ref(),
             );
+            assert(message.header.request != 0);
         }
 
         fn signal_notify_callback(signal: *Signal) void {
@@ -634,7 +637,12 @@ pub fn ContextType(
                     result_size,
                     reply,
                     batch_count,
-                ) catch unreachable;
+                ) catch std.debug.panic("client_result_callback: invalid batch: " ++
+                    "operation={s} reply.len={} batch_count={}", .{
+                    @tagName(operation),
+                    reply.len,
+                    batch_count,
+                });
 
                 var it: ?*Packet = packet;
                 while (it) |batched| {
--- a/src/clients/c/tb_client/context.zig
+++ b/src/clients/c/tb_client/context.zig
@@ -68,7 +68,7 @@ pub fn ContextType(
                 }
             }.is_valid,
             .element_size = struct {
-                fn element_size(_: void, vsr_operation: vsr.Operation) usize {
+                fn element_size(_: void, vsr_operation: vsr.Operation) u32 {
                     const operation = StateMachine.operation_from_vsr(vsr_operation).?;
                     return switch (operation) {
                         .pulse => unreachable,
@@ -79,7 +79,7 @@ pub fn ContextType(
                 }
             }.element_size,
             .alignment = struct {
-                fn alignment(_: void, vsr_operation: vsr.Operation) usize {
+                fn alignment(_: void, vsr_operation: vsr.Operation) u32 {
                     const operation = StateMachine.operation_from_vsr(vsr_operation).?;
                     return switch (operation) {
                         .pulse => unreachable,
@@ -101,7 +101,7 @@ pub fn ContextType(
                 }
             }.is_valid,
             .element_size = struct {
-                fn element_size(_: void, vsr_operation: vsr.Operation) usize {
+                fn element_size(_: void, vsr_operation: vsr.Operation) u32 {
                     const operation = StateMachine.operation_from_vsr(vsr_operation).?;
                     return switch (operation) {
                         .pulse => unreachable,
@@ -112,7 +112,7 @@ pub fn ContextType(
                 }
             }.element_size,
             .alignment = struct {
-                fn alignment(_: void, vsr_operation: vsr.Operation) usize {
+                fn alignment(_: void, vsr_operation: vsr.Operation) u32 {
                     const operation = StateMachine.operation_from_vsr(vsr_operation).?;
                     return switch (operation) {
                         .pulse => unreachable,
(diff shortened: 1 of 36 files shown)


src/state_machine.zig
```diff
@@ -709,7 +702,7 @@ pub fn StateMachineType(
             self: *const StateMachine,
             client_release: vsr.Release,
             operation: Operation,
-            input: []align(16) const u8,
+            input: []const u8,
```
batiati, line 774: Vopr seed: `./zig/zig build -Drelease vopr -- --lite 12293026746782977520 `

Cause: Vopr workload generating too many batched events.

Root cause: `StateMachine.input_valid` is not enough to validate a batched request. It validates each batch isolated, but we must also validate all batches combined and check if they fit into the commit limits.

I :heart: VOPR!

src/vsr/batch.zig
```diff
@@ -0,0 +1,804 @@
...
+const testing = std.testing;
+
+const stdx = @import("../stdx.zig");
+const assert = std.debug.assert;
+const maybe = stdx.maybe;
+
+const constants = @import("../constants.zig");
+const vsr = @import("../vsr.zig");
+
+const Postamble = packed struct(u32) {
```
sentientwaffle, line 11: Why `packed` instead of `extern`?

src/vsr/batch.zig
```diff
@@ -0,0 +1,804 @@
...
+        alignment: fn (Context, vsr.Operation) usize,
+    },
+) type {
+    return struct {
+        fn total_size(options: struct {
+            context: Context,
+            current_payload_size: usize,
+            current_batch_count: u16,
+            next_operation: vsr.Operation,
+            next_payload_size: usize,
```
sentientwaffle: Naming wise, are `current_payload_size` and `next_payload_size` using "payload" to refer to two different things?

The former is the cumulative size of the batch bodies. The latter is a single batch body.

What about:

```suggestion
            current_payload_size: usize,
            current_batch_count: u16,
            next_batch_operation: vsr.Operation,
            next_batch_size: usize,
```

:thinking: ...Though, that isn't great either. There are two different levels of batches -- the data we are encoding is a batch (of batches), and each chunk corresponding to an operation is itself a batch.

Can we disambiguate? What if we just (consistently) used a different synonym for the former? e.g. "parcel" / "protocol parcel" / `Operation.parcel` / `vsr/parcel.zig`?

src/vsr/batch.zig
```diff
@@ -0,0 +1,804 @@
...
+                    alignment,
+                );
+                // Assuming aligned buffers, so the first batch will not require padding.
+                assert(padding == 0 or options.current_payload_size > 0);
+                assert(std.mem.isAligned(
+                    options.current_payload_size + padding,
+                    alignment,
+                ));
+                break :padding padding;
+            };
```
sentientwaffle: Refactor so that the asserts cover both branches:

```suggestion
            const padding: u16 = padding: {
                if (options.next_payload_size == 0) {
                    break :padding 0;
                } else {
                    break :padding align_padding(
                        options.current_payload_size,
                        alignment,
                    );
                }
            };
            // Assuming aligned buffers, so the first batch will not require padding.
            assert(padding == 0 or options.current_payload_size > 0);
            assert(std.mem.isAligned(
                options.current_payload_size + padding,
                alignment,
            ));
```

sentientwaffle: The 256-byte message header is also part of the message -- should we assert that we are aligned relative to that too?

src/vsr/batch.zig
sentientwaffle, line 1: Per tigerstyle, avoid `usize` -- in some of the cases in this file (e.g. `align_padding()`), using `usize` might lead to unexpected differences when building on wasm with 32-bit.

Likewise for how we use pointer alignment, though that might be trickier. e.g. image a 32-bit client talking to a 64-bit replica.

src/vsr/batch.zig
```diff
@@ -0,0 +1,804 @@
...
+                body[body.len - @sizeOf(Postamble) ..],
+            ));
+            if (postamble.batch_count == 0) return error.BatchInvalid;
+            if (postamble.reserved != 0) return error.BatchInvalid;
+
+            const trailer_size = trailer_total_size(postamble.batch_count);
+            if (body.len < trailer_size) return error.BatchInvalid;
+
+            const payload_end = body.len - trailer_size;
+            const payload: []const u8 = body[0..payload_end];
```
sentientwaffle: More verbose, but better to be explicit:
```suggestion
            const payload_with_padding_size = body.len - trailer_size;
            const payload_with_padding: []const u8 = body[0..payload_with_padding_size];
```
(I am having trouble keeping track of which counts include padding :sweat_smile:)

src/vsr/batch.zig
sentientwaffle, line 11: Checking my understanding -- is the reason that the `Postamble` is at the end of the body rather than the beginning (since it is fixed size, unlike the list of trailer items) to avoid adding padding afterwards?

src/vsr/batch.zig
sentientwaffle, line 1: A module-level comment with some terminology definitions might be useful -- e.g. "element", "payload", "trailer", "trailer item", "postamble", "batch".

Also, maybe a simplified version of your diagram, to show the schema? In particular, pointing out the different alignments/paddings -- I find those hard to keep track of.

src/vsr/batch.zig
```diff
@@ -0,0 +1,804 @@
...
+            const aligned: []u8 = buffer[self.buffer_index + padding .. buffer.len - trailer_size];
+
+            // Trim the slice to the maximum number of elements that can fit.
+            const element_size = adapter.element_size(self.context, operation);
+            const size: usize = @divFloor(aligned.len, element_size) * element_size;
+
+            return aligned[0..size];
+        }
+
+        /// Records how many bytes were writen in the slice previously acquired by `writable()`.
```
sentientwaffle:
```suggestion
        /// Records how many bytes were written in the slice previously acquired by `writable()`.
```

src/vsr/batch.zig
```diff
@@ -0,0 +1,804 @@
...
+            }
+
+            const trailer_padding: usize = align_padding(
+                elements_size_total,
+                @alignOf(TrailerItem),
+            );
+            if (elements_size_total + trailer_padding != payload.len) {
+                return error.BatchInvalid;
+            }
+            if (trailer_padding > 0 and !stdx.zeroed(payload[payload.len - trailer_padding ..])) {
```
sentientwaffle: Simplify to
```suggestion
            if (!stdx.zeroed(payload[payload.len - trailer_padding ..])) {
```

src/vsr/batch.zig
```diff
@@ -0,0 +1,804 @@
...
+    reserved: u16,
+};
+
+const TrailerItem = packed struct(u32) {
+    /// The size in bytes of each batch.
+    /// The size of each element is operation-specific, and padding bytes may be added
+    /// at the beginning of the batch to maintain alignment.
+    /// Use `std.mem.alignForward` to compute the correct starting position for the required
+    /// alignment. Padding bytes are zeroed.
+    size: u24,
```
sentientwaffle, line 22: Maybe clarify in the comment that `size` doesn't include the padding?

src/vsr/batch.zig
```diff
@@ -0,0 +1,804 @@
...
+    comptime {
+        assert(@sizeOf(TrailerItem) == @sizeOf(Postamble));
+        assert(@alignOf(TrailerItem) == @alignOf(Postamble));
+    }
+};
+
+/// The trailer is a `[]BatchItem` containing the operation and the number of elements
+/// in each batch. To encode the trailer, `batch_count * @sizeOf(BatchItem)` bytes are
+/// needed, plus a `Postamble`.
+fn trailer_total_size(batch_count: u16) usize {
```
sentientwaffle: It is confusing that the "total trailer size" _doesn't_ include padding, while the `trailer_size` (returned by `total_size`) _does_ include padding. Maybe `trailer_unpadded_size`?

src/vsr/batch.zig
```diff
@@ -0,0 +1,804 @@
...
+    return (@as(usize, batch_count) * @sizeOf(TrailerItem)) + @sizeOf(Postamble);
+}
+
+fn align_padding(target: anytype, alignment: usize) u16 {
```
sentientwaffle: :thinking: Maybe `padding_for_alignment` or `alignment_padding`? `align_padding` makes it sound like we are aligning the _padding_, which isn't the case.

src/vsr/batch.zig
```diff
@@ -0,0 +1,804 @@
...
+            const trailer_size = trailer_total_size(self.batch_count);
+            assert(buffer.len >= self.buffer_index + trailer_size);
+
+            const trailer: []u8 = buffer[buffer.len - trailer_size ..];
+            const trailer_items: []TrailerItem = @alignCast(std.mem.bytesAsSlice(
+                TrailerItem,
+                trailer[0 .. trailer.len - @sizeOf(Postamble)],
+            ));
+            assert(trailer_items.len == self.batch_count);
+            if (constants.verify) {
```
sentientwaffle, line 547: Add an explanation that these asserts are paired with corresponding setup in `.writable()` -- without that context they don't make sense.

src/vsr/batch.zig
```diff
@@ -0,0 +1,804 @@
...
+            maybe(self.batch_count == 0);
+
+            // Takes into account extra trailer bytes that will need to be included.
+            const trailer_size: usize = trailer_total_size(
+                self.batch_count + 1,
+            );
+
+            const buffer: []u8 = self.buffer.?;
+            const padding: u16 = align_padding(
+                buffer[self.buffer_index..],
```
sentientwaffle: Passing a pointer to `align_padding` and using the result to inform our data layout makes me nervous -- could the result vary if by chance `buffer` was allocated on a particular offset? Could the result vary depending on whether we are a 32/64-bit build?

src/vsr/batch.zig
```diff
@@ -0,0 +1,804 @@
...
+                adapter.alignment(self.context, operation),
+            );
+            // The first batch must not include padding since the buffer is expected to be aligned.
+            assert(padding == 0 or self.buffer_index > 0);
+
+            if (buffer.len <= self.buffer_index + padding + trailer_size) {
+                // Insuficient space for one more batch.
```
sentientwaffle:
```suggestion
                // Insufficient space for one more batch.
```

src/state_machine/workload.zig
```diff
@@ -303,124 +402,436 @@ pub fn WorkloadType(comptime AccountingStateMachine: type) type {
...
+                                assert(limit >= count * @sizeOf(Event));
+
+                                break :size .{ count * @sizeOf(Event), count * @sizeOf(Result) };
+                            },
+                            .get_account_transfers, .get_account_balances => size: {
+                                const filter: *Event = self.single_event(
+                                    Event,
+                                    action_comptime,
+                                    writable,
+                                ) orelse break; // Insuficient space.
```
sentientwaffle:
```suggestion
                                ) orelse break; // Insufficient space.
```

src/state_machine/workload.zig
```diff
@@ -303,124 +402,436 @@ pub fn WorkloadType(comptime AccountingStateMachine: type) type {
...
+                                    @sizeOf(Event),
+                                    filter.limit * @sizeOf(Result),
+                                };
+                            },
+                            .query_accounts, .query_transfers => size: {
+                                const filter: *Event = self.single_event(
+                                    Event,
+                                    action_comptime,
+                                    writable,
+                                ) orelse break; // Insuficient space.
```
sentientwaffle:
```suggestion
                                ) orelse break; // Insufficient space.
```

src/vsr/batch.zig
```diff
@@ -0,0 +1,804 @@
...
+                .size_16 => @alignOf(u128),
+                .size_32 => @alignOf(u32),
+                .size_64 => @alignOf(u64),
+                .size_128 => @alignOf(u128),
+                .size_256 => @alignOf(u128),
+            };
+        }
+    };
+
+    fn run(options: struct {
```
sentientwaffle: Rewrote this function here since I had too many suggestions and wanted to experiment: https://github.com/tigerbeetle/tigerbeetle/commit/6d6492c2e7c6ab3f044e61b0fb33feb2ca62489d

Primary goal is to not rely on `BatchEncoder` when generating the data to be encoded. We are testing encoder, so we shouldn't trust it!

src/vsr/client.zig
```diff
@@ -21,13 +21,12 @@ pub fn ClientType(
         const Client = @This();
 
         pub const StateMachine = StateMachine_;
-        pub const DemuxerType = StateMachine.DemuxerType;
         pub const Request = struct {
             pub const Callback = *const fn (
                 user_data: u128,
-                operation: StateMachine.Operation,
+                operation: vsr.Operation,
```
sentientwaffle, line 27: ~Is the reason for `s/StateMachine/vsr` that using the state machine operations at the "top level" is essentially deprecated?~ Ah right, it is because we need to be able to pass `Operation.batched`.

src/vsr/batch.zig
batiati: But it does include paddind!
We really need comments here :grin:

src/vsr/batch.zig
batiati: Yes, since it's 4 bytes, payloads with `align(8)` and `align(16)` would require padding just at the beginning of the message.

Also, it makes sense as the entire trailer is written in the reverse order.

batiati: Just to be symmetrical with `TrailerItem`.

src/clients/c/tb_client/context.zig
```diff
@@ -56,6 +58,131 @@ pub fn ContextType(
...
+                        .pulse => unreachable,
+                        inline else => |comptime_operation| @alignOf(
+                            StateMachine.ResultType(comptime_operation),
+                        ),
+                    };
+                }
+            }.alignment,
+        });
+
+        fn result_expected_size(
```
sentientwaffle, line 127: Should this be `reply_body_expected_size`? So that it isn't confused with `StateMachine.Result`.

src/clients/c/tb_client/context.zig
```diff
@@ -240,6 +360,10 @@ pub fn ContextType(
         pub fn deinit(self: *Context) void {
             self.signal.stop();
             self.thread.join();
+
+            assert(self.submitted.pop() == null);
+            assert(self.pending.pop() == null);
```
sentientwaffle, line 365: Or:
```suggestion
            assert(self.submitted.empty());
            assert(self.pending.empty());
```

src/clients/c/tb_client/context.zig
```diff
@@ -290,317 +391,499 @@ pub fn ContextType(
...
+                    .context = {},
+                    .current_payload_size = root.batch_results_size,
+                    .current_batch_count = @max(1, root.batch_count) + 1,
+                    .next_batch_operation = vsr.Operation.from(StateMachine, operation),
+                    .next_batch_size = results_size,
+                });
+                // While the request size is constrained by `batch_size_limit`,
+                // the reply size is only limited by `message_body_size_max`.
+                if (encoded_results.payload_size + encoded_results.trailer_size >
+                    constants.message_body_size_max) continue;
```
sentientwaffle, line 523: These two conditions are nearly the same. Matching the order makes their similarities more obvious.

```suggestion
                if (encoded_events.payload_size + encoded_events.trailer_size >
                    self.batch_size_limit.?) continue;

                const encoded_results = BatchDecoder.encoded_total_size(.{
                    .context = {},
                    .current_payload_size = root.batch_results_size,
                    .current_batch_count = @max(1, root.batch_count) + 1,
                    .next_batch_operation = vsr.Operation.from(StateMachine, operation),
                    .next_batch_size = results_size,
                });
                // While the request size is constrained by `batch_size_limit`,
                // the reply size is only limited by `message_body_size_max`.
                if (encoded_results.payload_size + encoded_results.trailer_size >
                    constants.message_body_size_max) continue;
```

src/clients/c/tb_client/context.zig
```diff
@@ -290,317 +391,499 @@ pub fn ContextType(
...
+
+                if (root.batch_count == 0) {
+                    assert(root.batch_next == null);
+                    // Although the protocol allows `batch_count == 1`,
+                    // the client never packages a single packet into a batched request.
+                    // Instead, it sends a non-batched request to avoid wasting message
+                    // body bytes with the batch trailer.
```
sentientwaffle, line 530: Possibly worth discussing in the W+T, but it might make sense to just always send batched requests, to reduce client dimensionality.

src/clients/c/tb_client/context.zig
```diff
@@ -290,317 +391,499 @@ pub fn ContextType(
...
-                    => |tag| linked_chain_open: {
-                        const Event = StateMachine.EventType(tag);
-                        // Packet data isn't necessarily aligned.
-                        const events: [*]align(@alignOf(u8)) const Event = @ptrCast(data.?);
-                        const events_count: usize = @divExact(data_size, @sizeOf(Event));
-                        break :linked_chain_open events[events_count - 1].flags.linked;
-                    },
-                    else => false,
-                };
+        fn batch_merge(self: *Context) ?*Packet {
```
sentientwaffle, line 555: Why do we need to perform batching both here _and_ in `packet_enqueue`?

src/clients/c/tb_client/context.zig
```diff
@@ -290,317 +391,499 @@ pub fn ContextType(
...
-            return true;
+            assert(root.batch_events_size <= self.batch_size_limit.?);
+            assert(root.batch_results_size <= self.batch_size_limit.?);
+            return root;
         }
 
-        fn submit(self: *Context, packet: *Packet) void {
+        /// Sends the packet (the entire batched linked list of packets) through the vsr client.
+        /// Always called by the io thread.
+        fn packet_send(self: *Context, packet: *Packet) void {
```
sentientwaffle, line 630: Maybe call this `root` (as we do elsewhere) to hint that this is the beginning of a chain of packets?

src/clients/c/tb_client/context.zig
```diff
@@ -290,317 +391,499 @@ pub fn ContextType(
...
+
+        fn client_register_callback(user_data: u128, result: *const vsr.RegisterResult) void {
+            const self: *Context = @ptrFromInt(@as(usize, @intCast(user_data)));
+            assert(self.client.request_inflight == null);
+            assert(self.batch_size_limit == null);
+            assert(result.batch_size_limit > 0);
+
+            // The client might have a smaller message size limit.
+            maybe(constants.message_body_size_max < result.batch_size_limit);
+            self.batch_size_limit = @min(result.batch_size_limit, constants.message_body_size_max);
```
sentientwaffle, line 756: When is this true? During a test?

src/clients/c/tb_client/context.zig
```diff
@@ -290,317 +391,499 @@ pub fn ContextType(
...
+                @intFromEnum(eviction.header.reason),
+            });
+
+            self.eviction_reason = eviction.header.reason;
+
+            self.cancel_request_inflight();
+            self.cancel_queued_packets();
+        }
+
+        fn client_result_callback(
```
sentientwaffle, line 778: I know this was previously called `on_result`, but "result" typically refers to `StateMachine.Result`. Reply is more accurate:

```suggestion
        fn client_reply_callback(
```

src/clients/c/tb_client/context.zig
```diff
@@ -290,317 +391,499 @@ pub fn ContextType(
...
+                // the client never packages a single packet into a batched request.
+                assert(vsr_operation == .batched);
+                assert(packet.batch_next != null);
+                assert(packet.batch_count > 1);
+
+                var decoder = BatchDecoder.init({}, reply) catch {
+                    std.debug.panic("client_result_callback: invalid batch: " ++
+                        "reply.len={}", .{
+                        reply.len,
+                    });
```
sentientwaffle, line 827: Maybe log the `timestamp` and `packet.batch_count` as well, since we have those handy?

This would be a really bad error for a client to hit -- it _should_ be unreachable, but just in case, we should include anything which could possibly be used to troubleshoot.

src/clients/c/tb_client/context.zig
```diff
@@ -290,317 +391,499 @@ pub fn ContextType(
...
+                };
+                assert(packet.batch_count == decoder.batch_count());
+
+                var it: ?*Packet = packet;
+                while (it) |batched| {
+                    it = batched.batch_next;
```
sentientwaffle, line 833:
```suggestion
                var packet_iterator: ?*Packet = packet;
                while (packet_iterator) |batched| {
                    packet_iterator = batched.batch_next;
```

src/clients/c/tb_client/context.zig
```diff
@@ -290,317 +391,499 @@ pub fn ContextType(
...
             raw_user_data: u128,
-            op: StateMachine.Operation,
+            vsr_operation: vsr.Operation,
             timestamp: u64,
-            reply: []u8,
+            reply: []const u8,
         ) void {
             const user_data: UserData = @bitCast(raw_user_data);
             const self = user_data.self;
             const packet = user_data.packet;
```
sentientwaffle, line 786: Or maybe `root`?

src/clients/c/tb_client/packet.zig
```diff
@@ -55,11 +56,121 @@ pub const Packet = extern struct {
...
-            self.mutex.lock();
-            defer self.mutex.unlock();
-
+        /// Not thread safe, should be called only by the consumer thread.
+        pub fn empty(self: *const SubmissionQueue) bool {
             return self.fifo.count == 0;
         }
     };
+
+    const Phase = enum(u8) {
```
sentientwaffle, line 65: Should this be [`extern enum`](https://ziglang.org/documentation/master/#extern-enum)?

> By default, enums are not guaranteed to be compatible with the C ABI

(Or maybe that sentence doesn't apply when the enum integer type is explicitly declared.)

src/clients/c/tb_client/packet.zig
```diff
@@ -55,11 +56,121 @@ pub const Packet = extern struct {
...
+                assert(packet.batch_count == 0 or packet.batch_next != null);
+                assert(packet.batch_count == 0 or packet.batch_tail != null);
+                assert(packet.next == null);
+                maybe(packet.batch_count > 0);
+                maybe(packet.batch_events_size == 0);
+                maybe(packet.batch_results_size == 0);
+            },
+            .complete => {
+                // The packet pointer isn't available afer completed,
+                // it may be dealocated by the user;
```
sentientwaffle, line 148:
```suggestion
                // it may be deallocated by the user;
```

src/clients/c/tb_client/signal.zig
```diff
@@ -86,7 +87,7 @@ pub const Signal = struct {
     }
 
     fn wait(self: *Signal) void {
-        assert(!self.stop_requested());
+        maybe(self.stop_requested()); // Stop requested while processing the event.
```
sentientwaffle, line 90: Does this comment mean that the stop was requested during `on_signal_fn()`?

src/clients/c/test.zig
```diff
@@ -152,6 +157,33 @@ test "c_client echo" {
     while (repetition < repetitions_max) : (repetition += 1) {
         var completion = Completion{ .pending = concurrency_max };
 
+        const operation = operations[
+            prng.random().intRangeAtMost(
+                usize,
+                0,
+                operations.len - 1,
+            )
```
sentientwaffle, line 165:
```suggestion
            prng.random().uintAtMost(
                usize,
                operations.len - 1,
            )
```

src/clients/java/src/test/java/com/tigerbeetle/IntegrationTest.java
```diff
@@ -1306,32 +1305,47 @@ public void testAsyncTasks() throws Throwable {
             assertTrue(createAccountErrors.getLength() == 0);
 
             final var tasks = new CompletableFuture[TASKS_COUNT];
-            for (int i = 0; i < TASKS_COUNT; i++) {
+            for (int i = 0; i < TASKS_COUNT; i += 1) {
 
                 final var transfers = new TransferBatch(1);
                 transfers.add();
```
sentientwaffle, line 1311: Can this move into the below `else`  statement?

src/clients/java/src/test/java/com/tigerbeetle/IntegrationTest.java
```diff
@@ -2369,7 +2385,7 @@ public Server(final String label) throws IOException, Exception, InterruptedExce
             }
 
             this.process = new ProcessBuilder()
-                    .command(new String[] {exe, "start", "--addresses=0", "--development", tb_file})
```
sentientwaffle, line 2372: Why remove `--development`?

src/vsr.zig
```diff
@@ -266,6 +267,8 @@ pub const Operation = enum(u8) {
     pulse = 4,
     /// The value 5 is is reserved for release-upgrade requests.
     upgrade = 5,
+    /// Batched message that includes multiple state machine operations. See `vsr.batch`.
+    batched = constants.vsr_operations_reserved - 1,
```
sentientwaffle, line 271: Why not `= 6`?

src/testing/vortex/java_driver/src/main/java/Main.java
```diff
@@ -55,24 +68,42 @@ record Driver(Client client, Reader reader, Writer writer) {
   /**
    * Reads the next operation from stdin, runs it, collects the results, and writes them back to
    * stdout.
-   */
-  void next() throws IOException {
+      * @throws ExecutionException
+      * @throws InterruptedException
+      */
```
sentientwaffle, line 73:
```suggestion
   * @throws ExecutionException
   * @throws InterruptedException
   */
```

src/state_machine.zig
```diff
@@ -124,75 +124,13 @@ pub fn StateMachineType(
...
-                    // self.results to "consume" them from subsequent decode() calls.
-                    defer self.results = self.results[demuxed..];
-                    return mem.sliceAsBytes(self.results[0..demuxed]);
-                }
-            };
-        }
-
-        const batch_value_count_max = batch_value_counts_limit(config.message_body_size_max);
-
+        const tree_values_count_limit = tree_values_count_max(config.message_body_size_max);
```
sentientwaffle, line 127: Why "limit" rather than "max"? Typically we use "limit" to indicate a runtime upper bound, vs "max" for a comptime upper bound.

src/state_machine.zig
```diff
@@ -709,59 +771,83 @@ pub fn StateMachineType(
...
-                    if (input.len % event_size != 0) return false;
-                    if (input.len > batch_limit * event_size) return false;
+                    if (input.len % @sizeOf(Event) != 0) return false;
+                    if (input.len > batch_limit * @sizeOf(Event)) return false;
+                    if (input.len > 0 and
+                        !std.mem.isAligned(@intFromPtr(input.ptr), @alignOf(Event)))
+                    {
+                        log.info("Invalid input alignment for {s}", .{
+                            @tagName(operation),
+                        });
```
sentientwaffle, line 827: This seems like it should probably be `log.err` or at least `log.warn`, since (iiuc) it would indicate either a client or a replica bug. (And in case of the former, we should include the client release in the log message.)

Also, please include the replica index prefix (via `self.forest.grid.superblock.replica_index`).

src/state_machine.zig
```diff
@@ -709,59 +771,83 @@ pub fn StateMachineType(
...
             }
 
             return true;
         }
 
-        /// Updates `prepare_timestamp` to the highest timestamp of the response.
-        pub fn prepare(
-            self: *StateMachine,
+        /// Returns the logical time increment (in nanoseconds) for the highest
+        /// timestamp of the response.
```
sentientwaffle, line 837:
```suggestion
        /// Returns the number of nanoseconds that this prepare will consume.
```

The original wording was confusing -- it sounds like a mix of "this is a duration" and "this is an absolute timestamp". Not sure if my version is an improvement. :sweat_smile:

sentientwaffle (commented):
WIP

src/clients/java/src/test/java/com/tigerbeetle/IntegrationTest.java
batiati: I found divergences between Java and Go integrated tests, mainly because Go runs without the `--development` flag.

Anyway, it was an accidental commit, I intend to drop this flag for all integration tests in another PR.

src/clients/c/tb_client/signal.zig
batiati: You independently found the bug fixed at #2689 :smile:

src/clients/c/tb_client/packet.zig
batiati: It's not part of the `Packet.Extern` structure, so never exposed to the C ABI.

src/clients/c/tb_client/context.zig
batiati: Clients should be compatible if new versions advertise a larger message limit (e.g. `message_body_size_max` > 1MiB), or if we want to support running a debug client against a release server.

src/clients/c/tb_client/context.zig
batiati: We might want to improve this algorithm, `packet_enqueue` joins batches of the same operation while `batch_merge` joins multioperations. It keeps operations together without having to find where to insert between two linked list nodes.

batiati:
Ref: #2787

PR #2684, opened 2025-01-29
ci: test building the docs

fabioarnold:
With Docusaurus we tested if the docs would build during the node client test.
This tests building the docs during the smoke test.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,6 +23,7 @@ jobs:
       - run: ./zig/download.sh
       - run: ./zig/zig fmt --check .
       - run: ./zig/zig build check
+      - run: ./zig/zig build scripts -- ci --build-docs
       - run: ./.github/ci/docs_check.sh
 
   test_alpine:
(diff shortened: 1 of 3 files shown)


matklad (approved):

.github/workflows/ci.yml
```diff
@@ -23,6 +23,7 @@ jobs:
       - run: ./zig/download.sh
       - run: ./zig/zig fmt --check .
       - run: ./zig/zig build check
+      - run: ./zig/zig build scripts -- ci --build-docs
       - run: ./.github/ci/docs_check.sh
```
matklad, line 27: Would be so good to move these checks to Zig as well! The link-checker in particular should be easy I think!

fabioarnold: I could add link checking to `file_checker.zig`. Then for all html files check the href attributes if these files exist.

matklad: Yeah, that's exactly what I had in mind! Should be a follow up PR though!

PR #2685, opened 2025-01-29
docs: cleanups

matklad:
First batch of cleanups for docs, best reviewed per commit!


fabioarnold (approved):
I like the `cut_*` helpers.

PR #2686, opened 2025-01-29
docs: add link checker

fabioarnold:
Replaces `remark-validate-links` with a Zig implementation in `file_checker.zig`.
We also check for any insecure "http" links.

### DIFF
--- a/src/docs_website/src/file_checker.zig
+++ b/src/docs_website/src/file_checker.zig
@@ -260,43 +260,30 @@ const Link = struct {
 };
 
 const LinkIterator = struct {
-    line_number: u32 = 0,
-    line_iterator: std.mem.TokenIterator(u8, .scalar),
-    href_iterator: std.mem.TokenIterator(u8, .sequence),
+    line_number: u32 = 1,
+    remaining: []const u8,
 
     const href_prefix = "href=\"";
 
     fn init(html: []const u8) LinkIterator {
-        return .{
-            .line_iterator = std.mem.tokenizeScalar(u8, html, '\n'),
-            .href_iterator = std.mem.tokenizeSequence(u8, "", href_prefix),
-        };
+        return .{ .remaining = html };
     }
 
     fn next(self: *LinkIterator) ?Link {
-        const href = self.next_href() orelse return null;
-        return Link.parse(href);
-    }
+        const index = std.mem.indexOf(u8, self.remaining, href_prefix) orelse
+            return null;
+        const uri_start = index + href_prefix.len;
+        const uri_len = std.mem.indexOfScalar(u8, self.remaining[uri_start..], '"') orelse
+            return null;
+        const uri_end = uri_start + uri_len;
+        const uri_text = self.remaining[uri_start..][0..uri_len];
 
-    fn next_href(self: *LinkIterator) ?[]const u8 {
-        while (true) {
-            if (self.href_iterator.next()) |href| {
-                return std.mem.sliceTo(href, '"');
-            } else {
-                const line = self.next_line() orelse return null;
-
-                self.href_iterator = std.mem.tokenizeSequence(u8, line, href_prefix);
-                if (!std.mem.startsWith(u8, line, href_prefix)) {
-                    _ = self.href_iterator.next(); // Skip.
-                }
-            }
+        for (self.remaining[0..uri_start]) |c| {
+            if (c == '\n') self.line_number += 1;
         }
-    }
+        self.remaining = self.remaining[uri_end..];
 
-    fn next_line(self: *LinkIterator) ?[]const u8 {
-        const line = self.line_iterator.next() orelse return null;
-        self.line_number += 1;
-        return line;
+        return Link.parse(uri_text);
     }
 };
 
--- a/src/docs_website/src/file_checker.zig
+++ b/src/docs_website/src/file_checker.zig
@@ -138,7 +138,11 @@ const http_exceptions = std.StaticStringMap(void).initComptime(.{
 const https_exceptions = std.StaticStringMap(void).initComptime(.{
     .{"https://www.eecg.utoronto.ca/~yuan/papers/failure_analysis_osdi14.pdf"},
     .{"https://pmg.csail.mit.edu/papers/vr-revisited.pdf"},
+    .{"https://pmg.csail.mit.edu/papers/vr.pdf"},
     .{"https://www.infoq.com/presentations/LMAX/"},
+    .{"https://kernel.dk/io_uring.pdf"},
+    .{"https://research.cs.wisc.edu/wind/Publications/latent-sigmetrics07.pdf"},
+    .{"https://security.googleblog.com/2023/06/learnings-from-kctf-vrps-42-linux.html"},
 });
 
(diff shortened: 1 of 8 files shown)


matklad:
🥳

src/docs_website/src/file_checker.zig
```diff
@@ -70,5 +70,62 @@ fn check_files(arena: std.mem.Allocator, path: []const u8) !void {
...
+    .{"http://www.bailis.org/blog/linearizability-versus-serializability/"},
+});
+
+fn check_links(arena: std.mem.Allocator, html_path: []const u8) !void {
+    const dirname = std.fs.path.dirname(html_path).?;
+    const html = try std.fs.cwd().readFileAlloc(arena, html_path, file_size_max);
+
+    var line_it = std.mem.tokenizeScalar(u8, html, '\n');
+    var line_number: usize = 1;
+    while (line_it.next()) |line| {
```
matklad: This is a bit hard to read, as it mixes together link extraction and link validation, both of which are fiddly logic. 

Let's try wringing it as

```zig
var link_iterator = find_links(html);
while (link_iterator.next_link()) |link| {

}
```

where 

```zig
const link = struct {
    line_number: u32,
    uri: std.Uri,
};
```

maybe even extracting logic for validating a single link into a helper function, to keep "loop over links" control plane and "check one link" data plane separate.

src/docs_website/src/file_checker.zig
```diff
@@ -70,5 +70,62 @@ fn check_files(arena: std.mem.Allocator, path: []const u8) !void {
...
+        defer line_number += 1;
+        errdefer log.err("[link checker] error in {s}:{}", .{ html_path, line_number });
+
+        var link_it = std.mem.tokenizeSequence(u8, line, "href=\"");
+        if (!std.mem.startsWith(u8, line, "href=\"")) _ = link_it.next(); // Skip prefix
+        while (link_it.next()) |chunk| {
+            var url = std.mem.sliceTo(chunk, '"');
+
+            // Strip anchor
+            if (std.mem.lastIndexOfScalar(u8, url, '#')) |i| url = url[0..i];
```
matklad: We should also validate that anchors are valid!

src/docs_website/src/file_checker.zig
```diff
@@ -70,5 +70,62 @@ fn check_files(arena: std.mem.Allocator, path: []const u8) !void {
...
+            if (url.len == 0) continue;
+
+            if (std.mem.startsWith(u8, url, "http://")) {
+                if (https_exceptions.has(url)) continue;
+                log.err("Found insecure link: {s}", .{url});
+                return error.InsecureLink;
+            }
+
+            // Skip external link
+            if (std.mem.startsWith(u8, url, "https://")) continue;
```
matklad: Maybe add external link validation under a flag (just using http.CLient to fetch and check that its 200 ok)

matklad: Oh, and we also punctuate comments, so `// Strip external link.`:

https://github.com/tigerbeetle/tigerbeetle/blob/main/docs/TIGER_STYLE.md#naming-things

fabioarnold:
image

Link checker found an external 404 link.

src/docs_website/src/file_checker.zig
```diff
@@ -57,47 +63,243 @@ fn validate_dir(arena: std.mem.Allocator, path: []const u8) !void {
...
+            }
+        }
+    }
+
+    fn next_line(self: *LinkIterator) ?[]const u8 {
+        const line = self.line_iterator.next() orelse return null;
+        self.line_number += 1;
+        return line;
+    }
+};
```
matklad: Feels like this can be simpler if coded directly, instead of juggling two inner iterators?

```zig
const LinkIterator2 = struct {
    newline_count: u32 = 0,
    remaining: []const u8,

    const href_prefix = "href=\"";

    fn init(html: []const u8) LinkIterator {
        return .{
            .newline_count = 0,
            .remaining = html,
        };
    }

    fn next(self: *LinkIterator) ?Link {
        const index = std.mem.indexOf(u8, self.remaining, href_prefix) orelse
            return null;
        const uri_start = index + href_prefix.len;
        const uri_len = std.mem.indexOfScalar(u8, self.remaining[uri_start..], '"') orelse
            return null;
        const uri_end = uri_start + uri_len;
        const uri_text = self.remaining[uri_start..][0..uri_len];

        self.newline_count += std.mem.count(u8, self.remaining[0..uri_start], "\n");
        self.remaining = self.remaining[uri_end..];

        return Link.parse(uri_text);
    }
};
```

src/docs_website/src/file_checker.zig
```diff
@@ -57,47 +63,243 @@ fn validate_dir(arena: std.mem.Allocator, path: []const u8) !void {
...
+        return error.WrongStatusResponse;
+    }
+}
+
+fn check_anchor(context: FileValidationContext, target_path: []const u8, anchor: []const u8) !void {
+    assert(std.mem.endsWith(u8, target_path, ".html"));
+
+    const html = try read_file_cached(context.arena, context.dir, target_path);
+    const needle = try std.mem.concat(context.arena, u8, &.{ "id=\"", anchor, "\"" });
+    if (std.ascii.indexOfIgnoreCase(html, needle) == null) {
```
matklad: Shall we make this a case-sensitive comparison? I'd rather we fix our anchors to be exact than make the validator more lax!

matklad: Shall we make this a case-sensitive comparison? I'd rather we fix our anchors to be exact than make the validator more lax!

src/docs_website/src/file_checker.zig
```diff
@@ -57,47 +63,243 @@ fn validate_dir(arena: std.mem.Allocator, path: []const u8) !void {
...
+    try request.send();
+    try request.finish();
+    try request.wait();
+
+    if (request.response.status != std.http.Status.ok) {
+        return error.WrongStatusResponse;
+    }
+}
+
+fn check_anchor(context: FileValidationContext, target_path: []const u8, anchor: []const u8) !void {
```
matklad: might add some little-endian naming here?

check_links
check_link
check_link_external
check_link_fragment

matklad: might add some little-endian naming here?

check_links
check_link
check_link_external
check_link_fragment

src/docs_website/src/file_checker.zig
```diff
@@ -6,9 +6,15 @@ const assert = std.debug.assert;
 
 pub const file_size_max = 900 * 1024;
 
+// If this is set to true, we check if we get a 200 response for any external links.
+const check_links_external: bool = false;
```
fabioarnold, line 10: @matklad do you want this to be a command line option? Or is editing the source fine?

This is probably too flaky for the CI and takes too long. I guess we want to run this locally on occasion.

matklad: Being in source is fine! We should _not_ run this on CI, as CI is forbidden from being flaky!

But there's value in checking it every once in a while!

matklad (approved):

PR #2689, opened 2025-01-29
[tb_client] Fix incorrect assertion and clarify the API 

batiati:
Fix the assertion too tight, as `signal.wait()` may be called if shutdown is requested during the callback.
Clarify the API regarding the status `running`, `stop_requested`, and `shutdown` to disambiguate.

Related the CI failure:
https://github.com/tigerbeetle/tigerbeetle/actions/runs/12989943988/job/36224019457

### DIFF
--- a/src/clients/c/tb_client/signal.zig
+++ b/src/clients/c/tb_client/signal.zig
@@ -41,7 +41,7 @@ pub const Signal = struct {
 
     pub fn deinit(self: *Signal) void {
         assert(self.event != IO.INVALID_EVENT);
-        assert(self.stop_requested());
+        assert(self.status() == .shutdown);
 
         self.io.close_event(self.event);
         self.* = undefined;
@@ -50,15 +50,28 @@ pub const Signal = struct {
     /// Stops the current event.
     /// Further calls to "notify" will not fire the signal.
     /// Safe to call from multiple threads.
-    pub fn stop(self: *Signal) void {
+    pub fn shutdown(self: *Signal) void {
         const listening = self.listening.swap(false, .release);
         if (listening) {
             self.notify();
         }
     }
 
-    pub fn stop_requested(self: *const Signal) bool {
-        return !self.listening.load(.acquire);
+    pub fn status(self: *const Signal) enum {
+        running,
+        stop_requested,
+        shutdown,
+    } {
+        return switch (self.state.load(.acquire)) {
+            .shutdown => .shutdown,
+            .running,
+            .waiting,
+            .notified,
+            => if (self.listening.load(.acquire))
+                .running
+            else
+                .stop_requested,
+        };
     }
 
     /// Schedules the on_signal callback to be invoked on the IO thread.
@@ -86,7 +99,9 @@ pub const Signal = struct {
     }
 
     fn wait(self: *Signal) void {
-        assert(!self.stop_requested());
+        // It is not guaranteed to be `running` here, as another caller might have requested
+        // a stop during the callback.
+        assert(self.status() != .shutdown);
 
         const state = self.state.swap(.waiting, .acquire);
         self.io.event_listen(self.event, &self.completion, on_event);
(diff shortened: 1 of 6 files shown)


src/clients/c/tb_client/signal.zig
```diff
@@ -50,15 +50,28 @@ pub const Signal = struct {
...
             self.notify();
         }
     }
 
-    pub fn stop_requested(self: *const Signal) bool {
-        return !self.listening.load(.acquire);
+    pub fn status(self: *const Signal) enum {
+        running,
+        stop_requested,
+        shutdown,
```
matklad: Maybe `stop_requested` and `stopped` ? To make them more clearly related to each other.

src/clients/c/tb_client/signal.zig
```diff
@@ -86,7 +99,9 @@ pub const Signal = struct {
     }
 
     fn wait(self: *Signal) void {
-        assert(!self.stop_requested());
+        // It is not guaranteed to be `running` here, as another caller might have requested
+        // a stop during the callback.
+        assert(self.status() != .shutdown);
```
matklad: Let's cover this scenario by a test. Rather than adding a specific test here, I think we can change our `test "signal"` to be a fuzz-test of sorts --- run it ten times, where sometimes we randomly stop the signal?

src/clients/c/tb_client/signal.zig
matklad: And can we expand the doc comment for Signal to document the actual lifecycle?

src/clients/c/tb_client/signal_fuzz.zig
```diff
@@ -0,0 +1,135 @@
...
+}
+
+fn notify(context: *Context) void {
+    const delay = 10 * std.time.ns_per_us;
+    assert(std.Thread.getCurrentId() != context.main_thread_id);
+    while (context.signal.status() != .stopped) {
+        std.time.sleep(delay);
+
+        // Chance to stop the signal between notifications.
+        if (chance(context.random, stop_chance_percentage)) {
```
matklad: This is a data race!

Let's maybe not share the entire `*Context`, but rather pass only those parts that `notify` needs, to make it obvious at the call-site that we aren't sharing anything non-thread-safer.

src/clients/c/tb_client/signal_fuzz.zig
```diff
@@ -0,0 +1,135 @@
...
+    var prng = std.rand.DefaultPrng.init(args.seed);
+    const events_max = args.events_max orelse 10_000;
+
+    var context: Context = .{
+        .io = try IO.init(32, 0),
+        .main_thread_id = std.Thread.getCurrentId(),
+        .signal = undefined,
+        .random = prng.random(),
+    };
+    defer context.io.deinit();
```
matklad: This should be a relatively fast fuzzer, as it doesn't make much sense to do more than a dozen events within a single run. So let's add an overall-loop here:

```
for (0..iterations_total) {
    const io = try IO.init(32, 0)
    for (0..notifies_count) {
    }
}
```

That way, we'll test more `cancel` calls!

src/clients/c/tb_client/signal_fuzz.zig
```diff
@@ -0,0 +1,135 @@
...
+    while (context.signal.status() != .stopped) {
+        if (context.running_count >= events_max) {
+            if (context.stop_request.cmpxchgStrong(
+                .none,
+                .from_io_thread,
+                .acquire,
+                .monotonic,
+            ) == null) context.signal.stop();
+        }
+        try context.io.tick();
```
matklad: Let's also do, with some probability, `context.io.cancel_all` here?

batiati: It's not supported: After calling `IO.cancel_all`, it's safe to close the `event`, but the higher-level `Signal` never receives the completion and remains stuck in `.stop_requested`.

matklad (approved):

PR #2691, opened 2025-01-30
Update link to get_account_transfers.md

sekerez:
Currently returns a 404, see screenshot below (look at the url):
Screenshot 2025-01-30 at 9 52 45 AM


fabioarnold:
Thank you for your PR! We're in the process of adding a link checker that should catch these types of mistakes: https://github.com/tigerbeetle/tigerbeetle/pull/2686
Once it is merged, this particular issue should also be fixed. :)

Unfortunately, if I would merge your change as is it would break doc navigation on GitHub.

fabioarnold:
The link checker is now merged and the broken link is fixed. Thanks again for bringing this to our attention.

PR #2692, opened 2025-01-30
vsr: send withheld prepare_oks after syncing tables

chaitanyabhandari:
Resolves the failed seed `./zig/zig build -Drelease vopr -Dvopr-state-machine=testing -- 2381528437601413660` on e1d59062c0846a0e2f1697b4a60aa8aa993fb42b.

```C
[info] (cluster): no liveness, final cluster state (core=10000011):
[info] (cluster):  0   / .             6V  99/107/107C  15:109Jo  0/_0J!  15:109Wo <_24:_83> v1:1   126Ga  0G!   0G?   2/4Pp  0/3Rq
[info] (cluster):  1   \  .            6V  99/107/107C  78:109Jo  0/_0J!  78:109Wo <__0:__0> v1:1   126Ga  0G!   0G?
[info] (cluster):  2   /   .           5V  99/109/109C  78:109Jo  0/_0J!  78:109Wo <__0:__0> v1:1   108Ga  0G!   3G?   0/4Pp  0/3Rq
```
Replicas that haven't yet synced all the grid content only prepare_ok ops up till `op_checkpoint_trigger + pipeline_prepare_queue_max`. This ensures they don't falsely contribute to the durability of a _non_durable_ checkpoint they've synced to.
https://github.com/tigerbeetle/tigerbeetle/blob/e1d59062c0846a0e2f1697b4a60aa8aa993fb42b/src/vsr/replica.zig#L6362-L6372

Currently, we don't have logic to send these withheld prepare_oks past `op_checkpoint_trigger + pipeline_prepare_queue_max` when a replica finishes syncing all tables. As a result, R0 (a replica that finishes syncing its content after it steps up as primary) isn't able to advance its commit_min & commit_max to op=109, even though the ops 108 → 109 are durable on both R0 and R1.  This PR adds logic to send these withheld prepare_oks, similar to `send_prepare_oks_after_checkpoint`.
https://github.com/tigerbeetle/tigerbeetle/blob/e1d59062c0846a0e2f1697b4a60aa8aa993fb42b/src/vsr/replica.zig#L4657-L4658

### DIFF
--- a/src/vsr/replica.zig
+++ b/src/vsr/replica.zig
@@ -7891,6 +7891,20 @@ pub fn ReplicaType(
             self.send_prepare_oks_from(self.commit_max + 1);
         }
 
+        fn send_prepare_oks_after_syncing_tables(self: *Replica) void {
+            assert(self.status == .normal or self.status == .view_change or
+                self.status == .recovering_head);
+            assert(self.syncing == .idle);
+            assert(self.sync_tables == null);
+
+            const op_checkpoint_trigger =
+                vsr.Checkpoint.trigger_for_checkpoint(self.op_checkpoint()).?;
+            self.send_prepare_oks_from(@max(
+                self.commit_max + 1,
+                op_checkpoint_trigger + constants.pipeline_prepare_queue_max + 1,
+            ));
+        }
+
         fn send_prepare_oks_after_checkpoint(self: *Replica) void {
             assert(self.status == .normal or self.status == .view_change or
                 (self.status == .recovering and self.solo()));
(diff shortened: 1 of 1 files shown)


sentientwaffle (approved):
Great work!

I was initially surprised that we could call `send_prepare_oks_after_syncing_tables` from `status == .recovering_head or status == .view_change`, but `send_prepare_ok()` is just a noop in those cases, so it is fine.

chaitanyabhandari:
Thanks DJ! 

> send_prepare_ok() is just a noop in those cases, so it is fine.

Correct! Also, I initially thought we shouldn't be able to get to this branch while we were `recovering_head` at all, but since `state_machine_open_callback` → `sync_content` is called on startup (for replicas with non-zero `sync_op_max`), `recovering_head` replicas can hit it too.

PR #2698, opened 2025-02-03
Tracing: add metrics and statsd support

cb22:
### Overview
This PR adds supports for timing and metrics to the existing tracing infrastructure. The goal is to expose metrics at runtime to give operators visibility into the state of their cluster as well as to help with diagnosing any issues that arise.

Any trace that's recording using `trace.start()` now also gets aggregated internally, and periodically emitted via statsd if `--statsd=<host>:<port>` is specified in the CLI flags. 

Simple gauge metrics have also been added; these are aggregated on last-write-wins. Additionally, tracing (and timing) events are now fully typed (defined in `trace/event.zig`) with no more `anytype` payload. 

### Usage
```
$ ./tigerbeetle start --experimental --statsd=127.0.0.1:8125
```

```
$ nc -l -u 8125 # Pretend to be statsd.

# Some timings:
tigerbeetle.replica_commit_us.min:2735|g|#stage:prefetch
tigerbeetle.replica_commit_us.avg:2735|g|#stage:prefetch
tigerbeetle.replica_commit_us.max:2735|g|#stage:prefetch
tigerbeetle.replica_commit_us.sum:2735|c|#stage:prefetch
tigerbeetle.replica_commit_us.count:1|c|#stage:prefetch
...
# Some metrics:
tigerbeetle.table_count_visible:1|g|#tree:Transfer.debit_account_id,level:0
tigerbeetle.table_count_visible:0|g|#tree:Transfer.debit_account_id,level:1
tigerbeetle.table_count_visible:0|g|#tree:Transfer.debit_account_id,level:2
...
# Meta (:
tigerbeetle.metrics_emit_us.min:154|g|#
tigerbeetle.metrics_emit_us.avg:154|g|#
tigerbeetle.metrics_emit_us.max:154|g|#
tigerbeetle.metrics_emit_us.sum:154|c|#
tigerbeetle.metrics_emit_us.count:1|c|#
```

### Tracing vs Timing
The original (and still existing) `Event` type is further split into `EventTiming` and `EventTracing`: while tracing cares about cardinality of concurrency, timing cares about cardinality of aggregation.

The aggregations for timing and metrics are statically allocated across their cardinality, in the same way that tracing did originally.


src/state_machine.zig
```diff
@@ -37,6 +37,46 @@ const CreateTransferResult = tb.CreateTransferResult;
...
+
+    pub const TransferPending = .{
+        .timestamp = 20,
+        .status = 21,
+    };
+
+    pub const AccountBalance = .{
+        .timestamp = 22,
+    };
+};
```
cb22, line 78: Pulled this out of the StateMachineType, so it could be easily imported inside the Event code, as well as renamed `accounts` to `Account` to be more consistent with how we name things elsewhere...

src/tigerbeetle/benchmark_load.zig
```diff
@@ -125,17 +124,6 @@ pub fn main(
...
-    defer if (statsd_opt) |*statsd| statsd.deinit(allocator);
-
-    if (cli_args.statsd) {
-        statsd_opt = try StatsD.init(
-            allocator,
-            &io,
-            std.net.Address.parseIp4("127.0.0.1", 8125) catch unreachable,
-        );
-    }
-
```
cb22, line 138: The benchmark no longer supports statsd. It was only useful for us for a limited period of time - if we find we need it, we can add it back later with the new implementation.

src/trace.zig
```diff
@@ -455,15 +295,15 @@ test "trace json" {
...
     defer trace.deinit(std.testing.allocator);
 
-    trace.start(.replica_commit, .{ .foo = 123 });
-    trace.start(.compact_beat, .{});
-    trace.stop(.compact_beat, .{});
-    trace.stop(.replica_commit, .{ .bar = 456 });
+    trace.start(.{ .replica_commit = .{ .stage = .idle, .op = 123 } });
+    trace.start(.{ .compact_beat = .{ .tree = @enumFromInt(1), .level_b = 1 } });
+    trace.stop(.{ .compact_beat = .{ .tree = @enumFromInt(1), .level_b = 1 } });
+    trace.stop(.{ .replica_commit = .{ .stage = .idle, .op = 456 } });
```
cb22, line 396: Payloads have to be real now :smile:

src/trace/event.zig
```diff
@@ -0,0 +1,497 @@
+const std = @import("std");
+const assert = std.debug.assert;
+
+const constants = @import("../constants.zig");
+
+const CommitStageTag = @import("../vsr/replica.zig").CommitStage.CommitStageTag;
+
+const TreeEnum = tree_enum: {
```
cb22, line 8: While something resembling this does exist in the forest / grooves, pulling it out looked incredibly tricky vs using the tree_ids directly.

src/vsr/replica.zig
```diff
@@ -70,6 +70,41 @@ pub const ReplicaEvent = union(enum) {
     client_evicted: u128,
 };
 
+pub const CommitStage = union(enum) {
```
cb22, line 73: (same reason pulling this out as for tree_ids in StateMachine)

src/trace/event.zig
```diff
@@ -0,0 +1,497 @@
...
+
+    break :tree_enum @Type(.{ .Enum = .{
+        .tag_type = u64,
+        .fields = tree_fields,
+        .decls = &.{},
+        .is_exhaustive = true,
+    } });
+};
+
+fn enum_max(EnumOrUnion: type) u8 {
```
sentientwaffle, line 31: TIL `comptime` isn't required here!

src/trace/event.zig
```diff
@@ -0,0 +1,497 @@
...
+    const type_info = @typeInfo(EnumOrUnion);
+    assert(type_info == .Enum or type_info == .Union);
+
+    const Enum_ = if (type_info == .Enum)
```
sentientwaffle:
```suggestion
    const Enum = if (type_info == .Enum)
```

`Enum` is not a reserved word.

src/trace/event.zig
sentientwaffle, line 31: Kind of confusing that this actually returns `max + 1` not `max`. But I guess that is covered by the TODO on `stack()`.

src/trace/event.zig
```diff
@@ -0,0 +1,497 @@
...
+/// Base {Timing,Tracing} Event. This is further split up into two different Events that share the
+/// same tag: ones for timing and ones for tracing.
+///
+/// This is because there's a difference between tracing and aggregate timing. When doing tracing,
+/// the code needs to worry about the static allocation required for _concurrent_ traces. That is,
+/// there might be multiple `scan_tree`s, with different `index`es happening at once.
+///
+/// When timing, this is flipped on its head: the timing code doesn't need space for concurrency
+/// because it is called once, when an event has finished, and internally aggregates. The
+/// aggregiation is needed because there can be an unknown number of calls between flush intervals,
```
sentientwaffle:
```suggestion
/// aggregation is needed because there can be an unknown number of calls between flush intervals,
```

src/trace.zig
sentientwaffle, line 1: Update examples in module header comment.

src/trace/statsd.zig
```diff
@@ -0,0 +1,574 @@
...
+const RingBufferType = stdx.RingBufferType;
+
+const EventMetric = @import("event.zig").EventMetric;
+const EventMetricAggregate = @import("event.zig").EventMetricAggregate;
+const EventTiming = @import("event.zig").EventTiming;
+const EventTimingAggregate = @import("event.zig").EventTimingAggregate;
+
+const log = std.log.scoped(.statsd);
+
+/// A resonable value to keep the total length of the packet under a single MTU, for a local
```
sentientwaffle:
```suggestion
/// A reasonable value to keep the total length of the packet under a single MTU, for a local
```

src/trace/statsd.zig
```diff
@@ -0,0 +1,574 @@
...
+/// network.
+///
+/// https://github.com/statsd/statsd/blob/master/docs/metric_types.md#multi-metric-packets
+const packet_size_max = 1400;
+
+/// No single metric may be larger than this value. If it is, it'll be dropped with an error
+/// message. Since this is calculated at comptime, that means there's a bug in the calculation
+/// logic.
+const statsd_line_size_max = blk: {
```
sentientwaffle: For large blocks I prefer to use a more descriptive name, which gets repeated at the end when we "return":
```suggestion
const statsd_line_size_max = line_size_max: {
```

src/trace/statsd.zig
```diff
@@ -0,0 +1,574 @@
...
+    var len_max: u32 = 0;
+    while (iterator.next()) |value| {
+        // Getting a .none here means that either an Aggregate is null (impossible since they are
+        // set above) or, that the maximum buffer length of a single packet has been exceeded.
+        assert(value == .some);
+
+        len_max = @max(value.some.len, len_max);
+    }
+
+    break :blk len_max;
```
sentientwaffle:
```suggestion
    var size_max: u32 = 0;
    while (iterator.next()) |value| {
        // Getting a .none here means that either an Aggregate is null (impossible since they are
        // set above) or, that the maximum buffer length of a single packet has been exceeded.
        assert(value == .some);

        size_max = @max(value.some.len, size_max);
    }

    break :blk size_max;
```

src/trace/statsd.zig
```diff
@@ -0,0 +1,574 @@
...
+/// network.
+///
+/// https://github.com/statsd/statsd/blob/master/docs/metric_types.md#multi-metric-packets
+const packet_size_max = 1400;
+
+/// No single metric may be larger than this value. If it is, it'll be dropped with an error
+/// message. Since this is calculated at comptime, that means there's a bug in the calculation
+/// logic.
+const statsd_line_size_max = blk: {
+    var buffer: [packet_size_max]u8 = undefined;
```
sentientwaffle: Declare this later, closer to where it is actually needed.

src/trace/statsd.zig
```diff
@@ -0,0 +1,574 @@
...
+
+            return .{
+                .some = std.fmt.bufPrint(
+                    self.buffer,
+                    // TODO: Support counters.
+                    "tigerbeetle.{[name]s}:{[value]}|g|#{[tags]s}\n",
+                    .{ .name = field_name, .value = value, .tags = event_metric_tag_formatter },
+                ) catch {
+                    if (!@inComptime()) {
+                        log.err("metric line for {s} exceeeds buffer size", .{field_name});
```
sentientwaffle:
```suggestion
                        log.err("metric line for {s} exceeds buffer size", .{field_name});
```

src/trace/statsd.zig
```diff
@@ -0,0 +1,574 @@
...
+                    .{
+                        .name = field_name,
+                        .aggregation = @tagName(self.index_aggregation),
+                        .value = value,
+                        .statsd_type = statsd_type,
+                        .tags = tag_formatter,
+                    },
+                ) catch {
+                    if (!@inComptime()) {
+                        log.err("metric line for {s} exceeeds buffer size", .{field_name});
```
sentientwaffle:
```suggestion
                        log.err("metric line for {s} exceeds buffer size", .{field_name});
```

src/trace/statsd.zig
```diff
@@ -0,0 +1,574 @@
...
+    assert(statsd_line_size_max <= packet_size_max);
+    assert(messages_per_packet > 0);
+}
+
+/// This implementation emits on an open-loop: on the emit interval, it fires off up to
+/// max_packet_count UDP packets, without waiting for completions.
+///
+/// The emit interval needs to be large enough that the kernel will have finished processing them
+/// before emitting again. If not, an error will be logged.
+const max_packet_count = stdx.div_ceil(
```
sentientwaffle:
```suggestion
const packet_count_max = stdx.div_ceil(
```

src/trace/statsd.zig
```diff
@@ -0,0 +1,574 @@
...
+        // set above) or, that the maximum buffer length of a single packet has been exceeded.
+        assert(value == .some);
+
+        len_max = @max(value.some.len, len_max);
+    }
+
+    break :blk len_max;
+};
+
+const messages_per_packet = @divFloor(packet_size_max, statsd_line_size_max);
```
sentientwaffle: `packet_messages_max`?

src/trace/metric.zig
```diff
@@ -0,0 +1,130 @@
...
+
+        self.* = undefined;
+    }
+
+    /// Gauges work on a last-set wins. Multiple calls to .gauge() followed by an emit will result
+    /// in the last value being submitted.
+    pub fn gauge(self: *Metrics, event_metric: EventMetric, value: u64) void {
+        const timing_stack = event_metric.stack();
+        self.events_metric[timing_stack] = .{
+            .event = event_metric,
```
sentientwaffle: :thinking: Why do we need to include `event_metric` here -- can it be inferred from the index (i.e. the stack)?

src/trace/metric.zig
```diff
@@ -0,0 +1,130 @@
...
+            .value = value,
+        };
+    }
+
+    // Timing works by storing the min, max, sum and count of each value provided. The avg is
+    // calculated from sum and count at emit time.
+    //
+    // When these are emitted upstream (via statsd, currently), upstream must apply different
+    // aggregiations:
```
sentientwaffle:
```suggestion
    // aggregations:
```

src/trace/metric.zig
```diff
@@ -0,0 +1,130 @@
...
+        self.statsd.deinit(allocator);
+        allocator.free(self.events_timing);
+        allocator.free(self.events_metric);
+
+        self.* = undefined;
+    }
+
+    /// Gauges work on a last-set wins. Multiple calls to .gauge() followed by an emit will result
+    /// in the last value being submitted.
+    pub fn gauge(self: *Metrics, event_metric: EventMetric, value: u64) void {
```
sentientwaffle: Proxy this function via `Tracer` so that `manifest.zig` doesn't need to dig for it.

And maybe rename this "internal" version of `gauge()` to `record_gauge()` and `timing()` to `record_timing()`, to group them together as they are similar?

src/trace/statsd.zig
```diff
@@ -0,0 +1,574 @@
...
+    if (@typeInfo(StructOrVoid) == .Void) return {};
+
+    assert(@typeInfo(StructOrVoid) == .Struct);
+    const Struct = StructOrVoid;
+
+    var output: Struct = undefined;
+
+    for (std.meta.fields(Struct)) |field| {
+        const type_info = @typeInfo(field.type);
+        assert(type_info == .Int or type_info == .Enum);
```
sentientwaffle, line 401: If we used a signed int, then `minInt` would be the longest instead:
```suggestion
        assert(type_info == .Int or type_info == .Enum);
        assert(type_info != .Int or type_info.Int.signedness == .unsigned);
```

src/vsr/replica.zig
```diff
@@ -1249,6 +1287,11 @@ pub fn ReplicaType(
                     .id = replica_index,
                     .after = 500,
                 },
+                .trace_emit_timeout = Timeout{
+                    .name = "trace_emit_timeout",
+                    .id = replica_index,
+                    .after = 100,
+                },
```
sentientwaffle: Wouldn't `metrics_emit_timeout` be more accurate? Traces are emitted (logged) immediately.

src/vsr/replica.zig
```diff
@@ -821,6 +857,8 @@ pub fn ReplicaType(
             // can repair grid blocks if necessary:
             self.grid.open(grid_open_callback);
             self.invariants();
+
+            self.trace_emit_timeout.start();
```
sentientwaffle, line 864: Move this to the `transition_to_...` functions -- even for timeouts that run in all statuses, `timeout.start()` lives there. (See `ping_timeout` for example).

src/tigerbeetle/cli.zig
sentientwaffle, line 1: What do you think of:
- Add `--statsd` back to benchmark -- but this time, pass it through to the replica (like we do for `--trace`).
- Then in `integration_tests.zig`, in the `"benchmark/inspect smoke"`, pass `--statsd` with some random port. No need to listen on the port, I just want to have some test which actually includes us sending UDP packets. (Maybe pass `--trace=some_file.json` too, and delete the file afterward?)

src/tigerbeetle/cli.zig
```diff
@@ -953,30 +961,34 @@ fn parse_args_multiversion(multiversion: CLIArgs.Multiversion) Command.Multivers
 }
 
 /// Parse and allocate the addresses returning a slice into that array.
-fn parse_addresses(raw_addresses: []const u8) Command.Addresses {
-    var result: Command.Addresses = .{};
+fn parse_addresses(
+    raw_addresses: []const u8,
+    comptime flag: []const u8,
+    comptime Result: type,
+) Result {
```
sentientwaffle:
```suggestion
) Result {
    comptime assert(std.mem.startsWith(u8, flag, "--"));
```

src/tigerbeetle/cli.zig
```diff
@@ -953,30 +961,34 @@ fn parse_args_multiversion(multiversion: CLIArgs.Multiversion) Command.Multivers
 }
 
 /// Parse and allocate the addresses returning a slice into that array.
-fn parse_addresses(raw_addresses: []const u8) Command.Addresses {
-    var result: Command.Addresses = .{};
+fn parse_addresses(
+    raw_addresses: []const u8,
+    comptime flag: []const u8,
+    comptime Result: type,
```
sentientwaffle: Maybe call this `BoundedArray` so that it is clear what sort of type this is? Afaik there's no way to assert "this type was constructed by `stdx.BoundedArrayType` which would be even better.

src/trace/statsd.zig
```diff
@@ -0,0 +1,574 @@
...
+            const value = switch (self.index_aggregation) {
+                .min => values.duration_min_us,
+                .avg => @divFloor(values.duration_sum_us, values.count),
+                .max => values.duration_max_us,
+                .sum => values.duration_sum_us,
+                .count => values.count,
+                .sentinel => unreachable,
+            };
+
+            // Emit count and sum as counter metrics, and the rest as gagues. This ensures that the
```
sentientwaffle:
```suggestion
            // Emit count and sum as counter metrics, and the rest as gauges. This ensures that the
```

src/trace/statsd.zig
```diff
@@ -0,0 +1,574 @@
...
+            };
+            const event_metric_tag_formatter = TagFormatter{
+                .event = event_metric.event,
+            };
+
+            return .{
+                .some = std.fmt.bufPrint(
+                    self.buffer,
+                    // TODO: Support counters.
+                    "tigerbeetle.{[name]s}:{[value]}|g|#{[tags]s}\n",
```
sentientwaffle: I'm assuming we want to avoid an abbreviated prefix like `tb.`? I only consider it since it would let us fit more metrics in a packet.

sentientwaffle: If someone operates multiple TB clusters, and they accidentally point them both to the same statsd server, would the metrics be jumbled together? How can they avoid this?

I guess we could include `cluster` as a tag, though that is quite verbose. We could add a CLI flag to let the operator pick the metric prefix. But both of these seem like hacks. :disappointed:

sentientwaffle: ~Re:tags -- does any statsd server besides datadog's support this extension to the protocol?~
Ah, the [prometheus statsd exporter](https://github.com/prometheus/statsd_exporter) also supports it. Along with several other [competing incompatible tagging extensions](https://xkcd.com/927/) to statsd protocols. That's still less than ideal, but probably as good as we can do without just switching to prometheus.

---

Also, why not include a replica index tag in every metric?

src/trace/statsd.zig
```diff
@@ -0,0 +1,574 @@
...
+                },
+            };
+        }
+    };
+
+    const TimingsIterator = struct {
+        const Aggregation = enum { min, avg, max, sum, count, sentinel };
+        const TagFormatter = EventStatsdTagFormatterType(EventTiming);
+
+        events_timing: []?EventTimingAggregate,
```
sentientwaffle:
```suggestion
        events_timing: []const ?EventTimingAggregate,
```

src/trace/statsd.zig
```diff
@@ -0,0 +1,574 @@
...
+    }
+};
+
+const Iterator = struct {
+    const Output = ?union(enum) { none, some: []const u8 };
+    const MetricsIterator = struct {
+        const TagFormatter = EventStatsdTagFormatterType(EventMetric);
+
+        events_metric: []?EventMetricAggregate,
+        events_metric_emitted: []?EventMetricAggregate,
```
sentientwaffle:
```suggestion
        events_metric: []const ?EventMetricAggregate,
        events_metric_emitted: []const ?EventMetricAggregate,
```

src/trace/statsd.zig
```diff
@@ -0,0 +1,574 @@
...
+        allocator.free(self.events_metric_emitted);
+        allocator.free(self.buffer_completions_buffer);
+
+        self.* = undefined;
+    }
+
+    pub fn emit(
+        self: *StatsD,
+        events_metric: []?EventMetricAggregate,
+        events_timing: []?EventTimingAggregate,
```
sentientwaffle:
```suggestion
        events_metric: []const ?EventMetricAggregate,
        events_timing: []const ?EventTimingAggregate,
```

src/trace/statsd.zig
```diff
@@ -0,0 +1,574 @@
...
+
+        events_timing: []?EventTimingAggregate,
+
+        buffer: []u8 = undefined,
+        index: usize = 0,
+        index_aggregation: Aggregation = .min,
+
+        pub fn next(self: *TimingsIterator) Output {
+            defer {
+                self.index_aggregation = @enumFromInt(@intFromEnum(self.index_aggregation) + 1);
```
sentientwaffle: I'm surprised zig's `std.enums` doesn't have an iterator like this built-in!
The closest thing is `EnumSet.initFull().iterator()` but I think your approach is better.

src/trace/statsd.zig
```diff
@@ -0,0 +1,574 @@
...
+        }
+    };
+
+    const TimingsIterator = struct {
+        const Aggregation = enum { min, avg, max, sum, count, sentinel };
+        const TagFormatter = EventStatsdTagFormatterType(EventTiming);
+
+        events_timing: []?EventTimingAggregate,
+
+        buffer: []u8 = undefined,
```
sentientwaffle:
```suggestion
        buffer: []u8,
```

src/trace/statsd.zig
```diff
@@ -0,0 +1,574 @@
...
+                        &buffer_completion.completion,
+                        buffer_completion_written,
+                    );
+                },
+            }
+        } else {
+            self.buffer_completions.push_assume_capacity(buffer_completion);
+        }
+
+        @memcpy(self.events_metric_emitted, events_metric);
```
sentientwaffle: `copy_disjoint` + `.exact`?

src/trace/statsd.zig
```diff
@@ -0,0 +1,574 @@
...
+        allocator.free(self.buffer_completions_buffer);
+
+        self.* = undefined;
+    }
+
+    pub fn emit(
+        self: *StatsD,
+        events_metric: []?EventMetricAggregate,
+        events_timing: []?EventTimingAggregate,
+    ) void {
```
sentientwaffle:
```suggestion
    ) void {
            assert(events_metric.len == self.events_metric_emitted.len);

```

src/trace/statsd.zig
```diff
@@ -0,0 +1,574 @@
...
+                .events_metric_emitted = self.events_metric_emitted,
+                .buffer = &buffer,
+            },
+            .timings = .{
+                .events_timing = events_timing,
+                .buffer = &buffer,
+            },
+        };
+
+        var buffer_completion = self.buffer_completions.pop() orelse {
```
sentientwaffle: `assert(self.implementation == .udp);` (since `.log` synchronously restores the buffer completions).

src/trace/statsd.zig
```diff
@@ -0,0 +1,574 @@
...
+        log,
+    },
+
+    buffer_completions: BufferCompletionRing,
+    buffer_completions_buffer: []BufferCompletion,
+
+    events_metric_emitted: []?EventMetricAggregate,
+
+    /// Creates a statsd instance, which will send UDP packets via the IO instance provided.
+    pub fn init(allocator: std.mem.Allocator, io: *IO, address: std.net.Address) !StatsD {
```
sentientwaffle:
```suggestion
    pub fn init_udp(allocator: std.mem.Allocator, io: *IO, address: std.net.Address) !StatsD {
```

(To mirror `init_log()`.)

src/trace/metric.zig
sentientwaffle, line 1: Naming suggestions:

I'd suggest that instead of "timings" we say e.g. "histogram" or "distribution" -- we might want to output other values, e.g. distribution of block size, or distribution of prepare size.

Also, is confusing to me that when we say "metric" we mean "metrics that aren't timings/histograms".
I don't have a suggestion yet that would also work well for counters, though.

src/vsr/replica.zig
sentientwaffle: Will leave it as-is, since we are calling `trace.emit()` as `Tracer` is the container for all metrics.

src/trace/metric.zig
sentientwaffle: I will punt this for a future PR.

src/trace/metric.zig
sentientwaffle: Answer: In principle we could, but this is simpler.

src/vsr/replica.zig
sentientwaffle: Ah, that doesn't work, b/c we want to run this timeout even during status=recovering.

src/vsr/replica.zig
```diff
@@ -70,6 +70,41 @@ pub const ReplicaEvent = union(enum) {
     client_evicted: u128,
 };
 
+pub const CommitStage = union(enum) {
+    pub const CommitStageTag = std.meta.Tag(CommitStage);
```
matklad: Maybe just `Tag` so that the call-site reads `CommitStage.Tag`?

src/trace/event.zig
```diff
@@ -0,0 +1,496 @@
...
+                    }
+                }
+
+                return @unionInit(EventType, @tagName(tag), target_payload);
+            },
+        };
+    }
+};
+
+pub const EventTiming = union(Event.EventTag) {
```
matklad: Here as well, `Event.EventTag` looks funny, consider just `Event.Tag`

src/trace/event.zig
```diff
@@ -0,0 +1,496 @@
...
+
+    // Unlike with EventTracing, which neatly organizes related events underneath one another, the
+    // order here does not matter.
+    //
+    // TODO: This could be computed automatically, at least in the cases where all, or all except
+    // one of the values are enums.
+    //
+    // TODO: The enum logic is a bit wasteful. It considers the max value, so for enums that don't
+    // start at 0 (eg, the trees) it's not optimal.
+    pub fn stack(event: *const EventTiming) u32 {
```
matklad: I accidentally nerd-sniped myself here! One neat thing we can do, to double-check the logic, is to just generate _all_ events and check that there stacks don't collide:

0ab9c7415a878b512bfbb229f69078d71dc67f59 (updated with better names)

matklad: Unrelated, but is the stack still the right name here? Maybe `.index` or `.ordinal`, a we don't have real stacks for metrics?

src/trace.zig
```diff
@@ -284,63 +194,73 @@ pub const Tracer = struct {
             "\"ph\":\"{[event]c}\"," ++
             "\"ts\":{[timestamp]}," ++
             "\"cat\":\"{[category]s}\"," ++
-            "\"name\":\"{[name]s}{[data]}\"," ++
+            "\"name\":\"{[category]s} {[event_tracing]s} {[event_timing]}\"," ++
```
matklad:
```suggestion
            "\"name\":\"{[category]s} {[event_tracing]} {[event_timing]}\"," ++
```

src/trace/event.zig
```diff
@@ -0,0 +1,547 @@
...
+
+    /// Convert the base event to an EventTiming or EventMetric.
+    pub fn as(event: *const Event, EventType: type) EventType {
+        return switch (event.*) {
+            inline else => |source_payload, tag| {
+                const TargetPayload = std.meta.fieldInfo(EventType, tag).type;
+                const target_payload_info = @typeInfo(TargetPayload);
+                assert(target_payload_info == .Void or target_payload_info == .Struct);
+
+                var target_payload: TargetPayload = undefined;
```
matklad: Not sure if `undefined` is a valid value of `void` type in Zig. Maybe write this as

```
const target_payload = switch(@typeInfo(TargetPlaylod)) {
    .Void => {}, 
    .Struct => blk: {
        ...
    }
}
```

src/trace/event.zig
```diff
@@ -0,0 +1,547 @@
...
+        comptime fmt: []const u8,
+        options: std.fmt.FormatOptions,
+        writer: anytype,
+    ) !void {
+        _ = fmt;
+        _ = options;
+
+        switch (event.*) {
+            inline else => |data| {
+                try writer.print("{}", .{event_equals_format(data)});
```
matklad: Hm, it feels like there's one extra level of indirection here? This constructs `EventEqualsFormatterType` type and then invokes its `format` function, but we _already_ inside format. Can we remove `EventEqualsFormatterType` type and just call the formatting function, parametrized by data type, directly?


```diff
diff --git a/src/trace/event.zig b/src/trace/event.zig
index f1fc05e92..2a99df799 100644
--- a/src/trace/event.zig
+++ b/src/trace/event.zig
@@ -248,7 +248,7 @@ pub const EventTiming = union(Event.Tag) {
 
         switch (event.*) {
             inline else => |data| {
-                try writer.print("{}", .{event_equals_format(data)});
+                try format_data(data, writer);
             },
         }
     }
@@ -364,7 +364,7 @@ pub const EventTracing = union(Event.Tag) {
 
         switch (event.*) {
             inline else => |data| {
-                try writer.print("{}", .{event_equals_format(data)});
+                try format_data(data, writer);
             },
         }
     }
@@ -414,65 +414,36 @@ pub const EventMetric = union(enum) {
     }
 };
 
-/// Format EventTiming and EventMetric's payload (ie, the tags) with a space separated k=v format.
-fn EventEqualsFormatterType(comptime Data: type) type {
-    assert(@typeInfo(Data) == .Struct or @typeInfo(Data) == .Void);
-
-    if (@typeInfo(Data) == .Void) {
-        return struct {
-            data: Data,
-            pub fn format(
-                _: *const @This(),
-                comptime _: []const u8,
-                _: std.fmt.FormatOptions,
-                _: anytype,
-            ) !void {}
-        };
-    }
-
-    return struct {
-        data: Data,
-
-        pub fn format(
-            formatter: *const @This(),
-            comptime fmt: []const u8,
-            options: std.fmt.FormatOptions,
-            writer: anytype,
-        ) !void {
-            _ = fmt;
-            _ = options;
-
-            const fields = std.meta.fields(Data);
-            inline for (fields, 0..) |data_field, i| {
-                assert(data_field.type == bool or
-                    @typeInfo(data_field.type) == .Int or
-                    @typeInfo(data_field.type) == .Enum or
-                    @typeInfo(data_field.type) == .Union);
-
-                const data_field_value = @field(formatter.data, data_field.name);
-                try writer.writeAll(data_field.name);
-                try writer.writeByte('=');
-
-                if (@typeInfo(data_field.type) == .Enum or
-                    @typeInfo(data_field.type) == .Union)
-                {
-                    try writer.print("{s}", .{@tagName(data_field_value)});
-                } else {
-                    try writer.print("{}", .{data_field_value});
-                }
-
-                if (i != fields.len - 1) {
-                    try writer.writeByte(' ');
-                }
-            }
+pub fn format_data(
+    data: anytype,
+    writer: anytype,
+) !void {
+    const Data = @TypeOf(data);
+    if (Data == void) return;
+
+    const fields = std.meta.fields(Data);
+    inline for (fields, 0..) |data_field, i| {
+        assert(data_field.type == bool or
+            @typeInfo(data_field.type) == .Int or
+            @typeInfo(data_field.type) == .Enum or
+            @typeInfo(data_field.type) == .Union);
+
+        const data_field_value = @field(data, data_field.name);
+        try writer.writeAll(data_field.name);
+        try writer.writeByte('=');
+
+        if (@typeInfo(data_field.type) == .Enum or
+            @typeInfo(data_field.type) == .Union)
+        {
+            try writer.print("{s}", .{@tagName(data_field_value)});
+        } else {
+            try writer.print("{}", .{data_field_value});
         }
-    };
-}
 
-pub fn event_equals_format(
-    data: anytype,
-) EventEqualsFormatterType(@TypeOf(data)) {
-    return EventEqualsFormatterType(@TypeOf(data)){ .data = data };
+        if (i != fields.len - 1) {
+            try writer.writeByte(' ');
+        }
+    }
 }
 
 pub const EventTimingAggregate = struct {
```

src/trace/event.zig
```diff
@@ -0,0 +1,547 @@
...
+};
+
+pub const EventMetricAggregate = struct {
+    pub const ValueType = u64;
+
+    event: EventMetric,
+    value: ValueType,
+};
+
+test "EventTiming stack doesn't have collisions" {
```
matklad: Let's do the same for EventMetrics?

src/trace.zig
```diff
@@ -99,174 +99,84 @@ const std = @import("std");
...
+    }
+
+    pub fn start(tracer: *Tracer, event: Event) void {
+        const event_tracing = event.as(EventTracing);
+        const event_timing = event.as(EventTiming);
+        const stack = event_tracing.stack();
 
-        const stack = event.stack();
         assert(tracer.events_started[stack] == null);
         tracer.events_started[stack] = tracer.timer.read();
```
matklad: Unrelated to the PR, but I am not a fan of std.time.Timer API, it's needlessly stateful. Probably we should use `std.Instant`, or just make our own using our own real `Time`.

matklad: Related unrelated thing, I don't like how we convert between ns us and ms here and there. We will mess up units of measure at least once. 

So maybe stdx should grow our own stronger-typed wrappers:

```zig
const Instant = struct { 
    ...  

    pub fn now() Instant {}
    pub fn duration_since(now: Instant, earlier: Instant) Duration {}
}

const Duration = struct {
    nanoseconds: u64

    fn microseconds(duration: Duration) u64 {}
}
```

One thing tha Timer does is that it ensurer monotonicity, but at least Rust moved over to just saturating subtraction when converting Instant to duration for this. 

Or we can just use our own Time.

src/trace/metric.zig
```diff
@@ -0,0 +1,144 @@
...
+                    .duration_min_us = duration_us,
+                    .duration_max_us = duration_us,
+                    .duration_sum_us = duration_us,
+                    .count = 1,
+                },
+            };
+        } else {
+            const timing_existing = self.events_timing[timing_slot].?.values;
+            // Certain high cardinality data (eg, op) _can_ differ.
+            // TODO: Maybe assert and gate on constants.verify
```
matklad: Can assert at least that the tag is same. And can replace `== null` and `.?.` with

```
if (self.vents_timings[timing_slot]) |*event_record| {

} else {

}
```

src/trace.zig
```diff
@@ -367,81 +287,13 @@ pub const Tracer = struct {
...
-    };
-}
 
-fn struct_format(
-    data: anytype,
-    comptime cardinality: DataFormatterCardinality,
-) StructFormatterType(@TypeOf(data), cardinality) {
-    return StructFormatterType(@TypeOf(data), cardinality){ .data = data };
-}
+    pub fn emit(tracer: *Tracer) void {
```
matklad: `fn emit_metrics()` maybe?

src/trace/metric.zig
```diff
@@ -0,0 +1,144 @@
...
+const IO = @import("../io.zig").IO;
+const StatsD = @import("statsd.zig").StatsD;
+
+const EventTiming = @import("event.zig").EventTiming;
+const EventMetric = @import("event.zig").EventMetric;
+const EventTimingAggregate = @import("event.zig").EventTimingAggregate;
+const EventMetricAggregate = @import("event.zig").EventMetricAggregate;
+
+pub const Metrics = struct {
+    pub const StatsDOptions = ?struct {
```
matklad: Let's do 

```
union(enum) {
    disabled, 
    enabled: struct { io: *IO, address: std.net.Address }, 
}
```

With null at a call-site, it's unclear if it means enabled with default parameters or disabled completely.

src/trace/metric.zig
```diff
@@ -0,0 +1,144 @@
+const std = @import("std");
+const assert = std.debug.assert;
+
+const IO = @import("../io.zig").IO;
+const StatsD = @import("statsd.zig").StatsD;
+
+const EventTiming = @import("event.zig").EventTiming;
+const EventMetric = @import("event.zig").EventMetric;
+const EventTimingAggregate = @import("event.zig").EventTimingAggregate;
+const EventMetricAggregate = @import("event.zig").EventMetricAggregate;
```
matklad: Let's move the types here, they are unused in the `event.zig`, so this is needless indirection.

src/trace/metric.zig
matklad: Ah, reading the source code, `disabled` is actually just logging! So, yeah, worth naming here.

src/trace/metric.zig
matklad: ➕💯

src/trace/metric.zig
```diff
@@ -0,0 +1,144 @@
...
+
+pub const Metrics = struct {
+    pub const StatsDOptions = ?struct {
+        io: *IO,
+        address: std.net.Address,
+    };
+
+    events_metric: []?EventMetricAggregate,
```
matklad: Maybe get rid of `?` here, so that we don't waste bits for alignment, and can memset the literal zero byte for reset? We can use count!=0 for timings, and we can store value+1 for gauges?

Though, I guess `?` are also fine

src/trace/metric.zig
```diff
@@ -0,0 +1,144 @@
...
+
+const IO = @import("../io.zig").IO;
+const StatsD = @import("statsd.zig").StatsD;
+
+const EventTiming = @import("event.zig").EventTiming;
+const EventMetric = @import("event.zig").EventMetric;
+const EventTimingAggregate = @import("event.zig").EventTimingAggregate;
+const EventMetricAggregate = @import("event.zig").EventMetricAggregate;
+
+pub const Metrics = struct {
```
matklad: Do we need Metrics as separate struct? The heavy lifting is done by the Statsd anyway, so we might inline the metrics into tracer?

src/trace/statsd.zig
```diff
@@ -0,0 +1,599 @@
...
+        udp: struct {
+            socket: std.posix.socket_t,
+            io: *IO,
+            send_callback_error_count: u64 = 0,
+        },
+        log,
+    },
+
+    buffer_completions: BufferCompletionRing,
+    buffer_completions_buffer: []BufferCompletion,
```
matklad: This pair looks like `IOPS(BufferCompletion, packet_count_max)` in disguise?

src/trace/statsd.zig
```diff
@@ -0,0 +1,599 @@
...
+            }
+
+            stdx.copy_disjoint(
+                .inexact,
+                u8,
+                buffer_completion.buffer[buffer_completion_written..],
+                statsd_line,
+            );
+            buffer_completion_written += statsd_line.len;
+        }
```
matklad: Hm, this feels like a very complicated way to format the metrics, and it also goes through the memcpy. I am also not a fan of `const Output = ?union(enum) { none, some: []const u8 };` type we use internally. 

Can we do it more directly? Something like this:

```zig
        for (events_metric, self.events_metric_emitted) |new, *old| {
            if (new == null) continue;
            if (old.* != null and old.?.value == new.?.value) continue;
            old.* = new;
            offset += format_metric(completion.buffer[offset..], new.?) catch {
                self.emit_buffer(completion, offset);
                offset = 0;
                buffer = self.buffer_completions.pop() orelse return;
                offset += format_metric(completion.buffer[offset..], new.?) catch unreachable;
            };
        }
        for (events_timing) |new| {
            if (new == null) continue;
            offset += format_timing(completion.buffer[offset..], new.?) catch {
                self.emit_buffer(completion, offset);
                offset = 0;
                buffer = self.buffer_completions.pop() orelse return;
                offset += format_timing(completion.buffer[offset..], new.?) catch unreachable;
            };
        }

```


That is, write two loops, format directly into the buffer, advance offset if formatting succeeded and, otherwise, fetch a new buffer and try reformatting. 

So, the events that stride buffers get formatted twice, but then we don't memcpy stuff on the happy path. 


I don't like that this duplicates substantial logic for try format, rotate a buffer, format again, but, if we are adventurous, we could also do an outer inline-for-loop  over metrics and timings and branch on the type of the thing internally?

```
        inline for (.{events_metric, events_timing}) |things| {
            const Thing = @TypeOf(things[0]);
            for (things) |thing| {
                if (Thing == EventMetricAggregate) // extra break for the same value
            }
        }
```


We still need a test that each metric fits into a packet, and that all metrics don't occupy too much space, but asserting that seems easier than computing buffer sizes at the first place.

src/trace/metric.zig
sentientwaffle: I went with `udp` and `log` as the variant names, since that is more descriptive and also matches `StatsD`'s internal `implementation` tagged union.

src/trace/metric.zig
sentientwaffle: `statsd.zig` also depends on them, so moving them into `metric.zig` would create a circular dependency. I know that zig can resolve it just fine, but I like to avoid circular dependencies since they make reading the code more confusing.

sentientwaffle: Oh, I see in a later comment:

> Do we need Metrics as separate struct? The heavy lifting is done by the Statsd anyway, so we might inline the metrics into tracer?

So I will consider those together!

src/trace.zig
```diff
@@ -284,63 +194,73 @@ pub const Tracer = struct {
...
                 tracer.events_started[stack] = null;
-                tracer.write_stop(@intCast(stack), .{});
+                tracer.write_stop(@intCast(stack));
             }
         }
-    }
 
-    fn write_stop(tracer: *Tracer, stack: u32, data: anytype) void {
-        comptime assert(@typeInfo(@TypeOf(data)) == .Struct);
+        tracer.metrics.reset_all();
```
sentientwaffle: :thinking: Any idea why this reset is here?

src/trace/metric.zig
sentientwaffle: What if we go even farther? If we use a more structured approach to convert event data to slot index, we can get the backwards mapping too, which means we don't need to store the event data, just the value. Then when we go to emit the metric, we use the backwards mapping from the slot index to generate the event data.

I'm imagining something like this: (its not compiled/tested so it probably doesn't work as-is):

```
    const cardinalities = []struct { Event.Tag, []const u8, u32 }{
        .{ .replica_commit, "stage", enum_max(CommitStage.Tag) },
        .{ .compact_beat, "tree", enum_max(TreeEnum) },
        .{ .compact_beat, "level_b", @as(u32, constants.lsm_levels) },
        .{ .compact_mutable, "tree", enum_max(TreeEnum) },
        .{ .compact_mutable_suffix, "tree", enum_max(TreeEnum) },
        .{ .lookup, "tree", enum_max(TreeEnum) },
        .{ .lookup_worker, "tree", enum_max(TreeEnum) },
        .{ .scan_tree, "tree", enum_max(TreeEnum) },
        .{ .scan_tree_level, "tree", enum_max(TreeEnum) },
        .{ .scan_tree_level, "level", @as(u32, constants.levels) },
    };

    // Verify that each data field in the event is included in `cardinalities` exactly once.
    comptime {
        for (std.enums.values(Event.Tag)) |event_tag| {
            const EventData = std.meta.TagPayload(EventMetric, event_tag)
            if (@typeInfo(EventData) == .Struct) {
                for (std.meta.fieldNames(EventData)) |event_data_field| {
                    var hit: u32 = 0;
                    for (cardinalities) |cardinality| {
                        hit += @intFromBool(cardinality.0 == event_tag and
                            std.mem.eql(u8, cardinality.1, event_data_field));
                    }
                    assert(hit == 1);
                }
            } else {
                assert(EventData == void);
            }
        }
    }

    pub const slot_limits = array: {
        var array = std.enums.EnumArray(Event.Tag, u32).initFill(1);
        for (cardinalities) |cardinality| {
            const event_tag = cardinality.0;
            const event_field_limit = cardinality.2;
            array.set(event_tag, array.get(event_tag) * event_field_limit);
        }
        break :array array;
    };
```

(And extract most of that logic so that it is shared by `EventTiming`/`EventMetric`.)

... But now that I've written all of that it, it seems like the complexity may not be worth it.
Conceptually there's no reason to store the metric data at the slot index, since you can derive it. But in practice it is simpler and also our metric cardinality is not very high, so the memory savings are meaningless.

src/trace.zig
matklad: Probably just an accident!

Maybe the bug is that `Tracer.rest` is called reset rather than cancel? It is actively doing things, it's not simply resetting the memory. 

I am not sure what's our naming scheme for reset vs cancel, but my decision tree is:

* if it is async, then definitely cancel. 
* if it's simple data structure `clear` than it's reset.
* if it is synchronous, but is non-trivial in some way (side effectual, or actively removes resources from unrelated links lists), than it also should be cancel. 

Thinking bigger, maybe _nothing_ should be called `reset` and instead we should use `clear` for data structures and `cancel` for everything else?

src/trace/metric.zig
sentientwaffle: I think I will leave it as `?EventMetricAggregate`, for a couple reasons:
- I don't want to risk off-by-one's in our metrics, and
- this seems like something the compiler could optimize for us automatically (in the future). e.g. it could see that `EventMetricAggregate`'s tag does not require all 8 bits, and repurpose one.

src/trace.zig
```diff
@@ -284,67 +227,73 @@ pub const Tracer = struct {
...
 
-    fn write_stop(tracer: *Tracer, stack: u32, data: anytype) void {
-        comptime assert(@typeInfo(@TypeOf(data)) == .Struct);
-
+    fn write_stop(tracer: *Tracer, stack: u32) void {
         const writer = tracer.options.writer orelse return;
-        const time_now = std.time.Instant.now() catch unreachable;
-        const time_elapsed_ns = time_now.since(tracer.time_start);
-        const time_elapsed_us = @divFloor(time_elapsed_ns, std.time.ns_per_us);
+        const time_now = stdx.Instant.now();
```
matklad: This is micro-optimization, but I'd maybe pass `time_elapesd` in as a parameter here, so that in `stop` we don't lookup the current time twice

src/trace.zig
```diff
@@ -359,89 +308,68 @@ pub const Tracer = struct {
...
+    // * sum/count are considered counters for aggregation: they are added to the existing values.
+    //
+    // This matches the default behavior of the `g` and `c` statsd types respectively.
+    fn timing(tracer: *Tracer, event_timing: EventTiming, duration_us: u64) void {
+        const timing_slot = event_timing.slot();
+
+        if (tracer.events_timing[timing_slot]) |*event_timing_existing| {
+            const timing_existing = event_timing_existing.values;
+            // Certain high cardinality data (eg, op) _can_ differ.
+            // TODO: Maybe assert and gate on constants.verify
```
matklad: obsolete todo

matklad: (and the comment is wrong, we don't have op in EventTiming

src/trace.zig
```diff
@@ -284,67 +227,73 @@ pub const Tracer = struct {
...
-            .data = struct_format(data, .sparse),
-            .args = if (comptime @TypeOf(data) == @TypeOf(.{}))
-                "{}" // Serialize `.{}` as an empty object, not as an empty array.
-            else
-                std.json.Formatter(@TypeOf(data)){ .value = data, .options = .{} },
+            .timestamp = time_elapsed.microseconds(),
+            .event_tracing = event_tracing,
+            .event_timing = event_timing,
+            .args = std.json.Formatter(Event){ .value = event, .options = .{} },
         }) catch unreachable;
```
matklad: FWIW, I am slightly uneasy about this unreachable. We either should test that all events fit into the buffer, or maybe just `log.err` here instead of outright unreachable? Would be a pity if metrics kill us if we reach over 9000 ops or something, and it's not immediately obvious from the code that we are not overflowing this buffer here.

src/trace/statsd.zig
```diff
@@ -0,0 +1,463 @@
...
+                packet_count_max,
+            });
+        }
+
+        if (self.implementation == .udp and self.implementation.udp.send_callback_error_count > 0) {
+            log.warn(
+                "failed to send {} packets",
+                .{self.implementation.udp.send_callback_error_count},
+            );
+            self.implementation.udp.send_callback_error_count = 0;
```
matklad, line 212: Do we need to maintain the error count as a state? Why don't we immediately log error once we receive it? It shouldn't be _too_ many errors I think?

src/trace/statsd.zig
```diff
@@ -0,0 +1,463 @@
...
+        };
+
+        var buffer_written: usize = 0;
+        inline for (.{ events_metric, events_timing }) |events| {
+            for (events, 0..) |event_new_maybe, i| {
+                const event_new = event_new_maybe orelse continue;
+                if (@TypeOf(event_new) == EventMetricAggregate) {
+                    const event_old = &self.events_metric_emitted[i];
+                    if (event_old.* != null and event_old.*.?.value == event_new.value) continue;
+                    event_old.* = event_new;
```
matklad: This assignment feels redundant with 

```
        stdx.copy_disjoint(
            .exact,
            ?EventMetricAggregate,
            self.events_metric_emitted,
            events_metric,
        );
```

bellow. 

I am slightly confused as to what we are trying to do here, to be honest --- we have a check for not sending duplicate data here, but we _also_ reset metrics(aka gauges) on the tracer side as well.

Maybe:

* remove this `event_old.* = event_new;`
* remove memcopy below
* keep `@memset(tracer.events_metric, null);`

Than, the invariant that we send gauges that were set between two emit_metrics calls. Which feels nice in the face of udp packets dropped --- the current implementation might _hide_ a change in the gauge if the "edge triggered" packet, the packet when the change did happen, is dropped.

src/trace/statsd.zig
```diff
@@ -0,0 +1,463 @@
...
+
+/// This implementation emits on an open-loop: on the emit interval, it fires off up to
+/// packet_count_max UDP packets, without waiting for completions.
+///
+/// The emit interval needs to be large enough that the kernel will have finished processing them
+/// before emitting again. If not, an error will be logged.
+const packet_count_max = stdx.div_ceil(
+    EventMetric.slot_count + EventTiming.slot_count,
+    packet_messages_max,
+);
```
matklad, line 101: Let's add comptime assert that `packet_count_max` is not tooo large (256?). 

TBH, right now it is already `122` which feels like it on the larger size? That's 122 syscalls!

src/trace/statsd.zig
```diff
@@ -0,0 +1,463 @@
...
+                        .{ .timing = .{ .aggregate = event_new, .stat = .max } },
+                        .{ .timing = .{ .aggregate = event_new, .stat = .avg } },
+                        .{ .timing = .{ .aggregate = event_new, .stat = .sum } },
+                        .{ .timing = .{ .aggregate = event_new, .stat = .count } },
+                    },
+                    else => unreachable,
+                };
+
+                for (stats) |stat| {
+                    buffer_written += format_metric(
```
matklad: Ok, yeah, this feels much simpler now, but I got another crazy idea! 

What if we don't have a per-write buffer, but have just one huge, contiguous buffer for _all_ metrics?

Then, we just format _all_ metrics into this single buffer without any retries, and:


* Either maintain a bounded array of offsets of boundaries, to do one vectored write at the end
* Or maintain the running offset sent, and send a slice as soon as we have an mtg-worth of data

The drawback would be that we'd have to wait for _all_ packets to come through before starting the next loop, but that seems fine, as we _do_ expect all the packets to go through.

src/trace/statsd.zig
matklad: Although, I recall now that we use io_uring, we'll do one syscall effectively. Although now that I think even more about this, we setup io_uring with 128 slots:

```
command.io = try IO.init(128, 0);
```

So, we'll be burning almost entire turn of the ring just for all the UDP packets.

Hm, this doesn't compute for me: 122 packets * 1400 mtu is 153KiB, which is 5% of LOTR, or a LOTR every 30 seconds, given out emit metrics timeout of 100 ticks. It _feels_ like that's too much data. 

Are we really sending this much worst-case, or is there some error in the computation for the required amount of space?

src/trace/statsd.zig
sentientwaffle: I think the idea is to debounce it, so that you don't end up with (in the worst case) 122 logs/second if there is something messed up.

sentientwaffle: (e.g. if you give TB the wrong port, or the stats server is down, you wouldn't want to spam the logs.)

sentientwaffle: Interestingly, if you run the benchmark with `--statsd=...` but don't start any server on that UDP port, despite sending 75-80 packets/s, only 30-40 of those will report errors (`error.ConnectionRefused`)! I guess the others get dropped?

src/trace/statsd.zig
sentientwaffle: Example stat:
```
tb.compact_beat_merge_us.count:491|c|#cluster:00000000000000000000000000000000,replica:0,tree:Transfer.credit_account_id,level_b:0
```
(That cluster is a hex-encoded `u128`.)

We fit ~12 in a UDP packet. So I think 122 packets sounds about right.
In the benchmark, we send 75-80 packets/second. That number would go up as we fill levels.

Here is an example set of packets from a single `emit` during the benchmark: https://gist.github.com/sentientwaffle/c297490ce0ec9acec4023e6a7844365a

Also keep in mind that in practice each machine will often have its own statsd daemon, which means these UDP packets don't necessarily need to traverse the actual network.

---

Since io_uring is batching the syscalls I think we're okay there, though we might want to increase the ring size.

Overall I think this is a price well worth paying for observability! If it does turn out to be expensive I would suggest that we change the emit rate rather than reduce the metrics recorded (e.g. emit every 10 seconds or every 30 seconds instead). The high resolution metrics are really only useful if you are doing live metrics (e.g. a real-time animated chart) -- otherwise you don't need to be more granular than 1 minute anyway.

src/vsr/replica.zig
```diff
@@ -1294,7 +1294,7 @@ pub fn ReplicaType(
                 .trace_emit_timeout = Timeout{
                     .name = "trace_emit_timeout",
                     .id = replica_index,
-                    .after = 100,
+                    .after = 1000,
```
matklad, line 1297: This is every 15 seconds, right? Tick is 15 ms?

matklad (dismissed):

src/vsr/replica.zig
sentientwaffle: This is 10 seconds, since tick is 10ms:

```
    tick_ms: u63 = 10,
```

matklad: 🤯 I was 110% sure it is 15! One of my recurring thoughts is that "gosh, we should increase our tick from 15 ms to 16 ms, to be _exactly_ like a game". 

I wonder where did I get 15 from?

sentientwaffle:
(Retrying now that https://github.com/tigerbeetle/tigerbeetle/pull/2749 is in to get logs.)

src/trace/statsd.zig
```diff
@@ -96,7 +96,7 @@ comptime {
 /// The emit interval needs to be large enough that the kernel will have finished processing them
 /// before emitting again. If not, an error will be logged.
 const packet_count_max = stdx.div_ceil(
-    EventMetric.slot_count + EventTiming.slot_count,
+    EventMetric.slot_count + (EventTiming.slot_count * std.enums.values(TimingStat).len),
```
matklad, line 99: Ouch, great catch!

matklad (approved):

PR #2699, opened 2025-02-03
release 2025-02-03

cb22:


CHANGELOG.md
```diff
@@ -3,6 +3,75 @@
...
+
+- [#2681](https://github.com/tigerbeetle/tigerbeetle/pull/2681)
+
+  Remove zero-padding assert in the grid. After we read a block, we verify it has a valid header,
+  body, it was indeed the block we were expecting to find and assert that its sector padding is
+  zeroed.
+
+  This last step can fail due to disk corruption; sector padding is covered by any checksum! This
+  wasn't caught by the VOPR, since previously it would inject a storage fault that corrupts the
+  _entire_ sector.
```
sentientwaffle: This PR evolved so the original description is out of date. Maybe something like this instead:

```suggestion
  Consider blocks and prepares with nonzero padding to be corrupt.
  Previously blocks asserted zero padding, which can fail due to bitrot.
  
  Also fix a similar bug in the superblock copy index handling. The copy index is not covered
  by a checksum, so we must treat it carefully to avoid propagating bad data if it is corrupt.
  
  VOPR now injects single-bit errors into storage rather than whole-sector errors.
```

batiati (dismissed):

sentientwaffle (dismissed):

cb22:
Y'all too quick :smile: (the suggestion has leading whitespace, so rebased and removed it)

sentientwaffle (approved):

PR #2701, opened 2025-02-04
docs: Improve keyboard navigation

fabioarnold:
In search you can now use the up and down arrow keys to select page folders and enter to collapse and expand the folder. When selecting a folder, a preview of the page is opened in the right pane.

I deployed a preview version on https://tbdocs.fabioware.com.

https://github.com/user-attachments/assets/df78ed49-1ff3-4099-9e23-1d035afd46ee


batiati (approved):
Thanks for the detailed commit messages 😎

PR #2702, opened 2025-02-04
repl: use `StaticAllocator` on init and deinit

### LINKED ISSUE #2515: Use StaticAllocator in Repl.run

fidelicura:
>@brson at https://github.com/tigerbeetle/tigerbeetle/issues/2515#issuecomment-2632073831: Probably in `Repl.init` or `Repl.run`, it needs to wrap the given arena allocator in `StaticAllocator`, do its allocation, then call `transition_from_init_to_static`.

Add `static_allocator` field to `Repl`, wrap passed allocator (currently, arena allocator) with `StaticAllocator` in `Repl.init` and deinitialize it in `Repl.deinit`. Closes #2515 after merged #2542.

This is my first pull request and the most naive solution I've come up with, so please review it carefully. Thanks.

### DIFF
--- a/src/repl.zig
+++ b/src/repl.zig
@@ -5,6 +5,7 @@ const vsr = @import("vsr.zig");
 const stdx = vsr.stdx;
 const constants = vsr.constants;
 const IO = vsr.io.IO;
+const StaticAllocator = @import("static_allocator.zig");
 const Storage = vsr.storage.StorageType(IO);
 const StateMachine = vsr.state_machine.StateMachineType(
     Storage,
@@ -54,6 +55,8 @@ pub fn ReplType(comptime MessageBus: type, comptime Time: type) type {
         arguments: std.ArrayListUnmanaged(u8),
         message_pool: *MessagePool,
         io: *IO,
+
+        static_allocator: StaticAllocator,
 
         const Repl = @This();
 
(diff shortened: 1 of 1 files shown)


brson:
Thanks @fidelicura. Looks good, but let's make one change to be consistent with the patterns in `replica.zig`.

[In `replica.zig`](https://github.com/tigerbeetle/tigerbeetle/blob/main/src/vsr/replica.zig#L1273) the `deinit` function still takes `allocator` as an argument, and passes that allocator to its fields' `deinit` functions. It's functionally the same as what you have here, where the repl pulls the allocator out of `static_allocator`; but I think it's best we keep the consistent pattern of allocators being passed to both init and deinit. So please tweak this to keep passing the allocator to `deinit`.

brson (changes requested):
Per my previous comment.

fidelicura:
To be honest, it was a bit difficult to understand why the `static_allocator` field was needed at all: it only became clear when I saw `replica.deinit(replica.static_allocator.parent_allocator);` :smile:
I hope that what's needed now.

brson (approved):
lgtm thanks @fidelicura

PR #2703, opened 2025-02-05
vsr: remove logic to send both start_view and start_view_deprecated

chaitanyabhandari:
This is the second last PR in the series of the PRs that intends to roll out a new CheckpointState format (the first two were https://github.com/tigerbeetle/tigerbeetle/pull/2662 and https://github.com/tigerbeetle/tigerbeetle/pull/2600). It takes care of removing the logic to send and handle both `start_view` & `start_view_deprecated` messages. Replicas now panic when they receive the `start_view_deprecated` message.

The only thing left to do now is to remove the old checkpoint state format and the translation logic between `CheckpointStateOld` → `CheckpointState`. This can only be taken up when release `0.16.31` is rolled out (assuming no releases are skipped in the middle). Specifically, this translation logic can only be removed when our multiversion binaries stop bundling any release that writes `CheckpointStateOld` to disk (see https://github.com/tigerbeetle/tigerbeetle/pull/2703#discussion_r1945483335 for the reason).

### DIFF
--- a/src/message_pool.zig
+++ b/src/message_pool.zig
@@ -33,7 +33,6 @@ pub const Options = union(vsr.ProcessType) {
                 // Handle bursts.
                 // (e.g. Connection.parse_message(), or sending a ping when the send queue is full).
                 sum += 1;
-                sum += 1; // Extra burst from sending two SV variants.
 
                 // This conditions is necessary (but not sufficient) to prevent deadlocks.
                 assert(sum > 1);
@@ -106,10 +105,6 @@ pub const MessagePool = struct {
         pub const StartViewChange = CommandMessageType(.start_view_change);
         pub const DoViewChange = CommandMessageType(.do_view_change);
         pub const StartView = CommandMessageType(.start_view);
-        pub const StartViewDeprecated = CommandMessageType(.start_view_deprecated);
-        comptime {
-            assert(StartView == StartViewDeprecated);
-        }
         pub const RequestStartView = CommandMessageType(.request_start_view);
         pub const RequestHeaders = CommandMessageType(.request_headers);
         pub const RequestPrepare = CommandMessageType(.request_prepare);
--- a/src/vsr.zig
+++ b/src/vsr.zig
@@ -215,11 +215,6 @@ pub const Command = enum(u8) {
     request_blocks = 19,
     block = 20,
 
-    // Historical version of SV, with the CheckpointStateOld format. Currently, both
-    // `start_view_deprecated` and `start_view` are handled. Next release will ignore
-    // `start_view_deprecated`.
-    start_view_deprecated = 23,
-
     start_view = 24,
 
     // If a command is removed from the protocol, its ordinal is added here and can't be re-used.
@@ -227,6 +222,7 @@ pub const Command = enum(u8) {
         12, // start_view without checkpoint
         21, // request_sync_checkpoint
         22, // sync_checkpoint
+        23, // start_view with an older version of CheckpointState
     };
 
     comptime {
--- a/src/vsr/message_header.zig
+++ b/src/vsr/message_header.zig
@@ -88,7 +88,6 @@ pub const Header = extern struct {
             .commit => Commit,
             .start_view_change => StartViewChange,
             .do_view_change => DoViewChange,
-            .start_view_deprecated => StartView,
             .start_view => StartView,
             .request_start_view => RequestStartView,
             .request_headers => RequestHeaders,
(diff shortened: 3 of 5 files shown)


src/vsr/superblock.zig
```diff
@@ -1339,9 +1265,7 @@ pub fn SuperBlockType(comptime Storage: type) type {
 
                 const working = quorum.header;
 
-                // TODO: Remove the second condition when logic to translate CheckpointStateOld to
-                // CheckpointState is removed.
-                if (working.version != SuperBlockVersion and working.version != 1) {
+                if (working.version != SuperBlockVersion) {
```
sentientwaffle: Ahh, I think we can't remove this or the `CheckpointOld` translation yet -- not until our production release stops bundling any release that writes the old checkpoint.

Suppose you have a bundle of releases `[1, 2, 3]`. Release `1` writes `CheckpointOld`, release `2` can read `CheckpointOld` and write `Checkpoint`, and release `3` can only read `Checkpoint`.

If you try to upgrade a replica from `1 -> 3` with a cold-start replica (i.e. stop the replica then start it, rather than relying on our auto-upgrade), then what happens is 3 initially starts up and reads the superblock -- normally this is where it would see that the superblock's current release is `1` and execute that version -- but instead it fails because the checkpoint version is not recognized.

src/message_pool.zig
sentientwaffle, line 1: You can remove

```
sum += 1; // Extra burst from sending two SV variants.
```

Which, I just noticed is being added to `client`, not `replica`!

chaitanyabhandari: Oops, this was a bad miss on my part. :o

Wish I'd caught this one before, and I'm glad it hasn't caused any issues, perhaps because we have some wiggle room because of `sum += 1; // Handle bursts (e.g. Connection.parse_message)` in `replica`.

src/vsr/superblock.zig
chaitanyabhandari: Oh wow, I hadn't considered this cold start case at all! I thought that even in the case that you've mentioned, it would be guaranteed that `2` would be able to parse `CheckpointOld` and would be guaranteed to write `Checkpoint` before `3` starts up... I'm still wrapping my head around this though, even in the cold start case, we would be able to start up and upgrade to `3` only if `3` is durable in our superblock. 

IIUC, the issue here is that even though we upgrade from 1 → 2 → 3 when we rely on our auto upgrade, we only ever make the *highest* release bundled durable in the superblock. So, in the case a replica crashes and is started after `3` is made durable in the superblock but before upgrade from `1` → `2`, we _can_ directly go from `1` → `3`... Does my reasoning here sound correct @sentientwaffle?

sentientwaffle: I don't think that's quite it.

Let's use a concrete example -- our current production release is `0.16.26`.
Its "Oldest upgradable replica version" is `0.16.20`, which means releases `0.16.20 ... 0.16.25` are also included in the binary.

When you execute the binary, the first thing to run is `0.16.26`'s code -- e.g. if you run `tigerbeetle --help`, that is `0.16.26`'s help text.

When you run `tigerbeetle start`, it goes to open the replica as normal. Then in `Replica.open`, after we have opened the superblock:

```
            const release_target = self.superblock.working.vsr_state.checkpoint.release;
            assert(release_target.value >= self.superblock.working.release_format.value);
            log.info("superblock release={}", .{release_target});

            if (release_target.value != self.release.value) {
                self.release_transition(@src());
                return;
            }
```

That is, we check the `release` in the superblock checkpoint, and if it is not our current release, we (try to) execute that release instead.

The problem is that if (jumping back to our example release numbers) release `3` doesn't know how to interpret release `1`'s `CheckpointOld`, then the "opening the superblock" stage mentioned above will fail. If `3` _could_ interpret the `CheckpointOld`, then it would jump to release `1`, which in turn would upgrade to release `2`, which would upgrade to release `3` -- and since release `2` writes the new version of the checkpoint, release `3` would be fine.

sentientwaffle: (The above describes a cold start -- not the auto-upgrade where you just swap the binary but don't manually kill the old replica. I think the auto-upgrade would work fine, since it is the old replica that is doing the dispatching.)

chaitanyabhandari: > When you execute the binary, the first thing to run is 0.16.26's code -- e.g. if you run tigerbeetle --help, that is 0.16.26's help text.

Ah, my understanding was missing this crucial context! Its crystal clear now, thanks DJ. And this was a wonderful catch.

sentientwaffle (approved):

PR #2705, opened 2025-02-06
Upgrade to zig 0.14.1

batiati:
### Zig 0.14.1 upgrade.  

No `TODO(zig)` were evaluated, only the raw migration.
Plan to review each TODO in follow-up PRs.  

See commit messages for more details.

### DIFF
--- a/src/build_multiversion.zig
+++ b/src/build_multiversion.zig
@@ -21,11 +21,11 @@ const Target = union(enum) {
 
     pub fn parse(str: []const u8) !Target {
         const targets = .{
-            .{ "x86_64-linux", .{ .linux = .x86_64 } },
-            .{ "aarch64-linux", .{ .linux = .aarch64 } },
-            .{ "x86_64-windows", .{ .windows = .x86_64 } },
-            .{ "aarch64-windows", .{ .windows = .aarch64 } },
-            .{ "macos", .macos },
+            .{ "x86_64-linux", Target{ .linux = .x86_64 } },
+            .{ "aarch64-linux", Target{ .linux = .aarch64 } },
+            .{ "x86_64-windows", Target{ .windows = .x86_64 } },
+            .{ "aarch64-windows", Target{ .windows = .aarch64 } },
+            .{ "macos", Target.macos },
         };
 
         inline for (targets) |t| if (std.mem.eql(u8, str, t[0])) return t[1];
--- a/src/build_multiversion.zig
+++ b/src/build_multiversion.zig
@@ -20,12 +20,12 @@ const Target = union(enum) {
     macos, // Universal binary packing both x86_64 and aarch64 versions.
 
     pub fn parse(str: []const u8) !Target {
-        const targets = .{
-            .{ "x86_64-linux", Target{ .linux = .x86_64 } },
-            .{ "aarch64-linux", Target{ .linux = .aarch64 } },
-            .{ "x86_64-windows", Target{ .windows = .x86_64 } },
-            .{ "aarch64-windows", Target{ .windows = .aarch64 } },
-            .{ "macos", Target.macos },
+        const targets = [_]struct { []const u8, Target }{
+            .{ "x86_64-linux", .{ .linux = .x86_64 } },
+            .{ "aarch64-linux", .{ .linux = .aarch64 } },
+            .{ "x86_64-windows", .{ .windows = .x86_64 } },
+            .{ "aarch64-windows", .{ .windows = .aarch64 } },
+            .{ "macos", .macos },
         };
 
         inline for (targets) |t| if (std.mem.eql(u8, str, t[0])) return t[1];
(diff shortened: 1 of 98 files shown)


batiati:
Related: https://github.com/ziglang/zig/pull/22841

ivelsantos:
> ```
> src/aof.zig:853:39: error: expected type '[][]const u8', found '[][:0]const u8'
>         try aof_merge(allocator, paths[0 .. count - 2], "prepared.aof");
>                                  ~~~~~^~~~~~~~~~~~~~~~
> src/aof.zig:853:39: note: pointer type child '[:0]const u8' cannot cast into pointer type child '[]const u8'
> src/aof.zig:853:39: note: pointer sentinel '0' cannot cast into pointer sentinel 'unreachable'
> src/aof.zig:470:18: note: parameter type declared here
>     input_paths: [][]const u8,
> ```

With  `[][]const u8` you're giving `input_paths` a view into `paths[0 .. count-2]` that allows setting elements to null-terminated slices. `input_paths` should have `[]const []const u8` type.
Related issue: https://github.com/ziglang/zig/issues/23016

matklad:
Did a quick review, it all sounds reasonable to me!

Consider filing some commits which can work with both new and old Zig separately, to keep the diff smaller here, and to potentially simplify investigation should be discover something funny post merge. 

For example, [3f6339a](https://github.com/tigerbeetle/tigerbeetle/pull/2705/commits/3f6339a9c84927e7086406a85c4f0f4cadce2f05) and tokenize/tokenizeScalar changes can be landed already.

Actually, now that I think about it, vendoring AEGIS we can also just do already?

batiati:
> Consider filing some commits which can work with both new and old Zig separately, to keep the diff smaller here, and to potentially simplify investigation should be discover something funny post merge.

Done (#2812, #2814 and #2819)!
Much smaller diff now with better commit messages.

src/build_multiversion.zig
```diff
@@ -21,11 +21,11 @@ const Target = union(enum) {
 
     pub fn parse(str: []const u8) !Target {
         const targets = .{
```
matklad: I guess we can do `[_]struct { const []u8, Target }{` here?

batiati: yeah, nice usage for explicitly declared tuples.

src/state_machine.zig
```diff
@@ -7868,6 +7868,12 @@ test "deprecated operations" {
 // Sanity test to check the maximum batch size on a 1MiB message.
 // For a comprehensive test of all operations, see the `input_valid` test.
 test "StateMachine: batch_elements_max" {
+    // TODO: Zig 0.13.0 lazy compilation allowed creating a state machine
+    // with `test_min` config and a 1MiB message. With Zig 0.14.1, many
+    // assertions are hit during compile time.
+    // We should either move this test to another non-testing binary or isolate
+    // the code in another type that doesn’t require creating the state machine.
+    if (true) return error.SkipZigTest;
```
batiati, line 7876: I'd rather fix this TODO in a separate PR. I feel it's part of a bigger refactor, since we don't want to propagate constants as parameters to the statemachine anymore.

Maybe we could replace this test with `comptime` assertions.

src/io/darwin.zig
```diff
@@ -592,7 +592,28 @@ pub const IO = struct {
...
+                        error.SymLinkLoop => unreachable,
+                        error.NameTooLong => unreachable,
+                        error.FileNotFound => unreachable,
+                        error.NotDir => unreachable,
+                        error.NetworkUnreachable => unreachable,
+                        error.AddressNotAvailable => unreachable,
+                        error.SocketNotConnected => unreachable,
+                        error.UnreachableAddress => unreachable,
+                        else => |e| return e,
+                    };
```
matklad, line 616: Hm, I don't understand this change, https://github.com/ziglang/zig/pull/20223 was merged a while ago, right? So we should be fine with `send`, no?

matklad (commented):
LGTM overall, but I don't understand the reason fro this change:

https://github.com/tigerbeetle/tigerbeetle/pull/2705/commits/22db4e9adfddd38ee698f8a6809f5586fb51a1a0#r2182827426

I did read the linked issue/PR, and still don't get it. Might be a reading comprehension problem on my side :D

src/io/darwin.zig
matklad: Ok, got it now! https://github.com/ziglang/zig/pull/20223 fixes _only_ `sendto`, it doesn't fix `send`, so the upstream issue is incorrectly closed. I asked upstream to re-open

matklad (approved):

matklad:
Going to slap merge when ready to upgrade during sync!

src/io/darwin.zig
batiati: This change added `ConnectionRefused` to `sendto()`... Using `send()` will panic.

https://github.com/ziglang/zig/pull/20223/files#diff-503dcf04ec9ce2a1818ec55644fa34ff5918e86bd38a565c0186b201c06dd540

PR #2707, opened 2025-02-06
Devhub: Format fuzz count with commas

sentientwaffle:
Now the we have VOPR green, the numbers are larger.

2025-02-06T13:36:56,044036827-08:00

### DIFF
--- a/src/devhub/devhub.js
+++ b/src/devhub/devhub.js
@@ -142,7 +142,7 @@ async function mainSeeds() {
           <td><code>${record.command}</code></td>
           <td><time>${seedDuration}</time></td>
           <td><time>${seedFreshness} ago</time></td>
-          <td>${record.count}</td>
+          <td>${record.count.toLocaleString('en-US').replace(/,/g, '&nbsp;')}</td>
       `;
     tableDom.appendChild(rowDom);
   }


src/devhub/devhub.js
```diff
@@ -142,7 +142,7 @@ async function mainSeeds() {
           <td><code>${record.command}</code></td>
           <td><time>${seedDuration}</time></td>
           <td><time>${seedFreshness} ago</time></td>
-          <td>${record.count}</td>
+          <td>${record.count.toLocaleString()}</td>
```
matklad: FWIW, I've recently switched from toLocaleString to manual formatting for dates, because local-dependent formatting produced really weird dates (I mean, I think it picked US locale :) )

https://github.com/tigerbeetle/tigerbeetle/pull/2672

So let's maybe remove dependency on user's locale here and pass `en-US`? 

I also personally like just spaces as the separator:

image

so we might also want to do `/,/&nbsp;` on the result. 

Though, LGTM as is!

src/devhub/devhub.js
```diff
@@ -142,7 +142,7 @@ async function mainSeeds() {
           <td><code>${record.command}</code></td>
           <td><time>${seedDuration}</time></td>
           <td><time>${seedFreshness} ago</time></td>
-          <td>${record.count}</td>
+          <td>${record.count.toLocaleString('en-US').replace(/,/g, ' ')}</td>
```
matklad:
```suggestion
          <td>${record.count.toLocaleString('en-US').replace(/,/g, '&nbsp;')}</td>
```

we do want this to be non-breaking, otherwise the same number will be split over multiple lines.

sentientwaffle: Oops, that didn't occur to me!

matklad (approved):

src/devhub/devhub.js
matklad: It didn't occur to me either, it's just that when I was manually prepping the screenshots, I got the wrapping with plain spaces!

PR #2711, opened 2025-02-07
VSR/Journal: Fix misguided misdirect `fix`

sentientwaffle:
Fix a recovery correctness bug caused by a misdirected write in the WAL (discovered by the VOPR in https://github.com/tigerbeetle/tigerbeetle/pull/2677).

I recommend reviewing commit-by-commit.

## Bug

Scenario:

1. Write prepare/header `op=X view=V₁`.
2. Write prepare/header `op=X view=V₂`, but the prepare write is lost/misdirected.
3. (Checkpoint, so that `X` precedes our `op_checkpoint`, but is not yet overwritten.)
4. Crash. Recovery.
5. `op=X` is recovered with `decision=fix` (case `@M`) -- but we keep the wrong header!

## Fix

`op⌊` in the recovery table branched on `prepare.op < op_checkpoint`. This case was added to protect us from falling into `status=recovering_head` if we recover soon after checkpointing with a bunch of corrupt ops between `op_checkpoint` and `trigger_for_checkpoint(op_checkpoint)`.

But thanks to https://github.com/tigerbeetle/tigerbeetle/pull/2345, this case is not needed any more -- it is impossible for an op prior to our checkpoint to become our `op_head`.

Therefore we can merge recovery cases `@L`/`@M`, and remove the `op⌊` pattern. The former `@N` has been renamed to `@M`.

### DIFF
--- a/src/vsr/journal.zig
+++ b/src/vsr/journal.zig
@@ -1218,17 +1218,17 @@ pub fn JournalType(comptime Replica: type, comptime Storage: type) type {
         ///
         /// Recovery decision table:
         ///
-        ///   label                   @A  @B  @C  @D  @E  @F  @G  @H  @I  @J  @K  @L  @M  @N
-        ///   header valid             0   1   1   0   0   0   1   1   1   1   1   1   1   1
-        ///   header reserved          _   1   0   _   _   _   1   0   1   0   0   0   0   0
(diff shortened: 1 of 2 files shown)


matklad (approved):
I've been working on TigerBeetle for a couple of years at this point, but VOPR still gives me a sense of wonder! I am so glad that we _started_ the thing with thinking about testing!

PR #2716, opened 2025-02-10
inspect: start memory section

matklad:
The "memory" section gives a napkin math estimate for the memory usage. It is intentionally calculated manually from constants, rather than by running inits with some kind of fake allocator, to double check that we understand where the memory is going. That is, large discrepancies between what `inspect constants` prints and what we get in actual usage are possible, but it's not given that the bug is in inspect.

Start with giving the size of FreeSet.

### DIFF
--- a/src/tigerbeetle/inspect.zig
+++ b/src/tigerbeetle/inspect.zig
@@ -24,6 +24,8 @@ const SuperBlockQuorums = vsr.superblock.Quorums;
 const StateMachine = vsr.state_machine.StateMachineType(Storage, constants.state_machine_config);
 const BlockPtr = vsr.grid.BlockPtr;
 const BlockPtrConst = vsr.grid.BlockPtrConst;
+const CheckpointTrailer = vsr.CheckpointTrailerType(Storage);
+const Grid = vsr.GridType(Storage);
 const allocate_block = vsr.grid.allocate_block;
 const is_composite_key = vsr.lsm.composite_key.is_composite_key;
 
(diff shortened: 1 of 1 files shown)


src/tigerbeetle/inspect.zig
```diff
@@ -192,6 +194,29 @@ fn inspect_constants(output: std.io.AnyWriter) !void {
...
+
+    {
+        try print_header(output, 0, "free_set");
+        const hashmap_entries =
+            StateMachine.Forest.compaction_blocks_released_per_pipeline_max() +
+            Grid.free_set_checkpoints_blocks_max(datafile_size) +
+            CheckpointTrailer.block_count_for_trailer_size(vsr.ClientSessions.encode_size);
+        try output.print("{:.2}\n", .{std.fmt.fmtIntSizeBin(
+            // HashMap of block addresses.
+            stdx.div_ceil(hashmap_entries * 100, 80) * @sizeOf(u64) +
```
chaitanyabhandari: Shall we also put a small comment about `default_max_load_percentage` for the HashMap being 80? Or perhaps just use it from the standard library since its public? 
```C
pub const default_max_load_percentage = 80;
```

chaitanyabhandari:
Neat to see that the size of `blocks_released_prior_checkpoint_durability` is what we expected it to be. I actually never got around to calculating the size after the initial napkin math that I did for it _before_ I started implementing it. 😄 
```C
Memory (approximate):
datafile (on disk)  64TiB
free_set            220.05KiB
```

src/tigerbeetle/inspect.zig
matklad: Ah, excellent point! That's actually what I intended to do here, but looks like I've accidentally partially-evaluated the thing!

chaitanyabhandari (approved):

PR #2717, opened 2025-02-10
node: don't crash when a client closed concurrent

### LINKED ISSUE #2636: [tigerbeetle-node] Client destroy invocation panics thread during concurrent client method invocation
I've observed behavior I believe to be a bug considering the nature of the situation— I'll describe it below:
Invoking client.destroy while the client is concurrently attempting another process causes a thread panic.

For context— I am working on a solution to ensure that a connection from a general purpose server (where tigerbeetle-node client is present) is established successfully with a tigerbeetle server I've instantiated.

Due to a lack of connection information and retries flooding the console, I've put together a workaround which makes use of a race between `client.lookupAccounts` and a timeout.
The purpose of this race is to validate that a proper connection to the server was established in a reasonable amount of time.
If the timeout rejected (or lookupAccounts rejected) before / rather than the lookupAccounts resolving (Javascript Promises) then it would invoke client.destroy (to free up resources and end retries if it was a network error).

This invocation (of destroy) while another client method invocation was "pending" would yield:
```zsh
thread <int> panic: reached unreachable code
```
Which killed the process rather than propagating an error to the Javascript environment.

To reproduce:

> Environment Details:
MacOS (M3 Max): 15.3 Beta (24D5034f)
Node: v18.20.2
NPM: 10.5.0

1. Install the tigerbeetle-node client from npm into a project directory (I'm using v0.16.20)
```zsh
npm i tigerbeetle-node@0.16.20
```
2. _Mark the project as a module in package.json if using the sample code I provide_
__package.json__
```json
{
  "type": "module",
  "dependencies": {
    "tigerbeetle-node": "^0.16.20"
  }
}
```
3. Create a `repro.js` file with the following contents:
__repro.js__
```js
import { createClient } from 'tigerbeetle-node';

const client = createClient({
    cluster_id: 0n, // Valid filler data as no server is actually used in this example to emulate network errors.
    replica_addresses: ['3001'],
});

try {
    await Promise.race([
        client.lookupAccounts([0n]),
        new Promise((_, rej) => {
            setTimeout(() => {
                client.destroy();
                rej(
                    new Error(
                        'Failed to connect to tigerbeetle server in time.'
                    )
                );
            }, 3000);
        }),
    ]);
} catch (e) {
    console.log(e); // Expecting some form of network error either from the timeout or lookupAccounts
}
```
4. Run the `repro.js` file:
```zsh
node repro.js
```

If this is not something unique to my environment— you'll get the aforementioned log message:
```zsh
thread <int> panic: reached unreachable code
```

As of writing, my testing only finds this error when `destroy` is invoked concurrently with another pending result derived from an invocation of a client method which hasn't yet resolved.

I understand that typically you would not see this type of  implementation behavior, its likely expected (in the tigerbeetle-node client system design), that client implementations avoid concurrent method invocations and instead block and behave more synchronous / procedurally.
However when client method invocations are not blocking in their context (thennables, racing promises, etc)— this panic becomes unfitting and would be better suited as a propagated error to the Javascript environment which can be caught and handled.

matklad:
Problem: if JS calls client.destroy() while there are outstanding requests, the native part crashes with unreachable.

Cause: two bugs, actually!

_First_, `destroy` was using `napi_tsfn_abort` for cleaning-up the trampoline used to invoke the JS callback. But before that, client_deinint would actually enqueue several calls to that callback. We only should be actually destroying it once all callbacks are fired, that is, when node-side refcount reaches zero. That's exactly what release is for!

_Second_, using `release` uncovered an off-by-one error in our reference counting, with the function never released. That's because in the `init` we end up with ref-count 2, we get one from

```
1, // Number of acquires/threads that will be calling this TSFN.
```

and a redundant one from explicit acquire, which is removed.

The relevant node docs are at

<https://nodejs.org/api/n-api.html#n_api_reference_counting_of_thread_safe_functions>

I wish I didn't have to write a manual test here, and that instead we hade some sort of language-agnostic client test harness, some sort of vörtex where you'll only have to write a client-specific driver, and let the fuzzer find lifecycle bugs for you. But I'll leave that for future work!

Closes: #2636

### DIFF
--- a/src/clients/node/node.zig
+++ b/src/clients/node/node.zig
@@ -255,14 +255,15 @@ fn on_completion(
     _ = client;
     _ = timestamp;
 
-    switch (packet.status) {
-        .ok => {
-            switch (@as(Operation, @enumFromInt(packet.operation))) {
-                inline else => |op| {
-                    const event_count = @divExact(
-                        packet.data_size,
-                        @sizeOf(StateMachine.EventType(op)),
-                    );
+    const buffer_size: BufferSize = switch (@as(Operation, @enumFromInt(packet.operation))) {
+        inline else => |op| buffer_size: {
+            const event_count = @divExact(
+                packet.data_size,
+                @sizeOf(StateMachine.EventType(op)),
+            );
+
+            const result_count = switch (packet.status) {
+                .ok => result_count: {
                     const buffer: BufferType(op) = .{
                         .ptr = @ptrCast(packet),
                         .count = event_count,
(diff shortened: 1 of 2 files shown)


matklad:
Rather mystifying failure on windows, setting up a VM now!

matklad:
Ok, there's always one more bucket underneath!

batiati (approved):
🤌

Issue #2529, opened 2024-12-01
repl regression: repl no longer supports multi-line commands


ericphanson:
I tried copy-pasting
```
create_accounts id=1 code=10 ledger=700,
                id=2 code=10 ledger=700;
```
from https://docs.tigerbeetle.com/quick-start#5-create-accounts but it gives an error:
```sh
info(message_bus): connected to replica 0
TigerBeetle CLI Client 0.16.14+ff102ad
  Hit enter after a semicolon to run a command.
  Ctrl+D to exit.

Examples:
  create_accounts id=1 code=10 ledger=700 flags=linked|history,
                  id=2 code=10 ledger=700;
  create_transfers id=1 debit_account_id=1 credit_account_id=2 amount=10 ledger=700 code=10;
  lookup_accounts id=1;
  lookup_accounts id=1, id=2;
  get_account_transfers account_id=1 flags=debits|credits;
  get_account_balances account_id=1 flags=debits|credits;

> create_accounts id=1 code=10 ledger=700,
Fail near line 1, column 40:

create_accounts id=1 code=10 ledger=700,
                                        ^ Near here.

Expected key starting key-value pair. e.g. `id=1`
```
It works as two separate lines. I'm on macos fwiw.

matklad:
Thanks, this seems like it is a recent regression! I checked that it errors out on main, but does work on 0.16.1. 

cc @brson I guess this could have regressed in the PR that introduced better terminal editing for repl?

brson:
Also mentioned https://github.com/tigerbeetle/tigerbeetle/pull/2464#issuecomment-2489104777, missed at the time.

brson:
Bisected to c8ce36fece8f51e333da7917ebe11c5a02507dd0

brson:
In my review of the PR that changed this I did notice the change in behavior (https://github.com/tigerbeetle/tigerbeetle/pull/2155#issuecomment-2270001279), without further scrutiny.

brson:
Restoring the ability to put linefeeds into a single repl command is not trivial, while also keeping the ability to backarrow and insert characters, maintain command history.

Naively allowing commands to span multiple lines has some problems:

- It breaks the current logic for all arrow keys, that doesn't expect linefeeds. Probably surmountable with hard thinking.
- It raises the question of what to do if the user backarrows to the middle of the line, then presses enter, perhaps with a semicolon existing somewhere in the statement.

[...]

PR #2722, opened 2025-02-11
A few cleanups to repl docs

brson:
- Don't use multi-line repl commands. These don't currently work - cc https://github.com/tigerbeetle/tigerbeetle/issues/2529
- Don't use semicolons. They are still supported but not necessary and serve no function.
- Update quickstart with current repl output

### DIFF
--- a/src/repl.zig
+++ b/src/repl.zig
@@ -496,8 +496,7 @@ pub fn ReplType(comptime MessageBus: type, comptime Time: type) type {
                 \\  Ctrl+D to exit.
                 \\
                 \\Examples:
-                \\  create_accounts id=1 code=10 ledger=700 flags=linked|history,
-                \\                  id=2 code=10 ledger=700;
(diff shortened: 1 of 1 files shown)


batiati (commented):
Removing the line breaks reads better regardless having support to multi line statements, that's a good call.

However I'd keep the semicolons, because they will be necessary as soon we fix this regression, avoiding going back and forth in the examples and documentation.

brson:
> Removing the line breaks reads better regardless having support to multi line statements, that's a good call.
> 
> However I'd keep the semicolons, because they will be necessary as soon we fix this regression, avoiding going back and forth in the examples and documentation.

Done.

batiati (dismissed):
Thanks!!!

brson:
@batiati this needs another review.

The quick start docs changed and invalidated most of this pr. The only thing that remains is a single change to the repl help output.

batiati (approved):
>The quick start docs changed and invalidated most of this pr. The only thing that remains is a single change to the repl help output.

That’s truly designing the problem away!  :wink:

PR #2724, opened 2025-02-12
build: don't expose VSR module to dependents

rbino:
`addModule` was called also from `build_test_integration` which resulted in the VSR module built for integration tests being exposed (with hardcoded options) to dependents.

`build_test_integration` now calls `createModule` instead, resulting in a private module.

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -340,7 +340,7 @@ fn build_vsr_module(b: *std.Build, options: struct {
     vsr_options.addOption(bool, "config_aof_recovery", options.config_aof_recovery);
     vsr_options.addOption(config.HashLogMode, "hash_log_mode", options.hash_log_mode);
 
-    const vsr_module = b.addModule("vsr", .{
+    const vsr_module = b.createModule(.{
         .root_source_file = b.path("src/vsr.zig"),
     });
     vsr_module.addOptions("vsr_options", vsr_options);


matklad:
This makes sense, but I'd maybe do

```diff
diff --git a/build.zig b/build.zig
index a21517d19..bdd30e718 100644
--- a/build.zig
+++ b/build.zig
@@ -340,7 +340,7 @@ fn build_vsr_module(b: *std.Build, options: struct {
     vsr_options.addOption(bool, "config_aof_recovery", options.config_aof_recovery);
     vsr_options.addOption(config.HashLogMode, "hash_log_mode", options.hash_log_mode);
 
-    const vsr_module = b.addModule("vsr", .{
+    const vsr_module = b.createModule("vsr", .{
         .root_source_file = b.path("src/vsr.zig"),
     });
     vsr_module.addOptions("vsr_options", vsr_options);
```


At the moment, we do not intend to produce a proper Zig package out of TigerBeetle repository, so the question of what's exported is irrelevant. 

We _might_ want to export vsr at some point, but that'll require a thorough audit of what exactly we are exposing and how.

matklad:
Ah, I see the context now, https://tigerbeetle.slack.com/archives/C04375JUAMS/p1738536497929219,this is for building an external client!


🤔 I must admit, we don't really have a good story yet! Ideally, we should publish our pre-build C library (an `.a` file and an `.h` file), and let all external clients (and, actually, internal clients) to depend on that. 

That's still some way off though (in particular, we need to excise the C dependency there...).


So, we could expose the module for now, but this'll come without any guarantees. 

At that point though, I think you can more or less disregard our build.zig entirely, and create VSR module on your side?

rbino:
Ok, I was able to create the VSR module on my side with:
```zig
    const tigerbeetle_dep = b.dependency("tigerbeetle", .{});
    const vsr_mod = b.createModule(.{
        .root_source_file = tigerbeetle_dep.path("src/vsr.zig"),
    });
```

and declaring the `vsr_config` in my root

```zig
// TODO: how should we keep these in sync?
pub const vsr_options = .{
    .config_verify = true,
    .hash_log_mode = .none,
    .release = "0.16.27",
    .release_client_min = "0.15.3",
    .git_commit = null,
    .config_aof_recovery = false,
};
```

I'll change both module creations to `createModule`.

Regarding external clients, I'm currently using the Zig bindings (not the C ones) so publishing the prebuild C library would not help much there.

I guess my main question remains: is there a way to use the same VSR options all internal clients are using instead of having to keep them in sync on my side? Can the `vsr_options` be exposed in some way to external clients (and, does it make sense)? I'm still grasping if `vsr_options` would be something a client needs to change for some reason.

matklad:
>Regarding external clients, I'm currently using the Zig bindings (not the C ones) so publishing the prebuild C library would not help much there.

You _probably_ want to use the C library still: all our clients go through the C layer (although for some clients, eg, node, that's a big wonky). Sorry again that this all is too complicated at the moment! We should simplify this a bit, you are observing the sausage from inside here :P 

>I guess my main question remains: is there a way to use the same VSR options all internal clients are using instead of having to keep them in sync on my side? Can the vsr_options be exposed in some way to external clients (and, does it make sense)? I'm still grasping if vsr_options would be something a client needs to change for some reason.

Not really, or at least, not easily! The boundary we _intend_ to provide is the C client. I'd stick with copying options for now, that seems the simplest!


But if you don't want to do that at all, I'll also be fine with exposing the module here, with the understanding that that's not something that we guarantee, and something we can take back at any point!

matklad (approved):
Ok, merging, but, again, if you think we should do a public module let's do public module, I am not too hung up on this being private!

rbino:
> if you think we should do a public module let's do public module, I am not too hung up on this being private!

Given I've verified I'm able to build TigerBeetlex without the public module, I'd rather avoid forcing you to expose some kind of "public" API that is actually not guaranteed to be stable.

> You probably want to use the C library still: all our clients go through the C layer

I think the Java client is doing [something similar](https://github.com/tigerbeetle/tigerbeetle/blob/main/src/clients/java/src/client.zig#L19) to what I'm doing in TigerBeetlex. Going through the C API would probably not be a problem (Elixir NIFs are designed mainly for C interoperability), but being able to write all "native" code in Zig instead of C is a nice plus for me, purely from an ergonomic standpoint.

matklad:
Yeah, java is wonky in the same way node is! `vsr.tb_client` **is** actually our C client, but we are using it directly, instead of through the actual C API.

PR #2725, opened 2025-02-12
Various repl fixes and improvements.

brson:
These changes bring the TigerBeetle repl more inline with Python and Julia.

- Ctrl-C - previous behavior was odd when the cursor was in the middle of the buffer. Now the cursor moves to the end of the buffer, prints "^C", and starts a new command.
- Fix the logic for deduplicating history. Previous the comparison was wrong and treated "abc" and "a" as duplicates.
- Fix newlines when cursor is in the middle of the buffer. Previously if the line was longer than the screen, a newline would cause the following lines to be overwritten. This moves the cursor to the end of the buffer before emitting the newline.

Add a variety of standard navigation shortcuts:

- DEL - delete the character after the cursor
- Ctrl-B / Ctrl-F - back / forward
- Ctrl-P / Ctrl-N - history up / down
- Ctrl-A / Ctrl-E - move to start / end of input
- Ctrl-K - delete buffer after cursor
- Ctrl-L - clear screen

### DIFF
--- a/src/repl.zig
+++ b/src/repl.zig
@@ -495,9 +495,13 @@ pub fn ReplType(comptime MessageBus: type, comptime Time: type) type {
             };
 
             if (input.len > 0) {
-                if (repl.history.empty() or
-                    !std.mem.eql(u8, repl.history.tail_ptr_const().?[0..input.len], input))
-                {
+                const add_to_history = brk: {
+                    const last_entry = repl.history.tail_ptr_const() orelse break :brk true;
+                    const last_entry_str = std.mem.sliceTo(last_entry, '\x00');
+                    break :brk !std.mem.eql(u8, last_entry_str, input);
+                };
+
+                if (add_to_history) {
                     // NB: Avoiding big stack allocations below.
 
                     assert(input.len < repl_history_entry_bytes_with_nul);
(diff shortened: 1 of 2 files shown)


src/repl.zig
```diff
@@ -140,12 +140,31 @@ pub fn ReplType(comptime MessageBus: type, comptime Time: type) type {
 
                 switch (user_input) {
                     .ctrlc => {
-                        // Erase everything below the current cursor's position in case Ctrl-C was
-                        // pressed somewhere inside the buffer.
-                        try repl.terminal.print("^C\x1b[J\n", .{});
+                        // move to end of line, print "^C" and abort the command
```
cb22: Can you make this and all other comments full sentences (punctuation, grammar).

cb22 (commented):
Mostly LGTM! Just a nit about the comments, and it needs a rebase.

cb22 (approved):

cb22:
I couldn't resist adding a few others that I normally use quite often in a shell and are common too:
- Ctrl-Left / Ctrl-Right - same as Alt-B / Alt-F - move left / right by word.
- Home / End - same as Ctrl-A / Ctrl-E - go to start / end of line.

PR #2727, opened 2025-02-12
REPL Wishlist

### LINKED ISSUE #2726: repl wishlist
Just some notes about things the repl should do but doesn't. I am mostly using the Python and Julia repls for comparison.

## more nuanced Ctrl-D

Right now pressing Ctrl-D exits the repl immediately. This is not what other repls do, because Ctrl-D is also a shortcut for "delete".

Instead, iff the prompt is empty, exit. Otherwise behave like delete.

To exit while editing a buffer, user can press Ctrl-C to clear the buffer then Ctrl+D.

## other standard key combos

- Alt-B, Alt-F - move back and forward along word boundaries

raghunandanbhat:
closes #2726 

- [x] Nuanced Ctrl + D behavior 
- [x] Alt + B
- [x] Alt + F

### DIFF
--- a/src/repl.zig
+++ b/src/repl.zig
@@ -420,10 +420,54 @@ pub fn ReplType(comptime MessageBus: type, comptime Time: type) type {
                         repl.buffer.append_slice_assume_capacity(buffer_next);
                         buffer_index = repl.buffer.count();
                     },
+                    .altf => {
+                        const forward = move_forward_by_word(repl.buffer.slice(), buffer_index);
+                        terminal_screen.update_cursor_position(
+                            @as(isize, @intCast(forward - buffer_index)),
+                        );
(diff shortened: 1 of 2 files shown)


raghunandanbhat:
@brson, please review when you get a chance. thanks!

brson (changes requested):
Thanks @raghunandanbhat.

Looks good except one thing: the alt-f behavior is not the same as python/julia. In those repls, alt-f move the cursor to the index just past the end of the next word, while this implementation moves to the beginning of the next word. Can you take another look at how python (or julia) works and make this behave the same?

raghunandanbhat:
@brson fixed. please review.

brson (approved):

PR #2729, opened 2025-02-12
vsr: allow all replicas to repair during view change

chaitanyabhandari:


chaitanyabhandari:
After discussion with @sentientwaffle, we've decided to abandon this. 

The motivation behind being able to repair the write-ahead log during view changes was that it could potentially give us a great invariant -- "everything that can be repaired will be repaired", which also holds for grid blocks. However, implementing this would come at the cost of breaking some _other_ great invariants. 

Currently, we update DVC headers only when `view == log_view` (in other words, when the replica has been in normal status), which is the only time when a replica makes changes to its journal. Now, if we allow replicas to repair headers and prepares during view change, we would then have to update DVC headers more frequently such that the changes in the journal reflect in our DVC headers. In other words, if DVC headers are not updated more frequently, with our current logic for updating DVC headers, certain changes to the journal may *never* make it to DVC headers, leading to the cluster getting permanently deadlocked.  However, a key problem with updating DVC more frequently is, the *moment* you consult your journal for certain ops, you run the risk of leaking an outdated prepare into your DVC headers, and this breaks several invariants. For example, this may happen if a replica crashes, restarts, and loads an outdated, truncated op into its DVC headers. These are the invariants that are broken:

* The hash chain of the view change headers is broken (detected in view headers `ViewChangeHeadersSlice.verify`)
* Two DVCs from the same log_view may disagree on a header (detected in `DVCQuorum.verify`)

We've decided that coming up with a solution for this could significantly increase complexity of the code, and losing the aforementioned invariants would also make reasoning about correctness more difficult. Closing this PR.

PR #2732, opened 2025-02-13
RFC: replace slow web-based code review interface with git-native, offline-first solution. 

matklad:
# Review Summary

This PR adds a `git-review` command for doing offline-first reviews in git itself.

The command is _very_ rough and intentionally does only the bare minimum. I want to start using it
in anger before adding bells and whistles. Please refer to `git-review.zig` for docs aka
implementation details.

To review this particular PR:

* Clone the relevant branch (ideally, in a dedicated review worktree):

    ```
    $ git fetch
    $ git clone matklad/git-review
    ```
* Add your review comments in-code, starting each comment with `//?` marker.
* Amend the top review commit with your changes
* `git push --force-with-lease` it.

I then will pull your review, address it, resolve all comments marking them with `//? resolved.` and
will push my changes with `git push --force-with-lease` also.

Once you are happy with the state of the pull request, run

```
./zig/zig build git-review -- lgtm
```

to push a reverting commit upstream, and then leave an approving review on GitHub.


lewisdaly:
Just a minor thing I noticed - I would expect the file to be called `git_review.zig` (underscore instead of dash), based purely on the file naming conventions in the rest of the repo. Is there any reason for it to be a `-`?

matklad:
Maybe we should rename, yeah! The actual _binary_ should be named `git-review`, as that way, if you put it in your PATH, you'd be able to invoke it as `git review`.

jjmaestro:
Have you considered using / being compatible with [google/git-appraise](https://github.com/google/git-appraise)? It's a distributed code review system for git repos that uses [`git notes`](https://git-scm.com/docs/git-notes) to store the code review data.

I think it's probably a good match for what you want! Maybe you could be a Zig clone and/or integrate it into a git command like you are already doing.

My 2 cents! :)

matklad:
Yeah, I've looked at git-apprise in particular, and at the general "use git as a database" approach (https://github.com/git-bug/git-bug is another interesting project in that domain, and Gerrit can use git-notes as well). My conclusion there was that that'd be a better backend for something like GitHub style code review, but that that would only be a great fit for TigerBeetle, if there's a _really_ good implementation, as doing it yourself feels too hard. 

And git-apprise doesn't look like a great implementation:

* it feels somewhat abandoned
* I wasn't able to find integrations with popular editors. git-apprise seems to implement the right _data model_ for reviews, but it only focuses on data storage, not on presentation. There should be porcelain for `git appraise comment -f README.md -l 2 -m "Ah, so that's what this is" 1e6eb14c8014` plumbing, and I don't think there is?

I do very much hope that `jj` grows _something_ like this for the futures, as I think for [interdiff](https://gist.github.com/thoughtpolice/9c45287550a56b2047c6311fbadebed2) reviews you do want a rich code review data model.

---

*But* the key idea of this thing is that if we let go of interdiff requirements, and settle down for emulating _just_ GitHub, than there's a very low-tech solution --- not only use git as a database, but store the actual review as an actual commit with comments straight in the source code. This is much less rigid data model, with strongly-typed metadata, _but_ it gives editor integration for free.

jjmaestro:
@matklad awesome, thanks for the context!

sentientwaffle (approved):
I think there are some rough edges here, but the best way to find (and fix) them will be to actually try this flow out!

PR #2733, opened 2025-02-13
docs: cancel preview on escape

fabioarnold:
Hitting Escape should undo the previewing of search results, restoring the content that was viewed before searching.


fabioarnold:
I need to resolve some unexpected behavior in connection with the browser history first.

fabioarnold:
We can't just rely on the browser history to cancel search preview because after canceling users can still go forward in the browser history.

PR #2734, opened 2025-02-13
docs: spellchecking

fabioarnold:
Replaces the cspell shell script using Docker with [Vale](https://vale.sh) which is fetched as a static binary using Zig's package manager as part of the docs_website build.


src/docs_website/src/docs.zig
```diff
@@ -371,7 +371,12 @@ const Page = struct {
         docs: *std.Build.Step.WriteFile,
         search_index: *SearchIndex,
     ) !void {
+        const vale_step = std.Build.Step.Run.create(b, "run vale");
+        vale_step.addFileArg(website.vale_bin);
+        vale_step.addFileArg(b.path(self.path_source));
```
matklad: Hm, but this won't run on Changeling, and other things that are not part of the docs, right?

Maybe we should rather have a top-level step which runs vale on the root of the repo for all markdown files?

fabioarnold: True, that is the changelog, readme and internal docs I think (everything not part of the docs website).

I also noticed that Vale is quite slow when running on individual files (it's a large executable). So maybe I implement the spellcheck as a build option.

matklad (dismissed):

matklad:
Seems like this is going to fail:

```
 ../../CHANGELOG.md
 32:9   error  Did you really mean             Vale.Spelling 
               'loopback'?                                   
 32:48  error  Did you really mean             Vale.Spelling 
               'prepare_ok'?        
```

matklad (approved):

PR #2742, opened 2025-02-17
[tb_client] Rework the client handle

batiati:
This is (hopefully) the last major refactor I wanted to land for `tb_client` before considering the API stable.  

The main motivation is that the `tb_client` API is *almost* thread-safe: while `submit()` can be called from multiple application threads, `deinit()` requires synchronization. Worse, calling `submit()` after `deinit()` results in UB.  

To work around this, language clients currently guard the client handle with a mutex. However, this is suboptimal since, in practice, it makes the internal lock redundant!  

### This PR:  

Sorry for the rather large PR! I tried to keep it small, but the yak-shaving was inevitable.  
Best reviewed commit by commit.  

- Reworked the client handle as an intrusive struct, `ClientInterface`, which includes a VTable with function pointers to the state-machine specific implementation and the synchronization state. Since the language client is responsible for allocating it, `deinit()` can free all resources used by the client while leaving the interface itself intact, ensuring subsequent calls are safe.  

- Split `tb_client` into the idiomatic Zig API (bringing us one step closer to a proper Zig client) and `tb_client_exports`, which contains the FFI API.

- Reworked the build script.

- Updated clients.

### Next steps:  

- Implement the Logging callback in the language clients (the `tb_register_log_callback` function is still subject to API changes).
- Make the C client a first-class citizen among other clients, with documentation, examples, pre-built binaries, and, of course, a stable API.


src/clients/c/tb_client.h
```diff
@@ -249,68 +263,85 @@ typedef enum TB_PACKET_STATUS {
...
+    tb_client_t *client_out,
     const uint8_t cluster_id[16],
-    const char* address_ptr,
+    const char *address_ptr,
     uint32_t address_len,
-    uintptr_t on_completion_ctx,
-    void (*on_completion)(uintptr_t, tb_client_t, tb_packet_t*, uint64_t, const uint8_t*, uint32_t)
+    uintptr_t completion_ctx,
+    void (*completion_callback)(uintptr_t, tb_packet_t*, uint64_t, const uint8_t*, uint32_t)
 );
```
matklad, line 310: To get this to 90%, let's also pass in an allocator? This won't allow us to get rid of `libc` _immediately_, due to `@cImport`, but will keep API compatible with that. 

I suggest using Lua's version of allocator:


```
typedef void * (*lua_Alloc) (void *ud,
                             void *ptr,
                             size_t osize,
                             size_t nsize);
```

The `ud` would be `completion_ctx`, `osize` is old size, `nsize` is new size. 

https://www.lua.org/manual/5.3/manual.html

src/clients/go/pkg/native/tb_client.h
```diff
@@ -226,6 +226,20 @@ typedef enum TB_QUERY_FILTER_FLAGS {
     TB_QUERY_FILTER_REVERSED = 1 << 0,
 } TB_QUERY_FILTER_FLAGS;
 
+typedef struct tb_client_t {
+    uint64_t reserved[4];
```
matklad: Here and in the packet, let's call this `opaque` rather than `reserved`. `reserved` is supposed to be zero, and this here is used, we just don't want to explain how!

matklad: Would also be great to add a comment explaining that this is a pinned, unmovable, uncopyable struct.

src/clients/go/pkg/native/tb_client.h
```diff
@@ -226,6 +226,20 @@ typedef enum TB_QUERY_FILTER_FLAGS {
...
+} tb_client_t;
+
+typedef struct tb_packet_t {
+    void* user_data;
+    void* data;
+    uint32_t data_size;
+    uint16_t tag;
```
matklad: Maybe call this `user_tag`, to make it clear that this field is for the user?

src/clients/go/pkg/native/tb_client.h
```diff
@@ -249,68 +263,85 @@ typedef enum TB_PACKET_STATUS {
...
+    TB_LOG_DEBUG = 3,
+} TB_LOG_LEVEL;
 
 // Initialize a new TigerBeetle client which connects to the addresses provided and
 // completes submitted packets by invoking the callback with the given context.
-TB_STATUS tb_client_init(
-    tb_client_t* out_client,
+TB_INIT_STATUS tb_client_init(
+    tb_client_t *client_out,
     const uint8_t cluster_id[16],
```
matklad, line 305: wold be good to add a comment that this is little-endian

matklad (commented):
Makes _a lot_ of sense overall, but I still need a closer read here!

src/clients/c/tb_client.h
matklad: Decided not do do this: we need mmap to init io-uring anyway, so might as well use it for allocation.

src/clients/go/pkg/native/tb_client.h
batiati: ~Humm, I think the python implementation with a `dict` holding the `tb_client` can't guarantee stability. I'll confirm that.~

*EDIT:* That's fine; even if the `dict` instance moved values during resize, it holds just the reference to a `Client`.

src/clients/c/tb_client.h
```diff
@@ -226,6 +226,26 @@ typedef enum TB_QUERY_FILTER_FLAGS {
...
+// This struct must be "pinned" (not copyable or movable), as its address must remain stable
+// throughout the lifetime of the request.
+typedef struct tb_packet_t {
+    void* user_data;
+    void* data;
+    uint32_t data_size;
+    uint16_t user_tag;
+    uint8_t operation;
+    uint8_t status;
+    uint8_t reserved[32];
```
matklad: This one also wants to be opaque?

src/clients/c/tb_client.h
```diff
@@ -249,68 +269,87 @@ typedef enum TB_PACKET_STATUS {
...
+// been closed by `tb_client_deinit`.
+TB_CLIENT_STATUS tb_client_deinit(
+    tb_client_t *client
+);
+
+// Registers or unregisters the application log callback.
+TB_REGISTER_LOG_CALLBACK_STATUS register_log_callback(
+    void (*callback)(TB_LOG_LEVEL, const uint8_t*, uint32_t),
+    bool debug
 );
```
matklad, line 353: Let's maybe leave it to the follow up, but it seems that the current logging API doesn't spark joy:

* I think we hit UB when we access `logging.callback` and `logging.debug` outside of the mutex. The `debug` should be an atomic enum, which communicates the log level, including "is custom function set?".
* This is needlessly global state, it should be per client.
* The semantics of `register` is weird. If you try to replace the function, you get an error, but if you _clear_ and than set, than that's OK.  I think it's best to configure a per-client function _once_, by passing it to the init, and leave any dynamic switching logic to user's code. 
* As a rule of thumb, `bool` should not be used at the API level. We should use `enum LogLevel` here, even if we only support two levels now. 

But, yeah, this is mostly unrelated to the refactor here.

src/tigerbeetle/libtb_client.zig
matklad, line 1: I'd maybe keep this under `src/clients/c`? Took me a while to find where the exports happen.

src/tigerbeetle/libtb_client.zig
```diff
@@ -0,0 +1,31 @@
+//! Entry point for exporting the `tb_client` library.
+//! Used by language clients that rely on the shared or static library exposed by `tb_client.h`.
+//! For an idiomatic Zig API, use `vsr.tb_client` directly instead.
```
matklad, line 3: FWIW, I'd say that _all_ clients should go through the static or dynamic library (so, `.zig -> .h -> .zig` sandwich), just so that we only have one canonical interface. But no super strong opinion here.

build.zig
```diff
@@ -985,14 +1005,15 @@ fn build_go_client(
 
         const lib = b.addStaticLibrary(.{
             .name = "tb_client",
```
matklad, line 1007: We have 8 occurrences of `.name = "tb_client",` in build.zig, so we might as well return from `build_c_client`?

```
fn build_c_client(
) []struct {
   target: []const u8, 
   static_library: std.Build.LazyPath, 
   dynamic_library: std.Build.LazyPath, 
}
```


But can leave to follow up

matklad: This should also improve compile times quite a bit I think, currently we are redundantly rebuilding the same lib multiple times I think!

src/clients/c/tb_client/context.zig
```diff
@@ -669,19 +694,74 @@ pub fn ContextType(
...
                 result.reply.ptr,
                 @intCast(result.reply.len),
             );
         }
 
-        fn get_context(implementation: *ContextImplementation) *Context {
-            return @alignCast(@fieldParentPtr("implementation", implementation));
-        }
+        /// Functions called by `ClientInterface`, which is thread-safe.
+        const vtable = ClientInterface.VTable{
```
matklad: Nit, I but I'd maybe expressed this as

```diff
diff --git a/src/clients/c/tb_client/context.zig b/src/clients/c/tb_client/context.zig
index ac929af29..56bca2124 100644
--- a/src/clients/c/tb_client/context.zig
+++ b/src/clients/c/tb_client/context.zig
@@ -250,7 +250,11 @@ pub fn ContextType(
             };
             errdefer context.client.deinit(context.allocator);
 
-            ClientInterface.init(client_out, context, &Context.vtable);
+            ClientInterface.init(client_out, context, comptime &.{
+                .submit_fn = vtable_submit_fn,
+                .completion_context_fn = vtable_completion_context_fn,
+                .deinit_fn = vtable_deinit_fn,
+            });
             context.interface = client_out;
             context.submitted = .{
                 .name = null,
@@ -288,6 +292,59 @@ pub fn ContextType(
             };
         }
 
+        fn vtable_submit_fn(context: *anyopaque, packet_extern: *Packet.Extern) void {
+            const self: *Context = @ptrCast(@alignCast(context));
+            assert(self.signal.status() == .running);
+
+            // Packet is caller-allocated to enable elastic intrusive-link-list-based
+            // memory management. However, some of Packet's fields are essentially private.
+            // Initialize them here to avoid threading default fields through FFI boundary.
+            const packet: *Packet = packet_extern.cast();
+            packet.* = .{
+                .user_data = packet_extern.user_data,
+                .operation = packet_extern.operation,
+                .data_size = packet_extern.data_size,
+                .data = packet_extern.data,
+                .user_tag = packet_extern.user_tag,
+                .status = .ok,
+                .next = null,
+                .batch_next = null,
+                .batch_tail = null,
+                .batch_size = 0,
+                .batch_allowed = false,
+                .phase = .submitted,
+            };
+
+            // Enqueue the packet and notify the IO thread to process it asynchronously.
+            self.submitted.push(packet);
+            self.signal.notify();
+        }
+
+        fn vtable_completion_context_fn(context: *anyopaque) usize {
+            const self: *Context = @ptrCast(@alignCast(context));
+            return self.completion_context;
+        }
+
+        fn vtable_deinit_fn(context: *anyopaque) void {
+            const self: *Context = @ptrCast(@alignCast(context));
+            assert(self.signal.status() == .running);
+
+            self.signal.stop();
+            self.thread.join();
+
+            assert(self.submitted.pop() == null);
+            assert(self.pending.pop() == null);
+
+            self.io.cancel_all();
+
+            self.signal.deinit();
+            self.client.deinit(self.allocator);
+            self.message_pool.deinit(self.allocator);
+            self.io.deinit();
+
+            self.allocator.destroy(self);
+        }
+
         fn tick(self: *Context) void {
             if (self.eviction_reason == null) {
                 self.client.tick();
@@ -703,66 +760,6 @@ pub fn ContextType(
             );
         }
 
-        /// Functions called by `ClientInterface`, which is thread-safe.
-        const vtable = ClientInterface.VTable{
-            .submit_fn = struct {
-                fn submit_fn(context: *anyopaque, packet_extern: *Packet.Extern) void {
-                    const self: *Context = @ptrCast(@alignCast(context));
-                    assert(self.signal.status() == .running);
-
-                    // Packet is caller-allocated to enable elastic intrusive-link-list-based
-                    // memory management. However, some of Packet's fields are essentially private.
-                    // Initialize them here to avoid threading default fields through FFI boundary.
-                    const packet: *Packet = packet_extern.cast();
-                    packet.* = .{
-                        .user_data = packet_extern.user_data,
-                        .operation = packet_extern.operation,
-                        .data_size = packet_extern.data_size,
-                        .data = packet_extern.data,
-                        .user_tag = packet_extern.user_tag,
-                        .status = .ok,
-                        .next = null,
-                        .batch_next = null,
-                        .batch_tail = null,
-                        .batch_size = 0,
-                        .batch_allowed = false,
-                        .phase = .submitted,
-                    };
-
-                    // Enqueue the packet and notify the IO thread to process it asynchronously.
-                    self.submitted.push(packet);
-                    self.signal.notify();
-                }
-            }.submit_fn,
-            .completion_context_fn = struct {
-                fn completion_context_fn(context: *anyopaque) usize {
-                    const self: *Context = @ptrCast(@alignCast(context));
-                    return self.completion_context;
-                }
-            }.completion_context_fn,
-            .deinit_fn = struct {
-                fn deinit_fn(context: *anyopaque) void {
-                    const self: *Context = @ptrCast(@alignCast(context));
-                    assert(self.signal.status() == .running);
-
-                    self.signal.stop();
-                    self.thread.join();
-
-                    assert(self.submitted.pop() == null);
-                    assert(self.pending.pop() == null);
-
-                    self.io.cancel_all();
-
-                    self.signal.deinit();
-                    self.client.deinit(self.allocator);
-                    self.message_pool.deinit(self.allocator);
-                    self.io.deinit();
-
-                    self.allocator.destroy(self);
-                }
-            }.deinit_fn,
-        };
-
         fn operation_from_int(op: u8) ?StateMachine.Operation {
             inline for (allowed_operations) |operation| {
                 if (op == @intFromEnum(operation)) {
```

just to have fewer deep indents

src/clients/c/tb_client/packet.zig
```diff
@@ -35,33 +32,6 @@ pub const Packet = extern struct {
...
-
-            return self.fifo.pop();
-        }
-
-        /// Not thread safe, should be called only by the consumer thread.
-        pub fn empty(self: *const SubmissionQueue) bool {
-            return self.fifo.count == 0;
-        }
-    };
-
```
matklad, line 64: 🎉

src/clients/c/tb_client/context.zig
batiati: Cool, I like the `vtable_` prefix on the function's name!

src/tigerbeetle/libtb_client.zig
batiati: Not _all_ clients, for example, the Java and Node clients use the Zig idiomatic API to build their own JNI and NAPI modules.

src/clients/c/tb_client.h
batiati: Agreed! Let me add this bullet point on "next steps", as this hook still needs to be implemented on most language clients.

src/clients/c/tb_client.h
```diff
@@ -230,7 +230,7 @@ typedef enum TB_QUERY_FILTER_FLAGS {
 // This struct must be "pinned" (not copyable or movable), as its address must remain stable
 // throughout the lifetime of the client instance.
 typedef struct tb_client_t {
-    uint64_t opaque[4];
+    uint64_t opaque[3];
```
matklad: I'd maybe keep it as `4`, to have some extra bytes, and to be a power-of-two

matklad (commented):
LGTM, just some optional nits!

src/tigerbeetle/libtb_client.zig
batiati: Not a strong opinion here, but for me, having all entry points in `src/tigerbeetle/` makes it easier to identify which artifacts the "TigerBeetle" product produces, especially since `libtb_client` isn't just a pure VSR client.

src/clients/c/tb_client.h
batiati: I'm wondering ... could we use this extra byte to somehow validate the struct? Like a hash or a magic number, so we return `CLIENT_INVALID` instead of SEGFAULT'ng if the user pass a garbage pointer.

src/tigerbeetle/libtb_client.zig
matklad: Yeah, that's exactly my point, I think Java&Node should _also_ go through the same C API as everyone else. I don't think we win _that_ much from the API being idiomatic Zig, as idiomatic C is also quite nice to use from Zig. And it is quite nice to have just one facade.

src/clients/c/tb_client.h
matklad: Brilliant! Could write `0xBEE71E` there and assert! 

We should also write it at the _end_ of `init`, so that, if init bails part-way through, but the user doesn't handle the error and tries to use it, we'll still assert!

build.zig
batiati: I'll do in a follow up PR, as it may need a deeper refactor. 
As `build_c_client` builds for multiple targets, static and dynamic libraries, it's not easily reusable in build steps like `vortex` or `c_sample`.

matklad (approved):

PR #2747, opened 2025-02-18
docs: fix node samples deprecated behavior

NikitaIT:
1. `assert.equal` is deprecated since node v9
  - v9 https://nodejs.org/docs/latest-v9.x/api/assert.html#assert_assert_equal_actual_expected_message
  - latest https://nodejs.org/api/assert.html#assertequalactual-expected-message
2. `process` must be imported from `node:process` this is important for environments like deno and recommended for nodejs

It's probably better to rewrite this code in ES modules; new projects usually don't use requirejs.

### DIFF
--- a/src/clients/node/README.md
+++ b/src/clients/node/README.md
@@ -24,6 +24,7 @@ Now, create `main.js` and copy this into it:
 ```javascript
 const { id } = require("tigerbeetle-node");
 const { createClient } = require("tigerbeetle-node");
+const process = require("process");
 
 console.log("Import ok!");
 ```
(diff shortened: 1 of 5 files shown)


matklad:
LGTM, needs the following diff to get the tests passing:

```diff
diff --git a/src/clients/node/README.md b/src/clients/node/README.md
index d9dcdc1d6..43be520b6 100644
--- a/src/clients/node/README.md
+++ b/src/clients/node/README.md
@@ -24,6 +24,7 @@ Now, create `main.js` and copy this into it:
 ```javascript
 const { id } = require("tigerbeetle-node");
 const { createClient } = require("tigerbeetle-node");
+const process = require("process");
 
 console.log("Import ok!");
 ```

```

(you can run node tests with `./zig/zig build scripts -- ci --language=node`)

matklad (approved):

PR #2753, opened 2025-02-20
docs: fix `SecurityError` exception in search

fabioarnold:
When the `section.path` was empty we were creating URLs with double forward slashes such as: `"//"` and `"//#tigerbeetle"`. This lead to the following exception being thrown when trying to use these URLs with the `history.pushState` and `history.replaceState` APIs:

```
SecurityError: A history state object with URL 'http:' cannot be created in a document with origin 'http://localhost:8000' and URL 'http://localhost:8000/'.
```

Bonus: fix flashing search box while navigating by using CSS media query instead of JS to unhide the search box.

### DIFF
--- a/src/docs_website/assets/js/search.js
+++ b/src/docs_website/assets/js/search.js
@@ -109,7 +109,8 @@ function onSearchInput() {
     const summary = document.createElement("summary");
     details.appendChild(summary);
     summary.pageIndex = group.pageIndex;
-    summary.href = urlPrefix + "/" + group.hits[0].section.path + "/";
+    summary.href = urlPrefix + "/"
+    if (group.hits[0].section.path) summary.href += group.hits[0].section.path + "/";
     const p = document.createElement("p");
     summary.appendChild(p);
     p.innerText = pages[group.pageIndex].title;
@@ -125,7 +126,9 @@ function onSearchInput() {
         e.preventDefault();
         selectResult(a);
       });
-      a.href = urlPrefix + "/" + result.section.path + "/" + result.section.hash;
+      a.href = urlPrefix + "/";
+      if (result.section.path) a.href += result.section.path + "/";
+      a.href += result.section.hash;
       a.pageIndex = result.section.pageIndex;
       const h3 = document.createElement("h3");
       a.appendChild(h3);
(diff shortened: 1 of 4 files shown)


src/docs_website/assets/js/search.js
```diff
@@ -125,7 +126,9 @@ function onSearchInput() {
         e.preventDefault();
         selectResult(a);
       });
-      a.href = urlPrefix + "/" + result.section.path + "/" + result.section.hash;
+      a.href = urlPrefix + "/";
+      if (result.section.path) a.href += result.section.path + "/";
+      a.href += result.section.hash;
       a.pageIndex = result.section.pageIndex;
```
matklad, line 134: Should we have our own assert that the URL is valid&simplified?

fabioarnold: You mean after constructing `a.href` downstream in search.js or when generating the page paths in Zig?

matklad: Why not both? ^^

In general it's good to assert both when you produce a value, and when you consume a value: https://tigerbeetle.com/blog/2023-12-27-it-takes-two-to-contract/


And, as a rule of thumb, if there's some validity constraint, like absence of double-slashes, we _should_ assert it at least once. We've recently added similar asserts to file_checker:

https://github.com/tigerbeetle/tigerbeetle/blob/d8dc9bd3bd4add6476f29b0348895e466a3f879d/src/docs_website/src/file_checker.zig#L190-L195

fabioarnold: Yeah, this is nice on the Zig part.

This particular problem originates from normalizing our page paths from `"."` to `""` (changing the contract on the Zig side). So yeah, it wasn't nice that we previously relied on URLs like `"/./"` on the JS side. I discovered it through runtime testing (history API was throwing the exception).

If we assert at that particular place, the search result construction will fail intermediately on encountering an invalid URL which leads to a bad user experience. We will discover this only through runtime testing again. So I don't really see the benefit.

We should statically make sure that we only generate valid and simplified page paths (which you already did).

matklad: I think user experience would be fine! The moment someone hits this, they'll open an issue, and we'll fix this. We have a bunch of asserts in https://github.com/tigerbeetle/tigerbeetle/blob/main/src/devhub/devhub.js, and they don't fire because we've fixed all the issues. 

It _is_ customary for front-end to do "progressive degradation", where there's a bunch of errors in console, but the thing still basically works. But for us, I think we should apply TigerStyle here, _escalate_ those errors so that they produce an un-ignorable user-visible popup, and than make sure that the code is at zero bugs.

matklad (approved):

PR #2754, opened 2025-02-20
docs: add zig build docs command

matklad:
To have a `cd`! I tried using `--build-root` here, but hit https://github.com/ziglang/zig/issues/22956


fabioarnold (approved):
Works as expected

PR #2755, opened 2025-02-20
stdx: falky test

matklad:
On Windows, the two durations can be the same!

https://github.com/tigerbeetle/tigerbeetle/actions/runs/13435237543/job/37535934663#step:5:15

### DIFF
--- a/src/stdx.zig
+++ b/src/stdx.zig
@@ -876,7 +876,18 @@ test "Instant/Duration" {
     const instant_1 = Instant.now();
     const instant_2 = Instant.now();
     assert(instant_1.duration_since(instant_2).nanoseconds == 0);
-    assert(instant_2.duration_since(instant_1).nanoseconds > 0);
+    assert(instant_2.duration_since(instant_1).nanoseconds >= 0);
+
+    if (builtin.os.tag == .linux) {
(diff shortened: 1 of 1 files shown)


src/stdx.zig
```diff
@@ -876,7 +876,7 @@ test "Instant/Duration" {
     const instant_1 = Instant.now();
     const instant_2 = Instant.now();
     assert(instant_1.duration_since(instant_2).nanoseconds == 0);
-    assert(instant_2.duration_since(instant_1).nanoseconds > 0);
+    assert(instant_2.duration_since(instant_1).nanoseconds >= 0);
```
batiati, line 879: Oh! That's so unfortunate :cry:

The problem with `>= 0` is that we can't really assert that we are measuring a duration or always zero!
I'd suggest a sleep gated by `if (builtin.os.tag == .windows)`.

batiati: Or maybe, if  the goal is to assert `duration_since` we could just pass a dummy `instant_2` instead of `now()`.

matklad: Good idea about using a dummy instance of instant, implemented!

batiati (approved):

PR #2756, opened 2025-02-20
fix: 2 broken links on README.md

lewisdaly:
I noticed that the very first and last links on our `README.md` were broken!

Updated them based on taking a look at where the docs are now, however please let me know if my educated guesses were wrong.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # tigerbeetle
 
-*TigerBeetle is the financial transactions database designed for mission critical safety and performance to power the next 30 years of [OLTP](https://docs.tigerbeetle.com/concepts/oltp).*
+*TigerBeetle is the financial transactions database designed for mission critical safety and performance to power the next 30 years of [OLTP](https://docs.tigerbeetle.com/about/oltp).*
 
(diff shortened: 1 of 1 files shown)


matklad:
Yeah, we'll deploy the new docs on Monday, where this'll get broken in the opposite direction. But let's merge this and fix back on Monday!

matklad (approved):

matklad:
BTW, @lewisdaly I think it would excellent if you could review the new docs which we landed earlier this week, and which will go live on Monday! They aren't deployed yet, but you can build them locally as:

```
$ ./zig/zig build docs
$ python3 -m http.server -d src/docs_website/zig-out/
```

If you are on Mac, you'll also need

```
$ chmod +x ~/.cache/zig/p/1220418afe2846224487326841b0228b1c44e77bb8440c96190fcc3189a2df1688eb/vale
$ chmod +x ~/.cache/zig/p/1220c2506a07845d667e7c127fd0811e4f5f7591e38ccc7fb4376450f3435048d87a/bin/pandoc
```

PR #2760, opened 2025-02-21
vsr: simplify the idiom around flush_back 

matklad:
There's not material reason why we need a defer, it's only to underscore
that flusing is _usually_ the last thing to do.

Replace defers with more direct returns when possible.

### DIFF
--- a/src/vsr/replica.zig
+++ b/src/vsr/replica.zig
@@ -7850,6 +7850,7 @@ pub fn ReplicaType(
                 // This is safe only because the primary can verify against the prepare checksum.
                 if (self.journal.header_with_op(op)) |header| {
                     self.send_prepare_ok(header);
+                    if (self.loopback_queue != null) assert(self.journal.has_prepare(header));
                     self.flush_loopback_queue();
                 }
             }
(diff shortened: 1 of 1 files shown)


chaitanyabhandari:
@matklad Right now, the way we invoke `flush_loopback_queue` seems to be a mixture of these two patterns:
```C
            self.send_prepare_ok(message.header);
            return self.flush_loopback_queue();
```

```C
            self.send_prepare_ok(message.header);
            self.flush_loopback_queue();
```

Wondering if it would make sense to just use the second pattern (send message and invoke `flush_loopback_queue` _right_ after) everywhere? Or is there some other motivation behind the first pattern?

matklad:
`flush_loopback_queue` will call `on_message`, and that would typically be a recursion:

- you receive some message
- that causes you to send another message
- including a message to yourself
- so you call flush_loopback_queue
- and end up in on_message while still not exiting from the previous on_message

This is tricky control flow, and we want to minimize that. One way to simply it is to make sure that the recursive `flush` call is the last thing the replica is doing. The `return` underscores that. 


The two cases where we _don't_ do return are really exceptional. Ideally, we'd do return there as well, but we can't, as there's some work we need to do _after_ the message is handled.

chaitanyabhandari (approved):

Issue #2739, opened 2025-02-14
Elevated latencies during single-node faults


aphyr:
In TigerBeetle 0.16.17 through 0.16.27, single-node failures are frequently associated with higher latencies across all clients--often from three to five orders of magnitude. For instance, consider [this test](https://s3.amazonaws.com/jepsen.io/analyses/tigerbeetle-0.16.11/20250214T110802.529-single-node-latency-jump-3.zip). We killed one of three nodes, and saw latencies jump from 1-50 ms up to ~100 seconds. Latencies remained elevated until the node was restarted, almost a thousand seconds later.

![Image](https://github.com/user-attachments/assets/12b10263-0dc2-4be2-93da-1fd85894bd4d)

High latencies may recover spontaneously after tens to hundreds of seconds, or persist for thousands of seconds.

This may involve the failure of a non-leader node. It also seems more likely in smaller clusters, rather than large ones.

matklad:
Yup, we were handling node failure quite suboptimally! Available in the mathematical sense of eventually making progress, but not in the practical sense of finishing our job before the client gives up and leaves. It also wasn't a single specific issue, but rather the result of us mostly not having looked into the area. 

We landed a bunch of work recently to significantly improve the situation:

* https://github.com/tigerbeetle/tigerbeetle/pull/2820 added "performance mode" to our deterministic simulation testing, allowing us to actually _see_ the problem without having to spin up an entire cluster. 
* Then we have a whole bunch of improvements and outright bug fixes to our repair protocol:
  * https://github.com/tigerbeetle/tigerbeetle/pull/2957 is the key one, which paces repair not to a fixed timeout, but to the rate of progress of the cluster, 
  * https://github.com/tigerbeetle/tigerbeetle/pull/2972, https://github.com/tigerbeetle/tigerbeetle/pull/2950, https://github.com/tigerbeetle/tigerbeetle/pull/2950, https://github.com/tigerbeetle/tigerbeetle/pull/2628, https://github.com/tigerbeetle/tigerbeetle/pull/2626, https://github.com/tigerbeetle/tigerbeetle/pull/2625, https://github.com/tigerbeetle/tigerbeetle/pull/2623, https://github.com/tigerbeetle/tigerbeetle/pull/2620, https://github.com/tigerbeetle/tigerbeetle/pull/2613 --- various fixes and optimizations, many of them predate 0.16.27
[...]

PR #2761, opened 2025-02-21
VSR: Reverse ring replication for odd ops

matklad:
Alternate ring replication direction such that, if a single replica is down, the replicas after it learn that they are missing a prepare, and can repair it.

We originally reverted this commit as it wasn't playing well with our repair, which insisted on everything being done strictly in order. We have since relaxed our repair to be more concurrent:

* https://github.com/tigerbeetle/tigerbeetle/pull/2628 
* https://github.com/tigerbeetle/tigerbeetle/pull/2593

We don't yet have a deterministic performance test here yet, so I used the following script to test the behavior

```ts
#!/usr/bin/env -S deno run --allow-all
import $ from "jsr:@david/dax";

$.setPrintCommand(true);
const working_directory = "zig-out/cluster";
const replica_count = 6;
const replica_indexes = [...Array(replica_count)].map((_, index) => index);
const cluster = 0;
const addresses = [...Array(replica_count)].map((_, index) => `${3000 + index}`)
    .join(",");

await $`./zig/zig build -Drelease`;
try {
    await Deno.remove(working_directory, { recursive: true });
} catch {

}
await Deno.mkdir(working_directory);

await Promise.all(replica_indexes.map((replica) => {
    return $`./tigerbeetle format
            --cluster=${cluster} --replica=${replica} --replica-count=${replica_count}
            ${working_directory}/${cluster}_${replica}.tigerbeetle`;
}));

const dead_replica = 2;
const processes = [];
for (let replica = 0; replica < replica_count; replica += 1) {
    if (replica == dead_replica) continue;
    const process = $`./tigerbeetle start
            --addresses=${addresses}
            ${working_directory}/${cluster}_${replica}.tigerbeetle
            2> ./${working_directory}/${cluster}_${replica}.log`.spawn();
    processes.push(process);
}

await $`./tigerbeetle benchmark --print-batch-timings
         --id-order=random --seed=3 --transfer-batch-size=1000 --transfer-count=200_000
         --addresses=${addresses}
         --clients=4`;

console.log("ok.");
```

In all-healthy cluster, benchmark finishes in about 7 seconds for me, with or without bidirectional replication. 

If a single replica is down:

- with unidirectional, it takes either 7 seconds still (last replica is down) or takes a lot longer, with individual requests taking up to 12 seconds
- with bidirectional, it takes 30 seconds regardless of which replica is down.

### DIFF
--- a/src/testing/cluster/state_checker.zig
+++ b/src/testing/cluster/state_checker.zig
@@ -114,8 +114,9 @@ pub fn StateCheckerType(comptime Client: type, comptime Replica: type) type {
 
             if (replica.status != .recovering_head) {
                 const head_max = &state_checker.replica_head_max[replica_index];
-                const wal_headers = replica.superblock.storage.wal_headers();
-                const head_max_journal = wal_headers[head_max.op % constants.journal_slot_count];
+                const wal_prepares = replica.superblock.storage.wal_prepares();
+                const head_max_journal =
+                    wal_prepares[head_max.op % constants.journal_slot_count].header;
 
                 assert(replica.view > head_max.view or
                     (replica.view == head_max.view and (replica.op >= head_max.op or
--- a/src/vsr/replica.zig
+++ b/src/vsr/replica.zig
@@ -3182,8 +3182,7 @@ pub fn ReplicaType(
                 return self.commit_pipeline();
             }
 
-            // See replicate().
-            const ring_direction: i16 = if (prepare.message.header.op % 2 == 0) 1 else -1;
+            const ring_direction = replicate_direction(prepare.message);
 
             // The list of remote replicas yet to send a prepare_ok:
             var waiting: [constants.replicas_max]u8 = undefined;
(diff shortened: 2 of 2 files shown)


sentientwaffle (dismissed):

sentientwaffle:
This should help with https://github.com/tigerbeetle/tigerbeetle/issues/2739

sentientwaffle:
Wait, I'm confused!

> If a single replica is down:
>
> - with unidirectional, it takes either 7 seconds still (last replica is down) or takes a lot longer, with individual requests taking up to 12 seconds
> - with bidirectional, it takes 30 seconds regardless of which replica is down.

Is this saying that bidirectional is _slower_?

matklad:
Actually, that's not precise, let me describe it fully. Let's say we have replicas 0 thru 5, with 0 being the primary. Then, depending on the replica down, we'll get the following performance table:


|💀 |← |↔ |
|--|--|---|
|R1|∞ |30s|
|R2|∞ |30s|
|R3|7s|7 s|
|R4|7s|30s|
|R5|7s|30s|

(here, ∞ means not the literal infinity, but "too long to wait for test to finish", with _individual_ batch latencies larger than 10 seconds)

This is precisely what's expected: with unidirectional, the cluster basically doesn't notice if the last couple of replicas are eaten by the monsters, but a problem at the Strat of the ring leads to pathological behavior.

With bidirectional _both_ sides of the ring are equally important, so latency is elevated, unless the replica precisely the opposite of primal dies --- in that case, we still have replication quorum in either direction. 

(table from memory, but let me re-run the experiments just to confirm)

matklad:
Yup, the table above is correct!

matklad:
Will merge on Monday after the release, just in case.

matklad:
VOPR lite failure is false positive, I've added the fix here to keep the seed on the branch where it actually fails.

sentientwaffle (approved):

PR #2764, opened 2025-02-22
Spring cleaning: make sure that MessageBus does all io through IO

matklad:
MessageBus contained a bunch of platform specific code for working with sockets. Move all this code to IO, to make MessageBus a logical abstraction over byte streams, rather than a mixture of logical and physical. 

There's no _immediate_ benefits here, but this opens up MessageBus for fuzzing, as now you can just substitute its IO with a fake implementation.

### DIFF
--- a/src/message_bus.zig
+++ b/src/message_bus.zig
@@ -1,8 +1,6 @@
 const std = @import("std");
-const builtin = @import("builtin");
 const assert = std.debug.assert;
 const mem = std.mem;
-const os = std.os;
 const posix = std.posix;
 
 const constants = @import("constants.zig");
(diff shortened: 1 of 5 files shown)


src/message_bus.zig
```diff
@@ -1,12 +1,8 @@
 const std = @import("std");
-const builtin = @import("builtin");
 const assert = std.debug.assert;
 const mem = std.mem;
-const os = std.os;
 const posix = std.posix;
 
-const is_linux = builtin.target.os.tag == .linux;
-
```
matklad, line 9: Killing these two imports is the point! We still need posix for 

```
            const fd = try io.open_socket(address.any.family, posix.SOCK.STREAM, posix.IPPROTO.TCP);
```

I considered sinking that inside IO, so that it has `open_tcp_socket`, but figured that it's fine as-is.

batiati (approved):

PR #2768, opened 2025-02-25
docs: only spell check files checked into git

fabioarnold:
We want to restrict spell checking to files checked into git so we avoid checking build artifacts and failing the release CI: https://github.com/tigerbeetle/tigerbeetle/actions/runs/13502735251/job/37724965046

Downside of this is if we're working on a new Markdown file it is only spell checked until it's committed. However, in this case the CI will still catch it.

### DIFF
--- a/src/docs_website/build.zig
+++ b/src/docs_website/build.zig
@@ -25,7 +25,11 @@ pub fn build(b: *std.Build) !void {
 
     const check_spelling = std.Build.Step.Run.create(b, "run vale");
     check_spelling.addFileArg(vale_bin);
-    check_spelling.addArg("../.."); // Check all Markdown files in the repository.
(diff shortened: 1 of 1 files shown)


src/docs_website/build.zig
```diff
@@ -25,7 +25,11 @@ pub fn build(b: *std.Build) !void {
 
     const check_spelling = std.Build.Step.Run.create(b, "run vale");
     check_spelling.addFileArg(vale_bin);
-    check_spelling.addArg("../.."); // Check all Markdown files in the repository.
+    const md_files = try exec_stdout(b.allocator, &.{ "git", "ls-files", "../../**/*.md" });
```
matklad: Use b.run here:

https://github.com/tigerbeetle/tigerbeetle/blob/ba6828b5072ac8bd8ebbe68bd8ea8ae6157e4930/build.zig#L129

matklad (approved):

PR #2769, opened 2025-02-25
docs: wordsmith concepts

matklad:


matklad:
Thanks @MarinaWoo !

docs/concepts/oltp.md
```diff
@@ -5,89 +5,75 @@ could be payments, sales, car sharing rides, game scores, or API usage.
...
+  In January 2025 alone, it processed [16.9 billion payments.](https://www.npci.org.in/what-we-do/upi/product-statistics)
+- Cleaner energy and smart metering man energy is being traded by the kilowatt-hour.
+  Customer billing is every 15 or 30 minutes rather than at the end of the month.
 - Serverless APIs charge for usage by the second or per-request, rather than per month. (Today,
-  serverless billing at scale is often implemented using MapReduce, which makes it difficult or
-  impossible to offer customers real-time spending caps.)
+  serverless billing at scale is often implemented using [MapReduce](https://en.wikipedia.org/wiki/MapReduce).
+  This makes it difficult or impossible to offer customers real-time spending caps.)
+
+OLGP databases already struggle to keep up.
```
matklad, line 22: Removed the reference to contention here, as we are explains _why_ they struggle in the following two paragraphs.

docs/concepts/oltp.md
```diff
@@ -5,89 +5,75 @@ could be payments, sales, car sharing rides, game scores, or API usage.
...
-    [Visa Fact Sheet](https://usa.visa.com/dam/VCOM/global/about-visa/documents/aboutvisafactsheet.pdf)
-    reports that the network processed 276.3 billion transactions in 2023.
+You can only do as much business as your database supports. You need a core OLTP database capable of
+handling your transactions on your busiest days. And for decades to come.
 
-[^2]:
-    Visa.
-    [_A deep dive on Solana, a high performance blockchain network_](https://usa.visa.com/solutions/crypto/deep-dive-on-solana.html). 2023.
+TigerBeetle is designed to handle **1 million transactions per second**, to remove the risk of your
+business outgrowing your database.
```
matklad, line 74: Removed this whole TPS contest, for two reasons: 

- We ourselves don't yet have a clear benchmark which we can point to, as a measure of TPS. 
- The public numbers we use feel a bit dodge (Solana?), and they are going to change quickly, but the concepts should stand for eternity!

Instead, I worked-in the `risk` idea, as I think it is actually a rather key motivation for performance.

docs/concepts/README.md
```diff
@@ -1,13 +1,12 @@
 # Concepts
 
-This section is for anyone evaluating TigerBeetle, learning about it, or just curious.  It focuses
-on the big picture --- what is the problem that TigerBeetle solves and why it looks nothing like a
+This section is for anyone evaluating TigerBeetle, eager to learn about it, or curious. It focuses
+on the big picture and problem that TigerBeetle solves. As well as why it looks nothing like a
```
martijngonlag: I think this should be plural since TigerBeetle doesn't solve only one problem. 

```suggestion
on the big picture and problems that TigerBeetle solves. As well as why it looks nothing like a
```

docs/concepts/oltp.md
```diff
@@ -5,89 +5,75 @@ could be payments, sales, car sharing rides, game scores, or API usage.
 
 ## The World is Becoming More Transactional
 
-For the last 20-30 years, OLTP has been handled by general purpose databases like PostgreSQL, MySQL,
-and SQLite, which we refer to as Online General Purpose (OLGP) databases.
+For 20-30 years, general purpose databases like PostgreSQL, MySQL, and SQLite handled OLTP. We refer
```
martijngonlag: I think it would make more sense to change "For 20-30 years" here to "Historically". Time is only moving forward! What do you say @matklad?

docs/concepts/oltp.md
```diff
@@ -5,89 +5,75 @@ could be payments, sales, car sharing rides, game scores, or API usage.
...
 
-For example:
-
-- The UPI real-time payments switch in India processed 10 billion payments in 2019. The switch
-  processed 12 billion payments in the month of January 2024 alone.
-- With the moves to cleaner energy and smart metering, energy is being traded by the Kilowatt-hour
-  and billed to customers every 15 or 30 minutes rather than at the end of the month.
+- The UPI real-time payments switch in India processed 10 billion payments in the year 2019.
+  In January 2025 alone, it processed [16.9 billion payments.](https://www.npci.org.in/what-we-do/upi/product-statistics)
+- Cleaner energy and smart metering man energy is being traded by the kilowatt-hour.
```
martijngonlag:
```suggestion
- Cleaner energy and smart metering means energy is being traded by the kilowatt-hour.
```

docs/concepts/oltp.md
```diff
@@ -5,89 +5,75 @@ could be payments, sales, car sharing rides, game scores, or API usage.
...
-OLTP workloads today, which have increased by 3-4 orders of magnitude in the last 10 years alone.
+OLTP workloads have increased by 3-4 orders of magnitude in the last 10 years alone. For example:
 
-For example:
-
-- The UPI real-time payments switch in India processed 10 billion payments in 2019. The switch
-  processed 12 billion payments in the month of January 2024 alone.
-- With the moves to cleaner energy and smart metering, energy is being traded by the Kilowatt-hour
-  and billed to customers every 15 or 30 minutes rather than at the end of the month.
+- The UPI real-time payments switch in India processed 10 billion payments in the year 2019.
```
martijngonlag:
```suggestion
- The [UPI](https://en.wikipedia.org/wiki/Unified_Payments_Interface) real-time payments switch in India processed 10 billion payments in the year 2019.
```

docs/concepts/oltp.md
```diff
@@ -5,89 +5,75 @@ could be payments, sales, car sharing rides, game scores, or API usage.
...
+  This makes it difficult or impossible to offer customers real-time spending caps.)
+
+OLGP databases already struggle to keep up.
 
-**TigerBeetle is built to handle the scale of OLTP workloads today and for the decades to come.** It
-works well alongside a general purpose database. The latter holds data that isn't updated very
-frequently, such as account metadata, while TigerBeetle can race ahead processing transactions and
-giving your system unparalleled latency and throughput.
+**But TigerBeetle is built to handle the scale of OLTP workloads today and for the decades to come.**
+It works well alongside OLGP databases, holding data that isn't updated very frequently.
```
martijngonlag: I think this makes it flow a bit better!

```suggestion
It works well alongside OLGP databases, which holds infrequently updated data.
```

docs/concepts/oltp.md
```diff
@@ -5,89 +5,75 @@ could be payments, sales, car sharing rides, game scores, or API usage.
...
+- Row locks on hot accounts worsen when the transactions must execute across shards.
 
-Another approach to scaling OLTP systems is to use
-[MapReduce](https://en.wikipedia.org/wiki/MapReduce) for billing. Unfortunately, this makes it
-almost impossible to provide real-time balance reporting or real-time spending limits, because
-balances are only calculated periodically in batches. This is a poor user experience that cannot
-readily be fixed after the system is designed.
+Another approach to scaling OLTP systems is to use MapReduce for billing. But this makes it hard to
+provide real-time balance reporting or spending limits. It also creates a poor user experience
+that's hard to fix post system design.
```
martijngonlag, line 62: 🤔 We already touch on MapReduce on line 18. It might make sense to consolidate this information there instead?

martijngonlag (changes requested):

docs/concepts/oltp.md
matklad: Good one!

docs/concepts/oltp.md
matklad: I think it's actually good to circle here, because "why not map reduce" is a big idea, and worth repeating!

martijngonlag:
LGTM! 👍

martijngonlag (approved):

PR #2773, opened 2025-02-26
docs: add single page mode

fabioarnold:
A version of our docs that has all pages concatenated on a single page.

TODO:
- [x] write singe-page/index.html
- [x] add link to nav
- [x] exclude search from single page mode (can use Ctrl+F instead)
- [x] rewrite links in nav
- [x] rewrite links in content
- [x] update single page button styling

Bonus:
- tooltips for hot keys
- width and scroll position of the nav is preserved between page navigation again
- always focus the page content so arrow key scrolling works immediately 
- canonical links for pages
- style visited links
- move footnotes to the sections themselves


src/docs_website/pandoc/markdown-links.lua
```diff
@@ -6,6 +6,12 @@ function Link (link)
   local is_absolute = link.target:sub(1, 1) == "/"
   local is_anchor = link.target:sub(1, 1) == "#"
 
+  if link.target:sub(1, 29) == "https://docs.tigerbeetle.com/" then
+    link.target = link.target:gsub("https://docs.tigerbeetle.com/", "/")
+    is_external = false
+    is_absolute = true
+  end
+
```
batiati: Get rid of the constant `29` here : 

```lua
  local docs = "https://docs.tigerbeetle.com/"
  if link.target:sub(1, #docs) == docs then
    link.target = link.target:gsub(docs, "/")
    is_external = false
    is_absolute = true
  end
```

fabioarnold: So glad someone knows this language. 😄

batiati (approved):

PR #2779, opened 2025-02-28
docs: fix broken link to internals docs

evmar:


matklad (approved):
Thanks! This is somewhat expected fallout from the grad docs rewrite of last weeks!

PR #2781, opened 2025-02-28
Polish for Alex's ARCHITECTURE.md section

MarinaWoo:
@matklad - some edits for you. Mostly sentence length adjustment, some grammar/typo corrections, and then edits to make things read more easily. Please check I have not removed meaning!

### DIFF
--- a/docs/internals/ARCHITECTURE.md
+++ b/docs/internals/ARCHITECTURE.md
@@ -1,97 +1,92 @@
 # TigerBeetle Architecture
 
-This document is a short technical overview of the internals of TigerBeetle. It starts with the
-problem statement to motivate everything that follows, then gives the overview of TigerBeetle and
-provides motivation for a number of specific design decisions, and finally concludes with a list of
-prior art which inspired TigerBeetle.
+This document is a technical overview of the internals of TigerBeetle.  It includes a problem 
+statement, overview, motivation for design decisions and list of references that have inspired us.
 
 ## Problem Statement
 
 TigerBeetle is a database for workloads that:
 
-- are very contended and parallelize/shard poorly,
-- consist mostly of writes,
-- need very high throughput and moderately low latency,
-- require strong consistency guarantees,
-- and demand very high levels of durability.
+- are very contended and parallelize/shard poorly
+- consist mostly of writes
+- need very high throughput and moderately low latency
+- require strong consistency guarantees
+- demand very high levels of durability.
 
-At the moment, TigerBeetle is focused on financial transactions (transfers). Each transfer touches
-two accounts, and accounts are Pareto-distributed --- a few accounts are responsible for the bulk of
+At present, TigerBeetle is focused on financial transactions (transfers). Each transfer touches
+two accounts. Accounts are Pareto-distributed which means a few accounts are responsible for the bulk of
 the transfers. Multi-object transactions in combination with contention make traditional database
-architecture inefficient --- locking the two accounts and then going over the network to the
-application to compute the balances' update is slow, and, due to contention, this work can't be
+architecture inefficient. Locking the two accounts and then crossing the network to the
+application to compute the balances' update is slow. Due to contention, this work also can't be
 efficiently distributed across separate machines.
 
-While at the moment TigerBeetle implements only accounting logic, the underlying state machine can
-be swapped for a different one. TigerBeetle does only double-entry bookkeeping, but _can_ do
-anything that has the same combination of high performance and safety requirements, and high
-contention.
+TigerBeetle implements accounting logic but the underlying state machine can be swapped. TigerBeetle 
+does double-entry bookkeeping but _can_ do anything that has the same requirements.
 
 ## Overview
 
 TigerBeetle is a distributed system. It is a cluster of six replicas. The state of each replica is
-just a single file on disk, called the data file. Replicas use a consensus protocol to ensure that their
+a single file on disk, called the data file. Replicas use a consensus protocol to ensure that their
 respective data files store identical data.
 
 Specifically, TigerBeetle is a replicated state machine. The ground state of the system is an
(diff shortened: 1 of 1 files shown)


docs/internals/ARCHITECTURE.md
```diff
@@ -1,97 +1,92 @@
...
+architecture inefficient. Locking the two accounts and then crossing the network to the
+application to compute the balances' update is slow. Due to contention, this work also can't be
 efficiently distributed across separate machines.
 
-While at the moment TigerBeetle implements only accounting logic, the underlying state machine can
-be swapped for a different one. TigerBeetle does only double-entry bookkeeping, but _can_ do
-anything that has the same combination of high performance and safety requirements, and high
-contention.
+TigerBeetle implements accounting logic but the underlying state machine can be swapped. TigerBeetle 
+does double-entry bookkeeping but _can_ do anything that has the same requirements.
```
matklad: Let's keep the redundant repetition of "high performance, safety, extreme contention" --- while already mentioned above, it's worth circling around this more to make the key point sink.

docs/internals/ARCHITECTURE.md
```diff
@@ -1,97 +1,92 @@
...
+the rest of the state is reachable.
 
-When applying prepares to local state, replica doesn't overwrite any existing blocks in the data
-file, and instead writes new blocks only, maintaining the state of the superblock in memory. Once in
-a while, the superblock is atomically flushed to storage, forming a new checkpoint. When a replica
-crashes and restarts, it of course looses access to its previous in-memory superblock. Instead, it
-reads the previous checkpoint/superblock from storage, and reconstructs the lost state by replaying
-the suffix of the log of prepares after that checkpoint. Determinism guarantees that the replica
-ends up in the exact same state.
+When applying prepares to local state, replicsa don't overwrite any existing blocks in the data
```
matklad: replicas, not csa

docs/internals/ARCHITECTURE.md
```diff
@@ -1,97 +1,92 @@
...
+this is exactly what TigerBeetle does.
 
 ## Design Decisions
 
 ### Intentionality in the Design
 
-A lot of software is designed primarily via empirical feedback loop --- apply a small change to the
-software, and, judging the results, revert or double down. For TigerBeetle, we are trying something
-different: think from the first principles what the right solution _should_ be, and then use
+A lot of software is designed primarily via empirical feedback loop. You apply a small change to the
```
matklad, line 88: I like how you remove basically every `---` and just make it two separate sentences! So much better indeed!

docs/internals/ARCHITECTURE.md
```diff
@@ -103,127 +98,118 @@ hash map that stores accounts keyed by their IDs. Processing a transfer is then
...
-allocated somewhere on the heap and registered with an event loop, a `Box<dyn Future>`. Because
-TigerBeetle _can't_ heap allocate structures at runtime, it follows that each TigerBeetle "future"
-is represented by an explicit struct which is statically allocated as a field of a  component that
-owns the structure. So there's always a natural limit of how many concurrent tasks can be in flight!
-
-In summary, **static allocation is not the goal, but rather a forcing function to ensure that
-everything has a limit**, and a natural consequence of the limits.
+TigerBeetle also doesn't do arena allocation. It allocates an appropriately fixed-sized arena for 
+allocations at the start and fails with an 'out of memory' error if the limit is exceeded. In 
+TigerBeetle, static memory allocation is a consequence of everything having an explicit upper bound.
```
matklad: Here, the meaning got flipped! What I am trying to do here overall is contrast and compare:

Some systems use global statics (.bss), but TigerBeetle doesn't. 

Some systems use arena allocation, but TigerBeetle doesn't.

Instead, when TigerBeetle starts, it computes bla-bla-bla. 


So, the it in 

>  It allocates an appropriately fixed-sized arena for 
allocations at the start and fails with an 'out of memory' error if the limit is exceeded.

Should refer to a hypothetical application that uses arena allocation, not to TigerBeetle.

docs/internals/ARCHITECTURE.md
```diff
@@ -103,127 +98,118 @@ hash map that stores accounts keyed by their IDs. Processing a transfer is then
...
+
+When TigerBeetle starts, it computes the worst case memory allocation scenario for every "object" in 
+the system. After startup, no new objects are created therefore no dynamic memory allocation
```
matklad: >worst case memory allocation scenario for every "object" 

It's worth leaving the implication here: what TB actually computes is the _count_ of objects, and worst-case memory allocation is the result of the count. We actually can drop the memory part, and leave only the counts, as that's more important.

docs/internals/ARCHITECTURE.md
```diff
@@ -103,127 +98,118 @@ hash map that stores accounts keyed by their IDs. Processing a transfer is then
...
 Somewhat surprisingly, our experience is that static allocation also simplifies the system greatly.
-You need to spend more time up-front thinking through what data you need to store where, but, after
-the initial design, the interplay of the connected limits between the components tends to just work
-out.
+You spend more time thinking up-front but after the initial design the interplay tends to just work out.
 
-Of course static allocation makes the system faster, and greatly reduces latency variation, but this
-also is just a consequence of knowing the limits, the "physics" of the underlying system.
+Static allocation makes the system faster and greatly reduces latency variation. This is another 
+consequence of knowing the limits and "physics" of the underlying system.
```
matklad: Here, the `just` was actually a useful word --- what I am tryin to convey here is that while "performance" is a consequence of static allocation, and it is a _good_ consequence, but it is not the _why_ behind static allocations. 

I guess

> This is another  consequence of knowing the limits and "physics" of the underlying system, but not the main reason for choosing static allocation.

matklad (dismissed):

matklad (approved):

PR #2783, opened 2025-02-28
Assert the `cache_map.stash` count

batiati:
Since `ensureTotalCapacity` might allocate more than the exact capacity, it's necessary to assert that `stash.count() <= map_value_count_max`.

This change addresses two points:
- Use `assert` instead of `maybe` during compaction.
- Assert when adding or removing items from the stash.

For more context: I just hit the existing assert in `deinit()` during unrelated work while experimenting with different values for prefetch limits and was wondering why I didn’t hit the assertion earlier when the extra element was inserted.

### DIFF
--- a/src/lsm/cache_map.zig
+++ b/src/lsm/cache_map.zig
@@ -61,7 +61,7 @@ pub fn CacheMapType(
 
         pub const Options = struct {
             cache_value_count_max: u32,
-            map_value_count_max: u32,
+            stash_value_count_max: u32,
             scope_value_count_max: u32,
             name: []const u8,
         };
@@ -82,7 +82,7 @@ pub fn CacheMapType(
         options: Options,
 
         pub fn init(allocator: std.mem.Allocator, options: Options) !CacheMap {
-            assert(options.map_value_count_max > 0);
+            assert(options.stash_value_count_max > 0);
             maybe(options.cache_value_count_max == 0);
             maybe(options.scope_value_count_max == 0);
 
@@ -94,7 +94,7 @@ pub fn CacheMapType(
             errdefer if (cache) |*cache_unwrapped| cache_unwrapped.deinit(allocator);
 
             var stash: Map = .{};
-            try stash.ensureTotalCapacity(allocator, options.map_value_count_max);
+            try stash.ensureTotalCapacity(allocator, options.stash_value_count_max);
             errdefer stash.deinit(allocator);
 
             var scope_rollback_log = try std.ArrayListUnmanaged(Value).initCapacity(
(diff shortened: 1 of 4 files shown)


batiati:
Seed found by VOPR also fails in master:

`./zig/zig build -Drelease vopr -- 13232551907466493063`

src/lsm/cache_map.zig
```diff
@@ -300,7 +302,7 @@ pub fn CacheMapType(
         pub fn compact(self: *CacheMap) void {
             assert(!self.scope_is_active);
             assert(self.scope_rollback_log.items.len == 0);
-            maybe(self.stash.count() <= self.options.map_value_count_max);
+            assert(self.stash.count() <= self.options.map_value_count_max);
```
matklad: Huh, do you remember why we did maybe here in the first place?

matklad: Actually, this feels wrong? Consider a degenerate case, where the stash is sized normally, but `map` is sized to only have one element (almost no cache). In this situation, stash size _will_ be larger than map, no? 

Should we assert <= cache_value_count_max, and add a maybe in init to emphasize that map and stash sizes can be related in whatever way? 

Or maybe even assert that `map_count_max > cache_value_count`, for if the map is smaller, it is perhaps beneficial to just completely disable the cache?

batiati: > Huh, do you remember why we did maybe here in the first place?

No! That's why I left VOPR working on it during the weekend!
In fact I was very surprised to learn that [I authored this code change](https://github.com/tigerbeetle/tigerbeetle/pull/1584), and to quote Gandalf:

image

> Actually, this feels wrong? Consider a degenerate case, where the stash is sized normally, but map is sized to only have one element (almost no cache). In this situation, stash size will be larger than map, no?

We shouldn't rely on any extra capacity beyond `stash.ensureTotalCapacity` (unfortunately, there is no [ensureTotalCapacityPrecise](https://ziglang.org/documentation/0.13.0/std/#std.array_list.ArrayListAligned.ensureTotalCapacityPrecise) in hash maps).

If `maybe` is correct, we have a bug elsewhere, either calculating `map_value_count_max` too low or too many upserts happening.

matklad: Hm, the maybe seems correct: we compare _stash_ size with _map_ size, and map might be bigger than stash, no?

matklad: >In fact I was very surprised to learn that https://github.com/tigerbeetle/tigerbeetle/pull/1584,

Yeah, I've also blamed to that PR, and was surprised that reviewer didn't ask about the maybe. It probably was too obvious at that time :D 

But, anyway, I did convince myself that the maybe is correct here!

batiati: > Hm, the maybe seems correct: we compare _stash_ size with _map_ size, and map might be bigger than stash, no?

I'm missing something here!  Stash _is_ the map, we initialize it with that capacity, how it would be bigger?

https://github.com/tigerbeetle/tigerbeetle/blob/111e6fa905f9d18322d99615e03bb84b46b73565/src/lsm/cache_map.zig#L96-L98

matklad: Ah, no, you are of course right, I confused how we name the two... Should probably use `stash_value_count` in the public API to avoid synonyms!

🤔 Yeah, I _think_ that this assert is correct then!

matklad (approved):

PR #2784, opened 2025-02-28
docs: add a compatibility note about nested CSS selectors

### LINKED ISSUE #2777: Docs rendering funny on Linux Firefox/Chrome
Image

Image

This is on  Debian Bookworm, in Firefox and Chrome

(If no one else runs into this, feel free to close ... I am not actually using TB, just saw the blog post)

fabioarnold:
Screenshot 2025-02-28 at 14 09 03

This change adds a note that informs users of older browsers that this is not the way we intend the docs site to be rendered.

Addresses #2777.

### DIFF
--- a/src/docs_website/src/html/page.html
+++ b/src/docs_website/src/html/page.html
@@ -97,6 +97,7 @@
     <div class="resizer"></div>
     <main>
       <article>
+        <aside id="note-nested-css">Note: this site uses nested CSS selectors. The future is already here, it's just not evenly distributed yet! Support the latest version of your browser, which will render these properly.</aside>
         <a id="github-link" href="https://github.com/tigerbeetle/tigerbeetle">
           GitHub
           <svg width="16" height="16">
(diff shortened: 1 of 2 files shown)


src/docs_website/src/html/page.html
```diff
@@ -97,6 +97,7 @@
     <div class="resizer"></div>
     <main>
       <article>
+        <aside id="note-nested-css">Note: this site uses nested CSS selectors. The future is already here, it's just not evenly distributed yet! Support the latest version of your browser, which will render these properly.</aside>
```
matklad: Hm, let's maybe tone down a joke here a bit, a people with older browsers might be feeling somewhat dejected, and we should sweeten the pill. 

So I suggest going for the neutral:

> Note: this site uses nested CSS selectors. Please upgrade your browser or use the reader mode.

matklad (approved):

PR #2787, opened 2025-03-02
Multi-batch message encoder

batiati:
## Multi-batch message encoder

This PR is the _reincarnation_ of #2617 and #2682: the same spirit in a brand new body.  
No code is reused from the previous attempts (except for parts of the encoder/decoder), as it takes a completely different approach.

Batching consists of the application submitting multiple independent units of work of the same operation (the batch payload) within a single VSR message.
This amortizes network and consensus costs, improving performance in scenarios where highly concurrent user requests submit operations containing only a few events each, sharing the same physical request.

This PR replaces the state machine batching mechanism based on the `Create{Accounts,Transfers}Result.index` field (aka Demuxer) with a general-purpose batching encoder/decoder suitable for any operation. (See #2617 for more details about the motivation).

### How it works:

The client (tb_client library) is responsible for accumulating concurrent requests, encoding them into multi-batch messages, and sending them as a single VSR request.

Upon receiving a multi-batch request, the StateMachine decodes the message body, taking advantage of prefetching the entire payload in a single pass whenever possible, while executing each batch as an independent unit of work.

New operation numbers for multi-batch guarantee backward compatibility with old clients sending non-encoded messages.

Each batch may generate a number of results which are encoded in a multi-batch reply and sent back to the client.

The client decodes the multi-batch reply's body into multiple batch payloads and calls the respective user callback for each one.

### Encoding layout:

- Multi-batch messages use a portion at the end of the message body (the batch trailer) to encode batch metadata, so batched requests can hold fewer events than non-batched ones.

- The trailer size is always a multiple of the event/result size to keep system invariants.

- The batch trailer consists of an array of `u16` values representing the number of elements in each batch, plus a _postamble_ `u16` containing the total number of batches encoded in the message.

- The trailer has a variable length, depending on the number of batches, but always in multiples of the event/result size.

- To facilitate batch encoding, the trailer is stacked from the end of the message towards the beginning. For example, the last value in the array of element counts corresponds to the first batch.

- Unused elements in the trailer, required for padding, are filled with `maxInt(u16)`.

For more details on the message encoding format, please see `multi_batch.zig` comments.

https://github.com/tigerbeetle/tigerbeetle/blob/1df370356e3ec783a5a2c5618c8cc1f9c1e8f6bf/src/vsr/multi_batch.zig#L24-L41

### This PR

See the commit messages for the implementation details.

- StateMachine support with unit tests.
- tb_client support with unit and echo tests.
- MultiBatch encoder/decoder fuzzer.
- VOPR multibatch and multi client release support.
- Vortex (java driver only) support for async calls, testing multibatch requests.
- Client integration tests for concurrent lookups and queries.


src/vsr/replica.zig
```diff
@@ -5510,7 +5516,9 @@ pub fn ReplicaType(
                 return true;
             }
 
-            if (message.header.release.value > self.release.value) {
+            if (message.header.release.value > self.release.value and
+                !self.release.is_development())
+            {
```
batiati: I’m not particularly happy with this solution... but we need to test compatibility with clients <= `0.16.30` in a development cluster.

My first approach was to avoid relying on the `client_release` information for this, and instead create a new set of operations for multi-batch messages (`create_accounts_2`, `create_transfers_2`, etc.). But we already rely on `client_release` for other decisions (for example, `AccountFilterFormer`), so I think introducing a second mechanism for handling backwards compatibility would add unnecessary complexity right now.

Hopefully, we can deprecate all of this after a transition period (cutting the `minor = 17` release) and revisit this strategy after a proper code cleanup.

src/clients/c/tb_client/context.zig
```diff
@@ -430,113 +473,159 @@ pub fn ContextType(
...
-            const bytes_writen: u32 = bytes_writen: {
-                // Copy all batched packet event data into the message buffer.
-                const buffer: []u8 = message.buffer[@sizeOf(Header)..];
-                assert(buffer.len >= packet.batch_size);
+            const operation: StateMachine.Operation = operation_from_int(packet_list.operation).?;
+            const event_size: u32, const result_size: u32 = switch (operation) {
+                .pulse, .get_events => unreachable,
+                inline else => |operation_comptime| .{
+                    @sizeOf(StateMachine.EventType(operation_comptime)),
+                    @sizeOf(StateMachine.ResultType(operation_comptime)),
```
sentientwaffle: Earlier we use `event_size_bytes`/`result_size_bytes` -- should we be consistent or is there a reason to diverge here?

src/clients/c/tb_client/context.zig
```diff
@@ -647,35 +736,34 @@ pub fn ContextType(
...
-                        const batched_reply = demuxer.decode(event_offset, event_count);
-                        event_offset += event_count;
-
-                        if (!StateMachine.batch_logical_allowed.get(operation_comptime)) {
-                            assert(batched.batch_next == null);
-                            assert(batched_reply.len == reply.len);
-                        }
-
+                        if (batched != packet_list) batched.assert_phase(.batched);
+                        it = batched.multi_batch_next;
```
sentientwaffle: Maybe pull this up into

```
while (it) |batched| : (it = batched.multi_batch_next) {
```

to make it more obvious? Right now it is hiding in between a couple of assertions.

src/vsr/multi_batch_fuzz.zig
sentientwaffle, line 1: Also add this to `cfo.zig`'s task list!

src/stdx.zig
```diff
@@ -173,6 +173,29 @@ pub fn zeroed(bytes: []const u8) bool {
...
 }
 
+/// Similar to `std.mem.bytesAsSlice`, but allows buffers with inexact sizes,
+/// returning the largest possible slice that is less than or equal to the buffer length.
+/// Accepts only mutable slices, as it is intended for writing memory,
+/// where skipped bytes at the end can be safely ignored.
+pub fn buffer_as_slice(comptime T: type, buffer: []u8) []T {
+    const size = @divFloor(buffer.len, @sizeOf(T)) * @sizeOf(T);
+    return @alignCast(std.mem.bytesAsSlice(T, buffer[0..size]));
+}
```
sentientwaffle: :thinking: Is this a common enough idiom to warrant promotion to stdx?
Also I don't think that the function name is descriptive enough -- from the name alone I would expect it to behave more like `bytesAsSlice`.

src/testing/cluster.zig
```diff
@@ -297,7 +297,16 @@ pub fn ClusterType(comptime StateMachineType: anytype) type {
...
-                client.release = options.client_release;
+                if (options.client_release) |client_release| {
+                    client.release = client_release;
+                } else {
+                    // Choosing a release from the list of supported releases
+                    // the workload can simulate.
+                    const client_releases_supported = StateMachine.Workload.releases_supported;
+                    client.release = client_releases_supported[
+                        random.uintLessThan(usize, client_releases_supported.len)
+                    ];
```
sentientwaffle: This logic should move up a level, into `vopr.zig`.
Where practical, that is where the test parameter randomization takes place.

src/testing/vortex/java_driver/src/main/java/Main.java
```diff
@@ -55,24 +68,42 @@ record Driver(Client client, Reader reader, Writer writer) {
   /**
    * Reads the next operation from stdin, runs it, collects the results, and writes them back to
    * stdout.
-   */
-  void next() throws IOException {
+      * @throws ExecutionException
+      * @throws InterruptedException
+      */
```
sentientwaffle: Indentation?

src/testing/vortex/java_driver/src/main/java/Main.java
```diff
@@ -113,6 +144,62 @@ void createAccounts(Reader reader, Writer writer, int count) throws IOException
     writer.flush();
   }
 
+  void createAccountsAsync(Reader reader, Writer writer, int count)
```
sentientwaffle, line 151: I might not be understanding this right, but does this function potentially take a batch from the user and split it into multiple physical requests?

docs/reference/requests/lookup_accounts.md
sentientwaffle, line 1: Do we also need to change the warning in `query_accounts.md`?

src/vsr/replica.zig
sentientwaffle: :thinking: Yeah, this makes me nervous...

> I’m not particularly happy with this solution... but we need to test compatibility with clients <= 0.16.30 in a development cluster.

Which tests need this? 

Could we instead have `release.supports_multi_batching()`, and then have two different development releases `0.0.1`/`0.0.2` which return different values?

src/tigerbeetle/benchmark_load.zig
```diff
@@ -672,11 +719,46 @@ const Benchmark = struct {
...
-            .create_transfers => b.create_transfers_callback(client, result),
-            .lookup_accounts => b.validate_accounts_callback(client, @alignCast(result)),
-            .lookup_transfers => b.validate_transfers_callback(client, @alignCast(result)),
-            .get_account_transfers => b.get_account_transfers_callback(client, result),
+            .create_accounts => {
+                var reply_decoder = vsr.multi_batch.MultiBatchDecoder.init(
+                    result,
+                    .{ .element_size = @sizeOf(tb.CreateAccountsResult) },
+                ) catch unreachable;
+                assert(reply_decoder.batch_count() == 1);
```
sentientwaffle: Could this part of the logic be pulled out of the switch and shared? e.g. via `@sizeOf(StateMachine.ResultType(...))`.

sentientwaffle (commented):
WIP review

src/vsr/multi_batch.zig
```diff
@@ -0,0 +1,763 @@
...
+        });
+        assert(buffer.len >= trailer_size_min);
+
+        // The end of the buffer must be aligned with the trailer.
+        // If it isn't, reduce the buffer to maintain alignment.
+        const aligned_len = std.mem.alignBackward(
+            usize,
+            buffer.len,
+            @sizeOf(TrailerItem),
+        );
```
sentientwaffle, line 315: Is the length ever not already a multiple of `@sizeOf(TrailerItem)`? It would be great if this could be an assert instead.

src/vsr/multi_batch.zig
```diff
@@ -0,0 +1,763 @@
...
+        self.buffer_index += bytes_written;
+
+        const buffer: []u8 = self.buffer.?;
+        assert(self.buffer_index < buffer.len);
+
+        const trailer_size = trailer_total_size(.{
+            .batch_count = self.batch_count,
+            .element_size = self.options.element_size,
+        });
+        assert(buffer.len >= self.buffer_index + trailer_size);
```
sentientwaffle: I would swap the order of this clause so that it is more clear that it is a refinement of the previous assert.

src/vsr/multi_batch.zig
```diff
@@ -0,0 +1,763 @@
...
+        return slice[0..size];
+    }
+
+    /// Records how many bytes were written in the slice previously acquired by `writable()`.
+    pub fn add(self: *MultiBatchEncoder, bytes_written: u32) void {
+        assert(self.batch_count < multi_batch_count_max);
+        maybe(self.batch_count == 0);
+
+        const element_count: u16 = element_count: {
+            if (self.options.element_size == 0) {
```
sentientwaffle: What about naming these `written_bytes` and `written_element_count`, so that it is clear that this is not the cumulative element count.

src/vsr/multi_batch.zig
```diff
@@ -0,0 +1,763 @@
...
+        maybe(self.buffer_index == 0);
+
+        const trailer_size = trailer_total_size(.{
+            .batch_count = self.batch_count,
+            .element_size = self.options.element_size,
+        });
+
+        // For byte-aligned elements, padding may be required between the payload and the trailer.
+        const padding: u32 = self.buffer_index % @sizeOf(TrailerItem);
+        assert(padding < @sizeOf(TrailerItem));
```
sentientwaffle, line 438: Also to make this extra clear we can assert that either the padding is zero or the size of the element is one.

src/vsr/multi_batch.zig
```diff
@@ -0,0 +1,763 @@
...
+        // Filling in the extra alignment bytes with sentinels.
+        @memset(
+            trailer_items[0 .. trailer_items.len - self.batch_count],
+            TrailerItem.padding,
+        );
+
+        const postamble: *Postamble = @alignCast(@ptrCast(
+            buffer[self.buffer_index + padding + trailer_size - @sizeOf(Postamble) ..],
+        ));
+        postamble.batch_count = self.batch_count;
```
sentientwaffle: Instead set this as `postamble.*= ...` to make it clear that we are setting all fields.

src/vsr/multi_batch.zig
```diff
@@ -0,0 +1,763 @@
...
+
+    pub fn init(
+        /// The message body used, including the trailer.
+        body: []const u8,
+        options: Options,
+    ) error{MultiBatchInvalid}!MultiBatchDecoder {
+        // Supports zero-sized elements, or any power of two, including 2^0.
+        assert(options.element_size == 0 or std.math.isPowerOfTwo(options.element_size));
+
+        // Empty messages are considered valid.
```
sentientwaffle: I'm confused, where in the encoder do we create these empty messages?

src/vsr/multi_batch.zig
```diff
@@ -0,0 +1,763 @@
...
+        const batch_size = trailer_item.element_count * self.options.element_size;
+        assert(self.payload_index + batch_size <= self.payload.len);
+
+        const slice: []const u8 = self.payload[self.payload_index..][0..batch_size];
+        assert(slice.len > 0);
+        assert(slice.len % self.options.element_size == 0);
+        return slice;
+    }
+
+    pub fn move_next(self: *MultiBatchDecoder) bool {
```
sentientwaffle: As far as I can tell this doesn't need to be public.

src/vsr/multi_batch_fuzz.zig
```diff
@@ -0,0 +1,154 @@
...
+    const batch_element_size: u32 = std.math.pow(u32, 2, options.random.uintAtMost(u32, 8));
+    var batches = stdx.BoundedArrayType(u32, 8190){};
+    const batch_count_max = options.random.intRangeAtMost(
+        usize,
+        1,
+        batches.capacity(),
+    );
+
+    // Encoder will ignore and overwrite any existing content in the target buffer,
+    // at the end, both buffers must be equal.
```
sentientwaffle, line 67: Excellent idea!

sentientwaffle (commented):
WIP review

src/vsr/multi_batch_fuzz.zig
```diff
@@ -0,0 +1,154 @@
...
+        options.buffer_actual[0 .. options.buffer_actual.len - postamble_alignment];
+
+    // Generate the batch plan with element sizes from 2^0 to 2^8.
+    const batch_element_size: u32 = std.math.pow(u32, 2, options.random.uintAtMost(u32, 8));
+    var batches = stdx.BoundedArrayType(u32, 8190){};
+    const batch_count_max = options.random.intRangeAtMost(
+        usize,
+        1,
+        batches.capacity(),
+    );
```
sentientwaffle, line 64: Also add a randomized parameter to serve as another limit on `batch_element_count` -- as it is now, I think the vast majority of test runs will only use very few batches because it only takes a couple of large batches to use up all of the space.

src/vsr/multi_batch_fuzz.zig
sentientwaffle, line 1: Additionally fuzz the negative space, e.g.:

```diff
diff --git a/src/vsr/multi_batch_fuzz.zig b/src/vsr/multi_batch_fuzz.zig
index 71f33c958..e68379423 100644
--- a/src/vsr/multi_batch_fuzz.zig
+++ b/src/vsr/multi_batch_fuzz.zig
@@ -132,23 +132,52 @@ fn run_fuzz(options: struct {
     assert(encoder_bytes_written ==
         expect_payload_size + expect_padding + expect_trailer_size);
 
-    // Decode.
-    var decoder = try MultiBatchDecoder.init(
-        buffer_actual[0..encoder_bytes_written],
-        .{ .element_size = batch_element_size },
-    );
-    assert(expect_payload_size == decoder.payload.len);
+    {
+        // Decode.
+        var decoder = try MultiBatchDecoder.init(
+            buffer_actual[0..encoder_bytes_written],
+            .{ .element_size = batch_element_size },
+        );
+        assert(expect_payload_size == decoder.payload.len);
 
-    var payloads_decoded: u32 = 0;
-    for (batches.const_slice()) |batch_element_count| {
-        const batch_size: u32 = batch_element_count * batch_element_size;
-        const expect_batch: []u8 = buffer_expected[payloads_decoded..][0..batch_size];
-        payloads_decoded += batch_size;
+        var payloads_decoded: u32 = 0;
+        for (batches.const_slice()) |batch_element_count| {
+            const batch_size: u32 = batch_element_count * batch_element_size;
+            const expect_batch: []u8 = buffer_expected[payloads_decoded..][0..batch_size];
+            payloads_decoded += batch_size;
+
+            const decoded_batch = decoder.pop().?;
+            assert(batch_size == decoded_batch.len);
+            assert(std.mem.eql(u8, expect_batch, decoded_batch));
+        }
+        assert(payloads_decoded == decoder.payload.len);
+        assert(decoder.pop() == null);
+    }
 
-        const decoded_batch = decoder.pop().?;
-        assert(batch_size == decoded_batch.len);
-        assert(std.mem.eql(u8, expect_batch, decoded_batch));
+    // Verify that any flipped bit mutates the results or causes a decoding error
+    // (but never a panic).
+    for (0..32) |_| {
+        const byte_index = options.random.uintLessThan(usize, encoder_bytes_written);
+        for (0..@bitSizeOf(u8)) |bit_index| {
+            buffer_actual[byte_index] ^= @as(u8, 1) << @as(u3, @intCast(bit_index));
+            defer buffer_actual[byte_index] ^= @as(u8, 1) << @as(u3, @intCast(bit_index));
+
+            var decoder = MultiBatchDecoder.init(
+                buffer_actual[0..encoder_bytes_written],
+                .{ .element_size = batch_element_size },
+            ) catch continue;
+
+            var same: bool = true;
+            var payloads_decoded: u32 = 0;
+            for (batches.const_slice()) |batch_element_count| {
+                const batch_size: u32 = batch_element_count * batch_element_size;
+                const expect_batch: []u8 = buffer_expected[payloads_decoded..][0..batch_size];
+                payloads_decoded += batch_size;
+
+                const decoded_batch = decoder.pop().?;
+                same = same and std.mem.eql(u8, expect_batch, decoded_batch);
+            }
+            assert(!same);
+        }
     }
-    assert(payloads_decoded == decoder.payload.len);
-    assert(decoder.pop() == null);
 }
```

src/vsr/multi_batch.zig
batiati: Not anymore, it was to simplify handling operations such `pulse` and `get_events` as always multi-batch.

src/vsr/multi_batch.zig
batiati: No, the buffer size can be of any size (and we fuzz it with irregular sizes).
Only the final encoded message after calling `finish()` will be.

src/vsr/multi_batch_fuzz.zig
batiati: We do have `enum { zero, one, random }` for `batch_element_count`.

batiati (commented):
The way we handle legacy operations was reworked after #2824.
It's ready for another review.

src/clients/c/test.zig
```diff
@@ -133,16 +141,58 @@ test "tb_client echo" {
...
+            {
+                operation_current = operations[prng.index(operations)];
+            }
+            break :operation operation_current.?;
+        };
+
+        const event_size: u32, const event_request_max: u32 = switch (operation) {
+            .create_accounts => .{
+                @sizeOf(tb.Account),
+                @divExact(constants.message_body_size_max, @sizeOf(tb.Account)) - 1,
```
sentientwaffle, line 174: Is this -1 due to the trailer? If so it probably warrants a comment.

src/repl.zig
```diff
@@ -824,24 +830,25 @@ pub fn ReplType(comptime MessageBus: type, comptime Time: type) type {
         fn send(
             repl: *Repl,
             operation: StateMachine.Operation,
-            arguments: []const u8,
+            buffer: []u8,
+            payload_size: u32,
```
sentientwaffle: Would it be simpler to handle the multi batching encoding at the caller site so that we don't have to pass in a partially used buffer + count? Or alternatively we could pass in the whole array list.

src/state_machine.zig
```diff
@@ -554,39 +507,237 @@ pub fn StateMachineType(
...
+                    @divFloor(constants.message_body_size_max, result_size)
+                else
+                    @min(
+                        @divFloor(batch_size_limit, event_size),
+                        @divFloor(constants.message_body_size_max, result_size),
+                    );
+            }
+            assert(operation_is_multi_batch(operation));
+
+            const reply_trailer_size_min: u32 = vsr.multi_batch.trailer_total_size(.{
```
sentientwaffle, line 615: Would it be simpler if this function called `operation_result_max` for the reply part?

src/state_machine.zig
```diff
@@ -855,67 +1081,178 @@ pub fn StateMachineType(
...
+                .pulse, .get_events => return message_body_used.len == 0,
+                inline .create_accounts,
+                .create_transfers,
+                .lookup_accounts,
+                .lookup_transfers,
+                .get_account_transfers,
                 .get_account_balances,
                 .query_accounts,
                 .query_transfers,
                 => |operation_comptime| {
```
sentientwaffle: Instead of a more verbose switch statement, would it make sense to just do a conditional based off of `if (operation_is_multi_batch(...)_) { ... } else { ... }`?

src/state_machine.zig
```diff
@@ -855,67 +1081,178 @@ pub fn StateMachineType(
...
+
+            maybe(delta == 0);
+            self.prepare_timestamp += delta;
+        }
 
-            self.prepare_timestamp += switch (operation) {
+        /// Returns the logical time increment (in nanoseconds) for the highest
+        /// timestamp of the batch.
+        /// For multi-batch requests, this function expects a single, already decoded batch.
+        /// Inline function so that `operation` can be known at comptime.
```
sentientwaffle: I'm not sure what this comment means; I don't think this function can ever be called at compile time.

src/stdx.zig
```diff
@@ -177,6 +177,83 @@ pub fn zeroed(bytes: []const u8) bool {
     return byte_bits == 0;
 }
 
+/// Similar to `std.mem.bytesAsSlice`, but allows buffers with inexact sizes,
+/// returning the largest possible slice that is less than or equal to the buffer length.
+/// Differently from `std.mem.bytesAsSlice` that can return `[]align(1) T`, this function
+/// always `@alignCast` the result.
+pub fn bytes_as_slice(
+    comptime T: type,
+    comptime mode: enum { exact, inexact },
```
sentientwaffle: Maybe swap the argument order to mirror `copy_disjoint`?

src/tigerbeetle/benchmark_load.zig
```diff
@@ -367,18 +367,28 @@ const Benchmark = struct {
...
+                .event_size = @sizeOf(tb.Account),
+            });
         }
     }
 
     fn create_accounts_callback(b: *Benchmark, client_index: u32, result: []const u8) void {
         assert(b.stage == .create_accounts);
 
-        const create_accounts_results = std.mem.bytesAsSlice(tb.CreateAccountsResult, result);
+        const create_accounts_results: []const tb.CreateAccountsResult = stdx.bytes_as_slice(
```
sentientwaffle: Why use an explicit type signature? Seems unnecessarily verbose.
(Likewise elsewhere -- I think the return value for `bytesAsSlice`/`bytes_as_slice` is well-known?)

src/tigerbeetle/benchmark_load.zig
```diff
@@ -513,29 +542,45 @@ const Benchmark = struct {
...
+        const accounts_expected: []const tb.Account = stdx.bytes_as_slice(
+            tb.Account,
+            .exact,
+            accounts_expected_body,
+        )[0..accounts_count];
+        const accounts_actual: []const tb.Account = stdx.bytes_as_slice(
+            tb.Account,
+            .exact,
+            result,
+        );
```
sentientwaffle: Why use `bytes_as_slice` (rather than `bytesAsSlice`) -- isn't `bytesAsSlice` actually _stricter_ since it doesn't cast alignment? (Also it is less verbose.)

src/state_machine/auditor.zig
```diff
@@ -113,7 +113,7 @@ pub const AccountingAuditor = struct {
         account_id_permutation: IdPermutation,
         client_count: usize,
 
-        batch_create_transfers_limit: u32,
+        transfers_expiry_max: u32,
```
sentientwaffle: "expiry max" makes this sound like a timestamp to me. What about `pulse_expiries_max`?

src/vsr/multi_batch.zig
```diff
@@ -0,0 +1,732 @@
...
+    payload_index: u32,
+    batch_index: u16,
+
+    options: Options,
+
+    pub fn init(
+        /// The message body used, including the trailer.
+        body: []const u8,
+        options: Options,
+    ) error{MultiBatchInvalid}!MultiBatchDecoder {
```
sentientwaffle: What do you think of this refactor? My goal was to reduce the amount of manual offset calculations, and to make it super clear that checks like size/alignment are performed for each section.

```diff
diff --git i/src/vsr/multi_batch.zig w/src/vsr/multi_batch.zig
index 3ffba18d4..8a93c7202 100644
--- i/src/vsr/multi_batch.zig
+++ w/src/vsr/multi_batch.zig
@@ -121,53 +121,34 @@ pub const MultiBatchDecoder = struct {
         // Supports zero-sized elements, or any power of two, including 2^0.
         assert(options.element_size == 0 or std.math.isPowerOfTwo(options.element_size));
 
-        if (body.len < @sizeOf(Postamble)) return error.MultiBatchInvalid;
-        if (!std.mem.isAligned(
-            @intFromPtr(&body[body.len - @sizeOf(Postamble)]),
-            @alignOf(Postamble),
-        )) {
-            return error.MultiBatchInvalid;
-        }
+        const Parser = struct {
+            buffer: []const u8,
+            buffer_parsed: u32 = 0,
 
-        const postamble: *const Postamble = @alignCast(@ptrCast(
-            body[body.len - @sizeOf(Postamble) ..],
-        ));
-        if (postamble.batch_count == 0) return error.MultiBatchInvalid;
-        if (postamble.batch_count > multi_batch_count_max) return error.MultiBatchInvalid;
+            const Error = error{MultiBatchInvalid};
+            fn parse_suffix(parser: *@This(), comptime T: type, count: u32) Error![]const T {
+                const suffix_size = count * @sizeOf(T);
+                if (parser.buffer.len < suffix_size) return error.MultiBatchInvalid;
+                const suffix = parser.buffer[parser.buffer.len - suffix_size..];
+                const suffix_aligned = std.mem.isAligned(@intFromPtr(suffix.ptr), @alignOf(T));
+                if (!suffix_aligned) return error.MultiBatchInvalid;
+                parser.buffer = parser.buffer[0..parser.buffer.len - suffix_size];
+                parser.buffer_parsed += suffix_size;
+                return @alignCast(std.mem.bytesAsSlice(T, suffix));
+            }
+        };
 
+        var parser = Parser{ .buffer = body };
+        const postambles = try parser.parse_suffix(Postamble, 1);
+        const postamble = &postambles[0];
         const trailer_size = trailer_total_size(.{
             .element_size = options.element_size,
             .batch_count = postamble.batch_count,
         });
-        if ((options.element_size > 0 and body.len < trailer_size) or
-            (options.element_size == 0 and body.len != trailer_size))
-        {
-            return error.MultiBatchInvalid;
-        }
-        if (!std.mem.isAligned(
-            @intFromPtr(body[body.len - trailer_size ..].ptr),
-            @alignOf(TrailerItem),
-        )) {
-            return error.MultiBatchInvalid;
-        }
+        const trailer_items_used = try parser.parse_suffix(TrailerItem, postamble.batch_count);
+        const trailer_items_padding =
+            try parser.parse_suffix(u8, trailer_size - parser.buffer_parsed);
 
-        const trailer_items: []const TrailerItem = @alignCast(std.mem.bytesAsSlice(
-            TrailerItem,
-            body[body.len - trailer_size .. body.len - @sizeOf(Postamble)],
-        ));
-        if (trailer_items.len < postamble.batch_count) return error.MultiBatchInvalid;
-
-        const trailer_items_used = trailer_items[trailer_items.len - postamble.batch_count ..];
-        assert(trailer_items_used.len == postamble.batch_count);
-        if (trailer_items.len > postamble.batch_count) {
-            // Check the padding sentinel of the extra slots used for alignment.
-            const BackingInteger = @typeInfo(TrailerItem).Struct.backing_integer.?;
-            if (!std.mem.allEqual(
-                BackingInteger,
-                @ptrCast(trailer_items[0 .. trailer_items.len - postamble.batch_count]),
-                @bitCast(TrailerItem.padding),
-            )) return error.MultiBatchInvalid;
-        }
         const events_count_total: u32 = count: {
             var count: u32 = 0;
             for (trailer_items_used) |trailer_item| {
@@ -176,27 +157,28 @@ pub const MultiBatchDecoder = struct {
             break :count count;
         };
         if (options.element_size == 0 and events_count_total != 0) return error.MultiBatchInvalid;
-        const payload_size: u32 = std.math.mul(
-            u32,
-            events_count_total,
-            options.element_size,
-        ) catch |err| switch (err) {
+
+        const payload_size_or_overflow =
+            std.math.mul(u32, events_count_total, options.element_size);
+        const payload_size: u32 = payload_size_or_overflow catch |err| switch (err) {
             error.Overflow => return error.MultiBatchInvalid,
         };
 
         // For byte-aligned elements, padding may be required between the payload and the trailer.
-        const padding: u32 = @intCast(payload_size % @sizeOf(TrailerItem));
-        assert(padding < @sizeOf(TrailerItem));
-        assert(padding == 0 or options.element_size == 1);
-        if (!std.mem.allEqual(
-            u8,
-            body[body.len - padding - trailer_size ..][0..padding],
-            std.math.maxInt(u8),
-        )) return error.MultiBatchInvalid;
-        if (body.len != payload_size + padding + trailer_size) return error.MultiBatchInvalid;
+        const trailer_padding_size: u32 = @intCast(payload_size % @sizeOf(TrailerItem));
+        assert(trailer_padding_size < @sizeOf(TrailerItem));
+        assert(trailer_padding_size == 0 or options.element_size == 1);
+        const trailer_padding = try parser.parse_suffix(u8, trailer_padding_size);
+
+        for ([_][]const u8{ trailer_items_padding, trailer_padding }) |padding| {
+            if (!std.mem.allEqual(u8, padding, std.math.maxInt(u8))) return error.MultiBatchInvalid;
+        }
+
+        if (payload_size != body.len - parser.buffer_parsed) return error.MultiBatchInvalid;
+        assert(payload_size == parser.buffer.len);
 
         return .{
-            .payload = body[0..payload_size],
+            .payload = parser.buffer,
             .trailer_items = trailer_items_used,
             .payload_index = 0,
             .batch_index = 0,
```

src/state_machine.zig
```diff
@@ -3263,10 +4296,19 @@ pub fn StateMachineType(
...
+                operation_event_max(.create_accounts, batch_size_limit),
+                operation_event_max(.deprecated_create_accounts, batch_size_limit),
+            );
             assert(batch_create_accounts > 0);
+            assert(batch_create_accounts <= constants.batch_max.create_accounts);
+
+            const batch_create_transfers: u32 = @max(
+                operation_event_max(.create_transfers, batch_size_limit),
+                operation_event_max(.deprecated_create_transfers, batch_size_limit),
+            );
```
sentientwaffle, line 4300: Aren't these the same values as the `batch_max` constants?

src/state_machine.zig
```diff
@@ -752,13 +933,18 @@ pub fn StateMachineType(
...
         prefetch_context: PrefetchContext = .null,
 
         scan_lookup: ScanLookup = .null,
         scan_lookup_buffer: []align(16) u8,
-        scan_lookup_result_count: ?u32 = null,
+        scan_lookup_buffer_index: u32 = 0,
+        scan_lookup_results: stdx.BoundedArrayType(
+            u32,
+            vsr.multi_batch.multi_batch_count_max,
+        ) = .{},
```
sentientwaffle: I think this is a 256KiB allocation, which is much more than we actually need. Instead of `multi_batch_count_max` could this be a `@max` of operation batch max's?

Also we should move this to the heap, since even at its minimal size this is a bit large to be on the stack.

src/vsr/multi_batch.zig
```diff
@@ -0,0 +1,732 @@
...
+    comptime {
+        assert(@sizeOf(TrailerItem) == @sizeOf(Postamble));
+        assert(@alignOf(TrailerItem) == @alignOf(Postamble));
+    }
+};
+
+/// The maximum number of batches that can be encoded in a message.
+/// Since the multi-batch trailer has a variable length, the actual limit depends on the buffer
+/// size and the number of elements in each batch.
+pub const multi_batch_count_max: u16 = std.math.maxInt(u16);
```
sentientwaffle: Could this be `@min(maxInt(u16), message_body_size_max)`?

sentientwaffle (commented):
Review is still WIP, but posting what I have so far!

src/state_machine.zig
batiati: Ideally, we would have only `operation_batch_max`, calculating the `min(event, result)`, but since queries are a special case, we need both `operation_event_max` and `operation_result_max`.
In the current implementation, it's `operation_result_max` that calls this function for batchable operations, as the number of results also depends on the number of events we can send.

https://github.com/tigerbeetle/tigerbeetle/blob/19b013670f1e7ee425709d99e25c69cecc14eeba/src/state_machine.zig#L646-L647

src/state_machine.zig
batiati: No, it's based on `batch_size_limit` argument, which is the same in

```zig
const tree_values_count_max = tree_values_count(config.message_body_size_max);
```
but not in

```zig
const tree_values_count_limit = tree_values_count(options.batch_size_limit);
```

Actually, I've considered removing the `batch_max` constants completely, but they still have use to justify.

src/vsr/multi_batch.zig
```diff
@@ -50,6 +50,9 @@ const constants = @import("../constants.zig");
 const vsr = @import("../vsr.zig");
 
 const Postamble = packed struct(u16) {
```
sentientwaffle, line 52: I assume that the reason that we aren't going to reserve any space in the postamble is that if want to (e.g.) add flags in the future then we can just create new operations, is that right?

src/vsr/multi_batch.zig
```diff
@@ -0,0 +1,745 @@
...
+        return slice;
+    }
+
+    fn move_next(self: *MultiBatchDecoder) bool {
+        assert(self.trailer_items.len > 0);
+        maybe(self.payload.len == 0);
+
+        if (self.batch_index == self.trailer_items.len) {
+            assert(self.payload_index == self.payload.len);
+            return false;
```
sentientwaffle: I think this branch is unreachable. We only call this function from `pop` and it has already checked this condition.

src/vsr/multi_batch.zig
```diff
@@ -0,0 +1,745 @@
...
+                parser.buffer_parsed += suffix_size;
+                return stdx.bytes_as_slice(.exact, T, suffix);
+            }
+        };
+
+        var parser = Parser{ .buffer = body };
+        const postamble: *const Postamble = postamble: {
+            const slice = try parser.parse_suffix(Postamble, 1);
+            break :postamble &slice[0];
+        };
```
sentientwaffle, line 166:
```suggestion
        };
        if (postamble.batch_count == 0) return error.MultiBatchInvalid;
```
(Otherwise `trailer_total_size` can panic.)

This was caught by the [negative space fuzzing](https://github.com/tigerbeetle/tigerbeetle/pull/2787/files#r1996363952) -- could we get those into this branch?

src/vsr/multi_batch.zig
```diff
@@ -0,0 +1,745 @@
...
+        assert(self.trailer_items.len > 0);
+        assert(self.batch_index < self.trailer_items.len);
+        assert(self.payload_index <= self.payload.len);
+        maybe(self.payload.len == 0);
+
+        // Batch metadata is written from the end of the message, so the last
+        // element corresponds to the first batch.
+        const trailer_item: *const TrailerItem =
+            &self.trailer_items[self.trailer_items.len - self.batch_index - 1];
+        maybe(trailer_item.element_count == 0);
```
sentientwaffle: Since we immediately check this condition in an `if` statement, the `maybe` statement is redundant.

src/vsr/multi_batch.zig
```diff
@@ -0,0 +1,745 @@
...
+        if (self.batch_index == self.trailer_items.len) {
+            assert(self.payload_index == self.payload.len);
+            return null;
+        }
+        assert(self.batch_index < self.trailer_items.len);
+        assert(self.payload_index <= self.payload.len);
+
+        const batch_item: []const u8 = self.peek();
+        const moved = self.move_next();
+        maybe(moved);
```
sentientwaffle: We can remove `move_next` entirely and replace it with:
```suggestion
        self.batch_index += 1;
        self.payload_index += @intCast(batch_item.len);
```

src/vsr/multi_batch.zig
```diff
@@ -0,0 +1,745 @@
...
+    /// Returns the total number of bytes written (payload + trailer).
+    /// At least one batch must be inserted, and the encoder should not be used after
+    /// being finished.
+    pub fn finish(self: *MultiBatchEncoder) u32 {
+        assert(self.batch_count <= Postamble.batch_count_max);
+        // Empty messages are considered valid.
+        if (self.batch_count == 0) {
+            assert(self.buffer_index == 0);
+            return 0;
+        }
```
sentientwaffle: It seems like we are being inconsistent over whether it is legal to encode a multi batch with no batches. The encoder appears to allow it, but the documentation says it isn't allowed, and the decoder fails. e.g.:

```
test "batch: encode no batches" {
    const buffer_size = (1024 * 1024) - @sizeOf(vsr.Header); // 1MiB message.
    const buffer = try testing.allocator.alignedAlloc(u8, @alignOf(vsr.Header), buffer_size);
    var encoder = MultiBatchEncoder.init(buffer, .{ .element_size = 128 });
    const bytes_written = encoder.finish();

    var decoder = MultiBatchDecoder.init(
        buffer[0..bytes_written],
        .{ .element_size = 128 },
    ) catch unreachable;
    assert(decoder.batch_count() == 0);
}
```

src/vsr/multi_batch.zig
```diff
@@ -0,0 +1,745 @@
...
+    ));
+}
+
+const TestRunner = struct {
+    fn run(options: struct {
+        prng: *stdx.PRNG,
+        element_size: u32,
+        buffer: []align(16) u8,
+        batch_count: u16,
+        elements_per_batch: ?u16 = null,
```
sentientwaffle: Or `batch_elements`.

src/vsr/multi_batch.zig
```diff
@@ -0,0 +1,745 @@
...
+        const BoundedArray = stdx.BoundedArrayType(u16, std.math.maxInt(u16));
+        var expected: BoundedArray = .{};
+
+        const trailer_size = trailer_total_size(.{
+            .element_size = options.element_size,
+            .batch_count = options.batch_count,
+        });
+
+        // Cleaning the buffer first, so it can assert the bytes.
+        @memset(options.buffer, std.math.maxInt(u8));
```
sentientwaffle: Deleting this line doesn't change the test results. What assert is this referring to? If we don't need a particular value then we could randomize the buffer instead, or memset undefined.

src/vsr/multi_batch.zig
batiati: It was only natural that `Postamble` and `TrailerItem` had the same size (u16). To be honest, I haven't considered reserving extra space for future usage (like flags or encoding version).
Unless we can see practical use for it, I think we should keep operations as the primary mechanism for backward compatibility.

src/vsr/multi_batch.zig
batiati: That's true! We removed support here https://github.com/tigerbeetle/tigerbeetle/pull/2787#discussion_r1994336961, so it's illegal to encode zero batches (but we allow zero-element batches)

src/vsr/multi_batch_fuzz.zig
```diff
@@ -0,0 +1,181 @@
...
+            assert(std.mem.eql(u8, expect_batch, decoded_batch));
+        }
+        assert(payloads_decoded == decoder.payload.len);
+        assert(decoder.pop() == null);
+    }
+
+    // Verify that any flipped bit mutates the results or causes a decoding error
+    // (but never a panic).
+    for (0..32) |_| {
+        const byte_index = options.prng.int_inclusive(usize, encoder_bytes_written);
```
sentientwaffle: Since the rng is inclusive, this is off-by-one:
```suggestion
        const byte_index = options.prng.int_inclusive(usize, encoder_bytes_written - 1);
```

src/vsr/multi_batch_fuzz.zig
```diff
@@ -0,0 +1,181 @@
...
+            var same: bool = true;
+            var payloads_decoded: u32 = 0;
+            for (batches.const_slice()) |batch_element_count| {
+                const batch_size: u32 = batch_element_count * batch_element_size;
+                const expect_batch: []u8 = buffer_expected[payloads_decoded..][0..batch_size];
+                payloads_decoded += batch_size;
+
+                const decoded_batch = decoder.pop().?;
+                same = same and std.mem.eql(u8, expect_batch, decoded_batch);
+            }
```
sentientwaffle, line 178:
```suggestion
            }
            assert(decoder.pop() == null);
```

sentientwaffle (approved):

PR #2788, opened 2025-03-03
tidy: switch last few `@memcpy`s and ban them

cb22:


src/clients/go/go_bindings.zig
```diff
@@ -66,7 +66,7 @@ fn to_pascal_case(comptime input: []const u8, comptime min_len: ?usize) []const
             if (is_upper_case(word)) {
                 _ = std.ascii.upperString(output[len..], word);
             } else {
-                @memcpy(output[len..][0..word.len], word);
+                @memcpy(output[len..][0..word.len], word); // Bypass tidy's ban, for go_bindings.
```
cb22, line 69: This is the only complication, and I'm open to a better way of solving this!

`copy_disjoint()` doesn't work at comptime, because `disjoint_slices` -> `@intFromPtr` doesn't. But, you _can_ create overlapping slices at comptime so skipping the check if `@inComptime()` wouldn't be correct.

Having an exception did feel a bit better than having a manual copy loop, which I think could be more error prone than `@memcpy`. That said, as long as this stays at ~1 exception, it's probably OK :thinking:.

matklad: >But, you can create overlapping slices at comptime so skipping the check if @inComptime() wouldn't be correct.

I think built-in @memcpy doest this check at comptime as of 0.14.0, accoring to relnotes?

https://ziglang.org/download/0.14.0/release-notes.html#memcpy-Rules-Adjusted

>Implement aliasing check for comptime @memcpy; a compile error will now be emitted if the arguments alias.

cb22: I saw the `in-memory coercible` point above it, but missed that in the release notes! That's great, it means we could then get rid of this once we're on 0.14.

matklad (approved):

PR #2790, opened 2025-03-03
Use Stack (LIFO) instead of FIFO for free blocks during compaction for better temporal locality. 

toziegler:
LIFO (List In/First Out) instead of FIFO (First In/First Out) causes better temporal locality for free `blocks`  in `lsm/compaction.zig`, reducing memory bandwidth consumption.

The absolute performance gain is marginal (~2%):
 
LIFO: load accepted = 321803 tx/s
FIFO: load accepted = 315314 tx/s

However, this change reduces the memory bandwidth consumption during the benchmark by up to 20% (see plot)
Note that the first few seconds of the benchmark memory is pre-allocated, and at around 10 seconds, the benchmark begins.

[memory_bw.pdf](https://github.com/user-attachments/files/19056136/memory_bw.pdf)
 
Command to measure memory bandwidth: 

AMD:  `AMDuProfPcm -m memory -q -a -d 1000 ./tigerbeetle/tigerbeetle benchmark --cache-grid=32GiB --file=/dev/nvme2n1 --query-count=0`

### DIFF
--- a/src/lifo.zig
+++ b/src/lifo.zig
@@ -18,14 +18,18 @@ pub fn LIFOType(comptime T: type) type {
         name: ?[]const u8,
 
         // If the number of elements is large, the constants.verify check in push() can be too
-        // expensive. Allow the user to gate it. Could also be a comptime param?
+        // expensive. Allow the user to gate it.
         verify_push: bool,
 
-        pub fn init(count_max: u64, name: ?[]const u8, verify_push: bool) LIFO {
+        pub fn init(options: struct {
+            capacity: u64,
+            verify_push: bool,
+            name: ?[]const u8,
+        }) LIFO {
             return LIFO{
-                .name = name,
-                .count_max = count_max,
-                .verify_push = verify_push,
+                .name = options.name,
+                .count_max = options.capacity,
+                .verify_push = options.verify_push,
             };
         }
 
@@ -43,7 +47,7 @@ pub fn LIFOType(comptime T: type) type {
             self.count += 1;
         }
 
-        /// Returns the first element of the stack, and removes it.
+        /// Returns the first element of the LIFO list, and removes it.
         pub fn pop(self: *LIFO) ?*T {
             assert((self.count == 0) == (self.head == null));
 
@@ -54,7 +58,7 @@ pub fn LIFOType(comptime T: type) type {
             return node;
         }
 
-        /// Returns the first element of the stack, but does not remove it.
+        /// Returns the first element of the LIFO list, but does not remove it.
         pub fn peek(self: LIFO) ?*T {
             return self.head;
         }
@@ -88,7 +92,7 @@ pub fn LIFOType(comptime T: type) type {
 }
 
 test "LIFO: fuzz" {
-    // Fuzzy test to compare behavioud of LIFO against ArrayList (reference model).
+    // Fuzzy test to compare behavior of LIFO against std.ArrayList (reference model).
     comptime assert(constants.verify);
 
     const fuzz = @import("testing/fuzz.zig");
@@ -104,7 +108,7 @@ test "LIFO: fuzz" {
     const LIFO = LIFOType(Node);
 
     const node_count_max = 1024;
-    const events_max = 1 << 8;
+    const events_max = 1 << 10;
 
     const Event = enum { push, pop };
     const event_distribution = fuzz.DistributionType(Event){
@@ -123,7 +127,11 @@ test "LIFO: fuzz" {
     var nodes_free = try std.DynamicBitSetUnmanaged.initFull(allocator, node_count_max);
     defer nodes_free.deinit(allocator);
 
-    var stack = LIFO.init(node_count_max, "fuzz", true);
+    var lifo = LIFO.init(.{
+        .capacity = node_count_max,
+        .name = "fuzz",
+        .verify_push = true,
+    });
 
     // Reference model: a dynamic array of node IDs in LIFO order (last is the top).
     var model = try std.ArrayList(u32).initCapacity(allocator, node_count_max);
@@ -131,9 +139,9 @@ test "LIFO: fuzz" {
 
     // Run a sequence of randomized events.
     for (0..events_max) |_| {
-        std.debug.assert(model.items.len <= node_count_max);
-        std.debug.assert(model.items.len == stack.count);
-        std.debug.assert(model.items.len == 0 or !stack.empty());
+        assert(model.items.len <= node_count_max);
+        assert(model.items.len == lifo.count);
+        assert(model.items.len == 0 or !lifo.empty());
 
         const event = fuzz.random_enum(random, Event, event_distribution);
         switch (event) {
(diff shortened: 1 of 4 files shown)


src/lifo.zig
```diff
@@ -0,0 +1,149 @@
...
+
+    lifo.push(&one);
+    lifo.push(&two);
+    lifo.push(&three);
+    try testing.expectEqual(@as(?*Foo, &three), lifo.pop());
+    try testing.expectEqual(@as(?*Foo, &two), lifo.pop());
+    try testing.expectEqual(@as(?*Foo, &one), lifo.pop());
+    try testing.expectEqual(@as(?*Foo, null), lifo.pop());
+    try testing.expect(lifo.empty());
+}
```
matklad: Let's add a randomized test here! (if FIFO is missing one, would be good to add it there as well)

src/lifo.zig
```diff
@@ -0,0 +1,149 @@
...
+            if (constants.verify and self.verify_push) assert(!self.contains(elem));
+            assert(elem.next == null);
+            // Insert the new element at the head.
+            elem.next = self.head;
+            self.head = elem;
+            self.count += 1;
+        }
+
+        pub fn pop(self: *LIFO) ?*T {
+            const ret = self.head orelse return null;
```
matklad: `const result`, we use full names as per TigerStyle! (speaking as a former `res` person)

src/lifo.zig
```diff
@@ -0,0 +1,149 @@
...
+
+        // This should only be null if you're sure we'll never want to monitor `count`.
+        name: ?[]const u8,
+
+        // If the number of elements is large, the constants.verify check in push() can be too
+        // expensive. Allow the user to gate it. Could also be a comptime param?
+        verify_push: bool = true,
+
+        pub fn push(self: *LIFO, elem: *T) void {
+            if (constants.verify and self.verify_push) assert(!self.contains(elem));
```
matklad: Might add `assert((self.count == 0) == (self.head == null))` to push&pop.

src/lifo.zig
```diff
@@ -0,0 +1,149 @@
...
+            var iterator = self.head;
+            while (iterator) |elem| : (iterator = elem.next) {
+                if (elem == elem_needle) return true;
+            }
+            return false;
+        }
+
+        /// Remove an element from the stack. Asserts that the element is in the stack.
+        /// This operation is O(N); if done frequently, consider a different data structure.
+        pub fn remove(self: *LIFO, to_remove: *T) void {
```
matklad: Let's remove `remove`? And make `contains` private, for good measure. We can always generalize later, but better add one more speedbump on the path to accidentally O(N^2) something!

src/lifo.zig
```diff
@@ -0,0 +1,149 @@
...
+        }
+
+        pub fn empty(self: LIFO) bool {
+            return self.head == null;
+        }
+
+        /// Returns whether the linked list contains the given *exact element* (pointer comparison).
+        pub fn contains(self: *const LIFO, elem_needle: *const T) bool {
+            var iterator = self.head;
+            while (iterator) |elem| : (iterator = elem.next) {
```
matklad: Thinking about TigerStyle, WDYT about adding a `count_max` parameter to `init`, and use it to check that the count never overflows, and also to bound these sorts of loops, to tick to items from TigerStyle at the same time:

* always specify explicit limit up-front
* avoid unbounded iteration

Side quest: check if the same can be done for FIFO?

matklad:
@toziegler for institutional knowledge, can you add a short writeup to the comment for how to _measure_ memory bandwidth? 

I don't know how to do that, and I _will_ forget if you tell me, but I might remember that it is written down somewhere!

src/lifo.zig
toziegler: Will add a randomized test.

src/lifo.zig
toziegler: Fair enough. I have essentially copied the class from `FIFO`, should we change it there too? (that is where I have it from)

src/lifo.zig
toziegler: This is a good idea! I will change this here first and open a second PR for the `FIFO` (also to address your comments above there too)

src/lifo.zig
matklad: 😨 definitely should fix it there!

For bonus points, can ban `const|var ret|res =` strings in tidy.zig!

src/lifo.zig
toziegler: I added a fuzzy test that uses `ArrayList` as  a reference model. Is this what you expected?

src/lifo.zig
```diff
@@ -0,0 +1,216 @@
...
+        count_max: u64,
+
+        // This should only be null if you're sure we'll never want to monitor `count`.
+        name: ?[]const u8,
+
+        // If the number of elements is large, the constants.verify check in push() can be too
+        // expensive. Allow the user to gate it. Could also be a comptime param?
+        verify_push: bool,
+
+        pub fn init(count_max: u64, name: ?[]const u8, verify_push: bool) LIFO {
```
matklad:
```suggestion
        pub fn init(options: struct {
            capacity: u64, 
            verify_push: bool,
        }) LIFO {
```

Let's use named args pattern here: bools, numeric arguments, and optionals are usually better when named, and here we have all three.

And, as our stack _is_ statically sized, I think we don't need the `name`. So perhaps for LIFO the right design would be to say:

```
capacity: union(enum) {
    bounded: u32,
    unbounded: struct { name: []const u8 }
}
```

src/lifo.zig
```diff
@@ -0,0 +1,216 @@
...
+        /// Returns whether the linked list contains the given *exact element* (pointer comparison).
+        fn contains(self: *const LIFO, needle: *const T) bool {
+            var iterator = self.head;
+            var count: u64 = 1;
+            while (iterator) |node| : (iterator = node.next) {
+                if (node == needle) return true;
+                count += 1;
+                assert(count <= self.count_max);
+            }
+            return false;
```
matklad: Let's go for a for loop here to make it _syntactically_ obvious that the loop is bounded: 

```suggestion
            assert(self.count <= self.count_max);
            var next = self.head;
            for (0..self.count + 1) |_| {
                const node = next orelse return false;
                if (node == needle) return true;
                next = node.next;
            } else unreachable;
```

As a bonus, this also verifies that `self.count` is correct!

src/lifo.zig
```diff
@@ -0,0 +1,216 @@
...
+
+    var stack = LIFO.init(node_count_max, "fuzz", true);
+
+    // Reference model: a dynamic array of node IDs in LIFO order (last is the top).
+    var model = try std.ArrayList(u32).initCapacity(allocator, node_count_max);
+    defer model.deinit();
+
+    // Run a sequence of randomized events.
+    for (0..events_max) |_| {
+        std.debug.assert(model.items.len <= node_count_max);
```
matklad:
```suggestion
        assert(model.items.len <= node_count_max);
```

We in general are all other the place regarding qualified imports, but `assert` is one instance where we _always_ use it unqualified!

src/lifo.zig
toziegler: I opted for the named parameters and made `capacity` a mandatory parameter as we do not yet have an unbounded use case for `LIFO`.

matklad (approved):

PR #2797, opened 2025-03-05
vopr: search for correlated faults after safety mode

chaitanyabhandari:
This PR solves the following two VOPR failures:

#### ./zig/zig build -Drelease vopr -- 9161948212551073841

In this one, an uncommitted prepare is missing on more than a nack quorum of replicas (it is corrupted on R2, was never replicated to R0, and R1 goes into recovering_head state). Specifically, the cluster ends up getting stuck during view change, and all replicas continuously `on_do_view_change: view=22334 quorum received, awaiting repair`.

This is detected as a liveness failure at the end of safety mode, after which we expect replicas to reply to all requests and upgrade to the latest available release. We expect this because we are very careful about the parts of the grid & WAL where we inject corruptions, and we also take special measures to avoid more than a view change quorum of replicas to go into recovering_head status. Future work (https://github.com/tigerbeetle/tigerbeetle/pull/2708) will get rid of all of this logic that cautiously does storage fault injection.

For now, we roll out part of https://github.com/tigerbeetle/tigerbeetle/pull/2708, specifically the logic that tries to detect correlated faults at the end of safety mode as well (similar to how we do at the end of liveness mode). This would help shake off such false positives, where we accidentally end up injecting more faults than we should during safety mode.  Also, we add logic to the StateChecker to track uncommitted headers, so we can check for their existence in the VOPR.

#### ./zig/zig build -Drelease vopr -- 13232551907466493063

We end up violating one of our invariants at the end of liveness mode here -- a grid block that can be repaired (is present on the core) should be repaired.

Specifically, a replica that is supposed to help out with a block keeps emitting `on_request_blocks: ignoring remaining blocks; busy`. This happened because _no asynchronous callbacks were being invoked_, causing all grid repair read IOPs to get stuck. The ultimate root cause was that paused replicas weren't being unpaused while transitioning to liveness mode, leading to storage not being ticked!

### DIFF
--- a/src/testing/cluster/state_checker.zig
+++ b/src/testing/cluster/state_checker.zig
@@ -65,12 +65,6 @@ pub fn StateCheckerType(comptime Client: type, comptime Replica: type) type {
             errdefer allocator.free(replica_head_max);
             for (replica_head_max) |*head| head.* = .{ .view = 0, .op = 0 };
 
-            const pipeline_max = constants.pipeline_prepare_queue_max;
-            var uncommitted_headers: [pipeline_max]vsr.Header.Prepare = undefined;
-            for (0..pipeline_max) |slot| {
-                uncommitted_headers[slot] = vsr.Header.Prepare.reserve(options.cluster_id, slot);
-            }
-
             return StateChecker{
                 .node_count = @intCast(options.replicas.len),
                 .replica_count = options.replica_count,
@@ -260,7 +254,7 @@ pub fn StateCheckerType(comptime Client: type, comptime Replica: type) type {
             }
         }
 
-        pub fn committed_header_with_op(state_checker: *StateChecker, op: u64) vsr.Header.Prepare {
+        pub fn header_with_op(state_checker: *StateChecker, op: u64) vsr.Header.Prepare {
             assert(op < state_checker.commits.items.len);
             const commit = &state_checker.commits.items[op];
             assert(commit.header.op == op);
--- a/src/testing/cluster/state_checker.zig
+++ b/src/testing/cluster/state_checker.zig
@@ -65,6 +65,12 @@ pub fn StateCheckerType(comptime Client: type, comptime Replica: type) type {
             errdefer allocator.free(replica_head_max);
             for (replica_head_max) |*head| head.* = .{ .view = 0, .op = 0 };
 
+            const pipeline_max = constants.pipeline_prepare_queue_max;
+            var uncommitted_headers: [pipeline_max]vsr.Header.Prepare = undefined;
+            for (0..pipeline_max) |slot| {
+                uncommitted_headers[slot] = vsr.Header.Prepare.reserve(options.cluster_id, slot);
+            }
+
             return StateChecker{
                 .node_count = @intCast(options.replicas.len),
                 .replica_count = options.replica_count,
--- a/src/testing/cluster/state_checker.zig
+++ b/src/testing/cluster/state_checker.zig
@@ -260,7 +260,8 @@ pub fn StateCheckerType(comptime Client: type, comptime Replica: type) type {
             }
         }
 
-        pub fn header_with_op(state_checker: *StateChecker, op: u64) vsr.Header.Prepare {
+        pub fn committed_header_with_op(state_checker: *StateChecker, op: u64) vsr.Header.Prepare {
+            assert(op < state_checker.commits.items.len);
             const commit = &state_checker.commits.items[op];
             assert(commit.header.op == op);
             assert(commit.replicas.count() > 0);
--- a/src/vopr.zig
+++ b/src/vopr.zig
@@ -212,16 +212,21 @@ pub fn main() !void {
     } else {
         if (cli_args.lite) return;
 
+        // Safety mode ran out of ticks without completing its requests, so now we check whether it
+        // was correct to do so.
+        //
         // Heal current network partitions, disable future process, storage, and network faults
         // on all replicas. Run this fully-connected core of replicas for 500_000 ticks, which
         // should enough to ensure all faulty grid blocks, headers, or prepares that can be repaired
         // are repaired. After this, all we must be left with are correlated faults.
         {
             var replica: u8 = 0;
+            var core: Core = Core.initEmpty();
             while (replica < simulator.options.cluster.replica_count) : (replica += 1) {
-                simulator.core.set(replica);
+                core.set(replica);
             }
-            simulator.transition_to_liveness_mode(simulator.core);
+
+            simulator.transition_to_liveness_mode(core);
 
             tick = 0;
             while (tick < 500_000) : (tick += 1) simulator.tick();
@@ -240,13 +245,13 @@ pub fn main() !void {
 
     if (cli_args.lite) return;
 
-    simulator.core = random_core(
+    const core = random_core(
         simulator.prng,
         simulator.options.cluster.replica_count,
         simulator.options.cluster.standby_count,
     );
 
-    simulator.transition_to_liveness_mode(simulator.core);
+    simulator.transition_to_liveness_mode(core);
 
     // Liveness: a core set of replicas is up and fully connected. The rest of replicas might be
     // crashed or partitioned permanently. The core should converge to the same state.
@@ -594,7 +599,7 @@ pub const Simulator = struct {
         for (simulator.cluster.replicas) |*replica| {
             if (simulator.core.isSet(replica.replica)) {
                 for (replica.op_checkpoint() + 1..commit_max + 1) |op| {
-                    const header = simulator.cluster.state_checker.header_with_op(op);
+                    const header = simulator.cluster.state_checker.committed_header_with_op(op);
                     if (!replica.journal.has_prepare(&header)) return "pending journal";
                 }
                 // It's okay for a replica to miss some prepares older than the current checkpoint.
@@ -659,6 +664,10 @@ pub const Simulator = struct {
     /// context.
     pub fn transition_to_liveness_mode(simulator: *Simulator, core: Core) void {
         log.debug("transition_to_liveness_mode: core={b}", .{core.mask});
+        assert(simulator.core.count() == 0);
+        defer assert(simulator.core.count() > 0);
+
+        simulator.core = core;
 
         var it = core.iterator(.{});
         while (it.next()) |replica_index| {
@@ -731,6 +740,28 @@ pub const Simulator = struct {
         return quorums.view_change > core_replicas - core_recovering_head;
     }
 
+    fn core_repairable_replica(
+        simulator: *const Simulator,
+        comptime Replica: type,
+        replica: *const Replica,
+    ) bool {
+        if (!simulator.core.isSet(replica.replica)) return false;
+        if (replica.standby()) return false;
+        switch (replica.status) {
+            .normal => return true,
+            .recovering_head => return false,
+            .view_change => {
+                // Lagging replicas do not initiate WAL repair during view change.
+                if (vsr.Checkpoint.durable(replica.op_checkpoint_next(), replica.commit_max)) {
+                    return false;
+                } else {
+                    return true;
+                }
+            },
+            .recovering => unreachable,
+        }
+    }
+
     // Returns a header for a prepare which can't be repaired by the core due to storage faults.
     //
     // If a replica cannot make progress on committing, then it may be stuck while repairing either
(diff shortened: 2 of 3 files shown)


src/vopr.zig
```diff
@@ -624,11 +636,6 @@ pub const Simulator = struct {
     /// See https://tigerbeetle.com/blog/2023-07-06-simulation-testing-for-liveness for broader
     /// context.
     pub fn transition_to_liveness_mode(simulator: *Simulator) void {
```
sentientwaffle: Since the correct way to use this function is to set `simulator.core` immediately beforehand, how about we instead pass it the `core` bitset as an argument?

src/vopr.zig
```diff
@@ -637,6 +644,12 @@ pub const Simulator = struct {
             if (simulator.cluster.replica_health[replica_index] == .down) {
                 simulator.replica_restart(@intCast(replica_index), fault);
             }
+
+            const replica_health = simulator.cluster.replica_health[replica_index];
+            if (replica_health == .up and replica_health.up.paused) {
+                simulator.cluster.replica_unpause(@intCast(replica_index));
+            }
```
sentientwaffle, line 682: This bug is a great find!
(I am also surprised it took us this long to hit a seed that got stuck this way.)

src/vopr.zig
```diff
@@ -696,117 +709,104 @@ pub const Simulator = struct {
...
+        for (replicas_missing_ops, 0..) |*replicas_missing_op, index| {
+            replicas_missing_op.* = ReplicaSet.initEmpty();
+            const op = index + cluster_op_repair_min;
+            const header = simulator.cluster.state_checker.header_with_op(op);
+            for (simulator.cluster.replicas) |replica| {
+                if (replica.standby()) continue;
+                if (!simulator.core.isSet(replica.replica) or
+                    !replica.journal.has_prepare(&header))
+                {
+                    replicas_missing_ops[index].set(replica.replica);
```
```suggestion
                    replicas_missing_op.set(replica.replica);
```

src/testing/cluster/state_checker.zig
```diff
@@ -255,10 +269,18 @@ pub fn StateCheckerType(comptime Client: type, comptime Replica: type) type {
...
-            assert(commit.replicas.count() > 0);
-            return commit.header;
+            const header = blk: {
+                if (op < state_checker.commits.items.len) {
+                    const commit = &state_checker.commits.items[op];
+                    assert(commit.replicas.count() > 0);
+                    break :blk commit.header;
+                } else {
+                    const slot = op % constants.pipeline_prepare_queue_max;
+                    break :blk state_checker.uncommitted_headers[slot];
```
sentientwaffle: How do we know that this is the uncommitted prepare that the cluster is stuck on?
e.g. if we prepare `(view=1, op=10)` and then prepare `(view=2, op=10)`, the latter is in `uncommitted_headers`, but a `view=3` might select the former.

src/testing/cluster/state_checker.zig
```diff
@@ -93,6 +101,12 @@ pub fn StateCheckerType(comptime Client: type, comptime Replica: type) type {
             }
         }
 
+        pub fn on_new_prepare(state_checker: *StateChecker, prepare: *const Message.Prepare) void {
+            assert(prepare.header.command == .prepare);
```
sentientwaffle: Can we also assert that the op number does not decrease in a given slot?

sentientwaffle: Also can we assert that the op is not more than pipeline max ahead of the currently highest committed op.

src/vopr.zig
```diff
@@ -363,36 +363,41 @@ pub fn main() !void {
 
         if (requests_done and upgrades_done) break;
     } else {
+        if (cli_args.lite) return;
+
+        // Heal current network partitions, disable future process, storage, and network faults
```
sentientwaffle, line 218: Maybe add one more sentence to the start of this comment for context. Something along the lines of "safety mode ran out of ticks without completing its requests, so now we check whether it was correct to do so".

src/vopr.zig
```diff
@@ -363,36 +363,41 @@ pub fn main() !void {
...
     if (cli_args.lite) return;
 
-    simulator.transition_to_liveness_mode();
+    simulator.core = random_core(
+        simulator.prng,
+        simulator.options.cluster.replica_count,
+        simulator.options.cluster.standby_count,
+    );
+
+    simulator.transition_to_liveness_mode(simulator.core);
```
sentientwaffle: I was imagining that we could create the core with `const core = ...` then `transition_to_liveness_mode` would be responsible for assigning it to `simulator.core`. (And also it could assert the old core state?)

src/vopr.zig
```diff
@@ -715,117 +728,104 @@ pub const Simulator = struct {
...
-                            replica.commit_min,
-                        )) |op| {
-                            if (missing_prepare_op == null or op < missing_prepare_op.?) {
-                                missing_prepare_op = op;
-                            }
-                        }
-                    }
+        for (replicas_missing_ops, 0..) |*replicas_missing_op, index| {
+            replicas_missing_op.* = ReplicaSet.initEmpty();
+            const op = index + cluster_op_repair_min;
```
sentientwaffle: Instead of starting the iteration index from 0 could we start it from `cluster_op_repair_min`? Then we would be iterating the ops directly.

src/vopr.zig
```diff
@@ -835,26 +835,29 @@ pub const Simulator = struct {
     ) error{OutOfMemory}!?usize {
         assert(simulator.core.count() > 0);
 
-        var blocks_missing = std.ArrayList(struct {
-            replica: u8,
-            address: u64,
-            checksum: u128,
-        }).init(allocator);
+        const FaultyReplicas = std.StaticBitSet(constants.replicas_max + constants.standbys_max);
```
```suggestion
        const FaultyReplicas = std.StaticBitSet(constants.members_max);
```

src/vopr.zig
```diff
@@ -912,10 +920,10 @@ pub const Simulator = struct {
             }
         }
 
-        if (blocks_missing.items.len == 0) {
+        if (blocks_missing.count() == 0) {
```
sentientwaffle, line 982: Ah, so before if the same block was missing on multiple replicas then we would count it as missing multiple times.

src/vopr.zig
chaitanyabhandari: Ah, okay! I misunderstood the intention. Thanks for clarifying.

src/testing/cluster/state_checker.zig
```diff
@@ -65,6 +65,12 @@ pub fn StateCheckerType(comptime Client: type, comptime Replica: type) type {
             errdefer allocator.free(replica_head_max);
             for (replica_head_max) |*head| head.* = .{ .view = 0, .op = 0 };
 
+            const pipeline_max = constants.pipeline_prepare_queue_max;
+            var uncommitted_headers: [pipeline_max]vsr.Header.Prepare = undefined;
+            for (0..pipeline_max) |slot| {
+                uncommitted_headers[slot] = vsr.Header.Prepare.reserve(options.cluster_id, slot);
+            }
```
sentientwaffle: This code can be removed.

src/testing/cluster/state_checker.zig
```diff
@@ -254,7 +260,8 @@ pub fn StateCheckerType(comptime Client: type, comptime Replica: type) type {
             }
         }
 
-        pub fn header_with_op(state_checker: *StateChecker, op: u64) vsr.Header.Prepare {
+        pub fn committed_header_with_op(state_checker: *StateChecker, op: u64) vsr.Header.Prepare {
```
sentientwaffle: Since we are back to all of the state checkers headers being committed we can move that from the name.

src/vopr.zig
```diff
@@ -715,18 +737,26 @@ pub const Simulator = struct {
...
+        replica: *const Replica,
+    ) bool {
+        if (!simulator.core.isSet(replica.replica)) return false;
+        if (replica.standby()) return false;
+        switch (replica.status) {
+            .normal => return true,
+            .recovering_head => return false,
+            .view_change => {
+                // Lagging replicas do not initiate WAL repair during view change.
+                if (vsr.Checkpoint.durable(replica.op_checkpoint_next(), replica.commit_max)) {
```
sentientwaffle: This could instead simply be `return !vsr.Checkpoint.durable(replica.op_checkpoint_next(), replica.commit_max)`.

src/vopr.zig
```diff
@@ -736,96 +766,117 @@ pub const Simulator = struct {
...
+            if (!simulator.core_repairable_replica(Cluster.Replica, &replica)) continue;
+
+            cluster_commit_max = @max(cluster_commit_max, replica.commit_max);
+            cluster_log_view = @max(cluster_log_view, replica.log_view);
+            cluster_op_head = @max(cluster_op_head, replica.op);
+            if (op_repair_min == null) {
+                op_repair_min = replica.op_repair_min();
+            } else {
+                op_repair_min = @min(op_repair_min.?, replica.op_repair_min());
             }
```
sentientwaffle: If we initialize `op_repair_min` to `std.math.maxInt(u16)` then we can simplify this and avoid the branching.

src/vopr.zig
```diff
@@ -736,96 +766,117 @@ pub const Simulator = struct {
...
-        for (simulator.cluster.replicas) |*replica| {
-            if (simulator.core.isSet(replica.replica) and !replica.standby()) {
-                cluster_op_checkpoint = @max(
-                    cluster_op_checkpoint,
-                    replica.superblock.working.vsr_state.checkpoint.header.op,
-                );
+        var cluster_op_head: u64 = 0;
+        var cluster_commit_max: u64 = 0;
+        var cluster_log_view: u32 = 0;
+        var op_repair_min: ?u64 = 0;
```
sentientwaffle: I think this is a bug -- by initializing this to zero, the minimum will always be zero.

src/vopr.zig
```diff
@@ -736,96 +766,117 @@ pub const Simulator = struct {
...
+                op_repair_min = @min(op_repair_min.?, replica.op_repair_min());
             }
         }
+        const cluster_op_repair_min = op_repair_min.?;
+
+        // Use replicas with the largest log_view to infer uncommitted headers. Replicas with a
+        // smaller log_view may have an outdated version of the same uncommitted op. There may be
+        // at most a pipeline of uncommitted headers in the cluster.
+        const pipeline_max = constants.pipeline_prepare_queue_max;
+        var uncommitted_headers: [pipeline_max]vsr.Header.Prepare = undefined;
```
sentientwaffle: Maybe use an optional type (accessed later with `.?`) for the array entries just to be careful.

src/vopr.zig
```diff
@@ -736,96 +766,117 @@ pub const Simulator = struct {
...
+
+            cluster_commit_max = @max(cluster_commit_max, replica.commit_max);
+            cluster_log_view = @max(cluster_log_view, replica.log_view);
+            cluster_op_head = @max(cluster_op_head, replica.op);
+            if (op_repair_min == null) {
+                op_repair_min = replica.op_repair_min();
+            } else {
+                op_repair_min = @min(op_repair_min.?, replica.op_repair_min());
             }
         }
```
sentientwaffle, line 789: Can we assert that the `cluster_commit_max` is less than or equal to the `cluster_op_head`.

src/vopr.zig
```diff
@@ -736,96 +766,117 @@ pub const Simulator = struct {
...
-                    if (op_repair_min <= replica.commit_min) {
-                        if (simulator.smallest_missing_prepare_between(
-                            &replica,
-                            op_repair_min,
-                            replica.commit_min,
-                        )) |op| {
-                            if (missing_prepare_op == null or op < missing_prepare_op.?) {
-                                missing_prepare_op = op;
-                            }
+                if (cluster_commit_max < cluster_op_head) {
```
sentientwaffle: We already checked this condition above.

src/vopr.zig
```diff
@@ -736,96 +766,117 @@ pub const Simulator = struct {
...
+                }
+            }
+        }
 
-        const missing_op = if (missing_header_op) |op| op else missing_prepare_op.?;
-        const missing_header = simulator.cluster.state_checker.header_with_op(missing_op);
+        // Check whether any of the uncommitted headers is corrupted on more than a nack
+        // quorum of replicas. If so, the cluster cannot initiate repair or commit (see the
+        // awaiting_repair and complete_invalid cases in the DVCQuorum).
+        if (cluster_commit_max <= cluster_op_head) {
```
sentientwaffle: Hmm, I mentioned above that I thought this could be an assert, but maybe it can't be? When is `cluster_commit_max <= cluster_op_head` not true?

src/testing/cluster/state_checker.zig
chaitanyabhandari: Oops, nice catch! For this PR, I was trying to contain all related changes within a single commit but oh well... 😄

src/vopr.zig
chaitanyabhandari: Ah, I wanted this to be null! Great catch. 😶‍🌫️ 

Anyway, https://github.com/tigerbeetle/tigerbeetle/pull/2797#discussion_r1995775096 would obviate the need for this!

src/vopr.zig
chaitanyabhandari: Mhmmmmm, I actually think that it _can_ be an asserted, IIUC we always update commit_max & op_head in lockstep (via `set_op_and_commit_max`)? This was more an unconscious condition that I added, because I was adding it for `replica.commit_min < replica.commit_max` and `replica.op_repair_min() <= replica.commit_min`.

I'll add it!

chaitanyabhandari:
Thanks for the great comments @sentientwaffle, I'm much happier with how the changes look now! I've addressed the comments in ca06d7517e9007814ca06e51dba791f10045cae1.

src/vopr.zig
```diff
@@ -737,96 +766,125 @@ pub const Simulator = struct {
...
+            cluster_op_head - cluster_op_repair_min + 1,
+        );
+        errdefer allocator.free(replicas_missing_ops);
+
+        for (replicas_missing_ops, cluster_op_repair_min..) |*replicas_missing_op, op| {
+            replicas_missing_op.* = ReplicaSet.initEmpty();
+            const header = blk: {
+                if (op > cluster_commit_max) {
+                    const uncommitted_header = uncommitted_headers[op % pipeline_max].?;
+                    assert(uncommitted_header.op > cluster_commit_max);
```
sentientwaffle: Ah actually since we just checked `op > cluster_commit_max` we could check `uncommitted_header.op == op`.

src/vopr.zig
```diff
@@ -737,96 +766,125 @@ pub const Simulator = struct {
...
+        // Check whether any of the uncommitted headers is corrupted on more than a nack
+        // quorum of replicas. If so, the cluster cannot initiate repair or commit (see the
+        // awaiting_repair and complete_invalid cases in the DVCQuorum).
+        const nack_quorum = vsr.quorums(replica_count).nack_prepare;
+        for (cluster_commit_max..cluster_op_head + 1) |op| {
+            if (replicas_missing_ops[op - cluster_op_repair_min].count() >= nack_quorum) {
+                const header = blk: {
+                    if (op > cluster_commit_max) {
+                        const uncommitted_header = uncommitted_headers[op % pipeline_max].?;
+                        assert(uncommitted_header.op > cluster_commit_max);
```
sentientwaffle: Same as https://github.com/tigerbeetle/tigerbeetle/pull/2797/files#r1996043454

src/vopr.zig
chaitanyabhandari: Ack, done in https://github.com/tigerbeetle/tigerbeetle/pull/2797/commits/485051fa5a0608fbb299552a7e1b1fc73072a913.

sentientwaffle (approved):

src/testing/cluster/state_checker.zig
chaitanyabhandari: This version of the code was indeed vulnerable to this. It was solved via https://github.com/tigerbeetle/tigerbeetle/pull/2797/commits/20710a086c6af03eaac0ce4fa0b66f9bc840cb9c.

PR #2798, opened 2025-03-06
stdx: vendor PRNG

matklad:
Our usage of std.random.Random has been bugging me since forever for a bunch of tiny reasons:

- It's API is very wordy and it's easy to confuse inclusive/exclusive cases
- It's hard to extend it with natural convenience functions, such as `int_exponential`.
- We have three ways of modeling probabilities:
  - an int ≤ 100
  - a float ≤ 1.0
  - a float ≤ 100.0

  Ideally, we'd just use rational ratios uniformly, but that means
  that we again requre inconvenient extensions.
- PRNG algorithms tend to change, often for reasons not applicable to us (we need neither the fastest, nor the most secure PRNG). It's better if we rotate our PRNG algorithm at our own pace.

While working on this PR, I realized there's one more benefit: we now need an explicit pointer for *PRNG as oppose to vtable of std.Random, which makes it easier to see when PRNGs are independent and when they are not.


matklad:
I am quite happy with how this turned out and think we should do it!

- This was a surprise to me, but turns out you want to use inclusive ranges with random numbers. Most call-sites need inclusive, and of the remaining call-sites, _most_ are `.index` or `.chance` 
- We haven't _fully_ got rid of floating point --- `random_int_exponential` and `zipfian` need need floats, so we fall-back to std implementation there. 
- I am slightly unsure about `prng.bytes(u128)` function that returns a fully random int, but, looking at the usages, I think that's right. You use it when you need seed or fully random id, and there you _do_ treat things as random bytes, not as integers. I can rename it to `int`
- I choose PRNG as the canonical name, to not confuse it with std.Random, to underscore that this is not an OS-based source of randomness, and to leave the door open for arbtest-style FRNG (finite rng) to integrate with fuzzing.

src/stdx/prng.zig
```diff
@@ -0,0 +1,533 @@
...
+}
+
+/// Returns a random value of an enum.
+pub fn enum_uniform(prng: *PRNG, Enum: type) Enum {
+    const values = std.enums.values(Enum);
+    return values[prng.index(values)];
+}
+
+test enum_uniform {
+    const E = enum(u8) { a, b, c = 8 };
```
batiati: Why `c = 8`?
Please add a comment if it has any meaning for this test. Idem for `test enum_weighted`.

src/stdx/prng.zig
```diff
@@ -0,0 +1,533 @@
...
+const Snap = @import("../testing/snaptest.zig").Snap;
+const snap = Snap.snap;
+
+s: [4]u64,
+
+const PRNG = @This();
+
+/// A less than one rational number, used to specify probabilities.
+pub const Ratio = struct {
+    // Invariant: numerator < denomitator.
```
batiati: `ratio()` asserts `numerator <= denominator`
```suggestion
    // Invariant: numerator <= denomitator.
```

src/stdx/prng.zig
```diff
@@ -0,0 +1,533 @@
...
+        \\large=506 small=494
+    ).diff_fmt("large={} small={}", .{ large, small });
+}
+
+// Deliberately excluded from the API to normalize everything to closed ranges.
+// Somewhat surprisingly, closed ranges are more conveinient for generating random numbers:
+// - passing zero is not a subtle error
+// - passing intMax allows generating any integer
+// - at the call-site, inclusive is usually somewhat more obvious.
+pub const int_exclusive = void;
```
batiati: Suggestion: could we use `@compileError()` instead of `void`?

src/stdx/prng.zig
```diff
@@ -0,0 +1,533 @@
...
+    while (i < aligned_len) : (i += 8) {
+        var n = prng.next();
+        comptime var j: usize = 0;
+        inline while (j < 8) : (j += 1) {
+            target[i + j] = @as(u8, @truncate(n));
+            n >>= 8;
+        }
+    }
+
+    // Remaining. (cuts the stream)
```
batiati:
```suggestion
    // Remaining (cuts the stream).
```

batiati (commented):
> I am quite happy with how this turned out and think we should do it!

Indeed, it's delightful not having to deal with all those variants of int/uint biased/unbiased inclusive/exclusive when simply wanting a random integer :smiley: .

What is preventing us from replacing all usages of `std.rand.Random` with `stdx.PRNG` and implementing a `tidy` ban?

src/stdx/prng.zig
matklad: That's to make sure we properly convert from a randomly chosen index to a random discriminant, added a comment

src/stdx/prng.zig
matklad: Good catch!

src/stdx/prng.zig
matklad: 👍 I thought `@compileError` would not work here, but it seems it does!

matklad:
>What is preventing us from replacing all usages of std.rand.Random with stdx.PRNG and implementing a tidy ban? 

We still rely on std.Random for benchmarks (zipfian) and fuzzing (int_exponential), as those two distributions require floating point to draw from. 

I don't want to add floating point support to PRNG (as we use it both for testing and for production code).

And for bans, I think I'd rather keep it simple and either ban every single instance, or not ban anything. If we have to have some carve outs, I'd rather we not ban until we find a way to remove the carve-outs!

batiati:
> We still rely on std.Random for benchmarks (zipfian) and fuzzing (int_exponential), as those two distributions require floating point to draw from.
> I don't want to add floating point support to PRNG (as we use it both for testing and for production code).


I did grep for `std.rand` and tried the API myself, replacing everything I could with `stdx.PRNG`. 
Please take a look at the commit https://github.com/tigerbeetle/tigerbeetle/commit/38913314835a3523e87694819245e6884990e2d4

Only `clock.zig` and `time.zig` still require float randoms.
Is there any reason this PR didn't change the others?

Suggestions:

- Yes, please rename `prng.bytes` to `prng.int`!

- Rename `prng.range_inclusive` as `prng.int_range_inclusive` to align with `prng.int()` and `prng.int_inclusive()`.
  Not a super strong opinion here.

- `prng.index` is pretty handy, but sometimes we still need to resort to `prng.int_inclusive(size, count  - 1)` when we don't have a slice (for example, bounded arrays).

EDIT: Ah, `enum_weighted` is :man_cook: :pinched_fingers: :heart:

matklad:
>I did grep for std.rand and tried the API myself

😨 I was only looking for `std.Random.DefaultPrng`, I didn't realize that `std.rand.DefaultPrng` is also a thing, thanks!

batiati:
> > I did grep for std.rand and tried the API myself
> 
> 😨 I was only looking for `std.Random.DefaultPrng`, I didn't realize that `std.rand.DefaultPrng` is also a thing, thanks!

Feel free to cherry pick this commit

matklad:
Ok, I went ahead and banned the underlying integer API from std.Random, without banning Random as a whole!

>Yes, please rename prng.bytes to prng.int!

Done!

>Rename prng.range_inclusive as prng.int_range_inclusive to align with prng.int() and prng.int_inclusive().

I kept `range_inclusive`! The rationale here is that I intentionally do not want to provide any _options_ to have inclusive/exclusive/int/uint, etc. So I don't need `int_` prefix to disambiguate an `int_` range from some other range --- unsigned integer is your only option! The reason why we need `int_inclusive` and `int` is simply because we need some sort of a noun here. 

No strong opinion here, would be happy to do `int_range_inclusive` if you think this is on balance better!

>prng.index is pretty handy, but sometimes we still need to resort to prng.int_inclusive(size, count  - 1) when we don't have a slice (for example, bounded arrays).

Yup! You still often can get a slice from somewhere (https://github.com/tigerbeetle/tigerbeetle/pull/2798/commits/b2cc1570608d271a4119d3c70a3fc85309420bdf#diff-0c3113ecd6426b3e79e38198677f0ab7a62d5b3c9f40036b24e5afe8c2db6634R191), but there are cases where you need to do `.count() - 1`. Which I think is fine, and better than having to decide between exclusive and inclusive at the API method level.

batiati (approved):

PR #2799, opened 2025-03-06
Clients/Java: UInt128.asBytes disallow negative bigint

sentientwaffle:


src/clients/java/src/main/java/com/tigerbeetle/UInt128.java
```diff
@@ -150,9 +150,13 @@ public static BigInteger asBigInteger(final byte[] bytes) {
      * @return an array of 16 bytes representing the 128-bit value.
```
batiati:
```suggestion
     * @return an array of 16 bytes representing the 128-bit unsigned value.
```

batiati (commented):
Nice catch!

The .Net version already forbids negative values (`isUnsigned: true`)
https://github.com/tigerbeetle/tigerbeetle/blob/1dd3ec1bd787124b83a2ee2bd10bf30f94ce2ea4/src/clients/dotnet/TigerBeetle/UInt128Extensions.cs#L102-L115

I _think_ the Go version might need validation, as [`big.Int.SetBytes`](https://pkg.go.dev/math/big#Int.SetBytes) encodes as unsigned, while [big.Int.Bytes](https://pkg.go.dev/math/big#Int.Bytes) says "absolute value" (not sure what "absolute" means regarding the binary encoding ... will -1 be 1 or maxInt?).

batiati (approved):
:+1:

PR #2801, opened 2025-03-06
vsr: remove suprious resume in client_replies

matklad:
As far as I can tell, there's no specific reason to resume writing after read completes: read and writes race, when writing, we only check that the slot is not being written to.


sentientwaffle (approved):

matklad:
VOPR green at 170 657

PR #2806, opened 2025-03-08
docs: adds incrementing to the end of the batch in the example

IvoCrnkovic:
I think the end of the batch needs to be incremented after you submit the batch.


matklad:
Indeed, thanks!

matklad (approved):

PR #2809, opened 2025-03-10
[WIP] `create_transfers_with_balance`

batiati:
### Topics to discuss:

- The name `create_transfers_with_balance` may not fully capture its purpose, as it also includes the transfer's timestamp and amount.

- Extra lookups are performed when returning results, which could be avoided. However, we should prioritize not adding overhead for `create_transfers`.

- The result size is 160 bytes, not a power of two but evenly divisible by `message_body_max`, allowing 6552 transfers per request.

### TODO:

Clients, docs, unit tests, and the VOPR workload.

### Benchmark:

```bash
./tigerbeetle benchmark --transfer-batch-size=6552
1527 batches in 57.53 s
transfer batch size = 6552 txs
transfer batch delay = 0 us
load accepted = 173822 tx/s
batch latency p1 = 9 ms
batch latency p10 = 13 ms
batch latency p20 = 14 ms
batch latency p30 = 20 ms
batch latency p40 = 24 ms
batch latency p50 = 29 ms
batch latency p60 = 37 ms
batch latency p70 = 41 ms
batch latency p80 = 49 ms
batch latency p90 = 53 ms
batch latency p95 = 61 ms
batch latency p99 = 190 ms
batch latency p100 = 257 ms
```

```bash
./tigerbeetle benchmark --create-transfers-with-balance
1527 batches in 59.05 s
transfer batch size = 6552 txs
transfer batch delay = 0 us
load accepted = 169339 tx/s
batch latency p1 = 9 ms
batch latency p10 = 16 ms
batch latency p20 = 16 ms
batch latency p30 = 20 ms
batch latency p40 = 24 ms
batch latency p50 = 30 ms
batch latency p60 = 38 ms
batch latency p70 = 42 ms
batch latency p80 = 50 ms
batch latency p90 = 55 ms
batch latency p95 = 61 ms
batch latency p99 = 194 ms
batch latency p100 = 218 ms
```


src/state_machine.zig
```diff
@@ -2076,6 +2121,170 @@ pub fn StateMachineType(
...
+                        .timestamp = transfer.timestamp,
+                        .amount = transfer.amount,
+                        .debit_account_balance_debits_pending = dr_account.debits_pending,
+                        .debit_account_balance_debits_posted = dr_account.debits_posted,
+                        .debit_account_balance_credits_pending = dr_account.credits_pending,
+                        .debit_account_balance_credits_posted = dr_account.credits_posted,
+                        .credit_account_balance_debits_pending = cr_account.debits_pending,
+                        .credit_account_balance_debits_posted = cr_account.debits_posted,
+                        .credit_account_balance_credits_pending = cr_account.credits_pending,
+                        .credit_account_balance_credits_posted = cr_account.credits_posted,
```
sentientwaffle, line 2177: Including the balance for `exists` might be confusing since the balance is now different than it was when the transfer was initially created. Though then again not including it might be worse, since most applications are supposed to treat `ok` and `exists` identically.

src/state_machine.zig
```diff
@@ -518,6 +520,7 @@ pub fn StateMachineType(
             query_accounts = config.vsr_operations_reserved + 7,
             query_transfers = config.vsr_operations_reserved + 8,
             get_events = config.vsr_operations_reserved + 9,
+            create_transfers_with_balance = config.vsr_operations_reserved + 10,
```
sentientwaffle, line 523: My main concern with `create_transfers_with_balance` is that if an application mixes it with `create_transfers` then they will unintentionally and perhaps surprisingly sabotage their batching. I guess we could warn about this in the client since we can detect it automatically, though that seems a little bit awkward. But either way seems like they have to make an application wide decision.

Though now that I think about it I suppose the client could promote a `create_transfers` to a `create_transfers_with_balance` automatically in the client if it sees them mixed...

src/state_machine.zig
batiati: Honestly, neither option is good. Ideally, we would have to retrieve the balance from `account_events` and return it "as of" when the transfer succeeded. But we have plans to make history prunable, so not all transfers will have the history available.

In any case, I'd err to on side of not having the balance on `exists` rather than having the most current balance.

src/state_machine.zig
batiati: > Though now that I think about it I suppose the client could promote a create_transfers to a create_transfers_with_balance automatically in the client if it sees them mixed...

It's doable, we could rewrite the output in-place since the result of `create_transfers_with_balance` is always larger than `create_transfers`. But this logic is too specific to the state machine, which feels wrong to have at the `tb_client` layer!

I think we're ok with the application mixing operations.

batiati:
surpassed by https://github.com/tigerbeetle/tigerbeetle/pull/3080

PR #2811, opened 2025-03-11
bus: idiomatic log message fromat

matklad:
- start with process id
- add method name
- use key=value pairs

I've chosen to use decimal, rather than hex, for client id, as Peer prints them in decimal.


matklad:
Sample message:

```
2025-03-11 12:21:57.249Z warning(message_bus): 313120583216297804418423706816073194871: on_connect: error to=2 error.ConnectionRefused
```

cb22 (approved):

PR #2814, opened 2025-03-11
Changes required by Zig 0.14.0 but applicable in Zig 0.13.0

batiati:
In preparation for #2705.

### DIFF
--- a/src/io/windows.zig
+++ b/src/io/windows.zig
@@ -1394,8 +1394,17 @@ pub const IO = struct {
             };
         }
 
+        const set_end_of_file = @extern(
+            *const fn (
+                hFile: os.windows.HANDLE,
+            ) callconv(os.windows.WINAPI) std.os.windows.BOOL,
+            .{
+                .library_name = "kernel32",
+                .name = "SetEndOfFile",
+            },
+        );
         // Mark the moved file pointer (start + size) as the physical EOF.
-        const allocated = os.windows.kernel32.SetEndOfFile(handle);
+        const allocated = set_end_of_file(handle);
         if (allocated == os.windows.FALSE) {
             const err = os.windows.kernel32.GetLastError();
             return os.windows.unexpectedError(err);
--- a/src/vsr/journal.zig
+++ b/src/vsr/journal.zig
@@ -1525,7 +1525,7 @@ pub fn JournalType(comptime Replica: type, comptime Storage: type) type {
                     assert(prepare.?.operation == .reserved);
                     assert(header.?.checksum == prepare.?.checksum);
                     assert(
-                        header.?.checksum == Header.Prepare.reserved(cluster, slot.index).checksum,
+                        header.?.checksum == Header.Prepare.reserved_operation(cluster, slot.index).checksum,
                     );
                     assert(!journal.prepare_inhabited[slot.index]);
                     assert(journal.prepare_checksums[slot.index] == 0);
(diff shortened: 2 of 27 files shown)


src/vsr/message_header.zig
```diff
@@ -721,7 +721,7 @@ pub const Header = extern struct {
             return null;
         }
 
-        pub fn reserved(cluster: u128, slot: u64) Prepare {
+        pub fn reserved_operation(cluster: u128, slot: u64) Prepare {
```
matklad: Not a huge fan of this name, but nothing better comes to mind. Maybe `reserved_slot`? Anyway, shouldn't be a problem to rename later!

src/io/windows.zig
```diff
@@ -1394,8 +1394,19 @@ pub const IO = struct {
...
+        // by convention, used for building generic types.
+        const set_end_of_file = @extern(
+            *const fn (
+                hFile: os.windows.HANDLE,
+            ) callconv(os.windows.WINAPI) std.os.windows.BOOL,
+            .{
+                .library_name = "kernel32",
+                .name = "SetEndOfFile",
+            },
+        );
```
matklad, line 1405: I am curious here, what's the context of these being removed from std? As in, should we perhaps be using something else?

src/io/windows.zig
```diff
@@ -1394,8 +1394,19 @@ pub const IO = struct {
             };
         }
 
+        // Declaring the function with an alternative name because `CamelCase` functions are
+        // by convention, used for building generic types.
```
matklad: TBH, I'd remove the comment --- it is not particularly relevant to what we are doing here so it distracts the reader. If anyone tries to rename it to be more idiomatic, they'll learn why it is the way it is :)

matklad: But that would be great tidbit to include in the commit message: https://mislav.net/2014/02/hidden-documentation/

src/vsr/journal.zig
```diff
@@ -1525,7 +1525,8 @@ pub fn JournalType(comptime Replica: type, comptime Storage: type) type {
                     assert(prepare.?.operation == .reserved);
                     assert(header.?.checksum == prepare.?.checksum);
                     assert(
-                        header.?.checksum == Header.Prepare.reserved_operation(cluster, slot.index).checksum,
+                        header.?.checksum ==
+                            Header.Prepare.reserved_operation(cluster, slot.index).checksum,
```
matklad: Oh, maybe `Header.Prepare.reserve(cluster, slot.index)`? As a verb?

src/io/windows.zig
batiati: This:
https://github.com/ziglang/zig/commit/3095e83d1162081ba54ed73c59810bfe0fb3f025

batiati: > should we perhaps be using something else?

I've added a `TODO(zig)` for this one (planning to take a look after the migration)
https://github.com/tigerbeetle/tigerbeetle/pull/2814/commits/f240d7dcdc3622b09de6bdbd87f914b8b876d83b

src/vsr/journal.zig
batiati: `reserve` as a verb sounds more like an acquire/release mechanism.

maybe `as_reserved()`, `from_reserved()`, or the other way around, renaming the field `reserved` to something else like `reserved_bytes` or `padding`.

matklad (approved):

PR #2819, opened 2025-03-13
Remove anonymous structs

batiati:
In preparation for Zig 0.14.0
Ref https://ziglang.org/download/0.14.0/release-notes.html#Remove-Anonymous-Struct-Types-Unify-Tuples

### DIFF
--- a/src/clients/java/src/client.zig
+++ b/src/clients/java/src/client.zig
@@ -30,7 +30,7 @@ const jni_version = jni.jni_version_10;
 
 const global_allocator = std.heap.c_allocator;
 
-pub const std_options = .{
+pub const std_options: std.Options = .{
     .log_level = .debug,
     .logFn = tb.exports.Logging.application_logger,
 };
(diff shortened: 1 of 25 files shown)


src/clients/java/src/client.zig
```diff
@@ -30,7 +30,7 @@ const jni_version = jni.jni_version_10;
 
 const global_allocator = std.heap.c_allocator;
 
-pub const std_options = .{
+pub const std_options = std.Options{
```
matklad:
```suggestion
pub const std_options: std.Options = .{
```

I think that is more idiomatic, and I _think_ I've heard rumors of `Foo{` syntax going away. But let me double check _why_ we need this change, I haven't caught this in relnotes.

matklad: Ok, I see, we weren't assigning _any_ type before!

matklad (commented):
Yeah, let's move types to the left of `=` everywhere when it's _not_ cumbersome

matklad (approved):

PR #2820, opened 2025-03-13
vopr: add performance mode

matklad:
Nothing special here, just reduce most sources of randomness to get a
deterministic result.

This is not yet integrated with CFO --- there's some weirdness with client requests such that some requests are dropped even in a perfect cluster. I want to fix that before adding this to CI.

builds on top of #2817

### DIFF
--- a/src/vopr.zig
+++ b/src/vopr.zig
@@ -64,10 +64,12 @@ pub const std_options: std.Options = .{
 pub const tigerbeetle_config = @import("config.zig").configs.test_min;
 
 const cluster_id = 0;
+const MiB = 1024 * 1024;
 
 const CLIArgs = struct {
     // "lite" mode runs a small cluster and only looks for crashes.
     lite: bool = false,
+    performance: bool = false,
     ticks_max_requests: u32 = 40_000_000,
     ticks_max_convergence: u32 = 10_000_000,
     positional: struct {
@@ -86,6 +88,9 @@ pub fn main() !void {
     defer args.deinit();
 
     const cli_args = flags.parse(&args, CLIArgs);
+    if (cli_args.lite and cli_args.performance) {
+        return vsr.fatal(.cli, "--lite and --performance are mutially exclusive", .{});
+    }
 
     const seed_random = std.crypto.random.int(u64);
     const seed = seed_from_arg: {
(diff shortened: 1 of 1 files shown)


src/vopr.zig
```diff
@@ -237,50 +247,59 @@ pub fn main() !void {
...
-    tick = 0;
-    while (tick < cli_args.ticks_max_convergence) : (tick += 1) {
-        simulator.tick();
-        tick_total += 1;
-        if (simulator.pending() == null) {
-            break;
+    if (cli_args.lite) {
+        // Don't care about convergence
+    } else {
+        simulator.transition_to_liveness_mode(.{
```
chaitanyabhandari: Mhmmmm I was thinking, since `options_performance` disables all faults, isn't safety mode equivalent to having a fully connected core without any perturbations? In other words, do we need liveness mode for performance testing? 🤔

matklad: Ah, excellent point! So the liveness mode does two things:

* It switches the fault mode (most failures are disabled, but the remaining failures that are left become permanent).
* It also checks convergence. In the loop above, we only have checked that the cluster responded to each request, we haven't checked that for all replicas commit_min = commit_max. 

So my original thinking that we still want liveness in performance just for convergence, but of course we don't need it! The latency of requests is the only thing we care about (in the future, we might want to track "latency to convergence", but that needs to be disaggregated from the request latency). 

Rebased&simplified!

matklad:
Did a couple more tweaks here after experimenting with #2821:

* reduced latencies of storage to zero (c.f. #2823)
* reduced the number of clients to 6, to be bellow client_max (7 for vopr)
* disabled the standard vopr logs, as they get in the way

src/vopr.zig
```diff
@@ -427,6 +444,109 @@ fn options_lite(prng: *stdx.PRNG) Simulator.Options {
...
+        .in_flight_max = ReplySequence.stalled_queue_capacity,
+    });
+
+    return .{
+        .cluster = cluster_options,
+        .network = network_options,
+        .storage = storage_options,
+        .storage_fault_atlas = storage_fault_atlas,
+        .workload = workload_options,
+        // TODO Swarm testing: Test long+few crashes and short+many crashes separately.
```
chaitanyabhandari: This is just a TODO for `options_swarm`, no? Or does it also apply here since we're planning on doing performance mode with faults?

chaitanyabhandari (dismissed):
LGTM! Don't want to block the PR based on a nit (https://github.com/tigerbeetle/tigerbeetle/pull/2820#discussion_r1998791614). :)

src/vopr.zig
matklad: Good catch, removed!

chaitanyabhandari (approved):

PR #2823, opened 2025-03-14
vopr: allow fast writes

matklad:
In the current VOPR config, the minimum mean storage latency is 3 ticks,
which is 30 ms, which is just ridiculous. We generally expect almost all
IOPs to complete faster than in a single tick.

To model "faster than a tick" realistically, I change the simulator to
drain storage&network IO "within" each tick. As there could be multiple
immediately ready IOPs, the simulator randomizes the order in which
completions are invoked.

### DIFF
--- a/src/testing/cluster.zig
+++ b/src/testing/cluster.zig
@@ -446,6 +446,29 @@ pub fn ClusterType(comptime StateMachineType: anytype) type {
         }
 
         pub fn tick(cluster: *Cluster) void {
+            // Interleave storage and network steps, to allow for faster-than-a-tick IO.
+            while (true) {
+                var advanced = false;
+                advanced = cluster.network.step() or advanced;
+
+                for (
+                    cluster.storages,
+                    cluster.replica_health,
+                    cluster.replica_upgrades,
+                    0..,
+                ) |*storage, *health, *upgrade, i| {
+                    if (health.* == .up and health.*.up.paused) continue;
+                    // Upgrades immediately follow storage.step(), since upgrades occur at
+                    // checkpoint completion. (Downgrades are triggered separately – see
+                    // replica_restart()).
+                    advanced = storage.step() or advanced;
+                    if (upgrade.*) |_| cluster.replica_release_execute(@intCast(i));
+                    assert(upgrade.* == null);
+                }
+
+                if (!advanced) break;
+            }
+
             cluster.network.tick();
 
             for (cluster.clients, cluster.client_eviction_reasons) |*client, eviction_reason| {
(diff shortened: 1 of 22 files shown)


src/testing/cluster.zig
```diff
@@ -446,6 +446,29 @@ pub fn ClusterType(comptime StateMachineType: anytype) type {
         }
 
         pub fn tick(cluster: *Cluster) void {
+            // Interleave storage and network progres, to allow for faster-than-a-tick IO.
```
sentientwaffle:
```suggestion
            // Interleave storage and network progress, to allow for faster-than-a-tick IO.
```

src/testing/storage.zig
```diff
@@ -335,25 +335,60 @@ pub const Storage = struct {
         }
     }
 
-    pub fn tick(storage: *Storage) void {
-        storage.ticks += 1;
+    pub fn progress(storage: *Storage) bool {
```
sentientwaffle: Does this name use "progress" as a noun or a verb? i.e. "Is there progress?" vs "Go make progress.". It sort of seems like both at the same time right now.

src/testing/cluster.zig
sentientwaffle, line 1: There is a CFO seed, which can be fixed by:

```diff
@@ -843,6 +859,11 @@ pub fn ClusterType(comptime StateMachineType: anytype) type {
                 },
                 .sync_stage_changed => switch (replica.syncing) {
                     .idle => cluster.log_replica(.sync, replica.replica),
+                    .updating_checkpoint => {
+                        cluster.state_checker.check_state(replica.replica) catch |err| {
+                            fatal(.correctness, "state checker error: {}", .{err});
+                        };
+                    },
                     else => {},
                 },
                 .client_evicted => |client_id| cluster.cluster_on_eviction(client_id),
```

I think this PR causes state sync to finish fast enough in some cases that the state checker wasn't ever registering it, so it failed to capture the op jump here:

```zig
        pub fn check_state(state_checker: *StateChecker, replica_index: u8) !void {
            const replica = &state_checker.replicas[replica_index];
            if (replica.syncing == .updating_checkpoint) {
                // Allow a syncing replica to fast-forward its commit.
                //
                // But "fast-forwarding" may actually move commit_min slightly backwards:
                // 1. Suppose op X is a checkpoint trigger.
                // 2. We are committing op X-1 but are stuck due to a block that does not exist in
                //    the cluster anymore.
                // 3. When we sync, `commit_min` "backtracks", to `X - lsm_compaction_ops`.
                const commit_min_source = state_checker.commit_mins[replica_index];
                const commit_min_target =
                    replica.syncing.updating_checkpoint.header.op;
                assert(commit_min_source <= commit_min_target + constants.lsm_compaction_ops);
                state_checker.commit_mins[replica_index] = commit_min_target;
                return;
            }
```

src/testing/storage.zig
matklad: A verb, progrEss  (I think both your example are nouns though). 

It is ambiguous, I’ll think about a better name, and I am open for suggestions.

sentientwaffle: Ah yeah, both of my examples were nouns. Maybe `forward`? Though in the context of the network that might be confusing still. Maybe "step", though that is pretty close to "tick". "advance"?

matklad: Ah, step would be perfect, as this _is_ a counter-part of `run` from the next commit!

matklad: Yeah, I love this:

```zig
    pub fn run(storage: *Storage) void {
        while (storage.step()) {}
        storage.tick();
    }
```

sentientwaffle (dismissed):

matklad:
A new seed, haven't looked into it yet: `./zig/zig build -Drelease vopr -- 1559163846617790652`

matklad:
Green at 338 273!

src/testing/storage.zig
jorangreef: "advance" was also just what I was thinking—and then @sentientwaffle voiced that, and "step" looks even better!

sentientwaffle (approved):

PR #2824, opened 2025-03-16
Bump `release_client_min` to 0.16.4

batiati:
In preparation for #2787, remove backward compatibility with deprecated features:

- `register` without `RegisterRequest` (clients <v0.15.4) PR #1981 released on July 29, 2024.
- `forbid_zero_amounts` (clients <0.16.0) PR #2220 released on September 2, 2024.
- `AccountFilterFormer.required` (clients <0.16.2) PR #2300 released on September 16, 2024.
- `retry_transient_error` (<0.16.4) PR #2335 released on September 30, 2024.


src/state_machine.zig
```diff
@@ -854,33 +854,34 @@ pub fn StateMachineType(
 
         pub fn input_valid(
             self: *const StateMachine,
-            client_release: vsr.Release,
             operation: Operation,
             input: []align(16) const u8,
         ) bool {
+            // NB: This function should never accept `client_release` as an argument.
+            // Any public API changes must be introduced explicitly as a new `operation` number.
```
matklad, line 861: 🎉 ❤️

src/state_machine.zig
```diff
@@ -1816,11 +1770,15 @@ pub fn StateMachineType(
...
+            // NB: The `client_release` parameter may only be used for backwards compatibility
+            // with old clients' behavior.
+            // Any public API changes must be introduced explicitly as a new `operation` number.
+            _ = client_release;
             assert(op != 0);
-            assert(self.input_valid(client_release, operation, input));
+            assert(self.input_valid(operation, input));
             assert(timestamp > self.commit_timestamp or global_constants.aof_recovery);
             assert(input.len <= self.batch_size_limit);
+            maybe(client == 0); // Can be zero for `pulse`.
```
matklad: maybe ` if (client == 0) assert(operation == .pulse)` ?

src/state_machine.zig
```diff
@@ -1816,11 +1770,15 @@ pub fn StateMachineType(
             input: []align(16) const u8,
             output: *align(16) [constants.message_body_size_max]u8,
         ) usize {
-            _ = client;
+            // NB: The `client_release` parameter may only be used for backwards compatibility
+            // with old clients' behavior.
+            // Any public API changes must be introduced explicitly as a new `operation` number.
+            _ = client_release;
```
matklad: Hm, wasn't our plan to completely remove `client_release` from state machine's API?

src/state_machine.zig
```diff
@@ -2713,7 +2624,7 @@ pub fn StateMachineType(
             assert(cr_account.id == p.credit_account_id);
             assert(p.timestamp > dr_account.timestamp);
             assert(p.timestamp > cr_account.timestamp);
-            if (forbid_zero_amounts(client_release)) assert(p.amount > 0);
+            maybe(p.amount > 0);
```
matklad: `maybe(p.amount == 0);`, to callout the more surprising case. 

Also, I believe the line removed is a bug! Pending and posting/voiding transfer might be created by different client versions! 

So, a new client might create a pending transfer with zero amount, and then the old client might try to post it!

src/state_machine.zig
```diff
@@ -2729,17 +2640,13 @@ pub fn StateMachineType(
...
             const amount = amount: {
-                if (forbid_zero_amounts(client_release)) {
+                if (t.flags.void_pending_transfer) {
                     break :amount if (t.amount > 0) t.amount else p.amount;
                 } else {
-                    if (t.flags.void_pending_transfer) {
-                        break :amount if (t.amount > 0) t.amount else p.amount;
-                    } else {
-                        break :amount if (t.amount == std.math.maxInt(u128)) p.amount else t.amount;
-                    }
```
matklad, line 2739: Unrelated, but let's make this more symmetric:

```
            const amount = amount: {
                if (t.flags.void_pending_transfer) {
                    break :amount if (t.amount == 0) p.amount else t.amount;
                } else {
                    break :amount if (t.amount == std.math.maxInt(u128)) p.amount else t.amount;
                }
            };
```

src/state_machine.zig
```diff
@@ -2850,9 +2757,6 @@ pub fn StateMachineType(
             if (t2.flags.post_pending_transfer) {
```
matklad, line 2751: In the next `if (t2.flags.void_pending_transfer) {`, could you add `assert(t2.amount == p.amount)`?

matklad (commented):
First round in, but I'll need another look here

src/state_machine.zig
batiati: We're preserving the `client_release` infrastructure (e.g., the state machine accepting it for prefetch/commit and unit tests supporting multiple client versions) as it may still be used for minor changes in behavior, not in the API, such as when we allowed zeroed amounts.  

But I don't have a strong opinion here, we could instead enforce strict versioning by always introducing a new operation for any kind of change.

src/clients/java/java_bindings.zig
```diff
@@ -236,6 +236,7 @@ fn emit_enum(
     try buffer.writer().print(
         \\{[notice]s}
         \\package com.tigerbeetle;
+        \\import java.util.HashMap;
```
matklad: Should we use https://docs.oracle.com/javase/8/docs/api/java/util/EnumMap.html here?

src/state_machine.zig
batiati: After some thinking, it's better to remove `client_release` entirely from the state machine API, and let's try hard to avoid adding it back in the future (even for minor changes).

src/clients/java/java_bindings.zig
matklad: Ah, no that goes in the reverse direction. Yeah, let's leave it as a hash map, though, I think the most Occam's minimal solution here would be to generate a switch statement?

src/clients/node/src/bindings.ts
```diff
@@ -766,6 +761,11 @@ export enum CreateTransferError {
   * See [exceeds_debits](https://docs.tigerbeetle.com/reference/requests/create_transfers#exceeds_debits)
   */
   exceeds_debits = 55,
+
+  /**
+  * See [deprecated_18](https://docs.tigerbeetle.com/reference/requests/create_transfers#deprecated_18)
+  */
+  deprecated_18 = 18,
```
matklad: Hm, why node is different? For all other clients, we completely hide deprecations from the API, but we keep it here?

src/vsr/replica_test.zig
```diff
@@ -1496,45 +1496,6 @@ test "Cluster: scrub: background scrubber, fully corrupt grid" {
     try TestReplicas.expect_equal_grid(b1, b2);
 }
 
-// Compat(v0.15.3)
-test "Cluster: client: empty command=request operation=register body" {
```
matklad, line 1500: In replica.zig, you can now remove the code that's being tested by the test:

```
            if (prepare.header.size == @sizeOf(vsr.Header)) {
                // Old clients which don't send a RegisterRequest also don't check
                // `batch_size_limit`.
                result.* = .{
                    .batch_size_limit = 0,
                };
            } else {
```

```
            if (request.header.size == @sizeOf(vsr.Header)) {
                // Old clients don't send a RegisterRequest.
            } else {
```

src/vsr/message_header.zig
```diff
@@ -553,9 +553,7 @@ pub const Header = extern struct {
                     if (self.parent != 0) return "register: parent != 0";
                     if (self.session != 0) return "register: session != 0";
                     if (self.request != 0) return "register: request != 0";
-                    if (self.size != @sizeOf(Header) and // Compat(v0.15.3)
-                        self.size != @sizeOf(Header) + @sizeOf(vsr.RegisterRequest))
-                    {
+                    if (self.size != @sizeOf(Header) + @sizeOf(vsr.RegisterRequest)) {
```
matklad: Huh, this is a puzzle! If we do check this here, then we won't send an eviction message to the client, and instead silently drop the message. We do `.invalid` check before we do client_version_min check.

What are we to do? 🤔 

I suggest the following:

* Keep this code as is for at least a couple of releases more, such that, if people don't get our deprecation memo, they get a clean error message, rather than a silent no-answer
* But remove it afterwards: I think there's a lot of value in not admitting invalid values into the system, and, once almost everyone migrates, we shouldn't loose much from omitting the eviction message.

My hypothesis is that we won't be doing many similar "frame changes" in the future. If that turns out to be wrong, we need to change on message to move client version check before header validity check, but I think we won't need it, or at least won't need it yet!

src/tidy.zig
```diff
@@ -367,7 +367,7 @@ fn tidy_dead_declarations(
 }
 
 /// As we trim our functions, make sure to update this constant; tidy will error if you do not.
-const function_line_count_max = 459; // fn check in state_machine.zig
+const function_line_count_max = 413; // fn check in state_machine.zig
```
matklad, line 370: Nice win! We should get back to our function length at some point, but not today!

src/vsr/replica_test.zig
```diff
@@ -1496,45 +1496,6 @@ test "Cluster: scrub: background scrubber, fully corrupt grid" {
...
-        body: vsr.RegisterResult,
-    };
-
-    const reply = std.mem.bytesAsValue(Reply, client_bus.reply.?.buffer[0..@sizeOf(Reply)]);
-    try expectEqual(reply.header.command, .reply);
-    try expectEqual(reply.header.operation, .register);
-    try expectEqual(reply.header.size, @sizeOf(Reply));
-    try expectEqual(reply.header.request, 0);
-    try expect(stdx.zeroed(std.mem.asBytes(&reply.body)));
-}
```
matklad: Ah, maybe not remove this test altogether, but rather change the expectation? That is , assert that the client is evicted. 


That's my general rule of thumb for tests that stride versions: instead of testing just the old behavior, or just the new behavior, write the test as

```zig
for (.{.old, .new}) |behavior| {
    // normal test body //
}
```

and inside switch on `.behavior` to specifically underscore the _diff_ in the response!  

The spirit of this rule would've caught the "invalid before eviction" effect without us having to discover it by hand.

matklad:
As this is a non-trivial compatibility change, let's have a second pair of eyes here as well as per our review policy!

batiati:
> As this is a non-trivial compatibility change, let's have a second pair of eyes here as well as per our review policy!

summoning @sentientwaffle :tada:

matklad (dismissed):

src/vsr/replica_test.zig
```diff
@@ -1514,25 +1514,22 @@ test "Cluster: client: empty command=request operation=register body" {
         .request = 0,
         .command = .request,
         .operation = .register,
-        .release = releases[0].release,
+        .release = vsr.Release.from(.{ .major = 0, .minor = 15, .patch = 3 }),
```
```suggestion
        .release = vsr.Release.from(.{ .major = 0, .minor = 0, .patch = 9 }),
```

With this change we will fail with `client_release_too_low`,  which reflects how we actually expect it to work.

sentientwaffle: Also, if you manually change this release to be the the same as the cluster's release, then the cluster panics which seems wrong.

sentientwaffle (approved):
LGTM! This isn't technically a breaking API change, but should we mention it on https://github.com/tigerbeetle/tigerbeetle/issues/2231 anyway?

PR #2826, opened 2025-03-18
zig: simplify the idiom for creating comptime lists

matklad:
`++` is awkward to use when you want to add only a single element. But if you know the length up-front, you don't need `++`, you can assign array elements directly!


batiati (commented):
LGTM, just need to rebase it first.

batiati (approved):

PR #2827, opened 2025-03-19
docs: sync navigation when using the close button and some style fixes

fabioarnold:
* Synchronizes the sidebar navigation with the current page when the search is closed via its close button
* `<strong>` links were styled black instead of using the link color
* Add a style for `<blockquote>`:
image


cb22 (approved):
LGTM - but CI is currently blocked due to binary sized.

PR #2829, opened 2025-03-19
vsr: remove needless indirection

matklad:
Can use Grid.RepairTable directy here, we don't use *Replica pointer.

### DIFF
--- a/src/vsr/grid_blocks_missing.zig
+++ b/src/vsr/grid_blocks_missing.zig
@@ -20,8 +20,6 @@ const FIFOType = @import("../fifo.zig").FIFOType;
 const BlockPtrConst = *align(constants.sector_size) const [constants.block_size]u8;
 
 pub const GridBlocksMissing = struct {
-    const TableDataBlocksSet = std.StaticBitSet(constants.lsm_table_data_blocks_max);
-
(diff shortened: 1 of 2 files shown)


src/vsr/grid_blocks_missing.zig
```diff
@@ -212,12 +213,15 @@ pub const GridBlocksMissing = struct {
...
         checksum: u128,
     ) enum { insert, duplicate } {
         assert(queue.faulty_tables.count < queue.options.tables_max);
         assert(queue.faulty_blocks.count() ==
             queue.enqueued_blocks_single + queue.enqueued_blocks_table);
+        assert(table_bitset.capacity() == constants.lsm_table_data_blocks_max);
+        assert(table_bitset.count() == 0);
 
         var tables = queue.faulty_tables.peek();
         while (tables) |queue_table| : (tables = queue_table.next) {
```
sentientwaffle, line 227: In this loop we could also assert that `queue_table.data_box_received != table_bitset`.

sentientwaffle (dismissed):
One nit; otherwise this looks good to me.

sentientwaffle (approved):

PR #2832, opened 2025-03-20
CFO: Retain checkouts + cache across iterations

sentientwaffle:
Right now we force (each) CFO to recompile all fuzzers every iteration (so every ~hour) by deleting the working tree (which includes zig's binary cache).

This leads to a lot of extra compiling -- for the current PR's it is about 10 minutes for every 60 minutes of actual fuzzing. Since the branches don't change that often, this is unnecessary work -- we'd rather spend those 10 minutes fuzzing instead.

(When we are compiling, we aren't starting fuzzers, and compilation is only single-threaded, so the CFO machine spends that ~10 minutes very underutilized.)

Instead let's retain the cache and just garbage-collect it selectively when tasks are removed.

Note that this change requires updating the supervisor script.

### DIFF
--- a/src/scripts/cfo.zig
+++ b/src/scripts/cfo.zig
@@ -236,7 +236,6 @@ fn run_fuzzers(
 
         if (iteration_pull) {
             try run_fuzzers_prepare_tasks(&tasks, shell, gh_token);
-            try run_fuzzers_cleanup(&tasks, shell, gh_token);
 
             for (tasks.list.items) |*task| {
                 if (task.generation == tasks.generation) {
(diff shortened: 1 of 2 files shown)


matklad:
I am _slightly_ apprehensive here: stateless is good, stateful is hard, and I personally prefer avoiding dealing with the state and rather let some poor database developer get it right :P 

10 minutes compiling feels somewhat excessive to me, I wonder if there's some quick way to not do this many compiling? (or maybe to more effectively use build system to compile stuff in parallel). 

No super strong opinion here, caching probably _is_ the lesser of two weevils !

src/scripts/cfo.zig
```diff
@@ -683,6 +681,34 @@ fn run_fuzzers_prepare_tasks(tasks: *Tasks, shell: *Shell, gh_token: ?[]const u8
...
 
+/// Garbage-collect checkouts which don't have any running tasks.
+/// Otherwise we reuse checkouts as long as possible, to minimize time spent recompiling fuzzers.
+fn run_fuzzers_cleanup(tasks: *Tasks, shell: *Shell, gh_token: ?[]const u8) !void {
+    if (gh_token) |_| {
+        const working_directory = try shell.cwd.openDir("./working", .{ .iterate = true });
+        var checkouts = working_directory.iterate();
+        while (try checkouts.next()) |checkout| {
+            if (checkout.kind == .directory) {
+                if (std.mem.eql(u8, checkout.name, "main")) continue;
```
matklad: Can we collect `main`at least every 100th iteration or something? I _think_ `.zig-cache` will grow infinitely over wise, I don't think Zig implements GC of incremental compilation state?

sentientwaffle: That's a great idea! The supervisor script can just randomly delete the entire working directory periodically. That is much simpler than my approach, and it solves the state issue also.

src/scripts/cfo_supervisor.sh
```diff
@@ -10,10 +10,19 @@ git --version
...
+
     (
-        git clone https://github.com/tigerbeetle/tigerbeetle tigerbeetle
+        if ! [ -d ./tigerbeetle ]
+        then git clone https://github.com/tigerbeetle/tigerbeetle tigerbeetle
+        fi
+
         cd tigerbeetle
+        git fetch
+        git checkout -f origin/main
```
matklad: GitHub ate my comment :(

Let's use `git switch --discard-changes`:

- as a rule of thumb, scripts should use full flags, like `--force` instead of `-f`, to make sure the reader doesn't have second guess whether it is force, file, follow, or something else
- `git switch` is a more safe replacement for `git checkout`. In checkout, the argument can be _either_ a file path or a comittish, and git decides heuristically which is which. `switch` only operates on comitishes.

matklad (approved):

PR #2833, opened 2025-03-20
stdx: vendor bitset

matklad:
Vendor another type from stdx to std! The standard motivation applies:
we want to insualte us from std API chunrn, reduce the risk of
"meaningless" (to us) changes subtly breaking the sematics we rely on,
and align APIs with TIGER_STYLE.

The extra specific motivation in this case is that stdlib type provides
both `initFull` and `initEmpty` symmetrically, but `initFull` is a
binary size footgun --- you must include the entire bitset into the text
section of Elf binary! The `initEmpty` works by `memset`, rather than
`memcpy`, at runtime. And, although symmetrict, std only provides
`findFirstSet`, not `findFirstUnset`, which forces us to use initFull at
times.

This change reducess the release binary size on Linux by 38KiB.

Additionally, there are a couple of "nice to have" changes:

- std version comptime dispatches between a small integer-backed bitset,
  and a large array-backed one. We instead _assert_ that the bitset is
  reasonably small (if it isn't, you should use DynamicBitset, to avoid
  specializing on length).
- I decided to avoid odd-sized integers (eg, `u3`) as a backign integer
  for the bitset, to make sure we rely on simpler compiler machinery.
- It feels like we sometimes used a "flipped" bitset just to gain
  `findFirstSet`, now we rely on the natural empty semantics more often.
- I removed some extra API surface, like bidirectional iteration over
  set/unset. We _might_ end up adding it back later, but my hunch is
  that a minimal API would be enough.

### DIFF
--- a/src/iops.zig
+++ b/src/iops.zig
@@ -1,53 +1,56 @@
 const std = @import("std");
+const stdx = @import("./stdx.zig");
 const assert = std.debug.assert;
 
 /// Take a u8 to limit to 256 items max (2^8 = 256)
 pub fn IOPSType(comptime T: type, comptime size: u8) type {
-    const Map = std.StaticBitSet(size);
+    const Map = stdx.BitSetType(size);
+
     return struct {
         const IOPS = @This();
 
         items: [size]T = undefined,
-        /// 1 bits are free items.
-        free: Map = Map.initFull(),
+        busy: Map = .{},
 
         pub fn acquire(self: *IOPS) ?*T {
-            const i = self.free.findFirstSet() orelse return null;
-            self.free.unset(i);
+            const i = self.busy.first_unset() orelse return null;
+            self.busy.set(i);
             return &self.items[i];
         }
 
         pub fn release(self: *IOPS, item: *T) void {
             item.* = undefined;
             const i = self.index(item);
-            assert(!self.free.isSet(i));
-            self.free.set(i);
+            assert(self.busy.is_set(i));
+            self.busy.unset(i);
         }
 
         pub fn index(self: *IOPS, item: *T) usize {
-            const i = @divExact((@intFromPtr(item) - @intFromPtr(&self.items)), @sizeOf(T));
+            const i = @divExact(
+                (@intFromPtr(item) - @intFromPtr(&self.items)),
+                @sizeOf(T),
+            );
             assert(i < size);
             return i;
         }
 
         /// Returns the count of IOPs available.
         pub fn available(self: *const IOPS) usize {
-            return self.free.count();
+            return self.busy.capacity() - self.busy.count();
         }
 
-        pub fn total(self: *const IOPS) usize {
-            _ = self;
+        pub inline fn total(_: *const IOPS) usize {
             return size;
         }
 
         /// Returns the count of IOPs in use.
         pub fn executing(self: *const IOPS) usize {
-            return self.total() - self.available();
+            return self.busy.count();
         }
 
         pub const Iterator = struct {
             iops: *IOPS,
-            bitset_iterator: Map.Iterator(.{ .kind = .unset }),
+            bitset_iterator: Map.Iterator,
 
             pub fn next(iterator: *@This()) ?*T {
                 const i = iterator.bitset_iterator.next() orelse return null;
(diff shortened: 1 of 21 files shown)


src/stdx/bit_set.zig
```diff
@@ -0,0 +1,156 @@
...
+                        const bit = prng.int_inclusive(usize, N - 1);
+                        set.set(bit);
+                        model.set(bit);
+                    }
+                },
+                .unset => {
+                    if (N > 0) {
+                        const bit = prng.int_inclusive(usize, N - 1);
+                        set.set(bit);
+                        model.set(bit);
```
batiati:
```suggestion
                        set.unset(bit);
                        model.unset(bit);
```

src/stdx/bit_set.zig
```diff
@@ -0,0 +1,156 @@
+const stdx = @import("../stdx.zig");
+const std = @import("std");
+const assert = std.debug.assert;
+
+/// Take a u8 to limit to 255 items max (2^8 - 1 = 255).
+/// Use dynamic bitset for larger sizes.
+pub fn BitSetType(comptime with_capacity: u8) type {
+    return struct {
+        // While mathematicaly 0 and 1 are symmetric, we intentionally bias the API to use zeros
+        // default, as zero-initalization requces binary size.
```
batiati:
```suggestion
        // default, as zero-initialization reduces binary size.
```

src/iops.zig
```diff
@@ -1,53 +1,56 @@
...
+            assert(self.busy.is_set(i));
+            self.busy.unset(i);
         }
 
         pub fn index(self: *IOPS, item: *T) usize {
-            const i = @divExact((@intFromPtr(item) - @intFromPtr(&self.items)), @sizeOf(T));
+            const i = @divExact(
+                (@intFromPtr(item) - @intFromPtr(&self.items)),
+                @sizeOf(T),
+            );
```
batiati, line 32: Unnecessary line break, still under 100 columns.

src/iops.zig
```diff
@@ -1,53 +1,56 @@
...
-    const Map = std.StaticBitSet(size);
+    const Map = stdx.BitSetType(size);
+
     return struct {
         const IOPS = @This();
 
         items: [size]T = undefined,
-        /// 1 bits are free items.
-        free: Map = Map.initFull(),
+        busy: Map = .{},
```
batiati, line 13: :tada:  It's much nicer to have it inverted from `free/unset` to `busy/set` !

src/vsr/client_sessions.zig
```diff
@@ -174,16 +174,17 @@ pub const ClientSessions = struct {
...
                     .header = header.*,
                 };
             }
         }
 
-        assert(constants.clients_max - client_sessions.entries_free.count() ==
-            client_sessions.entries_by_client.count());
+        assert(
+            client_sessions.entries_present.count() == client_sessions.entries_by_client.count(),
+        );
```
batiati, line 187: :100:

src/vsr/journal.zig
```diff
@@ -240,10 +240,10 @@ pub fn JournalType(comptime Replica: type, comptime Storage: type) type {
             constants.journal_iops_write_max
         ][constants.sector_size]u8,
 
-        /// A set bit indicates a chunk of redundant headers that no read has been issued to yet.
-        header_chunks_requested: HeaderChunks = HeaderChunks.initFull(),
+        /// A set bit indicates a chunk of redundant headers that have a read has been issued.
```
batiati:
```suggestion
        /// A set bit indicates a chunk of redundant headers for which a read has been issued.
```

src/iops.zig
matklad: Yeah, I just find it poorly readable without a line break: it is a fraction, so we might as well make it look like a fraction!

src/stdx/bit_set.zig
matklad: Oops, great catch!

batiati (approved):

matklad:
Will merge after release on Monday, I want to get https://github.com/tigerbeetle/tigerbeetle/pull/2834 in first to make sure VOPR is green again.

With me switching polarity in couple of places, I _might_ have messed up set/unset somewhere, and want to get this somewhat more testing

batiati:
> Will merge after release on Monday, I want to get #2834 in first to make sure VOPR is green again.
> 
> With me switching polarity in couple of places, I _might_ have messed up set/unset somewhere, and want to get this somewhat more testing

Yeah, lets do it. I have double-checked all places where `.initFull()` was switched to `.{}`, but I also feel like it needs more testing.

I also rebased it to Zig 0.14.0 to get a preview if we would need much refactoring, and it almost compiles without changes.

matklad:
@batiati given that you are online, want to ✅ #2834? ^^

PR #2834, opened 2025-03-20
vopr: fix false positive

matklad:
Fixes a vopr seed causing a  crash during upgrade. I belive this is a pre-exising VOPR bug, which was uncovered by the tickless IO from #2823.

In production, release_execute is a `noreturn` function that immediately re-execs into the newer version.

In VOPR, `release_execute` just sets the flag. VOPR acts on the flag when it gets controll. Before #2823 storage.tick, and, after, storage.step executed IOPs and next_tick_queue callbacks after each other. So it was a possibility that an IOP triggers release_exectue, and a next_tick steps into it, but it wasn't hit before.

To fix it, ensure that storage.step "yields" between IOPs and next ticks. This is still perhaps not ideally, as we _should_ be yielding after every _individual_ IOP, but lets see if this is enough to get the thing fixed. I _think_ we should have a `next_tick` somewhere in commit loop to make sure that the "yield" does happen.

Seed: ./zig/zig build vopr -Dvopr-state-machine=testing -- 18445183419870469729

### DIFF
--- a/src/testing/storage.zig
+++ b/src/testing/storage.zig
@@ -358,28 +358,35 @@ pub const Storage = struct {
             write_count += 1;
         }
 
-        if (read_count == 0 and write_count == 0 and storage.next_tick_queue.empty()) {
-            return false;
-        }
+        // NB: to support upgrades in the simulator, we do not process next_tick_queue if we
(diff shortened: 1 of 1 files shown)


matklad:
Green at 1 143 860

batiati (approved):

PR #2837, opened 2025-03-21
vopr: properly fix upgrades during tickless IO

matklad:
In #2834, I fixed vopr failure by introducing a "yield" into the step function, between IOPs and next_tick, but that isn't really the right way to do that.

Instead, we should set a flag to notify step that reset happened, so that it can properly wind down.

This is how it used to work _before_ step, when `releae_execute_soon` would simply `.reset` the storage. This worker for tick, which simply procesesed events from the ready queue. But now that the step copies ready events to a local variable, we need some more active reset process.

Thanks chaitanyabhandari for pointing out the flaw in the original approach!

### DIFF
--- a/src/testing/storage.zig
+++ b/src/testing/storage.zig
@@ -146,6 +146,8 @@ pub const Storage = struct {
     options: Options,
     prng: stdx.PRNG,
 
+    status: enum { idle, stepping, resetting },
+
     /// `memory` always contains the pristine data as-written -- it does not include storage faults.
     memory: []align(constants.sector_size) u8,
     /// Set bits correspond to sectors that have ever been written to.
@@ -229,6 +231,7 @@ pub const Storage = struct {
             .size = size,
             .options = options,
             .prng = prng,
+            .status = .idle,
             .memory = memory,
             .memory_written = memory_written,
             .faults = faults,
@@ -239,6 +242,7 @@ pub const Storage = struct {
     }
 
     pub fn deinit(storage: *Storage, allocator: mem.Allocator) void {
+        assert(storage.status == .idle);
         storage.writes.deinit();
         storage.reads.deinit();
         allocator.destroy(storage.overlay_buffers);
@@ -250,6 +254,10 @@ pub const Storage = struct {
     /// Cancel any currently in-progress reads/writes.
     /// Corrupt the target sectors of any in-progress writes.
     pub fn reset(storage: *Storage) void {
+        assert(storage.status == .idle);
+        storage.status = .resetting;
+        defer storage.status = .idle;
+
         log.debug("Reset: {} pending reads, {} pending writes, {} pending next_ticks", .{
             storage.reads.count(),
             storage.writes.count(),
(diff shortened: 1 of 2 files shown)


src/testing/storage.zig
```diff
@@ -358,43 +387,63 @@ pub const Storage = struct {
...
+                .resetting => {},
             }
-            return true;
+        }
+
+        switch (storage.status) {
+            .idle => unreachable,
+            .stepping => storage.status = .idle,
+            .resetting => {
+                storage.status = .idle;
```
chaitanyabhandari, line 429: For my understanding, this is to take care of the case where the callbacks  that we invoke further enqueue some IO, right?

src/testing/storage.zig
```diff
@@ -842,6 +909,7 @@ pub const Storage = struct {
     }
 
     pub fn client_replies(storage: *const Storage) []const MessageRawType(.reply) {
```
chaitanyabhandari, line 911: _Seems_ like this would just be called storage is `idle`? I see invocations from:

* `test "format"` in `replica_format.zig`
* `core_missing_reply` in `vopr.zig`

Both of which aren't invoked from any callbacks, and so storage state can't be `stepping`?

src/testing/storage.zig
```diff
@@ -804,13 +868,15 @@ pub const Storage = struct {
         storage: *const Storage,
         copy_: u8,
     ) *const superblock.SuperBlockHeader {
+        assert(storage.status == .idle or storage.status == .stepping);
```
chaitanyabhandari: *Also* like this would just be called storage is idle? I see invocations from:

* test "format" in replica_format.zig

I guess a question to ask here is: are we trying to precise or future proof (such that these can be called from callbacks in the future)?

src/testing/storage.zig
```diff
@@ -784,6 +847,7 @@ pub const Storage = struct {
 
     /// Returns whether any sector in the area is corrupt.
     pub fn area_faulty(storage: *const Storage, area: Area) bool {
```
chaitanyabhandari, line 849: Also seems like it could be invoked only when status is `idle`? Invocations in:

* `core_missing_blocks` & `core_missing_reply` in `vopr.zig`
* A couple of tests in `replica.zig`

src/testing/storage.zig
matklad: Yeah, this branch is the "happy path" here: while we were stepping, someone asked us to reset, so we reset here. In reset, I want to assert that the state is idle, so I _make_ it idle here. This is sort-of `defer self.reset()`

src/testing/storage.zig
matklad: This is mostly for symmetry. Things like `wal_prepares` can be invoked when idle or when stepping, and I figured that I'd rather allow any getter to be called as long as we are not resetting. 

Ie, it wouldn't be _wrong_ to call `area_faulty` from stepping in the storage checker or something

chaitanyabhandari: Fair! So that answers my question in https://github.com/tigerbeetle/tigerbeetle/pull/2837/files#r2007906102, this is indeed to be future proof, so we _can_ call these in the future while stepping.

chaitanyabhandari: >  I'd rather allow any getter to be called as long as we are not resetting

Also, nit - what do you think about simplifying these to `assert(storage.status != .resetting)`?

matklad: Good call! I instinctively wrote an `or` because that's usually better with enums --- it makes sure that the code won't silently change semantics if we introduce a new variant. 

But here we _specifcally_ testing for "not resetting", rather than "being in a particular state", so `!= .resetting` _is_ clearer!

chaitanyabhandari (approved):
Elegant solution! LGTM.

PR #2840, opened 2025-03-21
VSR: Add replica/lsm/grid/journal metrics

sentientwaffle:
None of these have any tags (other than `replica`/`cluster`).


matklad (dismissed):
Do we want to add `grid_blocks_aquired`, or are we worrying about free-set popCount performance? 

Another interesting thin would be table count, but I think we already track that at per-tree granularity. 

And maybe `CommitStage`?

sentientwaffle:
Added `grid_blocks_acquired`!

The `replica_commit` timing event does include `CommitStage` as a tag. For the current commit stage though, these gauges are only emitted every 10 seconds -- at that resolution I don't think the _current_ `CommitStage` (at time of emit) is that useful.

matklad (approved):

cb22:
Ah, looks like the devhub-in-pr check caught this, but didn't yet gate merging!

PR #2841, opened 2025-03-22
vsr: fix crash in primary_repair_pipeline

matklad:
The last step of the repair for prospective primary is pipeline repair (loading prepares from WAL on disk into pipeline in memory). It is tricky because, when WAL read finishes, the replica might no longer be a primary!

So `repair_pipeline_read_callback` starts with a bunch of checks to ensure that we still should be repairing pipeline. It misses one case, when there's a header break betwene commit_max and op.

Seed: ./zig/zig build  vopr -Drelease -- 16932172197309204889

### DIFF
--- a/src/vsr/replica.zig
+++ b/src/vsr/replica.zig
@@ -7278,11 +7278,21 @@ pub fn ReplicaType(
                 return;
             }
 
+            if (self.journal.find_latest_headers_break_between(self.commit_max, self.op)) |range| {
+                log.debug("{}: repair_pipeline_read_callback: header break {}..{}", .{
+                    self.replica,
+                    range.op_min,
+                    range.op_max,
+                });
+                return;
+            }
+
             // We are in a state where we should be repairing the pipeline (cf. the end of repair).
             assert(self.status == .view_change);
             assert(self.primary_index(self.view) == self.replica);
             assert(self.commit_stage == .idle);
             assert(self.commit_min == self.commit_max);
+            assert(self.valid_hash_chain_between(self.commit_max, self.op));
 
             // But we still need to check that we are reparing the right prepare.
             const op = self.primary_repair_pipeline_op() orelse {


matklad:
Finally, a _real_ issue uncovered by tickles VOPR!

chaitanyabhandari:
Trying to understand what happened here more deeply! So, by the time `repair_pipeline_read_callback` is invoked on R, view change cycles through all replicas and R is a potential primary _again_, but its a still repairing its journal between self.commit_max → self.op?

matklad:
Correct!

src/vsr/replica.zig
```diff
@@ -7278,11 +7278,21 @@ pub fn ReplicaType(
                 return;
             }
 
+            if (self.journal.find_latest_headers_break_between(self.commit_max, self.op)) |range| {
```
chaitanyabhandari, line 7281: Thinking about this a bit more, are we still vulnerable to the scenario where we're a potential primary with:
* commit_min == commit_max,
* no missing headers between commit_max → op,
* _but_ missing *prepares* between commit_max → op

So, I'm wondering if the correct thing to do here is to check `if (self.primary_journal_repaired())`? This also mirrors our `assert(self.primary_journal_repaired());` in `primary_repair_pipeline`, where we *start* repairing our pipeline.

matklad: I _think_ we can safely embrace concurrency here! While we might not have all of the prepares, we've just read one, and, given that our hash-chain is intact, we can correctly insert it into the pipeline cache. 

`primary_journal_prepares_repaired` is best effort anyway --- with our storage fault model, we can't _really_ know our state of disk. We might _think_ we have everything, but there might be data corruption!

I am wondering even if we should embrace concurrency in the `fn repair`, and kick the pipeline repair even before journal's prepares are repaired? 

```diff
diff --git a/src/vsr/replica.zig b/src/vsr/replica.zig
index d3b8d2de1..bc056d972 100644
--- a/src/vsr/replica.zig
+++ b/src/vsr/replica.zig
@@ -6878,7 +6878,6 @@ pub fn ReplicaType(
                 if (self.log_view > self.superblock.staging.vsr_state.log_view) {
                     self.view_durable_update();
                 }
-                if (!self.primary_journal_prepares_repaired()) return;
 
                 if (self.commit_min == self.commit_max) {
                     if (self.commit_stage != .idle) {
@@ -6896,7 +6895,10 @@ pub fn ReplicaType(
                     switch (self.primary_repair_pipeline()) {
                         // primary_repair_pipeline() is already working.
                         .busy => {},
-                        .done => self.primary_start_view_as_the_new_primary(),
+                        .done => {
+                            if (!self.primary_journal_prepares_repaired()) return;
+                            self.primary_start_view_as_the_new_primary();
+                        },
                     }
                 }
             }
```

I think I'll keep the code simpler for now though!

chaitanyabhandari: Aha, of course! I'd forgot about the case where a prepare between commit_max → op gets corrupted _after_ we've repaired our entire journal but _before_ we step up as primary. Indeed, it makes sense to embrace concurrency here.

chaitanyabhandari (approved):
LGTM! Last I saw the state of the CFO on this, it was green at around a million seeds. Right now though, there's a failed seed `./zig/zig build -Drelease vopr -- 9347239620680081785` _but_ it also fails on main.

From the looks of it, it _seems_ like we've exhausted all messages? Will look at this on Monday. 🤯 
```C
thread 83829911 panic: attempt to use null value
/Users/chaitanyabhandari/Desktop/tigerbeetle/tigerbeetle/src/message_pool.zig:245:39: 0x1024dc8ef in send_message (vopr)
        const message = pool.free_list.?;
                                      ^
/Users/chaitanyabhandari/Desktop/tigerbeetle/tigerbeetle/src/testing/cluster/message_bus.zig:81:33: 0x1024ed6bb in send_reply_message_to_client (vopr)
        bus.network.send_message(message, .{
                                ^
/Users/chaitanyabhandari/Desktop/tigerbeetle/tigerbeetle/src/vsr/replica.zig:5849:50: 0x1024c40f3 in ignore_request_message_duplicate (vopr)
                self.send_reply_message_to_client(reply);
                                                 ^
/Users/chaitanyabhandari/Desktop/tigerbeetle/tigerbeetle/src/vsr/replica.zig:5640:54: 0x102481a43 in on_request (vopr)
            if (self.ignore_request_message_duplicate(message)) return true;
```

matklad:
>From the looks of it, it seems like we've exhausted all messages? Will look at this on Monday. 🤯

Aha, this is another bug in ticklers VOPR itself! In `PacketSimulator.step`, we dequeue up to 16 messages at a time, and then start processing them. But, once messages are dequeued, it is possible to submit _more_ messages.

PR #2846, opened 2025-03-24
vopr: re-do tickless VOPR

matklad:
This is a second take at https://github.com/tigerbeetle/tigerbeetle/pull/2823. The idea is the same: with many IOPs
ready at any given tick, we want to randomize the order in which IOPs
are executed.

The implementation is different --- rather than collecting a list of
ready IOPS and shuflling it, I add a helper function to pick a random
IOP directly off the queue. This keeps the queue as the real source of
truth about pending operations.

The "source of truth property" is important:

* For upgrades, where we clear the queue and expect that that removes
  all operations.
* For packet simulator, where we size message pool accoring to the size
  of the queue.

### DIFF
--- a/src/testing/fuzz.zig
+++ b/src/testing/fuzz.zig
@@ -4,6 +4,7 @@ const std = @import("std");
 const stdx = @import("../stdx.zig");
 const assert = std.debug.assert;
 const PRNG = stdx.PRNG;
+const ratio = stdx.PRNG.ratio;
 
 /// Returns an integer of type `T` with an exponential distribution of rate `avg`.
 /// Note: If you specify a very high rate then `std.math.maxInt(T)` may be over-represented.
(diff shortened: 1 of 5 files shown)


matklad:
@chaitanyabhandari hopefully this is the last time you have to review this :P Sorry about not getting this right the first, the second, and the third time!

src/testing/storage.zig
```diff
@@ -349,101 +319,54 @@ pub const Storage = struct {
...
+            .{ .read, .write }
+        else
+            .{ .write, .read };
+
+        for (order) |kind| switch (kind) {
+            .read => if (storage.reads.remove_ready(&storage.prng, storage.ticks)) |read| {
+                assert(read.ready_at_tick <= storage.ticks);
+                storage.read_sectors_finish(read);
+                advanced = true;
+                break;
```
chaitanyabhandari, line 346: Curious what the reason behind the `break` is here, could we not simply process one pending request from _each_ queue every time `step` is called? 🤔

src/testing/fuzz.zig
```diff
@@ -77,3 +78,108 @@ pub fn parse_seed(bytes: []const u8) u64 {
...
+        // All items in the queue, in an unspecified order.
+        pub fn const_slice(ready: *const ReadyQueue) []const T {
+            return ready.queue.items;
+        }
+
+        pub fn count(ready: *ReadyQueue) usize {
+            return ready.queue.items.len;
+        }
+
+        // For a subtree rooted at index, return the total number of items in the subtree as
```
chaitanyabhandari, line 146:
```suggestion
         // For a subtree rooted at index, return the total number of items in the subtree as
        // well as an index of a randomly uniformly drawn ready item.
        //
        // The implementation relies on a particular layout of std.PriorityQueue, and might require
        // vendoring it in the future.
```

chaitanyabhandari: Some typos in there.

src/testing/storage.zig
matklad: We could, but then we'll always have a perfect interleaving of reads and writes. Picking _one_ at random allows sequences like `rrrrwwrrrwwwww`

src/testing/fuzz.zig
matklad: Ah, my spell checker was off, thanks!

src/testing/storage.zig
chaitanyabhandari: Ah makes total sense, thank you for clarifying!

chaitanyabhandari:
This looks good to me, looking at all the assertions about storage state being removed was just so satisfying. Awesome job at pushing tickless VOPR to 90% (and beyond)!

I will approve this as soon as https://github.com/tigerbeetle/tigerbeetle/pull/2854 is merged, since CI is blocked right now.

chaitanyabhandari (approved):

PR #2850, opened 2025-03-24
vopr: various changes for performance mode

matklad:
Best review per-commit!

### DIFF
--- a/src/testing/cluster.zig
+++ b/src/testing/cluster.zig
@@ -235,8 +235,9 @@ pub fn ClusterType(comptime StateMachineType: anytype) type {
             errdefer allocator.free(replica_pools);
 
             // There may be more clients than `clients_max` (to test session eviction).
+            // +1 is for pulse which uses client_id = 0.
             const pipeline_requests_limit =
-                @min(options.cluster.client_count, constants.clients_max) -|
+                (@min(options.cluster.client_count, constants.clients_max) + @as(u8, 1)) -|
                 constants.pipeline_prepare_queue_max;
 
             for (replica_pools, 0..) |*pool, i| {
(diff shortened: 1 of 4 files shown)


src/testing/cluster.zig
```diff
@@ -235,8 +235,9 @@ pub fn ClusterType(comptime StateMachineType: anytype) type {
             errdefer allocator.free(replica_pools);
 
             // There may be more clients than `clients_max` (to test session eviction).
+            // +1 is for pulse which uses client_id = 0.
             const pipeline_requests_limit =
-                @min(options.cluster.client_count, constants.clients_max) -|
+                (@min(options.cluster.client_count, constants.clients_max) + @as(u8, 1)) -|
```
chaitanyabhandari, line 240: This is a great catch!

One observation though, I see that `pipeline_request_queue_max` is set using `clients_max`: https://github.com/tigerbeetle/tigerbeetle/blob/1c3dfec5c952ec628bdee39143748e14a24ddec8/src/constants.zig#L265-L268

So, when the # of clients chosen for VOPR runs is >= clients_max, the below assert in MessagePool could get triggered, no?
https://github.com/tigerbeetle/tigerbeetle/blob/1c3dfec5c952ec628bdee39143748e14a24ddec8/src/message_pool.zig#L49

Perhaps we need a `+1` in `pipeline_request_queue_max` as well?

chaitanyabhandari (approved):
Regarding https://github.com/tigerbeetle/tigerbeetle/pull/2850#discussion_r2012952211, I _just_ saw https://github.com/tigerbeetle/tigerbeetle/pull/2848/files and then also saw that the code snippet that I've attached already has a `+1` added. I was following the code locally, and hadn't pulled the latest changes. 

LGTM!

PR #2853, opened 2025-03-26
[Node] Handle "too much data" error.

batiati:
Fix a panic when running against `--development` clusters.  
The node client validates the batch size only against its local `message_size_max` constant, without accounting for the possibility that the server may be running with a smaller batch size.

### DIFF
--- a/src/clients/node/node.zig
+++ b/src/clients/node/node.zig
@@ -290,8 +290,8 @@ fn on_completion(
         .client_release_too_low,
         .client_release_too_high,
         .client_shutdown,
+        .too_much_data,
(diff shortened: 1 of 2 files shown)


lewisdaly:
So when re run in `--development` that's what causes the batch size to be smaller and then crashes the node client? Am I understanding that correctly?

batiati:
> So when re run in `--development` that's what causes the batch size to be smaller and then crashes the node client? Am I understanding that correctly?

Exactly. The bug happens for runtime configurations (e.g., using the `--development` flag or explicitly reduced messages with the `--limit_request` flag).

lewisdaly (approved):
thanks!

PR #2854, opened 2025-03-26
ci: block merging based on devhub pipeline result

cb22:
Fixes that #2840 got merged, even though the devhub check was failing.

(I'll fix that shortly too, just want to ensure that this works!)


cb22:
Whoops, didn't mean to assign just yet.

matklad:
(fwiw, I _almost_ reached the activation energy to redo our pipeline again, to make sure that the only thing we have there is `zig/download.sh && ./zig/zig build ci`).

matklad (approved):

PR #2855, opened 2025-03-26
fuzz: remove some code duplication

matklad:


src/lsm/tree_fuzz.zig
```diff
@@ -770,15 +770,11 @@ const Model = struct {
...
-    const avg_int: Int = if (prng.boolean())
-        // 1. We want to cause many collisions.
-        constants.lsm_growth_factor * 2048
-    else
-        // 2. We want to generate enough ids that the cache can't hold them all.
-        100 * constants.lsm_growth_factor * 2048;
-    return fuzz.random_int_exponential(prng, Int, avg_int);
+fn random_id(prng: *stdx.PRNG) u64 {
+    return fuzz.random_id(prng, u64, .{
+        .average_hot = constants.lsm_growth_factor * 2048,
```
matklad: @batiati `constants.lsm_growth_factor * 2048,` feels like way to much for the purpose of forcing collisions here. Do you have a suggestion for a better natural constant here? Something like `64` would work better here, imo?

batiati: Yes, it seems to have been just copied from fuzzer to fuzzer without much reasoning.

Specifically in `cache_map_fuzzer.zig`, I don't know why it should be based on `lsm_growth_factor` at all.
Here in `tree_fuzz.zig` would it be appropriate to use the same values we have in `forest_fuzz.zig`?

matklad: Adjusted the constants!

batiati (approved):

PR #2856, opened 2025-03-26
tidy: fix off by one

matklad:
bytes_read is a count, not an index, we don't neeed -1 there.

Found on the stream: https://www.youtube.com/live/Lvp3Le7R3OM?feature=shared&t=3375

### DIFF
--- a/src/tidy.zig
+++ b/src/tidy.zig
@@ -39,7 +39,7 @@ test "tidy" {
     // to read the files once.
     for (paths) |path| {
         const bytes_read = (try std.fs.cwd().readFile(path, buffer)).len;
-        if (bytes_read == buffer.len - 1) return error.FileTooLong;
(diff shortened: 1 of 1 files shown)


src/tidy.zig
```diff
@@ -39,7 +39,7 @@ test "tidy" {
     // to read the files once.
     for (paths) |path| {
         const bytes_read = (try std.fs.cwd().readFile(path, buffer)).len;
-        if (bytes_read == buffer.len - 1) return error.FileTooLong;
+        if (bytes_read >= buffer.len - 1) return error.FileTooLong;
```
batiati, line 42: I guess the -1 is to accommodate the \n char, isn't it?

Otherwise all this time we were enforcing 99 characters instead of 100 😱😳

batiati: Oh! its not the LineTooLong error👍 makes sense now

batiati (approved):

PR #2858, opened 2025-03-27
vopr: print dropped packets in VOPR

matklad:
Worth reviewing per-commit for a plot twist!

### DIFF
--- a/src/vsr/message_header.zig
+++ b/src/vsr/message_header.zig
@@ -239,6 +239,23 @@ pub const Header = extern struct {
         }
     }
 
+    pub fn format(
+        self: *const Header,
+        comptime fmt: []const u8,
+        options: std.fmt.FormatOptions,
+        writer: anytype,
+    ) !void {
+        switch (self.into_any()) {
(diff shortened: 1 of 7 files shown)


src/vsr/message_header.zig
```diff
@@ -239,6 +239,23 @@ pub const Header = extern struct {
...
     }
 
+    pub fn format(
+        self: *const Header,
+        comptime fmt: []const u8,
+        options: std.fmt.FormatOptions,
+        writer: anytype,
+    ) !void {
+        switch (self.into_any()) {
+            inline else => |header| return try std.fmt.formatType(
```
sentientwaffle, line 249: (Note that the "try" here is not strictly needed.)

sentientwaffle (approved):

src/vsr/message_header.zig
matklad: I like `return try` because it makes it clear that the operation is failable. 

I think there’s even a lang proposal to make it mandatory?

PR #2859, opened 2025-03-27
Update correcting-transfers.md example

shraddha38:
## Description
The transfers are happening from account `A` to accounts `X` and `Y`.
But in the explanation it says to accounts `Y` and `Z`
Screenshot 2025-03-27 at 2 11 45 PM

## Changes Proposed 

Corrected Example wordings in correcting-transfers.md 

- Let's say you had a couple of transfers, from account A to accounts X and Y:


batiati (approved):
Nice catch @shraddha38!  Thanks a lot!

PR #2860, opened 2025-03-27
[Fix] AccountFilter and QueryFilter timestamp range

batiati:
Improper validation of timestamps from query filters (`get_account_{transfers,balances}` and `query_{accounts,transfers}`) could cause the replica to crash if a value greater than a `u63` was provided. (The last bit of the timestamp is reserved for the tombstone flag.)


sentientwaffle (approved):
LGTM. We could alternatively check `== 0` inside `valid` instead of checking it at the caller.

batiati:
> LGTM. We could alternatively check `== 0` inside `valid` instead of checking it at the caller.

I explicitly left zero as invalid, since it means "no filter"

PR #2861, opened 2025-03-27
state machine: add basic fuzzer

cb22:
~This is probably still too basic to catch anything, but worth letting it brew on CFO for a bit.~

Before rebasing, it did catch #2860 almost immediately!

### DIFF
--- a/src/state_machine_fuzz.zig
+++ b/src/state_machine_fuzz.zig
@@ -64,6 +64,10 @@ pub fn main(allocator: std.mem.Allocator, args: fuzz.FuzzArgs) !void {
         };
         const request_buffer_operation = request_buffer_full[0..operation_size];
 
+        // This is checked in the state machine, but as of writing the fuzz, all that it cares about
+        // is the length. Maybe that changes in the future and you hit this assert.
+        assert(context.state_machine.input_valid(operation, request_buffer_operation));
+
         switch (operation) {
             .query_accounts, .query_transfers => {
                 const query_filter: *tb.QueryFilter = @ptrCast(request_buffer_operation);
@@ -132,10 +136,6 @@ pub fn main(allocator: std.mem.Allocator, args: fuzz.FuzzArgs) !void {
             .get_events, .pulse => continue,
         }
 
-        // This is checked in the state machine, but as of writing the fuzz, all that it cares about
-        // is the length. Maybe that changes in the future and you hit this assert.
-        assert(context.state_machine.input_valid(operation, request_buffer_operation));
-
         context.prepare(operation, request_buffer_operation);
 
         const reply_size = context.execute(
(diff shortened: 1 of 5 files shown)


sentientwaffle:
I removed the `fuzz state_machine` label b/c using a fuzz label which the CFO doesn't know about causes it to crash:

```
thread 87042 panic: reached unreachable code
/var/home/djg/C/bin/zig-linux-x86_64-0.13.0/lib/std/debug.zig:412:14: 0x11c802d in assert (scripts)
             ^
/tmp/tigerbeetle/src/scripts/cfo.zig:643:19: 0x11dbbd3 in run_fuzzers_prepare_tasks (scripts)
            assert(pr_fuzzers_count >= 2); // The canary and at least one different fuzzer.
                  ^
/tmp/tigerbeetle/src/scripts/cfo.zig:238:42: 0x11eb43a in run_fuzzers (scripts)
            try run_fuzzers_prepare_tasks(&tasks, shell, gh_token);
                                         ^
/tmp/tigerbeetle/src/scripts/cfo.zig:173:20: 0x11ee355 in main (scripts)
    try run_fuzzers(shell, gpa, gh_token_option, .{
                   ^
/tmp/tigerbeetle/src/scripts.zig:99:40: 0x12cc3e9 in main (scripts)
        .cfo => |args_cfo| try cfo.main(shell, gpa, args_cfo),
                                       ^
/var/home/djg/C/bin/zig-linux-x86_64-0.13.0/lib/std/start.zig:524:37: 0x11c7a85 in posixCallMainAndExit (scripts)
                                    ^
/var/home/djg/C/bin/zig-linux-x86_64-0.13.0/lib/std/start.zig:266:5: 0x11c75a1 in _start (scripts)
```

We can start using the label once this PR is merged.

src/state_machine_fuzz.zig
```diff
@@ -0,0 +1,151 @@
...
+                    },
+                };
+            },
+            .lookup_accounts, .lookup_transfers => {
+                const account_or_transfer_id: *u128 = @ptrCast(request_buffer_operation);
+                account_or_transfer_id.* = int_edge_biased(&prng, u128);
+            },
+
+            // TODO: Only reads are supported now, due to how create_* require compaction to be
+            // hooked up.
```
batiati, line 132: We could at least call `input_valid` for those operations.

src/state_machine_fuzz.zig
```diff
@@ -0,0 +1,151 @@
...
+                    .timestamp_max = int_edge_biased(&prng, u64),
+                    .limit = int_edge_biased(&prng, u32),
+                    .reserved = reserved,
+                    .flags = .{
+                        .reversed = prng.boolean(),
+                        .padding = if (prng.chance(.{ .numerator = 1, .denominator = 1000 }))
+                            int_edge_biased(&prng, u31)
+                        else
+                            0,
+                    },
```
batiati, line 95: There are many `if (field != 0)` when we build the scan in `get_scan_from_account_filter` and `get_scan_from_query_filter`.

Do we have significant occurrences of "zeros" here to stress the combination of these code paths?

batiati (commented):
:100:

src/state_machine_fuzz.zig
cb22: Good question! So testing `int_edge_biased`, we get about ~15k zeroes from 10 million integers. or ~0.15% chance of hitting a zero.

We could increase that?

cb22: Currently, it'll do 10m iterations in ~5 seconds, and it does 9 `int_edge_biased` calls for either `query_*` or `get_account_*`.

batiati: > We could increase that?

Having zeros "sometime" is good enough 👍

Since we are not asserting semantically the results, we dont require testing the combination off all possible scans here (for this purpose we have the scan fuzzer and the vopr).

batiati (approved):

PR #2821, opened 2025-03-14
vsr: quicker request protocol

matklad:
Old protocol:

- client round-robins the target replica during retries
- replica forwards a request to the primary _if_ the message is from the
  earlier view
- if the request is from the same view, it is dropped

This really doesn't work at all --- if a single message is dropped or
delayed, the client has to go through the entire ring unitl it re-sends
back to the primary.

New protocol:

- client always sends a message to a replica it thinks to be the primary.
- on retry, client sends a round-robin ping message to learn if a view
  change happened.
- backups never forward requests

Things intentionally omitted:

- When a client receives a pong and advances its view, it _does not_
  reset the on_request timeout. This is to spread out the thundering
  heard after a view change, when the cluster might have a backlog of
  prepares.
- Replicas never forward a request, as that is prone to amplify
  bandwidth when the cluster isn't fully healthy.
- Replicas _do not_ respond with a pong to a request with a wrong view.
  Because the client always sends request to the primary, and the
  primary might be down, it _has_ to send a ping to someone else, and
  it's that replica's job to inform the client about the view.
- The original `.register` request is not broadcast, and follows the same logic.
  This is a bit unintelligent, as client's initial view might be wrong,
  and broadcasting feels better. However, I am not sure our on_request
  timeout is set correctly: in local VOPR testing with very fast
  storage, somewhat faulty networks, and multiple clients, the clients
  often do useless retries, which lead to "request already
  queued/preparing" on the primary. I will look into the timeout
  separately, and, for now, use the same logic for .register as for all
  other messages, as that's just simpler.

Results for `./zig/zig build vopr -- --performance 92` with one client and 10% packet loss:

```
Before:

Messages:
ping                     10650   5.20MiB
pong                     9574    2.34MiB
commit                   3615    903.75KiB
prepare                  1563    578.34KiB
prepare_ok               1428    357.00KiB
request                  561     202.00KiB
reply                    280     106.84KiB
request_prepare          228     57.00KiB
request_headers          157     39.25KiB
headers                  139     501.00KiB
ping_client              66      16.50KiB
pong_client              59      14.75KiB
request_start_view       51      12.75KiB
start_view               51      138.50KiB
do_view_change           25      12.50KiB
total                    28447   10.41MiB

          PASSED (35075 ticks)
```

```
After:

Messages:
ping                     3210    1.57MiB
pong                     2890    722.50KiB
prepare                  1682    727.20KiB
prepare_ok               1518    379.50KiB
commit                   1140    285.00KiB
[...]

PR #2862, opened 2025-03-28
devhub: track repl execution time

cb22:
It's currently much slower than it should be, due to the request protocol and waiting for extra ticks. Target time is ~6ms.

### DIFF
--- a/src/scripts/devhub.zig
+++ b/src/scripts/devhub.zig
@@ -164,7 +164,7 @@ fn devhub_metrics(shell: *Shell, cli_args: CLIArgs) !void {
         break :blk stats_count;
     };
 
-    const startup_time_ms = blk: {
+    const startup_time_ms, const repl_single_command_ms = blk: {
         timer.reset();
 
         var process = try shell.spawn(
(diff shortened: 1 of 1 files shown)


matklad:
I wonder whether https://github.com/tigerbeetle/tigerbeetle/pull/2821 made any difference for repl?

cb22:
> I wonder whether https://github.com/tigerbeetle/tigerbeetle/pull/2821 made any difference for repl?

Ah, I didn't notice that had been merged! I don't think so, ~620ms was my original time a few weeks ago, and it seemed to be the same when I tested this against latest main a few hours ago.

cb22:
Protocol was the wrong choice of words in the description! It's not the protocol as such, but more some timings / ticks / delays in message bus that cause ~600ms latency. The remaining ~20ms is down to extra ticks. I'll revisit those shortly, but I can't seem to find my code from [here](https://tigerbeetleinc.slack.com/archives/C07QCNUL43E/p1741352977577879). :sweat_smile:

sentientwaffle (approved):

PR #2863, opened 2025-03-28
format: make concurrent and only write essential data

cb22:
This pulls the concurrent format out of #2368, and greatly speeds up the format time. On my machine, it goes from ~1s to <100ms:

```
/t/m/s/g/tigerbeetle/tigerbeetle main• ❱ time ./tigerbeetle format --replica-count=1 --replica=0 --cluster=0 /mnt/original
2025-03-28 15:15:38.147Z warning: a cluster id of 0 is reserved for testing and benchmarking, do not use in production
2025-03-28 15:15:38.147Z warning: omit --cluster=0 to randomly generate a suitable id

2025-03-28 15:15:38.148Z info(io): creating "original"...
2025-03-28 15:15:38.148Z info(io): allocating 1.06298828125GiB...
2025-03-28 15:15:38.982Z info(main): 0: formatted: cluster=0 replica_count=1

________________________________________________________
Executed in  836.10 millis    fish           external
   usr time  641.26 millis  281.00 micros  640.98 millis
   sys time  179.51 millis  138.00 micros  179.38 millis

/t/m/s/g/tigerbeetle/tigerbeetle main• ❱ sha256sum /mnt/original
19e942abb69c987dc4bed93656af4a37dce3b5d720a33dfab55a78938f746390  /mnt/original
```

vs

```
~/g/t/tigerbeetle-2 cb22/concurrent-format• ❱ time ./tigerbeetle format --replica-count=1 --replica=0 --cluster=0 /mnt/concurrent
2025-03-28 15:16:30.133Z warning: a cluster id of 0 is reserved for testing and benchmarking, do not use in production
2025-03-28 15:16:30.133Z warning: omit --cluster=0 to randomly generate a suitable id

2025-03-28 15:16:30.133Z info(io): creating "concurrent"...
2025-03-28 15:16:30.133Z info(io): allocating 1.06298828125GiB...
2025-03-28 15:16:30.228Z info(main): 0: formatted: cluster=0 replica_count=1

________________________________________________________
Executed in   97.90 millis    fish           external
   usr time   51.82 millis    0.00 micros   51.82 millis
   sys time   13.84 millis  979.00 micros   12.86 millis

~/g/t/tigerbeetle-2 cb22/concurrent-format• ❱ sha256sum /mnt/concurrent
19e942abb69c987dc4bed93656af4a37dce3b5d720a33dfab55a78938f746390  /mnt/concurrent
```

Most of the speedup comes from skipping writing the full block for the WAL: we've fallocated() the file, so we only need to write the header. Even if there was garbage data, it shouldn't be a problem.


matklad:
@cb22 should this have a reviewer? As in, is it ready for review?

cb22:
@matklad not yet! I'm not happy with how I'm using `format_wal_prepares` etc currently.

src/vsr/replica_format.zig
```diff
@@ -292,4 +382,14 @@ test "format" {
...
     }
+
+    // Lastly, verify the entire storage contents against a known good checksum for the given
+    // cluster, replica and replica count.
+    //
+    // This doesn't match the output from `tigerbeetle format ...` since the testing storage / slot
+    // counts are lower.
+    try std.testing.expectEqual(
+        vsr.checksum(storage.memory),
+        234838825150141811691318382366967529672,
```
cb22, line 416: Worth noting: this is the same value from before the changes, and the `sha256sum` of the datafile from before and after, on a 0.0.1 release is the same too (`19e942abb69c987dc4bed93656af4a37dce3b5d720a33dfab55a78938f746390`).

src/vsr/replica_format.zig
```diff
@@ -26,195 +22,286 @@ pub fn format(
...
-            self.formatting = false;
+            assert(self.formatting_superblock);
+            self.formatting_superblock = false;
         }
     };
 }
 
+/// Write a WAL header into the start of header_buffer, and return a pointer to the written header
+/// as well as the offset its matching prepare would go into.
+pub fn wal_header(cluster: u128, slot: u64, header_buffer: []u8) struct { *Header.Prepare, u64 } {
```
cb22: IMO, this is much simpler than the previous functions (`format_wal_headers` and `format_wal_prepares`) and doesn't really suit a fuzzer; I'd rather exhaustively check it below.

cb22:
@matklad much happier now! Ready for review.

src/vsr/replica_format.zig
```diff
@@ -26,195 +22,287 @@ pub fn format(
...
```
matklad: Hm, it feels like it's just two different functions fused together? 

```zig
pub fn wal_header(cluster: u128, slot: u64, header_buffer: []u8) *Header.Prepare
pub fn wal_prepare_offset(slot: u64) u64
```

matklad: And it also feels that maybe _without_ the first function the code would be more readable? 

```
if (slot == 0)
        Header.Prepare.root(cluster)
    else
        Header.Prepare.reserve(cluster, slot);
```

is the only logic of substance in this function, so perhaps we need to 

```
fn slot_header(cluster: u128, slot: u64) Header.Prepare
```

and then inline offset/buffer management at the call site, to remove a level of pointer intdirection?

src/vsr/replica_format.zig
```diff
@@ -26,195 +22,287 @@ pub fn format(
...
+    assert(header_buffer.len >= @sizeOf(Header.Prepare));
+
+    stdx.maybe(header_buffer.len % constants.sector_size == 0);
+
+    const prepare_offset = slot * constants.message_size_max;
+    assert(prepare_offset <= constants.journal_size_prepares);
+    assert(prepare_offset % @sizeOf(Header) == 0);
+    assert(prepare_offset % constants.sector_size == 0);
+
+    const header: *Header.Prepare = @alignCast(
```
matklad: If we keep this function, we should push alignment requirement into the signature

src/vsr/replica_format.zig
```diff
@@ -26,195 +22,287 @@ pub fn format(
...
-                    wal_offset,
+            // Direct I/O requires the buffer to be sector-aligned. Unlike the Prepares above that
+            // require a buffer per prepare, since they are spread out with zeros inbetween, the
+            // headers zone is contiguous so a single buffer will do.
+            const headers_buffer = try arena.alignedAlloc(
+                u8,
+                constants.sector_size,
+                vsr.sector_ceil(constants.journal_size_headers),
+            );
+            @memset(headers_buffer, 0);
```
matklad: In this memset and the one below, lets be more intentional and zero out only the padding bytes. Blank `memset` has a higher chance of being correct by accident, but it is hard to reason about, as it doesn't informed the user which specific bytes are padding.

src/vsr/replica_format.zig
```diff
@@ -26,195 +22,287 @@ pub fn format(
...
+
                 // Direct I/O requires the buffer to be sector-aligned.
-                const padding_buffer = try allocator.alignedAlloc(
+                const padding_buffer = try arena.alignedAlloc(
                     u8,
                     constants.sector_size,
                     vsr.Zone.size(.grid_padding).?,
                 );
-                defer allocator.free(padding_buffer);
+
```
matklad: drop blank line between alloc and memset

src/vsr/replica_format.zig
```diff
@@ -26,195 +22,287 @@ pub fn format(
...
             self: *ReplicaFormat,
-            allocator: std.mem.Allocator,
+            arena: std.mem.Allocator,
             storage: *Storage,
         ) !void {
-            assert(!self.formatting);
+            assert(!self.formatting and !self.formatting_superblock);
 
             // Direct I/O requires the buffer to be sector-aligned.
             const message_buffer =
```
matklad: remove \n after `=`, to keep formatting consistent. And no blank line before memset

src/vsr/replica_format.zig
```diff
@@ -26,195 +22,287 @@ pub fn format(
...
-            defer allocator.free(message_buffer);
+                try arena.alignedAlloc(
+                    u8,
+                    constants.sector_size,
+                    constants.message_size_max * constants.clients_max,
+                );
+
             @memset(message_buffer, 0);
 
             for (0..constants.clients_max) |slot| {
```
matklad: Why do we need a loop here? Isn't this effectively just a single big write? Do we need to write anything here at all, given that we no longer try to zero-out the entire file (for wal prepares, we only writing headers).

src/vsr/replica_format.zig
```diff
@@ -26,195 +22,287 @@ pub fn format(
...
+/// * 1 write that contains all of the headers,
+/// * constants.clients_max many replies,
+/// * 1 block for grid padding, maybe.
+pub const writes_max = constants.journal_slot_count + 1 + constants.clients_max + 1;
+
 fn ReplicaFormatType(comptime Storage: type) type {
     const SuperBlock = vsr.SuperBlockType(Storage);
     return struct {
         const ReplicaFormat = @This();
+        var formating_instance: ?*ReplicaFormat = null;
```
matklad: I do this sort of global mutable variable thing in `testing/fixtures.zig`, and, as that's testing code, that's probably.

But this is production code, so I think we shouldn't assume that there's only single global instance of formatting. Let's do the proper thing and define our own 

```zig
const Write = struct {
    write: Storage.Write, 
    replica_format: *ReplicaFormat,
}
```

src/vsr/replica_format.zig
cb22: > Do we need to write anything here at all

No! And we don't need to write anything for the padding either!

matklad:
Out of curiosity, if, before formatting, we do a single vectored write that splats a single random sector over the entire data file, how long would it take?

matklad (commented):
Ah, I think we can do a bit of online-verification here, to check the contract twice!

Define a `write_sectors` function on `ReplicaFormat`, which delegates to storage, but also flips bits in a bitset for sectors actually written (or maybe even flip them in the callback, if we still have access to zone/offset/size there)?

At the end of format, do a linear for loop over all sectors, and assert that sector being physically written according to the bitset is equivalent to sector being required to be written to.

src/vsr/replica_format.zig
```diff
@@ -26,195 +22,215 @@ pub fn format(
...
-        write: Storage.Write = undefined,
 
-        fn format_wal(
+        writes: [writes_max]Write = undefined,
+        writes_pending: u64 = 0,
+
+        fn queue_format_wal(
             self: *ReplicaFormat,
-            allocator: std.mem.Allocator,
+            arena: std.mem.Allocator,
```
matklad: ❤️

src/vsr/replica_format.zig
```diff
@@ -26,195 +22,215 @@ pub fn format(
...
-                    }
+            for (0..constants.journal_slot_count) |slot| {
+                // Direct I/O requires the buffer to be sector-aligned. Allocate a buffer for each
+                // sector in the arena, so they can be written concurrently.
+                const header_buffer = try arena.alignedAlloc(
+                    u8,
+                    constants.sector_size,
+                    constants.sector_size,
+                );
+                const header: *Header.Prepare = @alignCast(
```
matklad: No need to `@alignCast` here. 

The one below I think is necessary. We can get rid of it by casing then slicing, rather than slicing then casting, but I think the current way is better.

cb22:
> Out of curiosity, if, before formatting, we do a single vectored write that splats a single random sector over the entire data file, how long would it take?

I'd expect pretty close to sequential throughput speed your drive can manage, with fio on mine, that's ~300ms / 3.3GiB/s.

cb22:
> At the end of format, do a linear for loop over all sectors, and assert that sector being physically written according to the bitset is equivalent to sector being required to be written to.

Done! (With an exception of the superblock...)

matklad:
Dead code! (related to me adding KiB and friends to stdx).

src/vsr/replica_format.zig
```diff
@@ -27,205 +23,288 @@ pub fn format(
...
+pub const writes_max = constants.journal_slot_count + 1;
+
 fn ReplicaFormatType(comptime Storage: type) type {
     const SuperBlock = vsr.SuperBlockType(Storage);
     return struct {
         const ReplicaFormat = @This();
+        const SectorBitSet = std.bit_set.ArrayBitSet(
+            usize,
+            @divExact(data_file_size_min, constants.sector_size),
+        );
```
matklad: Let's use dynamic bitset here, we don't gain anything by adding another copy of bitset's code to our binary.

src/vsr/replica_format.zig
```diff
@@ -27,205 +23,288 @@ pub fn format(
...
+                    assert(self.sectors_written.isSet(sector));
+                } else if (sector_start >= vsr.Zone.wal_prepares.start() and
+                    (sector_start - vsr.Zone.wal_prepares.start()) %
+                        constants.message_size_max == 0 and
+                    sector_start < vsr.Zone.wal_prepares.start() + vsr.Zone.wal_prepares.size().?)
+                {
+                    assert(self.sectors_written.isSet(sector));
+                } else {
+                    assert(!self.sectors_written.isSet(sector));
+                }
```
matklad: These if conditions are somewhat complex, and it isn't obvious that we indeed handle all the cases exhaustively. 

Let maybe do something like this instead:

```
const zone = for (std.enums.values(vsr.Zone)) |zone| {
  if (sector in zone) break zone;
} else unreachable; 

switch zone {
}
```

to separate out the question of which zone the write belongs two, from the question of whether it should be written or not.

matklad (approved):

PR #2866, opened 2025-03-31
vopr: fix false positive

matklad:
When looking for available replies, we should only consider .up replicas, otherwise we are stepping into uninitialized memory!

Seed: ./zig/zig build -Drelease vopr -Dvopr-state-machine=testing -- 11119430380926947877

### DIFF
--- a/src/testing/cluster/state_checker.zig
+++ b/src/testing/cluster/state_checker.zig
@@ -5,6 +5,7 @@ const mem = std.mem;
 const constants = @import("../../constants.zig");
 const vsr = @import("../../vsr.zig");
 const stdx = @import("../../stdx.zig");
+const maybe = stdx.maybe;
 
 const message_pool = @import("../../message_pool.zig");
 const MessagePool = message_pool.MessagePool;
@@ -35,7 +36,10 @@ pub fn StateCheckerType(comptime Client: type, comptime Replica: type) type {
 
         replicas: []const Replica,
         clients: []const Client,
+        /// Tracks the latest reply for every non-evicted client.
+        client_replies: std.AutoArrayHashMapUnmanaged(u128, vsr.Header.Reply),
         clients_exhaustive: bool = true,
+        clients_register_op_latest: u64 = 0,
 
         /// The number of times the canonical state has been advanced.
         requests_committed: u64 = 0,
@@ -62,6 +66,10 @@ pub fn StateCheckerType(comptime Client: type, comptime Replica: type) type {
                 .replicas = commit_replicas,
             });
 
+            var client_replies: std.AutoArrayHashMapUnmanaged(u128, vsr.Header.Reply) = .{};
+            try client_replies.ensureTotalCapacity(allocator, constants.clients_max);
+            errdefer client_replies.deinit(allocator);
+
             const replica_head_max = try allocator.alloc(ReplicaHead, options.replicas.len);
             errdefer allocator.free(replica_head_max);
             for (replica_head_max) |*head| head.* = .{ .view = 0, .op = 0 };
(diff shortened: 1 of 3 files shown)


src/vopr.zig
```diff
@@ -1146,7 +1146,8 @@ pub const Simulator = struct {
         assert(simulator.core.count() > 0);
 
         var replies_latest = [_]?vsr.Header.Reply{null} ** constants.clients_max;
-        for (simulator.cluster.replicas) |replica| {
+        for (simulator.cluster.replicas, simulator.cluster.replica_health) |replica, health| {
+            if (health == .down) continue;
```
matklad: TBH, something feels off about this check:

* for any cluster run, imagine a run where every replica not in core is down
* then this loop is equivalent to a loop over only core replicas
* but then it tautologically returns true

Should we perhaps track required replies in some other way? Should we perhaps extend StateChecker's `on_message` to track replies and explicitly capture which replies we _should_ have?

chaitanyabhandari: Mhmmmmmm interesting case indeed! I was thinking that we could *also* solve this by not introducing additional logic in `on_message`, and instead just making sure the memory for replicas that are down is valid? 

I had done that in https://github.com/tigerbeetle/tigerbeetle/pull/2708; instead of calling `deinit` in `replica_crash` like we currently do, we call `replica_deinit` _before_ every invocation to `replica_open` (and also in `deinit` in cluster.zig). https://github.com/tigerbeetle/tigerbeetle/blob/74847f2f8fd1bb457fd66282562ee55e0c3fd7ff/src/testing/cluster.zig#L557-L569

Thoughts @matklad?

matklad: Oh, I like that! Having the memory available would also be super helpful for debugging the issues, and is worth doing just for that reason!

Though, separately from this, I guess I am unsure about the approach we have right now, where we trust _some_ replica's memory to retain the right answer. So, if replica's collude to erase their memory (due to a bug), we wouldn't notice it! I'd be more comfortable checking against _externalized_ state (messages sent), but that's a separate thing.

matklad: Ah, this doesn't quite work! I was hoping we can achieve an invariant that all cluster's replicas memories are always valid, but this doesn't actually work. A replica can fail to open, in which case we won't have any valid memory at all to keep around.

This happens in "Cluster: recovery: WAL prepare corruption (R=1, corrupt between checkpoint and head)"

matklad: Ok, take three in!

chaitanyabhandari: Yes, the invariant that all the replicas' memories are always valid doesn't actually work, for the case of wild VOPR even more so because `open` can fail there during VOPR runs as well, due to `WALCorrupt`! I had to introduce a new status called `dead` there, and ensure that we don't have `dead` replicas to `simulator.core`.

Also, indeed the case where the replica memory is untrustworthy due to a bug is a tricky one, I didn't understand that that was what you were referring to before! I'm going to review take 3 now.

src/testing/cluster/state_checker.zig
```diff
@@ -72,25 +80,66 @@ pub fn StateCheckerType(comptime Client: type, comptime Replica: type) type {
...
-            state_checker.commits.deinit();
+
             allocator.free(state_checker.replica_head_max);
+            state_checker.client_replies.deinit(allocator);
+            state_checker.commits.deinit();
+        }
+
+        pub fn on_client_eviction(state_checker: *StateChecker, client_id: u128) void {
+            const removed = state_checker.client_replies.swapRemove(client_id);
+            maybe(removed);
```
chaitanyabhandari, line 98: Mhmmmmm this is subtle! The way I understood it was that even when a client has been evicted from `replica.client_sessions` and removed from `state_checker.client_replies`, a replayed `register` message can cause it to get re-added to `replica.client_sessions`. However, it may not get added to `state_checker.client_replies` again due to the primary crashing (and no reply message being sent). Then, if the client is evicted from `replica.client_sessions` again, this `swamRemove` would return `False`!

Does that sound about right @matklad or is there a more straightforward case I'm missing?

chaitanyabhandari (approved):
LGTM; I like the approach of tracking new clients using `clients_register_op_latest`. This is also VOPR green at ~350k seeds, approving!

src/testing/cluster/state_checker.zig
matklad: I was thinking of a simpler case: Primary A adds a client to its client_sessions, evicting someone, and then crashes. Primary B takes over, and adds replays the eviction.

chaitanyabhandari: Ah yes, thank you!

PR #2867, opened 2025-03-31
clients: remove dependency on libc from the C client

matklad:
We only need libc for an allocator, and, given that we don't allocate after startup, we don't need a good alocator, anything would do!

So let's just use GPA!

We could further improve this by using per-client allocator with leak checking, but let's do things step-by-step, my primary goal here is reducing dimensionality via removing dependency on external code (libc).

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -917,7 +917,6 @@ fn build_vortex(
         .target = options.target,
         .optimize = options.mode,
     });
-    tb_client.linkLibC();
     tb_client.pie = true;
     tb_client.bundle_compiler_rt = true;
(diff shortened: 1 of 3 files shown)


matklad:
Ah, ok, this'll require some more thinking, it looks like we can't directly spawn threads without libc, as `pub var tls_image: TLSImage = undefined;` is uninitialized

batiati:
I tried it before, and indeed we need linking libc as a workaround for https://github.com/ziglang/zig/issues/17062

However, we don't use thread local storage, so it is quite possible to bypass this check vendoring the thread spawn function.

PR #2870, opened 2025-04-02
feat: fuzzy search

shraddha38:
**Fuzzy Search Implementation**
This pull request introduces a fuzzy search feature to enhance the user experience when searching through content.

Current Implementation - 
Prefix matching 

Proposed Changes:

1. Fuzzy Matching: Allows users to find results even with minor typos or partial matches.

2. One Bug fix as well 

The following assumption in the test is wrong because it expects true to true kinda scenario 
```try std.testing.expectEqualSlices(u8, c.query.const_slice(), c.query.const_slice());```
 changed it to 
```try std.testing.expectEqualSlices(u8, c.query.const_slice(), t.query.const_slice());```

### CODE REVIEW

File: src/repl/completion.zig [hunk 1]
```diff
@@ -128,47 +140,65 @@ test "completion.zig: Split buffer and complete" {
...
-            .prefix = try BoundedArray(u8, completion_entry_bytes).from_slice(
-                "create_accounts id=1 ",
-            ),
-            .suffix = try BoundedArray(u8, completion_entry_bytes).from_slice(" ledger=700"),
-            .query = try BoundedArray(u8, completion_entry_bytes).from_slice("co"),
-            .matches = try BoundedArray([]const u8, 5).from_slice(&.{"code"}),
-        },
-    };
+        matches: BoundedArray([]const u8, 10),
+    }{ .{
```
brson: The formatting of this nested brace has changed, and that is resulting in github seeing the whole block as a diff, making it tough to read. I wonder if you could restore the original formatting so the diff is reduced to just meaningful changes.

brson: I think it's because all subsequent lines become indented.

brson:
Nice idea. I'm going to think give it a try and think about it a bit.

src/repl/completion.zig
brson: I've pushed a commit to address this.

brson:
@shraddha38 I've run this a bit locally, compared it to other repls and thought about how I would use it, and I'm afraid I don't understand what the intended benefit of this behavior is.

As illustration, after this PR, I can type "b" at the repl:

```
> b
```

And then TAB:

```
> get_account_balances 
 get_account_balances  debit_account_id 
```

And I get all "keywords" that contain the letter "b".

The most useful way I see to use this is to "fish" for completions - e.g. press a random letter, then tab, just to see some valid keywords. But as a way to discover keywords this does not seem intuitive to me. And neither the python or julia repls have completions that behave this way. I would be afraid that this completion behavior would often provide irrelevant completions.

Please do let me know if there are uses of this I'm not seeing.

To be constructive, if I were to think about how to make the repl's keywords more discoverable I might add context-sensitivity to the completions. For example, if I am already entering a `create_accounts` command, then completions would only be for `create_accounts` field names. If I am trying to complete a request name, then completions would only be for request names (not fieldnames). If I am at the beginning of a blank line (or perhaps after a space in field position), then the repl might offer completions of all possible requests (this is something julia does, python does not).

brson (changes requested):

matklad:
Closing draft due to inactivity

PR #2874, opened 2025-04-02
vsr: call hook into vopr while sending message to client

chaitanyabhandari:
In 5de082c227b4ebb72aa11945a132004cb0bc5e26, we added new logic to track replies based on when they are externalized from replica to client. Turns out this new logic was not being exercised at all, since we only call the hook into the VOPR when messages are sent to other replicas, not when they are sent to the client!

Fix it by calling the hook when messages are sent to the client, this would lead to reply messages getting appropriately tracked.

Seed: ./zig/zig build -Drelease vopr -Dvopr-state-machine=testing -- 10711934386281284252

### DIFF
--- a/src/vopr.zig
+++ b/src/vopr.zig
@@ -1158,19 +1158,6 @@ pub const Simulator = struct {
     pub fn core_missing_reply(simulator: *const Simulator) ?vsr.Header.Reply {
         assert(simulator.core.count() > 0);
 
-        var replies_latest = [_]?vsr.Header.Reply{null} ** constants.clients_max;
-        for (simulator.cluster.replicas) |replica| {
-            for (replica.client_sessions.entries, 0..) |entry, entry_slot| {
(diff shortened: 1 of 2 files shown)


src/vopr.zig
```diff
@@ -1158,19 +1158,6 @@ pub const Simulator = struct {
...
-                if (entry.session != 0) {
-                    if (replies_latest[entry_slot] == null or
-                        replies_latest[entry_slot].?.op < entry.header.op)
-                    {
-                        replies_latest[entry_slot] = entry.header;
-                    }
-                }
-            }
-        }
-
```
matklad, line 1173: 🤦

chaitanyabhandari:
The failed seed `./zig/zig build -Drelease vopr -- 5733619977068526902` is unrelated to this one, it fails on main as well.

matklad (approved):

PR #2877, opened 2025-04-03
vopr: simplify time handling

matklad:
I don't see a strong reason why Replica must own the time. Injecting it as a pointer feels simpler.


matklad:
@sentientwaffle am I missing some reason why we didn't do `*Time` in the first place, and instead went with a more complex TimePointer approach?

matklad:
Ah, we can even completely remove the time field from replica it seems!

sentientwaffle (approved):
I think the only reason is that this code was written early on, before I had a good understanding of when a type should own versus borrow a dependency.

PR #2878, opened 2025-04-03
Avoid std.time in favor of our own Time

matklad:
Before this PR, tracer used raw os time, rather than our `Time` abstraction. This PR changes Tracer to use Time like everything else in the codebase, and involves _mostly_ threading types throughout. 

The motivation is a bit roundabout. For dynamic routing, I need to store timing information about prepares and prepare_oks, and that involves storing both "duration" and "instant" in the single data structure. Originally I went with `u64`, and of course confused myself over what is duration and what is instant. So I decided to switch to stdx.Instant/Duration, but the snag there was that `stdx.Instant.now` is non-deterministic. I considered adding two versions of this `now` function, but than figured out that that'd be confusing, and its' better to stick to the principle that there should be only one thing! 

One thing that the current PR doesn't do is that we use `std.Timer` in a couple of places, and that really should be using our `Time` as well, or perhaps even `Tracer` directly, but I figured out that that's a refactor for the future. 

Best reviewed per commit (and builds on top of #2877)

Official soundtrack of the pull request: https://open.spotify.com/album/0sMhDEURXDVVqiqh45DeAi?si=bdYAJ-mKS86bEbTOiVhTEg


src/lsm/groove.zig
```diff
@@ -1434,47 +1434,3 @@ pub fn GrooveType(
...
-            .optional = &[_][]const u8{},
-            .derived = .{},
-            .orphaned_ids = true,
-        },
-    );
-
-    std.testing.refAllDecls(Groove);
-    std.testing.refAllDecls(Groove.PrefetchWorker);
-    std.testing.refAllDecls(Groove.PrefetchContext);
-}
```
matklad, line 1480: Figured that I'd rather delete these kinds of test, as I don't think they are bring value relative to zig build check.

src/stdx.zig
```diff
@@ -923,23 +923,12 @@ pub fn array_print(
...
-
-    pub fn duration_since(now_: Instant, earlier: Instant) Duration {
-        if (now_.base.order(earlier.base) == .lt) {
-            return .{ .nanoseconds = 0 };
-        } else {
-            const elapsed_ns = now_.base.since(earlier.base);
-            return .{ .nanoseconds = elapsed_ns };
-        }
+    pub fn duration_since(now: Instant, earlier: Instant) Duration {
+        assert(now.base_ns >= earlier.base_ns);
```
matklad, line 929: Our time is for real monotonic, so we can assert this.

src/storage.zig
```diff
@@ -8,7 +8,7 @@ const stdx = vsr.stdx;
 const FIFOType = vsr.fifo.FIFOType;
 const constants = vsr.constants;
 
-pub fn StorageType(comptime IO: type) type {
+pub fn StorageType(comptime IO: type, comptime _Tracer: type) type {
```
matklad, line 11: I still don't know a good pattern for naming this to avoid shadowing.

src/trace.zig
```diff
@@ -112,304 +112,324 @@ pub const EventMetricAggregate = @import("trace/event.zig").EventMetricAggregate
...
-    }
-
-    pub fn cancel(tracer: *Tracer, event_tag: Event.Tag) void {
-        const stack_base = EventTracing.stack_bases.get(event_tag);
-        const cardinality = EventTracing.stack_limits.get(event_tag);
-        for (stack_base..stack_base + cardinality) |stack| {
-            if (tracer.events_started[stack]) |event_start| {
-                log.debug("{}: {s}: cancel", .{ tracer.replica_index, @tagName(event_tag) });
-
-                const event_end = stdx.Instant.now();
```
matklad, line 295: It is a pet-peeve of mine where `.now()` is called several times in the same function. I like to cache such reads, not for performance, but to make sure that things that logically happened at the same time get the same physical timestamp.

src/trace.zig
```diff
@@ -112,304 +112,324 @@ pub const EventMetricAggregate = @import("trace/event.zig").EventMetricAggregate
...
     try snap(@src(),
         \\[
-        \\{"pid":0,"tid":0,"ph":"B","ts":<snap:ignore>,"cat":"replica_commit","name":"replica_commit  stage=idle","args":{"stage":"idle","op":123}},
-        \\{"pid":0,"tid":4,"ph":"B","ts":<snap:ignore>,"cat":"compact_beat","name":"compact_beat  tree=Account.id","args":{"tree":"Account.id","level_b":1}},
-        \\{"pid":0,"tid":4,"ph":"E","ts":<snap:ignore>},
-        \\{"pid":0,"tid":0,"ph":"E","ts":<snap:ignore>},
+        \\{"pid":0,"tid":0,"ph":"B","ts":0,"cat":"replica_commit","name":"replica_commit  stage=idle","args":{"stage":"idle","op":123}},
+        \\{"pid":0,"tid":4,"ph":"B","ts":10000,"cat":"compact_beat","name":"compact_beat  tree=Account.id","args":{"tree":"Account.id","level_b":1}},
+        \\{"pid":0,"tid":4,"ph":"E","ts":20000},
+        \\{"pid":0,"tid":0,"ph":"E","ts":60000},
```
matklad, line 422: Bonus, no longer need cut-outs here!

src/vsr/replica.zig
```diff
@@ -1187,9 +1185,6 @@ pub fn ReplicaType(
                 .release_execute = options.release_execute,
                 .release_execute_context = options.release_execute_context,
                 .nonce = options.nonce,
-                // Copy the (already-initialized) time back, to avoid regressing the monotonic
-                // clock guard.
-                .time = self.time,
```
matklad, line 1192: Also glad to get rid of this hack.

src/shell.zig
```diff
@@ -152,10 +152,9 @@ pub fn open_section(shell: *Shell, name: []const u8) !Section {
 const Section = struct {
     ci: bool,
     name: []const u8,
-    start: stdx.Instant,
+    timer: std.time.Timer,
```
matklad, line 155: I don't want to import vsr in shell, so I refactored this one "for the worse" :D Doesn't matter for shell, I think.

sentientwaffle (approved):
Thanks for breaking this up into different commits! It made review much simpler.

matklad:
>Originally I went with u64, and of course confused myself over what is duration and what is instant. 

As a validation, I _immediatly_ found a new bug when porting routing to Instant/Duration!

PR #2880, opened 2025-04-03
vsr: adaptive routing based on randomized explore/exploit

matklad:
TigerBeetle replication routing protocol.

Eight fallacies of distributed computing:

1. The network is reliable;
2. Latency is zero;
3. Bandwidth is infinite;
4. The network is secure;
5. Topology doesn't change;
6. There is one administrator;
7. Transport cost is zero;
8. The network is homogeneous;

Robust tail principle:
- The same code should handle slow nodes and crashed nodes.
- A single crashed node should not cause retries and risk metastability.

Algorithm:

The replication route is V-shaped. Primary is in the middle, and it tosses each prepare at two
of its neighbors. Backups forward prepares to at most one further backup. Neighbors of the
primary have at least one more neighbor (in a six-replica cluster). If any single node fails,
the primary still gets a replication quorum. If the primary and backups disagree about the
replication route, the primary still gets a replication quorum.

Because topology changes, routes are dynamic. The primary broadcasts the current route in the
ping message. It's enough if the routing information is only eventually consistent.

To select the best route, primary uses outcome-focused explore-exploit approach. Every once in a
while the primary tries an alternative route. The primary captures replication latency for a
route (that is, the arrival time of prepare_ok messages). If the latency for an alternative
route is sufficiently better than current latency, the route is switched. Note that latency
includes both network and disk latency.

The experiment schedule is defined randomly. All replicas share the same RNG seed, so no
coordination is needed to launch an experiment!

To remove outliers, the experiment take two ops. op and op^1 are either both experimental, or
both non-experimental. Experimental route cost is the average of two costs. Active route cost
is maintained as exponential weighted moving average.

### DIFF
--- a/src/vsr/replica.zig
+++ b/src/vsr/replica.zig
@@ -434,6 +434,8 @@ pub fn ReplicaType(
             cache: PipelineCache,
         },
 
+        routing: vsr.Routing,
+
         /// When "log_view < view": The DVC headers.
         /// When "log_view = view": The SV headers. (Just as a cache,
         /// since they are regenerated for every request_start_view).
@@ -828,6 +830,7 @@ pub fn ReplicaType(
                 // completes.
                 self.log_view += 1;
                 self.view += 1;
+                self.routing.view_change(self.view);
                 self.primary_update_view_headers();
                 self.view_durable_update();
 
@@ -1285,6 +1288,11 @@ pub fn ReplicaType(
                     .capacity = constants.pipeline_prepare_queue_max +
                         options.pipeline_requests_limit,
                 } },
+                .routing = vsr.Routing.init(.{
+                    .replica = replica_index,
+                    .replica_count = replica_count,
+                    .standby_count = standby_count,
+                }),
                 .view_headers = vsr.Headers.ViewChangeArray.init(
                     self.superblock.working.vsr_headers().command,
                     self.superblock.working.vsr_headers().slice,
@@ -1381,6 +1389,7 @@ pub fn ReplicaType(
                 .aof = options.aof,
                 .replicate_options = options.replicate_options,
             };
+            self.routing.view_change(self.view);
 
             log.debug("{}: init: replica_count={} quorum_view_change={} quorum_replication={} " ++
                 "release={}", .{
@@ -1441,6 +1450,7 @@ pub fn ReplicaType(
 
         pub fn invariants(self: *const Replica) void {
             assert(self.journal.header_with_op(self.op) != null);
+            assert(self.view == self.routing.view);
         }
 
         /// Time is measured in logical ticks that are incremented on every call to tick().
@@ -1703,6 +1713,15 @@ pub fn ReplicaType(
                 .ping_timestamp_monotonic = message.header.ping_timestamp_monotonic,
                 .pong_timestamp_wall = @bitCast(self.clock.realtime()),
             }));
+
+            if (self.status == .normal and self.backup()) {
+                if (message.header.view == self.view and message.header.route != 0) {
+                    const route = self.routing.route_decode(message.header.route).?;
+                    if (!std.mem.eql(u8, route.const_slice(), self.routing.a.const_slice())) {
+                        self.routing.route_activate(route);
+                    }
+                }
+            }
 
             if (message.header.replica < self.replica_count) {
                 const upgrade_targets = &self.upgrade_targets[message.header.replica];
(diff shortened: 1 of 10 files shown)


src/vsr/replica.zig
```diff
@@ -3229,8 +3268,8 @@ pub fn ReplicaType(
 
             self.prepare_timeout.backoff(&self.prng);
 
-            assert(waiting_len <= self.replica_count);
-            for (waiting[0..waiting_len]) |replica| {
+            assert(waiting_count < self.replica_count);
```
matklad, line 3467: Changed this to be strict as we don't count ourselves

src/vsr/routing.zig
```diff
@@ -0,0 +1,877 @@
...
+
+/// The route we currently use.
+a: Route,
+a_cost: ?Cost, // Computed as exponential weighted moving average.
+
+/// The best alternative, which might be better or worse than a.
+b: Route,
+b_cost: ?Cost, // Computed as an average of two experiments.
+
+experiment_chance: Ratio = ratio(1, 20),
```
matklad, line 71: No good feeling for what's a good constant here, 5% seems ok.

src/vsr/routing.zig
```diff
@@ -0,0 +1,877 @@
...
+
+        op_history.prepare_ok[replica] = Duration.min(latency, latency_max);
+        op_history.present.set(replica);
+    }
+};
+
+const Cost = struct {
+    // Left-biased median latency matches replication quorum latency, and directly contributes to
+    // user-visible latency.
+    median: Duration,
```
matklad, line 111: Originally I parametrized the whole thing by the quorum size, but I think hard-coding median is the same in effect and leads to a very nice definition of cost.

src/vsr/routing.zig
```diff
@@ -0,0 +1,877 @@
...
+        return .{
+            .median = ewma_add_duration(old.median, new.median),
+            .maximum = ewma_add_duration(old.maximum, new.maximum),
+            .sum = ewma_add_duration(old.sum, new.sum),
+        };
+    }
+
+    fn ewma_add_duration(old: Duration, new: Duration) Duration {
+        return .{
+            .ns = @divFloor((old.ns * 4) + new.ns, 5),
```
matklad, line 188: 4, 1 coefficients are also just arbitrary reasonable numbers.

src/vsr/routing.zig
```diff
@@ -0,0 +1,877 @@
...
+        };
+    }
+};
+
+const Routing = @This();
```
matklad, line 193: I am trying out a new pattern where 

```
count This = @This();
```


Is the last type defined, which separates fields&types from methods.

src/vsr/routing.zig
```diff
@@ -0,0 +1,877 @@
...
+
+/// The best alternative, which might be better or worse than a.
+b: Route,
+b_cost: ?Cost, // Computed as an average of two experiments.
+
+experiment_chance: Ratio = ratio(1, 20),
+
+history: [history_max]OpHistory,
+
+pub const Route = stdx.BoundedArrayType(u8, constants.replicas_max);
```
matklad, line 75: I find it more readable to declare fields before types, as fields are more important.

src/vsr/routing.zig
```diff
@@ -0,0 +1,877 @@
...
+        .b = route,
+        .b_cost = null,
+
+        .history = .{OpHistory.root()} ** history_max,
+    };
+}
+
+pub fn view_change(routing: *Routing, view: u32) void {
+    assert(view > routing.view or (view == 0 and routing.view == 0));
+    assert(routing.history_empty());
```
matklad, line 224: The design is that the caller should explicitly call `.view_change` and `.history_reset`. `view_change` _can_ automagically reset the history, but I'd rather have this visible at the call-site.

src/vsr/routing.zig
```diff
@@ -0,0 +1,877 @@
...
+    assert(routing.route_valid(route));
+
+    routing.a = route;
+    assert(routing.a_cost == null);
+
+    routing.b = route;
+    assert(routing.b_cost == null);
+}
+
+fn route_valid(routing: *const Routing, route: Route) bool {
```
matklad, line 237: NB: this does depend on the current view.

src/vsr/routing.zig
```diff
@@ -0,0 +1,877 @@
...
+    std.mem.swap(
+        u8,
+        &route.slice()[primary_index],
+        &route.slice()[@divFloor(replica_count, 2)],
+    );
+
+    return route;
+}
+
+pub fn route_encode(routing: *const Routing, route: Route) u64 {
```
matklad, line 290: Originally I went for a crazy scheme that encoded a route in 255 as an index of permutation of `replica_count - 1` replicas. That was fancy enough to warrant a blog-post: https://matklad.github.io/2025/04/15/underusing-snapshot-testing.html. 

So I reconsidered and went for direct encoding of a list of replicas.

matklad:
Marking as ready for review, because it is!

It isn't ready for merge yet --- there's no code to handle transition to the new ping message. I'll add it separately once we hammer out the happy case details here!

matklad:
Wait, no ,let me rebase this on top of git-review tool. This feels like a good PR to test drive that crazy idea...

sentientwaffle (approved):

PR #2881, opened 2025-04-03
vopr: allow customizing network delay

matklad:
This is not used yet, but  I am going to use it for testing routing, in https://github.com/tigerbeetle/tigerbeetle/pull/2880.

There, I'd love to setup "ring" toplogy (in trems of packet delay), and I think the cleanest way to do that is to let the caller pass down a function.

An alterative solution would be passing down a _matrix_ of delay distributions, but that feels like it would be much less readable at the call site. Maybe its wrong, but we can always refactor back to matrix later!


matklad:
Call-site from that branch:

https://github.com/tigerbeetle/tigerbeetle/blob/34330fe24dc0f35f39a3810b45f4988ff6b14d78/src/vsr/routing.zig#L550-L553

chaitanyabhandari (approved):

PR #2882, opened 2025-04-03
zig: build ci

matklad:
Every time I subimt a PR, I fail either `zig fmt`, `zig build check`, or `zig build tidy`. I of course remember which one failed, and run that locally before the next PR. Sadly, I can remeber running two commands at most, and it is always the third one that fails!

So, let's introduce `zig build ci` which runs a reasonable set of checks.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,10 +20,27 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - run: shellcheck ./zig/download.sh
-      - run: ./zig/download.sh
-      - run: ./zig/zig fmt --check .
-      - run: ./zig/zig build check
-      - run: ./zig/zig build scripts -- ci --build-docs
+      - run: ./zig/download.ps1 && ./zig/zig build ci -- smoke
+
+  test:
+    strategy:
+      matrix:
+        include:
+          - { os: 'ubuntu-latest' }
+          - { os: 'ubuntu-latest-arm64' }
+          - { os: 'windows-latest' }
+          - { os: 'macos-latest' }
+          - { os: 'macos-13' }
+    runs-on: ${{ matrix.os }}
+    steps:
+      - run: git config --global core.autocrlf false
+      - if: matrix.os == 'ubuntu-latest' || matrix.os == 'ubuntu-latest-arm64'
+        run: | # Allow unshare for vortex.
+          sudo sysctl -w kernel.apparmor_restrict_unprivileged_unconfined=0
+          sudo sysctl -w kernel.apparmor_restrict_unprivileged_userns=0
+      - uses: actions/checkout@v4
+        with: { fetch-depth: 2147483647 } # Fetch full history for tidy.
+      - run: ./zig/download.ps1 && ./zig/zig build ci -- test
 
   test_alpine:
     runs-on: ubuntu-latest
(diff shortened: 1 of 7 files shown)


zig/download.ps1
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh
+echo `# <#`
+./zig/download.sh
+exit
+#> > $null
+./zig/download.bat
```
cb22, line 6: I'm not quite sure if I love this or hate this :smile:.

.github/workflows/ci.yml
```diff
@@ -20,10 +20,27 @@ jobs:
...
-      - run: ./zig/zig build scripts -- ci --build-docs
+      - run: ./zig/download.ps1 && ./zig/zig build ci -- smoke
+
+  test:
+    strategy:
+      matrix:
+        include:
+          - { os: 'ubuntu-latest' }
+          - { os: 'ubuntu-latest-arm64' }
+          - { os: 'windows-latest' }
```
cb22, line 31: Out-of-scope for PR, but perhaps we should add Windows ARM64 too.

.github/workflows/ci.yml
```diff
@@ -20,10 +20,27 @@ jobs:
...
+          - { os: 'macos-latest' }
+          - { os: 'macos-13' }
+    runs-on: ${{ matrix.os }}
+    steps:
+      - run: git config --global core.autocrlf false
+      - if: matrix.os == 'ubuntu-latest' || matrix.os == 'ubuntu-latest-arm64'
+        run: | # Allow unsare for vortex.
```
cb22:
```suggestion
        run: | # Allow unshare for vortex.
```

.github/workflows/ci.yml
```diff
@@ -225,7 +178,7 @@ jobs:
   core-pipeline:
     if: always() && github.event_name == 'merge_group'
     runs-on: ubuntu-latest
-    needs: [smoke, test_alpine, test_ubuntu, test_aof, test_windows, test_macos, clients, devhub]
+    needs: [smoke, test, test_alpine, test_aof, clients, devhub]
```
cb22: Would be nice if we could somehow drop `test_alpine` :thinking:.

build.zig
```diff
@@ -359,6 +366,147 @@ fn build_vsr_module(b: *std.Build, options: struct {
...
+) void {
+    const run_artifact = b.addRunArtifact(scripts);
+    run_artifact.addArgs(argv);
+    run_artifact.setEnvironmentVariable("ZIG_EXE", b.graph.zig_exe);
+    build_ci_check_status(run_artifact);
+    step_ci.dependOn(&run_artifact.step);
+}
+
+fn build_ci_check_status(step: *std.Build.Step.Run) void {
+    // XXX: The purpose here is not to check status (Zig does that automatically), but rather to
```
cb22: A bit of _careful_ Googling tells me `XXX:` is indeed used as a heads-up (I've always used it more as a `TODO`). That said, we don't use it anywhere else, so perhaps `NB: `?

zig/download.ps1
matklad: I mean, yes, but also consider that the alternative is adding ifs to YAML, which I'd argue is as elegant as putting bash inside powershell :P

.github/workflows/ci.yml
matklad: We don't support building for arm windows, and I'd rather add that only when someone asks. I am confident that we _could_ support whatever, but I want to minimize the time to administrate CI heterogeneity (my mental model is that any CI config has an independent probability of "flakes" for boring reasons).

build.zig
matklad: Yeah, let's use NB as more classy. They have semantic distinction of 

* NB: something unusual happens here
* XXX: something unusual happens here which will bite you in the back if you don't pay attention

but that's perhaps not worth it.

.github/workflows/ci.yml
matklad: I am confident that we can just drop it, which I did.

cb22 (approved):

PR #2887, opened 2025-04-09
LSM: Make object cache optional.

toziegler:
This update makes the `objects_cache` optional within the `Groove`. 
Previously, the `objects_cache` was always maintained, even in scenarios where it was not  used. 
For example, the `AccountEvents` Groove is append-only and is currently accessed only via scans, making the cache unnecessary. Specifically, this is the case if  `prefetch_entries_for_read_max` and `prefetch_entries_for_update_max` are both 0.
We avoid doing unnecessary work by disabling the cache when it isn’t required.

This change gives us roughly 10% overall performance:

Before: load accepted = 328302 tx/s
After:    load accepted = 355223 tx/s

### DIFF
--- a/src/lsm/groove.zig
+++ b/src/lsm/groove.zig
@@ -168,9 +168,9 @@ pub fn GrooveType(
     ///     Whether Groove should store objectless `id`s to prevent their reuse.
     ///     Should be `true` only if the object contains an `id` field.
     ///
-    /// - object_cache: bool:
+    /// - objects_cache: bool:
     ///     Whether Groove should have an ObjectCache.
-    ///     Should be `false` only if both `prefetch_entries_for_update_max` and 
+    ///     Should be `false` only if both `prefetch_entries_for_update_max` and
     ///     `prefetch_entries_for_read_max` are set to 0.
     comptime groove_options: anytype,
 ) type {
@@ -336,7 +336,7 @@ pub fn GrooveType(
 
     const has_scan = index_fields.len > 0;
 
-    const has_objects_cache = groove_options.object_cache;
+    const has_objects_cache = groove_options.objects_cache;
 
     // Verify groove index count:
     const indexes_count_actual = std.meta.fields(_IndexTrees).len;
@@ -563,7 +563,7 @@ pub fn GrooveType(
         /// keeping table_mutable as an array, and simplifying the compaction path
         /// is faster than trying to amortize and save memory.
         ///
-        /// Invariant: if there is an object_cache then if something is in the mutable or immutable 
+        /// Invariant: if there is an object_cache then if something is in the mutable or immutable
         /// table, it _must_ exist in our object cache.
         objects_cache: if (has_objects_cache) ObjectsCache else void,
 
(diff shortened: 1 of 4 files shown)


src/lsm/groove.zig
```diff
@@ -595,29 +602,42 @@ pub fn GrooveType(
...
-                // number of prefetches a bar may perform, excluding prefetches already accounted
-                // for by batch_value_count_limit.
-                .stash_value_count_max = constants.lsm_compaction_ops *
-                    (options.tree_options_object.batch_value_count_limit +
-                    options.prefetch_entries_for_read_max),
-
-                // Scopes are limited to a single beat, so the maximum number of entries in a
-                // single scope is batch_value_count_limit (total – not per beat).
-                .scope_value_count_max = options.tree_options_object.batch_value_count_limit,
+            groove.objects_cache = switch (has_objects_cache) {
```
matklad: Maybe `if` rather than switch?

src/lsm/groove.zig
```diff
@@ -556,9 +563,9 @@ pub fn GrooveType(
         /// keeping table_mutable as an array, and simplifying the compaction path
         /// is faster than trying to amortize and save memory.
         ///
-        /// Invariant: if something is in the mutable or immutable table, it _must_ exist in our
-        /// object cache.
-        objects_cache: ObjectsCache,
+        /// Invariant: if there is an object_cache then if something is in the mutable or immutable 
+        /// table, it _must_ exist in our object cache.
+        objects_cache: if (has_objects_cache) ObjectsCache else void,
```
matklad: Not sure which one is better, but consider making `ObjectCache` itself void: `const ObjectCache = if (has) CacheMap else void`. 

Than you could do 

```
if (ObjectCache != void) groove.objects_cache.reset();
```

which is perhaps a bit more self-documenting.

toziegler: Done :+1: do you think this is better? I am fine with both

matklad:
LGTM personally, but let's assign Rafael to make sure this makes sense overall!

src/lsm/groove.zig
batiati: Just a reminder that we can use Zig's return blocks in this nice pattern:

```zig
            groove.objects_cache = objects_cache: {
                if (ObjectsCache == void) {
                    // If there are no modifications or point lookups on the Groove then
                    // no `objects_cache` is needed.
                    assert(options.prefetch_entries_for_read_max == 0);
                    assert(options.prefetch_entries_for_update_max == 0);
                    break :objects_cache {};
                }

                break :objects_cache try ObjectsCache.init(...);
            };
```
As a TIGERSTYLE thing, we try to make early returns more evident and put the most natural `if` clause first (`==` over `!=` and `<` over `>=`).  
Nothing against if-else; it's still the best option for short statements like `if (x) a else b;`.

batiati (approved):
Nice work Tobi!!

src/lsm/groove.zig
toziegler: Somehow I completely missed your great comment this morning 😬 will fix it

PR #2888, opened 2025-04-10
Rename FIFO to Queue

matklad:
FIFO and LIFO are much easier to confuse with each other than Queue and Stack.


toziegler:
The rename and fuzzer LGTM. One thought I had was that most queues are bounded, should we make something like a `queue_option` that contains the capacity so that we can avoid the unbounded `while` loop? I could also make a separate PR and keep this as is.

toziegler (approved):

PR #2889, opened 2025-04-10
LSM: Remove set associative cache.

toziegler:
This PR removes the set associative cache from the `cache_map.zig`. While this change may initially seem surprising (or even controversial), the motivation are as follows: 

- Performance: No measurable performance impact from removal.
- Safety and simplicity: Reduces complexity and thus surface area of bugs (we had one or two memory issues in this component).

Also it prepares our codebase for a potential upcoming PR where we might introduce our own hash table implementation.

Baseline: load accepted = 327822 tx/s
After: load accepted = 330576 tx/s

### DIFF
--- a/src/lsm/cache_map.zig
+++ b/src/lsm/cache_map.zig
@@ -66,12 +66,6 @@ pub fn CacheMapType(
             name: []const u8,
         };
 
-        // The hierarchy for lookups is cache (if present) -> stash -> immutable table -> lsm.
-        // Lower levels _may_ have stale values, provided the correct value exists
-        // in one of the levels above.
-        // Evictions from the cache first flow into stash, with `.compact()` clearing it.
-        // When cache is null, the stash mirrors the mutable table.
-        cache: ?Cache,
         stash: Map,
 
         // Scopes allow you to perform operations on the CacheMap before either persisting or
@@ -86,13 +80,6 @@ pub fn CacheMapType(
             maybe(options.cache_value_count_max == 0);
             maybe(options.scope_value_count_max == 0);
 
-            var cache: ?Cache = if (options.cache_value_count_max == 0) null else try Cache.init(
-                allocator,
-                options.cache_value_count_max,
-                .{ .name = options.name },
-            );
-            errdefer if (cache) |*cache_unwrapped| cache_unwrapped.deinit(allocator);
-
             var stash: Map = .{};
             try stash.ensureTotalCapacity(allocator, options.stash_value_count_max);
             errdefer stash.deinit(allocator);
@@ -104,7 +91,6 @@ pub fn CacheMapType(
             errdefer scope_rollback_log.deinit(allocator);
 
             return CacheMap{
-                .cache = cache,
                 .stash = stash,
                 .scope_rollback_log = scope_rollback_log,
                 .options = options,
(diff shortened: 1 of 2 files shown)


matklad:
> Performance: No measurable performance impact from removal.

Is this only for default parameters of the benchmark, or does this hold for more or less all parameters (I imagine contention value can affect this a bit?)

toziegler:
> > Performance: No measurable performance impact from removal.
> 
> Is this only for default parameters of the benchmark, or does this hold for more or less all parameters (I imagine contention value can affect this a bit?)

Great question! Yes, the above was with the following benchmark command: 

```bash
./tigerbeetle benchmark --cache-grid=32GiB --file=/dev/nvme2n1 --query-count=0 
```
I also ran another benchmark using only 16 accounts, with this command:

```bash
./tigerbeetle benchmark --cache-grid=32GiB --file=/dev/nvme2n1 --query-count=0 --account-count=16
```


**Results**: 
baseline:                                  load accepted = 380566 tx/s
without set associative cache: load accepted = 382506 tx/s

From a first-principles perspective, I wouldn't expect a dramatic difference between the two. Hot accounts likely remain in the CPU cache regardless, and our values are relatively large—too big to benefit from tight packing within a single cache line (e.g. it is not possible to have 8 values in one cache line). The main potential of the SetAssociativeCache might come from fewer TLB lookups, since the SetAssociativeCache uses fewer virtual memory pages (denser allocation) than the hash map since we want to store all the values here.

toziegler:
Let me add that I suspect the SetAssociativeCache improved our older benchmarks because it required fewer instructions than Zig's hash table at that time. However, according to the GitHub history at least, zig's hash table has since been improved and now the performance is relative similar.

cb22:
@toziegler ah, could you test it the other way (ie, `--account-count=100_000`+) too. That's where I'd expect there might be a difference: the stash will be cleared at the bar boundary, so it'd have to be repopulated via grid cache (at best, or with an actual read at worst).

(IIRC the default accounts cache size is 1 million.)

toziegler:
@cb22, excellent point—your analysis is spot on:
```bash
--account-count=100000 
```
baseline load accepted = 305056 tx/s
this pr  load accepted = 280798 tx/s  (~ -10%)

I tried to stress the situation that you brought up even more  with this configuration:

```bash
--account-count=100000 --account-count-hot=16
```

baseline load accepted = 324904 tx/s
this pr load accepted = 280688 tx/s (~ -15%)


Overall, we see about a 10 - 15% performance regression in these use-cases  without the set-associative cache. Given the reduction in complexity, this drop is not too bad. However, @matklad and I have an idea to address the need to clear the hash table with our own hash table implementation.

So we should hold off merging this PR for now. I will work on a PoC for the custom hash table.

batiati:
@toziegler can we undo https://github.com/tigerbeetle/tigerbeetle/pull/2774 without reintroducing the TOCTOU bug?

toziegler:
> @toziegler can we undo #2774 without reintroducing the TOCTOU bug?

Good question, @batiati. I _think_ we could undo it in this case. As I understand it, the original bug came from having two sources of truth—cache and stash—and the need to do evictions which invalidates references we hold. With this change, we now have a single hash table and no evictions happening, so that issue shouldn't arise. 

The new hash table I mentioned earlier will be designed to keep pointers stable throughout a bar. The main idea is to perform batch lookups at the beginning of each beat. @matklad pointed out that this approach allows us to lock the pointers upfront and then run our state machine logic over all accounts in that beat. If anything changes unexpectedly, we can immediately trigger an assertion—similar to Zig’s lockPointers API—which makes it much easier to catch bugs early.

PR #2883, opened 2025-04-03
vopr: fix todo

matklad:
Make sure that vopr doesn't leak memory (there was one leak), and remove todo. 

Best reviewed per commit.

PR #2891, opened 2025-04-11
vsr: fix assertion

matklad:
As written, the assertion has both false postives and false negatives.

False positive: it is OK if buffer.ptr == headers.ptr + headers_size, these just means that the two were allocated one after another.

False negative: it's not enough to check that buffer.ptr < headers.ptr, buffer's size needs to be taken into account.

Convieniently, we already have an stdx helper just for that!

Notably, false positive was triggered at https://github.com/tigerbeetle/tigerbeetle/pull/2883


matklad:
Want to land this separately from #2883, as I have a hunch that this is not the only bug there, but it is a real bug, and I don't want to block the fix.

chaitanyabhandari (approved):
Great catch!

Issue #2879, opened 2025-04-03
`register_log_callback()` is exported incorrectly in C client?


tyranron:
In https://github.com/tigerbeetle/tigerbeetle/pull/2742 the [`register_log_callback()` function in C client](https://github.com/tigerbeetle/tigerbeetle/blob/0.16.34/src/clients/c/tb_client.h#L350) was added. It differs from other functions in `tb_client.h` that its name doesn't start with `tb_client_` prefix. However, in [`tb_client_exports.zig`](https://github.com/tigerbeetle/tigerbeetle/blob/0.16.34/src/clients/c/tb_client_exports.zig) all the functions doesn't have such prefix.

Trying to actually call the `register_log_callback()` in FFI code fails with [linkage error](https://github.com/tigerbeetle-rust/tigerbeetle-unofficial/actions/runs/13842038556/job/38731644449#step:5:73) (undefined symbol). [Correcting the exported name as `tb_client_register_log_callback()`](https://github.com/tigerbeetle-rust/tigerbeetle-unofficial/pull/49/commits/768087f7e3c84141b553f40e84de05e505c4a55c) worked for me, and there is no linkage error anymore.

I guess either it really should be exported as `tb_client_register_log_callback()` in `tb_client.h`, or tooling to be corrected, so the linker sees it with the current name.

batiati:
Hi @tyranron 
The log callback is pretty much experimental and unstable, we need to refactor this API and wire it to the language clients before we stabilize that! Please be aware this API will change: https://github.com/tigerbeetle/tigerbeetle/pull/2742#discussion_r1961443332.

However the linkage issue seems unrelated to the function name, since we indeed export it the `tb_client_` prefix:

https://github.com/tigerbeetle/tigerbeetle/blob/775e734cc707e602271db8fbf181111ccf96d825/src/tigerbeetle/libtb_client.zig#L28-L31

batiati:
Ah I see now, it's just the `tb_client.h` that is wrong! Thanks for catching it!

PR #2892, opened 2025-04-11
[tb_client] Fix export name tb_client_register_log_callback

batiati:
Ref https://github.com/tigerbeetle/tigerbeetle/issues/2879


matklad:
Can we add a test that everything exported has the proper prefix?

matklad (approved):
Approving to get this into release, but I'd still love to see some validation here. It is concerning that it doesn't fail, especially given that Python client, as far as I can tell, is trying to call `tb_client_register_log_callback` already

PR #2893, opened 2025-04-14
stdx: de-genericify Stack

matklad:
Move most of the code out of generic Stack into non-generic StackAny.

In terms of machine code, while before `.next` field pointed to `T`, now `.next` points to `.next` directly, which means things need not to be generic.

I keept the wrapping `StackType` just for some type safety. An alternative design would be to just have `StackAny`, but then at the call site you'll be able to pop a node of a wrong type.

See also https://github.com/ziglang/zig/commit/1639fcea43549853f1fded32aa1d711d21771e1c and hat tip to https://www.openmymind.net/Zigs-New-LinkedList-API/ for brining this to my attention.

### DIFF
--- a/src/message_pool.zig
+++ b/src/message_pool.zig
@@ -7,6 +7,7 @@ const constants = @import("constants.zig");
 
 const vsr = @import("vsr.zig");
 const Header = vsr.Header;
+const StackType = @import("./stack.zig").StackType;
 
 comptime {
     // message_size_max must be a multiple of sector_size for Direct I/O
@@ -21,10 +22,10 @@ pub const Options = union(vsr.ProcessType) {
     client,
 
     /// The number of messages allocated at initialization by the message pool.
-    fn messages_max(options: *const Options) usize {
+    fn messages_max(options: *const Options) u32 {
         return switch (options.*) {
             .client => messages_max: {
-                var sum: usize = 0;
+                var sum: u32 = 0;
 
                 sum += constants.replicas_max; // Connection.recv_message
                 // Connection.send_queue:
@@ -48,7 +49,7 @@ pub const Options = union(vsr.ProcessType) {
                 assert(replica.pipeline_requests_limit >= 0);
                 assert(replica.pipeline_requests_limit <= constants.pipeline_request_queue_max);
 
-                var sum: usize = 0;
+                var sum: u32 = 0;
 
                 const pipeline_limit =
                     constants.pipeline_prepare_queue_max + replica.pipeline_requests_limit;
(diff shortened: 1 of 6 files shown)


matklad:
We should do the same for Queue, but I'd rather do that separately, to minimize the diff. 


I've already went on a side-quest of updating some usizes/u64s to u32, don't want this to balloon larger :)

src/stack.zig
```diff
@@ -4,94 +4,131 @@ const assert = std.debug.assert;
...
+
+        // Insert the new element at the head.
+        link.next = self.head;
+        self.head = link;
+        self.count += 1;
+    }
+
+    fn pop(self: *StackAny) ?*StackLink {
+        assert((self.count == 0) == (self.head == null));
+
```
toziegler, line 92: I wonder if we want to have another assertion here too (i.e. the opposite): 
```zig
assert(self.count > 0) == (self.head !=null); 
```
The idea here is that if the count is larger than 0, and the head is null, we loose items. 
But not sure if needed (and I initially did not place this assert). What do you think?

src/message_pool.zig
```diff
@@ -169,8 +171,9 @@ pub const MessagePool = struct {
         }
     };
 
+    const FreeList = StackType(Message);
     /// List of currently unused messages.
-    free_list: ?*Message,
+    free_list: StackType(Message),
```
toziegler, line 176: Nice!

src/stack.zig
```diff
@@ -4,94 +4,131 @@ const assert = std.debug.assert;
...
+        assert(self.count <= self.capacity);
+        var next = self.head;
+        for (0..self.count + 1) |_| {
+            const link = next orelse return false;
+            if (link == needle) return true;
+            next = link.next;
+        } else unreachable;
+    }
+
+    fn reset(self: *StackAny) void {
```
toziegler: This comment is more for me to learn. 
Whenever we have a `reset` function I wonder if it would better to not have default initializations at the top, so that all the variables are enumerated in the reset and visible with one glance.

What is the preference here w.r.t. to default values?

toziegler (commented):
Left one small comment about another possible assertion that we could add, otherwise looks really good to me!

src/stack.zig
matklad: Isn't it tautological? `(self.count == 0) == (self.head == null)` is the same as `(self.count > 0) == (self.head !=null)`. The first one is `A <=> B`, the second one is `!A <=> !B`?

src/stack.zig
matklad: Dropped reset altogether, as we don't need it yet! 

For defaults, I think the fact we write `count = 0` is unidiomatic zig in general.

It would be incorrect for the caller to supply this default, so it shouldn't be a default at all: https://ziglang.org/documentation/0.14.0/#Faulty-Default-Field-Values

But:

* I think we need to wait for 0.14 to apply this conveniently, as this really wants new semantics for `.empty` syntax 
* I am not yet sure I agree with this in general, this is a _very_ subtle rule!

src/stack.zig
toziegler: Ah oh of course you are right, must have been to early for me.

toziegler (approved):

PR #2896, opened 2025-04-14
vopr: compute storage write, read and packet delay latencies with correct mean

chaitanyabhandari:
Currently, storage write, read, and packet delay latencies are computed using:

`min + fuzz.random_int_exponential(&self.prng, u64, mean - min)`

This results in a non-exponentially distributed latency if min is equal to mean. For example, if min=mean for packet delay is X, this leads to the packet delay for all packets to be computed as X, since random_int_exponential always returns zero (μ=0 for the exponential distribution). This is a bug!

The bug manifests in an interesting way: a liveness false positive in single-replica clusters, where non-core standbies are able to starve core standbys using their request_blocks & request_prepare messages. Specifically, in these cases, all request_prepare & request_blocks messages are directed to a single replica. These messages are sent at a fixed cadency. Also, minimum & mean network latency is equal, so our network latency is always computed the same. Therefore, it is possible to come up with a case where R1's requests *always* reach R0 when R2's request is being processed.

Note that in all other cases, where number of replicas is greater than 1, our repair would be robust since we randomly choose a replica to repair blocks/prepares from.

This PR resolves the following seeds:
* ./zig/zig build -Drelease vopr -Dvopr-state-machine=testing -- 10166148228885359786
* ./zig/zig build -Drelease vopr -- 5733619977068526902


src/testing/packet_simulator.zig
```diff
@@ -511,7 +511,7 @@ pub fn PacketSimulatorType(comptime Packet: type) type {
         fn packet_delay_default(self: *PacketSimulator, _: Packet, _: Path) u64 {
             const min = self.options.one_way_delay_min;
             const mean = self.options.one_way_delay_mean;
-            return min + fuzz.random_int_exponential(&self.prng, u64, mean - min);
+            return min + fuzz.random_int_exponential(&self.prng, u64, mean);
```
sentientwaffle: Shouldn't these be `@max(min, fuzz...)`? Otherwise the actual mean is `min + mean`?

src/lsm/forest_fuzz.zig
```diff
@@ -1117,9 +1117,9 @@ pub fn main(gpa: std.mem.Allocator, fuzz_args: fuzz.FuzzArgs) !void {
     try run_fuzz_ops(gpa, Storage.Options{
         .seed = prng.int(u64),
         .read_latency_min = 0,
-        .read_latency_mean = 0 + fuzz.random_int_exponential(&prng, u64, io_latency_mean),
+        .read_latency_mean = prng.range_inclusive(u64, 0, io_latency_mean),
```
sentientwaffle, line 1120: :thinking: Why change from exponential to uniform distribution?

chaitanyabhandari: This was meant to be for consistency, such that at all call sites, we choose the mean value uniformly distributed between a [min, max], since we internally use an exponential distribution using this mean anyway. Thoughts?

src/testing/packet_simulator.zig
chaitanyabhandari: Oops, I think I pushed an old version of the code. Thank you for spotting it, I'll update it!

src/lsm/forest_fuzz.zig
sentientwaffle: Oh I see -- yeah, if we are using an exponential distribution anyway then that seems fine.

sentientwaffle (dismissed):

chaitanyabhandari:
@sentientwaffle oops, found a failing test `Cluster: eviction: client_release_too_low` while running `zig build test` locally. 🤯 

 Taking a look at it. 

```C
[replica] (warn): 1: on_ping_client: ignoring unsupported client version; too low (client=12 version=0.0.9<0.0.10)
thread 104334751 panic: reached unreachable code
/Users/chaitanyabhandari/Desktop/tigerbeetle/tigerbeetle/zig/lib/std/debug.zig:412:14: 0x102bfa377 in assert (test)
             ^
/Users/chaitanyabhandari/Desktop/tigerbeetle/tigerbeetle/src/vsr/replica.zig:8090:19: 0x102d48ff3 in send_eviction_message_to_client (test)
            assert(self.primary());
                  ^
/Users/chaitanyabhandari/Desktop/tigerbeetle/tigerbeetle/src/vsr/replica.zig:5351:53: 0x102ce6107 in ignore_ping_client (test)
                self.send_eviction_message_to_client(
                                                    ^
/Users/chaitanyabhandari/Desktop/tigerbeetle/tigerbeetle/src/vsr/replica.zig:1598:40: 0x102ca7b43 in on_ping_client (test)
            if (self.ignore_ping_client(message)) return;
                                       ^
/Users/chaitanyabhandari/Desktop/tigerbeetle/tigerbeetle/src/vsr/replica.zig:1478:56: 0x102c59e8b in on_message (test)
                .ping_client => |m| self.on_ping_client(m),
                                                       ^
```

chaitanyabhandari:
@sentientwaffle fixed in https://github.com/tigerbeetle/tigerbeetle/pull/2896/commits/2573878ceb59c368b0ca6b4d5949870b2d46db87! New latency distribution exposed a rather obvious bug, where a `ping_client` message from a client with client_release_too_low/client_release_too_high reaches backups in view=0, before the cluster has gone through its first view change.

Still trying to wrap my head around why the VOPR hasn't found this one, it seems like such a low hanging fruit... 🤔

sentientwaffle (approved):

matklad:
> These messages are sent at a fixed cadency.

This feels like a replica-side bug, there should be some timeout randomization happening?

chaitanyabhandari:
@matklad Timeout randomization is certainly one way to tame this. However, as of now, we don't randomize _any_ of our timeouts, they're all fixed! I thought that that was intentional, and we were aiming to solve precisely this by using randomization while _choosing_ the replica that we repair blocks/prepares from. Is my understanding correct?

We _could_ add randomization to the `repair_timeout` and `grid_repair_message_timeout` as well, but IIUC, it would only help in this one-replica case?

matklad:
Hm, ok, yeah, I think randomizing the replica we are contacting should be enough here, doesn't make sense to make the code more complicated just for standbys!

PR #2897, opened 2025-04-15
stdx: align exhaustigen&prng API

matklad:
PRNG steals Int from exhaustigen, as that conveys more information.

Exhaustigen learns `fn index`, as that's conventien, and can be used for `enum_value`.

Also fix typos!

### DIFF
--- a/src/stdx/prng.zig
+++ b/src/stdx/prng.zig
@@ -12,7 +12,7 @@
 //! - isolate our test suite from stdlib API churn
 //! - isolate TigerBeetle from the churn in the PRNG algorithms
 //! - simplify and extend the API
-//! - remove dynamic-dispatch idirection (a minor bonus).
+//! - remove dynamic-dispatch indirection (a minor bonus).
 
 const std = @import("std");
 const stdx = @import("../stdx.zig");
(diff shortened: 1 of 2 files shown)


src/stdx/prng.zig
```diff
@@ -189,26 +189,26 @@ test fill {
...
-/// Genrate an unbiased, uniformly distributed integer r such that 0 ≤ r ≤ max.
+/// Generate an unbiased, uniformly distributed integer r such that 0 ≤ r ≤ max.
 ///
 /// No biased version is provided --- while biased generation is simpler&faster, the bias can be
 /// quite high depending on max!
-pub fn int_inclusive(prng: *PRNG, T: anytype, max: T) T {
-    comptime assert(@typeInfo(T).Int.signedness == .unsigned);
-    if (max == std.math.maxInt(T)) {
-        return prng.int(T);
+pub fn int_inclusive(prng: *PRNG, Int: anytype, max: Int) Int {
```
matklad, line 196: I am confused whether I should approvingly pat myself on the back for noticing a nice pattern elsewhere, and applying it to my code, or scold myself because "elsewhere" was written by past matklad, and apparently I've forgotten the pattern since!

src/stdx/prng.zig
```diff
@@ -503,7 +503,7 @@ test Combination {
 }
 
 /// An iterator style API for selecting a single element out of the given weighted sequence,
-/// without apriory knowledge about the total weight.
+/// without a priory knowledge about the total weight.
```
fabioarnold:
```suggestion
/// without a priori knowledge about the total weight.
```

src/stdx/prng.zig
```diff
@@ -333,16 +333,16 @@ test int {
     ));
 }
 
-fn test_bytes_T(T: type, want: Snap) !void {
+fn test_bytes_T(Int: type, want: Snap) !void {
```
fabioarnold: Should we also rename this function to `test_bytes_Int`?

src/stdx/prng.zig
matklad: I even looked up this in wikipedia and still messed it up, thanks!

src/stdx/prng.zig
matklad: Good catch, done!

fabioarnold (approved):

PR #2899, opened 2025-04-17
LSM: Skip full sort if incremental sorted runs are monotonic and prepare for k-way merge by tracking sorted runs.

toziegler:
This PR prepares for the k-way merge by tracking the sorted runs we produce on every beat. 

To simplify the review of the final k-way merge, this PR introduces tracking of sorted runs. Additionally, it includes an orthogonal optimization enabled by this tracking:

We already produce sorted runs incrementally with sort_suffix. With this PR, we now remember each sorted run's starting and ending indices. By checking the boundaries of these sorted runs, we can determine if they are strictly monotonically increasing. For example, if the end of the first sorted run is smaller than the beginning of the second, we can confirm that the entire table is sorted and skip the final sort.

Example:
```
[1, 2, 3   5, 6, 7]
^            ^ 
|              Second sorted run 
First sorted run
```

While the benchmark improvement is meager (~ 2%), this PR is a first step toward implementing the k-way merge, as tracking the sorted runs is required for that algorithm.

Benchmark baseline: 351437 tx/s
Benchmark this pr:   357342 tx/s

### DIFF
--- a/src/lsm/table_memory.zig
+++ b/src/lsm/table_memory.zig
@@ -237,8 +237,9 @@ pub fn TableMemoryType(comptime Table: type) type {
             var source_index: u32 = offset;
             var target_index: u32 = offset;
             while (source_index < source_count) {
-                const value = values[source_index];
-                values[target_index] = value;
+                if (source_index != target_index) {
+                    values[target_index] = values[source_index];
+                }
 
                 // If we're at the end of the source, there is no next value, so the next value
                 // can't be equal.
--- a/src/lsm/table_memory.zig
+++ b/src/lsm/table_memory.zig
@@ -7,6 +7,9 @@ const binary_search = @import("binary_search.zig");
 const stdx = @import("../stdx.zig");
 const maybe = stdx.maybe;
 
+// BUG: the sorted flag previously started sorted already.
+//       it was set to false if the values were not insert in sorted order
+//       Need to figure out a good way to do this.
 pub fn TableMemoryType(comptime Table: type) type {
     const Key = Table.Key;
     const Value = Table.Value;
@@ -17,8 +20,8 @@ pub fn TableMemoryType(comptime Table: type) type {
 
         pub const ValueContext = struct {
             count: u32 = 0,
-            /// When true, `values` is strictly ascending-ordered (no duplicates).
-            sorted: bool = true,
+            // When true, `values` is strictly ascending-ordered (no duplicates).
+            //sorted: bool = true,
         };
 
         const Mutability = union(enum) {
@@ -39,6 +42,22 @@ pub fn TableMemoryType(comptime Table: type) type {
                 snapshot_min: u64 = 0,
             },
         };
+
+        // `sort_suffix` breaks the `values` array into “sorted runs” (sorted sub‑arrays).
+        // Each SortedRun captures the start (min) and end (max) indices of one sorted run.
+        // We can exploit this knowledge to check if the full array is sorted in O(k) and avoid
+        // the final sort, which reduces tail latency.
+        // The number of calls to `sort_suffix` are determined by `constants.lsm_compaction_ops`.
+        const SortedRun = struct {
+            min: u32, // inclusive
+            max: u32, // exclusive
+        };
+        // Allow one extra sorted run beyond lsm_compaction_ops:
+        // Scans perform an ad‑hoc full-table sort, producing a full run.
+        // Then there could be still number lsm_compaction_ops producing the remaining sorted runs.
+        const sorted_runs_max = constants.lsm_compaction_ops + 1;
+        sorted_runs: [sorted_runs_max]SortedRun,
+        sorted_runs_count: u16,
 
         values: []Value,
         value_context: ValueContext,
@@ -62,6 +81,11 @@ pub fn TableMemoryType(comptime Table: type) type {
                     .mutable => .{ .mutable = .{} },
                     .immutable => .{ .immutable = .{} },
                 },
+                .sorted_runs = .{.{
+                    .min = std.math.maxInt(u32),
+                    .max = std.math.maxInt(u32),
+                }} ** sorted_runs_max,
+                .sorted_runs_count = 0,
                 .name = name,
 
                 .values = undefined,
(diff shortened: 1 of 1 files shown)


src/lsm/table_memory.zig
```diff
@@ -40,6 +40,22 @@ pub fn TableMemoryType(comptime Table: type) type {
...
         };
 
+        // `sort_suffix` breaks the `values` array into “sorted runs” (sorted sub‑arrays).
+        // Each SortedRun captures the start (min) and end (max) indices of one sorted run.
+        // We can exploit this knowledge to check if the full array is sorted in O(k) and avoid
+        // the final sort, which reduces tail latency.
+        // The number of calls to `sort_suffix` are determined by `constants.lsm_compaction_ops`.
+        const SortedRun = struct {
+            min: u32,
+            max: u32,
```
matklad:
```suggestion
            min: u32, // inclusive
            max: u32, // exclusive
```

We sometimes use both-including, like HeaderRange in journal or view Range in replica

src/lsm/table_memory.zig
```diff
@@ -84,6 +102,8 @@ pub fn TableMemoryType(comptime Table: type) type {
             };
 
             table.* = .{
+                .sorted_runs = [_]?SortedRun{null} ** sorted_runs_max,
```
matklad: will just

```suggestion
                .sorted_runs = .{null} ** sorted_runs_max,
```

work?

src/lsm/table_memory.zig
```diff
@@ -196,20 +218,84 @@ pub fn TableMemoryType(comptime Table: type) type {
...
+                table.sorted_runs[0] = SortedRun{ .min = 0, .max = table.count() };
+                table.sorted_runs_count = 1;
+            }
         }
 
         pub fn sort_suffix(table: *TableMemory) void {
             assert(table.mutability == .mutable);
             assert(table.mutability.mutable.suffix_offset <= table.count());
 
+            const min = table.mutability.mutable.suffix_offset;
```
matklad: call these run_min/run_max

src/lsm/table_memory.zig
```diff
@@ -196,20 +218,84 @@ pub fn TableMemoryType(comptime Table: type) type {
...
+                return false;
+            }
+
+            var skip_final_sort = true;
+
+            for (
+                table.sorted_runs[0..(table.sorted_runs_count - 1)],
+                table.sorted_runs[1..table.sorted_runs_count],
+            ) |maybe_sorted_run_a, maybe_sorted_run_b| {
+                const sorted_run_a = maybe_sorted_run_a orelse continue;
```
matklad: these `continue` are actually `unreachable`, right? Let's do this more directly, by not making `SortedRun`s optional, and instead initializing with `min = max = maxInt(u32)`

src/lsm/table_memory.zig
```diff
@@ -196,20 +218,84 @@ pub fn TableMemoryType(comptime Table: type) type {
...
+                const sorted_run_b = maybe_sorted_run_b orelse continue;
+
```
matklad: assert min < max for both runs, and a_max < b_min

🤔 

shouldn't we coalesce runs when we produce them actually? And maybe even get rid of `.sorted`, and add `fn sorted() { return run_count <= 1}`?

So, even if it is not _fully_ sorted, we still get fewer runs?

src/lsm/table_memory.zig
```diff
@@ -237,8 +237,9 @@ pub fn TableMemoryType(comptime Table: type) type {
             var source_index: u32 = offset;
             var target_index: u32 = offset;
             while (source_index < source_count) {
-                const value = values[source_index];
-                values[target_index] = value;
```
matklad, line 241: Surprised this helps!

matklad: I wonder if it makes to even pull this out in a separate loop that finds first mismatch?

src/lsm/table_memory.zig
toziegler: Did I understand it correctly from the comment below that we do not make it optional now?

src/lsm/table_memory.zig
toziegler: Assert :+1:  

Do you mean by coalesce that we combine to sorted runs into one without additional sort /merge? Is that what you mean by not fully sorted?   
```
[1256] [2345]  -> [12562345]
```
The loser tree will only work with fully sorted runs, so the above input would break the algorithm.

src/lsm/table_memory.zig
matklad: Yeah, these are separate suggestions:

* remove optionals in favor of invalid object that will blow up on use anyway
* Simplify array initialization by omitting the type

src/lsm/table_memory.zig
matklad: No, I mean that if we have one run 123, and produce the next run, 456, we then immediately notice that it’s just a single run [123456] rather that carrying [123][456] until the end of the bar

toziegler: That is nice, with the suggestion of `sorted()` it becomes quite clean.

toziegler:
I addressed most of your feedback - I moved the tracking and coalescing of sorted runs in the sort function, which simplified the surrounding code quite a bit. 

Dropping `.sorted` entirely will take a bit more work (that is also why the test fail now) - I want to make sure we do it cleanly. A quick note on the current logic for `.sorted`, for the record:

- We start with `.sorted == true`.
- On every put, we check if we need to invalidate that flag if the inserted value breaks the ordering.

I  _think_ the new `sorted` function must look like this `fn sorted(){return table.sorted_run_count == 1;}`. With `sorted_run_count == 0`, we could have values that are simply not sorted and have not created a sorted run yet.  
Because of that last point, I _think_ we need to begin with a single run like { .min = 0, .max = 0 }. And then expand the sorted run on `put`, if the values are inserted in sorted order. But this also feels convoluted if the `put` invalidates the sort order. We potentially have a sorted run that is `{0,10}{10, MAX}`. 


I’m also unsure how often this scenario really occurs. 
If the incoming values are always sorted,  isn’t that an invariant we could exploit and define explicitly in the Groove?

I need to think a bit more about how to make it clean.

matklad:
@toziegler do we want to close this one for now?

PR #2900, opened 2025-04-17
io: simplify darwing implementation

matklad:
Maybe I am missing something, but I think the new code is equivalent to the old one, and doesn't poke into implementation details of the Queue.

### DIFF
--- a/src/io/darwin.zig
+++ b/src/io/darwin.zig
@@ -77,14 +77,13 @@ pub const IO = struct {
     }
 
     fn flush(self: *IO, wait_for_completions: bool) !void {
-        var io_pending = self.io_pending.peek();
         var events: [256]posix.Kevent = undefined;
 
         // Check timeouts and fill events with completions in io_pending
         // (they will be submitted through kevent).
         // Timeouts are expired here and possibly pushed to the completed queue.
         const next_timeout = self.flush_timeouts();
-        const change_events = self.flush_io(&events, &io_pending);
+        const change_events = self.flush_io(&events);
 
         // Only call kevent() if we need to submit io events or if we need to wait for completions.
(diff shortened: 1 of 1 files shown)


chaitanyabhandari (dismissed):
Gosh, that was one convoluted way to pop completions from `io_pending`! This looks equivalent and correct to me.

src/io/darwin.zig
```diff
@@ -113,11 +112,6 @@ pub const IO = struct {
             );
 
             // Mark the io events submitted only after kevent() successfully processed them
-            self.io_pending.out = io_pending;
-            if (io_pending == null) {
-                self.io_pending.in = null;
-            }
-
             self.io_inflight += change_events;
             self.io_inflight -= new_events;
```
chaitanyabhandari, line 117: Ah, one thing that I _just_ realized, perhaps we can get rid of the `completion.next = null;` in the code snippet below, give that `pop` takes care of it now?

https://github.com/tigerbeetle/tigerbeetle/blob/61dc76b4f2e2627db3ffc36b077443a44c3d0d80/src/io/darwin.zig#L118-L123

chaitanyabhandari, line 117: Ah, one thing that I _just_ realized, perhaps we can get rid of the `completion.next = null;` in the code snippet below, give that `pop` takes care of it now?

https://github.com/tigerbeetle/tigerbeetle/blob/61dc76b4f2e2627db3ffc36b077443a44c3d0d80/src/io/darwin.zig#L118-L123

matklad: Great catch! This totally must be an assert now!

chaitanyabhandari: Ah indeed, even better. 😄

chaitanyabhandari (approved):

PR #2902, opened 2025-04-18
Set socket options for peer connections.

batiati:
Socket options such as keepalive, timeout, no_delay, and send/receive buffer size were previously set only during `listen()`, leaving clients and replica-to-replica sockets with default OS parameters.

### DIFF
--- a/src/io/common.zig
+++ b/src/io/common.zig
@@ -7,7 +7,7 @@ const assert = std.debug.assert;
 
 const is_linux = builtin.target.os.tag == .linux;
 
-pub const ListenOptions = struct {
+pub const SocketOptions = struct {
     rcvbuf: c_int,
     sndbuf: c_int,
     keepalive: ?struct {
@@ -17,6 +17,9 @@ pub const ListenOptions = struct {
     },
     user_timeout_ms: c_int,
     nodelay: bool,
+};
+
+pub const ListenOptions = struct {
     backlog: u31,
 };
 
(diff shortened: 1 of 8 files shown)


src/message_bus.zig
```diff
@@ -212,6 +212,16 @@ fn MessageBusType(comptime process_type: vsr.ProcessType) type {
...
                 .nodelay = constants.tcp_nodelay,
+            });
+            return fd;
+        }
+
+        fn init_tcp_listen(io: *IO, address: std.net.Address) !struct {
+            fd: IO.socket_t,
+            address: std.net.Address,
+        } {
+            const fd = try init_tcp(io, address.any.family);
```
matklad: Given that we've extracted init_tcp, we can now pass `fd` as a parameter, and simply the signature here to return just address.

Also, `errdefer close(fd)` is needed.

src/message_bus.zig
```diff
@@ -188,18 +191,15 @@ fn MessageBusType(comptime process_type: vsr.ProcessType) type {
...
             const fd = try io.open_socket(
-                address.any.family,
+                family,
                 posix.SOCK.STREAM,
                 posix.IPPROTO.TCP,
             );
             errdefer io.close_socket(fd);
 
-            const address_resolved = try io.listen(fd, address, .{
+            try io.socket_options(fd, .{
```
matklad: Instead of separate `io.socket_options` call, consider adding `options` parameter to `io.open_socket` so that, from the perspective of the caller, we atomically get the `fd` we want.

matklad:
Excellent catch! When moving this options to message bus, I was wondering whether its enough to set them when we `.listen` or whether setting them after `.accept` is also required. I convinced myself that doing it on `.listen` is enough, but of course we have the entire other side of connection, where we certainly need also need the options!

src/message_bus.zig
batiati: We can't do that because some options are TCP-specific, and we use `io.open_socket` for UDP connections.
Renamed `socket_options` to `tcp_options` instead, so we can add `udp_options` when/if required.

matklad: Hm, maybe we can `open_socket_tcp` / `open_socket_udp`, so that we also don't have to pass `posix.IPPROTO.TCP` which feels like it's at the wrong abstraction layer here?

matklad:
@batiati do we want to get this into Monday's release? Can do a careful review if we want, otherwise I will look with Monday eyes, don't fully trust weekend matklad!

batiati:
> @batiati do we want to get this into Monday's release? Can do a careful review if we want, otherwise I will look with Monday eyes, don't fully trust weekend matklad!

Sure, there is no need to rush this fix! It's raining so much in here that I'm forced to stare at the computer!😬☔️
Enjoy your weekend!

src/message_bus.zig
```diff
@@ -454,8 +447,7 @@ fn MessageBusType(comptime process_type: vsr.ProcessType) type {
                 // The first replica's network address family determines the
                 // family for all other replicas:
                 const family = bus.configuration[0].any.family;
```
matklad:
```suggestion
                const family = bus.configuration[replica].any.family;
```

src/message_bus.zig
```diff
@@ -1,7 +1,6 @@
 const std = @import("std");
 const assert = std.debug.assert;
 const mem = std.mem;
-const posix = std.posix;
```
matklad, line 4: 🎉

src/io/linux.zig
```diff
@@ -1454,10 +1454,27 @@ pub const IO = struct {
     pub const socket_t = posix.socket_t;
     pub const INVALID_SOCKET = -1;
 
-    /// Creates a socket that can be used for async operations with the IO instance.
-    pub fn open_socket(self: *IO, family: u32, sock_type: u32, protocol: u32) !socket_t {
+    /// Creates a UDP socket that can be used for async operations with the IO instance.
```
matklad:
```suggestion
    /// Creates a TCP socket that can be used for async operations with the IO instance.
```

src/io/common.zig
```diff
@@ -7,16 +7,19 @@ const assert = std.debug.assert;
...
         keepidle: c_int,
         keepintvl: c_int,
         keepcnt: c_int,
-    },
-    user_timeout_ms: c_int,
-    nodelay: bool,
+    } = null,
+    user_timeout_ms: c_int = 0,
+    nodelay: bool = false,
+};
```
matklad: FWIW, originally I didn't add defaults there by design, as per our 🐯-style:

> Explicitly pass options to library functions at the call site, instead of relying on the defaults. For example, write `@prefetch(a, .{ .cache = .data, .rw = .read, .locality = 3 });` over `@prefetch(a, .{});`. This improves readability but most of all avoids latent, potentially catastrophic bugs in case the library ever changes its defaults.

For tests, we can have `count test_options: TCPOptions = ...` to avoid verbosity.

src/io/common.zig
```diff
@@ -25,72 +28,78 @@ pub fn listen(
...
-    var addrlen: posix.socklen_t = @sizeOf(std.net.Address);
-    try posix.getsockname(fd, &address_resolved.any, &addrlen);
-    assert(address_resolved.getOsSockLen() == addrlen);
-    assert(address_resolved.any.family == address.any.family);
-
-    try posix.listen(fd, options.backlog);
-
-    return address_resolved;
+pub fn setsockopt(_fd: posix.socket_t, level: i32, option: u32, value: c_int) !void {
+    try posix.setsockopt(_fd, level, option, &std.mem.toBytes(value));
```
matklad:
```suggestion
pub fn setsockopt(fd: posix.socket_t, level: i32, option: u32, value: c_int) !void {
    try posix.setsockopt(fd, level, option, &std.mem.toBytes(value));
```

src/io/linux.zig
```diff
@@ -19,6 +19,7 @@ const parse_dirty_semver = stdx.parse_dirty_semver;
 const maybe = stdx.maybe;
 
 pub const IO = struct {
+    pub const TCPOptions = common.TCPOptions;
```
matklad, line 22: 🤌

matklad (approved):

PR #2903, opened 2025-04-21
release-2025-04-21

batiati:


CHANGELOG.md
```diff
@@ -3,6 +3,46 @@
...
+
+  Fix a bug where VOPR latencies were computed incorrectly when the minimum
+  and mean were equal.
+
+### Internals
+
+- [#2893](https://github.com/tigerbeetle/tigerbeetle/pull/2893)
+
+  Make the `stdx.StackType(T)` intrusive pointer non-generic (`*StackLink`) over the
+  pointee type (`*T`), while keeping the overall API generic for type safety.
```
matklad: Let me take a second attempt at writing a useful commit message here: 

> Improve binary size and compilation time by making implementation of intrusive stack non-generic.

For commit messages, I tend to more "what" the summary line, as, contrary to the popular opinion, "why" tends to be less useful than "what" there. You usually look at commit messages to find a change that broke something, and there "what" is more useful as it has higher chance to capture unintended consequences. And, in context, "why" usually follows. 

For user-facing changelings, we should lean more on the why side.

CHANGELOG.md
```diff
@@ -3,6 +3,46 @@
...
+
+- [#2898](https://github.com/tigerbeetle/tigerbeetle/pull/2898)
+
+  Make the Docs website pass the W3C HTML validation test.
+
+- [#2897](https://github.com/tigerbeetle/tigerbeetle/pull/2897)
+
+  Rename the generic type `T` to `Int` in `stdx.PRNG` to better convey its purpose.
+  Also fix a few typos in the comments.
```
matklad: FWIW, I'd drop this from the changelog. We needn't include _everything_, only the things people can notice and thing "wow, that's cool", or where they can learn from our mistakes.

matklad (approved):

PR #2905, opened 2025-04-21
build: compile scripts for zig build ci

matklad:


matklad:
Changed my mind and moved it to be a part of `zig build ci` so that this is triggered _not_ only locally, but on CI as well.

matklad:
```
2025-04-21 13:31:50.643Z error: process failed with ExecTimeout: mvn --batch-mode --file pom.xml --quiet install
java ci: 14m14.782s
2025-04-21 13:31:50.644Z error: ExecTimeout
```

Huh?

batiati:
> ```
> 2025-04-21 13:31:50.643Z error: process failed with ExecTimeout: mvn --batch-mode --file pom.xml --quiet install
> java ci: 14m14.782s
> 2025-04-21 13:31:50.644Z error: ExecTimeout
> ```

Probable cause is the replica panicking during the integration tests.
Let me see how we can expose the logs, similar to what we did in https://github.com/tigerbeetle/tigerbeetle/pull/2749.

batiati (approved):

matklad:
@batiati  I am wondering, can we remove `private static class Server implements AutoCloseable {` altogether, and rely on a port to TIgerBEetle process passed as an env argument? Then, `TmpTigerBeetle` will be the only code that manages a running test cluster, and it includes the logic of dumping stderr on process crash:

https://github.com/tigerbeetle/tigerbeetle/blob/7800915823f5960d2e6ab67c2a9fafe28a2c6bb0/src/testing/tmp_tigerbeetle.zig#L190

matklad:
This make it more painful to run the tests _without_ going through `java/ci.zig`, but perhaps we can have `source ./scripts/activate-tigerbeetle.sh` 🤣 (no)

PR #2906, opened 2025-04-21
build: hide stderr for successful build zig ci

matklad:
Hide step's stderr unless it fails, to prevent zig build ci output being dominated by VOPR logs. Sadly, this requires a somewhat creative approach as far as I can tell.

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -446,7 +446,7 @@ fn build_ci(
     }
     if (all or mode == .aof) {
         const aof = b.addSystemCommand(&.{"./.github/ci/test_aof.sh"});
-        build_ci_check_status(aof);
+        hide_stderr(aof);
         step_ci.dependOn(&aof.step);
     }
     inline for (&.{ CIMode.dotnet, .go, .java, .node, .python }) |language| {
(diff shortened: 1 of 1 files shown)


build.zig
```diff
@@ -495,16 +497,32 @@ fn build_ci_script(
...
     run_artifact.setEnvironmentVariable("ZIG_EXE", b.graph.zig_exe);
-    build_ci_check_status(run_artifact);
+    hide_stderr(run_artifact);
     step_ci.dependOn(&run_artifact.step);
 }
 
-fn build_ci_check_status(step: *std.Build.Step.Run) void {
-    // NB: The purpose here is not to check status (Zig does that automatically), but rather to
-    // prevent inheriting stdio, which causes everything to execute sequentially due to stdio lock!
-    step.stdio = .{ .check = .{} };
```
cb22, line 505: Without this, would the run steps now be sequential?

matklad: No, they are still run in parallel. `.addCheck` already sets stdio to `.check`, so the explicit `step.stdio = .{ .check = .{} };` is simply unnecessary.

cb22 (approved):

PR #2907, opened 2025-04-21
Add basic monitoring docs and track request timings

cb22:
This PR adds some basic documentation on monitoring (including things like what you should look out for at a system level), and adds a new metric: `replica_request` which currently corresponds to the `commit_dispatch: slow request` messages.

~Originally, the plan was to make `replica_request` be end-to-end - so things like a view change would show as a latency spike. However, this is complicated by the fact that the header's prepare timestamp is set by `primary_pipeline_prepare`, which will underestimate latency significantly.~

`replica_request` is now as end-to-end as it can be; it's based off the message timestamp, which happens as soon as replica sees a message. However, this is subject to a bit of measurement error: if a message comes in while CPU compaction is running, it'll exclude that time, but if it comes in before it'll include it. This causes a divergence from what the client sees, because the client sees a reply before compaction starts.

Additionally, `replica_request_local` is tracked which is the local timer (exact same values as the log messages) and `replica_request_execute` which is the end-to-end timing until the execute stage, at which point the reply is sent to the client.

### DIFF
--- a/docs/operating/monitoring.md
+++ b/docs/operating/monitoring.md
@@ -1,8 +1,7 @@
 # Monitoring
 
-TigerBeetle supports emitting metrics via StatsD, with
-[DogStatsD extensions](https://docs.datadoghq.com/developers/dogstatsd/datagram_shell?tab=metrics)
-for tags.
+TigerBeetle supports emitting metrics via StatsD, and uses the
+[DogStatsD format for tags.](https://docs.datadoghq.com/developers/dogstatsd/datagram_shell?tab=metrics)
 
 This requires a StatsD compatible agent running locally. The Datadog Agent works out of the
 box with its default configuration, as does Telegraf's [statsd plugin](https://github.com/influxdata/telegraf/blob/master/plugins/inputs/statsd/README.md),
@@ -14,7 +13,9 @@ You can enable emitting metrics by adding the following CLI flags to each replic
 ```
 --experimental --statsd=127.0.0.1:8125
 ```
-(the `--statsd=` only accepts IP addresses at present.)
+
+The `--statsd` argument must be specified as an `IP:Port` address (IPv4 or IPv6).  DNS names are not
+currently supported.
 
 All TigerBeetle metrics are namespaced under `tb.` and are tagged with `cluster` (the cluster ID
 specified at format time) and `replica` (the replica index). Specific metrics might have additional
--- a/docs/operating/monitoring.md
+++ b/docs/operating/monitoring.md
@@ -1,0 +1,67 @@
+# Monitoring
+
+TigerBeetle supports emitting metrics via StatsD, with
+[DogStatsD extensions](https://docs.datadoghq.com/developers/dogstatsd/datagram_shell?tab=metrics)
+for tags.
+
+This requires a StatsD compatible agent running locally. The Datadog Agent works out of the
+box with its default configuration, as does Telegraf's [statsd plugin](https://github.com/influxdata/telegraf/blob/master/plugins/inputs/statsd/README.md),
+with `datadog_extensions` enabled.
+
+You can enable emitting metrics by adding the following CLI flags to each replica, depending on your
+[deployment method](./deploying/):
+
+```
+--experimental --statsd=127.0.0.1:8125
+```
+(the `--statsd=` only accepts IP addresses at present.)
+
+All TigerBeetle metrics are namespaced under `tb.` and are tagged with `cluster` (the cluster ID
+specified at format time) and `replica` (the replica index). Specific metrics might have additional
+tags - you can see a full list of metrics and cardinality by running `tigerbeetle inspect metrics`.
+
+## Specific Metrics
+
+### Overall status
+The `replica_status` metric corresponds to the overall status of the replica. If it's anything other
+than 0, it should be alerted on as it indicates a non-normal status. The full values are:
+
+| Value | Status          | Explanation                                                                                                                                    |
+|-------|-----------------|------------------------------------------------------------------------------------------------------------------------------------------------|
+| 0     | normal          | The replica is functioning normally.                                                                                                           |
+| 1     | view_change     | The replica is doing a view change.                                                                                                            |
+| 2     | recovering      | The replica is recovering. Usually, this will be present on startup before immediately transitioning to normal.                                |
+| 3     | recovering_head | The replica's persistent state is corrupted, and it can't participate in consensus. It will try and recover from the remainder of the cluster. |
+
+### State sync status
+The `replica_sync_stage` metric corresponds to the state sync stage. If this is anything other than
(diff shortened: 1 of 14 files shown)


src/stdx/bit_set.zig
```diff
@@ -2,15 +2,15 @@ const stdx = @import("../stdx.zig");
 const std = @import("std");
 const assert = std.debug.assert;
 
-/// Take a u8 to limit to 255 items max (2^8 - 1 = 255).
-/// Use dynamic bitset for larger sizes.
-pub fn BitSetType(comptime with_capacity: u8) type {
+/// Takes a u16, but the maximum size is bounded by the Word selection below.
```
cb22: cc @matklad since you wrote the original comment, I think we were wasting a bit by limiting our capacity type!

src/stdx/bit_set.zig
```diff
@@ -2,15 +2,15 @@ const stdx = @import("../stdx.zig");
...
+/// Use a dynamic bitset for larger sizes.
+pub fn BitSetType(comptime with_capacity: u16) type {
     return struct {
         // While mathematicaly 0 and 1 are symmetric, we intentionally bias the API to use zeros
         // default, as zero-initialization reduces binary size.
         bits: Word = 0,
 
-        pub const Word = for (.{ u8, u16, u32, u64, u128, u256 }) |w| {
+        pub const Word = for (.{ u8, u16, u32, u64, u128, u256, u512 }) |w| {
```
cb22: Maybe this is fine for now, but metrics is already on ~283 packets.

src/vsr/replica.zig
```diff
@@ -4710,6 +4714,26 @@ pub fn ReplicaType(
...
             }
 
+            // In most cases, this should be equal to the latency the client sees when sending a
+            // request, less the language client overhead, MessageBus queue and network latency.
+            //
+            // Only time operations when:
+            // * Running with the real state machine - as otherwise there's a circular dependency,
+            // * and when the replica's status is .normal - otherwise things like WAL replay at
+            //   startup will skew these numbers.
+            if (StateMachine.Operation == @import("../state_machine.zig").Operation_ and
```
cb22: This is pretty ugly, but I couldn't think of a nice way to resolve it! There's the same problem technically for tree_id too. Making Tracer depend on a concrete StateMachine would introduce a circular dependency there...

src/stdx/bit_set.zig
matklad: Let's rather do 

```diff
diff --git a/src/trace/statsd.zig b/src/trace/statsd.zig
index dab744118..1a2ab788c 100644
--- a/src/trace/statsd.zig
+++ b/src/trace/statsd.zig
@@ -119,7 +119,8 @@ pub const StatsD = struct {
     },
 
     send_buffer: *[packet_count_max * packet_size_max]u8,
-    send_completions: IOPSType(IO.Completion, packet_count_max) = .{},
+    send_completions: [packet_count_max]IO.Completion = .{},
+    send_in_flight_count: u32 = 0,
 
     /// Creates a statsd instance, which will send UDP packets via the IO instance provided.
     pub fn init_udp(
@@ -191,7 +192,7 @@ pub const StatsD = struct {
         //
         // Keep it as a log, rather than assert, to avoid the common pitfall of metrics killing
         // the whole system.
-        if (self.send_completions.executing() != 0) {
+        if (self.send_in_flight_count != 0) {
             log.err("{}: {} / {} packets still in flight; skipping emit", .{
                 self.replica,
                 self.send_completions.executing(),
@@ -264,11 +265,14 @@ pub const StatsD = struct {
 
         var send_offset: u32 = 0;
         for (send_sizes.const_slice()) |send_size| {
-            const completion = self.send_completions.acquire() orelse {
+            if (self.send_in_flight_count >= self.send_completions.len) {
                 // This shouldn't ever happen, but don't allow metrics to kill the system.
                 log.err("{}: insufficient packets to emit any metrics", .{self.replica});
                 return;
-            };
+            }
+
+            const completion = self.send_completions[self.send_in_flight_count];
+            self.send_in_flight_count += 1;
             self.emit_buffer(completion, self.send_buffer[send_offset..][0..send_size]);
             send_offset += send_size;
         }
@@ -300,7 +304,8 @@ pub const StatsD = struct {
             assert(self.implementation == .udp);
             self.implementation.udp.send_callback_error_count += 1;
         };
-        self.send_completions.release(completion);
+        self.send_in_flight_count -= 1;
+        completion.* = undefined;
     }
 };
 
```

I think `u256` is roughly the right boundary to force thinking about using a different data structure, and, for metrics, we don't really need IOPS!

docs/operating/monitoring.md
```diff
@@ -0,0 +1,67 @@
+# Monitoring
+
+TigerBeetle supports emitting metrics via StatsD, with
+[DogStatsD extensions](https://docs.datadoghq.com/developers/dogstatsd/datagram_shell?tab=metrics)
+for tags.
```
batiati: Can we rephrase this to make it true for any `StatsD`-compatible server, and mention Datadog's agent as an example of use case?  
Currently, it gives the impression that we only support metrics via DogStatsD extensions.

docs/operating/monitoring.md
```diff
@@ -0,0 +1,67 @@
...
+box with its default configuration, as does Telegraf's [statsd plugin](https://github.com/influxdata/telegraf/blob/master/plugins/inputs/statsd/README.md),
+with `datadog_extensions` enabled.
+
+You can enable emitting metrics by adding the following CLI flags to each replica, depending on your
+[deployment method](./deploying/):
+
+```
+--experimental --statsd=127.0.0.1:8125
+```
+(the `--statsd=` only accepts IP addresses at present.)
```
batiati: WDYT?

```suggestion
The `--statsd` argument must be specified as an `IP:Port` address (IPv4 or IPv6).  
DNS names are not currently supported.
```

src/state_machine.zig
```diff
@@ -89,17 +89,46 @@ pub const tree_ids = struct {
     };
 };
 
+// Looking to make backwards incompatible changes here? Make sure to check release.zig for
+// `release_triple_client_min`.
+pub const Operation_ = enum(u8) {
```
batiati: I don’t mind using the `_` suffix for re-exposed types. The issue here is that we’re accessing `Operation_`  from outside the StateMachine (in `event.zig`), and that feels wrong.

**Suggestion:** declare the `Operation` enum in `tigerbeetle.zig` and re-export it to the StateMachine. In `event.zig` we can use `tb.Operation`.

src/vsr/replica.zig
batiati: See comment above!

cb22:
@batiati ready for another round of review!

batiati (approved):

PR #2908, opened 2025-04-22
scripts: Better error on empty readme template sections

brson:
When adding the Rust client CI, I got stuck for a long time setting up the README template, where I initially created empty code sections to be copied into the readme.

For reference, every client has sample projects, and some of that content is copied into the client readme. I initially began the rust client sample by writing this:

```rust
// section:imports

fn main() {
    // endsection:imports
    todo!();                                                                                                                                             
    // section:client
    // endsection:client 
```

But empty sections are not allowed, producing an obscure assertion failure. This change makes the error more explicit when sections are empty by printing a explanatory warning and then panicking.

Hopefully this will make it easier for the next person who adds a client.

### DIFF
--- a/src/scripts/client_readmes.zig
+++ b/src/scripts/client_readmes.zig
@@ -7,6 +7,7 @@
 
 const std = @import("std");
 const assert = std.debug.assert;
+const log = std.log;
 
 const stdx = @import("../stdx.zig");
 const Shell = @import("../shell.zig");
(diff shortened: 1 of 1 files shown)


matklad (dismissed):
Thanks!

brson:
@matklad I had to fix a merge conflict and need another review

matklad (approved):

PR #2909, opened 2025-04-22
devhub: detect and highlight a stale canary

fabioarnold:
When a canary seed is older than 2 hours, we color it red and let it blink.

### DIFF
--- a/src/devhub/devhub.js
+++ b/src/devhub/devhub.js
@@ -3,7 +3,7 @@
 // At the moment, it isn't clear what's the right style for this kind of non-Zig developer facing
 // code, so the following is somewhat arbitrary:
 //
-// - camelCase naming
+// - snake_case naming
 // - `deno fmt` for style
 // - no TypeScript, no build step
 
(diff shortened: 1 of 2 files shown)


src/devhub/devhub.js
```diff
@@ -109,42 +109,53 @@ async function mainSeeds() {
...
+    if (canery_is_stale) seed_success = false;
+    if (seed_success) {
+      row_dom.classList.add("success");
     } else {
       seedFailCount++;
     }
     if (record.commit_sha != commit_previous) {
       commit_previous = record.commit_sha;
-      rowDom.classList.add("group-start");
+      row_dom.classList.add("group-start");
```
matklad, line 133: 👍 

Might rename others snake case while we are at it?

fabioarnold: Everything 🐍?

matklad: Yup!

matklad: Sorry for not getting casing consistently initially!

fabioarnold: No worries! I thought we would incrementally switch to snake case.

matklad (approved):

PR #2911, opened 2025-04-22
devhub: add lewisdaly to the release rotation

fabioarnold:
image

This week's release has been handled so this should be perfect timing.


lewisdaly (approved):
bring it on!

PR #2913, opened 2025-04-22
ci: don't rely on default token permissions

matklad:
As per TigerStyle,

> Explicitly pass options to library functions at the call site, instead
> of relying on the defaults.

This applies quadruply to automation permissions!

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,4 +1,5 @@
 name: CI
+permissions: {}
 
 concurrency:
   group: core-${{ github.event.pull_request.number || github.ref }}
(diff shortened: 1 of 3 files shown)


cb22:
```
The workflow is not valid. .github/workflows/ci.yml (Line: 2, Col: 1): Unexpected value 'permissionos'
```

cb22 (approved):

cb22:
(whoops, I was going to say: we do have permissions on this repo limited to read tokens by default in the UI too, but this can't hurt!)

matklad:
Yeah, github is very byzantine when it comes to how its configuration works, so I thought we might apply layering for defense in depth here.

PR #2914, opened 2025-04-22
ci: remove some dead code

matklad:
~~At some point we've lost them! I am not actually sure when, I thought that was~~

~~https://github.com/tigerbeetle/tigerbeetle/commit/1836a9231e64b6c89b32aec75cf8288c5626c3d2~~

~~But it looks like they weren'd runnnig at that poin already?~~

Just drop some dead code.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,7 +12,6 @@ on:
     branches: ["main"]
 
 env:
-  CI: 1 # build.zig enables extra sanity checks if CI is set.
   GH_TOKEN: ${{ github.token }}
 
 jobs:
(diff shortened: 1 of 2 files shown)


matklad:
Ah, never mind, I got confused by our build.zig logic, we _are_ running them all-right:

```
    if (b.args == null) {
        steps.@"test".dependOn(steps.test_integration);
        steps.@"test".dependOn(steps.test_fmt);
    }
```

matklad:
Ok, yeah, this is just some dead code then!

lewisdaly (approved):
Cool, looks good!

PR #2916, opened 2025-04-22
docs: extend the Automatic Batching section

lewisdaly:


docs/coding/requests.md
```diff
@@ -67,15 +67,23 @@ In the default configuration, the maximum batch sizes for each request type are:
...
 - [Go](/src/clients/go/README.md#batching)
 - [Java](/src/clients/java/README.md#batching)
 - [.NET](/src/clients/dotnet/README.md#batching)
 - [Python](/src/clients/python/README.md#batching)
 
+### Automatic Batching
+
+TigerBeetle clients automatically batch events. There may be instances where your application logic 
+makes it hard to fill up the batches that you send to TigerBeetle, for example a multi-threaded web 
+server where each HTTP request is handled on a different thread.
```
batiati: Wondering if we should say "batch operations" instead of "batch events".

One thing that we don't mention anywhere in the docs: besides the scenario where the application cannot fill up a `create_{accounts, transfers}` batch, there's another scenario where operations are not batchable! For example, `query_*` takes a single filter per call, but concurrent calls can be packed into the same request as long as the combined size of the `limit` field fits within a single reply. Here we are batching results instead of events.

docs/coding/requests.md
```diff
@@ -67,15 +67,23 @@ In the default configuration, the maximum batch sizes for each request type are:
...
 - [.NET](/src/clients/dotnet/README.md#batching)
 - [Python](/src/clients/python/README.md#batching)
 
+### Automatic Batching
+
+TigerBeetle clients automatically batch events. There may be instances where your application logic 
+makes it hard to fill up the batches that you send to TigerBeetle, for example a multi-threaded web 
+server where each HTTP request is handled on a different thread. 
+
+In these cases, if you share a TigerBeetle client across threads (or tasks, depending on your
```
batiati: _"In these cases, **if** you share a TigerBeetle client ..."_
Can we be more assertive that a single TigerBeetle client should always be the first option? I think we should recommend multiple clients only if they are targeting different clusters (so one client per cluster).

There are other scenarios where having multiple clients can be beneficial. For example, if one client is dedicated to queries while the other handles OLTP work, it could take advantage of the replica prepare pipeline if there are enough requests simultaneously. However, benchmarking the specific workload is always required to make an informed decision.

batiati (commented):
Could you also take a look at this paragraph in the client's documentation:

https://github.com/tigerbeetle/tigerbeetle/blob/b596034e66e37a98a8c6efdf2bc4c1773c0a49f4/src/clients/dotnet/docs.zig#L41-L46

docs/coding/requests.md
lewisdaly: Would you like me to add in this part as well? 

```
There are other scenarios where having multiple clients can be beneficial. For example, if one client is dedicated to queries while the other handles OLTP work, it could take advantage of the replica prepare pipeline if there are enough requests simultaneously. However, benchmarking the specific workload is always required to make an informed decision.
```

I quite like it, but I also don't want to take away from the main point of automatic batching.

docs/coding/requests.md
lewisdaly: Is there a difference between operations and events? I just see the term `Events` everywhere especially in this file, so would prefer to stick to this term. 

Re. non-batchable events/operations, the docs do have this in the table just above the section:

https://github.com/tigerbeetle/tigerbeetle/blob/693c988d820dd02f97551b1008794f2bacf11dc5/docs/coding/requests.md?plain=1#L57-L68

Maybe I could split this out and add more detail, something like:


```
### Request Batching

For the `lookup_{accounts, transfers}` and `create_{accounts, transfers}` request types, 
events should be batched together into a single request.

### Result Batching

For request types which support a maximum batch size (such as `query_accounts`) of 1, 
TigerBeetle doesn't allow batching of requests, _but_ does provide an optimization where 
concurrent calls can be packed into the same request as long as the combined size of the
`limit` field fits within a single reply.

### Automatic Batching
...
```

src/clients/dotnet/docs.zig
```diff
@@ -38,12 +38,7 @@ pub const DotnetDocs = Docs{
 
     .examples = "",
 
-    .client_object_documentation =
-    \\The `Client` class is thread-safe and for better performance, a
-    \\single instance should be shared between multiple concurrent
-    \\tasks. Multiple clients can be instantiated in case of connecting
-    \\to more than one TigerBeetle cluster.
-    ,
+    .client_object_documentation = "",
```
lewisdaly, line 41: @batiati I actually removed this section for the dotnet client since it was already covered in the paragraph just before.

docs/coding/requests.md
batiati: > Is there a difference between operations and events? 

Let’s try to stick to the jargon we have defined:

- `Event` is what we send, `Result` is what we get back, and `Operation` is the entire thing.

- `Batch` is an array of events/results that an operation sends/returns in a single call.  
For example, `create_transfers` can batch up to 8189 transfers.

- `Multi-batch` is an array of batches that the client can group together and send/receive in the same `Request`.

> Maybe I could split this out and add more detail

I don't think we should make this distinction. What matters is:

- For operations like `create_transfers`, the best practice is for the application to pack as many events as possible into a batch, regardless of whether the client can do automatic _multi-batching_ for concurrent requests.

- For operations like `query_transfers`, the best practice is for the application to define a sane `limit` rather than always using the maximum. It's faster and leaves room for _multi-batching_ concurrent queries together. (That’s why, unlike other filters where zero means “all,” `limit=0` returns nothing. We wanted to force the user to think about the limit!)

The selling point of automatic multi-batching is to optimize highly concurrent applications, such as the example of an HTTP server that handles one request per thread.

But from the application’s point of view, each individual thread shouldn’t have to care about automatic batching at all, as the best practices apply for all cases.

docs/coding/requests.md
batiati: > Would you like me to add in this part as well?

I think we can just save it for future use!

batiati (commented):
@lewisdaly, the overall content LGTM.
Would you care to take a fresh look before we merge it? After all this time, I'm not sure if it's the final version, and perhaps you might want to add/rephrase something.

lewisdaly:
> @lewisdaly, the overall content LGTM. Would you care to take a fresh look before we merge it? After all this time, I'm not sure if it's the final version, and perhaps you might want to add/rephrase something.

@batiati I just made a single change since I was using the word 'events' instead of 'operations'.

lewisdaly:
@batiati ready for a last check!

batiati (approved):

PR #2917, opened 2025-04-23
Change Data Capture (CDC)

batiati:
## Change Data Capture (CDC)

This PR enables CDC and implements a connector to AMQP (Advanced Message Queuing Protocol) targets, such as RabbitMQ and other compatible brokers.

It's composed of three distinct but related areas:

#### 1. TigerBeetle `get_events` operation.

Defines and implements the StateMachine `get_events` API (filter and results).
Until now, this operation was only a stub for testing purposes.

This PR does not expose `get_events` to `tb_client` or language clients. It remains an internal, undocumented operation,  
giving us room for eventual breaking changes.

#### 2. AMQP 0.9.1 Client.

> [AMQP](https://www.amqp.org/) is an open standard for passing business messages between applications or organizations.

We chose to implement the AMQP 0.9.1 specification instead of AMQP 1.0 for two main reasons: AMQP 0.9.1 is much simpler and widely supported (e.g., RabbitMQ only recently added native AMQP 1.0 support).
AMQP 1.0 is a good candidate for future support, though, given its interoperability across vendors.

The full AMQP specification was generated in Zig from the [XML file](https://github.com/rabbitmq/amqp-0.9.1-spec/blob/main/xml/amqp0-9-1.extended.xml). However, our API is very limited, supporting only a subset of methods required for CDC.

We use TigerBeetle’s IO abstraction to efficiently publish batches of messages from static buffers.

It is **not** a general-purpose AMQP client as there is no error handling.  
Instead, we crash on error, relying on the operator to log the issue and restart the CDC job.

Secure (AMQPS) connections over TLS are not supported, but this can be easily worked around using TLS tunnels.

#### 3. CDC Job.

The CDC job is part of the `tigerbeetle` executable, and can be started by command line `./tigerbeetle cdc [options]`.

It uses the `vsr.Client` directly (not `tb_client`) since it’s a single-threaded process that shares the same `*IO` instance with the AMQP client.

The CDC job has two components running in parallel and swapping buffers:
- **Producer**: Calls TigerBeetle's `get_events` operation.
- **Consumer**: Publishes events in the AMQP server.

The CDC job is **stateless** as TigerBeetle does not track which events have been published. Instead, the AMQP server is responsible for maintaining state in a persistent queue (the _progress tracker queue_) used to store the last timestamp acknowledged, so during startup, the CDC job consumes from this queue and resumes streaming from that point.

The CDC job also requires exclusive access to an internal queue (the *locker queue*), which prevents more than one instance of the CDC job from publishing events at the same time.

Messages are published and confirmed in batches using AMQP transactions (via the `tx-select` and `tx-commit` methods). However, AMQP transactions do not guarantee full atomicity in all cases! For example, if rejected due to quotas, partial batches might be delivered. TigerBeetle guarantees **at least once** semantics and makes a best-effort attempt to avoid publishing duplicate events, but it is the consumer's responsibility to perform idempotency checks.

TigerBeetle events (the transfer plus the affected debit and credit accounts, with their "as-of" balances) are encoded as UTF-8 JSON objects and sent as individual messages to the _exchange_. Each message also includes additional metadata (such as timestamp, ledger, and codes) that consumers can use for routing or filtering.

### Message content.

https://github.com/tigerbeetle/tigerbeetle/blob/batiati-cdc/docs/operating/cdc.md

Message metadata and custom headers:

| Key                   | AMQP data type     | Description                              |
|-----------------------|--------------------|------------------------------------------|
| `event_type`          | `string`           | Identifies the type of event.<br/>One of `single_phase`, `two_phase_pending`, `two_phase_posted`, `two_phase_voided` or `two_phase_expired`.                          |
| `ledger`              | `long_uint`        | The ledger of the transfer and accounts. |
| `transfer_code`       | `short_uint`       | The transfer code.                       |
| `debit_account_code`  | `short_uint`       | The debit account code.                  |
| `credit_account_code` | `short_uint`       | The credit account code.                 |
| `app_id`              | `string`           | Constant `tigerbeetle`.                  |
| `content_type`        | `string`           | Constant `application/json`              |
| `delivery_mode`       | `short_short_uint` | Constant `2` which means _persistent_.   |
| `timestamp`           | `timestamp`  | The event timestamp.<br>It's the transfer's timestamp, except when `event_type == 'two_phase_expired'` when it's the expiry timestamp.<br>The AMQP standard is in seconds, so TigerBeetle timestamps are truncated. Use the payload timestamp for full precision.<br>         |

Message payload:

The payload is encoded as JSON without line breaks or spaces.
Long integers such as `u128` and `u64` are encoded as JSON strings to improve interoperability.

Here is a formatted version for readability.

```json
{
  "timestamp": "1745328372758695656",
  "type": "single_phase",
  "ledger": 2,
  "transfer": {
    "id": 9082709,
    "amount": 3794,
    "pending_id": 0,
    "user_data_128": "79248595801719937611592367840129079151",
    "user_data_64": "13615171707598273871",
    "user_data_32": 3229992513,
    "timeout": 0,
    "code": 20295,
    "flags": 0,
    "timestamp": "1745328372758695656"
  },
  "debit_account": {
    "id": 3750,
    "debits_pending": 0,
    "debits_posted": 8463768,
    "credits_pending": 0,
    "credits_posted": 8861179,
    "user_data_128": "118966247877720884212341541320399553321",
    "user_data_64": "526432537153007844",
    "user_data_32": 4157247332,
    "code": 1,
    "flags": 0,
    "timestamp": "1745328270103398016"
  },
  "credit_account": {
    "id": 6765,
    "debits_pending": 0,
    "debits_posted": 8669204,
    "credits_pending": 0,
    "credits_posted": 8637251,
    "user_data_128": "43670023860556310170878798978091998141",
    "user_data_64": "12485093662256535374",
    "user_data_32": 1924162092,
    "code": 1,
    "flags": 0,
    "timestamp": "1745328270103401031"
  }
}
```

### Command line options.

Run the command `./tigerbeegle cdc` to start the CDC job.

```bash
./tigerbeetle amqp --addresses=3000 \
    --cluster=0 \
    --host=127.0.0.1 \
    --user=guest \
    --password=guest \
    --vhost=/ \
    --publish-exchange=tigerbeetle  
```
Complete list of arguments:

| Option                      | Description                                                                                                                              | Required                                                                                 |
|-----------------------------|------------------------------------------------------------------------------------------------------------------------------------------|------------------------------------------------------------------------------------------|
| `addresses`                 | TigerBeetle cluster addresses.                                                                                                           | Required                                                                                 |
| `cluster`                   | TigerBeetle cluster ID.                                                                                                                  | Required                                                                                 |
| `host`              | AMQP host address in the format `ip:port`. Both IPv4 and IPv6 addresses are supported. If `port` is omitted, the default is `5672`. DNS names are not supported. | Required                                                                                 |
| `user`                 | The AMQP username.                                                                                                                       | Required                                                                                 |
| `password`             | The AMQP password. Only PLAIN authentication is supported.                                                                              | Required                                                                                 |
| `vhost`                | The AMQP virtual host name.                                                                                                              | Required                                                                                 |
| `publish-exchange`     | The exchange name. Must be a pre-existing exchange provided by the operator.                                                            | Optional.<br>May be omitted if `amqp-publish-routing-key` is present.                   |
| `publish-routing-key`  | The routing key used in combination with the exchange.                                                                                   | Optional.<br>May be omitted if `amqp-publish-exchange` is present.                      |
| `event-count-max`           | Maximum number of events fetched from TigerBeetle and published to the AMQP server per batch.                                           | Optional.<br>Defaults to `2730` if omitted.                                              |
| `idle-interval-ms`          | Time interval in milliseconds to wait before querying again when the last query returned no events. | Optional.<br>Defaults to `1000` ms if omitted.                                           |
| `timestamp-last`            | Overrides the last timestamp stored in the progress tracker queue.<br>It's a TigerBeetle timestamp with precision in nanoseconds. | Optional.<br>If omitted, the timestamp is retrieved from the progress tracking queue.    |

### Testing with RabbitMQ.

For trying the CDC job, first start a RabbitMQ instance using `docker` (configure the `-v` path if you want to persist configurations).

```bash
docker run -it --rm \
    --name rabbitmq \
    -p 5672:5672 -p 15672:15672 \
    -v ./rabbitmq_data:/var/lib/rabbitmq/mnesia \
    -e RABBITMQ_NODENAME='rabbitmq@localhost' \
    rabbitmq:3-management
```

Then, using the web management interface at [http://localhost:15672/](http://localhost:15672/)  
(default username and password: `guest` / `guest`), create a queue and bind it to an exchange (the same name referenced by the `--amqp-publish-exchange` option) to receive the published messages.

Once the exchange and queue are set up, start the CDC job targeting an existing TigerBeetle cluster  
and this RabbitMQ server.


embano1:
🤩 disclosure: Tigerbeetle and CDC fan here, also deeply into eventing (Kafka, AWS EventBridge, CloudEvents, etc)

Briefly skimmed the proposal, was wondering what you think about using CloudEvents as the envelop to standardize your event payload and make it easier to integrate with other brokers and platforms natively supporting CloudEvents?

Let me know if you want to chat/discuss, happy to brainstorm on what this could look like. FWIW, Debezium also [supports](https://debezium.io/documentation/reference/stable/integrations/cloudevents.html) CE.

sentientwaffle (approved):

src/cdc/amqp/protocol.zig
```diff
@@ -0,0 +1,1260 @@
...
+    /// normal operations.
+    InternalError = spec.HARD_ERROR_INTERNAL_ERROR,
+
+    _,
+};
+
+pub const FieldValueTag = enum(u8) {
+    boolean = 't',
+    short_short_uint = 'B',
+    short_uint = 'u',
```
alvinyan-bond, line 169: Some of these fields are "unofficial" in the AMQP-0.9.1 spec, in particular `u` and `i`. Many programmatic MQ clients do not support these fields - for example I haven't found any libraries in Golang that support this at all. This appears to be a known issue where the core AMQP-0.9.1 grammar is agreed-upon, but peripheral extensions around alternate formats for integers are not, so many clients just haven't implemented them.

While not the fault of this PR, this unfortunately makes it very painful to consume these CDC messages. Is there a specific reason these peripheral types are needed, vs just using `l` to capture the superset?

alvinyan-bond: Started a thread in Slack for this - this seems to be an issue with the Golang RabbitMQ clients in particular, other clients like Python's Pika and Node's amqplib do support these fields. I'm not actually sure if they are considered "peripheral/nonstandard" fields or not...

batiati: Thanks @alvinyan-bond!
I did a complete review of the data types we use. For clarification, `u` and `i` are not _unofficial_ types. Unfortunately, they are just not supported by the RabbitMQ Go client. On the other hand, the `L` type (signed 64-bit integer) is indeed divergent, and both `l` and `L` are decoded as signed by RabbitMQ.
https://www.rabbitmq.com/amqp-0-9-1-errata#section_3

Fixed by https://github.com/tigerbeetle/tigerbeetle/pull/3019.

Davidc2525:
Cómo lo implemento con Kafka ?

PR #2918, opened 2025-04-23
stdx: de-genericify Queue

matklad:
We recently did this for a Stack, let's apply the same trick to Queue, for the same reason: smaller binary&faster compile times.

Binary size for aarch64 is reduced from 3 641 672 to 3 639 688, shaving off two KiBs!

### DIFF
--- a/src/clients/c/tb_client/context.zig
+++ b/src/clients/c/tb_client/context.zig
@@ -15,7 +15,6 @@ const MultiBatchDecoder = vsr.multi_batch.MultiBatchDecoder;
 const MultiBatchEncoder = vsr.multi_batch.MultiBatchEncoder;
 
 const IO = vsr.io.IO;
-const QueueType = vsr.queue.QueueType;
 const message_pool = vsr.message_pool;
 
 const MessagePool = message_pool.MessagePool;
@@ -176,8 +175,8 @@ pub fn ContextType(
         completion_context: usize,
 
         interface: *ClientInterface,
-        submitted: QueueType(Packet),
-        pending: QueueType(Packet),
+        submitted: Packet.Queue,
+        pending: Packet.Queue,
 
         signal: Signal,
         eviction_reason: ?vsr.Header.Eviction.Reason,
@@ -271,14 +270,14 @@ pub fn ContextType(
                 .deinit_fn = &vtable_deinit_fn,
             });
             context.interface = client_out;
-            context.submitted = .{
+            context.submitted = Packet.Queue.init(.{
                 .name = null,
                 .verify_push = builtin.is_test,
-            };
-            context.pending = .{
+            });
+            context.pending = Packet.Queue.init(.{
                 .name = null,
                 .verify_push = builtin.is_test,
-            };
+            });
             context.completion_context = completion_ctx;
             context.completion_callback = completion_callback;
             context.eviction_reason = null;
@@ -363,7 +362,7 @@ pub fn ContextType(
         /// Calls the user callback when a packet (the entire batched linked list of packets)
         /// is canceled due to the client being either evicted or shutdown.
         fn packet_cancel(self: *Context, packet_list: *Packet) void {
-            assert(packet_list.next == null);
+            assert(packet_list.link.next == null);
             assert(packet_list.phase != .complete);
             packet_list.assert_phase(packet_list.phase);
 
@@ -457,7 +456,7 @@ pub fn ContextType(
 
             // Nothing inflight means the packet should be submitted right now.
             if (self.client.request_inflight == null) {
-                assert(self.pending.count == 0);
+                assert(self.pending.count() == 0);
                 packet.phase = .pending;
                 packet.multi_batch_count = 1;
                 packet.multi_batch_event_count = @intCast(batch.event_count);
@@ -466,10 +465,9 @@ pub fn ContextType(
                 return;
             }
 
-            var it = self.pending.peek();
-            while (it) |root| {
+            var it = self.pending.iterate();
+            while (it.next()) |root| {
                 root.assert_phase(.pending);
-                it = root.next;
 
                 if (root.operation != packet.operation) continue;
 
@@ -625,7 +623,7 @@ pub fn ContextType(
 
             // Prevents IO thread starvation under heavy client load.
             // Process only the minimal number of packets for the next pending request.
-            const enqueued_count = self.pending.count;
+            const enqueued_count = self.pending.count();
             const safety_limit = 8 * 1024; // Avoid unbounded loop in case of invalid packets.
             for (0..safety_limit) |_| {
                 const packet: *Packet = pop: {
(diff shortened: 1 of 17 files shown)


src/queue.zig
```diff
@@ -3,126 +3,220 @@ const assert = std.debug.assert;
...
-            if (self.in) |in| {
-                in.next = elem;
-                self.in = elem;
-            } else {
-                assert(self.out == null);
-                self.in = elem;
-                self.out = elem;
-            }
-            self.count += 1;
+        pub inline fn pop(self: *Queue) ?*T {
```
matklad, line 33: This is just a hunch, but I suspect that _inline_ here can _reduce_ compilation time, as it should lead to less code arriving at LLVM.

src/queue.zig
```diff
@@ -3,126 +3,220 @@ const assert = std.debug.assert;
...
-        pub fn peek_last(self: Queue) ?*T {
-            return self.in;
+        pub inline fn peek(self: *const Queue) ?*T {
+            const link = self.any.peek() orelse return null;
+            return @alignCast(@fieldParentPtr("link", link));
         }
 
-        pub fn peek(self: Queue) ?*T {
-            return self.out;
+        pub fn count(self: *const Queue) u64 {
```
matklad, line 48: I feel that we should use u32 for the size, as, if you have more than 4 billion elements in the Queue, something is very wrong. But I leave this for future work.

src/clients/c/tb_client/context.zig
```diff
@@ -466,10 +465,9 @@ pub fn ContextType(
                 return;
             }
 
-            var it = self.pending.peek();
-            while (it) |root| {
+            var it = self.pending.iterate();
+            while (it.next()) |root| {
```
batiati, line 469: That's nice! :100:

src/clients/c/tb_client/packet.zig
```diff
@@ -131,7 +135,7 @@ pub const Packet = extern struct {
         assert(@sizeOf(Packet) == @sizeOf(Extern));
         assert(@alignOf(Packet) == @alignOf(Extern));
 
-        // Asseting the fields are identical.
+        // Asserting the fields are identical.
```
batiati, line 138: You redeemed your typos by catching this one :grimacing:

src/io/windows.zig
```diff
@@ -145,12 +145,10 @@ pub const IO = struct {
     fn flush_timeouts(self: *IO) ?u64 {
         var min_expires: ?u64 = null;
         var current_time: ?u64 = null;
-        var timeouts: ?*Completion = self.timeouts.peek();
 
         // iterate through the timeouts, returning min_expires at the end
```
batiati: Unrelated, just spotted this (and many other non-tigerstyle comments) in this file!
Feel free to ignore this if you don't want to mix things.

```suggestion
        // Iterate through the timeouts, returning min_expires at the end.
```

src/queue.zig
```diff
@@ -3,126 +3,220 @@ const assert = std.debug.assert;
 
 const constants = @import("./constants.zig");
 
+const QueueLink = extern struct {
+    next: ?*QueueLink = null,
+};
+
 /// An intrusive first in/first out linked list.
 /// The element type T must have a field called "next" of type ?*T
```
batiati:
```suggestion
/// The element type T must have a field called "link" of type `QueueLink`.
```

This same comment is repeated in `stack.zig` as well.

src/queue.zig
batiati: Zig’s rule of thumb for `inline fn` is either to propagate `comptime` arguments or because a benchmark showed it’s worth using. However, we often see `inline fn` used in Zig’s standard library for simple dispatches like this one!

So, I think we can consider it a good practice :+1:

src/testing/fuzz.zig
```diff
@@ -88,6 +88,29 @@ pub fn parse_seed(bytes: []const u8) u64 {
...
+// Like `std.meta.DeclEnum`, but allows excluding specific things. Feed the result into
+// random_enum_weights for swarm testing public API of a data structure.
+pub fn DeclEnumExcludingType(T: type, exclude: []const std.meta.DeclEnum(T)) type {
+    const base = @typeInfo(std.meta.DeclEnum(T)).Enum;
+    var fields_filtered: [base.fields.len - exclude.len]std.builtin.Type.EnumField = undefined;
+    var i: usize = 0;
+    next_field: for (base.fields) |field| {
+        for (exclude) |excluded| {
+            if (std.mem.eql(u8, field.name, @tagName(excluded))) continue :next_field;
+        }
```
batiati, line 102: TIL it's valid in Zig 0.13 :heart:

src/testing/fuzz.zig
```diff
@@ -88,6 +88,29 @@ pub fn parse_seed(bytes: []const u8) u64 {
     };
 }
 
+// Like `std.meta.DeclEnum`, but allows excluding specific things. Feed the result into
+// random_enum_weights for swarm testing public API of a data structure.
+pub fn DeclEnumExcludingType(T: type, exclude: []const std.meta.DeclEnum(T)) type {
+    const base = @typeInfo(std.meta.DeclEnum(T)).Enum;
+    var fields_filtered: [base.fields.len - exclude.len]std.builtin.Type.EnumField = undefined;
```
batiati, line 97:
```suggestion
   assert(exclude.len < base.fields.len);
    var fields_filtered: [base.fields.len - exclude.len]std.builtin.Type.EnumField = undefined;
```

src/vsr/grid_scrubber.zig
```diff
@@ -79,7 +79,7 @@ pub fn GridScrubberType(comptime Forest: type) type {
             done: bool,
 
             /// "next" belongs to the FIFOs.
```
batiati: Should we update the "next" in this comment?

src/testing/fuzz.zig
matklad: Why wouldn't it be? I think I am not using anything fancy here, no?

src/io/windows.zig
matklad: Will re-do comments in a separate PR!

src/vsr/grid_scrubber.zig
matklad: Definitely. We actually might not need these comments anymore, as now we spell the type of the thing the link belongs to, but I think extra clarity won't hurt!

src/testing/fuzz.zig
batiati: My intuition would tell me the outer `for` loop would start at `0` again! But probably I'm mixing it with the new labelled continue on `switch` introduced in Zig 0.14.

matklad: Yeah, no, that's just standard labeled continue:

https://docs.oracle.com/javase/tutorial/java/nutsandbolts/branch.html#:~:text=A%20labeled%20continue

I tried rewriting it as

```zig
    for (base.fields) |field| {
        const field_excluded = for (exclude) |excluded| {
            if (std.mem.eql(u8, field.name, @tagName(excluded))) break true;
        } else false;
        if (!field_excluded) {
            fields_filtered[i] = field;
            i += 1;
        }
    }
```

But I don't really love how we flip modality here.

matklad: Related post I always think about when writing loops:

https://www.teamten.com/lawrence/programming/avoid-continue.html

batiati (dismissed):

batiati (approved):

PR #2919, opened 2025-04-24
fix typos

matklad:
Used https://github.com/crate-ci/typos


batiati:
That is awesome!

One thing we should remove from TigerStyle then:

> Comments after the end of a line _can_ be phrases, with no punctuation.

https://github.com/tigerbeetle/tigerbeetle/blob/main/docs/TIGER_STYLE.md#L340-L341

matklad:
I think it's still fine for those to be phrases, like 

https://github.com/tigerbeetle/tigerbeetle/blob/b09829e38c44c8ab670bcc65998ec606e9a55dc5/src/vsr.zig#L1338

batiati (approved):

PR #2920, opened 2025-04-24
git-review: add a smoke test

matklad:
Let's get a bit more serious here and start writing tests!

Also, let's print all pending comments using gcc file:line: message format, so that you can use you editor to jump to a particular unresolved review comment. This is an improvemnt over just `git show HEAD`, as that doesn't make distinctin between resolved and unresolved comments.


batiati (approved):
Thanks for the opportunity to try the new git-review workflow!!

PR #2924, opened 2025-04-28
devhub: no more blinking, copy commands to clipboard

fabioarnold:
image

### DIFF
--- a/src/devhub/devhub.js
+++ b/src/devhub/devhub.js
@@ -56,8 +56,9 @@ async function main_seeds() {
     "https://api.github.com/repos/tigerbeetle/tigerbeetle/issues?per_page=200";
 
   const [records, issues] = await Promise.all([
-    (async () => await (await fetch(data_url)).json())(),
-    (async () => await (await fetch(issues_url)).json())(),
+    (async () => await (await fetch(data_url, { cache: "no-cache" })).json())(),
+    (async () =>
+      await (await fetch(issues_url, { cache: "no-cache" })).json())(),
   ]);
 
   const pulls = issues.filter((issue) => issue.pull_request);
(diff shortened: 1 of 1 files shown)


src/devhub/devhub.js
```diff
@@ -56,8 +56,9 @@ async function main_seeds() {
     "https://api.github.com/repos/tigerbeetle/tigerbeetle/issues?per_page=200";
 
   const [records, issues] = await Promise.all([
-    (async () => await (await fetch(data_url)).json())(),
-    (async () => await (await fetch(issues_url)).json())(),
+    (async () => await (await fetch(data_url, { cache: "no-cache" })).json())(),
+    (async () =>
+      await (await fetch(issues_url, { cache: "no-cache" })).json())(),
```
matklad: Let's introduce a `fetch_json` helper here! This pattern of `(async => await await)()` looks less fun now than when it was originally written 🤣 

I _think_ we should be able to do just

```
await Promise.all([
    fetch_json(data_url),
    fetch_json(issues_url), 
])
```

which looks much tamer.

matklad: Not sure about `no-cache` here, shouldn't it be controlled by the server-side cache-control headers here?

fabioarnold: This was in response to the issue dj had where Chrome would restore the page, re-execute the script with a current timestamp, but fetch an old data.json from the cache.

matklad: Did you confirm that this fixes issue for Chrome? My thinking here is that the _correct_ behavior from Chrome would be to re-execute the `fetch` depending on the cache policy. If it doesn't do that, than it doesn't honor cache policy. So even the no-cache could have no effect?

fabioarnold: It's super tricky to simulate the actual situation where the memory saver was involved. There's chrome://discards to manually discard tabs. But this doesn't work with DevTools attached (to observe network fetches).

But I can see that the fetch now always has a response with a `transferSize` > 0 (couple hundred bytes). So it always checks with the server if the file has changed.
For the fetch call in `main_metrics` which doesn't have this cache option set it's going directly to disk.

That makes me think the option does have an effect and it should work as intended.

matklad (approved):

PR #2926, opened 2025-04-28
git-review: add split subcommand

matklad:
The idea is that you can edit the code in-place during reviewing, than
add everything, comments and code changes, into the review commit, and
then call `git review split` to split the commit in two, one with code
changes and one with just the review comments.

The main benefit here is that this ordres code changes _before_ the
review, so any further history rewriting becomes easier.

In other words, in a typical workflow wheer you make code-changes as a
commint atop of the review commit, rebasing this commit to be before the
review would result in nasty merge conflicts, as you are likely to
modify adjacent lines. The `git-review split` does the right thing
there.

### DIFF
--- a/src/git-review.zig
+++ b/src/git-review.zig
@@ -54,6 +54,7 @@ fn log_fn(
 const CLIArgs = union(enum) {
     new,
     status,
+    split,
     lgtm,
     pub const help =
         \\Usage:
@@ -65,6 +66,9 @@ const CLIArgs = union(enum) {
         \\
         \\  git review status
         \\        Check validity of the current review.
+        \\
+        \\  git review split
+        \\        Splits mixed review commit into code changes and review proper.
         \\
         \\  git review lgtm
         \\        Assert that all comments are resolved, revert the review commit, push to remote.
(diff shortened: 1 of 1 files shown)


src/git-review.zig
```diff
@@ -194,6 +199,75 @@ fn review_lgtm(shell: *Shell) !void {
     try shell.exec("git push --force-with-lease", .{});
 }
 
+fn review_split(shell: *Shell) !void {
+    if (!try has_review(shell)) {
+        log.err("no review in progress", .{});
+        return error.NoReview;
+    }
+    const commit = try shell.exec_stdout("git rev-parse HEAD", .{});
```
sentientwaffle, line 207: Before we start making any changes to the git tree, can we print this commit hash to `log.info`? That way if there's some kind of problem with the script, you can always easily restore the original state.

Maybe even print something like `To undo the split, run "git reset ..."`.

matklad: Yeah, that's exactly how it works, see `errdefer` on line 223. Operations before that are read-only.

matklad: Oh, are you talking about printing this unconditionally? Yeah, that's probably better: even if the script _thinks_ it worked correctly, it might be buggy! I actually originally just had an info log there, let me restore it back.

sentientwaffle: Yeah, even if there isn't a zig error, we might still not like the result

sentientwaffle (approved):

PR #2929, opened 2025-05-02
Release 2025 05 02

matklad:


cb22 (approved):

docs/internals/releases.md
```diff
@@ -57,16 +57,24 @@ The motivation for specific steps follows after.
...
    ```
-6. Ask someone else to approve the GitHub workflow.
 
-7. Ping release manager for the next week in Slack.
+7. From this point on, the CFO will be fuzzing the release branch over the weekend.
+
+8. On Monday (different release manager!) check that there are no VOPR failures on the release
+   branch.
+
+9. Trigger the release workflow via GitHub web interface.
```
batiati, line 73: Can we move step 9 to Friday, so the next release manager is responsible for approving on Monday, so instead of 10 "ask someone" we would have "10 - Approve it"?

matklad: I originally planned to do that, but than switched to the current approach:

* I am not sure, but I think GitHub timeouts workflows which are without approval for too long?
* Having pending-deploy in the GitHub web-ui is not OCD friendly. I'd be _very_ tempted to click the button right away just so that I don't have to think about release in progress. 

No strong feelings either way though, feel free to remix the release process as you see fit, the docs are more of a suggestion/reminder!

matklad: And sorry for accidentally tagging for review! Which actually is a demonstration why having a button in the GitHub UI at all might lead to misclicks before Monday :P

PR #2937, opened 2025-05-05
git-review: add command to send review comments as a email

matklad:
git-review email --recepient somebody@example.com` will parse the
current state of review and send a plain-text email to the specified
address. Currently, this is mostly just scaffolding. I confirmd that it
can, in fact, send an email, but there's a bunch of things yet to do to
make it fully useful:

- setup a review bot email account (I've been using my personal account
  for testing)
- actually call `git review send` in GHA
- add a gitdb to skip already emailed comments
- add nicer formatting for the actual mail

But I want to move one step at a time here, and the first step is
crafting of the relevant CURL command.

### DIFF
--- a/src/git-review.zig
+++ b/src/git-review.zig
@@ -353,7 +353,7 @@ const Review = struct {
                 // Extract '380' from
                 // @@ -379,0 +380,2 @@
                 _, const added = stdx.cut(line, " +").?;
-                const hunk_line_str, _ = stdx.cut(added, ",") orelse stdx.cut(added, " ").?;
+                const hunk_line_str, _ = stdx.cut(added, ",") orelse stdx.cut(added, " @@").?;
                 hunk_line = std.fmt.parseInt(u32, hunk_line_str, 10) catch unreachable;
             }
 
--- a/src/git-review.zig
+++ b/src/git-review.zig
@@ -51,11 +51,19 @@ fn log_fn(
     std.debug.print(prefix ++ format ++ "\n", args);
 }
 
+const EmailOptions = struct {
+    server: []const u8,
+    auth: []const u8,
+    sender: []const u8,
+    recepient: []const u8,
+};
+
 const CLIArgs = union(enum) {
     new,
     status,
     split,
     lgtm,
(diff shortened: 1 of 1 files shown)


src/git-review.zig
```diff
@@ -389,8 +429,46 @@ const Review = struct {
...
+    try shell.exec_options(.{ .stdin_slice = email },
+        \\curl --ssl-reqd {server} --user {auth}
+        \\  --mail-from {sender} --mail-rcpt {recepient}
+        \\  --upload-file -
+    , .{
+        .server = options.server,
+        .auth = options.auth,
+        .sender = options.sender,
+        .recepient = options.recepient,
+    });
```
matklad, line 469: Open to suggestions for how to send email better, but just curling seems the optimal solution to me. This works with a gmail account using https://myaccount.google.com/apppasswords

src/git-review.zig
```diff
@@ -317,7 +353,7 @@ const Review = struct {
                 // Extract '380' from
                 // @@ -379,0 +380,2 @@
                 _, const added = stdx.cut(line, " +").?;
-                const hunk_line_str, _ = stdx.cut(added, ",").?;
+                const hunk_line_str, _ = stdx.cut(added, ",") orelse stdx.cut(added, " ").?;
```
batiati:
```suggestion
                const hunk_line_str, _ = stdx.cut(added, ",") orelse stdx.cut(added, " @@").?;
```

src/git-review.zig
batiati: Relying on curl is an elegant solution, the same way we rely on git!

But how does it work?
~Who will be responsible for triggering the email~ (and to which recipient addresses)?

> actually call git review send in GHA

Never mind

matklad: Yeah, this will be triggered by GHA, which would pass all EmailOptions as a secrets. As too what server we'll use, I think we should be able to use gmail, so it'll be review tigerbeetle.com email and `smtps://smtp.gmail.com:465` server.

batiati (approved):

PR #2941, opened 2025-05-06
Improve memory usage of tests 10x and test runtime 10x

matklad:
Best reviewed per commit. Found two low-hanging hippopotami in our tests!

### DIFF
--- a/src/testing/cluster.zig
+++ b/src/testing/cluster.zig
@@ -7,6 +7,7 @@ const log = std.log.scoped(.cluster);
 const stdx = @import("../stdx.zig");
 const constants = @import("../constants.zig");
 const message_pool = @import("../message_pool.zig");
+const ratio = stdx.PRNG.ratio;
 const MessagePool = message_pool.MessagePool;
 const Message = MessagePool.Message;
 
@@ -105,6 +106,7 @@ pub fn ClusterType(comptime StateMachineType: anytype) type {
         };
 
         allocator: mem.Allocator,
+        prng: stdx.PRNG,
         options: Options,
         callbacks: Callbacks,
 
@@ -355,6 +357,7 @@ pub fn ClusterType(comptime StateMachineType: anytype) type {
 
             cluster.* = Cluster{
                 .allocator = allocator,
+                .prng = prng,
                 .options = options.cluster,
                 .callbacks = options.callbacks,
                 .network = network,
@@ -486,8 +489,7 @@ pub fn ClusterType(comptime StateMachineType: anytype) type {
                 cluster.replicas,
                 cluster.replica_times,
                 cluster.replica_health,
-                0..,
-            ) |*storage, *replica, *time, *health, i| {
+            ) |*storage, *replica, *time, *health| {
                 if (health.* == .up and health.*.up.paused) {
                     // Tick the time even in a paused state, to simulate VM migration.
                     time.tick();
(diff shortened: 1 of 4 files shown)


src/testing/cluster.zig
```diff
@@ -499,7 +501,9 @@ pub fn ClusterType(comptime StateMachineType: anytype) type {
                             replica.tick();
 
                             // For performance, don't run every tick.
-                            if (i % 100 == 0) JournalChecker.check(replica);
+                            if (cluster.prng.chance(ratio(1, 100))) {
+                                JournalChecker.check(replica);
+                            }
```
matklad, line 506: This one feels like it'll also make VOPR massively faster.

sentientwaffle: :facepalm: Ahh excellent catch on this typo!

sentientwaffle (approved):

vimpostor:
@matklad In your [blog post](https://matklad.github.io/2025/05/06/performance-profile-visualization-challenge.html) you asked about better tooling to find this single-line hotspot in a matter of minutes instead of hours and to share so publicly.

> Please post your solutions somewhere publicly, so that we all can learn to profile more efficiently!

Well for these single-line hotspots it is more a matter of seconds instead of minutes, if you use the right tool.

For this I used [hotspot](https://github.com/KDAB/hotspot) but you can use any [perf](https://perfwiki.github.io/main/) frontend you like.
I started the Zig test case from Hotspot (leaving everything at the default options and you don't need to configure anything for this to work, debuginfod will even download missing debug symbols), and your line of code immediately lights up like a christmas tree:

Go to the "By File" tab, sort by "Cycles:Pu (incl.)", navigate to the first non-Zig-Toolchain file, and the line is visible directly at the top:
Screenshot_20250507_120714

It's also visible in the flamegraph, but this is the easiest way to find the line directly.

Hope this helps and happy profiling!

velaskuez:
Hi, I'm also here from your blog post. I found the line of code in xcode instruments using the cycles CPU counter. The default setting of sampling every 1M events wasn't granular enough to find the specific line, only the function, so I reduced it to 750K.
Screenshot 2025-05-11 at 10 47 19
Screenshot 2025-05-11 at 10 47 50

matklad:
@vimpostor congratulations on winning the challenge:

https://matklad.github.io/2025/05/19/profiling-challenge-results.html

I wasn't able to reproduce the result because my HotSpot lacks the By File tab, but, well, can't argue with a screenshot!

vimpostor:
> I wasn't able to reproduce the result because my HotSpot lacks the By File tab

Indeed that's a relatively new feature: https://github.com/KDAB/hotspot/issues/665
It's from last year, but seems it hasn't made it into a release yet.

> “Cycles:Pu” (who the heck is Pu?) is arcane

I agree the descriptions are not self-explanatory, in this case this refers to the event modifiers as per [man perf-list(1)](https://man.archlinux.org/man/perf-list.1.en#EVENT_MODIFIERS), so the `cycles` event is modified with `:Pu`:
- `P` - use maximum detected precise level
- `u` - user-space counting

It wouldn't be FOSS, if it didn't have ridiculous syntax (and granted, it doesn't help to find the info when the `perf` man page is split in 34 [sic!] different sub man-pages), but at least it's not important to grok it in this case - these are just the perf defaults after all. The important difference between the two columns is just the "inclusive cost" vs "self cost" in this case.

javierhonduco:
A bit late to the game, but these are the results with https://github.com/javierhonduco/lightswitch, with the pprof output and using the pprof CLI with `go tool pprof -http=:8080 -lines profile.pb`:

image

Using the default profile format (SVG flamegraph) with `sudo lightswitch --pids=$(pidof test) --flamegraph-aggregation=all`:

image

As an aside, this profiling challenge unearthed a couple of issues that had to be fixed before Tigerbeetle could be introspected with this profiler, details in https://github.com/javierhonduco/lightswitch/issues/264

PR #2948, opened 2025-05-07
vsr: primary sends start_view message on checkpoint durability

chaitanyabhandari:


src/config.zig
```diff
@@ -111,7 +111,7 @@ const ConfigProcess = struct {
     tcp_nodelay: bool = true,
     direct_io: bool,
     journal_iops_read_max: usize = 8,
-    journal_iops_write_max: usize = 16,
+    journal_iops_write_max: usize = 32,
```
matklad, line 114: I assume this is intentional?

matklad (approved):

src/config.zig
chaitanyabhandari: This is already part of the branch actually, so I'm not sure why it came up as a change in my PR: https://github.com/tigerbeetle/tigerbeetle/commit/8954d83a66c9c40ccfd76726db81d9a878703c3c

Perhaps I changed the order of the commits. 🤔

PR #2950, opened 2025-05-07
vsr: decouple prepare repair from header breaks

chaitanyabhandari:
In this PR, we decouple the repair of prepares with header breaks, such that repairing prepares _always_ iterates from  _commit_min + 1 → op_ , and then from _op_repair_min → op_, skipping missing headers and ignoring hash chain breaks.

One motivation for this change is simplicity. We needn't compute the header breaks for repairing prepares (which currently ensures that we are only requesting a portion of the log that is correctly hash chained). Instead, we can simply fire off `request_prepare` and rely on the fact that we have guardrails in `repair_header` to _reject_ wrong prepares (which a replica would only request for in the rare case where view change has truncated some prepares it is unaware of). Additionally, we have guardrails in our commit path to prevent committing a prepare unless our hash chain is valid all the way up to the head op.

This would also increase the efficacy of our repair. Currently, a cluster which misses higher and higher prepares would only ensure `request_prepare` is sent for prepares that are part of a perfect hash chain, and so is playing against a moving target (the part of the log with a perfect hash chain). Imperfect hash chains are rare, since they happen during view changes, so we don't need to design our repair around them (especially with other guardrails in place for protecting us against the log forking).

Finally, I was thinking that this simplification could help us maintain some state like `repair_last_requested_uncommitted_op`. We can then use this state cycle through ops between _commit_min + 1 → op_ when we make our repair more aggressive, as opposed to requesting the same prepares multiple times (leading to the quadratic repair problem we solved).

---

<details>
<summary>Extended Motivation</summary>
<br>
Our header repair & prepare repair are intertwined. First, we find the latest header break from _op_repair_min → op_. Then, we use  range.op_max + 1 as the prepare to start repairing from. As the cluster makes progress, if a replica misses/drops higher & higher ops, the latest header break would evaluate to a higher op number. This impacts the efficacy of our current repair protocol.
<pre>

            if (self.journal.dirty.count > 0) {
                // Request and repair any dirty or faulty prepares.
                const op_min = if (header_break) |range|
                    range.op_max + 1
                else
                    self.op_repair_min();
                self.repair_prepares(op_min);
            }

</pre>

Specifically, we do try to repair ops from _commit_min + 1 → op_, to make sure we make progress on commits stalled on missing prepares. But, if _commit_min < op_min_ (the latest header break) and op_min keeps getting reset to a larger op (for example; commit_min=5, and op_min moves from 10 → 15 → 20, etc), then we would actually keep attempting to repair from _op_min → op_ as opposed to _commit_min → op_!
<pre>

            const repair_ranges_exclusive: [2]struct { u64, u64 } = .{
                .{ @max(op_min, self.commit_min + 1), self.op + 1 },
                .{ op_min, @max(self.commit_min + 1, op_min) },
            }; 
</pre>
</details>

### DIFF
--- a/src/vsr/replica.zig
+++ b/src/vsr/replica.zig
@@ -6828,11 +6828,7 @@ pub fn ReplicaType(
 
             if (self.journal.dirty.count > 0) {
                 // Request and repair any dirty or faulty prepares.
-                const op_min = if (header_break) |range|
-                    range.op_max + 1
-                else
-                    self.op_repair_min();
-                self.repair_prepares(op_min);
+                self.repair_prepares();
             }
 
             if (header_break != null and header_break.?.op_max > self.op_checkpoint()) {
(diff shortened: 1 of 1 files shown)


src/vsr/replica.zig
```diff
@@ -7342,16 +7338,50 @@ pub fn ReplicaType(
...
+                            }
+                        }
+                    } else {
+                        assert(!self.journal.faulty.bit(slot));
+                    }
+                }
+            }
+        }
+
+        pub fn repair_prepares(self: *Replica) void {
```
sentientwaffle: No need for `pub`.

src/vsr/replica.zig
```diff
@@ -7342,16 +7338,50 @@ pub fn ReplicaType(
             }
         }
 
-        fn repair_prepares(self: *Replica, op_min: u64) void {
+        /// Attempt to repair prepares between [op_min, op_max], skipping over ops for which we
+        /// don't have a header.
```
```suggestion
        /// don't have a header and disregarding hash chain breaks.
```

src/vsr/replica.zig
```diff
@@ -7376,52 +7406,16 @@ pub fn ReplicaType(
...
-            if (budget == 0) {
-                log.debug("{}: repair_prepares: waiting for IOP", .{self.replica});
-                return;
-            }
-
-            // Iterate through op_min..=self.op, but make sure to iterate commit_min+1..=self.op
-            // sub range first:
             // - our first priority is to commit further,
             // - afterwards, repair committed prepares which are at risk of getting evicted from
             //   the journal, to help repair any lagging replicas.
```
sentientwaffle, line 7414: Maybe give this list a new introduction comment? or reformat it so that it makes sense

src/vsr/replica.zig
```diff
@@ -7342,16 +7338,50 @@ pub fn ReplicaType(
...
-        fn repair_prepares(self: *Replica, op_min: u64) void {
+        /// Attempt to repair prepares between [op_min, op_max], skipping over ops for which we
+        /// don't have a header.
+        fn repair_prepares_between(self: *Replica, op_min: u64, op_max: u64) void {
             assert(self.status == .normal or self.status == .view_change);
-            assert(self.op_repair_min() <= op_min);
+            assert(self.repairs_allowed());
             assert(op_min <= self.op);
+            assert(op_min >= self.op_repair_min());
+            assert(op_min <= op_max);
```
sentientwaffle, line 7348:
```suggestion
            assert(op_min <= op_max);
            assert(op_max <= self.op);
```

chaitanyabhandari:
@sentientwaffle Addressed all the comments in https://github.com/tigerbeetle/tigerbeetle/pull/2950/commits/a6eb1f02be9eb6fbec6b53dbe363e47e7c2f625c! Apologies for not pushing a new commit, I force pushed out of habit. 😕

sentientwaffle (approved):
No worries -- for a PR this size a single commit is just fine anyhow!

PR #2953, opened 2025-05-08
vsr: ensure that replies are padded to sector boundary

matklad:


src/message_bus.zig
```diff
@@ -825,15 +825,20 @@ fn MessageBusType(comptime process_type: vsr.ProcessType) type {
...
-                        assert(message.header.size < sector_ceil);
-                        assert(message.buffer.len == constants.message_size_max);
-                        @memset(message.buffer[message.header.size..sector_ceil], 0);
+                if (process_type == .replica) {
+                    // These messages are written to disk (.request converted to .prepare first) at
+                    // a sector granularity, and so should be padded with zeros for determinism.
+                    if (message.header.command == .request or
+                        message.header.command == .prepare or
+                        message.header.command == .block
+                    // or message.header.command == .reply
```
matklad, line 834: Ah, so this one doesn't fail in the VOPR, because we zero out stuff in ClientReplies: 

```
                // Zero sector padding to ensure deterministic storage.
                const size = message.header.size;
                const size_ceil = vsr.sector_ceil(size);
                @memset(message.buffer[size..size_ceil], 0);
```

I am a bit concerned that there are two places in the code-base where we do the zeroing, and its not clear to me which one is the correct one...

matklad:
superseded by #2992

PR #2954, opened 2025-05-08
[git-review] Fix the line number parser.

batiati:
The number of lines may not be shown if it's 1.


matklad:
Might I interest you in reviewing https://github.com/tigerbeetle/tigerbeetle/pull/2937 which also has this fixed with a test? ^^

batiati:
Fixed by #2937

PR #2956, opened 2025-05-08
vsr: tickless VOPR take 3, real time

matklad:
This is the third installment in the tickless VOPR series, whose overall
goal is to make sure that the VOPR can:

* simulate IO that completes faster than in a tick
* arbitrary permute io requests around each other

Historically, VOPR specified IO delay in terms of _ticks_, so the
"tickless" logic required writing a fancy priority queue which could
pick an arbitrary event, if many events are ready at the same tick.

But that's a very round-about way to solve the problem! A more direct
solution is to model latencies in nanoseconds! There are _many_
nanoseconds, so a _random_ timestamp now properly shuffles events!

A side benefit here is that, for things like performance VOPR, its now
much easier to map simulated latencies to real-world equivalents.


matklad:
Sorry for back and forth on this one @chaitanyabhandari, but I _think_ now I got it for sure --- it just makes sense that we should specify delays in terms of time!

src/testing/storage.zig
```diff
@@ -322,39 +327,37 @@ pub const Storage = struct {
...
+        const write_ready_at_ns =
+            if (storage.writes.peek()) |write| write.ready_at.ns else std.math.maxInt(u64);
+        if (read_ready_at_ns <= storage.tick_instant().ns and
+            read_ready_at_ns <= write_ready_at_ns)
+        {
+            const read = storage.reads.remove();
+            storage.read_sectors_finish(read);
+            advanced = true;
+        } else if (write_ready_at_ns <= storage.tick_instant().ns and
+            write_ready_at_ns <= write_ready_at_ns)
```
chaitanyabhandari: This should be `write_ready_at_ns <= read_ready_at_ns` instead!

src/vopr.zig
```diff
@@ -418,11 +419,11 @@ fn options_swarm(prng: *stdx.PRNG) Simulator.Options {
 
         .seed = prng.int(u64),
 
-        .one_way_delay_mean = prng.range_inclusive(u16, 3, 10),
-        .one_way_delay_min = prng.int_inclusive(u16, 3),
+        .one_way_delay_min = range_inclusive_ms(prng, 0, 30),
```
chaitanyabhandari, line 422: _Really_ like that we have just `range_inclusive_ms` now, as opposed to both `range_inclusive` and `int_inclusive`! Which I now realize we could've just used prng.range_inclusive(u16, 0, X) for. 😄

chaitanyabhandari:
Took one pass and things look great to me (sans that one tiny comment); really like what we landed up with here, modeling everything with `Instant` & `Duration` seems just right! Will take another pass tomorrow.

Oh, it also looks like there's a VOPR failure: `./zig/zig build -Drelease vopr -- 12995992096107620390`

src/vopr.zig
matklad: Yeah, I also played with the idea of teaching `range_inclusive` about `Duration`, and allso adding `ms` constructor to `Duration`, so the we can write 

```
prng.range_inclusive(ms(0), ms(30));
```

Probably a good idea, but figured we can let that simmer for a bit.

src/testing/storage.zig
matklad: 🎖️

matklad:
Failure seems unrelated, but also real:

```
/Users/matklad/p/tb/work/src/vsr/replica.zig:4640:19: 0x102acafdb in commit_stall (vopr)
            assert(commit_min_min <= self.commit_min);
                  ^
/Users/matklad/p/tb/work/src/vsr/replica.zig:4283:42: 0x102a82a1b in commit_dispatch (vopr)
                    if (self.commit_stall() == .pending) return;
```

Something for release manger to look into potentially!

matklad:
cc @sentientwaffle I am logging off for today, but 

```
git switch -d 820cbd09772daf5c84ecc7f05c175d88bc86c8df && \
  ./zig/zig build -Drelease vopr -- 12995992096107620390
```

feels like a real issue with https://github.com/tigerbeetle/tigerbeetle/pull/3050 (probably just over tight asertion).

chaitanyabhandari (approved):

PR #2957, opened 2025-05-08
vsr: prepare repair optimizations

chaitanyabhandari:
Best reviewed per commit! This PR implements various improvement to prepare repair.

**_First, it improves detection latency for a missing prepare._**

Currently, we only detect and attempt to repair prepares at repair timeout (which is 500ms), or if we are _already_ repairing the WAL (`self.repair` is invoked only when when we are already in the repair process). This can be disastrous in a loaded cluster, because even if a _single_ prepare is missed, we only detect it at repair timeout. This delayed detection also stalls the commit progress, exacerbating the lag on an already lagging replica.

We now invoke `self.repair` from some _non repair_ paths as well, to speed up the detection of missing prepares. Specifically, we now invoke `self.repair` during `on_prepare`, to pace repair with the rate on incoming prepares in the cluster.

Also, reduce `repair_timeout` from 500ms → 100ms.

**_Second, it improves throughput of repair._**

Our repair process maintains a repair budget to ensure a lagging replica doesn't flood the network with repair messages. However, while currently _decrement_ the repair budget every time we send a repair message (like `request_headers`, `request_prepares`), we only _replenish_ the repair budget at repair timeout. This is a bug that impacts repair throughput, because if a replica exhausts its repair budget, it must wait for ~500ms for it to get replenished.

The fix is simple, we increment the budget every time we receive a `header` or `prepare` that we _actually_ repair from (to avoid handling duplicate messages). 

**_Third, we now cycle through the missing prepares, instead of requesting for the same ones over and over again._**

_During_ a repair timeout interval, we now iterate through the missing prepares and ensure we only send a `request_prepare` for a missing prepare _once_. At repair timeout, we cycle around and start iterating from the beginning again!

### DIFF
--- a/src/vopr.zig
+++ b/src/vopr.zig
@@ -958,9 +958,15 @@ pub const Simulator = struct {
         for (simulator.cluster.replicas) |replica| {
             if (!simulator.core_repairable_replica(Cluster.Replica, &replica)) continue;
 
-            cluster_commit_max = @max(cluster_commit_max, replica.commit_max);
+            if (replica.log_view > cluster_log_view) {
+                maybe(cluster_op_head > replica.op);
+                cluster_op_head = replica.op;
+            } else if (replica.log_view == cluster_log_view) {
+                cluster_op_head = @max(cluster_op_head, replica.op);
+            }
+
             cluster_log_view = @max(cluster_log_view, replica.log_view);
-            cluster_op_head = @max(cluster_op_head, replica.op);
+            cluster_commit_max = @max(cluster_commit_max, replica.commit_max);
             cluster_op_repair_min = @min(cluster_op_repair_min, replica.op_repair_min());
         }
         assert(cluster_commit_max <= cluster_op_head);
--- a/src/vsr/replica.zig
+++ b/src/vsr/replica.zig
@@ -472,12 +472,12 @@ pub fn ReplicaType(
         repair_messages_budget: u32 = constants.vsr_repair_message_budget_max + 1,
 
         // The next op between [commit_min + 1, op] to be requested during repair.
-        // Used to cycle through uncommitted prepares, to avoid resending the same missing prepares
+        // Used to cycle through uncommitted prepares, to avoid requesting the same missing prepares
         // repeatedly. Reset to commit_min + 1 when repair timeout is fired.
         repair_prepare_op_next_uncommitted: u64 = 0,
 
         // The next op between [op_repair_min, commit_min] to be requested during repair.
-        // Used to cycle through committed prepares, to avoid resending the same missing prepares
+        // Used to cycle through committed prepares, to avoid requesting the same missing prepares
         // repeatedly. Reset to op_repair_min when repair timeout is fired.
         repair_prepare_op_next_committed: u64 = 0,
 
@@ -2146,10 +2146,14 @@ pub fn ReplicaType(
 
                 log.debug("{}: on_repair: repairing journal", .{self.replica});
                 self.write_prepare(message, .repair);
-                // Write prepare adds it synchronously to in-memory pipeline cache.
-                // Optimistically start committing without waiting for the disk write to finish.
+
                 if (self.status == .normal and self.backup()) {
+                    // Write prepare adds it synchronously to in-memory pipeline cache.
+                    // Optimistically start committing without waiting for the disk write to finish.
                     self.commit_journal();
+
+                    // Increment budget and initiate repair so `repair_header`/`request_prepare`
+                    // network messages can be sent concurrently with the write IO for this prepare.
                     self.repair_messages_budget = @min(
                         self.repair_messages_budget + 1,
                         constants.vsr_repair_message_budget_max + 1,
@@ -3423,8 +3427,17 @@ pub fn ReplicaType(
                 // +1 to make it easier to decrement budget with -|.
                 constants.vsr_repair_message_budget_max + 1,
             );
-            self.repair_prepare_op_next_committed = self.op_repair_min();
-            self.repair_prepare_op_next_uncommitted = self.commit_min + 1;
+
+            // We intentionally omit setting `repair_header_op_next` to `op_repair_min`. This avoids
+            // fetching extraneous headers from `op_repair_min → header break` every time the
+            // repair_timeout is fired, since we have already attempted to fetch these headers once!
+            // If we encounter a break before `repair_header_op_next`, (for example if replica
+            // we requested from is down or if its headers message was dropped), we simply request
+            // the missing headers, which is a smaller range to request.
+            {
+                self.repair_prepare_op_next_committed = self.op_repair_min();
+                self.repair_prepare_op_next_uncommitted = self.commit_min + 1;
+            }
 
             self.repair();
         }
--- a/src/vsr/replica.zig
+++ b/src/vsr/replica.zig
@@ -474,12 +474,17 @@ pub fn ReplicaType(
         // The next op between [commit_min + 1, op] to be requested during repair.
         // Used to cycle through uncommitted prepares, to avoid resending the same missing prepares
         // repeatedly. Reset to commit_min + 1 when repair timeout is fired.
-        repair_uncommitted_op_next: u64 = 0,
+        repair_prepare_op_next_uncommitted: u64 = 0,
 
-        // The next next op between [op_repair_min, commit_min] to be requested during repair.
+        // The next op between [op_repair_min, commit_min] to be requested during repair.
         // Used to cycle through committed prepares, to avoid resending the same missing prepares
         // repeatedly. Reset to op_repair_min when repair timeout is fired.
-        repair_committed_op_next: u64 = 0,
+        repair_prepare_op_next_committed: u64 = 0,
+
+        // The op number which should be set as op_min for the next request_headers call.
+        // This is an optimization that ensures op_min is not always set to op_repair_min (reducing
+        // the number of headers requested).
+        repair_header_op_next: u64 = 0,
 
         /// Whether the primary has received a quorum of do_view_change messages for the view
         /// change. Determines whether the primary may effect repairs according to the CTRL
@@ -3406,8 +3411,8 @@ pub fn ReplicaType(
                 // +1 to make it easier to decrement budget with -|.
                 constants.vsr_repair_message_budget_max + 1,
             );
-            self.repair_committed_op_next = self.op_repair_min();
-            self.repair_uncommitted_op_next = self.commit_min + 1;
+            self.repair_prepare_op_next_committed = self.op_repair_min();
+            self.repair_prepare_op_next_uncommitted = self.commit_min + 1;
 
             self.repair();
         }
@@ -6838,7 +6843,6 @@ pub fn ReplicaType(
                         self.op,
                     },
                 );
-
                 self.repair_messages_budget -|= 1;
                 if (self.repair_messages_budget > 0) {
                     self.send_header_to_replica(
@@ -6847,13 +6851,15 @@ pub fn ReplicaType(
                             .command = .request_headers,
                             .cluster = self.cluster,
                             .replica = self.replica,
-                            // Pessimistically request as many headers as we might need.
-                            // Requesting/sending extra headers is inexpensive, and it may save us
-                            // extra round-trips to repair earlier breaks.
-                            .op_min = self.op_repair_min(),
+                            // Pessimistically request extra headers. Requesting/sending extra
+                            // headers is inexpensive, and it may save us extra round-trips to
+                            // repair earlier breaks.
+                            .op_min = @min(range.op_min, self.repair_header_op_next),
                             .op_max = range.op_max,
                         }),
                     );
+
+                    self.repair_header_op_next = @max(self.repair_header_op_next, range.op_max + 1);
                 }
             }
 
@@ -7466,7 +7472,7 @@ pub fn ReplicaType(
             // - afterwards, repair committed prepares which are at risk of getting evicted from
             //   the journal, to help repair any lagging replicas.
             const range_min_uncommitted = @max(
-                self.repair_uncommitted_op_next,
+                self.repair_prepare_op_next_uncommitted,
                 self.commit_min + 1,
             );
             const range_max_uncommitted = self.op;
@@ -7477,14 +7483,14 @@ pub fn ReplicaType(
                 )) |op| {
                     assert(op >= self.commit_min + 1);
                     assert(op <= self.op);
-                    assert(op >= self.repair_uncommitted_op_next);
+                    assert(op >= self.repair_prepare_op_next_uncommitted);
 
-                    self.repair_uncommitted_op_next = op + 1;
+                    self.repair_prepare_op_next_uncommitted = op + 1;
                 }
             }
 
             const range_min_committed = @max(
-                self.repair_committed_op_next,
+                self.repair_prepare_op_next_committed,
                 self.op_repair_min(),
             );
             const range_max_committed = self.commit_min;
@@ -7492,9 +7498,9 @@ pub fn ReplicaType(
                 if (self.repair_prepares_between(range_min_committed, range_max_committed)) |op| {
                     assert(op >= self.op_repair_min());
                     assert(op <= range_max_committed);
-                    assert(op >= self.repair_committed_op_next);
+                    assert(op >= self.repair_prepare_op_next_committed);
 
-                    self.repair_committed_op_next = op + 1;
+                    self.repair_prepare_op_next_committed = op + 1;
                 }
             }
 
--- a/src/vsr/replica.zig
+++ b/src/vsr/replica.zig
@@ -2143,7 +2143,14 @@ pub fn ReplicaType(
                 self.write_prepare(message, .repair);
                 // Write prepare adds it synchronously to in-memory pipeline cache.
                 // Optimistically start committing without waiting for the disk write to finish.
-                if (self.status == .normal and self.backup()) self.commit_journal();
+                if (self.status == .normal and self.backup()) {
+                    self.commit_journal();
+                    self.repair_messages_budget = @min(
+                        self.repair_messages_budget + 1,
+                        constants.vsr_repair_message_budget_max + 1,
+                    );
+                    self.repair();
+                }
             }
         }
 
@@ -10262,16 +10269,7 @@ pub fn ReplicaType(
             self.flush_loopback_queue();
 
             switch (trigger) {
-                .append => {},
-                // Replenish budget and continue immediately to repair the next prepare:
-                // This is an optimization to eliminate waiting until the next repair timeout.
-                .repair => {
-                    self.repair_messages_budget = @min(
-                        self.repair_messages_budget + 1,
-                        constants.vsr_repair_message_budget_max + 1,
-                    );
-                    self.repair();
-                },
+                .append, .repair => {},
                 // Continue repairing the primary's pipeline.
                 .pipeline => self.repair(),
                 .fix => unreachable,
--- a/src/vsr/replica.zig
+++ b/src/vsr/replica.zig
@@ -7394,14 +7394,22 @@ pub fn ReplicaType(
                     if (self.journal.dirty.bit(slot)) {
                         // Rebroadcast outstanding `request_prepare` every `repair_timeout` tick.
                         // Continue to request prepares until our budget is depleted.
+                        self.repair_messages_budget -|= 1;
+                        if (self.repair_messages_budget == 0) {
+                            log.debug("{}: repair_prepares: repair budget used", .{
+                                self.replica,
+                            });
+                            break;
+                        }
+
                         if (self.repair_prepare(op)) {
                             requested_op_max = op;
                             budget -= 1;
                             if (budget == 0) {
-                                log.debug("{}: repair_prepares: request budget used", .{
+                                log.debug("{}: repair_prepares: IO budget used", .{
                                     self.replica,
                                 });
-                                return requested_op_max;
+                                break;
                             }
                         }
                     } else {
@@ -7469,8 +7477,12 @@ pub fn ReplicaType(
             }
 
             const range_min_committed = @max(
-                // +1 as repair_requested_op_max_committed tracks op for max requested prepare.
-                self.repair_requested_op_max_committed + 1,
+                if (self.repair_requested_op_max_committed == 0)
+                    // Ensure we start from op=0 so we can repair corrupt root op.
+                    0
+                else
+                    // +1 as repair_requested_op_max_committed tracks op for max requested prepare.
+                    self.repair_requested_op_max_committed + 1,
                 self.op_repair_min(),
             );
             const range_max_committed = self.commit_min;
@@ -7478,7 +7490,7 @@ pub fn ReplicaType(
                 if (self.repair_prepares_between(range_min_committed, range_max_committed)) |op| {
                     assert(op >= self.op_repair_min());
                     assert(op <= range_max_committed);
-                    assert(op > self.repair_requested_op_max_committed);
+                    assert(op >= self.repair_requested_op_max_committed);
 
                     self.repair_requested_op_max_committed = op;
                 }
@@ -7629,13 +7641,10 @@ pub fn ReplicaType(
                     reason,
                 });
 
-                self.repair_messages_budget -|= 1;
-                if (self.repair_messages_budget > 0) {
-                    self.send_header_to_replica(
-                        self.choose_any_other_replica(),
-                        @bitCast(request_prepare),
-                    );
-                }
+                self.send_header_to_replica(
+                    self.choose_any_other_replica(),
+                    @bitCast(request_prepare),
+                );
             }
 
             return true;
--- a/src/vsr/replica.zig
+++ b/src/vsr/replica.zig
@@ -1792,7 +1792,12 @@ pub fn ReplicaType(
                 self.advance_commit_max(message.header.commit, @src());
                 assert(self.commit_max >= message.header.commit);
             }
-            defer if (self.backup()) self.commit_journal();
+            defer {
+                if (self.backup()) {
+                    self.commit_journal();
+                    self.repair();
+                }
+            }
 
             if (message.header.op > self.commit_min + 2 * constants.pipeline_prepare_queue_max) {
                 log.warn("{}: on_prepare: lagging behind the cluster prepare.op={} " ++
--- a/src/vsr/replica.zig
+++ b/src/vsr/replica.zig
@@ -471,6 +471,16 @@ pub fn ReplicaType(
         // +1 to make it easier to decrement budget with -|.
         repair_messages_budget: u32 = constants.vsr_repair_message_budget_max + 1,
 
+        // The maximum prepare op between [commit_min + 1, op] requested during repair.
+        // Used to cycle through uncommitted prepares, to avoid resending the same missing prepares
+        // repeatedly. Reset to commit_min when repair timeout is fired.
+        repair_requested_op_max_uncommitted: u64 = 0,
+
+        // The maximum prepare op between [op_repair_min, commit_min] requested during repair.
+        // Used to cycle through committed prepares, to avoid resending the same missing prepares
+        // repeatedly. Reset to op_repair_min -| 1 when repair timeout is fired.
+        repair_requested_op_max_committed: u64 = 0,
+
         /// Whether the primary has received a quorum of do_view_change messages for the view
         /// change. Determines whether the primary may effect repairs according to the CTRL
         /// protocol.
@@ -3376,11 +3386,15 @@ pub fn ReplicaType(
         fn on_repair_timeout(self: *Replica) void {
             assert(self.status == .normal or self.status == .view_change);
             assert(self.repair_messages_budget <= constants.vsr_repair_message_budget_max + 1);
+
             self.repair_messages_budget = @min(
                 (self.repair_messages_budget + constants.vsr_repair_message_budget_refill),
                 // +1 to make it easier to decrement budget with -|.
                 constants.vsr_repair_message_budget_max + 1,
             );
+            self.repair_requested_op_max_committed = self.op_repair_min() -| 1;
+            self.repair_requested_op_max_uncommitted = self.commit_min;
+
             self.repair();
         }
 
@@ -7343,7 +7357,9 @@ pub fn ReplicaType(
 
         /// Attempt to repair prepares between [op_min, op_max], skipping over ops for which we
         /// don't have a header and disregarding hash chain breaks.
-        fn repair_prepares_between(self: *Replica, op_min: u64, op_max: u64) void {
+        ///
+        /// Returns maximum op for which request_prepare has been sent.
+        fn repair_prepares_between(self: *Replica, op_min: u64, op_max: u64) ?u64 {
             assert(self.status == .normal or self.status == .view_change);
             assert(self.repairs_allowed());
             assert(op_min <= self.op);
@@ -7355,21 +7371,23 @@ pub fn ReplicaType(
             var budget = self.journal.writes.available();
             if (budget == 0) {
                 log.debug("{}: repair_prepares: waiting for IOP", .{self.replica});
-                return;
+                return null;
             }
 
+            var requested_op_max: ?u64 = null;
             for (op_min..op_max + 1) |op| {
                 if (self.journal.slot_with_op(op)) |slot| {
                     if (self.journal.dirty.bit(slot)) {
                         // Rebroadcast outstanding `request_prepare` every `repair_timeout` tick.
                         // Continue to request prepares until our budget is depleted.
                         if (self.repair_prepare(op)) {
+                            requested_op_max = op;
                             budget -= 1;
                             if (budget == 0) {
                                 log.debug("{}: repair_prepares: request budget used", .{
                                     self.replica,
                                 });
-                                return;
+                                return requested_op_max;
                             }
                         }
                     } else {
@@ -7377,6 +7395,8 @@ pub fn ReplicaType(
                     }
                 }
             }
+
+            return requested_op_max;
         }
 
         fn repair_prepares(self: *Replica) void {
@@ -7415,12 +7435,39 @@ pub fn ReplicaType(
             // - our first priority is to commit further,
             // - afterwards, repair committed prepares which are at risk of getting evicted from
             //   the journal, to help repair any lagging replicas.
-            if (self.commit_min + 1 <= self.op) {
-                self.repair_prepares_between(self.commit_min + 1, self.op);
+            const range_min_uncommitted = @max(
+                // +1 as repair_requested_op_max_uncommitted tracks op for max requested prepare.
+                self.repair_requested_op_max_uncommitted + 1,
+                self.commit_min + 1,
+            );
+            const range_max_uncommitted = self.op;
+            if (range_min_uncommitted <= range_max_uncommitted) {
+                if (self.repair_prepares_between(
+                    range_min_uncommitted,
+                    range_max_uncommitted,
+                )) |op| {
+                    assert(op >= self.commit_min + 1);
+                    assert(op <= self.op);
+                    assert(op > self.repair_requested_op_max_uncommitted);
+
+                    self.repair_requested_op_max_uncommitted = op;
+                }
             }
 
-            if (self.op_repair_min() <= self.commit_min) {
-                self.repair_prepares_between(self.op_repair_min(), self.commit_min);
+            const range_min_committed = @max(
+                // +1 as repair_requested_op_max_committed tracks op for max requested prepare.
+                self.repair_requested_op_max_committed + 1,
+                self.op_repair_min(),
+            );
+            const range_max_committed = self.commit_min;
+            if (range_min_committed <= range_max_committed) {
+                if (self.repair_prepares_between(range_min_committed, range_max_committed)) |op| {
+                    assert(op >= self.op_repair_min());
+                    assert(op <= range_max_committed);
+                    assert(op > self.repair_requested_op_max_committed);
+
+                    self.repair_requested_op_max_committed = op;
+                }
             }
 
             // Clean up out-of-bounds dirty slots so repair() can finish.
--- a/src/vsr/replica.zig
+++ b/src/vsr/replica.zig
@@ -1279,7 +1279,7 @@ pub fn ReplicaType(
                 .repair_timeout = Timeout{
                     .name = "repair_timeout",
                     .id = replica_index,
-                    .after = 50,
+                    .after = 10,
                 },
                 .repair_sync_timeout = Timeout{
                     .name = "repair_sync_timeout",
--- a/src/vsr/replica.zig
+++ b/src/vsr/replica.zig
@@ -2872,12 +2872,26 @@ pub fn ReplicaType(
 
             var op_min: ?u64 = null;
             var op_max: ?u64 = null;
+            var header_repaired = false;
             for (message_body_as_prepare_headers(message.base_const())) |*h| {
                 if (op_min == null or h.op < op_min.?) op_min = h.op;
                 if (op_max == null or h.op > op_max.?) op_max = h.op;
-                _ = self.repair_header(h);
+                if (self.repair_header(h)) {
+                    header_repaired = true;
+                }
             }
             assert(op_max.? >= op_min.?);
+
+            // Ensure we only replenish the repair budget if we use this message to repair at least
+            // one header. This guards us against double incrementing the budget in case of
+            // duplicate headers messages.
+            if (header_repaired) {
+                self.repair_messages_budget = @min(
+                    self.repair_messages_budget + 1,
+                    // +1 to make it easier to decrement budget with -|.
+                    constants.vsr_repair_message_budget_max + 1,
+                );
+            }
 
             self.repair();
         }
@@ -10233,9 +10247,16 @@ pub fn ReplicaType(
 
             switch (trigger) {
                 .append => {},
-                // If this was a repair, continue immediately to repair the next prepare:
+                // Replenish budget and continue immediately to repair the next prepare:
                 // This is an optimization to eliminate waiting until the next repair timeout.
-                .repair => self.repair(),
+                .repair => {
+                    self.repair_messages_budget = @min(
+                        self.repair_messages_budget + 1,
+                        constants.vsr_repair_message_budget_max + 1,
+                    );
+                    self.repair();
+                },
+                // Continue repairing the primary's pipeline.
                 .pipeline => self.repair(),
                 .fix => unreachable,
             }
--- a/src/vsr/replica.zig
+++ b/src/vsr/replica.zig
@@ -471,15 +471,15 @@ pub fn ReplicaType(
         // +1 to make it easier to decrement budget with -|.
         repair_messages_budget: u32 = constants.vsr_repair_message_budget_max + 1,
 
-        // The maximum prepare op between [commit_min + 1, op] requested during repair.
+        // The next op between [commit_min + 1, op] to be requested during repair.
         // Used to cycle through uncommitted prepares, to avoid resending the same missing prepares
-        // repeatedly. Reset to commit_min when repair timeout is fired.
-        repair_requested_op_max_uncommitted: u64 = 0,
+        // repeatedly. Reset to commit_min + 1 when repair timeout is fired.
+        repair_uncommitted_op_next: u64 = 0,
 
-        // The maximum prepare op between [op_repair_min, commit_min] requested during repair.
+        // The next next op between [op_repair_min, commit_min] to be requested during repair.
         // Used to cycle through committed prepares, to avoid resending the same missing prepares
-        // repeatedly. Reset to op_repair_min -| 1 when repair timeout is fired.
-        repair_requested_op_max_committed: u64 = 0,
+        // repeatedly. Reset to op_repair_min when repair timeout is fired.
+        repair_committed_op_next: u64 = 0,
 
         /// Whether the primary has received a quorum of do_view_change messages for the view
         /// change. Determines whether the primary may effect repairs according to the CTRL
@@ -3406,8 +3406,8 @@ pub fn ReplicaType(
                 // +1 to make it easier to decrement budget with -|.
                 constants.vsr_repair_message_budget_max + 1,
             );
-            self.repair_requested_op_max_committed = self.op_repair_min() -| 1;
-            self.repair_requested_op_max_uncommitted = self.commit_min;
+            self.repair_committed_op_next = self.op_repair_min();
+            self.repair_uncommitted_op_next = self.commit_min + 1;
 
             self.repair();
         }
@@ -7382,9 +7382,16 @@ pub fn ReplicaType(
             assert(op_max <= self.op);
 
             // Request enough prepares to utilize our max IO depth:
-            var budget = self.journal.writes.available();
-            if (budget == 0) {
+            var io_budget = self.journal.writes.available();
+            if (io_budget == 0) {
                 log.debug("{}: repair_prepares: waiting for IOP", .{self.replica});
+                return null;
+            }
+
+            // Don't check the repair budget for a potential primary in view change, use the full
+            // write bandwidth for repair.
+            if (self.status == .normal and self.repair_messages_budget == 0) {
+                log.debug("{}: repair_prepares: waiting for repair budget", .{self.replica});
                 return null;
             }
 
@@ -7394,21 +7401,22 @@ pub fn ReplicaType(
                     if (self.journal.dirty.bit(slot)) {
                         // Rebroadcast outstanding `request_prepare` every `repair_timeout` tick.
                         // Continue to request prepares until our budget is depleted.
-                        self.repair_messages_budget -|= 1;
-                        if (self.repair_messages_budget == 0) {
-                            log.debug("{}: repair_prepares: repair budget used", .{
-                                self.replica,
-                            });
-                            break;
-                        }
-
                         if (self.repair_prepare(op)) {
                             requested_op_max = op;
-                            budget -= 1;
-                            if (budget == 0) {
-                                log.debug("{}: repair_prepares: IO budget used", .{
-                                    self.replica,
-                                });
+
+                            io_budget -= 1;
+                            if (self.status == .normal) {
+                                self.repair_messages_budget -= 1;
+                                if (self.repair_messages_budget == 0) {
+                                    log.debug("{}: repair_prepares: repair budget used", .{
+                                        self.replica,
+                                    });
+                                    break;
+                                }
+                            }
+
+                            if (io_budget == 0) {
+                                log.debug("{}: repair_prepares: IO budget used", .{self.replica});
                                 break;
                             }
                         }
@@ -7458,8 +7466,7 @@ pub fn ReplicaType(
             // - afterwards, repair committed prepares which are at risk of getting evicted from
             //   the journal, to help repair any lagging replicas.
             const range_min_uncommitted = @max(
-                // +1 as repair_requested_op_max_uncommitted tracks op for max requested prepare.
-                self.repair_requested_op_max_uncommitted + 1,
+                self.repair_uncommitted_op_next,
                 self.commit_min + 1,
             );
             const range_max_uncommitted = self.op;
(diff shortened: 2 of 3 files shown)


src/vsr/replica.zig
```diff
@@ -7355,28 +7371,32 @@ pub fn ReplicaType(
...
                         // Continue to request prepares until our budget is depleted.
                         if (self.repair_prepare(op)) {
+                            requested_op_max = op;
                             budget -= 1;
                             if (budget == 0) {
                                 log.debug("{}: repair_prepares: request budget used", .{
                                     self.replica,
                                 });
-                                return;
+                                return requested_op_max;
```
sentientwaffle: Maybe simpler to replace this with `break`?

src/vsr/replica.zig
```diff
@@ -7415,12 +7459,39 @@ pub fn ReplicaType(
...
+                }
             }
 
-            if (self.op_repair_min() <= self.commit_min) {
-                self.repair_prepares_between(self.op_repair_min(), self.commit_min);
+            const range_min_committed = @max(
+                // +1 as repair_requested_op_max_committed tracks op for max requested prepare.
+                self.repair_requested_op_max_committed + 1,
+                self.op_repair_min(),
+            );
```
sentientwaffle: Doesn't this make it impossible for the replica to repair its root prepare? (op=0)

src/vsr/replica.zig
```diff
@@ -7415,12 +7459,39 @@ pub fn ReplicaType(
...
+            if (range_min_uncommitted <= range_max_uncommitted) {
+                if (self.repair_prepares_between(
+                    range_min_uncommitted,
+                    range_max_uncommitted,
+                )) |op| {
+                    assert(op >= self.commit_min + 1);
+                    assert(op <= self.op);
+                    assert(op > self.repair_requested_op_max_uncommitted);
+
+                    self.repair_requested_op_max_uncommitted = op;
```
sentientwaffle: In `repair_prepare`:
```
                self.repair_messages_budget -|= 1;
                if (self.repair_messages_budget > 0) {
                    self.send_header_to_replica(
                        self.choose_any_other_replica(),
                        @bitCast(request_prepare),
                    );
                }
            }

            return true;
```
So even if repair is out of budget, we might still cycle/skip past some messages by updating our `repair_requested_op_max_uncommitted` when we have not actually sent a `request_prepare`.

Maybe we should instead push that budget check up into the caller (and change it to an assert in `repair_prepare`).. what do you think?

chaitanyabhandari:
Ran some experiments with three systems:

* Old Repair Protocol
* New Repair Protocol with `self.repair()` invoked on the commit path (https://github.com/tigerbeetle/tigerbeetle/pull/2957/commits/8dac08b3e4a555f1213e8fc6a4f7cc2c604e0e69)
* New Repair Protocol with `repair_timeout` changed from 500ms → 100 ms

Some high level insights:
* New Repair Protocol with `self.repair()` invoked on the commit path (https://github.com/tigerbeetle/tigerbeetle/pull/2957/commits/8dac08b3e4a555f1213e8fc6a4f7cc2c604e0e69) only _marginally_ increases performance (or it remains nearly the same). Surprisingly, for the 30% packet loss case, it even falls into state sync. 
* New repair protocol (in both cases) is more chatty over the network in all cases, which is understandable because it is more aggressive. Surprisingly, most of the increase is due to `headers`, which means we're requesting more & more headers in the new protocol. For instance, for the no faults case, the new repair protocol exchanges 2.34MiB  of headers and the old protocol exchanges  172.75KiB of headers. Presumably, this is because we actually request _all_ headers from op_repair_min → header break, to _reduce_ the number of metadata calls we make (an optimization we took up recently in https://github.com/tigerbeetle/tigerbeetle/pull/2628), and it like this is not playing well with more aggressive repair. @sentientwaffle what are your thoughts about making this a bit more pessimistic again, same as it was before your PR, given we call repair much more aggressively now?
* New Repair Protocol with `repair_timeout` changed from 500ms → 100 ms (which is the *simplest* change) performs better than the old repair protocol, and the  New Repair Protocol with `self.repair()` invoked on the commit path (https://github.com/tigerbeetle/tigerbeetle/pull/2957/commits/8dac08b3e4a555f1213e8fc6a4f7cc2c604e0e69) in all cases. For instance in the one replica down case, it takes _3259 ticks vs. 6041 ticks_ in the old repair protocol.

<details>
<summary>No faults, ./zig/zig build vopr -Drelease -- --performance 4380260082931739075</summary>

#### Old repair

```C
Messages:
prepare                  1765    1.40MiB
prepare_ok               1761    440.25KiB
ping                     750     375.00KiB
pong                     750     187.50KiB
commit                   340     85.00KiB
request                  271     267.55KiB
reply                    263     168.42KiB
start_view               91      238.00KiB
request_headers          51      12.75KiB
headers                  51      172.75KiB
do_view_change           30      15.00KiB
request_prepare          24      6.00KiB
ping_client              11      2.75KiB
pong_client              11      2.75KiB
request_start_view       1       256B
total                    6170    3.33MiB


          PASSED (2074 ticks)
```

#### New Repair with repair in commit path


```C
Messages:
prepare                  1962    1.61MiB
prepare_ok               1944    486.00KiB
ping                     750     375.00KiB
pong                     750     187.50KiB
request_headers          625     156.25KiB
headers                  625     2.34MiB
commit                   340     85.00KiB
request_prepare          298     74.50KiB
request                  274     280.00KiB
reply                    263     218.45KiB
start_view               102     268.75KiB
do_view_change           30      15.00KiB
ping_client              14      3.50KiB
pong_client              14      3.50KiB
request_start_view       7       1.75KiB
total                    7998    6.06MiB

[info] (simulator): 
          PASSED (2085 ticks)
```


#### New repair with 100ms repair timeout

```C
Messages:
prepare                  2075    1.53MiB
prepare_ok               2050    512.50KiB
request_headers          875     218.75KiB
headers                  875     3.32MiB
ping                     720     360.00KiB
pong                     720     180.00KiB
request_prepare          438     109.50KiB
commit                   325     81.25KiB
request                  275     249.69KiB
reply                    265     187.39KiB
start_view               93      243.75KiB
do_view_change           30      15.00KiB
ping_client              15      3.75KiB
pong_client              15      3.75KiB
request_start_view       3       768B
total                    8774    6.97MiB


          PASSED (1995 ticks)
```

</details>

<details>
<summary>5% packet loss,  ./zig/zig build vopr -Drelease -- --packet-loss-ratio=5/100 --performance 8243974837816383288</summary>



#### Old Repair

```C
Messages:
prepare                  1621    671.25KiB
prepare_ok               1210    302.50KiB
ping                     930     465.00KiB
pong                     908     227.00KiB
commit                   375     93.75KiB
request                  295     126.36KiB
reply                    280     117.91KiB
request_prepare          164     41.00KiB
request_blocks           96      31.38KiB
request_headers          87      21.75KiB
headers                  85      293.50KiB
start_view               85      223.50KiB
block                    71      89.45KiB
ping_client              59      14.75KiB
pong_client              58      14.50KiB
do_view_change           25      12.50KiB
request_start_view       10      2.50KiB
total                    6359    2.68MiB


          PASSED (2636 ticks)
```

#### New Repair with repair in commit path

```C
Messages:
prepare                  1533    631.23KiB
prepare_ok               1306    326.50KiB
ping                     900     450.00KiB
pong                     885     221.25KiB
commit                   365     91.25KiB
request_headers          359     89.75KiB
headers                  335     1.20MiB
request                  321     133.00KiB
reply                    276     116.20KiB
request_prepare          181     45.25KiB
block                    130     130.45KiB
request_blocks           100     32.38KiB
start_view               93      244.75KiB
ping_client              85      21.25KiB
pong_client              84      21.00KiB
do_view_change           25      12.50KiB
request_start_view       19      4.75KiB
request_reply            18      4.50KiB
total                    7015    3.71MiB


          PASSED (2548 ticks)
```

#### New Repair with 100ms repair timeout

```C
Messages:
prepare                  1720    714.56KiB
prepare_ok               1656    414.00KiB
ping                     840     420.00KiB
pong                     821     205.25KiB
request_headers          707     176.75KiB
headers                  690     2.56MiB
request_prepare          411     102.75KiB
commit                   350     87.50KiB
request                  298     127.45KiB
reply                    277     127.71KiB
start_view               84      220.00KiB
ping_client              38      9.50KiB
pong_client              38      9.50KiB
do_view_change           25      12.50KiB
request_start_view       10      2.50KiB
total                    7965    5.13MiB


          PASSED (2361 ticks)
```
</details>



<details>
<summary>30% packet loss, ./zig/zig build vopr -Drelease -- --packet-loss-ratio=30/100 --performance 1589881801411644986</summary>

#### Old Repair


```C
Messages:
prepare                  2606    1.86MiB
ping                     2155    1.05MiB
prepare_ok               1950    487.50KiB
pong                     1778    444.50KiB
request_prepare          886     221.50KiB
request                  836     704.88KiB
commit                   810     202.50KiB
ping_client              624     156.00KiB
request_headers          574     143.50KiB
pong_client              513     128.25KiB
headers                  454     1.65MiB
reply                    350     203.12KiB
start_view               144     383.75KiB
request_start_view       58      14.50KiB
do_view_change           45      22.50KiB
request_blocks           44      14.13KiB
block                    43      104.22KiB

total                    13870   7.72MiB

          PASSED (6771 ticks)
```

#### New Repair with repair in commit path

```C
Messages:
ping                     4610    2.25MiB
pong                     3745    936.25KiB
block                    3122    6.39MiB
prepare                  2695    1.89MiB
request_blocks           1974    677.00KiB
prepare_ok               1718    429.50KiB
commit                   1570    392.50KiB
request                  1290    1.08MiB
ping_client              1150    287.50KiB
pong_client              951     237.75KiB
request_prepare          938     234.50KiB
request_headers          929     232.25KiB
headers                  762     2.74MiB
request_reply            405     101.25KiB
reply                    371     206.06KiB
start_view               320     876.25KiB
request_start_view       291     72.75KiB
do_view_change           75      67.50KiB
start_view_change        25      6.25KiB
total                    26941   18.99MiB

[info] (simulator):
          PASSED (14950 ticks)
```
#### New Repair with 100ms repair_timeout

```C
Messages:
prepare                  2751    2.02MiB
prepare_ok               2172    543.00KiB
ping                     1670    835.00KiB
pong                     1351    337.75KiB
request_prepare          1325    331.25KiB
request_headers          1302    325.50KiB
headers                  1055    3.97MiB
commit                   645     161.25KiB
request                  637     553.27KiB
ping_client              401     100.25KiB
reply                    356     301.02KiB
pong_client              339     84.75KiB
start_view               119     316.25KiB
do_view_change           40      20.00KiB
request_start_view       37      9.25KiB
total                    14200   9.82MiB


          PASSED (5131 ticks)
```

</details>

<details>
<summary>R=3 down, ./zig/zig build vopr -Drelease -- --replica-missing=3 --performance 5036667415013174002</summary>

#### Old Repair


```C
Messages:
prepare                  1795    1.57MiB
ping                     1625    812.50KiB
pong                     1300    325.00KiB
prepare_ok               1250    312.50KiB
request                  793     857.41KiB
commit                   730     182.50KiB
ping_client              581     145.25KiB
pong_client              421     105.25KiB
reply                    342     260.01KiB
request_headers          300     75.00KiB
request_prepare          281     70.25KiB
headers                  230     839.50KiB
block                    184     352.87KiB
request_blocks           160     51.31KiB
start_view               129     347.00KiB
request_start_view       39      9.75KiB
do_view_change           25      12.50KiB
total                    10185   6.21MiB


          PASSED (6041 ticks)
```

#### New Repair with repair in commit path

```C
prepare                  1731    1.41MiB
ping                     1575    787.50KiB
pong                     1259    314.75KiB
prepare_ok               1057    264.25KiB
request                  792     818.16KiB
commit                   715     178.75KiB
ping_client              580     145.00KiB
request_headers          456     114.00KiB
pong_client              430     107.50KiB
headers                  372     1.33MiB
reply                    347     255.20KiB
request_blocks           235     75.44KiB
request_prepare          225     56.25KiB
block                    181     336.09KiB
start_view               109     290.50KiB
request_reply            29      7.25KiB
do_view_change           25      12.50KiB
request_start_view       19      4.75KiB
total                    10137   6.43MiB

[info] (simulator): 
          PASSED (5888 ticks)
```

#### New repair with 100ms repair timeout

```C
Messages:
prepare                  2047    1.65MiB
prepare_ok               1679    419.75KiB
request_headers          1124    281.00KiB
ping                     925     462.50KiB
headers                  880     3.33MiB
request_prepare          869     217.25KiB
pong                     740     185.00KiB
commit                   455     113.75KiB
request                  323     314.63KiB
reply                    289     205.59KiB
start_view               90      236.25KiB
ping_client              87      21.75KiB
pong_client              72      18.00KiB
do_view_change           25      12.50KiB
total                    9605    7.41MiB
          PASSED (3259 ticks)
```

</details>

src/vsr/replica.zig
chaitanyabhandari: Ah, awesome catch! My understanding was that `repair_prepare` should return `False` in those cases and that we're returning `True` *only* when we're sending a `request_prepare`, but we don't! So another solution could also be just returning `False` there?
```C
                if (self.repair_messages_budget > 0) {
                    self.send_header_to_replica(
                        self.choose_any_other_replica(),
                        @bitCast(request_prepare),
                    );
                } else return false;
```

sentientwaffle: We could, though in that case we would loop through all the prepares even when the budget is exhausted and we could otherwise break early.

chaitanyabhandari: Ah yes, indeed! Your suggestion is spot on, I just hadn't wrapped my head around it completely. :)

src/vsr/replica.zig
chaitanyabhandari: Ah, _excellent_ catch, indeed it does! I have actually forgot to account for root prepare once more in the past, I will make sure to keep it in mind now. Also, I am very curious to see how this made it past `Cluster: recovery: WAL prepare corruption (R=3, corrupt root)"`. Let me investigate... and also come up with a better solution.

chaitanyabhandari: Mhmmmm `Cluster: recovery: WAL prepare corruption (R=3, corrupt root)` only checks if the _cluster_ makes progress when one of the replicas starts up with a corrupt root prepare. I will add a check there for whether the faulty bit for that sector is _actually_ unset or not.

src/vsr/replica.zig
```diff
@@ -7404,14 +7404,22 @@ pub fn ReplicaType(
                     if (self.journal.dirty.bit(slot)) {
                         // Rebroadcast outstanding `request_prepare` every `repair_timeout` tick.
                         // Continue to request prepares until our budget is depleted.
+                        self.repair_messages_budget -|= 1;
+                        if (self.repair_messages_budget == 0) {
+                            log.debug("{}: repair_prepares: repair budget used", .{
+                                self.replica,
+                            });
+                            break;
```
sentientwaffle: I think we shouldn't decrement the repair budget if `repair_prepare` returns false.

src/vsr/replica.zig
```diff
@@ -7479,16 +7487,20 @@ pub fn ReplicaType(
...
 
             const range_min_committed = @max(
-                // +1 as repair_requested_op_max_committed tracks op for max requested prepare.
-                self.repair_requested_op_max_committed + 1,
+                if (self.repair_requested_op_max_committed == 0)
+                    // Ensure we start from op=0 so we can repair corrupt root op.
+                    0
+                else
+                    // +1 as repair_requested_op_max_committed tracks op for max requested prepare.
+                    self.repair_requested_op_max_committed + 1,
```
sentientwaffle: This is a little awkward... would it be clearer to make these fields optional and then initialize them to null?

src/vsr/replica.zig
```diff
@@ -7404,14 +7404,22 @@ pub fn ReplicaType(
                     if (self.journal.dirty.bit(slot)) {
                         // Rebroadcast outstanding `request_prepare` every `repair_timeout` tick.
                         // Continue to request prepares until our budget is depleted.
+                        self.repair_messages_budget -|= 1;
+                        if (self.repair_messages_budget == 0) {
```
sentientwaffle: For potential primaries I think we don't want to check the budget. (Or at least that is how it works right now.)

src/vsr/replica.zig
chaitanyabhandari: Also not a fan of this logic, I had tried out a couple of different ways to organize this condition and this seemed to be best out of the worst... let me brainstorm a bit more to see how it can be made cleaner.

src/vsr/replica.zig
chaitanyabhandari: I was thinking, since we're incrementing the budget now when we receive this prepare, perhaps makes sense to decrement it as well?

Also, I'm not quite clear on why skipping the budget here helps, we _would_ be bottlenecked by the write IO anyway, no?

sentientwaffle: > I was thinking, since we're incrementing the budget now when we receive this prepare, perhaps makes sense to decrement it as well?

Ah that's a good point! Though the flip side is that if there is a potential primary, then the cluster is doing a view change, which means it is stuck in not making any other progress. So we want to get healthy and committing as soon as possible. Which means that the normal repair budget might not be enough in that case.

src/vsr/replica.zig
chaitanyabhandari: Ah, _of course_ you're right! These budget calculations have been confusing me somehow, I think I need a coffee stat. ☕

src/vsr/replica.zig
chaitanyabhandari: Mhmmmmm, I _think_ it also made less sense before because repair budget was 12 and the # of write IOPs were 8 so we'd be bottlenecked by the write IOPs anyway. Now they are 16, so we might as well use all the bandwidth that we have to repair during view change...

That reminds me, shall I also push a change to increase them to 32 as part of this PR? Since we've done that for OKTO on their custom branch anyway?

src/vsr/replica.zig
chaitanyabhandari: Changed it to track `repair_uncommitted_op_next` and `repair_committed_op_next` instead, in https://github.com/tigerbeetle/tigerbeetle/pull/2957/commits/ba7e3468ba8849c6aa23dd705122d75d151e5e0b. That seems cleaner to me and trivially handles the root prepare case!

src/vsr/replica.zig
```diff
@@ -7392,31 +7392,37 @@ pub fn ReplicaType(
...
-                            if (budget == 0) {
-                                log.debug("{}: repair_prepares: IO budget used", .{
+
+                            io_budget -= 1;
+                            if (io_budget == 0) {
+                                log.debug("{}: repair_prepares: IO budget used", .{self.replica});
+                                break;
+                            }
+
+                            self.repair_messages_budget -= @intFromBool(self.status == .normal);
```
sentientwaffle: I think we need to decrement `repair_messages_budget` before we check whether `io_budget` is zero.

sentientwaffle: Also if `repair_messages_budget == 0 and status == .view_change` (_before_ decrementing) then we will break even when we should continue.

chaitanyabhandari:
@sentientwaffle I re-ran the experiments with non-alternating replication: 
* The total size of messages sent is *smaller* than their counterpart with alternating replication, *but* still slightly chattier than the current replication. For example 4.61MiB vs. 3.33MiB for the no faults case, 3.44MiB vs. 2.68MiB in the 5% packet loss case. 
* For 30% packet loss case, 6.85MiB (New repair with 100ms repair timeout) vs. 7.72MiB (Old repair)
* *Surprisingly*, our new repair protocol _still_ goes into state sync in the 30% packet loss case, for reasons that aren't obviously clear to me. I'm going to look through the logs to figure out what the issue might be here, it is actually quite counterintuitive to me that we're repairing more aggressively but we instead lag *more*!

<details>
<summary>No faults, ./zig/zig build vopr -Drelease -- --performance 4380260082931739075</summary>

#### New repair with non-alternating replication

```C
Messages:
prepare                  1818    1.30MiB
prepare_ok               1808    452.00KiB
ping                     750     375.00KiB
pong                     749     187.25KiB
request_headers          375     93.75KiB
headers                  374     1.35MiB
commit                   325     81.25KiB
request                  273     229.31KiB
request_prepare          264     66.00KiB
reply                    262     186.63KiB
start_view               113     301.00KiB
do_view_change           30      15.00KiB
request_start_view       28      7.00KiB
ping_client              13      3.25KiB
pong_client              13      3.25KiB
total                    7195    4.61MiB


          PASSED (2024 ticks)
```

#### New repair with 100ms repair timeout and non-alternating replication

```C
Messages:
prepare                  1812    1.15MiB
prepare_ok               1802    450.50KiB
ping                     750     375.00KiB
pong                     750     187.50KiB
request_headers          442     110.50KiB
headers                  442     1.62MiB
commit                   325     81.25KiB
request_prepare          285     71.25KiB
request                  275     200.23KiB
reply                    264     173.48KiB
start_view               98      258.75KiB
do_view_change           30      15.00KiB
ping_client              15      3.75KiB
pong_client              15      3.75KiB
request_start_view       13      3.25KiB
total                    7318    4.66MiB


          PASSED (2048 ticks)
```

</details>

<details>
<summary>5% packet loss,  ./zig/zig build vopr -Drelease -- --packet-loss-ratio=5/100 --performance 8243974837816383288</summary>

#### New repair with non-alternating replication

```C
Messages:
prepare                  1548    669.86KiB
prepare_ok               1207    301.75KiB
ping                     930     465.00KiB
pong                     912     228.00KiB
commit                   380     95.00KiB
request                  333     160.39KiB
reply                    270     117.51KiB
block                    270     214.01KiB
request_headers          211     52.75KiB
request_blocks           201     65.78KiB
headers                  193     703.25KiB
request_prepare          129     32.25KiB
start_view               129     344.25KiB
ping_client              97      24.25KiB
pong_client              96      24.00KiB
request_start_view       54      13.50KiB
do_view_change           25      12.50KiB
request_reply            3       768B
total                    6988    3.44MiB


          PASSED (2700 ticks)
```

#### New repair with 100ms repair timeout and non-alternating replication

```C
Messages:
prepare                  1617    692.70KiB
prepare_ok               1566    391.50KiB
ping                     780     390.00KiB
pong                     763     190.75KiB
request_headers          393     98.25KiB
headers                  383     1.40MiB
commit                   330     82.50KiB
request_prepare          295     73.75KiB
request                  279     123.80KiB
reply                    268     126.46KiB
start_view               89      233.75KiB
do_view_change           25      12.50KiB
ping_client              19      4.75KiB
pong_client              19      4.75KiB
request_start_view       14      3.50KiB
total                    6840    3.78MiB


          PASSED (2184 ticks)
```

</details>

<details>
<summary>30% packet loss, ./zig/zig build vopr -Drelease -- --packet-loss-ratio=30/100 --performance 1589881801411644986</summary>

```C
Messages:
ping                     4350    2.12MiB
pong                     3501    875.25KiB
prepare                  2517    1.72MiB
block                    2253    4.27MiB
prepare_ok               1607    401.75KiB
request_blocks           1566    541.84KiB
commit                   1530    382.50KiB
request                  1266    1.10MiB
ping_client              1102    275.50KiB
pong_client              900     225.00KiB
request_prepare          702     175.50KiB
request_headers          662     165.50KiB
headers                  539     1.93MiB
request_reply            408     102.00KiB
reply                    368     297.28KiB
start_view               313     860.50KiB
request_start_view       276     69.00KiB
do_view_change           30      15.00KiB
total                    23890   15.43MiB


          PASSED (14040 ticks)
```

#### New repair with 100ms repair timeout and non-alternating replication

```C
Messages:
prepare                  2380    1.09MiB
prepare_ok               1887    471.75KiB
ping                     1610    805.00KiB
pong                     1289    322.25KiB
request_prepare          1167    291.75KiB
request_headers          865     216.25KiB
headers                  685     2.50MiB
commit                   615     153.75KiB
request                  591     312.17KiB
reply                    357     193.13KiB
ping_client              355     88.75KiB
pong_client              286     71.50KiB
start_view               138     370.25KiB
request_start_view       64      16.00KiB
do_view_change           45      22.50KiB
total                    12334   6.85MiB


          PASSED (4921 ticks)
```

</details>

src/vsr/replica.zig
chaitanyabhandari: Great catch! In https://github.com/tigerbeetle/tigerbeetle/pull/2957/commits/2ead82849f9c3fbb33c8bfc3a37bd91574bff72e, I just made it so that we don't check `repair_messages_budget == 0` for replicas in `view_change` at all.

chaitanyabhandari:
Since deterministic performance testing showed that the amount of headers traffic generated these optimizations is significantly higher than the current traffic, aebecc667e0bdde9286fdd975659cff033fa8a8c implements an optimization for `request_headers` that reduces this traffic.

For example, for the no faults case, we now exchange `403.50KiB` worth of _headers_, as opposed to `2.34MiB`, a ~6x improvement! On main, the old repair protocol exchanges `172.75KiB` worth of _headers_. This is understandable since repair is more aggressive now, given we correctly replenish repair budget _and_ we invoke repair during the commit path as well.

Similarly, for the 5% packet loss case, we now exchange `234.25KiB` as opposed to `1.40MiB` of _headers_. On main, the old repair protocol exchanges `293.50KiB` worth of headers, which is _more_ than what it is currently!


<details>

<summary>No faults, ./zig/zig build vopr -Drelease -- --performance 4380260082931739075</summary>

```C
Messages:
prepare                  1747    1.01MiB
prepare_ok               1738    434.50KiB
ping                     750     375.00KiB
pong                     749     187.25KiB
commit                   325     81.25KiB
request_headers          285     71.25KiB
headers                  276     403.50KiB
request                  271     178.33KiB
reply                    261     155.52KiB
request_prepare          211     52.75KiB
start_view               87      227.00KiB
do_view_change           30      15.00KiB
ping_client              11      2.75KiB
pong_client              11      2.75KiB
request_start_view       2       512B
total                    6754    3.14MiB


          PASSED (2023 ticks)
```

</details>


<details>

<summary>5% packet loss, ./zig/zig build vopr -Drelease -- --packet-loss-ratio=5/100 --performance 8243974837816383288</summary>

```C
Messages:
prepare                  1581    680.09KiB
prepare_ok               1206    301.50KiB
ping                     1140    570.00KiB
pong                     1110    277.50KiB
commit                   445     111.25KiB
request                  390     194.83KiB
block                    361     360.53KiB
request_blocks           303     100.88KiB
reply                    283     121.48KiB
request_headers          157     39.25KiB
ping_client              154     38.50KiB
pong_client              152     38.00KiB
headers                  141     234.25KiB
request_prepare          128     32.00KiB
start_view               102     271.00KiB
request_reply            55      13.75KiB
request_start_view       28      7.00KiB
do_view_change           25      12.50KiB
total                    7761    3.32MiB


          PASSED (3341 ticks)
```

</details>

chaitanyabhandari:
@sentientwaffle The changes in this PR make sense to me from first principles:
* The budget replenishment would certainly be an improvement over the status quo and increase throughput, because we currently don't replenish budget at all, before the repair_timeout fires.
* Cycling through the prepares also makes sense, to ensure that even though we replenish budget, we don't repeatedly request the same missing prepares.

However, they somehow don't seem to improve performance in the *absolute* sense, i.e. the reduce # of ticks our DPT experiments... the *why* of which is something that is not entirely clear to me yet.

The absolute performance did increase when we reduced `repair_timeout` (which we saw in https://github.com/tigerbeetle/tigerbeetle/pull/2957#issuecomment-2867164325 when we reduced it from 500ms → 100ms), but it was at quite a high cost of extra messages being sent. To find a reasonable middle ground for network utilization, I also did some experiments with `repair_timeout` changed from 500ms → 250ms.

It improves the performance (# of ticks) in all experiments we ran. However, it comes at the cost of extra messages sent in most cases (except the case with one replica down):

| Experiment    | # of ticks (old) |  # of ticks (new)  | Messages (old) | Messages (new) | Headers (old)| Headers (new) |
| -------- | ------- |  ------- | ------- | ------- | ------- | ------- |
| No faults  | 2074   | 1990| 3.33MiB | 3.77MiB | 172.75KiB |  556.25KiB|
| 5% packet loss | 2636    | 2408  | 2.68MiB | 3.10MiB| 293.50KiB  | 695.50KiB |
| 30% packet loss    | 6771   | 6041 | 7.72MiB | 7.89MiB |1.65MiB  | 1.71MiB |
|  One replica (R3) down | 6041   | 3805| 6.21MiB | 5.10MiB| 839.50KiB | 786.25KiB |


<details>
<summary>No faults, ./zig/zig build vopr -Drelease -- --performance 4380260082931739075</summary>

```C
Messages:
prepare                  1861    1.36MiB
prepare_ok               1844    461.00KiB
ping                     720     360.00KiB
pong                     720     180.00KiB
commit                   325     81.25KiB
request_headers          298     74.50KiB
headers                  295     556.25KiB
request                  273     242.88KiB
reply                    262     187.62KiB
request_prepare          247     61.75KiB
start_view               93      243.50KiB
do_view_change           30      15.00KiB
ping_client              13      3.25KiB
pong_client              13      3.25KiB
request_start_view       3       768B
total                    6997    3.77MiB


          PASSED (1990 ticks)
```
</details>

<details>
<summary>5% packet loss, ./zig/zig build vopr -Drelease -- --packet-loss-ratio=5/100 --performance 82439748378163832885</summary>

```C
Messages:
prepare                  1661    701.91KiB
prepare_ok               1608    402.00KiB
ping                     865     432.50KiB
pong                     842     210.50KiB
commit                   360     90.00KiB
request                  295     132.14KiB
request_headers          294     73.50KiB
headers                  286     695.50KiB
reply                    278     128.70KiB
request_prepare          247     61.75KiB
start_view               83      217.00KiB
ping_client              35      8.75KiB
pong_client              34      8.50KiB
do_view_change           25      12.50KiB
request_start_view       3       768B
total                    6916    3.10MiB


          PASSED (2408 ticks)
```
</details>

<details>

<summary>30% packet loss, ./zig/zig build vopr -Drelease -- --packet-loss-ratio=30/100 --performance 1589881801411644986</summary>

```C
Messages:
prepare                  2627    2.07MiB
prepare_ok               2049    512.25KiB
ping                     1945    972.50KiB
pong                     1589    397.25KiB
request_prepare          911     227.75KiB
request                  756     805.89KiB
commit                   735     183.75KiB
request_headers          662     165.50KiB
headers                  546     1.71MiB
ping_client              544     136.00KiB
pong_client              453     113.25KiB
reply                    359     272.48KiB
start_view               141     374.00KiB
request_start_view       58      14.50KiB
do_view_change           45      22.50KiB
total                    13420   7.89MiB


          PASSED (6041 ticks)
```
</details>

<details>
<summary>.R=3 down, ./zig/zig build vopr -Drelease -- --replica-missing=3 --performance 5036667415013174002</summary>

```C
Messages:
prepare                  1975    1.60MiB
prepare_ok               1496    374.00KiB
ping                     1065    532.50KiB
pong                     841     210.25KiB
request_prepare          546     136.50KiB
commit                   515     128.75KiB
request_headers          460     115.00KiB
request                  431     409.38KiB
headers                  340     786.25KiB
reply                    320     289.95KiB
ping_client              195     48.75KiB
start_view               179     490.00KiB
pong_client              144     36.00KiB
request_start_view       89      22.25KiB
do_view_change           25      12.50KiB
total                    8621    5.10MiB


          PASSED (3805 ticks)
```

</details>

chaitanyabhandari:
One thing here that has me baffled is why https://github.com/tigerbeetle/tigerbeetle/pull/2957/commits/0e2467c7891ddd8683dbf217d2bbfcb34c63929f isn't contributing to any improvement in the performance. The way that I understood it, proactively calling `repair`  via the commit path (when we find a missing header/prepare) should behave equivalently to reducing the `repair_timeout`, since ultimately, both entail invoking `repair`. In fact, my understanding is that https://github.com/tigerbeetle/tigerbeetle/pull/2957/commits/0e2467c7891ddd8683dbf217d2bbfcb34c63929f should be even _more_ aggressive than merely reducing the timeout. And since its more aggressive, it should help us repair faster (and reduce # of ticks in the DPT experiments)...

But the actual behavior is _completely_ defying my understanding, there's no improvement due to that commit! 🤯

chaitanyabhandari:
I investigated why 0e2467c7891ddd8683dbf217d2bbfcb34c63929f isn't leading to better performance, and found that even though it entails aggressively calling repair, it also falls prey to an issue where we call `request_headers` for a _small_ range of headers. In other words, we are too eager to quickly detect every prepare that's missing, so we _first_ fetch the header for that prepare, and _then_ the prepare itself. And we do this for most missing prepares! **_Essentially, this approach does not allow for batching headers_**. 

So, based on this new understanding, I think that we should _not_ rely on calling `repair` in `on_prepare`/on the commit path, and revert 0e2467c7891ddd8683dbf217d2bbfcb34c63929f. Instead, we should reduce the `repair_timeout` to a lower value that improves our detection latency, and also allows for batching headers. Of course, it shouldn't be too chatty over the network (for example, 250ms as we saw in https://github.com/tigerbeetle/tigerbeetle/pull/2957#issuecomment-2874123200). I will also post the numbers for 100ms so we can take a well informed call for this!

chaitanyabhandari:
https://github.com/tigerbeetle/tigerbeetle/pull/2957/commits/e60fcd4687ccbf4be2c168fc61dbf2613b9947a8 improves the header-tracking optimization, and makes the overhead of reducing repair_timeout from 500ms → 100ms much more tractable! For the `30% packet loss` and `One replica down` case, we are now _less_ chatty than the protocol on `main`. And of course, we are still faster in all cases.

I _think_ we can safely reduce the repair_timeout to 100ms now. Thoughts @sentientwaffle?


### repair_timeout reduced from 500ms → 100ms

| Experiment    | # of ticks (old) |  # of ticks (new)  | Messages (old) | Messages (new) | Headers (old)| Headers (new) |
| -------- | ------- |  ------- | ------- | ------- | ------- | ------- |
| No faults  | 2074   | 2033| 3.33MiB | 4.22MiB | 172.75KiB |  625.00KiB|
| 5% packet loss | 2636    | 2314  | 2.68MiB | 2.98MiB| 293.50KiB  | 492.00KiB |
| 30% packet loss    | 6771   | 5209 | 7.72MiB | 6.95MiB |1.65MiB  | 668.50KiB |
|  One replica (R3) down | 6041   | 3373| 6.21MiB | 4.47MiB| 839.50KiB | 521.50KiB |

### repair_timeout reduced from 500ms → 250ms


| Experiment    | # of ticks (old) |  # of ticks (new)  | Messages (old) | Messages (new) | Headers (old)| Headers (new) |
| -------- | ------- |  ------- | ------- | ------- | ------- | ------- |
| No faults  | 2074   | 2001| 3.33MiB | 3.90MiB | 172.75KiB | 518.00KiB|
| 5% packet loss | 2636    | 2341  | 2.68MiB | 2.80MiB| 293.50KiB  | 408.00KiB |
| 30% packet loss    | 6771   | 5997 | 7.72MiB | 5.58MiB |1.65MiB  | 488.50KiB |
|  One replica (R3) down | 6041   | 3656| 6.21MiB | 4.44MiB| 839.50KiB | 396.50KiB |

sentientwaffle:
Excellent! How does this look when combined with [adaptive routing](https://github.com/tigerbeetle/tigerbeetle/pull/2880)? The alternating routing that we use right now is less than idea for repair.

chaitanyabhandari:
Ah, _great_ idea to get a feel of how this will fare with adaptive routing! Numbers coming soon...

matklad:
>I think we can safely reduce the repair_timeout to 100ms now.

Absolutely no context here as I am still stuck polling the message bus, but my thinking here was that:

* we _keep_ repair timeout somewhat large
* but we ensure that it is only a backstop for an idle cluster
* and otherwise we actively trigger repair every time we receive a prepare

to relate our repair rate to our prepare rate. 

EDIT: disregard, I've now read the message above 🤣

matklad:
>so we first fetch the header for that prepare, and then the prepare itself. And we do this for most missing prepares! Essentially, this approach does not allow for batching headers.


Can we speed this up by sending a _prepare_ whenever a replcia requests a single missing header? That is:

* I am a replica
* I receive a `.request_headers`
* "Oh, look, they only want _one_ header" -- notice I
* I send the prepare directly, rather than the header.

chaitanyabhandari:
@sentientwaffle here's the numbers from adaptive routing, even smaller overheads there!


### repair_timeout reduced from 500ms → 100ms, adaptive routing enabled

| Experiment    | # of ticks (old) |  # of ticks (new)  | Messages (old) | Messages (new) | Headers (old)| Headers (new) |
| -------- | ------- |  ------- | ------- | ------- | ------- | ------- |
| No faults  | 1797   | 1648| 3.10MiB | 3.79MiB | 242.25KiB |  571.50KiB|
| 5% packet loss | 2222    | 1778  | 2.19MiB | 2.38MiB| 291.00KiB | 416.75KiB |
| 30% packet loss    | 5862   | 3968 | 5.93MiB | 5.12MiB |1.46MiB  | 610.75KiB|
|  One replica (R3) down | 2107   | 2039| 2.75MiB | 2.97MiB| 174.75KiB | 197.25KiB |

---

Also @matklad, _interesting_ idea about sending a prepare when a single header is requested! I think it would help this case, however, I have a few apprehensions that I will have to give some more thought to:

* We now increment the repair budget in `on_headers`, when we receive the headers that we needed for repairs, but with this optimization accounting for the budget becomes a bit tricky. We'd probably have to special case one-header `request_headers`, such that they don't use the budget at all? But that seems a bit messy to me...
* The issue that I had observed was not necessarily requesting _one header_, could be >1 as well (but a small number), so this optimization may not help _all_ cases. Or, we'd have to come up with some threshold below which a replica can send _mutiple_ prepares to a replica, if its asking for a small number of headers.

Actually, one orthogonal thing that comes to mind here is, can we eliminate the round trip for fetching headers completely? What if we only have a repair prepares phase where we check for _missing headers_, and we straightaway request prepares using op. Then, the receiving replica could return whatever prepare it has corresponding to the op? Not a well thought idea though, so it might be complete gibberish... 😄

chaitanyabhandari:
@sentientwaffle Here are the results aggregated over 100 VOPR runs, trends are very similar to what we observed before:
* This PR improves performance in all scenarios tested (reduces # of ticks), marginally in the `No faults` case (_2064 → 2019_), but quite significantly with faults (_3024 → 2300_ in `5% packet loss`,  _6537 → 3503_ in `R5 down`).
* In the `No faults` case, we see a noticeable increase in the messages sent over the network (_3.25MiB → 4.22MiB_). In all other cases, this PR either decreases the messages sent (_5.58MiB → 4.51MiB_ in  `R5 down` cases, and  _7.95MiB → 6.40MiB_ in `30% packet loss`), or increasing them only marginally (_4.10MiB → 4.40MiB_ in `5% packet loss`).

| Experiment    | Ticks p50 (old, new) | Ticks p99 (old, new) | Msgs p50 (old, new) | Msgs p99 (old, new) |
| -------- | ------- |  ------- | ------- | ------- |
| No faults  | 2064 → 2019   |2217 → 2132 |3.25MiB → 4.22MiB|3.59MiB →  4.62MiB|
| 5% packet loss | 3024 → 2300    | 3596 →  2489 |4.10MiB → 4.40MiB | 5.08MiB → 4.98MiB| 
| 30% packet loss    | 6663  → 5142   | 7876 → 5675| 7.95MiB → 6.40MiB | 9.38MiB →  7.28MiB|
| 50% packet loss    | 13874 → 10394   | 19659 → 15146| 11.67MiB → 9.20MiB | 13.74MiB → 11.25MiB|
|  R0 down | 4845 → 4071   |5350 → 4499| 4.39MiB → 4.84MiB | 5.19MiB → 5.24MiB |
|  R1 down | 4918 → 4123   |5609 → 4385| 4.58MiB → 4.97MiB| 5.52MiB → 5.48MiB|
|  R2 down | 4801 → 4063   |5746 → 4441 |4.36MiB  → 4.81MiB | 5.66MiB → 5.31MiB|
|  R3 down | 5754 → 3335   |6476 → 3618 | 5.63MiB → 4.49MiB | 6.54MiB → 4.96MiB|
|  R4 down | 6533 → 2654   |7440  → 2881| 6.65MiB →  4.73MiB| 7.94MiB → 5.41MiB|
|  R5 down | 5649 → 3283   |6537 → 3503| 5.58MiB → 4.51MiB| 6.84MiB → 5.04MiB|

chaitanyabhandari:
I implemented a proof of concept for the optimization in https://github.com/tigerbeetle/tigerbeetle/pull/2957#issuecomment-2876921396, and tested it out on the above scenarios. For each case it only increased the performance by a very small amount. Upon further investigation, I found that it was because the optimization was only applying to a _small_ fraction (~10%) of `request_headers` requests where only a single header was being requested. 

 A more proactive repair (driven by `on_prepare`) requires more thinking about the design, and since the current PR has gone through several iterations and is certainly an improvement over the status quo, perhaps we should consider getting these changes in? Thoughts @sentientwaffle?

chaitanyabhandari:
@sentientwaffle Experimented with a more aggressive repair where `repair` is called in _every_ `on_prepare` (https://github.com/tigerbeetle/tigerbeetle/pull/2957/commits/2eba3f19cd813613180de736cc6f8cfc66e8700e).

First, I implemented a proof of concept for the optimization in https://github.com/tigerbeetle/tigerbeetle/pull/2957#issuecomment-2876921396, and tested it out on the above scenarios, (along with a _500ms repair timeout_). For each case it only increased the performance by a very small amount. Upon further investigation, I found that it was because the optimization was only applying to a _small_ fraction (~10%) of `request_headers` requests where only a single header was being requested. So, it became clear that with a _large_ repair_timeout, implementing proactive repair such that it gives the equivalent gains to a reduced repair_timeout would require some more thinking.


Then, I had a hunch that our repair budget, the (new) state we maintain for repairing prepares (which only cycles around to the beginning on repair_timeout), along with the (new) optimizations for reducing `request_headers` traffic, would make the cost of calling `repair` in _every_ `on_prepare` tractable, even with a small, _100ms repair_timeout_! 

Ran some experiments, and indeed! Nearly the same performance as https://github.com/tigerbeetle/tigerbeetle/pull/2957#issuecomment-2877956332, with nearly the same network cost!

| Experiment    | Ticks p50 (old, new) | Ticks p99 (old, new) | Msgs p50 (old, new) | Msgs p99 (old, new) |
| -------- | ------- |  ------- | ------- | ------- |
| No faults  | 2064 → 2007   |2217 → 2118 |3.25MiB → 4.41MiB|3.59MiB →  4.88MiB|
| 5% packet loss | 3024 → 2307    | 3596 →  2461 |4.10MiB → 4.70MiB | 5.08MiB → 5.11MiB| 
| 30% packet loss    | 6663  → 5090   | 7876 → 5591| 7.95MiB → 6.50MiB | 9.38MiB →  7.36MiB|
| 50% packet loss    | 13874 → 10364   | 19659 → 13571| 11.67MiB → 9.17MiB | 13.74MiB → 10.60MiB|
|  R0 down | 4845 → 3968   |5350 → 4334| 4.39MiB → 4.90MiB | 5.19MiB → 5.39MiB |
|  R1 down | 4918 → 4020   |5609 → 4385| 4.58MiB → 5.12MiB| 5.52MiB → 5.54MiB|
|  R2 down | 4801 → 3968   |5746 → 4247 |4.36MiB  → 4.80MiB | 5.66MiB → 5.38MiB|
|  R3 down | 5754 → 3256   |6476 → 3507 | 5.63MiB → 4.65MiB | 6.54MiB → 5.24MiB|
|  R4 down | 6533 → 2726   |7440  → 2991| 6.65MiB →  4.87MiB| 7.94MiB → 5.46MiB|
|  R5 down | 5649 → 3224   |6537 → 3478| 5.58MiB → 4.63MiB| 6.84MiB → 5.03MiB|



cc: @matklad

src/vsr/replica.zig
```diff
@@ -471,6 +471,21 @@ pub fn ReplicaType(
         // +1 to make it easier to decrement budget with -|.
         repair_messages_budget: u32 = constants.vsr_repair_message_budget_max + 1,
 
+        // The next op between [commit_min + 1, op] to be requested during repair.
+        // Used to cycle through uncommitted prepares, to avoid resending the same missing prepares
```
```suggestion
        // Used to cycle through uncommitted prepares, to avoid requesting the same missing prepares
```
This does also prevent us from resending the same missing prepares, but the direct purpose of this field is to prevent redundant requests.

(Likewise for the below comment.)

src/vsr/replica_test.zig
```diff
@@ -118,6 +118,12 @@ test "Cluster: recovery: WAL prepare corruption (R=3, corrupt root)" {
 
     try c.request(1, 1);
     try expectEqual(t.replica(.R_).commit(), 1);
+
+    const r0 = t.replica(.R0);
+    const r0_storage = &t.cluster.storages[r0.replicas.get(0)];
+    const fault_offset = vsr.Zone.wal_prepares.offset(0);
+    const fault_sector = @divExact(fault_offset, constants.sector_size);
+    try expect(!r0_storage.faults.isSet(fault_sector));
```
sentientwaffle: Would `storage.area_faulty(.{ .wal_prepares = 0 })` work?

src/vsr/replica.zig
```diff
@@ -3376,11 +3417,15 @@ pub fn ReplicaType(
...
             assert(self.status == .normal or self.status == .view_change);
             assert(self.repair_messages_budget <= constants.vsr_repair_message_budget_max + 1);
+
             self.repair_messages_budget = @min(
                 (self.repair_messages_budget + constants.vsr_repair_message_budget_refill),
                 // +1 to make it easier to decrement budget with -|.
                 constants.vsr_repair_message_budget_max + 1,
             );
+            self.repair_prepare_op_next_committed = self.op_repair_min();
+            self.repair_prepare_op_next_uncommitted = self.commit_min + 1;
```
sentientwaffle: I think we also need to reset `repair_header_op_next`.

src/vsr/replica.zig
```diff
@@ -2128,7 +2148,14 @@ pub fn ReplicaType(
...
                 // Write prepare adds it synchronously to in-memory pipeline cache.
                 // Optimistically start committing without waiting for the disk write to finish.
-                if (self.status == .normal and self.backup()) self.commit_journal();
+                if (self.status == .normal and self.backup()) {
+                    self.commit_journal();
+                    self.repair_messages_budget = @min(
+                        self.repair_messages_budget + 1,
+                        constants.vsr_repair_message_budget_max + 1,
+                    );
+                    self.repair();
```
sentientwaffle, line 2161: So is the idea of incrementing this here (and kicking off the repair) rather then after we have written to the journal that we can get a request in flight while we are still writing to disk?

sentientwaffle (commented):
Note that devhub has a failing seed too:

```
./zig/zig build -Drelease vopr -- 12126348985487966327
```

src/vsr/replica_test.zig
chaitanyabhandari: Ah, right! `area_faulty` does _precisely_ what I want, thanks DJ. 👍🏼

src/vsr/replica.zig
chaitanyabhandari: Ack, that was my intended logic here, to pipeline disk IO & sending repair messages over the network!

chaitanyabhandari: Actually, let me add a comment there so its clear. :)

src/vsr/replica.zig
chaitanyabhandari: Mhmmmm, that's actually what I _started_ with here, setting `repair_header_op_next` to `op_repair_min()`... However, it actually doesn't play very well with our reduced `repair_timeout`. Specifically, at every `repair_timeout`, we would fetch all the headers from `op_repair_min → range.op_max`, which would be wasted work (which showed up as increased `headers` traffic in the DPT experiments), since we are guaranteed to have requested headers up till `repair_header_op_next` _once_.

Instead, I thought that what we _could_ do here is to make it so that `repair_header_op_next` always increases. Now, we requests extra headers from `repair_header_op_next → range.op_max`. And if we do have a break _before_ `repair_header_op_next`, then we just send a request from `op_min → op_max`.

```C
                    self.send_header_to_replica(
                            .command = .request_headers,
                            .cluster = self.cluster,
                            .replica = self.replica,
                            // Pessimistically request extra headers. Requesting/sending extra
                            // headers is inexpensive, and it may save us extra round-trips to
                            // repair earlier breaks.
                            .op_min = @min(range.op_min, self.repair_header_op_next),
                            .op_max = range.op_max,
                        }),
```

sentientwaffle: Ahh, I see what you mean.  So our first repair request we'll cover many headers, but after that the range shrinks down. And this gets restored naturally as the replica makes progress.

chaitanyabhandari: Ack, precisely what I was thinking!

chaitanyabhandari:
Addressed review in https://github.com/tigerbeetle/tigerbeetle/pull/2957/commits/71b20058889efa3ea5ff8a0abca216493ee10017! 

Also, `./zig/zig build vopr -- 12126348985487966327` is a crash in the `core_missing_prepare` logic for handling uncommitted headers... Seems like an old regression unrelated to this PR! It _doesn't_ fail on main though, I will investigate this more tomorrow.
```C
thread 136595404 panic: attempt to use null value
/Users/chaitanyabhandari/Desktop/tigerbeetle/tigerbeetle/src/vopr.zig:1003:86: 0x1048940ef in core_missing_prepare (vopr)
                    const uncommitted_header = uncommitted_headers[op % pipeline_max].?;
                                                                                     ^
/Users/chaitanyabhandari/Desktop/tigerbeetle/tigerbeetle/src/vopr.zig:820:54: 0x104892f47 in cluster_recoverable (vopr)
        } else if (try simulator.core_missing_prepare(gpa)) |header| {
                                                     ^
/Users/chaitanyabhandari/Desktop/tigerbeetle/tigerbeetle/src/vopr.zig:318:50: 0x1048980af in main (vopr)
            if (try simulator.cluster_recoverable(gpa)) {
                                                 ^
/Users/chaitanyabhandari/Desktop/tigerbeetle/tigerbeetle/zig/lib/std/start.zig:524:37: 0x10489c037 in main (vopr)
```

src/vsr/replica.zig
chaitanyabhandari: Also included the reasoning in a comment, since its important it doesn't _just_ live in my head. 😄 
https://github.com/tigerbeetle/tigerbeetle/blob/71b20058889efa3ea5ff8a0abca216493ee10017/src/vsr/replica.zig#L3431-L3441

chaitanyabhandari:
`./zig/zig build vopr -- 12126348985487966327` is _interesting_, but certainly unrelated to the current PR! Seems like to calculate `cluster_op_head`, we just iterate through all replicas in the core and do `cluster_op_head = @max(cluster_op_head, replica.op);`. However, this could be an issue because we could end up assigning cluster_op_head to a _truncated_ op. This is precisely what happened here, `cluster_op_head=144` was selected from R2, which is on log_view=10547. But op=144 was truncated in view=10548!

```C
0: commit_max=143, log_view=10550, op=143, op_repair_min=120
2: commit_max=143, log_view=10547, op=144, op_repair_min=120
3: commit_max=143, log_view=10548, op=143, op_repair_min=120
4: commit_max=143, log_view=10543, op=143, op_repair_min=120
0: fetching header for uncommitted op=144
0: no header for uncommitted op=144
thread 137539227 panic: attempt to use null value
/Users/chaitanyabhandari/Desktop/tigerbeetle/tigerbeetle/src/vopr.zig:0:0: 0x102b8fc43 in cluster_recoverable (vopr)
/Users/chaitanyabhandari/Desktop/tigerbeetle/tigerbeetle/src/vopr.zig:318:50: 0x102b967bf in main (vopr)
            if (try simulator.cluster_recoverable(gpa)) {
```

The fix is simple (https://github.com/tigerbeetle/tigerbeetle/pull/2957/commits/2a8870df9efbbd173635b9d118b25f561de748c5), choose the cluster's head while adhering to log_view.

src/vopr.zig
```diff
@@ -958,9 +958,15 @@ pub const Simulator = struct {
         for (simulator.cluster.replicas) |replica| {
             if (!simulator.core_repairable_replica(Cluster.Replica, &replica)) continue;
 
-            cluster_commit_max = @max(cluster_commit_max, replica.commit_max);
+            if (replica.log_view > cluster_log_view) {
+                maybe(cluster_op_head > replica.op);
+                cluster_op_head = replica.op;
+            } else if (replica.log_view == cluster_log_view) {
+                cluster_op_head = @max(cluster_op_head, replica.op);
+            }
```
sentientwaffle, line 966: Excellent catch!

src/vsr/replica.zig
```diff
@@ -2862,13 +2893,27 @@ pub fn ReplicaType(
...
+            // duplicate headers messages.
+            if (header_repaired) {
+                self.repair_messages_budget = @min(
+                    self.repair_messages_budget + 1,
+                    // +1 to make it easier to decrement budget with -|.
+                    constants.vsr_repair_message_budget_max + 1,
+                );
+            }
+
             self.repair();
```
sentientwaffle, line 2917: Would it make sense to only call repair here if we have just incremented the budget?

sentientwaffle (approved):
(Sorry, meant to approve!)

src/vsr/replica.zig
chaitanyabhandari: It _could_ be that we already have a non-zero budget even if we don't increment it, and in that case we'd miss out on an opportunity to repair. Perhaps better to call it regardless? 🤔

PR #2958, opened 2025-05-09
inspect: add tigerbeetle inspect op

matklad:
Shows checkponts&triggers around the given op:

```
$ tigerbeetle inspect op 624894739
checkpoint          op                  trigger             prepare_max         checkpoint_next
624894719      +20  624894739      +12  624894751      +16  624894767      +912 624895679

$ tigerbeetle inspect op 624894751
checkpoint          trigger             op                  prepare_max         checkpoint_next
624894719      +32  624894751      +0   624894751      +16  624894767      +912 624895679
```

### DIFF
--- a/src/tigerbeetle/cli.zig
+++ b/src/tigerbeetle/cli.zig
@@ -138,6 +138,9 @@ const CLIArgs = union(enum) {
     const Inspect = union(enum) {
         constants,
         metrics,
+        op: struct {
+            positional: struct { op: u64 },
+        },
         superblock: struct {
             positional: struct { path: []const u8 },
         },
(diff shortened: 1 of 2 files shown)


scriptforfivem:
I'm impressed by the modularity of this codebase. Makes it very maintainable.

lewisdaly:
Can you help me understand the ordering here? In your above example:
- `624894739` prints `checkpoint, op, trigger, prepare_max, checkpoint_next`
- `624894751` prints `checkpoint, trigger, op, prepare_max, checkpoint_next` 

is that intentional? i.e. why is `checkpoint` always first and `trigger` and `op` seem to change places?

matklad:
Everything is printed in order. If CLI arg op is before the trigger, it goes first. If it is after the trigger, it goes second.

I don’t have any special handing for “what if op is exactly trigger/checkpoint/prepare_max”, but hopefully +0 in the output makes it clear that two entries are in fact the same!

lewisdaly:
Ah cool, thanks for clarifying

lewisdaly (approved):

PR #2962, opened 2025-05-10
Add support for balancing with pending

bovle:
I ran into issues when trying to temporarily enforce `debits_must_not_exceed_credits` on an account with pending debits.

The problem was that the `balancing_credit` flag didn't take pending debits into account.

I've added a `balancing_debit_with_pending` and a `balancing_credit_with_pending` flag to fix this.

Let me know what you think. If it makes sense i can add some more test/documentation.

### DIFF
--- a/src/state_machine.zig
+++ b/src/state_machine.zig
@@ -6358,7 +6358,7 @@ test "create_transfers: per-transfer balance invariant with pending" {
         \\ account A3  0  0  0  0  _  _  _ _ L1 C1   _   _ C<D _ _ _ _ _ ok
         \\ commit create_accounts
         \\
-        \\ setup A2 10 30 0 0
+        \\ setup A2 0 50 10 0
         \\
         \\ transfer T1 A1 A2 41   _  _  _  _    0 L1 C1 LNK   _ _   _   _   _ _ _ _ _ _ _ _ linked_event_failed
         \\ transfer T2 A2 A3  1   _  _  _  _    0 L1 C1 LNK PEN _   _ _   _ _ _ _ BDRP _ _ _ exceeds_debits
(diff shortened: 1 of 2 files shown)


batiati:
Hi @bovle, thank you for the PR!

Can you please help me to understand what do you mean by "_The problem was that the balancing_credit flag didn't take pending debits into account_" ? Because in fact, we do take the pending balance into account (to avoid double spending), so I couldn't figure out what use case are you trying to solve exactly.


> I ran into issues when trying to temporarily enforce `debits_must_not_exceed_credits`

For reference, we've tried something similar in the past (https://github.com/tigerbeetle/tigerbeetle/pull/2326), just to realize that we could already validate balances for specific transfers using the [balance invariant recipe](https://docs.tigerbeetle.com/coding/recipes/balance-invariant-transfers/), which can be adapted for many similar use cases (e.g., credit limits, quotas, capping, etc).

bovle:
Hi @batiati 

Thanks for the comments!

The problem I have is that the [balancing_credit](https://docs.tigerbeetle.com/reference/transfer/#flagsbalancing_credit) flag considers pending_credits and not pending_debits: `credit_account.credits_pending + credit_account.credits_posted ≤ credit_account.debits_posted`
I think the [tests](https://github.com/tigerbeetle/tigerbeetle/pull/2962/files#diff-af5dc615b5a3545b695dbff58d1ded942a93236c3f4314dfc4f08a41707553b2R6353-R6417) I have added showcase the problem well. If the debit account has pending debits the current balance invariant recipe will not stop the transfer.

bovle:
Hi again @batiati 

Anything I can do to move this forward? Should I bring this pr out of draft state?
I was thinking that I could reduce this change to a single flag: "consider_pending" or something that can only be used in conjunction with the balancing_XXX flags.
Let me know what you think

PR #2967, opened 2025-05-12
docs: denser quick start

matklad:
Supper important to be short here, let's hide away two platforms by default.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,8 @@
 Run a single-replica cluster on Linux (or [other platforms](https://docs.tigerbeetle.com/start/)):
 
 ```console
-$ curl -Lo tigerbeetle.zip https://linux.tigerbeetle.com && unzip tigerbeetle.zip && ./tigerbeetle version
+$ curl -Lo tigerbeetle.zip https://linux.tigerbeetle.com && unzip tigerbeetle.zip
+$ ./tigerbeetle version
 $ ./tigerbeetle format --cluster=0 --replica=0 --replica-count=1 --development 0_0.tigerbeetle
 $ ./tigerbeetle start --addresses=3000 --development 0_0.tigerbeetle
 ```
--- a/docs/start.md
+++ b/docs/start.md
@@ -35,7 +35,7 @@ curl -Lo tigerbeetle.zip https://mac.tigerbeetle.com && unzip tigerbeetle.zip
 
 ```console
 powershell -command "curl.exe -Lo tigerbeetle.zip https://windows.tigerbeetle.com; Expand-Archive tigerbeetle.zip .
-./tigerbeetle version
+.\tigerbeetle version
 ```
 </details>
 
--- a/docs/start.md
+++ b/docs/start.md
@@ -10,22 +10,34 @@ solve in the [Concepts](./concepts/) section, but let's make some real transacti
 
 TigerBeetle is a single, small, statically linked binary.
 
-You can download a pre-built Linux binary from `tigerbeetle.com`:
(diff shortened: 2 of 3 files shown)


matklad:
image

matklad:
image

matklad:
I think that's already an improvement, but, @fabioarnold, would be great if we bring the website version to 92% by:

* adding a smidgen of JavaScript to pop-open the right tab based on the browser's operating system, if we can get that. I'd check https://rustup.rs to see how they do detection. 
* Playing with CSS here. At minimum, I think we should tone down the styling, we don't really need these huge borders _here_ I think. The minimal GitHub's rendering looks better. Or maybe don't mess with CSS, and actually add a bit of JS progressive enhancement to go and replace consecutive details with a tabbed interface? I think this thick "button" thing around `Linux` would actually look great if it were `Linux | macOS | Windows` toggle . https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/tab_role#example seems like a good worked example here.

README.md
```diff
@@ -17,7 +17,8 @@
 Run a single-replica cluster on Linux (or [other platforms](https://docs.tigerbeetle.com/start/)):
 
 ```console
-$ curl -Lo tigerbeetle.zip https://linux.tigerbeetle.com && unzip tigerbeetle.zip && ./tigerbeetle version
+$ curl -Lo tigerbeetle.zip https://linux.tigerbeetle.com && unzip tigerbeetle.zip
+$ ./tigerbeetle version
```
matklad, line 21: At the moment, `./tigerbeetle version` gets cut off, but I think it's worth putting on a line by itself, to teach the user "this is how I check if TB is operational".

fabioarnold:
rustup.rs checks `navigator.platform`. It's a bit of messy string checking, maybe not as bad as checking the user agent.
Since we're using CSP we would need to add content hashes of these one-off scripts to our <head> section. From what I've gathered is you can't do a tabbed control without JS, right?

docs/start.md
```diff
@@ -10,22 +10,34 @@ solve in the [Concepts](./concepts/) section, but let's make some real transacti
...
-# Windows
-powershell -command "curl.exe -Lo tigerbeetle.zip https://windows.tigerbeetle.com; Expand-Archive tigerbeetle.zip .; .\tigerbeetle version"
+</details>
+
+<details>
+<summary>Windows</summary>
+
+```console
+powershell -command "curl.exe -Lo tigerbeetle.zip https://windows.tigerbeetle.com; Expand-Archive tigerbeetle.zip .
+./tigerbeetle version
```
fabioarnold: Haven't tested the powershell script. But it was a backslash before.

```suggestion
.\tigerbeetle version
```

fabioarnold (dismissed):

matklad:
>From what I've gathered is you can't do a tabbed control without JS, right?

I _think_ you can, if you try hard enough, https://css-tricks.com/the-lost-css-tricks-of-cohost-org/, but I don't think there's a good solution that doesn't require swimming upstream the Niagara Falls. So let's do JS! I think we can stuff it together with the rest of docs javascript? We also could inline it in the docs themselves, but it won't work on GitHub or in vim, and will make the docs less readable. We _can_ add `class="platform-toggle"` in the docs, but I'd rather we detect it just from the shape, or looking at the `summary`'s contents.

matklad:
We had another one in Installing, fixed that as well!

fabioarnold (approved):

PR #2970, opened 2025-05-13
bus: pull-based message bus

matklad:
**Note:** PR was substantially reworked & simplified, here's the new description:

Before this commit, replica had to process every message that the bus
throws at replica. Now, replica gains the ability to "suspend" certain
messages in bus' receive buffer, to return to them later.

This is used to:

* postpone prepare messages until there's a journal write available
* postpone block messages until there's a grid repair write available

If a message is suspended, the bus won't be recving into the socket,
propagating backpressure all the way up to the sending replica.

To ameliorate head-of-line blocking issues, backpressure applies to
specific commands. If there are _many_ messages in a particular buffer,
they all will be considered.

If a replica suspends a message, it must call bus.ready_to_receive at
some later point to guarantee that messages are delivered.

Old Description below (also available in a blog form here: https://matklad.github.io/2025/05/14/scalar-select-aniti-pattern.html): 

---

Before this comment, message bus invoked an on_message callback every
time a recv syscall returned enough bytes to form a valid message. Now,
messages need to be explicitly pulled out of the bus via receive_message
method.

  ## Background

Pulling events is in general a good pattern for asynchronous systems.
Imagine a generic evented system which works like:

    for message in incoming_messages:
        process(message)

Note that because `process` takes time, by the time it finishes and we
are ready to pull the next message from the pile, there might be _many_
messages in the pile already! And usually there are _some_ things to be
gained from exploiting the pile as a whole, rather than looking at it as
just a sequence of messages:

* You can prioritize, and select the more important messages out of the
  pile, even if they arrived later.
* You can coalesce and remove "duplicate" messages (e.g., taking only
  the latest state update).
* You can batch message processing. For example, processing all updates
  in one go and sending only a single change notification.

These points don't significantly apply to TigerBeetle, but there's one
issue that does apply, backpressure!

  ## Precarious Prepare Pileup

In the today's push-based setup, Replica's `on_message` is called
immediately upon receiving the message, and the replica might deal with
the message immediately. This is not always possible, the replica might
lack the required resources.

One very specific instance of this that we observed in real-world
deployments is message bus receiving a whole bunch of tiny prepares in
one go, lets say, 20 of them. When that happens, the bus calls
on_message 20 times in a row. But because we only have 8 IOPs in the
journal, most of the prepares are dropped on the floor, which creates
then causes extra repair and prepare traffic.

The behavior we want here is for the replica to _only_ pull a prepare
message out of the message bus when it has a spare IOP, and that's
exactly what's achieved here.

  ## Implementation

The core of the change is that MessageBus loses the `on_message`
callback, and instead gains a

  fn receive_message(bus: *Bus) ?*Message

instead. The replcia can now poll the bus when it is ready to process a
message. The fully general solution here would expose something like

    const MessageIterator = struct {
        fn peek(it: *const MessageIterator) ?*const Header,
        fn next_take(it: *MessageIterator) *Message,
        fn next_skip(it: *MessageIterator) *const Message,
    }

to allow the replica to inspect all available messages and receive them
selectively, but:

- I don't think we need to go _that_ far
- and, even if we need, it'll be a relatively uninteresting diff on top
  of the present commit.

So instead there's just `peek_message`, `receive_message`, which only
allow stalling the processing, rather than being selective. And we only
stall if we don't have a write IOP available, so the stalls should be
short, and I wouldn't worry about us delaying pings too much.

The interesting question is _when_ should the replica call
`bus.receive_message()`? How do we notify about readiness? The proposed
answer is "all the time". On every iteration of the event loop, before
we go into the kernel to sleep, we invoke

    fn idle(replica: *Replica) void

callback which then polls the bus.

The nice thing about this solution is that we avoid function pointes or
weird stacks: `replcia.idle()` is a _direct_ call, and we do it right
from main, there's nothing above us on the stack.

The not-so-nice thing is that this _seems_ inefficient. The previous
approach, with on_message callback, notified us precisely! We can
restore that by keeping the on_message callback, but then letting the
replica poll for the message from that callback. This would be an
incomplete solution! If the replica decides to postpone receiving a
prepare, it still needs to poll the bus later, when it is ready to
process the message. So the full solution here would be to pull from the
bus in the on_message callback, and _also_ in the `write_prepare` /
`write_block` callback. But that would create ugly backtraces, where
there's unrelated stuff above in the stack. To fix that, one option
would be just to set a `need_to_pull_message_bus` _flag_ in the
`write_prepare_callback`. And then, when we have nothing else on the
call-stack, we can check the flag and poll the bus. But then this
becomes equivalent to the idle solution!

This line of thinking convinced myself that the loop with `.idle` in
main is the right approach.

What I _don't_ like about this solution is that it make the VOPR event
loop even more complex, but, given that the production loop looks fine,
I am ok with that.

### DIFF
--- a/src/message_buffer.zig
+++ b/src/message_buffer.zig
@@ -1,7 +1,6 @@
 const stdx = @import("./stdx.zig");
 const std = @import("std");
 const assert = std.debug.assert;
-const maybe = stdx.maybe;
 
 const vsr = @import("./vsr.zig");
 const MessagePool = @import("message_pool.zig").MessagePool;
@@ -11,29 +10,40 @@ const constants = vsr.constants;
 
 /// MessageBuffer is the interface between a MessageBus and a Replica for passing batches of
 /// messages while minimizing copies. It handles message framing, but doesn't do IO directly.
+///
+/// It is a producer-consumer ring buffer of bytes, with two twists:
+/// - consumer can skip over or "suspend" certain slices, to return to them later.
+/// - producer validates bytes against a checksum, and this validation is sticky: a message is
+///   validated once, even if it is skipped many times.
+///
+/// Invariant: suspend_size ≤ process_size ≤ advance_size ≤ receive_size
 pub const MessageBuffer = struct {
     /// The buffer passed to the kernel for reading into. This is Message rather than []u8 to
     /// enable zero-copy fast path. If a recv syscall reads exactly one message, no copying occurs.
     message: *Message,
-    /// For the case where a single recv fetched more than one message, the amount of bytes
-    /// consumed from the beginning of message.buffer.
-    consume_size: u32 = 0,
+
+    /// Suspended bytes, always a number of full messages.
+    suspend_size: u32 = 0,
+    /// Processed (consumed or suspended) bytes, always a number of full messages.
+    process_size: u32 = 0,
+    /// Bytes covered by a valid checksum, a number of full messages and maybe a header.
+    advance_size: u32 = 0,
     /// The amount of bytes received from the kernel.
-    /// Only consume_size..receive_size byte range of message.buffer is initialized.
     receive_size: u32 = 0,
 
     /// Which peer we are receiving from, inferred from the received messages. Set during parsing,
     /// and used by the message bus to map connections to replicas and clients.
     peer: vsr.Peer = .unknown,
 
-    // Parsing state machine:
-    // - valid_header: Header checksum is correct.
-    // - valid_body: Body checksum is correct, a full message can be consumed.
-    // - invalid: An error occurred, and the MessageBus should terminate connection.
-    //            Can be set by replica to indicate semantic errors, such as wrong cluster.
-    valid_header: bool = false,
-    valid_body: bool = false,
+    // An error occurred, and the MessageBus should terminate connection.
+    // Can be set by replica to indicate semantic errors, such as wrong cluster.
     invalid: ?InvalidReason = null,
+
+    iterator_state: enum {
+        idle,
+        after_peek,
+        after_consume_suspend,
+    } = .idle,
 
     const InvalidReason = enum {
         header_checksum,
@@ -42,6 +52,12 @@ pub const MessageBuffer = struct {
         body_checksum,
         misdirected,
     };
+
+    fn invariants(buffer: *MessageBuffer) void {
+        assert(buffer.suspend_size <= buffer.process_size);
+        assert(buffer.process_size <= buffer.advance_size);
+        assert(buffer.advance_size <= buffer.receive_size);
+    }
 
     pub fn init(pool: *MessagePool) MessageBuffer {
         return .{ .message = pool.get_message(null) };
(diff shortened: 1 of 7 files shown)


src/io/darwin.zig
```diff
@@ -43,39 +46,39 @@ pub const IO = struct {
...
+    /// Use with run_for_ns_setup:
+    ///
+    ///     io.run_for_ns_setup(timeout);
+    ///     while (try io.run_for_ns()) {
+    ///         // Can schedule extra work here.
+    ///     }
+    pub fn run_for_ns(self: *IO) !bool {
+        if (self.run_for_ns_timed_out) {
+            self.run_for_ns_completion = undefined;
+            self.run_for_ns_timed_out = false;
```
jorangreef: Nit: Swap lines 77 and 76, so that after the check of `self.run_for_ns_timed_out` we immediately then toggle it (to move related variables closer together in space/time).

matklad:
Still hacking on this, but the implementation strategy is _much_ simpler now:

- no more idle callback!
- instead, we just naturally extend batched message interface introduced recently to allow for skipping individual messages
- and the replica explicitly pokes message bus after finishing journal/block writes to signal that it is ready to receive new messages, if the bus has some.

matklad:
Marking as ready for review! 

I need to do a self-review tomorrow with fresh eyes, but I am overall happy here!

src/message_buffer.zig
sentientwaffle, line 1: What would be some interesting metrics to track related to the pull based message bus? e.g.:

- current length of queue
- number of times we suspend a message

src/vsr/replica.zig
sentientwaffle, line 1: A possible performance consideration:
- if one of the replicas in the ring is repairing then it is more likely to be starved for journal write iops...
- ...which could lead to it delaying replication and stalling the cluster

Though with [adaptive-routing](https://github.com/tigerbeetle/tigerbeetle/pull/2880) that won't be so much of an issue since there are multiple paths that the prepare could take through the cluster.

src/testing/cluster/message_bus.zig
```diff
@@ -70,6 +72,15 @@ pub const MessageBus = struct {
         bus.pool.unref(message);
     }
 
+    pub fn resume_needed(bus: *MessageBus) bool {
+        return bus.suspended;
+    }
+
+    pub fn resume_receive(bus: *MessageBus) void {
```
sentientwaffle, line 79: Can we assert these transitions?

sentientwaffle (commented):
(WIP review)

src/message_buffer.zig
matklad: We definitely should log.warn whenever we suspend a message, and there's existing log for the current length of the queue if it is too long. 

I am not sure we need a metric though: given that we now log _every_ suspended message, a metric would be strictly less information. 

Or do we still want to count it just to have a redundant information flow? I don't have a good feeling for what's good metric at all, I must confess!

src/vsr/replica.zig
matklad: Yeah, I _think_ this is working as intended. A falling-behind replica _should_ try to slow-down the cluster. The cluster _should_ route-around an individual falling-behind replcia. If a quorum of replicas are slow, then the entire cluster should slow down. 

I _think_ even without adaptive routing the behavior improves: we do slow the cluster down somewhat, but that's presumably preferable to the replcia falling behind completely.

But that's a good call. I recall that I was able to reproduce dropped messages scenario rather reliably when running replicas off tmpsfs, will see if this PR makes the situation better!

src/testing/cluster/message_bus.zig
matklad: I don't think so? The contract is that you can call `.resume_receive` whenever, there isn't a wrong time to call it. And in the callback we do assert that the resume was requested. That's on the level of message bus. 

On the level of replica it is tricky. We do want to enforce that, if we suspend a message, we eventually call `resume_receive`, to ensure we don't deadlock. I don't see a super great way to solve that, but my assert in tick did in fact caught a bug where I was missing a resumption call. 

Or am I misunderstanding the kind of assertions you have in mind?

src/vsr/replica.zig
matklad: Ok, so I had trouble reproducing messages dropped due to IOPs until chaitanya pointed out that we now have bidirectional replication. If I disable that, I do observe some prepares dropped due to IOPs with the old code, and don't observe dropped prepares with this change. 

It doesnt' make much of a difference for the overall behavior though --- in my testing, some replicas enter state sync regardless.

matklad:
Started looking into what's up with Java, but probably won't solve this until tomorrow, have an IronBeetle stream coming up!

matklad:
Yup, was a real bug around client evictions: by the message buffer contract, the entire buffer must be drained, but the client breaks out of the loop if it is evicted, instead of consuming all the messages.

src/message_buffer.zig
```diff
@@ -43,6 +53,12 @@ pub const MessageBuffer = struct {
         misdirected,
     };
 
+    fn invariants(buffer: *MessageBuffer) void {
+        assert(buffer.suspend_size <= buffer.process_size);
+        assert(buffer.process_size <= buffer.advance_size);
+        assert(buffer.advance_size <= buffer.receive_size);
```
sentientwaffle, line 60: Can we assert:  
```
if (buffer.invalid) |_| {
    // sizes are zero, iterator state is idle
}
```

src/message_buffer.zig
```diff
@@ -239,24 +301,52 @@ pub const MessageBuffer = struct {
...
         return message.ref();
     }
+
+    pub fn suspend_message(buffer: *MessageBuffer, header: *const Header) void {
+        assert(buffer.iterator_state == .after_peek);
+        assert(buffer.advance_size - buffer.process_size >= header.size);
+        assert(buffer.invalid == null);
+        defer buffer.iterator_state = .after_consume_suspend;
+
+        assert(buffer.suspend_size <= buffer.process_size);
```
sentientwaffle: A sanity-check, and then let's be certain than we are suspending what we think we're suspending:

```suggestion
        defer buffer.iterator_state = .after_consume_suspend;

        assert(header.size <= constants.message_size_max);
        assert(std.mem.eql(
            u8,
            std.mem.asBytes(header),
            buffer.message.buffer[buffer.process_size..][0..@sizeOf(Header)],
        ));

        assert(buffer.suspend_size <= buffer.process_size);
```

src/message_buffer.zig
```diff
@@ -54,77 +70,56 @@ pub const MessageBuffer = struct {
...
-        buffer.consume_size = 0;
-        buffer.valid_header = false;
-        buffer.valid_body = false;
+        buffer.iterator_state = .idle;
         buffer.invalid = reason;
+        buffer.invariants();
     }
 
     /// Advances the parsing state machine.
     /// Idempotent, but eagerly called whenever available_slice changes.
```
sentientwaffle: `available_slice` doesn't exist anymore

src/message_buffer.zig
```diff
@@ -176,56 +171,123 @@ pub const MessageBuffer = struct {
...
+                return true;
+            }
+        }
+        return false;
+    }
+
+    /// MessageBuffer is also an iterator which must be driven to completion.
+    /// A call to next_header must be immediately followed by a call to consume_message
+    /// or suspend_message.
+    pub fn next_header(buffer: *MessageBuffer) ?Header {
```
sentientwaffle, line 234: I think this is worth adding as it is maybe surprising:
```suggestion
    pub fn next_header(buffer: *MessageBuffer) ?Header {
        maybe(buffer.invalid != null);
```

src/message_buffer.zig
```diff
@@ -176,56 +171,123 @@ pub const MessageBuffer = struct {
...
+        switch (buffer.iterator_state) {
+            .idle, .after_consume_suspend => {},
+            else => unreachable,
+        }
+
+        const valid_unprocessed = buffer.advance_size - buffer.process_size;
+        if (valid_unprocessed >= @sizeOf(Header)) {
```
sentientwaffle, line 243: `assert(buffer.invalid != null);`

src/message_buffer.zig
```diff
@@ -239,24 +301,52 @@ pub const MessageBuffer = struct {
...
+        assert(buffer.suspend_size <= buffer.process_size);
+
+        if (buffer.suspend_size < buffer.process_size) {
+            // Move from this:
+            // |  bytes  |    hole     |    message    |     bytes    |
+            //           ^suspend_size ^process_size                 ^receive_size
+            //
+            // To this:
+            // | bytes |    message    |     hole      |     bytes    |
+            //                         ^suspend_size   ^process_size ^receive_size
```
sentientwaffle:
```suggestion
            // Move from this:
            // |  bytes  |    hole     |    message    |     bytes    |
            //           ^suspend_size ^process_size                  ^receive_size
            //
            // To this:
            // | bytes |    message    |     hole      |     bytes    |
            //                         ^suspend_size   ^process_size  ^receive_size
```

src/message_bus.zig
```diff
@@ -344,6 +351,51 @@ fn MessageBusType(comptime process_type: vsr.ProcessType) type {
...
+            };
+            assert(bus.resume_receive_submitted);
+            bus.resume_receive_submitted = false;
+            maybe(bus.connections_suspended.empty());
+
+            // Steal the queue to avoid an infinite loop.
+            var connections_suspended = bus.connections_suspended;
+            bus.connections_suspended.reset();
+
+            while (connections_suspended.pop()) |connecton| {
```
sentientwaffle: `s/connecton/connection`

sentientwaffle (approved):

PR #2971, opened 2025-05-15
docs: small improvements to the recipes examples

snth:
A few small changes for things I noticed while reading through the recipes. They're mostly stylistic except for the balance-invariant recipe which I think was incorrect before. For the other ones I hope they make things overall more consistent and a bit easier to read.

The commit messages have some explanations of what I was trying to achieve with each edit.


docs/coding/recipes/balance-invariant-transfers.md
```diff
@@ -18,11 +26,11 @@ This recipe requires three accounts:
...
+| -: | ------------: | -------------: | -----------: | ---------: | --------------------------------------------------: |
+|  1 |        Source |    Destination |    123       |          - | [`linked`](../../reference/transfer.md#flagslinked) |
+|  2 |   Destination |        Control | `AMOUNT_MAX` |          - | [`linked`](../../reference/transfer.md#flagslinked), [`pending`](../../reference/transfer.md#flagspending), [`balancing_debit`](../../reference/transfer.md#flagsbalancing_debit) |
+|  3 |             - |              - |            - |          2 | [`void_pending_transfer`](../../reference/transfer.md#flagsvoid_pending_transfer) |
 
-When the destination account's final credits do not exceed its debits, the chain will succeed.
-When the destination account's final credits exceeds its debits, transfer `2` will fail with `exceeds_debits`.
+When the destination account's credits after transaction 1 do not exceed its debits, the chain will succeed.
+When the destination account's credits after transaction 1 exceed its debits, transfer `2` will fail with `exceeds_debits`.
```
sentientwaffle, line 36: Good catch!

docs/coding/recipes/balance-invariant-transfers.md
```diff
@@ -18,11 +26,11 @@ This recipe requires three accounts:
...
-|  1 |        Source |    Destination |    123 |          - | [`linked`](../../reference/transfer.md#flagslinked) |
-|  2 |   Destination |        Control |      1 |          - | [`linked`](../../reference/transfer.md#flagslinked), [`pending`](../../reference/transfer.md#flagspending), [`balancing_debit`](../../reference/transfer.md#flagsbalancing_debit) |
-|  3 |             - |              - |      0 |          2 | [`void_pending_transfer`](../../reference/transfer.md#flagsvoid_pending_transfer) |
+| Id | Debit Account | Credit Account |       Amount | Pending Id |                                               Flags |
+| -: | ------------: | -------------: | -----------: | ---------: | --------------------------------------------------: |
+|  1 |        Source |    Destination |    123       |          - | [`linked`](../../reference/transfer.md#flagslinked) |
+|  2 |   Destination |        Control | `AMOUNT_MAX` |          - | [`linked`](../../reference/transfer.md#flagslinked), [`pending`](../../reference/transfer.md#flagspending), [`balancing_debit`](../../reference/transfer.md#flagsbalancing_debit) |
```
sentientwaffle, line 32: I think this was already correct with `amount=1`. (`AMOUNT_MAX` works as well, but I think `amount=1` is more clear about the intent.)

The second transfer succeeds if and only if its amount is reduced to `0` (by `balancing_debit`). If the second transfer's final amount is nonzero, then the control account would end up with `debits=0 credits={amount}` (since it initially had `0` credits), which would break its `credits_must_not_exceed_debits` invariant.

See also https://github.com/tigerbeetle/tigerbeetle/blob/444f65b4ec074a36b6030c872a486806bb4521f4/src/state_machine_tests.zig#L1794-L1796 for a test which implements this pattern.

matklad:
Merged as a part of 3186 and released already:

https://github.com/tigerbeetle/tigerbeetle/releases/tag/0.16.55#:~:text=3186

Thanks for the PR and sorry that it took us so long to respond!

PR #2973, opened 2025-05-16
bus: introduce ReceiveBuffer

matklad:
This is mostly behavior preserving refactoring which extracts "parsing" related  parts of MessageBus into a sans-io ReceiveBuffer. As a result, now the replica the drives parsing loop, allowing for batch processing (prioritization & selective back-pressure), although the actual changes to processing are left for the follow-up.

This seems like a good refactor anyway, as it allows us to fuzz this component easily, and reduces the amount of code in the MessageBus, which is one of our more complicated components.

### DIFF
--- a/src/testing/cluster.zig
+++ b/src/testing/cluster.zig
@@ -124,7 +124,7 @@ pub fn ClusterType(comptime StateMachineType: anytype) type {
         replica_count: u8,
         standby_count: u8,
 
-        clients: []Client,
+        clients: []?Client,
         client_pools: []MessagePool,
         /// Updated when the *client* is informed of the eviction.
         /// (Which may be some time after the client is actually evicted by the cluster.)
@@ -285,11 +285,11 @@ pub fn ClusterType(comptime StateMachineType: anytype) type {
             @memset(client_eviction_reasons, null);
 
             const client_id_permutation = IdPermutation.generate(&prng);
-            var clients = try allocator.alloc(Client, options.cluster.client_count);
+            var clients = try allocator.alloc(?Client, options.cluster.client_count);
             errdefer allocator.free(clients);
 
             for (clients, 0..) |*client, i| {
-                errdefer for (clients[0..i]) |*c| c.deinit(allocator);
+                errdefer for (clients[0..i]) |*c| c.*.?.deinit(allocator);
                 client.* = try Client.init(
                     allocator,
                     .{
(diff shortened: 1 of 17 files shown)


src/receive_buffer.zig
```diff
@@ -0,0 +1,371 @@
...
+        if (!header.valid_checksum()) {
+            buffer.invalidate(.header_checksum);
+            return;
+        }
+
+        buffer.valid_header = true;
+
+        // To avoid dropping outgoing messages, set the peer as soon as we can,
+        // and not when we receive a full message.
+        const message_peer = header.peer_type();
```
matklad: Ideally, I'd love to do this inside Replica.on_messages, so that ReceiveBuffer doesn't have to know about the content of the messages at all, but the problem with that is that, by the time on_messages returns and the bus gets a chance to setup connection's peer, the replica will have already sent response messages!

src/vsr/client.zig
```diff
@@ -188,8 +189,21 @@ pub fn ClientType(
...
             const self: *Client = @fieldParentPtr("message_bus", message_bus);
+            while (buffer.consume_message(self.message_bus.pool)) |message| {
+                defer self.message_bus.unref(message);
+
+                if (message.header.cluster != self.cluster) {
+                    buffer.invalidate(.header_cluster);
+                    return;
+                }
+                self.on_message(message);
+                if (self.evicted) break;
```
matklad, line 202: I _think_ it was actually possible to hit the `assert(!self.evicted)` assert in `on_message` before --- production message bus _can_  push several messages onto a client in one go.

src/testing/cluster.zig
```diff
@@ -460,6 +460,13 @@ pub fn ClusterType(comptime StateMachineType: anytype) type {
                 var advanced = false;
                 advanced = cluster.network.step() or advanced;
 
+                for (cluster.clients, cluster.client_eviction_reasons) |*client, eviction_reason| {
+                    if (client.* != null and eviction_reason != null) {
+                        client.*.?.deinit(cluster.allocator);
+                        client.* = null;
+                    }
+                }
```
matklad, line 468: We no longer can `client.deinit` right from the eviction callback, so that now sets a flag (via eviciton_reason), and the actual deinit happens here.

sentientwaffle (approved):

PR #2979, opened 2025-05-19
ci: speedup lsm_scan_fuzz

matklad:
It runs out of 30min budget on CFO.

The fuzzer is slow because it is quadratic worst-case: every commit adds a number of items to the database, and each query is linear in the database.

I don't want to reduce the number of commits, as we want to fill LSM levels. So instead I just reduce the frequency of queries: we'll still be probabilistically calling every query at every point!

The actual formula to compute query probability is hairy, and I stopped understanding it the moment I typed the last closing `)`. I do trust the comptime tests though that it is not always returning zero!

Seed: ./zig/zig build -Drelease fuzz -- lsm_scan 1112800611640318623

### DIFF
--- a/src/lsm/scan_fuzz.zig
+++ b/src/lsm/scan_fuzz.zig
@@ -6,6 +6,7 @@ const constants = @import("../constants.zig");
 const fuzz = @import("../testing/fuzz.zig");
 const stdx = @import("../stdx.zig");
 const vsr = @import("../vsr.zig");
+const Ratio = stdx.PRNG.Ratio;
 const ratio = stdx.PRNG.ratio;
 
 const log = std.log.scoped(.lsm_scan_fuzz);
@@ -567,6 +568,7 @@ const Environment = struct {
         storage: *Storage,
         prng: *stdx.PRNG,
         commits_max: u32,
+        query_chance: Ratio,
     ) !void {
         assert(commits_max > 0);
         log.info("commits = {}", .{commits_max});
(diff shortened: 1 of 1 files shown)


src/lsm/scan_fuzz.zig
```diff
@@ -640,9 +642,11 @@ const Environment = struct {
             }
             try env.commit();
 
-            for (&query_specs, &env.model_matches) |*query_spec, *query_matches| {
-                const query_results_count = try env.run_query(query_spec, query_matches);
-                assert(query_results_count == query_matches.count()); // Sanity-check.
+            if (env.prng.chance(query_chance)) {
```
batiati, line 645: Overall this makes sense,
I think we could either gate individual `query_specs` (i.e., push the `if` down), or at least ensure that we always run the queries after the last commit.

It feels wrong to skip queries after commits, since that's the point of running the same query multiple times against an ever-changing LSM.

matklad: Makes sense to run queries once more at the end, to make sure there are no wasted commits, added that. 

> It feels wrong to skip queries after commits, since that's the point of running the same query multiple times against an ever-changing LSM.

It does feel right to me! Commits change the data across two dimensions:

1) One specific commits might create some strange LSM tree state that would make query to fail, but that state is transient and would be _fixed_ by subsequent commits (eg, something about queries not handling table memory correct)
2) The aggregate of all commits just pushing enough data onto the LSM tree to create an interesting level structure. 

To catch 1), you ideally do want to run queries after every commit. But if you do that, you can't have too many commits, and if you have too few commits, you compromise 2. 

But running queries probabilistically allows us to have both 1 and two. Suppose that a fuzzer fails if run after every commit on a particular seed. Then, with randomization, the probability that it still fails is only 10% --- we might miss just the right moment to do the query. But we still have 10% probability of failure, so we'll still hit this case eventually. _And_ because we are running queries only 10% of the time, we can now push 10x more data into our LSM tree, to uncover more error of type-2. 

I actually think that this might be a general pattern fro fuzzers: 

If you want to speed up fuzzer, don't reduce the event count, reduce the frequency of checks. You want _fewer_ checks on _larger_ state, as the number of checks is something you can rack up by just running more instances of the fuzzer, but the state size is what it is, it can't go larger than events_max.

batiati (approved):

PR #2980, opened 2025-05-19
VSR/Grid: Fix release assert in `fulfill_block`

sentientwaffle:
We may receive a block (that we didn't ask for) which is from a newer release than we are on.

VOPR found this on my local branch; I'm not sure why it doesn't ever find this on main.


matklad:
>We may receive a block (that we didn't ask for)

I think we don't do misdirects or anything that can cause that to happen in VOPR? We probably should though.

matklad (approved):

sentientwaffle:
Ahh right, that explains it. In my local branch it is possible for a replica to backtrack its release due to reformatting.

PR #2981, opened 2025-05-22
vopr: assert that replicas agree on commits

matklad:
StateChecker tracks the sequence of commits made logically by the cluster, by observing individual's replicas commits and advancing the watermark.

Unless I am mistaken, if the water mark is not advanced (that is, a different replcia committed first), we _don't_ check that we commit the same prepare. This feels like an oversight.

### DIFF
--- a/src/testing/cluster/state_checker.zig
+++ b/src/testing/cluster/state_checker.zig
@@ -184,8 +184,8 @@ pub fn StateCheckerType(comptime Client: type, comptime Replica: type) type {
             const commit_b = replica.commit_min;
 
             const header_b = replica.journal.header_with_op(replica.commit_min);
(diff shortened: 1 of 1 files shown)


chaitanyabhandari (approved):
Nice catch! And even nicer simplification to the below, the old condition was _much_ harder for me to understand.

https://github.com/tigerbeetle/tigerbeetle/blob/6f389f6c7710234df897a6fc7708482ee7cdc61b/src/testing/cluster/state_checker.zig#L187-L188

matklad:
>And even nicer simplification to the below, the old condition was much harder for me to understand.

Yeah, some years ago we flipped our idiom for `assert(an implies b)` from `assert(!a or b)` to `if (a) assert(b)`, but we have some old-form leftovers left. 

should actually blog about it...

PR #2982, opened 2025-05-22
aof: single code path for production & testing

chaitanyabhandari:
This PR parametrizes the AOF by IO, so that the test AOF and the production AOF exercise the same code path (barring the IO of course). The goal is to make sure the production AOF code is rigorously fuzzed, and not just tested using basic tests!

We _still_  use [writeAll](https://github.com/ziglang/zig/blob/0.13.0/lib/std/fs/File.zig#L1284) & [readAll](https://github.com/ziglang/zig/blob/0.13.0/lib/std/fs/File.zig#L1142) to write to & read from the AOF, but they've now been moved to `io/common.zig` under `write_blocking` & `read_blocking`, which are in turn invoked by `io/linux.zig`, `io/darwin.zig`, and `io/windows.zig`.

```C
pub fn write_blocking(fd: posix.fd_t, buffer: []const u8) posix.WriteError!void {
    const file = std.fs.File{ .handle = fd };
    return file.writeAll(buffer);
}
```

Similarly, the logic from the test AOF for writing to the in-memory AOF has been moved to `testing/io.zig` under `write_blocking` & `read_blocking`. 
```C
    pub fn write_blocking(self: *IO, fd: fd_t, source: []const u8) posix.WriteError!void {
        assert(fd < self.files.len);

        const file_index = @as(u32, @intCast(fd));
        const target = self.files[file_index].buffer;
        const offset = self.files_offsets[file_index];

        assert(offset + source.len <= target.len);

        stdx.copy_disjoint(.exact, u8, target[offset..][0..source.len], source);

        self.files_offsets[file_index] += source.len;
    }
```

This _has_ led to some code duplication across the three platform-specific IO files which I am not particularly thrilled about, but IIUC we have no choice if we want to make sure we store that the AOF is parametrized with the _IO_ & _file descriptor_ so it can be used for both test & production AOF. The reason is that while the test IO _does_ use the _file descriptor_ to index into the in-memory buffer to write/read into, the production IO needn't do that!


src/io/darwin.zig
```diff
@@ -1017,4 +1017,19 @@ pub const IO = struct {
             else => |e| e,
         };
     }
+
+    pub fn write_now(_: *IO, fd: posix.fd_t, buffer: []const u8) posix.WriteError!void {
```
matklad: Let's call this `read_blocking`. The `_now` convention is four functions that can return `WouldBlock / null`

src/io/common.zig
```diff
@@ -103,3 +103,18 @@ pub fn tcp_options(
...
+
+pub fn read_blocking(fd: posix.fd_t, buffer: []u8, offset: u64) posix.PReadError!usize {
+    const file = std.fs.File{ .handle = fd };
+    return file.preadAll(buffer, offset);
+}
+
+pub fn close_blocking(fd: posix.fd_t) void {
+    const file = std.fs.File{ .handle = fd };
+    file.close();
+}
```
cb22: Hmmm, it would be nice to make it clear that these (and below) _only_ exist for the AOF and have no place being used in any other code.

We could break naming, and go with something like `aof_blocking_{write,read,close}`?

src/testing/aof.zig
```diff
@@ -1,156 +0,0 @@
-const std = @import("std");
```
cb22, line 1: :tada:

src/testing/cluster.zig
```diff
@@ -355,14 +351,56 @@ pub fn ClusterType(comptime StateMachineType: anytype) type {
...
+                    constants.sector_size,
+                    // Arbitrary value.
+                    32 * 1024 * 1024,
+                );
+                errdefer allocator.free(buffer);
+
+                aof_io_file[0] = .{ .buffer = buffer };
+                aof_io.* = try IO.init(aof_io_file, allocator, .{
+                    .seed = options.cluster.seed,
+                    .larger_than_logical_sector_read_fault_probability = ratio(
```
cb22: Perhaps, let's make a `Ratio.zero()` constructor or similar that returns this - I had to do a double check when reading it as to what exactly it's doing :smile:.

src/testing/cluster.zig
```diff
@@ -505,7 +551,12 @@ pub fn ClusterType(comptime StateMachineType: anytype) type {
                     switch (health.*) {
                         .up => |up| {
                             assert(!up.paused);
+
                             replica.tick();
+                            aof_io.run() catch |err| {
+                                log.err("{}: io.run() failed: error={}", .{ replica.replica, err });
```
cb22: `std.debug.panic`?

src/testing/io.zig
```diff
@@ -33,19 +34,31 @@ pub const IO = struct {
     const Queue = QueueType(Completion);
 
     files: []const File,
+
+    // Maintained for appending to the end of the file via `write_blocking`.
+    files_offsets: []usize,
```
cb22:
```suggestion
    files_offsets: []u64,
```

(or `u32`!)

src/testing/io.zig
```diff
@@ -220,4 +236,67 @@ pub const IO = struct {
...
+            context,
+            callback,
+            completion,
+            .fsync,
+            .{ .fd = fd },
+            struct {
+                fn do_operation(io: *IO, op: anytype) FsyncError!void {
+                    _ = io; // autofix
+                    _ = op; // autofix
+                }
```
cb22:
```suggestion
                fn do_operation(_: *IO, _: anytype) FsyncError!void {
                }
```

src/testing/io.zig
```diff
@@ -220,4 +236,67 @@ pub const IO = struct {
...
+
+        self.files_offsets[file_index] += source.len;
+    }
+
+    pub fn close_blocking(_: *IO) void {}
+
+    pub fn read_blocking(self: *IO, fd: fd_t, target: []u8, offset: u64) posix.PReadError!usize {
+        assert(fd < self.files.len);
+
+        const source = self.files[@as(u32, @intCast(fd))].buffer;
```
cb22: Perhaps split out like `file_index` above to make it clearer.

src/aof.zig
```diff
@@ -6,20 +6,20 @@ const constants = @import("constants.zig");
...
-const StateMachine = vsr.state_machine.StateMachineType(Storage, constants.state_machine_config);
 
-const Header = vsr.Header;
-const Client = vsr.ClientType(StateMachine, MessageBus, vsr.time.Time);
 const log = std.log.scoped(.aof);
 
 const magic_number: u128 = 0xbcd8d3fee406119ed192c4f4c4fc82;
 
+const IOTest = @import("testing/io.zig").IO;
+const IOProd = @import("io.zig").IO;
```
cb22: Is there a better 4 letter word than `Prod` that we can use?

src/testing/io.zig
```diff
@@ -220,4 +236,67 @@ pub const IO = struct {
...
+        const offset = self.files_offsets[file_index];
+
+        assert(offset + source.len <= target.len);
+
+        stdx.copy_disjoint(.exact, u8, target[offset..][0..source.len], source);
+
+        self.files_offsets[file_index] += source.len;
+    }
+
+    pub fn close_blocking(_: *IO) void {}
```
cb22: We could track if a file is closed in its struct, and set that here. Then, on `deinit`, assert all files are closed!

src/io/common.zig
cb22: Also, we should make it clear it's `readAll` and `writeAll` semantics in the name!

src/testing/cluster.zig
chaitanyabhandari: I really like the `Ratio.zero` idea! 🎉

src/testing/io.zig
chaitanyabhandari: Ah yes, great idea!

src/io/common.zig
chaitanyabhandari: Ack regarding retaining `readAll` and `writeAll` semantics in the name!

> Hmmm, it would be nice to make it clear that these (and below) only exist for the AOF and have no place being used in any other code.

Mhmmmm good point... I was wondering if they'd also be useful in other (not core database) code that we write, but I believe we can just use `std.fs.File` _directly_ there (the way we are already in `multiversion.zig`)! Let me change them to `aof_blocking_pread_all`, `aof_blocking_write_all`, and  `aof_blocking_close`.

src/aof.zig
chaitanyabhandari: Mhmmmmmm, I did go back and forth on this naming for a bit and reverted to an old one that I had rejected: `const IOProd = @import("io.zig").IO` → `const IO = @import("io.zig").IO`, and `AOFType(comptime IO: type)` → `AOFType(comptime _IO: type)`. 

Thoughts?

chaitanyabhandari:
@cb22 Comments addressed in 58463328d6d9ba618cc743fec0aff165a1504786!

src/aof.zig
cb22: Ah I see. Here's an idea, let me know if this is workable - the only things that require those top level types (I _think_) are the aof_* functions and `AOFReplayClient`. We could move the aof_* functions under `AOFReplayClient`, and get rid of the global IO imports entirely - scope it under `AOFReplayClient` where needed, and then we can have `AOFType(comptime IO: type)` since there'll be no conflict.

Another option, and one I'm not sure about, is how weirdly we deal with IO itself. Unlike everything else that's an explicit type parameter, IO is kinda this implicitly defined one, based on OS, but with an explicit testing variant. Maybe we could unify those somehow :thinking:.

chaitanyabhandari: Ah, this is _indeed_ confusing! The top-level type is actually used in `aof_merge`, `aof_iterator`, `AOFReplayClient`, and `test "aof write / read"`. Specifically, we use it to initialize `var io = try IO.init(32, 0);` in `main` and then pass the IO instance to `aof_merge`, `aof_iterator`, _and_ `AOFReplayClient`!

I'm thinking about ways to simplify this...

chaitanyabhandari:
@cb22 took a stab at a refactor that removes the top-level IO imports, in https://github.com/tigerbeetle/tigerbeetle/pull/2982/commits/eff71e42fbf72029587df33a7b8dea3cb38058b3! Thoughts? 😄

cb22 (dismissed):
Looks great!

chaitanyabhandari:
@cb22 could you please re-approve? DJ pointed out that due to a compile failure in the scan fuzzer, this PR was breaking the CFO. We then discovered that our CI does not run `./zig/zig build fuzz -- smoke` anymore! I will raise a separate PR to add it to the `./zig/zig build ci -- smoke` step of the CI, so we can detect regressions in the fuzzers.

For now, I have pushed a fix for the fuzz failures, and ran `./zig/zig build fuzz -- smoke` locally (which succeeded):
```C
info(fuzz): done in 45.308s
```

cc: @sentientwaffle

cb22 (approved):

PR #2988, opened 2025-05-26
ci: fix release_validate to run on release

matklad:
It's important that release validation uses releases's own code. That used to be just the `release` branch, but, with two-phase releases, this is no longer true, so we need to check the tag specifically.

### DIFF
--- a/.github/workflows/release_validate.yml
+++ b/.github/workflows/release_validate.yml
@@ -20,7 +20,13 @@ jobs:
     steps:
       - uses: actions/checkout@v4
         with:
-          ref: release # Use the 'release' branch even if triggered by cron.
+          ref: release
+      # Run on the latest tag. The release branch is ahead of release tag over the weekend.
+      - run: |
+          TAG=$(gh release list --limit 1 | cut -f1)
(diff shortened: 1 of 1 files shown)


matklad:
Ok, it seems to work:

https://github.com/tigerbeetle/tigerbeetle/actions/runs/15258907812/job/42912794189

.github/workflows/release_validate.yml
```diff
@@ -20,7 +20,13 @@ jobs:
     steps:
       - uses: actions/checkout@v4
         with:
-          ref: release # Use the 'release' branch even if triggered by cron.
+          ref: release
+      # Run on the latest tag. The release branch is ahead of release tag over the weekend.
+      - run: |
+          TAG=$(gh release list --limit 1 | cut -f1)
```
matklad, line 26: Ideally, we'd pass that as a `ref` above, but I couldn't get it to work, and the docs are strategically ambiguous regrarding whether that's possible at all.

matklad:
This _hasn't_ bit us yet, as usually version skew here is not problematic, but it will fire eventually!

cb22 (approved):

PR #2989, opened 2025-05-26
ci: check release determinism and remove dependencies on `zip` & `unzip`

cb22:
This has been there for ages, but not enabled! #2988 reminded me of it again.

In summary, rebuild the release zip when validating the release, to ensure the artifact's checksum matches the freshly built checksum.

### DIFF
--- a/src/scripts/ci.zig
+++ b/src/scripts/ci.zig
@@ -114,18 +114,17 @@ fn validate_release(shell: *Shell, gpa: std.mem.Allocator, language_requested: ?
         // Zig only guarantees release builds to be deterministic.
         if (std.mem.indexOf(u8, artifact, "-debug.zip") != null) continue;
 
-        // TODO(Zig): Determinism is broken on Windows:
-        // https://github.com/ziglang/zig/issues/9432
-        if (std.mem.indexOf(u8, artifact, "-windows.zip") != null) continue;
-
         const checksum_downloaded = try shell.exec_stdout("sha256sum {artifact}", .{
             .artifact = artifact,
         });
 
         shell.popd();
-        const checksum_built = try shell.exec_stdout("sha256sum zig-out/dist/tigerbeetle/{artifact}", .{
-            .artifact = artifact,
-        });
+        const checksum_built = try shell.exec_stdout(
+            "sha256sum zig-out/dist/tigerbeetle/{artifact}",
+            .{
+                .artifact = artifact,
+            },
+        );
         try shell.pushd_dir(tmp_dir.dir);
 
         // Slice the output to suppress the names.
--- a/src/scripts/ci.zig
+++ b/src/scripts/ci.zig
@@ -138,7 +138,10 @@ fn validate_release(shell: *Shell, gpa: std.mem.Allocator, language_requested: ?
         }
     }
 
-    try shell.exec("unzip tigerbeetle-x86_64-linux.zip", .{});
+    const zip_file = try std.fs.cwd().openFile("tigerbeetle-x86_64-linux.zip", .{});
+    defer zip_file.close();
+    try std.zip.extract(std.fs.cwd(), zip_file.seekableStream(), .{});
+
     const version = try shell.exec_stdout("./tigerbeetle version --verbose", .{});
     assert(std.mem.indexOf(u8, version, tag) != null);
     assert(std.mem.indexOf(u8, version, "ReleaseSafe") != null);
--- a/src/scripts/ci.zig
+++ b/src/scripts/ci.zig
@@ -142,7 +142,7 @@ fn validate_release(shell: *Shell, gpa: std.mem.Allocator, language_requested: ?
         }
     }
 
-    try shell.zip_tigerbeetle_extract("tigerbeetle-x86_64-linux.zip");
+    try shell.unzip_executable("tigerbeetle-x86_64-linux.zip", "tigerbeetle");
 
     const version = try shell.exec_stdout("./tigerbeetle version --verbose", .{});
     assert(std.mem.indexOf(u8, version, tag) != null);
--- a/src/scripts/ci.zig
+++ b/src/scripts/ci.zig
@@ -137,7 +137,7 @@ fn validate_release(shell: *Shell, gpa: std.mem.Allocator, language_requested: ?
         }
     }
 
-    try shell.extract_tigerbeetle_zip("tigerbeetle-x86_64-linux.zip");
+    try shell.zip_tigerbeetle_extract("tigerbeetle-x86_64-linux.zip");
 
     const version = try shell.exec_stdout("./tigerbeetle version --verbose", .{});
     assert(std.mem.indexOf(u8, version, tag) != null);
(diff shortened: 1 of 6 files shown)


matklad:
ahaha, I actually filed #2988 exactly because I also was trying to fix this

src/scripts/ci.zig
```diff
@@ -105,51 +105,45 @@ fn validate_release(shell: *Shell, gpa: std.mem.Allocator, language_requested: ?
...
-    // try shell.exec_zig("build scripts -- release --build  --run-number={run_number} " ++
-    //     "--sha={sha} --language=zig", .{
-    //     .run_number = run_number,
-    //     .sha = sha,
-    // });
-    // for (artifacts) |artifact| {
-    //     // Zig only guarantees release builds to be deterministic.
-    //     if (std.mem.indexOf(u8, artifact, "-debug.zip") != null) continue;
-
-    //     // TODO(Zig): Determinism is broken on Windows:
```
cb22, line 128: This issue has been closed, yay!

src/scripts/ci.zig
```diff
@@ -105,51 +105,45 @@ fn validate_release(shell: *Shell, gpa: std.mem.Allocator, language_requested: ?
...
+            "sha256sum zig-out/dist/tigerbeetle/{artifact}",
+            .{
+                .artifact = artifact,
+            },
+        );
+        try shell.pushd_dir(tmp_dir.dir);
+
+        // Slice the output to suppress the names.
+        if (!std.mem.eql(u8, checksum_downloaded[0..64], checksum_built[0..64])) {
+            // Still run the code, but suppress failing until a release cycle has taken place.
```
cb22, line 132: NB: Change this to a panic once the next release, built with the new Zip code, has gone out.

cb22 (commented):
Trying to get `zip` to generate deterministic output was more painful than writing our own! As a side effect, we now no longer depend on the zip & unzip CLI utils for building, which is nice.

src/shell.zig
```diff
@@ -924,3 +924,131 @@ pub fn iso8601_to_timestamp_seconds(shell: *Shell, datetime_iso8601: []const u8)
...
+pub fn zip_tigerbeetle_create(
+    shell: *Shell,
+    zip_file: std.fs.File,
+    exe_path: []const u8,
+    max_size: u64,
+) !void {
+    var zip_file_writer = std.io.countingWriter(zip_file.writer());
+
+    const tigerbeetle = try shell.cwd.readFileAlloc(shell.arena.allocator(), exe_path, max_size);
+    const tigerbeetle_dos_mtime = unix_to_dos_timestamp((try shell.cwd.statFile(exe_path)).mtime);
```
matklad: I don't think it's worth it to preserve any extra metadata here. TigerBeetle binary is content-defined, `./tigerbeetle version` should tell you anything you know. So let's hard-code this to something like 1 Jan 2001 or something (I think using UNIX epoch here might break? Only vague memories!)

src/shell.zig
```diff
@@ -924,3 +924,131 @@ pub fn iso8601_to_timestamp_seconds(shell: *Shell, datetime_iso8601: []const u8)
         .{ .datetime_iso8601 = datetime_iso8601 },
     ), 10);
 }
+
+/// Extracts a TigerBeetle zip and ensures the output has correct permissions.
+pub fn zip_tigerbeetle_extract(shell: *Shell, relative_path: []const u8) !void {
```
matklad: Let's keep this non-too-specialized for tigerbeetle! I maybe want to move `shell.zig` to `stdx`. 

So maybe

```
pub fn unzip_executable(shell: *Shell, zip: []const u8, executable_name: []const u8)
```

and 

```
pub fn zip_executable
```

src/shell.zig
```diff
@@ -924,3 +924,131 @@ pub fn iso8601_to_timestamp_seconds(shell: *Shell, datetime_iso8601: []const u8)
...
+        .signature = std.zip.local_file_header_sig,
+        .version_needed_to_extract = zip_version_20,
+        .flags = .{ .encrypted = false, ._ = 0 },
+        .compression_method = .deflate,
+        .last_modification_time = tigerbeetle_dos_mtime.time,
+        .last_modification_date = tigerbeetle_dos_mtime.date,
+        .crc32 = crc32,
+        .compressed_size = @intCast(tigerbeetle_deflated.len),
+        .uncompressed_size = @intCast(tigerbeetle.len),
+        .filename_len = @intCast(exe_path.len),
```
matklad: Aren't we leaking the entire path here? Shouldn't we take basename at least, or maybe explicitly pass in the file name as an argument?

src/shell.zig
```diff
@@ -924,3 +924,131 @@ pub fn iso8601_to_timestamp_seconds(shell: *Shell, datetime_iso8601: []const u8)
...
+
+pub fn zip_tigerbeetle_create(
+    shell: *Shell,
+    zip_file: std.fs.File,
+    exe_path: []const u8,
+    max_size: u64,
+) !void {
+    var zip_file_writer = std.io.countingWriter(zip_file.writer());
+
+    const tigerbeetle = try shell.cwd.readFileAlloc(shell.arena.allocator(), exe_path, max_size);
```
matklad: We are _inside_ shell, so it makes sense to put a bit more effort to avoid space leaks. Use `shell.gpa` for allocating buffers

src/shell.zig
cb22: We could, but not currently since `exe_path` is actually `exe_name` which is just `tigerbeetle{.exe}`. But I'll split it as a separate parameter.

src/shell.zig
cb22: > I don't think it's worth it to preserve any extra metadata here.

I thought so too, but then running `ls -l` when `tigerbeetle` is in a directory looks ugly! How about making it explicitly passed in and not relying on `mtime`?

matklad: 👍

src/shell.zig
```diff
@@ -924,3 +924,141 @@ pub fn iso8601_to_timestamp_seconds(shell: *Shell, datetime_iso8601: []const u8)
...
+
+    return .{ .time = time, .date = date };
+}
+
+pub fn zip_executable(
+    shell: *Shell,
+    zip_file: std.fs.File,
+    input: struct {
+        executable_name: []const u8,
+        executable_mtime: stdx.Instant,
```
matklad: Instant would be a "wrong" type to use here. Instant is intentionally not anchored at any given point in time: comparing two instants makes sense, but absolute value of `.ns` is somewhat meaningless. Here, we want a unix timestamp, we don't have a dedicated type for that, so I suggest sticking with `u64` nanoseconds, what we do for Transfers and Accounts.

cb22: Makes sense - I missed the intention behind `Instant`.

I'm hesitant to use a raw numerical type directly, since I've introduced enough bugs doing that :smile:.

Especially because there are two layers of conversion present (git giving timestamp in seconds). I went with threading DateTimeUTC through directly...

matklad: Oh, right, that's the perfect type here actually!

matklad (approved):

PR #2990, opened 2025-05-27
stdx: refine BoundedArrayType API

matklad:
Rename append_assume_capacity to push, to be shorter and consistent with Queue&Stack.

Also make it return `?void`, so that the caller has to `.?` explicitly, signaling a pre-condition here.

I am a bit torn on void+callee assert over void?+caller .?, but I prefer either to `assume_capacity`, and doing from caller to callee is the easier future move.

### DIFF
--- a/src/build_multiversion.zig
+++ b/src/build_multiversion.zig
@@ -588,13 +588,13 @@ fn build_multiversion_body(shell: *Shell, options: struct {
             break :blk offset;
         };
 
-        releases.append_assume_capacity(past_release);
-        checksums.append_assume_capacity(past_checksum);
-        offsets.append_assume_capacity(offset);
-        sizes.append_assume_capacity(past_size);
-        flags_.append_assume_capacity(past_flag);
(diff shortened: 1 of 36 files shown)


batiati:
> I am a bit torn on void+callee assert over void?+caller .?, but I prefer either to assume_capacity, and doing from caller to callee is the easier future move.

I think a `?void` return like this isn't much "idiomatic"... besides we are dealing with a "BoundedArray" which makes it kind of expected that `push()` would panic if out of bounds.

I would go with `fn push() void` instead, and let the caller verify the array `len` if necessary (or maybe a `fn try_push() bool` variant).

matklad:
Makes sense, scaled this back to just a rename!

batiati (dismissed):

cb22:
Aha my AOF path beat this in :smile: let's hold off merging until tomorrow - gives the CFO more focus when `release` and `main` are pointing to the same commit!

batiati (approved):

PR #2991, opened 2025-05-27
io: reduce code duplication when erasing callback's type

matklad:


src/io/linux.zig
```diff
@@ -1888,4 +1789,26 @@ pub const IO = struct {
             }
         }
     }
+
+    fn erase_type(
+        comptime Context: type,
+        comptime Result: type,
+        comptime callback: fn (
+            context: Context,
+            comppletion: *Completion,
```
batiati: Typo: completion.

src/io/linux.zig
```diff
@@ -1888,4 +1789,26 @@ pub const IO = struct {
             }
         }
     }
+
+    fn erase_type(
```
batiati: Suggestion: Wouldn't `erase_callback` be better?

Not super strong here, but the `_type` suffix makes the reader question whether it should be named `EraseType(...)`, until realizing that despite the name, it returns a function pointer, not a type!

matklad: Went for `erase_types`. Type erasure is [a term of art](https://en.wikipedia.org/wiki/Type_erasure), and that's what we are doing here. But it indeed clashes with with our type suffix, but hopefully plural makes is clearer (we are erasing the context and the result types!)

batiati (approved):

PR #2992, opened 2025-05-27
vsr: cleanup, handle message padding uniformly

matklad:
Message that is transferred as n bytes on the wire might take up to n + (sector_padding - 1) when written to disk, because disks are block devices operating on sector granularity.

For most messages we used to fix this in the message bus, but for replies the fix was applied right before writing it to disk.

Now, all the message types are handled uniformly by the Replica's `on_messages`, where we can also guarantee that message ref-count is zero, and so it is valid to mutate the underlying buffer.

### DIFF
--- a/src/vsr/replica.zig
+++ b/src/vsr/replica.zig
@@ -1536,9 +1536,11 @@ pub fn ReplicaType(
 
                 assert(message.references == 1);
 
+                // Avoid leaking sector padding for messages written to a block device:
                 if (message.header.command == .request or
                     message.header.command == .prepare or
-                    message.header.command == .block)
+                    message.header.command == .block or
+                    message.header.command == .reply)
                 {
                     const sector_ceil = vsr.sector_ceil(message.header.size);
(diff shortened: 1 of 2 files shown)


src/vsr/replica.zig
```diff
@@ -10598,12 +10606,12 @@ pub fn ReplicaType(
...
             };
 
-            stdx.copy_disjoint(.exact, u8, request.body_used(), body);
-            @memset(request.buffer[request.header.size..vsr.sector_ceil(request.header.size)], 0);
-
             request.header.set_checksum_body(request.body_used());
             request.header.set_checksum();
 
+            stdx.copy_disjoint(.exact, u8, request.body_used(), body);
+            @memset(request.buffer[request.header.size..vsr.sector_ceil(request.header.size)], 0);
```
matklad, line 10613: Moved after setting body checksum for offensive programming: If we accidentally memset actual message body, I want the checksum to blow up.

sentientwaffle (approved):

PR #2993, opened 2025-05-27
rustclient: Add tests and crate docs

brson:
Tests for the Rust client. Also crate docs.


src/clients/rust/src/lib.rs
```diff
@@ -33,6 +33,9 @@ pub struct Client {
     client: *mut tbc::tb_client_t,
 }
 
+unsafe impl Send for Client {}
+unsafe impl Sync for Client {}
```
matklad, line 349: Good catch!

src/clients/rust/src/lib.rs
```diff
@@ -373,6 +376,7 @@ pub struct Account {
 bitflags! {
     #[repr(transparent)]
     #[derive(Copy, Clone, Debug, Default)]
+    #[derive(Eq, PartialEq, Ord, PartialOrd)]
```
matklad: Oh, love this pattern of separating out value-like derives onto a separate line!

Should we throw in `Hash` as well for good measure?

src/clients/rust/tests/tests.rs
```diff
@@ -1,33 +1,53 @@
...
 use std::process::{Child, Command, Stdio};
+use std::sync::{Arc, Barrier};
+use std::sync::{Mutex, MutexGuard};
+
+use futures::executor::block_on;
 
 use tigerbeetle as tb;
 
-fn random_seed() -> u64 {
+// Using the std hasher for rng to avoid the dep on rand crate
```
matklad:
```suggestion
// Using the std hasher for rng to avoid the dep on rand crate.
```

src/clients/rust/tests/tests.rs
```diff
@@ -1,33 +1,53 @@
...
+    let a = random_u64();
+    let b = random_u64();
+    let a = a as u128;
+    let b = b as u128;
+    a | b << 64
+}
+
+// Prevent multiple tests from running in parallel -
+// tests run a full tigerbeetle and that's a lot of resources.
+static TEST_MUTEX: Mutex<()> = Mutex::new(());
```
matklad: Not a fan of this, the tests should be fast! Would the following approach work for running the tests in parallel:

* There's just a single database instance per test process
* This instance is created lazily (could use https://doc.rust-lang.org/stable/std/sync/struct.OnceLock.html)
* The database is spawned with `"--addresses=0"`, so there's no need to run Drop for it, the database process will exit itself when the test process exits.
* Individual tests concurrently hit this database. 
* To make sure that individual tests do not step on each other's toes, each test generates random ids using the `id()` function. 

Ah, but we seem to miss the `id()` function? We should add one, see the JS impmlementaiton here:

https://github.com/tigerbeetle/tigerbeetle/blob/75f77b8b3280ce2f289cf42ae928945190fe4a2a/src/clients/node/src/index.ts#L161-L191

and the docs here:

https://docs.tigerbeetle.com/coding/data-modeling/#tigerbeetle-time-based-identifiers-recommended


Basically, the ID is a timestamp concatenated to a random nonce, and binary encoded in such a way that natural sorting corresponds to timestamp sorting. We do this to allow quickly checking "have we seen this ID before" in the database --- if the timestamp part is larger than any previously seen timestamp, the ID is guaranteed to be new. And, if clients have approximately correct clock, we'll always be hitting this hot path.

src/clients/rust/tests/tests.rs
```diff
@@ -302,3 +332,286 @@ fn smoke() -> anyhow::Result<()> {
...
+                            user_data_128: 0,
+                            user_data_64: 0,
+                            user_data_32: 0,
+                            reserved: tb::Reserved::default(),
+                            ledger: TEST_LEDGER,
+                            code: TEST_CODE,
+                            flags: tb::AccountFlags::History,
+                            timestamp: 0,
+                        }])
+                        .await?;
```
matklad, line 529: Another tricky, but important to cover case, is when the client is concurrently closed while there's an outstanding request in flight.

matklad (commented):
LGTM overall, but let's try to run the tests concurrently against the single instance of a database, a-la https://calpaterson.com/against-database-teardown.html

src/clients/rust/tests/tests.rs
brson: That case should not be possible in the rust client: the client cannot be dropped while requests are outstanding as requests hold a reference to the client.

The only similar case I can think of is that the rust requests can be dropped before the underlying tb_client request has completed and that's a tricky corner case that isn't tested.

brson: It could be possible to submit requests, drop their futures, drop the client, before the underlying tb_client has completed the request. I can write a test that tries to test that.

src/clients/rust/src/lib.rs
brson: I think this formatting only got through `cargo fmt` because it's in a macro...

brson: I've added the `Hash` impls.

src/clients/rust/tests/tests.rs
brson: Still working on this.

brson:
@matklad I have pushed an implementation of `id`. I have tried to make the rust version very intentional about how it handles all corner cases and it should be pretty awesome.

Still have not reworked the test suite but will.

brson:
Note that supporting dates prior to the unix epoch involves compromises - most especially the `LazyLock` around the global id generator would not be needed because it could be initialized statically if we could say timestamps equal to the unix epoch are not allowed. Could just panic in that case.

edit: well there are other ways to get rid of the `LazyLock` while still supporting dates prior to the epoch. I'll try.

edit: I removed the LazyLock.

(The ts and python impls of id don't seem to account for such dates and probably just do whatever)

src/clients/rust/tests/tests.rs
brson: Done.

src/clients/rust/tests/tests.rs
brson: Done.

brson:
@matklad I have addressed all reviews now.

brson:
I pushed a minor commit that removes `ClientStatus`. This was unused as it shouldn't be possible for the rust client to see an uninitialized/deinitialized client handle, and all cases assert this is true.

brson:
Pushed a few more tests.

brson:
@matklad I have also pushed the API docs to this branch.

brson:
note to self: i need to test that the paging algorithm i wrote in the docs actually works

edit: this is done

brson:
There's a bug here with the test db that i need to fix. I thought `--addresses=0` would tell tigerbeetle to delete the database on exit, but that is not the case. Currently the test suite leaves databases in the tmp directory.

edit: this is done

brson:
I also wrote an example of how to execute a paged range query since I found it not as obvious as the TigerBeetle docs suggest.

brson:
I pushed a commit that again changes how the test database is created. It's now at a single defined location (`target/tmp/0_0.testdb.tigerbeetle`) and just persists forever.

brson:
@matklad this branch is super done and there's a lot more to look at since last you reviewed.

brson:
after this is merged i plan to do whatever is necessary to publish the rust client crate, _then_ get the website docs/examples finished.

brson:
Ok I added one more big set of tests: that the conversions from C status codes and flags to Rust types is correct and complete. The big reason to do this is to catch future additions to the C values, as that would not otherwise trigger a compilation failure.

The cost for this though is I had to add conversions from the Rust status types back to C types. Some of these conversions are necessary for the vortex driver anyway.

src/clients/rust/src/time_based_id.rs
```diff
@@ -0,0 +1,314 @@
...
+fn random_u80() -> u128 {
+    random_u128() & U80_MASK
+}
+
+fn random_u128() -> u128 {
+    let a = random_u64();
+    let b = random_u64();
+    let a = a as u128;
+    let b = b as u128;
+    a | b << 64
```
matklad:
```suggestion
    a | (b << 64)
```

so that the reader doesn't have to wander about the order of operators in Rust.

src/clients/rust/src/time_based_id.rs
```diff
@@ -0,0 +1,314 @@
...
+}
+
+fn random_u64() -> u64 {
+    std::hash::Hasher::finish(&std::hash::BuildHasher::build_hasher(
+        &std::collections::hash_map::RandomState::new(),
+    ))
```
matklad, line 167: I am sooo waiting for Rust std gaining `getrandom`!

src/clients/rust/src/time_based_id.rs
```diff
@@ -0,0 +1,314 @@
...
+    let mut guard = GLOBAL_GENERATOR.lock().expect("global tbid generator");
+    match *guard {
+        None => {
+            *guard = Some(TbidGenerator::new());
+            drop(guard);
+            id()
+        }
+        Some(ref mut generator) => generator.next(),
+    }
+}
```
matklad, line 41: That's for a future improvement, but we might want to have something like

```rust
pub fn ids(count: usize) -> impl ExactSizeIterator<Item=u128>
```

to amortize those mutex lock/unlocks. Or maybe we don't. As I've recently realized, tigerbeetle client library isn't really the right place for id generating function, as ids should be generated end-to-end (on a mobile phone/website, rather than at the API gateway). And at the _end_ you aren't going to generate 8k transfers in one go anyway.

src/clients/rust/src/time_based_id.rs
```diff
@@ -0,0 +1,314 @@
...
+        id_prev = id_next
+    }
+
+    let id_start = id_start.unwrap();
+    let id_end = id_end.unwrap();
+
+    TestResult { id_start, id_end }
+}
+
+#[cfg(test)]
```
matklad: Slight preference for defining 

```
#[cfg(test)]
mod tests {
  fn helper() { ... }

  #[test]
  fn test() { ... }
}
```

here, to avoid individual cfg gates, and to make it abundantly clear which code relates to tests and which is production.

src/clients/rust/src/time_based_id.rs
```diff
@@ -0,0 +1,314 @@
...
+        random: 0,
+    };
+
+    let mut count = 0;
+
+    assert_ids_monotonic_and_multiple_timestamps(|| {
+        let past_time_base = now - Duration::from_millis(2);
+        let past_time = past_time_base + Duration::from_millis(count);
+        count += 1;
+        idgen.next_from_system_time(past_time)
```
matklad: Nice setup for making the code testable here!

src/clients/rust/src/conversions.rs
```diff
@@ -1,13 +1,13 @@
 pub use super::*;
 
 #[rustfmt::skip]
-impl From<tbc::tb_create_accounts_result_t> for CreateAccountResult {
-    fn from(other: tbc::tb_create_accounts_result_t) -> CreateAccountResult {
+impl From<u32> for CreateAccountResult {
```
matklad, line 4: I think we should constrain `CreateAccountResult` and other enums to only include known variants. 

We do pretty strict version checks server-side, so if we ever materialize a code that's not known to the client, that'll be a pretty big bug for us (if an old client connects to the new cluster, we'll take care to only send old codes to it). But that's not really related to the PR at hand!

matklad (dismissed):
This looks great!

I will try to do another once-over review to the whole thing, as I have a nagging feeling that _maybe_ we can simplify some types/code generation here, but that's not really related to the current changes!

matklad:
>I will try to do another once-over review to the whole thing, as I have a nagging feeling that maybe we can simplify some types/code generation here, but that's not really related to the current changes

Ok, yeah, I think we _can_ simplify things substantially. As Rafael suggested, I think we can skip the types in `tb_client.rs` and directly generate the public types in `lib.rs`, with `Reserved<N>`, `bitflags!` and what not. One danger there is that the diff between source C/Zig types and our Rust wrappers are bigger, so we run the danger of messing up the sizes, alignment and what not by accident. But _that_ I think we can combat _that_ by also generating the tests! Right now, we have a hand-written

```
    assert_eq!(
        std::mem::size_of::<Account>(),
        std::mem::size_of::<tbc::tb_account_t>()
    );
```

We could instead _generate_

```
    assert_eq!(
        std::mem::size_of::<Account>(),
        128,
    );

```

where we get `128` straight from the horses mouth, by calling `@sizeOf` in the generator on the Zig side! And the same goes for align_of and ofsset_of. 

But that's a follow up PR

src/clients/rust/src/time_based_id.rs
brson: We could also expose the generator type for the user to instantiate and synchronize (or not) however they want.

brson: Even if the servers themselves shouldn't be generating ids, the id functions have to live and be distributed somewhere. The client libraries seem as good a place as any for now if we don't want everybody reimplementing `id`.

src/clients/rust/src/time_based_id.rs
brson: Done.

src/clients/rust/src/conversions.rs
brson: I've removed the `Unknown` variants and changed the conversions to panics.

brson:
> > I will try to do another once-over review to the whole thing, as I have a nagging feeling that maybe we can simplify some types/code generation here, but that's not really related to the current changes
> 
> Ok, yeah, I think we _can_ simplify things substantially. As Rafael suggested, I think we can skip the types in `tb_client.rs` and directly generate the public types in `lib.rs`, with `Reserved<N>`, `bitflags!` and what not. One danger there is that the diff between source C/Zig types and our Rust wrappers are bigger, so we run the danger of messing up the sizes, alignment and what not by accident. But _that_ I think we can combat _that_ by also generating the tests! Right now, we have a hand-written
> 
> ```
>     assert_eq!(
>         std::mem::size_of::<Account>(),
>         std::mem::size_of::<tbc::tb_account_t>()
>     );
> ```
> 
> We could instead _generate_
> 
> ```
>     assert_eq!(
>         std::mem::size_of::<Account>(),
>         128,
>     );
> ```
> 
> where we get `128` straight from the horses mouth, by calling `@sizeOf` in the generator on the Zig side! And the same goes for align_of and ofsset_of.
> 
> But that's a follow up PR

OK, we can save it for a follow up.

I am not enthusiastic of leaning hard on code generation - I think it makes the code harder to read, document, and maintain. In general I do not like generated code being exposed in an API surface, and prefer to generate code, then wrap it with hand-written code that looks beautiful and is easy to read. I think the approach here - generating straight-forward C-like types (no bitflags, Reserved etc) then hiding them is a pleasing balance. And Rust at least has a strong enough type system that everything I can imagine breaking in the future will break at compile time or in tests.

I'm willing to try code-generating even more in the future though if asked.

brson:
I am going to remove the step here in `collect_results` that converts the abbreviated results returned by the server into full results for all input events. That's the last task here I'm aware of.

brson:
Well I pushed another commit that marks `handle_message` as unsafe since it will convert a byte buffer to a buffer of any `Copy` type, and it sure seems like there's potential for UB there, particularly wrt misalignment. Sadly this propagates outward and makes all the request method bodies completely unsafe (internally). I did not mark `create_packet` as unsafe because afaict it is not directly unsafe; it can be used though to create a packet that will crash tb_client if passed to it.

Edit: actually I've un-pushed this for now. I'm pretty confident that function _is_ unsafe, but also that I'm kinda sleepy, and that safety is maintained at the module boundary anyway so it's not super important. Could use 2nd opinion.

matklad:
>In general I do not like generated code being exposed in an API surface, and prefer to generate code, then wrap it with hand-written code that looks beautiful and is easy to read

The way I usually resolve this tradeoff is by generating code that looks beautiful and is easy to read! _Most_ generated code looks ugly, but I think that's simply because there's no cultural expectation that generated code should be as good as hand written one, but it certainly could. And I think in general it's worth putting the effort to generate readable code, because someone _will_ have to read it while debugging. In rust-analyzer, the API _surface_ for syntax trees is mostly generated, but pretty, as I have to look at it fairly frequently. 

The way I see it, the two-layer structure with a hand-written facade itself hinders readability a bit for folks who look at the source code, and now they have to think both about the facade and the generated code, there's fundamentally one more level of indirection.

The problem with generating readable code is that of course the code of the generator itself becomes harder to read (though shelling out to `rustfmt` at the end can help a lot), but I think there complexity is OK: while the generator code is complex, because you just see the results, you can treat it mostly as a black box. 

So I'd say, let's try to cut out a layer here, while _keeping_ the code easy to read? If that doesn't work out, I am fine with the present two-layer structure, but its worth trying to simplify here imo!

brson:
I pushed the change to stop materializing results.

I think it's the right direction, more TigerStyle, and points the way to some other improvements. It does make the API harder to use for many cases, and I do think the obvious thing most people are going to want to do is to iterate over results for all the inputs. I've written detailed docs about the return values of the request methods, and included copy-pastable example iterators over the materialized results.

That commit also derives more traits for all the public types, as indicated by some of the new test cases. I think I was going to have to make this change for vortex anyway.

brson:
I think it's worth trying to land this large amount of work that is a big improvement to what's in-tree right now.

But I am also considering changing both arguments to the request methods, one for performance reasons, one for ergonomic.

## Convert the `&[...]` arguments to `Vec<...>`

After removing the extra allocation and work involved in materializing the results, it makes me question whether we should also make this change for efficiency sake.

The signature of `create_accounts` looks like:

```
    pub fn create_accounts<'s>(                                                                    
        &'s self,                                                                                  
        events: &[Account],                                                                        
    ) -> impl Future<Output = Result<Vec<CreateAccountsResult>, PacketStatus>> + use<'s> { 
```

`events` is only a slice for ergonomic reasons, and the function immediately clones it. The efficient version passes ownership of `events` into `create_accounts` and then back out again:

```
    pub fn create_accounts<'s>(                                                                    
        &'s self,                                                                                  
        events: Vec<Account>,                                                                        
    ) -> impl Future<Output = (Vec<Account>, Result<Vec<CreateAccountsResult>, PacketStatus>>) + use<'s> { 
```

The return type here is pretty ugly for callers to work with.

## Don't capture the `&self` lifetime in the return future

Not having that lifetime in the return value would give callers more options for where they could store or send it. The only behavioral change would be that request futures may resolve to `ClientShutdown`.

matklad:
>Not having that lifetime in the return value would give callers more options for where they could store or send it. The only behavioral change would be that request futures may resolve to ClientShutdown.

I think that's the way to go, I think we do want to allow dropping the client with some futures in-flight!

>events is only a slice for ergonomic reasons, and the function immediately clones it.

I am wondering what would be the _most_ effective here? I guess, immediately writing to an allocated packet? Perhaps the optimal thing to do would be to accept an IntoIterator here?

brson:
> I think that's the way to go, I think we do want to allow dropping the client with some futures in-flight!

I pushed this commit too.

matklad (approved):

src/clients/rust/src/time_based_id.rs
```diff
@@ -0,0 +1,314 @@
...
+///   sequentially from the previous ID until the clock catches up with the future timestamp.
+///
+/// - The system time is prior to the Unix epoch. IDs will be generated
+///   sequentially starting at the Unix epoch (plus a base random number).
+//
+// References:
+//
+// - https://github.com/tigerbeetle/tigerbeetle/blob/75f77b8b3280ce2f289cf42ae928945190fe4a2a/src/clients/node/src/index.ts#L161-L191
+// - https://github.com/ulid/spec
+pub fn id() -> u128 {
```
michabp, line 31: Maybe this could be a newtype `TigerBeetleId(u128)` with a `next` function since calling `SystemTime::now()` is expensive and often sub-transactions with the same timestamp are being generated in the same function.

Or otherwise, an additional method could be publicly exposed to generate the next id with same timestamp from an `u128`.

Generally, the `GLOBAL_GENERATOR` Mutex gives a false sense of security while also degrading performance since id generation will usually happen on frontend as well as for some edge cases in different distributed services.

As a developer, I'd rather like to have full control over id generation. But this doesn't necessarily have to be adressed in this crate as I can just roll my own.

PR #2995, opened 2025-05-27
Ruby client

trevorrjohn:
Generate Ruby native extension from zig. 

This uses zig to generate a library that can be compiled with the ruby gem. There is a simple entry point that will load all the ruby classes and make them available to use in the ruby runtime. 

### TODOS

- [x] Build client with build.zig
- [x] Ruby gem that loads the generated library
- [x] Enums
- [x] Structs
- [x] multi-platform support for ruby headers
- [ ] Samples
- [ ] Tests
**Operations**
- [x] lookup_accounts
- [x] submit_accounts
~~- [ ] get_change_events~~
- [x] create_transfers
- [x] lookup_transfers
- [x] get_account_transfers
- [x] get_account_balances
- [x] query_accounts
- [x] query_transfers
**Client Functions**
- [x] init
- [x] submit
- [x] deinit
- [ ] register_log_callback


build.zig
```diff
@@ -1469,6 +1476,72 @@ fn build_node_client(
...
+        });
+
+        // TODO: Remove hardcoding link to Ruby library
+        if (comptime std.mem.eql(u8, platform[0], "x86_64-linux-gnu.2.27")) {
+            shared_lib.addIncludePath(.{ .cwd_relative = "/home/trevor/.asdf/installs/ruby/3.4.1/include/ruby-3.4.0" });
+            shared_lib.addIncludePath(.{ .cwd_relative = "/home/trevor/.asdf/installs/ruby/3.4.1/include/ruby-3.4.0/x86_64-linux" });
+        } else if (comptime std.mem.eql(u8, platform[0], "aarch64-macos")) {
+            shared_lib.addIncludePath(.{ .cwd_relative = "/Users/tj/.asdf/installs/ruby/3.5-dev/include/ruby-3.5.0+1" });
+            shared_lib.addIncludePath(.{ .cwd_relative = "/Users/tj/.asdf/installs/ruby/3.5-dev/include/ruby-3.5.0+1/arm64-darwin24" });
+        }
```
brson: I used the following code here and it seemed to work:

```
        const ruby_include_dirs = b.run(&.{                                                                       
            "ruby", "-rrbconfig", "-e",                                                                           
            "puts RbConfig::CONFIG.values_at(\"rubyhdrdir\", \"rubyarchhdrdir\")",                                
        });                                                                                                       
        var ruby_include_dir_iter = std.mem.tokenize(u8, ruby_include_dirs, "\n");                                
        while (ruby_include_dir_iter.next()) |dir| {                                                              
            shared_lib.addIncludePath(.{ .cwd_relative = dir });                                                  
        }  
```

brson:
I think I succeeded in building the tigerbeetle gem by doing something like

```
zig build clients:ruby
cd src/clients/ruby && gem build
```

I tried to run `src/clients/ruby/bin/console` but failed with this error:

```
/home/brian/.homes/dev/.rbenv/versions/3.4.4/lib/ruby/3.4.0/bundled_gems.rb:82:in 'Kernel.require': cannot load such file -- tigerbeetle/tigerbeetle (LoadError)
        from /home/brian/.homes/dev/.rbenv/versions/3.4.4/lib/ruby/3.4.0/bundled_gems.rb:82:in 'block (2 levels) in Kernel#replace_require'
        from /home/brian/.homes/dev/tigerbeetle2/src/clients/ruby/lib/tigerbeetle.rb:7:in '<top (required)>'
        from /home/brian/.homes/dev/.rbenv/versions/3.4.4/lib/ruby/3.4.0/bundled_gems.rb:82:in 'Kernel.require'
        from /home/brian/.homes/dev/.rbenv/versions/3.4.4/lib/ruby/3.4.0/bundled_gems.rb:82:in 'block (2 levels) in Kernel#replace_require'
        from bin/console:5:in '<main>'
```

Forgive my ignorance of ruby and please tell me if there are obvious commands I should run to exercise this code a bit.

brson:
This looks like a good start, and TigerBeetle wants a Ruby client. I encourage you to keep pushing forward on this as I try to understand the PR better.

I see there is a sizable native component in zig and that the code looks pretty tricky, so I'll need to review it carefully.

I am still reading here and trying to write some ruby code that uses it.

trevorrjohn:
> I think I succeeded in building the tigerbeetle gem by doing something like
> 
> ```
> zig build clients:ruby
> cd src/clients/ruby && gem build
> ```
> 
> I tried to run `src/clients/ruby/bin/console` but failed with this error:
> 
> ```
> /home/brian/.homes/dev/.rbenv/versions/3.4.4/lib/ruby/3.4.0/bundled_gems.rb:82:in 'Kernel.require': cannot load such file -- tigerbeetle/tigerbeetle (LoadError)
>         from /home/brian/.homes/dev/.rbenv/versions/3.4.4/lib/ruby/3.4.0/bundled_gems.rb:82:in 'block (2 levels) in Kernel#replace_require'
>         from /home/brian/.homes/dev/tigerbeetle2/src/clients/ruby/lib/tigerbeetle.rb:7:in '<top (required)>'
>         from /home/brian/.homes/dev/.rbenv/versions/3.4.4/lib/ruby/3.4.0/bundled_gems.rb:82:in 'Kernel.require'
>         from /home/brian/.homes/dev/.rbenv/versions/3.4.4/lib/ruby/3.4.0/bundled_gems.rb:82:in 'block (2 levels) in Kernel#replace_require'
>         from bin/console:5:in '<main>'
> ```
> 
> Forgive my ignorance of ruby and please tell me if there are obvious commands I should run to exercise this code a bit.

So I created a command, in `src/clients/ruby` called `./bin/setup` that will compile the ruby native extension. Then you can run `ruby test.rb` or `./bin/console` to load the gem and use the ruby console. 

```sh
zig build clients:ruby
cd src/clients/ruby && ./bin/setup
```

build.zig
trevorrjohn: Yeah I think that might be the best path, however we need to provide headers for the different versions of ruby that we want to support. This would only get whatever version of ruby you have installed. 

Long term I might include the headers in the gem and then tell zig build where to find it and compile versions that support each ruby version. I need to explore more here about what versions of ruby we should support and if there will be breaking changes that we will need to manage.

trevorrjohn: @brson I ended up having to go another route with this. Because ruby needs to be compiled to generate the headers for a platform we cannot simply copy headers over. (My belief was that ruby cannot be compiled with the zig toolchain, however I didn't try this and don't even know if it is a good idea.)

I ended up writting a shim file that would act like a header file. It is a little brittle as the function definitions have to match what ruby will define, however I believe all the functions we are using are pretty stable. It will cause segfaults if we get this wrong unfortunately.

trevorrjohn:
Closed in favor of #3034

PR #2996, opened 2025-05-27
CLI: Add `tigerbeetle recover` subcommand

sentientwaffle:
# Overview

If a replica('s data file) is permanently lost (e.g. the SSD fails) currently the only way to bring a new replica into the cluster is by running `tigerbeetle format` and allowing the new replica to state sync.

But this is unsafe -- the newly-formatted replica may nack prepares which its previous incarnation acked -- a correctness bug. (In other words, it could cause committed requests to become un-committed!)

This PR adds a new subcommand `tigerbeetle recover`, which creates a new data file (like `tigerbeetle format`) but prevents the data loss hazard.

# Details

When `tigerbeetle recover` is run:

1. The recovery process send `pipeline_prepare_queue_max` requests (noops) to the cluster.
2. Once those have committed, it creates the new data file. The data file is identical to `tigerbeetle format`'s output _except_ that `vsr_state.view == client.view + 2` (where `client.view` is the view number of the client at the end of committing the requests).
3. The recovery process exits. Now running `tigerbeetle start` as normal will work.

Explanation for `view + 2`:
- We don't want to join in the same view, since the replica might have participated in it before being lost, and we can't remember any promises we made.
- Likewise, we don't want to go to `view + 1` -- if we were the first to collect a SVC quorum before being lost, we might have sent a DVC. Since we don't remember, we must skip past `view + 1` to ensure that we don't send a _different_ DVC. (We have the invariant that if a replica sends a DVC for a given view, then all DVC's it sends for that view will be identical.)

Some other notes:

- Reformatting standbys is not implemented/permitted. Since they aren't part of consensus anyway, just use the regular `format` command.
- The cluster needs a healthy (view-change) quorum for this process to work.
- ...Therefore this won't work for 2-replica clusters. (Or single replica clusters, of course!)
- If the cluster _doesn't_ have such a quorum, it will (correctly) remain unavailable.

## Upgrades

One tricky thing here (particularly from an operational standpoint) is: when you reformat a replica what release should you tag the new data file with?

- In the vopr we can just tag with the oldest release, which always works because that release is included in all three of the releases that the vopr runs through.
- In a real cluster you wouldn't want to do that because the oldest release might not be included in the binaries that are now deployed since clusters can be very long lived.
- Ideally you would be able to tag it with the current release that is deployed to the remainder of the healthy cluster.
- But note that if a newer binary has been deployed but the replicas have not yet updated to that release, the reformat should use the release that the replicas are actually running.

## Testing

- [X] `replica_test.zig` scenarios
- [X] VOPR: Some percentage of all process restarts will reformat the replica before restarting.
- [X] Integration test (to test CLI) (Implemented in follow-up: https://github.com/tigerbeetle/tigerbeetle/pull/3018)

## Other possible approaches

My initial approach (not in this PR) for recovery was:

1. `tigerbeetle recovery` formats a data file (no request step). The superblock includes a new `recovering` flag, which is set.
2. Recovering replica waits for quorum of pings to learn view number (and thus, primary) as well as the `checkpoint_op`.
3. Recovering replica transitions to `recovering_head` for the `view+1`.
4. Send a SVC for the new view with a "force" flag. (To override the SVC quorum requirement.)
5. Wait for a SV for that (or a newer) view.
6. When we enter the new view properly, unset our `recovery` flag.

That approach would have some benefits relative to the one implemented in this PR:

- Much simpler reformatting. (In particular, reformatting is offline (like `format`).)
- Much simpler testing setup (i.e. take a look at the `src/testing/cluster.zig` changes to make online reformatting work).

...And some drawbacks:
- More complicated in the replica.
- In particular, requires multiple `view_durable_update`s as we make progress.
- Requires a new replica status.
- Requires a new flag on SVC messages to allow the recovering replica to force a view change.
- Requires a new flag in the superblock.
- Requires a new op number in the superblock. This is used to avoid nacking ops which predate our recovery. (This might have been avoided with the `view+2` idea, which I didn't figure out until later.)

I don't think either approach is an obvious winner, but I have a preference for the one implemented in this PR, as I think avoiding the extra superblock flag is worth it!


src/vsr/replica.zig
```diff
@@ -7955,6 +7957,16 @@ pub fn ReplicaType(
...
 
+            if (self.release.value < message.header.release.value and
+                self.replica == message.header.replica)
+            {
+                // Don't replica messages on a newer release than us if we were the one who
+                // originally sent it. This can happen if our release backtracked due to being
+                // reformatted.
+                log.warn("{}: replicate: ignoring prepare from newer release", .{self.replica});
+                return;
+            }
```
sentientwaffle, line 7968: This seems like a hack, but I didn't see a solution that is _less_ of a hack! The problem is that reformatting might regress our version, then we replicate a prepare which we previously originated (`message.header.replica == self.replica`). This trips:

```
            // Compare the release in this message to ours only if we authored the message.
            if (message.header.replica == self.replica) {
                assert(message.header.release.value <= self.release.value);
            }
```

in `send_message_to_replica_base`.

src/integration_tests.zig
```diff
@@ -523,184 +523,6 @@ test "in-place upgrade" {
...
-    };
-
-    const seed = std.crypto.random.int(u64);
-    log.info("seed = {}", .{seed});
-    var context = try Context.init(.{ .seed = seed });
-    defer context.deinit();
-
-    try context.run();
-}
-
```
matklad: Was this lost by accident?

src/vsr/replica.zig
matklad: >The problem is that reformatting might regress our version

Can we fix this directly?

* In the documentation, require that the operator uses the release at least as new as the one when the replica crashed.
* In the VOPR, don't restart the replica into an earlier release. 

I'd be strongly in favor of that _if_ we could assert this. But I think we can't: the release in Reply is not cluster's release, so, after the round of noops, we can't actually know which release the cluster is at. We could do best-effort check via `PongClient`, but that's not a guarantee. 

Relatedly, what happens if, due to operator's mistake, we stat on a _higher_ release? I think that's fine, right? There, we'll be using newer release for the client, and so we won't be able to commit noops. 


🤔 What if we do this: 

* leave as is for the current PR
* but in a follow up introduce `release_primary` field to `Reply` and `Prepare`
* and then add a check that, when restoring data file, we restore into the latest known release?

This is not so much to fix this issue, which I don't think is _that_ important (though, reducing dimensionality is nice), but rather to gain `release_primary` into the reply, which I _think_ might be useful to debug issues? We _promise_ to keep semantics fixed, based on client's release, but, if we do make a bug there, it surely would help to know which release actually executed the prepare?

src/integration_tests.zig
sentientwaffle: It was committed by accident, as it was nowhere near done!

matklad (dismissed):

matklad (approved):

PR #2997, opened 2025-05-27
ci: run fuzzers & vopr

chaitanyabhandari:
https://github.com/tigerbeetle/tigerbeetle/commit/1836a9231e64b6c89b32aec75cf8288c5626c3d2 removed the `./zig/zig build fuzz -- smoke` from the CI, but neither `./zig/zig build ci -- smoke` nor `./zig/zig build ci -- test` runs our fuzzers! 

This PR adds an explicit `fuzz` job to the CI, which invokes `./zig/zig build ci -- fuzz`. This would help detect regressions in the fuzzers and VOPR.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,6 +22,13 @@ jobs:
       - run: shellcheck ./zig/download.sh
       - run: ./zig/download.ps1 && ./zig/zig build ci -- smoke
 
+  fuzz:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - run: shellcheck ./zig/download.sh
+      - run: ./zig/download.ps1 && ./zig/zig build ci -- fuzz
+
   test:
     strategy:
       matrix:


sentientwaffle (approved):

matklad:
This should be in scope for

```
./zig/zig build ci -- test
```

and

```
./zig/zig build ci
```

(that is, default) 

The design principle here is that we shouldn't multiply entry points. And the `ci test` is the entry point for "running all finite-in-time" checks.

chaitanyabhandari:
Thanks for bringing this up @matklad, I _was_ a bit torn about whether this should be added as a separate step or just add it to `zig build ci -- test` or `zig build ci -- smoke`, due to a principle similar to the one you described! Ultimately, I got thrown off by the `fuzzer too slow for the smoke mode` log message thrown by the segmented array fuzzer, so I assumed _that_ might be the reason that you didn't want to add it to `zig build ci -- smoke`. In hindsight, I should've just asked you. 😄 

I raise another PR to add this to `zig build ci -- test`.

matklad:
Yeah, it's also true that fuzzers should be fast enough in smoke mode! So, if that message is printed, the right move is to spend 10 minutes profiling the fuzzer trying to find a 10x perf improvement, or otherwise just reducing the event count!

PR #2998, opened 2025-05-28
devhub: track debug build time

cb22:
We track the release build time currently, but at least personally, I only occasionally build with `-Drelease` locally. Track the debug build time too, so we know how much time we have [to get a coffee](https://xkcd.com/303/) between builds.

(unfortunately the joke doesn't land as well with Zig's fast compile times :smile:)


matklad:
Yup, would be nice to capture when Zig switches to self-hosted backend!

matklad (approved):

PR #2999, opened 2025-05-28
ci: run fuzzers with smaller stacks 

matklad:
Two changes here:

* Set stack size for vopr and fuzz to 4*MiB. Zig's default stack size is
  16*MiB, but we want to make sure that we have ample free space. Given
  that we don't have a billion of dependencies, and that recursion is
  banned by TigerStyle, our stack size should be pretty shallow.

* Eagerly fault-in the stack on startup for fuzzers, to, ahem, scare out
  the ghosts out of our CFO machines:
  https://ziggit.dev/t/stack-probe-puzzle/10291/1

  This _is_ a work-around, but, based on the fact that it _only_ affects
  forest_fuzz which does a lot of allocation and de-allocation, I do
  think there isn't any issue with our code, and, given that the root
  cause hasn't reveal itself after 3 days of investigation, it seems
  better to just table this for now.

### DIFF
--- a/src/fuzz_tests.zig
+++ b/src/fuzz_tests.zig
@@ -13,6 +13,8 @@ comptime {
     assert(constants.storage_size_limit_max == tigerbeetle_config.process.storage_size_limit_max);
 }
 
+const MiB = 1024 * 1024;
+
 pub const std_options: std.Options = .{
     .log_level = .info,
     .log_scope_levels = &[_]std.log.ScopeLevel{
@@ -56,6 +58,8 @@ const CLIArgs = struct {
 };
 
 pub fn main() !void {
+    probe_stack(3 * MiB);
+
     var gpa_allocator: std.heap.GeneralPurposeAllocator(.{}) = .{};
     const gpa = gpa_allocator.allocator();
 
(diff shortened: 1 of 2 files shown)


src/fuzz_tests.zig
```diff
@@ -56,6 +58,8 @@ const CLIArgs = struct {
 };
 
 pub fn main() !void {
+    probe_stack(3 * MiB);
```
matklad, line 61: Proof of compiler's work to check stack: 

https://godbolt.org/z/Ez9r69rW6

matklad:
I've also added RAM limiting here, as that's obviously a great idea, but 30 seconds later it didn't seem _obviously_ good to me, so I backed it out!

src/fuzz_tests.zig
cb22: We set the stack size to 4MiB and only probe 3MiB?

matklad: If we probe 4Mib, we'll probably hit the guard page already, as there's already something on the stack

cb22 (approved):

PR #3001, opened 2025-05-29
ci: limit fuzzer processes to 20GiB of RAM

matklad:
This is done in-process, rather than in cfo_supervisor via unsare, because in-process limiting is actually better! The reason why we need unsare is that we can't really reliably wait for children to finish, and _have_ to do that from the outside!

### DIFF
--- a/src/testing/fuzz.zig
+++ b/src/testing/fuzz.zig
@@ -1,10 +1,15 @@
 //! Utils functions for writing fuzzers.
 
+const builtin = @import("builtin");
 const std = @import("std");
 const stdx = @import("../stdx.zig");
 const assert = std.debug.assert;
 const PRNG = stdx.PRNG;
 const ratio = stdx.PRNG.ratio;
+
+const GiB = 1024 * 1024 * 1024;
+
+const log = std.log.scoped(.fuzz);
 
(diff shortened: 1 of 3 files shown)


matklad:
I want to land this a couple of days after https://github.com/tigerbeetle/tigerbeetle/pull/2999, just to change fewer variables at a time!

src/testing/fuzz.zig
```diff
@@ -217,3 +222,14 @@ pub fn ReadyQueueType(T: type) type {
         }
     };
 }
+
+pub fn limit_ram() void {
+    if (builtin.target.os.tag != .linux) return;
+
+    std.posix.setrlimit(.AS, .{
+        .cur = 20 * GiB,
+        .max = 20 * GiB,
```
matklad, line 231: 20 comes from CacheMap (I managed to get it to use slightly more than 16, so 20 gives some head room). 

I _think_ we should reduce this to 8/16GiB, by rewriting our fuzzers to do better coverage with smaller parameters. 

_But_ we do need that large of state space for the current implementation of CacheMap fuzzer, and I don't want to touch that until we remove the cache from CacheMap.

cb22 (approved):

PR #3008, opened 2025-06-01
compaction: finish one tree's bar quota before compacting the next

chaitanyabhandari:
In compaction, at the beginning of every bar, we iterate through each tree and pair of levels (for example Accounts tree and level 3 → 4) and calculate the number of input values to compact. During every _beat_ of the bar, we compact each tree and pair of levels by a fraction of the total input values. This ensures predictability by making sure that compaction does roughly the same amount of work every beat, but also comes with some caveats. Every beat, we:

* Suspend multiple compactions, and carry over their state to the subsequent beat
* Perform multiple instruction-cache context switches to different trees' merge functions, which are type-specialized
* Perform multiple data-cache context switches to different trees' compaction state

This PR changes the way that we schedule compactions. Now, we move on to a subsequent compaction only when the current compaction is run to completion, i.e. when its bar quota is exhausted. This involves pacing compaction using the _total bytes of input values across all trees_, as opposed to using the _number of input values_. This new way of compaction scheduling is not prone to the aforementioned caveats, but also affords us some new optimizations & code simplifications:

* _Memory efficiency_. **Forest compaction now reserves 18MiB of memory for its blocks as opposed to 76.5MiB**! Since we only carry over a single compaction's state to the next beat, we can carry over *more* state! So, we now carry over all the index & data blocks that we read during the current beat, to the next beat. Earlier, we used to discard these blocks and re-read them in the next beat (hopefully from cache, but there was no guarantee).
* We also got rid of some state in compaction, namely _wasted values_ (as a side effect of the above optimization), and also the logic to avoid over-reads at the end of the beat (since over-reads can simply be carried over to the next beat, now). 

---

#### 1-replica TB cluster (`i4i.16xlarge` EC2 instance)

`./tigerbeetle benchmark --query-count=0 --transfer-count=50_000_000` with our default grid cache size, _~10% increase in throughput, and also a reduction in p50 & p99 latencies_:
```C
load accepted = 182667 tx/s → 201422 tx/s
batch latency p1 = 15 ms → 15 ms
batch latency p50 = 41 ms → 34 ms
batch latency p99 = 184 ms → 177 ms
batch latency p100 = 194 ms → 197 ms
```

`./tigerbeetle benchmark --query-count=0 --transfer-count=50_000_000` with a 64MiB cache to simulate pressure on the grid cache, _~20% increase in throughput, and also a reduction in p50, p99 & p100 latencies_:
```C
load accepted = 164094 tx/s → 196907 tx/s
batch latency p1 = 15 ms → 15 ms
batch latency p50 = 46 ms → 35 ms
batch latency p99 = 184 ms →  175 ms
batch latency p100 = 203 ms → 191 ms
```

#### 6-replica TB cluster (`i4i.xlarge` EC2 instance)

`/tigerbeetle benchmark --query-count=0 --transfer-count=1_000_000_000 --clients=8`, _~3.5% increase in throughput, reduction in p1 & p50 latencies (p100 seems like an outlier here)_

```C
transfer batch size = 8189 txs
transfer batch delay = 0 us
load accepted = 117901 tx/s → 121101 tx/s 
batch latency p1   = 289 ms → 245 ms
batch latency p50  = 557 ms → 554 ms
batch latency p99  = 857 ms → 881 ms
batch latency p100 = 1310 ms → 3122 ms
```


chaitanyabhandari:
_Slightly_ better CPU utilization with this change as well, since we aren't context switching as much anymore.

cpu_usage_plot

src/constants.zig
```diff
@@ -658,8 +658,8 @@ comptime {
 // runtime-known number of free blocks.
 //
 // For simplicity for now, size IOPS to always be available.
-pub const lsm_compaction_queue_read_max = 8;
-pub const lsm_compaction_queue_write_max = 8;
+pub const lsm_compaction_queue_read_max = 16;
```
chaitanyabhandari, line 661: The motivation for increasing this was a back-of-the-envelope that I did to calculate how these constants are related to the amount of the write IO we drive to disk:

I did the calculations with 10ms latency for disk reads and writes, and 500us merge latency (based on some experiments that I'd done earlier). With these, and assuming a 50% grid cache hit rate, we need 16 blocks for reading input value blocks & 16 blocks for writing output value blocks (a total of ~16MiB as opposed to the old ~70MiB) to sustain ~1GiB/s write bandwidth to the disk.

src/lsm/compaction.zig
```diff
@@ -66,15 +66,16 @@ pub const compaction_tables_input_max = 1 + constants.lsm_growth_factor;
...
+/// The minimum number of blocks required for a single beat of a single compaction.
+///
+/// Counterintuitively, this number isn't 6 (one value and index block for the two input tables and
+/// one output table). Instead it is:
+/// One index and one value block for the output table, one index block for the two input tables,
+/// and `lsm_compaction_queue_read_max/2` value blocks for the two input tables.
+///
+/// The reason is that we carry over the output index block and all input blocks to the next beat.
+pub const compaction_block_count_beat_min: u32 =
+    (1 + 1) + (1 + 1) + constants.lsm_compaction_queue_read_max;
```
chaitanyabhandari, line 75: It _did_ feel a bit sad that a single compaction just can't run with 6-blocks worth of memory (which was the old constant)!

This is okay for production because we are much more memory efficient than we were before (we use ~18 MiB now as opposed to ~75MiB). *However*, one caveat of this increase is that the *minimum* memory that the tree fuzzer utilizes increases from 9MiB → 40MiB. Consequently, we now vary `block_count_min` from [block_count_min,  block_count_min * 16], instead of [block_count_min,  block_count_min * 64], to maintain a somewhat similar *maximum*.


```C
    const block_count_min =
        stdx.div_ceil(constants.lsm_levels, 2) * compaction_block_count_beat_min;

    const block_count = if (prng.chance(ratio(1, 5)))
        block_count_min
    else
        prng.range_inclusive(u32, block_count_min, block_count_min * 16);
```

src/lsm/compaction.zig
```diff
@@ -415,11 +404,12 @@ pub fn CompactionType(
         // Beat-scoped fields:
         // ------------------
         pool: ?*ResourcePool = null,
-        callback: ?*const fn (*ResourcePool) void = null,
+        callback: ?*const fn (*ResourcePool, u64) void = null,
 
         grid_reservation: ?Grid.Reservation = null,
         table_builder_value_block: ?*ResourcePool.Block = null,
 
+        values_in_beat: u32 = 0,
```
chaitanyabhandari: We need this because `counters` & `quotas` are _both_ bar scoped, and we need a beat-scoped variable to maintain the number of values that were consumed in a beat (CompactionSchedule needs it to adjust the input values while resuming the next compaction).

Although, now that I think about it, *perhaps* we can make `counters` beat-scoped, use it for the same purpose, while our current assertions will still assert equivalent behaviour? 🤔

src/lsm/compaction.zig
chaitanyabhandari: However, if this overhead seems a bit too high for the fuzzer, I was thinking that we *could* expose a boolean to control whether we want to carry-over blocks to the next beat or not, and then the fuzzer could opt out of this behavior? Seemed a bit overkill to me, what are your thoughts?

src/lsm/forest.zig
```diff
@@ -536,6 +538,10 @@ pub fn ForestType(comptime _Storage: type, comptime groove_cfg: anytype) type {
             // On the last beat of the bar, make sure that manifest log compaction is finished.
             if ((last_beat or last_half_beat) and op > constants.lsm_compaction_ops) {
                 forest.manifest_log.compact_end();
+                if (forest.manifest_log.grid_reservation) |grid_reservation| {
+                    forest.manifest_log.grid.forfeit(grid_reservation);
```
chaitanyabhandari: This (and the similar, compaction-related code above), is the caveat that I described in the PR write-up. Our free-set reservation-forfeiting is akin to two-phase locking, wherein once we forfeit one out of multiple active reservations, we can't reserve another one. So, I had to move this logic out from ManifestLog & Compaction to the forest.

chaitanyabhandari:
Looking good on DevHub too!

Screenshot 2025-06-16 at 9 59 14 AM

src/config.zig
```diff
@@ -124,8 +124,8 @@ const ConfigProcess = struct {
     clock_epoch_max_ms: u64 = 60000,
     clock_synchronization_window_min_ms: u64 = 2000,
     clock_synchronization_window_max_ms: u64 = 20000,
-    grid_iops_read_max: u64 = 16,
-    grid_iops_write_max: u64 = 16,
+    grid_iops_read_max: u64 = 32,
+    grid_iops_write_max: u64 = 32,
```
matklad, line 129: Intentional?

src/constants.zig
matklad: >I did the calculations with 10ms latency for disk reads and writes, a

Isn't it more like 1ms? 

https://github.com/sirupsen/napkin-math

Sequential SSD write, +fsync (8KiB) | 1 ms | 10 MiB/s | 100 ms | 2 min
-- | -- | -- | -- | --

matklad: Would be good to add napkin math formulas to the comment, to make it clear that this particular constant comes out of thicker air than usual!

chaitanyabhandari: New napkin math with latencies calculated after running a benchmark on `i4i.xlarge` machine on AWS:
* Read IO latency: 8ms
* Write IO latency: 2ms
* Merge latency: 75us

Assuming a cache hit rate of 50% & that compaction has 16 read IOPs available, consider this series of events:

* T=0ms
  * Read 4 level A & 4 level B value blocks from cache, 4 level A & 4 level B value blocks from disk
  * Total writes underway=0, Total reads underway=8
* T=0.3ms (75us * 4 for merging 4 pairs of value blocks)
  *  Start writing 4 value blocks to disk
  *  Read 2 level A & 2 level B value blocks from cache, 2 level A & 2 level B value blocks from disk
  * Total writes underway=4, Total reads underway=12
* T=0.45ms (75us * 2 for merging 2 pairs of value blocks)
  * Start writing 2 value blocks to disk
  * Read 1 level A & 1 level B value blocks from cache, 1 level A & 1 level B value blocks from disk
  * Total writes underway=6, Total reads underway=14
* T=0.525ms (75us * 1 for merging 2 pairs of value blocks)
  * Start writing 1 value block to disk, 
  * Read 1 level A value block from cache, 1 level B value block from disk 
  * Total writes underway=7, total reads underway=15
* T=2.525ms
  *  Finish writing 7 output blocks, wait for 15 reads from disk to finish
  * Total writes underway=0, total reads underway=15
* T=8ms
  * Finish reading 7 value blocks from level A, 7 value blocks from level B
  * Start read loop again
*  T=8.6ms
   *  Start writing 7 value blocks to disk
*  T=10.6ms
   *  Finish writing 7 value blocks to disk

_By T=10.6ms, we have finished writing 14 value blocks to disk. Therefore, we write ~1320 value blocks to disk within 1 second, which means we drive ~660MiB/s of write IO to the disk!_

chaitanyabhandari: @matklad based on my napkin math (apologies that it's a long one 🥲), it _looks_ like we're actually bottlenecked by the read IO latency, and that if we want to drive a 1GiB of write IO to the disk, we'd have to _further double_ the read IOPs (i.e. 4x what we have on main). 

Would love to get your feedback on the above napkin math though, specifically whether the cache hit rate of 50% seems reasonable or not, because calculation is heavily dependent on that!

chaitanyabhandari:
Debugged `./zig/zig build -Drelease vopr -- 6165036051955770325` and it isn't a legitimate liveness issue, the seed passes when ticks-max-convergence is increased from 10_000_000 → 10_010_000. It passes on main with fewer # of ticks, but that wouldn't be a fair comparison since the two runs aren't equivalent (new compaction scheduling that this PR introduces is a source of non-determinism):

```C
./zig/zig build -Drelease vopr -- --ticks-max-convergence=10_010_000 6165036051955770325 (this branch)

 0   \ .        14V 119/123/123C   0:123Jo  0/10J!   0:123Wo <__0:123> v1:1   169Ga  0G!   4G?

          PASSED (10072468 ticks)

./zig/zig build -Drelease vopr -- --ticks-max-convergence=10_010_000 6165036051955770325 (main)

 2   \   .     705V 159/164/164C 133:164Jo  0/_1J! 132:163Wo <__0:__0> v3:3   143Ga  0G!   0G?

          PASSED (2084937 ticks)
```

matklad:
> 10_000_000

Can we cut that 10x to 1_000_000 ^^ ? While unrelated, it might be useful to learn what takes so much time and optimize that, while we have a convenient repro on hand!

chaitanyabhandari:
Good idea @matklad, in this particular case I _think_ the root cause could be that the mean write latency is _quite_ high (~1 second), that'd slow things down a fair bit. 
```C
          write_latency_min=3
          write_latency_mean=96
          write_fault_probability=10/100
```

src/lsm/compaction.zig
matklad: In general, it's OK for fuzzers to use a lot of memory, so I wouldn't worry about that. We _might_ want to start pushing fuzz memory usage down in the future, but we are not there yet!

As long as we are not hitting this limit, we should be fine: https://github.com/tigerbeetle/tigerbeetle/blob/main/src/fuzz_tests.zig#L59

src/lsm/compaction.zig
```diff
@@ -66,15 +66,16 @@ pub const compaction_tables_input_max = 1 + constants.lsm_growth_factor;
...
-/// re-read in the next beat, and the compaction is paced in such a way as to end the beat only when
-/// an output value block is full.
-pub const compaction_block_count_bar_max: u32 = 1;
-
-// The minimum number of blocks required for a single beat of a single compaction:
-// one index and one value block for two input and one output table.
-pub const compaction_block_count_beat_min: u32 = (1 + 1) * 3;
+/// The minimum number of blocks required for a single beat of a single compaction.
+///
+/// Counterintuitively, this number isn't 6 (one value and index block for the two input tables and
```
matklad: It is counter-intuitive relative to the original implementation, it is rather intuitive for the new one, so we can drop this from the comment. And maybe re-order comment:

```
// Compaction needs to carry over all in-flight output and input blocks, hence the ....
```

src/lsm/forest.zig
```diff
@@ -808,28 +821,20 @@ fn CompactionScheduleType(comptime Forest: type, comptime Grid: type) type {
         pool: ResourcePool,
         next_tick: Grid.NextTick = undefined,
         callback: ?*const fn (*Forest) void = null,
+        bar_input_bytes_count: u64 = 0,
```
matklad: Assert state transitions for this field. I _think_ it should be zero at the end/start of the bar? 

Also, `_bytes_count`  is just `_size` in our naming conventions!

src/lsm/forest.zig
```diff
@@ -872,65 +877,74 @@ fn CompactionScheduleType(comptime Forest: type, comptime Grid: type) type {
...
                     inline for (comptime std.enums.values(Forest.TreeID)) |tree_id| {
-                        const resumed = self.compaction_at(level_b, tree_id)
-                            .compaction_dispatch_enter(&self.pool, beat_resume_callback);
+                        const tree = Forest.tree_info_for_id(tree_id);
+                        const Value = tree.Tree.Table.Value;
+                        const compaction = self.compaction_at(level_b, tree_id);
+
+                        const resumed = compaction.beat_commence(.{
+                            .callback = beat_resume_callback,
+                            .values_count = stdx.div_ceil(
```
matklad: Inconsistent API!

We pass value_count _in_, but then we'll get size back in the resume_callback. We should pick one for `compaction.zig` abstraction layer, and stick to it. I _think_ value could makes more sense there, so it's only compaction schedule who inter-converts between the two.

src/lsm/forest.zig
```diff
@@ -872,65 +877,74 @@ fn CompactionScheduleType(comptime Forest: type, comptime Grid: type) type {
...
+                            self.bar_input_bytes_count += (bar_input_values * @sizeOf(Value));
                         }
                     }
                 }
             }
 
+            const beats_total = half_bar;
+            const beats_done = compaction_beat % half_bar;
+            const beats_remaining = beats_total - beats_done;
+            self.beat_input_bytes_count = stdx.div_ceil(
```
matklad: Can assert that this is non-decreasing? We can only revise our estimates down!

src/lsm/forest.zig
matklad: Assert in `beat_commence` that `value_count > 0`?

Actually, I think we don't need this whole

```
            for (0..constants.lsm_levels) |level_b| {
                if (level_active(.{ .level_b = level_b, .op = op })) {
```

Loop anymore? We run _one_ compaction, so the forest can run it directly, instead of looping through all of them. And then, in `beat_commence`, we can simplify the signature and just assert that there's work to do, removing the `ceased` variant?

The original code was set-up to enable the parallel loop structure, thinking about running multiple compaction in parallel, but we no longer need/want that, so we can simplify stuff. 

Though, not _super_ sure whether that'll be a simplification or not! But worth thinking about more radical change here!

chaitanyabhandari: Mhmmmm, that's an approach that I'd considered initially, that we'd _only_ have the loop when we want to start a _new_ compaction (either when the bar starts, or when we have exhausted the beat quota). However, I'd decided not to do it since it involved storing extra state in CompactionSchedule, for the active tree ID & level B, which would then be used to fetch the compaction to run. 

This was when I was _first_ designing this, but I'm (hopefully) wiser now, so let me think about this once again to see if there's a potential simplification there...

src/lsm/forest.zig
chaitanyabhandari: Mhmmmmm didn't quite understand what you meant by this,  since `self.beat_input_size` should be `0` when `beat_start` is invoked. Included this in https://github.com/tigerbeetle/tigerbeetle/pull/3008/commits/9c0296b509d356531c70b3149f98d3213d3c8d26 but do let me know if you had something else in mind!

src/lsm/forest.zig
chaitanyabhandari: Made the `beat_commence` API consistent in https://github.com/tigerbeetle/tigerbeetle/pull/3008/commits/9c0296b509d356531c70b3149f98d3213d3c8d26, great catch there!

src/lsm/forest.zig
chaitanyabhandari: Good catch, added assertions for `bar_input_size` & `beat_input_size` in https://github.com/tigerbeetle/tigerbeetle/pull/3008/commits/9c0296b509d356531c70b3149f98d3213d3c8d26!

src/config.zig
matklad: @toziegler hath saith that 32 is the number, so this makes sense!

(roughly, SSDs have _a lot_ of internal concurrency, and we are quite under budget there).

But, just to double-check, if you _just_ increase `grid_iops_read_max`/`grid_iops_write_max`, without any other changes in this branch, how does that affect performance? 

When looking at the benchmark, I want to disentangle algorithmic changes from iop limit changes!

chaitanyabhandari: _Interestingly_, no effect on performance when I ran `./tigerbeetle benchmark --query-count=0 --transfer-count=50_000_00` on a single-replica `i4i.16xlarge`. Perhaps as Tobi said, we're already saturating the disk on those instances!

What's surprising to me though, is that the results differ _quite_ a bit from the experiments I'd ran on `i4i.16xlarge` (but a different VM) earlier! For example, throughput that I measured earlier on `main` was `182667 tps` but this time it was `116469 tps`. That's _huge_ variability. The only change here is that I've now switched to using spot instances to save cost, perhaps AWS penalizes performance there? Although, it could just be variability due to different hardware, but its quite huge... 🤔 

#### Grid IOPs = 16
```C
load accepted = 116469 tx/s
batch latency p1   = 39 ms
batch latency p50  = 66 ms
batch latency p99  = 208 ms
batch latency p100 = 327 ms
```

#### Grid IOPs increased from 16 → 32, Compaction IOPs increased from 8 → 16
```C
load accepted = 111943 tx/s
batch latency p1   = 40 ms
batch latency p50  = 69 ms
batch latency p99  = 208 ms
batch latency p100 = 332 ms
```

#### Compaction Scheduling optimizations + Grid IOPs increased from 16 → 32, Compaction IOPs increased from 8 → 16

```C
load accepted = 124627 tx/s
batch latency p1   = 39 ms
batch latency p50  = 59 ms
batch latency p99  = 204 ms
batch latency p100 = 433 ms
```

src/lsm/forest.zig
chaitanyabhandari: Thought about this a bit more and realized that I don't particularly like the current code from an understandability perspective! 

When we look at the compaction scheduling code, it is not immediately clear _what_ the compaction scheduling algorithm is. Specifically, it is non trivial to understand that we run a compaction to completion, before moving to the next compaction. Also, tricky to understand the forest _resumes_ the _suspended_ compaction in the next beat, we rely on the iteration order of `inline for (comptime std.enums.values(Forest.TreeID))` for that...

Ideally, I'd want the compaction scheduler to just call `beat_commence` on the _suspended_ compaction, but I haven't yet been able to come up with a clean way to do that. It looks like to start a new compaction (either at the beginning of the bar or during a beat when the current one is exhausted), we need a way to fetch the _next_ compaction. This could either be the next active level within a tree, or the first active level within the next tree! But this sounds awfully similar to the iteration logic we have already... so, I find myself stumped. 😅 
```C
      for (0..constants.lsm_levels) |level_b| {
	      if (level_active(.{ .level_b = level_b, .op = op })) {
		      inline for (comptime std.enums.values(Forest.TreeID)) |tree_id| {
```

src/lsm/compaction.zig
```diff
@@ -350,17 +339,14 @@ pub fn CompactionType(
         /// `dropped` values can vary between the replicas.
         ///
         /// Counters obey accounting equation of compaction:
-        ///     out = in - (dropped + wasted)
+        ///     out = in - dropped
```
matklad, line 342: I think the counters are now deterministic, now that we don't have wasted? If that's true, comment needs updating!

src/lsm/compaction.zig
```diff
@@ -415,11 +401,12 @@ pub fn CompactionType(
         // Beat-scoped fields:
         // ------------------
         pool: ?*ResourcePool = null,
-        callback: ?*const fn (*ResourcePool) void = null,
+        callback: ?*const fn (*ResourcePool, u16, u64) void = null,
```
matklad: Specify names for u16 and u64 parameters, as a help for the reader.

src/lsm/compaction.zig
```diff
@@ -648,6 +647,15 @@ pub fn CompactionType(
                 .beat = 0,
                 .bar = quota_bar,
             };
+
+            log.debug("{s}:{}: bar_commence: " ++
```
matklad: will fit on a single line

src/lsm/compaction.zig
```diff
@@ -819,43 +831,61 @@ pub fn CompactionType(
...
+        /// The entry point to the actual compaction work for the beat. Called by the forest.
         pub fn beat_commence(
             compaction: *Compaction,
-            beats_remaining: u64,
-        ) void {
+            options: struct {
+                callback: *const fn (*ResourcePool, u16, u64) void,
+                values_count: u64,
+            },
+        ) enum { active, ceased } {
```
matklad: Let's use naming convention from replica.zig commit protocol:

* `pending` --- I need to do some asynchronous work, I've arranged to call your callback some time in the future
* `ready` --- I've done all the required work synchronously, callback won't be called.

src/lsm/compaction.zig
matklad: Hm, no, actually something else's going on here. 

I _think_ we are still not fully deterministic, as replicas can read ahead by different amounts, so the comment is broadly correct. 

_But_ the way we actually update counters does make them deterministic: 

```
fn merge_callback() {
            compaction.quotas.done += consumed_total;
            compaction.counters.in += consumed_total;
            compaction.counters.dropped += merge_result.dropped;
            compaction.values_in_beat += consumed_total;
}
```

This doesn't make much sense. The thinking between separating quotas from countries was that quotas are updated logically, on merge, while counters are updated eagerly, when we do actual IO (in read_block_callback). That way, we can _assert_ that the two are consistent, without this being just a trivial functional dependency. 

But here, I think we make them exactly equal? 

I'd rather we keep updating counters in read_callback, if we _are_ non-deterministic with respect to reads!

src/lsm/compaction.zig
```diff
@@ -1513,7 +1447,6 @@ pub fn CompactionType(
             assert(block.stage == .read_value_block);
             stdx.copy_disjoint(.exact, u8, block.ptr, value_block);
             block.stage = .read_value_block_done;
-            compaction.counters.in += Table.data_block_values_used(block.ptr).len;
```
matklad: This one I'd rather keep (re the counters/quotes discussion above)

src/lsm/compaction.zig
matklad: counters are bar scoped, but quotas already carry both bar and beat!

Let's use this for quotas (bar_done is the old done, beat_done is values_in_beat)

```
            beat: u64 = 0,
            beat_done: u64 = 0,
            bar: u64 = 0,
            bar_done: u64 = 0,
```

src/lsm/manifest_log.zig
```diff
@@ -851,9 +851,8 @@ pub fn ManifestLogType(comptime Storage: type) type {
             assert(manifest_log.read_callback == null);
             assert(manifest_log.write_callback == null);
 
-            if (manifest_log.grid_reservation) |grid_reservation| {
-                manifest_log.grid.forfeit(grid_reservation);
-                manifest_log.grid_reservation = null;
+            if (manifest_log.grid_reservation) |_| {
+                // Grid reservation is is forfeited by the forest.
```
matklad:
```suggestion
                // Grid reservation is is forfeited by the forest because ...
```

Always explain Y!

matklad (commented):
LGTM overall, left some more comments inline!

Could you also add a documentation commit to CompactionSchedule explaining the schedule?

src/lsm/compaction.zig
```diff
@@ -819,43 +831,61 @@ pub fn CompactionType(
...
-        /// - determine our quota in terms of minimum number of input values to process,
-        /// - reserve grid blocks for the output.
+        /// The entry point to the actual compaction work for the beat. Called by the forest.
         pub fn beat_commence(
             compaction: *Compaction,
-            beats_remaining: u64,
-        ) void {
+            options: struct {
+                callback: *const fn (*ResourcePool, u16, u64) void,
+                values_count: u64,
```
matklad: Oh, nit, callback last:

```
values_count: u64
callback:
```

src/lsm/compaction.zig
```diff
@@ -876,22 +906,8 @@ pub fn CompactionType(
...
-            log.debug("{s}:{}: beat_commence: quota_done={} quota_beat={} quota_bar={} " ++
-                "blocks_reserved={} values_wasted={}", .{
-                compaction.tree.config.name,
-                compaction.level_b,
-                compaction.quotas.done,
-                compaction.quotas.beat,
-                compaction.quotas.bar,
-                beat_blocks_max,
-                compaction.counters.wasted,
-            });
```
matklad, line 894: It is _soooo_ nice to get rid of this!

src/lsm/compaction.zig
chaitanyabhandari: Ah, great suggestion, much cleaner this way! Thank you. :)

chaitanyabhandari:
Thanks for the great comments @matklad, most of them addressed in https://github.com/tigerbeetle/tigerbeetle/pull/3008/commits/f1e50aeb37426cccbf400fcc621d3d193851b7ef, with the exception of the comment for CompactionSchedule, which I'll take up as a separate commit. 😄

src/lsm/compaction.zig
chaitanyabhandari: Mhmmmmm, I thought about this for a bit and realized that I didn't _fully_ understand the `counters` field... Specifically, I didn't properly understand the _non-determinism_ aspect of it!

IIUC, updating `compaction.counters.in` in `read_block_callback` as opposed to `merge_callback` would be tricky: we may read more blocks than we actually merge in a particular beat, and the ones that aren't merged are just carried over to the next beat. So, if we want to update `compaction.counters.in` in `read_block_callback`, we would have to maintain another field (akin to `wasted` but named differently since we don't actually "waste" any reads now), where we would also have to maintain which values we carried over to the next beat... 

Thoughts @matklad?

chaitanyabhandari: Now that I think about it, embarrassingly enough, I feel like I don't _fully_ understand how our compaction is deterministic across replicas! 

Specifically I was trying to reason about a scenario where the read for value blocks V1, V2 from level B finishes in different orders on replica R1 & R2, but read of value blocks V3, V4 from level B finish in the same order. In that case, couldn't R1 & R2 construct construct _different_ output value blocks: R1 is merging (V1, V3) and R2 is merging (V2, V3), and both of them output a different V' that they write to the same address?

Surely I am missing something fundamental here. 😅

chaitanyabhandari:
@matklad I took a stab at removing the `beat_table_done` status and collapsing it into `beat_quota_done`, and carrying over unfinished value blocks to the next beat, in https://github.com/tigerbeetle/tigerbeetle/pull/3008/commits/6ffed5deffb2f4cd63a7ad2a7ea363a9dd27949d. I _think_ it has simplified things further:
* Reduces # of states
* Logic for over-reads is not required anymore because as soon as we reach `beat_quota_done`, we stop reading value blocks and simply wait for inflight IO to finish
* Logic to write value & index blocks in `compaction_dispatch_beat_quota_done` is simpler, since it now only caters to writing value & index blocks if the bar quota is complete (as opposed to making sure we always write a value block every beat)

However, one caveat with this change is that the compaction counters _may_ not be consistent at the end of a beat, they are only consistent at the end of a bar, which is not very interesting since it doesn't track non-determinism. This is because we may do _merge_ work (which updates `counters.in`) within a beat, but actually _write_ the value block (which updates `counters.out`) in the _next_ beat. Also, as discussed in https://github.com/tigerbeetle/tigerbeetle/pull/3008#discussion_r2178692798, moving the update of `counters.in` to `read_block_callback` would require adding back another parameter to track what we're carrying over to the next beat.

I prefer simplification of the code flow over maintaining compaction counters, so one option could be to remove those counters. The other option is to add another parameter to track what we're carrying over to the next beat (similar to `wasted`), but that seems complex to me... 

Would love to hear your thoughts about this!

src/lsm/forest.zig
chaitanyabhandari: Didn't find a clean enough way to get rid of this loop and make the compaction schedule more obvious *just* by reading the code... but we have a comment to explain the schedule now: https://github.com/tigerbeetle/tigerbeetle/pull/3008/commits/5a1689e5e8bea5e888bc0cf6add12b25c5aad5dd!

matklad:
>However, one caveat with this change is that the compaction counters may not be consistent at the end of a beat, they are only consistent at the end of a bar, which is not very interesting since it doesn't track non-determinism. 

On the contrary, it is _very_ interesting! Counters give us a very powerful on-line verification that the entire compaction worked correctly. Internally, compaction is a rather complicated flow. But if you directly count every value that gets in from disk, and every value that gets out to disk, and the numbers of consistent, you have a very strong guarantee that the complicated middle part didn't mess things up! This is beautiful --- we have this very complex ball of asynchrony, and we can check that it worked correctly by just tallying up the water that it sucks in and then lets out!

So, I think we should keep the counters, but, yes, it is _crucial_ that we count things as soon as they leave disk (in read_callback), and not at any later time. If we do it later, the counters won't detect a bug where we redundantly read a data block, for example. 

Now, it's a choice whether we want to assert consistency of counters between beats, or only at the end of the bar. I _think_ doing it only at the bar boundary is fine. But we can do this at the beat as well! The "missing" values are exactly those in `level_X_value_blocks`!

So I think we can even strengthen asserts a bit here?

```diff
diff --git a/src/lsm/compaction.zig b/src/lsm/compaction.zig
index afd3627dd..fb5b72297 100644
--- a/src/lsm/compaction.zig
+++ b/src/lsm/compaction.zig
@@ -690,6 +690,9 @@ pub fn CompactionType(
 
                 return 0;
             } else {
+                if (compaction.table_info_a == .immutable) {
+                    compaction.counters.in == compaction.table_info_a.?.immutable.len;
+                }
                 return compaction.quotas.bar;
             }
         }
@@ -813,6 +816,20 @@ pub fn CompactionType(
                 assert(manifest_added_value_count == manifest_removed_value_count);
                 assert(manifest_added_value_count > 0);
             } else {
+                // Cross-check against each over:
+                // - the number of values we should have processed as per original manifest.
+                // - the number of values we processed, accoring to manifest changes.
+                // - individually counted processed values.
+                var values_planned: u64 = switch (compaction.table_info_a.?) {
+                    .immutable => |values| values.len,
+                    .disk => |table_info| table_info.table_info.value_count,
+                };
+                for (compaction.range_b.?.tables.const_slice()) |table_info| {
+                    values_planned += table_info.table_info.value_count;
+                }
+                assert(values_planned == compaction.counters.in);
+                assert(values_planned == manifest_removed_value_count);
+
                 assert(manifest_added_value_count == compaction.counters.out);
                 assert(manifest_removed_value_count == compaction.counters.in);
                 assert(manifest_removed_value_count - manifest_added_value_count ==
@@ -934,8 +951,19 @@ pub fn CompactionType(
             }
 
             assert(compaction.block_queues_empty_output());
-            // We may be carrying over some input blocks to the next beat.
-            maybe(compaction.block_queues_empty_input());
+            var values_in_flight: u64 = 0;
+            for (.{
+                compaction.level_a_value_block.iterator(),
+                compaction.level_b_value_block.iterator(),
+            }) |it| {
+                while (it.next()) |block| {
+                    values_in_flight += Table.data_block_values_used(block.ptr).len;
+                }
+            }
+            values_in_flight -= compaction.level_a_position.value;
+            values_in_flight -= compaction.level_b_position.value;
+            assert(compaction.counters.out + values_in_flight ==
+                compaction.counters.in - compaction.counters.dropped);
 
             assert(compaction.pool.?.idle());
             maybe(compaction.pool.?.blocks_acquired() > 0);
@@ -1446,6 +1474,7 @@ pub fn CompactionType(
             assert(block.stage == .read_value_block);
             stdx.copy_disjoint(.exact, u8, block.ptr, value_block);
             block.stage = .read_value_block_done;
+            compaction.counters.in += Table.data_block_values_used(block.ptr).len;
             compaction.compaction_dispatch();
         }
 
@@ -1555,8 +1584,6 @@ pub fn CompactionType(
 
             compaction.quotas.bar_done += consumed_total;
             compaction.quotas.beat_done += consumed_total;
-            compaction.counters.in += consumed_total;
-
             compaction.counters.dropped += merge_result.dropped;
 
             assert(compaction.quotas.bar_done <= compaction.quotas.bar);
```

src/lsm/compaction.zig
chaitanyabhandari: We add a value block to our queue when we _start_ reading it, not when we _finish_ reading it, and that's the key!

And the order in which we queue value block addresses up into the queue is deterministic across replicas because the order of value block addresses within the index blocks is of course the same.

src/lsm/compaction.zig
chaitanyabhandari: Fixed in https://github.com/tigerbeetle/tigerbeetle/pull/3008/commits/d3bccacf3f3f0ddf6aa0a1b46682cbb28c22b7ca.

sentientwaffle:
Removed cfo labels since cfo is now running zig 14, while this branch is still on zig 13.

matklad:
re added the label as the branch is, in fact, on zig 14 already!

src/lsm/forest.zig
```diff
@@ -443,7 +441,7 @@ pub fn ForestType(comptime _Storage: type, comptime groove_cfg: anytype) type {
             // balances out, because we expect to naturally do less other compaction work on the
             // last beat.
             // The first bar has no manifest compaction.
-            if ((last_beat or last_half_beat) and op > constants.lsm_compaction_ops) {
```
matklad, line 446: Hm, why `op > constants.lsm_compaction_ops` condition is gone? It handles "The first bar has no manifest compaction." case

chaitanyabhandari: Ah, I was meaning to add a GitHub comment on that change but forgot it! I removed it because it is already taken care of in the ManifestLog:
https://github.com/tigerbeetle/tigerbeetle/blob/5bd9395e47db2ee6465fda91b0633e9ac149ecaf/src/lsm/manifest_log.zig#L688-L694

I spent some time thinking about whether `last_beat or last_half_beat` condition also belongs _within_ ManifestLog, such that it is responsible for pacing its own compaction (as we do pass `op` to `compact`), and the forest should just call `forest.manifest_log.compact`... but decided that it belongs in a separate PR.

src/lsm/forest.zig
```diff
@@ -915,8 +907,34 @@ fn CompactionScheduleType(comptime Forest: type, comptime Grid: type) type {
             const beats_total = half_bar;
             const beats_done = compaction_beat % half_bar;
             const beats_remaining = beats_total - beats_done;
+
             self.beat_input_size = stdx.div_ceil(self.bar_input_size, beats_remaining);
 
+            // This is akin to a dry run for the actual compaction work that is going to happen
```
matklad, line 924: FWIW, in my head `commence` by itself has exactly this semantics, of doing prep before the actual work starts. And commence happens "in parallel". The originator of this naming is I believe dj, with

```
            inline for (std.meta.fields(Grooves)) |field| {
                @field(forest.grooves, field.name).open_commence(&forest.manifest_log);
            }

            forest.manifest_log.open(manifest_log_open_event, manifest_log_open_callback);

```

which follows the same pattern (we notify _all_ trees that we are beginning the opening procedure. Once we notified all trees, we start feeding them with individual events from the manifest in an interleaved order)

src/lsm/forest.zig
```diff
@@ -915,8 +907,34 @@ fn CompactionScheduleType(comptime Forest: type, comptime Grid: type) type {
...
+                            compaction.beat_commence(
+                                stdx.div_ceil(beat_input_size, @sizeOf(Value)),
+                            );
+
+                            beat_input_size -|= (compaction.quotas.beat * @sizeOf(Value));
+                        }
+                    }
+                }
+                assert(beat_input_size == 0);
+            }
```
matklad, line 968: Yeah, this feels much better, and I think matches the characteristics of the reservation we had before. It is still tantalizing that this is not _quite_ optimal though 😉 

We are running compactions sequentially, so if we do just a single reservation, per beat, which is then shared across compactions, we'll avoid even more fragmentation (the reservation then can be stuffed into the resource pool I think?). 

Alternatively, maybe we can do one-at-a-time reservation, and then just relax the asserts in the reservations? I _think_ it is ok if manifest holds onto a reservation, while we are creating and forfeiting one more for compactions? 

Though, between the two I'd probably pick one per-beat shared reservation!

matklad: https://mmapped.blog/posts/29-plan-execute -- related post

src/lsm/forest.zig
chaitanyabhandari: Ah yes, great pattern indeed! Love that we started with this and came back to it... what goes around comes back around. :)

src/lsm/forest.zig
chaitanyabhandari: Mhmmmmmm, one big reservation per beat _does_ sound better to reduce fragmentation, let me try it out! *Should* be straightforward with the block calculation living in CompactionScheduler, as opposed to within Compaction...

Also, I realized that I don't *deeply* understand the reasoning behind our reservation asserts, and *why* we don't allow for the acquire, acquire, release, acquire pattern! I should look at it a bit more closely to understand.

src/lsm/forest.zig
```diff
@@ -911,31 +969,55 @@ fn CompactionScheduleType(comptime Forest: type, comptime Grid: type) type {
...
+                },
+            }
+
             self.beat_resume();
         }
 
         fn beat_finish(self: *CompactionSchedule) void {
             assert(self.callback != null);
+            assert(self.beat_input_size == 0);
+            if (self.grid_reservation) |reservation| {
```
matklad: This can assert that the reservation is != null, no?

src/lsm/compaction.zig
```diff
@@ -819,174 +886,136 @@ pub fn CompactionType(
...
 
         // While beat_commence is called by the forest sequentially for each compaction, to get
         // deterministic grid reservations, each compaction completes its own beat's work
         // asynchronously
         fn beat_complete(compaction: *Compaction) void {
-            assert(compaction.stage == .beat_table_done);
+            assert(compaction.stage == .beat_quota_done);
             switch (compaction.table_builder.state) {
-                .no_blocks => {},
+                .no_blocks, .index_and_data_block => {},
```
matklad: Can assert that `data_block` is not full? This is actually tricky! If a block gets full right at the beat boundary, we can write it out either in the current beat or in the next beat. It doesn't matter which we choose, but it's impertative is that the choice is the same across replicas!

src/lsm/compaction.zig
```diff
@@ -819,174 +886,136 @@ pub fn CompactionType(
...
-            assert(compaction.quotas.done <= compaction.quotas.beat);
+            assert(compaction.block_queues_empty_output());
+            // We may be carrying over some blocks from the previous beat.
+            maybe(compaction.block_queues_empty_input());
+
+            if (compaction.move_table) assert(compaction.quotas.bar_exhausted());
+
+            // Run the compaction up to completion of the bar quota, if possible.
+            const values_remaining = (compaction.quotas.bar - compaction.quotas.bar_done);
+            const quota_beat = @min(values_count, values_remaining);
```
matklad:
```suggestion
            compaction.quotas.beat = @min(values_count, values_remaining);
```

to avoid introducing an alias

src/lsm/compaction.zig
```diff
@@ -1250,22 +1288,12 @@ pub fn CompactionType(
                     value_block.stage = .free;
                     compaction.pool.?.block_release(value_block);
                 } else {
-                    if (!compaction.output_blocks.full()) {
-                        const write = compaction.pool.?.writes.acquire().?;
```
matklad, line 1254: TBH, I like how we have a separate statement for resource acquisition: 

```
const write = compaction.pool.?.writes.acquire().?;
```


If that `.?` fires in production, it would be helpful if it points at a specific line!

In general, my rule-of-thumb is that anything that's not pure function, and can fail or side-effect, is worth extracting into a statement of its own! (so, perhaps it's even worth doing the same for address!)

src/lsm/compaction.zig
```diff
@@ -1274,89 +1302,53 @@ pub fn CompactionType(
...
+                assert(compaction.table_builder.state == .no_blocks);
             }
 
             if (compaction.output_blocks.count > 0) {
                 return;
             }
+
             switch (compaction.table_builder.state) {
-                .no_blocks => {},
+                .no_blocks, .index_and_data_block => {},
```
matklad: Same here, assert that if we have a data block, it is not full

src/lsm/compaction.zig
```diff
@@ -1619,13 +1593,14 @@ pub fn CompactionType(
             assert(compaction.level_b_position.value <= Table.data.value_count_max);
             assert(compaction.table_builder.value_count <= Table.data.value_count_max);
 
-            if (compaction.table_info_a.? == .immutable) {
-                compaction.counters.in += merge_result.consumed_a;
-            }
+            const consumed_total = merge_result.consumed_a + merge_result.consumed_b;
```
matklad: alt naming:

```suggestion
            const consumed_ab = merge_result.consumed_a + merge_result.consumed_b;
```

src/lsm/compaction.zig
```diff
@@ -1744,13 +1730,24 @@ pub fn CompactionType(
...
+                    const data_blocks_count =
+                        index_schema.data_blocks_used(index_block.ptr);
+                    if (compaction.level_b_position.value_block == data_blocks_count) {
                         compaction.level_b_position.index_block += 1;
                         compaction.level_b_position.value_block = 0;
+
+                        const popped = compaction.level_b_index_block.pop().?;
+                        assert(popped == index_block);
+                        compaction.read_value_block_release_table(index_block.ptr);
+                        index_block.stage = .free;
```
matklad, line 1775: Hm, do we _need_ to change the logic around index blocks? Previously, an index block was freed immediately once we scheduled the read of the last value block therein, and the last block was tagged with `last_block_in_the_table` so that we can advance position. 

It was done that way to not hang onto the index block longer than usual. Is something changed with new pacing that there's a good reason to hang onto the block? Given that we have only _one_ index block slot `level_b_index_block` available, this means that our level_b pipeline drains out at the boundary between the tables, as we can only start reading the next table once the previous one is fully merged!

chaitanyabhandari: Mhmmmm, great question, I had actually forgotten about _why_ I had this changed this logic! Took me a while to recall it but I finally remember now. 

This change is required for our new way of calculating which value block to read next to work. The new way doesn't rely on compaction-level state for `level_a_value_block_next`, `level_b_value_block_next`, but it also requires us to enforce a crucial invariant: _Value block queues can only contain value blocks for a single index block_

The reason for this is that we calculate the next block to read using the below. If we pop the index block when we issue the read for the final index block, and we update `compaction.level_a_position.value_block` to `0`, `level_a_value_block_next` will be computed as an incorrect index! It is important that we only set `compaction.level_a_position.value_block` to `0` when the value block queue is drained... so, we have to couple popping of the index block with the popping of the last value block from the queue!

```C
                const level_a_value_block_next =
                    compaction.level_a_position.value_block +
                    @as(u32, @intCast(compaction.level_a_value_block.count));
```

src/lsm/compaction.zig
chaitanyabhandari: The rule-of-thumb makes a lot of sense to me! Indeed, it'd be better to understand which line the assertion failure comes from, I will separate the statement out.

src/lsm/forest.zig
chaitanyabhandari: There's a code path in `beat_start` → `beat_finish` where we don't create a grid reservation!
```C
        pub fn beat_start(self: *CompactionSchedule, callback: Forest.Callback, op: u64) void {
            assert(self.pool.idle());
            assert(self.callback == null);
            assert(self.beat_input_size == 0);

            self.callback = callback;
            if (op < constants.lsm_compaction_ops or
                self.grid.superblock.working.vsr_state.op_compacted(op))
            {
                self.beat_finish();
                return;
            }
```

src/lsm/compaction.zig
chaitanyabhandari: Mhmmmmm my understanding was actually that if during the the final merge, the value block gets full *and* the beat quota is fulfilled, then we are sure to write out the value block in the _next_ beat. This is because we mark the state as `beat_quota_done` in `merge_advance_position`, and _then_ when we invoke `compaction_dispatch` we go straight to `compaction_dispatch_beat_quota_done` and wait for the currently pending IO to complete (without issuing a write for the block that currently got full).

So, at the end of beat, the data block could be either full or not, no? 🤔

src/lsm/compaction.zig
chaitanyabhandari: Added a comment for why we do this: https://github.com/tigerbeetle/tigerbeetle/pull/3008/commits/1c97e35d7901c65e0d2ac97c757f779df2fe475d

I _think_ we could increase the number of `level_b_index_block` slots (as opposed to adding back code for maintaining `last_block_in_the_table`, `level_a_value_block_next`, `level_b_value_block_next`, `level_a_index_block_next`, `level_b_index_block_next`), as the current way reduces dimensionality... Thoughts @matklad?

src/lsm/compaction.zig
matklad: If _that_ is the case, we can assert that the value block is empty iff we exhausted the bar quota? It doesn't matter much _which_ beat we do this in, but we should assert ether way. 

Although, doing it in the earlier bit feels more natural, just not to get data get cold in the ram.

src/lsm/compaction.zig
matklad: Hm, I'd maybe try to get by with a single index block in-flight! We don't _really_ need two, I would think, as tables are pretty big, and we could be reading the next index block while we are finishing reading the previous value blocks. Saves us 0.5MiB of RAM!

But I don't think it's too important, adding one extra block seems fine by me (or maybe we don't even need an extra block because switching between tables is rare? Needs benchmarking!)

On the other hand, we already had the code to do that, and it _should_ still work, right? Like, there's nothing in the new pacing that _requires_ us to compute the indexes differently? 

So, I'd maybe spend some more cycles thinking about how we can get by without an extra block here, but it's your call whether it is worth it!

chaitanyabhandari: > On the other hand, we already had the code to do that, and it should still work, right? Like, there's nothing in the new pacing that requires us to compute the indexes differently?

Mhmmmmm, no this is not a side effect of the new compaction pacing, but rather a _separate_ refactor that attempts to remove last_block_in_the_table, level_a_value_block_next, level_b_value_block_next, level_a_index_block_next, level_b_index_block_next from compaction! 

> So, I'd maybe spend some more cycles thinking about how we can get by without an extra block here, but it's your call whether it is worth it!

Right, my understanding was also that switching between tables is rare, so removing the optimization where we read the index table concurrently with merging the last value block of the table _shouldn't_ be too harmful. However, I had a different plan in mind to achieve something similar -- we _could_ read in all index blocks at _beginning_ of the bar, and just store the _value block addresses + checksums_ instead of the complete index blocks. Was thinking about picking this separately from this PR though!

chaitanyabhandari: This one was based on a discussion had with Fed a while back, wherein he said that index blocks are quite sparse, and so we _could_ make it so that we use a _single block's worth of memory_ and store the value block addresses + checksums for the entire bar directly! I still need to calculate the upper bound on the memory that we may need for all the value block addresses + checksums, though.

src/lsm/compaction.zig
chaitanyabhandari: Agreed that writing it earlier feels better in case the data block gets full... however, the case where we fulfill beat quota _and_ fill up the data block to the brim should be rare, no? Happy to reconsider if you feel strongly about flushing the data block at the end of the beat though, though it did feel good to get rid of some complexity in `compaction_dispatch_beat_quota_done` using the reasoning that the above case feels rare. :D 

I have tightened the asserts around table builder state at the end of the beat in https://github.com/tigerbeetle/tigerbeetle/pull/3008/commits/254e1367098d0cbaf7275272bb92f29b96dc0667, do let me know if it looks good to you.

cb22:
(temporarily removing the fuzz labels here to give more cycles to main + release!)

src/lsm/compaction.zig
```diff
@@ -819,174 +886,137 @@ pub fn CompactionType(
...
                 assert(compaction.table_builder.state == .no_blocks);
+                assert(compaction.table_builder_value_block == null);
                 assert(compaction.table_builder_index_block == null);
+                assert(compaction.block_queues_empty_input());
             }
 
-            compaction.stage = .paused;
             const pool = compaction.pool.?;
             const callback = compaction.callback.?;
-            compaction.pool = null;
```
matklad, line 941: Hm, the original idea was to keep compaction.pool scoped to the beat, so that pool is null outside of compaction actually doing the work. But I guess we can make it bar scoped as well.

We should assert that pool is null in assert_between_bars then. (previously, we were asserting that pool is null between beats in fn idle).

Though, I do have a slight preference for the design where pool is beat-scoped, and where `grid_reservation` is a field on the pool, such that `pool` encapsulates all the resources that Forest handles down to an individual compaction.

src/lsm/compaction.zig
matklad: Hm, I think the code as written has a pretty nasty non-determinism bug, at least in theory!

Consider this sequence of events:

* We run the merge_callback, which fills _exactly_ a value block, but doesn't exhaust quota
* We write, or _not_ write these two blocks based on `if (compaction.output_blocks.spare_capacity() >= 2) {`
* We run another merge_callback, which now does exhaust the beat quota. But, a twist! The stars are _so_ perfectly aligned, that this new merge callback perfectly annihilates all the input values (so, all values are dropped)
* So we enter the `compaction_dispatch_beat_quota_done` loop, where we _don't_ write out the table builder (unless we are also at the bar quota, but assume we aren't)

This sort of thing is why it's worthwhile to be extremely pedantic about all state transitions and leant towards nailing them as precisely as possible, even if that means writing more code (c.f. how we've lost assertion that `pool` is null)

src/lsm/compaction.zig
chaitanyabhandari: Good idea regarding moving `grid_reservation` into `pool`, I do recall you mentioned it before but it slipped my mind amidst all our discussions, apologies! Done in https://github.com/tigerbeetle/tigerbeetle/pull/3008/commits/6103db128974fb2a49448ad78cbd10bee9133c54.

src/lsm/compaction.zig
chaitanyabhandari: Mhmmmm the sequence of events that you describe _certainly_ seems possible... Consequently, we _could_ have two replicas where one of them writes this value block at the _end_ of the compaction of op=X and the other one writes it at the _beginning_ of the compaction of op=X+1. So, the storage on the two replicas at the end of op=X would _indeed_ be different... However, I'm trying to think of a case where this would be problem. Specifically, do we rely on the fact that our storage is perfectly identical at the end of _each_ beat? IIUC, we would still converge to the same storage state at the end of the half-bar (at the latest, and best earlier whenever this compaction's bar quota is finished), no? Or perhaps I am missing something in the scenario that you described, does it somehow lead to a case where two replicas irrecoverably diverge? 🤔

chaitanyabhandari:
New numbers for scale test on a realistic TB cluster! Since the cache here is very large (24 GiB), the optimization in this PR are not _as_ effective. The reason is that our optimizations to carry over all index & value blocks to the next beat benefit performance if the compaction on _main_ goes to the _disk_ to re-read those blocks. In case of a large cache, that effect becomes diminished since most blocks that compaction needs (and has read in the previous beat), are likely to be found in cache! 

Essentially, this PR gives bigger performance gains on _smaller_ TB deployments, which have a smaller cache. For example, I imagine it being helpful for the small serverless tiers!

#### 6-replica TB cluster (`i4i.xlarge` EC2 instance)

`/tigerbeetle benchmark --query-count=0 --transfer-count=1_000_000_000 --clients=8`, with a 24GiB grid cache:  _~3.5% increase in throughput, reduction in p1 & p50 latencies (p100 seems like an outlier here)_

```C
transfer batch size = 8189 txs
transfer batch delay = 0 us
load accepted = 117901 tx/s → 121101 tx/s 
batch latency p1   = 289 ms → 245 ms
batch latency p50  = 557 ms → 554 ms
batch latency p99  = 857 ms → 881 ms
batch latency p100 = 1310 ms → 3122 ms
```

src/lsm/compaction.zig
matklad: I think in that case the respective blocks might end up assigned different addresses? Eg, imagine that the latter beat is also the one that runs manifest log compaction, so, for the replica which ends up _not_ writing the block in the earlier beat, the address it _would_ have written the block to would be allocated to manifest log instead.

chaitanyabhandari: AH INDEED! Oh _gosh_, this is a particularly nasty one... What a phenomenal catch @matklad.

I am thinking about how to solve this. One thing that immediately comes to mind is per-bar grid reservation for compaction? Let me try to give it some more thought... 

_Or_ simply make sure that at the end of the beat in `compaction_dispatch_beat_quota_done`, we check if we have any full value/index blocks on hand and we make sure we flush them before we move on to the next beat.

chaitanyabhandari: _Or_ simply make sure that at the end of the beat in `compaction_dispatch_beat_quota_done`, we check if we have any full value/index blocks on hand and we make sure we flush them before we move on to the next beat.

chaitanyabhandari: Mhmmmmm, I realized that per-bar grid reservation would _also_ be bad for fragmentation _unless_ we also have the ManifestLog compaction blocks included in that reservation... If we don't, then at the first beat of the half-bar we reserve a big chunk of forest compaction blocks, and then in the _last_ beat of the half-bar, we reserve a smaller number of ManifestLog compaction blocks. Wasted blocks in the _forest compaction grid reservation_ would lead to fragmentation in our grid block allocations.

I actually spent some time implementing a single, bar-scoped grid reservation for _both_ the ManifestLog & Forest compactions, because the prospect of minimizing fragmentation even further was just so alluring! But then embarrassingly realized while testing that its not possible _at all_, ManifestLog & Forest sharing a compaction, because they run concurrently and would contend for the same addresses. 🤦🏼

chaitanyabhandari: @matklad we always write full value & index blocks at the end of the beat now, making that decision deterministic across replicas: https://github.com/tigerbeetle/tigerbeetle/pull/3008/commits/ab8042fe8f284665438bc5efff10531e28f48a26

I realized that another to solve this _could_ have been to making sure we invoke forest compaction _before_ ManifestLog compaction, which would ensure that the grid reservations in the next beat would include the free addresses from the previous beat. So, we would make sure a value block that is written in op=X on one replica, and op=X+1 on another replica always gets the same address... but I guess that's a bit too fragile & subtle a logic to rely on.

chaitanyabhandari:
Also had another look at the memory footprint since I realized I never quantified it... _Forest compaction now reserves `18MiB` of memory for its blocks as opposed to `76.5MiB` before!_

matklad (approved):
❤️

PR #3011, opened 2025-06-03
devhub: track how efficient is the VOPR

matklad:
It's important to notice if, for some reason, VOPR becomes 10x slower (or faster). We already have all the info to compute that, only need a bit of GUI:

* take the count from seed record for the main branch
* divide by the time elapsed since the seed was tested

For stability wait for at least 100 seeds

### DIFF
--- a/src/devhub/devhub.js
+++ b/src/devhub/devhub.js
@@ -91,7 +91,7 @@ async function main_seeds() {
   const table_dom = document.querySelector("#seeds>tbody");
   let commit_previous = undefined;
   let seed_fail_count = 0;
-  let vpm = undefined; // VOPRS per minute
+  let vpm = undefined; // VOPRs per minute
 
   for (const record of records) {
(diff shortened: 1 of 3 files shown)


matklad:
image

src/devhub/devhub.js
```diff
@@ -91,8 +91,16 @@ async function main_seeds() {
   const table_dom = document.querySelector("#seeds>tbody");
   let commit_previous = undefined;
   let seed_fail_count = 0;
+  let vpm = undefined; // VOPRS per minute
```
cb22:
```suggestion
  let vpm = undefined; // VOPRs per minute
```

:smile:

cb22 (dismissed):

fabioarnold:
image

A little speedometer would be great. 😁

cb22 (approved):

PR #3012, opened 2025-06-04
replica: add primary and status information to all log lines

cb22:
When glancing at server logs, it's really helpful to know which replica is the primary. This can already be figured out by knowing the topology and the view number, but this makes it much quicker.

Additionally, include the status if it's not `.normal`.

(I briefly considered making replica have a `self.log.info` & co, but figured this was more in keeping with the existing spirit...)

### DIFF
--- a/src/vsr/replica.zig
+++ b/src/vsr/replica.zig
@@ -176,6 +176,36 @@ pub fn ReplicaType(
             replica: *Replica,
         };
 
+        const LogPrefix = struct {
+            replica: u8,
+            status: Status,
+            primary: bool,
+
+            pub fn format(
+                self: LogPrefix,
+                comptime fmt: []const u8,
+                options: std.fmt.FormatOptions,
+                writer: anytype,
+            ) !void {
+                _ = fmt;
+                _ = options;
+                try writer.print("{}", .{self.replica});
+
+                var status_character: u8 = switch (self.status) {
+                    .normal => 'n',
+                    .view_change => 'v',
+                    .recovering => 'r',
+                    .recovering_head => 'h',
+                };
+
+                if (self.primary) {
+                    status_character = std.ascii.toUpper(status_character);
+                }
+
+                try writer.print("{c}", .{status_character});
+            }
+        };
+
         /// We use this allocator during open/init and then disable it.
         /// An accidental dynamic allocation after open/init will cause an assertion failure.
         static_allocator: StaticAllocator,
@@ -660,7 +690,7 @@ pub fn ReplicaType(
             const replica_count = self.superblock.working.vsr_state.replica_count;
             if (replica >= options.node_count or replica_count > options.node_count) {
                 log.err("{}: open: no address for replica (replica_count={} node_count={})", .{
-                    replica,
+                    self.log_prefix(),
                     replica_count,
                     options.node_count,
                 });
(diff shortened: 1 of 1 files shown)


src/vsr/replica.zig
```diff
@@ -169,6 +169,29 @@ pub fn ReplicaType(
             replica: *Replica,
         };
 
+        const ReplicaLogPrefix = struct {
+            replica: *Replica,
```
sentientwaffle: :thinking: Could we use `@fieldParentPtr` instead or would that not work?

src/vsr/replica.zig
```diff
@@ -169,6 +169,29 @@ pub fn ReplicaType(
...
+                _ = options;
+                try writer.print("{}", .{self.replica.replica});
+
+                if (self.replica.primary_index(self.replica.view) == self.replica.replica) {
+                    try writer.print(" (primary)", .{});
+                }
+
+                if (self.replica.status != .normal) {
+                    try writer.print(" ({s})", .{@tagName(self.replica.status)});
+                }
```
sentientwaffle: Can we take a cue from the vopr and encode status as a single character, and is-primary as another character? e.g.

normal primary:
```
5N!
```

or view-change backup:
```
5V_
```

(or something similar).

These prefixes will be part of debug logs for VOPR as well, so being compact is very useful.

src/vsr/replica.zig
cb22: I could only get it to work if `ReplicaLogPrefix` is not zero sized, otherwise `self` is undefined.

sentientwaffle: Maybe just use a dummy padding variable then? Seems nice to avoid the self-pointer anyhow.

sentientwaffle: Another possibility is to add a replica method `log_prefix` which returns `[2]u8` and format it with `{s}`.

src/vsr/replica.zig
cb22: Tied this in with `log_prefix()` above, so now, it uses `n` / `v` / `r` / `h` for the states, in lowercase if a backup and uppercase if a primary.

src/vsr/replica.zig
```diff
@@ -10885,6 +10978,28 @@ pub fn ReplicaType(
                 (tables_index_block_count + tables_value_block_count +
                     self.state_machine.forest.manifest_log.log_block_checksums.count));
         }
+
+        pub fn log_prefix(self: *const Replica) [2]u8 {
+            var log_message: [2]u8 = undefined;
+
+            log_message[0] = switch (self.replica) {
+                0...9 => 48 + self.replica,
+                else => unreachable,
```
sentientwaffle: Ah wait, I think this is not unreachable -- standbys mean the max replica index is actually 12.

cb22: Hmm ok, think I'll make `log_prefix` return a struct with `format`, so we don't have to waste a character each time to add the extra info.

src/vsr/replica.zig
```diff
@@ -2723,9 +2766,9 @@ pub fn ReplicaType(
 
             // Otherwise, cancel in progress commit and prepare to sync.
             log.mark.debug(
-                \\{}: on_start_view_set_checkpoint: sync started view={} checkpoint={}..{}
+                \\{s}: on_start_view_set_checkpoint: sync started view={} checkpoint={}..{}
```
```suggestion
                \\{}: on_start_view_set_checkpoint: sync started view={} checkpoint={}..{}
```

sentientwaffle (dismissed):

src/vsr/replica.zig
cb22: Fixed and rebased.

sentientwaffle (approved):

Issue #3007, opened 2025-06-01
Type annotations in Python client


stenczelt:
# summary

The python client does not comply with static type checkers like `mypy`, making `tigerbeetle` a headache as a dependency in downstream packages requiring passing a type checker. 

Let's type-annotate the client and ship like that, or provide an accompanying package of stub files at least.

# current behaviour

On version `0.16.42`. 

with `main.py`
```python
import tigerbeetle

print(tigerbeetle.Account(id=1, ledger=1, code=1))
```

we have

```
$ mypy main.py         
main.py:1: error: Skipping analyzing "tigerbeetle": module is installed, but missing library stubs or py.typed marker  [import-untyped]
main.py:1: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
```

# expected behaviour

Success criteria:
- Running `mypy` on a program importing `tigerbeetle` passes
- CI enforces `mypy` passing on `src/clients/python/src/tigerbeetle` with `--strict` flag enabled _(good to have)_

# proposed solution & questions

One of 
1. Type-annotate the package and publish with a `py.typed` marker
2. create a separate `tigerbeetle_stubs` package which can be installed from PyPi

I'd prefer (1), so there is only one dependency to be managed.

This is easy enough to do for `client.py` and `lib.py`, but I am not sure how to do for `bindings.py` which says it's auto-generated from `tb_client_header.zig`. How is this generated, do we have to make the generator produce type-annotated code? Or can we maybe just write a stub for `bindings.py` to be shipped that together with the rest of the code? In the latter case, having the CI enforce `mypy` passing on the code should require developers to keep the `bindings.pyi` stub up to date if they change `tb_client_header.zig`.

stenczelt:
may relate to #2925 as well

cb22:
Hi @stenczelt 

Yes, having proper typing natively on our Python client would be ideal!

The contents for `bindings.py` are generated by `python_bindings.zig`. It's not the easiest code to read, since it's Zig generating Python, so let me know if you have any questions there. I think we'd prefer to have the bindings generate the types too, just to minimize the chance of things getting out of sync.

(If you feel like putting up a PR even for just `client.py` and `lib.py`, that'd be a great starting point!)

stenczelt:
Great, thanks for the review and merging #3014, I see it's included in the `0.16.56` release now.

PR #3014, opened 2025-06-05
Typed python client

stenczelt:
# summary 

Type annotations for the python client, for #3007. The client code is now passing `mypy` in "strict" mode.

# details

Updated the hand-written python client code, and the zig code generating the client bindings as well. Both are type annotated now, with backwards compatibility up to Python 3.7. I've also added `mypy` to `src/clients/python/ci.zig` so need to continue passing this when running the CI.

# questions

## CI running on commits
I've edited the github actions settings so the CI is ran on my own branches. How to correctly set this up so it's running on all my commits without this making it into the upstream for you?

## enum defaults in dataclasses

The bindings include dataclasses with `IntEnum` attributes. These are created with default 0, but the type checker realises this is looser than the `IntEnum` class defined. I've noticed that for all enums exposed, 0 is `.OK` so I've used that - this would break if you write/expose other enums where 0 is not `OK`. **What do you want/recommend?**
- inspect the enum and find what 0 corresponds to and print that? (how to do this in Zig?)
- don't add a default
the latter I'd only do for enums, sine for flags, and general numbers 0 seems to typically mean NONE/no-filtering.

see https://github.com/stenczelt/tigerbeetle/blob/6b909ae1e02194f96c4e027dc74da4d75930ddc3/src/clients/python/python_bindings.zig#L320-L325

and 

https://github.com/stenczelt/tigerbeetle/blob/6b909ae1e02194f96c4e027dc74da4d75930ddc3/src/clients/python/src/tigerbeetle/bindings.py#L246-L249

## ~decorators~

~I've removed a few decorators, e.g. from the log handler, which to my understanding was not actually doing anything. The tests still passed. Please check if you agree with this, or let me know if there is some behaviour expected under the hood that I've not noticed (also let's add a test for these if so).~

### DIFF
--- a/src/clients/python/ci.zig
+++ b/src/clients/python/ci.zig
@@ -68,6 +68,9 @@ pub fn tests(shell: *Shell, gpa: std.mem.Allocator) !void {
         try shell.env.put("TB_ADDRESS", tmp_beetle.port_str.slice());
         try shell.exec("python3 main.py", .{});
     }
+
+    // We are checking type annotations of the entire package.
+    try shell.exec("python3 -m mypy . --strict", .{});
 }
 
 pub fn validate_release(shell: *Shell, gpa: std.mem.Allocator, options: struct {
--- a/src/clients/python/src/tigerbeetle/client.py
+++ b/src/clients/python/src/tigerbeetle/client.py
@@ -4,11 +4,16 @@ import asyncio
 import ctypes
 import logging
 import os
+import sys
 import threading
 import time
 from collections.abc import Callable  # noqa: TCH003
 from dataclasses import dataclass
 from typing import Any
+if sys.version_info >= (3, 11):
+    from typing import Self
+else:
+    from typing_extensions import Self
 
 from . import bindings
 from .lib import tb_assert, c_uint128
@@ -17,11 +22,11 @@ logger = logging.getLogger("tigerbeetle")
 
 
 class AtomicInteger:
-    def __init__(self, value=0):
+    def __init__(self, value: int = 0) -> None:
         self._value = value
         self._lock = threading.Lock()
 
-    def increment(self):
+    def increment(self) -> int:
         with self._lock:
             self._value += 1
             return self._value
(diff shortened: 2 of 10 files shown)


stenczelt:
hey @cb22, this is ready for review now! I've built the package locally and were satisfied with it.

matklad:
@stenczelt just a heads up, we have https://systemsdistributed.com conference this week, so it might take longer than usual to get to the proper review!

stenczelt:
@cb22 / @matklad I've brought this up to date with the latest changes on `main` and got it passing tests, and checked compatibility with python 3.7 to 3.13. From my side this is ready for a review again.

src/clients/python/ci.zig
```diff
@@ -68,6 +68,9 @@ pub fn tests(shell: *Shell, gpa: std.mem.Allocator) !void {
         try shell.env.put("TB_ADDRESS", tmp_beetle.port_str.slice());
         try shell.exec("python3 main.py", .{});
     }
+
+    // test type annotations of the entire package
```
cb22: As part of [TigerStyle](https://github.com/tigerbeetle/tigerbeetle/blob/main/docs/TIGER_STYLE.md) please could you make all comments full sentences / grammatically correct.

src/clients/python/src/tigerbeetle/client.py
```diff
@@ -61,7 +66,8 @@ def id() -> int:
     if time_ms <= time_ms_last:
         time_ms = time_ms_last
     else:
-        id._time_ms_last = time_ms
+        # fixme: this is a very non-pythonic solution, and discouraged
```
cb22: Can't remember why I did this as opposed to declaring it as a global... Might have been performance, or not wanting to pollute the namespace when using `from ... import *`.

Anyhow, feel free to fix it!

src/tidy.zig
```diff
@@ -743,11 +743,11 @@ test "tidy no large blobs" {
...
-        .{".bat"},     .{".c"},   .{".cs"},    .{".csproj"}, .{".css"},  .{".go"},
-        .{".h"},       .{".hcl"}, .{".html"},  .{".java"},   .{".js"},   .{".json"},
-        .{".md"},      .{".mod"}, .{".props"}, .{".ps1"},    .{".py"},   .{".rs"},
-        .{".service"}, .{".sln"}, .{".sum"},   .{".svg"},    .{".toml"}, .{".ts"},
-        .{".txt"},     .{".xml"}, .{".yml"},   .{".zig"},    .{".zon"},
+        .{".bat"},     .{".c"},     .{".cs"},    .{".csproj"}, .{".css"},  .{".go"},
+        .{".h"},       .{".hcl"},   .{".html"},  .{".java"},   .{".js"},   .{".json"},
+        .{".md"},      .{".mod"},   .{".props"}, .{".ps1"},    .{".py"},   .{".rs"},
+        .{".service"}, .{".sln"},   .{".sum"},   .{".svg"},    .{".toml"}, .{".ts"},
+        .{".txt"},     .{".typed"}, .{".xml"},   .{".yml"},    .{".zig"},  .{".zon"},
```
cb22: You can remove `.typed` from here, since it's specified explicitly below.

src/clients/python/src/tigerbeetle/client.py
```diff
@@ -156,7 +162,8 @@ def _c_on_completion(completion_ctx, packet, timestamp, bytes_ptr, len_):
 
         if packet[0].status != bindings.PacketStatus.OK.value:
             inflight_packet.response = PacketError(repr(bindings.PacketStatus(packet[0].status)))
-            tb_assert(inflight_packet.on_completion is not None)
+            if inflight_packet.on_completion is None:
```
cb22, line 177: What was the reason behind changing the `tb_assert`?

cb22 (commented):
Thanks @stenczelt - and sorry about the delayed review! This looks great, just a few nits and questions.

src/clients/python/src/tigerbeetle/client.py
stenczelt: `mypy` does not accept that `inflight_packet.on_completion` is surely `not None` after `tb_assert(inflight_packet.on_completion is not None)` so complains about the usage of it on the next line. :(

src/tidy.zig
stenczelt: very well, done!

src/clients/python/ci.zig
stenczelt: ok, done!

src/clients/python/src/tigerbeetle/client.py
stenczelt: I've made it into using a class under the hood, same interface as before. Plus added test cases for it.

cb22 (approved):

cb22:
Big thanks @stenczelt! These are great to have.

